PR #6463, opened 2020-01-14
find_program: Always use USERPROFILE instead of HOME

nirbheek:
On MSYS2 and MSYS, Python reads HOME instead of USERPROFILE, which
gets the path wrong.

Serves me right for not writing a test!!


xclaesse:
LGTM, once CI passed.

Issue #5888, opened 2019-09-05
run_tests.py: windows ninja >= 1.9 not detected


scivision:
On Windows 10 (and not on Linux) I notice the new run_tests.py Ninja >= 1.9 detection didn't seem to detect that I do indeed have the ninja 1.9 binary. I have manually checked that the run_tests.py 
`ninja_cmd = detect_ninja(v)` is indeed version 1.9.0 when running the command below via a `print(ninja_cmd)` statement added.

This issue is only happening when running Meson self-tests, and I didn't attempt to debug further. The print(ninja_cmd) only outputs once. 

Here's an example using latest master  caec875fe   (i'm assuming it probably existed since added to run_tests.py)

```
python .\run_project_tests.py --only common                                                                                                                                                                
C:\msys64\mingw64\bin\ninja.EXE   # <== due to my added print(ninja_cmd), only outputs once


c      : cc (gcc 9.2.0 "cc (Rev1, Built by MSYS2 project) 9.2.0")
cpp    : c++ (gcc 9.2.0 "c++ (Rev1, Built by MSYS2 project) 9.2.0")
cs     : not found
cuda   : nvcc (nvcc 10.1.243)
d      : not found
fortran: gfortran (gcc 9.2.0 "GNU Fortran (Rev1, Built by MSYS2 project) 9.2.0")
java   : not found
objc   : cc (gcc 9.2.0)
objcpp : c++ (gcc 9.2.0)
rust   : not found
swift  : not found
vala   : not found

Checking that configuring works...
Checking that building works...
Checking that testing works...
Checking that installing works...
cc: fatal error: cannot execute 'cc1obj': CreateProcess: No such file or directory                                      compilation terminated.
c++: fatal error: cannot execute 'cc1objplus': CreateProcess: No such file or directory                                 compilation terminated.

Running common tests.

Don't have ninja >= 1.9, enabling timestamp resolution workaround
Don't have ninja >= 1.9, enabling timestamp resolution workaround
Don't have ninja >= 1.9, enabling timestamp resolution workaround
Don't have ninja >= 1.9, enabling timestamp resolution workaround
Don't have ninja >= 1.9, enabling timestamp resolution workaround
Don't have ninja >= 1.9, enabling timestamp resolution workaround
Don't have ninja >= 1.9, enabling timestamp resolution workaround
Don't have ninja >= 1.9, enabling timestamp resolution workaround
Don't have ninja >= 1.9, enabling timestamp resolution workaround
[...]

PR #6465, opened 2020-01-14
run_tests: Fix detection of ninja 1.9

nirbheek:
`get_backend_commands()` doesn't get called when we run tests as
subprocesses, so detect ninja on import. This should speed up CI.

Fixes https://github.com/mesonbuild/meson/issues/5888


nirbheek:
Before and after Windows CI times:
before-after-meson-ci

PR #6468, opened 2020-01-15
Update Users.md

amitdo:
Remove "(not the default yet)" from GLib, GStreamer, Orc and Pango


amitdo:
Meson is also the solo build system for GStreamer, but currently only in its master branch.

tp-m:
> Meson is also the default build system for GStreamer, but currently only in its master branch.

It's also the recommended build system for the stable branch (1.16), even if it still ships with an autotools build in 1.16 too, so I think it's ok to update it for GStreamer as well.

amitdo:
You're right.

https://gstreamer.freedesktop.org/releases/1.16/

>The Meson build is now feature-complete (\*) and it is now the recommended build system on all platforms. The Autotools build is scheduled to be removed in the next cycle.

I'll update the PR soon.

amitdo:
Done.

Issue #6470, opened 2020-01-15
meson 0.53.0 py_installation.dependency('dbusmock') leads to: ERROR: Function does not take positional arguments.


DimStar77:
**Describe the bug**
The error was detected while building folks (reported there too, but presumably this is a meson issue):
https://gitlab.gnome.org/GNOME/folks/issues/119

**To Reproduce**
A minimalistic meson.build could be:

```
project('folks', [ 'vala', 'c' ],
  version: '0.13.1',
  license: 'LGPL2.1+',
  meson_version: '>= 0.49',
)

pymod = import('python')
py_installation = pymod.find_installation('python3')
python_dbusmock = py_installation.dependency('dbusmock')
```

**Expected behavior**
Like earlier meson version: pass

**system parameters**
Tested on openSUSE Tumbleweed; last version in use was meson 0.51.2, where it works. Latest to be tested is 0.53.0 - where it fails

ailin-nemui:
probable fix: delete this line https://github.com/mesonbuild/meson/blob/master/mesonbuild/modules/python.py#L351

DimStar77:
> probable fix: delete this line

fixes folks at least; I'm just building a whole gnome stack to see if something new would pop up

xclaesse:
git blame points the finger to https://github.com/mesonbuild/meson/pull/6341. The fix is indeed to remove @noPosargs, if someone wants to open a PR.

mensinda:
The thing is, that adding any positional arguments doesn't do anything here since `args` is never used inside the function. Also, the dependency name is hardcoded to `python` [here](https://github.com/mesonbuild/meson/blob/master/mesonbuild/modules/python.py#L48). So it does make semantically sense to add `@noPosargs` here unless I have fundamentally missed something.

xclaesse:
@mensinda Oops, you're right. The doc is misleading, it means that it has the same keyword arguments than dependency(), not same positional arguments. The name of the dependency is hardcoded to "python" and you cannot change that.

mensinda:
So, to further clarify, passing an argument to `py_installation.dependency()` should be an error, since this method will only ever return a `python` dependency (for embedding python itself, etc.) and **not** `dbusmock` or anything else.

That being said, I am not sure if it makes sense to just "break" existing meson projects (although they are arguably already broken). So it might make sense to print a warning like:

```dependency does not take any positional arguments.
It always returns a dependency for python itself.
This will become a hard error in the future.
```

[...]

PR #6471, opened 2020-01-15
python: Fix ERROR: Function does not take positional arguments

DimStar77:
Fixes issue #6470


mensinda:
Just removing `@noPosargs` doesn't seem to be the right solution to me. Instead, there should be st least a warning that adding positional arguments doesn't do anything and that this will be a hard error in the future.

See [my other comment](https://github.com/mesonbuild/meson/issues/6470#issuecomment-574725443) for more detail.

Issue #1608, opened 2017-04-11
Arbitrary overriding of binaries


jpakkane:
Originally from #763.

Coming back to this, let's think through what the requirements here are. I can think of the following (with an example of what it could look like):

-  find an executable from the system: find_program('bob')
- find an executable that comes from the system or subproject: find_program('bob', fallback : ['fooproj', 'bob_exe']
- find an executable that is overridden in cross compile: find_program('objcopy'), which would look up cross file and return armhf-something-objcopy unless native : true is defined

Any other cases we need to handle?

nirbheek:
As a note, we also want the fallback to be invoked if the version doesn't match too, see #1609.

But the biggest blocker for GStreamer on Windows right now is that we need a way to be able to pass executables from a subproject as replacements for programs in `PATH` for the GNOME module. Basically, if we have glib as a subproject, we want to use the `glib-mkenums` and `glib-compile-resources`, etc, provided by it.

ldrumm:
I was discussing basically this issue on IRC, looking for advice. I've implemented a basic simple PoC on
https://github.com/ldrumm/meson/tree/find_program-fallback with an example repository of its usage at https://github.com/ldrumm/meson-find-program-fallback.git

I'd be happy to take this on given guidance on design (particularly re the cross-compiling scenarios)

xclaesse:
Please open a PR so we can do proper review comments.

We can have syntax like `find_program('foo', fallback : 'foo_subproject')`. When doing the fallback, it would simply configure the subproject, then assume the subproject called override_find_program(). This avoid having to pass a variable name too. This is what I'm doing with dependency() too in https://github.com/mesonbuild/meson/pull/6203.

amyspark:
This is affecting pkg-config usage in Windows, if it's provided through a wrap + `override_find_program` the `dependency` call won't use it. The only workaround is to disable pkg-config altogether through the environment variable (or a native or cross-file).

xclaesse:
I believe this issue can be closed, we have automatic fallback to subprojects for a while now.

pkg-config is special, the problem is the find_program logic that does fallbacks is in the interpreter and PKGConfigDependency object does not have access to the interpreter when it needs to find pkg-config.

bonzini:
For pkgconfig the best solution really would be to embed a Python rewrite in Meson itself. Considering the extra complication of potentially recursive invocation of pkg-config (dependency -> find_program -> subproject -> dependency), let's just make this a documentation issue.

PR #6472, opened 2020-01-15
[WIP] Rough example of find_program(..., fallback)

ldrumm:
This PR is to start discussion on what needs to be done for https://github.com/mesonbuild/meson/issues/1608

### DIFF
--- a/mesonbuild/interpreter.py
+++ b/mesonbuild/interpreter.py
@@ -2342,7 +2342,12 @@ permitted_kwargs = {'add_global_arguments': {'language', 'native'},
                                            'variables',
                                            },
                     'executable': build.known_exe_kwargs,
-                    'find_program': {'required', 'native', 'version', 'dirs'},
+                    'find_program': {'required',
+                                     'native',
+                                     'version',
+                                     'dirs',
+                                     'fallback',
+                                     },
                     'generator': {'arguments',
                                   'output',
                                   'depends',
@@ -3467,11 +3472,12 @@ external dependencies (including libraries) must go to "dependencies".''')
     # the host machine.
     def find_program_impl(self, args, for_machine: MachineChoice = MachineChoice.HOST,
                           required=True, silent=True, wanted='', search_dirs=None,
-                          version_func=None):
+                          version_func=None, fallback=None):
         args = mesonlib.listify(args)
 
         extra_info = []
-        progobj = self.program_lookup(args, for_machine, required, search_dirs, extra_info)
+        progobj = self.program_lookup(
+                args, for_machine, required, search_dirs, extra_info, fallback)
         if progobj is None:
             progobj = self.notfound_program(args)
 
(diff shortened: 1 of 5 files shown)


mesonbuild/interpreter.py
```diff
@@ -3095,6 +3112,9 @@ def find_program_impl(self, args, for_machine: MachineChoice = MachineChoice.HOS
                 return ExternalProgramHolder(dependencies.NonExistingExternalProgram())
         return progobj
 
+
+
+    @FeatureNewKwargs('find_program', '???', ['fallback'])
```
ignatenkobrain: __[Flake8]__ too many blank lines (3) ([see on Sider](https://sider.review/gh/repos/19784232/pulls/6472#issue-5064127))

<details>

| Rule |
| - |
| E303 |

</details>
<sub>posted by Sider</sub>

mesonbuild/interpreter.py
```diff
@@ -825,6 +825,11 @@ class ExecutableHolder(BuildTargetHolder):
     def __init__(self, target, interp):
         super().__init__(target, interp)
 
+class ExecutableDependencyHolder(ExecutableHolder):
+    def found(self):
+        return True
+
```
xclaesse: not used

mesonbuild/build.py
```diff
@@ -1317,8 +1317,8 @@ def __init__(self, args, kwargs):
         exe = args[0]
         if hasattr(exe, 'held_object'):
             exe = exe.held_object
-        if not isinstance(exe, (Executable, dependencies.ExternalProgram)):
-            raise InvalidArguments('First generator argument must be an executable.')
+#       if not isinstance(exe, (Executable, dependencies.ExternalProgram, ExecutableDependency)):
+#           raise InvalidArguments('First generator argument must be an executable.')
```
xclaesse: why?

mesonbuild/interpreter.py
```diff
@@ -3076,6 +3083,16 @@ def find_program_impl(self, args, for_machine: MachineChoice = MachineChoice.HOS
...
             progobj = ExternalProgramHolder(prog)
+        if progobj is None and fallback is not None:
+            kwargs = {
+                'required': required,
+                'silent': silent,
+                'wanted': wanted,
+                'search_dirs': search_dirs,
+                'fallback': fallback,
+            }
+            progobj = self.dependency_fallback(args[0], kwargs)
```
xclaesse: dependency_fallback() assumes the variable you take from the subproject is a dependency object. This should fail if you're trying to take an executable/program object.

mesonbuild/interpreter.py
xclaesse: 0.54.0 for now.

mesonbuild/interpreter.py
xclaesse: You'll need to write a function almost exactly the same, but with messages tweaked to speak about programs instead of dependency, and do not call self.get_subproject_dep() but instead just something like `self.subprojects[dirname].get_variable_method([varname], {})` and verify it is an Executable or ExternalProgramHolder.

mesonbuild/interpreter.py
```diff
@@ -2089,7 +2094,7 @@ def get_cross_property_method(self, args, kwargs):
                                            'version',
                                            },
                     'executable': build.known_exe_kwargs,
-                    'find_program': {'required', 'native', 'version', 'dirs'},
+                    'find_program': {'required', 'native', 'version', 'dirs', 'fallback'},
```
xclaesse: It will also need `default_options`.

mesonbuild/interpreter.py
```diff
@@ -3114,7 +3134,7 @@ def func_find_program(self, node, args, kwargs):
         for_machine = self.machine_from_native_kwarg(kwargs)
         return self.find_program_impl(args, for_machine, required=required,
                                       silent=False, wanted=wanted,
-                                      search_dirs=search_dirs)
+                                      search_dirs=search_dirs, fallback=kwargs.get('fallback', None))
```
xclaesse: I would pass the whole kwargs, that you can pass down to dependency_fallback() instead of creating a new kwarg dict there, because you'll need to also pass down `default_options` for example.

mesonbuild/interpreter.py
```diff
@@ -3168,10 +3188,13 @@ def get_subproject_dep(self, display_name, dirname, varname, kwargs):
         except InvalidArguments:
             pass
 
-        if not isinstance(dep, DependencyHolder):
+        if not isinstance(dep, (ExecutableHolder, DependencyHolder)):
```
xclaesse: You can't accept both. When doing a dependency() fallback it must accept only DependencyHolder, and when doing a find_program() fallback it must accept only ExecutableHolder (and ExternalProgramHolder too I guess, but it would be weird to fallback to find a system program...).

xclaesse: But I agree it's better to add support for both code paths here than duplicating all the code.

mesonbuild/interpreter.py
xclaesse: Sorry, saw below that you hacked get_subproject_dep() too. But you still need to differentiate both code paths, hopefully without duplicating too much code.

xclaesse: Oh, and you need the program is found, but the version does not match (the case `if not is_found` below), you also need to try to fallback too.

ldrumm:
@xclaesse Thanks for the feedback. Clearly most of the code's pretty rough (WIP), but if you're happy with the general idea, I'm going to take another pass at this with tests and documentation. It also needs to be rebased.

tristan957:
Closing since we will have `find_program()` fallback in 1.3.0 with version comparisons and default options.

PR #6474, opened 2020-01-16
RPM macros: Add *_no_vpath macros.

besser82:
Some projects may need to be build in different configurations from the same source tree.  Having macros with a hardcoded vpath makes that rather inconvenient, as one would need reduplicate and update those macros by hand in the spec file.

Having reusable macros at hand makes it easy, for meson developers, as they just need to update the macros at the known location, and for package maintainers, as they can use the *_no_vpath macros and simply add their target build-dir for each different build config from the same source-tree.

### DIFF
--- a/data/macros.meson
+++ b/data/macros.meson
@@ -8,7 +8,7 @@
         if [ -n "$ncpus_max" ] && [ "$ncpus_max" -gt 0 ] && [ "$MESON_BUILD_NCPUS" -gt "$ncpus_max" ]; then MESON_BUILD_NCPUS="$ncpus_max"; fi; \\\
         if [ "$MESON_BUILD_NCPUS" -gt 1 ]; then echo "--num-processes $MESON_BUILD_NCPUS"; fi)
 
-%meson \
+%meson_no_vpath \
     %set_build_flags \
     %{shrink:%{__meson} \
         --buildtype=plain \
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
Umm, wait, doesn't Meson _mandate_ out of tree builds? And this change doesn't make sense anyway, since you could redefine the `%_vpath_*` macros to be in-tree if you wanted anyway...

besser82:
> Umm, wait, doesn't Meson _mandate_ out of tree builds? And this change doesn't make sense anyway, since you could redefine the `%_vpath_*` macros to be in-tree if you wanted anyway...

I'm not about doing in-tree builds here.  Let me explain a bit, what I am targeting with this:

Let's say you have a package `libfoo`.  This lib can build in several configurations.  One of the configurations builds a `libfoo.so.x` and another builds a `libfoo.so.y`.

As `libfoo` is a common system library, one wants to build to build both SO-versions, to have `libfoo.so.x` in the distribution repos for pre-built legacy software, while using `libfoo.so.y` for the software packaged for the distro.

The easiest way to build both SO-versions is to run the build from the same source in two different build-trees:

```
%meson_no_vpath $src $build-so_x
%meson_no_vpath $src $build-so_y
%meson_build_no_vpath -C $build-so_x
%meson_build_no_vpath -C $build-so_y
…
```

AFAIK, rpm does not support redefining of macros, as they are sourced *before* parsing the spec file.

Does that make sense to you, @Conan-Kudo ?

Conan-Kudo:
> AFAIK, rpm does not support redefining of macros, as they are sourced before parsing the spec file.

It does if you use `%define`. Each `%define` is evaluated lazily (e.g. when that line in the spec is being executed).

besser82:
> > AFAIK, rpm does not support redefining of macros, as they are sourced before parsing the spec file.
> 
> It does if you use `%define`. Each `%define` is evaluated lazily (i.e. when that line in the spec is being executed).

Does that even apply within the same section of a spec file?

```
…
%build
%define %vpath_builddir build_so_x
%meson
%meson_build
%define %vpath_builddir build_so_y
%meson
%meson_build
…
```

I'm not sure whether that will be guaranteed to work…

Conan-Kudo:
Yes, it's supposed to.

besser82:
> Yes, it's supposed to.

Undocumented "feature" or well defined and documented behaviour?  I'm tending to think it's the first…

Conan-Kudo:
[From rpm.org](https://rpm.org/user_doc/macros.html#builtin-macros):
> Note that %define and %global differ in more ways than just scope: the body of a %define’d macro is lazily expanded (ie when used), but the body of %global is expanded at definition time. It’s possible to use %%-escaping to force lazy expansion of %global.

Though I still think you're probably solving this problem the wrong way.

ignatenkobrain:
Use define or global instead.

besser82:
> It does if you use `%define`. Each `%define` is evaluated lazily (i.e. when that line in the spec is being executed).

I just tested:  `%define` and `%global` work both.

There is just one caveat:

`%global _vpath_builddir %{_vpath_builddir}-my_suffix` does not work directly…

@ignatenkobrain and @Conan-Kudo Would you agree, we may change the direct use of `%{_vpath_builddir}` in the meson macros to `%{meson_builddir}`, and define `meson_builddir` to be `_vpath_builddir` by default?

That way one could redefine `meson_builddir` to `_vpath_builddir-my_suffix` pretty easily.

ignatenkobrain:
It is very same with optflags macro. You need to echo it in RPM shell...

PR #6475, opened 2020-01-16
RPM macros: Abstract %_vpath_builddir to %meson_builddir.

besser82:
With this small non-intrusive change one can customize the build-dir used by the %meson* macros by redefining it inside of the spec file with %define or %global, like:

```
%global meson_builddir %{_vpath_builddir}_my_suffix
%global meson_builddir my_prefix_%{_vpath_builddir}
```
***

This seems to be the cleaner solution over #6474.


ignatenkobrain:
I still don't see any reason for that. The whole idea of vpath macro was standardization.

PR #4845, opened 2019-01-29
qt: Only look for a framework on macOS if building for macOS

nirbheek:
When building for iOS, the Qt binaries only contain static libraries and headers. No framework.

With this, Meson can successfully compile and link to Qt on iOS.

Also, print the full path to qmake found.

PR #6481, opened 2020-01-17
Fix dependency('qt') crash with old Qt on OSX

jon-turney:
Don't fail with 'KeyError: QMAKE_XSPEC' if `qmake -query` doesn't report QMAKE_XSPEC. (Seen with Qt 4.8.7)

### DIFF
--- a/.github/workflows/ci_frameworks.yml
+++ b/.github/workflows/ci_frameworks.yml
@@ -71,7 +71,7 @@ jobs:
     - run: brew install pkg-config ninja gcc
     - run: brew tap cartr/qt4
     - run: brew install qt@4
-    - run: meson setup "test cases/frameworks/4 qt" build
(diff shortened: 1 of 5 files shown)


jon-turney:
Look like this has been broken since #4845.

You can install qt4 for osx using:
```
brew tap cartr/qt4
brew install qt@4
```
but this is hard to test in CI since homebrew doesn't allow you to have both qt4 and qt5 in PATH.

jon-turney:
> but this is hard to test in CI since homebrew doesn't allow you to have both qt4 and qt5 in PATH.

It looks like we have alredy have some similar testing in `.github/workflows/ci_frameworks.yml`, I'll see if I can extend that to cover this.

dcbaker:
This look good to me, and I'm ready to merge if you are @jon-turney

jon-turney:
> This look good to me, and I'm ready to merge if you are @jon-turney

Yes, thanks.

PR #6483, opened 2020-01-18
msvc: add prefix to build type arguments

PlushBeaver:
When using Clang compiler and MSVC linker (at least), linker arguments in compiler command-line must be passed via `-Wl`, otherwise Clang will reject the entire command with an error like this:

```
clang: @lib/XXX.rsp
clang: error: no such file or directory: '/OPT:REF'
```

### DIFF
--- a/mesonbuild/linkers.py
+++ b/mesonbuild/linkers.py
@@ -781,6 +781,9 @@ class VisualStudioLikeLinkerMixin:
         super().__init__(*args, **kwargs)
         self.machine = machine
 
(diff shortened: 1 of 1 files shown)


nirbheek:
@dcbaker I'd like to get this into a 0.53.2 release this weekend, do you think you could take a look at the consequences for this? If not, I'll dig into it and refresh my understanding of how all this works and review it :smile:

PlushBeaver:
JFYI, this patch received some field testing from DPDK project, for instance: http://mails.dpdk.org/archives/dev/2020-January/156559.html.

PR #6484, opened 2020-01-18
add zrythm to Users.md

alex-tee:


jpakkane:
Thanks, just note in the future that doc-only changes should have `[skip ci] ` at the end of their commit messages.

PR #6488, opened 2020-01-20
Users.md: add qboot

elmarco:
It's worth to list a meson project building a firmware.


jpakkane:
Thanks, just remember in the future that doc-only changes should have `[skip ci]` in their commit messages so CI is not run.

PR #6490, opened 2020-01-20
Support multiple args to message() and warning()

xclaesse:


tp-m:
I've been wanting this as well :)

xclaesse:
@tp-m could you please give it a review then, so it can be merged? Thank :D

PR #6491, opened 2020-01-20
Set source and build dir envvars when running dist.

jpakkane:


xclaesse:
Should update documentation too, there is no mention of env vars.

jpakkane:
They are mentioned in the reference tables already, though.

xclaesse:
LGTM

PR #6139, opened 2019-11-06
wrap.py: fail gracefully if no internet connection

scivision:
fixes #6130 

MWE to reveal problem / fix (make sure your computer is offline first):

```meson
project('offline')
dep = dependency('aaa', required: false, fallback: ['gtest', 'gtest_dep'])
```

also added type annotation in wrap.py, fixing other import/syntax issues thereby discovered.
To type check just part of Meson, in general do like:

```sh
mypy mesonbuild/wrap --follow-imports=skip
```

PR #6495, opened 2020-01-22
Fix git colors on windows again and again

nirbheek:
commit e1c0ee5ddbd6f0a9455bc98f10f5a1bfc6b98c78:

```
wrap: Use uppercase for global constants
This makes things much clearer and follows PEP8.
```

commit 10f205130e6d68c59e27b30326853c562296477e:

```
wrap: Look for git only once at startup
No need to repeatedly call `shutil.which`.
```

commit 1d366dace68ef241f42d3f6e01e6ca4f46f8efe9:

```
wrap: Always use a wrapper for running git
We need this for the next commit.
```

commit 793fa60f79e82f5fab6743bd96fcd71253484b5f:

```
wrap: Redirect stdin to DEVNULL when calling git
Without this git messes up the console and ANSI colors stop working on
Windows inside cmd.exe. Also add the same to all git calls. 
```

commit 8411b0559dc11fb55c18bcb0252b289b08f40bbe:

```
wrap: Re-set the console mode after calling git
`git submodule update --recursive` calls git clone recursively, and on
Windows it will undo the console mode we set in mlog and cause ANSI
colors to stop working. We could set it again only when we call that,
but we will definitely miss other instances where this could happen
in the future and regress.
```

commit 77f733daab6ffab07d8c7b50790a43371abd6018:

```
msubprojects: Rename 'git' to 'git_output'
No functional changes. Split out from the next commit for ease of
reading.
```

commit fca98e484f05b2f6d4b67c08aa8590b91a4ee0d9:

```
Move git helper out into mesonlib for reuse
Reuse the git helper for `meson wrap` and `meson subprojects` so we
don't need to maintain the same git-colors-on-windows workarounds in
multiple places.
```

This broke in https://github.com/mesonbuild/meson/pull/6139. Missed this when I opened https://github.com/mesonbuild/meson/pull/6255

### DIFF
--- a/mesonbuild/wrap/wrap.py
+++ b/mesonbuild/wrap/wrap.py
@@ -43,9 +43,9 @@ except ImportError:
     has_ssl = False
     API_ROOT = 'http://wrapdb.mesonbuild.com/v1/'
 
-req_timeout = 600.0
-ssl_warning_printed = False
-whitelist_subdomain = 'wrapdb.mesonbuild.com'
+REQ_TIMEOUT = 600.0
+SSL_WARNING_PRINTED = False
+WHITELIST_SUBDOMAIN = 'wrapdb.mesonbuild.com'
 
 
 def quiet_git(cmd: T.List[str], workingdir: str) -> T.Tuple[bool, str]:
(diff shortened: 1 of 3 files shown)


mesonbuild/wrap/wrap.py
```diff
@@ -266,44 +276,38 @@ def get_git(self) -> None:
...
-                subprocess.check_call([git, 'submodule', 'update',
-                                       '--init', '--checkout', '--recursive', *depth_option],
-                                      cwd=self.dirname)
+                verbose_git(['submodule', 'update', '--init', '--checkout', '--recursive', *depth_option],
+                            self.dirname, check=True)
+                # git submodule update --recursive calls git clone recursively,
+                # and on Windows it will undo the console mode we set in mlog
+                # and cause ANSI colors to stop working. Set it again.
+                if platform.system().lower() == 'windows':
+                    mlog._windows_ansi()
```
xclaesse: Is this still needed now that you set stdin to DEVNULL? Why would it be needed here and not anywhere else?

nirbheek: I clarified the comment, should make more sense now.

xclaesse: Ok I see, now you moved it to the common function, makes sense.

xclaesse:
The CI failure is unrelated.

Issue #6427, opened 2020-01-07
meson 0.53.0 gives "TypeError: Forward references must evaluate to types. Got NotImplemented."


imphil:
With the newly-released 0.53.0 we get the following error on Ubuntu 16.04 with its default Python ~3.5.1~ 3.5.2 (for x86). The previous version of meson worked perfectly.

```
meson -Dtarget=sim-verilator -Dot_version=opentitan-snapshot-20191101-1-434-gb99db7f -Ddev_bin_dir=/home/vsts/work/1/a/build-bin/sw/device/sim-verilator -Dhost_bin_dir=/home/vsts/work/1/a/build-bin/sw/host --cross-file=/tmp/toolchain.dhA3HP.txt /home/vsts/work/1/a/build-out/sw/sim-verilator
Traceback (most recent call last):
  File "/usr/local/bin/meson", line 11, in <module>
    load_entry_point('meson==0.53.0', 'console_scripts', 'meson')()
  File "/usr/local/lib/python3.5/dist-packages/pkg_resources/__init__.py", line 489, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/local/lib/python3.5/dist-packages/pkg_resources/__init__.py", line 2852, in load_entry_point
    return ep.load()
  File "/usr/local/lib/python3.5/dist-packages/pkg_resources/__init__.py", line 2443, in load
    return self.resolve()
  File "/usr/local/lib/python3.5/dist-packages/pkg_resources/__init__.py", line 2449, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/usr/local/lib/python3.5/dist-packages/mesonbuild/mesonmain.py", line 25, in <module>
    from . import mconf, mdist, minit, minstall, mintro, msetup, mtest, rewriter, msubprojects, munstable_coredata
  File "/usr/local/lib/python3.5/dist-packages/mesonbuild/mconf.py", line 16, in <module>
    from . import coredata, environment, mesonlib, build, mintro, mlog
  File "/usr/local/lib/python3.5/dist-packages/mesonbuild/environment.py", line 32, in <module>
    from . import compilers
  File "/usr/local/lib/python3.5/dist-packages/mesonbuild/compilers/__init__.py", line 102, in <module>
    from .compilers import (
  File "/usr/local/lib/python3.5/dist-packages/mesonbuild/compilers/compilers.py", line 389, in <module>
[...]

PR #6498, opened 2020-01-23
typing: Fix compatibility with Python 3.5.2

nirbheek:
Explicitly use the type instead of the string `'NotImplemented'` which still works with Python 3.5.2

Fixes https://github.com/mesonbuild/meson/issues/6427.

### DIFF
--- a/mesonbuild/compilers/compilers.py
+++ b/mesonbuild/compilers/compilers.py
@@ -386,7 +386,7 @@ class RunResult:
         self.stdout = stdout
         self.stderr = stderr
 
-class CompilerArgs(T.MutableSequence[str]):
+class CompilerArgs(collections.abc.MutableSequence):
     '''
     List-like class that manages a list of compiler arguments. Should be used
     while constructing compiler arguments from various sources. Can be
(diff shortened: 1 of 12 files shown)


mesonbuild/compilers/compilers.py
```diff
@@ -386,7 +386,7 @@ def __init__(self, compiled, returncode=999, stdout='UNDEFINED', stderr='UNDEFIN
         self.stdout = stdout
         self.stderr = stderr
 
-class CompilerArgs(T.MutableSequence[str]):
+class CompilerArgs(collections.abc.MutableSequence):
```
dcbaker: This changes the meaning from mypy's point of veiw, you've actually taken away the type annotation, because now mypy doesn't see this as a container of strings, it sees it as a container of Any.

I guess in the short term this is fine, in the long run I'd rather see something like @mensinda's annotation remover. There is a way to publish multiple wheels of a python package for different versions of python as well, so we could in theory publish a 3.5 version with the annotations stripped and a 3.6+ version with the annotations.

nirbheek: Maybe we can use a variable that changes based on the python version?

dcbaker: That might be possible. On the other hand, python 3.5 is EOL, so I think I'd prefer doing some kind of post processing if we really need to keep supporting it indefinitely

PR #6499, opened 2020-01-23
docs: Rename "mapfile" to "verfile"

freddy77:
The script passed to the linker to usually specify symbol versions
is called version script. Map file is a linker output file that
tell mapping between symbols and final addresses.

Signed-off-by: Frediano Ziglio <fziglio@redhat.com>


elmarco:
Imho, this is too late. Most projects out there use .map file, gnulib manual and Urlich Drepper "How To Write Shared Libraries" too.

freddy77:
Yes, apparently somebody call these file `Export Map` files.

PR #6503, opened 2020-01-24
ci: Add Arch Linux OS test

mensinda:
Add Arch Linux as a test environment to GitHub Actions.

@scivision The scalapack test is failing for some reason even though it is installed in the docker image from [the AUR](https://aur.archlinux.org/packages/scalapack). I am not sure if this is an issue with my image or a meson bug.


mensinda:
Fixed the preliminary CMake lookup for some special cases.

jpakkane:
Would it be possible to use prebuilt and cached images? Running a full install cycle on every build seems a bit wasteful.

mensinda:
> Would it be possible to use prebuilt and cached images? Running a full install cycle on every build seems a bit wasteful.

I am not sure what you mean by that. I am using the [official Arch Linux image](https://hub.docker.com/_/archlinux/) as a base and am installing all the necessary packages on top of it (+ a system update since Arch is a rolling release distro).

The `useradd` stuff is required since the arch image is *very* minimal and you are not allowed to run `makepkg` as root. I didn't want to use another image as the base since this would likely increase the image size even more.

mensinda:
Just to make sure we are talking at the same thing: The `install.sh` script is only run when the image is **built** in the Dockerfile. The [resulting image](https://hub.docker.com/r/mensinda/arch) is then pulled for the CI.

PR #6504, opened 2020-01-24
Fix required dependency() not failing when wrap-mode=nofallback

xclaesse:
When a dependency is required, not found on the system, and its fallback
is disabled with --wrap-mode=nofallback, meson should abort instead of
returning not-found.


tp-m:
Might be good to mention the "required" bit in the commit/MR summary as well, I was confused there for a second when I first saw it :)

ndufresne:
Tested locally, thanks !

PR #6506, opened 2020-01-24
Qt5: Do not use system qmake if not specified in cross file

xclaesse:
There is a comment saying we do it because we used to do it. But it's
wrong and lead to using system library when cross compiling.

Factor out the code we use to find pkg-config, because it is the same
use-case.

### DIFF
--- a/mesonbuild/dependencies/base.py
+++ b/mesonbuild/dependencies/base.py
@@ -380,6 +380,25 @@ class ExternalDependency(Dependency, HasNativeKwarg):
                         raise DependencyException(m.format(self.name, not_found, self.version))
                     return
 
+    # Create an iterator of options
+    def search_tool(self, name, display_name, default_names):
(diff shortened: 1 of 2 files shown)


ndufresne:
Tested locally, thanks!

jpakkane:
Test failure is relevant:

```
    def test_qt5dependency_qmake_detection(self):
        # Confirm that the dependency was found with qmake

...

        mesonlog = self.get_meson_log()
        self.assertRegex('\n'.join(mesonlog),
>                        r'Run-time dependency qt5 \(modules: Core\) found: YES .* \((qmake|qmake-qt5)\)\n')
E       AssertionError: Regex didn't match: 
```

ndufresne:
Thanks @jpakkane and @xclaesse, very that we figure-out why I had to much trouble with my cross-builds on limited rootfs.

PR #5728, opened 2019-07-27
Avoid building tests by default (ninja only)

NNemec:
For one of my projects, building the tests of a third party library causes tricky problems and is not worth the trouble. The easiest way to avoid this is to follow up a comment in the mesonbuild sources:

"XXX: Sometime in the future these should be built only before running tests."

This change achieves this for the ninja backend in a quite straightforward way, since most of the mechanism was already in place.

I did try hard to get it working for the Visual Studio / msbuild backend but ultimately gave up. All attempts to realize a separate meta-project for building the tests separately had side effects that broke CI and required a ever growing cascade of fixes in other places of the code. So ultimately I added a line to exactly preserve the behavior for Visual Studio.

PR #6511, opened 2020-01-25
Don't build tests by default in ninja

### LINKED ISSUE #1704: Test target depends on all executables
Say I have a meson build that defines several executables including a unittest-suite executable. Then I define one test for the one unittest executable.

Example:

    executable('some_frontend', 'foo_etc.cc')
    executable('some_frontend2', 'foo_etc2.cc')
    # some more executable definitions
    ut = executable('ut', 'some_sources.cc')
    test('unittest', ut)

When I execute:

    $ ninja-build ut
    $ ninja-build test

Expected behavior: The `ninja-build test` just executes the unittest executable since the test only depends on that executable.

Actual behavior: The `ninja-build test` also builds all other executables that are defined in the meson build file.

This is caused because meson generates the ninja test target like this:

    build test: CUSTOM_COMMAND all PHONY

The expected behavior would require the generation of something like:

    build test: CUSTOM_COMMAND ut PHONY

### LINKED ISSUE #2518: test executable always built even when build_by_default is false
I have an executable used for testing that has build_by_default set to false because the sources for the test executable are three times the size of the rest of the code and most of the time I don't need it built.

If I create a test() that uses the executable it is always built. I would expect it to only get built if I run ninja test.

dcbaker:
This is a rebased version of #5728, with the changes I requested applied on top. The idea is that tests and benchmarks are removed from the all target, and are placed in their own `meson-build-tests` and `meson-build-benchmarks` targets, which the `test` and `benchmark` targets depend on. This allows us to avoid building test executables by default.

I'd like to follow this up with a change that allows targets that are not installed to not build by default.

Fixes #2518
Fixes #1704

### DIFF
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -1002,6 +1002,20 @@ class Backend:
                 result[dep.get_id()] = dep
         return result
 
+    def get_benchmark_targets(self):
+        result = OrderedDict()
+        # Get all targets used as test executables and arguments.
+        for t in self.build.get_benchmarks():
+            for arg in [t.exe] + t.cmd_args:
+                if hasattr(arg, 'held_object'):
+                    arg = arg.held_object
+                if isinstance(arg, (build.CustomTarget, build.BuildTarget)):
+                    result[arg.get_id()] = arg
+            for dep in t.depends:
+                assert isinstance(dep, (build.CustomTarget, build.BuildTarget))
+                result[dep.get_id()] = dep
+        return result
+
     @lru_cache(maxsize=None)
     def get_custom_target_provided_by_generated_source(self, generated_source):
         libs = []
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -992,12 +992,9 @@ class Backend:
         # built only before running tests.
         for t in self.build.get_tests():
             exe = unholder(t.exe)
-            if isinstance(exe, (build.CustomTarget, build.BuildTarget)):
-                result[exe.get_id()] = exe
-            for arg in unholder(t.cmd_args):
-                if not isinstance(arg, (build.CustomTarget, build.BuildTarget)):
-                    continue
-                result[arg.get_id()] = arg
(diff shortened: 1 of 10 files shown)


scivision:
So is this a really short summary of what this does?

* I have a "large" project with lots of tests that only devs, not end-users that may build this project are interested in
* whether using "large" directly via `meson build` or as a Meson subproject, I don't want to build tests by default

this appears to be done new for this PR by:
```
executable('test_shaky', build_by_default : false)
```

such that test_shaky is not built by

```sh
ninja -C build
# or 
meson install -C build
```

but would be built by

```sh
meson test -C build
# or
ninja -C build test
```
---

Currently my approach to this is to enclose user-annoying tests like:

```meson
if get_option('dev')
  executable('test_shaky')
endif
```

dcbaker:
Ideally this would have a follow up of "don't build targets that aren't installed by default", so that if you had helper libraries only used in tests those also wouldn't get built by default. This is for me, motivated by the fact that mesa has a lot of test-only helper libraries that all have to be carefully guarded, and not having to do that guarding would be nice.

dcbaker:
I've adde a bunch of stuff to make working with run_project_tests more pleasant, as I had to run a few tests over and over on windows to figure some of the problems out.

emersion:
What's the status of this PR? Apart from rebasing, what are the outstanding issues?

jpakkane:
Featurewise this is ok, but the failing tests need to be fixed at least.

dcbaker:
I need to find a chance to rebase it and fix the failing tests, just not right at the top of queue right now

gaudetja:
> I need to find a chance to rebase it and fix the failing tests, just not right at the top of queue right now

I came across this PR and created https://github.com/mesonbuild/meson/pull/11317 to try and help with rebasing and running tests. Everything seems to be consistent between `master` and this branch as far as tests passing on my development machine. Thank you for creating this fix years ago!

bonzini:
Closing in favor of #11317, though even that one might be unnecessary given that meson has `meson-test-prereq` and `meson-benchmark-prereq`.

Issue #5962, opened 2019-09-26
Cross compilation requires both host and build compilers


keith-packard:
add_languages succeeds only if it finds compilers for both build and host machines. A project which only does cross-compilation shouldn't require a compiler for the build machine.

QuLogic:
Duplicate of #2392, essentially.

Issue #5862, opened 2019-08-29
True cross compile WITHOUT native compiler should be supported


mrneo240:
Hello, 

It's incredibly inconvenient there is not a proper way to have Meson not check for a native compiler that is running Meson.

This would be very handy when using docker or any other tailored toolchain solution for cross compiling.

I heard the argument that "oh you might need to compile generators", that's fine but let me declare that I don't need that functionality and continue.

Forcing a cross compile to have a native compiler is anti user

QuLogic:
Duplicate of #2392.

mrneo240:
So, after 2 years this is still a known issue?

xclaesse:
Issues don't fix themself magically. Did you write a fix we can reviews and merge?

mrneo240:
I would be happy to write one, if we can have a discussion with the maintainers to allow a user explicitly stating they won't be using or needing a local compiler.

ViliusSutkus:
> user explicitly stating they won't be using or needing a local compiler.

Shouldn't this be a default? CMake, Autotools (and others?) do not need local compilers.

jpakkane:
The simple solution is that the build definition is processed as usual, and if any sort of attempt is made to do something native (look up deps, define a native target etc) the build is immediately aborted.

Issue #2392, opened 2017-09-29
Don't require a native compiler when doing only cross-compilation


rhd:
I have a windows machine with the latest `.msi` installed - version `0.42.1`.

I set my path like this:

```
PATH=C:\Xilinx\SDK\2017.1\bin;C:\Xilinx\SDK\2017.1\gnu\microblaze\nt\bin;C:\Xili
nx\SDK\2017.1\gnu\arm\nt\bin;C:\Xilinx\SDK\2017.1\gnu\microblaze\linux_toolchain
\nt64_be\bin;C:\Xilinx\SDK\2017.1\gnu\microblaze\linux_toolchain\nt64_le\bin;C:\
Xilinx\SDK\2017.1\gnu\aarch32\nt\gcc-arm-linux-gnueabi\bin;C:\Xilinx\SDK\2017.1\
gnu\aarch32\nt\gcc-arm-none-eabi\bin;C:\Xilinx\SDK\2017.1\gnu\aarch64\nt\aarch64
-linux\bin;C:\Xilinx\SDK\2017.1\gnu\aarch64\nt\aarch64-none\bin;C:\Xilinx\SDK\20
17.1\gnu\armr5\nt\gcc-arm-none-eabi\bin;C:\Xilinx\SDK\2017.1\tps\win64\cmake-3.3
.2\bin;C:\Xilinx\DocNav;
```

This is my cross file (I've tried also tried full paths but it makes no difference:
```
$ cat armr5_cross.txt 
[binaries]
c = ['ccache', 'armr5-none-eabi-gcc']
cpp = ['ccache', 'armr5-none-eabi-g++']
ar = 'armr5-none-eabi-ar'
strip = 'armr5-none-eabi-strip'

[properties]
c_args = ['-DARMR5', '-mcpu=cortex-r5', '-mfloat-abi=softfp', '-mfpu=vfpv3-d16', '-nostdlib']
cpp_args = ['-DARMR5', '-mcpu=cortex-r5', '-mfloat-abi=softfp', '-mfpu=vfpv3-d16', '-nostdlib']

[host_machine]
system = 'notlinux'
cpu_family = 'arm'
cpu = 'r5'
endian = 'little'
```

Meson prints this error:

```
"c:\Program Files\Meson\meson.exe" build-r5 --cross-file=armr5_cross.txt
The Meson build system
Version: 0.42.1
Source dir: c:\work\phg
Build dir: c:\work\phg\build-r5
Build type: cross build
[...]

Issue #4402, opened 2018-10-22
How to cross compile for ARM with MSVC?


mstorsjo:
Normally when compiling with MSVC, you select the cross compilation target ("host" in the cross file) by running a script that sets up the environment (`PATH`, `LIB`) according to the desired host environment; the `cl.exe` in path is then the one for the chosen architecture, and `LIB` points to the directories containing libraries for this architecture.

If trying to run meson in such an environment, the default native compiler is also detected as `cl` from the path, a dialog box saying "The image file c:\...\sanitycheckc.exe is valid, but is for a machine type other than the current machine" appears, and meson fails with the following error message:

```
meson.build:25:0: ERROR:  Could not invoke sanity test executable: [WinError 216] This version of %1 is not compatible with the version of Windows you're running. Check your computer's system information and then contact the software publisher.
```

How is one supposed to set up cross compilation with Meson with the MSVC compiler, when you can only have one compiler in `PATH` and one `LIB` set of directories in the environment at a time?

jon-turney:
I think you must use a [cross-file](http://mesonbuild.com/Cross-compilation.html) to specify the cross-toolchain.

mstorsjo:
Ok, how do I specify a different `LIB` env var for the cross compiler vs the native one then?

dcbaker:
It might be worth it to just create the awful wrappers ourselves and ship them with meson to avoid this coming up again and again?

jon-turney:
Having a fix for #2392 would help, if we don't actually need to use the host compiler...

mstorsjo:
> Yes, this is terrible on many different levels but that is sadly the current limitation we have.

Ok, thanks for spelling it out!

Btw, a minor addition to your blogpost about awful things, on the topic of spawning processes. In addition to the quoting mess, the exec() functions are highly misleading, as they look like their posix counterparts, but doesn't behave the same (as you can't replace the current process, exec() spawns a new child process, but doesn't wait for it to exit, and instead exits the current process with a zero return code).

> Having a fix for #2392 would help, if we don't actually need to use the host compiler...

Yup, for at least my current case, this would be enough.

mstorsjo:
> * create wrapper bat files for cl and linker that first invoke the cross toolchain's `vcvarsall.bat` and then the actual cross compiler and put those in the cross file

[...]

PR #6512, opened 2020-01-25
Don't always require a native compiler

jon-turney:
Language additions and various fixes so a `meson.build` can be written which doesn't require a native compiler.

`add_languages()` gains a `native:` keyword, indicating if a native or cross compiler is required. 
`project()` gains a`languages:` and `native_languages:` keywords.
Add to CI applicable tests using the MSVC ARM64 cross-compiler only.

Closes: #2392, Closes: #5862, Closes: #5962
Mitigates: #4402. See also #3969


jon-turney:
@Ericson2314 I'm particularly uncertain about the last few commits.  Your review of those would be appreciated.

jpakkane:
Rather than adding a new kwarg, could we instead try to find both, fail gracefully if only one of them is found and then if the user tries to define a target with a missing language exit with a hard error?

Ericson2314:
I aslso want to split this and first fix `add_language` and only after `project`. @jon-turney if you would be so kind to split the PR like that I'd be happy to help you simplify things. I'll admit I'm surprised and confused by what I think are just `project` changes, but I'm more confident I know how to make `add_languages` simple. After that maybe we'll both together realize how to make `project` simple too :).

jon-turney:
> Rather than adding a new kwarg, could we instead try to find both, fail gracefully if only one of them is found and then if the user tries to define a target with a missing language exit with a hard error?

For the avoidance of doubt, I don't thinkthis is possible for `add_languages()`, as that has a return value, which could be amibiguous with lazy detection.

I think even if "'we'" implement that lazy detection, there should also be a syntax for being explicit about which machine(s) languages are used with.

The surface for interacting with compiler detection is not limited to just defining targets, there is also (at least) accessing compiler objects (`meson.get_compiler()`) and the `build|host_machine` objects.

> I also want to split this and first fix `add_language` and only after `project`. @jon-turney if you would be so kind to split the PR like that I'd be happy to help you simplify things.

Good idea.  I split that off as #6532.

jon-turney:
Split off #6636 to discuss a machine detection issue.

jon-turney:
Refined this in a similar way to add_languages(native:) in #6532, so (existing) `project('whatever', 'c')` means language 'c' is required for the host machine, and detected (but not required) for the build machine.

This is still WIP.  Some test failures are expected and addressed by #6636.

mesonbuild/ast/introspection.py
```diff
@@ -124,6 +124,10 @@ def func_project(self, node: BaseNode, args: T.List[TYPE_nvar], kwargs: T.Dict[s
 
         self.coredata.set_options(options)
         self.func_add_languages(None, proj_langs, None)
+        for kw in ['languages', 'native_languages']:
+            if kw in kwargs:
+                v = [kwargs.get(kw)]  # type: T.List[TYPE_nvar]
+                self.func_add_languages(None, v, None)
```
Ericson2314: `func_add_languages` currently iterates through both `MachineChoice.BUILD`, and `MachineChoice.HOST`. We should give `func_add_languages` a `for_machine` argument instead.

jon-turney: We can't (simply) change the signature of `func_add_languages()`, because it's invoked by the parser function dispatcher.  Certainly an internal `add_languages()` method could be added.

I don't know what use (if any) the introspector makes of the language information.  At this remove in time I no longer know what problem I thought adding this was fixing (reverting it doesn't make any tests fail, now).

Ericson2314: > Certainly an internal `add_languages()` method could be added.

That sounds good.

> I don't know what use (if any) the introspector makes of the language information.

Even if it does none, I think using a helper function is still good. In that case, it avoids a gotcha going forward if we do try to make it use that info.

jon-turney: I have factored this out as a helper function, `_add_languages()` as you suggest.

mesonbuild/ast/introspection.py
```diff
@@ -123,11 +123,12 @@ def func_project(self, node: BaseNode, args: T.List[TYPE_nvar], kwargs: T.Dict[s
...
         self.coredata.set_options(options)
-        self.func_add_languages(None, proj_langs, None)
+        self._add_languages(proj_langs, MachineChoice.HOST)
+        self._add_languages(proj_langs, MachineChoice.BUILD)
         for kw in ['languages', 'native_languages']:
             if kw in kwargs:
-                v = [kwargs.get(kw)]  # type: T.List[TYPE_nvar]
-                self.func_add_languages(None, v, None)
+                self._add_languages([kwargs.get(kw)], MachineChoice.HOST)
+                self._add_languages([kwargs.get(kw)], MachineChoice.BUILD)
```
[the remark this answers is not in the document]
Ericson2314: Ah, I forgot about positional vs keword args. Yeah the logic should be like `interpreter.py`. Right now the new keyword ways is not done altogether.

mesonbuild/ast/introspection.py
```diff
@@ -137,17 +142,20 @@ def do_subproject(self, dirname: str) -> None:
             return
 
     def func_add_languages(self, node: BaseNode, args: T.List[TYPE_nvar], kwargs: T.Dict[str, TYPE_nvar]) -> None:
-        args = self.flatten_args(args)
         for for_machine in [MachineChoice.BUILD, MachineChoice.HOST]:
```
Ericson2314: And we can do something based off of `kwargs` for this list too? Granted that a pre-existing bug we could also punt for a future PR. I know you've done tons of these revisions already :).

jon-turney: I think I omitted that logic, because as far as I can see, the native kwarg is ignored in the introspector for all other meson functions which support it (so presumably isn't signficant to introspection)

mesonbuild/ast/introspection.py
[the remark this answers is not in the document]
jon-turney: Done

jon-turney:
Looking at this again, I don't immediately see why the introspector's `_add_lanaguages()` needs to do anything.  Does the introspector make use of the compiler detected anywhere?

mensinda:
The introspector needs to know all the supported languages to add the `c_args`, `cpp_args` options, so that the result of `meson configure meson.build` is the same as the output of `meson configure <fresh build dir>` in 99% of all cases.

There are currently no other uses for languages in the introspection code (as far as I can remember).

mesonbuild/ast/introspection.py
mensinda: Adding `native` support here would be nice, if it would change the introspection output (what meson options printed with `meson configure` without a build dir). The other functions don't use `native` since it wouldn't make any difference to the introspection output (as far as I know).

jon-turney: It seems a bit odd that running the introspector on an unconfigured meson.build will detect compilers, but I guess that makes sense.

Ericson2314:
I think all the split-out parts have been merged so this can be rebased?

jon-turney:
Updated for CI changes, relies on https://github.com/mesonbuild/cidata/pull/4

Ericson2314:
It seems like there is still a lot of testing-related changes that while great, belongs in a separate PR?

Ericson2314:
The changes to meson itself do look good though!

jon-turney:
> It seems like there is still a lot of testing-related changes that while great, belongs in a separate PR?

You'll have to help me understand what parts you have in mind here.

Ericson2314:
@jon-turney everything to do with windows arm cross compilation we could land as a follow-up PR, right?

Ericson2314:
I had just been asking about splitting the PR on principle, but now I have concrete reason to split it up: I'd like to back-port the non-Windows bits to https://github.com/NixOS/nixpkgs/issues/86229 which fix cross builds for Nixpkgs/NixOS after I inadvertently broken them cleaning some things up.

The Windows bits wouldn't hurt, of course, but the smaller the back-port the better.

jon-turney:
> @jon-turney everything to do with windows arm cross compilation we could land as a follow-up PR, right?

There is nothing here which is specific to "windows arm cross compilation".

This PR contains a feature "Don't always require a native compiler".  `Contributing.md` says new features require tests.  Building in an environment where only a Windows ARM cross-compiler is available is an instance of that, and is used to test that feature.

> I had just been asking about splitting the PR on principle, but now I have concrete reason to split it up: I'd like to back-port the non-Windows bits

Still have no idea what parts you think can be separately merged.  Please give concrete examples.

Ericson2314:
> Building in an environment where only a Windows ARM cross-compiler is available is an instance of that, and is used to test that feature.

How about test on linux using `CC_FOR_BUILD=garbage` to prevent it from finding a working native C compiler for now? Then later we fix windows cross testing.

> Still have no idea what parts you think can be separately merged. Please give concrete examples.

I'll comment.

azure-pipelines.yml
```diff
@@ -53,6 +53,11 @@ jobs:
           arch: x64
           compiler: msvc2019
           backend: vs2019
+        vc2019armninjacross:
+          arch: arm64
+          compiler: msvc2019
+          backend: ninja
+          extraargs: --cross arm64cl.txt
```
Ericson2314, line 60: Save for windows cross later

ci/run.ps1
```diff
@@ -74,4 +74,4 @@ echo "=== Start running tests ==="
 # Starting from VS2019 Powershell(?) will fail the test run
 # if it prints anything to stderr. Python's test runner
 # does that by default so we need to forward it.
-cmd /c 'python 2>&1' run_tests.py --backend $env:backend
+cmd /c "python 2>&1 run_tests.py --backend $env:backend $env:extraargs"
```
Ericson2314, line 77: Save changes in this file for windows cross later

cross/arm64cl.txt
```diff
@@ -0,0 +1,14 @@
...
+windres = 'rc'
+
+[properties]
+c_args = ['-DWINAPI_FAMILY=WINAPI_FAMILY_APP']
+
+[host_machine]
+system = 'windows'
+cpu_family = 'aarch64'
+cpu = 'armv8'
+endian = 'little'
```
Ericson2314, line 14: Save for windows cross later

mesonbuild/environment.py
```diff
@@ -1094,7 +1094,7 @@ def sanitize(p):
                 linker = self._guess_win_linker(['link'], cls, for_machine)
                 return cls(
                     compiler, version, for_machine, is_cross, info, exe_wrap,
-                    target, linker=linker)
+                    target, full_version=cl_signature, linker=linker)
```
Ericson2314, line 1097: Save for windows cross later

mesonbuild/interpreter.py
```diff
@@ -2954,8 +2961,19 @@ def func_project(self, node, args, kwargs):
...
+                raise InterpreterException('project() may not have both language positional arguments and languages:.')
+            self.add_languages(mesonlib.stringlistify(kwargs.get('languages', [])), True, MachineChoice.HOST)
+        else:
+            self.add_languages(proj_langs, True, MachineChoice.HOST)
+
+        if 'native_languages' in kwargs:
+            self.add_languages(mesonlib.stringlistify(kwargs.get('native_languages', [])), True, MachineChoice.BUILD)
+        else:
+            self.add_languages(proj_langs, False, MachineChoice.BUILD)
+
```
Ericson2314, line 2976: When there are positional languages and `native_languages`, there is no error, but there should be.

How about:
```suggestion

        if len(proj_langs) > 0 and ('native_languages' in kwargs or 'languages' in kwargs):
                raise InterpreterException('project() may not have both language positional arguments and languages:.')
        self.add_languages(proj_langs, True, MachineChoice.BUILD)
        self.add_languages(proj_langs, True, MachineChoice.HOST)
        self.add_languages(mesonlib.stringlistify(kwargs.get('native_languages', [])), True, MachineChoice.BUILD)
        self.add_languages(mesonlib.stringlistify(kwargs.get('languages', [])), True, MachineChoice.HOST)
```

Ericson2314:
Let me just temper the Windows stuff by saying it does look good to me, (and I am excited about it!), and if @jpakkane and others are happy with merging it all at once, it's fine.

jpakkane:
The documentation should probably say that most projects should use they keywordless versions as they work 99% of the time. The other versions should only be used for projects with special needs.

Ericson2314:
@jpakkane that would be a bit of friction for Nixpkgs, where we don't give access to native compilers in cross builds unless explicitly requested.

Maybe after a few cycles we could make the list just be for the host platform in cross builds?

jpakkane:
Let's look at this through an example. Suppose we have something simple (like zlib) that builds only one library. Its build file should be something like:

```meson
project('foo', 'c')
... = static_library('foo', ...)
... = declare_dependency(...)
``´

This definition **must** work. It **must** work for native building. It **must** work for cross building. It should be the default recommended way of doing things. You should only ever need to specify native/cross compiler requirements _only if_ you build and run code during the build.

The automatic approach of "try to create both compilers, ignore if you can't instantiate a native compiler when cross building but error out when the build file tries to build a native target" is the common solution to the common use case. It makes the common case simple at the cost of making the uncommon one slightly more expensive (specifically by delaying the printing of errors to point of use rather than point of trying to instantiate the compiler). This is a good tradeoff.

File: mesonbuild/interpreter.py (hunk 8 above)
Comment (Ericson2314) [reply mesonbuild/interpreter.py:2976]: Oh nevermind I misread the message

**Comment by Ericson2314** [comment]
> This definition must work. It must work for native building. It must work for cross building. It should be the default recommended way of doing things. You should only ever need to specify native/cross compiler requirements only if you build and run code during the build.

We agree.

> try to create both compilers, ignore if you can't instantiate a native compiler when cross building but error out when the build file tries to build a native target

This introduces the same sort of implicit state / non-determinism that environment variables do. We also want users to be able to reproduce their builds, and we want the meson file to say what is needed, not under-approximate it because some things happen to work when extra compilers beyond what is requested are available.

I'm advocating a very small change to your plan, and only after a breaking change warning period:

##### native build
```meson
project('foo', *args)
```
is the same as
```meson
project('foo', languages: args, native_languages: args)
```

##### cross build
```meson
project('foo', *args)
```
is the same as
```meson
project('foo', languages: args, native_languages: [])
```

This means if you don't care about cross, you don't need to care about `native_languages`, even if you build and run code during the build. Only if you do care about cross, *and* use `native: true`, do you need to use `native_languages`.

\[This also means `languages` is no longer necessary.\]

jon-turney:
> The documentation should probably say that most projects should use they keywordless versions as they work 99% of the time. The other versions should only be used for projects with special needs.

I've adjusted the documentation to try to address this.

jon-turney:
Split out the introspector fix for add_languages(native:) as #7428

nirbheek:
@jon-turney would be nice to get this for 0.56, is it possible to split this out into seperate PRs so it's easier to merge?

jon-turney:
> @jon-turney would be nice to get this for 0.56, is it possible to split this out into seperate PRs so it's easier to merge?

I'll see what I can do, time permitting.

jon-turney:
> How about test on linux using `CC_FOR_BUILD=garbage` to prevent it from finding a working native C compiler for now? Then later we fix windows cross testing.

`CC_FOR_BUILD=garbage` fails with a hard error trying to invoke `garbage`, which is probably what's wanted for `CC_FOR_BUILD=typoed-compiler-command`.

jon-turney:
I've dropped the Windows ARM64 cross-only environment testing and replaced that with an arm cross-only environment test.
I've also dropped the additional kwargs for `project()`, since that simplifies things, and was obviously a sticking point.

Ericson2314 (approved):
Thanks for returning to this! And, of course, after, I'm still excited about that Windows cross stuff.

Ericson2314 (approved):

nirbheek (approved):
Went through it all, LGTM :)

jon-turney:
Split out Windows ARM64 cross-only environment testing as #7718.

jon-turney:
With hindsight, moving tests which require a native compiler to a separate suite was probably a mistake.  Now we have `test.json`, we could probably mark tests which require native tools in some way there.  That would also open up the way to cross-testing running suites other than `common`.

nirbheek:
Yeah, that sounds like a good idea to me.

PR #6513, opened 2020-01-25
tests: Prevent multiple-definition of symbols

ernestask:
With GCC 10, -fno-common becomes default behavior, meaning that any
subtly-broken code will be broken not so subtly anymore.

This commit changes the linkage to variables declared in headers to
external and, where needed, adds additional definitions in other
compilation units.

### DIFF
--- a/test cases/common/219 source set configuration_data/all.h
+++ b/test cases/common/219 source set configuration_data/all.h
@@ -3,5 +3,7 @@ extern void g(void);
 extern void h(void);
 extern void undefined(void);
 
-/* No extern here to get a common symbol */
-void (*p)(void);
+/* Defined in nope.c and f.c,
+ * value depends on the source set and configuration used.
+ */
+extern void (*p)(void);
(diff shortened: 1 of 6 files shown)


ernestask:
The context is me trying to re-enable running tests in Fedora (Rawhide has GCC 10). There’s still plenty of failing ones, but I’ll address them later in one way or another.

test cases/common/219 source set configuration_data/all.h
```diff
@@ -4,4 +4,4 @@ extern void h(void);
 extern void undefined(void);
 
 /* No extern here to get a common symbol */
-void (*p)(void);
+extern void (*p)(void);
```
jon-turney: Comment on the line above needs updating somehow, since it's now in direct contradiction to the declaration.

test cases/common/220 source set dictionary/all.h
```diff
@@ -4,4 +4,4 @@ extern void h(void);
 extern void undefined(void);
 
 /* No extern here to get a common symbol */
-void (*p)(void);
+extern void (*p)(void);
```
jon-turney: ditto

test cases/common/219 source set configuration_data/all.h
ernestask: Oh, hurr, my bad.

PR #6515, opened 2020-01-25
mtest: allow test processes to react on termination

camilo-celis:
Since neither SIGKILL nor SIGSTOP can be caught nor ignored it could be useful to terminate a test process via another signal first. Allowing some additional cleanup of a test. 

As an example, I would like to report a backtrace of the current state of a test when it times out as it could help understand an underlying problem. Currently, because the process running the test is terminated via SIGKILL it is not possible to do this.

### DIFF
--- a/docs/markdown/snippets/termination_signal_for_tests.md
+++ b/docs/markdown/snippets/termination_signal_for_tests.md
@@ -1,0 +1,6 @@
+## Changed the signal used to terminate a test process (group)
+
+A test process (group) is now terminated via SIGTERM instead of SIGKILL
+allowing the signal to be handled. However, it is now the responsibility of
+the custom signal handler (if any) to ensure that any process spawned by the
+top-level test processes is correctly killed.
--- a/mesonbuild/mtest.py
+++ b/mesonbuild/mtest.py
@@ -578,14 +578,27 @@ class SingleTestRunner:
             if is_windows():
                 subprocess.run(['taskkill', '/F', '/T', '/PID', str(p.pid)])
             else:
+
+                def _send_signal_to_process_group(pgid : int, signum : int):
+                    """ sends a signal to a process group """
+                    try:
+                        os.killpg(pgid, signum) # type: ignore
+                    except ProcessLookupError:
+                        # Sometimes (e.g. with Wine) this happens.
+                        # There's nothing we can do (maybe the process
+                        # already died) so carry on.
+                        pass
+
+                # Send a termination signal to the process group that setsid()
+                # created - giving it a chance to perform any cleanup.
+                _send_signal_to_process_group(p.pid, signal.SIGTERM)
+
+                # Make sure the termination signal actually kills the process
+                # group, otherwise retry with a SIGKILL.
                 try:
-                    # Kill the process group that setsid() created.
-                    os.killpg(p.pid, signal.SIGKILL)  # type: ignore
-                except ProcessLookupError:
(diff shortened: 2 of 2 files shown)


jpakkane:
This fails for cases where something is capturing SIGTERM for whatever reason. This should be done with something like:

- send SIGTERM
- wait at most 0.5 seconds
- if process not dead, send SIGKILL

camilo-celis:
> This fails for cases where something is capturing SIGTERM for whatever reason. This should be done with something like:
> 
> * send SIGTERM
> * wait at most 0.5 seconds
> * if process not dead, send SIGKILL

Good point, I sort of had that in mind but thought that the following subprocess `communicate` takes care of that case.

https://github.com/mesonbuild/meson/blob/dbad64cf34ba43b24dc2d71bd260f137acc37ea5/mesonbuild/mtest.py#L589-L595

jpakkane:
Reading the actual code, it is actually even more tricky.

The current code sends SIGKILL to the process and all its children.

The new code sends SIGTERM to the process and all its childred and then, if needed, SIGKILL to the process.

The difference here is that the child processes are not sent a SIGKILL any more so they might remain alive.

camilo-celis:
@jpakkane I gave your suggestion an initial stab. Now, we should be re-trying to kill the process group with SIGKILL if SIGTERM did not cut it.

jpakkane:
This gets even trickier. Suppose your main test executable dies with SIGTERM but some of its children do not. They would remain alive in this case because they would not get a SIGKILL, only a SIGTERM. So maybe something like:

- send SIGTERM to main process only
- wait 1 s for main process to exit
- send SIGKILL to process group
- if process had not died in the previous wait step, wait for it again

Though I don't know what happen to process groups when the "master" process is killed. I don't have the time to check the reference docs for this today, so if someone can point to an online reference, please do.

camilo-celis:
@jpakkane considering that specific case it does get trickier. But I am don't know if I follow the suggestion here. I thought that when a parent process is terminated all its children are re-parent. So, sending an additional SIGKILL to the process group would not have the intended effect. 

If my assumption is correct here, on the scenario you mention we would end up in the same case: having children processes be re-parent as children of the init process, and hence becoming zombies. 

I might be way off here, so please correct me if I am misunderstanding something.

jpakkane:
Based on some research, what we want to achieve seems to be impossible to do reliably:

- it is not possible to send a signal to members of a process group except the leader
- when the leader dies, its children get reparented and thus can't be reached any more

The approach in this MR would work, if we can assume that people who write signal handlers that ignore SIGTERM are expected to know what they are doing and be responsible for shutting down properly.

camilo-celis:
@jpakkane it looks like that is the case. I gave it also some thought but could not come up with a good solution. I stumble upon the _subreaper_ concept, but this might not be very portable. 

What about if from meson's perspective we cannot safely assume that people who write signal handlers are responsible for shutting down properly the test process we could choose which signal to be sent as part of a meson test argument? We would default to SIGKILL and allow SIGTERM hoping that by doing this explicitly a developer will be more responsible.

jpakkane:
That gets too complicated. Add a release note snippet explaining the behavioural change and then we can merge this. We can revert it if people raise issues (which seems unlikely).

camilo-celis:
@jpakkane alright. Added a snippet. Not 100% happy with the wording so feedback is appreciated.

PR #5859, opened 2019-08-28
Add --cross-file-constant and --native-file-constant arguments

xclaesse:
Config files already support interpolation, this allows setting
variables from the command line too. The common use-case is passing the
path the an SDK or toolchain.

PR #6517, opened 2020-01-26
Add variables in native/cross files

xclaesse:
This adds a new section `[variables]` to define variables that will be
replaced in other sections values using the @varname@ pattern.

Also:
- Removed type annotation for return value of from_config_parser()
  because it force values to be str, but could be lists too.
- It is an hard error to have an unknown section name, this should
  avoid hard to find mistake when there is a typo in a section name.
- Fix error messages saying 'cross file' when it's actually in the
  native file.
- Allow upper case in entries, by default configparser lowercase() them
  all, which makes impossible to have mixed case in binaries names, or
  upper case variables.

### DIFF
--- a/docs/markdown/Cross-compilation.md
+++ b/docs/markdown/Cross-compilation.md
@@ -328,6 +328,21 @@ myvar = meson.get_cross_property('somekey')
 # myvar now has the value 'somevalue'
 ```
 
+## Constants
+
+Since *0.54.0* constants can be defined in the `[constants]` section and
+used in other sections as `@varname@`. If provided, it must be the first section
+in the file. Values must all be strings, integers, booleans and lists are not
+allowed.
+
+```ini
+[constants]
+somepath = '/path/to/toolchain'
+[binaries]
+c = '@somepath@/gcc'
+cpp = '@somepath@/g++'
+```
+
 ## Cross file locations
 
 As of version 0.44.0 meson supports loading cross files from system locations
--- a/docs/markdown/Native-environments.md
+++ b/docs/markdown/Native-environments.md
@@ -76,6 +76,11 @@ expressing all of these configurations in monolithic configurations would
 result in 81 different native files. By layering them, it can be expressed by
 just 12 native files.
 
+## Constants
+
+Since *0.54.0* constants can be defined in the `[constants]` section and
+used in other sections as `@varname@`. See [here](Cross-compilation.md#constants)
+for details.
 
 ## Native file locations
 
--- a/docs/markdown/snippets/cross_file_constants.md
+++ b/docs/markdown/snippets/cross_file_constants.md
@@ -1,0 +1,13 @@
+## Cross and native file constants
+
+Constants can be defined in the `[constants]` section and used in other
+sections as `@varname@`. See [here](Cross-compilation.md#constants)
+for details.
+
+```ini
+[constants]
+somepath = '/path/to/toolchain'
+[binaries]
+c = '@somepath@/gcc'
+cpp = '@somepath@/g++'
+```
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -230,6 +230,8 @@ class UserFeatureOption(UserComboOption):
 def load_configs(filenames: T.List[str]) -> configparser.ConfigParser:
     """Load configuration files from a named subdirectory."""
     config = configparser.ConfigParser(interpolation=None)
+    # Do not lowercase all keys
+    config.optionxform = str
     config.read(filenames)
     return config
 
--- a/mesonbuild/envconfig.py
+++ b/mesonbuild/envconfig.py
@@ -79,30 +79,83 @@ CPU_FAMILES_64_BIT = [
 ]
 
 class MesonConfigFile:
-    @classmethod
-    def from_config_parser(cls, parser: configparser.ConfigParser) -> T.Dict[str, T.Dict[str, T.Dict[str, str]]]:
+    variable_format = 'meson'
+    regex = mesonlib.get_variable_regex(variable_format)
+
+    @staticmethod
+    def check_value(kind, entry, value, constants):
+        if isinstance(value, str):
+            value, missing = mesonlib.do_replacement(MesonConfigFile.regex, value, MesonConfigFile.variable_format, constants)
+            if missing:
+                var_list = ", ".join(map(repr, sorted(missing)))
+                raise EnvironmentException(
+                    "The constant(s) {} in the entry {} of {} file are not "
+                    "present in the [constants] section.".format(var_list, entry, kind))
+        elif not isinstance(value, (int, bool)):
+            raise EnvironmentException('Malformed value in {} file variable {}.'.format(kind, entry))
+        return value
+
+    @staticmethod
+    def check_entry(kind, entry):
+        if ' ' in entry or '\t' in entry or "'" in entry or '"' in entry:
+            raise EnvironmentException('Malformed variable name {} in {} file.'.format(entry, kind))
+
+    @staticmethod
+    def eval_value(kind, entry, value):
+        # Windows paths...
+        value = value.replace('\\', '\\\\')
+        try:
+            return eval(value, {'__builtins__': None}, {'true': True, 'false': False})
+        except Exception:
+            raise EnvironmentException('Malformed value in {} file variable {}.'.format(kind, entry))
+
+    @staticmethod
+    def eval_section(kind, s, constants):
+        section = {}
+        for entry, value in s.items():
+            MesonConfigFile.check_entry(kind, entry)
+            res = MesonConfigFile.eval_value(kind, entry, value)
+
+            if isinstance(res, list):
+                tmp = []
+                for i in res:
+                    tmp.append(MesonConfigFile.check_value(kind, entry, i, constants))
+                res = tmp
+            else:
+                res = MesonConfigFile.check_value(kind, entry, res, constants)
+
+            section[entry] = res
+        return section
+
+    @staticmethod
+    def eval_constants_section(kind, s):
+        section = {}
+        for entry, value in s.items():
+            MesonConfigFile.check_entry(kind, entry)
+            res = MesonConfigFile.eval_value(kind, entry, value)
+            if not isinstance(res, str):
+                m = 'Constant {} in {} file must be string'
+                raise EnvironmentException(m.format(entry, kind))
+            section[entry] = res
+        return section
+
+
+    @staticmethod
+    def from_config_parser(kind: str, parser: configparser.ConfigParser):
         out = {}
-        # This is a bit hackish at the moment.
-        for s in parser.sections():
-            section = {}
-            for entry in parser[s]:
-                value = parser[s][entry]
-                # Windows paths...
-                value = value.replace('\\', '\\\\')
-                if ' ' in entry or '\t' in entry or "'" in entry or '"' in entry:
-                    raise EnvironmentException('Malformed variable name %s in cross file..' % entry)
-                try:
-                    res = eval(value, {'__builtins__': None}, {'true': True, 'false': False})
-                except Exception:
-                    raise EnvironmentException('Malformed value in cross file variable %s.' % entry)
-
-                for i in (res if isinstance(res, list) else [res]):
-                    if not isinstance(i, (str, int, bool)):
-                        raise EnvironmentException('Malformed value in cross file variable %s.' % entry)
-
-                section[entry] = res
-
-            out[s] = section
+        constants = {}
+        for i, name in enumerate(parser.sections()):
+            s = parser[name]
+            if name == 'constants':
+                if i != 0:
+                    raise EnvironmentException("'constants' section must be first")
+                res = MesonConfigFile.eval_constants_section(kind, s)
+                # do_replacement() wants a tuple (value, desc) as values
+                constants = {k: (v, None) for k, v in res.items()}
+            elif name in ['binaries', 'properties', 'host_machine', 'paths']:
+                out[name] = MesonConfigFile.eval_section(kind, s, constants)
+            else:
+                raise EnvironmentException('Unknown {!r} section in {} file'.format(name, kind))
         return out
 
 class HasEnvVarFallback:
--- a/mesonbuild/environment.py
+++ b/mesonbuild/environment.py
@@ -542,7 +542,7 @@ class Environment:
         ## Read in native file(s) to override build machine configuration
 
         if self.coredata.config_files is not None:
-            config = MesonConfigFile.from_config_parser(
+            config = MesonConfigFile.from_config_parser('native',
                 coredata.load_configs(self.coredata.config_files))
             binaries.build = BinaryTable(config.get('binaries', {}))
             paths.build = Directories(**config.get('paths', {}))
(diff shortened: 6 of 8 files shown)


jpakkane:
I don't like this at all either in general or this UI in particular. Cross and native files are designed to be simple and unambiguous lists of facts. Adding all this functionality just to be able to save writing a few `/usr/bin`s to the cross file (which is written once and then rarely changes) is unnecessary complexity. 

Furthermore:

- `DEFAULT` is a poor name, it conveys no meaning to new users
- The interpolation syntax is completely different from the one we already have (`@VARNAME@`)
- The use of Python evals is an implementation detail that should not be leaked out. The original code had an explicit comment about this, but it seems have disappeared during refactoring.
- The name of the MR is misleading as this adds interpolation functionality rather than fixes existing functionality.

In fact any use of `eval` in Meson is something that should be eliminated. I only used them because I was too lazy to reimplement string/array parsing. Fortunately grepping says that there is only one instance in the code base remaining.

xclaesse:
It is a bit annoying that you get late in the fight with such radical opposition. You should have voiced your concerns months ago: https://github.com/mesonbuild/meson/pull/5859.

Currently cross files are a mess because sysroot/toolchain path is repeated everywhere (every binary, many c/cpp args). Those path are often not just /usr/bin, but a long path to where user decided to install the SDK. At the very least we need a way to have them written only in one place, and ideally it should come from the command line.

jpakkane:
This is not to say that we can't have some sort of an interpolation functionality. It's just that the approach listed here has noticeable downsides as listed above.

So something like this might be possible:

```
[constants]
prefix = '/usr/bin/bob'

[binaries]
c = '@PREFIX@/gcc'
```

Simply exposing Python internals is not nice, because they usually contain a ton of functionality that we don't really need but would then have to commit to supporting forever.

xclaesse:
@jpakkane ok, I implemented it as you suggested, more work but we don't depend on python interpolation any more.

tp-m:
(Random comment from the sidelines: I don't know if we have to be consistent with `configure()` files here in terms of the replacement pattern, but I do agree we shouldn't use the python mechanism. I think it would be nicer to use the scheme used in e.g. pkg-config files, which is `${prefix}`, just in case that hasn't been considered yet :))

xclaesse:
Unit test failure is not relevant, I've seen other PRs failing the same way.

@tp-m I personally have a slight preference for the @varname@ syntax, and tbh I was already using it myself to `sed` my crossfile.txt.in. And since I did all the work already, I would say it's too late unless someone has strong arguments against.

jpakkane:
Looks good, apart from the fact that the section should be called `constants` rather than `variables`, because they are, well, constants. `variables` implies that their values can change.

xclaesse:
Well, they won't be constant any more once we get to https://github.com/mesonbuild/meson/pull/5859

marc-h38:
> Well, they won't be constant any more once we get to #5859

... because their value will be defined once in the cross/native file, then _changed_ by the command line?

I'd rather document this as "constant defined first on the command line; immutable so canNOT be changed later in the cross/native file." And also print a warning in that case because:

https://github.com/mesonbuild/meson/pull/5859#issuecomment-582105298

> better to have things in one well defined place. Adding maximal flexibility and tens of ways to define everything becomes confusing very quickly. The one thing which I absolutely hated about CMake is that you can never be sure what the value of anything is, because there are a gazillion ways to inject things

nirbheek:
I like `@name@` because it is the natural extension of our string formatting rules which has `@1@`, etc. I'm sure some day we will add string formatting support for our dicts, and it will probably look like this.

I am not so sure about being able to override cross-file constants from the command-line. What I _would_ like is to be able to say "this constant needs to be provided by the user, and it's required". i.e., no 'default value' that is 'overriden'.

marc-h38:
> What I would like is to be able to say "this constant needs to be provided by the user, and it's required". i.e., no 'default value' that is 'overriden'.

"Even more constant", why not.

Now do you think solving the other, pure copy/paste problem _inside_ the cross/native files requires a completely independent and different solution? (I don't).

nirbheek:
> Now do you think solving the other, pure copy/paste problem _inside_ the cross/native files requires a completely independent and different solution?

No idea what you mean by "completely independent and different solution". I think we can add syntax for representing 'constants that must be provided by the user' in cross and native files.

marc-h38:
> No idea what you mean by "completely independent and different solution". 

1. This PR is solving a copy/paste problem inside files.

2. #5859 solves a fairly different problem: let each user point at his SDK location on the command line.

@xclaesse is solving both problems with a single solution. Do you think he should not? Do you think the solution to problem 1 and problem 2 should be independent from each other?

Focusing on the SDK example: if cross-files cannot be shipped with a default install directory that can be overriden on the command line, then maybe solution 1 and solution 2 should look different enough not to mislead me into thinking I can override that default location on the command-line.

xclaesse:
@nirbheek Hmm, so what you suggest is something like that:
```
[constants]
prefix= 'toolchain/usr/bin'
[binaries]
c = '@SDK_PATH@/@prefix@/clang'
```

Where `meson --cross-file-variable prefix=foo` would be an error because prefix is constant, omitting `--cross-file-variable SDK_PATH=/path` would be an error too because it must be provided by user. Is that what you mean? In that case I'm not sure `[constants]` section is even needed.

What I had in mind is a override mechanism:
```
[variables]
SDK_PATH = '/opt/android-ndk'
[binaries]
c = '@SDK_PATH@/bin/aarch64-linux-android-clang'
```
Where `meson --cross-file-variable SDK_PATH=/path` would override the default value, but omitting it would still work, assuming you installed the NDK in that path.

nirbheek:
> What I had in mind is a override mechanism:
> 
> ```
> [variables]
> SDK_PATH = '/opt/android-ndk'
> [binaries]
> c = '@SDK_PATH@/bin/aarch64-linux-android-clang'
> ```
> 
> Where `meson --cross-file-variable SDK_PATH=/path` would override the default value, but omitting it would still work, assuming you installed the NDK in that path.

The default value here makes no sense. You're either distributing a cross file for someone to plug in their own values into (think gst-build), or you're shipping something where you know the location of the NDK because it's a tightly-controlled environment (think cerbero).

What I was thinking of, and I suspect this is closer to what Jussi wants, is:

```
[constants]
sdk_path = get_value('sdk-path')

[binaries]
c = '@sdk_path@/bin/aarch64-linux-android-clang'
```

Then you'd call `meson setup --cross-file cross-file.txt --cross-file-value sdk-path=/opt/android-ndk` or something like that. The names of things don't matter here. The idea being that it's totally clear from looking at the cross file which values are already set, and which ones need to be provided by the user.

Mixing the two is just going to lead to hard-to-figure out cross files. With this cross file format, you don't need any documentation at all. It is self-evident which values need to be provided by the user. 

In your case, the situation looks simple right now because there's only one 'variable'. In sufficiently complex native/cross files, there will be many, and the user will have to guess (or you will have to document it somewhere), which is a complexity and usability trap. What if they forget? What if they don't set the right ones? They'll get a weird unrecognizable error.

xclaesse:
> ```
> [constants]
> sdk_path = get_value('sdk-path')
> 
> [binaries]
> c = '@sdk_path@/bin/aarch64-linux-android-clang'
> ```

get_value() here is clearly not obvious at all what it means, I don't like having pseudo-code in a config file.

But I get your point that you could easily have some `@foo@` hidden in the cross file that user will only find when meson will abort because a variable is not defined. I think that can be solved with just leaving empty values in the `[variables]` section and enforcing that all @foo@ substitutions must have `foo` defined in `[variables]` even if left with no value. My example would become:
```
[variables]
SDK_PATH =
[binaries]
c = '@SDK_PATH@/bin/aarch64-linux-android-clang'
```
And that would produce error if `--cross-file-variable SDK_PATH=/path` is ommitted.

The reason why I like being able to set a default value, is because some SDK does have a default install location, especially on Windows. Even on Linux I think in multiarch case I think you could have `/usr/lib/<multiarch>`. But that's secondary.

nirbheek:
> I think that can be solved with just leaving empty values in the `[variables]` section and enforcing that all @foo@ substitutions must have `foo` defined in `[variables]` even if left with no value.

I thought of that at first, but I wasn't sure if it was clear enough to users that a value was expected for that variable/constant. This is entering bikeshed territory, of course. As long as we agree about the general mechanism, I'm not going to be picky :)

> The reason why I like being able to set a default value, is because some SDK does have a default install location, especially on Windows

Windows 'default paths' are a lie, most of the time you need to read an environment variable to find the actual location (Intel Media SDK), or do a complicated lookup (Visual Studio SDKs), or the user needs to specify the path with no way for us to guess (Qt5 SDK, which ships separate binaries for different versions and different compilers and architectures). We've had to handle all these situations in Cerbero.

I actually cannot think of a single situation in which a hard-coded default will work for 90% of users of an SDK.

xclaesse:
Fair enough about SDK path that never have a default. Another case that I think could use a default is cpu_family could be a variable that default to aarch64 but can be set to arm. But not sure that will flight neither.

I wouldn't mind if variables that have non empty value in the cross file cannot be overridden by command line. But I feel it's going to be unclear limitation that users won't understand and will complain.

nirbheek:
> I wouldn't mind if variables that have non empty value in the cross file cannot be overridden by command line. But I feel it's going to be unclear limitation that users won't understand and will complain.

I'm sure we can come up with a syntax that clearly differentiates constants that are defined inside the cross file vs constants that are expected to be provided by the user. If there's no ambiguity, people won't be confused. They may ask us to add the ability to override constants, but we can tackle that based on the use-cases they provide. That's nothing new for us :wink:

marc-h38:
> I actually cannot think of a single situation in which a hard-coded default will work for 90% of users of an SDK

The installer of every single SDK (or similar SW) I ever used had a default install path that you could just use. While you clearly have much more experience, it's truly fascinating ours don't overlap.

nirbheek:
> The installer of every single SDK (or similar SW) I ever used had a default install path that you could just use. While you clearly have much more experience, it's truly fascinating ours don't overlap.

Yes, and every single one of them has multiple versions, and they're all parallel installable. Which one do you want to default to? Oh, you picked the wrong one, so the user now has to figure out how to change it. Except that they wanted you to pick the one that was in the env var for it, but we can't.

And if your SDK is simple enough to not need multiple versions and parallel installability, it usually comes as a Zip file.

Like I said, there is no such thing as a 'default install path' for SDKs, and it's best that no one pretends that such a thing exists. :smile:

marc-h38:
> Which SDK do you want to default to?

The one for the system you're shipping the cross/ files for. Obviously.

> Like I said, there is no such thing as a 'default install path' for SDKs, and it's best that no one pretends that such a thing exists. 

Your knowledge of everything is admirable.

xclaesse:
> I'm sure we can come up with a syntax that clearly differentiates constants that are defined inside the cross file vs constants that are expected to be provided by the user. If there's no ambiguity, people won't be confused.

Well... please suggest one ;-) I personally don't see why we should introduce such arbitrary limitation, I find it pretty elegant to have a `[variables]` section that could have a default value and can all be overriden from command line. Let's not make things more complicated than it needs to be.

germandiagogomez:
Just my two cents here. I really think, from a user point of view, it is better to pass explicit variables since this makes things clear. Since defaults are not a real thing for SDKs, I think that the get_variable('sdk_path') is what I would find more clear: 

1. it shows that you are getting something from outside.
2. you just need to set it once at configuration time
3. you have a conscious decision of what sdk you are setting. Nothing implicit.

Anyway, I am not been following this issue for a long time. So forgive a bit my intrusion into the conversation.

xclaesse:
Note that all discussion around how setting those variables from command line is out of scope for this PR, it should actually go to https://github.com/mesonbuild/meson/pull/5859. I separated the PRs because I know there are objections on that part.

Here we only allow variable substitutions in the cross/native file. The only thing blocking this PR is a pure naming question: constants VS variables. I personally think "variables" is the common way of calling this. That's how we call everything in Meson even when they are immutable, I see no point in inventing new names, just to waste my time for a PR ready to merge. Since they will be settable from outside the file, I think calling them constants is totally misleading.

marc-h38:
> That's how we call everything in Meson even when they are immutable,

That's a fair point, unfortunately undermined by all the rest:

> I see no point in inventing new names,

It's new only to meson; has been in every computer language since forever.

> just to waste my time for a PR ready to merge.

Not helping.

> Since they will be settable from outside the file, I think calling them constants is totally misleading.

In every computer language, a constant is "a value that never changes" while the program runs. This is one of the rare terms that's totally consistent, simply and well defined and universally understood across the entire field. Where and when that single value is defined is secondary.

xclaesse:
We is really bikesheding here, 'const int x;' is often called a *constant variable* (Interestingly, French wikipedia article does it a lot, but not English wikipedia, maybe it's my mother language bias).
- In the context of Meson, everything is constant, and we still call them variables (e.g. `get_variable()`).
- In the context of a cross file they won't be constant since their value can be changed from outside. I proposed `--cross-file-variable key=value` command line, are you recommending to rename `--cross-file-constant key=value`?
 
Anyway, if I'm the only one to think this is useless bikeshedding, I'll just rename it in the PR, I was not expecting wasting my life on this.

@jpakkane that's your last word?

marc-h38:
A bikeshed/waste of time would be about some obscure symbol name buried in the source code. This is going to be front and center in the user interface - which developers have an unfortunate tendency not to treat differently. UI is hard because it's difficult to put yourself in the shoes of someone who knows less / wants to know as little as possible to get his job done - especially with one of the "[universal despised](https://mesonbuild.com/Design-rationale.html)" build systems. It's much easier for me because I'm very ignorant compared to you :-)

Yes there will be some inconsistencies, but how annoying/minor will they really be? I don't find it shocking that `get_variable()` can return a constant.

> In the context of a cross file they won't be constant since their value can be changed from outside.

It won't change while the program runs, so that's still a constant with all its benefits from a programming perspective: https://www.google.com/search?q=immutability+safety

> recommending to rename --cross-file-constant key=value?

I'm tempted to say yes but I haven't thought about it much and I won't; this may surprise you but I have limited time too :-)

QuLogic:
> In the context of Meson, everything is constant, and we still call them variables (e.g. `get_variable()`).

That is not true. In Meson, the _values_ are immutable (except maybe dictionaries, I'm not up-to-date on that.) The names that we assign to those values are not constant, and thus are called variables.

xclaesse:
@QuLogic which is exactly what would happen with cross file variables, their value can be changed from command line.

Anyway, since I'm the only one to understand it like that, I'll update it to call it a constant. Feels wrong, but whatever.

marc-h38:
> which is exactly what would happen with cross file variables, their value can be changed from command line.

"Changed" except the "former" value was (please correct me) never visible and is totally impossible to access from any `meson.build` code no matter what meson loophole you try to find. If correct then I'm genuinely sorry you don't see the difference it makes.

QuLogic:
Just to clarify, I have made no pronouncement on what the things in native/cross files should be called, just on what existing things in Meson do. I have not had any time to review this PR or its effects.

TheQwertiest:
FWIW, I'm in a `variable` camp. Reason is simple: meson does not use `const` in it's API. Every method that deals with variables (be it const or non-const) uses `variable`/`variables` and never ever uses `const` or `constant` (even in documentation).
In other words, meson language does *not* have a proper `const` concept, so it should not be introduced unless there are plans to rework other parts of meson as well or a *VERY* good reason that requires introduction of such concept.

xclaesse:
In the light of https://github.com/mesonbuild/meson/pull/6815, I'm not sure any more this is the right approach. Adding support for `+` operator on strings/lists has numerous advantages.

I still believe we need a `[variables]` (call it constants if you want) section, but it could be working like that instead:
```
[variables]
sysroot = '/path'
common_args = ['--sysroot=' + sysroot, '-DFOO']

[properties]
c_args = common_args
cpp_args = common_args + ['-DBAR']

[binaries]
c = sysroot + '/bin/gcc'
cpp = sysroot + '/bin/g++'
```

Where sysroot could be set from command line: `meson build --cross-file-variable sysroot=/toolchain`.

I personally think it's stretching the file format a lot, but it would be really handy in practice.

PR #6519, opened 2020-01-26
Adding more Meson sample templates.

michaelbadcrumble:
I have produced new templates for Meson build systems existing project template library. I also added a feature note, remembered to update test_templates, moved Meson specific generator into its template file and added docstrings in `minit.py`.  The programming languages I selected `java`, `Objective-C++`,  `C#`, and `Cuda`.

At some point, I will produce a final batch of templates for what I consider to be the hardest programming languages to write template definitions for, `Vala`, and  `Swift`.


michaelbadcrumble:
Ping @scivision?

PR #6521, opened 2020-01-27
added cwd kwarg to run_command

### LINKED ISSUE #6584: Add working directory as an optional KWARG to run_command
Basically passes an optional cwd kwarg to the underlying subprocess open.

the PR is here \:
https://github.com/mesonbuild/meson/pull/6521

I think this adds a lot of flexibility when interfacing with legacy scripts and tools

peterino2:
Exactly what the commit name sounds like allows a command to run in a specified CWD

Let me know if this is a good idea or one that is just right out.. If so I can add unit tests etc... 

Fixes #6584

### DIFF
--- a/docs/markdown/External-commands.md
+++ b/docs/markdown/External-commands.md
@@ -35,11 +35,13 @@ Since 0.54.0
 You can also pass a path to the working directory for a given command
 
 ```meson
-results = run_command('command', 'arg1', 'arg2', cwd: 'path/to/wd')
+results = run_command('command', 'arg1', 'arg2', workdir: 'path/to/working_directory')
+# can be combined with meson directives
+results = run_command('command', 'arg1', 'arg2', workdir: meson.source_root() + '/path/in/source')
 ```
 
 
-The `run_command` function returns an object that can be queried for
+The `run_command` function returns an object that can be  for
 return value and text written to stdout and stderr. The `strip` method
 call is used to strip trailing and leading whitespace from
 strings. Usually output from command line programs ends in a newline,
--- a/docs/markdown/External-commands.md
+++ b/docs/markdown/External-commands.md
@@ -31,6 +31,14 @@ env.set('FOO', 'bar')
 run_command('command', 'arg1', 'arg2', env: env)
 ```
 
+Since 0.54.0
+You can also pass a path to the working directory for a given command
+
+```meson
+results = run_command('command', 'arg1', 'arg2', cwd: 'path/to/wd')
+```
+
+
 The `run_command` function returns an object that can be queried for
 return value and text written to stdout and stderr. The `strip` method
 call is used to strip trailing and leading whitespace from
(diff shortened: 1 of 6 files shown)


mesonbuild/interpreter.py
```diff
@@ -2448,7 +2449,7 @@ def validate_arguments(self, args, argcount, arg_types):
                 if not isinstance(actual, wanted):
                     raise InvalidArguments('Incorrect argument type.')
 
-    @FeatureNewKwargs('run_command', '0.50.0', ['env'])
+    @FeatureNewKwargs('run_command', '0.50.0', ['env', 'cwd'])
```
eli-schwartz: The feature is new to 0.54.0, not 0.50.0, so add a new decorator rather than appending to an old one.

searzocom:
Ok I got a unit test in and updated the docs. Let me know if that makes sense

jpakkane:
The test should take a File object as an input and verify that the path is correctly translated to the new cwd as in #6585.

Also, `cwd` is not a particularly great name for this, because it is an acronym for "current working directory". Maybe something like `workdir` instead?

jon-turney:
The documentation for `run_command` in the reference manual should probably be also be updated to mention this kwarg.

It also goes to some lengths there to mention that the command is run in an *unspecified* directory.

searzocom:
> The test should take a File object as an input and verify that the path is correctly translated to the new cwd as in #6585.
> 
> Also, `cwd` is not a particularly great name for this, because it is an acronym for "current working directory". Maybe something like `workdir` instead?

good idea

searzocom:
> The documentation for `run_command` in the reference manual should probably be also be updated to mention this kwarg.
> 
> It also goes to some lengths there to mention that the command is run in an _unspecified_ directory.

can you elaborate a bit more? Is the default working directory not defined by default?

jon-turney:
I just meant that "The command is run from an unspecified directory" at https://github.com/mesonbuild/meson/blob/master/docs/markdown/Reference-manual.md#run_command needs updating.

searzocom:
> I just meant that "The command is run from an unspecified directory" at https://github.com/mesonbuild/meson/blob/master/docs/markdown/Reference-manual.md#run_command needs updating.

that makes sense thanks!

docs/markdown/External-commands.md
```diff
@@ -35,11 +35,13 @@ Since 0.54.0
...
 ```meson
-results = run_command('command', 'arg1', 'arg2', cwd: 'path/to/wd')
+results = run_command('command', 'arg1', 'arg2', workdir: 'path/to/working_directory')
+# can be combined with meson directives
+results = run_command('command', 'arg1', 'arg2', workdir: meson.source_root() + '/path/in/source')
 ```
 
 
-The `run_command` function returns an object that can be queried for
+The `run_command` function returns an object that can be  for
```
jon-turney: Was 'queried' removed in error here?

eli-schwartz:
> The test should take a File object as an input and verify that the path is correctly translated to the new cwd as in #6585.

This never happened. :(

This will need rebasing on master, in addition to the test tweak. See https://mesonbuild.com/Yaml-RefMan.html for changes to how the reference manual is built.

PR #6523, opened 2020-01-27
Only provide 64 bit MSIs going forward.

jpakkane:


elmarco:
Win10 (or even win8) do not support x86 version? Any link?

jpakkane:
They do support 32 bit exes just fine. But since all Windows 10 versions have 64 bit support, there's no point in providing a 32 bit one. Everyone can use the 64 bit version.

elmarco:
I can see in MSDN there are win10 x86 & x64 versions. I don't know where you got the idea that they all use 64 bit version.

jpakkane:
Gah! Why must everything always be broken? Some points related to this:

- Is anyone actually using 32 bit Windows 10? I've never seen one in the wild.
- I don't have access to said 32 bit Windows 10 so I could not create 32 bit installers even if I wanted to.

elmarco:
fwiw VLC on windows is still 32 bit download by default (https://www.videolan.org/vlc/download-windows.html). Firefox still ships 32 bit versions too.

jpakkane:
That is different. Shipping 32 bit apps is totally fine. This is about 64 bit support in the platform. The bitness of your build system's exe has nothing to do with the binaries it outputs.

elmarco:
sure, if you want to only support meson on 64 bit host, that is quite fine to me.

jpakkane:
It's not even about that. It is about providing MSI installers only for 64 bit host. People can still use 32 bit Meson via Python if they so choose.

Building the installers is a pain point, because automating things on Windows is not the world's greatest experience. A fair share of the work needed to get a new release out goes to making the MSI installers.

michaelbadcrumble:
That makes sense.  MIS for the 64-bit host, and have 32-bit as an optional MIS package for anyone still using a 32-bit host.

Issue #6485, opened 2020-01-18
[RFC] Delegate module maintainership


xclaesse:
I would like to start by thanking the enormous work @jpakkane is doing with the Meson project. Being a maintainer is hard and ungrateful work, I'm really pleased by the time and effort he has always put in the project. If meson is so successful today it's thanks to him first.

I've personally been involved in the Meson project for about 2 years now, and while the project gets a continuous stream of contribution from everyone, I feel we are underpowered regarding maintainership and reviewing of patches. It is often a frustrating bottleneck.

I would like to suggest the idea of delegating maintainership of meson modules to individuals, who will be in charge of reviewing/approving patches touching their module. This means they would be responsible to approve or reject new methods and features in their module. That way @jpakkane can focus on the core of Meson.

The side effect of this proposal is to allow more flexibility to experimental modules that would not always match the same stability standards than the rest of Meson, as long as it has a clear maintainer and popular use-cases. Yes I'm thinking about a "cargo" module.

Why modules?
- They have self-contained code base, it's easy to know who is responsible for each PR.
- They have dedicated doc page, where we can make it clear when it's unstable/experimental/deprecated/etc.
- Easy to just drop/deprecate the code altogether if it turns out the experiment is a failure, or has been replaced by better alternatives.
- Easy to document who is in charge for each module, so contributers know who to talk to.
- In most projects this would be achieved by external plugins, but I personally don't like spreading the code base, and that would require to design stable public API which is not really the biggest strength of python.

Examples:
- I hereby offer becoming maintainer of pkgconfig module. :-P
- @nirbheek has offered to maintain a future cargo module.
- @mensinda is already maintaining cmake module.
- I'm sure people will step-in for other modules, comment below ;-)

Of course this won't solve everything, but could be a good start. What do you think?

marc-h38:
BTW: https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners 

>CODEOWNERS are automatically requested for review when someone opens a pull request that modifies code that they own."

> Yes I'm thinking about a "cargo" module.

`Cargo integration in Meson via an "unstable-cargo" module (try 2)` #2617

xclaesse:
> BTW: https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners
> 
> > CODEOWNERS are automatically requested for review when someone opens a pull request that modifies code that they own."

Yep, we should use that.

> > Yes I'm thinking about a "cargo" module.
[...]

PR #5031, opened 2019-03-07
Kconfig Module

bonzini:
Add kconfig integration such that existing projects can use the ".config" file from a kconfig system to conditionally compile in or out features.

The code is an updated version of the original submission in pull request #1617. Compared to the original, it uses the dictionary objects that have since been introduced in meson, and it lets you pass the input as a file object. This makes for a more flexible and easier to use API.

PR #1617, opened 2017-04-13
Kconfig Module

schultetwin:
Add kconfig integration such that existing projects can use the ".config" file from a kconfig system to conditionally compile in or out features.

PR #6524, opened 2020-01-27
Kconfig: unquote quoted strings

rossburton:
If a kconfig file sets a string value:

  CONFIG_FOO="bar"

Then the loaded dictionary returned by this module looks like this:

{
    'CONFIG_FOO': '"BAR"'
}

Whilst this is arguably correct it isn't helpful, as most code that
wants to use the strings will want to unquote them.  Help them out by
unquoting whilst parsing.

CC @marc-h38


jpakkane:
LGTM, other people can comment on the breakage potential.

marc-h38:
I originally asked for this in #5031 but now I'm having second thoughts and I'm afraid I seriously underestimated what we were getting into.

My concern is symbol typing in general and losing type information in particular.

Before this patch the user can easily "introspect" whether any Kconfig symbol is a string or not by... checking whether it's quoted. Gross but possible. This patch makes it difficult/impossible to check whether a parameter is a string versus an int or something else.

Before and after this patch all values are returned as Python strings. Maybe this could/should change?

Does such "introspection" matter in meson? Maybe not. I don't know.

The added `CONFIG_VAL_STRING1=foo` test line prompted me to test how kconfig silently "autocorrects" unquoted strings in `.config`.  Whereas Kconfiglib adds the missing quotes, the kernel's Kconfig silently drops the unquoted value and replaces it with an empty string. (So maybe Kconfiglib is not "bug for bug compatible" after all ;-). In other words, this new, unquoted `CONFIG_VAL_STRING1=foo` test is inconsistent with the behaviour of the kernel's Kconfig (v5.5rc7)

marc-h38:
https://www.kernel.org/doc/html/latest/kbuild/kconfig-language.html states:
> There are only two basic types: tristate and string; the other types are based on these two


What does "based on" mean exactly? I don't know. At least it means you can't enter any random input, there is some type checking happening (I just tested again that too to be sure).

I also had a look at what @ulfalizer does in https://github.com/ulfalizer/Kconfiglib. Unsurprisingly, it parses Kconfig files and remembers the type of each symbol, see `self._set_type(node, t0)` in `Kconfig._parse_properties(node)`. Then it performs type checks on default values in `_check_sym_sanity()`. Then `_load_config() `performs type checks when parsing a `.config` file. Last but not least, the `Symbol.type()` public interface makes the type available to any user of the library.

While meson.kconfig needs nowhere near that much type logic, it could probably use some of it. I think this particular example proves what was already discussed in #1617: the risk that this brand new module will just keep slowly duplicating Kconfiglib.

In #1617 there seems to have been an assumption that Kconfiglib would have to be embedded in meson. Hell no: I _already_ have Kconfiglib (otherwise how would I be using Kconfig?!), let me please control the kconfig version I want/need. I think meson.Kconfig should depend on but of course not "vendor" Kconfiglib.

Another option could be for (another?) meson.kconfig depending on and supercharged with kconfiglib to be hosted outside meson. Maybe inside kconfiglib even, @ulfalizer? How practical/realistic would it be for a meson module to be outside the meson repo? @xclaesse of #6485, any clue?

rossburton:
I did dig around to try and find a formal document of the Kconfig format to ensure the right types were returned but couldn't see anything. Also the tristate would presumably need a new object to be passed around.

dcbaker:
I think we're treading into interesting ground, and maybe we need @jpakkane's opinion. In my mind modules having external dependencies (as long as we're really, really, sure we need them) is fine. Many of them already have external dependencies on binaries, the windows module assumes windres, the qt module assumes various qt binaries, etc. In this case I think that it may be smarter to simply rely on kconfiglib for the kconfig module and not reinvent the wheel (and all of the bugs that come with it). It's not a core meson feature, so asking those who need it to add a python dependency wouldn't be any worse than asking them to add a binary dependency.

jpakkane:
Depending on kconfig binaries is ok, but not Python libs because:

- there are people who will install exes but not Python bindings (do not ask me why, there just are)
- external libs do not work with custom Python binaries

dcbaker:
Why do people like to hurt themselves so much?

jpakkane:
It seems that kconfiglib is only in PyPi, not in Debian. So that's out.

marc-h38:
> Depending on kconfig binaries is ok, but not Python libs because...

This left me confused. Thank you @dcbaker for filling some of my knowledge gaps offline. From https://mesonbuild.com/Getting-meson.html
> We provide an MSI installer on the GitHub release page that can be used to install both Meson and Ninja at once for Windows. It also contains an **embedded** copy of Python,..

Long story short there are Meson users who do **not** use Python (and don't want to use it).

Fascinating.

PS: I suspect they'd rather not use Meson either but don't really have a choice.

bonzini:
For what it's worth, it's okay for QEMU to adopt the kernel Kconfig's behavior and drop the quotes for strings, but we are currently relying on being able to parse quote-less strings.

One thing to be careful about is that "y" and y would return the same value after unquoting; and returning true and false for `y` and `n` would not be enough because we still could not distinguish `m` from `"m"`.

Perhaps instead we could add kconfig methods `unquote_str`, `to_bool` and `to_int`? To_bool would also accept `m` and keep it as a string, but incorrect types would gove an error. I also thought of a `parse` method that takes a dictionary and gives back another, but it would have the same issue of conflating `m` and `"m"`.

bonzini:
... Or even rename the module to something like `keyval`.

bonzini:
Thinking more about it, I think the best course of action is indeed to rename the module to keyval and decouple it from kconfig syntax. kconfig users can use `configuration_data(keyval.load(...))` and `get_unquoted` to access strings. Later on, arguments such as `bool_true`, `bool_false`, `quoted_strings` can be added without breaking backwords compatibility to make life simpler for kconfig users, but they can be considered separately.

marc-h38:
> adopt the kernel Kconfig's behavior and drop the quotes for strings, 

Could you expand that a bit? Just curious.

> I think the best course of action is indeed to rename the module to keyval and decouple it from kconfig syntax.

Sounds good, see you in #6599

bonzini:
> > adopt the kernel Kconfig's behavior and drop the quotes for strings,
>
> Could you expand that a bit? Just curious.

Our config file contains strings, but it's generated by a shell script (the bools are generated based on dependencies as in Kconfig) and is passed to both Make and Meson, so it's using unquoted strings. We don't have quotes in those strings however, so it would be okay to unconditionally unquote strings if they start and end with `"`. My preferred option however is to rename the module, though (of course :)).

bonzini:
Closing because the syntax supported by the `keyval` module is not linked to Kconfig anymore.

PR #6528, opened 2020-01-27
Fix changed pkg config path

dcbaker:


mesonbuild/coredata.py
```diff
@@ -717,16 +717,20 @@ def set_default_options(self, default_options, subproject, env):
...
-                    if k not in pkg_config_paths:
-                        pkg_config_paths.append(k)
-                options['pkg_config_path'] = pkg_config_paths
+        p_env = os.environ.get('PKG_CONFIG_PATH')
+        if p_env:
+            # PKG_CONFIG_PATH may contain duplicates, which must be
+            # removed, else a duplicates-in-array-option warning arises.
+            p_list = list(OrderedSet(p_env.split(':')))
+            if env.first_invocation:
+                options['pkg_config_path'] = p_list
```
xclaesse: It was already like that before, but this means the environment variable takes precedence over value set with `-Dpkg_config_path`. I don't think this is right, for CFLAGS it is the opposite.

xclaesse:
Hmmm, but since -Dpkg_config_path should take precedence, you can't know if `options['pkg_config_path']` is difference because the env changed or the user overridden the env by setting -Dpkg_config_path at first invocation.

Probably for the same reason we don't warn on changing CFLAGS env?

mesonbuild/coredata.py
dcbaker: No, this is setting the default values, if there is a command line value it will overwrite the environment variables. Just like CFLAGS. I just tested to be sure.

dcbaker:
I think it's worth telling a user that the PKG_CONFIG_PATH environment variable isn't the same as what meson is going to use, whether they set it original via PKG_CONFIG_PATH or via -Dpkg_config_path

mesonbuild/coredata.py
xclaesse: oh ok then.

xclaesse:
> I think it's worth telling a user that the PKG_CONFIG_PATH environment variable isn't the same as what meson is going to use, whether they set it original via PKG_CONFIG_PATH or via -Dpkg_config_path

Hmm, we don't warn for CFLAGS and friends, why is PKG_CONFIG_PATH different? Arguably we could warn for them all.

But if you do:
```
export PKG_CONFIG_PATH=FOO
meson --Dpkg_config_path=BAR builddir
```
then it will warn at every reconfigure because the option will always be different to the env. Right?

dcbaker:
Yes it would warn you in that case, I guess the question is whether other's feel it's important to warn about. @nirbheek, you asked for this originally, do you have a preference now?

xclaesse:
Wondering if we should warn if both are set tbh. Not only for PKG_CONFIG_PATH but also all others like CFLAGS.

jpakkane:
The first two commits seem unrelated...

xclaesse:
> The first two commits seem unrelated...

I think it's a rebase issue, those commits are already in master.

dcbaker:
I've rebased on master to get the unrelated commits to disappear.

mesonbuild/coredata.py
```diff
@@ -746,16 +746,20 @@ def set_default_options(self, default_options, subproject, env):
...
-                pkg_config_paths = []
-                for k in p_env.split(':'):
-                    if k not in pkg_config_paths:
-                        pkg_config_paths.append(k)
-                options['pkg_config_path'] = pkg_config_paths
+        p_env = os.environ.get('PKG_CONFIG_PATH')
+        if p_env:
+            # PKG_CONFIG_PATH may contain duplicates, which must be
+            # removed, else a duplicates-in-array-option warning arises.
+            p_list = list(OrderedSet(p_env.split(':')))
```
lazka: not sure if this is relevant here, but Windows uses ";" not ":"

xclaesse: We had that conversation, I don't remember where... There are multiple places that does that mistake and fixing it caused issues in the CI that we did not understand. Definitely something that needs to be fixed at some point.

PR #6530, opened 2020-01-28
Fix Renesas Compiler

alanNz:
Fixed Renesas RX Family compiler to work with latest meson, updated cross-file, fixed assembly file use

### DIFF
--- a/cross/ccrx.txt
+++ b/cross/ccrx.txt
@@ -10,11 +10,13 @@ strip = 'rlink'
 [properties]
 # The '--cpu' option with the appropriate target type should be mentioned
 # to cross compile c/c++ code with ccrx,.
-c_args      = ['--cpu=rx600']
-cpp_args    = ['--cpu=rx600']
+c_args      = ['-cpu=rx600']
+cpp_args    = ['-cpu=rx600']
+c_link_args = []
+cpp_link_args = []
 
 [host_machine]
-system = 'bare metal'      # Update with your system name - bare metal/OS.
+system = 'bare metal'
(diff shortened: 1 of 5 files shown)


jpakkane:
As per [the documentation](https://mesonbuild.com/Contributing.html#keeping-pull-requests-up-to-date) please don't merge trunk to your branch. It should have only one commit with your new code.

dcbaker:
@alanNz, if you do a `git rebase` on master, you should be able to easily get rid of the back merges, once those are taken care of this is ready to land.

alanNz:
@dcbaker I've just rebased here and tested it, cheers

cross/ccrx.txt
```diff
@@ -6,15 +6,53 @@ c = 'ccrx'
...
+  '-goptimize',
+  '-nostuff',
+  '-define=_REENTRANT,_FEVAL=0,_HAS_C9X_FAST_FMA=0,RX63_BOARD=CUSTOM631,__DBL8,DEVICE_PACKAGE_100_PIN,GHW_SINGLE_CHIP'
+  ]
+c_link_args = [
+  '-show=ALL',
+  '-list=device.map',
+  '-form=absolute',
+  '-nologo',
+  '-optimize',
```
jpakkane: If this does the same as `-O2` does on GCC, it should not be listed in the cross file. All optimization and debug flags should come from the compiler class based on build settings, not from cross files.

alanNz: good point changed now to less info

jpakkane:
Thanks.

PR #6368, opened 2019-12-18
cmake: Use new JSON trace format for CMake >= 3.17

mensinda:
Add support for the new JSON trace format, which is easier to parse (it is just JSON) and there is no longer any ambiguity with spaces. So we no longer have to do arcane magic tricks to determine which space is part of a path and which space separates arguments (at least once we have 3.17 as the minimum CMake version).

CMake merge request: https://gitlab.kitware.com/cmake/cmake/merge_requests/4102 (**currently open**)

The merge request is still open, but it should be in the final phases of the review and I don't expect any changes to the actual format. I am opening this PR already in the hope of getting this change into 0.53.0 just in case the CMake MR is merged before 22/12.

PR #6531, opened 2020-01-28
cmake: Add support for the new JSON trace format

mensinda:
This is basically #6368, without the refactoring portion. It should be safe to merge this PR once the first CMake RC is out.

Related CMake MRs:
- https://gitlab.kitware.com/cmake/cmake/merge_requests/4169
- https://gitlab.kitware.com/cmake/cmake/merge_requests/4242

### DIFF
--- a/docs/markdown/CMake-module.md
+++ b/docs/markdown/CMake-module.md
@@ -12,6 +12,11 @@ following functions will then be available as methods on the object
 with the name `cmake`. You can, of course, replace the name `cmake`
 with anything else.
 
+It is generally recommended to use the latest Meson version and
+CMake >=3.17 for best compatibility. CMake subprojects will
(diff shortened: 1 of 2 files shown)


dcbaker:
This all looks reasonable to me, Since it's presumably more reliable to use JSON than the human readable format, maybe add something to the documentation?

mensinda:
Yes, the JSON format is a lot more reliable (that's why I have added it to CMake :smile:).

I have updated the docs to recommend CMake >=3.17.

mensinda:
Merging, since [CMake 3.17.0-rc1 is released](https://cmake.org/cmake/help/latest/release/3.17.html#command-line), with the new trace format. Breaking changes and/or removal is now very unlikely.

PR #6532, opened 2020-01-28
Add add_languages(native:)

jon-turney:
Add add_languages(native:)

Split out from #6512.

### DIFF
--- a/mesonbuild/interpreter.py
+++ b/mesonbuild/interpreter.py
@@ -2927,9 +2927,9 @@ external dependencies (including libraries) must go to "dependencies".''')
             return self.add_languages(args, required, self.machine_from_native_kwarg(kwargs))
         else:
             # absent 'native' means 'both' for backwards compatibility
-            mlog.warning('add_languages is missing native:, assuming languages are required for both host and build.',
+            mlog.warning('add_languages is missing native:, assuming languages are wanted for both host and build.',
                          location=self.current_node)
-            success = self.add_languages(args, required, MachineChoice.BUILD)
+            success = self.add_languages(args, False, MachineChoice.BUILD)
             success &= self.add_languages(args, required, MachineChoice.HOST)
             return success
 
--- a/mesonbuild/interpreter.py
+++ b/mesonbuild/interpreter.py
@@ -2080,7 +2080,7 @@ _base_test_args = {'args', 'depends', 'env', 'should_fail', 'timeout', 'workdir'
 
 permitted_kwargs = {'add_global_arguments': {'language', 'native'},
                     'add_global_link_arguments': {'language', 'native'},
-                    'add_languages': {'required'},
+                    'add_languages': {'required', 'native'},
                     'add_project_link_arguments': {'language', 'native'},
                     'add_project_arguments': {'language', 'native'},
                     'add_test_setup': {'exe_wrapper', 'gdb', 'timeout_multiplier', 'env', 'is_default'},
@@ -2908,11 +2908,13 @@ external dependencies (including libraries) must go to "dependencies".''')
         self.build.projects[self.subproject] = proj_name
         mlog.log('Project name:', mlog.bold(proj_name))
         mlog.log('Project version:', mlog.bold(self.project_version))
-        self.add_languages(proj_langs, True)
+        self.add_languages(proj_langs, True, MachineChoice.BUILD)
+        self.add_languages(proj_langs, True, MachineChoice.HOST)
         self.set_backend()
         if not self.is_subproject():
             self.check_stdlibs()
 
+    @FeatureNewKwargs('add_languages', '0.54.0', ['native'])
     @permittedKwargs(permitted_kwargs['add_languages'])
     @stringArgs
     def func_add_languages(self, node, args, kwargs):
@@ -2921,7 +2923,15 @@ external dependencies (including libraries) must go to "dependencies".''')
             for lang in sorted(args, key=compilers.sort_clink):
                 mlog.log('Compiler for language', mlog.bold(lang), 'skipped: feature', mlog.bold(feature), 'disabled')
             return False
-        return self.add_languages(args, required)
+        if 'native' in kwargs:
+            return self.add_languages(args, required, self.machine_from_native_kwarg(kwargs))
+        else:
+            # absent 'native' means 'both' for backwards compatibility
+            mlog.warning('add_languages is missing native:, assuming languages are required for both host and build.',
+                         location=self.current_node)
+            success = self.add_languages(args, required, MachineChoice.BUILD)
+            success &= self.add_languages(args, required, MachineChoice.HOST)
+            return success
 
     def get_message_string_arg(self, arg):
         if isinstance(arg, list):
@@ -3009,9 +3019,8 @@ external dependencies (including libraries) must go to "dependencies".''')
         self.validate_arguments(args, 0, [])
         raise Exception()
 
-    def add_languages(self, args: T.Sequence[str], required: bool) -> bool:
-        success = self.add_languages_for(args, required, MachineChoice.BUILD)
-        success &= self.add_languages_for(args, required, MachineChoice.HOST)
+    def add_languages(self, args: T.Sequence[str], required: bool, for_machine: MachineChoice) -> bool:
+        success = self.add_languages_for(args, required, for_machine)
         if not self.coredata.is_cross_build():
             self.coredata.copy_build_options_from_regular_ones()
         return success
--- a/docs/markdown/Reference-manual.md
+++ b/docs/markdown/Reference-manual.md
@@ -54,9 +54,9 @@ Like `add_global_arguments` but the arguments are passed to the linker.
   bool add_languages(*langs*)
 ```
 
-Add support for new programming languages. Equivalent to having them
-in the `project` declaration. This function is usually used to add
-languages that are only used on some platforms like this:
+Add programming languages used by the project. Equivalent to having them in the
+`project` declaration. This function is usually used to add languages that are
+only used under some conditions, like this:
 
 ```meson
 project('foobar', 'c')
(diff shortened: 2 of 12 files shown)


Ericson2314:
Thanks!!

So I previously wrote `Interpreter.machine_from_native_kwarg` to handle the `native` kwarg everywhere it occurs. Can you try using it?

jon-turney:
> So I previously wrote `Interpreter.machine_from_native_kwarg` to handle the `native` kwarg everywhere it occurs. Can you try using it?

Nice! That didn't exist when I originally wrote this. Updated appropriately.

Do you want to explore the issues mentioned in https://github.com/jon-turney/meson/commit/97b9e974f8c69fdc34ee0560a9c9c651d0a1fbcd in this PR, I should I raise a separate one for that?

jpakkane:
From #6512:

> For the avoidance of doubt, I don't think this is possible for add_languages(), as that has a return value, which could be amibiguous with lazy detection.

For the common case it would not be:

- when building natively, there is one case
- when cross compiling, 99% of the time you care about the cross compiler, so the return value would be true if it was found and false otherwise

jon-turney:
> From #6512:
> 
> > For the avoidance of doubt, I don't think this is possible for add_languages(), as that has a return value, which could be amibiguous with lazy detection.
> 
> For the common case it would not be:
> 
>    * when building natively, there is one case
> 
>    * when cross compiling, 99% of the time you care about the cross compiler, so the return value would be true if it was found and false otherwise

I don't know what change this is asking me to make in this PR.

Please note that this PR doesn't change the existing meaning of `add_languages()` when `native:` is not present:  We try to find compilers for the specified languages for both build and host machines, and return false unless all are present.

mesonbuild/interpreter.py
```diff
@@ -2954,9 +2963,8 @@ def func_exception(self, node, args, kwargs):
         self.validate_arguments(args, 0, [])
         raise Exception()
 
-    def add_languages(self, args: T.Sequence[str], required: bool) -> bool:
-        success = self.add_languages_for(args, required, MachineChoice.BUILD)
-        success &= self.add_languages_for(args, required, MachineChoice.HOST)
+    def add_languages(self, args: T.Sequence[str], required: bool, for_machine: MachineChoice) -> bool:
+        success = self.add_languages_for(args, required, for_machine)
         if not self.coredata.is_cross_build():
             self.coredata.copy_build_options_from_regular_ones()
```
Ericson2314: We should move this logic to the caller.

mesonbuild/interpreter.py
```diff
@@ -2866,7 +2868,14 @@ def func_add_languages(self, node, args, kwargs):
...
             return False
-        return self.add_languages(args, required)
+        if 'native' in kwargs:
+            return self.add_languages(args, required, self.machine_from_native_kwarg(kwargs))
+        else:
+            # absent 'native' means 'both' for backwards compatibility
+            mlog.warning('add_languages is missing native:, assuming languages are required for both host and build.',
+                         location=self.current_node)
+            return (self.add_languages(args, required, MachineChoice.BUILD) and
+                    self.add_languages(args, required, MachineChoice.HOST))
```
Ericson2314: Uh oh! Now we sometimes return `bool` and sometimes `(bool, bool)`. If that logic from below is moved up, that would handle the native case, but in the cross case we could have independent failures. I think we just want to `&&` the responses.

Or, to be honest based on the 99% we could just default to "host" if no argument is given. That may not match old behavior but does match the other `native` keyword args.

mesonbuild/interpreter.py
jon-turney: Which logic?  Which call-site? Why?

mesonbuild/interpreter.py
jon-turney: >  we sometimes return `bool` and sometimes `(bool, bool)`

No, we sometimes return `bool` and sometimes return the logical-and of a pair of `bool`s, which is a `bool`.

> I think we just want to `&&` the responses.

The logical-and operator in python is `and`, which is used in the line you quote.

Ericson2314: Oh sorry I somehow missed that!

jpakkane:
> I don't know what change this is asking me to make in this PR.

What I meant was that we should do something like:

- if building natively, nothing changse
- if cross building then the success depends on whether the cross compiler is found
- thus if required is true and cross compiler is found, try to find native compiler, too, but if it is not found then carry on as usual (trying to build a native target with this language is a hard error)
- if cross building and required is false, if the cross compiler is not found, do not even attempt to find the native compiler and if it is found, behave as above

This would make the common case of cross compiling without needing to build native code generators work without any changes in build definition files.

Ericson2314:
> do not even attempt to find the native compiler and if it is found, behave as above

And if the host platform one is found, not the build platform one which isn't attempted?

I guess this is fine, though I hop we can eventually transition to the simpler semantics of no kwarg is the same as `native: false`.

jon-turney:
> I guess this is fine, though I hope we can eventually transition to the simpler semantics of no kwarg is the same as `native: false` that the others have.

That would certainly be simpler!

> What I meant was that we should do something like:

Not sure if this is rejecting the idea of having an explicit syntax, or just suggesting a more complex approach for the case when `native:` is omitted, but assuming the latter

>    * if building natively, nothing changse
> 
>    * if cross building then the success depends on whether the cross compiler is found

I think it's not actually necessary to change how the return value is determined in the cross-case (which adds another point of backwards incompatibility), just how required: is applied.

I also found it easier to write documentation for the behaviour that way.

>    * thus if required is true and cross compiler is found, try to find native compiler, too, but if it is not found then carry on as usual (trying to build a native target with this language is a hard error)

Added a failing test to cover that (although I don't think it's possible yet to execute things in an environment which exercise that).
 
>    * if cross building and required is false, if the cross compiler is not found, do not even attempt to find the native compiler and if it is found, behave as above

jon-turney:
> > * thus if required is true and cross compiler is found, try to find native compiler, too, but if it is not found then carry on as usual (trying to build a native target with this language is a hard error)
> 
> Added a failing test to cover that (although I don't think it's possible yet to execute things in an environment which exercises that).

Trying that test with #6512, in an enviroment without a native compiler, shows that we don't currently produce a helpful error message in that situation:

```
The Meson build system
Version: 0.53.999
Source dir: C:\cygwin64\wip\meson\test cases\failing\99 no native compiler
Build dir: C:\cygwin64\wip\meson\test cases\failing\99 no native compiler\_build
Build type: cross build
Project name: no native compiler
Project version: undefined
Build machine cpu: x86_64
Host machine cpu family: aarch64
Host machine cpu: armv8
Target machine cpu family: aarch64
Target machine cpu: armv8
Compiler for language c for the build machine not found.
meson.build:11: WARNING: add_languages is missing native:, assuming languages are required for both host and build.
Compiler for language c for the build machine not found.
  File "C:\cygwin64\wip\meson\mesonbuild\mesonmain.py", line 129, in run
  File "C:\cygwin64\wip\meson\mesonbuild\msetup.py", line 245, in run
    app.generate()
  File "C:\cygwin64\wip\meson\mesonbuild\msetup.py", line 159, in generate
    self._generate(env)
  File "C:\cygwin64\wip\meson\mesonbuild\msetup.py", line 192, in _generate
    intr.run()
  File "C:\cygwin64\wip\meson\mesonbuild\interpreter.py", line 4275, in run
    super().run()
  File "C:\cygwin64\wip\meson\mesonbuild\interpreterbase.py", line 412, in run
  File "C:\cygwin64\wip\meson\mesonbuild\interpreterbase.py", line 436, in evaluate_codeblock
    raise e
  File "C:\cygwin64\wip\meson\mesonbuild\interpreterbase.py", line 430, in evaluate_codeblock
    self.evaluate_statement(cur)
  File "C:\cygwin64\wip\meson\mesonbuild\interpreterbase.py", line 441, in evaluate_statement
  File "C:\cygwin64\wip\meson\mesonbuild\interpreterbase.py", line 788, in function_call
    return func(node, posargs, kwargs)
  File "C:\cygwin64\wip\meson\mesonbuild\interpreterbase.py", line 285, in wrapped
  File "C:\cygwin64\wip\meson\mesonbuild\interpreterbase.py", line 174, in wrapped
  File "C:\cygwin64\wip\meson\mesonbuild\interpreter.py", line 3461, in func_executable
    return self.build_target(node, args, kwargs, ExecutableHolder)
  File "C:\cygwin64\wip\meson\mesonbuild\interpreter.py", line 4469, in build_target
    target = targetclass(name, self.subdir, self.subproject, for_machine, sources, objs, self.environment, kwargs)
  File "C:\cygwin64\wip\meson\mesonbuild\build.py", line 1543, in __init__
    self.get_using_msvc()):
  File "C:\cygwin64\wip\meson\mesonbuild\build.py", line 1293, in get_using_msvc
    linker, _ = self.get_clink_dynamic_linker_and_stdlibs()
  File "C:\cygwin64\wip\meson\mesonbuild\build.py", line 1269, in get_clink_dynamic_linker_and_stdlibs
    raise AssertionError(m.format(self.name))
AssertionError: Could not get a dynamic linker for build target 'main'
```

Not quite sure how to improve that. As `BuildTarget.process_compilers()` says "It's ok if no compilers can compile it [an item in the sources list], because users are expected to be able to add arbitrary non-source files to the sources list." (so we plough on and fail looking for a linker...)

jon-turney:
> AssertionError: Could not get a dynamic linker for build target 'main'

This only occurs on Windows.  Otherwise, a far more informative error is produced.

```
ERROR: No specified compiler can handle file main.c
```

But this isn't actually new, the following currently produces the same assert with master.

```
project('p')
executable('main', 'main.c')
```

jpakkane:
> Not quite sure how to improve that. As BuildTarget.process_compilers() says "It's ok if no compilers can compile it [an item in the sources list], because users are expected to be able to add arbitrary non-source files to the sources list."

The only solution I can think of for now is to keep a list of known source file suffixes (union of all supported source files) and if no compiler is found for a given file, its suffix is checked against this list of suffixes. If the suffix is in the list, exit with an error.

jon-turney:
> The only solution I can think of for now is to keep a list of known source file suffixes (union of all supported source files) and if no compiler is found for a given file, its suffix is checked against this list of suffixes. If the suffix is in the list, exit with an error.

Yeah, that seems workable.

jpakkane:
LGTM. Anyone else have comments on this?

Issue #6391, opened 2019-12-29
Generated headers are not order-only deps for generated source files


kjellahl:
The description of executable() in the reference manual, https://mesonbuild.com/Reference-manual.html, says

>With the Ninja backend, Meson will create a build-time order-only dependency on all generated input files

I suppose this is also how library() is supposed to work. I can't get this to work as
expected with library() when both the source file (.cc file) and the header files are generated.

For .cc files in the source tree (not generated) it works as expected. The generated
headers become order-only dependencies in the build.ninja file (listed after '||').
For .cc files generated with custom_target(), all generated header files, listed as
sources in library(), become implicit dependencies (listed after '|').

We have projects with many generated .cc files and many generated header files.
The way Meson handles those files, all the generated .cc files are recompiled when
a generated header file has been changed, whether they include the changed header
file or not.

I use meson 0.52.1 and ninja 1.9.0 on Ubuntu 19.10.

kjellahl:
A patch such as this one fixes the issue for me.
```
diff --git a/mesonbuild/backend/ninjabackend.py b/mesonbuild/backend/ninjabackend.py
index 824b9584..d81f870a 100644
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -578,8 +578,7 @@ int dummy;
             if self.environment.is_llvm_ir(src):
                 o = self.generate_llvm_ir_compile(target, src)
             else:
-                o = self.generate_single_compile(target, src, True,
-                                                 header_deps=header_deps)
+                o = self.generate_single_compile(target, src, True, [], header_deps)
             obj_list.append(o)
 
         use_pch = self.environment.coredata.base_options.get('b_pch', False)
```
Can it have bad side effects in other cases? Was the generated headers deliberately
added as implicit dependencies to the generated sources? It has been like that since
commit ccf63459c6c9090518ce9164c67bc12de010bc21 2014-08-24.

kjellahl:
If the present behavior is deliberate, is there a workaround?
It causes many unnecessary recompilations. It's obviously contrary to what's
said in Meson's documentation, that it's meant to be extremely fast.

marc-h38:
> all the generated .cc files are recompiled when a generated header file has been changed, whether they include the changed header file or not.

This seems to be the real issue: why are these generated .cc files depending _at all_ on headers files not included and irrelevant?

[...]

PR #6535, opened 2020-01-28
docs: explain the purpose of order-only deps for generated files [skip ci]

marc-h38:
The documentation of "order-only" dependencies is limited and their
various purposes are especially not clear. See issue #6391 for a recent
example, search the internet for many more. So mention the particular
purpose here while making the documentation barely longer.


docs/markdown/Reference-manual.md
```diff
@@ -547,11 +547,12 @@ libraries (`.so`, `.dll`, etc) will be linked.
...
-all generated input files, including unknown files. For all input
-files (generated and non-generated), Meson uses the [dependency
-file](https://ninja-build.org/manual.html#ref_headers) generated by
-your compiler to determine when to rebuild sources. The behavior is
-similar for other backends.
+all generated input files, including unknown files. This is needed
+to bootstrap the generation of the real dependencies in the
+[depfile](https://ninja-build.org/manual.html#ref_headers)
+generated by your compiler to determine when to rebuild sources.
+Meson relies on this dependency file for all input files, generated
```
dcbaker: I know this that the original says this, but I 'm pretty sure that `meson` is supposed to be `ninja`

marc-h38: I meant "meson _indirectly_ relies on..." based on the last sentence which is: "The behavior is similar for other backends."

I can either add the missing "indirectly" or change to ninja, whatever works.

dcbaker: I thinks `meson` ... `other backends`. is confusing, as meson is not a backend, ninja is. I'd be just as happy to drop the `other backends part`

marc-h38:
Also submitted similar https://github.com/ninja-build/ninja/pull/1725/files

docs/markdown/Reference-manual.md
marc-h38: I (mis?)understand the "similar behavior for other **backends**" at the end of this paragraph to be referring to: "For the Ninja **backend**, Meson will create a build-time order-only dependency..." at the top of this paragraph.

On the other hand, you seem to (mis?)understanding this "similar behavior for other backends" to be referring to the "Meson relies on..." sentence immediately before that I'm modifying, correct?

My modification admittedly creates a little more distance between the top and end of this paragraph.

The git log traces this paragraph back to... a mass import from the wiki in 2017 :-(

jpakkane: This was most likely originally written by me and it was wrong, it should have said Ninja.

marc-h38:
> it should have said Ninja.

Done. Re-opening to retry CI.

PR #6536, opened 2020-01-28
Refactor CI cross-testing

jon-turney:
Refactor and simplify cross-testing in CI:

* Make the crossfile selection part of the CI job configuration (so we can more easily add more CI jobs for different cross-toolchains)
* Make the `platform_fix_name()`  machinery for toolchains which don't have gcc-like filename conventions also apply to cross-toolchain testing

### DIFF
--- a/run_cross_test.py
+++ b/run_cross_test.py
@@ -15,43 +15,25 @@
 # limitations under the License.
 
 '''Runs the basic test suite through a cross compiler.
-Not part of the main test suite because of two reasons:
 
-1) setup of the cross build is platform specific
-2) it can be slow (e.g. when invoking test apps via wine)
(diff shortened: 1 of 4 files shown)


run_cross_test.py
```diff
@@ -15,43 +15,25 @@
 # limitations under the License.
 
 '''Runs the basic test suite through a cross compiler.
-Not part of the main test suite because of two reasons:
 
-1) setup of the cross build is platform specific
-2) it can be slow (e.g. when invoking test apps via wine)
+This is now just a wrapper around run_project_tests.py with specific arguments
+'''
```
Ericson2314: Nice! But at least soon after this PR I'd like to see this wrapper go and the CI scripts just do the right thing instead.

PR #6537, opened 2020-01-29
ghwt: allow user to specify branch

UnoccupiedColonist:
New feature:
- Users may specify a branch to download a specific version of a project.  If a branch is not specified the original behavior remains.  If the specified branch does not exist, the script prints valid branch names and exits with an error.

Bug fixes:
- Subprojects now go to the subproject directory specified in the wrap file (if it exists) instead of always going to the default directory
- Fixed hash report swap if download did not match expected


jpakkane:
Thanks.

PR #6538, opened 2020-01-29
cmake: dependency: Add formatted defines and libraries

blobfish:
Had a build missing '-D' and '-l'('el'). The code had '-I'('eye') for includes, so I figured this was the place to fix, but I am unfamiliar with the code. Thanks.

### DIFF
--- a/mesonbuild/dependencies/base.py
+++ b/mesonbuild/dependencies/base.py
@@ -1420,8 +1420,9 @@ class CMakeDependency(ExternalDependency):
 
             # Try to use old style variables if no module is specified
             if len(libs) > 0:
-                self.compile_args = list(map(lambda x: '-I{}'.format(x), incDirs)) + defs
-                self.link_args = libs
+                self.compile_args = list(map(lambda x: '-I{}'.format(x), incDirs))
+                self.compile_args += list(map(lambda x: '-D{}'.format(x), defs))
+                self.link_args = list(map(lambda x: '-l{}'.format(x), libs))
                 mlog.debug('using old-style CMake variables for dependency {}'.format(name))
                 mlog.debug('Include Dirs:         {}'.format(incDirs))
                 mlog.debug('Compiler Definitions: {}'.format(defs))


mensinda:
This is the right place, but there are a lot of problems with this. For instance, `PACKAGE_DEFINITIONS` usually already start with `-D`, so this current fix would break other projects. The same thing can happen with `PACKAGE_LIBRARIES` and in 90% of the cases the full path to the `.so` is provided (instead of only the name).

Do you have a sample project or the link to the dependency where the current code fails?

blobfish:
On Wed, 29 Jan 2020 08:37:28 -0800
Daniel Mensinger <notifications@github.com> wrote:

> This is the right place, but there are a lot of problems with this.
> For instance, `PACKAGE_DEFINITIONS` usually already start with `-D`,
> so this current fix would break other projects. The same thing can
> happen with `PACKAGE_LIBRARIES` and in 90% of the cases the full path
> to the `.so` is provided (instead of only the name).
>
> Do you have a sample project or the link to the dependency where the
> current code fails?
>

I kind of figured it wasn't going to be this easy. The
linking missing '-l' was for a self compiled version of OpenCASCade on
both debian buster and bullseye. The missing '-D' was for a self
compiled version of VTK on debian bullseye. So it is kind of hard to
provide an example and the problems are probably rooted in broken
install scripts. I will attach cmake config files for both
dependencies. Maybe there is a clue there? Thanks.

blobfish:
If something like this works:
cmake_minimum_required(VERSION 3.1)
project(mesonOccTest VERSION 0.1.0)
find_package(OpenCASCADE REQUIRED)
include_directories(${OpenCASCADE_INCLUDE_DIR})
add_executable(mesonOccTest main.cpp)
target_link_libraries(mesonOccTest ${OpenCASCADE_LIBRARIES})

shouldn't something like this also work?:
project('mesonOcctTest', 'cpp',
  version : '0.1.0',
  license : 'GPL-3+',
  meson_version : '>=0.50',
  default_options : ['warning_level=3', 'cpp_std=c++11']
)
occt = dependency('OpenCASCADE', version : '>=7.3')
executable('mesonOcctTest', 'main.cpp',
  dependencies : [occt]
)


I will put some more time on this if someone, who is familiar with the cmake meson interaction, wants to guide me. Otherwise, I will close the pull request. Thanks.

mensinda:
For handling defines, you could take a look at [this line](https://github.com/mesonbuild/meson/blob/master/mesonbuild/dependencies/base.py#L1410).

I am not sure if anything can be done with the missing `-l` situation. If the value is not a path to a `.so`, or starts with `-l` then it is likely a target name. For `OpenCASCADE` specifically, I would recommend to manually specify the CMake targets you want (see the `modules` kwarg for CMake dependencies). 

Additionally, it would be a good idea to add print a note in the logs when the old-style variables are used but targets also exist [similar to this error here](https://github.com/mesonbuild/meson/blob/master/mesonbuild/dependencies/base.py#L1385).

blobfish:
> I am not sure if anything can be done with the missing `-l` situation. If the value is not a path to a `.so`, or starts with `-l` then it is likely a target name. For `OpenCASCADE` specifically, I would recommend to manually specify the CMake targets you want (see the `modules` kwarg for CMake dependencies).

I tried this a while back, maybe different now, and I lost the include directories as the occt cmake doesn't use target properties for includes.
<br><br>

> For handling defines, you could take a look at [this line](https://github.com/mesonbuild/meson/blob/master/mesonbuild/dependencies/base.py#L1410).

I was using 'old cmake' so that code doesn't apply, or am I confused?
<br><br>

Anyway. I will continue to look for an acceptable hack that doesn't involve altering meson code. Thanks for the input and your work on meson. I will close this pull request.

PR #6540, opened 2020-01-29
dependencies: Convert qt to DependencyFactory

dcbaker:
This splits the qt dependency into three separate dependencies: qmake,
pkg-config, and extra-frameworks.

This has an additional advantage that the qt module now uses the normal
find method, instead of creating a new dependency each time.


jpakkane:
:crossed_swords: conflicts. But seems ok, apart from the LGTM issues.

dcbaker:
This landed from a different PR already.

PR #6547, opened 2020-01-30
dependencies/base: Remove config-tool fallback to path

dcbaker:
This is conceptually similar to #6506, but covers all config tools
instead of qmake. As far as I'm aware there is no useful case for
picking a config-tool from path instead of from the cross file, and it's
the source of bugs more often than not, It's been marked deprecated for
a while, so lets remove it.


dcbaker:
Closed in favor of #7099

PR #6549, opened 2020-01-30
dependencies: netcdf always look for netcdf-fortran for fortran

dcbaker:
Otherwise it tries to link with the C bindings and fails. This is why
the test is broken on archlinux.

@mensinda 
@scivision


scivision:
This works for me on Ubuntu 18.04 and `apt install libnetcdff-dev`

PR #6550, opened 2020-01-30
Remove vs2015 jobs as Azure will stop supporting it.

jpakkane:


ci/azure-steps.yml
```diff
@@ -38,28 +38,16 @@ steps:
      DownloadFile -Source 'https://github.com/mesonbuild/cidata/raw/master/win32/pkg-config.exe' -Destination $(System.WorkFolder)\pkg-config.exe
      DownloadFile -Source 'https://download.microsoft.com/download/D/B/B/DBB64BA1-7B51-43DB-8BF1-D1FB45EACF7A/msmpisdk.msi' -Destination msmpisdk.msi
      DownloadFile -Source 'https://download.microsoft.com/download/D/B/B/DBB64BA1-7B51-43DB-8BF1-D1FB45EACF7A/MSMpiSetup.exe' -Destination MSMpiSetup.exe
-     if ($env:compiler -ne 'msvc2015') {
-       Start-Process msiexec.exe -ArgumentList '/i msmpisdk.msi /quiet' -Wait
```
jon-turney: The code inside this conditional should stay, since the condition is `-ne 'msvc2015'`?

jon-turney:
There's no MSVC x86 job after this removal, perhaps one of the remaining MSVC jobs should be changed to x86?

jpakkane:
If anyone knows about x86 VS setups, please help with this. I moved all of the tasks back to 64 bit, because I could not make them work with x86.

scivision:
References (FYI):

* VS2015 lifecycle / EOL [Oct. 2025](https://docs.microsoft.com/en-us/visualstudio/releases/2019/servicing#support-for-older-versions-of-visual-studio)
* Azure VS2015 dropped [March 23, 2020](https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops#use-a-microsoft-hosted-agent)

PR #6552, opened 2020-01-30
Fix prefix dependent option defaults

### LINKED ISSUE #6546: get_option(localstatedir,, sharedstatedir and sysconfdir) are relative
Those options should be `/var` and `/etc` according to existing/closed issues, merge requests and documentation, but i've had for ages relative values `var` and `etc`.

I need to pass the absolute paths to configuration_data(), but i can neither pass the value given by get_option(), nor concatenate with get_option('prefix').

What should be done ?

jon-turney:
Closes #6546

### DIFF
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -501,7 +501,7 @@ class CoreData:
         # Create builtin options with default values
         self.builtins = {}
         for key, opt in builtin_options.items():
-            self.builtins[key] = opt.init_option()
+            self.builtins[key] = opt.init_option(key, default_prefix())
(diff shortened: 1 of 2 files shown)


jpakkane:
This needs a test.

jon-turney:
> This needs a test.

The commit "Extend test_prefix_dependent_defaults unit test to cover default case" extends the `test_prefix_dependent_defaults` unit test in way which demonstrates the failure reported in issue #6546.

nirbheek:
Nice, my script caught that this PR wasn't added to the milestone while the issue was.

PR #6560, opened 2020-01-31
modules: Add a wayland module

dcbaker:
Currently this only adds a wrapper around wayland-scanner for handling
the setup and some of the trickiness of that tool. I noticed that a
number of projects (mesa, weston, gnome, a couple of test tools we have
internally) are all porting around different code for trying to find the
scanner and point at files in the protocol directory, and decided that a
module to handle this would be a welcome addition.

Obviously this currently doesn't have any tests or documentation, but I have ported a few internal tools and [mesa](https://gitlab.freedesktop.org/dbaker/mesa/tree/meson-use-wayland-module) to use this. At this point I'm hoping to get some feedback on the interface and whether this would be useful. Looking at both weston and mutter I see that they could probably use this to delete a lot of code, though I haven't tried to port either of those.

### DIFF
--- a/mesonbuild/modules/wayland.py
+++ b/mesonbuild/modules/wayland.py
@@ -1,0 +1,115 @@
+# Copyright © 2020 The Meson development team
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import typing as T
+from pathlib import PurePath
+
+from . import ExtensionModule
+from . import ModuleReturnValue
+from .. import build
+from .. import mlog
+from ..interpreterbase import permittedKwargs, stringArgs
+from ..mesonlib import MesonException, File, MachineChoice, version_compare
+
+if T.TYPE_CHECKING:
+    from ..interpreter import ModuleState, Interpreter
+
+
+class WaylandModule(ExtensionModule):
+
+    def __init__(self, interpreter: 'Interpreter'):
+        super().__init__(interpreter)
+
+        # wayland-scanner has a pkg-config file, we should use that to find
+        # wayland-scanner, as it's more robust
+        scanner_dep  = interpreter.dependency_impl(
+            'wayland-scanner', 'wayland-scanner', {'required': True, 'native': True})
+        scanner_name = scanner_dep.variable_method([], {'pkgconfig': 'wayland_scanner'})
+        self.scanner = interpreter.find_program_impl(
+            [scanner_name], silent=False, for_machine=MachineChoice.BUILD)
+
+        # Older version of wayland-scanner take different arguments
+        self.old_scanner = version_compare(scanner_dep.held_object.version, '< 1.15')
+
+        # We need this to find the location where upstream protocols are installed
+        self.protocols_dep = interpreter.dependency_impl(
+            'wayland-protocols', 'wayland-protocols', {'required': True})
+
+    @stringArgs
+    @permittedKwargs(build.CustomTarget.known_kwargs | {'is_protocol'})
+    def scanner_target(self, state: 'ModuleState', args: T.Sequence[str], kwargs: T.Dict[str, T.Any]) -> ModuleReturnValue:
+        """Wrapper the wayland-scanner tool, and create custom targets.
+
+        This provides a wrapper around wayland-scanner, simplifying things
+        like where to find the installed protocols and setting up paths
+        correctly. It can also auto generate an output name based on teh
+        input name, if an output name isn't provided, which is fine for
+        generating C files, though it may not be for generating headers.
+        """
+        if len(args) != 2:
+            raise MesonException('scanner_target requires 1 arguments')
+
+        name = args[0]
+        if not isinstance(name, str):
+            raise MesonException('First argument must be a string')
+
+        valid_types = {'client-header', 'server-header', 'private-code', 'public-code'}
+        type_ = args[1]
+        if not isinstance(type_, str):
+            raise MesonException('Second argument must be a string')
+        if type_ not in valid_types:
+            raise MesonException('Second argument must be one of {}'.format(' '.join(sorted(valid_types))))
+
+        # map new argument to old arguments if necessary (so we only accept the
+        # newer versions), error if the wayland-scanner is too old for the
+        # requested argument
+        if self.old_scanner:
+            if type_ == 'private-code':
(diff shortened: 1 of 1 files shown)


dcbaker:
Okay, I ported weston too, https://gitlab.freedesktop.org/dbaker/weston/tree/meson-wl-module

mesonbuild/modules/wayland.py
```diff
@@ -0,0 +1,115 @@
...
+class WaylandModule(ExtensionModule):
+
+    def __init__(self, interpreter: 'Interpreter'):
+        super().__init__(interpreter)
+
+        # wayland-scanner has a pkg-config file, we should use that to find
+        # wayland-scanner, as it's more robust
+        scanner_dep  = interpreter.dependency_impl(
+            'wayland-scanner', 'wayland-scanner', {'required': True, 'native': True})
+        scanner_name = scanner_dep.variable_method([], {'pkgconfig': 'wayland_scanner'})
```
xclaesse: I think you should make the pkg-config dependency optional, and fallback to set scanner_name to "wayland-scanner"

mesonbuild/modules/wayland.py
```diff
@@ -0,0 +1,115 @@
...
+        """Wrapper the wayland-scanner tool, and create custom targets.
+
+        This provides a wrapper around wayland-scanner, simplifying things
+        like where to find the installed protocols and setting up paths
+        correctly. It can also auto generate an output name based on teh
+        input name, if an output name isn't provided, which is fine for
+        generating C files, though it may not be for generating headers.
+        """
+        if len(args) != 2:
+            raise MesonException('scanner_target requires 1 arguments')
```
xclaesse: 2 arguments

mesonbuild/modules/wayland.py
xclaesse: One reason for this is wayland could be built as subproject I guess? In that case the pkg-config file won't exist yet but wayland project should have called override_find_program().

mesonbuild/modules/wayland.py
```diff
@@ -0,0 +1,115 @@
...
+
+        # Older version of wayland-scanner take different arguments
+        self.old_scanner = version_compare(scanner_dep.held_object.version, '< 1.15')
+
+        # We need this to find the location where upstream protocols are installed
+        self.protocols_dep = interpreter.dependency_impl(
+            'wayland-protocols', 'wayland-protocols', {'required': True})
+
+    @stringArgs
+    @permittedKwargs(build.CustomTarget.known_kwargs | {'is_protocol'})
```
xclaesse: Do we really want to expose every current and future kwargs of custom_target()? Do they all make sense? Maybe you whould copy your own list here.

mesonbuild/modules/wayland.py
```diff
@@ -0,0 +1,115 @@
...
+        # map new argument to old arguments if necessary (so we only accept the
+        # newer versions), error if the wayland-scanner is too old for the
+        # requested argument
+        if self.old_scanner:
+            if type_ == 'private-code':
+                type_ = 'code'
+            elif type_ == 'public-code':
+                raise MesonException('public-code type requires wayland scanner >=1.15')
+
+        if isinstance(kwargs['input'], list):
```
xclaesse: This would  backtrace in case input kwarg is missing, instead of raising a proper error message.

mesonbuild/modules/wayland.py
```diff
@@ -0,0 +1,115 @@
...
+            if len(kwargs['input']) != 1:
+                raise MesonException('Only one input file allowed.')
+            input_ = kwargs['input'][0]
+        else:
+            input_ = kwargs['input']
```
xclaesse: Missing check that it's a string.

mesonbuild/modules/wayland.py
```diff
@@ -0,0 +1,115 @@
...
+        if kwargs.pop('is_protocol', False):
+            p = self.protocols_dep.variable_method([], {'pkgconfig': 'pkgdatadir'})
+            input_ = (PurePath(p) / input_).as_posix()
+
+        if 'output' in kwargs:
+            output = kwargs['output']
```
xclaesse: check its type

ppaalanen:
I like standardizing how to find `wayland-scanner`. We have debated how to do that right in the past. The version of the scanner should be the same as the version of libwayland-client/server. Lower scanner version is ok too, but higher is not. I don't think we have ever actually enforced that anywhere.

Cross-compilation requires the user to build a build-arch version of wayland-scanner manually somewhere, and then get build systems to use that. The one thing one does not want is to have to set a per-project build option to point to the scanner. Anything else goes: PATH, build-arch .pc file, crossfile... whatever works.

mesonbuild/modules/wayland.py
```diff
@@ -0,0 +1,115 @@
...
+            raise MesonException('Second argument must be one of {}'.format(' '.join(sorted(valid_types))))
+
+        # map new argument to old arguments if necessary (so we only accept the
+        # newer versions), error if the wayland-scanner is too old for the
+        # requested argument
+        if self.old_scanner:
+            if type_ == 'private-code':
+                type_ = 'code'
+            elif type_ == 'public-code':
+                raise MesonException('public-code type requires wayland scanner >=1.15')
```
ppaalanen: Isn't this reversed?

See https://gitlab.freedesktop.org/wayland/wayland/commit/9b76def674f0a0b99f5850cc800017fc4e03af7d where `code` implies `public-code`, while `private-code` is a new feature.

mesonbuild/modules/wayland.py
```diff
@@ -0,0 +1,115 @@
...
+        if 'output' in kwargs:
+            output = kwargs['output']
+        else:
+            output_stem = PurePath(PurePath(input_).name).stem
+
+            if type_.endswith('code'):
+                output_suffix = 'c'
+            else:
+                output_suffix = 'h'
+            output ='{}.{}'.format(output_stem, output_suffix)
```
ppaalanen: Should the header case not also take client vs. server into account?

mesonbuild/modules/wayland.py
```diff
@@ -0,0 +1,115 @@
...
+
+        if isinstance(kwargs['input'], list):
+            if len(kwargs['input']) != 1:
+                raise MesonException('Only one input file allowed.')
+            input_ = kwargs['input'][0]
+        else:
+            input_ = kwargs['input']
+        if kwargs.pop('is_protocol', False):
+            p = self.protocols_dep.variable_method([], {'pkgconfig': 'pkgdatadir'})
+            input_ = (PurePath(p) / input_).as_posix()
```
ppaalanen: I find `is_protocol` wording confusing, because everything is about protocol here.

Is there a reason to make a shorthand for `wayland-protocols` specifically? Other protocol repositories exist too, and I'm ok with the shorthand for `wayland-protocols` (which might not always be required), but how does one use extensions from other projects?

E.g. Weston installs some XML files and announces the location via `libweston-MAJOR-protocols.pc` file `pkgdatadir`. There is also https://github.com/swaywm/wlr-protocols .

nirbheek:
Superseded by https://github.com/mesonbuild/meson/pull/9985.

PR #6562, opened 2020-01-31
ci: Add Fedora and OpenSUSE images

mensinda:
I have tried to fix as many test failures as possible on these images, however, I didn't manage to fix all of them. Specifically the `hdf5`, `netcdf`, `mpi` and `scalapack` tests just won't work on Fedora and OpenSUSE. Unfortunately, I don't know enough about these dependencies to fix them (ping @scivision, @dcbaker).

The docker images are hosted here: https://hub.docker.com/u/mensinda

### DIFF
--- a/.github/workflows/os_comp.yml
+++ b/.github/workflows/os_comp.yml
@@ -36,3 +36,27 @@ jobs:
       run: ./run_tests.py
       env:
         CI: '1'
+
+  fedora:
+    name: Fedora
+    runs-on: ubuntu-latest
+    container: mensinda/fedora:latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Run tests
+      run: ./run_tests.py
+      env:
+        CI: '1'
+        SKIP_SCIENTIFIC: '1'
+
+  opensuse:
+    name: OpenSUSE
+    runs-on: ubuntu-latest
+    container: mensinda/opensuse:latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Run tests
+      run: ./run_tests.py
+      env:
+        CI: '1'
+        SKIP_SCIENTIFIC: '1'
--- a/ci/ciimage/fedora/Dockerfile
+++ b/ci/ciimage/fedora/Dockerfile
@@ -1,0 +1,4 @@
+FROM fedora:latest
+
+ADD install.sh /usr/sbin/docker-fedora-install
+RUN docker-fedora-install
--- a/ci/ciimage/fedora/install.sh
+++ b/ci/ciimage/fedora/install.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+pkgs=(
+  python python-setuptools python-wheel python-pip python-pytest-xdist pygobject3 python3-devel python2-devel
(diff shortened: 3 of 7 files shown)


scivision:
I have noticed a few prior notes and instances of NetCDF or Scalapack causing CI problems. I don't think I have time in the near future to setup VMs for the various OS to fix these issues. So maybe it's better to skip NetCDF and Scalapack? Or even remove them from Meson? 

The long term plan is to "fix" these quirky HPC-oriented frameworks by using CMake from Meson via CMake `find_package(... COMPONENTS ...)` where I and others have worked about robust FindFoo.cmake over the years.  I think I will do that in 2Q2020. This will fix a lot of libraries in general.

mensinda:
Skipping the tests, when the dependency isn't found shouldn't be a problem. However, with `hdf5` and `netcdf` the really annoying problem is that they **are found** but **don't compile**. I can work around this by uninstalling the packages, but this would only hide the bug.

Also, CMake `COMPONENTS` support will be added with #6569.

scivision:
In general this type of Fortran error is caused by missing include directories:

```
    1 | use netcdf
      |    1
Fatal Error: Cannot open module file ‘netcdf.mod’ for reading at (1): No such file or directory
```

it may well be another case of messed up pkg-config files. 

In CMake FindFoo.cmake a typical technique is to:

1. find_library, find_path
2. check_fortran_source_compiles() using the libraries found

and only declaring the package found if both are OK. Maybe this is a pattern that could be in DependencyFactory?

mensinda:
I am not sure if the `DependencyFactory` can be rewritten to support a logical and. We could add a new `DependencyChain` class for this (only found if all dependencies are found). However, I am not sure if this would be overengineered. Just doing both checks in the system dependency class might be cleaner (at least for now).

scivision:
OK so do you mean within each dependency, do a `run_command` like check using Meson internals?

A typical example is Lapack, where many versions are not uncommon to be installed. There could be a for loop inside that tries to compile with each one until it finds one that compiles

mensinda:
Yes, for specific dependencies that require this (and would return a false positive otherwise). Or, we make the lookup good enough so this isn't required :)

scivision:
OK what I've been doing is for each project I have a meson_deps/meson.build that does like

```meson
cc = meson.get_compiler('c')
foreach l : ['foo','bar','baz']
   foo = dependency(l, required: false, disabler: true)
   if cc.links('foo.c', dependencies: foo)
     break
   endif
   foo = disabler()
endforeach
```

so far I just copy that project to project as needed. But OK that would be a pathway forward to do it internal to Meson dependency()

mensinda:
For now, I went forward with removing the relevant packages from the images and allowing the framework to skip the tests.

These changes should be undone once the dependencies are fixed.

scivision:
I think that's a good idea. Netcdf and Scalapack dependencies are a bit shaky for now on less common use cases. As a few PRs are merged, we'll have a better basis to make these dependencies more solid.

mensinda:
OK the CI is green, and this is now ready to merge from my perspective. Does anyone have any objections / missing packages in the images?

PR #6565, opened 2020-02-01
meson: performance optimize

marcelhollerbach:
With this commit, 2 performance bottlenecks are resolved.
- Caching the result of can_compile removes 11s of efl configure time
- Using a lookup dict instead of a list is also speeding the ninja generation up by several seconds. Additionally, using the knowledge if this flag is in pre or post, is shaving off a few seconds of runtime.

Additionally, the usage of --profile-self is fixed in this commit.

### DIFF
--- a/mesonbuild/compilers/compilers.py
+++ b/mesonbuild/compilers/compilers.py
@@ -731,6 +731,7 @@ class Compiler:
         return repr_str.format(self.__class__.__name__, self.version,
                                ' '.join(self.exelist))
 
+    @lru_cache(maxsize=None)
     def can_compile(self, src) -> bool:
         if hasattr(src, 'fname'):
             src = src.fname
--- a/mesonbuild/compilers/compilers.py
+++ b/mesonbuild/compilers/compilers.py
@@ -15,6 +15,7 @@
 import contextlib, os.path, re, tempfile
 import collections.abc
 import typing as T
+from functools import lru_cache
 
 from ..linkers import StaticLinker, GnuLikeDynamicLinkerMixin, SolarisDynamicLinker
 from .. import coredata
@@ -445,6 +446,9 @@ class CompilerArgs(collections.abc.MutableSequence):
                  iterable: T.Optional[T.Iterable[str]] = None):
         self.compiler = compiler
         self.__container = list(iterable) if iterable is not None else []  # type: T.List[str]
+        self.__seen_args = set()
+        for arg in self.__container:
+            self.__seen_args.add(arg)
 
     @T.overload                                # noqa: F811
     def __getitem__(self, index: int) -> str:  # noqa: F811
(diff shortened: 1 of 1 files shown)


xclaesse:
Looks good to me, gst-build went from 1.23m to 1.20m, so not really much in that case, but your case could be more intensive.

marcelhollerbach:
Cool! Thank you for your feedback, i will check tomorrow if i can find more performance bottlenecks :)

xclaesse:
I think one big improvement we could do, is adding more batch API like get_supported_arguments(), and parallelize those checks. I've always been wanting to do that.

xclaesse:
Gave it a try, almost too easy to be true: https://github.com/mesonbuild/meson/pull/6568

xclaesse:
oh interesting, when parallelizing this PR makes it actually slower. 

Both this PR and https://github.com/mesonbuild/meson/pull/6568 applied: 48s
Only https://github.com/mesonbuild/meson/pull/6568 applied: 32s

This is measured with simply `rm -rf builddir && time meson builddir`, but seems to give stable numbers.

marcelhollerbach:
I reworked now again the CompilerArgs thing a bit, as i found some slowness there due to the "in" usage on lists, i replaced it with a dict, which boosts things up another time. However, the continues removal and adding of this is rather annoying, maybe we can build up some tree like structure, and return a handmade iterator later on, where we walk this tree like structure of pre and posts, and just jump all the elements that we have already discovered earlier/later. But thats something for later on, if nothing else fails in the tests, i am going to leave it like it is now. (This should also again make gst-build faster :))

mesonbuild/compilers/compilers.py
```diff
@@ -445,6 +446,9 @@ def __init__(self, compiler: T.Union['Compiler', StaticLinker],
                  iterable: T.Optional[T.Iterable[str]] = None):
         self.compiler = compiler
         self.__container = list(iterable) if iterable is not None else []  # type: T.List[str]
+        self.__fast_loopup = {}
+        for arg in self.__container:
+            self.__fast_loopup[arg] = True
```
dcbaker: why don't we just replace `self.__container` with a `mesonlib.OrderedDict`, so we can get the log(n) lookup, but not have to keep two containers in sync?

marcelhollerbach: Mhm, i tried to give this a try. However, i cannot implement then `__getitem__` `__setitem__` etc.
But as i said in my commit for this, this is not the end, there can be way more to improve this, maybe without two containers, but a other data struct.

dcbaker: Okay. I'm still not crazy about having two data structures, but we probably need a different datastructure, maybe something like a trie? But that's for another time I guess.

marcelhollerbach: Yeah - i totally agree on that, I will not have much time in the next few months, i hope i can revisit this, and build the structure i was thinking about.

nirbheek:
I split out the `--profile-self` fix into https://github.com/mesonbuild/meson/pull/6611 so we can get it merged.

marcelhollerbach:
Sure, i am just wondering what is up with this PR, the CI failure is a timeout, where i am quite sure that i did not cause it (as ci on my repository passed). Is there anything else blocking this here?

nirbheek:
I will let @xclaesse answer that, since he's the one who has been reviewing this :)

jpakkane:
Maybe it was a fluke. Try rebasing to get CI to rerun.

marcelhollerbach:
@jpakkane Seems to have worked, thank you. Now the question is, is there anything left for this ?

jpakkane:
Why is it called a `loopup` and not `lookup`? In fact neither of those is a very good name, maybe something like `processed` or `seen_args` or something?

For clarity it should be a set, not a dictionary, since we only care about membership.

marcelhollerbach:
Hi! Thank you! I changed it to a set, and fixed the name.

mesonbuild/compilers/compilers.py
```diff
@@ -619,6 +632,7 @@ def __iadd__(self, args: T.Iterable[str]) -> 'CompilerArgs':
         Add two CompilerArgs while taking into account overriding of arguments
         and while preserving the order of arguments as much as possible
         '''
+        this_round_added = {} # a dict that contains a value, when the value was added this round
```
jpakkane: This should also be a set. Other than that looks good.

marcelhollerbach: Sorry that i did not catch that initially. Fixed now :)

PR #4904, opened 2019-02-09
Parallelize compiler.get_supported_* functions

Akaricchi:
I've taken the `concurrent.futures` approach because it required minimal changes to the code structure. I also had to introduce a lock object into `mlog` to prevent the debug log lines from different tests from interspersing.

Related to #3635

Issue #3635, opened 2018-05-26
Speed-up compiler checks by parallelizing where possible


nirbheek:
Some compiler checks can be parallelized, for example:

* `get_supported_arguments()`
* `get_supported_link_arguments()`
* `cross_compute_int()`: there are two branches at each guess, so we can run both at the same time, which will improve the worst-case time.

We should also add map-like API for batch-processing several `has_function()` and `has_header()` checks at once and sets values in a `configuration_data()` object, usually of the form:

```meson
cdata = configuration_data()
if cc.has_function(func_name, prefix: includes)
   cdata.set(func_name.to_upper().underscorify())
endif
```

Usually this requires lambda support (or similar) to be able to do this, but since the function in this case is almost always the same, we can hard-code it in Meson. It would be a good default in case we want to extend it later anyway.

xclaesse:
One simple way to parallelize has_header() and has_function() is to add get_supported_headers() and get_supported_functions(). That would definitely speedup that kind of foreach loop: https://gitlab.gnome.org/GNOME/glib/blob/master/meson.build#L260.

bluetech:
I have copied the list of 47 headers from @xclaesse's link to an empty project, hacked in parallelism using `concurrent.futures` and timed `meson setup`. The results I get (laptop with 2 cores, minimum of multiple runs) are:

- Same thread: `0m0.904s`.
- `ThreadPoolExecutor(1)`: `0m0.932s`.
- `ThreadPoolExecutor(2)`: `0m0.687s`.
- `ThreadPoolExecutor(3)` and up: same as 2 threads.

xclaesse:
Could you run that same project but with cc.compile('#include <@0@>'.format(header)) instead of cc.has_header(), please? That could give some useful data for issue #2246.

bluetech:
For reference, what I measured is an implementation of `get_supported_headers`.

Regarding whether parallelization is worth it, I am not sure. Since the checks are performed in an imperative fashion, there does not tend to be many batches to perform in parallel.

What is a good example of a meson project where the configure phase takes a long time? Maybe glib? I'd be interested to get a sense of how long `meson setup` of such a project takes.

tp-m:
[gst-build](https://cgit.freedesktop.org/gstreamer/gst-build/) with `--wrap-mode=forcefallback` perhaps

nirbheek:
> hacked in parallelism using concurrent.futures
...
[...]

PR #6568, opened 2020-02-01
Parallelize get_supported_arguments()

xclaesse:
From my testing this drops gst-build configure from 1.22m down to 1.09m.

### DIFF
--- a/mesonbuild/interpreter.py
+++ b/mesonbuild/interpreter.py
@@ -42,6 +42,7 @@ import collections
 from itertools import chain
 import functools
 import typing as T
+import concurrent.futures as conc
 
 import importlib
 
@@ -1615,10 +1616,15 @@ class CompilerHolder(InterpreterObject):
     @permittedKwargs({})
     def get_supported_arguments_method(self, args, kwargs):
         args = mesonlib.stringlistify(args)
-        supported_args = []
-        for arg in args:
-            if self.has_argument_method(arg, kwargs):
-                supported_args.append(arg)
+        with conc.ThreadPoolExecutor() as executor:
+            futures = []
+            for arg in args:
+                f = executor.submit(self.has_argument_method, arg, kwargs)
+                futures.append((f, arg))
+            supported_args = []
+            for f, arg in futures:
+                if f.result():
+                    supported_args.append(arg)
         return supported_args
 
     @permittedKwargs({})


xclaesse:
Even down to 32s when not applying PR https://github.com/mesonbuild/meson/pull/6565.

mesonbuild/interpreter.py
```diff
@@ -1615,10 +1616,15 @@ def has_multi_arguments_method(self, args, kwargs):
...
     def get_supported_arguments_method(self, args, kwargs):
         args = mesonlib.stringlistify(args)
-        supported_args = []
-        for arg in args:
-            if self.has_argument_method(arg, kwargs):
-                supported_args.append(arg)
+        with conc.ThreadPoolExecutor() as executor:
+            futures = []
+            for arg in args:
+                f = executor.submit(self.has_argument_method(arg, kwargs))
```
QuLogic: How can this even work? You just submitted the result of the check to the thread pool, not the check itself.

xclaesse: It submit them all, then call result() on each. That should block until they complete, no?

xclaesse: Forget me, I see what you mean... Why does it make a big difference in time... Maybe I'm just to tired... :P

xclaesse: Voilà fixed, the gain is much less impressive, but still 10s less. Should not code during the evening after FOSDEM, too tired :P

xclaesse:
Note that many places in GStreamer are not using get_supported_args() but should. I'll see if that can improve the gain here.

My previous numbers where wrong because a few optional subprojects actually where backtracing and thus skipped which was speeding the configure of course.

marcelhollerbach:
Mhm, i have tested efl build:
- 6565 & this: 22s
- only 6565: 26s
- only this: 35s

I think it is quite a good idea to go with both?
EFL does not use get_supported_args much, so this will not have a that big impact. EFLs configure time is most of the time dominated by processing giant source lists, which is why can_compile has such a big impact for us. (Maybe i can refactor things in add_deps, to only process the sources of a dependency once ...)

mesonbuild/interpreter.py
```diff
@@ -1615,10 +1616,15 @@ def has_multi_arguments_method(self, args, kwargs):
     @permittedKwargs({})
     def get_supported_arguments_method(self, args, kwargs):
         args = mesonlib.stringlistify(args)
-        supported_args = []
-        for arg in args:
-            if self.has_argument_method(arg, kwargs):
-                supported_args.append(arg)
+        with conc.ThreadPoolExecutor() as executor:
```
nirbheek: concurrent.futures has a lot of overhead, and it's not really needed either. We're just running processes and waiting for them to exit. We already benchmarked all the various methods in https://github.com/mesonbuild/meson/issues/3635 and `subprocess.Popen` is twice as fast as this or async/await.

nirbheek: Also the same approach was discussed in https://github.com/mesonbuild/meson/pull/4904. I really think we should reorganize the code to allow using Popen instead of trying to keep the changes minimal and picking a parallelization method with a lot of overhead.

xclaesse: Totally forgot about those 2, thanks! I noticed that gstreamer has many places where it's not using get_supported_args(), need to fix that first to get proper benchmark numbers.

xclaesse:
This PR is superseded by https://github.com/mesonbuild/meson/pull/6664.

PR #6569, opened 2020-02-01
WIP: cmake: Boost CMake dependency backend and COMPONENTS support

mensinda:
Adds the new boost CMake dependency backend. It is a lot smaller than the existing boost mechanism and should also be a lot more reliable due to the mature CMake ecosystem.

Additionally, support for specifying `find_package()` `COMPONENTS` is added. This new feature is required for the `BoostCMakeDependency`.

### DIFF
--- a/docs/markdown/Dependencies.md
+++ b/docs/markdown/Dependencies.md
@@ -173,6 +173,10 @@ it automatically.
     cmake_dep = dependency('ZLIB', method : 'cmake', modules : ['ZLIB::ZLIB'])
 ```
 
+Support for adding additional `COMPONENTS` for the CMake `find_package` lookup
+is provided with the `components` kwarg (*introduced in 0.54.0*). All specified
+componets will be passed directly to `find_package(COMPONENTS)`.
+
 It is also possible to reuse existing `Find<name>.cmake` files with the
 `cmake_module_path` property. Using this property is equivalent to setting the
 `CMAKE_MODULE_PATH` variable in CMake. The path(s) given to `cmake_module_path`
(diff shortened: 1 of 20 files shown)


docs/markdown/Dependencies.md
```diff
@@ -173,6 +173,10 @@ it automatically.
     cmake_dep = dependency('ZLIB', method : 'cmake', modules : ['ZLIB::ZLIB'])
 ```
 
+Support for adding additional `COMPONENTS` for the CMake `find_package` lookup
+is provided with the `components` kwarg (*introduced in 0.54.0*). All specified
+componets will be passed directly to `find_package(COMPONENTS)`.
+
```
dcbaker: In the past we've mapped this to the existing `modules` arugment instead of `components`, even if cases (like llvm-config) when `components` is what the underlying tool calls them. I'd like to keep that so we don't have two arguments that do the same thing

mensinda: I agree that this makes semantically more sense, however, we are currently using `modules` already for selecting the imported CMake targets (think `Zlib::Zlib` or `KF5::KTextEditor`). Changing the meaning of the `modules` kwarg would be a breaking change (possible to do with a deprecation period and a concrete plan, but out of the scope of this PR).

If I were to write the CMake backend from scratch, I would also use `modules` for `COMPONENTS` and add an extra kwarg for CMake targets.

scivision:
as a reference, here's the work I did for adding CMake COMPONENTS

https://github.com/mesonbuild/meson/compare/master...scivision:cm_comp

docs/markdown/Dependencies.md
dcbaker: ahhhhh. Okay, that makes sense. We can leave this as-is then

mensinda:
OK, I am giving up. Even CMake can't find boost so that it just works. At this point, it is easier to rewrite the system dependency...

PR #6574, opened 2020-02-03
Cleaning project readme

michaelbadcrumble:
Cleaning the project readme, and untangling Information that are not relevant, added some links to the Meson build system website.  Hopefully, my changes did not make things more confusing.

Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ or by typing `meson --help`.
 #### Contributing
 
 We love code contributions. See the [contribution
-page](https://mesonbuild.com/Contributing.html) on the web site for
+page](https://mesonbuild.com/Contributing.html) on the website for
 details.
 
 
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ the Python 3 version of `pip`.
 
 If you wish you can install it locally with the standard Python command:
 
-```sh
+```console
 python3 -m pip install meson
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -58,27 +59,22 @@ You can omit either of the two directories, and Meson will substitute
...
 To compile, cd into your build directory and type `ninja`. To run unit
 tests, type `ninja test`.
 
-Install is the same but it can take an extra argument:
-
-`DESTDIR=/destdir/path ninja install`
-
-`DESTDIR` can be omitted. If you are installing to system directories,
-you may need to run this command with sudo.
-
```
dcbaker: I'm not sure that dropping the note about DESTDIR is a good idea, specifically because meson uses DESTDIR as an environment variable, not as an argument to make like you'd do with autotools. It could certainly be *clearer*, but I'm not in favor of dropping it.

README.md
```diff
@@ -58,27 +59,22 @@ You can omit either of the two directories, and Meson will substitute
 the current directory and autodetect what you mean. This allows you to
 do things like this:
 
-`cd source_root; mkdir builddir; cd builddir; meson ..`
-
-or
-
-`cd source_root; mkdir builddir; meson builddir`
+```zsh
```
dcbaker: I think this should be `console` not `zsh`

README.md
michaelbadcrumble: I understand that but there is already some documentation on the DESTDIR environment variable, so that is why I added a link to the [running Meson page](https://mesonbuild.com/Quick-guide.html#using-meson-as-a-distro-packager).

README.md
michaelbadcrumble: sorry about that making the quick fix.

michaelbadcrumble: ok change made. But I accidentally forgot the [skip ci] in the commit message, so sorry.

Issue #6586, opened 2020-02-05
Meson can't import dependency ddbc from dub


Rogni:
**Describe the bug**
Meson can't import dependency ddbc from dub. Perhaps the reason is the library dependencies and their location ( https://github.com/mesonbuild/meson/blob/master/mesonbuild/dependencies/base.py#L1682 find libs only in [build_name](https://github.com/mesonbuild/meson/blob/master/mesonbuild/dependencies/base.py#L1682) folder ) 

**To Reproduce**
https://github.com/Rogni/meson_debug_dub sample project

```
git clone https://github.com/Rogni/meson_debug_dub.git meson_debug_dub
git clone https://github.com/mesonbuild/meson.git meson
dub fetch ddbc
dub build ddbc
cd meson_debug_dub
DC=ldc2 python3 ../meson/meson.py build
```

https://travis-ci.org/Rogni/meson_debug_dub/builds/646334749#L318 travis log
(i'm added warning to fork of meson if library in DubDependency not found https://github.com/Rogni/meson/blob/debug_dub/mesonbuild/dependencies/base.py#L1649 ) 

**Expected behavior**
ddbc founded as dependency

**system parameters**
* Is this a [cross build](https://mesonbuild.com/Cross-compilation.html) or just a plain native build (for the same computer)? 
plain native build
* what operating system (e.g. MacOS Catalina, Windows 10, CentOS 8.0, Ubuntu 18.04, etc.) 
Mint 19.1 tessa
* what Python version are you using e.g. 3.8.0 
3.6.9
* what `meson --version`
0.53.0 from python3-pip
* what `ninja --version` if it's a Ninja build
1.9.0.git.kitware.dyndep-1.jobserver-1

Rogni:
In host device ( Mint 19.1 tessa )
```
$ DC=ldc2 python3 ../debug_dub_in_meson/meson.py build
The Meson build system
Version: 0.53.999
Source dir: /home/mak/Рабочий стол/Dlang/debug_meson_project
Build dir: /home/mak/Рабочий стол/Dlang/debug_meson_project/build
Build type: native build
Project name: TestDubDependency
Project version: undefined
D compiler for the host machine: ldc2 (llvm 1.8.0 "LDC - the LLVM D compiler (1.8.0):")
D linker for the host machine: ldc2 GNU ld.bfd 2.30
Host machine cpu family: x86_64
Host machine cpu: x86_64
[...]

PR #6587, opened 2020-02-05
Fix import dependency from dub

Rogni:
#6586 
fix DubDependency._find_right_lib_path
if default_path is path to library and directory  of lib don't have subdirectory '.dub/build' -> return default_path

### DIFF
--- a/mesonbuild/dependencies/base.py
+++ b/mesonbuild/dependencies/base.py
@@ -1646,6 +1646,7 @@ class DubDependency(ExternalDependency):
                     if lib_path:
                         self.link_args.append(lib_path)
                     else:
+                        mlog.warning('Library not found: {}'.format(file))
                         self.is_found = False
 
     def get_compiler(self):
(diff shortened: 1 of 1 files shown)


jpakkane:
Would be possible to add a test for this?

You should also squash the commits here, the latter commit completely undoes the first one, there is no point in having it.

Rogni:
in last commit also i'm added checking default_path. its try to fix feature dependency (method: dub) if subdependency have fullpath to lib in target/linkerFiles ( https://github.com/Rogni/meson/blob/debug_dub/mesonbuild/dependencies/base.py#L1645 ). it was discussed in #6586

jpakkane:
The first commit has only this:

   +                         mlog.warning('Library not found: {}'.format(file))

and the second commit has this:

    -                        mlog.warning('Library not found: {}'.format(file))

Which is the exact opposite. You can see this in the entire changelist, since it only contains the other parts of the second commit.

Rogni:
in the first commit, I added a warning for the log in travis build https://travis-ci.org/Rogni/meson_debug_dub/builds/646334749#L318 and on local device. after fix this warning is no longer needed

jpakkane:
Thanks.

Issue #3005, opened 2018-02-02
exclude_files as a regex for install_subdir ?


Salamandar:
Everything is in the title ;)
Right now the relative path is kinda restrictive.

infirit:
I think unix-style glob patterns in combination with ``fnmatch`` would imho be a much more appropriate solution. ``fnmatch.filter`` can also be called on an iterable like the lists return by ``os.walk`` but this requires a little bit more work. I may have a go and add this to minstall.py next week.

PR #6590, opened 2020-02-05
Use unix style glob patterns to exlcude file and dirs for install_subdir

infirit:
This is regarding #3005. Regular expressions is way overkill but if people can come up with a use case that can not be handled by unix style glob patterns this should be revisited.

I tried to ask on irc but it seems dead but, the test for install_subdir appear to not test if the result actually match what is expected. If I am wrong can someone point me how it does this? unittests as they are now pass so at least there is not regression :wink: 

I am however not sure how I can have meson throw a warning when a glob pattern is used before version it is introduced. I can add it as part of the docs.

My specific use case is that I like to exclude Makefile.am from a fairly large(ish) project. Instead of specifying 20 or so lines in exclude file I wanted to just say ``*Makefile.am``


QuLogic:
Installed files should be compared to the list in `installed_files.txt`.

PR #6591, opened 2020-02-06
mesonlib: Set stdin to DEVNULL for all programs run by us

nirbheek:
Otherwise there's a high likelihood that some program run by us will mess up the console settings and break ANSI colors. F.ex., running `uname` in the Visual Studio 2019 x86 developer prompt using `run_command()` does this.

### DIFF
--- a/mesonbuild/mesonlib.py
+++ b/mesonbuild/mesonlib.py
@@ -1084,10 +1084,15 @@ def Popen_safe(args: T.List[str], write: T.Optional[str] = None,
     if 'stdin' not in kwargs:
         kwargs['stdin'] = subprocess.DEVNULL
     if sys.version_info < (3, 6) or not sys.stdout.encoding or encoding.upper() != 'UTF-8':
-        return Popen_safe_legacy(args, write=write, stdout=stdout, stderr=stderr, **kwargs)
-    p = subprocess.Popen(args, universal_newlines=True, close_fds=False,
-                         stdout=stdout, stderr=stderr, **kwargs)
-    o, e = p.communicate(write)
+        p, o, e = Popen_safe_legacy(args, write=write, stdout=stdout, stderr=stderr, **kwargs)
+    else:
+        p = subprocess.Popen(args, universal_newlines=True, close_fds=False,
+                             stdout=stdout, stderr=stderr, **kwargs)
+        o, e = p.communicate(write)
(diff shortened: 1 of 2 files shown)


mesonbuild/mesonlib.py
```diff
@@ -1080,6 +1080,10 @@ def Popen_safe(args: T.List[str], write: T.Optional[str] = None,
                **kwargs: T.Any) -> T.Tuple[subprocess.Popen, str, str]:
     import locale
     encoding = locale.getpreferredencoding()
+    # Redirect stdin to DEVNULL otherwise the command run by us here might mess
+    # up the console and ANSI colors will stop working on Windows.
+    if 'stdin' not in kwargs:
+        kwargs['stdin'] = subprocess.DEVNULL
```
xclaesse: We have many more places where we use Popen and friends directly instead of this wrapper, I think. Should we take this opportunity to convert them all to use a meson wrapper function? I remember you did the same change for git invokations in wrap.py recently.

xclaesse: btw, for git you also added that hack:

```python
    if platform.system().lower() == 'windows':
        mlog._windows_ansi()
```

Sounds like something we should write once, and use everywhere.

dcbaker: Shouldn't that just happen in the mlog module at the top scope?

nirbheek: I was trying to not call `mlog._windows_ansi()` on every subprocess invocation since that's already what we spend the most amount of time on during configure.. I'll measure it and get some actual numbers.

Closing stdin is good hygiene for `run_command()` anyway to avoid the program expecting something on stdin and blocking configure.

The mlog module only calls `_windows_ansi()` only once, while the problem here is that the console mode gets reset by subprocesses sometimes, and the way to detect that also involves setting it, so we can just set it.

jpakkane:
LGTM, if other people don't have outstanding issues, feel free to merge.

mesonbuild/mesonlib.py
nirbheek: On both my old Windows machine and my new Windows machine, calling `mlog._windows_ansi()`  100,000 times in a loop takes ~2-3 seconds. That's basically negligible. I'll call it on every subprocess invocation.

PR #6595, opened 2020-02-06
Revert "Make 'default_library' per-subproject builtin option"

jpakkane:
This reverts commit 3ba0073df6014d4b594df052f8e1fa75ce17b20e.

It was merged without authorization and with open issues. Putting this here as a reminder that the issues must be fixed before the release or they will be reverted.


xclaesse:
That's not true.

jmansion:
It really sucks that this is reverted.
At the moment, I'm having to make lots of local changes to the protobuf wrap which does not work with MSVC on Windows because it does not have all the necessary defines to make sure the imports and exports are set up.  Consequently the DLLs build but it has no exported symbols and MS link does not create an import lib - so then the whole thing falls to bits.
I'd argue that actually a static library is a much safer default anyway and its certainly policy at some big C++ users (eg Bloomberg for a start) to use dynamic linking during dev only and deploy a static binary.
This change (and TBH an ability to pass other overrides to subprojects) is very handy, especially where this sort of change in dynamic library behaviour and/or symbol visibility is important.

dcbaker:
And the policy for most of the projects I work on is you cannot under any circumstances statically link an external dependency, so no matter what default we pick someone is going to be unhappy.

I'm trying to get this fixed so that you can set the link mode using the command line and native/cross files instead of doing it in the source language. I *think* it's just the source language changes that are going to be reverted. We're also planning to have a site-wide auto-loaded configuration, with those two things in place you should be able to get the behavior you want of statically linking all subprojects by default.

xclaesse:
This is not going to be reverted.

Nobody has objection on having default_library per subproject, @jpakkane only objected that we should make MORE options per subproject. This PR added the needed infrastructure to make it happen for more options (e.g. c_std, werror, etc) but it would still require more work (I have a WIP branch). I think we can do it step by step instead of waiting to get everything ready in one shot.

@dcbaker with Meson master you can already set it from command line which would override anything set in build definition.

jpakkane:
The objection I had has more to do with the UI we put in for this. Having a gazillion locations for it is not really nice because it is easy to get confused. Having all the options in one place, in the cross/native file as per #6597 is nicer because that centralizes them in one place.

Yes, I am aware that this stance is contrast to how default options for projects work. That design seemed to make sense when I wrote it but in hindsight maybe it was not that great after all. :(

xclaesse:
I explained already why it is not enough to define those options in native/cross file. Again, it is rarely user's choice.

- A project could want to statically link a subproject by default.
- A project often knows that one of its subprojects does not build with werror.
- The C/C++ std is a property of each (sub)project, not something the user can decide.

So defining those per-subproject built-in option from the native/cross file is a valid use-case but is not the only one. I personally think it's actually the less common one. Limiting those options to only be settable from native/cross file is a n arbitrary limitation that introduce a huge inconsistency. And so far everyone who commented on this matter agreed with that.

jpakkane:
> A project could want to statically link a subproject by default.

You can't really rely on this. Suppose your has subprojects A and B, both of which depend on C. A wants to build it shared and B wants to build it static. Which one will be chosen? The one whose subproject was ran first. These sorts of cases can only be reasonably solved with an override in the native/cross file which won't be in the next release.

> The C/C++ std is a property of each (sub)project, not something the user can decide.

This is true for C, not so much for C++, where people do questionable things like have ifdefs in their headers that depend on the chosen C++ version. Those are (sadly) ODR violations waiting to happen. For C setting the version in each subproject is the default thing you want to do, for C++ it is the opposite and you'd want to impose the same version over all subprojects. This makes it hard to design a good UI for both.

xclaesse:
> > A project could want to statically link a subproject by default.
> 
> You can't really rely on this. Suppose your has subprojects A and B, both of which depend on C. A wants to build it shared and B wants to build it static. Which one will be chosen? The one whose subproject was ran first. These sorts of cases can only be reasonably solved with an override in the native/cross file which won't be in the next release.

First one win, it has always been like that for subproject() and dependency() default_options kwarg. It is also like that when you have foo 1.0 on your system and you have 2 calls to dependency('foo') for version >=1.0 and >=1.1. In practice that's a corner case that rarely happens, but it's main project's responsibility to deal with the order of its subprojects. User can already override those per-subproject options using `-D` command line which is completely consistent with any other option. Doing it in the cross/native file would be nice, but not essential at all, there is zero reason those option are any different to others.

> > The C/C++ std is a property of each (sub)project, not something the user can decide.
> 
> This is true for C, not so much for C++, where people do questionable things like have ifdefs in their headers that depend on the chosen C++ version. Those are (sadly) ODR violations waiting to happen. For C setting the version in each subproject is the default thing you want to do, for C++ it is the opposite and you'd want to impose the same version over all subprojects. This makes it hard to design a good UI for both.

I'm not not a C++ expert, but many people commented already saying mixing C++ versions is actually most of the time perfectly fine. It makes perfect sense for a master project to limit itself to C++11 but have optional subprojects that require C++14, the user will never know such detail. Similar stories are even more common in C world, happens concretely in many GNOME projects.

TheQwertiest:
>  It makes perfect sense for a master project to limit itself to C++11 but have optional subprojects that require C++14, the user will never know such detail

Doubly so for subprojects that are shared libraries with C interface: they might have entirely different compilation flags or even entirely different underlying language (e.g. a Rust library with C interface).

Issue #3001, opened 2018-01-31
Generic overrider functionality


jpakkane:
There are a ton of niggly one-off things people want to do which are reasonable to do but tricky and laborious to do by editing build definitions. Some edits can not even be done because they would touch subprojects. Examples of these include:

 - changing the build type of some target to be shared or static
 - change optimization level of one target (because, for example, compiler bugs)
 - install or not install files from a given target or an entire subproject
 - etc

All of these should be doable without changing the build definition files. One solution would be to add a Meson switch to read and use a file with these definitions.

dcbaker:
@jpakkane, I'm interested in getting this done sooner than later, so I'm willing to work on it. Do you have a specific format in mind for this?

jpakkane:
There are really only three design requirements:

 - must be human readable
 - must be human writable
 - should be some file format we already know how to parse if possible

This would probably need to be some sort of a hierarchical file format. Is INI file sufficient for this? I can't really tell. However one should be able to do at least the following:

 - set the default value for an option
 - override the value of an option for all subprojects
 - override the value of an option for a specific subproject
 - override the value of an option for a specific target in a specific subproject

A good example for this would be the language standard version. People have expressed requirements to build some subprojects always with `gnu99` or somesuch regardless of what the superproject uses.

Similarly you should be able to override `shared_library`, `static_library` and `library` (at least) to instead build some other kind of library type.

germandiagogomez:
This functionality would be very welcome by me after fighting today 2 of them (not the optimization one) and finding it was not possible to do anything about it :)

xclaesse:
If we go with the "per-subproject builtin options" solution, the only thing missing here is a "per-target options" which could be fixed by adding `library(..., default_options:['optimization=0'])`.

IMHO, that solution seems easier to implement and rely on existing option infrastructure instead of inventing another file format.

xclaesse:
Actually, I just discovered that we already have override_options kwarg. So per-subproject options is the only thing we miss here. Unless there is something I missed?

jpakkane:
The use case I had in mind (for the generic overrider) was overriding functionality in subprojects _without needing to change the build definition files_  of said subproject. Like suppose a subproject has `shared_library(...)` but you, for whatever reason, would want it to be `static_library(...)` so you could change it with an override file only and the original code remains immutable.

[...]

PR #5489, opened 2019-06-13
Compiler options per lang

Ericson2314:
A current rather untyped storage of options is one of the things that
contributes to the options code being so complex. This takes a small
step in synching down by storing the compiler options in dicts per
language.

Future work might be replacing the langauge strings with an enum, and
defaultdict with a custom struct, just like `PerMachine` and
`MachineChoice`.

PR #6597, opened 2020-02-07
Set project and meson options in cross/native files

dcbaker:
This is a partial implementation of #3001, but not everything.

This series allows meson builtins and project options to be specified in the cross and/or native files. This should make it easier to set up subprojects in a reliable way, as all of the options can be kept in a native/cross file. With file layering a projects can even ship preconfigured options, that the end user can overlay with their own cross/native files.

For mesa we'd like to couple this with the ability to set some builtin options (namely default_library) on a per-subproject basis to control wraps in a predictable way.


dcbaker:
@marc-h38, I think this was what you were wanting.

xclaesse:
I haven't read the whole thing yet, but one thing I'm wondering is why do you split options in different sections?

I was expecting a simpler format, like the one we already use in `<builddir>/meson-private/cmd_line.txt`. For example:
```ini
[options]
prefix = '/opt/gnome'
default_library = 'both'
ffmpeg:default_library = 'static'
glib:gtk_doc = 'disabled'
```
That way you can load that into a single dict that would match exactly as if the user passed them with `-D`.

I think one common use-case is when you have only a cross file, and I think you should be able to set all options from there. But then what happens if you have both a cross file and a native file with options defined in both? I guess we should abort with an error if the same option is set in both files. That would be similar to what we currently do when passing both `--default-library` and `-Ddefault_library`. Of course `build.` options must only be in cross file.

dcbaker:
@xclaesse I really want to get rid of the one off format in cmd_line.txt and replace it with a native/cross file.

> I haven't read the whole thing yet, but one thing I'm wondering is why do you split options in different sections?

Composition, for one. cross and native files can be composed by passing the argument more than once. Since each component is it's own section you can re-use a cross file in more than one build. Say you need zlib in two different projects, and you want to statically link it in both cases, now you can have a single zlib file.

For meson builtin versus project arguments it makes sense to split them as well, because they're conceptually different. Project arguments by definition apply only to one project, meson builtins (at least generally) apply to all projects, so being able to specify them separately makes sense to me.

> But then what happens if you have both a cross file and a native file with options defined in both? 

This is clearly defined. If you have both a cross and native file `project options` from the native file are ignored, `meson options` that have a per-machine value like `pkg_config_path` are read from the native-file and applied to the build machine. values from the cross file are applied to the host arguments.

jpakkane:
I'm so glad you posted this, because I was going to start working on it myself today. :)

It would be great to get all functionality needed for configuration so we can merge #6595 to get a general solution for all options rather than a one-off feature for one of them. One possible syntax change comes to mind. Rather than doing this:

```ini
[meson options]
c_std = 'c99'
[zlib:meson options]
default_library = 'static'
```

could we instead do this:


```ini
[meson options]
c_std = 'c99'
zlib:default_library = 'static'
```

dcbaker:
I thought about that, but I really dislike it because it makes things more verbose if you have more than one options:

```ini
[meson options]
zlib:foo = 'bar'
zlib:bar = 'foo'
zlib:thing = false
```

vs

```ini
[zlib:meson options]
foo = 'bar'
bar = 'foo'
thing = false
```

jpakkane:
> I'm not sure exactly what parts you don't like, but all of the machinery he put in place except in the interpreter changes is needed to make per-subproject builtins work

Per-project values of options should only be settable via the native/cross file. There are usually more than one of these, so concentrating them all in one place makes it easier to review.

dcbaker:
> Per-project values of options should only be settable via the native/cross file. There are usually more than one of these, so concentrating them all in one place makes it easier to review.

I think we should be able to set them via the command line. We allow setting project options on the command line, and we're going to get no end of complaints about `-Dsub:foo=c` works but `-Dsub:c_std=c99` doesn't.

nirbheek:
> We allow setting project options on the command line, and we're going to get no end of complaints about `-Dsub:foo=c` works but `-Dsub:c_std=c99` doesn't.

It's also going to be very inconsistent to not be able to set arguments on the command-line like that. We already have a bunch of unavoidable inconsistency in our argument handling, let's not increase it.

Setting arguments on the command-line vs a file is also much clearer for CI purposes since it's _right there_ in the yaml file, and it shows up in the CI logs for debugging.

scivision:
My use cases are a simpler subset. It's really useful to be able to set all these in the native/cross files. 
Would this make it possible to set the Ninja  binary location as well, to avoid use of NINJA or PATH environment variable?

dcbaker:
@scivision That would be a separate change, because I don't think we need any infrastructure changes for that, we just need to find ninja using the `find_program` implementation.

xclaesse:
> ```ini
> [meson options]
> c_std = 'c99'
> zlib:default_library = 'static'
> ```

I still strongly prefer that syntax over splitting into different sections. We use the `subproject:opt` syntax everywhere, it's consistent with command line options and with output of `meson configure`. It also can trivially be parsed into a dict used as default values to `options.cmd_line_options`, we already have that code when we parse cmd_line.txt.

I also do not agree this supersede cmd_line.txt, we still have to remember options set from command line and from `meson configure -Dfoo=bar`. Both concepts are complementary, and IMHO should share as much code and file format as possible.

Ultimately what I would like to see happening is projects shipping a set of cross files with their source code, that contains default values to build for specific targets (e.g. Android NDK, Linux distro mingw, etc). That often means specific values for a set of options, but still allowing them to be overriden from command line by the user.

xclaesse:
Also, nitpicking, but I don't think we need "meson" in `[meson options]`, we know it's a meson file. We don't have `[meson binaries]`, etc.

marc-h38:
> Also, nitpicking, but I don't think we need "meson" in [meson options], we know it's a meson file.

While pretty much everything around here is meson-related and the term "meson" is almost as vague and overloaded as "build", I suspect `[meson options]` was a relatively specific hence useful reference to `meson_options.txt`, was it not?

xclaesse:
> I suspect `[meson options]` was a relatively specific hence useful reference to `meson_options.txt`, was it not?

meson_options.txt only contains project options, so that would be a misleading reference since in the current proposal `[meson options]` actually does NOT contain project options, afaiu.

marc-h38:
> `meson_options.txt` only contains project options,

Mmmm... so this PR wants to add `[project options]` and `[meson options]` and out of these two the one that is more closely related to `meson_options.txt` is... `[project options]`?!

> `meson_options.txt` only contains project options,

... yet `meson_options.txt` is not mentioned in https://mesonbuild.com/Reference-manual.html#project , only in https://mesonbuild.com/Build-options.html ?!

Am I really the only one worried about the concepts and terminology? As a sanity check, do you guys try from time to time to discuss these terms and concepts with developers not familiar with meson?

xclaesse:
> Am I really the only one worried about the concepts and terminology? As a sanity check, do you guys try from time to time to discuss these terms and concepts with developers not familiar with meson?

I'm glad you volunteer to improve our documentation, there is certainly a lot to do :D

marc-h38:
@xclaesse to a reasonable extent I do volunteer to improve the documentation, in fact I already started that a little bit in other places. However no one can improve the documentation if developers use the terms "meson options" and "project options" in one place but vice-versa in another (really?), in that case it's a lost cause.
Before writing any documentation, the design must have clearly defined concepts and _language_ first. Super vague words like "build" and "meson" are not good names to start with. I also volunteer to find and suggest names (if not too late) but the high-level concepts must be summarized in a few sentences first, otherwise the concepts are flawed anyway.

xclaesse:
I never checked what the doc says exactly. From `meson configure` there are those terms: `Core options`, `Backend options`, `Base options`, `Compiler options`, `Directories`, `Testing options`, and `project options`. core+directories+testing are often referred as builtin options (that's how they are named in the code) because those options are always present.

meson_options.txt are **project options** because they are the options defined by the project, and not defined by meson itself.

dcbaker:
> I still strongly prefer that syntax over splitting into different sections. We use the subproject:opt syntax everywhere, it's consistent with command line options and with output of meson configure. It also can trivially be parsed into a dict used as default values to options.cmd_line_options, we already have that code when we parse cmd_line.txt.

And I'm strongly in favor of having separate sections for each project. I think it makes sense because of composition (ie, mesa will probably ship a config file for just zlib for windows, but unless you're doing a windows->windows build you'll need to compose it with another cross file) and it makes it clear, at a glance, exactly what projects are being modified by the file.

> Also, nitpicking, but I don't think we need "meson" in [meson options], we know it's a meson file. We don't have [meson binaries], etc.

How about "builtin options" then? I'd like to make it clear that one section is for meson level arguments, and the other is for options defined in your meson_options.txt, which I guess should be called "build options" as it is in the docs?

xclaesse:
> And I'm strongly in favor of having separate sections for each project. I think it makes sense because of composition (ie, mesa will probably ship a config file for just zlib for windows, but unless you're doing a windows->windows build you'll need to compose it with another cross file) and it makes it clear, at a glance, exactly what projects are being modified by the file.

Composition works just fine with a single section, they will be merged.

> > Also, nitpicking, but I don't think we need "meson" in [meson options], we know it's a meson file. We don't have [meson binaries], etc.
> 
> How about "builtin options" then? I'd like to make it clear that one section is for meson level arguments, and the other is for options defined in your meson_options.txt, which I guess should be called "build options" as it is in the docs?

The simple fact that it's hard to name is already a proof we should not split options, because whatever name we pick it will not be understood by the user without reading long documentation. If we insist on splitting options, it should at least follow the naming we have in `meson configure`, so I would call them `[core options]` and `[project options]`. For subproject options maybe e.g. `[zlib options]`.

dcbaker:
> Composition works just fine with a single section, they will be merged.

Certainly, but if you have three files

```ini
[zlib:builtin options]
c_std = 'c99'
c_args = ['-someopt']
```
```ini
[expat:builtin options]
c_std = 'gnu99'
```
```ini
[meson options]
c_std = 'c11'
cpp_std = 'c++11'
```

It's super clear what is being done in each one, the first only *only* affects zlib, I can see that just by reading the section header, the second one only affects expat, and the thrid only affects the primary project. compare that to:

```ini
[builtin options]
zlib:c_std = 'c99'
zlib:c_args = ['-someopt']
```
```ini
[builtin options]
expat:c_std = 'gnu99'
```
Where I can't tell at a glance that a file affects more than one project, but lets have a bigger more complicated example:

```ini
[zlib:builtin options]
c_std = 'c99'
c_args = ['-foo', '-bar', 'something else']
b_ndebug = true
b_lto = false

[expat:builtin options]
c_std = 'gnu99'
c_args = ['-foo']
c_link_args = ['-Wl,--arg']
b_ndebug = true

[builtin options]
b_ndebug = true
c_std = 'c11'
```

vs

```ini
```ini
zlib:c_std = 'c99'
zlib:c_args = ['-foo', '-bar', 'something else']
zlib:b_ndebug = true
zlib:b_lto = false
expat:c_std = 'gnu99'
expat:c_args = ['-foo']
expat:c_link_args = ['-Wl,--arg']
b_ndebug = true
c_std = 'c11'
```

The first way is much more readable to me, and it forces you to keep per-project options sorted, you would end up with
```ini
[builtin options]
c_std = 'c99'
zlib:b_ndeug = false
c_args = ['foo', 'bar']
c_link_args = ['-Wl,-something']
expat:c_std = 'c11'
zlib:c_args = ['-foo']
expat:b_ndebug=false
```

> The simple fact that it's hard to name is already a proof we should not split options, because whatever name we pick it will not be understood by the user without reading long documentation. If we insist on splitting options, it should at least follow the naming we have in meson configure, so I would call them [core options] and [project options]. For subproject options maybe e.g. [zlib options].

That's a terrible argument. Naming things is one of the hardest things in computer science, by that logic no software should be written at all!

But I don't think we're going to solve this by arguing about it anymore, clearly I like blue bikesheds and you like yellow ones. Maybe some other people have opinions?

xclaesse:
> ```ini
> zlib:c_std = 'c99'
> zlib:c_args = ['-foo', '-bar', 'something else']
> zlib:b_ndebug = true
> zlib:b_lto = false
> expat:c_std = 'gnu99'
> expat:c_args = ['-foo']
> expat:c_link_args = ['-Wl,--arg']
> b_ndebug = true
> c_std = 'c11'
> ```

Exactly what I have in mind, pretty readable to me, it's a matter of coding style, like in any language. It's also exactly what I already have in script that calls meson with all options set.

> But I don't think we're going to solve this by arguing about it anymore, clearly I like blue bikesheds and you like yellow ones. Maybe some other people have opinions?

Indeed, I think it's pretty clear we'll never agree, so we need others to vote :-)

marc-h38:
> I'm glad you volunteer to improve our documentation, there is certainly a lot to do :D

Monday morning, fresh mind let's give it a go.


- ~~Meson~~ **Built-in options**: options all meson users get whether they use them or not
https://mesonbuild.com/Builtin-options.html


- ~~Build~~, ~~Project~~ **User options**: options created and named by (a certain type of) meson users: the people who design `meson.build` and `meson_options.txt` files.
https://mesonbuild.com/Build-options.html

I found the name "User" in the first sentence of that page, didn't take me much effort. "User" is IMHO a relatively good negation of "built-in" but please suggest better ones.


Nothing gets called "build options" because in a meson context, it's hard to come with a name more vague, less "searchable" and generally less useful than "build". 

In the same meson context, the name... "meson" (!) options is slightly less vague but barely. The name "meson BUILTINS" is tolerated in the rare case of potential confusion with some other sort of builtins.

BTW "build" is already used in the build/host/target autoconf triplet which is very relevant to... cross files! Another, huge reason not to overload it with anything else in the same area. It's a bit too late to change autoconf names, like 30 years late. There's also a "build step", a "build directory" and I bet we could find other "build"/nameless things (most of them also much older than meson). 

In case anyone doubts "build" is an empty placeholder, see how "build options" sometimes refers to not-builtins-options, example here: https://mesonbuild.com/Builtin-options.html
> two kinds of options: build options provided by the build files and built-in options that..

In some other places "build options" refers to ALL option kinds! Example: https://mesonbuild.com/Configuring-a-build-directory.html
> Since 0.50.0, it is also possible to get a list of all build options by invoking meson configure


- "**Project**" vs "**Global**" options: names only relevant to subprojects and used only in subproject related documentation, nowhere else. Valid examples:
https://mesonbuild.com/Reference-manual.html#project
> _Global_ options such as buildtype can only be specified in the master project, settings in subprojects are ignored....

https://mesonbuild.com/Subprojects.html#build-options-in-subproject

The names "Project and "Global" are forbidden outside subproject-related topics to avoid overloading and intolerable confusion.

I really, really hope I didn't misunderstand the concepts here. If I have please correct the concepts and do not discuss the names I just suggested as gold standards, that would be pointless. I may have misunderstood the concepts because... they don't have good, standard names yet :-)


https://martinfowler.com/bliki/TwoHardThings.html

dcbaker:
@nirbheek, @jpakkane, do either of you have strong opinions regarding the "one section per project" or "one sections for all projects" debate that @xclaesse and I are having?

dcbaker:
Once we have the big bikesheding debate worked out I'll also rename the sections to match our documentation, `builtin-in options` and `user options`.

marc-h38:
A bikeshed is in the backyard. This is the user interface = the front porch.

jpakkane:
I agree that `meson_options.txt` is not a good name, because the file specifies options for the _project_. In my defense I made that choice back when there were zero other users (or team members) to ask comments for. Oh well, you can't win them all, I guess.

I personally prefer the format where each subproject has its own top level section, for pretty much the same reasons as listed above: it makes it easy to see options affecting each subproject at a glance.

I agree that `[meson options]` is a bit of a misnomer and if we had a word for "options common to all projects including subdir, language standards etc" we should probably start using that consistently. The problem being what should it be called? The only ones I could come up with were `common options` and `shared options` but neither of those feels like a good overall term.

dcbaker:
I've rebased this on master, and changed the `meson options` section to `built-in options`, as that seems a little clearer to me.

xclaesse:
What about not inventing words, and simply use `[options]` for all options that does not have a `subproject:` prefix, and `[subproject options]` for all options that have `subproject:` prefix?

```
[options]
default_library=shared
some_main_project_option=enabled

[foo options]
# Maybe drop `foo:` prefix here, and automatically prefix them all with first word of section title.
# But IMHO it's better to keep the prefix because that's how they are shown in `meson configure`
foo:default_library=static
foo:some_subproject_option=enabled
```

dcbaker:
I don't think we can safely drop the `prefix:`. We guarantee that a project wont have a `:` in the name but we don't do that for any other character. you could name your project `zlib options` if you wanted. I'm not totally opposed to merging the two sections, but I didn't "make up" built-in, that's literally the name of the section in the manual for options that meson provides https://mesonbuild.com/Builtin-options.html.

@jpakkane, do you have a strong preference on one section or two? If you don't have a preference for two I think I'm going to merge them into one section.

xclaesse:
> I didn't "make up" built-in, that's literally the name of the section in the manual for options that meson provides https://mesonbuild.com/Builtin-options.html.

Ah, good point. In the doc built-in options contains core, base, compiler options.

dcbaker:
Right, the intention was that all of the core, base, and compiler options would go in the `built-in options` sections, all of the options defined in `meson_options.txt` would be in the `project options` sections. Does that seem good, or would you still prefer one `options` section?

jpakkane:
Fine by me.

dcbaker:
I'd like to target this for 0.55 or later, I think there's enough in 0.54 as is, and this is probably going to be the source of a lot of bugs.

jpakkane:
This being the case we'd probably need to get the part that allows you to set options in build files from #6595 and merge it for this release.

Ericson2314:
This would benefit I believe benefit from a better separation of parsing files and business logic, especially if we are fine holding off until `0.55`. I got started on that separation #5489, after which I want to improve `Properties` and get rid of Environment.cmd_line_options` by merging in the command line into `Properties` (perhaps using logic for unioning `Properties` from #6815).

I think a good order of things is:

1. #5489
2. Improve Properties and compiler options with a `Language` enum and `PerLanguaeg` a la `MachineChoice` and `PerMachine`.
3. This PR

Also concurrently after step 2:

3. #6815 
4. Get rid of `Environment.cmd_line_options`

And finally after both threads

5. More structure for subprojects, thinking about default options, etc.

dcbaker:
I think @xclaesse and I both have different, unfinished branches to make compiler options per subproject as well, although both of those make the env.comand_line_options even grosser IIRC (I know mine does, maybe the other didn't?)

Ericson2314:
@dcbaker Well, then I better hurry up getting rid of `Environment.cmd_line_options`! :)

dcbaker:
Okay, I've rebased this, and done some small documentation cleanups, but otherwise this is pretty much the same as before. I'd like to move toward landing this sooner rather than later if possible. @Ericson2314, @xclaesse, @jpakkane

TheQwertiest:
How would be the collision between native and cross resolved? For example, what would be result of combining the following two files?
native:
```
[built-in options]
backend=ninja
```
cross
```
[built-in options]
backend=vs2019
```

TheQwertiest:
Regarding `(if doing a cross build the options from the native file will be ignored)`: why is this the case? This would somewhat limit the abilities of this feature.
For example it won't be possible to use a common file for specifying options only:
`meson --native-file=options.ini --native-file=native_toolchain.ini`
`meson --native-file=options.ini --native-file=native_toolchain.ini --cross-file=cross_toolchain.ini`
one would have to move `options.ini` from `--native-file` to `--cross-file` depending on the build type.

nirbheek:
> How would be the collision between native and cross resolved

I think this should be an error.

docs/markdown/Machine-files.md
```diff
@@ -86,12 +86,61 @@ In addition to special data that may be specified in cross files, this
...
 `meson.get_external_property()`
 
-## Properties
+### Project specific options
 
-*New for native files in 0.54.0*
+*New in 0.55.0*
 
-The properties section can contain any variable you like, and is accessed via
-`meson.get_external_property`, or `meson.get_cross_property`.
```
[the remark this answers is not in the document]
dcbaker: Oops, that looks like a bad rebase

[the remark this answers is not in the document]
dcbaker: ah, I see what happened. This section was duplicated. I've split this change out into a separate commit.

docs/markdown/Machine-files.md
```diff
@@ -86,12 +86,61 @@ In addition to special data that may be specified in cross files, this
...
+```ini
+[built-in options]
+c_std = 'c99'
+```
+
+You can set some meson built-in options on a per-subproject basis.
+
+The current list is:
+- default_library
+- werror
```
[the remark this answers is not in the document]
dcbaker: Command line, machine file, build system definitions; in that order. I've updated the docs.

dcbaker:
The problem with it being an error is that some options are allowed to be different for build (native) and host (cross) targets, so we need to be able to set at least some options in the native file.

dcbaker:
@xclaesse, sorry for the late response on this. That wont work because we need the options from both the native and the cross file for options that are allowed to be set on a per machine basis, like pkg-config path. I've added a test for this in the latest version.

mesonbuild/coredata.py
```diff
@@ -747,24 +748,39 @@ def set_default_options(self, default_options, subproject, env):
...
+                if project:
+                    k = '{}:{}'.format(project, k)
+                cmd_line_options[k] = v
+
         # Override project default_options using conf files (cross or native)
         for k, v in env.paths.host:
             if v is not None:
                 cmd_line_options[k] = v
 
+        # For cross builds we need to get the build specifc options
```
Ericson2314, line 778: I am tempted to always do this, but I think I'll wait to simplify things until we can get rid of `cmd_line_options` and `env.cmd_line_options` altogether.

dcbaker: we certainly could do it always, but it wouldn't affect anything.

xclaesse:
> @xclaesse, sorry for the late response on this. That wont work because we need the options from both the native and the cross file for options that are allowed to be set on a per machine basis, like pkg-config path. I've added a test for this in the latest version.

Hm, you would have to indeed prefix them with `build.`, good point.

docs/markdown/Machine-files.md
```diff
@@ -84,14 +84,63 @@ command line will override any options in the native file. For example, passing
...
 
-## Properties
+### Project specific options
 
-*New for native files in 0.54.0*
+*New in 0.55.0*
 
-The properties section can contain any variable you like, and is accessed via
-`meson.get_external_property`, or `meson.get_cross_property`.
+Path options are not allowed, those must be set in the `[paths]` section.
```
xclaesse, line 93: Or rather we could deprecate the `[paths]` section and have them into the built-in options section? I personally don't like imposing to the user arbitrary split of options. Sounds like imposing an implementation detail into our API.

docs/markdown/Machine-files.md
```diff
@@ -84,14 +84,63 @@ command line will override any options in the native file. For example, passing
...
+An incomplete list of options is:
+- pkg_config_path
+- cmake_prefix_path
+
+For setting options in supbprojects use the `<subproject>:project options`
+section instead.
+
+```ini
+[project options]
+build-tests = true
```
xclaesse, line 108: IIRC the cross file parser force to have quotes for all values, do you have a unit test that verify that actually works?

mesonbuild/environment.py
```diff
@@ -563,13 +585,29 @@ def __init__(self, source_dir, build_dir, options):
...
         self.properties = properties.default_missing()
         self.paths = paths.default_missing()
+        self.user_options = user_options
+        self.meson_options = meson_options.default_missing()
+
+        # Ensure that no paths are passed via built-in options:
+        if '' in self.meson_options.host:
+            for each in ['bindir', 'datadir', 'includedir', 'infodir', 'libdir',
+                         'libexecdir', 'localedir', 'localstatedir', 'mandir',
+                         'prefix', 'sbindir', 'sharedstatedir', 'sysconfdir']:
```
xclaesse, line 605: mconf.py already has a dir_option_names list, maybe we should move that list into coredata to maintain it in a single place?

mesonbuild/coredata.py
```diff
@@ -747,24 +748,39 @@ def set_default_options(self, default_options, subproject, env):
...
+        if env.meson_options.host != env.meson_options.build and subproject in env.meson_options.build:
+            for k in builtin_options_per_machine.keys():
```
xclaesse, line 780: There are also compiler options that are per machine, is that handled too?

Ericson2314: Oh oops! I knew something was off...I of all people should have caught that since I made the relevant refactor!

dcbaker:
Okay, I think I've addressed everything except whether we should deprecated the `[paths]` section (which I'd rather do in a separate MR if possible).

@xclaesse, @jpakkane, @Ericson2314, @nirbheek, what else needs to be done for this to land?

docs/markdown/Machine-files.md
```diff
@@ -64,8 +92,8 @@ An incomplete list of internally used programs that can be overridden here is:
 
 ### Paths and Directories
 
-As of 0.50.0 paths and directories such as libdir can be defined in the native
-file in a paths section
+As of 0.50.0 paths and directories such as libdir can be defined in the cross
+and native file in a paths section. These should be strings.
```
Ericson2314, line 96:
```suggestion
As of 0.50.0 paths and directories such as libdir can be defined in native
and cross files in a "paths" section. These should be strings.
```
I like native before cross, like build before host.

dcbaker:
@xclaesse, @jpakkane, @Ericson2314, @nirbheek, I feel like we're ready to land this, anyone object?

docs/markdown/Machine-files.md
```diff
@@ -5,24 +5,52 @@ documentation on the common values used by both, for the specific values of
...
+- arrays
+- booleans
+
+A string is specified single quoted:
+```ini
+[section]
+option = 'value'
+```
+
+An array is enclosed in square brackets, and must consist of strings or booleans
```
nirbheek, line 21: Do we also support implicitly defining an array with a single element using string syntax like we do for the `[binaries]` section? I have a slight preference for not doing that if we do, but if we decide to support that it should be documented.

xclaesse:
I strongly dislike that syntax:
```
[sub:built-in options]
default_library=static
```
And I strongly dislike the way too complex implementation that I still do not really understand, compared to the trivial solution I proposed.

That being said, I ran the unit tests I wrote for https://github.com/mesonbuild/meson/pull/7293 and they passed, and it seems I'm alone in disliking this PR, so... lgtm ;-)

mesonbuild/interpreter.py
```diff
@@ -2754,6 +2754,15 @@ def do_subproject(self, dirname: str, method: str, kwargs):
 
         default_options = mesonlib.stringlistify(kwargs.get('default_options', []))
         default_options = coredata.create_options_dict(default_options)
+
+        # We want to overwrite the keyword arguments with options from the
+        # cross/native file. Take the parent options first, then overrise that
```
TheQwertiest, line 2759: Typo?
```suggestion
        # cross/native file. Take the parent options first, then override that
```

xclaesse:
Would be nice in a future PR to deprecate `[paths]` section since they are built-in options after all, and also deprecate compiler options in `[properties]` section.

xclaesse:
Actually I deprecating compiler options from [properties] is a merge blocker because that native file silently ignore the value from [properties]. We need at least a warning if not abort when both are provided.

```
[built-in options]
c_args='-DBAR'

[properties]
c_args='-DFOO'
```

xclaesse:
Hm, `optimization=1` prints that error:
```
meson.build:1:0: ERROR: Value "1" for combo option "Optimization level" is not one of the choices. Possible choices are: "0", "g", "1", "2", "3", "s".
```

It has to be `optimization='1'` with the quotes (same thing with my PR tbh), wondering if we should make it work or at least print a proper error message.

dcbaker:
> Hm, optimization=1 prints that error:

Oh, yeah, that seems like we should either have a good error message or allow numbers to be unquoted. I think we should allow them unquoted to make `1 != '1'`. In mesa we do have a case where we have a combo that takes numbers as strings, so I think that makes more sense, and I'm going to implement it that way.

dcbaker:
Oh, interesting, currently combo options must be strings....

Hmm, that makes me more inclined to to fix the error message for now, as allowing integers/floats in combo options would be a bigger change.

xclaesse:
I think UserComboOption is wrong because it checks the value is string only in __init__ but that should be done in validate_value(). tbh I would just add `value = str(value).to_lower()` in validate_value(), that would also make it work for example for combos that have 'true' in their choices. But that can (and should) be a separate PR anyway.

dcbaker:
> 

I think UserComboOption is wrong because it checks the value is string only in init but that should be done in validate_value(). tbh I would just add value = str(value).to_lower() in validate_value(), that would also make it work for example for combos that have 'true' in their choices. But that can (and should) be a separate PR anyway.

I agree, for right now I'm updating the error messages to be more helpful, and we can square away the type stuff after that.

dcbaker:
@xclaesse, I've deprecated the `[paths]` section and the `<lang>_args` in the `[properties]`.

docs/markdown/Machine-files.md
```diff
@@ -5,24 +5,59 @@ documentation on the common values used by both, for the specific values of
...
+internally by meson, or by the `find_program` function.
+
+These values must be either strings or an array of strings
 
 Compilers and linkers are defined here using `<lang>` and `<lang>_ld`.
 `<lang>_ld` is special because it is compiler specific. For compilers like
 gcc and clang which are used to invoke the linker this is a value to pass to
 their "choose the linker" argument (-fuse-ld= in this case). For compilers
 like MSVC and Clang-Cl, this is the path to a linker for meson to invoke,
-such as `link.exe` or `lld-link.exe`. Support for ls is *new in 0.53.0*
```
TheQwertiest, line 25: So support for `ls` was deprecated? Darn! :P

docs/markdown/Machine-files.md
```diff
@@ -84,21 +121,76 @@ command line will override any options in the native file. For example, passing
 
 In addition to special data that may be specified in cross files, this
 section may contain random key value pairs accessed using the
-`meson.get_external_property()`
+`meson.get_external_property`, or `meson.get_cross_property`.
```
TheQwertiest, line 124:
```suggestion
`meson.get_external_property()` or `meson.get_cross_property()`.
```

But, IMO, we shouldn't mention `meson.get_cross_property()` here at all, since we have `get_external_property()` which is superset of `get_cross_property()`

cross/c2000.txt
```diff
@@ -12,7 +12,7 @@ cpu_family = 'c2000'
 cpu = 'c28x'
 endian = 'little'
 
-[properties]
+[built-in options]
 needs_exe_wrapper = true
```
xclaesse, line 16: needs_exe_wrapper is still a property.

docs/markdown/snippets/project_options_in_machine_files.md
```diff
@@ -0,0 +1,51 @@
...
+
+Additionally meson level options can be set in the same way, using the
+`[built-in options]` section.
+
+```ini
+[built-in options]
+c_std = 'c99'
+```
+
+Subprojects works too, although only `default_library` can currently be set:
```
TheQwertiest, line 47:
```suggestion
Subprojects works too, although only `default_library` and `werror` can currently be set:
```
Because https://github.com/mesonbuild/meson/pull/6795

docs/markdown/Machine-files.md
```diff
@@ -5,24 +5,59 @@ documentation on the common values used by both, for the specific values of
 one or the other see the [cross compilation](Cross-compilation.md) and [native
 environments](Native-environments.md).
 
+## Data Types
+
+There are three basic data types in a machine file:
+- strings
+- arrays
+- booleans
+- integers
```
xclaesse, line 14: There are three hard problems in programming:
- naming things
- off by one error

docs/markdown/Machine-files.md
```diff
@@ -5,24 +5,59 @@ documentation on the common values used by both, for the specific values of
...
+[section]
+option = ['value']
+```
+
+A boolean must be either `true` or `false`, and unquoted.
+```ini
+option = false
+```
+
+An integer must be either an unquoted numeric constant;
```
TheQwertiest, line 34:
```suggestion
An integer must be an unquoted numeric constant.
```

docs/markdown/Machine-files.md
xclaesse: It's not done in the machine file, but probably a stringlistify() somewhere when it gets the value of binaries. Array options works like that too, `c_args='-DFOO'` works.

docs/markdown/Machine-files.md
```diff
@@ -84,21 +121,76 @@ command line will override any options in the native file. For example, passing
```
TheQwertiest, line 121: Drive-by (for line 120):
```diff
- *New in native files in 0.54.0*, always in cross files.
+ *available in native files since 0.54.0*
+ *always available in cross files*
```

docs/markdown/Machine-files.md
xclaesse, line 34: 'either' here implies there is another syntax possible, but there is none, right?

docs/markdown/snippets/project_options_in_machine_files.md
dcbaker: I swear I added `werror` already, I must have accidentally rebased it out.

docs/markdown/Machine-files.md
```diff
@@ -40,8 +75,8 @@ llvm-config = '/usr/lib/llvm8/bin/llvm-config'
 Cross example:
 
 ```ini
-c = '/usr/bin/i586-mingw32msvc-gcc'
-cpp = '/usr/bin/i586-mingw32msvc-g++'
+c = ['ccache', '/usr/bin/i586-mingw32msvc-gcc']
+cpp = ['ccache', '/usr/bin/i586-mingw32msvc-g++']
```
xclaesse, line 79: Not sure we want to change this example, a simple string works and IIRC there is an open issue somewhere to prepend ccache automatically when available.

docs/markdown/Machine-files.md
TheQwertiest: This line needs to be edited, since it contradicts the line above:
```
### Paths and Directories
*Deprecated in 0.55.0* use the built-in section instead.
```

xclaesse: Now that you did deprecate [paths] that line should be removed.

docs/markdown/Machine-files.md
```diff
@@ -84,21 +121,76 @@ command line will override any options in the native file. For example, passing
...
+
+Being able to set project specific options in a cross or native file can be
+done using the `[project options]` section of the specific file (if doing a
+cross build the options from the native file will be ignored)
+
+An incomplete list of options is:
+- pkg_config_path
+- cmake_prefix_path
+
+For setting options in supbprojects use the `<subproject>:project options`
```
TheQwertiest, line 144:
```suggestion
For setting options in subprojects use the `<subproject>:project options`
```

docs/markdown/Machine-files.md
```diff
@@ -84,21 +121,76 @@ command line will override any options in the native file. For example, passing
...
+
+Path options are not allowed, those must be set in the `[paths]` section.
+
+Being able to set project specific options in a cross or native file can be
+done using the `[project options]` section of the specific file (if doing a
+cross build the options from the native file will be ignored)
+
+An incomplete list of options is:
+- pkg_config_path
+- cmake_prefix_path
```
xclaesse, line 142: Those are not project options, they are built-in, right?

docs/markdown/Machine-files.md
```diff
@@ -84,21 +121,76 @@ command line will override any options in the native file. For example, passing
...
+```ini
+[built-in options]
+c_std = 'c99'
+```
+
+You can set some meson built-in options on a per-subproject basis, such as
+`default_library` and `werror`. The order of precedence is:
+1) Command line
+2) Machine file
+3) build system definitions
```
TheQwertiest, line 168:
```suggestion
3) Build system definitions
```

docs/markdown/Machine-files.md
xclaesse, line 144: nitpicking: I would add `[]` around it.

docs/markdown/snippets/project_options_in_machine_files.md
```diff
@@ -0,0 +1,51 @@
+## Project and Meson options can be set in native or cross files
```
xclaesse, line 1: Meson -> built-in here, since that's the naming we picked.

mesonbuild/coredata.py
```diff
@@ -160,10 +160,16 @@ def __init__(self, description, choices: T.List[str], value, yielding=None):
...
+                _type = 'number'
+            else:
+                _type = 'string'
             optionsstring = ', '.join(['"%s"' % (item,) for item in self.choices])
-            raise MesonException('Value "{}" for combo option "{}" is not one of the choices.'
-                                 ' Possible choices are: {}.'.format(
-                                     value, self.description, optionsstring))
+            raise MesonException('Value "{}" (of type "{}") for combo option "{}" is not one of the choices.'
+                                 ' Possible choices are (as string): {}.'.format(
+                                     value, _type, self.description, optionsstring))
```
xclaesse, line 172: Any type other than str is forbidden atm. I would just move the check from `__init__()` into `validate_value()`:

```
            if not isinstance(value, str):
                raise MesonException('Combo choice elements must be strings.')
```

mesonbuild/coredata.py
```diff
@@ -749,26 +756,52 @@ def set_default_options(self, default_options, subproject, env):
...
                 if (k not in builtin_options or builtin_options[k].yielding) \
                         and optinterpreter.is_invalid_name(k, log=False):
                     continue
-                k = subproject + ':' + k
+                k = '{}:{}'.format(subproject, k)
             cmd_line_options[k] = v
 
-        # Override project default_options using conf files (cross or native)
-        for k, v in env.paths.host:
-            if v is not None:
```
xclaesse, line 767: [paths] is deprecated but should still work, are you sure that code can be deleted?

docs/markdown/Machine-files.md
```diff
@@ -84,21 +121,76 @@ command line will override any options in the native file. For example, passing
...
+option from the parent if the parent has a setting but the subproject
+doesn't, even when here is a default set meson language.
+
+
+```ini
+[built-in options]
+default_library = 'static'
+```
+
+will make subprojects use default_library as static.
```
TheQwertiest, line 186: Edited for clarity:

``````suggestion
Note that the order of precedence is the following:
1) Command line
2) Machine file
3) Build system definitions

You can set some meson built-in options on a per-subproject basis, such as
`default_library` and `werror`:

```ini
[zlib:built-in options]
default_library = 'static'
werror = false
```

Per-subproject options use the same order of precedence as the root project options.
Subproject will also inherit options from the parent project if it was not specified
specifically for the subproject. For example the following:

```ini
[built-in options]
default_library = 'static'
```

will make subprojects use `default_library` as static.
``````

File: mesonbuild/coredata.py (hunk 24 above)
Comment (xclaesse) [reply mesonbuild/coredata.py:767]: nevermind, you load them into user_options, good.

File: docs/markdown/snippets/project_options_in_machine_files.md (hunk 14 above)
Comment (TheQwertiest) [reply docs/markdown/snippets/project_options_in_machine_files.md:47]: It *was* added, but not here (rather in the other part of the doc) =)

File: docs/markdown/Machine-files.md (hunk 17 above)
Comment (TheQwertiest) [inline docs/markdown/Machine-files.md:121]: Drive-by:

```diff
- *New in native files in 0.54.0*, always in cross files.
+ *available in native files since 0.54.0*
+ *always available in cross files*
```

docs/markdown/snippets/project_options_in_machine_files.md
```diff
@@ -0,0 +1,51 @@
...
+
+A new set of sections has been added to the cross and native files, `[project
+options]` and `[<subproject_name>:project options]`, where `subproject_name`
+is the name of a subproject. Any options that are allowed in the project can
+be set from this section. They have the lowest precedent, and will be
+overwritten by command line arguments
```
TheQwertiest, line 7:
```suggestion
overwritten by command line arguments.
```

docs/markdown/snippets/project_options_in_machine_files.md
```diff
@@ -0,0 +1,51 @@
...
+Will result in the option foo having the value `other val`,
+
+```console
+meson build --native-file my.ini -Dfoo='different val'
+```
+
+Will result in the option foo having the value `different val`,
+
+
+for subprojects this looks like:
```
TheQwertiest, line 32:
```suggestion
Subproject options are assigned like this:
```

docs/markdown/snippets/project_options_in_machine_files.md
TheQwertiest: Removed extra `s` as well.
```suggestion
Subprojects work too, although only `default_library` and `werror` can currently be set:
```

mesonbuild/environment.py
```diff
@@ -553,13 +554,31 @@ def __init__(self, source_dir, build_dir, options):
...
+                    store[project] = config.get(section, {})
+
         if self.coredata.config_files is not None:
             config = MesonConfigFile.from_config_parser(
                 coredata.load_configs(self.coredata.config_files))
             binaries.build = BinaryTable(config.get('binaries', {}))
-            paths.build = Directories(**config.get('paths', {}))
             properties.build = Properties(config.get('properties', {}))
 
+            # Don't run this if there are cross file, we don't want to use the
```
TheQwertiest, line 572:
```suggestion
            # Don't run this if there is a cross file, we don't want to use the
```

mesonbuild/interpreter.py
```diff
@@ -2754,6 +2754,15 @@ def do_subproject(self, dirname: str, method: str, kwargs):
...
         default_options = mesonlib.stringlistify(kwargs.get('default_options', []))
         default_options = coredata.create_options_dict(default_options)
+
+        # We want to overwrite the keyword arguments with options from the
+        # cross/native file. Take the parent options first, then override that
+        # with the subproject specific options
+        for n in ['', dirname]:
+            default_options.update(self.environment.meson_options.host.get(n, {}))
+            default_options.update({'build.{}'.format(k): v for k, v in self.environment.meson_options.build.get(n, {}).items()
+                                    if k in coredata.builtin_options_per_machine})
```
xclaesse, line 2764: You prepend `build.` here and also in set_default_options(). Here you set them in default_options, but then you also set them in set_default_options()... I'm lost.

The code in set_default_options() and do_subproject() is already waaay to complicated, I would **STRONGLY** recommend refactoring your code to update self.cmd_line_options in environment.py as I did in myhttps://github.com/mesonbuild/meson/pull/7293. That simplify a lot the implementation of this PR and makes it obvious that options in machine files are handled  the same way as if passed on the command line.

TheQwertiest:
@dcbaker , sorry for all the nit-picks :P

docs/markdown/Machine-files.md
dcbaker: that's just a copy and paste error :)

docs/markdown/Machine-files.md
dcbaker: The point was that a list is also accepted, as it came up recently that you can pass these as lists not just as strings, and using ccache was a reasonable example. If you've got a different example using a list I'll change it.

docs/markdown/Machine-files.md
dcbaker: That looks like something went wrong in a rebase, it's in the wrong section and the paragraph before it about separate host vs build options are missing.

dcbaker:
Okay, I think I resolved most or all of the docs and comments nits. It's getting late here, so I'll look at the rest tomorrow.

mesonbuild/interpreter.py
dcbaker: I think when @Ericson2314 and I talked about this (it was a while ago on IRC) we were talking about removing env.cmd_line_options all together and using the individual options instead.

docs/markdown/Machine-files.md
TheQwertiest: Still not fixed tho :P

docs/markdown/Machine-files.md
```diff
@@ -5,24 +5,59 @@ documentation on the common values used by both, for the specific values of
...
 - paths
 - properties
+- project options
+- built-in options
 
 ### Binaries
 
 The binaries section contains a list of binaries. These can be used
-internally by meson, or by the `find_program` function:
+internally by meson, or by the `find_program` function.
```
TheQwertiest, line 51:
```suggestion
internally by meson or by the `find_program` function.
```

docs/markdown/Machine-files.md
```diff
@@ -84,21 +121,79 @@ command line will override any options in the native file. For example, passing
 
 In addition to special data that may be specified in cross files, this
 section may contain random key value pairs accessed using the
-`meson.get_external_property()`
+`meson.get_external_property()`, or `meson.get_cross_property()`.
```
TheQwertiest, line 124:
```suggestion
`meson.get_external_property()` or `meson.get_cross_property()`.
```
And my comment about unneeded `get_cross_property` still stand, so the following would be even better:
```suggestion
`meson.get_external_property()`.
```

mesonbuild/interpreter.py
xclaesse: I don't know what's wrong with `env.cmd_line_options` and I would like to discuss it if there are issues raised since I'm the one who wrote that code ;-) But in any case, that's not a reason to make the code more complicated than it could be in the meantime.

btw, I'm not saying your code is wrong. It does the right thing in all the tests I've done, I just think it could do the same thing in a simpler and more maintainable way. From past experience I know that set_default_options() is a can of worms with many subtle edge cases, better avoid messing with it unless it's for rewriting that mess :D

dcbaker:
@xclaesse, @Ericson2314: I went ahead and started doing the refactor that xclaesse suggested, and ended up removing env.cmd_line_options, because it was simpler to get rid of it. What I came up with was a set of structures mirroring the ones in coredata, but that store the raw (not a UserOption) value passed in. My approach folds the command line and the machine files and Env creation time, so don't need to keep a copy of the command line arguments in env to keep mutating. In general it's a nice cleanup I think, although if we were to ensure all options were instantiated at CoreData initialization time we could make this even simpler, though there's tradeoffs to that so I haven't gone that far.

xclaesse:
Big refactor like that should go into separate PR, it's huge to review now... tbh if you stick to the implementation I suggested it would already be merged by now. Not saying it's wrong, but it will take me a big effort to try to understand that diff :(

mesonbuild/coredata.py
```diff
@@ -741,87 +747,58 @@ def set_options(self, options, *, subproject='', warn_unknown=True):
...
+        lang_prefixes = tuple('{}_'.format(l) for l in all_languages)
+        # split arguments that can be set now, and those that cannot so they
+        # can be set later, when they've been initialized.
+        for k, v in default_options.items():
+            if k.startswith(lang_prefixes):
+                lang, key = k.split('_', 1)
+                for machine in MachineChoice:
+                    if key not in env.compiler_options[machine].get(lang, {}):
+                         if lang not in env.compiler_options[machine]:
+                             env.compiler_options[machine][lang] = {}
```
Ericson2314, line 769: Aha! These use `defaultdict` so you don't need to do that, the dict will be recreated on demand!

mesonbuild/coredata.py
```diff
@@ -741,87 +747,58 @@ def set_options(self, options, *, subproject='', warn_unknown=True):
...
+                          env.meson_options.build.get(subproject, {}).items()):
+            if k in builtin_options_per_machine:
+                options[make_key('build.{}'.format(k))] = v
+
+        options.update({make_key(k): v for k, v in env.user_options.get(subproject, {}).items()})
+
+        # Some options (namely the compiler options) are not preasant in
+        # coredata until the compiler is fully initialized. As such, we need to
+        # put those options into env.meson_options, only if they're not already
+        # in there, as the machine files and command line have precendence.
```
Ericson2314, line 791: Maybe this easier left as a further refactor, but if not: Wouldn't it be easier to just *always* put all config items in `env.meson_options` and then initialize the actual options using `env.meson_options` alone (no CLI, no default options, etc. to worry about) later?

mesonbuild/coredata.py
```diff
@@ -741,87 +747,58 @@ def set_options(self, options, *, subproject='', warn_unknown=True):
...
+        for k, v in default_options.items():
+            if k in builtin_options and not builtin_options[k].yielding:
+                continue
+            for machine in MachineChoice:
+                if machine is MachineChoice.BUILD and not self.is_cross_build():
                     continue
-            options[k] = v
+                if subproject not in env.meson_options[machine]:
+                    env.meson_options[machine][subproject] = {}
```
Ericson2314, line 799: I'd use a `defaultdict` for this too (see previous comment) so you don't need to do this.

mesonbuild/coredata.py
```diff
@@ -1094,23 +1066,25 @@ def add_to_argparse(self, name: str, parser: argparse.ArgumentParser, prefix: st
...
-    # Directories
-    ('prefix',     BuiltinOption(UserStringOption, 'Installation prefix', default_prefix())),
-    ('bindir',     BuiltinOption(UserStringOption, 'Executable directory', 'bin')),
-    ('datadir',    BuiltinOption(UserStringOption, 'Data file directory', 'share')),
-    ('includedir', BuiltinOption(UserStringOption, 'Header file directory', 'include')),
-    ('infodir',    BuiltinOption(UserStringOption, 'Info page directory', 'share/info')),
-    ('libdir',     BuiltinOption(UserStringOption, 'Library directory', default_libdir())),
-    ('libexecdir', BuiltinOption(UserStringOption, 'Library executable directory', default_libexecdir())),
-    ('localedir',  BuiltinOption(UserStringOption, 'Locale data directory', 'share/locale')),
+BUILTIN_DIR_OPTIONS = OrderedDict([
```
Ericson2314, line 1071: Ah thanks, I kept on getting confused with all the similarly named fields and constants :)

mesonbuild/environment.py
```diff
@@ -560,15 +563,61 @@ def __init__(self, source_dir, build_dir, options):
...
             config = MesonConfigFile.from_config_parser(
                 coredata.load_configs(self.coredata.config_files))
             binaries.build = BinaryTable(config.get('binaries', {}))
-            paths.build = Directories(**config.get('paths', {}))
             properties.build = Properties(config.get('properties', {}))
 
+            # Don't run this if there are any cross files, we don't want to use
+            # the native values if we're doing a cross build
+            if not self.coredata.cross_files:
+                load_options('project options', user_options)
```
Ericson2314, line 611: What exactly is this? Is it something we could do with that defaulting logic instead?

Ericson2314:
@xclaesse Well yes it's nice to keep features and refactors separate in general....but I'm even more excited about the refactor than I am about the feature.

I hoped there would have been more deleted code because of it, but I think that can still be gotten in later refactors. (For example, we can read *all* the env vars at once in `environment`, and then ban `os.environ` from the rest of the code base, which will not only clean things up but give us huge correctness/purity benefits).

xclaesse:
I haven't read the code, but at first glance it adds more code than it removes, I'm really not convinced at it's the right move. Separating options into more structures sounds like a step backward, one thing I always wanted to do is merging all those dicts into a single one, it makes a lot of code uselessly more complicated.

But I'm open to discussions to the pros&cons, but IMHO it's not at all the right place here. I really want to get options into cross file, I currently using this into projects, but this PR is not impossible to review.

Ericson2314:
Well perhaps @jpakkane needs to way in here, but if there are still doubts on the refactor, let's just make a PR without the feature and try to get those deleted lines up----that is the true test of this approach vs @xclaesse's PR, and I'm happy to put my money where my mouth is, writing additional commits to the refactor to delete more stuff.

xclaesse:
Sure I'm not against the refactor, there is a lot of code that could be better. If that could be splitted that would help a lot to see the advantages of the new design.

mesonbuild/coredata.py
dcbaker: won't that cause problems for the `default_missing()` method?

mesonbuild/coredata.py
dcbaker: The refactor I'd like to do is to ensure that all options are initialized before we get here, so that we don't need to do any of this, we can just put them straight into coredata and be done with it.

mesonbuild/environment.py
dcbaker: This allows you to put project options (ie,` -Duse_systemd`) in your machine file. If we're doing a "native" build, we load those values out of the native file, if we're doing a cross build, load them from the cross file. We don't store them on a per-subproject basis because we only consider the file for the type of build we're doing.

dcbaker:
@xclaesse, I'm not sure what the refactor you want is then? I thought you wanted all of the logic to be in Environment, with none in do_subproject or coredata.set_default_options.

Ericson2314:
@dcbaker I think @xclaesse was separately on one hand being skeptical about the refactor, and on the other wishing it was separate than the functionality change. I guess from the last comment the second is the primary concern?

jpakkane:
One thing that jumped at me is that all of the cross files seem to use the new builtin option syntax for things like `c_args`. At least one of the common ones should stick with the old definition to ensure we don't break backwards compatibility. We must keep the old cross files working for at least a release or two.

dcbaker:
I've added a test for both using the legacy paths section and the legacy `lang_args` in `[properties]` (which, surprise, were both broken). I've removed the warning test that was failing, the warnings are being generated before the `project()` call starts, and the test harness doesn't log that for some reason. I've also done the `defaultdict` switch as requested, and fixed the LGTM warning.

dcbaker:
The mac failure looks spurious to me, but msys and suse have exactly the same issue which I can't reproduce locally. Anyone have any ideas?

xclaesse:
I would like to put this on hold until https://github.com/mesonbuild/meson/pull/7521 gets merged, otherwise it will badly conflict. I did not read the latest version of this PR yet, but I would strongly recommend rebasing on top of https://github.com/mesonbuild/meson/pull/7521 and following the implementation idea I gave earlier. (Sorry, I know I'm grumpy)

dcbaker:
Okay, my turn to be grumpy, really fucking grumpy. I've been trying to merge this for almost a year, you asking me to rebase again on something you opened yesterday... let me just say I don't have nice words for you so I'm going to stop there.

dcbaker:
I know what the problem is, it's the code for handling the environment variables. I just need to decide how to fix it.

Ericson2314:
@dcbaker care to explain? I changed a lot of that code so I'm curious what the problem is---I agree it could be better.

Ericson2314:
Ah OK that makes sense. Yeah we really need a better solution than all this `default_missing` and `miss_defaulting` crap I wrote long ago.

dcbaker:
I don't know what the right solution is really, but since environment variables are per-machine anyway (due to the _FOR_BUILD) variants, it was a bug in my implementation to set them before calling default_missing()

dcbaker:
@jpakkane, @mensinda: I finally have all of the bugs worked out of this. Do either of you have any more objections before I merge?

dcbaker:
@mensinda, done.

dcbaker:
Sigh, more regressions.

dcbaker:
Okay, I think I have them all resolved now.

nirbheek:
Test failure is unrelated.

dcbaker:
I'm going to merge then. \o/

xclaesse:
@dcbaker I'm sorry I've been too arch last week. I know you worked a lot on this PR but to be honest I've said many times I don't like neither the syntax nor the implementation. I have suggested multiple times a much simpler implementation that could have been merged months ago. I personally need this feature and like it a lot, but I'm not the one to give the +1 on it given that I'm not happy with how it's done.

That being said, cool that it got merged, I'll read that code while rebasing my branch.

xclaesse:
You got PKG_CONFIG_PATH wrong, the env override value from native file, and it warn uselessly on reconfigure if you have the value in native file. The warning should be removed, it never worked (was dead code before this PR). It requires more work to actually do it properly. I'm preparing a fix.

mensinda:
Wait, do I understand this correctly: Env variables override stuff in native/cross files? If yes, why?

xclaesse:
It's a bug in this PR, I'll fix it.

mensinda:
OK, thanks for clarifying. I thought this was a general thing in meson for a second :)

xclaesse:
Actually, a comment suggests it's intentional: https://github.com/mesonbuild/meson/pull/6597/files#diff-638d32e2abc720f45ee2279d5befe1e5R690

@dcbaker what's the rational to make that change? Seems weird that -Dpkg_config_path=foo overrides PKG_CONFIG_PATH but not setting pkg_config_path in machine file.

xclaesse:
I don't think `test_builtin_options_subprojects_inherits_parent_override()` is correct, and I think it's a regression.

Before this PR if you run `meson builddir -Ddefault_library=both` in `test cases/common/230 persubproject options` it pass, but with this PR it raises `ERROR: Assert failed: Parent should override default_library`. At least it's consistent with the behaviour when setting default_library=both in the machine file, but I'm not sure it's the right behaviour.

Ericson2314:
I did also think the precedence we CLI > config files > env vars.

xclaesse:
`test cases/common/230 persubproject options` also now print that warning:

`|WARNING: In subproject sub1: Unknown options: "sub1:build.pkg_config_path, sub1:pkg_config_path, sub1:prefix"`

xclaesse:
It also print that warning when the main project sets options in project():

|WARNING: In subproject sub1: Unknown options: "sub1:buildtype, sub1:cpp_eh, sub1:cpp_std, sub1:prefix, sub1:sub1:test_option"

xclaesse:
@nirbheek @Ericson2314 Since you gave approval to this PR, did you make a in depth review of it an approve this change? Do you agree with @dcbaker that his approach is better and more maintainable than https://github.com/mesonbuild/meson/pull/7293? Since I get no support from anyone when pushing for a simpler solution, and if it is a well thought decision, then I'll have to accept the decision of the majority.

PR #6598, opened 2020-02-07
run_project_tests: Fix it to actually work

dcbaker:
You can't sort TestDefs at the moment. Also, don't do an inplace sort while we're at it.


dcbaker:
@mensinda, I think I addressed everything.

dcbaker:
I'll merge as soon as the CI turns green. Thanks for looking at it!

PR #6599, opened 2020-02-07
Stabilize unstable-kconfig, rename it to keyval

bonzini:
Discussions in #6524 have shown that there are various possible uses of the kconfig module and even disagreements in the exact file format between Python-based kconfiglib and the tools in Linux.  Instead of trying to reconcile them, just rename the module to something less suggestive and leave any policy to meson.build files.

In the future it may be possible to add some kind of parsing through keyword arguments such as bool_true, quoted_strings, etc. and possibly creation of key-value lists too.  For now, configuration_data objects provide an easy way to access quoted strings.  Note that Kconfig stores
false as "absent" so it was already necessary to write `x.has_key('abc')` rather than the more compact `x['abc']`.  Therefore, having to use configuration_data does not make things much more verbose.


marc-h38:
While I didn't get a chance to try this, I really the idea of evolving this into something more generic and simpler than a specialized Kconfig module. Pretty sure "keyval" is how GNU Make treats .config files BTW.

I had a look at the current test cases and I found none that uses strings (= what prompted #6524). Adding a few strings would make the test cases a bit more "realistic". Both quoted and unquoted strings. There's also no hex number and all the integers are only one character long.

Note I'm not suggesting adding `get_unquoted()` or any other type-specific test code, that would be for another PR if any. I'm merely concerned about the test cases being too simplistic.

In fact and to help make sure this module doesn't ever become kconfig-specialized again, I think it would be good to change or add non-Kconfig syntax in tests: lowercase keys not starting with CONFIG_, 'n' boolean value, etc.

bonzini:
@marc-h38 can you please write the test cases and place them in a branch I can pull and incorporate into mine? Thanks!

jpakkane:
Could we keep this named `unstable-keyval` for a release or two? It has not had that much use yet.

xclaesse:
IMHO we should not use `unstable-` prefix in module names, because when you'll eventually remove the prefix that won't be backward compatible change. Unless we add an alias mechanism to make both `import('unstable-keyval')` and `import('keyval')` work when it will be stabilized.

That being said, keeping the unstable notice in the doc for a few release would make sense indeed.

bonzini:
> IMHO we should not use unstable- prefix in module names, because when you'll eventually remove the prefix that won't be backward compatible change. Unless we add an alias mechanism to make both import('unstable-keyval') and import('keyval') work when it will be stabilized

Right, for this reason we would like to have the module stabilized before we switch to Meson. Another possibility perhaps could be an `allow_unstable` keyword argument. That would only leave the ugly warning.

I think there aren't many other ways to implement this module if we agree not to do any parsing whatsoever, which is why I proposed to stabilize it immediately. Note that the change is not in semantics or implementation; the original code was fine but the `kconfig` name suggested that it was trying to do more than it actually did.

xclaesse:
Even if we add some parsing, that will be done with extra kwarg passed to load() method to select which parsing format, etc.

bonzini:
Indeed, from the backwards-compatibility POV it should be safe.

nirbheek:
I thought the idea was always to have `import('unstable-foo')` keep working even after module `foo` would become stable. Otherwise what's the point in having it available at all? People aren't going to use something that will break their old build files regardless of whether they put in the effort to get it stable or not.

jpakkane:
Making `import('unstable-foo')` fall back to `import('foo')` if it is not found is a fairly straightforward change in `func_import`.

mesonbuild/modules/keyval.py
```diff
@@ -21,9 +21,9 @@
 
 import os
 
-class KconfigModule(ExtensionModule):
+class KeyvalModule(ExtensionModule):
 
-    @FeatureNew('Kconfig Module', '0.51.0')
+    @FeatureNew('Keyval Module', '0.55.0')
```
xclaesse: Let's be optimistic, next release is 0.54.0

xclaesse:
Does someone volunteer to become maintainer of this module? Please add yourself in CODEOWNERS.

xclaesse:
> Making `import('unstable-foo')` fall back to `import('foo')` if it is not found is a fairly straightforward change in `func_import`.

I'm not personally a big fan of "unstable" in API, because in practice nobody cares. If someone needs the feature they will still be using it and we should still do as much as we can to maintain stability. The side effect of having `import('unstable_keyval')` in our API today is we'll still see that in projects years after it has been marked stable.

That being said, we can (and should) indeed add that fallback, we have other 'unstable' modules already.

bonzini:
> Does someone volunteer to become maintainer of this module? Please add yourself in CODEOWNERS.

I do.

> If someone needs the feature they will still be using it and we should still do as much as we can to maintain stability.

Yes, I agree. It's painful for downstream projects and should be avoided unless something is really, really unstable.

marc-h38:
> I'm not personally a big fan of "unstable" in API, because in practice nobody cares.

Has there been a survey? I do.

I wouldn't spend so much time around here otherwise.

bonzini:
> Has there been a survey? I do.

I think what he means is that people needing unstable APIs/modules will use them nevertheless, and won't care that they are unstable. I am taking a wil guess that you would participate anyway even if the module was named `kconfig`. :)

marc-h38:
> I think what he means is that people needing unstable APIs/modules will use them nevertheless, and won't care that they are unstable. 

I'm using `unstable-kconfig` for a brand new, experimental project. I would certainly not for a "production" project. In fact I would certainly not use meson at all for a production project that requires kconfig.

You can't survey the users you don't even have.

bonzini:
Any decision here? Should I redo this with unstable-keyval? (QEMU would still not switch to Meson until the warning is there).

jpakkane:
Let's go with `unstable-keyval` so we can get this in the next release. We can do the `unstable-foo` -> `foo` fallback (at the latest) when the keyval module goes stable.

Maybe the release note snippet could say "we expect this module to become stable once it has some usage experience, specifically in the next or the following release" or something like that.

bonzini:
@jpakkane CI failures seem transient, they happened even before the keyval tests ran.

bonzini:
This has missed 0.54.0. If I rebase for 0.55.0 should I rename the method to `keyval` or keep `unstable-keyval`?

jpakkane:
Let's do unstable now, we can change it before the actual release if needed.

elmarco:
@jpakkane ping?

docs/markdown/Keyval-module.md
```diff
@@ -1,15 +1,15 @@
 ---
-short-description: Unstable kconfig module
+short-description: keyval module
```
elmarco, line 2: The doc should be updated too, it's still unstable..

docs/markdown/Keyval-module.md
```diff
@@ -23,20 +23,23 @@ chosen the configuration options), output a ".config" file.
 The module may be imported as follows:
 
 ``` meson
-kconfig = import('unstable-kconfig')
+keyval = import('keyval')
```
elmarco, line 26: unstable-keyval

elmarco:
@jpakkane good to go?

PR #6600, opened 2020-02-07
minstall: Make --only-changed less verbose

nirbheek:
When `--only-changed` is passed, we only want to know about files that were newly-installed. Everything else is noise. The full list can always be found in `install-log.txt` anyway. Sample output:

```
ninja: Entering directory `.'
ninja: no work to do.
Preserved 667 unchanged files, see meson-logs\install-log.txt for the full list
```

### DIFF
--- a/mesonbuild/minstall.py
+++ b/mesonbuild/minstall.py
@@ -214,6 +214,7 @@ class Installer:
         self.did_install_something = False
         self.options = options
         self.lf = lf
+        self.preserved_file_count = 0
 
     def should_preserve_existing_file(self, from_file, to_file):
         if not self.options.only_changed:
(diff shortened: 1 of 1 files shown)


jpakkane:
```
  File "run_project_tests.py", line 755, in <listcomp>
    gathered_tests = [(name, gather_tests(Path('test cases', subdir)), skip) for name, subdir, skip in all_tests]
  File "run_project_tests.py", line 542, in gather_tests
    all_tests.sort()
TypeError: '<' not supported between instances of 'TestDef' and 'TestDef'
```

????????

jpakkane:
Those ifs bend my mind into a pretzel. They should not work, but yet they do.

Anyhow, for usability's sake maybe the code should count the number of skipped files, and print a summary like `42 files skipped because they were already installed` if the number is bigger than zero?

nirbheek:
> Anyhow, for usability's sake maybe the code should count the number of skipped files, and print a summary like `42 files skipped because they were already installed` if the number is bigger than zero?

I like that idea! I was thinking about something like this myself.

nirbheek:
Updated according to reviews.

xclaesse:
Perfect

PR #6602, opened 2020-02-08
boost: System dependency rewrite

mensinda:
This PR significantly improves the handling of the boost library tags and also ensures that the found libraries are always compatible (have the same ABI tag). The current setup can also be extended to filter for additional features (static linking with the runtime, etc.).

Additionally, BOOST_ROOT is better supported (it is now guaranteed that all found files belong to a single root directory).

Finally, `boost.py` is now fully annotated (and checked with mypy).

### DIFF
--- a/.github/workflows/lint_mypy.yml
+++ b/.github/workflows/lint_mypy.yml
@@ -30,4 +30,4 @@ jobs:
       with:
         python-version: '3.x'
     - run: python -m pip install mypy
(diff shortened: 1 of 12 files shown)


mensinda:
The CI is finally green, so this PR is ready for review.

@jpakkane Could you please add `libboost-regex-dev` to your eoan image? This package is required for the static boost_regex library.

jpakkane:
Updated, please add the same to `ci/ciimage/eoan/Dockerfile`.

jon-turney:
Updating #6620, one thing I noticed was that the output for `test cases/failing/68 wrong boost module` seems to have changed (I'm guessing due to this) and no longer identifies the individual boost modules which couldn't be found (`abc` in that case).

PR #2617, opened 2017-11-12
Cargo integration in Meson via an "unstable-cargo" module (try 2)

nirbheek:
This was made primarily so we can build librsvg with Meson, which now works: https://git.gnome.org/browse/librsvg/log/?h=wip/meson, and it actually required fewer hacks than I expected it to. 

The integration can and should definitely be better, but I think the best way to get there is to have something that works right now and slowly improve Cargo till we have everything we need. Cargo is not going to magically grow API for giving greater control to build systems; we have to work towards it.

I have been talking to @alexcrichton about the things that Meson would need to integrate better with Cargo, and he has been very receptive and encouraging of our efforts, and in the need for adding features to Cargo that allow Meson to control what it does a bit better.

Here's what you can do right now:

'Extract' build outputs from `Cargo.toml` files in your source tree with the `unstable-cargo` module and mostly use them like you would any other target. For example:

```meson
cargo = import('unstable-cargo')
# Build the staticlib
librsvg_internals = cargo.static_library('rsvg_internals', toml : 'Cargo.toml')

librsvg = library('rsvg-' + rsvg_api_major, librsvg_sources,
                  dependencies : [librsvg_deps, librsvg_internals])
```

You can also specify the list of sources if you wish:

```meson
cargo.executable('mesonbin',
                 toml : 'Cargo.toml',
                 # This is optional, since ninja keeps track of dependencies on source files.
                 sources : ['src/main.rs'],
                 # Yes, installation works
                 install : true)
```

The sources list is optional because cargo will output a nice `depfile` for us and ninja can then keep track of when to rebuild the target.

Meson TODO:
* Cross-compilation
* Cargo tests and benchmarks
* cdylibs or rust-specific library types
* `ninja dist` does not support vendoring crate sources yet (needed for librsvg)

Cargo TODO:
* https://github.com/rust-lang/cargo/pull/4711
* `CARGO_TARGET_DIR` should be an argument `--target-dir`
* Options to set the output library/executable name (prefix/suffix); perhaps also in `Cargo.toml`
* In general, the ability to more accurately control build outputs from via cargo arguments
  * This is where our work comes in; something half-working is the best way to figure out what is needed to make it work well ;)
* Dry-run downloading of sources by `cargo vendor` so we can ship a list of files instead of the sources themselves (cc @alexlarsson)

Cargo long-term TODO:
* Output `rustc` invocations needed to build a crate so we can run everything ourselves
[...]

PR #6603, opened 2020-02-09
Document the project policy on mixing build systems. [skip ci]

jpakkane:


docs/markdown/Mixing-build-systems.md
```diff
@@ -0,0 +1,51 @@
+# Meson's policy on mixing multiple build systems in one build directory
+
+Meson has been designed with the principle that all dependencies are
+either provided by "the platform" via a mechanism such as Pkg-Config
+or that they are built as Meson subprojects under the main
+project. There are several projects that would like to mix build
+systems, that is, build dependencies in the same build directory in
+the same build directory by having one build system call the other.
```
jon-turney: "in the same build directory in the same build directory" (duplicated)

docs/markdown/Mixing-build-systems.md
```diff
@@ -0,0 +1,51 @@
...
+
+This page lists the Meson project's stance on mixing build system. The
+tl/dr version is that while we do provide some functionality for this
+use case, it only works for simple cases. Anything more complex can
+not be made reliable. Thus the support for these use cases is not
+guaranteed to work (in fact it is almost certainly guaranteed to *not*
```
mensinda: I agree with making our policy on mixing build systems clear, but I think this part is a bit too harsh. Would something like `Thus the support for these use cases is unlikely to work.` also work?

Also maybe consider changing `it only works for simple cases` to `it only somewhat reliably works for simple cases`. IMHO if these mechanisms can't at least be classified to work "somewhat reliable" for small/simple projects (which *does* reflect the current status for CMake) then there is no point in having them :)

docs/markdown/Mixing-build-systems.md
```diff
@@ -0,0 +1,51 @@
...
+work), and even if it works today there are no guarantees that it will
+keep working in any future version.
+
+## The definition of "buils system mixing"
```
tp-m: Typo ("buils")

xclaesse:
I'm not sure if this means cargo module as proposed in https://github.com/mesonbuild/meson/pull/2617 would be accepted or not.

nirbheek:
> I'm not sure if this means cargo module as proposed in #2617 would be accepted or not.

I would interpret this as meaning "go ahead and merge it", because it meets all these requirements.

docs/markdown/Mixing-build-systems.md
```diff
@@ -0,0 +1,55 @@
+# Meson's policy on mixing multiple build systems in one build directory
+
+Meson has been designed with the principle that all dependencies are
+either provided by "the platform" via a mechanism such as Pkg-Config
+or that they are built as Meson subprojects under the main
+project. There are several projects that would like to mix build
+systems, that is, build dependencies in the same build directory as
+the other build system by having one build system call the other. The
+build diectories do not necessarily need to be inside each other, but
```
nirbheek: *build directories

docs/markdown/Mixing-build-systems.md
```diff
@@ -0,0 +1,55 @@
...
+that is the common case.
+
+This page lists the Meson project's stance on mixing build system. The
+tl/dr version is that while we do provide some functionality for this
+use case, it only works for simple cases. Anything more complex can
+not be made reliable and trying to do that would burden Meson
+developers with an effectively infinite maintenance burder. Thus these
```
nirbheek: *burden

docs/markdown/Mixing-build-systems.md
```diff
@@ -0,0 +1,55 @@
...
+Meson has been designed with the principle that all dependencies are
+either provided by "the platform" via a mechanism such as Pkg-Config
+or that they are built as Meson subprojects under the main
+project. There are several projects that would like to mix build
+systems, that is, build dependencies in the same build directory as
+the other build system by having one build system call the other. The
+build diectories do not necessarily need to be inside each other, but
+that is the common case.
+
+This page lists the Meson project's stance on mixing build system. The
```
nirbheek: *build systems

jpakkane:
The point of this PR was to make it possible to merge things like #2617. It does not mean that they will be automatically and blindly merged, though. They must be reviewed as usual and in addition every piece of functionality (mostly modules) that falls under this rule **must** start its documentation by referring to this guide (like the CMake module does in this commit).

PR #6606, opened 2020-02-10
Exclude .hg* from dist tarballs

### LINKED ISSUE #6575: .hgtags included in dist archive
**Describe the bug**

When running `ninja dist` the resulting `tar.xz` includes the `.hgtags` files that distros (namely debian) will mark as a packaging error.

**To Reproduce**

Run `ninja dist` on a mercurial repository that has at least 1 tag.

**Expected behavior**
That the archive does not have the `.hgtags` file in it.  I just did an `hg archive` and unfortunately saw that `.hgtags` was there as well.

**system parameters**
* what operating system (e.g. MacOS Catalina, Windows 10, CentOS 8.0, Ubuntu 18.04, etc.) Linux, Debian Unstable
* what Python version are you using e.g. 3.8.0: 3.7.6
* what `meson --version`: 0.53.1
* what `ninja --version` if it's a Ninja build: 1.9.0

rlaager:
Debian's `lintian` checker complaints when upstream tarballs contain an
`.hgtags` file.

This excludes `.hg[a-z]*`.  This is mostly consistent with the git
handling in del_gitfiles() which deletes `.git*`.  `hg archive --help` shows
an example of `-X ".hg*"`.  However, instead of `.hg*`, I have used
`.hg[a-z]*` to keep the automatically added `hg_archival.txt`.  This file
may be useful to link the tarball to the Mercurial revision for either
manual inspection or in case any code interprets it for a `--version` or
similar.

This also excludes `.hgignore` and other things like `.hgflow`, which seems
desirable.

Fixes #6575

### DIFF
--- a/mesonbuild/mdist.py
+++ b/mesonbuild/mdist.py
@@ -151,7 +151,16 @@ def create_dist_hg(dist_name, archives, src_root, bld_root, dist_sub, dist_scrip
     xzname = tarname + '.xz'
     gzname = tarname + '.gz'
     zipname = os.path.join(dist_sub, dist_name + '.zip')
-    subprocess.check_call(['hg', 'archive', '-R', src_root, '-S', '-t', 'tar', tarname])
+    # Note that -X interprets relative paths using the current working
+    # directory, not the repository root, so this must be an absolute path:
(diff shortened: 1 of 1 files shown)


jpakkane:
Thanks.

PR #6608, opened 2020-02-10
test cases/python3/3 cython: fix dependency files

gerion0:
The Cython generation depends on cstorer.pxd as well, so add that as
dependency.

A detection by cython itself [does not exist](https://github.com/cython/cython/issues/1214).


jpakkane:
Thanks.

PR #6610, opened 2020-02-11
configure_file: Also copy timestamps to avoid useless rebuilds

nirbheek:
If we always copy files over without timestamps, we're forcing a (probably unnecessary) rebuild. Also include a test for this.


xclaesse:
lgtm

Issue #4496, opened 2018-11-08
meson install with --only-changed fails if the target binary is running


mteyssier:
I got some issue while trying to install a binary which is already running.
Somewhat looks like #578

```
$ /install/bin/sample &
$ meson install --only-changed
ninja: Entering directory `.'
ninja: no work to do.
Preserving existing file /install/bin/sample.
Traceback (most recent call last):
  File "/usr/bin/meson", line 29, in <module>
    sys.exit(mesonmain.main())
  File "/usr/lib/python3/dist-packages/mesonbuild/mesonmain.py", line 379, in main
    return run(sys.argv[1:], launcher)
  File "/usr/lib/python3/dist-packages/mesonbuild/mesonmain.py", line 284, in run
    return minstall.run(remaining_args)
  File "/usr/lib/python3/dist-packages/mesonbuild/minstall.py", line 514, in run
    installer.do_install(datafilename)
  File "/usr/lib/python3/dist-packages/mesonbuild/minstall.py", line 346, in do_install
    self.install_targets(d)
  File "/usr/lib/python3/dist-packages/mesonbuild/minstall.py", line 490, in install_targets
    install_name_mappings, verbose=False)
  File "/usr/lib/python3/dist-packages/mesonbuild/scripts/depfixer.py", line 404, in fix_rpath
    fix_elf(fname, new_rpath, verbose)
  File "/usr/lib/python3/dist-packages/mesonbuild/scripts/depfixer.py", line 342, in fix_elf
    with Elf(fname, verbose) as e:
  File "/usr/lib/python3/dist-packages/mesonbuild/scripts/depfixer.py", line 117, in __init__
    self.bf = open(bfile, 'r+b')
OSError: [Errno 26] Text file busy: '/install/bin/sample'
Exception ignored in: <bound method Elf.__del__ of <mesonbuild.scripts.depfixer.Elf object at 0x7f36fa0737f0>>
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/mesonbuild/scripts/depfixer.py", line 132, in __del__
    if self.bf:
AttributeError: 'Elf' object has no attribute 'bf'
```

Running `meson install` without `--only-changed` works fine.

If I understand correctly, in `install_targets` from `minstall.py`:
- `do_copyfile` is called, the file didn't change so it is preserved (not removed, not copied, no issue yet).
- however `depfixer.fix_rpath` is still called and tries to open the file with write access which fails because the binary is running.

Is this use case supported?
[...]

PR #6612, opened 2020-02-11
meson install perf improvements and new option --quiet

nirbheek:
With these, `meson install` with gst-build on my Windows machine goes from taking 15.3 seconds to 4 seconds. `meson install --only-changed` goes from 11.6 seconds to 0.9 seconds.

Also fixes https://github.com/mesonbuild/meson/issues/4496.

### DIFF
--- a/mesonbuild/minstall.py
+++ b/mesonbuild/minstall.py
@@ -229,7 +229,7 @@ class Installer:
         to_time = os.stat(to_file).st_mtime
         return from_time <= to_time
 
-    def do_copyfile(self, from_file, to_file):
+    def do_copyfile(self, from_file, to_file, makedirs=None):
         outdir = os.path.split(to_file)[0]
         if not os.path.isfile(from_file) and not os.path.islink(from_file):
             raise RuntimeError('Tried to install something that isn\'t a file:'
@@ -246,6 +246,11 @@ class Installer:
                 self.preserved_file_count += 1
                 return False
             os.remove(to_file)
+        elif makedirs:
+            # Unpack tuple
+            dirmaker, outdir = makedirs
+            # Create dirs if needed
+            dirmaker.makedirs(outdir, exist_ok=True)
         print('Installing %s to %s' % (from_file, outdir))
         if os.path.islink(from_file):
             if not os.path.exists(from_file):
@@ -378,8 +383,7 @@ class Installer:
             outfilename = get_destdir_path(d, i[1])
             mode = i[2]
             outdir = os.path.dirname(outfilename)
-            d.dirmaker.makedirs(outdir, exist_ok=True)
-            if self.do_copyfile(fullfilename, outfilename):
+            if self.do_copyfile(fullfilename, outfilename, makedirs=(d.dirmaker, outdir)):
                 self.did_install_something = True
             set_mode(outfilename, mode, d.install_umask)
 
@@ -388,9 +392,8 @@ class Installer:
             full_source_filename = m[0]
             outfilename = get_destdir_path(d, m[1])
             outdir = os.path.dirname(outfilename)
-            d.dirmaker.makedirs(outdir, exist_ok=True)
             install_mode = m[2]
-            if self.do_copyfile(full_source_filename, outfilename):
+            if self.do_copyfile(full_source_filename, outfilename, makedirs=(d.dirmaker, outdir)):
                 self.did_install_something = True
             set_mode(outfilename, install_mode, d.install_umask)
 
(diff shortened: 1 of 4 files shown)


mesonbuild/scripts/depfixer.py
```diff
@@ -26,6 +26,9 @@
 DT_SONAME = 14
 DT_MIPS_RLD_MAP_REL = 1879048245
 
+# Global cache for tools
+INSTALL_NAME_TOOL = False
```
jpakkane: Shouldn't this be `INSTALL_NAME_TOOL = shutil.which('install_name_tool')` so it would get resolved on import? This would get of the `if` branching lookup code below.

nirbheek: In that case we will call `shutil.which` uselessly on non-mac platforms, which can take a while if your `PATH` is very large on Windows (which will happen eventually once you've installed enough dev tools, frameworks, SDKs, etc). 

F.ex., on my brand-new Windows machine with an NVMe drive and just 22 entries in `PATH`, 1000 `shutil.which` calls in a loop take 2.7s. However, on my old Windows machine with a Samsung Evo 950 SSD and 48 entries in `PATH`, the same takes 30 seconds! I expect it to be much slower on rotating drives.

That's 30ms per `shutil.which` (when the disk is fully idle), which is a pretty significant startup-time cost (and unnecessary I/O) in the no-op `meson install --only-changed` workflow (which is equivalent to the "one gstreamer plugin changed, and I just want to install that" workflow) for doing unnecessary work.

mesonbuild/scripts/depfixer.py
```diff
@@ -432,11 +432,12 @@ def fix_jar(fname):
...
-    # Static libraries never have rpaths
-    if fname.endswith('.a'):
+    # Static libraries, import libraries, debug information, headers, etc
+    # never have rpaths
+    if fname.endswith(('.a', '.lib', '.pdb', '.h', '.hpp')):
         return
-    # DLLs and EXE never have rpaths
-    if fname.endswith('.dll') or fname.endswith('.exe'):
+    # DLLs and EXE currently do not need runtime path fixing
+    if fname.endswith(('.dll', '.exe')):
```
xclaesse: nitpicking: why not adding dll/exe in the same list above?

nirbheek: I guessed the original code kept it separate because they were skipped for semantically different reasons (don't have RPATHs vs [we haven't implemented it yet](https://nibblestew.blogspot.com/2019/05/emulating-rpath-on-windows-via-binary.html)), so I kept it that way.

mesonbuild/minstall.py
```diff
@@ -321,6 +321,7 @@ def do_copydir(self, data, src_dir, dst_dir, exclude, install_mode):
                 abs_dst = os.path.join(dst_dir, filepart)
                 if os.path.isdir(abs_dst):
                     print('Tried to copy file %s but a directory of that name already exists.' % abs_dst)
+                    sys.exit(1)
```
xclaesse: I think it should raise MesonException() instead? There seems to be a mix of `raise RuntimeError` and `exit(1)` in minstall.py. But probably could go as follow-up cleanup.

nirbheek: minstall.py does not import `MesonException`. That sort of rewrite should probably happen separately. 

The difference between `RuntimeError` and `print()` + `exit(1)` is that the former is for programming errors and the latter is for expected errors.

mesonbuild/minstall.py
```diff
@@ -219,6 +221,10 @@ def __init__(self, options, lf):
         self.lf = lf
         self.preserved_file_count = 0
 
+    def log(self, msg):
+        if not self.options.quiet:
+            print(msg)
```
xclaesse: Shouldn't we use mlog.log() instead to reuse the existing logger mechanism? That gives as bonus the opportunity to use colors later.

nirbheek: Yes, we could! But minstall does not use mlog, I'd classify that as a rewrite/cleanup that should happen separately :)

xclaesse: indeed, doesn't have to be done here.

mesonbuild/minstall.py
xclaesse: That probably go together with using mlog, because raising MesonException is actually going to use mlog in mesonmain.py.

mesonbuild/scripts/depfixer.py
xclaesse: Hm, maybe... I think it was more because the one who write that didn't know the tuple syntax, so it would become quickly a long condition (I just learnt it now).

xclaesse: But since you're mixing multiple semantics in the first condition already, I would sneark exe/dll too. But again, just a little nitpicking, feel free to ignore :-)

mesonbuild/scripts/depfixer.py
jpakkane: This is completely reasonable. Please add it as a comment above the variable declaration, otherwise someone might accidentally "optimize" it away in the future.

PR #6614, opened 2020-02-12
cmake: Fix crash when no C++ compiler is not installed (fixes #6559)

### LINKED ISSUE #6559: meson claims Unknown compiler(s): ['c++', 'g++', 'clang++', 'pgc++', 'icpc']
This is kinda weird: I have a very small dummy meson.build which errors out with:

> meson.build:7:0: ERROR: Unknown compiler(s): ['c++', 'g++', 'clang++', 'pgc++', 'icpc']

Whereas it's correct I don't have any of those compilers on my system, there is also no indication in my meson.build that this would be needed.

The sample meson.build is:

```
project('sample',
	'c',
	version: '1.0.0',
	meson_version: '>= 0.47',
)

dep = dependency('doesnotexist', required: false)
```

mensinda:
No test, since this would require the absence of a C++ compiler.

### DIFF
--- a/mesonbuild/cmake/executor.py
+++ b/mesonbuild/cmake/executor.py
@@ -23,6 +23,7 @@ import re
 import os
 import shutil
 import ctypes
+import textwrap
 
 from .. import mlog, mesonlib
 from ..mesonlib import PerMachine, Popen_safe, version_compare, MachineChoice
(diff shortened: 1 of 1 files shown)


mesonbuild/cmake/executor.py
```diff
@@ -23,6 +23,7 @@
 import os
 import shutil
 import ctypes
+import textwrap
```
jon-turney: Is this reindentation also part of the fix?

mensinda: No, should I make a seperate PR/commit?

nirbheek:
You should be able to emulate the absence of a C++ compiler by setting the `CPP` env var to `false` or something like that, no?

mesonbuild/cmake/executor.py
nirbheek: Next time, yes, I'll just merge this for now :)

PR #6457, opened 2020-01-13
environment: Replace LD with <LANG>LD

dcbaker:
The rust code is ugly, because rust is annoying. It doesn't invoke a
linker directly (unless that linker is link.exe or lld-link.exe),
instead it invokes the C compiler (gcc or clang usually) to do it's
linking. Meson doesn't have good abstractions for this, though we
probably should because some of the D compilers do the same thing.
Either that or we should just call the c compiler directly, like vala
does.

This changes the public interface for meson, which we don't do unless we
absolutely have to. In this case I think we need to do it. A fair number
of projects have already been using 'ld' in their cross/native files to
get the ld binary and call it directly in custom_targets or generators,
and we broke that. While we could hit this problem again names like
`cld` and `cppld` are far less likely to cause collisions than `ld`.
Additionally this gives a way to set the linker on a per-compiler basis,
which is probably in itself very useful.

Fixes #6442

PR #6618, opened 2020-02-13
docs: Update documentation about linker selection

nirbheek:
We missed this in https://github.com/mesonbuild/meson/pull/6457

@jpakkane we will need to regenerate the website after merging this so that the incorrect documentation in the release notes is fixed: https://mesonbuild.com/Release-notes-for-0-53-0.html#generic-overrider-for-dynamic-linker-selection

### DIFF
--- a/docs/markdown/Release-notes-for-0.53.0.md
+++ b/docs/markdown/Release-notes-for-0.53.0.md
@@ -74,13 +74,31 @@ flags passed via language flags and hoped things worked out. In meson 0.52.0
 meson started detecting the linker and making intelligent decisions about
 using it. Unfortunately this broke choosing a non-default linker.
 
-Now there is a generic mechanism for doing this, you may use the LD
-environment variable (with normal meson environment variable rules), or add
-the following to a cross or native file:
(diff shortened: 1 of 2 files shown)


dcbaker:
I am very opposed to this as-is. Retroactively changing the release notes because we changed some behavior later is not what release notes are for, the [actual documentation](https://mesonbuild.com/howtox.html#page-description) is correct given the current behavior.

I'm not opposed to adding some sort of note: "This behavior was changed after 0.53.0, see the main documentation for the current behavior" blurb, but I don't think changing release notes ex post facto.

nirbheek:
> I'm not opposed to adding some sort of note: "This behavior was changed after 0.53.0, see the main documentation for the current behavior" blurb, but I don't think changing release notes ex post facto.

How about we document the behaviour in both versions? See updated commit. 

The release notes are only for 0.53.0 because we don't change API in stable/point releases. Now that we have, we need to document it in _some_ release notes somewhere, and this is the only one we have.

nirbheek:
>  I'm just concerned that people with the misfortune of running 0.53.0 will be confused if we claim that that 0.53.1 behavior is the 0.53 behavior.

Makes sense. Thanks for reviewing!

Issue #6622, opened 2020-02-13
Better error message:  ￼Write ￼Preview Describe the bug  This error message should be better. The error refers to a custom_target.  ERROR: Command cannot have '@INPUT@', since no input files were specified


haata:
**Describe the bug**

This error message should be better. The error refers to a custom_target.

```
ERROR: Command cannot have '@INPUT@', since no input files were specified
```
Ideally the file which file (and maybe which line) the custom_target was defined on.

**To Reproduce**
```
custom_target(
   'my_target',
   command: ['echo', '@INPUT@'],
   input: [],
)
```

**Expected behavior**
Show the filename (possibly the line number) as well as the name of the custom_target with the error.

**system parameters**

Python 3.6
meson 0.53.0
ninja 1.7.2

denizzzka:
Catched strictly same message but if `input:` contains object returned process() of generator. This is behaviour by design?

PR #6624, opened 2020-02-13
custom_target: Do substitutions in command while in interpreter

xclaesse:
Substitutions were already done in backend, but then we lost the
interpreter context for error messages.

Fixes: #6622


jpakkane:
This seems to break all tests that depend on the outputs of custom targets:

```
../test cases/common/13 pch/generated/pch/prog.h:1:36: fatal error: generated_customTarget.h: No such file or directory
```

xclaesse:
Yes it's more complicated that it seems. I have up for now.

PR #5978, opened 2019-09-30
Make parser errors print relative path to meson.build file

NickeZ:
Implemented this for fun.

Fixes: #5946

PR #6627, opened 2020-02-14
Consistently report file locations relative to cwd

jon-turney:
Consistently report file locations in errors and warnings relative to cwd, rather than the mixture of usually relative to meson source directory, but sometimes relative to cwd we have after #5978


jon-turney:
- [ ] rename `subdir` -> `filename` everywhere relevant
- [ ] unit test for warning file paths
- [ ] Snippet?

jon-turney:
After this we can probably drop making reported paths relative to cwd from mlog, and which will then just report the path to the meson source root as originally given on the command line (i.e. relative to cwd or absolute, but uncanonicalized).

jon-turney:
In the alternative, revert #5978 for consistent behaviour.

PR #6628, opened 2020-02-14
Update mesonlib.py

EdouardThiel:
I added expanduser_in_envpath() and called it in Popen_safe(), so as to solve the problem of ~ in $PATH.
Now, if we set for instance PATH="$PATH:~/.local/bin", meson setup builddir will find ninja (when user installed).
Checked on linux, not windows.
Hope it helps, cheers Edouard


nirbheek:
I'm going to close this PR, sorry! But we should definitely still document the `NINJA` env var somewhere.

PR #6630, opened 2020-02-15
modules/python: Do not error out if positional args are passed

nirbheek:
Also update the documentation.

Fixes https://github.com/mesonbuild/meson/issues/6470

### DIFF
--- a/docs/markdown/Python-module.md
+++ b/docs/markdown/Python-module.md
@@ -103,8 +103,9 @@ need to add `dependencies : py_installation.dependency()`, see [][`dependency()`
 python_dependency py_installation.dependency(...)
 ```
 
-This method accepts the same arguments as the standard [dependency] function and
-the following additional keyword arguments:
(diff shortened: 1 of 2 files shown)


xclaesse:
Hmmm I though we decided to keep it hard error. Do we have people complaining? The doc update is necessary indeed.

nirbheek:
> Hmmm I though we decided to keep it hard error. Do we have people complaining

I mean, people _have_ complained. Just because they've fixed the error in the master branches of their projects doesn't mean users aren't going to hit this when they build the existing releases of those projects.

bonzini:
> Just because they've fixed the error in the master branches of their projects doesn't mean users aren't going to hit this when they build the existing releases of those projects.

Right, if Meson wants to change this it has to do one of the following:

* introduce a mechanism where the version specified in `project()` affects not just warnings for unavailable features, but also behavior for (at least some) deprecated features

* encourage projects to ship all of Meson together with their releases (which is basically the Autotools model).

Issue #6633, opened 2020-02-16
Linker detecting on Windows with ninja


ur4t:
**Describe the bug**

With clang (not clang-cl) targeting msvc, meson generates `build.ninja` file using `libfoo.dll.a` (mingw style) to link, while actually generated import library file is `libfoo.lib` (along with `libfoo.dll`).

**To Reproduce**

`meson.build` 
```meson
project('demo', 'cpp')
executable('demo', 'demo.cpp', link_with: shared_library('foo', 'foo.cpp'))
```
`foo.h` 
```cpp
#include <cstdio>

#ifdef __EXPORTS
#define DllFunc __declspec(dllexport)
#else
#define DllFunc __declspec(dllimport)
#endif

extern "C" DllFunc void foo();
```
`foo.cpp`
```cpp
#define __EXPORTS
#include "foo.h"
void foo() { puts("Test!"); }
```
`demo.cpp`
```cpp
#include "foo.h"

int main() { foo(); }
```
Sections of generated `build.ninja`
```ninja
rule cpp_LINKER
 command = clang++ @$out.rsp
 rspfile = $out.rsp
 rspfile_content = $ARGS -Wl,/MACHINE:X64 -Wl,/OUT:$out $in $LINK_ARGS
 description = Linking target $out.

build demo.exe: cpp_LINKER demo@exe/demo.cpp.obj | foo@sha/libfoo.dll.symbols
 LINK_ARGS = "-Wl,/nologo" "-Wl,/release" "-Wl,/nologo"
 "-Wl,/DEBUG" "-Wl,/PDB:demo.pdb" "libfoo.dll.a" "-mconsole"
```

**Expected behavior**

Sections of working `build.ninja`
```ninja
rule cpp_LINKER
 command = clang++ @$out.rsp
 rspfile = $out.rsp
 rspfile_content = $ARGS -Wl,/MACHINE:X64 -Wl,/OUT:$out $in $LINK_ARGS
 description = Linking target $out.

build demo.exe: cpp_LINKER demo@exe/demo.cpp.obj | foo@sha/libfoo.dll.symbols
 LINK_ARGS = "-Wl,/nologo" "-Wl,/release" "-Wl,/nologo"
 "-Wl,/DEBUG" "-Wl,/PDB:demo.pdb" "libfoo.lib" "-mconsole"
```

**System parameters**
* Clang 9.0.0 Target: x86_64-pc-windows-msvc
* MSVC 19.24
* Windows 10
[...]

PR #6634, opened 2020-02-16
Fixed msvc style linker detecting bug on windows.

ur4t:
Actually `get_clink_dynamic_linker_and_stdlibs()` in `mesonbuild/build.py` returns a `Compiler`, but the return value is used in name of `linker` everywhere. Hard to fix further, so I added some comments useful for debugging and refactoring later.

### DIFF
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -1263,6 +1263,8 @@ You probably should put it in link_with instead.''')
                     if dl != linker.language:
                         stdlib_args += all_compilers[dl].language_stdlib_only_link_flags()
                         added_languages.add(dl)
+                # Type of var 'linker' is Compiler. 
+                # Pretty hard to fix because the return value is passed everywhere
                 return linker, stdlib_args
 
         m = 'Could not get a dynamic linker for build target {!r}'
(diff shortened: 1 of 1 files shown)


ur4t:
Close #6633.

nirbheek:
This looks correct to me. This combination (clang-not-cl with link.exe) is regressing because we don't test it on the CI right now. Would you be interested in adding that? Ideally as a GitHub action.

ur4t:
> 
> 
> This looks correct to me. This combination (clang-not-cl with link.exe) is regressing because we don't test it on the CI right now. Would you be interested in adding that? Ideally as a GitHub action.

I've manually tested windows test cases with the combination. Run fixed meson and ninja together in a script, both "clang-not-cl with link.exe" and "clang-cl with lld-link" tested,  in vs developer shell x64 native.

#### Meson failed cases: ####
- 5 resources `ERROR: Problem encountered: MESON_SKIP_TEST build_to_src has spaces`
- 12 resources with custom targets `ERROR: Problem encountered: MESON_SKIP_TEST build_to_src has spaces`
- 14 resources with custom target `ERROR: Could not find Windows resource compiler`
- 15 resource scripts with duplicate filenames `ERROR: Could not find Windows resource compiler`

#### Ninja failed cases: ####
- 6 vs module defs
- 7 dll versioning
- 9 vs module defs generated
- 10 vs module defs generated custom target
- 16 gui app

Having compared generated `build.ninja` and read logs, I found the following problems:
- With "clang-not-cl with link.exe", all generated `build.ninja` uses clang as linker, linker arguments are passed with `-Wl,XXX`. That is the nix style I've found in `build.ninja` generated by meson on linux.
- Case 6,9,10,16 are all linker errors, because clang as linker does not recognize vs def files.
- Case 7, import library files' names are not implicitly disclaimed in linker arguments, so generated .lib files are named versioned (along with versioned .dll files). But when linking executables, linker asks unversioned .lib files.

dcbaker:
@nirbheek, it's not a regression. Before 0.52 you couldn't use clang with link.exe, it was hard-coded to use a gnu-style linker. Just like running vanilla clang (not apple clang) on mac. I'm certain there are a lot of bugs in those cases.

PR #1751, opened 2017-05-06
Fix caching of external dependencies of various types

nirbheek:
All our cached_dep magic was totally useless since we ended up using the same identifier for native and cross deps. Just nuke all this `cached_dep` code since it is very error-prone and improve the identifier generation instead.

For instance, this is broken *right now* with the `type_name` kwarg.

Add a bunch of tests to ensure that all this actually works...

Closes https://github.com/mesonbuild/meson/issues/1736

Also list deps properly in mesonintrospect.

PR #1248, opened 2016-12-22
Force a Ninja regeneration before each project test

nirbheek:
We have no test coverage for regeneration at all, which is why issues like #1246 slide by without us noticing. With this, we will run a regen on every test during `ninja test` after it has been compiled. This will not affect test times too much since the regen will not rebuild anything at all since there have been no source changes.

PR #6636, opened 2020-02-16
Redetect machines when languages change

jon-turney:
A modification to `test cases/common/123 cpp and asm/meson.build` which demonstrates a problem with machine detection:

In the MSYS2 `MSYSTEM=MINGW32` environment (64-bit MSYS2 but with a `i686-w64-mingw32` targeted gcc as `gcc`)  this test fails, as it (incorrectly) tries to build retval-x86_64.S using an x86 compiler.

(See https://github.com/mesonbuild/meson/pull/6512/commits/4cd3015828766fb3bb297ea5a177de62889d7792 for discussion)

I was puzzled for a while why this only reproduced locally but not in CI, but this is due to the fact that it only configures incorrectly the *first* time (in CI, the `meson.build` is touched to force a reconfigure when building.  I assume some information is preserved in CoreData which causes the machine to be detected correctly when reconfiguring?)

I'm not suggesting the second commit here should be applied as is.  I guess we could always change from 'configure; reconfigure & build; reconfigure & test; install; clean' to something like  'configure; build; clean; reconfigure & build; reconfigure & test; install; clean', but doing that for all tests would make things even slower...

### DIFF
--- a/mesonbuild/interpreter.py
+++ b/mesonbuild/interpreter.py
@@ -2216,7 +2216,9 @@ class Interpreter(InterpreterBase):
         self.build_def_files = [os.path.join(self.subdir, environment.build_filename)]
         if not mock:
             self.parse_project()
+        self._redetect_machines()
 
+    def _redetect_machines(self):
         # Re-initialize machine descriptions. We can do a better job now because we
         # have the compilers needed to gain more knowledge, so wipe out old
         # inference and start over.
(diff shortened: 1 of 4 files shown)


jon-turney:
> I guess we could always change from 'configure; reconfigure & build; reconfigure & test; install; clean' to something like 'configure; build; clean; reconfigure & build; reconfigure & test; install; clean', but doing that for all tests would make things even slower...

After looking back at #1248 and #1751 a bit, I think we don't need to do quite that, adding an extra build step seems to be enough (so sequence is 'configure; build; reconfigure & build; reconfigure & test; install; clean').

- [ ] quantify the impact this has on total test time.

jpakkane:
LGTM are there still things needed to be done for this?

jon-turney:
> LGTM are there still things needed to be done for this?

Not that I am aware of.

PR #6637, opened 2020-02-16
Implement symbolextractor on windows + some cleanups/fixes

nirbheek:
commit 431283b35dff5b43ecc65fe19ae581b3c6d8b9be:

```
symbolextractor: Correctly filter undefined symbols on macOS
-g is --extern-only and -P is --format=posix. We were missing
--defined-only for some reason, which we pass to `nm` on Linux.
This avoids having to manually filter later.
```

commit 6fe7af58091f0cb9c7a4c41e6f714d5a8e9c9ad7:

```
symbolextractor: Print a warning if required tools not found
Also write out a dummy symbols file if the tool wasn't found or didn't
work instead of just spewing an exception.
```

commit 901bbc36d902c8146482295827e5464070bbf898:

```
symbolextractor: Support passing arguments to tools
This is how we parse all env vars for tools in Meson. Do the same here
too for consistency.
```

commit feb82e0f0f365d66199d697f1d99e109e2c6a700:

```
symbolextractor: Add typing hints
```

commit 5c0be89ba0ffb09e37d43f0e20fe180be39e3fe6:

```
ninjabackend: List PDBs in output list for targets
This is more correct, and forces the target(s) to be rebuilt if the
PDB files are missing. Increases the minimum required Ninja to 1.7,
which is available in Ubuntu 16.04 under backports.

We can't do the same for import libraries, because it is impossible
for us to know at configure time whether or not an import library will
be generated for a given DLL.
```

commit 0cbf397b4d3c03d2ad977c91ce3058b4cf96127a:

```
tests: Ensure that executable and library are named differently
On Windows, the basename is used to determine the name of the PDB
file. So for a project called myproject, we will create myproject.dll
and myproject.exe, both of which will have myproject.pdb. This is
a file collision. Instead, append `_test`, similar to the C# template.

Fixes AllPlatformTest.test_templates on MSVC. This became a hard error
when we started listing PDBs in the implicit outputs list of ninja
targets.

Do the same for a test that was making the same mistake.
```

commit 1c32a1f81ee5dbdbf519ddcd56c9b3753d158b99:

```
ninjabackend: Minor refactoring
```

commit 63c837fa669ea2917399ab1741a2f47744f511b4:

```
ninjabackend: Pass the import library to SHSYM
We actually use this while linking on Windows, and hence we need to
extract symbols from this file, and not the DLL.

However, we cannot pass it instead of the DLL because it's an optional
output of the compiler. It will not be written out at all if there are
no symbols in the DLL, and we cannot know that at configure time. This
means we cannot describe it as an output of any ninja target, or the
input of any ninja target. We must pass it as an argument without
semantic meaning.
```

commit 129ccd8469b97421ba1cd634911fd8b158281da5:

```
unit tests: Skip if pkg-config is not found
Of course, this does not skip on the CI, but helps on Windows.
```

commit b95dffc905ac1aade86bbf3749c5322156e9182e:

```
unit tests: Make assertBuildNoOp check stricter
We also need to verify that no CustomBuild targets were rebuilt.
```

commit f97ff1b901e3d1492e75232c1068e59d92dc009a:

```
unit tests: Add a test for reconfigure causing no-op build
meson setup && ninja && touch meson.build && ninja

should only reconfigure but not cause anything to be rebuilt.
```

commit 57b38563e92d69f9fbd7bee3dde54386f2146f3b:

```
unit tests: Add a test for the symbolchecker script
When a source file for a library is changed without adding new extern
symbols, only that library should be rebuilt. Nothing that uses it
should be relinked.

Along the way, also remove trailing `.` in all Ninja rule
descriptions. It's very confusing to see messages like:

    Linking target mylib.dll.

It's confusing that the period at the end of that is not part of the
filename. Instead of removing that period manually in the tests (which
feels wrong!) just don't print it at all.
```

commit fae49c582a93881e2b194eaf5f8679c3f02324e5:

```
symbolextractor: Add a Windows implementation
Supports both MSVC and MinGW toolchains. Checks for MSVC first, then
falls back to MinGW.
```

commit 6b8851776a39300b7414424466051b43bfa3a0a5:

```
symbolextractor: Add support for clang-cl
Requires the latest LLVm 9.0 release which implements the `-list`
argument to `llvm-lib` and ships with an implementation of `nm` called
`llvm-nm`.
```

commit 5ed20400b1d36448658e145c5ba947ae79a44bff:

```
symbolextractor: Add support for Cygwin
```

### DIFF
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -668,7 +668,7 @@ int dummy;
         (srcs, ofilenames, cmd) = self.eval_custom_target_command(target)
         deps = self.unwrap_dep_list(target)
         deps += self.get_custom_target_depend_files(target)
-        desc = 'Generating {0} with a {1} command.'
+        desc = 'Generating {0} with a {1} command'
         if target.build_always_stale:
             deps.append('PHONY')
         if target.depfile is None:
@@ -764,7 +764,7 @@ int dummy;
             target_name = 'meson-{}'.format(self.build_run_target_name(target))
             elem = NinjaBuildElement(self.all_outputs, target_name, 'CUSTOM_COMMAND', [])
             elem.add_item('COMMAND', cmd)
-            elem.add_item('description', 'Running external command %s.' % target.name)
+            elem.add_item('description', 'Running external command %s' % target.name)
             elem.add_item('pool', 'console')
             # Alias that runs the target defined above with the name the user specified
             self.create_target_alias(target_name)
@@ -789,7 +789,7 @@ int dummy;
     def generate_coverage_rules(self):
         e = NinjaBuildElement(self.all_outputs, 'meson-coverage', 'CUSTOM_COMMAND', 'PHONY')
         self.generate_coverage_command(e, [])
-        e.add_item('description', 'Generates coverage reports.')
+        e.add_item('description', 'Generates coverage reports')
         self.add_build(e)
         # Alias that runs the target defined above
         self.create_target_alias('meson-coverage')
@@ -798,21 +798,21 @@ int dummy;
     def generate_coverage_legacy_rules(self):
         e = NinjaBuildElement(self.all_outputs, 'meson-coverage-xml', 'CUSTOM_COMMAND', 'PHONY')
         self.generate_coverage_command(e, ['--xml'])
-        e.add_item('description', 'Generates XML coverage report.')
+        e.add_item('description', 'Generates XML coverage report')
         self.add_build(e)
         # Alias that runs the target defined above
         self.create_target_alias('meson-coverage-xml')
 
         e = NinjaBuildElement(self.all_outputs, 'meson-coverage-text', 'CUSTOM_COMMAND', 'PHONY')
         self.generate_coverage_command(e, ['--text'])
-        e.add_item('description', 'Generates text coverage report.')
+        e.add_item('description', 'Generates text coverage report')
         self.add_build(e)
         # Alias that runs the target defined above
         self.create_target_alias('meson-coverage-text')
 
         e = NinjaBuildElement(self.all_outputs, 'meson-coverage-html', 'CUSTOM_COMMAND', 'PHONY')
         self.generate_coverage_command(e, ['--html'])
-        e.add_item('description', 'Generates HTML coverage report.')
+        e.add_item('description', 'Generates HTML coverage report')
         self.add_build(e)
         # Alias that runs the target defined above
         self.create_target_alias('meson-coverage-html')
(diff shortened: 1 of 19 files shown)


nirbheek:
Hmmmm, test fails on `clang-cl` in a weird way. I guess I need to set it up locally to figure out what's happening.

nirbheek:
All tests pass, ready for review! Super-excited about this PR. gst-build on windows is much nicer now.

jpakkane:
Good stuff, thanks. There should probably be a snippet describing the minimum Ninja version requirement bump.

run_tests.py
```diff
@@ -45,15 +45,13 @@
...
-            break
+    # is fixed
+    NINJA_CMD = detect_ninja('1.9')
+    if NINJA_CMD is not None:
+        NINJA_1_9_OR_NEWER = True
+    else:
+        mlog.warning('Found ninja <1.9, tests will run slower', once=True)
+        NINJA_CMD = detect_ninja()
 if NINJA_CMD is None:
     raise RuntimeError('Could not find Ninja v1.6 or newer')
```
jpakkane: This should probably now say 1.7?

mesonbuild/backend/ninjabackend.py
```diff
@@ -2300,12 +2309,17 @@ def generate_pch(self, target, header_deps=None):
...
+        return os.path.join(targetdir, target.get_filename() + '.symbols')
+
     def generate_shsym(self, target):
-        target_name = target.get_filename()
         target_file = self.get_target_filename(target)
-        targetdir = self.get_target_private_dir(target)
-        symname = os.path.join(targetdir, target_name + '.symbols')
+        symname = self.get_target_shsym_filename(target)
         elem = NinjaBuildElement(self.all_outputs, symname, 'SHSYM', target_file)
+        # The library we will actually link to, which is an import library on Windows (not the DLL)
```
jpakkane: To be pedantic, shouldn't this say that the import library is used with VS (and maybe clang?) rather than Windows in general

nirbheek: The ability to link to a DLL is a hack in GCC where it tries to guess what the import library would look like for the DLL (including guessing the calling convention), and so on. It's really bad to do that.

You should _always_ use the import library even with GCC because the import library can contain, f.ex., stubs that refer to multiple DLLs, or constants that aren't present in the DLL at all, etc.

So to be ultra-pedantic, DLLs should never be used for linking!

nirbheek:
> Good stuff, thanks. There should probably be a snippet describing the minimum Ninja version requirement bump.

Thanks! Snippet added.

mesonbuild/backend/ninjabackend.py
nirbheek: Also, in meson we always use the import library for linking on Windows. We don't have a way to link directly to the DLL at all.

PR #6641, opened 2020-02-17
cmake: Ignore files that violate subproject isolation (fixes #6640)

### LINKED ISSUE #6640: CMake subproject fails with "Tried to form an absolute path to a source dir"
**Describe the bug**
I have a simple Meson project that uses a CMake subproject. I pass an additional CMake option in using `cmake_options`, specifying `BOOST_ROOT`.

When configuring, I see the subproject getting configured with CMake correctly, but as soon as the configuration succeeds, Meson quits with an error:

> subprojects/mySubproject/meson.build:0:0: ERROR: Tried to form an absolute path to a source dir. You should not do that but use relative paths instead.

(The error message is followed by an explanation on how to use `include_directories`.)
Note that the error message points to a nonexistent file `subprojects/mySubproject/meson.build` in line 0.

**To Reproduce**

Here's my `meson.build`:

```
project(
    'myApp',
    'cpp',
    version : '0.0.1'
)

cmake = import('cmake')

boost_root = meson.source_root() + '/myBoost'

message('Configuring mySubproject with BOOST_ROOT = ' + boost_root)
proj = cmake.subproject('mySubproject', cmake_options : [ '-DBOOST_ROOT=' + boost_root ])
```

**Expected behavior**
It should either provide better diagnostics, or not fail.

**system parameters**
* Native build
* OS: Oracle Linux 7.7 (based on CentOS)
* Python: 3.7.4
* Meson version: 0.53.1
* Ninja version: 1.9.0

mensinda:


docs/markdown/Contributing.md
```diff
@@ -170,6 +170,13 @@ Various constructs containing `?` are used to indicate platform specific
 filename variations (e.g. `?so` represents the platform appropriate suffix for a
 shared library)
 
+* `fails_on.txt` specifies the stage, where a failing tests is expected to fail.
+If this file is present, the test case must fail specified stage. Possible stages
+are:
+  - meson
+  - build
+  - test
+
```
jon-turney: It might be worthwhile to make this slightly more generic (e.g. a key=value in dict in a json file or whatever), so we don't have to add yet another file for yet another test configuration variation in future.

Or maybe you could just put this test in 'test cases/failing'? (since no validation is currently done of how those failing tests are failing (so failing due to missing cmake is indistinguishable))

mensinda: I am planning to refactor the testing framework (for the next big release probably), so that there is only one big test definition file and we don't have `installed_files.txt`, `test_args.txt`, `setup_env.json`, `test_matrix.json` and `fails_on.txt`. So, having one extra file now is only a bit more work for me later on.

I can't put this in the failing directories, since it fails on build, and not on configure.

jpakkane: You can do something like:

```meson
if not find_program('cmake': required: false).found()
  executable('bob', 'a_non_compiling_source.cpp')
endif
```

mensinda: Moved it to `failing build` for now.

PR #6642, opened 2020-02-18
Skip Py2 tests correctly also on other systems than Debian

mimi1vx:


dcbaker:
systemd says you need the `os-version` file, so pretty much every Linux system is going to have it. We should just drop that condition and run the pkg-config code always (or at least when there is pkg-config) for Windows we should probably be checking if `py -2` works. For mac I'm not sure.

mimi1vx:
@dcbaker all modern distributions use `/etc/os-release` :D but probably better test  here will be use `sys.plattform`

eli-schwartz:
```
if is_windows():
    do_stuff()
if os.path.exists('/etc/os-release'):
    do_other_stuff()
```

The second check should just be `elif True:`...

On any Debian system that doesn't have os-release, the python2 version will be older than Bullseye, which the code was designed to check for in the first place.

I agree this should just key off of the platform, and unconditionally check for python2 via pkg-config on all non-Windows systems, since all non-Windows systems have pkg-config and meson already expects pkg-config to exist.

jpakkane:
> unconditionally check for python2 via pkg-config 

With `pkg-config --exists python2`?

dcbaker:
Probably something like:

```python
if is_windows():
    has_py2 = subprocess.call(['py', '-2'], ...)
else:
    has_py2 = subprocess.call(['pkg-config', '--exists', 'python2'], ...)
```

mimi1vx:
> Probably something like:
> 
> ```python
> if is_windows():
>     has_py2 = subprocess.call(['py', '-2'], ...)
> else:
>     has_py2 = subprocess.call(['pkg-config', '--exists', 'python2'], ...)
> ```

ant then check for  python2 devel files ?

dcbaker:
pkg-config is checking for the devel files, and on windows you can't split them (I think)

mimi1vx:
looks like somebody implemented Your idea :D :+1:

PR #6643, opened 2020-02-18
docs: Fix Dub section being inside CMake section

nh2:
It should be a `##` heading, not a `###` heading.

Also make heading label consistent with the others.


jpakkane:
Thanks, but note in the future that doc-only changes should have `[skip ci]` at the end of their commit messages so CI is not run.

Issue #6646, opened 2020-02-18
summary() broken with --reconfigure --debug and a bool variable


nh2:
**Describe the bug**

`meson.build`:

```meson
summary({'asdf': true})
```

Run:

```
rm -rf build && meson setup build && meson setup build --debug --reconfigure
```

The first `setup` works, the `setup --debug --reconfigure` breaks, as shown in this output:

```
serkezet 1.0

    asdf: True

Found ninja-1.8.2 at /nix/store/v3rwd0gqjzvr8zppxy3w0nli2q7n0j54-ninja-1.8.2/bin/ninja
Traceback (most recent call last):
  File "/nix/store/xcar77agh5khbzcr1npmac56qbk1wgxy-meson-0.53.1/lib/python3.6/site-packages/mesonbuild/mesonmain.py", line 129, in run
    return options.run_func(options)
  File "/nix/store/xcar77agh5khbzcr1npmac56qbk1wgxy-meson-0.53.1/lib/python3.6/site-packages/mesonbuild/msetup.py", line 245, in run
    app.generate()
  File "/nix/store/xcar77agh5khbzcr1npmac56qbk1wgxy-meson-0.53.1/lib/python3.6/site-packages/mesonbuild/msetup.py", line 159, in generate
    self._generate(env)
  File "/nix/store/xcar77agh5khbzcr1npmac56qbk1wgxy-meson-0.53.1/lib/python3.6/site-packages/mesonbuild/msetup.py", line 220, in _generate
    coredata.update_cmd_line_file(self.build_dir, self.options)
  File "/nix/store/xcar77agh5khbzcr1npmac56qbk1wgxy-meson-0.53.1/lib/python3.6/site-packages/mesonbuild/coredata.py", line 836, in update_cmd_line_file
    config['options'].update(options.cmd_line_options)
  File "/nix/store/i1hh6prcyfqd066jdh4zwqfs58mqhdfw-python3-3.6.6/lib/python3.6/_collections_abc.py", line 841, in update
    self[key] = other[key]
  File "/nix/store/i1hh6prcyfqd066jdh4zwqfs58mqhdfw-python3-3.6.6/lib/python3.6/configparser.py", line 1237, in __setitem__
    self._parser._validate_value_types(option=key, value=value)
  File "/nix/store/i1hh6prcyfqd066jdh4zwqfs58mqhdfw-python3-3.6.6/lib/python3.6/configparser.py", line 1177, in _validate_value_types
[...]

PR #6647, opened 2020-02-18
coredata: Fix crash when cmd_line_options contains booleans

xclaesse:
`meson setup --debug` puts a bool object in cmd_line_options dictionary
that we have to convert to string.

Fixes: #6646

### DIFF
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -850,6 +850,9 @@ def read_cmd_line_file(build_dir, options):
         # literal_eval to get it into the list of strings.
         options.native_file = ast.literal_eval(properties.get('native_file', '[]'))
 
+def cmd_line_options_to_string(options):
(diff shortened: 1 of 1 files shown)


mesonbuild/coredata.py
```diff
@@ -850,6 +850,9 @@ def read_cmd_line_file(build_dir, options):
         # literal_eval to get it into the list of strings.
         options.native_file = ast.literal_eval(properties.get('native_file', '[]'))
 
+def cmd_line_options_to_string(options):
+    return {k: v if isinstance(v, str) else str(v) for k, v in options.cmd_line_options.items()}
```
dcbaker: str(str)` is fast pathed in the python anyway, I think its clearer and easier if we skip the ternary and `isinstance` and just  do:

`{k: str(v) for k, v in options...}`

xclaesse: Good point, I made that change now.

PR #6649, opened 2020-02-18
docs: Improve visibility for `.enabled()` etc., add example

nh2:
Following discussion with @xclaesse on IRC.

### DIFF
--- a/docs/markdown/Build-options.md
+++ b/docs/markdown/Build-options.md
@@ -83,14 +83,29 @@ Currently supported in
 - `disabled` do not look for the dependency and always return 'not-found'.
 
 When getting the value of this type of option using `get_option()`, a special
-object is returned instead of the string representation of the option's value.
-That object has three methods returning boolean and taking no argument:
-`enabled()`, `disabled()`, and `auto()`.
+[feature option object](Reference-manual.md#feature-option-object)
+is returned instead of the string representation of the option's value.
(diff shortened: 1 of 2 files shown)


docs/markdown/Reference-manual.md
```diff
@@ -2000,6 +2000,16 @@ Dictionaries are available since 0.47.0.
 Since 0.48.0 dictionaries can be added (e.g. `d1 = d2 + d3` and `d1 += d2`).
 Values from the second dictionary overrides values from the first.
 
+### Feature option object
```
xclaesse: This should be moved in the `## Returned objects` section just below.

jon-turney: also, there should be a link to this from the text in `get_option()` above which mentions these objects.

nh2: Both done.

nh2:
Feedback addressed, please have another look.

docs/markdown/Build-options.md
```diff
@@ -94,6 +94,21 @@ if d.found()
 endif
 ```
 
+To check the value of the feature, the object has three methods
+returning boolean and taking no argument:
```
jon-turney: "returning a boolean"

nh2: Done.

PR #6653, opened 2020-02-18
Fix `compiler.get_linker_id` and documentation

dcbaker:
The documentation leaves out a lot of possible values, and is wrong in a number of cases.

On the other end calling `meson.get_compiler('java').get_linker_id()` would cause an uncaught python exception.

### DIFF
--- a/docs/markdown/Reference-tables.md
+++ b/docs/markdown/Reference-tables.md
@@ -33,16 +33,26 @@ These are return values of the `get_id` (Compiler family) and
 
 ## Linker ids
 
-These are return values of the `get_linker_id` (Linker family) method in a compiler object.
+These are return values of the `get_linker_id` method in a compiler object.
 
-| Value     | Linker family                              |
-| -----     | ---------------                            |
-| ld.bfd    | GNU Compiler Collection                    |
-| {ld.bfd,lld} | Clang non-Windows                       |
-| link      | MSVC, Clang-cl, clang Windows              |
-| pgi       | Portland/Nvidia PGI                        |
-| {ld.bfd,gold,xild}  | Intel compiler non-Windows       |
-| xilink    | Intel-cl Windows                           |
+| Value     | Linker family                               |
+| -----     | ---------------                             |
+| ld.bfd    | The GNU linker                              |
+| ld.gold   | The GNU linker                              |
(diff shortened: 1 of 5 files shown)


docs/markdown/Reference-tables.md
```diff
@@ -33,16 +33,23 @@ These are return values of the `get_id` (Compiler family) and
...
+| lld       | The LLVM linker, with the GNU interface     |
+| link      | MSVC linker                                 |
+| pgi       | Portland/Nvidia PGI                         |
+| lld-link  | The LLVM linker, with the MSVC interface    |
+| xilink    | Used with Intel-cl only, MSVC like          |
+| rlink     | The Renesas linker, used with CCrx only     |
+| armlink   | The ARM linker (arm and armclang compilers) |
+| optlink   | optlink                                     |
+| APPLE ld  | Apple ld64                                  |
+| solaris   | Solaris and illumos                         |
```
nirbheek: Maybe we should have a unit test for this in the same way that we do for the compiler ids?

Also, since the other commit makes `get_linker_id()` return compiler ids in some cases, should we note that somewhere here?

dcbaker: Probably. I'm actually annoyed that I missed this change, because the ids are extremely inconsistent as I never meant them to be exposed in meson language...

nirbheek: I think we can change just "APPLE ld", "pgi" and "solaris" for some semblance of consistancy. At least the rest kinda match the linker binary name.

dcbaker: I'd really also like to rename lld to ld.lld, since there's also lld-link and there's an lld that acts like apple's ld (although we don't support that ATM)

nirbheek: I think we can do that.

dcbaker: I've added a new patch to make those changes, and to hopefully make the linkers more robust against being accidentally renamed (moved the name into the class)

dcbaker:
I'm not sure about the tests. Those work well because they can be fully automated, but there isn't a container with all of the names in it for the linkers, they're class attributes. Which means we'd have to import them all and do a comparison. I don't know if that is useful or not?

PR #6654, opened 2020-02-18
Add .found() method to cmakesubproject

dcbaker:
Otherwise it's impossible to figure out what happened with `cmake.subproject(..., required : false)`. Also, normal subprojects have this, so not having it is very strange.


wak-google:
Thanks

PR #6656, opened 2020-02-19
Update one extra use of unversioned python

seb128:
Thanks Stefano Rivera <stefanor@ubuntu.com> for the patch


jpakkane:
Test failure seems relevant:

```
Program python3 found: YES (/usr/bin/python3)

meson.build:20:2: ERROR: Assert failed: Didn't load python from config file

meson.build:20:2: ERROR: Assert failed: Didn't load python from config file
A full log can be found at /home/runner/work/meson/meson/tmp86r_ijzv/meson-logs/meson-log.txt

Build started at 2020-02-19T08:57:09.509698
```

seb128:
Indeed, a bit weird though, the patch is applied to the ubuntu package and tests are not failing there...

LocutusOfBorg:
looks like #6703 might be the best solution

seb128:
let's close this one then, it needed work anyway

PR #6658, opened 2020-02-19
Update minit.py

michaelbadcrumble:
Doing some cleaning for the `meson init` command, created constant variables for any values that are repeated, and added missing info for the `meson init --help` flag.

I was wondering if any new programming languages will be added to Meson build so I may create a new batch for them?

If no new programming languages are being added at this time, I was also wondering if there are some other parts of Meson I can work on?

### DIFF
--- a/mesonbuild/minit.py
+++ b/mesonbuild/minit.py
@@ -40,7 +40,8 @@ we currently have one meson template at this time.
 from mesonbuild.templates.mesontemplates import create_meson_build
 
 FORTRAN_SUFFIXES = ('.f', '.for', '.F', '.f90', '.F90')
-LANG_SUFFIXES = {'.c', '.cc', '.cpp', '.cs', '.cu', '.d', '.m', '.mm', '.rs', '.java'}.add(FORTRAN_SUFFIXES)
+LANG_SUFFIXES = {'.c', '.cc', '.cpp', '.cs', '.cu', '.d', '.m', '.mm', '.rs', '.java'}
+LANG_SUFFIXES.add(FORTRAN_SUFFIXES)
 LANG_SUPPORTED = {'c', 'cpp', 'cs', 'cuda', 'd', 'fortran', 'java', 'rust', 'objc', 'objcpp'}
 
 DEFAULT_PROJECT = 'executable'
--- a/mesonbuild/minit.py
+++ b/mesonbuild/minit.py
@@ -217,7 +217,7 @@ def add_arguments(parser):
     parser.add_argument("-b", "--build", action='store_true', help="build after generation")
     parser.add_argument("--builddir", default='build', help="directory for build")
     parser.add_argument("-f", "--force", action="store_true", help="force overwrite of existing files and directories.")
-    parser.add_argument('--type', default='executable', choices=('executable', 'library'), help="project type. default: {} based project".format(DEFAULT_PROJECT))
+    parser.add_argument('--type', default=DEFAULT_PROJECT, choices=('executable', 'library'), help="project type. default: {} based project".format(DEFAULT_PROJECT))
     parser.add_argument('--version', default=DEFAULT_VERSION, help="project version. default: {}".format(DEFAULT_VERSION))
 
 def run(options) -> int:
--- a/mesonbuild/minit.py
+++ b/mesonbuild/minit.py
@@ -15,6 +15,7 @@
 """Code that creates simple startup projects."""
 
 from pathlib import Path
+from enum import Enum
 import subprocess
 import shutil
 import re
@@ -42,10 +43,11 @@ FORTRAN_SUFFIXES = ('.f', '.for', '.F', '.f90', '.F90')
 LANG_SUFFIXES = ('.c', '.cc', '.cpp', '.cs', '.cu', '.d', '.m', '.mm', '.rs', '.java') + FORTRAN_SUFFIXES
 LANG_SUPPORTED = ('c', 'cpp', 'cs', 'cuda', 'd', 'fortran', 'java', 'rust', 'objc', 'objcpp')
 
-DEFAULT_EXE = 'executable'
-DEFAULT_LIB = 'library'
 DEFAULT_PROJECT = 'executable'
 DEFAULT_VERSION = '0.1'
+class DEFAULT_TYPES(Enum):
+    EXE = 'executable'
+    LIB = 'library'
 
 UNREACHABLE_CODE_ERROR = 'Unreachable code'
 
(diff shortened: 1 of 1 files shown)


mesonbuild/minit.py
```diff
@@ -15,118 +15,125 @@
...
-there is currently only one meson template at this time.
+we currently have one meson template at this time.
 '''
 from mesonbuild.templates.mesontemplates import create_meson_build
 
 FORTRAN_SUFFIXES = ['.f', '.for', '.F', '.f90', '.F90']
+LANG_SUFFIXES = ['.c', '.cc', '.cpp', '.cs', '.cu', '.d', '.m', '.mm', '.rs', '.java'] + FORTRAN_SUFFIXES
+LANG_SUPPORTED = ('c', 'cpp', 'cs', 'cuda', 'd', 'fortran', 'java', 'rust', 'objc', 'objcpp')
+TYPE_EXE = 'executable'
+TYPE_LIB = 'library'
```
dcbaker: These two should probably be an enum?

mesonbuild/minit.py
```diff
@@ -15,118 +15,125 @@
...
+from mesonbuild.templates.ctemplates import (create_exe_c_sample, create_lib_c_sample)
 
 '''
-there is currently only one meson template at this time.
+we currently have one meson template at this time.
 '''
 from mesonbuild.templates.mesontemplates import create_meson_build
 
 FORTRAN_SUFFIXES = ['.f', '.for', '.F', '.f90', '.F90']
+LANG_SUFFIXES = ['.c', '.cc', '.cpp', '.cs', '.cu', '.d', '.m', '.mm', '.rs', '.java'] + FORTRAN_SUFFIXES
```
dcbaker: Both of these should be sets not lists, as we're only doing membership tests.

mesonbuild/minit.py
```diff
@@ -148,7 +155,7 @@ def autodetect_options(options, sample: bool = False):
     if not options.srcfiles:
         srcfiles = []
         for f in (f for f in Path().iterdir() if f.is_file()):
-            if f.suffix in (['.c', '.cc', '.cpp', '.cs', '.cu', '.d', '.m', '.mm', '.rs', '.java'] + FORTRAN_SUFFIXES):
+            if f.suffix in (LANG_SUFFIXES):
```
dcbaker: the parens here are now unnecessary.

mesonbuild/minit.py
```diff
@@ -201,13 +208,12 @@ def add_arguments(parser):
...
-                        help="project language. default: autodetected based on source files")
-    parser.add_argument("-b", "--build", help="build after generation", action='store_true')
-    parser.add_argument("--builddir", help="directory for build", default='build')
+    parser.add_argument("-l", "--language", choices=LANG_SUPPORTED, help="project language. default: autodetected based on source files")
+    parser.add_argument("-b", "--build", action='store_true', help="build after generation")
+    parser.add_argument("--builddir", default='build', help="directory for build")
     parser.add_argument("-f", "--force", action="store_true", help="force overwrite of existing files and directories.")
-    parser.add_argument('--type', default='executable', choices=['executable', 'library'])
-    parser.add_argument('--version', default='0.1')
+    parser.add_argument('--type', default='executable', choices=['executable', 'library'], help="project type. default: executable based project")
```
dcbaker: if you make this an enum as I suggested you can use the `type` parameter here to have argparse do the conversion for you.

mesonbuild/minit.py
```diff
@@ -201,13 +208,12 @@ def add_arguments(parser):
...
+    parser.add_argument('--version', default='0.1', help="project version. default: 0.1")
```
dcbaker: In this case I think you can use `help="project version, default: %(default)s"` to have the default value be auto updated.

michaelbadcrumble: This seems like an interesting suggestion but I’m not sure if it makes sense, it feels like a nice to have.

mesonbuild/minit.py
michaelbadcrumble: Made those into sets for membership test.

mesonbuild/minit.py
michaelbadcrumble: Ok, remove redundant braces.

mesonbuild/minit.py
michaelbadcrumble: Is there a particular conversation for enums?

dcbaker: I can live without the enums, but it's just nice from a readability perspective to say "ah, these are the value values."

mesonbuild/minit.py
dcbaker: To be clear, what this does is (In case it did't make sense) is that argparse will do the string interpolation for us, so if someone changes the default version in the python code the help message will match. I would really prefer not to be repeating ourselves when the framework eliminates the need to do so.

michaelbadcrumble: removed hardcoded values for more add less work in the long run.

mesonbuild/minit.py
michaelbadcrumble: making use of enums via enum class.

michaelbadcrumble:
Ping @dcbaker .

michaelbadcrumble:
Ok @dcbaker I believe the changes have been applied correctly.

mesonbuild/minit.py
```diff
@@ -15,118 +15,130 @@
...
 
 '''
-there is currently only one meson template at this time.
+we currently have one meson template at this time.
 '''
 from mesonbuild.templates.mesontemplates import create_meson_build
 
-FORTRAN_SUFFIXES = ['.f', '.for', '.F', '.f90', '.F90']
+FORTRAN_SUFFIXES = ('.f', '.for', '.F', '.f90', '.F90')
+LANG_SUFFIXES = ('.c', '.cc', '.cpp', '.cs', '.cu', '.d', '.m', '.mm', '.rs', '.java') + FORTRAN_SUFFIXES
```
dcbaker: these are tuples, not sets, sets are in the form:
```python
{'.f', '.for', '.F'}
```

Otherwise this looks good.

michaelbadcrumble: I was wondering if I missed something.  Change applied.

mesonbuild/minit.py
```diff
@@ -15,118 +15,130 @@
...
 
 '''
-there is currently only one meson template at this time.
+we currently have one meson template at this time.
 '''
 from mesonbuild.templates.mesontemplates import create_meson_build
 
-FORTRAN_SUFFIXES = ['.f', '.for', '.F', '.f90', '.F90']
+FORTRAN_SUFFIXES = ('.f', '.for', '.F', '.f90', '.F90')
+LANG_SUFFIXES = {'.c', '.cc', '.cpp', '.cs', '.cu', '.d', '.m', '.mm', '.rs', '.java'}.add(FORTRAN_SUFFIXES)
```
dcbaker: `.add` doesn't return anything. I think you want `{...} | FORTAN_SUFFIXES`

mesonbuild/minit.py
michaelbadcrumble: There was an issue with adding to an assigned set structure. But I fixed it.

mesonbuild/minit.py
michaelbadcrumble: Ok applied suggested change.  I also changed the Fortran suffix structure to set.

PR #6664, opened 2020-02-20
WIP: Add batch get_supported_headers/functions() and parallelize them

xclaesse:
With minimal changes in GLib to use the new APIs, configure time goes from 0m11,477s to 0m8,604s with this branch. There are still a few places where glib could use batch APIs, but I did the 3 main ones (args, headers, functions).


xclaesse:
gst-build with no changes goes from 53,194s to 44,837s which is surprisingly good given that I did not change to use get_supported_headers/function() and I've seen some GStreamer modules not even using get_supported_arguments() where it could. At least this time I verified it's actually building the same set of subprojects, so those numbers should be real.

xclaesse:
With recent optimizations landed in master, gst-build improved a lot:
master: 0m38,683s
master+PR: 0m31,980s

PR #6438, opened 2020-01-09
Postprocessors for the meson python code

mensinda:
This PR introduces a modular python code post-processor for meson. The main use-case here would be to support older versions of python (specifically 3.5.2) while using newer python features, which can be automatically removed or replaced with an older equivalent.

This should resolve #6427 and give us most of the 3.6 features (see #6297) while maintaining python 3.5 compatibility (at the cost of a "compiler"/"transpiler").

PR #6666, opened 2020-02-20
Refactor run_projectests.py

mensinda:
Some cleanup in `run_projectests.py` and merging of the many test definition files into a single `test.json`.


jpakkane:
⚔️ conflicts.

run_project_tests.py
```diff
@@ -106,10 +154,12 @@ def display_name(self) -> str:
             return '{}   ({})'.format(self.path.as_posix(), self.name)
         return self.path.as_posix()
 
-    def __lt__(self, other: T.Any) -> T.Union[bool, type(NotImplemented)]:
+    def __lt__(self, other: T.Any) -> bool:
```
dcbaker: This change is wrong, all of the rich comparison methods return a Untion of bool and NotImplemented

run_project_tests.py
```diff
@@ -106,10 +154,12 @@ def display_name(self) -> str:
...
         if isinstance(other, TestDef):
             # None is not sortable, so replace it with an empty string
-            return (self.path, self.name or '') < (other.path, other.name or '')
+            s_id = int(self.path.name.split(' ')[0])
+            o_id = int(other.path.name.split(' ')[0])
+            return (s_id, self.path, self.name or '') < (o_id, other.path, other.name or '')
```
dcbaker: do we really need to compare path if we're already comparing the number in the path?

run_project_tests.py
```diff
@@ -360,83 +355,60 @@ def run_test_inprocess(testdir):
...
             try:
-                return _run_test(testdir, build_dir, install_dir, extra_args, compiler, backend, flags, commands, should_fail)
+                return _run_test(test, build_dir, install_dir, extra_args, compiler, backend, flags, commands, should_fail)
             finally:
                 mlog.shutdown() # Close the log file because otherwise Windows wets itself.
 
-def pass_prefix_to_test(dirname):
-    if '39 prefix absolute' in dirname:
+def pass_prefix_to_test(dirname: Path):
+    if '39 prefix absolute' in dirname.name:
```
dcbaker: While you're hear, the return type is `bool`, and this can be simplified to simply 
```python
def pass_prefix_to_test(dirname: Path) -> bool:
    return '39 prefix absolute' not in dirname.name
```

run_project_tests.py
mensinda: We already had cases that the same number was used twice after merging two PR's, and the name will be `Null` for non test matrix tests.

run_project_tests.py
mensinda: This is the only way to make both mypy and python 3.5.2 happy. See https://github.com/mesonbuild/meson/pull/6438#issuecomment-584644255

run_project_tests.py
dcbaker: Ah, okay that makes sense.

run_project_tests.py
dcbaker: <rant>I'm seriously tired of bending over backwards to try to make ancient versions of ubuntu work with modern code bases </rant>.

I've seen mypy complain that bool is not right in that case, maybe they've changed it or an old version doesn't catch it?

mensinda: I am running mypy 0.761 and Github actions also passes. Alternatively, we could get rid of all python 3.5 issues with #6438 :)

dcbaker: For now I guess I'll drop my complaint. I would really like to have that patch and be able to take advantage of 3.6 improvements. There's a lot of them :)

run_project_tests.py
mensinda: Even better, I removed all those functions and replaced them with a `do_not_pass` entry in `test.json`.

run_project_tests.py
```diff
@@ -160,17 +206,8 @@ def setup_commands(optbackend):
...
-
-def platform_fix_name(fname: str, compiler, env) -> str:
-    # canonicalize compiler
-    if (compiler in {'clang-cl', 'intel-cl'} or
-       (env.machines.host.is_windows() and compiler == 'pgi')):
-        canonical_compiler = 'msvc'
-    else:
-        canonical_compiler = compiler
-
+# TODO try to eliminate or at least reduce this function
```
jon-turney: Yeah, it's horrible :) I think more could be removed by adding 'lib' and 'so' installed files types, but that can certainly be done as a follow-up.

run_project_tests.py
```diff
@@ -145,6 +145,7 @@ def __init__(self, path: Path, name: T.Optional[str], args: T.List[str], skip: b
         self.skip = skip
         self.env = os.environ.copy()
         self.installed_files = []  # type: T.List[InstalledFile]
+        self.do_not_pass = []      # type: T.List[str]
```
jon-turney: Perhaps 'do_not_use_option' or somthing.  This name could cause confusion that the test is expected to fail.

mensinda: Done

mensinda:
I have updated the docs. My English is not the best, so there might be some errors.

run_project_tests.py
mensinda: Yeah, I don't want to do this in this PR. I might look into this later.

docs/markdown/Contributing.md
```diff
@@ -163,28 +163,136 @@ Any tests that require more thorough analysis, such as checking that certain
...
+      { "opt1": "bad"                    }
+    ]
+  }
+}
+```
+
+##### env
+
+The `env` key contains a dictionary which specifies additional
+environment variables to be set during the configure step of the test. `@ROOT@`
```
dcbaker: Maybe it would be better to use `@SOURCE_DIR@` like we do in the meson language? I don't know, I'm just thinking out loud.

jon-turney:
The markdown for the documentation table doesn't seem to be quite right: see https://mesonbuild.com/Contributing.html#installed

mensinda:
I am aware. There are currently some chages to it in #6911.

Issue #6294, opened 2019-12-05
gst-plugins-base fails on MacOS due to -fobjc-arc check


pbor:
gst-plugins-base fails on MacOS with the following error:

```
Compiler for Objective-C supports arguments -fobjc-arc: NO 

gst-libs/gst/gl/meson.build:683:4: ERROR: Problem encountered: ARC is required for building
```

This happens because of this check:

https://gitlab.freedesktop.org/boxerab/gst-plugins-bad/commit/38c3d4ca7b8e562e1396b902bb6dea07155ed3aa

Digging into the logs, the failure is:

```
Working directory:  /Users/jenkins/jhbuild/checkout/gst-plugins-base-1.16.2/build/meson-private/tmpicamerlo
Command line:  cc -L/Users/jenkins/DCV_JHBUILD/inst/lib /Users/jenkins/jhbuild/checkout/gst-plugins-base-1.16.2/build/meson-private/tmpicamerlo/testfile.m -o /Users/jenkins/jhbuild/checkout/gst-plugins-base-1.16.2/build/meson-private/tmpicamerlo/output.obj -pipe -c -O0 -Werror=unknown-warning-option -Werror=unused-command-line-argument -Werror=ignored-optimization-argument -fobjc-arc 

Code:
 int i;

Compiler stdout:
 
Compiler stderr:
 clang: error: argument unused during compilation: '-L/Users/jenkins/DCV_JHBUILD/inst/lib' [-Werror,-Wunused-command-line-argument]

Compiler for Objective-C supports arguments -fobjc-arc: NO 

gst-libs/gst/gl/meson.build:683:4: ERROR: Problem encountered: ARC is required for building

```

So, basically it seems to fail because there are LDFLAGS in the env (which are set by jhbuild in this case).

Should meson be able to sanitize the env for these kind of checks?

nacho:
I just had the same issue when compiling gst-plugins-base. If I look at the code from mesonbuild I see the following line:

`        myargs = ['-Werror=unknown-warning-option', '-Werror=unused-command-line-argument']`

https://github.com/mesonbuild/meson/blob/master/mesonbuild/compilers/mixins/clang.py#L64

If I change that line to look like this:

`        myargs = ['-Werror=unknown-warning-option']`

then the check works as expected. Thoughts?

nacho:
@nirbheek how are you folks getting this built on osx? I would have expected a fail in your side as well...

nirbheek:
[...]

PR #6668, opened 2020-02-20
macOS: Remove more unused linkerlike args

nirbheek:
`-L` and `-headerpad_max_install_names` are both linker arguments that are commonly passed in CFLAGS too.

Closes https://github.com/mesonbuild/meson/issues/6294

Also remove an unused variable in the unit tests.


nirbheek:
Ubuntu 16.04 failure seems unrelated.

PR #6672, opened 2020-02-21
docs: +1 line explaining why 'ninja test' does build_by_default [skip ci]

### LINKED ISSUE #1949: build_by_default:false is ignored by mesontest.
I have many targets, all flagged as "build_by_default: false". When I do ninja <target_name> it correctly builds only what is required for that target. But if I do mesontest <test_name> it seems to build everything.

marc-h38:
Fixes #1949

### DIFF
--- a/docs/markdown/Reference-manual.md
+++ b/docs/markdown/Reference-manual.md
@@ -311,7 +311,8 @@ following.
 
 - `build_by_default` *(added 0.38)* causes, when set to true, to
   have this target be built by default, that is, when invoking plain
-  `ninja`; the default value is false
+  `ninja` or `ninja test` or anything that depends on the default target.
+   The default value is false
   *(changed in 0.50)* if `build_by_default` is explicitly set to false, `install`
   will no longer override it. If `build_by_default` is not set, `install` will
   still determine its default.


docs/markdown/Reference-manual.md
```diff
@@ -311,7 +311,8 @@ following.
 
 - `build_by_default` *(added 0.38)* causes, when set to true, to
   have this target be built by default, that is, when invoking plain
-  `ninja`; the default value is false
+  `ninja` or `ninja test` or anything that depends on the default target.
+   The default value is false
```
nirbheek: This sentence is pretty hard to understand! How about:

```
- `build_by_default` *(added 0.38)* when set to `true`, this target will be built by default; 
  which means it will be automatically built when `ninja` is called without any arguments
  or when a target that depends on the [default target](https://ninja-build.org/manual.html#_default_target_statements) 
  `all` is specified, such as `ninja test`. The same behavior applies for other backends. 
  The default value is `false`.
```

marc-h38: Thanks @nirbheek for the feedback. I was trying to keep it as short as possible but if that's not a major concern then your text looks great to me. Considering you basically wrote it all would you like to submit this yourself in a new PR so `git blame` reflects reality? This would also save me taking credit for the "applies to other backends" bit that I know absolutely nothing about :-) If you're too busy and prefer me to do it that's fine too, up to you.

jon-turney: This describes the behaviour with ninja, then says "the same behavour applies for other backends", so perhaps the phrasing could be simplified to not mention the backend at all?

marc-h38: I like sample commands like `ninja test` because they get the point across clearly and briefly. This being said, maybe @nirbheek 's text can be rephrased and mention ninja fewer times.

jpakkane:
As there does not seem to be a consensus, I'm going to close this to reduce the clutter on the MR page. Feel free to reopen a new one with the improved wording. Thanks.

PR #6676, opened 2020-02-22
cmake2meson fix if nesting

ydirson:
cmake2meson would break with cmake statements like "if((A STREQUAL x) AND (NOT (B STREQUAL y)))" because it can't handle such nesting.  The "elseif" handling logic was even more simplistic, whereas it could reuse the if') logic.  This PR takes care of both problems.


jpakkane:
Thanks.

Issue #6669, opened 2020-02-21
Allow emscripten to be used as non cross compiler


Cloudef:
Emscripten can produce binaries that you can actually run. Flat out refusing to compile kinda sucks. Similarly I hope cross-compilation is also not forced on non native toolchains, since it's possible to run non-native binaries by using binfmt and qemu-static for example.

nirbheek:
How do you run binaries produced by emscripten? Is there an exe_wrapper or something? If so, what's wrong with cross-compiling with an exe_wrapper?

Cloudef:
Emscripten produces node scripts that has the node hashbang on them. In practice they work as any normal binary.

nirbheek:
Well, then you'll have to paste more information about what fails and how it fails. From the information you've given right now it should already work, or at least it seems like it should be easy to fix :)

I noticed that we also do not have any CI with this, so once this is fixed, ideally we'd want that to make sure that it keeps working and doesn't break.

Cloudef:
https://github.com/mesonbuild/meson/blob/1682058decb5f16ce1b78121ee587c024e9cfa17/mesonbuild/compilers/c.py#L140 what currently happens is this. Even if you remove these checks it will later fail at linker stuff. Removing the "emscripten detection" all together makes it work though.

nirbheek:
Sounds like you're well-placed to make a pull request? :)

Cloudef:
It looks like @jpakkane is the original author of the emscripten support. So before I do any PRs, I'd want to ask what would be the best approach to go here, as in to not break other people's build envs.

Cloudef:
@jpakkane 
I created a cross file with contents:
```
[properties]
needs_exe_wrapper=false
```
But trying to build pixman gives:
```
Build started at 2020-03-04T04:03:23.998316
Main binary: /usr/bin/python
Build Options: -Db_lto=true -Db_pie=true -Dloongson-mmi=disabled -Dvmx=disabled -Darm-simd=disabled -Dneon=disabled -Diwmmxt=disabled -Dmips-dspr2=disabled -Dgtk=disabled -Dprefix=/usr -Dlibexecdir=lib -Dsbindir=bin -Dauto_features=disabled -Dbuildtype=plain -Ddefault_library=static -Dwrap_mode=nodownload '--cross-file /etc/emscripten.toolchain.meson'
Python system: Linux
The Meson build system
Version: 0.53.2
[...]

PR #6678, opened 2020-02-23
Allow emscripten to be used as non cross-compiler

Cloudef:
https://github.com/mesonbuild/meson/issues/6669

### DIFF
--- a/mesonbuild/compilers/c.py
+++ b/mesonbuild/compilers/c.py
@@ -136,8 +136,6 @@ class AppleClangCCompiler(ClangCCompiler):
 class EmscriptenCCompiler(LinkerEnvVarsMixin, EmscriptenMixin, BasicLinkerIsCompilerMixin, ClangCCompiler):
     def __init__(self, exelist, version, for_machine: MachineChoice,
                  is_cross: bool, info: 'MachineInfo', exe_wrapper=None, **kwargs):
-        if not is_cross:
-            raise MesonException('Emscripten compiler can only be used for cross compilation.')
         ClangCCompiler.__init__(self, exelist=exelist, version=version,
                                 for_machine=for_machine, is_cross=is_cross,
                                 info=info, exe_wrapper=exe_wrapper, **kwargs)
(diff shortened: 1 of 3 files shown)


nirbheek:
Are you planning on adding CI too? Should be relatively easy to setup a GitHub action to test this. Makes it more likely that this PR will be merged (we know for sure that it works), and also ensures that the backend won't break in the future.

jpakkane:
There is a bit of a semantic issue here. In Meson native compilation means "targetting the current CPU + OS combo", which emscripten does not seem to do as it targets JS + WASM. Basically you should be able to have a target with one C file and one C++ file, set the C compiler to GCC, the C++ compiler to Emscripten, build the project and run it. Is it possible with this? It does not seem to be, since gcc produces elf files.

In Meson there is a way to specify that cross build executables are directly runnable, which is to set `needs_exe_wrapper` to `false`. It is used when "cross compiling" 32 bit binaries on 64 bit platforms. Can you test if it works for you (without this patch).

Cloudef:
> Basically you should be able to have a target with one C file and one C++ file, set the C compiler to GCC, the C++ compiler to Emscripten, build the project and run it. Is it possible with this? It does not seem to be, since gcc produces elf files.

This seems like a very specific use case... Which would be already possible by setting CC and CXX accordingly, however LD is gonna trip it out, but still solvable by either making the project support this case (since I don't see why such specific case would be wanted outside some project) or writing wrapper `ld` that picks the right linker.

> In Meson there is a way to specify that cross build executables are directly runnable, which is to set needs_exe_wrapper to false. It is used when "cross compiling" 32 bit binaries on 64 bit platforms. Can you test if it works for you (without this patch).

Do I have to write a cross file for that, or can I pass that one parameter from command line?

jpakkane:
You need a cross file.

jpakkane:
This MR probably won't be merged as is because it changes the semantics of what a "native build" is. That would need a fair bit more discussion and the issue raised here can be worked around with `needs_exe_wrapper=false`. If it does not work then that is a bug that needs to be fixed.

Thus closing to reduce clutter on the MR page. For further discussion please open a bug if necessary. Thanks.

PR #6685, opened 2020-02-24
azure ci: Test x86 Visual Studio builds again

nirbheek:
This was dropped when the VS2015 images were removed from Azure.


nirbheek:
Hah, a completely different test is failing on master...

jpakkane:
The Cygwin failure seems to be the same one as for all the others, so merging.

PR #6687, opened 2020-02-24
gnome.py: Determine g-ir-scanner and g-ir-compiler paths from pkgconfig

aduskett:
Because GOI must first run and scan a binary when creating .gir files, cross-compilation is quite tricky! Mainly because if the binary is cross-compiled, you would not typically be able to run or scan a cross-compiled binary from the host system. Because of this significant limitation, both Buildroot and Openembedded have decided to use several wrappers of which runs the native scanner using qemu to create the binaries. These wrappers live in the staging directory.

Currently, meson hard-codes the paths of these utilities, which results in cross-compiled environments running the host versions of these tools instead of the wrappers installed in the staging directory.

GObject-introspection provides the appropriate paths to these utilities via pkg-config, so it would be best to use the results of pkg-config instead of hard coding these paths.

### DIFF
--- a/mesonbuild/modules/gnome.py
+++ b/mesonbuild/modules/gnome.py
@@ -736,15 +736,30 @@ class GnomeModule(ExtensionModule):
         if kwargs.get('install_dir'):
             raise MesonException('install_dir is not supported with generate_gir(), see "install_dir_gir" and "install_dir_typelib"')
 
-        giscanner = self.interpreter.find_program_impl('g-ir-scanner')
-        gicompiler = self.interpreter.find_program_impl('g-ir-compiler')
-
         girtargets = [self._unwrap_gir_target(arg, state) for arg in args]
 
         if len(girtargets) > 1 and any([isinstance(el, build.Executable) for el in girtargets]):
             raise MesonException('generate_gir only accepts a single argument when one of the arguments is an executable')
 
         self.gir_dep, pkgargs = self._get_gir_dep(state)
+        # find_program is needed in the case g-i is built as subproject.
+        # In that case it uses override_find_program so the gobject utilities
+        # can be used from the build dir instead of from the system.
+        # However, GObject-introspection provides the appropriate paths to
+        # these utilities via pkg-config, so it would be best to use the
+        # results from pkg-config when possible.
+        gi_util_dirs_check = [state.environment.get_build_dir(), state.environment.get_source_dir()]
+        giscanner = self.interpreter.find_program_impl('g-ir-scanner')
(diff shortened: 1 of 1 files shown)


aduskett:
Just another note:

These paths have been in the .pc file since 1.30, which is as old as you can get for a stable gobject-introspection release.

mesonbuild/modules/gnome.py
```diff
@@ -736,15 +736,15 @@ def generate_gir(self, state, args, kwargs):
         if kwargs.get('install_dir'):
             raise MesonException('install_dir is not supported with generate_gir(), see "install_dir_gir" and "install_dir_typelib"')
 
-        giscanner = self.interpreter.find_program_impl('g-ir-scanner')
-        gicompiler = self.interpreter.find_program_impl('g-ir-compiler')
-
```
xclaesse: I think you should still keep that in the case it's not cross build, and fallback to use pkg-config.

aduskett: Is there a reason for this? Gobject-introspection has provided these paths since release, and they work not only on cross-building but also host-building.

aduskett: Also, I think you have it backward when cross-building, the above would still find host/usr/bin/g-ir-scanner, whereas using pkg-config set's the appropriate paths.

xclaesse: find_program is needed in the case g-i is built as subproject. In that case ou uses override_find_program so we can use them from the build dir instead of from system.

aduskett: Ah that makes sense, although this still leads to the problem above.

If giscanner = self.interpreter.find_program_impl('g-ir-scanner') is first, then the host directory will be found first in a cross-compiled environment and the pkg-config code will never get executed.

xclaesse: Right, I guess you need to check the path of the returned program, if it's within build dir or source dir then you should use it, otherwise use pkg-config.

aduskett: I updated the code, it's a bit long-winded but it should work for both scenarios.

aduskett: The first failure is due to a boost issue.

The second failure I am not sure why it's crashing. It's still using /usr/bin/g-ir-scanner, even though I check to see if the goi utils are in the source or build path. Any help would be appreciated!

xclaesse:
Looks good to me now, could you squash your commits please?

xclaesse:
LGTM, thanks.

PR #6690, opened 2020-02-25
mesonbuild: Add mcompile command

### LINKED ISSUE #6670: [Feature request] Add `meson compile` command
Proposal:
Backend-agnostic build command `meson compile` - counterpart of `cmake --build`.

Reasoning:
- Simplifies `meson` usage and integration: the same `meson` commands could be invoked for all backends.
- Reduces the amount of knowledge required to use `meson` out-of-the-box (i.e. with default `ninja` backend). Since user won't have to know anything about ninja (apart from the fact that it's needed).

Proposed interface (based on `cmake --build):
```
usage: meson compile [-h] [-C WD] [--clean] [--backend-args ARGS] [args [args ...]]

positional arguments:
  args                            optional list of targets to build

optional arguments:
-h, --help                      show this help message and exit
-C WD                          directory to cd into before running
--clean                          invokes clean command of the native tool.
--backend-args ARGS    arguments to pass to the native tool.
```

dcbaker:
This is tested working with both msbuild and ninja/samu. Since our xcode
support is pretty much broken I didn't bother.

Fixes #6670

### DIFF
--- a/docs/markdown/snippets/meson_compile_command.md
+++ b/docs/markdown/snippets/meson_compile_command.md
@@ -1,0 +1,21 @@
+## Backend agnostic compile command
+
+A new `meson compile` command has been added to support backend agnostic
+compilation. It accepts two arguments, `-j` and `-l`, which are used if
+possible (`-l` does nothing with msbuild). A `-j` or `-l` value < 1 lets the
+backend decide how many threads to use. For msbuild this means `-m`, for
+ninja it means passing no arguments.
+
+```console
+meson builddir --backend vs
+meson compile -C builddir -j0  # this is the same as `msbuild builddir/my.sln -m`
+```
+
+```console
+meson builddir
+meson compile -C builddir -j3  # this is the same as `ninja -C builddir -j3`
+```
+
+Additionally `meson compile` provides a `--clean` switch to clean the project.
+
+A complete list of arguments is always documented via `meson compile --help`
--- a/mesonbuild/mcompile.py
+++ b/mesonbuild/mcompile.py
@@ -1,0 +1,123 @@
+# Copyright 2020 The Meson development team
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Entrypoint script for backend agnostic compile."""
+
+import os
+import pathlib
+import shutil
+import sys
+import typing as T
+
+from . import mlog
+from . import mesonlib
+from .mesonlib import MesonException
+
+if T.TYPE_CHECKING:
+    import argparse
+
+
+def add_arguments(parser: 'argparse.ArgumentParser') -> None:
+    """Add compile specific arguments."""
+    parser.add_argument(
+        '-j', '--jobs',
+        action='store',
+        default=0,
+        type=int,
+        help='The number of worker jobs to run (if supported). If the value is less than 1 the build program will guess.'
+    )
+    parser.add_argument(
+        '-l', '--load-average',
+        action='store',
+        default=0,
+        type=int,
+        help='The system load average to try to maintain (if supported)'
+    )
+    parser.add_argument(
+        '--clean',
+        action='store_true',
+        help='Clean the build directory.'
+    )
+    parser.add_argument(
+        '-C',
+        action='store',
+        dest='builddir',
+        type=pathlib.Path,
+        required=True,
+        default='.',
+        help='The directory containing build files to be built.'
+    )
+
+
+def run(options: 'argparse.Namespace') -> int:
+    bdir = options.builddir  # type: pathlib.Path
+    if not bdir.exists():
+        raise MesonException('Path to builddir {} does not exist!'.format(str(bdir.resolve())))
+    if not bdir.is_dir():
+        raise MesonException('builddir path should be a directory.')
+
+    cmd = []  # type: T.List[str]
+    runner = None  # type T.Optional[str]
+    slns = list(bdir.glob('*.sln'))
+
+    if (bdir / 'build.ninja').exists():
+        runner = os.environ.get('NINJA')
+        if not runner:
+            if shutil.which('ninja'):
+                runner = 'ninja'
+            elif shutil.which('samu'):
+                runner = 'samu'
+
+        if runner is None:
+            raise MesonException('Cannot find either ninja or samu.')
+
+        cmd = [runner, '-C', bdir.as_posix()]
+
+        # If the value is set to < 1 then don't set anything, which let's
+        # ninja/samu decide what to do.
+        if options.jobs > 0:
+            cmd.extend(['-j', str(options.jobs)])
+        if options.load_average > 0:
+            cmd.extend(['-l', str(options.load_average)])
+        if options.clean:
+            cmd.append('clean')
+
+    # TODO: with python 3.8 this could be `elif slns := bdir.glob('*.sln'):`
+    elif slns:
+        assert len(slns) == 1, 'More than one solution in a project?'
+
+        sln = slns[0]
+        cmd = ['msbuild', str(sln.resolve())]
+
+        # In msbuild `-m` with no number means "detect cpus", the default is `-m1`
+        if options.jobs > 0:
+            cmd.append('-m{}'.format(options.jobs))
+        else:
+            cmd.append('-m')
+
+        if options.load_average:
+            mlog.warning('Msbuild does not have a load-average switch, ignoring.')
+        if options.clean:
+            cmd.extend(['/t:Clean'])
+
+    # TODO: xcode?
+    else:
+        raise MesonException(
+            'Could not find any runner or backend for directory {}'.format(bdir.resolve().as_posix()))
+
+    mlog.log('Found runner:', runner)
+
+    p, *_ = mesonlib.Popen_safe(cmd, stdout=sys.stdout.buffer, stderr=sys.stderr.buffer)
+
+    return p.returncode
--- a/run_unittests.py
+++ b/run_unittests.py
@@ -4412,6 +4412,21 @@ recommended as it is not supported on some platforms''')
         else:
             self.assertEqual(expected_lines, out_lines)
 
+    def test_meson_compile(self):
+        """Test the meson compile command."""
+        prog = 'trivialprog'
+        if is_windows():
+            prog = '{}.exe'.format(prog)
+
+        testdir = os.path.join(self.common_test_dir, '1 trivial')
+        self.init(testdir)
+        self._run([*self.meson_command, 'compile', '-C', self.builddir])
+        # If compile worked then we should get a program
+        self.assertPathExists(os.path.join(self.builddir, prog))
+
+        self._run([*self.meson_command, 'compile', '-C', self.builddir, '--clean'])
+        self.assertPathDoesNotExist(os.path.join(self.builddir, prog))
+
 
 class FailureTests(BasePlatformTests):
     '''
--- a/.github/workflows/lint_mypy.yml
+++ b/.github/workflows/lint_mypy.yml
@@ -30,4 +30,4 @@ jobs:
       with:
         python-version: '3.x'
     - run: python -m pip install mypy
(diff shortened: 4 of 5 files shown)


docs/markdown/snippets/meson_compile_command.md
```diff
@@ -0,0 +1,17 @@
...
+ninja it means passing no arguments.
+
+```console
+meson builddir --backend vs
+meson compile builddir -j0  # this is the same as `msbuild builddir/my.sln -m`
+```
+
+```console
+meson builddir -j 3
+meson compile builddir  # this is the same as `ninja -C builddir -j3`
```
xclaesse: -j3 is on wrong line

mesonbuild/mcompile.py
```diff
@@ -0,0 +1,127 @@
...
+        action='store_true',
+        help='Print debug information for the meson compile command.'
+    )
+    parser.add_argument(
+        'builddir',
+        type=pathlib.Path,
+        nargs='?',
+        default='.',
+        help='The directory containing build files to be built.'
+    )
```
xclaesse: I think it should be -C to be consistent with test and install commands

mesonbuild/mcompile.py
```diff
@@ -0,0 +1,127 @@
...
+        # If the value is set to < 1 then don't set anything, which let's
+        # ninja/samu decide what to do.
+        if options.jobs > 0:
+            cmd.extend(['-j', str(options.jobs)])
+        if options.load_average > 0:
+            cmd.extend(['-l', str(options.load_average)])
+
+    # TODO: with python 3.8 this could be `elif sln := bdir.glob('*.sln'):`
+    elif list(bdir.glob('*.sln')):
+        sln = list(bdir.glob('*.sln'))[0]
```
xclaesse: You glob it twice

xclaesse:
Would be nice to have a unit test too.

mesonbuild/mcompile.py
dcbaker: Without the python 3.8 walrus operator (the TODO comment) there isn't a really good way not to glob twice. We could make the VS code the first if, but I suspect most people use ninja, we could do a nested if:
```python
if <build.ninja>:
    ...
else:
    <glob>
    if glob:
        ...
    else:  # xcode
        ...
```
Or we could lose the elif, and have two if statements, but that would mean even if we've already decided to use ninja we still have to glob. I felt like just adding a TODO comment and doubling the glob was the cleanest solution.

mesonbuild/mcompile.py
dcbaker: Ah, I added that for meson test, but I always thought people disliked that syntax. In retrospect I dislike it too. So I can either add support to meson install and meson test to not use -C as well (ie, both -C or a positional would work, but not at the same time), or I can add it here, preferences?

docs/markdown/snippets/meson_compile_command.md
dcbaker: done

dcbaker:
I've added a unittest.

TheQwertiest:
I would argue that `--clean` argument should be added as well, since there is no other way to clean backend-agnosticly.

mesonbuild/mcompile.py
TheQwertiest: `-C` should remain for the compatibility sake, if nothing else (there are tools that use a positional instead as well, but -C usage is pretty widespread). I have no opinion regarding positional arg though.

dcbaker:
--clean added.

mesonbuild/mcompile.py
xclaesse: I'm not a big fan of -C, but current status is a mess:
- setup: positional
- configure: positional
- dist: -C
- install: -C 
- introspect: positional
- init: --builddir
- test: -C

The trend seems to be that old commands use positional but new commands use -C. I think -C is pretty common (e.g. git, ninja, make) and it has the advantage of being more explicit in the case your command has other positional arguments. `meson compile foo bar`, does it build the `bar` target in `foo` builddir or the opposite?

mesonbuild/mcompile.py
xclaesse: What about:
```python
def _get_compile_cmd():
    if <build.ninja>:
        return _get_ninja_cmd()
    <glob>
    if glob:
        return _get_sln_cmd(glob)
    if ...
def _run():
  cmd = _get_compile_cmd()
  return subprocess.run(cmd)

```

mesonbuild/mcompile.py
```diff
@@ -0,0 +1,136 @@
...
+        nargs='?',
+        default='.',
+        help='The directory containing build files to be built.'
+    )
+
+
+def _run(options: 'argparse.Namespace') -> int:
+    if options.debug:
+        mlog.log_file = sys.stderr
+        mlog.set_verbose()
```
xclaesse: Shouldn't we call mlog.initialize()?

mesonbuild/mcompile.py
```diff
@@ -0,0 +1,136 @@
...
+        if options.clean:
+            cmd.extend(['/t:Clean'])
+
+    # TODO: xcode?
+    else:
+        mlog.error('Could not find any runner or backend for directory {}'.format(bdir.resolve().as_posix()))
+        return 1
+
+    mlog.debug('Running command {}'.format(cmd))
+    p = subprocess.run(cmd)
```
xclaesse: Should we use mesonlib.Popen_safe() here?

mesonbuild/mcompile.py
dcbaker: I don't have a strong opinion, but I'll commit to make the all of the commands do the same thing (without remove the current behavior of course), so, opinions on what the "right" thing do do is? @xclaesse, @nirbheek, @jpakkane

mesonbuild/mcompile.py
dcbaker: I don't think so, we're not using a logfile in the sense of a file, but putting everything on stderr, and then letting the subprocess call write to our stdout and stderr. If we initialize it'll try to open `sys.stderr`, or we'll end up writing files in the build directory? We only need this to get options of lower status than warning anyway.

mesonbuild/mcompile.py
dcbaker: I guess we can, but I don't think it's necessary. We're not passing anything to stdin or capturing the output, we're just letting it be redirected to the terminal.

mesonbuild/mcompile.py
xclaesse: Hm, right, but since it's not logging into a real file, I'm not sure mlog.shutdown() will do. Any specific reason you want to print to stderr instead of stdout, btw? Ideally mlog.warning() should go to stderr but mlog.log() should go to stdout, I guess.

Also, verbose seems to be the default, I think you should call mlog.set_quiet() in else clause.

mesonbuild/mcompile.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+def _run(options: 'argparse.Namespace') -> int:
+    if options.debug:
+        mlog.log_file = sys.stderr
+        mlog.set_verbose()
+
+    bdir = options.builddir  # type: pathlib.Path
+    if not bdir.exists():
+        mlog.error('Path to builddir {} does not exist!'.format(bdir.as_posix()))
```
xclaesse: mlog.error() does not abort AFAIK. I think you should raise a MesonException instead, they are catched in mesonmain.py to print them as error.

mesonbuild/mcompile.py
```diff
@@ -0,0 +1,136 @@
...
+        if not runner:
+            if shutil.which('ninja'):
+                runner = 'ninja'
+            elif shutil.which('samu'):
+                runner = 'samu'
+            mlog.debug('found runner {}'.format(runner))
+
+        if runner is None:
+            mlog.error('Cannot find either ninja or samu.')
+            return 1
```
xclaesse: Shouldn't we standardize on raising a MesonException in case of error? I had similar questioning for minstall.py tbh, where it's a mix IIRC.

mesonbuild/mcompile.py
xclaesse: I don't remember what are the extra hacks we have in Popen_safe(), maybe it's not required here.

mesonbuild/mcompile.py
```diff
@@ -0,0 +1,136 @@
...
+    p = subprocess.run(cmd)
+
+    return p.returncode
+
+
+def run(options: 'argparse.Namespace') -> int:
+    try:
+        return _run(options)
+    finally:
+        mlog.shutdown()
```
xclaesse: shutdown is already done in mesonmain.py

run_unittests.py
```diff
@@ -4347,7 +4347,7 @@ def test_summary(self):
                          A list: string
                                  1
                                  True
-                     empty list: 
+                     empty list:
```
xclaesse: That's unrelated change, and I'm not sure it's correct. Was it failing the CI here?

dcbaker: No, I have my editor set up to trim trailing whitespace, so every time I save this it gets deleted. Is there a good reason that the summary code prints the whiltespace, or could we just remove it and not have it in the test either?

xclaesse: It's purely an implementation detail, feel free to fix it ;-)

mesonbuild/mcompile.py
dcbaker: Okay, I'll just use MesonException.

mesonbuild/mcompile.py
dcbaker: done

run_unittests.py
xclaesse: Did it there, while at it: https://github.com/mesonbuild/meson/pull/6692

mesonbuild/mcompile.py
dcbaker: I'll just use Popen_safe, so that the next person to look at this doesn't ask "why didn't they use popen_safe?" and I don't have to say "I don't remember, that was ... ago."

mesonbuild/mcompile.py
dcbaker: I don't think we need this at all, actually if we just let mesonmain handle it.

mesonbuild/mcompile.py
dcbaker: I noticed that there's a `Path.match()` function that is probably a better fit anyway, it's lazy (first match causes it to return true, so I used that instead for the elif

xclaesse: Ok, that's good enough. I know nothing about VS backend, but are we sure any .sln file is good, or should we let user tell which one to use somehow?

mesonbuild/mcompile.py
```diff
@@ -0,0 +1,124 @@
...
+        nargs='?',
+        default='.',
+        help='The directory containing build files to be built.'
+    )
+
+
+def run(options: 'argparse.Namespace') -> int:
+    bdir = options.builddir  # type: pathlib.Path
+    if not bdir.exists():
+        raise mesonlib.MesonException('Path to builddir {} does not exist!'.format(str(bdir.resolve())))
```
xclaesse: nitpicking: AFAIK, everywhere else in meson we do `from mesonlib import MesonException`.

mesonbuild/mcompile.py
```diff
@@ -0,0 +1,124 @@
...
+    if not bdir.is_dir():
+        raise mesonlib.MesonException('builddir path should be a directory.')
+
+    cmd = []  # type: T.List[str]
+    runner = None  # type T.Optional[str]
+    if (bdir / 'build.ninja').exists():
+        mlog.debug('Found build.ninja, using ninja.')
```
xclaesse: mlog.debug() only print into log_file, so this will have no effect. Should be mlog.log() I guess?

mesonbuild/mcompile.py
```diff
@@ -0,0 +1,124 @@
...
+        '-l', '--load-average',
+        action='store',
+        default=0,
+        type=int,
+        help='The system load average to try to maintain (if supported)'
+    )
+    parser.add_argument(
+        '--debug',
+        action='store_true',
+        help='Print debug information for the meson compile command.'
```
xclaesse: This option is not used any more.

dcbaker:
I removed all of the mlog.debug, and now just have one mlog for the runner selected. I also switched to using the -C option.

docs/markdown/snippets/meson_compile_command.md
```diff
@@ -0,0 +1,17 @@
...
+ninja it means passing no arguments.
+
+```console
+meson builddir --backend vs
+meson compile builddir -j0  # this is the same as `msbuild builddir/my.sln -m`
+```
+
+```console
+meson builddir
+meson compile builddir -j3  # this is the same as `ninja -C builddir -j3`
```
xclaesse: Doc needs to be  updated to use -C

mesonbuild/mcompile.py
xclaesse: Did you forget to fix it, or do you think it's better like this? tbh I don't have strong opinion, as you want ;-)

jpakkane:
As new functionality this is ok, go ahead and merge once all the details have been worked out.

mesonbuild/mcompile.py
dcbaker: I probably forgot. I need to update the docs anyway, so I'll send another version.

xclaesse:
Looks all good to me, LGTM once CI is green. Thanks for your work!

docs/markdown/snippets/meson_compile_command.md
```diff
@@ -0,0 +1,17 @@
+## Backend agnostic compile command
+
```
TheQwertiest: Missing `--clean` mention.

TheQwertiest:
Note: CI failures are relevant.

xclaesse:
As followup PR I just realised we could use this in must.py instead of forcing Ninja.

PR #6692, opened 2020-02-25
Summary improvements

xclaesse:
How this looks like with gst-build. Note that it is a 100% automatic summary section, no changes are needed in gst-build to produce that summary.

summary

### DIFF
--- a/docs/markdown/Reference-manual.md
+++ b/docs/markdown/Reference-manual.md
@@ -1239,9 +1239,10 @@ pair doesn't appear twice. All sections will be collected and printed at
 the end of the configuration in the same order as they have been called.
 
 Keyword arguments:
+- `section` title to group a set of key/value pairs.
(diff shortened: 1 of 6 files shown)


tp-m:
Nitpick, but have you also tried how it looks if you column-align the comment after the `YES|NO` ?

```
Foo: YES  2 warnings
Bar: NO   Feature 'bar' disabled
```

xclaesse:
Not sure it makes much difference:
summary

Also as an implementation PoV I did that screenshot by using `"NO "` string, but it's not something generally easy to provide alignment of values.

QuLogic:
If all you have is YES/NO, then the alignment seems a no-brainer. Sure it makes little difference here, but consider a run of alternating YES/NO with messages. The zig-zag would be quite annoying. If there are other outputs though, then that's a little different trade-off.

PR #6697, opened 2020-02-26
compilers/mixins/clike.py: prefer .a shlibext when --default-library=static

aduskett:
The existing library search order makes sense for environments with shared or a
mix of shared and static libraries. However, in a cross-compile environment
with a prebuilt cross-toolchain, both static and shared versions of the sysroot
libraries.

This environment presents a problem when a user builds an entirely static build
where there won't be any shared libraries at runtime because ld throws
the error: "ld: attempted static link of a dynamic object."

### DIFF
--- a/mesonbuild/compilers/mixins/clike.py
+++ b/mesonbuild/compilers/mixins/clike.py
@@ -857,6 +857,9 @@ class CLikeCompiler:
         elif env.machines[self.for_machine].is_cygwin():
             shlibext = ['dll', 'dll.a']
             prefixes = ['cyg'] + prefixes
+        elif env.coredata.get_builtin_option('default_library') == 'static':
+            # Linux/BSDs
+            shlibext = ['a']
         else:
             # Linux/BSDs
             shlibext = ['so']


xclaesse:
I don't think this is correct, and I'm surprised the CI pass, we are missing tests for this area it seems...

It is conceptually wrong because default_library=static means you want to build static libraries, but not that you want to depend on static libraries. So at least it must be a new separate option.

With this patch `cc.find_library('foo', static : false)` will return the static library when you have both `.so` and `.a` libraries available. That's not what the user requested, and could also fail to link because it will be missing dependencies of library foo. And if libfoo.a is not found then is abort instead of using the .so as requested.

To be able to think about a proper solution, I would need a precise use-case. Would be nice to provide a minimal test project that shows the issue, so we can include it in our test suite too.

aduskett:
Sure, I believe http://autobuild.buildroot.net/results/db1740b4777f436324218c52bc7b08e5c21b667d is what caused the creation of this patch.

I can create a simple docker buildroot test if you are willing to take a look, it would be much appreciated!

xclaesse:
I mean a stripped down meson.build that shows the issue. Like those we have in `test cases/common/`.

aduskett:
Ah, I can give it a try.

yann-morin-1998:
@xclaesse

> It is conceptually wrong because default_library=static means you want to build static libraries, but not that you want to depend on static libraries. So at least it must be a new separate option.
[snip]
> To be able to think about a proper solution, I would need a precise use-case. Would be nice to provide a minimal test project that shows the issue, so we can include it in our test suite too.

The purpose of that patch is exactly to never ever link with shared libs, even if present.

Buildroot is a (meta-)buildsystem for embedded devices. It grabs the sources of packages, 
builds a toolchain (or downloads a prebuilt one), and configures+builds packages. In some cases, we may want to only link with static libs.

So, this is not merely about providing a package with a meson.build that wants to link static libs, it is more about the builder above the package instructung it to only link to static libs (and also only generate static libs, the original purpose of that option).

So in Buildroot we came up with this solution: if we want to generate only static libs, it's also because we only want to link with static libs. Hence this patch.

Now, I agree with you, this is a shortcut for us, that illustrates our needs. If you the upstream want it implemented otherwise (a separate option?) is OK for us too; we'd adapt.

(unless I misunderstood your question...)

xclaesse:
Thanks @yann-morin-1998, I agree this is a quick work around that works for your specific use-case, but it's not suited for upstream since it clash with other valid use-cases.

When both shared and static libraries are available for a dependency, and `static` keyword argument is not provided, we implicitly prefer shared libraries. We could add a new option `-Dprefer_static=true` (naming is always hard) to instead prefer static. See below for considerations about pkg-config case. In find_library() case it won't always work because you need to explicitly add dependencies of your dependency when using static libraries.

When only static is available (or `-Dprefer_static=true`), and static` keyword argument is not provided, pkg-config should re-run itself with --static option: See https://github.com/mesonbuild/meson/pull/6629 for some considerations about that case.

yann-morin-1998:
> We could add a new option `-Dprefer_static=true` (naming is always hard) to instead prefer static.

Yeah, naming is hard ;-) but in this case, 'prefer' is not strong enough: we want to have an option that says "only use static, never shared". (note: conversely, we'd also want an option "only use shared, never static"). 'prefer' implies that if static is not usable, then that'd be OK to use shared, which in our case isn't. A good option would probably be: `-Duse_libs=ordered-list`, like:
* `-Duse_libs=shared` use shared only, fail if missing
* `-Duse_libs=shared,static` use shared if available, fallback to static, fail if both missing  _(the default value, to keep the current behaviour?)_
* `-Duse_libs=static` use static only, fail if missing
* `-Duse_libs=static,shared` use static if available, fallback to shared, fail if both missing

That obviously does not solve the problem of inherited dependencies (but pkg-config is supposed to provide them in the `Requires` and `Requires.private` fields; I'm not sure how that would be handled in `find_library()` though, but you're the experts!).

aduskett:
Yes, I think the problem here is that we want an upstreamable way to force meson to fail if a static library isn't present, even if a shared library is present.

xclaesse:
Ok, I get it now, so you want a force-static rather than a prefer-static. I guess the `use_libs` (again, naming stuff...) you described is the best option. That seems like a can of worms though.

> That obviously does not solve the problem of inherited dependencies (but pkg-config is supposed to provide them in the `Requires` and `Requires.private` fields; I'm not sure how that would be handled in `find_library()` though, but you're the experts!).

Spoiler alert, it won't work with find_library() and even with pkg-config [it's tricky](https://github.com/mesonbuild/meson/pull/6629#issuecomment-589812351).

aduskett:
How about the option: -Donly_shared? Would that be accepted?

PR #6700, opened 2020-02-26
Python usage fix for tests from Stefano Rivera.

jpakkane:
50% of the fixes in https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=952610.

Let's see if this works in master.


dcbaker:
This is working on my gentoo system: https://github.com/mesonbuild/meson/pull/6703

PR #6703, opened 2020-02-26
Fix python2 rename

dcbaker:
The old guidance from the python foundation was that "python" should point to 2.x, and that when 2.x EOLs that "python" should go away, that guidance has changed to "python" should now point to either "python2" or "python3". Given that we need to update some tests. This is complicated by the fact that macOS hasn't been updated, it still only has a "python" binary that is python2

I've also tried to update some of the "do you have python2" checks to be more generic.


dcbaker:
@jpakkane, I think this is ready to merge, and should superceed both #6700 and #6656

PR #6706, opened 2020-02-27
Fix 'linuxlike/14 static dynamic linkage' on Cygwin

jon-turney:
Update the expected output from nm to match changes in binutils 2.34.

### DIFF
--- a/test cases/linuxlike/14 static dynamic linkage/verify_static.py
+++ b/test cases/linuxlike/14 static dynamic linkage/verify_static.py
@@ -13,7 +13,7 @@ def handle_common(path):
 def handle_cygwin(path):
     """Handle the Cygwin case."""
     output = subprocess.check_output(['nm', path]).decode('utf-8')
-    if 'I __imp_zlibVersion' in output:
+    if (('I __imp_zlibVersion' in output) or ('D __imp_zlibVersion' in output)):
         return 1
     return 0
 


test cases/linuxlike/14 static dynamic linkage/verify_static.py
```diff
@@ -13,7 +13,8 @@ def handle_common(path):
 def handle_cygwin(path):
     """Handle the Cygwin case."""
     output = subprocess.check_output(['nm', path]).decode('utf-8')
-    if 'I __imp_zlibVersion' in output:
+    if (('I __imp_zlibVersion' in output) or
+       ('D __imp_zlibVersion' in output)):
```
dcbaker:
```suggestion
    if (('I __imp_zlibVersion' in output) or
        ('D __imp_zlibVersion' in output)):
```

The alignment looks funny here.

jon-turney: Ooops, yes.  That's what I get for not using the right editor :)

jon-turney:
> Thanks for fixing this

No problem.  

Please feel free to ping me for prompt attention when something has changed in cygwin that is causing a breakage, otherwise I may not notice until the next time I try to use master.

jon-turney:
> Update the expected output from nm to match changes in binutils 2.34.

FTR, this change is a regression in binutils, affecting 2.33 and 2.34. See:

https://sourceware.org/ml/binutils/2020-02/msg00553.html
https://sourceware.org/bugzilla/show_bug.cgi?id=24511

PR #6712, opened 2020-03-01
Implementation template factory and add Java 'meson.build' for Java developers.

michaelbadcrumble:
I have implemented a Meson template factory so we can add more programming languages if needed. 
Any new programming languages must implement `create_library(self)` and `create_executable(self)` methods.

I also added a `meson.build` template for Java projects.


michaelbadcrumble:
This PR is now split into there own PRs

#6724 
#6723 
#6722

PR #6713, opened 2020-03-01
Clean 'meson dist' command

michaelbadcrumble:
Changed Python2 %s syntax to Python3 format and added comments.

### DIFF
--- a/mesonbuild/mdist.py
+++ b/mesonbuild/mdist.py
@@ -102,9 +102,6 @@ def run_dist_scripts(src_root, bld_root, dist_root, dist_scripts):
             sys.exit(1)
 
 def is_git(src_root):
-    '''
-    Check to see if we are in a 'git' project
-    '''
     _git = os.path.join(src_root, '.git')
     return os.path.isdir(_git) or os.path.isfile(_git)
 
@@ -126,9 +123,6 @@ def git_clone(src_root, distdir):
 
 
 def create_dist_git(dist_name, archives, src_root, bld_root, dist_sub, dist_scripts, subprojects):
-    '''
-    Create a Meson dist release using Git
-    '''
     distdir = os.path.join(dist_sub, dist_name)
     git_clone(src_root, distdir)
     for path in subprojects:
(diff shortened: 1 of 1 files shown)


mesonbuild/mdist.py
```diff
@@ -33,6 +33,10 @@
                      'zip': '.zip'}
 
 def add_arguments(parser):
+    '''
+      here we are adding arguments that can be used with
```
nirbheek: You don't need to start each comment with "here we are", because it's redundant.

mesonbuild/mdist.py
```diff
@@ -233,7 +285,11 @@ def determine_archives_to_generate(options):
         sys.exit('No archive types specified.')
     return result
 
+
 def run(options):
+    '''
+    this is a runner function for 'meson dist'
```
nirbheek: You don't need to say "this is a", just say `runner function for 'meson dist'`. Same for all the rest.

michaelbadcrumble: Sorry, my brain was on autopilot, will fix this.

mesonbuild/mdist.py
michaelbadcrumble: Problem fixed.

jpakkane:
Thanks, but I'm going to have to be a slight contrarian here w.r.t. the comments. See for example this:

```
def is_hg(src_root):
    '''
    Check to see if its a 'hg' project.
    '''
```

This comment provides no extra information, it is simply saying the same thing as the function name. Thus it is visual noise that should be removed. If one wants to clarify that this is a "project" then the better approach would be to rename the function to `is_hg_project` instead. (I don't think it is necessary in this case.)

The same goes for the other added comment blocks.

michaelbadcrumble:
Needless comments removed.

nirbheek:
Did you close this PR on purpose?

michaelbadcrumble:
I wasn’t particularly happy with that PR.  So I created #6725 since I originally wanted to cherry-pick Python2 syntax.  Also, note-to-self no PR should be created late at night or without having a cup coffee first.

PR #6714, opened 2020-03-01
Using python3 print format to python2 %

michaelbadcrumble:
Doing some cleaning in the install command, simply switch Python2 % for Python3 format.

### DIFF
--- a/mesonbuild/minstall.py
+++ b/mesonbuild/minstall.py
@@ -248,7 +248,7 @@ class Installer:
                 raise RuntimeError('Destination {!r} already exists and is not '
                                    'a file'.format(to_file))
             if self.should_preserve_existing_file(from_file, to_file):
-                append_to_log(self.lf, '# Preserving old file %s\n' % to_file)
+                append_to_log(self.lf, '# Preserving old file {}\n'.format(to_file))
                 self.preserved_file_count += 1
                 return False
             os.remove(to_file)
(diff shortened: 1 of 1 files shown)


mesonbuild/minstall.py
```diff
@@ -181,6 +183,9 @@ def restore_selinux_contexts():
 
 
 def get_destdir_path(d, path):
+    '''
+    here we dist directory path
```
nirbheek: What does this comment mean?

mesonbuild/minstall.py
```diff
@@ -76,7 +76,9 @@ def __exit__(self, exception_type, value, traceback):
             append_to_log(self.lf, d)
 
 def is_executable(path, follow_symlinks=False):
-    '''Checks whether any of the "x" bits are set in the source file mode.'''
+    '''
+    Here checks whether any of the "x" bits are set in the source file mode.
```
QuLogic: Do not add "Here" to the front; in fact it should be imperative tense, see [PEP257](https://www.python.org/dev/peps/pep-0257/).

mesonbuild/minstall.py
michaelbadcrumble: This comment was a mistake and is removed.

mesonbuild/minstall.py
michaelbadcrumble: Fixed.

michaelbadcrumble: Thanks for the reference.

nirbheek:
Did you close this PR on purpose?

michaelbadcrumble:
I wasn’t particularly happy with that PR.  So I created #6725 since I originally wanted to cherry-pick Python2 syntax.  Also, note-to-self no PR should be created late at night or without having a cup coffee first.

PR #6716, opened 2020-03-01
Adding -C to 'meson init' command

michaelbadcrumble:
Provided a `-C` argument so users can specify the root directory.


xclaesse:
Seems reasonable to me. LGTM.

PR #6717, opened 2020-03-02
Document the details of optimization+debug vs buildtype

nirbheek:
Otherwise you have to hunt through the source code. Specifically, this is defined in `mesonbuild/coredata.py`: `set_buildtype_from_others()` and `set_others_from_buildtype()`

### DIFF
--- a/docs/markdown/Builtin-options.md
+++ b/docs/markdown/Builtin-options.md
@@ -87,6 +87,23 @@ Using the option as-is with no prefix affects all machines. For example:
 | werror                               | false         | Treat warnings as errors                                       | no             |
 | wrap_mode {default, nofallback,<br>nodownload, forcefallback} | default | Wrap mode to use                            | no             |
 
+For setting optimization levels and toggling debug, you can either set the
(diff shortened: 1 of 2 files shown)


TheQwertiest:
Test failures are relevant: you have to somehow limit the scope of the section data in `test_builtin_options_documented`.
Current parsing code assumes that the section contains only a single table.
The relevant code is probably this:
```
# Extract the content for this subsection
sub_end = sub2.start() if sub2 is not None else len(content)
subcontent = content[sub1.end():sub_end]
# Find the list entries
arches = [m.group(1) for m in re.finditer(r"^\| (\w+) .* \|", subcontent, re.MULTILINE)]
```

nirbheek:
I rewrote the tests. They were actually completely broken. Fixed now.

TheQwertiest:
> They were actually completely broken.

Sorry for being a Python amateur  \<(．＿．)\>

nirbheek:
> Sorry for being a Python amateur <(．＿．)>

No, your analysis was correct 😄. I was talking about the outer code which had some broken assumptions and had to be rewritten.

PR #6718, opened 2020-03-02
coredata: init_builtins should always call libdir_cross_fixup

dabrain34:
As the interpreter reinitialize the coredate builtins, need to call
again libdir_cross_fixup


jpakkane:
This MR caused a regression and might get reverted. See #7076.

dcbaker:
would it be possible with my site-config series to just drop the fixup all together and just ship a site-config file for debian?

jpakkane:
The Debian fix needs to happen now, and I don't want it to have any out-of-tree patches. So I'll merge the revert. It has a regression test so we know later fixes won't break.

PR #6721, opened 2020-03-02
pkgconfig: Fix deprecation message (fixes #6720)

### LINKED ISSUE #6720: Unable to cross build glib
[my-meson-cross-file.txt](https://github.com/mesonbuild/meson/files/4275963/my-meson-cross-file.txt)
**Describe the bug**
The configure step fails during the configuration of glib:

```
  File "/home/scerveau/DEV/EXTERNAL/meson/mesonbuild/modules/pkgconfig.py", line 71, in _check_generated_pc_deprecation
    location=obj.generated_pc_warn[1])
  File "/home/scerveau/DEV/EXTERNAL/meson/mesonbuild/mlog.py", line 279, in deprecation
    return _log_error('deprecation', *args, **kwargs, is_error=True)
  File "/home/scerveau/DEV/EXTERNAL/meson/mesonbuild/mlog.py", line 253, in _log_error
    if location is not None:
AttributeError: 'types.SimpleNamespace' object has no attribute 'filename'
```

**To Reproduce**
Checkout gst-build and try to cross build with the given cross-file attached
[my-meson-cross-file.txt](https://github.com/mesonbuild/meson/files/4275967/my-meson-cross-file.txt)

**Expected behavior**

To finish the configuration step without failing

**system parameters**
* Is this a [cross build](https://mesonbuild.com/Cross-compilation.html) or just a plain native build (for the same computer)? yes

Ubuntu 18.04

3.6.9
* what `meson --version`
master b190a20a
1.8.2

dabrain34:
Deprecation warning was unable to display because of patch
c8f8d58273a40d94c820dccab54a7ae2d948cb8a

### DIFF
--- a/mesonbuild/modules/pkgconfig.py
+++ b/mesonbuild/modules/pkgconfig.py
@@ -526,8 +526,7 @@ class PkgConfigModule(ExtensionModule):
             for lib in deps.pub_libs:
                 if not isinstance(lib, str) and not hasattr(lib, 'generated_pc'):
                     lib.generated_pc = filebase
-                    location = types.SimpleNamespace(subdir=state.subdir,
-                                                     lineno=state.current_lineno)
+                    location = state.current_node
                     lib.generated_pc_warn = [name, location]
         return ModuleReturnValue(res, [res])
 


mesonbuild/modules/pkgconfig.py
```diff
@@ -526,7 +526,7 @@ def parse_variable_list(stringlist):
             for lib in deps.pub_libs:
                 if not isinstance(lib, str) and not hasattr(lib, 'generated_pc'):
                     lib.generated_pc = filebase
-                    location = types.SimpleNamespace(subdir=state.subdir,
+                    location = types.SimpleNamespace(filename=state.subdir,
                                                      lineno=state.current_lineno)
```
xclaesse: Seems weird, I guess filename should be a file and not a dir. If I understand correctly it should be `filename = os.path.join(state.subdir, environment.build_filename)`, no?

xclaesse: @jon-turney ^

jon-turney: Yes.  It seems like I missed an instance of this, despite checking several time.

jon-turney: Perhaps the ModuleState passed to a module snippet should have a location member we would use here instead, as a more general solution.

Not being caught earlier also indicates that this warning doesn't have any test coverage.

dabrain34: I added the `build_filename` to the `ModuleState` but I was wondering if the build_filename can be different from `meson.build`

mesonbuild/modules/pkgconfig.py
```diff
@@ -526,7 +526,7 @@ def parse_variable_list(stringlist):
             for lib in deps.pub_libs:
                 if not isinstance(lib, str) and not hasattr(lib, 'generated_pc'):
                     lib.generated_pc = filebase
-                    location = types.SimpleNamespace(subdir=state.subdir,
+                    location = types.SimpleNamespace(filename=os.path.join(state.subdir, state.build_filename),
```
jon-turney: Oh, if we already have the relevant node, this can just be `location = self.current_node`

dabrain34: I'm not sure to follow you, `current_node` is not defined

jon-turney: Bah, I mean `location = state.current_node` it's an attribute of the ModuleState object.

dabrain34: even better indeed.

jon-turney:
> Not being caught earlier also indicates that this warning doesn't have any test coverage

I attempted to write a `meson-warning` test for this warning, but looking at the history of this warning didn't give any simple examples.

PR #6722, opened 2020-03-02
Implement sample factory

michaelbadcrumble:
I have implemented a Meson template factory so we can add more programming languages if needed. 
Any new programming language sample generators must implement `create_library(self)` and `create_executable(self)` methods found in `SampleImpl`.


mesonbuild/templates/samplefactory.py
```diff
@@ -0,0 +1,50 @@
...
+        elif options.language == 'java':
+            return JavaProject(options)
+        elif options.language == 'd':
+            return DlangProject(options)
+        elif options.language == 'rust':
+            return RustProject(options)
+        elif options.language == 'cuda':
+            return CudaProject(options)
+        elif options.language == 'fortran':
+            return FortranProject(options)
```
Akaricchi: I'd recommend something like this instead
```python
return {
    'c': CProject,
    'cpp': CppProject,
    .....
}[options.language](options)
```

michaelbadcrumble: Like the suggestion.  I never have seen `[options.language](options)` syntax before.  Is it compatible with Python 3.5?

Akaricchi: It's not any special syntax. You're just instantiating a dict object `{...}`, then indexing it `[...]`, which returns one of the class objects, which you then call/instantiate `(...)`. Ought to be working for as long as Python had dict literals. You can rewrite it like that for clarity:

```python
classes = {
    'c': CProject,
    'cpp': CppProject,
    .....
}
proj_class = classes[options.language]
proj = proj_class(options)
return proj
```

michaelbadcrumble: I will see if I can get that to work otherwise thanks for the suggestion.

michaelbadcrumble:
Ping @jpakkane and or @dcbaker .

mesonbuild/minit.py
```diff
@@ -24,35 +24,23 @@
...
 FORTRAN_SUFFIXES = {'.f', '.for', '.F', '.f90', '.F90'}
 LANG_SUFFIXES = {'.c', '.cc', '.cpp', '.cs', '.cu', '.d', '.m', '.mm', '.rs', '.java'} | FORTRAN_SUFFIXES
 LANG_SUPPORTED = {'c', 'cpp', 'cs', 'cuda', 'd', 'fortran', 'java', 'rust', 'objc', 'objcpp'}
 
-DEFAULT_PROJECT = 'executable'
-DEFAULT_VERSION = '0.1'
+class DEFAULT_VALUES(Enum):
+    PROJECT = 'executable'
+    VERSION = '0.1'
+
```
dcbaker: This is a bad place to use an enum, these aren't homogenous types.

mesonbuild/templates/samplefactory.py
```diff
@@ -0,0 +1,40 @@
...
+from mesonbuild.templates.rusttemplates import RustProject
+from mesonbuild.templates.javatemplates import JavaProject
+from mesonbuild.templates.cudatemplates import CudaProject
+from mesonbuild.templates.objctemplates import ObjCProject
+from mesonbuild.templates.cpptemplates import CppProject
+from mesonbuild.templates.cstemplates import CSharpProject
+from mesonbuild.templates.ctemplates import CProject
+
+
+class SampleFactory(object):
```
dcbaker: Don't use `object` here, that's a python 2.x requirement and is just legacy baggage at this point.

mesonbuild/templates/samplefactory.py
```diff
@@ -0,0 +1,40 @@
...
+            'cpp': CppProject,
+            'cs': CSharpProject,
+            'cuda': CudaProject,
+            'objc': ObjCProject,
+            'objcpp': ObjCppProject,
+            'java': JavaProject,
+            'd': DlangProject,
+            'rust': RustProject,
+            'fortran': FortranProject
+        }[options.language](options)
```
dcbaker: There is absolutely no reason for this to be a class, just take the static function and make it a top level function.

mesonbuild/minit.py
michaelbadcrumble: Ok, #6762 should have all of the issues fixed.

michaelbadcrumble:
Ok, #6762 should have all of the issues fixed.  So closing this PR to reduce clutter.

PR #6725, opened 2020-03-02
Cherry-picking - (rm python2 % add python3 .format)

michaelbadcrumble:
Cherry-picking Python2 % syntax from Meson commands.

### DIFF
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -963,8 +963,7 @@ This will become a hard error in a future Meson release.''')
         if self.gnu_symbol_visibility != '':
             permitted = ['default', 'internal', 'hidden', 'protected', 'inlineshidden']
             if self.gnu_symbol_visibility not in permitted:
-                raise InvalidArguments('GNU symbol visibility arg {} not one of: {}',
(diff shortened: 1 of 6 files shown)


mesonbuild/build.py
```diff
@@ -964,7 +963,7 @@ def process_kwargs(self, kwargs, environment):
         if self.gnu_symbol_visibility != '':
             permitted = ['default', 'internal', 'hidden', 'protected', 'inlineshidden']
             if self.gnu_symbol_visibility not in permitted:
-                raise InvalidArguments('GNU symbol visibility arg %s not one of: %s',
+                raise InvalidArguments('GNU symbol visibility arg {} not one of: {}',
                                        self.symbol_visibility, ', '.join(permitted))
```
jpakkane: This is missing `.format`. TBF the original was broken as well..

michaelbadcrumble: Added the missing `.format`.

michaelbadcrumble: 😕 So.  Should there be a test to catch this?

jpakkane: Given the way Python works, it's not really feasible.

Issue #1889, opened 2017-06-03
subprojects always inherit c_std and cpp_std from the parent project


Timmmm:
If you have a main project that uses c11 for example, and then a subproject needs gnu99 for example, the subproject will actually be compiled with c11.

I'm not sure meson pays any attention to language versions in subprojects.

jpakkane:
The language version is a global option that is shared by every target. There are languages (such as C++) where combining different language versions fails.

Timmmm:
But it only fails in some fairly limited cases. I feel like it should be respected but you can print a warning something like:

    Warning: Mixing C++ versions c++11 and gnu++14 may cause hard-to-find bugs.

If you really don't want to allow this then at the very least it should print a warning like this:

    Warning: Subproject has different C++ version. Language versions are global options shared between all targets. Ignoring 'gnu++14' and using 'c++11'.

germandiagogomez:
I am having exactly the same problem but with another variation.

I have a project, the default_options is c++14 or c++1z depending on wether my build is for ios or desktop due to this bug: https://bugreports.qt.io/browse/QTBUG-63747

I do something like this (@jpakkane correct me if I am wrong in the way to approach this):

```python
project('myproject', 'cpp', version : '0.5')


if host_machine.system() == 'darwin' and meson.is_cross_build()
  # Workaround exactly this bug: https://bugreports.qt.io/browse/QTBUG-63747
  add_project_arguments(['-std=c++14'], language : 'cpp')
  add_project_link_arguments(['-std=c++14'], language : 'cpp')
else
  add_project_arguments(['-std=c++1z'], language : 'cpp')
  add_project_link_arguments(['-std=c++1z'], language : 'cpp')
endif
```

So the net result of this is that now, due to project not having default options (since it is conditional), my subproject gets no c++11/14 flag and cannot be compiled. Can I define default_options conditionally? I tried but meson complained that project must be the first thing in the project.

nirbheek:
Setting aside cpp_std, I think c_std should never be inherited by subprojects.

dpirch:
I think it is ok that config options (build type, warning options...) are inherited from the parent project. The problem here is that **c_std shouldn't be a config option at all**.

It does not make sense that the user can set the C standard version via *meson configure* - it is not a  preference by the user who is compiling the project, but determined by authors of the project, similar to dependencies.

[...]

PR #6729, opened 2020-03-03
WIP: Make compiler options per-subproject

xclaesse:
This is unfinished work, as reference because that's a popular request. I'll try to continue working on that when possible.

### DIFF
--- a/test cases/common/229 persubproject options/meson.build
+++ b/test cases/common/229 persubproject options/meson.build
@@ -1,10 +1,24 @@
-project('persubproject options', 'c', default_options : ['default_library=both'])
+project('persubproject options', 'c', 'cpp',
+  default_options : [
+    'default_library=both',
+    'c_args=-DPARENT',
+    'cpp_std=c++98'
+  ]
+)
 
 assert(get_option('default_library') == 'both', 'Parent default_library should be "both"')
+assert(get_option('c_args') == ['-DPARENT'], 'Parent c_args are wrong')
 
 # Check it build both by calling a method only both_libraries target implement
 lib = library('lib1', 'foo.c')
 lib.get_static_lib()
 
-subproject('sub1')
-subproject('sub2', default_options : ['default_library=static'])
+sub1 = subproject('sub1')
+sub2 = subproject('sub2',
+  default_options : [
+    'default_library=static',
+    'c_args=-DSUB2_FROM_PARENT',
+  ]
+)
+
+test('mixing-cpp-version', executable('test1', 'main.cpp', link_with : sub2.get_variable('libcpp11')))
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -167,12 +167,14 @@ class Backend:
 
     def get_base_options_for_target(self, target):
         return OptionOverrideProxy(target.option_overrides,
+                                   target.subproject,
                                    self.environment.coredata.builtins,
                                    self.environment.coredata.base_options)
 
     def get_compiler_options_for_target(self, target):
         return OptionOverrideProxy(
             target.option_overrides,
+            target.subproject,
             self.environment.coredata.compiler_options[target.for_machine])
 
     def get_option_for_target(self, option_name, target):
@@ -615,7 +617,7 @@ class Backend:
         # Compile args added from the env: CFLAGS/CXXFLAGS, etc, or the cross
         # file. We want these to override all the defaults, but not the
         # per-target compile args.
-        commands += self.environment.coredata.get_external_args(target.for_machine, compiler.get_language())
+        commands += copt_proxy[compiler.get_language() + '_args'].value
         # Always set -fPIC for shared libraries
         if isinstance(target, build.SharedLibrary):
             commands += compiler.get_pic_args()
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -2522,7 +2522,8 @@ https://gcc.gnu.org/bugzilla/show_bug.cgi?id=47485'''))
             commands += self.build.get_global_link_args(linker, target.for_machine)
             # Link args added from the env: LDFLAGS. We want these to override
             # all the defaults but not the per-target link args.
-            commands += self.environment.coredata.get_external_link_args(target.for_machine, linker.get_language())
+            copt_proxy = self.get_compiler_options_for_target(target)
+            commands += copt_proxy[linker.get_language() + '_args'].value
 
         # Now we will add libraries and library paths from various sources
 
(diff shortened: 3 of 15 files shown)


test cases/common/229 persubproject options/meson.build
```diff
@@ -1,10 +1,24 @@
...
-subproject('sub2', default_options : ['default_library=static'])
+sub1 = subproject('sub1')
+sub2 = subproject('sub2',
+  default_options : [
+    'default_library=static',
+    'c_args=-DSUB2_FROM_PARENT',
+  ]
+)
+
+test('mixing-cpp-version', executable('test1', 'main.cpp', link_with : sub2.get_variable('libcpp11')))
```
TheQwertiest: There are no ABI compatibility guarantees for c++98 and c++11 AFAIK. There *are* guarantees for c++11, c++14 and c++17 though.
I.e. it's safe to mix c++11, c++14 and c++17, but not with c++03/c++98.

TheQwertiest: This is UB AFAIK, since it's not safe to mix c++98/03 with c++11/c++14/c++17. It *is* safe to mix  c++11/c++14/c++17 though.
This would've been alright if `sub2` was a dynamic library with C interface.

TheQwertiest: I'm too lazy to look through every compiler, but here is the proof for `gcc`:

> The C++98 language is ABI-compatible with the C++11 language, but several places in the Standard Library break compatibility. This makes it potentially dangerous to link C++98 objects with C++11 objects. 
Source: https://gcc.gnu.org/wiki/Cxx11AbiCompatibility

xclaesse: > This would've been alright if `sub2` was a dynamic library with C interface.

That's the intention, yes. I think it's fine because void foo(void) is simple enough to have the same ABI, but it should have `extern "C"` to make it explicit.

TheQwertiest: In this test it's built and linked as a static library `default_library=static`, and there is `std::shared_ptr` in `sub2/foo.cpp`. 

Either way, it seems that my gcc knowledge on this subject is flawed, since there is a direct confirmation of c++03/c++11+ ABI compatibility by `libstdc++` author: https://stackoverflow.com/a/49119902/6310318

xclaesse: > In this test it's built and linked as a static library `default_library=static`, and there is `std::shared_ptr` in `sub2/foo.cpp`.

Oh good point, haven't noticed that was static linked. Needs to be fixed indeed.

> Either way, it seems that my gcc knowledge on this subject is flawed, since there is a direct confirmation of c++03/c++11+ ABI compatibility by `libstdc++` developer: https://stackoverflow.com/a/49119902/6310318

That's gcc specific, if I understand correctly. Wouldn't be surprised that's in the category of undefined in the spec but all compilers actually keep compat.

TheQwertiest: > Wouldn't be surprised that's in the category of undefined in the spec but all compilers actually keep compat.

C++ standard doesn't know anything about ABI at all :D
Regarding other compilers: clang claims to be ABI compatible with GCC (thus theoretically it should handle C++XX compatibility the same); MSVC is ABI compatible across C++14/C++17 (https://www.reddit.com/r/cpp/comments/alsoqx/c_binary_compatibility_and_painfree_upgrades_to/efhj7kp)

jpakkane:
As functionality this is good, but just to be sure they should only be settable in the native file (or equivalent). Putting them in meson configure makes the output explode (not to mention passing them as arguments to `subproject` et al).

xclaesse:
Having them from command line and in `subprojects(default_options:...)` is the whole point. It makes no sense to mean to threat those options differently.

jpakkane:
That has bad and confusing failure modes. Suppose you have a project that does this:

```meson
subproject'('foo', cpp_std=c++17 <or whatever>)
```

What is the value of `cpp_std` for that subproject now? It can be 17. It can be any other value. It can get *different* values on different machines, even (in the extreme case) different values on the _same_ machine.

How is this possible? Well you may have a _second_ subproject call somewhere else in your project that gets called first under some circumstances. You can't really know. The end result is the wrong value for the option and a very confused and angry user.

Setting these values from within build files is simple and easy _assuming_ that your project setup is simple but it gets unwieldy when the project size grows. Setting them in a file is the exact opposite: a bit inconvenient for small projects but an absolute life saver for large projects. Since the tendency seems to be to go towards ever bigger nonhomogeneous projects, we should be very careful in our design to that we don't screw up the configuration experience for large projects.

xclaesse:
We already have `default_options` mechanism for project() and subproject() and `-Dsub:opt=val`, we are just reusing the same system, that's consistent. It makes no sense to me to treat compiler options differently. You already bringed that idea of limiting per subproject default_library to native file, and everyone disagreed.

However, setting sub:cpp_std=c++17 in your native file (when that gets supported) override any default options set in project() and subproject(), so user is in full control in the end.

jpakkane:
There is a big difference between project options and general options. The former are not shared between projects and isolation between them is natural unless you explicitly `yield`. The latter are different in that the default assumption is that they are the same between different subprojects. If you have 15 different subprojects, then you have potentially 15 different toggle options for the language standard. The only way to make that work in any reasonable way is to define them in one, and only one, place.

xclaesse:
Deciding things like cpp_std is not something the user will usually decide. The whole point here is to fix cases like this:

- My project requires c++14 so I have `project(..., default_options=['cpp_std=c++14'])`.
- I optionally depend on a subproject that itself requires c++17. That subproject has `project(..., default_options=['cpp_std=c++17'])`.

Currently the master project force the subproject to use c++14 and it fails to build. With this branch, the subproject will be built with c++17 and everyone is happy, the user does not have to know that detail and does not even have to write a native file.

TheQwertiest:
Another thing to consider is `Werror`. You can't fix warnings in the project you don't own.

xclaesse:
It has been discussed a lot in https://github.com/mesonbuild/meson/issues/1889

xclaesse:
> Another thing to consider is `Werror`. You can't fix warnings in the project you don't own.

That's also a case the user does not always has to handle. In many cases the master project knows its subproject generate many warnings and can call `dependency('foo', fallback: ['sub', 'foo_dep'], default_options : ['werror=false'])`.

xclaesse:
> Putting them in meson configure makes the output explode

But that's required to make those options discoverable. Here is how it looks like on gst-build, probably the biggest subprojects consumer we've got: [options-list.txt](https://github.com/mesonbuild/meson/files/4284605/options-list.txt). As you can see, I already did the work months ago to make those options sorted by subproject. I think it's not that bad. I have an alias that pipe meson configure to less, we should make that automatic like many git commands.

dcbaker:
For native/cross projects when/if this lands I can rebase #6597, or we can land that first and ensure that these changes work with that as well.

xclaesse:
Closing this since the code changed a lot, replaced by https://github.com/mesonbuild/meson/pull/7521.

PR #6730, opened 2020-03-03
qt5: Add has_tools() method

xclaesse:


xclaesse:
This is needed to fix broken code in GStreamer: https://gitlab.freedesktop.org/gstreamer/gst-plugins-base/-/blob/master/tests/examples/overlay/meson.build#L15. That check does not work on Ubuntu because the tool is called "moc" instead of "moc-qt5", and won't work in the case the pkg-config version doesn't match `moc --version` which happens when cross compiling and the qt version on the system does not match the version from sysroot.

PR #6731, opened 2020-03-03
add setup keyword in example commands

michaelbadcrumble:


jpakkane:
Thanks, but remember that doc-only changes should have `[skip ci]` in their commit messages.

PR #6733, opened 2020-03-04
interpreter: add 'name' method to BuildTargetHolder

dabrain34:
As any child of ObjectHolder might need the name of the object,
provides a method to get name.
This is useful in gst-build to display the plugin name and not
the filename.

### DIFF
--- a/docs/markdown/snippets/build_target_older_name.md
+++ b/docs/markdown/snippets/build_target_older_name.md
@@ -1,0 +1,2 @@
+## Added 'name' method
+Build target objects (as returned by executable(), library(), ...) now have a name() method.
--- a/mesonbuild/interpreter.py
+++ b/mesonbuild/interpreter.py
@@ -771,6 +771,7 @@ class BuildTargetHolder(TargetHolder):
         super().__init__(target, interp)
         self.methods.update({'extract_objects': self.extract_objects_method,
                              'extract_all_objects': self.extract_all_objects_method,
+                             'name': self.name_method,
                              'get_id': self.get_id_method,
                              'outdir': self.outdir_method,
                              'full_path': self.full_path_method,
@@ -824,6 +825,12 @@ class BuildTargetHolder(TargetHolder):
     @permittedKwargs({})
     def get_id_method(self, args, kwargs):
         return self.held_object.get_id()
+
+    @FeatureNew('name', '0.54.0')
+    @noPosargs
+    @permittedKwargs({})
+    def name_method(self, args, kwargs):
+        return self.held_object.name
 
 class ExecutableHolder(BuildTargetHolder):
     def __init__(self, target, interp):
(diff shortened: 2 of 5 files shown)


xclaesse:
I would be surprised every held object have a name attribute. This will lead to crash for those that don't. I'm not against adding name method on objects that have one, but it needs to be fine grained IMHO.

xclaesse:
Also need to add that method to the doc of each object that gain it.

dabrain34:
I was surprised it wasnt already available. But yeah mypy fails for unknown reason right now.
I wanted to be more generic and avoid duplicate codes
So I can add it to the `BuildTargetHolder` at least in a first step

mesonbuild/interpreter.py
```diff
@@ -825,6 +826,11 @@ def extract_all_objects_method(self, args, kwargs):
     def get_id_method(self, args, kwargs):
         return self.held_object.get_id()
 
+    @noPosargs
+    @permittedKwargs({})
+    def name_method(self, args, kwargs):
```
xclaesse: missing `@featureNew('name', '0.54.0')`

test cases/common/1 trivial/meson.build
```diff
@@ -20,7 +20,7 @@ if meson.is_cross_build()
 endif
 
 exe = executable('trivialprog', sources : sources)
-
+assert(exe.name() == 'trivialprog', 'The object name is incorrect')
```
xclaesse: It's fine, but just FYI, the 2 argument to assert() is not required any more. It will print `exe.name() == 'trivialprog' assertion failed` automatically, which is just as good.

xclaesse: marking as resolved, it doesn't have to be changed ;-)

xclaesse:
Also missing a release notes snippet. Other than those small nitpicking, it's good to go.

docs/markdown/snippets/build_target_older_name.md
```diff
@@ -0,0 +1,3 @@
+## Added 'name' property
```
xclaesse: It's not a property but a method.

docs/markdown/snippets/build_target_older_name.md
```diff
@@ -0,0 +1,3 @@
...
+As any child of BuildTargetHolder might need the name of the object,
```
xclaesse: BuildTargetHolder is the internal class name, it's not user visible. I would say something like:

Build target objects (as returned by executable(), library(), ...) now have a name() method.

mesonbuild/interpreter.py
```diff
@@ -825,6 +826,12 @@ def extract_all_objects_method(self, args, kwargs):
     def get_id_method(self, args, kwargs):
         return self.held_object.get_id()
 
+    @FeatureNew('custom_target.to_list', '0.54.0')
```
xclaesse: wrong copy paste: to_list

dabrain34: oups

docs/markdown/snippets/build_target_older_name.md
dabrain34: :+1:

xclaesse:
LGTM when CI is green.

PR #6735, opened 2020-03-04
fix compiler flag order on visual studio

ericLemanissier:
currently, all "-LIBPATH:..." arguments are treated as gcc's -L, so they are prependded
they end up before the '/link' argument, so they are ignored


nirbheek:
Unit test? This will break very easily otherwise :)

ericLemanissier:
Unfortunately it's not the correct solution, it breaks non C compilers which don't have the linker_to_compiler_args method.

ericLemanissier:
this problem seems solved in recent meson version

PR #6736, opened 2020-03-04
Mesonlib type annotations

dcbaker:
This series is mainly aimed at adding type annotations to the mesonlib module, although it does have a few more fixups in it. The biggest change is splitting the unholder functionality out of listify, I've done this because listify is already basically impossible to correctly type, and splitting out the unholdering code makes it much easier.

### DIFF
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -26,7 +26,7 @@ from . import mlog
 from .mesonlib import (
     File, MesonException, MachineChoice, PerMachine, OrderedSet, listify,
     extract_as_list, typeslistify, stringlistify, classify_unity_sources,
-    get_filenames_templates_dict, substitute_values, has_path_sep,
+    get_filenames_templates_dict, substitute_values, has_path_sep, unholder
 )
 from .compilers import Compiler, is_object, clink_langs, sort_clink, lang_suffixes
 from .linkers import StaticLinker
@@ -856,7 +856,7 @@ just like those detected with the dependency() function.''')
         if dfeature_debug:
             dfeatures['debug'] = dfeature_debug
         if 'd_import_dirs' in kwargs:
-            dfeature_import_dirs = extract_as_list(kwargs, 'd_import_dirs', unholder=True)
+            dfeature_import_dirs = unholder(extract_as_list(kwargs, 'd_import_dirs'))
             for d in dfeature_import_dirs:
                 if not isinstance(d, IncludeDirs):
                     raise InvalidArguments('Arguments to d_import_dirs must be include_directories.')
@@ -1083,7 +1083,7 @@ You probably should put it in link_with instead.''')
         return isinstance(self, StaticLibrary) and not self.need_install
 
     def link(self, target):
-        for t in listify(target, unholder=True):
+        for t in unholder(listify(target)):
             if isinstance(self, StaticLibrary) and self.need_install and t.is_internal():
                 # When we're a static library and we link_with to an
                 # internal/convenience library, promote to link_whole.
@@ -1105,7 +1105,7 @@ You probably should put it in link_with instead.''')
             self.link_targets.append(t)
 
     def link_whole(self, target):
-        for t in listify(target, unholder=True):
+        for t in unholder(listify(target)):
             if isinstance(t, (CustomTarget, CustomTargetIndex)):
                 if not t.is_linkable_target():
                     raise InvalidArguments('Custom target {!r} is not linkable.'.format(t))
@@ -1261,7 +1261,7 @@ You probably should put it in link_with instead.''')
                     if dl != linker.language:
                         stdlib_args += all_compilers[dl].language_stdlib_only_link_flags()
                         added_languages.add(dl)
-                # Type of var 'linker' is Compiler. 
+                # Type of var 'linker' is Compiler.
                 # Pretty hard to fix because the return value is passed everywhere
                 return linker, stdlib_args
 
@@ -1372,7 +1372,7 @@ class Generator:
                 raise InvalidArguments('Capture must be boolean.')
             self.capture = capture
         if 'depends' in kwargs:
-            depends = listify(kwargs['depends'], unholder=True)
+            depends = unholder(listify(kwargs['depends']))
             for d in depends:
                 if not isinstance(d, BuildTarget):
                     raise InvalidArguments('Depends entries must be build targets.')
@@ -2054,7 +2054,7 @@ class CustomTarget(Target):
         return bdeps
 
     def flatten_command(self, cmd):
-        cmd = listify(cmd, unholder=True)
+        cmd = unholder(listify(cmd))
         final_cmd = []
         for c in cmd:
             if isinstance(c, str):
@@ -2082,7 +2082,7 @@ class CustomTarget(Target):
 
     def process_kwargs(self, kwargs, backend):
         self.process_kwargs_base(kwargs)
-        self.sources = extract_as_list(kwargs, 'input', unholder=True)
+        self.sources = unholder(extract_as_list(kwargs, 'input'))
         if 'output' not in kwargs:
             raise InvalidArguments('Missing keyword argument "output".')
         self.outputs = listify(kwargs['output'])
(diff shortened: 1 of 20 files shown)


mesonbuild/build.py
```diff
@@ -2411,18 +2410,18 @@ def get_target_name(self):
 class ConfigurationData:
     def __init__(self):
```
bluetech:
```suggestion
    def __init__(self) -> None:
```

Otherwise it's not type-checked.

mesonbuild/mesonlib.py
```diff
@@ -982,49 +1036,57 @@ def replace_if_different(dst, dst_tmp):
...
+def unholder(item: T.List['ObjectHolder[_T]']) -> T.List[_T]: ...
+
+@T.overload
+def unholder(item: T.List[_T]) -> T.List[_T]: ...
+
+@T.overload
+def unholder(item: T.List[T.Union[_T, 'ObjectHolder[_T]']]) -> T.List[_T]: ...
+
+def unholder(item):
+    """Get the held itme of an object holder of list of object holers."""
```
bluetech:
```suggestion
    """Get the held item of an object holder or list of object holders."""
```

(I think...)

mesonbuild/mesonlib.py
```diff
@@ -982,49 +1036,57 @@ def replace_if_different(dst, dst_tmp):
...
+@T.overload
+def unholder(item: T.List['ObjectHolder[_T]']) -> T.List[_T]: ...
+
+@T.overload
+def unholder(item: T.List[_T]) -> T.List[_T]: ...
+
+@T.overload
+def unholder(item: T.List[T.Union[_T, 'ObjectHolder[_T]']]) -> T.List[_T]: ...
+
+def unholder(item):
```
bluetech: If you want the *implementation* to be type-checked, you need to annotate it as well (with a superset of all the individual overloads). Might be too annoying though.

mesonbuild/mesonlib.py
```diff
@@ -982,49 +1036,57 @@ def replace_if_different(dst, dst_tmp):
...
 
-def listify(item: T.Any,
-            flatten: bool = True,
-            unholder: bool = False) -> T.List[T.Any]:
+
+@T.overload
+def unholder(item: 'ObjectHolder[_T]') -> _T: ...
+
+@T.overload
+def unholder(item: T.List['ObjectHolder[_T]']) -> T.List[_T]: ...
```
bluetech: You might consider using `Iterable` for the input. Then is more flexible and provides a guarantee to the caller that the list won't be modified. If `Iterable` doesn't work then `Sequence` usually does.

mesonbuild/mesonlib.py
```diff
@@ -1332,57 +1409,63 @@ def detect_subprojects(spdir_name, current_dir='', result=None):
...
-class OrderedSet(collections.abc.MutableSet):
+
+class OrderedSet(T.MutableSet[_T]):
     """A set that preserves the order in which items are added, by first
     insertion.
     """
-    def __init__(self, iterable=None):
-        self.__container = collections.OrderedDict()
+    def __init__(self, iterable: T.Optional[T.Iterable[_T]] = None):
+        # typing.OrderedDict is new in 3.7.2, so we can't use that, but we can
```
bluetech: Since you will only be using `typing.OrderedDict` during type checking, you *can* actually use it if you want, since it is defined in the stubs that mypy uses. Understandable if you don't want to though.

mesonbuild/mesonlib.py
```diff
@@ -1332,57 +1409,63 @@ def detect_subprojects(spdir_name, current_dir='', result=None):
...
 
-class OrderedSet(collections.abc.MutableSet):
+
+class OrderedSet(T.MutableSet[_T]):
     """A set that preserves the order in which items are added, by first
     insertion.
     """
-    def __init__(self, iterable=None):
-        self.__container = collections.OrderedDict()
+    def __init__(self, iterable: T.Optional[T.Iterable[_T]] = None):
```
bluetech:
```suggestion
    def __init__(self, iterable: T.Optional[T.Iterable[_T]] = None) -> None:
```

mesonbuild/mesonlib.py
```diff
@@ -1332,57 +1409,63 @@ def detect_subprojects(spdir_name, current_dir='', result=None):
...
+        # typing.OrderedDict is new in 3.7.2, so we can't use that, but we can
+        # use MutableMapping, which is fine in this case.
+        self.__container = collections.OrderedDict()  # type: T.MutableMapping[_T, None]
         if iterable:
             self.update(iterable)
 
-    def __contains__(self, value):
+    def __contains__(self, value: T.Any) -> bool:
```
bluetech:
```suggestion
    def __contains__(self, value: object) -> bool:
```

run_unittests.py
```diff
@@ -686,12 +686,17 @@ def test_listify(self):
...
         self.assertEqual([holder1, 2], listify([holder1, 2]))
         self.assertEqual([holder1, 2, 3], listify([holder1, 2, [3]]))
-        self.assertEqual([1], listify(holder1, unholder=True))
-        self.assertEqual([1], listify([holder1], unholder=True))
-        self.assertEqual([1, 2], listify([holder1, 2], unholder=True))
-        self.assertEqual([1, 2, 3], listify([holder1, 2, [holder3]], unholder=True))
-        # Unholding doesn't work recursively when not flattening
-        self.assertEqual([1, [2], [holder3]], listify([holder1, [2], [holder3]], unholder=True, flatten=False))
+
+    def test_unholder(self):
```
bluetech:
```suggestion
    def test_unholder(self) -> None:
```

... if you want the test to be type-checked. Usually a good way to validate the type annotations.

mesonbuild/mesonlib.py
```diff
@@ -109,22 +118,31 @@ def check_direntry_issues(direntry_array):
...
 # by accident.
 import threading
 an_unpicklable_object = threading.Lock()
 
+
 class MesonException(Exception):
     '''Exceptions thrown by Meson'''
 
+    file = None    # type: T.Optional[str]
+    lineno = None  # type: T.Optional[str]
```
bluetech: Just in case this is a mistake, and just from the names, I would expect `linono` and `colno` to be `int`s.

mesonbuild/build.py
dcbaker: I just went and read up on this. The rules are absolutely stupid, for anyone else wondering:

If your `__init__` has arguments, and those are typed you may omit the return type and still get typing, if the `__init__` does not  have arguments (other than self), you must include the `-> None`.

mensinda: Wow... In that case, should we just keep it consistent and add `-> None` to every init?

mesonbuild/mesonlib.py
dcbaker: We run mypy with python 3.5 though, and it gets grumpy about undefined type, so I'm going to just use MutableMapping. Since the actual OrderedDict is a private implementation detail anyway.

mesonbuild/mesonlib.py
dcbaker: The docs say not to do this, and when I tried mypy complains about any signature I give it, complaining that one or more overloads doesn't match that signature. I'm going to leave this plain.

mesonbuild/mesonlib.py
dcbaker: You are correct, those should be ints.

mesonbuild/mesonlib.py
dcbaker: Interesting, I didn't realize that `Any` and `object` were not equivalent. For anyone else curious, Any basically means "dont type check this", object means "assume that the thing in this parameter is a descendant of `object` and don't allow using methods that `object` doesn't define.

mesonbuild/mesonlib.py
dcbaker: We seem not not use a return type on `__init__` when it has arguments, so for now I'm going to leave this as-is, I'd rather do a batch '-> None:` to all `__init__`s if we're going to do that.

mesonbuild/build.py
bluetech: > I just went and read up on this. The rules are absolutely stupid, for anyone else wondering

I wouldn't call them stupid, they do make sense.

> Wow... In that case, should we just keep it consistent and add -> None to every init?

This is what I would do. No reason to treat `__init__` as a special case.

mesonbuild/mesonlib.py
bluetech: > The docs say not to do this

The mypy docs at least don't say not to do this: https://mypy.readthedocs.io/en/latest/more_types.html#function-overloading

> I'm going to leave this plain.

Sure, sometimes it's not worth it.

PR #6743, opened 2020-03-05
intro: List symlinks in installdata (fixes #6734)

### LINKED ISSUE #6734: meson introspect --installed does not list symlinks
##  Problem

When enumerating installed artifacts with `meson introspect --installed` (with `ninja` on Linux, probably elsewhere as well), a `shared_library()` such as `libfoo.so.1.2.3` will appear, but auto-generated (and still installed) symlinks like `libfoo.so.1` and `libfoo.so` are not. Since they are nevertheless installed, I'd expect them to be enumerated.

## Reproduction

For instance,
```bash
cd "test cases/linuxlike/7 library versions"
meson build/
meson introspect --installed --indent build/
```
produces 
```json
{
    "/home/meson/test cases/linuxlike/7 library versions/build/libsome.so.1.2.3": "/usr/local/lib64/libsome.so.1.2.3",
    "/home/meson/test cases/linuxlike/7 library versions/build/libnoversion.so": "/usr/local/lib64/libnoversion.so",
    "/home/meson/test cases/linuxlike/7 library versions/build/libonlyversion.so.1.4.5": "/usr/local/lib64/libonlyversion.so.1.4.5",
    "/home/meson/test cases/linuxlike/7 library versions/build/libonlysoversion.so.5": "/usr/local/lib64/libonlysoversion.so.5",
    "/home/meson/test cases/linuxlike/7 library versions/build/libmodule.so": "/usr/local/lib64/libmodule.so"
}
```
yet installation with `cd build && ninja && ninja install` would place the symlinks `libsome.so` and `libsome.so.0` side-by-side with `libsome.so.1.2.3`.

## Expected

It would have been nice if there was a symmetry between `ninja install`'s actions and the accounting that `meson introspect --installed` gives. It's much easier to detect and remove symlinks from a list of installed targets that include them, than determining from a list of installed files which excluded symlinks transitively point into them.

Alternately, a new `--installed-extra` flag could be added, or a `--dry-run --verbose --json` flag set could be added to `meson install`, or some other machine-readable form to present that information.

mensinda:


jpakkane:
Crap, those travis failures caused me not to see this actual macOS failure:

```
{'libmodule.dylib': '/usr/lib/libmodule.dylib',
?             ^^^^^                       ^^^^^
+ {'libmodule.so': '/usr/lib/libmodule.so',
?             ^^                       ^^
-  'libnoversion.dylib': '/usr/lib/libnoversion.dylib',
?                ^^^^^                          ^^^^^
+  'libnoversion.so': '/usr/lib/libnoversion.so',
?                ^^                          ^^
```

PR #6750, opened 2020-03-06
[WIP] mesonbuild: allow multiple build dirs via -C option

asartori86:
It is common to have several `builddir`s set up for a project. This patch allows deploying the `builddir`s in one command
```
meson.py compile -C build-gcc,build-clang
```

it is just a proposal, if you like the feature I would like to extend it to the `meson test` command as well.

### DIFF
--- a/mesonbuild/mcompile.py
+++ b/mesonbuild/mcompile.py
@@ -68,10 +68,10 @@ def run(options: 'argparse.Namespace')-> int:
             ret += single_run(options, pathlib.Path(b))
         return ret
     else:
-        return single_run(options, pathlib.Path(b))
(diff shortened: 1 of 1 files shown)


eli-schwartz:
I feel this would be inconsistent since you also need to run `meson setup` manually, multiple times. And this doesn't really add any utility over

```
ninja -C builddir1 && ninja -C builddir2 && ninja -C builddir3
```

It saves maybe a fractional bit of typing at best...

Anyway it seems evident that no one on the core team at the time was interested enough in the change to respond or consider merging it. So I'm going to consider this a wont-implement.

...

BTW, implementation note: this would have broken build directories that contained commas in them (a valid directory name character).

Issue #6752, opened 2020-03-06
buildtype set in project() default_options overrides -Doptimization and -Ddebug set via command-line args


nirbheek:
So, f.ex., if you have:

```meson
project('foobar')
```

And you configure with: `meson setup _builddir -Doptimization=2 -Ddebug=true`, you get:

```meson
get_option('buildtype') == 'debugoptimized'
get_option('debug') == true
get_option('optimization') == '2'
```

*However*, if you have this:

```meson
project('foobar', default_options: ['buildtype=debug'])
```

And you again configure with: `meson setup _builddir -Doptimization=2 -Ddebug=true`, you get:

```meson
get_option('buildtype') == 'debug'
get_option('debug') == true
get_option('optimization') == '0'
```

This happens because we start with a dict containing the command-line arguments, then use `dict.setdefault()` to set default values from the native file and `default_options`. This works fine for options which only have one 'name'. Unfortunately, for buildtype arguments, this ends up being equivalent to: `-Doptimization=2 -Ddebug=true -Dbuildtype=debug`. The fix is to reverse the order in which things are set, working on a PR for this.

xclaesse:
I'm tempted to instead apply `buildtype` first regardless of the order the user uses, just like we do for `prefix`.

nirbheek:
> I'm tempted to instead apply `buildtype` first regardless of the order the user uses, just like we do for `prefix`.

I'm not sure I like that. I do agree that setting both buildtype and optimization/debug on the command-line is a suboptimal configuration. I suggest that we warn about it.

xclaesse:
Or even abort on it like we do for `--buildtype foo -Dbuildtype=bar`.

Issue #6742, opened 2020-03-05
-Dbuildtype silently overwites setting from -Doptimization


bredelings:
**Describe the bug**
When specifying both -Dbuildtype=plain and -Doptimization=3, meson selects buildtype=plain and optimization=0.  No warning is given.  Also, the order of the arguments appears not to matter.

**Expected behavior**
I expect that when specifying -Doptimization=3 _after_ -Dbuildtype=plain, that the optimization level would be 3.

If this behavior is not acceptable, then I expect a message like "You specified both -Doptimization and -Dbuildtype.  Ignoring -Doptimization".

bredelings:
This bug report is wrong.  The order _does_ matter.  I think I was confused, because I did not think of that at first.

xclaesse:
I think the order is kind of undefined, -D options goes into a python dict which only guarantee order with python >=3.6, and we depend on 3.5. So it could be working by chance most of the time.

In any case, I don't like depending on the ordering, I think we should apply buildtype first, then override with debug/optimization. This is a case that always bothered me because it's undocumented what happens AFAIK.

bredelings:
Hmm... how about if you give two different values for the same option, such as -Dbuildtype=plain -Dbuildtype=release?  Currently debian packages automatically supply -Dbuildtype=plain, so the ability to override by specifying -Dbuildtype=release afterwards is important.

xclaesse:
Good point. I think we would need to double check if we always put this key:Val in an OrderedDict.

bredelings:
Along these same lines, the list of options as reported in meson-log.txt does not match the order they are given on the command line.

nirbheek:
I recently found a different bug in the same code: `buildtype` set in project `default_options` was overriding `debug` and `optimization` values passed on the command line. I have a WIP fix for that and the bug described in this issue.

nirbheek:
> This is a case that always bothered me because it's undocumented what happens AFAIK.

I've partially documented it in https://github.com/mesonbuild/meson/pull/6717

nirbheek:
https://github.com/mesonbuild/meson/pull/6753

bredelings:
ANother issue is that I think setting -Doptimization=3 sets -Db_ndebug=true
I might be wrong about that, but it that is right, then a warning about that would also be good.

Issue #4686, opened 2018-12-29
Passing both --buildtype=debug{,optimized} and --debug results in no debug CFLAGS


perror:
I recently noticed that using the options `--buildtype=debug --debug` with `gcc` on a gnulinke platform was not triggering the usual debug options (aka `-g`). This issue refer to [glib merge request #566](https://gitlab.gnome.org/GNOME/glib/merge_requests/566).

It seems that the default flags for debug are defined as _empty_... I suggest the following patch (but it might be not enough to cover all, I noticed a lot of empty `debug` and `debugoptimized` options which are not filled properly in these templates.

```
diff --git a/mesonbuild/compilers/compilers.py b/mesonbuild/compilers/compilers.py
index 2a5c9763..0aa8080a 100644
--- a/mesonbuild/compilers/compilers.py
+++ b/mesonbuild/compilers/compilers.py
@@ -127,7 +127,7 @@ def is_library(fname):
     return suffix in lib_suffixes
 
 gnulike_buildtype_args = {'plain': [],
-                          'debug': [],
+                          'debug': ['-O0', '-g'],
                           'debugoptimized': [],
                           'release': [],
                           'minsize': [],

```

Let me know if you want me to fill all the defaults which I believe not to be properly filled (obviously, the `debugoptimized` is the second that comes in mind).

nirbheek:
The code there is a bit confusing, but `-g` is added [lower down](https://github.com/mesonbuild/meson/blob/master/mesonbuild/compilers/compilers.py#L333):

```python
clike_debug_args = {False: [],
                    True: ['-g']}
```

What version of meson are you using, what are you testing, what do you expect, and what do you observe?

nirbheek:
Verified, passing `--buildtype=debug` or `debugoptimized` and `--debug` leads to no debug flags being set.

copelnug:
Actually, the problem is simply with `--debug`. It seems the problem is in the coredata.py file where the functions `get_builtin_option_action` and  `get_builtin_option_default` use the third ([2]) field of the built-in options. 

While reading those function, it become apparent that `debug` being `True` by default create a design problem. This has a couple of impacts:
1. The implementation of `get_builtin_option_action` will do the opposite of what the option name suggest (putting `False` at debug).
2. Even if it did what was suggested by the name, it would be redundant.
3. If it did what was suggested by the name, there would be no way to disable that option (except the special case of buildtype).

## Impact
The problem seems to concern three options and make them behave strangely at setup:
[...]

PR #6753, opened 2020-03-06
Fix some build type option parsing bugs

nirbheek:
commit 7cd9fcef125a9ad6f7416f647e70155e205c1c9d:

```
coredata: Convert all option parsing to OrderedDict()
This ensures that options are always parsed in the order in which they
were specified on the command-line, even with Python 3.5, and
non-CPython implementations compatible with CPython 3.5 and 3.6.
```
Closes https://github.com/mesonbuild/meson/issues/6742

commit 6ee7450aaf99b24343eb3680457830dea718151b:

```
coredata: Set default options as cmdline args that override each other
The previous code was assuming that options do not depend on each
other, and that you can set defaults using `dict.setdefault()`. This
is not true for `buildtype` + `optimization`/`debug`, so we add
defaults + overrides in the right order and use the options parsing
code later to compute the values.

Includes a test.
```
Closes https://github.com/mesonbuild/meson/issues/6752

commit b52420cc4c292feeff710f56d1ea36bab3cea973:

```
coredata: Passing an option is supposed to set it, not flip the default
With the current logic passing `--debug` will actually be parsed as
`-Ddebug=false`, which is absolutely not what is expected.

There is no case in which the presence of a boolean option in `--foo`
form will mean 'I want feature foo disabled', regardless of the
*default* value of that option.

Also includes a test.
```
Closes https://github.com/mesonbuild/meson/issues/4686

commit 4dc0f5a1631d36cd015dc0071a3cd4730cb0aad8:

```
coredata: Warn on usage of both -Dbuildtype and -Doptimization/-Ddebug
It may not be obvious to users that these two ways to set build-types
override each other and specifying both is redundant, and conflicts
are resolved based on whichever is specified later.
```
Closes https://github.com/mesonbuild/meson/issues/6742

### DIFF
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -718,6 +718,13 @@ class CoreData:
             self.copy_build_options_from_regular_ones()
 
     def set_default_options(self, default_options, subproject, env):
+        # Warn if the user is using two different ways of setting build-type
+        # options that override each other
+        if 'buildtype' in env.cmd_line_options and \
+           ('optimization' in env.cmd_line_options or 'debug' in env.cmd_line_options):
+            mlog.warning('Recommend using either -Dbuildtype or -Doptimization + -Ddebug. '
+                         'Using both is redundant since they override each other. '
+                         'See: https://mesonbuild.com/Builtin-options.html#build-type-options')
         cmd_line_options = OrderedDict()
         # Set project default_options as if they were passed to the cmdline.
         # Subprojects can only define default for user options and not yielding
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -718,13 +718,8 @@ class CoreData:
             self.copy_build_options_from_regular_ones()
 
     def set_default_options(self, default_options, subproject, env):
-        # Set defaults first from conf files (cross or native), then
-        # override them as nec as necessary.
-        for k, v in env.paths.host:
-            if v is not None:
-                env.cmd_line_options.setdefault(k, v)
-
-        # Set default options as if they were passed to the command line.
+        cmd_line_options = OrderedDict()
+        # Set project default_options as if they were passed to the cmdline.
         # Subprojects can only define default for user options and not yielding
         # builtin option.
         from . import optinterpreter
@@ -734,7 +729,17 @@ class CoreData:
                         and optinterpreter.is_invalid_name(k, log=False):
                     continue
                 k = subproject + ':' + k
-            env.cmd_line_options.setdefault(k, v)
+            cmd_line_options[k] = v
+
+        # Override project default_options using conf files (cross or native)
+        for k, v in env.paths.host:
+            if v is not None:
+                cmd_line_options[k] = v
+
+        # Override all the above defaults using the command-line arguments
+        # actually passed to us
+        cmd_line_options.update(env.cmd_line_options)
+        env.cmd_line_options = cmd_line_options
 
         # Create a subset of cmd_line_options, keeping only options for this
         # subproject. Also take builtin options if it's the main project.
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -1001,9 +1001,10 @@ class BuiltinOption(T.Generic[_T, _U]):
         return self.opt_type(self.description, **keywords)
 
     def _argparse_action(self) -> T.Optional[str]:
-        if self.default is True:
-            return 'store_false'
-        elif self.default is False:
+        # If the type is a boolean, the presence of the argument in --foo form
+        # is to enable it. Disabling happens by using -Dfoo=false, which is
+        # parsed under `args.projectoptions` and does not hit this codepath.
+        if isinstance(self.default, bool):
             return 'store_true'
         return None
 
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -668,7 +668,7 @@ class CoreData:
         return len(self.cross_files) > 0
 
     def strip_build_option_names(self, options):
-        res = {}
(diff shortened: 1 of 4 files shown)


nirbheek:
This adds a new deprecation, so needs approval from @jpakkane

nirbheek:
Test failure looks relevant. I bet if I retry it will go away. Must be a dict order thing somewhere that I missed.

xclaesse:
IMHO, if we want to deprecated `get_option('buildtype')` then we should deprecate that option completely, from command line too. But it is so common to check buildtype option in every project, I feel it's going to cause more troubles than needed.

nirbheek:
> IMHO, if we want to deprecated `get_option('buildtype')` then we should deprecate that option completely, from command line too. But it is so common to check buildtype option in every project, I feel it's going to cause more troubles than needed.

I think having it available on the command-line as a short-cut for the more verbose `optimization` and `debug` options is a good thing, and it's a good way to guide people towards good 'choices' for combinations of those two options. I wouldn't really mind if it was deprecated too, but I see no compelling reason for it.

On the other hand, _every single user_ of `get_option('buildtype')` is wrong.

F.ex., glib uses it to check when to disable cast checks by comparing it with the string `'release'`, which is going to be completely wrong when you're cross-compiling to embedded, where you'll use `-Doptimization=s` (`'minsize'`) or `2` (`'custom'`) and not `-Doptimization=3` (which generates larger code).

Glib also uses it to check when debugging is enabled to enable some debug infrastructure, which is _very_ wrong, but that code was written before the `debug` option was available and it has been copied everywhere.

nirbheek:
I have dropped the deprecation from this PR since it needs more discussion. Will open a separate one for that.

mesonbuild/coredata.py
```diff
@@ -1001,9 +1001,10 @@ def init_option(self, name: str, value: T.Optional[T.Any], prefix: str) -> _U:
         return self.opt_type(self.description, **keywords)
 
     def _argparse_action(self) -> T.Optional[str]:
-        if self.default is True:
-            return 'store_false'
-        elif self.default is False:
+        # If the type is a boolean, the presence of the argument in --foo form
+        # is to enable it. Disabling happens by using -Dfoo=false, which is
+        # parsed under `args.projectoptions` and does not hit this codepath.
+        if isinstance(self.default, bool):
```
xclaesse: While this is obviously the right thing to do, it also means --debug, --errorlogs and --stdsplit are useless since they are the default. They are still useful in their -D form, so I don't think it's an issue.

mesonbuild/coredata.py
```diff
@@ -718,6 +718,13 @@ def set_options(self, options, *, subproject='', warn_unknown=True):
             self.copy_build_options_from_regular_ones()
 
     def set_default_options(self, default_options, subproject, env):
+        # Warn if the user is using two different ways of setting build-type
+        # options that override each other
+        if 'buildtype' in env.cmd_line_options and \
+           ('optimization' in env.cmd_line_options or 'debug' in env.cmd_line_options):
+            mlog.warning('Recommend using either -Dbuildtype or -Doptimization + -Ddebug. '
+                         'Using both is redundant since they override each other. '
+                         'See: https://mesonbuild.com/Builtin-options.html#build-type-options')
```
xclaesse: Would be nice to add a title above the buildtype/debug/optimization table so we can link directly to that part of the doc.

mesonbuild/coredata.py
```diff
@@ -718,6 +718,13 @@ def set_options(self, options, *, subproject='', warn_unknown=True):
             self.copy_build_options_from_regular_ones()
 
     def set_default_options(self, default_options, subproject, env):
+        # Warn if the user is using two different ways of setting build-type
+        # options that override each other
+        if 'buildtype' in env.cmd_line_options and \
+           ('optimization' in env.cmd_line_options or 'debug' in env.cmd_line_options):
+            mlog.warning('Recommend using either -Dbuildtype or -Doptimization + -Ddebug. '
```
xclaesse: I'm not native English speaker, but I don't think "Recommend using" is the proper formulation here. Shouldn't it be "It is recommended to use either ..." ?

xclaesse:
@nirbheek Looks good to me, I only had nitpicking comments.

The inverted logic of --debug is really bad, if you're planning a 0.53.3 that should definitely go in it, IMHO.

mesonbuild/coredata.py
nirbheek: They're not entirely useless: they can still override `default_options:` in `project()`. But yes.

mesonbuild/coredata.py
nirbheek: Good point.

mesonbuild/coredata.py
nirbheek: I'm not a native English speaker either, but as I understand it, here `Recommend using` is a shortened assertive form, similar to what's used in newspaper headlines. It's short for `We recommend using` where `we` is obvious from context. Words like `It is` and `We` are just noise in short messages.

nirbheek:
> The inverted logic of --debug is really bad, if you're planning a 0.53.3 that should definitely go in it, IMHO.

Yeah, but it's also a very old bug, and I don't think we will do a 0.53.3. There's no stable distro releases around this time that will benefit from a point release, afaik?

mesonbuild/coredata.py
nirbheek: Wow, I just realized that I thought I already did that. I added the anchor in the URL but didn't add it to the documentation.

mesonbuild/coredata.py
xclaesse: ok, if you say so. It doesn't matter much :-)

Issue #6128, opened 2019-11-03
Wrong cpu  detection for 'x86_64' windows


sz-CHEN:
In ‘x64 Native Tools Command Prompt for VS 2017’, the cpu detected by meson  is 'x86' ？

``` shell
**********************************************************************
** Visual Studio 2017 Developer Command Prompt v15.9.16
** Copyright (c) 2017 Microsoft Corporation
**********************************************************************
[vcvarsall.bat] Environment initialized for: 'x64'
```

```shell
D:\opt\glib-2.62.2\_build>meson .. --buildtype=release --prefix="D:\opt\glib\"
The Meson build system
Version: 0.52.0
Source dir: D:\opt\glib-2.62.2
Build dir: D:\opt\glib-2.62.2\_build
Build type: native build
Project name: glib
Project version: 2.62.2
C compiler for the host machine: cl (msvc 19.16.27034)
C linker for the host machine: link 19.16.27034
C++ compiler for the host machine: cl (msvc 19.16.27034)
C++ linker for the host machine: link 19.16.27034
Host machine cpu family: x86
Host machine cpu: x86
```

And I tried in vs2013 , 2015, got the same result.

scivision:
I don't see this with VS2019--I didn't try with VS2017.

```
The Meson build system
Version: 0.52.0
Source dir: C:\meson\test cases\common\1 trivial
Build dir: c:\temp\11
Build type: native build
Project name: trivial test
Project version: undefined
C compiler for the host machine: cl (msvc 19.23.28106.4)
C linker for the host machine: link 19.23.28106.4
Host machine cpu family: x86_64
Host machine cpu: x86_64
``

sz-CHEN:
> I don't see this with VS2019--I didn't try with VS2017.
> 
> ```
> The Meson build system
> Version: 0.52.0
> Source dir: C:\meson\test cases\common\1 trivial
> Build dir: c:\temp\11
> Build type: native build
> Project name: trivial test
> Project version: undefined
> C compiler for the host machine: cl (msvc 19.23.28106.4)
> C linker for the host machine: link 19.23.28106.4
> Host machine cpu family: x86_64
> Host machine cpu: x86_64
> ``
> ```

May some other env

sz-CHEN:
> I don't see this with VS2019--I didn't try with VS2017.
> 
> ```
> The Meson build system
> Version: 0.52.0
> Source dir: C:\meson\test cases\common\1 trivial
[...]

PR #6759, opened 2020-03-09
Improve regular expression for MSVC compiler target arch detection

### LINKED ISSUE #6757: meson will fail when use the Chinese version of VS2017
**Describe the bug**
1. install  VS2017 with zh_cn
2. make a meson.build 
3.  exec  meson  build . when will show:
        meson.build:1:0: ERROR: Failed to detect MSVC compiler target architecture: 'cl /?' output is
���� x64 �� Microsoft (R) C/C++ �Ż������� 19.16.27035 ��

**To Reproduce**
" cl /?"  output :   
用于 x64 的 Microsoft (R) C/C++ 优化编译器 19.16.27035 版

there is a bug  in environment.py:

                cl_signature = lookat.split('\n')[0]
                match = re.search('.*(x86|x64|ARM|ARM64)$', cl_signature)
       
  This expression may fail in a non-english version.

**Expected behavior**
 match = re.search('.*(x86|x64|ARM|ARM64)', cl_signature)

**system parameters**

lb90:
Fixes #6757

### DIFF
--- a/mesonbuild/environment.py
+++ b/mesonbuild/environment.py
@@ -1063,7 +1063,7 @@ class Environment:
                     m = 'Failed to detect MSVC compiler version: stderr was\n{!r}'
                     raise EnvironmentException(m.format(err))
                 cl_signature = lookat.split('\n')[0]
-                match = re.search('.*(x86|x64|ARM|ARM64)$', cl_signature)
+                match = re.search('.*(x86|x64|ARM|ARM64)( |$)', cl_signature)
                 if match:
                     target = match.group(1)
                 else:


mesonbuild/environment.py
```diff
@@ -1063,7 +1063,7 @@ def sanitize(p):
                     m = 'Failed to detect MSVC compiler version: stderr was\n{!r}'
                     raise EnvironmentException(m.format(err))
                 cl_signature = lookat.split('\n')[0]
-                match = re.search('.*(x86|x64|ARM|ARM64)$', cl_signature)
+                match = re.search('.*(x86|x64|ARM|ARM64)( |$)', cl_signature)
```
nirbheek: If we set `VSLANG=1033` in the env, that will fix this issue too. It will force MSVC to use the eng codepage and eliminate locale-specific issues in general.

lb90: The problem is that VSLANG is not always honoured by cl.exe. See https://github.com/mesonbuild/meson/issues/6128#issuecomment-551734903 or https://bugreports.qt.io/plugins/servlet/mobile#issue/QTCREATORBUG-20327/comment/400818

nirbheek: Ah, interesting, thanks :)

PR #6153, opened 2019-11-08
Raise an exception if we cannot detect the MSVC compiler target architecture

lb90:
Discussed in #6128

PR #6760, opened 2020-03-09
Improve code for MSVC linker target arch detection

lb90:
While the code for the compiler received some fixes,
the code for the linker did not. Apply the fixes to
the linker counterpart.

Applies #5852, #6153 and #6759 to the linker arch
detection


lb90:
This needs some work.

jpakkane:
As we have a lot of clutter on our MR page, I'm going to close this. Feel free to reopen once you have wo new version. Thanks.

PR #6761, opened 2020-03-09
pkgconfig: Document conflicts keyword

Finii:
[why]
Sometimes one want to set the 'Conflicts:' field in .pc files.

This is possible by using the 'conflicts' keyword argument in the
pkgconfig module. The feature is not documented on mesonbuild.org,
though.

But a warning is issued:
  WARNING: Passed invalid keyword argument "conflicts".
  WARNING: This will become a hard error in the future.

History:

It has been added along with kwarg 'url' with commit
  309041918 pkgconfig: Add missing 'URL' and 'Conflicts' entries

Later the kwargs check has been introduced with
  80d665e8d Converted some modules.
but both 'url' and 'conflicts' were missing.

With commit
   2acf737b pkgconfig: Document url keyword
the 'url' kwarg has been added to the checks, but not 'conflicts'.

[how]
Add 'conflicts' to the allowed kwargs.

Signed-off-by: Fini Jastrow <ulf.fini.jastrow@desy.de>

### DIFF
--- a/docs/markdown/Pkgconfig-module.md
+++ b/docs/markdown/Pkgconfig-module.md
@@ -58,6 +58,7 @@ keyword arguments.
   generating the uninstalled pkg-config file. Since *0.54.0*
 - `dataonly` field. (*since 0.54.0*) this is used for architecture-independent
    pkg-config files in projects which also have architecture-dependent outputs.
+- `conflicts` (*since 0.36.0, incorrectly issued a warning prior to 0.54.0*) list of strings to be put in the `Conflicts` field.
 
 Since 0.46 a `StaticLibrary` or `SharedLibrary` object can optionally be passed
 as first positional argument. If one is provided a default value will be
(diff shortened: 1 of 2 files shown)


docs/markdown/Pkgconfig-module.md
```diff
@@ -58,6 +58,7 @@ keyword arguments.
   generating the uninstalled pkg-config file. Since *0.54.0*
 - `dataonly` field. (*since 0.54.0*) this is used for architecture-independent
    pkg-config files in projects which also have architecture-dependent outputs.
+- `conflicts` (*since 0.36.0*) list of strings to be put in the `Conflicts` field.
```
nirbheek: We should also note that prior to 0.54, Meson was incorrectly emitting a warning when this kwarg was being used.

Finii: Added. Force push.

xclaesse:
Nice catch, thanks for the fix!

Issue #6726, opened 2020-03-02
All files report 0% coverage


vedranmiletic:
**Describe the bug**
All files have 0% coverage after running `ninja coverage`.

**To Reproduce**
Take spdlog for example.

```
git clone https://github.com/gabime/spdlog.git
cd spdlog
meson -Dlibrary_type=shared -Denable_benchmarks=true -Denable_tests_ho=true -Db_coverage=true builddir
cd builddir
ninja
ninja coverage
```

**Expected behavior**
Nonzero percents since some tests exist.

**system parameters**
* plain native build
* Manjaro stable
* Python version 3.8.1
* meson 0.53.1
* ninja 1.10.0

Note: If I'm doing something wrong and this causes percents to remain at zero, please correct me and I'll update the relevant documentation to clarify how it should be done.

jpakkane:
Are you running the tests with `ninja test` before generating the coverage report?

vedranmiletic:
> Are you running the tests with `ninja test` before generating the coverage report?

See, that's why I put the note at the end. Thanks, I'll submit a pull request with the documentation update soonish.

michaelbadcrumble:
Hey @vedranmiletic as a result of this issue I just wrote a tip to help other users from making the same mistake, adding this to my notebook, thanks.

## [***Meson Tip***] Proper usage of `ninja coverage`

Have you ever had all coverage results return with 0% after
running `ninja coverage`?

Why?  Will if you just run coverage without running test there will
not be any data for coverage reports, like the example below first
run tests then generate coverage results.

```console
meson setup -Db_coverage=true builddir
meson compile -C builddir
meson test -C builddir
ninja coverage -C builddir
```

--------

Hopefully, you have coverage and let me know if this helped,
thanks.

vedranmiletic:
> Hey @vedranmiletic as a result of this issue I just wrote a tip to help other users from making the same mistake, adding this to my notebook, thanks.
> ## [_**Meson Tip**_] Proper usage of `ninja coverage`
> 
> Have you ever had all coverage results return with 0% after
> running `ninja coverage`?
> 
> Why? Will if you just run coverage without running test there will
> not be any data for coverage reports, like the example below first
> run tests then generate coverage results.
> 
> ```
> meson setup -Db_coverage=true builddir
> meson compile -C builddir
> ninja test -C builddir
> ninja coverage -C builddir
> ```
> 
> Hopefully, you have coverage and let me know if this helped,
> thanks.

[...]

PR #6763, opened 2020-03-09
Document the requirement to run tests before coverage

vedranmiletic:
Should help new users in situations like #6726.


jpakkane:
Thanks, just note in the future that doc-only changes should have `[skip ci]` at the end of their commit messages so CI is not run.

vedranmiletic:
> Thanks, just note in the future that doc-only changes should have `[skip ci]` at the end of their commit messages so CI is not run.

Thanks. Indeed, should have remembered to do that, I think you already told me to do this last time.

vedranmiletic:
@jpakkane any ETA when will these changes be visible on mesonbuild.com?

PR #6764, opened 2020-03-10
Add Glorytun in Users.md

angt:


jpakkane:
Thanks, but do note in the future that doc-only changes should have `[skip ci]` in their commit message so CI is not run.

PR #6765, opened 2020-03-10
mintro: AST JSON printer

mensinda:
Adds the `--ast` command to `meson introspect` to dump the raw meson AST as JSON.

### DIFF
--- a/docs/markdown/IDE-integration.md
+++ b/docs/markdown/IDE-integration.md
@@ -29,16 +29,16 @@ watch for changes in this directory to know when something changed.
 
 The `meson-info` directory should contain the following files:
 
-| File                            | Description                                                         |
-| ----                            | -----------                                                         |
-| `intro-benchmarks.json`         | Lists all benchmarks                                                |
-| `intro-buildoptions.json`       | Contains a full list of meson configuration options for the project |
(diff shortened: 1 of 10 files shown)


mensinda:
The both linter CI's are failing on checkout (for whatever reason) and I don't know why OpenSUSE didn't start yet.

mensinda:
Ping. The CI error seems to be unrelated.

jpakkane:
What is the eventual use case for this?

mensinda:
This was discussed in the IRC a while back (I think around December 2019). Essentially someone wants to do a more advanced analysis of meson files (which options influence which dependency, etc.), than what we are willing to provide in the introspection API.

Also, this might be useful for IDE's to do better syntax highlighting, since they now no longer have to implement there own lexer + parser for meson.

mensinda:
Updated the version to 0.55.0

mensinda:
ping :)

jpakkane:
The documentation should say that the format is not guaranteed to be stable.

mensinda:
Rebased and updated the docs.

mensinda:
The CI failure is unrelated (the same happens on master)

PR #6766, opened 2020-03-10
WIP: Simplify module API

xclaesse:
Module API has a few issues:
- We can call any method not starting with underscore from meson.build.
While it is good practice, it is also little known trick. Over time a
few private methods have been added without underscore (e.g.
QtBaseModule.parse_qrc). Prevent this by making ExtensionModule a
subclass of InterpreterObject which has a 'methods' dictionary to make
its public API explicit. This is also more consistent with the rest of
Meson design.
- 'snippets' bypass the check that modules should not modify
internal state. It is also little known trick often misunderstood (e.g.
fs module has 'generate_dub_file' for no reason). It is useless because
it just pass interpreter object as argument which is already available as
attribute on ExtensionModule base class. This removes snippets methods
special case.
- ModuleReturnValue is rarely needed, we can just check the type of the
return value and special case them and allow all methods to directly
return a any type that we can holderify.
- Some modules export some InterpreterObject (e.g.
CMakeSubprojectHolder) which feels inconsistent because their main
ExtensionModule object is not an InterpreterObject (i.e. not using
self.methods dict) and they don't have the 'state' argument which is
essential to keep modules behind a common API. Fix this by adding
ModuleObject, subclass of InterpreterObject, that can be used for every
objects exported by modules, not only the main ExtensionModule.

--
I only ported the FS module as an example, until we agree on the API. All modules will need to be ported afterward.

### DIFF
--- a/mesonbuild/interpreter.py
+++ b/mesonbuild/interpreter.py
@@ -31,7 +31,6 @@ from .interpreterbase import InterpreterException, InvalidArguments, InvalidCode
 from .interpreterbase import InterpreterObject, MutableInterpreterObject, Disabler, disablerIfNotFound
 from .interpreterbase import FeatureNew, FeatureDeprecated, FeatureNewKwargs
 from .interpreterbase import ObjectHolder
-from .modules import ModuleReturnValue
 from .cmake import CMakeInterpreter
 
 from pathlib import Path, PurePath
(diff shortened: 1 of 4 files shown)


jpakkane:
This has unfortunately been a victim of priorisation. However on a general level any and all work on improving the module API is welcome. It was never really designed because there was no clear requirements on what it should actually do.

xclaesse:
Created a new PR: https://github.com/mesonbuild/meson/pull/8460.

PR #6770, opened 2020-03-11
interpreter: Correctly ignore def files in build directory

xclaesse:


xclaesse:
TIL: With python < 3.6 Path().resolve() raise exception if the path does not exist. Good that we have CI with python 3.5 :-)

jpakkane:
Strange.

```
Succeeded test: test cases/common/132 build by default targets in tests
ninja: error: build.ninja:148: multiple rules generate ../test cases/common/134 include order/ctsub/main.h [-w dupbuild=err]
```

xclaesse:
I guess it's adding `main.h` twice in `self.build_def_files` because `os.path.normpath(f.relative_name())` and `f.as_posix()` are not returning the same string for the same file.

xclaesse:
Yep, that was it, it added both `ctsub/main.h` and `ctsub\main.h`, didn't realize normpath() does not the same as as_posix(). Fixed now.

jpakkane:
This needs a test.

xclaesse:
@jpakkane test added. btw see https://gitlab.freedesktop.org/gstreamer/gst-build/-/issues/85 for the context, the bug happens in really specific case.

nirbheek:
Tests are failing, not sure if the failures are relevant.

mensinda:
The two Azure failures seem relevant (`RuntimeError: No vcxproj matching '<TargetName>clean</TargetName>' in 'D:\\a\\1\\s\\tmpjuux23ie\\srctree\\_build'`), the OS Comp tests fail on checkout so maybe a rebase can fix this?

xclaesse:
@mensinda @nirbheek I think remaining CI failure is irrelevant.

mensinda:
Yes, Azure is such a pain since it is constantly updating and breaking stuff.

mesonbuild/interpreter.py
```diff
@@ -2422,14 +2422,14 @@ def add_build_def_file(self, f):
...
-                f = rel_path
-            elif not mesonlib.relpath(f, start=builddir).startswith('..'):
+            srcdir = Path(self.environment.get_source_dir())
+            builddir = Path(self.environment.get_build_dir())
+            f = Path(f).resolve()
+            if builddir in f.parents:
                 return
+            if srcdir in f.parents:
+                f = f.relative_to(srcdir)
+            f = os.path.normpath(f.as_posix())
```
nirbheek, line 2432: Why do we need `normpath()`? Didn't we already call `resolve()` which is a superset of that?

xclaesse: because pathlib and normpath does not normalize the same way, so I was getting duplicates in the build_def_files list. IIRC that was causing CI of this PR to fail on Windows.

xclaesse: See the doc:
- as_posix: Return a string representation of the path with forward slashes (/)
- normpath: On Windows, it converts forward slashes to backward slashes

nirbheek: Ah, but then why use `f.as_posix()` at all? `str(f)` will always use the correct path separator.

xclaesse: ah, didn't though about `str(f)`, I updated my patch, let's see if that pass CI.

PR #6772, opened 2020-03-11
docs: Document that ldc now supports linker overriding [skip ci]

dcbaker:
I forgot to put this in the d linker cleanups, as I implemented it.


dcbaker:
I have a patch to turn this on for gdc as well, but I need to write tests for that. For dmd I have patches, but they aren't quite working yet, and they're crazy.

Issue #5814, opened 2019-08-19
‘optimization’ affects buildtype


kloetzl:
Here is some behavior which I find very confusing.

    project('helloworld', 'c', version : '0.2', license : 'GPL3',
        default_options : [
            'buildtype=release',
            'optimization=2',
            'b_ndebug=if-release'
        ]
    )

Setting `optimization` to 2 disables the `release` buildtype. Even setting it explicitly doesn't help.

    % meson setup --buildtype=release dir
    …
    Option buildtype is: custom [default: release]

This also breaks the `if-release` feature. I think the above configuration simply should result in a `release` buildtype with `-O2` instead of `-O3`.

jpakkane:
The semantics of buildtype are roughly "if release, set options x, y and z to values a, b and c". You should either set all individual options yourself or just set the build type.

This is due to backwards compatibility. If we were to design this from scratch, buildtype would not be a standalone option, just a shortcut to set the other options.

kloetzl:
Thank you for looking into this. I think there are multiple facets to the issue.

> The semantics of buildtype are roughly "if release, set options x, y and z to values a, b and c".
> 
> If we were to design this from scratch, buildtype would not be a standalone option, just a shortcut to set the other options.

Well, the behavior you are describing is a shortcut, just as it should be in my opinion. If release use `-O3`, unless overridden. Above, I want to change `-O3` to `-O2`.

Now the weird thing is that setting `optimization` changes the buildtype. I don't see why it should. Furthermore, note that all these options are *default* options. Thus configuring the build via the commandline should take precedence. Interestingly, `--buildtype=release` doesn't work, but `-D optimization=3` does.

chuling:
Agreed, I just want to change the optimization level from O3 to O2, but the buildtype 
will be changed to custom, not keep release.

I think that maybe other opts should affect buildtype, but optimization should not, it's very special.

jpakkane:
As said, that is due to backwards compatibility. But if at all possible do not depend on buildtype, but instead look at specific settings instead.

bonzini:
This works these days

```
  buildtype                      release                          [plain, debug, debugoptimized,   Build type to use                                            
                                                                   release, minsize, custom]                                                                    
  optimization                   2                                [plain, 0, g, 1, 2, 3, s]        Optimization level                                           
```

PR #6774, opened 2020-03-12
RFC: Deprecate get_option('buildtype'), use optimization/debug instead

nirbheek:
This was long overdue.

Closes https://github.com/mesonbuild/meson/issues/5814

----

Continuing discussion from https://github.com/mesonbuild/meson/pull/6753 with @xclaesse 

> IMHO, if we want to deprecated `get_option('buildtype')` then we should deprecate that option completely, from command line too. But it is so common to check buildtype option in every project, I feel it's going to cause more troubles than needed.

I think having it available on the command-line as a short-cut for the more verbose optimization and debug options is a good thing, and it's a good way to guide people towards good 'choices' for combinations of those two options. I wouldn't really mind if it was deprecated too, but I see no compelling reason for it.

On the other hand, every single user of `get_option('buildtype')` is wrong.

F.ex., glib uses it to check when to disable cast checks by comparing it with the string 'release', which is going to be completely wrong when you're cross-compiling to embedded, where you'll use `-Doptimization=s` (`'minsize'`) or `2` (`'custom'`) and not `-Doptimization=3` (which generates larger code).

Glib also uses it to check when debugging is enabled to enable some debug infrastructure, which is very wrong, but that code was written before the debug option was available and it has been copied everywhere.

### DIFF
--- a/docs/markdown/Builtin-options.md
+++ b/docs/markdown/Builtin-options.md
@@ -103,7 +103,10 @@ two-way mapping:
 | release        | false | 3            |
 | minsize        | true  | s            |
 
-All other combinations of `debug` and `optimization` set `buildtype` to `'custom'`.
+All other combinations of `debug` and `optimization` set `buildtype` to
+`'custom'`. Hence, inside your build files when using `get_option()`, you
+should always fetch `optimization` and/or `debug`. *Since 0.54*,
+`get_option('builtype')` is deprecated.
 
 ## Base options
 
--- a/docs/markdown/snippets/deprecate_get_option_buildtype.md
+++ b/docs/markdown/snippets/deprecate_get_option_buildtype.md
@@ -1,0 +1,35 @@
+## `get_option('buildtype')` is deprecated by `optimization` and `debug`
+
+Build files have been doing anti-patterns such as checking whether debugging is
+enabled with `get_option('buildtype').startswith('debug')`. This has been
+incorrect *[since 0.48]* where we added separate options for `-Doptimization`
+(choice) and `-Ddebug` (boolean) which also set `buildtype` for
+backward-compatibility.
+
+However, many [combinations of these two options](Builtin-options.html#build-type-options)
+such as `-Doptimization=g -Ddebug=true` actually set **`buildtype='custom'`**.
+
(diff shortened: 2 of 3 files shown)


nirbheek:
Another reason to not deprecate `-Dbuildtype` is because `b_vscrt`'s default value is `from_buildtype`, which derives from that. It would be a regression to derive from `-Ddebug` because it's fairly common to link to the release CRT while having debugging enabled. F.ex., you need to do this when linking to MinGW to avoid cross-CRT issues.

xclaesse:
Regarding GLib, I know ebassi has really strong opinion about buildtype: https://bugzilla.gnome.org/show_bug.cgi?id=794790.

nirbheek:
> Regarding GLib, I know ebassi has really strong opinion about buildtype: https://bugzilla.gnome.org/show_bug.cgi?id=794790.

I'll be opening a merge request to glib, let's see how that goes. I understand what he's saying, and I think my changes will be compatible with that.

nirbheek:
Ah, also, `b_ndebug` has `if-release` as a possible value, so we cannot deprecate `-Dbuildtype` entirely either. Ideally this should be deprecated too. I think the idea of `release` is fundamentally broken right now. You don't want `-Doptimization=3` for release builds on _all_ platforms. Maybe we should just define that `optimization` in `['2', '3', 's']` will set `buildtype=release`.

nirbheek:
Hmm `b_vscrt=from_buildtype` and `b_ndebug=if-release`  are also broken in a specific way: they're dependent on `buildtype`, but the mapping is not clear, and the build file usually wants to know the _actual_ value that will be used.

nirbheek:
So there's a couple of problems here:

1. People are using `get_option('buildtype').startswith('debug')` instead of `get_option('debug')`, which is completely wrong ever since `-Ddebug` was added (years ago). We should warn about this case.

2. `buildtype=release` has the semantic meaning 'release build' (f.ex., `b_ndebug=if-release`), and build files are using that to enable/disable features. But that's broken when you want to set `-Doptimization` to anything other than `3` because build files will get `custom`. We need a fix for this: maybe we should set `buildtype=release` when `-Doptimization={2,3,s}`.

PR #6778, opened 2020-03-12
Improve usability of get_option('buildtype')

nirbheek:
Alternative to https://github.com/mesonbuild/meson/pull/6774.

commit a18658d8e11ed3504aa700df6a9e78a3cd9460f4:

```
coredata: Set buildtype=release from others in more cases
`buildtype=release` has the semantic meaning 'release build' (f.ex.,
`b_ndebug=if-release`), and build files are using that to enable/disable
features.

However, that's broken when you want to set `-Doptimization` to
anything other than `3` for a 'release build', for example when
shipping for embedded or mobile, because build files will get
`custom`.

Setting `buildtype` to `release` when `optimization` is `2` or `s`
(and `debug=false`) is more useful than setting it to `custom`.
```

commit 3348a0fd1a09ab0768f487661ad1eabe295a6725:

```
Warn when detecting debug with get_option('buildtype')
Build files should always use `get_option('debug')` since `buildtype`
is set to `custom` in many cases.
```

### DIFF
--- a/docs/markdown/Builtin-options.md
+++ b/docs/markdown/Builtin-options.md
@@ -91,19 +91,31 @@ Using the option as-is with no prefix affects all machines. For example:
 For setting optimization levels and toggling debug, you can either set the
 `buildtype` option, or you can set the `optimization` and `debug` options which
 give finer control over the same. Whichever you decide to use, the other will
-be deduced from it. For example, `-Dbuildtype=debugoptimized` is the same as
+be deduced from it. For example, `-Dbuildtype=debugoptimized` will also set
 `-Ddebug=true -Doptimization=2` and vice-versa. This table documents the
-two-way mapping:
+mapping from `buildtype` to `debug` and `optimization`:
 
 | buildtype      | debug | optimization |
 | ---------      | ----- | ------------ |
-| plain          | false | 0            |
 | debug          | true  | 0            |
+| minsize        | true  | s            |
 | debugoptimized | true  | 2            |
 | release        | false | 3            |
-| minsize        | true  | s            |
+| plain          | false | 0            |
 
-All other combinations of `debug` and `optimization` set `buildtype` to `'custom'`.
+This table documents the reverse mapping:
+
+| debug | optimization | buildtype      |
+| ----- | ------------ | ---------      |
+| true  | 0            | debug          |
+| true  | s            | minsize        |
+| true  | 2            | debugoptimized |
+| false | 0            | plain          |
+| false | s            | release *(since 0.54)* |
+| false | 2            | release *(since 0.54)* |
+| false | 3            | release        |
+
+All other combinations of `debug` and `optimization` set `buildtype` to `custom`.
 
 ## Base options
 
--- a/mesonbuild/interpreter.py
+++ b/mesonbuild/interpreter.py
@@ -2833,7 +2833,7 @@ external dependencies (including libraries) must go to "dependencies".''')
 
     @stringArgs
     @noKwargs
-    def func_get_option(self, nodes, args, kwargs):
+    def func_get_option(self, node, args, kwargs):
         if len(args) != 1:
             raise InterpreterException('Argument required for get_option.')
         optname = args[0]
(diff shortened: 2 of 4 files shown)


jpakkane:
I have thought about this and how about this alternative approach:

- Keep the option name as currently.
- Setting it to a new value sets the corresponding values of opt, debug, etc
- Setting opt or debug value does _not_ change the builtype do custom
- Instead, we print a warning (at configure time?) if the option values do not match the buildtype

nirbheek:
I'm not sure, it seems like the user will be stuck with that behaviour and not know what to do. If they want to pass `-Doptimization=s` because they're targeting mobile, should they not use `-Doptimization` at all? Should they pass the appropriate flags through `c_args` manually?

jpakkane:
This is a conceptual problem, really. What do we want `buildtype` to _actually_ mean?

- a specific set of option values that are always enforced to specific values
- a logical intent specifying what the the user intends to use the build for, but which does not mandate flags
- a shortcut for setting some options to specific values and going to custom if they are changed (the current approach)

Thus at least the following behaviours could reasonably be considered:

- setting buildtype sets the opt/debug/etc options and changing them changes the type to custom
- setting buildtype sets the opt/debug/etc options but changing them to other values does not change the buildtype to custom
- setting buildtype sets the opt/debug/etc options and trying to change them to other values is blocked

The problem with the last one is that release builds could reasonably use either `O2` or `O3`. It seems that this would need to be settable somehow.

xclaesse:
I've been thinking about this issue and I think I like the idea of deprecating `get_option('buildtype')`. I think `--buildtype release` is a nice shorthand for users who just want a build that they can deploy, with sane defaults. But build definition should not rely on it and instead use debug/optimization pair, like @nirbheek did in GLib: https://gitlab.gnome.org/GNOME/glib/-/blob/master/meson.build#L225.

xclaesse:
I mean we could deprecate `buildtype` completely and just keep `--buildtype` and `-Dbuildtype` as command line shorthand.

docs/markdown/Builtin-options.md
```diff
@@ -91,19 +91,31 @@ Using the option as-is with no prefix affects all machines. For example:
...
 | ---------      | ----- | ------------ |
-| plain          | false | 0            |
 | debug          | true  | 0            |
+| minsize        | true  | s            |
 | debugoptimized | true  | 2            |
 | release        | false | 3            |
-| minsize        | true  | s            |
+| plain          | false | 0            |
+
+This table documents the reverse mapping:
```
xclaesse, line 106: I don't think the reverse mapping is needed to document if we deprecate `get_option('buildtype')`.

jpakkane:
If we remove `buildtype` then that removes a semantic meaning of debugness. For example one might want to add extra checks and loggings in a debug ("development") build. You can't really get the same thing by querying `debug` as it is almost certainly enabled for release builds as well.

Is this an issue in practice? I don't really know. This is the downside of not having a sales/marketing department whose job it is to find out what users actually want and need.

mesonbuild/interpreter.py
```diff
@@ -2843,9 +2843,17 @@ def func_get_option(self, nodes, args, kwargs):
...
         elif isinstance(opt, coredata.UserOption):
-            return opt.value
+            opt = opt.value
+        if optname == 'buildtype':
+            class BuildTypeString(str):
+                def startswith(self, token):
+                    if token == 'debug':
+                        mlog.warning("Use get_option('debug') to check whether debugging is enabled "
+                                     "instead of get_option('buildtype').startswith('debug')", location=node)
+                    super().startswith(token)
```
xclaesse, line 2855: I think if we are going to deprecate some usage of buildtype, better deprecate it completely. That's more consistent.

jpakkane:
That Glib check is actually pretty much what I was talking about. Really you'd want to enable or disable things based on _what the build will be used for_ rather than _what optimization setting is set_.

But as usual if you always want to use the features based on optimization, then adding an extra option is pointless and wasteful.

xclaesse:
> If we remove `buildtype` then that removes a semantic meaning of debugness. For example one might want to add extra checks and loggings in a debug ("development") build. You can't really get the same thing by querying `debug` as it is almost certainly enabled for release builds as well.

I think in that case your project should add its own option. For example GStreamer has glib-checks, glib-asserts and extra-checks to have fine grained levels of debuggability.

PR #6780, opened 2020-03-12
interpreterbase: Add warning when built-in object method has no kwargs

xclaesse:


jpakkane:
This would work better as decorator, like the one already in `interpreterbase` (though that one hard errors, so it can't be used here yet).

xclaesse:
@jpakkane done.

jpakkane:
Test failure seems relevant:

```
mesonbuild/ast/interpreter.py:319:43: error: Argument 1 to "flatten_kwargs" of "AstInterpreter" has incompatible type "Dict[BaseNode, BaseNode]"; expected "Dict[str, Union[Union[Union[str, int, float, bool], List[Any], Dict[Any, Any], InterpreterObject, ObjectHolder[Any]], BaseNode]]"
Found 1 error in 1 file (checked 27 source files)
```

xclaesse:
@jpakkane remaining CI failure are not relevant, afaik.

jpakkane:
Indeed.

Issue #2155, opened 2017-08-03
Duplicate include directories


zajo:
The way I've setup my projects I get a lot of duplicate -I options passed to the compiler. This seems wrong, is it possible for the include_directories support to delete the duplicates?

jpakkane:
We should be doing that already. Which version of Meson are you using?

zajo:
0.41.2. However, I incorrectly reported that it is for the -I options, it is for the -isystem.

zajo:
I tried to fix this myself but it seems it isn't trivial. I tried adding '-isystem' to dedup2_prefixes in compilers.py, but I see that later in _can_dedup it won't dedup anything that is defined by what comes after it (per the comment at lines 472-479 in 0.42.0 distro). So I'm still stuck with tons of -isystem duplicates.

nirbheek:
See also https://github.com/mesonbuild/meson/issues/2156. We might have to implement that mechanism because currently we only support deduping individual arguments, not argument-value pairs.

zajo:
Any development on this issue? I'm thinking removing duplicate `-I` and `-isystem` can even be done at the very end, even after the command line has been converted to a string?

nirbheek:
No progress so far, would you like to work on this? I can give you the relevant context if you wish.

zajo:
I looked into this some time ago, and figured that the normal mechanisms for coping with duplicate options don't work for command line args that read the next arg, like `-I` or `-isystem`. Could you point me to the code that composes the final command line for `C` or `C++` compilers? I think I'd be able to figure it out from there.

nirbheek:
You will want to change how `CompilerArgs()` in `compilers/compilers.py` does dedup by making it read the next argument too.

Compiler arguments are composed for C/C++ compilers in `_generate_single_compile()` in `mesonbuild/backend/ninjabackend.py`.

ClausKlein:
Does this **de-duplicate** realy work?

I tested with a something like that:
```meson
jsoncpp_include_directories = include_directories(['include', 'dir1', 'dir2/', 'dir2/../dir1', 'dir1/../dir1', 'dir2', 'dir1/', 'include'])
````
If I understand the code right, cls.dedup2_prefixes are never de-duplicated.

```patch
bash-5.0$ git diff
diff --git a/mesonbuild/compilers/compilers.py b/mesonbuild/compilers/compilers.py
index e13256e8..5867620a 100644
--- a/mesonbuild/compilers/compilers.py
[...]

PR #6782, opened 2020-03-13
Use os.path.normpath() for include paths

ClausKlein:
This make relative pathes shorter an too give a chance to
de-duplicate -isystem flags just like -I flags.

Fix common test case 203 for OSX build host too.

Related to https://github.com/mesonbuild/meson/issues/2155

### DIFF
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -2013,7 +2013,6 @@ https://gcc.gnu.org/bugzilla/show_bug.cgi?id=47485'''))
             # own target build dir.
             if not isinstance(i, (build.CustomTarget, build.CustomTargetIndex)):
                 continue
-            # NOTE: if we want to use os.path.realpath() for include path only! CK
             idir = os.path.normpath(self.get_target_dir(i))
             if not idir:
                 idir = '.'
@@ -2028,11 +2027,9 @@ https://gcc.gnu.org/bugzilla/show_bug.cgi?id=47485'''))
     def generate_inc_dir(self, compiler, d, basedir, is_system):
         # Avoid superfluous '/.' at the end of paths when d is '.'
         if d not in ('', '.'):
-            # NOTE: if we want to use os.path.realpath() for include path only! CK
             expdir = os.path.normpath(os.path.join(basedir, d))
         else:
             expdir = basedir
-        # NOTE: if we want to use os.path.realpath() for include path only! CK
         srctreedir = os.path.normpath(os.path.join(self.build_to_src, expdir))
         sargs = compiler.get_include_args(srctreedir, is_system)
         # There may be include dirs where a build directory has not been
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -2013,7 +2013,8 @@ https://gcc.gnu.org/bugzilla/show_bug.cgi?id=47485'''))
             # own target build dir.
             if not isinstance(i, (build.CustomTarget, build.CustomTargetIndex)):
                 continue
-            idir = self.get_target_dir(i)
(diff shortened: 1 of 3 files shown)


ClausKlein:
see too the https://github.com/ninja-build/ninja/issues/1251#issuecomment-598684012

mesonbuild/mesonlib.py
```diff
@@ -270,8 +270,10 @@ def from_absolute_file(fname: str) -> 'File':
     def rel_to_builddir(self, build_to_src: str) -> str:
         if self.is_built:
             return self.relative_name()
+            #NO! return os.path.realpath(self.relative_name())
         else:
             return os.path.join(build_to_src, self.subdir, self.fname)
+            #NO! return os.path.realpath(os.path.join(build_to_src, self.subdir, self.fname))
```
jpakkane: What are these comments for? I don't understand what they try to say?

test cases/common/203 function attributes/meson.build
```diff
@@ -65,6 +64,7 @@ endif
 if host_machine.system() != 'darwin'
   attributes += 'alias'
   attributes += 'visibility'
+  attributes += 'alloc_size'
```
jpakkane: This seems like an unrelated change, it should be in its own MR.

mesonbuild/backend/ninjabackend.py
```diff
@@ -2013,7 +2013,8 @@ def get_custom_target_dir_include_args(self, target, compiler):
             # own target build dir.
             if not isinstance(i, (build.CustomTarget, build.CustomTargetIndex)):
                 continue
-            idir = self.get_target_dir(i)
+            # NOTE: if we want to use os.path.realpath() for include path only! CK
+            idir = os.path.normpath(self.get_target_dir(i))
```
jpakkane: The comment does not make sense. Also, please don't add initials or any such tags to the code. They are just noise, that's what `git blame` is for. The same goes for the other two comments below

It would probably also be better to make `get_target_dir` return normalised paths. This way it fixes the issue for all call sites.

mesonbuild/backend/ninjabackend.py
```diff
@@ -2013,7 +2013,8 @@ def get_custom_target_dir_include_args(self, target, compiler):
             # own target build dir.
             if not isinstance(i, (build.CustomTarget, build.CustomTargetIndex)):
                 continue
-            idir = self.get_target_dir(i)
+            # NOTE: if we want to use os.path.realpath() for include path only! CK
```
dcbaker: I don't understand what this comment even means, (I do know what `os.path.realpath` does). Also, please don't sign your comments, that's what `git blame` is for.

test cases/common/203 function attributes/meson.build
ClausKlein: without that, i can't run the tests

mesonbuild/mesonlib.py
ClausKlein: My intension was to use realpath() , but there is still a ninja bug.
see https://github.com/ninja-build/ninja/issues/1251#issuecomment-599541385

mesonbuild/backend/ninjabackend.py
ClausKlein: see https://github.com/ninja-build/ninja/issues/1251

mesonbuild/backend/ninjabackend.py
ClausKlein: done

mesonbuild/backend/ninjabackend.py
ClausKlein: done

mesonbuild/mesonlib.py
ClausKlein: done

jpakkane:
Thanks.

PR #6785, opened 2020-03-15
Add builtin docdir option

### LINKED ISSUE #825: Add docdir builtin option
`docdir` is defined by default to be `/usr/local/share/doc/<yourpkg>`, or in other words, `<datarootdir>/doc/<yourpkg>`. I tried to implement this myself just like `sysconfdir` was, but I'm uncertain how to do the `<yourpkg>` bit by default.

Flow-It:
This PR is supposed to close #825 

As discussed in the linked issue, there is no reliable way to extract the project part, if we wanted docdir to be `/usr/share/doc/<my_project>`. So instead of idling around, I propose to add the docdir option with the simple `/usr/share/doc` default, and let users be responsible to append the correct path to their (sub)project. I agree, the solution is not perfect, but with this PR we achieve a bit of simplification like this small example shows:
```
# previously
my_docdir = get_option('prefix') / get_option('datadir') / 'doc' / meson.project_name()
# now
my_docdir = get_option('prefix') / get_option('docdir') / meson.project_name()
```

As this is my first PR: I've seen many commit messages prefixed with the subsystem they are touching, but not all commits follow that. There is also no mention about this style in the contribution guidelines. Please let me know if the commit messages are fine or if I should rename them.


xclaesse:
I'm not a big fan of adding a built-in option not used by Meson itself, and not used by projects that don't have documentation.

What's the use-case exactly? What can't you just have a project option for this, relative to prefix?

Flow-It:
I'd say it is common enough for projects to have documentation. Hence, having a builtin docdir option is a convenience feature first. However, I see at least one more argument, and that is consistency. Having the documentation directory as a project option will probably lead to different namings for the same thing. Project A might call it docdir, project B might call it doc**s**dir, ... The same consistency argument applies when comparing to other build systems. autotools has it (although they go with the `/usr/share/doc/<my_project>` version and CMake also has the GNUInstallDirs module.

I don't know meson well enough, but are all currently existing builtin options used by meson itself? I doubt that.

jpakkane:
Regardless you don't need to add prefix, because relative install paths are interpreted relative to prefix anyway. So this works:

```meson
my_docdir = get_option('datadir') / 'doc' / meson.project_name()
```

There is a bit of a problem with subprojects, though. This option should have different value for every subproject, otherwise the docs for subprojects end up in the master project's doc dir. We currently don't do subproject specific options.

Flow-It:
Thanks for the hint with `prefix`. It doesn't affect my arguments though. So, what is the plan going forward? Are subproject-specific options planned and we should stall this PR until then, or is a `docdir` option not wanted at all and we can close it for good?

jpakkane:
Maybe we could add a `docrootdir`, which would by default point to `share/doc`? Then people could do stuff like:

```meson
install_XXX(... install_dir: get_option('docrootdir') / 'myproj')
```

Would that be useful? It would work without problems with subprojects at least.

eli-schwartz:
I see no advantage to having a docdir setting if it is going to require manual fixups anyway. If it defaulted to `get_option('datadir') / 'doc' / meson.project_name()` and applied per-subproject, projects could modify it in `project(default_options: ...)` if they really needed to.

I'd say stall it until then.

Flow-It:
Calling it `docrootdir` feels inconsistent, because we use `datadir` instead of `datarootdir` as well. I guess the option is not that important to justify merging a half-baked solution. So following all the input given here it's probably good to revisit this PR when a decision regarding subproject-specific options has been made.

jpakkane:
Ok, let's close this for now to reduce clutter on the MR page.

PR #6787, opened 2020-03-15
Generators can depend on custom targets as well as build targets.

jpakkane:
There's no reason not to, this is just an oversight.

### DIFF
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -1369,7 +1369,7 @@ class Generator:
         if 'depends' in kwargs:
             depends = unholder(listify(kwargs['depends']))
             for d in depends:
-                if not isinstance(d, BuildTarget):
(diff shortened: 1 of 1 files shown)


jpakkane:
I always forget that that is a thing you can do. 🤦‍♂ Fixed.

mesonbuild/build.py
```diff
@@ -1369,7 +1369,7 @@ def process_kwargs(self, kwargs):
         if 'depends' in kwargs:
             depends = unholder(listify(kwargs['depends']))
             for d in depends:
-                if not isinstance(d, BuildTarget):
+                if not (isinstance(d, BuildTarget, CustomTarget)):
```
obilaniu: You put the first parenthesis at the wrong place; I recommended `not isinstance(d, (BuildTarget, CustomTarget))`

jpakkane:
Test failure is unrelated.

PR #6793, opened 2020-03-18
Fix exe_wrapper usage in EmptyExternalProgram case

richardweinberger:
If no exe_wrapper is set in the meson cross file the exe_wrapper
object will be an instance of EmptyExternalProgram.
So, found is True and prorgram is an empty list.

This will cause meson to tun the compiler sanity check because
it checks only for self.is_cross and self.exe_wrapper being
not None.

I ran into that situation while cross compiling for ia32 on a
x64_64 host. The host had no ia32 userspace installed, so the
self test failed.

As workaround I currently set exe_wrapper to 'true'.

Signed-off-by: Richard Weinberger <richard@nod.at>


jpakkane:
Thanks.

PR #6795, opened 2020-03-18
Make werror per subproject option

xclaesse:


test cases/common/229 persubproject options/subprojects/sub2/foo.c
```diff
@@ -1,5 +1,9 @@
 int foo(void);
 
+#ifdef __GNUC__
```
dcbaker: Doesn't this mean the test only actually works on glibc?

xclaesse: The CI fails with MSVC that doesn't support `#warning`. Since this change is compiler agnostic, it doesn't matter, IMO.

dcbaker: I guess thats fair

jpakkane:
Given that the way subproject options work is still not worked out (see f.e. #6595) it's not really nice to add even more of these in the mix.

xclaesse:
I'm sorry to fix issues, should I stop? Could you please just close #6595, it should not get merged.

xclaesse:
Failed CI seems unrelated.

@jpakkane can this get a freeze exception since it's popular request and is a safe change now that per-subproject default_library is accepted. @dcbaker already approved the implementation.

PR #6798, opened 2020-03-18
Site-wide native file configuration

dcbaker:
This allows a special native file called `siteconfig.ini` to be placed
in a user or system wide location on Unix systems for default
configuration. My initial thought of it's usefulness is to allow distros
that use non standard lib locations to set those system wide in
configuration rather than us having to fix up libdir paths.

Ideally this would be followed up by a similar cross configuration
setting, but I'm unsure of how to do that, given the complexities of
cross compilation.

### DIFF
--- a/docs/markdown/Native-environments.md
+++ b/docs/markdown/Native-environments.md
@@ -25,6 +25,30 @@ persistent environment:
 All of the rules about cross files and changed settings apply to native files
 as well, see [here](Cross-compilation.md#changing-cross-file-settings)
 
+## Site local configuration
+
+*New in 0.54.0*
+
+Meson allows system administrators, OS distributions, and users to set
+default policy using a special native file, `siteconfig.ini`. This is looked
+up in all of the search paths meson searches. This applies to Unix-like OSes
+only.
+
+It is recommended that a minimal number of settings be put in this config. A
+good thing to put in this are non standard paths, for example, if a
+distribution uses Debian style libdirs (`/usr/lib/<host-triple>`) they could
+use this config for x86_64 Linux with glibc:
+
+```ini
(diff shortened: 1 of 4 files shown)


docs/markdown/Native-environments.md
```diff
@@ -25,6 +25,30 @@ persistent environment:
...
+only.
+
+It is recommended that a minimal number of settings be put in this config. A
+good thing to put in this are non standard paths, for example, if a
+distribution uses Debian style libdirs (`/usr/lib/<host-triple>`) they could
+use this config for x86_64 Linux with glibc:
+
+```ini
+[paths]
+libdir = 'x64_64-linux-gnu'
```
eli-schwartz: Shouldn't this be `lib/x64_64-linux-gnu`?

dcbaker: err, yeah it should

ePirat:
Are the meson native and crossfile actually ini files? While they have a ini-like syntax I don't think they are strictly ini files, given the arrays for example?

dcbaker:
ini files aren't formally defined anywhere. We may add additional meaning to the values, but we don't allow invalid characters in the keys or values so I'd say that they are valid ini files, as much as anything is.

ePirat:
Sure it's not really defined but for example ini files usually have no quotes for their values, while meson cross and native files enforce them. Additionally I've never seen square brackets used in any ini file to specific an array value.

I get that ini is not specified so you can call whatever you want an ini, just noting IMO it's confusing to call it .ini unless it really uses such a permissive parser as is usually used for
ini files.

xry111:
Is there any progress about this feature?

PR #6802, opened 2020-03-19
Fix failing rust CI

xclaesse:
rustc 1.42 now dynamically link to libstd so its path is needed in
LD_LIBRARY_PATH.


jpakkane:
Failure is unrelated.

PR #6363, opened 2019-12-18
Fix legacy env var support with cross

Ericson2314:
This fixes us to match Autoconf's behavior; see [the issue](https://github.com/mesonbuild/meson/issues/3969). It does however still keep it so the cross is the only way to control the host platform in cross builds.

CC @dcbaker

Fixes #3969

PR #6806, opened 2020-03-19
Naturally use env vars a bit more to match Autoconf -- contains #6363

Ericson2314:
PR #6363 makes it so our interpretation of env vars no longer clashes
with Autoconf's: if both Meson and Autoconf would read and env var, both
would do the same things with the value they read.

However, there were still cases that Autoconf would read an env var when
Meson wouldn't:

 - Autoconf would use `CC` in cross builds too

 - Autoconf would use `CC_FOR_BUILD` in native builds too.

There's no reason Meson can't also do this--if native cross files
overwrite rather than replace env vars, cross files can also overwrite
rather than replace env vars.

Because variables like `CC` are so ubiquitous, and because ignoring them
in cross builds just makes those builds liable to break (and things more
complicated in general), we bring Meson's behavior in line with
Autoconf's.

I can personally vouch having trouble keeping the cross files as comprehensive as the environment variables in https://github.com/NixOS/nixpkgs . I would love to fix that some day, but in the meantime it would be quite convenient if things that didn't make it to the cross file would work because of the environment variables.

CC @ePirat

### DIFF
--- a/run_unittests.py
+++ b/run_unittests.py
@@ -6328,11 +6328,10 @@ c = ['{0}']
         testdir = os.path.join(self.unit_test_dir, '61 identity cross')
         env = {
             'CC_FOR_BUILD': '"' + os.path.join(testdir, 'build_wrapper.py') + '"',
+            'CC': '"' + os.path.join(testdir, 'host_wrapper.py') + '"',
         }
         crossfile = tempfile.NamedTemporaryFile(mode='w')
-        crossfile.write('''[binaries]
-c = ['{0}']
-'''.format(os.path.join(testdir, 'host_wrapper.py')))
+        crossfile.write('')
         crossfile.flush()
         self.meson_cross_file = crossfile.name
         # TODO should someday be explicit about build platform only here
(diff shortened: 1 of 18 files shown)


run_unittests.py
```diff
@@ -6326,11 +6326,10 @@ def test_identity_cross_env(self):
         testdir = os.path.join(self.unit_test_dir, '61 identity cross')
         env = {
             'CC_FOR_BUILD': '"' + os.path.join(testdir, 'build_wrapper.py') + '"',
+            'CC': '"' + os.path.join(testdir, 'host_wrapper.py') + '"',
         }
         crossfile = tempfile.NamedTemporaryFile(mode='w')
-        crossfile.write('''[binaries]
-c = ['{0}']
-'''.format(os.path.join(testdir, 'host_wrapper.py')))
+        crossfile.write('')
```
Ericson2314: I don't have to change this test, as this is a non-breaking change, but doing so makes it less redundant with the previous test.

PR #6815, opened 2020-03-20
Add support for composing properties with layering

### LINKED ISSUE #6813: Is there any way to compose *_args when layering cross files?
Love the cross file layering support, has simplified a significant amount of duplication in my cross files.

I remember seeing in the documentation a comment along the lines of "the primary intention of layering is composing native and cross files, not overriding settings". In the spirit of composition, I'm wondering if there is a way to compile the compile/link flags instead of just overriding their values completely when layering?

Consider this set of flags:

```
c_args = [ '-mcpu=cortex-m4', '-mfloat-abi=hard', '-mfpu=fpv4-sp-d16', '-mabi=aapcs', '-mthumb',]
```

My ideal scenario would be to extract the hard floating point flags into a standalone file, and some how add these flags to `c_args` instead of completely overriding the previous cross files. Perhaps this is as simple in implementation as changing '=' to '+='. If something like this is reasonable, I'm happy to take a stab at the implementation.

```
c_args += ['-mfloat-abi=hard', '-mfpu=fpv4-sp-d16',]
```

With this kind of implementation, I could provide cross files that allow a user to select soft or hard FP support. You could compose a cross build by selecting from the options:

* ARM cross file base (defines toolchain, common flags)
* Cortex-M4 cross file (defines cortex-M4 specific flags, override `cpu`)
* Hard-floating point cross file (adds specific flags)

I know I can just make hard/soft floating point a build option that is only honored when `is_cross_build() == true` and the compiler's ID is `arm-gnu-none-eabi`. But I'm trying my best to keep cross-compilation details out of the build system as much as possible.

phillipjohnston:
This commit adds support for using `+=` with properties to enable better composition with native/cross file properties. This can be used for layering compiler/linker arguments as described in the documentation.

Updates have been made to the native file + cross file documentation in the user manual.

Unit tests have been added to test three different cases:

- Define a property in a base, append with another file
- Define a property in a base, overwrite with another file
- Append a property that isn't defined yet

Fixes #6813

### DIFF
--- a/docs/markdown/Cross-compilation.md
+++ b/docs/markdown/Cross-compilation.md
@@ -229,14 +229,82 @@ bindir = 'bin'
 
 This will be overwritten by any options passed on the command line.
 
+## Loading Multiple Cross Files
+
 Since meson 0.52.0 it is possible to layer cross files together. This
 works like native file layering: the purpose is to compose cross files
 together, and values from the second cross file will replace those
 from the first.
 
+### Cross File Layering Example
+
+For example, you might define a base ARM cross-compilation file:
+
+```
+[binaries]
+c = 'arm-none-eabi-gcc'
+cpp = 'arm-none-eabi-c++'
+ar = 'arm-none-eabi-ar'
+strip = 'arm-none-eabi-strip'
+
+[properties]
+needs_exe_wrapper = true
+
+[host_machine]
+system = 'none'
+cpu_family = 'arm'
+# CPU should be redefined in child cross files - this is a placeholder
+# that will be used in case a child file does not override this setting
+cpu = 'arm-generic'
+endian = 'little'
+```
+
+And layer it with a Cortex-M4 hard floating point cross-file:
+
+```
+[properties]
+c_args = [ '-mcpu=cortex-m4', '-mfloat-abi=hard', '-mfpu=fpv4-sp-d16', '-mabi=aapcs', '-mthumb',]
+c_link_args = [ '-mcpu=cortex-m4', '-mfloat-abi=hard', '-mfpu=fpv4-sp-d16', '-mabi=aapcs', '-mthumb',]
+cpp_args = [ '-mcpu=cortex-m4', '-mfloat-abi=hard', '-mfpu=fpv4-sp-d16', '-mabi=aapcs', '-mthumb',]
+cpp_link_args = [ '-mcpu=cortex-m4', '-mfloat-abi=hard', '-mfpu=fpv4-sp-d16', '-mabi=aapcs', '-mthumb',]
+
+[host_machine]
+cpu = 'cortex-m4'
+```
+
+These files can be layered by adding multiple `cross-file` arguments during the configure step:
+
+```
+$ meson buildresults --cross-file=build/cross/arm.txt --cross-file=build/cross/cortex-m4_hardfloat.txt
+```
+
+### Composing Properties With Layering
+
+Entries in the `properties` section allow *composition* by appending to a value:
+
+```
+[properties]
+my_prop += ['append_this']
+```
+
+If the property is not already defined, it will be created. You must honor the `+=` usage when layering multiple native files together. If you include three different files and the last one assigns (`=`) instead of composes (`+=`), the value will be overwritten.
+
+For example, consider a Cortex-M4 cross file which defines compilation flags:
+
+```
+[properties]
+c_args = [ '-mcpu=cortex-m4', '-mabi=aapcs', '-mthumb',]
+endian = 'little'
+```
+
+We could create additional cross-files that add the appropriate flags for either hard or soft floating-point support. Using composition (`+=`), we can ensure that these flags are *added* to previously defined values instead of *replacing* the values.
+
+```
+[properties]
+c_args += ['-mfloat-abi=hard', '-mfpu=fpv4-sp-d16']
+```
 
 ## Starting a cross build
-
 
 Once you have the cross file, starting a build is simple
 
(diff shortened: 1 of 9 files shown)


phillipjohnston:
I'm a C/C++/assembly guy primarily, so please let me know if there's something I can do better with Python.

Ericson2314:
I (and @dcbaker too if I remember) would appending be the default, and replacing as done today be the explicit opt-in.

phillipjohnston:
That would be my preference too @Ericson2314. What's the desired syntax for indicating replacement? Happy to adjust.

phillipjohnston:
Thinking on this over the weekend, one potential issue with changing the default is that we already have built-in association with how += and = operates in the Meson build file context. So we would be breaking consistency on that front. From my perspective, that introduces a point of confusion into the language.

xclaesse:
I'm not a big fan of this proposal, because those files are more or less "standard" ini file. Extending the syntax to add such operation is stretching it a lot IMHO.

That being said, I agree with the functionality. Maybe we should instead create a brand new file format for them instead, based on Meson interpreter, just like what we have for meson_options.txt. But that could be overkill...

phillipjohnston:
Hi @xclaesse,

I understand that perspective and the goal of sticking to the `.ini` format. I did not consider that before making these changes. 

I don't agree myself that "+=" is "stretching it a lot", at least compared to inventing a new file format.

In the interest of compromise while *not* creating a new file format, what if we leveraged sections instead? We could also merge in the preference of @dcbaker and @Ericson2314 to have composing be the default behavior.

Perhaps `[properties]` composes by default, and `[properties-overrides]` will override previously-specified values.

xclaesse:
I don't really have strong opinion on this tbh. concat lists when composing files is probably the good compromise, it avoids adding `+` and `+=` expressions in a ini file.

Ericson2314:
If you could write this logic as a way to union two `Properties`, I will be happy. I will want to reuse that for getting rid of `Environment.cmd_line_options` by also storing it as a Properties which is unioned with the rest CC @dcbaker.

dcbaker:
I'm also not sure this is a great idea, it's adding logic to what is an otherwise declarative syntax. I'm not sure how we'd go about solving this either. I guess one way we could do this is add a command line option like `--merge-config-files`, and then load each file one at a time and merge any keys in common, though I'm not sure that's much better. Alternatively (just throwing things at the wall), we could add a script to meson do do that merging, so you could do `meson merge-config -o merged.ini 1.ini 2.ini 3.ini` and then just use merged.ini.

What we absolutely cannot do is change the behavior to concat lists by default, that is absolutely going to break things for someone. (me being one such someone)

@Ericson2314, I'd be super happy to get rid of Environment.cmd_line_options, that code is getting pretty scary.

phillipjohnston:
@dcbaker "not a great idea" referring to the += or to the idea of using a section to indicate flags to merge/overwrite?

> What we absolutely cannot do is change the behavior to concat lists by default, that is absolutely going to break things for someone. 

Heard.

Ericson2314:
`--merge-config-files` with a deprecation cycle sounds pretty good to me.

xclaesse:
We could also just add key/value in the config file to configure the merge policy. 

```
merge_policy='append'

[properties]
c_args = ['-Dfoo']
```

xclaesse:
Or

```
[merge policy]
list = 'append'
string = 'override'
[properties]
c_args = ['-Dfoo']
```

phillipjohnston:
I like `merge policy` the best so far.

One downside to both `--merge-config-files` and `merge_policy` vs finer-grained control, whether by section or by list/string specific settings, is that we're setting a blanket behavior policy. I can definitely see cases where I might want to append to some properties, and override other ones completely. That flexibility is probably valuable to others as well.

xclaesse:
We can go meta... what's the merge policy of the merge policy?

Ericson2314:
Yeah I don't think it can go in the files themselves; that's some crazy OOP nonsense. Too many pointless edge cases to think about.

phillipjohnston:
Too bad for `+=`, the simplest solution with no edge cases :'(.

This PR has now gone beyond what I'm able to support with my Python abilities or understanding of the team's desires. Shall I close this PR and file an issue to capture the conversation? Or does somebody want to commandeer this?

phillipjohnston:
I have one last pitch for a simple approach like `+=` vs `=`.

When adding `+` to the left side of the `=`, it gets parsed as part of the key in the `.ini` file (e.g., `c_flags +` becomes the key). Is there an acceptable "tag" that we could prepend to the key name to indicate a merge that isn't `+`, but would still allow this functionality to be added to cross/native files under the existing architecture?

Ericson2314:
@phillipjohnston the `--merge-config-files` solution should be almost as simple? I believe no one is *against* that?

phillipjohnston:
@Ericson2314 I don't know the proper place to put such an argument in the current architecture, but I can give it a shot first before I declare defeat :).

Ericson2314:
@phillipjohnston 

> I can give it a shot first before I declare defeat :).

Alright! :)

You can add the argument in `mesonbuild/msetup.py`. Look at how other flags are handled and you should be able to follow the logic from there to `mesonbuild/environment.py`. Feel free to ask questions, of course.

dcbaker:
One thing I think is important is leaving "files replace other files" behavior in some form, that behavior is also useful in some cases. so maybe the better thing would for the command line to be:
`--machine-file-policy={merge,replace}` with `replace` being the current (and default) behavior, and `merge` being the new behavior.

phillipjohnston:
@Ericson2314: thank you for the pointers
@dcbaker: acknowledged

xclaesse:
Throwing ideas there in favor of adding `+` operator: https://github.com/mesonbuild/meson/pull/6517#issuecomment-603471250.  I'm starting to like the idea actually.

PR #6816, opened 2020-03-20
project test junit schema + a few more uses

dcbaker:
This adds a schema for the json format we use in the framework tests, and a unittest which runs the schema against the tests (using the jsonschema python module). This also adds a couple of matrix options for some of the framework tests so we can test all of the different methods of discovering those dependencies.

This schema is also useful for IDE users, as it can be used for error checking and auto completion. With VSCode for example, I can get both.

I suspect that this has a chicken and egg problem with the CI images.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -23,3 +23,5 @@ indent_size = 2
 [meson.build]
 indent_size = 2
 
+[*.json]
+indent_size = 2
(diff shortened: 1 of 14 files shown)


dcbaker:
Maybe it's smarter to have a specific github action for validating the schema instead of a unit test? anyone have opinions?

mensinda:
IMHO, testing things across as many platforms as possible is always better. So we should rather make the CI image update process a lot less painefull :)

jpakkane:
>  So we should rather make the CI image update process a lot less painful

Which it should now be.

dcbaker:
I think this is all working now, except for a couple of unrelated bugs

PR #4570, opened 2018-11-30
wrap: Add support for applying a list of patch files

xclaesse:
This replaces my previous idea from PR https://github.com/mesonbuild/meson/pull/4382.

Patch files are commonly used, that's what distro use and it makes sense to be able to just copy the patch file from distro for example. It does not replace the `patch_filename` tarball we already have because that one serve a different purpose, tailored for wrapdb (doesn't require external command-line tool, less subject to conflict when used to only drop meson.build files).

PR #6818, opened 2020-03-21
Wrap: add local files support via *_filename

### LINKED ISSUE #6732: Wish: add ability to use local patch files
It is impossible to use cmake projects if they placed not in root directory of subproject.

For example look into LLVM project repository:
https://github.com/llvm/llvm-project/
Each of its subdir is cmake projects but it is impossible to use them from Meson if llvm-project will be wrapped as subproject.

mensinda:
This PR adds the `source_path` and `patch_path` entries. This approach is more readable and user friendly than the current approach with the `packagecache` directory.

fixes #6732


jpakkane:
I'm not really a fan of this approach. In Meson we have aimed to minimize the number of customization points whenever possible. This does not really save all that much, because either you need to copy your stuff to the packagecache dir _or_ you need to write the full paths to the cross file.

The advantage of packagecache is that it is machine independent. You can `scp` the source tree from one machine to another and everything still works. Arbitrary paths don't really provide for that.

mensinda:
These are *not* arbitrary paths. All source files have to be inside the `subprojects` directory, so `scp` still works. So the `_path` variables are basically the packagecache variant without the need to specify a dummy URL and a checksum.

IMHO this is much cleaner because `*_path` covers a completely different use case. The existing variable triple should (arguably) only be used for fetching remote content.

mensinda:
> In Meson we have aimed to minimize the number of customization points whenever possible.

True, but there is also:

> Meson is an open source build system meant to be both extremely fast, and, even more importantly, as **user friendly** as possible.

I wouldn't call specifying 3 variables, of which one does absolutely nothing, and having to put your archive in a folder with `cache` in the name user friendly :)

My main point here is that fetching remote sources and providing sources within the project are two separate things and thus should be treated differently.

jpakkane:
> I wouldn't call specifying 3 variables, of which one does absolutely nothing, and having to put your archive in a folder with cache in the name user friendly

On the other hand there is consistenty. Currently when you come to any Meson project you _know_ where all the dependencies are without needing to read any project documentation or catting wrap files. This sort of consistency is, among other things, why suprojects are always in the `subprojects` directory rather than scattered about the source tree. Some people dislike this at first and complain (I know this because they always complain about it to me). Then they get used to it and don't even notice it any more.

mensinda:
My main problem with the current setup is that it feels like a hack/bug. You *must* set a *dummy* URL, and maintain the checksum (which is arguably unnecessary, because git should do that for you). The only relevant information is stored in `*_filename`.

I would also argue that it is far more inconsistent to store vendored files inside a directory that was originally designed to be a simple cache maintained by meson. Thus, it is not possible to simply add the entire `packagecache` directory to `.gitignore`, but a whitelist for the non cache files also has to be maintained. You also can't immediately tell which files are automatically downloaded by meson, and which are provided by the repository.

And again, downloaded files and local files should be treated differently IMHO.

I am not against storing local patches/sources in a fixed location (the current restriction is somewhere in the root subprojects directory), so could we then introduce a new folder (`packagefiles`, `packagelocal`, `localpackages` or `localfiles`) were all vendored patches/sources can be stored? In that case, we could use `*_filename` to specify which file inside this new directory should be used, if neither `*_url` nor `*_hash` are present.

jpakkane:
We can do that. I'd still keep the hashes and would validate them. It's a good safeguard against accidental bugs.

Maybe the wrap file should be different as well? Something like [wrap_local]` (I know, not a very good name either)?

mensinda:
> We can do that. I'd still keep the hashes and would validate them. It's a good safeguard against accidental bugs.

Could we make this feature optional then? I personally don't want to maintain a hash for such files in my git repo :)

Also, I think that the biggest source of bugs is that patches are only applied once. If you change the patch and update the hash nothing happens. I don't know how to elegantly fix this, other than recreating the directory when the wrap or patch file changes.

> Maybe the wrap file should be different as well?

I don't think so since this would require that all files (sources and patches) are either downloaded or provided locally. I would say that the main use case for local files would be to download the main sources from the web and just locally maintain the patch.

mensinda:
Updated. Now `*_filename` (in the absence of `*_url`) is used for `subprojects/packagefiles`. Also, `*_hash` is optional in this setup.

If someone has a better name than `packagefiles`, please let me know.

denizzzka:
>> We can do that. I'd still keep the hashes and would validate them. It's a good safeguard against accidental bugs.

> Could we make this feature optional then? I personally don't want to maintain a hash for such files in my git repo :)

Already now we can accidentally skip hash checking by using wrap-git instead of wrap-file.
So, it is need an option that disables ability to download external repositories without hash checking.
But for a local patches introduced in this PR hash checking looks redundant - its files already covered by hashes of the current repository.

mensinda:
ping :)

denizzzka:
I really, really hope that this PR will be included into the next version!

docs/markdown/Wrap-dependency-system-manual.md
```diff
@@ -74,9 +74,11 @@ revision = head
 - `source_url` - download url to retrieve the wrap-file source archive
 - `source_filename` - filename of the downloaded source archive
 - `source_hash` - sha256 checksum of the downloaded source archive
+- `source_path` - *(since 0.54.0)* path to a local source file (replaces `source_url`, `source_filename`, and `source_hash`)
```
xclaesse, line 77: 55 :(

docs/markdown/Wrap-dependency-system-manual.md
```diff
@@ -74,9 +74,11 @@ revision = head
...
 - `patch_url` - download url to retrieve an optional overlay archive
 - `patch_filename` - filename of the downloaded overlay archive
 - `patch_hash` - sha256 checksum of the downloaded overlay archive
+- `patch_path` - *(since 0.54.0)* path to a local patch file (replaces `patch_url`, `patch_filename`, and `patch_hash`)
```
xclaesse, line 81: here too

mesonbuild/wrap/wrap.py
```diff
@@ -383,18 +384,44 @@ def download(self, what: str, ofname: str) -> None:
...
+                mlog.log('Using', mlog.bold(self.packagename), what, 'from cache.')
+                return cache_path
+
+            if not os.path.isdir(self.cachedir):
+                os.mkdir(self.cachedir)
+            self.download(what, cache_path)
             return cache_path
+        elif what + '_path' in self.wrap.values:
+            from ..interpreterbase import FeatureNew
+            FeatureNew('Local wrap patch files via {}_path'.format(what), '0.54.0').use('')
```
xclaesse, line 405: You should pass down current interpreter's subproject, it's not always the main project that use that new feature.

mesonbuild/wrap/wrap.py
```diff
@@ -383,18 +384,44 @@ def download(self, what: str, ofname: str) -> None:
...
+                raise WrapException('File "{}" does not exist'.format(path))
 
-        if not os.path.isdir(self.cachedir):
-            os.mkdir(self.cachedir)
-        self.download(what, cache_path)
-        return cache_path
+            path = path.resolve()
+            try:
+                path.relative_to(Path(self.subdir_root))
+            except Exception:
```
xclaesse, line 419: It raises ValueError

mesonbuild/wrap/wrap.py
```diff
@@ -383,18 +384,44 @@ def download(self, what: str, ofname: str) -> None:
...
+            if os.path.exists(cache_path):
+                self.check_hash(what, cache_path)
+                mlog.log('Using', mlog.bold(self.packagename), what, 'from cache.')
+                return cache_path
+
+            if not os.path.isdir(self.cachedir):
+                os.mkdir(self.cachedir)
+            self.download(what, cache_path)
             return cache_path
+        elif what + '_path' in self.wrap.values:
```
xclaesse, line 403: Should raise exception here if `what + '_filename' in self.wrap.values`.

mesonbuild/wrap/wrap.py
```diff
@@ -383,18 +384,44 @@ def download(self, what: str, ofname: str) -> None:
...
-            os.mkdir(self.cachedir)
-        self.download(what, cache_path)
-        return cache_path
+            path = path.resolve()
+            try:
+                path.relative_to(Path(self.subdir_root))
+            except Exception:
+                raise WrapException('{}_path must be within the root subprojects directory'.format(what))
+
+            return path.as_posix()
```
xclaesse, line 422: We should call self.check_hash() if we have `what + '_hash' in self.wrap.values`, just in case.

mesonbuild/wrap/wrap.py
mensinda: Fixed.

mensinda:
It looks like I accidentally force pushed a previous version of the PR with `*_path` instead of `*_filename`. I restored the new version and fixed the points in the code review that still apply.

xclaesse:
Looks good to me, but CI fails and it seems related.

docs/markdown/snippets/wrap_patch.md
```diff
@@ -0,0 +1,6 @@
+## Local wrap source and patch files
+
+It is now possible to use the `patch_filename` and `source_filename` value in a
```
denizzzka, line 3: Can it be also added `patch_directory` to avoid usage of archiever?

xclaesse: You mean a subtree with meson.build files, or actual .patch files? For the former I did https://github.com/mesonbuild/meson/pull/6920/commits/4a12c4394e8e14ea90ac263590ab24e796f84cbd, but I agree allowing a tree would be better. For the latter I did that: https://github.com/mesonbuild/meson/pull/4570.

denizzzka: I am about `patch_filename = some_patch.zip` archive content. Here is not need to use archives because single-file URL isn't used.

mensinda: This won't happen in this PR but may be implemented in a future PR. However, renaming `patch_filename` is highly unlikely, since this would be a breaking change with very little benefit.

denizzzka: > However, renaming patch_filename is highly unlikely, since this would be a breaking change with very little benefit.

Why? (As I understand) `patch_filename` is introduced in this PR

mensinda: No, they already exist: https://mesonbuild.com/Wrap-dependency-system-manual.html#wrap-format

denizzzka:
Everything worked fine for me in this PR

PR #6819, opened 2020-03-21
Add property to disable compiler sanity checks during cross compilation.

jpakkane:
Ping @keith-packard can you test if this fixes your issue?

### DIFF
--- a/cross/ubuntu-armhf.txt
+++ b/cross/ubuntu-armhf.txt
@@ -18,6 +18,8 @@ cpp_args = '-DMESON_TEST_ISSUE_1665=1'
 has_function_printf = true
 has_function_hfkerhisadf = false
 
+skip_sanity_check = true
+
 [host_machine]
 system = 'linux'
 cpu_family = 'arm'
--- a/docs/markdown/snippets/skipsanity.md
+++ b/docs/markdown/snippets/skipsanity.md
@@ -1,0 +1,9 @@
+## Skip sanity tests when cross compiling
+
+For certain cross compilation environments it is not possible to
+compile a sanity check application. This can now be disabled by adding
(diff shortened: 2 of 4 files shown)


Ericson2314:
Hmm, there already was some stuff where it wouldn't try to link anything for cross. Not against this in principle, but curious what the symptom of not having it is.

keith-packard:
> Hmm, there already was some stuff where it wouldn't try to link anything for cross. Not against this on principle (I like being able to elide all environment reflection by hard-coding), but curious what the symptom of not having it is.

This variable disables the initial sanity checks at meson startup, which were otherwise unconditionally attempting to compile, link and execute a simple program using the target compiler.

keith-packard:
> Ping @keith-packard can you test if this fixes your issue?

Yup, works great! Thanks much.

Ericson2314:
> This variable disables the initial sanity checks at meson startup, which were otherwise unconditionally attempting to compile, link and execute a simple program using the target compiler.

Did you have an exe wrapper? If not, this is definitely a bug we should also make an issue for.

jpakkane:
Note that this is a case of building your own libc. The compiler provides _nothing_, even `_start` or `memcpy` or anything like that. Thus it is not possible to even link an executable.

Ericson2314:
@jpakkane Agreed, but I recall the intent was to use the presence/absence of the exe wrapper as a heuristic for whether linking ought to work in the sanity check.

jpakkane:
In this particular case exe_wrapper is defined and it is being used to run the test suite. This is a very niche use case only encountered by unusual projects.

Ericson2314:
Ah OK, well that explains it. Glad this option exists, glad it's not a bug after all!

PR #6822, opened 2020-03-22
Do not try to mix gcc with link.exe.

jpakkane:


jpakkane:
Test failure is unrelated, I checked that the unit test this touches passed on both.

This is not the final, correct solution for this problem, but it makes all tests green, which is more important for now.

PR #6826, opened 2020-03-23
compilers: disable ccache when using temporary source files

hwti:
When doing a compile test with a testfile.c, ccache fails since the path is random.
So it's better to disable it, to avoid reporting this as a cache miss.


jpakkane:
Thanks.

PR #6827, opened 2020-03-23
Document and test env vars a bit better

Ericson2314:
Since this PR was first opened, just about everything it was supposed to do already happened, so now it is just refining the docs and tests for existing behavior.

### DIFF
--- a/docs/markdown/howtox.md
+++ b/docs/markdown/howtox.md
@@ -12,12 +12,12 @@ When first running Meson, set it in an environment variable.
 $ CC=mycc meson <options>
 ```
 
-Note that environment variables like `CC` only works in native builds.
-The `CC` refers to the compiler for the host platform, that is the
-compiler used to compile programs that run on the machine we will
-eventually install the project on. The compiler used to build things
-that run on the machine we do the building can be specified with
-`CC_FOR_BUILD`. You can use it in cross builds.
+Note that environment variables like `CC` only refer to the host
+platform in cross builds.  That is, `CC` refers to the compiler used to
+compile programs that run on the machine we will eventually install the
+project on. The compiler used to build things that run on the machine we
+do the building can be specified with `CC_FOR_BUILD`. You can use it in
+cross builds.
 
 Note that environment variables are never the idiomatic way to do
 anything with Meson, however. It is better to use the native and cross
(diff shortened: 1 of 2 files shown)


run_unittests.py
```diff
@@ -6338,11 +6338,10 @@ def test_identity_cross_env(self):
         testdir = os.path.join(self.unit_test_dir, '61 identity cross')
         env = {
             'CC_FOR_BUILD': '"' + os.path.join(testdir, 'build_wrapper.py') + '"',
+            'CC': '"' + os.path.join(testdir, 'host_wrapper.py') + '"',
         }
         crossfile = tempfile.NamedTemporaryFile(mode='w')
-        crossfile.write('''[binaries]
-c = ['{0}']
-'''.format(os.path.join(testdir, 'host_wrapper.py')))
+        crossfile.write('')
```
Ericson2314: I don't have to change this test, as this is a non-breaking change. However, doing so makes it less redundant with the previous test, which exclusive uses the native and cross files with no environment variables.

Ericson2314:
@ePirat Let's not forget about this :).

docs/markdown/howtox.md
```diff
@@ -12,15 +12,16 @@ When first running Meson, set it in an environment variable.
 $ CC=mycc meson <options>
 ```
 
-Note that environment variables like `CC` only works in native builds. The `CC`
-refers to the compiler for the host platform, that is the compiler used to
-compile programs that run on the machine we will eventually install the project
-on. The compiler used to build things that run on the machine we do the
-building can be specified with `CC_FOR_BUILD`. You can use it in cross builds.
+Note that environment variables like `CC` only refer to the host platform in
+cross builds.  That is, the `CC` refers compiler used to compile programs that
```
ePirat, line 16: > That is, the `CC` refers compiler used

should probably be "refers to the compiler used".

Ericson2314: Thanks

mesonbuild/envconfig.py
```diff
@@ -119,9 +119,9 @@ def get_env_var_pair(for_machine: MachineChoice,
         # compiling we fall back on the unprefixed host version. This
         # allows native builds to never need to worry about the 'BUILD_*'
         # ones.
-        ([var_name + '_FOR_BUILD'] if is_cross else [var_name]),
+        [var_name + '_FOR_BUILD'] + ([] if is_cross else [var_name]),
         # Always just the unprefixed host verions
-        ([] if is_cross else [var_name]),
+        [var_name],
```
jtojnar, line 124: Done in https://github.com/mesonbuild/meson/commit/e03c4d4199ba47e008a57bce03cc5477347763f3

Ericson2314: Oh nice! Half way there!

dcbaker (approved):

dcbaker:
Updating the tests to help with env2cross seems like a great idea

PR #6829, opened 2020-03-23
Fix resource file finding

### LINKED ISSUE #6801: FileNotFoundError: [Errno 2] \\Meson\\mesonbuild\\dependencies\\data\\CMakePathInfo.txt
I am trying to install the [BGP Scanner](https://gitlab.com/Isolario/bgpscanner) with meson, I got this failure. Could anyone provide an advice for this issue? Thanks! 

```
C:\Users\icherepa\Desktop\bgpscanner> meson --buildtype=release ..
The Meson build system
Version: 0.53.2
Source dir: C:\Users\icherepa\Desktop\bgpscanner
Build dir: C:\Users\icherepa\Desktop
Build type: native build
Project name: bgpscanner
Project version: 2.31
C compiler for the host machine: gcc (gcc 6.3.0 "gcc (MinGW.org GCC-6.3.0-1) 6.3.0")
C linker for the host machine: gcc ld.bfd 2.28
Host machine cpu family: x86
Host machine cpu: x86
Found pkg-config: C:\MinGW\bin\pkg-config.EXE (0.23)
Found CMake: C:\Program Files\CMake\bin\cmake.EXE (3.17.0-rc3)
  File "mesonbuild\mesonmain.py", line 129, in run
  File "mesonbuild\msetup.py", line 245, in run
  File "mesonbuild\msetup.py", line 159, in generate
  File "mesonbuild\msetup.py", line 192, in _generate
  File "mesonbuild\interpreter.py", line 4167, in run
  File "mesonbuild\interpreterbase.py", line 412, in run
  File "mesonbuild\interpreterbase.py", line 436, in evaluate_codeblock
  File "mesonbuild\interpreterbase.py", line 430, in evaluate_codeblock
  File "mesonbuild\interpreterbase.py", line 443, in evaluate_statement
  File "mesonbuild\interpreterbase.py", line 1064, in assignment
  File "mesonbuild\interpreterbase.py", line 441, in evaluate_statement
  File "mesonbuild\interpreterbase.py", line 788, in function_call
  File "mesonbuild\interpreterbase.py", line 285, in wrapped
  File "mesonbuild\interpreterbase.py", line 285, in wrapped
  File "mesonbuild\interpreterbase.py", line 285, in wrapped
  [Previous line repeated 2 more times]
  File "mesonbuild\interpreterbase.py", line 155, in wrapped
  File "mesonbuild\interpreterbase.py", line 174, in wrapped
  File "mesonbuild\interpreter.py", line 3236, in func_dependency
  File "mesonbuild\interpreter.py", line 3283, in dependency_impl
  File "mesonbuild\dependencies\base.py", line 2234, in find_external_dependency
  File "mesonbuild\dependencies\base.py", line 1104, in __init__
  File "mesonbuild\dependencies\base.py", line 1158, in _get_cmake_info
  File "mesonbuild\dependencies\base.py", line 1533, in _call_cmake
  File "mesonbuild\dependencies\base.py", line 1507, in _setup_cmake_dir
  File "pathlib.py", line 1229, in read_text
  File "pathlib.py", line 1215, in open
  File "pathlib.py", line 1071, in _opener
FileNotFoundError: [Errno 2] No such file or directory: 'C:\\Program Files\\Meson\\mesonbuild\\dependencies\\data\\CMakePathInfo.txt'
```

I don't know how my issue is related to https://github.com/mesonbuild/meson/issues/5858 (there is also no solution answer).

dcbaker:
This is a small but significant change, in particular it changes meson to use setuptools pkg_resource for finding data files. This is important because pkg_resource understands all of the odd corner cases of where python might install data files in different cases (ziplib, wheel, egg, running from the directory, etc). We already have a dependency on setuptools for it's entrypoint functions (and to install meson at all), so this isn't a new dependency, but it is another place we depend on setuptools.

Fixes #6801

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -30,6 +30,8 @@ long_description = Meson is a cross-platform build system designed to be both as
 
 [options]
 python_requires = >= 3.5.2
+setup_requires =
+  setuptools
 
 [options.extras_require]
 progress =
--- a/mesonbuild/cmake/interpreter.py
+++ b/mesonbuild/cmake/interpreter.py
@@ -15,6 +15,8 @@
 # This class contains the basic functionality needed to run any interpreter
 # or an interpreter-based tool.
 
+import pkg_resources
+
 from .common import CMakeException, CMakeTarget
 from .client import CMakeClient, RequestCMakeInputs, RequestConfigure, RequestCompute, RequestCodeModel
 from .fileapi import CMakeFileAPI
@@ -779,7 +781,7 @@ class CMakeInterpreter:
             raise CMakeException('Unable to find CMake')
         self.trace = CMakeTraceParser(cmake_exe.version(), self.build_dir, permissive=True)
 
-        preload_file = Path(__file__).resolve().parent / 'data' / 'preload.cmake'
+        preload_file = pkg_resources.resource_filename('mesonbuild', 'cmake/data/preload.cmake')
 
         # Prefere CMAKE_PROJECT_INCLUDE over CMAKE_TOOLCHAIN_FILE if possible,
         # since CMAKE_PROJECT_INCLUDE was actually designed for code injection.
@@ -1024,7 +1026,7 @@ class CMakeInterpreter:
         root_cb.lines += [function('project', [self.project_name] + self.languages)]
 
         # Add the run script for custom commands
-        run_script = '{}/data/run_ctgt.py'.format(os.path.dirname(os.path.realpath(__file__)))
+        run_script = pkg_resources.resource_filename('mesonbuild', 'cmake/data/run_ctgt.py')
         run_script_var = 'ctgt_run_script'
         root_cb.lines += [assign(run_script_var, function('find_program', [[run_script]], {'required': True}))]
 
--- a/mesonbuild/dependencies/base.py
+++ b/mesonbuild/dependencies/base.py
@@ -28,6 +28,8 @@ import typing as T
 from enum import Enum
 from pathlib import Path, PurePath
 
+import pkg_resources
+
 from .. import mlog
 from .. import mesonlib
 from ..compilers import clib_langs
@@ -1508,8 +1510,8 @@ class CMakeDependency(ExternalDependency):
         build_dir = self._get_build_dir()
 
         # Insert language parameters into the CMakeLists.txt and write new CMakeLists.txt
-        src_cmake = Path(__file__).parent / 'data' / cmake_file
-        cmake_txt = src_cmake.read_text()
+        # Per the warning in pkg_resources, this is *not* a path and os.path and Pathlib are *not* safe to use here.
+        cmake_txt = pkg_resources.resource_string('mesonbuild', 'dependencies/data/' + cmake_file).decode()
 
         # In general, some Fortran CMake find_package() also require C language enabled,
         # even if nothing from C is directly used. An easy Fortran example that fails


mensinda:
You missed [`run_ctgt.py`](https://github.com/mesonbuild/meson/blob/master/mesonbuild/cmake/interpreter.py#L1027). Also, wouldn't this change mean that setuptools is now also required at runtime? As far as I can tell nothing in `mesonbuild` currently depends on setuptools, so this would be a new dependency.

dcbaker:
We require it at runtime too, the meson interface is generated with setuptools' `console_scripts` field, which requires `pkg_resource` at runtime already.

```sh
cat `which meson`
```

```py
#!/usr/bin/env python3
__requires__ = 'meson'
import re
import sys
from pkg_resources import load_entry_point

if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
    sys.exit(
        load_entry_point('meson', 'console_scripts', 'meson')()
    )
```

incidently this is why using `sys.argv[0]` is dangerous in python, and you should use `__file__` instead :)

dcbaker:
Updated for `run_ctgt.py`

jpakkane:
An alternative approach would be to not have data files but instead put them in py files like so:

```python
preload_cmake = '''
<contents here>
'''
```

This one is guaranteed to always work, though you do lose syntax highlighting for editing.

dcbaker:
I feel like you're giving up a lot putting them as strings inside a python file, syntax highlighting, built-in editor linting, auto completion, and the fact that you're going to end up with a python file like:

```py
CMAKE_STRING_TO_WRITE_TO_FILE = """\
<cmake>
"""
```

If we didn't already rely on setuptools I'd probably say we need to do something like that, but it feels like an ugly hack when we already require setuptools and it has a ready made solution exactly for this problem.

setup.cfg
```diff
@@ -30,6 +30,8 @@ long_description = Meson is a cross-platform build system designed to be both as
 
 [options]
 python_requires = >= 3.5.2
+setup_requires =
```
bluetech, line 33: If you use it at runtime, shouldn't it be `install_requires`?

I would imagine having `setuptools` in `setup_requires` would not have any effect since setuptools is itself what interprets `setup_requires`, though I'm not 100% sure.

dcbaker: Normally I would agree with you, but I think debian splits pkg_resource out of setuptools.

dcbaker:
If we could rely on python 3.7 or allowed external dependencies I would be completely on board with using importlib.metadata. setuptools is our only external dependency, and I think @jpakkane really wants to keep it that way. Maybe we can make an exception for backport functionality since one could use a newer version of python if they don't want dependencies?

jpakkane:
Meson has to work out of the box on a standard Python install without external deps. That is a given. This can be fixed later as we raise our required Python version. We should be able to bump it soonish as the new Ubuntu LTS is out.

eli-schwartz:
I had the same thought, but before I tried doing anything about it I was trying to understand why Path isn't sufficient to begin with:

```
05:43 PM <eschwartz> dcbaker: basically, I'm wondering why you need to import pkg_resources and use that -- I would think there should be a more lightweight solution.
05:45 PM <@dcbaker> I'm trying to remember exactly, but it's something to do with the way the meson.exe is created
05:45 PM <@dcbaker> the modules are actually installed someplace like c:\\program files\\python3.7\\site-packages\\
05:51 PM <eschwartz> lol, but how is Path(__file__) reporting something totally different then?
05:57 PM <eschwartz> dcbaker: so either way this doesn't have anything to do with needing to extract files from a zipped egg or anything?
05:59 PM <@dcbaker> Like I said, I don't remember exactly what the problem was, and I don't have my windows laptop handy to remind myself :)
```

Does anyone remember why this is? This PR unfortunately doesn't preserve any knowledge about how it happened, other than "corner cases". It's difficult to know whether importlib.metadata "could" be used, without knowing "why" pkg_resources fixed it, because it sounds like the problem is some odd implementation detail.

bluetech:
> Meson has to work out of the box on a standard Python install without external deps.

I was thinking `meson` could conditionally use `importlib.metadata`/`importlib_metadata` if available, and fall back to `pkg_resources` if not available. This way newer Python versions don't have to pay the "`pkg_resources` tax".

Just to demonstrate the issue more, in current meson master,

```
$ time ./meson.py --version
0.54.999

real	0m0.350s
user	0m0.317s
sys	0m0.031s
```

After applying this diff:

<details>

```diff
diff --git a/mesonbuild/cmake/interpreter.py b/mesonbuild/cmake/interpreter.py
index 125f18b76..997f02742 100644
--- a/mesonbuild/cmake/interpreter.py
+++ b/mesonbuild/cmake/interpreter.py
@@ -15,7 +15,7 @@
 # This class contains the basic functionality needed to run any interpreter
 # or an interpreter-based tool.
 
-import pkg_resources
+# import pkg_resources
 
 from .common import CMakeException, CMakeTarget
 from .client import CMakeClient, RequestCMakeInputs, RequestConfigure, RequestCompute, RequestCodeModel
diff --git a/mesonbuild/dependencies/base.py b/mesonbuild/dependencies/base.py
index c0ec089db..92850e0ba 100644
--- a/mesonbuild/dependencies/base.py
+++ b/mesonbuild/dependencies/base.py
@@ -28,7 +28,7 @@ import typing as T
 from enum import Enum
 from pathlib import Path, PurePath
 
-import pkg_resources
+# import pkg_resources
 
 from .. import mlog
 from .. import mesonlib
```
</details>

the time is:

```
real	0m0.244s
user	0m0.213s
sys	0m0.030s
```

So more than 100ms startup overhead.

Note that the `./meson.py` above doesn't use the `entry_points` mechanism. That mechanism loads `pkg_resources` in some situations as @dcbaker mentioned, specifically when not installed through a Python "wheel". Virtualenvs created by `virtualenv` have `wheel` pre-installed, while virtualenvs created by `python -m venv` don't. My system meson (Arch Linux) unfortunately seems to use the slow method. Perhaps I should report it to them since it probably slows down the startup of a lot of command line programs:

<details>
<summary>List of binaries on my system</summary>

```
$ grep --files-with-matches 'from pkg_resources import load_entry_point' /bin | sort
borg
borgfs
chardetect
docker-compose
easy_install
easy_install-3.8
f2py
f2py3
f2py3.8
isympy
jsonschema
lit
mako-render
markdown_py
meson
mid3cp
mid3iconv
mid3v2
moggsplit
mutagen-inspect
mutagen-pony
pygmentize
pyrsa-decrypt
pyrsa-encrypt
pyrsa-keygen
pyrsa-priv2pub
pyrsa-sign
pyrsa-verify
tox
tox-quickstart
tqdm
virtualenv
virtualenv3
wheel
youtube-dl
```

xclaesse:
I think one significant code path we should optimise is `meson --internal <script>` because that's called for each custom_target() on Windows. I tested and it takes >250ms just for a no-op.

DimStar77:
I just run into this issue of pkg_resources being a new runtime dep for meson.

at openSUSE we have something called 'ring0' which we use to monitor how easily we can still bootstrap a distro (and it's terrible) - by now, meson is already part of that bootstrap cycle (due to libeconf, other consumers like rpm, util-linux) are interested in converting to meson, but you know more there than me :)

Now with this recent change, I need python-setuptools in the base bootstrap and setuptools itself brings more deps, making a distro bootstrap even more complicated (generally, that stuff all depends on each other to be able to get started in building a distro)

eli-schwartz:
Meson doesn't just accidentally use setuptools. Distutils isn't a suitable replacement if you have entry_points.

mensinda:
We should still try to get rid of setuptools inside mesonbuild itself, IMHO. This way, we wouldn't require it when running the `meson.py` directly (be it from a release tarball or `git clone`), which would improve the bootstrapping issue.

Then the only place where we would still use setuptools would be from the automatically-generated `EASY-INSTALL-ENTRY-SCRIPT` script (see the `setup.py`) with all the fun `load_entry_point` stuff.

eli-schwartz:
I have some vague memory that someone suggested they weren't sure anything but pkg_resources would handle the right "edge cases", not even importlib. But I'm still tremendously curious what the actual case is that needs handling, and why Path isn't sufficient. Does anyone here use Windows and understand just what the MSI installer was doing?

mensinda:
#6801 seems to be the issue that was fixed with this.

I guess the most portable solution would be to store all the data files as strings in a `*.py`, like suggested in https://github.com/mesonbuild/meson/pull/6829#issuecomment-602907488. This file could be automatically generated from the current data files, so we wouldn't lose as much convenience.

PR #6830, opened 2020-03-23
docs/markdown: Add a new page with common cross/native file options [skip ci]

dcbaker:
Rather than having two separate sections with duplicated information
lets just have one for the common settings, and only document sections
specific to each file in separately


phillipjohnston:
Thanks for doing this.

dcbaker:
This was originally part of a larger more complicated series that probably wont land any time soon, but this seemed like a worthwhile change on its own :)

Issue #5910, opened 2019-09-10
Meson misdetects lld as "APPLE ld unknown version" with the llvm-mingw cross toolchain


Akaricchi:
Here's the toolchain in question: https://github.com/mstorsjo/llvm-mingw

Linker invocation:
```
 % bin/x86_64-w64-mingw32-c11 -o test.exe test.c -v
clang version 9.0.0 (trunk 356111) (llvm/trunk 356109)
Target: x86_64-w64-windows-gnu
Thread model: posix
InstalledDir: /data/llvm-mingw/bin
 "/data/llvm-mingw/bin/clang-9" -cc1 -triple x86_64-w64-windows-gnu -emit-obj -mrelax-all -disable-free -disable-llvm-verifier -discard-value-names -main-file-name test.c -mrelocation-model pic -pic-level 2 -mthread-model posix -fmath-errno -masm-verbose -mconstructor-aliases -munwind-tables -target-cpu x86-64 -dwarf-column-info -debugger-tuning=gdb -momit-leaf-frame-pointer -v -resource-dir /data/llvm-mingw/lib/clang/9.0.0 -internal-isystem /data/llvm-mingw/lib/clang/9.0.0/include -internal-isystem /data/llvm-mingw/x86_64-w64-mingw32/include -internal-isystem /data/llvm-mingw/include -fdebug-compilation-dir /data/llvm-mingw -ferror-limit 19 -fmessage-length 239 -fobjc-runtime=gcc -fseh-exceptions -fdiagnostics-show-option -fcolor-diagnostics -o /tmp/test-ae289e.o -x c test.c -faddrsig
clang -cc1 version 9.0.0 based upon LLVM 9.0.0svn default target x86_64-unknown-linux-gnu
#include "..." search starts here:
#include <...> search starts here:
 /data/llvm-mingw/lib/clang/9.0.0/include
 /data/llvm-mingw/x86_64-w64-mingw32/include
 /data/llvm-mingw/include
End of search list.
test.c:1:1: warning: variable named 'main' with external linkage has undefined behavior [-Wmain]
int main;
^
1 warning generated.
[...]

PR #6836, opened 2020-03-24
Clarify a workaround for older versions of LLD for MinGW

mstorsjo:
The previous diagnosis (that the wrapper that injects the -target
option to clang caused the issues) doesn't seem right; the issue
was that LLD's MinGW frontend (which is invoked when ld.lld is given
a COFF argument to the -m option) didn't handle the --version argument
before LLD 9.0.0.

This clarifies the diagnosis from #5910 (ping @Akaricchi).

### DIFF
--- a/mesonbuild/environment.py
+++ b/mesonbuild/environment.py
@@ -854,7 +854,9 @@ class Environment:
             linker = LLVMDynamicLinker(
                 compiler, for_machine, comp_class.LINKER_PREFIX, override, version=v)  # type: DynamicLinker
         elif e.startswith('lld-link: '):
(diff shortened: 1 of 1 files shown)


dcbaker:
Just FYI, in the future when you're making doc only or comment only changes add `[skip ci]` to the end of your commit message so it wont run through the CI. It has to be in the commit messages, not in the MR title.

I'll merge this as soon as Akaricchi says it looks good.

Akaricchi:
Yeah this seems right. I can't reproduce the bad behavior on 9.0.0.

Issue #6453, opened 2020-01-12
link_language has no effect


burp:
A patch from issue 1534 introduced link_language, but I believe that it is not working and the introduced test case does not test link_language, see https://github.com/mesonbuild/meson/issues/1534

The introduced patch does not seem to work and the supplied test case does not seem to be a proper test of the functionality. This is because the C component is only supplied as a library and it will therefore select the Fortran compiler for linking. From a discussion in #mesonbuild we figured out that "as far as executable() can tell, it is fortran-exclusive code, and shared libraries don't have a language type."

A minimal example is:

project('tproj', ['fortran','cpp'])
executable('run', ['test.f90','test.cpp'], link_language: 'fortran')

then it generates rule cpp_LINKER with the gcc command instead of gfortran

dcbaker:
just to be clear, the cpp_LINKER rule *should* be generated with g++, however there should also be a fortran_LINKER rule with gfortran (or whatever), and your target should be using the fortran_LINKER, not the cpp_LINKER.

@scivision, you're probably interested

scivision:
OK I thought it did work for some cases, but maybe I fooled myself or it was not really that general.

freddy77:
I'm trying to do a similar thing forcing a C linker with some C++ sources using `both_libraries` but Meson is going a warning that `link_language` is not a valid keyword despite documentation states should be.

dcbaker:
`link_language` is only valid for executables. It should only be necessary for executables because meson cannot always figure out which language implements the main() function.

What are you trying to achieve by using the C linker instead of the C++ linker?

freddy77:
To make sure I don't depend on C++ libraries. It's supported by autoconf so why you need to limit this feature? Shared libraries are not much different from executables, usually there's a flag/number in the binary to change it. From documentation (https://mesonbuild.com/Reference-manual.html)

```
Executable supports the following keyword arguments. Note that just like the positional arguments above, these keyword arguments can also be passed to shared and static libraries.
```

ebraminio:
FWIW we like to have this in HarfBuzz also, we were able to choose our linker like this https://github.com/harfbuzz/harfbuzz/blob/748e1cf/src/Makefile.am#L120-L129 in autotools and implement the same in cmake https://github.com/harfbuzz/harfbuzz/blob/e860000/CMakeLists.txt#L481-L488 but not in meson apparently.

PR #6838, opened 2020-03-24
Link language in libraries

dcbaker:
It turns out the `link_language` field basically doesn't work at all, more than that if it did actually work it would work on more than just executable (which is all it was supposed to work on). As it is there are a couple of legitimate reasons to use this on libraries, namely libraries written in !C (C++ for example), that have C linkage, and the project wants to enforce that.

This series adds tests for the bug, fixes the bug, and remove the broken warning about using `link_language` in non-executables. It also updates the docs to explain when and why you would want to use it.

At least the first 4 patches are candidates for 0.54,though we might as well pull the whole thing in.

Fixes: #6453

### DIFF
--- a/docs/markdown/Reference-manual.md
+++ b/docs/markdown/Reference-manual.md
@@ -600,8 +600,12 @@ be passed to [shared and static libraries](#library).
   depends on such as a symbol visibility map. The purpose is to
   automatically trigger a re-link (but not a re-compile) of the target
   when this file changes.
-- `link_language` since 0.51.0 makes the linker for this target
-  be for the specified language. This is helpful for multi-language targets.
+- `link_language` since 0.51.0 (broken until 0.55.0) makes the linker for this
+  target be for the specified language. It is generally unnecessary to set
+  this, as meson will detect the right linker to use in most cases. There are
+  only two cases where this is needed. One, your main function in an
+  executable is not in the language meson picked, or second you want to force
+  a library to use only one ABI.
 - `link_whole` links all contents of the given static libraries
   whether they are used by not, equivalent to the
   `-Wl,--whole-archive` argument flag of GCC, available since 0.40.0.
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -82,6 +82,7 @@ buildtarget_kwargs = set([
     'override_options',
     'sources',
     'gnu_symbol_visibility',
+    'link_language',
 ])
 
 known_build_target_kwargs = (
@@ -92,7 +93,7 @@ known_build_target_kwargs = (
     rust_kwargs |
     cs_kwargs)
 
-known_exe_kwargs = known_build_target_kwargs | {'implib', 'export_dynamic', 'link_language', 'pie'}
+known_exe_kwargs = known_build_target_kwargs | {'implib', 'export_dynamic', 'pie'}
 known_shlib_kwargs = known_build_target_kwargs | {'version', 'soversion', 'vs_module_defs', 'darwin_versions'}
 known_shmod_kwargs = known_build_target_kwargs | {'vs_module_defs'}
 known_stlib_kwargs = known_build_target_kwargs | {'pic'}
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -1217,7 +1217,7 @@ You probably should put it in link_with instead.''')
 
         See: https://github.com/mesonbuild/meson/issues/1653
         '''
-        langs = []
+        langs = [] # type: T.List[str]
 
         # User specified link_language of target (for multi-language targets)
         if self.link_language:
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -1219,10 +1219,6 @@ You probably should put it in link_with instead.''')
         '''
         langs = [] # type: T.List[str]
 
-        # User specified link_language of target (for multi-language targets)
-        if self.link_language:
-            return [self.link_language]
-
         # Check if any of the external libraries were written in this language
         for dep in self.external_deps:
             if dep.language is None:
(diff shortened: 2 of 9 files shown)


ebraminio:
0.54 is released apparently, sad that it didn't have this and versions here should be updated I guess.

mesonbuild/build.py
```diff
@@ -1240,6 +1237,12 @@ def get_clink_dynamic_linker_and_stdlibs(self):
         # Populate list of all compilers, not just those being used to compile
         # sources in this target
         all_compilers = self.environment.coredata.compilers[self.for_machine]
+
+        # If the user set the link_language, just return that.
+        if self.link_language:
+            comp = all_compilers[self.link_language]
+            return comp, comp.language_stdlib_only_link_flags()
+
```
freddy77, line 1245: Why not something like

```python
        # did user override clink_langs for this target?
        link_langs = [self.link_language] if self.link_language else clink_langs
```
and use `link_langs` below instead of `clink_langs` ? Nothing wrong with code above but code below returns a more friendly error in case the compiler was not found.

freddy77: No, that change I proposed is not enough. I would need to skip the part of code calling `language_stdlib_only_link_flags`.

ePirat:
Sounds like a good change, I have a need for this too (for transitive dependencies that need C++ when statically linking) and currently had to resort to a dummy cpp file as workaround for this.

docs/markdown/Reference-manual.md
```diff
@@ -600,8 +600,12 @@ be passed to [shared and static libraries](#library).
...
   automatically trigger a re-link (but not a re-compile) of the target
   when this file changes.
-- `link_language` since 0.51.0 makes the linker for this target
-  be for the specified language. This is helpful for multi-language targets.
+- `link_language` since 0.51.0 (broken until 0.55.0) makes the linker for this
+  target be for the specified language. It is generally unnecessary to set
+  this, as meson will detect the right linker to use in most cases. There are
+  only two cases where this is needed. One, your main function in an
+  executable is not in the language meson picked, or second you want to force
+  a library to expose only one ABI, such as using C++ with `extern "C"`.
```
freddy77, line 608: I think this comment for C++ is misleading. This option does not help for this, `extern "C"` is enough and adding `link_language` won't forbid the library to export also C++ symbols. What will prevent is to import symbols from standard C++ libraries. In different cases you want to use C++ as a C with additional features but for dependencies or other reasons you don't want to use more than that.

The `extern "C"` in C++ allows to: 

1. import C symbols as C symbols (so for instance you can use `printf` from C++)
2. export C++ symbols as C symbols (so for instance you can use a `my_printf` written in C++ from a C program)

dcbaker: The C linker will die a painful death if you have any C++ symbols exposed, which is something a number of people have wanted: write in C++ but ensure only C symbols are exposed by linking with the C linker.

freddy77: Maybe I don't follow. I tried with your branch this:

- meson.build
```python
project('foo', 'cpp', 'c')

shared_library('foo', ['main.cpp'], link_language : 'c')
```
- main.cpp
```c++
void foo(int) { }
```
It compiles, the shared object file contains C++ symbols `objdump -TC libfoo.so` gives:
```

libfoo.so:     file format elf64-x86-64

DYNAMIC SYMBOL TABLE:
0000000000000000  w   D  *UND*  0000000000000000 __cxa_finalize
0000000000000000  w   D  *UND*  0000000000000000 _ITM_registerTMCloneTable
0000000000000000  w   D  *UND*  0000000000000000 _ITM_deregisterTMCloneTable
0000000000000000  w   D  *UND*  0000000000000000 __gmon_start__
00000000000010f9 g    DF .text  000000000000000a foo(int)
```
`ninja` is using the wanted linker:
```
$ ninja -v
[1/2] ccache c++ -Ifoo@sha -I. -I.. -I/home/user/install/include -fdiagnostics-color=always -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wnon-virtual-dtor -g -fPIC -MD -MQ 'foo@sha/main.cpp.o' -MF 'foo@sha/main.cpp.o.d' -o 'foo@sha/main.cpp.o' -c ../main.cpp
[2/2] cc  -o libfoo.so 'foo@sha/main.cpp.o' -Wl,--as-needed -Wl,--no-undefined -shared -fPIC -Wl,--start-group -Wl,-soname,libfoo.so -Wl,--end-group
```

dcbaker: Consider this (assume there's a match .h):

```cpp
extern "C" int foo() { return 1; }
```

This only exposes C linkage symbols, you can link this with the C linker and everything is good.

But this:

```cpp
#include <string>

std::string bar() { return std::string{"foo"}; }
extern "C" int foo() { return 1; }
```

Will not link with the c linker. This is the case I'm trying to describe.

freddy77: Yes, I perfectly understood. But the linker is not failing because the module is exporting `bar` but because is importing `std::string::string`. So maybe just changing `a library to expose only one ABI` to `a library to import only one ABI` in your description would be enough.

jpakkane:
⚔️ conflicts. Apart from that, is this ready?

dcbaker:
I think this is ready other than conflicts and the one line in the documentation. Do you have preferences on that?

freddy77:
> I think this is ready other than conflicts and the one line in the documentation. Do you have preferences on that?

@dcbaker, are you asking to me or @jpakkane ?

dcbaker:
Sure, I'll add that.

freddy77:
> I've updated it to say "import or export" as it is useful for both cases (the exporting case is the bug that I set out to solve.)

but the example you made had nothing to do with the export

ePirat:
@dcbaker I agree with @freddy77 that saying its about exports is misleading as it's really about the imports, not the exports?

The only reliable way to ensure to not expose C++ symbols seems to me is to build a shared library, and link that library to another shared library that is C-only and includes all public headers and link it with the C linker.

ePirat:
That said this is a very advanced setting and most people using it will probably already know what they need it for.

dcbaker:
Updated again, I've removed the example and replaced "import or export" with "use"

PR #6842, opened 2020-03-25
Do not call restorecon if there are no files to install

### LINKED ISSUE #6452: restorecon called with empty input
**Describe the bug**
If there is nothing to install, `ninja install` shows a slightly misleading restorecon error 

**To Reproduce**
Basically empty `meson.build`
```
project('empty')
```
and then:
```
meson build
cd build
ninja install
```
yields:
```
[0/1] Installing files.
Failed to restore SELinux context of installed files...
Standard output:

Standard error:
restorecon: SELinux: Could not get canonical path for  restorecon: No such file or directory.

Nothing to install.
```

**Expected behavior**
No error message.

**System parameters**
Seen with the Fedora 31 Meson package (meson-0.52.1-1.fc31.noarch).

**Note**
Note the extra space before restorecon in the message. What happens is that restorecon is called with an empty input in `restore_selinux_contexts()`
```
out, err = proc.communicate(input=b'\0'.join(os.fsencode(f) for f in selinux_updates) + b'\0')
```
telling it to process a file with name '', where it shouldn't be called at all, because `selinux_updates` is empty. Should be easily fixable with a `if not selinux_updates: return`.

michaelkuhn:
Fixes #6452

### DIFF
--- a/mesonbuild/minstall.py
+++ b/mesonbuild/minstall.py
@@ -171,6 +171,10 @@ def restore_selinux_contexts():
         # If we don't have restorecon, failure is ignored quietly.
         return
 
+    if not selinux_updates:
+        # If the list of files is empty, do not try to call restorecon.
+        return
+
     with subprocess.Popen(['restorecon', '-F', '-f-', '-0'],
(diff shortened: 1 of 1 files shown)


jpakkane:
Thanks.

PR #6863, opened 2020-03-28
mtest: Avoid over commiting futures

ndufresne:
This patch will prevent creating as many repeats as the total requested
number. This reduce memory pressure when a large number if requested and
will ensure that we stop processing if one of the parallel running test
failed.

Before this patch, we often endup queue all the futures before the first
failures and wait for all the futures in drain_futures(). This is not
very convenient when trying to reproduce a rarely failing test.

Note that the parallel test runners seems to have many more issues. I wrote this patch to unblock my work and to open the discussion on the proper approach forward.

 - CTRL+C does not work
 - future.cancel() does to kill the process
- There is no "forever" feature
- After this patch there is no longer a way to run for all the repeats (e.g. to gather a failure/success ratio)

### DIFF
--- a/mesonbuild/mtest.py
+++ b/mesonbuild/mtest.py
@@ -946,6 +946,8 @@ Timeout:            %4d
                             executor = conc.ThreadPoolExecutor(max_workers=self.options.num_processes)
                         f = executor.submit(single_test.run)
                         futures.append((f, numlen, tests, visible_name, i))
+                        if len(futures) > self.options.num_processes:
+                            self.drain_future(futures.pop(0))
                     if self.options.repeat > 1 and self.fail_count:
                         break
                 if self.options.repeat > 1 and self.fail_count:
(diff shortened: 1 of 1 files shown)


mesonbuild/mtest.py
```diff
@@ -946,6 +946,8 @@ def run_tests(self, tests: T.List['TestSerialisation']) -> None:
                             executor = conc.ThreadPoolExecutor(max_workers=self.options.num_processes)
                         f = executor.submit(single_test.run)
                         futures.append((f, numlen, tests, visible_name, i))
+                        if len(futures) > self.options.num_processes:
+                            self.drain_future(futures.pop(0))
```
jpakkane: This has a negative impact on parallelism. If we have a long running test, this will block until that is finished but no new work can be started.

ndufresne: That can easily be covered by adding an '* 2` on the number of queued process. But with current implementation, it means you'll have to wait double that time before it stops on error.

As I said in the message, I'm looking for feedback on what was the intent of this implementation. There is not enough implemented for me to guess what was expected here. Also, the strict ordering implemented here makes everything much more difficult in term of parallelism. So if you agree to drop that, we can likely just rewrite the entire thing with proper futures API.

ndufresne:
Any feedback ? What about we run and sync each iterations separately ? So basically we'd commit a full iteration, regardless how many there is, let that iteration run completely and only trigger the next iteration if it worked ?

That might not be ideal if one wants to hammer on exactly 1 test. Or maybe we should just go fully async / out of order ?

jpakkane:
What is the resource that actually runs out currently? There are a lot of users for Python's executors, it should be able to handle a few hundred to a thousand futures just fine.

> This is not very convenient when trying to reproduce a rarely failing test.

Would you not tell `meson test` to run only that test in a loop? Or does your bug only appear if you run all the previous tests as well (even though they are in different processes)?

ndufresne:
> What is the resource that actually runs out currently? There are a lot of users for Python's executors, it should be able to handle a few hundred to a thousand futures just fine.
> 
> > This is not very convenient when trying to reproduce a rarely failing test.
> 
> Would you not tell `meson test` to run only that test in a loop? Or does your bug only appear if you run all the previous tests as well (even though they are in different processes)?

That you have one or more test does not matter. With the current implementation, it will always run through all the iterations. The code that tries to stop never worked.

bonzini:
Closing to clean up test targets PRs; the code has been completely rewritten to use async and this fixed a lot of the issues in the description (such as Ctrl-C and cancellation).

PR #6872, opened 2020-03-30
docs: Described library command in Build targets section [skip ci]

vedranmiletic:
Build targets document should address best practices since this is the place where the new users usually get started.


vedranmiletic:
Thanks!

PR #6875, opened 2020-03-30
ci: Fix Azure timeouts with self-hosted dependency bundle

mensinda:
There are some Azure CI failures (*again* :unamused:), more specifically random timeouts. These seem to be caused *somewhere* in the setup code (which takes more than 40min in some instances). So, this PR aims to identify which part of the setup causes this problem.

---

The results below show that downloading boost (up to 30min download time) causes the timeout. This PR (along with mesonbuild/cidata#2) solves this problem by putting all external resources into one big image and hosting it on GitHub.

This PR depends on mesonbuild/cidata#2.


mensinda:
Test results from [vc2019x64ninja](https://dev.azure.com/jussi0947/jussi/_build/results?buildId=5912&view=logs&j=5015ea34-1c35-577b-b206-c8bea8465a58&t=d96bae21-2af5-55ef-6693-add22ffd7f07):

| Download                        | Time  |
| ------------------------------- | :---: |
| `ninja-win.zip`                 | 00:02 |
| `pkg-config.exe`                | 00:01 |
| `msmpisdk.msi`                  | 00:00 |
| `MSMpiSetup.exe`                | 00:00 |
| `boost_1_72_0-msvc-14.1-64.exe` | 29:53 |
| `dmd.2.091.0.windows.zip`       | 00:04 |

Entire setup process: 35:34

So we are currently spending more than 30min on setting up our CI environment and downloading boost alone eats 30min in some instances. I suspect that there are some server-side issues on sourceforge.net where the boost exe is hosted.

jon-turney:
I think a MSVC boost is (now) pre-installed in the azure worker image, so perhaps we don't need to install it at all?

mensinda:
Looks like it: https://github.com/actions/virtual-environments/blob/master/images/win/Windows2016-Readme.md#boost

I will try a run without packaging boost and see how it goes.

mensinda:
It builds but the tests fail for some reason with the preinstalled boost version...

mensinda:
Nevermind, I just had to update `Path` to make the builtin boost version work. I am still in favor of bundling all dependencies in one place since it reduces the potential points of failure.

jpakkane:
mesonbuild/cidata#2 has been merged. Is this ready now?

mensinda:
Now it is ready (assuming the CI stays green). I have moved the CI bundle from my personal for to https://github.com/mesonbuild/cidata/releases/tag/ci1.

PR #6877, opened 2020-03-30
backend/vs: Fix link of wmain / wWinMain based apps

### LINKED ISSUE #6698: --backend=vs2019 EntryPointSymbol incorrect for 'UNICODE' projects
**Describe the bug**
vcxproj files generated by meson have an incorrect EntryPointSymbol for UNICODE applications (ie. those that have an entry point of wmain (console) or wWinMain (gui)).

```
2>MSVCRTD.lib(exe_main.obj) : error LNK2019: unresolved external symbol main referenced in function "int __cdecl invoke_main(void)" (?invoke_main@@YAHXZ)
2>.\main.exe : fatal error LNK1120: 1 unresolved externals
```

Removing EntryPointSymbol from all generated .vcxproj files works around the issue.

Ninja builds do not have this issue.

**To Reproduce**
meson.build:
``` meson
project('wmain', 'c')

executable('main', 'wmain.c')
```
wmain.c:
``` c
#include <stdio.h>

int wmain(int argc, wchar_t *argv[]) {
    printf("Hello, World!\n");
}
```

**Expected behavior**
main.exe should be generated instead of a link error.

**system parameters**
This is a native build, with Visual Studio 2019, on Windows 10, with meson and ninja from the official 0.53.1 MSI (I looked at the code, and it's still the same in master on github).

peterh:
Executables may have an entry point of wmain or wWinMain. These
executables link successfully with ninja. Rather than add more flags to
executable() in meson.build, remove the EntryPointSymbol override. This
makes the vs backend behave more like the ninja backend.

Fixes #6698

### DIFF
--- a/mesonbuild/backend/vs2010backend.py
+++ b/mesonbuild/backend/vs2010backend.py
@@ -744,19 +744,16 @@ class Vs2010Backend(backends.Backend):
 
     def gen_vcxproj(self, target, ofname, guid):
         mlog.debug('Generating vcxproj %s.' % target.name)
-        entrypoint = 'WinMainCRTStartup'
         subsystem = 'Windows'
         self.handled_target_deps[target.get_id()] = []
         if isinstance(target, build.Executable):
             conftype = 'Application'
             if not target.gui_app:
                 subsystem = 'Console'
-                entrypoint = 'mainCRTStartup'
         elif isinstance(target, build.StaticLibrary):
             conftype = 'StaticLibrary'
         elif isinstance(target, build.SharedLibrary):
             conftype = 'DynamicLibrary'
-            entrypoint = '_DllMainCrtStartup'
         elif isinstance(target, build.CustomTarget):
             return self.gen_custom_target_vcxproj(target, ofname, guid)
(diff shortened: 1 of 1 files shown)


peterh:
The one failure here appears to be a timeout of the test infrastructure, not an actual failure. Is there a way to ask for a retry?

jpakkane:
This is probably the correct fix but may cause unexpected failures. Do you have any idea if this could case regressions for certain project setups?

peterh:
Given that the removed values are the ACP defaults, I can't imagine that removing them makes any difference for existing projects. Especially since the ninja back-end doesn't have this at all (I expect build systems to typically use ninja for performance reasons. From my perspective, the vcxproj files are only to make debugging easier).

I have occasionally wanted to set EntryPointSymbol for a DLL, but there's no way to set it to a non-default value in meson. I suppose it's possible that someone might have a workflow that involves generating the vcxproj with meson and then modifying it with sed (instead of adding the value with a tool that groks XML) before building, but that seems unlikely to me.

PR #6896, opened 2020-04-03
qt dependency: Pick debug libraries based on b_vscrt

nirbheek:
`b_vscrt` has existed forever, and is the canonical source for which CRT to link to, and hence whether to use the debug libraries or not.


nirbheek:
Azure job is a runner failure, needs that job to be restarted by @jpakkane (or I can re-push and force all jobs to re-run)

jpakkane:
Let's try to run this this again once the CI fix merge is in.

PR #6900, opened 2020-04-04
ci: Add CUDA image (Arch based)

mensinda:
Also, a small fix for the OpenSUSE image.


mensinda:
The `OS Comp Tests / CUDA` test failure is expected, since the image is not yet pushed.

PR #6901, opened 2020-04-04
ci: Travis: Use the new mesonbuild/eoan image

mensinda:


jpakkane:
Dub seems to be failing:

    test cases/d/11 dub/meson.build:8:0: ERROR: Dependency "urld" not found

mensinda:
It looks more like there is a compiler mismatch: `ERROR: urld found but it wasn't compiled with ldc`. Is DMD supposed to be the default compiler for the image?

mensinda:
Fixed. The CI is green now.

PR #6903, opened 2020-04-04
Always perform cross build if an exe wrapper is needed

oleavr:
To revive support for cross-compiling x86_64 binaries for Apple's iOS
Simulator from a macOS build machine.

Seeing as the same assumption is made in other places, I also made the
same adjustments there.


oleavr:
Hmm, sorry, this definitely needs more work, or perhaps an entirely different fix.

jpakkane:
Ok, closing this then to reduce clutter on the MR page. Please reopen when ready.

PR #6910, opened 2020-04-05
Always perform cross build if an exe wrapper is needed

oleavr:
To revive support for cross-compiling x86_64 binaries for Apple's iOS
Simulator from a macOS build machine.

(Second attempt, hoping I got the corner-cases figured out now 🤞)

### DIFF
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -734,7 +734,7 @@ class Backend:
                 # E.g. an external verifier or simulator program run on a generated executable.
                 # Can always be run without a wrapper.
                 test_for_machine = MachineChoice.BUILD
-            is_cross = not self.environment.machines.matches_build_machine(test_for_machine)
+            is_cross = not self.environment.machines.matches_build_machine(test_for_machine) or self.environment.need_exe_wrapper(test_for_machine)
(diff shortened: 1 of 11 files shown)


mesonbuild/backend/backends.py
```diff
@@ -734,7 +734,7 @@ def create_test_serialisation(self, tests):
                 # E.g. an external verifier or simulator program run on a generated executable.
                 # Can always be run without a wrapper.
                 test_for_machine = MachineChoice.BUILD
-            is_cross = not self.environment.machines.matches_build_machine(test_for_machine)
+            is_cross = not self.environment.machines.matches_build_machine(test_for_machine) or self.environment.need_exe_wrapper(test_for_machine)
```
jpakkane, line 737: This is actually wrong. The correct way to check for this is `env.is_cross_build()`. The logic for deciding whether cross building should be in one place only (and the correct logic for cross building is "has a cross file", this logic changed recently and there are apparently places we missed).

oleavr: Yay. So this problematic cleverness of trying to short-circuit the cross build is a thing of the past?

oleavr:
Superseded by #6931.

PR #6909, opened 2020-04-05
boost: Fix library version (fixes #6899)

mensinda:
No test case since this requires two boost versions in the image and this is currently not doable with any existing distros.

PR #6911, opened 2020-04-05
ci: Add Ubuntu Bionic image

mensinda:
Also includes some documentation fixes and additional options for skipping tests if a tool version is to old.

### DIFF
--- a/.github/workflows/images.yml
+++ b/.github/workflows/images.yml
@@ -31,6 +31,7 @@ jobs:
           - { name: CUDA (on Arch),   id: cuda     }
           - { name: Fedora,           id: fedora   }
           - { name: OpenSUSE,         id: opensuse }
+          - { name: Ubuntu Bionic,    id: bionic   }
           - { name: Ubuntu Eoan,      id: eoan     }
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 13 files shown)


mensinda:
About the current failures in the new image:
- `1 boost` is expected to fail until #6909 is merged
- `23 hotdoc` fails with an `UnicodeEncodeError` exception
- two `_changes_cause_rebuild` fail, which looks really concerning to me
- also `LinuxlikeTests.test_pic` fails
- `test_guessed_linker_dependencies` fails on `self.assertRebuiltTarget`

Can someone check if these errors also occur on a real system, if someone is already running Ubuntu Bionic or has a Bionic VM (or does anyone already know why the tests fail)?

Similar to #6900, `OS Comp Tests / Ubuntu Bionic` fails because it cannot test what doesn't exist yet.

jpakkane:
Dammit, I upgrade my last Bionic machine (well, work machine) just today. I'll have to see if I have a VM image lying somwhere.

mensinda:
Thanks, otherwise I will set up one. Let's hope I didn't discover some sort of regression :)

jpakkane:
Hotdoc can be ignored. Nobody is using it on Bionic, it's too new for that.

mensinda:
The pylint error is unrelated and the bionic test fails because the image does not exist yet.

jpakkane:
I tried rerunning the tests but they still failed.

mensinda:
@jpakkane Which test are you concerned about? The pylint failure is a known issue and `OS Comp Tests / Ubuntu Bionic (pull_request)` fails because the docker image does not exist (yet) and will only be pushed once this PR is merged.

PR #6913, opened 2020-04-06
travis: Fix Travis OSX build

mensinda:


mensinda:
Halfway through, this turned a bit into a refactoring since writing bash scripts in the `.travis.yml` is neither readable nor fun.

PR #6915, opened 2020-04-06
Handle BuildTarget as custom_target() argument

### LINKED ISSUE #6914: custom_target() doesn't accept SharedLibrary as command argument
When using a library as command argument, you get a RuntimeError:
is of unknown type <class 'mesonbuild.build.SharedLibrary'>

```
lib = library('foo', sources)
abidw = find_program('abidw')

custom_target('abi',
  output : 'libfoo.abi',
  command : [abidw, lib]
)
```

That seems enough:
```diff
diff --git a/mesonbuild/backend/backends.py b/mesonbuild/backend/backends.py
index 687b12265..b220bc730 100644
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -960,7 +960,7 @@ class Backend:
         # Evaluate the command list
         cmd = []
         for i in target.command:
-            if isinstance(i, build.Executable):
+            if isinstance(i, (build.SharedLibrary, build.SharedModule, build.Executable)):
                 cmd += self.exe_object_to_cmd_array(i)
                 continue
             elif isinstance(i, build.CustomTarget):
```

elmarco:
Should fix  #6914


xclaesse:
Sounds good, will see if CI gets green.

elmarco:
> Sounds good, will see if CI gets green.

ok except:
E: Failed to fetch https://storage.googleapis.com/bazel-apt/dists/stable/jdk1.8/binary-amd64/Packages.gz  Hash Sum mismatch

test cases/common/93 selfbuilt custom/meson.build
```diff
@@ -14,3 +14,11 @@ hfile = custom_target('datah',
 main = executable('mainprog', 'mainprog.cpp', hfile)
 
 test('maintest', main)
+
+lib = library('libtool', 'tool.cpp')
+
+ctlib = custom_target('ctlib',
+  output : 'ctlib.out',
+  capture : true,
+  command : [main, lib],
```
jpakkane, line 23: Custom targets are not built by default, thus this is not actually run. And it would probably not run on Windows because shared libs are nor directly runnable (IIRC, and they might not be runnable on Linux either but I don't know the specifics).

A better test would use the generated file as an argument to some other command (such as a script that checks that the output file is actually generated).

elmarco: yes, the point was rather to check that meson accept the argument, not really to run it. I have updated it so it will run by default with a basic test (not perfect though, I don't think we need to do much more).

obilaniu: @jpakkane It's exceedingly rare, but Linux shared libraries can be given entry points and then be executed. In particular, GNU libc `/lib64/libc.so.6` and the dynamic loader `/lib64/ld-linux-x86-64.so.2`, executed with no arguments, print information about themselves and exit, and obviously `/lib64/ld-linux-x86-64.so.2` serves as the "interpreter" of all dynamically-linked programs on x86-64 Linux.

PR #6917, opened 2020-04-06
cmake: support cmake config file syntax

dabrain34:
remove useless assert when parsing cmake syntax config files.
tested with openjpeg cmake config files.

### DIFF
--- a/mesonbuild/mesonlib.py
+++ b/mesonbuild/mesonlib.py
@@ -908,15 +908,21 @@ def do_replacement(regex: T.Pattern[str], line: str, variable_format: str,
             return var
     return re.sub(regex, variable_replace, line), missing_variables
 
+def do_define(regex: T.Pattern[str], line: str, confdata: 'ConfigurationData', variable_format: str) -> str:
+    def get_cmake_define(line: str, confdata: 'ConfigurationData') -> str:
+        arr = line.split()
+        define_value=[]
+        for token in arr[2:]:
+            try:
+                (v, desc) = confdata.get(token)
+                define_value += [v]
+            except KeyError:
+                define_value += [token]
+        return ' '.join(define_value)
 
-def do_define(line: str, confdata: 'ConfigurationData', mesondefine: bool) -> str:
     arr = line.split()
-    if mesondefine:
-      if len(arr) != 2:
-        raise MesonException('#mesondefine does not contain exactly two tokens: %s' % line.strip())
-    else:
-      if len(arr) not in (2, 3):
-        raise MesonException('#cmakedefine does not contain two or tree tokens: %s' % line.strip())
+    if variable_format == 'meson' and len(arr) != 2:
+      raise MesonException('#mesondefine does not contain exactly two tokens: %s' % line.strip())
 
     varname = arr[1]
     try:
@@ -931,12 +937,27 @@ def do_define(line: str, confdata: 'ConfigurationData', mesondefine: bool) -> st
     elif isinstance(v, int):
         return '#define %s %d\n' % (varname, v)
     elif isinstance(v, str):
-        return '#define %s %s\n' % (varname, v)
+        if variable_format == 'meson':
+            result = v
+        else:
+            result = get_cmake_define(line, confdata)
+        result = '#define %s %s\n' % (varname, result)
+        (result, missing_variable) = do_replacement(regex, result, variable_format, confdata)
+        return result
     else:
         raise MesonException('#mesondefine argument "%s" is of unknown type.' % varname)
 
 def do_conf_str (data: list, confdata: 'ConfigurationData', variable_format: str,
                  encoding: str = 'utf-8') -> T.Tuple[T.List[str],T.Set[str], bool]:
+    def line_is_valid(line : str, variable_format: str):
+      if variable_format == 'meson':
+          if '#cmakedefine' in line:
+              return False
+      else: #cmake format
+         if '#mesondefine' in line:
+            return False
+      return True
+
     # Only allow (a-z, A-Z, 0-9, _, -) as valid characters for a define
     # Also allow escaping '@' with '\@'
     if variable_format in ['meson', 'cmake@']:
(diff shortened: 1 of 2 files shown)


mesonbuild/mesonlib.py
```diff
@@ -877,7 +877,6 @@ def do_replacement(regex: T.Pattern[str], line: str, variable_format: str,
         start_tag = '${'
         backslash_tag = '\\${'
     else:
-        assert variable_format == 'meson'
```
mensinda, line 880: Why do you want to remove this assert?

mesonbuild/mesonlib.py
```diff
@@ -911,8 +910,8 @@ def variable_replace(match: T.Match[str]) -> str:
 
 def do_mesondefine(line: str, confdata: 'ConfigurationData') -> str:
     arr = line.split()
-    if len(arr) != 2:
-        raise MesonException('#mesondefine does not contain exactly two tokens: %s' % line.strip())
+    if len(arr) < 2 and len(arr) > 3:
```
mensinda, line 913: There is no way this expression can ever be true. No array can have less than two **and** more than three elements at the same time.

dabrain34: it can if a statement is only having mesondefine. I was about to remove this exception as it prevents the use of this statement;

```
#cmakedefine _FILE_OFFSET_BITS @_FILE_OFFSET_BITS@
```
with cdata defines as

configure_file(input: 'config.h.cmake.in', output : 'config.h', configuration : cdata, format : 'cmake@')

mesonbuild/mesonlib.py
dabrain34: because it prevents the use of 'cmake@'  format.

mensinda:
Could you please describe the problem you are currently having in more detail? Preferably with a simple test project that demonstrates the issue.

mensinda:
See the [meson documentation](https://mesonbuild.com/Configuration.html) for more information about `configure_file()`.

dabrain34:
```
#cmakedefine _FILE_OFFSET_BITS @_FILE_OFFSET_BITS@
```
with cdata defines as

```
configure_file(input: 'config.h.cmake.in', output : 'config.h', configuration : cdata, format : 'cmake@')
```

With this statement in the meson build file, I was raising two assert, one with the variable_format and another one with the number of line's elements

mesonbuild/mesonlib.py
mensinda: Then instead of removing the assert (which is there for a reason to catch bugs), you should fix it instead to also allow the `cmake@` format. Maybe something like `assert variable_format in ['meson', 'cmake@']`.

mesonbuild/mesonlib.py
mensinda: You are not getting my point that this current if statement is pointless since it is logically impossible to ever be true. No number can be less than 2 **and** more than 3 at the same time. Are you sure the `and` isn't supposed to be an `or`?

Also, while `#cmakedefine` can indeed have more than two parameters, `#mesondefine`  is more strict and only allows the format `#mesondefine FOO`. So fixing `#cmakedefine` should not also affect `#mesondefine`.

mensinda:
Also, please make sure that this is a proper fix and `#cmakedefine` now works as specified in the [official CMake docs](https://cmake.org/cmake/help/latest/command/configure_file.html).

For adding tests, I would recommend splitting `do_conf_file`, so that all the file handling is still done in `do_conf_file` and all the string substitution is done in a new function. This will make adding tests in `run_unittests.py` much easier since you no longer have to work with extra files.

mesonbuild/mesonlib.py
dabrain34: :+1:

mesonbuild/mesonlib.py
dabrain34: > You are not getting my point that this current if statement is pointless since it is logically impossible to ever be true. No number can be less than 2 **and** more than 3 at the same time. Are you sure the `and` isn't supposed to be an `or`?
I really wanted to avoid unconsciously the `assert` :wink: Fixed !
> 
> Also, while `#cmakedefine` can indeed have more than two parameters, `#mesondefine` is more strict and only allows the format `#mesondefine FOO`. So fixing `#cmakedefine` should not also affect `#mesondefine`.
Thanks for this clarification, I did not know about `mesondefine` format.

dabrain34:
Working on a unit test for that.

dabrain34:
I just landed a fixup including a unit test for the three config file format supported.

When i'm trying to test the undef usecase, I'm facing an error that the var is not available in confdata 
>           (v, desc) = confdata.get(varname)
E           TypeError: 'NoneType' object is not iterable

Do you know @mensinda how could I have this use case addressed ?

mensinda:
This is most likely because the function expects a `ConfigurationData` object and not a dict.

mensinda:
Also, please split `do_conf_file` into two functions (`do_conf_file` and `do_conf_str` maybe) as described above. This way you can avoid all the temporary files in the unit tests and the code will be more reusable and much cleaner.

dabrain34:
hope its good (I secretely rant on unit_tests env :smiling_imp: )

mesonbuild/mesonlib.py
```diff
@@ -959,13 +958,25 @@ def do_conf_file(src: str, dst: str, confdata: 'ConfigurationData', variable_for
...
+
+def do_conf_file(src: str, dst: str, confdata: 'ConfigurationData', variable_format: str,
+                 encoding: str = 'utf-8') -> T.Tuple[T.Set[str], bool]:
+    try:
+        with open(src, encoding=encoding, newline='') as f:
+            data = f.readlines()
+    except Exception as e:
+        raise MesonException('Could not read input file %s: %s' % (src, str(e)))
+
+    (result, missing_variables, confdata_useless) = do_conf_str (data, confdata, variable_format, encoding)
```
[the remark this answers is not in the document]
dabrain34: c coder habits ;)

dabrain34:
I'm not sure to follow you, regarding the last request.

What do you mean by failing test ? Any example in mind ?

There is an exception if there is more than 2 or 3 tokens with cmakedefine. 

Do you want to catch the exception ?

I'd appreciate a code snippet to follow your mind :)

mensinda:
I was asking for two (or more) different tests. So, it should be tested that an exception is raised (via assertRaises for instance) when the format is wrong. Also, the example I have provided should *work* not fail, because it works with CMake (and produce the same output as CMake).

mensinda:
@dcbaker That CI job is currently broken everywhere (CUDA is fixed in master). I am going to look into the Azure failures tomorrow.

PR #6920, opened 2020-04-07
Add basic support to build common configure/make projects

xclaesse:
This adds a new Meson module "unstable_external_project" to build standard configure/make subprojects. It is similar conceptually to the cargo module we already discussed a lot.

The concept:
-  An autotools (or any configure/make) subproject requires a `meson.build` file at the root source dir that would use the `unstable_external_project` module to define configure options and extract dependencies objects that can then be used in the rest of Meson project.
- During meson's setup phase, the configure (or autogen.sh) script is executed out-of-tree, in the build directory.
- During configure, PKG_CONFIG_PATH is set to point to `<builddir>/meson-uninstalled` so the subproject can use pkg-config to get information about dependencies built by meson.
- During configure, language specific flags are passed from meson using CC, CFLAGS, etc environment. This works even for cross compilation.
- At build time, `make` in called by an helper script. The helper script generates a depfile by simply listing all files in the subproject so ninja knows when to rebuild.
- `make install` is then called with `DESTDIR` pointing to a directory inside build directory. That means that all files are in a predictable state and location.
- At installation phase, the DESTDIR is simply copied as-is since it already has the right prefix/libdir/etc structure.
- The module provides a `dependency()` method that returns an InternalDependency object that has `-I` and `-L` args pointing to the DESTDIR locations. Those dependency objects can be used to link meson code against those libraries.
- When a source file is modified, ninja will execute `make` custom target, and we assume the external build system (often autotools) is smart enough to reconfigure itself if needed.

Current limitations:
- Cannot run executables from the external project because LD_LIBRARY_PATH won't be set appropriately.
- Cannot run any executable built by meson that links on a library from an external project, for the same reason. I guess it should be possible to set the rpath, but I'm not sure how that part works.


jpakkane:
No! This will not be done. Supporting Autotools as a subproject is not tenable, it is nothing but an endless source of maintenance nightmares. An absolute requirement for adding this sort of thing is that the build system has builtin support for it (which CMake kind of has) or an active upstream development team willing to do this (which Cargo almost has). Autotools has neither. If you ever find yourself in this situation the way to approach is one of:

- convert the build system to Meson and submit it to wrapdb
- use a build aggregator such as Conan or Cerbero
- build and install to your own staging directory by hand the old fashioned way

The best we could do is to add some staging-like functionality like with Conan but it would still need to keep Autotools fully separate and isolated. Support for building Autotools projects as Meson subprojects will NOT BE ACCEPTED UNDER ANY CIRCUMSTANCES. Not now, not ever!

xclaesse:
Can you please be a little bit open minded and accept you are not the only one using and developing meson? I'm a bit tired of your dictator behavior. At the very least please provide concrete reason why it won't work.

This module is already more advanced than the cmake (e.g. support cross build) one with like 1000 times less code.

xclaesse:
This is a module that I offer to maintain, feel free to ignore any pr and issue related to it if you don't have time, which I perfectly understand. I think by my chin l countless contributions to meson project I deserve more respect.

elmarco:
So it support for cmake, but not for autotools?  autotools also "kind of have" subproject support.

Nothing is perfect, but I bet @xclasse didn't just work on that for fun (like most meson contributions!)

xclaesse:
@elmarco There is a cmake module that is horribly complicated and totally failed with every single cmake project I tried. I made a few patches to make it kind of work with a few project, but gave up because it's just too complicated and error prone. It got accepted in meson because @jpakkane wrongly think it's the good approach. But the previous version of my PR supported cmake as well, and it was working for all cmake projects and even in cross build case, with just about 200 lines of code pretty trivial to maintain.

mensinda:
As the maintainer of the (admittedly) large and "horribly complicated" CMake submodule, I would like to add the approach used there (generating a `meson.build` from introspection information) is the only feasible one I have found to make CMake subprojects work. It also has the advantage of *complete* control over the build process. See our IRC discussion a few weeks ago about the state of pkg-config in CMake.

However, I am not pretending that this approach is perfect (yes code size is an issue, but there is not much that can be done about it since three different CMake APIs have to be supported). I would also like to know which CMake projects you tried, so I can add support for them.

With that out of the way, it is clear that the CMake approach can never work with autotools (lack of introspection, etc). So `uninstalled.pc` files seem to be the only mechanism that could work. This PR is also a clear improvement to #4321, since the subproject is no longer compiled during the build step.

I would also like to add that the initial CMake subprojects also only worked for simple and small projects (no custom targets, etc.). So I disagree with closing this PR, but it should also be clear from the documentation that this feature is experimental (already given) and may only work for small projects and break in unexpected ways.

@jpakkane even from your comment above, I can't tell why you are against this module. From my point of few, there are a lot of similarities to the CMake module: It doesn't affect the main codebase, requires no additional meson features to function, and has a clear maintainer willing to improve and fix it over time.

Yes, autotoos does not support introspection in any way, but in my opinion, this is not a requirement for the approach in this PR. However, I am also not a big fan of calling `make` inside of `ninja` since it can easily break the subproject encapsulation and interfere with the meson build directory. Though, the risk of this happening is relatively low since you have to actively design a project to do so. Clearly warning about the implications about this in the documentation should be enough.

That being said, I do have some general questions regarding this PR:
- what happens if `./configure` ignores CFLAGS, CC, CXX, etc.
- how are incompatible compile flags handled (for instance there is the [`/MT`, `/MD` nonsense on Windows](https://docs.microsoft.com/en-us/cpp/build/reference/md-mt-ld-use-run-time-library?view=vs-2019))
- can the subprojects be installed
- regarding `dependency_method`
  - is there no better way of handling dependencies besides guessing the include and lib dir 
  - will this work on Windows, or even on Ubuntu with its (lets called it interesting choice) for installed libraries
  - Is it possible to check the validity of the dependency name

jpakkane:
The basic problem is the fact that if the project has dependencies of its own that do not come from the system (especially ones you build yourself) then passing those in is problematic. This same problem is with all foreign subprojects, though.

The bigger conceptual issue is that projects might not work if they are run from an install location that does not match their `prefix`. Hardcoding paths to `/usr/share/libname` is the usual culprit.

But going back to the staging case I mentioned above, could this be done entirely within the wrap file? That is, specify all the things needed in that, rather than in the build file? This would fit with the theme of moving `fallback` data inside the wrap files somehow. Then it would be easier to make clear to people that the use of external projects is fully a "staging" event (run at Meson configure time) and not a "build" event (run during `ninja`). This also naturally tells people that all the really difficult things (sharing deps) between projects does not really work (and yes, people have asked for this with the CMake integration, for all of LLVM no less).

> Can you please be a little bit open minded and accept you are not the only one using and developing meson?

I may have been to hasty in closing this. Sorry about that. But the unfortunate reality of managing any project is limiting the scope of what you do. Trying to support every use case and combination eventually leads to a incoherent unusable monstrosity like has happened with so many build systems thus far. Sometimes it means saying no to features, even if they were reasonable just because they would, for example, cause an ongoing maintenance burden.

xclaesse:
@mensinda Sorry if I sounded harsh on you cmake module, wanted the make the point that not working in all cases and having a lot of maintenance burden is not a reason for not doing it. Things does improve over time, we get a lot of contributions. ;-)

>     * what happens if `./configure` ignores CFLAGS, CC, CXX, etc.

Does it happen? I Haven't see a configure-like project not supporting those env, but I guess we could list in the doc that limitation. One possible solution if that happen is to add `{cc}` and `{cflags}` tokens to `add_project()` options so they get replaced by the compiler/flags, assuming the configure script would take them from command line instead of env.

Note that one relatively big requirement is to be able to build out-of-tree, that's something many custom configure-like scripts, or even badly written autotools, gets wrong. But in practice anything packaged in a distro does support out-of-tree build.

>     * how are incompatible compile flags handled (for instance there is the [`/MT`, `/MD` nonsense on Windows](https://docs.microsoft.com/en-us/cpp/build/reference/md-mt-ld-use-run-time-library?view=vs-2019))

Not sure to understand what you mean here. I just pass all flags from meson to CFLAGS env, if there are incompatible flags then it won't work for Meson neither.

>     * can the subprojects be installed

Yes, at build time a custom_target() will do `make install` with DESTDIR set inside Meson's build directory. At install time meson will simply copy as-is that folder since it already has the correct final layout (assuming configure script respected prefix/libdir/etc).

>     * regarding `dependency_method`
>       
>       * is there no better way of handling dependencies besides guessing the include and lib dir

It's not guessing since we pass the prefix and libdir, we know the lib is there. That being said, it would be nice to use the pkg-config file if the subproject creates one. The problem is Meson needs those flags at configure time, and most of the time the .pc file will be generated only at build time. That's the reason why in my previous PR I did the build at meson's configure time. The other reason why I stopped using pkg-config files from the subproject is because it requires setting PKG_CONFIG_SYSROOT_DIR to fixup all paths prefixed by the DESTDIR, but in a cross build scenario the user already can set sys_root in his cross file to another location. Unfortunately pkg-config does not support having .pc files coming from multiple sysroots. That's a limitation we could solve by implementing pkg-config directly in meson (which is a crazy plan I had for a long time already).

>       * will this work on Windows, or even on Ubuntu with its (lets called it interesting choice) for installed libraries

I haven't tested it on Windows with real projects, but IIRC the CI of this PR does run on Windows. On ubuntu it obviously works since that's my distro. Not sure why you think Ubuntu could cause issues, I'm passing down to configure script the prefix/libdir from Meson.

>       * Is it possible to check the validity of the dependency name

Unfortunately not, because the library won't exist until build stage, and the dependency is needed at configure stage.

> The basic problem is the fact that if the project has dependencies of its own that do not come from the system (especially ones you build yourself) then passing those in is problematic. This same problem is with all foreign subprojects, though.

Many autotools project will use pkg-config for that, and it will pick -uninstalled.pc files generated by meson.

> The bigger conceptual issue is that projects might not work if they are run from an install location that does not match their `prefix`. Hardcoding paths to `/usr/share/libname` is the usual culprit.

That's true and it's the biggest limitation that I documented, we cannot execute any binaries from those subprojects, because it won't find its libraries. That limitation could be lifted if we set stuff like LD_LIBRARY_PATH, and rpath for exe built by meson. Regarding hardcoding paths like `/usr/share` the same can be said about any executable built by Meson too, not every exe can be run uninstalled.

> But going back to the staging case I mentioned above, could this be done entirely within the wrap file? That is, specify all the things needed in that, rather than in the build file? This would fit with the theme of moving `fallback` data inside the wrap files somehow. Then it would be easier to make clear to people that the use of external projects is fully a "staging" event (run at Meson configure time) and not a "build" event (run during `ninja`). This also naturally tells people that all the really difficult things (sharing deps) between projects does not really work (and yes, people have asked for this with the CMake integration, for all of LLVM no less).

I went away from building at meson configure time because now that we have -uninstalled.pc files we actually can manage dependencies pretty well. We can have an A->B->C where B is built with autotools and A/C are built by meson.

We could add the info in the wrap file, that's actually what my very first attempt did years ago, but it's not flexible at all. You could have to pass different configure args depending if it's cross build, or per platform, depending on project options, etc...  Having the power of meson interpreter helps a lot. It's also how cmake module works.

> > Can you please be a little bit open minded and accept you are not the only one using and developing meson?
> 
> I may have been to hasty in closing this. Sorry about that. But the unfortunate reality of managing any project is limiting the scope of what you do. Trying to support every use case and combination eventually leads to a incoherent unusable monstrosity like has happened with so many build systems thus far. Sometimes it means saying no to features, even if they were reasonable just because they would, for example, cause an ongoing maintenance burden.

I agree with that, and that's also why I opted for a solution self contained into a module, with really few changes in meson core. Most build systems allows 3rd parties to have their own extension modules, or even program into the build system language. Meson does not allow that, for good reason, but I think one trade off is to allow more "exotic" use-cases into upstreamed modules.

xclaesse:
> I haven't tested it on Windows with real projects, but IIRC the CI of this PR does run on Windows. 

Obviously you would need msys2 or some kind of unix shell, as always with autotools. If you don't have that, then of course your only choice is to port the dependency to Meson natively.

jpakkane:
> You could have to pass different configure args depending if it's cross build, or per platform, depending on project options, etc

Can you give a concrete example of this. What specific things would one want to set and how?

xclaesse:
> > You could have to pass different configure args depending if it's cross build, or per platform, depending on project options, etc
> 
> Can you give a concrete example of this. What specific things would one want to set and how?

One example is to bind a meson feature option to --enable-foo --disable-foo, the other case I faced is when cross compiling you sometimes need to set cached values for autotools:
```meson
mod = import('unstable_external_project')

extra_options = []
feature = get_option('foo')
if feature.enabled()
  extra_options += '--enable-foo'
elif feature.disabled()
  extra_options += '--disable-foo'
endif

env = environment()
if meson.is_cross_build()
  # Workaround broken autoconf test in cross compilation
  env.set('ac_cv_func_malloc_0_nonnull', 'yes')
  env.set('ac_cv_func_realloc_0_nonnull', 'yes')
endif

p = mod.add_project('configure',
  options : [
    '--prefix={prefix}',
    '--libdir=' + join_paths('{prefix}', '{libdir}'),
    '--includedir=' + join_paths('{prefix}', '{includedir}'),
  ] + extra_options,
  env : env,
)
```

Another example I face is when cross compiling I had to help a bit configure's poor detection code with stuff like:
`

xclaesse:
For feature binding, I'll probably add a method on the module that takes a boolean or feature option and return a --disable-foo --enable-foo string, the code snippet above does not scale much otherwise.

xclaesse:
Quick notes:
- I should replace the `'{prefix}'` syntax with `'@PREFIX@'` to be consistent with other APIs.
- Rename `options` kwarg to `configure_options` because we could want to add `build_options` too, I got told that kbuild could need `make K=V` syntax.

xclaesse:
>     * I should replace the `'{prefix}'` syntax with `'@PREFIX@'` to be consistent with other APIs.
>     * Rename `options` kwarg to `configure_options` because we could want to add `build_options` too, I got told that kbuild could need `make K=V` syntax.

done.

rhd:
I love this idea. Dealing with third party code is such a pain and porting everything to meson is sadly not an option.  @xclaesse thanks for taking the time to write (and support) this. I hope it gets upstreamed.

mesonbuild/modules/unstable_external_project.py
```diff
@@ -0,0 +1,250 @@
...
+        # On Windows if the prefix is "c:/foo" and DESTDIR is "c:/bar", `make`
+        # will install files into "c:/bar/c:/foo" which is an invalid path.
+        # Work around that issue by removing the drive from prefix.
+        self.prefix = os.path.splitdrive(self.prefix)[1]
+
+        # self.prefix is an absolute path, so we cannot use it in
+        # os.path.join(something, self.prefix).
+        self.rel_prefix = self.prefix[1:]
+
+        self.make = self.interpreter.find_program_impl(os.environ.get('MAKE', 'make'))
```
mensinda, line 64: Shouldn't `os.environ` be avoided?

mesonbuild/modules/unstable_external_project.py
```diff
@@ -0,0 +1,250 @@
...
+                ]
+        if self.verbose:
+            cmd.append('--verbose')
+
+        target_kwargs = {'output': '{}.stamp'.format(self.name),
+                         'depfile': '{}.d'.format(self.name),
+                         'command': cmd + ['@OUTPUT@', '@DEPFILE@'],
+                         'console': True,
+                         }
+        self.target = build.CustomTarget(self.name,
```
mensinda, line 172: `self.target` is not defined in `__init__`.

mesonbuild/scripts/externalproject.py
```diff
@@ -0,0 +1,88 @@
...
+    def write_stampfile(self):
+        with open(self.stampfile, 'w') as f:
+            pass
+
+    def build(self):
+        make_cmd = [self.make]
+        if not self.verbose:
+            make_cmd.append('--quiet')
+
+        build_cmd = make_cmd + ['-j%d' % multiprocessing.cpu_count()]
```
mensinda, line 53: Maybe use `.format()` or `+ str()` for consistency.

test cases/common/236 external project/libfoo/libfoo.c
```diff
@@ -0,0 +1,8 @@
+#include "libfoo.h"
+
+int func(void);
```
mensinda, line 3: Maybe also add a `func.h` instead of forward declaring to test that include paths work?

mensinda (commented):
This looks mostly good implementation wise (see the few minor nitpicks below). Also, please add typing and consider switching to `pathlib`

mesonbuild/modules/unstable_external_project.py
xclaesse: Is there another way to get $MAKE from env? I got told that it's commonly passed by env, like PKG_CONFIG, etc.

mesonbuild/modules/unstable_external_project.py
xclaesse: Is that a problem? I could maybe call _get_targets() from __init__(), is that better?

mesonbuild/scripts/externalproject.py
xclaesse: Good point, will update.

mesonbuild/modules/unstable_external_project.py
mensinda: If I remember correctly, there was a function that automatically gets the correct value from the env and machine files. Not sure how exactly it worked and where it is.

test cases/common/236 external project/libfoo/libfoo.c
xclaesse: right, done.

mesonbuild/modules/unstable_external_project.py
xclaesse: Ok you're right, find_program_impl('make') will now return the command from MAKE env, or machine file.

mesonbuild/modules/unstable_external_project.py
mensinda: It's mostly to so that all `self.*` are declared in `__init__()`. Otherwsie, accessing them before they are set in another function will result in an error. Also, IDEs will be able to do better autocompletion :)

I was thinking of setting `self.target = None` and then checking (or asserting) that `self.target` was set when it is used.

xclaesse: ok, done.

xclaesse:
> This looks mostly good implementation wise (see the few minor nitpicks below). Also, please add typing and consider switching to `pathlib`

Thanks a lot for the review, still need to make CI pass...

About pathlib, one thing I don't like with it is we have strings everywhere, so it requires wrapper all of them with pathlib.Path() where simple os.path.join() works directly with strings. Also it's significantly more complicated, it's well illustrated by the class diagram on the doc: https://docs.python.org/3/library/pathlib.html. Seriously, what are all those different objects, what 'pure' can possibly mean?

mensinda:
> It's of course a bit of chicken and egg problem

And that's exactly why I was suggesting to use pathlib for new code :)

> I remember there were bugs in Python 3.5 that we still have to support

Yes, but path joining and basic file IO works fine even with 3.5. `Path.resolve()` was the primary 3.5 pain point if I remember correctly.

> I thought pathlib is something mostly on the TODO list when we upgrade to newer Python, no?

It's already primarily used in new code (like the boost dependency) and basically everywhere in the codebase (although not consistently). Also, I am planning on refactoring all my CMake code to use pathlib (specifically to avoid issues with drive letters on Windows).

Also, I am hoping to "solve" the minimum python version issue with #7518, so we can upgrade faster to new python versions.

xclaesse:
This is exactly why I was scared to use pathlib: https://docs.python.org/3/library/os.path.html#os.path.splitdrive: `Changed in version 3.6: Accepts a path-like object.`. That means that for every single API I use, I need to carefully read the doc to see when it added support for pathlib.

mensinda:
I would say that everything in `os.*` is a bad example because there is a working replacement in `pathlib` and that already covers 95% of all use cases that you want to do with paths. And for the remaining 5% where you have to use other APIs you can just do `str(path)` or `path.as_posix()` depending on the situation.

For instance, this code

```python
self.prefix = os.path.splitdrive(self.prefix)[1]
self.rel_prefix = self.prefix[1:]
```
can be rewritten as (which is also cleaner IMHO):

```python
self.prefix = self.prefix.relative_to(self.prefix.drive + self.prefix.root)
```

mesonbuild/modules/unstable_external_project.py
```diff
@@ -0,0 +1,251 @@
...
+        self.includedir = self.env.coredata.get_builtin_option('includedir')
+
+        # On Windows if the prefix is "c:/foo" and DESTDIR is "c:/bar", `make`
+        # will install files into "c:/bar/c:/foo" which is an invalid path.
+        # Work around that issue by removing the drive from prefix.
+        self.prefix = os.path.splitdrive(self.prefix)[1]
+
+        # self.prefix is an absolute path, so we cannot use it in
+        # os.path.join(something, self.prefix).
+        self.rel_prefix = self.prefix[1:]
```
mensinda, line 62: Is this also the case when `self.prefix` is `c:Program Files`?

xclaesse: It's valid without `/` ? Damned Windows path :( Any idea how to make it relative in a portable way then?

mensinda: There is [this](https://stackoverflow.com/questions/23955968/path-with-no-slash-after-drive-letter-and-colon-what-does-it-point-to), so it probably isn't what it is intended but it sort of works.

`self.prefix = self.prefix.relative_to(self.prefix.drive + self.prefix.root)` is actually a protable way to solve this, since `self.prefix.root` will evaluate to an empty string in this case :P

xclaesse: Ok, you mark a point for justifying moving to pathlib :D

And drive+root is anchor as far as I understand.

mensinda: True... Somehow I missed anchor.

xclaesse: Updated to use pathlib, let me know if it's correct.

mesonbuild/modules/unstable_external_project.py
```diff
@@ -0,0 +1,253 @@
...
+        self.build_machine = build_machine
+        self.host_machine = host_machine
+        self.configure_command = configure_command
+        self.configure_options = configure_options
+        self.cross_configure_options = cross_configure_options
+        self.verbose = verbose
+        self.user_env = env
+
+        self.name = self.subdir.name
+        self.src_dir = Path(self.env.get_source_dir(), self.subdir)
```
mensinda, line 49: This is by no means wrong, but `Path(self.env.get_source_dir()) / self.subdir` would also work (only the first operand has to be a `Path` object).

This is just in case you didn't know already and boils down to personal preference though, so use what you like better.

xclaesse: I knew, thanks. To be fair I was lazy and started by simply s/os.path.join/Path/ :)

xclaesse:
@mensinda I extended a big requirements & limitations doc sections.

I disabled test on Windows because the dummy configure script is too dumb. It needs to know cl.exe command line, how to build a windows import lib, etc... I don't know how that works. I think that's something that can be added later when someone who knows Windows wants to contribute.

I also disabled test for cross compilation because it hits a pkg-config bug/limitation. They cannot use -uninstalled.pc files when PKG_CONFIG_SYSROOT_DIR is set. Cross compilation still works if the subproject does not need dependencies from Meson.

I think it's better to start with smaller scope and extend over time.

xclaesse:
@mensinda Do you have any idea why that Ubuntu bionic CI fails here? I don't understand that python backtrace :/

mensinda:
The `self.build_dir` shouldn't be a `Path` right? It looks like it's a problem with the file descriptors, but I have no idea what is going on here... Maybe try not using `/dev/null`?

xclaesse:
For reference, I opened a pkgconf issue to find a solution to the PKG_CONFIG_SYSROOR_DIR limitation: https://todo.sr.ht/~kaniini/pkgconf/20

xclaesse:
> The `self.build_dir` shouldn't be a `Path` right?

It's a string from command line options.

> It looks like it's a problem with the file descriptors, but I have no idea what is going on here... Maybe try not using `/dev/null`?

When self.verbose is True, I want ninja to output stdout from make command. It makes stuff more debugable, and also if you're compiling a big external project (why not webkit...) it would be weird to see no output for a long period. I'm suspecting that's an old Python bug, works fine in all other CI jobs. I'll just remove `verbose: true` from the unit test for now.

xclaesse:
Ok, CI passed now, the remaining failure is unrelated.

mensinda:
> I think it's better to start with smaller scope and extend over time.

True, this is also how the CMake module initially started :)

OK, now the only thing missing is typing *\*ducks away\**

But seriously, could you please at least do the script? Especially with #7657 also in the pipeline where I add type annotations for all other scripts.

Otherwise, this PR looks good to me, assuming @jpakkane agrees with the API and gives his final approval.

xclaesse:
@mensinda typing done.

xclaesse:
> For reference, I opened a pkgconf issue to find a solution to the PKG_CONFIG_SYSROOR_DIR limitation: https://todo.sr.ht/~kaniini/pkgconf/20

Call me crazy, I even made a MR for pkg-config... https://gitlab.freedesktop.org/pkg-config/pkg-config/-/merge_requests/7

mensinda (approved):
Thanks, LGTM.

xclaesse:
Thanks a lot @mensinda for your review.

docs/markdown/snippets/external_project.md
```diff
@@ -0,0 +1,24 @@
+## External projects
+
+A new experimental module `unstable_external_project` has been added to build
+code using other build systems than Meson. Currently only suppoting projects
```
elmarco, line 4: supporting

xclaesse: Thanks, fixed.

jpakkane:
The module is called external project but in reality it is specifically for Autotools (or imitations thereof). Maybe it should be called `unstable_autotools` to match the CMake module? Otherwise people are expecting to use it for any possible build system (plain makefiles, qmake, waf, what have you).

You should mark yourself as the maintainer for this in the code owners file.

The documentation should begin with a link to https://mesonbuild.com/Mixing-build-systems.html.

xclaesse:
It's called external_project because I got the inspiration from CMake's similar feature: https://cmake.org/cmake/help/latest/module/ExternalProject.html.

It's not strictly specific to autotools, currently it can be any build system where you have a configure script (could be custom ./configure bash script or ./autogen.sh, etc) that generate Makefile. In the future it can easily be extended to support other build tools than `make`. I'm expecting to extend for example with syntax like `add_project(..., build_command: find_program('ninja'))` in the case the configure script generates a ninja file instead of Makefile. I'm also expecting to extend the configure script name with some special cases like `qmake`, `waf` (https://github.com/mesonbuild/meson/issues/7638), `cargo`, `kbuild`, and will do some special treatments if needed. But before doing more complex uses cases, I would like to start small to get feedback.

For example, I've been using earlier version of this PR to build cmake projects before we had a specific cmake module. I would have to discusss with @mensinda what he thinks about adding back the special case for cmake in the `external_project` module, because it has multiple advantages like support cross compilation.

> You should mark yourself as the maintainer for this in the code owners file.

Good point, done.

> The documentation should begin with a link to https://mesonbuild.com/Mixing-build-systems.html.

Ok, copied the text from cmake module.

mensinda:
> I would have to discusss with @mensinda what he thinks about adding back the special case for cmake in the external_project module, because it has multiple advantages like support cross compilation.

First of all, cross-compilation support is still planned for the CMake module, I just didn't get around to implementing it yet :)

What are the other advantages `unstable_external_project` has?

In principle, I don't have anything against also supporting CMake, if there is a big note in the docs to only use `unstable_external_project` as a last resort if there is a fundamental issue that is impossible to solve with the CMake module. This is because the CMake module also has some advantages:

- complete control over the generated `meson.build` (see the new [options object](https://mesonbuild.com/CMake-module.html#cmake-options-object))
- does not call an external `make` or `ninja`

xclaesse:
@mensinda Note that I'm not suggesting we should do it in any near future, just that it is technically possible to do it. Your module has undeniably advantages too. I haven't used the cmake module since a long time, and I'm sure it involved a lot since then, but the issues I had with it were:
- Works only with some cmake projects. Every cmake API has to be translated by Meson so when an API/syntax is not supported it won't work. I guess that also means it's highly depending on cmake not breaking their API, or even adding new APIs, and output that gets parsed.
- cmake configuration does not have all the compiler args from meson, which is why cross compilation does not work, but I'm sure it causes issues in native compilation too. I did a patch to pass those args but then they are added to every generated meson target which caused issues. I suggested a long time ago that we should add for example `plain: true` kwarg to targets to tell that meson should not add any args to that target because cmake has put them already. Dunno if there are other solutions you envisaged?
- Complexity, it has about 10x more LOC, and a significant number of new issues/PRs are about cmake module. Which is really good to see so many contributions.

I think both solutions have pros & cons, the best approach really depends on user's use-case IMHO.

jpakkane:
We could, in theory, have a general external project module/method, but it would require creating an API that can be used to define configuration for all of them. This will most likely be very complicated and, at the very least, we don't have a good understanding of how it should look like (unknown unknowns and all that). Thus it would seem prudent to do specific modules first and once we have real world usage and possibly more than two different modules, then we can do the general one (and it may turn out that it is not actually needed for whatever reason).

xclaesse:
That's why it's called unstable. But I think it's generic enough to actually work with most half decent build systems. In the end it's really just passing the command line for configure/build. cmake's external project support is generic like that too.

xclaesse:
Ok, I think I replied to all concerns, the CI is green, time to merge.

Thanks a lot @mensinda for your review!

PR #6922, opened 2020-04-07
Add .to_string() method to Feature Objects

sgallagher:
Signed-off-by: Stephen Gallagher <sgallagh@redhat.com>

### DIFF
--- a/mesonbuild/interpreter.py
+++ b/mesonbuild/interpreter.py
@@ -79,6 +79,7 @@ class FeatureOptionHolder(InterpreterObject, ObjectHolder):
         self.methods.update({'enabled': self.enabled_method,
                              'disabled': self.disabled_method,
                              'auto': self.auto_method,
+                             'to_string': self.to_string_method,
                              })
 
     @noPosargs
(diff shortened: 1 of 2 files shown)


mesonbuild/interpreter.py
```diff
@@ -96,6 +97,19 @@ def disabled_method(self, args, kwargs):
...
+    def to_string_method(self, args, kwargs):
+        if self.held_object.is_enabled():
+            return 'enabled'
+        elif self.held_object.is_disabled():
+            return 'disabled'
+        elif self.held_object.is_auto():
+            return 'auto'
+
+        # Should never reach here
+        return 'unknown'
```
xclaesse, line 111: I think you should raise an exception instead of silently returning a value user won't expect.

sgallagher: Sure, I can do that.

sgallagher:
> Implementation looks good, except for a small nitpicking. Missing a release notes snippet and `@featurenew`.

I'll look into that.

> However, what's your use case?
I want to print all of the `meson_options.txt` values in my `summary()` call, but I can't print the value of a Feature object directly.

ePirat:
Wouldn't this mean people could accidentally do

```
if my_feature == "enabled"
```
and it would work, even though thats not how one is supposed to check this?

sgallagher:
> Wouldn't this mean people could accidentally do
> 
> ```
> if my_feature == "enabled"
> ```
> 
> and it would work, even though thats not how one is supposed to check this?

I don't *think* this would automatically do the string conversion; I think you would need to do `if my_feature.to_string() == 'enabled'`

At which point, it's not an accident...

ePirat:
Oh right I forgot meson does not implicitly convert, usually. So probably fine.

mesonbuild/interpreter.py
jpakkane: This still needs fixing. Since this is new functionality, this MR should add a release note snippet and update the `build-options` documentation page.

jpakkane: ping

sgallagher:
Sorry, I've been swamped with more important things for a while. I'll see if I can clean this up this week.

PR #6930, opened 2020-04-09
Fix architecture detection logic for VS backend to allow ARM64

jenatali:
VS2019 has support for ARM64 builds, but Meson can't target them.


jpakkane:
Thanks.

PR #6933, opened 2020-04-10
Fix outdated cross-compilation checks

oleavr:
(Original PR got closed inadvertently whilst updating it, sorry for the noise.)

### DIFF
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -768,7 +768,7 @@ class Backend:
                 # E.g. an external verifier or simulator program run on a generated executable.
                 # Can always be run without a wrapper.
                 test_for_machine = MachineChoice.BUILD
-            is_cross = not self.environment.machines.matches_build_machine(test_for_machine)
+            is_cross = self.environment.is_cross_build(test_for_machine)
             if is_cross and self.environment.need_exe_wrapper():
                 exe_wrapper = self.environment.get_exe_wrapper()
             else:
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -682,7 +682,9 @@ class CoreData:
                 if type(oldval) != type(value):
                     self.user_options[name] = value
 
-    def is_cross_build(self) -> bool:
+    def is_cross_build(self, when_building_for: MachineChoice = MachineChoice.HOST) -> bool:
+        if when_building_for == MachineChoice.BUILD:
+            return False
         return len(self.cross_files) > 0
 
     def strip_build_option_names(self, options):
--- a/mesonbuild/environment.py
+++ b/mesonbuild/environment.py
@@ -635,8 +635,8 @@ class Environment:
         self.coredata.meson_command = mesonlib.meson_command
         self.first_invocation = True
 
-    def is_cross_build(self) -> bool:
-        return self.coredata.is_cross_build()
+    def is_cross_build(self, when_building_for: MachineChoice = MachineChoice.HOST) -> bool:
+        return self.coredata.is_cross_build(when_building_for)
 
     def dump_coredata(self):
         return coredata.save(self.coredata, self.get_build_dir())
@@ -899,7 +899,7 @@ class Environment:
     def _detect_c_or_cpp_compiler(self, lang: str, for_machine: MachineChoice) -> Compiler:
         popen_exceptions = {}
         compilers, ccache, exe_wrap = self._get_compilers(lang, for_machine)
-        is_cross = not self.machines.matches_build_machine(for_machine)
+        is_cross = self.is_cross_build(for_machine)
         info = self.machines[for_machine]
 
         for compiler in compilers:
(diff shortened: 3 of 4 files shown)


mesonbuild/backend/backends.py
```diff
@@ -734,7 +734,7 @@ def create_test_serialisation(self, tests):
                 # E.g. an external verifier or simulator program run on a generated executable.
                 # Can always be run without a wrapper.
                 test_for_machine = MachineChoice.BUILD
-            is_cross = not self.environment.machines.matches_build_machine(test_for_machine)
+            is_cross = self.environment.is_cross_build() if test_for_machine == MachineChoice.HOST else False
```
jpakkane, line 737: This should only call `is_cross_build`. _All_ logic on deciding whether a build is a cross build _must_ be inside that function. Splitting it around like this is breaking DRY. If you need extra logic then it should go in a different named variable.

oleavr: @jpakkane Great point! Thank you. Just force-pushed a new version.

mesonbuild/environment.py
```diff
@@ -635,7 +635,9 @@ def create_new_coredata(self, options):
         self.coredata.meson_command = mesonlib.meson_command
         self.first_invocation = True
 
-    def is_cross_build(self) -> bool:
+    def is_cross_build(self, when_building_for: MachineChoice = MachineChoice.HOST) -> bool:
+        if when_building_for == MachineChoice.BUILD:
+            return False
```
jpakkane, line 640: This is wrong and splits the logic again. The logic must be inside `coredata.is_cross_build` and it is _exactly_ whether the user has defined a cross file or not. We have documented that publicly. What is the problem you are trying to work around by having this extra logic here?

oleavr: > This is wrong and splits the logic again.

Oops, sorry, I misunderstood the architecture.

> What is the problem you are trying to work around by having this extra logic here?

The reason I assumed there would be a need to differentiate is that (unless I'm mistaken) Meson may create up to two instances of each compiler, one instance for the build machine and one for the host machine. I assumed that the instance for the build machine should be constructed with `is_cross=False` (for the compilers that have `is_cross`). The reason I assumed this is that implementations may choose to use the `exe_wrapper` when `self.is_cross == True`, and it seems weird to build a binary for the build machine and try to run it using the configured wrapper (if set).

jpakkane:
After thinking about this I realized what the confusion was about. We unfortunately have two different usages for `is_cross` in the code base. It can either mean "are we doing a cross build or a regular one" or "is this compiler going to be a native compiler or a cross compiler". I kept confusing the two.

This whole thing would become more understandable if we changed all `is_cross` variable names here to either `is_cross_build` or `is_cross_compiler`. This would make things unambiguous.

Ericson2314:
@jpakkane So "are we doing a cross build or a regular one" is a question I think should almost never need to ask---what does it achieve?, whereas "is this compiler going to be a native compiler or a cross compiler" is fine question---even better are the question "is this platform weird enough to need a exe-wrapper?", and follow up "if so, do we have one?". Anything that steers people away from the first question is fine with me.

CC @dcbaker

jpakkane:
In the bug linked above, this MR is found to fix iOS emulator builds.

jpakkane:
The selfbuilt custom test fails with this:

```
ERROR: 
Cannot use target checkarg as a generator because it is built for the
host machine and no exe wrapper is defined or needs_exe_wrapper is
true. You might want to set `native: true` instead to build it for
the build machine.
```

FWICT tell this has been fixed in master. Please rebase so CI is run. This should be ready once that works.

dcbaker:
I think this fixes #7041 as well.

cpackham:
Looks like this also fixes  #7125

mesonbuild/coredata.py
```diff
@@ -682,7 +682,9 @@ def merge_user_options(self, options):
                 if type(oldval) != type(value):
                     self.user_options[name] = value
 
-    def is_cross_build(self) -> bool:
+    def is_cross_build(self, when_building_for: MachineChoice = MachineChoice.HOST) -> bool:
+        if when_building_for == MachineChoice.BUILD:
+            return False
```
nirbheek, line 687: This is not correct. This will break, f.ex., building with MSVC and targeting x64-UWP when building on an x64 machine. The machine architecture is not the final say on whether we're cross compiling.

oleavr: @nirbheek This is when building a binary for the build machine, not the host machine.

nirbheek: Specifically, in the UWP case, the target machine is x64, but the compiler is a cross-compiler (you cannot run its output because it depends on libraries that are not available), and the build is a cross build (for the same reason).

oleavr: I thought targets that produce binaries for the *build* machine would always produce outputs that the build machine can run? Such as a tool for generating code, which could be used to compile binaries for the host. (Which obviously the build machine may not be able to run.) Either we're talking about different things, or I'm missing something.

Ericson2314: @nirbheek the two choices here are "build" and "host". I think this is correct and we are just getting hung up on the naming.

Ericson2314: In particular, there is absolutely no comparing of platforms and what-not here. Basically the logic is "host machine + have cross file => is cross". You could have the native and cross files be identical and it would still say is cross.

nirbheek: Ahh, I misunderstood what this code is doing, my bad. Also my second comment was posted before I saw your reply to my comment, sorry for the confusion. Github didn't update with your comment.

tbodt:
When can I expect a release that includes this?

PR #6935, opened 2020-04-10
ci: Fix cuda image

j-carl:
Install missing package gtest.


mensinda:
There is a more general issue with the CUDA image. Specifically that the arguments do not appear to be passed correctly, thus the non CUDA tests are not skipped. See the difference in OS Comp Tests and the CI image builder.

PR #6940, opened 2020-04-10
wrap: Add fallback urls

xclaesse:
It can happen that a server is temporaly down, tarballs often have
many mirrors available so we should be able to add at least one fallback
mirror in wrap files.


jpakkane:
This most definitely needs doc updates. Preferably also a test, but network things are usually not really doable.

xclaesse:
Pushed to soon, I'll add doc. I don't think we have tests that actually download, do we? Does file:// URL works to fake download inside meson tree?

xclaesse:
Ok, we even already had a unit test for file://. Doc and test now added.

PR #6941, opened 2020-04-10
backend: Fix file name confilcts by using the complete path

### LINKED ISSUE #6928: ninja/cmake issue when importing large subproject
**Describe the bug**
I am using meson to compile a complicated project. Part of this project involves sharing C++ source between my x86 spec PC and an ESP32 micro-controller. When I attempt to connect the ESP32 SDK's cmake project to my meson project as a subproject, ninja stops with error:

`ERROR: Multiple producers for Ninja target "subprojects/esp32/55ac3a2@@cm_mbedcrypto@sta/bignum.c.o". Please rename your targets.`

**To Reproduce**
You'll need the ESP-IDF toolchain, setup guide here:
https://docs.espressif.com/projects/esp-idf/en/latest/esp32/get-started/

I've created an example repo with a trivial ESP32 main implementation as an example here:
https://github.com/Michael-Densberger/meson_esp32_issue_example

**Expected behavior**
I expected that the cmake subproject would build and I could use custom meson targets to interact with the idf.py wrapper that is used by the ESP-IDF toolchain to manage the micro-controller. Indeed, calling `idf.py flash monitor` from the subproject/esp32 dir results in a compiling and running binary on my ESP32 micro-controller. 

To clarify:
1. using cmake through the idf.py wrapper:
    works with no issues.
2. using cmake through meson results in:
    `ERROR: Multiple producers for Ninja target "subprojects/esp32/55ac3a2@@cm_mbedcrypto@sta/bignum.c.o". Please rename your targets.`

**system parameters**
* Native build
* An as of this morning up-to-date Arch Linux
* Python 3.8.2
* Meson 0.54.0
* Ninja 1.10.0

mensinda:
fixes #6928

### DIFF
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -468,8 +468,8 @@ class Backend:
             source = 'meson-generated_' + os.path.relpath(rel_src, targetdir)
         else:
             if os.path.isabs(rel_src):
-                # Not from the source directory; hopefully this doesn't conflict with user's source files.
-                source = os.path.basename(rel_src)
+                # Use the absolute path directly to avoid file name conflicts
+                source = rel_src
             else:
                 source = os.path.relpath(os.path.join(build_dir, rel_src),
                                          os.path.join(self.environment.get_source_dir(), target.get_subdir()))


mensinda:
Both CI failures are unrelated.

PR #6944, opened 2020-04-10
Build private directory name from output file name.

jpakkane:
The output name is guaranteed to be unique already so use that for the priv dir. Originally I though about using `.priv` suffic, but this one is shorter. Collision probability is low, as `,p` is an uncommon thing to have in your output file name.

### DIFF
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -227,7 +227,7 @@ class Backend:
         return self.build_to_src
 
     def get_target_private_dir(self, target):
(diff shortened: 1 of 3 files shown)


jpakkane:
This works now. What do people think?

textshell:
It's always nice to have the intended changes and why they are done spelt out a bit to do drive by reviews.
Also the description talks about ",p" but the code seems to use ".p"?

jpakkane:
> Also the description talks about ",p" but the code seems to use ".p"?

Had to change it because DMD  seems to use comma as a command line separator (somehow) and it broke things.

PR #6952, opened 2020-04-11
opts: Allow `-` and `not` in meson_options.txt (fixes #6948)

### LINKED ISSUE #6948: `meson_options.txt` cannot be used to specify negative integer values
*Actual:*
When trying to specify a negative number in `meson_options.txt`, I get the error `Arguments may only be string, int, bool, or array of those.`.

*Expected:*
No error is raised, and negative values are accepted.

*Rationale:*
These options are ideal for setting defaults for build configuration via `#mesondefine`. It's fairly common for integer options to treat zero and negative values as having special meanings. And even if not, sometimes you'd like to have negative values as options.

*Workaround:*
- Add a second option for the special meaning (ugh!)
- Pass the option as a string, and use `#define` instead of `#mesondefine` (ugh!)

*Details:*
The error is raised here:
https://github.com/mesonbuild/meson/blob/6e865fc08d0d13b3846d00e80c9f3ea090645fb8/mesonbuild/optinterpreter.py#L170

The culprit looks like it may be here: https://github.com/mesonbuild/meson/blob/6e865fc08d0d13b3846d00e80c9f3ea090645fb8/mesonbuild/mparser.py#L107

mensinda:


Ericson2314:
Was it possible to do negative numbers before? This is a little bit of a comparability hazard for any project that relied on only getting natural number user input.

mensinda:
You can already set integer options to negative numbers with `meson configure -Dint_opt=-1`. Just setting a negative default value does not work.

Also, you can restrict the number range to natural numbers with `min: 0`.

xclaesse:
Looks good to me, lgtm.

xclaesse:
Hmm, for some reason I cannot see the failing unit test on my phone, is it related? I'll check later when in pc.

xclaesse:
Same CI failure on master, so let's merge this.

PR #6956, opened 2020-04-11
References-tables.md: repair broken table

Vanadiae:
At https://mesonbuild.com/Reference-tables.html#compiler-and-linker-selection-variables, the table is broken and it seems that the only difference with other well-working tables on this page is the semi-column ":" that are present under Language.


mensinda:
Thanks. Please put `[skip ci]` in the commit message for future documentation only changes.

PR #6957, opened 2020-04-11
azure: Do not skip boost tests

mensinda:


mensinda:
Fixed by manually setting the `BOOST_ROOT` environment variable. It is for some reason no longer set by default.

The other CI errors are unrelated.

nirbheek:
Does not apply to 0.54 branch, do you want to make a new PR for this?

mensinda:
No, this should be fine (just fixes CI stuff). I only added it to the milestone to keep track of it and to ensure that boost is OK before 0.54.1 is released.

PR #6961, opened 2020-04-12
cmake: Preserve include directory order (fixes #6959)

### LINKED ISSUE #6959: Differing ninja target behavior when accessing cmake subproject via sub_proj.target() method versus via __CMake_build dir
**Describe the bug**
I am importing a large cmake projectwith quite a few targets. If I attempt to use Meson to retrieve a cmake target from within the subproject and execute it, I get errors. If I use ninja directly on the __CMake_build dir I can run the same target with no issues.

**To Reproduce**
I've created a repo to showcase an example ESP/meson integrated project but at the moment it's serving as an example of the problem behavior:
https://github.com/Michael-Densberger/meson_esp32

To produce the behavior I ran:
`mkdir /opt/meson_esp32`
`cd /opt/meson_esp32/`
`git clone git@github.com:Michael-Densberger/meson_esp32.git .`
`meson esp32/ build/ --cross-file=esp32/cross.txt`
`ninja -C build/ flash_from_cmake_root` <- Works and flashes no problem.
`ninja -C build/ flash` <- Build errors that vary. 

Note flash and flash_from_cmake_root are Meson `run_targets` that I defined in `meson.build` in the repo.

**Expected behavior**
The flash from the top level which is a `cmake.subproject` target would behave the same as the flash target in the lower level subproject `__CMake_build` folder.

**system parameters**
* Cross build.
* Up-to-date Arch linux.
* Python: 3.8.2
* Meson: 0.54.0
* Ninja: 1.10.0

mensinda:
I think I went a bit overboard with the test case :smile:.

### DIFF
--- a/mesonbuild/cmake/interpreter.py
+++ b/mesonbuild/cmake/interpreter.py
@@ -22,7 +22,7 @@ from .executor import CMakeExecutor
 from .traceparser import CMakeTraceParser, CMakeGeneratorTarget
 from .. import mlog
 from ..environment import Environment
-from ..mesonlib import MachineChoice, version_compare
+from ..mesonlib import MachineChoice, OrderedSet, version_compare
 from ..compilers.compilers import lang_suffixes, header_suffixes, obj_suffixes, lib_suffixes, is_header
 from enum import Enum
 from functools import lru_cache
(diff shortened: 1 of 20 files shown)


mensinda:
CI errors are unrelated.

PR #6966, opened 2020-04-13
Corrected prefix path default in documentation.

snyball:
Using meson version `0.53.2` prefix defaulted to `/usr/local`, not `/usr/local/` (note the trailing slash)


jpakkane:
Thanks, just note in the future that doc-only changes should have `[skip ci]` in their commit message so CI is not run.

PR #6967, opened 2020-04-13
RFC: generator_targets

jpakkane:
This MR has two main things:

- a sample project showing what the UX for defining and using generator_targets could be like
- just enough implementation to make the simplest case work

Eventually you should be able to to everything with generator_targets that you could with a plain generator. In fact the former should be implementable with the latter. It just needs more work to handle things like using the output of one generator target as input to another.

Comments welcome.


mesonbuild/interpreter.py
```diff
@@ -2189,6 +2196,8 @@ def _get_native() -> str:
                                   'depfile',
                                   'capture',
                                   'preserve_path_from'},
+                    'generator_target': {'output',
+                                         'sources'},
```
nirbheek, line 2200: Here we define the permitted kwargs for this function as `output` and `sources`, but then later we also use `command` and `input`? 

From what I understand, the main gripe people have with `custom_target()` is that you have to define it over and over again even if you just want to call the same transformation on different files. So they try to use `generator()` and trip over its deficiencies: that it's called repeatedly if you use the output of `gen_object.process()` in multiple places and that you can't chain it.

So I was imagining that `generator_target()` would similarly only take `command:` + `output:` kwargs, etc (no `input:`) and return an object with, say, a `run()` method that takes a `input:` kwarg (a list), which would create a target with the specified inputs and outputs. Something like:

```meson
gen = generator_target('gen1',
    command: [find_program('script.py'), '@INPUT@', '@OUTPUT@'],
    output: ['@BASENAME@.c'])
# This will generate common1.c common2.c
common_sources = gen.run(input: ['common1.in', 'common2.in'])
# This will generate source1.c source2.c, and reuse common1.c common2.c
executable('some-tool1', gen.run(input: ['source1.in', 'source2.in']), common_sources)
# This will reuse source1.c, generate source3.c, and reuse common1.c common2.c
executable('some-tool2', gen.run(input: ['source1.in', 'source3.in']), common_sources)
```

nirbheek: We will probably want to provide smarter syntax for the `command:` and `output:` templates, since there are a bunch of common cases to be covered. For example, commands that only take one input and one output vs commands that can take multiple inputs and multiple outputs, and commands that can take multiple inputs that are then converted to exactly those many outputs, etc.

jpakkane: Alteratively one could do this:

```meson
g = generator(...)
generator_target('foo', 'input1.txt',
  generator: g)
```

The difference to custom target being that the output goes in the output dir rather than current build dir.

jpakkane:
Updated this so that a generator target takes an external generator object, just like regular generators. This should hopefully convey better that generator targets are just generators that work standalone.

test cases/common/234 generator target/gen1/meson.build
```diff
@@ -0,0 +1,12 @@
...
+  arguments: ['@INPUT@', '@OUTPUT@'],
+)
+
+xml1 = generator_target('gen1',
+  input: ['first.component',
+          'second.component',
+          'third.component',
+         ],
+  generator: g1
+)
```
ppaalanen, line 12: Trying to understand the goal here, how does this make things better than using a `custom_target()` with multiple outputs? Is the point that the tool (`comp2xml.py`) does not need to handle multiple inputs in one invocation?

ppaalanen (commented):
> Updated this so that a generator target takes an external generator object, just like regular generators. This should hopefully convey better that generator targets are just generators that work standalone.

Is that a useful or intuitive concept? I'm not sure. The problem comes from people not understanding that `generator()` is not "standalone", so adding more API to make a "standalone generator" seems backwards to me. "This thing does not do what you expect, so use it with this other thing to get what you expected."

test cases/common/234 generator target/gen1/meson.build
jpakkane: > how does this make things better than using a custom_target() with multiple outputs

A custom target consists of only one external process call. Most generator programs can not handle multiple inputs and outputs and must instead be called once per input file. This is that, basically "like a generator, but output everything to the current build dir".

PR #6968, opened 2020-04-13
Merge wraps from subprojects into wraps from main project

xclaesse:
wraps from subprojects are now merged into the list of wraps from main
    project, so they can be used to download dependencies of dependencies
    instead of having to promote wraps manually. If multiple projects
    provides the same wrap file, the first one to be configured wins.
    
    This also fix usage of sub-subproject that don't have wrap files. We can
    now configure B when its source tree is at
    `subprojects/A/subprojects/B/`. This has the implication that we cannot
    assume that subproject "foo" is at `self.subproject_dir / 'foo'` any
    more.

--
This PR is based on top of https://github.com/mesonbuild/meson/pull/6902, so that one must be merged first.


jpakkane:
All of this is good stuff except for this bit:

> If multiple projects provides the same wrap file, the first one to be configured wins.

This means that if the order in which subprojects are processed changes, the dependency used will change. This is in strict contrast with how the subproject and wrap concept has been designed. The main goal has always been that you can know which dependencies you are using by looking in one, and only one, place: the top level `subprojects` dir. Spreading the deps around the project tree like this becomes unwieldy quite quickly.

Some ways of fixing this would be:

- instead of using the wrap file directly, copy it to the top level `subprojects` dir and use it from there
- write a "forwarding wrap file" in the top level dir that points to the actual wrap file to use (a sort of a conceptual symlink if you will)

xclaesse:
Note that the code is still downloaded into the main project' subproject dir, so it's clear which code it's blusing, unless it's a subprojects that has no wrap file in which case we have no choice but using it in place. But some kind of symlink wrap can't hurt I agree.

jpakkane:
If the autoupdate is only used when the user passes a command line argument and we always extract the files in the top level `subprojects` dir then I guess it could work. It should extract the files in a subdirectory that matches the dependency name, though so it will be picked up on subsequent runs. (i.e. it shoudl extract to `glib` and not `glib-x.y.z`. But then it fails if it provides multiple deps with different names. 🤔

xclaesse:
@jpakkane what about simply copying the wrap file used into the main project, so the user can easily review and commit it in case it matters? That will also ensure that stuff like 'meson subprojects update' will work.

Providing multiple dependencies in the same wrap file is fine, with the `[provide]` section added in https://github.com/mesonbuild/meson/pull/6902. Btw that PR is a prerequisite.

xclaesse:
@jpakkane I added a commit to copy wrap file into main project. Keep in mind that this PR is based on https://github.com/mesonbuild/meson/pull/6902, most commits come from there.

jpakkane:
There seems to be a requirement for both use cases, autopromoting for convenience and not autopromoting so all dependencies can be inspected by hand. So we probably need some sort of a toggle. Which one should be the default? Wrap download does the former at least.

Alternatively we could add a new command like `meson wrap autopromote` that does this task so it is convenient but is only done when the user invokes it manually.

xclaesse:
@jpakkane what's the use-case for not wanting auto promote? Note that this PR does copy the .wrap file it used to main project so it can be inspected more easily.

xclaesse:
I added `--wrap-mode=nopromote` to disable this new behavior.

xclaesse:
@nirbheek @jpakkane any objection on this now?

jpakkane:
I got some random thoughts. They are not review as such, more of a bunch of things that came to mind.

- If the used version of a subproject changes, it should still work assuming the project has set up their dependency versions properly (i.e. it fails if the other version is too old)
- people might want to have a way to "freeze" promoted subprojects for the duration of each build dir, this could be done with the forwarding wrap files as discussed above
- we may need to change the command line arguments if it turns out that someone needs to have downloading enabled but promoting disabled

But, again, conceptually this is ok.

xclaesse:
>     * If the used version of a subproject changes, it should still work assuming the project has set up their dependency versions properly (i.e. it fails if the other version is too old)

Not sure to understand what you mean here. 

>     * people might want to have a way to "freeze" promoted subprojects for the duration of each build dir, this could be done with the forwarding wrap files as discussed above

This PR copies the .wrap into the parent project, that has the effect of freezing it until the user delete the wrap file, and one the subproject has been downloaded it won't be updated until the user calls `meson subprojects update`. So I think it's pretty well frozen already.

>     * we may need to change the command line arguments if it turns out that someone needs to have downloading enabled but promoting disabled

`--wrap-mode=nopromote` still leaves download enabled. But it is true that there is no way to combine wrap modes, I don't see a use-case for that atm, but if some complains we could change wrap_mode to an array option, for exampe: `--wrap-mode=nopromote,forcefallback`.

phillipjohnston:
As described in #7788, the existing wrap file promotion model is insufficient since 0.55 introduced the ability to reference local source archives and patch archives/directories. I definitely do not want to manually copy/promote source archives and patch files across every repository that uses a subproject; that's why I wanted to localize it in a reusable subproject :).

I downloaded @xclaesse's PR and tested it out locally, and it does exactly what I am expecting. I hope it gets merged soon, since there doesn't actually seem to be any real gates.

xclaesse:
Actually there is an issue: Since this PR copies the .wrap file into parent project, on reconfigure it will use the .wrap from parent project and expect packagefiles to be in parent project too. I agree with @phillipjohnston that we should not in general have to copy everything into the parent project, so maybe by default the .wrap file should be be copied neither.

Another thing I noticed with this PR is it actually gets hard to know from where the wrap came. When using this with gst-build I noticed that the dependency graph can gets pretty deep: gst-build > gst-plugins-base > pang > cairo > freetype2 > fontconfig.

What about not copying .wrap files, but instead add the info of which .wrap has been used in the Subprojects summary? Not sure exactly how best that can be presented, but here is an abstract of how it looks like currently with gst-build:
```
  Subprojects
               gst-plugins-bad: YES 7 warnings
              gst-plugins-base: YES 3 warnings
              gst-plugins-good: YES
                gst-plugins-rs: NO Feature 'rs' disabled
```

We could append on each line `(needed by foo)` where `foo` is the subproject that configured the sub-subproject?

jpakkane:
It seems to me that my suggestion from above would work in this case as well:

> write a "forwarding wrap file" in the top level dir that points to the actual wrap file to use (a sort of a conceptual symlink if you will)

In this way the info where the dep comes from is always in the top level subprojects dir. If it is only printed in the console, it risks getting lost in the scrollback.

phillipjohnston:
@jpakkane, just making sure I understand: you preference for a "forwarding wrap file" vs picking up wraps in the sub-subproject is because you are forced to explicitly indicate which particular wrap you want in the event of a duplicate?

xclaesse:
> It seems to me that my suggestion from above would work in this case as well:
> 
> > write a "forwarding wrap file" in the top level dir that points to the actual wrap file to use (a sort of a conceptual symlink if you will)

You're right, that would fix the problem indeed. I still think it would be useful information to know which project pulls each subproject, but that should be done in separate PR then. I'll update this PR.

xclaesse:
@jpakkane added a commit that adds a new type of wrap file:
```
[wrap-redirect]
filename = path/to/another/foo.wrap
```

The parser will recursively parse the linked wrap file instead of this one. The `packagefiles` directory is set to the location where the final wrap file is found.

I'm not sure that new type of wrap file should be documented, maybe it should be kept as internal implementation detail? I'm wondering if this mechanism could be abused in some way? What do you think?

phillipjohnston:
@xclaesse @jpakkane really like the new approach after trying it out on my projects.

mesonbuild/wrap/wrap.py
```diff
@@ -97,25 +98,34 @@ def __init__(self, fname: str):
...
+            self.config.read(self.filename)
         except configparser.Error:
             raise WrapException('Failed to parse {}'.format(self.basename))
         self.parse_wrap_section()
+        if self.type == 'redirect':
+            # [wrap-redirect] have a `filename` value pointing to the real wrap
+            # file we should parse instead. It can be either absolute path or
+            # relative to the current wrap file location.
+            dirname = os.path.dirname(self.filename)
+            self.filename = os.path.join(dirname, self.values['filename'])
```
jpakkane, line 126: This should validate that the value points to an actual subproject and does not try to do anything nasty like having `..` in it. Because if we don't check, people will abuse the crap out of it. The path should thus be of the type `subprojects/foo/subprojects/bar` or it should have only the project names like `foo/bar` or, if we had the syntax from #7513,  `['foo', 'bar']`.

There should also be a test for the forwarding wrap file by itself where the dependency info is static and in Git.

xclaesse: Added unit tests and a few sanity checks:  filename must be within current subprojects/ dir, must point to a .wrap file that exists.

I'm against the form `foo/bar` because it looks like a path but it's not, that's too confusing. Took me a while to understand your suggestion. `['foo', 'bar']` is a bit better but not currently possible. `subprojects/foo/subprojects/bar` it should be in that form but it's actually hard to enforce because each subproject could have subprojects_dir set to a different value. Those redirect wrap files are parsed at the very beginning of the main project interpreter, at that point it has no idea of any subprojects.

xclaesse:
CI failures seems unrelated.

mesonbuild/wrap/wrap.py
jpakkane: > it should be in that form but it's actually hard to enforce because each subproject could have subprojects_dir set to a different value

Let me tell you a little secret. There is at least one other place in the code that does this kind of traversal and it completely ignores the fact that `subprojects_dir` can be changed. Yet, no-one has ever complained or filed a bug. The only people who use that seem to be existing corporate monorepo thingies that will never be used as subprojects themselves. In fact, we could even deprecate that setting, it is only meant as a transitory helper.

Summa summarum: the actual syntax is not that important, but it really must validate that nested subprojects come from the respective subprojects dir rather than some random dir elsewhere.

xclaesse: Ok then... updated the PR to enforce it.

jpakkane:
But it :crossed_swords: conflicts now. :cry:

xclaesse:
@jpakkane rebased again now.

phillipjohnston:
Thank you @xclaesse and @jpakkane, I'm grateful for this change.

jpakkane:
It's pretty much all xclaesse. I'm mostly a grumpy merge blocker. :P

Issue #6969, opened 2020-04-13
Bad command name on run_command error message 


adalessandro:
Calling `run_command` with a non-found program returned by `find_program`, returns an ugly error message.

`meson.build` file:
```
project('test', 'c')

foo_program = find_program('foo_program', required : false)
run_command(foo_program)
```
Log:
```
$ python --version
Python 3.8.2

$ meson build
The Meson build system
Version: 0.54.999
[...]
Build type: native build
Project name: test
Project version: undefined
C compiler for the host machine: ccache cc (gcc 9.3.0 "cc (Arch Linux 9.3.0-1) 9.3.0")
C linker for the host machine: cc ld.bfd 2.34
Host machine cpu family: x86_64
Host machine cpu: x86_64
Program foo_program found: NO

meson.build:4:0: ERROR: command <NonExistingExternalProgram 'nonexistingprogram' -> [None]> not found or not executable
[...]
```

System details are described in the code snippet above.

adalessandro:
After chatting about this issue in the PR [0], I think that the proper solution would be storing the program's name in the `NonExistingExternalProgram` instances, and showing it in the current error message.

I'll submit a patch for this soon.

[0] https://github.com/mesonbuild/meson/pull/6970

adalessandro:
This patch prints the program's name on the error message:
https://github.com/adalessandro/meson/commit/bcd8b0473ebf50920a5759d4a2b24db3c106e084

Same test setup as before:

```
$ meson build
The Meson build system
Version: 0.54.999
[...]
Build type: native build
Project name: test
Project version: undefined
C compiler for the host machine: ccache cc (gcc 9.3.0 "cc (Arch Linux 9.3.0-1) 9.3.0")
C linker for the host machine: cc ld.bfd 2.34
Host machine cpu family: x86_64
Host machine cpu: x86_64
Program foo_program found: NO

meson.build:3:0: ERROR: command 'nonexistingprogram' not found or not executable
[...]
```

adalessandro:
This patch modifies `find_program()` to store the program's name even if it's not found:
https://github.com/adalessandro/meson/commit/4f7d62bcee76e2b78ba2b753f5742a27dcdbae6c

This way, the `NonExistingExternalProgram` will have the target program's name stored instead of the default name `nonexistingprogram`, as printed in the previous comment.

Example test with a single program name argument:

```
$ cat meson.build 
project('test', 'c')
foo_program = find_program('foo_program', required : false)
[...]

PR #6970, opened 2020-04-13
interpreter: Fix bad command name on run_command error message

adalessandro:
Calling run_command with a non-found program returned by find_program, returns an ugly error message. Fix this by improving the returned message for ExternalProgram.

See issue https://github.com/mesonbuild/meson/issues/6969

### DIFF
--- a/mesonbuild/interpreter.py
+++ b/mesonbuild/interpreter.py
@@ -2581,6 +2581,8 @@ external dependencies (including libraries) must go to "dependencies".''')
                       ' and therefore cannot be used during configuration'
                 raise InterpreterException(msg.format(progname, cmd.description()))
             if not cmd.found():
+                if isinstance(cmd, dependencies.ExternalProgram):
+                    raise InterpreterException('Tried to use not-found external program')
                 raise InterpreterException('command {!r} not found or not executable'.format(cmd))
         elif isinstance(cmd, CompilerHolder):
             exelist = cmd.compiler.get_exelist()


mesonbuild/interpreter.py
```diff
@@ -2581,6 +2581,8 @@ def run_command_impl(self, node, args, kwargs, in_builddir=False):
                       ' and therefore cannot be used during configuration'
                 raise InterpreterException(msg.format(progname, cmd.description()))
             if not cmd.found():
+                if isinstance(cmd, dependencies.ExternalProgram):
+                    raise InterpreterException('Tried to use not-found external program')
```
xclaesse, line 2585: Nitpicking: I would the external program's name in the message: cmd.get_name().

adalessandro: Right, I'd rather return the external program's name. But the `cmd.get_name()` will return `nonexistingprogram` as it's an instance of `NonExistingExternalProgram`.
This commit fixes the bad error message. Further patches could deal with the implementation itself to enable showing the non-found program's name in the message.

adalessandro: More specifically, the `NonExistingExternalProgram` instance is being created in the `Interpreter::find_program_impl` method with an empty name argument, which defaults to `nonexistingprogram`. This could be changed in order to store the name of the target program, even if it hasn't been found. I wonder if there's any reason not to do so.

xclaesse: That can be improved by returning `NonExistingExternalProgram(args[0])` in `func_find_program()` and `find_program_impl()`

xclaesse: Or even `', '.join(args)` instead of `args[0]`

adalessandro: Exactly, I agree. I'll submit another PR setting the proper name. I'd rather use the joined args array, as that is more correct than just the first program name.

adalessandro: After chatting about this, I'm closing this PR as I think the proper solution would be storing the program's name in the `NonExistingExternalProgram` instances, and showing it in the current error message. @xclaesse thanks!

PR #6971, opened 2020-04-13
Freedesktop module

dcbaker:
This is a work in process to add a module for creating freedesktop data files. There's a bunch of them, people need them, and providing a built-in method for generating and installing them to the correct place that is tested and validated seems pretty dang useful.

This is not completely as-is, the .desktop file is pretty complete I think, but the appstream generation is super limited still. The tests work on my system, but probably fail somewhere (windows?). There's also limited documentation. I've added automatic validation generation to the targets as well, which I think is pretty slick.

I'm mostly hoping to get some feedback before I spend too much more time on this.

### DIFF
--- a/docs/markdown/Freedesktop-module.md
+++ b/docs/markdown/Freedesktop-module.md
@@ -1,0 +1,112 @@
+# Freedesktop Module
+
+This module provides helper functions for freedesktop standards used primarily on *nix systems.
+
+*New in 0.55.0*
+
+## File generators
+
+### desktop_file()
+
+```meson
+  custom_target desktop_file(exe, ...)
+```
+
+Helper for generating a [.desktop file](https://specifications.freedesktop.org/desktop-entry-spec/latest/index.html)
+
+Creates a custom_target that will be installed if the executable used to
+generate it is also installed. Takes one positional argument which is an
+executable or a disabler. If a disabler is passed or the executable is not
+going to be installed then nothing is returned.
+
+The `Exec` field is automatically populated with the installed path to the executable.
+
+The `Terminal` field is set based on the value of the Executable's `gui_app` field.
+
+This function has the following keyword arguments:
+
+- `name` The name of the program. This is the value of the `Name` field in
+  the .desktop file. If this field is not set then the name of the executable
+  is used.
+
+- `comment` The value of the `Comment` field in the .desktop file. If this
+  value is not provided then the field is unset.
+
+- `generic_name` is the `GenericName` in the .desktop file. If this value is
+  not provided then the field is unset.
+
+- `icon` is the `Icon` value in the .desktop file. Meson does not validate
+  this value in anyway. If not provided the field is unset.
+
+- `categories` a meson list of categories for this program. These values will
+  be joined into a single semicolon delimited string in the .desktop file.
+  Meson does not validate this value in any way. If not provided the field is
+  unset.
+
+- `install_name` The name of the .desktop file without the file extension. If
+  this is unset the name of the value of `name` or the name of the executable
+  will be used, in that order.
+
+- `mimetype` A meson list of mimetypes that this program handles. These
+  values will be joined into a single semicolon delimited string in the
+  .desktop file. Meson does not validate these values in any way. If not
+  provided the field is unset.
+
+- `dbus` A boolean that populates the `DBusActivatable` field in the .desktop
+file. Defaults to false
+
+- `no_display` A boolean for the `NoDisplay` field.
+
+- `only_show_in` A list of strings to populate the `OnlyShowIn` field.
+
+- `not_show_in` A list of strings to populate the `NotShowIn` field.
+
+- `startup_notify` A Boolean for to populate the `StartupNotify` field.
+
+- `startup_wm_class` A Boolean for to populate the `StartupWMClass` field.
+
+- `implements` A list of strings that will be `;` joined to populate the
+  `Implements` field.
+
+- `keywords` A list of strings that will be `;` joined to populate the
+  `Keywords` field.
+
+- `extra_args` A dict of string to string|list of string|boolean's for
+  extended attributes. The keys must start with `X-`. Lists will be converted
+  in `;` delimited strings. If a value needs to be formatted differently the
+  caller must format it themselves and pass the value as a string. Meson
+  validates only that all keys start with `X-`, it is up to the caller to
+  ensure that these keys are correctly formatted.
+
+- `test_target` A boolean. If this is set to true then a test target to
+  validate the .desktop file is also generated. If this is true then the
+  program `desktop-file-validate` is required
+
+- `actions` A list of dictionaries for additional actions for the .desktop
+  file. If this value is not provided then no additional actions are set.
+
+The actions dictionaries have the following signature:
+
+- `exec` Additional arguments to pass to the Executable. The path to the
+  executable is provided, these are treated as extra arguments. Required.
+
+- `name` the name of the action. Required.
+
+- `icon` Icon for the action, uses the same logic as the main `icon`
+  argument. Optional.
+
+```meson
+fdo = import('freedesktop')
+
+exe = executable(..., gui_app : true, install : true)
+fdo.desktop_file(
+  exe,
+  generic_name : 'IDE',
+  actions : [
+    {
+      name : 'Import',
+      exec : ['--import', '%f'],
+    }
+  ]
+)
+```
--- a/docs/markdown/snippets/freedesktop_module.md
+++ b/docs/markdown/snippets/freedesktop_module.md
@@ -1,0 +1,15 @@
+## A new freedesktop module for common freedesktop operations
+
+This module include useful functionality for projects that use freedesktop standards, such as .desktop files.
+
+```meson
+exe = executable(..., gui_app : true, install : true)
+
+fdo = import('freedesktop')
+exe_desktop = fdo.desktop_file(
+    exe,
+    name : 'UnicornAwesomeApp',
+    generic_name : 'Email',
+    ...
+)
+```
--- a/docs/sitemap.txt
+++ b/docs/sitemap.txt
@@ -49,6 +49,7 @@ index.md
 			Windows-module.md
 			Cuda-module.md
 			Kconfig-module.md
+			Freedesktop-module.md
 		Java.md
 		Vala.md
 		D.md
--- a/mesonbuild/interpreter.py
+++ b/mesonbuild/interpreter.py
@@ -4450,12 +4450,12 @@ Try setting b_lundef to false instead.'''.format(self.coredata.base_options['b_s
             results.append(s)
         return results
 
-    def add_target(self, name, tobj):
+    def add_target(self, name: str, tobj: build.BuildTarget) -> None:
         if name == '':
             raise InterpreterException('Target name must not be empty.')
         if name.strip() == '':
             raise InterpreterException('Target name must not consist only of whitespace.')
-        if name.startswith('meson-'):
+        if not tobj.internal_target and name.startswith('meson-'):
             raise InvalidArguments("Target names starting with 'meson-' are reserved "
                                    "for Meson's internal use. Please rename.")
         if name in coredata.forbidden_target_names:
--- a/mesonbuild/interpreter.py
+++ b/mesonbuild/interpreter.py
@@ -2352,7 +2352,7 @@ class Interpreter(InterpreterBase):
 
         if isinstance(item, build.CustomTarget):
             return CustomTargetHolder(item, self)
-        elif isinstance(item, (int, str, bool, Disabler)) or item is None:
+        elif isinstance(item, (int, str, bool, Disabler, mesonlib.File)) or item is None:
             return item
         elif isinstance(item, build.Executable):
             return ExecutableHolder(item, self)
@@ -2397,7 +2397,7 @@ class Interpreter(InterpreterBase):
                 self.process_new_values(v.sources[0])
             elif hasattr(v, 'held_object'):
                 pass
-            elif isinstance(v, (int, str, bool, Disabler)):
+            elif isinstance(v, (int, str, bool, Disabler, mesonlib.File)):
                 pass
             else:
                 raise InterpreterException('Module returned a value of unknown type.')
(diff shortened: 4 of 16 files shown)


dcbaker:
@ximion, I started thinking about this again I saw your metainfo creator. I thought you might be interested.

ximion:
Looks nice! I'll probably be able to take a much closer look at the code later (just quickly skimmed through it).

> There's also limited documentation.

With that, I can most certainly help, especially in case of AppStream and most likely in case of other XDG specs. What information are you missing?

dcbaker:
> With that, I can most certainly help, especially in case of AppStream and most likely in case of other XDG specs. What information are you missing?

Oh, I meant *I* need to write meson documentation for the new code :)

ximion:
> Oh, I meant I need to write meson documentation for the new code :)

Haha, yeah, that makes sense - apparently I am so used to people complaining about XDG docs that I now assume that implicitly, even if it isn't the case :P

docs/markdown/Freedesktop-module.md
```diff
@@ -0,0 +1,47 @@
...
+
+## File generators
+
+### desktop_file()
+
+```meson
+  datafile desktop_file(exe, ...)
+```
+
+Creates a an datafile that will be installed if the executable used to
```
xclaesse, line 15: typo: "a an"

docs/markdown/Freedesktop-module.md
```diff
@@ -0,0 +1,47 @@
...
+
+- `comment` The value of the `Comment` field in the .desktop file. If this value is not provided then the field is unset.
+
+- `generic_name` is the `GenericName` in the .desktop file. If this value is not provided then the field is unset.
+
+- `icon` is the `Icon` value in the .desktop file. Meson does not validate this value in anyway. If not provided the field is unset.
+
+- `categories` a meson list of categories for this program. These values will be joined into a single semicolon delimited string in the .desktop file. Meson does not validate this value in any way. If not provided the field is unset.
+
+- `install_name` The name of the .desktop file without the file extension. If this is unset the name of the value of `name` or the name of the executable will be used, in that order.
```
xclaesse, line 35: "the name of the value of `name`" sounds weird. Do you mean "the value of the `name` field or the name of the executable..." ?

docs/markdown/snippets/freedesktop_module.md
```diff
@@ -0,0 +1,15 @@
...
+```meson
+exe = executable(..., gui_app : true, install : true)
+
+fdo = import('freedesktop')
+exe_desktop = fdo.desktop_file(
+    exe,
+    name : 'UnicornAwesomeApp',
+    generic_name : 'Email',
+    ...
+)
```
xclaesse, line 14: Would be nice to have the example in the module documentation too.

docs/sitemap.txt
```diff
@@ -49,6 +49,7 @@ index.md
 			Windows-module.md
 			Cuda-module.md
 			Kconfig-module.md
+			Freedesktop-module.md
```
xclaesse, line 52: I think it needs to be added in `docs/theme/extra/templates/navbar_links.html` too.

mesonbuild/modules/freedesktop.py
```diff
@@ -0,0 +1,206 @@
...
+    e = et.SubElement(parent, tag, **kwargs)
+    e.text = text
+    return e
+
+
+class FreedesktopModule(ExtensionModule):
+
+    @permittedKwargs({'name', 'generic_name', 'icon', 'categories',
+                      'actions', 'install_name', 'mimetype', 'comment',
+                      'dbus'})
```
xclaesse, line 70: dbus kwarg is missing from doc.

mesonbuild/modules/freedesktop.py
```diff
@@ -0,0 +1,206 @@
...
+    def desktop_file(self, state: 'ModuleState',
+                     args: T.Tuple[T.Union['ExecutableHolder', 'Disabler']],
+                     kwargs: T.Dict[str, T.Any]) -> ModuleReturnValue:
+        if len(args) != 1:
+            raise MesonException('Only one positinal argument allowed.')
+
+        if isinstance(args[0], Disabler):
+            return ModuleReturnValue(None, [])
```
xclaesse, line 78: If an argument is a disabler the interpreter should not even call your method, are you sure this can happen?

mesonbuild/modules/freedesktop.py
```diff
@@ -0,0 +1,206 @@
...
+                raise MesonException('"{}" must be a string if defined'.format(kkey))
+            values[vkey] = value
+
+
+def optional_list_arg(kwargs: T.Dict[str, T.Any], kkey: str,
+                      values: T.Dict[str, str], vkey: str) -> None:
+        value = listify(kwargs.get(kkey, []))
+        if value:
+            if any(not isinstance(i, str) for i in value):
+                raise MesonException('"{}" must be a list of strings if defined'.format(kkey))
```
xclaesse, line 56: stringlistify() does that check for you IIRC.

mesonbuild/modules/freedesktop.py
```diff
@@ -0,0 +1,206 @@
...
+            raise MesonException('Only one positinal argument allowed.')
+
+        if isinstance(args[0], Disabler):
+            return ModuleReturnValue(None, [])
+
+        exe = args[0].held_object  # type: Executable
+        if not isinstance(exe, Executable):
+            raise MesonException('First argument must be an Executable')
+
+        name = kwargs.get('name', '')  # type: str
```
xclaesse, line 84: You don't validate that name is a string.

mesonbuild/modules/freedesktop.py
```diff
@@ -0,0 +1,206 @@
...
+        if not isinstance(dbus, bool):
+            raise MesonException('dbus must be a boolean.')
+        values['DBusActivatable'] = str(dbus).lower()
+
+        actions = listify(kwargs.get('actions', []))
+        if actions:
+            for a in actions:
+                if not isinstance(a, dict):
+                    raise MesonException('"actions" must be a list of dicts if defined.')
+            values['Actions'] = ';'.join(a['action'] for a in actions)
```
xclaesse, line 118: You should validate that 'action' is in the dict and is a string.

mesonbuild/modules/freedesktop.py
```diff
@@ -0,0 +1,206 @@
...
+
+        # This must happen after values['Action'], but before any action
+        # sections are added.
+        config['Desktop Entry'] = values
+
+        for action in actions:
+            if 'action' not in action:
+                raise MesonException('missing required "action" entry in action "{}"'.format(action))
```
xclaesse, line 126: You validate after having using it.

mesonbuild/modules/freedesktop.py
```diff
@@ -0,0 +1,206 @@
...
+            if 'name' not in action:
+                raise MesonException('missing required "name" entry in action "{}"'.format(action))
```
xclaesse, line 128: Need to validate it's a string too.

mesonbuild/modules/freedesktop.py
```diff
@@ -0,0 +1,206 @@
...
+            avalues = {}
+            avalues['Name'] = action['name']
+
+            exec_ = listify(action['exec'])
```
xclaesse, line 132: You don't validate that 'exec' is in the dict, and you could use stringlistify() to validate they are strings.

mesonbuild/modules/freedesktop.py
```diff
@@ -0,0 +1,206 @@
...
+            if any(not isinstance(a, str) for a in exec_):
+                raise MesonException('"exec" must be a list of strings if defined.')
+            avalues['Exec'] =  ' '.join([values['Exec'], *exec_])
+
+            optional_string_arg(action, 'icon', avalues, 'Icon')
+
+            config['Desktop Action {}'.format(action['action'])] = avalues
+
+        install_name = kwargs.get('install_name', name)
```
xclaesse, line 141: You don't validate it's a string.

mesonbuild/interpreter.py
```diff
@@ -703,9 +703,12 @@ def get_custom_install_mode(self):
...
 
 class DataHolder(InterpreterObject, ObjectHolder):
-    def __init__(self, data):
+    def __init__(self, data, subproject: str):
         InterpreterObject.__init__(self)
-        ObjectHolder.__init__(self, data)
+        ObjectHolder.__init__(self, data, subproject)
+        self.methods.update({
+            'sources': self.sources_method,
+        })
```
xclaesse, line 711: Hmm, Data object is not documented. The only place it's currently used is the return value of install_data() that is documented as being void... I'm not against adding this method, but we need to add a section under "returned objects" of reference manual.

xclaesse:
@dcbaker It's all just small nitpickings, overall it looks good to me. I haven't looked at appstream part since it's not documented, and I don't know what they are.

One general question I've got: are there any escaping we should do? Does it work when the install path of the exe has spaces for example?

xclaesse:
Also, desktop files can have many more fields, even unrestricted `X-` fields. e.g. from random file on my distro, I see `X-Ubuntu-Gettext-Domain`, `X-Flatpak-RenamedFrom`, `StartupNotify`. Should we provide a freeform dictionary  kwarg for them? Like `extra_entries: {'StartupNotify': 'true', ...}`.

Oh, and I see you hardcode `values['Type'] = 'Application'`, are there other types of desktop files we could want to support? Mostly out of curiosity, they can always be added later.

xclaesse:
> One general question I've got: are there any escaping we should do? Does it work when the install path of the exe has spaces for example?

Self replying: Yes there are quoting/escaping considerations we should take care of: https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s07.html.

btw, linking to the spec, at least in the implementation code, could be nice.

dcbaker:
> Oh, and I see you hardcode values['Type'] = 'Application', are there other types of desktop files we could want to support? Mostly out of curiosity, they can always be added later.

There are two other types currently, `link` and `directory`. I don't think that either make sense for meson, but maybe someone else has a different opinion.

mesonbuild/interpreter.py
dcbaker: The more I think about it, the more I think this should be a custom_target. We want it to be updated whenever any of the inputs are updated anyway.

dcbaker:
> Also, desktop files can have many more fields, even unrestricted X- fields. e.g. from random file on my distro, I see X-Ubuntu-Gettext-Domain, X-Flatpak-RenamedFrom, StartupNotify. Should we provide a freeform dictionary kwarg for them? Like extra_entries: {'StartupNotify': 'true', ...}.

I think we should support all of the official keys as builtin functionality, but `X-` keys, sure, lets just put those in a a free form dictionary. We can even enforce that they're `X-` fields.

dcbaker:
Updated to return a CustomTarget, also added a keyword to automatically generate tests. I think the .desktop stuff is pretty much complete baring any bugs in my code (There is a bug I found in the desktop validator and sent an MR for). The appstream function still needs work I think.

docs/markdown/Freedesktop-module.md
```diff
@@ -0,0 +1,112 @@
...
+```meson
+fdo = import('freedesktop')
+
+exe = executable(..., gui_app : true, install : true)
+fdo.desktop_file(
+  exe,
+  generic_name : 'IDE',
+  actions : [
+    {
+      name : 'Import',
```
xclaesse, line 107: keys must be quoted strings here.

xclaesse: Unless we introduce intermediary object like `action = fd.action('Import', exec : [...])`

xclaesse:
@dcbaker could you explain why you opted for a custom target instead of generating the file at configure time? Note that some modules uses helper script run at build time and they put them into mesonbuild/scripts/ so they can be invoked as `meson --internal scriptname args` instead of mangling it in the module py file. Also you could use pickle to serialise your data instead of json, probably faster and safer regarding escaping issues you could have.

dcbaker:
Because I want to be able to pass the location of the .desktop file to the appstream() method (it needs to point at a .desktop file). Maybe it could be a run_target instead of a custom_target?

xclaesse:
@dcbaker I mean you could generate the .desktop file right away during configure, instead of delaying to build time, no? It's pretty much a configure_file() kind of operation as far as I can tell.

dcbaker:
I said run_target, but I mean run_command. I just need a value to pass from the .desktop generator to something else.

xclaesse:
Why not just writing the file directly in FreedesktopModule.desktop_file() and return `mesonlib.File.from_built_file(self.subdir, output)` just like configure_file() does?

eli-schwartz:
As a matter of curiosity, why would one wish to use this instead of using some sort of generator program to create a file and check it into git? For a fully static file I wouldn't want to waste build times (even very minor quantities of it) on having meson convert it from one format to another.

And at least I don't really understand the benefit of generating a .desktop file with a trivial file format using keyword arguments to a meson function that just become lines of the form `kwargname=kwargval1;kwargval2;kwargval3;kwargval4;`.

xclaesse:
Not a blocker, can be added later, but a post_install() method would make sense, similar to https://github.com/mesonbuild/meson/pull/8272. As suggested here: https://github.com/mesonbuild/meson/issues/8268#issuecomment-771782087.

xclaesse:
Right, that issue has been discussed above already, https://github.com/mesonbuild/meson/pull/6971#issuecomment-613713458. I think it make sense to have kwarg for the most common keys, but add a way to add free form str:str dictionary to extend it, and probably not limit that to `X-`.

ximion:
A dictionary won't cut it though - many of the tags in MetaInfo files have attributes (such as versions, comparison attributes, types, ...) and even more are actually deeply nested. A dict is flat - unless you want to nest dictionaries, but that complexity will probably get out of hand rather quickly, and will probably be harder to read than just writing the data outright.

eli-schwartz:
I will reiterate and expound on my belief that meson is supposed to configure files based on *build configuration options*, compile source code, install data, orchestrate tool processing, etc. -- it is not supposed to natively implement some file format and generate data files by transpiling meson dictionaries into XML or ini files. And as a side effect, completely waste time during `meson configure` by creating files that are a textbook case of "something that should be checked into git".

If the highly simple Freedesktop .desktop file format needs special tools to create them without syntax errors, meson or Freedesktop should provide a FAQ encouraging people to use some independent tool that helps you do that and produces files suitable for, again, checking into git.

A meson Freedesktop module should limit itself to:
- providing post-install integration for update-desktop-database / update-mime-database
- providing `install_foo()` methods that know *where* to install a source file better than the generic `install_data()` does
- optionally, `install_foo()` can be really clever and add a `meson test` rule to run `desktop-file-validate` or `appstreamcli validate` etc.

Please, please, please, do NOT get meson into the business of becoming a source code authoring system.

PR #6977, opened 2020-04-14
i18n: generate .mo files with folder structure

### LINKED ISSUE #6973: i18n: generate .mo files on build
I'm currently looking into i18n support for several Python applications. They work fine on installation, but it's not possible to test them without installation:
Right now the `.mo` files are only generated when running `ninja install`. If you want to test a translation, this is a bit annoying. Better would be, if `ninja` would generate the `.mo` files in a folder style, which actually represents how they are going to be installed. Example:
```
builddir/
    po/
        locale/
            lang/
                LC_MESSAGES/
                    domain.mo
po/
    lang.po
```
This way Python projects could [set the localedir](https://docs.python.org/3/library/gettext.html#gettext.bindtextdomain) when running without an installation for testing purposes. In the `meson.build`:
```meson
localedir_devel = join_paths(meson.build_root(), 'po', 'locale')
```

I know there is the `domain-gmo` call for ninja, however this doesn't fix the problem. Since the `.mo` files need to be generated anyway, I think it's better to do it during the build and just copy it on installation than building them on installation.

stephanlachnit:
The goal of this PR is to add a new target called `pkgname-generate-mo`, which generates the `.mo` files in such a folder structure that it can be used with [Python's gettext.bindtextdomain() function](https://docs.python.org/3/library/gettext.html#gettext.bindtextdomain). This allows for translations to be tested without workarounds or an installation.

This currently is a draft because I have some question how you handle certain things:
- The `pkgname-gmo` call is now kinda useless, it should be eventually removed. Do you want to keep it for compatibility reasons? How should I proceed here?
- The Documentation isn't updated yet (depends on the point above).

Closes #6973.


mesonbuild/scripts/gettext.py
```diff
@@ -66,6 +66,16 @@ def gen_gmo(src_sub, bld_sub, langs):
                                '-o', os.path.join(bld_sub, l + '.gmo')])
     return 0
 
+def generate_mo(src_sub, bld_sub, pkgname, langs):
+    for l in langs:
+        srcfile = os.path.join(src_sub, l + '.po')
+        outfile = os.path.join(bld_sub, l, 'LC_MESSAGES',
+                               pkgname + '.mo')
+        os.makedirs(os.path.dirname(outfile), exist_ok=True)
```
stephanlachnit, line 74: Is this safe to assume? Can we create subdirs in the builddir this way without worrying? I guess in the real world this _should_ never happen in this case, but just to be sure^^

PR #6980, opened 2020-04-15
Fix CMake Windows system library deps

fooishbar:
These three commits fix linking of system libraries when declared as dependencies of CMake targets we pick up as Meson dependencies. In particular, when linking Clang into Mesa via CMake, we were dropping all of the system library dependencies such as `ole32` and `uuid` … except if Meson was executed from the source directory, in which case we would retain `version` as a bare argument (since the Mesa source root has a file called `VERSION`), which would get passed through to the linker without a suffix qualification and fail the build.

This PR attempts to do something more reasonable with system-library dependencies under Windows.

cc @dcbaker @jenatali

### DIFF
--- a/mesonbuild/dependencies/base.py
+++ b/mesonbuild/dependencies/base.py
@@ -1413,6 +1413,7 @@ class CMakeDependency(ExternalDependency):
         # Set dependencies with CMake targets
         # recognise arguments we should pass directly to the linker
         reg_is_lib = re.compile(r'^(-l[a-zA-Z0-9_]+|-pthread|-delayload:[a-zA-Z0-9_\.]+|[a-zA-Z0-9_]+\.lib)$')
+        reg_is_maybe_bare_lib = re.compile(r'^[a-zA-Z0-9_]+$')
         processed_targets = []
         incDirs = []
         compileDefinitions = []
@@ -1482,6 +1483,14 @@ class CMakeDependency(ExternalDependency):
                         targets += [j]
                     elif reg_is_lib.match(j) or os.path.exists(j):
                         libraries += [j]
+                    elif mesonlib.is_windows() and reg_is_maybe_bare_lib.match(j):
+                        # On Windows, CMake library dependencies can be passed as bare library names,
+                        # e.g. 'version' should translate into 'version.lib'. CMake brute-forces a
+                        # combination of prefix/suffix combinations to find the right library, however
+                        # as we do not have a compiler environment available to us, we cannot do the
+                        # same, but must assume any bare argument passed which is not also a CMake
+                        # target must be a system library we should try to link against
+                        libraries += ["{}.lib".format(j)]
 
                 processed_targets += [curr]
 
--- a/mesonbuild/dependencies/base.py
+++ b/mesonbuild/dependencies/base.py
@@ -1411,7 +1411,8 @@ class CMakeDependency(ExternalDependency):
                                       'Valid targets are:\n{}'.format(name, list(self.traceparser.targets.keys())))
 
         # Set dependencies with CMake targets
-        reg_is_lib = re.compile(r'^(-l[a-zA-Z0-9_]+|-pthread)$')
+        # recognise arguments we should pass directly to the linker
+        reg_is_lib = re.compile(r'^(-l[a-zA-Z0-9_]+|-pthread|-delayload:[a-zA-Z0-9_\.]+|[a-zA-Z0-9_]+\.lib)$')
         processed_targets = []
         incDirs = []
         compileDefinitions = []
--- a/mesonbuild/dependencies/base.py
+++ b/mesonbuild/dependencies/base.py
@@ -1481,7 +1481,9 @@ class CMakeDependency(ExternalDependency):
                 for j in otherDeps:
                     if j in self.traceparser.targets:
                         targets += [j]
-                    elif reg_is_lib.match(j) or os.path.exists(j):
+                    elif reg_is_lib.match(j):
+                        libraries += [j]
+                    elif os.path.isabs(j) and os.path.exists(j):
                         libraries += [j]
                     elif mesonlib.is_windows() and reg_is_maybe_bare_lib.match(j):
                         # On Windows, CMake library dependencies can be passed as bare library names,
@@ -1491,6 +1493,8 @@ class CMakeDependency(ExternalDependency):
                         # same, but must assume any bare argument passed which is not also a CMake
                         # target must be a system library we should try to link against
                         libraries += ["{}.lib".format(j)]
+                    else:
+                        mlog.warning('CMake: Dependency', mlog.bold(j), 'for', mlog.bold(name), 'target', mlog.bold(self._original_module_name(curr)), 'was not found')
 
                 processed_targets += [curr]
 


mesonbuild/dependencies/base.py
```diff
@@ -1477,10 +1480,23 @@ def _detect_dep(self, name: str, modules: T.List[T.Tuple[str, bool]], components
...
                         libraries += [j]
+                    elif mesonlib.is_windows() and not is_abs_path:
+                        if not any([j.endswith(x) for x in ('.lib', '.dll')]):
+                            if self.static:
+                                j += ".lib"
+                            else:
+                                j += ".dll"
+                        libraries += [j]
+                    else:
+                        mlog.warning('CMake: Dependency', mlog.bold(j), 'for', mlog.bold(name), 'target', mlog.bold(self._original_module_name(curr)), 'was not found')
```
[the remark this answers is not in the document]
mensinda: No, there are a lot of (let's call them "interesting") CMake projects that do "fun" stuff (even in their dependency packages). So making this into a hard error will likely break projects that are currently working just fine.

Also, we didn't have any issues with silently dropping stuff until now (no false positive/broken libraries), so upgrading it to a warning is probably a good idea but making this an error would require seeing how many projects are hit with this warning first.

mesonbuild/dependencies/base.py
```diff
@@ -1477,10 +1480,23 @@ def _detect_dep(self, name: str, modules: T.List[T.Tuple[str, bool]], components
                     otherDeps += [x for x in tgt.properties['IMPORTED_LINK_DEPENDENT_LIBRARIES'] if x]
 
                 for j in otherDeps:
+                    is_abs_path = os.path.split(j)[0]
```
mensinda, line 1483: Why is this variable named `is_abs_path` when `os.path.split` is used? Are you sure you didn't want to use `os.path.isabs()`?

mesonbuild/dependencies/base.py
```diff
@@ -1411,7 +1411,10 @@ def _detect_dep(self, name: str, modules: T.List[T.Tuple[str, bool]], components
                                       'Valid targets are:\n{}'.format(name, list(self.traceparser.targets.keys())))
 
         # Set dependencies with CMake targets
-        reg_is_lib = re.compile(r'^(-l[a-zA-Z0-9_]+|-pthread)$')
+        if mesonlib.is_windows():
+            reg_is_lib = re.compile(r'(^-delayload:[a-zA-Z0-9_\.]+|[a-zA-Z0-9_]+\.(lib|dll))$')
```
mensinda, line 1415: Could you combine this into one big regex? `is_windows()` Is also true on MSYS and doesn't necessarily imply the use of Microsofts compiler.

mesonbuild/dependencies/base.py
```diff
@@ -1477,10 +1480,23 @@ def _detect_dep(self, name: str, modules: T.List[T.Tuple[str, bool]], components
...
                 for j in otherDeps:
+                    is_abs_path = os.path.split(j)[0]
                     if j in self.traceparser.targets:
                         targets += [j]
-                    elif reg_is_lib.match(j) or os.path.exists(j):
+                    elif reg_is_lib.match(j):
+                        libraries += [j]
+                    elif is_abs_path and os.path.exists(j):
                         libraries += [j]
+                    elif mesonlib.is_windows() and not is_abs_path:
```
mensinda, line 1490: This branch will break if there is random junk (stray link flags, etc.) in one of the properties. Could you please make the check more restrictive, so that only valid libraries are added?

mesonbuild/dependencies/base.py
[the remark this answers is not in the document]
fooishbar: Thanks! That makes sense.

mesonbuild/dependencies/base.py
fooishbar: So just have a single regex for all systems, matching all of -lfoo, -delayload:foo, foo.lib and foo.dll?

mesonbuild/dependencies/base.py
fooishbar: How do I know what a valid library is? I don't have a compiler object to hand, since the dependency doesn't intrinsically include a language.

mesonbuild/dependencies/base.py
mensinda: Yes.

mesonbuild/dependencies/base.py
fooishbar: Thanks, I missed `isabs` in the `os.path` documentation somehow. But the variable is more or less what it claims to be: `os.path.split()` splits into a two-element `(dirname,basename)` tuple, so for absolute and relative paths the `[0]` element will evaluate to a truthy value, and not for bare arguments (e.g. `version` and `uuid`).

mesonbuild/dependencies/base.py
mensinda: I am not sure that there is a good solution. But we should avoid having stuff like `-Dfoobar=stuff.dll` in the link flags. Do you have a link to the CMake documentation on how exactly it is handled there?

If not, then maybe use some sort of regex to check if library name looks OK (starts with an alphanumerical char, etc.)

\<rant\>Who thought it was a good idea to allow plain names like `version` (that are also practically indistinguishable from CMake targets) as valid library names?\</rant\>

fooishbar:
The three CI failures (`checkarg` failing in a cross-compilation environment which doesn't use CMake, increased Clang 10.x strictness on Windows) all seem unrelated.

mesonbuild/dependencies/base.py
mensinda: No problem, that's what code review is for :)

mensinda:
Yes, these CI failures are currently a know issue.

mesonbuild/dependencies/base.py
fooishbar: > I am not sure that there is a good solution. But we should avoid having stuff like -Dfoobar=stuff.dll in the link flags.

OK, so how about the regex as you mentioned, restricting this behaviour to only bare names (`^[a-zA-Z0-9_]+$`)?

> Do you have a link to the CMake documentation on how exactly it is handled there?

It brute-forces across a range of possible prefix/suffix combinations and tries a compiler with them to see which one will work.

> <rant>Who thought it was a good idea to allow plain names like version (that are also practically indistinguishable from CMake targets) as valid library names?</rant>

Yes, I definitely have some opinions on CMake + LLVM + Windows after the past couple of weeks.

mensinda: > OK, so how about the regex as you mentioned, restricting this behaviour to only bare names (^[a-zA-Z0-9_]+$)?

Yes, this regex should work for now. But please also add a comment explaining the situation, otherwise, there will be confusion in the future.

fooishbar: Thanks a lot for all the great suggestions. I've implemented them all now and pushed.

mesonbuild/dependencies/base.py
```diff
@@ -1479,8 +1481,25 @@ def _detect_dep(self, name: str, modules: T.List[T.Tuple[str, bool]], components
...
+                        # e.g. 'version' should translate into 'version.lib' or 'version.dll' depending
+                        # on the link mode. CMake brute-forces a combination of prefix/suffix
+                        # combinations to find the right library, however as we do not have a compiler
+                        # environment available to us, we cannot do the same, but must assume any
+                        # bare argument passed which is not also a CMake target must be a library we
+                        # should try to link against
+                        if self.static:
+                            libraries += ["{}.lib".format(j)]
+                        else:
+                            libraries += ["{}.dll".format(j)]
```
jenatali, line 1499: I wasn't aware that you could link directly to .dlls… I'm pretty sure you have to link to a .lib always, it's just that a static .lib contains code, where a dynamic .lib only contains stubs which add entries to the import tables.

fooishbar: Oh, OK. So should we always just pass `.lib` in this case?

jenatali: Yeah, I think so. I think I've seen tools that can generate a corresponding import lib from from a dll, but the linker doesn't take dlls directly as inputs.

mesonbuild/dependencies/base.py
```diff
@@ -1411,7 +1411,9 @@ def _detect_dep(self, name: str, modules: T.List[T.Tuple[str, bool]], components
                                       'Valid targets are:\n{}'.format(name, list(self.traceparser.targets.keys())))
 
         # Set dependencies with CMake targets
-        reg_is_lib = re.compile(r'^(-l[a-zA-Z0-9_]+|-pthread)$')
+        # recognise arguments we should pass directly to the linker
+        reg_is_lib = re.compile(r'(^-l[a-zA-Z0-9_\.]+|-pthread|-delayload:[a-zA-Z0-9_\.]+|[a-zA-Z0-9_]+\.(lib|dll))$')
```
jenatali, line 1415: I think dll could be removed from here as well, but it probably doesn't cause any harm.

fooishbar: Gotcha, done this as well. Thanks!

PR #6991, opened 2020-04-18
compilers: Silence warning about gnu_inline with clang

nirbheek:
The warning is due to a change in behaviour in Clang 10 and newer:

https://releases.llvm.org/10.0.0/tools/clang/docs/ReleaseNotes.html#c-language-changes-in-clang

This was already fixed for clang++, but not for clang for some reason. It was also fixed incorrectly; by adding `extern` instead of moving from `-Werror` to `-Werror=attributes`.

Should fix the remaining failure in the CI (msys2 clangx64ninja):

```
Compiler stderr:
 D:\a\1\s\b 14fe268f30\meson-private\tmplco6n27n\testfile.cpp:1:23: error: 'gnu_inline' attribute without 'extern' in C++ treated as externally available, this changed in Clang 10 [-Werror,-Wgnu-inline-cpp-without-extern]
inline __attribute__((gnu_inline)) int foo(void) { return 0; }
                      ^
1 error generated.

Compiler for C++ supports function attribute gnu_inline: NO 

test cases\common\203 function attributes\meson.build:90:2: ERROR: Assert failed: clang: gnu_inline
```

### DIFF
--- a/mesonbuild/compilers/c_function_attributes.py
+++ b/mesonbuild/compilers/c_function_attributes.py
@@ -127,7 +127,4 @@ CXX_FUNC_ATTRIBUTES = {
          'static int (*resolve_foo(void))(void) { return my_foo; }'
          '}'
          'int foo(void) __attribute__((ifunc("resolve_foo")));'),
-    # Clang >= 10 requires the 'extern' keyword
-    'gnu_inline':
-        'extern inline __attribute__((gnu_inline)) int foo(void) { return 0; }',
 }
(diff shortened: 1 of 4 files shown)


dcbaker:
This wasn't changed for clang, and only for clang++ because it doesn't apply to clang. The change is specific to the peculiarity of using `gnu_inline` in C++ code, the C behavior didn't actually change. [Clang patch discussion](https://reviews.llvm.org/D67414).

Either way it probably makes more sense to use `-Werror=attributes` anyway.

Issue #6918, opened 2020-04-06
Required ninja version is inconsistent


tbodt:
https://github.com/mesonbuild/meson/blob/00104b1081cd5076451d83524f4f9bfa75750cf6/mesonbuild/backend/ninjabackend.py#L297
https://github.com/mesonbuild/meson/blob/00104b1081cd5076451d83524f4f9bfa75750cf6/mesonbuild/backend/ninjabackend.py#L306
https://github.com/mesonbuild/meson/blob/8bdc5c2dd564fc685e82f67480ad0adc43c823b2/mesonbuild/environment.py#L154
Is it 1.5, 1.5.1, or 1.7?

xclaesse:
At least for Meson >=0.54 the minimum required version is 1.7, according to changelog. We should indeed update all those places, ideally storing the minimum version requires in a single place.

eli-schwartz:
Fixed in #6993...

PR #6993, opened 2020-04-19
fix ninja version inconsistent

michaelbadcrumble:
This should fix (#6918)


eli-schwartz:
> fix (#6918)

The () here is pointless and prevented autoclosing the linked issue.

PR #6995, opened 2020-04-19
switch python2 %s for python3 .format

michaelbadcrumble:


QuLogic:
This is not a Python 2 vs Python 3 thing.

michaelbadcrumble:
I know but both syntactic styles shouldn’t be mixed. I find it very tempting to use `f-strings`.

PR #7002, opened 2020-04-20
find_program: Fixes when the program has been overriden by executable

xclaesse:
- ExternalProgramHolder has path() method while CustomTargetHolder and
  BuildTargetHolder have full_path().
- The returned ExternalProgramHolder's path() method was broken, because
  build.Executable object has no get_path() method, it needs the
  backend.
- find_program('overriden_prog', version : '>=1.0') was broken because
  it needs to execute the exe that is not yet built. Now assume the
  program has the (sub)project version.
- If the version check fails, interpreter uses
  ExternalProgramHolder.get_name() for the error message but
  build.Executable does not implement get_name() method.


QuLogic:
s/overriden/overridden/g

PR #7003, opened 2020-04-21
WIP: Create enum to use instead of strings for language

Ericson2314:
I think this should help prevent bugs. Hope others can help me with the refactor! Pushed it to the main repo for sake of collaboration.


jpakkane:
I have not had time to go through the code, but conceptually this is good.

PR #7008, opened 2020-04-21
Document formal Meson grammar

Flow-It:
My main motivation for this is that I wanted to lookup Meson grammar to use something similar for my own DSL project, but there was no formal definition to be found. I argue having such a formal definition could prove useful in the future as a reference when working on the parser component.

This is all still WIP, but I wanted to share the current status with you here to hopefully get some early feedback.

The grammar is expressed in EBNF, but I found there is no universally agreed upon syntax for EBNF, so here the same syntax Python uses for its grammar description is used.

Expression precedence was taken from C++ and I'm 99% sure that the same precedence rules apply in Meson, but that is a thing I have to check conclusively.

Most of the TODOs and FIXMEs are more hints to me, and will hopefully soon disappear, but if someone wants to voice his opinion on certain questions I'm happy for the input.

### DIFF
--- a/docs/markdown/Syntax.md
+++ b/docs/markdown/Syntax.md
@@ -588,3 +588,64 @@ FAQ](FAQ.md#why-is-meson-not-just-a-python-module-so-i-could-code-my-build-setup
 because of this limitation you find yourself copying and pasting code
 a lot you may be able to use a [`foreach` loop
 instead](#foreach-statements).
+
+Grammar
+--
+
+This is the full Meson grammar, as it is used to parse Meson build definition files:
+
+```
+program = (NEWLINE | stmt)*
+stmt = (expression_stmt | selection_stmt | iteration_stmt | jump_stmt) NEWLINE
+selection_stmt = 'if' condition NEWLINE (stmt)* ('elif' condition NEWLINE (stmt)*)* ['else' (stmt)*] 'endif'
+iteration_stmt = 'foreach' identifier_list ':' identifier NEWLINE (stmt)* 'endforeach'
+condition = conditional_expr // TODO classic way would be condition = expression, but we can't have assignments in conditions. Should that be covered by the grammar (syntax) or not
+identifier_list = identifier | (identifier_list ',' identifier)
+jump_stmt = 'break' | 'continue' // TODO can currently only occur inside a iteration_stmt body, but that is not reflected here
+parameter_list = [arg_list] | [kwarg_list] | (arg_list ',' kwarg_list) // TODO could be more concise
+arg_list = expr_list
+kwarg_list = kwarg | (kwarg_list ',' kwarg)
+kwarg = string_literal ':' expr // TODO or could the key here actually be an expression?
+expression_stmt = expr
+expr_list = assignment_expr | (expr_list ',' assignment_expr)
+expr = assignment_expr
+expr_list = assignment_expr | expression_list ',' assignment_expr // TODO currently unused, do we need it somewhere?
+assignment_expr = conditional_expr | (logical_or_expr assignment_op assignment_expr)
+assignment_op = '=' | '*=' | '/=' | '%=' | '+=' | '-='
+conditional_expr = logical_or_expr | (logical_or_expr '?' expr ':' assignment_expr
+logical_or_expr = logical_end_expr | (logical_or_expr 'or' logical_and_expr')
+// FIXME in between here would be xor, if we had it
+logical_and_expr = equality_expr | (logical_and_expr 'and' equality_expr)
+equality_expr = relational_expr | (equality_expr equality_op relational_expr)
+equality_op = '==' | '!='
+relational_expr = additive_expr | (relational_expr relational_op additive_expr)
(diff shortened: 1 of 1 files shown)


mensinda:
Thanks for doing this. Also, #6765 might also be interesting to you since it will give you the AST for a `meson.build`.

mensinda:
Also, please put `[skip ci]` in the commits for documentation only changes.

docs/markdown/Syntax.md
```diff
@@ -588,3 +588,64 @@ FAQ](FAQ.md#why-is-meson-not-just-a-python-module-so-i-could-code-my-build-setup
...
+stmt = (expression_stmt | selection_stmt | iteration_stmt | jump_stmt) NEWLINE
+selection_stmt = 'if' condition NEWLINE (stmt)* ('elif' condition NEWLINE (stmt)*)* ['else' (stmt)*] 'endif'
+iteration_stmt = 'foreach' identifier_list ':' identifier NEWLINE (stmt)* 'endforeach'
+condition = conditional_expr // TODO classic way would be condition = expression, but we can't have assignments in conditions. Should that be covered by the grammar (syntax) or not
+identifier_list = identifier | (identifier_list ',' identifier)
+jump_stmt = 'break' | 'continue' // TODO can currently only occur inside a iteration_stmt body, but that is not reflected here
+parameter_list = [arg_list] | [kwarg_list] | (arg_list ',' kwarg_list) // TODO could be more concise
+arg_list = expr_list
+kwarg_list = kwarg | (kwarg_list ',' kwarg)
+kwarg = string_literal ':' expr // TODO or could the key here actually be an expression?
```
dcbaker, line 608: The answer is yes this can be an expression

docs/markdown/Syntax.md
```diff
@@ -588,3 +588,64 @@ FAQ](FAQ.md#why-is-meson-not-just-a-python-module-so-i-could-code-my-build-setup
...
+expression_stmt = expr
+expr_list = assignment_expr | (expr_list ',' assignment_expr)
+expr = assignment_expr
+expr_list = assignment_expr | expression_list ',' assignment_expr // TODO currently unused, do we need it somewhere?
+assignment_expr = conditional_expr | (logical_or_expr assignment_op assignment_expr)
+assignment_op = '=' | '*=' | '/=' | '%=' | '+=' | '-='
+conditional_expr = logical_or_expr | (logical_or_expr '?' expr ':' assignment_expr
+logical_or_expr = logical_end_expr | (logical_or_expr 'or' logical_and_expr')
```
dcbaker, line 616: I think you have a typo here end -> and

docs/markdown/Syntax.md
```diff
@@ -588,3 +588,64 @@ FAQ](FAQ.md#why-is-meson-not-just-a-python-module-so-i-could-code-my-build-setup
...
+nonzero_digit = NON_ZERO_DIGIT
+octal_literal = '0o' (octal_digit)*
+octal_digit = OCTAL_DIGIT
+hex_literal = '0x' (hex_digit)*
+hex_digit = HEX_DIGIT
+string_literal = "'" char_seq "'" // TODO multiline strings
+char_seq = (string_char)* // FIXME escapes
+boolean_literal = 'true' | 'false'
+array_literal = '[' expr_list ']' // TODO not assignment expressions unfortunately
+dictionary_literal = '{' kwarg_list '}'
```
dcbaker, line 650: Something's wrong here, either dictionary or kwargs are wrong.

A dictionary looks like this:
```meson
key2 = 'key2'
{'key' : 'value', key2 : 'value2'}
```

This is because you can use variables (as long as they're strings) as keys in dicts, but not as keywords.

but keyword arguments look like this:

```meson
dependency('foo', version : '>= 1.0')
```

docs/markdown/Syntax.md
Flow-It: The question "what kind of expression is allowed here" appears in multiple places. I think I go with the base expression, which can be assignment expression. Something lilke `func(kwarg: a = 'abc')` is syntactically valid, it's just that Meson evaluates assignment expressisions to `None`, so the type system errors out here.

Flow-It:
> Also, please put `[skip ci]` in the commits for documentation only changes.

I did with the follow up commits, but the pipelines started anyway. What am I missing?

eli-schwartz:
> I did with the follow up commits, but the pipelines started anyway. What am I missing?

Not every CI system supports "[skip ci]", so they will start anyway but then a python script will check the commit message, notice it should be skipped, and exit successfully after merely seconds instead of tens of minutes.

Flow-It:
I think I have all now, so open for official review. I kept the individual commits, so changes could be followed easier, but I suggest squashing them all together before merging.

jpakkane:
This should mention that this is not set in stone and that we reserve the right to change the actual syntax (at least until 1.0 is out).

Flow-It:
I added a note about the possibility of the syntax changing. I feel that giving examples of how the syntax might change is important, but it's impossible to list everything there.

PR #7014, opened 2020-04-21
dependency: log cached or skipped dependencies with reference to modules

eli-schwartz:
If the dependency is a special dependency which takes modules, the modules get cached separately and probably reference different pkg-config files. It's very plausible that we have multiple dependencies, using different modules. For example:

```
  Run-time dependency qt5 (modules: Core) found: YES 5.14.2 (pkg-config)
  Dependency qt5 skipped: feature gui disabled
```

Obviously this makes no sense, because of course we found qt5 and even used it. The second line is a lot more readable if it shows this:

```
  Dependency qt5 (modules: Widgets) skipped: feature gui disabled
```

Similar confusion abounds in the case where a module is found in the cache -- which module, exactly, has been found here:

```
  Dependency qt5 found: YES 5.14.2 (cached)
```

Rewrite the dependency function to *consistently* pass around (and use!) the display_name even in cases where we know it isn't anonymous (this is just more correct anyway), and make it serve a dual purpose by also appending the list of modules just like we do for pretty-printing that a dependency has just been found for the first time.

### DIFF
--- a/mesonbuild/interpreter.py
+++ b/mesonbuild/interpreter.py
@@ -3406,7 +3406,7 @@ external dependencies (including libraries) must go to "dependencies".''')
         display_name = name if name else '(anonymous)'
         mods = extract_as_list(kwargs, 'modules')
         if mods:
-            display_name += ' (modules: {})'.format(', '.join(mods))
+            display_name += ' (modules: {})'.format(', '.join(str(i) for i in mods))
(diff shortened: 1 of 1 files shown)


eli-schwartz:
CI failure because if I'm reading this correctly, a unittest fails *the wrong way* due to a module argument being a non-string?

PR #7015, opened 2020-04-22
Adjust regex to handle cases such as C:/Program Files/

georgev93:
I noticed the existing reg_start pattern wouldn't allow the following case to pass:

['C:/Program', 'Files', '(x86)/folder/include']

The new reg_start allows for spaces in the first folder, or in the first folder after the drive.

### DIFF
--- a/mesonbuild/cmake/traceparser.py
+++ b/mesonbuild/cmake/traceparser.py
@@ -655,7 +655,7 @@ class CMakeTraceParser:
 
         # Try joining file paths that contain spaces
 
-        reg_start = re.compile(r'^([A-Za-z]:)?/.*/[^./]+$')
+        reg_start = re.compile(r'^([A-Za-z]:)?/(.*/)*[^./]+$')
         reg_end = re.compile(r'^.*\.[a-zA-Z]+$')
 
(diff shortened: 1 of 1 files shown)


georgev93:
`/path with space/file.txt` would match the proposed reg_start, but _guess_file would still return the correct output because not matching reg_start isn't a necessary exit condition. `/path with space/file.txt` would be properly joined because it matches the reg_end pattern. Just as `C:/Program Files/folder/include` would be properly joined because it is a file path that exists (even though it still matches reg_start)

mensinda:
I still have to say no to the current approach. You are right that it probably wouldn't matter in this example but this change would make the last part of the regex practically pointless. Please make the regex more explicit instead. Maybe use something like `(.*/[^./]+|[^./]+|)`?

georgev93:
Ah, I see what you're talking about. How about the following pattern:

`^([A-Za-z]:)?/(.*/)*[^./]+$`

This should allow spaces in the first directory after the drive while also rejecting:

- anything with a `.` after the last `/`
- anything ending with `/`

Does this properly capture your design intent?

PR #7018, opened 2020-04-22
mtest: Generate JUnit results for tests

dcbaker:
JUnit is a widely supported (and mostly standard) test result format. A number of services can consume it directly or with standard plugins. Gitlab and jenkins are two common consumers of this format.

This adds a dependency on lxml for unittests (but not to run meson), to validate the generated XML against the JUnit schema.


dcbaker:
@jpakkane, does this look good?

PR #7019, opened 2020-04-23
Support VS /MP option to speed up Visual Studio builds

jenatali:
Adds a b_vsmp option which defaults to true (hopefully I did this right). When set, it adds /MP to the cl.exe command line used by Visual Studio (via the MultiProcessorCompilation vcxproj).

In order for this to actually work, we also need to not override the output filename for compiled objects. Unfortunately, this can introduce name collisions for files with the same name coming from separate directories. To avoid this, leave the filename as just the output directory *only* when the input file is in the same directory as the build. This is good enough for most cases.

This speeds up build times significantly in most cases, though in some, it can slow things down (hence the option), due to multi-process parallelism both at the project level *and* the file level, resulting in too many processes to make sufficient forward progress. This is a very uncommon problem.

### DIFF
--- a/docs/markdown/Build-options.md
+++ b/docs/markdown/Build-options.md
@@ -218,6 +218,9 @@ project('my_project', 'c', default_options: ['backend_startup_project=my_exe'])
 executable('my_exe', ...)
 ```
 
+The backend\_multithreaded option can be set to set the "multi-processor
+compilation" (`/MP`) option for visual studio targets to speed up builds.
+
 ### Ninja
 
 #### Max links
--- a/docs/markdown/snippets/visualstudio_multithreading.md
+++ b/docs/markdown/snippets/visualstudio_multithreading.md
@@ -1,0 +1,5 @@
+## Visual Studio Multiprocessor Compilation support
+
+The `backend_multithreaded` option can now be set for Visual Studio backends
+to set the `/MP` flag, speeding up compilation for targets with a lot of
+source files. Defaults to true.
--- a/mesonbuild/backend/vs2010backend.py
+++ b/mesonbuild/backend/vs2010backend.py
@@ -887,6 +887,8 @@ class Vs2010Backend(backends.Backend):
                 ET.SubElement(clconf, 'ExceptionHandling').text = 'false'
             else: # 'sc' or 'default'
                 ET.SubElement(clconf, 'ExceptionHandling').text = 'Sync'
+        if self.environment.coredata.backend_options['backend_multithreaded']:
+            ET.SubElement(clconf, 'MultiProcessorCompilation').text = 'true'
         # End configuration
         ET.SubElement(root, 'Import', Project=r'$(VCTargetsPath)\Microsoft.Cpp.props')
         generated_files, custom_target_output_files, generated_files_include_dirs = self.generate_custom_generator_commands(target, root)
@@ -1274,7 +1276,15 @@ class Vs2010Backend(backends.Backend):
                     self.add_additional_options(lang, inc_cl, file_args)
                     self.add_preprocessor_defines(lang, inc_cl, file_defines)
                     self.add_include_dirs(lang, inc_cl, file_inc_dirs)
-                    ET.SubElement(inc_cl, 'ObjectFileName').text = "$(IntDir)" + self.object_filename_from_source(target, s)
+                    obj_filename = self.object_filename_from_source(target, s)
+                    # For source files in the current directory, just let VS pick their output filename
+                    # If the output filename isn't just a directory, it'll prevent it from being parallelized
+                    # with other source files in the same project
+                    # However, for source files not in the current directory, continue to customize the filename
+                    # to prevent collisions for files with the same name from different directories
+                    if obj_filename == s.fname + '.obj':
+                        obj_filename = ''
+                    ET.SubElement(inc_cl, 'ObjectFileName').text = "$(IntDir)" + obj_filename
             for s in gen_src:
                 if path_normalize_add(s, previous_sources):
                     inc_cl = ET.SubElement(inc_src, 'CLCompile', Include=s)
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -552,6 +552,10 @@ class CoreData:
                 UserStringOption(
                     'Default project to execute in Visual Studio',
                     '')
+            self.backend_options['backend_multithreaded'] = \
+                UserBooleanOption(
+                    'Use /MP switch for multithreaded compilation with cl.exe',
+                    True)
 
     def get_builtin_option(self, optname: str, subproject: str = '') -> T.Union[str, int, bool]:
         raw_optname = optname


nirbheek:
As I understand it, `/MP` is mostly only useful for C++ projects, right? Maybe we should enable it by default for only that?

jenatali:
> As I understand it, `/MP` is mostly only useful for C++ projects, right? Maybe we should enable it by default for only that?

@nirbheek Yeah, for C and C++. I forget sometimes that VS can support other languages like C#. And I'm not a Meson expert unfortunately, just trying to address minor inconveniences that I encounter while using it. I'm happy to iterate if someone wants to point me to the correct places to scope this down.

jpakkane:
If there is no harm in having MP on in other languages, let's just always default it to true.

The more interesting question would be how to enable this for the Ninja backend. Is it just a matter of adding compiler flags or do you need to do something else? Is it enough for us to tell people to set the ninja backend's maximum link process count to a lower value via the corresponding option?

jenatali:
Visual Studio does parallelism differently from Ninja. Ninja parallelizes at the file level by default, while MSBuild only parallelizes at the project level. In order to get file-level parallelization for C++ compiling, there's functionality built into the MSVC compiler to spawn multiple processes there. If you use MSBuild to drive (e.g.) clang-cl, you can't get file-level parallelism, because clang-cl doesn't support /MP.

So, Ninja doesn't need this switch, only the actual VS compiler does.

jenatali:
Basically, without this patch, if you have a projects with a lot of files not broken into smaller libraries, Visual Studio projects from Meson take a *lot* longer to build than equivalent Ninja projects.

nirbheek:
That's not the complete story though. `/MP` was created specifically with C++ in mind because single-file compilation can often take a _very_ long time because of templates or compiler-complex C++14/17 features. So with Ninja depending on the details of your project, you might still see improvements from this.

As a trivial example, let's say you have 24 cores and you build a target with 8 source files that depends on 4 targets with 4 sources each. Most of your cores will be idle while building this.

I think we should implement `/MP` for ninja too, but disable it by default.

jenatali:
Hm, I'm not following that. Ninja compiles each file individually, where MSBuild passes a group of files at once. `/MP` doesn't do anything if you pass a single file to the command line, it only forks out to multiple processes if there's multiple files.

jpakkane:
I went through the documentation and based on my understanding this is what happens in your example. Please correct me if I got something wrong.

> As a trivial example, let's say you have 24 cores and you build a target with 8 source files that depends on 4 targets with 4 sources each. Most of your cores will be idle while building this.

The current behaviour is that the solution builder launches the four dependency targets, with each taking one core (4 cores used at once). When all these are done, it will run the 8 file project, which will use only one core.

With this change the solution builder will launch all 4 (?) dependency targets and all will use multiple processes (4x4, meaning 16 in total). After those are done, the master project is built and it will use 8 cores (because it has 8 files).

If this is the case then adding `/MP` to Ninja would bring no benefit. The obvious problematic case with this is if you have 10 dependency projects with 10 files each, which would lead many processes. In this case you'd need something like the Make jobserver protocol to make optimal use of cores but that's something the VS dev team has to provide via the toolchain, we can't add that after the fact.

jenatali:
Yes, exactly. This happens when building LLVM via CMake with Visual Studio, for example - you end up with too many processes and the system slows to a crawl, making less progress than if you had less parallelism.

But as you say, this is a flaw with Visual Studio, and why this should be optional. I think defaulting to "on" is the right choice here (again matching CMake), since in almost all instances it ends up being a significant win.

You can also configure parallelism both at the CL.exe /MP level (e.g. /MP4) and at the MSBuild.exe /m level (e.g. /m4) to override the default scaling to all cores, which can help. The VS UI doesn't expose options for adding these numbers, but command lines do.

nirbheek:
> Hm, I'm not following that. Ninja compiles each file individually, where MSBuild passes a group of files at once. `/MP` doesn't do anything if you pass a single file to the command line, it only forks out to multiple processes if there's multiple files.

Ouch. So my understanding was incorrect. That's disappointing.

I agree with the PR as-is then.

jpakkane:
Ok, there are two things that come to mind:

- as this is a new feature, it should have a release note snippet under `docs/markdown/snippets`
- since this is a backend specific feature, it should be defined as a backend option rather than a base option

Currently the VS backend does not have options, but you can look at how `backend_max_links` is done specified in the Ninja backend.

tomaszduda23:
I would prefer if you use old way for resolving file name conflicts. https://github.com/mesonbuild/meson/blob/52eab4b00674b160ba1bdaec8ceeb1114143063d/mesonbuild/backend/vs2010backend.py. It seems to be more general. It should be improve to handle:
- item.cpp and item.c
- Item.cpp and item.cpp

jpakkane:
⚔️ conflicts.

jenatali:
Sorry for leaving this sitting for so long, I finally decided to try to actually get this in :)

@jpakkane:

> as this is a new feature, it should have a release note snippet under docs/markdown/snippets

Done

> since this is a backend specific feature, it should be defined as a backend option rather than a base option

Done - looks like VS actually gained a backend option since I originally prototyped this, so I had a pattern to follow

@tomaszduda23:

> I would prefer if you use old way for resolving file name conflicts.

Visual Studio will refuse to multithread compilations if there's an object name set. If there's an object name, it means it needs to use a dedicated cl.exe instance for that source file, so it can pass the specific object name. If there's just a directory, then MSBuild gathers all of the source files with that directory, and passes them to cl.exe at once, which means the /MP command line option can actually multi-thread their compilation.

The comment should make it clear what's going on - if the name was just going to end up being <objdir>\<objname>.obj (because there were no conflicts), then just strip the <objname>.obj bit and let VS append it while multithreading.

jpakkane:
Rebasing should get the tests to pass, they have been fixed in master.

One thing that I wonder is whether `backend_multithreaded` is a suitable name for this, as it implies using multithreading in general whereas this change is about multithreaded linking in particular. Maybe something like `backend_mt_linking`instead?

jenatali:
This isn't about multithreaded linking, it's technically about multiprocess compilation. Specifically cl.exe (not link.exe) spawning multiple processes for multiple input files at once, instead of running them sequentially one at a time.

jpakkane:
Ah, right. That should be an ok name then. However the test failure seems relevant:

   LINK : fatal error LNK1104: cannot open file 'common@sta\main.cc.obj' [D:\a\1\s\b b17497fc84\aarch64@exe.vcxproj]

Dunno if it is relevant, but the private directory seems to be of the old type (`common@sta`) rather than the new one (`common.lib.p` or thereabouts). That is the naming convention for the Ninja backend at least, dunno if the VS backend was ever updated to the new naming scheme...

jenatali:
Yeah, I see that... Seems there's some more dependencies on the obj filename than I expected or understood. Let me see if I can figure it out...

jenatali:
Ah, I see it. Any change I make would need to be made in `object_filename_from_source` directly, so that `get_outputs` and `determine_ext_objs` reference the correct files. Unfortunately, VS's naming scheme for output files is to drop the source extension (e.g. `foo.c` becomes `foo.obj`) where Meson's appends the obj extension (`foo.c` becomes `foo.c.obj`). If we want parallelization, we'd need to change Meson's obj naming scheme. Not sure if that's a good idea.

On the other hand, I've recently learned of an MSBuild-level parallelism feature: https://devblogs.microsoft.com/cppblog/improved-parallelism-in-msbuild/

Considering it's experimental and opt-in, it probably doesn't make sense to enable it by default for Meson, but it might make sense to hook it up to this switch and just change the default. Or, I can just abandon this pull request and tell VS Meson developers to set the environment variable.

jpakkane:
> If we want parallelization, we'd need to change Meson's obj naming scheme. Not sure if that's a good idea.

It means that having `foo.c` and `foo.cpp` in the same project no longer works because the files would clobber each other.

jenatali:
Yep, good point. Thanks for humoring me, but I think this should get closed in favor of just using the MSBuild feature, either with an environment variable for now, or waiting until they flip it to on-by-default, which I expect won't be long.

jpakkane:
Np, thanks for looking into this.

PR #7021, opened 2020-04-23
Fix cross-compile in Windows with MSVC and UWP

nirbheek:
So, configuration without a native compiler was supposed to work according to the 0.54 release notes, but it doesn't actually work for mandatory compilers. It only works for optional compilers added with `add_languages()`. This fixes that, and also fixes a bug in linker detection with MSVC.

I'd like to get this into the stable point release if possible.

commit e2fe0b1132b5567d35e50617cf92c1f182a0a722:

```
Allow cross-compile without a native compiler
```

commit 4112c37fd3f5608f2e522b9bb5c7fc82397c2014:

```
windows linker: Only guess if we don't know target cpu_family
Unlike `cl.exe, MSVC `link.exe` has a 'default' target machine, and
the actual target machine can be selected with `/MACHINE:`. The
'default' is `X86` if you use the x86 native or cross toolchain and
it's `X64` if you use the x64 native or cross toolchain.

So, if you call `vcvarsall.bat x86_arm64` which is the x86 cross
toolchain for arm64, `link.exe` will default to `X86`, not `ARM64`.
It has to be selected by the build system by passing `/MACHINE:ARM64`

So that's what we do now. In the native build case when a native file
is not passed, the auto-detect will always be correct, and in the
cross build case we will always know the `cpu_family`, so this fixes
it for all cases.
```

commit a54397746a1fb9c47ade7be8f47380606d5495aa:

```
envconfig: Don't log a useless message
When there's no env var candidates to check, we print a useless
message into the log.
```

### DIFF
--- a/run_unittests.py
+++ b/run_unittests.py
@@ -9008,18 +9008,16 @@ def convert_args(argv):
         pytest_args += ['-k', ' or '.join(test_list)]
     return pytest_args
 
+def running_a_single_test(argv, cases):
+    'Check whether we got an argument for running an individual test'
+    for arg in argv:
+        for case in cases:
+            if arg.startswith(case) and '.' in arg and case in arg.split('.'):
+                return True
+    return False
+
 def main():
     unset_envs()
-    try:
-        import pytest # noqa: F401
-        # Need pytest-xdist for `-n` arg
-        import xdist # noqa: F401
-        pytest_args = ['-n', 'auto', './run_unittests.py']
-        pytest_args += convert_args(sys.argv[1:])
-        return subprocess.run(python_command + ['-m', 'pytest'] + pytest_args).returncode
-    except ImportError:
-        print('pytest-xdist not found, using unittest instead')
-    # All attempts at locating pytest failed, fall back to plain unittest.
     cases = ['InternalTests', 'DataTests', 'AllPlatformTests', 'FailureTests',
              'PythonTests', 'NativeFileTests', 'RewriterTests', 'CrossFileTests',
              'TAPParserTests',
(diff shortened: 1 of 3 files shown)


nirbheek:
CC: @jon-turney since he implemented that feature.

jon-turney:
> So, configuration without a native compiler was supposed to work according to the 0.54 release notes, but it doesn't actually work for mandatory compilers. It only works for optional compilers added with `add_languages()`

Only the changes to `add_languages()` were merged for 0.54. The rest of the work is in (to add it to `project()`) is in #6512.

Some of your fixes here are probably needed in addition to that work, since I was only testing with Windows cross to ARM64 there.

nirbheek:
Hmm! That's a lot of changes, and most of what I put in this PR is already in that one. I'll leave this PR open for just the UWP bits then :smile:

jpakkane:
Is this good to go now?

nirbheek:
> Is this good to go now?

No, we should merge https://github.com/mesonbuild/meson/pull/6512 first, and then I'll rebase this on top of that.

nirbheek:
Just needs a Windows ARM64 (UWP) job and it'll be ready for merging.

jon-turney:
> Just needs a Windows ARM64 (UWP) job and it'll be ready for review/merging.

I'd suggest altering the job added in #7718 to fit your needs.

run_unittests.py
```diff
@@ -9008,25 +9008,36 @@ def convert_args(argv):
         pytest_args += ['-k', ' or '.join(test_list)]
     return pytest_args
 
+def running_a_single_test(argv, cases):
```
jon-turney, line 9011: The name doesn't match what this function actually does: check if any arguments which looks like test names are present. Perhaps it should count them, and only avoid test parallelization if there is exactly one?

Great change to have, though!

nirbheek: I split this out into https://github.com/mesonbuild/meson/pull/7724 and incorporated your suggestion :)

nirbheek:
Hum, this PR is not needed anymore because we always set the linker machine as the compiler machine since 06dcbd50eea47b3182081527ea1c0ada01d4d847.

```python
        self.target = target
        self.is_64 = ('x64' in target) or ('x86_64' in target)
        # do some canonicalization of target machine
        if 'x86_64' in target:
            self.machine = 'x64'
        elif '86' in target:
            self.machine = 'x86'
        else:
            self.machine = target
        self.linker.machine = self.machine
```

PR #7027, opened 2020-04-24
modules: python: Allow python detection without distutils module

chouquette:
If distutils isn't present but not required by the called, we can still
return a valid python installation

This allows the python module to be used on build machines with the bare minimal python install: https://code.videolan.org/chouquette/vlc/-/jobs/386119

### DIFF
--- a/mesonbuild/modules/python.py
+++ b/mesonbuild/modules/python.py
@@ -266,10 +266,13 @@ import sys
 install_paths = sysconfig.get_paths(scheme='posix_prefix', vars={'base': '', 'platbase': '', 'installed_base': ''})
 
 def links_against_libpython():
-    from distutils.core import Distribution, Extension
-    cmd = Distribution().get_command_obj('build_ext')
-    cmd.ensure_finalized()
-    return bool(cmd.get_libraries(Extension('dummy', [])))
+    try:
+        from distutils.core import Distribution, Extension
+        cmd = Distribution().get_command_obj('build_ext')
+        cmd.ensure_finalized()
+        return bool(cmd.get_libraries(Extension('dummy', [])))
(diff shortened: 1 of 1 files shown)


eli-schwartz:
Why would you want to return a valid python installation while incorrectly reporting the status of link_libpython? This would break the actual intended use of the feature.

Python.org doesn't document any sort of valid "bare minimal python install", the stdlib must be installed without question if you want to reliably use it for, well, anything.

jpakkane:
More specifically, are you building something with Python (such as bindings) or do you just need Python to run scripts?

eli-schwartz:
Yeah, good point. If you just need python to run scripts, then just use find_program to find your *.py file and run it. This is natively supported if the script is executable and your operating system knows how to open the script using the right interpreter (this should always be the case on Unix)... and specifically because Windows users always have issues with being unable to run scripts properly, meson makes sure this natively works there too:

> Meson will also autodetect scripts with a shebang line and run them with the executable/interpreter specified in it both on Windows (because the command invocator will reject the command otherwise) and Unixes (if the script file does not have the executable bit set). Hence, you must not manually add the interpreter while using this script as part of a list of commands.

And meson will detect a python shebang and run it using its own python installation, since it knows it has one.

This works fine if you only need the stdlib, and in your case you claim you don't even have the stdlib, so you probably don't have other system modules you're depending on and therefore absolutely whichever python meson is using will be fine for you.

chouquette:
Hi,

Using `find_program` was my first idea, but the python module logic to find python regardless of its version felt more flexible, as the goal is indeed to run a few python scripts. 

I didn't realize find_program could also be used to find scripts, so I'll use that instead.

Thanks!

PR #7020, opened 2020-04-23
gnome: Fix usage of gobject-introspection as subproject

xclaesse:

PR #7028, opened 2020-04-24
gnome: fix resolution of gi scanner and compiler in Windows

ylatuya:
Taking the variable directly from pkg-config returns a plain string
with the path to the python script: g-ir-scanner or g-ir-compiler.
This has to be converted to an ExternalProgram so it's properly
called using python.exe


xclaesse:
Makes sense, yes. Since there are more issues with related code and I already had a PR that would conflict with this, I included your change in my PR there: https://github.com/mesonbuild/meson/pull/7020, if you don't mind.

ylatuya:
Cool, I will close this one.

PR #7031, opened 2020-04-24
compilers: convert unix include directory arg to msvc format

boxerab:
This change allows me to build gstreamer on windows using external dependency specified by .pc file.

cc @xclaesse

### DIFF
--- a/mesonbuild/compilers/mixins/visualstudio.py
+++ b/mesonbuild/compilers/mixins/visualstudio.py
@@ -246,6 +246,8 @@ class VisualStudioLikeCompiler(metaclass=abc.ABCMeta):
                 else:
                     i = '/I' + i[10:]
             # -pthread in link flags is only used on Linux
+            elif i.startswith('-I'):
+                result.append('/I' + i[2:])
             elif i == '-pthread':
                 continue
             result.append(i)


xclaesse:
I looks right to me, but @nirbheek  knows msvc better I think.

mesonbuild/compilers/mixins/visualstudio.py
```diff
@@ -246,6 +246,8 @@ def unix_args_to_native(cls, args: T.List[str]) -> T.List[str]:
                 else:
                     i = '/I' + i[10:]
             # -pthread in link flags is only used on Linux
+            elif i.startswith('-I'):
+                result.append('/I' + i[2:])
```
nirbheek, line 250: `-I` and `/I` are identical. What problem did you face for which you had to change this?

nirbheek: Note that for _all_ command-line utilities shipped with Visual Studio, you can pass `/ARG` or `-ARG` and they will be parsed the same.

boxerab: Thanks. This fix is for getting flags from standard *.pc file, which use unix-style arguments.
Unfortunately, `-I` argument does not get recognized without this change.

nirbheek: Can you paste the compiler error you're getting or whatever? You're saying "it does not work" but it's not clear what you're talking about since obviously gst-build and cerbero both work on Windows with MSVC.

boxerab: gst-build doesn't find the include file from the .pc file without this change.

nirbheek: This is not true on my system:

```
Found pkg-config: C:\gstreamer\1.0\dangling_x86_64.msvc\bin\pkg-config.EXE (0.29.2)
Determining dependency 'glib-2.0' with pkg-config executable 'C:\\gstreamer\\1.0\\dangling_x86_64.msvc\\bin\\pkg-config.EXE'
PKG_CONFIG_PATH:
Called `C:\gstreamer\1.0\dangling_x86_64.msvc\bin\pkg-config.EXE --modversion glib-2.0` -> 0
2.62.5
PKG_CONFIG_PATH:
Called `C:\gstreamer\1.0\dangling_x86_64.msvc\bin\pkg-config.EXE --cflags glib-2.0` -> 0
-IC:/gstreamer/1.0/dangling_x86_64.msvc/include/glib-2.0 -IC:/gstreamer/1.0/dangling_x86_64.msvc/lib/glib-2.0/include
PKG_CONFIG_PATH:
Called `C:\gstreamer\1.0\dangling_x86_64.msvc\bin\pkg-config.EXE glib-2.0 --libs` -> 0
-LC:/gstreamer/1.0/dangling_x86_64.msvc/lib -lglib-2.0 -lintl
PKG_CONFIG_PATH:
Called `C:\gstreamer\1.0\dangling_x86_64.msvc\bin\pkg-config.EXE glib-2.0 --libs` -> 0
-LC:/gstreamer/1.0/dangling_x86_64.msvc/lib -lglib-2.0 -lintl
None of 'PKG_CONFIG_PATH' are defined in the environment, not changing global flags.
Run-time dependency glib-2.0 found: YES 2.62.5
```

And then everything builds fine.

boxerab: Happy to report that you are correct! After clean build, all .pc file args are recognized by Meson.
Sorry for the noise.

Issue #1281, opened 2017-01-04
Allow dry-run installation


Nikratio:
It'd be great if there was a way to support a "dry run" installation. At the moment, calling `ninja -d install` isn't helpful at all, because it just shows that it would call Meson.

nirbheek:
This is not possible to implement in Meson because Ninja doesn't actually run anything in dry-run (`-n`) mode (internally, it implements a fake class that just returns `true` without doing anything), so no Meson scripts would be called at all.

Maybe we can implement a `dry-install` Ninja target that will call all the scripts in a dry-run mode.

mine260309:
I would like to see this function as well, e.g. I would like to see which service files are going to be installed to the system directory, without actually installing them.

braindevices:
any progress on this? I just want see same output as `make -n install`.

xclaesse:
Should be pretty easy to implement now that we have 'meson install' command.

PR #7032, opened 2020-04-24
minstall: Add --dry-run option

xclaesse:
Closes: #1281

### DIFF
--- a/mesonbuild/minstall.py
+++ b/mesonbuild/minstall.py
@@ -54,6 +54,7 @@ if T.TYPE_CHECKING:
         quiet: bool
         wd: str
         destdir: str
+        dry_run: bool
 
 
 symlink_warning = '''Warning: trying to copy a symlink that points to a file. This will copy the file,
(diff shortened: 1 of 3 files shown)


mesonbuild/minstall.py
```diff
@@ -217,6 +224,22 @@ def check_for_stampfile(fname):
...
+    os.chmod = noop
+    os.chown = noop
+    os.makedirs = noop
+    os.mkdir = noop
+    os.remove = noop
+    os.symlink = noop
+    shutil.copy = noop
+    shutil.copyfile = noop
+    shutil.copystat = noop
+    depfixer.fix_rpath = noop
```
jpakkane, line 240: I don't like monkey patching the running process like this. One thing I have wanted to do for a while is to run some, preferably all, Meson tool invocations in unittests as direct function calls rather than spawning processes. If you do that, these changes persist (at least I think they do) and can cause interesting things in the caller later on.

nirbheek: Yes, that's probably exactly why the tests fail right now. I think we should use a wrapper for all these functions and do nothing if dry_run is set. Monkey-patching is genius, but also brittle.

xclaesse: Ok, I agree. I wanted to make sure we can't sneak an os.copy somewhere without noticing, but it's cleaner to keep a wrapper indeed. Patch updated.

nt8r:
What's the status of this? It would be nice to have #1281 finished.

nirbheek:
> What's the status of this? It would be nice to have #1281 finished.

There are conflicts and all the tests fail. The approach itself looks fine to me, so it's a matter of fixing those and then inspecting the `meson install --dry-run` output and ensuring that it's the same as `meson install`.

The `os.mkdir -> self.mkdir` etc changes are pretty good for implementing a future 'verbose' mode too, where we print _everything_ we do.

jpakkane:
```
mesonbuild/minstall.py:83:5: error: Function is missing a type annotation for one or more arguments
mesonbuild/minstall.py:287:19: error: "ArgumentType" has no attribute "dry_run"
mesonbuild/minstall.py:339:13: error: Cannot call function of unknown type
```

and a bunch of other ones after that.

xclaesse:
I don't know how to fix them, @dcbaker ^

mesonbuild/minstall.py
```diff
@@ -75,20 +77,27 @@ def add_arguments(parser: argparse.Namespace) -> None:
                         help='Do not print every file that was installed.')
     parser.add_argument('--destdir', default=None,
                         help='Sets or overrides DESTDIR environment. (Since 0.57.0)')
+    parser.add_argument('--dry-run', '-n', action='store_true',
+                        help='Doesn\'t actually install, but print logs.')
 
 class DirMaker:
-    def __init__(self, lf: T.TextIO):
+    def __init__(self, lf: T.TextIO, makedirs):
```
dcbaker, line 84: This is a nasty corner case of typing, passing callables with keyword arguments. the best you can do here (right now) is:
```python
T.Callable[..., None]
```

mesonbuild/minstall.py
```diff
@@ -276,6 +285,27 @@ def __init__(self, options: 'ArgumentType', lf: T.TextIO):
         self.lf = lf
         self.preserved_file_count = 0
 
+        dry_run = options.dry_run
+        noop = lambda *args, **kwargs: None
+        Popen_noop = lambda *args, **kwargs: (types.SimpleNamespace(returncode=0), '', '')
+        run_exe_noop = lambda *args, **kwargs: 0
```
dcbaker, line 291: I'm going to expect mypy to choke on these, and require you to not use lambdas, but I might surprised. Especially the `SimpleNamespace` thing

xclaesse:
@dcbaker I fixed it by open coding each function, should be easier to get type checking like that.

PR #7035, opened 2020-04-25
Recursively call _guess_files in the event that some portion of the c…

georgev93:
…orrect file path resolves to a file path that actually exists on the host machine.

For instance if C:/Program Files (x86)/folder is passed to _guess_files, it would resolve to ['C:/Program Files', '(x86)/folder'] since C:/Program Files is an actual file location that can exist.

### DIFF
--- a/mesonbuild/cmake/traceparser.py
+++ b/mesonbuild/cmake/traceparser.py
@@ -660,25 +660,40 @@ class CMakeTraceParser:
 
         fixed_list = []  # type: T.List[str]
         curr_str = None  # type: T.Optional[str]
+        path_found = False # type: bool
 
         for i in broken_list:
             if curr_str is None:
                 curr_str = i
+                path_found = False
             elif os.path.isfile(curr_str):
                 # Abort concatenation if curr_str is an existing file
                 fixed_list += [curr_str]
                 curr_str = i
+                path_found = False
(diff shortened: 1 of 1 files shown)


mensinda:
The CI error is related. See [this line](https://github.com/mesonbuild/meson/pull/7035/checks?check_run_id=618548452#step:7:1195) in the logs.

If you are on Windows, I would strongly recommend upgrading to CMake >= 3.17 since the `_guess_files` code patch is no longer needed there.

georgev93:
Ah, it does look like I caused that CI error. Also you are correct regarding Cmake 3.17 fixing this issue. I am fine with this PR being closed without merge unless you think something should be done to handle this case.

mesonbuild/cmake/traceparser.py
```diff
@@ -660,6 +660,7 @@ def _guess_files(self, broken_list: T.List[str]) -> T.List[str]:
 
         fixed_list = []  # type: T.List[str]
         curr_str = None  # type: T.Optional[str]
+        path_found = False # type: bool
```
mensinda, line 663: `path_found` currently isn't reset in the other branches (`elif os.path.isfile(curr_str)`, etc.).

georgev93:
Good catch!

Issue #7029, opened 2020-04-24
Better error message for test()


fryeb:
Hi, 
I'm fairly new to meson, and while adding tests to my project I ran into the following error message:
```
..\tests\meson.build:2:0: ERROR: Incorrect number of arguments
```
After a bit of research I was able to narrow down the issue: `test` expects two arguments, not one. I might've solved the issue sooner had the error message read something like:
```
..\tests\meson.build:2:0 ERROR: test() expects 2 arguments but only 1 was given.
```
Admittedly this is a small issue, but I imagine it shouldn't be hard to fix and could help new users a fair bit.

xclaesse:
PR welcome :-)

PR #7040, opened 2020-04-26
Clarified error message for test().

fryeb:
Potential fix for: https://github.com/mesonbuild/meson/issues/7029#issue-606367289

Changes the error message for  `test()` so it now reports the expected and actual number of parameters.

### DIFF
--- a/mesonbuild/interpreter.py
+++ b/mesonbuild/interpreter.py
@@ -3761,7 +3761,7 @@ This will become a hard error in the future.''' % kwargs['input'], location=self
 
     def add_test(self, node, args, kwargs, is_base_test):
         if len(args) != 2:
-            raise InterpreterException('test expects 2 arguments, {} given'.format(len(args))
+            raise InterpreterException('test expects 2 arguments, {} given'.format(len(args)))
         if not isinstance(args[0], str):
             raise InterpreterException('First argument of test must be a string.')
         exe = args[1]
--- a/mesonbuild/interpreter.py
+++ b/mesonbuild/interpreter.py
@@ -3761,7 +3761,7 @@ This will become a hard error in the future.''' % kwargs['input'], location=self
 
     def add_test(self, node, args, kwargs, is_base_test):
         if len(args) != 2:
-            raise InterpreterException(r'test expects 2 arguments, only {len(args)} given')
+            raise InterpreterException('test expects 2 arguments, {} given'.format(len(args))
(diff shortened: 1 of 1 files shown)


mesonbuild/interpreter.py
```diff
@@ -3761,7 +3761,7 @@ def unpack_env_kwarg(self, kwargs) -> build.EnvironmentVariables:
 
     def add_test(self, node, args, kwargs, is_base_test):
         if len(args) != 2:
-            raise InterpreterException('Incorrect number of arguments')
+            raise InterpreterException(r'test expects 2 arguments, only {len(args)} given')
```
SuperWig, line 3764:
```suggestion
            raise InterpreterException(r'test expects 2 arguments, {len(args)} given')
```
"test expects 2 arguments, only 3 given" doesn't really make that much sense.

dcbaker, line 3764: This isn't doing what you're intending it to do. the `r` prefix in python makes a string a `raw string`. Python assumes you mean exactly what you see (no escaping, no unicode interpretation, etc). What I think you meant to do is use an `f` prefix, an `f-string` does inline interpolation, ie:

```python
>>> a = 'foo'
>>> print(f'I have {a}')
I have foo
>>> print(r'I have {a}')
I have {a}
```

Unfortunately f-strings are a python 3.6 feature and meson still needs to work with 3.5, so you'll need to use `str.format()` instead:

```python
raise InterpreterException(r'test expects 2 arguments, {} given'.format(len(args))
```

mesonbuild/interpreter.py
```diff
@@ -3761,7 +3761,7 @@ def unpack_env_kwarg(self, kwargs) -> build.EnvironmentVariables:
 
     def add_test(self, node, args, kwargs, is_base_test):
         if len(args) != 2:
-            raise InterpreterException('Incorrect number of arguments')
+            raise InterpreterException('test expects 2 arguments, {} given'.format(len(args))
```
dcbaker, line 3764:
```suggestion
            raise InterpreterException('test expects 2 arguments, {} given'.format(len(args)))
```

You're missing a closing paren.

fryeb: Yeah I noticed, serves me right for not running the tests locally first.

dcbaker:
lint failure is expected, as long as everything else is green I'll merge this. Thanks again.

PR #7043, opened 2020-04-27
Make `override_find_program` respect `native` keyword

Ericson2314:


xclaesse:
You need to add `navive` kwarg to override_find_program then, right?

Ericson2314:
@xclaesse the idea was to just get the `for_machine` from the override program, but this breaks the native test where they are mixing build and host machines. I think we'll have to do something similar to @jon-turney's PR where there is a native keyword arg: if it's set it must match the program, but if it's unset and it's a native build, then both platforms are overridden for back compat.

Ericson2314:
Longer term, instead of all these hacks for native compilation, what we need is:

- A short-hand to say "give me a `native: true` and `native: false` copy of this

- ninja to deduplicate build rules so "redundant" things like `native: true` and `native: false` copies (but not limited to that) don't result in building the same thing twice. The best thing to do would be some build rule content-addressing like Nix does.

xclaesse:
Note that override_dependency() has the native kwarg.

Ericson2314:
@xclaesse Ah thanks. I will model on that.

Ericson2314:
OK, there is a failure having to do with partial deps I cannot figure out.

jpakkane:
You may want to rebase this, CI on master was broken for a while but is now fixed.

Ericson2314:
@jpakkane will do!

bonzini:
Closing in favor of #15404.

PR #7047, opened 2020-04-27
boost: Only use usage-requirements defines (fixes #7046)

### LINKED ISSUE #7046: Can't build projects dynamically linked to boost.log
Since this change:

https://github.com/mesonbuild/meson/blob/7c68fe80083fd72a100998578417b223e3704af5/mesonbuild/dependencies/boost.py#L791

I can't compile anymore some applications dynamically linked to boost.log:

```
/usr/include/boost169/boost/log/detail/config.hpp:225: error: "BOOST_LOG_DLL" redefined [-Werror]
 #       define BOOST_LOG_DLL

<command-line>: note: this is the location of the previous definition
```

I think that `-DBOOST_LOG_DLL` should not specified (see https://www.boost.org/doc/libs/1_72_0/libs/log/doc/html/log/rationale/namespace_mangling.html) but only `-DBOOST_LOG_DYN_LINK=1`

Tested with boost from 1.69 to 1.72 on CentOS and MacOS.

If you confirm that `-DBOOST_LOG_DLL` can be removed I can provide a PR, just let me know .

mensinda:


tools/boost_names.py
```diff
@@ -43,10 +43,10 @@
...
     def __init__(self, name: str, shared: T.List[str], static: T.List[str], single: T.List[str], multi: T.List[str]):
         self.name = name
-        self.shared = shared
-        self.static = static
-        self.single = single
-        self.multi = multi
+        self.shared = sorted(set(shared))
+        self.static = sorted(set(static))
+        self.single = sorted(set(single))
+        self.multi = sorted(set(multi))
```
dcbaker, line 49: sorted returns an iterator, not a list. you need to wrap this in list as well to avoid changing the type (and running into strange errors if you try to use these more than once.

tools/boost_names.py
```diff
@@ -117,28 +137,36 @@ def get_libraries(jamfile: Path) -> T.List[BoostLibrary]:
...
+                elif colon_counter >= 4:
+                    usage_requirements += [j]
+
             # Get shared / static defines
             shared: T.List[str] = []
             static: T.List[str] = []
             single: T.List[str] = []
             multi: T.List[str] = []
-            for j in parts:
+            for j in usage_requirements + project_usage_requirements:
```
dcbaker, line 155: You might consider `itertools.chain` instead of concatenating lists, to avoid building a concrete list.

tools/boost_names.py
mensinda: Actually, `sorted` does return a list: https://docs.python.org/3/library/functions.html#sorted. You can verify this with `type(sorted([3, 2, 1])).__name__`.

tools/boost_names.py
mensinda: Thanks, I will keep it in mind for the next time, but in this case, I would go for the better readability with the `+`, since I don't expect any of the lists to exceed 10 elements and this is a code generator which is only run once every boost release (at most).

tools/boost_names.py
dcbaker: interesting, I could have sworn that it returned and interable. Maybe it did in python2? :shrug:

PR #7050, opened 2020-04-28
CI: pin pylint

jon-turney:
Pin pylint version to workaround https://github.com/PyCQA/pylint/issues/3524

### DIFF
--- a/.github/workflows/lint_mypy.yml
+++ b/.github/workflows/lint_mypy.yml
@@ -19,7 +19,8 @@ jobs:
     - uses: actions/setup-python@v1
       with:
         python-version: '3.x'
-    - run: python -m pip install pylint
+    # pylint version constraint can be removed when https://github.com/PyCQA/pylint/issues/3524 is resolved
+    - run: python -m pip install pylint==2.4.4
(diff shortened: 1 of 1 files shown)


mensinda:
Alternatively, passing `-j1` also works and more threads won't really help anyway on the CI.

jpakkane:
There should be a comment explaining the version requirement and the criterion for removing it.

.github/workflows/lint_mypy.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     - uses: actions/setup-python@v1
       with:
         python-version: '3.x'
-    - run: python -m pip install pylint
+    - run: python -m pip install pylint==2.4.4
```
jameshilliard, line 22: Probably better to just use `pylint!=2.5.0` since upstream is aware of this bug and will likely fix it next release.

jon-turney:
> There should be a comment explaining the version requirement and the criterion for removing it.

Done

dcbaker:
We can change things up later if we want to, but for now lets turn the CI green again.

PR #7052, opened 2020-04-28
default_options as dictionaries

dcbaker:
This allows using dictionaries for default_options and override_options:

```meson
project(
  'myproject',
  default_options : {'cpp_std' : 'c++11'}
)
```


mesonbuild/build.py
```diff
@@ -462,10 +462,13 @@ def process_kwargs_base(self, kwargs):
                    continue
             self.option_overrides_base[k] = v
 
-    def parse_overrides(self, kwargs) -> dict:
+    def parse_overrides(self, kwargs: T.Dict[str, object]) -> T.Dict[str, str]:
+        overrides = kwargs.get('override_options', [])
+        if isinstance(overrides, dict):
+            return overrides
```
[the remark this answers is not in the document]
xclaesse: Side tracking: this function is basically a duplicate of create_options_dict(), maybe it can be removed?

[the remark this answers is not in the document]
dcbaker: The FeatureNew is handled in the interpreter, which I fell like is the right place to handle that. see `func_build_target`

dcbaker:
I made the cleanups the @xclaesse suggested, by merging the function in build with the one in coredata. I also created a private helper to handle the FeatureNew/FeatureDeprecated warnings. I've added a commit to deprecate using an array, because it isn't nearly as intuitive of a syntax, and that way maybe by 1.0 we can remove it.

mesonbuild/interpreter.py
```diff
@@ -2705,6 +2704,13 @@ def func_gettext(self, nodes, args, kwargs):
     def func_option(self, nodes, args, kwargs):
         raise InterpreterException('Tried to call option() in build description file. All options must be in the option file.')
 
+    def _handle_default_options(self, options: T.Union[str, T.List[str], T.Dict[str, str]]) -> 'T.OrderedDict[str, str]':
+        if isinstance(options, dict):
+            FeatureNew('Dictionary argument to default_options', '0.55.0').use(self.subproject)
+        else:
+            FeatureDeprecated('Array argument to default_options, use dict instead.', '0.50.0').use(self.subproject)
```
xclaesse, line 2711: I don't think we should deprecate, I don't see the problem in supporting both forms. No need to annoy users with deprecations warnings for no reason.

dcbaker: Having two ways to do something in the long run just means more code to maintain and more code to break. The deprecation warning wont trigger until you set the `meson_version : >=0.55.0` anyway. But I could drop it I guess.

xclaesse: IMHO, small and widely used things like this deprecation would be more burden than maintaining those 10 lines of code in Meson. I also find the list of strings syntax more compact and easier to write, so there is no clear benefit to use dict.

jpakkane: This is a long standing and much used feature that would trigger many, many deprecation warnings. Since the implementation is so simple, I guess we can keep this working, at least for the moment. We can deprecate it later if needed.

jpakkane:
:crossed_swords: conflicts.

dcbaker:
rebased and dropped the deprecation patch as well as a patch that was superseded by another to do the same thing.

jpakkane:
That's a lot of CI red already.

docs/markdown/Reference-manual.md
```diff
@@ -443,8 +443,9 @@ arguments:
   that override those set in the subproject's `meson_options.txt`
   (like `default_options` in [`project()`](#project), they only have
   effect when Meson is run for the first time, and command line
-  arguments override any default options in build files)
-- `fallback`: specifies a subproject fallback to use in case the
+  arguments override any default options in build files). *(since 0.56.0)*
+  a dictionary of string: string mappings is also accepted.
```
xclaesse, line 447: nitpicking: I would remove the space `string:string`, maybe it's just me but I was reading `string: string` as a colon in the sentence.

docs/markdown/Reference-manual.md
```diff
@@ -655,10 +656,10 @@ be passed to [shared and static libraries](#library).
   target by overriding the default. By default on Windows this is
   `exe` and on other platforms it is omitted. Set this to `[]`, or omit
   the keyword argument for the default behaviour.
-- `override_options` *(since 0.40.0)*: takes an array of strings in the same format as
-  `project`'s `default_options` overriding the values of these options
-  for this target only.
-- `gnu_symbol_visibility` *(since 0.48.0)*: specifies how symbols should be exported, see
+- `override_options` takes an array of strings or dictionary in the same
+  format as `project`'s `default_options` overriding the values of these
+  options for this target only, since 0.40.0. Dictionary new in 0.56.0
```
xclaesse, line 661: I find the "since" version a bit confusing. What about keeping the same text as before, and just appending `*Since 0.56.0* can be a dictionary too.`? I think we append new behaviors in the doc in many other places like that.

Or reuse the phrasing you did below: `or, since 0.56.0, a dictionary of strings`

docs/markdown/Reference-manual.md
```diff
@@ -1579,11 +1580,13 @@ arguments:
...
    (like `default_options` in `project`, they only have effect when
    Meson is run for the first time, and command line arguments override
-   any default options in build files). *(since 0.54.0)*: `default_library`
-   built-in option can also be overridden.
- - `version`: works just like the same as in `dependency`.
-   It specifies what version the subproject should be, as an example `>=1.0.1`
- - `required` *(since 0.48.0)*: By default, `required` is `true` and
+   any default options in build files).
+   *Since 0.54.0* `default_library` built-in option can also be overridden.
+   *Since 0.56.0* a Dictionary may be used instead of an array
```
xclaesse, line 1585: here you have a 3rd way of adding dictionary documentation. Would be nice to harmonize across all default_options docs. FWIW, this is the format I like the more.

dcbaker:
I still want to fix this, but it really needs my build.py refactors to land first to be feasable, and this needs to be completely rewritteng.

PR #7054, opened 2020-04-28
Fix incremental debug builds in VS

GustavoLCR:
Fixes `Warning LNK4075: Ignoring '/INCREMENTAL' due to '/RELEASE' specification`

### DIFF
--- a/mesonbuild/backend/vs2010backend.py
+++ b/mesonbuild/backend/vs2010backend.py
@@ -1192,7 +1192,7 @@ class Vs2010Backend(backends.Backend):
         # /nologo
         ET.SubElement(link, 'SuppressStartupBanner').text = 'true'
         # /release
-        if (self.buildtype not in ['debug', 'debugoptimized']):
+        if not self.environment.coredata.get_builtin_option('debug'):
             ET.SubElement(link, 'SetChecksum').text = 'true'
 
         meson_file_group = ET.SubElement(root, 'ItemGroup')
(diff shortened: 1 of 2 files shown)


mesonbuild/backend/vs2010backend.py
```diff
@@ -1192,7 +1192,8 @@ def gen_vcxproj(self, target, ofname, guid):
         # /nologo
         ET.SubElement(link, 'SuppressStartupBanner').text = 'true'
         # /release
-        ET.SubElement(link, 'SetChecksum').text = 'true'
+        if (self.buildtype not in ['debug', 'debugoptimized']):
+            ET.SubElement(link, 'SetChecksum').text = 'true'
```
jpakkane, line 1196: Is this "debug" checking for debug info or for debug CRT?

If the former, this should check for ` debug` and `b_vscrt` if the latter.

GustavoLCR: its checking for debug info

jpakkane: In that case the code should check for `self.environment.coredata.get_builtin_option('debug')` instead of the build type.

jpakkane:
Test failure seems relevant:

```
>               self.assertTrue(pe.verify_checksum(), msg=msg)
E               AssertionError: False is not true : PE file: 'D:\\a\\1\\s\\tmpmc0atbfx\\mycpplib.dll', compiler: 'msvc', linker: 'link'
```

jpakkane:
Thanks.

PR #7061, opened 2020-04-29
Allow building with b_coverage set to true when clang is being used r…

georgev93:
…egardless of linker selection.


georgev93:
I was building a project with the following configuration:

(x86_64 Windows)
C compiler: clang
C++ compiler: clang++
Linker: link.exe
b_coverage: true

The meson setup failed with the Error: Linker link doesn't implement coverage data generation. My workflow involves using llvm-cov with the .gcda and .gcno files, so the linker doesn't need to support coverage data generation.

georgev93:
The failure seems unrelated. Any thoughts, @jpakkane?

PR #7068, opened 2020-05-01
docs/Users: add Vala Language Server

Prince781:
See https://github.com/benwaffle/vala-language-server


mensinda:
Thanks, please put `[skip ci]` for doc only changes into the commit next time.

PR #7072, opened 2020-05-03
Allow overriding g-ir-scanner and g-ir-compiler binaries.

jameshilliard:
This is useful when one needs to force meson to use wrappers for cross compilation.

### DIFF
--- a/mesonbuild/modules/gnome.py
+++ b/mesonbuild/modules/gnome.py
@@ -406,11 +406,19 @@ class GnomeModule(ExtensionModule):
             kwargs = {'native': True, 'required': True}
             holder = self.interpreter.func_dependency(state.current_node, ['gobject-introspection-1.0'], kwargs)
             self.gir_dep = holder.held_object
-            if self.gir_dep.type_name == 'pkgconfig':
+            giscanner = state.environment.lookup_binary_entry(MachineChoice.HOST, 'g-ir-scanner')
+            if giscanner is not None:
+                self.giscanner = ExternalProgram.from_entry('g-ir-scanner', giscanner)
+            elif self.gir_dep.type_name == 'pkgconfig':
                 self.giscanner = ExternalProgram('g_ir_scanner', self.gir_dep.get_pkgconfig_variable('g_ir_scanner', {}))
-                self.gicompiler = ExternalProgram('g_ir_compiler', self.gir_dep.get_pkgconfig_variable('g_ir_compiler', {}))
             else:
                 self.giscanner = self.interpreter.find_program_impl('g-ir-scanner')
+            gicompiler = state.environment.lookup_binary_entry(MachineChoice.HOST, 'g-ir-compiler')
+            if gicompiler is not None:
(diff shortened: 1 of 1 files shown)


jpakkane:
Who knows the most about this? @xclaesse?

xclaesse:
Hmm, there are 4 places that tool can come from:
- Overridden by gobject-introspection as subproject
- Path set by user in cross/native file
- Path set in pkgconfig file
- In system $PATH

However, since a pkgconfig file is required and always contains a path to those tools, I'm not sure to understand use cases for having it in cross/native file and lookup in system PATH. Why do you need a wrapper?

jameshilliard:
> Why do you need a wrapper?

You need a qemu wrapper to cross compile with these tools.

mesonbuild/modules/gnome.py
```diff
@@ -406,12 +406,22 @@ def _get_gir_dep(self, state):
             kwargs = {'native': True, 'required': True}
             holder = self.interpreter.func_dependency(state.current_node, ['gobject-introspection-1.0'], kwargs)
             self.gir_dep = holder.held_object
-            if self.gir_dep.type_name == 'pkgconfig':
-                self.giscanner = ExternalProgram('g_ir_scanner', self.gir_dep.get_pkgconfig_variable('g_ir_scanner', {}))
-                self.gicompiler = ExternalProgram('g_ir_compiler', self.gir_dep.get_pkgconfig_variable('g_ir_compiler', {}))
+            giscanner = state.environment.lookup_binary_entry(MachineChoice.HOST, 'g-ir-scanner')
+            if giscanner is not None:
+                self.giscanner = ExternalProgram.from_bin_list(state.environment, MachineChoice.HOST, 'g-ir-scanner')
```
xclaesse, line 411: This does the lookup twice, better do `ExternalProgram.from_entry('g-ir-scanner', giscanner)` here, right?

mesonbuild/modules/gnome.py
```diff
@@ -406,12 +406,22 @@ def _get_gir_dep(self, state):
...
             else:
-                self.giscanner = self.interpreter.find_program_impl('g-ir-scanner')
-                self.gicompiler = self.interpreter.find_program_impl('g-ir-compiler')
+                if self.gir_dep.type_name == 'pkgconfig':
```
xclaesse, line 413: nitpicking: we can save an indentation level with `elif`

mesonbuild/modules/gnome.py
jameshilliard: Hmm, not sure but I used [this](https://github.com/mesonbuild/meson/blob/c53b6379597be5961b4e69e7f187608452874e4c/mesonbuild/environment.py#L574-L577) as an example.

mesonbuild/modules/gnome.py
jameshilliard: changed to use `elif`

mesonbuild/modules/gnome.py
jameshilliard: Updated with your suggestion.

xclaesse:
OK, fine with me.

PR #7075, opened 2020-05-04
Let .pc files provide rpaths.

dankegel:
Marks injected buildtime rpaths with the suffix /./., and then removes only those when installing.

### DIFF
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -59,12 +59,13 @@ class InstallData:
         self.mesonintrospect = mesonintrospect
 
 class TargetInstallData:
-    def __init__(self, fname, outdir, aliases, strip, install_name_mappings, install_rpath, install_mode, optional=False):
+    def __init__(self, fname, outdir, aliases, strip, install_name_mappings, build_rpath, install_rpath, install_mode, optional=False):
         self.fname = fname
         self.outdir = outdir
         self.aliases = aliases
         self.strip = strip
         self.install_name_mappings = install_name_mappings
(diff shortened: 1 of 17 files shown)


dankegel:
@michaelkuhn, I also tested by verifying that this patch rescues spack's gtkplus build when dropped in as a replacement for spack's existing kludge var/spack/repos/builtin/packages/meson/rpath-0.54.patch.

dankegel:
This is still a WIP.  
e.g. it still needs to remove rpaths added by the user's build_rpath, not just the ones synthesized by meson itself.

jpakkane:
Perhaps it would be better if the function that determines rpaths would return a list of strings to be removed (that is, those added by Meson itself for build dir usage) which would then be passed to the rpath fixer script somehow.

dankegel:
Probably, thanks.

dankegel:
It would clearly be preferable to avoid the /./. mark in the actual RPATH.

Instead of marking rpaths in build_rpath_args, I could hand them to the caller (generate_link) which would save them in target; then generate_target_install would save them in InstallTargetData,
and install_targets() could pass them to fix_rpath()

I see three options for how to get the rpaths out of build_rpath_args:
- pass it a list to add to as a side effect
- have it return the pair (args, build_rpaths)
- refactor it into two functions, gather_rpaths() and build_rpath_args()

Maybe I'll try one of those approaches.  Comments welcome.

dankegel:
Closing in favor of https://github.com/mesonbuild/meson/pull/7103

PR #7079, opened 2020-05-05
dependency: Add dep.get_include_paths() function.

drmoose:
When dependencies are installed in odd or nonstandard locations,
external scripts such as those used in `custom_target` or `run_command`
can break.

A concrete example is [pybind's mkdoc.py][1], which uses
[clang.cindex][2] to scan a codebase and generate a header file
containing python docstrings for a C++ project. This takes similar
arguments to the clang compiler, and can fail with a fatal error if it
encounters an `#include<>` it cannot resolve.

This is currently achievable by adding a separate python script, but
that script ends up reimplementing the entire meson dependency
infrastructure, since it requires searching cmake, pkg-config and the
subprojects directory to be reliable across environments.

This PR adds a `dep.get_include_paths()` to dependency objects, providing
a list of folders for such tools to search for headers. This list can be
added to the argv passed to functions such as `run_command` and
`custom_target`.

  [1]: https://github.com/pybind/pybind11/blob/master/tools/mkdoc.py
  [2]: https://github.com/llvm-mirror/clang/blob/master/bindings/python/clang/cindex.py


dcbaker:
I'm not super thrilled about this approach, we already have a method to get a dependency with just include directories. I wonder if it would make more sense to have a way to get an `include_directories` back from a dependency, and then allow coverting include directories into strings for custom_targets like build_targets.

@jpakkane, what do you think?

drmoose:
My first thought was to just provide a way to convert dependencies to arrays generally, where the array would just be the collection of compiler flags, so that my `custom_target` could be something like `custom_target('whatever', command: ['mkdoc.py'] + dep.partial_dependency(include: true) + sources)`. Honestly the only reason I didn't go that way with this PR to begin with was the line in contributing.md about preferring specific solutions to general frameworks.

drmoose:
I've now written a workaround for this problem that involves unpickling `meson-private/build.dat` so I can get the information I need, which is an unconscionably hideous solution.

I've updated the PR to address merge conflicts. Please let me know if I need to rewrite it using a different approach, or otherwise how I can address this problem.

PR #7083, opened 2020-05-05
interpreter: Don't assign duplication and new feature warning to the …

### LINKED ISSUE #7080: Deprecation warnings are reversed
**Describe the bug**
When I use `python3` module with min_version where it was not deprecated yet, I get deprecation warning and a “feature too new” warning:

```
DEPRECATION: Project targeting '>= 0.47' but tried to use feature deprecated since '0.48.0': python3 module
Build targets in project: 0
WARNING: Project specifies a minimum meson_version '>= 0.47' but uses features which were added in newer versions:
 * 0.48.0: {'python3 module'}
WARNING: Deprecated features used:
 * 0.48.0: {'python3 module'}
```

When I bump the min_version to 0.48 both warnings disappear.

**To Reproduce**
```
project('almanah', 'c', meson_version: '>= 0.47')

python3 = import('python3')
python = python3.find_python().path()
```

**Expected behavior**
* Do not show deprecation warning with min_version < 48
* Show deprecation warning with min_version ≥ 48
* Never show “feature too new” warning.

**system parameters**
* native build
* NixOS
* Python 3.7.7
* Meson 0.54.1

dcbaker:
…same variable

Currently The Deprecated and New features checkers share an attribute
through a base class that should be per class. We need to duplicate this
and move it into each of the sublcasses

Fixes #7080

### DIFF
--- a/mesonbuild/interpreterbase.py
+++ b/mesonbuild/interpreterbase.py
@@ -18,6 +18,7 @@
 from . import mparser, mesonlib, mlog
 from . import environment, dependencies
 
+import abc
 import os, copy, re
 import collections.abc
 from functools import wraps
@@ -212,7 +213,7 @@ class permittedKwargs:
             return f(*wrapped_args, **wrapped_kwargs)
         return wrapped
 
-class FeatureCheckBase:
+class FeatureCheckBase(metaclass=abc.ABCMeta):
     "Base class for feature version checks"
 
     # In python 3.6 we can just forward declare this, but in 3.5 we can't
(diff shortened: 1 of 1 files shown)


dcbaker:
I found another, related bug that should probably be fixed in 0.54.2 as well... deprecation warnings don't work

mensinda:
Should `FeatureDeprecated` even have a version check? Wouldn't it make more sense to print the deprecation message unconditionally?

dcbaker:
I don't think so, I get a lot of PRs to remove "deprecated" features that are the only way (or right way) to do something in an older version of meson I need to support.

mensinda:
Fair point, but then this logic should also apply to all `mlog.deprecation` messages (which are shown unconditionally).

dcbaker:
We probably should replace a fair number of `mlog.deprecation` with `FeatureDeprecation` anyway.

Issue #7082, opened 2020-05-05
Should meson specifiy -release for DLang projects with the `plain` buildtype?


Cogitri:
Right now meson doesn't specify `-release` on the compiler command line when compiling with the `plain` buildtype. This is problematic, since the `-release` flag doesn't do unwanted optimizations distros want to avoid, it disables asserts, quoting the LDC docs: `-release | Disables asserts, invariants, contracts and boundscheck` [0]. Since distros usually use the `plain` buildtype, but don't want debug asserts, contracts (input validation for functions, also via asserts) etc. in their binaries it'd make sense to specify `-release` even with the `plain` buildtype IMHO.

dcbaker:
It sounds like `-release` should be tied t the badly named `b_ndebug` flag then.

Cogitri:
Ah yes, that's probably the right course of action.

PR #7084, opened 2020-05-05
n_debug support for D lang

dcbaker:
This makes two changes, first is that it generalizes n_debug, which previously always set `-DNDEBUG`, which of course doesn't make sense outside of a few languages. Instead this is now set in the compiler class with the `get_disable_assert_args` method. The second is to turn on n_debug for D language.

For D this means setting the `-release` flag (or similar for ldc2 and gdc). This turns off assertions, contracts, and array bounds checking in some cases.

Fixes: #7082


jpakkane:
I'd really like to hear from D users if this behaviour makes sense for them conceptually. I guess maybe @ximion would know?

denizzzka:
> I'd really like to hear from D users if this behaviour makes sense for them conceptually.

I use D and I agreed with this changes

Also maybe it makes sense (to avoid compilers defaults changes in future) add debug options:

`-debug` (Compile in debug code) for DMD and
`-d-debug` for LDC (can also be specified without options, in which case it enables all debug checks (i.e. asserts, boundschecks, contracts and invariants) as well as acting as -d-debug=1.)

denizzzka:
ldc2 also supports `minsize` option by passing `--Oz` or `-Oz`
I am daily use it from cross file without problems.

jpakkane:
Merging then. Thanks.

PR #7088, opened 2020-05-06
gnome.generate_gir: Fix missing include directories

xclaesse:
This revert a part of #7020 because it was using gir_inc_dirs
before it is set. Properly fix typelib_includes instead that was working
only when g-i is a pkgconfig dependency.


jon-turney:
Thanks.

This fixes the regression in the corpus test https://travis-ci.org/github/jon-turney/meson-corpus-test/builds/684034686

Is it possible to add some kind of test to cover this?

xclaesse:
@jon-turney Added a unit test for this specific case, grilo was passing a generated gir in `includes` kwarg.

jpakkane:
Test failure seems relevant:

```
Failed test during install: 'test cases/frameworks/7 gnome'
Reason: 
Extra file usr/share/gir-1.0/MesonDep3-1.0.gir found.
Extra file usr/lib/libdep3lib.so found.
Extra file usr/lib/girepository-1.0/MesonDep3-1.0.typelib found.
```

xclaesse:
@jpakkane fixed that now, I think remaining failing test is unrelated, I see it failing on master too.

PR #7099, opened 2020-05-07
ConfigToolDependency: Don't fallback to system tool when cross compiling

xclaesse:
The system tool is always the wrong thing to use and cause hard to debug
issues when trying to link system libraries with cross built binaries.

The ExternalDependency base class already had a method to deal with
this, used by PkgConfigDependency and QtBaseDependency, so it should
make things more consistent.

### DIFF
--- a/docs/markdown/Dependencies.md
+++ b/docs/markdown/Dependencies.md
@@ -242,6 +242,9 @@ libgcrypt_dep = dependency('libgcrypt', version: '>= 1.8')
 gpgme_dep = dependency('gpgme', version: '>= 1.0')
 ```
 
+*Since 0.55.0* Meson won't search $PATH any more for a config tool binary when
+cross compiling if the config tool did not have an entry in the cross file.
+
 ## AppleFrameworks
 
 Use the `modules` keyword to list frameworks required, e.g.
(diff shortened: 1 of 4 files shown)


dcbaker:
I have a PR for this already, https://github.com/mesonbuild/meson/pull/6547

We need to fix the CI because it only passes currently because of this behavior.

dcbaker:
You're also changing behavior, so we at the very least need a snippet for this.

xclaesse:
@dcbaker oh haven't noticed your PR, sorry. I now copied your doc into this PR, and skipped the failing unit test when cross compiling. I think it's better to use the base class search_tool() method for consistency, even if it does pretty much the same thing. Is that's OK with you?

Ericson2314:
Yeah this one deletes more code so I guess I prefer it :D.

xclaesse:
As far as I can tell, failing unit test is unrelated, happens on master too.

dcbaker:
@xclaesse, I agree that yours is better, I don't think `search_tool` existed when I wrote mine.

dcbaker:
I just merged code that should fix the failing tests.

xclaesse:
Thanks :)

PR #7103, opened 2020-05-08
Let .pc files and LDFLAGS provide rpaths.

### LINKED ISSUE #2567: Please provide a user-configurable, global way to specify RPATH/RUNPATH for the whole project.
Currently meson does not provide a way to specify a custom rpath, it removes rpath in install step even if it was specified via LDFLAGS.
https://github.com/NixOS/nixpkgs/pull/28444#issuecomment-324033323

Please note: I'm aware about {build,install}_rpath, these are project or target options.
I'm talking about something like cmake's CMAKE_INSTALL_RPATH which can be easily passed to cmake while configuring. Even autohell behaves fine with such use case.
What I'm asking for is an option like --meson-install-rpath (like --prefix) which would allow users (project consumers, not developers) to build and install packages into a custom prefix _without_ messing with meson.build .

This is needed when you want to install something into an arbitrary prefix to avoid overwriting system stuff, eg. you want to build something with new shiny features straight from git, but you don't want to break your OS.

Please consider adding such a feature.
 
PS: LD_RUN_PATH et al is not an option, the rest of the system should not know anything "prefixed" installs.

### LINKED ISSUE #4027: Meson strips rpaths set via pkgconfig on install
If a .pc file specifies -Wl,-rpath, it probably has a good reason to.
Thus it is surprising that meson strips those rpaths on install.
The workaround suggested by https://github.com/mesonbuild/meson/issues/1411
is awkward as it requires knowing what rpath the library needed.
[Edit: and it requires editing the meson config, which is awkward when
you're just trying to package, say, gstreamer to install to /opt/foo.
What's the point of meson's --prefix option if it forbids rpaths pointing into that prefix?]

Since rpath in .pc files are frowned upon in linux distributions, this is likely to only
affect third-party software that installs libraries to /opt.

If meson is making a policy statement that .pc files must not contain rpath options, it should say so clearly, and possibly abort when such .pc files are encountered

dankegel:
Like https://github.com/mesonbuild/meson/pull/7075, but instead of marking the injected rpaths, remembers them.

Fixes #4027

Oh, and what the heck, fix #2567 en passant.


test cases/unit/40 external, internal library rpath/built library/meson.build
```diff
@@ -15,7 +15,9 @@ l = shared_library('bar_built', 'bar.c',
                    install: true,
                    dependencies : [foo_system_dep, faa_pkg_dep])
 
-if host_machine.system() == 'darwin'
-  e = executable('prog', 'prog.c', link_with: l, install: true)
+if host_machine.system() == 'darwin' or host_machine.system() == 'linux'
+  e = executable('prog', 'prog.c', link_with: l, install: true,
+        install_rpath: '$ORIGIN/..' / get_option('libdir'),
```
dankegel, line 20: $ORIGIN is invalid on mac, this bit needs adjusting.

eli-schwartz:
Is this just about pkg-config files, or is it intended to also preserve rpath which is set by e.g. LDFLAGS? The latter is a concern for yocto/buildroot/nixos/guix etc. as discussed in #2567.

dankegel:
Letting .pc files provide rpaths implies preserving rpaths set via LDFLAGS, I believe, so this is intended to also solve the yocto/buildroot/etc issue (which is how I got into this).

eli-schwartz:
Sounds great... what do you think about using a slightly less confusing commit message, then?

dankegel:
I'd feel great about it, but I'd want to add a unit test for the particular use case in mind.

Care to suggest one?

eli-schwartz:
Seems like implementing essentially the same as test_usage_pkgconfig_prefixes, but by setting val1's project to build with env1['LDFLAGS'], might do the trick?

dankegel:
You mean instead of using the install_rpath argument to shared_library?

dankegel:
I would prefer to keep this pull request simple, so would like to handle the global LDFLAGS rpath feature later.

This pull request is ready for review.

jpakkane:
From what I can tell, the `rpath_dirs_to_update` argument to `build_rpath_args` is only used as an out-parameter. This is strongly discouraged in general. We don't use out parameters at all in Meson, because they make code hard to reason about.

The better approach for this is to change the return value of the method to a tuple `(rpath_args, rpath_dirs_added_to_run_from_build_dir)`, or words to that effect.

dankegel:
Roger, wilco. I used the out parameter out of sheer laziness in the proof-of-concept.

jpakkane:
Since this is a behavioural change, it should get a release note snippet under `docs/markdown/snippets`.

dankegel:
All issues resolved.

nickbroon:
Is there a work-around for older versions of meson?

I have project that makes use of external third-party libraries (via dependency() ) that provide a rpath in their .pc pkgconfig files. meson strips these rpaths, when installing my project and thus makes it unusable.
Is there some meson snippet that can extract the rpath from the dependency gotten from the .pc file, that can then be passed to install_rpath: and thus avoid the meson broken stripping behaviour?

eli-schwartz:
Why not just bump the minimum meson version?

nickbroon:
I'm targeting a specific distribution version that only has an older version of meson available. I'm reluctant to have to backport and support a newer version of meson. (as this might impact other packages in distribution build).

eli-schwartz:
If they have an older version of meson and won't upgrade, that's because they have a stable release policy. If so, they won't accept new versions of your software either. So it really doesn't matter if you bump the minimum meson version.

People building outside of the distribution repository infrastructure can locally download the latest version of meson e.g. using `pip install meson`.

...You say the code currently doesn't work at all on older versions of meson. So no one is currently using it on these ~~stale~~ old distros. Bumping the minimum meson version doesn't prevent anyone from doing something they used to be able to do. All it does is correctly communicate the pre-existing constraints. It's definitely the correct thing to do if your code depends on this PR.

nickbroon:
I'm not submitting my software to the stable distribution ,but my software does specifically targeted that stable distribution, and has be able to build on it. (That is the software forms part of overlay for a stable distribution with older version of meson).

As there appears to be no work around for this meson bug, I guess I'll have to include a newer version of meson as part of the overlay build collection.

dankegel:
That's what I had to do.

mesonbuild/backend/backends.py
```diff
@@ -443,6 +444,21 @@ def _libdir_is_system(libdir, compilers, env):
...
+        args = []
+        # FIXME: is there a better way?
+        for lang in ['c', 'cpp']:
+            try:
+                args.extend(self.environment.coredata.get_external_link_args(target.for_machine, lang))
+            except Exception:
+                pass
+        for arg in args:
+            if arg.startswith('-Wl,-rpath='):
+                for dir in arg.replace('-Wl,-rpath=','').split(':'):
```
jameshilliard, line 458: This match pattern is insufficient, I've followed up with a fix in #7472.

PR #7109, opened 2020-05-08
vs: add static_from_buildtype to b_vscrt

peterh:
I compile my code with the static CRT, to simplify deployment.

I was using `b_vscrt=mt`. I recently tried to add `-D_DEBUG` to my debug builds, but my code failed to link because `mt` does not export all the required symbols. `from_buildtype` is almost what I wanted, except from_buildtype always links with the DLL CRT.

It's not clear if `static_from_buildtype` is the correct way to go about this, but it's a starting point for discussion.

(Note that I snuck in a fix for backend=vs when using `from_buildtype` and `buildtype=debugoptimized` -- meson previously generated MultiThreaded (non-dll), which did not match ninja or the documentation)

### DIFF
--- a/docs/markdown/Builtin-options.md
+++ b/docs/markdown/Builtin-options.md
@@ -126,23 +126,23 @@ platforms or with all compilers:
 | b_sanitize  | none          | see below               | Code sanitizer to use |
 | b_staticpic | true          | true, false             | Build static libraries as position independent |
 | b_pie       | false         | true, false             | Build position-independent executables (since 0.49.0)|
-| b_vscrt     | from_buildtype| none, md, mdd, mt, mtd, from_buildtype | VS runtime library to use (since 0.48.0) |
+| b_vscrt     | from_buildtype| none, md, mdd, mt, mtd, from_buildtype, static_from_buildtype | VS runtime library to use (since 0.48.0) (static_from_buildtype since 0.56.0) |
 
 The value of `b_sanitize` can be one of: `none`, `address`, `thread`,
 `undefined`, `memory`, `address,undefined`.
 
 <a name="b_vscrt-from_buildtype"></a>
-The default value of `b_vscrt` is `from_buildtype`. In that case, the following
-table is used internally to pick the CRT compiler arguments based on the value
-of the `buildtype` option:
+The default value of `b_vscrt` is `from_buildtype`. The following table is used
+internally to pick the CRT compiler arguments for `from_buildtype` or
+`static_from_buildtype` *(since 0.56)* based on the value of the `buildtype` option:
 
-| buildtype      | Visual Studio CRT |
-| --------       | ----------------- |
-| debug          | `/MDd`            |
-| debugoptimized | `/MD`             |
-| release        | `/MD`             |
-| minsize        | `/MD`             |
-| custom         | error!            |
+| buildtype      | from_buildtype | static_from_buildtype |
+| --------       | -------------- | --------------------- |
+| debug          | `/MDd`         | `/MTd`                |
+| debugoptimized | `/MD`          | `/MT`                 |
+| release        | `/MD`          | `/MT`                 |
+| minsize        | `/MD`          | `/MT`                 |
+| custom         | error!         | error!                |
 
 ### Notes about Apple Bitcode support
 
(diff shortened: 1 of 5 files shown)


nirbheek:
> (Note that I snuck in a fix for backend=vs when using `from_buildtype` and `buildtype=debugoptimized` -- meson previously generated MultiThreaded (non-dll), which did not match ninja or the documentation)

Could you split that out so we can get it in immediately? That seems obviously correct :)

peterh:
> Could you split that out so we can get it in immediately? That seems obviously correct :)

Done: #7146 (and rebased this pull request on top of that one)

peterh:
Rebased and force-pushed to avoid the merge conflict. Is there anything else I can do to help this along?

docs/markdown/Builtin-options.md
```diff
@@ -123,23 +123,23 @@ platforms or with all compilers:
 | b_sanitize  | none          | see below               | Code sanitizer to use |
 | b_staticpic | true          | true, false             | Build static libraries as position independent |
 | b_pie       | false         | true, false             | Build position-independent executables (since 0.49.0)|
-| b_vscrt     | from_buildtype| none, md, mdd, mt, mtd, from_buildtype | VS runtime library to use (since 0.48.0) |
+| b_vscrt     | from_buildtype| none, md, mdd, mt, mtd, from_buildtype, static_from_buildtype | VS runtime library to use (since 0.48.0) (static_from_buildtype since 0.55.0) |
```
nirbheek, line 126: s/0.55/0.56

docs/markdown/Builtin-options.md
```diff
@@ -123,23 +123,23 @@ platforms or with all compilers:
...
-| buildtype      | Visual Studio CRT |
-| --------       | ----------------- |
-| debug          | `/MDd`            |
-| debugoptimized | `/MD`             |
-| release        | `/MD`             |
-| minsize        | `/MD`             |
-| custom         | error!            |
+The default value of `b_vscrt` is `from_buildtype`. The following table is used
+internally to pick the CRT compiler arguments for `from_buildtype` or
+`static_from_buildtype` based on the value of the `buildtype` option:
```
nirbheek, line 134: This should also mention that it's since 0.56 (yes, repeating is good)

nirbheek (commented):
I think this also needs changes in:

`mesonbuild/dependencies/ui.py` -- pick the correct qt mt libs (I think this requires someone to build Qt with /MT and /MTd to test, so maybe it's enough to just do mlog.warning when b_vscrt is mt/mtd)

`mesonbuild/dependencies/boost.py` -- pick the correct boost mt libs (this is already picking it based on a kwarg, so I think we need to change behavior there)

LGTM otherwise. Needs @jpakkane approval for the new API.

peterh:
> I think this also needs changes in:
> 
> `mesonbuild/dependencies/ui.py` -- pick the correct qt mt libs (I think this requires someone to build Qt with /MT and /MTd to test, so maybe it's enough to just do mlog.warning when b_vscrt is mt/mtd)

I installed (but didn't build anything with) Qt to check, and they only have d and not-d versions of their libraries. I'm not familiar at all with Qt so I don't know which side (static or not-static) should be warned about, if any. The current code checks for both mdd (dll) and mtd (static), so maybe the pre-built Qt libraries work either way and no warning is necessary?

> `mesonbuild/dependencies/boost.py` -- pick the correct boost mt libs (this is already picking it based on a kwarg, so I think we need to change behavior there)

This is already picking based on the return value of `get_crt_compile_args`, so that should continue to work with my patch. (I didn't change the set of possible return values from this function).

> LGTM otherwise. Needs @jpakkane approval for the new API.

Updated 0.55 to 0.56, and added 0.56 in the location requested.

Thanks for the review.

nirbheek:
> I installed (but didn't build anything with) Qt to check, and they only have d and not-d versions of their libraries.

Yes, that's exactly the problem. Qt only ships /MD and /MDd versions of their binaries, and the dependency only handles this case 
(see: `_get_modules_lib_suffix()`). You have to build Qt yourself to get /MT or /MTd, so no one has tested it yet.

Honestly don't know what we should do here. I guess we can keep it as-is. It's not a new problem.

> This is already picking based on the return value of `get_crt_compile_args`, so that should continue to work with my patch

Hmm, the code is a bit confusing but I see what you mean. Looks ok.

nirbheek:
@jpakkane code looks fine, and behaviour-wise I think this new value for b_vscrt is a good idea. What do you think?

jpakkane:
Seems reasonable. This needs a rebase, though, to rerun the test suite.

jpakkane:
Test failures are unrelated.

PR #7116, opened 2020-05-10
Fix builtin check in has_function() with GCC 10 on Windows

### LINKED ISSUE #7113: mingw: has_function() MSYS2 workaround no longer works with gcc 10
**Describe the bug**

has_function for stpcpy and posix_memalign now return true on Windows with gcc 10

My guess is that this is because the workaround here:

https://github.com/mesonbuild/meson/blob/39ca178d21a72a8f7560b657b517fb73f81b4a89/mesonbuild/compilers/mixins/clike.py#L738

no longer works because gcc 10 now supports __has_builtin.

**To Reproduce**

```meson
project('glib', 'c', 'cpp',
  version : '2.65.0',
)

cc = meson.get_compiler('c')
cc.has_function('posix_memalign', prefix : '#include <stdlib.h>')
```

```meson
The Meson build system
Version: 0.54.1
Source dir: C:/msys64/home/user/foo
Build dir: C:/msys64/home/user/foo/_build
Build type: native build
Using 'PKG_CONFIG_PATH' from environment with value: 'C:\\msys64\\mingw64\\lib\\pkgconfig;C:\\msys64\\mingw64\\share\\pkgconfig'
Using 'PKG_CONFIG_PATH' from environment with value: 'C:\\msys64\\mingw64\\lib\\pkgconfig;C:\\msys64\\mingw64\\share\\pkgconfig'
Project name: glib
Project version: 2.65.0
C compiler for the host machine: ccache cc (gcc 10.1.0 "cc (Rev1, Built by MSYS2 project) 10.1.0")
C linker for the host machine: cc ld.bfd 2.34
C++ compiler for the host machine: ccache c++ (gcc 10.1.0 "c++ (Rev1, Built by MSYS2 project) 10.1.0")
C++ linker for the host machine: c++ ld.bfd 2.34
Host machine cpu: x86_64
Checking for function "posix_memalign" : YES
Build targets in project: 0

Found ninja.EXE-1.10.0 at C:\msys64\mingw64\bin/ninja.EXE
```

**Expected behavior**

```meson
Checking for function "posix_memalign" : NO
```

lazka:
The builtin check had a special case that if a header was provided and
the function wasn't defined, it would ignore the builtin to avoid
non-functional builtins (for example __builtin_posix_memalign in MSYS2).

GCC 10 gained support for __has_builtin() which now skipps this check
and because __has_builtin(__builtin_posix_memalign) returns true the
non functional builtin is now reported as available.

To get the old behaviour back move the special case in front of the actual
availability check.

Fixes #7113

### DIFF
--- a/mesonbuild/compilers/mixins/clike.py
+++ b/mesonbuild/compilers/mixins/clike.py
@@ -727,24 +727,29 @@ class CLikeCompiler:
         # need to look for them differently. On nice compilers like clang, we
         # can just directly use the __has_builtin() macro.
         fargs['no_includes'] = '#include' not in prefix
-        fargs['__builtin_'] = '' if funcname.startswith('__builtin_') else '__builtin_'
+        is_builtin = funcname.startswith('__builtin_')
+        fargs['is_builtin'] = is_builtin
+        fargs['__builtin_'] = '' if is_builtin else '__builtin_'
         t = '''{prefix}
         int main(void) {{
+
+        /* With some toolchains (MSYS2/mingw for example) the compiler
+         * provides various builtins which are not really implemented and
+         * fall back to the stdlib where they aren't provided and fail at
+         * build/link time. In case the user provides a header, including
+         * the header didn't lead to the function being defined, and the
+         * function we are checking isn't a builtin itself we assume the
+         * builtin is not functional and we just error out. */
+        #if !{no_includes:d} && !defined({func}) && !{is_builtin:d}
+            #error "No definition for {__builtin_}{func} found in the prefix"
+        #endif
+
         #ifdef __has_builtin
             #if !__has_builtin({__builtin_}{func})
                 #error "{__builtin_}{func} not found"
             #endif
         #elif ! defined({func})
-            /* Check for {__builtin_}{func} only if no includes were added to the
-             * prefix above, which means no definition of {func} can be found.
-             * We would always check for this, but we get false positives on
-             * MSYS2 if we do. Their toolchain is broken, but we can at least
-             * give them a workaround. */
-            #if {no_includes:d}
-                {__builtin_}{func};
-            #else
-                #error "No definition for {__builtin_}{func} found in the prefix"
-            #endif
+            {__builtin_}{func};
         #endif
         return 0;
         }}'''


lazka:
Tests are failing because ninja doesn't support gcc 10 yet, see https://github.com/ninja-build/ninja/pull/1774

dcbaker:
This all looks reasonable, and turning the CI green is a plus. I wonder if we really should consider either taking builtins out of has_function, or at least adding a `allow_builtins` flag.

nirbheek:
I liked @ePirat's suggestion of letting this function be as-is and creating two new compiler checks: `has_builtin()` and `has_prototype()` or something like that.

lazka:
Thanks!

> Tests are failing because ninja doesn't support gcc 10 yet, see [ninja-build/ninja#1774](https://github.com/ninja-build/ninja/pull/1774) (we'll probably patch this in MSYS2 shortly)

We partly reverted the GCC 10 commit in MSYS2 that broke ninja, so ninja parsing GCC depfiles works again.

> This all looks reasonable, and turning the CI green is a plus

CI is still red though :/

lazka:
This broke has_function() for some cases with clang, but only because it exposed another bug:

It tries to link some code which on clang always fails to compile, I guess this wasn't noticed because only on Window is sizeof(long) != sizeof(void *):
```
error: cast from pointer to smaller type 'long' loses information
            long b = (long) a;
                     ^~~~~~~~
1 error generated.
```

I'll look into this.

mesonbuild/compilers/mixins/clike.py
```diff
@@ -727,24 +727,29 @@ def has_function(self, funcname, prefix, env, *, extra_args=None, dependencies=N
...
         fargs['no_includes'] = '#include' not in prefix
-        fargs['__builtin_'] = '' if funcname.startswith('__builtin_') else '__builtin_'
+        is_builtin = funcname.startswith('__builtin_')
+        fargs['is_builtin'] = is_builtin
+        fargs['__builtin_'] = '' if is_builtin else '__builtin_'
         t = '''{prefix}
         int main(void) {{
+
+        /* With some toolchains (MSYS2/mingw for example) the compiler
+         * provides various builtins which are not really implemented and
```
mstorsjo, line 737: Just for the record (commenting on the comment, even if it was here before) - this isn't really toolchain specific - but mingw is maybe one of the more common platforms where more functions are missing (hence the need for build system checks). The compilers provide their builtins regardless of what happens to be available in the platform headers (the compiler doesn't know what, and platform headers evolve independently from compilers).

lazka: Thanks, good to know. The previous comment I replaced here said the toolchain is broken, and I wasn't sure what to blame.. :)

PR #7117, opened 2020-05-11
boost: Try finding libraries with the matching arch (fixes #7110)

### LINKED ISSUE #7110: Boost rewrite breaks build
Consider the following meson.build file
```
project('test_link', 'cpp', default_options : ['cpp_std=c++17'])

boost_dep = dependency('boost', modules : ['system'])

sources = files(
        'main.cpp'
)

a_out = executable('a.out', sources, dependencies : [boost_dep])
```
(the file `main.cpp` can just contain `int main(){}`.)

This no longer builds.  The behavior was introduced with 08224dafcba1b694fb624553e7d84deb565aae22.

The problem is that instead of `-lboost_system`, the file `/usr/lib/libboost_system.so.1.72.0` is passed as such to the linker:
```
[2/2] c++  -o a.out 'a.out@exe/main.cpp.o' -Wl,--as-needed -Wl,--no-undefined -Wl,--start-group /usr/lib/libboost_system.so.1.72.0 -Wl,--end-group
FAILED: a.out 
c++  -o a.out 'a.out@exe/main.cpp.o' -Wl,--as-needed -Wl,--no-undefined -Wl,--start-group /usr/lib/libboost_system.so.1.72.0 -Wl,--end-group
/usr/lib/gcc/x86_64-pc-linux-gnu/10.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: error: /usr/lib/libboost_system.so.1.72.0: incompatible target
```

mensinda:
No test since this is highly system-specific. I will try to reproduce this with a Gentoo OS Comp test in the future.

### DIFF
--- a/mesonbuild/dependencies/boost.py
+++ b/mesonbuild/dependencies/boost.py
@@ -412,10 +412,19 @@ class BoostDependency(ExternalDependency):
                 break
 
     def run_check(self, inc_dirs: T.List[BoostIncludeDir], lib_dirs: T.List[Path]) -> bool:
+        mlog.debug('  - potential library dirs: {}'.format([x.as_posix() for x in lib_dirs]))
+        mlog.debug('  - potential include dirs: {}'.format([x.path.as_posix() for x in inc_dirs]))
+
         #   2. Find all boost libraries
         libs = []  # type: T.List[BoostLibraryFile]
         for i in lib_dirs:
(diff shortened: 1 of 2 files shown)


floppym:
I'll give this a spin.

floppym:
This seems to work for me.

PR #7119, opened 2020-05-12
Performance optimize ninja backend

marcelhollerbach:
This PR overall shaves off 10s from the build time in normal efl, and roughly 2 min. when building a special configuration of efl.

The details for each change are described in the commits themselfs.

### DIFF
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -2004,6 +2004,10 @@ https://gcc.gnu.org/bugzilla/show_bug.cgi?id=47485'''))
             curdir = '.'
         return compiler.get_include_args(curdir, False)
 
+    @lru_cache(maxsize=None)
+    def get_normpath_target(self, source) -> str:
+        return os.path.normpath(source)
+
     def get_custom_target_dir_include_args(self, target, compiler):
         custom_target_include_dirs = []
         for i in target.get_generated_sources():
@@ -2012,7 +2016,7 @@ https://gcc.gnu.org/bugzilla/show_bug.cgi?id=47485'''))
             # own target build dir.
             if not isinstance(i, (build.CustomTarget, build.CustomTargetIndex)):
                 continue
-            idir = os.path.normpath(self.get_target_dir(i))
+            idir = self.get_normpath_target(self.get_target_dir(i))
             if not idir:
                 idir = '.'
             if idir not in custom_target_include_dirs:
--- a/mesonbuild/compilers/compilers.py
+++ b/mesonbuild/compilers/compilers.py
@@ -138,11 +138,15 @@ def is_llvm_ir(fname):
         fname = fname.fname
     return fname.split('.')[-1] == 'll'
 
+@lru_cache(maxsize=None)
+def cached_by_name(fname):
+    suffix = fname.split('.')[-1]
+    return suffix in obj_suffixes
+
 def is_object(fname):
     if hasattr(fname, 'fname'):
         fname = fname.fname
-    suffix = fname.split('.')[-1]
-    return suffix in obj_suffixes
+    return cached_by_name(fname)
 
 def is_library(fname):
     if hasattr(fname, 'fname'):
--- a/mesonbuild/compilers/compilers.py
+++ b/mesonbuild/compilers/compilers.py
@@ -16,7 +16,6 @@ import contextlib, os.path, re, tempfile
 import collections.abc
 import itertools
 import typing as T
-from functools import lru_cache
 
 from ..linkers import (
     GnuLikeDynamicLinkerMixin, LinkerEnvVarsMixin, SolarisDynamicLinker,
@@ -466,9 +465,6 @@ class CompilerArgs(collections.abc.MutableSequence):
                  iterable: T.Optional[T.Iterable[str]] = None):
         self.compiler = compiler
         self.__container = list(iterable) if iterable is not None else []  # type: T.List[str]
-        self.__seen_args = set()
-        for arg in self.__container:
-            self.__seen_args.add(arg)
 
     @T.overload                                # noqa: F811
     def __getitem__(self, index: int) -> str:  # noqa: F811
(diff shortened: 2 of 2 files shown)


marcelhollerbach:
Mhm, it seems the azure pipeline failes undependend from that ... :/

mesonbuild/compilers/compilers.py
```diff
@@ -138,11 +139,15 @@ def is_llvm_ir(fname):
         fname = fname.fname
     return fname.split('.')[-1] == 'll'
 
+@lru_cache(maxsize=None)
+def cached_by_name(fname):
+    suffix = fname.split('.')[-1]
+    return suffix in obj_suffixes
```
dcbaker, line 145: Could you also try chainging `obj_suffixes` from a tuple to a set? that will make the lookup time log(O) instead of n(O)?

mesonbuild/compilers/compilers.py
```diff
@@ -462,9 +467,50 @@ def __init__(self, compiler: T.Union['Compiler', StaticLinker],
...
+    # Note: This function is designed to work without delete operations, as deletions are worsening the performance a lot.
+    def flush_pre_post(self):
+        pre_flush = deque()
+        pre_flush_set = set()
+        post_flush = deque()
+        post_flush_set = set()
+
+        #The two lists are here walked from the front to the back, in order to not need removals for deduplication
+        self.pre.reverse()
+        for a in self.pre:
```
dcbaker, line 485: You're not reusing the value of `self.pre` in sorted form, use `for a in sorted(self.pre)`, which is faster due to not doing an in-pace sort.

mesonbuild/compilers/compilers.py
```diff
@@ -462,9 +467,50 @@ def __init__(self, compiler: T.Union['Compiler', StaticLinker],
...
+            dedup = self._can_dedup(a)
+            if a not in pre_flush_set:
+                pre_flush.appendleft(a)
+                if dedup == 2:
+                    pre_flush_set.add(a)
+
+        self.post.reverse()
+        for a in self.post:
```
dcbaker, line 493: same here.

mesonbuild/compilers/compilers.py
```diff
@@ -462,9 +467,50 @@ def __init__(self, compiler: T.Union['Compiler', StaticLinker],
...
+        #pre and post will overwrite every element that is in the container
+        #only copy over args that are in __container but not in the post flush or pre flush set
+
+        for a in self.__container:
+            if a not in post_flush_set and a not in pre_flush_set:
+                pre_flush.append(a)
+
+        self.__container = list(pre_flush) + list(post_flush)
+        self.pre.clear()
+        self.post.clear()
```
dcbaker, line 509: IT doesn't look like these are being used outside of this function, so they can just be local variables, right?

mesonbuild/compilers/compilers.py
```diff
@@ -462,9 +467,50 @@ def __init__(self, compiler: T.Union['Compiler', StaticLinker],
...
+
+    def __iter__(self):
+        self.flush_pre_post()
+        return self.__container.__iter__();
```
dcbaker, line 513:
```suggestion
        return iter(self.__container)
```

Don't call dunder methods directly (that's what the methods that start and end with `__` are called) unless you're doing something tricky with subclassing. Use the builtin function instead.

mesonbuild/backend/ninjabackend.py
```diff
@@ -2002,6 +2002,10 @@ def get_build_dir_include_args(self, target, compiler):
             curdir = '.'
         return compiler.get_include_args(curdir, False)
 
+    @lru_cache(maxsize=None)
+    def get_normpath_target(self, source):
+        return os.path.normpath(source)
```
dcbaker, line 2007:
```suggestion
def get_normpath_target(self, source: str) -> str:
        return os.path.normpath(source)
```

Please use type annotations for the new functions if possible.

mesonbuild/compilers/compilers.py
marcelhollerbach: That does not really change anything. The tuple is small enough that it is not relevant i guess.

mesonbuild/compilers/compilers.py
marcelhollerbach: Cool, thx.

mesonbuild/compilers/compilers.py
marcelhollerbach: Done.

mesonbuild/compilers/compilers.py
marcelhollerbach: Mhm, i am not sure what you mean, pre and post are filled in __iadd__, and just flushed in here to __container.

mesonbuild/compilers/compilers.py
marcelhollerbach: Okay wait, calling reserve() is faster here than calling sorted(self.pre, reverse=True) (we are talking about 0.01s in avg., so not much)

marcelhollerbach:
Okay, i updated a few things, i left the reverse in favour of sorted(...) as it turned out to be faster...

mesonbuild/compilers/compilers.py
dcbaker: oops. I really meant to say `reversed()` and said `sorted()`. The difference is obviously going to be bigger the larger the deque is, but generally in python you want to avoid calling `.sort()` or `.reverse()` on a container unless you need to iterate over that value multiple times.

mesonbuild/compilers/compilers.py
dcbaker: Never mind then, I missed that.

mesonbuild/compilers/compilers.py
marcelhollerbach: Okay yeah, that helped. Thank you :)

marcelhollerbach:
Btw. if someone wants to toy around with performance stuff, I uploaded my playground project. Which just contains an insane big about of dep's sources and libraries, its not meant to ever be fast, but you can nicely see where the most times goes, esp. with tree-like dependency trees.

https://github.com/marcelhollerbach/meson-giant

marcelhollerbach:
@dcbaker Is there anything from your requested changes that I missed ?

marcelhollerbach:
Is there any other comment ? I would like to further optimize a few things, however, more work is building up on that ...

jon-turney:
It looks like this regresses building zrythm @ tag v0.8.200 (and maybe some other stuff in the corpus which I haven't looked at yet).  Compare:

https://travis-ci.com/github/jon-turney/meson-corpus-test/jobs/346393426
https://travis-ci.com/github/jon-turney/meson-corpus-test/jobs/344566632

bisection lands on this PR.

On a brief investigation, it looks like the position of `-I.` in the command line is altered, which causes config.h from a subproject, rather than the project, to be included, which is missing the expected define.

marcelhollerbach:
Looking now, thank you for the report :)

marcelhollerbach:
Mhm, I just probebly found the problem, can you give https://github.com/marcelhollerbach/meson/tree/fast-fix a quick check ? I cannot right now, will in a hour or so, sorry for breaking that :(

marcelhollerbach:
https://github.com/marcelhollerbach/meson/commit/32ab12ba9d4a538bd8fdfed39a22f852b027589c

^ This is fixing it. Will create a test case tomorrow...

jon-turney:
> can you give https://github.com/marcelhollerbach/meson/tree/fast-fix a quick check

Sorry, I was having a nap 💤 :)

> This is fixing it.

Confirmed that zrythm builds again with that.  Thanks!

marcelhollerbach:
Np, now its time for my nap :)

Issue #7017, opened 2020-04-22
Test suite fails when using GCC 10


DimStar77:
**Describe the bug**
GCC10 is nearing release and various distros are sarting preparation work. Meson's test suite has been identified as a component that is failing when using gcc19

```
[  873s] Running compile:
[  873s] Working directory:  /home/abuild/rpmbuild/BUILD/meson-0.54.0/b 536484cc32/meson-private/tmpsv_1caih
[  873s] Command line:  gfortran /home/abuild/rpmbuild/BUILD/meson-0.54.0/b 536484cc32/meson-private/tmpsv_1caih/testfile.f90 -pipe -cpp -E -P -D_FILE_OFFSET_BITS=64 -cpp -P -O0 
[  873s] 
[  873s] Code:
[  873s]  
[  873s]         #ifdef __has_include
[  873s]          #if !__has_include("omp.h")
[  873s]           #error "Header 'omp.h' could not be found"
[  873s]          #endif
[  873s]         #else
[  873s]          #include <omp.h>
[  873s]         #endif
[  873s] Compiler stdout:
[  873s]  
[  873s] 
[  873s] 
[  873s]         #ifdef 0
[  873s]          #if !0("omp.h")
[  873s]           #error "Header 'omp.h' could not be found"
[  873s]          #endif
[  873s]         #else
[  873s]          #include <omp.h>
[  873s]         #endif
[  873s] 
[  873s] Compiler stderr:
[  873s]  /home/abuild/rpmbuild/BUILD/meson-0.54.0/b 536484cc32/meson-private/tmpsv_1caih/testfile.f90:2:0:
[  873s] 
[  873s]     2 |         #ifdef __has_include
[  873s]       | 
[  873s] Error: "__has_include" used outside of preprocessing directive
[  873s] /home/abuild/rpmbuild/BUILD/meson-0.54.0/b 536484cc32/meson-private/tmpsv_1caih/testfile.f90:2:0: Error: missing '(' before "__has_include" operand
[  873s] /home/abuild/rpmbuild/BUILD/meson-0.54.0/b 536484cc32/meson-private/tmpsv_1caih/testfile.f90:2:0: Error: operator "__has_include" requires a header-name
[  873s] /home/abuild/rpmbuild/BUILD/meson-0.54.0/b 536484cc32/meson-private/tmpsv_1caih/testfile.f90:3:0:
[  873s] 
[  873s]     3 |          #if !__has_include("omp.h")
[  873s]       | 
[...]

PR #7120, opened 2020-05-12
Fix has_function() for clang on 64bit Windows

lazka:
has_function() tries to link an example program using the function
to see if it is available, but with clang on 64bit Windows this
example always already failed at the compile step:

error: cast from pointer to smaller type 'long' loses information
            long b = (long) a;

This is due to long!=pointer with LLP64

Change from "long" to "long long" which is min 64bit and should always
fit a pointer. While "long long" is strictly a C99 feature every
non super ancient compiler should still support it.

### DIFF
--- a/mesonbuild/compilers/mixins/clike.py
+++ b/mesonbuild/compilers/mixins/clike.py
@@ -658,7 +658,7 @@ class CLikeCompiler:
         # is not run so we don't care what the return value is.
         main = '''\nint main(void) {{
             void *a = (void*) &{func};
-            long b = (long) a;
+            long long b = (long long) a;
             return (int) b;
         }}'''
         return head, main


lazka:
Not sure if there is a better fix, and I wasn't sure what I'm allowed to include in that context so I went with __int64/_WIN64 which should be provided by the compiler.

dcbaker:
Gah, pre C99 is such a mess :(

Can we just change `long` to `long long` across the board?

lazka:
> Can we just change `long` to `long long` across the board?

For Windows I'd say yes, based on [0]

[0] https://web.archive.org/web/20060411150011/http://msdn2.microsoft.com/en-us/library/s3f49ktz(VS.80).aspx

lazka:
(The remaining errors are due to #7017 (GCC 10 changes) and not MSYS2 specific)

nirbheek:
I think it's ok to change to `long long` everywhere. There will be no loss of information, since it's required to be 64-bit on all platforms.

lazka:
Changed to long long, updated commit message, all is green (as green as it can be atm)

nirbheek:
> all is green (as green as it can be atm)

Yeah, we should fix that, but not in this PR.

PR #899, opened 2016-10-12
pkgconfig: Handle library names starting with 'lib'

nirbheek:
Sometimes people want the library to start with `lib` everywhere, which is achieved by setting `name_prefix` to `''` and the target name to `'libfoo'`. In that case, try to get the `pkg-config` `'-lfoo'` arg correct. Also, don't warn about anything on Windows.

The other commits add a unittest for this. Had to make some changes to `PkgConfigDependency` to make this work.

PR #7121, opened 2020-05-12
build: always use lib prefix for GCC import libraries

yselkowitz:
Even when name_prefix: '' is specified for Windows, the GCC import library should still have the customary 'lib' prefix, and if name_prefix: 'cyg' is for some reason specified for Cygwin, it should definitely not also apply to the import library.

### DIFF
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -1743,7 +1743,7 @@ class SharedLibrary(BuildTarget):
         elif env.machines[self.for_machine].is_windows():
             suffix = 'dll'
             self.vs_import_filename = '{0}{1}.lib'.format(self.prefix if self.prefix is not None else '', self.name)
-            self.gcc_import_filename = '{0}{1}.dll.a'.format(self.prefix if self.prefix is not None else 'lib', self.name)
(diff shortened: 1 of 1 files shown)


dcbaker:
@jon-turney, could you look at this?

jon-turney:
git blame shows that I wrote those lines in #4480 for reasons which obviously made sense to me at the time.

jon-turney:
@yselkowitz makes a convincing argument about what the behaviour should be to match the platform conventions.

Modulo the test case failures, where expected files need updating, I think the issues here are:

- This reverts how #4480 makes linking with `shared_library('libfoo', name_prefix : '')` via pkg-config work as expected.  #899 is unclear about why this that is preferred to `shared_library('foo', name_prefix : 'lib')`.  Maybe we need some special handling to make these two cases equivalent?

- If multiple libraries exist with the same name, but different name_prefix:, the implib names will clash.  Perhaps meson is being too flexible here and it doesn't make sense to allow anything other than 'lib', '', or [] as a name_prefix: (as anything else should be part of the library name?), so we should (initally) warn about that, with a plan to eventually disallow other values?

PR #7122, opened 2020-05-12
docs: Add a Howto about the null dependency [skip ci]

dcbaker:


docs/markdown/howtox.md
```diff
@@ -260,3 +260,26 @@ The `cmake_module_path` property is only needed for custom CMake scripts. System
...
+```meson
+my_dep = dependency('', required : false)
+if host_machine.system() in ['freebsd', 'netbsd', 'openbsd', 'dragonfly']
+  my_dep = dependency('some dep', required : false)
+elif host_machine.system() == 'linux'
+  my_dep = dependency('some other dep', required : false)
+endif
+
+# Last ditch effort!
+if no my_dep.found()
```
jon-turney, line 282: no -> not

docs/markdown/howtox.md
```diff
@@ -260,3 +260,26 @@ The `cmake_module_path` property is only needed for custom CMake scripts. System
...
 
 More information can be found [here](Dependencies.md#cmake)
+
+## Get a default not-found dependency?
+
+```meson
+null_dep = dependency('', required : false)
+```
+
+This can be used in cases where you want a default value, but might override it
```
jon-turney, line 270: The example could perhaps also illustrate that 'might', showing that it's safe to use that default in targets (e.g if all the detection was under some option, and the default value ends up being used if that option was off)

PR #7123, opened 2020-05-12
Use FeatureDeprecated (and other cleanups)

dcbaker:
This little series was meant to mainly replace a number of `mlog.deprecation` calls with `FeatureDeprecated`, along the way I did a little cleanup.

Suggested by @mensinda

### DIFF
--- a/mesonbuild/optinterpreter.py
+++ b/mesonbuild/optinterpreter.py
@@ -16,10 +16,11 @@ import re
 import functools
 import typing as T
 
-from . import mparser
+from . import compilers
 from . import coredata
 from . import mesonlib
-from . import compilers
+from . import mparser
+from .interpreterbase import FeatureNew
 
 forbidden_option_names = set(coredata.builtin_options.keys())
 forbidden_prefixes = [lang + '_' for lang in compilers.all_languages] + ['b_', 'backend_']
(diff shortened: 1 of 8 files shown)


mesonbuild/optinterpreter.py
```diff
@@ -200,11 +201,8 @@ def evaluate_statement(self, node):
             raise OptionException('Only calls to option() are allowed in option files.')
         (posargs, kwargs) = self.reduce_arguments(node.args)
 
-        # FIXME: Cannot use FeatureNew while parsing options because we parse
-        # it before reading options in project(). See func_project() in
-        # interpreter.py
-        #if 'yield' in kwargs:
-        #    FeatureNew('option yield', '0.45.0').use(self.subproject)
```
mensinda, line 207: There should be an assertion here that `project_meson_versions` is set, since assumptions about the initialization order break easily during a refactoring (especially if there are no comments).

dcbaker: I've done the best I can to address this in the latest version of the series. There is a corner case that's in our test suite that makes it kinda hard to solve generally, so the approach I have is a bit of a hack (I really don't want to rewrite the way this detection works completely), that corner is if a FeatureNew/FeatureDeprecated is triggered in an argument to project(), such as
```meson
project('foo', version : run_command(..., check : true))
```
So I've changed the type of project_meson_versions to a defaultdict so there is always a value, and added a comment to project() implementation

mensinda: Ah, I didn't see this corner case immediately...

jon-turney:
It looks like after this test case `warning/1 version for string div` sometimes doesn't output the expected  `tried to use feature introduced in '0.49.0': / with string arguments`? No idea why that might be!

PR #7132, opened 2020-05-13
symbolextractor: Do not store the size of code objects

nirbheek:
This will almost always change and cause a relink of everything. Our
other symbol extractor implementations do not store this either. We
only need to store the size of data objects, since that necessitates
a relink due to copy relocations.

Drastically reduces the amount of relinking required in gstreamer and
gtk on Linux.

~~This is a very safe and obvious fix, so I'm marking it for the stable release.~~

### DIFF
--- a/mesonbuild/scripts/symbolextractor.py
+++ b/mesonbuild/scripts/symbolextractor.py
@@ -113,7 +113,10 @@ def gnu_syms(libfilename: str, outfilename: str):
             continue
         line_split = line.split()
         entry = line_split[0:2]
-        if len(line_split) >= 4:
+        # Store the size of symbols pointing to data objects so we relink
+        # when those change, which is needed because of copy relocations
+        # https://github.com/mesonbuild/meson/pull/7132#issuecomment-628353702
+        if line_split[1].upper() in ('B', 'G', 'D') and len(line_split) >= 4:
             entry += [line_split[3]]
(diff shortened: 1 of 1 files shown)


jpakkane:
IIRC We originally did not store the size but it was changed when someone filed a MR because it was needed in some cases. Don't remember the details now, though.

nirbheek:
Seems like it was added in https://github.com/mesonbuild/meson/pull/5211. This can't be entirely true, because that would mean that changing any code inside a publicly-exported function would break ABI. Need to figure out what cases cause that error to happen.

nirbheek:
Okay, so apparently this happens because of [copy relocations](https://lists.fedoraproject.org/pipermail/devel/2015-September/214129.html) of data from shared libraries when linking executables without PIE.

The space needed for copy relocations [only changes if the size of a data object such as an array changes](https://docs.oracle.com/cd/E19120-01/open.solaris/819-0690/chapter4-84604/index.html), or if the type of the data object changes.

So we should only store the sizes for the following symbols:

```
"B"
"b" The symbol is in the BSS data section.

"D"
"d" The symbol is in the initialized data section.

"G"
"g" The symbol is in an initialized data section for small objects.
```

The rest is code objects (not data objects), and changes in the size of those won't affect copy relocations.

nirbheek:
Implemented.

Issue #7133, opened 2020-05-13
Crash in gnome module when using -Dlayout=flat


ocrete:
It seems that -Dlayout=flat doesn't work with the gnome module.

```
$ meson  build -Dlayout=flat
The Meson build system
Version: 0.54.0
Source dir: /home/tester/collabora/libnice
Build dir: /home/tester/collabora/libnice/build
Build type: native build
Project name: libnice
Project version: 0.1.16.1
C compiler for the host machine: ccache cc (gcc 10.1.1 "cc (GCC) 10.1.1 20200507 (Red Hat 10.1.1-1)")
C linker for the host machine: cc ld.bfd 2.34-2
Host machine cpu family: x86_64
Host machine cpu: x86_64
Checking for function "clock_gettime" : YES 
Has header "arpa/inet.h" : YES 
Has header "net/in.h" : NO 
Has header "netdb.h" : YES 
Has header "ifaddrs.h" : YES 
Has header "unistd.h" : YES 
Checking for function "poll" : YES 
Checking for function "getifaddrs" : YES 
Compiler for C supports arguments -fno-strict-aliasing: YES 
Message: warning level: 1
Message: werror enabled: false
Found pkg-config: /usr/bin/pkg-config (1.6.3)
Run-time dependency gio-2.0 found: YES 2.64.2
Run-time dependency gthread-2.0 found: YES 2.64.2
Message: Crypto librar requested: auto
Run-time dependency gnutls found: YES 3.6.13
Run-time dependency gstreamer-base-1.0 found: YES 1.16.2
Run-time dependency gupnp-igd-1.0 found: YES 0.2.5
Library m found: YES
Program g-ir-scanner found: YES (/usr/bin/g-ir-scanner)
Program gen-def.py found: YES (/usr/bin/python3 /home/tester/collabora/libnice/nice/gen-def.py)
Program gen-map.py found: YES (/usr/bin/python3 /home/tester/collabora/libnice/nice/gen-map.py)
Compiler for C supports link arguments -Wl,--version-script,/home/tester/collabora/libnice/nice/libnice.ver: YES 
Found pkg-config: /usr/bin/pkg-config (1.6.3)
Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/mesonbuild/mesonmain.py", line 131, in run
    return options.run_func(options)
  File "/usr/lib/python3.8/site-packages/mesonbuild/msetup.py", line 245, in run
    app.generate()
[...]

PR #7135, opened 2020-05-13
Add test for gnome.generate_gir with a flat layout

ocrete:
This is a test that illustrates #7133


ocrete:
I replace the mkdir with os.makedirs().. but I get a new error now with the path not being extract correctly from the dependency.

PR #7136, opened 2020-05-14
gfortran: Fix has_header implementation with GCC 10

nirbheek:
__has_include is not accepted as a pre-processor directive in Fortran
code since GCC 10.

Closes https://github.com/mesonbuild/meson/issues/7017


nirbheek:
@scivision I'm going to merge this for the stable release since it's very safe. Not sure if it's the best fix, but it isn't more wrong than it was before at least :)

scivision:
Thanks this works for GCC 10

Issue #6115, opened 2019-10-29
"cross compiling" to native (build) machine leads to different install paths for libraries


baryluk:
I am using `meson 0.52.0` on Debian Linux testing amd64.

Set some build machine stuff:

```shell
export CC="/usr/bin/x86_64-linux-gnu-gcc-${GCCVERSION}"
export CXX="/usr/bin/x86_64-linux-gnu-g++-${GCCVERSION}"
export PKG_CONFIG="/usr/bin/x86_64-linux-gnu-pkg-config"
```

Prepare cross-file:

```shell
cat > "${MESONCROSSFILE_64}" <<EOF
[binaries]
c = '/usr/bin/x86_64-linux-gnu-gcc-${GCCVERSION}'
cpp = '/usr/bin/x86_64-linux-gnu-g++-${GCCVERSION}'
llvm-config = '/usr/bin/llvm-config-${LLVMVERSION}'
strip = '/usr/bin/x86_64-linux-gnu-strip'
EOF
```

Do a build and install without cross-file:

```shell
cd "${DRMREPO}"
rm -rf "${DRMBUILDDIR}/build-amd64-opt/"
meson "${DRMBUILDDIR}/build-amd64-opt/" \
  --buildtype=plain \
  --prefix="${INSTALLDIR}/build-amd64-opt/install"
ninja -C "${DRMBUILDDIR}/build-amd64-opt/"
ninja -C "${DRMBUILDDIR}/build-amd64-opt/" install
```

Relevant output:
```
...
Build type: native build
Project name: libdrm
Project version: 2.4.100
C compiler for the host machine: /usr/bin/x86_64-linux-gnu-gcc-9 (gcc 9.2.1 "x86_64-linux-gnu-gcc-9 (Debian 9.2.1-15) 9.2.1 20191027")
C linker for the host machine: GNU ld.bfd 2.33.1
Host machine cpu family: x86_64
Host machine cpu: x86_64
Run-time dependency threads found: YES 
Found pkg-config: /usr/bin/x86_64-linux-gnu-pkg-config (0.29)
...
Installing intel/libdrm_intel.so.1.0.0
  to /home/user/mesa-git/installdir/build-amd64-opt/install/lib/x86_64-linux-gnu
...
Installing /home/user/mesa-drm-git/builddir/build-amd64-opt/meson-private/libdrm_amdgpu.pc
[...]

PR #7137, opened 2020-05-14
unit tests: Fix broken test_cross_libdir test

nirbheek:
test_cross_libdir() was broken because we were passing `--libdir=lib` when invoking meson.

We didn't notice that https://github.com/mesonbuild/meson/issues/6115 broke because of this.

With the second commit in this PR, this fixes https://github.com/mesonbuild/meson/issues/6115.


nirbheek:
This PR will cause the CI to fail because it fixes a broken test that was incorrectly passing.

nirbheek:
That should fix it.

PR #7138, opened 2020-05-14
ui/qt: use new directory layout for qt on android

ystreet:
Now follows ios and other platform directory layouts.  Moves from
separate android_$arch directories to every library containing a _$arch
suffix.  e.g. libQt5Core_x86.a in a single directory.

cc @nirbheek

### DIFF
--- a/mesonbuild/dependencies/ui.py
+++ b/mesonbuild/dependencies/ui.py
@@ -440,6 +440,9 @@ class QtBaseDependency(ExternalDependency):
                     suffix += '_armeabi-v7a'
                 elif cpu_family == 'aarch64':
                     suffix += '_arm64-v8a'
+                else:
+                    mlog.warning('Android target arch {!r} for Qt5 is unknown, '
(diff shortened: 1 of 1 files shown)


mesonbuild/dependencies/ui.py
```diff
@@ -426,6 +429,17 @@ def _get_modules_lib_suffix(self, is_debug):
...
+            if self.env.machines[self.for_machine].is_android():
+                cpu_family = self.env.machines[self.for_machine].cpu_family
+                if cpu_family == 'x86':
+                    suffix += '_x86'
+                elif cpu_family == 'x86_64':
+                    suffix += '_x86_64'
+                elif cpu_family == 'arm':
+                    suffix += '_armeabi-v7a'
+                elif cpu_family == 'aarch64':
+                    suffix += '_arm64-v8a'
```
nirbheek, line 442: Maybe we want to print a warning if the cpu_family is unknown?

PR #7150, opened 2020-05-14
Docs: Make the suggested use for the not-found dependency better

dcbaker:
@jon-turney suggested this, and it makes more sense to me as well.

### DIFF
--- a/docs/markdown/howtox.md
+++ b/docs/markdown/howtox.md
@@ -271,6 +271,7 @@ This can be used in cases where you want a default value, but might override it
 later.
 
 ```meson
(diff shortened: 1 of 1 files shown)


docs/markdown/howtox.md
```diff
@@ -279,7 +280,9 @@ elif host_machine.system() == 'linux'
 endif
 
 # Last ditch effort!
```
jon-turney, line 282: This comment seems to be stale now?

dcbaker: yup, removed.

PR #7151, opened 2020-05-14
run_*test.py: print meson version at start of test

scivision:
When building a general Meson project, the Meson version is printed first. I find this very handy, and it would be even more useful to have the Meson version printed at the start of Meson's own run_*test.py scripts, as they can take over an hour to run, so you want to be sure you're testing the intended Meson version. 

I also saw some PEP8 issues including missing import and unused variable that I fixed in the second commit.

## New functionality

* print Meson version at start of each Meson internal run_*tests.py execution.

### DIFF
--- a/run_unittests.py
+++ b/run_unittests.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import typing as T
 import stat
 import subprocess
 import re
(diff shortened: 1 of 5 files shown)


run_unittests.py
```diff
@@ -6036,7 +6037,7 @@ def test_introspect_installed(self):
             }
             the_truth_2 = {'/usr/lib/libsome.dylib',
                            '/usr/lib/libsome.0.dylib',
-            }
+                           }
```
nirbheek, line 6040: nitpick: Maybe this block should match the indentation of `the_truth`?

scivision: `flake8` with default settings (not overridden by me) complained about it--maybe it was a false positive?

nirbheek: I'm saying it should be:

```py
the_truth_2 = {
    '/usr/lib/libsome.dylib',
    '/usr/lib/libsome.0.dylib',
}
```

Like `the_truth` above it. Flake8 should accept this.

scivision: For me, flake8 says `E124 closing bracket does not match visual indentation` but it does match project style, so I put it as you mentioned, thanks

PR #7153, opened 2020-05-15
travis/macos: Restore the old pkg-config behaviour

nirbheek:
We need to test both "have pkg-config" and "don't have pkg-config" pathways on macOS, which is why pkg-config was only installed in one branch based on `--unity=on/off`.

This broke in https://github.com/mesonbuild/meson/pull/6913.

### DIFF
--- a/ci/travis_install.sh
+++ b/ci/travis_install.sh
@@ -7,9 +7,11 @@ msg() { echo -e "\x1b[1;32mINFO:  \x1b[37m$*\x1b[0m"; }
 if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
   msg "Running OSX setup"
   brew update
+  # Run one macOS build with pkg-config available (pulled in by qt), and the
+  # other (unity=on) without pkg-config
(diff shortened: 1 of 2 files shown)


ci/travis_install.sh
```diff
@@ -7,9 +7,11 @@ msg() { echo -e "\x1b[1;32mINFO:  \x1b[37m$*\x1b[0m"; }
 if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
   msg "Running OSX setup"
   brew update
+  # Run one macOS build with pkg-config available (pulled in by qt), and the
+  # other (unity=on) without pkg-config
   brew install qt ldc llvm ninja
   if [[ "$MESON_ARGS" =~ .*unity=on.* ]]; then
-    which pkg-config || brew install pkg-config
+    which pkg-config && chmod -x `which pkg-config`
```
mensinda, line 14: Maybe use `rm $(which pkg-config)` to better communicate the intent?

PR #4010, opened 2018-08-11
Purge much `is_cross` and `<things>_cross` without changing user interfaces---includes on #5263

Ericson2314:
This is a draft of how I think cross should work. Compilers and targets track the machine they run on / target, rather than `is_cross`/`want_cross`. This is usually done with a `for_machine` field and `MachineChoice` enum. `MachineChoice` is abstract, representing the "build/host/target"' name, rather than the definition of the machines themselves. This allows the same code paths to be shared in the native and cross cases, vastly reducing the space of code paths.

An example probably makes this clearer. Consider the situation of a single installed C target. As it is installed, it must be for the host machine (or target, but it's not currently possible to make things for the target platform). In the native case we have a single compiler for the build and host machines. In the cross case we have different ones for each.

As the code stands today, the host compiler goes in `compilers` in the native case or `cross_compilers` in the cross case, and the target fishes out the compiler it needs from either `compilers` in the native case, or `cross_compilers` in the target case. So we have two conditional bits of code: where to put the compiler, and where to fetch the compiler, that should cancel out, but nothing really *ensures* that they in fact do. With the code in this PR, it's always unconditionally stored in the same places (`core_data.compiler`, `build.for_host.compilers`). This ensures the same behavior by construction.

It's true with enough work Meson will work either way, so this sort of refactor might seem superfluous. But is also helps *users* of Meson catch cross-related bugs in their `meson.build` files. Currently, we error in the cross case when someone tries to install a native (build platform) target, or link together cross and native targets. But in the native case we have no idea when those things would happen. But with this change since we track build machine vs host machine vs target machine in *all* cases, we can always when a build platform target is installed, or when a host machine and build machine target are linked together. These errors become warnings in the native case (for backwards comparability), helping users that rarely cross-compile their code detect this bugs sooner.

I hope in opening this now, I can make myself clearer in the issues, and better motivate incremental #3993. If this looks good, it can hopefully be broken up into additional manageable issues. It also lays a foundation upon which https://github.com/mesonbuild/meson/issues/3972 https://github.com/mesonbuild/meson/issues/3969 would be natural to implement.

Fixes #5102

PR #7157, opened 2020-05-16
WIP: gobject-introspection cross with pkg-config

Ericson2314:
gobject-introspection is infamous for being difficult to get working
with cross compilation. But the situation has improved recently with
https://gitlab.gnome.org/GNOME/gobject-introspection/-/merge_requests/64,
thanks to @kanavin from the Yocto project. The upshot is now only
*some*, not all, of the executables need to be run on the host platform
with the exe wrapper.

On the meson side, there have been two attempts to fix things for cross:

- https://github.com/mesonbuild/meson/pull/2965from @kanavin and the
Yocto projejct, which because it predates
https://github.com/mesonbuild/meson/pull/4010 had to be somewhat hacky

- https://github.com/mesonbuild/meson/pull/7072 recently merged which
allows specifying some binaries with a cross file.

But, I think we can make a more seamless user interface that won't
require extra config, like for native builds. gobject-introspection
provides the binaries in its pkg-config file, and Meson now cleanly
supports separate `native: true` and `native: false` pkg-config paths
and lookup.

We should just need to add separate `native: true` and `native: false`
deps, and I have done that, but I am not sure exactly which should be
used when. I am coming at this as a distro maintainer not even
particularly involved with gnome things, and so I'll need some advice on
what to do next.

CC @jtojnar

### DIFF
--- a/mesonbuild/modules/gnome.py
+++ b/mesonbuild/modules/gnome.py
@@ -29,8 +29,8 @@ from . import get_include_args
 from . import ExtensionModule
 from . import ModuleReturnValue
 from ..mesonlib import (
-    MachineChoice, MesonException, OrderedSet, Popen_safe, extract_as_list,
-    join_args, unholder,
+    MachineChoice, MesonException, OrderedSet, Popen_safe, PerMachine,
+    extract_as_list, join_args, unholder,
 )
 from ..dependencies import Dependency, PkgConfigDependency, InternalDependency, ExternalProgram
 from ..interpreterbase import noKwargs, permittedKwargs, FeatureNew, FeatureNewKwargs
@@ -45,7 +45,7 @@ gresource_dep_needed_version = '>= 2.51.1'
 native_glib_version = None
 
 class GnomeModule(ExtensionModule):
-    gir_dep = None
+    gir_dep = PerMachine(None, None)
 
     @staticmethod
     def _get_native_glib_version(state):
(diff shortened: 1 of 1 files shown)


kanavin:
This would have to be carefully tested in a cross-environment. For reference, this is how Yocto project patches meson currently to support the cross-case (which is for an older meson version so will have to be rebased to master):
http://git.yoctoproject.org/cgit/cgit.cgi/poky/tree/meta/recipes-devtools/meson/meson/0002-gobject-introspection-determine-g-ir-scanner-and-g-i.patch

Also note the prepending of PKG_CONFIG_SYSROOT_DIR.

Ericson2314:
So for starters, which of `giscanner` and `gicompiler` can be run natively, and which need to dlopen the library and thus require the exe wrapper?

kanavin:
Unfortunately it's not as "simple" as that. Here's what Yocto does with the wrappers.

giscanner is run natively, but still requires *three* wrappers. One is for running giscanner itself (note that it takes two other wrappers as arguments):

`       # Write out a wrapper for g-ir-scanner itself, which will be used when building introspection files
        # for glib-based packages. This wrapper calls the native version of the scanner, and tells it to use
        # a qemu wrapper for running transient target binaries produced by the scanner, and an include directory
        # from the target sysroot.
        cat > ${B}/g-ir-scanner-wrapper << EOF
#!/bin/sh
# This prevents g-ir-scanner from writing cache data to $HOME
export GI_SCANNER_DISABLE_CACHE=1

g-ir-scanner --lib-dirs-envvar=GIR_EXTRA_LIBS_PATH --use-binary-wrapper=${STAGING_BINDIR}/g-ir-scanner-qemuwrapper --use-ldd-wrapper=${STAGING_BINDIR}/g-ir-scanner-lddwrapper --add-include-path=${STAGING_DATADIR}/gir-1.0 --add-include-path=${STAGING_LIBDIR}/gir-1.0 "\$@"
EOF
        chmod +x ${B}/g-ir-scanner-wrapper
`
Second is for running the transient target binary through qemu (or similar) emulation:
`       # Write out a qemu wrapper that will be given to gi-scanner so that it
        # can run target helper binaries through that.
        qemu_binary="${@qemu_wrapper_cmdline(d, '$STAGING_DIR_HOST', ['\\$GIR_EXTRA_LIBS_PATH','.libs','$STAGING_DIR_HOST/${libdir}','$STAGI
NG_DIR_HOST/${base_libdir}'])}"
        cat > ${B}/g-ir-scanner-qemuwrapper << EOF
#!/bin/sh
# Use a modules directory which doesn't exist so we don't load random things
# which may then get deleted (or their dependencies) and potentially segfault
export GIO_MODULE_DIR=${STAGING_LIBDIR}/gio/modules-dummy

$qemu_binary "\$@"
if [ \$? -ne 0 ]; then
    echo "If the above error message is about missing .so libraries, then setting up GIR_EXTRA_LIBS_PATH in the recipe should help."
    echo "(typically like this: GIR_EXTRA_LIBS_PATH=\"$""{B}/something/.libs\" )"
    exit 1
fi
EOF
`
Third is for ldd:
`       # Write out a wrapper to use instead of ldd, which does not work when a binary is built
        # for a different architecture
        cat > ${B}/g-ir-scanner-lddwrapper << EOF
#!/bin/sh
prelink-rtld --root=$STAGING_DIR_HOST "\$@"
EOF
        chmod +x ${B}/g-ir-scanner-lddwrapper
`

kanavin:
Basically the bottom line is that you cannot use a 'generic exe wrapper' and need to write custom specific wrappers for everything, and instruct meson to use them one way or another.

Ericson2314:
@kanavin Thanks that super helpful! I have a few related questions that are not meson-specific, so I guess at the risk of strewing about the conversation I will head back to the issue?

Ericson2314:
Where is `prelink-rtld` from BTW? I built Yocto's version of prelink and only got a `prelink`.

kanavin:
http://git.yoctoproject.org/cgit/cgit.cgi/prelink-cross/tree/src/rtld/Makefile.am?h=cross_prelink I guess?

Ericson2314:
Oh I was on the wrong branch-----despite the right branch being merged into the branch I was on.

jameshilliard:
@aduskett who maintains gobject-introspection in buildroot should probably review this as well.

Ericson2314:
If @aduskett would be so interested, I'd suggest starting with https://gitlab.gnome.org/GNOME/gobject-introspection/-/merge_requests/227 (my pared-down PR just for more testing) and, if that looks good, https://gitlab.gnome.org/GNOME/gobject-introspection/-/merge_requests/224.

aduskett:
> If @aduskett would be so interested (and it would be great to have a buildroot person taking a look!), I'd suggest looking upstream with https://gitlab.gnome.org/GNOME/gobject-introspection/-/merge_requests/227 (my pared-down PR just for more testing) and, if that looks good, https://gitlab.gnome.org/GNOME/gobject-introspection/-/merge_requests/224.
> 
> The main thing blocking me from finishing this PR off is a desire to not have such a big difference from upstream.

Yeah, I can take a look hopefully later today.

PR #7158, opened 2020-05-16
symbolextractor: add OpenBSD support

ajacoutot:


jpakkane:
Note that #7103 changed the behaviour of rpath filtering so it may be possible that this will cause problems on OpenBSD. I wanted to merge that first but was not fast enough. If possible @ajacoutot please verify that current master works on OpenBSD.

ajacoutot:
@jpakkane  I will test that the rpath filtering change did not introduce any regression on OpenBSD. If it does, I will follow up in an issue/PR.
Thanks :-)

Issue #7004, opened 2020-04-21
meson: error: unrecognized arguments: --default_library build-shared


dota17:
**Describe the bug**

When I use `default_library` to run like this:

`meson --buildtype ${BUILD_TYPE} --default_library ${LIB_TYPE} . build-${LIB_TYPE}`

it tells me that:
> meson: error: unrecognized arguments: --default_library build-shared

So, I think it should be `default-library`, but not `default_library`

**system parameters**

* what `meson --version`
0.51.1
* what `ninja --version` if it's a Ninja build
1.9.0

dcbaker:
Yes, the argument should be:
```console
meson --default-library shared
```

You can use `meson setup --help` to see all of the builtin options.

norihiro:
I'm reading [HTML document](https://mesonbuild.com/Builtin-options.html) but it is not clearly mentioned. Do you have a plan to update the document?

PR #7162, opened 2020-05-18
Fix typo in default-library docs

williamfgc:
Related to #7004 
Just fixing leftover typo in doc as I came across with static compilations.


xclaesse:
That's not a typo, options have underscore not dash. It's confusing because built in options can also be passed on command line as --default-library as well as -Ddefault_library.

williamfgc:
@xclaesse thanks for clarifying. Just bringing awareness that docs can be improved on this topic. Would help meson newcomers like myself. Thanks!

jpakkane:
Closing this to reduce clutter on the MR page. Thanks. If you want to create a new MR that updates the comment string in the table to explain the issue, feel free to do so. Note, though, that doc-only changes should have `[skip ci]` at the end of their commit messages so CI is not run (it is quite slow).

PR #7164, opened 2020-05-18
Limit the filename length

### LINKED ISSUE #7159: meson create filename longer than 255
**Describe the bug**
When work with project have deep folders, meson tend to create very long filename. For example the gst-build projects. The ccache report `Failed to create temporary file for ... File name too long`
https://gitlab.freedesktop.org/gstreamer/gst-build/-/issues/97
**To Reproduce**

**Expected behavior**
meson should check the filename and path length. When it exceed 255 use hash value of it instead.

**system parameters**
* plain native build 
* ubuntu 19.10
* Python 3.7.5
* `meson --version` 0.54.1
* `ninja --version` 1.9.0.git.kitware.dyndep-1.jobserver-1

braindevices:
try to fix #7159 
When the project has very deep folder, object_filename_from_source() tends to create very long filename exceeding linux filename limit of 255.  Here, I try to impose a max_filename_len to it. Once it exceed the length, it truncate the filename and add sha256 of the filename to the front to make sure the resulting length == max_filename_len. The hash value length is 64, thus the max_filename_len should at least 65 otherwise the filename length is actually increased. I also set the maximum value to 230 to give it some room for the postfix.

### DIFF
--- a/docs/markdown/Builtin-options.md
+++ b/docs/markdown/Builtin-options.md
@@ -86,7 +86,7 @@ Using the option as-is with no prefix affects all machines. For example:
 | warning_level {0, 1, 2, 3}           | 1             | Set the warning level. From 0 = none to 3 = highest            | no             |
 | werror                               | false         | Treat warnings as errors                                       | no             |
 | wrap_mode {default, nofallback,<br>nodownload, forcefallback} | default | Wrap mode to use                            | no             |
-| max_filename_len {>=65, <=250} | 155 | maximum object filename | no
+| max_filename_len {>=65, <=230} | 155 | maximum object filename | no
 
 <a name="build-type-options"></a>
 For setting optimization levels and toggling debug, you can either set the
--- a/docs/markdown/Builtin-options.md
+++ b/docs/markdown/Builtin-options.md
@@ -86,6 +86,7 @@ Using the option as-is with no prefix affects all machines. For example:
 | warning_level {0, 1, 2, 3}           | 1             | Set the warning level. From 0 = none to 3 = highest            | no             |
 | werror                               | false         | Treat warnings as errors                                       | no             |
 | wrap_mode {default, nofallback,<br>nodownload, forcefallback} | default | Wrap mode to use                            | no             |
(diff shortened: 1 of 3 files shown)


braindevices:
why msys2 always fail?
```
error: could not open file /var/cache/pacman/pkg/msys2-runtime-3.1.4-1-x86_64.pkg.tar.zst: Child process exited with status 127
error: failed to commit transaction (cannot open package file)
Errors occurred, no packages were upgraded.
```
It does not seem to related to my commits.

jpakkane:
There should not be an option for the length. We should always do the right thing rather than bother end users with that.

eli-schwartz:
> why msys2 always fail?

Sounds like an MSYS2 bug (their pacman binary is compiled without zstd support, but is trying to download system updates using zstd compression).

dabrain34:
I got this issue fixed by reducing to 65 the len of filename. https://gitlab.freedesktop.org/gstreamer/gst-build/-/issues/9

I'm not sure about the right approach as the default len of 130 was not a good number here. 

But first the filename should be kept and not replaced by a hash for human understanding and then maybe we could detect the len by platform but here the issue is related to ccache and not the system

I open an issue on ccache side:

https://github.com/ccache/ccache/issues/593

xclaesse:
We already use hash to shorten some paths, see _get_id_hash() in build.py, can se reuse that method?

dabrain34:
yeah but on what condition ?

dabrain34:
https://github.com/ccache/ccache/blob/8386a5a714d623c12e1aeff802916e64f5c17de4/src/legacy_util.cpp#L620

Here ccache takes the orginal filename and put addtional chars to it ;)

braindevices:
@dabrain34 the host name can be very long, this is hard to coop then.

eli-schwartz:
I rediscovered the issue in #10788 and fixed it by unconditionally switching to hash prefixes, but only when the depth is >5 and hashing the additional components reduces the length. This shouldn't be a problem anymore.

PR #6620, opened 2020-02-13
Add a mechanism for validating meson output in tests

jon-turney:
Add a mechanism for validating meson output in tests.

In particular, this allows checking that `faiiling-meson` and `warning-meson` tests are actually exercising the intended code-paths and failing with the expected messages (_spoiler_: some of them aren't).

There's also some unit tests which just exist to verify the expected output occurs (e.g. `test_warning_location`), which could be moved to use this mechanism.

PR #6203, opened 2019-11-19
Add meson.override_dependency()

xclaesse:
Similar to meson.override_find_program() but overrides the result of the
dependency() function.

Also ensure that dependency() always returns the same result when
looking for the same dependency, this fixes cases where parts of the
project could be using a system library and other parts use the library
provided by a subproject.

PR #7172, opened 2020-05-19
Make the expected output check mandatory for failing-meson and warning-meson tests

jon-turney:
A follow-up to #6620, this adds the expected output for tests added since that work was started, and makes it mandatory to have expected output for failing-meson and warning-meson tests, so it will (hopefully) get added for new tests when they are added.

### DIFF
--- a/run_project_tests.py
+++ b/run_project_tests.py
@@ -593,18 +593,16 @@ def _run_test(test: TestDef, test_build_dir: str, install_dir: str, extra_args, 
 
     return testresult
 
-def gather_tests(testdir: Path) -> T.List[TestDef]:
+def gather_tests(testdir: Path, stdout_mandatory: bool) -> T.List[TestDef]:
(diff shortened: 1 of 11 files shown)


jon-turney:
Note: the current output for failing-meson tests 36, 67, 84 is a bit wonky, as noted in https://github.com/mesonbuild/meson/pull/6203#issuecomment-619432227

dcbaker:
This looks really nice overall. I really like the work going into the test.json format.

jpakkane:
This caused massive test failures for some reason even though it got green in CI, so I had to revert this out of master. Please reopen. I'll get the other master test issues fixed in the mean time (except msys, that requires an update to msys).

jon-turney:
I think at least part of the problem is that this collides with #7123, which changes the text of a warning, which we were now validating.

jon-turney:
No option to re-open this PR, so created a new one: #7197

Issue #553, opened 2016-05-24
Ambiguous threads dependency


LRN:
`dependency('threads')` is poorly-documented.

Looking at the code (0.31.0), it seems that all that `dependency('threads')` does is create a dependency object that has `needs_threads() == True`.

`needs_threads() == True` makes the implementation add `compiler.thread_flags()` and `compiler.thread_link_flags()` to the flags that are actually used.

`compiler.thread_flags()` is `-pthread` for C and an empty string for C++
`compiler.thread_link_flags()` is `-pthread` for C and an empty string for C++

`<threading-rant>`

This does not seem like a good idea. A better approach is to focus on threading APIs.

I.e. `meson.build` would request threading API `X` for language `Y`. Meson will then do appropriate lookups and provide appropriate flags.

Threading APIs that i know of:

Name  | OS | library      | flags     | languages
------|----|--------------|-----------|----------
POSIX | GNU| libc         | -pthread? | C/C++, bindings
POSIX |*BSD| libpthread   | -lpthread | C/C++, bindings
POSIX |W32 | libpthread   | -lpthread | C/C++, bindings
W32   |W32 | libkernel32  | -lkernel32| C/C++, bindings
Python|Any | Python stdlib|           | Python

and so on for C#, Rust and other languages (Fortran?).

That said, it doesn't make much sense to check for these, as, for example, Python threading API is always part of Python stdlib, and W32 threading API is always available on W32 (-lkernel32 is hardcoded into GCC spec files, IIRC, you have to make an *effort* to *not* to link to it; dunno about MSVC).

POSIX threads is the only API that needs actual checking (as it might be in libc or in libpthread or somewhere else, in exotic OSes, and it might need special compiler flags).

Projects can also support multiple threading APIs (glib supports pthreads and W32 threads, at least) in their code. Which one is used is usually decided at configuration time by a configure option.

[...]

Issue #2835, opened 2017-12-27
Feature: MKL dependency


alyst:
It would be nice to support [Intel Math Kernel Library](https://software.intel.com/en-us/mkl) via `dependency('mkl')`, which would replicate the logic of [Intel MKL Link Advisor](https://software.intel.com/en-us/articles/intel-mkl-link-line-advisor).

arteymix:
MKL now provides a pkg-config file that covers static/dynamic linkage and 32/64-bit integer interfaces since 2018.1 release.

https://software.intel.com/en-us/articles/intel-math-kernel-library-intel-mkl-and-pkg-config-tool

I'll investigate if this issue is still pertinent.

rgommers:
> I'll investigate if this issue is still pertinent.

This issue is still relevant, although it's larger than only MKL - generic BLAS/LAPACK dependency detection is needed. From comments in gh-2847:

_FTR the "correct" way of solving this is to add dependency('BLAS') so the fix is implemented only once in upstream, rather than copypasting the detection logic from one meson.build file to another._

...

_It would be great if you could work on BLAS/LAPACK/MKL dependency classes in Meson so that everyone benefits from being able to find and use them. The BoostDependency object is a good start in mesonbuild/dependencies/. Poke me if you have questions!_

dcbaker:
Is this different than the scalapack dependency we already have?

rgommers:
That's a good start, but it looks different to me. This is what we'd want I think: https://github.com/mesonbuild/meson/pull/5186#issuecomment-479176488. To me it looks like the ScaLAPACK support is a good start, but not the same. ScaLAPACK itself is a subset of LAPACK redesigned for high-performance distributed computing. From reading the scalapack dependency code and the docs for it, it's unclear to me if today `dependency('mkl')` standalone works.

For more context: I'm looking into what we'd need to use Meson in SciPy. There are multiple versions of BLAS and LAPACK, and the ideal interface is being able to give a list of ones that are accepted, in order of preference:
```
# There's 64-bit flavors (ILP64, e.g. 'openblas_ilp64') that we'd need for SciPy too
blas_dep = dependency('mkl', 'openblas', 'blis', 'accelerate', 'atlas', 'netlib')
lapack_dep = dependency('mkl', 'openblas', 'libflame', 'accelerate', 'atlas', 'netlib')
```
[...]

PR #7174, opened 2020-05-19
Query the threading model from dependency('threads')

dcbaker:
Fixes: #553


docs/markdown/Reference-manual.md
```diff
@@ -2456,15 +2456,17 @@ an external dependency with the following methods:
    - sources: any compiled or static sources the dependency has
 
  - `get_variable(cmake : str, pkgconfig : str, configtool : str,
-   internal: str, default_value : str, pkgconfig_define : [str, str])`
+   internal: str, mesont: str, default_value : str, pkgconfig_define : [str, str])`
```
mensinda, line 2459: Typo: `mesont` --> `meson`

dcbaker:
Fixed a couple more typos (as well as the one mensida caught). Also made the test more accurate (fixed mingw)

docs/markdown/snippets/thread_model.md
```diff
@@ -0,0 +1,27 @@
...
+implementation it is a bit different:
+
+```meson
+dep_threads = dependency('threads', method : 'cmake')
+if dep_threads.get_variable(cmake : 'CMAKE_USE_WIN32_THREADS_INIT') == '1':
+   # using windows threads
+elif dep_threads.get_variable(cmake : 'CMAKE_USE_PTHREADS_INIT') == '1':
+   # using pthreads.
+endif
+```
```
jpakkane, line 27: Having two different ways of querying this information is not nice. There should only be one that works with all methods. Maybe we could do something like:

```
thread_dep = dependency('threads')
if thread_dep.threading_type() == 'posix'
  ...
endif
```

And where calling `threading_type` on a dependency that is not an actual thread dep would be a hard error?

dcbaker: I really dislike adding special methods to generic types that only work on one specific class.

What if you have code like:
```meson
dep_threads = dependency('', required : false)
if <something>
    dep_threads = dependency('threads')
endif
....

if thread_dep.threading_type() == 'posix'
   # what happens here?
endif
```

Do we implement the method for all dependency objects, and return a value like `''`? That sounds like a really good way to cause hard to detect bugs. If we don't implement it though, in the case above you *have* to ensure that `dep_thread` is actually a thread dep. That takes away some of the nice polymorphic properties. But with my proposal you could do:

```meson
dep_threads = dependency('', required : false)
if <something>
    dep_threads = dependency('threads')
endif

if dep_threads.get_variable(meson : 'model', cmake : 'CMAKE_USE_PTHREADS_INIT', default_value : '0') in ['posix', '1']
  ...
if dep_threads.get_variable(meson : 'model', cmake : 'CMAKE_USE_WIN#@_ THREADS_INIT', default_value : '0') in ['win32', '1']
  ...
endif
```

jpakkane: OTOH I dislike adding kwargs that are unintuitive and only used for a special case, especially if they user needs to do extra work to get the thing they really want. One approach could be to be able to specify an value that is returned if the dependency is not a thread dep. Something along the lines of:

```meson
# Returns 'something' if the dependency is not a thread dep.
t = some_dep.threading_type('something')
```

dcbaker: Realistically though when are you going to use cmake for thread dependencies? meson will *always* return something using the built-in finder unless you specifically ask it to use cmake. that means for 99% of users:
```meson
thread_model = dep_threads.get_variable(meson : 'model')
if thread_model == 'posix'
   ...
elif thread_model == 'win32'
  ...
endif
```
is completely sufficient.

Just because we haven't yet found other uses for the `meson` keyword to get_variable doesn't mean we wont. We wrap up a lot of dependenices in meson with their own special finder methods, I suspect we could use it to expose a mess of stuff from boost, for example.

jpakkane: The reason I don't like adding general functionality that might get used in the future is that surprisingly often it will never be used. This is exactly YAGNI. If there are at least two (preferably three) use cases that are known and added, then it is of course a different case altogether.

michaelbadcrumble:
Should this be like an introspection thing because as a Meson build user this seems strange just to query the threading dependency just to see what type it is. If you think about it kinda adds some none constant functionality.

dcbaker:
No, this isn't an introspection thing. If you write code that calls `pthread_init`, you *need* the threading implementation to be pthreads. If your code calls `CreateThread` you *need* to have win32 threads.

docs/markdown/snippets/thread_model.md
rgommers: I just ran into this issue. SciPy code is using `std::thread`, and on Windows with MSVC the build failed because everything assumed that _if_ the threads dependency is available, `pthread.h` should be included. 

> If there are at least two (preferably three) use cases that are known and added, then it is of course a different case altogether.

Perhaps to break the deadlock: I think this "what's the specific implementation" API would also be helpful for better BLAS and LAPACK support - those have many flavors, and we'd like a generic detection for them (xref https://github.com/mesonbuild/meson/issues/2835#issuecomment-792600589). And then we do need to know which one was found, because they for example use different Fortran ABI's.

jpakkane: Good, now we have an actual use case. What is it that you actually would need from this? And would something simple like this work:

```meson
dep = dependency('threads')
if dep.get_variable('threading_type', 'dummyfallback') == ...
```

rgommers: The SciPy code uses `std::thread`, so is generic (the easy case here compared to C). So all I need is for my code in `meson.build` to emit a single define that tells the C++ whether to include `pthread.h`. It looks like:
```C++
#include <thread>
#ifdef POCKETFFT_PTHREADS
#  include <pthread.h>
#endif
```
So something like:
```meson
if dep.get_variable('threading_type') == 'pthread'`
  cpp_flags += '-DPOCKETFFT_PTHREADS'
endif
```
would be enough.

I'd be happy with any syntax here. Reading the above thread, the one thing that would be concerning is if the code I'd have to write in `meson.build` would need to know whether it was Meson or CMake having done the detection - that seems like something that should be handled internally in Meson and not leak to the user.

Here is how I had to (unsatisfactorily) work around the problem now, by simply assuming `pthreads` is not installed when building with MSVC (because I think the regular `<thread>` is always available): https://github.com/scipy/scipy/pull/16957/commits/027635e2cca032d0596c228483424c1939c8c005

In the commit I point to I could have worked a bit harder and try-compile some code that does `#include <pthread.h>`, but that seems to defeat the purpose of the generic thread dependency a bit.

tristan957:
It sounds like @jpakkane is good to go on the direction @dcbaker originally took this PR?

Issue #894, opened 2016-10-11
Add valadoc support


nirbheek:
There is no clean way to install Vala documentation generated with a custom target (#893), so we really need this.

Found while investigating #890

inigomartinez:
I've been porting [gitg](https://wiki.gnome.org/Apps/Gitg) and it also uses `valadoc` to generate documentation. I also tried to use a `custom_target`, but it seems clear to me that a custom `valadoc` support is necessary.

nielsdg:
Note that you _can_ generate Valadoc with a `custom_target()` these days, but that it's still a bit cumbersome to write. Feel free to comment on #7177 to make sure it works for your use case.

grindhold:
bump :) is there anything happening on this issue? I've just re-encountered the issue.

grindhold:
bump :) is there anything happening on this issue? I've just re-encountered the issue.

eli-schwartz:
See the linked PR: #7177

PR #7177, opened 2020-05-20
gnome: Add support for Valadoc

nielsdg:
See https://github.com/mesonbuild/meson/issues/894


nielsdg:
A bit more explanation: I maintain several Vala projects which at the moment define a custom target to accomplish the same thing, which is a bit of an annoyance. So I used those as basis for this PR.

* https://gitlab.gnome.org/GNOME/folks/-/blob/master/docs/devhelp/meson.build
* https://gitlab.gnome.org/GNOME/folks/-/blob/master/docs/gtkdoc/meson.build
* https://gitlab.gnome.org/GNOME/gnome-contacts/-/blob/master/docs/meson.build

jpakkane:
LGTM, but this needs a release note snippet.

docs/markdown/Gnome-module.md
```diff
@@ -294,7 +294,26 @@ library.
 Returns a custom dependency that can be included when building other
 VAPI or Vala binaries.
 
-*Added 0.36.0*
```
dcbaker, line 297: I don't think this should be removed, right?

mesonbuild/modules/gnome.py
```diff
@@ -1689,5 +1689,93 @@ def generate_vapi(self, state, args, kwargs):
...
+    @FeatureNew('gnome.valadoc', '0.55.0')
+    def valadoc(self, state, args, kwargs):
+        build_dir = os.path.join(state.environment.get_build_dir(), state.subdir)
+        source_dir = os.path.join(state.environment.get_source_dir(), state.subdir)
+
+        if not args:
+            raise MesonException('valadoc takes at least one positional argument')
+
+        # Build the command line
+        cmd = [ self.interpreter.find_program_impl('valadoc') ]
```
dcbaker, line 1703: I don't see the rest of the gnome module using spaces around `[` and `]` like this, and we don't do it anywhere else in the code base.

mesonbuild/modules/gnome.py
```diff
@@ -1689,5 +1689,93 @@ def generate_vapi(self, state, args, kwargs):
...
+        cmd += [ '--directory=' + os.path.join(build_dir, package) ]
+
+        # Inputs
+        inputs = self.interpreter.source_strings_to_files(args)
+        for f in inputs:
+            if isinstance(f, mesonlib.File):
+                cmd += [ f.absolute_path(state.environment.source_dir,
+                                         state.environment.build_dir) ]
+            else:
+                raise InterpreterException('valadoc inputs can only be strings or file objects')
```
dcbaker, line 1763: I think not support CustomTarget, CustomTargetIndex, and ConfigureFiles here is going to be a limitation that needs to be fixed at some point so we should probably do it now.

Aside -- we've reimplemented the "handle strings, files, CustomTargets, CustomTargetIndexes, and ConfigureFiles" enough times we should probably have a helper somewhere for it

mesonbuild/modules/gnome.py
```diff
@@ -1689,5 +1689,93 @@ def generate_vapi(self, state, args, kwargs):
...
+
+        # Combine into a custom target
+        target_name = package + '-valadoc'
+        custom_kwargs = {
+             'command': cmd,
+             'input': inputs,
+             'output': package,
+             # 'depends': vapi_depends,
```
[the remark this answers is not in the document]
nielsdg: No longer needed

mesonbuild/modules/gnome.py
nielsdg: Fixed!

docs/markdown/Gnome-module.md
nielsdg: Correct. Should be fixed now!

tristan957:
Wondering if this would fit better in a Vala module. Is Vala trying to break out of the shadow of GNOME or will it always be tied to GNOME?

nielsdg:
I think the amount of projects using it outside of GNOME is probably negligible. Tbf, you _could_ move it into its own module since it's also one of the supported languages Meson, but I'm not sure if it's worth the effort

dcbaker:
Failures look relavent

nielsdg:
Failure should be fixed now :-)

lw64:
Any updates?

eli-schwartz:
The very first non-author comment is "LGTM but needs a release snippet" and there is still no release snippet.

lw64:
I see. there is a snippets folder.

eli-schwartz:
@nielsdg any update on a release notes snippet (and rebasing against master at the same time)?

lw64:
This just came up as an issue in the Vala repo: valadoc is also used to generate GIR files that contain the documentation, and it would be good to have an easy way to do that as well. Would it be hard to add an optional parameter  to install the GIR file?

Issue #7175, opened 2020-05-20
Unknown CPU family 'ip35' on IRIX 6.5.30


unxmaal:
Regarding this message: WARNING: Unknown CPU family 'ip35', please report this at https://github.com/mesonbuild/meson/issues/new with theoutput of `uname -a` and `cat /proc/cpuinfo`

`$ uname -a
IRIX64 yellow 6.5 07202013 IP35 mips IP35 Irix`

IRIX has no /proc/cpuinfo, but this can be derived from `hinv -v`. 

`hinv
4 700 MHZ IP35 Processors
CPU: MIPS R16000 Processor Chip Revision: 2.1
FPU: MIPS R16010 Floating Point Chip Revision: 2.1
Main memory size: 8192 Mbytes`

(Yes, Meson works on IRIX. No, this isn't a joke.)

jpakkane:
Since none of us has an Irix machine (that I know of), patches to fix this would be welcome. The change should be done in `detect_cpu_family` function in `environment.py`. It should return either `mips' or `mips64` depending on whether it is a 32 or 64 bit userland.

PR #7180, opened 2020-05-21
Updated to resolve issue identifying SGI CPUs on IRIX systems

unxmaal:
Resolves issue #7175

### DIFF
--- a/mesonbuild/envconfig.py
+++ b/mesonbuild/envconfig.py
@@ -64,7 +64,9 @@ known_cpu_families = (
     'wasm32',
     'wasm64',
     'x86',
-    'x86_64'
+    'x86_64',
+    'ip30',
+    'ip35'
 )
 
 # It would feel more natural to call this "64_BIT_CPU_FAMILES", but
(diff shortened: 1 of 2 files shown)


mesonbuild/envconfig.py
```diff
@@ -64,7 +64,9 @@
     'wasm32',
     'wasm64',
     'x86',
-    'x86_64'
+    'x86_64',
+    'ip30',
+    'ip35'
```
dcbaker, line 69: I don't think this is right, I think these should be caught and converted to mips or mips64. These are for cpu families, not specific cpus. Think x86 vs i486, i586, i686.

mesonbuild/envconfig.py
```diff
@@ -299,6 +301,10 @@ def is_hurd(self) -> bool:
         """
         return self.system == 'gnu'
 
+    def is_irix(self) -> bool:
+        """Machine is IRIX?"""
+        return self.system == 'irix64'
```
dcbaker, line 306: I know that irix64 is already pretty old, but maybe this should be `return self.system.startswith('irix')` just in case someone tries to use meson on an old iris 32bit system?

mesonbuild/envconfig.py
dcbaker: Unless I'm completely misremembering what was in the old SGI machines?

unxmaal: No, this is a good call. I'll fix it.

jpakkane: The `detect_cpu_family` function should be updated to handle these instead.

jpakkane:
ping

unxmaal:
Will have some time over the weekend to resolve.

PR #7181, opened 2020-05-21
Added ability to specify target in `meson compile`

### LINKED ISSUE #6740: [Feature request] Add ability to specify target(/s) to `meson compile`
Proposal:
Add ability to specify target(/s) to `meson compile`.

Reasoning:
- Removes most of the cases where direct invocation of `ninja`/`msbuild` is needed.
- (relative) Feature parity with `cmake --build`.

Proposed interface:
```
usage: meson compile [args [args...]]

positional arguments:
  args                            optional list of targets to build
```

Extracted from #6670.

TheQwertiest:
Implements and fixes #6740

Target name has the following syntax: `[PATH/]FOO[:TARGET_TYPE]`:
- `compile FOO` - works if there is only a single instance of FOO in introspection, errors out otherwise.
- `compile FOO:TARGET_TYPE` - works if there is only a single instance of FOO in introspection with TARGET_TYPE.
- `compile PATH/FOO:TARGET_TYPE` - always works

PATH - relative path from meson root to the target (e.g `./foo` or `subdir/subdir/foo`).
FOO - target name from meson syntax (e.g. `shared_library('foo', ...)`)
TARGET_TYPE - target type (`shared_library`, `static_library`, `jar`, `run` and etc).

Supported backends: msvs and ninja.
Limitations: 
- msvs: `run_target` is not supported

### DIFF
--- a/docs/markdown/Commands.md
+++ b/docs/markdown/Commands.md
@@ -136,30 +136,49 @@ meson configure builddir -Doption=new_value
 *(since 0.54.0)*
 
 ```
-$ meson compile [-h] [-j JOBS] [-l LOAD_AVERAGE] [--clean] [-C BUILDDIR]
+$ meson compile [-h] [--clean] [-C BUILDDIR] [-j JOBS] [-l LOAD_AVERAGE]
                 [--verbose] [--ninja-args NINJA_ARGS] [--vs-args VS_ARGS]
+                [TARGET [TARGET ...]]
 ```
 
 Builds a default or a specified target of a configured meson project.
 
 ```
+positional arguments:
+  TARGET                                Targets to build. Target has the
+                                        following format: [PATH_TO_TARGET/]TARGE
+                                        T_NAME[:TARGET_TYPE].
+
 optional arguments:
   -h, --help                            show this help message and exit
+  --clean                               Clean the build directory.
+  -C BUILDDIR                           The directory containing build files to
+                                        be built.
   -j JOBS, --jobs JOBS                  The number of worker jobs to run (if
                                         supported). If the value is less than 1
                                         the build program will guess.
   -l LOAD_AVERAGE, --load-average LOAD_AVERAGE
                                         The system load average to try to
-                                        maintain (if supported)
-  --clean                               Clean the build directory.
-  -C BUILDDIR                           The directory containing build files to
-                                        be built.
+                                        maintain (if supported).
   --verbose                             Show more verbose output.
   --ninja-args NINJA_ARGS               Arguments to pass to `ninja` (applied
                                         only on `ninja` backend).
   --vs-args VS_ARGS                     Arguments to pass to `msbuild` (applied
                                         only on `vs` backend).
 ```
+
+`--verbose` argument is available since 0.55.0.
+
+#### Targets
+
+*(since 0.55.0)*
+
+`TARGET` has the following syntax `[PATH/]NAME[:TYPE]`, where:
+- `NAME`: name of the target from `meson.build` (e.g. `foo` from `executable('foo', ...)`).
+- `PATH`: path to the target relative to the root `meson.build` file. Note: relative path for a target specified in the root `meson.build` is `./`.
+- `TYPE`: type of the target. Can be one of the following: 'executable', 'static_library', 'shared_library', 'shared_module', 'custom', 'run', 'jar'.
+  
+`PATH` and/or `TYPE` can be ommited if the resulting `TARGET` can be used to uniquely identify the target in `meson.build`.
 
 #### Backend specific arguments
 
@@ -191,6 +210,16 @@ meson compile -C builddir
 Execute a dry run on ninja backend with additional debug info:
 ```
 meson compile --ninja-args=-n,-d,explain
+```
+
+Build three targets: two targets that have the same `foo` name, but different type, and a `bar` target:
+```
+meson compile foo:shared_library foo:static_library bar
+```
+
+Produce a coverage html report (if available):
+```
+meson compile coverage-html
 ```
 
 ### dist
--- a/mesonbuild/mcompile.py
+++ b/mesonbuild/mcompile.py
@@ -14,9 +14,11 @@
 
 """Entrypoint script for backend agnostic compile."""
 
-import argparse
+import json
+import re
 import sys
 import typing as T
+from collections import defaultdict
 from pathlib import Path
 
 from . import mlog
@@ -26,10 +28,13 @@ from .mesonlib import MesonException
 from mesonbuild.environment import detect_ninja
 from mesonbuild.coredata import UserArrayOption
 
+if T.TYPE_CHECKING:
+    import argparse
+
 def array_arg(value: str) -> T.List[str]:
     return UserArrayOption(None, value, allow_dups=True, user_input=True).value
 
-def validate_builddir(builddir: Path):
+def validate_builddir(builddir: Path) -> None:
     if not (builddir / 'meson-private' / 'coredata.dat' ).is_file():
         raise MesonException('Current directory is not a meson build directory: `{}`.\n'
                              'Please specify a valid build dir or change the working directory to it.\n'
@@ -42,7 +47,93 @@ def get_backend_from_coredata(builddir: Path) -> str:
     """
     return coredata.load(str(builddir)).get_builtin_option('backend')
 
-def get_parsed_args_ninja(options: 'argparse.Namespace', builddir: Path):
+def parse_introspect_data(builddir: Path) -> T.Dict[str, T.List[dict]]:
+    """
+    Converts a List of name-to-dict to a dict of name-to-dicts (since names are not unique)
+    """
+    path_to_intro = builddir / 'meson-info' / 'intro-targets.json'
+    if not path_to_intro.exists():
+        raise MesonException('`{}` is missing! Directory is not configured yet?'.format(path_to_intro.name))
+    with path_to_intro.open() as f:
+        schema = json.load(f)
+
+    parsed_data = defaultdict(list) # type: T.Dict[str, T.List[dict]]
+    for target in schema:
+        parsed_data[target['name']] += [target]
+    return parsed_data
+
+class ParsedTargetName:
+    full_name = ''
+    name = ''
+    type = ''
+    path = ''
+
+    def __init__(self, target: str):
+        self.full_name = target
+        split = target.rsplit(':', 1)
+        if len(split) > 1:
+            self.type = split[1]
+            if not self._is_valid_type(self.type):
+                raise MesonException('Can\'t invoke target `{}`: unknown target type: `{}`'.format(target, self.type))
+
+        split = split[0].rsplit('/', 1)
+        if len(split) > 1:
+            self.path = split[0]
+            self.name = split[1]
+        else:
+            self.name = split[0]
+
+    @staticmethod
+    def _is_valid_type(type: str) -> bool:
+        # Ammend docs in Commands.md when editing this list
+        allowed_types = {
+            'executable',
+            'static_library',
+            'shared_library',
+            'shared_module',
+            'custom',
+            'run',
+            'jar',
+        }
+        return type in allowed_types
+
+def get_target_from_intro_data(target: ParsedTargetName, builddir: Path, introspect_data: dict) -> dict:
+    if target.name not in introspect_data:
+        raise MesonException('Can\'t invoke target `{}`: target not found'.format(target.full_name))
+
+    intro_targets = introspect_data[target.name]
+    found_targets = []
+
+    resolved_bdir = builddir.resolve()
+
+    if not target.type and not target.path:
+        found_targets = intro_targets
+    else:
+        for intro_target in intro_targets:
+            if (intro_target['subproject'] or
+                    (target.type and target.type != intro_target['type'].replace(' ', '_')) or
+                    (target.path
+                         and intro_target['filename'] != 'no_name'
+                         and Path(target.path) != Path(intro_target['filename'][0]).relative_to(resolved_bdir).parent)):
+                continue
+            found_targets += [intro_target]
+
+    if not found_targets:
+        raise MesonException('Can\'t invoke target `{}`: target not found'.format(target.full_name))
+    elif len(found_targets) > 1:
+        raise MesonException('Can\'t invoke target `{}`: ambigious name. Add target type and/or path: `PATH/NAME:TYPE`'.format(target.full_name))
+
+    return found_targets[0]
+
+def generate_target_names_ninja(target: ParsedTargetName, builddir: Path, introspect_data: dict) -> T.List[str]:
+    intro_target = get_target_from_intro_data(target, builddir, introspect_data)
+
+    if intro_target['type'] == 'run':
+        return [target.name]
+    else:
+        return [str(Path(out_file).relative_to(builddir.resolve())) for out_file in intro_target['filename']]
+
+def get_parsed_args_ninja(options: 'argparse.Namespace', builddir: Path) -> T.List[str]:
     runner = detect_ninja()
     if runner is None:
         raise MesonException('Cannot find ninja.')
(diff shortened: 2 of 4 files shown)


TheQwertiest:
CI failures do not seem relevant: all failed configurations failed on the "Update MSYS" step

mesonbuild/mcompile.py
```diff
@@ -27,9 +29,45 @@
 if T.TYPE_CHECKING:
     import argparse
 
+def generate_ninja_target_name(target_name: str, builddir: Path, introspect_data: dict) -> str:
+    for target in introspect_data:
+        if (not target['subproject'] and target['name'] == target_name):
```
dcbaker, line 34: please don't wrap if statements in parens unless they span multiple lines, this is python not C :)

mesonbuild/mcompile.py
```diff
@@ -27,9 +29,45 @@
...
+            if (target_type == 'run'):
+                return target_name
+            elif (target_type == 'custom'):
+                # Custom targets do not have a clear mapping to `ninja` targets
+                raise MesonException('Can\'t invoke target `{}`: target of type `{}` can\'t be invoked directly on this backend'.format(target_name, target_type))
+            else:
+                if (target['filename'] == 'no_name'):
+                    raise MesonException('Can\'t invoke target `{}`: unexpected target configuration: missing filename'.format(target_name))
+                 
+                return str(Path(target['filename'][0]).relative_to(builddir.resolve()))
```
dcbaker, line 45: shouldn't this be in the form:
```python
elif target['filename'] == 'no_name':
     raise ...
else:
     return ...
```

mesonbuild/mcompile.py
```diff
@@ -83,14 +120,22 @@ def run(options: 'argparse.Namespace') -> int:
 
         cmd = [runner, '-C', bdir.as_posix()]
 
+        if options.targets and options.clean:
+            raise MesonException('`target` and `clean` can\'t be used simultaneously')
```
dcbaker, line 124: argparse has builtin functionality for this: https://docs.python.org/3/library/argparse.html#mutual-exclusion

mesonbuild/mcompile.py
```diff
@@ -83,14 +120,22 @@ def run(options: 'argparse.Namespace') -> int:
...
+        if options.targets:
+            with (bdir/'meson-info'/'intro-targets.json').open() as f:
```
dcbaker, line 126: `/` is an operator, and we always use spaces around operators.

mesonbuild/mcompile.py
```diff
@@ -83,14 +120,22 @@ def run(options: 'argparse.Namespace') -> int:
...
+                schema = json.load(f)
+            for target in options.targets:
+                cmd.extend([generate_ninja_target_name(target, bdir, schema)])
```
dcbaker, line 129:
```suggestion
            cmd.extend([generate_ninja_target_name(t, bdir, schema) for t in options.targets])
```

mesonbuild/mcompile.py
```diff
@@ -105,11 +150,19 @@ def run(options: 'argparse.Namespace') -> int:
...
 
-        if options.load_average:
-            mlog.warning('Msbuild does not have a load-average switch, ignoring.')
+        if options.targets and options.clean:
+            raise MesonException('`target` and `clean` can\'t be used simultaneously')
+        if options.targets:
+            with Path(bdir/'meson-info'/'intro-targets.json').open() as f:
+                schema = json.load(f)
+            for target in options.targets:
+                cmd.extend(['/t:{}'.format(generate_msvs_target_name(target, bdir, schema))])
```
dcbaker, line 159:
```suggestion
            cmd.extend(['/t:{}'.format(generate_msvs_target_name(t, bdir, schema)) for t in options.targets])
```

TheQwertiest:
Thanks for the review! As you might've guessed I'm not very experienced in Python ^^"...

mesonbuild/mcompile.py
TheQwertiest: Well, I prefer making a clear distinction between the purpose of each if\else group.
E.g. I would've used `elif` here if it was handling the same condition parameter like this:
```
if target.name == 'a':
  ...
elif target.name == 'b':
  ...
```
I can change it if you want though, since it's just a style choice.

mesonbuild/mcompile.py
TheQwertiest: Thanks for the tip! Fixed

mesonbuild/mcompile.py

mesonbuild/mcompile.py

mesonbuild/mcompile.py

TheQwertiest:
@dcbaker , what do you think about replacing backend divination code with introspection data parsing? This should make backend detection more robust.

mesonbuild/mcompile.py
TheQwertiest: This results in `error: argument TARGET: not allowed with argument --clean`, when invoking like this: `meson compile -C dir --clean`.
It seems that `TARGET` is always defined as present even it's just an empty default value.

TheQwertiest:
@dcbaker , seems I'll have to rework this PR a bit due to the new info from @jpakkane (https://github.com/mesonbuild/meson/issues/6740#issuecomment-632719896)

mesonbuild/mcompile.py
dcbaker: hmmmm, I'm not sure off the top of my head. It may not be possible to make a switch argument and positional argument mutally exclusive. You may need to either drop my suggestion or use something like `-t/--target=foo`.

I'm fine with whichever way you want to go. Though if decide to not use the mutually exclusive arguments, can we move the argument exclusive checking into common code so each backend function doesn't need to implement it?

mesonbuild/mcompile.py
dcbaker: I guess either way then

TheQwertiest:
Reworked PR a bit: 
- Added handling of non-unique names by adding the following syntax to target name: https://github.com/mesonbuild/meson/issues/6740#issuecomment-632725990
- Fixed target execution in subdirs on VS backend.
- Converted parsed introspection data to dict from array to avoid needless data iterations.

mesonbuild/mcompile.py
TheQwertiest: Removed `mutual_exclusion` and extracted the check to the common section.

TheQwertiest:
How do I fix the typing error? 
https://github.com/mesonbuild/meson/pull/7181/checks?check_run_id=700324080#step:5:8
If I add typing after the variable name like this: `var: typing = ...`, it will fail on another agent with lower Python version :\

TheQwertiest:
Test failures seems to be unrelated

TheQwertiest:
@dcbaker , Thansk!
PS: Force pushes are for test-fixing only (trying to divine output library names...), i.e. no changes to mcompile.
PPS: Should I squish/clean the PR?

TheQwertiest:
Rebased to remove conflicts

jpakkane:
Needs a release note snippet and the documentation needs to be updated, at least the way of defining build target names.

TheQwertiest:
> Needs a release note snippet

Will be done.

> documentation needs to be updated

There is no documentation for `meson compile` at all.
I was planning on making a separate `Meson commands` page in a separate PR, that lists all commands with short description and links to the respective pages (if any).

TheQwertiest:
Added a release snippet.
The proper documentation page for all commands will be added in a separate PR (if you allow me to =)).

TheQwertiest:
Wait! I've just noticed that target types from introspection have different names from targets in `meson.build` (e.g. `custom_target` in meson.build vs `custom` in introspection). Should I use those instead? If so, than should there be a mapping for meson.build target types such as `library`, `both_libraries` or `build_target`?

TheQwertiest:
@jpakkane , https://github.com/mesonbuild/meson/pull/7217 contains documentation for `meson compile` and this PR (among other things).

nioncode:
> 
> Limitations:
>    *  ninja: custom_target is not supported
>    *  msvs: run_target is not supported
> 

@TheQwertiest why is this the case, i.e. what's the problem with supporting those? It would be nice to have full support for all targets.

TheQwertiest:
@nioncode , I agree, but it's not possible (yet?):
See the following comments:
https://github.com/mesonbuild/meson/blob/547e29718b67d6b1fc22d42e846fee8bc106f475/mesonbuild/mcompile.py#L185
https://github.com/mesonbuild/meson/blob/547e29718b67d6b1fc22d42e846fee8bc106f475/mesonbuild/mcompile.py#L176

nioncode:
Regarding msbuild: I'm fairly certain that you can simply build the `vcxproj` directly with `msbuild` as I have done in my old PR: https://github.com/nioncode/meson/blob/8008cc97ee8a7a02b393dc62e7513d6be6b9ffc8/mesonbuild/mcompile.py#L58

Regarding ninja: isn't this what this PR is there for and should thus provide a way to build a custom_target? At least in my cmake projects I have various custom_target/run_target like commands that I execute with `cmake --build` all the time. Is there anything really preventing us from mapping a target name to the correct ninja commands to invoke? I haven't touched meson code for over a year, so I might be completely wrong, but I think it should be possible to know the commands to run for each target, otherwise the build.ninja file could not be generated.

TheQwertiest:
> I'm fairly certain that you can simply build the vcxproj directly with msbuild

Yes, but that won't build project dependencies (AFAIK). It will build only the specified project,

> Regarding ninja: isn't this what this PR is there for and should thus provide a way to build a custom_target? 

This PR is for all targets not just custom_target. Custom target is just the odd man out.

Regarding custom target: I've checked the resulting ninja file and there is nothing for the custom target itself. Only for the files it produces, So to make it work one has to gather all custom_target's output and convert them to ninja targets.

nioncode:
For my use case (which always involves building on Windows with msbuild + Linux with ninja), this basically means that mcompile can not be used for either `run_target` or `custom_target` (because either of those will fail on one of my platforms), which is unfortunate. I think we should go the extra mile here and gather the required dependencies for vs and invoke them in addition to the `run_target` (or you sure that the dependencies are not being run then? I really can't remember from my PR, but I think those properly work with the vcxproj that we generate in the vs backend) and doing something similar for ninja and `custom_target`.

Does mcompile support `alias_target`s? If this works, then this might be good enough, since we can suggest to wrap `run_target` and `custom_target` in an `alias_target` then.

TheQwertiest:
Regarding msvs, I stand corrected: I've just retested and it seems to work even when targeting individual projects. I swear it was not working before! >_<
Maybe something there was a bug in vs backend deps formation somewhere, i.e. it was not a vs problem but rather a problem with vcxproj generation.

TheQwertiest:
There is another problem though: manual vcxproj invocation requires a separate msbuild run. I.e. you can't invoke `shared_library` target and `run_target` via a single msbuild invocation.

TheQwertiest:
Added support for custom_targets in ninja backend: it was actually quite simple. My bad.

nioncode:
Looks great.

> There is another problem though: manual vcxproj invocation requires a separate msbuild run. I.e. you can't invoke shared_library target and run_target via a single msbuild invocation. It will also make all run_target builds sequential. And this will require a total rewrite of mcompile.

Couldn't this be solved by running all `run_target`s in sequence before doing the 'final' msbuild call that builds everything else (if there is a target remaining)? This shouldn't be too hard, since it would basically be:
```python
for t in run_targets:
  run_msbuild(t)
run_msbuild(remaining_targets)
```

Maybe it is also possible to list multiple vcxproj on the command line (this would allow to collapse the msbuild invocations into a maximum of 2 calls), but I'm not sure.

TheQwertiest:
> Maybe it is also possible to list multiple vcxproj on the command line (this would allow to collapse the msbuild invocations into a maximum of 2 calls), but I'm not sure.

This's not possible =(
They even have a specific error message for that (smth like "only one project can be specified").

> Couldn't this be solved by running all run_targets in sequence before doing the 'final' msbuild call that builds everything else (if there is a target remaining)?

It will still require:
- a total separation of vs and ninja invocation logic.
- a total rewrite of vs logic.

And it creates another potential problem (if/when https://github.com/mesonbuild/meson/pull/7183 is merged): user can no longer deduce what's actually happening inside, so backend args that he passes might not be applicable anymore if he specifies even a single run_target (since it might be passed not to a single `msbuild .sln` invocation, but to multiple `msbuild .vcxproj` invocations as well).

PS: Yes, these are all excuses for me not to do extra work =)

nioncode:
Maybe it is good enough for now to simply say: if a `run_target` is specified (for vs only), no other targets are allowed? Then at least it is possible to run each target one after another.

I don't really understand your points about the backend_args PR though. If I supply backend_args, I'd assume that they apply to every msbuild invocation that is being done. If I don't want to build everything with the same settings, I'd do multpile `meson compile` invocations.

jpakkane:
> Maybe it is good enough for now to simply say: if a run_target is specified (for vs only), no other targets are allowed? Then at least it is possible to run each target one after another.

Fine by me.

TheQwertiest:
@jpakkane , all done. The only question remaining is `TARGET_TYPE` naming: currently intro target types are used, which are not the same as meson target types (differences in **bold**):

current (based on introspection):
- 'executable',
- 'static_library',
- 'shared_library',
- 'shared_module',
- 'custom',
- 'run',
- 'jar',

meson.build:
- 'executable',
- 'static_library',
- 'shared_library',
- 'shared_module',
- **'custom_target',**
- **'run_target',**
- 'jar',
- **'library'**,
- **'both_libraries'**,
- **'build_target'**,

TheQwertiest:
One solution would be just to list all intro-style types in docs.

docs/markdown/Commands.md
```diff
@@ -132,32 +132,50 @@ meson configure builddir -Doption=new_value
...
+                [TARGET [TARGET ...]]
 ```
 
 Builds a default or a specified target of a configured meson project.
 
 ```
+positional arguments:
+  TARGET                                Targets to build. Target has the
+                                        following format: [PATH_TO_TARGET/]TARGE
+                                        T_NAME[:TARGET_TYPE]
```
dcbaker, line 145: You should document that target is new in 0.55.0

mesonbuild/mcompile.py
```diff
@@ -40,61 +43,184 @@ def get_backend_from_coredata(builddir: Path) -> str:
     """
     return coredata.load(str(builddir)).get_builtin_option('backend')
 
-def get_parsed_args_ninja(options: 'argparse.Namespace', builddir: Path):
+def parse_introspect_data(builddir: Path) -> dict:
```
dcbaker, line 46: Please make the returned type match the type of `parsed_data`

mesonbuild/mcompile.py
```diff
@@ -40,61 +43,184 @@ def get_backend_from_coredata(builddir: Path) -> str:
...
+        # Ammend docs in Commands.md when editing this list
+        allowed_types = [
+            'executable',
+            'static_library',
+            'shared_library',
+            'shared_module',
+            'custom',
+            'run',
+            'jar',
+        ]
```
dcbaker, line 93: you should use a set rather than a list here, since order doesn't matter and you're just doing membership tests.

docs/markdown/Commands.md
TheQwertiest: Fixed

mesonbuild/mcompile.py

mesonbuild/mcompile.py

TheQwertiest:
@dcbaker , what do you think about target type naming discrepancy?

dcbaker:
You mean the meson may generate conflicting target names and your solution of `NAME[:TYPE]`?

TheQwertiest:
I mean https://github.com/mesonbuild/meson/pull/7181#issuecomment-644069430 =)
I.e. (intro target type name) != (meson.build target type name)

dcbaker:
The only concern I have about that is that introspect is listing what the build_target *is* (shared_library, static_library, etc), but meson.build is listing what it *could be*. that means that the values passed to meson compile could change based on your configuration, which seems less than ideal. ie,
```console
meson setup build -Ddefault_library=static
meson compile build librarytarget:shared_library
```
could fail to compile because `librarytarget` is a `static_library`

I'm not really concerned about the `run` vs `run_target` differences.

TheQwertiest:
> The only concern I have about that is that introspect is listing what the build_target is

Right, and it doesn't look like I can check if it was specified as `both_libraries` in `meson.build` (not through introspection files anyway).

> I'm not really concerned about the run vs run_target differences.

Which was actually my main concern :D

TheQwertiest:
Rebased and squashed (since it's a huge PITA to rebase otherwise)

TheQwertiest:
@dcbaker , is there anything else I need to fix to make it mergeable?

jpakkane:
⚔️ conflicts.

TheQwertiest:
Rebased

TheQwertiest:
PS: Not sure how to fix CI failure - it seems that it can't find `meson` binary on Macos and Windows agents. Is there a way to require python packages from tests?

jpakkane:
Let's merge this with the caveat that if it turns out in actual usage that there is a major design flaw, we can do a breaking fix in the next release.

PR #7182, opened 2020-05-21
compilers: add fetching of define list for clang

sigexp:
Simmilar to gcc, the list of pre-processor defines can be fetched with
`-dM -E` option. The way cpu_family is determined on linux relies on
this list.

Fixes incorrect value of cpu_family on linux, when crosscompiling:

```
CC="clang -m32" meson ./build
```

Signed-off-by: Yevhenii Kolesnikov <yevhenii.kolesnikov@globallogic.com>

### DIFF
--- a/mesonbuild/compilers/c.py
+++ b/mesonbuild/compilers/c.py
@@ -86,9 +86,10 @@ class ClangCCompiler(ClangCompiler, CCompiler):
     _C18_VERSION = '>=8.0.0'
 
     def __init__(self, exelist, version, for_machine: MachineChoice,
-                 is_cross, info: 'MachineInfo', exe_wrapper=None, **kwargs):
+                 is_cross, info: 'MachineInfo', exe_wrapper=None,
+                 defines: T.Optional[T.List[str]] = None, **kwargs):
         CCompiler.__init__(self, exelist, version, for_machine, is_cross, info, exe_wrapper, **kwargs)
-        ClangCompiler.__init__(self)
+        ClangCompiler.__init__(self, defines)
         default_warn_args = ['-Wall', '-Winvalid-pch']
         self.warn_args = {'0': [],
                           '1': default_warn_args,
--- a/mesonbuild/compilers/mixins/clang.py
+++ b/mesonbuild/compilers/mixins/clang.py
@@ -42,9 +42,10 @@ clang_optimization_args = {
 }  # type: T.Dict[str, T.List[str]]
 
 class ClangCompiler(GnuLikeCompiler):
-    def __init__(self):
+    def __init__(self, defines: T.Optional[T.Dict[str, str]]):
         super().__init__()
         self.id = 'clang'
+        self.defines = defines or {}
         self.base_options.append('b_colorout')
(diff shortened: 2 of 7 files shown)


sigexp:
This fixes the issue reported in Mesa:
https://gitlab.freedesktop.org/mesa/mesa/-/issues/3020

sigexp:
> This pull request **introduces 6 alerts** when merging [6f9fb2f](https://github.com/mesonbuild/meson/commit/6f9fb2fb4cacf6bfe0528c073f29397970e85a12) into [29ef447](https://github.com/mesonbuild/meson/commit/29ef4478df6d3aaca40c7993f125b29409be1de2) - [view on LGTM.com](https://lgtm.com/projects/g/mesonbuild/meson/rev/pr-ccb275ecdb43a45dcd58163bb5d970eee69ff673)
> 
> **new alerts:**
> 
>     * 3 for Wrong number of arguments in a call
> 
>     * 3 for Wrong number of arguments in a class instantiation

Oh, I forgot about Fortran and Obj-C. I don't know much about those languages, so for now, for those compilers I just added a placeholder, supplying an empty list instead.

mesonbuild/compilers/mixins/clang.py
```diff
@@ -42,9 +42,10 @@
 }  # type: T.Dict[str, T.List[str]]
 
 class ClangCompiler(GnuLikeCompiler):
-    def __init__(self):
+    def __init__(self, defines: T.Dict[str, str]):
```
dcbaker, line 45:
```suggestion
    def __init__(self, defines: T.Optional[T.Dict[str, str]]):
```

Since this can be `None`, right?

mesonbuild/compilers/mixins/clang.py
```diff
@@ -56,6 +57,14 @@ def __init__(self):
     def get_colorout_args(self, colortype: str) -> T.List[str]:
         return clang_color_args[colortype][:]
 
+    def has_builtin_define(self, define: str) -> bool:
+        return define in self.defines
+
+    def get_builtin_define(self, define: str) -> T.Optional[str]:
+        if define in self.defines:
+            return self.defines[define]
+        return None
```
dcbaker, line 66:
```suggestion
        return self.defines.get(define)
```

Same thing, just less code.

mesonbuild/compilers/c.py
```diff
@@ -86,9 +86,10 @@ class ClangCCompiler(ClangCompiler, CCompiler):
     _C18_VERSION = '>=8.0.0'
 
     def __init__(self, exelist, version, for_machine: MachineChoice,
-                 is_cross, info: 'MachineInfo', exe_wrapper=None, **kwargs):
+                 is_cross, info: 'MachineInfo', exe_wrapper=None,
+                 defines=None, **kwargs):
```
dcbaker, line 90:
```suggestion
                 defines: T.Optional[T.List[str]] = None, **kwargs):
```

mesonbuild/compilers/mixins/clang.py
sigexp: Yes, this looks better, thanks.

PR #7183, opened 2020-05-21
Added ability to pass arguments to backend in `meson compile`

TheQwertiest:
Adds ability to pass arguments directly to backend build tool:
```
usage: meson compile [--vs-args=args] [--ninja-args=args]
```

These arguments  use the following syntax:

If you only pass a single string, then it is considered to have all values separated by commas. Thus invoking the following command:

```
$ meson compile --ninja-args=foo,bar
```

would set the value to an array of two elements, foo and bar.

If you need to have commas in your string values, then you need to pass the value with proper shell quoting like this:

```
$ meson compile "--ninja-args=['a,b', 'c,d']"
```

### DIFF
--- a/docs/markdown/Commands.md
+++ b/docs/markdown/Commands.md
@@ -137,7 +137,7 @@ meson configure builddir -Doption=new_value
 
 ```
 $ meson compile [-h] [-j JOBS] [-l LOAD_AVERAGE] [--clean] [-C BUILDDIR]
-                [--verbose]
+                [--verbose] [--ninja-args NINJA_ARGS] [--vs-args VS_ARGS]
 ```
 
 Builds a default or a specified target of a configured meson project.
@@ -155,6 +155,30 @@ optional arguments:
   -C BUILDDIR                           The directory containing build files to
                                         be built.
   --verbose                             Show more verbose output.
+  --ninja-args NINJA_ARGS               Arguments to pass to `ninja` (applied
+                                        only on `ninja` backend).
+  --vs-args VS_ARGS                     Arguments to pass to `msbuild` (applied
+                                        only on `vs` backend).
+```
+
+#### Backend specific arguments
+
+*(since 0.55.0)*
+
+`BACKEND-args` use the following syntax:
+
+If you only pass a single string, then it is considered to have all values separated by commas. Thus invoking the following command:
+
+```
+$ meson compile --ninja-args=-n,-d,explain
+```
+
+would add `-n`, `-d` and `explain` arguments to ninja invocation.
+
+If you need to have commas or spaces in your string values, then you need to pass the value with proper shell quoting like this:
+
+```
+$ meson compile "--ninja-args=['a,b', 'c d']"
 ```
 
 #### Examples:
@@ -162,6 +186,11 @@ optional arguments:
 Build the project:
 ```
 meson compile -C builddir
+```
+
+Execute a dry run on ninja backend with additional debug info:
+```
+meson compile --ninja-args=-n,-d,explain
 ```
 
 ### dist
--- a/docs/markdown/snippets/add_compile_backend_arg.md
+++ b/docs/markdown/snippets/add_compile_backend_arg.md
@@ -1,0 +1,26 @@
+## Added ability to specify backend arguments in `meson compile`
+
+It's now possible to specify backend specific arguments in `meson compile`.
+
+Usage: `meson compile [--vs-args=args] [--ninja-args=args]`
+
+```
+  --ninja-args NINJA_ARGS    Arguments to pass to `ninja` (applied only on `ninja` backend).
+  --vs-args VS_ARGS          Arguments to pass to `msbuild` (applied only on `vs` backend).
+```
+
+These arguments use the following syntax:
+
+If you only pass a single string, then it is considered to have all values separated by commas. Thus invoking the following command:
+
+```
+$ meson compile --ninja-args=-n,-d,explain
+```
+
+would add `-n`, `-d` and `explain` arguments to ninja invocation.
+
+If you need to have commas or spaces in your string values, then you need to pass the value with proper shell quoting like this:
+
+```
+$ meson compile "--ninja-args=['a,b', 'c d']"
+```
--- a/mesonbuild/mcompile.py
+++ b/mesonbuild/mcompile.py
@@ -14,6 +14,7 @@
 
 """Entrypoint script for backend agnostic compile."""
 
+import argparse
 import sys
 import typing as T
 from pathlib import Path
@@ -23,10 +24,11 @@ from . import mesonlib
 from . import coredata
 from .mesonlib import MesonException
 from mesonbuild.environment import detect_ninja
+from mesonbuild.coredata import UserArrayOption
 
-if T.TYPE_CHECKING:
-    import argparse
-    
+def array_arg(value: str) -> T.List[str]:
+    return UserArrayOption(None, value, allow_dups=True, user_input=True).value
+
 def validate_builddir(builddir: Path):
     if not (builddir / 'meson-private' / 'coredata.dat' ).is_file():
         raise MesonException('Current directory is not a meson build directory: `{}`.\n'
@@ -58,31 +60,31 @@ def get_parsed_args_ninja(options: 'argparse.Namespace', builddir: Path):
         cmd.append('-v')
     if options.clean:
         cmd.append('clean')
-    
+
     return cmd
 
 def get_parsed_args_vs(options: 'argparse.Namespace', builddir: Path):
     slns = list(builddir.glob('*.sln'))
     assert len(slns) == 1, 'More than one solution in a project?'
-    
+
     sln = slns[0]
     cmd = ['msbuild', str(sln.resolve())]
-    
+
     # In msbuild `-m` with no number means "detect cpus", the default is `-m1`
     if options.jobs > 0:
         cmd.append('-m{}'.format(options.jobs))
     else:
         cmd.append('-m')
-    
+
     if options.load_average:
         mlog.warning('Msbuild does not have a load-average switch, ignoring.')
     if not options.verbose:
         cmd.append('/v:minimal')
     if options.clean:
         cmd.append('/t:Clean')
-    
+
     return cmd
-    
+
 def add_arguments(parser: 'argparse.ArgumentParser') -> None:
     """Add compile specific arguments."""
     parser.add_argument(
@@ -117,7 +119,18 @@ def add_arguments(parser: 'argparse.ArgumentParser') -> None:
         action='store_true',
         help='Show more verbose output.'
     )
-
+    parser.add_argument(
+        '--ninja-args',
+        type=array_arg,
+        default=[],
+        help='Arguments to pass to `ninja` (applied only on `ninja` backend).'
+    )
+    parser.add_argument(
+        '--vs-args',
+        type=array_arg,
+        default=[],
+        help='Arguments to pass to `msbuild` (applied only on `vs` backend).'
+    )
 
 def run(options: 'argparse.Namespace') -> int:
     bdir = options.builddir  # type: Path
@@ -128,8 +141,10 @@ def run(options: 'argparse.Namespace') -> int:
     backend = get_backend_from_coredata(bdir)
     if backend == 'ninja':
         cmd = get_parsed_args_ninja(options, bdir)
+        cmd += options.ninja_args
     elif backend.startswith('vs'):
         cmd = get_parsed_args_vs(options, bdir)
+        cmd += options.vs_args
     else:
         # TODO: xcode?
         raise MesonException(
--- a/run_unittests.py
+++ b/run_unittests.py
@@ -4637,12 +4637,27 @@ recommended as it is not supported on some platforms''')
 
         testdir = os.path.join(self.common_test_dir, '1 trivial')
         self.init(testdir)
+
         self._run([*self.meson_command, 'compile', '-C', self.builddir])
         # If compile worked then we should get a program
         self.assertPathExists(os.path.join(self.builddir, prog))
 
         self._run([*self.meson_command, 'compile', '-C', self.builddir, '--clean'])
         self.assertPathDoesNotExist(os.path.join(self.builddir, prog))
+
+        # `--$BACKEND-args`
+
+        if self.backend is Backend.ninja:
+            self.init(testdir, extra_args=['--wipe'])
+            # Dry run - should not create a program
+            self._run([*self.meson_command, 'compile', '-C', self.builddir, '--ninja-args=-n'])
+            self.assertPathDoesNotExist(os.path.join(self.builddir, prog))
+        elif self.backend is Backend.vs:
+            self.init(testdir, extra_args=['--wipe'])
+            self._run([*self.meson_command, 'compile', '-C', self.builddir])
+            # Explicitly clean the target through msbuild interface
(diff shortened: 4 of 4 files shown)


dcbaker:
I'm really not crazy about this. Why have a backend agnostic interface if you're just going to pass backend specific arguments anyway?

TheQwertiest:
@dcbaker , because backends often provides features that are unique to those backends. And if some user will need this unique feature once, he'll have to forfeit the use of backend-agnostic method completely.

Simple example: ninja's build output is very brief and it's not enough for proper build logs (e.g. on CI), this can be fixed by using `--verbose` flag. But neither xcode nor msvs need this flag, because their default output is verbose enough (includes all compile commands). So, if I want to use `meson` with `ninja` backend, I'll always have to use `ninja` directly, even if it's because of a single flag.

Someone can object by saying, "why would you use backend-agnostic method then, if you already have to specify backend-specific logic anyway?". Because backend-agnostic methods are not just `ninja -C` or `msbuild -C`, it contains other logic for hiding backend-specific interface behind backend-agnostic interface: targets, multithread/multiprocess build, clean command and maybe more in the future. I mean it's not even possible to invoke `msbuild -C builddir`  directly, user has to find the `.sln` first.

By limiting the user from passing *any* backend-specific arguments we are forcing them to replicate the whole `meson compile` logic manually (if they want to support multiple backends).

TheQwertiest:
PS: I don't know why `"-n"` in `_run([*self.meson_command, 'compile', '-C', self.builddir, '--backend_arg="-n"'])` is interpreted as a separate positional argument... 
Any help in solving this would be greatly appreciated.

dcbaker:
We already allow passing options that are commonly useful but not supported by all backends. For example, we have a load-average, which msbuild doesn't support, only ninja does. I'd be perfectly happy to have a --verbose/-v switch that msbuild ignores but ninja uses. At some point though if you want to get into the nitty options of each backend you just need to call that backend directly. calling `msbuild /opt /otheropt /thing foo` is much nicer than `meson compile --backend-opt="/opt" --backend-opt="/otheropt" --backend-opt="/thing" foo`

TheQwertiest:
> For example, we have a load-average, which msbuild doesn't support, only ninja does. 

And I don't think such options should've been added at all :P

> I'd be perfectly happy to have a --verbose/-v switch that msbuild ignores but ninja uses.

This would be really confusing, since msbuild has a verbose output option as well, but it's reeeeeeeeally verbose (like outputting internal build tool information or listing all the search paths during linking stage and etc). So it would be a `verbose` option that applies *only* to ninja, despite msbuild having the same option in it's interface.

> msbuild /opt /otheropt /thing foo is much nicer than meson compile --backend-opt="/opt" --backend-opt="/otheropt" --backend-opt="/thing" foo

That's exactly what I'm talking about! In the very basic scenario it would *not* be just `msbuild /opt`, it would be at least:
```
slns = list(bulddir.glob('*.sln'))
exec(f'msbuild -C{builddir} {slns[0]} /opt')
```
And that's without even talking about target functionality, which will require fetching and parsing introspection data and target name deduction, i.e. at the very least least this non-trivial method:
```
def generate_msvs_target_name(target_name: str, builddir: Path) -> str:
    with (builddir / 'meson-info' / 'intro-targets.json').open() as f:
        introspect_data= json.load(f)
    for target in introspect_data:
        if not target['subproject'] and target['name'] == target_name:
            target_type = target['type']
            if target_type == 'run':
                raise MyException('Can\'t invoke target `{}`: target of type `{}` can\'t be invoked directly on this backend'.format(target_name, target_type))
targets-in-solutions-by-using-msbuild-exe
            return re.sub('[\%\$\@\;\.\(\)\']', '_', target['id'])

    raise MyException('Can\'t invoke target `{}`: target not found'.format(target_name))

def generate_ninja_target_name(target_name: str, builddir: Path, introspect_data: dict) -> str:
    ....
```

dcbaker:
I don't think we're going to get any further, I'll let @jpakkane or @nirbheek chime in. I think the meson compile should neciessarily only work in the simple cases through clean, obvious abstractions, and when you want to do something really backend specific you should just call the backend directly. But maybe they'll disagree with me.

TheQwertiest:
Removed dependency on #7181 for brevity.

dcbaker:
I've thought about this some more, and I can go for the idea of passing arguments to the backend, but I really dislike the interface. Could we go for a more posix like interface, ala:
```console
meson compile -j5 -- --verbose
```
argparse has support for this using `argparse.REMAINDER`

TheQwertiest:
Sure! I will look into it.

TheQwertiest:
Rebased to remove conflicts

TheQwertiest:
@dcbaker , regretfully, there are problems when integrating *argparse.remainder* approach with positionals from https://github.com/mesonbuild/meson/pull/7181 :
`argparse` is kinda bugged - when it parses the command line it works like this (when talking about optional args):
1. Parse options that start with `-`/`--`. If option is without dashes goto 2.
2. Parse all positionals. If dashed option is encountered goto 1.

The problem is that `argparse.REMAINDER` is considered to be a positional and on step 2 it consumes *all* remaining arguments (whether dashed options or not). So for example the following parser:
```py
parser.add_argument('pos',nargs='*')
parser.add_argument('--opt', action='store', default=0,type=int)
parser.add_argument('remainder', nargs=argparse.REMAINDER)
```
will parse the line `pos1 pos2 pos3 --opt=2 -- other args` as 
```py
{
  'pos': ['pos1','pos2','pos3'],
  'opt': 0,
  'remainder': ['--opt=2', '--', 'other', 'args']
}
```

I've tried googling, but it seems there is no way to make it work =(

TheQwertiest:
This could be theoretically fixed by extending `def run(options: 'argparse.Namespace')` and parent methods. I.e. parse `--` in the root `argparse` and pass the remainder separately as an additional argument to `run()`. Not sure if it's worth it though.

TheQwertiest:
Made `backend_arg` an explicit argument, because of the problem mentioned above (which can't be solved without changing the parent argument handler code).

PS: I've discovered, that meson supports arrays in cmd invocations, so it's not *that* bad:
`meson compile --backend-arg=[--arg1=a, --arg2=b]`

jpakkane:
> meson compile --backend-arg=['--arg1=a', '--arg2=b']

Errr, how can that work? The shell should swallow the brackets AFAICT.

TheQwertiest:
Because typo :D
`--backend-arg="['--arg1=a', '--arg2=b']"`

TheQwertiest:
As per: https://mesonbuild.com/Build-options.html
Or do options have a separate parsing logic from other arguments?

xclaesse:
Hmm, personally I prefer `meson compile -- --arg1=a --arg2=b` over `meson compile --backend-arg="['--arg1=a', '--arg2=b']"`, I think it's worth patching CommandLineParser.run() method in mesonmain.py for that.

xclaesse:
otoh, just thinking out loud here, what about per-backend options instead so we can keep writing backend independent commands? Something like: `meson compile --ninja-args="['--opt1', '--opt2']" --vs-args="['--opt3', '--opt4']"` ?

TheQwertiest:
@xclaesse, @jpakkane , decided to go with `--ninja-args`/`--vs-args` approach.

nirbheek:
The `--ninja-args` etc approach is nice for ninja-specific args like `-d keeprsp` but I think `--verbose` should be an argument common to all backends. It's a very generic option that could translate to all backends.

TheQwertiest:
@nirbheek , see https://github.com/mesonbuild/meson/pull/7183#issuecomment-632308331
TLDR: 
- Default verbosity level is different for all backends.
- Ninja default output is too brief, but verbose is alright.
- MSVS verbose output is too verbose, but default is alright.

nirbheek:
Specifically, I am suggesting that `--verbose` should translate to `--verbose` for ninja and to `-verbosity:detailed` (or similar) for msbuild.

nirbheek:
@TheQwertiest yes, but I see `--verbose` as "give me more information" and `--quiet" as "give me less information". In that sense, `--verbose` should map to `ninja --verbose` and `msbuild -verbosity:detailed`. We could also add `--quiet`, which makes `msbuild` quieter but does nothing for ninja (because there is no way to make it quieter).

TheQwertiest:
`-verbosity:detailed` is needlessly verbose and results in HUGE logs even for simple projects. It's 300kb vs 4kb on a simple `11 subdir` test.

nirbheek:
> `-verbosity:detailed` is needlessly verbose and results in HUGE logs even for simple projects. It's 300kb vs 4kb on a simple `11 subdir` test.

Then maybe the user shouldn't pass that? :)

It just seems like if we've come up with a system that requires passing `--ninja-args=-v` for a simple backend-agnostic option like `-v`, we've kinda failed our users.

TheQwertiest:
@nirbheek , my points are the following:
- ninja default output is useless for CI.
- msvc default output is ok for CI.
- msvc `-verbosity:detailed` (and above) results only in huge unreadable walls of texts that are not needed in any usual scenario.
- Thus, I can't use `--verbose` for both backends anyway, and have to use `if backend ==`s.
- Applying `verbose` only to one backend doesn't make any sense and counter-intuitive(especially considering this PR).

> We could also add --quiet, which makes msbuild quieter but does nothing for ninja (because there is no way to make it quieter).

This is exactly the reason I've created this PR. If something is backend-specific it should be specified as such. Creating a lot of `--option (works only in backend B)`s will only make API more convoluted and harder to reason about.

nirbheek:
Then maybe by default we should call `msbuild -verbosity:quiet` so that `--verbose` actually makes sense for both backends?

nirbheek:
> Well, that's how it is for most multiple-backend-supporting tools that I've seen (e.g. conan, cmake from more recent ones), they have actually a specific argument for controlling verbosity for each backend.

I don't think that's a good argument. `ninja` vs `ninja -v` is a good interface. `meson compile` vs `meson compile --ninja-args=-v` is _so bad_.

TheQwertiest:
> Then maybe by default we should call msbuild -verbosity:quiet so that --verbose actually makes sense for both backends?

`quiet` results in no output at all :D
`minimal` though is not that bad (I've only tested it on a really small project though).
So it might not be a bad idea.
But I don't think that it should be a part of this PR.

> Why should anyone want to use meson compile at all?

Because users don't want to know about backend. They want so that simple `configure && make` (or `meson setup && meson compile` in our case) will `just work (c)`. They want to spend less time learning build system and spend more time actually coding. Ninja knowledge should not be required (or forced).

> meson compile vs meson compile --ninja-args=-v is so bad.

It was either this or `meson compile -- -v`. The later is less verbose, but is less flexible as well.

nirbheek:
> Because users don't want to know about backend. They want so that simple `configure && make` (or `meson setup && meson compile` in our case) will _just work (c)_. They want to spend less time learning build system and spend more time actually coding. Ninja knowledge should not be required (or forced upon users).

Counter-point is that if our interface for `meson compile` is hard to use, most people will just use `ninja` in their documentation instead (default backend), and newbies will be confused about the difference. See, f.ex: CMake and their infinite backends. 

We should make `meson compile` as much of a drop-in replacement for `ninja` as possible, and if that means making `msbuild` not be horribly verbose by default, I think that's a feature not a bug.

>  `minimal` though is not that bad (I've only tested it on a really small project though).
> So it might not be a bad idea.
> But I don't think that it should be a part of this PR.

Sure.

TheQwertiest:
> if that means making msbuild not be horribly verbose by default, I think that's a feature not a bug.

Agreed =)
I might even open a PR that does that (sounds like 10 LOC worth).

xclaesse:
> Then maybe by default we should call `msbuild -verbosity:quiet` so that `--verbose` actually makes sense for both backends?

> `quiet` results in no output at all :D
> `minimal` though is not that bad (I've only tested it on a really small project though).

+1 for `-verbosity:minimal` by default. Then `--verbose` could just not pass `-verbosity:minimal` to vs backend. That way both backends stay roughly on the same verbosity level. That's exactly what an abstraction layer is supposed to do: hide backend differences as much as possible.

TheQwertiest:
@xclaesse , I agree with that, but `--verbose` will be in a separate PR =)

nirbheek:
> Then `--verbose` could just not pass ``

Agreed, and for consistency, it could just pass `-verbosity:normal` when `--verbose` is passed.

docs/markdown/Commands.md
```diff
@@ -149,6 +150,26 @@ optional arguments:
...
+  --vs-args VS_ARGS                     Arguments to pass to `msbuild` (applied
+                                        only on `vs` backend).
+```
+
+`BACKEND-args` use the following syntax:
+
+If you only pass a single string, then it is considered to have all values separated by commas. Thus invoking the following command:
+
+```
+$ meson compile --ninja-args=foo,bar
```
xclaesse, line 164: I would use a real example `--ninja-args=-k 0`, or is it `--ninja-args="-k 0"` or  `--ninja-args="['-k', '0']"`` ??

docs/markdown/Commands.md
```diff
@@ -149,6 +150,26 @@ optional arguments:
...
+```
+
+would set the value to an array of two elements, `foo` and `bar`.
+
+If you need to have commas in your string values, then you need to pass the value with proper shell quoting like this:
```
xclaesse, line 169: commas or spaces.

mesonbuild/mcompile.py
```diff
@@ -24,9 +26,28 @@
...
+    elif isinstance(value, list):
+        newvalue = value
+    else:
+        raise argparse.ArgumentTypeError('"{}" should be a string array, but it is not'.format(newvalue))
+
+    for i in newvalue:
+        if not isinstance(i, str):
+            raise argparse.ArgumentTypeError('String array element "{0}" is not a string.'.format(str(newvalue)))
+
+    return newvalue
```
xclaesse, line 49: I would like to not duplicate that code from UserArrayOption. I think it can be replaced by something like:
```
def array_arg(value: str) -> T.List[str]:
  return coredata.UserArrayOption(None, value, user_input=True).value
```

docs/markdown/Commands.md
xclaesse: Actually not, --ninja-args="a b, c" should be fine and be parsed as `a b` and `c`, right?

mesonbuild/mcompile.py

docs/markdown/Commands.md
TheQwertiest: Fixed

docs/markdown/Commands.md
TheQwertiest: Added a better(?) example.

docs/markdown/Commands.md
TheQwertiest: > Actually not, --ninja-args="a b, c" should be fine and be parsed as a b and c, right?

Yep, but it still requires additional quoting for the value though?

docs/markdown/snippets/add_compile_backend_arg.md
```diff
@@ -0,0 +1,26 @@
+## Added ability to specify backend arguments in `meson compile`
+
+It's now possible to specify backend specific arguments in `meson compile`.
+
+Usage: `meson compile [--vs-args--args] [--ninja-args=args]`
```
xclaesse, line 5: typo `--vs-args--args` -> `--vs-args=args`

docs/markdown/Commands.md
xclaesse: yep, worth mentioning it needs quotes.

docs/markdown/snippets/add_compile_backend_arg.md
TheQwertiest: Thx, fixed

TheQwertiest:
Rebased and squashed.

Issue #5438, opened 2019-05-31
Built-in Boost dependency takes precedence over explicitly provided pkg-config dependency


agurtovoy:
I'm using meson together with conan/pkg-config (on Windows). One of the project dependencies is Boost, and conan dutifully generates the corresponding `boost.pc` file, but meson ends up trying to use the built in `BoostDependency` class before even checking whether the `boost.pc` file exists:

```Python
def _build_external_dependency_list(name, env: Environment, kwargs: Dict[str, Any]) -> list:
    # First check if the method is valid
    if 'method' in kwargs and kwargs['method'] not in [e.value for e in DependencyMethods]:
        raise DependencyException('method {!r} is invalid'.format(kwargs['method']))

    # Is there a specific dependency detector for this dependency?
    lname = name.lower()
    if lname in packages:
        # Create the list of dependency object constructors using a factory
        # class method, if one exists, otherwise the list just consists of the
        # constructor
        if getattr(packages[lname], '_factory', None):
            dep = packages[lname]._factory(env, kwargs)
        else:
            dep = [functools.partial(packages[lname], env, kwargs)]
        return dep

    candidates = []

    # If it's explicitly requested, use the dub detection method (only)
    if 'dub' == kwargs.get('method', ''):
        candidates.append(functools.partial(DubDependency, name, env, kwargs))
        return candidates

    # If it's explicitly requested, use the pkgconfig detection method (only)
    if 'pkg-config' == kwargs.get('method', ''):
        candidates.append(functools.partial(PkgConfigDependency, name, env, kwargs))
        return candidates

```

Other pkg-config dependencies are picked up as expected. Here's my `meson.build`:

```
project('xxx', 'cpp',
    version: '2.2.1',
    meson_version: '>=0.50')

sources = files(
    'logger.cpp',
    'dll_main.cpp'
    )

inc = include_directories( '.', 'resources' )
shared_library('logger', sources, include_directories: inc, 
    dependencies: [
        dependency('pantheios'),
        dependency('boost'),
        dependency('stlsoft'),
        dependency('zlib')
    ])
```

Thoughts?

dcbaker:
I don't have a .pc from my boost install on Linux, and looking at their [trac](https://svn.boost.org/trac10/ticket/1094), they don't have pkg-config support. Is conan generating the .pc file for you?

agurtovoy:
[...]

PR #7184, opened 2020-05-22
Add boost_root kwarg support to `dependency('boost')`

### LINKED ISSUE #249: dependency('boost') should have a kwarg for the boost root directory
Hi,

Would it be possible to add a path argument to dependency() for the times when we have libs installed in non-standard places? I see that boost will look at the env variable BOOST_ROOT, but it would be nice to not have to rely on the env.

Thanks!

cmcneish:
Resolves #249

This commit implements a `boost_root` kwarg for boost dependencies.

Systems I work on typically have several boost installations per user: one (old) system-wide version, and one or more with varying { versions x ABIs } installed by a user (e.g. by [spack](https://github.com/spack/spack)).

Meson (and CMake) will often result in the compiler picking up headers from a user-installed version, but libraries from the system, resulting in linker errors at the very end of a build and some frustration.
This PR allows `boost_root` to be specified inline or in a configuration file in a way that's persistent, which prevents these issues.

Outstanding questions:
1) I'm not familiar enough with how the testing (CI and regular) works to add tests for this functionality. Any ideas? I'm thinking I could symlink a boost install somewhere else in CI and look for it there. 
2) Is what I did with the new kwargs parameters correct?
    a) Can I add `boost_root` to `permitted_kwargs` for `dependency`? It's not clear to me how to add it for just the `dependency('boost')` case.
    b) What do I set the version in `@FeatureNewKwargs` to, in general? The next release number, as I've done here?
3) I'm trying to copy the `dirs` behavior from e.g. `find_program`, where `boost_root` will fall back to the system installation if not found. But, boost_root isn't plural. But, the env var BOOST_ROOT also can take multiple optional paths and also isn't plural. Any opinions here (or am I just overthinking things)?
4) I'm assuming that custom installs are generally not going to be split in ways that require use of BOOST_INCLUDEDIR + BOOST_LIBRARYDIR. (Certainly, I don't need them for the use case I have in mind). Should I just bite the bullet and implement them here too?

### DIFF
--- a/docs/markdown/Dependencies.md
+++ b/docs/markdown/Dependencies.md
@@ -289,7 +289,10 @@ use those to link against your targets.
 
 If your boost headers or libraries are in non-standard locations you
 can set the BOOST_ROOT, BOOST_INCLUDEDIR, and/or BOOST_LIBRARYDIR
-environment variables.
+environment variables. *(added in 0.55.0)* You can also set the argument
+`boost_root` to select a boost installation. BOOST_ROOT does not
+fall back to `boost_root` or system boost directories, however
+`boost_root` will fall back to system directories if not found.
 
 You can set the argument `threading` to `single` to use boost
 libraries that have been compiled for single-threaded use instead.
--- a/mesonbuild/dependencies/boost.py
+++ b/mesonbuild/dependencies/boost.py
@@ -367,7 +367,7 @@ class BoostDependency(ExternalDependency):
         self.arch = environment.machines[self.for_machine].cpu_family
         self.arch = boost_arch_map.get(self.arch, None)
 
-        # Prefere BOOST_INCLUDEDIR and BOOST_LIBRARYDIR if preset
+        # Prefer BOOST_INCLUDEDIR and BOOST_LIBRARYDIR if present
         boost_manual_env = [x in os.environ for x in ['BOOST_INCLUDEDIR', 'BOOST_LIBRARYDIR']]
         if all(boost_manual_env):
             inc_dir = Path(os.environ['BOOST_INCLUDEDIR'])
@@ -390,8 +390,10 @@ class BoostDependency(ExternalDependency):
         elif any(boost_manual_env):
             mlog.warning('Both BOOST_INCLUDEDIR *and* BOOST_LIBRARYDIR have to be set (one is not enough). Ignoring.')
 
-        # A) Detect potential boost root directories (uses also BOOST_ROOT env var)
-        roots = self.detect_roots()
+        # A) Detect potential boost root directories (uses BOOST_ROOT env var, then boost_root kwarg (if present) and system paths)
+        kwarg_boost_root = mesonlib.extract_as_list(kwargs, 'boost_root')
+
+        roots = self.detect_roots(kwarg_boost_root)
         roots = list(mesonlib.OrderedSet(roots))
 
         # B) Foreach candidate
@@ -592,9 +594,7 @@ class BoostDependency(ExternalDependency):
             libs += [BoostLibraryFile(i)]
         return [x for x in libs if x.is_boost()]  # Filter out no boost libraries
 
-    def detect_roots(self) -> T.List[Path]:
-        roots = []  # type: T.List[Path]
-
+    def detect_roots(self, kwarg_boost_roots: T.List[str]) -> T.List[Path]:
         # Add roots from the environment
         for i in ['BOOST_ROOT', 'BOOSTROOT']:
             if i in os.environ:
@@ -602,8 +602,18 @@ class BoostDependency(ExternalDependency):
                 paths = [Path(x) for x in raw_paths]
                 if paths and any([not x.is_absolute() for x in paths]):
                     raise DependencyException('Paths in {} must be absolute'.format(i))
-                roots += paths
-                return roots  # Do not add system paths if BOOST_ROOT is present
+                return paths  # Do not add system paths if BOOST_ROOT is present
+
+        roots = []  # type: T.List[Path]
+
+        # Add roots from kwarg boost_root parameter
+        if kwarg_boost_roots:
+            paths = [Path(x) for x in kwarg_boost_roots]
+            if paths and any([not x.is_absolute() for x in paths]):
+                print([not x.is_absolute() for x in paths])
+                print(paths)
+                raise DependencyException('Paths in "boost_root" kwarg must be absolute')
+            roots += paths  # Prefer but do not require paths provided in boost_root
 
         # Add roots from system paths
         inc_paths = [Path(x) for x in self.clib_compiler.get_default_include_dirs()]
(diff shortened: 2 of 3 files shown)


mensinda:
I am not sure if adding a `boost_root` kwarg is the right approach here. IMHO it would make much more sense to implement this with [native](https://mesonbuild.com/Native-environments.html) and [cross](https://mesonbuild.com/Cross-compilation.html) files. So `BoostDependency` should read these entries from `environment.properties[self.for_machine]` instead of a kwarg. This way, the boost root is still persistent but you avoid storing system specific stuff in your `meson.build`.

I hope the following answers your questions:

1. The boost tests are located in `test cases/frameworks/1 boost`. You could add a new (optional) meson option that specifies the boost root and passes it to the kwarg. However, this won't be trivial since boost isn't installed in a fixed location.
2. a) Yes this is correct. b) Yes, the next meson release should always be used. Something doesn't make it into the next release the version number has to be updated.
3. I would say you are overthinking things :) I would generally expect that a user only ever wants to specify one path in `BOOST_ROOT` and `boost_root` so it won't matter anyway. But if you need to set more paths (for some reason but not sure why you would ever want to do that) meson will support it. In short, I don't think that it matters if it's called `boost_root` or `boost_roots`.
4. Yes, the should also be supported for consistency.

If we do end up with a new kwarg it should at least be something that works for all dependencies (`dependency_root` or something like that). Otherwise we will end up with a `*_root` kwarg for every dependency.

Also, since 0.54.0, meson should only pick up boost from one prefix, if it doesn't please open an issue :)

PS: The msys2 CI is known to be broken.

TheQwertiest:
@mensinda , the problem is that `boost` dependency is a special snow-flake anyway. It does not allow to specify the find method directly, so user can't just use `method: pkg-config` and have to use hacks to specify boost paths (i.e. with environment variable) (https://github.com/mesonbuild/meson/issues/5438).
In other words, user can't contain all boost dependency handling logic in `meson.build` files, they have to add workarounds above meson (e.g. by adding additional script for invoking meson). Adding ability to specify boost paths will fix that (user will still have to find and parse pkg-config manually, but it can be done inside `meson.build`).

mensinda:
@TheQwertiest I am not against specifying boost paths. I just think that doing it in a `meson.build` is the wrong approach. The boost root dir is something specific to a machine and should thus be in the cross or native file and not in the `meson.build` which should work for *all* platforms.

As for `*.pc` support, the main reason why it doesn't work is that no one has implemented (or rather whitelisted it yet). Adding basic support should only be a few lines of code. However, I have no idea how to implement modules with `*.pc` files or how to select the correct `*.pc` files for windows. Again, [this total mess](https://www.boost.org/doc/libs/1_73_0/more/getting_started/windows.html#library-naming) is still a huge pain.

TheQwertiest:
> I just think that doing it in a meson.build is the wrong approach. The boost root dir is something specific to a machine and should thus be in the cross or native file and not in the meson.build which should work for all platforms

That's not how it works for all other deps though. Usually a user only have to specify `PKG_CONFIG_PATH` which points to all `.pc` files. And that's enough for working with all deps except for `boost`, which requires a specific env variable. I.e. the only things that're specified in machine files are paths to `.pc` files (which are often static) and *not* actual paths to dependency binaries (which are often dynamic). And boost requires paths to *binaries*, so run-time  `.pc` file parsing is needed.

mensinda:
@TheQwertiest sure, but this is *not* what *this* PR is about (as in it does not have anything to do with pkg-config). My argument is that if we have a system-specific path somewhere in meson, it should be in the cross/native file and not in the `meson.build` where it is even more out of place.

Just to be sure we are talking about the same thing: All this PR does is to add a way to set the boost root directory (not even the binaries itself) for the system boost dependency in a way that does not rely on environment variables.

Also, most users don't have system `*.pc` files (and also wouldn't want to write their own `*.pc` files). So adding a way to specify the boost root directory is still valid even if `*.pc` files are supported.

TheQwertiest:
> Just to be sure we are talking about the same thing: All this PR does is to add a way to set the boost root directory (not even the binaries itself) for the system boost dependency in a way that does not rely on environment variables.

Yep. =)
I was trying to prove that allowing settings *root path* in `meson.build` is a needed feature and that a machine-file might not be the right place to setup dependency rootpaths.

> This isn't the case. Meson only has to know the boost root directory (or the include directory and the library dir if you are doing something fancy). Specifying paths to binaries should not be required

Sorry, my wording was bad, by *binaries path* I meant *root path* (i.e. a path to directory containing `bin`, `include` and etc).

mensinda:
OK, but why is the `meson.build` better suited for setting the boost root over the native / corss file? I honestly don't get it. As far as I can tell, the `meson.build` should be independent from the underlying system (it should work on Windows, Mac, and every Linux). Hardcoding a system-specific path in there is far from platform-independent.

Native / Corss files, on the other hand, are written by the user who builds the software for a specific system (which is not necessarily the author of the project). So it makes far more sense to add `boost_root` there. I get that this also is not ideal, but it is a far better solution since the `PKG_CONFIG_PATH` is also set there and not in the `meson.build`.

TheQwertiest:
> Hardcoding a system-specific path in there is far from platform-independent.

I think, there might be a slight miscommunication. This is the usage scenario I'm thinking of:
```py
# Preconditions: 
# - there is a `boost.pc` in PKG_CONFIG_PATH
# - `boost.pc` contains `root path`
# - `root path` points to a properly structured directory
root_path = run_command('find_boost_pc_from_pkg_config_path_and_get_root_path')
dep_boost = dependency('boost', root_path: root_path )
```

jpakkane:
The build files should be as much as possible about specifying _what_ dependencies you require. The question of _how_ the requirements are fulfilled should be done (again, as much as possible) outside the build files.

mensinda:
> This is not ideal, obviously, but at least it won't require a build system (e.g. conan) to special-case boost dependency.

We should fix conan support propperly, rather than telling users to relly on such a hack (that will probably break on non conan systems). I would go so far as to say that it is a meson bug, if you require `run_command` to get the boost_root and you can't just set it in the cross / native file (assuming you are not doing something completely exotic).

cmcneish:
Thanks for all the comments. I'm working on a native/cross file implementation.


> Also, since 0.54.0, meson should only pick up boost from one prefix, if it doesn't please open an issue

Yeah, I think my cases of split detections are all from versions pre 0.54.


For

> OK, but why is the meson.build better suited for setting the boost root over the native / cross file?

and

> The build files should be as much as possible about specifying what dependencies you require.

some arguments in favor of `boost_root`:

- `find_program` and `find_library` have `dirs` parameters that do approximately the same thing as described. This could be the parameter name for the `dependency_root` idea suggested above, for consistency.
- On the systems I use, due to filesystems shared between users, what and where a dependency is can be the same thing for all users of a given project.
- Unfortunately `--native-file` is another argument to remember to pass in the same vein as `BOOST_ROOT`. (It is still nicer in that I don't have to track down the path to my boost install every time, and the presence of the native file is a decent reminder.)

mensinda:
Closing since this is replaced by #7210.

PR #7188, opened 2020-05-23
Remove stray print call.

jpakkane:
The current logs are unreadable because all text printed by failing meson logs gets spammed to stdout.


TheQwertiest:
Thanks! It was really painful to parse the logs >_<

PR #7189, opened 2020-05-23
mcompile: replaced backend divination code + cleanup

TheQwertiest:
- Replaced backend divination code with coredata parsing (`meson compile` should be used only with meson configured folders anyway).
- Extracted backend specific code to separate functions.
- Improved(?) error message when invoked on non-configured directory.

### DIFF
--- a/mesonbuild/mcompile.py
+++ b/mesonbuild/mcompile.py
@@ -14,8 +14,6 @@
 
 """Entrypoint script for backend agnostic compile."""
 
-import os
-import shutil
 import sys
 import typing as T
 from pathlib import Path
--- a/mesonbuild/mcompile.py
+++ b/mesonbuild/mcompile.py
@@ -24,6 +24,7 @@ from . import mlog
 from . import mesonlib
 from . import coredata
 from .mesonlib import MesonException
+from mesonbuild.environment import detect_ninja
 
 if T.TYPE_CHECKING:
     import argparse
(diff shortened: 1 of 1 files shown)


TheQwertiest:
CI failures seem unrelated:
```
Failed test during configure: 'test cases/warning/1 version for string div'
```

mesonbuild/mcompile.py
```diff
@@ -27,6 +28,20 @@
 if T.TYPE_CHECKING:
     import argparse
 
+def get_backend_from_introspect(builddir: Path) -> str:
+    """
+    Gets `backend` option value from introspection data
+    """
+    path_to_intro = builddir / 'meson-info' / 'intro-buildoptions.json'
```
xclaesse, line 35: I think you should get it from coredata file instead. Should be faster to pickle that file than parsing json.

TheQwertiest:
Changes in the latest commit:
- Replaced backend divination code with coredata parsing instead of introspect.
- Extracted backend specific code to separate functions.
- Improved(?) error message when invoked on non-configured directory.

mesonbuild/mcompile.py
```diff
@@ -15,19 +15,80 @@
...
+                             'meson version. Please regenerate it in this case.'.format(builddir))
+
+def get_backend_from_coredata(builddir: Path) -> str:
+    """
+    Gets `backend` option value from coredata
+    """
+    return coredata.load(str(builddir)).get_builtin_option('backend')
+
+def get_parsed_args_ninja(options: 'argparse.Namespace', builddir: Path):
+    runner = os.environ.get('NINJA')
```
xclaesse, line 48: Better use detect_ninja() from environment.py here.

mesonbuild/mcompile.py
```diff
@@ -15,19 +15,80 @@
...
 from .mesonlib import MesonException
 
 if T.TYPE_CHECKING:
     import argparse
+    
+def validate_builddir(builddir: Path):
+    if (not builddir.is_dir() or
+        not (builddir / 'meson-private').is_dir() or
+        not (builddir / 'meson-info').is_dir() or
+        not (builddir / 'meson-info' / 'meson-info.json').is_file()):
```
xclaesse, line 35: Why do you check each subdir? Can't you just check if coredata.dat exists?

TheQwertiest: Haha, fair enough (I've just blindly copy pasted this check from mintro)

xclaesse: Right, my 2 remaining nitpickings are not your fault, was already like that, but while you're here it would be nice to cleanup :)

TheQwertiest: I like to leave the code in the better state than I've found it, so all nit-picks are appreciated.

mesonbuild/mcompile.py

mesonbuild/mcompile.py

xclaesse:
Looks good to me, thanks.

PR #7190, opened 2020-05-23
Give user control of option skip_sanity_check

Neumann-A:
Either give the user control over `skip_sanity_check` or print an error why the user is doing something stupid. This PR uses the former approach

### DIFF
--- a/mesonbuild/interpreter.py
+++ b/mesonbuild/interpreter.py
@@ -3169,13 +3169,13 @@ external dependencies (including libraries) must go to "dependencies".''')
         return success
 
     def should_skip_sanity_check(self, for_machine: MachineChoice) -> bool:
-        if for_machine != MachineChoice.HOST:
-            return False
-        if not self.environment.is_cross_build():
-            return False
         should = self.environment.properties.host.get('skip_sanity_check', False)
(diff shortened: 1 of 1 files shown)


mensinda:
Please rebase this PR on master to rerun the CI. This should get rid of the CI failures.

vlj:
What is the status of this PR ? (there is a long chain of PR dependencies for vcpkg...)

c72578:
@jpakkane Please have a look at this PR and merge if OK from your side.

jpakkane:
Could you please explain a bit more what is happening here? What is the actual problem is. "Give control of option skip_sanity_check" is not very informative by itself.

Neumann-A:
@jpakkane: there are cases where the option is simply ignored by meson. 

The previous code:
```
        if for_machine != MachineChoice.HOST:	
            return False	
        if not self.environment.is_cross_build():	
            return False
```
is returning an early false without even checking if the user provided the `skip_sanity_check` option.
As such if the user provided `skip_sanity_check=true` meson still runs the sanity check if on of the early returns fires. (Which is the case for arm crosscompiles on windows.)

PR #7202, opened 2020-05-25
 opts: Allow string concatenation (fixes #7199) 

### LINKED ISSUE #7199: meson_options.txt don't allow splitting long strings nicely
Currently `meson_options.txt` don't provide a way to split long strings in multiple lines. If I try to use the `+` operator, the syntax is rejected. My use case is the `description` value. Descriptions can be long to be useful, but I don't like writing anything beyond the 80-column barrier and would rather break this long line.

mensinda:


docs/markdown/snippets/options_string_concat.md
```diff
@@ -0,0 +1,14 @@
+## String concatenation in meson_options.txt
+
+It is not possible to use string concatenation (with the `+` opperator) in the
```
xclaesse, line 3: Not->now

xclaesse:
Otherwise looks good to me

docs/markdown/snippets/options_string_concat.md
mensinda: Fixed, Thanks.

TheQwertiest:
Why not use python\C\C++ syntax instead? I.e. when two consequent strings are treated as one :
```
option(
  'test' 'option',
  type : 'string',
  value : 'optval',
  description : 'An option with a very long description'
                      'that does something in a specific context'
)
```

mensinda:
Because you can't do that in plain meson either and `meson_options.txt` just uses a subset of normal meson code (which means that it also heavily relies on the existing AST / interpreter).

TheQwertiest:
Then it should be added in meson interpreter as well :P Joking (or am I? :D)

mensinda:
Sure, but not in this PR :)

xclaesse:
LGTM.

PR #7205, opened 2020-05-26
[RFC] Add 'prebuilt' keyword for gtest/gmock dependencies

falconindy:
I don't believe this is in a mergeable state (it's at least missing
documentation), but I'd like to submit it for feedback as to whether or not the
approach is acceptable/wanted, particularly with regard to the addition of a
new keyword. This could, alternatively, be fixed by just making the 'system'
look for sources first rather than prebuilt.

---

gmock and gtest can be built from source or included as prebuilt
dependencies. This only works when the method is declared as 'system',
but the detection is inflexible in letting the user choose between the
prebuilt lib and the source when both are present -- the prebuilt is
always favored. This change allows that to be overridden.

Why is this useful? Exported C++ ABIs are never stable. Therefore,
prebuilt libraries that might exist in a distrbution can be strictly
incompatible with a meson-configured build. As an example, a distro
might generally build/package with libstdc++, but a project may want to
use -stdlib=libc++, which would result in linker errors when it comes
time to include the prebuilt gtest/gmock dependencies.


jpakkane:
Are there systems that provide both prebuilt and source versions at the same time or is it always one or the other?

falconindy:
Arch Linux packages both the source and the .so. You can see the file list here:

https://www.archlinux.org/packages/community/x86_64/gtest/

(See the "Package Contents" section at the bottom)

falconindy:
Debian has a "libgtest-dev" package which only includes a static lib and the pkg-config files, but it depends on the "googletest" package which contains all the source.

jpakkane:
A design question is that if you set `prebuilt: true` and the prebuilt one is not available but the sources are, should it fallback to those or error out. Or, in other words, would `prefer_prebuilt` be a better name?

falconindy:
Yeah, that's a good point. Making this a suggestion rather than a requirement would have nicer failure behavior.

falconindy:
I've also discovered that Arch Linux has decided to ship gtest sources under a path that meson doesn't consider -- /usr/src/googletest/src. There's no standard here, only convention. Sigh. Going to chase down the Arch side first to see if it's reasonable to move the sources to /usr/src/gtest/src to match our /usr/src/gmock/src.

jpakkane:
Has the arch directory thing been resolved?

falconindy:
I raised this, but we didn't really come to a conclusion because there's literally no standard here. Would it be terrible to add a third path to the lookup?

PR #7207, opened 2020-05-26
Recognize Arduino .ino files as C++ files

phillipjohnston:
Renaming .ino files is not an option when working with projects that must use the IDE. Meson should allow their use with the C++ compiler.


phillipjohnston:
I understand if this is undesirable. As another option, we could update the documentation to explain how to use .ino files, perhaps by creating a .cpp file which includes the .ino file:

```
#include "Arduino.h"
#include "../../examples/default_to_serial/default_to_serial.ino"
```

PR #7208, opened 2020-05-26
Fix lack of space after 'Cflags:' in pkgconfig files

Soapux:


xclaesse:
LGTM once CI is green.

out of curiousity, does it cause issues or is it only for consistency with other fields?

Soapux:
Not sure if it causes any issues, I just thought I'd make it consistent.  I had noticed the spacing changed when upgrading to 0.54.

PR #7210, opened 2020-05-27
Add boost_root support to properties files

cmcneish:
Replaces #7184

This commit implements `boost_root`, `boost_includedir`, and `boost_librarydir` variable support to native and cross properties files. The search order is currently environment variables, then
these variables, and finally a platform-dependent search.

Why does `BOOST_ROOT` require an absolute path? It dates to the very beginning of the boost dependency (https://github.com/mesonbuild/meson/commit/af4e3b6671618d468da0fdaae769a904ad448810) and `BOOST_LIBRARYDIR` / `BOOST_INCLUDEDIR` don't currently have the same restriction. I haven't added it this restriction to `boost_root` found in property files, but it's I'll go ahead and add it for a good reason. I've primarily left it out to ease testing:

I still don't really know what to do by way of testing. It seems to me that there are two main goals:
1. are we actually using the new parameters
2. is search order correct (env -> properties -> system search)

For 1., I put together two test cases that find a faked version.hpp. They use relative paths to find their `version.hpp`s, which is why omitting the absolute path requirement above is helpful.

I don't know how to implement 2. in the current test structure, hints would be appreciated. It looks like `BOOST_ROOT` itself isn't getting tested except incidentally by the Windows CI, unless I'm missing something.


mensinda:
This looks mostly good, but I have a conceptual issue with how the native/cross-file `boost_*` and the `BOOST_*` env vars are handled:

First of all, I don't think that env vars should override entries in the cross/native files. I am not 100% sure if this is also the case in other parts of meson, but my intuition is that all entries in the native/cross-files should take precedence over environment variables. Again I am not 100% sure about this so ping @jpakkane, @Ericson2314.

Secondly, If `boost_*` is present in the machine file then it should behave like the `BOOST_*` so that *only* the `boost_*` entries are considered without any fallback to the builtin lookup. This is because a lot of users want to specify a *specific* boost and not just adding an additional search path. Accidentally finding the system boost version would lead to hard to catch bugs and a lot of user frustration.

So the search order should be:
1. `boost_*` machine file entries (abort if boost is not found with those)
2. `BOOST_*` env vars (abort if boost is not found with those)
3. All the other system detection methods

> I still don't really know what to do by way of testing. It seems to me that there are two main goals:

Your test case is sufficient for 1. You could add some empty `libboost_regex.so.0.0.1`, `boost_regex-vc142-mt-gd-x64-0_00.lib` (as well as the mac equivalent) into your lib dirs and let meson find them.

Testing 2. seems too fragile for our CI setup. We have enough CI breakages due to system changes without checking some complex fallback logic (and yes, we are heavily dependent on `BOOST_ROOT` on the Windows CI) :)

> Why does `BOOST_ROOT` require an absolute path?

Relative paths just don't make sense most of the time and are fragile at best. Consider this example for instance:

```bash
~ $ export BOOST_ROOT='./boost'
~ $ cd my_project
~/my_project $ meson build  # The boost is installed in ~/boost, but meson searches in ~/my_project/boost
```

So, relative paths in `BOOST_*` are most likely a user error, and I guess we just forgot to check `BOOST_{LIBRARY,INCLUDE}DIR`. So please check that all paths are absolute. This will complicate testing though...

To fix your tests, I would recommend checking for `{native,cross}file.ini.in` in [`run_project_tests.py`](https://github.com/mesonbuild/meson/blob/master/run_project_tests.py#L498) and using [`mesonlib.do_conf_file`](https://github.com/mesonbuild/meson/blob/master/mesonbuild/mesonlib.py#L1000) to replace `@ROOT@` with the current test root.

cmcneish:
> I don't think that env vars should override entries in the cross/native files

Done, order is now cross/native, then env vars, then pkg-config for Conan, then system search. Property files and env vars, if specified, should not fall back to anything else.

> A lot of users want to specify a specific boost and not just adding an additional search path

Done, neither cross/native and env var settings now fall back to anything else. (Previously, putting in  one of the split-root variables but not the other resulted in a warning. Now, it raises a `DependencyException`)

> [empty libraries with matching fake versions]

Two problems, see after the break

> Testing 2. seems too fragile for our CI setup.

It would still be nice to have this kind of test in the future, since control flow around detection success and failure is fairly opaque.

> [no relative paths]

All paths are now required to be absolute, newly including `BOOST_LIBRARYDIR` + `BOOST_INCLUDEDIR`

> `run_project_tests.py`

This worked, thank you for the suggestion.
I wasn't entirely sure where to put this configuration step in that file--test-specific logic only seems to happen in `skippable`, but that isn't the right place for this. Let me know if you have better ideas here.

---

1. I don't know what to name the fake boost_regex so that all of the windows CI tests pass. 

I'm currently getting
```
  - filtered library list:
  - found:
  - NOT found:
    - boost_regex
```
for the single file `boost_regex-vc142-mt-gd-x64-0_01.lib`. Do I also need the `.dll`? Do I need to vary `-vc*-` and `-x64-`? Does anything interesting happen with vc versus mingw? The first pipelines failure happens after around half an hour, so I'm reluctant to just try things out for these.

2. *.dylib is being preferred to *.so on linux; as far as i can tell this isn't something I introduced. I don't expect anyone to hit this in sane environments, however it's plausible with irresponsible NFS usage. I'll try and fix it soon.

run_project_tests.py
```diff
@@ -614,6 +623,10 @@ def gather_tests(testdir: Path, stdout_mandatory: bool) -> T.List[TestDef]:
         if test_def_file.is_file():
             test_def = json.loads(test_def_file.read_text())
 
+        # Perform setup for boost tests that require absolute paths in nativefile.ini
+        if str(t.path).endswith('32 boost root') or str(t.path).endswith('33 boost split root'):
```
mensinda, line 627: This should work for all tests if the `nativefile.ini.in` is present, not just boost.

run_project_tests.py
```diff
@@ -603,6 +604,14 @@ def force_regenerate():
 
     return testresult
 
+def set_boost_nativefile_paths(test: TestDef) -> None:
+    script_dir = os.path.abspath(os.path.split(__file__)[0])
+
+    confdata = ConfigurationData()
+    confdata.values = {'ROOT': (script_dir, 'project base directory')}
+
+    mesonlib.do_conf_file(str(test.path / 'nativefile.ini.in'), str(test.path / 'nativefile.ini'), confdata, 'meson')
```
mensinda, line 613: Please put the generated files into the build dir and not into the project dir. Polluting the project dir can still be a pain for running tests locally, even if they are in the .gitignore.

You will have access to the build dir if you move the function call to here: https://github.com/mesonbuild/meson/blob/master/run_project_tests.py#L498. Maybe you could make a generic `detect_cross_args() -> Path, Path` function that handles the configuration and detection of any cross/native files.

run_project_tests.py
cmcneish: Fixed, see `detect_parameter_files()`

cmcneish:
Looks like everything was fine on Windows, except dlls weren't supposed to have the patch number in their name. I've moved the version checking up to the minor number, which seems to be working (the cygwin failure looks unrelated, and it passes the tests I've added).

I've moved pre-test native/cross file configuration like you suggested.

I looked at only picking up `*.dylib` on darwin but it looks to me like more effort than it's worth, at least as part of this PR. Instead, I've dropped the `.dylib` files entirely in favor of the (versioned) `.so` files.

nirbheek:
@jpakkane does this look ok interface-wise?

mesonbuild/dependencies/boost.py
```diff
@@ -367,36 +367,26 @@ def __init__(self, environment: Environment, kwargs):
...
+        # First, look for paths specified in a property file
+        props = self.env.properties[self.for_machine]
+        boost_property_env = [props.get('boost_includedir'), props.get('boost_librarydir'), props.get('boost_root')]
+        if any(boost_property_env):
+            self.detect_boost_property_file(props)
+            return
 
-            self.is_found = self.run_check([boost_inc_dir], [lib_dir])
+        # Next, look for paths in the environment
+        boost_manual_env = [x in os.environ for x in ['BOOST_INCLUDEDIR', 'BOOST_LIBRARYDIR', 'BOOST_ROOT', 'BOOSTROOT']]
```
Ericson2314, line 378: You should use `get_env_var` from `envconfig`, which will add the `_FOR_BUILD` suffix for the native boost in cross builds, as is proper.

Ericson2314:
Also, rather than having separate but very similar code paths for the config files and environment variables, could we just look in one then the other per-key? I know that means one could put some things in config files and other things in env vars, but I don't really care.

My general view is that rather than peek at `os.environ` all over the codebase, we should have a list of all environment we use to initialize the meson `environment` prior to loading the config files, and `os.environ` should be banned in the rest of the codebase. The upshot of that is I think having separate env vs config file code paths elsewhere is just asking for things to drift apart, and just too much effort.

cmcneish:
> get_env_var

Makes sense to me; I've removed my use of `os.environ` in favor. A lot of the rest of meson is still using `os.environ`, however.

> [R]ather than having separate but very similar code paths for the config files and environment variables, could we just look in one then the other per-key?

I did pull some out into `BoostDependency.detect_split_root`; the rest is fairly tightly tied to error reporting or actual differences in processing (retrieval of parameters, and handling of both `BOOST_ROOT` and `BOOSTROOT` versus just `boost_root`, for example).
Seems to me that there's relatively little benefit to combining these functions (at least as part of this pull request) if so much complexity is going to be pulled into the caller.

mesonbuild/dependencies/boost.py
cmcneish: Thank you.

Ericson2314:
Well, I suppose it is fine for now. Maybe if/when we do front-load the env vars as I describe, will also store where each setting came from to make this sort of specialized error message easier.

nirbheek:
> My general view is that rather than peek at `os.environ` all over the codebase, we should have a list of all environment we use to initialize the meson `environment` prior to loading the config files, and `os.environ` should be banned in the rest of the codebase

I completely agree. Besides fixing re-setup corner-cases when the `compile` stage has a different env than the `setup` phase, this will also help us on Windows for:

1. Adding support for cross-compiling with MSVC while having a native MSVC compiler
2. Support building with MSVC using a command-line arg like CMake instead of requiring the user to open a cmd prompt and run `vcvarsall.bat` themselves

Since in these cases we will have to parse the env set by `vcvarsall.bat` ourselves and store it somewhere.

jpakkane:
> does this look ok interface-wise?

Yep, feel free to merge once you are happy with the implementation.

mensinda:
@cmcneish Could you please rebase this on master to rerun the CI? Unfortunately, this PR didn't make it into 0.55, so could you also change the version to 0.56?

cmcneish:
Rebased, and version changed to 0.56.

I also changed some of the language to talk about "machine files" instead of "property files" since that seems to be more common.

@mensinda will this need anything else after CI runs?

mensinda:
Thanks, this looks good and the CI is green. Merging.

For next time, please try to reduce the number of commits by squashing some of them together. Especially when a commit reverts/changes something in a previous commit.

Since the final changes in this PR are small enough, I am going to squash everything into a single commit.

cmcneish:
Thanks

> For next time, please try to reduce the number of commits by squashing some of them together. Especially when a commit reverts/changes something in a previous commit.

Sure, sorry about that

Issue #6927, opened 2020-04-09
Meson find_program fails to find python when installed from the Windows Store in 0.54.0


jenatali:
**Describe the bug**
In 0.54.0, a change was committed (`64432e94481dbd55bedf92f21610bc5ed6b52655`) to ignore the Windows 10 "Execution Aliases" directory, to prevent finding e.g. python.exe when it's just a redirect to the Store. But once you install Python from the store, that redirect actually becomes the program that's run. So, running python.exe from the execution alias, in order to run meson.py, then fails to find a program named python.exe.

**To Reproduce**
1. Install Python from the Windows store.
2. Install/clone meson.
3. Run meson on a script including `find_program("python")`.

**Expected behavior**
Python is found.

Consider checking whether the current meson instance is being run from the execution alias before excluding that directory from program search. People setting up new development machines now that this exists may actually *want* to use the Windows Store version of python, to prevent similar issues with other tools finding the dummy python.exe instead of the real installation.

jpakkane:
The only real Windows install I have is an old Windows 7 I use for playing games, so unfortunately I can't really fix this (and even if I could I would not be able to properly test it). Patches are welcome, though.

PR #6459, opened 2020-01-14
find_program: Ignore programs in the WindowsApps directory

nirbheek:
The latest Windows 10 release in May 2019 added zero-sized files that act as stubs which when launched from cmd.exe spawn the Windows Store to install those apps. This also includes python.exe and python3.exe:

https://devblogs.microsoft.com/python/python-in-the-windows-10-may-2019-update/

Unfortunately, `import('python').find_installation('python3')` will then think that python3.exe is available on Windows. Or, worse, if the user has a fresh installation of Windows 10 and then installs the Python 3 using the official installer (not the Windows Store app), we will *still* pickup this stub because it will be first in `PATH`.

Always remove the WindowsApps directory from `PATH` while searching.

First reported at https://gitlab.freedesktop.org/gstreamer/cerbero/issues/223

PR #7214, opened 2020-05-27
Fix python3 installed from the windows app store

d-scott-phillips:
Windows has a python3.exe in the WindowsApps directory that either
redirects you to the app store to install python, or dispatches to the
real python3.exe if it's installed. To allow using this python, don't
exclude WindowsApps from path searches (but do order it last), and
ignore any error in resolving the executable path to the dispatcher.

Closes https://github.com/mesonbuild/meson/issues/6927

### DIFF
--- a/mesonbuild/dependencies/base.py
+++ b/mesonbuild/dependencies/base.py
@@ -22,6 +22,7 @@ import json
 import shlex
 import shutil
 import stat
+import sys
 import textwrap
 import platform
 import typing as T
@@ -1847,14 +1848,22 @@ class ExternalProgram:
         # Ensure that we use USERPROFILE even when inside MSYS, MSYS2, Cygwin, etc.
         if 'USERPROFILE' not in os.environ:
             return path
-        # Ignore executables in the WindowsApps directory which are
-        # zero-sized wrappers that magically open the Windows Store to
-        # install the application.
+        # The WindowsApps directory is a bit of a problem. It contains
+        # some zero-sized .exe files which have "reparse points", that
+        # might either launch an installed application, or might open
+        # a page in the Windows Store to download the application.
+        #
+        # To handle the case where the python interpreter we're
+        # running on came from the Windows Store, if we see the
+        # WindowsApps path in the search path, replace it with
+        # dirname(sys.executable).
         appstore_dir = Path(os.environ['USERPROFILE']) / 'AppData' / 'Local' / 'Microsoft' / 'WindowsApps'
         paths = []
         for each in path.split(os.pathsep):
             if Path(each) != appstore_dir:
                 paths.append(each)
+            elif 'WindowsApps' in sys.executable:
+                paths.append(os.path.dirname(sys.executable))
         return os.pathsep.join(paths)
 
     @staticmethod
--- a/mesonbuild/interpreter.py
+++ b/mesonbuild/interpreter.py
@@ -41,6 +41,7 @@ import shutil
 import uuid
 import re
 import shlex
+import stat
 import subprocess
 import collections
 import functools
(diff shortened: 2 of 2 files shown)


mesonbuild/interpreter.py
```diff
@@ -2512,7 +2512,14 @@ def add_build_def_file(self, f):
         elif os.path.isfile(f) and not f.startswith('/dev'):
             srcdir = Path(self.environment.get_source_dir())
             builddir = Path(self.environment.get_build_dir())
-            f = Path(f).resolve()
+            f = Path(f)
+            try:
+                f = Path(f).resolve()
+            except:
```
dcbaker, line 2518: It looks like from the docs this should either be `except FileNotFoundError` or, if we're hitting a really bad case `except (FileNotFoundError, RuntimeError)`

mesonbuild/dependencies/base.py
```diff
@@ -1847,15 +1847,19 @@ def _windows_sanitize_path(path: str) -> str:
...
+        # installed application.
         appstore_dir = Path(os.environ['USERPROFILE']) / 'AppData' / 'Local' / 'Microsoft' / 'WindowsApps'
         paths = []
+        appstore_paths = []
         for each in path.split(os.pathsep):
             if Path(each) != appstore_dir:
                 paths.append(each)
-        return os.pathsep.join(paths)
+            else:
+                appstore_paths.append(each)
```
nirbheek, line 1861: Why are we adding `appstore_dir` multiple times in the end? Shouldn't we just add it once?

mesonbuild/interpreter.py
```diff
@@ -2512,7 +2513,16 @@ def add_build_def_file(self, f):
...
             builddir = Path(self.environment.get_build_dir())
-            f = Path(f).resolve()
+            try:
+                f = Path(f).resolve()
+            except OSError as e:
+                if e.errno == errno.EINVAL:
+                    # Magic WindowsApps executables aren't resolvable,
+                    # but still work... try making the path absolute
+                    # and hope for the best.
+                    f = Path(f).absolute()
```
nirbheek, line 2523: We need to find a way to detect whether the python wrapper in WindowsApps is an actual python or will just open a store link. Otherwise, https://github.com/mesonbuild/meson/pull/6459 will happen again. One way might be to check if the path is zero-sized or not:

```python
>>> pathlib.Path('C:/Users/nirbheek/AppData/Local/Microsoft/WindowsApps/python3.exe')
>>> s = p.stat()
>>> s
os.stat_result(st_mode=33279, st_ino=1970324836975681, st_dev=708247912, st_nlink=1, st_uid=0, st_gid=0, st_size=0, st_atime=1589576759, st_mtime=1589576759, st_ctime=1580153409)
>>> s.st_size
0
```

nirbheek: To clarify, the same also happens with `python.exe`. On my machine this is what it shows when I don't have the store app installed. Please check what happens on your machine :)

d-scott-phillips: If reordering the search path and overlooking the path.resolve() error isn't acceptable, maybe an alternative approach would be to special-case looking for python or python3. In that case we could drop the WindowsApps path from the search as now, but then also inject dirname(sys.executable) at the end of the search paths.

nirbheek: > In that case we could drop the WindowsApps path from the search as now, but then also inject dirname(sys.executable) at the end of the search paths.

I like this idea.

nirbheek: Poke :)

mesonbuild/dependencies/base.py
```diff
@@ -1847,13 +1848,21 @@ def _windows_sanitize_path(path: str) -> str:
...
+        # running on came from the Windows Store, if we see the
+        # WindowsApps path in the search path, replace it with
+        # dirname(sys.executable).
         appstore_dir = Path(os.environ['USERPROFILE']) / 'AppData' / 'Local' / 'Microsoft' / 'WindowsApps'
         paths = []
         for each in path.split(os.pathsep):
-            if Path(each) != appstore_dir:
+            if Path(each) == appstore_dir and 'WindowsApps' in sys.executable:
+                paths.append(os.path.dirname(sys.executable))
+            else:
```
eli-schwartz, line 1865: So, 
```
if Path(each) == appstore_dir and 'WindowsApps' not in sys.executable:
    paths.append(each)
```

Seems broken. Either make it an if inside of an if, or make the else into an elif.

d-scott-phillips: oops, good point. fixed

PR #5245, opened 2019-04-11
[WIP] Make ninja backend only use response files when needed

jon-turney:
Make ninja backend only use response files when the command line is long enough to need them.

This makes the output of 'ninja -v' useful more often (something like `cl @exec@exe/main.c.obj.rsp` is not very useful if you don't have the response file to look at; this is helpful when the response file (and/or the VM containing it) has been deleted by the time you get to look at the error) (see also discussion in https://github.com/ninja-build/ninja/pull/1223)

Also, per https://github.com/mesonbuild/meson/issues/271#issuecomment-378998318, writing rsp files on Windows is moderately expensive.

This also prevents rules which aren't used by any build statement from being emitted in the `build.ninja` file.

PR #7215, opened 2020-05-27
Use rsp files when linking many files with gnulike linker

### LINKED ISSUE #7212: Response files needed sometimes even on non-windows systems
A user porting www.flintlib.org to meson ran into a commandline length limit
when it tried to link > 3300 files into a single library, with error
```
ninja: fatal: posix_spawn: Argument list too long
```
https://www.in-ulm.de/~mascheck/various/argmax/ 
says cygwin has the tightest limit on commandline length, but FreeBSD also has a smallish limit.
I was able to reproduce the error on FreeBSD with the following script:
```
#!/bin/sh
set -ex
n=1
echo "project('toolong', 'c')" > meson.build
echo "srcs = [" >> meson.build
while test $n -lt 12000
do
    echo "int f$n(int x) { return x + $n; }" > foo$n.c
    echo " 'foo$n.c'," >> meson.build
    n=$(expr $n + 1)
done
echo "]" >> meson.build
echo "library('toolong', srcs)" >> meson.build
meson . btmp
cd btmp
ninja -v
```

CMake successfully avoids the problem by using a response file, and meson should, too.
(Though ideally it should avoid them when not needed, as they do make debugging a little harder.)

dankegel:
Even non-windows systems have commandline limits, so when linking
"too many" files, use a response file.  (Don't use them on normal-sized
link commands; they're obscure and would make developers unhappy.)

This patch uses a lame definition of "too many", and doesn't address static libraries,
but it may be enough for now.

Partially fixes #7212

### DIFF
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -1589,6 +1589,11 @@ int dummy;
                 self.add_rule(NinjaRule(rule, command, args, description,
                                         rspable=compiler.can_linker_accept_rsp(),
                                         extra=pool))
+                # Also output a LINKER_SHORT rule for short commandlines on non-windows systems
+                rule = '{}_LINKER_SHORT{}'.format(langname, self.get_rule_suffix(for_machine))
+                self.add_rule(NinjaRule(rule, command, args, description,
+                                        rspable=False,
+                                        extra=pool))
 
         args = [ninja_quote(quote_func(x)) for x in self.environment.get_build_command()] + \
             ['--internal',
@@ -1709,8 +1714,9 @@ https://gcc.gnu.org/bugzilla/show_bug.cgi?id=47485'''))
         else:
             deps = 'gcc'
             depfile = '$DEPFILE'
+        want_rsp = mesonlib.is_windows()
         self.add_rule(NinjaRule(rule, command, args, description,
-                                rspable=compiler.can_linker_accept_rsp(),
+                                rspable=(want_rsp and compiler.can_linker_accept_rsp()),
                                 deps=deps, depfile=depfile))
 
     def generate_pch_rule_for(self, langname, compiler):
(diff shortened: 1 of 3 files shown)


mesonbuild/backend/ninjabackend.py
```diff
@@ -2595,7 +2602,15 @@ def generate_link(self, target, outname, obj_list, linker, extra_args=None, stdl
...
-        elem = NinjaBuildElement(self.all_outputs, outname, linker_rule, obj_list, implicit_outs=implicit_outs)
+        # On Windows systems, always use rsp files.
+        # On other systems, only use them for long lines passed to dynamic linker (for now)
+        # FIXME: don't hardcode commandline length limit; do define it better
+        max_len = 1000
+        if linker_base == 'STATIC' or mesonlib.is_windows() or len(obj_list) > max_len:
+            rule = linker_rule
+        else:
+            rule = linker_short_rule
+        elem = NinjaBuildElement(self.all_outputs, outname, rule, obj_list, implicit_outs=implicit_outs)
```
jpakkane, line 2613: Extract the decision logic and its corresponding comment to its own method called `should_use_rsp_file` or something like that.

marc-h38:
More background info at https://groups.google.com/d/msg/mesonbuild/419IYNF0PAk/m7teKUbqAAAJ

jon-turney:
You might find the test case added in #5245 a useful basis for adding a test for this.

I suspect this won't work in some edge cases, as the quoting rules for response files aren't quite the same as shell, and meson has a pile of Windows-specific hacks to attempt to deal with that.

dankegel:
Thank you.  I did a cursory search for test cases, but missed that one.

jon-turney:
> a useful basis for adding a test for this.

Just to be clear, by that I meant this commit: https://github.com/mesonbuild/meson/pull/5245/commits/e69382a48a2d40ee07c366f11b29262d9f12a057. Sorry for any miscommunication.

dankegel:
Closing in favor of https://github.com/mesonbuild/meson/pull/7227

PR #7217, opened 2020-05-28
Added docs for all meson commands + corresponding unit test

TheQwertiest:
- Added a commands page describing all listed top level meson commands.
- Added a unit-tests that checks that all commands were documented.

Some problems that might need to be solved in this PR or maybe a later PR:
- There's [a page](https://mesonbuild.com/Creating-releases.html#page-description) that describes `ninja dist` which basically does the same as `meson dist`. It can either be removed altogether or amended to use `meson dist` instead of `ninja`.
- This page should be probably integrated in the [overview](https://mesonbuild.com/Running-Meson.html). But I can't think of a good way to do that currently. Overview page also uses ninja as *the way*, should be probably amended.
- `meson install` command links to [a page](https://mesonbuild.com/Installing.html), which again uses `ninja` as *the way* to install. Should be probably amended to use `meson` command instead.

### DIFF
--- a/docs/markdown/Commands.md
+++ b/docs/markdown/Commands.md
@@ -14,6 +14,18 @@ The most common workflow is to run [`setup`](#setup), followed by [`compile`](#c
 For the full list of all available options for a specific command use the following syntax:
 `meson COMMAND --help`
 
+### configure
+
+Reconfigures existing meson project.
+
+Examples:
+- `meson configure builddir`: list all available options.
+- `meson configure builddir -Doption=new_value`: change a single option
+
+Most arguments are the same as in [`setup`](#setup).
+
+Note: reconfiguring project will not reset options to their default values (even if they were changed in `meson.build`).
+
 ### compile
 
 *(since 0.54.0)*
--- a/docs/markdown/Commands.md
+++ b/docs/markdown/Commands.md
@@ -79,7 +79,7 @@ Noteable options:
 
 *(since 0.45.0)*
 
-Generates a basic meson setup from the template.
+Creates a basic set of build files based on a template.
 
 Example: `meson init -C sourcedir`
 
@@ -103,11 +103,13 @@ Notable introspect commands:
 
 *(since 0.47.0)*
 
-Installs the project to the folder specified in `setup`.
+Installs the project to the prefix specified in `setup`.
 
-Example: `meson install -C builddir`
+Examples: 
+- `meson install -C builddir`
+- `DESTDIR=/path/to/staging/area meson install -C builddir`
 
-See [this page](Installing.md) for more info.
+See [the installation documentation](Installing.md) for more info.
 
 ### rewrite
 
@@ -115,7 +117,7 @@ See [this page](Installing.md) for more info.
 
 Modifies the meson project.
 
-See [this page](Rewriter.md) for more info.
+See [the meson file rewriter documentation](Rewriter.md) for more info.
 
 ### setup
 
@@ -125,7 +127,7 @@ Configures a build directory for the meson project.
 
 Example: `meson setup builddir`
 
-See [this page](Running-Meson.md#configuring-the-build-directory) for more info.
+See [meson introduction page](Running-Meson.md#configuring-the-build-directory) for more info.
 
 ### subprojects
 
@@ -147,10 +149,10 @@ Run tests for the configure meson project.
 
 Example: `meson test -C builddir`
 
-See [this page](Unit-tests.md) for more info.
+See [the unit test documentation](Unit-tests.md) for more info.
 
 ### wrap
 
 An utility to manage WrapDB dependencies.
 
-See [this page](Using-wraptool.md) for more info.
+See [the WrapDB tool documentation](Using-wraptool.md) for more info.
--- a/docs/markdown/Commands.md
+++ b/docs/markdown/Commands.md
@@ -1,0 +1,156 @@
+# Command-line commands
+
+There are two different ways of invoking Meson. First, you can run it directly
+from the source tree with the command `/path/to/source/meson.py`. Meson may
+also be installed in which case the command is simply `meson`. In this manual
+we only use the latter format for simplicity.
+
+Meson is invoked using the following syntax:
+`meson [COMMAND] [COMMAND_OPTIONS]`
+
+This section describes all available commands and some of their notable options.
+
+For full list of available options for a specific command use the following syntax:
+`meson COMMAND --help`
+
+### compile
+
+*(since 0.54.0)*
+
+Builds a default or specified target of a configured meson project.
+
+Syntax: `meson compile [TARGET [TARGET...]] [options]`  
+Eample: `meson compile -C builddir`
+
+Note: specifying a target is available since 0.55.0.
+
+`TARGET` has the following syntax: `[PATH/]NAME[:TYPE]`.
+`NAME`: name of the target from `meson.build` (e.g. `foo` from `executable('foo', ...)`).
+`PATH`: path to the target relative to the root `meson.build` file. Note: relative path for a target specified in the root `meson.build` is `./`.
+`TYPE`: type of the target (e.g. `shared_library`, `executable` and etc)
+
+`PATH` and/or `TYPE` can be ommited if the resulting `TARGET` can be used to uniquely identify the target in `meson.build`.
+
+For example targets from the following code:
+```meson
+shared_library('foo', ...)
+static_library('foo', ...)
+executable('bar', ...)
+```
+can be invoked with `meson compile foo:shared_library foo:static_library bar`.
+
+Note: the following limitations for backends apply:
+- `ninja`: `custom_target` is not supported.
+- `vs`: `run_target` is not supported.
+
+Notable options:
+- `--jobs JOBS`: The number of worker jobs to run (if supported). If the value is less
+                 than 1 the build program will guess.
+- `--clean`: Cleans the build directory.
+
+### dist
+
+*(since 0.52.0)*
+
+Generates a release archive from the current source tree.
+
+Example: `meson dist -C builddir`
+
+This creates a file called `projectname-version.tar.xz` in the build
+tree subdirectory `meson-dist`. This archive contains the full
+contents of the latest commit in revision control including all the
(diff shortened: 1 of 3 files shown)


run_unittests.py
```diff
@@ -1464,7 +1464,6 @@ def test_all_functions_defined_in_ast_interpreter(self):
         astint = AstInterpreter('.', '', '')
         self.assertEqual(set(interp.funcs.keys()), set(astint.funcs.keys()))
 
-
```
nirbheek, line 1467: Spurious newline removal?

docs/markdown/Commands.md
```diff
@@ -0,0 +1,156 @@
...
+
+### install
+
+*(since 0.47.0)*
+
+Installs the project to the folder specified in `setup`.
+
+Example: `meson install -C builddir`
+
+See [this page](Installing.md) for more info.
```
nirbheek, line 110: Links using text like `this page` is considered an anti-pattern because it makes it harder for blind users to see what the link is for. Should use something like `See [the installation documentation](Installing.md) for more info`. Same for below.

docs/markdown/Commands.md
```diff
@@ -0,0 +1,156 @@
...
+Notable introspect commands:
+- `--buildoptions`: List all build options.
+- `--dependencies`: List external dependencies.
+- `--targets`: List top level targets.
+
+### install
+
+*(since 0.47.0)*
+
+Installs the project to the folder specified in `setup`.
```
nirbheek, line 106: Should also mention `DESTDIR` as an example here?

docs/markdown/Commands.md
```diff
@@ -0,0 +1,156 @@
...
+(especially the one tagged for the release).
+
+Noteable options:
+- `--include-subprojects`: Include source code of subprojects that have been used for the build.
+
+### init
+
+*(since 0.45.0)*
+
+Generates a basic meson setup from the template.
```
nirbheek, line 82: The phrase `meson setup` here can be confusing to newcomers. Probably best to use a different word than 'setup'. Such as: `Creates a basic set of build files based on a template`

docs/markdown/Commands.md
```diff
@@ -0,0 +1,156 @@
...
+
+There are two different ways of invoking Meson. First, you can run it directly
+from the source tree with the command `/path/to/source/meson.py`. Meson may
+also be installed in which case the command is simply `meson`. In this manual
+we only use the latter format for simplicity.
+
+Meson is invoked using the following syntax:
+`meson [COMMAND] [COMMAND_OPTIONS]`
+
+This section describes all available commands and some of their notable options.
```
nirbheek, line 11: I think for newbies, it might be best to list the most common commands and also the order in which you'd use them (generally). F.ex.:

```suggestion
This section describes all available commands and some of their notable options. The most common workflow is to run [`setup`](#setup), followed by [`compile`](#compile), and then [`install`](#install).
```

TheQwertiest: We already have `Overview` though that describes the usual workflow.

docs/markdown/Commands.md
xclaesse: instead of "folder" I would say "prefix" here. +1 about mentioning DESDIR too.

docs/markdown/Commands.md
```diff
@@ -0,0 +1,156 @@
...
+
+Manages subprojects of the meson project.
+
+Syntax: `meson subprojects SUBPROJECTS_COMMAND [subprojects_command_options]`
+
+Notable subprojects commands:
+- `download`: Ensure subprojects are fetched, even if not in use. Already downloaded subprojects
+              are not modified. This can be used to pre-fetch all subprojects and avoid
+              downloads during configure.
+- `foreach`: Execute a command in each subproject directory.
```
xclaesse, line 142: There are other commands too.

docs/markdown/Commands.md
```diff
@@ -0,0 +1,156 @@
...
+static_library('foo', ...)
+executable('bar', ...)
+```
+can be invoked with `meson compile foo:shared_library foo:static_library bar`.
+
+Note: the following limitations for backends apply:
+- `ninja`: `custom_target` is not supported.
+- `vs`: `run_target` is not supported.
+
+Notable options:
```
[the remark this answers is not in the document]
TheQwertiest: Well, it's a decision that you (that is core meson developers) will have to make :P
I can go either way - document everything or document only important things. But it should be noted that even the current linked doc pages do not describe every available option, so there will be another decision to make: should there be a doc for every option here or should they be placed in a relevant page?

docs/markdown/Commands.md
```diff
@@ -0,0 +1,156 @@
...
+*(since 0.54.0)*
+
+Builds a default or specified target of a configured meson project.
+
+Syntax: `meson compile [TARGET [TARGET...]] [options]`  
+Eample: `meson compile -C builddir`
+
+Note: specifying a target is available since 0.55.0.
+
+`TARGET` has the following syntax: `[PATH/]NAME[:TYPE]`.
```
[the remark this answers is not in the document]
TheQwertiest: It's not (which is noted in PR description >:( ). But @jpakkane in that PR noted that he wanted the docs for `meson compile` to be amended and we didn't have any. I interpreted that as "I approve this PR, but docs are needed in some way". 

I can remove this part, wait till this PR is merged and then add it back in the `compile target` PR, if you want.

docs/markdown/Commands.md
nirbheek: Yes, but we still have to cater for people that show up to this page as their first piece of documentation from Google. Duplication is good sometimes.

run_unittests.py
TheQwertiest: Fixed

docs/markdown/Commands.md
TheQwertiest: Fixed

docs/markdown/Commands.md
TheQwertiest: Fixed

docs/markdown/Commands.md
TheQwertiest: Fixed

TheQwertiest:
PS: Sorry for the repeated pushes, but I currently don't have a machine where I can run those tests locally =(

docs/markdown/Commands.md
TheQwertiest: Fixed

TheQwertiest:
Welp, unit test worked: I've missed `configure` command. Amended.

TheQwertiest:
@nirbheek , @xclaesse , there is one remaining question so that this issue could be ready to be merged (aside from yet-to-be-merged `targets` from `meson compile`, which I will remove): 
- Should this page document *all* command options? 
  - If so, what should be done for pages that it links to (e.g. https://mesonbuild.com/Unit-tests.html)?
One way would be to only make short description here and point to those separate pages for more detailed info.

nirbheek:
How hard would it be to make it so that the unit test errors out if a new option is added to any command (or a new command is added) without documenting it here? There's some prior art for this in the unit tests already (for the reference tables).

If that can be done, I think it's good to document everything here and not worry about duplication.

TheQwertiest:
> How hard would it be to make it so that the unit test errors out if a new option is added to any command (or a new command is added) without documenting it here? There's some prior art for this in the unit tests already (for the reference tables).

Dunno :D
The main problem would be parsing the page, so options and sub-commands(if any) need to have some unique identifiable format. But I can try to implement it. (yay! more Py3.5 regex T_T)

TheQwertiest:
Added test, hopefully they will fail in the expected way :D

TheQwertiest:
Note: this test does *not* check the arguments of subcommands (e.g. `meson rewrite target --help`)

TheQwertiest:
Yup, it works: https://github.com/mesonbuild/meson/pull/7217/checks?check_run_id=728026565#step:4:839

Now onto the documenting part,,,

TheQwertiest:
All done, apart from the issues that are described in PR description.

TheQwertiest:
Another thing: I didn't mark any arguments with `(since version 0.x)`, since I don't have enough motivation to do that for *ALL* of those.

TheQwertiest:
It should be also noted, that RegExp is pretty dumb, so if someone will add an optional/positional/command/value/whatever name that contains some unexpected magical characters, the test will still pass, despite new argument being undocumented.

TheQwertiest:
@nirbheek , A random thought: what about making this doc page generation dynamic? This would be the most correct and maintainable way and will remove the need to copy-paste info from `meson cmd --help`.
Actually... are there any dynamic docs in meson right now?

TheQwertiest:
So each section would be a template like
```
### Command
Short description

%cmd_output% < placeholder for replacement

Examples...
Additional description...
```
And the unit test will only need to check for section names (which could be also made dynamic, but that would be harder to implement)

TheQwertiest:
ping?

TheQwertiest:
Replaced manually edited `--help` with an actual `--help` output. This greatly simplifies the documenting process and makes the unit test actually reliable.

Note: the test will fail for now, since `argparse` `choices` are not sorted in any order and may be different with every invocation. I will try to fix it in all respective commands.

TheQwertiest:
All done. CI failures are irrelevant (https://github.com/mesonbuild/meson/issues/7307).

docs/markdown/Commands.md
```diff
@@ -0,0 +1,604 @@
...
+                  [--wrap-mode {default,nofallback,nodownload,forcefallback}]
+                  [--pkg-config-path PKG_CONFIG_PATH]
+                  [--build.pkg-config-path BUILD.PKG_CONFIG_PATH]
+                  [--cmake-prefix-path CMAKE_PREFIX_PATH]
+                  [--build.cmake-prefix-path BUILD.CMAKE_PREFIX_PATH]
+                  [-D option] [--clearcache]
+                  [builddir]
+```
+
+Reconfigures existing meson project.
```
nirbheek, line 45: I think this needs more clarification. `reconfigure` has a completely different meaning with other build systems (aka, re-setup) :)

TheQwertiest: Fixed

PR #7220, opened 2020-05-28
wrap: Add patch_directory support

xclaesse:
I'm just wondering if we should force patch_directory inside `packagefiles` subdir?

### DIFF
--- a/docs/markdown/snippets/wrap_patch.md
+++ b/docs/markdown/snippets/wrap_patch.md
@@ -4,3 +4,11 @@ It is now possible to use the `patch_filename` and `source_filename` value in a
 `.wrap` file without `*_url` to specify a local source / patch file. All local
 files must be located in the `subprojects/packagefiles` directory. The `*_hash`
 entries are optional with this setup.
+
+## Local wrap patch directory
+
+Wrap files can now specify `patch_directory` instead of `patch_filename` in the
+case overlay files are local. Every files in that directory, and subdirectories,
+will be copied to the subproject directory. This can be used for example to add
+`meson.build` files to a project not using Meson build system upstream.
+The patch directory must be placed in `subprojects/packagefiles` directory.
--- a/mesonbuild/wrap/wrap.py
+++ b/mesonbuild/wrap/wrap.py
@@ -126,9 +126,6 @@ class PackageDefinition:
             m = 'Missing key {!r} in {}'
             raise WrapException(m.format(key, self.basename))
 
-    def has_patch(self) -> bool:
-        return 'patch_filename' in self.values
-
 def load_wrap(subdir_root: str, packagename: str) -> PackageDefinition:
     fname = os.path.join(subdir_root, packagename + '.wrap')
     if os.path.isfile(fname):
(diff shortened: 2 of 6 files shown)


mensinda:
> I'm just wondering if we should force patch_directory inside `packagefiles` subdir?

Yes, I would say we should do that. But then I would recommend renaming `packagefiles` to something diffrerent (`localresources` maybe? Naming things is hard).

xclaesse:
I agree the naming is a bit unfortunate, but I don't think it's worth adding yet another directory.

xclaesse:
PR updated to use packagefiles directory.

mensinda:
Rather than adding a new directory, I wanted to suggest to rename the existing `packagefiles` dir, since it hasn't been released yet.

xclaesse:
oh, forgot that it was not yet released, good point.

xclaesse:
@mensinda tbh, I don't think the naming is that bad. One advantage is that `packagecache` and `packagefiles` are sorted together in when listing dirs in subprojects/. Any objection on merging this PR?

mensinda:
Implementation wise, this PR is good to go. I also don't have any strong feelings regarding the directory name (directory sorting is indeed a good point).

Since this is technically a new feature I would say that @jpakkane should have the last word whether `packagefiles` should stay or not.

Also, it would be a good idea to explicitly document that this *only* works for patches. Otherwise, we are likely to get bug reports :)

docs/markdown/snippets/wrap_patch.md
```diff
@@ -4,3 +4,9 @@ It is now possible to use the `patch_filename` and `source_filename` value in a
 `.wrap` file without `*_url` to specify a local source / patch file. All local
 files must be located in the `subprojects/packagefiles` directory. The `*_hash`
 entries are optional with this setup.
+
+## Local wrap patch directory
+
+Wrap files can now specify `patch_directory` instead of `patch_filename` in the
+case overlay files are local. The patch directory must be placed in
+`subprojects/packagefiles` directory.
```
jpakkane, line 12: "Overlay" is not an immediately obvious term. This should explain briefly what it does (e.g. "copies files in this dir unaltered over the source dir").

xclaesse: done.

xclaesse: Note that the doc already uses the term 'overlay' for this, see https://mesonbuild.com/Wrap-dependency-system-manual.html#specific-to-wrapfile. But I agree it's better to clarify in the release snippet too.

mesonbuild/wrap/wrap.py
```diff
@@ -422,13 +418,22 @@ def get_file_internal(self, what: str) -> str:
...
     def apply_patch(self) -> None:
-        path = self.get_file_internal('patch')
-        try:
-            shutil.unpack_archive(path, self.subdir_root)
-        except Exception:
-            with tempfile.TemporaryDirectory() as workdir:
-                shutil.unpack_archive(path, workdir)
-                self.copy_tree(workdir, self.subdir_root)
+        if 'patch_filename' in self.wrap.values:
+            path = self.get_file_internal('patch')
```
jpakkane, line 422: This should check if `patch_directory` is also defined and error out if it is. Otherwise the file is ambiguous.

xclaesse: Good point, fixed now.

PR #7229, opened 2020-05-31
glib gtkdoc fixes for Cygwin

jon-turney:
Fixes problems building glib documentation as reported at https://cygwin.com/pipermail/cygwin-apps/2020-May/040144.html

Test case in a separate PR #7274, since it also exposes some pre-existing problems.

### DIFF
--- a/mesonbuild/scripts/gtkdochelper.py
+++ b/mesonbuild/scripts/gtkdochelper.py
@@ -16,7 +16,7 @@ import sys, os
 import subprocess
 import shutil
 import argparse
-from ..mesonlib import MesonException, Popen_safe, is_windows, split_args
+from ..mesonlib import MesonException, Popen_safe, is_windows, is_cygwin, split_args
 from . import destdir_join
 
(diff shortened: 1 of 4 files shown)


mesonbuild/scripts/gtkdochelper.py
```diff
@@ -65,6 +65,9 @@ def gtkdoc_run_check(cmd, cwd, library_paths=None):
             library_paths.extend(env['LD_LIBRARY_PATH'].split(os.pathsep))
         env['LD_LIBRARY_PATH'] = os.pathsep.join(library_paths)
 
+    if is_windows():
+        cmd.insert(0, sys.executable)
```
nirbheek, line 69: Hm, didn't we already do this above `if is_windows() or is_cygwin():`?

jon-turney: Huh.  I thought I removed that, since we don't want it in the cygwin case.  Well spotted.

FIxed.

PR #7230, opened 2020-05-31
mesonlib.py: refactored detect_vcs() to use pathlib.Path

a1phabeta:


jpakkane:
According to the failing test, you need to update the annotations:

```
mesonbuild/mesonlib.py:555:18: error: Incompatible types in assignment (expression has type "Path", variable has type "str")
mesonbuild/mesonlib.py:556:47: error: "str" has no attribute "parents"
```

PR #7231, opened 2020-05-31
cmake: Add more advanced subproject configuration options

mensinda:
This is done with the new cmake subprojects options object that is similar to the already existing configuration data object. It is consumed by the new `options` kwarg of the cmake.subproject function.


docs/markdown/CMake-module.md
```diff
@@ -103,7 +144,37 @@ and supports the following methods:
    the subproject. Usually `dependency()` or `target()` should be
    preferred to extract build targets.
  - `found` returns true if the subproject is available, otherwise false
-   *new in in 0.53.2*
+   *new in meson 0.53.2*
+
+### `cmake options` object
+
+This object is returned by the `subproject_options()` function and consumed by
```
jpakkane, line 151: This should be "method", not "function".

mensinda: Just to clarify, are all root functions of a module methods? Up until now, I have only used the term "method" for functions/methods of an object and not a module.

If "method" should also for module functions/methods I will change it for all occoruences in the CMake module docs.

jpakkane: I don't know if "method" is the correct term here or even if an actual proper term even exists. The difference I had in mind is that "functions" look like this:

    freestading_function()

whereas methods are always invoked via something else like this:

    some.bound_function()

That is clearly the case when `some` is an object but it's less clear when it is a "module" or something similar. Anyhow, since this would require changing all the docs, let's go with this for now.

mensinda: OK, sure. Does this PR get a dev freeze exception then :)

Issue #7218, opened 2020-05-28
Want --wrap-mode=forcefallback:glib,foo,bar


tp-m:
**Describe the bug**
Currently there is `--wrap-mode=forcefallback` but this will then apply to *all* possible fallback subprojects.

Often what people want is to only force certain fallbacks, like e.g. force to use the included glib or ffmpeg subproject but not any others.

**To Reproduce**
Can be reproduced by building [gst-build](https://gitlab.freedesktop.org/gstreamer/gst-build) with `meson --wrap-mode=force-fallback:glib` while a recent glib plus development headers are installed.

**Expected behavior**
With `meson --wrap-mode=force-fallback:glib` I would expect the build to use the glib subproject instead of the system-installed glib.

tp-m:
**For bonus points**
Perhaps also allow `--force-fallback=glib,-openh264` to force `glib` but disallow fallbacks to be used for `openh264`

xclaesse:
I totally agree it would be nice to have a per-subproject setting here.

I'm not really a big fan of adding complex syntax for --wrap-mode command line, just throwing ideas, what about splitting those options: `--force-fallback=glib --no-fallback=openh264` ? Seems more explicit to me.

tp-m:
I agree, that seems much nicer!

Would you allow passing a comma-separated list like `--force-fallback=glib,libffi` ?

xclaesse:
> Would you allow passing a comma-separated list like `--force-fallback=glib,libffi` ?

Yes that's the idea, we already have support for list options.

MathieuDuponchelle:
@xavierclaessens , good thing you prefer a separate option because adding this to `wrap-mode` would actually have been quite a pain: meson registers wrap-mode with argparse as an argument with a fixed set of choices, I don't think it would have been possible to teach argparse to accept `forcefallback:foo,bar` as one of the choices.

I've named the option `--force-fallback-for=`, I like it a bit better than `--force-fallback`. Regarding `no-fallback` / `no-fallback-for`, it shouldn't be too difficult to implement either, but to do things right one will also probably want to perform some validation at command line parsing time, in order to make sure the same dependency isn't listed both as `no-fallback` and `force-fallback`, which would be ambiguous.

One could also not error out and instead just add / remove from one list and to the other as options are parsed (like `-W` / `-Wno` options are parsed by compilers), but I don't know how much access we have to that information at the meson level.

PR #7242, opened 2020-06-02
interpreter: add support for --force-fallback-for

MathieuDuponchelle:
This new command line option allows specifying dependencies for
which to force fallback.

See the documentation for more information

Fixes: #7218

### DIFF
--- a/docs/markdown/Builtin-options.md
+++ b/docs/markdown/Builtin-options.md
@@ -86,6 +86,7 @@ Using the option as-is with no prefix affects all machines. For example:
 | warning_level {0, 1, 2, 3}           | 1             | Set the warning level. From 0 = none to 3 = highest            | no             |
 | werror                               | false         | Treat warnings as errors                                       | no             |
 | wrap_mode {default, nofallback,<br>nodownload, forcefallback} | default | Wrap mode to use                            | no             |
+| force_fallback_for                   | []            | Force fallback for those dependencies                          | no             |
 
 <a name="build-type-options"></a>
 For setting optimization levels and toggling debug, you can either set the
--- a/docs/markdown/Subprojects.md
+++ b/docs/markdown/Subprojects.md
@@ -212,6 +212,9 @@ the following command-line options:
     calls, and those are meant to be used for sources that cannot be
     provided by the system, such as copylibs.
 
+    This option may be overriden by `--force-fallback-for` for specific
+    dependencies.
+
 * **--wrap-mode=forcefallback**
 
     Meson will not look at the system for any dependencies which have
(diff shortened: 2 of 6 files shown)


tp-m:
What about its little brother `--no-fallback-for=xyz`? :)

MathieuDuponchelle:
> What about its little brother `--no-fallback-for=xyz`? :)

See my comment @ https://github.com/mesonbuild/meson/issues/7218#issuecomment-637821295 :)

xclaesse:
Looks good to me. I prefer without the `-for` suffix but I can live with that.

There is a corner case we should maybe clarify: `--wrap-mode=nodownload --force-fallback-for=glib`. In that case it will work only if glib has already been downloaded. Maybe just mention that in the doc?

MathieuDuponchelle:
> Looks good to me. I prefer without the -for suffix but I can live with that.

I implemented it, I get to choose :P More seriously, in case someone wants to give a tie breaker feel free to comment, not too hard to change

> There is a corner case we should maybe clarify: --wrap-mode=nodownload --force-fallback-for=glib. In that case it will work only if glib has already been downloaded. Maybe just mention that in the doc?

Good point, I only mentioned the interaction with `nofallback`, I'll add a quick note

dabrain34:
I like this change 
I think it would make sense to define this force fallback also in a given project
 
I mean if I understand well, you need to pass the `--force-fallback-for=glib` to force the use of fallback glib to the configure command line

xclaesse:
> I think it would make sense to define this force fallback also in a given project `meson.build` file as a default and overriden by the command line if necessary

Do you have a specific use-case for that in mind? I don't think we should hold this PR for that, we can find a way later if needed. 

Looks all good to me, LGTM.

xclaesse:
Oh, mmm, using this new feature and I just realized that you need to pass `glib-2.0` (the dependency  name) and not `glib` (the subproject/wrap name). Shouldn't we allow both?

dcbaker:
This also broke CI since it added a new command line option (which we have tests for apparently)

xclaesse:
Should be fixed by https://github.com/mesonbuild/meson/pull/7335

xclaesse:
@MathieuDuponchelle What do you thing about https://github.com/mesonbuild/meson/pull/7334 ?

MathieuDuponchelle:
Thanks for the fix @xclaesse , not sure why CI didn't complain here! Re #7334 it seems safe to me, and indeed friendlier.

xclaesse:
Not your fault, it's a new test that got merged after this pr ran it's ci.

dabrain34:
> > I think it would make sense to define this force fallback also in a given project `meson.build` file as a default and overriden by the command line if necessary
> 
> Do you have a specific use-case for that in mind? I don't think we should hold this PR for that, we can find a way later if needed.

Yes the idea would be to control that for a given dependency we prefer the fallback. Let's imagine we we are building libdrm for gstreamer plugin with a specific revision with a specific project, would be nice to have a way to force always the subropoject fallback and not have to pass always the command line in order to avoid the use of the system lib

> 
> Looks all good to me, LGTM.

MathieuDuponchelle:
@dabrain34 , not sure I see the point, why not just use subproject() then?

xclaesse:
@dabrain34 if for your project you always want to fallback on libdrm, I would suggest to write a native file (or cross file depending on your case) and set all options needed, such as `force_fallback_for='libdrm'` in `[built-in options]` section, once we land https://github.com/mesonbuild/meson/pull/6597. That way you only have 1 config file to pass on the command line. Does that make sense for your use-case?

dabrain34:
It might have misunderstood as I was confused by the previous error message
So if I want a project to use an external library version, you have to include the subproject before right ?

PR #7243, opened 2020-06-02
Allow passing an array to library(vs_module_defs)

dcbaker:
There are some cases where we want to use a def file with one compiler,
but not another. For example, you might want to use one with msvc,
clang-cl, and intel-cl, but not with mingw. Previously this required
duplicating the library call, one for the compilers with the def file,
and one without. This allows us to pass an array, possibly of zero
elements.

### DIFF
--- a/docs/markdown/Reference-manual.md
+++ b/docs/markdown/Reference-manual.md
@@ -1461,9 +1461,13 @@ extra keyword arguments.
   or if it's not a list, the specified value will be used for setting
   both compatibility version and current version. If unspecified, the
   `soversion` will be used as per the aforementioned rules.
-- `vs_module_defs` a string, a File object, or Custom Target for a
-  Microsoft module definition file for controlling symbol exports,
-  etc., on platforms where that is possible (e.g. Windows).
+- `vs_module_defs` a string, a File object, Custom Target, or an array of
+  those for a Microsoft module definition file for controlling symbol
+  exports, etc., on platforms where that is possible (e.g. Windows). If a
+  an array is passed it should have a length of 1 or 0. An empty array may
+  used if module defs are only needed in some situations, such as not using
+  one with mingw.
+  *(new in 0.55.0)* Array options.
 
 ### shared_module()
 
@@ -1485,10 +1489,8 @@ you will need to set the `export_dynamic` argument of the executable to
 
 Supports the following extra keyword arguments:
 
-- `vs_module_defs`, *(Added 0.52.0)*, a string, a File object, or
-  Custom Target for a Microsoft module definition file for controlling
-  symbol exports, etc., on platforms where that is possible
-  (e.g. Windows).
+- `vs_module_defs`, *(Added 0.52.0)*, See the `vs_module_defs` entry of
+  [`shared_library`](#shared_library) for more information.
 
 **Note:** Linking to a shared module is not supported on some
 platforms, notably OSX.  Consider using a
--- a/docs/markdown/snippets/empty_vs_module_defs.md
+++ b/docs/markdown/snippets/empty_vs_module_defs.md
@@ -1,0 +1,9 @@
+## library vs_module_defs may be an array
+
+It is sometimes desirable to use a .def file only in some situations, such as
+with msvc, but now with mingw. Previously this meant defining a duplicate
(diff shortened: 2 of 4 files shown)


jon-turney:
Did you consider allowing an empty string to have this special meaning?

dcbaker:
I thought about it, but in most cases we use an empty array to mean "no value"

jon-turney:
> but in most cases we use an empty array to mean "no value"

True (except where we don't) :)

I guess what I find irrregular about this is the 0 or 1-item list part.  How about permitting `vs_module_defs:` to be (something that can be coerced to) a file, `[]`, or maybe `''`?

dcbaker:
Here's my thought on the 1 or 0 length thing, this should work:
```meson
def = []
# our .def file isn't used with mingw
if host_machine.system() == 'windows' and cc.get_id() != 'gcc'
    def += 'def.def'
endif
```
Otherwise you'll get an annoying error about how `[]` is valid, and `def.def` is valid, but `['def.def']` is not.

jon-turney:
> this should work

Well, perhaps.  I think it would be more normal to write `=` rather than `+=` ,  since it doesn't make much sense to construct `vs_module_defs:` by concatenation, since it can only be a single value?

I'm just worried that this is a bit inconsistent with other places where values and empty lists are accepted, but non-empty lists aren't (e.g `library(name_suffix:)`?

Or if list flattening everywhere means that values and single-item lists are syntactically equivalent, then that should be mentioned somewhere other than the documentation for this keyword.

nirbheek:
The other day I was thinking how we should standardize on a value that signifies "use the default value for this kwarg", similar to how in Python you can pass `None` as the value for all kwargs to mark them as 'unset'.

`[]` is not very discoverable for that (and I say this as the person who started this trend in the first place...). `false` is slightly better, but it is also a valid value for a kwarg. Maybe we want to make a new constant?

Another question is, why not use something like

```meson
library_kwargs = {}
if cc.get_id() == 'msvc'
  library_kwargs += {'vs_module_defs': 'libfoo.def'}
endif
library(... kwargs: library_kwargs)
```

This does the same thing without duplicating the `library()` call.

jpakkane:
Much like `disabler` we could perhaps have a `kwargunsetter` object, whose only use is that if it is passed to a kwarg, the outcome is the same as if the kwarg had not been written in the code at all?

nirbheek:
> Much like `disabler` we could perhaps have a `kwargunsetter` object, whose only use is that if it is passed to a kwarg, the outcome is the same as if the kwarg had not been written in the code at all?

We could generalize it to have an object `unsetter()` which unsets whatever you set it to, whether that's a variable or a kwarg.

jpakkane:
That would be problematic in the following case:

```
opt_kwarg = something
if something_else
   opt_kwarg = unsetter()
endif
some_function(..., foo: opt_kwarg)
```

If unsetting unsets the variable, then the last line leads to a missing variable error.

nirbheek:
True, that's a bad interaction. I'm not sure I like `kwargunsetter()` as-is. Feels like a super-specific thing that already has a different solution.

jon-turney:
> If unsetting unsets the variable, then the last line leads to a missing variable error.

Not intractable, if

```
opt_kwarg = something
if something_else
   opt_kwarg = none()
endif
some_function(..., foo: opt_kwarg)
```
where `none()` returns a value which is distinguished from all other values, and a kwarg with that value is defined to behave the same as if was absent.

But I'm not sure this is an improvement.  How is `[none()]` or `['something', none()]` supposed to behave? and this has drifted pretty far from the original problem, which already has a solution with kwarg dicts.

xclaesse:
> ```
> opt_kwarg = something
> if something_else
>    opt_kwarg = none()
> endif
> some_function(..., foo: opt_kwarg)
> ```

Can be done, but not pretty:
```meson
d = {'foo': something}
if something_else
  d = {}
endif
some_function(..., kwargs: d)
```

bonzini:
FWIW I had a similar issue with `static: false` being different from nothing at all, and solved it with `kwargs`.

dcbaker:
I think we'd discussed having a `none()` or `default()` for solving some of these cases. I find the `kwargs` thing really hard to follow as it adds another level of indirection, but this clearly is not the solution we want.

PR #7227, opened 2020-05-31
Make ninja backend use response files when needed on linux

dankegel:
This is https://github.com/mesonbuild/meson/pull/5245 rebased to master, with a little linux support from https://github.com/mesonbuild/meson/pull/7215.

Goal is to fix https://github.com/mesonbuild/meson/issues/7212 

Note: the order in which github displays commits here seems... odd.  Here's what git log says (newest first):
```
6de31e19 ninja: rip off the rsp bandaid
acdbea35 ninja: long commandlines cannot warn, that breaks unit/68 static link
bf3e760d ninja: Add ninja variable DEPFILE_UNQUOTED with unquoted DEPFILE value
d56cff53 ninja: Specifically implement gcc rspfile style quoting
b6ffcaf8 backend: There is no need for 'compiler-specific escaping' anymore
03afad82 ninja: Implement Windows-style command line quoting
286c8d82 ninja: Quoting in rspfile depends on the compiler, not the shell
bf7df9ed ninja: Push ninja and shell quoting down into NinjaRule
a268556a 145 special chars: Document and optionally disable cases that currently fail, provide more info on failure.
2c781370 Extended test case for special characters to compiler arguments
da74dcbf tidy up wording in test
becfc801 ci: run some builders with response file threshold set to zero
eeec4cbe ninja: response file threshold now more accurate, overridable.  Corresponding test more accurate, too.
261da124 docs: add snippet about response file change
5ff0e967 gnulike linkers (and ar) accept rsp files.
fdba4398 ninja: don't warn on rsp files for windows
64750928 ninja: always use rsp files on Windows... for now.
ed53bd43 ninja: Refcount rsp and non-rsp rule usage separately
44477697 ninja: Expose response file rules in compdb
baeae882 ninja: Only use response files when needed
9ca5c2c2 ninja: Rename 'rule' -> 'rulename' in NinjaBuildElement 
683209bb Add a test case for very long command lines
```

PR #7219, opened 2020-05-28
Escaping special characters compiler arguments redux

dankegel:
This is https://github.com/mesonbuild/meson/pull/5322 plus a commit or two.

Just checking where the failures are by abusing the merge request process... please ignore (unless you're interested in https://github.com/mesonbuild/meson/pull/5245 )

Issue #271, opened 2015-09-28
Support rspfile in ninja backend


jimon:
On windows there is a command line limit, so longer calls from ninja with just fail with "CreateProcess: The parameter is incorrect."

To reproduce it just do :

     project('app', 'cpp')
     executable('app', 'test1.cpp', 'test2.cpp', ... , 'test1000.cpp')

And generate build.ninja file. In this case backend will generate one link build command with all 1000 object files. And total length of this command will be longer that maximum command line length on windows.

To fix it you need to use [rspfile in ninja](http://martine.github.io/ninja/manual.html#ref_rule)

jimon:
Awesome ! :+1:

nico:
We were talking about this issue elsewhere; I wanted to chime in on the fix here:

Re meson#271:

```
>>> len('test1.cpp') * 1000
9000
```

It sounds like meson sends link commands through the shell which has a limit of 8k. If you don't send links through the shell, the limit is 32k.

Also, the example here is combining compile and link of 1000 cpp files in one invocation. In practice, it's more likely to have 1000 compile actions taking cc -> o and then one link taking 1000 o files.

Writing rsp files on Windows is expensive. If you unconditionally use them for every compile, each compile will be a bit slower. In chromium, with a distributed build system that makes the actual compile fast, stopping to use rsp files for compiles made the build 50% or so faster (https://bugs.chromium.org/p/chromium/issues/detail?id=796021#c7) since rspfiles are written serially by ninja but the compiles run in parallel.

So I'd recommend that you don't use rspfiles by default for compiles (for links it's ok since there are way fewer of them). I don't know if meson makes a compile/link distinction and if that's feasible to implement; but I thought I should mention it.

PR #7245, opened 2020-06-03
Make ninja backend only use response files when needed, on linux too

dankegel:
(This is #5245 rebased to master, with a few fixes, a little quote testing from #7219, and a little rspfile support detection added.  It supersedes the less tidy versions in #7215 and #7227.)

Make ninja backend only use response files when the command line is long enough to need them.

This makes the output of 'ninja -v' useful more often (something like cl @exec@exe/main.c.obj.rsp is not very useful if you don't have the response file to look at; this is helpful when the response file (and/or the VM containing it) has been deleted by the time you get to look at the error) (see also discussion in ninja-build/ninja#1223)

Also, per #271 (comment), writing rsp files on Windows is moderately expensive.

This also prevents rules which aren't used by any build statement from being emitted in the build.ninja file.

It also enables response files on Linux, thereby fixing #7212  

Unit test test_internal_include_order is skipped on msys2 when rsp files are forced,
otherwise it would fail because run_unittests.py's get_compdb() fails to handle rspfile quoting.

### DIFF
--- a/docs/markdown/snippets/response-files.md
+++ b/docs/markdown/snippets/response-files.md
@@ -1,0 +1,7 @@
+## Response files enabled on Linux, reined in on Windows
+
+Meson used to always use response files on Windows,
+but never on Linux.
+
+It now strikes a happier balance, using them on both platforms,
+but only when needed to avoid command line length limits.
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -28,7 +28,7 @@ from .. import build
 from .. import dependencies
 from .. import mesonlib
 from .. import mlog
-from ..compilers import CompilerArgs, VisualStudioLikeCompiler
+from ..compilers import CompilerArgs
 from ..mesonlib import (
     File, MachineChoice, MesonException, OrderedSet, OptionOverrideProxy,
     classify_unity_sources, unholder
@@ -607,29 +607,13 @@ class Backend:
 
     @staticmethod
     def escape_extra_args(compiler, args):
-        # No extra escaping/quoting needed when not running on Windows
-        if not mesonlib.is_windows():
-            return args
+        # all backslashes in defines are doubly-escaped
         extra_args = []
-        # Compiler-specific escaping is needed for -D args but not for any others
-        if isinstance(compiler, VisualStudioLikeCompiler):
-            # MSVC needs escaping when a -D argument ends in \ or \"
-            for arg in args:
-                if arg.startswith('-D') or arg.startswith('/D'):
-                    # Without extra escaping for these two, the next character
-                    # gets eaten
-                    if arg.endswith('\\'):
-                        arg += '\\'
-                    elif arg.endswith('\\"'):
-                        arg = arg[:-2] + '\\\\"'
-                extra_args.append(arg)
-        else:
-            # MinGW GCC needs all backslashes in defines to be doubly-escaped
-            # FIXME: Not sure about Cygwin or Clang
-            for arg in args:
-                if arg.startswith('-D') or arg.startswith('/D'):
-                    arg = arg.replace('\\', '\\\\')
-                extra_args.append(arg)
+        for arg in args:
+            if arg.startswith('-D') or arg.startswith('/D'):
+                arg = arg.replace('\\', '\\\\')
+            extra_args.append(arg)
+
         return extra_args
 
     def generate_basic_compiler_args(self, target, compiler, no_warn_args=False):
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -121,14 +121,14 @@ class NinjaRule:
         outfile.write('\n')
 
 class NinjaBuildElement:
-    def __init__(self, all_outputs, outfilenames, rule, infilenames, implicit_outs=None):
+    def __init__(self, all_outputs, outfilenames, rulename, infilenames, implicit_outs=None):
         self.implicit_outfilenames = implicit_outs or []
         if isinstance(outfilenames, str):
             self.outfilenames = [outfilenames]
         else:
             self.outfilenames = outfilenames
-        assert(isinstance(rule, str))
-        self.rule = rule
+        assert(isinstance(rulename, str))
+        self.rulename = rulename
         if isinstance(infilenames, str):
             self.infilenames = [infilenames]
         else:
@@ -166,7 +166,7 @@ class NinjaBuildElement:
         implicit_outs = ' '.join([ninja_quote(i, True) for i in self.implicit_outfilenames])
         if implicit_outs:
             implicit_outs = ' | ' + implicit_outs
-        line = 'build {}{}: {} {}'.format(outs, implicit_outs, self.rule, ins)
+        line = 'build {}{}: {} {}'.format(outs, implicit_outs, self.rulename, ins)
         if len(self.deps) > 0:
             line += ' | ' + ' '.join([ninja_quote(x, True) for x in self.deps])
         if len(self.orderdeps) > 0:
@@ -901,8 +901,8 @@ int dummy;
         self.build_elements.append(build)
 
         # increment rule refcount
-        if build.rule != 'phony':
-            self.ruledict[build.rule].refcount += 1
+        if build.rulename != 'phony':
+            self.ruledict[build.rulename].refcount += 1
 
     def write_rules(self, outfile):
         for r in self.rules:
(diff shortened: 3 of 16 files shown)


dankegel:
I tested with gst-build on ubuntu 20.04.  Comparing the install that resulted from stock meson,
```
meson -Ddoc=disabled -Dgtk_doc=disabled -v --buildtype=debug build
```
with the one that resulted from this branch and forcing mostly rspfiles,
```
MESON_RSP_THRESHOLD=400 ~/meson/meson.py -Ddoc=disabled -Dgtk_doc=disabled --buildtype=debug build
```
yielded exactly one difference: .pc files had an extra space, e.g.

```
-Cflags:-I${includedir}
+Cflags: -I${includedir}
```

I'm not sure that's a problem worth fixing; opinions welcome.

test cases/common/145 special characters/btmp2/meson-private/sanitycheckc.c
```diff
@@ -0,0 +1 @@
+int main(void) { int class=0; return class; }
```
jon-turney, line 1: Should this file have got committed?

test cases/common/145 special characters/meson.build
```diff
@@ -42,29 +42,47 @@ gen2 = custom_target('gen2',
...
+# set to true if you want the builder to go green, false to see the bugs in action
+skipbugs = true
+cc = meson.get_compiler('c')
 
 foreach s : special
-  args = '-DCHAR="@0@"'.format(s[1])
-  e = executable('arg-string-' + s[0], 'arg-string-test.c', c_args: args)
-  test('arg-string-' + s[0], e, args: s[1])
+  if skipbugs and (s[0] == 'hash') and (cc.get_id() == 'msvc')
+    message('passing the hash char fails always at runtime with msvc (CHAR is "="?!, arg is ok)')
```
jon-turney, line 64: Fascinating.  I don't recall tripping over that before.

According to https://docs.microsoft.com/en-us/cpp/build/reference/d-preprocessor-definitions?view=vs-2019, `'#` is permitted instead of `=` in `/D`, which is probably related somehow.

docs/markdown/snippets/response-files.md
```diff
@@ -0,0 +1,7 @@
+## Response files enabled on Linux, reined in on Windows
+
+Meson used to always use response files on Windows,
+but never on Linux.
+
+It now strikes a happier balance, using them on both platforms,
+but only when needed to avoid commandline length limits.
```
jon-turney, line 7: `commandline` should be hyphenated or two words.

mesonbuild/backend/backends.py
```diff
@@ -607,29 +607,13 @@ def create_msvc_pch_implementation(self, target, lang, pch_header):
...
-                extra_args.append(arg)
-        else:
-            # MinGW GCC needs all backslashes in defines to be doubly-escaped
-            # FIXME: Not sure about Cygwin or Clang
-            for arg in args:
-                if arg.startswith('-D') or arg.startswith('/D'):
-                    arg = arg.replace('\\', '\\\\')
-                extra_args.append(arg)
+        for arg in args:
+            if arg.startswith('-D') or arg.startswith('/D'):
```
jon-turney, line 613: Patch commentary says we are doing this to all arguments, but I think maybe I was just confused.

mesonbuild/backend/ninjabackend.py
```diff
@@ -93,7 +93,7 @@ def gcc_rsp_quote(s):
 # variables (or variables we use them in) is interpreted directly by ninja
 # (e.g. the value of the depfile variable is a pathname that ninja will read
 # from, etc.), so it must not be shell quoted.
-raw_names = {'DEPFILE_UNQUOTED', 'DESC', 'pool', 'description'}
+raw_names = {'DEPFILE_UNQUOTED', 'DESC', 'pool', 'description', 'targetdep'}
```
jon-turney, line 114: The commentary doesn't explain why this variable fits into the category outlined in the comment above.  Does it? Perhaps the reasoning in the comment is incorrect? Or this change just makes things work? :)

jon-turney:
So, one thing I vaguely remember, which should probably be mentioned in comments somewhere is that ninja has some special handling for variables it "knows" contains pathnames (e.g. $out) and does some platform-specific quoting on them itself when the appear in commands.

mesonbuild/backend/ninjabackend.py
dankegel: You removed it by accident when you made raw_names a global, I think.
I just put it back because otherwise the rust tests all fail on rust's 'ninja -t clean' rules.
It was added by 49e9e017.

dankegel:
I don't have enough grasp of the situation at the moment to write the comment you're proposing.
But https://github.com/ninja-build/ninja/issues/53 is interesting reading, and hints at e.g. why ninja has to be the one to generate the .rsp files.

mesonbuild/backend/ninjabackend.py
```diff
@@ -219,8 +219,8 @@ def write(self, outfile):
         else:
             rspfile_quote_func = gcc_rsp_quote
 
-        for rsp in ([''] if self.refcount else [] +
-                    ['_RSP'] if self.rsprefcount else []):
+        for rsp in (([''] if self.refcount else []) +
+                    (['_RSP'] if self.rsprefcount else [])):
```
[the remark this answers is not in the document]
dankegel: That looks good.  Done.

test cases/common/145 special characters/meson.build
dankegel: Yep.  I removed the test for # and left a comment behind.

docs/markdown/snippets/response-files.md
dankegel: Done.

mesonbuild/backend/backends.py
dankegel: Fixed patch comment.

mesonbuild/backend/ninjabackend.py
dankegel: I disappeared the bug, and added a line in the patch commentary about what breaks if you remove that raw name.

test cases/common/145 special characters/btmp2/meson-private/sanitycheckc.c
dankegel: Whoops, no.  Fixed.

dankegel:
I think it's ready to be merged...

mesonbuild/backend/ninjabackend.py
```diff
@@ -877,13 +1064,15 @@ def generate_rules(self):
...
              'regenerate',
-             ninja_quote(quote_func(self.environment.get_source_dir())),
-             ninja_quote(quote_func(self.environment.get_build_dir()))]
+             self.environment.get_source_dir(),
+             self.environment.get_build_dir(),
+             '--backend',
+             'ninja']
         self.add_rule(NinjaRule('REGENERATE_BUILD',
-                                c + ['--backend', 'ninja'], [],
+                                c, [],
```
jon-turney, line 1075: Huh?

jon-turney:
@dankegel Thanks again for picking this up and getting it across the finish line. :)

* The release note snippet should probably mention `ninja`, since the behaviour for other backends is not changed (not sure what it currently is).
* Hijacking a cross-test for MESON_RSP_THRESHOLD=0 is perhaps not ideal, as fewer tests are run in that configuration.

scivision:
Thank you! Prior to this PR, this was a pet peeve for Meson on Windows, making it tedious to diagnose compiler commands emitted from Meson.

mesonbuild/backend/backends.py
mikkelfj: I'm not sure exactly how Meson output and Ninja processing results in a CreateProcess call on Windows but I suspect that the issue is with general quoting on not specifically the -D option although it is more likely to cause problems.

The Windows CreateProcess requires a single command line string of combined arguments. The rules are that backslash is not escaped unless followed by a double quote. Double quote always need escaping by prefixing a backslash. This means doubling the number of backslashes and adding one more if there is a sequence of backslashes followed by a double quote, or just inserting a backslash if there is a single double quote.

Also note that a pair of double quotes collapse into a single double quote, but with the above rules each of these would be prefixed by a backslash thus avoiding that collapse. The pair would be treated differently prior to 2008.

http://daviddeley.com/autohotkey/parameters/parameters.htm

mikkelfj: Ninja probably already do this for arguments it believes to be files.

jon-turney: > I suspect that the issue is with general quoting on not specifically the -D option although it is more likely to cause problems.

This suspicion is correct.

> The rules are that backslash is not escaped unless followed by a double quote. Double quote always need escaping by prefixing a backslash. This means doubling the number of backslashes and adding one more if there is a sequence of backslashes followed by a double quote, or just inserting a backslash if there is a single double quote.

This is now implemented in cmd_quote() in ninjabackend.py.  We didn't need it before this PR, because we always used response files.

I believe if you look at the git history of this function, you'll see that it's intended to make things like the following work as intended:

```
c_args: '-DPATH=C:\now\thats\a\path'
```

unfortunately, this means that the following doesn't work as intended:

```
c_args: '-DCHAR=\n'
```

I think additions to the meson language would be needed to allow both to be expressed unambigiously.

PR #7250, opened 2020-06-04
Add libeconf to users.md

sysrich:
libeconf is now using meson as its primary/default build system, so we'd like to see ourselves on the meson users list .


mensinda:
Please put `[skip ci]` into the commit for doc only changes next time.

PR #7254, opened 2020-06-05
cmake: Subprojects support CMAKE_PREFIX_PATH (fixes #7249)

### LINKED ISSUE #7249: -Dcmake_prefix_path does not work for CMake subprojects
This works for both CMake dependencies and CMake subprojects:
```
$ export CMAKE_PREFIX_PATH=<somedir>
$ meson <builddir>
```
And this works for CMake dependencies only:
```
meson -Dcmake_prefix_path=<somedir> <builddir>
```
I think it should be mentioned in the docs to avoid confusion.

mensinda:
fixes #7249


nirbheek:
Is this good to go?

mensinda:
Yes, it is (from my part).

PR #7256, opened 2020-06-05
Fix OpenMP on Visual Studio

peterh:
Using --backend=ninja with Visual Studio results in many warnings (one for each static library included in an executable). Using --backend=vs2019 results in OpenMP not working.

Fix both of these issues.

Example meson.build file demonstrating these issues:
```meson
project('tmp', 'c')

openmp = dependency('openmp')

libone = static_library('libone', 'fileone.c', dependencies: openmp)
libtwo = static_library('libtwo', 'filetwo.c', dependencies: openmp)

executable('main', 'main.c', link_with: [libone, libtwo])
```

### DIFF
--- a/mesonbuild/dependencies/misc.py
+++ b/mesonbuild/dependencies/misc.py
@@ -97,7 +97,8 @@ class OpenMPDependency(ExternalDependency):
             for name in header_names:
                 if self.clib_compiler.has_header(name, '', self.env, dependencies=[self], disable_cache=True)[0]:
                     self.is_found = True
-                    self.compile_args = self.link_args = self.clib_compiler.openmp_flags()
+                    self.compile_args = self.clib_compiler.openmp_flags()
+                    self.link_args = self.clib_compiler.openmp_link_flags()
                     break
             if not self.is_found:
                 mlog.log(mlog.yellow('WARNING:'), 'OpenMP found but omp.h missing.')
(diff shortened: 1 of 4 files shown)


mesonbuild/dependencies/misc.py
```diff
@@ -97,7 +97,9 @@ def __init__(self, environment, kwargs):
             for name in header_names:
                 if self.clib_compiler.has_header(name, '', self.env, dependencies=[self], disable_cache=True)[0]:
                     self.is_found = True
-                    self.compile_args = self.link_args = self.clib_compiler.openmp_flags()
+                    self.compile_args = self.clib_compiler.openmp_flags()
+                    if self.clib_compiler.get_id() != 'msvc':
+                        self.link_args = self.clib_compiler.openmp_flags()
```
dcbaker, line 102: I don't think this is the right place to solve this, I think this should be solved the same way that the Thread dependency does, with two separate methods in the compiler classes an `openmp_flags` and `openmp_link_flags`. Just leave the non MSVC compilers returning the values of `openmp_flags()`, and change MSVC to return `[]`

peterh: That was my first instinct. Then I saw `if self.clib_compiler.get_id() == 'pgi'` a few lines above in the same function and took the easy way out.

I'm glad to know that my instincts were correct, at least. Force pushed with the requested change.

PR #7263, opened 2020-06-07
Fix new compiler warnings in test code

jpakkane:


jpakkane:
Test failure is unrelated.

Issue #7265, opened 2020-06-08
0.54.1 regression for Fortran generated library()


scivision:
**Describe the bug**
Commit bd8d079473d introduced the Meson 0.54.1 regression for library() created from Fortran generated source files.
This regression happens for "static" or "shared" library().

---

With Windows 10 and Gfortran 10.1.0:

The "good" Meson 0.54.0 command line is:

```sh
gfortran "-Imod3@sha" "-I." "-I.." "-fdiagnostics-color=always" "-pipe" "-Wall" "-g" "-Jmod3@sha" -o mod3@sha/mod3.f90.obj -c "C:/Users/code/temp/7 generated/build/mod3.f90"
```

The "bad" Meson 0.54.1 command line is:

```sh
gfortran "-Imod3@sha" "-I." "-I.." "-fdiagnostics-color=always" "-pipe" "-Wall" "-g" "-Jmod3@sha" -o "C:_Users_code_temp_7 generated_build_mod3.f90.obj" -c "C:/Users/code/temp/7 generated/build/mod3.f90"
```

--- 

On Ubuntu 20.04 with Gfortran 9.3.0 or Gfortran 10.0.1 the regression is **not** apparent.

Meson 0.54.0 emits:

```sh
gfortran -Imod3@sha -I. -I.. -fdiagnostics-color=always -pipe -Wall -g -D_FILE_OFFSET_BITS=64 -fPIC -Jmod3@sha -o 'mod3@sha/mod3.f90.o' -c '/mnt/c/Users/code/temp/7 generated/build/mod3.f90'
```

Meson 0.54.1 emits:
```sh
 gfortran -Imod3@sha -I. -I.. -fdiagnostics-color=always -pipe -Wall -g -D_FILE_OFFSET_BITS=64 -fPIC -Jmod3@sha -o 'mod3@sha/_mnt_c_Users_code_temp_7 generated_build_mod3.f90.o' -c '/mnt/c/Users/code/temp/7 generated/build/mod3.f90'
```

**To Reproduce**
I updated the Meson project test "test cases/fortran/7 generated" in PR #7266 as this is an important case to test.

**Expected behavior**

**system parameters**

The broken setup is:

* Gfortran 10.1.0 (MinGW via MSYS2)
* native build
* Windows 10
* Python: 3.7.7
* Meson: 0.54.1 introduced the regression
* Ninja: 1.10.0

A **working** setup is:

* Gfortran 9.3.0 or Gfortran 10.0.1
[...]

PR #7266, opened 2020-06-08
0.54.1 regression: add library() check for Fortran configure_file()

scivision:
Meson 0.54.1 has a regression noted in #7265. This additional test code reveals the issue, but does not fix the issue. This is an important use case that may only break on Windows, but it's good to add this test in general.

### DIFF
--- a/test cases/fortran/7 generated/meson.build
+++ b/test cases/fortran/7 generated/meson.build
@@ -10,6 +10,8 @@ conf_data.set('THREE', 3)
 
 outfile = configure_file(
       input : 'mod3.fpp', output : 'mod3.f90', configuration : conf_data)
+# Manually build absolute path to source file to test
+# https://github.com/mesonbuild/meson/issues/7265
 three = library('mod3', meson.current_build_dir() / 'mod3.f90')
 
 templates_basenames = ['mod2', 'mod1']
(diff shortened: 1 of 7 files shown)


test cases/fortran/7 generated/meson.build
```diff
@@ -6,6 +6,11 @@ project('generated', 'fortran')
 conf_data = configuration_data()
 conf_data.set('ONE', 1)
 conf_data.set('TWO', 2)
+conf_data.set('THREE', 3)
+
+outfile = configure_file(
+      input : 'mod3.fpp', output : 'mod3.f90', configuration : conf_data)
+three = library('mod3', meson.current_build_dir() / 'mod3.f90')
```
nirbheek, line 13: Why aren't you using `outfile` here directly instead of creating a string to the output?

nirbheek:
@scivision poke :)

test cases/fortran/7 generated/meson.build
nirbheek: I see, you want to test that an absolute path to a source file works correctly.

nirbheek:
> The Cygwin failure should go away on a rebase but this one seems relevant:

Yes, that's what the test is checking for. The commit I just pushed should fix it.

scivision:
Sorry for my delay. Yes I should have more simply done

```meson
three = library('mod3', sources: outfile)
```

it was just a simple oversight.

nirbheek:
> it was just a simple oversight.

I think that would've used a relative path (and if it doesn't, it might've changed to use a relative path in the future), so creating the absolute path by hand is correct. I added a comment saying that when I added my commit :)

PR #7272, opened 2020-06-08
compilers: Add UNIX large file support for MinGW

seungha-yang:
MinGW gcc supports _FILE_OFFSET_BITS=64 and we need to set it for
MinGW as well

cc: @nirbheek

### DIFF
--- a/mesonbuild/compilers/compilers.py
+++ b/mesonbuild/compilers/compilers.py
@@ -1246,10 +1246,10 @@ def get_largefile_args(compiler):
     '''
     Enable transparent large-file-support for 32-bit UNIX systems
     '''
-    if not (compiler.info.is_windows() or compiler.info.is_darwin()):
+    if not (compiler.get_argument_syntax() == 'msvc' or compiler.info.is_darwin()):
         # Enable large-file support unconditionally on all platforms other
-        # than macOS and Windows. macOS is now 64-bit-only so it doesn't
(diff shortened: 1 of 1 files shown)


mesonbuild/compilers/compilers.py
```diff
@@ -1246,10 +1246,10 @@ def get_largefile_args(compiler):
     '''
     Enable transparent large-file-support for 32-bit UNIX systems
     '''
-    if not (compiler.info.is_windows() or compiler.info.is_darwin()):
+    if not (compiler.get_argument_syntax() == 'msvc' or compiler.info.is_darwin()):
```
nirbheek, line 1249: This will also match `clang-cl`, does that not support `_FILE_OFFSET_BITS=64`?

tp-m: I think `clang-cl` relies on the Windows SDK and doesn't provide its own like MinGW does, so I would expect `clang-cl` to behave the same as msvc's `cl` here?

So looks correct to me as-is.

seungha-yang: I guess clang-cl doesn't support `_FILE_OFFSET_BITS` (I couldn't find any reference)

fwiw, glib's meson file was referenced here
https://gitlab.gnome.org/GNOME/glib/blob/master/meson.build#L141-147

PR #7274, opened 2020-06-08
Add gtkdoc test covering gtkdoc-scangobj

jon-turney:
Add a trivial GObject-derived class to the gtkdoc test.

This test adds coverage of `gtkdoc-scangobj` being invoked by gtkdochelper (because a typesfile exists, generated by `gtkdoc-scan --rebuild-types`)

### DIFF
--- a/test cases/frameworks/10 gtk-doc/doc/baz/baz-docs.xml
+++ b/test cases/frameworks/10 gtk-doc/doc/baz/baz-docs.xml
@@ -1,0 +1,34 @@
+<?xml version="1.0"?>
+<!DOCTYPE book PUBLIC "-//OASIS//DTD DocBook XML V4.1.2//EN"
+               "http://www.oasis-open.org/docbook/xml/4.1.2/docbookx.dtd">
+<book id="index" xmlns:xi="http://www.w3.org/2003/XInclude">
+  <bookinfo>
+    <title>Baz Reference Manual</title>
+    <releaseinfo>for Baz 1.0</releaseinfo>
+    <authorgroup>
+      <author>
+        <firstname>Testy</firstname>
+        <surname>McTestFace</surname>
+        <affiliation>
+          <address>
+            <email>unknown@example.com</email>
+          </address>
+        </affiliation>
+      </author>
+    </authorgroup>
+    <copyright>
+      <year>2020</year>
+      <holder>The Baz Organization</holder>
+    </copyright>
+  </bookinfo>
+
+  <reference id="baz">
+    <title>Baz library</title>
+    <partintro>
+      <para>
+        This documents the Baz library.
+      </para>
+    </partintro>
+    <xi:include href="xml/baz_trivial.xml"/>
+  </reference>
+</book>
--- a/test cases/frameworks/10 gtk-doc/doc/baz/meson.build
+++ b/test cases/frameworks/10 gtk-doc/doc/baz/meson.build
@@ -1,0 +1,7 @@
+gnome.gtkdoc('baz',
+  src_dir : [],
+  main_xml : 'baz-docs.xml',
+  dependencies: libbaz_dep,
+  scan_args: '--rebuild-types',
+  check: true,
+  install : true)
--- a/test cases/frameworks/10 gtk-doc/doc/meson.build
+++ b/test cases/frameworks/10 gtk-doc/doc/meson.build
@@ -8,3 +8,5 @@ subdir('foobar1')
 subdir('foobar2')
 subdir('foobar3')
 subdir('foobar4')
+
+subdir('baz')
--- a/test cases/frameworks/10 gtk-doc/meson.build
+++ b/test cases/frameworks/10 gtk-doc/meson.build
@@ -12,6 +12,7 @@ assert(gnome.gtkdoc_html_dir('foobar') == 'share/gtk-doc/html/foobar', 'Gtkdoc i
 inc = include_directories('include')
 
 subdir('include')
+subdir('src')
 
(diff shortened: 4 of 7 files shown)


jon-turney:
The failure here is interesting:

```
[4/4] /usr/bin/python3 /root/meson.py --internal gtkdoc 
'--sourcedir=/root/test cases/frameworks/10 gtk-doc' '--builddir=/root/b 
e24876abc6' --subdir=doc/baz --headerdirs= --mainfile=baz-docs.xml 
--modulename=baz --moduleversion= --mode=auto 
--gtkdoc-scan=/usr/bin/gtkdoc-scan 
--gtkdoc-scangobj=/usr/bin/gtkdoc-scangobj 
--gtkdoc-mkdb=/usr/bin/gtkdoc-mkdb --gtkdoc-mkhtml=/usr/bin/gtkdoc-mkhtml 
--gtkdoc-fixxref=/usr/bin/gtkdoc-fixxref --scanargs=--rebuild-types 
--content-files= --cc=gcc --ld=gcc '--cflags=-I./src/. -I../test 
cases/frameworks/10 gtk-doc/src/. -I/usr/include/glib-2.0 
-I/usr/lib/x86_64-linux-gnu/glib-2.0/include' '--ldflags='"'"'-L/root/b 
e24876abc6/src'"'"' '"'"'-Wl,-rpath,/root/b e24876abc6/src'"'"' -lbaz 
-lgobject-2.0 -lglib-2.0'
FAILED: doc/baz/baz-decl.txt 
/usr/bin/python3 /root/meson.py --internal gtkdoc '--sourcedir=/root/test 
cases/frameworks/10 gtk-doc' '--builddir=/root/b e24876abc6' 
--subdir=doc/baz --headerdirs= --mainfile=baz-docs.xml --modulename=baz 
--moduleversion= --mode=auto --gtkdoc-scan=/usr/bin/gtkdoc-scan 
--gtkdoc-scangobj=/usr/bin/gtkdoc-scangobj 
--gtkdoc-mkdb=/usr/bin/gtkdoc-mkdb --gtkdoc-mkhtml=/usr/bin/gtkdoc-mkhtml 
--gtkdoc-fixxref=/usr/bin/gtkdoc-fixxref --scanargs=--rebuild-types 
--content-files= --cc=gcc --ld=gcc '--cflags=-I./src/. -I../test 
cases/frameworks/10 gtk-doc/src/. -I/usr/include/glib-2.0 
-I/usr/lib/x86_64-linux-gnu/glib-2.0/include' '--ldflags='"'"'-L/root/b 
e24876abc6/src'"'"' '"'"'-Wl,-rpath,/root/b e24876abc6/src'"'"' -lbaz 
-lgobject-2.0 -lglib-2.0'
Building documentation for baz
ERROR: Error in gtkdoc helper script:
ERROR: ['/usr/bin/gtkdoc-scangobj', '--types=/root/b 
e24876abc6/doc/baz/baz.types', '--module=baz', '--run=', '--cflags=-I./src/. 
-I../test cases/frameworks/10 gtk-doc/src/. -I/usr/include/glib-2.0 
-I/usr/lib/x86_64-linux-gnu/glib-2.0/include', "--ldflags='-L/root/b 
e24876abc6/src' '-Wl,-rpath,/root/b e24876abc6/src' -lbaz -lgobject-2.0 
-lglib-2.0", '--cc=gcc', '--ld=gcc', '--output-dir=/root/b 
e24876abc6/doc/baz'] failed with status 1
gcc: error: cases/frameworks/10: No such file or directory
gcc: error: gtk-doc/src/.: No such file or directory
2020-06-08 23:17:06,533:scangobj.py:execute_command:1196:WARNING:Compiling 
scanner failed: 1, command: gcc -I./src/. -I../test cases/frameworks/10 
gtk-doc/src/. -I/usr/include/glib-2.0 
-I/usr/lib/x86_64-linux-gnu/glib-2.0/include -c -o baz-scan.o baz-scan.c
```

The script has an argument of the form `--cflags=-I@BUILD_ROOT@ -I@SOURCE_ROOT@`, which (in this case), after substitution, contains spaces, which aren't escaped or quoted properly (compare with the `--ldflags` argument).

jon-turney:
> The failure here is interesting:
>
> [...]
>
> The script has an argument of the form `--cflags=-I@BUILD_ROOT@ -I@SOURCE_ROOT@`, which (in this case), after substitution, contains spaces, which aren't escaped or quoted properly (compare with the `--ldflags` argument).

Both `--cflags` and `--ldflags` arguments are being made in the gnome module using `mesonlib.join_args` (which uses `shlex.quote` or it's platform equivalent), but the substitutions aren't yet made, so any spaces in them don't get quoted.

I guess this quoting needs to be deferred to the backend, but the way that this is just building the `command:` for a `CustomTarget` under the hood, this seems not a straightforward change.

nirbheek:
Would it help if gtk-doc could take arguments using rspfiles?

jon-turney:
> Would it help if gtk-doc could take arguments using rspfiles?

I don't think so.  The problem is that the quoting happens too early (when building the custom target), before substitution of meson path placeholders (in the backend), because we are building these odd 'argument for a subprocess which contains an argument list for a subsubprocess' strings.

If we're talking about changing gtk-doc, then what might work is allowing `--cflags=-I@BUILD_ROOT@ --cflags=-I@SOURCE_ROOT@` (if it isn't already) (which are then concatenated inside gtk-doc for the gcc it runs), because I think then we wouldn't need `join_args`.

(There's probably also a generic problem here with `custom_target()` that certain complex command lines like this can't be represented to it in a way which let's them get quoted correctly on all platforms.)

Issue #7276, opened 2020-06-08
Meson will find build machine binaries for host if the build machine and host machine are the same


dcbaker:
If you have a case where the build machine == host machine (say x86_64-linux and x86_64-linux) then meson will try to use binaries from the build machine for the host unconditionally. There are however cases where this isn't safe, such as with cmake. When building chromeos this exact situation happens, and the default /usr/bin/cmake is selected, which in turn finds dependencies on the build machine, and tries to use them in the host machine, which obviously will compile, but results in an unusable system.

Gcenx:
Is it possible to disable this change in some manner without needing to patch meson?

Say I want to use a cross file that builds for darwin but I’m running darwin and rather use the native pkg-config/cmake etc instead of specifying these binaries within the cross file?

dcbaker:
No, there isn't a way to do that currently. What exactly are you doing, cross compiling from macOS to iOS or similar?

Gcenx:
Some meson based projects using macports had issues compiling, for some reason using a cross file got the projects building correctly for the desired architecture namely i386.

I’d assume adding all required utility’s into the cross files is desired, but would patching allow_default_for_cross=False to True break cross-compilation that’s not darwin?

PR #7277, opened 2020-06-08
Don't try to use build machine dependency finders for host

dcbaker:
Even if the build machine == host machine, we can't use dependencies from the build machine for the host machine. This can happen when using a system to build a new machine image. for example when compiling chromeos. It's completely possible to build an x86_64-linux host for an x86_64-linux build machine, but not have all of the same runtime dependencies. In particular, if meson tries to use cmake from the build machine for host dependencies it will result in trouble.

Fixes: #7276

### DIFF
--- a/mesonbuild/cmake/executor.py
+++ b/mesonbuild/cmake/executor.py
@@ -61,25 +61,7 @@ class CMakeExecutor:
             return
 
     def find_cmake_binary(self, environment: Environment, silent: bool = False) -> T.Tuple['ExternalProgram', str]:
-        from ..dependencies.base import ExternalProgram
-
-        # Create an iterator of options
-        def search():
-            # Lookup in cross or machine file.
-            potential_cmakepath = environment.lookup_binary_entry(self.for_machine, 'cmake')
-            if potential_cmakepath is not None:
-                mlog.debug('CMake binary for %s specified from cross file, native file, or env var as %s.', self.for_machine, potential_cmakepath)
-                yield ExternalProgram.from_entry('cmake', potential_cmakepath)
-                # We never fallback if the user-specified option is no good, so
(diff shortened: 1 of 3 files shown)


dcbaker:
This is conceptually similar to #7099

mensinda:
I am not 100% sure if this is the right approach for CMake since this would only change a few default search paths at best.

Using CMake from the build machine should always be OK (and probably the best solution) once we support propper CMake cross-compilation with [toolchain files](https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html).

dcbaker:
Changing the default search paths is 100% the problem. /usr/bin/cmake will find cmake files in /usr, which are completely wrong almost all of the time. It's also super inconsistent with the way we find other binaries, which if not specified in the cross file are automatically treated as "not found"

mensinda:
In that case, we can also set [CMAKE_FIND_ROOT_PATH](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_ROOT_PATH.html), [CMAKE_SYSROOT](https://cmake.org/cmake/help/latest/variable/CMAKE_SYSROOT.html), etc. I would argue that CMake *is* a special binary since it has it's own cross-compilation logic.

Wouldn't this also mean that we would need to invoke CMake with an exec wrapper (so `wine /path/to/cmake.exe` for mingw-w64)?

I am fine with changing the default variables by using the host CMake binary, but I would like to keep the option open for using the build CMake binary with a toolchain file in the future.

dcbaker:
What this does is gives binaries a way to opt out of the default for cross builds through a boolean, so in the future you could change that boolean to something like `allow_default=bool(has_toolchain_file)`, which seems like a reasonable thing to do.

PR #7279, opened 2020-06-09
tell variable name when erroring on bad combo option

### LINKED ISSUE #7269: Report which command line variable has wrong value
Sometimes projects change the type of a meson option, for example from boolean to feature.
Meson then gives an error like this:
```
meson.build:1:0: ERROR: Value "false" for combo option is not one of the choices. Possible choices are: "enabled", "disabled", "auto".
```
Which means I have to spend some time figuring out which option changed.
Would it be possible to mention which option has a wrong value? For example by changing the first sentence of the error to: _Value "false" for combo option **some_option** is not one of the choices._

scivision:
fixes #7269

test by:

meson.build
```meson
project('foo')
message(get_option('m1'))
```

meson_options.txt
```meson
option('m1', type: 'combo', choices: ['a', 'b'])
```

then do

```sh
meson build -Dm1=c
```

now results in 

> meson.build:1:0: ERROR: Value "c" for combo option "m1" is not one of the choices. Possible choices are: "a", "b".

### DIFF
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -161,7 +161,9 @@ class UserComboOption(UserOption[str]):
     def validate_value(self, value):
         if value not in self.choices:
             optionsstring = ', '.join(['"%s"' % (item,) for item in self.choices])
-            raise MesonException('Value "%s" for combo option is not one of the choices. Possible choices are: %s.' % (value, optionsstring))
+            raise MesonException('Value "{}" for combo option "{}" is not one of the choices.'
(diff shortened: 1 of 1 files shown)


jon-turney:
Nice.

> test by:

I think you could turn this into a test case in `test cases\failing` by the setting the invalid option and the noting expected line in stdout in test.json.

scivision:
Yes I also think the other options could be fixed as well I was just in a big hurry and doing this when I had a spare moment. And then we could add test cases for each of the variable types.

PR #7283, opened 2020-06-09
compilers: corretify deduplication direction

marcelhollerbach:
so: when building compile args, meson is deduplicating flags. When a
compiler argument is appended, a later appearance of a dedup'ed is going
to remove a earlier one. If the argument is prepended, the element
*before* the new one is going to be removed. And that is where the
problem reported in https://github.com/mesonbuild/meson/pull/7119 is
coming in. In the revision linked there, the order of replacement in the
prepend case was revesered.

With this patch, we restore this behaviour again.


marcelhollerbach:
I'll add a testcase tomorrow. Just wanted to add the PR here, in case more projects than just zrythem have issues.

marcelhollerbach:
Now with a testcase ẃith the failed case.

I splitted the testcase off, to ensure that we do not accidently flush the CompilerArgs too early, so we ensure that we have duplicates when flushing the first time in self.pre.

jon-turney:
I think a project test for the scenario which exposed this failure (assuming https://github.com/mesonbuild/meson/pull/7119#issuecomment-641416810 is right: having a subproject which exposes in it's include path an include file whose name clashes with one in the superproject) might also be a good idea, since it seems like the kind of thing which might get broken (kind of surprised we don't have one already, really)

PR #7292, opened 2020-06-10
mintro: enforced order for subcommands

TheQwertiest:
This is needed for deterministic output for #7217

### DIFF
--- a/mesonbuild/mintro.py
+++ b/mesonbuild/mintro.py
@@ -53,7 +53,7 @@ class IntroCommand:
 def get_meson_introspection_types(coredata: T.Optional[cdata.CoreData] = None,
                                   builddata: T.Optional[build.Build] = None,
                                   backend: T.Optional[backends.Backend] = None,
-                                  sourcedir: T.Optional[str] = None) -> 'collections.OrderedDict[str, IntroCommand]':
+                                  sourcedir: T.Optional[str] = None) -> 'T.Mapping[str, IntroCommand]':
     if backend and builddata:
         benchmarkdata = backend.create_test_serialisation(builddata.get_benchmarks())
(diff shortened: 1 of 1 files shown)


mesonbuild/mintro.py
```diff
@@ -53,7 +53,7 @@ def __init__(self,
 def get_meson_introspection_types(coredata: T.Optional[cdata.CoreData] = None,
                                   builddata: T.Optional[build.Build] = None,
                                   backend: T.Optional[backends.Backend] = None,
-                                  sourcedir: T.Optional[str] = None) -> 'T.OrderedDict[str, IntroCommand]':
+                                  sourcedir: T.Optional[str] = None) -> 'collections.OrderedDict[str, IntroCommand]':
```
dcbaker, line 56:
```suggestion
                                  sourcedir: T.Optional[str] = None) -> 'T.Mapping[str, IntroCommand]':
```
Might be what you want.

dcbaker: It's not as exact, but it is correct and exists in all versions of python we try to support.

TheQwertiest: But would that be any different from using T.Dict (as it was in the beginning)? As in using a different type, which has the same interface as our target type.

dcbaker: Because Dict is a type of Mapping, and OrderedDict is a type of Mapping, but OrderedDict is not a type of Dict (or vice versa)

TheQwertiest: Fixed

dcbaker:
dang it, I meant to squash that into one commit :(

TheQwertiest:
Now meson will have my (allegedly) bad git practices etched in its history for ever >_<

PR #7293, opened 2020-06-10
Add [options] section to native/cross files

xclaesse:
This is a counter proposal for https://github.com/mesonbuild/meson/pull/6597. Since I had this code laying around, let's push it as PR for reference. The main advantage is the implementation is simpler. This also does not split project and built-in options in the machine file which is IMHO an arbitrary useless distinction.

Putting it here for reference, I'm not really expecting it to be merged but maybe it can help tackling the remaining issue for per machine compiler options.

### DIFF
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -230,7 +230,7 @@ class UserFeatureOption(UserComboOption):
 
 def load_configs(filenames: T.List[str]) -> configparser.ConfigParser:
     """Load configuration files from a named subdirectory."""
-    config = configparser.ConfigParser(interpolation=None)
+    config = CmdLineFileParser()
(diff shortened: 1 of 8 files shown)


Ericson2314:
I like separating the different types of options. If we could at "parse time" identify everything by whitelists and separate them, I think it would make the current code / featureset a lot simpler. Having separate sections does that splitting work for us.

Also, I think as far as deprecation cycles go, it's easier to from that to this than vice-versa: it's comparatively easier to allow everything in each section and deprecate the other types of sections. But if we switch from this to that, all the complexity of deciding what option goes aware can't go away until the deprecation period is finished.

Basically, I think this is only easier because of the messiness around options that already exists, messiness that it would be nice to move away from.

PR #7297, opened 2020-06-11
docs: remove duplicated entry for gnome.generate_gir() documentation

takaswie:
The documentation of gnome.generate_gir() has duplicated entry for
dependencies parameter. As a fix, this patch removes the entry added
recently.


jpakkane:
Thanks.

Issue #7301, opened 2020-06-11
[REGRESSION] UnboundLocalError in cmake configure_package_config_file()


p01arst0rm:
**Describe the bug**
when specifying `install_dir : join_paths(libdir, 'cmake')`  meson
panics in function `configure_package_config_file()`  ;- line 
257 of `mesonbuild\modules\cmake.py`

**To Reproduce**
```
cmake_conf = configuration_data()
cmake_conf.set_quoted('foo', 'bar')

cmake.configure_package_config_file(
    name : 'foolib',
    input : 'foolib.cmake.in',
    install_dir : join_paths(libdir, 'cmake'),
    configuration : cmake_conf)
```

**Expected behavior**
generate a configured cmake package config file

**system parameters**
- native build
- Windows 10 (Current latest patch KB4538461)
- python version 3.7.1
- meson version 0.54.1
- ninja Not Applicable.

p01arst0rm:
cc @dcbaker :)

PR #7302, opened 2020-06-11
modules/cmake: Fix setting install_dir

dcbaker:
ensure that `install_dir` is always assigned.

Fixes: #7301


dcbaker:
@mensinda, are those cmake failures on windows expected?

mensinda:
Yes, since they are the result of an Azure update (stil a meson bug, but none that was found before).

PR #7303, opened 2020-06-11
envconfig: Add [constants] section in machine files

xclaesse:
Machine files already supports `+` operator as an implementation detail,
since it's using eval(). Now make it an officially supported feature and
add a way to define constants that are used while evaluating an entry
value.

### DIFF
--- a/docs/markdown/Machine-files.md
+++ b/docs/markdown/Machine-files.md
@@ -8,9 +8,82 @@ environments](Native-environments.md).
 ## Sections
 
 The following sections are allowed:
+- constants
 - binaries
 - paths
 - properties
+
+### constants
+
+*Since 0.55.0*
+
+String and list concatenation is supported using the `+` operator, joining paths
+is supported using the `/` operator.
+Entries defined in the `[constants]` section can be used in any other section
+(they are always parsed first), entries in any other section can be used only
+within that same section and only after it has been defined.
+
+```ini
+[constants]
+toolchain = '/toolchain'
+common_flags = ['--sysroot=' + toolchain / 'sysroot']
+
+[properties]
+c_args = common_flags + ['-DSOMETHING']
+cpp_args = c_args + ['-DSOMETHING_ELSE']
+
+[binaries]
+c = toolchain / 'gcc'
+```
+
+This can be useful with cross file composition as well. A generic cross file
+could be composed with a platform specific file where constants are defined:
+```ini
+# aarch64.ini
+[constants]
+arch = 'aarch64-linux-gnu'
+```
+
+```ini
+# cross.ini
+[binaries]
+c = arch + '-gcc'
+cpp = arch + '-g++'
+strip = arch + '-strip'
+pkgconfig = arch + '-pkg-config'
+...
+```
+
+This can be used as `meson setup --cross-file aarch64.ini --cross-file cross.ini builddir`.
+
+Note that file composition happens before the parsing of values. The example
+below results in `b` being `'HelloWorld'`:
+```ini
+# file1.ini:
+[constants]
+a = 'Foo'
+b = a + 'World'
+```
+
+```ini
+#file2.ini:
+[constants]
+a = 'Hello'
+```
+
+The example below results in an error when file1.ini is included before file2.ini
+because `b` would be defined before `a`:
+```ini
+# file1.ini:
+[constants]
+b = a + 'World'
+```
+
+```ini
+#file2.ini:
+[constants]
+a = 'Hello'
+```
 
 ### Binaries
 
--- a/docs/markdown/snippets/machine_file_constants.md
+++ b/docs/markdown/snippets/machine_file_constants.md
@@ -1,0 +1,20 @@
+## Machine file constants
+
+Native and cross files now support string and list concatenation using the `+`
+operator, and joining paths using the `/` operator.
+Entries defined in the `[constants]` section can be used in any other section.
+An entry defined in any other section can be used only within that same section and only
+after it has been defined.
+
+```ini
+[constants]
+toolchain = '/toolchain'
+common_flags = ['--sysroot=' + toolchain + '/sysroot']
+
+[properties]
+c_args = common_flags + ['-DSOMETHING']
+cpp_args = c_args + ['-DSOMETHING_ELSE']
+
+[binaries]
+c = toolchain + '/gcc'
+```
--- a/mesonbuild/envconfig.py
+++ b/mesonbuild/envconfig.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import configparser, os, subprocess
+import os, subprocess
 import typing as T
 
 from . import mesonlib
@@ -82,33 +82,6 @@ CPU_FAMILES_64_BIT = [
     'wasm64',
     'x86_64',
 ]
-
-class MesonConfigFile:
-    @classmethod
-    def from_config_parser(cls, parser: configparser.ConfigParser) -> T.Dict[str, T.Dict[str, T.Dict[str, str]]]:
-        out = {}
-        # This is a bit hackish at the moment.
-        for s in parser.sections():
-            section = {}
-            for entry in parser[s]:
-                value = parser[s][entry]
-                # Windows paths...
-                value = value.replace('\\', '\\\\')
-                if ' ' in entry or '\t' in entry or "'" in entry or '"' in entry:
-                    raise EnvironmentException('Malformed variable name {} in cross file..'.format(entry))
-                try:
-                    res = eval(value, {'__builtins__': None}, {'true': True, 'false': False})
-                except Exception:
-                    raise EnvironmentException('Malformed value in cross file variable {}.'.format(entry))
-
-                for i in (res if isinstance(res, list) else [res]):
-                    if not isinstance(i, (str, int, bool)):
-                        raise EnvironmentException('Malformed value in cross file variable {}.'.format(entry))
-
-                section[entry] = res
-
-            out[s] = section
-        return out
 
 def get_env_var_pair(for_machine: MachineChoice,
                      is_cross: bool,
--- a/run_unittests.py
+++ b/run_unittests.py
@@ -3915,7 +3915,7 @@ recommended as it is not supported on some platforms''')
         with tempfile.NamedTemporaryFile(mode='w', delete=False) as crossfile:
             crossfile.write(textwrap.dedent(
                 '''[binaries]
-                pkgconfig = r'{0}'
+                pkgconfig = '{0}'
 
                 [properties]
 
@@ -3945,7 +3945,7 @@ recommended as it is not supported on some platforms''')
                 pkgconfig = 'pkg-config'
 
                 [properties]
-                pkg_config_libdir = [r'{0}']
+                pkg_config_libdir = ['{0}']
 
                 [host_machine]
                 system = 'linux'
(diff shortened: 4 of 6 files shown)


mesonbuild/envconfig.py
```diff
@@ -86,8 +86,10 @@ class MesonConfigFile:
     @classmethod
     def from_config_parser(cls, parser: configparser.ConfigParser) -> T.Dict[str, T.Dict[str, T.Dict[str, str]]]:
         out = {}
+        constants = {'true': True, 'false': False}
```
dcbaker, line 89: Don't we need to parse the constants section first to avoid a race between reading constants and evaluating sections? I think we need a test where `[constants]` isn't the first section.

Actually, taking that further, don't we need to read constants from each machine file, in order, before parsing any other sections in any other cross file to make sure that layering/overriding works correctly? We probably need a test for that too.

docs/markdown/Machine-files.md
```diff
@@ -8,10 +8,34 @@ environments](Native-environments.md).
...
+common_flags = ['--sysroot=' + toolchain + '/sysroot']
+
+[properties]
+c_args = common_flags + ['-DSOMETHING']
+cpp_args = c_args + ['-DSOMETHING_ELSE']
+
+[binaries]
+c = toolchain + '/gcc'
+```
+
```
dcbaker, line 38: Another example where I can think to use this (with layering is something like):
```ini
[constants]
arch = 'aarch64-linux-gnu'
```

```ini
[binaries]
c = arch + '-gcc'
cpp = arch + '-g++'
strip = arch + '-strip'
pkgconfig = arch + '-pkg-config'
# etc
```

You don't necessarily need to document this, but I'm certain other people will think of the same thing.

xclaesse: Yes that's the kind of usage I had in mind too, probably worth mentioning it explicitly in the doc, I'll add.

The other usage I still have in mind is adding `--cross-file-constant toolchain=/path/to/my/toolchain`. The idea is a project could ship their cross file and user would only have to specify from command line where they installed the SDK toolchain without having to write extra file. That's in a separate PR that will need to be adapted to now syntax proposed here: https://github.com/mesonbuild/meson/pull/5859.

mesonbuild/envconfig.py
dcbaker: On point 2 I think the configparser will actually handle that, but a test wouldn't be the worst thing to have :)

xclaesse: In the doc I wrote:
```
Entries defined in the `[constants]` section can be used in any other section
after its definition (i.e. should usually be first section)
```
The idea is you could write:
```ini
[host_machine]
# constants can't be used here
system = 'linux'
...
[constants]
path='/foo'
...
[binaries]
# constants can be used here
c = path + '/gcc'
...
```
But I agree that wouldn't make much sense. I'm also not sure the ordering of sections is guaranteed in case for example the first file does not have [constants] section and 2nd file has [constants] is first section... I think the easiest solution here is to guarantee we parse [constants] first whatever is the ordering in the file itself.

xclaesse: Changed the code to parse constants first.

docs/markdown/Machine-files.md
xclaesse: Mentioned composition in the doc and modified the unit test to verify it works.

mesonbuild/envconfig.py
dcbaker: The order is insertion order guaranteed using and `OrderedDict`. and I wrote a little test for myself, it will store the sections in order, by file. so if file1 has a `[binaries]` section, and file2 has a `[constants]` section and a `[binaries]` section, `[binaries]` will be iterated before `[constants]` I think it's going to be an endless source of headaches to not guarantee that `[constants]` will be read first.

xclaesse:
Note: as a follow-up PR it could be interesting to also add support for `/` operator because I'm expecting this feature to be used a lot for building paths. `c = toolchain / 'gcc'` is nicer than `c = toolchain + '/gcc'`.

mesonbuild/envconfig.py
xclaesse: Ok, thanks for confirming what I though. So my updated patch is the only reliable way I think. Right?

xclaesse:
@dcbaker I believe the failing CI is unrelated.

@jpakkane any objection on the feature?

dcbaker:
There's a couple of cmake failures on windows that mensinda is working on, so if it's those then yeah, unrelated.

run_unittests.py
```diff
@@ -4702,6 +4702,37 @@ def test_link_language_linker(self):
         self.assertRegex(contents, r'build main(\.exe)?.*: c_LINKER')
         self.assertRegex(contents, r'build (lib|cyg)?mylib.*: c_LINKER')
 
+    def test_cross_file_constants(self):
+        filenames = []
+        with tempfile.NamedTemporaryFile(mode='w', delete=False) as crossfile:
```
dcbaker, line 4707: There are a bunch of caveats to using `NamedTemporaryFile` like this on windows. There's  a `temp_filename` helper in run_unittests.py that's safe to use on windows, although it does always delete the file, you could probably do something like:

```python
with temp_filename() as cfile1, temp_filename() as cfile2:
    with open(cfile1, 'w') as f:
        ...
    with open(cfile2, 'w') as f:
        ....
    parser = ...
    values = ...
self.assert ...
```

docs/markdown/Machine-files.md
```diff
@@ -8,10 +8,53 @@ environments](Native-environments.md).
...
 - properties
 
+### constants
+
+*Since 0.55.0*
+
+String and list concatenation is supported using the `+` operator.
+Entries defined in the `[constants]` section can be used in any other section,
+entries in any other section can be used only within that same section and only
+after it has been defined.
```
dcbaker, line 23: Probably should document that `[constants]` is guaranteed to be read before any other sections.

run_unittests.py
xclaesse: There are many tests that use NamedTemporaryFile. The problem on Windows is you cannot read the file while it's open, so you cannot use it in a `with` clause, but that leaks temp files I guess.

Updated the test using your pattern.

docs/markdown/Machine-files.md
xclaesse: done.

nirbheek:
> Note: as a follow-up PR it could be interesting to also add support for `/` operator because I'm expecting this feature to be used a lot for building paths. `c = toolchain / 'gcc'` is nicer than `c = toolchain + '/gcc'`.

I actually think that the `/` operator is critical for use of this feature. In the interest of getting things going, I agree we should merge this first and then add `/` later, and block the 0.55 release on that. How does that sound? :)

TheQwertiest:
It should be noted though, that this would make machine files no longer compatible with `ini` format (I've seen a lot of machine files that were called `*.ini` and used as such).

xclaesse:
> I actually think that the `/` operator is critical for use of this feature. In the interest of getting things going, I agree we should merge this first and then add `/` later, and block the 0.55 release on that. How does that sound? :)

Getting `+` operator was a cheap trick we got for free, side effect of using `eval()`. If you've got no ideas how to implement `/` operator, seems to require writing a full custom parser, it's order of magnitude more complicated, but probably the right think to do anyway. But I wouldn't hold my breath on that.

xclaesse:
One easy trick I could think of is adding this into constants: `{'join_paths': os.path.join}` so we can have `c = join_paths(toolchain, 'gcc')`

xclaesse:
> It should be noted though, that this would make machine files no longer compatible with `ini` format: I've seen a lot of machine files that were called `*.ini` and used as such (hell, even this PR uses `*.ini` extension in examples).

I agree it's stretching the format a lot, and that's why I wasn't happy with this in the beginning. But I got convinced that the benefits by far outweigh the cons.

- We already support python-like syntax for string lists.
- The `+` operator feels natural with strings and lists.
- In my past projects I always get super long list of flags duplicated between `c_args` and `cpp_args`.
- With toolchain/sysroot installed in a long path, duplicating that path everywhere makes things hard to read (binaries, cflags, pkg_config_libdir, etc...).

nirbheek:
> It should be noted though, that this would make machine files no longer compatible with `ini` format: I've seen a lot of machine files that were called `*.ini` and used as such (hell, even this PR uses `*.ini` extension in examples).

Hmm, I think if people are writing/reading meson cross files as `.ini` files, they can choose to not use `+`. If they're expecting to consume arbitrary cross files as INI files that's their problem.

We should probably pick a different extension for our examples though.

TheQwertiest:
Mind you, I was not saying that I'm against this PR =)
One alternative would be is to add smth like the following to the docs:
```
Syntax used in machine files is a super-set of the basic `ini`-format (<https://en.wikipedia.org/wiki/INI_file>): all *base*(basic?) `ini` files will be readable by meson, but not all machine files will be readable with external `ini` parsers.
```

xclaesse:
@nirbheek @dcbaker Ok, replaced the usage of eval() with Meson's own mparser, and now it supports joining path with `/` operator :D

TheQwertiest:
@xclaesse , what would (and should) happen in the following scenarios?

file1.ini:
```ini
[constants]
a = 'a1'
b = a + 'b1'
```

file2.ini:
```ini
[constants]
a = 'a2'
```

file3.ini:
```ini
[constants]
b = a + 'b3'
```

`--n-f=file1.ini --n-f=file2.ini`: what is value of `b`?
`--n-f=file3.ini --n-f=file2.ini`: is it a valid scenario?

xclaesse:
Override happens before parsing values.

> `--n-f=file1.ini --n-f=file2.ini`: what is value of `b`?
`a='a2'` and `b='a2b1'`.

> `--n-f=file3.ini --n-f=file2.ini`: is it a valid scenario?

Error because `b` is defined before `a`.

TheQwertiest:
> Override happens before parsing values.

Nice! This should be probably mentioned in the docs =)

xclaesse:
> > Override happens before parsing values.
> 
> Nice! This should be probably mentioned in the docs =)

done.

docs/markdown/Machine-files.md
```diff
@@ -8,10 +8,83 @@ environments](Native-environments.md).
...
 - properties
 
+### constants
+
+*Since 0.55.0*
+
+String and list concatenation is supported using the `+` operator, joining paths
+is supported using the `/` operator.
+Entries defined in the `[constants]` section can be used in any other section
+(there are always parsed first), entries in any other section can be used only
```
dcbaker, line 23:
```suggestion
(they are always parsed first), entries in any other section can be used only
```

mesonbuild/envconfig.py
```diff
@@ -83,31 +83,69 @@
...
+                   (isinstance(l, list) and isinstance(r, list)):
+                    return l + r
+            elif node.operation == 'div':
+                if isinstance(l, str) and isinstance(r, str):
+                    return os.path.join(l, r)
+        raise EnvironmentException('Unsupported node type')
+
+    @staticmethod
+    def from_config_parser(filenames: T.List[str]) -> T.Dict[str, T.Dict[str, T.Dict[str, str]]]:
+        return MesonConfigFile(filenames).sections
```
dcbaker, line 148: This should be a class method and return `cls(filesnames).sections`

docs/markdown/snippets/machine_file_constants.md
```diff
@@ -0,0 +1,20 @@
+## Machine file constants
+
+Native and cross files now support string and list concatenation using the `+`
+operator, and joining paths using the `/` operator.
+Entries defined in the `[constants]` section can be used in any other section,
+entries in any other section can be used only within that same section and only
+after it has been defined.
```
dcbaker, line 7: you changed from plural `entries` to singular `it`, I think you probably want:
```suggestion
Entries defined in the `[constants]` section can be used in any other section.
An entry defined in any other section can be used only within that same section and only
after it has been defined.
```

mesonbuild/envconfig.py
dcbaker: From a design point of view though this is kinda icky, class that has two methods (that should be private/protected), some member variables, and is only called via a static method that instantiates an instance of itself, and returns a member. I think it would be nicer to make this a function. Even more importantly, this really belongs in coredata not in envconfig, as that avoids the need for the local import.

jpakkane:
Feature-wise this is ok. Feel free to merge once you are happy with the implementation.

mesonbuild/envconfig.py
xclaesse: Right, restructured all that now.

PR #7306, opened 2020-06-12
cmake: Fix handling of path seperators (fixes #7294)

### LINKED ISSUE #7294: Issue with env CMAKE_PREFIX_PATH and the host’s native path separator
**Describe the bug**
Building a project which dependencies ([libigl](https://github.com/libigl/libigl) and [alembic](https://github.com/alembic/alembic)) only provides cmake config files. They add their cmake path to the CMAKE_PREFIX_PATH environment variable using the host's native path separator (`:` in my case). Other projects that are using pure cmake finds these just fine. But meson can only find them separately, i.e. if only one of them is listed.

It will work if I change the separator to a `;`, then it will find both. Reading the cmake docs, [CMAKE_PREFIX_PATH](https://cmake.org/cmake/help/v3.17/variable/CMAKE_PREFIX_PATH.html) is defined as a semicolon-separated list. But reading the [search-procedure](https://cmake.org/cmake/help/v3.17/command/find_package.html#search-procedure) in the find-package documentation it looks like it is both. I.e. CMAKE_PREFIX_PATH defined from the commandline by `-D` or in a CMakeLists.txt is semicolon-separated, the environment variant of it is host specific (`:` for unix and `;` for windows).

Digging through the logs, it looks like meson does not take this in to consideration.

E.g. a snippet from the `CMakeLists.txt` meson generates when searching for alembic
```cmake
list(APPEND TMP_PATHS_LIST ${CMAKE_PREFIX_PATH})
#...
list(APPEND TMP_PATHS_LIST $ENV{CMAKE_PREFIX_PATH})
#...
```
**To Reproduce**

This is simplified version of the command I'm using:
```
 CMAKE_PREFIX_PATH=$LIBIGL_PACKAGE_ROOT/share/libigl/cmake:$ALEMBIC_LIBRARY_PATH/cmake/Alembic meson build -Dbuildtype=release
```
Where `LIBIGL_PACKAGE_ROOT`  contains the path to where the libigl package is installed (e.g. `/home/plattfot/libigl/2.0.0`) and `ALEMBIC_LIBRARY_PATH` contains the library path (e.g. `/home/plattfot/alembic/1.7.7/lib`). 

And the simplified meson.build file:
```
project('cmake-issue')
alembic_dep = dependency('Alembic', include_type: 'system')
igl_dep = dependency('libigl', modules: ['igl::core'], include_type: 'system')
```
Here follows some examples where it managed to find one of them or both.

This will find alembic, but fail to find libigl:
```
 CMAKE_PREFIX_PATH=$ALEMBIC_LIBRARY_PATH/cmake/Alembic  meson build -Dbuildtype=release
```

This will find libigl, but fail to find alembic (Need to change the order in the `meson.build`):
```
 CMAKE_PREFIX_PATH=$LIBIGL_PACKAGE_ROOT/share/libigl/cmake meson build -Dbuildtype=release
```

And this will find both, note changed the `:` to a `;`:
```
 CMAKE_PREFIX_PATH="$LIBIGL_PACKAGE_ROOT/share/libigl/cmake;$ALEMBIC_LIBRARY_PATH/cmake/Alembic" meson build -Dbuildtype=release
```
And is the current one I'm using to workaround the issue.

**Expected behavior**
Expect meson to find the dependencies with the colon separated environment variable `CMAKE_PREFIX_PATH`. As that is the behavior when using pure cmake.

**system parameters**
native build
cent6 with non standard install of dependencies
python-3.7.4
* what `meson --version`
0.54.2
1.9.0

mensinda:


nirbheek:
Conflicts and CI failures seem relevant

mensinda:
Should be fixed now. There was some cache that didn't get cleard, which caused issues.

nirbheek:
I think the CI failure in `UnusedMissingReturn / windows` is because `pip install ninja` doesn't download the binary anymore. It tries to build it, and fails.

mensinda:
Yeah, all CI errors are unrelated. The UnusedMissingReturn error now happens on all builds (see #7309 for instance).

nirbheek:
Looks like the latest 1.10 ninja update does not ship with Windows builds: 

https://pypi.org/project/ninja/1.10.0/#files

The previous version did:

https://pypi.org/project/ninja/1.9.0.post1/#files

So I guess we should pin the version. I'll open a PR for it.

nirbheek:
https://github.com/mesonbuild/meson/pull/7312

PR #7308, opened 2020-06-12
Make CompilerArgs subclassable

dcbaker:
The `CompilerArgs` class is a rather complicated beast, with deep rooted Gnu and C assumptions. There's already some workaround in the code for handling the way that D compilers handle linker arguments. It really doesn't make sense to keep special casing it in one giant mega class. I've tried to simplify things by putting a bit of code that is hopefully universally usable (like linking the same library) in a base class, while putting to the C-like logic in a subclass. To make it easy to ensure that you have the right arguments, the `Compiler` and `StaticLinker` classes now has a method to return the proper `CompilerArgs` iinstance.


dcbaker:
> My only question would be: how does this affect performance? I expect it should make things faster to have fewer runtime checks. The CompilerArgs class is in the hot path for our build.ninja generation, so it would be good to measure the improvement!

I measured the `d/7 multilib` test, and I could pretty consistantly get about .1 seconds better run time with this series than the master it was branched from. That's obviously not a huge difference, but it's not a huge project either.

nirbheek:
> I measured the `d/7 multilib` test, and I could pretty consistantly get about .1 seconds better configure time with this series than the master it was branched from. That's obviously not a huge difference, but it's not a huge project either.

As another datapoint, I tested with gst-build and the configure times were identical before and after.  (within +/- 0.1s). Ah well, maybe some day someone will rewrite it in C ;)

dcbaker:
I wouldn't really expect a C/C++/ObjC/ObjC++ based project to get faster, it's still going through basically the same path as before (but with the D workaround removed). for any other language it should be faster as it's not getting sucked through the Clike de duplication path.

jpakkane:
LGTM, feel free to merge once you are happy with the implementation.

Issue #7307, opened 2020-06-12
CMake test failures on CI due to Azure image update


TheQwertiest:
Two cmake module tests are failing on azure agents.

#### `test cases/cmake/5 object library`:
**Failure:** can't find zlib headers:
```
5 object library\subprojects\cmobjlib\libb.cpp(2): fatal error C1083: Cannot open include file: 'zlib.h': No such file or directory
```
**Reason:** `C:\Program Files\PostgreSQL\12\bin` was added to `PATH` on the agent, which contains `zlib.lib`, but not respective headers. It seems PATH has a priority over pkg-config in CMake.
</br>
#### `test cases/cmake/2 advanced`:
**Failure:**
```
 Traceback (most recent call last):
   File "D:\a\1\s\mesonbuild\mesonmain.py", line 131, in run
     return options.run_func(options)
   File "D:\a\1\s\mesonbuild\msetup.py", line 245, in run
     app.generate()
   File "D:\a\1\s\mesonbuild\msetup.py", line 159, in generate
     self._generate(env)
   File "D:\a\1\s\mesonbuild\msetup.py", line 192, in _generate
     intr.run()
   File "D:\a\1\s\mesonbuild\interpreter.py", line 4469, in run
     super().run()
   File "D:\a\1\s\mesonbuild\interpreterbase.py", line 493, in run
     self.evaluate_codeblock(self.ast, start=1)
   File "D:\a\1\s\mesonbuild\interpreterbase.py", line 518, in evaluate_codeblock
     raise e
   File "D:\a\1\s\mesonbuild\interpreterbase.py", line 511, in evaluate_codeblock
     self.evaluate_statement(cur)
   File "D:\a\1\s\mesonbuild\interpreterbase.py", line 526, in evaluate_statement
     self.assignment(cur)
   File "D:\a\1\s\mesonbuild\interpreterbase.py", line 1179, in assignment
     value = self.evaluate_statement(node.value)
   File "D:\a\1\s\mesonbuild\interpreterbase.py", line 528, in evaluate_statement
     return self.method_call(cur)
   File "D:\a\1\s\mesonbuild\interpreterbase.py", line 923, in method_call
     return obj.method_call(method_name, args, self.kwargs_string_keys(kwargs))
   File "D:\a\1\s\mesonbuild\interpreter.py", line 1813, in method_call
     value = fn(self.interpreter, state, args, kwargs)
[...]

PR #7309, opened 2020-06-12
 cmake: fix definitions with interface libraries (fixes #7299) 

### LINKED ISSUE #7299: target_compile_definitions from import cmake project not working correctly
I try to use this cmake library https://github.com/felixguendling/cista
In CmakeLists line: target_compile_definitions(cista INTERFACE CISTA_${CISTA_HASH}=1) passed to the compiler incorrectly.
On the cmake site https://cmake.org/cmake/help/latest/command/target_compile_definitions.html says that prefix -D not required, but meson doesn’t think so and pass to compiler two definitions at once: CISTA_FNV1A=1 -DCISTA_FNV1A=1 - ???
Change line to target_compile_definitions(cista INTERFACE -DCISTA_${CISTA_HASH}=1)
fix this problem.

But there is another library that does similar things and everything works fine:
target_compile_definitions(spdlog PUBLIC SPDLOG_COMPILED_LIB)
I think the problem is with INTERFACE or with parameter = 1

mensinda:
Test case for #7299


mensinda:
The CI errors are unrelated, see #7312 and #7307. Travis fails in OSX because of clock skew.

PR #7312, opened 2020-06-13
ci: Pin ninja to 1.9.0.post1 for now

nirbheek:
See: https://github.com/mesonbuild/meson/pull/7306#issuecomment-643606736


mensinda:
We could also use the prebuilt ninja exe going forward (although we would have to unzip it): https://github.com/ninja-build/ninja/releases

But pinning the version also seems like a good fix for now.

nirbheek:
> We could also use the prebuilt ninja exe going forward (although we would have to unzip it): https://github.com/ninja-build/ninja/releases

We used to do that, and we moved to pip to make things simpler.

PR #7315, opened 2020-06-13
Adding Swift templates to Meson

michaelbadcrumble:
Finally found some time to implemented Swift programming language templates. I also added a note and updated the test case to find the swift compiler.

### DIFF
--- a/run_unittests.py
+++ b/run_unittests.py
@@ -3368,6 +3368,11 @@ int main(int argc, char **argv) {
         except EnvironmentException:
             pass
         try:
+            env.detect_swift_compiler(MachineChoice.HOST)
+            langs.append('swift')
+        except EnvironmentException:
+            pass
+        try:
             env.detect_d_compiler(MachineChoice.HOST)
             langs.append('d')
         except EnvironmentException:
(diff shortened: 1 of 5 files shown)


michaelbadcrumble:
Ping @dcbaker and or @jpakkane.

run_unittests.py
```diff
@@ -3404,6 +3404,12 @@ def test_templates(self):
                 langs.append('rust')
             except EnvironmentException:
                 pass
+        if is_osx():
```
dcbaker, line 3407: Everything looks good here except this, swift works on oses other than macos.

michaelbadcrumble: So. Dose it work on Linux?

michaelbadcrumble: Ok did some digging and turns out Swift can run on Linux.

michaelbadcrumble: It’s not looking good doc. What do I do from here @dcbaker ?

dcbaker:
one of the documented commands tests is failing, you should be able to fix that pretty easily

michaelbadcrumble:
I have opened a new PR for swift template support with the latest frock to the Meson build project and was finally able to find that the documented commands test case. Should be here (#7339).

Also, there is an issue with the Swift support please go see the issue (#7337), thanks.

PR #7316, opened 2020-06-13
Disable failing cmake jobs on VS2017

nirbheek:
These only fail when building with msvc/clang-cl on the VS2017-Win2016
image. See: https://github.com/mesonbuild/meson/issues/7307

Attempt at making https://github.com/mesonbuild/meson/pull/7314 more specific.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -15,6 +15,8 @@ jobs:
 - job: vs2017
   pool:
     vmImage: VS2017-Win2016
+  variables:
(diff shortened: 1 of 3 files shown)


mensinda:
Sorry, I only just got onto my PC.

As for msys: those tests are skipped anyway since zlib isn't installed/found by CMake. The same goes for the vs2019 tests. The only "windows" tests that do not skip the test anyway is cygwin, which is it's own os for meson as far as I know.

Also, the way the skip with env vars is handled here is really hacky...

Issue #7317, opened 2020-06-14
Documentation issue in FAQ/generated headers


laolux:
**Describe the bug**
The example code in the FAQ section [How do I tell Meson that my sources use generated headers?](https://mesonbuild.com/FAQ.html#how-do-i-tell-meson-that-my-sources-use-generated-headers) does not work, even when filling in the missing parts. This is due to `CustomTargetHolder` and `list` being different types which do not play well together. The fix is very easy and should be incorporated to the documentation.

**To Reproduce**
Making a minimal working example from the code fragment in the FAQ does not allow me to run `meson build` and gives the error

    meson.build:5:0: ERROR: Invalid use of addition: can only concatenate list (not "CustomTargetHolder") to list

Here the files of my MWE.
First `meson.build`, as close to the example fragment as possible.

    project('example library', 'c')
    libfoo_gen_headers = custom_target('gen-headers', input: 'foo-template.h', output: 'foo-gen.h', command: ['cp', '@INPUT@', '@OUTPUT@'])
    libfoo_sources = files('foo-utils.c', 'foo-lib.c')
    libfoo = library('foo', sources: libfoo_sources + libfoo_gen_headers)

Next the dummy files `foo-utils.c`, `foo-lib.c` and `foo-template.h` for very simple trying at home.
`foo-utils.c`

    #include<foo-gen.h>
  
    int add(int a, int b){
        return a+b;
    }

`foo-lib.c`

    #include <foo-gen.h>
  
    int multiply(int a, int b){
        return a*b;
    }

`foo-template.h`

    int add(int a, int b);
    int multiply(int a, int b);

**Expected behavior**
`meson.build` should run without errors. This can be achieved by adjusting the `meson.build` with

    libfoo = library('foo', sources: libfoo_sources + libfoo_gen_headers.to_list())

**system parameters**
* Is this a [cross build](https://mesonbuild.com/Cross-compilation.html) or just a plain native build (for the same computer)? plain native build
* what operating system: Fedora 32
* what Python version: 3.8.3
* what `meson --version`: 0.54.2
* what `ninja --version` if it's a Ninja build: 1.10.0

dcbaker:
Yeah, that's completely bogus, what that should say is:

```meson
libfoo_gen_headers = custom_target('gen-headers', ..., output: 'foo-gen.h')
libfoo_sources = files('foo-utils.c', 'foo-lib.c')
# Add generated headers to the list of sources for the build target
[...]

PR #7318, opened 2020-06-14
Documentation, FAQ.md, generated_headers

laolux:
Fixed code fragment in the "How do I tell Meson that my sources use generated headers?" section in the FAQ.
Now this fragment will work without throwing an error.
Fixes issue  #7317

### DIFF
--- a/docs/markdown/FAQ.md
+++ b/docs/markdown/FAQ.md
@@ -432,7 +432,7 @@ sources in the build target:
 libfoo_gen_headers = custom_target('gen-headers', ..., output: 'foo-gen.h')
 libfoo_sources = files('foo-utils.c', 'foo-lib.c')
 # Add generated headers to the list of sources for the build target
-libfoo = library('foo', sources: libfoo_sources + libfoo_gen_headers.to_list())
+libfoo = library('foo', sources: [libfoo_sources + libfoo_gen_headers])
 ```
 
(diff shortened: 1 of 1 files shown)


docs/markdown/FAQ.md
```diff
@@ -432,7 +432,7 @@ sources in the build target:
 libfoo_gen_headers = custom_target('gen-headers', ..., output: 'foo-gen.h')
 libfoo_sources = files('foo-utils.c', 'foo-lib.c')
 # Add generated headers to the list of sources for the build target
-libfoo = library('foo', sources: libfoo_sources + libfoo_gen_headers)
+libfoo = library('foo', sources: libfoo_sources + libfoo_gen_headers.to_list())
```
jpakkane, line 435: Actually this should be `sources: [libfoo_sources, libfoo_gen_headers]`. People should not need to have to use `to_list` for these things, that would be bad usability.

laolux: I agree, it does look a lot cleaner. I added a new commit to update it accordingly.
Originally I wanted to amend the original commit, but not sure how to do so in github.
Hope this is fine, otherwise I can start out fresh and make a new, corrected pull request with only one commit.

jpakkane: I can squash it for now. Just remember in the future that doc-only changes should have `[skip ci]` at the end of their commit messages so CI is not run.

PR #7319, opened 2020-06-14
ci: Use test.json to skip VS2017 azure tests

mensinda:
See also #7307 #7314 #7316

cc @nirbheek


nirbheek:
This is nicer. I had forgotten that `test.json` can be used to do this. :)

PR #7326, opened 2020-06-15
Intel compiler enhancements

scivision:
* allow testcases/fortran/7generated to pass with IntelCl on Windows--CL-based Windows Fortran compilers generally are easier to use with static linking as shared linking requires special non-standard Fortran syntax just for CL. So like other such Meson Fortran project tests, we default to static linking.

* add a fallback version-finding regex, as Intel has moved to four-digit year-based versions with their new, no-cost [oneAPI compilers](https://software.intel.com/content/www/us/en/develop/tools/oneapi/hpc-toolkit.html).

Example:
```sh
ifort /logo
```
```
Intel(R) Visual Fortran Intel(R) 64 Compiler for applications running on Intel(R) 64, Version 2021.1 Beta Build 20200306
```

```sh
icl
```
```
Intel(R) C++ Intel(R) 64 Compiler for applications running on Intel(R) 64, Version 2021.1 Beta Build 20200306
```

### DIFF
--- a/test cases/fortran/7 generated/meson.build
+++ b/test cases/fortran/7 generated/meson.build
@@ -1,15 +1,15 @@
 # Tests whether fortran sources files created during configuration are properly
 # scanned for dependency information
 
-project('generated', 'fortran')
+project('generated', 'fortran',
+  default_options : ['default_library=static'])
 
 conf_data = configuration_data()
 conf_data.set('ONE', 1)
 conf_data.set('TWO', 2)
 conf_data.set('THREE', 3)
 
-outfile = configure_file(
-      input : 'mod3.fpp', output : 'mod3.f90', configuration : conf_data)
+configure_file(input : 'mod3.fpp', output : 'mod3.f90', configuration : conf_data)
 # Manually build absolute path to source file to test
 # https://github.com/mesonbuild/meson/issues/7265
 three = library('mod3', meson.current_build_dir() / 'mod3.f90')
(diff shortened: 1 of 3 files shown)


test cases/fortran/7 generated/meson.build
```diff
@@ -1,18 +1,19 @@
...
 conf_data.set('TWO', 2)
 conf_data.set('THREE', 3)
 
 outfile = configure_file(
-      input : 'mod3.fpp', output : 'mod3.f90', configuration : conf_data)
+  input : 'mod3.fpp', output : 'mod3.f90', configuration : conf_data)
 # Manually build absolute path to source file to test
 # https://github.com/mesonbuild/meson/issues/7265
-three = library('mod3', meson.current_build_dir() / 'mod3.f90')
+three = library('mod3', outfile)
```
nirbheek, line 16: This change should not be here, right? https://github.com/mesonbuild/meson/pull/7266#issuecomment-643917454

test cases/fortran/7 generated/meson.build
```diff
@@ -1,18 +1,19 @@
...
+  default_options : ['default_library=static'])
 
 conf_data = configuration_data()
 conf_data.set('ONE', 1)
 conf_data.set('TWO', 2)
 conf_data.set('THREE', 3)
 
 outfile = configure_file(
-      input : 'mod3.fpp', output : 'mod3.f90', configuration : conf_data)
+  input : 'mod3.fpp', output : 'mod3.f90', configuration : conf_data)
```
nirbheek, line 13: Spurious whitespace? We use a single tab for function call continuation elsewhere.

PR #7327, opened 2020-06-15
Add --type to `subprojects foreach`

fiesh:
This allows restricting a `foreach` call to subprojects of a certain
type.  This is useful when, for example, only `git` repositories should
be `foreach`ed since we want to run a `git` command on them.

### DIFF
--- a/mesonbuild/msubprojects.py
+++ b/mesonbuild/msubprojects.py
@@ -169,6 +169,8 @@ def download(wrap, repo_dir, options):
         mlog.log('  ->', mlog.red(str(e)))
 
 def foreach(wrap, repo_dir, options):
+    if options.type and wrap.type != options.type:
+        return
     mlog.log('Executing command in {}'.format(repo_dir))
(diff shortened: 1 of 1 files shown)


xclaesse:
I have a slight preference to add `--git` instead of `--type git`. I don't know if there is a use-case for that, but that would allow doing `--git --hg` to go through all git and hg but not others.

fiesh:
> I have a slight preference to add `--git` instead of `--type git`. I don't know if there is a use-case for that, but that would allow doing `--git --hg` to go through all git and hg but not others.

That's how I had implemented it originally, but then I figured a flag for all kind of types would be more uniform.  I'm open to both, and `--git` is slightly shorter and thus better than `--type git`.  This is countered by the disadvantage of having four options `--file`, `--git`, `--hg`, and `--svn`.

xclaesse:
bah, I don't have strong opinion, I'm fine either way. Anyone else has preference?

tristan957:
Closing since we have `--types`

PR #7328, opened 2020-06-15
docs: replaced ninja with appropriate meson commands [skip ci]

TheQwertiest:
Replaced direct `ninja` invocations with equivalent `meson` commands.

Reasoning:
- Users don't and shouldn't care about the backend in most cases. It's an implementation detail. One of the most frequent question that I was asked when introducing meson is `what is ninja?`/`is ninja part of the meson?`/`why do I have to invoke ninja/.sln directly?`
- Backend-agnostic meson methods cover 99% of common ninja scenarios (1% being specifying targets(https://github.com/mesonbuild/meson/pull/7181) and backend arguments(https://github.com/mesonbuild/meson/pull/7183)). And most of these methods provide additional functionality not available via direct ninja invocation (most prominent example would be `meson test`).
- Meson should be about meson promotion and not ninja promotion =)
- No one ever invokes backend directly when using CMake. Everyone uses `cmake --build`/`cmake --install` and etc. Having the same approach in meson will ease the transition for people moving  away from CMake.

Leftover `ninja TARGET` invocations will be mesonified if/when https://github.com/mesonbuild/meson/pull/7181 is merged.

### DIFF
--- a/docs/markdown/Commands.md
+++ b/docs/markdown/Commands.md
@@ -183,21 +183,7 @@ optional arguments:
   --no-tests             Do not build and test generated packages.
 ```
 
-This creates a file called `projectname-version.tar.xz` in the build
-tree subdirectory `meson-dist`. This archive contains the full
-contents of the latest commit in revision control including all the
-submodules (recursively). All revision control metadata is removed.
-Meson then takes
-this archive and tests that it works by doing a full compile + test +
-install cycle. If all these pass, Meson will then create a SHA-256
-checksum file next to the archive.
-
-**Note**: Meson behaviour is different from Autotools. The Autotools
-"dist" target packages up the current source tree. Meson packages
-the latest revision control commit. The reason for this is that it
-prevents developers from doing accidental releases where the
-distributed archive does not match any commit in revision control
-(especially the one tagged for the release).
+See [notes about creating releases](Creating-releases.md) for more info.
 
 #### Examples:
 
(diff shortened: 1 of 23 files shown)


scivision:
This topic is something I found a tripping point when learning Meson. The `meson compile` abstracts away Ninja for almost all common uses as you note. So I think this is a good change.

TheQwertiest:
I wonder, should I make (in a separate PR) a *meson-ification* pass through various tests as well (e.g. run_unit_tests.py)? This would improve `meson CMD` test coverage.

jpakkane:
:crossed_swords: conflicts.

TheQwertiest:
Rebased

eli-schwartz:
This PR and its followup were terrible ideas. Now users instead complain that "I tried to run `meson compile <target>` per the instructions and it returns super weird errors" and get told "oh yeah, that's incredibly dependent on which version of meson you use".

Because you're trying to **_blackhole_** information about the backend.

Also:
>     * No one ever invokes backend directly when using CMake.

What a... lovely unqualified absolutism, this is definitively untrue. Especially if you use ninja instead of Make, and ninja doesn't support $MAKEFLAGS and is diametrically opposed on religious principle to adding a $NINJAFLAGS feature and samurai doesn't want to support $SAMUFLAGS other than -v/-j.

"Well, just use --ninja-args" -- or, if I know what backend I'm using and care about passing arguments to it, I could just not bother using a very complicated, fiddly, version-dependent wrapper that doesn't actually help me avoid hardcoding backend-specific info.

tl;dr at least #7397 should be reverted. The people who will get confused by that one don't even get an error message stating that the "compile" command doesn't exist.

PR #7332, opened 2020-06-16
depfixer: zero the path if rpath is empty

### LINKED ISSUE #7331: RUNPATH is not cleared when install_rpath is empty
The documentation says:

```build_rpath - a string to add to target's rpath definition in the build dir, but which will be removed on install```

I have a project `x` with a subproject `y` and use `executable(..., install_rpath: '')`.

When I run `strings` on the installed binary, `I still see the rpath set by Meson`, although `readelf` says RUNPATH is missing, as it should be.

To check why this happens, I replaced `''` with `'abcd'` and this is what the path inside the binary looks like now (`$ORIG` has been replaced with `abcd\0`):

```LIBC_2.28^@GLIBC_2.17^@abcd^@IN/subprojects/y^@_ITM_deregisterTMCloneTable^@__gmon_start__^@_ITM_registerTMCloneTable^@^@^@```

This happens because depfixer.py doesn't zero the remaining length of the string, when rpath is changed, and doesn't clear the string when rpath is removed.

In some cases, it's undesirable to have things like names of subprojects (which are often the names of libraries), in a binary.

dimkr:
Fixes #7331.

### DIFF
--- a/mesonbuild/scripts/depfixer.py
+++ b/mesonbuild/scripts/depfixer.py
@@ -324,6 +324,15 @@ class Elf(DataSizes):
 
         if len(old_rpath) < len(new_rpath):
             sys.exit("New rpath must not be longer than the old one.")
+
+        # Clear the old rpath: if we're going to remove the entry, the path will
+        # remain in the binary, unused. However, it may diclose information,
+        # like the fact the binary was built with Meson, and the subprojects
+        # used. If we're going to change the rpath, we want to trim trailing
+        # junk (after the terminating \0), for the same reason.
+        self.bf.seek(rp_off)
+        self.bf.write(b'\0' * len(old_rpath))
+
         # The linker does read-only string deduplication. If there is a
         # string that shares a suffix with the rpath, they might get
         # dedupped. This means changing the rpath string might break something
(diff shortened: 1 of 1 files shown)


jpakkane:
The test failures are relevant. This approach probably zeroes out too much data.

dankegel:
In case it's helpful, I wrote a test case to verify that the build directory
name didn't show up in the executable, see
https://github.com/mesonbuild/meson/pull/7336/commits/f473752fc18511f3b8f47923820d773af00c1a35

Feel free to use that in your pull request if you like.

dimkr:
> The test failures are relevant. This approach probably zeroes out too much data.

I'll fix them.

> Feel free to use that in your pull request if you like.

Thanks!

dimkr:
The problem was addition of RUNPATH directories using LDFLAGS, link_args, etc'. I changed the scope of this PR, so it clears the string only if all directories are directories added by Meson, like subproject directories.

dimkr:
It's a very problematic change. The failure `./test cases/unit/7 run installed` raises a good point: there may be references to substrings of the string holding RUNPATH, e.g. the addition of `puts("foo")` might print nothing, if RUNPATH contains `subprojects/foo` and this string was replaced with `\0`s.

I wonder if `puts("$ORIGIN")` breaks after installation, too, even without my changes.

Issue #7337, opened 2020-06-17
Swift projects are generating strange files


michaelbrockus:
**Describe the bug**
Was adding Swift templates to Meson build when this happened. I also get strange files.

```console
The Meson build system
Version: 0.54.999
Source dir: /private/var/folders/cx/hf5684cj1dbckp0gjqphzyfw0000gp/T/tmpp9jze1qv
Build dir: /private/var/folders/cx/hf5684cj1dbckp0gjqphzyfw0000gp/T/tmpp9jze1qv/builddir
Build type: native build
Project name: tmpp9jze1qv
Project version: 0.1
Swift compiler for the host machine: swiftc (llvm 5.1.3)
Swift linker for the host machine: swiftc ld64 530
Host machine cpu family: x86_64
Host machine cpu: x86_64
Build targets in project: 2

Found ninja-1.10.0 at /Users/mike/Library/Python/3.8/bin/ninja

[1/5] Compiling Swift source /private/var/folders/cx/hf5684cj1dbckp0gjqphzyfw0000gp/T/tmpp9jze1qv/tmpp9jze1qv.swift
[2/5] Linking static target libtmpp9jze1qv.a
[3/5] Compiling Swift source /private/var/folders/cx/hf5684cj1dbckp0gjqphzyfw0000gp/T/tmpp9jze1qv/tmpp9jze1qv.swift
[4/5] Compiling Swift source /private/var/folders/cx/hf5684cj1dbckp0gjqphzyfw0000gp/T/tmpp9jze1qv/tmpp9jze1qv_test.swift
[5/5] Compiling Swift source 
FAILED: tmpp9jze1qv_test 
/usr/local/opt/python@3.8/bin/python3.8 /Users/mike/Desktop/github/meson/meson.py --internal dirchanger tmpp9jze1qv_test.p swiftc -o /private/var/folders/cx/hf5684cj1dbckp0gjqphzyfw0000gp/T/tmpp9jze1qv/builddir/tmpp9jze1qv_test -L /private/var/folders/cx/hf5684cj1dbckp0gjqphzyfw0000gp/T/tmpp9jze1qv/builddir -emit-executable tmpp9jze1qv_test.o /private/var/folders/cx/hf5684cj1dbckp0gjqphzyfw0000gp/T/tmpp9jze1qv/builddir/libtmpp9jze1qv.a 
duplicate symbol '_main' in:
    tmpp9jze1qv_test.o
[...]

PR #7339, opened 2020-06-17
Add Swift templates attempt 2

michaelbadcrumble:
Adding Swift templates to Meson build.


TheQwertiest:
The easier (and a proper) way to update Commands.md is to copy paste the output of 

```bash
COLUMNS=80 meson CMD --help
```

This way you won't have to *guess* the newlines and stuff.

michaelbadcrumble:
I’m starting to feel uncomfortable about this PR since there is an issue here #7337 also it seems that the command checker was just added so I’m going to what for that to mature and be documented before trying again.

PR #7342, opened 2020-06-17
ci: skip some CMake tests on VS2019

mensinda:

PR #7343, opened 2020-06-17
ci: Remove PostgreSQL from PATH

nirbheek:
It's one of the causes of the cmake test failures, and it's also plaguing the VS2019 jobs now because of the image update.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -15,8 +15,6 @@ jobs:
 - job: vs2017
   pool:
     vmImage: VS2017-Win2016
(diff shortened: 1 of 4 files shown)


nirbheek:
Need to merge either this or https://github.com/mesonbuild/meson/pull/7342.

nirbheek:
```
<mensinda> Sure, this is indeed the case, but constantly fixing the CI gets really annoying with time. So, yeah we can go with your PR and let's hope that MS doesn't break our setup again...
```

PR #7346, opened 2020-06-17
Made Commands.md dynamically generated

TheQwertiest:
Replaced `docs/markdown/Commands.md` with dynamic generation from template:
- Added a helper script `tools/regenerate_docs.py`:
  This script generates `docs/markdown/Commands.md` from `docs/markdown_dynamic/Commands.md`, using the current `meson.py` to get the relevant data. Template rendering is implemented via `jinja2` Python package.
- Added a call to `tools/regenerate_docs.py` from `docs/meson.build` (which is used to generate site docs during release).
- Rewrote corresponding unit-test: now it only checks two things:
  - All commands are present (i.e. section names).
  - All commands have corresponding place-holders.


TheQwertiest:
PS: I can rewrite the test proeprly if the `jinja2` requirement can be satisfied somehow.

TheQwertiest:
Another option would be to move this script invocation to `meson/docs/meson.build` and template to `docs/markdown`. This way there would be no need to test any options at all (only commands presence), since docs will always contain the up-to-date info. There won't be the need for user to manually invoke the script as well.

TheQwertiest:
Added a test to enforce template usage: test checks if the `markdown/Commands.md` is consistent with `templates/Commands.md`

TheQwertiest:
Rebased to resolve conflicts

jpakkane:
Fully dynamic is better. Whenever something can be done so it works without manual labor, it should be done.

TheQwertiest:
@jpakkane , done. But please check if it works on your machine as well, since site generation (an upload) is not covered by any tests (AFAIK).

jpakkane:
It currently fails because `regenerate_docs.py` does not have the executable bit set.

Testing it yourself is simple:

- Install hotdoc (most easily via pip)
- go in the `docs` subdir
- run `../meson.py . builddir && ninja -C builddir`

TheQwertiest:
Strange, it works for me :\
I'll add the flag anyway =)

TheQwertiest:
Added a flag, renamed `templates` to `markdown_dynamic` (for better discoverability) and rebased for good measure :D

PR #7349, opened 2020-06-17
Update commands used in CI examples [skip ci]

michaelbadcrumble:
Commands used in examples seem to be outdated. It should be fixed now.


TheQwertiest:
You are slowly recreating the following PR: https://github.com/mesonbuild/meson/pull/7328

michaelbadcrumble:
Oh sorry, I promise that the style guide is the last one.

TheQwertiest:
> Oh sorry, I promise that the style guide is the last one.

Haha, I didn't mean it like that =)
But rather that you don't need to make PRs that only convert `ninja` commands to `meson`, since it's already covered by the mentioned PR.

michaelbadcrumble:
Ok then in that case I will close pull request that would have conflicts against PR: #7328

PR #7355, opened 2020-06-18
Update Users.md

lock042:
Update Users.md with the add of Siril (www.siril.org) that is migrating from autotools to meson.


jpakkane:
Thanks, just note in the future that doc-only changes should have `[skip ci]` at the end of their commit messages, so CI is not run.

lock042:
Oh sorry? Indeed I thought it was strange to run all CI for minor changes ;). Thanks for the TIP.
Also, I'm not dev of GIMP but you could note that gimp is migrating to meson too.

Cheers, and thanks for your wonderful tool.

PR #7356, opened 2020-06-18
path inclusion fix for windows

p01arst0rm:
Fixes an issue with path splitting on windows. whitespace is required in some default windows paths, but it is not escaped. to properly allow the whitespace to be included in the file path, the entire -DCMAKE_PREFIX_PATH value needs to be quoted ( with `"` ). this results in the correct path being passed. 

The command also split the values of CMAKE_PREFIX_PATH based on `:`, meaning it would break the full path extension. This causes problems if you are trying to include a prefix on a non-local disk drive (aka `D:\`).

another point to highlight is how setting environment variables works in windows.

if i was to:
`SET CMAKE_PREFIX_PATH=C:\Progam files\boost;C:\Program Files\zlib`

the resulting string that would be passed to cmake inside meson would be:

-DCMAKE_PREFIX_PATH=C:\Progam Files\boost;C:\Program Files\zlib`

which would cause an error, as you have not escaped the spaces.

if i was to:
`SET CMAKE_PREFIX_PATH="C:\Progam files\boost;C:\Program Files\zlib"`

meson would internally store this as:
`env_pref_path ['"C:\Progam Files\boost', 'C:\Program Files\zlib"']`

as `env_pref_path` is appended to `prefix_paths`, this could cause a runtime error.

say for example ;- 
`prefix_paths=['C:\Program Files']`

the `env_prev_path` append would result in the following being sent to cmake internally:
`-DCMAKE_PREFIX_PATH=C:\Progam Files\;"C:\Progam Files\boost;C:\Program Files\zlib"`

this also causes a runtime error.

these changes should fix this from happening, but ideally `"` should also be stripped 
from the paths, mayhaps other special characters along with it.

### DIFF
--- a/mesonbuild/envconfig.py
+++ b/mesonbuild/envconfig.py
@@ -145,6 +145,24 @@ def get_env_var(for_machine: MachineChoice,
         var, value = ret
         return value
 
+def get_env_path(for_machine: MachineChoice,
+                is_cross: bool,
+                var_name: str) -> T.Tuple[T.Optional[str], T.Optional[str]]:
+    ret = get_env_var_pair(for_machine, is_cross, var_name)
+    if ret is None:
+        return None
+    else:
+        var, value = ret
+        mlog.debug(ret)
+        # Filter based on path seperator
+        if value.find(';'):
+            value = value.split(r';')
+        else:
+            value = value.split(r':')
+        # Filter out empty strings
+        value = [x for x in value if x]
+        return value
+
 class Properties:
     def __init__(
             self,
--- a/mesonbuild/cmake/executor.py
+++ b/mesonbuild/cmake/executor.py
@@ -28,7 +28,7 @@ import textwrap
 from .. import mlog, mesonlib
 from ..mesonlib import PerMachine, Popen_safe, version_compare, MachineChoice
 from ..environment import Environment
-from ..envconfig import get_env_var
+from ..envconfig import get_env_path
 
 if T.TYPE_CHECKING:
     from ..dependencies.base import ExternalProgram
@@ -64,13 +64,11 @@ class CMakeExecutor:
             return
 
         self.prefix_paths = self.environment.coredata.builtins_per_machine[self.for_machine]['cmake_prefix_path'].value
-        env_pref_path = get_env_var(
+        env_pref_path = get_env_path(
             self.for_machine,
             self.environment.is_cross_build(),
             'CMAKE_PREFIX_PATH')
         if env_pref_path is not None:
-            env_pref_path = re.split(r':|;', env_pref_path)
-            env_pref_path = [x for x in env_pref_path if x]  # Filter out empty strings
             if not self.prefix_paths:
                 self.prefix_paths = []
             self.prefix_paths += env_pref_path


jpakkane:
Test failures seem relevant:

```
test cases/unit/65 cmake parser/meson.build:3:0: ERROR: Dependency "mesontest" not found, tried pkgconfig and cmake

___________________ AllPlatformTests.test_cmake_prefix_path ____________________
[gw0] linux -- Python 3.7.5 /usr/bin/python3
self = <run_unittests.AllPlatformTests testMethod=test_cmake_prefix_path>
    @skip_if_no_cmake
    def test_cmake_prefix_path(self):
        testdir = os.path.join(self.unit_test_dir, '64 cmake_prefix_path')
>       self.init(testdir, extra_args=['-Dcmake_prefix_path=' + os.path.join(testdir, 'prefix')])
```

p01arst0rm:
wait what i never closed this

p01arst0rm:
> Test failures seem relevant:
> 
> ```
> test cases/unit/65 cmake parser/meson.build:3:0: ERROR: Dependency "mesontest" not found, tried pkgconfig and cmake
> 
> ___________________ AllPlatformTests.test_cmake_prefix_path ____________________
> [gw0] linux -- Python 3.7.5 /usr/bin/python3
> self = <run_unittests.AllPlatformTests testMethod=test_cmake_prefix_path>
>     @skip_if_no_cmake
>     def test_cmake_prefix_path(self):
>         testdir = os.path.join(self.unit_test_dir, '64 cmake_prefix_path')
> >       self.init(testdir, extra_args=['-Dcmake_prefix_path=' + os.path.join(testdir, 'prefix')])
> ```

i can't understand why this test is failing? the string passed to cmake seems perfectly fine

```
Called `/usr/sbin/cmake -G Ninja -DNAME=mesontest -DARCHS= -DCOMPS= --trace-expand --trace-format=json-v1 --no-warn-unused-cli --trace-redirect=cmake_trace.txt . -DCMAKE_PREFIX_PATH="/__w/meson/meson/test cases/unit/64 cmake_prefix_path/prefix"` in /__w/meson/meson/tmpf7tuesua/meson-private/cmake_mesontest -> 0
```

p01arst0rm:
these tests fail on windows as it is with no changes. as meson currently is, this is the following path given to cmake on windows

```
"C;\\Users\\p01ar\\git\\upstream\\test cases\\unit\\65 cmake parser\\prefix\";\"C:\\Users\\p01ar\\git\\upstream\\test cases\\unit\\65 cmake parser\\prefix\";C:/Program Files;C:/Program Files (x86);C:/Program Files/CMake;C:/Program Files (x86)/MesonTemp"
```

scivision:
The auto-issue closing can happen on force pushes where the original change gets overwritten. It's a known GitHub quirk.

p01arst0rm:
new commit fixes env_pref_path; tested on Windows 10 (patch KB4538461)

p01arst0rm:
log of the change to the path:
```
Using 'CMAKE_PREFIX_PATH' from environment with value: 'C:\\Program Files(x86);C:\\Program Files;C:\\Program Files (x86);'
('CMAKE_PREFIX_PATH', 'C:\\Program Files(x86);C:\\Program Files;C:\\Program Files (x86);')
OLD PATH: ["C","\Program Files(x86)","C","\Program Files","C","\Program Files (x86)"]
NEW PATH: ['C:\\Program Files(x86)', 'C:\\Program Files', 'C:\\Program Files (x86)']
Extracting basic cmake information
Try CMake generator: auto
```

p01arst0rm:
bumping PR as was never merged/closed.

xclaesse:
It has conflicts, can you rebase please?

mesonbuild/envconfig.py
```diff
@@ -145,6 +145,24 @@ def get_env_var(for_machine: MachineChoice,
...
+    if ret is None:
+        return None
+    else:
+        var, value = ret
+        mlog.debug(ret)
+        # Filter based on path seperator
+        if value.find(';'):
+            value = value.split(r';')
+        else:
+            value = value.split(r':')
```
xclaesse, line 161: Shouldn't this be `value.split(os.pathsep)`? Also, why using `r''` for split() but not find()?

Issue #7358, opened 2020-06-19
Unknown CPU family 'sh4'


malisal:
I noticed there is no support for Renesas SuperH CPU architecture.

Even though the warning ("Unknown CPU family") is reported, the builds works just fine. I have attached the relevant output for reference.

```
$ uname -a
Linux test 5.6.7 #1 Fri May 20 14:02:13 CEST 2020 sh4 GNU/Linux
$ cat /proc/cpuinfo
machine		: RTS7751R2D
processor	: 0
cpu family	: sh4
cpu type	: SH7751R
cut		: unknown
cpu flags	: fpu ptea
cache type	: split (harvard)
icache size	:  4KiB (2-way)
dcache size	:  4KiB (2-way)
address sizes	: 29 bits physical
bogomips	: 120.00
```

PR #7359, opened 2020-06-19
Add SuperH SH-4 support

dcbaker:
Fixes: #7358


jpakkane:
Should the family be `superh` instead, so it would cover [all the different SuperH processors](https://en.wikipedia.org/wiki/SuperH#Models)?

dcbaker:
I thought about that, except that GCC differentiates them with -m flags like it does for x86 and x86_64, and I couldn't get a clear answer whether you could compile code that works on all superH chips, ie, could you build a binary that runs on both sh-2 and sh-4, which is pretty much how we break down cpu_family. (ie, you can build code that runs on both a 486 and a brand new i7, and we call that x86)

PR #7361, opened 2020-06-20
Handle cmake dependencies which require a specified version

jekstrand:
Some CMake packages fail to find at all if no version is specified. This PR adds a cmake_version parameter to dependency() to allow you to specify the requested version.

This is my first meson PR and I have absolutely no notion whatsoever that this is the "right" way to do things.  Feel free to tell me how to do it properly. :smile_cat:


mesonbuild/dependencies/base.py
```diff
@@ -1085,6 +1085,7 @@ def __init__(self, name: str, environment: Environment, kwargs, language: T.Opti
         if self.cmakeinfo is None:
             raise self._gen_exception('Unable to obtain CMake system information')
 
+        version = kwargs.get('cmake_version', '')
```
mensinda, line 1088: There should be a check that `version` is a `str` (since this is user input)

mensinda:
There is also https://mesonbuild.com/Contributing.html with some more in-depth contributing guidelines.

mesonbuild/dependencies/base.py
jekstrand: Should be fixed in the latest version.

jekstrand:
I just pushed a new version with the following changes:
 - Added a release notes snippet
 - Added documentation
 - Renamed the new kwarg to `cmake_package_version`
 - Added a check to throw an exception if it's not a string

Regarding the name of the kwarg, I'm open to suggestions.  `cmake_package_version`, while explicit, seems a bit verbose.  Maybe simply `requested_version`?  It's not quite the same as `version` which contains a comparison.  As much as I'd like to make it auto-magic and just try to parse `version` and try to figure it out, that seems highly error-prone especially with as stupid custom as CMake stuff tends to be, hence making it a different kwarg.

mensinda:
OK, the test error looks weird, not sure what is going on here. Please try to fix this, since we can't merge a PR with a broken test. It could also be a caching issue since all other dependencies are cached (caching is done around `interpreter.py:3557`).

jekstrand:
Yeah, I get that we can't merge with a failing test.  I'll try to get it fixed.  This just isn't the only thing on my plate and my emotional energy for fighting with CMake is finite. :frowning:

dcbaker:
Closed in favor of: #7846

as @jekstrand asked me to get this landed for him.

PR #7367, opened 2020-06-22
Support native armel/armhf builds on arm64 hosts.

markhindley:
arm64 (aarch64) can natively build and run code for armel and armhf architectures (analagously to amd64 and i386). This patch add support for that.

Mark


markhindley:
What should I do about the failed CI tests? AFAICS they have just timed out.

markhindley:
@jpakkane: Just wondering if there is anything else I should do to facilitate review of this MR?

PR #7370, opened 2020-06-23
Solaris fixes

alanc:
The primary changes are adding Solaris support to symbolextractor (by using the GNU binutils and wrapping the GNU version), and adding asneeded & pie support to the Solaris linker mixin.

The rest are fixing up various test cases to run correctly on Solaris.

### DIFF
--- a/mesonbuild/linkers.py
+++ b/mesonbuild/linkers.py
@@ -1098,6 +1098,13 @@ class SolarisDynamicLinker(PosixDynamicLinkerMixin, DynamicLinker):
             return args
         return self._apply_prefix('--whole-archive') + args + self._apply_prefix('--no-whole-archive')
 
+    def get_pie_args(self) -> T.List[str]:
+        # Available in Solaris 11.2 and later
+        return ['-z', 'type=pie']
+
+    def get_asneeded_args(self) -> T.List[str]:
+        return self._apply_prefix(['-z', 'ignore'])
+
     def no_undefined_args(self) -> T.List[str]:
         return ['-z', 'defs']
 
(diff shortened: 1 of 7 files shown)


mesonbuild/linkers.py
```diff
@@ -1098,6 +1098,13 @@ def get_link_whole_for(self, args: T.List[str]) -> T.List[str]:
             return args
         return self._apply_prefix('--whole-archive') + args + self._apply_prefix('--no-whole-archive')
 
+    def get_pie_args(self) -> T.List[str]:
+        # Available in Solaris 11.2 and later
+        return ['-z', 'type=pie']
```
dcbaker, line 1103: It doesn't look like the version in openindiana has this option (although my vm is a little out of date). If they haven't added it, you'll need to check to see if this works before using it.

dcbaker:
And it looks like I was slow on pressing submit :/

mesonbuild/linkers.py
alanc: You appear to be right about the illumos-based OS's.  I don't see any other examples of checking linker options in linkers.py - any suggestions on how to do this?

alanc: Done in #7422

mesonbuild/scripts/symbolextractor.py
```diff
@@ -121,6 +121,13 @@ def gnu_syms(libfilename: str, outfilename: str):
         result += [' '.join(entry)]
     write_if_changed('\n'.join(result) + '\n', outfilename)
 
+def solaris_syms(libfilename: str, outfilename: str):
+    # gnu_syms() works with GNU nm & readelf, not Solaris nm & elfdump
+    origpath = os.environ['PATH']
+    os.environ['PATH'] = '/usr/gnu/bin:' + origpath
+    gnu_syms(libfilename, outfilename)
+    os.environ['PATH'] = origpath
```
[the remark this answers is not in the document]
alanc: Done in #7422

PR #7373, opened 2020-06-24
MPI: corrections to generalize across platforms and MPI implementations

### LINKED ISSUE #7045: Using Intel MPI with GCC
**Describe the bug**
It seems that in Meson 0.54.1, it is not possible anymore to use the mpi dependency to detect IntelMPI when using gcc.
This worked in 0.53 and is probably due to the change in behavior in baca2cd07d632dd5d944b2b364a297a0adf42028

**To Reproduce**
```
project('mpitest', 'c')
dependency('mpi')
```
with Intel `mpicc` in `$PATH`.

1. run meson with CC=icc: succeeds
2. run meson with CC=gcc
```
mpicc found: NO
Run-time dependency MPI found: NO (tried pkgconfig and config-tool)
```

**Expected behavior**
Meson should use the available `mpicc` to find the correct MPI implementation independent of the compiler.
I am also not sure if the comment
`# OpenMPI, which doesn't work with intel` in the code is accurate. I remember using OpenMPI with icc working just fine. Detecting whether we have Intel or OpenMPI based on compiler type therefore does not work.

**system parameters**
* native build
* CentOS 7.7
* Python 3.7.3
* Meson 0.45.1
* 1.9.0.git.kitware.dyndep-1.jobserver-1

scivision:
fixes #7045

Tested on:

* Windows MSYS2 with MS-MPI and GCC  (new!)
* OpenMPI and GCC
* IntelMPI and GCC (Linux)
* IntelMPI and Intel compilers (Windows, Linux)

---

Also made minor correction to default Intel compiler options on Windows


mesonbuild/dependencies/mpi.py
```diff
@@ -196,6 +207,33 @@ def _sanitize_version(self, out: str) -> str:
         return out
 
 
+class MSMPIConfigToolDependency(_MPIConfigToolDependency):
```
dcbaker, line 220: Does it really make sense to inherit the `_MPIConfigToolDependency` here? It seems like you're just working around it. maybe it would make more sense to just inherit `ConfigToolDependency` directly?

mesonbuild/dependencies/mpi.py
```diff
@@ -54,23 +57,45 @@ def mpi_factory(env: 'Environment', for_machine: 'MachineChoice',
...
+                cls_tools = [(MSMPIConfigToolDependency, tool_names)]
+        else:
+            # Linux, MacOS: try the environment variables for the tools first;
+            # fall back to hardcoded names
+
+            # first, IntelMPI
+            # https://software.intel.com/content/www/us/en/develop/documentation/mpi-developer-reference-linux/top/command-reference/compilation-commands.html
             if language == 'c':
-                tool_names = [os.environ.get('I_MPI_CC'), 'mpiicc']
+                tool_names = [os.environ.get('I_MPI_CC'), 'mpigcc']
```
dcbaker, line 91: I don't have an `mpigcc` installed from openmpi or from intel?

mesonbuild/dependencies/mpi.py
```diff
@@ -79,14 +104,13 @@ def mpi_factory(env: 'Environment', for_machine: 'MachineChoice',
...
 
-        tool_names = [t for t in tool_names if t]  # remove empty environment variables
-        assert tool_names
-
-        nwargs['tools'] = tool_names
-        candidates.append(functools.partial(
-            cls, tool_names[0], env, nwargs, language=language))
+        for cls, names in cls_tools:
+            args = nwargs.copy()  # else candidates interfere with each other
+            args['tools'] = [t for t in names if t]
```
dcbaker, line 111: I think we should keep the assert here.

mesonbuild/dependencies/mpi.py
```diff
@@ -195,6 +217,33 @@ def _sanitize_version(self, out: str) -> str:
...
+        if not self.is_found:
+            return
+
+        args = self.get_config_value(['-show'], 'link and compile args')
+        self.compile_args = self._filter_compile_args(args)
+        self.link_args = self._filter_link_args(args)
+
+    def _sanitize_version(self, out: str) -> str:
+        # disregard dummy "out" input as MinGW + MS-MPI doesn't work like other platforms
+        if not shutil.which('mpiexec'):
```
dcbaker, line 237: This seems... odd. The whole point of the ConfigToolDependency is to do this kind of stuff safely, and with machine file support. This really makes me think that using `ConfigToolDependency` as the parent is the right thing to do.

test cases/frameworks/17 mpi/main.cpp
```diff
@@ -1,11 +1,17 @@
...
+    int L;
+
+    char version[MPI_MAX_LIBRARY_VERSION_STRING];
+    MPI_Get_library_version(version, &L);
+    std::cout << version << std::endl;
+
     MPI::Init(argc, argv);
     if (!MPI::Is_initialized()) {
-        printf("MPI did not initialize!\n");
+        std::cerr << "MPI did not initialize!\n";
```
dcbaker, line 14: presumably you should be using `std::endl` here as well.

mesonbuild/dependencies/mpi.py
scivision: mpigcc is from IntelMPI as per the link in the source code just above. It's present on Linux.

mesonbuild/dependencies/mpi.py
scivision: OK I didn't refresh my understanding after the major refactoring you did some months ago. That would be better than abusing the factory function this way.

mesonbuild/dependencies/mpi.py
scivision: yes

test cases/frameworks/17 mpi/main.cpp
TheQwertiest: `std::endl` usage is considered bad practice though, since it forces a flush to output (it's more chars to type as well =)).
Not that it matters in this case (being an error message and all that).

scivision: I suppose here a flush could be beneficial, considering parallel tests, the flush would make it a bit more likely the stderr message is proximate to other same-test messages.

scivision:
This needs more work. There are some issues with MS-MPI that may need to be split into a separate PR. I will try to split off MS-MPI since it didn't really work anyway, and this PR fixes regressions unrelated to MS-MPI

scivision:
I simplified this PR. The compilers + MPI + OS cited above were just retested by me to work.
As noted above, there are future improvements that can be made, but for now this increases the number of MPI setups that work while simplifying and clarifying the code logic

nordmoen:
We just encountered a challenge that this PR should be able to fix. Is there anything that needs help with before being accepted?

dcbaker:
It needs to be rebased at the very least

rcoacci:
I just ran into this issue. My use case is using Intel compilers (`icc`/`icpc`) with OpenMPI. My solution for now was to "trick" meson by setting `I_MPI_CC=/usr/bin/mpicc` (which is OpenMPI).
I _really_ think one should not assume much with regards to the combination of compilers and MPI implementations. IMHO the best thing to do is to either use `MPICC` (and friends) env vars to detect which MPI implementation to use or rely on `/usr/bin/mpicc` (and friends), as those are _de facto_ standards. In particular, *DON'T* rely on compiler specific variables (I_MPI_CC), pkg-config, etc. Or at least use those as fallback only.
Also, if a particular compiler+MPI combination doesn't work, it should be the user's responsibility to set the variables to get the correct combination.

PR #7377, opened 2020-06-24
docs: Document string path building with examples

nirbheek:
Also document that line continuation didn't work before 0.50.

90c9b868b20b11bb089fc5e0c634d5ed76fea0cb fixed it.

### DIFF
--- a/docs/markdown/Syntax.md
+++ b/docs/markdown/Syntax.md
@@ -16,9 +16,10 @@ statements* and *includes*.
 Usually one Meson statement takes just one line. There is no way to
 have multiple statements on one line as in e.g. *C*. Function and
 method calls' argument lists can be split over multiple lines. Meson
-will autodetect this case and do the right thing. In other cases you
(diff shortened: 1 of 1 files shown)


docs/markdown/Syntax.md
```diff
@@ -136,6 +137,24 @@ str2 = 'xyz'
...
+```meson
+joined = '/usr/share' / 'projectname'    # => /usr/share/projectname
+joined = '/usr/local' / '/etc/name'      # => /etc/name
+
+joined = 'C:\\foo\\bar' / 'builddir'     # => C:/foo/bar/builddir
+joined = 'C:\\foo\\bar' / 'D:\\builddir' # => D:/builddir
+```
+
+Note that this is equivalent to using [`join_paths()`](Reference-manual.md#join_paths),
+which is deprecated.
```
xclaesse, line 156: There is no deprecation warning in https://mesonbuild.com/Reference-manual.html#join_paths, I don't think it should be considered deprecated, IMHO both are valid.

Issue #7371, opened 2020-06-23
LLVMDependencyConfigTool fails with exception if `llvm-config --shared-mode` returns in error


ydirson:
While configuring mesa for cross-compiling I hit the following stack trace (with 0.53.2):

```
;; This buffer is for text that is not saved, and for Lisp evaluation.
;; To create a file, visit it with C-x C-f and enter text in its buffer.

| Checking for function "strtof" : YES
| Checking for function "mkostemp" : YES
| Traceback (most recent call last):
|   File ".../mesonbuild/mesonmain.py", line 129, in run
|     return options.run_func(options)
|   File ".../mesonbuild/msetup.py", line 245, in run
|     app.generate()
|   File ".../mesonbuild/msetup.py", line 159, in generate
|     self._generate(env)
|   File ".../mesonbuild/msetup.py", line 192, in _generate
|     intr.run()
|   File ".../mesonbuild/interpreter.py", line 4168, in run
|     super().run()
|   File ".../mesonbuild/interpreterbase.py", line 412, in run
|     self.evaluate_codeblock(self.ast, start=1)
|   File ".../mesonbuild/interpreterbase.py", line 436, in evaluate_codeblock
|     raise e
|   File ".../mesonbuild/interpreterbase.py", line 430, in evaluate_codeblock
|     self.evaluate_statement(cur)
|   File ".../mesonbuild/interpreterbase.py", line 451, in evaluate_statement
|     return self.evaluate_if(cur)
|   File ".../mesonbuild/interpreterbase.py", line 530, in evaluate_if
|     self.evaluate_codeblock(i.block)
|   File ".../mesonbuild/interpreterbase.py", line 436, in evaluate_codeblock
|     raise e
|   File ".../mesonbuild/interpreterbase.py", line 430, in evaluate_codeblock
|     self.evaluate_statement(cur)
|   File ".../mesonbuild/interpreterbase.py", line 443, in evaluate_statement
|     return self.assignment(cur)
|   File ".../mesonbuild/interpreterbase.py", line 1064, in assignment
|     value = self.evaluate_statement(node.value)
|   File ".../mesonbuild/interpreterbase.py", line 441, in evaluate_statement
|     return self.function_call(cur)
|   File ".../mesonbuild/interpreterbase.py", line 788, in function_call
|     return func(node, posargs, kwargs)
|   File ".../mesonbuild/interpreterbase.py", line 285, in wrapped
|     return f(*wrapped_args, **wrapped_kwargs)
[...]

PR #7379, opened 2020-06-24
depenencies/llvm: Handle llvm-config --shared-mode failing

dcbaker:
Fixes: #7371
Fixes: #7878


dcbaker:
looks like in at least a few configurations this fails. That's probably OS related, I'll change this to allow skipping in CI

PR #7381, opened 2020-06-25
wrap: add user agent when downloading files

eli-schwartz:
This lets servers know when they're being used by meson. It also avoids issues where the Independent JPEG Group decided to ban the "Python-urllib" default user agent.

Fixes https://github.com/mesonbuild/libjpeg/issues/9

### DIFF
--- a/mesonbuild/wrap/wrap.py
+++ b/mesonbuild/wrap/wrap.py
@@ -29,6 +29,7 @@ import typing as T
 
 from pathlib import Path
 from . import WrapMode
+from .. import coredata
 from ..mesonlib import git, GIT, ProgressBar, MesonException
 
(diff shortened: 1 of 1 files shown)


scivision:
CMake does something roughly similar, a UserAgent of `curl/$curl_version` as from [CMake source](https://gitlab.kitware.com/yrHeTaTeJlb/cmake/-/commit/14aff4daf45a3cf0af59de60a464af9abe6a013c) and confirmed with CMake 

```
file(DOWNLOAD ${url} result.hml)
```

> curl/7.69.0

eli-schwartz:
Oh hmm, so there was also an older issue for that?

xclaesse:
LGTM

PR #7385, opened 2020-06-26
backend/vs: Fix C/C++ intellisense and "Go to File Action"

XNerv:
backend/vs: Fix C/C++ intellisense and "Go to File Action"

Intellisense and "Go to File Action" breaks when you try to edit/view a header
file that include a header of one of the target dependencies, but, the target
builds without any issue.

The problem occurs because the fields in the : [[ Project Property Pages  -->
C/C++  -->  General  -->  Additional Include Directories ]] and 
[[ Project Property Pages  --> C/C++  -->  General  -->  Preprocessor Definitions ]]
was not correctly set with the exported include directories and defines of the target
dependencies.

### DIFF
--- a/mesonbuild/backend/vs2010backend.py
+++ b/mesonbuild/backend/vs2010backend.py
@@ -983,9 +983,10 @@ class Vs2010Backend(backends.Backend):
                     else:
                         define = arg[2:]
                     # De-dup
-                    if define in file_defines[l]:
-                        file_defines[l].remove(define)
-                    file_defines[l].append(define)
+                    if define not in file_defines[l]:
+                        file_defines[l].append(define)
+                    if define not in target_defines: 
+                        target_defines.append(define)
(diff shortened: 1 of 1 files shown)


jpakkane:
Test failure seems relevant:

```
"D:\a\1\s\b 45d2573121\local arg test.sln" (default target) (1) ->
"D:\a\1\s\b 45d2573121\prog@exe.vcxproj" (default target) (2) ->
(ClCompile target) -> 
  d:\a\1\s\test cases\common\22 target arg\prog.cc(2): fatal error C1189: #error:  "Wrong local argument set" [D:\a\1\s\b 45d2573121\prog@exe.vcxproj]

    0 Warning(s)
    1 Error(s)
```

jpakkane:
Ping. Please fix the issue so this can be merged. Thanks.

nirbheek:
Fixed in https://github.com/mesonbuild/meson/pull/7729.

PR #7392, opened 2020-06-29
Use internal version of run_ctgt.py to support msi

### LINKED ISSUE #7390: CMake subprojects don't work when using Meson from .msi
When running Meson from the msi installation, it will fail a Cmake subproject build because it doesn't have `mesonbuild/cmake/data/preload.cmake` nor does it have `mesonbuild/cmake/data/run_ctgt.py`.

`mesonbuild/cmake/data/preload.cmake`:
This can just be added to the msi. Not too difficult and hopefully I can send in that PR soon.

`mesonbuild/cmake/data/run_ctgt.py`:
If this is simply added to the msi, it will violate the "no python required" aspect of the msi. This can be added as a hidden import to the msi, but then it needs to be called with `meson --internal run_ctgt ...`.

georgev93:
Fixes #7390


mensinda:
Did you test if the generated msi from this PR works for you? I am on Linux so testing this myself is a bit tricky.

georgev93:
Correct. I have it building a project with a cmake subproject which contains custom build targets. Everything seems to check out!

PR #7396, opened 2020-06-30
Cleaned up Reference docs for consistency [skip ci]

TheQwertiest:
The inconsistency was bugging me forever =)
Strange `*introduced in v0.xx*` placement was confusing as well.

Cleaned up using the following rules:
- All changes that appeared in certain meson versions were marked with `*(since 0.xx.x)*` (instead of `*(added in 0.xx.x)*`, `*in 0.xx*` and etc).
- Kwargs and arguments has the following formatting:
```
- `kwarg1_name` *(since 0.xx.x)*: kwarg1_description.
- `kwarg2_name`: *(since 0.xx.x)* **(deprecated)** kwarg2_description.
```

- Changes that appeared starting from a certain version have the following formatting:
```
 *(since 0.xx.x)* Description of changes.
```

- Methods use the following formatting:
```
### Method

`method(args)`

*(since 0.xx.x)*

Method description
```

### DIFF
--- a/docs/markdown/Reference-manual.md
+++ b/docs/markdown/Reference-manual.md
@@ -15,19 +15,19 @@ afterwards](#returned-objects).
 Adds the positional arguments to the compiler command line. This
 function has two keyword arguments:
 
-- `language` specifies the language(s) that the arguments should be
+- `language`: specifies the language(s) that the arguments should be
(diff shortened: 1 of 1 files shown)


jpakkane:
Good stuff, thanks.

PR #7397, opened 2020-06-30
Replaced `ninja` with `meson` [skip ci]

TheQwertiest:
Continuation of #7328


eli-schwartz:
Please revert this commit as it's incredibly misleading and deceptive to users to tell them to use `meson compile <target>`.

```
01:29 PM <ScottHDev> But I get : `ERROR: Neither directory contains a build file meson.build.` Seems like meson parses the two arguments as directory names.
01:30 PM <ScottHDev> (I'm following this document page https://mesonbuild.com/Localisation.html)
```

This definitely does not work robustly, and if you're using common but mildly old versions of meson you get completely nonsensical error messages with virtually no hope of debugging the cause, and documentation that tries to avoid telling you what you can use instead.

gregsskyles:
Does the documentation need the usual `since version 0.54.0` (example only, just guessing which version is applicable for this issue) disclaimer?

PR #7412, opened 2020-07-04
doc: Improve documentation of [provide] section in wrap files

xclaesse:
Explicitly document the behaviour of dependency('foo-1.0', required:
false).

### DIFF
--- a/docs/markdown/Wrap-dependency-system-manual.md
+++ b/docs/markdown/Wrap-dependency-system-manual.md
@@ -136,25 +136,60 @@ thousands of lines of code. Once you have a working build definition,
 just zip up the Meson build files (and others you have changed) and
 put them somewhere where you can download them.
 
-Meson build patches are only supported for wrap-file mode. When using
-wrap-git, the repository must contain all Meson build definitions.
+Prior to *0.55.0* Meson build patches were only supported for wrap-file mode.
+When using wrap-git, the repository must contain all Meson build definitions.
+Since *0.55.0* Meson build patches are supported for any wrap modes, including
+wrap-git.
 
 ## `provide` section
 
 *Since *0.55.0*
 
 Wrap files can define the dependencies it provides in the `[provide]` section.
-When a wrap file provides the dependency `foo` any call do `dependency('foo')`
-will automatically fallback to that subproject even if no `fallback` keyword
-argument is given. It is recommended for subprojects to call
-`meson.override_dependency('foo', foo_dep)`, dependency name can then be added into
-the special `dependency_names` entry which takes comma separated list of dependency
-names. For backward compatibility with subprojects that does not call
-`meson.override_dependency()`, the variable name can be provided in the wrap file
-with entries in the format `dependency_name = variable_name`,
-where `dependency_name` usually match the corresponding pkg-config name and
-`variable_name` is the name of a variable defined in the subproject that should
-be returned for that dependency.
+
+```
+[provide]
+dependency_names = foo-1.0
+```
+
+When a wrap file provides the dependency `foo-1.0`, as above, any call to
+`dependency('foo-1.0')` will automatically fallback to that subproject even if
+no `fallback` keyword argument is given. A wrap file named `foo.wrap` implicitly
+provides the dependency name `foo` even when the `[provide]` section is missing.
+
+Optional dependencies, like `dependency('foo-1.0', required: get_option('foo_opt'))`
+where `foo_opt` is a feature option set to `auto`, will not fallback to the
+subproject defined in the wrap file, for 2 reasons:
+- It allows for looking the dependency in other ways first, for example using
+  `cc.find_library('foo')`, and only fallback if that fails:
+```
+# this won't use fallback defined in foo.wrap
+foo_dep = dependency('foo-1.0', required: false)
+if not foo_dep.found()
+  foo_dep = cc.find_library('foo', has_headers: 'foo.h', required: false)
+  if not foo_dep.found()
+    # This will use the fallback
+    foo_dep = dependency('foo-1.0')
+    # or
+    foo_dep = dependency('foo-1.0', required: false, fallback: 'foo')
+  endif
+endif
+```
+- Sometimes not-found dependency is preferable to a fallback when the feature is
+  not explicitly requested by the user. In that case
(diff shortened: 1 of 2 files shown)


docs/markdown/Wrap-dependency-system-manual.md
```diff
@@ -136,25 +136,55 @@ thousands of lines of code. Once you have a working build definition,
...
+```
+- Sometimes not-found dependency is preferable to a fallback when the feature is
+  not explicitly requested by the user. In that case
+  `dependency('foo-1.0', required: get_option('foo_opt'))` will only fallback
+  when the user sets `foo_opt` to `enabled` instead of `auto`.
+
+This mechanism assumes the subproject calls `meson.override_dependency('foo-1.0', foo_dep)`
+so Meson knows which dependency object should be used as fallback.
+Since that method was introduced in version *0.54.0*, maybe projects does not yet
+make use of it. In that case the variable name can be provided in the wrap file
```
randy408, line 186: Perhaps "projects may not make use of it"?

xclaesse: oh it's a typo, I meant "**many** projects does not..."

xclaesse: fixed.

textshell: Maybe usesomething like:
> Since that method was introduced in version *0.54.0*, as a transitional aid for projects that do not yet
make use of it.

docs/markdown/Wrap-dependency-system-manual.md
```diff
@@ -136,25 +136,55 @@ thousands of lines of code. Once you have a working build definition,
...
-names. For backward compatibility with subprojects that does not call
-`meson.override_dependency()`, the variable name can be provided in the wrap file
-with entries in the format `dependency_name = variable_name`,
-where `dependency_name` usually match the corresponding pkg-config name and
-`variable_name` is the name of a variable defined in the subproject that should
-be returned for that dependency.
+
+```
+[provide]
+dependency_names = foo-1.0
```
textshell, line 152: I think using examples with "foo-1.0" makes it harder to concentrate on the provided name semantics since a lot of pkgconfig library names don't follow that scheme.

xclaesse: Hm, I made that change specifically because most pkgconfig libraries follows that scheme (glib-2.0, gstreamer-1.0, gtk+-3.0, ...) but that could be a bias because I'm from GNOME world?

xclaesse: Looking at /usr/lib/x86_64-linux-gnu/pkgconfig/ you're actually right, most don't... I'm biaised :D

But still, if your pkgconfig file is named `foo` and you have `foo.wrap` you don't even need the `[provide]` section, it automatically fallback to a wrap with the same name as your dependency.

docs/markdown/Wrap-dependency-system-manual.md
tp-m: "Many projects ... does" still needs fixing :)

I think this section would benefit from some rewording/reflow though, along the lines of textshell's suggestion above. The text will probably stay like that for quite some time, so should still make sense when someone reads it in a year or two.

docs/markdown/Wrap-dependency-system-manual.md
textshell: maybe use foo-glib as example dependency that would need to be added explicitly?

docs/markdown/Wrap-dependency-system-manual.md
```diff
@@ -136,25 +136,55 @@ thousands of lines of code. Once you have a working build definition,
...
+    foo_dep = dependency('foo-1.0')
+    # or
+    foo_dep = dependency('foo-1.0', required: false, fallback: 'foo')
+  endif
+endif
+```
+- Sometimes not-found dependency is preferable to a fallback when the feature is
+  not explicitly requested by the user. In that case
+  `dependency('foo-1.0', required: get_option('foo_opt'))` will only fallback
+  when the user sets `foo_opt` to `enabled` instead of `auto`.
```
jpakkane, line 181: Perhaps this should mention that if the desired behavior is to use it even for auto, you can do something like:

    dependency('foo-1.0', required: get_option('foo_opt') != 'disabled')

xclaesse: I don't think that would work because when option is set to disabled that would pass `required: false` which is not the same thing.

I did not mention that part clearly enough, but see the example code with find_library above, it does: `dependency('foo-1.0', required: false, fallback: 'foo')`. When you specify the fallback explicitly it will use it even for optional dependencies. So with an `auto` option you would do: `dependency('foo-1.0', required: get_option('foo_opt'), fallback: 'foo')`. I think it would be better to instead add support for that syntax: `dependency('foo-1.0', required: get_option('foo_opt'), fallback: true)`, what do you think?

xclaesse: Or you can code it like that:
```

if not get_option('foo_opt').disabled()
  foo_dep = dependency('foo-1.0', required: get_option('foo_opt'))
else
  foo_dep = dependency('', required: false)
endif
```

xclaesse: Note also that with `dependency('foo-1.0', required: get_option('foo_opt') != 'disabled')` when option is `auto` and the fallback subproject fails it would abort instead of returning not-found.

docs/markdown/Wrap-dependency-system-manual.md
xclaesse: you're right, updated.

docs/markdown/Wrap-dependency-system-manual.md
xclaesse: Added explicit mention of `fallback: 'foo'` trick for now.

docs/markdown/Wrap-dependency-system-manual.md
xclaesse: Hm, I've been using foo-1.0 across the whole text, I don't think foo-glib makes much difference. IMHO `foo-1.0` is common enough for a dependency name that does not match the project name.

PR #7418, opened 2020-07-06
CUDA fixes

SoapGentoo:
@dcbaker @nirbheek @jpakkane

### DIFF
--- a/mesonbuild/compilers/cuda.py
+++ b/mesonbuild/compilers/cuda.py
@@ -46,14 +46,33 @@ class CudaCompiler(Compiler):
 
     @classmethod
     def _to_host_flags(cls, flags, phase='compiler'):
-        return list(map(partial(cls._to_host_flag, phase=phase), flags))
+        # since version 6, GCC is *VERY* allergic to passing
+        # '-isystem/usr/include' on the command line, see for instance
+        #
+        #   https://gcc.gnu.org/bugzilla/show_bug.cgi?id=70129#c1
+        #
+        # and while nvcc doesn't care about this, the fact that
+        # it happily passes '-isystem/usr/include' to GCC for
+        # compiling host code means we need to filter it.
+        FILTER = frozenset(['-isystem/usr/include']) if phase == 'compiler' else frozenset()
+        prefiltered_flags = filter(lambda x: x not in FILTER, flags)
+        return list(map(partial(cls._to_host_flag, phase=phase), prefiltered_flags))
 
     @classmethod
     def _to_host_flag(cls, flag, phase):
         if not flag[0] in ['-', '/'] or flag[:2] in cls._universal_flags[phase]:
             return flag
 
-        return '-X{}={}'.format(phase, flag)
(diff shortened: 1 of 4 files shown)


armintoepfer:
Finally no more `sed` hacking

mesonbuild/compilers/cuda.py
```diff
@@ -46,18 +46,40 @@ def __init__(self, exelist, version, for_machine: MachineChoice,
...
 
-        return '-X{}={}'.format(phase, flag)
+        if phase == 'compiler':
+            TRANSLATE_ARGS = {}
+        else:
+            assert phase == 'linker'
+            # nvcc passes '-Xlinker=-pthread' straight to ld, which
+            # explodes without remapping to '-Xlinker=-lpthread', as
+            # -pthread is purely a GCC flag.
+            TRANSLATE_ARGS = {'-pthread': '-lpthread'}
```
nirbheek, line 73: This seems slightly overengineered :) why not do:

```python
if phase == 'linker' and flag == '-pthread':
    flag = '-lpthread'
```

mesonbuild/compilers/cuda.py
```diff
@@ -46,18 +46,40 @@ def __init__(self, exelist, version, for_machine: MachineChoice,
...
+        # since version 6, GCC is *VERY* allergic to passing
+        # '-isystem/usr/include' on the command line, see for instance
+        #
+        #   https://gcc.gnu.org/bugzilla/show_bug.cgi?id=70129#c1
+        #
+        # and while nvcc doesn't care about this, the fact that
+        # it happily passes '-isystem/usr/include' to GCC for
+        # compiling host code means we need to filter it.
+        FILTER = frozenset(['-isystem/usr/include']) if phase == 'compiler' else frozenset()
+        prefiltered_flags = filter(lambda x: x not in FILTER, flags)
```
nirbheek, line 58: We discussed this on IRC too, but if `-isystem` is broken with GCC, maybe we should never emit it inside the Boost dependency when using GCC? It seems wrong that we're filtering a flag that we ourselves emit elsewhere.

mesonbuild/compilers/cuda.py
SoapGentoo: I'll address this

mesonbuild/compilers/cuda.py
SoapGentoo: The problem is, the `-I -> -isystem` rewrite happens at a later phase, when the boost dependency doesn't have control over it anymore. and only `-isystem/usr/include` breaks GCC, not `-I/usr/include`

jpakkane:
What's the status on this one? I don't have sufficient experience with Cuda to do a meaningful review.

SoapGentoo:
Let's close this for now. We need to overhaul this in a number of ways:
- the CUDA compiler doesn't honour most normal settings (ndebug etc)
- we need to be more intelligent with filtering the system include paths (which requires running the host compiler)
- we need to add support for Ampere

PR #7422, opened 2020-07-08
Solaris fixes revisited

alanc:
Fixes issues raised by @dcbaker  during review of #7370 and an additional issue (linker version info showing gcc's instead) noticed while testing.

Tested on both Solaris 11.4 & OpenIndiana 2020.04.

### DIFF
--- a/mesonbuild/environment.py
+++ b/mesonbuild/environment.py
@@ -930,9 +930,15 @@ class Environment:
                 cls = GnuBFDDynamicLinker
             linker = cls(compiler, for_machine, comp_class.LINKER_PREFIX, override, version=v)
         elif 'Solaris' in e or 'Solaris' in o:
+            for line in (o+e).split('\n'):
+                if 'ld: Software Generation Utilities' in line:
+                    v = line.split(':')[2].lstrip()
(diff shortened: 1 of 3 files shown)


jpakkane:
How badly do you want this for 0.55.0? The release freeze is ongoing and this is on the cusp of acceptance (if it touched any non-solaris code paths it would not be accepted, for example).

alanc:
As the Solaris package maintainer, I'm okay with waiting - I've applied [the patches we need to 0.54.3 in our packages],(https://github.com/oracle/solaris-userland/tree/master/components/python/meson/patches) and 0.55 should mostly just work for us.  Those on the open source side of the Solaris fence, such as OpenIndiana distros may find this more urgent, since they'll be getting ninja files generated with invalid flags for their linker.  (Though before my patches they instead would have seen meson dying with an exception due to not having implemented `get_pie_args`, so presumably aren't testing 0.55 yet, though they do seem to be [shipping 0.54.2 in OpenIndiana](https://github.com/OpenIndiana/oi-userland/tree/oi/hipster/components/developer/meson).)

PR #7424, opened 2020-07-08
coredata: Make warning_level per subproject builtin option

xclaesse:


xclaesse:
Failing unit test is caused by https://github.com/mesonbuild/meson/pull/7492

nirbheek:
lgtm, this is necessary for being able to use subprojects without spammy warnings. Same reason as `werror`.

PR #7425, opened 2020-07-09
string: add substring method

dabrain34:
This method aims to offer a simple way to 'substring'
an existing string with start and end values.


docs/markdown/Reference-manual.md
```diff
@@ -1977,6 +1977,9 @@ are immutable, all operations return their results as a new string.
 - `startswith(string)`: returns true if string starts with the string
   specified as the argument
 
+- `substring(start,end)`: returns a substring specified from start to end.
```
TheQwertiest, line 1980:
```suggestion
- `substring(start,end)` *(since 0.56.0)*: returns a substring specified from start to end.
```

docs/markdown/Reference-manual.md
```diff
@@ -1977,6 +1977,9 @@ are immutable, all operations return their results as a new string.
...
+  Both values are optional as 'foobar'.substring() returns 'foobar'.
```
TheQwertiest, line 1981:
```suggestion
  Both `start` and `end` arguments are optional, so, for example, `'foobar'.substring()` will return `'foobar'`.
```

test cases/common/38 string operations/meson.build
```diff
@@ -101,3 +101,11 @@ assert('\\\\n' == bs_bs_n, 'Four backslash broken before n')
 assert('\\\\\n' == bs_bs_nl, 'Five backslash broken before n')
 assert('\\\\' == bs_bs, 'Double-backslash broken')
 assert('\\' == bs, 'Backslash broken')
+
+mysubstring='foobarbaz'
+assert(mysubstring.substring() == 'foobarbaz', 'substring is broken')
+assert(mysubstring.substring(0) == 'foobarbaz', 'substring is broken')
+assert(mysubstring.substring(1) == 'oobarbaz', 'substring is broken')
+assert(mysubstring.substring(-5) == 'foobarbaz', 'substring is broken')
```
tp-m, line 109: What's the rationale for allowing negative `start` values? Why not just error out?

test cases/common/38 string operations/meson.build
```diff
@@ -101,3 +101,11 @@ assert('\\\\n' == bs_bs_n, 'Four backslash broken before n')
...
+assert(mysubstring.substring(1, 4) == 'oob', 'substring is broken')
+assert(mysubstring.substring(1,-5) == 'oobarbaz', 'substring is broken')
```
tp-m, line 111: I would expect an `end` value of `-5` to mean `5 characters from the end`, like python would do for `mysubstring[1:-5]` (which would return `oob` here). Otherwise it should just be a hard error imho.

docs/markdown/Syntax.md
```diff
@@ -220,6 +220,14 @@ is_x86 = target.startswith('x86') # boolean value 'true'
 is_bsd = target.to_lower().endswith('bsd') # boolean value 'true'
 ```
 
+#### .substring()
+
+```meson
+target = 'x86_FreeBSD'
+platform = target.substring(0, 3) # prefix string value 'x86'
+system = target.substring(4) # suffix string value 'FreeBSD'
+```
```
tp-m, line 229: Somewhat unfortunate example, because here `split('_')` would arguably be a better fit, but doesn't matter I guess :)

xclaesse:
I think it would be useful to allow negative index like python, to mean relative from the end. 
`'libfoo.so'.substring(3:-3)`would give `foo`.

test cases/common/38 string operations/meson.build
dabrain34: because i wanted to avoid errors and I followed javascript sbstring doc (not sure its a good argument :p)

@xclaesse I can implement, so negative end value means pos = len - value right ?

docs/markdown/Syntax.md
dabrain34: That's why I implemented this because gstdockingstation.split('gst') is giving you ockination :)

dabrain34: I agree but as a new manner to do it and be sure of of the result :)

test cases/common/38 string operations/meson.build
dabrain34: done

mesonbuild/interpreterbase.py
```diff
@@ -1026,6 +1026,28 @@ def string_method_call(self, obj: str, method_name: str, posargs: T.List[TYPE_nv
...
             return mesonlib.version_compare(obj, cmpr)
+        elif method_name == 'substring':
+            if len(posargs) > 2:
+                raise InterpreterException('substring() takes maximum two arguments.')
+            start = 0
+            end = len(obj)
+            if len (posargs) > 0:
+              start = int(posargs[0])
+            if len (posargs) > 1:
+                end = int(posargs[1])
```
xclaesse, line 1037: You should check type of args:
```
if not isinstance(posargs[0], int):
    raise InterpreterException()
```

mesonbuild/interpreterbase.py
```diff
@@ -1026,6 +1026,28 @@ def string_method_call(self, obj: str, method_name: str, posargs: T.List[TYPE_nv
...
+            if start < 0:
+                start = 0
```
xclaesse, line 1039: start could also be relative to end.

mesonbuild/interpreterbase.py
```diff
@@ -1026,6 +1026,28 @@ def string_method_call(self, obj: str, method_name: str, posargs: T.List[TYPE_nv
...
+            if end < 0:
+                if abs(end) <= len(obj):
+                    end = len(obj) - abs(end)
+                else:
+                    end = 0
```
xclaesse, line 1044: you don't need to implement it yourself, python will do that logic for you already. tbh you don't need to do any checks on start/end values, python does the right thing for any value ;-)

mesonbuild/interpreterbase.py
```diff
@@ -1026,6 +1026,28 @@ def string_method_call(self, obj: str, method_name: str, posargs: T.List[TYPE_nv
...
+            # swap if start > stop
+            if start > end:
+                s = start
+                start = end
+                end = s
```
xclaesse, line 1049: Why swapping? IMHO `'abc'.substring(1, 0)` should return empty string, like python does.

run_unittests.py
```diff
@@ -2015,6 +2015,10 @@ def test_libdir_must_be_inside_prefix(self):
         else:
             self.assertRaises(subprocess.CalledProcessError, self.setconf, '-Dlibdir=/opt', False)
 
+    def test_string_operations(self):
+      testdir = os.path.join(self.common_test_dir, '38 string operations')
+      self.init(testdir)
```
xclaesse, line 2020: you don't need to write unit test here, everything in common/ will be configured/build by default.

xclaesse: btw, you can just append your tests to `common/38 string operations` where we already test other string methods. It's faster on the CI, it saves meson invocations on hundred small tests.

mesonbuild/interpreterbase.py
dabrain34: I took it from javascript as I should have taken from python ;)

run_unittests.py
dabrain34: what is the command to just run this unit test, I was unable to find it.

xclaesse: I personally have a `~/.local/bin/meson` being a symlink to `meson.py` from source. Then I can just use meson command in test dirs. Dunno if there are better ways.

mesonbuild/interpreterbase.py
xclaesse: ah, didn't know JS does that. I guess it depends on the language you're used to... I'm certainly biased toward python, maybe that's not the behaviour others would expect?

run_unittests.py
TheQwertiest: `./meson.py ./run_unit_tests.py test_string_operations`

dabrain34: yes i removed it at the end ... and thx @xavierclaessens I was dumb as I can run it directly in the folder ... :stuck_out_tongue:

test cases/common/38 string operations/meson.build
```diff
@@ -101,3 +101,11 @@ assert('\\\\n' == bs_bs_n, 'Four backslash broken before n')
...
 assert('\\\\' == bs_bs, 'Double-backslash broken')
 assert('\\' == bs, 'Backslash broken')
+
+mysubstring='foobarbaz'
+assert(mysubstring.substring() == 'foobarbaz', 'substring is broken')
+assert(mysubstring.substring(0) == 'foobarbaz', 'substring is broken')
+assert(mysubstring.substring(1) == 'oobarbaz', 'substring is broken')
+assert(mysubstring.substring(-5) == 'arbaz', 'substring is broken')
+assert(mysubstring.substring(1, 4) == 'oob', 'substring is broken')
+assert(mysubstring.substring(1,-5) == 'oob', 'substring is broken')
```
xclaesse, line 111: I think it could also have a few tests with out of range values: (1,0), (0,100), etc.

docs/markdown/Syntax.md
xclaesse, line 229: Since 0.56

xclaesse: Could also document behaviour of negative and out-of-range values

xclaesse: You forgot that:

> Could also document behaviour of negative and out-of-range values

docs/markdown/Syntax.md
```diff
@@ -220,6 +220,26 @@ is_x86 = target.startswith('x86') # boolean value 'true'
...
+
+```meson
+# Similar to the Python str[start:end] syntax
+target = 'x86_FreeBSD'
+platform = target.substring(0, 3) # prefix string value 'x86'
+system = target.substring(4) # suffix string value 'FreeBSD'
+```
+
+The method accepts negative values where a negative `start` always returns
+empty string and a negative `end` returns from `start` to `len(string) - end`.
```
xclaesse, line 235: Negative start also means relative to end, in python `'abc'[-2:-1]` gives `b`.

dabrain34: I did not find any clear documentation so I was not sure about this behavior, thanks for the clarification. I hope the doc is now clear enough

xclaesse:
Looks all good to me know, thanks! LGTM after release freeze.

docs/markdown/Syntax.md
```diff
@@ -220,6 +220,26 @@ is_x86 = target.startswith('x86') # boolean value 'true'
 is_bsd = target.to_lower().endswith('bsd') # boolean value 'true'
 ```
 
+#### .substring()
+
+Since 0.56.0, you can extract a substring from a string.
```
TheQwertiest, line 225:
```suggestion
*(since 0.56.0)*

Extracts a substring from a string.
```

xclaesse:
@dabrain34 could you rebase to see if CI goes green please?

xclaesse:
Since substring() has been proposed by @jpakkane on IRC, I'm assuming it's an approved feature. Merging now, thanks.

PR #7427, opened 2020-07-09
summary: Wrap lines when printing lists

xclaesse:
When a list_sep is provided (e.g. ', ') all items are printed on the
same line, which gets ugly on very long lists (e.g. list of plugins
enabled).


dabrain34:
dont you think it should be configurable instead in the summary call ?

xclaesse:
I'm not sure. I don't like hard coding an arbitrary number coming from another century, but making it configurable means that if 2 projects picks different value you'll have inconsistent output. 80 chars still makes sense today, it's the default size of most terminals (at least on Linux) when the window is not maximized.

We can always make it configurable later, maybe for example choosing the max value of all summary() calls.

dabrain34:
yes you're right. It would make sense to configure it on meson command line livel, otherwise it does not make any sense, inside the projects.(I dont like neither to hardcode values like that)

xclaesse:
Updated to use `shutil.get_terminal_size().columns` instead of hard coding 80 chars.

tp-m:
> Updated to use `shutil.get_terminal_size().columns` instead of hard coding 80 chars.

perfect! :heart:

PR #7428, opened 2020-07-09
Handle add_languages(native:) in introspector

jon-turney:
This should have been part of #6532

### DIFF
--- a/mesonbuild/ast/introspection.py
+++ b/mesonbuild/ast/introspection.py
@@ -123,7 +123,8 @@ class IntrospectionInterpreter(AstInterpreter):
         options = {k: v for k, v in self.environment.meson_options.host[''].items() if k.startswith('backend_')}
 
         self.coredata.set_options(options)
-        self.func_add_languages(None, proj_langs, None)
+        self._add_languages(proj_langs, MachineChoice.HOST)
+        self._add_languages(proj_langs, MachineChoice.BUILD)
 
     def do_subproject(self, dirname: str) -> None:
         subproject_dir_abs = os.path.join(self.environment.get_source_dir(), self.subproject_dir)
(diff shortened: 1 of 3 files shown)


mensinda:
This should have a unit test if possible (just search for `introspect` in `run_unittests.py` for templates).

jpakkane:
Ping.

jon-turney:
> This should have a unit test if possible

Reading https://github.com/mesonbuild/meson/pull/6512#issuecomment-599989205, I think the test needed here is to verify that e.g. c_args/build.c_args options are present in introspection output appropriately.

mesonbuild/ast/introspection.py
```diff
@@ -137,17 +138,24 @@ def do_subproject(self, dirname: str) -> None:
...
-                if isinstance(lang, StringNode):
-                    assert isinstance(lang.value, str)
-                    lang = lang.value
-                if not isinstance(lang, str):
-                    continue
-                lang = lang.lower()
-                if lang not in self.coredata.compilers[for_machine]:
-                    self.environment.detect_compiler_for(lang, for_machine)
+        if 'native' in kwargs:
+            native = kwargs.get('native', False)
```
mensinda, line 142: You should `flatten_kwargs` the `kwargs` first.

jon-turney: Done

mensinda (approved):
LGTM once the CI is green

mesonbuild/ast/introspection.py
jon-turney: ... and it looks like it was never working correctly due to that, so I had the expected test result upside-down?

Ericson2314 (approved):

PR #7429, opened 2020-07-10
[RFC] Add --enable/--disable syntax sugar for features

haasn:
Since the default 'feature' type makes the special names 'enabled' and
'disabled' significant, we have reason to justify hard-coding
--enable-foo and --disable-foo as -Dfoo=enabled and -Dfoo=disabled
respectively.

At the risk of increasing API inconsistency, I think this change is a
net gain because it:

- avoids breaking muscle memory or 'expectation' for new users coming to
  meson from autotools etc., thus increasing ease of transitioning.

- removes one of the barriers for meson's adoption in the real world,
  where persuasive individuals do sadly end up expressing distate of its
  "ugly cmake-style syntax".

Due to existing built-in options like --libdir -> -Dlibdir, there *is*
some precedent to supplying a nicer "front-end" options where we can.

TODO:
- [ ] Update documentation


tp-m:
:heart: 

This is one of the things autotools got right, it's just much nicer UI.

Should the same be done for boolean properties?

Should we have `--with-*` for other types of properties?

TheQwertiest:
There is was one (potentially huge) problem: back-compat breakage if someone has a `--enable-foo` for option name.

xclaesse:
I like the feature. My comments:
- You probably need to replace `-` with `_` since we kind of implicitly encourage using underscore in options, for example we have a pretty common `gtk_doc` option that would probably be set using `--enable-gtk-doc`. But then projects who use dash it won't work for them.
- It means we won't be allowed to add any builtin option that start with `enable_` that are not of type feature... I think that's too corner case to happen (famous last words).
- We won't be allowed to add any --enable-* option to other meson commands, you probably limit that rewrite if args[0] is "setup" or "configure". It also seems really unlikely to happen.

xclaesse:
> There is one (potentially huge) problem: back-compat breakage if someone has a `--enable-foo` for option name.

That would be asking for problems... Currently they would write `-D--enable-foo=value`... add that to the too unlikely to care list :D

TheQwertiest:
> Currently they would write -D--enable-foo=value

Oh right, I've forgotten that we use `-DOPT_NAME` and not `--OPT_NAME` for assigning user option values.

xclaesse:
oh, and we probably should show this in --help message somehow?

xclaesse:
> Should the same be done for boolean properties?

I don't think we can't have --foo rewrote as --Dfoo=true, that would conflict if a project has an option with the same name as any of our command args. For example if a project has an option called `wipe` or `version`.

> Should we have `--with-*` for other types of properties?

You mean a generic `--with-foo=bar` to `-Dfoo=bar`?

tp-m:
>     * You probably need to replace `-` with `_` since we kind of implicitly encourage using underscore in options, for example we have a pretty common `gtk_doc` option that would probably be set using `--enable-gtk-doc`. But then projects who use dash it won't work for them.

I think meson should normalise this anyway, and treat dashes/underscores as equivalent.

`--enable-gtk_doc` is just wrong imho :)

haasn:
> Should we have `--with-*` for other types of properties?

One semi-interesting aspect of doing this is that due to the egalitarian nature of this syntactical translation, we'd now have `--with-libdir=/path` as an alternative to the already existing sugar `--libdir=/path`, which is perhaps the point where the redundancy gets confusing (three variants).

xclaesse:
> I think meson should normalise this anyway, and treat dashes/underscores as equivalent.

I agree, it should be done in the PR as well to avoid --enable-gtk_doc indeed.

haasn:
> > I think meson should normalise this anyway, and treat dashes/underscores as equivalent.
> 
> I agree, it should be done in the PR as well to avoid --enable-gtk_doc indeed.

I suppose that means we aren't worried about existing projects that have both `foo_bar` and `foo-bar` as separate options? (But that probably also falls under 'obnoxious enough to be worth breaking')

If I understand you correctly, -Dfoo-bar should also normalize to -Dfoo_bar?

haasn:
Updated with `--help` syntax added (by adding special 'dummy' arguments `enable-option` and `disable-option`, which won't actually get parsed, merely for their help output), and documentation.

I'm not sure if I should put 'since version 0.55' or 'since version 0.56' in the documentation (or omit it altogether), since I see you're in the process of making that release. Do you think this would be included?

xclaesse:
> I suppose that means we aren't worried about existing projects that have both `foo_bar` and `foo-bar` as separate options? (But that probably also falls under 'obnoxious enough to be worth breaking')

IMHO, yes.
 
> If I understand you correctly, -Dfoo-bar should also normalize to -Dfoo_bar?

IMHO, yes. Probably even `get_option('foo-bar')` should work for option `foo_bar`. Actually it should be normalized from the beginning in optinterpreter.py to parse `option('foo-bar')` immediately as `foo_bar`.

xclaesse:
> I'm not sure if I should put 'since version 0.55' or 'since version 0.56' in the documentation (or omit it altogether), since I see you're in the process of making that release. Do you think this would be included?

Definitely 0.56, we are at rc2 already for 0.55, nothing else than major regression will be merged until final release (Sunday).

xclaesse:
Wow, defining an option twice is not even an error currently... optinterpreter should check that.
```
option('foo-bar', type: 'boolean')
option('foo-bar', type: 'boolean') # This should already be an error, but it's not
option('foo_bar', type: 'boolean') # This should become an error with this PR
```

haasn:
There is one interesting special case that, as a result of normalization, the previously legal option `c-foo` will now become illegal (due to the `c_` prefix being reserved).

xclaesse:
Hmm, it's not only `c_` that is currently reserved, it's `<any lang>_` as well as `b_` and `backend_`. See `is_invalid_name()` in optinterpreter.py. Those reserved "namespace" have always been a bit annoying, adding any language breaks backward compat with projects who have options that start with the same name.

xclaesse:
See how a simple feature becomes something complicated with backward compat considerations... Now I remember why we never added --enable-foo before :D

haasn:
> See how a simple feature becomes something complicated with backward compat considerations... Now I remember why we never added --enable-foo before :D

There's nothing saying we need `-` -> `_` normalization everywhere just to support `--enable-foo`. As far as I'm concerned those are separate changes that I'm being asked to merge into a single PR.

xclaesse:
I think that's merge blocker for consistency, yes. As said above, you would have to do --enable-gtk_doc otherwise which is stupid. But if you just do s/-/_/ then it's not working for projects who use "foo-bar", so you have to go all the way down to making - and _ equivalent everywhere...

haasn:
> Those reserved "namespace" have always been a bit annoying, adding any language breaks backward compat with projects who have options that start with the same name.

OTOH, given that the motivation behind this normalization is the fact that you already recommend using underscores instead of dashes, presumably the risk of somebody using `reservedprefix-foo` as an existing option is already low.

Side note, a reason *for* this normalization is that it actually *increases* consistency across projects. I *can* actually name projects (including my own) that use `foo-bar` style option naming instead of `foo_bar`, which after this change will no longer "surprise" seasoned users that expected them to be using `foo_bar` instead.

tp-m:
Maybe we could issue a warning in the upcoming release for these corner case option names, to get a sense of whether it's actually an issue or not? (seems very unlikely to me, but who knows)

haasn:
I added a proof of concept of what a "normalize all options" commit could look like. Not sure I'm happy with the result, but it's something at the very least.

PR #543, opened 2016-05-23
compilers: Add the buildtype args for checks that perform linking

nirbheek:
This is required for checking for compiler checks that involve linking to a static library with MSVC. Without this, MSVC errors out since no CRT is specified.

These also don't hurt with other compilers since they simply force the check to use debugging mode which should not do optimizations and change the result of the check.

PR #7432, opened 2020-07-11
Fix native builds on Windows ARM64 machines

nirbheek:
I made the mistake of always selecting the debug CRT for compiler checks on Windows 4 years ago: https://github.com/mesonbuild/meson/pull/543 https://github.com/mesonbuild/meson/pull/614

The idea was to always build the tests with debugging enabled so that the compiler doesn't optimize the tests away. But we stopped doing that a while ago, and also the debug CRT has no relation to that.

We should select the CRT in the same way that we do for building targets: based on the options.

On Windows ARM64, the debug CRT for ARM64 isn't always available, and the release CRT is available only after installing the runtime package. Without this, we will always try to pick the debug CRT even when --buildtype=debugoptimized or release.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -13,6 +13,7 @@ variables:
 jobs:
 
 - job: vs2017
+  timeoutInMinutes: 120
   pool:
     vmImage: VS2017-Win2016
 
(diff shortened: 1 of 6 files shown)


nirbheek:
CC: @seungha-yang

seungha-yang:
I can confirm this will fix my [gst-build](https://gitlab.freedesktop.org/gstreamer/gst-build) build issue (native ARM64 with Surface Pro X + Visual Studio 2019). Thanks @nirbheek

Issue #7400, opened 2020-07-01
Windows 10 python3 installed from Microsoft store not found by meson


tksuoran:
**Describe the bug**
I have freshly installed Windows 10.
I have python3 installed from Microsoft store. _(edit: tested both python 3.7 and python 3.8)_
I have meson installed from latest release / .msi _(edit: also tested with meson installed with pip in python virtual environment)_
In my meson.build I have `python3 = import('python').find_installation('python3')`. (edit: also tested with `python` instead of `python3`)
I run meson in `x64 Native Tools Command Prompt for VS 2019`.
Running meson setup gives me this message:
`src\erhe\gl\meson.build:1:0: ERROR: python3 not found`
Meanwhile, running `python3 --version` in the same console works fine and outputs `Python 3.7.8`

**Expected behavior**
I would expect python3 to be found.

**system parameters**
* native build
* Windows 10 Pro, Version 2004, OS build 19041.329
* Python 3.7.8 (tags/v3.7.8:4b47a5b6ba, Jun 28 2020, 10:03:53) [MSC v.1916 64 bit (AMD64)] on win32
* 0.54.3 (meson --version)
* 1.10.0 (ninja --version)

tksuoran:
Some more info, just in case it is any use:
```
**********************************************************************
** Visual Studio 2019 Developer Command Prompt v16.6.2
** Copyright (c) 2020 Microsoft Corporation
**********************************************************************
[vcvarsall.bat] Environment initialized for: 'x64'

C:\Program Files (x86)\Microsoft Visual Studio\2019\Community>echo %PATH%
[...]

PR #7436, opened 2020-07-12
Python test upgrade

scivision:
The "test cases/python" tests were skipping all tests on MSVC.
Now with MSVC and Windows Store Python

```
Succeeded test (without install): test cases/python/1 basic
Skipping: test cases/python/2 extmodule
Skipping: test cases/python/3 cython
Skipping: test cases/python/4 custom target depends extmodule
Succeeded test (without install): test cases/python/5 modules kwarg
```

This is relevant for finding errors with MSVC, particularly for Windows Store Python as in #7400.

Also, tests 2 and 4 were failing with Anaconda Python on Ubuntu 20.04. Now, all 5 tests succeed for this setup.

Long, nested if statements were simplified for clarity.

Setuptools is not always present, but distutils is part of Python stdlib.

As usual, this can be quickly tested by:

```sh
python run_project_tests.py --only python
```

### DIFF
--- a/run_project_tests.py
+++ b/run_project_tests.py
@@ -361,11 +361,10 @@ def _run_ci_include(args: T.List[str]) -> str:
     if not args:
         return 'At least one parameter required'
     try:
-        file_path = Path(args[0])
-        data = file_path.open(errors='ignore', encoding='utf-8').read()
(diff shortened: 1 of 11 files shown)


scivision:
This one image failure is just a timeout. It could be these extra tests running are just tipping past the 60 minute Azure time limit.

Issue #7437, opened 2020-07-12
meson build -Db_coverage=true fails on harfbuzz when meson is upgraded to 0.55


ebraminio:
`meson build -Db_coverage=true` fails on harfbuzz when meson is upgraded to 0.55 and gives,  https://github.com/harfbuzz/harfbuzz/runs/863065338

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/mesonbuild/mesonmain.py", line 131, in run
    return options.run_func(options)
  File "/usr/local/lib/python3.8/dist-packages/mesonbuild/msetup.py", line 245, in run
    app.generate()
  File "/usr/local/lib/python3.8/dist-packages/mesonbuild/msetup.py", line 159, in generate
    self._generate(env)
  File "/usr/local/lib/python3.8/dist-packages/mesonbuild/msetup.py", line 215, in _generate
    intr.backend.generate()
  File "/usr/local/lib/python3.8/dist-packages/mesonbuild/backend/ninjabackend.py", line 518, in generate
    self.generate_coverage_rules()
  File "/usr/local/lib/python3.8/dist-packages/mesonbuild/backend/ninjabackend.py", line 991, in generate_coverage_rules
    self.generate_coverage_command(e, [])
  File "/usr/local/lib/python3.8/dist-packages/mesonbuild/backend/ninjabackend.py", line 975, in generate_coverage_command
    for compiler in target.compilers.values():
AttributeError: 'CustomTarget' object has no attribute 'compilers'
```

I rechecked manually that wasn't the case with the prior version as this happened with the meson upgrade on pip.

**To Reproduce**
`git clone https://github.com/harfbuzz/harfbuzz && cd harfbuzz && meson build -Db_coverage=true`, doesn't happen before 0.55 but happens after it

**system parameters**
* Is this a [cross build](https://mesonbuild.com/Cross-compilation.html) or just a plain native build (for the same computer)? No
* what operating system, Ubuntu latest https://github.com/harfbuzz/harfbuzz/runs/863065338 which runs https://github.com/harfbuzz/harfbuzz/blob/master/.github/workflows/linux-ci.yml#L11 but happened on another bot with 20.04 also and locally on Manjaro/ArchLinux.
* what Python version are you using, 3.8.2
* what `meson --version`, 0.55.0
* what `ninja --version` if it's a Ninja build, 1.10.0

ebraminio:
[...]

PR #7411, opened 2020-07-04
ninjabackend: fix no attribute 'compilers'

phongt:
When meson.build configure with run_target() there will be an error
mesonbuild/backend/ninjabackend.py", line 977, in generate_coverage_command
    for compiler in target.compilers.values():
AttributeError: 'RunTarget' object has no attribute 'compilers'

Add the check attribute for avoiding unexpected failure.

PR #7439, opened 2020-07-12
Check for "compilers" attribute in generate_coverage_command

georgev93:
When building a meson project with Clang, if there is a "run_target" specified in the meson.build file it will enumerate as a target without a compiler when generate_coverage_command runs. Without this tweak, Meson will crash because it can't find the compiler of the run_target target.


georgev93:
Addresses #7437

georgev93:
Closing this pull request because I now see that @phongt fixed this in #7411 (using continue instead of break, which continue is the much better option). #7411 seems like a **really** important PR, because otherwise it looks like meson will crash when -Db_coverage is true and any run_target or custom_target is defined in the build.

PR #7440, opened 2020-07-12
[WIP] CI: Port MSYS2 from azure-pipelines to github actions

lazka:
This uses the [msys2/setup-msys2@v2](https://github.com/msys2/setup-msys2) to set up MSYS2. The upside being that you don't have to handle the MSYS2 install/setup quirks (that's the plan at least)

----

There is still one test failing `test cases/common/234 very long command line` and I haven't  figured out why. Maybe someone has an idea.


scivision:
common/234 is not specifically an MSYS2 issue, it happens with MSVC etc. and [requires Python >= 3.6 and a registry edit to fix](https://docs.python.org/3/using/windows.html#removing-the-max-path-limitation).

This test needs to be made a little more programmatic to avoid the intermittent failures. I tested on my Windows machine and this test failed at 260 character path length.

lazka:
We patched the MSYS2 Python to be long paths aware as well, so maybe the registry entry is missing on the GHA Windows VMs vs azure pipelines?

lazka:
Could it be that other things in the stack need to be fixed to support long paths? ninja, gcc etc?

scivision:
Ninja will throw error if the path gets too long (the error says it's too long). MinGW seems to give weird assembler errors. I was getting all 3 types of errors intermittently till I trimmed the path length down.

I guess the idea from the end user perspective is I'd rather not assume their system has been patched, as most are probably not needing to deal with long paths.

lazka:
Sounds good. I'm hoping Windows will switch this on by default at some point in the future.

lazka:
Sounds good. I'm hoping Windows will switch this on by default at some point in the future.

lazka:
Switched to the just released msys2/setup-msys2@v2

PR #7441, opened 2020-07-12
Replace `dynamic linker` with `linker`

lanza:
This header erroneously referred to the dynamic linker while the paragraph talks about
the "link editor." Change the title to account for the difference.


jpakkane:
Thanks. In the future please note that doc-only changes should have `[skip ci]` in their commit messages so CI is not run.

PR #7442, opened 2020-07-12
qt module: add qresource support to compile_translations

eli-schwartz:
A common pattern in Qt5 applications is to embed translations in the executable using the qresource system. In this case, the list of translation files is already available in the .qrc file and there's no good reason to duplicate this info in meson.build.

Let compile_translations optionally take a qrc input, in which case it will go straight to generating the relevant translations and rcc-generated .cpp, and directly return the thing users actually care about -- the .cpp for linking.

### DIFF
--- a/docs/markdown/snippets/qt_compile_translations_from_qrc.md
+++ b/docs/markdown/snippets/qt_compile_translations_from_qrc.md
@@ -1,0 +1,19 @@
+## Qt5 compile_translations now supports qresource preprocessing
+
+When using qtmod.preprocess() in combination with qtmod.compile_translations()
+to embed translations using rcc, it is no longer required to do this:
+
+```meson
+ts_files = ['list', 'of', 'files']
+qtmod.compile_translations(ts_files)
+# lang.qrc also contains the duplicated list of files
+lang_cpp = qtmod.preprocess(qresources: 'lang.qrc')
+```
+
+Instead, use:
+```meson
+lang_cpp = qtmod.compile_translations(qresource: 'lang.qrc')
+```
+
+which will automatically detect and generate the needed compile_translations
+targets.
--- a/mesonbuild/modules/qt.py
+++ b/mesonbuild/modules/qt.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import os
+import shutil
 from .. import mlog
 from .. import build
 from ..mesonlib import MesonException, extract_as_list, File, unholder, version_compare
@@ -59,7 +60,7 @@ class QtBaseModule(ExtensionModule):
             self.rcc = NonExistingExternalProgram(name='rcc' + suffix)
             self.lrelease = NonExistingExternalProgram(name='lrelease' + suffix)
 
-    def parse_qrc(self, state, rcc_file):
+    def qrc_nodes(self, state, rcc_file):
         if type(rcc_file) is str:
             abspath = os.path.join(state.environment.source_dir, state.subdir, rcc_file)
             rcc_dirname = os.path.dirname(abspath)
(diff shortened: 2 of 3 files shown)


jpakkane:
Is there a web page that describes how this is done with other build systems (so presumably qmake and CMake)? It would be useful for review, because I don't have that much experience with Qt's localisation system.

eli-schwartz:
I'm not sure there's any official guidance on this. But here's a couple of horror stories instead.

What I've see elsewhere is just passing a cmake `FILE(GLOB TS_FILES *.ts)` to `qt5_add_translation()`, then separately invoke `configure_file(foo.qrc ${CMAKE_CURRENT_BINARY_DIR} COPYONLY)` and rely on automatic dependency relationships parsed from the .qrc to ensure files are built in the right order.

You can do this with meson too, it's just... ugly:

```
# configure-time globbing
ts_sources = run_command(['python', '-c',
    'from glob import glob; print("\\n".join(glob("*.ts")), end="")'
]).stdout().split('\n')

qt5_mod.compile_translations(
    ts_files: ts_sources,
)

lang_qrc = qt5_mod.preprocess(
    qresources: configure_file(input: 'lang.qrc', output: 'lang.qrc', copy: true),
    rcc_extra_arguments: rcc_opts,
)
```

qmake is essentially the same thing, with the caveat that you need a pile of manual rules around `QMAKE_EXTRA_COMPILERS += lrelease` (this is more or less a custom generator) which write to the source tree, and the qrc file isn't copied over.

The "official" qmake guide is to have the developer run lupdate on the *.pro file, which apparently parses all TRANSLATIONS values and generates both .ts and .qm in the source tree.
The qmake wiki recommends this: https://wiki.qt.io/Automating_generation_of_qm_files
Note the use of system() to ensure it gets generated while running qmake, instead of using a QMAKE_EXTRA_COMPILERS generator.

mesonbuild/modules/qt.py
```diff
@@ -196,15 +204,43 @@ def preprocess(self, state, args, kwargs):
...
+            if ts_files:
+                raise MesonException('qt.compile_translations: Cannot specify both ts_files and qresource')
+            if os.path.dirname(qresource) != '':
+                raise MesonException('qt.compile_translations: qresource file name must not contain a subdirectory.')
+            qresource = File.from_built_file(state.subdir, qresource)
+            infile_abs = os.path.join(state.environment.source_dir, qresource.relative_name())
+            outfile_abs = os.path.join(state.environment.build_dir, qresource.relative_name())
+            os.makedirs(os.path.dirname(outfile_abs), exist_ok=True)
+            shutil.copyfile(infile_abs, outfile_abs)
+            shutil.copystat(infile_abs, outfile_abs)
```
dcbaker, line 223: couldn't you just use `shutil.copy2(infile_abs, outfile_abs)` here?

eli-schwartz: I could, the code I copied it from used this method though.

I'll take a look at reviewing meson's use of copyfile and copystat rather than copy2, while I'm at it.

eli-schwartz: https://github.com/mesonbuild/meson/pull/7619

docs/markdown/snippets/qt_compile_translations_from_qrc.md
```diff
@@ -0,0 +1,6 @@
+## Qt5 compile_translations now supports qresource preprocessing
+
+When using qtmod.preprocess() in combination with qtmod.compile_translations()
+to embed translations using rcc, it is now possible to tell the latter to
+accept a qresource file directly, autodetect the needed translations, and
+return the generated rcc source, without using preprocess().
```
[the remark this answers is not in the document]
jpakkane: Thinking further, this info should probably go in the Qt 5 module documentation page so it remains easily findable in the future.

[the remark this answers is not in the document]
eli-schwartz: I updated the doc in both places.

jpakkane:
I have no personal experience with this part of Qt5 translations either. If someone has comments, we can still change this until the next major release.

PR #7446, opened 2020-07-13
common/234: avoid intermittent failure by dynamic path length generation

scivision:
This is needed to fix #7440 and in general to avoid shaky test failures because of exceeding 260 characters for full file name. This is not just for MSYS2, but also for MSVC and other Windows compilers.

260 character limit on Windows for Meson because:

1. Python < 3.6
2. need to modify Windows 10 registry to break 260 character full pathname limit

### DIFF
--- a/test cases/common/234 very long commmand line/meson.build
+++ b/test cases/common/234 very long commmand line/meson.build
@@ -6,6 +6,10 @@ if build_machine.system() == 'windows'
   # cmd.exe: 8kb
   # CreateProcess: 32kb
   limit = 32767
+  # NOTE: filename limit is 260 characters unless
+  # 1. Python >= 3.6 is being used
+  # 2. Windows 10 registry has been edited to enable long pathnaems
+  # ninja backend uses absolute filenames, so we ensure they don't exceed 260.
 elif build_machine.system() == 'cygwin'
   # cygwin-to-win32: see above
   # cygwin-to-cygwin: no limit?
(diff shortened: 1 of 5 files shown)


test cases/common/234 very long commmand line/seq.py
```diff
@@ -1,6 +1,19 @@
 #!/usr/bin/env python3
+"""
```
jon-turney, line 2: Perhapy it would make sense to rename this script, since it's no longer doing the same thing as coreutils `seq`.

test cases/common/234 very long commmand line/seq.py
```diff
@@ -1,6 +1,19 @@
...
+generate sequence of filename that does not exceed MAX_LEN=260
+for Python < 3.6 and Windows without modified registry
+"""
 
 import sys
+import string
 
-for i in range(int(sys.argv[1]), int(sys.argv[2])):
-    print(i)
+name_len = 260 - len(sys.argv[2]) - 4 - 39 - 4 - 2
```
jon-turney, line 10: It seems like this could be zero or less, if `meson.build_root()` is long enough.  This test can't work in that case? (or if it's small enough that we can't generate enough unique filenames?)

test cases/common/234 very long commmand line/meson.build
```diff
@@ -6,6 +6,9 @@ if build_machine.system() == 'windows'
   # cmd.exe: 8kb
   # CreateProcess: 32kb
   limit = 32767
+  # NOTE: filename limit is 260 characters unless
```
jon-turney, line 9: If I've understood correctly what this change is doing, this comment should probably mention that the ninja backend will use absolute filenames, so we go through these gynmastics to arrange that they don't exceed that limit.

test cases/common/234 very long commmand line/seq.py
scivision: I've made it name_gen.py

PR #7448, opened 2020-07-13
RFC: stabilize keyval module

elmarco:
We have experimented with the module for about a year in a qemu
branch (https://wiki.qemu.org/Features/Meson), and we would like to
start moving the build system to meson. For that, keyval should have
the stability guarantees.


docs/markdown/Keyval-module.md
```diff
@@ -23,7 +23,7 @@ chosen the configuration options), output a ".config" file.
 The module may be imported as follows:
 
 ``` meson
-keyval = import('unstable-keyval')
+keyval = import('keyval')
```
xclaesse, line 26: I would like to take this opportunity to change the import implementation to allow both, so projects using "unstable-keyval" won't stop working.

elmarco: @xclaesse good idea, see my additional commit. I wonder if I should make it a different MR though.

jpakkane:
As this is new stable functionality it needs a release note snippet. Apart from that it's good to go.

elmarco:
Many thanks! (& ping @bonzini)

Now qemu just needs a meson release so we can start pushing meson conversion. Fwiw, we could in theory start doing it as early as in one month (https://wiki.qemu.org/Planning/5.1).

However, if there are no meson release plans before, say, 3 months, we are going to focus on other work and postpone the meson conversion to the _next-next_ merge window (likely in 2021..).

What are the chances to see a release of meson in about a month? Probably a bit too early for the traditional ~3 month schedule I guess?

jpakkane:
Traditionally we had releases roughly once a month. It got longer as more regressions were found in the releases that took work. However it seems that the current release is fairly smooth (knock on wood) so we could have a release out sooner.

xclaesse:
@elmarco you can already use `import('unstable-keyval')` and it will continue to work forever, I don't think you need to wait for a release.

elmarco:
> @elmarco you can already use `import('unstable-keyval')` and it will continue to work forever, I don't think you need to wait for a release.

Well, for the maintainers point-of-view, there is still no official release with API stability. I mean, keyval API could change in meson in git before the next release.

xclaesse:
That's really being picky IMHO ;-)

But I'm all for getting more regular Meson releases, I think we have been waiting too much in the last releases. The more we wait, the more new stuff gets added to master, the more we could get regressions, etc... it's a circle...

bonzini:
Yes I have a beautiful patch for adding the stable name but it is too big to fit in this margin... Ehm, to send it while I am on vacation. I will open a PR next week. For QEMU the only issue would be the ugly warning if you are on an older release, but it's solvable since we'll embed a Meson submodule for at least a year or two to smoothen the transition.

PR #7451, opened 2020-07-13
qt module: rcc supports depfiles now, given a recent enough version of Qt5

eli-schwartz:
Add depfile support to generated targets for Qt >= 5.14.

Move warning into the module init itself, to check if the version is too old before issuing. Also tweak the wording itself, to advise upgrading to a suitable version of Qt5 instead of advising to wait for a Qt bug to be fixed.

### DIFF
--- a/mesonbuild/modules/qt4.py
+++ b/mesonbuild/modules/qt4.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from .. import mlog
 from .qt import QtBaseModule
 
 
@@ -23,6 +22,4 @@ class Qt4Module(QtBaseModule):
 
 
 def initialize(*args, **kwargs):
-    mlog.warning('rcc dependencies will not work properly until this upstream issue is fixed:',
-                 mlog.bold('https://bugreports.qt.io/browse/QTBUG-45460'), fatal=False)
     return Qt4Module(*args, **kwargs)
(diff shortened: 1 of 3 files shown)


mesonbuild/modules/qt4.py
```diff
@@ -23,6 +23,4 @@ def __init__(self, interpreter):
 
 
 def initialize(*args, **kwargs):
-    mlog.warning('rcc dependencies will not work properly until this upstream issue is fixed:',
```
nirbheek, line 26: `mlog` is not used anymore in this module, so the import can be removed. That's what the lgtm warning is about.

mesonbuild/modules/qt5.py
```diff
@@ -23,6 +23,4 @@ def __init__(self, interpreter):
 
 
 def initialize(*args, **kwargs):
-    mlog.warning('rcc dependencies will not work reliably until this upstream issue is fixed:',
```
nirbheek, line 26: `mlog` is not used anymore in this module, so the import can be removed. That's what the lgtm warning is about.

mesonbuild/modules/qt4.py
eli-schwartz: Oh, nice.

textshell:
This breaks builds using Qt 5.14.2. Also the log message and the code disagree if this is supported since Qt 5.14 or 5.15

eli-schwartz:
I noticed that yesterday and wrote a patch to fix it, but I haven't gotten around to submitting it yet. I'm terribly sorry for the typo. :/

PR #7454, opened 2020-07-14
doc: Add warning about --force-fallback-for [skip ci]

xclaesse:


xclaesse:
I did a real life test for this, when building this with `--force-fallback-for=glib`:
```meson
executable('app', 'foo.c',
  dependencies : [
    dependency('gstreamer-1.0'),
    dependency('glib-2.0')
  ]
)
```

It actually produce no error, here is the link command, note how it pass glib twice:
```
[2/2] cc  -o app app.p/foo.c.o -Wl,--as-needed -Wl,--no-undefined -Wl,--start-group subprojects/glib/glib/libglib-2.0.so.0.6500.0 /usr/lib/x86_64-linux-gnu/libgstreamer-1.0.so /usr/lib/x86_64-linux-gnu/libgobject-2.0.so /usr/lib/x86_64-linux-gnu/libglib-2.0.so -Wl,--end-group '-Wl,-rpath,$ORIGIN/subprojects/glib/glib' -Wl,-rpath-link,/home/xclaesse/programmation/meson-test/builddir/subprojects/glib/glib
```

But resulting binary links only on the system version, so if you were wanted locally patched version of glib it won't work:
```
	linux-vdso.so.1 (0x00007ffe0e1cd000)
	libgstreamer-1.0.so.0 => /usr/lib/x86_64-linux-gnu/libgstreamer-1.0.so.0 (0x00007f3fa1ee1000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f3fa1cef000)
	libgobject-2.0.so.0 => /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0 (0x00007f3fa1c8f000)
	libgmodule-2.0.so.0 => /usr/lib/x86_64-linux-gnu/libgmodule-2.0.so.0 (0x00007f3fa1c89000)
	libglib-2.0.so.0 => /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0 (0x00007f3fa1b60000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f3fa1a11000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f3fa1a09000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f3fa19e6000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f3fa205e000)
	libffi.so.7 => /usr/lib/x86_64-linux-gnu/libffi.so.7 (0x00007f3fa19da000)
	libpcre.so.3 => /lib/x86_64-linux-gnu/libpcre.so.3 (0x00007f3fa1967000)
```

xclaesse:
CC @tp-m @MathieuDuponchelle

xclaesse:
oh, and yes, I don't need to specify `fallback` kwarg for glib-2.0 there, I added `[provide]` section in glib.wrap so it does implicit fallback :-D

tp-m:
Good to document this. I wonder if concrete examples would make this gotcha easier to grok than the generic `foo` and `bar` deps?

xclaesse:
I can s/foo/glib/ and s/bar/gstreamer/ but I wasn't sure those are well known projects out there... do you have other more common use-case? otoh, it doesn't hurt neither.

xclaesse:
Ok, changed to glib/gstreamer, it's likely to be one of the biggest user of this feature after all.

PR #7458, opened 2020-07-16
ninjabackend: check if target has compiler attribute

marcelhollerbach:
otherwise we are getting errors like:
```
  File "/usr/local/lib/python3.6/dist-packages/mesonbuild/mesonmain.py", line 131, in run
  File "/usr/local/lib/python3.6/dist-packages/mesonbuild/msetup.py", line 245, in run
    app.generate()
  File "/usr/local/lib/python3.6/dist-packages/mesonbuild/msetup.py", line 159, in generate
    self._generate(env)
  File "/usr/local/lib/python3.6/dist-packages/mesonbuild/msetup.py", line 215, in _generate
    intr.backend.generate()
  File "/usr/local/lib/python3.6/dist-packages/mesonbuild/backend/ninjabackend.py", line 518, in generate
    self.generate_coverage_rules()
  File "/usr/local/lib/python3.6/dist-packages/mesonbuild/backend/ninjabackend.py", line 991, in generate_coverage_rules
    self.generate_coverage_command(e, [])
  File "/usr/local/lib/python3.6/dist-packages/mesonbuild/backend/ninjabackend.py", line 975, in generate_coverage_command
    for compiler in target.compilers.values():
AttributeError: 'RunTarget' object has no attribute 'compilers'
```
This also adds a custom target to the trivial project to showcase the
problem.

### DIFF
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -972,6 +972,8 @@ int dummy;
         targets = self.build.get_targets().values()
         use_llvm_cov = False
         for target in targets:
+            if not hasattr(target, 'compilers'):
+                continue
             for compiler in target.compilers.values():
                 if compiler.get_id() == 'clang' and not compiler.info.is_darwin():
                     use_llvm_cov = True
--- a/run_unittests.py
+++ b/run_unittests.py
@@ -1272,7 +1272,6 @@ class InternalTests(unittest.TestCase):
 
         self.assertFalse(errors)
 
-
 @unittest.skipIf(is_tarball(), 'Skipping because this is a tarball release')
 class DataTests(unittest.TestCase):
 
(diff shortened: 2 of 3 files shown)


test cases/common/1 trivial/meson.build
```diff
@@ -19,6 +19,14 @@ if meson.is_cross_build()
   test('native exe in cross build', native_exe)
 endif
 
+echo = find_program('echo')
```
jpakkane, line 22: Using shell commands in tests is forbidden. This would need to be a e.g. Python script instead.

This test also does not run the test with coverage enabled so it's not really testing this case FWICT.

Finally, since this is getting fairly complex, it should not be in the test case of `1 trivial`, which is all about testing the simplest of cases.

marcelhollerbach:
Okay, i reworked this. 

The meson test is now simply taken from "109 generatorcustom" which is already generating source codes.
The test is executed with coverage beeing true, which would obviously fail here.

(The conditions under which the tests are skipped, are copied from the previous coverage test)

eli-schwartz:
This should be added to the 0.55.1 milestone.

Issue #7443, opened 2020-07-12
Update to 0.55 breaks two ports in VCPKG with "ModuleNotFoundError: No module named 'pkg_resources'"


Neumann-A:
Error:
```
Traceback (most recent call last):
  File "D:/downloads/tools/meson/meson-0.55.0/meson.py", line 26, in <module>
    from mesonbuild import mesonmain
  File "D:\downloads\tools\meson\meson-0.55.0\mesonbuild\mesonmain.py", line 25, in <module>
    from . import mconf, mdist, minit, minstall, mintro, msetup, mtest, rewriter, msubprojects, munstable_coredata, mcompile
  File "D:\downloads\tools\meson\meson-0.55.0\mesonbuild\mconf.py", line 16, in <module>
    from . import coredata, environment, mesonlib, build, mintro, mlog
  File "D:\downloads\tools\meson\meson-0.55.0\mesonbuild\build.py", line 26, in <module>
    from . import dependencies
  File "D:\downloads\tools\meson\meson-0.55.0\mesonbuild\dependencies\__init__.py", line 15, in <module>
    from .boost import BoostDependency
  File "D:\downloads\tools\meson\meson-0.55.0\mesonbuild\dependencies\boost.py", line 25, in <module>
    from .base import DependencyException, ExternalDependency, PkgConfigDependency
  File "D:\downloads\tools\meson\meson-0.55.0\mesonbuild\dependencies\base.py", line 32, in <module>
    import pkg_resources
ModuleNotFoundError: No module named 'pkg_resources'
```
CI link: https://github.com/microsoft/vcpkg/pull/12393/checks?check_run_id=863306764

Any Ideas how to fix it?

Voskrese:
witch pkg-config

Neumann-A:
@Voskrese: pkgconfig set via env variable `PKG_CONFIG`. And it schould be `<Msysroot>/usr/bin/pkg-config.exe`.

mensinda:
This might be due to #6829, where we switched to `pkg_resources` (which is part of `setuptools`). The reasoning for using this was that we already depend on `setuptools` for the meson entry point, so using this here shouldn't cause any errors. What does your current meson setup look like? Did you install it via pip?

cc @dcbaker

Voskrese:
hmm.. deps setuptools ? why?

Neumann-A:
> Did you install it via pip?

No direct download from github + extract. Nothing more. If I need to pip install something I could try that.

[...]

PR #7461, opened 2020-07-16
Remove the setuptools dependency with mesondata.py

mensinda:
This PR gets rid of the `setuptools` dependency inside `mesonbuild`. Instead `tools/gen_data.py` is used to generate the `mesondata.py` script, where all the data files are stored as strings. A unit test is used to compare the checksums of the stored and actual files to ensure that `mesondata.py` is always up-to-date.

See also:
- #6801
- #6829
- #7443


scivision:
I didn't try this myself but FWIW I generally agree with this approach, as Meson has a substantial Python < 3.7 userbase at this time.
For my own packages where most users are on Python >= 3.7 I use importlib.resources with fallback to pkg_resources. But almost all my package users have setuptools while Meson has a much larger subset without setuptools.
Nice work

Issue #6027, opened 2019-10-11
-Wl,-rpath-link for secondary dependencies causes linking failure when there are two prefixes


lantw44:
This is a regression of Meson 0.52.0. https://github.com/mesonbuild/meson/pull/5647 breaks it and reverting 7b9c348102792030859ed0001a51416506a0a092 fixes the problem. Tag the author @SoapGentoo and reviewers @jpakkane, @dcbaker and @textshell .

It is typical for a development environment to have at least two prefixes, one for the development work under the home directory and one for libraries installed on the system. We have to ensure that libraries under the development prefix is always preferred to libraries under the system prefix. Otherwise, the linker may report undefined reference errors because the version provided by the system may be too old.

The way https://github.com/mesonbuild/meson/pull/5647 adds `-Wl,-rpath-link` breaks the development environment managed by JHBuild seriously because everything linking to GLib is likely to trigger undefined reference errors with GNU ld. LLVM lld isn't affected because it just ignores `-Wl,-rpath-link`. There are two problems in the current implementation of `-Wl,-rpath-link` in `PkgConfigDependency`:

1. It only adds `-Wl,-rpath-link` for `secondary_libpaths`. It is possible for a secondary dependency to come from `prefix_libpaths`, so it should add `-Wl,-rpath-link` for `prefix_libpaths` as well. `prefix_libpaths` should come before `secondary_libpaths` because it should be preferred.
2. It appears on the command line too early. In file `mesonbuild/backend/ninjabackend.py`, class `NinjaBackend`, function `generate_link`, `linker.get_dependency_link_args` is called before `linker.build_rpath_args`. This means `-Wl,-rpath-link` arguments for external dependencies returned by `linker.get_dependency_link_args` come before `-Wl,-rpath-link` arguments for internal dependencies returned by `linker.build_rpath_args`, which is wrong because we should always prefer internal libraries to external libraries.

Let's see an example here. I was trying to build `pixman` with JHBuild.

 * The development prefix `/home/lantw44/gnome/devinstall/lib` has GLib, GTK, Pango, FriBidi.
 * The system prefix `/usr/local/lib` has PCRE, libintl, libiconv.

Here are the error messages:
```
FAILED: demos/gradient-test 
[...]

PR #7472, opened 2020-07-18
backends: fix rpath match pattern

jameshilliard:
Since -Wl,-rpath= is not the only valid rpath ldflags syntax we need to try and match all valid rpath ldflags.

### DIFF
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -14,6 +14,7 @@
 
 from collections import OrderedDict
 from functools import lru_cache
+from pathlib import Path
 import enum
 import json
 import os
@@ -455,10 +456,35 @@ class Backend:
                 args.extend(self.environment.coredata.get_external_link_args(target.for_machine, lang))
             except Exception:
                 pass
+        # Match rpath formats:
+        # -Wl,-rpath=
+        # -Wl,-rpath,
+        rpath_regex = re.compile(r'-Wl,-rpath[=,]([^,]+)')
+        # Match solaris style compat runpath formats:
+        # -Wl,-R
+        # -Wl,-R,
+        runpath_regex = re.compile(r'-Wl,-R[,]?([^,]+)')
+        # Match symbols formats:
+        # -Wl,--just-symbols=
+        # -Wl,--just-symbols,
+        symbols_regex = re.compile(r'-Wl,--just-symbols[=,]([^,]+)')
         for arg in args:
-            if arg.startswith('-Wl,-rpath='):
-                for dir in arg.replace('-Wl,-rpath=','').split(':'):
+            rpath_match = rpath_regex.match(arg)
+            if rpath_match:
+                for dir in rpath_match.group(1).split(':'):
                     dirs.add(dir)
+            runpath_match = runpath_regex.match(arg)
+            if runpath_match:
+                for dir in runpath_match.group(1).split(':'):
+                    # The symbols arg is an rpath if the path is a directory
+                    if Path(dir).is_dir():
+                        dirs.add(dir)
+            symbols_match = symbols_regex.match(arg)
+            if symbols_match:
+                for dir in symbols_match.group(1).split(':'):
+                    # Prevent usage of --just-symbols to specify rpath
+                    if Path(dir).is_dir():
+                        raise MesonException('Invalid arg for --just-symbols, {} is a directory.'.format(dir))
         return dirs
 
     def rpaths_for_bundled_shared_libraries(self, target, exclude_system=True):
--- a/run_unittests.py
+++ b/run_unittests.py
@@ -6423,19 +6423,34 @@ class LinuxlikeTests(BasePlatformTests):
         self.init(yonder_dir)
         self.build()
         self.install(use_destdir=False)
-        self.new_builddir()
 
-        # Build an app that uses that installed library.
-        # Supply the rpath to the installed library via LDFLAGS
-        # (as systems like buildroot and guix are wont to do)
-        # and verify install preserves that rpath.
-        env = {'LDFLAGS': '-Wl,-rpath=' + yonder_libdir,
-               'PKG_CONFIG_PATH': os.path.join(yonder_libdir, 'pkgconfig')}
-        self.init(testdir, override_envvars=env)
-        self.build()
-        self.install(use_destdir=False)
-        got_rpath = get_rpath(os.path.join(yonder_prefix, 'bin/rpathified'))
-        self.assertEqual(got_rpath, yonder_libdir)
+        # Since rpath has multiple valid formats we need to
+        # test that they are all properly used.
+        rpath_formats = [
+            ('-Wl,-rpath=', False),
+            ('-Wl,-rpath,', False),
+            ('-Wl,--just-symbols=', True),
+            ('-Wl,--just-symbols,', True),
+            ('-Wl,-R', False),
+            ('-Wl,-R,', False)
+        ]
+        for rpath_format, exception in rpath_formats:
+            # Build an app that uses that installed library.
+            # Supply the rpath to the installed library via LDFLAGS
+            # (as systems like buildroot and guix are wont to do)
+            # and verify install preserves that rpath.
+            self.new_builddir()
+            env = {'LDFLAGS': rpath_format + yonder_libdir,
+                   'PKG_CONFIG_PATH': os.path.join(yonder_libdir, 'pkgconfig')}
+            if exception:
+                with self.assertRaises(subprocess.CalledProcessError):
+                    self.init(testdir, override_envvars=env)
+                break
+            self.init(testdir, override_envvars=env)
+            self.build()
+            self.install(use_destdir=False)
+            got_rpath = get_rpath(os.path.join(yonder_prefix, 'bin/rpathified'))
+            self.assertEqual(got_rpath, yonder_libdir, rpath_format)
 
     @skip_if_not_base_option('b_sanitize')
     def test_pch_with_address_sanitizer(self):


jameshilliard:
This fixes a bug in #7103 which causes failures like this:
http://autobuild.buildroot.net/results/f2f/f2f912416ab11d454479ef3d22ed6b757207f84f/build-end.log

dankegel:
Cool.  (Is --rpath-link actually required in the wild?  cf. https://github.com/mesonbuild/meson/issues/6027 ?)

jameshilliard:
> Is --rpath-link actually required in the wild?

I think it's mostly used for more exotic use cases, from what I can tell we should preserve it if set as it's assumed whatever is setting it knows it's needed.

jameshilliard:
FYI this was [applied](https://github.com/buildroot/buildroot/commit/ae26a5b539242ca5952a1e4f45a43d2ff64856bb) in buildroot a few days ago and it appears to have eliminated all the rpath autobuilder failures we were seeing without it.

dankegel:
Directories added by rpath-link do not get added to the runtime rpath, so they should not be added to the dir list, I think.

Also, that regex is pretty complex, could use a comment?

jameshilliard:
> Directories added by rpath-link do not get added to the runtime rpath, so they should not be added to the dir list, I think.

Yeah, you're right looks like, there was a bug in my test where the first build directory was being used for all subsequent tests which caused the test to miss that rpath-link doesn't set rpath.

> Also, that regex is pretty complex, could use a comment?

Added

run_unittests.py
```diff
@@ -6423,19 +6423,23 @@ def test_global_rpath(self):
...
-        self.assertEqual(got_rpath, yonder_libdir)
+        # Since rpath has multiple valid formats we need to
+        # test that they are all properly used.
+        rpath_formats = ['-Wl,-rpath=', '-Wl,-rpath,']
+        for rpath_format in rpath_formats:
+            # Build an app that uses that installed library.
+            # Supply the rpath to the installed library via LDFLAGS
+            # (as systems like buildroot and guix are wont to do)
+            # and verify install preserves that rpath.
+            self.new_builddir()
```
jameshilliard, line 6435: Had to move this inside the loop otherwise only the first `rpath_format` in the `rpath_formats` list gets built.

mesonbuild/backend/backends.py
```diff
@@ -455,9 +455,12 @@ def get_external_rpath_dirs(self, target):
                 args.extend(self.environment.coredata.get_external_link_args(target.for_machine, lang))
             except Exception:
                 pass
+        # Match -Wl,-rpath= and -Wl,-rpath, formats
+        rpath_regex = re.compile(r'-Wl,-rpath?[=,]([^,@\-][^,]+)')
```
dankegel, line 459: The regex looks overly complicated still, and I'm not sure the h should be optional.

Can't it just be something like r'-Wl-rpath[=,](.*)' ?  If not, *that* is what the comment should explain.

jameshilliard: > I'm not sure the h should be optional

Yeah, had forgotten to remove that, updated.

> Can't it just be something like r'-Wl-rpath=,' ? If not, that is what the comment should explain.

Well the rest is just the capture group for matching the actual rpath paths that come after `-Wl,-rpath=` or `-Wl,-rpath,` args.

dankegel: My comment may not have come through properly.  Why is the capture group pattern so complicated?  Shouldn't it just be (.*) ?

jameshilliard: So parsing this correctly seems to be significantly more complex than I was expecting, it appears [this](https://github.com/bminor/binutils-gdb/blob/binutils-2_34/ld/lexsup.c#L1070-L1148) is the way ld parses the options which I suppose we want to try and fully emulate.

This one seems particularly annoying since we would have to actually do a path lookup:
```
For compatibility with other ELF linkers, if the -R option is followed by a directory name, rather than a file name, it is treated as the -rpath option.
```

dankegel: Oh, man, I forgot about that case.

Frankly, I would just fix the cases buildroot tickles, and file a followon bug for the others...

mesonbuild/backend/backends.py
```diff
@@ -455,10 +456,27 @@ def get_external_rpath_dirs(self, target):
...
+        # Match rpath formats:
+        # -Wl,-rpath=
+        # -Wl,-rpath,
+        rpath_regex = re.compile(r'-Wl,-rpath[=,]([^,]+)')
+        # Match symbols formats:
+        # -Wl,--just-symbols=
+        # -Wl,--just-symbols,
+        # -Wl,-R
+        # -Wl,-R,
+        symbols_regex = re.compile(r'-Wl,(?:--just-symbols[=,]|-R[,]?)([^,]+)')
```
jameshilliard, line 468: I've added a regex match for this alternate rpath format.

mesonbuild/backend/backends.py
```diff
@@ -455,10 +456,27 @@ def get_external_rpath_dirs(self, target):
...
-                for dir in arg.replace('-Wl,-rpath=','').split(':'):
+            rpath_match = rpath_regex.match(arg)
+            if rpath_match:
+                for dir in rpath_match.group(1).split(':'):
                     dirs.add(dir)
+            symbols_match = symbols_regex.match(arg)
+            if symbols_match:
+                for dir in symbols_match.group(1).split(':'):
+                    # The symbols arg is an rpath if the path is a directory
+                    if Path(dir).is_dir():
```
jameshilliard, line 478: For the alternate format we have to check if the path is a directory.

mesonbuild/backend/backends.py
jameshilliard: Well I've got that one added with a path check and it seems to pass in tests.

jameshilliard:
> -R is a Solaris linker option. gold does not accept --just-symbols as a synonym for -R.

[Docs](https://sourceware.org/binutils/docs/ld/Options.html) indicate the normal binutils ld does accept both at least, not sure about gold.

From LD options:
```
-R filename
--just-symbols=filename

   Read symbol names and their addresses from filename, but do not relocate it or include it in the output. This allows your output file to refer symbolically to absolute locations of memory defined in other programs. You may use this option more than once.

   For compatibility with other ELF linkers, if the -R option is followed by a directory name, rather than a file name, it is treated as the -rpath option.
```

dankegel:
Yes, I've read the man page, and looked at the code.

There is no way any sane person would use --just-symbols as a synonym for Solaris's ld's -R option.

It is an non-option.  It is silly.  Accepting it would be wrong.   It would taint you with its silliness.  Seriously.

jameshilliard:
>  I don't think gold accepts --just-symbols as a synonym for -R.

Ok, I've verified that gold hard errors if you pass a directory through using `--just-symbols`, however the default(bfd I think) linker accepts it just fine as described by the docs. In any case since we do a directory check before using the path as a rpath any usage of `--just-symbols` with gold will not be included since the path will not be a directory as gold would hard error on a directory.

dankegel:
Doesn't matter if gnu ld accidentally accepts that option; that was not their intent. They were supporting solaris's -R option by overloading their existing option named -R.

In short, sir, I cannot say LGTM if you include that option.  It is wrong from all but the most superficial viewpoint.

Uh, I confess, spending too much time on Twitter may have warped me.  But seriously, ixnay with the ustjay-ymbolsay.

dankegel:
If @ianlancetaylor commented on why gold doesn't treat --just-symbols as a synonym for -R, would that change your mind?

jameshilliard:
> Doesn't matter if gnu ld accidentally accepts that option; that was not their intent. They were supporting solaris's -R option by overloading their existing option named -R.

Hmm, `--just-symbols` is just the long opt version of `-R` from what I understand, seems strange to me that they would have different behavior here(well aside from the obvious naming confusion).

In any case I don't think that's all that relevant for this particular patch as this is just for handling existing external rpath linker flags properly.

> If @ianlancetaylor commented on why gold doesn't treat --just-symbols as a synonym for -R, would that change your mind?

If we want meson to prevent `--just-symbols` being used as a synonym for `-R` then that's somewhat out of scope for this PR IMO as it should be a hard error that is triggered well before this stage of compilation. I don't think having rpath behavior inconsistent with the compiler is a good idea as that could lead to obscure bugs.

jameshilliard:
By the way I wonder if this:

https://github.com/mesonbuild/meson/blob/d0c68dc11507a47b9b85de508e023d9590d60565/mesonbuild/backend/backends.py#L452-L453

should be changed to:

https://github.com/mesonbuild/meson/blob/d0c68dc11507a47b9b85de508e023d9590d60565/mesonbuild/compilers/compilers.py#L82-L83

mesonbuild/backend/backends.py
```diff
@@ -455,10 +456,35 @@ def get_external_rpath_dirs(self, target):
...
+                for dir in runpath_match.group(1).split(':'):
+                    # The symbols arg is an rpath if the path is a directory
+                    if Path(dir).is_dir():
+                        dirs.add(dir)
+            symbols_match = symbols_regex.match(arg)
+            if symbols_match:
+                for dir in symbols_match.group(1).split(':'):
+                    # Prevent usage of --just-symbols to specify rpath
+                    if Path(dir).is_dir():
+                        raise MesonException('Invalid arg for --just-symbols, {} is a directory.'.format(dir))
```
jameshilliard, line 487: Actually seems it's easy enough to blacklist `-Wl,--just-symbols` from being used to set an rpath.

dankegel: Mentioning -Wl,--just-symbols at all in this patch just clutters what should have been a tiny fix.

Better to consider the abuse of -Wl,--just-symbols an uncaught error.  I've never seen it misused in that way, have you?

jameshilliard: > Mentioning -Wl,--just-symbols at all in this patch just clutters what should have been a tiny fix.

I think we're a bit past that point since we also have to handle `-R`.

> Better to consider the abuse of -Wl,--just-symbols an uncaught error.

But it's not a build error in meson at all when used with the default gnu ld bfd linker, I don't want what should be a build time error turn into a runtime error. What I want to avoid is rpath being set using `-Wl,--just-symbols` and meson not recognizing the inconsistent state/behavior it is in. I much prefer meson to hard error if it can not proceed in a sane way as this greatly simplifies tracking down bugs such as bad/unsupported compiler flags being passed to meson, especially in complex meta build systems like buildroot which have a lot of intertwined packages.

> I've never seen it misused in that way, have you?

No, but the docs for `ld` very explicitly spelling out this `--just-symbols` behavior is supported by `ld` worries me enough that someone might be using this option that we really should catch it one way or another.

dankegel:
Again, have you ever seen it in the wild?

Meson is chock full of approximations that do not matter.  Ignoring misguided uses of -Wl,--just-symbols is appropriate here because that misuse occurs so rarely in the wild (possibly never) and because gold being the default on some platforms will tend to flense away any such crazy misuses that do occur.

jameshilliard:
> Again, have you ever seen it in the wild?

It's not something I've personally run into yet but it's also not something that's easy to search for usage of.

>  Ignoring misguided uses of -Wl,--just-symbols is appropriate here because that misuse occurs so rarely in the wild (possibly never) and because gold being the default on some platforms will tend to flense away any such crazy misuses that do occur.

Well maybe it would get caught by gold, but from my understanding gold is not commonly used/supported on complex embedded meta build systems like buildroot.

In any case I think if the intention is to not support `-Wl,--just-symbols` being used to set rpath then we should not go halfway and allow it to be accidentally used in ways meson can't handle properly. Since we already have to maintain a directory check for the `-Wl,-R` rpath special case that is effectively identical to the one used to blacklist `-Wl,--just-symbols` I don't see this as a major maintenance issue for meson.

ianlancetaylor:
I'm not sure what the issues are here.

`-R file` is equivalent to `--just-symbols file`.

`-R directory` is equivalent to `--rpath directory`.

`--just-symbols directory` is an error.  It makes no sense.

`--rpath file` is implausible, but is accepted as the rpath is a run-time feature, not a link-time feature.

I suppose it's true that for GNU ld `--just-symbols` acts like `-rpath` when the argument is a directory, but I think that's a bug.

Historically, `-R` with the meaning of `--just-symbols` came first.  `--just-symbols` was the long option synonym of `-R`.  When ELF support was added the `-rpath` option was added.  Since the Solaris linker used `-R` for that, and since `-R directory` didn't make sense, I decided that we should treat `-R directory` as `-rpath directory`.

```
Wed Jul 20 15:49:27 1994  Ian Lance Taylor  (ian@sanguine.cygnus.com)

	* ld.h (args_type): Add field rpath.
	* lexsup.c (S_ISDIR): Define if not already defined.
	(parse_args): Add support for -rpath.  If -R is used to name a
	directory, treat it as -rpath for Solaris compatibility.
	* emultempl/elf32.em (gld${EMULATION_NAME}_before_allocation):
	Pass command_line.rpath to bfd_elf32_size_dynamic_sections.
	* ldmain.c (main): Initialize command_line.rpath to NULL.
	* ld.texinfo: Document -rpath option.
```

There was never any intent for `--just-symbols directory` to add the directory to the rpath.  That was an accident.

jameshilliard:
> I suppose it's true that for GNU ld --just-symbols acts like -rpath when the argument is a directory, but I think that's a bug.

Yeah, so the basic issue is that meson needs to detect any external ldflags passed in that cause an rpath to be set in a binary(so that they are differentiated from mesons internally generated rpaths properly and aren't accidentally stripped by the install process). 

Now because `--just-symbols` can be used to set rpath in some cases my opinion is that we should either handle that case and detect+use the rpath being set by `--just-symbols` or just hard error when it is detected that `--just-symbols` is set to a directory and would thus create a rpath, if we really want to prevent `--just-symbols` from setting rpath I think we should have meson hard error rather than silently proceed which could cause unexpected difficult to debug inconsistent rpath stripping bugs.

The original bug(which meson itself did not catch) that prompted me to make this PR is basically the same type of bug one could hit if a rpath set by `--just-symbols` were just not detected at all, the only difference was in this case that meson wasn't handling/detecting `-Wl,-rpath,` syntax properly.

dankegel:
Have you tried your updated patch on the buildroot fleet to see if any project triggers your new fatal error message?

jameshilliard:
> Have you tried your updated patch on the buildroot fleet to see if any project triggers your new fatal error message?

No, but we're not likely to be hitting it at the moment in buildroot since buildroot does additional rpath sanity checks itself which would likely catch it. In buildroot the most likely source of a bug involving `--just-symbols` down the line IMO would be one coming from ldflags pulled in from an external package.

The fatal error message is handy because it makes these sorts of bugs much quicker to trace, we hit a lot of compiler flag and build system related issues in general and it's often not immediately clear what the root cause is when the build error only gets caught by something like the buildroot rpath sanity checker, it's much easier to trace issues like this when they throw fatal errors with clear error messages immediately and early in the package compilation process.

dankegel:
I defer to anyone who runs builds across such a large fleet of projects.  Onwards!

dankegel:
In case it wasn't clear: LGTM.   I apologize for missing the alternate spellings of the option in my original patch, and am really happy with this pull request.

Issue #5191, opened 2019-04-01
rpath contains paths to static libs


TheQwertiest:
Create a project with ninja backend and mirror layout:
```meson
# subdir_1
my_st_lib = static_library('my_st_lib', ...)
# subdir_2
my_sh_lib = shared_library('my_sh_lib', link_with: my_st_lib, ...)
```
`my_sh_lib`'s rpath will contain path to `subdir_1` (e.g. `$ORIGIN/../subdir_1`), which AFAIK is pointless.

TheQwertiest:
If this is not an intended behaviour, I can make a PR that fixes it.

xphoenix:
Quite annoying staff in some cases. My scenario - Im building EFI app under linux. With using clang it is finnally becomes as simple as set the right triplet:

```
boot_x86_64_exe = executable(
  'boot.x86_64.efi', boot_x86_64_sources,
  include_directories: boot_headers,
  link_with: [lib_elc, boot_libs_crypto, boot_libs_compress],

  c_args: ['-target','x86_64-pc-win32-coff','-fno-stack-protector','-fshort-wchar','-mno-red-zone','-D__LIB_ENV_EFI=1'],
  link_args: ['-target','x86_64-pc-win32-coff','-fuse-ld=lld','-nodefaultlibs','-nostartfiles','-Wl,-subsystem:efi_application','-Wl,-entry:efi_main']
)
```

However meson messes up flags adding rlink staff even if all libraries are ```static```. Also it makes no sens to add ```rlink``` flags with ```x86_64-pc-win32-coff``` and ```-fuse-ld=lld``` you will get output:
```
lld-link: warning: ignoring unknown argument '-rpath'
lld-link: warning: ignoring unknown argument '-rpath-link'
lld-link: warning: ignoring /lib since it's not the first argument
lld-link: error: could not open '$ORIGIN/': No such file or directory
....
```
After a bit of grep i've made a little patch in Backends class:
```
    @lru_cache(maxsize=None)
    def get_link_dep_subdirs(self):
        result = OrderedSet()
        for i in self.link_targets:
            if isinstance(i, StaticLibrary): continue # here is it, skip static libs
            result.add(i.get_subdir())
            result.update(i.get_link_dep_subdirs())
        return result
```
However something tells me it is not a really proper way to fix the issue :(

PR #7474, opened 2020-07-19
build: Skip rpath subdir if target is a static library

richardweinberger:
When building the list of rpaths from all targets in subdirs,
skip static libraries.
These libraries don't need a rpath and are often used to
organize a project.

This solution is taken from:
https://github.com/mesonbuild/meson/issues/5191
And authored by:
https://github.com/xphoenix

Signed-off-by: Richard Weinberger <richard@nod.at>


jpakkane:
This will be in #7475 so closing.

PR #7476, opened 2020-07-19
Fix dirs typo in Vala docs

wb9688:
`find_library()` accepts a `dirs` argument, not a `dir` argument.


jpakkane:
Thanks. Just note in the future that doc-only changes should have `[skip ci]` in their commit messages so CI is not run.

PR #7477, opened 2020-07-20
mtest: report failures and overall summary after test(s) logs

camilo-celis:
When running multiple test suites that include large amount of tests we have found that it is more beneficial to (1) report a summary of results at **after** all logs from tests have been written to stdout, and (2) in case of failure report a small summary of only those tests. This could be a subjective topic but I thought we could benefit from easying it out the process spotting failures.

### DIFF
--- a/mesonbuild/mtest.py
+++ b/mesonbuild/mtest.py
@@ -794,6 +794,7 @@ class TestHarness:
     def __init__(self, options: argparse.Namespace):
         self.options = options
         self.collected_logs = []  # type: T.List[str]
+        self.collected_failures = []  # type: T.List[str]
         self.fail_count = 0
         self.expectedfail_count = 0
         self.unexpectedpass_count = 0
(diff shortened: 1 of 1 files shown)


jpakkane:
Seems reasonable. Anyone else have comments?

camilo-celis:
For completeness: as an example when running all GStreamer tests with `meson test -C build/ -v` if a 2 test fails we would get: 
```
. . .
106/109 pipelines_stress                        OK      10.79 s 
107/109 generic_sinks                           OK       4.08 s 
108/109 gst_gstcpp                              OK       0.02 s 
109/109 libs_gstlibscpp                         OK       0.01 s 

Summary of Failures:

  3/109 gst_gstabi                              FAIL     0.26 s (exit status 1)
  9/109 gst_gstclock                            FAIL     0.25 s (exit status 1)


Ok:                  107
Expected Fail:         0
Fail:                  1
Unexpected Pass:       0
Skipped:               1
Timeout:               0

```

PR #7478, opened 2020-07-20
vim: Syntax-highlight continue and break statements

nirbheek:


data/syntax-highlighting/vim/syntax/meson.vim
```diff
@@ -32,8 +32,9 @@ set cpo&vim
 
 " http://mesonbuild.com/Syntax.html
 syn keyword mesonConditional	elif else if endif
-syn keyword mesonRepeat	foreach endforeach
+syn keyword mesonRepeat		foreach endforeach
```
eli-schwartz, line 35: What's the purpose of this diff hunk?

nirbheek: A tab for editor alignment:
image

ignatenkobrain: I think depends on tab size :)

eli-schwartz: Meh. this file contains `" vim:set sw=2 sts=2 ts=8 noet:` so the change is correct and needed if I open it in vim, but git log seems to align tabs including the leading `-+ ` for the diff context. Sorry...

nirbheek: No worries, thanks for eyeballs :)

PR #7481, opened 2020-07-20
Make meson recognize the Qualcomm LLVM toolchain

antonysigma:
Meson calls `path/to/clang++ --version` to guess which build toolchain the
user has picked to build the source code. For the Qualcomm LLVM toolchain,
the output have an unusual output as shown below:

```
clang version 8.0.12
Snapdragon LLVM ARM Compiler 8.0.12 (based on llvm.org 7.0+)
Target: arm-unknown-linux-gnueabi
Thread model: posix
Repository: (ssh://git-hexagon-aus.qualcomm.com:...)
InstalledDir: /pkg/qct/software/llvm/release/arm/8.0.12/bin
```

Another unusual pattern is the output of `path/to/ld.qcld --version`:

```
ARM Linker from Snapdragon LLVM ARM Compiler Version 8.0.12
ARM Linker based on LLVM version: 8.0
```

The Meson logic is modified accordingly so that Meson can correctly
determine toolchain as "LLVM aarch64 cross-compiler on GNU/Linux64 OS".

This is the expected output of
`meson --native-file native_file.ini --cross-file cross_file.ini build/aarch64-debug/`:

```
...
C++ compiler for the host machine: ... (clang 8.0.12 "clang version 8.0.12")
C++ linker for the host machine: ... ld.lld 8.0.12
...
```

Reference: https://developer.qualcomm.com/software/snapdragon-llvm-compiler-android

### DIFF
--- a/mesonbuild/compilers/mixins/clang.py
+++ b/mesonbuild/compilers/mixins/clang.py
@@ -113,6 +113,11 @@ class ClangCompiler(GnuLikeCompiler):
         # (and other gcc-like compilers) cannot. This is becuse clang (being
         # llvm based) is retargetable, while GCC is not.
         #
+
+        # qcld: Qualcomm Snapdragon linker, based on LLVM
+        if linker == 'qcld':
+            return  ['-fuse-ld=qcld']
+
         if shutil.which(linker):
             if not shutil.which(linker):
                 raise mesonlib.MesonException(
--- a/mesonbuild/environment.py
+++ b/mesonbuild/environment.py
@@ -52,6 +52,7 @@ from .linkers import (
     GnuBFDDynamicLinker,
     GnuGoldDynamicLinker,
     LLVMDynamicLinker,
+    QualcommLLVMDynamicLinker,
     MSVCDynamicLinker,
     OptlinkDynamicLinker,
     PGIDynamicLinker,
@@ -888,9 +889,12 @@ class Environment:
             check_args += override
 
         _, o, e = Popen_safe(compiler + check_args)
-        v = search_version(o)
+        v = search_version(o + e)
         if o.startswith('LLD'):
             linker = LLVMDynamicLinker(
+                compiler, for_machine, comp_class.LINKER_PREFIX, override, version=v)  # type: DynamicLinker
+        elif 'Snapdragon' in e and 'LLVM' in e:
+            linker = QualcommLLVMDynamicLinker(
                 compiler, for_machine, comp_class.LINKER_PREFIX, override, version=v)  # type: DynamicLinker
         elif e.startswith('lld-link: '):
             # The LLD MinGW frontend didn't respond to --version before version 9.0.0,
@@ -1087,7 +1091,7 @@ class Environment:
                 return cls(
                     compiler, version, for_machine, is_cross, info, exe_wrap,
                     target, linker=linker)
-            if 'clang' in out:
+            if 'clang' in out or 'Clang' in out:
                 linker = None
 
                 defines = self.get_clang_compiler_defines(compiler)
--- a/mesonbuild/linkers.py
+++ b/mesonbuild/linkers.py
@@ -927,6 +927,12 @@ class ArmClangDynamicLinker(ArmDynamicLinker):
     def import_library_args(self, implibname: str) -> T.List[str]:
         return ['--symdefs=' + implibname]
 
+class QualcommLLVMDynamicLinker(LLVMDynamicLinker):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # ARM Linker from Snapdragon LLVM ARM Compiler
+        self.id = 'ld.qcld'
 
 class PGIDynamicLinker(PosixDynamicLinkerMixin, DynamicLinker):
 


nirbheek:
These changes look small and safe to me.

antonysigma:
Thank you @nirbheek for reviewing the draft. I have changed it to a normal PR.

mesonbuild/compilers/mixins/gnu.py
```diff
@@ -306,7 +306,7 @@ def get_include_args(self, path: str, is_system: bool) -> T.List[str]:
 
     @classmethod
     def use_linker_args(cls, linker: str) -> T.List[str]:
-        if linker not in {'gold', 'bfd', 'lld'}:
+        if linker not in {'gold', 'bfd', 'lld', 'qcld'}:
```
nirbheek, line 309: In what way is qcld different than the other linkers?

nirbheek: Basically just asking so you can add a comment about it here.

antonysigma: Thanks. I added a comment here.

mesonbuild/compilers/mixins/gnu.py
```diff
@@ -306,7 +306,9 @@ def get_include_args(self, path: str, is_system: bool) -> T.List[str]:
 
     @classmethod
     def use_linker_args(cls, linker: str) -> T.List[str]:
-        if linker not in {'gold', 'bfd', 'lld'}:
+        if linker not in {'gold', 'bfd', 'lld',
+                # qcld: Qualcomm Snapdragon linker, based on LLVM.
+                'qcld'}:
```
antonysigma, line 311: Comment added here to describe what is 'qcld'.

dcbaker: Does this linker actually work with gnu? I think we probably want to put this in the clang implementation.

mesonbuild/environment.py
```diff
@@ -888,8 +888,8 @@ def _guess_nix_linker(self, compiler: T.List[str], comp_class: T.Type[Compiler],
             check_args += override
 
         _, o, e = Popen_safe(compiler + check_args)
-        v = search_version(o)
-        if o.startswith('LLD'):
+        v = search_version(o + e)
+        if o.startswith('LLD') or 'LLVM' in e:
```
dcbaker, line 892: This seems off to me. If qualcom sees fit to not identify their linker as lld, i'd rather have a separate class for it, even if it inherits the lld class.

mesonbuild/environment.py
```diff
@@ -1087,7 +1087,7 @@ def sanitize(p):
                 return cls(
                     compiler, version, for_machine, is_cross, info, exe_wrap,
                     target, linker=linker)
-            if 'clang' in out:
+            if 'clang' in out or 'Clang' in out:
```
dcbaker, line 1090: Is the qualcom clang identical to vanilla clang (at least in terms of command line arguments)? If it's not we should do like we do with Apple's clang and have a separate representation.

mesonbuild/environment.py
antonysigma: Yes please. I don't know Meson enough to write up a separate class for baremetal LLVM linker variants though. Could you please pass me some pseudo-code to start with?

antonysigma: > Does this linker actually work with gnu? I think we probably want to put this in the clang implementation.

Oh. That explains why Meson (w/o my patch) classified the Qualcomm linker as Apple. Again, I am happy to rewrite the patch if you can pass me some pseudo code of the derived class.

mesonbuild/environment.py
antonysigma: Yes it is. Qualcomm calls it the "baremetal" LLVM toolchain for Snapdragon chipsets.

mesonbuild/compilers/mixins/gnu.py
antonysigma: Oh. That explains why Meson (w/o my patch) classified the Qualcomm linker as Apple's clang. I am happy to rewrite the patch if you can pass me some pseudo code of the derived class.

mesonbuild/environment.py
antonysigma: Oh. That explains why Meson (w/o my patch) classified the Qualcomm linker as Apple's clang. Again, I am happy to rewrite the patch if you can pass me some pseudo code of the derived class.

mesonbuild/environment.py
antonysigma: Yes it is. Qualcomm calls it the "baremetal" LLVM toolchain for Snapdragon chipsets.

mesonbuild/environment.py
dcbaker: What I'd do here then is something like (in mesonbuild/linkers.py):

```python
class QualcommLLVMDynamicLinker(LLVMDynamicLinker):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.id = 'ld.qcld'  # or you could refactor LLVMDynamicLinker to not hardcode ld.lld
```

Then you need to add it to the all list and do the proper finding thing for it in env.

mesonbuild/environment.py
dcbaker: Okay, then this seems fine.

mesonbuild/compilers/mixins/gnu.py
dcbaker: in https://github.com/mesonbuild/meson/blob/master/mesonbuild/compilers/mixins/clang.py#L111 you want to add something like:

```python
if linker == 'qcld':
    return ['-fuse-ld=qcld']
```

mesonbuild/compilers/mixins/clang.py
```diff
@@ -113,6 +113,11 @@ def use_linker_args(cls, linker: str) -> T.List[str]:
         # (and other gcc-like compilers) cannot. This is becuse clang (being
         # llvm based) is retargetable, while GCC is not.
         #
+
+        # qcld: Qualcomm Snapdragon linker, based on LLVM
+        if linker == 'qcld':
+            return  ['-fuse-ld=qcld']
+
```
antonysigma, line 120: @dcbaker Here is the changes you suggested.

mesonbuild/linkers.py
```diff
@@ -927,6 +927,12 @@ def export_dynamic_args(self, env: 'Environment') -> T.List[str]:
     def import_library_args(self, implibname: str) -> T.List[str]:
         return ['--symdefs=' + implibname]
 
+class QualcommLLVMDynamicLinker(LLVMDynamicLinker):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # ARM Linker from Snapdragon LLVM ARM Compiler
+        self.id = 'ld.qcld'
```
antonysigma, line 935: Here it is. Since `ld.lld` is far more popular than `ld.qcld`, I prefer not to disturb the default behavior of `LLVMDyanamicLinker`.

PR #7484, opened 2020-07-20
prevent disabler() object from overwriting arrays

### LINKED ISSUE #7107: disabler() objects convert arrays into non-arrays
When an array is used to contain a bunch of built objects, e.g. executable()s , if any of those objects becomes a disabler() object, the array stops being an array, and syntactic breakage occurs.

For example:
```python
project('list-of-disablers', 'c',
        version : '2.35-rc2',
        license : 'GPLv2+')
exes = []

exes += executable(
  'foo',
  'foo.c')

exes += executable(
  'foo2',
  'foo.c',
  dependencies : disabler())

exes += executable(       #here
  'foo3',
  'foo.c')
```
This fails with:
```
../meson.build:16:0: ERROR: The += operator currently only works with arrays, dicts, strings or ints 
```
because fails because `exes` is suddenly not an array any more, but a disabler() object. This is just not terribly useful. Instead of changing the types, disabled object need to keep their type, but just get disabled.

meson-0.54.0-1.fc32.noarch

Apache-HB:
fixes #7107

### DIFF
--- a/mesonbuild/interpreterbase.py
+++ b/mesonbuild/interpreterbase.py
@@ -810,9 +810,7 @@ The result of this is undefined and will become a hard error in a future Meson r
         assert(isinstance(node, mparser.PlusAssignmentNode))
         varname = node.var_name
         addition = self.evaluate_statement(node.value)
-        if is_disabler(addition):
-            self.set_variable(varname, addition)
-            return
+
         # Remember that all variables are immutable. We must always create a
         # full new variable and then assign it.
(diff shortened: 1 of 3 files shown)


nirbheek:
We recently hit this downstream in gstreamer too: https://gitlab.freedesktop.org/gstreamer/gst-plugins-bad/-/merge_requests/1519

I think it's a safe change.

keszybz:
Thanks!

PR #7490, opened 2020-07-22
InternalDependency: Add as_link_whole() method

xclaesse:


docs/markdown/snippets/as_link_whole.md
```diff
@@ -0,0 +1,12 @@
+## `dep.as_link_whole()`
+
+Dependencies created with `declare_dependency()` now has new method `as_link_whole()`.
+It returns a copy of the dependency object with all link_with arguments changed
+to link_whole. This is useful for example for fallback dependency from a
+subproject built with `default_library=static`.
```
jpakkane, line 6: This should probably mention what happens if the dependency is a shared library, i.e. nothing at all.

jpakkane:
Other than that and rebase to fix conflicts, LGTM.

docs/markdown/snippets/as_link_whole.md
xclaesse: Actually currently that would try to link_whole the shared library and abort with: `meson.build:5:0: ERROR: <SharedLibrary foo@sha: libfoo.so> is not a static library.`. I think it's preferable to have an error than silently doing a dynamic link, don't you think?

Thinking about it, maybe we could automatically convert BothLibraries to StaticLibrary however. That could be done in general for any `library(..., link_whole: both_libraries(...))` . What do you think?

dcbaker: I personally like an error for something we don't want to support. And it's easier to remove the error case later than to change it to an error.

For both_libraries I think we really need to think about how to make them more usable. maybe add `as_static` and `as_shared` to the library object for getting a specific one?

xclaesse: @dcbaker agreed, we can always lift an error later if needed, but we can't do the opposite, so better stay safe when in doubt.

About `as_shared` and `as_static` I agree and wanted them as well, but that can be done in separate PR I think, ok?

xclaesse: I added in the doc that link_with objects must be static libraries otherwise it raise an error.

jpakkane:
Failure seems relevant:

 ```
Compiler stderr:
 D:/a/msys64/mingw32/bin/../lib/gcc/i686-w64-mingw32/10.2.0/../../../../i686-w64-mingw32/bin/ld.exe: cannot find -ldl
collect2.exe: error: ld returned 1 exit status
```

xclaesse:
> Failure seems relevant:
> 
> ```
> Compiler stderr:
> D:/a/msys64/mingw32/bin/../lib/gcc/i686-w64-mingw32/10.2.0/../../../../i686-w64-mingw32/bin/ld.exe: cannot find -ldl
> collect2.exe: error: ld returned 1 exit status
> ```

Not related, it's failing in master too.

docs/markdown/snippets/as_link_whole.md
dcbaker: @xclaesse yes, absolutely in a different MR.

PR #7491, opened 2020-07-22
RFC: logs: Prepend current subproject name to all messages

xclaesse:
Meson used to prepend '|' for each nested subproject to distinguish in
the logs where a subproject start and ends. It is more useful to print
the current subproject name.

Also print the call stack when starting a new subproject to better see
which subproject chain leads to to.

--
This is an RFC, I'm not sure this should be the default output, or if we should have a command line option for that?

### DIFF
--- a/mesonbuild/mlog.py
+++ b/mesonbuild/mlog.py
@@ -201,7 +201,7 @@ def process_markup(args: T.Sequence[T.Union[AnsiDecorator, str]], keep: bool) ->
             arr.append(str(arg))
     return arr
 
-def force_print(*args: str, **kwargs: T.Any) -> None:
+def force_print(*args: str, nested: str, **kwargs: T.Any) -> None:
     if log_disable_stdout:
         return
     iostr = io.StringIO()
@@ -210,7 +210,7 @@ def force_print(*args: str, **kwargs: T.Any) -> None:
 
     raw = iostr.getvalue()
     if log_depth:
-        prepend = log_depth[-1] + '| '
+        prepend = log_depth[-1] + '| ' if nested else ''
         lines = []
         for l in raw.split('\n'):
             l = l.strip()
@@ -250,6 +250,7 @@ def log(*args: T.Union[str, AnsiDecorator], is_error: bool = False,
 
 def _log(*args: T.Union[str, AnsiDecorator], is_error: bool = False,
         **kwargs: T.Any) -> None:
+    nested = kwargs.pop('nested', True)
     arr = process_markup(args, False)
     if log_file is not None:
         print(*arr, file=log_file, **kwargs)
@@ -257,7 +258,7 @@ def _log(*args: T.Union[str, AnsiDecorator], is_error: bool = False,
     if colorize_console():
         arr = process_markup(args, True)
     if not log_errors_only or is_error:
-        force_print(*arr, **kwargs)
+        force_print(*arr, nested=nested, **kwargs)
 
 def log_once(*args: T.Union[str, AnsiDecorator], is_error: bool = False,
              **kwargs: T.Any) -> None:
--- a/mesonbuild/mlog.py
+++ b/mesonbuild/mlog.py
@@ -210,7 +210,7 @@ def force_print(*args: str, **kwargs: T.Any) -> None:
 
     raw = iostr.getvalue()
     if log_depth:
-        prepend = log_depth[-1] + '|'
+        prepend = log_depth[-1] + '| '
         raw = prepend + raw.replace('\n', '\n' + prepend, raw.count('\n') - 1)
 
     # _Something_ is going to get printed.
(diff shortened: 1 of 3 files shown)


xclaesse:
See the result on gst-build: https://imgur.com/cWkXg3T

mensinda:
I am not a huge fan of this change since it makes the output a lot noisier and the indentation is all over the place. Could you make the indentation uniform by calculating the max. project name length?

So I would vote for an opt-in for printing the subproject name. However, the call stack is really useful and can be printed regardless.

Also, having something like `Executing ... method` in the call stack is confusing since this suggests something about custom methods/functions (and meson doesn't have those) at first glance.

tristan957:
I think calculating the longest name could get pretty ugly if you are using subprojects of 2+ levels in depth. Right `|` is really manageable but if you start getting into 2 or 3 layers of subprojects with this new method, you are going to have a bad time.

```
subproj1|...
subproj1|subproj2|...
```

tp-m:
I'd be happy with just the current project's name being printed, i.e.
```
|gst-build|..
||gstreamer|..
|||glib|..
||||libffi|..
```

xclaesse:
FYI, here is how it looks like with this PR:

Capture d’écran de 2021-03-08 08-18-27

xclaesse:
I rebased this PR in case people want to give it a try.

> I'd be happy with just the current project's name being printed, i.e.
> 
> ```
> |gst-build|..
> ||gstreamer|..
> |||glib|..
> ||||libffi|..
> ```
IMHO the stack of `|` is not really needed when printing the current subproject. Omitting it makes things a bit better aligned. Note that this PR also print the stack when starting a subproject: `x264|Executing subproject gst-plugins-ugly:x264 method meson` so you know x264 subproject has been called from -ugly.

tp-m:
My comment was in response to the previous comment not about the MR implementation, sorry for the confusion.

The formatting you suggest works for me as well, though I wonder if my proposal would make the "subproject indent level" clearer. No strong feelings, I'm just happy that the subproject name gets printed :)

xclaesse:
@jpakkane @dcbaker what's your opinion on this? I personally still think it helps a lot to debug dependency tree when there are many fallbacks. It makes a bit longer debug lines, and stuff are a bit less aligned, but IMHO that's minor cons.

dcbaker:
I'm fine with this I'm the current iteration (though i haven't looked at the code). The one thing that would be nice is some indication of why a subproject was configured, which we have now, even if that's just an mlog.debug message i can get from the on disk log, unless i missed that :)

xclaesse:
@jpakkane we already do that.

@dcbaker it's already pretty clear no? What extra info do you have in mind?
```
|Looking for a fallback subproject for the dependency graphene-gobject-1.0
|
||Executing subproject graphene method meson 
```

xclaesse:
One thing that could change is to not print the stack of `|` when it's an empty line.

jpakkane:
At least there should be spaces around the pipe (probably only the last one between the subprojects and the actual message). The current one looks cramped.

xclaesse:
Ok, did a few tweaks:
- Do not print 'subproject|' on empty lines
- Do not print 'subproject|' for the message 'Executing subproject'
- Do not print "method meson" (this one is unrelated, but I always found it confusing)

xclaesse:
Capture d’écran de 2021-03-10 14-25-43

tristan957:
This is looking much better. What does two layers of subprojects look like?

dcbaker:
@xclaesse I missed that, looks good to me.

mesonbuild/mlog.py
```diff
@@ -370,10 +375,10 @@ def format_list(input_list: T.List[str]) -> str:
...
-def nested() -> T.Generator[None, None, None]:
+def nested(name: str = '') -> T.Generator[None, None, None]:
     global log_depth
-    log_depth += 1
+    log_depth.append(name)
     try:
         yield
     finally:
-        log_depth -= 1
+        log_depth.pop()
```
dcbaker, line 384: nit: pop is for returning a value, `del log_depth[-1]` is more correct.

xclaesse:
> This is looking much better. What does two layers of subprojects look like?

It only print the current subproject, not the whole stack. But the line "executing subproject..." now prints the whole stack.

mesonbuild/mlog.py
xclaesse: Does it really change anything? IMHO pop() is more readable.

xclaesse:
CI failure seems unrelated, any objection on merging this now?

tristan957:
@xclaesse did you end up adding space on both sides of the pipe? I am fine with this as is, not that my opinion super matters :).

xclaesse:
> @xclaesse did you end up adding space on both sides of the pipe? I am fine with this as is, not that my opinion super matters :).

Only after the pipe, as shown in https://github.com/mesonbuild/meson/pull/7491#issuecomment-795954861.

PR #7492, opened 2020-07-22
visualstudio: warning_level 0 should not map to /W1

xclaesse:
In every other compilers level 0 maps to no argument at all.


nirbheek:
lgtm

PR #7494, opened 2020-07-24
Fix two warnings

keszybz:


jpakkane:
This is probably caused by the change of output:

```
expected "test cases/warning/2 languages missing native/meson.build:2: WARNING: add_languages is missing native:, assuming languages are wanted for both host and build." not found in stdout 
```

keszybz:
Thanks, the tests pass now.

PR #7499, opened 2020-07-24
tests/common/227: don't have symlink in Git

scivision:
Windows Git users with symlinks have a **constantly Git dirty Meson repo** from
this "a_symlink".  This is true even when the Windows install has Developer Mode and symlinks enabled via Windows group policy.

This change generates a symlink in the build directory when the test is run, instead of tracking the symlink in Git. 
Now, there is no longer a constantly Git dirty Meson project directory on Windows.

While CMake has the native capability to create symlinks, Meson doesn't have this (it's trivially easy to add, but I think I remember a discussion where it wasn't desired?)

### DIFF
--- a/test cases/common/227 fs module/meson.build
+++ b/test cases/common/227 fs module/meson.build
@@ -18,7 +18,9 @@ assert(not fs.exists('nonexisting'), 'Nonexisting file was found.')
 is_git_checkout = fs.exists('../../../.git')
 
 if not is_windows and build_machine.system() != 'cygwin' and is_git_checkout
-  assert(fs.is_symlink('a_symlink'), 'Symlink not detected.')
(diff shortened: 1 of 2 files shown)


test cases/common/227 fs module/meson.build
```diff
@@ -18,7 +18,9 @@ assert(not fs.exists('nonexisting'), 'Nonexisting file was found.')
 is_git_checkout = fs.exists('../../../.git')
```
nirbheek, line 18: I think we can get rid of `is_git_checkout` entirely from this file now that the symlink is generated at configure time.

scivision: Thanks I removed that now unneeded code.

PR #7502, opened 2020-07-26
Update syntax info doc page

michaelbrockus:
I have written a better paragraph for this recommendation on naming variables in Meson.


jpakkane:
:crossed_swords: conflicts. Also please squash and rebase the commits so there is only one.

jpakkane:
There are now 5 commits, when there should be only one. Do not merge master into your branch, [use rebase instead](https://mesonbuild.com/Contributing.html#keeping-pull-requests-up-to-date).

Issue #3047, opened 2018-02-12
Meson needs native understanding of linker scripts


hughsie:
Meson doesn't seem to understand much about linkers; most of the projects porting to meson do something like this:

    mapfile = 'libgusb.ver'
    vflag = '-Wl,--version-script,@0@/@1@'.format(meson.current_source_dir(), mapfile)
    gusb = shared_library(
      'gusb',
      sources : [
        'gusb-context.c',
      ],
      soversion : lt_current,
      version : lt_version,
      dependencies : [
        libusb,
      ],
      link_args : vflag,
      link_depends : mapfile,
      install : true
    )

Which is fine, unless you discover that OS-X doesn't support `--version-script` : https://github.com/hughsie/libgusb/issues/11

Rather than have to do evil things to the meson.build file to support OS-X, I would much prefer fr meson to know what a version-script is, rather than just dumping it in a command line. In the case of OS-X the version script could just be ignored. Ideas welcome, thanks!

nirbheek:
Just for the record, could you explain why you need a version script instead of using symbol visibility prototypes?

hughsie:
The version script provides a record of the ABI/API over time for each version.

inigomartinez:
Several gnome and freedesktop packages are also using version scripts, so this feature would be very appreciated.

I would also like to point out that trying to workaround this [doesn't seem to be an easy task](https://bugzilla.gnome.org/show_bug.cgi?id=794365#c19), given the number of existing linkers and that their behaviour does not seem to be always the same.

npmccallum:
I'm also using linker scripts in my projects. I'd very much appreciate this.

emersion:
On macOS, Clang supports `-exported_symbols_list`, which is a list of exported symbols. Here is an example `meson.build` that supports both: https://github.com/mkhl/mrsh/blob/f94361fdff16f3a19a764691a418fb6734a634ad/meson.build#L40-L48

I'd like to help with this issue, however I'm not sure what the best design is. Make Meson parse version scripts and convert it to a symbol list when required (not all platforms support symbol versioning)? Or the other way around: make Meson upgrade symbol lists to version scripts if necessary (GCC doesn't support symbol lists), allow users to provide both if they want to have symbol versioning?

jpakkane:
[...]

PR #7505, opened 2020-07-27
Do not enclose link arguments between start-group and end-group

antonysigma:
**Summary**
This patch intends to fix the bug when user introduce linker script using the following build rules.

```meson
library('some_library',
  sources: ...
  link_args: ['-Wl,-T,path/to/linker_script.dl'],
  dependencies: some_library,
)
```

**What went wrong**
Previously, meson will enclose the argument `-Wl,-T,path/to/linker_script.dl' between the two flags,
`--start-group` and `--end-group`, and the LLVM linker will fail with the following error:

```
LLVM ERROR: Assertion `0` failed with : Linker script inside start group and end group not supported
```

**What the patch fixes**

The patch resolves the above bug by enclosing only the consecutive libraries (I called them sub-groups) with the above `--*-group` pair. Here, it is assumed that the library/libraries in each sub-groups does/do not have circular symbol dependencies outside the sub-groups.

**Alternate solution doesn't work**

If the user only wants to "augment" the system linker script, it is possible to define meson build rules with the following pseudo-code. 
https://ftp.gnu.org/old-gnu/Manuals/ld-2.9.1/html_node/ld_3.html

```meson
library('some_library',
  sources: ...
  objects: files('path/to/linker_script.dl'),
  dependencies: some_library,
)
```

Such behaviour is described in https://man7.org/linux/man-pages/man1/ld.1.html:
> If the linker cannot recognize the format of an object file, it will assume that it is a linker script.  A script specified in this way augments the main linker script used for the link.

However, when user introduce the prefix `-Wl,-T,' to the path of the linker script, it is very clear that the user intents to "replace" the system linker script. So, the above trick does not apply here.


antonysigma:
Hi everyone, here is the draft proposal to fix the linker error:

```
LLVM ERROR: Assertion `0` failed with : Linker script inside start group and end group not supported
```

I would like to hear your opinion on how to improve the patch above.

a1xndr:
Is there any update on this? This is problem is lso affecting QEMU's builds on the oss-fuzz infrastructure, where we rely on a linker-script.

antonysigma:
Hi @a1xndr , thank you for commenting on this draft PR. I am currently asking everyone for comments / revisions on my patch to resolve the issue "linker script cannot be enclosed in --start-group and --end-group" with Clang. And we definitely join force to polish it.

First, let me address your questions:

> Is there any update on this? This is problem is lso affecting QEMU's builds on the oss-fuzz infrastructure, where we rely on a linker-script.

This draft PR is yet to be picked up by the core developers of the project. The patch works on my own machine for Qualcomm Snapdragon LLVM toolchain, but feel free to test it on your enviroment with QEMU. Or, do you mean my patch did break your environment?

> "Make meson recognize the Qualcomm LLVM toolchain"
> Should this commit be in a different PR?

Yes, it was merged already, so I definitely need to rebase my commits in this draft PR. I will do it later today.

antonysigma:
One more thing, I was under the impression that in order for my PR to be merged, it must meet the following criteria:
1. it must not import non-native Python modules/libraries in Python3 environment;
2. to support a new toolchain (e.g. Snapdragon LLVM), the patch should create a separate derived class, rather than patching over existing classes.
3. patch must not break other existing build environment, e.g. NVIDIA nvcc toolchain, as shown in the TravisCI results below.

mensinda:
@antonysigma Yes, you are correct about the requirements for the PR. Sadly, it happens from time to time that PRs are forgotten about. If there is no response in (roughly) over 2 weeks it is OK to write a ping comment or ask in #mesonbuild for a review.

Also, having a green CI increases the chances of a review :)

Sadly I don't know enough about the meson compiler infrastructure to comment on whether or not the approach here is correct. @dcbaker should know more here. However, there is also #3047 which tracks the current support (or lack thereof) of linker scripts in meson.

antonysigma:
Thank you @mensinda for pointing out the related issue #3047. Yes definitely would like to know if there's any elegant solution to the problem by updating the Meson syntax. I will contribute my thoughts there.

@a1xndr I have rebased my commits. Please feel free to check it out and test it on your build environment.

bonzini:
This has the problem that it breaks the libraries in multiple groups. Since the PR ccannot work as is, and is old, I'm closing it.

PR #7513, opened 2020-07-29
Fix wrap file and machine file syntax inconsistency

xclaesse:
commit b6365eb7edf893cb64c3f120370984212109485f

    doc: Update wrap file format

commit fe1d3eeb7bc83aa7f17da044375366dc80eed644

    wrap: Add special meson.wrap file
    
    That file is used as base for every other wrap file. It can be used for
    example to have every subproject version in a central place.

commit 2a239738904daa904e0162178982bfbb4065d419

    wrap: Allow same syntax as machine files
    
    For consistency it is better to quote values in wrap files and use
    python style lists. It also allows using + and / operators as well as
    [constants] section which is useful for example to write the project
    version in a single place and use it in both source_url and
    fallback_source_url.
    
    For backward compatibility a legacy mode is added to
    ConfigFileInterpreter that disable completely value parsing if it
    fails to parse the first value (e.g. it is unquoted).

commit 9f049c6cd03e54a3b79a014f0d417c565eb2a0f3

    coredata: Move file parsing from ConfigFileInterpreter to ConfigFileParser
    
    ConfigFileParser is responsible to parse files and
    ConfigFileInterpreter is responsible to interpret values.

commit 24582c4dab763cc74a11a64f2b71db2278116ec7

    coredata: Rename CmdLineFileParser and MachineFileParser
    
    They will be used for all kind of config files for consistency,
    including wrap files.


docs/markdown/Wrap-dependency-system-manual.md
```diff
@@ -47,24 +47,35 @@ Currently Meson has four kinds of wraps:
...
 of wrap, followed by properties describing how to obtain the sources, validate
-them, and modify them if needed. An example wrap-file for the wrap named
-`libfoobar` would have a filename `libfoobar.wrap` and would look like this:
+them, and modify them if needed.
+
+Since *0.56.0* the same syntax as [machine files](Machine-files.md) is supported
+and recommended for values:
+- Quoted string values (e.g. `directory='libfoobar-1.0'`).
+- Python-style list values (e.g. `program_names=['prog1', 'prog2']`)
+- `+` and `/` operators and `[constants]` section (e.g. `source_url='http://example.com' / project_name + 'tar.gz'`)
```
eli-schwartz, line 56: Does the `/` filename operator work as expected when used with *urls* on Windows?

And `tar.gz` should be `.tar.gz`.

xclaesse: Argh, indeed it fails on Windows CI: `AssertionError: 'http://mirror.invalid\\project-1.0.tar.gz' != 'http://mirror.invalid/project-1.0.tar.gz'`.

xclaesse: It is actually a bug in machine files too, '/' operator was never supposed to add '\'. interpreterbase.py does that for the / operator in meson.build files: `os.path.join(*args).replace('\\', '/')`.

Fixed both.

xclaesse: Actually replaced in both places using posixpath.join() since that will preserve any existing `\` char in the path.

eli-schwartz: It's not clear to me that there's a benefit for urls vs. just hardcoding the `/` character and using `+`. Paths are logically justified specifically and solely because cross-platform handling has nuances above and beyond "just do string concatenation".

IMHO, YMMV.

xclaesse: It has the cosmetic advantage that `'foo/' / 'bar'` and `'foo' / 'bar'` both yields the same string `'foo/bar'`, avoiding the double slash.

xclaesse: and when used with variables, `foo / bar` is nicer than `foo + '/' + bar`

xclaesse: (unsolved this thread because it's unclear if we can use posixpath.join() here)

xclaesse: Ok, reverted to use `os.path.join(l, r).replace('\\', '/')`. It is of course not fool proof for URLs but works fine in simple cases.

xclaesse:
Btw, one think I would like to do when combined with https://github.com/mesonbuild/meson/pull/6968 is having the main project's meson.wrap be composed too, in that order: `[subproject_meson_wrap, main_meson_wrap, 'project.wrap']`. That way, if the subproject's subproject uses a constant for the wrap's revision it can be overriden by main project without copying the whole .wrap.

xclaesse:
CI failure is not related.

jpakkane:
The `meson.wrap` commit seems to be unrelated, at least it does not match the MR description. Maybe it should be its own MR?

xclaesse:
> The `meson.wrap` commit seems to be unrelated, at least it does not match the MR description. Maybe it should be its own MR?

You're right, removed that part now, I'll open another PR once this one lands.

jpakkane:
This change will affect wrapd, so pinging @legeana.

legeana:
If I understand this correctly this will most definitely not work without https://github.com/mesonbuild/mesonwrap support as vanilla https://docs.python.org/3/library/configparser.html is used to parse `upstream.wrap` files there.

xclaesse:
Note that old syntax is still supported, should be fine if wrapdb is not ported right away.

legeana:
We need to make it clear that the new syntax can't be used right now.

This new format, is it something custom to Meson? If so we need to make the parsing code public (from the API perspective) to the mesonwrap project and import/use it from there directly. The alternative of reimplementing this logic doesn't sound good to me. For instance this is the parsing code I implemented: https://github.com/mesonbuild/mesonwrap/blob/master/mesonwrap/ini.py and it's not compatible.

xclaesse:
technically its still a configparser compliant format. It's just the values that gets extra parsing to evaluate quoted strings, lists, etc. 

Do you actually need to read wrap files? I think wrapdb would only generate them?

legeana:
Example: https://github.com/mesonbuild/sdl2_ttf/blob/2.0.12/upstream.wrap
Resulting file: https://wrapdb.mesonbuild.com/v1/projects/sdl2_ttf/2.0.12/1/get_wrap

At bare minimum `mesonwrap` adds certain lines to the file. Since `[provides]` section was introduced this can no longer be achieved with a string concatenation as new lines can be added to the wrong section.

Secondly `mesonwrap` doesn't blindly accept a pull request, there are certain checks that run and verify that the submission makes sense. You can take a look at this code here: https://github.com/mesonbuild/mesonwrap/blob/master/mesonwrap/tools/reviewtool.py It should be able to extract checksum, source file, etc.

The point is whatever solution is introduced it must have 2 properties:
1. The format must be parseable and generatable programmatically.
2. The code that does that must be accessible by the `mesonwrap` project. Currently it is https://docs.python.org/3/library/configparser.html but if there is a custom implementation in the core Meson project we have to introduce a dependency.

On the (2) point I think it's a good idea already. There are certain things that I would love to use in `mesonwrap` but are hard to directly include. For instance I had to reimplement the code that downloads and parses the wrap. Since the goal of doing this is to catch non-functional wrap it would have been better to use the original code.

What do you think?

eli-schwartz:
Seems like it shouldn't be a big problem to have mesonwrap depend on meson and import a parser with a stable API from `mesonbuild.*`

legeana:
I am happy with this idea as long as we declare the API stable, or rather maintained within the https://github.com/mesonbuild scope.
Does this API supports modifying and converting the structure back to string?

tristan957:
I really like this PR. What would it take to finish it?

I agree that the current dependency_names/program_names is gross compared to `['prog1', 'prog2']`

PR #7517, opened 2020-07-30
Interpreter: Fix c_stdlib usage

xclaesse:
- Exceptions raised during subproject setup were ignored.
- Allow c_stdlib in native file, was already half supported.
- Eliminate usage of subproject variable name by overriding
  '<lang>_stdlib' dependency name.


mesonbuild/interpreter.py
```diff
@@ -2567,21 +2567,25 @@ def get_variables(self):
         return self.variables
 
     def check_stdlibs(self):
-        for for_machine in MachineChoice:
+        machine_choices = [MachineChoice.HOST]
+        if self.coredata.is_cross_build():
+            machine_choices.append(MachineChoice.BUILD)
+        for for_machine in machine_choices:
```
dcbaker, line 2573: not something you need to fix, but we should probably have a `coredata.for_machines()` method that iterates all the valid machines in a build, I seem to remember a lot of code like this around.

xclaesse:
Note that this adds `force_fallback` kwarg for dependency() but atm only for internal usage since it's not listed in `permittedKwargs`. It could be interesting to expose that kwarg in the future?

mesonbuild/interpreter.py
```diff
@@ -2567,21 +2567,25 @@ def get_variables(self):
...
-                    self.build.stdlibs.host[l] = subproj.get_variable_method([depname], {})
                 except KeyError:
-                    pass
-                except InvalidArguments:
-                    pass
+                    continue
+                if len(di) == 1:
+                    FeatureNew.single_use('stdlib without variable name', '0.56.0', self.subproject)
+                kwargs = {'fallback': di,
+                          'native': for_machine == MachineChoice.BUILD,
```
dcbaker, line 2583: you should use `is` and `is not` with enums instead of `==` and `!=`

mesonbuild/interpreter.py
xclaesse: Yeah, thought the same and was surprised to not find it.

run_unittests.py
```diff
@@ -5029,6 +5029,27 @@ def test_wrap_git(self):
             self.build()
             self.run_tests()
 
+    def test_nostdlib(self):
```
dcbaker, line 5032: this should be in the `NativeFileTests` object, which has helper methods for writing machine files.

mesonbuild/interpreter.py
xclaesse: Changed, but it's not the only place it uses ==, does it make a difference?

run_unittests.py
xclaesse: It does both native and cross. tbh this test is trivial enough to not need any helper. You can fake cross build with empty cross file, and it cross build to the same platform.

xclaesse: (was leaking the tmp file, there are other tests doing that... anyway, fixed this one)

dcbaker: The helpers though don't used NamedTemprorary file which has all kinds of caveats on windows and we shouldn't be using.

xclaesse: Well, if that doesn't work, no sure how our CI works then...
```
meson$ git grep "tempfile.TemporaryDirectory" | wc -l
30
```

That said, thinking about it I could write the file into builddir instead.

dcbaker: At least use the `temp_filename` helper? I'd be happy with that. It doesn't have the caveats that NamedTemporaryFile does.

xclaesse: Already updated to use file in build dir ;)

xclaesse:
CI failure is not related.

xclaesse:
Thanks!

PR #7518, opened 2020-07-30
Meson AppImage releases

mensinda:
This PR adds an automatic [AppImage](https://appimage.org) builder for meson. Releasing with AppImages would allow us to support ancient operating systems (Ubuntu 12.04 for instance) far longer, while still working with newer dependencies (mostly python). It could also help with bootstrapping and broken python installations.

The resulting AppImage should work on almost all systems.

Included dependencies:
- python 3.8
- ninja 1.10.0
- CMake 3.18.0
- pkg-config 0.29.2

 The AppImage produced by this PR can be [downloaded here](https://github.com/mensinda/meson/releases/download/BEST_RELEASE/Meson.AppImage) for testing.

Stuff to address in followup PRs:
- cross-compiling Python, etc. to arm, riscv, etc.
- maybe a custom runtime
- documentation
- running project tests on the AppImage

Note: The required [`appimagetool`](https://github.com/AppImage/AppImageKit/releases) isn't automatically downloaded and has to be manually passed as a command-line flag.


mesonbuild/environment.py
```diff
@@ -160,18 +160,31 @@ def detect_ninja(version: str = '1.7', log: bool = False) -> str:
...
+    # Always use AppImage ninja (if running in AppImage mode)
+    appimage = os.environ.get('MESON_AppRun_APPIMAGE', None)
+    if appimage:
+        ninja = [appimage, '--apprun-ninja']
+        vers = check_ninja_vers(ninja)
+        if vers:
+            return (ninja, vers)
+        return None
+
+    env_ninja = os.environ.get('NINJA', None)
```
lygstate, line 184: If more NINJA wrapper are coming, here would be more and more complicated

mesonbuild/mdist.py
```diff
@@ -184,15 +184,15 @@ def run_dist_steps(meson_command, unpacked_src_dir, builddir, installdir, ninja_
     if subprocess.call(meson_command + ['--backend=ninja', unpacked_src_dir, builddir]) != 0:
         print('Running Meson on distribution package failed')
         return 1
-    if subprocess.call([ninja_bin], cwd=builddir) != 0:
+    if subprocess.call(ninja_bin, cwd=builddir) != 0:
```
lygstate, line 187: Not a bin anymore, it's a list

jpakkane:
The main conceptual question is whether these Appimages are going to be "best effort" kinda things or something we intend to officially support. The latter adds to the maintenance burden.

eli-schwartz:
What's the advantage of this over some python freeze program (e.g. pyinstaller) to create a meson executable, then expect the user to use the ninja standalone binaries available from the ninja project, and the cmake binaries from the cmake project? Especially cmake is probably better off distributed by kitware, and would pointlessly bloat the binary release for users who don't want or need cmake.

jpakkane:
To get things rolling, how about the following:

- remove CMake from the image
- instead of us providing AppImages (with all the maintenance burden that entails) instead start by saying something like "if you need an AppImage of Meson for use on old system, you can generate it yourself by running the script <scriptname> and copying the end result <somefile>"

Than later if there is real world usage and requirements we can expand on this?

mensinda:
I have updated it to no longer include CMake and there are now also some major changes to the mechanism itself:

- Patching software to work around `LD_LIBRARY_PATH` is no longer needed (we are using rpaths now with `$ORIGIN`)
- The runtime now extracts itself into `build/meson-runtime` on the first `meson setup` call. All future calls now also use the extracted runtime.
- We are now using a custom -- fully statically linked -- AppImage runtime based on my [libRuntime](https://github.com/mensinda/libRuntime)
- I have tested the generated runtime / "AppImage" and it runs on Ubuntu 10.04
- An Alpine Linux docker image is used to generate the runtime, since Alpine uses [musl](https://musl.libc.org) instead of glibc.

You can try the current x86_64 AppImage / Meson runtime here: [runtime.tar.gz](https://github.com/mesonbuild/meson/files/6941798/runtime.tar.gz) (zipped, since GitHub won't let me upload ist otherwise)

mensinda:
> instead of us providing AppImages (with all the maintenance burden that entails) instead start by saying something like "if you need an AppImage of Meson for use on old system, you can generate it yourself by running the script and copying the end result "

The main point of providing AppImages / Meson runtimes is that it is convenient. You can run the script yourself, but then you have to install all the `-dev` dependencies for python and the wait until python compiles from source. If we provide the script, we might as well provide the AppImage / runtime as well.

jpakkane:
The requirements for this are pretty much:

- There is an actual need for these binaries
- Building one on release is a matter of running at most one command
- Someone volunteers to take on the ongoing maintenance burden

If all three are met then sure, we can provide these.

mensinda:
1. Basically everyone who complains when we bump our min Python version requirements
2. This is already the case. Just run the shell script next to the build.py. It will take a bit though, since we are building python from source
3. That would be me.

Another big advantage we get out of this is that we might be able to switch to newer python versions sooner because we now have a convienient, standalone alternative.

On 6 August 2021 00:08:50 CEST, Jussi Pakkanen ***@***.***> wrote:
>The requirements for this are pretty much:
>
>- There is an actual need for these binaries
>- Building one on release is a matter of running at most one command
>- Someone volunteers to take on the ongoing maintenance burden
>
>If all three are met then sure, we can provide these.
>
>-- 
>You are receiving this because you authored the thread.
>https://github.com/mesonbuild/meson/pull/7518#issuecomment-893839263
-- 

mensinda:
> Building one on release is a matter of running at most one command

I have added a GitHub actions release script, so it now is just a case of waiting 15min. See https://github.com/mensinda/meson/releases/tag/0.0.13-runtime and https://github.com/mensinda/meson/runs/3269603623?check_suite_focus=true

jpakkane:
Please rebase so CI is rerun.

mensinda:
The LGTM alert is not really helpful in this case... (Maybe we should add a UUID module to meson :D)

eli-schwartz:
> 1. Basically everyone who complains when we bump our min Python version requirements

I *still* think this could and should be solved the same way we do it for Windows... run the PyInstaller command to create an executable that unpacks its own copy of a) the python interpreter, b) the packaged module (mesonbuild).

What does AppImage bring as an advantage, other than bundling its own copy of ninja, pkg-config, and cmake? None of which have to do with python, and:
- ninja is already available in a fairly stable manner unlike python+meson woes,
- pkg-config is certainly available on the distro already, so this amounts to trying to insist that specifically pkgconf is used or something?
- cmake is big and fat and Kitware already provides their own standalone binaries, which most meson users don't need anyway and has been taken off the table for this discussion.

This AppImage is trying to do too much, and the actual stated advantages of it are only advantages over "just use your OS python", not over "pyinstaller binaries".

Using pyinstaller would have the advantage of it being architecturally identical to a `meson.pyz` zipapp or the existing Windows PyInstaller frozen binary, and not having to maintain 25 (?) new files to maintain custom infrastructure, a desktop file (???), a custom AppImage runtime maintained in-house, etc.

I would love to see a PyInstaller Linux build of meson, but I really think AppImage is too complicated by design.

eli-schwartz:
```
$ ./meson.runtime setup --wipe /home/eschwartz/git/archlinux/pacman/stable builddir/
Using exsisting meson-runtime in builddir//meson-runtime
The Meson build system
Version: 0.59.99
Source dir: /home/eschwartz/git/archlinux/pacman/stable
Build dir: /tmp/builddir
Build type: native build
  File "/tmp/builddir/meson-runtime/usr/lib/python3.9/site-packages/mesonbuild/mesonmain.py", line 227, in run
  File "/tmp/builddir/meson-runtime/usr/lib/python3.9/site-packages/mesonbuild/msetup.py", line 299, in run
  File "/tmp/builddir/meson-runtime/usr/lib/python3.9/site-packages/mesonbuild/msetup.py", line 202, in generate
  File "/tmp/builddir/meson-runtime/usr/lib/python3.9/site-packages/mesonbuild/msetup.py", line 219, in _generate
  File "/tmp/builddir/meson-runtime/usr/lib/python3.9/site-packages/mesonbuild/interpreter/interpreter.py", line 337, in __init__
  File "/tmp/builddir/meson-runtime/usr/lib/python3.9/site-packages/mesonbuild/interpreterbase/interpreterbase.py", line 104, in load_root_meson_file
  File "/tmp/builddir/meson-runtime/usr/lib/python3.9/site-packages/mesonbuild/mparser.py", line 530, in parse
  File "/tmp/builddir/meson-runtime/usr/lib/python3.9/site-packages/mesonbuild/mparser.py", line 810, in codeblock
  File "/tmp/builddir/meson-runtime/usr/lib/python3.9/site-packages/mesonbuild/mparser.py", line 804, in line
  File "/tmp/builddir/meson-runtime/usr/lib/python3.9/site-packages/mesonbuild/mparser.py", line 535, in statement
  File "/tmp/builddir/meson-runtime/usr/lib/python3.9/site-packages/mesonbuild/mparser.py", line 545, in e1
  File "/tmp/builddir/meson-runtime/usr/lib/python3.9/site-packages/mesonbuild/mparser.py", line 508, in accept
  File "/tmp/builddir/meson-runtime/usr/lib/python3.9/site-packages/mesonbuild/mparser.py", line 499, in getsym
  File "/tmp/builddir/meson-runtime/usr/lib/python3.9/site-packages/mesonbuild/mparser.py", line 206, in lex
  File "/tmp/builddir/meson-runtime/usr/lib/python3.9/site-packages/mesonbuild/mparser.py", line 44, in decode_match
LookupError: unknown encoding: unicode_escape
$ ls builddir
meson-info/  meson-logs/  meson-private/
$ ./meson.runtime setup --wipe /home/eschwartz/git/archlinux/pacman/stable builddir/
The Meson build system
Version: 0.59.99
Source dir: /home/eschwartz/git/archlinux/pacman/stable
Build dir: /tmp/builddir
Build type: native build
  File "/tmp/builddir/meson-runtime/usr/lib/python3.9/site-packages/mesonbuild/mesonmain.py", line 227, in run
  File "/tmp/builddir/meson-runtime/usr/lib/python3.9/site-packages/mesonbuild/msetup.py", line 299, in run
  File "/tmp/builddir/meson-runtime/usr/lib/python3.9/site-packages/mesonbuild/msetup.py", line 202, in generate
  File "/tmp/builddir/meson-runtime/usr/lib/python3.9/site-packages/mesonbuild/msetup.py", line 219, in _generate
  File "/tmp/builddir/meson-runtime/usr/lib/python3.9/site-packages/mesonbuild/interpreter/interpreter.py", line 337, in __init__
  File "/tmp/builddir/meson-runtime/usr/lib/python3.9/site-packages/mesonbuild/interpreterbase/interpreterbase.py", line 104, in load_root_meson_file
  File "/tmp/builddir/meson-runtime/usr/lib/python3.9/site-packages/mesonbuild/mparser.py", line 530, in parse
  File "/tmp/builddir/meson-runtime/usr/lib/python3.9/site-packages/mesonbuild/mparser.py", line 810, in codeblock
  File "/tmp/builddir/meson-runtime/usr/lib/python3.9/site-packages/mesonbuild/mparser.py", line 804, in line
  File "/tmp/builddir/meson-runtime/usr/lib/python3.9/site-packages/mesonbuild/mparser.py", line 535, in statement
  File "/tmp/builddir/meson-runtime/usr/lib/python3.9/site-packages/mesonbuild/mparser.py", line 545, in e1
  File "/tmp/builddir/meson-runtime/usr/lib/python3.9/site-packages/mesonbuild/mparser.py", line 508, in accept
  File "/tmp/builddir/meson-runtime/usr/lib/python3.9/site-packages/mesonbuild/mparser.py", line 499, in getsym
  File "/tmp/builddir/meson-runtime/usr/lib/python3.9/site-packages/mesonbuild/mparser.py", line 206, in lex
  File "/tmp/builddir/meson-runtime/usr/lib/python3.9/site-packages/mesonbuild/mparser.py", line 44, in decode_match
LookupError: unknown encoding: unicode_escape
```

It seems completely broken with --wipe, which deletes the entire directory and then the AppImage runtime tries to load itself from there anyway and falls over.

Then once there is nothing there at all, it still crashes, with the same traceback but without mentioning "Using exsisting meson-runtime in builddir//meson-runtime" (typo intentional), but I am not sure why it wants to use something nonexistent...

Trying to manage the lifecycle of a temporary extracted runtime strikes me as a bad move. If we want to have an extracted version of the standalone program, this is a long-ago solved problem: ship both python and mesonbuild/ in a tarball and ask users to extract it into `/opt/meson/`

mensinda:
The main advantage of the AppImage is that it contains everything Meson needs to run and that it is runnable pretty much everywhere (including ancient distros) because the runtime is a statically linked binary and it bundles libc (so literally *everything* needed to run meson). pyinstaller binaries still depend on the host libc.

Additionally, we need a stable way to invoke the python from ninja, since we guarantee that it exists (and by also with the min required python version from Meson). I currently don't know how to achieve this with pyinstaller (though I also did not look into it in too much detail).

> a desktop file (???)

Yeah, required by AppImage...

I am not denying that AppImages (and specifically this solution) is more complicated, but it is also more user-friendly.

mensinda:
> If we want to have an extracted version of the standalone program, this is a long-ago solved problem: ship both python and mesonbuild/ in a tarball and ask users to extract it into /opt/meson/

Except that you now need root privileges to install meson. Thanks for pointing out the bug with `--wipe`.

mensinda:
As an alternative: We could drop the runtime and AppImage part and just ship the AppDir (without the special *.desktop stuff). Then the users just needs to add `fakebin/` to PATH or call `fakebin/meson`.

You can currently easily get the same result for testing if you run `--runtime-setup <somewhere>`. This would still have all the advantages of the AppImage but is less user-friendly.

eli-schwartz:
> Additionally, we need a stable way to invoke the python from ninja, since we guarantee that it exists (and by also with the min required python version from Meson). I currently don't know how to achieve this with pyinstaller (though I also did not look into it in too much detail).

Isn't this just `meson runpython` ?

> because the runtime is a statically linked binary and it bundles libc

The traditional Linux way to be super compatible with everything is to build on the oldest libc you intend to support, or to make it an easy choice, build on CentOS 7.

Another option should be, since IIRC pyinstaller bundles up the python interpreter you ran it with -- run pyinstaller from a custom built python which is statically linked to glibc, in Modules/Setup it appears you can tell python to include all modules in the interpreter directly rather than lib-dynload.

eli-schwartz:
> Except that you now need root privileges to install meson.

You can optionally choose to install it to ~/opt/meson instead, and not need root. I don't know why people (this is not the first time I've seen that objection) persist in somehow thinking that a ***standalone binary*** which for convention and convenience gets installed to /opt (and symlinked to /usr/bin) so that it's convenient for all users on a system to access, must therefore ***require*** root as a mandatory part of the installation process, without which the standalone binary will crash.

mensinda:
> Isn't this just meson runpython ?

Then we have to teach meson to only generate rules that use `meson runpython` for invoking custom targets.

> The traditional Linux way to be super compatible with everything is to build on the oldest libc you intend to support, or to make it an easy choice, build on CentOS 7.

That's also how AppImages are usually built, but I wanted to support as many systems as possible (even those with a completely different libc (musl for instance)).

> Another option should be, since IIRC pyinstaller bundles up the python interpreter you ran it with -- run pyinstaller from a custom built python which is statically linked to glibc, in Modules/Setup it appears you can tell python to include all modules in the interpreter directly rather than lib-dynload.

That would be great. Do you know the configure flag because I couldn't find a way to do this...

eli-schwartz:
> That would be great. Do you know the configure flag because I couldn't find a way to do this...

edit the file `Modules/Setup`

Add to the top: `*static*`

Start uncommenting lines describing individual modules, e.g. 

```diff
diff --git a/Modules/Setup b/Modules/Setup
index 02cfb67518..a6b7873d98 100644
--- a/Modules/Setup
+++ b/Modules/Setup
@@ -1,4 +1,5 @@
 # -*- makefile -*-
+*static*
 # The file Setup is used by the makesetup script to construct the files
 # Makefile and config.c, from Makefile.pre and config.c.in,
 # respectively.  Note that Makefile.pre is created from Makefile.pre.in
@@ -179,13 +180,13 @@ _symtable symtablemodule.c
 #_testinternalcapi _testinternalcapi.c -I$(srcdir)/Include/internal -DPy_BUILD_CORE_MODULE  # Python internal C API test module
 #_random _randommodule.c -DPy_BUILD_CORE_MODULE	# Random number generator
 #_elementtree -I$(srcdir)/Modules/expat -DHAVE_EXPAT_CONFIG_H -DUSE_PYEXPAT_CAPI _elementtree.c	# elementtree accelerator
-#_pickle _pickle.c	# pickle accelerator
+_pickle _pickle.c	# pickle accelerator
 #_datetime _datetimemodule.c	# datetime accelerator
 #_zoneinfo _zoneinfo.c	# zoneinfo accelerator
 #_bisect _bisectmodule.c	# Bisection algorithms
 #_heapq _heapqmodule.c	# Heap queue algorithm
 #_asyncio _asynciomodule.c  # Fast asyncio Future
-#_json -I$(srcdir)/Include/internal -DPy_BUILD_CORE_BUILTIN _json.c	# _json speedups
+_json -I$(srcdir)/Include/internal -DPy_BUILD_CORE_BUILTIN _json.c	# _json speedups
 #_statistics _statisticsmodule.c # statistics accelerator
 
 #unicodedata unicodedata.c    # static Unicode character database
@@ -211,10 +212,10 @@ _symtable symtablemodule.c
 
 # Socket module helper for SSL support; you must comment out the other
 # socket line above, and possibly edit the SSL variable:
-#SSL=/usr/local/ssl
-#_ssl _ssl.c \
-#	-DUSE_SSL -I$(SSL)/include -I$(SSL)/include/openssl \
-#	-L$(SSL)/lib -lssl -lcrypto
+SSL=/usr/local/ssl
+_ssl _ssl.c \
+	-DUSE_SSL -I$(SSL)/include -I$(SSL)/include/openssl \
+	-L$(SSL)/lib -lssl -lcrypto
 
 # The crypt module is now disabled by default because it breaks builds
 # on many systems (where -lcrypt is needed), e.g. Linux (I believe).
```

run `make python`, and the resulting ./python module has a few things that used to be `lib-dynload/_{foo}.cpython-39-x86_64-linux-gnu.so` which are now builtin, have no __file__ attribute, and don't need to be copied around or dlopened.

eli-schwartz:
```diff
From bf547ae1af86f17adda054eb3267f4e7b95fee00 Mon Sep 17 00:00:00 2001
From: Eli Schwartz <eschwartz@archlinux.org>
Date: Sun, 15 Aug 2021 20:00:57 -0400
Subject: [PATCH] first version of python-static which successfully ran meson

built using

./configure --prefix /does/not/exist --disable-shared
make python LIBS=

./python interpreter loads no modules, only .py scripts, ld-linux.so
links to libc.so and libm.so (on versions of glibc < 2.34 where this is
a separate shared lib) but that can be configured away once this is
serious

LIBS= will prevent overlinking to libcrypt.so, which I don't have a
static version of, but again hold off until serious
---
 Modules/Setup | 39 ++++++++++++++++++++-------------------
 1 file changed, 20 insertions(+), 19 deletions(-)

diff --git a/Modules/Setup b/Modules/Setup
index 02cfb67518..f25987d3be 100644
--- a/Modules/Setup
+++ b/Modules/Setup
@@ -1,4 +1,5 @@
 # -*- makefile -*-
+*static*
 # The file Setup is used by the makesetup script to construct the files
 # Makefile and config.c, from Makefile.pre and config.c.in,
 # respectively.  Note that Makefile.pre is created from Makefile.pre.in
@@ -169,36 +170,36 @@ _symtable symtablemodule.c
 
 # Modules that should always be present (non UNIX dependent):
 
-#array arraymodule.c	# array objects
+array arraymodule.c	# array objects
 #cmath cmathmodule.c _math.c -DPy_BUILD_CORE_MODULE # -lm # complex math library functions
-#math mathmodule.c _math.c -DPy_BUILD_CORE_MODULE # -lm # math library functions, e.g. sin()
-#_contextvars _contextvarsmodule.c  # Context Variables
-#_struct _struct.c	# binary structure packing/unpacking
+math mathmodule.c _math.c -DPy_BUILD_CORE_MODULE # -lm # math library functions, e.g. sin()
+_contextvars _contextvarsmodule.c  # Context Variables
+_struct _struct.c	# binary structure packing/unpacking
 #_weakref _weakref.c	# basic weak reference support
 #_testcapi _testcapimodule.c    # Python C API test module
 #_testinternalcapi _testinternalcapi.c -I$(srcdir)/Include/internal -DPy_BUILD_CORE_MODULE  # Python internal C API test module
-#_random _randommodule.c -DPy_BUILD_CORE_MODULE	# Random number generator
+_random _randommodule.c -DPy_BUILD_CORE_MODULE	# Random number generator
 #_elementtree -I$(srcdir)/Modules/expat -DHAVE_EXPAT_CONFIG_H -DUSE_PYEXPAT_CAPI _elementtree.c	# elementtree accelerator
-#_pickle _pickle.c	# pickle accelerator
+_pickle _pickle.c	# pickle accelerator
 #_datetime _datetimemodule.c	# datetime accelerator
 #_zoneinfo _zoneinfo.c	# zoneinfo accelerator
 #_bisect _bisectmodule.c	# Bisection algorithms
 #_heapq _heapqmodule.c	# Heap queue algorithm
 #_asyncio _asynciomodule.c  # Fast asyncio Future
-#_json -I$(srcdir)/Include/internal -DPy_BUILD_CORE_BUILTIN _json.c	# _json speedups
+_json -I$(srcdir)/Include/internal -DPy_BUILD_CORE_BUILTIN _json.c	# _json speedups
 #_statistics _statisticsmodule.c # statistics accelerator
 
-#unicodedata unicodedata.c    # static Unicode character database
+unicodedata unicodedata.c    # static Unicode character database
 
 
 # Modules with some UNIX dependencies -- on by default:
 # (If you have a really backward UNIX, select and socket may not be
 # supported...)
 
-#fcntl fcntlmodule.c	# fcntl(2) and ioctl(2)
+fcntl fcntlmodule.c	# fcntl(2) and ioctl(2)
 #spwd spwdmodule.c		# spwd(3)
 #grp grpmodule.c		# grp(3)
-#select selectmodule.c	# select(2); not on ancient System V
+select selectmodule.c	# select(2); not on ancient System V
 
 # Memory-mapped files (also works on Win32).
 #mmap mmapmodule.c
@@ -207,14 +208,14 @@ _symtable symtablemodule.c
 #_csv _csv.c
 
 # Socket module helper for socket(2)
-#_socket socketmodule.c
+_socket socketmodule.c
 
 # Socket module helper for SSL support; you must comment out the other
 # socket line above, and possibly edit the SSL variable:
-#SSL=/usr/local/ssl
-#_ssl _ssl.c \
-#	-DUSE_SSL -I$(SSL)/include -I$(SSL)/include/openssl \
-#	-L$(SSL)/lib -lssl -lcrypto
+SSL=/usr/local/ssl
+_ssl _ssl.c \
+	-DUSE_SSL -I$(SSL)/include -I$(SSL)/include/openssl \
+	-L$(SSL)/lib -l:libssl.a -l:libcrypto.a
 
 # The crypt module is now disabled by default because it breaks builds
 # on many systems (where -lcrypt is needed), e.g. Linux (I believe).
@@ -229,7 +230,7 @@ _symtable symtablemodule.c
 #termios termios.c	# Steen Lumholt's termios module
 #resource resource.c	# Jeremy Hylton's rlimit interface
 
-#_posixsubprocess _posixsubprocess.c  # POSIX subprocess module helper
+_posixsubprocess _posixsubprocess.c  # POSIX subprocess module helper
 
 # Multimedia modules -- off by default.
 # These don't work for 64-bit platforms!!!
@@ -256,7 +257,7 @@ _symtable symtablemodule.c
 #_sha3 _sha3/sha3module.c
 
 # _blake module
-#_blake2 _blake2/blake2module.c _blake2/blake2b_impl.c _blake2/blake2s_impl.c
+_blake2 _blake2/blake2module.c _blake2/blake2b_impl.c _blake2/blake2s_impl.c
 
 # The _tkinter module.
 #
@@ -329,7 +330,7 @@ _symtable symtablemodule.c
 
 
 # Helper module for various ascii-encoders
-#binascii binascii.c
+binascii binascii.c
 
 # Fred Drake's interface to the Python parser
 #parser parsermodule.c
@@ -338,7 +339,7 @@ _symtable symtablemodule.c
 # Andrew Kuchling's zlib module.
 # This require zlib 1.1.3 (or later).
 # See http://www.gzip.org/zlib/
-#zlib zlibmodule.c -I$(prefix)/include -L$(exec_prefix)/lib -lz
+zlib zlibmodule.c -I$(prefix)/include -L$(exec_prefix)/lib -l:libz.a
 
 # Interface to the Expat XML parser
 # More information on Expat can be found at www.libexpat.org.
-- 
2.32.0
```

With this patch plus a static copy of openssl and perhapsCC=musl-gcc there is a serious thought to meson shipping with a single-file executable version of python, plus the standard library of pure python.

jpakkane:
I tried building the image but it did not seem to work. First it gave this:

```
packaging/appimage/build_x86_64.sh: line 8: [: : integer expression expected
```

Then eventually it failed with this:

```
OK: 283 MiB in 66 packages
adduser: uid '0' in use
sudo: unknown user: meson
sudo: error initializing audit plugin sudoers_audit
packaging/appimage/build_x86_64.sh: line 10: /home/jpakkane/src/appimage/packaging/appimage/../../appimage/package.sh: No such file or directory
```

PR #7521, opened 2020-07-31
Per subproject compiler options

xclaesse:
This PR is based on top of https://github.com/mesonbuild/meson/pull/7574

commit 38d457da78337d25147f69dbc6fb6f8f5c9a6928
Date:   Fri Aug 7 19:01:07 2020 -0400

    tests: Move persubproject test to linuxlike
    
    c++98 is not supported by msvc tests, and it's not testing anything
    platform specific any way.

commit dbd17345e6c96265731c729158d03c8948c0a809
Date:   Fri Jul 31 13:10:14 2020 -0400

    Make compiler options per subproject

commit d248d29bd5a961a722a778d5a7315866fa7224b0
Date:   Fri Jul 31 13:07:11 2020 -0400

    Propagate the subproject for which we take options
    
    This is in preparation to have per subproject compiler options.

commit 51df1a446d39cacfc477e8f3e466c29d32662bfc
Date:   Fri Jul 31 01:00:23 2020 -0400

    coredata: flatten compiler_options dict
    
    This is a revert of https://github.com/mesonbuild/meson/pull/5489 that
    was made for no clear purpose and benefices. This fixes Xc16CCompiler,
    C2000CCompiler, C2000CPPCompiler, CudaCompiler and EmscriptenMixin that
    were not ported.
    
    This is in preparation to add per-subproject compiler options that will
    also include a subproject prefix to option names.


xclaesse:
This replaces https://github.com/mesonbuild/meson/pull/6729.

I really dislike the direction that option handling has taken, especially for compiler options. It became a mess all over the place. Why spliting dictionary and then flatten the lang everywhere... makes no sense. I reverted all that back to having a normal key/value dictionary for compiler options.

Ericson2314:
Highly against all this flattening. I'm sorry, but It's a terrible idea to have less structured data structures.

Ericson2314:
I'm glad you have at least established how the line count can go down. I will hold myself to those -30 lines to justify the increased nesting in future refactors.

xclaesse:
The data structure is not well thought, it's flattening it everywhere, it makes no sense. Better keep it flat from the beginning. It's not only a matter of line count but also readability, many lines are less complex to understand.

dcbaker:
@xclaesse, traditionally on Unix environment variables override config files, no?

I'm not sure I agree that this makes things simpler, it simply moves the complexity around by using Proxies, which is conceptually harder to understand (It took me a while to figure out what your code was doing), and certainly more expensive. like @Ericson2314 I'm not convinced this is the right direction either. It makes more sense to me to make the consumers do a two step lookup than effectively having a two key system in a string. ie: ['c']['std'] vs ['c_std']. The latter means that you can't short circuit the lookup if you already know the language you're using and need multiple values, and it means that we do expensive string construction instead of cheap dict lookups (for references constructing 'c_std' using str.format is roughly half as fast as the double lookup). It feels like trying to make our internal data structures match the command line arguments, instead of putting them in an easy to use format.

xclaesse:
About performance, I would have expected my way to be faster because it costed an extra dict lookup and it was calling _flatten_pair_iterator() everywhere. But after testing, both are exactly identical, at least in the time it takes to configure gst-build with all its subprojects (the biggest meson configure I know). So let's keep that consideration out of the way. I think with small data structures (would be surprised any project has >100 options) whatever you do, it's never going to be a problem.

About using a proxy, we already have that mechanism for option overrides, nothing new there. tbf, I used that because I'm lazy but I could revert all compilers back to use option names that include the language.

Besides the fact that I find my code simpler (that's not an objective measurements) it's also required to support per subproject compiler option. For builtin and project options, the subproject is a prefix "<subproject>:" in the option name, that's assumed in many places all over the places. I'm sure you can find other tricks, adding yet another level of dictionary for the subproject... but IMHO that's getting way too much.

In the end I would like to go the exact opposite direction: flatten all option structures to have a single dictionary in coredata. There are many places where we have to look in each directory to find the option, a single lookup is simpler. We already guarantee that keys are unique, because that's what the user actually type: `-D[<subproject>:][build.][lang_]optname`. But that's for hypothetical future.

> @xclaesse, traditionally on Unix environment variables override config files, no?

Dunno... I don't have a strong opinion either way tbh. But currently we have that: -Dpkg_config_path > PKG_CONFIG_PATH > machine-file. I thing env should be first or last, but not in the middle. It is also inconsistent with -Dc_args > machine-file > CFLAGS.

xclaesse:
Note this PR also fix a few backend bugs where it was not using get_compiler_options_for_target(), so not picking target overrides.

Ericson2314:
Let's please start with something we can all agree on. I opened https://github.com/mesonbuild/meson/issues/7571 to discuss stuff along those lines.

xclaesse:
I splitted part of this PR into https://github.com/mesonbuild/meson/pull/7574, to go step by step.

jpakkane:
One thing that came to mind was that there are clear use cases for not having per-subproject options as well as having them. These may even change depending on the top level project where the subproject is being used. So how about adding a top level option as an array of strings that specify which subprojects should have their own options? By default subprojects would not have their own options. In this way end users are not surprised when some projects have their own settings as they have to manually enable those.

xclaesse:
I think that's related to one thing @dcbaker wanted to address in his PR https://github.com/mesonbuild/meson/pull/6597 that my PR here actually reverts: https://github.com/mesonbuild/meson/pull/7574/commits/f3b8e13883e5e4e4c00500694091568aa5439cb0. I'm not actually sure about the desired behavior, so I opted to revert back to the 0.55.0 behavior, such change should be its own PR if we want to change it.

So, the question is: Should setting option (-D or machine file) on the main project override the default set on subproject? For example if a subproject has `project(..., default_options: ['static_library=static'])` and I configure the main project with `-Dstatic_library=shared`, should the subproject build static or shared? In 0.55 it would build static unless you pass `-Dsubproject:default_library=shared` but currently in master, with @dcbaker changes, it would build shared.

@jpakkane I think if we do like @dcbaker did (currently in master) that would mostly solve your concern, right?

xclaesse:
> I'm not sure I agree that this makes things simpler, it simply moves the complexity around by using Proxies

I was not a big fan if that trick neither, so I updated my commit that flatten lang dict to also revert all compilers to use the full option name, including the lang prefix. While doing that I noticed that Xc16CCompiler, C2000CCompiler, C2000CPPCompiler, CudaCompiler and EmscriptenMixin were broken since https://github.com/mesonbuild/meson/pull/5489.

xclaesse:
> The latter means that you can't short circuit the lookup if you already know the language you're using and need multiple values, and it means that we do expensive string construction instead of cheap dict lookups (for references constructing 'c_std' using str.format is roughly half as fast as the double lookup).

My initial testing was not showing measurable difference, but I was tricked by CPU frequency scaling on my laptop, the variance was too big to make conclusions. On a much more powerful desktop machine, I get more consistent numbers. Comparing current master and this branch, gst-build goes from 20.8s to 20.2s roughly. I don't consider that really meaningful, but that's 3% faster. Dictionaries lookups does have a cost too.

PR #7524, opened 2020-08-03
"Downgrade" warning when regenerating after version bump

keszybz:
There is nothing to "warn" about, this is a completely routine
occurence. OTOH, when something is corrupted, we should warn. Keep
the red color and "WARNING:" prefix in that case.

Example output:
$ ninja -C build
Regenerating configuration from scratch: Build directory has been generated with Meson version 0.55.999, which is incompatible with current version 0.56.0.
The Meson build system
Version: 0.56.0
...


keszybz:
travis failed with "4/5 Boost extralib test FAIL           0.12s (killed by signal 11 SIGSEGV)", looks unrelated.

nirbheek:
lgtm!

Issue #7515, opened 2020-07-29
0.55 considers SKIP in tap output as failed


matthiasclasen:
GTK has a some tests that produce output like

```
1..227
...
ok 57 /FinalizeObject/GtkConstraint
ok 58 /FinalizeObject/GtkConstraintGuide
ok 59 /FinalizeObject/GtkConstraintLayoutChild # SKIP Skipping GtkConstraintLayoutChild
ok 60 /FinalizeObject/GtkConstraintLayout
ok 61 /FinalizeObject/GtkCssProvider
...
ok 227 /FinalizeObject/GtkWindowHandle
```

that is, 227 ok's and an exit code of 0.

And meson 0.55.0 on my local system says:

```
3/3 gtk:gtk+object / objects-finalize FAIL           11.85s (exit status 0)
```
If I shorten the test to leave out all the cases that emit a # SKIP comment, meson 0.55 is happy again.

0.54 in our ci thinks everything is just fine.

I also think that everything with our testsuite is just fine, and meson regressed.

smcv:
Bisection says this regressed in 0c3bb153.

ptomato:
Despite the fix being included in 0.55.1, this still doesn't seem to work. Any skipped tests still cause the test suite to fail. I believe this should be reopened.

ptomato:
Despite the above, it seems to have been fixed eventually in 0.55.3, so :shrug:

nirbheek:
> Despite the above, it seems to have been fixed eventually in 0.55.3, so 🤷

Yes, the discussion happened on https://github.com/mesonbuild/meson/pull/7618.

PR #7525, opened 2020-08-03
mtest: TestResult.SKIP is not a failure

smcv:
If some but not all tests in a run were skipped, then the overall result
is given by whether there were any failures among the non-skipped tests.

Resolves: https://github.com/mesonbuild/meson/issues/7515


jpakkane:
This needs a test. Can you get one of the original test failure logs, create a test executable that just prints that and then turn that into a test?

smcv:
> This needs a test. Can you get one of the original test failure logs, create a test executable that just prints that and then turn that into a test?

I added a minimal reproducer inline in `meson.build`, and a more realistic example using one of the GNOME tests that regressed like you said (the report of #7515 was for GTK, but the corresponding Debian bug was for gjs which compiles a lot more quickly, so I used a gjs test). Feel free to delete one or the other if you don't want both.

Instead of an executable that prints the output, I added a simple implementation of cat(1) and a text file with the output, so that it can be reused for other TAP sequences.

smcv:
Rebased on master to avoid CI failure (#7535).

PR #7529, opened 2020-08-03
Remove uneeded `parse_cmd_line_options`

Ericson2314:
The new way of doing things, thanks to @dcbaker, is making it so
`Environment` combines the command line options and config files, and
then coredata and just ingestion 1 source of raw data to initialize the
actual options. This dramatically simplifies things by making
information flow through one path not many.

`parse_cmd_line_options` is a vestige of the old way of coredata having
to crawl over a bunch of different sources of information and repeatedly
parse them, and is no longer needed.

CC @dcbaker @xclaesse

### DIFF
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -1023,24 +1023,6 @@ def create_options_dict(options):
         result[key] = value
     return result
 
-def parse_cmd_line_options(args):
-    args.cmd_line_options = create_options_dict(args.projectoptions)
-
-    # Merge builtin options set with --option into the dict.
-    for name in chain(
-            builtin_options.keys(),
-            ('build.' + k for k in builtin_options_per_machine.keys()),
-            builtin_options_per_machine.keys(),
-    ):
-        value = getattr(args, name, None)
-        if value is not None:
-            if name in args.cmd_line_options:
-                cmdline_name = BuiltinOption.argparse_name_to_arg(name)
-                raise MesonException(
-                    'Got argument {0} as both -D{0} and {1}. Pick one.'.format(name, cmdline_name))
-            args.cmd_line_options[name] = value
-            delattr(args, name)
-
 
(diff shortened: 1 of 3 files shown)


xclaesse:
Did you even read that code? grep where `cmd_line_options` is used?

Ericson2314:
OK no i didn't realize this thing mutated it's argument so something else could access the field. Still that's an absolute rediculous convoluted approach, and whatever bits of this code are sill needed should be moved environment where the CLI and config files are combined.

xclaesse:
I agree that function should not modify options but rather return the dict. It could eventually be moved into Environment too, but would have to check the places it's used first.

xclaesse:
FYI, I know you won't like it, but I'm refactoring the changes from @dcbaker there: https://github.com/mesonbuild/meson/pull/7521/commits/709054218d2c1505fc40bb6eb2fe11bd8ca8f521. We should please take a step back and find the most maintainable code possible to handle options. It is currently a giant mess that takes days of work to figure out how every part works.

Ericson2314:
@xclaesse I agree it's not good right now, but I have read that and I am still think you are taking the code in the wrong direction---a local optimum but we should hop further afield to find a better optimum.

One of the the underlying issue is the division of labor between `environment` and `coredata` is currently a complete unprincipled mess. Any long-term plan should fix that.

xclaesse:
> One of the the underlying issue is the division of labor between `environment` and `coredata` is currently a complete unprincipled mess. Any long-term plan should fix that.

With the change I'm proposing there, the separation is Environment handles "raw" options, in the format as typed by the user in command line, default_options, etc (i.e. a single str:str dictionary). coredata is then responsible to parse those raw options into the proper data structure.

Ericson2314:
Well it sounds like there are some things we are all agree on which we should do first before any struct flattening.

PR #7539, opened 2020-08-04
avoid spurious warning from get_target_filename

bonzini:
In the situation covered by the new test case, get_target_filename is
invoked in the process of retrieving (via get_target_private_dir) the
custom target's @BUILD_DIR@; because the custom target has more than
one output, a spurious warning is printed.

However, that substitution is not used in the commands.  Avoid the
problem by not invoking get_target_private_dir.

This seems to be a regression from 0.54.3 to 0.55.0.

Signed-off-by: Paolo Bonzini <pbonzini@redhat.com>

### DIFF
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -1955,9 +1955,11 @@ https://gcc.gnu.org/bugzilla/show_bug.cgi?id=47485'''))
             source_target_dir = os.path.join(self.build_to_src, override_subdir)
         else:
             source_target_dir = self.get_target_source_dir(target)
-        relout = self.get_target_private_dir(target)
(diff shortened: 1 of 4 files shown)


nirbheek:
Wasn't this already fixed by https://github.com/mesonbuild/meson/pull/7494?

nirbheek:
The test case is still probably worth committing though. Needs an update, however.

bonzini:
Yes it should . Is that series marked for 0.55.1? I will rebase and include the test case only.

jpakkane:
Yes. You can see the milestone assignment on the right side column of each MR page.

nirbheek:
I've extracted the test into https://github.com/mesonbuild/meson/pull/7583. Closing.

PR #7546, opened 2020-08-05
cmake: Use the DEBUG config when linking to the debug CRT

nirbheek:
The `debug` builtin option does not control whether or not the debug CRT is used. Without this fix, when `buildtype=debugoptimized` or when `debug=true` + `b_vscrt=md`, we will try to link to the debug libraries found via cmake while linking with `/release`, which will cause a link failure.

CC: @mensinda

### DIFF
--- a/mesonbuild/dependencies/base.py
+++ b/mesonbuild/dependencies/base.py
@@ -1450,7 +1450,10 @@ class CMakeDependency(ExternalDependency):
                     cfgs = [x for x in tgt.properties['IMPORTED_CONFIGURATIONS'] if x]
                     cfg = cfgs[0]
 
-                is_debug = self.env.coredata.get_builtin_option('debug');
+                is_debug = self.env.coredata.get_builtin_option('buildtype') == 'debug'
+                if 'b_vscrt' in self.env.coredata.base_options:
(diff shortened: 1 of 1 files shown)


mesonbuild/dependencies/base.py
```diff
@@ -1450,7 +1450,10 @@ def _detect_dep(self, name: str, modules: T.List[T.Tuple[str, bool]], components
                     cfgs = [x for x in tgt.properties['IMPORTED_CONFIGURATIONS'] if x]
                     cfg = cfgs[0]
 
-                is_debug = self.env.coredata.get_builtin_option('debug');
+                is_debug = self.env.coredata.get_builtin_option('buildtype') == 'debug'
```
mensinda, line 1453: Shouldn't this still be `'debug'` instead of `'buildtype'`?

mensinda: also, this would currently miss `debugoptimized`

nirbheek: No, like I said in my commit message, `debugoptimized` uses the release CRT unless you override with `b_vscrt`. This is documented: https://mesonbuild.com/Builtin-options.html#b_vscrt-from_buildtype

This code is identical to what we do when detecting the Qt dependency and elsewhere.

PR #7556, opened 2020-08-06
msvc: enable /std:c11 flag

### LINKED ISSUE #7554: Support c_std=c11 for newer MSVC 2019
Although I haven't found any mention of this in the [16.7 release notes](https://docs.microsoft.com/en-us/visualstudio/releases/2019/release-notes#16.7.0) `/std:c11` (and only `c11`) is supported for this version, this enables C11 _Generic support which has been a [preview feature](https://developercommunity.visualstudio.com/comments/1039772/view.html) for a while.

scivision:
fixes #7554

Dylan, should I be copying over the "winlibs" from get_options in VisualStudioLikeCCompilerMixin? 

Before this change, a trivial meson.build + main.c would issue:

```sh
[1/2] "cl" "-Imain.exe.p" "-I." "-I.." "/MDd" "/nologo" "/showIncludes" "/W2" "/ZI" "/Ob0" "/Od" "/RTC1" "/Fdmain.exe.p\main.c.pdb" /Fomain.exe.p/main.c.obj "/c" ../main.c
[2/2] "link"  /MACHINE:x64 /OUT:main.exe main.exe.p/main.c.obj "/nologo" "/release" "/nologo" "/DEBUG" "/PDB:main.pdb" "/SUBSYSTEM:CONSOLE" "kernel32.lib" "user32.lib" "gdi32.lib" "winspool.lib" "shell32.lib" "ole32.lib" "oleaut32.lib" "uuid.lib" "comdlg32.lib" "advapi32.lib"
````

with this change:

```sh
[1/2] "cl" "-Imain.exe.p" "-I." "-I.." "/MDd" "/nologo" "/showIncludes" "/W2" "/std:c11" "/ZI" "/Ob0" "/Od" "/RTC1" "/Fdmain.exe.p\main.c.pdb" /Fomain.exe.p/main.c.obj "/c" ../main.c
[2/2] "link"  /MACHINE:x64 /OUT:main.exe main.exe.p/main.c.obj "/nologo" "/release" "/nologo" "/DEBUG" "/PDB:main.pdb" "/SUBSYSTEM:CONSOLE" "kernel32.lib" "user32.lib" "gdi32.lib" "winspool.lib" "shell32.lib" "ole32.lib" "oleaut32.lib" "uuid.lib" "comdlg32.lib" "advapi32.lib"
```

Unlike other compilers, MSVC doesn't error when given an unknown option, so MSVC < 16.7 just ignores this option.

## Example

With MSVC 16.7, this fails to build without /std:c11 and works with it.

meson.build

```meson
project('vc11', 'c', default_options : ['c_std=c11'])

executable('main', 'main.c')
```

main.c

```c
// https://www.ibm.com/support/knowledgecenter/en/SSGH2K_13.1.2/com.ibm.xlc131.aix.doc/language_ref/genericselection.html
#include <stdio.h>

#define myfunction(X) _Generic((X), \
long double:myfunction_longdouble, \
default:myfunction_double, \
float:myfunction_float \
)(X)
void myfunction_longdouble(long double x){printf("calling %s\n",__func__);}
void myfunction_double(double x){printf("calling %s\n",__func__);}
void myfunction_float(float x){printf("calling %s\n",__func__);}

int main()
{
  long double ld;
  double d;
  float f;
  myfunction(ld);
  myfunction(d);
  myfunction(f);
}
```

### DIFF
--- a/mesonbuild/compilers/c.py
+++ b/mesonbuild/compilers/c.py
@@ -353,6 +353,26 @@ class VisualStudioCCompiler(MSVCCompiler, VisualStudioLikeCCompilerMixin, CCompi
                            info, exe_wrap, **kwargs)
         MSVCCompiler.__init__(self, target)
 
+    def get_options(self):
+        opts = super().get_options()
+        c_stds = ['none', 'c89', 'c99', 'c11']
+        opts.update({
+            'std': coredata.UserComboOption(
+                'C language standard to use',
+                c_stds,
+                'none',
+            ),
+        })
+        return opts
+
+    def get_option_compile_args(self, options):
+        args = []
+        std = options['std']
+        # As of MVSC 16.7, /std:c11 is the only valid C standard option.
+        if std.value in {'c11'}:
+            args.append('/std:' + std.value)
+        return args
+
 
 class ClangClCCompiler(ClangClCompiler, VisualStudioLikeCCompilerMixin, CCompiler):
     def __init__(self, exelist, version, for_machine: MachineChoice,


mesonbuild/compilers/c.py
```diff
@@ -353,6 +353,28 @@ def __init__(self, exelist, version, for_machine: MachineChoice,
...
+        })
+        return opts
+
+    def get_option_compile_args(self, options):
+        args = []
+        std = options['std']
+        # As of MVSC 16.7, /std:c11 is the only valid C standard option.
+        # default is c99-like, with c11 requiring /std:c11
+        # Add more as MSVC adds C standards in the future
+        if std.value in ('c11'):
```
dcbaker, line 374: you should use a set, not a tuple for membership tests like this, since sets are log time, and tuples or linear time for lookups.

mesonbuild/compilers/c.py
```diff
@@ -353,6 +353,28 @@ def __init__(self, exelist, version, for_machine: MachineChoice,
...
+            ),
+        })
+        return opts
+
+    def get_option_compile_args(self, options):
+        args = []
+        std = options['std']
+        # As of MVSC 16.7, /std:c11 is the only valid C standard option.
+        # default is c99-like, with c11 requiring /std:c11
+        # Add more as MSVC adds C standards in the future
```
dcbaker, line 373: Are you sure about this? I thought the default is C89 like, not C99 like (no designated initializers come to mind)

dcbaker:
I'm not sure about the winlis atm. Honestly we need to do some work to the "compilers generate their own options idea" there's some real hacks around the because we need some of that information to detect/pick the linker.

mesonbuild/compilers/c.py
scivision: Oh thanks I need to check into that thank you

mesonbuild/compilers/c.py
Akaricchi: More importantly, `('c11')` **is not** a tuple literal, it's a string literal wrapped in parentheses (so, equivalent to just `c11`). So this code can't work as intended. You must write `('c11',)` to produce a single-element tuple.

mesonbuild/compilers/c.py
scivision: based on some [toy examples](https://github.com/scivision/c11-examples) and docs, there are a scattershot of C99 features in MSVC 19.27 as required to support C++.  For example C99 `restrict` is in MSVC 19.27 but not 19.26.

scivision:
so based on my tests with various Visual Studio 2019 versions, it seems that allowing c89 and c99 to be silently accepted but not pass any new flags, and make c11 emit `/std:c11` is consistent with VS behavior through 16.7/19.27.

dcbaker:
Cool, vs2019 keeps surprising me in good ways.

dcbaker:
@nirbheek, how do you feel about this for 55.1?

nirbheek:
Yes, sounds good to me.

mesonbuild/compilers/c.py
```diff
@@ -353,6 +353,26 @@ def __init__(self, exelist, version, for_machine: MachineChoice,
...
+            ),
+        })
+        return opts
+
+    def get_option_compile_args(self, options):
+        args = []
+        std = options['std']
+        # As of MVSC 16.7, /std:c11 is the only valid C standard option.
+        if std.value in {'c11'}:
+            args.append('/std:' + std.value)
```
nirbheek, line 373: Why aren't we checking the version before passing this option? What happens if you pass this option to a version of MSVC that doesn't support it?

nirbheek: FWIW, MSVC 16.7 is `cl` version `19.27.XXXXX`. 16.6 was `19.26.XXXXX`.

nirbheek: I think we should have: 

1. A test that uses generics and uses `c_std=c11`, and is skipped on MSVC versions that don't support c11
2. A test that builds with `--werror` and uses `c_std=c11` unconditionally on some warning-free code, and is built on all MSVC versions

scivision: For cv < 19.27, the /std:c11 flag is ignored by cl.exe with typical unknown flag warning.
Yes I agree there should be a test.
Separately, this same flag is coming in CMake 3.18.2.

scivision: It would be more meaningful to the end user that they're doing something known not to be supported with older MSVC rather than emit compiler errors.

nirbheek: > For cv < 19.27, the /std:c11 flag is ignored by cl.exe with typical unknown flag warning.

I think we should either error out that the c std is not supported on the current MSVC, or we should not pass `/std:c11` at all.

I think I prefer the latter since this is a preview feature and it's not clear what c11 features are supported out of the box with MSVC and which require the flag, and what features are not available even when the flag is passed.

nirbheek: > It would be more meaningful to the end user that they're doing something known not to be supported with older MSVC rather than emit compiler errors.

Maybe we can emit a warning? I think we already do this for c++11 and c++14, f.ex..

nirbheek: >  Separately, this same flag is coming in CMake 3.18.2.

I am happy to block 0.55.1 on this being fixed :)

nirbheek: @scivision any updates here? Should we try to get this into 0.55.1?

scivision: I may not have time for a while, maybe leave this out of 0.55.1.

PR #7527, opened 2020-08-03
cmake: resolve IMPORTED executables in custom commands (fixes #7509)

mensinda:

PR #7557, opened 2020-08-06
cmake: Fix failing test case. Followup to #7527.

mensinda:


mensinda:
Merging since the CI is green

Issue #7495, opened 2020-07-24
Regression in 0.55: Compiling Vala requires C.


tim77:
**Describe the bug**
Since Meson 0.55 update we got error during build for Vala projects:
```
ERROR: Compiling Vala requires C. Add C to your project languages and rerun Meson.
```

**To Reproduce**
- [Koschei state](https://koschei.fedoraproject.org/package/gradio?collection=f33)
- [Build log](https://kojipkgs.fedoraproject.org/work/tasks/7138/47667138/build.log)
- [meson.build](https://github.com/haecker-felix/Gradio/blob/master/meson.build)

**Expected behavior**
Should successfully compiles as before.

**system parameters**
* plain native build
* Fedora 32
* Python 3.8.5
* meson 0.55.0
* ninja 1.10.0

Look like something similar already happened before https://github.com/mesonbuild/meson/issues/631.

jon-turney:
I suspect that perhaps the regression here is that the order the languages appear in `project()` is now significant.

xclaesse:
The language name is case-sensitive, with lower-case 'c' it works in my testing.

xclaesse:
But I can confirm it was working with Meson 0.54.0 somehow...

xclaesse:
Wondering why meson error on that instead of just adding 'c' language implicitly...

xclaesse:
Ok, `add_languages_for()` used to check if 'c' lang has been added after `lang = lang.lower()` but now it does it before.

tim77:
@jpakkane @jon-turney Thank you! It works with lowercase `c`.

PR #7558, opened 2020-08-06
interpreter: Lower case languages before checking if 'c' is in the list

xclaesse:
Fixes: #7495

### DIFF
--- a/mesonbuild/interpreter.py
+++ b/mesonbuild/interpreter.py
@@ -3246,6 +3246,7 @@ external dependencies (including libraries) must go to "dependencies".''')
         return should
 
     def add_languages_for(self, args, required, for_machine: MachineChoice):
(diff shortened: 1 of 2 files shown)


jon-turney:
I'm a bit surprised that languages are accepted case insensistively, and the documentation for `project()` seems to indicate they should be lower case, so I don't know if a warning or documentation change is appropriate here.

Also, altering some existing test case to cover using uppercase 'C' as a language would probably be a good idea.

Thanks for looking into this.

xclaesse:
IMHO, we can recommend lower case and it doesn't cost much to accept both, no need to annoy users with that detail. I added a unit test.

jpakkane:
> it doesn't cost much to accept both

Except that causes bugs like these sometimes. :(

Issue #7555, opened 2020-08-06
meson finds hdf5 dependency on CentOS 7 but fails to link libhdf5


williamfgc:
**Describe the bug**
meson finds hdf5 dependency on CentOS 7 from epel hdf5-devel-1.8.12-11.el7.x86_64 `yum install hdf5-devel`, but fails to link to `libhdf5.so`

**To Reproduce**
meson.build
```
hdf5_dep = dependency('hdf5',
                      language: 'c',
                      required: false,
                      not_found_message: 'NCIO dependency HDF5 not found')
```
Config:
`Run-time dependency HDF5 for c found: YES`

Linker is missing a reference to `libhdf5.so` variant when running `ninja`

```
[12/18] Linking target meson-out/libncio.so
FAILED: meson-out/libncio.so 
c++  -o meson-out/libncio.so meson-out/libncio.so.p/core_NCIO.cpp.o meson-out/libncio.so.p/core_DataDescriptor.cpp.o meson-out/libncio.so.p/helper_ncioHelperString.cpp.o meson-out/libncio.so.p/transport_Transport.cpp.o meson-out/libncio.so.p/transport_TransportNull.cpp.o meson-out/libncio.so.p/transport_TransportHDF5.cpp.o -Wl,--as-needed -Wl,--no-undefined -Wl,-O1 -shared -fPIC -Wl,--start-group -Wl,-soname,libncio.so -pthread -Wl,-z,relro -ldl -lm -Wl,-rpath -Wl,/usr/lib64 -L/usr/lib64 -lsz -lz -Wl,--end-group '-Wl,-rpath,$ORIGIN/' -Wl,-rpath-link,/home/ncio/build/meson-out
meson-out/libncio.so.p/transport_TransportHDF5.cpp.o: In function `ncio::transport::TransportHDF5::DoClose()':
TransportHDF5.cpp:(.text+0x2ff): undefined reference to `H5Fclose'
meson-out/libncio.so.p/transport_TransportHDF5.cpp.o: In function `ncio::transport::TransportHDF5::TransportHDF5(std::string const&, ncio::openmode, std::map<std::string, std::string, std::less<std::string>, std::allocator<std::pair<std::string const, std::string> > > const&)':
TransportHDF5.cpp:(.text+0x49a): undefined reference to `H5open'
TransportHDF5.cpp:(.text+0x4a1): undefined reference to `H5P_CLS_FILE_ACCESS_g'
[...]

PR #7559, opened 2020-08-06
Add h5cc shlib argument to correctly link hdf5 libraries without pkg-config support

williamfgc:
Older versions of hdf5 (e.g. 1.8) don't have a pkg-config `hdf5.pc` file. Meson uses hdf5's `h5cc` binary for `DependencyMethods.AUTO`.

This PR adds the corresponding arguments to respect "static" key in the dependency function for hdf5:
`h5cc -shlib -show` for shared dependencies
`h5cc -noshlib -show` for static dependencies
Currently `-show` shows only static dependencies by default.

Partially addresses #7555. HDF5 static libraries are not distributed with `fPIC`.

### DIFF
--- a/mesonbuild/dependencies/hdf5.py
+++ b/mesonbuild/dependencies/hdf5.py
@@ -105,7 +105,8 @@ class HDF5Dependency(ExternalDependency):
                 prog = ExternalProgram(wrappers[lang], silent=True)
                 if not prog.found():
                     return
-                cmd = prog.get_command() + ['-show']
+                shlib_arg = '-noshlib' if kwargs.get('static', False) else '-shlib'
+                cmd = prog.get_command() + [shlib_arg, '-show']
                 p = subprocess.run(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, universal_newlines=True, timeout=15)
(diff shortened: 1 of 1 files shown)


mesonbuild/dependencies/hdf5.py
```diff
@@ -105,7 +105,8 @@ def __init__(self, environment, kwargs):
                 prog = ExternalProgram(wrappers[lang], silent=True)
                 if not prog.found():
                     return
-                cmd = prog.get_command() + ['-show']
+                shlib_arg = '-noshlib' if kwargs.get('static', False) else '-shlib'
+                cmd = prog.get_command() + [shlib_arg + ' -show']
```
dcbaker, line 109: I think this needs to be, `cmd = prog.get_command() + [shlib_arg', '-show']`, otherwise the quoting will be wrong.

williamfgc: @dcbaker thanks for the review. Yes, it should be a list. I just fixed it.

dcbaker: Thanks, I'll go ahead and merge this.

PR #7561, opened 2020-08-07
qt dependency: do not require all the tools automatically

### LINKED ISSUE #5582: Qt5 modules should not always require uic
Qt5 can be built without `uic` (if widgets are disable). This is quite common for embedded use-cases with QtQuick only applications.

Currently meson fails because it always requires `uic`. I'm not sure what the correct fix is.
Maybe handle this like `lrelease`?

eli-schwartz:
The compilers_detect function is only used in the qt module, which checks every time before using a specific compiler, if it is found.

e.g.

```
meson.build:10:6: ERROR: MOC sources specified and couldn't find moc-qt5, please check your qt5 installation
```

In fact, the current check means we never even hit this error to begin with, because we previously died on the uninformative error:

```
meson.build:10:6: ERROR: Program(s) ['moc'] not found or not executable
```

which doesn't actually tell the user why this matters, and is all around a waste of time.

Fixes #5582

### DIFF
--- a/mesonbuild/dependencies/ui.py
+++ b/mesonbuild/dependencies/ui.py
@@ -234,13 +234,13 @@ class QtBaseDependency(ExternalDependency):
         def gen_bins():
             for b in bins:
                 if self.bindir:
-                    yield os.path.join(self.bindir, b), b, False
+                    yield os.path.join(self.bindir, b), b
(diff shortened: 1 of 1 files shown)


xclaesse:
lgtm.

PR #7568, opened 2020-08-08
Use uname -p to detect x86_64 on QNX.

elahav:
On QNX/x86_64 "uname -m" reports "x86pc" (for both the 32- and 64 bit variants of x86). To see the actual architecture name we need to use "uname -p", like the *BSDs.


dcbaker:
This looks safe to add to the next stable, so I've tagged it.

PR #7572, opened 2020-08-08
CI: Port MSYS2 from azure-pipelines to github actions

lazka:
This uses the [msys2/setup-msys2@v2](https://github.com/msys2/setup-msys2) to set up MSYS2. The upside being that you don't have to handle the MSYS2 install/setup quirks (that's the plan at least)


lazka:
@scivision still failing in one case. Any ideas?

scivision:
I am not sure if it is a defect in the algorithm used to generate long test filenames. I tried to make it platform agnostic but perhaps I fell short. This is the test that is failing.

lazka:
ok, turned out I assumed MESON_RSP_THRESHOLD env var was set in the old CI, but it wasn't actually used.

I've also installed pytest-xdist now in a second commit, which makes things a bit faster.

Issue #7573, opened 2020-08-08
[bisected] Can no longer override subproject defaults in superproject's project() statement


Akaricchi:
**Describe the bug**
In Meson 0.55.0 and below, it was possible to override a subproject's default options inside the superproject's `project()` statement, like so:
```meson
project('superproject', 'c',
  default_options : [
      'subproject:test_option=false'])
```

In current git master, this is broken. Configuring such a project for the first time results in the following (irrelevant output omitted):
```
$ grep test_option subprojects/subproject/meson_options.txt 
option('test_option', type : 'boolean', value : true, description : 'Test option. Superproject overrides default to "false"')

$ meson setup build

... omitted ...

WARNING: Unknown options: "subproject:test_option"
The value of new options can be set with:
meson setup <builddir> --reconfigure -Dnew_option=new_value ...

... omitted ...

|Executing subproject subproject method meson 
|
|WARNING: In subproject subproject: Unknown options: "subproject:subproject:test_option"
|The value of new options can be set with:
|meson setup <builddir> --reconfigure -Dnew_option=new_value ...

... omitted ...

Option subproject:test_option is: true [default: false]

$ meson configure build | grep test_option
  subproject:test_option     true           [true, false]       Test option. Superproject overrides default to "false"
```

It seems to recognize `false` as the new "default", however it still applies the old default value to a fresh build directory, as if `-Dtest_option=true` was passed explicitly. If an old build directory is `--wipe`d, any such options are also reverted to their subproject defaults, unless they have been explicitly overriden at any point during the build dir's lifetime.

There is also a new spurious warning
```
|WARNING: In subproject subproject: Unknown options: "subproject:subproject:test_option"
```

This seems to happen for *every* option overriden via the `project()` statement.

**To Reproduce**
Here is a minimal test case: 
[meson-test-subproject-option-override-from-superproject.zip](https://github.com/mesonbuild/meson/files/5046588/meson-test-subproject-option-override-from-superproject.zip)

**Expected behavior**
After configuring a build directory for the attached test case, `subproject:test_option` should be `false`. There should not be any warnings about missing options.

**system parameters**
* meson 0.55.999 (1ddd2b4cb)

Akaricchi:
Bisected to `591e6e94b9fccfc49ee7093cb21735a27fd64005`, part of #6597. CC @dcbaker

xclaesse:
[...]

PR #7574, opened 2020-08-09
Refactor handling of machine file options

xclaesse:
This solves spurious warnings I reported in #6597, fix a regression reported in #7573, and make behaviour of PKG_CONFIG_PATH consistent with other env like CFLAGS.

### DIFF
--- a/mesonbuild/ast/introspection.py
+++ b/mesonbuild/ast/introspection.py
@@ -125,7 +125,7 @@ class IntrospectionInterpreter(AstInterpreter):
                         self.do_subproject(i)
 
         self.coredata.init_backend_options(self.backend)
-        options = {k: v for k, v in self.environment.meson_options.host[''].items() if k.startswith('backend_')}
+        options = {k: v for k, v in self.environment.raw_options.items() if k.startswith('backend_')}
 
         self.coredata.set_options(options)
         self._add_languages(proj_langs, MachineChoice.HOST)
--- a/mesonbuild/compilers/compilers.py
+++ b/mesonbuild/compilers/compilers.py
@@ -28,7 +28,7 @@ from ..mesonlib import (
     Popen_safe, split_args, LibType
 )
 from ..envconfig import (
-    Properties, get_env_var
+    get_env_var
 )
 
 from ..arglist import CompilerArgs
@@ -1246,8 +1246,7 @@ def get_args_from_envvars(lang: str,
 def get_global_options(lang: str,
                        comp: T.Type[Compiler],
                        for_machine: MachineChoice,
-                       is_cross: bool,
-                       properties: Properties) -> 'OptionDictType':
+                       is_cross: bool) -> 'OptionDictType':
     """Retreive options that apply to all compilers for a given language."""
     description = 'Extra arguments passed to the {}'.format(lang)
     opts = {
@@ -1267,11 +1266,7 @@ def get_global_options(lang: str,
         comp.INVOKES_LINKER)
 
     for k, o in opts.items():
-        user_k = lang + '_' + k
-        if user_k in properties:
-            # Get from configuration files.
-            o.set_value(properties[user_k])
-        elif k == 'args':
+        if k == 'args':
             o.set_value(compile_args)
         elif k == 'link_args':
             o.set_value(link_args)
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -735,7 +735,9 @@ class CoreData:
         for k, v in options.items():
             if k.startswith('build.'):
                 k = k.split('.', 1)[1]
-            res[k] = v
+                res.setdefault(k, v)
+            else:
+                res[k] = v
         return res
 
     def copy_build_options_from_regular_ones(self):
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -658,6 +658,23 @@ class CoreData:
             for k1, v1 in v0.items():
                 yield (k0 + k1, v1)
 
+    @classmethod
+    def insert_build_prefix(cls, k):
+        idx = k.find(':')
+        if idx < 0:
+            return 'build.' + k
+        return k[:idx + 1] + 'build.' + k[idx + 1:]
+
+    @classmethod
+    def is_per_machine_option(cls, optname):
+        if optname in BUILTIN_OPTIONS_PER_MACHINE:
+            return True
+        from .compilers import compilers
+        for lang_prefix in [lang + '_' for lang in compilers.all_languages]:
+            if optname.startswith(lang_prefix):
+                return True
+        return False
+
     def _get_all_nonbuiltin_options(self) -> T.Iterable[T.Dict[str, UserOption]]:
         yield self.backend_options
         yield self.user_options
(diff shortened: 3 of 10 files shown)


xclaesse:
Even went a step further than #7573: A subproject can itself have `project(..., default_options: ['another_sub:option=value'])`.

jpakkane:
@nirbheek should this go to 0.55.1 as well? Seems like it should.

xclaesse:
No the regression is in master, 0.55 is fine.

Ericson2314:
I still think this is too much flattening, sorry. That we can categorize the raw options by looking at, e.g. the language prefix, `build.`, etc. before we know what the available options are, is a good feature we can use to validate things earlier (e.g you cannot call your own option `c_foo`, as we already detect). And it's also a key property used to clean up the structured stuff under @dcbaker's and my plan (very easy to match up structured options with structured raw options, when the structuring is the same) so let's not get rid of it just yet.

xclaesse:
I think the numerous regressions proved me right, that PR is just too complicated. That PR got merged mostly unreviewed, worse it got merged against my explicit objection. I really don't understand why you're on a quest for the most complicated structure possible where I showed you easiest and more correct solutions.

Ericson2314:
@xclaesse I would say the number one mantra I've seen pay off again and again is: carefully design the data structures to maintain invariants, and the code will write itself. (Perhaps the earliest version of this "Show me your flowchart and conceal your tables, and I shall continue to be mystified. Show me your tables, and I won't usually need your flowchart; it'll be obvious.")

There are definitely problems with the code, but I think in you're completely misdiagnosing them in trying to flatten the data structures.

xclaesse:
Fact remains: my code works, currently master does not. And about flattening lang dict you would have to show me how to implement per subproject compiler options. Until then, I do have solutions to actual problems, you have theorical design that serves no purpose.

Let's agree we'll never agree on this topic ;-)

xclaesse:
> @xclaesse I would say the number one mantra I've seen pay off again and again is: carefully design the data structures to maintain invariants, and the code will write itself. (Perhaps the earliest version of this "Show me your flowchart and conceal your tables, and I shall continue to be mystified. Show me your tables, and I won't usually need your flowchart; it'll be obvious.")

Funny because that's exactly why we have to flatten the structure. A flat structure does not need explanation, it's obvious. But current structure you have to explain multiple dicts, some per-machine, some per-language, you have to explain why keys does not match the name of the option as user sees it... 

Compare https://github.com/mesonbuild/meson/pull/7293 and https://github.com/mesonbuild/meson/pull/6597, then tell me which one has code that "write itself".

dcbaker:
>  you have to explain why keys does not match the name of the option as user sees it...

This is a classic misthink in compiler/interpreter design. The thing the user inputs is a thing that is easy for them to type and remember, our internal IR (which is frankly what this is) should be in a for useful for our use/consumption. having our keys stored as `zlib:build.c_std` instead of `options['zlib'].build['c']['std']` may seem appealing, but there's all sorts of downsides to that kind of format, if I want to get all of the options for the zlib subproject's C compiler on the builds I would do:
```python
for k, v in options.items():
  if not k.startswith('zlib:'):
    continue
  k = k.strip('zlib:')
  if not k.startswith('build.')
    continue
  k = k.strip('.build')
  if k.startswith('c_'):
     continue
  yield k
```

as opposed to:

```python
yield from options['zlib'].build['c']
```

Making our internal representation match what the user input is 100% backwards. It's half the reason that my MR was so huge and messy, is because our internal IR looks too much like what the user inputs, and is thus hard to work with. I absolutely and against flattening the structure because instead of having multiple dicts, we have a key that is multiple keys encoded as a string.

xclaesse:
@dcbaker the problem is you try to optimize for a case that does not exist. We never need that code snippet you wrote. But we do have in MANY places _flatten_pair_iterator(). A flat structure is actually what is the more convenient to consume, like it or not.

dcbaker:
Sure, but you still have essentially a flat dictionary acting as a nested one, because you have a 4 part key:
`<Subproject>:[build.]<lang>_<option>`. We could store that using a tuple, `(subproject, build/host, lang, option)` but that just proves the point that the flat dictionary is the wrong abstraction, because we still have to do things like:
```python
for k, v in options if k[0] == 'zlib' and k[1] == HostMachine.BUILD and k[2] == 'c':
   ...
```

We really do often need to get all the options for just [sub]projectA, or all of the BUILD options, nesting these data structures puts this data in the form we want to work with.

dcbaker:
And also, there is a bug with environment variables, but PKG_CONFIG_PATH is the only one that works correctly, environment variables should override config files not the other way around. That's in keeping with the UNIX philosophy. So actaully, we can probably delete a *lot* of code out of the compilers by moving all of that handling into Environment with PKG_CONFIG_PATH

xclaesse:
Consistency and backward compatibility are more important than an obscure Unix philosophy. Change them all or don't change anything but in both cases that's for a separate PR where we can discuss it explicitly instead of sneaking in it in an interview PR.

Anyway, of you really believe your code is better, fine, I give you responsibility of maintaining it, but please fix regressions I mentioned and implement per subprojects compiler options. Must be easy given how great that data structure is.

nirbheek:
> Must be easy given how great that data structure is.

This kind of snark is unnecessary, Xavier. It's not going to make it easier to work amicably 🙂

xclaesse:
Just to make it clear to anyone reviewing this PR, the conversation above got side tracker from unrelated discussion. This PR is NOT about flattening the lang dictionary, that's another topic.

Here I revert part of https://github.com/mesonbuild/meson/pull/6597 and implement the safer and simpler solution that I suggested in https://github.com/mesonbuild/meson/pull/7293, but keeping @dcbaker's proposed machine file syntax (i.e. categories like `[zlib:built-in options]`).

dcbaker (commented):
I still really think this is going backwards, I don't like it, and I'll leave it at that.

But, I'm super concerned about allow subprojects to alter other subprojects options in their default_options.

Imagine that I have a project, and I use subprojects A and B. If A uses C, and C uses D, and D *could* use B but doesn't, and has `B:foo=false` in it's default_options, then when my project initializes B then it will have `foo=false`. That's going to be a nightmare for an end user and we'll end up fielding bugs like "WTF? my subproject B is getting default options set and I can't figure out why!"

I know that can already happing with D calling `subproject('B', override_options: ...)`, but this can happen even if D doesn't use B.

I'd much rather if subproject option in other subprojects `default_options` was ignored, hopefully with a warning.

xclaesse:
> (...) but this can happen even if D doesn't use B.

That's a good point, haven't thought about that.

Note that here I'm just reverting to how it was working in latest release because we got a regression report: https://github.com/mesonbuild/meson/issues/7573. Maybe this should be discussed further in a separate issue/PR.

I personally actually like that we can do `project(..., default_options: ['sub:opt=true'])` because it is much cleaner to have it on top of your project together with other options rather than buried down into `dependency('sub', default_options: ['opt=true'])`. An extra advantage is the main project can set options on all its subprojects at the very top to solve ordering issues. I didn't know it was possible before getting the regression report, but now I'm planning to actually use that. If anything, maybe we should even deprecate default_options in dependency() and subproject() in favor of that.

I think the questions are: Do we want to allow that only on the main project? If yes, in you example above, `project('D', default_options: ['B:foo=false'])` should warn or silently ignore that line? A warning is annoying because B has no way to know if it is used as main project or subproject within project() function, does it? Silently ignoring is also annoying because it's not backward compatible change I think, and in the case A->B->C we force A to know about C instead of using the default set by B which uses C directly.

dcbaker:
In my mind the answer is "yes, only the root/super/main project should be able to set subproject options in it's `project()` call". I know that is a deviation from what happened before, but I want to stress we never documented or intended to make that work before, so defining that behavior in such a way that it doesn't have side effects is a good thing. Keeping meson easy to understand and work with is important, having side effects from calling into a subproject is a potentially annoying side effect.

xclaesse:
I don't think we should restrict to main project:
- If a subproject does it we cannot warn that the option is ignored, because the project has no way to know if it's being used as main or subproject.
- If the option is set in a subproject's project() and silently ignored that seems super surprising to me, users won't understand that.
- subprojects can already override_options when invoking a subsubproject, so better do it at the top. Hiding it in override_option is worse.
- Whatever we choose, I think it's simply out of scope for this PR. Restoring the behavior as it currently is in 0.55.3 stable release is the obvious first step. From there we can always decide to change the behavior in a future PR. Notice that I do not document this feature just yet, to leave the door open.

dcbaker:
No, no, it is in scope. Because it was never documented, it was previously undefined behavior. It happened to work, but there was no guarantee it would. If you turn it on now then it's defined behavior and it's a regression to turn it off later.

@jpakkane, @nirbheek, @mensinda  can we get some input on this please.

xclaesse:
So it was working in all previous releases and that's not a problem, but if it keeps working in 0.56 then suddenly it becomes defined behavior? That makes no sense to me...

dcbaker:
What doesn't make sense? Projects have unintended behavior all the time, it wasn't documented, it wasn't intended. it *just happened to work*. When that happens they have to make a choice between saying "that seem useful, safe, and not like a maintenance burden, let's support that officially", or to say "that did happen to work, but it's not a feature, it's a bug." I'm saying for the main project it's fine, let's keep it. But for subprojects to override other subproject options in their default_options? that's a bug. If they really need to set those options universally them can do it in the dependency() call, *without* side effects.

xclaesse:
Yes a decision needs to be made at some point, but it makes no sense to rush it in this PR just a few days before release. I intentionally did not document this feature either way to leave the door open for when we get an agreement.

dcbaker:
A commit called "turn this feature on" and a testcase for it is as much documentation as a note in the docs.

Akaricchi:
> But for subprojects to override other subproject options in their default_options? that's a bug. If they really need to set those options universally them can do it in the dependency() call, without side effects.

I'm probably misunderstanding something here, but how is a subproject overriding another subproject's options in the `dependency()` or `subproject()` call any better than it doing that in the `project()` call...?

jpakkane:
One of the design points of Meson has always been to try to have as few places for configuration options as possible, preferably there is only one canonical place. This is, for example, why all wrap info must come from a single place at the top level directory rather than wherever around the file system tree.

Because of this I don't really like having subproject options in `project`. They can already be defined in dependency and subproject calls, and these are as close to the point of call as possible. They are also easy to grep for. Having them in `project` also gives off a vibe of spooky-action-at-a-distance.

Ericson2314:
@jpakkane Am I interpreting you correctly that you are asking for the feature to be restricted even more than @dcbaker is?

dcbaker:
@Akaricchi, in the project call they can effect a subproject even if they don't invoke it. If I use subprojects A and B, and B can use A but won't won't in my case, and B has options for A, then those options will be used when I Invoke A if I Invoke B first.

xclaesse:
Clearly everyone has a different opinion, and again, this is OUT OF SCOPE for this PR (I can drop the unit test if you want). I opened https://github.com/mesonbuild/meson/issues/7854 to discuss what we should do. Let's please focus on getting this PR actually reviewed in time for 0.56 release.

dcbaker:
Everything except the bit of functionality we've been debating has my ack.

xclaesse:
@dcbaker ahhh, I think I misunderstood something! After running some tests on 0.55.3 I realize now that I was wrong, setting sub-subproject options in a subproject was NOT working before. Only setting them in the main project was working and regressed in master. With that, I totally agree the last patch in the PR was wrongly adding a new behavior instead of reverting to previous behavior as I was assuming. I dropped that commit now, sorry for the confusion.

Issue #7534, opened 2020-08-04
Dependencies that share a fallback subproject should always look in the system first


nirbheek:
This is a regression, I am pretty sure this worked correctly in the past.

Let's say you have a subproject called `gst-plugins-base` that has multiple pkgconfig libraries which are dependent on feature options, say: `gstreamer-gl-x11-1.0`, `gstreamer-gl-wayland-1.0`, `gstreamer-gl-egl-1.0`.

If there's a project that consumes this as follows:

```meson
gst_gl_x11 = dependency('gstreamer-gl-x11-1.0', required: false, fallback: ['gst-plugins-base', 'gst_gl_x11_dep'])
gst_gl_wayland = dependency('gstreamer-gl-wayland-1.0', required: false, fallback: ['gst-plugins-base', 'gst_gl_wayland_dep'])
gst_gl_egl = dependency('gstreamer-gl-egl-1.0', required: false, fallback: ['gst-plugins-base', 'gst_gl_egl_dep'])
```

If `gstreamer-gl-x11-1.0.pc` is not found, and the fallback subproject is not available, then meson won't look for the remaining ones either:

```
Run-time dependency gstreamer-gl-x11-1.0 found: NO (tried pkgconfig and cmake)
Looking for a fallback subproject for the dependency gstreamer-gl-x11-1.0
Subproject directory not found and gst-plugins-base.wrap file not found
Subproject  subprojects/gst-plugins-base is buildable: NO (disabling)
Dependency gstreamer-gl-x11-1.0 from subproject subprojects/gst-plugins-base found: NO
Dependency gstreamer-gl-wayland-1.0 from subproject subprojects/gst-plugins-base found: NO
Dependency gstreamer-gl-egl-1.0 from subproject subprojects/gst-plugins-base found: NO
```

FWICT, this is happening because of this code in `dependency_impl()`:

```python
# If the dependency has already been configured, possibly by
# a higher level project, try to use it first.
if has_fallback:
    dirname, varname = self.get_subproject_infos(kwargs)
    if dirname in self.subprojects:
        return self.get_subproject_dep(name, display_name, dirname, varname, kwargs)
```

CC: @xclaesse via `git blame` 😉

xclaesse:
I think it has always been like that, but I changed so much code in latest release that it could have regressed.

I guess it would make sense to add a condition there: if subprojects[dirname].fount()

xclaesse:
[...]

PR #7580, opened 2020-08-11
interpreter: Don't force fallback when subproject failed to configure

xclaesse:
Fixes: #7534


nirbheek:
Fix looks simple enough, and has a test. I'm pulling this into the stable release.

PR #7587, opened 2020-08-13
mtest: Allow filtering tests by subproject

nirbheek:
You could always specify a list of tests to run by passing the names as arguments to `meson test`. If there were multiple tests with that name (in the same project or different subprojects), all of them would be run. Now you can:

1. Run all tests with the specified name from a specific subproject: `meson test subprojname:testname`
1. Run all tests defined in a specific subproject: `meson test subprojectname:`

Also forbid ':' in test names. We already forbid this elsewhere, so should not be a big deal.

### DIFF
--- a/docs/markdown/snippets/mtest_test_list_subprojects.md
+++ b/docs/markdown/snippets/mtest_test_list_subprojects.md
@@ -1,0 +1,18 @@
+## `meson test` can now filter tests by subproject
+
+You could always specify a list of tests to run by passing the names as
+arguments to `meson test`. If there were multiple tests with that name (in the
+same project or different subprojects), all of them would be run. Now you can:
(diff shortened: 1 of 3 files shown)


xclaesse:
Can't you already do that with `meson test --suite <subproject>` ?

jpakkane:
Ping. Xclaesse's question remains unaswered.

nirbheek:
> Can't you already do that with `meson test --suite <subproject>` ?

Ewww! I just looked at the code and this is an undocumented side-effect of the fact that test suites are transformed to `subproject:suitename` when writing the test serialization. It is not an intended effect at all, and it's also not very discoverable. I don't think we should start using it as the recommended way to do this.

PR #7591, opened 2020-08-14
docs: Clarify some FAQs in the Tutorial [skip ci]

nirbheek:
Clarifications that were observed to be needed when seeing a user try the Tutorial for the first time at:
https://www.twitch.tv/videos/709237047?t=50m20s


xclaesse:
lgtm

Issue #7590, opened 2020-08-13
Meson warns about use of a feature even if its use is limited for higher versions anyway


ebraminio:
Meson complain about use of `summary` even if is hide behind `meson.version().version_compare('>=0.53')` for older versions compatibility:

```
WARNING: Project specifies a minimum meson_version '>= 0.47.0' but uses features which were added in newer versions:
 * 0.53.0: {'summary'}
```

I've used a polyfill for summary https://github.com/harfbuzz/harfbuzz/commit/58209c86a0d2d536ce3d35953ffe340762ad1f48#diff-969b60ad3d206fd45c208e266ccfed38R399 so can have `summary` of 0.53 but have something similar on older version but meson complains about it anyway.

**To Reproduce**
```
project('test', meson_version: '>= 0.47.0')

if meson.version().version_compare('>=0.53')
  summary({'Feature': true}, bool_yn: true, section: 'Section 1')
else
  message('test')
endif
```

**Expected behavior**
I understand if you don't see it possible to workaround. Maybe at least you can turn the warn to something not that scary like "make sure the use is behind proper version checking", or maybe you can have someway to check you are already behind version checking guard?

From a very unrelated context, something similar for example happens with Android development lint that when some part of a code isn't behind a proper Android version check the linter warns the developer but that will be resolved if the code is behind a version checking guard. Android Studio linter needs an immediate surrounding guard (or some annotation or something) as the complexity of Java/Kotlin codes but guess you are able to do something even better in meson.

**system parameters**
* Is this a [cross build](https://mesonbuild.com/Cross-compilation.html) or just a plain native build (for the same computer)? No
* what operating system, Linux
* what Python version are you using e.g. 3.8.3
* what `meson --version` 0.55.0
* what `ninja --version` 1.10.0

PR #7594, opened 2020-08-14
Special case meson.version().version_compare() statement

xclaesse:
when that statement gets evaluated, the interpreter remembers the
    version target and if it was part of the evaluation of a `if` condition
    then the target meson version is temporally overriden within that
    if-block.

Fixes: #7590

### DIFF
--- a/mesonbuild/interpreterbase.py
+++ b/mesonbuild/interpreterbase.py
@@ -458,7 +458,7 @@ class InterpreterBase:
         # meson.version().compare_version(version_string)
         # If it was part of a if-clause, it is used to temporally override the
         # current meson version target within that if-block.
-        self.tmp_meson_version = None # type: str
+        self.tmp_meson_version = None # type: T.Optional[str]
 
     def load_root_meson_file(self) -> None:
         mesonfile = os.path.join(self.source_root, self.subdir, environment.build_filename)
--- a/mesonbuild/interpreterbase.py
+++ b/mesonbuild/interpreterbase.py
@@ -54,6 +54,9 @@ TYPE_var = T.Union[TYPE_elementary, list, dict, InterpreterObject, ObjectHolder]
 TYPE_nvar = T.Union[TYPE_var, mparser.BaseNode]
 TYPE_nkwargs = T.Dict[T.Union[mparser.BaseNode, str], TYPE_nvar]
 
+class MesonVersionString(str):
+    pass
+
 # Decorators for method calls.
 
 def check_stringlist(a: T.Any, msg: str = 'Arguments must be strings.') -> None:
@@ -451,6 +454,11 @@ class InterpreterBase:
         # Current node set during a function call. This can be used as location
         # when printing a warning message during a method call.
         self.current_node = None  # type: mparser.BaseNode
+        # This is set to `version_string` when this statement is evaluated:
+        # meson.version().compare_version(version_string)
+        # If it was part of a if-clause, it is used to temporally override the
+        # current meson version target within that if-block.
+        self.tmp_meson_version = None # type: str
 
     def load_root_meson_file(self) -> None:
         mesonfile = os.path.join(self.source_root, self.subdir, environment.build_filename)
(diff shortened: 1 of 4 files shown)


xclaesse:
@jpakkane what do you think about this trick?

xclaesse:
Will add doc & release notes if we agree on API of course.

nirbheek:
I like this a lot. Is there a way to return the `MesonVersionHolder` object from the usual clause too? `meson.version().version_compare()`? 

That way it will work for all existing build files without any changes, and without upping the meson version requirement.

jpakkane:
What's the point of returning a special object rather than a plain `bool`?

xclaesse:
Being able to special case it in evaluate_if

mensinda:
I am not sure how useful this will be (since it doesn't look very flexible). As far as I can tell this will only work if it is directly used in an if clause?

Also, what happens in the following scenarios:

```meson
if not meson.version_compare('>=0.49')
  # do not use the feature
else
  # will this warn if a meson >=0.49 feature is used?
endif
```

```meson
if (false or meson.version_compare('>=0.49')) and true
  # will this warn if a meson >=0.49 feature is used?
endif
```

xclaesse:
> I am not sure how useful this will be (since it doesn't look very flexible). As far as I can tell this will only work if it is directly used in an if clause?

Yes it's not flexible, and I wish I have a better idea to implement this tbh. Ideas welcome! But it is really useful already, because most of the time (at least everywhere I've seen) you need to do something only if meson version supports it, you'll do that in a `if` block with a direct version check. This is aimed for that kind of code: https://gitlab.freedesktop.org/gstreamer/gstreamer/-/blob/master/meson.build#L586

> 
> Also, what happens in the following scenarios:
> 
> ```meson
> if not meson.version_compare('>=0.49')
>   # do not use the feature
> else
>   # will this warn if a meson >=0.49 feature is used?
> endif
> ```

This will actually not even work, because `meson.version_compare()` does not return a boolean, so the `not` will be rejected. You would have to write that as `if meson.version_compare('<0.49')` and the `else` case will use the `meson_version` from your `project()`.

> ```meson
> if (false or meson.version_compare('>=0.49')) and true
>   # will this warn if a meson >=0.49 feature is used?
> endif
> ```

Won't work, you can't do any boolean operation with the return value of `meson.version_compare()` other than passing it directly to `if` or `elif` clauses. I'm not particularly happy with that behavior, but I didn't had better idea... :/

jpakkane:
Maybe a better approach would be to print the warning message only if the function is actually called? IOW so that this snippet:

```meson
if false
    some_future_function()
endif
```

Would never print a warning.

xclaesse:
@jpakkane if course that won't print the warning, but that's not the point here. I should have started with the use-case actually:

GStreamer wants to keep compatibility with meson 0.48, so it has `project(..., meson_version: '>=0.48')`. `summary()` has been added in version `0.53`, so it does:
```
if meson.version().version_compare('>= 0.54')
  summary()
endif
```
When user actually has meson 0.54, it will do that summary() call and get that warning: `|WARNING: Project targeting '>= 0.48' but tried to use feature introduced in '0.53.0': summary.`

My proposed solution here is do do:
```
if meson.version_compare('>= 0.54')
  summary()
endif
```
That would temporally override `meson_version: '>=0.48'` and consider that within that if-block `meson_version` is now `>=0.54`.

xclaesse:
Note that one big downside is that `meson.compare_version()` itself is new API for 0.56, so that would be only a solution for the future. Ideally we would need something that would work with older versions too... :/

mensinda:
Ideally, we wouldn't directly use `str`, `int`, and `bool` in the interpreter, but use objects for everything instead. This would make this so much easier since `meson.version()` could then return something subclassed from the custom string class with some special logic for the `version_compare` function. This would also make typing and https://github.com/mesonbuild/meson/blob/dce60c8f58798e0457a9bb24198adb039a0e6fdd/mesonbuild/interpreterbase.py#L886-L910 much easier since everything would be an `InterpreterObject` (maybe at the cost of some performance though).

\</rant\>

xclaesse:
@jpakkane @mensinda ok, got a better idea! It fix my concern about the initial proposal: It was requiring new API to fix warning about usage of new API... which was only a solution for the future but not for existing code.

Now existing code that does `if meson.version().version_compare('>= 0.55')` will work, without using any new syntax.  The interpreter detect that statement, and if it was done during evaluation of a if condition then it override the target meson version during evaluation of the if block.

The only issue I see with this new approach is that negation won't work as expected: `if not meson.version().version_compare('>= 0.55')` will evaluate that block with meson target `>=0.55`.

xclaesse:
@mensinda I agree usage of python built-in types is unfortunate, but I think it's too late to change that now :(

bonzini:
>     if false
>         some_future_function()
>     endif

This should warn because otherwise you might miss warnings for new-version constructs that are only used on Windows hosts. Unless specially flagged with `version_compare()`, those should warn on all hosts, even those for which that code becomes dead.

xclaesse:
It does not warn, just like with any interpreter, the code in a false condition is not evaluated. Same thing with python.

bonzini:
> the code in a false condition is not evaluated

Warning about missing features can be done at parsing time though, it doesn't need evaluation.

mesonbuild/interpreterbase.py
```diff
@@ -451,6 +454,11 @@ def __init__(self, source_root: str, subdir: str, subproject: str):
         # Current node set during a function call. This can be used as location
         # when printing a warning message during a method call.
         self.current_node = None  # type: mparser.BaseNode
+        # This is set to `version_string` when this statement is evaluated:
+        # meson.version().compare_version(version_string)
+        # If it was part of a if-clause, it is used to temporally override the
+        # current meson version target within that if-block.
+        self.tmp_meson_version = None # type: str
```
mensinda, line 461: This should be `T.Optional[str]`
```suggestion
        self.tmp_meson_version = None # type: T.Optional[str]
```

mensinda (approved):
LGTM, assuming the CI is green.

xclaesse:
Thanks, merged.

PR #7595, opened 2020-08-14
Fix option handling in machine files

dcbaker:
There are two distinct problems I'm aiming to fix:
1) environment variables vs config files

Currently this is handled rather haphazardly, resulting in environment variables overriding config in some, but not all cases. It should be this order (environment variables overriding config files) because that's the unix tradition, and doing anything else would violate the principle of least surprise)

2) Setting subproject options in the superproject default_options

I don't think we meant to allow this, but we did and it should continue to work. I've marked it for deprecation because we already have a documented way to do this via `subproject(..., default_options : ...)`, which is nicer anyway because you can control that in the flow of the program.

This is an alternative proposal to #7574

### DIFF
--- a/mesonbuild/compilers/compilers.py
+++ b/mesonbuild/compilers/compilers.py
@@ -21,13 +21,9 @@ from functools import lru_cache
 from .. import coredata
 from .. import mlog
 from .. import mesonlib
-from ..linkers import LinkerEnvVarsMixin
 from ..mesonlib import (
     EnvironmentException, MachineChoice, MesonException,
-    Popen_safe, split_args
-)
-from ..envconfig import (
-    Properties, get_env_var
+    Popen_safe
 )
 from ..arglist import CompilerArgs
 
@@ -79,22 +75,7 @@ for _l in clink_langs + ('vala',):
 clink_suffixes += ('h', 'll', 's')
 all_suffixes = set(itertools.chain(*lang_suffixes.values(), clink_suffixes))
 
-# Languages that should use LDFLAGS arguments when linking.
-languages_using_ldflags = {'objcpp', 'cpp', 'objc', 'c', 'fortran', 'd', 'cuda'}
-# Languages that should use CPPFLAGS arguments when linking.
-languages_using_cppflags = {'c', 'cpp', 'objc', 'objcpp'}
 soregex = re.compile(r'.*\.so(\.[0-9]+)?(\.[0-9]+)?(\.[0-9]+)?$')
-
-# Environment variables that each lang uses.
-cflags_mapping = {'c': 'CFLAGS',
-                  'cpp': 'CXXFLAGS',
-                  'cuda': 'CUFLAGS',
-                  'objc': 'OBJCFLAGS',
-                  'objcpp': 'OBJCXXFLAGS',
-                  'fortran': 'FFLAGS',
-                  'd': 'DFLAGS',
-                  'vala': 'VALAFLAGS',
-                  'rust': 'RUSTFLAGS'}
 
 # All these are only for C-linkable languages; see `clink_langs` above.
 
@@ -536,11 +517,6 @@ class Compiler(metaclass=abc.ABCMeta):
         """
         return []
 
-    def get_linker_args_from_envvars(self,
-                                     for_machine: MachineChoice,
-                                     is_cross: bool) -> T.List[str]:
-        return self.linker.get_args_from_envvars(for_machine, is_cross)
-
     def get_options(self) -> T.Dict[str, coredata.UserOption]:
         return {}
 
@@ -909,74 +885,17 @@ def get_largefile_args(compiler):
     return []
 
 
-def get_args_from_envvars(lang: str,
-                          for_machine: MachineChoice,
-                          is_cross: bool,
-                          use_linker_args: bool) -> T.Tuple[T.List[str], T.List[str]]:
-    """
-    Returns a tuple of (compile_flags, link_flags) for the specified language
-    from the inherited environment
-    """
-    if lang not in cflags_mapping:
-        return [], []
-
-    compile_flags = []  # type: T.List[str]
-    link_flags = []     # type: T.List[str]
-
-    env_compile_flags = get_env_var(for_machine, is_cross, cflags_mapping[lang])
-    if env_compile_flags is not None:
-        compile_flags += split_args(env_compile_flags)
-
-    # Link flags (same for all languages)
-    if lang in languages_using_ldflags:
-        link_flags += LinkerEnvVarsMixin.get_args_from_envvars(for_machine, is_cross)
-    if use_linker_args:
-        # When the compiler is used as a wrapper around the linker (such as
-        # with GCC and Clang), the compile flags can be needed while linking
-        # too. This is also what Autotools does. However, we don't want to do
-        # this when the linker is stand-alone such as with MSVC C/C++, etc.
-        link_flags = compile_flags + link_flags
-
-    # Pre-processor flags for certain languages
-    if lang in languages_using_cppflags:
-        env_preproc_flags = get_env_var(for_machine, is_cross, 'CPPFLAGS')
-        if env_preproc_flags is not None:
-            compile_flags += split_args(env_preproc_flags)
-
-    return compile_flags, link_flags
-
-
-def get_global_options(lang: str,
-                       comp: T.Type[Compiler],
-                       for_machine: MachineChoice,
-                       is_cross: bool,
-                       properties: Properties) -> T.Dict[str, coredata.UserOption]:
+def get_global_options(lang: str, comp: T.Type[Compiler], for_machine: MachineChoice,
+                       env: 'Environment') -> T.Dict[str, coredata.UserOption]:
     """Retreive options that apply to all compilers for a given language."""
     description = 'Extra arguments passed to the {}'.format(lang)
-    opts = {
+    return {
         'args': coredata.UserArrayOption(
             description + ' compiler',
-            [], split_args=True, user_input=True, allow_dups=True),
+            env.compiler_options[for_machine][lang].get('args', []),
+            split_args=True, user_input=True, allow_dups=True),
         'link_args': coredata.UserArrayOption(
             description + ' linker',
-            [], split_args=True, user_input=True, allow_dups=True),
+            env.compiler_options[for_machine][lang].get('link_args', []),
+            split_args=True, user_input=True, allow_dups=True),
     }
-
-    # Get from env vars.
-    compile_args, link_args = get_args_from_envvars(
-        lang,
-        for_machine,
-        is_cross,
-        comp.INVOKES_LINKER)
-
-    for k, o in opts.items():
-        user_k = lang + '_' + k
-        if user_k in properties:
-            # Get from configuration files.
-            o.set_value(properties[user_k])
-        elif k == 'args':
-            o.set_value(compile_args)
-        elif k == 'link_args':
-            o.set_value(link_args)
-
-    return opts
--- a/mesonbuild/compilers/compilers.py
+++ b/mesonbuild/compilers/compilers.py
@@ -892,10 +892,8 @@ def get_global_options(lang: str, comp: T.Type[Compiler], for_machine: MachineCh
     return {
         'args': coredata.UserArrayOption(
             description + ' compiler',
-            env.compiler_options[for_machine][lang].get('args', []),
-            split_args=True, user_input=True, allow_dups=True),
+            [], split_args=True, user_input=True, allow_dups=True),
         'link_args': coredata.UserArrayOption(
             description + ' linker',
-            env.compiler_options[for_machine][lang].get('link_args', []),
-            split_args=True, user_input=True, allow_dups=True),
+            [], split_args=True, user_input=True, allow_dups=True),
     }
--- a/mesonbuild/environment.py
+++ b/mesonbuild/environment.py
@@ -139,6 +139,7 @@ COMPILER_OPTS_TO_ENV = {
 
 OTHER_OPTS_TO_ENV = {
     'pkg_config_path': 'PKG_CONFIG_PATH',
+    'cmake_prefix_path': 'CMAKE_PREFIX_PATH',
 }  # type: T.Dict[str, str]
 
 def detect_gcovr(min_version='3.3', new_rootdir_version='4.2', log=False):
@@ -649,9 +650,13 @@ class Environment:
                 if p_env_pair is not None:
                     p_env_var, p_env = p_env_pair
 
+                    if mesonlib.is_windows():
+                        splitter = lambda p: p.split(';')
+                    else:
+                        splitter = lambda p: re.split(r':|;', p)
                     # PKG_CONFIG_PATH may contain duplicates, which must be
                     # removed, else a duplicates-in-array-option warning arises.
-                    p_list = list(mesonlib.OrderedSet(p_env.split(':')))
+                    p_list = list(mesonlib.OrderedSet(splitter(p_env)))
 
                     # Environment variables override config
                     meson_options[for_machine][''][key] = p_list
--- a/mesonbuild/environment.py
+++ b/mesonbuild/environment.py
@@ -124,6 +124,23 @@ build_filename = 'meson.build'
 
 CompilersDict = T.Dict[str, Compiler]
 
+COMPILER_OPTS_TO_ENV = {
+    'c': 'CFLAGS',
+    'cpp': 'CXXFLAGS',
+    'cuda': 'CUFLAGS',
+    'objc': 'OBJCFLAGS',
+    'objcpp': 'OBJCXXFLAGS',
+    'fortran': 'FFLAGS',
+    'd': 'DFLAGS',
+    'vala': 'VALAFLAGS',
+    'rust': 'RUSTFLAGS',
+    'pkg_config_path': 'PKG_CONFIG_PATH',
+}  # type: T.Dict[str, str]
+
+OTHER_OPTS_TO_ENV = {
+    'pkg_config_path': 'PKG_CONFIG_PATH',
+}  # type: T.Dict[str, str]
+
 def detect_gcovr(min_version='3.3', new_rootdir_version='4.2', log=False):
     gcovr_exe = 'gcovr'
     try:
@@ -561,7 +578,8 @@ class Environment:
         user_options = collections.defaultdict(dict)  # type: T.DefaultDict[str, T.Dict[str, object]]
 
         # meson builtin options, as passed through cross or native files
-        meson_options = PerMachineDefaultable()  # type: PerMachineDefaultable[T.DefaultDict[str, T.Dict[str, object]]]
+        meson_options = PerMachineDefaultable(
+            collections.defaultdict(dict), collections.defaultdict(dict))  # type: PerMachineDefaultable[T.DefaultDict[str, T.Dict[str, object]]]
 
         ## Setup build machine defaults
 
@@ -577,8 +595,9 @@ class Environment:
         _base_options = {}  # type: T.Dict[str, object]
 
         # Per language compiler arguments
-        compiler_options = PerMachineDefaultable()  # type: PerMachineDefaultable[T.DefaultDict[str, T.Dict[str, object]]]
-        compiler_options.build = collections.defaultdict(dict)
+        compiler_options = PerMachineDefaultable(
+            collections.defaultdict(dict),
+            collections.defaultdict(dict))  # type: PerMachineDefaultable[T.DefaultDict[str, T.Dict[str, object]]]
 
         ## Read in native file(s) to override build machine configuration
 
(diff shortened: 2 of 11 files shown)


xclaesse:
I personally strongly disagree with 1, machine file must override env, just like command line. They are not config files from /etc but passed in command line, that must be exactly as if passed in command line. What does others think?

About 2 I don't have opinion whether or not it should be deprecated, I would like to hear from people using that trick if they have reasons to not use default_options in subproject() or dependency(). I guess there is a slight benefit to have all defaults in the top of project instead of buried where it actually check for the suboroject.

xclaesse:
That being said, I personally strongly prefer my fix of course, not going in my direction means I won't be able to rebase it and I'll have to abandon my branches dealing with options. I hope you have plans to fix others regressions and per subproject compiler option.

But I like dealing with env in a single place for compilers, that's really neat!

dcbaker:
I understand why you disagree, and to be honest, I don't feel like there is a "right" answer to whether env before conf or vice versa. That being said almost everyone has agreed that env overrides conf (unix tradition and all that), so not honoring that makes meson stand out and have surprising behavior. the sort of surprising behavior that someone (probably a distro maintainer) will complain about.

mesonbuild/environment.py
```diff
@@ -682,27 +700,56 @@ def move_compiler_options(properties: Properties, compopts: T.Dict[str, T.Defaul
...
+
+            compiler_args_from_env = False
+            for key, envkey in COMPILER_OPTS_TO_ENV.items():
+                p_env_pair = get_env_var_pair(for_machine, self.coredata.is_cross_build(), envkey)
+                if p_env_pair is not None:
+                    p_env = p_env_pair[1]
+
+                    if self.first_invocation:
+                        # Environment variables override config
+                        self.compiler_options[for_machine][key]['args'] = split_args(p_env)
```
mensinda, line 731: The same issue as the previous comment.

mesonbuild/environment.py
```diff
@@ -682,27 +700,56 @@ def move_compiler_options(properties: Properties, compopts: T.Dict[str, T.Defaul
...
+            # Handle CPPFLAGS, which are only relavent for a couple of languages
+            p_env_pair = get_env_var_pair(for_machine, self.coredata.is_cross_build(), 'CPPFLAGS')
+            if p_env_pair is not None:
+                p_env = split_args(p_env_pair[1])
+                for l in ['c', 'cpp', 'objc', 'objcpp']:
+                    compargs = self.compiler_options[for_machine][l]
+                    # If the compiler args were set from the environment extend
+                    # them with CPPFLAGS, if they came from somewhere else
+                    # replace them.
+                    if compiler_args_from_env and 'args' in compargs:
```
mensinda, line 743: Shouldn't all this be done only during the `self.first_invocation` or am I missing something?

mesonbuild/environment.py
```diff
@@ -682,27 +700,56 @@ def move_compiler_options(properties: Properties, compopts: T.Dict[str, T.Defaul
...
-                    mlog.warning(
-                        p_env_var,
-                        'environment variable does not match configured',
-                        'between configurations, meson ignores this.',
-                        'Use -Dpkg_config_path to change pkg-config search',
-                        'path instead.'
-                    )
+                p_env = split_args(p_env_pair[1])
+                for l in ['objcpp', 'cpp', 'objc', 'c', 'fortran', 'd', 'cuda']:
+                    self.compiler_options[for_machine][l]['link_args'] = p_env
```
mensinda, line 752: Same `self.first_invocation` issue as the previous comment.

mensinda:
First of all, it is great that all the env var handling is now centralized but I also strongly disagree that environment variables should overwrite machine file entries for a few reasons:

- it changes the current behavior, which is also surprising and would at least require a release note snippet.
- There are some people (at least me) who set `CC` and `CXX` in their `.zshrc` (for building AUR packages, etc.). If I want to change that I can currently use a machine file. With this change I always have to remember to change the environment variables *and* pass the machine file for all the other stuff I want to set.
- machine files are, IMHO, not UNIX config files. Yes, meson can load then from [some default locations](https://mesonbuild.com/Native-environments.html#native-file-locations) for convenience, however, I don't know a distribution that ships with `/usr/share/meson/native` and most people are likely still going to pass them explicitly. And if I pass something explicitly via the command-line I expect it to overwrite environment variables.

As for the flattening language dicts debate, there are pros and cons for either solution and I don't have a strong opinion for what is the right one (as long as all the regressions are fixed).

tp-m:
> Setting subproject options in the superproject default_options
> 
> I don't think we meant to allow this, but we did and it should continue to work. I've marked it for deprecation because we already have a documented way to do this via `subproject(..., default_options : ...)`, which is nicer anyway because you can control that in the flow of the program.

I have never made use of this feature (didn't know we could do this, actually), but I feel we should keep it and keep supporting it, because there might be multiple tiers of `subproject()`ing and the superproject might not control the actual `subproject()` call because it's done from a subproject, but might still have an interest in overriding options for it.

xclaesse:
Thanks @mensinda, exactly what I think too. That also confirms what I though, your comment there was not an actual review of that PR: https://github.com/mesonbuild/meson/pull/6597#pullrequestreview-459628355. So it really got merged unreviewed.

Note that here the question is not about flattening the lang dict in coredata, that's another debat, and I rekon it can be argued that it the lang dict has advantages too. No, here the question is to compare https://github.com/mesonbuild/meson/pull/6597 and https://github.com/mesonbuild/meson/pull/7293 and decide which approach has more maintainable code. I don't understand why I'm the only one to step up for this, I hope people here have actually read those PRs and understand the difference. But if everyone really think @dcbaker approach is better, then I'll have to accept it I guess...

mensinda:
> So it really got merged unreviewed.

I don't think that this is entirely correct since there were already 2 other approvals. For more context: I wasn't really an expert in the machine file/options/environment variables code (and arguably I am still not), so I have only reviewed the files where I am a code owner and only briefly checked if the other parts roughly made sense. I requested some changes since I found a minor issue in the `mintro` code: https://github.com/mesonbuild/meson/pull/6597#pullrequestreview-459583488 (and yes, I should not have written the last sentence of this comment). Once this was fixed I approved the PR since there were already 2 other approvals.

Now looking back, I should have definitely been more clear about that I have only really reviewed the `mintro` code. I will make sure that I communicate a (more or less) partial PR review more clearly in the future.

xclaesse:
@mensinda makes sense, you indeed get requested automatically for review when touching certain parts of code source. Thanks for the clarification :-) I should myself have used the "request changes" github feature to make it clear that I did not approve those changes, since my comments got buried in long history.

dcbaker:
It seems like most people think that machine files should override env, and I don't like it, but that seems to be the project concensus. The only last question I have before I go that route, since most people seem to agree on that, is what happens if/when we land a site-native file? (there's an open <R for this) which is automatically loaded for your native file. that would silently override any env var settings you have. I'd also like to get @eli-schwartz (as a distro person) and @jpakkane's opinion. Since currently only pkg-config and cmake interact with machine files, I'd like to make the decision once and not have to change it after 0.56.

mesonbuild/environment.py
```diff
@@ -682,27 +700,56 @@ def move_compiler_options(properties: Properties, compopts: T.Dict[str, T.Defaul
...
         for for_machine in MachineChoice:
-            p_env_pair = get_env_var_pair(for_machine, self.coredata.is_cross_build(), 'PKG_CONFIG_PATH')
+            for key, envkey in OTHER_OPTS_TO_ENV.items():
+                p_env_pair = get_env_var_pair(for_machine, self.coredata.is_cross_build(), envkey)
+                if p_env_pair is not None:
+                    p_env_var, p_env = p_env_pair
+
+                    # these may contain duplicates, which must be removed, else
+                    # a duplicates-in-array-option warning arises.
+                    p_list = list(mesonlib.OrderedSet(re.split(r':|;', p_env)))
```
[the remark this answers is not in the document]
dcbaker: I didn't even think about drive leters, but you're right, I'll fix that.

mesonbuild/environment.py
dcbaker: The more I think about it, the more I want to wrap this entire block in `if self.first_invocation`, hmm.

xclaesse: fwiw, I do have plans to remove all those `self.first_invocation`, meson can reevaluate options and pick new defaults since we have `meson-private/cmd_line.txt` that tells what are the options given by the user. This is a recurring issue people constantly raise, changing default_options has no effect until you --wipe.

eli-schwartz:
I'd be inclined to think files which are loaded via `meson setup --native-file <profile-name|file>` constitute the equivalent of `-Dopt=val` passed on the command line, just in different form. It should override environment variables.

A file e.g. /etc/meson/distro-defaults.conf which is automatically loaded, however, should emphatically NOT be treated like a native file specified on the command line; it must not operate in the background, per default, to override environment settings.

So in order, the precedence is:
- OS/user `~/.config` **defaults** autoloaded in the background
- per-**environment** settings
- per-**invocation** command-line arguments


An argument for native files being overridden by environment options would need to argue the case that it's more similar to OS/`~/.config` settings than to `-Dopt=val`. And it should argue from the perspective of how people communicate their intent to the build system, not from "but the implementation format uses files instead of strings".

@dcbaker does this make sense?

dcbaker:
@eli-schwartz, yeah that makes sense. I still think it should be the other way, but since I'm the only one I'll make machine files override environment variables.

Ericson2314:
@eli-schwartz said exactly what I wanted to say. Thanks!

@dcbaker one precedent is response files---specifying configure files on the command line is not *that* different than putting CLI args in a separate file, except the former is more structured and beautiful. And response files do take precedence over env vars. How do you feel about that?

mesonbuild/environment.py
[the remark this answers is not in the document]
dcbaker: I've refactored this to move the environments before machine file consumption, and wrapped the env reading in first_invocation.

dcbaker:
Alright, I've addresed the override order, env is now overriden by machine files, I've also fixed the first_invocation issues.

Ericson2314:
@xclaesse 

> I don't think this is the right approach. I think we should first merge #7574 and then refactor handling of env vars on top of that, because that's the really good part of this PR.

I'm glad we agree that agree the refactors are more important than this PR, but I think PR does great work in making the current nested data structures work better. I know we'll disagree but for that reason I rather see *this* PR merged first, and then take a look at your PR on top of it.

dcbaker:
And fix the lgtm warnings

Ericson2314:
@dcbaker is there anything we can do to in fact make the automatic ones lower priority?

dcbaker:
@Ericson2314 yes. The good news is that hasn't landed yet, and I don't plan to push on it till after we get the refactor of Environment and CoreData done

Ericson2314:
Oh great, didn't realize that hadn't landed yet.

mesonbuild/environment.py
```diff
@@ -650,6 +716,7 @@ def move_compiler_options(properties: Properties, compopts: T.Dict[str, T.Defaul
         ## Read in cross file(s) to override host machine configuration
 
         if self.coredata.cross_files:
+            read_env_vars(MachineChoice.HOST)
```
Ericson2314, line 719: Any reason why the native file `if`-block doesn't also begin with this?

mesonbuild/environment.py
```diff
@@ -628,6 +647,54 @@ def move_compiler_options(properties: Properties, compopts: T.Dict[str, T.Defaul
...
+                p_env = split_args(p_env_pair[1])
+                for l in ['objcpp', 'cpp', 'objc', 'c', 'fortran', 'd', 'cuda']:
+                    compiler_options[for_machine][l]['link_args'] = p_env
+
+
+        # Environment variables are only read on first invocation
+        if self.first_invocation:
+            if self.coredata.cross_files:
+                read_env_vars(MachineChoice.BUILD)
+            read_env_vars(MachineChoice.HOST)
```
Ericson2314, line 696: I think you want
```suggestion
            read_env_vars(MachineChoice.BUILD)
            if self.coredata.cross_files:
                read_env_vars(MachineChoice.Host)
```
so that the defaulting logic will copy the build configuration to host configuration just like when there is a native file but not cross file.

mesonbuild/environment.py
```diff
@@ -668,6 +734,11 @@ def move_compiler_options(properties: Properties, compopts: T.Dict[str, T.Defaul
             split_compiler_options(meson_options.host, MachineChoice.HOST)
             move_compiler_options(properties.host, compiler_options.host)
 
+        if not meson_options.host:
+            meson_options.host = None
+        if not compiler_options.host:
+            compiler_options.host = None
```
Ericson2314, line 740: If you don't initialize the `PerMachineDefaultable` above, you won't need this.

mesonbuild/compilers/compilers.py
```diff
@@ -79,23 +75,8 @@
...
-# Environment variables that each lang uses.
-cflags_mapping = {'c': 'CFLAGS',
-                  'cpp': 'CXXFLAGS',
-                  'cuda': 'CUFLAGS',
-                  'objc': 'OBJCFLAGS',
-                  'objcpp': 'OBJCXXFLAGS',
-                  'fortran': 'FFLAGS',
-                  'd': 'DFLAGS',
-                  'vala': 'VALAFLAGS',
-                  'rust': 'RUSTFLAGS'}
```
xclaesse, line 97: I'm not sure if those lists are better in compilers or environment. I have a slight preference in keeping this here (capitalize the dict name is fine), but it can be seen both ways I guess.

mesonbuild/compilers/compilers.py
```diff
@@ -909,74 +885,17 @@ def get_largefile_args(compiler):
...
+def get_global_options(lang: str, comp: T.Type[Compiler], for_machine: MachineChoice,
+                       env: 'Environment') -> T.Dict[str, coredata.UserOption]:
     """Retreive options that apply to all compilers for a given language."""
     description = 'Extra arguments passed to the {}'.format(lang)
-    opts = {
+    return {
         'args': coredata.UserArrayOption(
             description + ' compiler',
-            [], split_args=True, user_input=True, allow_dups=True),
+            env.compiler_options[for_machine][lang].get('args', []),
```
xclaesse, line 895: You don't need to set the value here, it's going to be set again in add_lang_args(). Other compiler options leave their default value unset, I think it's more consistent to not use env.compiler_options here but only in coredata.

mesonbuild/compilers/compilers.py
```diff
@@ -909,74 +885,17 @@ def get_largefile_args(compiler):
...
-        lang,
-        for_machine,
-        is_cross,
-        comp.INVOKES_LINKER)
-
-    for k, o in opts.items():
-        user_k = lang + '_' + k
-        if user_k in properties:
-            # Get from configuration files.
-            o.set_value(properties[user_k])
```
xclaesse, line 976: Do we still take values from properties somewhere? We deprecated [paths] in machine files, but I don't remember what we decided about c_args in [properties], IMHO it should be deprecated too, and incorporated into env.compiler_options, but I'm not sure that has been done.

mesonbuild/environment.py
```diff
@@ -124,6 +124,24 @@
...
+    'c': 'CFLAGS',
+    'cpp': 'CXXFLAGS',
+    'cuda': 'CUFLAGS',
+    'objc': 'OBJCFLAGS',
+    'objcpp': 'OBJCXXFLAGS',
+    'fortran': 'FFLAGS',
+    'd': 'DFLAGS',
+    'vala': 'VALAFLAGS',
+    'rust': 'RUSTFLAGS',
+    'pkg_config_path': 'PKG_CONFIG_PATH',
```
xclaesse, line 137: I don't think 'pkg_config_path' should be here.

mesonbuild/environment.py
```diff
@@ -561,7 +579,8 @@ def __init__(self, source_dir, build_dir, options):
         user_options = collections.defaultdict(dict)  # type: T.DefaultDict[str, T.Dict[str, object]]
 
         # meson builtin options, as passed through cross or native files
-        meson_options = PerMachineDefaultable()  # type: PerMachineDefaultable[T.DefaultDict[str, T.Dict[str, object]]]
+        meson_options = PerMachineDefaultable(
+            collections.defaultdict(dict), collections.defaultdict(dict))  # type: PerMachineDefaultable[T.DefaultDict[str, T.Dict[str, object]]]
```
xclaesse, line 583: More complicated structure again, no idea what that means... at least put a comment to tell what it contains.

mesonbuild/environment.py
```diff
@@ -628,6 +647,54 @@ def move_compiler_options(properties: Properties, compopts: T.Dict[str, T.Defaul
...
+        def read_env_vars(for_machine: MachineChoice) -> None:
+            for key, envkey in OTHER_OPTS_TO_ENV.items():
+                p_env_pair = get_env_var_pair(for_machine, self.coredata.is_cross_build(), envkey)
+                if p_env_pair is not None:
+                    p_env_var, p_env = p_env_pair
+
+                    if mesonlib.is_windows():
+                        splitter = lambda p: p.split(';')
+                    else:
+                        splitter = lambda p: re.split(r':|;', p)
```
xclaesse, line 659: There was a link to a bug in the code you deleted above regarding that split. I haven't read what it's about, but I guess it's better to keep it. I find it suspect that we split on both chars, ideally it should be only os.pathsep.

mesonbuild/environment.py
```diff
@@ -124,6 +124,24 @@
...
+    'objc': 'OBJCFLAGS',
+    'objcpp': 'OBJCXXFLAGS',
+    'fortran': 'FFLAGS',
+    'd': 'DFLAGS',
+    'vala': 'VALAFLAGS',
+    'rust': 'RUSTFLAGS',
+    'pkg_config_path': 'PKG_CONFIG_PATH',
+}  # type: T.Dict[str, str]
+
+OTHER_OPTS_TO_ENV = {
```
xclaesse, line 140: I'm not a big fan of "OTHER" in the name. other than what? I'm also not a big fan of each list of env dealt with duplicated but slightly different code below.

Wondering if we could consolidate all this into a single mapping:
```
ENV_TO_OPTION = {
  'CFLAGS': ['c_args'],
  ...
  'LDFLAGS': [lang + '_link_args' for lang in languages_using_ldflags]
  ...
  'PKG_CONFIG_PATH': ['pkg_config_path']
}
```
That structure can be processed in a single loop, do the value splitting if `key.endswith('_PATH')`

mesonbuild/compilers/compilers.py
Ericson2314: @xclasses The idea is we *really* don't want arbitrary code accessing environment variables. All config, by it from CLI, env vars, or configuration files, should be gathered in one place, and the rest of the code should not have to care about where the raw options have come from.

Having completely different code paths for the different ways to configure has been a huge source of bugs in the past, and I believe is one of the reasons environment variables have gotten such a bad rep around here. (Though don't get me wrong, they are still iffy in general.)

xclaesse: I'm all for having the code in a single place. The point here is the mapping language<>env is more a compiler matter. The code actually taking the value is an Envorinment.py matter. But I guess that can be argued both ways, I don't have strong opinion against this change.

mesonbuild/environment.py
mensinda: Specifically in CMake `;` is used as a list separator on all platforms since this is a core language "feature". So we have to expect that all env vars and config options related to CMake can use `;` as a list separator.

xclaesse: Ok, worth adding a comment there mentioning cmake then.

mesonbuild/compilers/compilers.py
dcbaker: We deprecated both the entire paths sections and setting language args in [properties] allready.

mesonbuild/environment.py
dcbaker: You're right, I keep rebasing that back *in* somehow :(

mesonbuild/environment.py
dcbaker: I'll add a comment to explain, but I've actually just moved the creation of the defaultdicts here, rather than being conditional.

mesonbuild/environment.py
dcbaker: Oh, you're right about the cflags/ldflags stuff for sure. In fact, using a dict is the wrong data structure, a list of tuples works much nicer.

I'm not so sure combining the _PATH variables together makes sense though. We do end up duplicated a couple of lines of code (about 4) and end up with two loops. But I think that's preferable to having an extra if path on every iteration of one loop. Look at the version I'm pushing now and see what you think.

dcbaker: Yeah, `OTHER` isn't good. I changed it to `PATH_OPTS_TO_ENV`.

mesonbuild/compilers/compilers.py
dcbaker: I like moving it just to prevent spaghetti and the X imports Y imports X problem. I can make this module "private" if that makes it better?

mesonbuild/environment.py
dcbaker: This allows us to delete a ton of:
```python
if compiler_options.host is None:
    compiler_options.host = defaultdict(dict)
```

Which I think is nice.

mesonbuild/environment.py
dcbaker: yup, fixed in the latest version

dcbaker:
I know I didn't get to everything, but my battery is almost dead and it's getting late.

mesonbuild/compilers/compilers.py
dcbaker: the `move_compiler_options` closure does that.

PR #7597, opened 2020-08-15
find_library: Print type of library not found

nirbheek:
If we can't find a static library, we should say that. It's confusing otherwise.


nirbheek:
Test failure is spurious.

PR #7600, opened 2020-08-15
Add C2x option support.

alexrp:
`c2x`/`gnu2x` were added in GCC 9 / Clang 9.

Also updated the list of LLVM tool versions while I was here, since LLVM 10 has since been released.


mesonbuild/compilers/c.py
```diff
@@ -147,6 +151,7 @@ class AppleClangCCompiler(ClangCCompiler):
 
     _C17_VERSION = '>=10.0.0'
     _C18_VERSION = '>=11.0.0'
+    _C2X_VERSION = '>=12.0.0'
```
alexrp, line 154: Note that this is guesswork. I don't have a Mac to check, so I have no actual idea if Xcode 12 ships with the Apple equivalent of Clang 9.

dcbaker:
@nirbheek @jpakkane, do either of you have a macbook that has a new enough version of xcode to check this?

nirbheek:
Is this safe enough to go into the stable release?

michaelbrockus:
@nirbheek we probably won’t know tell we have a test case to validate it works.

PR #7602, opened 2020-08-16
backends: check external rpaths for all languages using ldflags

jameshilliard:


jameshilliard:
@dankegel Might be a good idea for you to review this.

dankegel:
A review comment on https://github.com/mesonbuild/meson/pull/7472 already suggested this, and it sounded good to me back then, too!

PR #7603, opened 2020-08-16
unittests: continue on exception instead of break

jameshilliard:


jameshilliard:
@jpakkane the test was partially broken and was skipping some tests, the code it is testing is not broken however

Issue #7601, opened 2020-08-16
CMake module: false behaviour of CMakeTraceParser._guess_files()


gerion0:
**Describe the bug**
Meson fails to build a CMake project with `add_custom_command`.
I'll start with a minimal example, then the description is much easier.

**To Reproduce**

Execute this steps:
```
$ mkdir mwe && cd mwe
$ mkdir subprojects
$ git clone https://github.com/metabrainz/libmusicbrainz subprojects/libmusicbrainz
```
Then create the following meson.build file:
```
project('mwe', 'cpp',
  version : '0.1',
  default_options : ['warning_level=3',
                     'cpp_std=c++17'])

cmake = import('cmake')
libmusicbrainz_proj = cmake.subproject('libmusicbrainz')
libmusicbrainz = libmusicbrainz_proj.dependency('musicbrainz5')
```
Building this project fails (I'm using meson git head):
```
% ~/sourcecode/meson/meson.py build
The Meson build system
Version: 0.55.999
Source dir: /home/ci3nt/test/mwe
Build dir: /home/ci3nt/test/mwe/build
Build type: native build
Project name: mwe
Project version: 0.1
C++ compiler for the host machine: ccache c++ (gcc 9.3.0 "c++ (Gentoo 9.3.0 p2) 9.3.0")
C++ linker for the host machine: c++ ld.bfd 2.33.1
Host machine cpu family: x86_64
Host machine cpu: x86_64

|Executing subproject libmusicbrainz method cmake 
|
|Found CMake: /usr/bin/cmake (3.16.5)
|
||Configuring the build directory with CMake version 3.16.5
||Running: -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER=c++ -G Ninja -DCMAKE_INSTALL_PREFIX=/usr/local
||  - build directory:          /home/ci3nt/test/mwe/build/subprojects/libmusicbrainz/__CMake_build
||  - source directory:         /home/ci3nt/test/mwe/subprojects/libmusicbrainz
||  - trace args:               --trace --trace-expand --no-warn-unused-cli --trace-redirect=cmake_trace.txt
||  - preload file:             /home/ci3nt/test/mwe/build/meson-private/data/preload.cmake
||  - disabled policy warnings: [CMP0025, CMP0047, CMP0056, CMP0060, CMP0065, CMP0066, CMP0067, CMP0082, CMP0089]
||
||Running with trace output on.
[...]

PR #7604, opened 2020-08-16
cmake/traceparser: warn user for too old cmake

gerion0:
CMake <3.17 is error prone in some cases. Warn the user about that.

I'm not entirely happy with this commit, since it results in a lot of warnings, because `_guess_files` is often called multiple times.
However, `_guess_files` is the place, where the errors can occur and Meson needs to guess and the warning will appear less and less in the future.

See also: #7601

### DIFF
--- a/docs/markdown/CMake-module.md
+++ b/docs/markdown/CMake-module.md
@@ -237,3 +237,11 @@ myProject.cmake.in:
 
 set(MYVAR VAR)
 ```
+
+### Additional hints
+
+Meson needs to parse the CMake build files in certain cases to mimic its behaviour, for example for the CMake function: `add_custom_command`.
+It therefore executes CMake and interprets its output.
+Before CMake 3.17, this output is ambiguous in some cases.
+Meson therefore needs to guess the correct meaning and may fail in doing this.
+If you encounter this problem, the recommended way is to upgrade CMake to a minimum version of 3.17.
--- a/mesonbuild/cmake/traceparser.py
+++ b/mesonbuild/cmake/traceparser.py
@@ -655,6 +655,12 @@ class CMakeTraceParser:
         if self.trace_format != 'human':
             return broken_list
 
+        mlog.log_once(
+            'Detected version of CMake',
+            mlog.bold(self.cmake_version), 'is below 3.17.',
+            'This can cause problems in the build, see https://mesonbuild.com/CMake-module.html#additional-hints.'
+        )
+
         # Try joining file paths that contain spaces
 
         reg_start = re.compile(r'^([A-Za-z]:)?/(.*/)*[^./]+$')


mensinda:
I don't think that a warning is reasonable here since there are a lot of projects that work (more or less) fine with CMake <3.17. A short note (as in `mlog.log`) when selecting the CMake executable in `executor.py` which references the (also updated docs with more details) would be more reasonable.

You can print a warning/log only once by passing `once=True` as a kwarg to the `mlog` functions.

gerion0:
> I don't think that a warning is reasonable here since there are a lot of projects that work (more or less) fine with CMake <3.17. A short note (as in `mlog.log`) when selecting the CMake executable in `executor.py` which references the (also updated docs with more details) would be more reasonable.

I have changed the log level to log instead of warning and have extended the documentation. However, I think executor is the wrong place for the message. First there already is a check for 3.17 in traceparser that needs to be dublicated, when warn in executor and second is the point where the guessing happens the `_guess_files` function in traceparser. So as a Meson developer (or anyone who looks through the code) it is much more understandable, if the warning for ambiguous parsing is emitted in exactly this part where the guessing is done.

> You can print a warning/log only once by passing `once=True` as a kwarg to the `mlog` functions.

Actually not. All other log levels of `mlog` have the `once` kwarg, but `log` has another function `log_once`. Anyway, I have used it.

mensinda:
> So as a Meson developer (or anyone who looks through the code) it is much more understandable, if the warning for ambiguous parsing is emitted in exactly this part where the guessing is done.

The first problem with this is that `_guess_files` is not the only place where different code paths based on CMake versions are executed. This is because the CMake trace format is not the only CMake feature that has changed. There is, for instance, `--trace-redirect` which was introduced in CMake 3.16 and the CMake file API which was introduced in CMake 3.14 (see the [`CMakeInterpreter`](https://github.com/mesonbuild/meson/blob/master/mesonbuild/cmake/interpreter.py#L848)).

The second problem is that this message will also be printed for CMake dependencies. There `_guess_files` is far less fragile since the file paths are far more predictable and have file extensions in 99.9% of all cases. So, printing this message there would not be helpful for the end-user.

Also, as an end-user, I would expect meson to warn/notify me about CMake version requirements close to where CMake was detected and not after calling CMake and somewhere in the middle of the logs for processing a CMake dependency/subproject.

Just hardcoding CMake 3.17 in `CMakeExecutor` might indeed not have been the best call, however, the `CMakeExecutor` could take a new (optional) `recommended_version` argument which would be more flexible. Alternatively, the version check can also be put in the `CMakeInterpreter` with the other version checks.

eli-schwartz:
This is obsolete since #9906 because we now warn you for cmake <3.17 everywhere.

PR #7606, opened 2020-08-17
interpreter: Don't used cache if fallback is forced

hdhauk:
This PR fixes a bug where forcing fallback of dependencies doesn't work if that dependency is already cached.


xclaesse:
That's not a bug, if we already returned the system dependency previously, we cannot fallback anymore. That would mix system and subprojects version of the same library.

xclaesse:
You probably miss fallback argument in a previous call to dependency. You can instead add the [provide] section in your wrap file to ensure it always implicitly have the fallback information. See https://mesonbuild.com/Wrap-dependency-system-manual.html#provide-section

hdhauk:
Gotcha! Thanks for the insight

PR #7607, opened 2020-08-17
Various speedups from profiling QEMU's meson.build

bonzini:
These patches speed up Meson by about 15% on a QEMU build (or even 25% when running under cProfile). In general the code doesn't become worse, or when it does it was already not very beautiful or Pythonic.


xclaesse:
At first glance it looks good changes.

I'm wondering how do you measure, because when I use --profile-self, or just `time meson setup ...` I get widely different result each run. I full gst-build takes between 50-60 sec, that's a 20% error margin...

Also, as always with performance, small innocent changes that have a real unsuspected impact once profiling. I'm wondering if we could setup some reliable CI to know when we regress significantly on performance.

bonzini:
> I'm wondering how do you measure, because when I use --profile-self, or just time meson setup ... I get widely different result each run. I full gst-build takes between 50-60 sec, that's a 20% error margin...

For each individual patch, I'm using `python -m cProfile meson setup` and checking the time spent on individual calls (and the number of calls to the callees) over a few different runs. I also timed `meson setup` several times and the variance is +/- 0.5s (so the improvement from the patches is statistically significant).

xclaesse:
ok, got tricked by frequency scaling on my laptop. When testing on a more powerful desktop machine, I get stable numbers.

mesonbuild/arglist.py
```diff
@@ -140,12 +140,15 @@ def flush_pre_post(self) -> None:
 
         #pre and post will overwrite every element that is in the container
         #only copy over args that are in _container but not in the post flush or pre flush set
+        if len(pre_flush_set) or len(post_flush_set):
```
dcbaker, line 143:
```suggestion
        if pre_flush_set or post_flush_set:
```

using `len()` to check the emptiness of a container is an anit-pattern, and much slower than just `if container`

mesonbuild/backend/ninjabackend.py
```diff
@@ -114,21 +114,25 @@ def get_rsp_threshold():
 # from, etc.), so it must not be shell quoted.
 raw_names = {'DEPFILE_UNQUOTED', 'DESC', 'pool', 'description', 'targetdep'}
 
+NINJA_QUOTE_BUILD = re.compile("[$ :\n]")
+NINJA_QUOTE_VAR = re.compile("[$ \n]")
```
dcbaker, line 118:
```suggestion
NINJA_QUOTE_BUILD = re.compile(r"[$ :\n]")
NINJA_QUOTE_VAR = re.compile(r"[$ \n]")
```

Values passed to `re` methods as regular expressions should be "raw" strings.

PR #7609, opened 2020-08-18
Fix mapping of apple compilers in cmake

dcbaker:
The Apple situation is more complicated than it seems. There aren't actually that many linkers that work on macOS, so everyone just used ld64 (there is lld.ld64 as well, but we don't support that). This means that the current code will misdetect ICC and GCC as AppleClang. To make this even more complicated there is also vanilla clang for macOS, which also uses ld64, which means we can't even do `if cc.id == 'clang' and cc.linker.id == 'ld64'`, because that will still catch things it shouldn't. The only way to do this, literally, is to do an isinstance check against the compiler instance.

To make this work correctly for objc and ojbc++ I've added apple wrappers that do nothing but denote that this is the Apple version. maybe someday they'll do more than that.


mesonbuild/environment.py
```diff
@@ -1535,7 +1537,10 @@ def _detect_objc_or_objcpp_compiler(self, for_machine: MachineInfo, objc: bool)
                     exe_wrap, defines, linker=linker)
             if 'clang' in out:
                 linker = None
-                comp = ClangObjCCompiler if objc else ClangObjCPPCompiler
+                if 'Apple' in out:
+                    comp = AppleClangObjCCompiler if lang == 'c' else AppleClangObjCPPCompiler
```
nirbheek, line 1541: There's no 'lang' defined here. That's causing the CI failure.

dcbaker: oops, you're right of course, the objc variant uses a bool. Which I should know because I wrote...

nirbheek:
Since https://github.com/mesonbuild/meson/pull/7549 went into 0.55.1, should this go into 0.55.2?

mensinda:
I would say yes.

PR #7615, opened 2020-08-19
RFC: Rewrite buildtype as debug+optimization pair in user inputs

xclaesse:
This deprecate get_option('buildtype') to avoid the trap when it's set
    to 'custom'. It is still useful for the user to write
    -Dbuildtype=release as a way to declare intention rather than having to
    dive into optimization level details.
    
    Note that this makes mixing buildtype and debug/optimization in command
    line or in default_options a hard error, but not when using buildtype in
    command line and debug/optimization in default_options (or vice versa).
    In other words, `default_options: ['buildtype=debugoptimized']` with
    `-Ddebug=false` is not an error and results in buildtype='custom',
    debug=false and optimization=2.

Note that this is based on top of https://github.com/mesonbuild/meson/pull/7574, only the top commit is relevant.

If we agree on this behavior, doc & tests will have to be adapted.

### DIFF
--- a/mesonbuild/ast/introspection.py
+++ b/mesonbuild/ast/introspection.py
@@ -120,7 +120,7 @@ class IntrospectionInterpreter(AstInterpreter):
                         self.do_subproject(i)
 
         self.coredata.init_backend_options(self.backend)
-        options = {k: v for k, v in self.environment.meson_options.host[''].items() if k.startswith('backend_')}
+        options = {k: v for k, v in self.environment.raw_options.items() if k.startswith('backend_')}
 
         self.coredata.set_options(options)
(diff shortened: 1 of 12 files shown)


eli-schwartz (approved):
I like it, people shouldn't be doing this. It just makes current recommendations into policy, anyway. :)

xclaesse:
It's actually more broken than I thought, buildtype is used by compilers in many places which won't work if a target uses `override_options['debug=false']` for example.

bonzini:
Done mostly by commit d37d649b08b832d52fa684bc0506829fb40d5261 and refined later (see for example commit eccab11e535abe45e22e04333a29319c11972f0b).

PR #7618, opened 2020-08-20
Backport to 0.55 mtest: TestResult.SKIP is not a failure

3v1n0:
This is a pretty critical issue for bulding packages using TAP tags, so please backport the PR #7525.


jpakkane:
The real issue was that #7525 was not tagged for 0.55.1. @nirbheek what would be the correct way to fix this?

dcbaker:
I think the correct way to fix this is to add it to the 0.55.2 milestone, which I've done. if that's changed @nirbheek will tell us :)

PR #7626, opened 2020-08-21
Virtualization <3 Meson [skip ci]

bonzini:
There are a couple new users of Meson that might be worth mentioning.

502 commits later, Libvirt and QEMU have both switched their build systems!


amitdo:
The alphabetic order is wrong here.

amitdo:
The alphabetic order is wrong here.

Issue #7633, opened 2020-08-24
Passing base options in non -D-style fashion fails


blu-base:
**Describe the bug**

The the `--option=value` and `--option value` argument styles are not recognized for the base options. Only `-Doption=value` works.

However, the manual states that all three options should work for options (and it does in most cases).

I found this Stackoverflow question, which reported this behavior: https://stackoverflow.com/questions/63549328
Since i can confirm this SO issue, i had the impression, this behavior is not intended. A brief search did reveal similar reports.

**To Reproduce**

To reproduce, use any meson build setup. It shows even in the default, try this:

```
mkdir optiontest && cd optiontest
meson init
meson build --b_lto=true
```
```
usage: meson [-h] {setup,configure,dist,install,introspect,init,test,wrap,subprojects,help,rewrite,compile} ...
meson: error: unrecognized arguments: --b_lto=true
```

Moreover, ` meson build  --b_lto true` does not recognize the argument as well.

However, `meson build -Db_lto=true` works as intended, and configured successfully.

This issue does not just occur with b_lto.
   
  
**Expected behavior**

The following three commands should have the same result:

- `meson build --b_lto=true`
- `meson build --b_lto true`
- `meson build -Db_lto=true` 

**system parameters**

* OpenSUSE Tumbleweed
* python: 3.8.5
* meson 0.55.1
* ninja 1.10.0

-----

Alternatively, if `-D option` is the intended way to specify options, then the manual at this [location](https://mesonbuild.com/Builtin-options.html) might be changed to point out the suggested way, and the other styles only might work.  `meson setup --help` is showing only `[-D option]` for general use as well.

xclaesse:
`--option` only applies to built-in options. So the first 2 tables: directories and core options.

blu-base:
I see, i did not notice the same head levels of Universial options and Base options.

Would it be reasonable to propose a change in https://github.com/mesonbuild/meson/blob/master/docs/markdown/Builtin-options.md to point this out?

Since some options have a different label in `--option` and `-D option`, considering the labels with underscores (in the tables).
compare the specified arguments of `meson setup --help` with the table in https://github.com/mesonbuild/meson/blob/master/docs/markdown/Builtin-options.md


I would propose something like this:

[...]

PR #7634, opened 2020-08-24
Clarify the use of -Doption and alternatives for Builtin options passed as arguments [skip ci]

blu-base:
The -Doption=value style argument works for all builtin options,
and compiler options passed to meson. And some universal options,
have additional ways to be passed. However, base options (and compiler
options) do not have exactly the same ways of passing as universal
options.

This change adds a few pieces of information, which might get lost,
if the manual is not read serially.

Please See #7633 for reference.


jpakkane:
Thanks.

PR #7639, opened 2020-08-25
Fix typo when fetching buildtype option for compiler checks

nirbheek:
This typo happened in https://github.com/mesonbuild/meson/pull/7432 and wasn't noticed because I didn't add a test for it. Rectified now.

If we don't specify the CRT, MSVC will pick `/MT` by default (!?) and link to `libcmt.lib`. This actually *breaks* UWP because `libcmt.lib` is not available by default when building for UWP.

Was noticed here: https://github.com/cisco/libsrtp/pull/505


seungha-yang:
Still works fine with Windows arm64 native build (gst-build) 👍

Issue #7631, opened 2020-08-24
[meson|cmake] Unable to select _DEBUG configuration for other platforms 


Neumann-A:
due to #7546 there is currently no way to select `_DEBUG` configurations for other platforms than windows if buildtype is set to `custom` or `plain `or for windows if `b_vscrt='none'` 

**cc:** @mensinda @nirbheek 

maybe meson should get it things together what `debug` means and maybe decided that there is need for another option which is called `debuginfo` instead of the current convoluted mess of options. Or simply let the user provide a cmake configuration mapping via a native_file or builtin-option

nirbheek:
Hm, that's true. We should only do the above magic when the `b_vscrt` option is available. On other platforms, we should just use the `debug` option.

PR #7640, opened 2020-08-25
cmake: Don't link DEBUG to CRT when not building with MSVC

nirbheek:
cmake: Don't link DEBUG to CRT when not building with MSVC

is_debug doesn't just control the CRT, it also controls the 'debug
configuration' which is unrelated to the CRT setting on non-MSVC.

Fixes https://github.com/mesonbuild/meson/issues/7631

### DIFF
--- a/mesonbuild/dependencies/base.py
+++ b/mesonbuild/dependencies/base.py
@@ -1450,10 +1450,12 @@ class CMakeDependency(ExternalDependency):
                     cfgs = [x for x in tgt.properties['IMPORTED_CONFIGURATIONS'] if x]
                     cfg = cfgs[0]
 
-                is_debug = self.env.coredata.get_builtin_option('buildtype') == 'debug'
                 if 'b_vscrt' in self.env.coredata.base_options:
+                    is_debug = self.env.coredata.get_builtin_option('buildtype') == 'debug'
                     if self.env.coredata.base_options['b_vscrt'].value in ('mdd', 'mtd'):
                         is_debug = True
+                else:
+                    is_debug = self.env.coredata.get_builtin_option('debug')
                 if is_debug:
                     if 'DEBUG' in cfgs:
                         cfg = 'DEBUG'


mesonbuild/dependencies/base.py
```diff
@@ -1450,10 +1450,12 @@ def _detect_dep(self, name: str, modules: T.List[T.Tuple[str, bool]], components
                     cfgs = [x for x in tgt.properties['IMPORTED_CONFIGURATIONS'] if x]
                     cfg = cfgs[0]
 
-                is_debug = self.env.coredata.get_builtin_option('buildtype') == 'debug'
                 if 'b_vscrt' in self.env.coredata.base_options:
+                    is_debug = self.env.coredata.get_builtin_option('buildtype') == 'debug'
                     if self.env.coredata.base_options['b_vscrt'].value in ('mdd', 'mtd'):
                         is_debug = True
```
Neumann-A, line 1456: Still missing case `b_vscrt='none'` if I understand that correctly. Probably should fallback to `is_debug = self.env.coredata.get_builtin_option('debug')` in this case.

nirbheek: When `b_vscrt` is `none`, then we just don't pass any CRT flags to the compiler. That means the compiler will use whatever is the default, which will usually be `/MD` or `/MT`, both of which are release CRTs.

On platforms where `b_vscrt` is not supported, it won't be in `coredata.base_options` at all, so this case is only hit when building with MSVC and friends.

nirbheek: Basically, if `-Db_vscrt=none` and we try to follow the `debug` option with MSVC, the CMake dependency will use the debug CRT and Meson will use the release CRT and everything will blow up at link time or runtime.

nirbheek:
I tried adding a test for this, but the only test we have for the cmake dependency type is a linuxlike test that doesn't work on Windows.

mesonbuild/dependencies/base.py
Neumann-A: > That means the compiler will use whatever is the default

assuming `c(pp)_args` is empty.... never thought about `buildtype plain/custom` with completely custom build flags? 

> Meson will use the release CRT

as you stated: Meson will use nothing. It is in the responsibility of the user to supply the correct build flags within the environment in this case.

nirbheek: buildtype `custom` does not mean "completely custom build flags". It means the user has set debug/optimized to values that [do not match the mapping in the buildtype table](https://mesonbuild.com/Builtin-options.html#build-type-options). If you can set `--debug` you can and should also set `-Db_vscrt` instead of passing in the CRT args using `$lang_args`.

With buildtype `plain`, the user may indeed set any $lang_args, and we have no idea what CRT they're going to use. Maybe we should look at `$lang_args` in that case.

In neither case we should look at the value of `debug`.

Issue #7610, opened 2020-08-18
OpenSuse Github Action fails when building coverage


nirbheek:
Regression since https://github.com/mesonbuild/meson/pull/7598.

```
2020-08-15T14:51:35.7932388Z lcov: LCOV version 1.0
2020-08-15T14:51:35.7932500Z genhtml: LCOV version 1.0
2020-08-15T14:51:35.7932621Z Capturing coverage data from /__w/meson/meson/tmpsvqtwj2o
2020-08-15T14:51:35.7933349Z Can't locate JSON.pm in @INC (you may need to install the JSON module) (@INC contains: /usr/lib/perl5/site_perl/5.30.1/x86_64-linux-thread-multi /usr/lib/perl5/site_perl/5.30.1 /usr/lib/perl5/vendor_perl/5.30.1/x86_64-linux-thread-multi /usr/lib/perl5/vendor_perl/5.30.1 /usr/lib/perl5/5.30.1/x86_64-linux-thread-multi /usr/lib/perl5/5.30.1 /usr/lib/perl5/site_perl) at /usr/bin/geninfo line 63.
2020-08-15T14:51:35.7933705Z BEGIN failed--compilation aborted at /usr/bin/geninfo line 63.
2020-08-15T14:51:35.7933837Z Traceback (most recent call last):
```

nirbheek:
This has been fixed already. Not sure by what!

PR #7641, opened 2020-08-25
ci: Fix coverage tests on opensuse

nirbheek:
Fixes https://github.com/mesonbuild/meson/issues/7610


mensinda:
#7642 should get rid of most errors since they are caused by doxygen and hotdoc changes.

mensinda:
#7648 should fix the remaining arch issues

nirbheek:
This is not needed any longer.

PR #7642, opened 2020-08-25
test: Add 'dir' support for installed files in test.json

mensinda:
This is useful for automatically generated docs (doxygen, hotdoc) with a lot of generated files that may differ with different versions of the generator.


jpakkane:
This is a feature addition and those should not go to stable point releases.

mensinda:
It is, technically, a new feature, however, I would argue that it should still be included in the stable release:

- it only affects mesons `run_project_tests.py` setup (no production `mesonbuild` code was changed)
- it makes the tests more robust
  - the current test setup fails with the latest hotdoc version
  - the latest doxygen version also has incompatible changes (which is even worse, since this means that it is currently not possible to fix the test code for both Ubuntu and Arch)
- this PR (or something similar) is needed for #7641, which would then have to be pulled from 0.55.2 as well
  - if #7641 lands in master before the 0.55.2 release, the CI images will be updated. This would result in failing tests for the stable branch in both the CI build and on production systems with updated hotdoc and Doxygen versions (which is at least Arch)

nirbheek:
I agree with mensindia.

Issue #7654, opened 2020-08-27
Regression: bad order of includes for D (fix with MR)


second-reality:
**Describe the bug**
From commit 93c3ec7e2dd6d425baff5fd80e5a46c88d152cb0 (between 0.54.3 and 0.55.0), include directories in D are in reverse order on command line.

Problem was seen because we have the same module that is implemented by two different files, selectively included to have one or the other implementation. From meson >= 0.55.0, we can't compile our project since include orders is not respected.

**To Reproduce**
[DIncludeBadOrder.zip](https://github.com/mesonbuild/meson/files/5138951/DIncludeBadOrder.zip)
rm -rf bin && ~/meson/meson.py bin && ninja -C bin -v

In output, notice that -I../first and -I../second are inverted, resulting in compilation error.

**Expected behavior**
-I in correct order.

** Correction ideas **
This bug does not affect C/C++, and all "clike" compilers.
When introducing commit with regression, a work was done on clike.py to make compilers work with this (prepend, dedup, that kind of stuff). Seems like it is missing completely from d.py, and that result in regression.

**system parameters**
* Debian buster
* Python 3.7.3
* `meson --version`: 0.55.1
* ninja: 1.8.2

second-reality:
Proposed a MR: #7655

PR #7655, opened 2020-08-27
D: fix include order

second-reality:
Commit 93c3ec7e introduced a new way to handle deduplication with
compiler args. This resulted in D includes to be reversed.

See #7654

### DIFF
--- a/mesonbuild/compilers/d.py
+++ b/mesonbuild/compilers/d.py
@@ -19,6 +19,7 @@ from ..mesonlib import (
     EnvironmentException, MachineChoice, version_compare,
 )
 
+from ..arglist import CompilerArgs
 from .compilers import (
     d_dmd_buildtype_args,
     d_gdc_buildtype_args,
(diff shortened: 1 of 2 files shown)


mesonbuild/compilers/d.py
```diff
@@ -581,7 +585,7 @@ def _get_compiler_check_args(self, env, extra_args, dependencies, mode='compile'
         elif not isinstance(dependencies, list):
             dependencies = [dependencies]
         # Collect compiler arguments
-        args = self.compiler_args()
+        args = DCompilerArgs(self)
```
mensinda, line 588: Shouldn't `self.compiler_args()` still work, since you have overwritten it below?

second-reality: In current version, it does not.

From what I understood, compiler_args is overriden for "clike" languages, from which C/C++ inherits.
A mechanism was added in commit introducing regression where some options are defined as "deduplicated" or "prepended". I did the same kind of thing with D, by introducing DCompilerArgs (which inherits CompilerArgs).

Without this, unit test does not pass.
Maybe @dcbaker  who introduced all this has a comment (or better fix) for this.

mensinda: You did override `compiler_args` in `DCompiler`, so `self.compiler_args()` will use your version from `DCompiler`.

second-reality: Yes, just realised now I missed what you said.

You are totally right, I pushed a new version without this.

mensinda (approved):
LGTM

PR #7657, opened 2020-08-28
typing: Strict type annotations

mensinda:
This PR adds the `run_mypy.py` script for easy local type checking. It runs mypy in 2 modes:

- the "normal" mode (which is currently used) with all the files and modules we currently check
- the "strict" mode with better type checking which (currently) only runs on a subset of modules

There is also some minor refactoring done in the first commit to make typing interpreter base a lot less painful and the code easier to reason about.

---

## Update:

all modules are now checked with the "strict" mode, "normal" mode is now gone.

### DIFF
--- a/mesonbuild/arglist.py
+++ b/mesonbuild/arglist.py
@@ -164,7 +164,7 @@ class CompilerArgs(collections.abc.MutableSequence):
     def __getitem__(self, index: slice) -> T.MutableSequence[str]:  # noqa: F811
         pass
 
-    def __getitem__(self, index):  # noqa: F811
+    def __getitem__(self, index):  # type: ignore  # noqa: F811
         self.flush_pre_post()
         return self._container[index]
 
@@ -176,7 +176,7 @@ class CompilerArgs(collections.abc.MutableSequence):
     def __setitem__(self, index: slice, value: T.Iterable[str]) -> None:  # noqa: F811
         pass
 
-    def __setitem__(self, index, value) -> None:  # noqa: F811
+    def __setitem__(self, index, value) -> None:  # type: ignore  # noqa: F811
         self.flush_pre_post()
         self._container[index] = value
 
@@ -242,7 +242,7 @@ class CompilerArgs(collections.abc.MutableSequence):
             new = self.copy()
         else:
             new = self
-        return self.compiler.unix_args_to_native(new._container)
+        return T.cast(T.List[str], self.compiler.unix_args_to_native(new._container))
 
     def append_direct(self, arg: str) -> None:
         '''
--- a/mesonbuild/arglist.py
+++ b/mesonbuild/arglist.py
@@ -164,7 +164,7 @@ class CompilerArgs(collections.abc.MutableSequence):
     def __getitem__(self, index: slice) -> T.MutableSequence[str]:  # noqa: F811
         pass
 
-    def __getitem__(self, index):  # type: ignore  # noqa: F811
+    def __getitem__(self, index: T.Union[int, slice]) -> T.Union[str, T.MutableSequence[str]]:  # noqa: F811
         self.flush_pre_post()
         return self._container[index]
 
@@ -176,9 +176,9 @@ class CompilerArgs(collections.abc.MutableSequence):
     def __setitem__(self, index: slice, value: T.Iterable[str]) -> None:  # noqa: F811
         pass
 
-    def __setitem__(self, index, value) -> None:  # type: ignore  # noqa: F811
+    def __setitem__(self, index: T.Union[int, slice], value: T.Union[str, T.Iterable[str]]) -> None:  # noqa: F811
         self.flush_pre_post()
-        self._container[index] = value
+        self._container[index] = value  # type: ignore  # TODO: fix 'Invalid index type' and 'Incompatible types in assignment' erros
 
     def __delitem__(self, index: T.Union[int, slice]) -> None:
         self.flush_pre_post()
@@ -314,7 +314,7 @@ class CompilerArgs(collections.abc.MutableSequence):
         new += self
         return new
 
-    def __eq__(self, other: T.Any) -> T.Union[bool]:
+    def __eq__(self, other: object) -> T.Union[bool]:
         self.flush_pre_post()
         # Only allow equality checks against other CompilerArgs and lists instances
         if isinstance(other, CompilerArgs):
--- a/mesonbuild/arglist.py
+++ b/mesonbuild/arglist.py
@@ -242,7 +242,7 @@ class CompilerArgs(collections.abc.MutableSequence):
             new = self.copy()
         else:
             new = self
-        return T.cast(T.List[str], self.compiler.unix_args_to_native(new._container))
+        return self.compiler.unix_args_to_native(new._container)
 
     def append_direct(self, arg: str) -> None:
         '''
--- a/mesonbuild/ast/introspection.py
+++ b/mesonbuild/ast/introspection.py
@@ -296,7 +296,7 @@ class IntrospectionInterpreter(AstInterpreter):
         return None
 
     def is_subproject(self) -> bool:
-        return str(self.subproject) != ''
+        return self.subproject != ''
 
     def analyze(self) -> None:
         self.load_root_meson_file()
--- a/mesonbuild/ast/introspection.py
+++ b/mesonbuild/ast/introspection.py
@@ -25,10 +25,11 @@ from ..build import BuildTarget, Executable, Jar, SharedLibrary, SharedModule, S
 from ..mparser import BaseNode, ArithmeticNode, ArrayNode, ElementaryNode, IdNode, FunctionNode, StringNode
 import typing as T
 import os
+import argparse
 
 build_target_functions = ['executable', 'jar', 'library', 'shared_library', 'shared_module', 'static_library', 'both_libraries']
 
-class IntrospectionHelper:
+class IntrospectionHelper(argparse.Namespace):
     # mimic an argparse namespace
     def __init__(self, cross_file: str):
         self.cross_file = cross_file  # type: str
--- a/mesonbuild/ast/introspection.py
+++ b/mesonbuild/ast/introspection.py
@@ -21,7 +21,7 @@ from .. import compilers, environment, mesonlib, optinterpreter
 from .. import coredata as cdata
 from ..mesonlib import MachineChoice
 from ..interpreterbase import InvalidArguments, TYPE_nvar
-from ..build import Executable, Jar, SharedLibrary, SharedModule, StaticLibrary
+from ..build import BuildTarget, Executable, Jar, SharedLibrary, SharedModule, StaticLibrary
 from ..mparser import BaseNode, ArithmeticNode, ArrayNode, ElementaryNode, IdNode, FunctionNode, StringNode
 import typing as T
 import os
@@ -173,7 +173,7 @@ class IntrospectionInterpreter(AstInterpreter):
             'node': node
         }]
 
-    def build_target(self, node: BaseNode, args: T.List[TYPE_nvar], kwargs_raw: T.Dict[str, TYPE_nvar], targetclass) -> T.Optional[T.Dict[str, T.Any]]:
+    def build_target(self, node: BaseNode, args: T.List[TYPE_nvar], kwargs_raw: T.Dict[str, TYPE_nvar], targetclass: T.Type[BuildTarget]) -> T.Optional[T.Dict[str, T.Any]]:
         args = self.flatten_args(args)
         if not args or not isinstance(args[0], str):
             return None
@@ -295,7 +295,7 @@ class IntrospectionInterpreter(AstInterpreter):
         return None
 
     def is_subproject(self) -> bool:
-        return self.subproject != ''
+        return str(self.subproject) != ''
 
     def analyze(self) -> None:
         self.load_root_meson_file()
--- a/mesonbuild/ast/introspection.py
+++ b/mesonbuild/ast/introspection.py
@@ -32,9 +32,13 @@ build_target_functions = ['executable', 'jar', 'library', 'shared_library', 'sha
 class IntrospectionHelper(argparse.Namespace):
     # mimic an argparse namespace
     def __init__(self, cross_file: str):
+        super().__init__()
         self.cross_file = cross_file  # type: str
         self.native_file = None       # type: str
         self.cmd_line_options = {}    # type: T.Dict[str, str]
+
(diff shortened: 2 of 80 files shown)


lazka:
Just wondering, why don't set the mypy options in setup.cfg? Otherwise editors that integrate with mypy wont use the right options.

mensinda:
Because there are (at least currently) two sets of mypy options being used. The more strict ones are only used by a subset of files checked by mypy.

I know that this is a mess, but introducing typing to meson is still very much WIP and far from done or prefect.

lazka:
You can specify multiple sets of options for different packages/modules with wildcards: https://mypy.readthedocs.io/en/stable/config_file.html#config-file-format

mensinda:
Now, that I look at the remaining errors, there is not much work left to convert all the remaining files to the more strict rules...

mensinda:
OK, the parameters are now all in `.mypy.ini` and all modules now work with the strict mypy settings.

dcbaker:
I forgot to mention that I didn't get through interpreterbase.py yet.

mesonbuild/compilers/mixins/intel.py
```diff
@@ -97,7 +97,7 @@ def openmp_flags(self) -> T.List[str]:
         else:
             return ['-openmp']
 
-    def compiles(self, *args, **kwargs) -> T.Tuple[bool, bool]:
+    def compiles(self, *args: T.Any, **kwargs: T.Any) -> T.Tuple[bool, bool]:
```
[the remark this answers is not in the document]
mensinda: There are also other places with `*args, **kwargs`. What would be the correct approach in general?

mesonbuild/dependencies/boost.py
```diff
@@ -339,12 +339,12 @@ def get_link_args(self) -> T.List[str]:
...
 
 class BoostDependency(ExternalDependency):
-    def __init__(self, environment: Environment, kwargs):
+    def __init__(self, environment: Environment, kwargs: T.Dict[str, T.Any]) -> None:
         super().__init__('boost', environment, kwargs, language='cpp')
         self.debug = environment.coredata.get_builtin_option('buildtype').startswith('debug')
         self.multithreading = kwargs.get('threading', 'multi') == 'multi'
 
-        self.boost_root = None
+        self.boost_root = None  # type: Path
```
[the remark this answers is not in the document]
mensinda: You are right, it technically is more correct this way. I am going to change it.

However, there are already other places where this is used and we use `strict_optional = False`. If we want to be more strict in these cases, we should set this flag to `True` and fix all the errors.

mesonbuild/mcompile.py
```diff
@@ -32,7 +32,7 @@
     import argparse
 
 def array_arg(value: str) -> T.List[str]:
-    return UserArrayOption(None, value, allow_dups=True, user_input=True).value
+    return T.cast(T.List[str], UserArrayOption(None, value, allow_dups=True, user_input=True).value)
```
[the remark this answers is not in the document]
mensinda: The cast is required because of `follow_imports = skip` and thus `value` is of type `Any`, which is an error with the current mypy options. I will drop it if `follow_imports = silent` fixes this.

mesonbuild/mtest.py
```diff
@@ -1096,9 +1096,9 @@ def get_pretty_suite(self, test: 'TestSerialisation') -> str:
             s = "+".join(TestHarness.split_suite_string(s)[1] for s in test.suite)
             if s:
                 rv += ":"
-            return rv + s + " / " + test.name
+            return rv + s + " / " + T.cast(str, test.name)
         else:
-            return test.name
+            return T.cast(str, test.name)
```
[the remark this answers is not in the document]
mensinda: This was also because of `follow_imports = skip`

mesonbuild/arglist.py
```diff
@@ -176,7 +176,7 @@ def __setitem__(self, index: int, value: str) -> None:  # noqa: F811
     def __setitem__(self, index: slice, value: T.Iterable[str]) -> None:  # noqa: F811
         pass
 
-    def __setitem__(self, index, value) -> None:  # noqa: F811
+    def __setitem__(self, index, value) -> None:  # type: ignore  # noqa: F811
```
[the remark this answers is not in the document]
mensinda: I fixed this, but then I get the following erros in the assignment:

```
mesonbuild/arglist.py:181:25: error: Invalid index type "Union[int, slice]" for "List[str]"; expected type "int"
mesonbuild/arglist.py:181:34: error: Incompatible types in assignment (expression has type "Union[str, Iterable[str]]", target has type "str")
```

mensinda:
I have resolved most issues and switched to `follow_imports = silent`. However, this required adding even more type hints to other modules which increased the size of the PR....

mensinda:
Rebased on top of #7677.

scivision:
Wow that's very impressive that you got mypy to pass regardless of that `typing.Any` was used in a few places. There are some cases in my own projects that are 1/1000 complexity of mason that `typing.Any` is just the best choice.
Before this PR what I've had to do is wade through the hundreds of mypy messages trying to make sure that I didn't add a new type error. 
So if we find that it's appropriate to eliminate some of those Any's in the future we can do that incrementally. This is a big step towards making the internal quality of Mason better, and once we get a release out where Python 3.6 minimum isn't causing user mayhem, we can start to take advantage of 3.6 type hinting syntax.

mesonbuild/mcompile.py
```diff
@@ -169,9 +171,9 @@ def generate_target_name_vs(target: ParsedTargetName, builddir: Path, introspect
 
     # Normalize project name
     # Source: https://docs.microsoft.com/en-us/visualstudio/msbuild/how-to-build-specific-targets-in-solutions-by-using-msbuild-exe
-    target_name = re.sub('[\%\$\@\;\.\(\)\']', '_', intro_target['id'])
+    target_name = re.sub('[\%\$\@\;\.\(\)\']', '_', intro_target['id'])  # type: str
     rel_path = Path(intro_target['filename'][0]).relative_to(builddir.resolve()).parent
-    if rel_path != '.':
+    if rel_path != Path('.'):
```
scivision, line 176: Nice catch

scivision (approved):
I see there is a simple merge conflict but having scanned all this I think we should get this merged as many other PRs will conflict due to large PR size. This PR has already caught algorithmic errors revealed during the PR editing. 
My approach to this might have been to put typing.Any at key points just to get loose options with MyPy passing. However, this PR's approach is much better as it's jumped straight into checking.

mesonbuild/compilers/mixins/intel.py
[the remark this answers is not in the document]
dcbaker: My approach in the series I've got that adds typing for these is to make the signatures of the base `Compiler` and all subclasses match exactly, it's more verbose but it's also clearer.

mesonbuild/arglist.py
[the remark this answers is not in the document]
dcbaker: hmmm, I guess just leave as ignore then.

mesonbuild/ast/introspection.py
```diff
@@ -21,20 +21,25 @@
...
+from ..build import BuildTarget, Executable, Jar, SharedLibrary, SharedModule, StaticLibrary
 from ..mparser import BaseNode, ArithmeticNode, ArrayNode, ElementaryNode, IdNode, FunctionNode, StringNode
 import typing as T
 import os
+import argparse
 
 build_target_functions = ['executable', 'jar', 'library', 'shared_library', 'shared_module', 'static_library', 'both_libraries']
 
-class IntrospectionHelper:
+class IntrospectionHelper(argparse.Namespace):
```
dcbaker, line 32: GAh, What I'd given to have `Protocol`s

mesonbuild/backend/ninjabackend.py
```diff
@@ -2997,7 +2997,7 @@ def generate_ending(self):
         elem = NinjaBuildElement(self.all_outputs, deps, 'phony', '')
         self.add_build(elem)
 
-    def get_introspection_data(self, target_id, target):
+    def get_introspection_data(self, target_id: str, target: build.Target) -> T.List[T.Dict[str, T.Union[bool, str, T.List[T.Union[str, T.Dict[str, T.Union[str, T.List[str], bool]]]]]]]:
```
dcbaker, line 3000: This is a place that without `TypedDict`, I'd be happy to use `T.Any` as the dict value. This sort of stuff is exactly what `TypedDict` is for.

mesonbuild/build.py
```diff
@@ -369,22 +372,22 @@ def __init__(self, name, subdir, subproject, build_by_default, for_machine: Mach
         if not hasattr(self, 'typename'):
             raise RuntimeError('Target type is not set for target class "{}". This is a bug'.format(type(self).__name__))
 
-    def __lt__(self, other: T.Any) -> T.Union[bool, type(NotImplemented)]:
+    def __lt__(self, other: object) -> T.Union[bool, type(NotImplemented)]:
```
dcbaker, line 375: Does the `type(NotImplemented)` thing actually work? It looks like mypy is smart enough to understand that `-> bool`, so maybe just stick with that?

dcbaker: I know I started the `NotImplemented` in return type thing...

mesonbuild/dependencies/base.py
```diff
@@ -2541,7 +2541,7 @@ def factory_methods(methods: T.Set[DependencyMethods]) -> T.Callable[['FactoryTy
 
     This helps to make factory functions self documenting
     >>> @factory_methods([DependencyMethods.PKGCONFIG, DependencyMethods.CMAKE])
-    >>> def factory(env: Environment, for_machine: MachineChoice, kwargs: T.Dict[str, T.Any], methods: T.Set[DependencyMethods]) -> T.List[DependencyType]:
+    >>> def factory(env: Environment, for_machine: MachineChoice, kwargs: T.Dict[str, T.Any], methods: T.List[DependencyMethods]) -> T.List[DependencyType]:
```
dcbaker, line 2544: This needs a rebase I think

mesonbuild/envconfig.py
```diff
@@ -172,17 +172,17 @@ def __init__(self, system: str, cpu_family: str, cpu: str, endian: str):
         self.endian = endian
         self.is_64_bit = cpu_family in CPU_FAMILES_64_BIT  # type: bool
 
-    def __eq__(self, other: T.Any) -> 'T.Union[bool, NotImplemented]':
-        if self.__class__ is not other.__class__:
+    def __eq__(self, other: object) -> 'T.Union[bool, NotImplemented]':
+        if self.__class__ is not other.__class__ or not isinstance(other, MachineInfo):
```
dcbaker, line 176: This is being too clever to begin with, I think we should either just do:
```python
if isinstance(other, MachineInfo)
```
or
```python
if isinstance(other, type(self))
```

mesonbuild/envconfig.py
```diff
@@ -172,17 +172,17 @@ def __init__(self, system: str, cpu_family: str, cpu: str, endian: str):
...
         return \
             self.system == other.system and \
             self.cpu_family == other.cpu_family and \
             self.cpu == other.cpu and \
             self.endian == other.endian
 
-    def __ne__(self, other: T.Any) -> 'T.Union[bool, NotImplemented]':
-        if self.__class__ is not other.__class__:
+    def __ne__(self, other: object) -> 'T.Union[bool, NotImplemented]':
+        if self.__class__ is not other.__class__ or not isinstance(other, MachineInfo):
```
dcbaker, line 185: This one too.

mesonbuild/mesonlib.py
```diff
@@ -327,23 +327,23 @@ class OrderedEnum(Enum):
     """
     An Enum which additionally offers homogeneous ordered comparison.
     """
-    def __ge__(self, other):
-        if self.__class__ is other.__class__:
+    def __ge__(self, other: object) -> bool:
+        if self.__class__ is other.__class__ and isinstance(other, OrderedEnum)and isinstance(self.value, int) and isinstance(other.value, int):
```
dcbaker, line 331: See the (too clever) also, 3.6 adds `IntEnum` which can replace this I think.

mesonbuild/mesonlib.py
```diff
@@ -1115,7 +1115,7 @@ def unholder(item: T.List[_T]) -> T.List[_T]: ...
 @T.overload
 def unholder(item: T.List[T.Union[_T, 'ObjectHolder[_T]']]) -> T.List[_T]: ...
 
-def unholder(item):
+def unholder(item):  # type: ignore  # TODO fix overload (somehow)
```
dcbaker, line 1118: I think the right thing to do to fix a lot of these to is to insist that all flattening happen in the interpreter, that way all other functions can safely assume they're being passed flat data structures. That's another refactor for another time thought

dcbaker (changes requested):
I've looked through 53 of 79 files now, It's looking pretty good now, just a few things I've noticed.

mesonbuild/build.py
mensinda: I just did a `s/other: T.Any/other: object/g` on all files and didn't check the return type. I am not sure if `type(NotImplemented)` works, but a straight `-> bool` would be IMHO better anyway.

mesonbuild/ast/introspection.py
mensinda: I know, this is ugly but it works.

mesonbuild/backend/ninjabackend.py
mensinda: There is https://github.com/python/typing/tree/master/typing_extensions which could give us both `Protocol` and `TypedDict` for 3.6+

mesonbuild/dependencies/base.py
mensinda: Done

mesonbuild/mesonlib.py
mensinda: Then I would leave this as `type: ignore` for now.

jpakkane:
@dcbaker is this good for you know? If yes, just go ahead and merge this.

dcbaker (approved):
Looks good. Thanks for working on this and putting up with all my commentary :)

mesonbuild/backend/ninjabackend.py
dcbaker: I think that might be a good idea. If we always guard typing_extensions behind `if T.TYPE_CHECKING` then it would only be a requirement when using mypy.

scivision:
tremendous! thank you!!!11!!!!!1!

PR #7658, opened 2020-08-28
Reword unsupported compiler option warning

### LINKED ISSUE #7544: Confusing warning for unsupported compiler options
**Describe the bug**

If a compiler option is not supported (e.g. `c_std` on MSVC) Meson will say it's an unknown option, this can be confusing.

**To Reproduce**
```
project('example', 'c')
```
```bash
$ meson build; cd build
$ meson configure -Dc_std=c99
WARNING: Unknown options: "c_std"
The value of new options can be set with:
meson setup <builddir> --reconfigure -Dnew_option=new_value ...
```

**Expected behavior**
Indicate that the compiler does not support the option:
```bash
WARNING: Compiler does not support: "c_std"
```

**system parameters**
* Windows, MSVC
* Python 3.8.0
* meson 0.55.0

Apache-HB:
Fixes #7544


jpakkane:
This fails because the test suite assumes that error messages have a certain form:

```
self.assertIn('Unknown options: "bad, foo, wrong_link_args"', out)
E       AssertionError: 'Unknown options: "bad, foo, wrong_link_args"' not
```

You need to change the expected strings, too. You should also squash away the intermediate fixup commits so there is only one commit total.

dcbaker:
I'm sorry, but this is just wrong. that path is used by all options, not just compiler options.

jpakkane:
I'm going to close this as the issue mentioned by dcbaker must be fixed. In addition the branch has many spurious merges and reverts that should not be there so this would need to be completely reworked anyway. See https://mesonbuild.com/Contributing.html#keeping-pull-requests-up-to-date

Feel free to open a new mr once the issues are fixed. Thanks.

PR #7667, opened 2020-08-30
Be stricter when detecting Windows/Cygwin

### LINKED ISSUE #7552: run_command('git') fails when using MSYS2: 'utf-8' codec can't decode byte 0x90 in position 2: invalid start byte
**Describe the bug**
meson run_command('git') fails when using MSYS2 MINGW64 on Windows: 'utf-8' codec can't decode byte 0x90 in position 2: invalid start byte

```
sebas@ASUS-i7 MINGW64 /c/Users/sebas/workspace/siril
$ meson _build/
The Meson build system
Version: 0.54.3
Source dir: /c/Users/sebas/workspace/siril
Build dir: /c/Users/sebas/workspace/siril/_build
Build type: native build
Using 'PKG_CONFIG_PATH' from environment with value: '/mingw64/lib/pkgconfig:/mingw64/share/pkgconfig'
Using 'PKG_CONFIG_PATH' from environment with value: '/mingw64/lib/pkgconfig:/mingw64/share/pkgconfig'
Project name: siril
Project version: 0.99.3
C compiler for the host machine: cc (gcc 10.2.0 "cc.exe (Rev1, Built by MSYS2 project) 10.2.0")
C linker for the host machine: cc ld.bfd 2.34
C++ compiler for the host machine: c++ (gcc 10.2.0 "c++.exe (Rev1, Built by MSYS2 project) 10.2.0")
C++ linker for the host machine: c++ ld.bfd 2.34
Host machine cpu: x86_64

meson.build:53:0: ERROR: Program or command 'git' not found or not executable
```

I think it's a return of the bug that was fixed 2 years ago https://github.com/mesonbuild/meson/pull/3399 Fix encoding bug on Windows

And/or the same as reported here https://code.videolan.org/videolan/dav1d/-/issues/267 Build fails on Windows when using MSYS2

**To Reproduce**
Follow the steps from my bug report on Siril: https://gitlab.com/free-astro/siril/-/issues/519 Meson fails on Windows: ../meson.build:53:0: ERROR: Program or command 'git' not found or not executable

**system parameters**
* plain native build (for the same computer) but using MSYS2 MINGW64
* Windows 10
* Python 3.8.5
* meson 0.54.3
* (ninja 1.10.0)

Best regards,
Sébastien

lazka:
This removes the check for "mingw" for platform.system(). The only case I know
where "mingw" is return is if using a msys Python under a msys2 mingw environment.
This combination is not really supported by meson and will result in weird errors,
so remove the check.

The second change is checking sys.platform for cygwin instead of platform.system().
The former is document to return "cygwin", while the latter is not and just
returns uname().

While under Cygwin it uname() always starts with "cygwin" it's not hardcoded in MSYS2
and starts with the environment name. Using sys.platform is safer here.

Fixes #7552

### DIFF
--- a/mesonbuild/envconfig.py
+++ b/mesonbuild/envconfig.py
@@ -211,13 +211,13 @@ class MachineInfo:
         """
         Machine is windows?
         """
-        return self.system == 'windows' or 'mingw' in self.system
+        return self.system == 'windows'
 
     def is_cygwin(self) -> bool:
         """
         Machine is cygwin?
         """
-        return self.system.startswith('cygwin')
+        return self.system == 'cygwin'
 
     def is_linux(self) -> bool:
         """
--- a/mesonbuild/environment.py
+++ b/mesonbuild/environment.py
@@ -429,10 +429,9 @@ def detect_cpu(compilers: CompilersDict):
     return trial
 
 def detect_system():
-    system = platform.system().lower()
-    if system.startswith('cygwin'):
+    if sys.platform == 'cygwin':
         return 'cygwin'
-    return system
+    return platform.system().lower()
 
 def detect_msys2_arch():
     if 'MSYSTEM_CARCH' in os.environ:
--- a/mesonbuild/mesonlib.py
+++ b/mesonbuild/mesonlib.py
@@ -502,11 +502,11 @@ def is_openbsd() -> bool:
 
 def is_windows() -> bool:
     platname = platform.system().lower()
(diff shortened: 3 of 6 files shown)


nirbheek:
@lazka should this go into the 0.55.2 stable release or have you already patched it in MSYS and it's not needed?

lazka:
No need to backport, 0.56 is fine.

Issue #7666, opened 2020-08-30
[cmake module] Directories of generated headers are ignored instead of created


w0xel:
**Describe the bug**
When a header is generated in the CMake project, but the path does not exist yet, meson outputs a warning and fails during compilation instead of creating the directory.

**To Reproduce**
See example repository here: https://github.com/w0xel/meson_cmake_module_bug_example/

**Expected behavior**
Meson should create the directory and add it to the include_directories of the target if the directory is within the CMake build directory. 

**system parameters**
* native build
* Ubuntu 20.04
* Python 3.8.2
* meson 0.55.1 (meson git doesn't work either as of 6fc06734)
* ninja 1.10.0
* CMake 3.18.1

I'm currently trying to get https://github.com/krpc/krpc to work for an education project btw., this issue is the last missing piece.

PR #7670, opened 2020-08-30
CMake module: Allow paths of generated CMake sources for include directories

w0xel:
Fixes: https://github.com/mesonbuild/meson/issues/7666

Meson will then just assume directories of generated files will appear during build.


mensinda:
This needs a test case. You can probably add it to `test cases/cmake/8 custom command`.

w0xel:
Your call to create a test case was right, as it always is, as explicitly creating the directory is indeed required, if the cmake-project does not do this. I also used the `${CMAKE_COMMAND} -E copy` instead of `mycpy`, as it will create the fully path required for the copy implicitly (the new meson code does only create `some`, but `some/directory` is required in the CMake project).

This test case reflects what happens in krpc.

w0xel:
Also deleted an unnecessary newline that slipped in...

mensinda (approved):
LGTM, assuming the CI is green

PR #7671, opened 2020-08-30
Bump minimum supported Python version to 3.6. Closes #6297.

### LINKED ISSUE #6297: Minimum supported Python 3 version bump
At some point we should update the minimum version of Python 3. Here's what the distros currently have:

 - Debian stable, 3.7
 - Ubuntu Xenial, 3.5
 - Ubuntu Bionic, 3.6
 - RHEL 8, 3.6???
 - SLES 15, I could not find out by googling

Xenial is _probably_ old enough that we don't have to keep supporting it. 3.6 seems to be widely available, but is that so. Are there platforms where we'd need to support 3.5?

jpakkane:


mensinda:
Bionic (18.04) does have python 3.6, so we don't need to remove it, or am I missing something?

https://packages.ubuntu.com/search?keywords=python3&searchon=names&suite=bionic&section=all

jpakkane:
That was unintentional. Fixed.

mensinda:
Is keeping Ubuntu 16.04 around intentional?

ignatenkobrain:
CentOS has 3.6, so fine fro RHEL/Fedora POV.

+1

nirbheek:
I would like to add a warning for people who are using Python 3.5 for one release before we do this. If for no other reason than to give people a heads-up.

xclaesse:
+1 for having a warning cycle.

jpakkane:
This goes in immediately after 0.56.0 is done.

jpakkane:
Apparently Azure conf also needs updating:

```
Meson works correctly only with python 3.6+.
You have python 3.5.4 (v3.5.4:3f56838, Aug  8 2017, 02:17:05) [MSC v.1900 64 bit (AMD64)].
Please update your environment
```

scivision (approved):
I think this is a good deprecation cycle / approach

alanc:
Shouldn't [docs/markdown/Getting-meson.md](https://github.com/mesonbuild/meson/blob/master/docs/markdown/Getting-meson.md) also be updated to list 3.6 instead of 3.5??

jpakkane:
It should. #8155.

PR #7672, opened 2020-08-31
mcompile: Also support -v for verbose

nirbheek:
This matches `meson test`, and there's really no other meaning that could be attributed to this, since you would call `meson` to get the version, not the `compile` sub-command.


mensinda (approved):

nirbheek:
CI failure was due to a timeout downloading CI data. Unrelated and can be ignored here.

PR #7673, opened 2020-08-31
Add a notice about Python 3.5 support

nirbheek:
This will be printed in bold at the end of interactive meson sub-commands that won't be parsed by a program. Specifically: setup, compile, test, and install.

`NOTICE: You are using Python 3.5 which is EOL. Starting with v0.57, Meson will require Python 3.6 or newer`

### DIFF
--- a/mesonbuild/mesonmain.py
+++ b/mesonbuild/mesonmain.py
@@ -38,7 +38,7 @@ class CommandLineParser:
         self.commands = {}
         self.hidden_commands = []
         self.parser = argparse.ArgumentParser(prog='meson', formatter_class=self.formatter)
-        self.subparsers = self.parser.add_subparsers(title='Commands',
+        self.subparsers = self.parser.add_subparsers(title='Commands', dest='command',
                                                      description='If no command is specified it defaults to setup command.')
         self.add_command('setup', msetup.add_arguments, msetup.run,
                          help_msg='Configure the project')
@@ -111,6 +111,7 @@ class CommandLineParser:
         return 0
 
     def run(self, args):
+        print_py35_notice = False
         # If first arg is not a known command, assume user wants to run the setup
         # command.
         known_commands = list(self.commands.keys()) + ['-h', '--help']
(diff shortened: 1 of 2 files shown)


xclaesse:
This should come with a big fat warning in release notes too.

mensinda:
We could also backport this up to older meson versions (maybe up to 0.50) with a new point release just for this, to ensure that more people will see this message.

nirbheek:
> We could also backport this up to older meson versions (maybe up to 0.50) with a new point release just for this

I doubt any distros are going to pull that into their stable distros. I think pulling it into 0.55.2 is the best we can do.

mesonbuild/mesonmain.py
```diff
@@ -119,14 +120,22 @@ def run(self, args):
...
             parser = self.parser
+            command = None
 
         args = mesonlib.expand_arguments(args)
         options = parser.parse_args(args)
 
+        if command is None:
+            command = options.command
+
+        if command in ('setup', 'compile', 'test', 'install') and sys.version_info < (3, 6):
```
mensinda, line 136: We can probably also add `'configure'` here.

nirbheek: I thought about that, but then I thought that there might be some IDE plugin that's parsing `meson configure` output. The rest have more-or-less free-form output.

mensinda: the `meson configure` output is **not** supposed to be parsed (I already [told the eclipse devs](https://bugs.eclipse.org/bugs/show_bug.cgi?id=548350)). We have an introspection API for that.

nirbheek: Yes, but the people using Python 3.5 are probably unable to update anything other than Meson itself. Arguably an unsupported configuration, but why take the chance. Anyone who runs `meson configure` is also going to run at least `setup` 🙂

mensinda: Just saying that we already had changes in the `meson configure` format and the only project that bothered constantly updating their fragile regex setup is/was eclipse...

nirbheek: That's proof to me that we should not put the notice at the end of `configure`! 😄

nirbheek:
This was ACKed by Jussi on IRC a few days ago, so merging.

xclaesse:
You should add a release note.

nirbheek:
> You should add a release note.

Yes, was going to, and did: https://github.com/mesonbuild/meson/pull/7681

PR #7674, opened 2020-08-31
Add support for the CompCert C Compiler

archi:
Hello Meson developers! I added some basic/preliminary support for the [CompCert C Compiler](https://github.com/AbsInt/CompCert).
The code is essentially a copy of the XC16 code, plus some adaptions.

From my perspective the most questionable behaviour is the automatic filtering of some GCC flags:
CompCert uses GCC as preprocessor, assembler and linker. Hence many GCC flags are supported, but need to be prefixed with a `-WUl,` [linker], `-Wp,` [preprocessor] or `-Wa,` [assembler] for CompCert's driver to pass them to the correct tool.
Since my current objective is to evaluate [picolibc](https://github.com/picolibc/picolibc), I automatically prefix those flags which are used by picolibc's build files.

Additionally, some GCC flags are safe to be ignored by CompCert (such as the `-fno-builtin`), so these are also filtered (again because picolibc has them hardcoded in its build files).

I suppose this isn't perfect on first try (plus I don't usually develop in python), so just let me know what to adapt :)

Best regards, Sebastian

### DIFF
--- a/cross/ccomp-armv7a.txt
+++ b/cross/ccomp-armv7a.txt
@@ -1,0 +1,13 @@
+[binaries]
+c = 'ccomp'
+ar = 'ccomp'
+strip = 'strip'
+
+[built-in options]
+c_args      = ['-target armv7a-eabi', '-fall']
+
+[host_machine]
+system = 'bare metal'      # Update with your system name - bare metal/OS.
+cpu_family = 'arm'
+cpu = 'Cortex-A9'
+endian = 'little'
--- a/cross/ccomp-armv7a.txt
+++ b/cross/ccomp-armv7a.txt
@@ -4,7 +4,7 @@ ar = 'ccomp'
 strip = 'strip'
 
 [built-in options]
-c_args      = ['-target armv7a-eabi', '-fall']
+c_args      = ['-target', 'armv7a-eabi', '-fall']
 
 [host_machine]
 system = 'bare metal'      # Update with your system name - bare metal/OS.
--- a/docs/markdown/snippets/add_compcert_compiler.md
+++ b/docs/markdown/snippets/add_compcert_compiler.md
@@ -1,3 +1,3 @@
 ## Added CompCert C compiler
 
-Added support for the CompCert formally-verified C compiler (https://github.com/AbsInt/CompCert).
+Added experimental support for the [CompCert formally-verified C compiler](https://github.com/AbsInt/CompCert). The current state of the implementation is good enough to build the [picolibc project](https://github.com/picolibc/picolibc) with CompCert, but might still need additional adjustments for other projects.
--- a/docs/markdown/snippets/add_compcert_compiler.md
+++ b/docs/markdown/snippets/add_compcert_compiler.md
@@ -1,0 +1,3 @@
+## Added CompCert C compiler
+
+Added support for the CompCert formally-verified C compiler (https://github.com/AbsInt/CompCert).
--- a/mesonbuild/compilers/c.py
+++ b/mesonbuild/compilers/c.py
@@ -576,7 +576,7 @@ class CompCertCCompiler(CompCertCompiler, CCompiler):
         return ['-O0']
 
     def get_output_args(self, target):
-        return ['-o%s' % target]
+        return ['-o{}'.format(target)]
 
     def get_werror_args(self):
         return ['-Werror']
--- a/mesonbuild/compilers/c.py
+++ b/mesonbuild/compilers/c.py
@@ -22,6 +22,7 @@ from .c_function_attributes import C_FUNC_ATTRIBUTES
 from .mixins.clike import CLikeCompiler
 from .mixins.ccrx import CcrxCompiler
 from .mixins.xc16 import Xc16Compiler
+from .mixins.compcert import CompCertCompiler
 from .mixins.c2000 import C2000Compiler
 from .mixins.arm import ArmCompiler, ArmclangCompiler
 from .mixins.visualstudio import MSVCCompiler, ClangClCompiler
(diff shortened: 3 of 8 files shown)


archi:
Ah, the unused import and the `self` vs `cls` are rather obvious. For the conflicting attributes I have to take a look at the violated rules on how to fix them (that part was just copy&pasted with minor adaptions).

Not sure what caused the travis/mac build to silently fail, seems like instance was just sitting there, waiting for homebrew?

dcbaker:
Ignore the conflicting base class thing. That one's my fault

archi:
Thanks @dcbaker. Is there anything I have to do here, or should I just wait @jpakkane to review the patch?

jpakkane:
> Additionally, some GCC flags are safe to be ignored by CompCert (such as the -fno-builtin), so these are also filtered (again because picolibc has them hardcoded in its build files).

This seems a bit suspicious. Typically people really like their compiler arguments and get very upset if they get removed (even if unnecessary). This seems like a thing that should be fixed in picolibc build files, because if they add these flags unconditionally then they will break with all compilers that don't support them.

archi:
This was intended to offer a more easy "drop in replacement". But I can understand this isn't always the best choice and remove it :)

This part is okay, though?

```
# As of CompCert 20.04, these arguments should be passed to the underlying gcc linker (via -WUl,<arg>)
# There are probably (many) more, but these are those used by picolibc
```

mesonbuild/compilers/c.py
```diff
@@ -554,6 +555,36 @@ def get_include_args(self, path, is_system):
...
+        return opts
+
+    def get_option_compile_args(self, options):
+        return []
+
+    def get_no_optimization_args(self):
+        return ['-O0']
+
+    def get_output_args(self, target):
+        return ['-o%s' % target]
```
dcbaker, line 579:
```suggestion
        return ['-o{}'.format(target)]
```

I think we're trying to migrate away from using the % formatters.

mesonbuild/compilers/mixins/compcert.py
```diff
@@ -0,0 +1,127 @@
...
+
+# As of CompCert 20.04, these arguments should be passed to the underlying gcc linker (via -WUl,<arg>)
+# There are probably (many) more, but these are those used by picolibc
+ccomp_args_to_wul = [
+        r"^-ffreestanding$",
+        r"^-r$"
+] # type: T.List[str]
+
+class CompCertCompiler:
+    def __init__(self):
```
dcbaker, line 55:
```suggestion
    def __init__(self) -> None:
```

mesonbuild/linkers.py
```diff
@@ -230,6 +230,18 @@ def get_output_args(self, target: str) -> T.List[str]:
...
+
+    def __init__(self, exelist: T.List[str]):
+        super().__init__(exelist)
+        self.id = 'ccomp'
+
+    def can_linker_accept_rsp(self) -> bool:
+        return False
+
+    def get_output_args(self, target: str) -> T.List[str]:
+        return ['-o {}'.format(target)]
```
dcbaker, line 243: This seems off to me, this would be equivalent to in bash doing:
```sh
$CC "-o ${target}"
```
I'm assuming you either want to pas this as a two element list, or to remove the space between them?

dcbaker (changes requested):
Sorry it took me so long to get around to this. This looks pretty good overall, there's just a couple of small things I see.

Additionally we need a release snippet doc, and the ids need to be added to the documentation as well, there's a table of them.

dcbaker:
Oh, and ignore that last warning from LGTM. It's fine.

archi:
> Sorry it took me so long to get around to this. 
No problem :)

> This looks pretty good overall, there's just a couple of small things I see.
Thanks. I have committed all of your suggestions.

> Additionally we need a release snippet doc, and the ids need to be added to the documentation as well, there's a table of them.

Okay, makes sense. I've added a release note snippet, and `ccomp` to the compiler+linker tables. Let me know if you'd like anything beyond that.

mesonbuild/compilers/mixins/compcert.py
```diff
@@ -0,0 +1,127 @@
...
+import re
+import typing as T
+
+if T.TYPE_CHECKING:
+    from ...environment import Environment
+
+ccomp_buildtype_args = {
+    'plain': [''],
+    'debug': ['-O0 -g'],
+    'debugoptimized': ['-O0 -g'],
```
dcbaker, line 27:
```suggestion
    'debug': ['-O0', '-g'],
    'debugoptimized': ['-O0', '-g'],
```

dcbaker (changes requested):
Thanks for fixing all of that. There's just one last thing I spotted and this will be ready to land

docs/markdown/snippets/add_compcert_compiler.md
```diff
@@ -0,0 +1,3 @@
+## Added CompCert C compiler
+
+Added support for the CompCert formally-verified C compiler (https://github.com/AbsInt/CompCert).
```
jpakkane, line 3: This should use the markdown format instead:

    Added support for the [CompCert formally-verified C compiler](https://github.com/AbsInt/CompCert).

Also, it would be useful if there were a couple of sentences describing the state of the implementation, possible gotchas (if any) and so on.

archi: State/gotchas of the compiler support implementation inside meson, or state/gotchas of the actual compiler? The `add_nvidia_hpc_sdk_compilers.md` isn't very verbose, either, so I sticked to that non-verbosity ;) But could add a blurb if that's preferred.

jpakkane: It's more of a general "if you use this, expect roughly this sort of quality". So whether the implementation is "thoroughly tested with a large set of programs", "fairly good, should be expected to work on most projects" or "somewhat experimental, works with some projects but may have issues".

dcbaker (approved):
This looks good to me. When @jpakkane is happy with the documentation this is ready to land I think.

docs/markdown/snippets/add_compcert_compiler.md
archi: Thanks for the clarification. The snippet has been updated.

cross/ccomp-armv7a.txt
```diff
@@ -0,0 +1,13 @@
+[binaries]
+c = 'ccomp'
+ar = 'ccomp'
+strip = 'strip'
+
+[built-in options]
+c_args      = ['-target armv7a-eabi', '-fall']
```
archi, line 7: Oh, I assume this would be better written as `['-target', 'armv7a-eabi', '-fall']`? @dcbaker?

jpakkane: Yes. All command lines should be defined as fully separated arrays, not strings.

archi: Okay, adapted

jpakkane:
Thanks.

PR #7677, opened 2020-09-01
Submit/fix dependency factory type

dcbaker:
I've done my best to fix the dependency_factory annotations. I'm not entirely happy with the result, but it does shut mypy up and it is what the mypy documentation says to do.


mensinda (approved):
LGTM, the typing is even less intrusive than I feared.

PR #7681, opened 2020-09-02
docs: Add a snippet for python 3.5 deprecation [skip ci]

nirbheek:


mensinda (approved):

docs/markdown/snippets/python35-eol.md
```diff
@@ -0,0 +1,15 @@
+## Python 3.5 support will be dropped in the next release
+
+The final [Python 3.5 release was 3.5.10 in September](https://www.python.org/dev/peps/pep-0478/#id4).
+This release series is now End-of-Life (EOL). The only LTS distribution that
+still only ships Python 3.5 is Ubuntu 16.04, which will be
+[EOL in April 2021](https://ubuntu.com/about/release-cycle).
```
dhewg, line 6: This is unfortunately not true:
Debian 9 "Stretch" (oldstable) is supported up to June 30, 2022 [0] and ships with python 3.5 [1].

This is an actual issue for OpenWrt and its buildbots, which are running debian oldstable. My patch to drop support for python 3.5 because of meson was rejected for that reason [2].

Meson is popular and widely used by now, please don't be so eager to remove support for older distros. Especially CI boxes aren't always updated to the latest and greatest.

[0] https://wiki.debian.org/LTS
[1] https://packages.qa.debian.org/p/python3-defaults.html
[2] http://patchwork.ozlabs.org/project/openwrt/patch/20210216072108.26107-1-a.heider@gmail.com/

eli-schwartz: Perhaps, but even so it remains true that python 3.5 itself is EOL.

Are you just making a point that the wording here should be tweaked? Do you want meson to reintroduce support for end of life versions of python?

(Why do you need to upgrade meson on Debian oldstable anyway? The entire distro base is incredibly outdated and frozen in amber.)

dhewg:
No, I don't want meson to re-add python 3.5 support ;)
I'm just stating that it was a bit too eager to remove IMHO, hopefully this can be avoided in the future.

While debian oldstable is indeed outdated, in this case it just serves as a build host for the OpenWrt distro. Meson is part of its package collection to build other packages. I'll try to keep meson at < 0.57 so it keeps working, hopefully no packages require it anytime soon.

docs/markdown/snippets/python35-eol.md
xclaesse: TBH, I agree that python EOL does not matter, we are not a web service.  What matters is keeping a viable code base that uses new features from the language instead of piling hacks on top of hacks. So there is a balance to be found, and I think 3.6 is not unreasonable at this time. At some point it gets OpenWrt's responsability to upgrade their buildbots if they want to remain a viable project.

eli-schwartz: It does matter only in the sense that once it's EOL it's fair game for people to say "we really need this new feature, let's bump the minimum version".

jpakkane:
> I'm just stating that it was a bit too eager to remove IMHO, hopefully this can be avoided in the future.

When should it be removed then, in your opinion? Only until oldstable goes out of support? Even further if people keep on using it? Something else?

dhewg:
Well it's project specific, but in my humble opinion the commit message got it right - check the widely used LTS distros ;) It's just that you missed debian by accident (but I too found another debian wiki page first which had another already expired date).

I've seen similar approaches with mesa bumping the minimum required meson version or projects dropping support for older gcc version etc. That sounds sensible to me. But anyway, hopefully we can avoid issues while sticking to 0.56 until the OpenWrt buildbots get updated (no idea when that will be).

jpakkane:
> It's just that you missed debian by accident

To be honest we did not miss it by accident. It's just that oldstable was not even considered, only stable. That is so old that it is unlikely that compiling modern software (that requires >= 0.57) your other dependencies are going to be so out of data that you'd need to update them too.

dhewg:
Oh, alright. Oldstable is the single debian LTS at this moment though, see the link from my first post.

Maybe OpenWrt is special in this regard, but there're just a few essential host requirements [0]. It basically just needs to be able to build a cross compiler and then goes from there. Most of the host packages don't matter, like >95% of its work is cross compiling target binaries. But by now quite a few packages require meson so python3 is one of the few that do matter.

[0] https://github.com/openwrt/openwrt/blob/master/include/prereq-build.mk

PR #7685, opened 2020-09-02
tests: add 'd/12 dub static library' test case

ljmf00:
This will cover the behavior of adding a dependency with 'staticLibrary'
target type from dub. The test succeed if it links and find the main function
symbol, otherwise, should fail, reporting missing that.

Signed-off-by: Luís Ferreira <contact@lsferreira.net>


ljmf00:
Related to this merged tag: https://github.com/mesonbuild/dubtestproject/pull/3 .

jpakkane:
The new test fails:

```
test cases/d/12 dub static library/meson.build:8:0: ERROR: Dependency "dubtestproject:test3" not found
```

If this was a question of upstream things being out of date, rebasing to current master and force pushing should fix it.

ljmf00:
> The new test fails:
> 
> ```
> test cases/d/12 dub static library/meson.build:8:0: ERROR: Dependency "dubtestproject:test3" not found
> ```
> 
> If this was a question of upstream things being out of date, rebasing to current master and force pushing should fix it.

I already rebased with `master` and force pushed, but `CI image builder / Fedora (pull_request)` job seems failing for unrelated reasons.

PR #7686, opened 2020-09-02
Revert "Do not automatically set warning flags if buildtype is 'plain'"

### LINKED ISSUE #7399: Meson silently ignores compiler's warn_args for buildtype=plain
**Describe the bug**

If the `buildtype` option is set to `plain`, meson will ignore the compiler's `warn_args` (as stated in [this line](https://github.com/mesonbuild/meson/blob/d42cd735a4dc894d8e898a5f9e81029f6eb5364c/mesonbuild/backend/backends.py#L641). This way, `warning_level=<any value>` is completely thrown out (although `werror=true` still sends a `-Werror` in GCC/Clang).

In addition to that, there are two things I would like to point out:
  1. **Documentation:** There's nothing obvious describing what is each buildtype's expected scenarios. I only found out the idea behind `plain` because of a line in ["Quick-Guide"](https://mesonbuild.com/Quick-guide.html#using-meson-as-a-distro-packager), while the table in [Core Options](https://mesonbuild.com/Builtin-options.html#core-options) doesn't give any clue about it. I would appreciate some more and better documenting on this (I could help with a PR but probably would end up causing some rework, since someone would have to explain to me the internal differences between the builttypes, which is almost the same as writing the doc itself).

  2. **The current behaviour:** While discussing in `#mesonbuild`, it was mentioned that maybe it wasn't a good choice of behaviour. First because we now sort of have two ways to enable/disable the same thing (i.e. either by defining `warning_level=0` or by having `buildtype=plain`), even if one of them is some sort of a superset of the other. For this, it was stated that if someone wanted to both disable meson flags _and_ warnings, we could simply recommend to use `--buildtype=plain --warnlevel 0`. The second is that it is completely non-overrideable: if you use `buildtype=plain`, there's no way to reapply `warning_level`'s flags. For this one, maybe trying to change this could lead to unnecessary additional complexity, but further discussion could also lead to a simple solution for this.

**To Reproduce**
Have any simple meson C or C++ project, configure builddir with `buildtype=plain` and `-Dwarning_level=<number greater than 0>`. Then use `ninja -C <builddir> --verbose` to see that no `-Wall`, `-Wextra` or `-Wpedantic` flags go into compilation lines.

**Expected behavior**

There are three kind of behaviors I would expect:
- Meson still uses `warning_level` in `plain` builds;
- Meson warns me about they being ignored because it is a `plain` build;
- Meson doesn't complain of manually inserting `-Wall` on a `plain` build (but keep warning on any other buildtype).

Additionally, `werror` (as well as other flags) should be consistent with the decided behavior.

**system parameters**
* **Is this a [cross build](https://mesonbuild.com/Cross-compilation.html) or just a plain native build (for the same computer)?**
  Plain (for the same computer).
* **what operating system (e.g. MacOS Catalina, Windows 10, CentOS 8.0, Ubuntu 18.04, etc.)**
  Linux 5.7.5-arch1-1

* **what Python version are you using e.g. 3.8.0**
  3.8.3

* **what `meson --version`**
  0.54.3

* **what `ninja --version` if it's a Ninja build**
  1.10.0

eli-schwartz:
This reverts commit 8ee1c9a07a3a35e3ed262fbc358fd86c257a966e.

No rationale was given for this change prior to merging. After the fact it was described as desired by distro packagers, however as a distro packager I believe this commit hurts me.

From a distro packaging perspective, we primarily care about one thing: previously building code should still build. Hence, -Werror is bad for our use case. meson handles this via -D werror which is different from -D warning_level and as long as the former is disabled, warnings in the build do not cause harm (but may attract review attention for upstream to fix).

buildtype is a completely unrelated concern, and the intention for =plain is to disable debug or optimization settings that result in codegen differences and thus different built artifacts. This must not happen in distro builds because the *distributed programs* should conform to policy settings.

Unfortunately, completely disabling warnings happens silently, and cannot be overridden even if you really, really believe you know what you're doing. It is thematically broken, since use of add_project_arguments() to add more -W flags is not likewise ignored. But if you try to add -Wall in that manner, meson will lecture you to use warning_level which you cannot do. And if you have custom warning flags which depend on options enabled by judicious use of -Wall via default_options: 'warning_level=1', then you end up with generated warnings complaining about your command line rather than your code, such as:

```
cc1: warning: ‘-Wformat-y2k’ ignored without ‘-Wformat’ [-Wformat-y2k]
cc1: warning: ‘-Wformat-extra-args’ ignored without ‘-Wformat’ [-Wformat-extra-args]
cc1: warning: ‘-Wformat-zero-length’ ignored without ‘-Wformat’ [-Wformat-zero-length]
cc1: warning: ‘-Wformat-contains-nul’ ignored without ‘-Wformat’ [-Wformat-contains-nul]
cc1: warning: ‘-Wformat-security’ ignored without ‘-Wformat’ [-Wformat-security]
```

which then break the build with -Werror.

Throughout all this, a buildtype of "plain" does *not* disable -D werror=true, which is the part where distro builds actually break down!

Users who both wish to disable debug/optimization codegen, *and* disable warning commentary, are encouraged to do so by doing both, not by doing one and having the other be assumed.

Fixes #7399

### DIFF
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -663,7 +663,7 @@ class Backend:
         # we weren't explicitly asked to not emit warnings (for Vala, f.ex)
         if no_warn_args:
             commands += compiler.get_no_warn_args()
-        elif self.get_option_for_target('buildtype', target) != 'plain':
+        else:
             commands += compiler.get_warn_args(self.get_option_for_target('warning_level', target))
         # Add -Werror if werror=true is set in the build options set on the
         # command-line or default_options inside project(). This only sets the


mensinda:
I agree with reverting this change (is this behavior even documented?), however, this should also have a test case.

eli-schwartz:
A test case for what?

jpakkane:
The reason this is like this is that at least Debian and Fedora to roughly this:

```
CFLAS='-g -O2 -Wall -D__FORTIFY=something etc` meson <options>
```

The point of plain was (at least originally) to mean that Meson does not add any optional flags like those (and not just debug and optimization levels) and that end users are expected to add all of them by hand. Yes, this is a bit stupid, but that is how Autotools, CMake et al work and people's work flows have been tuned to set things up this way.

No, I don't like this either, but that is the state of things as of now. Maybe we could change it so that it won't print the warnings for -Wall if buildtype is plain. Or something else.

eli-schwartz:
Since meson and Debian/Fedora are agreed on passing -Wall, there should be no functional changes by default. But I still think if they want that, then they should pass `-D warning_level=0` too. It's not correct to conflate `buildtype=plain` with `option_profile=linuxdistros`.

As far as not warning on buildtype=plain and -Wall, this doesn't solve the problem where it's literally impossible to do:

```
meson setup builddir/ -D debug=false -D optimization=0 -D warning_level=1
```

because the explicitly passed warning level is ignored. If you want to tie buildtype to the warning level, then *first* merge this PR, *then* teach warning_level to default to 0 if buildtype is plain but otherwise default to 1.

In the meantime projects are actually failing to build because meson provides NO escape hatch here.

There is no breakage caused by merging this PR. Debian/Fedora continue to build packages as they used to, and can add `-D warning_level=0` at their leisure to prevent -Wall being passed twice.

eli-schwartz:
The only correct way to handle "Debian wishes to disable all warnings, then enable -Wall via CFLAGS" is to add `add_project_warningflags()` which acts like `add_project_arguments()` but does nothing if warning_level=0.

That's beyond the scope of this bugfix PR though.

bonzini:
Additionally, the `-Wall` that some distros include in CFLAGS is an "additive" flag. Distros do that to enable _additional_ warnings, in case the configure script or CMakeLists do not do so. There is no reason why distros would want Meson to drop a project's own carefully crafted set of compiler warning options, as long as it includes `-Wall`.

For Meson projects they could in principle do so using `-Dwarning_level=1`, but they actually cannot due to this very same issue. So I agree that the commit should be reverted.

jpakkane:
@ignatenkobrain how does this change look like from Fedora's point of view?

nirbheek:
fwiw, I agree with @eli-schwartz here as well.

ignatenkobrain:
Probably fine with me.

PR #7690, opened 2020-09-03
ci: update to use latest recommendations

scivision:
The "unused_args" CI .yml was updated in #7689

This just uses the latest Actions recommended by Github and removes redundant variables "CI" that are already global by default now.

### DIFF
--- a/.github/workflows/os_comp.yml
+++ b/.github/workflows/os_comp.yml
@@ -25,7 +25,6 @@ jobs:
     - name: Run tests
       run: LD_LIBRARY_PATH=/usr/local/share/boost/1.69.0/lib/:$(rustc --print sysroot)/lib:$LD_LIBRARY_PATH python3 run_tests.py
       env:
-        CI: '1'
         XENIAL: '1'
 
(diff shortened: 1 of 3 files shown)


.github/workflows/os_comp.yml
```diff
@@ -25,7 +25,6 @@ jobs:
     - name: Run tests
       run: LD_LIBRARY_PATH=/usr/local/share/boost/1.69.0/lib/:$(rustc --print sysroot)/lib:$LD_LIBRARY_PATH python3 run_tests.py
       env:
-        CI: '1'
```
mensinda, line 28: Why remove this? There is still special logic in (at least) `run_project_tests.py` for this env var.

mensinda: Ah, sorry ignore this. I should have read your initial comment more carefully...

scivision: Yes I brought Actions to Meson early; it took several months for some ordinary features to get added to Actions.

PR #7694, opened 2020-09-04
introspect: add target ids for executable/args/depends to test and benchmark JSON

bonzini:
Add the paths to the dependencies, as computed by the backend, to the introspection data for tests and benchmarks.  Without this information, IDEs must update the entire build before running any test.  They can now instead selectively build the test executable itself and anything that is needed to run it.

### DIFF
--- a/run_unittests.py
+++ b/run_unittests.py
@@ -4281,6 +4281,7 @@ recommended as it is not supported on some platforms''')
             ('suite', list),
             ('is_parallel', bool),
             ('protocol', str),
+            ('depends', list),
         ]
 
         buildoptions_keylist = [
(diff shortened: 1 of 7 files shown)


mensinda:
Why use the full path instead of the target ID? This would make lookups in the target introspection easier (while, admittedly, requiring another lookup to get the path).

bonzini:
Good idea. Would that work with CustomTargetIndex dependencies?

mensinda:
I don't know. I am not sure that we even produce introspection information for custom targets... But if we don't, adding them would be a good idea anyway.

bonzini:
Yes, we do. But we don't allow depends to be CustomTargetIndex so it's all good. :)

mensinda:
Glad to hear that :)

Anyway, this needs a test case (see `test_introspect_json_dump` for instance).

mensinda:
Also, `dependencies` might not be the best name here, since it can be easily confused with external dependencies (see `intro-dependencies.json`). Also, the docs need a `*(added in 0.56.0)*` note for the new key.

run_unittests.py
```diff
@@ -4338,12 +4338,22 @@ def assertKeyTypes(key_type_list, obj):
...
+            print(json.dump(i, sys.stdout))
+            for group in i['target_sources']:
+                src_to_id.update({os.path.relpath(src, testdir): i['id']
+                                  for src in group['sources']})
+
         # Check Tests and benchmarks
         tests_to_find = ['test case 1', 'test case 2', 'benchmark 1']
+        deps_to_find = {'test case 2': [src_to_id['t3.cpp']]}
         for i in res['benchmarks'] + res['tests']:
             assertKeyTypes(test_keylist, i)
```
mensinda, line 4353: please update `test_keylist` accordingly.

bonzini:
Thanks @mensinda for the review. I have made the patch more comprehensive, including all required target ids and not just the extra dependencies.

mensinda (approved):
LGTM, once the CI is green

PR #7708, opened 2020-09-08
IntrospectInterpreter: allow dictionaries with nonconstant keys

bonzini:
Something like {a: foo} currently stymies the IntrospectionInterpreter and breaks introspection of the source directory.  The fix is just to walk the keys and return a dummy dictionary.

### DIFF
--- a/mesonbuild/ast/interpreter.py
+++ b/mesonbuild/ast/interpreter.py
@@ -192,6 +192,16 @@ class AstInterpreter(interpreterbase.InterpreterBase):
         self.evaluate_statement(node.trueblock)
         self.evaluate_statement(node.falseblock)
 
+    def evaluate_dictstatement(self, node: mparser.DictNode) -> TYPE_nkwargs:
+        (arguments, kwargs) = self.reduce_arguments(node.args, resolve_key_nodes=False)
+        assert (not arguments)
+        self.argument_depth += 1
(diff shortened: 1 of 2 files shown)


bonzini:
Would be nice to have in 0.55.2.

mesonbuild/ast/interpreter.py
```diff
@@ -192,6 +192,16 @@ def evaluate_ternary(self, node: TernaryNode) -> None:
         self.evaluate_statement(node.trueblock)
         self.evaluate_statement(node.falseblock)
 
+    def evaluate_dictstatement(self, node: mparser.DictNode) -> T.Dict[str, T.Any]:
```
mensinda, line 195: This should return `TYPE_nkwargs`, not `T.Dict[str, T.Any]`.

mensinda:
Besides this, this LGTM and should be good for 0.55.2

mesonbuild/ast/interpreter.py
bonzini: I don't understand why, but I'll oblige. :)

mensinda: It's because that is also the exact return type from the `InterpreterBase` and otherwise `mypy` would be very unhappy once #7657 is merged :)

mensinda (approved):
LGTM

Issue #7645, opened 2020-08-26
WINDRES environment variable not respected in Ninja build


francesco-st:
## Problem statement

When setting the WINDRES environment variable Meson does not actually use its value in the Ninja build file.

For example when using the variable:

```sh
export WINDRES="windres -F pe-i386"
```
the arguments `-F pe-i386` does not appear in the ninja build file and windres is not called with the `-F pe-i386` arguments as requested.

Please note that when the command `meson setup` is given Meson acknowledges the WINDRES variable. Here an excerpt from the output:
```
Using 'WINDRES' from environment with value: 'windres -F pe-i386'
Windows resource compiler: GNU windres (GNU Binutils) 2.33.1
```

## Reproduce the problem

Please find in attachment a minimal test case to illustrate the problem. If the variable WINDRES is set and exported, it is ignored by Meson in the ninja build file.

[windres-test.tar.gz](https://github.com/mesonbuild/meson/files/5095774/windres-test.tar.gz)

## Expected behavior

the WINDRES environment variable should be used to call windres in the ninja build file.

## system parameters

- plain native build
- the OS is Windows 10 on x64 using MSYS2
- Meson 0.55.1 (provided by the MSYS package mingw32/mingw-w64-i686-meson)
- Python version is 3.8.5 (provided by the MSYS package mingw32/mingw-w64-i686-python)
- Ninja version is 1.10.0 (provided by the MSYS package mingw32/mingw-w64-i686-ninja)

nirbheek:
This is a regression in 0.55.

PR #7709, opened 2020-09-08
Fix searching of tools such as windres via the env on Windows

nirbheek:
```
Fix picking up tools with args from the env on Windows

This was a regression in https://github.com/mesonbuild/meson/pull/7059.
We do not need to do a full search for all windows special cases. We
only want to check whether the command provided is actually a full
path to a script so we can add the interpreter.
```
Fixes https://github.com/mesonbuild/meson/issues/7645

```
Add some type hints to the ExternalProgram class

```

### DIFF
--- a/mesonbuild/dependencies/base.py
+++ b/mesonbuild/dependencies/base.py
@@ -1804,7 +1804,14 @@ class ExternalProgram:
             if mesonlib.is_windows():
                 cmd = self.command[0]
                 args = self.command[1:]
-                self.command = self._search_windows_special_cases(name, cmd) + args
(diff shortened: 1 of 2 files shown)


nirbheek:
> ERROR: Clock skew detected. File /private/var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/tmpu_elnpp1/build1/../../../../../../../../Users/travis/build/mesonbuild/meson/test cases/common/1 trivial/meson.build has a time stamp 1136.2147s in the future.

Unrelated CI failure.

Issue #7627, opened 2020-08-21
meson install does not print a FAILED: message when an install script fails


m1027:
When calling meson install, my custom install.sh, added by a subdir, is
searched and detected but silently not executed! When instead calling ninja
install, it works.

Is it something trivial?  I am using meson-0.54.2, ninja-1.10.0.

How to reproduce:

    mkdir -p main/subdir/my_module/scripts
    
    cd main
    
    cat <<EOF > meson.build
    project('main', 'c')
    subdir('subdir/my_module')
    EOF
    
    cat <<EOF > subdir/my_module/meson.build
    meson.add_install_script('scripts/install.sh')
    EOF
    
    cat <<EOF > subdir/my_module/scripts/install.sh
    #!/bin/bash
    exit 1
    EOF
    
    meson build
    
    meson install -C build
    # (the script's exit 1 does not produce and error but should)
    
    ninja -C build install
    # (works as expected and exits with error)

nirbheek:
```
$ meson install -C _build
ninja: Entering directory `_build'
ninja: no work to do.
Running custom install script 'bash C:/projects/tmp/install-script-subdir/main/subdir/my_module/scripts/install.sh'
I WAS RUN
$ echo $?
1
```

I added an `echo` to the script. So the script is run, and `meson install` exits with the status code 1. Only thing is that it doesn't print a `FAILED:` message.

m1027:
Ok, I tested the case with *two* subsequent subdir modules now.  Both install
scripts terminating with exit 1:

As expected: 'meson install' stops after the first script terminating with
error.

HOWEVER: Due to the missing FAILED message the user might not realize that
subsequent scripts have not been called! I consider this quite *serious*.

PR #7710, opened 2020-09-08
minstall: Print a big FAILED when a script fails

nirbheek:
We don't run any further scripts when this happens, so we need to print a big error. The exit code was already correct.

Fixes https://github.com/mesonbuild/meson/issues/7627

The output now looks like:

```
$ /c/projects/meson/meson.git/meson.py install -C _build
ninja: Entering directory `_build'
ninja: no work to do.
Running custom install script 'bash C:/projects/tmp/install-script-subdir/main/subdir/my_module/scripts/install.sh'
I WAS RUN
FAILED: install script 'bash C:/projects/tmp/install-script-subdir/main/subdir/my_module/scripts/install.sh' exit code 1, stopped
```

### DIFF
--- a/mesonbuild/minstall.py
+++ b/mesonbuild/minstall.py
@@ -437,11 +437,13 @@ class Installer:
             self.log('Running custom install script {!r}'.format(name))
             try:
                 rc = subprocess.call(script + args, env=child_env)
-                if rc != 0:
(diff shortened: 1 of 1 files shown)


xclaesse:
LGTM.

Issue #2354, opened 2017-09-21
Eliminate error about build directory being from a previous version of Meson


pwithnall:
In my auto-builder, I frequently get errors like this:
```
Meson encountered an error:
Build directory has been generated with Meson version 0.42.0, which is incompatible with current version 0.42.1.
Please delete this build directory AND create a new one.
FAILED: build.ninja 
```

Why can’t Meson delete the build directory itself, if it knows it’s out of date? This causes builds to fail spuriously every time someone upgrades the Meson version in GNOME’s JHBuild moduleset.

QuLogic:
Dupe of #2026.

nirbheek:
See also: https://github.com/mesonbuild/meson/pull/1662 which actually implements this.

pwithnall:
#1662 looks like what I want, thanks!

> Dupe of #2026.

Not quite. I Meson to do whatever it takes to get the build to succeed; an error message which includes the old configuration information is just as useless to me as an error message which doesn’t.

nirbheek:
I'll rebase and update #1622 then.

YaLTeR:
I'm hitting this on my project, except I'm doing `ninja install` as usual and getting a super cryptic exception in the middle of meson's source:
```
└─ ninja install
[0/1] Installing files.
Installing meson-out/org.gnome.gitlab.YaLTeR.VideoTrimmerDevel.desktop to /home/yalter/source/rs/video-trimmer/install/share/applications
Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/mesonbuild/mesonmain.py", line 131, in run
    return options.run_func(options)
  File "/usr/lib/python3.8/site-packages/mesonbuild/minstall.py", line 542, in run
    installer.do_install(datafilename)
  File "/usr/lib/python3.8/site-packages/mesonbuild/minstall.py", line 362, in do_install
    self.install_targets(d)
  File "/usr/lib/python3.8/site-packages/mesonbuild/minstall.py", line 515, in install_targets
    depfixer.fix_rpath(outname, t.rpath_dirs_to_remove, install_rpath, final_path,
AttributeError: 'TargetInstallData' object has no attribute 'rpath_dirs_to_remove'
FAILED: meson-install 
/usr/bin/meson install --no-rebuild
ninja: build stopped: subcommand failed.
```
Only upon trying `meson configure` I find out that my meson has updated and now refuses to work with the build directory:
```
└─ meson configure

[...]

PR #7717, opened 2020-09-10
minstall: Add version field to install data

nirbheek:
```
mtest: Refactor test data checking

```

```
minstall: Add version field to install data

And check the install data in the same way that mtest checks
serialisation data.
```
Fixes https://github.com/mesonbuild/meson/issues/2354


nirbheek:
Classic, testing on Windows didn't test installation because default backend is vs ;)

mensinda (approved):
LGTM

PR #7718, opened 2020-09-10
Add a test run for MSVC ARM64 cross

jon-turney:


mesonbuild/environment.py
```diff
@@ -1321,7 +1321,7 @@ def sanitize(p):
                 linker = self._guess_win_linker(['link'], cls, for_machine)
                 return cls(
                     compiler, version, for_machine, is_cross, info, exe_wrap,
-                    target, linker=linker)
+                    target, full_version=cl_signature, linker=linker)
```
nirbheek, line 1324: Hmm, what's this for? Is it for the same reason as https://github.com/mesonbuild/meson/pull/7021/commits/c01c4f01075f34ba22f6314473f5e44a4bfb8c53 for ARM64 detection?

jon-turney: No.  I think full_version is only used in logging, so this is just so we can see that the cross-compiler is being used.

jon-turney:
It should probably be noted that this test job is running in a slightly precarious environment: the cross-compiler `cl` is in the PATH, so will be (incorrectly) detected as the build machine compiler as well, but we've arranged things so we never use it (so we can never end up trying to run it's products, which will fail).

To fix that, we would probably need to pull the call to vcvars.bat out of install.ps1 (for this case) and put it into the machine files, as discussed a bit at https://github.com/mesonbuild/meson/issues/4402#issuecomment-690275118

test cases/common/122 llvm ir and assembly/meson.build
```diff
@@ -34,6 +34,9 @@ foreach lang : ['c', 'cpp']
       ml = find_program('ml', required: false)
     elif cpu == 'x86_64'
       ml = find_program('ml64', required: false)
+    elif cpu == 'aarch64'
+      # XXX: no idea what assembler we should use here
```
nirbheek, line 38: armasm64 is what you want: https://gitlab.freedesktop.org/gstreamer/meson-ports/libffi/-/blob/meson/src/meson.build#L104-132

test cases/common/121 shared module/meson.build
```diff
@@ -1,5 +1,14 @@
 project('shared module', 'c')
 
+# Windows on ARM doesn't appear to support the toolhelp API we use to emulate
+# runtime symbol resolution in this test.
+#
+# XXX: This should be written as a check if the macro
+# WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP) evaluates to true?
+if host_machine.system() == 'windows' and ['aarch64', 'arm'].contains(host_machine.cpu_family())
+  error('MESON_SKIP_TEST windows on ARM does not support this test.')
+endif
```
nirbheek, line 10: Loading of modules is done via `LoadPackagedLibrary`, and then to run it, you need to bundle the DLLs as assets (that part is not needed because we can't run cross-compiled tests anyway).

cross/arm64cl.txt
```diff
@@ -0,0 +1,14 @@
+[binaries]
+c = 'cl'
+cpp = 'cl'
+ar = 'lib'
+windres = 'rc'
+
+[properties]
+c_args = ['-DWINAPI_FAMILY=WINAPI_FAMILY_APP']
```
nirbheek, line 8: We should add `WindowsApp.lib` in c_link_args, that gives us the actual configuration people cross-compile to ARM64 with. Also needs cpp_args and cpp_link_args?

mstorsjo: These options aren't mandatory when targeting ARM64 windows - you can build full traditional win32 apps for ARM64, you don't need to target UWP.

And since some time ago, you don't need either `-DWINAPI_FAMILY=WINAPI_FAMILY_APP` (or `-D_ARM_WINAPI_PARTITION_DESKTOP_SDK_AVAILABLE=1`) when targeting ARM32 either, you can build totally regular desktop apps for that nowadays as well. (Since WinSDK 10.0.17763.0, the default define `_ARM_WINAPI_PARTITION_DESKTOP_SDK_AVAILABLE` in ucrt/corecrt.h has been changed from 0 to 1.)

test cases/common/122 llvm ir and assembly/meson.build
jon-turney: Thanks.

test cases/common/121 shared module/meson.build
jon-turney: The issue here isn't `LoadLibrary()` not being available for UWP, but the toolhelp API, which we use to grovel over all loaded modules, looking at their exported symbols, to emulate run-time linkage.  I rewrote this a bit so it does a more specific test for that.

jon-turney:
> To work around this in Cerbero, we use a native file like so:

I'm kind of surprised this works.  If we have `project('foo', 'c'), won't meson try to run `false` to determine what kind of compiler it is, and stop when fails?

But then I added it and it seems to work in the test run, so idk what's going on.

cross/arm64cl.txt
mstorsjo: Fair enough - just wanted to be sure there's no misunderstanding that it's required for this target.

jon-turney:
> I'm kind of surprised this works. If we have `project('foo', 'c')`, won't meson try to run `false` to determine what kind of compiler it is, and stop when fails?
> 
> But then I added it and it seems to work in the test run, so idk what's going on.

Only fatal if the compiler is required, of couse.

Added reporting of the native compilers found to project test output (when cross-compiling) so it can be seen this is working.

nirbheek:
> Only fatal if the compiler is required, of couse.
> 

Yeah, in Cerbero we now default to 'no build compiler' when cross-compiling except if a project specifically requests it. Only one project out of the dozens we build has needed it so far: fribidi.

test cases/common/121 shared module/meson.build
nirbheek: Makes sense :)

PR #7723, opened 2020-09-11
Fixes for meson subprojects update

xclaesse:


mesonbuild/msubprojects.py
```diff
@@ -131,6 +131,7 @@ def update_git(wrap, repo_dir, options):
     # is needed for http case were new remote branches wouldn't be discovered
     # otherwise. After this command, FETCH_HEAD is the revision we want.
     try:
+        git_output(['remote', 'update', 'origin'], repo_dir)
         git_output(['fetch', 'origin', revision], repo_dir)
```
nirbheek, line 135: Shouldn't we 'just' call `git fetch origin` here?

eli-schwartz: > Fetch only the revision we need, this avoids fetching useless branches

So now we fetch utterly everything, we can just remove the following line...

`git remote update origin` is identical to invoking `git fetch origin`, but previously (a long time ago) `git remote update` was the only way to fetch multiple remotes.

xclaesse: Ok, changed to `git fetch origin`. I think we still need both calls however, consider this situation:

I already have 1.18 branch locally but it is outdated, `git fetch origin` will update `origin/1.18` but `git reset --hard 1.18` will reset to the local version. And I cannot do `git reset --hard origin/1.18` because that fails if 1.18 is a tag and not a branch. This is why I need `git fetch origin 1.18` to set FETCH_HEAD to the right commit, so `git reset --hard FETCH_HEAD` will always work regardless if revision is a tag, branch or commit id.

I think I also need `git fetch origin <revision>` in the case revision is a commit id not part of any references, don't think `git fetch origin` pulls that commit in that case, does it? But I admit this is corner case.

eli-schwartz: This is all already the case before your commit. The question is why you need to fetch the entire remote repo before fetching the specific revision you're interested in. `git fetch origin 1.18` without first `git fetch origin`, should both fetch the content and set FETCH_HEAD, surely?

nirbheek:
There's another thing that needs fixing. Currently, `meson subprojects update --reset` outputs this for every repo: `No local changes to save`. We should suppress that. Either by calling `git stash` only if the tree is dirty, or by only printing its output when it stashed something.

mesonbuild/msubprojects.py
xclaesse: The problem is when 1.18 is a new branch on remote, `git fetch origin 1.18` will fetch the commit but won't create the reference locally, so `git checkout 1.18` fails.

xclaesse:
> `No local changes to save`

Fixed now.

mesonbuild/msubprojects.py
nirbheek: Is this conversation resolved now? I am not sure I understand these intricacies of git.

mesonbuild/msubprojects.py
```diff
@@ -47,6 +47,9 @@ def git_output(cmd, workingdir):
     return quiet_git(cmd, workingdir, check=True)[1]
 
 def git_stash(workingdir):
+    out = git_output(['diff', '-s'], workingdir)
+    if not out.strip():
+        return
```
nirbheek, line 52: This will discard changes that have been staged but not committed.

nirbheek: Actually this will _always_ discard output. `-s` is the opposite of `-p`, and by default it will cause `git diff` to output nothing. Probably want to add a test for this now that we know this can break? :)

xclaesse: Ooops, that's a typo because I initially wrote `git status -s`. Changed to --stat. Good catch!

Yes we definitely need tests, but it's a bit complicated to simulate those git repositories. Do you know if we have helpers somewhere already? Sadly git wraps is largely untested atm I think.

mesonbuild/msubprojects.py
xclaesse: It's unfortunate to have 2 fetches, I'm sure we can find some git magic to avoid that. But IMHO it's not that bad for now. Unless @eli-schwartz has an idea?

nirbheek: `git fetch origin` (without any more arguments) is always supposed to do the right thing. Why aren't we just calling that? I am not sure I understood that from the conversation.

mesonbuild/msubprojects.py
nirbheek: No, this change is not enough. It will discard changes that have been staged. Observe:

```
$ vim meson.py
$ git diff --stat
 meson.py | 1 +
 1 file changed, 1 insertion(+)
$ git add meson.py
$ git diff --stat
$
```

nirbheek: Also `git status -s` will also show untracked files, which we *do* want to ignore.

nirbheek: `git status --short --untracked-files=no --ignore-submodules=untracked` looks like the best option, I think.

eli-schwartz: `git diff --quiet HEAD`

By default git-diff(1) shows the difference between the working tree and the staging index. Adding --stat merely suppresses the line diff, which you promptly ignore for performance reasons I assume. It still returns "some" stdout in and *only* in situations where git-diff(1) would.

- `--quiet` tells git-diff(1) to completely avoid emitting output. Instead, it emits a useful return code for scripting.
- Passing HEAD as the diff target will make the comparison be between the working tree and the HEAD commit; it therefore shows both staged and unstaged changes.

mesonbuild/msubprojects.py
eli-schwartz: > The problem is when 1.18 is a new branch on remote, `git fetch origin 1.18` will fetch the commit but won't create the reference locally, so `git checkout 1.18` fails.

This works for me. I don't understand your problem but I continue to believe the commit is wrong.

Consider:

```
$ git branch -d -r origin/pythonbump
Deleted remote-tracking branch origin/pythonbump (was 754fbef3d).
$ git checkout pythonbump --
fatal: invalid reference: pythonbump
$ git fetch origin pythonbump
From https://github.com/mesonbuild/meson
 * branch                pythonbump -> FETCH_HEAD
 * [new branch]          pythonbump -> origin/pythonbump
$ git checkout pythonbump --
Previous HEAD position was b09efdbbe Revert "Do not automatically set warning flags if buildtype is 'plain'"
Branch 'pythonbump' set up to track remote branch 'pythonbump' from 'origin'.
Switched to a new branch 'pythonbump'
```

xclaesse: Sorry, I think I've been too confusing. There are many cases:
- revision is a new remote branch that does not exist locally
  * `git fetch origin newbranch && git checkout newbranch` works
- revision is a branch that exists locally but got updated remotely
  * `git fetch origin newbranch && git checkout newbranch` does not work because it only checkout the local outdated branch.
  * `git fetch origin newbranch && git checkout newbranch && git reset --hard FETCH_HEAD` works.
- revision is a tag
  * `git fetch origin newtag && git checkout newtag` does not work because the tag did not get created locally.
  * `git fetch origin && git checkout newtag && git reset --hard FETCH_HEAD` does not work because FETCH_HEAD is not set.
  * `git fetch origin && git fetch origin newtag && git checkout newtag && git reset --hard FETCH_HEAD` works.

That last command actually works in all those cases without having to know if revision is a tag or branch.

xclaesse: > `git fetch origin` (without any more arguments) is always supposed to do the right thing. Why aren't we just calling that? I am not sure I understood that from the conversation.

You get a problem in case 2.2 and 2.3 above: you don't know to which commit you need to update an existing outdated branch. You could do `git reset --hard origin/somebranch` but that does not work if the revision (`somebranch` here) is tag.

xclaesse: > > The problem is when 1.18 is a new branch on remote, `git fetch origin 1.18` will fetch the commit but won't create the reference locally, so `git checkout 1.18` fails.
> 
> This works for me. I don't understand your problem but I continue to believe the commit is wrong.

Sorry, I meant if 1.18 is a new tag actually. Your command examples won't work in the case pythonbump is a tag:
```
$ git fetch origin pythonbump
 * tag               pythonbump -> FETCH_HEAD
$ git checkout pythonbump
error: pathspec 'pythonbump' did not match any file(s) known to git
```

mesonbuild/msubprojects.py
xclaesse: > No, this change is not enough. It will discard changes that have been staged. Observe:
Good point.

> Also `git status -s` will also show untracked files, which we _do_ want to ignore:
Right, that's why I changed to git diff.

@eli-schwartz thanks for the tip, I think `git diff --quiet HEAD` would be best then, I'll update.

mesonbuild/msubprojects.py
eli-schwartz:
```
f() {
    git fetch origin +refs/tags/*$1:refs/tags/*$1 +refs/heads/*$1:refs/remotes/origin/*$1
}
f 0.55.3
f pythonbump
```

Does set FET_HEAD, even though it doesn't print it out.

```
case $(git cat-file -t FETCH_HEAD) in
    tag)
        ref_to_use=0.55.3
        ;;
    commit)
        ref_to_use=origin/0.55.3
        ;;
esac
```

mesonbuild/msubprojects.py
xclaesse: Fixed, and covered by unit test.

mesonbuild/msubprojects.py
eli-schwartz:
```
git fetch --refmap '+refs/heads/*:refs/remotes/origin/*' --refmap '+refs/tags/*:refs/tags/*' origin <ref>
```

More accurate than passing the refspec as the ref to fetch. refmap lets us add a non-default tag map for fetching.

Continue with `git cat-file -t FETCH_HEAD` to see the correct way to handle checking it out. Note tags are always going to be a detached HEAD.

xclaesse: @eli-schwartz almost.... but fails in the case revision is a commit id. Forgot to document that case in my list above. The good news is I know have a unit test covering that too.

xclaesse: @eli-schwartz ohh, with your updated refmap command it pass all my tests, nice! Note that `git cat-file -t FETCH_HEAD` always return 'commit' but it works without doing that, we never need `origin/` AFAIK.

eli-schwartz: If it returns "commit" then the fetched content was a branch, if it returns "tag" then the fetched content was a tag.

I guess you could, in fact, `git reset --hard FETCH_HEAD` though. No need to figure out if 0.55.3 needs to be referred to as `origin/` if you just use `FETCH_HEAD`.

xclaesse: Updated this PR with your suggestion @eli-schwartz. It works in all cases I've thought about and has the advantage of fetching a single reference instead of updating everything which could save some bandwidth. Thanks!

mesonbuild/msubprojects.py
```diff
@@ -125,11 +131,16 @@ def update_git(wrap, repo_dir, options):
         mlog.log(mlog.red(e.output))
         mlog.log(mlog.red(str(e)))
         return False
-    # Fetch only the revision we need, this avoids fetching useless branches and
-    # is needed for http case were new remote branches wouldn't be discovered
-    # otherwise. After this command, FETCH_HEAD is the revision we want.
```
eli-schwartz, line 130: Can we put back this comment, and update the commit message to be more truthful about the change being performed?

eli-schwartz (changes requested):

mesonbuild/msubprojects.py
xclaesse: I've put back part of that message, I think the http part was BS. Updated commit message too.

eli-schwartz: I agree the http part seems quite odd and it makes sense to take advantage of touching this code to *also* fix the comment.

nirbheek:
CI failure is spurious, clock skew on macOS.

PR #7724, opened 2020-09-11
unit tests: Don't use pytest when running single tests

nirbheek:
```
unit tests: Don't use pytest when running single tests

On my machine this spawns 24 processes and then runs like the single
test I asked it to run. With this change, running a single test goes
from 7 seconds to less than a second.
```

```
unit tests: Print total time taken for running tests

```

Split out from https://github.com/mesonbuild/meson/pull/7021.


jon-turney (approved):

nirbheek:
Test failure is spurious.

PR #7725, opened 2020-09-11
Fix Korean MSVC compiler target arch detection

LBPHacker:
This fixes an issue essentially identical to #6757, except affecting users of the Korean locale (and possibly a different version of MSVC, if the claim in sthagen/meson#64 is to be believed). The fix involves allowing closing parenthesis to terminate the architecture substring, beside space and EOL. The signature line this adds support for is:
```
Microsoft (R) C/C++ 최적화 컴파일러 버전 19.27.29111(x64)
```

### DIFF
--- a/mesonbuild/environment.py
+++ b/mesonbuild/environment.py
@@ -1311,7 +1311,7 @@ class Environment:
                     m = 'Failed to detect MSVC compiler version: stderr was\n{!r}'
                     raise EnvironmentException(m.format(err))
                 cl_signature = lookat.split('\n')[0]
-                match = re.search('.*(x86|x64|ARM|ARM64)( |$)', cl_signature)
+                match = re.search(r'.*(x86|x64|ARM|ARM64)([^_A-Za-z0-9]|$)', cl_signature)
                 if match:
                     target = match.group(1)
                 else:


nirbheek:
What's the output of the following commands for you?

```
> cl /logo --version
> set VSLANG=1033
> cl /logo --version
```

LBPHacker:
Both `cl` invocations, before and after setting `VSLANG`, result in this:
```
Microsoft (R) C/C++ 최적화 컴파일러 버전 19.27.29111(x64)
Copyright (c) Microsoft Corporation. All rights reserved.

cl : 명령줄 warning D9002 : 알 수 없는 '/logo' 옵션을 무시합니다.
cl : 명령줄 warning D9002 : 알 수 없는 '--version' 옵션을 무시합니다.
cl : 명령줄 error D8003 : 소스 파일 이름이 없습니다.
```
... which suggests that it completely ignores the environment variable. Based on the warnings, it also seems to be ignoring `/logo` and `--version`. I don't think I've done anything incorrectly.

Furthermore, I've discovered by chance that the x86 version uses yet another format string, giving this for `cl /?`:
```
x86용 Microsoft (R) C/C++ 최적화 컴파일러 버전 19.27.29111
```
Needless to say, this invalidates my current fix. I guess the regexp could be extended to allow all of `[^_A-Za-z0-9]` (anything I missed?) to terminate the architecture substring.

This is getting out of hand :(

nirbheek:
> ... which suggests that it completely ignores the environment variable. 

Yeah, that's not good.

> Based on the warnings, it also seems to be ignoring `/logo` and `--version`

This is fine, it's basically what we do during compiler detection. Need to force the compiler to give us output that identifies it.

Since VSLANG didn't work, I think your PR is correct for now, but it is likely broken in some other language.

I think we should run each of our VS Azure jobs with different codepages enabled. That way we'll be sure that this doesn't regress in the future, and that other languages also work. @LBPHacker can you try doing that? You want to look at these files:

```
azure-pipelines.yml
ci/azure_steps.yml
ci/run.ps1
```

Adding a `chcp <value>` somewhere in there might work? It will need that codepage pre-installed. I am not sure if the azure images have that.

LBPHacker:
I'll apply the fix I mentioned to correctly extract `x86` from the x86 signature line in my previous comment then. I can't imagine any signature which `(x86|...)([^_A-Za-z0-9]|$)` wouldn't match correctly.

As for the chcp settings, sure, I'll look into those, but they're uncharted territory for me. (As a result, I'm not quite sure how the active codepage would have any effect on cl.) One thing I can tell from a quick test though is that chcping away from 437 on a machine on which that's the default has no effect on the output of `cl /?` in any way, including the language it's in. I don't currently have access to the machine that produced the Korean signature lines, but I'll try `chcp` there when I do; maybe it'll behave differently.

LBPHacker:
I tried `chcp` on the Korean machine and it didn't affect `cl /?` at all. I also haven't been able to get output in any language other than the "underlying language" (i.e. not just UI) of VS, which I in turn haven't been able to control in any way other than by installing a version of Windows with the desired "underlying language". To be frank, I'm not sure what "underlying" means here; I define it as "the language the installer is in". In fact, locales and really the entirety of Windows is beyond me, so I'd be in over my head if I tried to update the Azure jobs to use different locales.

@nirbheek I'd like to just stick to fixing the x86 problem (see above) if that's alright.

LBPHacker:
Can someone tell me whether the msys tests fail due to something I've done, and if they do, what exactly? I see `Expected file usr\lib\modules\libnosyms.dll.a missing.`, which I can't quite trace back to this fix.

nirbheek:
The MSYS2 failures are unrelated to this PR. They are happening on other PRs too.

PR #7727, opened 2020-09-11
project tests: Verify that UWP cross binaries use the right arch

nirbheek:
```
Add build machine compilers to project tests compiler report

When cross-compiling, also report build machine compilers in the
compiler report produced by run_project_tests.py
```

```
project tests: Rename a variable for clarity

```

```
Rename [properties] to [built-in options]

This is the new syntax. Avoids loud deprecation warnings.
```

```
project tests: Also check whether introspect works

```

```
project tests: Verify that UWP cross binaries use the right arch

This is a test for https://github.com/mesonbuild/meson/pull/7021, to
verify that `link.exe` uses the correct architecture when targeting
ARM64. Can be extended to other cross targets later.
```


run_project_tests.py
```diff
@@ -1176,6 +1177,20 @@ def check_format():
                     continue
                 check_file(root / file)
 
+def get_exe_arch(exepath):
```
jon-turney, line 1180: The patch commentary isn't very clear about what pieces of information this is checking are identical, and there's not really anything UWP-specific about this.

I guess ultimately this is checking that `host_machine.cpu_family()` matches the architecture of a built executable (which should be trivially true when we are not cross-compiling and the cpu has been detected from the compiler, but is a useful test to do when cross-compiling)

I think this could be written as a stand-alone test, in a way similar to `windows/16 gui app`.

The rest of this patch set seems like good cleanup.

nirbheek: > I guess ultimately this is checking that host_machine.cpu_family() matches the architecture of a built executable 

Yes, the bug I was seeing earlier was that the default arch for link.exe is either x86 or x64.

If you pass `x86`, `x86_x64`, `x86_arm` or `x86_arm64` to `vcvarsall.bat`, you will get a link.exe that defaults to outputting for `X86`. 

If you pass `amd64`, `amd64_x86`, `amd64_arm` or `amd64_arm64` to `vcvarsall.bat`, you will get a link.exe that defaults to outputting for `X64`.

So we particularly need to verify that we're not accidentally outputting for the wrong arch when cross-compiling with UWP, and so this test only implements it for MSVC (with `dumpbin`).

> I think this could be written as a stand-alone test, in a way similar to windows/16 gui app.

I thought of that, but we can't always run tests when cross-compiling (needs exe wrapper), so I'd have to either write it as an install script (which means it won't get run when we are using the vs backend).

I guess it can be a `custom_target` that takes the executable as an argument and is `build_by_default: true`, and is in the "common tests" directory. Does that sound better to you?

jon-turney: > I thought of that, but we can't always run tests when cross-compiling (needs exe wrapper), so I'd have to either write it as an install script (which means it won't get run when we are using the vs backend).

Ah, I see. We could run the test in this case, because it doesn't need to execute the trivial host executable, but we don't in general, unless there is an exe wrapper, because we don't know that.

> I guess it can be a `custom_target` that takes the executable as an argument and is `build_by_default: true`, and is in the "common tests" directory. Does that sound better to you?

If that works, sure.

nirbheek: I was thinking, we could actually add a kwarg to `test()` like `needs_exe_wrapper:` which defaults to `true`. This is a somewhat common use-case. Maybe that should be done separately though :)

nirbheek:
> This pull request **introduces 1 alert** when merging [ed56e0e](https://github.com/mesonbuild/meson/commit/ed56e0eaf955b40be358e179a539853148c42a5e) into [e9fea0f](https://github.com/mesonbuild/meson/commit/e9fea0fecfc14c43d609fa53bbf92a97549ae2db) - [view on LGTM.com](https://lgtm.com/projects/g/mesonbuild/meson/rev/pr-f67f8f3600ba904bf658fd002354e8c27c13f689)

Fixed.

cross/arm64cl.txt
```diff
@@ -4,7 +4,7 @@ cpp = 'cl'
 ar = 'lib'
 windres = 'rc'
 
-[properties]
+[built-in options]
```
jon-turney, line 7: This renaming should probably be done in all the files in cross/ which have this section?

jon-turney (approved):

cross/arm64cl.txt
nirbheek: Other cross files already do that correctly.

PR #7729, opened 2020-09-12
backend/vs: Add include directories to projects as well

vvainola:
- "Go To Document" action previously only worked on c/cpp files which had the include directories set but it was not possible to move from header to another header.

Fixes same thing as in https://github.com/mesonbuild/meson/pull/7385

### DIFF
--- a/mesonbuild/backend/vs2010backend.py
+++ b/mesonbuild/backend/vs2010backend.py
@@ -966,9 +966,7 @@ class Vs2010Backend(backends.Backend):
             t_inc_dirs = [self.relpath(self.get_target_private_dir(target),
                                        self.get_target_dir(target))]
             if target.implicit_include_directories:
-                t_inc_dirs += ['.']
-            if target.implicit_include_directories:
-                t_inc_dirs += [proj_to_src_dir]
+                t_inc_dirs += ['.', proj_to_src_dir]
             args += ['-I' + arg for arg in t_inc_dirs]
 
         # Split preprocessor defines and include directories out of the list of
@@ -983,9 +981,8 @@ class Vs2010Backend(backends.Backend):
                     else:
                         define = arg[2:]
                     # De-dup
-                    if define in file_defines[l]:
-                        file_defines[l].remove(define)
-                    file_defines[l].append(define)
+                    if define not in file_defines[l]:
+                        file_defines[l].append(define)
                 elif arg.startswith(('-I', '/I')) or arg == '%(AdditionalIncludeDirectories)':
                     file_args[l].remove(arg)
                     # Don't escape the marker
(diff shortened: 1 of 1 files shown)


mesonbuild/backend/vs2010backend.py
```diff
@@ -966,9 +966,7 @@ def gen_vcxproj(self, target, ofname, guid):
             t_inc_dirs = [self.relpath(self.get_target_private_dir(target),
                                        self.get_target_dir(target))]
             if target.implicit_include_directories:
-                t_inc_dirs += ['.']
-            if target.implicit_include_directories:
-                t_inc_dirs += [proj_to_src_dir]
+                t_inc_dirs += ['.', proj_to_src_dir]
```
nirbheek, line 969: This change seems unrelated, should be split into a new commit.

mesonbuild/backend/vs2010backend.py
```diff
@@ -983,9 +981,8 @@ def gen_vcxproj(self, target, ofname, guid):
                     else:
                         define = arg[2:]
                     # De-dup
-                    if define in file_defines[l]:
-                        file_defines[l].remove(define)
-                    file_defines[l].append(define)
+                    if define not in file_defines[l]:
+                        file_defines[l].append(define)
```
nirbheek, line 985: This change also seems unrelated, but it's cleanup like the above change, so can be clubbed with it.

mesonbuild/backend/vs2010backend.py
```diff
@@ -1025,8 +1025,6 @@ def gen_vcxproj(self, target, ofname, guid):
         if len(target_args) > 0:
             target_args.append('%(AdditionalOptions)')
             ET.SubElement(clconf, "AdditionalOptions").text = ' '.join(target_args)
-
-        target_inc_dirs.append('%(AdditionalIncludeDirectories)')
```
nirbheek, line 1029: I am not sure I understand why this has been removed.

nirbheek (commented):
Also, could you please make sure that your commit message body is wrapped at 70 characters? That's the standard with git. Thanks!

mesonbuild/backend/vs2010backend.py
vvainola: Split cleanup into own commit.

mesonbuild/backend/vs2010backend.py
vvainola: Split cleanup into own commit.

mesonbuild/backend/vs2010backend.py
vvainola: Split cleanup into own commit.

nirbheek:
Thanks!

xclaesse:
This seems to have broken CI in master.

nirbheek:
This code is not called on MSYS2 at all.

PR #7730, opened 2020-09-13
mcompile: fix "meson compile" for 32-bit MSVC build

vvainola:
Currently "meson compile" fails to "error MSB4126: The specified solution configuration "debug|x86" is invalid" in 32-bit builds.

### DIFF
--- a/mesonbuild/mcompile.py
+++ b/mesonbuild/mcompile.py
@@ -137,7 +137,7 @@ def generate_target_names_ninja(target: ParsedTargetName, builddir: Path, intros
     else:
         return [str(Path(out_file).relative_to(builddir.resolve())) for out_file in intro_target['filename']]
 
-def get_parsed_args_ninja(options: 'argparse.Namespace', builddir: Path) -> T.List[str]:
+def get_parsed_args_ninja(options: 'argparse.Namespace', builddir: Path) -> T.Tuple[T.List[str], T.Optional[T.Dict[str, str]]]:
     runner = detect_ninja()
     if runner is None:
         raise MesonException('Cannot find ninja.')
@@ -164,7 +164,7 @@ def get_parsed_args_ninja(options: 'argparse.Namespace', builddir: Path) -> T.Li
 
     cmd += options.ninja_args
 
-    return cmd
+    return cmd, None
 
 def generate_target_name_vs(target: ParsedTargetName, builddir: Path, introspect_data: dict) -> str:
     intro_target = get_target_from_intro_data(target, builddir, introspect_data)
@@ -179,7 +179,7 @@ def generate_target_name_vs(target: ParsedTargetName, builddir: Path, introspect
         target_name = str(rel_path / target_name)
     return target_name
 
-def get_parsed_args_vs(options: 'argparse.Namespace', builddir: Path) -> T.List[str]:
+def get_parsed_args_vs(options: 'argparse.Namespace', builddir: Path) -> T.Tuple[T.List[str], T.Optional[T.Dict[str, str]]]:
     slns = list(builddir.glob('*.sln'))
     assert len(slns) == 1, 'More than one solution in a project?'
     sln = slns[0]
@@ -229,9 +229,13 @@ def get_parsed_args_vs(options: 'argparse.Namespace', builddir: Path) -> T.List[
 
     cmd += options.vs_args
 
-    return cmd
+    # Remove platform from env so that msbuild does not pick x86 platform when solution platform is Win32
+    env = os.environ.copy()
+    del env['PLATFORM']
 
-def get_parsed_args_xcode(options: 'argparse.Namespace', builddir: Path) -> T.List[str]:
+    return cmd, env
+
+def get_parsed_args_xcode(options: 'argparse.Namespace', builddir: Path) -> T.Tuple[T.List[str], T.Optional[T.Dict[str, str]]]:
     runner = 'xcodebuild'
     if not shutil.which(runner):
         raise MesonException('Cannot find xcodebuild, did you install XCode?')
@@ -265,7 +269,7 @@ def get_parsed_args_xcode(options: 'argparse.Namespace', builddir: Path) -> T.Li
         pass
 
     cmd += options.xcode_args
-    return cmd
+    return cmd, None
 
 def add_arguments(parser: 'argparse.ArgumentParser') -> None:
     """Add compile specific arguments."""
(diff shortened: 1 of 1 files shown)


mesonbuild/mcompile.py
```diff
@@ -227,6 +227,12 @@ def get_parsed_args_vs(options: 'argparse.Namespace', builddir: Path) -> T.List[
 
     cmd += options.vs_args
 
+    # Add platform from sln file because msbuild assumes that for 32-bit build
+    # the platform is x86 when it should be Win32. The platform is extracted from first
+    # line that contains buildtype|platform e.g. debug|Win32
+    platform = re.search('(?<=(\|))[a-zA-Z0-9]*(?=(\n))', open(sln, 'r').read()).group()
+    cmd.append('-p:Platform=' + platform)
```
nirbheek, line 234: The error you're seeing is because the visual studio command prompt sets the `Platform` env var to `x64` for 64-bit and `x86` for 32-bit, and MSBuild picks it up. The correct solution here would be to unset that before calling `msbuild`. Our solutions only contain a single `config|arch` anyway.

nirbheek: Looking at the code, the best way is probably to return `cmd, env` from all `get_parsed_args_*()` functions and pass it via the `env=` kwarg to `Popen_safe`.

`get_parsed_args_ninja()` would return `cmd, None` and `gst_parsed_args_vs()` would return `cmd, env` where `env = os.environ.copy() \n del env['Platform']` or similar.

vvainola: Changed as suggested, seems to work fine. CI build would probably be nice because test scripts still do not work with 32-bit dev prompt since they call msbuild directly rather than "meson compile".

dcbaker (approved):
looks good to me, @nirbheek?

nirbheek:
LGTM.

> CI build would probably be nice because test scripts still do not work with 32-bit dev prompt since they call msbuild directly rather than "meson compile".

We also don't have a 32-bit visual studio build, but we could convert the vc2017x64vs build to 32-bit.

There are conflicts so it needs to be manually rebased.

nirbheek:
> We also don't have a 32-bit visual studio build, but we could convert the vc2017x64vs build to 32-bit.

This PR implements that: https://github.com/mesonbuild/meson/pull/7730. Can be merged after this PR.

mesonbuild/mcompile.py
```diff
@@ -137,7 +137,7 @@ def generate_target_names_ninja(target: ParsedTargetName, builddir: Path, intros
     else:
         return [str(Path(out_file).relative_to(builddir.resolve())) for out_file in intro_target['filename']]
 
-def get_parsed_args_ninja(options: 'argparse.Namespace', builddir: Path) -> T.List[str]:
+def get_parsed_args_ninja(options: 'argparse.Namespace', builddir: Path) -> T.Tuple[T.List[str], T.Dict[str, str]]:
```
nirbheek, line 140: Looking at it again, this should actually be `T.Tuple[T.List[str], T.Optional[T.Dict[str, str]]]`, because `env` can be `None`. Same for the other uses of `env` below. I am surprised that mypy doesn't warn about this, though. @dcbaker what do you think?

dcbaker: Yeah, this should either return `cmd, {}`, or the Dict needs to be optional. I'm surprised that mypy isn't warning about this too..

dcbaker: If it were my I'd probably just return an empty dict you the calling code doesn't need to handle the None case.

mesonbuild/mcompile.py
```diff
@@ -265,7 +269,7 @@ def get_parsed_args_xcode(options: 'argparse.Namespace', builddir: Path) -> T.Li
         pass
 
     cmd += options.xcode_args
-    return cmd
+    return cmd, None
```
dcbaker, line 272: Same thing here.

mesonbuild/mcompile.py
dcbaker: Looking at the code, this happens to work because the receiving code takes an `Optional[Dict[str, str]]`, so either way I guess, but there will be less code to rework if you just always return `{}` instead of None

nirbheek: The two are not equivalent. In Python, when you set a kwarg to `None` that's considered to be unset (by convention), so `env=None` means "default: inherit process env" while `env={}` means "use an empty env". `env={}` will break everything.

mesonbuild/mcompile.py
nirbheek: https://github.com/mesonbuild/meson/pull/7730#discussion_r490138659

mesonbuild/mcompile.py
nirbheek: I've pushed a commit that fixes this, and also rebased it on top of https://github.com/mesonbuild/meson/pull/7730.

nirbheek:
Failures on 16.04, Arch, Fedora, Opensuse, Bionic:

```
FAILED run_unittests.py::AllPlatformTests::test_clang_tidy - subprocess.Calle
FAILED run_unittests.py::AllPlatformTests::test_clang_format - subprocess.Cal
FAILED run_unittests.py::AllPlatformTests::test_coverage - subprocess.CalledP
FAILED run_unittests.py::AllPlatformTests::test_coverage_html - subprocess.Ca
FAILED run_unittests.py::AllPlatformTests::test_coverage_xml - subprocess.Cal
FAILED run_unittests.py::AllPlatformTests::test_coverage_complex - subprocess
FAILED run_unittests.py::AllPlatformTests::test_coverage_text - subprocess.Ca
```

(will add more when other tests finish)

These are happening because in an effort to get the 32-bit VS build working, I had to convert the tests to use the new meson wrappers.

nirbheek:
> I guess those should indeed be in the list of targets.

I looked into this, and these ninja-only targets really don't fit into `intro-targets.json`. Going to just skip that for now and make it use `--ninja-args`.

nirbheek:
This is such a weird failure. CI always shows one of these two:

```
6>CustomBuild:
           File "D:\a\1\s\test cases\native\27 pipeline\depends\copyrunner.py", line 7, in <module>
             subprocess.check_call([prog, infile, outfile])
           File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\subprocess.py", line 358, in check_call
             retcode = call(*popenargs, **kwargs)
           File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\subprocess.py", line 339, in call
           File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\subprocess.py", line 800, in __init__
             restore_signals, start_new_session)
           File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\subprocess.py", line 1207, in _execute_child
             startupinfo)
         PermissionError: [WinError 32] The process cannot access the file because it is being used by another process
```

```
8>CustomBuild:
           File "D:\a\1\s\test cases\native\27 pipeline\depends\copyrunner.py", line 7, in <module>
             subprocess.check_call([prog, infile, outfile])
           File "C:\hostedtoolcache\windows\Python\3.5.4\x64\lib\subprocess.py", line 266, in check_call
             retcode = call(*popenargs, **kwargs)
           File "C:\hostedtoolcache\windows\Python\3.5.4\x64\lib\subprocess.py", line 247, in call
           File "C:\hostedtoolcache\windows\Python\3.5.4\x64\lib\subprocess.py", line 676, in __init__
             restore_signals, start_new_session)
           File "C:\hostedtoolcache\windows\Python\3.5.4\x64\lib\subprocess.py", line 957, in _execute_child
             startupinfo)
         FileNotFoundError: [WinError 2] The system cannot find the file specified
```

nirbheek:
I removed the CI stuff, will open a new PR for those. Let's get this in, we've verified that it will work if we add 32-bit VS CI.

PR #7733, opened 2020-09-14
WIP: Add support for linking against static Qt install

hdhauk:
Linking against a statically built Qt installation is mostly the same as
the dynamic case, with the exception of plugins (both regular an QML-plugins),
which needs to be manually imported and require several extra linker flags.

Note: This lacks any test for this new functionality, as I assume I'd need a static Qt install to test against. Any pointers to how I can add tests for this?

### DIFF
--- a/docs/markdown/Qt5-module.md
+++ b/docs/markdown/Qt5-module.md
@@ -1,7 +1,7 @@
 # Qt5 module
 
 The Qt5 module provides tools to automatically deal with the various
-tools and steps required for Qt. The module has two methods.
+tools and steps required for Qt. The module has several methods.
 
 ## preprocess
 
@@ -50,6 +50,29 @@ This method takes the following keyword arguments:
 - `method`: method used to find the Qt dependency (`auto` by default).
 
 *Since: 0.54.0*
+
+## is_static
+This method returnes `true` if the Qt install is statically built.
+
+*Since 0.56.0*
+
+## static_plugins_dep
+When linking against a statically built Qt installation, plugins also needs static linking. The `static_plugins_dep` return a dependecy object that holds the nessecary
+linker flags for static linking, and generates files that includes the plugins using Qt's `Q_IMPORT_PLUGIN`. Normal plugins (those found in the `plugins` folder) must
+needs to be specified manually, while QML plugins are automatically detected from QML and JavaScript files.
+
+The method takes the following keyword arguments:
+- `qresources`, rcc source file used for scanning for QML plugins
+- `plugins`, list of plugins to import
+
+```meson
+qt5 = import('qt5')
+if qt5.is_static()
+  plugins_dep = qt5.static_plugins_dep(plugins: ['qsqlite', 'qjpeg'], qresource: 'resources.qrc')
+endif
+```
+*Since 0.56.0*
+
 
 ## Dependencies
 
--- a/mesonbuild/dependencies/ui.py
+++ b/mesonbuild/dependencies/ui.py
@@ -258,7 +258,7 @@ class QtBaseDependency(ExternalDependency):
                     care = out
                 else:
                     care = err
-                return care.split(' ')[-1].replace(')', '')
+                return care.split(' ')[-1].replace(')', '').rstrip()
 
             p = interp_obj.find_program_impl([b], required=False,
                                              version_func=get_version,
@@ -323,11 +323,14 @@ class QtBaseDependency(ExternalDependency):
         # Try to detect moc, uic, rcc
         # Used by self.compilers_detect()
         self.bindir = self.get_pkgconfig_host_bins(core)
+        self.install_prefix = core.get_pkgconfig_variable('exec_prefix', {})
+        self.libdir = core.get_pkgconfig_variable('libdir', {})
         if not self.bindir:
             # If exec_prefix is not defined, the pkg-config file is broken
-            prefix = core.get_pkgconfig_variable('exec_prefix', {})
-            if prefix:
-                self.bindir = os.path.join(prefix, 'bin')
+            if self.install_prefix:
+                self.bindir = os.path.join(self.install_prefix, 'bin')
+        self.qtconfig = core.get_pkgconfig_variable('qt_config', {}).split(' ')
+        self.static = 'static' in self.qtconfig
 
     def search_qmake(self) -> T.Generator['ExternalProgram', None, None]:
         for qmake in ('qmake-' + self.name, 'qmake'):
@@ -378,7 +381,17 @@ class QtBaseDependency(ExternalDependency):
         libdir = qvars['QT_INSTALL_LIBS']
         # Used by self.compilers_detect()
         self.bindir = self.get_qmake_host_bins(qvars)
+        self.libdir = libdir
+        self.install_prefix = qvars['QT_INSTALL_PREFIX']
         self.is_found = True
+
+        qconfig_path = os.path.join(qvars['QT_INSTALL_PREFIX'], 'mkspecs', 'qconfig.pri')
+        with open(qconfig_path) as qconfigfile:
+            line_identifier = 'QT.global.enabled_features = '
+            for line in qconfigfile.readlines():
+                if line.startswith(line_identifier):
+                    self.qconfig = line.replace(line_identifier, '').split(' ')
+                    self.static = 'static' in self.qconfig
 
         # Use the buildtype by default, but look at the b_vscrt option if the
         # compiler supports it.
--- a/mesonbuild/modules/qt.py
+++ b/mesonbuild/modules/qt.py
@@ -12,12 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import json
 import os
+import re
 import shutil
 from .. import mlog
 from .. import build
-from ..mesonlib import MesonException, extract_as_list, File, unholder, version_compare
-from ..dependencies import Dependency, Qt4Dependency, Qt5Dependency, NonExistingExternalProgram
+from ..mesonlib import MesonException, extract_as_list, File, unholder, version_compare, Popen_safe
+from ..dependencies import Dependency, Qt4Dependency, Qt5Dependency, NonExistingExternalProgram, ExternalDependency
 import xml.etree.ElementTree as ET
 from . import ModuleReturnValue, get_include_args, ExtensionModule
 from ..interpreterbase import noPosargs, permittedKwargs, FeatureNew, FeatureNewKwargs
@@ -59,6 +61,15 @@ class QtBaseModule(ExtensionModule):
             self.uic = NonExistingExternalProgram(name='uic' + suffix)
             self.rcc = NonExistingExternalProgram(name='rcc' + suffix)
             self.lrelease = NonExistingExternalProgram(name='lrelease' + suffix)
+
+
+    def _is_static(self, env, method, required=True):
+        kwargs = {'required': required, 'modules': 'Core', 'method': method}
+        qt = _QT_DEPS_LUT[self.qt_version](env, kwargs)
+        if qt.found():
+            return qt.static
+
+        return False
 
     def qrc_nodes(self, state, rcc_file):
         if type(rcc_file) is str:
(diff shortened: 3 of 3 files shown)


stefanha:
@hdhauk Regarding tests, how about including mock .prl files and checking that the code generates the expected QT_IMPORT_PLUGIN output and linker arguments? That way a full Qt static install isn't needed.

docs/markdown/Qt5-module.md
```diff
@@ -51,6 +51,29 @@ This method takes the following keyword arguments:
...
+
+*Since 0.56.0*
+
+## static_plugins_dep
+When linking against a statically built Qt installation, plugins also needs static linking. The `static_plugins_dep` return a dependecy object that holds the nessecary
+linker flags for static linking, and generates files that includes the plugins using Qt's `Q_IMPORT_PLUGIN`. Normal plugins (those found in the `plugins` folder) must
+needs to be specified manually, while QML plugins are automatically detected from QML and JavaScript files.
+
+The method takes the following keyword arguments:
+- `qresources`, rcc source file used for scanning for QML plugins
```
stefanha, line 65: A `qmlfiles` argument is needed for applications that distribute loose QML files instead of bundling them into the binary with the Qt resource compiler.

docs/markdown/Qt5-module.md
```diff
@@ -51,6 +51,29 @@ This method takes the following keyword arguments:
...
+- `plugins`, list of plugins to import
+
+```meson
+qt5 = import('qt5')
+if qt5.is_static()
+  plugins_dep = qt5.static_plugins_dep(plugins: ['qsqlite', 'qjpeg'], qresource: 'resources.qrc')
```
stefanha, line 71: s/qresource/qresources/

docs/markdown/Qt5-module.md
hdhauk: I can certainly add that, although bundling loose QML files would kind of defeat the purpose of linking statically.

stefanha:
I tested with Qt 5.14.2 static built using MinGW-w64. The following backtrace was printed:

```shell
 File "meson/mesonbuild/modules/qt.py", line 352, in static_plugins_dep
    link_args.extend(self._link_args(qt, prl))
  File "meson/mesonbuild/modules/qt.py", line 288, in _link_args
    with open(prl) as prl:
FileNotFoundError: [Errno 2] No such file or directory: 'mxe/usr/x86_64-w64-mingw32.static/qt5/qml/QtQuick/Layouts/libqquicklayoutsplugin.prl'
$ ls mxe/usr/x86_64-w64-mingw32.static/qt5/qml/QtQuick/Layouts/
libqquicklayoutsplugin.a  plugins.qmltypes          qmldir                    qquicklayoutsplugin.prl
```

`lib` is being prepended to the .prl filename. This may be related to Windows cross-compilation but I haven't had time to investigate yet. I'll post more info when I'm able.

stefanha:
On Wed, Sep 16, 2020 at 9:25 AM Halvor Haukvik <notifications@github.com>
wrote:

> *@hdhauk* commented on this pull request.
> ------------------------------
>
> In docs/markdown/Qt5-module.md
> <https://github.com/mesonbuild/meson/pull/7733#discussion_r489255720>:
>
> > @@ -51,6 +51,29 @@ This method takes the following keyword arguments:
>
>  *Since: 0.54.0*
>
> +## is_static
> +This method returnes `true` if the Qt install is statically built.
> +
> +*Since 0.56.0*
> +
> +## static_plugins_dep
> +When linking against a statically built Qt installation, plugins also needs static linking. The `static_plugins_dep` return a dependecy object that holds the nessecary
> +linker flags for static linking, and generates files that includes the plugins using Qt's `Q_IMPORT_PLUGIN`. Normal plugins (those found in the `plugins` folder) must
> +needs to be specified manually, while QML plugins are automatically detected from QML and JavaScript files.
> +
> +The method takes the following keyword arguments:
> +- `qresources`, rcc source file used for scanning for QML plugins
>
> I can certainly add that, although bundling loose QML files would kind of
> defeat the purpose of linking statically.
>
Exposing QML files makes it easy for users to edit them, which is desirable
for some applications.

Using a static Qt build is still useful because it avoids the need to
manage lots of DLL files, so I don't see static Qt and loose QML files as
mutually exclusive.

hdhauk:
> I tested with Qt 5.14.2 static built using MinGW-w64. The following backtrace was printed:
> 
> ```shell
>  File "meson/mesonbuild/modules/qt.py", line 352, in static_plugins_dep
>     link_args.extend(self._link_args(qt, prl))
>   File "meson/mesonbuild/modules/qt.py", line 288, in _link_args
>     with open(prl) as prl:
> FileNotFoundError: [Errno 2] No such file or directory: 'mxe/usr/x86_64-w64-mingw32.static/qt5/qml/QtQuick/Layouts/libqquicklayoutsplugin.prl'
> $ ls mxe/usr/x86_64-w64-mingw32.static/qt5/qml/QtQuick/Layouts/
> libqquicklayoutsplugin.a  plugins.qmltypes          qmldir                    qquicklayoutsplugin.prl
> ```
> 
> `lib` is being prepended to the .prl filename. This may be related to Windows cross-compilation but I haven't had time to investigate yet. I'll post more info when I'm able.

You're right, the I guess the `lib`-prefix is only used for unix.

hdhauk:
> On Wed, Sep 16, 2020 at 9:25 AM Halvor Haukvik ***@***.***> wrote: ***@***.**** commented on this pull request. ------------------------------ In docs/markdown/Qt5-module.md <[#7733 (comment)](https://github.com/mesonbuild/meson/pull/7733#discussion_r489255720)>: > @@ -51,6 +51,29 @@ This method takes the following keyword arguments: *Since: 0.54.0* +## is_static +This method returnes `true` if the Qt install is statically built. + +*Since 0.56.0* + +## static_plugins_dep +When linking against a statically built Qt installation, plugins also needs static linking. The `static_plugins_dep` return a dependecy object that holds the nessecary +linker flags for static linking, and generates files that includes the plugins using Qt's `Q_IMPORT_PLUGIN`. Normal plugins (those found in the `plugins` folder) must +needs to be specified manually, while QML plugins are automatically detected from QML and JavaScript files. + +The method takes the following keyword arguments: +- `qresources`, rcc source file used for scanning for QML plugins I can certainly add that, although bundling loose QML files would kind of defeat the purpose of linking statically.
> Exposing QML files makes it easy for users to edit them, which is desirable for some applications. Using a static Qt build is still useful because it avoids the need to manage lots of DLL files, so I don't see static Qt and loose QML files as mutually exclusive.

Gotcha! I'll add an `qmlfiles` kwarg for that.

hdhauk:
@stefanha I've addressed some of your points. Was the simple test I've added along the lines you were suggesting?

stefanha:
@hdhauk Yes, that way the prl file parsing is covered by a unit test.

stefanha:
@hdhauk Your latest branch (b7e08c1acd00db59093a8b077c41942083542342) cross-compiles successfully with MinGW-w64. I tested both qresources and qmlfiles.

Here is a meson.build snippet:
```meson
if qt5.is_static()
  plugins_dep = qt5.static_plugins_dep(qresources : 'resources.qrc')
else
  plugins_dep = dependency('', required : false)
endif
```

I wonder if there is a more elegant way of expressing this. For example, qt5.static_plugins_dep() could check is_static() internally and return dependency('', required : false) when Qt is dynamically linked. That way the user's meson.build is simpler:

```meson
plugins_dep = qt5.static_plugins_dep(qresources : 'resources.qrc')
```

What do you think?

docs/markdown/Qt5-module.md
```diff
@@ -63,6 +63,7 @@ needs to be specified manually, while QML plugins are automatically detected fro
 
 The method takes the following keyword arguments:
 - `qresources`, rcc source file used for scanning for QML plugins
+- `qmlfiles`, sources files used for scanning for QML plugins
```
stefanha, line 66: It's slightly clearer to document that these are QML files rather than just "source files":

```diff
+ - `qmlfiles`, QML source files to scan for QML plugins
```

hdhauk:
> @hdhauk Your latest branch ([b7e08c1](https://github.com/mesonbuild/meson/commit/b7e08c1acd00db59093a8b077c41942083542342)) cross-compiles successfully with MinGW-w64. I tested both qresources and qmlfiles.
> 
> Here is a meson.build snippet:
> 
> ```meson
> if qt5.is_static()
>   plugins_dep = qt5.static_plugins_dep(qresources : 'resources.qrc')
> else
>   plugins_dep = dependency('', required : false)
> endif
> ```
> 
> I wonder if there is a more elegant way of expressing this. For example, qt5.static_plugins_dep() could check is_static() internally and return dependency('', required : false) when Qt is dynamically linked. That way the user's meson.build is simpler:
> 
> ```meson
> plugins_dep = qt5.static_plugins_dep(qresources : 'resources.qrc')
> ```
> 
> What do you think?

Yeah, it should probably internally check if qt is static anyway. I have gone back and forth a lot on what should be the syntax for this, but I think your suggestion probably is the cleanest so far. Given that there is 'static' in the method name, it shouldn't be too surprising for people that it doesn't do anything when Qt is dynamically built.

wahjam (approved):
Thank you, this is great.

I have tested MinGW-w64 cross-compilation.

stefanha (approved):

stefanha:
> Thank you, this is great.
> 
> I have tested MinGW-w64 cross-compilation.

Oops, that was me. I was logged into the wrong GitHub account ;).

stefanha:
@hdhauk A related issue: when I build without explicitly specifying the platform plugin (e.g. "qwindows") the resulting executable fails during Qt initialization. Have you seen this too?

A qmake build automatically includes the platform plugin and it would be nice to do this in the meson qt5 module too, but I think it's a separate issue and shouldn't hold up this pull request.

hdhauk:
> @hdhauk A related issue: when I build without explicitly specifying the platform plugin (e.g. "qwindows") the resulting executable fails during Qt initialization. Have you seen this too?
> 
> A qmake build automatically includes the platform plugin and it would be nice to do this in the meson qt5 module too, but I think it's a separate issue and shouldn't hold up this pull request.

Hmm, yeah that might be useful. I haven't really used qmake much, mostly Qbs -- and in Qbs I believe you have to add stuff like qwindows yourself.

stefanha:
@jpakkane I have tested Halvor's pull request. Does the code look good to you?

mesonbuild/modules/qt.py
```diff
@@ -263,3 +274,118 @@ def compile_translations(self, state, args, kwargs):
...
+        return [x.fname for x in nodes if os.path.splitext(x.fname)[1] in ['.qml', '.js']]
+
+
+    def _find_prl(self, dir, name):
+        templates = ['{}.prl', 'lib{}.prl']
+        for template in templates:
+            path = os.path.join(dir, template.format(name))
+            if os.path.exists(path):
+                return path
+        raise MesonException('failed to find .prl file for plugin \"{}\"'.format(name))
```
mensinda, line 332: Escaping `"` shouldn't be required here, because you are using `'` type stings.

mensinda:
I don't know enough about QT (and even less about static QT) to comment on whether this code is correct, however, it does look good to me from a general python and meson point of view.

There are, however, still some general points that have to be addressed:

- This is new functionality, so it needs a release note snippet.
- The CI must be green (maybe try a rebase if the errors are unrelated).
- Is this code actually tested in our CI setup / is there a simple way to install a static qt setup?

hdhauk:
> I don't know enough about QT (and even less about static QT) to comment on whether this code is correct, however, it does look good to me from a general python and meson point of view.
> 
> There are, however, still some general points that have to be addressed:
> 
>     * This is new functionality, so it needs a release note snippet.
> 
>     * The CI must be green (maybe try a rebase if the errors are unrelated).
> 
>     * Is this code actually tested in our CI setup / is there a simple way to install a static qt setup?

I'll have look later and attempt to address your two first points. The 3rd I'm unsure about. Qt doesn't release statically built binaries, and they tend to be at least a couple of hundred megabytes big. The plugins are also kind of annoying to check if you get right, as missing plugins (if you forget to use the `Q_IMPORT_PLUGIN`) give runtime errors instead of compile-time ones.

mensinda:
The problem is that there should be some way to test this code in the CI setup, so we don't accidentally break it further down the line. Ideally, we could do this with some (or at least one) of our existing CI docker images (see [the doc](https://mesonbuild.com/MesonCI.html) for a very rough overview). Adding a new CI image just for static QT would also work.

You can ask me if you need help with the CI/docker image setup. If you don't have the time to setup/fix a docker image we need, at the very least, some sort of instructions on how to set up such a static QT environment so we can check that it works locally or add a new docker image down the line later on.

jpakkane:
Do you need to have a prl file for this to work? I'd really want to have this work without external magic qmake files, especially given that Qt is going to CMake and Qmake will be phased out eventually.

hdhauk:
@jpakkane As far as I can tell, the bundled cmake-files also reads the config from the prl files.There even is a line with the tag `QMAKE_PRL_LIBS_FOR_CMAKE` in all the prls. As you pointed out, this might of course be subject to change since Qt are phasing out qmake, but I don't know of any other way of obtaining the right flags.

jpakkane:
The test failure seems relevant:

```
  File "/__w/meson/meson/mesonbuild/dependencies/ui.py", line 215, in __init__
    self.from_text = self._qmake_detect(mods, kwargs)
  File "/__w/meson/meson/mesonbuild/dependencies/ui.py", line 389, in _qmake_detect
    with open(qconfig_path) as qconfigfile:
FileNotFoundError: [Errno 2] No such file or directory: '/usr/mkspecs/qconfig.pri'
 ```

**Comment by ntonnaett** [comment]
What is the state of this? Could Meson get the information of the prl files in a different way? For me it looks more like the prl logic gets transferred to cmake someday (or pkgconf?).

File: mesonbuild/dependencies/ui.py [hunk 5]
```diff
@@ -378,8 +381,18 @@ def _qmake_detect(self, mods, kwargs):
         libdir = qvars['QT_INSTALL_LIBS']
         # Used by self.compilers_detect()
         self.bindir = self.get_qmake_host_bins(qvars)
+        self.libdir = libdir
+        self.install_prefix = qvars['QT_INSTALL_PREFIX']
         self.is_found = True
 
+        qconfig_path = os.path.join(qvars['QT_INSTALL_PREFIX'], 'mkspecs', 'qconfig.pri')
```
ntonnaett, line 388:
```suggestion
        qconfig_path = os.path.join(qvars['QT_INSTALL_ARCHDATA'], 'mkspecs', 'qconfig.pri')
```
To my understanding https://doc.qt.io/qt-6/qmake-environment-reference.html says that mkspecs is in the data dir and because it's architecture-dependent QT_INSTALL_ARCHDATA should be used.

PR #7734, opened 2020-09-14
Test meson wrappers on the CI, and test 32-bit VS

nirbheek:
```
project tests: Test that the wrappers actually work

```

```
ci: Convert the vc2017 x64 job to x86

So we test the VS backend on 32-bit. Would've caught, f.ex.,
```
https://github.com/mesonbuild/meson/pull/7730


nirbheek:
Merged into https://github.com/mesonbuild/meson/pull/7730.

PR #7735, opened 2020-09-14
machinefiles: Allow keys to be stored case insensitive

### LINKED ISSUE #7731: options in cross/native files cannot store capitals
`ConfigParser.configparser` by default stores all keys in sections as lower case strings. We need to change that.

dcbaker:
This is required to make the various keys in the [user options] section
work the same as they do in the meson_options.txt file, where we don't
have any rules about case sensitivity.

There is some risk here. Someone may be relying on this lower by default
behavior, and this could break their machine files.

Fixes #7731

### DIFF
--- a/docs/markdown/Machine-files.md
+++ b/docs/markdown/Machine-files.md
@@ -5,6 +5,9 @@ documentation on the common values used by both, for the specific values of
 one or the other see the [cross compilation](Cross-compilation.md) and [native
 environments](Native-environments.md).
 
+*Changed in 0.56.0* Keys within sections are now case sensitive. This is
+*required to make project options work correctly.
+
 ## Data Types
 
 There are four basic data types in a machine file:
--- a/docs/markdown/snippets/machine-files-case-insensitve.md
+++ b/docs/markdown/snippets/machine-files-case-insensitve.md
@@ -1,0 +1,6 @@
+## Machine file keys are stored case sensitive
+
+Previous the keys were always lowered, which worked fine for the values that
+were allowed in the machine files. With the addition of per-project options
+we need to make these sensitive to case, as the options in meson_options.txt
+are sensitive to case already.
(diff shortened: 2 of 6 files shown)


docs/markdown/Machine-files.md
```diff
@@ -5,6 +5,9 @@ documentation on the common values used by both, for the specific values of
 one or the other see the [cross compilation](Cross-compilation.md) and [native
 environments](Native-environments.md).
 
+*Changed in 0.56.0* Keys within sections are now case sensitive. This is
+*required to make project options work correctly.
```
xclaesse, line 9: The extra `*` is a typo I guess?

docs/markdown/snippets/machine-files-case-insensitve.md
```diff
@@ -0,0 +1,6 @@
+## Machine file keys are stored case sensitive
+
+Previous the keys were always lowered, which worked fine for the values that
```
xclaesse, line 3: *Previously

mesonbuild/coredata.py
```diff
@@ -848,13 +848,17 @@ def emit_base_options_warnings(self, enabled_opts: list):
...
-    def __init__(self):
+    def __init__(self) -> None:
         # We don't want ':' as key delimiter, otherwise it would break when
         # storing subproject options like "subproject:option=value"
         super().__init__(delimiters=['='], interpolation=None)
 
+    def optionxform(self, option: str) -> str:
+        # Don't call str.lower() on keys
+        return option
+
```
xclaesse, line 859: Strictly speaking making keys case sensitive is not backward compatible, but I don't think it's a big deal in this case. However, I'm wondering if we should do the opposite and make optinterpreter case insensitive. I don't think it is desired to have 2 identical options that differ only by case. We already discussed to also normalize - to _ there.

dcbaker: I thought about that too, and didn't know which was better, since that was also a backwards incompatible change. It would also mean making get_option case insensitive.

mensinda: There are cases where you want case sensitive variables in machine files. Specifically, for specifying (case sensitive) CMake variables in the machine files that are then forwarded to CMake. This is required / very annoying to work around otherwise.

xclaesse: ok, LGTM then, we can always change options to be case insensitive later.

PR #7736, opened 2020-09-14
options: Handle updates to choices in options

### LINKED ISSUE #7386: New choice in array option is ignored
**Describe the bug**
When a new choice "newchoice" is added to an existing array option "foo", _meson configure_ will not list the new choice in the "Possible Values" column, and _meson -Dfoo=newchoice_ complains that "newchoice" is not valid.

`ERROR: Options "newchoice" are not in allowed choices: "oldchoice"`

**To Reproduce**
[choice.tar.gz](https://github.com/mesonbuild/meson/files/4839477/choice.tar.gz)

1. Run _meson build_
2. Edit meson_options.txt as follows:
```
-        choices : ['oldchoice'],
+        choices : ['oldchoice', 'newchoice'],
```
3. Run _meson configure build_. The "foo" option should report "newchoice" as a Possible Value.

**Expected behavior**
The new choice should be considered by meson as a possible value and accepted when modifying the option with -D.

**system parameters**
* Operating system: Gentoo Linux
* Python version: 3.7
* Meson version: 0.52.1 (tested the master branch as well)

dcbaker:
Currently if you change the `choices` field in the meson_options.txt
file, no update will be done until `meson setup --wipe` is called. Now
if the choices change then the options will be properly merged.

If the currently select value is still valid it is guaranteed to be
kept, if it is now invalid the new default value will be used and a
warning will be printed.

Fixes #7386

### DIFF
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -689,14 +689,24 @@ class CoreData:
     def get_external_link_args(self, for_machine: MachineChoice, lang):
         return self.compiler_options[for_machine][lang]['link_args'].value
 
-    def merge_user_options(self, options: T.Dict[str, T.Union[str, bool, int]]) -> None:
+    def merge_user_options(self, options: T.Dict[str, UserOption[T.Any]]) -> None:
         for (name, value) in options.items():
             if name not in self.user_options:
                 self.user_options[name] = value
-            else:
-                oldval = self.user_options[name]
-                if type(oldval) != type(value):
-                    self.user_options[name] = value
+                continue
+
+            oldval = self.user_options[name]
+            if type(oldval) != type(value):
+                self.user_options[name] = value
+            elif oldval.choices != value.choices:
+                # If the choices have changed, use the new value, but attempt
+                # to keep the old options. If they are not valid keep the new
+                # defaults but warn.
+                self.user_options[name] = value
+                try:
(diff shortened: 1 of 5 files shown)


mensinda (approved):
LGTM (once the CI is green)

dcbaker:
I think all of the macos failures are unrelated, but I can't kick travis jobs, so I have no idea.

jpakkane:
Interestingly, neither have I. There used to be a rerun button but no more. The failures that I see are caused by clock skew and the other Mac job was successful.

mensinda:
This PR seems to have broken `test_options_with_choices_changing` on OSX (see https://travis-ci.org/github/mesonbuild/meson/jobs/733093209)

mensinda:
After looking at the PR again, I should say that the `test_options_with_choices_changing` from this PR didn't work to begin with on OSX.

dcbaker:
Gah. The macos ci is so unreliable I didn't see this. Of course, I have no idea why this would work in windows and Linux but not macos. I'm going to guess a test bug

jpakkane:
Should this be reverted then?

dcbaker:
I think it's a bug in the test. I just finished getting my Mac working again Friday, and I'll get a fix figured out tomorrow. I'll open a fresh issue.

dcbaker:
See #7835

PR #7737, opened 2020-09-14
Submit/pylint add a few more warnings

dcbaker:
The only thing I've actually caught is a couple of `if len(container)`, but the rest of these are pretty useful as well.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -7,5 +7,9 @@ score=no
 [MESSAGES CONTROL]
 disable=all
 enable=
+    assert-on-tuple,
+    dangerous-default-value,
+    deprecated-lambda,
+    compare-to-zero,
     len-as-condition,
     unreachable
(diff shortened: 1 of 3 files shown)


.pylintrc
```diff
@@ -6,4 +6,10 @@ score=no
 
 [MESSAGES CONTROL]
 disable=all
-enable=unreachable
\ No newline at end of file
+enable=
+    assert-on-tuple,
```
mensinda, line 10: Does this even work?

```bash
find mesonbuild -name "*.py" -type f -exec grep 'assert(' {} \;
```

prints a ton of results.

dcbaker: You probably need to feel for `assert(.*,` to find what it's actually searching for

mensinda: Oh, right. I was thinking of the `superfluous-paren` warnings, nevermind.

nirbheek:
The MSYS2 failures are fixed. A rebase should do it.

PR #7738, opened 2020-09-14
CI: don't update MSYS2

lazka:
This should use the package database from the installer,
from the release choco provides and result in fewer potential
changes overall.


jpakkane:
LGMT. If CI is green I'll merge this unless someone has issues.

lazka:
for future reference, the underlying issue motivating this change has been fixed -> https://github.com/msys2/MINGW-packages/pull/7032

PR #7740, opened 2020-09-14
Allow blocking/forcing automatic subproject search

bonzini:
Sometimes, distros want to configure a project so that it does not use any bundled library.  In this case, `meson.build` might want
to do something like this, where slirp is a combo option with values auto/system/internal:
    
      slirp = dependency('', required: false)
      if get_option('slirp') != 'internal'
        slirp = dependency('slirp',
                           required: get_option('slirp') == 'system')
      endif
      if not slirp.found()
        slirp = subproject('libslirp', ...) .variable('...')
      endif
    
Note that we cannot use "fallback" because the "system" value should never look for a subproject.

This worked until 0.54.x, but in 0.55.x this breaks because of the automatic subproject search.  In fact the desired effect here is backwards compared to the policy of doing an automatic search on `required: true`; instead we only want to do the search if `required` is false!

It would be possible to look for the dependency with  `required: false` and issue the error manually, but it's ugly and it may produce an error message that looks "different" from Meson's.

Instead, with this change it is possible to achieve this effect in an even simpler way:
    
      slirp = dependency('slirp',
                         required: get_option('slirp') != 'auto',
                         fallback: get_option('slirp') == 'system' ? false : ['slirp', 'libslirp_dep'])

### DIFF
--- a/docs/markdown/Reference-manual.md
+++ b/docs/markdown/Reference-manual.md
@@ -454,15 +454,15 @@ Dependencies can also be resolved in two other ways:
 * by a fallback subproject which, if needed, will be brought into the current
   build specification as if [`subproject()`](#subproject) had been called.
   The subproject can be specified with the `fallback` argument.  Alternatively,
-  if the `fallback` argument is absent and `required` is `true` or
-  [`enabled`](Build-options.md#features), *since 0.55.0* Meson will
+  if the `fallback` argument is absent, *since 0.55.0* Meson can
   automatically identify a subproject as a fallback if a wrap file
   [provides](Wrap-dependency-system-manual.md#provide-section) the
   dependency, or if a subproject has the same name as the dependency.
   In the latter case, the subproject must use `meson.override_dependency` to
   specify the replacement, or Meson will report a hard error.  See the
   [Wrap documentation](Wrap-dependency-system-manual.md#provide-section)
-  for more details.
+  for more details.  This automatic search can be controlled using the
+  `allow_fallback` keyword argument.
 
 This function supports the following keyword arguments:
 
@@ -471,7 +471,16 @@ This function supports the following keyword arguments:
   (like `default_options` in [`project()`](#project), they only have
   effect when Meson is run for the first time, and command line
   arguments override any default options in build files)
-- `fallback`: manually specifies a subproject
+- `allow_fallback` (boolean argument, *since 0.56.0*): specifies whether Meson
+  should automatically pick a fallback subproject in case the dependency
+  is not found in the system.  If `true` and the dependency is not found
+  on the system, Meson will fallback to a subproject that provides this
+  dependency. If `false`, Meson will not fallback even if a subproject
+  provides this dependency.  By default, Meson will do so if `required`
+  is `true` or  [`enabled`](Build-options.md#features); see the [Wrap
+  documentation](Wrap-dependency-system-manual.md#provide-section)
+  for more details.
+- `fallback` (string or array argument): manually specifies a subproject
   fallback to use in case the dependency is not found in the system.
   This is useful if the automatic search is not applicable or if you
   want to support versions of Meson older than 0.55.0.  If the value is an
--- a/docs/markdown/Reference-manual.md
+++ b/docs/markdown/Reference-manual.md
@@ -440,33 +440,52 @@ system) with the given name with `pkg-config` and [with
 CMake](Dependencies.md#cmake) if `pkg-config` fails. Additionally,
 frameworks (OSX only) and [library-specific fallback detection
 logic](Dependencies.md#dependencies-with-custom-lookup-functionality)
-are also supported. This function supports the following keyword
-arguments:
+are also supported.
+
+Dependencies can also be resolved in two other ways:
+
+* if the same name was used in a `meson.override_dependency` prior to
+  the call to `dependency`, the overriding dependency will be returned
+  unconditionally; that is, the overriding dependency will be used
+  independent of whether an external dependency is installed in the system.
+  Typically, `meson.override_dependency` will have been used by a
+  subproject.
+
+* by a fallback subproject which, if needed, will be brought into the current
+  build specification as if [`subproject()`](#subproject) had been called.
+  The subproject can be specified with the `fallback` argument.  Alternatively,
+  if the `fallback` argument is absent and `required` is `true` or
+  [`enabled`](Build-options.md#features), *since 0.55.0* Meson will
+  automatically identify a subproject as a fallback if a wrap file
+  [provides](Wrap-dependency-system-manual.md#provide-section) the
+  dependency, or if a subproject has the same name as the dependency.
+  In the latter case, the subproject must use `meson.override_dependency` to
+  specify the replacement, or Meson will report a hard error.  See the
+  [Wrap documentation](Wrap-dependency-system-manual.md#provide-section)
+  for more details.
+
+This function supports the following keyword arguments:
 
 - `default_options` *(since 0.37.0)*: an array of default option values
   that override those set in the subproject's `meson_options.txt`
   (like `default_options` in [`project()`](#project), they only have
   effect when Meson is run for the first time, and command line
   arguments override any default options in build files)
-- `fallback`: specifies a subproject fallback to use in case the
-  dependency is not found in the system. The value is an array
-  `['subproj_name', 'subproj_dep']` where the first value is the name
+- `fallback`: manually specifies a subproject
+  fallback to use in case the dependency is not found in the system.
+  This is useful if the automatic search is not applicable or if you
+  want to support versions of Meson older than 0.55.0.  If the value is an
+  array `['subproj_name', 'subproj_dep']`, the first value is the name
   of the subproject and the second is the variable name in that
   subproject that contains a dependency object such as the return
   value of [`declare_dependency`](#declare_dependency) or
   [`dependency()`](#dependency), etc. Note that this means the
   fallback dependency may be a not-found dependency, in which
   case the value of the `required:` kwarg will be obeyed.
-  *(since 0.54.0)* `'subproj_dep'` argument can be omitted in the case the
-  subproject used `meson.override_dependency('dependency_name', subproj_dep)`.
-  In that case, the `fallback` keyword argument can be a single string instead
-  of a list of 2 strings. *Since 0.55.0* the `fallback` keyword argument can be
-  omitted when there is a wrap file or a directory with the same `dependency_name`,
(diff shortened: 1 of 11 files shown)


xclaesse:
You should use a `feature` option instead, they have auto/enabled/disabled values, and use it like that: `slirp = dependency('slirp', required: get_option('slirp'))`. Distros will usually build with `-Dauto_features=enabled --wrap-mode=nofallback` which will have the same effect as setting it to system in your case.

bonzini:
A feature option is not appropriate if "disabled" is never a valid value. I guess I could use an auto/enabled combo and `required: get_option('slirp') == 'enabled'` but this doesn't do what I need, because...

> Distros will usually build with -Dauto_features=enabled --wrap-mode=nofallback which will have the same effect as setting it to system in your case

... the key word is usually. For example QEMU in RHEL uses _some_ bundled libraries but not all because there are no other users in the whole distro. Besides, that requires using wrap, which some parent projects may not desire (because they don't want to rely on upstream git repositories, for example); or `override_dependency`, which some subprojects may not desire to remain compatible with old versions of Meson.

Also this was changed in 0.55.0 without any deprecation period, and it generally breaks all uses of `required: true` if the subproject is found automatically but it doesn't use either wrap or `override_dependency`.

xclaesse:
I'm surprised rhel allows subprojects when building their package, it means it fetch the dependency from git during package build? I was sure all distros had strict requirement for building packages offline.

xclaesse:
Anyway... I understand your case now, and I'm sorry to have broken that in 0.55, I would have never thought someone had that kind of requirement.

I actually like having fallback being a boolean, I already wanted the opposite use-case, fallback: true, to use the automatic fallback when required=false. Since fallback information are now in wrap file [provide] section, we could deprecated any non-boolean value.

bonzini:
Subprojects in this case are (will be) git submodules and l part of the tarball.

I certainly can add support for `fallback: true,`, that's a great idea.

bonzini:
Done. Tests should be fairly comprehensive, let's see how the CI goes.

bonzini:
Regarding the CI failure, it could be just an overzealous unit test but still: is it intended that the Travis .6 job (which is a macOS one) doesn't have pkg-config installed? I don't see anything in the configuration that would justify that difference between the two macOS jobs.

bonzini:
Regarding the CI failure, it could be just an overzealous unit test but still: is it intended that the Travis .6 job (which is a macOS one) doesn't have pkg-config installed? I don't see anything in the configuration that would justify that difference between the two macOS jobs.

docs/markdown/Reference-manual.md
```diff
@@ -436,33 +436,46 @@ system) with the given name with `pkg-config` and [with
...
-  `['subproj_name', 'subproj_dep']` where the first value is the name
+- `fallback` (boolean argument, *since 0.56.0*): specifies whether Meson
+  should automatically pick a fallback subproject in case the dependency
+  is not found in the system.  By default, Meson will do so if `required` is
+  `true` or  [`enabled`](Build-options.md#features), or if a same-named
+  subproject has already been added with [`subproject()`](#subproject).
+- `fallback` (string or array argument): manually specifies a subproject
+  fallback to use in case the dependency is not found in the system.
+  This is useful if the automatic search is not applicable or if you
+  want to support older versions of Meson.  If the value is an array
```
xclaesse, line 467: Instead of vague 'older' it should say 'older than 0.55.0'.

docs/markdown/Reference-manual.md
```diff
@@ -436,33 +436,46 @@ system) with the given name with `pkg-config` and [with
 CMake](Dependencies.md#cmake) if `pkg-config` fails. Additionally,
 frameworks (OSX only) and [library-specific fallback detection
 logic](Dependencies.md#dependencies-with-custom-lookup-functionality)
-are also supported. This function supports the following keyword
-arguments:
+are also supported.
+
+Dependencies can also be resolved by a fallback subproject.
+Versions of Meson before 0.55.0 will never automatically search for
+a fallback.  *Since 0.55.0* Meson can look for a wrap file or a directory
```
xclaesse, line 443: IMHO it's not clear what 'automatically search' means here.

Maybe instead of repeating the mechanism here, it could just link to the doc we already have:
```
Dependencies can also be resolved by a fallback subproject. See
[Wrap documentation](Wrap-dependency-system-manual.md#provide-section) and `fallback` keyword argument.
```

docs/markdown/Reference-manual.md
```diff
@@ -436,33 +436,46 @@ system) with the given name with `pkg-config` and [with
...
   (like `default_options` in [`project()`](#project), they only have
   effect when Meson is run for the first time, and command line
   arguments override any default options in build files)
-- `fallback`: specifies a subproject fallback to use in case the
-  dependency is not found in the system. The value is an array
-  `['subproj_name', 'subproj_dep']` where the first value is the name
+- `fallback` (boolean argument, *since 0.56.0*): specifies whether Meson
+  should automatically pick a fallback subproject in case the dependency
+  is not found in the system.  By default, Meson will do so if `required` is
+  `true` or  [`enabled`](Build-options.md#features), or if a same-named
```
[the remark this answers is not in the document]
xclaesse: I think this part of the doc could be more explicit of what it does if true and if false.

What about:
```
If `true` and the dependency is not found on the system, Meson will fallback to a subproject that provides this dependency, even if the dependency is optional. If `false` Meson will not fallback even if a subproject provides this dependency. See [Wrap documentation](Wrap-dependency-system-manual.md#provide-section).
```

docs/markdown/Reference-manual.md
xclaesse: We could even go a step further and deprecate this behavior.

docs/markdown/snippets/fallback_bool.md
```diff
@@ -0,0 +1,5 @@
+## `dependency` accepts a boolean argument
```
xclaesse, line 1: missing 'fallback' somewhere in that title.

docs/markdown/snippets/fallback_bool.md
```diff
@@ -0,0 +1,5 @@
...
+
+Passing `fallback: true` to `dependency` enables automatic fallback to
+a subproject with the same name as the dependency, while `fallback: false`
+blocks automatic search for a subproject that can fulfill the dependency.
```
xclaesse, line 5: It's not clear what 'automatic' means here (to people not aware of this feature) and it's not only about 'same name' but more generally a subproject that _provides_ a dependency.

I would simply copy/paste the text suggested in the doc above:
```
`fallback` keyword argument now accept boolean value. If `true` and the dependency is not found on the system, Meson will fallback to a subproject that provides this dependency, even if the dependency is optional. If `false` Meson will not fallback even if a subproject provides this dependency. See [Wrap documentation](Wrap-dependency-system-manual.md#provide-section).
```

docs/markdown/Reference-manual.md
xclaesse: btw, that wrap documentation page explains why optional dependencies does not use automatic fallback. Would make sense to add a note there that it can now be forced by setting `fallback: true`.

xclaesse (commented):
Mostly nitpicking comments about doc. The implementation is a bit hairy but does the right thing and I don't have ideas to make is nicer, so let's go with that.

docs/markdown/Reference-manual.md
xclaesse: Actually the wrap doc mention the trick of setting `fallback: 'foo'` which is now equivalent of setting `fallback: true`.

docs/markdown/Reference-manual.md
bonzini: Deprecation of the string argument is probably a good idea. The array argument needs to stay in order to support older versions of Meson.

docs/markdown/Reference-manual.md
bonzini: On one hand wrap files and subprojects are somewhat related, on the other hand I think the reference manual should be self contained. I adjusted the documentation according to your other suggestions though.

nirbheek:
> is it intended that the Travis .6 job (which is a macOS one) doesn't have pkg-config installed? I don't see anything in the configuration that would justify that difference between the two macOS jobs.

Yes it is on purpose because macOS does not ship with pkg-config installed (but it's available via Homebrew) so we need to test both configurations.

bonzini:
Ping?

docs/markdown/Reference-manual.md
```diff
@@ -436,33 +436,51 @@ system) with the given name with `pkg-config` and [with
...
+are also supported.
+
+Dependencies can also be resolved by a fallback subproject; if needed,
+the subproject will be brought into the current build specification as if
+[`subproject()`](#subproject) had been called.  *Since 0.55.0* Meson will
+automatically identify a subproject as a fallback if a wrap file
+[provides](Wrap-dependency-system-manual.md#provide-section)
+the dependency, or if a subproject has the same name as the dependency.
+In the latter case, the subproject must use
+`meson.override_dependency('dependency_name', subproj_dep) to
```
jpakkane, line 448: A backtick is missing here.

jpakkane:
What happens if someone has done an override_dependency and then someone calls `dependency(..., fallback: false)`? Should it error out? Not use the dep? Use the dep? At the very least this behaviour must be documented.

jpakkane:
The default way to handle these things in Meson is to detect them and make it a hard error. It can be made to do something else later if people have a need for it.

bonzini:
`override_dependency` always uses the overriding dependency. `fallback` only checks whether a subproject is entered automatically, it has no effect if a subproject was entered before through either `subproject` or another dependency.

Note that, even without this PR, a `dependency('bar', fallback: [ 'bar', 'bar_dep'])` is ignored if a subproject `foo` has already defined an overriding dependency for `bar`.

jpakkane:
I'm not a huge fan of having bool and string types for the kwarg. However if the intent is to gradually phase out forms other than boolean, this might work for the transition period.

jpakkane:
Or maybe it should be its own kwarg? `allow_fallback`?

bonzini:
The string one should definitely be phased out, the array one might remain.

jpakkane:
In that case it should go in its own kwarg. It is more explicit and readable. In addition this improves backwards (or forwards?) compatibility. Any project that would use a boolean would no longer be buildable with an older Meson even if it otherwise would work. An extra kwarg will give a warning but it will work.

With that fixed and the documentation updated to explicitly say that if someone has done an override for the dependency name, it will be used, even if the value of `allow_fallback` is `false`. With those changes done, LGTM.

PR #7741, opened 2020-09-14
Add type annotations to the linkers module

dcbaker:
There isn't too much to do here, but mypy did notice a bug in the code. So there's that.


mensinda (approved):
LGTM

PR #7743, opened 2020-09-15
envconfig: fix return type of get_env_var_pair

dcbaker:
It doesn't return (None, None), it returns just None.


Ericson2314 (approved):

Ericson2314:
This is very trivial, and entirely fixing my mistakes. Thanks!

PR #7752, opened 2020-09-17
Support identifying msvc dependency prefix with a path starting with a backslash

mstorsjo:
With certain cross compilation wrappings of MSVC, the paths use backslashes, but without the leading drive name, so allow the path to start with any path separator.

### DIFF
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -464,8 +464,13 @@ int dummy;
         # different locales have different messages with a different
         # number of colons. Match up to the the drive name 'd:\'.
         # When used in cross compilation, the path separator is a
-        # forward slash rather than a backslash so handle both.
-        matchre = re.compile(rb"^(.*\s)([a-zA-Z]:\\|\/).*stdio.h$")
+        # forward slash rather than a backslash so handle both; i.e.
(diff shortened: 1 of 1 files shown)


mesonbuild/backend/ninjabackend.py
```diff
@@ -465,7 +465,10 @@ def detect_vs_dep_prefix(self, tempfilename):
         # number of colons. Match up to the the drive name 'd:\'.
         # When used in cross compilation, the path separator is a
         # forward slash rather than a backslash so handle both.
-        matchre = re.compile(rb"^(.*\s)([a-zA-Z]:\\|\/).*stdio.h$")
+        # With certain cross compilation wrappings of MSVC, the paths
+        # use backslashes, but without the leading drive name, so
+        # allow the path to start with any path separator.
```
nirbheek, line 470: Please add an example of such paths so that people know what to test against when changing this code.

mstorsjo: Done; added an example of the existing clang-cl cross compilation case, and one example for the case that the patch adds support for.

nirbheek:
You need to rebase on top of master to fix CI failures.

mstorsjo:
> You need to rebase on top of master to fix CI failures.

Done, thanks!

PR #7753, opened 2020-09-17
Some fixes for CompCert

archi:
Some fixes for #7674:

1. Like with gcc's `ld`, also use the `group_start` code to create a
   `--start-group`/`--end-group`
2. xc16 tricked into believing the 'link_whole' was about `--*-group`,
   but it should use gcc's `--whole-archive` instead.
3. Not clear what the get_lib_prefix should really do, but for picolibc
   it seems I want just `''`.

The problem with picolibc was that the `-l` would be prefixed to a lib
like `picolib/libm/libm.a`. Though of course the `-l` would be necessary
for just a plain `m` (that's what I assumed this would be used for).
I think this might need some clarification from the meson devs ;-)

### DIFF
--- a/mesonbuild/compilers/mixins/clike.py
+++ b/mesonbuild/compilers/mixins/clike.py
@@ -32,7 +32,7 @@ from pathlib import Path
 from ... import arglist
 from ... import mesonlib
 from ... import mlog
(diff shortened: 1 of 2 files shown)


nirbheek:
You need to rebase on top of master to fix CI failures.

PR #7757, opened 2020-09-18
Extend the Curses dependency to use config tools and hand rolled search

### LINKED ISSUE #7693: Improve curses detection
Currently, meson offers custom lookup for the `curses` library by searching for `ncurses` and `ncursesw` via pkg-config. And if that was enough, the world would be a better place. Unfortunately, the curses library has a long and twisted history that makes it a nightmare to use. There's different types of curses in use today (taken from [this](https://github.com/autoconf-archive/autoconf-archive/blob/master/m4/ax_with_curses.m4) autoconf script):

- Standard curses conforming to [X/Open Curses](https://publications.opengroup.org/c094) and System V UNIX, release 4 (SVr4).
- Enhanced curses with wide char support.
- Curses with the obsolete bits i.e. functions used on old BSDs and SVr4.
- "new" curses with support for colors and mice and panels and such.

As for implementations, there's a handful (I think):

- SVr4 / X/Open curses, still used on solaris/omnios (though omnios includes ncurses in base).
- ncurses, which implements SVr4 curses plus the colors and stuff.
- ncursesw, superset of ncurses that supports widechar things. May or may not be split out as a seperate library/header.
- PDcurses, which supports wide stuff and colours. Haven't looked deep into it.
- NetBSD curses, which is its own [can of worms](https://invisible-island.net/ncurses/ncurses-netbsd.html).

Then there's the header file. Because of all these libraries, you can expect to find the "right" one in various places:

```c
#include <ncursesw/ncurses.h>
#include <ncurses/ncurses.h>
#include <ncurses/curses.h>
#include <ncurses.h>
#include <curses.h>
```

Sometimes the bottom headers are just symlink/include wrappers around the top.

As mentioned, OS maintainers may split ncursesw into its own lib, so now you have to check these:

- `-lncursesw`
- `-lncurses`
- `-lcursesw`

But that's not all, you can also see ncurses 5 and 6 on the same machine due to backwards incompatible
changes. So append a 5 or 6 to each one.

ncurses started generating pkg-config files c. [2009](https://invisible-island.net/ncurses/NEWS.html#index-t20090103) to fix this, but some OSs don't include them. Sometimes they'll bundle ncurses-config to help, but this also affected by both the wide char split and the 5/6 split, so you have to search for `ncursesw?(5-6)?-config`.

So yeah, its a little confusing. I'm actually hitting this because of htop-dev/htop#80, and I resorted to making [this monstrosity](https://git.sgregoratto.me/ncurses-meson-dep/file/ncurses/meson.build.html):

```meson
# Ultimate ncurses detection 100% real no fake
nc_libs = [
	'ncursesw',
	'ncurses',
	'curses'
]
nc_headers = [
	'ncursesw/ncurses.h',
	'ncursesw.h',
	'ncurses/ncurses.h',
	'ncurses/curses.h',
	'ncurses.h',
	'curses.h',
]
nc_progs = [
	'ncursesw6',
	'ncursesw5',
	'ncurses6',
	'ncurses5',
]
wide_incantations = [
	'-D_XOPEN_SOURCE_EXTENDED',
	'-DNCURSES_WIDECHAR',
]

found = false
foreach h : nc_headers
	if cc.has_header(h)
		found = true
		conf.set('CURSES_HEADER', h)
		break
	endif
endforeach
if not found
	error('ncurses header not found')
endif

foreach l : nc_libs
	curses = dependency(l, required: false)
	if curses.found()
		subdir_done()
	endif
endforeach
foreach p : nc_progs
	# Method is whitelisted for specific libraries, darn.
	#curses = dependency(p, method: 'config', required: false)
	#if curses.found()
	#	subdir_done()
	#endif
	name = p + '-config'
	prog = find_program([name,
		'/bin' / name,
		'/usr' / 'bin' / name,
		'/usr' / 'local' / 'bin' / name],
		required: false
	)
	if not prog.found()
		continue
	endif
	libs	= run_command(prog, '--libs-only-l').stdout().strip()
	cflags	= run_command(prog, '--cflags').stdout().strip()
	version = run_command(prog, '--version').stdout().strip()

	cflags = ' '.join([cflags] + wide_incantations)
	curses	= declare_dependency(
		compile_args: cflags,
		link_args: libs,
		version: version,
	)
	subdir_done()
endforeach
foreach l : nc_libs
	curses = cc.find_library(l, required: false)
	if curses.found()
		subdir_done()
	endif
endforeach

error('ncurses library not found')
```

And even then it will probably break on some system. As @eli-schwartz put it

> honestly this cries out for the meson builtin dependency resolver for curses to be improved.

dcbaker:
Like many other classic unix libraries curses is a portability pain, there's a bunch of slightly different implementations, that are found different ways. Meson already provides abstractions for making other widely available but not always easy to find libraries easier to use.

This adds all of the methods, and is tested on linux. I can test it on illumos/OpenIndiana and FreeBSD later if needed.

This doesn't resolve the fact that users will likely need to know which header they should `#include`. But that's a bigger change to dependencies.

Fixes #7693

### DIFF
--- a/mesonbuild/dependencies/misc.py
+++ b/mesonbuild/dependencies/misc.py
@@ -410,7 +410,7 @@ def curses_factory(env: 'Environment', for_machine: 'MachineChoice',
     candidates = []  # type: T.List[T.Callable[[], Dependency]]
 
     if DependencyMethods.PKGCONFIG in methods:
-        pkgconfig_files = ['ncurses', 'ncursesw']
+        pkgconfig_files = ['ncursesw', 'ncurses', 'curses']
         for pkg in pkgconfig_files:
             candidates.append(functools.partial(PkgConfigDependency, pkg, env, kwargs))
 
--- a/mesonbuild/dependencies/misc.py
+++ b/mesonbuild/dependencies/misc.py
@@ -404,7 +404,23 @@ class ShadercDependency(ExternalDependency):
         return [DependencyMethods.SYSTEM, DependencyMethods.PKGCONFIG]
 
 
-@factory_methods({DependencyMethods.PKGCONFIG})
+class CursesConfigToolDependency(ConfigToolDependency):
+
+    """Use the curses config tools."""
+
+    tool = 'curses-config'
+    # ncurses5.4-config is for macOS Catalina
+    tools = ['ncursesw6-config', 'ncursesw5-config', 'ncurses6-config', 'ncurses5-config', 'ncurses5.4-config']
+
+    def __init__(self, name: str, env: 'Environment', kwargs: T.Dict[str, T.Any], language: T.Optional[str] = None):
+        super().__init__(name, env, kwargs, language)
+        if not self.is_found:
+            return
+        self.compile_args = self.get_config_value(['--cflags'], 'compile_args')
+        self.link_args = self.get_config_value(['--libs'], 'link_args')
+
+
+@factory_methods({DependencyMethods.PKGCONFIG, DependencyMethods.CONFIG_TOOL})
 def curses_factory(env: 'Environment', for_machine: 'MachineChoice',
                    kwargs: T.Dict[str, T.Any], methods: T.List[DependencyMethods]) -> T.List[T.Callable[[], 'Dependency']]:
     candidates = []  # type: T.List[T.Callable[[], Dependency]]
@@ -413,6 +429,9 @@ def curses_factory(env: 'Environment', for_machine: 'MachineChoice',
         pkgconfig_files = ['ncursesw', 'ncurses', 'curses']
         for pkg in pkgconfig_files:
             candidates.append(functools.partial(PkgConfigDependency, pkg, env, kwargs))
+
+    if DependencyMethods.CONFIG_TOOL in methods:
+        candidates.append(functools.partial(CursesConfigToolDependency, 'curses', env, kwargs))
 
     return candidates
 
--- a/mesonbuild/dependencies/misc.py
+++ b/mesonbuild/dependencies/misc.py
@@ -420,7 +420,61 @@ class CursesConfigToolDependency(ConfigToolDependency):
         self.link_args = self.get_config_value(['--libs'], 'link_args')
 
 
-@factory_methods({DependencyMethods.PKGCONFIG, DependencyMethods.CONFIG_TOOL})
+class CursesSystemDependency(ExternalDependency):
+
+    """Curses dependency the hard way.
+
+    This replaces hand rolled find_library() and has_header() calls. We
+    provide this for portability reasons, there are a large number of curses
+    implementations, and the differences between them can be very annoying.
+    """
+
+    def __init__(self, name: str, env: 'Environment', kwargs: T.Dict[str, T.Any]):
+        super().__init__(name, env, kwargs)
+
+        candidates = [
+            ('ncursesw',  ['ncursesw/ncurses.h', 'ncurses.h']),
+            ('ncurses',  ['ncurses/ncurses.h', 'ncurses/curses.h', 'ncurses.h']),
+            ('curses',  ['curses.h']),
+        ]
+
+        # Not sure how else to elegently break out of both loops
+        for lib, headers in candidates:
+            l = self.clib_compiler.find_library(lib, env, [])
+            if l:
+                for header in headers:
+                    h = self.clib_compiler.has_header(header, '', env)
+                    if h[0]:
+                        self.is_found = True
+                        self.link_args = l
+                        # Not sure how to find version for non-ncurses curses
+                        # implementations. The one in illumos/OpenIndiana
+                        # doesn't seem to have a version defined in the header.
+                        if lib.startswith('ncurses'):
+                            v, _ = self.clib_compiler.get_define('NCURSES_VERSION', '#include <{}>'.format(header), env, [], [self])
+                            self.version = v.strip('"')
+
+                        # Check the version if possible, emit a wraning if we can't
+                        req = kwargs.get('version')
+                        if req:
+                            if self.version:
+                                self.is_found = mesonlib.version_compare(self.version, req)
+                            else:
+                                mlog.warning('Cannot determine version of curses to compare against.')
+
+                        if self.is_found:
+                            mlog.debug('Curses library:', l)
+                            mlog.debug('Curses header:', header)
+                            break
+            if self.is_found:
+                break
+
+    @staticmethod
+    def get_methods() -> T.List[DependencyMethods]:
+        return [DependencyMethods.SYSTEM]
+
+
+@factory_methods({DependencyMethods.PKGCONFIG, DependencyMethods.CONFIG_TOOL, DependencyMethods.SYSTEM})
 def curses_factory(env: 'Environment', for_machine: 'MachineChoice',
                    kwargs: T.Dict[str, T.Any], methods: T.List[DependencyMethods]) -> T.List[T.Callable[[], 'Dependency']]:
     candidates = []  # type: T.List[T.Callable[[], Dependency]]
@@ -432,6 +486,9 @@ def curses_factory(env: 'Environment', for_machine: 'MachineChoice',
 
     if DependencyMethods.CONFIG_TOOL in methods:
         candidates.append(functools.partial(CursesConfigToolDependency, 'curses', env, kwargs))
+
+    if DependencyMethods.SYSTEM in methods:
+        candidates.append(functools.partial(CursesSystemDependency, 'curses', env, kwargs))
 
     return candidates
 
--- a/mesonbuild/dependencies/misc.py
+++ b/mesonbuild/dependencies/misc.py
@@ -489,11 +489,15 @@ def curses_factory(env: 'Environment', for_machine: 'MachineChoice',
         for pkg in pkgconfig_files:
             candidates.append(functools.partial(PkgConfigDependency, pkg, env, kwargs))
 
-    if DependencyMethods.CONFIG_TOOL in methods:
-        candidates.append(functools.partial(CursesConfigToolDependency, 'curses', env, kwargs))
+    # There are path handling problems with these methods on msys, and they
+    # don't apply to windows otherwise (cygwin is handled seperately from
+    # windows)
+    if not env.machines[for_machine].is_windows():
+        if DependencyMethods.CONFIG_TOOL in methods:
+            candidates.append(functools.partial(CursesConfigToolDependency, 'curses', env, kwargs))
 
-    if DependencyMethods.SYSTEM in methods:
-        candidates.append(functools.partial(CursesSystemDependency, 'curses', env, kwargs))
+        if DependencyMethods.SYSTEM in methods:
+            candidates.append(functools.partial(CursesSystemDependency, 'curses', env, kwargs))
 
     return candidates
 
--- a/mesonbuild/dependencies/misc.py
+++ b/mesonbuild/dependencies/misc.py
@@ -433,6 +433,7 @@ class CursesSystemDependency(ExternalDependency):
         super().__init__(name, env, kwargs)
 
         candidates = [
+            ('pdcurses', ['pdcurses/curses.h']),
             ('ncursesw',  ['ncursesw/ncurses.h', 'ncurses.h']),
             ('ncurses',  ['ncurses/ncurses.h', 'ncurses/curses.h', 'ncurses.h']),
             ('curses',  ['curses.h']),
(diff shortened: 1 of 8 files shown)


dcbaker:
There's also a zlib fix in here, I can split that out if others prefer.

dcbaker:
Interestingly, it looks like msys is hitting the "which headers" problem...

mesonbuild/dependencies/misc.py
```diff
@@ -404,16 +404,92 @@ def get_methods():
...
+        for lib, headers in candidates:
+            l = self.clib_compiler.find_library(lib, env, [])
+            if l:
+                for header in headers:
+                    h = self.clib_compiler.has_header(header, '', env)
+                    if h[0]:
+                        self.is_found = True
+                        self.link_args = l
+                        # Not sure how to find version for non-ncurses curses implementations.
+                        # The on in illumos doesn't seem to have a version defined in the header.
```
The-King-of-Toasters, line 450: typo

test cases/frameworks/31 curses/meson.build
```diff
@@ -1,9 +1,13 @@
...
-curses = dependency('curses', required: false)
+curses = dependency('curses', required: false, method : get_option('method'), version : '>= 0')
 if not curses.found()
   error('MESON_SKIP_TEST: Curses library not found')
 endif
 
 exec = executable('basic', 'main.c', dependencies: curses)
 # didn't run the test because in general graphics fail on CI
+
+# this should fain
```
The-King-of-Toasters, line 11: Typo

The-King-of-Toasters:
I have access to a couple of systems, will mark off when tested:

- [x] OpenBSD - current and stable (uses system).
- [x] OmniOS (can use pkg-config, config-tool and system).
- [ ] Arch Linux.
- [ ] FreeBSD.

The-King-of-Toasters:
I have access to a couple of systems, will mark off when tested:

- [x] OpenBSD - current and stable (uses system).
- [x] OmniOS (can use pkg-config, config-tool and system).
- [ ] Arch Linux.
- [ ] FreeBSD.

nirbheek (approved):
LGTM, minus typos already pointed out.

lygstate:
Also, hope  msys2/mingw to be tested.

lygstate:
@dcbaker do you have time to update the pull request?

dcbaker:
Yes, I have the comments addressed locally, in just trying to get an msys environment set up to figure out why that's failing

lygstate:
> Yes, I have the comments addressed locally, in just trying to get an msys environment set up to figure out why that's failing

maybe I can help the msys environment, I've already steup it.

lygstate:
```
2020-09-18T20:33:47.0655611Z Source dir: D:\a\1\s\test cases\frameworks\31 curses
2020-09-18T20:33:47.0655960Z Build dir: D:\a\1\s\b efdeae105d
2020-09-18T20:33:47.0656322Z Build type: native build
2020-09-18T20:33:47.0656605Z Project name: curses
2020-09-18T20:33:47.0656884Z Project version: undefined
2020-09-18T20:33:47.0657356Z C compiler for the host machine: cc (gcc 10.2.0 "cc (Rev1, Built by MSYS2 project) 10.2.0")
2020-09-18T20:33:47.0657807Z C linker for the host machine: cc ld.bfd 2.35
2020-09-18T20:33:47.0658199Z Host machine cpu family: x86
2020-09-18T20:33:47.0658494Z Host machine cpu: x86
2020-09-18T20:33:47.0658943Z ncursesw6-config found: YES (sh D:\a\msys64\mingw32\bin\ncursesw6-config) 6.2.20200212
2020-09-18T20:33:47.0659413Z Run-time dependency curses found: YES 6.2.20200212
2020-09-18T20:33:47.0659925Z Dependency curses found: NO found 6.2.20200212 but need: '> 1000000' (cached)
2020-09-18T20:33:47.0660557Z WARNING: Project targeting '>= 0.51.1' but tried to use feature introduced in '0.53.0': assert function without message argument.
2020-09-18T20:33:47.0661082Z Build targets in project: 1
2020-09-18T20:33:47.0661653Z WARNING: Project specifies a minimum meson_version '>= 0.51.1' but uses features which were added in newer versions:
2020-09-18T20:33:47.0662328Z  * 0.53.0: {'assert function without message argument'}
2020-09-18T20:33:47.0662753Z WARNING: Deprecated features used:
2020-09-18T20:33:47.0663072Z  * 0.48.0: {'python3 module'}
2020-09-18T20:33:47.0663451Z  * 0.55.0: {'cmake_options arg in subproject'}
2020-09-18T20:33:47.0663697Z 
2020-09-18T20:33:47.0663987Z Found ninja-1.10.1 at D:\a\msys64\mingw32\bin\ninja.EXE
2020-09-18T20:33:47.0664988Z [1/2] "cc" "-Ibasic.exe.p" "-I." "-I..\test cases\frameworks\31 curses" "-I/mingw32/include/ncursesw" "-I/mingw32/include" "-fdiagnostics-color=always" "-pipe" "-D_FILE_OFFSET_BITS=64" "-Wall" "-Winvalid-pch" "-g" "-D_XOPEN_SOURCE=600" "-D_POSIX_C_SOURCE=199506L" -MD -MQ basic.exe.p/main.c.obj -MF "basic.exe.p\main.c.obj.d" -o basic.exe.p/main.c.obj "-c" "../test cases/frameworks/31 curses/main.c"
2020-09-18T20:33:47.0666017Z FAILED: basic.exe.p/main.c.obj 
2020-09-18T20:33:47.0666917Z "cc" "-Ibasic.exe.p" "-I." "-I..\test cases\frameworks\31 curses" "-I/mingw32/include/ncursesw" "-I/mingw32/include" "-fdiagnostics-color=always" "-pipe" "-D_FILE_OFFSET_BITS=64" "-Wall" "-Winvalid-pch" "-g" "-D_XOPEN_SOURCE=600" "-D_POSIX_C_SOURCE=199506L" -MD -MQ basic.exe.p/main.c.obj -MF "basic.exe.p\main.c.obj.d" -o basic.exe.p/main.c.obj "-c" "../test cases/frameworks/31 curses/main.c"
2020-09-18T20:33:47.0668018Z ../test cases/frameworks/31 curses/main.c:1:10: fatal error: curses.h: No such file or directory
2020-09-18T20:33:47.0668507Z     1 | #include "curses.h"
2020-09-18T20:33:47.0668792Z       |          ^~~~~~~~~~
2020-09-18T20:33:47.0669129Z compilation terminated.
2020-09-18T20:33:47.0669439Z ninja: build stopped: subcommand failed.
2020-09-18T20:33:47.0670093Z  
2020-09-18T20:33:47.0670309Z 
2020-09-18T20:33:47.0671771Z CI COMMAND ci_include:
2020-09-18T20:33:47.0722277Z Included file D:\a\1\s\b efdeae105d\build.ninja:
2020-09-18T20:33:47.0722991Z # This is the build file for project "curses"
2020-09-18T20:33:47.0723956Z # It is autogenerated by the Meson build system.
2020-09-18T20:33:47.0724462Z # Do not edit by hand.
2020-09-18T20:33:47.0724869Z 
2020-09-18T20:33:47.0725252Z ninja_required_version = 1.7.1
```

didn't found this header

lygstate:
Hi, I've found the cause of the msys2 error:

```

lygstate@DESKTOP-94PU0GB MINGW64 ~
$ pkg-config ncursesw --cflags
-D_XOPEN_SOURCE=600 -D_POSIX_C_SOURCE=199506L -IC:/CI-Tools/msys64/mingw64/include/ncursesw

lygstate@DESKTOP-94PU0GB MINGW64 ~
$ ncursesw6-config --cflags
-D_XOPEN_SOURCE=600 -D_POSIX_C_SOURCE=199506L -I/mingw64/include/ncursesw -I/mingw64/include


```
The ncursesw6-config are showing the msys2 style path, but pkg-config ncursesw --cflags showing the windows style path, and the mingw/gcc compiler just recoginize the windows sytle path, didn't recognize the msys2 style path, so we should return the windows style path instead.

lygstate:
also, lack of pdcurses detection on msys2

lygstate:
@dcbaker Hi, I've fixes the msys2 for you pull request and add pdcurses  support on msys2, you can cherry pick my patches
at 
https://github.com/mesonbuild/meson/pull/7773

lygstate:
I didn't fixes the reviews for your pull reqeust, I create that just used to verify if the CI passed

dcbaker:
@lazka I'm not sure what to do about the config-tool thing. Is there someway for us to patch up the path so that it works correctly?

dcbaker:
@lazka: is there something we can do to fix the config tools (or should I open a bug against msys somewhere?)

@lygstate I pulled your patches in, but I split the adding of pdcurses and disabling of configtool on windows into two separate patches.

lygstate:
> @lazka: is there something we can do to fix the config tools (or should I open a bug against msys somewhere?)
> 
> @lygstate I pulled your patches in, but I split the adding of pdcurses and disabling of configtool on windows into two separate patches.

No problem, I found my pull request still have  problem, and I know why,  on msys2/mingw when the curses are detect by system
we need add the include path manually, Don't know if other system also need that.

dcbaker:
Not that I've seen macos, linux, freebsd, and openindiana all just work.

lygstate:
> Not that I've seen macos, linux, freebsd, and openindiana all just work.

Yeap, mingw maybe a differen things, but from qemu, there must some platform need the include

lygstate:
I see why things goes wrong
The mingw path convert procedure are not properly handled
```
    def _convert_mingw_paths(self, args: T.List[str]) -> T.List[str]:
        '''
        Both MSVC and native Python on Windows cannot handle MinGW-esque /c/foo
        paths so convert them to C:/foo. We cannot resolve other paths starting
        with / like /home/foo so leave them as-is so that the user gets an
        error/warning from the compiler/linker.
        '''
        if not mesonlib.is_windows():
            return args
        converted = []
        for arg in args:
            pargs = []
            # Library search path
            if arg.startswith('-L/'):
                pargs = PurePath(arg[2:]).parts
                tmpl = '-L{}:/{}'
            elif arg.startswith('-I/'):
                pargs = PurePath(arg[2:]).parts
                tmpl = '-I{}:/{}'
            # Full path to library or .la file
            elif arg.startswith('/'):
                pargs = PurePath(arg).parts
                tmpl = '{}:/{}'
            elif arg.startswith(('-L', '-I')) or (len(arg) > 2 and arg[1] == ':'):
                # clean out improper '\\ ' as comes from some Windows pkg-config files
                arg = arg.replace('\\ ', ' ')
            if len(pargs) > 1 and len(pargs[1]) == 1:
                arg = tmpl.format(pargs[1], '/'.join(pargs[2:]))
            converted.append(arg)
        return converted

```

So we have no need to disable CursesConfigToolDependency on windows
Just need to fixes _convert_mingw_paths, and this is also whey qemu building with meson failed.

lygstate:
@dcbaker I've rebase and update with https://github.com/mesonbuild/meson/pull/7773 to fixes the CI, and from that we know the existing mingw path handling are error prone. we need use cygpath to do proper path convert.

So the pkg-config and custom config-tool both need use that.

lazka:
> @lazka: is there something we can do to fix the config tools (or should I open a bug against msys somewhere?)

I've filed https://github.com/msys2/MINGW-packages/issues/7025

lygstate:
> > @lazka: is there something we can do to fix the config tools (or should I open a bug against msys somewhere?)
> 
> I've filed [msys2/MINGW-packages#7025](https://github.com/msys2/MINGW-packages/issues/7025)

I've fixes it consistence with pkg-config in pull request https://github.com/mesonbuild/meson/pull/7773
So there is no need the mingw bug anymore

lazka:
>  So there is no need the mingw bug anymore

I assume fixing it doesn't hurt though? I've done so in https://github.com/msys2/MINGW-packages/pull/7026

lygstate:
> > So there is no need the mingw bug anymore
> 
> I assume fixing it doesn't hurt though? I've done so in [msys2/MINGW-packages#7026](https://github.com/msys2/MINGW-packages/pull/7026)

Not hurt:)

lygstate:
@dcbaker I've updated https://github.com/mesonbuild/meson/pull/7773#partial-pull-merging and fixes all compiling issues under win32, please have a look and cherry-pick it

jpakkane:
How does this tie in with #7773 then? Should both of them be merged or only one? If one, which one?

dcbaker:
What I think I'd like to do, is just work around the windows issue for now and merge this, then we'll take the msys issues @lygstate is trying to fix one at a time in a separate MR. does that sound reasonable?

lygstate:
> What I think I'd like to do, is just work around the windows issue for now and merge this, then we'll take the msys issues @lygstate is trying to fix one at a time in a separate MR. does that sound reasonable?
> 
> The big difference between them is that 7773 is trying to fix a bunch of msys issues as well (that break using this dependency)

reasonable

dcbaker:
Okay, I've gone ahead and merged this, let me go take a closer look at 7773 now.

PR #7758, opened 2020-09-18
dependencies/hdf5: Convert to a dependency_factory

dcbaker:
Instead of a mega dependency that does everything, use a dependency
factory for config-tool and pkg-config

### DIFF
--- a/docs/markdown/snippets/hdf5_dependnecy_improvements.md
+++ b/docs/markdown/snippets/hdf5_dependnecy_improvements.md
@@ -1,0 +1,6 @@
+## HDF5 dependency improvements
+
+HDF5 has been improved so that the internal representations have been split.
+This allows selecting pkg-config and config-tool dependencies separately.
+Both work as proper dependencies of their type, so `get_variable` and similar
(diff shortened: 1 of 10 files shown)


dcbaker:
@scivision I'm not sure what to do about the macos failures. It's find gfortran and using that with apple clang, which is of course resulting in failures due to include paths. At the same time forcing the compilers to be the same doesn't seem good.

docs/markdown/snippets/hdf5_dependnecy_improvements.md
```diff
@@ -0,0 +1,6 @@
+## HD5F dependency improvements
+
+HD5F has been improved so that the internal representations have been split.
```
QuLogic, line 3:
```suggestion
## HDF5 dependency improvements

HDF5 has been improved so that the internal representations have been split.
```

jpakkane:
LGTM. Is this ready to merge?

dcbaker:
I think so. I'm going to go ahead and merge it.

PR #7759, opened 2020-09-18
Type annotations for a lot of the compilers package

dcbaker:
This is a rather crazy amount of code, but it gets us a long way to getting the compiler package annotated. It gets the base compiler annotated (and adds a bunch of the missing methods to the base Compiler class that are defined in hgher level classes), this is important for cases where a compiler is returned (such as in the environment module). I have not gotten to the per-language compilers yet. For one thing, several of them (C, C++ and D in particular) are all very complex, and would take a lot of time and careful review. For another this MR is already huge and is going to be painful to rebase.

There is one patch in here that could probably land separately, and that's the `Use a distinct type for ocmpile/link results`. That in itself is a good cleanup, and mypy helped me spot a lot of places where we weren't handling that correctly.


mesonbuild/compilers/compilers.py
```diff
@@ -505,42 +511,54 @@ def get_display_language(cls) -> str:
     def get_default_suffix(self) -> str:
         return self.default_suffix
 
-    def get_define(self, dname, prefix, env, extra_args, dependencies) -> T.Tuple[str, bool]:
+    def get_define(self, dname: str, prefix: str, env: 'Environment',
+                   extra_args: T.Sequence[str], dependencies: T.Sequence['Dependency'],
+                   disable_cache: bool = False) -> T.Tuple[str, bool]:
```
nirbheek, line 516: Should we convert all these to `@abc.abstractmethod`?

mesonbuild/compilers/compilers.py
```diff
@@ -697,21 +719,13 @@ def cached_compile(self, code, cdata: coredata.CoreData, *, extra_args=None, mod
...
-                # This way the object can be serialized
-                tokeep = ['args', 'commands', 'input_name', 'output_name',
-                          'pid', 'returncode', 'stdo', 'stde', 'text_mode']
-                todel = [x for x in vars(p).keys() if x not in tokeep]
-                for i in todel:
-                    delattr(p, i)
-                p.cached = False
                 cdata.compiler_check_cache[key] = p
                 yield p
+                p.cached = True
```
nirbheek: This change is incorrect, but I see that you've fixed it in [a later commit](https://github.com/mesonbuild/meson/pull/7759/commits/e4fb6d02d9aa418296033ff2bb587c21854fccbe) so we can let it be.

mesonbuild/compilers/mixins/arm.py
```diff
@@ -127,30 +133,37 @@ def compute_parameters_with_absolute_paths(self, parameter_list: T.List[str], bu
...
         # Using the regular expression from environment.search_version,
         # which is used for searching compiler version
         version_regex = r'(?<!(\d|\.))(\d{1,2}(\.\d+)+(-[a-zA-Z0-9]+)?)'
-        linker_ver = re.search(version_regex, ver_str)
-        if linker_ver:
-            linker_ver = linker_ver.group(0)
+        _linker_ver = re.search(version_regex, ver_str)
+        if _linker_ver:
+            linker_ver = _linker_ver.group(0)
         if not mesonlib.version_compare(self.version, '==' + linker_ver):
```
nirbheek, line 167: `if not _linker_ver` won't this error out saying `linker_ver` is not defined?

mesonbuild/backend/ninjabackend.py
```diff
@@ -2647,10 +2647,10 @@ def guess_external_link_dependencies(self, linker, target, commands, internal):
 
         guessed_dependencies = []
         # TODO The get_library_naming requirement currently excludes link targets that use d or fortran as their main linker
-        if hasattr(linker, 'get_library_naming'):
-            search_dirs = tuple(search_dirs) + tuple(linker.get_library_dirs(self.environment))
```
nirbheek, line 2651: I am not sure I understand this. Does mypy raise errors about `hasattr` usage?

mesonbuild/compilers/mixins/clike.py
```diff
@@ -241,10 +286,10 @@ def get_library_dirs(self, env, elf_class = None):
                     # Skip the file if we can't read it
                     pass
 
-        return tuple(retval)
+        return retval
```
nirbheek, line 290: If we're using lru_cache, then we should either return a tuple (or some other immutable sequence) or make sure that everything that consumes it copies before mutating the list, otherwise the cached value will change.

mesonbuild/compilers/mixins/clike.py
```diff
@@ -764,7 +869,7 @@ def has_function(self, funcname, prefix, env, *, extra_args=None, dependencies=N
             extra_args = []
 
         # Short-circuit if the check is already provided by the cross-info file
-        varname = 'has function ' + funcname
+        varname = 'has function ' + typename
```
nirbheek, line 870: Hmm, why are we renaming from funcname to typename in this function?

mesonbuild/compilers/mixins/clike.py
```diff
@@ -1165,19 +1287,21 @@ def thread_flags(self, env):
     def thread_link_flags(self, env: 'Environment') -> T.List[str]:
         return self.linker.thread_flags(env)
 
-    def linker_to_compiler_args(self, args):
-        return args
+    def linker_to_compiler_args(self, args: T.List[str]) -> T.List[str]:
+        return list(args)
```
nirbheek, line 1289: `args.copy()` might be clearer?

mesonbuild/compilers/mixins/gnu.py
```diff
@@ -367,9 +390,9 @@ def has_arguments(self, args, env, code, mode):
                 result = False
             if self.language in {'c', 'objc'} and 'is valid for C++/ObjC++' in p.stderr:
                 result = False
-        return result, p.cached
+            return result, p.cached
```
nirbheek, line 393: Not sure I understand this change.

nirbheek (commented):
Great work, should help a lot in catching issues!

mesonbuild/backend/ninjabackend.py
dcbaker: mypy doesn't (and according to Guido Rossum) probably never will support hasattr, because it doesn't know what the annotations for the function would be.

mesonbuild/compilers/compilers.py
dcbaker: The problem with abstractmethod is children cannot be instantiated without a concrete implementation, so we'd have to reimplement the same "not supported" method for every language. I think e should just use abstractmethod for cases where every language needs to implement the method

mesonbuild/compilers/mixins/clike.py
dcbaker: Because typename is what the base compiler calls it and I just copied the signature from there and touched up the function body. I can change then to funcname if you prefer

mesonbuild/compilers/mixins/gnu.py
dcbaker: Ah, I can change that. At one point in the series I was doing something clever to set the cached variable, but then changes my mind and this didn't get changed back

mesonbuild/compilers/mixins/arm.py
dcbaker: Ah, yup. Good catch

mesonbuild/compilers/mixins/clike.py
dcbaker: Actually, tuples are mutable too. Really we need to copy the return value to ensure that if a caller mutates it, then they won't cause problems. Or document that callers need to copy before mutation. I'll double check that before the next version

mesonbuild/compilers/mixins/clike.py
dcbaker: I need to think about this, it's actually possible that you can get a `CompilerArgs` here... Need to look in the backend code

mesonbuild/mesonlib.py
```diff
@@ -34,6 +34,8 @@
     from .compilers.compilers import CompilerType
     from .interpreterbase import ObjectHolder
 
+    FileOrString = T.Union['File', str]
```
mensinda, line 37: Is there a specific reason/benefit why this is inside the `T.TYPE_CHECKING` if while the? If not, would you consider moving this out, so that it doesn't have to be quoted in other parts of the codebase?

IMHO it isn't really intuitive what style is required (especially for new python developers) if `T.TYPE_CHECKING` blocks in other files have to be checked.

mesonbuild/coredata.py
```diff
@@ -32,10 +32,11 @@
 
 if T.TYPE_CHECKING:
     from . import dependencies
-    from .compilers import Compiler  # noqa: F401
+    from .compilers.compilers import Compiler, CompileResult  # noqa: F401
     from .environment import Environment
 
     OptionDictType = T.Dict[str, 'UserOption[T.Any]']
+    CompilerCheckCacheKey = T.Tuple[T.Tuple[str, ...], str, str, T.Tuple[str, ...], str]
```
mensinda, line 39: Same question from `mesonlib.py`

mesonbuild/compilers/compilers.py
```diff
@@ -544,70 +585,125 @@ def get_linker_args_from_envvars(self,
...
     def unix_args_to_native(cls, args: T.List[str]) -> T.List[str]:
         "Always returns a copy that can be independently mutated"
-        return args[:]
+        return args.copy()
 
     @classmethod
     def native_args_to_unix(cls, args: T.List[str]) -> T.List[str]:
         "Always returns a copy that can be independently mutated"
-        return args[:]
+        return args.copy()
```
mensinda, line 682: Is there a specific reason for switching to `.copy()`? According to [StackOverflow](https://stackoverflow.com/questions/2612802/list-changes-unexpectedly-after-assignment-how-do-i-clone-or-copy-it-to-prevent) the old slicing should be faster.

mesonbuild/compilers/mixins/clike.py
```diff
@@ -111,14 +119,48 @@ class CLikeCompiler:
 
     """Shared bits for the C and CPP Compilers."""
 
+    if T.TYPE_CHECKING:
+        can_compile_suffixes = set()  # type: T.Set[str]
+        exelist = []  # type: T.List[str]
+        for_machine = mesonlib.MachineChoice.HOST
+        id = ''
+        ignore_libs = ()  # typoe: T.Tuple[str, ...]
```
mensinda, line 127: `typoe` --> `type`

mensinda (commented):
This looks really good overall, however, I also have a few nitpicks.

mesonbuild/compilers/mixins/clike.py
nirbheek: The tuple type is a container which is not mutable, but it can contain other mutable types. Tuples of strings are completely immutable. I think it's easier to assert that all functions that return a string sequence are tuples using mypy than to assert that they are returning _a copy of a list_. So I think it would be safer to always return a tuple from all our compiler APIs.

mesonbuild/compilers/mixins/clike.py
nirbheek: I think it should be `funcname` because that's actually what it is, the base compiler prototype should be changed here :)

mesonbuild/compilers/compilers.py
nirbheek: I think arg conversion is also a hotspot, so the speed difference may be significant.

dcbaker: Python is my primary programming language, and I had to look up with the heck `[:]` does. It's extremely non-obvious, but anyone can see `foo.copy()` and reasonably understand what that does.

I'm suspicious of those benchmarking numbers, especially because python 3 has a bunch of optimizations to make using methods faster that python2 lacked. And that answer doesn't use `list.copy()`, but `copy.copy`, since the former was added in python3, and not present in python2.

```python
from timeit import timeit

a = [str(a) for a in range(50)]

print('slice:', timeit(lambda: a[:]))
print('copy:', timeit(lambda: a.copy()))
print('new list:', timeit(lambda: list(a)))
```
and here's my results:

python3.6:
```
slice: 0.3625514349998866
copy: 0.36979652300010457
new list: 0.4572163679999903
```

python3.7
```
slice: 0.346616264999966
copy: 0.3320672290001312
new list: 0.41092001400011213
```

python3.8
```
slice: 0.30240244600008737
copy: 0.2990029269999468
new list: 0.319297583999969
```

python3.9
```
slice: 0.3030145709999488
copy: 0.26365815300005124
new list: 0.3073842159999458
```

Note that timeit calls the callable 1,000,000 times. The difference in performance between copy() and slicing is minimal (and varies, sometimes one is faster than the other.)

mesonbuild/compilers/compilers.py
dcbaker: This isn't incorrect per se, but its entirely too clever, which is why I dropped it.

mesonbuild/compilers/mixins/arm.py
dcbaker: Actually, this code is all silly, we already have the linker version in self.linker.version... I'm going to just change it to do that.

mesonbuild/compilers/mixins/clike.py
dcbaker: :/ It's just really gross to treat tuples as immutable lists, because that's not what they're for.

mesonbuild/compilers/mixins/clike.py
dcbaker: no, this can only be a list, so `.copy()` is right

mesonbuild/mesonlib.py
dcbaker: Ḃecause it's only used by mypy? It seems really strange to me to define or import things outside of an if type_checking block that are only used for type checking.

mesonbuild/compilers/mixins/clike.py
dcbaker: I've wrapped the cached function as a private/public pair. If you're deadset on using a tuple I guess we can do that, but I really dislike using `tuple[str, ...]`, because that's a code smell.

mesonbuild/compilers/compilers.py
nirbheek: Thanks for doing the testing, lgtm :)

mesonbuild/compilers/mixins/clike.py
nirbheek: We really need some kind of immutable sequence type to return from all our compiler methods. We've hit the 'oh no we accidentally returned a mutable copy and changed it for everyone' issue too many times over the years. Most recently: https://github.com/mesonbuild/meson/pull/6852, and we'll definitely hit it again.

We can do it later (i.e., not in this PR), but Typing seems like an ideal opportunity to fix this.

mesonbuild/mesonlib.py
mensinda: The thing is, we are currently doing this inconsistently and I dislike using strings because of missing syntax highlighting. Also, you could argue that it is (kind of) used in the type hints since they are part of the code. Python will complain if you have a typo in an annotation without strings for instance.

dcbaker: Until 3.7 when they realized it was a terrible idea for the interpreter to actually interpret the type hints (because it has memory and startup time considerations). in the 3.7 with `__future__.annotations` the interpreter treats them as strings always. Eventually (like all __future__ stuff) that will become the default. I wish we could drop 3.6 support already and just use the 3.7+ syntax :)

mesonbuild/compilers/mixins/clike.py
dcbaker: As a pythonista... It isn't really possible, python is all mutable all the time. returning copies really is the best we can do. I really wish mypy had a `const[]` annotation so we could catch people mutating stuff we don't want them to without the runtime costs of using properties for that.

dcbaker:
I've made a few more changes (that are pretty big). It dawned on me last night that rather than re-implement all of the methods from `Compiler` in the various mixins (and risk bitrotting), I could do something clever with mypy to get mypy to check them against the base `Compiler` class.

In python3 all classes are "new" classes, meaning the inherit from `object`. so `class Foo is class Foo(object)`. with the TYPE_CHECKING thing we can do this:

```python
if T.TYPE_CHECKING:
    from compilers import Compiler
else:
    Compiler = object

class Mixin(Compiler): ...
```

now, for run time this has 0 effect, `Compiler` is just an alias of `object`, but for mypy it will check them Mixins as if they're instances of `Compiler`, and viola, I found all the bugs.

I've also added type annotaions for ObjC and ObjC++, because I wanted to make sure that this cleverness wasn't going to break things for mypy when we actually started putting the mixins together. And, I found a bug. of course.

mesonbuild/compilers/mixins/gnu.py
```diff
@@ -164,14 +170,14 @@ def get_buildtype_args(self, buildtype: str) -> T.List[str]:
 
     @abc.abstractmethod
     def get_optimization_args(self, optimization_level: str) -> T.List[str]:
-        raise NotImplementedError("get_optimization_args not implemented")
```
mensinda, line 167: Why are these exceptions removed?

dcbaker: they're redundant with `abc.abstracetmethod`, you can't instantiate a class that has non-overloaded abstract methods, the `ABCMeta` metaclass prevents it.

mensinda: OK, thanks for explaining

dcbaker:
yay, mypy + pylint found my bugs for me.

mesonbuild/compilers/mixins/clike.py
mensinda: There is `Final` in [typing_extensions](https://mypy.readthedocs.io/en/stable/final_attrs.html)...

mensinda (approved):
LGTM

mesonbuild/compilers/mixins/clike.py
dcbaker: Which isn't quite what I want unfortunately, `Final` allows you to say "don't let a subclass overwrite this", so you can do things like:

```python
class Base:
   cache: Final[T.Dict[str, str]] = {}  # this needs to be shared by all subclasses

class Child(Base):
    cache: T.Dict[str, str] = {}   # mypy error!
```

dcbaker: Actually, I just had a thought. We probably could get mypy to do this with a Protocol, if we had a list-like protocol that lacked any of the mutation methods....

dcbaker: Yes, in my testing that seems to work... I'm going to play with that some more.

nirbheek: ❤

dcbaker:
@nirbheek, is this okay with you?

nirbheek:
LGTM, we should merge asap.

dcbaker:
Done. Thanks for reading through all of this.

PR #6367, opened 2019-12-18
Properly handle the case of linking static library with custom targets

yshui:
To avoid reintroduce the problem fixed by 484b72136995, when a static
library is linked with a custom target that is not installed, we need to
use link_whole(). However link_whole() explicitly doesn't work with
custom targets.

So this commit instead generates a warning when such a case is detected.

This commit also avoids a call to is_internal() on
CustomTarget/CustomTargetIndex objects, thus fixes #6365.

Signed-off-by: Yuxuan Shui <yshuiv7@gmail.com>

PR #7769, opened 2020-09-22
Custom target link

### LINKED ISSUE #6365: AttributeError: 'CustomTargetIndex' object has no attribute 'is_internal'
**Describe the bug**
meson crashes with the error in the title

**To Reproduce**
```meson
project('bug', 'c')
t = custom_target('dummy', output: [ 'libdummy.a' ], command: [ 'echo', 'dummy' ])
dep = declare_dependency(link_with: t[0])

l = static_library('l', 'src/empty.c', dependencies: dep, install: true)
```

**system parameters**
* ArchLinux
* Python: 3.8.0
* meson: 0.52.1

Sahnvour:
Continuation of #6367, fixes #6365.

I had to implement both `is_internal` and `extract_all_objects_recurse` in the end because they're needed later during execution. It works as is, but I'm not sure it's correct @nirbheek.

The test initially called `echo dummy` as an external command but this fails on windows, so I figured python should be OK instead.

### DIFF
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -1130,10 +1130,16 @@ You probably should put it in link_with instead.''')
 
     def link(self, target):
         for t in unholder(listify(target)):
-            if isinstance(self, StaticLibrary) and self.need_install and t.is_internal():
-                # When we're a static library and we link_with to an
-                # internal/convenience library, promote to link_whole.
-                return self.link_whole(t)
+            if isinstance(self, StaticLibrary) and self.need_install:
+                if isinstance(t, (CustomTarget, CustomTargetIndex)):
+                    if not t.should_install():
+                        mlog.warning('Try to link an installed static library target {} with a custom target '
+                                     'that is not installed, this might cause problems when you try to use '
+                                     'this static library'.format(self.name))
+                elif t.is_internal():
+                    # When we're a static library and we link_with to an
+                    # internal/convenience library, promote to link_whole.
+                    return self.link_whole(t)
             if not isinstance(t, (Target, CustomTargetIndex)):
                 raise InvalidArguments('{!r} is not a target.'.format(t))
             if not t.is_linkable_target():
@@ -2274,6 +2280,18 @@ class CustomTarget(Target):
     def get_all_link_deps(self):
         return []
 
+    def is_internal(self) -> bool:
+        if not self.should_install():
+            return True
+        for out in self.get_outputs():
+            # Can't check if this is a static library, so try to guess
+            if not out.endswith(('.a', '.lib')):
+                return False
+        return True
+
+    def extract_all_objects_recurse(self):
+        return self.get_outputs()
+
     def type_suffix(self):
         return "@cus"
 
@@ -2418,6 +2436,15 @@ class CustomTargetIndex:
         suf = os.path.splitext(self.output)[-1]
         if suf == '.a' or suf == '.dll' or suf == '.lib' or suf == '.so':
             return True
+
+    def should_install(self) -> bool:
+        return self.target.should_install()
+
+    def is_internal(self) -> bool:
+        return self.target.is_internal()
+
+    def extract_all_objects_recurse(self):
+        return self.target.extract_all_objects_recurse()
 
 class ConfigureFile:
 
--- a/test cases/common/215 link custom/meson.build
+++ b/test cases/common/215 link custom/meson.build
@@ -57,3 +57,22 @@ d2_i = declare_dependency(link_whole: clib[0])
 
 exe4_i = executable('prog4_i', 'prog.c', dependencies: d2_i)
 test('linkwhole2_i', exe2_i)
+
+# Link with custom target
+
+dummy = static_library('dummy', 'dummy.c')
+
+custom_prog = find_program('custom_target.py')
+t = custom_target('custom', input: dummy, output: 'libcustom.a', command: [custom_prog, '@INPUT@', '@OUTPUT@'])
+
+dep1 = declare_dependency(link_with: t)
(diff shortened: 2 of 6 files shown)


mesonbuild/build.py
```diff
@@ -1113,10 +1113,16 @@ def is_internal(self):
...
+            if isinstance(self, StaticLibrary) and self.need_install:
+                if isinstance(t, (CustomTarget, CustomTargetIndex)):
+                    if not t.should_install():
+                        mlog.warning('Try to link an installed static library target {} with a custom target '
+                                     'that is not installed, this might cause problems when you try to use '
+                                     'this static library'.format(self.name))
+                elif t.is_internal():
+                    # When we're a static library and we link_with to an
+                    # internal/convenience library, promote to link_whole.
+                    return self.link_whole(t)
```
dcbaker, line 1125: This doesn't seem right, you're not linking if it's a `CustomTarget` or `CustomTargetIndex`

Sahnvour:
The failure on Travis is a timeout.

mesonbuild/build.py
```diff
@@ -2402,6 +2408,15 @@ def is_linkable_target(self):
         if suf == '.a' or suf == '.dll' or suf == '.lib' or suf == '.so':
             return True
 
+    def should_install(self) -> bool:
+        return False
```
nirbheek, line 2412: Shouldn't this do `return self.target.should_install()`?

mesonbuild/build.py
```diff
@@ -2402,6 +2408,15 @@ def is_linkable_target(self):
...
+
+    def is_internal(self) -> bool:
+        return False
```
nirbheek, line 2415: Shouldn't this method also be implemented on `CustomTarget` and proxy to it in the same way as `should_install`()?

mesonbuild/build.py
```diff
@@ -2402,6 +2408,15 @@ def is_linkable_target(self):
...
+
+    def extract_all_objects_recurse(self):
+        return self.get_outputs()
```
nirbheek, line 2418: Same comment as `is_internal()`.

Sahnvour:
Thanks, updated with your feedback.

mesonbuild/build.py
```diff
@@ -2402,6 +2414,15 @@ def is_linkable_target(self):
         if suf == '.a' or suf == '.dll' or suf == '.lib' or suf == '.so':
             return True
 
+    def should_install(self) -> bool:
+        return self.target.should_install()
+
+    def is_internal(self) -> bool:
+        return self.target.is_internal()
```
nirbheek, line 2421: Hmm, looking at the code, `t.is_internal()` is never called on a CustomTarget or CustomTargetIndex, why do we implement this method? Same above.

mesonbuild/build.py
```diff
@@ -2402,6 +2414,15 @@ def is_linkable_target(self):
...
+
+    def extract_all_objects_recurse(self):
+        return self.target.extract_all_objects_recurse()
```
nirbheek, line 2424: Same, this patch doesn't seem to be using this new method?

nirbheek: Oh, I see you're saying they're used later during execution.

mesonbuild/build.py
```diff
@@ -2257,6 +2263,12 @@ def get_link_dep_subdirs(self):
     def get_all_link_deps(self):
         return []
 
+    def is_internal(self) -> bool:
+        return False
```
nirbheek, line 2267: `is_internal()` is defined on `BuildTarget` as:

```py
def is_internal(self):
    return isinstance(self, StaticLibrary) and not self.need_install
```

I think in this case we can do something like:

```py
def is_internal(self) -> bool:
    if not self.should_install():
        return True
    for out in self.get_outputs():
        # Can't check if this is a static library, so try to guess
        if not out.endswith(('.a', '.lib')):
            return False
    return True
```

Sahnvour:
Updated, the failing tests are timed out.

test cases/common/215 link custom/meson.build
```diff
@@ -57,3 +57,15 @@ d2_i = declare_dependency(link_whole: clib[0])
 
 exe4_i = executable('prog4_i', 'prog.c', dependencies: d2_i)
 test('linkwhole2_i', exe2_i)
+
+# Link with custom target
+
+py_mod = import('python')
+py3 = py_mod.find_installation()
+t = custom_target('dummy', output: [ 'libdummy.a' ], command: [ py3, '--version' ])
```
jpakkane, line 65: This is not testing the behaviour properly. The custom target does not actually seem to generate a library so the link operations below are useless and in fact they should fail. It's strange that they don't. A proper test should do something like:

- build a static library called dummy which has a function `inner_lib_func` or something
- create a custom target that takes the dummy lib as input and creates a different lib, say `libcustom` by copying the input lib file to the output lib file
- the static library `lib1` should have a function `outer_lib_func` that calls `inner_lib_func`
- there should be a test binary `exe1` that links against `lib1` (and only that) and calls `outer_lib_func` and there should be a test that runs this
- there should also be a `exe2` that does the same for `lib2`

Sahnvour: Updated following your advice. As is noted in common test case 28, it's not possible to emulate `cp` on windows so it will be skipped in case meson can't find it.

jpakkane: Use of shell commands in Meson is forbidden Write a Python script that does the copying instead. It works on all platforms.

Sahnvour: Updated.

nirbheek:
CI failures are unrelated. I'm not even sure how the mypy failure got through.

PR #7773, opened 2020-09-23
Curses

lygstate:
Fixes for https://github.com/mesonbuild/meson/pull/7757


mesonbuild/dependencies/base.py
```diff
@@ -327,6 +333,66 @@ def __init__(self, type_name, environment, kwargs, language: T.Optional[str] = N
         HasNativeKwarg.__init__(self, kwargs)
         self.clib_compiler = detect_compiler(self.name, environment, self.for_machine, self.language)
 
+    def convert_path_posix_to_native(self, p):
+        return convert_path_posix_to_native(p)
+
+    def get_include_paths_from_suffix(self, include_suffix):
+        global msystem
```
dcbaker, line 340: I'll look over this more, but you don't need global here. You only need global when you want to *assign* to something from the global namespace, like:

```python
value = None

def set_value(value):
    global value
    value = foo
```

lygstate: OK, this is also the place I need considerating sys_root and CMAKE_PREFIX_PATH, please give me suggestion or solution

mesonbuild/dependencies/base.py
```diff
@@ -132,6 +133,9 @@ def __repr__(self):
         s = '<{0} {1}: {2}>'
         return s.format(self.__class__.__name__, self.name, self.is_found)
 
+    def get_native_args(self, args):
```
dcbaker, line 136:
```suggestion
    def get_native_args(self, args: T.List[str]) -> T.List[str]:
```

mesonbuild/dependencies/base.py
```diff
@@ -308,6 +312,8 @@ def __init__(self, kwargs):
     def get_for_machine_from_kwargs(self, kwargs):
         return MachineChoice.BUILD if kwargs.get('native', False) else MachineChoice.HOST
 
+msystem = detect_msystem()
```
dcbaker, line 315: I think it probably makes more sense to add an `is_msys` to the `MachineInfo` class in `envconfig.py` and possibly one in `mesonlib.py` as well. then in the dependency object itself you can do something like:

```python
if self.environment.machines[self.for_machine].is_msys():
    ....
````

File: mesonbuild/dependencies/base.py [hunk 4]
```diff
@@ -327,6 +333,66 @@ def __init__(self, type_name, environment, kwargs, language: T.Optional[str] = N
...
+        if not mesonlib.is_windows():
+            return args
+        converted = []
+        for arg in args:
+            prefix = ''
+            p = arg
+            if arg.startswith('-isystem'):
+                prefix = '-isystem'
+                p = arg[8:]
+            elif arg.startswith('-I') or arg.startswith('/I'):
```
dcbaker, line 375:
```suggestion
            elif arg.startswith(('-I'. '/I')):
```

mesonbuild/dependencies/base.py
```diff
@@ -2058,6 +2093,38 @@ def get_path(self):
     def get_name(self):
         return self.name
 
+class Msys2PathConverter():
+
+    def __init__(self):
```
dcbaker, line 2098:
```suggestion
    def __init__(self) -> None:
```

mesonbuild/dependencies/base.py
```diff
@@ -2058,6 +2093,38 @@ def get_path(self):
...
+        self.cygpath = ExternalProgram('cygpath', silent=True)
```
dcbaker, line 2099: If we don't find cygpath then meson will stop completely, is that an okay behavior in this case?

mesonbuild/dependencies/base.py
```diff
@@ -2058,6 +2093,38 @@ def get_path(self):
...
+
+    def win2posix(self, path):
```
dcbaker, line 2101:
```suggestion
    def win2posix(self, path: str) -> str:
```

mesonbuild/dependencies/base.py
```diff
@@ -2058,6 +2093,38 @@ def get_path(self):
...
+        """Convert a Windows path to a msys path"""
+        p, output, error = Popen_safe(self.cygpath.command + ['-u', path], write='', stdin=subprocess.PIPE)
+        posix_path = output.strip()
+        return posix_path
+
+    def posix2win(self, path):
```
dcbaker, line 2107:
```suggestion
    def posix2win(self, path: str) -> str:
```

mesonbuild/dependencies/base.py
dcbaker: You may also want and `is_mingw` method, after reading through this more

dcbaker (changes requested):
I've merged the basic dependency with pkg-config only for windows, so it would be good to rebase.

mesonbuild/dependencies/base.py
lygstate: its ok， cygpath are the base component of msys

mesonbuild/dependencies/base.py
lygstate: What does for_machine means? does for_machine means the target platform? for this, we are talking about native platform, not the target platform, that's different.

mesonbuild/dependencies/base.py
lygstate: > If we don't find cygpath then meson will stop completely, is that an okay behavior in this case?

silent=True means won't stop completely

mesonbuild/dependencies/base.py
dcbaker: `for_machine` is usually a `mesonlib.MachineChoice`, which means that it tells whether this compiler/dependency targets the build machine or the host machine. You could, for example, be using an msys compiler as your "build" compiler and msvc as your "host" compiler, for example when building for arm based windows.

mesonbuild/dependencies/base.py
dcbaker: no, that's what `required` does. `silent` means don't print a found message on the command line.

dcbaker: But anyway, I think that answers the question.

mesonbuild/dependencies/base.py
dcbaker: This is the only thing left I'm not sure about, everything else looks good. @Ericson2314, do you have an opinion on whether this is fine or we should add it to the `MachineInfo` class?

mesonbuild/dependencies/base.py
lygstate: The constructor of ExternalProgram doesn't have the required option, do you talking about the 
ExternalDependency class constructor?
```
class ExternalProgram:
    windows_exts = ('exe', 'msc', 'com', 'bat', 'cmd')
    # An 'ExternalProgram' always runs on the build machine
    for_machine = MachineChoice.BUILD

    def __init__(self, name: str, command: T.Optional[T.List[str]] = None,
                 silent: bool = False, search_dir: T.Optional[str] = None,
                 extra_search_dirs: T.Optional[T.List[str]] = None):
```

mesonbuild/dependencies/base.py
lygstate: > This is the only thing left I'm not sure about, everything else looks good. @Ericson2314, do you have an opinion on whether this is fine or we should add it to the `MachineInfo` class?

I still don't know how to add it to the MachineInfo class. Because I am retrieving host info, not the guest info

lygstate: ping

lazka: `__CYGWIN__` is defined in both msys2/cygwin, `__MSYS__` is defined only in msys2.

lygstate: > I did a bit of googling, you can use the compiler to determine whether you're using a mingw compiler, because `__MINGW32__` will bedefined. I'm not sure about msys. But the more I think about it, the more I think we need to handl this in the MachineInfo class, because if we don't then this _only_ works for the build machine.
> 
> @lazka, is there an easy way for us to figure out if we're running in an msys2 environment from the compiler? or some other way than just "is an env var defined"? I want to be able to make sure that we're associating the correct machine (the host or the build) with msys.

Gotcha, using the compiler to detecting __MINGW32__ is a good idea, I may use that, but din't know where to start yet.
So if the platform is windows, and dtecting the compiler macro, then we using cygpath to converting paths?
Or we could implement the path convert directly using python script?

lygstate: The cygpath belongs to msys2 runtime, so if msys2 present, it always present. unless we doesn't want to using msys2 runtime at all
https://packages.msys2.org/package/msys2-runtime?repo=msys&variant=x86_64

lygstate: > `__CYGWIN__` is defined in both msys2/cygwin, `__MSYS__` is defined only in msys2.

So all of these can be deteced by using the compiler

lazka: I'm not a fan of hardcoding msys2 specifics in meson, but other than that: if windows and the path starts with "/mingw" and cygpath is in PATH would be a good indicator.

lygstate: > I'm not a fan of hardcoding msys2 specifics in meson, but other than that: if windows and the path starts with "/mingw" and cygpath is in PATH would be a good indicator for converting paths. But not 100% sure what you are trying to do here.

I am trying to converting path from /mingw32/include/a to windows' style absolute path, sometimes the mingw compiler doesn't recognize the path like -I/mingw32/include/a

lygstate: > I'm not a fan of hardcoding msys2 specifics in meson, but other than that: if windows and the path starts with "/mingw" and cygpath is in PATH would be a good indicator for converting paths (or maybe the path not existing, but the converted path existing). But not 100% sure what you are trying to do here.

by the way, if cygpath are already present, then the is no need convert it by myself.

lygstate: Hi, I am confusing, does MachineInfo system standfor? are the system stand for the target system or the host system?

def detect_machine_info(env: T.Optional['Environment'] = None, compilers: T.Optional[CompilersDict] = None) -> MachineInfo:
    """Detect the machine we're running on

    If compilers are not provided, we cannot know as much. None out those
    fields to avoid accidentally depending on partial knowledge. The
    underlying ''detect_*'' method can be called to explicitly use the
    partial information.
    """
    if env and compilers:
        print('Start print')
        for compiler in compilers.values():
            v, _ = compiler.get_define('__MINGW32__',  '', env, [], [])
            print(compiler)
            print(v)

    return MachineInfo(
        detect_system(),
        detect_cpu_family(compilers) if compilers is not None else None,
        detect_cpu(compilers) if compilers is not None else None,
        sys.byteorder)

I can using compiler to retrive the __MINGW32__ macro info

lygstate: How to setup the native build for detect_system()

lygstate: ping for suggestion. there is no system such as mingw and msys, how to add it and dosn't affect things?

Ericson2314: > Hi, I am confusing, does MachineInfo system standfor? are the system stand for the target system or the host system?

@lygstate we have only tried to auto-detect assuming native compilation, but I am not too pleased with the way auto-detection works today (guess platform, find compiler, guess platform harder using compiler). If you have some other ideas I'd certainly be open to it

Ericson2314: @dcbaker I think I would prefer it go in `MachineInfo`. I wish we didn't have any top-level `is_*` functions either, but I guess I never removed those.

lygstate: > @dcbaker I think I would prefer it go in `MachineInfo`. I wish we didn't have any top-level `is_*` functions either, but I guess I never removed those.

no problem, but how? maybe we need a more parameter such as vendor in MachineInfo? cygwin/msys2 almost the same but different vendor; msvc/intel-for-msvc/clang-for-msvc almost the same but different vendor;  msys2/mingw are also a compier for windows? how to distinguish them properly in machineInfo?

lygstate: ping for help

jpakkane:
What is the status of this now? Everyone happy or are changes still requested?

lygstate:
> What is the status of this now? Everyone happy or are changes still requested?
Don't know how to implement is_mingw in mesonlib.MachineChoice, I need help

bonzini:
I don't think it needs to be a blocker for 0.56.0

mesonbuild/dependencies/base.py
jpakkane: If what @lazka says above is true and this issue is fixed in msys2, how much of this MR is still needed?

jpakkane:
Reposting my question from above:

If what @lazka says above is true and this issue is fixed in msys2, how much of this MR is still needed?

lygstate:
> Reposting my question from above:
> 
> If what @lazka says above is true and this issue is fixed in msys2, how much of this MR is still needed?

QEMU now detecting curl by themself, so this is not needed by qemu anymore, but that's not means the curses detection in meson aregood engouh, anyway I can close this pull request

The following code are the qemu curses/iconv detecting code, they are related.

```

iconv = not_found
curses = not_found
if have_system and not get_option('curses').disabled()
  curses_test = '''
    #include <locale.h>
    #include <curses.h>
    #include <wchar.h>
    int main(void) {
      wchar_t wch = L'w';
      setlocale(LC_ALL, "");
      resize_term(0, 0);
      addwstr(L"wide chars\n");
      addnwstr(&wch, 1);
      add_wch(WACS_DEGREE);
      return 0;
    }'''

  curses_dep_list = targetos == 'windows' ? ['ncurses', 'ncursesw'] : ['ncursesw']
  foreach curses_dep : curses_dep_list
    if not curses.found()
      curses = dependency(curses_dep,
                          required: false,
                          method: 'pkg-config',
                          static: enable_static)
    endif
  endforeach
  msg = get_option('curses').enabled() ? 'curses library not found' : ''
  curses_compile_args = ['-DNCURSES_WIDECHAR']
  if curses.found()
    if cc.links(curses_test, args: curses_compile_args, dependencies: [curses])
      curses = declare_dependency(compile_args: curses_compile_args, dependencies: [curses])
    else
      msg = 'curses package not usable'
      curses = not_found
    endif
  endif
  if not curses.found()
    has_curses_h = cc.has_header('curses.h', args: curses_compile_args)
    if targetos != 'windows' and not has_curses_h
      message('Trying with /usr/include/ncursesw')
      curses_compile_args += ['-I/usr/include/ncursesw']
      has_curses_h = cc.has_header('curses.h', args: curses_compile_args)
    endif
    if has_curses_h
      curses_libname_list = (targetos == 'windows' ? ['pdcurses'] : ['ncursesw', 'cursesw'])
      foreach curses_libname : curses_libname_list
        libcurses = cc.find_library(curses_libname,
                                    required: false,
                                    static: enable_static)
        if libcurses.found()
          if cc.links(curses_test, args: curses_compile_args, dependencies: libcurses)
            curses = declare_dependency(compile_args: curses_compile_args,
                                        dependencies: [libcurses])
            break
          else
            msg = 'curses library not usable'
          endif
        endif
      endforeach
    endif
  endif
  if not get_option('iconv').disabled()
    foreach link_args : [ ['-liconv'], [] ]
      # Programs will be linked with glib and this will bring in libiconv on FreeBSD.
      # We need to use libiconv if available because mixing libiconv's headers with
      # the system libc does not work.
      # However, without adding glib to the dependencies -L/usr/local/lib will not be
      # included in the command line and libiconv will not be found.
      if cc.links('''
        #include <iconv.h>
        int main(void) {
          iconv_t conv = iconv_open("WCHAR_T", "UCS-2");
          return conv != (iconv_t) -1;
        }''', args: config_host['GLIB_CFLAGS'].split() + config_host['GLIB_LIBS'].split() + link_args)
        iconv = declare_dependency(link_args: link_args, dependencies: glib)
        break
      endif
    endforeach
  endif
  if curses.found() and not iconv.found()
    if get_option('iconv').enabled()
      error('iconv not available')
    endif
    msg = 'iconv required for curses UI but not available'
    curses = not_found
  endif
  if not curses.found() and msg != ''
    if get_option('curses').enabled()
      error(msg)
    else
      warning(msg + ', disabling')
    endif
  endif
endif
```

PR #7779, opened 2020-09-24
Add a non-mutable list type for mypy checking

dcbaker:
this adds a *type checking* dependency on `typing_extensions` for python < 3.8. For runtime there is no extra dependency.

The `ImmutableListProtocol` I've added is an immutable (doesn't implement any mutation methods) list protocol (think typescript or go interface), that`list` implements. So we can return a list as one of these protocols, but if anyone tries to use one of the mutation methods mypy will complain.

Obviously this is a much more radical approach to using mypy than we've been doing so far, since so far we've just been annotating things. What do you guys think of this kind of use of mypy?


mensinda:
I like this change, however, it would be nice to also have a check in `run_mypy.py` to ensure that `typing_extensions` is installed (if required) with a clear error message and not a stack trace.

Also, this needs a rebase...

eli-schwartz:
AFAICT all the meaningful bits of this PR landed in #8724, can we close this as obsolete?

PR #7782, opened 2020-09-25
Improve documentation about using shell in custom_target()  [skip ci]

jon-turney:
Add a note about the portability of using shell constructs in the custom_target() command.

Even after this change, there's some conflict between these different parts of the documentation, and it's a bit unclear what the intent is.

We say only a single command is _permitted_, but don't do anything to enforce that, and take steps in the ninja backend to make `&&` work as expected (so of course, people end up using it).

https://github.com/mesonbuild/meson/blob/dff40ca259c396568eeb4d05c534781ca148f8e7/mesonbuild/backend/ninjabackend.py#L178-L180

### DIFF
--- a/docs/markdown/Custom-build-targets.md
+++ b/docs/markdown/Custom-build-targets.md
@@ -29,14 +29,15 @@ it does for source generation.
 
 See [Generating Sources](Generating-sources.md) for more information on this topic.
 
-## Details on compiler invocations
+## Details on command invocation
 
 Meson only permits you to specify one command to run. This is by
 design as writing shell pipelines into build definition files leads to
-code that is very hard to maintain. If your compilation requires
+code that is very hard to maintain. If your command requires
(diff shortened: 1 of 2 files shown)


jon-turney:
> We say only a single command is _permitted_, but don't do anything to enforce that, and take steps in the ninja backend to make `&&` work as expected

This also interacts badly with command pickling, as if the command ends up in the build.ninja, it's executed by ninja with `sh -c`, but if it ends up in a pickle, it's executed by meson directly with `execve()`.  This causes particular problems on Cygwin, where we more often end up wrapping a command (due to the need to do PATH manipulation) (e.g. see [1]).

See https://github.com/jon-turney/meson/commit/1157e7fa60d62080f69c6460fcb9c3200d8289cd for a test case which demonstrates this failure.

[1] https://cygwin.com/pipermail/cygwin-apps/2020-May/040109.html

PR #7787, opened 2020-09-27
compilers: Detect invalid command line options for intel windows compilers

IanH0073:
Using has_arguments for msvc as a guide, add invalid argument detection to the intel windows compilers, based on the presence of diagnostic #10006 in stderr or stdout.

(The Intel compilers will continue to accept options with the same name that use the "-xx" form rather than "/xx".)

### DIFF
--- a/mesonbuild/compilers/mixins/intel.py
+++ b/mesonbuild/compilers/mixins/intel.py
@@ -195,3 +195,11 @@ class IntelVisualStudioLikeCompiler(VisualStudioLikeCompiler):
 
     def get_optimization_args(self, optimization_level: str) -> T.List[str]:
         return self.OPTIM_ARGS[optimization_level]
+
+    def has_arguments(self, args: T.List[str], env: 'Environment', code: str, mode: str) -> T.Tuple[bool, bool]:
+        warning_text = 'command line warning #10006'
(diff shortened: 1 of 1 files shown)


dcbaker:
Are you actually going a case where has_argument doesn't work? It's supposed to be passing `/Qdiag-error:10006` which should make that a hard error

IanH0073:
Yes - for example with "test cases/fortran/1 basic", building with current Intel Fortran on Windows.  /Qdiag-error:10006 is not in the command line testing the validity of a particular command line argument.

I see an attempt to set /Qdiag-error:10006 and friends in intel.py:160 compiles(...), but this is not on the execution path for argument checking (compile**s** vs compile confusion?).

dcbaker:
Can you test #7795 and see if that resolves the issue for you? I've done a bit of refactoring to try to make it more obvious where to put extra args for compiler testing in that series. I haven't gotten out windows to check, but with those changes the tests work correctly on ifort for linux.

IanH0073:
Yes - the move of the /Qdiag-error arguments from the previous intel.py compiles(...) into intel.py get_compiler_check_args(...) means that argument checking now works with ifort on Windows.

PR #7791, opened 2020-09-27
Add win_subsystem kwarg. Closes #7765.

### LINKED ISSUE #7765: Meson does not support Windows kernel drivers
**Describe the bug**
Meson assumes a binary state of affairs when building Windows applications based on
the `gui_app` kwarg: either you're building a GUI app or you're building a console app.
This is translated to either a `/SUBSYSTEM:WINDOWS` or a `/SUBSYSTEM:CONSOLE`
in the ninja files respectively. The reality is that `link` supports a variety of different
subsystems ranging from kernel drivers to EFI applications and drivers which are not buildable
with meson because of this.

**To Reproduce**
I don't think this requires any specific project files for reproduction as the outlined behavior
should be apparent by opening the generated `build.ninja` of any Windows application project.

**Expected behavior**
Meson features a `subsystem` kwarg in `executable` that allows the user to select between
one of the subsystems supported by `link`.

**system parameters**
* Native build targeting Windows
* Windows
* Python version 3.8.6
* Meson version 0.55.3
* Ninja version 1.10.1

jpakkane:


mesonbuild/compilers/mixins/gnu.py
```diff
@@ -219,6 +219,17 @@ def get_gui_app_args(self, value: bool) -> T.List[str]:
             return ['-mwindows' if value else '-mconsole']
         return []
 
+    def get_win_subsystem_args(self, value: str) -> T.List[str]:
+        if self.info.is_windows() or self.info.is_cygwin():
+            if 'windows' in value:
+                return ['-mwindows']
+            elif 'console' in value:
+                return ['-mconsole']
```
mensinda, line 227: Shouldn't there be a warning for something like `windows,6.0`, since the `6.0` part is ignored, or is this irrelevant for MinGW?

mesonbuild/compilers/compilers.py
```diff
@@ -877,6 +877,9 @@ def gnu_symbol_visibility_args(self, vistype: str) -> T.List[str]:
     def get_gui_app_args(self, value: bool) -> T.List[str]:
         return []
 
+    def get_win_subsystem_args(self, value: str) -> T.List[str]:
+        return []
```
dcbaker, line 881: Wouldn't it make more sense to raise an `EnvironmentException` here? Since not setting the value is likely to lead to someone getting unexpected results?

mesonbuild/compilers/mixins/gnu.py
jpakkane: GCC documentation does not say anything about specifying a version. So I don't really know. I'm guessing people who care about this detail will only use MSVC anyway.

jon-turney: These things are controllable via `ld` flags: see `--subsystem`, `--major-subsystem-version`, `--minor-subsystem-version` in `man ld`. 

If you look at the output of `gcc -dumpspecs` you can see that `-mwindows` is a convenient shortcut for `--subsystem windows` and some libraries.

test cases/windows/16 gui app/meson.build
```diff
@@ -6,13 +6,15 @@ project('gui_app_test', 'c')
 #
 
 console_lib = static_library('main', 'console_prog.c')
-executable('console', 'dummy.c', link_with: console_lib, gui_app: false)
+executable('console', 'dummy.c', link_with: console_lib, win_subsystem: 'console')
+executable('console2', 'dummy.c', link_with: console_lib, gui_app: false)
```
jon-turney, line 10: perhaps `tester` should be run on the output executable to verify the subsystem is set correctly.

docs/markdown/Reference-manual.md
```diff
@@ -673,6 +674,12 @@ be passed to [shared and static libraries](#library).
 - `pie` *(since 0.49.0)*: build a position-independent executable
 - `native`: is a boolean controlling whether the target is compiled for the
   build or host machines. Defaults to false, building for the host machine.
+- `win_subsystem` *(since 0.56.0)* specifies the subsystem type to use
+  on the Windows platform. Typical values include `console` for text
```
jon-turney, line 678: Unclear if numeric values are expected to work here, so perhaps `is a string specifying the subsystem...`?

test cases/windows/16 gui app/meson.build
jpakkane: AFAIK it should fail to link if you use WinMain on a console app or plain main on a Windows app.

mesonbuild/compilers/mixins/gnu.py
jpakkane: Rewriting this but currently msys download page is unresponsive so I can't actually test it locally but instead have to use CI. :(

jpakkane:
That should be all done now. @dcbaker this probably means that you need to add some sort of a gui kwarg to the freedesktop module. Is that ok?

dcbaker:
Yeah, that's fine. I was just sorta piggy backing on it because it was convenient.

bonzini:
The instant deprecation is annoying, it triggers about 60 times when building QEMU. Because 0.55.x doesn't have the `win_subsystem` kwarg at all, it means we cannot support 0.55.x and 0.56.x at the same time without playing games with `kwargs` (and frankly those games are _not_ why I choose Meson for my project).

Now, for QEMU the situation is not _that_ bad: we distribute our own copy of Meson (which makes it much easier to stay on the bleeding edge), we haven't even had a release yet that required Meson, and we probably want to adopt 0.56.0 ASAP anyway due to #7760. But then, distros may not want to use the embedded QEMU, and in any case we want to work with upstream as much as possible to share our pain points even for the future.

I'll create a pull request to restrict the deprecation message to `meson_version: '>=0.56.0'` for a couple releases, but long term there should be at least three different kinds of deprecation:

* cases where Meson is doing something that will break immediately (e.g. also in 0.56.0, replacing colons in test names)

* cases where something willl be removed in the future because it was a bad idea

* cases where the syntax has a simple replacement.

For the last two, it's probably time for Meson to have a formal deprecation policy. The policy should specify:

* a minimum period between warning and removal

* a minimum preriod between introducing the alternative and warning (unless a high-enough `meson_version` is specified).

mensinda:
> The instant deprecation is annoying, it triggers about 60 times when building QEMU.

Would the situation improve if all deprecations were only logged once by default? I understand that the situation is suboptimal, but I see no reason *not* to deprecate an old function immediately if there is a new and better alternative. As for restricting the warning to `meson_version: '>=0.56.0'`, I think this is a good idea, however, this restriction should be dropped after `<insert number>` of releases or `<insert number>` of releases before the feature is removed.

bonzini:
> I see no reason not to deprecate an old function immediately if there is a new and better alternative

Yeah in the end I agree and `@FeatureDeprecatedKwargs` does what I need, so I used it in #7872. Making the warning independent of `meson_version()` can be done later (possibly by not passing any version to the decorator?).

> this restriction should be dropped after `<insert number>` of releases or `<insert number>` of releases before the feature is removed.

That `<insert number>` is exactly why having a policy would be great. :)

PR #7792, opened 2020-09-27
esclude generate file from clang-tidy processing

mdionisio:
by default run_clang_tidy process al file in compile_commands.json but
the file generated has to be esclude like already done from
manual_clangformat

### DIFF
--- a/mesonbuild/scripts/clangtidy.py
+++ b/mesonbuild/scripts/clangtidy.py
@@ -15,6 +15,8 @@
 import pathlib
 import subprocess
 import shutil
+import os
+import re
 from concurrent.futures import ThreadPoolExecutor
 import typing as T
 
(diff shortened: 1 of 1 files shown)


jpakkane:
This is a bit problematic because there are people who want to run clang-tidy on their generated code to find bugs, particularly the people who write the generators.

mdionisio:
In my option there are more people that want to validate its own code than people that want to validate the generated code.
But In any case if the scope of this modules is to validate any code, also the already present function 
`def manual_clangformat(srcdir_name: str, builddir_name: str) -> int:`
is buggy because file in build folder is skipped
`if strf.startswith(builddir_name):
                continue`

jpakkane:
Let's do this now for consistency with clang-format. If someone files an issue, we can create a more complex solution.

PR #7794, opened 2020-09-28
typing: CMake module

mensinda:
This PR fully annotates the meson CMake module. Additionally, there are some minor changes in `run_mypy.py` in the first two commits.

### DIFF
--- a/mesonbuild/cmake/client.py
+++ b/mesonbuild/cmake/client.py
@@ -17,7 +17,6 @@
 
 from .common import CMakeException, CMakeConfiguration, CMakeBuildFile
 from .executor import CMakeExecutor
-from ..environment import Environment
 from ..mesonlib import MachineChoice
 from .. import mlog
 from contextlib import contextmanager
@@ -25,6 +24,9 @@ from subprocess import Popen, PIPE, TimeoutExpired
 import typing as T
 import json
 import os
+
+if T.TYPE_CHECKING:
+    from ..environment import Environment
 
 CMAKE_SERVER_BEGIN_STR = '[== "CMake Server" ==['
 CMAKE_SERVER_END_STR = ']== "CMake Server" ==]'
@@ -36,7 +38,7 @@ CMAKE_MESSAGE_TYPES = {
     'progress': ['cookie'],
     'reply': ['cookie', 'inReplyTo'],
     'signal': ['cookie', 'name'],
-}
+}  # type: T.Dict[str, T.List[str]]
 
 CMAKE_REPLY_TYPES = {
     'handshake': [],
@@ -44,16 +46,16 @@ CMAKE_REPLY_TYPES = {
     'compute': [],
     'cmakeInputs': ['buildFiles', 'cmakeRootDirectory', 'sourceDirectory'],
     'codemodel': ['configurations']
-}
+}  # type: T.Dict[str, T.List[str]]
 
 # Base CMake server message classes
 
 class MessageBase:
-    def __init__(self, msg_type: str, cookie: str):
+    def __init__(self, msg_type: str, cookie: str) -> None:
         self.type = msg_type
         self.cookie = cookie
 
-    def to_dict(self) -> dict:
+    def to_dict(self) -> T.Dict[str, T.Union[str, T.List[str], T.Dict[str, int]]]:
         return {'type': self.type, 'cookie': self.cookie}
 
     def log(self) -> None:
@@ -62,21 +64,21 @@ class MessageBase:
 class RequestBase(MessageBase):
     cookie_counter = 0
 
-    def __init__(self, msg_type: str):
+    def __init__(self, msg_type: str) -> None:
         super().__init__(msg_type, self.gen_cookie())
 
     @staticmethod
-    def gen_cookie():
+    def gen_cookie() -> str:
         RequestBase.cookie_counter += 1
         return 'meson_{}'.format(RequestBase.cookie_counter)
 
 class ReplyBase(MessageBase):
-    def __init__(self, cookie: str, in_reply_to: str):
+    def __init__(self, cookie: str, in_reply_to: str) -> None:
         super().__init__('reply', cookie)
         self.in_reply_to = in_reply_to
 
 class SignalBase(MessageBase):
-    def __init__(self, cookie: str, signal_name: str):
+    def __init__(self, cookie: str, signal_name: str) -> None:
         super().__init__('signal', cookie)
         self.signal_name = signal_name
 
@@ -86,7 +88,7 @@ class SignalBase(MessageBase):
 # Special Message classes
 
 class Error(MessageBase):
-    def __init__(self, cookie: str, message: str):
+    def __init__(self, cookie: str, message: str) -> None:
         super().__init__('error', cookie)
         self.message = message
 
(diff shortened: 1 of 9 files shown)


mensinda:
The CI error is unrelated

mesonbuild/cmake/client.py
```diff
@@ -62,21 +64,21 @@ def log(self) -> None:
 class RequestBase(MessageBase):
     cookie_counter = 0
 
-    def __init__(self, msg_type: str):
+    def __init__(self, msg_type: str) -> None:
```
dcbaker, line 67: we should really decide what we want to do with `__init__`, mypy assumes that the return type is `None`, the only case you need the return annotation is for cases like `__init__(self) -> None`, where there are no arguments, and thus mypy can't infer that the method is supposed to be typed.

mesonbuild/cmake/client.py
```diff
@@ -103,19 +105,21 @@ def log(self) -> None:
...
-    def __init__(self, cookie: str):
+    def __init__(self, cookie: str) -> None:
         super().__init__('progress', cookie)
 
     def log(self) -> None:
         pass
 
 class MessageHello(MessageBase):
-    def __init__(self, supported_protocol_versions: T.List[dict]):
+    def __init__(self, supported_protocol_versions: T.List[T.Dict[str, int]]) -> None:
```
dcbaker, line 115: Could you add a TODO comment here about using a typed dict?

mesonbuild/cmake/client.py
```diff
@@ -315,15 +323,15 @@ def resolve_type_reply(self, data: dict) -> ReplyBase:
...
+    def resolve_reply_cmakeInputs(self, data: T.Dict[str, T.Any]) -> ReplyCMakeInputs:
         files = []
         for i in data['buildFiles']:
             for j in i['sources']:
                 files += [CMakeBuildFile(j, i['isCMake'], i['isTemporary'])]
         return ReplyCMakeInputs(data['cookie'], data['cmakeRootDirectory'], data['sourceDirectory'], files)
 
     @contextmanager
-    def connect(self):
+    def connect(self) -> T.Generator[None, None, None]:
```
dcbaker, line 334: This can be simplified as `T.Iterator[None]`, but using the Generator is fine too. (Generator is an Iterator I think)

mesonbuild/cmake/client.py
mensinda: I would vote for always use `-> None` because it is required for all other functions with no return type and it is also sometimes required for `__init__` itself. So the most consistent thing we could do would be to always use `-> None` IMHO.

mesonbuild/cmake/common.py
```diff
@@ -111,22 +119,22 @@ def log(self) -> None:
...
+        self.install                 = data.get('hasInstallRule', False)                    # type: bool
+        self.install_paths           = list(set(data.get('installPaths', [])))              # type: T.List[str]
+        self.link_lang               = data.get('linkerLanguage', '')                       # type: str
+        self.link_libraries          = _flags_to_list(data.get('linkLibraries', ''))        # type: T.List[str]
+        self.link_flags              = _flags_to_list(data.get('linkFlags', ''))            # type: T.List[str]
+        self.link_lang_flags         = _flags_to_list(data.get('linkLanguageFlags', ''))    # type: T.List[str]
+        # self.link_path             = data.get('linkPath', '')                             # type: str
+        self.type                    = data.get('type', 'EXECUTABLE')                       # type: str
+        # self.is_generator_provided = data.get('isGeneratorProvided', False)               # type: bool
+        self.files                   = []                                                   # type: T.List[CMakeFileGroup]
```
dcbaker, line 137: I'm not crazy about lining up `=` signs like this, I don't think it improves readability (as opposed to say lining up data in lists of tuples or dicts), but... I don't care too much so if you like it leave it.

mesonbuild/cmake/interpreter.py
```diff
@@ -185,8 +189,10 @@ def artifact(self, name: str) -> T.Optional[T.Union['ConverterTarget', 'Converte
             keys += [self._rel_artifact_key(i), os.path.basename(i), self._base_artifact_key(i)]
         return self._return_first_valid_key(keys)
 
-    def generated(self, name: str) -> T.Optional[T.Union['ConverterTarget', 'ConverterCustomTarget']]:
-        return self._return_first_valid_key([self._rel_generated_file_key(name), self._base_generated_file_key(name)])
+    def generated(self, name: str) -> T.Optional['ConverterCustomTarget']:
+        res = self._return_first_valid_key([self._rel_generated_file_key(name), self._base_generated_file_key(name)])
+        assert res is None or isinstance(res, ConverterCustomTarget)
```
dcbaker, line 194: Is this assertion really needed here? Shouldn't `_return_first_valid_key` ensure that instead?

mesonbuild/cmake/interpreter.py
```diff
@@ -958,71 +979,72 @@ def analyse(self) -> None:
...
+                    self.project_name = j_0.name
+                for k_0 in j_0.targets:
                     # Avoid duplicate targets from different configurations and known
                     # dummy CMake internal target types
-                    if k.type not in skip_targets and k.name not in added_target_names:
-                        added_target_names += [k.name]
-                        self.targets += [ConverterTarget(k, self.env)]
+                    if k_0.type not in skip_targets and k_0.name not in added_target_names:
+                        added_target_names += [k_0.name]
+                        self.targets += [ConverterTarget(k_0, self.env)]
```
dcbaker, line 991: gahh, the lack of scoping in python makes this so painful...

dcbaker (changes requested):
I had a few nits, but only the one about the assert that I really want to address, the other's are "take it or leave it" kinda comments.

mesonbuild/cmake/client.py
mensinda: This data directly comes from CMake and is never really checked in any way (otherwise there would have to be *many* checks and the entire CMake code would be twice as long). We are trusting that CMake always returns the right types here.

I guess it won't hurt, but I am also not sure that we gain anything from it either.

mesonbuild/cmake/common.py
mensinda: It really improved the readability for me though. Not aligning is not big deal for one or two assignments, but helps a lot if there is a huge block.

I guess this is just a difference in taste :)

mesonbuild/cmake/interpreter.py
mensinda: I agree. I already tried coming up with better names in most instances but decided to give up here and just use numbers instead...

mesonbuild/cmake/interpreter.py
mensinda: Yes. `_return_first_valid_key` is just a fancy dict lookup (for a dict of type `T.Dict[str, T.Union['ConverterTarget', 'ConverterCustomTarget']]`) and does no further checking.

This assertion works because of the `add` logic, which ensures that only `ConverterCustomTarget` objects are stored for any keys generated in the `generated` function.

mensinda: So, in short, `_return_first_valid_key` can also return both types in other instances, but it would be a bug (and blow up further down the line anyway) if it were to return a `ConverterTarget` in the scope of `generated`.

I have to admit that this is not the best piece of software I have ever designed...

dcbaker: ahhh, okay. that makes sense.

mesonbuild/cmake/client.py
mensinda: I would leave it as `Generator` because that is also what is used in `nested` in `mlog.py` and it really shouldn't make any difference.

dcbaker (approved):

PR #7795, opened 2020-09-28
Full annotations for the Compiler package

dcbaker:
Mainly this adds full annotations to the compiler package, though there are a few refactors to make things easier. The biggest changes I think are those to the `OptionOverrideProxy` to make it a full replacement for `Dict`, and some refactoring around the `CLikeCompiler._build_wrapper`, which I've moved down into Compiler and added a `get_compiler_check_args` method, which is the hook to add extra arguments to compile checks.


mesonbuild/compilers/c.py
```diff
@@ -129,16 +146,17 @@ def get_options(self):
...
         if std.value != 'none':
             args.append('-std=' + std.value)
         return args
 
-    def get_option_link_args(self, options):
+    def get_option_link_args(self, options: 'OptionDictType') -> T.List[str]:
         if self.info.is_windows() or self.info.is_cygwin():
-            return options['winlibs'].value[:]
+            # without a typedict mypy can't understand this.
+            return T.cast(T.List[str], options['winlibs'].value[:])
```
mensinda, line 159: Maybe use something like this here (and in all other instances) instead?

```suggestion
            res = options['winlibs'].value[:]
            assert isinstance(res, list)
            for i in res:
                assert isinstance(i, str)
            return res
```

Or would the impact of the explicit check be too significant here?

mesonbuild/compilers/compilers.py
```diff
@@ -1047,6 +1051,148 @@ def get_has_func_attribute_extra_args(self, name: str) -> T.List[str]:
     def name_string(self) -> str:
         return ' '.join(self.exelist)
 
+    @abc.abstractclassmethod
```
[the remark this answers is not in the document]
dcbaker: That looks like an autocomplete error on my part

mesonbuild/compilers/compilers.py
```diff
@@ -1047,6 +1051,148 @@ def get_has_func_attribute_extra_args(self, name: str) -> T.List[str]:
...
+        raise EnvironmentError('{} does not implement get_optimization_args'.format(self.id))
+
+    def get_module_incdir_args(self) -> T.Tuple[str, ...]:
+        raise EnvironmentError('{} does not implement get_module_incdir_args'.format(self.id))
+
+    def get_module_outdir_args(self, path: str) -> T.List[str]:
+        raise EnvironmentError('{} does not implement get_module_outdir_args'.format(self.id))
+
+    def module_name_to_filename(self, module_name: str) -> str:
+        raise EnvironmentError('{} does not implement module_name_to_filename'.format(self.id))
```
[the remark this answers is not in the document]
dcbaker: The *module* methods are only implemented by fortran, so I'd rather not implement them for every langague. But I have changed `get_optimization_args` to be abstract, it looks like java is the only language that doesn't implement that.

mesonbuild/compilers/c.py
dcbaker: I'm not sure what the right thing to do is here. I'd really like to use a TypedDict for this, but the way we dynamically construct that list makes it rather difficult to do that.

mensinda: I am also not 100% certain what is better, but with the asserts the types are at least checked and it will blow up immediately if something changes.

dcbaker: I went ahead and changed to asserts, it's not a particularly hot path so it shouldn't be a problem to get a little bit more safety.

dcbaker:
The last ci failure looks unrelated

mensinda (approved):
LGTM

PR #7796, opened 2020-09-28
CODEOWNERS: don't use * [skip ci]

dcbaker:
strangely enough the syntax is:
/foo/* (files in foo, but not in foo/dir/)
/foo/ (files in foo and any subfolder of foo (recursively)


xclaesse:
Good to know thanks

PR #7798, opened 2020-09-29
cmake: switch to pathlib (fixes #7322)

### LINKED ISSUE #7322: [CMake] subproject panic when including non "C:\" libraries
**Describe the bug**
CMake subprojects will panic when trying to include libraries
not installed to the `C:` drive

**To Reproduce**
build a CMake subproject, make sure it has a dependency
not on the `C:` drive. Discovery of bug included building `LibKF5Attica`
as a subproject, with `QT5` as a dependency on drive `D:`. 
Native CMake build completed successfully with this setup.

**Expected behavior**
Build the subproject.

**system parameters**
native build
Windows 10 (Current latest patch KB4538461)
python version 3.7.1
meson version 0.54.1
ninja Not Applicable.

mensinda:
This PR removes all `os.path.*` calls from `mesonbuild.cmake` and uses `Path` objects instead.

I have added this to `0.56.0` since #7322 should automatically be fixed with this.

### DIFF
--- a/mesonbuild/_pathlib.py
+++ b/mesonbuild/_pathlib.py
@@ -1,0 +1,49 @@
+# Copyright 2020 The Meson development team
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sys
+import typing as T
+
+# Python 3.5 does not have the strict kwarg for resolve and always
+# behaves like calling resolve with strict=True in Python 3.6+
+#
+# This module emulates the behavior of Python 3.6+ by in Python 3.5 by
+# overriding the resolve method with a bit of custom logic
+#
+# TODO: Drop this module as soon as Python 3.5 support is dropped
+
+if T.TYPE_CHECKING:
+    from pathlib import Path
+else:
+    if sys.version_info.major <= 3 and sys.version_info.minor <= 5:
+
+        # Inspired by https://codereview.stackexchange.com/questions/162426/subclassing-pathlib-path
+        import pathlib
+        import os
+
+        # Can not directly inherit from pathlib.Path because the __new__
+        # operator of pathlib.Path() returns a {Posix,Windows}Path object.
+        class Path(type(pathlib.Path())):
+            def resolve(self, strict: bool = False) -> 'Path':
+                try:
+                    return super().resolve()
+                except FileNotFoundError:
+                    if strict:
+                        raise
+                    return Path(os.path.normpath(str(self)))
+
+    else:
+        from pathlib import Path
+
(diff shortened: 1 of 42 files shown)


mensinda:
The LGTM errors are intentional. I know, the entire `_pathlib.py` module is a bit of a hack, but this is the only way to make `resolve()` work in Python 3.5...

mensinda:
The Travis CI error is unrelated since MacOS fails during setup (again).

mesonbuild/_pathlib.py
```diff
@@ -0,0 +1,50 @@
...
+#
+# This module emulates the behavior of Python 3.6+ by in Python 3.5 by
+# overriding the resolve method with a bit of custom logic
+#
+# TODO: Drop this module as soon as Python 3.5 support is dropped
+
+if T.TYPE_CHECKING:
+    from pathlib import Path
+else:
+    if sys.version_info.major <= 3 and sys.version_info.minor <= 5:
```
dcbaker, line 29: I think this can just be  `if sys.version_info <= (3, 5)`

mesonbuild/_pathlib.py
```diff
@@ -0,0 +1,50 @@
...
+
+        # Inspired by https://codereview.stackexchange.com/questions/162426/subclassing-pathlib-path
+        # from pathlib import Path as _Path_, _windows_flavour, _posix_flavour
+        import pathlib
+        import os
+
+        class Path(type(pathlib.Path())):
```
dcbaker, line 36: Why this this not just `Path(pathlib.Path)`

dcbaker (approved):
I've got a couple of questions about the new module, and I only skimmed the cmake changes, but it looks pretty good to me

mesonbuild/_pathlib.py
mensinda: Because `pathlib.Path` is special... The `__new__` operator of `pathlib.Path` creates either a `PosixPath` or a `WindowsPath` object. Inheriting from `pathlib.Path` directly can work if you replicate the `__new__` operator which seems way messier to me.

mesonbuild/_pathlib.py
mensinda: At least not in this form. Comparing with a tuple is apparently not that straightforward...

```
Python 3.5.10 (default, Sep  6 2020, 09:43:22) 
[GCC 10.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> sys.version_info <= (3, 5)
False
>>> sys.version_info <= (3, 4)
False
>>> sys.version_info <= (3, 6)
True
>>> sys.version_info <= (3, 5, 10)
False
>>> sys.version_info <= (3, 5, 11)
True
```

So we would have to compare to `(3, 5, 999)` which is worse than the explicit `major` and `minor` solution IMHO because of the magic number. I am not 100% sure if `< (3, 6)` will always work, so I would like to keep the current version since it has the least potential for surprises.

mesonbuild/_pathlib.py
dcbaker: Ahhh, okay. I hadn't looked at the implementation of path lib, I assumed it was like os.path where Path was implemented in private modules and selectively imported. Can you add a comment here? With that I think this is good to go

mesonbuild/_pathlib.py
dcbaker: Okay

mesonbuild/_pathlib.py
mensinda: Done, thanks for the review.

dcbaker (approved):

PR #7803, opened 2020-10-01
deprecated get_configtool_variable and get_pkgconfig_variable

dcbaker:
The get_variable method is able to do everything they do and more,
making it generally more useful. Let's tell people to stop using the old
ones.


xclaesse:
LGTM, thanks.

PR #7806, opened 2020-10-01
linkers: Drop -pie on macOS

roolebo:
Projects that specify b_pie=true clutter build logs with the warning:

  clang: warning: argument unused during compilation: '-pie' [-Wunused-command-line-argument]

No option is needed to produce PIE binaries because ld64 is making PIE
executables on 10.7 and above by default, as documented in ld(1).

Signed-off-by: Roman Bolshakov <r.bolshakov@yadro.com>


jpakkane:
Thanks.

PR #7812, opened 2020-10-03
Never run clang-format / clang-tidy against directories

tryone144:
Currently the auto-generated `clang-format` target fails when the source directory contains a directory matching one of the `lang_suffixes`, e.g. a directory named `test.h/`. This is due to `pathlib.Path.glob()` not only returning files but also returning directories.

Both `clang-format` and `clang-tidy` fail when handed a directory. This PR explicitly checks if the "file" to check is actually a directory and skips calling `clang-format` and `clang-tidy` for those. Unit tests have been adapted accordingly.

### DIFF
--- a/run_unittests.py
+++ b/run_unittests.py
@@ -4212,9 +4212,11 @@ recommended as it is not supported on some platforms''')
         if is_osx():
             raise unittest.SkipTest('Apple ships a broken clang-tidy that chokes on -pipe.')
         testdir = os.path.join(self.unit_test_dir, '70 clang-tidy')
+        dummydir = os.path.join(testdir, 'dummydir.h')
         self.init(testdir, override_envvars={'CXX': 'c++'})
         out = self.run_target('clang-tidy')
         self.assertIn('cttest.cpp:4:20', out)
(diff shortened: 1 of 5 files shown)


run_unittests.py
```diff
@@ -4212,9 +4216,16 @@ def test_clang_tidy(self):
...
+        headerdir = os.path.join(testdir, 'dir.h')
+        try:
+            os.mkdir(headerdir)
+            self.init(testdir, override_envvars={'CXX': 'c++'})
+            out = self.run_target('clang-tidy')
+            self.assertIn('cttest.cpp:4:20', out)
+            self.assertNotIn(headerdir, out)
+        finally:
+            if os.path.exists(headerdir):
+                os.rmdir(headerdir)
```
jpakkane, line 4228: Tests should not modify the source directory. Create the directory and add a `dummy.dat` file that says something like "This file only exists here to get its enclosing directory into Git." and add it to Git instead.

tryone144: Updated both testcases to include static directory `dummydir.h`.

jpakkane:
Thanks.

Issue #589, opened 2016-06-12
xcode generator fails when using files object


theothermatt:
When using a files object (created with files('src.c',...)), the xcode generator fails with the following error:

```
Traceback (most recent call last):
  File "/usr/local/Cellar/meson/0.31.0/lib/python3.5/site-packages/mesonbuild/mesonmain.py", line 254, in run
    app.generate()
  File "/usr/local/Cellar/meson/0.31.0/lib/python3.5/site-packages/mesonbuild/mesonmain.py", line 158, in generate
    g.generate(intr)
  File "/usr/local/Cellar/meson/0.31.0/lib/python3.5/site-packages/mesonbuild/backend/xcodebackend.py", line 88, in generate
    self.generate_pbx_build_file()
  File "/usr/local/Cellar/meson/0.31.0/lib/python3.5/site-packages/mesonbuild/backend/xcodebackend.py", line 234, in generate_pbx_build_file
    idval = self.buildmap[s]
KeyError: 'tests/fileA.c'
```

I've set up a sample repo, which this error is from, at https://github.com/theothermatt/mesontest.
In this example, `fileA.c` is not in the `tests` directory as the error suggests, it's in the `src` directory, where a `meson.build` file creates a files object.

This works when using the ninja or vs2010 backends.

Meson version:  0.31.0

jpakkane:
Unfortunately the XCode backend lags a bit behind the other two. Mostly because its file format is awful enough to make your eyes bleed blood.

nirbheek:
… but also because no one has stepped up to maintain it. @nioncode has been working on the Visual Studio backends and they're in much better shape in comparison. So basically, help is appreciated for better health of this backend!

mensinda:
Fixed with #7813.

PR #7813, opened 2020-10-04
Fix xcode subdir files

OskarSigvardsson:
I ran into bug #589 recently, and I looked into it and the fix is simple enough: when converting from a `mesonlib.File` object to a string, it didn't take subdirs into account, which is fixed by doing a `os.path.join(s.subdir, s.fname)`. 

I looked through the test case folder and couldn't find anything in `common/` that specifically tested for this (though I might have just missed it, there's a lot of tests in there), so I added a new test for this specific bug. 

This is my first time contributing to meson, please let me know if I did anything wrong :)

### DIFF
--- a/mesonbuild/backend/xcodebackend.py
+++ b/mesonbuild/backend/xcodebackend.py
@@ -264,7 +264,7 @@ class XCodeBackend(backends.Backend):
 
             for s in t.sources:
                 if isinstance(s, mesonlib.File):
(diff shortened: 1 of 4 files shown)


mensinda (approved):
> This is my first time contributing to meson, please let me know if I did anything wrong :)

Thanks, everything looks OK (and the CI timeouts are unrelated).

PR #7817, opened 2020-10-06
compilers: Enable C++20 for Intel C++ Compiler.

vinsonlee:
Intel C++ Compiler 19.1 has C++20 features.

https://software.intel.com/content/www/us/en/develop/articles/intel-c-compiler-191-for-linux-release-notes-for-intel-parallel-studio-xe-2020.html

Signed-off-by: Vinson Lee <vlee@freedesktop.org>


jpakkane:
Thanks.

PR #7818, opened 2020-10-06
python.dependency() is not respecting 'required' kwarg

xclaesse:


dcbaker (approved):

xclaesse:
CI failure is unrelated.

mensinda:
This PR seems to have broken `PythonTests.test_versions` on `vc2017x86ninja`.

xclaesse:
Oops, you're right. That test assumes that requires is False by default which is inconsistent with dependency() function...  I think we should add `required: false`, is that fine with your?

xclaesse:
Should be fixed by https://github.com/mesonbuild/meson/pull/7821

PR #7821, opened 2020-10-07
Tests: py.dependency() now has required:true by default

xclaesse:
It used to ignore the required argument and got fixed to be consistent
with dependency() function.


mensinda (approved):
LGTM (if the relevant CI is green).

PR #7822, opened 2020-10-08
Replace ad-hoc find program functionality with centralized, cached mechanism

dcbaker:
This is an attempt to make a single, centralized find program mechanism, with caching and all that niceness that *isn't* part of the interpreter.

I'm fairly happy with the solution, though there are a few things that are a big ugly, and a few things I could use input on.

This is a draft because there's a bug I can't figure out, and one that is a design change that I don't know if it's safe.

The bug I can't figure out is somewhere in the ast code, and I can't make heads nor tails of it.

The behavioral change is that because programs are cached, overrides can cause the behavior of programs to change on reconfigure. I'm not sure if it's safe, and on reload I think we need to purge all overridden programs from the cache, but programs found through normal means can remain. @xclaesse, I know you've done most of the work on overrides, does that sound correct? (this is why test 28 and 187 are failing on Linux)


mesonbuild/programs.py
```diff
@@ -0,0 +1,489 @@
...
+
+    @functools.lru_cache()
+    def get_version(self) -> T.Optional[str]:
+        cmd = self.get_command() + self.version_arg
+        p, out, err = mesonlib.Popen_safe(cmd)
+        if p.returncode != 0:
+            raise mesonlib.MesonException('Failed running "{}"'.format(' ' .join(cmd)))
+        if not out:
+            out = err
+        match = re.search(r'([0-9][0-9\.]+)', out)
```
mensinda, line 311: That regex should be configurable and probably a bit more complex. I know that you are trying to solve this with `ExternalProgram*` but making the regex an optional parameter would make this more general and even simplify `ExternalProgramCMake` since the `get_version` basically only differs in the regex

See https://github.com/mesonbuild/meson/blob/master/run_project_tests.py#L1255-L1275 for existing version checks for instance.

mensinda (changes requested):
I really like that `find_program` is moved out of the interpreter and more centralized, but I don't like that it is added to `Environment` (which is already way to big and does way too much) instead.

jpakkane:
:crossed_swords: conflicts. And a lot of the tests fail. :(

dcbaker:
Yeah, I've been rebasing this locally and making fixes. Hopefully I'll have something ready or at least close enough to be reviewable today

dcbaker:
The commit history is still really messy, but I think I have all the bugs worked out (it works for me™, at least).

PR #7823, opened 2020-10-08
msubprojects: Handle change of URL in wrap-git

xclaesse:


mesonbuild/msubprojects.py
```diff
@@ -120,11 +120,33 @@ def update_git(wrap, repo_dir, options):
...
-        mlog.log('  -> No revision specified.')
+        mlog.log('  -> No revision or URL specified.')
         return True
+    try:
+        origin_url = git_output(['remote', 'get-url', 'origin'], repo_dir).strip()
+    except GitException as e:
+        mlog.log('  -> Failed to determine current origin URL in', mlog.bold(repo_dir))
+        mlog.log(mlog.red(e.output))
+        mlog.log(mlog.red(str(e)))
+        return False
```
nirbheek, line 136: How will this handle the case where a wrap switches from a tarball to git?

xclaesse: oh, haven't thought about that, but it's dangerous case. All git commands will apply to the parent project if parent is in git!

So we have to check if there is a .git file/dir in the subproject before doing anything. If not then it should print error message, that requires user intervention. I don't think we should `rm -rf` the directory and clone the new repo, there is a risk of data loss if the user modified anything.

xclaesse: @nirbheek updated this PR to skip update if it's not a git repo, unless --reset is specified in which case it deletes the whole directory and clone it again.

nirbheek:
Travis failure is spurious and also happening in master. mypy failure is fixed in master.

PR #7829, opened 2020-10-11
summary: Add from which subproject each subproject have been called

xclaesse:
meson

### DIFF
--- a/mesonbuild/interpreter/interpreter.py
+++ b/mesonbuild/interpreter/interpreter.py
@@ -999,7 +999,8 @@ class Interpreter(InterpreterBase, HoldableObject):
                 raise InterpreterException(f'Subproject {subp_name} version is {pv} but {wanted} required.')
         self.active_projectname = current_active
         self.subprojects.update(subi.subprojects)
-        self.subprojects[subp_name] = SubprojectHolder(subi, subdir, warnings=subi_warnings)
+        self.subprojects[subp_name] = SubprojectHolder(subi, subdir, warnings=subi_warnings,
+                                                       callstack=self.subproject_stack)
         # Duplicates are possible when subproject uses files from project root
         if build_def_files:
             self.build_def_files.update(build_def_files)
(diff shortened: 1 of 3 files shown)


xclaesse:
I find that summary a bit too much loaded, but really useful information. Anyone has a better idea to present this?

tp-m:
Looks ok to me, not sure how to make this 'lighter' without adding an option and not printing it by default (not sure that's warranted here though).

Alternative would be to print the 'from' hierarchy (e.g. `gst-plugins-good -> libsoup` for `libpsl`) in the main log when entering the subproject (don't remember if we do that already)

jpakkane:
Looks nice indeed. The only thing that is a bit weird is the `(from pango)` which is a bit confusing. Maybe `(requested by pango)`? It gets a bit long-winded, though.

xclaesse:
> Alternative would be to print the 'from' hierarchy (e.g. `gst-plugins-good -> libsoup` for `libpsl`) in the main log when entering the subproject (don't remember if we do that already)

I think we should do that too, yes. I was also wondering if I could add the hierarchy in summary too:
meson

xclaesse:
Or move the call stack in the key side:
meson

jpakkane:
Or maybe in tabular form so we can also add the info of a forwaring wrap file. Something like:

```
subproject            requested by          using from

foo                   main                        main
bar                   bar                          foo
```

Where bar is a subproject inside foo. Maybe "main" could even be left empty, so only non-standard ones are printed?

xclaesse:
Rebased this PR, it prints the summary like screenshot from https://github.com/mesonbuild/meson/pull/7829#issuecomment-706727226.

Any objection?

xclaesse:
ping?

xclaesse:
@jpakkane @eli-schwartz @tristan957 any objection?

tristan957:
What are your thoughts on the table-based output?

xclaesse:
> What are your thoughts on the table-based output?

IMHO it takes more space, a summary should be more compact. But that can be revisited later too I guess.

tristan957:
Then the design is fine with me!

PR #7833, opened 2020-10-11
ci: fix broken CI images

mensinda:
Updates the install scripts to account for changes in Arch and Fedora.

Also fixes the CUDA version detection in absence of a `/path/to/cuda/version.txt` file by looking for `CUDART_VERSION` in `cuda_runtime_api.h`.

Generally, it should be checked before each release that the last scheduled (Sunday night) [CI image builder job](https://github.com/mesonbuild/meson/actions?query=workflow%3A%22CI+image+builder%22) is green. A red job could indicate a failed test run in an up-to-date system.

### DIFF
--- a/ci/ciimage/.gitignore
+++ b/ci/ciimage/.gitignore
@@ -1,2 +1,3 @@
 /build_*
 /test_*
+/user.sh
--- a/ci/ciimage/arch/install.sh
+++ b/ci/ciimage/arch/install.sh
@@ -26,7 +26,7 @@ PACMAN_OPTS='--needed --noprogressbar --noconfirm'
 # Patch config files
 sed -i 's/#Color/Color/g'                            /etc/pacman.conf
 sed -i 's,#MAKEFLAGS="-j2",MAKEFLAGS="-j$(nproc)",g' /etc/makepkg.conf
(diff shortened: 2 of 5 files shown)


jpakkane:
There seems to be a linting error:

```
 tools/boost_names.py:51:54: error: Variable "builtins.NotImplemented" is not valid as a type
tools/boost_names.py:51:54: note: See https://mypy.readthedocs.io/en/latest/common_issues.html#variables-vs-type-aliases
tools/boost_names.py:56:54: error: Variable "builtins.NotImplemented" is not valid as a type
tools/boost_names.py:56:54: note: See https://mypy.readthedocs.io/en/latest/common_issues.html#variables-vs-type-aliases
tools/boost_names.py:74:54: error: Variable "builtins.NotImplemented" is not valid as a type
tools/boost_names.py:74:54: note: See https://mypy.readthedocs.io/en/latest/common_issues.html#variables-vs-type-aliases
```

mensinda:
This error was already fixed by #7834 and would go away after a rebase.

jpakkane:
LGTM, good to merge after a rebase so CI is green (apart from the Travis mac thing).

PR #7836, opened 2020-10-12
[RFC] mtest: use asyncio instead of concurrency.futures

bonzini:
This pull request rewrites the SingleTestRunner and TestHarness to use asyncio for subprocesses.

The main advantage is better control on cancellation; with the current code, KeyboardInterrupt was never handled by the thread executor so the code that tried to handle it in SingleTestRunner only worked for non-parallel tests.  Likewise, SIGTERM was not handled and left tests running. (In addition, if the test exited with exit status 0 in response to SIGTERM, it was not marked as a failure).

With this pull request, mtest gains a new `INTERRUPT` TestResult and SIGINT will return it for the oldest-running test. SIGTERM instead will return it for all running tests and will not start any new test. Because a picture is worth a thousand works, here are some Asciinema recordings:
* [sending SIGINT to `meson test` (unpatched)](https://asciinema.org/a/QdFqzfET6mhxZk6BcokKo1pDx)
* [sending SIGINT to `meson test` (patched)](https://asciinema.org/a/TwPwquCNiZFtfB7SjOzrAm0K1)
* [sending SIGTERM to `meson test` (patched)](https://asciinema.org/a/WsK001eLfvPueEzz7PESSDq2G)

The pull request is split between a refactoring part and a new feature part, so it can be split easily if desired.

### DIFF
--- a/mesonbuild/mtest.py
+++ b/mesonbuild/mtest.py
@@ -34,7 +34,6 @@ import subprocess
 import sys
 import tempfile
 import textwrap
-import threading
 import time
 import typing as T
 import xml.etree.ElementTree as et
@@ -680,7 +679,7 @@ class SingleTestRunner:
     async def _run_subprocess(self, args: T.List[str], *, timeout: T.Optional[int],
                               stdout: T.IO, stderr: T.IO,
                               env: T.Dict[str, str], cwd: T.Optional[str]) -> T.Tuple[int, TestResult, T.Optional[str]]:
-        async def kill_process(p: subprocess.Popen, f: asyncio.Future) -> T.Optional[str]:
+        async def kill_process(p: asyncio.subprocess.Process) -> T.Optional[str]:
             # Python does not provide multiplatform support for
             # killing a process and all its children so we need
             # to roll our own.
@@ -694,39 +693,30 @@ class SingleTestRunner:
 
                     # Make sure the termination signal actually kills the process
                     # group, otherwise retry with a SIGKILL.
-                    await try_wait_one(f, timeout=0.5)
-                    if f.done():
+                    await try_wait_one(p.wait(), timeout=0.5)
+                    if p.returncode is not None:
                         return None
 
                     os.killpg(p.pid, signal.SIGKILL)
 
-                await try_wait_one(f, timeout=1)
-                if f.done():
+                await try_wait_one(p.wait(), timeout=1)
+                if p.returncode is not None:
                     return None
 
                 # An earlier kill attempt has not worked for whatever reason.
                 # Try to kill it one last time with a direct call.
                 # If the process has spawned children, they will remain around.
                 p.kill()
-                await try_wait_one(f, timeout=1)
-                if f.done():
+                await try_wait_one(p.wait(), timeout=1)
+                if p.returncode is not None:
                     return None
                 return 'Test process could not be killed.'
             except ProcessLookupError:
                 # Sometimes (e.g. with Wine) this happens.  There's nothing
                 # we can do, probably the process already died so just wait
                 # for the event loop to pick that up.
-                await f
+                await p.wait()
                 return None
-
-        def wait(p: subprocess.Popen, loop: asyncio.AbstractEventLoop, f: asyncio.Future) -> None:
-            try:
-                p.wait()
-                loop.call_soon_threadsafe(f.set_result, p.returncode)
-            except BaseException as e:   # lgtm [py/catch-base-exception]
-                # The exception will be raised again in the main thread,
-                # so catching BaseException is okay.
-                loop.call_soon_threadsafe(f.set_exception, e)
 
         # Let gdb handle ^C instead of us
         if self.options.gdb:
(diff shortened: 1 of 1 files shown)


bonzini:
(According to CI there seems to be a race that I cannot figure out. Not a big deal for a generic review of the idea).

bonzini:
It seems like mtest occasionally never exits on Window I'll try to bisect to find what is wrong.

nirbheek:
You are my hero.

bonzini:
So it seems that asyncio.subprocess was the result of the breakage on Windows. Trying again without.

nirbheek:
> It seems like mtest occasionally never exits on Window I'll try to bisect to find what is wrong.

I ran the code in mtest-asyncio-full on the `test_suite_selection` unit test for 100 iterations and it didn't hang even once. Very strange that it hung on the CI.

mesonbuild/mtest.py
```diff
@@ -643,9 +673,107 @@ def run(self) -> TestRun:
...
+            except ProcessLookupError:
+                # Sometimes (e.g. with Wine) this happens.  There's nothing
+                # we can do, probably the process already died so just wait
+                # for the event loop to pick that up.
+                await f
+                return None
+
+        def communicate(p: subprocess.Popen, loop: asyncio.AbstractEventLoop, f: asyncio.Future) -> None:
+            try:
+                p.communicate()
```
dcbaker, line 722: why are you using communicate if you're not getting back the stdout or stderr? Wouldn't it be better to use `Popen.wait`?

bonzini: The coffee that was there before also used `communicate`, but yeah `wait` is better.

bonzini:
I don't have any more parts of the patches that I can bisect, but it looks like ProactorEventLoop is responsible for the issue. I'll wait for [this CI run](https://dev.azure.com/jussi0947/jussi/_build/results?buildId=8072) to finish, and then trigger another without it.

OTOH this failing run has Python 3.5.4, therefore the issue with ProactorEventLoop may fix itself once Windows CI is updated to 3.6.

scivision:
This is a side comment--bonzini is far more experienced with this than me.

My experience is with ~ 1000 users, mostly Windows, Python >= 3.6..3.8, and that ProactorEventLoop and subprocess worked well for that user group.  In my programs, I explicitly selected ProactorEventLoop for Windows via `if` statement.

### Question

Is it worth having (do we already have?) a Windows CI on Python 3.6, and another Windows CI on Python 3.9, since **Python 3.8 changed Windows [default event loop to Proactor](https://docs.python.org/3/library/asyncio-policy.html#asyncio.DefaultEventLoopPolicy)**. Python 3.6 and 3.7 [default event loop is Selector on Windows](https://docs.python.org/3.7/library/asyncio-policy.html#asyncio.DefaultEventLoopPolicy)

bonzini:
We would use Proactor anyway once we switch to a new-enough Python that fixed the bugs I was encountering. But yeah Windows should have one CI on the oldest possible Python version and one on a newer version.

mesonbuild/mtest.py
```diff
@@ -645,9 +675,107 @@ def run(self) -> TestRun:
...
+                # we can do, probably the process already died so just wait
+                # for the event loop to pick that up.
+                await f
+                return None
+
+        def wait(p: subprocess.Popen, loop: asyncio.AbstractEventLoop, f: asyncio.Future) -> None:
+            try:
+                p.wait()
+                loop.call_soon_threadsafe(f.set_result, p.returncode)
+            except BaseException as e:   # lgtm [py/catch-base-exception]
```
dcbaker, line 726: I think I know why you're catching `BaseException`, but since this is normally a bad thing, could you add a comment explaining why you're catching it.

dcbaker (approved):
I have on small nit, but otherwise this looks great. I'll be happy to merge this as soon as we get 0.56.0 out the door, and the bump to python 3.6 happens (which should be right after 0.57 development starts).

mesonbuild/mtest.py
bonzini: Sure. Note that if the CI passes I wouldn't mind going straight to the version that uses asyncio subprocesses, on which case the catching of BaseException disappears.

dcbaker: I'd really prefer to use asyncio.subprocess if it will work, just because it's going to be less maintenance for us. If the performance is worse, maybe we can consider using a faster event loop in CI, like the libuv one.

jpakkane:
Python 3.5 is dead in master now.

bonzini:
Looks like 3.6 is okay with SelectorEventLoop, I will add the commits that switch to Proctor once the run is over.

jpakkane:
Since I am paranoid and the CI system has been revamped, please rebase so the new CI runs this as well. Thanks.

bonzini:
Not sure why `test cases/d/9 features` is failing but it's the same on several of my other pull requests (sometimes for `unity=off` and sometimes for `unity=on`).

dcbaker:
I'm seeing that too. Macos just seems generally unreliable unfortunately

bonzini:
FWIW I have the next part ready which is to read logs via asyncio pipes instead of temporary files, and parsing TAP asynchronously (instead of waiting until the subprocess terminates). That alone doesn't add much value, but it is an important step towards reporting TAP results one by one; doing that for multiple concurrent tests would be almost impossible without asyncio. I might open a draft PR as soon as this one is merged, so as to get reviews.

PR #7837, opened 2020-10-12
pkgconfig: InternalDependency's ext_deps should be private by default

xclaesse:
This fixes a regression introduced by
https://github.com/mesonbuild/meson/pull/7488.

InternalDependency's ext_deps previously where simply ignored, but that
PR has effect to add many more public Requires in generated pc files.

### DIFF
--- a/mesonbuild/modules/pkgconfig.py
+++ b/mesonbuild/modules/pkgconfig.py
@@ -173,7 +173,7 @@ class DependenciesHelper:
         for t in link_whole_targets:
             self._add_link_whole(t, public)
         # And finally its external dependencies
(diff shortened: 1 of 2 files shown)


jpakkane:
Could a test be added to this?

xclaesse:
There is actually already a unit test, that asserts the opposite... that's why the CI failed. I need to update the test.

xclaesse:
@jpakkane test is now changed to assert that declare_dependency(..., dependencies: [foo]) goes into private instead of public.

xclaesse:
@jpakkane CI failure is unrelated.

PR #7841, opened 2020-10-13
gnome: generate cpp gresource source in cpp projects

### LINKED ISSUE #7839: gnome compile_resources doesn't work in C++ projects
**Describe the bug**
`gnome.compile_resources` form the GNOME module always generates C source files. Those files can't be added to C++ executables.

**To Reproduce**
```
project('example', 'cpp')

gnome = import('gnome')

gtkmm = dependency('gtkmm-3.0')

gresources = gnome.compile_resources('gresources', 'main.gresource.xml')

executable(meson.project_name(), gresources, 'main.cpp', dependencies: gtkmm)
```
Gives the following error:
```
$ meson _build
The Meson build system
Version: 0.55.3
Source dir: example
Build dir: example/_build
Build type: native build
Project name: example
Project version: undefined
C++ compiler for the host machine: ccache c++ (gcc 10.2.0 "c++ (Debian 10.2.0-13) 10.2.0")
C++ linker for the host machine: c++ ld.bfd 2.35.1
Host machine cpu: x86_64
Run-time dependency gtkmm-3.0 found: YES 3.24.2

meson.build:9:0: ERROR: No host machine compiler for "gresources.c"
```

**Expected behavior**
A C++ source file should be generated named `gresources.cpp` instead of the C source file. It's possible to do that with `run_command('glib-compile-resources', '--target=gresources.cpp', '--generate-source', 'main.gresource.xml')`.

**system parameters**
* Is this a cross build: No
* what operating system: `Debian GNU/Linux bullseye/sid`
* what Python version are you using: `Python 3.8.6`
* what `meson --version`: `0.55.3`
* what `ninja --version` if it's a Ninja build: `1.10.1`

mimi89999:
Closes #7839

### DIFF
--- a/mesonbuild/modules/gnome.py
+++ b/mesonbuild/modules/gnome.py
@@ -157,8 +157,14 @@ class GnomeModule(ExtensionModule):
             output = args[0] + '.gresource'
             name = args[0] + '_gresource'
         else:
-            output = args[0] + '.c'
-            name = args[0] + '_c'
+            if 'c' in state.environment.coredata.compilers.host.keys():
+                output = args[0] + '.c'
+                name = args[0] + '_c'
+            elif 'cpp' in state.environment.coredata.compilers.host.keys():
+                output = args[0] + '.cpp'
+                name = args[0] + '_cpp'
+            else:
+                raise MesonException('Compiling GResources into code is only supported in C and C++ projects')
 
         if kwargs.get('install', False) and not gresource:
             raise MesonException('The install kwarg only applies to gresource bundles, see install_header')


mesonbuild/modules/gnome.py
```diff
@@ -157,8 +157,12 @@ def compile_resources(self, state, args, kwargs):
             output = args[0] + '.gresource'
             name = args[0] + '_gresource'
         else:
-            output = args[0] + '.c'
-            name = args[0] + '_c'
+            if 'cpp' in state.project_args.keys():
```
xclaesse, line 160: I think this should rather check in `state.environment.coredata.compilers.keys()`. I'm not sure `state.project_args` contains anything unless the project does `add_project_arguments()`. Also I think it's preferable to still generate .c file in case both c and cpp compilers are available, I assume C is faster to compile, no?

jpakkane:
Is compiling with C++ something upstream officially supports (as opposed to being something that works)?

mesonbuild/modules/gnome.py
mimi89999: Checking `state.environment.coredata.compilers.keys()` won't work. `state.environment.coredata.compilers` contains several `OrderedDict` (two in my test). I don't know which one to choose.

xclaesse: Ah right, it's per-machine. Try: `state.environment.coredata.compilers.host.keys()`

mimi89999: Changed.
> Also I think it's preferable to still generate .c file in case both c and cpp compilers are available, I assume C is faster to compile, no?

Why? C is generally faster to compile, but for the resource file, the difference might not be noticeable. That requires testing. Also, generating the second resource source file might take longer than what you will gain on compile time. Then, there is also the issue with different flags for compilers...

jpakkane: There are at least two reasons. First of all C is more tested so using it is the more reliable option.

Second of all if we do use the C++ compiler and plain C people notice that, they will complain and file bugs. As an example suppose you have a plain C project that uses a subproject that is implemented in C++. In this case the resource file would be compiled with a C++ compiler even though the user probably does not want that. This also changes the linker from C to C++, which is again something people might not want. IIRC there are platforms where having an unnecessary dependency on libstdc++ is a big no-no.

xclaesse: > generating the second resource source

No, only generate one of course if you have a C compiler you generate .c and if not but have C++ then you generate .cpp instead.

> This also changes the linker from C to C++, which is again something people might not want. IIRC there are platforms where having an unnecessary dependency on libstdc++ is a big no-no.

Very true, it's indeed worse than I thought. For example gtk+ uses gnome.compile_resources() and has C++ compiler for a single unit test that checks that gtk headers works with C++ compiler, linking gtk on libstdc++ will definitely be rejected.

mimi89999:
How can this issue be solved then?
Check if there is cpp and not c?

jpakkane:
Check if C compiler exists. If yes, generate C. If not, check for C++ compiler. If it exists, use C++. Otherwise error out I guess?

mimi89999:
I updated my PR. The error message is not great, but I don't have better ideas...

mesonbuild/modules/gnome.py
```diff
@@ -157,8 +157,14 @@ def compile_resources(self, state, args, kwargs):
             output = args[0] + '.gresource'
             name = args[0] + '_gresource'
         else:
-            output = args[0] + '.c'
-            name = args[0] + '_c'
+            if 'c' in state.environment.coredata.compilers.host.keys():
+                output = args[0] + '.c'
+                name = args[0] + '_c'
+            if 'cpp' in state.environment.coredata.compilers.host.keys():
```
xclaesse, line 163: elif

xclaesse:
LGTM when CI passed, thanks for the contribution!

dcbaker:
Just the one macos test that fails in ci no one can replicate locally. So CI is green

PR #7843, opened 2020-10-13
A few fixups for rust

dcbaker:


mesonbuild/backend/ninjabackend.py
```diff
@@ -2793,7 +2794,7 @@ def generate_link(self, target, outname, obj_list, linker, extra_args=None, stdl
             # now have them. Once slashes are forbidden, remove this bit.
             target_slashname_workaround_dir = os.path.join(
                 os.path.dirname(target.name),
-                self.get_target_dir(target))
+                sela.get_target_dir(target))
```
nirbheek, line 2797: typo

nirbheek:
lgtm other than the typo, but maybe we want to add some tests that exercise this code?

dcbaker:
Hmmmm. I've been trying to come up with a good, generic way to test that arguments from the args end up in the backend. Unfortunately I'm having trouble coming up with something generic, since we check that all of those arguments are valid before passing them down. Do you have any ideas?

nirbheek:
I was thinking more that it would be nice to have the build failures anholt was seeing as project tests. Is that possible?

jpakkane:
LGTM, ok to merge when others are happy.

dcbaker:
Well some of anholt's problems are due to cross compiling with rust, and us handling the exe_wrapper incorrectly in 0.55.x (I fixed that in the typifying of the compiler package I think). Let me go look at his project again and see what I can do there.

mesonbuild/environment.py
```diff
@@ -1739,7 +1738,10 @@ def detect_rust_compiler(self, for_machine):
                 elif 'link' in override[0]:
                     linker = self._guess_win_linker(
                         override, RustCompiler, for_machine, use_linker_prefix=False)
+                    # rustc takes linker arguments undercorator, and does
```
anholt, line 1741: "undecorated"?

dcbaker: yeah, that should be "undecorated", but maybe "Without a prefix" would be better, since we generally refer to the (for example) `-Wl,` that gcc uses to forward linker args as a prefix.

jpakkane:
Is this ready now?

jpakkane:
```
>       raise EnvironmentException(msg)
E       mesonbuild.mesonlib.EnvironmentException: Unable to detect linker for compiler "gcc -fuse-ld=gold -Wl,--version"
E       stdout: 
E       stderr: collect2 version 10.2.1 20200825 [revision c0746a1beb1ba073c7981eb09f55b3d993b32e5c]
E       [cannot find ld] -plugin /usr/lib64/gcc/x86_64-suse-linux/10/liblto_plugin.so -plugin-opt=/usr/lib64/gcc/x86_64-suse-linux/10/lto-wrapper -plugin-opt=-fresolution=/tmp/ccol4ZPF.res -plugin-opt=-pass-through=-lgcc -plugin-opt=-pass-through=-lgcc_s -plugin-opt=-pass-through=-lc -plugin-opt=-pass-through=-lgcc -plugin-opt=-pass-through=-lgcc_s --build-id --eh-frame-hdr -m elf_x86_64 -dynamic-linker /lib64/ld-linux-x86-64.so.2 /usr/lib64/gcc/x86_64-suse-linux/10/../../../../lib64/crt1.o /usr/lib64/gcc/x86_64-suse-linux/10/../../../../lib64/crti.o /usr/lib64/gcc/x86_64-suse-linux/10/crtbegin.o -L/usr/lib64/gcc/x86_64-suse-linux/10 -L/usr/lib64/gcc/x86_64-suse-linux/10/../../../../lib64 -L/lib/../lib64 -L/usr/lib/../lib64 -L/usr/lib64/gcc/x86_64-suse-linux/10/../../../../x86_64-suse-linux/lib -L/usr/lib64/gcc/x86_64-suse-linux/10/../../.. --version -lgcc --push-state --as-needed -lgcc_s --pop-state -lc -lgcc --push-state --as-needed -lgcc_s --pop-state /usr/lib64/gcc/x86_64-suse-linux/10/crtend.o /usr/lib64/gcc/x86_64-suse-linux/10/../../../../lib64/crtn.o
E       collect2: fatal error: cannot find ‘ld’
E       compilation terminated.
```

jpakkane:
Should only need a text check fix any more:

```
test cases/failing/76 link with shared module on osx/meson.build:8:0: ERROR: target links against shared modules. This is not permitted on OSX
expected "test cases/failing/76 link with shared module on osx/meson.build:8:0: ERROR: target links against shared modules." not found in stdout 
```

Issue #7842, opened 2020-10-13
Failed link with boost using ldc2


Rogni:
**Describe the bug**
https://travis-ci.org/github/Rogni/MesonLinkWithBoostBug/builds/735436192#L307

error link boost using ldc2 in last versions meson (0.54, 0.55)

Error: 
```
ldc2  -of=app app.p/main.d.o app.p/external.cpp.o -L=--allow-shlib-undefined -link-defaultlib-shared /usr/lib/x86_64-linux-gnu/libboost_thread.so.1.65.1 -L=-lstdc++ -L=-lstdc++
Error: unrecognized file extension 1
```

**To Reproduce**
reproduced in versions 0.54 and 0.55
in version 0.53.2 boost linked as `LINK_ARGS = -L=-L--allow-shlib-undefined -L=--start-group -L=-lboost_thread -L=-lstdc++ -L=--end-group`

https://github.com/Rogni/MesonLinkWithBoostBug

meson.build
```
project('MesonLinkBoostBug', ['d', 'cpp'])
src = ['main.d', 'external.cpp']
boost_dep = dependency('boost', modules: 'thread')
executable('app', src, dependencies: boost_dep)
```

**Expected behavior**
maybe need append prefix `-L=` to `/usr/lib/x86_64-linux-gnu/libboost_thread.so.1.65.1`

**system parameters**
* native build
* Linux Mint 20 Ulyana
* Python 3.8.5
* what `meson --version` reproduced in versions 0.54.3, 0.55.3
* what `ninja --version`: 1.10.0.git.kitware.jobserver-1

dcbaker:
This is also broken with dmd, only gdc currently works.

mensinda:
Is this a general bug when passing absolute paths to libraries, or is this boost specific?

Rogni:
tried 
```
zlib = dependency('zlib', method: 'cmake')

executable('app', src, dependencies: zlib)
```

in build.ninja:
```
LINK_ARGS = -L=--allow-shlib-undefined -link-defaultlib-shared /usr/lib/x86_64-linux-gnu/libz.so -L=-lstdc++
```
app linked

dcbaker:
it has to do with getting version numbers, ldc and dmd expect the file to end in `.so` (or whatevert)  not `.so.X.Y.Z`, but if you pass them as `-L=mylib.so.1.2.3` that works. And it's really easy for the DMD and LDC to do that. so that's what I did my my PR.

PR #7845, opened 2020-10-13
compilers/d: Correctly pass shared libraries to DMD and LDC2

dcbaker:
They don't accept .so's (and presumably .dlls or .dylibs) to be passed
to the compiler directly, they must be passed as -L=foo

Fixes: #7842


test cases/d/12 link directly to lib/meson.build
```diff
@@ -0,0 +1,14 @@
+project('MesonLinkBoostBug', ['d', 'cpp'])
+
+src = ['main.d', 'external.cpp']
+
+boost_dep = dependency('boost', required : false)
```
mensinda, line 5: The boost dep will not link to any libraries unless a specific module (which is not header only) is specified. So, this won't link to anything, which is OK in this case because `boost::optional` is header only, but this won't help to reproduce the bug.

dcbaker: gah. @#$# header only "libraries"

dcbaker: alright, now using threads for absolutely no reason :)

dcbaker:
gahhh, of course something goes wrong with ancient versions of boost :(

Rogni:
without system in modules: build filed
https://travis-ci.org/github/Rogni/MesonLinkWithBoostBug/builds/735569685#L335

with system:
https://travis-ci.org/github/Rogni/MesonLinkWithBoostBug/builds/735570359#L322
build success

jpakkane:
Something is still wrong:

```
"dmd"  -of=app.exe app.exe.p/main.d.obj app.exe.p/external.cpp.obj "-L=/DEBUG" "-L=/PDB:app.pdb" "-L=/nologo" "-L=/release" "-L=/nologo" "-m64" "-L=phobos64.lib" "-L=C:/hostedtoolcache/windows/Boost/1.72.0/x86_64/lib/boost_system-vc142-mt-gd-x64-1_72.lib" "-L=C:/hostedtoolcache/windows/Boost/1.72.0/x86_64/lib/boost_thread-vc142-mt-gd-x64-1_72.lib"
   Creating library app.lib and object app.exp
main.d.obj : error LNK2019: unresolved external symbol "int __cdecl char_to_int(char const * const)" (?char_to_int@@YAHQEBD@Z) referenced in function _Dmain
  Hint on symbols that are defined and could potentially match:
    "int __cdecl char_to_int(char const *)" (?char_to_int@@YAHPEBD@Z)
app.exe : fatal error LNK1120: 1 unresolved externals
Error: linker exited with status 1120
```

dcbaker:
I *think*, but haven't gotten out windows to check, that dmd is selecting a different language standard mangling rules. There's a switch for it.

dcbaker:
gah, it appears that MSVC is more strict about constness than GCC/clang, making the pointer itself const as well as the data fixes the issue.

jpakkane:
```
/usr/bin/ld: app.p/external.cpp.o: relocation R_X86_64_32 against `.bss' can not be used when making a PIE object; recompile with -fPIE
/usr/bin/ld: final link failed: nonrepresentable section on output
```

dcbaker:
Yeah, I'm not sure how to fix this. D doesn't currently have support for pie (and I've looked at it and can't figure out how to get it working with non gdc).

jpakkane:
Maybe we need to detect this setup and skip the test in that case?

Rogni:
should we disable test if D compiller is gdc ?

dcbaker:
That's just covering up a real problem that's going to bite people in production, I guess maybe that's better than not having this at all.

PR #7848, opened 2020-10-14
Add pkgconfig directory from $prefix/$libdir to PKG_CONFIG_LIBDIR

### LINKED ISSUE #7814: Expecting meson to add $prefix/$libdir/pkgconfig/ to --pkg-config-path by default
# Describe the bug

When I use:

    meson setup builddir -Dprefix=/tmp/install

I would expect meson to actually do this:

    meson setup builddir -Dprefix=/tmp/install --pkg-config-path "/tmp/install/lib/x86_64-linux-gnu/pkgconfig/"

where `/tmp/install` is the `prefix` and `lib/x86_64-linux-gnu/pkgconfig` is the pkgconfig location in my system. (The `x86_64-linux-gnu` is a debian specific component to `$libdir`  as far as I know, in other systems it may be `/tmp/install/lib//pkgconfig/`)

So, in other words, I would expect meson to tell pkg-config to look for dependencies in the installation prefix,
since I may have installed some dependencies previously in that same prefix.

# To Reproduce

I've set up a minimal directory structure of two projects, one (`proj1`) installing a .pc file and one (`proj2`) that looks for it. My issue refers to where packages look for dependencies, so it is seen in `proj2`.

[reproduce-missing-prefix-pkg-config-path.tar.gz](https://github.com/mesonbuild/meson/files/5326031/reproduce-missing-prefix-pkg-config-path.tar.gz)

    tar xzf reproduce-missing-prefix-pkg-config-path.tar.gz
    cd reproduce-missing-prefix-pkg-config-path.tar.gz
    ./test.sh

# Expected behavior

As mentioned, I would expect meson to tell pkg-config to look for dependencies in the installation prefix.

To me, repeating the prefix (`/tmp/install`) component is error-prone and finding out the `lib/x86_64-linux-gnu/pkgconfig` subdirectory (which is system dependant in meson) can be hard. More importantly, I fail to see a reason why someone would *not* want to have the install prefix in the `PKG_CONFIG_PATH`. I may not be seeing the full picture, though.

# system parameters

* Is this a cross build or just a plain native build? A native build
* Operating system: Ubuntu 20.04
* what Python version are you using: 3.8.2
* what `meson --version`: 0.55.3
* what `ninja --version` if it's a Ninja build: 1.10.0

zeehio:
With this pull request, meson will now also search for pkg-config dependencies (for the host machine) under the installation prefix.

Closes #7814

So if this PR was accepted the following call:

    meson setup builddir -Dprefix=/tmp/install

Would behave like this call in previous meson versions:

    meson setup builddir -Dprefix=/tmp/install --pkg-config-path "/tmp/install/lib/x86_64-linux-gnu/pkgconfig/"

So with this PR, we would avoid typing `--pkg-config-path "/tmp/install/lib/x86_64-linux-gnu/pkgconfig/"`. 

When installing to a custom prefix, some pkg-config libraries may already exist there from other meson projects. It makes sense to me that meson looks by default for pkg-config entries where it may have installed them.

Asking the user to provide that path is unconvenient, especially because finding out the libdir used by meson (e.g. `lib/x86_64-linux-gnu/pkgconfig`, `lib/pkgconfig`) may not be straightforward.

Besides, I fail to see a reason why someone would *not* want to have the install prefix in the `PKG_CONFIG_PATH`.

### DIFF
--- a/mesonbuild/dependencies/base.py
+++ b/mesonbuild/dependencies/base.py
@@ -660,6 +660,14 @@ class PkgConfigDependency(ExternalDependency):
         env['PKG_CONFIG_PATH'] = new_pkg_config_path
 
         pkg_config_libdir_prop = environment.properties[for_machine].get_pkg_config_libdir()
+        # Add the pkgconfig directory from the installation prefix to PKG_CONFIG_LIBDIR:
+        if for_machine == MachineChoice.HOST:
+            if pkg_config_libdir_prop is None:
+                pkg_config_libdir_prop = []
+            pkg_config_libdir_prefix = os.path.join(
+                environment.get_prefix(), environment.get_libdir(), "pkgconfig")
+            if os.path.isdir(pkg_config_libdir_prefix):
+                pkg_config_libdir_prop.append(pkg_config_libdir_prefix)
         if pkg_config_libdir_prop:
             new_pkg_config_libdir = ':'.join([p for p in pkg_config_libdir_prop])
             env['PKG_CONFIG_LIBDIR'] = new_pkg_config_libdir
--- a/docs/markdown/snippets/search_pkgconfig_dependencies_in_install_prefix.md
+++ b/docs/markdown/snippets/search_pkgconfig_dependencies_in_install_prefix.md
@@ -1,0 +1,5 @@
+## Search for pkgconfig dependencies in the installation prefix
+
+Meson will now also search for pkg-config dependencies for the host machine under
+the installation prefix.
+
--- a/run_unittests.py
+++ b/run_unittests.py
@@ -6920,6 +6920,26 @@ class LinuxlikeTests(BasePlatformTests):
             self.build()
 
     @skipIfNoPkgconfig
+    def test_pkgconfig_in_prefix(self):
+        '''
+        '''
+        with tempfile.TemporaryDirectory() as temp_prefixdirname:
+            # build library
+            testdirbase = os.path.join(self.unit_test_dir, '85 pkg-config in prefix')
+            proj1dir = os.path.join(testdirbase, 'proj1')
+            self.init(proj1dir, extra_args=['--prefix=' + temp_prefixdirname,
+                                            ], default_args=False)
+            self.build()
+            self.install(use_destdir=False)
+
+            # build user of library
+            proj2dir = os.path.join(testdirbase, 'proj2')
+            self.new_builddir()
+            self.init(proj2dir, extra_args=['--prefix=' + temp_prefixdirname,
+                                            ], default_args=False)
+            self.build()
+
+    @skipIfNoPkgconfig
     def test_static_archive_stripping(self):
         '''
         Check that Meson produces valid static archives with --strip enabled
--- a/test cases/unit/85 pkg-config in prefix/proj1/meson.build
+++ b/test cases/unit/85 pkg-config in prefix/proj1/meson.build
@@ -1,0 +1,13 @@
+project('proj1', ['c'], version : '1.0.0')
+
+pkg = import('pkgconfig')
+
+pkg.generate(
+  name: meson.project_name(),
+  description: 'Project 1',
+  libraries: [],
+  subdirs: [],
+  version: meson.project_version(),
+  filebase : meson.project_name()
+)
+
--- a/test cases/unit/85 pkg-config in prefix/proj2/meson.build
+++ b/test cases/unit/85 pkg-config in prefix/proj2/meson.build
@@ -1,0 +1,3 @@
+project('proj2', ['c'], version : '1.0.0')
+
+dependency('proj1')


jpakkane:
I don't know enough about pkg-config setup to say much (apart from the failing tests). Who does know? @xclaesse? @dcbaker?

xclaesse:
It would make a lot of sense in most cases, but I think there is a reason why not. I can cross compile on x86 PC for an ARM device and still have `prefix=/usr` because that's the final destination once installed on device, but install the project with `DESTDIR=/my/arm/sysroot`, then rsync `/my/arm/sysroot` to the devices's `/`. Under that scenario with this PR meson would look into my system x86 libraries to find arm libraries, and that's not going to end well.

dcbaker:
I agree with @xclaesse, turning this on by default will break working setups. I'm not opposed to the behavior, but it needs to have an opt in mechanism.

xclaesse:
It does have an opt in mechanism: `--pkg-config-path "/prefix/lib/pkgconfig/"`, I'm not sure it's worth adding an option just for saving a bit of typing, seems too corner case to me.

zeehio:
Thank you very much for your review


@xclaesse, it's not about saving a bit of typing it's about having to tell meson its own default settings.

The existing opt in mechanism (passing `--pkg-config-path "/prefix/lib/pkgconfig/") can be challenging to get right, because meson uses different libdir default values per system (e.g. we would need `--pkg-config-path "/prefix/lib/x86_64-linux-gnu/pkgconfig/"` on Debian based systems). Manually looking what's the pkg-config-path meson chose by default to then tell meson to consider it in its search sounds against common user friendliness to me.

I have fixed my PR so all tests pass (let's see when they finish...) and I have hidden this feature behind `--pkg-config-path-append-prefix`. A long option name for a corner case feature that makes sense in most cases seems fair enough to me.

The decision is yours anyway, and I appreciate your time and consideration

Thanks again!

mensinda:
If we go ahead with this, then this mechanism should also work for `cmake_prefix_path` at the same time.

jpakkane:
Experience has shown that boolean options usually grow the need to specify some third value as well. Is there ever as case where you'd want to only search the install prefix? This would imply that the argument should be something like `--pkg-search-from` with values `system`, `install` and `both`.

zeehio:
I have added cmake search support (and unit test) as suggested by @mensinda (Thanks for the idea!) and changed the option flag to `--dependency-search-prefix`

    meson setup builddir -Dprefix=/tmp/install --dependency-search-prefix=true

I'll think if I can come up with a better name for the option, I am open to suggestions.

With respect to @jpakkane 's question regarding the "only search in the install prefix", I lack the cmake and pkg-config experience to tell. While here we are using `PKG_CONFIG_PATH` to specify the install prefix as an additional directory I know there is also `PKG_CONFIG_LIBDIR` to specify where to locate system directories, and that is possible to set in a cross file directory, according to @xclaesse .

mensinda:
At least for CMake, restricting default search paths is possible by setting a few variables (see [`CMAKE_FIND_USE_SYSTEM_ENVIRONMENT_PATH`](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_USE_SYSTEM_ENVIRONMENT_PATH.html)) but this is likely not in the scope of this PR. However, since it is theoretically possible and potentially usefull to *only* search in the install prefix, going with @jpakkane suggestion would make sense.

The CMake part of the PR LGTM.

I have no idea about pkg-config though.

zeehio:
In the last commit I implemented the `--dependency-search-from=system` and `--dependency-search-from=both` options as suggested by @jpakkane  and @mensinda (Thanks to both). I leave the `--dependency-search-from=install` option as further work for another pull request, since I lack the pkg-config and cmake skills to do it properly and it would take me several more iterations.

If you prefer the `--pkg-search-from=` instead of `--dependency-search-from=` name I can change that as well, but the `--pkg-search-from` makes me think it is a `pkg-config` specific option, while the `dependency-search-from` sounds more generic.

mensinda (approved):
I would opt for something with `default` in the name since the current option could imply something more strict (maybe `default_search_paths` or `default_search_source`)?

Besides this option name bikeshedding, the PR LGTM and the current name is also not terrible.

mensinda:
Also, when you rename the variable, please consider squashing some commits to keep the git history clean. It is not a real issue in this case, since the PR is relatively small and we can just `Squash and merge` but this can be problematic for larger PR and should ideally be avoided.

zeehio:
With respect to the option name, I would go for something that contains `default` (as suggested by @mensinda ),  `dependency` (because it relates to finding dependencies) and search/find/paths/sources, since that is what it does. I suggest `--default-dependency-sources=`

With respect to its value: A combo is better than a logical. I wonder if a list (with `system` and `local` as possibilities) would be better than a combo: If there was another option we are not seeing the option `both` would stop having the expected semantic. And right now it would have a more explicit meaning: `--default-dependency-sources=system,local` seems more clear to me than `--default-dependency-sources=both`.

@jpakkane I am happy to implement any option. Whatever you decide will be fine for me.

zeehio:
Summary of the current status:

- This PR allows the user to choose where pkg-config and cmake will search for dependencies by default: either in the system (native or as defined in a cross file) and/or in the local installation prefix.
- It the option `--default-dependency-sources`, with values either `system`, `local` or `both`.
- The default is `system` preserving backwards compatibility.
- Support for `both` is implemented in this PR.
- Support for `local` is left to be implemented on a future PR, if it is ever needed.
- Closes #7814

mensinda (approved):
This PR didn't make it into 0.56.x due to the dev freeze, but should be good to go now.

However, since this is new functionality, the final aprooval of @jpakkane is needed here.

jpakkane:
What do @xclaesse and @dcbaker think of the new approach here?

mensinda:
ping @dcbaker @xclaesse

mensinda:
Ping again. Also, there are some conflicts now :(

zeehio:
If you don't mind I'd rather rebase after the review, so I only do it once.

I'm a bit surprised for how long code reviews take in meson. Meson would benefit of better patch management and timings, in my opinion. I guess it's a matter of bandwidth and ensuring code quality. Anyway, we all do the best we can, so thanks for all your efforts (to all of you).

mesonbuild/dependencies/base.py
```diff
@@ -655,6 +655,12 @@ def setup_env(env, environment, for_machine, extra_path=None):
         sysroot = environment.properties[for_machine].get_sys_root()
         if sysroot:
             env['PKG_CONFIG_SYSROOT_DIR'] = sysroot
+        # Add the pkgconfig directory from the installation prefix to PKG_CONFIG_PATH:
+        if environment.coredata.builtins_per_machine[for_machine]['default_dependency_sources'].value in ('install', 'both'):
```
eli-schwartz, line 659: Why accept it if it's not handled the way you'd expect it to be? Actually setting this to "install" would result in behavior which is surprising now, and not backwards-compatible if it gets implemented later.

eli-schwartz (changes requested):

Issue #7825, opened 2020-10-08
CUDA compiler header check support


zzzoom:
I'm trying to detect the presence of CUB in the CUDA toolkit with something similar to:

```meson
if not meson.get_compiler('cuda').has_header('cub/version.cuh')
    cuda_includes += include_directories('cub-1.8.0')
endif
```
However, using `has_header` or `has_header_symbol` on a CUDA compiler with Meson 0.55.3 results in:

`ERROR: Language Cuda does not support header symbol checks.`

Is this due to an `nvcc` limitation or just because nobody has implemented it yet?

jpakkane:
Probably the latter.

zzzoom:
The cuda compiler has a [definition of `has_header_symbol`](https://github.com/mesonbuild/meson/blob/e36f713a7f5cc10d3f8adc7ae1c73ef6cce51082/mesonbuild/compilers/cuda.py#L169), however it calls `super().has_header_symbol` in Compiler, which always throws.

dcbaker:
It looks like the `has_header_symbol` has been broken since it landed :(

It looks like it just needs the call to super() and the check of super()'s return value to be removed. But I don't have an machine that runs cuda to check on. If you can test and want to send an MR I'll review it for you.

PR #7849, opened 2020-10-14
compilers/cuda: Fix has_header_symbol check

zzzoom:
Header symbol checks for the CUDA compiler are broken (#7825) because the check was taken from the C++ compiler code. This doesn't work because `CPPCompiler` derives from `CLikeCompiler`, which implements C preprocessor checks in `has_header_symbol`, while `CudaCompiler` derives from `Compiler`, whose `has_header_symbol` throws.

This PR fixes these problems by:
- Not calling `super.has_header_symbol` anymore.
- Checking for C preprocessor macros. (with code copied from `CLikeCompiler`)

### DIFF
--- a/mesonbuild/compilers/cuda.py
+++ b/mesonbuild/compilers/cuda.py
@@ -170,16 +170,29 @@ class CudaCompiler(Compiler):
                           env: 'Environment', *,
                           extra_args: T.Optional[T.List[str]] = None,
                           dependencies: T.Optional[T.List['Dependency']] = None) -> T.Tuple[bool, bool]:
-        result, cached = super().has_header_symbol(hname, symbol, prefix, env, extra_args=extra_args, dependencies=dependencies)
-        if result:
-            return True, cached
         if extra_args is None:
             extra_args = []
         fargs = {'prefix': prefix, 'header': hname, 'symbol': symbol}
+        # Check if it's a C-like symbol
+        t = '''{prefix}
+        #include <{header}>
+        int main(void) {{
+            /* If it's not defined as a macro, try to use as a symbol */
+            #ifndef {symbol}
+                {symbol};
+            #endif
+            return 0;
+        }}'''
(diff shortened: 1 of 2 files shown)


mensinda:
This should have a test case. See https://github.com/mesonbuild/meson/tree/master/test%20cases/cuda for inspiration.

mesonbuild/compilers/cuda.py
```diff
@@ -170,16 +170,29 @@ def has_header_symbol(self, hname: str, symbol: str, prefix: str,
...
+        t = '''{prefix}
+        #include <{header}>
+        int main(void) {{
+            /* If it's not defined as a macro, try to use as a symbol */
+            #ifndef {symbol}
+                {symbol};
+            #endif
+            return 0;
+        }}'''
+        found, cached = self.compiles(t.format(**fargs), env, extra_args=extra_args, dependencies=dependencies)
```
dcbaker, line 186: instead of `format(**mapping)`, you can use `format_map(mapping)`, though it's just a suggestion, as we do exactly what you're doing elsewhere.

dcbaker (approved):
This seems like it's probably right. It would be good to have a test for it

zzzoom:
> This should have a test case. See https://github.com/mesonbuild/meson/tree/master/test%20cases/cuda for inspiration.

Would things break if I just added cuda to https://github.com/mesonbuild/meson/tree/master/test%20cases/common/107%20has%20header%20symbol/meson.build ?

mensinda:
All CUDA tests should ideally be in the cuda directory to separate the specific tests from the common tests that should always work. We only have one CI checking CUDA, so putting the test into `common` would require skipping the check when CUDA is not installed. Such setups should be avoided if possible.

For an easy start, you could just copy the entire `107 has header symbol` folder into the cuda dir and adjust the test to only test for CUDA stuff.

zzzoom:
Ok, all done.

mensinda (approved):
LGTM and all CI errors are unrelated. Merging.

PR #7855, opened 2020-10-16
Allow to cross-compile with C sharp

kgarlinski:
When trying to cross-compile for ARM in Yocto with a C sharp compiler
one would get a KeyError: 'cs_COMPILER'. This can be fixed by creating
the C sharp compile rules for all machines, not just for the build machine.

### DIFF
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -1892,8 +1892,7 @@ https://gcc.gnu.org/bugzilla/show_bug.cgi?id=47485'''))
                 self.generate_java_compile_rule(compiler)
             return
         if langname == 'cs':
-            if self.environment.machines.matches_build_machine(compiler.for_machine):
-                self.generate_cs_compile_rule(compiler)
+            self.generate_cs_compile_rule(compiler)
             return
         if langname == 'vala':
             self.generate_vala_compile_rules(compiler)


dcbaker (approved):
This seems correct to me, though it would be nice to have a test case.

jpakkane:
Just pointing that this has not been done earlier because IIRC C# outputs are platform agnostic. It is the same for native and cross compilation just like it is with Java. It's getting late here so I don't have the time to check, but it should be verified that the implementation for C# should be roughly the same as for Java (unless there are complications, of course).

dcbaker:
I thought about that too, but it doesn't really make sense to force them to be native when you're cross compiling. Have likely could use the same change

jpakkane:
There is a behevioural difference, though. LInking cross compiled libraries to native programs (and vice versa) does not work. For C# and Java it does work just fine. If you need some library during build and after install, you have to build it twice even though the end result is identical.

dcbaker:
Then java an c# targets probably should always be native and never allow cross, which would also fix this issue, right?

kgarlinski:
C# is slightly different from Java in that regard that it can be compiled for a specific architecture only. Usually the only thing that it does is that it limits the architectures that it can be loaded on (because of some native interop), but it does not have to be cross compiled to use that. Microsoft's C++/CLI allows to mix and link native (C++/C) and managed (ex. C#) code in a single library, and you need to cross compile for that. We are not using this case here, and I do not know if meson even supports that.

The reason that we need a cross compilation here is the one that I have mentioned in the PR. We are building a complete system image using OpenEmbedded (Yocto) build system, and by default only the things that got cross-compiled by it's build system can go inside this image; it does not allow things that were native-compiled. We have used meson because it is easier to integrate into Yocto workflow, but to date we had to carry this single patch to allow us to continue with this workflow. If our programs have used only managed code, then theoretically it would probably be possible to hack every single recipe to perform native-only compilation, but it is much easier for us to just do cross-compiling. It would not work if we were to mix it with native code though, as the Open Embedded's build system detects libraries used and would fail during packaging because of the wrong architecture.

If you are completely against the idea for cross-compiling for languages like C# and Java, then I think that meson should at least fail gracefully and not with an exception like it does not.

dcbaker:
We don't support mixing native and C# code, but it's completely possible that we could. @jpakkane, I think given that you can do that It makes sense to allow C# as cross. What do you think?

jpakkane:
You can mix Java and native code as well with JNI. We don't really support that yet, though. It is already possible somehow as people are creating Android apps with Meson.

kgarlinski:
JNI is different in that regard, that you link the native library and Java code (ex. *.jar file) separately; using C++/CLI you just link it into single *.dll library file. The gluing of native/managed code magic is done by the compiler, hence technically it needs to be cross when not building for your host architecture. Again, as I said before, it is not our use case; we just want to easily embed C# binaries into our system using OpenEmbedded (Yocto) build system and simply cross-compiling it is the simplest thing for us to do when using this operating system building framework.

@jpakkane, if you are against this patch then please just say so and close this PR. I do not want to force our use case to meson if it is against it's philosophy or some other reason.

jpakkane:
I'm not against this feature. Cross compiling C# is something we definitely want to support. It's more of a question of us understanding what needs to be done and how. For example I did not know that C# assemblies can have native code.

PR #7858, opened 2020-10-16
Add support for parsing rust unit tests

dcbaker:
this is a pretty basic first step for rust testing. What I'd like to follow up with is a rust module method that can be passed a rust build target, and create a new executable and test target with thaat information. But for now getting support for getting the test outputs and putting them in the junit xml that meson test generates seems like a good start.

### DIFF
--- a/docs/markdown/Reference-manual.md
+++ b/docs/markdown/Reference-manual.md
@@ -1740,6 +1740,7 @@ test(..., env: nomalloc, ...)
     to record the outcome of the test).
   - `tap`: [Test Anything Protocol](https://www.testanything.org/).
   - `gtest` *(since 0.55.0)*: for Google Tests.
+  - `rust` *(since 0.56.0)*: for native rust tests
 
 - `priority` *(since 0.52.0)*:specifies the priority of a test. Tests with a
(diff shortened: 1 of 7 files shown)


anholt:
Shouldn't we also have a passing rust unit test, to make sure we're parsing its output successfully?

anholt:
I looked into the arguments of rust test binaries, and this does seem like the right way to be running and parsing them, at least for now (apparently there's json output for tests, but that's a nightly feature)

dcbaker:
There's several features that are "nightly only" that would be really nice to have (there's one to tell you what linker rustc is going to invoke, which would make the other bug just go away :(

I saw the json too, and wanted to use that. It looks like that landed in 2016 (https://github.com/rust-lang/rust/pull/30711)?

dcbaker:
I've added both a case of no failing, and of all skip. It seems that we don't have a way with our test.json to give expected test results yet. I'm not sure I want to go down that path right now though.

anholt:
rust test support lgtm, I'll leave the second commit to more of a meson insider.

dcbaker:
#7860 contains this, with fixes. Closing in favor of that.

PR #7860, opened 2020-10-16
Add a rust module

dcbaker:
This is now ready for review.

The goal here is to add  a place to put some helpers to make rust and meson play nice together. Presumably this would be the place to put cargo integration into meson as well.

What I have right now is a single helper for building rust tests, it takes an existing rust build target, and creates a new executable build target with the same sources and arguments, but adds the `--test` argument, as well as the `--format=pretty` option, and sets the test protocol to `'rust'`. This makes turning on rust unit tests super simple, as the updated test case shows.

I've chosen not to add this as unstable because we're going to start using this in mesa as soon as it lands.

### DIFF
--- a/docs/markdown/Rust-module.md
+++ b/docs/markdown/Rust-module.md
@@ -1,0 +1,35 @@
+---
+short-description: Rust language integration module
+authors:
+    - name: Dylan Baker
+      email: dylan@pnwbakers.com
+      years: [2020]
+...
+
+# Unstable Rust module
+
+*(new in 0.57.0)*
+
+**Note** Unstable modules make no backwards compatible API guarantees.
+
+The rust module provides helper to integrate rust code into meson. The goal
+is to make using rust in meson more pleasant, while still remaining mesonic,
+this means that it attempts to make rust work more like meson, rather than
+meson work more like rust.
+
+## Functions
+
+### test(name: string, target: library | executable, dependencies: []Dependency)
+
+This function creates a new rust unittest target from an existing rust based
+target, which may be a library or executable. It does this by copying the
+sources and arguments passed to the original target and adding the `--test`
+argument to the compilation, then creates a new test target which calls that
+executable, using the rust test protocol.
+
+This accepts all of the keyword arguments as the
+[`test`](Reference-manual.md#test) function except `protocol`, it will set
+that automatically.
+
+Additional, test only dependencies may be passed via the dependencies
+argument.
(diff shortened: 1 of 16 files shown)


dcbaker:
Ah, also note that I used a bunch of python 3.6 stuff in this, so some test failures on platforms still using 3.5 are expected. Since 3.5 support is being dropped for 0.57, I went ahead and used nice, new stuff.

dcbaker:
Anyone interested in looking at this?

nirbheek:
I am really glad that you're working on this! I want to review this, but I am not sure when I will be able to get to it. 

> I've chosen not to add this as unstable because we're going to start using this in mesa as soon as it lands.

Is this because you want the API to be fixed before you start using it in Mesa or you're worried about breakage from the rename? `import('unstable-rust')` will continue to work even after we remove the prefix and call the module `rust`, so that shouldn't be a problem.

However, if you want the API to be finalized before merging, that would make things harder. Often actual usage shows that an API break is needed. But maybe that's ok, and we can just wrap it into our usual deprecation cycle, like we did for the Python module.

dcbaker:
Ah, I didn't realize the `unstable-rust` would continue to work if it was renamed. I can change it then, as that was my biggest concern.

dcbaker:
I'll change to `unstable-rust` and I think I'll split some of the groundwork patches out as they should be safe to merge separately.

bonzini:
Regarding "_mtest: Handle subtest results as a dict_",  the pull request at #8029 is already adding a lot of the infrastructure to store TAP test results as individual objects (e.g. `TAPParser.Test` objects).

dcbaker:
And if 8029 lands first I'll rebase on top of it. This has been opened for quite a while (and the rust test type changes which included the dict change were opened a long time before that).

jpakkane:
And that one is waiting for this one. Which one do people want to see landed first?

bonzini:
I'm fine with this one going in first if @dcbaker rebases it.

dcbaker:
@jpakkane, @bonzini, rebased and passing CI (minus cygwin which still hasn't been fixed).

docs/markdown/Rust-module.md
```diff
@@ -0,0 +1,32 @@
+---
+short-description: Rust language integration module
+authors:
+    - name: Dylan Baker
+      email: dylan@pnwbakers.com
+      years: [2020]
+...
+
+# Rust module
```
jpakkane, line 9: This should mention that it is `unstable-rust` rather than `rust`, much like `unstable-simd`.

jpakkane:
The module needs a release note snippet.

Issue #7611, opened 2020-08-18
windows: master chokes on `c_std=gnu99` setting in meson.build


boxerab:
These lines are from the latest `Pango` `meson.build`:

```
      # We only need c99, but glib needs GNU-specific features
          # https://github.com/mesonbuild/meson/issues/2289
          'c_std=gnu99',
        ],
```

version `0.54` of meson handles it correctly, but not master.

xclaesse:
What's the error message?

rabindra-harlalka:
> What's the error message?
```
The Meson build system
Version: 0.55.999
Source dir: C:\Work\...\src\pango
Build dir: C:\Work\...\src\pango\build
Build type: native build
Project name: pango
Project version: 1.45.1

C:\Work\...\src\pango\meson.build:1:0: ERROR: Value "gnu99" (of type "string") for combo option "C language standard to use" is not one of the choices. Possible choices are (as string): "none", "c89", "c99", "c11".
```

dcbaker:
What compiler are you using?

rabindra-harlalka:
MSVC. Running from Visual Studio x64 command prompt; backed in the default (ninja)

xclaesse:
Ok, gave it a try, got :
```
meson.build:1:0: ERROR: Value "gnu89" (of type "string") for combo option "C language standard to use" is not one of the
 choices. Possible choices are (as string): "none", "c89", "c99", "c11".
```

Which makes sense, MSVC clearly does not support GNU C. Wondering why that ever worked, but indeed it was working with 0.55.

boxerab:
> 
> Which makes sense, MSVC clearly does not support GNU C. Wondering why that ever worked, but indeed it was working with 0.55.

a miracle! So, just need to replace `gnu` with `c` .

xclaesse:
cc @dcbaker since that's likely a regression from https://github.com/mesonbuild/meson/pull/6597.

@jpakkane did we have a policy that setting unknown c_std should gracefully fallback to some default value?

xclaesse:
Confirmed that GLib with MSVC is indeed broken now.

xclaesse:
For context, glib does in many places `#if defined(__GNUC__)`. So if the compiler supports gnu99 it uses it, but it can be built without it too.

xclaesse:
[...]

PR #7862, opened 2020-10-17
Make MSVC accept gnu11 as a language standard version. Closes: #7611.

jpakkane:
Not actually tested ...

### DIFF
--- a/mesonbuild/compilers/c.py
+++ b/mesonbuild/compilers/c.py
@@ -428,7 +428,11 @@ class VisualStudioCCompiler(MSVCCompiler, VisualStudioLikeCCompilerMixin, CCompi
 
     def get_options(self) -> 'OptionDictType':
         opts = super().get_options()
-        c_stds = ['none', 'c89', 'c99', 'c11']
+        c_stds = ['none', 'c89', 'c99', 'c11',
+                  # Need to have these to be compatible with projects
+                  # that set c_std to e.g. gnu99.
+                  # https://github.com/mesonbuild/meson/issues/7611
+                  'gnu89', 'gnu90', 'gnu9x', 'gnu99', 'gnu1x', 'gnu11']
         opts.update({
             'std': coredata.UserComboOption(
                 'C language standard to use',
@@ -442,8 +446,8 @@ class VisualStudioCCompiler(MSVCCompiler, VisualStudioLikeCCompilerMixin, CCompi
         args = []
         std = options['std']
         # As of MVSC 16.7, /std:c11 is the only valid C standard option.
-        if std.value in {'c11'}:
-            args.append('/std:' + std.value)
+        if std.value in {'c11', 'gnu11'}:
+            args.append('/std:c11')
         return args
 
 


xclaesse:
That's not enough AFAIK, it should also accept gn89, gnu99, basically anything other compilers could use. Also could add a comment there that it is a work around until we have a proper solution, and link to relevant issue.

jpakkane:
Updated. Thinking more, maybe the language standard version should not be set by each compiler? Instead the set of values could always be the same, the compiler classes would just error out or fallback on versions they don't support.

xclaesse:
@jpakkane the tricky part is for projects like glib that uses GNU features, but with lots of #ifdef can also build without. Some other projects could instead require GNU compiler and simply fail otherwise.

dcbaker (commented):
I think it would be best to give some kind of warning that gnuXX was selected, but cXX is going to be used, and that this behavior is deprecated and will change in the future.

dcbaker:
I have a patch that moves the std down to the CCompiler or CPPCompiler, asks then reach complier updates the choices it supports. I think that's what you were thinking about?

xclaesse:
> I think it would be best to give some kind of warning that gnuXX was selected, but cXX is going to be used, and that this behavior is deprecated and will change in the future.

I'm fine with deprecating, but only if we have an alternative which we currently don't have for 0.56.0 release. Unless `c_std=c99` does enable GNU extensions on gcc/clang?

dcbaker:
I'm pretty sure that c99 on gcc doesn't provide gnu extensions. At least a warning that you're falling back and things may not work would be good

xclaesse:
The problem is that would be a warning that glib and many other project will have no way to get ride of. A warning with no alternative is not good. But I'm all for a warning once we can provide an alternative syntax for that use case.

dcbaker:
But not having the warning is going to leave people who's project *cant* fall back to c99 saying "meson configured, but it always fails to compile, what gives!". The warning could well be "requested gnu99, but only c99 available. This may or may not work." We have similar warnings for requesting C++11 with msvc, as it only has C++98 and C++14, so we auto promote C++11 to C++14.

Here's the cleanup I mentioned earlier: #7866

dcbaker:
@jpakkane, I noticed in #7866 that clang-cl has the exact same problem. If you would, could you just merge that instead? if not, could you pick the clang-cl and warnings patches over to this?

nirbheek:
> But not having the warning is going to leave people who's project _cant_ fall back to c99 saying "meson configured, but it always fails to compile, what gives!".

This is how meson has always behaved before 0.56. By making gnuXY not fallback to cXY on compilers that don't support it, we are changing the meaning of the `c_std=` argument in a backwards-incompatible way with no way around it for projects.

I would argue that the vast majority of cross-platform projects do not want to be strict about gnustd vs cstd because they handle that using `#ifdef` in the code, so we should _definitely not_ change the default behaviour. However, some projects may indeed want to require gnustd (and these probably don't care about cross-platform support, so they won't even hit the issue where a compiler doesn't support a gnustd), and for that we should add new syntax.

dcbaker:
No, that's how msvc and clang-cl have behaved. If you pass `c_std=gnu24` to gcc it will straight up error. The problem here is that MSVC and Clang-CL didn't implement standard handling at all, and that resulted in them accepting anything. But no other C compiler's have that behavior. Frankly, what does 'c_std=gnu99' mean if it doesn't mean "I need gnu99?" because if it doesn't mean "gnu99 or else" I'm going to have to change a lot of projects in the xorg camp because we're relying on gnu99 means gnu99 not c99.

nirbheek:
> The problem here is that MSVC and Clang-CL didn't implement standard handling at all, and that resulted in them accepting anything

Yes, and they still don't really implement standard handling (c11 support is incomplete, older stds will never be implemented, and all this is experimental). MSVC implements random bits of POSIX and GNU libc APIs, so if it helps, you can look at it as "msvc supports all the standards, as best as you can expect". So the old implementation was _more correct_ than what we have today.

But you're doing all the work, so it's your call. I won't say anything more on the subject.

nirbheek:
0.56.1 will be fixed with https://github.com/mesonbuild/meson/pull/7946

0.57 can be discussed in https://github.com/mesonbuild/meson/pull/7866

Closing.

PR #7866, opened 2020-10-18
Put the Compiler standard option in the Language mixin

dcbaker:
This has a nice effect on code duplication, plus ensures that all compilers expose the {lang}_std option, even if the only option is none. This prevents cases where a project says "I need gnu99" but the compiler doesn't support it, and meson acts like it does.

This also cleans up the options, as some use (incorrectly) `c_std`, when they should use just `std`

### DIFF
--- a/mesonbuild/compilers/c.py
+++ b/mesonbuild/compilers/c.py
@@ -462,6 +462,25 @@ class ClangClCCompiler(ClangClCompiler, VisualStudioLikeCCompilerMixin, CCompile
                            full_version=full_version)
         ClangClCompiler.__init__(self, target)
 
+    def get_options(self) -> 'OptionDictType':
+        # Clang-cl can compile up to c99, but doesn't have a std-swtich for
+        # them. Unlike recent versions of MSVC it doesn't (as of 10.0.1)
+        # support c11
+        opts = super().get_options()
+        c_stds = ['none', 'c89', 'c99',
+                  # Need to have these to be compatible with projects
+                  # that set c_std to e.g. gnu99.
+                  # https://github.com/mesonbuild/meson/issues/7611
+                  'gnu89', 'gnu90', 'gnu9x', 'gnu99']
+        opts.update({
+            'std': coredata.UserComboOption(
+                'C language standard to use',
+                c_stds,
+                'none',
+            ),
+        })
+        return opts
+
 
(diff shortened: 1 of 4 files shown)


dcbaker:
Hmmm, now that I think about it, clang-cl has the `/clang` option for proxying options to clang itself. We might be able to do `/clang:-std=c99`, etc. Which would be better.

Let's hold off on merging this until I have a chance to look at that.

mesonbuild/compilers/c.py
```diff
@@ -420,6 +420,11 @@ def get_options(self) -> 'OptionDictType':
     def get_option_compile_args(self, options: 'OptionDictType') -> T.List[str]:
         args = []
         std = options['std']
+        if std.value.startswith('gnu'):
+            mlog.warning(
+                'cl.exe does not actually support gnu standards, and meson '
+                'will instead demote to the nearest ISO C standard. This '
+                'may cause compilation to fail.', once=True)
```
nirbheek, line 427: We should not add this warning till we have a replacement for this workflow.

Please, this is really really bad. We should **never** be spammy about warnings. It makes `--fatal-meson-warnings` unusable by projects and makes it harder for us to deprecate things because people just ignore warnings.

dcbaker: I've replaced the warning with a log() message instead, as the message seems useful to me (all of my projects that use gnu99 will fail to compile with c99, so giving me *something* is nice.) I've also replaced a similar warning in the IntelCl compiler with a log() message instead.

I'll get to adding a fallback mechanism later, but for now is this sufficient to land?

dcbaker:
The LGTM is a lie, i've moved some things around, so it thinks i've "fixed" things and "broken" things, but I've just moved existing warnings from here to there.

jpakkane:
```
  File "D:\a\1\s\mesonbuild\backend\ninjabackend.py", line 843, in generate_target
    elem = self.generate_link(target, outname, obj_list, linker, pch_objects, stdlib_args=stdlib_args)
  File "D:\a\1\s\mesonbuild\backend\ninjabackend.py", line 2789, in generate_link
    commands += linker.get_option_link_args(self.environment.coredata.compiler_options[target.for_machine][linker.get_language()])
  File "D:\a\1\s\mesonbuild\compilers\c.py", line 411, in get_option_link_args
    libs = options['winlibs'].value.copy()
KeyError: 'winlibs'
```

PR #7868, opened 2020-10-18
Add delays to test for timestamp issues on mac.

jpakkane:


dcbaker (approved):
I was worried this was the kind of change that would be required :/

jpakkane:
Well let's see if it works first. Can't really think of what else could be causing this. :disappointed:

dcbaker:
Nirbheek had suggested moving the macos tests to GitHub actions. I think I'll give that a try soonish

PR #7872, opened 2020-10-19
switch gui_app deprecation to FeatureDeprecatedKwargs

bonzini:
The deprecation message for "gui_app" is appearing for every target
rather than just once, and even if the required version is older
than 0.56.0.  Use @FeatureDeprecatedKwargs to fix both issues.


xclaesse:
Lgtm once CI is green.

dcbaker (approved):
Awesome, thanks!

PR #7874, opened 2020-10-20
cmake: some subproject related fixes

mensinda:
Both commits don't add new functionality and should be OK for 0.56.0

---

cmake: Always create missing includes in build dir

There really isn't any reason to not always create missing
include directories inside the build dir. Just restricting
this to generate generated sources should work in an ideal
world, however, there exists lots of suboptimal CMake code
where this assumption is not always true.

---

cmake: Disable the new (CMake 3.16) PCH support

Subprojects that use the CMake PCH feature will cause
compilation/linker errors. The CMake PCH support
should thus be disabled until this can be properly
translated to meson.

### DIFF
--- a/mesonbuild/cmake/data/preload.cmake
+++ b/mesonbuild/cmake/data/preload.cmake
@@ -4,6 +4,9 @@ endif()
 
 set(MESON_PS_LOADED ON)
 
+cmake_policy(PUSH)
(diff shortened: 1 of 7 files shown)


dcbaker (approved):
This seems fine to me, and I think it's fine for 0.56.0-rc2.

PR #7875, opened 2020-10-20
pkgconfig: Define libdir and includedir in -uninstalled.pc files

xclaesse:
This fixes glib-2.0-uninstalled.pc file.
GLib does `extra_cflags : ['-I${libdir}/glib-2.0/include']` because some
of its headers gets installed there. But when used uninstalled that path
makes no sense and pkg-config aborts because ${libdir} is not defined.
This cannot be worked around by GLib because Meson does not allow
setting different `extra_cflags` for -uninstalled.pc, and does not allow
setting libdir in `uninstalled_variables`.


xclaesse:
Since we guarantee that prefix, libdir and includedir are always defined, better guarantee it for -uninstalled.pc files too, even if it means giving dummy values.

PR #7876, opened 2020-10-20
compiler: allow non-built internal dependencies as arguments

bonzini:
Allow methods on the compiler object to receive internal dependencies, as long as they only specify compiler/linker arguments or other dependencies that satisfy the same requirements.

This is useful if you're using internal dependencies to add special "-D" flags such as -DNCURSES_WIDECHAR, -D_XOPEN_SOURCE_EXTENDED or -DGLIB_STATIC_COMPILATION.

### DIFF
--- a/test cases/linuxlike/2 external library/meson.build
+++ b/test cases/linuxlike/2 external library/meson.build
@@ -18,6 +18,12 @@ int main(void) {
 '''
 
 assert(cc.links(linkcode, args : '-lz', name : 'Test link against zlib'), 'Linking test failed.')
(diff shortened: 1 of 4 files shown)


test cases/linuxlike/2 external library/meson.build
```diff
@@ -18,6 +18,12 @@ int main(void) {
 '''
 
 assert(cc.links(linkcode, args : '-lz', name : 'Test link against zlib'), 'Linking test failed.')
+d1 = declare_dependency(link_args: '-lz')
```
jpakkane, line 21: This should probably also have compile args for completeness.

PR #7877, opened 2020-10-20
setup.cfg: Add pytest discovery args

dcbaker:
We have a single giant file for our tests, but a number of files that
match pytest's default discovery globs. To fix that, let's tell pytest
what to do.

This means you can just `pytest` and get the right results. It also
helps IDE's like vscode correctly identify tests.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -39,3 +39,6 @@ progress =
 
 [tool:pytest]
 python_classes =
+python_files =
+    run_unittests.py
+


scivision (approved):
This is a good idea, to speed up pytest a bit and help the IDEs. 

Long-term, PyTest recommends pyproject.toml over setup.cfg, but this is a recent recommendation. Not everyone may have PyTest >= 6.0 yet for pyproject.toml. We can save long for a future reorganization of setup.{py,cfg} / pyproject.toml.

PR #7887, opened 2020-10-23
interpreter: Add get_keys function for configuration_data

brainelectronics:
would be beneficial to iterate over all configuration_data keys. getter function already exists, but not made available to the user

### DIFF
--- a/docs/markdown/Reference-manual.md
+++ b/docs/markdown/Reference-manual.md
@@ -2408,8 +2408,11 @@ page](Configuration.md) It has three methods:
 
 - `has(varname)`: returns `true` if the specified variable is set
 
-- `keys()`*(since 0.57.0)*: returns all keys of the configuration data object
-  as an iteratable list.
+- `keys()`*(since 0.57.0)*: returns an alphabetically sorted array of keys of
+  the configuration data object.
+
+  You can iterate over this array with the [`foreach`
+  statement](Syntax.md#foreach-statements).
 
 - `merge_from(other)` *(since 0.42.0)*: takes as argument a different
   configuration data object and copies all entries from that object to
--- a/docs/markdown/Reference-manual.md
+++ b/docs/markdown/Reference-manual.md
@@ -2408,7 +2408,7 @@ page](Configuration.md) It has three methods:
 
 - `has(varname)`: returns `true` if the specified variable is set
 
-- `keys()`*(since 0.57.0)*: returns an alphabetically sorted array of keys of
+- `keys()`*(since 0.57.0)*: returns an array of keys of
   the configuration data object.
 
   You can iterate over this array with the [`foreach`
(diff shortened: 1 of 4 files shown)


xclaesse:
We added `keys()` method on dictionary in Meson 0.56.0, so I think it would be more consistent to rename this without the `get_` prefix.

mesonbuild/interpreter.py
```diff
@@ -401,6 +402,9 @@ def get_unquoted_method(self, args, kwargs):
     def get(self, name):
         return self.held_object.values[name] # (val, desc)
 
+    def get_keys_method(self, args, kwargs):
```
dcbaker, line 405: You also need to add a `FeatureNew` decorator to this.

dcbaker (changes requested):
Along with what @xclaesse said, you also need to add a release snippet, and update the documentation (the reference manual, and if there's a page on configuration data that too.

Oh, and a test.

bonzini:
You probably want to sort the returned keys, see also #7900.

docs/markdown/Reference-manual.md
```diff
@@ -2408,6 +2408,9 @@ page](Configuration.md) It has three methods:
 
 - `has(varname)`: returns `true` if the specified variable is set
 
+- `keys()`*(since 0.57.0)*: returns all keys of the configuration data object
+  as an iteratable list.
```
bonzini, line 2412: I would say "array" since it's not only iteratable but also indexable. Same in the release notes.

brainelectronics: after you mentioned this idea, I directly found a usecase for it. Changed it accordingly and updated all necessary stuff. Thanks for the hint :)

docs/markdown/Reference-manual.md
```diff
@@ -2408,6 +2408,12 @@ page](Configuration.md) It has three methods:
 
 - `has(varname)`: returns `true` if the specified variable is set
 
+- `keys()`*(since 0.57.0)*: returns an alphabetically sorted array of keys of
```
xclaesse, line 2411: Note sure we should specify the ordering here, sounds more like an implementation detail. I fear this could open a can of worms if people expect Unicode ordering, which would be leaking python retail in our API.

brainelectronics: Got your point, I'll change the description to just "sorted", so it is mentioned that is sorted, but not further specified how.
```suggestion
- `keys()`*(since 0.57.0)*: returns a sorted array of keys of
```

xclaesse: I would even not mention ordering at all, we don't mention ordering of dict neither that's a design decision to leave that undefined. As an implementation detail however I agree we need to make sure the order is stable across reconfigure.

dcbaker (approved):
This looks good to me, and can be merged once the 0.57 window opens.

PR #7894, opened 2020-10-24
Update unstable CUDA Module to support newest toolkits and drivers.

obilaniu:
With much inspiration from CMake's [own logic](https://github.com/Kitware/CMake/blob/master/Modules/FindCUDA/select_compute_arch.cmake).


obilaniu:
@jpakkane The CI failures appear unrelated, and the code being modified is low risk (in an unstable module) and _mostly_ additive.

May I call for the PR's merger before 0.56.0 is tagged?

jpakkane:
After RC only bug fixes are accepted in master. Sorry.

Issue #7114, opened 2020-05-10
dependency(include_type: 'system') support is not complete on windows


Sahnvour:
To suppress warnings coming from system/dependency headers, the `dependency()` function has the `include_type: 'system'` kwarg.
This was added in #5953.

MSVC supports a similar mechanism via the `/export` flag, described for example [here](https://devblogs.microsoft.com/cppblog/broken-warnings-theory/). It was mentionned in the PR but not implemented as far as I can tell.

The feature is also unused when configuring for clang-cl. I guess it is detected as some kind of msvc, which it is, but am I correct in thinking meson could emit `-Xclang -isystem...` in this case ?

Sahnvour:
Sorry I missed the comment you linked in the merge request.

I get your point, however this missing feature is making it tedious to write cross-platform code with higher standards than external libraries it's using. One has to work around this with proxy includes and/or macros everywhere.

Sahnvour:
Would you be open to allow "experimental" msvc flags for this, accompanied by a warning from meson for example? Working around this is quite cumbersome.

jpakkane:
Is there any info on when that might go from an experimental to an official flag?

And how would the warnings work? The simple solution would be `/external:w0 /external:I foo/bar`? And please tell me there is a spaceless version of that include path arg.

Sahnvour:
I couldn't find information about this becoming official (only documented, as mentionned in https://github.com/mesonbuild/meson/pull/5953#issuecomment-536329993). It seems to have been added and not touched/publicised much since.

My proposal of a warning was only as a compromise between offering feature-parity between compilers, and acknowledging the experimental status of the flag, if meson is being conservative.

PR #7897, opened 2020-10-24
clang-cl: add a translation pass for `-isystem` args to work

Sahnvour:
Related to #7114.

I'm not sure how to please the type checker on this CI failure.

### DIFF
--- a/mesonbuild/compilers/mixins/visualstudio.py
+++ b/mesonbuild/compilers/mixins/visualstudio.py
@@ -26,6 +26,7 @@ from ... import mlog
 
 if T.TYPE_CHECKING:
     from ...environment import Environment
+    from ...dependencies import Dependency
     from .clike import CLikeCompiler as Compiler
 else:
     # This is a bit clever, for mypy we pretend that these mixins descend from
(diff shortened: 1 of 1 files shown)


mesonbuild/compilers/mixins/visualstudio.py
```diff
@@ -26,6 +26,7 @@
 
 if T.TYPE_CHECKING:
     from ...environment import Environment
+    from ..dependencies import Dependency
```
dcbaker, line 29: this should be:
```python
from ...dependencies import Dependency
```

That will make the type checker happy I think.

dcbaker (changes requested):
I think with the import resolved the type checker will be happy. I haven't had a chance to get windows out and test this yet, and I haven't looked to see if we have tests that should now run with clang-cl that skip

mesonbuild/compilers/mixins/visualstudio.py
Sahnvour: Right, thanks.

Sahnvour:
This should be good to go, despite the spurious failures.

dcbaker (approved):

dcbaker (approved):

Sahnvour:
Failures are related to Codecov.
Thanks for the approval @dcbaker, any chance this could be merged in the near future ? That would really help having cleaner code.

eli-schwartz:
> Failures are related to Codecov.

Codecov should never be able to fail. It's just run as a post-test thing to upload coverage results for both successful and failing CI.

PR #7901, opened 2020-10-26
add fs.expand method

bonzini:
Globbing is already available through non-portable and disgusting hacks involving `run_command`. Just do it right in the filesystem module instead.


xclaesse:
I think wildcard is something meson choose to not support by design. The reason is that adding new files won't trigger reconfigure. This would allow doing `executable('myapp', fs.expand('*.c'))`, right?

bonzini:
> That being said, doing it in a module could be a middle ground, IMHO.

Yes, that was my impression with respect to the whole of the `fs` module. As an extreme example, it does not add dependencies for everything that is checked with `fs.isdir`. I'll add the remark to the docs for the `fs` module in general.

bonzini:
FWIW my usecase is that QEMU looks for files in `default-configs/targets/*.mak` and each of those becomes an executable. The check can be overridden with an option, so it's always possible to have reproducible (independent of the filesystem) build recipes by specifying the option.

jpakkane:
Since it does globbing, maybe it should also be called "glob".

mesonbuild/modules/fs.py
```diff
@@ -32,13 +32,14 @@ def __init__(self, interpreter: 'Interpreter') -> None:
         super().__init__(interpreter)
         self.snippets.add('generate_dub_file')
 
-    def _absolute_dir(self, state: 'ModuleState', arg: str) -> Path:
+    def _absolute_dir(self, state: 'ModuleState', arg: T.Optional[str]=None) -> Path:
         """
```
dcbaker, line 36: There should be spaces around the = when using type annotations.

mesonbuild/modules/fs.py
```diff
@@ -69,6 +70,23 @@ def expanduser(self, state: 'ModuleState', args: T.Sequence[str], kwargs: dict)
...
+            arg = Path(p).expanduser()
+            # Path.glob does not support absolute paths
+            if arg.is_absolute():
+                drive = Path(arg.parts[0])
+                p = str(arg.relative_to(drive))
+                result.update(str(x) for x in drive.glob(p))
+            else:
+                result.update(str(x.relative_to(base)) for x in base.glob(p))
+        return ModuleReturnValue(sorted(result), [])
+
```
dcbaker, line 89: I really feel like we should have an `mlog.warn(once=true)` here about the dangers of globbing.

dcbaker (changes requested):
I'm not convinced including this is a good idea. Is one thing to say "here's rope" and another to say, "here's a noose." If we provide it, people will use it, and the reliability will go down. Making it hard to do the wrong thing is a good thing.

Anyway, I've said my peace on the matter, I have a few comments on the code.

jpakkane:
I admit I'm also very torn on this. OTOH source globbing has all the problems mentioned but on the other hand it is useful in certain fringe cases like this. I guess we could make the globber intentionally fail if you give it a pattern that matches source files, but that's being a bit mean...

jpakkane:
For the QEMU case specifically, I'd create a project option that is an array of strings with the allowed values being those ones. If the list of programs changes rarely, this is a fairly good choice. Dunno how well that plays together with the existing system.

bonzini:
Making the list of targets the "choices" for the option is a possibility, but it is unnecessary duplication between the filenames and the `meson.build`. I'd rather avoid having to modify two places when one will do (this was one of the advantages I saw in Meson from the very beginning), and to be honest I'd end up using `run_command` instead.

I don't like unconditional warnings because they dilute the value of warnings and just make users angry (an example is the existing warning about subdir outputs of `custom_target`, and this one would be even worse). The whole of the filesystem module is about fringe cases, so with the new big fat warning in the manual I think we can just ignore users who practice shooting themselves in the foot.

However, one possibility could be to detect `fs.expand` in a build target function and warn in that case. Easily worked around with

```
     my_sources = fs.expand('*.c')
     executable('aa', sources: my_sources)
```

but at some point you just have to realize you're fighting in mud with a pig&mdash;and it's enjoying it.

jpakkane:
One idea from the past was to call it `unreliable_glob` to discourage use.

bonzini:
That's better than a warning, but still places unnecessary burden (cognitive in this case) on the user. There's nothing unreliable per se.

I will play with the trick of detecting source lists.

dcbaker:
For the source list trick, you might consider having this return a special kind of list at the interpreter level, `GlobListHolder`, which isn't accepted by `build_target` for sources.

bonzini:
> you might consider having this return a special kind of list at the interpreter level, GlobListHolder, which isn't accepted by build_target for sources

It may require a bit of surgery due to flattening of positional arguments. I'll probably generalize `@noArgsFlattening` to custom preprocessing of positional arguments.

bonzini:
@dcbaker Here it is: https://github.com/bonzini/meson/commit/fs-expand-warn but I honestly don't think it's worth the very ugly code (e.g. to pass the new list type unscated through holderify). Anybody that uses `executable('foo', fs.expandglob('*.c'))` will realize it's a bad idea very soon.

jpakkane:
If you can iterate over the globbed list then you can do something like:

```meson
filez = []
foreach f in globbed_stuff
  filez += f
endforeach
executable('globbed', filez)
```

bonzini:
> If you can iterate over the globbed list then you can do something like:

Yes, any sufficiently determined person will be able to bypass the warning.

Or are you suggesting that the returned value of `fs.expandglob` should _only_ be usable in `foreach`? I might actually buy that, especially because it's very easy to rollback the limitation in the future.

dcbaker:
I guess the other option (which is much harder to get around) is to have this function return `mesonlib.File` objects, and add an `is_glob` field to them. I don't know, I'm just throwing out ideas.

bonzini:
I finally got something decent [here](https://github.com/bonzini/meson/commit/fs-expand-warn). By returning a wrapper for `list` instead of a subclass, I can fool args flattening without making the code too ugly.

I am still not sure it's a good idea so I've not yet included it in the PR, but if it's okay with you guys I'll add it.

Issue #7473, opened 2020-07-19
`meson test` with a specific test should only build what is needed for the test


alex-tee:
it's an usual workflow of mine to fix something and retest to see if it's fixed (using a particular test), but i have 30 tests now and i need to wait for all of them to finish linking, when i already specified the test i want to run

I think meson should be smart here and only build what's needed for the test (or suite) i specified

alex-tee:
this is actually not a big issue. I was linking a static lib to each test so it was taking a while. using `both_libraries()` when creating a library for my shared code and doing `.get_shared_lib()` for the tests and `.get_static_lib()` for my executable seems to work fine. tests link really fast now

bonzini:
All the information that is needed here is already available in `TestSerialisation` for use by introspection, so it should not be hard to do.

dcbaker:
I have an opened pr for this, there's just a couple of bugs I haven't worked out yet

bonzini:
Is that #6511? It seems a bit different, in particular one is "what to build by default" and the other one is "what to rebuild" (you can rebuild more than needed and do it by default to make things safer, but the ergonomics are worse). For a large project I don't mind building tests by default, but I don't want `meson test --suite unit` to rebuild thousands of files while I haven't even gotten the unit tests to pass for example. 

My idea for this issue was to invoke ninja with the specific targets included in `depends` and test arguments).

dcbaker:
I guess my pr gets half there, since it changes meson test to depend on a target that builds only the tests and their dependencies.

PR #7902, opened 2020-10-26
mtest: only build what is needed for the tests

bonzini:
It is a usual workflow to fix something and retest to see if it is fixed using a particular test.  When tests start to become numerous, it becomes time consuming for "meson test" to relink all of them (and in fact rebuild the whole project) where the user has already specified the tests they want to run, as well as the tests' dependencies.

Teach meson to be smart and only build what is needed for the test (or suite) that were specified.

Fixes: #7473

### DIFF
--- a/docs/markdown/snippets/meson_test_depends.md
+++ b/docs/markdown/snippets/meson_test_depends.md
@@ -1,0 +1,17 @@
+## `meson test` only rebuilds test dependencies
+
+Until now, `meson test` rebuilt the whole project independent of the
+requested tests and their dependencies.  With this release, `meson test`
+will only rebuild what is needed for the tests or suites that will be run.
+This feature can be used, for example, to speed up bisecting regressions
+using commands like the following:
+
+    git bisect start <broken commit> <working commit>
+    git bisect run meson test <failing test name>
+
+This would find the broken commit automatically while at each step
+rebuilding only those pieces of code needed to run the test.
+
+However, this change could cause failures if dependencies are not
+specified correctly in `meson.build`.
+
--- a/mesonbuild/minstall.py
+++ b/mesonbuild/minstall.py
@@ -19,7 +19,6 @@ from glob import glob
 from .scripts import depfixer
 from .scripts import destdir_join
 from .mesonlib import is_windows, Popen_safe
-from .mtest import rebuild_all
 from .backend.backends import InstallData
 from .coredata import major_versions_differ, MesonVersionMismatchException
 from .coredata import version as coredata_version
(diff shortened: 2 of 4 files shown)


bonzini:
The failure is weird since it's a SIGSEGV in a test. It shouldn't be affected by this patch. Let's see if it's transient...

nickbroon:
This is great. I have targets that consists of 100s of file and have LTO enabled by default so take a long time build, but are unneeded by UT tests. This would allow to rapidly build tests during a iterative develop work.

bonzini:
ping...

docs/markdown/snippets/meson_test_depends.md
```diff
@@ -0,0 +1,8 @@
+## `meson test` only rebuilds test dependencies
+
+Until now, "meson test" rebuilt the whole project independent of the requested
+tests and their dependencies.  With this release, "meson test" will
+only rebuild what is needed for the tests or suites that will be run.
+This could cause failures if dependencies are not specified correctly in
+`meson.build`.
+
```
jpakkane, line 8: Maybe something like the following could be added to this to make the usefulness of this feature immediately obvious:

```
This feature can be used, for example, to speed up bisecting regressions using commands like the following:

git bisect start <broken commit> <working commit>
git bisect run meson test <failing test name>

This would find the broken commit automatically while at each step rebuilding only those pieces
of code needed to run the test.
```

bonzini: Sounds good, I adjusted the release notes.

mesonbuild/minstall.py
```diff
@@ -532,6 +531,23 @@ def install_targets(self, d):
                     else:
                         raise
 
+def rebuild_all(wd: str) -> bool:
```
wereii, line 534: `Path` and `environment` are undefined, they aren't imported in `minstall.py` (as they are in `mintro.py`). 
This fails `meson install`.

https://github.com/mesonbuild/meson/blob/2f836e3accd112a52bb8d3ef5d44b7e0fb692fcf/mesonbuild/minstall.py#L534

Made a PR #8136

bonzini: oops, thanks.

xclaesse:
I think this PR causes this regression: https://gitlab.freedesktop.org/ocrete/libnice/-/jobs/7348274

CC @ocrete

xclaesse:
`os.path.relpath(f, wd)` returns path with `\` seperator on Windows, but ninja targets always uses `/` separator.

Issue #7903, opened 2020-10-27
Invoking gtest executable when working directory is set


skartashev:
**Describe the bug**
Problem running "meson test" when workdir is set and protocol is 'gtest'

**To Reproduce**
[gtest_failure.zip](https://github.com/mesonbuild/meson/files/5443832/gtest_failure.zip)

```
meson build && cd build
mkdir tests
meson test
```
**Expected behavior**
File _'build/test/simple-test.xml'_ must appears.
Seems like problem is in gtest arguments: wrong _--gtest_output_ argument. 
Now gtest-output looks like[mtest.py]:

```
            if self.test.workdir:
                gtestname = '{}:{}'.format(self.test.workdir, self.test.name)
            extra_cmd.append('--gtest_output=xml:{}'.format(gtestname))
```

But it must looks like:
```
            if self.test.workdir:
                gtestname = '{}/{}'.format(self.test.workdir, self.test.name)
            extra_cmd.append('--gtest_output=xml:{}.xml'.format(gtestname))
```

**system parameters**
Windows 10
Python 3.8.5
0.56.0.rc2 (latest master)
ninja 1.10.1

skartashev:
#7904 https://github.com/mesonbuild/meson/pull/7904

PR #7904, opened 2020-10-27
Fix gtest invoking while workdir is set

skartashev:
Fixes: https://github.com/mesonbuild/meson/issues/7903

### DIFF
--- a/mesonbuild/mtest.py
+++ b/mesonbuild/mtest.py
@@ -697,7 +697,7 @@ class SingleTestRunner:
 
         extra_cmd = []  # type: T.List[str]
         if self.test.protocol is TestProtocol.GTEST:
-            gtestname = '{}.xml'.format(self.test.name)
+            gtestname = '{}'.format(self.test.name)
             if self.test.workdir:
                 gtestname = '{}/{}'.format(self.test.workdir, self.test.name)
             extra_cmd.append('--gtest_output=xml:{}.xml'.format(gtestname))
--- a/mesonbuild/mtest.py
+++ b/mesonbuild/mtest.py
@@ -697,9 +697,9 @@ class SingleTestRunner:
 
         extra_cmd = []  # type: T.List[str]
         if self.test.protocol is TestProtocol.GTEST:
-            gtestname = '{}'.format(self.test.name)
+            gtestname = self.test.name
             if self.test.workdir:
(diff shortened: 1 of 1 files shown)


mesonbuild/mtest.py
```diff
@@ -697,10 +697,10 @@ def preexec_fn() -> None:
 
         extra_cmd = []  # type: T.List[str]
         if self.test.protocol is TestProtocol.GTEST:
-            gtestname = '{}.xml'.format(self.test.name)
+            gtestname = '{}'.format(self.test.name)
             if self.test.workdir:
-                gtestname = '{}:{}'.format(self.test.workdir, self.test.name)
-            extra_cmd.append('--gtest_output=xml:{}'.format(gtestname))
+                gtestname = '{}/{}'.format(self.test.workdir, self.test.name)
```
dcbaker, line 702: This should use `os.path.join`, instead of hard coding `/`.

mesonbuild/mtest.py
```diff
@@ -697,10 +697,10 @@ def preexec_fn() -> None:
 
         extra_cmd = []  # type: T.List[str]
         if self.test.protocol is TestProtocol.GTEST:
-            gtestname = '{}.xml'.format(self.test.name)
+            gtestname = '{}'.format(self.test.name)
```
dcbaker, line 700: Either leave the .xml here, or replace this with `gtestname = self.test.name`

dcbaker (changes requested):
This looks like legitimate bug fix, so we should try to get this into 0.56.0, I've got a few nits I'd like changed first.

mesonbuild/mtest.py
skartashev: Fixed

mesonbuild/mtest.py
skartashev: Fixed

skartashev:
[gtest_failure.zip](https://github.com/mesonbuild/meson/files/5453719/gtest_failure.zip)

Added test with/without workdir

dcbaker (approved):

dcbaker:
looks good, thanks! I'll merge when CI comes back green (I don't see why it wouldn't).

dcbaker:
the swift failures in the 16.04 image are not related to this MR.

PR #7907, opened 2020-10-28
dist: Allow packaging subproject in same git repo as main project

xclaesse:


xclaesse:
CC @thiblahute

mesonbuild/mdist.py
```diff
@@ -109,8 +123,20 @@ def git_clone(src_root, distdir):
...
-    subprocess.check_call(['git', 'clone', '--shared', src_root, distdir])
+    repo_root = git_root(src_root)
+    if repo_root.samefile(src_root):
+        os.makedirs(distdir)
+        subprocess.check_call(['git', 'clone', '--shared', src_root, distdir])
+    else:
+        subdir = Path(src_root).relative_to(repo_root)
+        tmp_distdir = distdir + '-tmp'
+        if os.path.exists(tmp_distdir):
+            shutil.rmtree(tmp_distdir)
```
dcbaker, line 134: probably need the `mesonlib.safe_rmtree` (or whatever it's called)

xclaesse: Right, fixed now.

docs/markdown/Creating-releases.md
```diff
@@ -58,3 +58,20 @@ for example when done in CI that already does its own testing.
...
+source code from that subdir and not the rest of the main project or other
+subprojects.
+
+For example:
+```sh
+git clone https://github.com/myproject
+cd myproject/subprojects/mysubproject
+meson builddir
+meson dist -C builddir
+```
```
jpakkane, line 77: This should also mention what the outcome will be. Something like "this will create foo.tar.gz and/or baz.tar.gz in directories xyz".

This should also be rebased so CI is rerun. There's quite a lot of red currently.

xclaesse: Done.

jpakkane: All the red is still there ... 😞

jpakkane:
This needs a test. Otherwise it will probably break at the first refactoring that touches the code.  :D

With that, this looks ok to merge.

PR #7908, opened 2020-10-28
Add fs.relative_to() function. Closes #2184

### LINKED ISSUE #2184: Provide a way to get the path to a directory relative to another directory
Basically a `meson.build` equivalent for `os.path.relpath()`.

This would be quite useful for passing to the `install_rpath:` kwarg. For instance if you have a private library in `pkglibdir_abs` = `/usr/lib64/pkgname` which is used by a binary in `bindir_abs` = `/usr/bin`, you want to pass `install_rpath: '../lib64/pkgname'` instead of `pkglibdir_abs`.

zeehio:
The usefulness of a `rel_path()` function has been discussed on #2184.

I have a personal project where a binary installed at `$prefix/$bindir` would look for a file on `$prefix/$datadir/subdirectory`. With `rel_path()` I can compute `/../share/subdirectory` or whatever it is the path to `$prefix/$datadir/subdirectory` from `$prefix/$bindir` and pass that as a  `#define` variable to the `executable()`, avoiding hardcoding the prefix.

### DIFF
--- a/docs/markdown/Fs-module.md
+++ b/docs/markdown/Fs-module.md
@@ -243,8 +243,9 @@ fs.relative_to('/prefix/lib/foo', '/prefix')  # 'lib/foo'
   fallback.
 
 - `if_within`: A path. `relative_to(path1, path2, if_within: path3)` returns
-  the relative path of `path1 with respect to `path2` if `path1` is found in `path3`,
-  otherwise it returns `path1` unchanged.
+  the relative path of `path1` with respect to `path2` if `path1` is found in `path3`.
+  Otherwise, if `allow_absolute: True` and `path1` is an absolute path, `path1` is
+  returned. If neither of those conditions are fullfilled an error is raised.
 
 Examples: 
 
--- a/docs/markdown/Fs-module.md
+++ b/docs/markdown/Fs-module.md
@@ -239,7 +239,8 @@ fs.relative_to('/prefix/lib/foo', '/prefix')  # 'lib/foo'
   following the rules of the build machine (native: `true`) or the host (native: `false`).
 
 - `allow_absolute`: If set to `true`, instead of giving an error, `relative_to()`
-  will return the original path. Useful as a fallback
+  will return the original path. Useful if an absolute path is a reasonable
+  fallback.
 
 - `if_within`: A path. `relative_to(path1, path2, within: path3)` returns
   the relative path of `path1 with respect to `path2` if `path1` is found in `path3`,
--- a/docs/markdown/Fs-module.md
+++ b/docs/markdown/Fs-module.md
@@ -242,7 +242,7 @@ fs.relative_to('/prefix/lib/foo', '/prefix')  # 'lib/foo'
   will return the original path. Useful if an absolute path is a reasonable
   fallback.
 
-- `if_within`: A path. `relative_to(path1, path2, within: path3)` returns
+- `if_within`: A path. `relative_to(path1, path2, if_within: path3)` returns
   the relative path of `path1 with respect to `path2` if `path1` is found in `path3`,
   otherwise it returns `path1` unchanged.
 
--- a/docs/markdown/Fs-module.md
+++ b/docs/markdown/Fs-module.md
@@ -220,12 +220,6 @@ fs.stem('foo/bar/baz.dll.a')  # baz.dll
 *since 0.64.0*
 
 Given two paths, returns a version of the first path relative to the second.
-If a path is given as the 'within' argument, the function will return the first path
-unchanged if it is not inside 'within'.
-
-The `relative_to` function is system dependent. It accepts the `native: true` argument
-to compute relative paths on the build system. Otherwise it uses the host system.
-
 
 Examples: 
 
@@ -233,13 +227,36 @@ Examples:
 # On windows:
 fs.relative_to('c:\\proj1\\foo', 'c:\\proj1\\bar')  # '..\\foo'
 fs.relative_to('c:\\proj1\\foo', 'd:\\proj1\\bar')  # ERROR, since the relative path does not exist
-fs.relative_to('c:\\proj1\\foo', 'd:\\proj1\\bar', within: 'd:\\')  # 'c:\\proj1\\foo', within allows to get an absolute path
 
 # On other systems:
 fs.relative_to('/prefix/lib', '/prefix/bin')  # '../lib'
 fs.relative_to('/prefix/lib/foo', '/prefix')  # 'lib/foo'
-fs.relative_to('/usr/lib/foo', '/usr/bin', within: '/usr')  # '../lib/foo'
-fs.relative_to('/usr/lib/foo', '/usr/bin', within: '/var')  # '/usr/lib/foo'
+```
+
+`relative_to` has some optional keyword arguments:
+
+- `native`: Can be set to `true` or `false`. Whether the relative paths are computed
+  following the rules of the build machine (native: `true`) or the host (native: `false`).
+
+- `allow_absolute`: If set to `true`, instead of giving an error, `relative_to()`
+  will return the original path. Useful as a fallback
+
+- `if_within`: A path. `relative_to(path1, path2, within: path3)` returns
+  the relative path of `path1 with respect to `path2` if `path1` is found in `path3`,
+  otherwise it returns `path1` unchanged.
+
+Examples: 
+
+```meson
+# On windows:
+fs.relative_to('c:\\proj1\\foo', 'c:\\proj2\\bar', if_within: 'c:\\proj2')  # ERROR, the relative path is outside the "within" argument
+fs.relative_to('c:\\proj1\\foo', 'c:\\proj2\\bar', if_within: 'c:\\proj2', allow_absolute: true)  # 'c:\\proj1\\foo'
+fs.relative_to('c:\\proj1\\foo', 'd:\\proj1\\bar', allow_absolute: true)  # 'c:\\proj1\\foo'
+
+# On other systems:
+fs.relative_to('/project1/lib/foo', '/usr/bin', if_within: '/usr')  # ERROR: '/project1/lib/foo' not in '/usr'
+fs.relative_to('/project1/lib/foo', '/usr/bin', if_within: '/usr', allow_absolute: 'true')  # '/project1/lib/foo'
+fs.relative_to('/usr/lib/foo', '/usr/bin', if_within: '/usr')  # '../lib/foo'
 ```
 
 
--- a/docs/markdown/Fs-module.md
+++ b/docs/markdown/Fs-module.md
@@ -204,7 +204,6 @@ suffix
 fs.stem('foo/bar/baz.dll')  # baz
 fs.stem('foo/bar/baz.dll.a')  # baz.dll
 ```
-
 ### read
 - `read(path, encoding: 'utf-8')` *(since 0.57.0)*:
    return a [string](Syntax.md#strings) with the contents of the given `path`.
@@ -215,6 +214,21 @@ fs.stem('foo/bar/baz.dll.a')  # baz.dll
    specified by `path` changes, this will trigger Meson to reconfigure the
    project. If the file specified by `path` is a `files()` object it
    cannot refer to a built file.
+
+### relative_to
+
+*since 0.64.0*
+
+Given two absolute paths, returns a version of the first path relative to the second.
+If a path is given as the 'within' argument is specified, the function will return the
+first path unchanged if it is not inside 'within'.
+
+```meson
+fs.relative_to('/prefix/lib', '/prefix/bin')  # '../lib'
+fs.relative_to('/prefix/lib/foo', '/prefix')  # 'lib/foo'
+fs.relative_to('/usr/lib/foo', '/usr/bin', within: '/usr')  # '../lib/foo'
+fs.relative_to('/usr/lib/foo', '/usr/bin', within: '/var')  # '/usr/lib/foo'
+```
 
 
 ### copyfile
--- a/docs/markdown/Fs-module.md
+++ b/docs/markdown/Fs-module.md
@@ -219,7 +219,7 @@ fs.stem('foo/bar/baz.dll.a')  # baz.dll
 
 *since 0.64.0*
 
-Given two absolute paths, returns a version of the first path relative to the second.
+Given two paths, returns a version of the first path relative to the second.
 If a path is given as the 'within' argument, the function will return the first path
 unchanged if it is not inside 'within'.
 
--- a/docs/markdown/Fs-module.md
+++ b/docs/markdown/Fs-module.md
@@ -220,10 +220,22 @@ fs.stem('foo/bar/baz.dll.a')  # baz.dll
 *since 0.64.0*
 
 Given two absolute paths, returns a version of the first path relative to the second.
-If a path is given as the 'within' argument is specified, the function will return the
-first path unchanged if it is not inside 'within'.
+If a path is given as the 'within' argument, the function will return the first path
+unchanged if it is not inside 'within'.
+
+The `relative_to` function is system dependent. It accepts the `native: true` argument
+to compute relative paths on the build system. Otherwise it uses the host system.
+
+
+Examples: 
 
 ```meson
+# On windows:
+fs.relative_to('c:\\proj1\\foo', 'c:\\proj1\\bar')  # '..\\foo'
+fs.relative_to('c:\\proj1\\foo', 'd:\\proj1\\bar')  # ERROR, since the relative path does not exist
+fs.relative_to('c:\\proj1\\foo', 'd:\\proj1\\bar', within: 'd:\\')  # 'c:\\proj1\\foo', within allows to get an absolute path
+
+# On other systems:
 fs.relative_to('/prefix/lib', '/prefix/bin')  # '../lib'
 fs.relative_to('/prefix/lib/foo', '/prefix')  # 'lib/foo'
 fs.relative_to('/usr/lib/foo', '/usr/bin', within: '/usr')  # '../lib/foo'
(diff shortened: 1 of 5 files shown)


mensinda:
There is already the [FS-Module](https://mesonbuild.com/Fs-module.html) and this function should be added there instead.

mensinda:
Also please consider using `pathlib` instead of `os.path.*`. We are (admittedly slowly) trying to replace all `os.path.*` code with `pathlib` and new code should thus use `pathlib` whenever possible.

zeehio:
Following your advice I have added a `fs.relative_to` function using `pathlib`.

While doing that I found out that the use of `os.path.relpath` does not have a clear translation to `pathlib` (see [pathlib note about this](https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module) as well as [this stack overflow question](https://stackoverflow.com/q/38083555/446149)). I believe the implementation I came up with is clear enough. While I think there may be room for improvement in pathlib to make it more convenient for our use case, I believe the semantic the pathlib implementation provides for `relative_to()` is more simple and easy to understand than `os.path.relpath()`.

Python 3.9 introduced `PurePath().is_relative_to()` but since meson depends on older python versions I resorted to try/except a couple of times.

Python 3.10 has an open pull request to improve `relative_to()` covering our use case https://github.com/python/cpython/pull/19813. If it is merged, our code could be simplified.

I limited the meson implementation to absolute paths. Relative paths may be supported later if desired.

docs/markdown/Fs-module.md
```diff
@@ -194,3 +194,20 @@ Returns the last component of the path, dropping the last part of the suffix
...
+
+Given two absolute paths, returns a version of the first path relative to the second.
+If a path is given as the 'within' argument is specified, the function will return the
+first path unchanged if it is not inside 'within'.
+
+```meson
+fs.relative_to('/prefix/lib', '/prefix/bin')  # '../lib'
+fs.relative_to('/prefix/lib/foo', '/prefix')  # 'lib/foo'
+fs.relative_to('/usr/lib/foo', '/usr/bin', within: '/usr')  # '../lib/foo'
+fs.relative_to('/usr/lib/foo', '/usr/bin', within: '/var')  # '/usr/lib/foo'
```
jpakkane, line 210: This should also specify what happens on Windows paths that are on different drives:

```meson
fs.relative_to('c:\tmp', 'd:\something') # ??? (should return the latter path maybe?)

Comment (zeehio) [reply docs/markdown/Fs-module.md:222]: I have improved Windows support in relative_to based on your comment.

Since proper absolute/relative path computation is system dependent (e.g. pathlib's PurePath does not detect absolute Windows paths on a Linux system), I've added a `native` argument, so the user can choose if he/she needs the relative path for the host or build system.

Besides, I have added tests for windows paths and clarified possible error messages.

This behaviour is now documented:

    fs.relative_to('c:\\proj1\\foo', 'd:\\proj1\\bar')  # ERROR, since the relative path does not exist. Consider using "within".
    fs.relative_to('c:\\proj1\\foo', 'd:\\proj1\\bar', within: 'd:\\')  # 'c:\\proj1\\foo' ,  "within" allows us to get an absolute path if we are not within that directory:

In my opinion it makes sense that `relative_to` always returns a relative path unless we explicitly allow for absolute paths to be returned (using `within`).

**Comment by zeehio** [comment]
Ping. If you don't mind I will rebase once I have the ok of this contribution. (I'd rather rebase only once)

File: docs/markdown/Fs-module.md [hunk 2]
```diff
@@ -194,3 +194,32 @@ Returns the last component of the path, dropping the last part of the suffix
...
+to compute relative paths on the build system. Otherwise it uses the host system.
+
+
+Examples: 
+
+```meson
+# On windows:
+fs.relative_to('c:\\proj1\\foo', 'c:\\proj1\\bar')  # '..\\foo'
+fs.relative_to('c:\\proj1\\foo', 'd:\\proj1\\bar')  # ERROR, since the relative path does not exist
+fs.relative_to('c:\\proj1\\foo', 'd:\\proj1\\bar', within: 'd:\\')  # 'c:\\proj1\\foo', within allows to get an absolute path
```
jpakkane, line 216: This functionality is good, but the UI as it stands is a bit confusing. I had to read the documentation several times in order to understand what it does and why, Now this might just be a question of me being stupid, but it would seem prudent to try to get a nicer experience.

As an example, if the kwarg were called `if_within` or `is_within` its meaning would be more clear. It would indicate a condition rather than a general parameter of some sort.

Secondly, this is problematic in the Windows case if you don't directly control the paths yourself. If you get the file strings as arguments from somewhere, with values `c:\foo` and g:\bar` you can't say that an absolute path is fine without extracting the drive letter as a substring. This is unnecessary work for users. Having a kwarg like `allow_absolute_if_drive_letters_are_different` would be simpler.

Finally, the `native` approach seems a bit limited as people may need to build different kinds of paths so it might be better to instead have a kwarg `path_type` with the values `build`, `host`, `unix` and `windows` with `host` being the default.

bonzini: `fs` seems to be more obviously related to native paths. It may make sense to have path-related methods on the various machine objects, but that would be a different change (instead, I don't see the benefit in hard-coding that one is dealing with POSIX or Windows paths).

Regarding the syntax, I'd propose something like

```
fs.relative_to('c:\\proj1\\foo', 'c:\\proj1\\bar')  # '..\\foo'
fs.relative_to('c:\\proj1\\foo', 'c:\\proj2\\bar', within: 'c:\\proj2)  # ERROR, the relative path is outside the "within" argument
fs.relative_to('c:\\proj1\\foo', 'd:\\proj1\\bar')  # ERROR, for Windows the default "within" is the drive
fs.relative_to('c:\\proj1\\foo', 'c:\\proj2\\bar', within: 'c:\\proj2', allow_absolute: true)  # 'c:\\proj1\\foo'
fs.relative_to('c:\\proj1\\foo', 'd:\\proj1\\bar', allow_absolute: true)  # 'c:\\proj1\\foo'
```

tristan957:
Would be nice to get this in for 0.59 if possible.

Is there any reason this can't be implemented using `os.path.relpath()`?

eli-schwartz:
Did you read @jpakkane's review comments explaining the unrelated reason why?

tristan957:
If you are referring to the drive issues on Windows, I now see how that would be a problem for os.path.relpath (after testing locally). If that is what you are referring to.

dcbaker:
It would be nice if this gets rebased if it moved the `typed_` args decorators as well.

docs/markdown/Fs-module.md
eli-schwartz: As a matter of curiosity -- what's the use case for doing anything other than erroring out when trying to compute the relative path between two different drives?

Python considers this to be a ValueError, which doesn't seem illogical.

docs/markdown/Fs-module.md
```diff
@@ -194,3 +194,32 @@ Returns the last component of the path, dropping the last part of the suffix
 fs.stem('foo/bar/baz.dll')  # baz
 fs.stem('foo/bar/baz.dll.a')  # baz.dll
 ```
+
+### relative_to
+
+*since 0.57.0*
+
+Given two absolute paths, returns a version of the first path relative to the second.
```
eli-schwartz, line 202: Why must they be absolute paths? We generally assume that a relative path is relative to the current meson.build context.

docs/markdown/Fs-module.md
```diff
@@ -194,3 +194,32 @@ Returns the last component of the path, dropping the last part of the suffix
...
+If a path is given as the 'within' argument, the function will return the first path
+unchanged if it is not inside 'within'.
```
eli-schwartz, line 204: Why does within need to be a path? It's contextually meaningless unless it is a Windows drive letter, and in that case it can just as well be a kwarg to return the first path unchanged "if it cannot be rendered as a relative path".

mesonbuild/modules/fs.py
```diff
@@ -89,6 +89,73 @@ def as_posix(self, state: 'ModuleState', args: T.Sequence[str], kwargs: dict) ->
...
+            if part_to != part_from:
+                break
+            parts_root.append(part_to)
+        root = path_class(parts_root[0]) / path_class("/".join(parts_root[1:]))
+
+        # Then find how many levels do we need to go from path_from to the root:
+        num_levels = len(path_from.parts) - len(root.parts)
+        # And build the final path, going from path_from to the root and then to path_to:
+        path = path_class("/".join(num_levels*['..'])) / path_to.relative_to(root)
+        return ModuleReturnValue(str(path), [])
```
eli-schwartz, line 156: This extremely hairy, scary code seems to be entirely a result of a single devastating design flaw:

`os.path.relpath()` is the one true function for computing relations between paths; meanwhile, pathlib is often opinionatedly wrong and this is one of those cases. The proper situation in which pathlib should be used is as an os.PathLike implementation passed to `os.path.*` functions, and:
- `relative_to()`
- in general any concrete Path() (as opposed to the PurePath() functions which do not provide I/O) method is probably horrifically broken and should be avoided, most particularly .exists() as it may arbitrarily traceback instead of returning False when the answer is "it doesn't exist"

eli-schwartz (changes requested):
This code seems very complicated and confusing to use, and it feels like that is mostly because it erroneously considers pathlib to be a good thing.

mesonbuild/modules/fs.py
zeehio: I agree, however I was asked to move to `pathlib` on the first review here: https://github.com/mesonbuild/meson/pull/7908#issuecomment-717816900. That led me down a rabbit hole in pathlib reaching a python enhancement pull request yet to be merged... (Also liked in this PR)

To be fully honest, after seeing differences of opinions across meson maintainers without a clear path forward for me to follow and get the feature merged, I paused pushing for this. This has also blocked my adoption of meson on a project where I wanted to use this feature.

I understand everyone's opinions and you all have very valid points. I also appreciate the care and time you all put into your review. I just did not have the time to keep pushing for something that I had no idea if it would ever be merged.

Hopefully there will at some point be a consensus of what the meson interface should be and the requirements for the behaviour in corner cases. Once that's clear, then I --or anyone else interested-- will be able to provide a straightforward implementation.

Thanks to all of you for meson and for all your time. I feel bad for criticising your work when you are all giving it for free to the community, I hope you understand I say this with my best intentions, hoping that it helps you to consider possible management improvements in the future. If I am the only one feeling like this, please ignore me. But please take note if you find more people feeling the same way.

zeehio:
After some time upstream, python has merged 

```python
import pathlib
p = pathlib.Path("/a").relative_to("/b", walk_up=True)
```

For python 3.12

- https://github.com/python/cpython/pull/19813


I have backported the pull request at:

    python -m pip install backports-pathlib-relative-to

```python
from backports.pathlib import pathlib

pathlib.Path("/a").relative_to("/b", walk_up=True)
```

Backporting this `walk_up=` argument back to python 3.7.

If you agree to that backported dependency I'd be happy to rebase the pull request one more time.

eli-schwartz:
No backports, Meson needs to run on systems with python 3.7 and before pip is available.

Can we just use:
```
path_1 = pathlib.Path("/a")
path_2 = pathlib.Path("/b")

os.path.relpath(path_1, path_2)
```

as I suggested back in April?

docs/markdown/Fs-module.md
zeehio: Done

docs/markdown/Fs-module.md
zeehio: Since @bonzini was proposing the `within` argument here 

- https://github.com/mesonbuild/meson/issues/2184#issuecomment-674451979

he probably can see useful use cases to justify this

mesonbuild/modules/fs.py
zeehio: I changed it to `os.path.relpath()`

docs/markdown/Fs-module.md
zeehio: I guess sometimes one would prefer to have a relative path if possible, but would accept an absolute path otherwise rather than an error

docs/markdown/Fs-module.md
zeehio: @bonzini Do we need the `within` if we have `allow_absolute`?

zeehio: Ah yes, I see what you expect to have with each of those arguments

docs/markdown/Fs-module.md
zeehio: In this other comment https://github.com/mesonbuild/meson/pull/7908#discussion_r577851659,  @bonzini suggests to have both `within` and `allow_absolute`. `within` is convenient for his examples in that comment.

docs/markdown/Fs-module.md
zeehio: In my opinion, both the `within` and `allow_absolute` arguments are a consequence of other design decisions of the meson language:

- `allow_absolute` would not be necessary if it was easy to do something like:

```
try:
    something = fs.relative_to(path1, path2)
except:
    something = path1
```

- `within` would not be necessary if it was easy to do:

```
if fs.is_relative_to(path1, within):
    path1 = path1.relative_to(path2)
```

Are you sure you want to use arguments in `fs.relative_to()` to implement those features instead of working towards a `try/except` and an `is_relative_to()` functionality? It seems to me you are trying to fix a language issue with arguments, but maybe I miss some overarching goal

zeehio: This is the `relative_to` current behaviour from a functional perspective. I have implemented most of your suggestions:

### relative_to

*since 0.64.0*

Given two paths, returns a version of the first path relative to the second.

Examples: 

```meson
# On windows:
fs.relative_to('c:\\proj1\\foo', 'c:\\proj1\\bar')  # '..\\foo'
fs.relative_to('c:\\proj1\\foo', 'd:\\proj1\\bar')  # ERROR, since the relative path does not exist

# On other systems:
fs.relative_to('/prefix/lib', '/prefix/bin')  # '../lib'
fs.relative_to('/prefix/lib/foo', '/prefix')  # 'lib/foo'
```

`relative_to` has some optional keyword arguments:

- `native`: Can be set to `true` or `false`. Whether the relative paths are computed
  following the rules of the build machine (native: `true`) or the host (native: `false`).

- `allow_absolute`: If set to `true`, instead of giving an error, `relative_to()`
  will return the original path.  Useful if an absolute path is a reasonable
  fallback.

- `if_within`: A path. `relative_to(path1, path2, within: path3)` returns
  the relative path of `path1 with respect to `path2` if `path1` is found in `path3`,
  otherwise it returns `path1` unchanged.

Examples: 

```meson
# On windows:
fs.relative_to('c:\\proj1\\foo', 'c:\\proj2\\bar', if_within: 'c:\\proj2')  # ERROR, the relative path is outside the "within" argument
fs.relative_to('c:\\proj1\\foo', 'c:\\proj2\\bar', if_within: 'c:\\proj2', allow_absolute: true)  # 'c:\\proj1\\foo'
fs.relative_to('c:\\proj1\\foo', 'd:\\proj1\\bar', allow_absolute: true)  # 'c:\\proj1\\foo'

# On other systems:
fs.relative_to('/project1/lib/foo', '/usr/bin', if_within: '/usr')  # ERROR: '/project1/lib/foo' not in '/usr'
fs.relative_to('/project1/lib/foo', '/usr/bin', if_within: '/usr', allow_absolute: 'true')  # '/project1/lib/foo'
fs.relative_to('/usr/lib/foo', '/usr/bin', if_within: '/usr')  # '../lib/foo'
```

docs/markdown/Fs-module.md
zeehio: `if_within` needs to be a path because you may want a relative directory as long as it is within a specific subdirectory, and otherwise error or return the original absolute path. For further functional discussion I would suggest this conversation https://github.com/mesonbuild/meson/pull/7908#discussion_r555361018

docs/markdown/Fs-module.md
```diff
@@ -216,6 +215,51 @@ fs.stem('foo/bar/baz.dll.a')  # baz.dll
...
+`relative_to` has some optional keyword arguments:
+
+- `native`: Can be set to `true` or `false`. Whether the relative paths are computed
+  following the rules of the build machine (native: `true`) or the host (native: `false`).
+
+- `allow_absolute`: If set to `true`, instead of giving an error, `relative_to()`
+  will return the original path. Useful if an absolute path is a reasonable
+  fallback.
+
+- `if_within`: A path. `relative_to(path1, path2, within: path3)` returns
```
xclaesse, line 245:
```suggestion
- `if_within`: A path. `relative_to(path1, path2, if_within: path3)` returns
```

docs/markdown/Fs-module.md
```diff
@@ -216,6 +215,51 @@ fs.stem('foo/bar/baz.dll.a')  # baz.dll
...
+  the relative path of `path1 with respect to `path2` if `path1` is found in `path3`,
+  otherwise it returns `path1` unchanged.
+
+Examples: 
+
+```meson
+# On windows:
+fs.relative_to('c:\\proj1\\foo', 'c:\\proj2\\bar', if_within: 'c:\\proj2')  # ERROR, the relative path is outside the "within" argument
```
xclaesse, line 253: Is that an error? Doc above says " if `path1` is found in `path3`, otherwise it returns `path1` unchanged", so shouldn't this return `c:\\proj1\\foo`?

zeehio: The example was right, the docs were wrong.

If `path1` is not found on `path3`, one can do two things: either we return `path1` or we error.

- Choosing to always error is the easy path, but it is not convenient to users because as far as I know meson doesn't have a try/catch.
- Returning `path1` as was documented may lead to confusion if both `path1` and `path2` happen to be relative paths. In that scenario we could not always know if the returned path is just `path1` we gave because it wasn't contained in `path3` or if it was relative to `path2` because it was contained in `path3`. Therefore the only safe behaviour is to return `path1` if it is absolute, or raise an error. But for consistency, absolute paths can only be returned if `allow_absolute : true`, so that needs to be set as well.

I've pushed a commit to clarify the documentation as well

tristan957:
My PR has been merged. If you think you need other functionality, rebase it on master.

PR #7909, opened 2020-10-28
Relax restrictions on detecting boost dependency

kjmeagher:
This does two things:
* allows the .so files to be symlinks
* searches `lib` and `lib64` in `BOOST_ROOT` even if it finds lib
  directories from the compiler

The first condition is needed for the homebrew on macs because certian
boost modules are provided by seperate packages and are put together in
a directory with symlinks to the .so files. The both conditions are
needed for high performace computing environments where dependencies are
often provided in nonstandard directories with symlinks

### DIFF
--- a/mesonbuild/dependencies/boost.py
+++ b/mesonbuild/dependencies/boost.py
@@ -382,7 +382,7 @@ class BoostDependency(SystemDependency):
         # Finally, look for paths from .pc files and from searching the filesystem
         self.detect_roots()
 
-    def check_and_set_roots(self, roots: T.List[Path]) -> None:
+    def check_and_set_roots(self, roots: T.List[Path], use_system: bool) -> None:
         roots = list(mesonlib.OrderedSet(roots))
         for j in roots:
             #   1. Look for the boost headers (boost/version.hpp)
@@ -394,7 +394,7 @@ class BoostDependency(SystemDependency):
             if not inc_dirs:
                 continue
 
-            lib_dirs = self.detect_lib_dirs(j)
+            lib_dirs = self.detect_lib_dirs(j, use_system)
             self.is_found = self.run_check(inc_dirs, lib_dirs)
             if self.is_found:
                 self.boost_root = j
@@ -436,7 +436,7 @@ class BoostDependency(SystemDependency):
         if paths and any([not x.is_absolute() for x in paths]):
             raise DependencyException('boost_root path given in machine file must be absolute')
 
-        self.check_and_set_roots(paths)
+        self.check_and_set_roots(paths, use_system=False)
 
     def run_check(self, inc_dirs: T.List[BoostIncludeDir], lib_dirs: T.List[Path]) -> bool:
         mlog.debug('  - potential library dirs: {}'.format([x.as_posix() for x in lib_dirs]))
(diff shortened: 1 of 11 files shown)


mesonbuild/dependencies/boost.py
```diff
@@ -576,9 +576,6 @@ def detect_lib_dirs(self, root: Path) -> T.List[Path]:
         system_dirs = [x for x in system_dirs if mesonlib.path_is_in_root(x, root)]
         system_dirs = list(mesonlib.OrderedSet(system_dirs))
 
-        if system_dirs:
-            return system_dirs
-
```
mensinda, line 581: Removing this effectively makes the above code where `system_dirs` useless, which might break existing setups.

mesonbuild/dependencies/boost.py
```diff
@@ -652,7 +649,7 @@ def filter_libraries(self, libs: T.List[BoostLibraryFile], lib_vers: str) -> T.L
     def detect_libraries(self, libdir: Path) -> T.List[BoostLibraryFile]:
         libs = []  # type: T.List[BoostLibraryFile]
         for i in libdir.iterdir():
-            if not i.is_file() or i.is_symlink():
```
mensinda, line 655: Symlinks were excluded on purpose to avoid duplicate libraries in the detection logic and because symlinks usually remove information. Please either ensure that all symlinks are resolved or that "real" paths of a library are always preferred.

mensinda (changes requested):
I have nothing against these changes in principle, but the implementation needs some work.

Also, a test case would be nice.

deepbluev7:
The symlink thing was something that confused me for quite a while. I used cget to build boost, which by default symlinks all libraries into a directory. This means there are only symlinked libraries in the lib folder. I worked around this for now, but this PR would have saved me quite a bit of time. I do agree though, that non-symlinked libraries should be preferred.

mesonbuild/dependencies/boost.py
kjmeagher: I changed it to use both the system_dirs and the hard-coded lib directories

mesonbuild/dependencies/boost.py
kjmeagher: I changed it to resolve all the symlinks and use set() to remove duplicates

eli-schwartz: Why not make libs into a set from the very beginning?

kjmeagher: No good reason. I changed it to use a set from the begining

mesonbuild/dependencies/boost.py
```diff
@@ -561,7 +557,7 @@ def detect_lib_dirs(self, root: Path) -> T.List[Path]:
         # Filter out paths that don't match the target arch to avoid finding
         # the wrong libraries. See https://github.com/mesonbuild/meson/issues/7110
         if not self.arch:
-            return dirs + subdirs
+            return system_dirs + dirs + subdirs
```
mensinda, line 560: `system_dirs` should also be returned here in an `elif` branch since they would be ignored otherwise and should not be subject to the filtering logic below

mensinda (commented):
I just got one more issue, but then this is good to go.

mesonbuild/dependencies/boost.py
```diff
@@ -615,15 +611,19 @@ def filter_libraries(self, libs: T.List[BoostLibraryFile], lib_vers: str) -> T.L
...
         for i in libdir.iterdir():
-            if not i.is_file() or i.is_symlink():
+            if not i.is_file():
                 continue
             if not any([i.name.startswith(x) for x in ['libboost_', 'boost_']]):
                 continue
 
-            libs += [BoostLibraryFile(i)]
-        return [x for x in libs if x.is_boost()]  # Filter out no boost libraries
+            libs.add(i.resolve())
```
eli-schwartz, line 621: When I suggested making it into a set from the beginning, I was kind of assuming the logical next step is to consolidate this. You no longer need multiple complex filtering passes, so...

```suggestion
            libs.add(BoostLibraryFile(i.resolve()))
```

And then you don't need to assign blibs and make the return line refer to a different variable. This would help reduce the size of the diff... :)

mesonbuild/dependencies/boost.py
kjmeagher: After reading #7110 I think I might be going about this wrong let me try a different approach

kjmeagher: I re-wrote it to work explicitly what I was trying to do. Now when searching BOOST_ROOT it won't use the system_dirs, but when looking in other directories like /usr/local it will

mesonbuild/dependencies/boost.py
kjmeagher: I made the requested change

mesonbuild/dependencies/boost.py
```diff
@@ -679,11 +682,6 @@ def detect_roots(self) -> None:
         else:
             tmp = []  # type: T.List[Path]
 
-            # Homebrew
-            brew_boost = Path('/usr/local/Cellar/boost')
-            if brew_boost.is_dir():
-                tmp += [x for x in brew_boost.iterdir()]
-
```
mensinda, line 686: What is the reason behind removing this?

kjmeagher: On homebrew systems, boost should be detected in `/usr/local/lib` instead of the cellar. Looking in the cellar made sense before when the logic couldn't follow symlinks. But since this patch adds support for following symlinks, it now finds boost in /usr/local/lib like it should. In addition, event if it somehow got to checking the cellar it wouldn't find boost_python because it is in a different homebrew formula.

mensinda (approved):

PR #7913, opened 2020-10-29
gnome: Handle libraries that are not in the current build dir

ssssam:
The generate_gir() function previously assumed all library
inputs were in the current build dir. This would fail if they
weren't.

My particular use case for this fix was porting Rhythmbox to Meson. (https://gitlab.gnome.org/GNOME/rhythmbox/-/merge_requests/86). The original build system generates GIR files in the `./bindings/gi` subdir, separately from the actual library files. That doesn't work with Meson. I can work around this by building GIR files along with the libraries but ... I don't see a reason that the other approach should be broken.

### DIFF
--- a/mesonbuild/modules/gnome.py
+++ b/mesonbuild/modules/gnome.py
@@ -554,15 +554,16 @@ class GnomeModule(ExtensionModule):
             else:
                 # Because of https://gitlab.gnome.org/GNOME/gobject-introspection/merge_requests/72
                 # we can't use the full path until this is merged.
+                libpath = os.path.join(girtarget.get_subdir(), girtarget.get_filename())
                 if isinstance(girtarget, build.SharedLibrary):
+                    # need to put our output directory first as we need to use the
+                    # generated libraries instead of any possibly installed system/prefix
+                    # ones.
+                    ret += ["-L@BUILD_ROOT@/{}".format(os.path.dirname(libpath))]
                     libname = girtarget.get_basename()
                 else:
-                    libname = os.path.join("@PRIVATE_OUTDIR_ABS_%s@" % girtarget.get_id(), girtarget.get_filename())
+                    libname = os.path.join("@BUILD_ROOT@/{}".format(libpath))
                 ret += ['--library', libname]
-                # need to put our output directory first as we need to use the
-                # generated libraries instead of any possibly installed system/prefix
-                # ones.
-                ret += ["-L@PRIVATE_OUTDIR_ABS_%s@" % girtarget.get_id()]
                 # Needed for the following binutils bug:
                 # https://github.com/mesonbuild/meson/issues/1911
                 # However, g-ir-scanner does not understand -Wl,-rpath
--- a/run_project_tests.py
+++ b/run_project_tests.py
@@ -870,6 +870,10 @@ def skippable(suite, test):
     if test.endswith('15 llvm'):
         return True
 
+    # This test breaks with gobject-introspection <= 1.58.1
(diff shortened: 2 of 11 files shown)


mesonbuild/modules/gnome.py
```diff
@@ -553,15 +553,16 @@ def _scan_gir_targets(self, state, girtargets):
             else:
                 # Because of https://gitlab.gnome.org/GNOME/gobject-introspection/merge_requests/72
                 # we can't use the full path until this is merged.
+                libpath = os.path.join('/', girtarget.get_subdir(), girtarget.get_filename())
                 if isinstance(girtarget, build.SharedLibrary):
+                    # need to put our output directory first as we need to use the
+                    # generated libraries instead of any possibly installed system/prefix
+                    # ones.
+                    ret += ["-L@BUILD_ROOT@%s" % os.path.dirname(libpath)]
```
jpakkane, line 561: Do not start `libpath` with `/`. It is very confusing because it looks like you are building an absolute path instead. Have it in the format string instead. Text formatting should also be done with format rather than `%` as we are trying to phase out the latter.

Thus this should be something like:

```meson
libpath = os.path.join(girtarget.get_subdir(), girtarget.get_filename())
# and
ret += ["-L@BUILD_ROOT@/%s".format(os.path.dirname(libpath))]
```

ssssam: Thanks, that is better.

ssssam:
This breaks on Ubuntu Bionic:

```
Build-time dependency gobject-introspection-1.0 found: YES 1.56.1
PKG_CONFIG_PATH: 
Called `/usr/bin/pkg-config --variable=g_ir_scanner gobject-introspection-1.0` -> 0
/usr/bin/g-ir-scanner
Got pkgconfig variable g_ir_scanner : /usr/bin/g-ir-scanner
Program g_ir_scanner found: YES (/usr/bin/g-ir-scanner)
PKG_CONFIG_PATH: 
Called `/usr/bin/pkg-config --variable=g_ir_compiler gobject-introspection-1.0` -> 0
/usr/bin/g-ir-compiler
Got pkgconfig variable g_ir_compiler : /usr/bin/g-ir-compiler
Program g_ir_compiler found: YES (/usr/bin/g-ir-compiler)

test cases/frameworks/34 gir static lib/subdir/gir/meson.build:17:6: ERROR: Static libraries can only be introspected with GObject-Introspection >=1.58.1
```

Seems the issue is that the testcase I've added can't be built with old versions of gobject-introspection. So perhaps the solution is to skip the test in that situation -- is there a preferred way to do this?

jpakkane:
You'll probably need to use `run_command`, mangle the output and use `version_compare`.

ssssam:
I tried to do this, as follows:

   * add `if test.endswith('34 gir static lib'): return True` in `run_project_tests.py`
   * check version of `gobject-introspection-1.0` in meson.build
   * fail with `error('MESON_SKIP_TEST gobject-introspection is too old to support static libraries')` if it's too old

For some reason this is reported [as an error](https://github.com/mesonbuild/meson/pull/7913/checks?check_run_id=1388027356) instead of skipping the test -- am I missing something obvious here?

jpakkane:
There is a failsafe. You need to add that test directory to the list of tests that are allowed to skip tests. I don't remember offhand where it is, but grepping should find it fairly fast.

ssssam:
Aha, I found the failsafe already, but then I added it below the `if (linux) return False` line. :facepalm:

PR #7914, opened 2020-10-29
Feature/extend string checks with isupper islower functions

brainelectronics:
As functions to convert strings to either upper or lower cases exist, there should also be a function to check whether a string is actually upper or lower case, as it is available in python.

### DIFF
--- a/docs/markdown/Reference-manual.md
+++ b/docs/markdown/Reference-manual.md
@@ -2058,6 +2058,10 @@ are immutable, all operations return their results as a new string.
 
 - `to_upper()`: creates an upper case version of the string.
 
+- `is_lower()`: returns true if all characters in the string are lowercase.
+
+- `is_upper()`: returns true if all characters in the string are uppercase.
+
 - `underscorify()`: creates a string where every non-alphabetical
   non-number character is replaced with `_`.
 
--- a/docs/markdown/Syntax.md
+++ b/docs/markdown/Syntax.md
@@ -202,6 +202,14 @@ upper = target.to_upper() # t now has the value 'X86_FREEBSD'
 lower = target.to_lower() # t now has the value 'x86_freebsd'
 ```
 
+#### .is_upper(), .is_lower()
+
+```meson
+target = 'x86_freebsd'
+upper = target.is_upper() # boolean value 'false'
+lower = target.is_lower() # boolean value 'true'
+```
+
 #### .to_int()
 
 ```meson
--- a/docs/markdown/snippets/additional_string_checks.md
+++ b/docs/markdown/snippets/additional_string_checks.md
@@ -1,0 +1,4 @@
+## Additional checks for strings
+
+Strings can now be checked if they contain only uppercase or only lowercase
+letterss
--- a/docs/markdown/snippets/additional_string_checks.md
+++ b/docs/markdown/snippets/additional_string_checks.md
@@ -1,4 +1,4 @@
 ## Additional checks for strings
 
 Strings can now be checked if they contain only uppercase or only lowercase
-letterss
+letters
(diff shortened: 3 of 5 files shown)


jpakkane:
What is the use case for this? When and how would you need to know whether a string is in a certain case?

brainelectronics:
we use this to perform specific checks in our build system. There are upper case configs, which are predefined by DevOps and need to be selected from a specific content range, like controller type and some flags, and others, user defined configs, which are free to choose.
We currently need to maintain a list of "DevOps" provided configs and values, and crawl the config for those, see my other PR #7887. This is work we would like to give away to the backend.
With this PR we could crawl them as before and evaluate the configs value only if the key is upper case and apply the config value without checking otherwise

jpakkane:
> There are upper case configs, which are predefined by DevOps and need to be selected from a specific content range, like controller type and some flags

Depending on the specifics of your setup, a bettter solution could be to have those in a [project option](https://mesonbuild.com/Build-options.html). That way Meson will automatically enforce that the value is legitimate, you don't have to do that yourself later on with string matching.

brainelectronics:
I'll have a look at that in more detail. But it looks like a big breaking change if we would use the build options. 

Currently we use something like the attched code for a project. More than one of these can be added per project, e.g. we copy the `general_config` for `sample_app_DEBUG` and just change the `BUILD_TYPE` to debug. Just an example, so we get two applications with the same modules and configs, except the DEBUG flag. 
As our users can configure all of this we just want to check the devops stuff (which is in upper case) or to check everything of the user, which we aligned on has to be lower case

```meson
target_name = 'sample_app'

general_config = configuration_data({
    'target_file_version' : '5.0.0',        # Version of this target meson file
    'controller': 'STM32F071xB',            # Defines the controller that is used. Based on the controller the correct compile flags are added. e.g, stm32f302xc, stm32f10x_md
    'BUILD_TYPE' : 'release',               # release or debug, influences which vendor library will be linked
    'BUILD_UNIT_TESTS' : true,              # Builds the native unit tests for this target
    'BUILD_CONTROLLER_UNIT_TESTS' : false,  # Builds the controller specific unit tests for this target
    'BUILD_APPLICATION' : true,             # Builds the files as an application
    'BUILD_LIBRARY' : false,                # Builds the files as a library
    'INCLUDE_VENDOR_LIB' : false,           # Include vendor lib for build, only required for ARM Cortex Cores (M0, M3, M4, M7) not for specific like STM32F071xB
    'RUN_PREBUILD' : true,                  # Run prebuild script before compilation of application
    'PREBUILD_SCRIPT' : '/../buildsystem/prebuild.py',  # path relative to meson build_root
    'PREBUILD_OPTIONS' : '--a --b --c',     # Required if 'RUN_PREBUILD' is set as true, set as '' if no options used
    'RUN_POSTBUILD' : true,                 # Run posbuild script after compilation of application
    'POSTBUILD_SCRIPT' : '/../buildsystem/postbuild.py',  # path relative to meson build_root
    'POSTBUILD_OPTIONS' : '--d --e --f',     # Required if 'RUN_POSTBUILD' is set as true, set as ''

    # Include some modules
    'dio' : true,                       # digital_io.c
    'irq' : true,                       # interrupt.c
    # not mentioned modules or modules set to false won't be used
    # Add further files based on your needs
})

# apply general_config for target_name
application_targets += {target_name : general_config}

# something more here

another_target_name = 'sample_app_DEBUG'
# copy content of general_config
another_config = general_config
another_config.set('BUILD_TYPE', 'debug')
# cflags and cflags_native are extended with '-g' if 'BUILD_TYPE' is 'debug'

application_targets += {another_target_name : another_config}
```

brainelectronics:
In the end it would also be consistent with the already existing function `to_upper()` and `to_lower()`

PR #7922, opened 2020-10-31
Interpreter: Fix nested subsubproject detection

xclaesse:
A sub-subproject can be configured directly from
`subprojects/foo/subprojects/bar/` in the case `bar` is in the same git
repository as `foo` and not downloaded separately into the main
project's `subprojects/`. In that case the nested subproject violation
code was wrong because it is allowed to have more than one "subprojects"
in path (was not possible before Meson 0.56.0).

Example:
- self.environment.source_dir = '/home/user/myproject'
- self.root_subdir = 'subprojects/foo/subprojects/bar'
- project_root = '/home/user/myproject/subprojects/foo/subprojects/bar'
- norm = '/home/user/myproject/subprojects/foo/subprojects/bar/file.c'

We want `norm` path to have `project_root` in its parents and not have
`project_root / 'subprojects'` in its parents. In that case we are sure
`file.c` is within `bar` subproject.

### DIFF
--- a/mesonbuild/interpreter.py
+++ b/mesonbuild/interpreter.py
@@ -4789,21 +4789,6 @@ This will probably not work.
 Try setting b_lundef to false instead.'''.format(self.coredata.base_options['b_sanitize'].value),
                          location=self.current_node)
 
-    def evaluate_subproject_info(self, path_from_source_root, subproject_dir):
-        depth = 0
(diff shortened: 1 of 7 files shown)


jpakkane:
:crossed_swords: conflicts. Also, should this go to 0.56.1?

PR #7924, opened 2020-11-01
dependencies: Put pkgconfig options before operands

lanodan:
This fixes building with meson when the POSIX behavior of getopt is used (stopping at the first non-option), such as when GNU libc is used with `POSIXLY_CORRECT=1` defined.

### DIFF
--- a/mesonbuild/dependencies/base.py
+++ b/mesonbuild/dependencies/base.py
@@ -869,9 +869,13 @@ class PkgConfigDependency(ExternalDependency):
 
     def _set_libs(self):
         env = None
-        libcmd = [self.name, '--libs']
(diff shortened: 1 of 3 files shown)


xclaesse:
I don't understand why, but that breaks CI... Maybe pkgconf does not support that order?

lanodan:
> I don't understand why, but that breaks CI... Maybe pkgconf does not support that order?

It does, I'm using this patch on my system and I use pkgconf. Sadly the CI is horrible to read.

dcbaker:
I think you're causing that mock function to fall through to the end and implicitly return none

jpakkane:
⚔️ conflicts.

PR #7926, opened 2020-11-02
wrap: Fix git exception not catched by interpreter

xclaesse:
This is a regression introduced in Meson 0.56.0, it was fatal
error when optional dependencies fails to download their fallback
subproject.

### DIFF
--- a/mesonbuild/wrap/wrap.py
+++ b/mesonbuild/wrap/wrap.py
@@ -31,7 +31,8 @@ import textwrap
 from .._pathlib import Path
 from . import WrapMode
 from .. import coredata
-from ..mesonlib import verbose_git, quiet_git, GIT, ProgressBar, MesonException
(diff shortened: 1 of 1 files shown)


mesonbuild/wrap/wrap.py
```diff
@@ -186,6 +187,13 @@ def get_directory(subdir_root: str, packagename: str) -> str:
         return wrap.directory
     return packagename
 
+# Wrapper to convert GitException to WrapException catched in interpreter.
```
dcbaker, line 190: This should be a docstring. Also "caught" not "catched"

dcbaker (approved):
A couple of small nits

mesonbuild/wrap/wrap.py
xclaesse: ok, fixed.

PR #7927, opened 2020-11-02
Condense test directory names again.

jpakkane:


jpakkane:
None of the failures are relevant. Merging this will cause some breakage for outstanding MRs but it can't really be helped. In the future we should probably do this just before rc1.

PR #7928, opened 2020-11-02
vim: Drop backwards compatibility with Vim 5.

a-vrma:
We should do this because:

- Vim 5.8, the last minor release of Vim 5, came out in 2001.
- Even the runtime files maintained in the Vim repository, which are very careful about compatibility, removed Vim 5 support in 2016.
- It means less code paths and less lines of code.

Also I didn't keep the check for `did_meson_syn_inits` because I don't know what exactly that is referring to; there doesn't seem to be a variable with that name in any of the Vim scripts.


dcbaker (approved):
Seems fine to me

Issue #7929, opened 2020-11-03
[0.56.0] sysprof-capture-4.pc only referencing -pthread as Libs.private


DimStar77:
While building an entire GNOME 3.38.1 stack using meson 0.56.0, a glitch showed up.

libsoup now fails to build with a build error like:

```
[   52s] /usr/lib64/gcc/x86_64-suse-linux/10/../../../../x86_64-suse-linux/bin/ld: /usr/lib64/libsysprof-capture-4.a(sysprof-collector.c.o): in function `collector_init_cb':
[   52s] /home/abuild/rpmbuild/BUILD/sysprof-3.38.1/x86_64-suse-linux/../src/libsysprof-capture/sysprof-collector.c:465: undefined reference to `pthread_key_create'
```

I tracked this a bit down, and could identify that sysprof-capture-4.pc differs between sysprof being built with meson 0.55 vs 0.56. The diff is:

```
--- sysprof-capture-4.pc-meson0.55	2020-11-03 11:39:39.612272235 +0100
+++ sysprof-capture-4.pc-meson0.56	2020-11-03 11:39:32.528213799 +0100
@@ -7,5 +7,6 @@
 Name: sysprof-capture-4
 Description: The static capture library for tools that generate profiling capture data
 Version: 3.38.1
-Libs: -L${prefix}/lib64 -lsysprof-capture-4 -pthread
-Cflags: -I${includedir}/sysprof-4 -pthread
+Libs: -L${prefix}/lib64 -lsysprof-capture-4
+Libs.private: -pthread
+Cflags: -I${includedir}/sysprof-4
```

This seems to make sense in relation to the libsoup failure: libsoup is being built as dynamic library, which makes Libs.private to be ignore (only used for static builds)

xclaesse:
It looks like you're static linking `libsysprof-capture-4.a` but don't pass `--static` to pkg-config when you checked the `sysprof-capture-4 ` dependency, so it's missing the `-pthread` from `Libs.private`. Is that possible?

xclaesse:
Hm, but I see that library is only static, so this could be actually 2 bugs:
- When generating pc file for static only libraries meson used to move everything to public in the pc file, not sure why that does not seems to work any more.
- When meson checks for a dependency and can only find a static library, it should re-run pkg-config with --static. I believe that has never worked.

https://gitlab.gnome.org/GNOME/sysprof/-/blob/master/src/libsysprof-capture/meson.build#L46

[...]

PR #7930, opened 2020-11-03
pkgconfig: Make external deps of static library public

xclaesse:
This fix a regression caused by
https://github.com/mesonbuild/meson/pull/7837, it wanted to make
InternalDependency's external_deps private but has side effect of making
StatisLibrary's external_deps private too.

It is technically correct to make them private, but Meson used to make
StaticLibrary deps public so they are usable without `pkg-config
--static` when we know there is only a static library available.

Fixes: #7929.


DimStar77:
I tested the GNOME 3.38.1 build with this meson patch and the issue sysprof/libsoup has disappeared

PR #7931, opened 2020-11-03
Introduce `meson.read_file` to read a file as a string

### LINKED ISSUE #7890: Simplify version extraction from a file
Providing some way to replace the following pattern would be welcome:

```meson
project('foo', 'c',
  version: run_command(
    find_program('python'), '-c', 'import sys;print(open(sys.argv[1]).read())', files('VERSION')
  ).stdout().strip(),
)
```

Drawbacks with this current approach:
- we can't move the `run_command()` in a variable before `project()`, which likely implies this is currently a nasty hack where we're probably not supposed to do that
- some system may have `python3` and not `python`, and the other way around is also true
- using `run_command(find_program('cat'), ...)` instead of `python` would be fatal on Windows
- having a dedicated Python script is quite an overhead, and doesn't address the external subprocess hack to just read a file

**Rationale**

So the current scenario we're in is the following: we are managing our releases through Github, where making a release is basically simply pushing a tag. So we have the following primitive `make-release.sh` script:

```sh
#!/bin/sh

#
# Release process:
# 1. on a clean git state, run this script with the new version as argument
# 2. check the last commit and tag
# 3. git push && git push --tags
#

set -eu

if [ $# -ne 1 ]; then
	echo "Usage $0 <major.minor.micro>"
	exit 1
fi

cd "$(dirname $0)"

if ! git diff-index --quiet HEAD; then
	echo "Git index is not clean"
	exit 1
fi

set -x
VERSION="$1"
echo "$VERSION" > VERSION
git add VERSION
git commit -m "Release $VERSION"
git tag "v$VERSION"
```

The `VERSION` file is basically the channel of communication between this script and `meson.build`

**Alternative considered**

- hot-patching `meson.build` within the release script, but it's not exactly cleaner to say the least
- using `dist` scripts with `meson`; unfortunately we're running git commands at an unfortunate timing, and it involves quite some logistic
- requesting in the release process to patch the `meson.build` with the new version, and then extracting the version with the meson introspection with tools like `jq` in the release script; this creates an annoying dependency to `jq`, and it also makes the release process much more clumsy than just running a script

**How it could look like**

Here are a two usage suggestions that would simplify and make this workflow much more reliable, which I believe are not very intrusive:

- `project('foo', 'c', version: files('VERSION'))` (it's part of the source files after all)
- `project('foo', 'c', version: meson.get_file_content('VERSION'))`

ldrumm:
This is an RFC patch that should fix #7890

This patch introdocues the ability to read files at configure time, but
has some restrictions:
    - binary files are not supported (I don't think this will prove a
    problem, and if people are wanting to do something with binary
    files, they should probably be shelling out to their own script.
    - Only files outside the build directory allowed. This limitation
      should prevent build loops
Given that reading an arbitrary file at configure time can affect the
configuration in almost arbitrary ways, meson should force a reconfigure
when the given file changes. I made this non-configurable, but this is
probably better decided by someone more familiar with meson's core
principles.


mesonbuild/interpreter.py
```diff
@@ -2263,6 +2264,42 @@ def _get_native() -> str:
             else:
                 return _get_native()
 
+    @permittedKwargs({'encoding', 'allow_absolute'})
+    @FeatureNew('meson.get_file_contents', '0.58.0')
+    def get_file_contents_method(
+            self, args: T.Sequence[str], kwargs: dict) -> str:
```
dcbaker, line 2270: that should be `kwags: T.Dict[str, T.Any]

mesonbuild/interpreter.py
```diff
@@ -2263,6 +2264,42 @@ def _get_native() -> str:
...
+        '''
+        Read a file from the source tree and return its value as a decoded
+        string. If the encoding is not specified, the file is assumed to be
+        utf-8 encoded. Paths must be relative by default (to prevent accidents)
+        and are forbidden to be read from the build directory (to prevent build
+        loops)
+        '''
+        if len(args) != 1:
+            raise InterpreterException('expected single positional <path> arg')
+        path = args[0]
```
dcbaker, line 2280: you should enforce this is either a string or a `mesonlib.File`, and that if it's a file it's not built.`

mesonbuild/interpreter.py
```diff
@@ -2263,6 +2264,42 @@ def _get_native() -> str:
...
+            self, args: T.Sequence[str], kwargs: dict) -> str:
+        '''
+        Read a file from the source tree and return its value as a decoded
+        string. If the encoding is not specified, the file is assumed to be
+        utf-8 encoded. Paths must be relative by default (to prevent accidents)
+        and are forbidden to be read from the build directory (to prevent build
+        loops)
+        '''
+        if len(args) != 1:
+            raise InterpreterException('expected single positional <path> arg')
```
dcbaker, line 2279: A `mesonlib.File` object should be acceptable as well.

mesonbuild/interpreter.py
```diff
@@ -2263,6 +2264,42 @@ def _get_native() -> str:
...
+        path = args[0]
+        encoding = kwargs.get('encoding', 'utf-8')
```
dcbaker, line 2281: enforce that this is a string

mesonbuild/interpreter.py
```diff
@@ -2263,6 +2264,42 @@ def _get_native() -> str:
...
+        allow_absolute = kwargs.get('allow_absolute', False)
+        source_dir = self.interpreter.environment.get_source_dir()
+        build_dir = self.interpreter.environment.get_build_dir()
+        if not allow_absolute and os.path.isabs(path):
+            raise InterpreterException('paths must be relative')
```
dcbaker, line 2286: probably should add `unless the allow_abslolute` keyword is set. I'm not sure that we really need a public keyword for this though.

mesonbuild/interpreter.py
```diff
@@ -2263,6 +2264,42 @@ def _get_native() -> str:
...
+        path = os.path.join(source_dir, path)
+        if mesonlib.path_is_in_root(Path(path), Path(build_dir), resolve=True):
+            raise InterpreterException('path must not be in the build tree')
+        try:
+            with open(path, 'rb') as f:
+                data = f.read().decode(encoding)
```
dcbaker, line 2292: you can save yourself an extra decode by passing the encoding to `open()`:

```python
with open(path, 'r', encoding=encoding) as f:
     data = f.read()`
```
And you really don't need the rest of this in the context manager or in the try block.

mesonbuild/interpreter.py
```diff
@@ -2263,6 +2264,42 @@ def _get_native() -> str:
...
+                # Reconfigure meson when this file changes as it can contain
+                # data used by any part of the build configuration (e.g.
+                # `project(..., version: meson.get_file_contents('VERSION')` or
+                # `configure_file(...)`
+                self.interpreter.add_build_def_file(path)
+                return data
+        except UnicodeDecodeError:
+            raise InterpreterException("decoding failed for %r" % path)
```
dcbaker, line 2300: I think we're making an effort not to use the old % formatters. Since we have python 3.6 now you can do:
```python
f'decoding fail for {path}'
```

test cases/common/241 get_file_contents/meson.build
```diff
@@ -0,0 +1,5 @@
+project(
+  'get-file-contents',
+  'c',
+  version: meson.get_file_contents('VERSION').strip(),
+)
```
dcbaker, line 5: probably should add:

```meson
assert(meson.project_version().version_compare('== 1.2.0'), 'version was not rad correctly)
```

docs/markdown/Reference-manual.md
```diff
@@ -1958,6 +1958,17 @@ the following methods.
...
 
+- `get_file_contents(path, encoding: 'utf-8', allow_absolute: false)`: return a
+  [string](Syntax.md#strings) with the contents of the given path. If the
+  encoding keyword argument is not specified, the file specified by `path` is
+  assumed to be utf-8 encoded. Binary files are not supported.
+  The provided paths should be relative to the current
+  `meson.current_source_dir()` unless `allow_absolute` is given as `true`, in
+  which case any accessible path outside the build directory is accepted. If the
+  file specified by `path` changes, this will trigger meson to reconfigure the
+  project.
```
dcbaker, line 1969: needs `*(new in 0.57.0)*`

dcbaker (changes requested):
This looks really useful, and generally correct. There's a bunch of type enforcement that should be going on that isn't, which is easy to overlook because so much of the existing code doesn't do it. Thanks for writing the reference-manual documentation. It might be nice to add something to the `How do I..." page as well.

I noticed that `mesonlib.File` objects (returned by the mesonlang `files()` function) aren't supported, but they should be. I also wonder if this would be better called `read_file`? I know that's a "what color to paint the bikeshed kind of question".

test cases/common/241 get_file_contents/meson.build
ldrumm: good idea

mesonbuild/interpreter.py
ldrumm: This is deliberate. I keep the rest in the try block because I don't want to create a dependency on the file if we fail to decode it

mesonbuild/interpreter.py
ldrumm: I didn't realise we had access to f-strings. Will update

docs/markdown/Reference-manual.md
ldrumm: will do

ldrumm:
> This looks really useful, and generally correct. There's a bunch of type enforcement that should be going on that isn't, which is easy to overlook because so much of the existing code doesn't do it. Thanks for writing the reference-manual documentation. It might be nice to add something to the `How do I..." page as well.

Yes. I think the "how do I" would be useful - including the example from the linked #7890 is probably a good bet. I'll add a section.
> 
> I noticed that `mesonlib.File` objects (returned by the mesonlang `files()` function) aren't supported, but they should be. 
I've not actually used them yet myself, but I'll take a look and add them to the next iteration

> I also wonder if this would be better called `read_file`? I know that's a "what color to paint the bikeshed kind of question".

Yes. I'm not keen on the name either. `read_file` probably *is* better. I just copied from the linked ticket

mesonbuild/interpreter.py
dcbaker: But you raise an uncaught exception if the decode fails, so no dependency will be added. Unless I'm missing something.

ldrumm:
On Tue Nov 3, 2020 at 11:22 PM GMT, Dylan Baker wrote:
> @dcbaker commented on this pull request.
>
>
>
> > +        '''
> + if len(args) != 1:
> + raise InterpreterException('expected single positional <path> arg')
> + path = args[0]
> + encoding = kwargs.get('encoding', 'utf-8')
> + allow_absolute = kwargs.get('allow_absolute', False)
> + source_dir = self.interpreter.environment.get_source_dir()
> + build_dir = self.interpreter.environment.get_build_dir()
> + if not allow_absolute and os.path.isabs(path):
> + raise InterpreterException('paths must be relative')
> + path = os.path.join(source_dir, path)
> + if mesonlib.path_is_in_root(Path(path), Path(build_dir),
> resolve=True):
> + raise InterpreterException('path must not be in the build tree')
> + try:
> + with open(path, 'rb') as f:
> + data = f.read().decode(encoding)
>
> But you raise an uncaught exception if the decode fails, so no
> dependency will be added. Unless I'm missing something.
>
Yep. I saw it's cleaner just to do this after the block. I was just being
intransigent
My next iteration has this:

```python

        try:
            with open(path, 'r', encoding=encoding) as f:
                data = f.read()
        except UnicodeDecodeError:
            raise InterpreterException(f'decoding failed for {path}')
        # Reconfigure meson when this file changes as it can contain data used
        # by any part of the build configuration (e.g. `project(..., version:
        # meson.read_file('VERSION')` or `configure_file(...)`
        self.interpreter.add_build_def_file(path)
        return data
```

dcbaker:
That looks like what I'd expect. I'm just very cautious putting things in try blocks, I've been bitten many times by a large try block catching things I didn't expect it to.

mesonbuild/interpreter.py
ldrumm: done

mesonbuild/interpreter.py
ldrumm: done

mesonbuild/interpreter.py
ldrumm: done. Please check the logic here (especially around `subdir` handling. I added this to the tests, but I'd like it double-checked

mesonbuild/interpreter.py
ldrumm: done

mesonbuild/interpreter.py
ldrumm: done. I think the `allow_absolute` is a useful escape hatch with defaults that prevent accidents, but I'm happy to remove this logic if you're sure it's not warranted

test cases/common/241 get_file_contents/meson.build
ldrumm: done. More tests added too

mesonbuild/interpreter.py
ldrumm: done

mesonbuild/interpreter.py
dcbaker: I don't know for sure, but we don't have any other enforcement of absolute/not absolute. @jpakkane?

docs/markdown/Reference-manual.md
```diff
@@ -1958,6 +1958,17 @@ the following methods.
...
 
+- `read_file(path, encoding: 'utf-8', allow_absolute: false)`: return a
+  [string](Syntax.md#strings) with the contents of the given `path`. If the
+  `encoding` keyword argument is not specified, the file specified by `path` is
+  assumed to be utf-8 encoded. Binary files are not supported. The provided
+  paths should be relative to the current `meson.current_source_dir()` unless
+  `allow_absolute` is given as `true`, in which case any accessible path outside
+  the build directory is accepted. If the file specified by `path` changes, this
+  will trigger meson to reconfigure the project. If the file specified by `path
+  is a meson `files()'
```
dcbaker, line 1969: the quoting here looks funny

mesonbuild/interpreter.py
```diff
@@ -2263,6 +2264,66 @@ def _get_native() -> str:
             else:
                 return _get_native()
 
+    @permittedKwargs({'encoding', 'allow_absolute'})
+    @FeatureNew('meson.read_file', '0.58.0')
+    def read_file_method(
+            self, args: T.Sequence[str], kwargs: T.Dict[str, T.Any]) -> str:
```
dcbaker, line 2270: args needs to be updated. iIRC there's a `FileorString` alias in mesonlib for this.

dcbaker (changes requested):
This all looks correct. There's something off in the release snippet around quoting, and this still needs a release notes snippet

dcbaker (approved):
Looks good to me. Thanks for all the work!

mesonbuild/interpreter.py
xclaesse: IMHO we should drop allow_absolute, most APIs transparently accept both relative and absolute AFAIK.

dcbaker:
Ah, you need to quote "FileOrString", it's only declared when type checking

jpakkane:
This functionality is useful on its own but if we look at the main usage:

> project('myproj', 'c', 'cpp', version: meson.read_file('VERSION').strip())

This is utilitarian (as in it works) but it is not particularly pretty. Since this is a fairly common thing to want to do, it would be better if the UI was smoother (on the eyes at least).

Some alternative implementations:

- If the string passed in to `version` points to a file on disk (i.e. has value `"VERSION"`) then read the contents of that file as the version (as here)
- add a `version_file` kwarg, and if both `version` and `version_file` are specified Meson reads both and exits with a hard error if they have different values (I'm told this is needed for backwards compatibility for some projects)

In both of these cases the actual functionality of loading files should then either go to the `fs` module or as a method on a `file` object.

jpakkane:
Oh, and since strings in Meson are UTF-8, that is the only text encoding that should be supported. Encoding problems should be a hard error.

jpakkane:
Note that if this is added to the `fs` module you should be able to do something like:

```meson
version: import('fs').load_file('VERSION').strip()
```

dcbaker:
@jpakkane, I disagree on the encoding. This may be useful for wraps, and we had to relax the encoding of configure file inputs because of wraps, and some projects using weird encodings. I'd say let's default to utf-8, but leave support for non-utf-8

jpakkane:
Well the Meson way is to only support utf-8 at start, make everything else a hard error and wait for someone to complain with an actual use case.

jpakkane:
Based on various discussions (mostly on IRC) it seems that people prefer to have the higher level `version_file` kwarg for `project`, so let's go with that.

- if both `version` and `version_file` are specified they *must* have the same version value, if they are different then it should be a hard error
- the version file should be only one line, and it may have a newline at the end of it
- the version file should be added to the list of files, who cause reconfigurations when they are edited

Having a separate general purpose mechanism of reading file contents is also ok, but it should be done in a separate PR and the functionality should be in the `fs` module.

ldrumm:
> Based on various discussions (mostly on IRC) it seems that people prefer
> to have the higher level `version_file` kwarg for `project`, so let's go
> with that.
>
> - if both `version` and `version_file` are specified they *must* have
> the same version value, if they are different then it should be a hard
> error
> - the version file should be only one line, and it may have a newline at
> the end of it
> - the version file should be added to the list of files, who cause
> reconfigurations when they are edited
>
> Having a separate general purpose mechanism of reading file contents is
> also ok, but it should be done in a separate PR and the functionality
> should be in the `fs` module.
>
Sounds reasonable. I should have some bandwidth to do this in a week or so.

tp-m:
> Based on various discussions (mostly on IRC) it seems that people prefer to have the higher level `version_file` kwarg for `project`, so let's go with that.
> 
>     * `version_file`s value should be a string, like `version_file: 'VERSION'`
> 
>     * if both `version` and `version_file` are specified they _must_ have the same version value, if they are different then it should be a hard error

Just in case it hasn't been suggested yet (sorry if I missed the discussion): why not just allow:
```
project('foo', version: file('VERSION'), ...)
```
as well, without adding a new 'version_file` kwarg?

jpakkane:
That works too, though it should be `version: files('VERSION')` as there is no `file` function. Perhaps we should add one?

dcbaker:
I'd really prefer the `version : files(...)` to having a new keyword, but if we have the new keyword I'd prefer to make passing both an error. I don't see any legitimate reason to pass both at the same time.

eli-schwartz:
Simpler `files()` implementation just for project version is available in the just-linked PR.

ldrumm:
On Fri Jan 29, 2021 at 4:43 PM GMT, Jussi Pakkanen wrote:

> Closed #7931 via #8262.
>
> --
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/mesonbuild/meson/pull/7931#event-4267055291


Looks good. At some point soon I'll rewrite this to add a `read()` method to the
`fs` module where it can be used for general purpose file read operations.

jpakkane:
Didn't mean to close this, Github's autocloser did it all by itself. But just open a new one once you are ready.

PR #7932, opened 2020-11-03
Bare exceptions are no longer allowed [NFC]

ldrumm:
The last instances of

    try:
        ...
    except:
        ...

were removed in bf98ffca. The sideci.yml file was updated, but the
flake8 config still allows this. Ensure that flake8 tests fail if this
questionable construct appears again.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -24,8 +24,6 @@ ignore =
     E731,
     # E741: ambiguous variable name 'l'
     E741,
-    # E722: do not use bare except'
-    E722,
     # W504: line break after binary operator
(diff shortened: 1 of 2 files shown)


dcbaker:
This looks good to me. Could you also add this to the `.pylintrc` file? I think it's called `bare-except`.

dcbaker (approved):

ldrumm:
> This looks good to me. Could you also add this to the `.pylintrc` file? I think it's called `bare-except`.

done

Issue #5100, opened 2019-03-17
Rust: Support for edition=2018


strega-nil:
Like `c_std` and `cpp_std`, meson should add an option that allows one to set the `rust_edition` on a per-project basis.

PR #7934, opened 2020-11-03
rust: implement support for --edition

dcbaker:
Using the std option, so now `rust_std=..` will work. I've chosen to use
"std" even though rust calls these "editions", as meson refers to
language versions as "standards", which makes meson feel more uniform.

Fixes: #5100

### DIFF
--- a/docs/markdown/snippets/rust_std_option.md
+++ b/docs/markdown/snippets/rust_std_option.md
@@ -1,0 +1,5 @@
+## Rust now has a a standard option
+
+Rust calls these `editions`, however, meson generally refers to such language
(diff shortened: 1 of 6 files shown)


docs/markdown/snippets/rust_std_option.md
```diff
@@ -0,0 +1,5 @@
+## Rust now has a a standard option
+
+Rust calls these `editions`, but I've chosen to use `std`, as this is what
+meson uses for all other languages, even when those languages use a different
+term. Currently accepted values or 2015 and 2018
```
jpakkane, line 5: Release note snippets should not be written in the first person. Instead they should say something like "but in Meson they are called ...".

PR #7937, opened 2020-11-04
symbolextractor: Add FreeBSD support

monwarez:
This almost the same implementation of OpenBSD, except that FreeBSD version of `nm` have the `-defined-only` option


jpakkane:
Thanks.

PR #7939, opened 2020-11-04
docs: Recommend to use SPDX license identifiers

### LINKED ISSUE #7905: Recommend to use licence identifiers from the SPDX license list
Since #2942 the `project()` function supports declaring a license (or multiple licenses), which is quite useful. However, currently there is no recommended format for specifying licenses:

https://mesonbuild.com/Reference-manual.html#project
> `license:` takes a string or array of strings describing the license(s) the code is under. Usually this would be something like `license : 'GPL2+'`, but if the code has multiple licenses you can specify them as an array like this: `license : ['proprietary', 'GPL3']`. **Note that the text is informal and is only written to the dependency manifest. Meson does not do any license validation, you are responsible for verifying that you abide by all licensing terms.** You can access the value in your Meson build files with `meson.project_license()`.

Unfortunately some license identifiers are ambiguous. E.g. `GPL3` can mean "GNU General Public License v3.0 only" or "GNU General Public License v3.0 or later".

Therefore I propose to update the Meson documentation and all examples to recommend using a license identifier form the [SPDX license list](https://spdx.org/licenses/). For the previous example this would resolve the ambiguity by using either `GPL-3.0-only` or `GPL-3.0-or-later`.

The Software Package Data Exchange® (SPDX®) is a Linux Foundation Project and describes itself as:
> An open standard for communicating software bill of material information, including components, licenses, copyrights, and security references. SPDX reduces redundant work by providing a common format for companies and communities to share important data, thereby streamlining and improving compliance.

More information (advantages, details, etc.) can be found here:
- https://spdx.dev/resources/use/#identifiers
- https://spdx.dev/ids/

If you agree that the SPDX short-form license identifiers would be a good fit I would also offer to make a PR if you want.

primeos:
Some license identifiers are ambiguous (e.g. "GPL3"). The SPDX license
identifiers avoid this by providing standardized and unique identifiers
(e.g. "GPL-3.0-only" or "GPL-3.0-or-later" for the previous example).

Because SPDX short-form identifiers are also both human- and
machine-readable we should recommend them in the documentation.

More information (advantages, details, etc.) can be found here:
- https://spdx.dev/resources/use/#identifiers
- https://spdx.dev/ids/

Fix #7905.

---

I've also updated the test to ensure the "-2.0" suffix won't cause any issues though this is completely optional as I don't see why it should actually cause any issue (but it seemed like a good idea regardless).

### DIFF
--- a/docs/markdown/Reference-manual.md
+++ b/docs/markdown/Reference-manual.md
@@ -1398,13 +1398,14 @@ Project supports the following keyword arguments.
   specific options are used normally even in subprojects.
 
 
-- `license`: takes a string or array of strings describing the
-  license(s) the code is under. Usually this would be something like
-  `license : 'GPL2+'`, but if the code has multiple licenses you can
-  specify them as an array like this: `license : ['proprietary',
-  'GPL3']`. Note that the text is informal and is only written to
-  the dependency manifest. Meson does not do any license validation,
-  you are responsible for verifying that you abide by all licensing
+- `license`: takes a string or array of strings describing the license(s) the
+  code is under. To avoid ambiguity it is recommended to use a standardized
+  license identifier from the [SPDX license list](https://spdx.org/licenses/).
+  Usually this would be something like `license : 'GPL-2.0-or-later'`, but if
+  the code has multiple licenses you can specify them as an array like this:
+  `license : ['proprietary', 'GPL-3.0-only']`. Note that the text is informal
+  and is only written to the dependency manifest. Meson does not do any license
+  validation, you are responsible for verifying that you abide by all licensing
   terms. You can access the value in your Meson build files with
   `meson.project_license()`.
 
--- a/test cases/common/166 get project license/meson.build
+++ b/test cases/common/166 get project license/meson.build
@@ -1,8 +1,8 @@
-project('bar', 'c', license: 'Apache')
+project('bar', 'c', license: 'Apache-2.0')
 
 executable('bar', 'bar.c')
 
 license = meson.project_license()[0]
-if license != 'Apache'
-    error('The license should be Apache, but it is: ' + license)
+if license != 'Apache-2.0'
+    error('The license should be Apache-2.0, but it is: ' + license)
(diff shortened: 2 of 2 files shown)


dcbaker (approved):

docs/markdown/Reference-manual.md
```diff
@@ -1398,13 +1398,14 @@ Project supports the following keyword arguments.
...
-  `license : 'GPL2+'`, but if the code has multiple licenses you can
-  specify them as an array like this: `license : ['proprietary',
-  'GPL3']`. Note that the text is informal and is only written to
-  the dependency manifest. Meson does not do any license validation,
-  you are responsible for verifying that you abide by all licensing
+- `license`: takes a string or array of strings describing the license(s) the
+  code is under. To avoid ambiguity it is recommended to use a standardized
+  license identifier from the [SPDX license list](https://spdx.org/licenses/).
+  Usually this would be something like `license : 'GPL-2.0-or-later'`, but if
+  the code has multiple licenses you can specify them as an array like this:
```
jtojnar, line 1405: As you mention, the array is ambiguous as well so I would first recommend `AND` and `OR` connectives in this sentence. The array syntax can be described with the description that it  is actually informal. Eventually, we might want to deprecate the informal license.

jpakkane:
Thanks. In the future please add `[skip ci]` at the end of the commit message so CI is not run. Thanks.

docs/markdown/Reference-manual.md
primeos: I'm :+1: on that but it's a more minor thing, linking to https://spdx.org/licenses/ already partly covers it (at least if one looks more into the SPDX standard), and I'm not sure if the consensus would be to deprecate the array (as it would obviously be a breaking change for some). Feel free to open another PR/issue for that or briefly continue the discussion in the already closed issue (if the maintainers don't mind; in this case it'd seem ideal for a quick feedback if/what we want to do about this).

PR #7949, opened 2020-11-06
Fix clang-tidy return value reporting

### LINKED ISSUE #7948: Meson does not report clang-tidy exit code if invoked "manually (i.e. if run-clang-tidy(.py) is not in PATH)
Meson does not report clang-tidy exit code if invoked "manually (i.e. if run-clang-tidy(.py) is not in PATH). This is caused by ignoring the return value of `manual_clangformat()` in `clangformat()` within `mesonbuild/scripts/clangtidy.py`.

Flowdalic:
In case clang-tidy is invoked manually, i.e. if run-clang-tidy(.py) is
not found, Meson would not report the return value. This is caused by
ignoring the return value of manual_clangformat() in clangformat()
within mesonbuild/scripts/clangtidy.py.

Even though only more recent-versions of clang-tidy actually report an
non-zero exit code if errors are found, there is no reason Meson
shouldn't simply report any error codes it received from clang-tidy.

Fixes #7948.

### DIFF
--- a/mesonbuild/scripts/clangtidy.py
+++ b/mesonbuild/scripts/clangtidy.py
@@ -49,8 +49,7 @@ def clangformat(srcdir_name: str, builddir_name: str) -> int:
         return subprocess.run([run_clang_tidy, '-p', builddir_name, '^(?!' + re.escape(builddir_name + os.path.sep) +').*$']).returncode
     else:
         print('Could not find run-clang-tidy, running checks manually.')
-        manual_clangformat(srcdir_name, builddir_name)
-    return 0
+        return manual_clangformat(srcdir_name, builddir_name)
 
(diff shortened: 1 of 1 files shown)


Flowdalic:
There is a failed CI job, but it has no failed tests: https://travis-ci.org/github/mesonbuild/meson/jobs/741951389#L8148

Not sure what to make of it. Is this a fluke? Or caused by changes introduced by this PR?

jpakkane:
That count is for integration tests. This fails in the unit test phase earlier:

```
======================================================================
FAIL: test_nested_cmake_rebuild (__main__.AllPlatformTests)
----------------------------------------------------------------------
```

It is unrelated to this change, though, so merging. Thanks.

PR #7952, opened 2020-11-07
Add thinlto support. Closes #7493.

### LINKED ISSUE #7493: Add support for ThinLTO when using Clang and LLD
**Describe the bug**
ThinLTO has significant improvements under the "Full" LTO path.
It scales incredibly well from a memory consumption point of view and gives incremental and threaded linking to a project.
ThinLTO has long been stable with other build systems also supporting it now.
https://clang.llvm.org/docs/ThinLTO.html

**To Reproduce**
Easy enough to reproduce,
`CC=clang CC_LD=lld meson -Db_lto=true <remaining options>`

Meson will only set `-flto` for the compile args in this instance

**Expected behavior**
I expect meson to check which version of clang and lld is being used, and if it is above a certain version (ThinLTO was supported back with clang 3.9, but probably shouldn't be enabled there) then by default enable the thin lto option with `-flto=thin`

Additionally a nice feature might be an option to set LTO flavour. This way people can still build with `-flto=full` for their final releases but keep thin around for their development and regression testing.
`-Db_lto_flavor=full, -Db_lto_flavor=thin`

**system parameters**
* This isn't a cross-building problem
* Linux at the very least effected, MacOS might matter as well, unsure if thinlto is supported under Windows
* Python version doesn't matter
* `git HEAD` for meson, checked the source tree and it isn't supporting thinlto
* Ninja doesn't matter here.

jpakkane:
Dunno if changing the type of `b_lto` is bad or not. In the short term there may be a few issues, but it is "correct" in the long term.

### DIFF
--- a/docs/markdown/snippets/thinlto.md
+++ b/docs/markdown/snippets/thinlto.md
@@ -1,0 +1,27 @@
+## Add support for thin LTO
+
+The `b_lto` option has been updated and now can be set to the value
+`thin`. This enables [thin
+LTO](https://clang.llvm.org/docs/ThinLTO.html) on all compilers where
+it is supported. At the time of writing this means only Clang.
+
+This change is potentially backwards incompatible. If you have
+examined the value of `b_lto` in your build file, note that its type
+has changed from a boolean to a string. Thus comparisons like this:
+
+```meson
+if get_option('b_lto')
+...
+endif
+```
+
+need to be changed to something like this instead:
+
+```meson
+if get_option('b_lto') == 'true'
+...
+endif
+```
+
+This should not affect any comman line invocations as configuring LTO
+with `-Db_lto=true` still works and behaves the same way as before.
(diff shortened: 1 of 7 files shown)


dcbaker:
I've thought about this too, and I think we should make lto a compiler argument, not a built-in. There's not just thin lto, but lto threads, and cases where you might use two languages, but only one supports lto

jpakkane:
I'd say LTO is more like the optimization level option and the semantics are more of "if lto is set, and the current language/compiler supports it, then use it". IIRC you can do LTO between C, C++ and Rust now and it would be inconvenient to have to toggle it on in three different places.

Note that this is set up so that compilers fall back to regular lto if thinlto is not supported.

dcbaker:
I assume you can only lto between LLVM projects? Ie, clang and rustc?

dcbaker (approved):

jpakkane:
> I assume you can only lto between LLVM projects? Ie, clang and rustc?

Yes. I don't think you could ever LTO-combine GCC and LLVM as their intermediate formats are completely different.

PR #7955, opened 2020-11-08
Fix "generator doesn't stop" on Windows

### LINKED ISSUE #7947: "generator doesn't stop" building harfbuzz in MSYS2/MinGW32 on Win7, but works on Win10
**Describe the bug**
I'm using the [media-autobuild suite](https://github.com/m-ab-s/media-autobuild_suite/) (a set of batch and shell scripts which keeps an MSYS2 envirponment up-to-date before building software) to build several audio / video tools, up to an ffmpeg combining previously built libraries of different codecs.
One of the used libraries is harfbuzz, which MABS builds using meson. This worked regularly, even a few weeks ago, but suddenly it started failing (possibly after an MSYS2 update) on one PC which still runs Windows 7, even after completely re-installing MSYS2 and MinGW, but works with the same set of host software on another PC running Windows 10.
There are other projects (e.g. dav1d) which are still built successfully using meson.

**To Reproduce**
[logs.zip](https://github.com/mesonbuild/meson/files/5501015/logs.zip) collected by MABS, I added some possibly meson specific files (most interesting logs will be in harfbuzz-git/build-32bit).

I am not able to produce a minimum source to reproduce this issue, because I am not a C/C++ developer, only a user of the media-autobuild suite. So I can only hope that the logs with the stack trace help to at least guess what may cause this issue, and to learn what it even means, to support the MABS team finding a possible solution.

**Expected behavior**
Prepare and run compilation, as it used to work a few weeks ago.

**system parameters**
* Windows 7 SP1 German, MSYS2/MinGW
* Python 3.8.6-5 provided by MSYS2
* meson 0.56.0-1 provided by MSYS2
* ninja 1.10.1-1 provided by MSYS2

lazka:
When TemporaryDirectory() cleans up on __exit__ it sometimes throws
OSError noting that the dir isn't empty. This happens after the
first yield in this generator and leads to the exception being handled
which leads to a second yield.

contextlib.contextmanager() fails then since the function it wraps is only
allowed to yield once.

Fix this by not yielding again in the error case.

Fixes #7947

### DIFF
--- a/mesonbuild/compilers/compilers.py
+++ b/mesonbuild/compilers/compilers.py
@@ -785,7 +785,7 @@ class Compiler(metaclass=abc.ABCMeta):
             # On Windows antivirus programs and the like hold on to files so
             # they can't be deleted. There's not much to do in this case. Also,
             # catch OSError because the directory is then no longer empty.
-            yield None
+            return
 
     @contextlib.contextmanager
     def cached_compile(self, code: str, cdata: coredata.CoreData, *,


lazka:
imo a TemporaryDirectory alternative using windows_proof_rmtree would be nice.

anthrotype:
Can you please cut a new meson release with this fix? The current stable release from 30 October does not include this. Thanks! 🙏

eli-schwartz:
This is a simple and obvious bugfix and it makes sense to include it in the next maintenance release. Tagged for the milestone so we don't forget.

That being said, this will not be getting released today, sorry.

anthrotype:
It's ok, thank you!

guy-rouillier:
I'm believe I'm encountering this same issue trying to build a different package on msys2.  Did a system update to msys2-runtime 3.1.7-8, same result. meson 0.57.1

```
$ meson --prefix=/usr build
The Meson build system
Version: 0.57.1
Source dir: D:/Projects/xed
Build dir: D:/Projects/xed/build
Build type: native build
Project name: xed
Project version: 2.6.2
C compiler for the host machine: cc (gcc 10.2.0 "cc (Rev9, Built by MSYS2 project) 10.2.0")
C linker for the host machine: cc ld.bfd 2.36.1
Compiler for language c for the build machine not found.
Host machine cpu: x86_64
Found pkg-config: D:\msys64\mingw64\bin/pkg-config.EXE (1.7.3)
  File "D:/msys64/mingw64/lib/python3.8/site-packages/mesonbuild/mesonlib/universal.py", line 1570, in cleanup
    super().cleanup()
  File "D:/msys64/mingw64/lib/python3.8/tempfile.py", line 831, in cleanup
    self._rmtree(self.name)
  File "D:/msys64/mingw64/lib/python3.8/tempfile.py", line 813, in _rmtree
    _shutil.rmtree(name, onerror=onerror)
  File "D:/msys64/mingw64/lib/python3.8/shutil.py", line 740, in rmtree
    return _rmtree_unsafe(path, onerror)
  File "D:/msys64/mingw64/lib/python3.8/shutil.py", line 622, in _rmtree_unsafe
    onerror(os.rmdir, path, sys.exc_info())
  File "D:/msys64/mingw64/lib/python3.8/shutil.py", line 620, in _rmtree_unsafe
    os.rmdir(path)
OSError: [WinError 145] The directory is not empty: 'D:/Projects/xed/build/meson-private/tmpchm7tobp'

```

PR #7960, opened 2020-11-10
wrap: Add support do download into a destination directory

xclaesse:
It is common for projects to want to run their CI in offline mode, or at
least reduce as much as possible download time during each job. Some
projects uses `meson subprojects download` during docker image
generation and store the `subprojects/` directory directly into the
docker image. When a job is run, the cache is copied back into
subprojects/ directory.

One difficulty with that workflow is to exclude
subprojects that are part of the main project's git repository. Keeping
only tarballs for wrap-file also helps reducing docker image size.

### DIFF
--- a/docs/markdown/Subprojects.md
+++ b/docs/markdown/Subprojects.md
@@ -313,6 +313,13 @@ configure can be performed offline. The command-line `meson
 subprojects download` can be used for that, it will download all
 missing subprojects, but will not update already fetched subprojects.
 
+*Since 1.2.0* `--dest` argument can be used to specify in which directory
+subprojects should be downloaded. The destination directory will contain only
+the `packagecache` directory for `wrap-file` and their patches (they are not extracted),
+and complete checkout for other wrap types. This is used to store a cache of
+subprojects (e.g. in CI docker images) that can then be copied into project's
+`subprojects/` directory to build offline (i.e. with `--wrap-mode=nodownload`).
+
 ### Update subprojects
 
 *Since 0.49.0*
--- a/mesonbuild/msubprojects.py
+++ b/mesonbuild/msubprojects.py
@@ -55,6 +55,9 @@ if T.TYPE_CHECKING:
         apply: bool
         save: bool
 
+    class DownloadArguments(Arguments):
+        dest: T.Optional[str]
+
 ALL_TYPES_STRING = ', '.join(ALL_TYPES)
 
 def read_archive_files(path: Path, base_path: Path) -> T.Set[Path]:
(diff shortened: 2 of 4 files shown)


docs/markdown/Subprojects.md
```diff
@@ -292,6 +292,13 @@ offline. The command-line `meson subprojects download` can be used for that, it
 will download all missing subprojects, but will not update already fetched
 subprojects.
 
+*Since 0.56.0* `--dest` argument can be used to specify in which directory
```
jpakkane, line 295: This should be 0.57.

jpakkane:
Test failure seems relevant too:

```
 test cases/common/154 wrap file should not failed/meson.build:6:0: ERROR: Missing Meson file in /__w/meson/meson/test cases/common/154 wrap file should not failed/subprojects/foo-1.0/meson.build
```

jpakkane:
Ping.

tristan957:
So it seems like this PR might be semi-related to removing redirect wraps, which I am still in favor of. I think this could be closed in favor of that other PR.

mesonbuild/msubprojects.py
```diff
@@ -457,6 +460,13 @@ def checkout(self) -> bool:
 
     def download(self) -> bool:
         self.log(f'Download {self.wrap.name}...')
+        options = T.cast('DownloadArguments', self.options)
```
tristan957, line 463: Can you add `from __future__ import annotations` to remove the quotes?

docs/markdown/Subprojects.md
```diff
@@ -313,6 +313,13 @@ configure can be performed offline. The command-line `meson
 subprojects download` can be used for that, it will download all
 missing subprojects, but will not update already fetched subprojects.
 
+*Since 1.2.0* `--dest` argument can be used to specify in which directory
+subprojects should be downloaded. The destination directory will contain only
```
tristan957, line 317: "only the"

tristan957 (commented):
Looks pretty reasonable. Perhaps @eli-schwartz wants to take a look too?

mesonbuild/msubprojects.py
eli-schwartz: No. This isn't an annotation.

eli-schwartz: (Also the file already uses that future import, because all files have it.)

docs/markdown/Subprojects.md
xclaesse: fixed

docs/markdown/Subprojects.md
xclaesse: 1.2.0... 2 years passed...

tristan957:
Looks like this needs a rebase if you can get back to it

tristan957:
Seems related to https://mesonbuild.com/Release-notes-for-1-3-0.html#new-environment-variable-meson_package_cache_dir

xclaesse:
I think `MESON_PACKAGE_CACHE_DIR` supersedes this indeed. There are 2 main differences still:
- We don't currently cache git repositories. 
- We don't have a way to download without extracting wraps. I think that would be better served by adding e.g. `meson subprojects download --cache-only`.

xclaesse:
Let's close this PR, the 2 point above should be dealt in separate PRs.

Issue #7964, opened 2020-11-11
dynamic linking on macos with rust doesn't work


dcbaker:
We do not produce valid cdylibs on macos.

PR #7965, opened 2020-11-11
Migrate osx CI from travis to github actions

dcbaker:
Which will hopefully be more reliable. I've also fixed a number of tests along the way, and enabled rust on macos, only to discover that dynamic linking doesn't work (#7964). I've also split some of the jobs up into smaller jobs (unit tests separate from project tests) this should make human parsing of the jobs easier.


scivision:
This is timely given the new Travis quota, including for open source projects: https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

dcbaker:
Yeah, looks like we'll need migrate all of our jobs off travis

PR #7966, opened 2020-11-11
msvc: enable /std:c17 flag

jpark37:
MS blog post about improved C support: https://devblogs.microsoft.com/cppblog/c11-and-c17-standard-support-arriving-in-msvc/

I have tested this change against some C code that used static_assert, and did not compile previously.

Extends pattern laid out by #7556.


jpark37:
Forgot to create a branch.

Issue #7599, opened 2020-08-15
msvc /std:c11 flag needs some improvements


nirbheek:
Continuing from https://github.com/mesonbuild/meson/pull/7556#discussion_r469265068.

PR #7967, opened 2020-11-11
msvc: enable /std:c17 flag

jpark37:
MS blog post about improved C support: https://devblogs.microsoft.com/cppblog/c11-and-c17-standard-support-arriving-in-msvc/

I have tested this change against some C code that used static_assert, and did not compile previously.

Extends pattern laid out by #7556.

### DIFF
--- a/mesonbuild/compilers/c.py
+++ b/mesonbuild/compilers/c.py
@@ -410,6 +410,9 @@ class VisualStudioLikeCCompilerMixin(CompilerMixinBase):
 
 class VisualStudioCCompiler(MSVCCompiler, VisualStudioLikeCCompilerMixin, CCompiler):
 
+    _C11_VERSION = '>=19.28'
+    _C17_VERSION = '>=19.28'
+
     def __init__(self, exelist: T.List[str], version: str, for_machine: MachineChoice,
                  is_cross: bool, info: 'MachineInfo', target: str,
                  exe_wrapper: T.Optional['ExternalProgram'] = None,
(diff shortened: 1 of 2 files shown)


scivision:
Great! While implementing this, should we also work to fix #7599? The #7556 I made has the shortcomings noted in #7599

jpark37:
Added support for gnu options. I think this should be enough to handle the major regression. Maybe someone else can handle compiler versioning in another PR?

gnu89 ->
gnu99 ->
gnu11 -> /std:c11
gnu17 -> /std:c17
gnu18 -> /std:c17

jpark37:
Copied comment from fd6585ecc41e1c731224cac08a3f6ef6f49ab184.

Left out gnu1x since it's not listed in the documentation: https://mesonbuild.com/Builtin-options.html

scivision:
Okay cool I think part of the issue was perhaps with older visual studio. Perhaps those that were commenting on the bugs introduced by my merged PR will be able to assess this.

jpark37:
I looked around and adapted an existing version check pattern. I verified 19.28 worked, and 19.29 was too high for VS 16.8.0. I hope there aren't any 19.28 C++ compilers under 16.7.x, but I don't want to downgrade my Visual Studio to check.

jpakkane:
The way compiler versions was reworked in a major way, so this now :crossed_swords: conflicts.

jpark37:
Conflicts resolved.

Now that gnu9x/gnu1x are there, I left them in instead of taking them out.

In my testing, the mlog.log line was failing because it was trying to forward the once argument to print, which it doesn't support. I've replaced both mlog.log lines with mlog.log_once, which seems to work.

dcbaker (approved):
LGTM, thanks for rebasing this :)

dcbaker:
The test failures look relevant. I think for old versions you'll need to either skip that test, or change the standard it uses to c99

jpark37:
Attempting to modify the test to use c99 instead of c11. Hopefully that will clean up the test failures.

PR #7969, opened 2020-11-11
MSVC: Don't warn when projects add /Wall flag

drobilla:
See commit message for details.

### DIFF
--- a/mesonbuild/interpreter/interpreter.py
+++ b/mesonbuild/interpreter/interpreter.py
@@ -2807,7 +2807,7 @@ class Interpreter(InterpreterBase, HoldableObject):
         # see e.g.
         # https://github.com/mesonbuild/meson/issues/3275#issuecomment-641354956
         # https://github.com/mesonbuild/meson/issues/3742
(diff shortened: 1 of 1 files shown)


jpakkane:
Looking at MSDN docs, `/Wall` is a setting that has even more stuff than `/W4`, meaning it really should be called `/W5` for consistency. That would break things, since we specify that `warning_level` 4 means "everything". *sigh* Why does _everything_ have to be needlessly complicated?

Could we add the value `everything` to warning level instead that would explicity mean absolutely everything on every compiler?

drobilla:
> Could we add the value everything to warning level instead that would explicitly mean absolutely everything on every compiler?

I hope so (where "absolutely everything" is a little bit tricky but doable, I think).  I'm going to take a stab at that as well, this was just some low-hanging fruit.  That will be a bit of work, though, current meson warning levels are nowhere near everything for any compiler.

drobilla:
Supplanted by https://github.com/mesonbuild/meson/pull/10554

I guess I'll leave this open until the issue is resolved on master one way or another.

PR #7980, opened 2020-11-13
Add doc on how to add the [provide] entry to wrapdb entries. [skip ci]

jpakkane:


dcbaker (approved):

docs/markdown/Adding-new-projects-to-wrapdb.md
```diff
@@ -132,6 +141,24 @@ Under macOS the command is the following:
...
+ogg = ogg_dep
+```
+
+The `ogg` part on the left refers to the dependency name, which should
+be the same as its Pkg-Config name. `ogg_dep` on the right refers to
+the variable in the build definitions that provides the dependency.
+Most commonly it holds the result of a `declare_dependency` call. If a
+variable of that name is not defined, Meson will exit with a hard
+error.
+
```
xclaesse, line 161: Could probably link to https://mesonbuild.com/Wrap-dependency-system-manual.html#provide-section.

PR #5682, opened 2019-07-18
Perl dependency

ailin-nemui:

PR #7981, opened 2020-11-13
Add a perl dependency

dcbaker:
This is still WIP.

### DIFF
--- a/mesonbuild/dependencies/__init__.py
+++ b/mesonbuild/dependencies/__init__.py
@@ -31,6 +31,7 @@ from .misc import (
 )
 from .platform import AppleFrameworks
 from .ui import GnuStepDependency, Qt4Dependency, Qt5Dependency, WxDependency, gl_factory, sdl2_factory, vulkan_factory
+from .languages import PerlDependency
 
 
 # This is a dict where the keys should be strings, and the values must be one
@@ -48,6 +49,9 @@ packages.update({
 
     'boost': BoostDependency,
     'cuda': CudaDependency,
+
+    # programing/scripting languages
+    'perl': PerlDependency,
 
     # per-file
     'coarray': coarray_factory,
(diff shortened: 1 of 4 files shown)


eli-schwartz:
Prior art: #5682

grimmy:
I've done something similar which can be found at https://keep.imfreedom.org/gplugin/gplugin/file/tip/perl5/meson.build.

I stumbled across this because I had to add a `compiler.find_library` for `libperl.so` because some distributions (like Debian) don't ship the `libperl.so` symlink with the normal Perl package.  However, that then broke my script on FreeBSD that use's a non standard path for `libperl.so` and puts it in `/usr/local/lib/perl5/5.32/mach/CORE/libperl.so`.

To deal with that, I was considering parsing the path out of the `ldopts` and passing it to `compiler.find_library` but right now I'm leaning towards a hack of checking if we're on FreeBSD and just not running `compiler.find_library` because we know based on the way that FreeBSD installs packages, that the `ldopts` returned from `perl -MExtUtils::Embed -e ldopts` will work.

I'll probably end up implementation that solution for now, but would love to see this just handled via this pull request.

dcbaker:
@grimmy: I think the biggest thing missing (I can't/wont do) is a test case, if you wanted to provide one I'd be happy to get this ready to merge for 0.57.0

ailin-nemui:
does this work easily with cross compilation?

by the way and embperl dependency is a good first step, but building modules would also be nice to have

dcbaker:
That's a really good question I didn't think to test.

dcbaker:
I can't find any instructions on how to link to perl when cross compiling. I assume you need to run the host perl for that.

ailin-nemui:
for reference here is our current Perl detection code: https://github.com/irssi/irssi/blob/master/meson.build#L315
here is an example for cross compiling Perl to termux-android: https://github.com/irssi/irssi/blob/master/docs/example-cross-android-aarch64.txt#L9

if you have any further questions that I didn't answer on IRC you can also ask

tristan957:
This would be pretty nice to have for Postgres, which has a ton of Perl.

PR #7982, opened 2020-11-14
Unittest fixups

dcbaker:
Here's a few fixups for the unit tests. Mostly these make things more resilient, but there's also a few simplifications in here too.


dcbaker:
Anyone object to merging this?

PR #5499, opened 2019-06-17
New builtin-option to support partial linking for shared libraries

sompen:
Added a new builtin-option called 'sharedlib-linkmodel' for being able to generate a partial linked object using a shared_library() target.
- when this option value is set to 'standard', the existing shared library linking method is used as is and is the default value.
- when this option value is set to 'partial', required compiler options for generating partial linked object will be selected and few of the 'standard' linking options are skipped.
Currently the support is added for armcc, armclang and gnu compilers, a warning will be generated if this option is set to 'partial' while using any other compilers.

PR #7983, opened 2020-11-14
Add prelinking support for static libraries.

jpakkane:
This is still fairly WIP and supports only GCC.

### DIFF
--- a/docs/markdown/snippets/prelink.md
+++ b/docs/markdown/snippets/prelink.md
@@ -1,0 +1,6 @@
+## Add support for prelinked static libraries
+
+The static library gains a new `prelink` keyword argument that can be
+used to prelink object files in that target. This is currently only
+supported for the GNU toolchain, patches to add it to other compilers
+are most welcome.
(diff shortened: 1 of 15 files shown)


docs/markdown/snippets/prelink.md
```diff
@@ -0,0 +1,5 @@
+## Add support for prelinked static libraries
+
+The static library gains a new `prelink` keyword argument that can be
+used to prelink object files in that target. This is mostly useful in
+embedded projects.
```
marc-h38, line 5: I think partial linking is less rare in embedded projects _only_ because static libraries are less rare in embedded projects. I would not mention "embedded" here. This new feature is only about static libraries anyway.

I found the two use cases listed in #5499 very interesting, I think it would be much better to mention these: 
- namespaces more or less equivalent to `dllexport` but for static libraries 
- partial LTO which is better than no LTO at all when building with different toolchains.

jpakkane:
This should be ready now. Please speak up if you have comments.

PR #7990, opened 2020-11-16
gnome: Drop use of volatile in GLib type functions

pwithnall:
See https://gitlab.gnome.org/GNOME/glib/-/issues/600

`volatile` was previously mistakenly used in GLib to indicate that a
variable was accessed atomically or otherwise multi-threaded. It’s not
meant for that, and up to date compilers (like gcc-11) will rightly warn
about it.

Drop the `volatile` qualifiers.

See also http://isvolatileusefulwiththreads.in/c/.

Signed-off-by: Philip Withnall <pwithnall@endlessos.org>

### DIFF
--- a/mesonbuild/modules/gnome.py
+++ b/mesonbuild/modules/gnome.py
@@ -1430,7 +1430,7 @@ class GnomeModule(ExtensionModule):
 GType
 %s@enum_name@_get_type (void)
 {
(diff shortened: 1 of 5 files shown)


xclaesse:
Makes sense, LGTM once CI is green.

PR #7991, opened 2020-11-17
Clarify tutorials

michaelbrockus:
Just review the tutorials real quick to see if where not spreading bad practices and found a few spots. The GTK demo that was there previously well be considered outdated when GTK4 is released so to be compatible with GTK3 and GTK4 I updated the snippet, next is the fact that the word setup was missing in places where that should have shown up in the command.


docs/markdown/Tutorial.md
```diff
@@ -29,10 +29,13 @@ example. First we create a file `main.c` which holds the source. It
...
 ```c
-#include<stdio.h>
+#include <stdio.h>
 
+//
+// main is where all program execution starts
+//
-  printf("Hello there.\n");
+  puts("Hello there.");
```
jpakkane, line 38: This may be confusing to newcomers. Most people know about `printf`(the canonical helloworld uses printf, after all) but knowledge about `puts` is less common. This is not even a performance improvement, as compilers will do this transformation automatically:

https://gcc.godbolt.org/z/Kxcqhb

jpakkane:
This MR also has a merge commit from master to this branch. Please remove it.

docs/markdown/Tutorial.md
michaelbrockus: I was actually more focused on making the snippet more secure but I well happy to change it back to `printf`.

PR #7994, opened 2020-11-17
doc: add much-needed accuracy to pip installation instructions [skip ci]

eli-schwartz:
- mention installing from local sources, not PyPI
- warn against --user installs, which too often screw up users that then cannot install projects because ~/.local won't be in sudo's PYTHONPATH
- advise installing with sudo -- current versions of pip assume --user for you rather than failing with permission errors, which is great unless, like meson, there are compelling reasons to need to install as root


QuLogic:
`meson install` will ask polkit to elevate permissions, so you don't need to use `sudo` anyway.

PR #7995, opened 2020-11-17
Subprojects recommendation 

michaelbrockus:
I find believe this recommendation would have a nice home in your style recommendation page.


jpakkane:
That is actually not true. `dependencies` is the recommend way to do things. If your wrap files have the necessary `[provide]` entries, everything will work automatically both when compiling your own and when using system dependencies.

For future reference, do not merge master in your pull requests, keep them up to date with rebase instead:

https://mesonbuild.com/Contributing.html#keeping-pull-requests-up-to-date

PR #7996, opened 2020-11-17
pkgconfig: Fix linking to a custom target

xclaesse:
When generating pkgconfig file for a library that links to an
uninstalled static library built by custom_target() Meson was crashing
when trying to access some attributes that does not exist on that class.

Also fix is_internal() implementation, it only really make sense on a
CustomTargetIndex or if CustomTarget has only a single output.

### DIFF
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -2588,13 +2588,12 @@ class CustomTarget(Target, CommandBase):
         return []
 
     def is_internal(self) -> bool:
-        if not self.should_install():
-            return True
-        for out in self.get_outputs():
-            # Can't check if this is a static library, so try to guess
-            if not out.endswith(('.a', '.lib')):
-                return False
-        return True
+        '''
+        Returns True iif this is a not installed static library.
+        '''
+        if len(self.outputs) != 1:
+            return False
+        return CustomTargetIndex(self, self.outputs[0]).is_internal()
 
     def extract_all_objects_recurse(self) -> T.List[T.Union[str, 'ExtractedObjects']]:
         return self.get_outputs()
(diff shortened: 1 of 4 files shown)


xclaesse:
This fix meson crash seen in GStreamer: https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/issues/133#note_696355.

CC @nirbheek

gdesmott:
Does not seem to work here: https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/issues/133#note_706950

jpakkane:
The test failure is hard to pick up from the test output but fairly straightforward when run locally:

```
Failed test during install: 'test cases/common/45 pkgconfig-gen'
Reason: 
Extra file usr/lib/pkgconfig/simple6.pc found.
```

xclaesse:
Rebased and hopefully fixed the CI. Targetting 0.60.3 because it's a bugfix.

mesonbuild/build.py
```diff
@@ -2764,7 +2763,11 @@ def should_install(self) -> bool:
         return self.target.should_install()
 
     def is_internal(self) -> bool:
-        return self.target.is_internal()
+        '''
+        Returns True iif this is a not installed static library
+        '''
+        suf = os.path.splitext(self.output)[-1]
+        return (suf == '.a' or suf == '.lib') and not self.should_install()
```
dcbaker, line 2770:
```suggestion
        return suf in {'.a' or '.lib'} and not self.should_install()
```

xclaesse:
The Windows failure seems unrelated, happens on master too.

mesonbuild/build.py
xclaesse: done.

nirbheek:
LGTM, rebased to see if fixes from master fix the CI.

PR #7998, opened 2020-11-17
Update Tutorial.md

michaelbrockus:


jpakkane:
Thanks. Just note in the future that doc-only changes should have `[skip ci]` at the end of their commit messages, so CI is not run.

Issue #8002, opened 2020-11-19
'once' is an invalid keyword argument for print()


Neumann-A:
```
Traceback (most recent call last):
  File "D:\downloads\tools\meson\meson-91876b40316962620c1705ae14075ab46f8dd644\mesonbuild\mesonmain.py", line 132, in run
    return options.run_func(options)
  File "D:\downloads\tools\meson\meson-91876b40316962620c1705ae14075ab46f8dd644\mesonbuild\msetup.py", line 253, in run
    app.generate()
  File "D:\downloads\tools\meson\meson-91876b40316962620c1705ae14075ab46f8dd644\mesonbuild\msetup.py", line 161, in generate
    self._generate(env)
  File "D:\downloads\tools\meson\meson-91876b40316962620c1705ae14075ab46f8dd644\mesonbuild\msetup.py", line 200, in _generate
    intr.run()
  File "D:\downloads\tools\meson\meson-91876b40316962620c1705ae14075ab46f8dd644\mesonbuild\interpreter.py", line 4726, in run
    super().run()
  File "D:\downloads\tools\meson\meson-91876b40316962620c1705ae14075ab46f8dd644\mesonbuild\interpreterbase.py", line 519, in run
    self.evaluate_codeblock(self.ast, start=1)
  File "D:\downloads\tools\meson\meson-91876b40316962620c1705ae14075ab46f8dd644\mesonbuild\interpreterbase.py", line 544, in evaluate_codeblock
    raise e
  File "D:\downloads\tools\meson\meson-91876b40316962620c1705ae14075ab46f8dd644\mesonbuild\interpreterbase.py", line 537, in evaluate_codeblock
    self.evaluate_statement(cur)
  File "D:\downloads\tools\meson\meson-91876b40316962620c1705ae14075ab46f8dd644\mesonbuild\interpreterbase.py", line 552, in evaluate_statement
    self.assignment(cur)
  File "D:\downloads\tools\meson\meson-91876b40316962620c1705ae14075ab46f8dd644\mesonbuild\interpreterbase.py", line 1219, in assignment
    value = self.evaluate_statement(node.value)
  File "D:\downloads\tools\meson\meson-91876b40316962620c1705ae14075ab46f8dd644\mesonbuild\interpreterbase.py", line 554, in evaluate_statement
    return self.method_call(cur)
[...]

PR #8003, opened 2020-11-19
mlog: make mlog helper take once keyword argument

dcbaker:
We really want to have this in the log method as well.

Fixes: #8002

### DIFF
--- a/mesonbuild/mlog.py
+++ b/mesonbuild/mlog.py
@@ -218,7 +218,15 @@ def _debug_log_cmd(cmd: str, args: T.List[str]) -> None:
 def cmd_ci_include(file: str) -> None:
     _debug_log_cmd('ci_include', [file])
 
+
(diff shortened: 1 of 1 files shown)


dcbaker:
the LGTM warning is stupid. It's complaining that I'm returning a function that returns None. but it looks like this:
```python
def _fun1() -> None: ...
def _fun2() -> None: ...

def fun(thing: bool = true) -> None:
   if thing:
    return _fun1()
  return _fun2()
```

*Not* returning here is odd.

bonzini:
Why not use `else` as in the code you are moving? Or `return`ing on a separate line would work too.

dcbaker:
it's generally considered a code smell in python to use an else when you return from a single if, both because it creates unnecessary indentation, and also because there is a small cost to the else block.

PR #8011, opened 2020-11-20
Python 3.6 cleanups

dcbaker:
Here's a few cleanups now that we require 3.6.

Most of these are just removing < 3.6 checks, or updating minimum checks to be accurate. The big one is the removal of the `_pathlib` module, which fixed some behavior in 3.5 to act like 3.6+. Since we have 3.6 we don't need this anymore.


mensinda (approved):
LGTM, once the bot issues are fixed

PR #8012, opened 2020-11-21
Dependency recommendation

michaelbrockus:
I'm wondering if there is a case when someone would use the `subproject` method. Other than that I believe this should be close to what is recommended for the Meson build system.

### DIFF
--- a/docs/markdown/Style-guide.md
+++ b/docs/markdown/Style-guide.md
@@ -26,7 +26,7 @@ The `dependency` function is the recommended way to handle dependencies. If your
 have the necessary `[provide]` entries, everything will work automatically both when
 compiling your own and when using system dependencies.
 
-You should only need `subprojects` when you need to extract non dependencies/programs.
+You should only need `subproject` when you need to extract non dependencies/programs.
 
 ## Naming options
 
--- a/docs/markdown/Style-guide.md
+++ b/docs/markdown/Style-guide.md
@@ -22,7 +22,7 @@ Let say you would like to refer to your executable so something like `my_exe`.
 
 ## Dependency usage
 
-The `dependency` function is is the recommend as a way to handle dependenceis. If your wrap files
+The `dependency` function is is the recommend as a way to handle dependencies. If your wrap files
 have the necessary `[provide]` entries, everything will work automatically both when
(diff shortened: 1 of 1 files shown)


dcbaker:
You should only need subprojects when you need to extract non dependencies/programs.

michaelbrockus:
Ok. I believe this small note should do the trick.

dcbaker (approved):

docs/markdown/Style-guide.md
```diff
@@ -20,6 +20,14 @@ in Meson build scripts as identifiers for variable.
 
 Let say you would like to refer to your executable so something like `my_exe`.
 
+## Dependency usage
+
+The dependency method is the recommend as a way to handle dependenceis. If your wrap files
```
eli-schwartz, line 25: "is the recommend as a way"

Could use some rewording...

Maybe "is the recommended way"?

"dependenceis" -- typo

eli-schwartz (changes requested):

docs/markdown/Style-guide.md
jpakkane: This should say "dependency function" rather than method. And the word dependency should be in backticks so it gets formatter properly. So something like:

    The `dependency` function is

There is also a typo later in the sentence: "dependenceis" instead of "dependencies".

michaelbrockus: Already issues fixed.

eli-schwartz: No it is not. My first comment is still applicable

michaelbrockus: Which part?

eli-schwartz: > "is the recommend as a way"
> 
> Could use some rewording...
> 
> Maybe "is the recommended way"?

michaelbrockus: Oh. Well go and fix that sorry got a lot of stuff to do.

michaelbrockus: Already. Hopefully I covered all of the issues.

eli-schwartz: Thanks.

eli-schwartz (approved):

PR #8013, opened 2020-11-21
C++ module support

jpakkane:
This MR adds support for compiling C++ modules, which are similar to Fortran modules requiring one to scan the contents of files. This is done with the new Ninja dynamic dependency support that is only available in the latest releases. There are several limitations.

- only works with Visual Studio 2019 developer preview
- only supports basic "ixx" modules, not module partitions et al

There are also open questions such as where should the generated module files go. Should they go in target private dirs or to a single common dir? The latter is easier and works if we can rely on the fact that no module of a given name is generated in two different places. Currently all those are dumped in build dir root, because MS has not yet documented the compiler flags that specify where those should go (if said compiler flags exist at all).


jpakkane:
This should pass all the tests now. The actual functionality is fairly small, but the machinery is important and can be used to make Fortran compilation finally reliable.

PR #8014, opened 2020-11-22
Bump Ninja requirement to 1.8.

jpakkane:
Buster and Bionic both have 1.8. Not sure about the various RHELs, but since 1.8 should be newer than Python 3.5 which we no longer support.


jpakkane:
FTR Python 3.6 was released 2016-12-23 and Ninja 1.8 was released 2017-9-4. RHEL 7 was released before both and RHEL 8 was released after both. Same for SLES 12 and 15.

eli-schwartz:
Which new features of ninja do you want to start using?

jpakkane:
Nothing in particular, this is just bumping the base requirement. No point in keeping support for old versions that are not reasonably in use anywhere.

bonzini:
RHEL has 1.8.2. Probably the minimum requirement should be 1.8.2. since 1.8.0 and 1.8.1 seem to be release candidates.

scivision:
For separate reasons I have to always use Ninja >= 1.10 and this hasn't been a problem for me. I.e. requiring 1.8 is also not a problem for me.

obilaniu:
A quick reminder that the top-level README.md should also be updated, since it currently speaks of Ninja 1.7 and Python 3.5 (another stale mention).

PR #8025, opened 2020-11-25
mypy: use declarative mypy.ini for file paths

scivision:
MyPy can use the mypy.ini "files" parameter to set paths (files or directories) to process. This can be seen as "best practices" for using declarative rather than imperative parameters for the MyPy CI test.

### DIFF
--- a/run_mypy.py
+++ b/run_mypy.py
@@ -6,50 +6,18 @@ import argparse
 from pathlib import Path
 import typing as T
 
-modules = [
-    # fully typed submodules
-    'mesonbuild/ast',
-    'mesonbuild/cmake',
-    'mesonbuild/compilers',
-    'mesonbuild/scripts',
-    'mesonbuild/wrap',
-
-    # specific files
-    'mesonbuild/arglist.py',
-    # 'mesonbuild/coredata.py',
-    'mesonbuild/dependencies/boost.py',
-    'mesonbuild/dependencies/hdf5.py',
-    'mesonbuild/dependencies/mpi.py',
-    'mesonbuild/envconfig.py',
-    'mesonbuild/interpreterbase.py',
-    'mesonbuild/linkers.py',
-    'mesonbuild/mcompile.py',
-    'mesonbuild/mesonlib.py',
-    'mesonbuild/minit.py',
-    'mesonbuild/mintro.py',
-    'mesonbuild/mlog.py',
-    'mesonbuild/modules/fs.py',
-    'mesonbuild/mparser.py',
-    'mesonbuild/msetup.py',
-    'mesonbuild/mtest.py',
-    'mesonbuild/optinterpreter.py',
-
-    'run_mypy.py',
-    'tools'
-]
 
 def check_mypy() -> None:
     try:
         import mypy
     except ImportError:
-        print('Failed import mypy')
-        sys.exit(1)
+        raise SystemExit('Failed import mypy')
 
 def main() -> int:
     check_mypy()
 
     root = Path(__file__).absolute().parent
-    args = []  # type: T.List[str]
+    args: T.List[str] = []
 
     parser = argparse.ArgumentParser(description='Process some integers.')
     parser.add_argument('-p', '--pretty', action='store_true', help='pretty print mypy errors')
@@ -64,10 +32,10 @@ def main() -> int:
 
     print('Running mypy (this can take some time) ...')
     p = subprocess.run(
-        [sys.executable, '-m', 'mypy'] + args + modules,
+        [sys.executable, '-m', 'mypy'] + args,
         cwd=root,
     )
     return p.returncode
 
 if __name__ == '__main__':
-    sys.exit(main())
+    raise SystemExit(main())
--- a/.github/workflows/lint_mypy.yml
+++ b/.github/workflows/lint_mypy.yml
@@ -31,4 +31,4 @@ jobs:
       with:
         python-version: '3.x'
     - run: python -m pip install mypy
-    - run: python run_mypy.py
+    - run: python -m mypy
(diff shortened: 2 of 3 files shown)


run_mypy.py
```diff
@@ -64,10 +32,10 @@ def main() -> int:
...
     p = subprocess.run(
-        [sys.executable, '-m', 'mypy'] + args + modules,
+        [sys.executable, '-m', 'mypy'] + args,
         cwd=root,
     )
     return p.returncode
 
 if __name__ == '__main__':
-    sys.exit(main())
+    raise SystemExit(main())
```
dcbaker, line 41: if we're just going to put the arguments in the .mypy.ini anyway, and not use run_mypy in CI, do we even need the run_mypy script? @mensinda, opinions?

dcbaker (approved):
other than whether we need run_mypy.py at all, this looks good to me.

mensinda:
I am not a big fan of the current solutions, since all files are on the same line (which is also one of the main reasons I created the `run_mypy.py` script). So, unless this can be changed and the end result is as easy to maintain as the current list in `run_mypy` I am against this change.

Is there some actual tangible benefit from moving everything into `mypy.ini` besides some abstract "best practices"? We could theoretically remove `run_mypy.py` entirely since the base functionality is then provided with the config file. However, the script does provide some additional features that might not be provided with just mypy + `.ini`:

What would happen if you were to accidentally run `python -m mypy mesonbuild/` (the current script provides an error)? Can you run mypy from a different directory other than the project root of meson?

scivision:
Yes I was just trying to make things neater as I was the one who made the original long, messy mypy command line that was in the CI originally.

* yes, the current mypy.ini syntax requires all paths as one line without `\n`
* MyPy, Flake8, etc. use the configuration file in the directory from which they are executed. So if I run flake8 or mypy from a different project directory it uses that other project's .mypy.ini or .flake8 etc. 

The idea with this .mypy.ini `files` parameter is that users and CI can just type `mypy` at the top level directory and it just works without needing helper scripts. The same idea used by flake8, etc.  Other such tools act the same way when called from other directories.

run_mypy.py
scivision: The run_mypy.py benefit (ability to run mypy from any directory) can be accomplished by

```sh
mypy --config-file /path/to/meson/.mypy.ini
```

mensinda:
> yes, the current mypy.ini syntax requires all paths as one line without \n

This is a dealbreaker IMHO. At least until the majority of `mesonbuild` is fully annotated.

> The idea with this .mypy.ini files parameter is that users and CI can just type mypy at the top level directory and it just works without needing helper scripts.

I would argue that there isn't really much of a difference from typing `mypy` to typing `./run_mypy.py`. With the added benefit that `/path/to/run_mypy.py` is shorter (and easier to type) than `mypy --config-file /path/to/meson/.mypy.ini`.

I guess the wrapper script is a bit counter-intuitive for new users already familiar with mypy, but the wrapper script is literally called `run_mypy`...

lazka:
Another thing to consider is that IDEs like VSCode call mypy/flake8 directly for providing their inline linting reports. So this brings the test results in line with what the editor shows.

Re the file list approach: Would moving things to a blocklist help? I use something like this in another project:

```
[mypy]
python_version = 3.6

[mypy-setup]
ignore_errors = True

[mypy-tests.*]
ignore_errors = True
```

mensinda:
> Re the file list approach: Would moving things to a blocklist help?

I have not counted the files but we should be around roughly 50% (+- 10%) type annotations currently. Moving to a blacklist approach would be more maintainable than one line of `files`, but also much more clunky.

lazka:
yeah, it would look something like this:

```
[mypy-skip_ci,ghwt,run_project_tests,run_unittests,run_tests,meson,__main__,run_meson_command_tests,run_cross_test]
ignore_errors=True

[mypy-*.dependencies.base,*.dependencies.coarrays,*.dependencies.cuda,*.dependencies.dev,*.dependencies.misc,*.dependencies.platform,*.dependencies.scalapack,*.dependencies.ui]
ignore_errors=True

[mypy-*.build,*.coredata,*.depfile,*.environment,*.interpreter,*.mconf,*.mdist,*.mesonmain,*.minstall,*.msubprojects,*.munstable_coredata,*.optinterpreter,*.rewriter]
ignore_errors=True

[mypy-*.templates.*]
ignore_errors=True

[mypy-*.modules.*]
ignore_errors=True

[mypy-*.modules.fs]
ignore_errors=False

[mypy-*.backend.*]
ignore_errors=True
```

mensinda:
I guess linting is a valid point, but it should already work (are you encountering any issues currently?) for all the currently fully annotated files since all the settings are already in the `.mypy.ini`. The only problem should be the files that are currently ignored.

I would propose that we keep the current setup since it is easy to maintain (more than one line and does not mix white with blacklists) until the mypy coverage has increased considerably.

lazka:
> I guess linting is a valid point, but it should already work (are you encountering any issues currently?) for all the currently fully annotated files since all the settings are already in the `.mypy.ini`. The only problem should be the files that are currently ignored.

It's just red squiggly lines everywhere in some files (a mix of mypy and flake8 errors), that's all.

run_mypy.py
eli-schwartz: > The run_mypy.py benefit (ability to run mypy from any directory)

Why is this even considered "a benefit"?

mensinda: Well, at least for me, when I am working in the `work area` and I can just run `../run_mypy.py`

jpakkane:
Are there still issues afoot or is this ok to merge?

mensinda (changes requested):
I am personally against this, because this change will put all files/modules into a single line which is hard/annoying to maintain.

So, I am against merging this, unless there are some changes to the `.mypy.ini` syntax to allow declaring modules similar to the `run_mypy.py` script.

scivision:
Well this is a bit of bikeshedding / purism PR anyway. It could be deferred until type annotation is more complete to reconcile opinions.

dcbaker:
I agree that we should wait. If we change this then vscode wont run mypy against packages that aren't in the mypy list, which means if a module is partially annotated I wont get notifications that something is wrong in that module. That seems like a step back.

PR #8029, opened 2020-11-25
mtest: asynchronous TAP parsing, improved progress report

bonzini:
The next part after #8001. Adds:
* to the progress report a clock and, for TAP tests, the number of passed+run subtests.
* to the final test report for TAP tests, the number of subtests that have passed.

[Demo available here](https://asciinema.org/a/LIimxKSIyh8TNwq9U076vq0WU).

### DIFF
--- a/mesonbuild/mtest.py
+++ b/mesonbuild/mtest.py
@@ -1051,6 +1051,10 @@ class SingleTestRunner:
                 return self.test.exe_runner.get_command() + self.test.fname
         return self.test.fname
 
+    @property
+    def visible_name(self) -> str:
+        return self.runobj.name
+
     async def run(self) -> TestRun:
         cmd = self._get_cmd()
         self.runobj.start()
@@ -1254,7 +1258,8 @@ class TestHarness:
             options.wrapper = current.exe_wrapper
         return current.env.get_env(os.environ.copy())
 
-    def get_test_runner(self, test: TestSerialisation, name: str) -> SingleTestRunner:
+    def get_test_runner(self, test: TestSerialisation) -> SingleTestRunner:
+        name = self.get_pretty_suite(test)
         options = deepcopy(self.options)
         if not options.setup:
             options.setup = self.build_data.test_setup_default_name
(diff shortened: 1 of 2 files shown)


jpakkane:
Sorry for the delay in reviewing, but there is a lot of stuff in this one. 👍

Overall it looks nice, but there are minor issues in the formatting that jump at me. Specifically the alignments:

```
85/86 qemu:unit / test-char                           OK             5.47s (38 tests passed)                           
86/86 qemu:unit / ptimer-test                         OK             12.90s (528 tests passed)   
```

The times and numbers of tests run do not line up, which makes the output unpleasant to read. Come to think of it, the parentheses seem to be a bit superfluous as well. Maybe something like this instead:

```
85/86 qemu:unit / test-char                           OK              5.47s   38 subtests passed
86/86 qemu:unit / ptimer-test                         OK             12.90s  528 subtests passed
```

That is, pad the time with spaces to tens of seconds and the test count to hundreds. I'm guessing tests that take more than 99 seconds are rare as are tests that have more than 999 subtests.

bonzini:
The parentheses (and left justification of the number of tests) are there because they align with the `(killed by signal)` message (also demonstrated in the asciinema recording).

Padding the time is a good idea, but the formatting is the same as in current Meson master. Should I open a separate pull request for it or modify this one? I guess it depends also on the fate of #7860 since I'm waiting for that one before rebasing again.

jpakkane:
> Should I open a separate pull request for it or modify this one? 

You can put it in this one. No need to complicate things needlessly.

bonzini:
For now I have only rebased on top of the Rust module, I'll add the alignment tweak next.

bonzini:
The LGTM failure is for `NamedTuple`, which is from the `typing` module. I can use it as `T.NamedTuple` but it's uglier than just using the class name.

dcbaker:
Please don't double import it, `T.` is not a lot of typing anyway.

bonzini:
Ok, it was more a matter of looking untidy than extra typing. I'll fix it.

dcbaker:
I don't think using namespaces is untidy, I tend to view polluting the local exported namespace as worse. Probably because I'm primarily a python programmer and think about things like that, lol

eli-schwartz:
> Ok, it was more a matter of looking untidy than extra typing. I'll fix it.

What's really untidy is going against the project coding style for a couple of lines while being surrounded by lines of a different, conflicting coding style in the same file and class. :p

jpakkane:
LGTM, are there still things that people want to see fixed?

Issue #8030, opened 2020-11-25
build-RPATH order wrong if dependencies themself have RPATHs


Finii:
**Describe the bug**
The project is library `foo` which usually resides in a special directory (not in the usual search path). We want to build and test `foo`.

This `foo` depends on another library `bar` that also resides in the same special directory. `bar` is found through its pkgconfig file; and that files specifies a loader search path via `-Wl,-rpath`.

When `foo` is built and tested the test executable dynamically links potentially a wrong library (not the currently build one but the older system installed one).

The problem is how Meson assembles the linker rpath arguments. It first adds all linking arguments found from the dependencies' pkgconfig files. And afterwards it adds the path(s) to the current build artifacts:

```
c++  -o tests/mytest tests/mytest.p/mytest.cc.o
    -Wl,--as-needed -Wl,--no-undefined
    -Wl,--start-group src/libfoo.so.0.8.15 /tmp/special/dir/libbar.so -Wl,--end-group
    -Wl,-rpath,/tmp/special/dir
    '-Wl,-rpath,$ORIGIN/../src'
    -Wl,-rpath-link,/home/fini/git/foo/build/src
```

Note the rpath with `$ORIGIN` comes last and thus is at the end of the search path.
That rpath option need to come first instead.

**To Reproduce**
A simple repository can be found here: https://github.com/Finii/foo
It has instructions how to reproduce.

The current behavior leaves us with this build-rpath:
```
$ ldd build/tests/mytest
        linux-vdso.so.1 (0x00007fffbe78e000)
        libfoo.so.0.8.15 => /tmp/special/dir/libfoo.so.0.8.15 (0x00007f7d1c312000)
        libbar.so.47.11 => /tmp/special/dir/libbar.so.47.11 (0x00007f7d1c30d000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f7d1c101000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f7d1c31e000)

$ readelf -d build/tests/mytest | grep path
 0x000000000000001d (RUNPATH)            Library runpath: [/tmp/special/dir:$ORIGIN/../src]
```

**Expected behavior**
The artifacts of the current build should come always first in the search path, like so:

```
$ ldd build/tests/mytest
        linux-vdso.so.1 (0x00007fffe0781000)
[...]

PR #8031, opened 2020-11-25
ninjabackend: Correct RPATH order

Finii:
[why]
If we build and test a library we need to make sure that we find the
currently build library object first, before an older system installed
one.
This can be broken if the library in question is installed in a custom
path, and another library we depend on also is installed there.

[how]
Just move the rpath to the current build artifacts to the front.

Solves #8030.

Signed-off-by: Fini Jastrow <ulf.fini.jastrow@desy.de>


jpakkane:
This needs a test.

Finii:
> This needs a test.

Is that a request in my direction? :->
The "setup" is relatively complicated, which is why I created a whole [how-to-reproduce repository](https://github.com/Finii/foo).
I can try of course, but I would rather leave this to ppl more experienced with the Meson test setup.

Please let me know how to proceed. Thank you, Fini

mhier:
Any news on this? We really need this...

jpakkane:
As said this needs a test. It should be implemented in `run_unittests.py´ under `LinuxlikeTests` and should do roughly this:

- have a custom .pc file for a dependency that just adds an rpath linker argument to a known value like ´/usr/lib/dummy`
- create a new project that imports said dependency, `PKG_CONFIG_PATH` needs to be set to the directory where your dummy .pc file is
- in the project build one shared lib and one executable that links to it
- it should also use the dummy dependency so its rpath flag is used
- build the project
- extract the rpath from the exe file, verify that both of the rpaths are there and that the build dir one (it always starts with `$ORIGIN`) comes before the system one

Overall the Meson file should look something like this:

```meson
project('rpathorder', 'c')

dummy_dep = dependency('rpathdummy')
l = shared_library('helper', 'helper.c')
executable('rpathtest', 'rpathtest.c', link_with: l, dependencies: dummy_dep)
```

Finii:
* rebase on master
* added test

I added the test not in `linuxlike` but rather in `unit`, because `10 build_rpath` (which approx checks the same thing, just without pkgconfig) also resides there.

In fact I copied that test to make this new test. First I incorporated the pkgconfig test into `10 build_rpath` but that was not so nice when `@skipIfNoPkgconfig` needs to be supported.

Thank you, Fini

Finii:
Force push: Correct whitespace error

run_unittests.py
```diff
@@ -6733,6 +6733,31 @@ def test_build_rpath(self):
...
+        # C program RPATH
+        build_rpath = get_rpath(os.path.join(self.builddir, 'prog'))
+        self.assertEqual(build_rpath, '$ORIGIN/sub:/foo/bar:/foo/dummy')
+        self.install()
+        install_rpath = get_rpath(os.path.join(self.installdir, 'usr/bin/prog'))
+        self.assertEqual(install_rpath, '/baz:/foo/dummy')
+        # C++ program RPATH
+        build_rpath = get_rpath(os.path.join(self.builddir, 'progcxx'))
+        self.assertEqual(build_rpath, '$ORIGIN/sub:/foo/bar:/foo/dummy')
+        self.install()
```
jpakkane, line 6757: You are running install twice. Once is sufficient. So something like:

- build
- check C build dir rpath
- check C++ build dir rpath
- install
- check C install rpath
- check C++ install rpath

Finii:
Removed superfluous second install.

Created PR #8222 where it is also removed from the testcase that has been used as copy source for this.

PR #8032, opened 2020-11-25
Cleanup specialized packaging (msi, zipapp)

eli-schwartz:
- move msi scripts to generic "packaging directory"
- create script to produce zipapp without immense clutter (running `python -m zipapp` on the project root will zip up `.git/`)
- remove `__main__.py` as it's both confusing in its current location and kind of useless. There are better ways to produce zipapps.


eli-schwartz:
This PR results in one less file in the project root. :D

dcbaker (approved):

eli-schwartz:
I didn't realize there was a test for the previous method. :p Updated the script a bit, and wired it up to the test.

PR #8035, opened 2020-11-26
CI: Move all remaining jobs from travis-ci to GHA

lazka:
Some changes:

* Set HOME to /root, since github mounts its own HOME and 'wine'
  (because of permissions) and 'dub' (can't find packages) don't
  like that.
* Remove the seccomp option, doesn't seem to be needed.


mensinda:
I am not entirely sure if we should get rid of travis entirely since GHA also have a limit of 20 parallel jobs: https://docs.github.com/en/free-pro-team@latest/actions/reference/usage-limits-billing-and-administration#usage-limits

Maybe just reduce the number below 5 on Travis, so that we still have some more breathing room on GHA?

lazka:
The last commit on master was pushed 3 hours ago and travis-ci is still not finished. travis-ci currently has currently 5700 jobs queued, about the same as on every other weekday at this time for the last month. So imo, GHA will always be faster even if there is some queuing involved.

lazka:
@mensinda since you are here already, while working on this I noticed that env_vars.sh in https://github.com/mesonbuild/meson/runs/1460462406 is failing but errors are ignored there.

mensinda:
OK, I wasn't really aware of the current Travis situation. In that case a full move to GHA would indeed be better.

Looks like there were some changes to DMD, and the installation path has changed in the image...

lazka:
Another thing to consider is managing the docker images here https://github.com/orgs/mesonbuild/packages instead of on docker hub: https://docs.github.com/en/free-pro-team@latest/actions/guides/publishing-docker-images#publishing-images-to-github-packages

Maybe it makes things faster (though I assume GH does some caching for docker hub, so maybe there is no difference)

mensinda:
Pulling the images is around 1:30 min. I haven't benchmarked it but I doubt that this time can be decreased significantly when switching to the GH thing.

That being said, I am not against moving more stuff from other services to GH (reduces potential points of failure), but have currently not really the time to test this.

lazka:
yeah, did a quick test here, and there is no difference between dockerhub and github: https://github.com/lazka/docker-perf-test/runs/1467804339?check_suite_focus=true never mind then :)

dcbaker:
travis has been waiting hours to run the same jobs, and it only seem to run one job at a time, which means that the parallelism is awful.

dcbaker:
The only concern I have, is with this change we aren't testing on xenial anymore, which means we're not testing the oldest supported versions of things, otherwise I think this is generally a positive change.

eli-schwartz:
I this just about the python version or is there more to it?

dcbaker:
There's been several people who are super concerned that ancient versions of Ubuntu keep working, I would be fine with just making sure at least one of our tests uses python 3.6, but others may have other opinions.

lazka:
@dcbaker there are no xenial tests on travis, this PR doesn't change that.

dcbaker:
So there isn't, the travis interface just lies and say Xenial. This is fine with me then. I'm going to go ahead and merge so we can stop waiting on travis.

lazka:
Thanks!

It's all docker on Xenial, so not a complete lie :)

PR #8044, opened 2020-11-28
Update Vala Preprocessor section in documentation [skip ci] - Closes #8040

### LINKED ISSUE #8040: Vala preprocessor documentation
**Describe the bug**

Following the instructions at https://mesonbuild.com/Vala.html#working-with-the-vala-preprocessor does not work as described.

**To Reproduce**

Adding add_project_arguments('-DSOME_FLAG', language: 'c') did not set that symbol for vala code, it seems vala requires a space '-D SOME_FLAG' however if we use add_project_arguments('-D SOME_FLAG', language: 'c') that results in an error probably due to it being quoted.

add_project_arguments('-D', 'SOME_FLAG', language: 'vala') seems to work as expected

**Question**

Is the version that works reliable? By this I mean is it guaranteed that those arguments will be added in that order everytime?

Should I just update the docs and submit a pull request or this actually a bug?

**system parameters**

* Fedora 33
* Python 3.9.0
* Meson 0.55.3
* Ninja 1.10.1

JerryCasiano:


astavale:
Thanks, looks good. Can you:

1. Add [skip ci] to the end of the commit title - see [Contributing - documentation](https://mesonbuild.com/Contributing.html#documentation)

2. Add something about the C pre-processor, e.g.:

> If you need to pass an argument to the C pre-processor then specify the language as `c`. For example to set `FUSE_USE_VERSION` to `26` use:
> 
> ```meson
> add_project_arguments('-DFUSE_USE_VERSION=26', language: 'c')
> ```
>

JerryCasiano:
I think that should do it.

Let me know if any other changes are needed.

Thanks.

PR #8048, opened 2020-11-28
Do not test aarch64 as the assembly has not been written.

jpakkane:
Originally from Debian:

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=975411

### DIFF
--- a/test cases/common/119 llvm ir and assembly/meson.build
+++ b/test cases/common/119 llvm ir and assembly/meson.build
@@ -1,7 +1,7 @@
 project('llvm-ir', 'c', 'cpp')
 
 cpu = host_machine.cpu_family()
-supported_cpus = ['arm', 'aarch64', 'x86', 'x86_64']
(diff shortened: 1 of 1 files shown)


mimi89999:
Isn't it easier to just write that test?

jpakkane:
> Isn't it easier to just write that test?

Patches welcome.

mimi89999:
I think it will just be:
```
.text
.globl SYMBOL_NAME(square_unsigned)
# ifdef __linux__
.type square_unsigned, %function
#endif

SYMBOL_NAME(square_unsigned):
    mul x1, x0, x0
    mov x0, x1
    ret
```
but I need to check that on an aarch64 device.

mimi89999:
@jpakkane It's exactly that. Just tested it. https://github.com/mesonbuild/meson/pull/8051

jpakkane:
Great, thanks. Closing this one.

PR #8052, opened 2020-11-29
Use release for dependencies if MSVC runtime is MD or MT

vvainola:
Specifying b_vscrt=MD should make dependencies use release build
even if buildtype is debug. The debug runtime in MSVC is much slower
and it is useful to be able to run own code in debug but the dependencies
are still in release. 

Before, using buildtype=debug, b_vscrt=MD with a cmake dependency caused
the debug version of dependency to be linked which will use MDd and which
could cause a runtime error.

### DIFF
--- a/mesonbuild/cmake/interpreter.py
+++ b/mesonbuild/cmake/interpreter.py
@@ -383,7 +383,9 @@ class ConverterTarget:
                     cfgs += [x for x in tgt.properties['CONFIGURATIONS'] if x]
                     cfg = cfgs[0]
 
-                is_debug = self.env.coredata.get_option(OptionKey('debug'));
+                is_debug = self.env.coredata.get_option(OptionKey('debug'))
+                if OptionKey('b_vscrt') in self.env.coredata.options:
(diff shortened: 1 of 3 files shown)


jpakkane:
:crossed_swords: conflicts.

jpakkane:
This is mixing two different things due to the unfortunate naming convention. The Meson option `debug` means whether to build with debug info enabled or not. In Visual Studio debug means using the "debug runtime" that has extra validity checks.

vvainola:
Exactly and now meson is picking wrong libs because MD/MT is always used in release builds but can be used also while debugging. Currently this basically means that using MD/MT with debug is not possible because then you get mixed run times.

vvainola:
The logic for MSVC should be based on VSCRT and for unix on the debug info.

vvainola:
I added minimalistic example here https://github.com/vvainola/Mixed-meson-VSCRT. Qt dependency currently tries to look for Qt5Cored if buildtype is debug even though b_vscrt=md. Using Qt library then causes crash because own code uses MD whereas Qt uses MDD

vvainola:
This seems to be fixed in master like it is in this PR

PR #8072, opened 2020-12-07
msubprojects: Ignore worktrees when checking out

nirbheek:
Otherwise it's not possible to share git subprojects via worktrees when creating a worktree of a git repository that uses meson subprojects.

The downside is that the user needs to be careful while adding commits to each tree's index.

### DIFF
--- a/mesonbuild/msubprojects.py
+++ b/mesonbuild/msubprojects.py
@@ -85,7 +85,7 @@ def git_reset(repo_dir, revision):
     return True
 
 def git_checkout(repo_dir, revision, create=False):
-    cmd = ['checkout', revision, '--']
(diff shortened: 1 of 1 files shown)


xclaesse:
I'm not using worktree that much (I should), but it makes sense, I trust you. LGTM.

gregsskyles:
Re: "The downside is that the user needs to be careful while adding commits to each tree's index."

Does this merit some cautionary text somewhere in the documentation?

nirbheek:
> Does this merit some cautionary text somewhere in the documentation?

No, this is a case that will only occur if the user manually chooses to use worktrees with subprojects, so they should be aware of the downsides of worktrees. It will never happen if you don't run git commands yourself on your subprojects to use worktrees.

Issue #8074, opened 2020-12-07
generated pkg-config files (.pc) may be broken due to variable reordering


michaelolbrich:
**Describe the bug**
Since meson 0.56.0 the generated .pc files may be broken: If a variable references another variable then that variable must be defined first, otherwise pkg-config will fail. I've seen this with glib. The generated file contains:
```
gobject_query=${bindir}/gobject-query
glib_mkenums=${bindir}/glib-mkenums
bindir=${prefix}/bin
glib_genmarshal=${bindir}/glib-genmarshal
```
Or another arbitrary order of these variable. Unless `bindir` is the first variable, pkg-config fails with:
`Variable 'bindir' not defined in '.../lib/pkgconfig/glib-2.0.pc'`

**To Reproduce**
Just build glib or split out the pkg-config stuff from here:
https://github.com/GNOME/glib/blob/master/glib/meson.build#L391

**Expected behavior**
The variable order must be preserved.

**system parameters**
I've first seen this on Ubuntu 16.04.7 LTS
I think all Python 3.5 Versions are affected.
I think Python 3.7 is not affected, but I may just be lucky in this case.
This happens since bcf369ea3c6ac9d48759a9a11304a853dfdab5ff and seems to be a problem with ordering and dictionaries. With 3.7 the order seems stable but I don't know enough about the internals, so I'm not sure if it is always correct.

xclaesse:
I think this is indeed a problem or dict ordering with python3.5. @michaelolbrich can you please verify https://github.com/mesonbuild/meson/pull/8076 fix the issue?

PR #8076, opened 2020-12-07
pkgconfig: Respect variable ordering when passed as list

xclaesse:
This fix a regression introduced in Meson 0.56.0 when using python 3.5.
Also mention in documentation that using a meson dict does not guarantee
ordering.

Fixes: #8074.


michaelolbrich:
It works for the existing string array syntax. But with the new dictionary syntax, the dict is created elsewhere, so this is still broken.

xclaesse:
@michaelbrockus Yes that's why I updated the doc, with the dict syntax we cannot guarantee ordering.

michaelolbrich:
Oh, I missed that part. So, this looks good as is.

PR #8083, opened 2020-12-08
Add Qt6 module

### LINKED ISSUE #8056: Support of Qt6 import
Hi,
Qt6 RC is out, it would be nice to add the support of this version to meson.

z3ntu:
This is basically a copy of the Qt5 module but from what I've tried (on 3 related but distinct projects) it seems to work properly.

```
Found qmake: /usr/bin/qmake-qt6 (6.0.0)
Run-time dependency qt6 (modules: Core, DBus, Gui, Network, Widgets) found: YES 6.0.0 (qmake-qt6)
```

Fixes #8056 

I'll see what I can do about tests & docs

edit: Tests are a bit difficult - e.g. `test_qtdependency_pkgconfig_detection` because no pkg-config files seem to be built for Qt6 (by default?). `test_qt5dependency_qmake_detection` could be copy-pasted for qt6.

`docs/markdown/Qt6-module.md` could be a stub referencing Qt5-module.md as everything's literally the same. `docs/markdown/Dependencies.md` can get the "Qt4 & Qt5" header expanded to include Qt6

edit 2: Qt should be found using CMake now as CMake is the primary build system for Qt6 now - qmake is deprecated and pkg-config as mentioned above isn't generated (at least right now in Arch Linux). This will have to be implemented still

### DIFF
--- a/ci/ciimage/arch/install.sh
+++ b/ci/ciimage/arch/install.sh
@@ -12,7 +12,7 @@ pkgs=(
   itstool gtk3 java-environment=8 gtk-doc llvm clang sdl2 graphviz
   doxygen vulkan-validation-layers openssh mercurial gtk-sharp-2 qt5-tools
   libwmf valgrind cmake netcdf-fortran openmpi nasm gnustep-base gettext
-  python-lxml hotdoc rust-bindgen
+  python-lxml hotdoc rust-bindgen qt6-base qt6-tools
   # cuda
 )
 
--- a/docs/markdown/snippets/qt6_module.md
+++ b/docs/markdown/snippets/qt6_module.md
@@ -1,0 +1,8 @@
+## Qt6 module
+
+A module for Qt6 is now available with the same functionality as the Qt5
+module.
+
+Currently finding Qt6 is only available via `qmake` as pkg-config files aren't
+generated (see [QTBUG-86080](https://bugreports.qt.io/browse/QTBUG-86080)) and
+CMake support is not available for this module yet.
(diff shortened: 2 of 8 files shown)


jpakkane:
It would be nice to get this under proper testing, but at least Debian unstable does not have Qt6 packages yet...

z3ntu:
Arch Linux has it in the repos: https://www.archlinux.org/packages/extra/x86_64/qt6-base/

jpakkane:
There is no reason why Qt6 could not ship .pc files given that 5 did. This may be worth opening an upstream bug.

eli-schwartz:
There is an upstream bug here: https://bugreports.qt.io/browse/QTBUG-86080

jpakkane:
Ok, this needs two fixes then:

- a simple test (we can't run it in CI ATM but you should test it on your own machine)
- a release note snippet that says Qt 6 support is in, but it is experimental and that reliable support depends on the pkg-config bug with a link to the actual bug report linked above

z3ntu:
> a simple test

Should I copy-paste `test_qt5dependency_qmake_detection` for qt6?

> a release note snippet

Can do

Also I've written this for cmake support https://github.com/z3ntu/meson/commit/a91e27767c8c9090ae4ff60e06be2b1f63086f80 which isn't pretty and very incomplete but basically works (with the hardcoded components in there)

jpakkane:
> Should I copy-paste test_qt5dependency_qmake_detection for qt6?

Seems ok.

jpakkane:
Ping. The RC freeze is coming up, it would be nice to get this in the next release.

z3ntu:
Done, the test succeeds on my Arch Linux system (`./run_unittests.py -k test_qt6dependency_qmake_detection -v`), hopefully the CI likes it as well

PR #8087, opened 2020-12-10
Add option for thinLTO

### LINKED ISSUE #7493: Add support for ThinLTO when using Clang and LLD

### LINKED ISSUE #7820: meson does not support parallel linking for lto enabled builds
As per GCC documentation, it supports -flto[=n], and If you specify the optional n, the optimization and code generation done at link time is executed in parallel using n parallel jobs by utilizing an installed make program. The environment variable MAKE may be used to override the program used. 

As per meson built-in options b_lto can either be true or a false, and it would be great if meson supports passing "n" as well.

dcbaker:
This adds two separate builtin options. I'm not sure how else to solve this, given that we have two separate toggles to add, and we can't break the existing options.

Fixes #7493
Fixes #7820

### DIFF
--- a/docs/markdown/snippets/lto_threads.md
+++ b/docs/markdown/snippets/lto_threads.md
@@ -1,0 +1,7 @@
+## Knob to control LTO thread
+
+Both the gnu linker and lld support using threads for speading up LTO, meson
+now provides a knob for this: `-Db_lto_threads`. Currently this is only
+supported for clang and gcc. Any positive integer is supported, `0` means
+`auto`. If the compiler or linker implemnets it's on `auto` we use that,
+otherwise the number of threads on the machine is used.
--- a/docs/markdown/Builtin-options.md
+++ b/docs/markdown/Builtin-options.md
@@ -128,6 +128,7 @@ available on all platforms or with all compilers:
 | b_lundef      | true           | true, false                                                      | Don't allow undefined symbols when linking                                    |
 | b_lto         | false          | true, false                                                      | Use link time optimization                                                    |
 | b_lto_threads | 0              | Any integer*                                                     | Use multiple threads for lto. *(Added in 0.57.0)*                             |
+| b_lto_mode    | default        | default, thin                                                    | Select between lto modes, thin and default. *(Added in 0.57.0)*               |
 | b_ndebug      | false          | true, false, if-release                                          | Disable asserts                                                               |
 | b_pch         | true           | true, false                                                      | Use precompiled headers                                                       |
 | b_pgo         | off            | off, generate, use                                               | Use profile guided optimization                                               |
@@ -140,6 +141,8 @@ The value of `b_sanitize` can be one of: `none`, `address`, `thread`,
 `undefined`, `memory`, `address,undefined`.
 
 * < 0 means disable, == 0 means automatic selection, > 0 sets a specific number to use
+
+LLVM supports `thin` lto, for more discussion see [LLVM's documentation](https://clang.llvm.org/docs/ThinLTO.html)
 
 <a name="b_vscrt-from_buildtype"></a>
 The default value of `b_vscrt` is `from_buildtype`. The following table is
--- a/docs/markdown/Builtin-options.md
+++ b/docs/markdown/Builtin-options.md
@@ -119,30 +119,33 @@ no options.
 The following options are available. Note that they may not be
 available on all platforms or with all compilers:
 
-| Option      | Default value | Possible values         | Description |
-| ----------- | ------------- | ---------------         | ----------- |
-| b_asneeded  | true          | true, false             | Use -Wl,--as-needed when linking |
-| b_bitcode   | false         | true, false             | Embed Apple bitcode, see below |
-| b_colorout  | always        | auto, always, never     | Use colored output |
-| b_coverage  | false         | true, false             | Enable coverage tracking |
-| b_lundef    | true          | true, false             | Don't allow undefined symbols when linking |
-| b_lto       | false         | true, false             | Use link time optimization |
-| b_ndebug    | false         | true, false, if-release | Disable asserts |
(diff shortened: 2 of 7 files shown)


jpakkane:
Could we make it so that we can remove the `b_lto` option later? Such as having also the value `none` for `b_lto_mode` and print a warning whenever someone does `get_option('b_lto')`? Then we could remove it after some releases.

Similarly, could we always determine the number of LTO threads automatically given that we know the maximum number of parallel link jobs, so the number of threads would be num_cores / num_link_jobs? It would work locally at least, not sure about distributed builds, though.

dcbaker:
So my approach with the threads is to use the the number of virtual threads (cores + hyper threads) if the linker doesn't implement it's own auto jobs (basically just old versions of gcc don't). if it does implement it's own, I just let it make it's own decisions, its probably smarter than us anyway. I'm suspect that's going to be more efficient than us trying to guess how many compile/link jobs are going to be running at a time and trying to optimize for that.

I actually like having the mode separate from turning lto on and off, I could do something like this in my clang native file:
```ini
[binaries]
c = ['clang']
cpp = ['clang++']

[built-in options]
b_lto_mode = 'thin'
```
That way I get thin lto whenever lto gets turned on with clang, but i'm not actually turning on lto itself.

jpakkane:
> basically just old versions of gcc don't

In that case could we not remove the option and do the automatic thing and say that it is suboptimal for old versions of GCC. I suspect not that many people are using it anyway.

satmandu:
Could we have a b_lto_mode thin fail over to auto if using gcc? 

In [Chromebrew](https://github.com/skycocker/chromebrew) we are trying to set default options and it would be nice to be able to just set sane optimized defaults for most builds which don't need to be flipped over if we have to use clang.

(ChromeOS is build with libraries which require lld, due to Google's use of ```section `.relr.dyn'``` so  we do have to flip over to using lld a fair bit, and since using LTO with LLD requires clang, it would be nice to just flip over CC/CXX without also having to muck with b_LTO as well for every single package which requires this change.)

bonzini:
Do you get up to M*N processes if you have `ninja -jM` and N LTO jobs?

dcbaker:
@bonzini my understanding is that gcc and clang use threads not processes internally, so you'd end up with M processes and N threads

bonzini:
Should you add a separate pool for linking jobs to avoid running them in parallel?

jpakkane:
There already is one. :D (On the Ninja backend at least.)

jpakkane:
I could not come up with a backwards compatible way of doing this, so let's go with this, with the following:

- if `thin` is not available then fall back to full LTO
- take out the number of lto threads option

The latter can be sort-of-emulated with max linker jobs. If we need something fancier then we probably need to tie these two together somehow (and also with VS and possibly also the VS backend). But we should only do that after we have somewhat wider usage experience and know where the actual bottlenecks are.

Does this sound ok, or are there issues still?

nickbroon:
Can you give an example of controlling the max number of linker threads?

I have a project that links just a single very large executable, using LTO, and the difference between `-flto` and `-flto=[n]` is very noticeable (when setting n to `nproc`).

jpakkane:
From what I can tell the linkers should enable multithreading automatically. If not, then we should set that by default so it works out of the box. So maybe to something like `number_of_processors/max_link_jobs`?

dcbaker:
No, they only enable it if you tell them to AFAICT

dcbaker:
the commit you asked me to drop added support for controlling that, and would give them their `auto` option if you set the number of jobs to `0`, otherwise the number you asked for.

jpakkane:
I misread that then. Well at the very least the behaviour now should be what `auto` should have done.

Related, it would be nice if there was a way to tell Ninja that some process is going to use more cores than one so it could use that info to drive process spawning.

dcbaker:
Okay, threads support added back.

To be clear, the default for threads is now 0, for clang 0 means `pick a number internally`, for gcc we convert that to `auto` if the version of gcc supports that, or `number of logical cores` if not.

nickbroon:
That behaviour for `b_lto_threads` seems reasonable, predictable and without surprise. 👍

nickbroon:
Is there anything else needed for the PR before it can be considered for merge, for 0.57? Are the 2 failed checks expected?

jpakkane:
⚔️ conflicts.

dcbaker:
Fixed.

docs/markdown/snippets/lto_threads.md
```diff
@@ -0,0 +1 @@
+## Supported
```
jpakkane, line 1: This should probably say something more. :)

PR #8090, opened 2020-12-11
Use currently detected Ninja version as the minimum required.

jpakkane:
This simplifies the implementation of scanner based modules in Fortran and C++.


eli-schwartz:
How? If meson needs to be capable of generating a compatible `build.ninja` for systems where the detected ninja is `1.8.2`, then meson cannot be simplified however you're thinking, in fact it would grow more complex due to needing to support multiple implementations of $functionality. The generated build.ninja could be simplified, but I'm not sure why that's an accomplishment?

This WILL be inconvenient/break things if meson detects one implementation of ninja, assumes it can use brand new features, and then the user invokes a different implementation with a lower compatibility level. I use @michaelforney's samu all the time:

```
$ samu --version
1.9.0
```

But ninja is installed too (it must be, due to shortcomings in pacman, as a dependency of meson).

```
$ ninja --version
1.10.2
```

Since meson will prefer ninja over samu, this PR will make build.ninja require 1.10.2, which then prevents me from running samu to build.

jpakkane:
C++ (and Fortran if you do it properly) modules require Ninja's dynamic dependency support, which is only available in 1.10. We only want to require that if we actually build with those. Now unfortunately the minimum version must be written in the file before any targets are processed so we don't actually know yet if they are used.

In the module branch I go the other way and only enable module scanning if Ninja is new enough. This is the simple solution, which unfortunately assumes that Ninja versions never go backwards.

eli-schwartz:
If meson won't even be checking to see if it's *using* the new features before raising the minimum required version, I think that means meson should learn how to delay writing the build.ninja until it has all the info it needs.

jpakkane:
It turns out that you can use newer stuff even if your minimum advertised version is older. The end result looks a bit silly but works, so closing.

PR #8071, opened 2020-12-07
Add support for driving lld-link indirectly through clang on Windows

e820:
Fixes #8064

This PR adds support for using clang with a non-GNU compatible `lld` instance, which results in correct
host machine identification in cross-compilation scenarios aswell as support for using `lld-link` indirectly
through `clang` by selecting `lld` through `c_ld`.

PR #8094, opened 2020-12-14
Make win_subsystem a linker property

e820:
This is more of a little cleanup amendment to #8071 which moves `get_win_subsystem_args` to the linker side
instead. `get_gui_app_args` is left mostly unchanged as both clang and gcc have special options to handle those two
subsystems directly from the compiler driver.

### DIFF
--- a/mesonbuild/compilers/mixins/clang.py
+++ b/mesonbuild/compilers/mixins/clang.py
@@ -19,7 +19,7 @@ import shutil
 import typing as T
 
 from ... import mesonlib
-from ...linkers import AppleDynamicLinker, ClangClDynamicLinker
+from ...linkers import AppleDynamicLinker
 from ..compilers import CompileCheckMode
 from .gnu import GnuLikeCompiler
 
--- a/mesonbuild/compilers/mixins/clang.py
+++ b/mesonbuild/compilers/mixins/clang.py
@@ -109,12 +109,6 @@ class ClangCompiler(GnuLikeCompiler):
             # Shouldn't work, but it'll be checked explicitly in the OpenMP dependency.
             return []
     
-    def get_win_subsystem_args(self, value: str) -> T.List[str]:
-        if self.info.is_windows() and not self.info.is_cygwin() and isinstance(self.linker, ClangClDynamicLinker):
-            return [f'-Wl,/subsystem:{value}']
-        
-        return super().get_win_subsystem_args(value)
-
     @classmethod
     def use_linker_args(cls, linker: str) -> T.List[str]:
         # Clang additionally can use a linker specified as a path, which GCC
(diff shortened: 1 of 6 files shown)


dcbaker (approved):
Thanks, this is a nice cleanup. Once the CI turns green I'll merge this.

The macos jobs are notoriously unreliable unfortunately.

e820:
> Thanks, this is a nice cleanup. Once the CI turns green I'll merge this.
> 
> The macos jobs are notoriously unreliable unfortunately.

Yea, they're certainly annoying :b.

dcbaker:
I give up. it's the same stupid tests. Theres the MPI test and one D language test that are always the cuprits. And of course no one can reproduce on their local systems...

mesonbuild/linkers.py
```diff
@@ -781,6 +802,13 @@ def get_allow_undefined_args(self) -> T.List[str]:
             return self._apply_prefix('--allow-shlib-undefined')
         return []
 
+    def get_win_subsystem_args(self, value: str) -> T.List[str]:
+        m = env.machines[self.for_machine]
+        if not m.is_windows() or m.is_cygwin():
+            return []
+
+        return self._apply_prefix([f'-subsystem:{value}'])
```
mstorsjo, line 806: This broke linking on windows with LLD in mingw mode. See above (lines 785), "Representation of LLVM's ld.lld linker. This is only the gnu-like linker, not the apple like or link.exe like linkers." - so this should use the GNU style syntax `--subsystem,windows` etc.

I'll make a PR that moves that code up from GnuDynamicLinker into GnuLikeDynamicLinkerMixin so that the same code also is used for the gnu-like uses of lld on windows.

mesonbuild/compilers/mixins/clang.py
```diff
@@ -109,12 +109,6 @@ def openmp_flags(self) -> T.List[str]:
             # Shouldn't work, but it'll be checked explicitly in the OpenMP dependency.
             return []
     
-    def get_win_subsystem_args(self, value: str) -> T.List[str]:
-        if self.info.is_windows() and not self.info.is_cygwin() and isinstance(self.linker, ClangClDynamicLinker):
-            return [f'-Wl,/subsystem:{value}']
```
mstorsjo, line 114: I see that the code that ended up breaking lld for mingw targets was moved from here. The bit that was lost in the move was the `isinstance(self.linker, ClangClDynamicLinker)` bit, which is why it worked before.

mesonbuild/linkers.py
e820: Good catch! I must have been under the impression that the ld compatible driver supported the link-esque subsystem syntax for some reason.

PR #8095, opened 2020-12-14
Extend the C++ module scanner to handle Fortran, too.

jpakkane:
This will probably fail to the tests, but let's just get this out there for people to test and comment on. So ping all Fortran users, such as @scivision.

The main advantage of this approach is that we can drop all the prescan hacks we have and instead do Fortran automatically and reliably out of the box. I.e. moving module definitions from one file to another should just work without needing to run Meson reconfiguration.

### DIFF
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -117,7 +117,7 @@ raw_names = {'DEPFILE_UNQUOTED', 'DESC', 'pool', 'description', 'targetdep', 'dy
 NINJA_QUOTE_BUILD_PAT = re.compile(r"[$ :\n]")
 NINJA_QUOTE_VAR_PAT = re.compile(r"[$ \n]")
 
-def ninja_quote(text, is_build_line=False):
+def ninja_quote(text: str, is_build_line=False) -> str:
     if is_build_line:
         quote_re = NINJA_QUOTE_BUILD_PAT
     else:
@@ -872,7 +872,11 @@ int dummy;
         self.generate_shlib_aliases(target, self.get_target_dir(target))
         self.add_build(elem)
 
-    def should_scan_target(self, target):
+    def should_use_dyndeps_for_target(self, target: 'build.BuildTarget') -> bool:
+        if mesonlib.version_compare(self.ninja_version, '<1.10.0'):
+            return False
+        if 'fortran' in target.compilers:
+            return True
         if 'cpp' not in target.compilers:
             return False
         # Currently only the preview version of Visual Studio is supported.
@@ -883,18 +887,16 @@ int dummy;
             return False
         if mesonlib.version_compare(cpp.version, '<19.28.28617'):
             return False
-        if mesonlib.version_compare(self.ninja_version, '<1.10.0'):
-            return False
         return True
 
     def generate_dependency_scan_target(self, target, compiled_sources, source2object):
-        if not self.should_scan_target(target):
+        if not self.should_use_dyndeps_for_target(target):
             return
         depscan_file = self.get_dep_scan_file_for(target)
         pickle_base = target.name + '.dat'
         pickle_file = os.path.join(self.get_target_private_dir(target), pickle_base).replace('\\', '/')
         pickle_abs = os.path.join(self.get_target_private_dir_abs(target), pickle_base).replace('\\', '/')
-        rule_name = 'cppscan'
+        rule_name = 'depscan'
         scan_sources = self.select_sources_to_scan(compiled_sources)
         elem = NinjaBuildElement(self.all_outputs, depscan_file, rule_name, scan_sources)
         elem.add_item('picklefile', pickle_file)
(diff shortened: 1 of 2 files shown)


mesonbuild/scripts/depscan.py
```diff
@@ -15,12 +15,23 @@
...
+cpp_import_re = re.compile('\w*import ([a-zA-Z0-9]+);')
+cpp_export_re = re.compile('\w*export module ([a-zA-Z0-9]+);')
+
+FORTRAN_INCLUDE_PAT = r"^\s*#?include\s*['\"](\w+\.\w+)['\"]"
+FORTRAN_MODULE_PAT = r"^\s*\bmodule\b\s+(\w+)\s*(?:!+.*)*$"
+FORTRAN_SUBMOD_PAT = r"^\s*\bsubmodule\b\s*\((\w+:?\w+)\)\s*(\w+)"
+FORTRAN_USE_PAT = r"^\s*use,?\s*(?:non_intrinsic)?\s*(?:::)?\s*(\w+)"
+
+fortran_module_re = re.compile(FORTRAN_MODULE_PAT)
+fortran_use_re = re.compile(FORTRAN_USE_PAT)
```
dcbaker, line 34: These should really all be all caps per PEP8, as they're module constants.

mesonbuild/backend/ninjabackend.py
```diff
@@ -907,10 +909,15 @@ def select_sources_to_scan(self, compiled_sources):
         # in practice pick up C++ and Fortran files. If some other language
         # requires scanning (possibly Java to deal with inner class files)
         # then add them here.
+        all_suffixes = set()
+        for s in compilers.lang_suffixes['cpp']:
+            all_suffixes.add(s)
+        for s in compilers.lang_suffixes['fortran']:
+            all_suffixes.add(s)
```
dcbaker, line 916:
```suggestion
        all_suffixes = set(compilers.lang_suffixes['cpp']) | set(compilers.lang_suffixes['fortran'])
```

dcbaker (changes requested):
I have a few small nits, but overall this looks good to me. Probably should get scivision to give his input though.

mesonbuild/scripts/depscan.py
```diff
@@ -34,9 +45,19 @@ def __init__(self, pickle_file: str, outfile: str, sources: T.List[str]):
...
-        for line in pathlib.Path(fname).read_text().split('\n'):
-            import_match = import_re.match(line)
-            export_match = export_re.match(line)
+        suffix = os.path.splitext(fname)[1][1:]
+        if suffix in lang_suffixes['fortran']:
+            self.scan_fortran_file(fname)
+        elif suffix in lang_suffixes['cpp']:
+            self.scan_fortran_file(fname)
+        else:
+            sys.exit('Can not scan files with suffix .{}.'.format(suffix))
```
scivision, line 54: I think `raise SystemExit` is preferred to `sys.exit`.  However, in this case should it rather be `raise RuntimeError`?

mesonbuild/scripts/depscan.py
```diff
@@ -15,12 +15,23 @@
...
 
 from ..backend.ninjabackend import TargetDependencyScannerInfo
+from ..compilers.compilers import lang_suffixes
 
-import_re = re.compile('\w*import ([a-zA-Z0-9]+);')
-export_re = re.compile('\w*export module ([a-zA-Z0-9]+);')
+cpp_import_re = re.compile('\w*import ([a-zA-Z0-9]+);')
+cpp_export_re = re.compile('\w*export module ([a-zA-Z0-9]+);')
+
+FORTRAN_INCLUDE_PAT = r"^\s*#?include\s*['\"](\w+\.\w+)['\"]"
```
scivision, line 28: This could be in a separate PR, but to speed this up, the `#include` syntax isn't used by any compiler I've found. I added this when I was overhauling the regex for Fortran, but since realized no current compiler allows this syntax. 

Suggest:
```python
FORTRAN_INCLUDE_PAT = r"^\s*include\s*['\"](\w+\.\w+)['\"]"
```

scivision:
Most of the Fortran project tests fail with this setup (only one I tried):

* Ninja 1.10.2
* GCC 10.2.0 (MSYS2)
* Windows 10

Typical build-time error:

```
ninja: build stopped: dependency cycle: use_syntax.exe.p/circle.mod -> use_syntax.exe.p/circle.mod.
```

scivision:
3cd6586 fails at a list | list I think maybe it's meant to have those variables be sets?

jpakkane:
Finally it passes all project tests. I'll fix the mypy issues later.

scivision:
Yes this works for me with Gfortran 10 as well. 

I can't effectively test with the Intel Windows compiler yet, as the Intel oneAPI "gold" official release a few weeks ago actually broke a bunch of things that previously worked throughout the beta period, including independently breaking Meson and CMake.

https://community.intel.com/t5/Intel-oneAPI-HPC-Toolkit/Gold-oneAPI-release-broken-but-betas-worked-fine-Linux-Windows/m-p/1236017#M7421

Intel has confirmed the issues and referred to internal team.

jpakkane:
That is unrelated to this issue, so merging.

scivision:
yes that's correct, thanks

PR #8096, opened 2020-12-14
cmake: Revert to using self.for_machine instead of MachineChoice.BUILD (fixes #8028)

### LINKED ISSUE #8028: host zlib is found using a cross file.
**Describe the bug**
When using a cross file which is not setting pkg_config_libdir, meson find the host version of zlib with

```
MESON build-cross --cross-file=path_to/my-cross-file.txt
```

**To Reproduce**

Meson file:
```
project('hellomeson', 'c')
cc = meson.get_compiler('c')
libz_dep = dependency('zlib', required : false)
if not libz_dep.found()
warning('unable to find libz')
endif
executable('hello-meson', 'src/hello.c')
```

Cross file:

```
[host_machine]
system = 'linux'
cpu_family = 'aarch64'
cpu = 'arm64'
endian = 'little'

[properties]
c_args = []
cpp_args = []
objc_args = []
objcpp_args = []
c_link_args = []
cpp_link_args = []
objc_link_args = []
objcpp_link_args = []
pkg_config_libdir = ['__unknown_sysroot__']


[binaries]
c = ['aarch64-linux-gnu-gcc']
cpp = ['aarch64-linux-gnu-g++']
ar = ['aarch64-linux-gnu-ar']
pkgconfig = 'pkg-config'
strip = ['aarch64-linux-gnu-strip']
```

**Expected behavior**
To not find any zlib or at least not the one from my host and then use it..

**system parameters**
cross build
Ubuntu 18.04
3.6.9
* what `meson --version`
master
1.8.2

mensinda:


jpakkane:
This could use a test to prevent re-regressions.

mensinda:
@jpakkane I tried to come up with one but this is not that trivial:

- Our cross files are already "fixed" since the required combination of CMake variables is already specified
- This change is basically just us saying "do not automatically find CMake and thus do not use CMake as a dependency backend automatically"

PR #8099, opened 2020-12-15
Fix the subsystem options for lld in mingw mode

mstorsjo:
lld in gnu like mode (e.g. for mingw) needs these options in
the same for as gnu ld, thus remove the lld specific code bit
and move the code for gnu like options into GnuLikeDynamicLinkerMixin.

This unbreaks linking with lld for mingw targets after
2fb4d1f7512a4ec125883fd65115ab33a5b06b6f.

CC @dcbaker @e820

### DIFF
--- a/mesonbuild/linkers.py
+++ b/mesonbuild/linkers.py
@@ -668,6 +668,18 @@ class GnuLikeDynamicLinkerMixin:
 
         return (args, rpath_dirs_to_remove)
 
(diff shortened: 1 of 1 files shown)


e820:
This looks good, as I said in my comment on the closed PR this was an oversight on my part. Thanks.

e820 (approved):

dcbaker (approved):

dcbaker:
Thanks for fixing this!

PR #8101, opened 2020-12-15
coredata: Add missing nopromote wrap_mode choice

xclaesse:


jpakkane:
This could use a test, given that the functionality has never been actually used yet.

jpakkane:
ping

xclaesse:
> This could use a test, given that the functionality has never been actually used yet.

Indeed, yet another proof that if it's not tested it's not working... Added the test now.

PR #8103, opened 2020-12-15
Fix some layering violations and add some type annotations in the interpreter

dcbaker:
This was peeled out of my rust module series, as it's not strictly related, and should be easy enough to review and land now.


mesonbuild/build.py
```diff
@@ -140,9 +187,9 @@ def __init__(self, environment: environment.Environment):
         self.projects_link_args = PerMachine({}, {})  # type: PerMachine[T.Dict[str, T.List[str]]]
         self.tests = []                               # type: T.List['Test']
         self.benchmarks = []                          # type: T.List['Test']
-        self.headers = []
-        self.man = []
-        self.data = []
+        self.headers: T.List[Headers] = []
+        self.man: T.List[Man] = []
+        self.data = []                                # type: T.List[Data]
```
mensinda, line 192: This should also be the new syntax now...

mensinda (commented):
LGTM, besides the one minor issue below.

mesonbuild/build.py
dcbaker: Ah, good catch. I've fixed that, and updated the Build initializer to use only the new syntax for consistency.

mensinda (approved):
LGTM, assuming the CI is green

Issue #8089, opened 2020-12-11
cmake subproject dependencies not working


kcgen:
**Describe the bug**
DOSBox Staging is transitioning to meson. We depend on FluidSynth (2.x, a CMake project), which in turn depends on Glib-2.0 (a Meson project), both of which aren't guaranteed to exist on all our supported platforms - so we're written Meson wraps for them.

Unfortunately we're struggling to provide the FluidSynth subproject with the Glib-2.0 dependencies generated from the Glib-2.0 subproject.

``` python
glib_dep = dependency('glib-2.0', fallback : ['glib'])
gthread_dep = dependency('gthread-2.0', fallback : ['glib'])
```

All goes well if Glib doesn't exist -- Meson falls back to the `glib` wrap and reports the dependencies are found:

``` text
Run-time dependency glib-2.0 found: NO (tried pkgconfig and cmake)
Looking for a fallback subproject for the dependency glib-2.0
|Executing subproject glib method meson 
|Project name: glib
|Project version: 2.67.1
... 
Dependency glib-2.0 found: YES 2.67.1 (overridden)
Dependency gthread-2.0 found: YES 2.67.1 (overridden)
```

Great - so Meson is taking care of that for us. Onto FluidSynth..

``` text
Executing subproject fluidsynth method cmake 
||Configuring the build directory with CMake version 3.16.3
||-- Checking for modules 'glib-2.0>=2.6.5;gthread-2.0>=2.6.5'
||--   No package 'glib-2.0' found
||--   Package 'glib-2.0', required by 'gthread-2.0', not found
||  A required package was not found
```

So despite Meson telling us it has control of `glib-2.0 found: YES 2.67.1` and ` gthread-2.0 found: YES 2.67.1`, its still unable to provide them to the cmake project, despite configuring it with these dependencies:

``` python
fluidsynth_proj = cmake.subproject('fluidsynth')
fluidsynth_proj.dependency('glib-2.0', fallback: ['glib'])
fluidsynth_proj.dependency('gthread-2.0', fallback: ['glib'])

fluidsynth_dep = fluidsynth_proj.dependency(
  'fluidsynth',
  version : '>=2',
  fallback: ['fluidsynth'],
  dependencies: ['glib-2.0', 'gthread-2.0']
)
```

**To Reproduce**

See attached tarball consisting of tiny 19-line `meson.build` and two wraps to reproduce the issue.

[...]

PR #8104, opened 2020-12-15
cmake: Enable support for using -uninstalled.pc files

mensinda:
This PR enables the usage of `*-uninstalled.pc` files for CMake subprojects.

The test case is a bit hacky since CMake just does not like spaces in the `pkg-config` output...

Related issue: #8089


jpakkane:
⚔️ conflicts. If CMake can't handle spaces then it might make sense to run it as a unittest by copying the entire tree to a spaceless dir and then running it by hand.

mensinda:
Done, should hopefully pass now...

jpakkane:
Msys fails. :(

mensinda:
Seems like there are even more problems with `\` in paths (which CMake also does not correctly handle)...

I guess the only reliable way to fix this would be to always use POSIX paths for *everything*. I am afraid that I won't get this running in time for 0.57 on windows...

kcgen:
@mensinda , just curious if this will make in the queue for 0.58?

tristan957:
@kiwixz wanna take this PR over?

kiwixz:
I tried rebasing it, but after 3499 commits on master there is not much that fits.  I see you're already taking care of the relative_to in #12087; so whats left is basically just the hack to remove spaces from paths (if that still needed with newer CMake ?).  If someone where to work on this, they should probably start from scratch and just salvage the test cases.

tristan957:
@kiwixz thanks for taking a look. I will close the PR.

PR #8105, opened 2020-12-16
Fix clang-cl not being detected properly if c contains a path

e820:
Fixes a small issue I came across when looking into making meson drive `clang-cl` properly on unix build machines. Previously if you supplied a full path to the `clang-cl` binary in a cross-file `_detect_c_or_cpp_compiler` would erroneously try to detect the compiler as a gnu-compatible `clang` instance as it missed the cl check and passed `--version` instead of `"/?"` which ends up not printing the cl compatibility string.

CC @dcbaker

### DIFF
--- a/mesonbuild/environment.py
+++ b/mesonbuild/environment.py
@@ -1196,7 +1196,7 @@ class Environment:
                 compiler = [compiler]
             compiler_name = os.path.basename(compiler[0])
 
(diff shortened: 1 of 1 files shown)


dcbaker (changes requested):
This looks correct to me. We need a test in `run_unttests.py` or this is going to get regressed again at some point.

lukester1975:
Looks like this should be closed thanks to #9361.

bonzini:
#9361 was merged years ago, closing.

PR #8106, opened 2020-12-17
environment.py: Detect all mips* architectures

persianpros:
We have more than those values, like:

mipsel
mipsel-nf
mips32el
mips33el-nf

So lets just detect them all.

### DIFF
--- a/mesonbuild/environment.py
+++ b/mesonbuild/environment.py
@@ -377,8 +377,8 @@ def detect_cpu_family(compilers: CompilersDict) -> str:
         trial = 'x86_64'
     elif trial in {'sun4u', 'sun4v'}:
         trial = 'sparc64'
-    elif trial in {'mipsel', 'mips64el'}:
-        trial = trial.rstrip('el')
+    elif trial.startswith('mips'):
+        trial = 'mips'
     elif trial in {'ip30', 'ip35'}:
         trial = 'mips64'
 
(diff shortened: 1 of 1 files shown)


mesonbuild/environment.py
```diff
@@ -377,8 +377,8 @@ def detect_cpu_family(compilers: CompilersDict) -> str:
         trial = 'x86_64'
     elif trial in {'sun4u', 'sun4v'}:
         trial = 'sparc64'
-    elif trial in {'mipsel', 'mips64el'}:
-        trial = trial.rstrip('el')
+    elif trial.startswith('mips'):
+        trial = 'mips'
```
eli-schwartz, line 381: This no longer returns 'mips64' in situations where it used to, so I'm not sure how you expect this to work.

eli-schwartz (changes requested):

mesonbuild/environment.py
persianpros: Oh I forgot that I just tested the 32 bit, let me send another one soon.

eli-schwartz:
Please don't close PRs like that whenever changes are requested. Just push an update to the current PR.

persianpros:
Sure I do, I just didn't keep my fork for the last one so I had to fork again.

Sorry for any inconvenience.

PR #8109, opened 2020-12-17
VS: copy dll dependencies from import libraries next to executable

vvainola:
- Copy dependent dlls next to executable targets automatically. This allows
 executables that are linked with import libraries to be runnable automatically in
 the build directory. This makes it easier to use found libraries from e.g. conan
 or vcpkg.
- Add option "b_copy_deps" that by default copies the dlls but it has
 an option to use symlink in case there are many dependencies that copying
 the dlls would take needlessly much space when building multiple configurations.
 When installing the dlls are always copied to install dir.
Base options seems a bit wrong for this as this is not a flag for the compiler
but this is not backend specific either.

How can I implement tests for this? How to get an import library for linking
and test that trying to run the target does not work without this because
the dll is not found next to executable and it does not exist in PATH.

### DIFF
--- a/docs/markdown/Builtin-options.md
+++ b/docs/markdown/Builtin-options.md
@@ -126,7 +126,8 @@ platforms or with all compilers:
 | b_sanitize  | none          | see below               | Code sanitizer to use |
 | b_staticpic | true          | true, false             | Build static libraries as position independent |
 | b_pie       | false         | true, false             | Build position-independent executables (since 0.49.0)|
-| b_vscrt     | from_buildtype| none, md, mdd, mt, mtd, from_buildtype, static_from_buildtype | VS runtime library to use (since 0.48.0) (static_from_buildtype since 0.56.0) |
+| b_vscrt     | from_buildtype| none, md, mdd, mt, mtd, from_buildtype, static_from_buildtype | VS runtime library to use (since 0.48.0) (static_from_buildtype since 0.56.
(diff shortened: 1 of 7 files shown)


zeehio:
It looks convenient. I may need something similar to this soon.

I was thinking of just  setting `meson configure --bindir bin --libdir bin...`. I believe that option would also work since that would install the DLL next to the executable as far as I understand. What use case does your pull request address that I am missing?

vvainola:
I'm not aware of any functionality in meson that it would automatically handle the dll copying or adding them to PATH to run exes.
 
The use case is for example when using cmake_toolchain_file with vcpkg that meson currently finds X.lib from vcpkg/installed/x64-windows/lib/X.lib and adds that to link args. However when trying to debug in Visual Studio it will error with "The code execution cannot proceed because X.dll was not found.". This PR adds that vcpkg-installed/x64-windows/bin/X.dll will also be copied next to executable targets so that running them in VS works

jpakkane:
We have not traditionally done this on purpose because this is an easy way to get into DLL hell. If system versions are updated but local ones are not, then things fail quite badly very quickly. The simpler (and I think recommended by most people) solution is to add your dependency prefix's `lib` directory to `PATH`. This should be fairly easy to do in VS and others (IIRC though, I have not done it myself).

This is a bit stupid and a deficiency on Windows's part. It is possible to create a manifest file for executable to tell where their dependency libraries should be looked up in. That would be a better solution for this problem, but I seem to remember that it had some strange limitations.

Issue #8115, opened 2020-12-21
depfixer doesn't dedup properly when new RPATH has multiple entries


alanc:
**Describe the bug**
Attempting to build gnome-settings-daemon 3.34.2 on Solaris 11.4 with meson 0.56 failed with:
```
Installing plugins/smartcard/gsd-smartcard to /export/alanc/gate/components/gnome/gnome-settings-daemon/build/prototype/sparc/usr/lib/sparcv9
New rpath must not be longer than the old one.
 Old: /usr/lib/sparcv9/gnome-settings-daemon-3.0:/usr/lib/mps/64:/usr/lib/mps/64:/usr/lib/mps/sparcv9
 New: /usr/lib/sparcv9/gnome-settings-daemon-3.0:/usr/lib/mps/64:/usr/lib/sparcv9/gnome-settings-daemon-3.0:/usr/lib/mps/64:/usr/lib/mps/sparcv9
```
which clearly has repeated path entries.

After debugging, I found the problem was the old path entries were added as a single string (`b'/usr/lib/sparcv9/gnome-settings-daemon-3.0:/usr/lib/mps/64'`) so the entries in it were not correctly compared against the new entries to add.

This simple change solved the problem for this package:

```
diff --git a/mesonbuild/scripts/depfixer.py b/mesonbuild/scripts/depfixer.py
index 18d70cca9..41b5e51ee 100644
--- a/mesonbuild/scripts/depfixer.py
+++ b/mesonbuild/scripts/depfixer.py
@@ -313,7 +313,8 @@ class Elf(DataSizes):
         # Only add each one once.
         new_rpaths = OrderedSet()  # type: OrderedSet[bytes]
         if new_rpath:
-            new_rpaths.add(new_rpath)
+            for rpath_dir in new_rpath.split(b':'):
+                new_rpaths.add(rpath_dir)
         if old_rpath:
             # Filter out build-only rpath entries
             # added by get_link_dep_subdirs() or
```

I'm now testing the build of the rest of our meson-built packages to verify it doesn't break them, and will then submit a pull request.

PR #8116, opened 2020-12-22
RPATH fixes

alanc:
Fixes issues with RPATH handling found when updating the Solaris meson packages to 0.56.0.

The first copies code from the GNU linker class to the equivalent code in the Solaris linker class.

The second makes the handling of old rpath entries match that of new rpath entries in depfixer (#8115)

### DIFF
--- a/mesonbuild/scripts/depfixer.py
+++ b/mesonbuild/scripts/depfixer.py
@@ -313,7 +313,7 @@ class Elf(DataSizes):
         # Only add each one once.
         new_rpaths = OrderedSet()  # type: OrderedSet[bytes]
         if new_rpath:
-            new_rpaths.add(new_rpath)
+            new_rpaths.update(new_rpath.split(b':'))
         if old_rpath:
             # Filter out build-only rpath entries
             # added by get_link_dep_subdirs() or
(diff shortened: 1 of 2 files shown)


mesonbuild/scripts/depfixer.py
```diff
@@ -313,7 +313,8 @@ def fix_rpathtype_entry(self, rpath_dirs_to_remove: T.List[bytes], new_rpath: by
         # Only add each one once.
         new_rpaths = OrderedSet()  # type: OrderedSet[bytes]
         if new_rpath:
-            new_rpaths.add(new_rpath)
+            for rpath_dir in new_rpath.split(b':'):
+                new_rpaths.add(rpath_dir)
```
eli-schwartz, line 317: `new_rpaths.update(new_rpath.split(b':'))`

eli-schwartz (changes requested):

dcbaker (approved):
LGTM, thanks Alan.

eli-schwartz (approved):

eli-schwartz:
Seems eminently reasonable to me (and I'm not sure why the second commit wasn't the case all along).

PR #8119, opened 2020-12-22
Implement support of dlang -makedeps switch

### LINKED ISSUE #8118: Support for DMD -makedeps switch
A new feature has recently been merged into DMD (see dlang/dmd#12011) to implement emission of dependency file in Makefile format.
This should be available in DMD v2.096, and in LDC when it will use the DMD frontend of the same version.

Meson can add this switch to the ninja D language rules in order to have correct dependencies and retrigger builds in an appropriate way.

rtbo:
Fix #8118

If DMD version is `>=2.096`, `-makedeps` switch can be emitted.
For LDC, it checks that the DMD frontend is `>=2.096`.

Kind of difficult to have automated test at the moment because even the current development version of DMD does not yet advertize the `2.096` version.
But according [DMD release policy](https://dlang.org/changelog/release-schedule.html), the PR implementing the change (dlang/dmd#12011 - merged Dec. 2nd 2020) should be in the `2.096` version.

I could however test the feature by tricking the version comparison.

### DIFF
--- a/mesonbuild/compilers/d.py
+++ b/mesonbuild/compilers/d.py
@@ -12,7 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import os.path, re, subprocess
+import os.path
+import re
+import subprocess
 import typing as T
 
 from ..mesonlib import (
@@ -80,8 +82,12 @@ class DmdLikeCompilerMixin(CompilerMixinBase):
     sharing between them as makes sense.
     """
 
-    def __init__(self, dmd_frontend_version: str):
-        self._frontend_version = dmd_frontend_version
+    def __init__(self, dmd_frontend_version: T.Optional[str]):
+        if dmd_frontend_version is None:
+            self._dmd_has_depfile = False
+        else:
+            # -makedeps switch introduced in 2.095 frontend
+            self._dmd_has_depfile = version_compare(dmd_frontend_version, ">=2.095.0")
 
     if T.TYPE_CHECKING:
         mscrt_args = {}  # type: T.Dict[str, T.List[str]]
@@ -135,8 +141,7 @@ class DmdLikeCompilerMixin(CompilerMixinBase):
         return 'deps'
 
     def get_dependency_gen_args(self, outtarget: str, outfile: str) -> T.List[str]:
-        # -makedeps switch introduced in 2.096 frontend
-        if version_compare(self._frontend_version, ">=2.096"):
+        if self._dmd_has_depfile:
             return [f'-makedeps={outfile}']
         return []
 
@@ -698,11 +703,13 @@ class GnuDCompiler(GnuCompiler, DCompiler):
 # the common features between LDC and DMD.
 # We need the complete version text because the match is not on first line
 # of version_output
-def find_ldc_dmd_frontend_version(version_output: str):
+def find_ldc_dmd_frontend_version(version_output: T.Optional[str]) -> T.Optional[str]:
+    if version_output is None:
+        return None
     version_regex = re.search(r'DMD v(\d+\.\d+\.\d+)', version_output)
-    if version_regex is not None and len(version_regex.groups()):
-        return version_regex.groups()[0]
-    return ''
+    if version_regex:
+        return version_regex.group(1)
+    return None
 
 class LLVMDCompiler(DmdLikeCompilerMixin, DCompiler):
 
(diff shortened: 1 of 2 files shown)


eli-schwartz:
Thank you. Note that it's not a meson requirement to open both a github issue and a github PR for every proposed change.

...

Do you know when the new DMD version is likely to be released?

mesonbuild/compilers/d.py
```diff
@@ -683,6 +692,17 @@ def get_linker_always_args(self) -> T.List[str]:
     def get_disable_assert_args(self) -> T.List[str]:
         return ['-frelease']
 
+# LDC uses the DMD frontend code to parse and analyse the code.
+# It then uses LLVM for the binary code generation and optimizations.
+# This function retrieves the dmd frontend version, which determines
+# the common features between LDC and DMD.
+# We need the complete version text because the match is not on first line
+# of version_output
+def find_ldc_dmd_frontend_version(version_output: str):
```
dcbaker, line 701:
```suggestion
def find_ldc_dmd_frontend_version(version_output: str) -> str:
```

mesonbuild/compilers/d.py
```diff
@@ -683,6 +692,17 @@ def get_linker_always_args(self) -> T.List[str]:
...
+    version_regex = re.search(r'DMD v(\d+\.\d+\.\d+)', version_output)
+    if version_regex is not None and len(version_regex.groups()):
+        return version_regex.groups()[0]
```
dcbaker, line 704: Two things here, using `len()` in python to check if a container is empty is considered an anti-pattern, there's a well defined conversion from container to bool in logiacl operations, so just `if version_regex.groups()` will test if it's empty or not. The problem with len() is that it's O(n), but using the bool conversion will short circuit as soon as a single member is found.

In thisc as though I think we can simplify this to:
```suggestion
    if version_regex:
        return version_regex.group(1)
```
Note that group 0 is the entire match.

mesonbuild/compilers/d.py
```diff
@@ -691,10 +711,11 @@ def __init__(self, exelist: T.List[str], version: str, for_machine: MachineChoic
                  exe_wrapper: T.Optional['ExternalProgram'] = None,
                  linker: T.Optional['DynamicLinker'] = None,
                  full_version: T.Optional[str] = None,
-                 is_cross: bool = False):
+                 is_cross: bool = False, version_output: str):
```
dcbaker, line 714: You can't have arguments without default values after those with default values, so either this needs to have a default value, or it needs to be moved before the ones with default value.

I think I'd be tempted to make `find_ldc_dmd_frontend_version` return an `Optional[str]`, and allow this to be `Optional[str] = None`, then for old versions you can just check if the value is None.

mesonbuild/compilers/d.py
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import os.path, subprocess
+import os.path, re, subprocess
```
dcbaker, line 15: We've been doing an inconsistent job of this, but I think we're moving to top level imports as one per line but multiple import in a `from module import ...`, i.e.:
```python
import re
import typing

from foo import a, b, c
```

dcbaker (changes requested):
Thanks for working on this! I've left a few review comments inline, but the approach looks good and once we get these resolved I'd be happy to get this merged.

rtbo:
@dcbaker thanks for the detailed review, I'll look into this.

> Do you know when the new DMD version is likely to be released?
> EDIT: apparently beginning of March.

I've seen that my PR has been merged into DMD `v2.095.0-beta.1`.
This might come earlier than I expected.

mesonbuild/compilers/d.py
dcbaker: Actually, the more I think about it, the more I think it would be better to just return a bool and store that. Then you can just do:
```meson
if self._dmd_has_depfile:
   return ['<depfile args>']
return []
```

eli-schwartz: In the Qt module, where we also need to determine whether depfiles are supported I set a class attribute rcc_supports_depfiles, and defined it to True in `__init__` if the version was new enough (otherwise logged a warning that dependencies would be unreliable).

Later in the function that handles rcc, I checked `self.rcc_supports_depfiles`.

So yes, I too would go this way.

dcbaker: I think the GDC code does this for several features as well.

dcbaker (approved):

dcbaker:
Merged, thanks!

rtbo:
Thank you.
Please note that `-makedeps` currently only work when compiling source file to object file (which I think is fine for Meson).
[Another PR](https://github.com/dlang/dmd/pull/12049) is in preparation to extend it to more targets, and to support special characters in filenames.
This one will likely not be before `2.096` though.

e.g. I didn't test, but the issued makefile may be broken if there is a space in the source path. I don't know how Ninja would handle that. Hopefully not by aborting the build.

PR #8122, opened 2020-12-23
Correct format for meson dist checksum and don't print

a-vrma:
The change makes `meson dist` follow the checksum format described at
https://www.gnu.org/software/coreutils/manual/html_node/md5sum-invocation.html#md5sum-invocation
and is compatible with busybox too.

I also removed the print since it doesn't make sense inside a function called
`create_hash`. It also contributes to the unnecessary output when a user runs
`meson dist`. [On my very simple project I get 55 lines of output and a few
warnings](https://0x0.st/iCOM.txt). There are other things we can do to improve
that (like calling `git` with `--quiet`) but this is a start.

I ran `AllPlatformTests.test_dist_git` and
`AllPlatformTests.test_dist_git_script` to test.

### DIFF
--- a/mesonbuild/mdist.py
+++ b/mesonbuild/mdist.py
@@ -48,7 +48,9 @@ def create_hash(fname):
     m = hashlib.sha256()
     m.update(open(fname, 'rb').read())
     with open(hashname, 'w') as f:
-        f.write('{} {}\n'.format(m.hexdigest(), os.path.basename(fname)))
+        # A space and an asterisk because that is the format defined by GNU coreutils
+        # and accepted by busybox and the Perl shasum tool.
+        f.write('{} *{}\n'.format(m.hexdigest(), os.path.basename(fname)))
     print(os.path.relpath(fname), m.hexdigest())
 
 
--- a/mesonbuild/mdist.py
+++ b/mesonbuild/mdist.py
@@ -51,7 +51,6 @@ def create_hash(fname):
         # A space and an asterisk because that is the format defined by GNU coreutils
         # and accepted by busybox and the Perl shasum tool.
         f.write('{} *{}\n'.format(m.hexdigest(), os.path.basename(fname)))
-    print(os.path.relpath(fname), m.hexdigest())
 
 
 def del_gitfiles(dirname):
@@ -284,4 +283,5 @@ def run(options):
     if rc == 0:
         for name in names:
             create_hash(name)
+            print('Created', os.path.relpath(name))
     return rc


mesonbuild/mdist.py
```diff
@@ -48,8 +48,7 @@ def create_hash(fname):
     m = hashlib.sha256()
     m.update(open(fname, 'rb').read())
     with open(hashname, 'w') as f:
-        f.write('{} {}\n'.format(m.hexdigest(), os.path.basename(fname)))
-    print(os.path.relpath(fname), m.hexdigest())
+        f.write('{}  {}\n'.format(m.hexdigest(), os.path.basename(fname)))
```
dcbaker, line 51: Please leave a comment here explaining why there are two spaces here, or it's going to get changed again in the future.

dcbaker (changes requested):
One request for a comment, otherwise this seems fine to me.

eli-schwartz:
The first commit makes sense and is an obvious bugfix. It currently produces invalid checksum files (GNU md5sum nevertheless does not warn on -w).

The second commit is less clear why it is needed...

mesonbuild/mdist.py
a-vrma: Good idea

a-vrma:
Originally, I wanted to reduce the output of `meson dist`
Looking at the message for commit f4d98a3e7, I realize why it was added. I still think printing the hex digest is unnecessary.

eli-schwartz:
It might be unnecessary, but on the other hand it's also not taking up another line. So I don't get the point of removing it.

a-vrma:
Ok I force-pushed changing 3 things:

1. Added a comment.
2. Switched to the 'binary mode' described at
   <https://www.gnu.org/software/coreutils/manual/html_node/md5sum-invocation.html>.
   I don't think there is any difference (tested on MSYS2, busybox-w32,
   GNU/Linux), but it is technically more cross-platform. The difference as
   I understand it is that text mode is supposed to open with mode `r` and
   binary mode is supposed to open with mode `rb`. But again, I couldn't detect
   any difference on Windows and POSIX doesn't differentiate between binary and
   text mode.
3. Kept the printing of the file name but took it out of the `check_hash`
   function.

eli-schwartz (approved):

eli-schwartz:
Note in the commit message you refer to check_hash instead of create_hash.

mesonbuild/mdist.py
```diff
@@ -284,4 +283,5 @@ def run(options):
     if rc == 0:
         for name in names:
             create_hash(name)
+            print('Created', os.path.relpath(name))
```
eli-schwartz, line 286: Printing that it is *created*, is indeed more sensible than printing the bare filename + hash.

a-vrma:
> Note in the commit message you refer to check_hash instead of create_hash.

Fixed now.

eli-schwartz:
Thanks!

PR #8124, opened 2020-12-23
WIP: externalprogram: Add program() method

xclaesse:

PR #8125, opened 2020-12-23
RFC: external-project: Add executable() method

dnicolodi:
Add an `executable()` method constructing an object pointing to a program built as part of the external project build and suitable to be passed to `meson.override_find_program()`. This allows to use external projects to build code generators to be used in other parts of the build.

Note that the program is run from the temporary install location in the build directory, thus it could fail to run if it depends on
resources located at specific absolute filesystem paths. In particular, binaries relying on shared libraries built as part of the external project would fail to load. This could maybe be solved using an appropriate `LD_LIBRARY_PATH` environmental variable, but no effort in this direction is done here.

This is just an RFC as I realize that abusing a `build.Executable` object to hold a reference to the program path is probably not the right thing to do. Documentation and tests will follow once this is resolved.


dnicolodi:
Superseded by #8124.

PR #8135, opened 2020-12-26
cmake: fix missing languages from CMake (fixes #8132)

### LINKED ISSUE #8132: CMake translation missing Objective-C++ files in wxWidgets
Attempting to build and link against wxWidgets on MacOS results in various linker errors. Looking up the symbols, they seem to be coming from the various objective-c++ `.mm` files in `src/osx/cocoa`. Compiling directly with CMake does seem to include those files. 

Unfortunately, I don't have a minimal repro handy. If that's needed, please let me know and I can try to make one. I've uploaded `meson.build`, `fileAPI.json` and `__CMake_build/cmake_trace.txt`, but if there's anything else I can provide I'd be happy to since this is blocking.

[cmake_objcpp_bug_logs.zip](https://github.com/mesonbuild/meson/files/5742882/cmake_objcpp_bug_logs.zip)

CMake module invocation: https://github.com/TypesettingTools/Aegisub/blob/meson-vs2019/meson.build#L125

System:
MacOS 11.1
Python 3.9.1
Meson 0.56.0
CMake 3.19.2

mensinda:


mensinda:
The cygwin error is unrelated since it is caused during the pip install stage...

PR #3998, opened 2018-08-09
Qt: Allow passing dependency objects to preprocess method (#3470, #3875)

equeim:

PR #8139, opened 2020-12-29
Qt: Pass include directories from dependency to moc compiler properly

stream009:
#3998 allows us to propagate project's dependency to the moc compiler, but it somehow is incomplete. It passes compile definitions properly but it passes include directories only if 

1. dependency has include directories in result of get_compile_args() which is true for some object (ex. Qt5Dependency) but false for some (ex. InternalDependency)
2. include directory is specified with '-I' (not -isystem)

This patch completes flow described above.

### DIFF
--- a/mesonbuild/modules/qt.py
+++ b/mesonbuild/modules/qt.py
@@ -188,9 +188,22 @@ class QtBaseModule(ExtensionModule):
         compile_args = []
         for dep in unholder(dependencies):
             if isinstance(dep, Dependency):
+
+                if hasattr(dep, 'include_directories'):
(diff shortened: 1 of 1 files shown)


jpakkane:
Please don't have have merges in a pull request, use rebase instead: https://mesonbuild.com/Contributing.html#keeping-pull-requests-up-to-date

jpakkane:
Ping. You'll need to remove the merge commit before this can be considered. Thanks.

eli-schwartz:
:crossed_swords: conflicts

Is this going to be rebased to not include merge commits?

PR #6030, opened 2019-10-11
Remove duplicated object files in static libraries

xclaesse:
When a static library link_whole to a bunch of other static libraries,
we have to extract all their objects recursively. But that could
introduce duplicated objects. ar is dumb enough to allow this without
error, but once the resulting static library is linked into an
executable or shared library, the linker will complain about duplicated
symbols.

PR #8151, opened 2021-01-04
transitively pass compile flags for internal dependency

bonzini:
Currently, Meson passes link arguments from both direct and indirect indirect dependencies.  However, only direct dependencies are used for compiler arguments.  The common case is a header file that includes headers from an external source, where those headers require a -I option to be found.
    
This was reported for QEMU as https://bugs.launchpad.net/qemu/+bug/1909256.

### DIFF
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -677,11 +677,12 @@ class Backend:
 
         return extra_args
 
-    def get_target_external_deps(self, target):
+    def get_target_external_deps(self, target, static_only):
         deps = target.get_external_deps()
         transitive_deps = OrderedSet()
         for d in target.get_dependencies():
-            if isinstance(d, build.StaticLibrary):
+            if isinstance(d, build.StaticLibrary) or \
+               (not static_only and isinstance(d, build.SharedLibrary)):
                 transitive_deps.update(d.get_external_deps())
         if transitive_deps:
             deps += list(transitive_deps)
@@ -744,7 +745,7 @@ class Backend:
         # added while generating the link command.
         # NOTE: We must preserve the order in which external deps are
         # specified, so we reverse the list before iterating over it.
-        for dep in reversed(target.get_external_deps()):
+        for dep in reversed(self.get_target_external_deps(target, static_only=False)):
             if not dep.found():
                 continue
 
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -677,6 +677,16 @@ class Backend:
 
         return extra_args
 
+    def get_target_external_deps(self, target):
+        deps = target.get_external_deps()
+        transitive_deps = OrderedSet()
+        for d in target.get_dependencies():
+            if isinstance(d, build.StaticLibrary):
+                transitive_deps.update(d.get_external_deps())
+        if transitive_deps:
+            deps += list(transitive_deps)
+        return deps
+
     def generate_basic_compiler_args(self, target: build.BuildTarget, compiler: 'Compiler', no_warn_args: bool = False) -> 'CompilerArgs':
         # Create an empty commands list, and start adding arguments from
         # various sources in the order in which they must override each other
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -2915,14 +2915,10 @@ https://gcc.gnu.org/bugzilla/show_bug.cgi?id=47485'''))
 
             commands += linker.get_target_link_args(target)
             # External deps must be last because target link libraries may depend on them.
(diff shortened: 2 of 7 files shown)


jpakkane:
The fact that header flags are not passed transitively but libraries are (in some cases) is intentional. Otherwise compiler flag counts explode in deep hierarchies. Because of this include paths must be exported manually, typically by adding the appropriate bits to a `declare_dependency`.

Libs are a bit stupid, because you need to add direct dependencies if, for example, you link to a static library.

bonzini:
I tried using the `declare_dependency` method, it works with `link_with` but not with `link_whole`:

```
project('Transitive link_whole', 'c')
libfoo = static_library('foo', 'foo.c')
foo = declare_dependency(link_whole: libfoo)

libbar = static_library('bar', 'bar.c', dependencies: foo)

# works
bar = declare_dependency(link_with: libbar, dependencies: foo)
executable('prog-with', sources: files('prog.c'), dependencies: [foo, bar])

# fails
bar = declare_dependency(link_whole: libbar, dependencies: foo)
executable('prog-whole', sources: files('prog.c'), dependencies: [foo, bar])
```

because the `bar` dependency brings in (via `libbar.a`) all the files linked into `libfoo.a`. The most elegant solution would be not to use whole-archive at all, and just include all the objects directly (removing duplicates). I'm not sure how easy it is, and whether you agree with the judgment.

roolebo:
@jpakkane This is somewhat similar to #6030. Please see the test case: https://github.com/roolebo/meson/commit/11fa339862c339158e980b73731f7dfbb8862398

roolebo:
Well, a fix for this could be:
```
diff --git a/mesonbuild/backend/ninjabackend.py b/mesonbuild/backend/ninjabackend.py
index b847c2cba..549c65ac2 100644
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -862,7 +862,8 @@ int dummy;
                     compiled_sources.append(s)
                     source2object[s] = o

-        obj_list += self.flatten_object_list(target)
+        if not isinstance(target, build.StaticLibrary):
+            obj_list += self.flatten_object_list(target)
         if is_unity:
             for src in self.generate_unity_files(target, unity_src):
                 o, s = self.generate_single_compile(target, src, True, unity_deps + header_deps
```

bonzini:
That seems like a weird patch. Is the static library complete with the patch? Note that having link_whole'd sub-libraries included in the static libraries themselves is fine. The problem is that this results in duplicate objects _when the libraries are linked in the executable_.

I hope to find some time for it next week.

roolebo:
> The problem is that this results in duplicate objects when the libraries are linked in the executable.

And the patch fixes exactly this issue. I added prints just before and after `obj_list += self.flatten_object_list(target)` and got:
```
['liba.a.p/subsystem-a_a.c.o']
['liba.a.p/subsystem-a_a.c.o']
['libb.a.p/subsystem-b_b.c.o']
['libb.a.p/subsystem-b_b.c.o', 'liba.a.p/subsystem-a_a.c.o']
['prog.p/prog.c.o']
['prog.p/prog.c.o']
```

I'm not claiming this is the best approach but it's something that resolves this particular test case.

bonzini:
Your patch is changing things for static libraries, not for executables. Are the link_whole'd sub-libraries still included in the super-library? If not, this still works in the case of QEMU (and by extension your testcase, which is distilled from QEMU) but not in general.

roolebo:
@bonzini Ok. I hope I see your point now. You say that super-library is a valid thing and shouldn't be broken but executable should flatten all objects from all libraries including super-libraries. Does it mean that we need to use object files from the static libraries directly in the executable?

But then I wonder why do we need static library as intermediate step? I'm sorry I can't get forget CMake approaches :)

bonzini:
> Does it mean that we need to use object files from the static libraries directly in the executable?

Yes. Not coincidentially this is what QEMU's old makefiles also did. Meson already has the necessary code in order to support `extract_all_objects()`.

> But then I wonder why do we need static library as intermediate step?

Because in QEMU the various executables (emulators, tools or tests) only need a subset of the static libraries that we build (authz, crypto, io, qom). So each subsystem is built into its own static library

roolebo:
@bonzini Thanks! I've seen `extract_all_objects()`. I'll try to come up with something.

> executables (emulators, tools or tests) only need a subset of the static libraries

My point was that CMake has special `OBJECT` libraries exactly for this: https://lists.gnu.org/archive/html/qemu-devel/2021-01/msg00989.html. Then there's no need to create archive with ar for unneeded static libraries that aren't going to be used/installed directly. That's why I saw it a bit different :) If meson had object_library target, one could implement libraries which are just a group of object files related to a subsystem.

PR #8158, opened 2021-01-05
rust: Accept generated sources for main.rs

### LINKED ISSUE #8157: Failure when using CustomTarget with a Rust executable
**Describe the bug**

Using a CustomTarget as the only source of a Rust executable makes meson fail.

**To Reproduce**

The self-contained `meson.build` is:

```
project('proj', 'rust')

main_rs = custom_target(
  'main_rs',
  output: 'main.rs',
  capture: true,
  command: ['echo',
            'fn main() { println!("hi") }'
	   ],
)

executable('main', [main_rs])
```

Then fails at `meson build` with the following stack trace:

```
  File "/usr/lib/python3.8/site-packages/mesonbuild/mesonmain.py", line 140, in run
  File "/usr/lib/python3.8/site-packages/mesonbuild/msetup.py", line 253, in run
    app.generate()
  File "/usr/lib/python3.8/site-packages/mesonbuild/msetup.py", line 161, in generate
    self._generate(env)
  File "/usr/lib/python3.8/site-packages/mesonbuild/msetup.py", line 223, in _generate
    intr.backend.generate()
  File "/usr/lib/python3.8/site-packages/mesonbuild/backend/ninjabackend.py", line 519, in generate
    self.generate_target(t)
  File "/usr/lib/python3.8/site-packages/mesonbuild/backend/ninjabackend.py", line 842, in generate_target
    linker, stdlib_args = self.determine_linker_and_stdlib_args(target)
  File "/usr/lib/python3.8/site-packages/mesonbuild/backend/backends.py", line 465, in determine_linker_and_stdlib_args
    l, stdlib_args = target.get_clink_dynamic_linker_and_stdlibs()
  File "/usr/lib/python3.8/site-packages/mesonbuild/build.py", line 1325, in get_clink_dynamic_linker_and_stdlibs
    raise AssertionError(m.format(self.name))
AssertionError: Could not get a dynamic linker for build target 'main'
```

**Expected behavior**

Expected success, and when executing `ninja -C build` an executable called `main` would be created.

**Notes**

After digging a bit it seems that generated sources are not taken into account by neither `is_rust_target` nor `generate_rust_target` in the ninjabackend. Failing to identify this as a rust target seems to be the main problem here.

**System parameters**

* Linux
* Meson 0.56.0
* Ninja 1.10.2

dcbaker:
This has lots of caveats, namely that rustc expects all of the sources
to exist in a single tree, not split across two (a build and source)
like meson does. You can use a generated source, but in that case it
must be exclusvively generated sources, in the same directory.

Fixes #8157

### DIFF
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -45,7 +45,7 @@ from ..mesonlib import (
 )
 from ..mesonlib import get_compiler_for_source, has_path_sep, OptionKey
 from .backends import CleanTrees
-from ..build import InvalidArguments
+from ..build import GeneratedList, InvalidArguments
 from ..interpreter import Interpreter
 
 if T.TYPE_CHECKING:
@@ -1581,12 +1581,23 @@ int dummy;
         args = rustc.compiler_args()
         # Compiler args for compiling this target
         args += compilers.get_base_compile_args(base_proxy, rustc)
+        self.generate_generator_list_rules(target)
         main_rust_file = None
         for i in target.get_sources():
             if not rustc.can_compile(i):
                 raise InvalidArguments('Rust target {} contains a non-rust source file.'.format(target.get_basename()))
             if main_rust_file is None:
                 main_rust_file = i.rel_to_builddir(self.build_to_src)
+        for g in target.get_generated_sources():
+            for i in g.get_outputs():
+                if not rustc.can_compile(i):
+                    raise InvalidArguments('Rust target {} contains a non-rust source file.'.format(target.get_basename()))
+                if isinstance(g, GeneratedList):
+                    fname = os.path.join(self.get_target_private_dir(target), i)
+                else:
+                    fname = i
+                if main_rust_file is None:
+                    main_rust_file = fname
         if main_rust_file is None:
             raise RuntimeError('A Rust target has no Rust sources. This is weird. Also a bug. Please report')
         target_name = os.path.join(target.subdir, target.get_filename())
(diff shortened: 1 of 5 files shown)


cmarcelo:
Couple of suggestions:

- Could meson fail when there's a mix of regular `sources` and `generated` with the message about not supporting that yet?

- This could also be documented in the Manual. There's a section for D, Java and Vala already there, having one of Rust could be a good place.

dcbaker:
It certainly could, but it would require almost as much work as just fixing it. We basically treat all rust files after the main one like c/C++ .h files, they're ignored.

dcbaker:
We really need to fix the generated sources problem anyway, and it's on my list of things todo.

dcbaker:
Also noticed that we weren't creating an order dependency, which is a problem.

jpakkane:
Test failure seems relevant:

```
The Meson build system
Version: 0.56.99
Source dir: D:\a\1\s\test cases\rust\11 generated main
Build dir: D:\a\1\s\b 482357f53b
Build type: native build
Project name: generated rust main
Project version: undefined
Rust compiler for the host machine: rustc -C linker=link (rustc 1.48.0)
Rust linker for the host machine: link link 14.28.29335.0
Rust compiler for the build machine: rustc -C linker=link (rustc 1.48.0)
Rust linker for the build machine: link link 14.28.29335.0
Build machine cpu: x86_64
Host machine cpu: x86_64
Target machine cpu family: x86_64
    compiler, _ = self.get_clink_dynamic_linker_and_stdlibs()
  File "D:\a\1\s\mesonbuild\build.py", line 1392, in get_clink_dynamic_linker_and_stdlibs
    raise AssertionError(m.format(self.name))
AssertionError: Could not get a dynamic linker for build target 'custom_target_main'
```

dcbaker:
Gah, we have two different functions for determining rust sources, one in the `BuildTarget` and one in the `NinjaBackend`! Must fix

dcbaker:
@jpakkane, is this okay to merge now?

PR #8159, opened 2021-01-06
Read and store all environment variables up front

dcbaker:
Currently, meson reads environment variables in various places, sometimes at configure time, sometimes at runtime, with various behaviors. This series seeks to unify all environment variable handling into a single place, and store them so that they persist across reconfigures. There are a lot of cleanups in here as well, but the end result I think is much nicer than what we had before (and is less code).


jpakkane:
Good stuff.

PR #8162, opened 2021-01-06
Add a wrapper to the rust module for bindgen

dcbaker:
It's pretty common to want to use bindgen with rust and meson, as bindgen makes writing unsafe wrappers for C libraries really easy. It's uses clang internally, and users need to pass include directories via `-I` arguments. Meson generally discourages this, and prefers `include_directories` instead. This series adds a bindgen wrapper to the rust module, which simplifies this process by simplifying the passing of arguments into bindgen, by correctly handling `include_directories` objects, and by automatically generating and using a depfile.

This is currently based on #8236

### DIFF
--- a/docs/markdown/Rust-module.md
+++ b/docs/markdown/Rust-module.md
@@ -3,7 +3,7 @@ short-description: Rust language integration module
 authors:
     - name: Dylan Baker
       email: dylan@pnwbakers.com
-      years: [2020]
+      years: [2020, 2021]
 ...
 
 # Unstable Rust module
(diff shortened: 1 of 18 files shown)


dcbaker:
@cmarcelo, @karolherbst, I know both of you have used bindgen in meson, does this look like a reasonable interface?

docs/markdown/Rust-module.md
```diff
@@ -33,3 +33,31 @@ that automatically.
...
+
+```meson
+rust = import('unstable-rust')
+
+inc = include_directories('..'¸ '../../foo')
+
+generated = rust.bindgen(
+    'myheader.h',
+    'generated.rs',
+    include_directories : [inc, include_directories('foo'),
```
cmarcelo, line 59: Question: In my use case, the C header also depends on generated headers (i.e. their directory is inside build dir). What would be a way to add them here?

Otherwise, the interface described here looks fine to me.

docs/markdown/Rust-module.md
```diff
@@ -33,3 +33,31 @@ that automatically.
...
+    args : ['--no-rustfmt-bindings'],
+    c_args : ['-DFOO=1'],
+)
```
karolherbst, line 62: I think some also might want to switch between static and shared library.

karolherbst:
Yeah.. checked how I used it and I think the interface should be fine.

docs/markdown/Rust-module.md
dcbaker: You're right, we do need to handle generated headers.

dcbaker:
I've updated this to handle generated headers, as well as multiple generated headers, and to generate a depfile, so using this is more reliable than hand rolling it! The API has changed slightly from:
```meson
rust.bindgen(
  'input.h',
  'output.rs'
  ...
)
```
to
```meson
rust.bindgen(
  input : [input_h, depends_h],
  output : 'output.rs'
  ...
)
```

jljusten:
@dcbaker: I was able to use this patchset to use bindgen to interface to a C library. I was able to call C functions from rust, and the C function had a callback function pointer in a structure. In my test, the callback target function was written in rust, and the C code successfully called back into the rust function.

dcbaker:
The Arch and Fedora CI images are failing because hotdoc isn't being properly setup now. The Windows CI fail is because there is a bindgen in path, but not clangd. The Eoan is the really annoying one, it seems to have been removed from the Ubuntu repos, so I guess we need to completely replace it with something else?

jpakkane:
That should probably be the latest non-LTS Ubuntu release, so Groovy instead of Eoan.

dcbaker:
I'm sorta off today, so I'll get around to wrangling groovy tomorrow and get this ready to merge properly. I guess we can merge the other one today if you like, and I'll get the rest of this sorted out tomorrow. It looks like from reading on azure there is a clangd somewhere, we just need to point an env var at it.

dcbaker:
@jpakkane I'd really like to get this into 0.57, we've got people wanting to use it in mesa, but I've got more unrelated ci problems

PR #8163, opened 2021-01-07
Cope with initial Popen_safe() throwing in _guess_win_linker() such that an explicit ld is used.

lukester1975:
Hello

I've been trying to use clang-cl with explicit environment variables pointing at my binaries; e.g.:

```
CC=/c/tools/msys64/mingw64/bin/clang-cl.exe
CXX=/c/tools/msys64/mingw64/bin/clang-cl.exe
C_LD=/c/tools/msys64/mingw64/bin/lld-link.exe
CXX_LD=/c/tools/msys64/mingw64/bin/lld-link.exe
```

#8105 is required to get the compiler detected correctly; that's fine.

However, linker detection also fails because there is no exception handling in the initial `Popen_safe()` check (for `lld-link.exe`) in `_guess_win_linker()` such that the second  `Popen_safe()` using my explicit linker is never run.

Don't know if it's the right answer, but here's my quick workaround if it's of any use!

Regards

### DIFF
--- a/mesonbuild/environment.py
+++ b/mesonbuild/environment.py
@@ -950,16 +950,19 @@ class Environment:
         if extra_args is not None:
             check_args.extend(extra_args)
 
-        p, o, _ = Popen_safe(compiler + check_args)
-        if o.startswith('LLD'):
-            if '(compatible with GNU linkers)' in o:
-                return LLVMDynamicLinker(
-                    compiler, for_machine, comp_class.LINKER_PREFIX,
-                    override, version=search_version(o))
-            elif not invoked_directly:
-                return ClangClDynamicLinker(
-                    for_machine, override, exelist=compiler, prefix=comp_class.LINKER_PREFIX,
-                    version=search_version(o), direct=False, machine=None)
+        try:
+            p, o, _ = Popen_safe(compiler + check_args)
+            if o.startswith('LLD'):
+                if '(compatible with GNU linkers)' in o:
+                    return LLVMDynamicLinker(
+                        compiler, for_machine, comp_class.LINKER_PREFIX,
+                        override, version=search_version(o))
+                elif not invoked_directly:
+                    return ClangClDynamicLinker(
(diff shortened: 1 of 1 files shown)


e820:
Unless I'm mistaken the actual fix would be to adjust this:

https://github.com/mesonbuild/meson/blob/4ff020759a3cdedb9d8ff1150ec09313096f6228/mesonbuild/environment.py#L1238

as it currently assumes that whatever `lld-link` instance you want to use resides in `PATH`.

lukester1975:
Personally I have no idea, I've only just started evaluating meson and trying to figure out how to get it to do what I want :)

I only spent a few hours looking at the code, but that first `Popen_safe` in `_guess_win_linker` remained very confusing to me (aside from the concept of trying to guess a linker when one is explicitly provided!). I wasn't sure what it was trying to accomplish but since if it throws then an explicit linker is never tried, wrapping it in a try was an easy "fix" for me.

Should I just open an issue instead?

(BTW once this is corrected, I think there's also a problem with building and running the sanity check: it does a compile+link in one step, which is ignoring an explicit linker. Also it looks like `clang-cl.exe` will use `link.exe` as the linker, and if that's not in the path...)

e820:
> I only spent a few hours looking at the code, but that first Popen_safe in _guess_win_linker remained very confusing to me (aside from the concept of trying to guess a linker when one is explicitly provided!). I wasn't sure what it was trying to accomplish but since if it throws then an explicit linker is never tried, wrapping it in a try was an easy "fix" for me.

There are multiple paths into the Windows linker detection logic based on whether your compiler is able to drive the selected linker or not - furthermore we have to differentiate between different kinds of `link`-style linkers.

> Should I just open an issue instead?

For now that might be the best idea, as far as I know @dcbaker is looking into a rewrite of the entire linker detection logic which is where we'd probably want to address all of the nits in the current implementation. That said I'm not sure how pressing this issue is so a bandaid fix might be just fine.

Issue #8165, opened 2021-01-07
dependency() on a macOS framework cannot be found while cross-compiling.


UncombedCoconut:
**Describe the bug**
I'm attempting to build a Meson project in an "[osxcross](https://github.com/tpoechtrager/osxcross)" setting (build system: Linux; host system: Darwin).
I've hit what looks like an obvious mistake in Meson's programming, so I'd like to run that by folks before providing a complete reproduction (see below). (However, depending on what Meson needs, I'll see whether it's possible to get a reduced test case.)

The meson.build file has a dependency() on Apple's "Foundation" framework, among others.
In the cross environment, what's reported is: `ERROR: Dependency "Foundation" not found, tried pkgconfig"`
On an actual Mac, the log shows Meson doing something reasonable:
```
Finding framework path by running:  cc -v -E - -pipe

Looking for framework physfs in /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/System/Library/Frameworks
```
Meson has a cross file available, so the same strategy ought to work for my osxcross build, but it doesn't bother.
The code responsible for this decision is one of these snippets in `mesonbuild/dependencies/base.py`
```
        # On OSX, also try framework dependency detector  
        if mesonlib.is_osx():
            candidates.append(functools.partial(ExtraFrameworkDependency, name, env, kwargs))
```
The implementation in `mesonbuild/mesonlib.py` is
```
def is_osx() -> bool:
    return platform.system().lower() == 'darwin'
```
As far as I can tell, it's impossible for this to succeed in a cross-compilation environment, because it's checking the **build's** platform rather than the host's.
I am able to get past this error by editing these scripts in `/usr/local/lib/python3.9/dist-packages/mesonbuild/`.

**To Reproduce**
So far, my answer is nasty, along the lines of "download an enormous Docker image, clone a 1GB git repo into it, and run Meson setup."
I'm sorry.
What are your needs in terms of reduced test cases?

**Expected behavior**
I would expect Meson to perform the same search "ExtraFrameworkDependency" search anytime the host system makes it appropriate to do so.

**system parameters**
* Is this a [cross build](https://mesonbuild.com/Cross-compilation.html) or just a plain native build (for the same computer)?
Yes.
* what operating system (e.g. MacOS Catalina, Windows 10, CentOS 8.0, Ubuntu 18.04, etc.)
Ubuntu 20.04 (Docker image).
* what Python version are you using e.g. 3.8.0
3.9.1
[...]

PR #8166, opened 2021-01-07
Look for macos Framework dependencies when cross compiling

dcbaker:
We currently use `mesonlib.is_osx()`, this isn't correct, as this is for the *build* machine, but we need to check for the machine that we are actually building for. This means that we can try frameworks when we shouldn't (cross compiling from osx -> !osx) and not when we should (cross compiling from !osx -> osx).

During this I added some type annotations, which caused mypy to notice some incorrect annotations in the boost module, as well as a legitimate bug (passing a method instead of the return value of that method)

Fixes # #8165


jpakkane:
:crossed_swords: conflicts.

dcbaker:
Ah, I love it when I fix the same bug in multiple branches, lol.

PR #8176, opened 2021-01-10
clang-cl with VS backend debug builds not very debuggable

lukester1975:
Looks like setting FavorSizeOrSpeed even with `/Od` means (some) optimization is taking place, making debugging problematic.

This is what I'm using locally to avoid setting FavorSizeOrSpeed with a debug build. Obviously my mod is not distinguishing between cl or clang-cl toolsets, but I suspect that's no bad thing!

Regards

Luke.

### DIFF
--- a/mesonbuild/backend/vs2010backend.py
+++ b/mesonbuild/backend/vs2010backend.py
@@ -1159,7 +1159,8 @@ class Vs2010Backend(backends.Backend):
         # Size-preserving flags
         if '/Os' in o_flags:
             ET.SubElement(clconf, 'FavorSizeOrSpeed').text = 'Size'
-        else:
+        # Note: setting FavorSizeOrSpeed with clang-cl conflicts with /Od and can make debugging difficult, so don't.
+        elif '/Od' not in o_flags:
             ET.SubElement(clconf, 'FavorSizeOrSpeed').text = 'Speed'
         # Note: SuppressStartupBanner is /NOLOGO and is 'true' by default
         self.generate_lang_standard_info(file_args, clconf)


mesonbuild/backend/vs2010backend.py
```diff
@@ -1071,7 +1071,8 @@ def gen_vcxproj(self, target, ofname, guid):
         # Size-preserving flags
         if '/Os' in o_flags:
             ET.SubElement(clconf, 'FavorSizeOrSpeed').text = 'Size'
-        else:
+        # Note: setting FavorSizeOrSpeed with clang-cl overrides /Od and makes debugging difficult.
+        elif not '/Od' in o_flags:
```
jpakkane, line 1075: This will potentially change in #8179 as we won't set `/Od` by default any more. How will it affect this change?

Also, the documentation says that there are three choices: https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.vcprojectengine.favorsizeorspeedoption?view=visualstudiosdk-2019

MSDN docs do not seem to say if you should have the value `none` in that case or is a missing value interpreted as `none`.

lukester1975: Doesn't seem that #8179 affects it. For `--buildtype=debug`:

| | 0.56.2 | marstaik:no_ZI | lukester1975:fix-clang-cl-vs-backend |
|-|-|-|-|
FavorSizeOrSpeed | Speed | Speed | not present |
BasicRuntimeChecks | EnableFastChecks | not present | EnableFastChecks |
Optimization | Disabled | not present | Disabled |

For the second point: a default "Hello World" vcxproj from VS doesn't have `FavorSizeOrSpeed` at all (i.e. debug or release). So missing is the same as `none`, I guess (though the IDE's wording is better - `neither`!).

lukester1975: FYI, using ninja there is no `/Ot` for a debug build:

```
"/MDd" "/nologo" "/showIncludes" "/W2" "/EHsc" "/std:c++14" "/permissive-" "/Zi" "/Ob0" 
"/Od" "/RTC1" "/Od" "/Ob0" "/Zi" "/arch:AVX"
```

VS backend:
```
/GS /W2 /Gy /Zi /Od /fp:precise /D "_MBCS" /WX- /Gd /MDd /EHsc /nologo /Ot
/diagnost "/EHsc" "/std:c++14" "/permissive-" "/arch:AVX" 
```

(ignoring include paths, pdbs and other intermediates).

For this particular example, the differences boil down to:

VS: `/GS /Gy /fp:precise /D "_MBCS" /WX- /Gd /Ot`

Ninja: `"/Ob0" "/RTC1" "/Zi"`

lukester1975:
Anything I can do to help get this merged in? I've rebased the branch.

In summary:

- Meson's VS backend is setting `/Ot` for debug builds
- Ninja backend is not
- `/Ot` with clang-cl is optimizing to some degree, despite `/Od` also being set
- Manifests as some stack frames being inlined and "unbreakpointable" and some variables being optimized away completey, hence unwatchable
- A VS created project is not setting `/Ot` for debug (no `FavorSizeOrSpeed` present at all)
- Still all true as of VS2022 with clang-cl 12

Many thanks

lukester1975:
Pretty please? I can pay 400 quatloos!

mesonbuild/backend/vs2010backend.py
```diff
@@ -1159,7 +1159,8 @@ def gen_vcxproj(self, target, ofname, guid):
         # Size-preserving flags
         if '/Os' in o_flags:
             ET.SubElement(clconf, 'FavorSizeOrSpeed').text = 'Size'
-        else:
+        # Note: setting FavorSizeOrSpeed with clang-cl conflicts with /Od and can make debugging difficult, so don't.
+        elif not '/Od' in o_flags:
```
eli-schwartz, line 1163: CI fails here because pylint dislikes the coding style.

```suggestion
        elif '/Od' not in o_flags:
```

eli-schwartz (changes requested):
This change seems to make sense, and it seems to be correct regardless of whatever some other PR potentially does... unsure why it has taken this long.

I do have one small nit to make CI pass.

mesonbuild/backend/vs2010backend.py
lukester1975: Cool, done.

Thanks

Issue #5183, opened 2019-03-31
MSVC does not always support /ZI, leading to noise in the build process (D9038) in these cases with --buildtype=debug


andlabs:
Right now `--buildtype=debug` uses `/ZI` with MSVC, meaning to enable PDB files with Link and Continue. However, this is not supported on all configurations, and `cl` will warn about that and treat the flag as if it was `/Zi` instead:

```
Z:\src\github.com\andlabs\libui>meson setup build
The Meson build system
Version: 0.50.0
Source dir: Z:\src\github.com\andlabs\libui
Build dir: Z:\src\github.com\andlabs\libui\build
Build type: native build
Project name: libui
Project version: undefined
Native C compiler: cl (msvc 18.00.30723)
Native C++ compiler: cl (msvc 18.00.30723)
Build machine cpu family: x86_64
Build machine cpu: x86_64
Windows resource compiler: Microsoft (R) Windows (R) Resource Compiler Version 6.3.9600.17200
Library user32 found: YES
Library kernel32 found: YES
Library gdi32 found: YES
Library comctl32 found: YES
Library uxtheme found: YES
Library msimg32 found: YES
Library comdlg32 found: YES
Library d2d1 found: YES
Library dwrite found: YES
Library ole32 found: YES
Library oleaut32 found: YES
Library oleacc found: YES
Library uuid found: YES
Library windowscodecs found: YES
Dependency threads found: YES
Build targets in project: 11
Found ninja-1.8.2 at 'C:\Program Files\Meson\ninja.EXE'
WARNING: This version of MSVC does not support cpp_std arguments
WARNING: This version of MSVC does not support cpp_std arguments

Z:\src\github.com\andlabs\libui>ninja --version
1.8.2

Z:\src\github.com\andlabs\libui>ninja -C build
ninja: Entering directory `build'
[1/76] Compiling C object ui@sha/common_attrlist.c.obj.
cl : Command line warning D9038 : /ZI is not supported on this platform; enabling /Zi instead
[2/76] Compiling C object ui@sha/common_attribute.c.obj.
cl : Command line warning D9038 : /ZI is not supported on this platform; enabling /Zi instead
(snipped; imagine this pattern repeating, interspersed with the various actual compiler warnings)
```

In this case, this is Visual Studio 2013 on 64-bit Windows 7 in the x64 Native Tools command line prompt.

[...]

PR #8179, opened 2021-01-10
Replacement of /ZI with /Zi on MSVC Debug

marstaik:
The /ZI flag adds in "Edit and Continue" debug information, which will cause massive slowdowns. It is not a flag that should be added by default to debug builds. It is also incompatible with MSVC LTO.

/Zi will still be added.

### DIFF
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -709,15 +709,16 @@ class Backend:
         commands += compiler.get_buildtype_args(self.get_option_for_target(OptionKey('buildtype'), target))
         commands += compiler.get_optimization_args(self.get_option_for_target(OptionKey('optimization'), target))
         commands += compiler.get_debug_args(self.get_option_for_target(OptionKey('debug'), target))
-        # MSVC debug builds have /ZI argument by default and /Zi is added with debug flag
-        # /ZI needs to be removed in that case to avoid cl's warning to that effect (D9025 : overriding '/ZI' with '/Zi')
-        if ('/ZI' in commands) and ('/Zi' in commands):
-            commands.remove('/Zi')
         # Add compile args added using add_project_arguments()
         commands += self.build.get_project_args(compiler, target.subproject, target.for_machine)
         # Add compile args added using add_global_arguments()
         # These override per-project arguments
         commands += self.build.get_global_args(compiler, target.for_machine)
+        # Using both /ZI and /Zi at the same times produces a compiler warning.
+        # We do not add /ZI by default. If it is being used it is because the user has explicitly enabled it.
+        # /ZI needs to be removed in that case to avoid cl's warning to that effect (D9025 : overriding '/ZI' with '/Zi')
+        if ('/ZI' in commands) and ('/Zi' in commands):
+            commands.remove('/Zi')
         # Compile args added from the env: CFLAGS/CXXFLAGS, etc, or the cross
         # file. We want these to override all the defaults, but not the
         # per-target compile args.
--- a/docs/markdown/snippets/msvc_no_zi.md
+++ b/docs/markdown/snippets/msvc_no_zi.md
@@ -1,0 +1,14 @@
+## "Edit and continue" (/ZI) is no longer used by default for Visual Studio
+
+Meson was adding the `/ZI` compiler argument as an argument for Visual Studio 
+in debug mode. This enables the `edit-and-continue` debugging in 
+Visual Studio IDE's.
+
+Unfortunately, it is also extremely expensive and breaks certain use cases such 
+as link time code generation. Edit and continue can be enabled by manually by 
+adding `/ZI` to compiler arguments.
+
+The `/ZI` argument has now been replaced by the `/Zi` argument for debug builds.
+
+If this is an important issue for you and would like a builtin toggle option, 
+please file an issue in the Meson bug tracker.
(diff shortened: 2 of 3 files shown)


jpakkane:
Still, this is an important feature for many people and we should have a way to toggle it on.

marstaik:
> 
> 
> Still, this is an important feature for many people and we should have a way to toggle it on.

I would assume if someone wants to add /ZI it would be added as a project level argument, or global level argument.
Its a very specific flag.

jpakkane:
FTR the flag was originally added [in this commit](https://github.com/mesonbuild/meson/commit/d23e59ed5f82a3f2d3b0d70faf71a6d4ca16eaaa). It was originally `/Zi`. There is no explanation on why that was changed in that commit, though. I probably copied it from somewhere, such as CMake's default compiler argument list.

jpakkane:
There should be at least `/Zi` in the list, because that is needed to generate debug info.

marstaik:
> 
> 
> There should be at least `/Zi` in the list, because that is needed to generate debug info.

/Zi is added in the debug arg list in:
https://github.com/mesonbuild/meson/blob/8f1d9bb7b0cca265276c46caf928c047870b34f1/mesonbuild/compilers/mixins/visualstudio.py#L90-L93

Unless for some reason that list is irrelevant

marstaik:
Relevant: https://github.com/mesonbuild/meson/issues/5183

vvainola:
If ZI is intended to be applied with global/project arguments, this part in backends.py should be removed or moved later after adding global/project arguments

    # MSVC debug builds have /ZI argument by default and /Zi is added with debug flag
    # /ZI needs to be removed in that case to avoid cl's warning to that effect (D9025 : overriding '/ZI' with '/Zi')
    if ('/ZI' in commands) and ('/Zi' in commands):
        commands.remove('/Zi')

jpakkane:
I tested this branch on MSVC and `/Zi` is used correctly. Thus this approach is doable though it needs the order change mentioned by vvainola above.  In addition there should a release note snippet about this change, something like:

```
## "Edit and continue" is no longer used by default for Visual Studio

Meson has added the `/ZI` compiler argument in debug mode for Visual Studio. This enables the `edit-and-continue` deguging in the Visual Studio IDE. Unfortunately it is also extremely expensive and breaks certain use cases such as link time code generation. Edit and continue can be enabled by adding `/ZI` to compiler arguments. If this is an important issue for you and would like a builtin toggle option, please file an issue in the Meson bug tracker.
```

Finally I'm going to remove this from the 0.56.3 milestone as point releases are only for regressions. This is a breaking behavioural change so it can't go in a point release.

mesonbuild/backend/backends.py
```diff
@@ -709,15 +709,15 @@ def generate_basic_compiler_args(self, target: build.BuildTarget, compiler: 'Com
...
-        # /ZI needs to be removed in that case to avoid cl's warning to that effect (D9025 : overriding '/ZI' with '/Zi')
-        if ('/ZI' in commands) and ('/Zi' in commands):
-            commands.remove('/Zi')
         # Add compile args added using add_project_arguments()
         commands += self.build.get_project_args(compiler, target.subproject, target.for_machine)
         # Add compile args added using add_global_arguments()
         # These override per-project arguments
         commands += self.build.get_global_args(compiler, target.for_machine)
+        # MSVC debug builds have /ZI argument by default and /Zi is added with debug flag
+        # /ZI needs to be removed in that case to avoid cl's warning to that effect (D9025 : overriding '/ZI' with '/Zi')
```
jpakkane, line 718: This comment is no longer up to date as `/ZI` is not added by default. Instead the comment should say something like "Using both /ZI and /Zi at the same times produces a compiler warning, We don't add /ZI by default so if it is being used it is because the user has explicitly enabled it so it should override /Zi.".

marstaik: Done

docs/markdown/snippets/msvc_argument_changes.md
```diff
@@ -0,0 +1,20 @@
...
+* Clang-CL `debug` buildtype arguments now match MSVC arguments
+* There is now no difference between `buildtype` flags and `debug` + `optimization` flags
+
+Meson was adding the `/ZI` compiler argument as an argument for Visual Studio in debug mode. This enables the `edit-and-continue` debugging in Visual Studio IDE's.
+Unfortunately, it is also extremely expensive and breaks certain use cases such as link time code generation. Edit and continue can be enabled by manually by adding `/ZI` to compiler arguments. The `/ZI` argument has now been replaced by the `/Zi` argument for debug builds.
+
+The /Od flag has been removed, as it is already the default in the MSVC compilers, and conflicts with other user options.
+
+/RTC1 conflicts with other RTC argument types as there are many different options, and has been removed by default.
+Run Time Checking can be enabled by manually adding `/RTC1` or other RTC flags of your choice.
```
jpakkane, line 15: Changing the scope of a MR and adding new functionality after review is not really nice. Changes like these should be done as separate MRs or, at the very least, be noticeably mentioned in the discussion page. Please restore `/RTC1` and let's do that in a separate MR properly. There seem to be a way to define Windows-specific things like RTC level in one place so that it is used with both MSVC and Clang both with C and c++. That should not block merging this one, though, as the`/ZI` change is clearly a separate issue.

marstaik: My apologies, I had originally thought about making them separate, but I thought they were related enough to not justify it.
I reverted and will make a new PR with the argument cleanup.

PR #8180, opened 2021-01-11
Add some Chicken-Libraries

michaelbrockus:
Adding some Chicken Libraries packages to the big list of users.

### DIFF
--- a/docs/markdown/Users.md
+++ b/docs/markdown/Users.md
@@ -22,10 +22,7 @@ topic](https://github.com/topics/meson).
  - [Bzip2](https://gitlab.com/federicomenaquintero/bzip2), the bzip2 compressor/decompressor
  - [Cage](https://github.com/Hjdskes/cage), a Wayland kiosk
  - [casync](https://github.com/systemd/casync), Content-Addressable Data Synchronization Tool
- - [Chicken Stack](https://github.com/chicken-libraries/chicken-stack), Chicken libraries classic Stack data structure
- - [Chicken Forward List](https://github.com/chicken-libraries/chicken-forward-list), Chicken libraries classic Forward List data structure
- - [Chicken Queue](https://github.com/chicken-libraries/chicken-queue), Chicken libraries classic Queue data structure
- - [Chicken DeQueue](https://github.com/chicken-libraries/chicken-double-queue), Chicken libraries double ended Queue data structure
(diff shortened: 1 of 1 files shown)


dnicolodi:
Uh?! These seems to be empty identical skeleton packages. What is the point to add them to the list?

michaelbrockus:
True but I only sent links to those that had been implemented. As team of
one, I'm doing everything I can to get them implemented.

eli-schwartz:
At the very least, if you're going to add an entire organization's worth of small utility packages which all use meson, you should only add it once and link to the organization, not individual projects.

I don't want to be updating users.md every single time you add another ***templated*** component within this greater whole as I don't think that scales well, and it makes it harder to notice other projects in a long list of "chicken this, chicken that".

That would lead to users.md visibility being inherently unbalanced in favor of whoever splits their project into the most subrepos.

docs/markdown/Users.md
```diff
@@ -22,10 +22,7 @@ topic](https://github.com/topics/meson).
  - [Bzip2](https://gitlab.com/federicomenaquintero/bzip2), the bzip2 compressor/decompressor
  - [Cage](https://github.com/Hjdskes/cage), a Wayland kiosk
  - [casync](https://github.com/systemd/casync), Content-Addressable Data Synchronization Tool
- - [Chicken Stack](https://github.com/chicken-libraries/chicken-stack), Chicken libraries classic Stack data structure
- - [Chicken Forward List](https://github.com/chicken-libraries/chicken-forward-list), Chicken libraries classic Forward List data structure
- - [Chicken Queue](https://github.com/chicken-libraries/chicken-queue), Chicken libraries classic Queue data structure
- - [Chicken DeQueue](https://github.com/chicken-libraries/chicken-double-queue), Chicken libraries double ended Queue data structure
+ - [Chicken Libraries](https://github.com/chicken-libraries), Collection of C libraries. Data structures, algorithms, and more.
```
dnicolodi, line 25: Typo: `s/libraries. Data/libraries, data/`

eli-schwartz: I'll just squash and merge myself with a streamlined commit message...

eli-schwartz: Ugh, missed skipping CI as a result.

Issue #8058, opened 2020-12-03
OpenMP dependency not found with Nvidia HPC SDK


nordmoen:
**Describe the bug**
When using `openmp = dependency('openmp')` with Nvidia HPC SDK and Meson `0.56.0` OpenMP is not detected.

**To Reproduce**
Have `openmp = dependency('openmp')` and run `CC=nvc meson setup builddir`

**Expected behavior**
Expected OpenMP to be detected since it is supported by Nvidia HPC SDK [according to their documentation](https://docs.nvidia.com/hpc-sdk/compilers/hpc-compilers-user-guide/index.html#openmp-use).

I expect the issue to be that Nvidia (previously PGI) still does not define `_OPENMP`. In the code there is a special handling of `pgi` and that should probably be extended to `nvidia_hpc`. As per the link above the new Nvidia HPC SDK compilers should support version `4.5` (except for the offloading capabilities).

**system parameters**
* Meson `0.56.0`
* Nvidia HPC SDK `20.7` and `20.9`

dcbaker:
@scivision I think you're the only one with access to the Nvidia compilers.

nordmoen:
The [Nvidia HPC SDK](https://developer.nvidia.com/hpc-sdk) should be accessible to everyone as long as you accept [Nvidia's EULA](https://docs.nvidia.com/hpc-sdk/eula/index.html).

scivision:
I don't have the Nvidia compiler anymore

dcbaker:
I think then we're at "patches welcome". I'd be happy to review something though.

nordmoen:
It seems like my initial assessment was wrong, the Nvidia HPC SDK does indeed support the `_OPENMP` macro, however, `meson` was unable to detect this due to incompatible flags used with `nvc` when preprocessing to detect `_OPENMP`. One solution is to add `nvc` flags to `PGICompiler` or create a new Nvidia HPC SDK mixin with the required flags.

The following preprocessor flags (the **order** is important) will result in `nvc` handling the `_OPENMP` macro and printing the expected result.
```python
def get_preprocess_only_args(self) -> T.List[str]:
    return ['-Mcpp', '-E']
```

nordmoen:
@dcbaker Could you or anyone else comment on the above proposed changes?

dcbaker:
Does that still work with the PGI compiler, if it does just flipping it makes sense, otherwise some kind of inheritance based override make sense.

[...]

PR #8181, opened 2021-01-11
Added workaround for Nvidia HPC SDK OpenMP support

nordmoen:
Tested with Nvidia HPC SDK `20.7`, `20.9` and `20.11` which all worked with this patch.

There also seems to be some confusion in the Nvidia documentation. Previous versions of the compiler suite only supported `OpenMP` version `4.5`, however, the [latest documentation](https://docs.nvidia.com/hpc-sdk/compilers/hpc-compilers-user-guide/index.html#openmp-use) talks about supporting some features of version `5.0` without further specification. This should not affect this change, but in the future updates to the hardcoded version could be added.

This should solve #8058.

### DIFF
--- a/mesonbuild/dependencies/misc.py
+++ b/mesonbuild/dependencies/misc.py
@@ -82,6 +82,15 @@ class OpenMPDependency(ExternalDependency):
             self.is_found = True
             self.compile_args = self.link_args = self.clib_compiler.openmp_flags()
             return
+        elif self.clib_compiler.get_id() == 'nvidia_hpc':
+            # Nvidia HPC SDK compilers (rebranded from PGI) does not define the '_OPENMP' macro, but
+            # should support OpenMP according to their documentation
+            # (https://docs.nvidia.com/hpc-sdk/compilers/hpc-compilers-user-guide/index.html#openmp-use)
+            # (https://docs.nvidia.com/hpc-sdk/compilers/hpc-compilers-user-guide/index.html#standards)
+            self.version = '4.5'
+            self.is_found = True
+            self.compile_args = self.link_args = self.clib_compiler.openmp_flags()
+            return
         try:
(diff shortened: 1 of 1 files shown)


nordmoen:
This change makes `test cases/Common/185 openmp` pass.

nordmoen:
I did some more checking and found that `nvc` actually defines `_OPENMP`, however, when `meson` checks for the value of `_OPENMP` the test file can't be compiled since `nvc` expects a newline at the end of the file.

```bash
Compiler stderr:                                                                 
 "/builddir/meson-private/tmp9s1gnvx2/testfile.c", line 
          7: warning: last line of file ends without a newline                   
  _OPENMP                                                                        
         ^                                                                       
```
If someone has any suggestion how to fix this I would gladly implement that instead of the hacky workaround introduced in this PR.

nordmoen:
The above turned out to be slightly wrong, `nvc` is more than happy to compile the file if the correct set of flags are passed in, together with the correct order.

To compile the `_OPENMP` macro file one has to invoke the following command `nvc -Mcpp -E -mp testfile.c`. The output of that command conforms to what `meson` expects, however, if we change the order to `nvc -E -Mcpp -mp testfile.c` - `nvc` will not output anything (but will preprocess the file fine).

My question then, should I close this PR and open a new with a change to the [`PGICompiler`](https://github.com/mesonbuild/meson/blob/master/mesonbuild/compilers/mixins/pgi.py) with:
```python
def get_preprocess_only_args(self) -> T.List[str]:
        return ['-Mcpp', '-E']
```

(Unfortunately, the above fix doesn't quite work as for some reason the invoked command ends up being: `Command line:  nvc /builddir/meson-private/tmp4r636_pn/testfile.c -Mcpp -E -Mcpp -O0 -mp` which seem odd based on the function)

PR #8185, opened 2021-01-12
CI: ciimage: fix arch linux image failing to install wxgtk

eli-schwartz:
There hasn't been any such package since the original addition of a gtk3 version of wxgtk... back in 2017. The "new" wxgtk2 package provided a virtual provides ever since, so people still depending on "wxgtk" would get the old gtk2 version. This virtual provides got dropped today, resulting in the package being uninstallable.

Resolve the provides to its canonical name, thus making it installable again.


dcbaker (approved):

dcbaker:
Ubuntu seems to be broken as well, getting 404's on everything, but that's not the fault of this PR. this looks fine to me, (and i've tested a similar patch) so feel free to merge.

PR #8191, opened 2021-01-12
Keep buildtype the same even if user changes debug and/or optimization.

jpakkane:
I have had an itch to do this for a long time but now it's finally done. Changing builtype automatically was originally my idea and it turned out not to be a great one. This should make it more sensible. In this way buildtype is more of an "intent for the build" but the exact details (e.g. -O3 instead of -O2) can still be tweaked.

### DIFF
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -582,8 +582,35 @@ class CoreData:
 
         if key.name == 'buildtype':
             self._set_others_from_buildtype(value)
-        elif key.name in {'debug', 'optimization'}:
-            self._set_buildtype_from_others()
+
+    def get_nondefault_buildtype_args(self):
+        result= []
+        value = self.options[OptionKey('buildtype')].value
+        if value == 'plain':
+            opt = '0'
+            debug = False
+        elif value == 'debug':
+            opt = '0'
+            debug = True
+        elif value == 'debugoptimized':
+            opt = '2'
+            debug = True
+        elif value == 'release':
+            opt = '3'
+            debug = False
+        elif value == 'minsize':
+            opt = 's'
+            debug = True
+        else:
+            assert(value == 'custom')
+            return []
+        actual_opt = self.options[OptionKey('optimization')].value
+        actual_debug = self.options[OptionKey('debug')].value
+        if actual_opt != opt:
+            result.append(('optimization', actual_opt, opt))
+        if actual_debug != debug:
+            result.append(('debug', actual_debug, debug))
+        return result
 
     def _set_others_from_buildtype(self, value: str) -> None:
         if value == 'plain':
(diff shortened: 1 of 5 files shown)


dnicolodi:
This is a bit counter to what @eli-schwartz was saying on IRC. Also it is a bit strange to have a build configured with `--debug=false` still be called `debug`. I admit that I don't like the build type concept and having three knobs control two dimensions is a bit of a mess. What is the build type concept good for? Wouldn't it be simpler to deprecate it?

dcbaker:
Honestly... can we just deprecate buildtype all together? the split optimization/debug flags make more sense anyway.

jpakkane:
> Honestly... can we just deprecate buildtype all together? the split optimization/debug flags make more sense anyway.

I don't think we can, as all of our documentation uses it and people are very much used to it. I personally do like it because a) it's nice shortcut to bring back settings to known values and b) there needs to be some sort of a semantic option specifying what the build is intended for.

As an example you can't use debug info generation as a thing to determine whether the build is for day to day development or building a package for end users. You need to have debug info for release builds after all because otherwise crash traces from the field would be useless.

dcbaker:
Alright, I'm good with this then.

dcbaker (approved):

mesonbuild/coredata.py
```diff
@@ -582,8 +582,35 @@ def set_option(self, key: OptionKey, value) -> None:
 
         if key.name == 'buildtype':
             self._set_others_from_buildtype(value)
-        elif key.name in {'debug', 'optimization'}:
-            self._set_buildtype_from_others()
+
+    def get_nondefault_buildtype_args(self):
```
dcbaker, line 586:
```suggestion
    def get_nondefault_buildtype_args(self) -> str:
```

mesonbuild/mconf.py
```diff
@@ -243,6 +243,17 @@ def print_default_values_warning():
             print('')
             print_default_values_warning()
 
+        self.print_nondefault_buildtype_options()
+
+    def print_nondefault_buildtype_options(self):
```
dcbaker, line 248:
```suggestion
    def print_nondefault_buildtype_options(self) -> None:
```

dcbaker (approved):
It'd be nice to add type annotations to new functions so we don't have to do that later, otherwise LGTM

eli-schwartz:
> This is a bit counter to what @eli-schwartz was saying on IRC.

For context, in IRC I mentioned that currently, buildtype just means a preset combination of debug + optimization. But I didn't say this is necessarily a good thing... I do agree it's valuable to have it as a quick alias, but it's probably quite fine to refrain from modifying the buildtype once you pass in explicit values from debug/optimization.

A **one-way** mapping buildtype -> debug/opt seems sufficient. No need for it to be a two-way mapping.

jpakkane:
I'll update the docs still about the usage intent.

marstaik:
Even if buildtype isn't removed, I would recommend removing the custom options associated with buildtype, and making it *really* a reference to optimization + debug.

IIRC there are build types that add additional settings, which is kind of eh.

jpakkane:
Updated the FAQ.

jpakkane:
>  I would recommend removing the custom options associated with buildtype, and making it really a reference to optimization + debug

If I was to do this from scratch today, that is exactly how it would be implemented. We can probably achieve that by gradually moving stuff away from buildtype to individual options.

amitdo:
>Setting the `buildtype' option to 

-->

>Setting the `buildtype` option to

PR #8192, opened 2021-01-12
Add type annotations to minstall (and some related cleanups)

dcbaker:
I started out wanted to add a new feature, but ended up generating 10 patches for making minstall work with mypy. Since this is already a lot of patches (and should be user transparent) I'm submitting this as its own.

Apart from the annotations there are the following changes:
 - replace lists with tuples. Since the length is fixed, the position of each element is significant, and the types differ by position, this is exactly what a tuple is for
 - split `InstallDir` into a `build` module class and add an `interpreter.InstallDirHolder`
 - don't add new attributes to existing objects, pass them separately

This does use the `typing_extensions` module *only* when running mypy (it is guarded by `T.TYPE_CHECKING`). This is not a runtime dependency, just a type checking dependency. It allows us to use some typing features added in later versions of python in older ones.

The LGTM comment is one where we can't really win, because we need to try to import the name from the builtin module, and fall back to typing_extensions if it doesn't exist.


mesonbuild/minstall.py
```diff
@@ -533,7 +563,7 @@ def install_targets(self, d):
             if file_copied:
                 self.did_install_something = True
                 try:
-                    depfixer.fix_rpath(outname, t.rpath_dirs_to_remove, install_rpath, final_path,
+                    depfixer.fix_rpath(outname, sorted(t.rpath_dirs_to_remove), install_rpath, final_path,
```
jpakkane, line 574: Just curious but why add `sorted` here`?

mesonbuild/backend/backends.py
```diff
@@ -80,26 +81,31 @@ def __init__(self, build_dir, trees):
...
+        self.targets: T.List[TargetInstallData] = []
+        self.headers: T.List[T.Tuple[str, str, T.Optional['FileMode']]] = []
+        self.man: T.List[T.Tuple[str, str, T.Optional['FileMode']]] = []
+        self.data: T.List[T.Tuple[str, str, T.Optional['FileMode']]] = []
+        self.po_package_name: str = ''
         self.po = []
-        self.install_scripts = []
-        self.install_subdirs = []
+        self.install_scripts: T.List[build.RunScript] = []
+        self.install_subdirs: T.List[T.Tuple[str, str, T.Optional['FileMode'], T.Tuple[T.Set[str], T.Set[str]]]] = []
```
mensinda, line 101: Nothing wrong with the annotation, but this tuple should really be a new class for better readability...

mesonbuild/minstall.py
dcbaker: `rpath_dirs_to_remove` is actually a set, while `fix_rpath` expects a list, and I wasn't sure if order mattered. But looking at it more it looks like this value is only used for a single `if x in container` check, so just passing a set would be better. I'll fix that.

mesonbuild/backend/backends.py
dcbaker: I made a type alias for this as well as the other ones a bit above this to reduce duplication there.

jpakkane:
Good stuff.

PR #8194, opened 2021-01-12
summary: align left, not align middle

eli-schwartz:
aligning along the left is, I think, what most projects want to do. Aligning along the middle looks subjectively ugly, and objectively prevents me from further indenting an element, e.g.

```
  Build information:
    prefix                : /usr
    sysconfdir            : /etc
        conf file         : /etc/myprogram.conf
```

### DIFF
--- a/docs/markdown/Reference-manual.md
+++ b/docs/markdown/Reference-manual.md
@@ -1344,16 +1344,16 @@ Output:
 My Project 1.0
 
   Directories
-             prefix: /opt/gnome
-             bindir: bin
-             libdir: lib/x86_64-linux-gnu
-            datadir: share
+    prefix         : /opt/gnome
+    bindir         : bin
+    libdir         : lib/x86_64-linux-gnu
+    datadir        : share
 
   Configuration
-       Some boolean: False
+    Some boolean   : False
     Another boolean: True
-        Some string: Hello World
-             A list: string
+    Some string    : Hello World
+    A list         : string
                      1
                      True
 ```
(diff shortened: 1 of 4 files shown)


dcbaker (approved):

eli-schwartz:
Did not originally realize the unittest checked output in run_unittests.py, since I thought there was a json file for this...

dnicolodi:
FWIIW, I really liked the right aligned labels. I think they are immensely easier to read if there are labels of widely different length.

eli-schwartz:
The advantage of middle aligning is that the value is very close by and you cannot possibly confuse which line things refer to.

The advantage of left aligning is that it's easier to find the start of the line, it's easier to compare multiple "use foo, use barify, use that_thing_with_the_long_name" and internalize they all begin with the word "use", it gives you more control, as a developer, over the message alignment, it matches what I think most people do anyway... and it's what the initial implementation of summary() did, before it got modified halfway through the implementation process itself by a second person without much discussion, so I don't feel it should be set in stone. :D

Personally, I don't find the "follow which line things are on" argument to be very convincing, as I don't find this hard (and certainly don't wish to optimize for it), but in pathological cases if you triple-click on the thing you're interested in, your terminal will probably highlight the entire line for you. A more interesting solution would be found in many GUI programs, which shade in alternating rows of a spreadsheet layout for contrast.

xclaesse:
Not sure why this has been changed, middle align was on purpose because it makes easy to match the key with the value. Left align makes it difficult because the key is sometimes far away from the value and you can easily skip a line.

> and objectively prevents me from further indenting an element

We have sections for that, I don't think you should be doing indent like that.

Anyway... it's really a matter of taste...

PR #8193, opened 2021-01-12
Fix misspells

MisterDA:
Noticed a few misspells in the documentation, ran a checker on the codebase.

PR #8195, opened 2021-01-13
gen_data.py: sort files when generating mesondata

eli-schwartz:
The current way this works is chaos since the tool might return files in any order and thus shuffle around the order of embedded files. This results in big diffs that cannot be easily reviewed.

Also regenerate the data according to the, going forward, canonical ordering algorithm.

Noticed in #8193

### DIFF
--- a/mesonbuild/mesondata.py
+++ b/mesonbuild/mesondata.py
@@ -1,4 +1,4 @@
-# Copyright 2020 The Meson development team
+# Copyright 2021 The Meson development team
 
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


mesonbuild/mesondata.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2020 The Meson development team
+# Copyright 2021 The Meson development team
```
dcbaker, line 1: Shouldn't this be `2020-2021`?

dcbaker (approved):

mesonbuild/mesondata.py
eli-schwartz: Not according to `gen_data.py` :p

But, really I'd think the copyright info is invalid and meaningless, generated files don't have license info -- only the source files do (or do not actually, in this case).

PR #8197, opened 2021-01-13
gen_data.py: even more sorting

eli-schwartz:
glob.glob() is not sorted, despite using shell-style wildcards, and the documentation does not mention this: https://bugs.python.org/issue21748

Recently, it does start mentioning "Whether or not the results are sorted depends on the file system." which does not really get to the heart of the matter...

This is causing fuzz too.

### DIFF
--- a/mesonbuild/mesondata.py
+++ b/mesonbuild/mesondata.py
@@ -30,7 +30,92 @@ if T.TYPE_CHECKING:
 # BEGIN Data section #
 ######################
 
-file_0_data_CMakeLists_txt = '''\
+file_0_data_preload_cmake = '''\
+if(MESON_PS_LOADED)
+  return()
+endif()
+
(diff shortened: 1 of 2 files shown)


mensinda (approved):
LGTM

QuLogic:
> Recently, it does start mentioning "Whether or not the results are sorted depends on the file system." which does not really get to the heart of the matter...

What it's stating is that it's not _guaranteed_ to be sorted, so if that's wanted, then it should be done explicitly (as done here.)

eli-schwartz:
Unfortunately:
- this is not mentioned in the `help(glob.glob)`, only in the online docs
- it will not be obvious to readers that this is an intentional difference from shell globbing, and this may not stick in their minds
- the documentation that does exist is not present in 3.7 and below

So quite frankly this is all still rather unsatisfactory on the part of python itself. :(

PR #8200, opened 2021-01-14
mtest: improvements to logging

bonzini:
This pull requests includes several improvements to logging. In particular:
* in --verbose mode, the JSON and text file logs now include the output of the tests
* also in --verbose mode, the console log (similar to ninja) does not mix the output of multiple tests that are being run in parallel
* the console log and the text file log employ slightly different formatting, so that each medium gets the best readability
* the error logs for --print-errorlogs are emitted during the first part of the output, and not limited to 10 tests anymore.

### DIFF
--- a/docs/markdown/snippets/meson_test_logs.md
+++ b/docs/markdown/snippets/meson_test_logs.md
@@ -1,0 +1,29 @@
+## New logging format for `meson test`
+
+The console output format for `meson test` has changed in several ways.
+The major changes are:
+
+* if stdout is a tty, `meson` includes a progress report.
+
+* if `--print-errorlogs` is specified, the logs are printed as test runs
+rather than afterwards.  All the error logs are printed rather than only
+the first ten.
+
+* if `--verbose` is specified and `--num-processes` specifies more than
+one concurrent test, test output is buffered and printed after the
+test finishes
+
+* the console logs include a reproducer command.  If `--verbose` is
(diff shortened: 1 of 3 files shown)


bonzini:
Some examples of the new-style verbose logging: [parallel](https://asciinema.org/a/qYHQSBRjZIYvwrYNVmugzGeu4), [--num-processes 1](https://asciinema.org/a/IJ4etyLDCqsTWrAMrt1x5VW6n), [--print-errorlogs](https://asciinema.org/a/6Vs9FZbHj3MdGKxLxC6p4DJqg).

bonzini:
I dropped the emoji because it failed on non-UTF8 locales. I replaced them with a colored `>>>`, which can be kept as a fallback even if emoji are added back later.

Apart from the cuteness factor, the purpose of emoji was to let people identify easily output lines with the command line, for cut-and-paste into a terminal. The `>>>` marker serves the same purpose.

jpakkane:
The cleanups look nice. Anyone want to comment on the new output format?

bonzini:
Ok, let me extract the cleanups so that we can fix #8212. There are other small bugfixes at the beginning of the branch.

jpakkane:
To reiterate, this looks fine to me. If anyone has reasons why these should not be merged speak now, or forever hold your peace (or file MRs to improve the situation later).

PR #8201, opened 2021-01-14
mtest: allow quickly interrupting the test run

bonzini:
The new behavior of interrupting the longest running test with Ctrl-C is useful when tests hang, but not when the run is completely broken for some reason. Psychology tells us that the user will compulsively spam Ctrl-C in this case, so exit if three Ctrl-C's are detected within a second.


dcbaker (approved):
I like this behavior (it's the one the test suite I wrote for OpenGL uses, lol). I usually just use ctrl + \ to fully kill things, but I don't see any harm in having ctrl + c three times quickly kill it too.

eli-schwartz (approved):
I've already tried spamming CTRL+C to kill a meson test run in the past, this commit rationale rings true for me.

Issue #7766, opened 2020-09-21
Hardcoded paths in the NEEDED section


plattfot:
**Describe the bug**

I have notice meson is passing in absolute paths to shared libraries instead of using `-L` `-l`. This have some implication for libraries that doesn't have the `SONAME` set, as the linkers in `binutils` will use the absolute path if the `SONAME` is not set. This results in hard coded paths to external libraries ends up in the `NEEDED` section. This is undesirable as it defeats the purpose of using shared libraries in the first place.

Using `-L` then `-l` does not result in this behavior. 

I encountered this when linking against `mkl-2020.1`, but we have a bunch of in-house libraries that lacks the `SONAME` section.

**To Reproduce**
Here is a simple example on what I mean

`foo.hpp`:
```
#pragma once

int foo();
```

`foo.cpp`:
```
#include "foo.hpp"

int foo()
{
  return 1;
}
```

`main.cpp`:
```
#include <iostream>
#include "foo.hpp"

int main()
{
  std::cout<<foo()<<std::endl;
  return 0;
}
```

I'm using `gcc-9.3.0` and `binutils-2.29.1` to compile the example:

1. Compiling and linking the executable `foo` with `libfoo.so` using
   the absolute path to the library:

```
mkdir -p /var/tmp/scratch/test/private/build/foo-1.0.0
g++ -o /var/tmp/scratch/test/private/build/foo-1.0.0/libfoo.so -shared foo.cpp
g++ -o /var/tmp/scratch/test/private/build/foo-1.0.0/foo main.cpp -Wl,--start-group /var/tmp/scratch/test/private/build/foo-1.0.0/libfoo.so -Wl,--end-group
```

2. Use the same library as in 1, but instead of setting the absolute
   to `libfoo.so`; use -L to add the search path and -l to specify the
   library name to search for.

```
g++ -o /var/tmp/scratch/test/private/build/foo-1.0.0/foo_Ll main.cpp -Wl,--start-group -L/var/tmp/scratch/test/private/build/foo-1.0.0 -lfoo -Wl,--end-group
```

3. Compile the library with a soname, `libfoo_soname`. Then compile
   and link the executable `foo_soname` in the same manner as 1.

```
g++ -o /var/tmp/scratch/test/private/build/foo-1.0.0/libfoo_soname.so -shared foo.cpp -Wl,-soname=libfoo.so
[...]

PR #8203, opened 2021-01-14
Do not use absolute path to shared libraries that don't have a SONAME set

hwti:
When there is no SONAME, the linker would embed the absolute path as DT_NEEDED in the
output, which will never work when using a sysroot, for example when cross-compiling.

Fixes: #7766


hwti:
I tried to avoid trying to parse random binaries as potential ELF file by adding `if not self.info.is_cygwin() and not self.info.is_windows() and not self.info.is_osx()` (which if I'm right could use `PosixDynamicLinkerMixin`, but won't use ELF).
Is this correct ?

I don't know how to implement test cases, maybe with run_unittests.py to check the linker command line in the generated meson.build, but meson will always use an soname on Linux.

pfirsich:
It seems this PR has been ignored. Please reconsider merging it, as I would find use in this as well.

vjardin:
Please what's the status of this merge request ? Why is it still in draft state for more that 1 year ?

cc: @dcbaker and @nirbheek

hwti:
> Please what's the status of this merge request ? Why is it still in draft state for more that 1 year ?
> 
> cc: @dcbaker and @nirbheek

I created this PR as draft since I'm not sure I did everything properly, see https://github.com/mesonbuild/meson/pull/8203#issuecomment-760300350.
So it was my way to create it as "RFC", but maybe it isn't the proper way.

PR #8205, opened 2021-01-14
Avoid accidental use of STANDALONE_WASM mode on compiler tests

brion:
Compiler tests, such as checking for atomics support, could fail
when compiling to WebAssembly multithreaded targets because the
compiler tests got compiled to 'output.wasm'.

Using the '.wasm' suffix in recent versions of emscripten engages
STANDALONE_WASM mode, which disables features that require a JS
runtime like shared memory.

This created false negatives on support of those features when
building a library to be linked into an executable that is not
in STANDALONE_WASM mode.

Changing these to 'output.o' will continue to produce WebAssembly
object files, but they will no longer be configured for standalone
runtime mode.

### DIFF
--- a/mesonbuild/compilers/mixins/emscripten.py
+++ b/mesonbuild/compilers/mixins/emscripten.py
@@ -43,7 +43,7 @@ class EmscriptenMixin(Compiler):
         if mode == 'link':
             suffix = 'js'
         else:
-            suffix = 'wasm'
+            suffix = 'o'
         return os.path.join(dirname, 'output.' + suffix)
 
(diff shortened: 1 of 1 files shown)


jpakkane:
This begs the question: will this break for some people who want the exact opposite? I.e. do we eventually need some way of specifying whether you want to build standalone or to the web?

brion:
> This begs the question: will this break for some people who want the exact opposite? I.e. do we eventually need some way of specifying whether you want to build standalone or to the web?

I assume there's already a way to specify compiler options which would allow them to pass the suitable option, or else specify the appropriate output suffix at the link stage?

Issue #7920, opened 2020-10-30
dependency('xyzzy', required: false, include_type: 'system') crashes meson 0.56.0, python 3.6


borg323:
**Describe the bug**
`xyzzy = dependency('xyzzy', required: false, include_type: 'system')` crashes meson.

**To Reproduce**
```
project('foo', 'cpp')

xyzzy = dependency('xyzzy', required: false, include_type: 'system')
```

**Expected behavior**
Clean exit

**system parameters**
* plain native build (for the same computer)
* Ubuntu 16.04, but also reported on other linux versions and windows
* Python version 3.5.1 but also reported with 3.6
* meson 0.56.0

borg323:
An example on windows: <https://ci.appveyor.com/project/LeelaChessZero/lc0/builds/36058219/job/8hmig6upn7dx3g5d#L109>

eli-schwartz:
Not reproducible for me on Arch Linux + meson 0.56.2

drmoose:
@eli-schwartz Arch's Python is too new (3.9.1 at time of writing). You need <= 3.6

eli-schwartz:
That information could be better highlighted in the description...

drmoose:
If it helps, here's a Dockerfile that reproduces the issue:

```Dockerfile
FROM ubuntu:bionic

RUN set -xe ;\
    apt-get update ;\
    apt-get install -yqq python3-pip pkg-config curl ;\
    rm -rf /var/lib/apt/lists ;\
    pip3 install ninja meson ;\
    mkdir -p /src/subprojects ; cd /src ;\
    echo "project('foo', 'cpp')" > meson.build ;\
    echo >> meson.build ;\
    echo "dependency('spdlog', required: false, include_type: 'system')" >> meson.build ;\
    curl -o subprojects/spdlog.wrap https://wrapdb.mesonbuild.com/v1/projects/spdlog/1.8.2/1/get_wrap

WORKDIR /src

CMD ["bash", "-c", "meson $(mktemp -d)"]
```

jpakkane:
I did the same with debootstrap rather than Docker and it worked:

```
Version: 0.56.99
Source dir: /home/jpakkane/meson/deptest
Build dir: /home/jpakkane/meson/build
Build type: native build
Project name: foo
Project version: undefined
C++ compiler for the host machine: c++ (gcc 7.3.0 "c++ (Ubuntu 7.3.0-16ubuntu3) 7.3.0")
[...]

PR #8216, opened 2021-01-17
Do not store config parser object in Wrap object. Closes: #7920.

jpakkane:


drmoose:
confirmed e30350c48 solves the problem in my test environment

PR #8218, opened 2021-01-17
MSVC and Clang-Cl Compiler Argument Cleanup

marstaik:
This PR performs some cleanup for the msvc and clang-cl arguments.

* "Disable Debug" (`/Od`) is no longer manually specified for optimization levels {`0`,`g`} (it is already the default for MSVC).
* "Run Time Checking" (`/RTC1`) removed from `debug` buildtype by default
  * Maybe an option can be added for msvc-like compilers?
* Clang-CL `debug` buildtype arguments now match MSVC arguments
* There is now no difference between `buildtype` flags and `debug` + `optimization` flags

I removed some redundancy that was in the compile args (`\Od` is used by default, and `\O2` implies `\Ob2`)

### DIFF
--- a/mesonbuild/compilers/mixins/visualstudio.py
+++ b/mesonbuild/compilers/mixins/visualstudio.py
@@ -60,31 +60,13 @@ vs64_instruction_set_args = {
     'neon': None,
 }  # T.Dicst[str, T.Optional[T.List[str]]]
 
-msvc_buildtype_args = {
-    'plain': [],
-    'debug': ["/RTC1"],
-    'debugoptimized': [],
-    'release': [],
-    'minsize': [],
-    'custom': [],
-}  # type: T.Dict[str, T.List[str]]
-
-# Clang-cl /Zi and /Z7 do the same thing
-# quoting the docs (https://clang.llvm.org/docs/MSVCCompatibility.html):
-#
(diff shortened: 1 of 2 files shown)


mesonbuild/compilers/mixins/visualstudio.py
```diff
@@ -183,7 +165,7 @@ def get_output_args(self, target: str) -> T.List[str]:
         return ['/Fo' + target]
 
     def get_buildtype_args(self, buildtype: str) -> T.List[str]:
-        return msvc_buildtype_args[buildtype]
+        return []
```
dcbaker, line 168: is override of the base method required at all now?

dcbaker (changes requested):
This looks fine, just one question about whether we need to override an inherited method.

mesonbuild/compilers/mixins/visualstudio.py
marstaik: Yes, if it is not implemented it will throw an exception.

dcbaker: cool

dcbaker (approved):
can you rebase this and resolve the conflicts, then ping me on matrix/irc and I'll merge this?

PR #8220, opened 2021-01-18
Base \MD[d] and \MT[d] off of 'debug' instead of 'buildtype'

marstaik:
Currently, the msvc the automatic crt flag detection is based off of `buildtype`, this PR would change it to be based off of `debug` for the following reasons.

* The [d] suffix on either `\MD` or `\MT` directly correlates to whether or not `_DEBUG` is defined, and many projects depend on `_DEBUG` being defined by msvc when debug information is actually built. Currently, `buildtype` will turn off the debug information on `debugoptimized`.
* Mixing custom `optimization` and `debug` levels causes `buildtype=custom`, which throws the following exception:
`Requested C runtime based on buildtype, but buildtype is "custom".`
  * Current behavior requires a user to manually specify the vs crt in this case, which adds complexity
  * I do not believe a default option should cause exceptions to be thrown.

### DIFF
--- a/docs/markdown/Builtin-options.md
+++ b/docs/markdown/Builtin-options.md
@@ -125,24 +125,21 @@ platforms or with all compilers:
 | b_pgo       | off           | off, generate, use      | Use profile guided optimization |
 | b_sanitize  | none          | see below               | Code sanitizer to use |
 | b_staticpic | true          | true, false             | Build static libraries as position independent |
-| b_pie       | false         | true, false             | Build position-independent executables (since 0.49.0)|
-| b_vscrt     | from_buildtype| none, md, mdd, mt, mtd, from_buildtype, static_from_buildtype | VS runtime library to use (since 0.48.0) (static_from_buildtype since 0.56.0) |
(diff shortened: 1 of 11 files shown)


jpakkane:
>  Currently, buildtype will turn off the debug information on debugoptimized.

There is an unfortunate clash of terminology here. Meson uses the unixy terms where "debug" means "-g", that is, debug info is generated. MSVC uses "debug" to mean whether the debug runtime is used (for GCC it would mean roughly the same as using `-D_GLIBXCC_DEBUG=1`). I'm guessing that you meant the latter here, because `debugoptimized` most definitely has debug info enabled. 

Because of this setting this option based on Meson's `debug` option does not seem right as it is enabled in release builds (because otherwise you can't debug stack traces from crashes in the field). See also #8191 for documentation update on what the `buildtype` option is meant to convey in the future.

> Current behavior requires a user to manually specify the vs crt in this case, which adds complexity

An alternative would be to change it so that for `custom` it adds no flags at all.

bruchar1:
As @jpakkane explained, this is wrong as 'debug' means 'generate debug info'. This PR should be closed.

PR #8226, opened 2021-01-20
Fixes for fortran: Include dirs for link_whole_targets and capital file suffix

jonaslb:
While working on some meson build systems for fortran projects, I've run into a few issues wrt dependency resolution. This PR should fix those issues I believe. There's a test too. 

Basically, some link targets were promoted to `link_whole_targets` when the build target had `install:true`, but `link_whole_targets` was not included in the loop that adds include directories for fortran static libraries. Further, some code-scanning code had file suffix checks that were case sensitive, but `.F90` is a common filename extension for fortran sources.

The projects that I work on make use of generated sources, so that's why the test does it too. This turned out not to be the source of the issues, but I left it in anyway. Hope it's fine.

### DIFF
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -45,6 +45,11 @@ if T.TYPE_CHECKING:
     InstallType = T.List[T.Tuple[str, str, T.Optional['FileMode']]]
     InstallSubdirsType = T.List[T.Tuple[str, str, T.Optional['FileMode'], T.Tuple[T.Set[str], T.Set[str]]]]
 
+# Languages that can mix with C or C++ but don't support unity builds yet
+# because the syntax we use for unity builds is specific to C/++/ObjC/++.
+# Assembly files cannot be unitified and neither can LLVM IR files
+LANGS_CANT_UNITY = ('d', 'fortran')
+
 class TestProtocol(enum.Enum):
 
     EXITCODE = 0
@@ -637,6 +642,9 @@ class Backend:
             unity_size = self.get_option_for_target(OptionKey('unity_size'), extobj.target)
 
             for comp, srcs in compsrcs.items():
+                if comp.language in LANGS_CANT_UNITY:
+                    sources += srcs
+                    continue
                 for i in range(len(srcs) // unity_size + 1):
                     osrc = self.get_unity_source_file(extobj.target,
                                                       comp.get_default_suffix(), i)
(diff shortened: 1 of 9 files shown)


jonaslb:
So I checked that the test should pass by running meson and ninja manually in the directory, but now it looks like it's failing in CI anyway. I will try to figure out why.

jonaslb:
Ok, the test is working in most of the CI - except the unity builds. Even though meson during configure says 
```FIXME FORTRAN is not supported in Unity builds yet, so FORTRAN sources in the 'main_exe' target will be compiled normally```, 
compiling still ends with an error 
`ninja: error: 'subprojects/static_hello/libstatic_hello.a.p/meson-generated_static_hello-unity0.f90.o', needed by 'libmainstatic.a', missing and no known rule to make it`.
So now I have to fix the unity stuff too...

**Comment by jonaslb** [comment]
The tests have all passed now, except for the azure ones which appear to be stuck as "queued", even after hours. So I've marked the PR as ready for review again :)

File: mesonbuild/backend/backends.py [hunk 1]
```diff
@@ -636,6 +637,10 @@ def determine_ext_objs(self, extobj, proj_dir_to_build_root):
             unity_size = self.get_option_for_target(OptionKey('unity_size'), extobj.target)
 
             for comp, srcs in compsrcs.items():
+                if comp.language in ('d', 'fortran'):
+                    # See langs_cant_unity in ninjabackend.py
```
dcbaker, line 641: Can we just move `langs_cant_unity` into the `backends` module and import it into `ninjabackend` instead of duplicating?

dcbaker:
@scivision, you're the resident fortran expert. Do the fortran changes look good?

mesonbuild/backend/backends.py
jonaslb: Hey, I moved `ninjabackend.NinjaBackend.langs_cant_unity` into `backends.LANGS_CANT_UNITY` in the most recent revision, as you suggested.

dcbaker: Thanks!

dcbaker (approved):
The python changes look good to me, you have my review for that. I am *not* qualified to consider the fortran changes though.

jonaslb:
I don't expect the changes to be controversial at all, they are only small bug fixes without which the small test I added doesn't compile. But thanks for reviewing! I suppose we'll just wait for scivision (or jpakkane i suppose?) to peek by for the final go.

scivision:
Thank you. I don't "install"Fortran code in the formal way as often as I might, so I didn't realize that was an issue. Also I haven't used Meson directly with pre-processed code much so I'm glad you caught that as well.

PR #8229, opened 2021-01-21
dependency: Add JDK system dependency

### LINKED ISSUE #8227: Add a Java system dependency for facilitating JNI development
I am working on Java bindings for a project. Java installations typically come with an `include` folder that has various headers for interacting with the JVM or the JNI. The `include` directory isn't symlinked into a default header location by default, no does it typically reside in the same location across systems (Fedora, Ubuntu, Windows, Mac, etc.).

I want to propose a Java system dependency that hooks up any libraries and include directories to your project.

I am interested in creating a PR for this myself.

tristan957:
The JDK system dependency is important for detecting JDK include paths
that may be useful when developing a JNI interface.

I am apprehensive of the PR due to the way that the Java version is detected in detect_java_compiler(). Looking at `/usr/bin/javac` is wrong to me, and a change should occur in which JAVA_HOME/bin/javac is looked at first and if that comes up empty then look at `/usr/bin/javac` I would be more than happy to make that a PR as well.

Fixes #8227

### DIFF
--- a/mesonbuild/wrap/wrap.py
+++ b/mesonbuild/wrap/wrap.py
@@ -512,7 +512,7 @@ class Resolver:
     def check_hash(self, what: str, path: str, hash_required: bool = True) -> None:
         if what + '_hash' not in self.wrap.values and not hash_required:
             return
-        expected = self.wrap.get(what + '_hash')
+        expected = self.wrap.get(what + '_hash').lower()
         h = hashlib.sha256()
         with open(path, 'rb') as f:
             h.update(f.read())
@@ -526,7 +526,7 @@ class Resolver:
         mlog.log('Downloading', mlog.bold(self.packagename), what, 'from', mlog.bold(srcurl))
         try:
             dhash, tmpfile = self.get_data(srcurl)
-            expected = self.wrap.get(what + '_hash')
+            expected = self.wrap.get(what + '_hash').lower()
             if dhash != expected:
                 os.remove(tmpfile)
                 raise WrapException('Incorrect hash for {}:\n {} expected\n {} actual.'.format(what, expected, dhash))


tristan957:
I am also not super tied to the name I picked `jdk`. Someone in irc suggested `jni`. I picked `jdk` over `jni` because there are other headers that don't necessarily apply to the JNI.

tristan957:
To be honest, I am a little confused by the test errors. At least they are all the same :/

mesonbuild/dependencies/dev.py
```diff
@@ -505,6 +506,59 @@ def get_methods():
...
+
+        java_home = os.environ.get('JAVA_HOME')
+        if java_home is None:
+            mlog.debug('$JAVA_HOME is unset')
+            self.is_found = False
+            return
+
+        self.version = environment.detect_java_compiler(self.for_machine).version
+
+        platform_include_dir = JDKSystemDependency.__machine_info_to_platform_include_dir(m)
```
dcbaker, line 523:
```suggestion
        platform_include_dir = self.__machine_info_to_platform_include_dir(m)
```

self can be used to access all methods of a class in python, and it is idiomatic to do so.

mesonbuild/dependencies/dev.py
```diff
@@ -505,6 +506,59 @@ def get_methods():
         return [DependencyMethods.SYSTEM]
 
 
+class JDKSystemDependency(ExternalDependency):
+    def __init__(self, environment: 'Environment', kwargs: T.Dict[str, T.Any]):
+        super().__init__('jdk', environment, kwargs)
+
+        m = self.env.machines[self.for_machine]
+
+        java_home = os.environ.get('JAVA_HOME')
```
dcbaker, line 515: I'd really like to move the JAVA_HOME detection into `envconfig.Properties` and `environment.Environment`, I just got all our environment variable handling centralized there. `BOOST_ROOT` is handled in a smiliar fashion.

This would also be handy if we were to change the `javac` detection to `$JAVA_HOME/bin/javac`, since you'd avoid doing multiple lookups.

mesonbuild/dependencies/dev.py
```diff
@@ -505,6 +506,59 @@ def get_methods():
...
+        java_home_include = pathlib.Path(java_home) / 'include'
+
+        if self.clib_compiler.get_argument_syntax() == 'gcc':
+            self.compile_args.append(f'-I{java_home_include}')
+            self.compile_args.append(f'-I{java_home_include / platform_include_dir}')
+            self.is_found = True
+        elif self.clib_compiler.get_argument_syntax() == 'msvc':
+            self.compile_args.append(f'/I{java_home_include}')
+            self.compile_args.append(f'/I{java_home_include / platform_include_dir}')
+            self.is_found = True
```
dcbaker, line 538: You shouldn't need to have the per compiler syntax, internally meson uses unix style arguments, and translates them into the style that the compiler takes natively in the backend.

mesonbuild/dependencies/dev.py
```diff
@@ -505,6 +506,59 @@ def get_methods():
...
+
+    @staticmethod
+    def get_methods():
```
dcbaker, line 541:
```suggestion
    def get_methods() -> T.List[DependencyMethods]:
```

mesonbuild/dependencies/dev.py
```diff
@@ -505,6 +506,59 @@ def get_methods():
...
+        return [DependencyMethods.SYSTEM]
+
+    @staticmethod
+    def __machine_info_to_platform_include_dir(m: MachineInfo) -> T.Optional[str]:
```
dcbaker, line 545: It doesn't look like you've imported `MachineInfo`, I'd suggest adding
```python
from ..envconfig import MachineInfo
```
inside the `if T.TYPE_CHECKING:` block, then change this to a forward declaration (python < 3.7 does that with a string) so:
````python
def __function(m: 'MachineInfo') ...
```

mesonbuild/dependencies/dev.py
```diff
@@ -505,6 +506,59 @@ def get_methods():
...
+
+        m = self.env.machines[self.for_machine]
+
+        java_home = os.environ.get('JAVA_HOME')
+        if java_home is None:
+            mlog.debug('$JAVA_HOME is unset')
+            self.is_found = False
+            return
+
+        self.version = environment.detect_java_compiler(self.for_machine).version
```
dcbaker, line 521: I need to look tomorrow, but there may already be an instance of the java compiler available to you here.

mesonbuild/environment.py
```diff
@@ -310,6 +311,15 @@ def detect_clangformat() -> T.List[str]:
...
 
+def detect_java() -> T.List[str]:
+    java_home = os.environ['JAVA_HOME']
+    if java_home is None:
+        return []
+
+    java_home = pathlib.Path(java_home) / 'bin' / 'java'
+
+    return [str(java_home), '--version']
+
```
dcbaker, line 322: This appears to be unused? leftovers from a different approach maybe?

dcbaker (changes requested):
This looks pretty good. I've got a few python style suggestions, and a a fix to your CI woes. Plus one big enhancement I'd like to see.

mesonbuild/dependencies/dev.py
dcbaker: This is why the tests are failing, btw.

mesonbuild/dependencies/dev.py
tristan957: Cool thanks.

mesonbuild/dependencies/dev.py
tristan957: Awesome fixed.

mesonbuild/dependencies/dev.py
tristan957: Are you referring to `_set_default_properties_from_env`?

tristan957: I think I am following you. HAving trouble getting a reference to Environment from detect_java_compiler() though.

tristan957: Nevermind I am dumb again

tristan957:
So I have responded to all comments except for the default compiler that you speak of. Hopefully the tests now pass.

mesonbuild/environment.py
```diff
@@ -1677,7 +1679,11 @@ def _detect_objc_or_objcpp_compiler(self, for_machine: MachineChoice, objc: bool
...
     def detect_java_compiler(self, for_machine):
-        exelist = self.lookup_binary_entry(for_machine, 'java')
+        java_home = self.properties[for_machine].get_java_home()
+        if java_home is not None:
+            exelist = [str(pathlib.Path(java_home) / 'bin' / 'javac')]
+        else:
+            exelist = self.lookup_binary_entry(for_machine, 'javac')
         info = self.machines[for_machine]
         if exelist is None:
             # TODO support fallback
```
tristan957, line 1689: Can this TODO be removed?

tristan957:
Looks like tests fail because JAVA_HOME isn't set and there doesn't seem to be javac installed in the base images which makes sense.

tristan957:
Leaving myself a note for tomorrow that the debug line "$JAVA_HOME is unset" isn't necessarily applicable anymore. Will fix.

mesonbuild/environment.py
dcbaker: yes, I think it can.

mesonbuild/environment.py
```diff
@@ -1677,7 +1679,11 @@ def _detect_objc_or_objcpp_compiler(self, for_machine: MachineChoice, objc: bool
         self._handle_exceptions(popen_exceptions, compilers)
 
     def detect_java_compiler(self, for_machine):
-        exelist = self.lookup_binary_entry(for_machine, 'java')
+        java_home = self.properties[for_machine].get_java_home()
+        if java_home is not None:
+            exelist = [str(pathlib.Path(java_home) / 'bin' / 'javac')]
+        else:
+            exelist = self.lookup_binary_entry(for_machine, 'javac')
```
dcbaker, line 1686: You might want to look at the C# detection logic below this, since we now potentially have two places to look ($JAVA_HOME/bin and $PATH), also, the `lookup_binary_entry` should be the first choice, the JAVA_HOME should be the second.

You might need to add an optional parameter to the `Environment._get_compilers` method, or do something clever with the `self.default_java` attribute. I'd probably add an optional keyword argument though. Maybe `prepend_candidates`? This is important because we guaranteed that if you put a compiler in your machine file or use CC=clang that we will either give you clang for your c compiler, or error out.

dcbaker (changes requested):
This is looking really nice, thanks for addressing all of my review feedback (I know there's a lot!). There's a few more things here, and then can you add an entry in `docs/markdown/dependencies.md` for `JDK`? Please make sure you document that it reads $JAVA_HOME or the java_home value in the `[properties]` section of the machine file.

mesonbuild/environment.py
dcbaker: Putting `lookup_binary_entry` first will also fix the test failures I think.

mesonbuild/envconfig.py
```diff
@@ -208,6 +208,14 @@ def get_cmake_use_exe_wrapper(self) -> bool:
         assert isinstance(res, bool)
         return res
 
+    def get_java_home(self) -> T.Optional[str]:
+        if 'java_home' not in self.properties:
+            return None
+
+        raw = self.properties['java_home']
+        assert isinstance(raw, str)
+        return raw
```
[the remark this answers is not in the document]
eli-schwartz: Assert is meant to check "things that can never happen unless the source code is wrong", and may get optimized out of the bytecode.

mesonbuild/environment.py
tristan957: Feels weird that we look for the includes in $JAVA_HOME, but look for the binary if it appears on $PATH. Because $JAVA_HOME/bin/javac isn't necessarily javac that is on the $PATH.

I am also wondering if maybe looking at $JAVA_HOME is dumb, and instead we should essentially do a `dirname $(dirname $(readlink -f $(which javac)))` to get to the include directory. This is easily supported using `pathlib`. This has the caveat that assumes javac on $PATH is a symlink, but maybe if it isn't a symlink we can fallback to the current behavior.

tristan957: This has the added benefit of relying on system tooling for toggling Java (alternatives on Fedora, etc.)

test cases/java/9 jdk/lib/meson.build
```diff
@@ -0,0 +1,7 @@
+sources = files(
+  'native.c',
+)
+
+library('jdkjava', sources,
+  dependencies : [jdk],
+)
```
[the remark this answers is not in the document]
tristan957: Oh you want me to hook it up to some Java code and maybe print hello world or something?

mesonbuild/environment.py
dcbaker: The lookup_binary_entry doesn't check $PATH, that's checked if someone says `JAVAC=/path/to/some/javac meson builddir` or sets `[binaries]javac = '/some/path'` in their machine file. That needs to be checked first.

After that I'm opened to doing whatever would be best (checking to see if javac is a symlink, looking in $JAVA_HOME/bin, whatever.

test cases/java/9 jdk/lib/meson.build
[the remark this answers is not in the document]
dcbaker: Sure, or just a main that calls a function that returns 0 is fine honestly, the equivalent of:

some_lib
```c
int func(void) { return 0; }
```
exe linked with some_lib:
```c
#include "some_lib.h"

int main(void) { return func(); }
```

mesonbuild/environment.py
tristan957: Oh ok had a misunderstanding. Now I get it. I will fix it appropriately later today. Thanks for your explanations

test cases/java/9 jdk/lib/meson.build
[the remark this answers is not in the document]
tristan957: Sure yea. I just need to update all the CI images to install jdk-devel on them

mesonbuild/environment.py
dcbaker: no problem, thanks for working on this!

jpakkane:
This feature _must_ also work out of the box even if `JAVA_HOME` is not set. That is, it should sniff the header locations from the system. That is, on for example Debian it needs to look under `/usr/lib/jvm`. The documentation should also be updated to reflect that `JAVA_HOME` is only needed if you want to override the builtin detection system.

Also, the JNI test should be updated to have a Java program that actually calls the C code to ensure that the full pipeline is working.

tristan957:
@jpakkane on Debian, what does `readlink -f $(which javac)` return?

jpakkane:
/usr/lib/jvm/java-15-openjdk-amd64/bin/javac

tristan957:
Sweet thanks.

I think the best approach is to look for javac in the following ways.

1. If JAVAC is set during setup of Meson
2. If JAVA_HOME is set
3. $(which javac)

Once we have found the correct javac. It should just be a matter of following the symlink and looking up the directory structure to get to the accompanying `include` directory.

jpakkane:
On Ubuntu at least the main JNI file is at `/usr/lib/jvm/java-15-openjdk-amd64/include/jni.h` and for example `libjava` is at `/usr/lib/jvm/java-15-openjdk-amd64/lib/libjava.so`.

tristan957:
Yep that is what I expect. Java installations have predictable file tree layouts.

tristan957:
gonna come back with a new pr. My forks master was messed up

nanthony21:
@tristan957 Did another PR get made? Having a hard time following the trail here. Would be awesome to have something that works for #8227 on all platforms.

tristan957:
@nanthony21 check the Meson 0.58 release notes

nanthony21:
Thanks @tristan957. Not sure if this is worthy of it's own issue but when I try this I get `Could not execute Java compiler "javac"`. No further useful information or stacktrace is in the log file. Presumably it's coming from https://github.com/mesonbuild/meson/blob/802f66f442d4ce8d5aadc6cbeb67ec7db2414da4/mesonbuild/environment.py#L1723 

Any idea what the cause of this is? I'm on Windows with `JAVA_HOME` set to the install location of jdk1.8.0. I also get this same error if I just try to add `java` to the projects `languages` kwarg.

Do I have to add `javac` to my `PATH`. I'm hoping that's not the case since I am just wanting to use the JNI as a c++ dependency anyway.

tristan957:
@nanthony21 the issue you're coming across is that java_home isn't used to provide the binary. You should setup a host file that tells meson where to find javac on your system. You can also set java_home in the same file if you wanted.

nanthony21:
Ok, thanks. I'll give the machine file a try. Would be great to one day have a more automatic way of finding the JNI. Using `dependency('jni')` works for me at least on Windows via CMake, but would be nice to not depend on CMake.

tristan957:
That's what dependency('jdk') does. It adds everything you need to the include path. But you have to tell Meson where javac is or java_home is, otherwise it can't set anything up for you.

If there is a bug, then open one and tag me. So far I haven't understood any bugs from what you've told me.

PR #8231, opened 2021-01-21
Allow passing external program and dependency objects to summary()

bonzini:
It is common to include in the summary whether a program or dependency was found. Simplify the code that does so, to avoid things such as

     prg.found() ? prg.full_path() : false


xclaesse:
Thanks, seems useful indeed. LGTM.

PR #8235, opened 2021-01-21
custom_target: Add env kwarg

xclaesse:


tristan957:
Will env valeus have @KEYWORD@ substitution like the arguments do?

docs/markdown/Reference-manual.md
```diff
@@ -372,6 +372,10 @@ following.
 - `install_mode` *(since 0.47.0)*: the file mode and optionally the
   owner/uid and group/gid
 - `output`: list of output files
+- `env` *(since 0.57.0)*: environment variables to set, such as `['NAME1=value1',
+   'NAME2=value2']`, or an [`environment()`
+   object](#environment-object) which allows more sophisticated
+   environment juggling. A dictionary is also accepted.
```
dcbaker, line 378: since dicts are always supported, maybe reword this the opposite way?
```markdown
environment variables to set, such as `{'NAME1': 'value1', 'NAME2': 'value2'}`, or ... A list of strings in the form `NAME=value` is also accepted.
```

dcbaker: I really feel like in the long run we probably want to deprecate the `['NAME=value']` syntax, as it's pretty magical and dicts or `environment()` objects are much nicer to work with anyway.

dcbaker (approved):
LGTM. One suggestion on the wording of the documentation, take it or leave it.

xclaesse:
> Will env valeus have @Keyword@ substitution like the command arguments do?

Not atm, could be added later I guess if needed? What's your use-case?

docs/markdown/Reference-manual.md
xclaesse: Right, better have the dict syntax first since that's what we should recommend now, made that change now. We could indeed deprecate the string list format, but I preferred to keep it here for consistency with other APIs.

tp-m:
Yay!

tristan957:
> > Will env valeus have @Keyword@ substitution like the command arguments do?
> 
> Not atm, could be added later I guess if needed? What's your use-case?

Was just thinking to be consistent with arguments.

tristan957:
@xclaesse could this be extended to run_target() as well?

Use case: I am building a Python extension, and want to add a run_target() that opens a Python shell with the PYTHONPATH pointing to the correct folder in the builddir.

xclaesse:
That would make sense, and made me realize run_target() uses completely different code path than custom_target() to run the command. Big code duplication that should get fixed too.

tristan957:
Cool. Code clean-up is a beautiful thing.

xclaesse:
@tristan957 FYI it's in https://github.com/mesonbuild/meson/pull/8305

tristan957:
Did this make it in for 0.57? Seems like 0.57 could be a fun release

xclaesse:
For run_target() it's not yet merged, not sure we'll have time since the freeze is this weekend. Unfortunately since it's a much bigger refactor than expected, it could have to be delayed to 0.58. Depends if someone can do the review...

tristan957:
That's cool. Nbd either way

PR #8236, opened 2021-01-21
Submit/rust fix generated sources in subdir

dcbaker:
Turns out that I didn't fix all the generated rust problems. The current code doesn't handle subdirs correctly with custom_targets or their indexes.

### DIFF
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -1598,7 +1598,7 @@ int dummy;
                 if isinstance(g, GeneratedList):
                     fname = os.path.join(self.get_target_private_dir(target), i)
                 else:
-                    fname = i
+                    fname = os.path.join(g.get_subdir(), i)
                 if main_rust_file is None:
                     main_rust_file = fname
                 orderdeps.append(fname)
--- a/test cases/rust/11 generated main/gen.py
+++ b/test cases/rust/11 generated main/gen.py
@@ -6,10 +6,14 @@ import argparse
 def main() -> None:
     parser = argparse.ArgumentParser()
     parser.add_argument('out')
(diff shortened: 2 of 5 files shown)


jljusten:
@dcbaker: These patches allowed me to run bindgen as a custom_target, and ninja would now attempt the build. But I found subsequent linking build step still failed. This may have been specific to my use case where I tried to produce a rust-library and link the C static library into the rust library.

By adding your
```
eaee775d9 backends/ninja: Fix linking rust targets to non-rust targets
```
patch from #8162 on top of these patches, I was able to get the link to work.

Should that patch's subject perhaps be tweaked to?
```
backends/ninja: Fix linking non-rust targets into rust targets
```

jpakkane:
How should these two MRs be merged? First one, then rebase the other or something else?

jljusten:
> How should these two MRs be merged? First one, then rebase the other or something else?

@jpakkane: The simplest would be to merge #8162, which includes these patches. (And then close this one)

If that PR still needs some time, then I'd recommend @dcbaker add the patch I referenced to this PR and merge it first. Hopefully #8162 won't be too far along because it seems like a much easier way to run bindgen, which would be helpful for projects wanting to mix rust and C.

Maybe @dcbaker would want to get your thoughts on how close to ready #8162 is to merge to find out if he should update this PR to add the 1 patch from the other PR.

dcbaker:
I think this one is ready to go and could land immediately, but it doesn't really matter. The problem for the other one is what to do about the Eaon CI.

jljusten:
@dcbaker: you don't think adding `backends/ninja: Fix linking non-rust targets into rust targets` into this PR is worth it? Hopefully #8162 will merge before the next release, and in that case kind of make this point irrelevant.

Issue #4919, opened 2019-02-13
Meson doesn't detect files ending in .C as C++


jml1795:
It seems that meson doesn't correctly detect files ending in the extension `.C` (note capital C) as C++ code. According to the [Gcc manual](https://gcc.gnu.org/onlinedocs/gcc-4.8.5/gcc/Overall-Options.html#index-file-name-suffix-71), this is an accepted file extension to be processed as C++. I realize that it isn't a common extension nor is it one I have ever seen outside my current project, but I cannot change the convention in my project at this time. Any chance we could add support for this in meson? I am happy to make the update if you can point me in the right direction.

I can get around this by passing `-xc++` but it would be nice if we could support it directly.

This can be demonstrated by running `meson init -l cpp` and changing the generated file's extension to `.C`.

I am running gcc 4.8.5 if that matters and meson off `07f1175c`. Thanks again.

QuLogic:
Generally the problem with this is that other systems are not necessarily case-sensitive. There are two incomplete PRs for this that never hit those corner cases #3409, #3415.

jml1795:
@QuLogic Thanks for the links. Is there a simple solution we can implement? Something to the effect of @nirbheek 's [proposal](https://github.com/mesonbuild/meson/pull/3409#discussion_r181566229) to be case sensitive on extensions and platform combinations where it matters and follow existing behavior elsewhere?

jpakkane:
Because of backwards compatibility, everything is terrible. There are Windows projects that have `.C` ending for C files because they originate from those systems. Then there are other projects that have `.C` for C++. Then, thanks to the magic of subprojects, you may have _both_ of these within one project and it is impossible to know whether a file with the extension `.C` should be treated as C or C++.

jml1795:
@jpakkane  Maybe overly simplistic, but could we allow a project-wide opt in? Something to the effect of:

```py
map_ext_to_language(ext: '.c', language: 'c')
map_ext_to_language(ext: '.C', language: 'cpp')
...
```
which would maintain current behavior if not overridden and override the extension detection mechanism if set. Not sure this solves the case of mixing conventions in subprojects, but maybe gets us closer?

nirbheek:
It's really much simpler to just rename your source files.

jpakkane:
It's also the "correct" long term solution and something you should do _anyway_. Granted, the transition does take some effort but in the end it's worth it.

scivision:
[...]

PR #3415, opened 2018-04-15
Meson recognize .C files as C sources while it is C++ Option 2

jeandet:
This PR adds .C as C++ extension and makes Meson case sensitive for
file extension detection only for C and C++ compilers.

See PR #3409 
@textshell if you have comments, you are welcome :)

PR #3409, opened 2018-04-14
Meson recognize .C files as C sources while it is C++

jeandet:
This PR adds .C as C++ extension and makes Meson case sensitive for
file extension detection.

I saw that while porting [SAMRAI](https://github.com/jeandet/SAMRAI) to meson they use .C as file extension for all their C++ sources.

PR #8239, opened 2021-01-22
added language: kwarg to files()

Volker-Weissmann:
We are discussing how to handle *.C files for nearly two years now.
I think we need to stop bike shedding and approve this PR.
This PR is fully backward compatible, does not introduce substantial amounts of tech-dept and solves a real-world problem (namely: some projects cannot be built without this patch).

Links:
https://github.com/mesonbuild/meson/issues/4919
https://github.com/mesonbuild/meson/pull/3409
https://github.com/mesonbuild/meson/pull/3415


dcbaker:
Hard NAK from me. This leads to all kinds of insanity, I had to take over a project in cmake that did the equivalent of `files('foo.c', lanaguage : 'objcpp')` once. This is the kind of foot gun insanity that meson *doesn't* do. Maybe @jpakkane will override me on this, but I'm **super** against this approach.

I'd be more open to a *specific* option for .C files, but not this.

Volker-Weissmann:
> Hard NAK from me. This leads to all kinds of insanity, I had to take over a project in cmake that did the equivalent of `files('foo.c', lanaguage : 'objcpp')` once. This is the kind of foot gun insanity that meson _doesn't_ do. Maybe @jpakkane will override me on this, but I'm **super** against this approach.
> 
> I'd be more open to a _specific_ option for .C files, but not this.

It IS a _specic_ option for .C files:
```python
        lang_clarification = kwargs['language'] if 'language' in kwargs else None
        if lang_clarification is not None:
            if lang_clarification not in ['c', 'cpp']:
                raise MesonException('The \'language\' argument has to be \'c\' or \'cpp\'.')
            for fname in args:
                suffix = os.path.splitext(fname)[1]
                if suffix not in ['.C', '.H']:
                    raise MesonException('The \'language\' argument is only available for *.C and *.H files.')
                return self.language == src.lang_clarification
```

eli-schwartz:
> I think we need to stop bike shedding and approve this PR.

That's an interesting interpretation of "this is a bad idea and we don't want it".

It's not bikeshedding if there is no actual bikeshedding, merely a difference of opinions about whether a feature is fundamentally a good idea to want. People are allowed to disagree about whether it is a good idea, without being accused of frivolous nitpicking over irrelevant and inconsequential side details.

That's all I wanted to say -- that I find your attempted rationale for why this PR should be accepted, to be in bad taste. Maybe there's a good reason to accept it, maybe not -- but this isn't that reason, so why go there?

Volker-Weissmann:
> > I think we need to stop bike shedding and approve this PR.
> 
> That's an interesting interpretation of "this is a bad idea and we don't want it".
> 
> It's not bikeshedding if there is no actual bikeshedding, merely a difference of opinions about whether a feature is fundamentally a good idea to want. People are allowed to disagree about whether it is a good idea, without being accused of _**and dismissed as**_ frivolous nitpicking over irrelevant and inconsequential side details.
> 
> That's all I wanted to say -- that I find your attempted rationale for why this PR should be accepted, to be in bad taste. Maybe there's a good reason to accept it, maybe not -- but this isn't that reason, so why go there?

Ok. Whatever. Feel free to disagree with my bikeshedding argument. My second argument was:

> This PR is fully backward compatible, does not introduce substantial amounts of tech-dept and solves a real-world problem (namely: some projects cannot be built without this patch).


So you need to come up with an argument how this breaks backwards compatibility, or how this introduce tech-dept, or how this does not solve a real-world problem.

My argument is the following:

Arguments for accepting this PR:

I can build my project without having to rename 20.000 files (yes I actually counted them), rename all references to those files and proof read the resulting commit with > 20000 lines to make sure that I don't introduce bugs (for example you cannot just use sed, because 
```
MyClass file
file.Count;
```
should not be replaced by
```
MyClass file
file.cppount;
```
.

Arguments for rejecting this PR:
It adds a total of 26 lines of code. Other than that? None I could think of.

xclaesse:
I'm personally in favor of this PR, seems pretty legit to me to be able to associate extensions to language. GLib used to have .c files that needs to be compiled by objc because of autotools issues, it was an issue when porting to meson until we could drop autotools support. Personally I would not restrict this to `.C`.

otoh, this is not a hill I want to die on.

Volker-Weissmann:
> I'm personally in favor of this PR, seems pretty legit to me to be able to associate extensions to language. GLib used to have .c files that needs to be compiled by objc because of autotools issues, it was an issue when porting to meson until we could drop autotools support. Personally I would not restrict this to `.C`.
> 
> otoh, this is not a hill I want to die on.

It IS restricted to .C and .H!

Volker-Weissmann:
run_mypy.py fails on this branch, but at least on my machine it also fails on master.

dcbaker:
If we're only going to allow .C and .H to be treated as C++, I'd rather have a per-project option set in default_options, ala:
```meson
project(
  ...,
  default_options : ['C_is_cpp=true]`
)
```
Or something like that.

Then if you try to use that option on Windows or mac we can either error out, or at least give a big warning (maybe in general give a big warning) that this might not work if your file system isn't case-sensitive.

That would solve my two biggest concerns, which is:
1. As soon as this lands someone's going to say "Why can't I treat .c files as objc files?
2. what do we do if you're on Windows or macOS which don't have case-sensitive file systems?

dcbaker:
Oh, yeah, I need to fix that mypy thing...

Volker-Weissmann:
> If we're only going to allow .C and .H to be treated as C++, I'd rather have a per-project option set in default_options, ala:
> 
> ```meson
> project(
>   ...,
>   default_options : ['C_is_cpp=true]`
> )
> ```
> 
> Or something like that.
>
> Then if you try to use that option on Windows or mac we can either error out, or at least give a big warning (maybe in general give a big warning) that this might not work if your file system isn't case-sensitive.


Honestly, I don't care. Write whatever you want and add it to this PR, as long as it ends up with meson being able to compile .C cpp-files, I'm fine with that. But make sure that no one rejects the PR by saying "What if I have c code in C files and cpp code in C files in the same project?"
Note that it both gcc and clang will NEVER view a .C as c, even if you pass -std=c90. This means that if you have
```
#ifdef __cplusplus
#error "should be c"
#endif
```
inside `file.C` it is absolutely impossible to compile this with gcc or clang.

> 
> That would solve my two biggest concerns, which is:
> 
>     1. As soon as this lands someone's going to say "Why can't I treat .c files as objc files?

Then we ask him if there is  >10000 files of legacy code that use .c for objc files. If not, we tell him to fuck off. If yes, we make it possible to write `files('file.c', language: 'objc')`.
 
> 
>     2. what do we do if you're on Windows or macOS which don't have case-sensitive file systems?

Correct me if I'm wrong (I know very little about the case insensitivity of Windows and macOS): You cannot build your project on Windows (at least not if you use a case insensitive file system). This is not mesons fault, nothing that meson could do can change that.

dcbaker:
> Correct me if I'm wrong (I know very little about the case insensitivity of Windows and macOS): You cannot build your project on Windows (at least not if you use a case insensitive file system). This is not mesons fault, nothing that meson could do can change that.

> 
Right, but cross platform is *very* important to meson, I get that you're concerned that renaming all of your files is going to be a huge pile of work for you that's error prone. I'm concerned about adding features to meson that are going to be a huge pile of work to maintain and error prone. We also tend to prefer solutions to specific problems than generic ones, so to me, an option you can put in your default_options is better.

> Then we ask him if there is >10000 files of legacy code that use .c for objc files. If not, we tell him to fuck off. If yes, we make it possible to write files('file.c', language: 'objc').

People ask us for user defined functions all the time, and we tell them no all the time. Meson is nice to use because we are willing to say "No, I know that would make your life easier, but that would make meson harder to maintain or harder to use for everyone." I know that can make us unpopular, that's just part of being a maintainer. I've been on the receiving end of that from jpakkane, and from other project maintainers. It sucks, but sometimes saying "Sorry, no." is important for the long term usability or maintainability of a project. I've also maintained projects where we never said no, and most of those projects got replaced because they became unmaintainable and unfixable.

Volker-Weissmann:
> Right, but cross platform is very important to meson

Yes, but we are talking about programs that are already non-cross-platform. If I put
```
#ifndef __linux__
#error "wrong os"
#endif
```
in my program, meson cannot build that cross-platform.

> I get that you're concerned that renaming all of your files is going to be a huge pile of work for you that's error prone. I'm concerned about adding features to meson that are going to be a huge pile of work to maintain and error prone.

It is a +51-7 patch, including the documentation. Yes it is an ugly hack. That is why I noted in the documentation that you should not use *.C files at all. I'm even fine with adding a "don't use *.C files" warning if you use *.C files. But in this case, I think that the amount of legacy code is high enough, that the pros greatly outweigh the benefits. Changing all *.C files in the world to *.cpp is way for work than maintaining a +51-7 patch.


> We also tend to prefer solutions to specific problems than generic ones, so to me, an option you can put in your default_options is better.

As I said, I'm fine with that.

> It sucks, but sometimes saying "Sorry, no." is important for the long term usability or maintainability of a project. 

Absolutely. Sometimes, but not in this case.
You also have to understand that programming is sometimes like being a peace negotiator in the middle east: You tell me "No, .C files are .c." and the guys at openfoam tell me "No, .C files are cpp", and I'm the idiot that is stuck in between. It is easy to say "meson needs to stay maintanable and you should not use .C files for cpp", if you are not the idiot that needs to convince openfoam to rename their legacy code. 



TLDR: A +51-7 hacky patch is less terrible than renaming every cpp *.C file in the world. To me, not supporting legacy .C-cpp files because this are not what you should do is like meson not supporting C and C++ because you should rewrite it in rust. 
Yes, if we could rewrite the whole world from scratch, we would neither need *.C files, nor C/C++. But as long as those legacy code exist, we have to make some(!) reasonable(!) sacrifices to support that legacy code.

docs/markdown/Reference-manual.md
```diff
@@ -837,6 +837,25 @@ Then you can use it in `bar2` like this:
 ```
 
 Meson will then do the right thing.
+Meson uses the file extension to guess the programming language.
+Because .C is sometimes used for 'c' code and sometimes for 'cpp'
+code, you must specify the `language: kwarg` like that:
```
eli-schwartz, line 842: Can, not must?

docs/markdown/Reference-manual.md
```diff
@@ -837,6 +837,25 @@ Then you can use it in `bar2` like this:
...
+```meson
+    files('fileA.C', language: 'cpp')
+    files('fileB.C', language: 'c')
+```
+Writing `files('fileB.C')` defaults to 'c', but doing this is
+deprecated, because in the future it might default to 'cpp', breaking
+your code.
```
eli-schwartz, line 849: No it won't, don't make predictions about meson development policy.

eli-schwartz (changes requested):
While I'm not sure whether to sigh and accept this for ease of adopters, or stick to ideals and say you need to rename your source files... this PR is unacceptably opinionated by documenting that it's reasonable to use `*.C` like this.

Before I can even consider agreeing to this PR, I need:
- the docs to be modified per individual comments
- meson to raise a configure warning if you do use this functionality, stating that it's a bad idea. This conveys to users that they're doing something wrong, which people porting legacy code can of course choose to ignore for the reason stated above, but provides a sharp incentive to people writing new code to NOT do this. It also guarantees that no one can get a warning-free build until they do rename all their files, additionally incentivizing them to not write new code matching the style of old code and thereby compounding the issue.

mensinda:
To contribute to the bikesheding and address the suggestion from @dcbaker: I am against adding a project option for this for two reasons:
- It adds an option that is useless in 99% of use cases
- It should **not** be configurable by the user, since this is a project requrement

To resolve this, I recommend adding a new (boolean) kwarg `C_is_cpp` (or something better, I am not the best with names), which removes the ambiguity since we are apparently not adding arbitrary language overwrites. This could potentially also simplify some of the code.

I also think that we should support this use case since `*.C` is listed as one potential extension for C++ files:
https://isocpp.org/wiki/faq/coding-standards#src-file-ext

dcbaker:
I'm fine with adding .C as c++ as long as:
1. whatever we do doesn't naturally open the door to treating .exe files as rust code (or .c as objective-c, or whatever)
2. we give fair warning that using .C is dangerous, and you shouldn't do it. (because case-insensitive file systems, which are a thing people are trying to add to linux, btw)

docs/markdown/Reference-manual.md
Volker-Weissmann: Ok, If you want.

docs/markdown/Reference-manual.md
Volker-Weissmann: Ok, I can remove that.

Volker-Weissmann:
> While I'm not sure whether to sigh and accept this for ease of adopters, or stick to ideals and say you need to rename your source files... this PR is unacceptably opinionated by documenting that it's reasonable to use `*.C` like this.

Quote from my docs:

> Using .C files for 'cpp' code is discouraged by meson,
> [iso-cpp](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#S-source)
> and [google](https://google.github.io/styleguide/cppguide.html#File_Names).
> 


> 
> Before I can even consider agreeing to this PR, I need:
> 
>     * the docs to be modified per individual comments
> 
>     * meson to raise a configure warning if you do use this functionality, stating that it's a bad idea. This conveys to users that they're doing something wrong, which people porting legacy code can of course choose to ignore for the reason stated above, but provides a sharp incentive to people writing new code to NOT do this. It also guarantees that no one can get a warning-free build until they do rename all their files, additionally incentivizing them to not write new code matching the style of old code and thereby compounding the issue.

I will do that.

Volker-Weissmann:
> To contribute to the bikesheding and address the suggestion from @dcbaker: I am against adding a project option for this for two reasons:
> 
>     * It adds an option that is useless in 99% of use cases

You are right.
> 
>     * It should **not** be configurable by the user, since this is a project requrement

Yes, absolutely.
> 
> 
> To resolve this, I recommend adding a new (boolean) kwarg `C_is_cpp` (or something better, I am not the best with names), which removes the ambiguity since we are apparently not adding arbitrary language overwrites. This could potentially also simplify some of the code.

I would prefer the `language` kwarg. If in the future we e.g. have problems with to other languages or two dialects using the same file extension, we can naturally do this using the `language` quark. Otherwise, we will have a `C_is_cpp` and a `java_is_volkers_java_fork` kwarg. Also `language: 'cpp'`  is easer to read/understand than `C_is_cpp`.
> 
> I also think that we should support this use case since `*.C` is listed as one potential extension for C++ files:
> https://isocpp.org/wiki/faq/coding-standards#src-file-ext

Don't tell ME that. @jpakkane  wanted to support *.C c-files. I don't care about *.C c-files.

Volker-Weissmann:
> I'm fine with adding .C as c++ as long as:
> 
>     1. whatever we do doesn't naturally open the door to treating .exe files as rust code (or .c as objective-c, or whatever)

Actually, I tried to implement exactly that, but realised that gcc does not except *.txt or *.rs or *.whatever files. Funny thing, the rustc compiler actually accepts all file extensions that I tested.
But yes, as I said, this PR is meant to support legacy code. I don't want to create more legacy code.
> 
>     2. we give fair warning that using .C is dangerous, and you shouldn't do it. (because case-insensitive file systems, which are a thing people are trying to add to linux, btw)

Agreed. I will do that.

xclaesse:
If it's limited to .C and .H why don't we just consider them as C++ in meson? The case on the FS does not matter, what matter is the case you use in your `meson.build` file.

xclaesse:
Note that Windows is case insensitive, but that does not mean it does not remember the case you used when creating the file. It's just that you can refer to that file with whatever case you want.

Volker-Weissmann:
> If it's limited to .C and .H why don't we just consider them as C++ in meson? The case on the FS does not matter, what matter is the case you use in your `meson.build` file. It's pretty trivial to make Meson compile `executable('app', 'file.C')` as C++ and `executable('app', 'file.c')` as C, is there any downside of just doing that?

This would be the perfect solution, especially because both clang and gcc treal all *.C files as C++, no mather what meson does.
The problem is that some crazy projects have .C with c-code and [jussi pakkanen](https://github.com/mesonbuild/meson/pull/3415#issuecomment-761902023) wants to support that.

> Note that Windows is case insensitive, but that does not mean it does not remember the case you used when creating the file. It's just that you can refer to that file with whatever case you want.

jpakkane:
One of the things that has been bugging me about this change is how much work would it actually take to rename all source files in a project (including changing the Makefiles). So [I created a script to do that](https://github.com/jpakkane/crename). Turns out it's not all that much. It has been my experience that one of the biggest reasons people oppose changes like these (even though they won't admit it) is that they want to avoid having to do the menial work themselves. If you just do it and tell them that the work is already done and only requires doing one commit, the opposition usually goes away.

Volker-Weissmann:
> One of the things that has been bugging me about this change is how much work would it actually take to rename all source files in a project (including changing the Makefiles). So [I created a script to do that](https://github.com/jpakkane/crename). Turns out it's not all that much. It has been my experience that one of the biggest reasons people oppose changes like these (even though they won't admit it) is that they want to avoid having to do the menial work themselves. If you just do it and tell them that the work is already done and only requires doing one commit, the opposition usually goes away.

Guess what, I also already wrote a script for that (altough your script is written nicer). But your (and mine) script does not work completely: You still need to change all `#include file.C` to `#include file.cpp`. You also need to change all the `#define INC_PATH file.C` to `#define INC_PATH file.cpp`, without changing `var_file.Cabc` to `var_file.cppabc` (yes I tried, and I had this problem). But this is still not enough! You need to track down every reference to `file.C` in your documentation or the documentation of someone else. And even if you did this, you now have additional walk when backporting stuff or moving stuff between different versions. A +52-6 line patch is definitely easier.

I still don't understand why you want to go through all these pain and effort instead of either mergin a 20 line patch, or doing what  xclaesse proposed: dropping support for c-code in .C files (already broken in gcc and clang).

Volker-Weissmann:
> One of the things that has been bugging me about this change is how much work would it actually take to rename all source files in a project (including changing the Makefiles). 

With the same argument, I could ask why it is so hard to rename the *.C files that contain plain c to *.c

Volker-Weissmann:
> One of the things that has been bugging me about this change is how much work would it actually take to rename all source files in a project (including changing the Makefiles). So [I created a script to do that](https://github.com/jpakkane/crename). Turns out it's not all that much. It has been my experience that one of the biggest reasons people oppose changes like these (even though they won't admit it) is that they want to avoid having to do the menial work themselves. If you just do it and tell them that the work is already done and only requires doing one commit, the opposition usually goes away.

Honestly, it isn't really about how much work renaming is, it is more about how with my PR, you can add meson support to a project, by just adding meson.build files, without touching anything existing. It would leave everything else, including the old build system untouched and reverting it would be as simple as deleting all meson.build files.

I think not having to do *any* change, just having to add meson.build files, is a huge organizational advantage, especially if you use the meson build system only experimental or besides another build system. Consider WrapDB for example: With this PR, the wrap/patch are just a few meson.build files, without this PR, the wrap is basically a fork of the originally project that is difficult to maintain. ( Why didn't I think of this argument before? This is a great argument! )

I should note that I don't want to port my project to meson, I want to add meson.build files, but also keep the old build system, because I'm not sure yet if meson is better then the old build system.

If this PR (or something similar) won't get approved, the simplest solution, would probably to tell the users "go use https://github.com/Volker-Weissmann/meson/tree/dot-C-files-are-C++" (this is what I'm currently telling them [here](https://develop.openfoam.com/Development/openfoam/-/issues/1984)), because it is simply easier to do. If it turns out that meson is a good build system for my project, I will maybe rename the C files so that one can use upstream meson. But I think we both agree that this solution is not terrible, but also not nice, because forks are generally a bad idea.

I know that meson does not like hacky solutions and that's a good thing, but when porting something to meson, a few hacks might be necessary at the beginning. Generally, meson makes it hard to add hacky solutions without either codegenning meson.build files or forking meson and thats why many people create Feature Request/PR's with stuff that one might find suboptimal (See also: Every discussion about turing completeness, user defined functions, user defined meson modules, or why meson isn't just a python module.).

jpakkane:
> I still don't understand why you want to go through all these pain and effort instead of either mergin a 20 line patch, or doing what xclaesse proposed: dropping support for c-code in .C files

This is a question that pops up every now and then and there are several reasons:

- Every feature added is an ongoing maintenance burden,
-The more options, toggles, switches and "one-off features" a project has, the less pleasant it becomes to use. Over the years Meson has had requests to add dozens upon dozens of these kinds of things. If we had accepted them all (after all, each one of them is "small and simple"I the project as a whole would now be a worse, more complicated thing as a whole.
- As an example, if we add support for this remapping, people _will_ want to have the same for arbitrary remappings, file bugs and merge requests and then complain when they are nak'ed. This is especially mentally draining for me as the maintainer, because I have to keep disappointing people over again. This is a maintenance burdern as well, even though it is not in any way technical.

Anyhow getting back to the actual issue, a major question is whether there are more C or C++ projects that use `.C` for their extensions. I don't have the numbers. If I had to guess then I'd probably go with C++. Based on that we could change it so that `.C` is taken as a C++ file. However that would still be a breaking change. Projects that use `.C` for C files that can be built with a current Meson would fail to build with future versions. Simply breaking these without warning is not nice, so we'd need to do something like this:

- print a deprecation warning when a project tries to use `.C` files ("this will change in the future, rename your files, it is already broken on gcc and clang, etc")
- have at least one release with this, see if anyone files bugs
- make the actual switch in a different release

Volker-Weissmann:
> > I still don't understand why you want to go through all these pain and effort instead of either mergin a 20 line patch, or doing what xclaesse proposed: dropping support for c-code in .C files
> 
> This is a question that pops up every now and then and there are several reasons:
> 
>     * Every feature added is an ongoing maintenance burden,
>       -The more options, toggles, switches and "one-off features" a project has, the less pleasant it becomes to use. Over the years Meson has had requests to add dozens upon dozens of these kinds of things. If we had accepted them all (after all, each one of them is "small and simple"I the project as a whole would now be a worse, more complicated thing as a whole.
> 
>     * As an example, if we add support for this remapping, people _will_ want to have the same for arbitrary remappings, 

Most compilers don't support arbitrary named input files.

> file bugs and merge requests and then complain when they are nak'ed. This is especially mentally draining for me as the maintainer, because I have to keep disappointing people over again. 

Telling people to fuck of is the fun part :wink: 

> This is a maintenance burdern as well, even though it is not in any way technical.
> 
> 
> Anyhow getting back to the actual issue, a major question is whether there are more C or C++ projects that use `.C` for their extensions. I don't have the numbers. If I had to guess then I'd probably go with C++. 

I would also guess that, especially because gcc and clang reject plain c in .C files.

> Based on that we could change it so that `.C` is taken as a C++ file. However that would still be a breaking change. Projects that use `.C` for C files that can be built with a current Meson would fail to build with future versions. Simply breaking these without warning is not nice, so we'd need to do something like this:
> 
>     * print a deprecation warning when a project tries to use `.C` files ("this will change in the future, rename your files, it is already broken on gcc and clang, etc")
> 
>     * have at least one release with this, see if anyone files bugs
> 
>     * make the actual switch in a different release

Wait a second, I'm adding this warning...

xclaesse:
> I would also guess that, especially because gcc and clang reject plain c in .C files.

I did not know that, seems gcc and clang compiles as C++
```
#ifdef __cplusplus
#error This is compiled as C++
#endif

int main()
{
}
```

```
$ gcc test.C
test.C:2:2: error: #error This is compiled as C++
    2 | #error This is compiled as C++
      |  ^~~~~
$ clang test.C
test.C:2:2: error: This is compiled as C++
#error This is compiled as C++
 ^
1 error generated.
```
But does not link with libc++
```
#include <iostream>

int main()
{
  std::cout << "Hello World" << std::endl;
}
```

```
$ clang test.C
/usr/bin/ld: /tmp/test-bddb2d.o: in function `main':
test.C:(.text+0xa): undefined reference to `std::cout'
$ gcc test.C
/usr/bin/ld: /tmp/ccPtUtcQ.o: in function `main':
test.C:(.text+0x12): undefined reference to `std::cout'
```

So I would be surprised any project actually do .C for C. We had multiple people requesting .C as C++ but did we ever heard anyone requesting .C as C ? IMHO we should just go ahead and just hardcode .C and .H as C++ in meson, it does not even need this `language` kwarg. Currently `.C` is simply not supported at all by meson, so there is no backward compat considerations.

xclaesse:
> Currently .C is simply not supported at all by meson, so there is no backward compat considerations.

I mean here that .C as C is not supported because meson actually build .C as C++, the ABI exposed is C++ and won't link with C code unless you add `extern "C"`.

Volker-Weissmann:
> > Currently .C is simply not supported at all by meson, so there is no backward compat considerations.
> 
> I mean here that .C as C is not supported because meson actually build .C as C++, the ABI exposed is C++ and won't link with C code unless you add `extern "C"`.

Slightly OT, but can you link me to some learning resources about those linking differences of C and C++?

xclaesse:
Simple test case:

test.C
```
int foo()
{
  return 0;
}
```

main.c
```
int foo();

int main(int argc, char *argv[])
{
  return foo();
}
```

meson.build
```
project('test', 'c')

executable('app', 'test.C', 'main.c')
```

Compilation error:
```
ninja: Entering directory `builddir'
[3/3] Linking target app
FAILED: app 
cc  -o app app.p/test.C.o app.p/main.c.o -Wl,--as-needed -Wl,--no-undefined
/usr/bin/ld: app.p/main.c.o: in function `main':
/home/xclaesse/programmation/meson-test/builddir/../main.c:5: undefined reference to `foo'
```

xclaesse:
It means you can't mix .C and .c and if you build a library it will have a C++ ABI. I'm sure there are millions other corner cases where compiling as C++ behind your back can create all kind of issues. I think any sane project would assume .C is C++.

mensinda:
Does anyone actually know a single meson project (or any project) that uses `*.C` as plain C?

jpakkane:
There was at least one person who had a project where all file names were in upper case because the code was originally written in DOS.

Volker-Weissmann:
I wrote the warning jussi wanted. I don't know if this PR, or a new PR is the best place for it, so I created a new branch here: https://github.com/Volker-Weissmann/meson/tree/warn-about-C-H but did not create a new PR.

This is the change:
https://github.com/Volker-Weissmann/meson/commit/2b39e86aa69707f949bd65d56b2027357e50ca99

xclaesse:
TBH I don't see a reason to warn about usage of .C, seems pretty legitimate file extension to me, even if I would personally never use it. Pretty clearly documented everywhere as C++ and all major compilers treat them as such. I really don't understand the debate here, Meson should just silently treat them as C++.

> There was at least one person who had a project where all file names were in upper case because the code was originally written in DOS.

gcc and clang will compile them as C++ anyway, so better make it explicit. Most C code builds as C++ just fine, they probably already rely on that.

jpakkane:
> This is the change: Volker-Weissmann@2b39e86

Do create a new MR with just that commit so it can be merged ASAP. It should not use its own logic for printing only once but instead the builtin `once` keyword argument.

The text is also a bit too direct. It should mention that `.C` is currently treated as a C source but it will be changed to be a C++ source in a future release. This is due to GCC and Clang already treat `.C` files as C++ and thus any files that are plain C but have an `.C` extension may break silently.

> I really don't understand the debate here, Meson should just silently treat them as C++

The problem is that currently we treat it as a C file, and if there is one thing that we've learned is that if you provide something, then someone, somewhere _will_ depend on it. Especially since there was a discussion (granted, only on IRC? several years ago) where someone had an old C project where all filenames were all uppercase. These sorts of changes require a deprecation warning for at least one release.

Volker-Weissmann:
> > This is the change: [Volker-Weissmann/meson@2b39e86](https://github.com/Volker-Weissmann/meson/commit/2b39e86)
> 
> Do create a new MR with just that commit so it can be merged ASAP. It should not use its own logic for printing only once but instead the builtin `once` keyword argument.
> 
I did that: https://github.com/mesonbuild/meson/pull/8249

> > I really don't understand the debate here, Meson should just silently treat them as C++
> 
> The problem is that currently we treat it as a C file, and if there is one thing that we've learned is that if you provide something, then someone, somewhere _will_ depend on it. Especially since there was a discussion (granted, only on IRC? several years ago) where someone had an old C project where all filenames were all uppercase. These sorts of changes require a deprecation warning for at least one release.

As always, there is a relevant xkcd:
https://xkcd.com/1172/

eli-schwartz:
So apparently gcc silently treats .C files as C++, on linux... but what does it do on Windows? Does it depend on how it's registered on the filesystem? Does it depend on how you name it on the command line invocation?

jpakkane:
The underlying issue will be fixed when we change the behaviour of `.C` files. As the proposed feature has a lot of opposition we are not going to add it so closing. Thanks.

Volker-Weissmann:
Everyone who can't wait for the two releases jpakkane talked about, you can use my fork as a temporal solution:

`git clone https://github.com/Volker-Weissmann/meson -b dot-C-files-are-C++`

It will be maintained at least until the two release cycles jpakkane talked about are over and we find a permanent solution in the upstream master branch.
("Maintaining" is easy, it is just a 4 line patch that I need to rebase every week or so)

PR #8242, opened 2021-01-22
Add ARM64EC as a new conceptual cpu type of arm64

mcfi:
ARM64EC is a new ARM64 ABI made by Microsoft. The ARM64EC binaries can be loaded in x64 processes on the latest Windows Insider Preview on ARM64, and they don't need to be emulated for the sake of performance.

To support the ARM64EC build target, a new conceptual arm64 cpu type 'arm64ec' is added. The cpu can be specified in cross files like below to generate msbuild solution/vcxproj files with platform set to ARM64EC.

```
[target_machine]
system = 'windows'
cpu_family = 'aarch64'
cpu = 'arm64ec'
endian = 'little'
```


jpakkane:
Test failures seem to be unrelated.

Issue #5406, opened 2019-05-18
Support generic POSIX C compiler


michaelforney:
When I try to run `meson` with `CC` set to a C compiler it doesn't know about (for instance pcc, tcc, or others), I get the error

	meson.build:1:0: ERROR: Unknown compiler(s): [['pcc']]

However, pcc follows the standard POSIX compiler interface, and is perfectly capable of building the project in question. There are quite a few C compilers in the world, and I think meson should allow the build to continue even if it doesn't know about the particular one the user specified.

Perhaps this could be done by falling back to a GenericPosixCCompiler, which supports the options specified in https://pubs.opengroup.org/onlinepubs/9699919799/utilities/c99.html.

jpakkane:
Even if this was done there would need to be code added to Meson that detects the compiler as something that works in the posix way based on its output. At this point you might as well add it to Meson as it does not take a lot of code, which is basically subclassing and overriding a few methods.

The bigger problem here is one of reliability and support. If we were to add a "posixlike" compiler then it really should work with all of them. We can't really support that since experience has shown that every compiler is its own unique snowflake in some way. One of the main principles of Meson is, and has always been, that we only add functionality that we know to work and can keep supporting in the future[1]. A generic "any compiler" backend does not really qualify for this.

[1] Yes, I do know our record on this is not perfect. But it is what we aim for.

michaelforney:
> Even if this was done there would need to be code added to Meson that detects the compiler as something that works in the posix way based on its output. At this point you might as well add it to Meson as it does not take a lot of code, which is basically subclassing and overriding a few methods.

The difference is that adding support for a particular compiler to meson solves the issue in one case, and adding support for the generic POSIX interface solves it in all cases. pcc and tcc are just two popular examples. Here's a few more:

* https://github.com/libfirm/cparser
* https://github.com/larmel/lacc
* https://sparse.wiki.kernel.org/index.php/Main_Page
* http://www.simple-cc.org/
* https://github.com/rui314/8cc
* https://git.sr.ht/~mcf/cc

The last one is my own C compiler I've been working on. Recently, I tried to use it to build a project using meson, and ended up just rewriting the build system in make, since that was easier than getting meson to work.

[...]

PR #8248, opened 2021-01-24
Add basic support for Tiny C Compiler

ret2libc:
This PR tries to add basic support tor tcc, Tiny C Compiler https://bellard.org/tcc/ .
I was able to use this to compile [Rizin](https://rizin.re) with tcc, however it probably still needs some adjustments because some unit tests still fail.

These are the failing tests: https://gist.github.com/ret2libc/1f386c7230e17a2c31be9ebf2c359523

Any hint on how to improve this support is welcome.

Do you suggest me to add `tcc` to the `mesonbuild/eoan` container and improve `os_comp.yaml` to test it?

### DIFF
--- a/mesonbuild/compilers/c.py
+++ b/mesonbuild/compilers/c.py
@@ -719,3 +719,79 @@ class C2000CCompiler(C2000Compiler, CCompiler):
         if path == '':
             path = '.'
         return ['--include_path=' + path]
+
+class TinyCCompiler(CCompiler):
+    LINKER_PREFIX = '-Wl,'
+
+    def __init__(self, exelist: T.List[str], version: str, for_machine: MachineChoice,
+                 is_cross: bool, info: 'MachineInfo',
+                 exe_wrapper: T.Optional['ExternalProgram'] = None,
+                 linker: T.Optional['DynamicLinker'] = None,
+                 full_version: T.Optional[str] = None):
+        self.id = 'tcc'
+        CCompiler.__init__(self, exelist, version, for_machine, is_cross,
+                           info, exe_wrapper, linker=linker, full_version=full_version)
+        self.can_compile_suffixes.add('s')
+        self.base_options = {
+            OptionKey(o) for o in ['b_ndebug', 'b_staticpic']}
+
+    def get_options(self) -> 'KeyedOptionDictType':
+        opts = CCompiler.get_options(self)
+        key = OptionKey('std', machine=self.for_machine, lang=self.language)
+        opts[key].choices = ['none', 'c99', 'c11']
+        return opts
+
+    def get_always_args(self) -> T.List[str]:
+        return []
+
+    def get_option_compile_args(self, options: 'KeyedOptionDictType') -> T.List[str]:
+        args = []
+        std = options[OptionKey('std', lang=self.language, machine=self.for_machine)]
+        if std.value != 'none':
+            args.append('-std=' + std.value)
+        return args
+
+    def get_dependency_gen_args(self, outtarget: str, outfile: str) -> T.List[str]:
+        return ['-MD', '-MF', outfile]
+
+    def get_optimization_args(self, optimization_level: str) -> T.List[str]:
+        optimization_args = {
+            '0': ['-O0'],
+            'g': ['-O0', '-g'],
+            '1': ['-O1'],
+            '2': ['-O2'],
+            '3': ['-O3'],
+            's': ['-Os'],
+        }  # type: T.Dict[str, T.List[str]]
+        return optimization_args[optimization_level]
+
+    def get_no_optimization_args(self) -> T.List[str]:
+        return self.get_optimization_args('0')
+
+    def get_pic_args(self) -> T.List[str]:
+        return []
+
+    def get_coverage_args(self) -> T.List[str]:
+        return []
+
+    def get_warn_args(self, level: str) -> T.List[str]:
+        warn_args = {
+            '0': [],
+            '1': ['-Wall'],
+            '2': ['-Wall', '-Werror'],
+            '3': ['-Wall', '-Werror', '-Wunsupported', '-Wwrite-strings']
+        }
+        return warn_args[level]
+
+    def get_buildtype_args(self, buildtype: str) -> T.List[str]:
+        return []
+
+    def get_debug_args(self, is_debug: bool) -> T.List[str]:
+        return ['-g'] if is_debug else []
+
+    def compute_parameters_with_absolute_paths(self, parameter_list: T.List[str], build_dir: str) -> T.List[str]:
+        for idx, i in enumerate(parameter_list):
+            if i[:2] == '-I' or i[:2] == '-L':
+                parameter_list[idx] = i[:2] + os.path.normpath(os.path.join(build_dir, i[2:]))
+
+        return parameter_list
--- a/mesonbuild/compilers/__init__.py
+++ b/mesonbuild/compilers/__init__.py
@@ -127,6 +127,7 @@ __all__ = [
     'VisualStudioCCompiler',
     'VisualStudioCPPCompiler',
     'CythonCompiler',
+    'TinyCCompiler',
 ]
 
 # Bring symbols from each module into compilers sub-package namespace
@@ -189,6 +190,7 @@ from .c import (
     CompCertCCompiler,
     C2000CCompiler,
     VisualStudioCCompiler,
+    TinyCCompiler,
 )
 from .cpp import (
     CPPCompiler,
--- a/mesonbuild/compilers/detect.py
+++ b/mesonbuild/compilers/detect.py
@@ -47,6 +47,8 @@ from ..linkers import (
     VisualStudioLinker,
     VisualStudioLikeLinkerMixin,
     WASMDynamicLinker,
+    TinyCDynamicLinker,
+    TinyCArLinker,
 )
 from .compilers import Compiler
 from .c import (
@@ -68,6 +70,7 @@ from .c import (
     CompCertCCompiler,
     C2000CCompiler,
     VisualStudioCCompiler,
+    TinyCCompiler,
 )
 from .cpp import (
     CPPCompiler,
@@ -324,6 +327,8 @@ def detect_static_linker(env: 'Environment', compiler: Compiler) -> StaticLinker
             return CcrxLinker(linker)
         if out.startswith('GNU ar') and ('xc16-ar' in linker or 'xc16-ar.exe' in linker):
             return Xc16Linker(linker)
+        if out.startswith('GNU ar') and compiler.id == 'tcc':
+            return TinyCArLinker(linker)
         if out.startswith('TMS320C2000') and ('ar2000' in linker or 'ar2000.exe' in linker):
             return C2000Linker(linker)
         if out.startswith('The CompCert'):
@@ -395,6 +400,8 @@ def _detect_c_or_cpp_compiler(env: 'Environment', lang: str, for_machine: Machin
         elif 'ccomp' in compiler_name:
             arg = '-version'
         elif 'cl2000' in compiler_name:
+            arg = '-version'
+        elif 'tcc' in compiler_name:
             arg = '-version'
         elif compiler_name in {'icl', 'icl.exe'}:
             # if you pass anything to icl you get stuck in a pager
@@ -620,6 +627,14 @@ def _detect_c_or_cpp_compiler(env: 'Environment', lang: str, for_machine: Machin
             cls = C2000CCompiler if lang == 'c' else C2000CPPCompiler
             env.coredata.add_lang_args(cls.language, cls, for_machine, env)
             linker = C2000DynamicLinker(compiler, for_machine, version=version)
+            return cls(
+                ccache + compiler, version, for_machine, is_cross, info,
+                exe_wrap, full_version=full_version, linker=linker)
+
+        if 'tcc ' in out and lang == 'c':
+            cls = TinyCCompiler
+            env.coredata.add_lang_args(cls.language, cls, for_machine, env)
+            linker = TinyCDynamicLinker(compiler, for_machine, cls.LINKER_PREFIX, [], version=version)
             return cls(
                 ccache + compiler, version, for_machine, is_cross, info,
                 exe_wrap, full_version=full_version, linker=linker)
--- a/mesonbuild/linkers/__init__.py
+++ b/mesonbuild/linkers/__init__.py
@@ -33,6 +33,7 @@ from .linkers import (
     C2000Linker,
     AIXArLinker,
     PGIStaticLinker,
+    TinyCArLinker,
     NvidiaHPC_StaticLinker,
 
     DynamicLinker,
@@ -63,6 +64,7 @@ from .linkers import (
     AIXDynamicLinker,
     OptlinkDynamicLinker,
     CudaLinker,
+    TinyCDynamicLinker,
 
     prepare_rpaths,
     order_rpaths,
@@ -90,6 +92,7 @@ __all__ = [
     'CompCertLinker',
     'C2000Linker',
     'AIXArLinker',
+    'TinyCArLinker',
     'PGIStaticLinker',
     'NvidiaHPC_StaticLinker',
 
--- a/mesonbuild/linkers/linkers.py
+++ b/mesonbuild/linkers/linkers.py
@@ -308,6 +308,10 @@ class AIXArLinker(ArLikeLinker):
     id = 'aixar'
     std_args = ['-csr', '-Xany']
 
+class TinyCArLinker(ArLikeLinker):
+    id = 'ar'
+    # TinyC does not support thin archives
+    std_args = ['csr']
 
 def prepare_rpaths(raw_rpaths: str, build_dir: str, from_dir: str) -> T.List[str]:
     # The rpaths we write must be relative if they point to the build dir,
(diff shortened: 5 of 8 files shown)


ret2libc:
`run_unittests.py::LinuxlikeTests::test_usage_pkgconfig_prefixes`: the TCC code does not add the rights `-L` necessary for TCC to find the libval1.so when looking at libval2.so

ret2libc:
With newer version, these are the remaining failing tests:
```
FAILED run_unittests.py::AllPlatformTests::test_custom_target_changes_cause_rebuild - AssertionError: Lists differ: [] != ['prog']
FAILED run_unittests.py::AllPlatformTests::test_source_changes_cause_rebuild - AssertionError: Lists differ: [] != ['prog']
FAILED run_unittests.py::AllPlatformTests::test_nostdlib - subprocess.CalledProcessError: Command '['/usr/bin/ninja', '-w', 'dupbuild=err', '-d', 'explain']' returned non-zero exit status 1.
FAILED run_unittests.py::AllPlatformTests::test_preprocessor_checks_CPPFLAGS - subprocess.CalledProcessError: Command '['/home/rschiron/projects/meson/venv/bin/python3', '/home/rschiron/p...
FAILED run_unittests.py::LinuxlikeTests::test_compiler_check_flags_order - subprocess.CalledProcessError: Command '['/home/rschiron/projects/meson/venv/bin/python3', '/home/rschiron/proje...
FAILED run_unittests.py::LinuxlikeTests::test_static_link - subprocess.CalledProcessError: Command '['/usr/bin/ninja', '-w', 'dupbuild=err', '-d', 'explain']' returned non-zero exit statu...
FAILED run_unittests.py::LinuxlikeTests::test_usage_pkgconfig_prefixes - subprocess.CalledProcessError: Command '['/usr/bin/ninja', '-w', 'dupbuild=err', '-d', 'explain']' returned non-ze...
```

With the following patch I'm also able to make `test_static_link` pass, however I am not sure whether it is the right fix or not. Tcc does not have `-Wl,--start-group`/`--end-group`, so it is important to put dependencies in order, or so it seems. I would wait a bit before committing this fix to the test because maybe it is caused by other issues in the TCC plugin I wrote:
```diff
diff --git a/test cases/unit/68 static link/meson.build b/test cases/unit/68 static link/meson.build
index dac17f810..aff13cd2b 100644
--- a/test cases/unit/68 static link/meson.build        
+++ b/test cases/unit/68 static link/meson.build        
@@ -22,10 +22,10 @@ test('test3-shared', executable('test3-shared', 'test3.c',
 func7_dep = cc.find_library('func7')
 func9_linkwhole_dep = cc.find_library('func9_linkwhole')
 test('test4-linkwhole', executable('test4-linkwhole', 'test4.c',
-  dependencies : [func7_dep, func9_linkwhole_dep]))
+  dependencies : [func9_linkwhole_dep, func7_dep]))
 func9_linkwith_dep = cc.find_library('func9_linkwith')
 test('test4-linkwith', executable('test4-linkwith', 'test4.c',
-  dependencies : [func7_dep, func9_linkwith_dep]))
+  dependencies : [func9_linkwith_dep, func7_dep]))
 
 # Verify that installed libfunc16.a is usable
 libfunc16_dep = cc.find_library('func16')
```

ret2libc:
By the way, to install Tiny C Compiler you can do:
```
          git clone https://repo.or.cz/tinycc.git
          cd tinycc
          git checkout mob
          chmod +x ./configure
          ./configure --prefix=/usr
          make
          sudo make install
```

dcbaker:
This looks pretty good, I can spot a few little style things, but nothing looks majorly wrong. It would be nice if we a test run with tcc, but I think the most likely place to put one is in Eoan and that's broken if you try to rebuild it, and I'm going to fix that tomorrow.

ret2libc:
> This looks pretty good, I can spot a few little style things, but nothing looks majorly wrong. It would be nice if we a test run with tcc, but I think the most likely place to put one is in Eoan and that's broken if you try to rebuild it, and I'm going to fix that tomorrow.

I'm still failing to understand why tcc is not re-called after touching a header file in test_source_changes_cause_rebuild

jpakkane:
That requires the compiler to support Make style dependency output generation similar to GCC's `-MD` and friends. So either those flags are missing or they get written to the wrong place or something similar.

dcbaker:
You need to implement the `get_dependency_gen_args` method, and maybe the `get_depfile_suffix` method

ghostmansd:
What a funny coincidence. I should have checked this before starting working on the same idea (https://github.com/ghostmansd/meson/tree/tcc). :-) Thanks for your work! Your version seems to be much more complete.

mesonbuild/compilers/c.py
```diff
@@ -710,3 +710,66 @@ def get_include_args(self, path: str, is_system: bool) -> T.List[str]:
...
+        return args
+
+    def get_optimization_args(self, optimization_level: str) -> T.List[str]:
+        return []
+
+    def get_no_optimization_args(self) -> T.List[str]:
+        return []
+
+    def get_pic_args(self) -> T.List[str]:
+        return []
```
ghostmansd, line 752: It seems that -fPIC is not supported. Maybe it's worth throwing a MesonException here?

mesonbuild/linkers.py
```diff
@@ -1372,3 +1372,88 @@ def get_soname_args(self, env: 'Environment', prefix: str, shlib_name: str,
...
+        return self._apply_prefix('--whole-archive') + args + self._apply_prefix('--no-whole-archive')
+
+    def get_coverage_args(self) -> T.List[str]:
+        return []
+
+    def export_dynamic_args(self, env: 'Environment') -> T.List[str]:
+        m = env.machines[self.for_machine]
+        if m.is_windows() or m.is_cygwin():
+            return self._apply_prefix('--export-all-symbols')
+        return self._apply_prefix('-export-dynamic')
```
ghostmansd, line 1398: It seems that the tcc's equivalent of "-Wl,--export-dynamic" is "-rdynamic" option.

https://lists.gnu.org/archive/html/libtool-patches/2014-07/msg00001.html

mesonbuild/compilers/c.py
```diff
@@ -710,3 +710,66 @@ def get_include_args(self, path: str, is_system: bool) -> T.List[str]:
...
+        return opts
+
+    def get_always_args(self) -> T.List[str]:
+        return []
+
+    def get_option_compile_args(self, options: 'KeyedOptionDictType') -> T.List[str]:
+        args = []
+        std = options[OptionKey('std', lang=self.language, machine=self.for_machine)]
+        if std.value != 'none':
+            args.append('-std=' + std.value)
```
ghostmansd, line 742: I've been under impression that -std option is simply ignored with tcc. Please, correct me if I'm wrong.

mesonbuild/compilers/c.py
```diff
@@ -710,3 +710,66 @@ def get_include_args(self, path: str, is_system: bool) -> T.List[str]:
...
+        return args
+
+    def get_optimization_args(self, optimization_level: str) -> T.List[str]:
+        return []
```
ghostmansd, line 746: It looks like that tcc supports -Ox notation, but I'm not sure whether it has some effects other than defining an __OPTIMIZE__ macro.

mesonbuild/compilers/c.py
```diff
@@ -710,3 +710,66 @@ def get_include_args(self, path: str, is_system: bool) -> T.List[str]:
...
+
+    def get_coverage_args(self) -> T.List[str]:
+        return []
+
+    def get_warn_args(self, level: str) -> T.List[str]:
+        warn_args = {
+            '0': [],
+            '1': ['-Wall'],
+            '2': ['-Wall', '-Werror'],
+            '3': ['-Wall', '-Werror', '-Wunsupported', '-Wwrite-strings']
```
ghostmansd, line 762: Maybe '-Wimplicit-function-declaration' should also be here?

ghostmansd: Never mind, it seems that one is included by -Wall already.

mesonbuild/linkers.py
ghostmansd: With tcc 0.9.27, it seems that '-Wl,--export-all-symbols' is the exact equivalent of '-rdynamic'.

mesonbuild/compilers/c.py
```diff
@@ -710,3 +710,66 @@ def get_include_args(self, path: str, is_system: bool) -> T.List[str]:
         if path == '':
             path = '.'
         return ['--include_path=' + path]
+
+class TCCCCompiler(CCompiler):
```
ghostmansd, line 714: It looks like tcc has some support for -MD and -MF options; if I'm not mistaken these can be set at get_dependency_gen_args.

mesonbuild/linkers.py
```diff
@@ -1372,3 +1372,88 @@ def get_soname_args(self, env: 'Environment', prefix: str, shlib_name: str,
...
+
+    def build_rpath_args(self, env: 'Environment', build_dir: str, from_dir: str,
```
ghostmansd, line 1400: This section should likely be re-considered; I'm not sure of SunOS or cygwin parts, quite likely the logic can be simplified.

mesonbuild/linkers.py
```diff
@@ -1372,3 +1372,88 @@ def get_soname_args(self, env: 'Environment', prefix: str, shlib_name: str,
...
+    def __init__(self, for_machine: mesonlib.MachineChoice,
+                 *, version: str = 'unknown version'):
+        super().__init__(['tcc'], for_machine, '-Wl,', [],
+                         version=version)
+
+    def get_allow_undefined_args(self) -> T.List[str]:
+        return []
+
+    def get_pic_args(self) -> T.List[str]:
+        return []
```
ghostmansd, line 1384: cf. previous comment w.r.t. -fPIC

eli-schwartz:
> It would be nice if we a test run with tcc, but I think the most likely place to put one is in Eoan and that's broken if you try to rebuild it, and I'm going to fix that tomorrow.

We don't need Ubuntu Eoan to run tcc unittests, it is packaged on Arch too.

mesonbuild/compilers/c.py
ret2libc: https://man.cx/tcc(1) here it says that -Ox options are ignored. That's why I did not list them here.

mesonbuild/compilers/c.py
ret2libc: I'm not sure...
```
$ tcc --help | grep std
  -std=c99     Conform to the ISO 1999 C standard (default).
  -std=c11     Conform to the ISO 2011 C standard.
```

mesonbuild/linkers.py
ret2libc: From man page:
```
-Wl,--export-all-symbols
-Wl,--export-dynamic
           Export global symbols to the dynamic linker. It is useful when a library opened with "dlopen()" needs to access executable symbols.

```
I'm not sure there is really a need to special case windows/cygwin, I probably copied that code from another linker around :)

mesonbuild/compilers/c.py
ret2libc: I am looking into this, because that is causing some unit tests to fail, however it is not working as I would expect yet.

mesonbuild/compilers/c.py
ghostmansd: I checked the code; it seems that the only thing it affects is __STDC_VERSION__. This, however, seems to be an already sufficient reason to support it. :-)

mesonbuild/compilers/c.py
ghostmansd: This is used to define `__OPTIMIZE__` only.

mesonbuild/linkers.py
ghostmansd: > I probably copied that code from another linker around
That's exactly the way I was doing the same thing. :-)

mesonbuild/compilers/c.py
ret2libc: so you think it makes sense to list the various `-Ox` here? If so I'll add.

mesonbuild/compilers/c.py
ret2libc: Raising an exception here makes *a lot* of unit tests fail. I think even tests that compile shared libraries fail if i put an exception.

mesonbuild/linkers.py
ret2libc: exception here works fine, it seems.

mesonbuild/compilers/c.py
ghostmansd: Well, it's basically the same as for `-std`: the option only defines one macro, but still it works when passed. I think yes, we should be consistent. Perhaps `-Og` must be translated into `-O0 -g`, but I'm not sure.

mesonbuild/linkers.py
ret2libc: I removed the SunOS part, but honestly i don't know how to improve this further without breaking stuff. I copied most of the body from gnu IIRC.

mesonbuild/compilers/c.py
ret2libc: Pushed a version with the -MD, -MF, however it causes other fails because of `assertBuildIsNoop`. Still not sure why.

ret2libc:
> > It would be nice if we a test run with tcc, but I think the most likely place to put one is in Eoan and that's broken if you try to rebuild it, and I'm going to fix that tomorrow.
> 
> We don't need Ubuntu Eoan to run tcc unittests, it is packaged on Arch too.

Is there already a way to run unit tests for a particular compiler that does not involve Eoan? I just see a matrix for different compilers in os_comp.yml, but it is only on eoan. The `arch` job in os_comp.yml does not seem to allow different CC for now... Just wanted to know if there is already a way to do this, otherwise I will adjust some jobs.

ret2libc:
> You need to implement the `get_dependency_gen_args` method, and maybe the `get_depfile_suffix` method

@dcbaker I have implemented the method in https://github.com/mesonbuild/meson/pull/8248/files#diff-7ada83df1272b6eaaff2fea8bbbb14ff9cd6baaf7de40c532fe95acf9bf73fbcR745, however now almost all tests that use `assertBuildIsNoop` fail because the test says stuff are recompiled when they should not. However, if I manually try to build those test directories and run ninja two times without touching any file, it doesn't recompile anything for me. Any idea where the problem might be?

ret2libc:
> > > It would be nice if we a test run with tcc, but I think the most likely place to put one is in Eoan and that's broken if you try to rebuild it, and I'm going to fix that tomorrow.
> > 
> > 
> > We don't need Ubuntu Eoan to run tcc unittests, it is packaged on Arch too.
> 
> Is there already a way to run unit tests for a particular compiler that does not involve Eoan? I just see a matrix for different compilers in os_comp.yml, but it is only on eoan. The `arch` job in os_comp.yml does not seem to allow different CC for now... Just wanted to know if there is already a way to do this, otherwise I will adjust some jobs.

BTW, there is problem in using tcc from arch linux. That's the release of tcc, which is from about 3 years ago and it was not updated since then. There are bugs that are only fixed in the `mob` branch on https://repo.or.cz/tinycc.git, where the development seems to be right now. For example, when using `-pthread` the `main` function is not identified, and probably other issues. I think it's better to test with updated tcc from the `mob` branch if that needs to be done.

dcbaker:
I bet @eli-schwartz can fix Arch ;)

eli-schwartz:
https://repo.or.cz/h/mob.html

This is unstable development crowd-sourced by allowing fully anonymous push access to the world. The "mob" user carries no guarantees of usability, safety, or "doesn't contain malware".

So I rather suspect this would need to be carefully vetted before being packaged. :/ Does the tcc community / project leads have the intention of nominating some state in that branch as being known-good, able to graduate to the master branch, and maybe being tagged as a release (even a beta release)?

ret2libc:
> https://repo.or.cz/h/mob.html
> 
> This is unstable development crowd-sourced by allowing fully anonymous push access to the world. The "mob" user/branch carries no guarantees of usability, safety, or "doesn't contain malware".
> 
> So I rather suspect this would need to be carefully vetted before being packaged. :/ Does the tcc community / project leads have the intention of nominating some state in that branch as being known-good, able to graduate to the master branch, and maybe being tagged as a release (even a beta release)?

I don't know :( I will send out an email on the mailing list, if there's anyone around, asking what are their intention.

ret2libc:
Only failing tests now:
* AllPlatformTests::test_preprocessor_checks_CPPFLAGS
* AllPlatformTests::test_nostdlib
* LinuxlikeTests::test_compiler_check_flags_order
* LinuxlikeTests::test_usage_pkgconfig_prefixes
* LinuxlikeTests::test_static_link

mesonbuild/compilers/c.py
```diff
@@ -710,3 +710,79 @@ def get_include_args(self, path: str, is_system: bool) -> T.List[str]:
...
```
dcbaker, line 714: I know this is kinda pedantic, but can we call this `TinyCCmopiler`? Having 4 `C`s in a row is hard to read, and when you expand it makes no sense `TinyCCompilerCCompiler` :)

dcbaker:
The compiler check order one is testing something that the `CCompileArgs` class does IIRC. You may need to implement one of those for TCC to pass that.

XVilka:
@ret2libc was there any answer to your question in the TinyCC mailing list? Could you maybe invite their developers here, to this PR to take a look?

Also, could you please rebase it on top of the current master?

mesonbuild/compilers/c.py
```diff
@@ -710,3 +710,79 @@ def get_include_args(self, path: str, is_system: bool) -> T.List[str]:
...
+    def get_optimization_args(self, optimization_level: str) -> T.List[str]:
+        optimization_args = {
+            '0': [],
+            'g': ['-O0', '-g'],
+            '1': ['-O1'],
+            '2': ['-O2'],
+            '3': ['-O3'],
+            's': ['-Os'],
+        }  # type: T.Dict[str, T.List[str]]
+        return optimization_args[optimization_level]
```
earboxer, line 757: > Note: GCC options -Ox, -fx and -mx are ignored.

https://bellard.org/tcc/tcc-doc.html#Option-summary

Looking at the tinycc source, the optimizations flag is parsed (as one would expect according to `tcc -hh`), but it doesn't seem to be used at all (as one would infer from the documentation).

earboxer:
> There are bugs that are only fixed in the mob branch on https://repo.or.cz/tinycc.git, where the development seems to be right now. For example, when using -pthread the main function is not identified, and probably other issues. I think it's better to test with updated tcc from the mob branch if that needs to be done.

> Does the tcc community / project leads have the intention of nominating some state in that branch as being known-good, able to graduate to the master branch, and maybe being tagged as a release (even a beta release)?

> was there any answer to your question in the TinyCC mailing list? 

There has not been a reply yet.

https://lists.nongnu.org/archive/html/tinycc-devel/2021-01/msg00221.html

tcc builds from the mob branch include the git hash in their version string with a prefix of "mob:"

```sh
./tcc -version # Latest mob branch
tcc version 0.9.27 mob:e97e108 (x86_64 Linux)
tcc -version # The official release from December 2017
tcc version 0.9.27 (x86_64 Linux)
```

This could be used to differentiate between versions: if it doesn't have the "mob" string in it, then it has XYZ known bugs/missing features.

But I think I agree with @TAAPArthur [who says](https://github.com/mesonbuild/meson/pull/8692#issuecomment-837377727)

> You know what you are getting into when dealing with development branches

It's nice to see some attempts toward https://github.com/mesonbuild/meson/issues/5406

averms:
Ubuntu and Debian package a git snapshot from 2020-08-14. Nix (the unstable channel) packages a snapshot from 2021-10-09. The 2020-08-14 snapshot has `-MM` and `-MF` and the 2021-10-09 snapshot has `-MD` and `-MMD` too.

@ret2libc are you still working on this? I would love to help when I get some free time.

ret2libc:
> Ubuntu and Debian package a git snapshot from 2020-08-14. Nix (the unstable channel) packages a snapshot from 2021-10-09. The 2020-08-14 snapshot has `-MM` and `-MF` and the 2021-10-09 snapshot has `-MD` and `-MMD` too.
> 
> @ret2libc are you still working on this? I would love to help when I get some free time.

I am focused on other things lately, so I don't have much time for moving this forward in the near future. AFAIR it should not require too much work to complete this PR though.

dcbaker:
There's a few things to rebase, the compiler and linker detection has moved out of environment and into compilers/detect.py and linkers/detect.py, respectively. That looks scary, but should actually be pretty simple, JFYI

ret2libc:
@averms did you already start working on this? Asking because I'm now having troubles with this branch because I need some newer feature and since I would have to rebase anyway I guess it would make sense to move this forward, but I'd like to avoid double work.

averms:
> @averms did you already start working on this? Asking because I'm now having troubles with this branch because I need some newer feature and since I would have to rebase anyway I guess it would make sense to move this forward, but I'd like to avoid double work.

No I have not.

ret2libc:
I'm fixing/skipping some tests... However, some of these tests require the `mob` version of tcc (which could change every day...). I'd like to know whether support for such a compiler would be accepted or not. In particular, I'm not sure I can add tests in CI for this, because arch has the 5-years old tcc.

Please help me to find the way forward.

XVilka:
So what was decided? It would be nice to have that mainline, after all everyone who wants to use TCC knows what they are dealing with.

ret2libc:
> So what was decided? It would be nice to have that mainline, after all everyone who wants to use TCC knows what they are dealing with.

Problem being that it is hard to even do some CI testing for it, because no distro ships it in the repos and you'd have to track a git repository that can change/break at any time, accessible by anyone.

eli-schwartz:
> While tcc commits at the mob branch do get reverted occasionally (by the maintainer),

Indicating that the maintainer is still active; indeed, I notice that the person who tagged the last release has committed to `mob` as recent as 2 months ago (no real activity this month anyway).

That being the case, I'm extremely curious to know -- what are the chances of the maintainer(s) being willing to periodically "bless" the mob branch by pushing known-good revisions to master?

I mentioned this above:

> So I rather suspect this would need to be carefully vetted before being packaged. :/ Does the tcc community / project leads have the intention of nominating some state in that branch as being known-good, able to graduate to the master branch, and maybe being tagged as a release (even a beta release)?

avih:
> what are the chances of the maintainer(s) being willing to periodically "bless" the mob branch by pushing known-good revisions to master?

TL;DR: impossible to tell.

This is basically a release. The last one was was some years ago, and mob had many commits since. It's been reqested to make a new release few times in recent years.

I also think soon would be a good time for a new release, but I can't speak for the maintainer.

If you want a "strategy" to choose a "safe" commit, I'd look for a commit by the maintainer (grischka), possibly with few further fixup commits, if there are any (he does tend to touch a lot of things in his commits, so sometimes it needs some followups).

XVilka:
For everyone caring about TinyCC (`tcc`) support (https://github.com/mesonbuild/meson/pull/8248), they seem to be closer to start crafting releases and recently bumped the version: https://github.com/TinyCC/tinycc/commit/086870addd9599ef74bed8e81ebb353da7836709

saidwho12:
Add support for TCC please this compiler is very fast

PR #8249, opened 2021-01-25
Warn about .C and .H files

Volker-Weissmann:
See https://github.com/mesonbuild/meson/pull/8239

### DIFF
--- a/mesonbuild/mesonlib/universal.py
+++ b/mesonbuild/mesonlib/universal.py
@@ -348,8 +348,20 @@ class FileMode:
             perms |= stat.S_ISVTX
         return perms
 
+already_warned_about_C_H = False
 class File:
     def __init__(self, is_built: bool, subdir: str, fname: str):
+        global already_warned_about_C_H
+        if not already_warned_about_C_H:
+            if fname.endswith(".C") or fname.endswith(".H"):
+                already_warned_about_C_H = True
+                mlog.warning("""You are using .C or .H files in your project. Do not do that.
+         If it is plain C code, use .h and .c instead, to avoid confusion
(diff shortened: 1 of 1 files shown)


jpakkane:
The error message does not get printed in the way you'd expect it to:

```
WARNING: You are using .C or .H files in your project. This is deprecated.
         Currently, Meson treats this as C code, but this
            might change in the future, breaking your build.
         You code also might be already broken on gcc and clang.
         See https://github.com/mesonbuild/meson/pull/8239 for the discussions.
```

The indentation is there. The simplest way to fix this would be to make the error message a top level constant rather than having it inside the method.

Volker-Weissmann:
> The error message does not get printed in the way you'd expect it to:
> 
> ```
> WARNING: You are using .C or .H files in your project. This is deprecated.
>          Currently, Meson treats this as C code, but this
>             might change in the future, breaking your build.
>          You code also might be already broken on gcc and clang.
>          See https://github.com/mesonbuild/meson/pull/8239 for the discussions.
> ```
> 
> The indentation is there. The simplest way to fix this would be to make the error message a top level constant rather than having it inside the method.

This is precisely the indentation that I wanted.

jpakkane:
Actually, thinking further this might be ok (because the length of warning is always the same as Meson is not localised) but it should still be moved to a top level variable, having text strings with weird indentation inside Python code blocks looks unclean.

Volker-Weissmann:
> Actually, thinking further this might be ok (because the length of warning is always the same as Meson is not localised) but it should still be moved to a top level variable, having text strings with weird indentation inside Python code blocks looks unclean.

Done

PR #8250, opened 2021-01-25
test: Make timeout <= 0 infinite duraction

xclaesse:


docs/markdown/Reference-manual.md
```diff
@@ -119,7 +119,9 @@ the following:
   environment juggling. *(since 0.52.0)* A dictionary is also accepted.
 - `exe_wrapper`: a list containing the wrapper command or script followed by the arguments to it
 - `gdb`: if `true`, the tests are also run under `gdb`
-- `timeout_multiplier`: a number to multiply the test timeout with
+- `timeout_multiplier`: a number to multiply the test timeout with.
+  *Since 0.57* if timeout_multiplier is `<= 0` the test has infinite duration,
+  previous versions of Meson were failing the test with a timeout immediately.
```
dcbaker, line 124:
```suggestion
  *Since 0.57* if timeout_multiplier is `<= 0` the test has infinite duration,
  in previous versions of Meson the test would fail with a timeout immediately.
```
Writing this in the past perfect feels more natural

docs/markdown/Reference-manual.md
```diff
@@ -1730,7 +1732,8 @@ test(..., env: nomalloc, ...)
 
 - `timeout`: the amount of seconds the test is allowed to run, a test
   that exceeds its time limit is always considered failed, defaults to
-  30 seconds
+  30 seconds. *Since 0.57* if timeout is `<= 0` the test has infinite duration,
+  previous versions of Meson were failing the test with a timeout immediately.
```
dcbaker, line 1736: Same thing here.

eli-schwartz:
While I agree it makes sense to support -1 as equivalent to 0, I don't think we should encourage people to do this with random values that all do the same thing. Maybe we could document only 0, and for <0 we could raise a warning "you probably meant 0, proceeding as though 0 was used"?

mesonbuild/mtest.py
```diff
@@ -492,7 +492,7 @@ def emit_progress(self, harness: 'TestHarness') -> None:
             spaces=' ' * TestResult.maxlen(),
             dur=int(time.time() - self.progress_test.starttime),
             durlen=harness.duration_max_len,
-            timeout=int(self.progress_test.timeout))
+            timeout=int(self.progress_test.timeout or -1))
```
dcbaker, line 495: This will replace 0 with -1, that's the desired behavior right?

dcbaker (approved):
A couple of suggestions on the wording in the docs, and one question. But otherwise this looks good.

mesonbuild/mtest.py
xclaesse: It can't be 0 at this point, timeout is either >0 or None. This is just the string that display progress, for infinite timeout it would be printed as `5/-1s`.

xclaesse:
> While I agree it makes sense to support -1 as equivalent to 0, I don't think we should encourage people to do this with random values that all do the same thing. Maybe we could document only 0, and for <0 we could raise a warning "you probably meant 0, proceeding as though 0 was used"?

I agree it does not make much sense to have `timeout: -42` but `-1` is pretty common so I wanted to support at least that. I don't think it's worth being too pedantic and warn for `<-1`.

dcbaker:
Doing something with values less that -1 seems like a good idea (less undefined behavior is always better in my book). I'm fine with the current behavior, but if not that then just raising an error for < -1 would be my second choice.

xclaesse:
@dcbaker agreed.

CI failure is unrelated, happens in master too. LGTM.

PR #8253, opened 2021-01-26
do not require git installed in order to use wraps

### LINKED ISSUE #2623: Subprojects don't work when git is missing on the system
I was trying to use meson on msys2 on Windows. Some things seemed to work, but meson reported
```
Dependency GTest found: NO
Also couldn't find a fallback subproject in subprojects/gtest for the dependency gtest

Meson encountered an error in file test/unittest/meson.build, line 1, column 0:
Dependency "gtest" not found
```
despite using
```meson
gtest = dependency('gtest', fallback : ['gtest', 'gtest_main_dep'], main : true)
```
and having a valid `subprojects/gtest.wrap` file:
```
[wrap-file]
directory = googletest-release-1.8.0

source_url = https://github.com/google/googletest/archive/release-1.8.0.zip
source_filename = gtest-1.8.0.zip
source_hash = f3ed3b58511efd272eb074a3a6d6fb79d7c2e6a0e374323d1e6bcbcc1ef141bf

patch_url = https://wrapdb.mesonbuild.com/v1/projects/gtest/1.8.0/4/get_zip
patch_filename = gtest-1.8.0-4-wrap.zip
patch_hash = 0b90fe055acbdb002a37dfb035184b306008b763931158497ef5dbaa8c7925af
```

After some debugging it turned out that `resolved = r.resolve(dirname)` was raising a `FileNotFoundError`. Further debugging revealed that the following line was a culprit
```python
def quiet_git(cmd, workingdir):
    pc = subprocess.Popen(['git', '-C', workingdir] + cmd,
                          stdout=subprocess.PIPE, stderr=subprocess.PIPE)
    ...
```
which was called from `resolve_git_submodule()` from `resolve()`.

Btw, since I have put some more debugging info to the sources, here's an interesting error that was properly intercepted:
```
fatal: Not a git repository (or any parent up to mount point /z)\nStopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).\n
```

After I installed `git`, everything started working (note that my code was not even in a git repository).

I believe that `meson`'s functionality should not depend on existence of `git` and that the failure to find `git` should be handled properly.

eli-schwartz:
It doesn't make sense to check for the presence of git every time we use it, but short-circuit any attempt to use a wrap right from the get-go because we are trying to be fancy with submodules.

If git is not installed, simply do not try to figure out whether the wrap is a submodule that can potentially be checked out/updated for the user. Just take it on faith that it isn't one.

Fixes #2623

### DIFF
--- a/mesonbuild/wrap/wrap.py
+++ b/mesonbuild/wrap/wrap.py
@@ -353,8 +353,10 @@ class Resolver:
             raise WrapException(m)
 
     def resolve_git_submodule(self) -> bool:
+        # Is git installed? If not, we're probably not in a git repository and
+        # definitely cannot try to conveniently set up a submodule.
         if not GIT:
-            raise WrapException('Git program not found.')
+            return False
         # Are we in a git repository?
         ret, out = quiet_git(['rev-parse'], self.subdir_root)
         if not ret:


dcbaker:
This seems reasonable to me, the other case is what if you're using hg or svn, not git?

eli-schwartz:
The actual type of wrap always triggers checks, e.g. `get_git()`, `get_hg()`, `get_svn()` check once more for the program.

This check comes first and tries to check whether your own project's VCS has the wrap additionally checked into git (in which case the submodule gets fetched, and the wrap is never downloaded).

I guess people using an hg wrap might feel a bit steamy about being told they need to install git in order to use wraps. :D

xclaesse:
LGTM, the failing CI seems unrelated.

PR #8256, opened 2021-01-26
intepreter: Allow using file objects for the script_name of add_*_script

dcbaker:
It's a bit silly and conveluted to have to call find_program on the
output of configure_file, so let's just allow passing files as the
script name.


dcbaker:
This also fixes a stack trace when passing a `File` in that position

PR #8257, opened 2021-01-26
xcode-backend: fix include paths

jmoguillansky-gpsw:
Add include paths from dependencies

### DIFF
--- a/mesonbuild/backend/xcodebackend.py
+++ b/mesonbuild/backend/xcodebackend.py
@@ -734,7 +734,9 @@ class XCodeBackend(backends.Backend):
                 else:
                     product_name = target.get_basename()
                 ldargs += target.link_args
+                cargs = []
                 for dep in target.get_external_deps():
+                    cargs += dep.get_compile_args()
                     ldargs += dep.get_link_args()
                 ldstr = ' '.join(ldargs)
                 valid = self.buildconfmap[target_name][buildtype]
@@ -751,6 +753,7 @@ class XCodeBackend(backends.Backend):
                     args = pargs + gargs + targs
                     if args:
                         langargs[langnamemap[lang]] = args
+                langargs['C'] += cargs
                 symroot = os.path.join(self.environment.get_build_dir(), target.subdir)
                 self.write_line('%s /* %s */ = {' % (valid, buildtype))
                 self.indent_level += 1


mesonbuild/backend/xcodebackend.py
```diff
@@ -751,6 +753,8 @@ def generate_xc_build_configuration(self):
                     args = pargs + gargs + targs
                     if args:
                         langargs[langnamemap[lang]] = args
+                cargs = list(dict.fromkeys(cargs)) # Remove duplicates, and preserve order
```
dcbaker, line 756: `dict.fromkeys()` isn't guaranteed to preserve order in python3.6, please use `mesonlib.OrderedSet` for this, it's also what we do elsewhere.

Also, is this actually safe? Sometimes arguments can be given multiple times, but actually matter, like inside a `--start-group/--end-group`.

mesonbuild/backend/xcodebackend.py
```diff
@@ -722,6 +722,7 @@ def generate_xc_build_configuration(self):
                 if links_dylib:
                     dep_libs = ['-Wl,-search_paths_first', '-Wl,-headerpad_max_install_names'] + dep_libs
                 dylib_version = None
+                cargs = []
```
dcbaker, line 725: I know it's a nit, but can you put this closer to where it's used?

dcbaker (changes requested):

mesonbuild/backend/xcodebackend.py
jmoguillansky-gpsw: thanks, you're right. Updated.  sorry for the delay

mesonbuild/backend/xcodebackend.py
jmoguillansky-gpsw: updated

mesonbuild/backend/xcodebackend.py
```diff
@@ -751,6 +753,7 @@ def generate_xc_build_configuration(self):
                     args = pargs + gargs + targs
                     if args:
                         langargs[langnamemap[lang]] = args
+                langargs['C'] += cargs
```
dcbaker, line 756: Should this really go into `langargs['C']` and not into `langargs[langnamemap[lang]]`?

Sorry I just noticed this.

dcbaker:
Other than that this look good to me.

mesonbuild/backend/xcodebackend.py
jmoguillansky-gpsw: we probably want to set these flags for C and C++.  I think the main concern is not to set these flags for other languages like objective-c, etc.

jmoguillansky-gpsw: https://developer.apple.com/library/archive/documentation/DeveloperTools/Reference/XcodeBuildSettingRef/1-Build_Setting_Reference/build_setting_ref.html

xcode spec has "OTHER_CFLAGS" and "OTHER_CPLUSPLUSFLAGS".

Default value of "OTHER_CPLUSPLUSFLAGS" is "OTHER_CFLAGS" so these flags should apply for C and C++

dcbaker: okay, That's probably fine for now then. If we run into a case where we need to apply them to ObjC we can handle that later.

dcbaker (approved):

dcbaker:
Thanks, merged.

PR #8258, opened 2021-01-26
Change OS Comp CI to be ubuntu rollling instead fo Eoan

dcbaker:
Eoan is not longer in the repos, and we don't really want to hard code to a specific release anyway, we want to be using the latest release, whether that's an LTS or not. Ubuntu provides a 'rolling' tag for this purpose in their docker images.

This also makes the CI less aggressive about starting, which means it doesn't start for docs only changes, so we have one less reason to add `[skip ci]` tags.


dcbaker:
Everything seems to be working except hotdoc, which the gi modules fails in. When I try to import that locally it says "no such module as libc++.so.6" (after i hack the crap out of it not to hide all errors). I'm not sure if that's what's happening on arch and fedora, nixos can introduce these sorts of problems itself.

dcbaker:
Okay, with enough forcing I have this working on nix. Gonna try the same on fedora and arch

eli-schwartz:
Sounds like quite the weird error. Also reminds me I really wanna just package this for Arch.

xclaesse:
I'm personally really against doing anything rolling in CI. Your env must be as frozen as possible. I don't think we want our CI breaking at random moment because Ubuntu decided to push a new release. That being said, we should of course drop Eoan and setup CI for 20.10. When 21.04 gets released we could drop 20.10 and setup 21.04, I don't think it's worth to keep CI for old non-LTS ubuntu releases.

eli-schwartz:
The Arch tests are rolling :p but I hear where you are coming from w.r.t. Ubuntu rolling vs. freezing.

Particularly, that one rolling CI env is probably all one really needs.

dcbaker:
I was told that the Eoan was supposed to be "the latest release of Ubuntu", it's almost as old as the LTS! clearly we're not keeping it updated.

We're also running fedora:latest and arch, which is rolling, and opensuse latest. so really ubuntu is the outlier here.

dcbaker:
everything seems to be passing now!

jpakkane:
\o/

Issue #8187, opened 2021-01-12
Executables with DLL deps passed to add_install_script() can't run on Windows


nirbheek:
If you have an `executable()` that has DLL deps in other builddir sub-dirs, then it can't be run by `meson.add_install_script()` on Windows because we do not do the same magic that we do for `custom_target()`, `test()`, etc to setup `PATH` to all the DLL dependencies of that executable.

Specifically, in `mesonbuild/backend/backends.py` we need to edit `Backend.generate_custom_install_script()` and have it do the same magic as `Backend.generate_custom_target()`. That is, it needs to call `self.as_meson_exe_cmdline()` which contains the necessary machinery, for instance it calls `self.determine_windows_extra_paths()`.

tp-m:
This causes issue on Windows with fontconfig's `fc-cache` tool running as part of the install procedure (where it then can't find the libs it needs), see https://gitlab.freedesktop.org/fontconfig/fontconfig/-/issues/260 (or https://gitlab.freedesktop.org/gstreamer/gst-build/-/issues/144)

Neumann-A:
I can also confirm this being also a problem in GTK running glib parts in a post install script. My solution currently is running it manually.

PR #8259, opened 2021-01-26
Fix executable as script on Windows

xclaesse:
On Windows this would fail because of missing DLL:
```
mylib = library(...)
exe = executable(..., link_with: mylib)
meson.add_install_script(exe)
```

The reason is on Windows we cannot rely on rpath to find libraries from
build directory, they are searched in $PATH. We already have all that
mechanism in place for custom_target() using ExecutableSerialisation
class, so reuse it for install/dist/postconf scripts too.

This has bonus side effect to also use exe_wrapper for those scripts.

Fixes: #8187

### DIFF
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -22,9 +22,10 @@ import os
 import pickle
 import re
 import shlex
-import subprocess
 import textwrap
 import typing as T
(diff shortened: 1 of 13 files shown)


xclaesse:
Failing CI seems unrelated, I think it's ready for review.

PR #8260, opened 2021-01-26
azure: be less agressive about starting piplines

dcbaker:


dcbaker:
This was just for testing some changes to the pipelines configuraiton.

PR #8262, opened 2021-01-27
Can read project version from a file.

### LINKED ISSUE #7890: Simplify version extraction from a file

jpakkane:
Fixes #7890

Does not close #7931 as it is about reading file contents directly.


mesonbuild/interpreter.py
```diff
@@ -3169,9 +3169,24 @@ def func_project(self, node, args, kwargs):
...
+            if len(version) != 1:
+                raise InvalidCode('Version argument is an array with more than one entry.')
+            version = version[0]
+        if isinstance(version, mesonlib.File):
+            self.add_build_def_file(version)
+            ifname = version.absolute_path(self.environment.source_dir,
+                                           self.environment.build_dir)
+            try:
+                with open(ifname, 'r') as versionfile:
+                    self.project_version = versionfile.readline().strip()
```
eli-schwartz, line 3183: Should we be allowing version files that have additional non-version lines in it? Yes, it's documented as such, but there does not seem to be a good reason to allow it, and IMHO it is confusing.

jpakkane: Either one is fine by me.

dcbaker: For mesa our version file is: `21.0.0-rc4\n`. I'd prefer to error if there's more than one line, we can always change it later if there's a compelling reason, and it's a pretty trivial change for a project if they have more than one line.

mesonbuild/interpreter.py
```diff
@@ -3169,9 +3169,24 @@ def func_project(self, node, args, kwargs):
...
         self.active_projectname = proj_name
-        self.project_version = kwargs.get('version', 'undefined')
-        if not isinstance(self.project_version, str):
-            raise InvalidCode('The version keyword argument must be a string.')
+        version = kwargs.get('version', 'undefined')
+        if isinstance(version, list):
+            if len(version) != 1:
+                raise InvalidCode('Version argument is an array with more than one entry.')
+            version = version[0]
+        if isinstance(version, mesonlib.File):
```
dcbaker, line 3177: `elif`?

jpakkane: Nope. That if just unpacks the item from the array. It is processed later. (`files` returns an array of files)

mesonbuild/interpreter.py
```diff
@@ -3169,9 +3169,28 @@ def func_project(self, node, args, kwargs):
...
+            self.add_build_def_file(version)
+            ifname = version.absolute_path(self.environment.source_dir,
+                                           self.environment.build_dir)
+            try:
+                ver_data = Path(ifname).read_text(encoding='utf-8').split('\n')
+            except FileNotFoundError:
+                raise InterpreterException('Version file not found.')
+            if len(ver_data) == 2 and ver_data[1] == '':
+                ver_data = ver_data[0:1]
```
eli-schwartz, line 3186: I would just `.rstrip('\n').split('\n')`, verify that it's exactly len() 1, and use that.

mesonbuild/interpreter.py
dcbaker: Oh, yeah, I see.

mesonbuild/interpreter.py
jpakkane: That accepts files that have multiple newlines at the end, which arguably should be rejected.

xclaesse: If your line ends with `\r\n` I think the `\r` would be included in the version string, no?

xclaesse:
```
>>> print('a\r\nb'.split('\n'))
['a\r', 'b']
>>> print('a\r\nb'.splitlines())
['a', 'b']
```

eli-schwartz: read_text() uses universal_newlines

mesonbuild/interpreter.py
```diff
@@ -3169,9 +3169,28 @@ def func_project(self, node, args, kwargs):
...
         self.active_projectname = proj_name
-        self.project_version = kwargs.get('version', 'undefined')
-        if not isinstance(self.project_version, str):
-            raise InvalidCode('The version keyword argument must be a string.')
+        version = kwargs.get('version', 'undefined')
+        if isinstance(version, list):
+            if len(version) != 1:
+                raise InvalidCode('Version argument is an array with more than one entry.')
+            version = version[0]
+        if isinstance(version, mesonlib.File):
```
xclaesse, line 3177: Missing FeatureNew()

mesonbuild/interpreter.py
xclaesse: oh didn't knowm good.

eli-schwartz: https://docs.python.org/3/library/functions.html#open-newline-parameter

mesonbuild/interpreter.py
jpakkane: Added. This is the first one of these that I have added, hopefully I got it right.

dcbaker: looks good.

PR #8275, opened 2021-01-30
Introduce `fs.read` to read a file as a string

ldrumm:
Following #7931 , this patch introduces the ability to read the contents
of a file to the fs module.

This patch introduces the ability to read files at configure time, but
has some restrictions:
    - binary files are not supported (I don't think this will prove a
    problem, and if people are wanting to do something with binary
    files, they should probably be shelling out to their own script).
    - Only files outside the build directory allowed. This limitation
      should prevent build loops.
Given that reading an arbitrary file at configure time can affect the
configuration in almost arbitrary ways, meson should force a reconfigure
when the given file changes. This is non-configurable, but this can
easily be changed with a future keyword argument.


ldrumm:
In #7931 it was suggested that `allow_absolute` is pointless and should be removed. I think that makes sense and will remove it from the present patch if a maintainer agress

jpakkane:
Go ahead and remove it.

The doc changes have several places where Meson is written in lowercase. It is a proper name, it should be written with a capital letter (unless you are referring to the command `meson` or somesuch).

jpakkane:
> mesonbuild/modules/fs.py:255:9: error: Call to untyped function "add_build_def_file" in typed context

This means that you need to add annotation to that function. This should be only thing you need, it should not cascade any further. You can run the test locally by running the `run_mypy.py` script.

docs/markdown/howtox.md
```diff
@@ -139,6 +139,23 @@ cdata.set('SOMETHING', txt)
...
+`The [fs module](#Fs-modules) offers the `read` function` which enables adding
+the contents of arbitrary files to configuration data (among other uses):
+
+```meson
+fs = import('fs')
+cdata = configuration_data()
+copyright = fs.read('LICENSE')
+cdata.set('COPYRIGHT', copyright)
+if build_machine.system() == 'linux'
+    os_release = fs.read('/etc/os-release', allow_absolute: true)
```
jpakkane, line 153: This still has `allow_absolute`.

Issue #8276, opened 2021-01-31
Target private directory is unconditionally added to compile include command line arguments


hazelnusse:
**Describe the bug**
Meson unconditionally adds target private include directories to compile command. Most of the time, this is either harmless (in case when there is nothing in the private directory), or desirable (in the case when there are generated files that the target needs to include and they reside in the private directory). However, when a user explicitly requests that implicit include directories be not be added to the compilation command, it is surprising to see a private include directory that isn't needed show up on the command line.

**To Reproduce**
Build any meson project, or for a simple example, see:
https://asciinema.org/a/VDtcePfch4psg1jqBsBQSUz33

`meson.build`:
```
project('foo', ['c'])
executable('foo', ['foo.c'], implicit_include_directories : false)
```
`foo.c`:
```
int main() { return 0; }
```

On Linux with the Ninja backend I see:
```
$ ninja -v
[1/2] cc -Ifoo.p -fdiagnostics-color=always -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -g -MD -MQ foo.p/foo.c.o -MF foo.p/foo.c.o.d -o foo.p/foo.c.o -c ../foo.c
[2/2] cc  -o foo foo.p/foo.c.o -Wl,--as-needed -Wl,--no-undefined
```
Of note here is that despite having `implicit_include_directories : false`, there is a `-Ifoo.p` compile command line argument. 

**Expected behavior**
If I set `implicit_include_directories : false`, I would expect compilation commands of my targets to only include the include paths of the explicitly declared dependencies.

Additionally, the `implicit_include_directories` is only respected by the ninja and vs2010 backends -- it is ignored by the vs2015, vs2017, vs2019, and xcode backends. This might not make sense to fix in this ticket but I thought I would mention it in case it turns out to makes sense to refactor all this logic into a common function which is depended upon by all the backends.

bonzini:
Given #8277 was closed, this should follow suit.

PR #8277, opened 2021-01-31
Make private include directories implicit

hazelnusse:
This addresses #8276.

Each target ends up with a private directory where object files and
other generated or otherwise intermediate files are placed. This commit
makes the private directory part of the set of implicit include
directories.

The default behavior of meson is unchanged. For targets declaring
implicit_include_directories: false, the private directory is no longer
present in the compile include directories path, which may cause
problems in cases where there are implicit dependencies upon headers
that end up in the private directory.

### DIFF
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -671,11 +671,11 @@ class Backend:
     def get_pch_include_args(self, compiler, target):
         args = []
         pchpath = self.get_target_private_dir(target)
-        includeargs = compiler.get_include_args(pchpath, False)
         p = target.get_pch(compiler.get_language())
         if p:
             args += compiler.get_pch_use_args(pchpath, p[0])
-        return includeargs + args
+            args += compiler.get_include_args(pchpath, False)
+        return args
 
     def create_msvc_pch_implementation(self, target, lang, pch_header):
         # We have to include the language in the file name, otherwise
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -2454,11 +2454,12 @@ https://gcc.gnu.org/bugzilla/show_bug.cgi?id=47485'''))
         # both Meson and Autotools in parallel as part of the transition.
         if target.implicit_include_directories:
             commands += self.get_source_dir_include_args(target, compiler)
-        if target.implicit_include_directories:
             commands += self.get_build_dir_include_args(target, compiler)
-        # Finally add the private dir for the target to the include path. This
-        # must override everything else and must be the final path added.
-        commands += compiler.get_include_args(self.get_target_private_dir(target), False)
+            # Finally add the private dir for the target to the include path.
+            # This must override everything else and must be the final path
+            # added.
+            commands += compiler.get_include_args(self.get_target_private_dir(target), False)
+
         return commands
 
     def generate_single_compile(self, target, src, is_generated=False, header_deps=None, order_deps=None):
--- a/mesonbuild/backend/vs2010backend.py
+++ b/mesonbuild/backend/vs2010backend.py
@@ -983,12 +983,15 @@ class Vs2010Backend(backends.Backend):
                 file_args[l] += args
         # The highest priority includes. In order of directory search:
         # target private dir, target build dir, target source dir
-        for args in file_args.values():
-            t_inc_dirs = [self.relpath(self.get_target_private_dir(target),
-                                       self.get_target_dir(target))]
-            if target.implicit_include_directories:
-                t_inc_dirs += ['.', proj_to_src_dir]
-            args += ['-I' + arg for arg in t_inc_dirs]
+        if target.implicit_include_directories:
+            for args in file_args.values():
+                t_inc_dirs = [
+                    self.relpath(self.get_target_private_dir(target),
+                                 self.get_target_dir(target)),
+                    '.',
+                    proj_to_src_dir,
+                ]
+                args += ['-I' + arg for arg in t_inc_dirs]
 
         # Split preprocessor defines and include directories out of the list of
         # all extra arguments. The rest go into %(AdditionalOptions).
--- a/test cases/common/235 pch implicit include directories/main.cpp
+++ b/test cases/common/235 pch implicit include directories/main.cpp
@@ -1,0 +1,5 @@
+#include <vector>
+#include <string>
+#include <map>
+
+int main() {}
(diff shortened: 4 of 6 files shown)


jpakkane:
One of the design goals of Meson has always been to not add extra code and toggles to special cases unless there is an actual need. This is one of them. Having the private dir always in search path is one of those things. Doing it is easy, simple and does the right thing almost always. There has not been a single case where this would have caused an issue (that I can remember at least). Yes it does add one superfluous include argument but it has not been shown to cause any problems in real world usage. It does have a performance hit but it is most likely unmeasurably small.

And contrary to the description, this does break existing projects in at least two ways:

- Any target that disables implicit directories but does use generated headers breaks
- Similarly any target that disables implicit directories but uses precompiled headers breaks

hazelnusse:
@jpakkane I wholeheartedly agree with not adding another option. 

I did some reading on the PCH case and it appears that it should be relatively easy to address by inspecting with something like:
```
        if target.implicit_include_directories or target.get_pch(compiler.language):
            commands += compiler.get_include_args(self.get_target_private_dir(target), False)
```

With regards to the generator, I'm not totally sure how or if that case is similarly detectable, any suggestions?

Assuming there are simple and maintainable approaches to address the two issues you mentioned, what do you think?

Separately, what do you think about the fact that only vs2010 and ninja backends support the implicit_include_directories flag? Would it make sense to refactor this logic into a common place so that all backends could have this functionality without duplicating the logic, or is there a reason only these two backends have seen this option added?

hazelnusse:
@jpakkane with regards to generated headers, if one follows the guidance here:
https://mesonbuild.com/Wrap-best-practices-and-tips.html#declare-generated-headers-explicitly

would the private directory get added to the list of includes through the dependency declaration, and hence not need the unconditional addition of the private directory to the include paths?

I guess perhaps this might be a matter of how much existing code might break due to not following the recommendations in the documentation?

hazelnusse:
In my experimentation I observed the private target directory is added to the includes list in more than one place:
1) https://github.com/mesonbuild/meson/blob/master/mesonbuild/backend/backends.py#L674
2) https://github.com/mesonbuild/meson/blob/master/mesonbuild/backend/ninjabackend.py#L2461

My latest commit refactors this slightly, and adds a test which indicates that at least for the PCH case, the private include directory is added when needed and left out when not:
```
[1/5] c++ -fdiagnostics-color=always -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wnon-virtual-dtor -g -MD -MQ e_no_pch.p/main.cpp.o -MF e_no_pch.p/main.cpp.o.d -o e_no_pch.p/main.cpp.o -c '../test cases/common/235 pch implicit include directories/main.cpp'
[2/5] c++  -o e_no_pch e_no_pch.p/main.cpp.o -Wl,--as-needed -Wl,--no-undefined
[3/5] c++ -fdiagnostics-color=always -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wnon-virtual-dtor -g -MD -MQ e_pch.p/myexe_pch.hpp.gch -MF e_pch.p/myexe_pch.hpp.gch.d -o e_pch.p/myexe_pch.hpp.gch -c '../test cases/common/235 pch implicit include directories/pch/myexe_pch.hpp'
[4/5] c++ -Ie_pch.p -fdiagnostics-color=always -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wnon-virtual-dtor -g -fpch-preprocess -include myexe_pch.hpp -MD -MQ e_pch.p/main.cpp.o -MF e_pch.p/main.cpp.o.d -o e_pch.p/main.cpp.o -c '../test cases/common/235 pch implicit include directories/main.cpp'
[5/5] c++  -o e_pch e_pch.p/main.cpp.o -Wl,--as-needed -Wl,--no-undefined
[0/1] Running all tests.
1/2 e_pch           OK              0.01s
2/2 e_no_pch        OK              0.00s
```
Step 1 is the compile step in the case without PCH and indicates that the private directory *is not* present, as desired.
Step 4 is the compile step in the case with PCH and indicates that the private directory *is* present, as desired.

However, there are build failures so clearly something is wrong with my logic. Any thoughts @jpakkane ?

jpakkane:
I'm sorry, but I don't really understand the underlying issue here. Why is it so important for you to get rid of this one include flag, which does not seem to be causing problems in practice?

hazelnusse:
@jpakkane The underlying issue is that unnecessary include paths are added to the compilation step. Admittedly this is very minor and in the vast majority of cases has no effect on the generated object file. I would say that it isn't so important for most people, I just happen to be kind of obsessive about things like this and when possible try to avoid them. An analogy might be including a standard library header that isn't strictly necessary -- usually there is no observable effect other than perhaps slightly increased compilation times.

I get that having the private include directory is unlikely to cause any real problems. However, I'm curious what would the justification for unconditionally adding this to the include paths, as opposed to only adding it when strictly necessary, which as far as I can tell is when there are generated sources or pre-compiled headers. In my head this is a very minor improvement/optimization, but perhaps you feel differently and would rather not have this include path be conditionally added only when strictly necessary. If that is indeed the case, I can stop pursuing this PR, but if you are open accepting a patch that is able to do this without breaking tests and seems simple/maintainable, I am willing to put in a bit more effort (and your guidance would be most welcome).

hazelnusse:
@jpakkane It is looking like all the tests are passing (or will be momentarily), including the new one I added. I'm not super confident that I've added sufficient testing for this however, any thoughts on how to expand/change the test I added to ensure there isn't a corner case I'm missing?

jpakkane:
> However, I'm curious what would justify unconditionally adding the private directory to the include paths, as opposed to only adding it when strictly necessary

In a way you have already answered your own question earlier in this thread. Making the change turned out to be more complicated than expected, caused build failures and required more special case code. More code means more work, more maintenance, more everything. The current approach is simple, straightforward, easy to understand and adds pretty much zero maintenance burden.

hazelnusse:
@jpakkane I understand. As I said, this is a minor optimization. In my opinion it would be better to eliminate the `include_implicity_directories` option altogether and have meson explicitly track dependencies such that if you need the include path you get it but if you don't need it, you don't. Perhaps there would be more interest in exploring that path? Certainly this would be more work but in the end might be less maintenance.

FWIW, neither CMake or Bazel have unconditional include directories on a per target basis, as far as I can tell.

jpakkane:
I am a big fan of simple solutions so let's stick with the current implementation. Thanks.

PR #8281, opened 2021-01-31
environment: fix typos

alialnu:
Fixed typos and reworded some sentences.

Signed-off-by: Ali Alnubani <alialnu@nvidia.com>

### DIFF
--- a/mesonbuild/environment.py
+++ b/mesonbuild/environment.py
@@ -1006,7 +1006,7 @@ class Environment:
     def _handle_exceptions(self, exceptions, binaries, bintype='compiler'):
         errmsg = 'Unknown {}(s): {}'.format(bintype, binaries)
         if exceptions:
-            errmsg += '\nThe follow exceptions were encountered:'
+            errmsg += '\nThe following exception(s) were encountered:'
             for (c, e) in exceptions.items():
                 errmsg += '\nRunning "{0}" gave "{1}"'.format(c, e)
         raise EnvironmentException(errmsg)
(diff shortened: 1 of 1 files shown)


mesonbuild/environment.py
```diff
@@ -1778,8 +1778,8 @@ def detect_rust_compiler(self, for_machine: MachineChoice) -> RustCompiler:
                     mlog.warning(
                         'Please do not put -C linker= in your compiler '
                         'command, set rust_ld=command in your cross file '
-                        'or use the RUST_LD environment variable. meson '
-                        'will override your seletion otherwise.')
+                        'or use the RUST_LD environment variable, otherwise meson '
+                        'will override your selection.')
```
eli-schwartz, line 1782: Typo fix noted.

But the sentence is longer and more unwieldy than it used to be.

alialnu: Thanks for taking a look @eli-schwartz,

Is the following better?
```
                        'or use the RUST_LD environment variable, otherwise'
                        'meson will override your selection.')
```

dcbaker: As the person who wrote the first version with typo, I think that your version reads better.

eli-schwartz: I agree the rewording looks better with the proposed change. I'm wondering whether the punctuation could still use some tweaking...

jpakkane: This MR already makes things better, so merging. Any further tweaks can be done in subsequent MRs.

Issue #8283, opened 2021-02-01
Sanitizer support doesn't properly check for compiler/linker capabilities


ericonr:
**Describe the bug**

Meson doesn't check if the `b_sanitize` values passed to it are sane for the compiler it's currently using, which ends up with failures in the build step instead of during the configuration step. This is specially visible on musl libc systems, where, at least for now, there is no sanitizer support.

**To Reproduce**

Simply run `CC=gcc ./run_unittests.py` on a musl system, it will error out with the following message:

```
[1/19] Precompiling header '../test cases/common/13 pch/c/pch/prog.h'
[2/19] Generating makeheader with a custom command
[3/19] Compiling C object c/prog.p/prog.c.o
[4/19] Linking target c/prog
FAILED: c/prog 
gcc  -o c/prog c/prog.p/prog.c.o -fsanitize=address -Wl,--as-needed -Wl,--allow-shlib-undefined
/usr/bin/ld: cannot find libasan_preinit.o: No such file or directory
/usr/bin/ld: cannot find -lasan
collect2: error: ld returned 1 exit status
```

From my understanding of the decorator around the tests that use sanitizers, they should be skipped when unsupported, but this doesn't happen.

More examples:

- With GCC 10, which here seems to not know about the `memory` sanitizer, it will also error out (during the build) if I configure any C project with `-Db_sanitize=memory`:

```
samu: job failed: cc -Ilibpurrlib.a.p -I. -I../../home/ericonr/mypro/purr-c -fdiagnostics-color=always -fsanitize=memory -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wextra -Wpedantic -g '-DGETTEXT_PACKAGE="purr-c"' '-DGETTEXT_DIR="/usr/local/share/locale"' -DUSE_LIBINTL -DHAVE_PROG_INVOCATION -DHAVE_SOCK_CLOEXEC_H -DHAVE_PIPE2 -fPIC -MD -MQ libpurrlib.a.p/encrypt.c.o -MF libpurrlib.a.p/encrypt.c.o.d -o libpurrlib.a.p/encrypt.c.o -c ../../home/ericonr/mypro/purr-c/encrypt.c
cc: error: unrecognized argument to '-fsanitize=' option: 'memory'
```

- With GCC 10, which does know about the `address` sanitizer, it will instead error out during linking:

```
[16/18] Linking target tests
[...]

PR #8284, opened 2021-02-01
tests/common: fix " 37 has function" test on musl systems.

ericonr:
This commit fixes the test that asserts on whether the lchmod() function
should have been detected as available by Meson. It does so by assuming
that on Linux systems not using glibc, the function will be available.

- fix comment about lchmod on Linux: musl has implemented the function
correctly since 2013, so the assumption in the test wasn't correct.
Furthermore, musl doesn't use glibc's stub mechanism.
- fix include to receive __GLIBC__ definition: including almost any
header in glibc will end up defining __GLIBC__, since most headers
include <features.h>. The <gnu/libc-version.h> header was probably
chosen because of its name, but its actual purpose is defining functions
for checking glibc version at runtime (instead of what the binary was
built with), so it isn't necessary to use it. Since it is a completely
non standard header, including it makes the test suite fail on musl due
to not finding the header.

This + fixing #8283 should allow `run_unittests.py` to run without any changes on musl.

### DIFF
--- a/test cases/common/37 has function/meson.build
+++ b/test cases/common/37 has function/meson.build
@@ -44,11 +44,13 @@ foreach cc : compilers
     error('Found non-existent function "hfkerhisadf".')
   endif
 
-  # With glibc on Linux lchmod is a stub that will always return an error,
-  # we want to detect that and declare that the function is not available.
-  # We can't check for the C library used here of course, but if it's not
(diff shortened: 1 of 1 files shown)


eli-schwartz:
This fixes running this test case with `CC=musl-gcc` for me. Seems reasonable.

PR #8288, opened 2021-02-02
introduce add_test_setup(exclude suites: ...) keyword argument

bonzini:
This new keyword argument makes it possible to run specific test setups only on a subset of the tests.  For example, to mark some tests as slow and avoid running them by default:
    
    add_test_setup('quick', exclude_suites: ['slow'], is_default: true)

It will then be possible to run the slow tests with `meson test --suite slow` (or one could add `add_test_setup('slow')` for symmetry, if so inclined).


bonzini:
I also had a patch adding `only_suites`, but I couldn't think of a use.

Issue #2025, opened 2017-07-03
install_man() does not take configuration_data


whot:
A common pattern is to have man pages in some a format that requires replacements on install (e.g. insert the project version). Currently `install_man()` does not take a `configuration_data` object or other configuration, so any man page that needs basic replacements needs to go through `configure_file()`.

Is this something intentional or just a missing feature?

nirbheek:
A bit of both? `install_man()` is just like `install_data()` in that it only installs pre-existing source files. It is true that it is very common to have substitutions in a man page before installation. We could add a `configuration_data:` keyword argument as a short-hand for passing it through `configure_file()`.

bonzini:
Not an exact match, but closing as the ever returning "`install_man` but for `custom_target`s" issue.

Issue #1550, opened 2017-04-02
install_man does not accept target output


keszybz:
```
  t1 = custom_target(
      'bootup.1',
      input : 'bootup.xml',
      output : 'bootup.1',
      command : [xsltproc, '-o', '@OUTPUT@'] + xsltproc_flags + [custom_man_xsl, '@INPUT@'])
  install_man(t1)
```

```
Meson encountered an error in file man/meson.build, line 210, column 2:
Arguments must be strings.
```

I would have expected this to work...

jpakkane:
In Meson we try to keep as much information about targets in one place. Whether and how to install should be defined in the target declaration command as is done for all other target types. So something like this:

    custom_target(..., install : true, install_dir : join_paths(get_option('mandir'), 'man1')

Otherwise you could get into strange cases such as having a target that has `install : false` but which is used in an explicit install command. The outcome of this is ambiguous.

keszybz:
I don't see this as ambiguous: the outcome of a target can be used as input to another target or targets, and each can have a separate install, install_dir setting. I can do this with any generated file, and man pages should be no different.

Open-coding install_dir requires a reimplementation of the extracting of man page directory. It's possible, but annoying, especially when the man page name and section comes from some list and must be extracted by splitting strings and such.

keszybz:
... in other words, install : true/false pertains to the rule, not to the (potentially one of many) output file.

nirbheek:
For multiple outputs of a custom target, we have a PR open to implement that already: https://github.com/mesonbuild/meson/pull/1469

realh:
AIUI this means you can't use `install_mans` if your manpages are generated eg from XML. Generated manpages are very common, and `install_mans` does useful stuff that the `install` parameter of `custom_target` doesn't. So it would help a lot if `custom_target` had some sort of `install_function` parameter.

dcbaker:
I think we should implement this. We allow passing custom_targets to install_data, so we should do it for install_man as well.

eli-schwartz:
This is even worse now, because if you try it you get a gigantic traceback:

```
Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/mesonbuild/mesonmain.py", line 131, in run
    return options.run_func(options)
  File "/usr/lib/python3.8/site-packages/mesonbuild/msetup.py", line 245, in run
    app.generate()
  File "/usr/lib/python3.8/site-packages/mesonbuild/msetup.py", line 159, in generate
[...]

PR #8291, opened 2021-02-02
Make installing generated man files easier

dcbaker:
Something that's long been a wart in meson is using man pages. If you man pages are static files you can pass them to `install_man`, which handles figuring out where to install the man pages to. For generated man pages (using either `configure_file` or `custom_target`) you must manually build the install directory. This is neither dry nor pleasant. We've discussed allowing passing these targets to `install_man`, but that creates an odd system where you either set `install : true; install_dir : ...` OR pass it to `install_man`, and violates the design goal of meson that all information about a target should be provided at target creation time.

This then, keeps all of those constraints by introducing an `'@MANDIR@'` special value that can be passed to `custom_target` and `configure_file`'s `install_dir` keyword arguments. Meson will then use the same logic it uses in `install_man` to calculate the correct install location for the target automatically.

For example:
```meson
custom_target(
  output : 'page.1',
  install : true,
  install_dir : '@MANDIR@',
)
```

Along the way I had to do a bit of refactoring, so there're some cleanups and type annotations and such in here as well. The only one of these that is significant is `Better documentation and argument checking for multiple install_dirs`, which adds and moves so argument validation logic around.

Fixes: #1550
Fixes: #2025

### DIFF
--- a/docs/markdown/Reference-manual.md
+++ b/docs/markdown/Reference-manual.md
@@ -302,6 +302,8 @@ false otherwise.
 - `install_dir`: the subdirectory to install the generated file to
   (e.g. `share/myproject`), if omitted or given the value of empty
   string, the file is not installed.
+  *(since 0.58.0)* the special `'@MANDIR@'` value can be used for generated
+  man pages. [See here for examples](howtox.md#Install-my-generated-man-pages).
 - `install_mode` *(since 0.47.0)*: specify the file mode in symbolic format
   and optionally the owner/uid and group/gid for the installed files.
 - `output`: the output file name. *(since 0.41.0)* may contain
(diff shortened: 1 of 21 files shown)


jpakkane:
Could (and indeed should) this be made general. That is so it would work with other things as well? So instead of this:

`´`
install_dir: get_option('datadir') / 'myproj'
```

you could have something like:

```
install_dir: '@datadir@/myproj'
```

dcbaker:
There is a difference here between `get_option('mandir')` and `@MANDIR@`, because the latter expands to `get_option('mandir') / man${section}'`

bonzini:
What happens if someone thinks its funny to specify `@MANDIR@` as their mandir, and an existing meson.build uses `get_option('mandir') / 'man1'`?

eli-schwartz:
What happens if someone wants to actually create a literal custom target output file named `@MESON_VARIABLE@` for a given value of documented meson variable replacements?

dcbaker:
I think given #8398 it probably makes more sense to not use a special value, and to use a module for this instead.

PR #8292, opened 2021-02-02
Make installing non-existing subdirs a supported feature

### LINKED ISSUE #2904: Provide a way to install an empty directory
Sometimes you need to create an empty directory (for third-party plugins, fonts, themes,…) as part of the software install. 
It does not seem currently possible. Maybe allow passing an empty subdir name to install_subdir to do the trick ?

whot:
`install_subdir()` with a non-existing subdir creates the directory in the
target directory. This seems like an implementation detail but is quite useful
to create new directories for e.g. configuration or plugins in the installed
locations.

Let's add a test for it and document it to make this behavior official.

Limitation: it can only create at the install_dir location, trying to create
nested subdirectories does not work and indeed creates the wrong directory
structure. That is a bug that should be fixed separately:

        install_subdir('blah',
                        install_dir: get_option('prefix'))
        install_subdir('sub/foobar',
                        install_dir: get_option('prefix'))
        install_subdir('foo/baz',
                        install_dir: get_option('prefix'))

        $ tree ../_inst
        ../_inst
        ├── baz
        ├── blah
        └── foobar

Fixes #2904

### DIFF
--- a/docs/markdown/Reference-manual.md
+++ b/docs/markdown/Reference-manual.md
@@ -1140,6 +1140,10 @@ Installs the entire given subdirectory and its contents from the
 source tree to the location specified by the keyword argument
 `install_dir`.
 
+If the subdirectory does not exist in the source tree, an empty directory is
+created in the specified location. *(since 0.45.0)* A newly created
+subdirectory may only be created in the keyword argument `install_dir`.
+
 The following keyword arguments are supported:
 
 - `exclude_files`: a list of file names that should not be installed.
@@ -1187,6 +1191,12 @@ share/
 ```text
 share/
   file1
+```
+
+`install_subdir('new_directory', install_dir : 'share')` creates
+```text
+share/
+  new_directory/
 ```
 
 ### is_disabler()
--- a/test cases/common/60 install subdir/test.json
+++ b/test cases/common/60 install subdir/test.json
@@ -11,6 +11,7 @@
     {"type": "file", "file": "usr/share/sub1/third.dat"},
     {"type": "file", "file": "usr/share/sub1/sub2/data2.dat"},
     {"type": "file", "file": "usr/share/sub2/one.dat"},
-    {"type": "file", "file": "usr/share/sub2/dircheck/excluded-three.dat"}
+    {"type": "file", "file": "usr/share/sub2/dircheck/excluded-three.dat"},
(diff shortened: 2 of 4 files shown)


test cases/common/60 install subdir/test.json
```diff
@@ -11,6 +11,7 @@
     {"type": "file", "file": "usr/share/sub1/third.dat"},
     {"type": "file", "file": "usr/share/sub1/sub2/data2.dat"},
     {"type": "file", "file": "usr/share/sub2/one.dat"},
-    {"type": "file", "file": "usr/share/sub2/dircheck/excluded-three.dat"}
+    {"type": "file", "file": "usr/share/sub2/dircheck/excluded-three.dat"},
+    {"type": "file", "file": "usr/share/new_directory"}
```
dcbaker, line 15: shouldn't this be type "dir" (or "directory", can't remember which)?

whot: fixed, thanks.

dcbaker:
the msys2 CI is apparently just broken ATM. It's failing in all of my MRs as well

lazka:
> the msys2 CI is apparently just broken ATM. It's failing in all of my MRs as well

sourceforge was broken yesterday, serving garbage files.

dcbaker:
yay sourceforge

docs/markdown/Reference-manual.md
```diff
@@ -1140,6 +1140,10 @@ Installs the entire given subdirectory and its contents from the
 source tree to the location specified by the keyword argument
 `install_dir`.
 
+If the subdirectory does not exist in the source tree, an empty directory is
+created in the specified location. A newly created subdirectory may only be
+created in the keyword argument `install_dir`.
+
```
dcbaker, line 1146: We should probably add a note that this is guaranteed to work after 0.57.0, and may work on older versions, just in case it turns out that didn't work in some ancient version of meson…

whot: I ran a simple script checking out every meson version and it works all the way back to including 0.45.0. Bisection then showed 8fe816101467e66792251b4f57e0ddddb537764a as the first commit supporting it.

I've added a `*(since 0.45.0)*` tag to the documentation now. I don't think we need to worry about stable updates to 2y old meson versions breaking this behaviour :)

PR #8297, opened 2021-02-04
xcode-backend: fix include paths

jmoguillansky-gpsw:
Add include paths from dependencies for all languages (c, c++, objc, objc++)

This is a continuation of https://github.com/mesonbuild/meson/pull/8257

### DIFF
--- a/mesonbuild/backend/xcodebackend.py
+++ b/mesonbuild/backend/xcodebackend.py
@@ -735,6 +735,7 @@ class XCodeBackend(backends.Backend):
                     product_name = target.get_basename()
                 ldargs += target.link_args
                 cargs = []
+                cargs.append('-I.')
                 for dep in target.get_external_deps():
                     cargs += dep.get_compile_args()
                     ldargs += dep.get_link_args()
@@ -753,7 +754,7 @@ class XCodeBackend(backends.Backend):
                     args = pargs + gargs + targs
                     if args:
                         langargs[langnamemap[lang]] = args
-                langargs['C'] += cargs
+                        langargs[langnamemap[lang]] += cargs
                 symroot = os.path.join(self.environment.get_build_dir(), target.subdir)
                 self.write_line('%s /* %s */ = {' % (valid, buildtype))
                 self.indent_level += 1


mesonbuild/backend/xcodebackend.py
```diff
@@ -735,6 +735,7 @@ def generate_xc_build_configuration(self):
                     product_name = target.get_basename()
                 ldargs += target.link_args
                 cargs = []
+                cargs.append('-I.')
```
dcbaker, line 738: this doesn't seem right, isn't this ignoring the implicit_include_paths argument?

jmoguillansky-gpsw:
Hi Dylan,
When looking at the build output in xcode, it first does cd to the project directory,
then it runs clang to compile the code.
This basically adds the project directory as an include path.
Otherwise I get a compile error.
I’m not sure if this is the best way though.
Is there an API for getting the project directory?

Thanks,

Jeff

mesonbuild/backend/xcodebackend.py
jmoguillansky-gpsw: Hi Dylan,
When looking at the build output in xcode, it first does cd to the source directory,
then it runs clang to compile the code.
This basically adds the source directory as an include path.
Otherwise I get a compile error.
I’m not sure if this is the best way though.
Is there an API for getting the source directory?

Thanks,
Jeff

dcbaker:
I'm not sure, but there's:
```
implicit_include_directories (since 0.42.0): a boolean telling whether Meson adds the current source and build directories to the include path, defaults to true.
```

I'm not sure if this is ignoring that option or not. It also seems odd that we wouldn't be adding the mirrored build dir at the same time.

The code in that does this in the ninja backend is:
```python
        if target.implicit_include_directories:
            commands += self.get_source_dir_include_args(target, compiler)
        if target.implicit_include_directories:
            commands += self.get_build_dir_include_args(target, compiler)
```

jmoguillansky-gpsw:
I tried using that API to get the source directory, but it returns a path relative to the build directory.
However, XCode does cd to the source directory before compiling the file, so I get a compile error (in other words it’s not adding the source directory as an include path).

I put the code here if you’d like to take a look:
https://github.com/jmoguillansky-gpsw/meson/pull/new/xcode-add-implicit-includes

jmoguillansky-gpsw:
Thanks @dcbaker ,
I submitted a new PR: https://github.com/mesonbuild/meson/pull/8303 for addressing the implicit includes.
However I'm not sure it's correct (please see my comment in the PR)...

PR #8299, opened 2021-02-04
Add a positional argumetn checking deocrator for the interpreter

dcbaker:
edit: We're looking to land this incrementally, starting with: #8304 

This should not be merged until after 0.57.0 is released.

Currently, we have a rather haphazard level of typing checking for arguments in the interpreter. Some functions and methods open code type checking and argument number checking, but not all of them do. Even when functions do said checking it's a ton of repetitive boilerplate that everyone has to implement. I propose instead using a function decorator for this. It works like this:
```python
@typed_pos_args('func_name', str, int, (CustomTargetHolder, CustomTargetIndexHolder)
def func(..., args : T.Tuple[str, int, T.Union[CustomTargetHolder, CustomTargetIndexHolder]], kwargs):
```
The function will detect two things, first that the number of arguments is correct (3), then it will detect that they have the correct type, and finally create a tuple of the arguments (rather than a list, because in tuples the index is relevant for the types). We can annotate the function appropriately, and both mypy is happy, and we get cheap, easy, painless type checking *before* the function actually starts looking at the positional arguments.

But what about optional arguments? We can handle that too!
```python
@typed_pos_ags('name', str, optargs=[str, (int, float)])
def func(..., args: T.Tuple[str, T.Optional[str], T.Optional[int, float], kwargs):
```
In this case we get the same level of type checking awesomeness, but optional values are optional, and we get a `None` for them. This has nice properties as we can do things like
```python
def func(...):
    name, pos, value = args
    if pos is not None:
        ...
```
And not have to resort to length checks. The decorator will also enforce that the positional arguments are given, and that no more than the maximum number of optional arguments are given.

But what about variadic functions? Got that covered too!
```python
@typed_pos_args('name', str, varargs=str)
def func(..., args: T.Tuple[str, T.List[str]], kwargs):
    ...
```
Notice that the decorator puts the variadic args in a list at the end? We can use this to separate mandatory arguments from variadiac ones. We can also set a minimum and maximum number of variadic arguments, and have functions with only variadic arguments.

Also, this code comes with unit tests, unlike the open coded versions.

Now, let's be clear, I've done my best to make sure that all the correct types are allowed. But I've certainly made mistakes, there are going to be some cases where what I've written breaks something. Therefore, I'd like to merge this as early in the 0.58 cycle as possible to get the most testing possible. And this is tightening the allowed types on many functions in meson that previously would take whatever was thrown at them.

I've tried to get as much of the code as possible converted over to this, but there is one case that this cannot handle, and that's the case of functions that are overloaded. There were only a couple of those that I could find, so I didn't feel like trying to add a bunch of infrastructure that would be more code than it removed. I've also not gone into the qt modules, both because that's where the overloaded functions are, and also because understanding that module is hard, and the documentation is lacking.

### DIFF
--- a/mesonbuild/interpreter.py
+++ b/mesonbuild/interpreter.py
@@ -3119,15 +3119,15 @@ external dependencies (including libraries) must go to "dependencies".''')
 
     @FeatureNewKwargs('add_languages', '0.54.0', ['native'])
     @permittedKwargs(permitted_kwargs['add_languages'])
-    @stringArgs
-    def func_add_languages(self, node, args, kwargs):
+    @typed_pos_args('add_languages', varargs=str)
+    def func_add_languages(self, node, args: T.Tuple[T.List[str]], kwargs):
         disabled, required, feature = extract_required_kwarg(kwargs, self.subproject)
         if disabled:
-            for lang in sorted(args, key=compilers.sort_clink):
+            for lang in sorted(args[0], key=compilers.sort_clink):
                 mlog.log('Compiler for language', mlog.bold(lang), 'skipped: feature', mlog.bold(feature), 'disabled')
             return False
         if 'native' in kwargs:
-            return self.add_languages(args, required, self.machine_from_native_kwarg(kwargs))
+            return self.add_languages(args[0], required, self.machine_from_native_kwarg(kwargs))
         else:
             # absent 'native' means 'both' for backwards compatibility
             tv = FeatureNew.get_target_version(self.subproject)
@@ -3135,8 +3135,8 @@ external dependencies (including libraries) must go to "dependencies".''')
                 mlog.warning('add_languages is missing native:, assuming languages are wanted for both host and build.',
                              location=self.current_node)
 
-            success = self.add_languages(args, False, MachineChoice.BUILD)
-            success &= self.add_languages(args, required, MachineChoice.HOST)
+            success = self.add_languages(args[0], False, MachineChoice.BUILD)
+            success &= self.add_languages(args[0], required, MachineChoice.HOST)
             return success
 
     @noArgsFlattening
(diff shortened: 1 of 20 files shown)


dcbaker:
Ah I've made mypy mad :)

mesonbuild/interpreter.py
```diff
@@ -2017,35 +1968,32 @@ def _process_script_args(
...
+    @typed_pos_args('meson.install_script', (str, mesonlib.File, ExecutableHolder, ExternalProgramHolder), varargs=(str, mesonlib.File, CustomTargetHolder, CustomTargetIndexHolder, ConfigureFileHolder, ExecutableHolder, ExternalProgramHolder))
+    def add_install_script_method(self, args: T.Tuple[T.Union[str,mesonlib.File, ExternalProgramHolder, ExecutableHolder], T.List[T.Union[str, mesonlib.File, CustomTargetHolder, CustomTargetIndexHolder, ConfigureFileHolder, ExecutableHolder, ExternalProgramHolder]]], kwargs: T.Dict[str, T.Any]) -> None:
         if isinstance(args[0], mesonlib.File):
             FeatureNew.single_use('Passing file object to script parameter of add_install_script',
                                   '0.57.0', self.interpreter.subproject)
         skip_if_destdir = kwargs.get('skip_if_destdir', False)
         if not isinstance(skip_if_destdir, bool):
             raise InterpreterException('skip_if_destdir keyword argument must be boolean')
-        script_args = self._process_script_args('add_install_script', args[1:], allow_built=True)
+        script_args = self._process_script_args('add_install_script', args[1], allow_built=True)
```
xclaesse, line 1979: Just reading randomly the diff to see how it works. I don't think that change is correct, it process only the first arg instead of all of them, no?

mesonbuild/interpreter.py
```diff
@@ -2194,17 +2135,10 @@ def override_find_program_method(self, args, kwargs):
...
     @permittedKwargs({'native'})
-    def override_dependency_method(self, args, kwargs):
-        if len(args) != 2:
-            raise InterpreterException('Override needs two arguments')
+    @typed_pos_args('override', str, DependencyHolder)
+    def override_dependency_method(self, args: T.Tuple[str, DependencyHolder], kwargs):
         name = args[0]
-        dep = args[1]
-        if not isinstance(name, str) or not name:
-            raise InterpreterException('First argument must be a string and cannot be empty')
```
xclaesse, line 2203: You still have to check name is not empty, otherwise you could override the special `''` dependency.

mesonbuild/interpreter.py
xclaesse: Forget that, I understood now, you change args to be a tuple.

mesonbuild/interpreter.py
dcbaker: You're right, I've changed that in the most recent version. The message did change a bit though.

dcbaker:
I should also mention that I'd be happy to merge this piecemeal, I know for example that the usage in the rust module is correct, so if people were happier to do an incremental merge we could merge the core infrastructure and individual commits.

xclaesse:
@dcbaker I think incremental merge would be better indeed. That's the kind of PR that conflict quickly and reviewing takes time because small details are hard to spot. I would love to get at least the new API and the unit tests merged asap, maybe even before0.57 release, so we can already make using it to new API mandatory, then we can port each module and interpreter incrementally.

dcbaker:
This has mostly landed at this point one piece at a time.

PR #8303, opened 2021-02-04
xcode-backend: add implicit includes + global link args

jmoguillansky-gpsw:


jmoguillansky-gpsw:
I'm not sure this logic is correct...
When compiling the file, xcode first does cd to the source directory, and the runs clang to compile the file.
However, these include paths seem to be relative to the build directory.  
So the include paths should perhaps be relative to the source directory instead (for xcode backend)?

jmoguillansky-gpsw:
it seems like get_source_dir_include_args and get_build_dir_include_args return the same include paths, just one of them is relative to source directory, and the other is relative to build directory, so I think this logic is ok...

jmoguillansky-gpsw:
updated: seems that "get_build_dir_include_args" provides the correct behavior for xcode backend

dcbaker (approved):
LGTM. thanks!

mesonbuild/backend/xcodebackend.py
```diff
@@ -734,8 +734,9 @@ def generate_xc_build_configuration(self):
                 else:
                     product_name = target.get_basename()
                 ldargs += target.link_args
+                linker, stdlib_args = self.determine_linker_and_stdlib_args(target)
+                ldargs += self.build.get_global_link_args(linker, target.for_machine)
```
dcbaker, line 738: oh, actually, while you're here... do we hand `self.build.get_project_link_args` in the xcode backend? might as well handle that too...

jmoguillansky-gpsw: updated

PR #8304, opened 2021-02-04
Type checking decorator incremental

dcbaker:
This is just the implementation and one user in the rust module from #8299

### DIFF
--- a/mesonbuild/interpreterbase.py
+++ b/mesonbuild/interpreterbase.py
@@ -18,10 +18,11 @@
 from . import mparser, mesonlib, mlog
 from . import environment, dependencies
 
-import abc
-import os, copy, re
-import collections.abc
 from functools import wraps
+import abc
+import collections.abc
+import itertools
+import os, copy, re
 import typing as T
 
 TV_fw_var = T.Union[str, int, float, bool, list, dict, 'InterpreterObject', 'ObjectHolder']
@@ -229,7 +230,9 @@ class permittedKwargs:
         return T.cast(TV_func, wrapped)
 
 
-def typed_pos_args(name: str, *types: T.Union[T.Type, T.Tuple[T.Type, ...]]) -> T.Callable[..., T.Any]:
+def typed_pos_args(name: str, *types: T.Union[T.Type, T.Tuple[T.Type, ...]],
+                   varargs: T.Optional[T.Union[T.Type, T.Tuple[T.Type]]] = None,
+                   min_varargs: int = 0, max_varargs: int = 0) -> T.Callable[..., T.Any]:
     """Decorator that types type checking of positional arguments.
 
     allows replacing this:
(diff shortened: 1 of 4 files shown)


mesonbuild/interpreterbase.py
```diff
@@ -228,6 +229,123 @@ def wrapped(*wrapped_args: T.Any, **wrapped_kwargs: T.Any) -> T.Any:
...
+
+            num_args = len(args)
+            num_types = len(types)
+            a_types = types
+
+            if varargs:
+                num_types += min_varargs
+                if max_varargs == 0 and num_args < num_types:
+                    raise InvalidArguments(f'{name} takes at least {num_types} arguments, but got {num_args}.')
+                elif max_varargs > 0 and num_args > num_types + max_varargs:
```
xclaesse, line 299: nitpicking: max_varargs is always >0 in this elif case.

I'm not sure `num_args > num_types + max_varargs` is correct because here `num_types` is `len(types) + min_varargs`. If you have between 1 and 2 varargs, it will raise error only if `num_args > 3` instead of `>2`. Unless I'm mistaken. In any case, you have no unit test where you define both max_varargs and min_varargs and you give 1 more arg than the max.

mesonbuild/interpreterbase.py
```diff
@@ -228,6 +229,123 @@ def wrapped(*wrapped_args: T.Any, **wrapped_kwargs: T.Any) -> T.Any:
...
+            elif num_args != num_types:
+                raise InvalidArguments(f'{name} takes exactly {num_types} arguments, but got {num_args}.')
+
+            for i, (arg, type_) in enumerate(itertools.zip_longest(args, a_types, fillvalue=varargs), start=1):
+                if not isinstance(arg, type_):
+                    if isinstance(type_, tuple):
+                        shouldbe = f'one of {", ".join(t.__name__ for t in type_)}'
+                    else:
+                        shouldbe = type_.__name__
+                    raise InvalidArguments(f'{name} argument {i} was of type {type(arg).__name__} but should have been {shouldbe}')
```
xclaesse, line 319: nitpicking: I think it's easier to read when quoting the types, e.g. `foo argument 2 was of type "float" but should have been "int"`

mesonbuild/interpreterbase.py
dcbaker: hmm, yeah, you're right. now to decide how to fix it :)

dcbaker: Actually, max_varargs isn't always >0 in that case, it's still 0 if there are enough arguments, But you're right on the second note, I added tests and they immediately failed :)

mesonbuild/interpreterbase.py
```diff
@@ -228,6 +229,123 @@ def wrapped(*wrapped_args: T.Any, **wrapped_kwargs: T.Any) -> T.Any:
...
+    ```
+    """
+    def inner(f: TV_func) -> TV_func:
+
+        @wraps(f)
+        def wrapper(*wrapped_args: T.Any, **wrapped_kwargs: T.Any) -> T.Any:
+            args = _get_callee_args(wrapped_args)[2]
+            assert isinstance(args, list), args
+
+            # These are implementation programming errors, end users should never see them.
```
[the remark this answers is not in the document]
dcbaker: agreed, done

mesonbuild/interpreterbase.py
```diff
@@ -228,6 +229,123 @@ def wrapped(*wrapped_args: T.Any, **wrapped_kwargs: T.Any) -> T.Any:
...
+            assert max_varargs >= 0, 'max_varrags cannot be negative'
+            assert optargs is None or varargs is None, \
+                'varargs and opt args not supported together becuase this would be ambigious'
```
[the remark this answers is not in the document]
dcbaker: fixed

mesonbuild/interpreterbase.py
dcbaker: agreed, done.

mesonbuild/interpreterbase.py
```diff
@@ -228,6 +229,124 @@ def wrapped(*wrapped_args: T.Any, **wrapped_kwargs: T.Any) -> T.Any:
...
+            num_args = len(args)
+            num_types = len(types)
+            a_types = types
+
+            if varargs:
+                num_types += min_varargs
+                if max_varargs == 0 and num_args < num_types:
+                    raise InvalidArguments(f'{name} takes at least {num_types} arguments, but got {num_args}.')
+                elif max_varargs != 0 and (num_args < num_types or num_args > num_types + max_varargs - min_varargs):
+                    raise InvalidArguments(f'{name} takes between {num_types} and {num_types + max_varargs - min_varargs} arguments, but got {len(args)}.')
```
xclaesse, line 301: I find it easier to read by adding more descriptive variables:
```suggestion
                min_args = num_types + min_varargs
                max_args = num_types + max_varargs
                if max_varargs == 0 and num_args < min_args:
                    raise InvalidArguments(f'{name} takes at least {min_args} arguments, but got {num_args}.')
                elif max_varargs != 0 and (num_args < min_args or num_args > max_args)):
                    raise InvalidArguments(f'{name} takes between {min_args} and {max_args} arguments, but got {num_args}.')
```

What do you think?

mesonbuild/interpreterbase.py
xclaesse: > Actually, max_varargs isn't always >0 in that case, it's still 0 if there are enough arguments

Right.

xclaesse:
Other than this suggestion, it's ready to go. Nice work, that's really going to improve our code! Looking forward in doing similar type checking for kwargs now :D

mesonbuild/interpreterbase.py
dcbaker: Seems good to me. I've updated that and updated the docstring to try to be more useful.

xclaesse:
LGTM once CI is green.

PR #8305, opened 2021-02-04
run_target: Add env kwarg

xclaesse:
Re-implement it in backend using the same code path as for
custom_target(). This for example handle setting PATH on Windows when
command is an executable.


jpakkane:
According to Git bisect, this change broke the cross file test: https://github.com/mesonbuild/meson/pull/8313/checks?check_run_id=1849734517

jpakkane:
Specifically, the code no longer checks that the exe wrapper is valid when it encounters a run target (and presumably a custom target).

xclaesse:
@jpakkane Sorry, I don't understand how this has not been catched by the CI before merging. Made a fix there: https://github.com/mesonbuild/meson/pull/8314

tristan957:
@xclaesse thanks for your work on this issue. I am excited to use it :).

PR #8309, opened 2021-02-05
lint run_unittests.py

eli-schwartz:
One minor error when already raising an assertion, and some inconsequential code cleanup.


dcbaker:
Might as well run pylint on run_unittests in ci as well?

eli-schwartz:
"lint" != "pylint"

I used flake8 to catch these. And this is only 8 out of 18 flake8 warnings, but the rest are basically arguing over when to use whitespace in arithmetic and these ones seemed to be no-brainers to fix.

dcbaker (approved):

Issue #5026, opened 2019-03-06
Add a range() function


pwithnall:
Like Python’s `range()`/`xrange()` functions, but usable from `meson.build`. This would be useful in combination with `foreach` for generating lists of things and operating on them.

My particular use case is to generate a list of all old stable versions (something like `range($first_stable_release, $next_stable_release, 2)`) so they can be passed as arguments to a build process.

michaelcadilhac:
Another use case (if that's appropriate): I have 164 test files, I'm running a script on each of them.  The file names are all something like `testN.txt`, with `N` in `1..164`, and I'd rather build the set of test files using a loop.

Issue #8310, opened 2021-02-06
meson --internal exe can add a lot of overhead


icecream95:
For custom targets where the output is captured (and in some other cases), Meson wraps the command with `meson --internal exe`. This call to meson adds overhead for each of these custom targets that are executed.

On my system, running `meson --internal exe -- echo` takes about 0.8 seconds. It seems that most of this is from importing all of the Python modules used by Meson.

This can add to about a minute of CPU time in total for larger projects, which is significant when doing a clean build with ccache.

Some solutions I can think of are:

- Add a lighter-weight script to replace `meson --internal exe`
- Add handling for `meson --internal exe` to the top of the meson executable, before everything is imported
- Using features of backends to redirect output where possible
- Discourage output capturing from being used
- Rewrite Meson in Rust :)

xclaesse:
I thought about this recently but did not made any measurement. Thanks for confirming my fear. I think option 2 is what we should do. If `sys.argv[1] == '--internal'` we should fast track to running the script without importing any other meson module.

dcbaker:
We should just avoid calling meson as the launcher for internal at all, there's no reason for it, and it avoids the need for hacks like looking in argv

dcbaker:
But also, avoiding capturing is best practice, and we should discourage it

xclaesse:
> We should just avoid calling meson as the launcher for internal at all, there's no reason for it, and it avoids the need for hacks like looking in argv

We already avoid it when possible, but it's required in a few cases, like setting env which is needed on Windows to set PATH.

xclaesse:
btw, IIRC there are people that embed python interpreter inside a standalone meson.exe, which is why we have to invoke our python scripts via meson and not a separate python script. But maybe that requirement is outdated?

I've been checking the modules we load, and one part of the problem is my fault. I added build.EnvironmentVariables object in the ExecutableSerialisation object we pickle. That make it load the whole thing basically... Working on it.

dcbaker:
@xclaesse we can do some clever stuff with serialization that might help. But avoiding calling meson whenever possible would be best

icecream95:
Ninja on posix systems uses `sh -c` to execute commands, so adding redirects to command lines works as expected.

This quick hack works for me:

```diff
diff --git a/mesonbuild/backend/ninjabackend.py b/mesonbuild/backend/ninjabackend.py
index ca17f19de..6226e6a0d 100644
--- a/mesonbuild/backend/ninjabackend.py
[...]

PR #8311, opened 2021-02-06
Add range() function

xclaesse:
Fixes: #5026

### DIFF
--- a/docs/markdown/Reference-manual.md
+++ b/docs/markdown/Reference-manual.md
@@ -1861,6 +1861,36 @@ which to build the targets.
 If you desire more specific behavior than what this command provides,
 you should use `custom_target`.
 
+### range()
+
+``` meson
+    rangeobject range(stop)
+    rangeobject range(start, stop[, step])
+```
+
+*Since 0.58.0*
+
+Return an opaque object that can be only be used in `foreach` statements.
+- `start` must be integer greater or equal to 0. Defaults to 0.
+- `stop` must be integer greater or equal to `start`.
+- `step` must be integer greater or equal to 1. Defaults to 1.
+
+It cause the `foreach` loop to be called with the value from `start` included
+to `stop` excluded with an increment of `step` after each loop.
+
+```meson
+# Loop 15 times with i from 0 to 14 included.
+foreach i : range(15)
+   ...
+endforeach
+```
+
+The range object can also be assigned to a variable and indexed.
+```meson
+r = range(5, 10, 2)
+assert(r[2] == 9)
+```
+
 ## Built-in objects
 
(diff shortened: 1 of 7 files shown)


tp-m:
Out of curiosity, where do you use this?

(I have a vague memory of wanting it once somewhere, but don't remember where.)

xclaesse:
@tp-m it has been requested a few times, out of memory I remember this one: https://gitlab.gnome.org/GNOME/glib/-/blob/master/docs/reference/meson.build#L5

In this particular case, I wanted to make a benchmark that runs a thousand custom_target() for https://github.com/mesonbuild/meson/issues/8310.

xclaesse:
@jpakkane this should be passing the CI now, if it's still on time for the freeze.

xclaesse:
CI failure is unrelated

docs/markdown/Reference-manual.md
```diff
@@ -1825,6 +1825,36 @@ which to build the targets.
...
+- `start` must be integer greater or equal to 0. Defaults to 0.
+- `stop` must be integer greater or equal to `start`.
+- `step` must be integer greater or equal to 1. Defaults to 1.
+
+It cause the `foreach` loop to be called with the value from `start` included
+to `stop` excluded with an increment of `step` after each loop.
+
+```meson
+# Loop 15 times with i from 0 to 14 included.
+foreach i : range(15)
```
tp-m, line 1847: Is there a particular reason or precedent elsewhere that made you go for a `[start, stop[` range rather than an inclusive `[start, stop]`? I personally find the latter more intuitive.

xclaesse: That's what Python does. Can be seen both ways, range(5) does 5 loops, from 0 to 4...

xclaesse: Same thing than common c loop for(i=0; I < stop; i++)

xclaesse: https://en.m.wikipedia.org/wiki/Off-by-one_error

bonzini:
I have no objections, but have you thought of using the Rust syntax `a..b`? The main advantage is future extensibility, like for substring/subarray extraction (for which it would be nice to have open-ended ranges) and inclusive ranges (`a..=b`). There is of course the disadvantage of having to change the lexer for a (still) niche feature.

xclaesse:
We already have substring, it's just a method on string object. I think it's important to keep lexer simple.

xclaesse:
@jpakkane @dcbaker any objection?

eli-schwartz:
It answers a need for something that cannot otherwise be conveniently done, so sure, I don't see why not.

PR #8314, opened 2021-02-07
run_target: Fix exe_wrapper detection

xclaesse:

PR #8317, opened 2021-02-08
Fix exe wrapper detection for run targets.

jpakkane:
More exhaustive version of #8314.


xclaesse:
Thanks for fixing the mess I left.

xclaesse:
I'll try to refactor this a bit more after the release because they're still inconsistencies in the multiple places we process command line args

PR #8321, opened 2021-02-09
run_target: some more fixes

xclaesse:


dcbaker (approved):
Thanks for adding a test case. This all looks good to me.

xclaesse:
> Thanks for adding a test case. This all looks good to me.

Running some more test, I realized that with 0.56.2:
- custom_target() unit test crash during at setup time
- If I remove `capture: true` it does not crash but fails with `/bin/sh: 1: configure: not found` during at build time.
- run_target() works and can find the `configure` script.

With current master, custom_target() with and without capture, and run_target() pass during setup, but all fails during build. More consistent bug is better, I guess...

Issue #8212, opened 2021-01-16
Meson test times out unless -v option is passed


alex-tee:
**Describe the bug**
`meson test` times out when it shouldn't, but `meson test -v` executes correctly. It looks like the test is somehow suspended and kept in a suspended state until it times out, because my program randomly stops producing output during the test.

**To Reproduce**
This is the test I am running from the [Zrythm repository](https://git.sr.ht/~alextee/zrythm), commit 99697cc5dddfd6584d780839aec7d4c1f6850dba.
`meson test -C build actions_arranger_selections`
It times out after 240 seconds.

The same issue occurs when doing `ninja -C build test`

When running the same test with `-v`, it finishes in 132.27s
`meson test -C build -v actions_arranger_selections`

*note: When using meson 0.55.1 (guix package), I do not experience this bug, and the test finishes without issues.*

**Expected behavior**
The test is expected to finish without timing out.

**system parameters**
* Is this a [cross build](https://mesonbuild.com/Cross-compilation.html) or just a plain native build (for the same computer)? Native
* what operating system (e.g. MacOS Catalina, Windows 10, CentOS 8.0, Ubuntu 18.04, etc.): Tested on GNU GuixSD and Arch Linux
* what Python version are you using: Python 3.8.2
* what `meson --version` Commit ea2f34e2860680bab82aa7c2538971ab788209ce
* what `ninja --version` if it's a Ninja build: 1.10.0

dcbaker:
@bonzini, this looks like a possible regression from the switch to asyncio.

bonzini:
A likely issue is that 755412b5261c448025d40214e3f2c0197e2e27bc awaits stdout before stderr---but the program hangs on a write to stderr, and therefore won't close stdout yet.

If so, #8200 fixes it.

bonzini:
@alex-tee Could you test the #8225 pull request, please?

alex-tee:
@bonzini looks like it passed

```
1/1 zrythm:actions / actions_arranger_selections        OK             156.17s
```

bonzini:
Fixed in master, please close.

PR #8322, opened 2021-02-09
mtest fixes for 0.57

bonzini:
* another deadlock similar to #8212
* avoid printing progress report as `123/-1s` when timeout is disabled
* allow `--wrapper` in combination with `--setup`, if the setup does not define a wrapper.
* avoid that asyncio tasks linger on timeout
* make non-parallel tests emit their output on the fly, similar to ninja console jobs


jpakkane:
Is this ready to merge or is there still updates to be done?

bonzini:
It's ready, I tested it on QEMU and Pekka also used it for his WIP Weston pull request. (I marked it as draft temporarily because I found a second bug and didn't want to make a second pull request until I had a patch for it as well).

Sorry for not thinking of testing timeouts more thoroughly earlier, since that's the most intrusive of the patches... error conditions are always the worst part.

PR #8323, opened 2021-02-10
aix: avoid -bsvr4 flag

peterh:
The svr4 linker flag causes issues, especially when compiling c++.
Replace `-z` options with the equivalent non-svr4 flags. When using
`-blibpath`, we must be careful to include the default system library
path, or the resulting executables will not be able to find libc.

This patch was suggested by @andreaskem in #7581.

### DIFF
--- a/mesonbuild/linkers.py
+++ b/mesonbuild/linkers.py
@@ -1278,25 +1278,38 @@ class AIXDynamicLinker(PosixDynamicLinkerMixin, DynamicLinker):
     id = 'ld.aix'
 
     def get_always_args(self) -> T.List[str]:
-        return self._apply_prefix(['-bsvr4', '-bnoipath', '-bbigtoc']) + super().get_always_args()
+        return self._apply_prefix(['-bnoipath', '-bbigtoc']) + super().get_always_args()
 
     def no_undefined_args(self) -> T.List[str]:
-        return self._apply_prefix(['-z', 'defs'])
+        return self._apply_prefix(['-bernotok'])
 
     def get_allow_undefined_args(self) -> T.List[str]:
-        return self._apply_prefix(['-z', 'nodefs'])
+        return self._apply_prefix(['-berok'])
 
     def build_rpath_args(self, env: 'Environment', build_dir: str, from_dir: str,
                          rpath_paths: str, build_rpath: str,
                          install_rpath: str) -> T.Tuple[T.List[str], T.Set[bytes]]:
-        all_paths = mesonlib.OrderedSet(['/opt/freeware/lib']) # for libgcc_s.a
-        for p in rpath_paths:
-            all_paths.add(os.path.join(build_dir, p))
-        if build_rpath != '':
-            all_paths.add(build_rpath)
+        all_paths = mesonlib.OrderedSet() # type: mesonlib.OrderedSet[str]
+        # install_rpath first, followed by other paths, and the system path last
         if install_rpath != '':
             all_paths.add(install_rpath)
-        return (self._apply_prefix([x for p in all_paths for x in ('-R', p)]), set())
+        if build_rpath != '':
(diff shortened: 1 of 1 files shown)


andreaskem:
Thank you very much for preparing this MR, I unfortunately got distracted along the way.

I tested this change and successfully built C++ and C executables on AIX 7.2.5 using GCC 8.3 from the Linux toolbox. The original issue I mentioned no longer occurs and these executables appear to function normally.

I encountered two other issues. Firstly, dependencies found via pkconf do not have their library path added to the `libpath` but this is most likely unrelated to this MR or AIX support in general. I might just misunderstand how this mechanism is supposed to work. Secondly, GCC seems to add `/opt/freeware/lib/gcc/powerpc-ibm-aix7.2.0.0/8/../../..` on its search path automatically before the paths added by the user. If I link e.g., `libxml2` using `find_library`, the flag `-lxml2` is placed on the linker command line. GCC then links `/opt/freeware/lib/libxml2.a` but with `libpath` looking like `/usr/lib:/lib:/opt/freeware/lib:...`, the library `/usr/lib/libxml2.a` is used at runtime leading to errors like
```
        0509-136   Symbol _GLOBAL__AIXI_libxml2_so (number 979) is not exported from
                   dependent module /usr/lib/libxml2.a(libxml2.so.2).
```
This can be worked around by `LIBPATH=/opt/freeware/lib`. Should `/opt/freeware/lib` be placed first if GCC is used to link? Not sure how this is should to work because I cannot quite wrap my head around how the `/opt/freeware/lib` stuff is supposed to be integrated in general, on AIX.

peterh:
Hmm. In general, I want to use the /usr/lib libraries, not the /opt/freeware/lib libraries, even when building with gcc. I suspect we should be picking up the library path at configure time, not hard coding it to /lib:/usr/lib:/opt/freeware/lib (or any other ordering).

The docs say that if you omit libpath, it adds all the -L values. But then how do we specify the rpath?

I'll dig deeper and see what I can come up with. Thanks for looking!

dcbaker:
Is this ready to merge, or do you want to wait?

peterh:
@dcbaker I'd like to wait a day or two. As much as I want this fix in, I would like to address the issues raised by andreskam first. (I was going to look at this today, but then meetings happened).

dcbaker:
sounds good.

peterh:
Okay, the force-pushed version (just now) changes the way the system path is calculated:

- the system path is placed last, same as when gcc invokes the linker without a libpath specified
- the path from gcc -print-search-dirs is preferred over the hard-coded `/usr/lib:/lib` path (on my machine, that includes `/usr/opt/freeware/lib` and a couple other gcc-specific paths ahead of `/usr/lib:/lib`)

As much as I'd prefer the system paths ahead of the freeware/lib dirs, I can't argue with "what gcc does by default".

@andreaskem does this version work for you?

peterh:
...if there are any other CI failures, I'll take a look tomorrow.

/me shakes fist in general direction of linters.

peterh:
Those two Azure failures have a 100% fail rate since January 24. I don't think they were caused by this merge request.

dcbaker:
No, it's caused by a regression in cygwin (not meson).

andreaskem:
@peterh You are awesome! The new version seems to work just fine, for me. I conducted the same testing procedure as before and did not have to muck with the LIBPATH environment variable anymore.

I am not happy that non-system paths are preferred, but that is how GCC seems to handle it on AIX. On the other hand, GCC is not a native package, itself. Maybe GCC will be packaged differently by IBM in the future, but your solution should gracefully handle that.
I dare not think about what `xlc/xlC` support should look like on AIX, though.

(I'm still not sure about my pkgconf problem, but that still seems completely unrelated)

Thank you very much!

PR #8324, opened 2021-02-10
Update SimpleStart.md

IvanVnucec:
Running `meson compile` command in `build` directory resulted in error "ERROR: Neither directory contains a build file meson.build.". Running `ninja` command in the `build` directory worked fine.

How to recreate: 
Follow the steps for Ubuntu https://mesonbuild.com/SimpleStart.html

System: Ubuntu 20.xx
Meson 0.53.2

### DIFF
--- a/docs/markdown/SimpleStart.md
+++ b/docs/markdown/SimpleStart.md
@@ -160,5 +160,5 @@ is put in the `build` subdirectory and can be run directly from there.
 ![All finished and ready to go](images/linux_alldone.png)
 
 The project is now ready for development. You can edit the code with
-any editor and it is rebuilt by going in the `build` subdirectory and
-executing the `meson compile` command.
+any editor and it is rebuilt by executing the `ninja` command in the
+`build` directory.


IvanVnucec:
Also [here](https://mesonbuild.com/Tutorial.html) I found a lot of outdated documentation.

eli-schwartz:
The `meson compile` command got added in meson 0.54, a lot of documentation assumes recent meson.

https://mesonbuild.com/Release-notes-for-0-54-0.html#backend-agnostic-compile-command

eli-schwartz:
BTW I continue to believe it was a mistake that the first argument to meson can be either a subcommand or a directory name, and depending on the version of meson a missing subcommand tries to run the `meson setup` subcommand with "compile" as the directory name.

IvanVnucec:
> BTW I continue to believe it was a mistake that the first argument to meson can be either a subcommand or a directory name, and depending on the version of meson a missing subcommand tries to run the `meson setup` subcommand with "compile" as the directory name.

But if you follow the steps in the provided link, you will end up installing 0.53.2 Meson version. I've followed the steps today.

jpakkane:
The text should keep using `meson compile` but note that for old versions of Meson you might need to run `ninja` instead.

IvanVnucec:
The problem is then if you follow the SimpleStart documentation, you will end up installing outdated version and if you continue to follow SimpleStart documentation for Ubuntu, you will be using wrong commands which would result in using wrong commands.

jpakkane:
I fixed this directly in master instead, so closing. Thanks.

Issue #8318, opened 2021-02-08
The TestProtocol enum does weird stuff in __str__, accessing other enum members via self


hroncok:
Hello. We are testing Python packages in Fedora with Python 3.10 to see what breaks. Packages using meson fail with:

```
Found ninja-1.10.2 at /usr/bin/ninja
  File "/usr/lib/python3.10/site-packages/mesonbuild/mesonmain.py", line 140, in run
    return options.run_func(options)
  File "/usr/lib/python3.10/site-packages/mesonbuild/msetup.py", line 253, in run
    app.generate()
  File "/usr/lib/python3.10/site-packages/mesonbuild/msetup.py", line 161, in generate
    self._generate(env)
  File "/usr/lib/python3.10/site-packages/mesonbuild/msetup.py", line 235, in _generate
    mintro.generate_introspection_file(b, intr.backend)
  File "/usr/lib/python3.10/site-packages/mesonbuild/mintro.py", line 470, in generate_introspection_file
    intro_info += [(key, val.func())]
  File "/usr/lib/python3.10/site-packages/mesonbuild/mintro.py", line 77, in <lambda>
    ('tests', IntroCommand('List all unit tests', func=lambda: list_tests(testdata))),
  File "/usr/lib/python3.10/site-packages/mesonbuild/mintro.py", line 342, in list_tests
    return get_test_list(testdata)
  File "/usr/lib/python3.10/site-packages/mesonbuild/mintro.py", line 336, in get_test_list
    to['protocol'] = str(t.protocol)
  File "/usr/lib/python3.10/site-packages/mesonbuild/backend/backends.py", line 59, in __str__
    if self is self.EXITCODE:
  File "/usr/lib64/python3.10/enum.py", line 146, in __get__
    raise AttributeError(
AttributeError: TestProtocol: no attribute 'EXITCODE'
```

This comes from:

https://github.com/mesonbuild/meson/blob/398df5629863e913fa603cbf02c525a9f501f8a8/mesonbuild/backend/backends.py#L52-L78

The `__str__` method accesses other enum members via `self.<name>`.

I've reported this as a regression to Python, because it used to work prior to 3.10.0a5 and no longer does. It seems however, that the author of enum doesn't think this is a proper API, and while they agree to keep it working in Python 3.10, they might end up adding a deprecation warning. https://bugs.python.org/issue43162

Possibly, the method could be simplified to:

```python
     def __str__(self) -> str: 
         return self.name.lower()
```

?

[...]

PR #8332, opened 2021-02-12
Don't access other enum members from self

hroncok:
This is deprecated in Python 3.10.
Instead, acces them from the class.

Fixes https://github.com/mesonbuild/meson/issues/8318

### DIFF
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -69,11 +69,12 @@ class TestProtocol(enum.Enum):
         raise MesonException('unknown test format {}'.format(string))
 
     def __str__(self) -> str:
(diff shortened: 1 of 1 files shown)


dcbaker:
@jpakkane we should get this in sooner than later, as the existing code will probably stop working in 3.12 or 3.13

jpakkane:
I'll merge this as soon as #8362 is in and this can be rebased to rerun the tests.

dcbaker:
@hroncok: could you rebase this, so we can get the unit tests working? I'd like to merge this ASAP

hroncok:
Done.

hroncok:
Thanks!

PR #8335, opened 2021-02-12
Support mixing static and generated Rust sources

dcbaker:
Rust has some different constraints than C like-languages. We can't compile each source file into an object file and link that together, instead the compiler opens the "main" Rust file, then opens additional files as they are used, this is similar to how python works, but without the ability overlay multiple paths using `$PYTHON_PATH` or `sys.path`. This means if we generate any rust sources (either with a `custom_target` or `configure_file`), then everything falls over as some sources are in the source directory, and some are in the build directory. Sometimes this limitation can be worked around by compiling one set of sources into an rlib and linking them into the other, but sometimes that doesn't work. It's also a pain, and not very friendly.

With this series that is fixed, at least when using `configure_file` and `custom_target` for the source generating. Meson will now detect that there is a mixture of generated and static sources, and add compile time rules to copy the static sources to the build tree before compiling in the build tree. I chose this route instead of reconfiguring, as running a few copy operations before recompiling is likely much faster than reconfiguring, which makes editing a single Rust source and rebuilding faster.


jpakkane:
That VS failure seems weird. Let's rebase to see if it was a fluke.

jpakkane:
The vs failure seems relevant:

```
Found ninja-1.10.0 at D:\a\ci_data\ninja.EXE
[1/7] "copy" "../test cases/rust/13 mixed generated static sources/src/main2.rs" "src/main2.rs"
FAILED: src/main2.rs 
"copy" "../test cases/rust/13 mixed generated static sources/src/main2.rs" "src/main2.rs"
CreateProcess failed: The system cannot find the file specified.
[2/7] "copy" "../test cases/rust/13 mixed generated static sources/src/main1.rs" "src/main1.rs"
FAILED: src/main1.rs 
"copy" "../test cases/rust/13 mixed generated static sources/src/main1.rs" "src/main1.rs"
CreateProcess failed: The system cannot find the file specified.
``´

Issue #8337, opened 2021-02-13
Unable to link simple CUDA project


andrzejc:
**Describe the bug**

Following the advice from https://github.com/mesonbuild/meson/issues/8334, I decided to convert my project to use built-in CUDA support rather than `generator()`. I was using meson testcases (https://github.com/mesonbuild/meson/tree/master/test%20cases/cuda) as an example how to do that.
When building a simple project (see the attachment) using built-in CUDA support, which includes `dependency('threads')`, the linking fails with the following output:

```
  [6/6] Linking target cuda-linking-bug
  FAILED: cuda-linking-bug 
  /usr/local/cuda-10.0/bin/nvcc  -o cuda-linking-bug cuda-linking-bug.p/test_main.cpp.o cuda-linking-bug.p/test_cuda.cu.o libcuda-linking-bug.a -Xlinker=-no-pie -Xlinker=-pthread -Xlinker=/usr/lib/x86_64-linux-gnu/libboost_thread.a -Xlinker=/usr/lib/x86_64-linux-gnu/libboost_system.a -Xlinker=/usr/lib/x86_64-linux-gnu/libboost_filesystem.a -Xlinker=-pthread -lcudart -lcublas -lcurand -Xlinker=-pthread -Xlinker=-pthread -lcudart -lcublas -lcurand -lstdc++ -lstdc++
  /usr/bin/ld: unrecognized option '-pthread'
  /usr/bin/ld: use the --help option for usage information
  collect2: error: ld returned 1 exit status
  ninja: build stopped: subcommand failed.
```

As it seems that meson is trying to pass `-pthread` compiler flag to `ld` invocation via `-Xlinker=` nvcc argument, I tried to workaround it by writing `nvcc-wrapper` script (included in the tarball) which removes this argument (or rather replaces it with `-Xcompiler=-pthread`, this doesn't matter for the linking phase). When using the wrapper the linking still fails:

```
  [6/6] Linking target cuda-linking-bug
  FAILED: cuda-linking-bug 
[...]

PR #8355, opened 2021-02-15
CUDA Compiler Fixes

obilaniu:
This is an extensive rewrite and hopefully hardening of the CUDA NVCC compiler object and its flag handling.

The class method `_to_host_flags()` is entirely rewritten and gains knowledge of the entire `nvcc` argument set in the [official documentation](https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html). The method now translates an array of mixed GCC/NVCC or MVSC/NVCC flags into a legal argument set for `nvcc`, exploiting `-Xcompiler` and `-Xlinker` to pass down foreign arguments into the compiler. It does the best it can to take into account and paper over the peculiarities of NVCC argument parsing.

Builds successfully the entire `test cases/cuda` testsuite on my machine (with CUDA Toolkit 10.2).

Related to #8337. @andrzejc, would you like to test out this branch?

### DIFF
--- a/mesonbuild/compilers/cuda.py
+++ b/mesonbuild/compilers/cuda.py
@@ -14,6 +14,7 @@
 
 import enum
 import os.path
+import string
 import typing as T
 
 from .. import coredata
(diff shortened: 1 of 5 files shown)


obilaniu:
CI failure is real - somewhere the DT_RPATH `$ORIGIN` is getting expanded into nothingness.

andrzejc:
> Related to #8337. @andrzejc, would you like to test out this branch?

Thank you @obilaniu, this is awesome, I'm definitely gonna give it a try.

obilaniu:
@andrzejc I have solved the CUDA CI failure. If you have Linux, this should definitely work. If you have Windows, it would be extremely valuable if you tried this out to ensure it does not break Windows CUDA.

andrzejc:
@obilaniu I just finished testing with Linux box, seems to work perfectly, thank you! I'll check on Windows tomorrow.

obilaniu:
@andrzejc With the rebase of an hour ago or yesterday?

andrzejc:
> @andrzejc With the rebase of an hour ago or yesterday?

with the rebase

andrzejc:
@obilaniu so I was curious about Windows and played with building with MSVC a moment ago and it also seems to work without any issues. I was able to compile & link my project with MSVC2017/CUDA 10.0 configuration and all the tests are passing just like they did with the generator approach. Awesome!

obilaniu:
@andrzejc Great! @jpakkane I think this is ready for review.

mesonbuild/compilers/cuda.py
```diff
@@ -60,23 +61,423 @@ def __init__(self, exelist: T.List[str], version: str, for_machine: MachineChoic
...
+        BS = "\\"
+        DQSQ = DQ+SQ+DQ
+        quotable = set(string.whitespace+'"$`\\')
+
+        if CM not in arg or not listmode:
+            if SQ not in arg:
+                """
+                If any of the special characters "$`\ or whitespace are present, single-quote.
+                Otherwise return bare.
+                """
```
dcbaker, line 88: These aren't docstrings, they should use the `# comment format`. A docstring must come directly after a `def` or `class` definition

obilaniu: > These aren't docstrings, they should use the `# comment format`. A docstring must come directly after a `def` or `class` definition

This is true, but triple-quoted strings elsewhere are simply assignments of a string literal to nothing in particular.

I can certainly convert them all to packs of `#`-prefixed comments.

mesonbuild/compilers/cuda.py
```diff
@@ -60,23 +61,423 @@ def __init__(self, exelist: T.List[str], version: str, for_machine: MachineChoic
...
+            '--prec-sqrt':                          '-prec-sqrt',                   # 4.2.7.10
+            '--fmad':                               '-fmad',                        # 4.2.7.11
+            '--Werror':                             '-Werror',                      # 4.2.8.9
+            '--diag-error':                         '-diag-error',                  # 4.2.8.11
+            '--diag-suppress':                      '-diag-suppress',               # 4.2.8.12
+            '--diag-warn':                          '-diag-warn',                   # 4.2.8.13
+            '--options-file':                       '-optf',                        # 4.2.8.17
+            '--time':                               '-time',                        # 4.2.8.18
+            '--qpp-config':                         '-qpp-config',                  # 4.2.8.19
+        }
```
dcbaker, line 279: I would avoid creating a huge dict like this inside a function body, since these will be recreated each time the function is called. These should be module or class level private.

mesonbuild/compilers/cuda.py
```diff
@@ -60,23 +61,423 @@ def __init__(self, exelist: T.List[str], version: str, for_machine: MachineChoic
...
+        flag_short2long_withargs = {v:k for k,v in flag_long2short_withargs.items()}
```
dcbaker, line 280: What are you trying to achieve here, a copy? dict has a `.copy()` method for that, which is probably faster than the comprehension.

mesonbuild/compilers/cuda.py
```diff
@@ -326,13 +732,24 @@ def get_crt_link_args(self, crt_val: str, buildtype: str) -> T.List[str]:
...
+        return self._to_host_flags(super().get_dependency_link_args(dep), _Phase.LINKER)
+
+    def get_ccbin_args(self, options: 'KeyedOptionDictType') -> T.List[str]:
+        key = OptionKey('ccbindir', machine=self.for_machine, lang=self.language)
+        try:
+            ccbindir = options[key].value
+            if isinstance(ccbindir, str) and ccbindir != '':
+                return [self._shield_nvcc_list_arg('-ccbin='+ccbindir, False)]
+            else:
+                return []
```
dcbaker, line 753: I'd pull the `if/else` out of the try/except block, since now if there were an unhandled `KeyError `in `self._shield_nvcc_list_arg`, it would get silently caught here.

dcbaker (changes requested):
I've got a few style nits, but nothing serious.

mesonbuild/compilers/cuda.py
obilaniu: @dcbaker A reverse map. It's `v:k for k,v`.

mesonbuild/compilers/cuda.py
obilaniu: I had this `try` only because I had an inexplicable `KeyError` with key `ccbindir` somewhere in the Arch CUDA CI, despite the flag being defaulted to empty and thus "absent", not changing the CI to try to use it, and the log messages all saying the CUDA tests had _passed_.

I'll try to strike it out.

obilaniu: @dcbaker Could you please have a look at that failure on the Arch CUDA runner and tell me what I'm doing wrong?

I introduced a new `-Dcuda_ccbindir=path` option, see `mesonbuild/compilers/cuda.py:582`, and it works for me in practice, but `run_unittests.py` chokes on it. I suspect it's bad mocking, but did I do everything I was supposed to do in order to register this new option?

dcbaker: That took me a while to figure out, the problem is that unittest is skipping some important stuff to get the compiler, replace the `getattr(env, 'detect_foo_comiler')` line with `env.detect_compiler_for(l, MachineChoice.HOST)`, that should fix it.

dcbaker (approved):

mesonbuild/compilers/cuda.py
obilaniu: @dcbaker You were right; The Arch CUDA runner has just passed.

PR #8358, opened 2021-02-16
Fix destdir detection for installation via polkit

xry111:
In 0.57.0 installation via polkit won't work anymore. `destdir` is
defaulted to an empty string ('') instead of None.  But polkit
installation code incorrectly tests `destdir is None`.


xclaesse:
This seems to be a regression since https://github.com/mesonbuild/meson/pull/8081, thanks for the fix.

We are trying to catch those regressions earlier by doing RC releases, we would appreciate if more people could test future Meson releases in advance to avoid this kind of issue. If you wish to help us, please follow `@mesonbuild` on twitter to be notified about RC releases, test your projects, and report issues before final release. Thanks.

PR #8362, opened 2021-02-16
unittests: expect mtest logs to be utf-8

bonzini:
I noticed a UnicodeDecodeError in the logs for #8332. This should fix it.

### DIFF
--- a/run_unittests.py
+++ b/run_unittests.py
@@ -2530,12 +2530,12 @@ class AllPlatformTests(BasePlatformTests):
         self.build()
         # Run tests without setup
         self.run_tests()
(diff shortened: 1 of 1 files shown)


jpakkane:
Did this change during test refactorings during the last cycle or is the actual error older? Or, in other words, does this need to go to 0.57.1?

bonzini:
The 0.57 changes did not affect the text logs, so I'm not sure why it was not visible before. I also am not sure why the problem would surface specifically after #8362; for that matter, it seems to me that `TestProtocol.__str__` is dead code (it is okay to have it for e.g. debugging purposes but it should never be invoked) so I don't think either of those are 0.57 material.

Issue #8357, opened 2021-02-15
Add `@CURRENT_SOURCE_DIR@` to `custom_command`


mdegans:
**Describe the bug**
Continuing from #7511

There may be a possible regression with 0.57 where `@CURRENT_SOURCE_DIR@` has been removed from `custom_target`. I realize that's not in the manual, but removing `CURRENT_SOURCE_DIR` broke my build and I can't figure out  a working replacement. 

`0.56.2`  generates this `build.ninja`
```
...
build proto/settings.pb.cc: CUSTOM_COMMAND ../proto/settings.proto | /usr/bin/protoc
 COMMAND = /usr/bin/protoc --proto_path=../proto --cpp_out=proto ../proto/settings.proto
 description = Generating$ settings_pb_cc$ with$ a$ custom$ command

build proto/settings.pb.h: CUSTOM_COMMAND ../proto/settings.proto | /usr/bin/protoc
 COMMAND = /usr/bin/protoc --proto_path=../proto --cpp_out=proto ../proto/settings.proto
 description = Generating$ settings_pb_h$ with$ a$ custom$ command
 ...
 ```
 `0.57` generates:
 ```
 build proto/settings.pb.cc: CUSTOM_COMMAND ../proto/settings.proto | /usr/bin/protoc
 COMMAND = /usr/bin/protoc --proto_path=@CURRENT_SOURCE_DIR@ --cpp_out=proto ../proto/settings.proto
 description = Generating$ settings_pb_cc$ with$ a$ custom$ command

build proto/settings.pb.h: CUSTOM_COMMAND ../proto/settings.proto | /usr/bin/protoc
 COMMAND = /usr/bin/protoc --proto_path=@CURRENT_SOURCE_DIR@ --cpp_out=proto ../proto/settings.proto
 description = Generating$ settings_pb_h$ with$ a$ custom$ command
 ```

**To Reproduce**

meson.build
```meson
protobuf = dependency('protobuf', version: '>=3.0.0')
protoc = find_program('protoc')

# many thanks to @marcelhollerbach for his solution.
# https://github.com/mesonbuild/meson/issues/7511#issuecomment-665470520

settings_pb_cc = custom_target('settings_pb_cc',
  input: ['settings.proto'],
  output: ['@BASENAME@.pb.cc'],
  command: [
    protoc,
    '--proto_path=@CURRENT_SOURCE_DIR@',
    '--cpp_out=@OUTDIR@',
    '@INPUT@',
  ]
)

[...]

PR #8363, opened 2021-02-16
Substitute @CURRENT_SOURCE_DIR@ in run_target() and custom_target()

xclaesse:
run_target() does some variable substitutions since 0.57.0. This is a
new behavior, and undocumented, caused by sharing more code with
custom_target(). More consistency is better, so document it now.

custom_target() was doing variable substitution in the past, because it
shared some code with generator(), but that was undocumented. Some
refactoring in 0.57.0 caused it to not replace @CURRENT_SOURCE_DIR@,
@SOURCE_DIR@, and @BUILD_DIR@ anymore. This patch adds back
@CURRENT_SOURCE_DIR@ and document it. It does not add back @SOURCE_DIR@
because it is duplicate with @SOURCE_ROOT@ that has a better name. Also
do not add back @BUILD_DIR@ which is duplicate of @PRIVATE_DIR@, and
not @BUILD_ROOT@ surprisingly, adding to the confusion.

### DIFF
--- a/docs/markdown/Reference-manual.md
+++ b/docs/markdown/Reference-manual.md
@@ -402,6 +402,13 @@ the following special string substitutions:
 - `@PLAINNAME@`: the input filename, without a path
 - `@BASENAME@`: the input filename, with extension removed
 - `@PRIVATE_DIR@` *(since 0.50.1)*: path to a directory where the custom target must store all its intermediate files.
+- `@SOURCE_ROOT@`: the path to the root of the source tree. Depending on the backend,
+  this may be an absolute or a relative to current workdir path.
+- `@BUILD_ROOT@`: the path to the root of the build tree. Depending on the backend,
+  this may be an absolute or a relative to current workdir path.
+- `@CURRENT_SOURCE_DIR@`: this is the directory where the currently
+  processed meson.build is located in.  Depending on the backend,
+  this may be an absolute or a relative to current workdir path.
 
 *(since 0.47.0)* The `depfile` keyword argument also accepts the
  `@BASENAME@` and `@PLAINNAME@` substitutions.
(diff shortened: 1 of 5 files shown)


xclaesse:
Note for future consistency improvements: We should deprecate `@SOURCE_DIR@` and `@BUILD_DIR@` in generator() and replace them with `@SOURCE_ROOT@` and `@PRIVATE_DIR@`.

xclaesse:
This is not strictly speaking regressions since those are all undefined behavior, but still tagging for 0.57.1 because people were depending on that implementation detail. See https://github.com/mesonbuild/meson/issues/8357.

docs/markdown/Reference-manual.md
```diff
@@ -402,6 +402,10 @@ the following special string substitutions:
 - `@PLAINNAME@`: the input filename, without a path
 - `@BASENAME@`: the input filename, with extension removed
 - `@PRIVATE_DIR@` *(since 0.50.1)*: path to a directory where the custom target must store all its intermediate files.
+- `@SOURCE_ROOT@`: the full path to the root of the source tree.
+- `@BUILD_ROOT@`: the full path to the root of the build tree.
+- `@CURRENT_SOURCE_DIR@`: this is the directory where the currently
+  processed meson.build is located in.
```
jpakkane, line 408: This should probably mention that it is a relative path from the cwd where the target was executed.

xclaesse: That's another weirdness we should eventually tackle some day. With vs backend they are absolute paths, I don't know why. But documenting it as `the full path` is indeed misleading.

jpakkane: Maybe something like "depending on the backend, this may be an absolute or a relative path" then?

xclaesse: done.

tp-m:
I'm about to use `@BUILD_ROOT@` with `custom_target` in fontconfig now, so would be nice to get it documented :+1:

Issue #8361, opened 2021-02-16
Loses flags when defining both CPPFLAGS and CFLAGS


jpakkane:
This is Debian bug https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=982865

The issue is that if you specify both `CFLAGS` and `CPPFLAGS`, then `CPPFLAGS` is ignored, probably because some variable is overwritten.

Bisection points to commit 4b0b44aface1cd9ed073733.

mbiebl:
duplicate of #8345 ?

evverx:
> duplicate of #8345 ?

@mbiebl thanks! Looks like it is a duplicate of that issues. I didn't notice that all the flags got lost there. I'll close that issue in favor of this one shortly.

@jpakkane Looking at https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=982865#34, given that it appears just building large projects like systemd (which seems to have caught all the C regressions with its CI in about five minutes) in real-life scenarios would help to catch issues like this, I think that instead of relying on people manually testing broken prereleases or trying to gate the meson package somewhere downstream it would be better to test meson much more thoroughly upstream. It's kind of being discussed in https://github.com/mesonbuild/meson/issues/8347

Issue #8345, opened 2021-02-15
Regression: the "sanitizer" flags passed via CFLAGS and CXXFLAGS are no longer passed to the linker


evverx:
**Describe the bug**
Since meson-0.57.0 was released `systemd` has been failing to compile on [CIFuzz](https://github.com/systemd/systemd/runs/1899186196?check_suite_focus=true) and [OSS-Fuzz](https://oss-fuzz-build-logs.storage.googleapis.com/log-9e1caee7-f12e-4d66-8f17-d026c57e54d5.txt) with
```
2021-02-14T21:23:53.3854577Z [481/664] clang++  -o fuzz-compress fuzz-compress.p/src_fuzz_fuzz-compress.c.o -L/usr/lib/clang/12.0.0/lib/linux -Wl,--as-needed -Wl,--allow-shlib-undefined '-Wl,-rpath,$ORIGIN/src/shared' -Wl,-rpath-link,/work/build/src/shared -Wl,--start-group src/shared/libsystemd-shared-247.so -lFuzzingEngine -Wl,--end-group
2021-02-14T21:23:53.3856578Z FAILED: fuzz-compress
2021-02-14T21:23:53.3858714Z clang++  -o fuzz-compress fuzz-compress.p/src_fuzz_fuzz-compress.c.o -L/usr/lib/clang/12.0.0/lib/linux -Wl,--as-needed -Wl,--allow-shlib-undefined '-Wl,-rpath,$ORIGIN/src/shared' -Wl,-rpath-link,/work/build/src/shared -Wl,--start-group src/shared/libsystemd-shared-247.so -lFuzzingEngine -Wl,--end-group
2021-02-14T21:23:53.4160985Z /usr/lib/gcc/x86_64-linux-gnu/5.4.0/../../../libFuzzingEngine.a(FuzzerTracePC.o): In function `size':
2021-02-14T21:23:53.4162332Z /usr/local/bin/../include/c++/v1/set:641: undefined reference to `__asan_report_load8'
2021-02-14T21:23:53.4164959Z /usr/lib/gcc/x86_64-linux-gnu/5.4.0/../../../libFuzzingEngine.a(FuzzerTracePC.o): In function `fuzzer::TracePC::HandleInline8bitCountersInit(unsigned char*, unsigned char*)':
2021-02-14T21:23:53.4166954Z /src/libfuzzer/FuzzerTracePC.cpp:61: undefined reference to `__asan_report_store1'
[...]

PR #8366, opened 2021-02-16
Environment: Fix passing envrionment both CPPFLAGS and CFLAGS

dcbaker:
Or other language flags that use CPPFLAGS (like CXXFLAGS). The problem
here is actually rather simple, `dict.setdefault()` doesn't work like I
thought it did, I thought it created a weak entry, but it actually is
equivalent to:
```python
if k not in dict:
    dict[k] = v
```
Instead we'll use an intermediate dictionary (a default dictionary
actually, since that makes things a little cleaner) and then add the
keys from that dict to self.options as applicable.

Test case written by Jussi, Fix by Dylan

Co-authored-by: Jussi Pakkanen
Fixes: #8361
Fixes: #8345

### DIFF
--- a/mesonbuild/environment.py
+++ b/mesonbuild/environment.py
@@ -788,6 +788,8 @@ class Environment:
             ]
         )
 
+        env_opts: T.DefaultDict[OptionKey, T.List[str]] = collections.defaultdict(list)
+
         for (evar, keyname), for_machine in itertools.product(opts, MachineChoice):
             p_env = _get_env_var(for_machine, self.is_cross_build(), evar)
             if p_env is not None:
@@ -805,7 +807,7 @@ class Environment:
                     p_list = list(mesonlib.OrderedSet(p_env.split(':')))
                 else:
                     p_list = split_args(p_env)
-                p_list = [e for e in p_list if e]  # filter out any empty eelemnts
+                p_list = [e for e in p_list if e]  # filter out any empty elements
 
                 # Take env vars only on first invocation, if the env changes when
                 # reconfiguring it gets ignored.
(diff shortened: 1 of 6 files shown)


mesonbuild/environment.py
```diff
@@ -805,7 +807,7 @@ def _set_default_options_from_env(self) -> None:
                     p_list = list(mesonlib.OrderedSet(p_env.split(':')))
                 else:
                     p_list = split_args(p_env)
-                p_list = [e for e in p_list if e]  # filter out any empty eelemnts
+                p_list = [e for e in p_list if e]  # filter out any empty elemnts
```
obilaniu, line 810: Awkward, but... `elements`.

dcbaker: gahh, my spell checker has decided that it will *not* not matter what I tell it check .py files anymore :cry:

obilaniu:
By the way, re: continuous integration for this PR, it would be worth it to find a way to abort [running tests](https://dev.azure.com/jussi0947/jussi/_build?definitionId=1&_a=summary) on Azure DevOps for branches that have been pushed (or force-pushed, at least).

dcbaker:
I thought we were already doing that! if we're not then yes, yes, would should find a way to do that.

obilaniu:
@dcbaker As you can see, _all_ the (force-)pushes you made are being tested at this present time. The Azure runners haven't taken only the branch tip, nor would they abort if you made another push.

dcbaker:
It looks like we're using the wrong type of pipelines, if we used PR triggers (https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#pr-triggers) then we'd get the behavior we wanted.

Issue #8354, opened 2021-02-15
regression 0.57.0 build.pkg_config_path broken


asavah:
**Describe the bug**
build.pkg_config_path is broken in 0.57.0 when crosscompiling.

**To Reproduce**
Using toolchain.meson hint file with following lines worked flawlessly in 0.56.x
```
[built-in options]
pkg_config_path = '/home/asavah/kross/build/asusb450eg/ovr/pkgconfig/:/home/asavah/kross/build/asusb450eg/rootfs/usr/lib/pkgconfig:/home/asavah/kross/build/asusb450eg/rootfs/usr/share/pkgconfig'
build.pkg_config_path = '/home/asavah/kross/host/lib/pkgconfig:/home/asavah/kross/host/share/pkgconfig:/usr/lib/pkgconfig:/usr/share/pkgconfig:/usr/lib/x86_64-linux-gnu/pkgconfig'
```

As per documentation build.pkg_config_path was used to specify locations of native .pc stuff

0.57.0 does some wierd appending of both build and host pkg_config_path

```
Library acl found: YES
Pkg-config binary for MachineChoice.HOST is cached.
Determining dependency 'blkid' with pkg-config executable '/home/asavah/kross/host/asusb450eg/bin/x86_64-asusb450eg-linux-gnu-pkg-config'
PKG_CONFIG_PATH: /home/asavah/kross/host/lib/pkgconfig:/home/asavah/kross/host/share/pkgconfig:/usr/lib/pkgconfig:/usr/share/pkgconfig:/usr/lib/x86_64-linux-gnu/pkgconfig
Called `/home/asavah/kross/host/asusb450eg/bin/x86_64-asusb450eg-linux-gnu-pkg-config --modversion blkid` -> 0
2.36.597
PKG_CONFIG_PATH: /home/asavah/kross/host/lib/pkgconfig:/home/asavah/kross/host/share/pkgconfig:/usr/lib/pkgconfig:/usr/share/pkgconfig:/usr/lib/x86_64-linux-gnu/pkgconfig
Called `/home/asavah/kross/host/asusb450eg/bin/x86_64-asusb450eg-linux-gnu-pkg-config --cflags blkid` -> 0
-I/home/asavah/kross/build/asusb450eg/rootfs/home/asavah/kross/host/include/blkid
[...]

Issue #8351, opened 2021-02-15
Regression: AssertionError: build.b_vscrt


Neumann-A:
```
Traceback (most recent call last):
  File "E:\vcpkg_cache\downloads\tools\meson\meson-5d651e76b5b6dba5dd03ffec609719b02d703bb7\mesonbuild\mesonmain.py", line 132, in run
    return options.run_func(options)
  File "E:\vcpkg_cache\downloads\tools\meson\meson-5d651e76b5b6dba5dd03ffec609719b02d703bb7\mesonbuild\msetup.py", line 275, in run
    app.generate()
  File "E:\vcpkg_cache\downloads\tools\meson\meson-5d651e76b5b6dba5dd03ffec609719b02d703bb7\mesonbuild\msetup.py", line 176, in generate
    env = environment.Environment(self.source_dir, self.build_dir, self.options)
  File "E:\vcpkg_cache\downloads\tools\meson\meson-5d651e76b5b6dba5dd03ffec609719b02d703bb7\mesonbuild\environment.py", line 643, in __init__
    self._load_machine_file_options(config, properties.build, MachineChoice.BUILD)
  File "E:\vcpkg_cache\downloads\tools\meson\meson-5d651e76b5b6dba5dd03ffec609719b02d703bb7\mesonbuild\environment.py", line 772, in _load_machine_file_options
    key = OptionKey.from_string(k).evolve(subproject=subproject, machine=machine)
  File "E:\vcpkg_cache\downloads\tools\meson\meson-5d651e76b5b6dba5dd03ffec609719b02d703bb7\mesonbuild\mesonlib\universal.py", line 2061, in evolve
    return OptionKey(
  File "E:\vcpkg_cache\downloads\tools\meson\meson-5d651e76b5b6dba5dd03ffec609719b02d703bb7\mesonbuild\mesonlib\universal.py", line 1959, in __init__
    _type = _classify_argument(self)
  File "E:\vcpkg_cache\downloads\tools\meson\meson-5d651e76b5b6dba5dd03ffec609719b02d703bb7\mesonbuild\mesonlib\universal.py", line 1913, in _classify_argument
    assert key.machine is MachineChoice.HOST, str(key)
AssertionError: build.b_vscrt
```
Used native files:
[meson-nativ-x64-windows.log](https://github.com/mesonbuild/meson/files/5982364/meson-nativ-x64-windows.log)
[...]

PR #8367, opened 2021-02-16
allow build.b_* options

dcbaker:
These continue to be ignored as they always have, but no longer raise an
error.

Fixes: #8351
Fixes: #8354

### DIFF
--- a/mesonbuild/environment.py
+++ b/mesonbuild/environment.py
@@ -769,13 +769,20 @@ class Environment:
                 subproject = ''
             if section == 'built-in options':
                 for k, v in values.items():
-                    key = OptionKey.from_string(k).evolve(subproject=subproject, machine=machine)
-                    self.options[key] = v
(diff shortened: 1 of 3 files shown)


jpakkane:
Now that we have native files it might make sense to prohibit setting build machine options in cross files (if I had realized that I would have made it an error when the feature was originally created). We can't break this immediately, but maybe we could print a deprecation warning that this might go away at some point.

dcbaker:
I'm proposing removing it in 0.60, since the correct solution has always worked, but I've fixed it to work correctly now, and added the deprecation warning.

jpakkane:
Test failures seem relevant.

dcbaker:
@jpakkane I have set the deprecation warning to mention 0.60 as the removal target, since the "right way" has always been available. Is that okay, or do you want me to change that?

jpakkane:
Please rebase. It should go green.

PR #8376, opened 2021-02-17
mtest: use / as path separator for ninja targets

bonzini:
os.path.relpath(f, wd) returns path with \ seperator on Windows, but ninja targets always uses / separator.

See for example https://gitlab.freedesktop.org/ocrete/libnice/-/jobs/7348274.

### DIFF
--- a/mesonbuild/mtest.py
+++ b/mesonbuild/mtest.py
@@ -1887,6 +1887,12 @@ def list_tests(th: TestHarness) -> bool:
     return not tests
 
 def rebuild_deps(wd: str, tests: T.List[TestSerialisation]) -> bool:
+    def convert_path_to_target(path: str) -> str:
+        path = os.path.relpath(path, wd)
+        if os.sep != '/':
+            path = path.replace(os.sep, '/')
+        return path
+
     if not (Path(wd) / 'build.ninja').is_file():
         print('Only ninja backend is supported to rebuild tests before running them.')
         return True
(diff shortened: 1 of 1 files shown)


mesonbuild/mtest.py
```diff
@@ -1901,7 +1907,7 @@ def rebuild_deps(wd: str, tests: T.List[TestSerialisation]) -> bool:
     intro_targets = dict()     # type: T.Dict[str, T.List[str]]
     for target in load_info_file(get_infodir(wd), kind='targets'):
         intro_targets[target['id']] = [
-            os.path.relpath(f, wd)
+            convert_path_to_target(f)
```
xclaesse, line 1910: We try to convert our code slowly to pathlib. You can use something like Path().relative_to().Unix()

xclaesse: Sorry it is as_posix()

bonzini: There is the following scary note in the pathlib documentation:

> Although os.path.relpath() and PurePath.relative_to() have some overlapping use-cases, their semantics differ enough to warrant not considering them equivalent.

I _think_ it refers to raising ValueError if (in the case above) `f` is not representable as a path from `wd`, which cannot happen here, but I wasn't sure if there's something else.

I could use `relative_to` and `as_posix()`, but it seemed hackish because ninja does not really accept POSIX paths as much as "Windows paths with forward slashes" (see for example drive letters). It's just that the slash is the only difference between relative POSIX paths and relative Windows paths. But I can also rewrite it that way if you prefer.

xclaesse: Bah, it's just nitpicking, was suggesting that because it's a one liner. But the result is really just the same :-)

LGTM.

PR #8381, opened 2021-02-17
Disable failing Cygwin GIR test.

jpakkane:
This will fail at first because I disabled only one as a test.


jpakkane:
It now fails with an underscore test, which was _definitely_ not there in the previous run. I checked. Merging this as it fixes a lot of the issues and will fix the underscore one later if it still occurs.

PR #8386, opened 2021-02-18
Summarize passed subtests even on overall failure

ldrumm:
It's nice to know how many subtests passed even if there was an overall
failure in a multi-subtest `test()` invocation

### DIFF
--- a/mesonbuild/mtest.py
+++ b/mesonbuild/mtest.py
@@ -898,17 +898,18 @@ class TestRun:
     def detail(self) -> str:
         if self.res is TestResult.PENDING:
             return ''
-        if self.returncode:
-            return returncode_to_status(self.returncode)
+        statuses = []
         if self.results:
             # running or succeeded
             passed = sum((x.result.is_ok() for x in self.results))
             ran = sum((x.result is not TestResult.SKIP for x in self.results))
             if passed == ran:
-                return '{} subtests passed'.format(passed)
+                statuses.append(f'{passed} subtests passed')
             else:
-                return '{}/{} subtests passed'.format(passed, ran)
-        return ''
+                statuses.append(f'{passed}/{ran} subtests passed')
+        if self.returncode:
+            statuses.append(returncode_to_status(self.returncode))
+        return '; '.join(statuses)
 
     def _complete(self, returncode: int, res: TestResult,
                   stdo: T.Optional[str], stde: T.Optional[str]) -> None:


bonzini:
I suspect you have a buggy TAP harness that returns nonzero exit code when some subtests fail?

The problem with the change is that the line becomes really, really long. What do you think about moving the "statuses" to a separate line, as in [this older mockup](https://asciinema.org/a/QhVvDrozZ4TJ6OHixgy5ohLXI).

ldrumm:
> I suspect you have a buggy TAP harness that returns nonzero exit code
> when some subtests fail?
It does exit with a status, but I don't think that is a bug. Neither classic TAP
nor TAP13 specify the exit status (they both mark TOOO at the bottom of their
respective pages[1][2]).

Google test exits with status too which is another multi-test protocol supported
by Meson.

> The problem with the change is that the line becomes really, really
> long. 
Here is the output from the present patch:

> 1/1 lit        ERROR           2.01s   68/101 subtests passed; exit status 1

That's 79 characters. I don't think that's egregious.

> What do you think about moving the "statuses" to a separate line,
> as in [this older
> mockup](https://asciinema.org/a/QhVvDrozZ4TJ6OHixgy5ohLXI)?

If you mean the "error blah blah" example, then that would be fine. I'm not
terribly attached to the formatting changes introduced by my patch, as long as
the pertinent information is present, so I'll willing adjust it to whatever is
most ergonomic.

[1] http://testanything.org/tap-specification.html#todo
[2] http://testanything.org/tap-version-13-specification.html#todo

bonzini:
The tap specification does not mention it, but the original harness, Perl's "prove", complains if the exit code is nonzero. I would leave out the exit code text if at least one test failed and the exit code does not represent being killed by a signal.

> If you mean the "error blah blah" example, then that would be fine

I mean having "X tests passed" on a separate line.

ldrumm:
Not all harnesses are equal and some of them return nonzero on overall failure. I don't think it's important what `prove` does because it's not the only one and in absense of a formal agreement across harnesses I think we should be graceful.

I'm honestly not fussed about the format as long as I can see the number of passed/total tests.

Do you have any preferences with the above said?

eli-schwartz:
If the TAP protocol eschews exit codes in favor of streaming messages, then surely that means the exit code is irrelevant... you could argue a non-zero exit code is a *lint* error, but that's all, I'd say.

ldrumm:
Perl's `prove` *warns* on nonzero exit, and has an `--ignore-exit` switch, so I think if we're using `prove` as our model, rather than the published spec we should behave in a similarly graceful manner:

```shell-session
$ prove --ignore-exit --exec '' sh -c <<EOF
#!/bin/sh

echo 1..3
echo ok 1 - foo
echo not ok 2 - bar
echo ok 3 - baz

exit 1
EOF
sh .. Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/3 subtests 

Test Summary Report
-------------------
sh (Wstat: 256 Tests: 3 Failed: 1)
  Failed test:  2
  Non-zero exit status: 1
Files=1, Tests=3,  0 wallclock secs ( 0.01 usr +  0.00 sys =  0.01 CPU)
Result: FAIL
```

> If the TAP protocol eschews exit codes in favor of streaming messages, then surely that means the exit code is irrelevant... you could argue a non-zero exit code is a lint error, but that's all, I'd say.

Yes. I don't think it's Meson's right to make a decision here. We should behave gracefully and print the number of passed tests regardless; I personally favour printing the exit status as it's useful information and I see no reason to behave differerently. nonzero exit status *always* means something went wrong, regardless of a higher-level streaming format so I think printing it is always helpful

ldrumm:
> @ldrumm are you going to implement this?
Sure. Sounds fine.

PR #8389, opened 2021-02-19
Add a script to run a single meson functional test case (with test.json support)

dcbaker:
It's often frustrating to me that if a single test regresses that uses a test.json I have to read the test.json and manually reconstruct that environment. That's silly, we have all the tools in place for this, we can just use them to create a single test runner than parses the test.json with the same logic that the main runner does. This is probably not super useful in CI, but as Meson developer being able to run a single test with its test.json (and even a single case from the test.json) is incredibly useful

### DIFF
--- a/run_mypy.py
+++ b/run_mypy.py
@@ -39,6 +39,7 @@ modules = [
     'mesonbuild/optinterpreter.py',
 
     'run_mypy.py',
+    'run_single_test.py',
     'tools'
 ]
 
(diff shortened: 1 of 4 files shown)


jpakkane:
Instead of creating a new script, could we instead extend the `--only` argument that `run_project_tests.py` already has. Something like:

    ./run_project_tests.py --only 'common/1 trivial'

Though that particular command line is not very good as you can't use tab completion for it.

dcbaker:
I tried that initially, It ended up being really ugly, because i basically needed all the code i already had, plus a ton of special cases in the script that were really fragile

PR #8390, opened 2021-02-20
tests/rust: Add rust case linking to zlib dependency

jljusten:
@dcbaker: I think this is a fairly minimal test case of the issue I mentioned on irc. I hope the test case is valid!

Note that I have a commented rust_args line in the meson.build file which probably doesn't belong in the final test case, but rather just shows one work-around I found.


jljusten:
Reworks:
 * Add MESON_SKIP_TEST error for darwin from polyglot test
 * Remove dependencies from executable

jljusten:
@dcbaker put this patch into #8403 along with a fix, so I will close this PR.

Issue #1635, opened 2017-04-16
Build RPATHs should not be overrideable with LD_LIBRARY_PATH, affects OSes using DT_RUNPATH such as FreeBSD and Debian


jpakkane:
Due to #1519. Should we revert or is there a fix?

nirbheek:
FWICT, the problem is in the order in which the runtime linker searches for libraries (from the manpage):

> If a shared object dependency does not contain a slash, then it is searched for in the following order:
>
> * (ELF only) Using the directories specified in the `DT_RPATH` dynamic section attribute of the binary if present and `DT_RUNPATH` attribute does not exist. Use of `DT_RPATH` is deprecated.
>
> * Using the environment variable `LD_LIBRARY_PATH` (unless the executable is being run in secure-execution mode; see below).  in which case it is ignored.
>
> * (ELF only) Using the directories specified in the `DT_RUNPATH` dynamic section attribute of the binary if present.

On some systems, including Debian Sid, setting `-Wl,-rpath` sets `DT_RUNPATH` instead of `DT_RPATH`, so `LD_LIBRARY_PATH` is searched first.

The only reliable fix is for `mesontest.py` to append the relevant test directories at the beginning of the `LD_LIBRARY_PATH` env variable. It is going to be a very rare use-case for someone to want to not use the libraries they just built when running the test program that they just built.

However, we can decide that this is too invasive a change right now, and just skip the test for this release.

nirbheek:
Continued from https://github.com/mesonbuild/meson/issues/2881:

> FreeBSD configures its clang to pass --enable-new-dtags to the linker by default, so RPATH gets converted to RUNPATH. If Debian is doing the same thing, this issue should also affect Debian users.

It does, and that's what the bug is about. We could disable it, but the way we rewrite rpaths on install means we will also set the install rpath with the old dtags, which breaks Debian and FreeBSD policy.

So the best solution would be to use the old dtags for build rpaths, and the new dtags (or whatever the default is for the linker) for install rpaths. Should be doable.

fooishbar:
> Now we have meson devenv (https://github.com/mesonbuild/meson/pull/8393), so users can manually workaround the issue by building and testing in the meson devenv environment. That is, users can run meson devenv ninja instead of ninja, and run meson devenv meson test instead of meson test.

[...]

PR #8393, opened 2021-02-21
Add `meson devenv` command and meson.add_devenv()

xclaesse:
New method `meson.add_devenv()` adds an [`environment()`](#environment) object
to the list of environments that will be applied when using `meson devenv`
command line. This is useful for developpers who wish to use the project without
installing it, it is often needed to set for example the path to plugins
directory, etc. Alternatively, a list or dictionary can be passed as first
argument.

``` meson
devenv = environment()
devenv.set('PLUGINS_PATH', meson.current_build_dir())
...
meson.add_devenv(devenv)
```

New command line has been added: `meson devenv -C builddir [<command>]`.
It runs a command, or open interactive shell if no command is provided, with
environment setup to run project from the build directory, without installation.

These variables are set in environment in addition to those set using `meson.add_devenv()`:
- `MESON_DEVENV` is defined to `'1'`.
- `PKG_CONFIG_PATH` includes the directory where Meson generates `-uninstalled.pc`
  files.
- `PATH` includes every directory where there is an executable that would be
  installed into `bindir`. On windows it also includes every directory where there
  is a DLL needed to run those executables.

### DIFF
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -1491,3 +1491,22 @@ class Backend:
             }]
 
         return []
+
+    def get_devenv(self) -> build.EnvironmentVariables:
+        env = build.EnvironmentVariables()
+        extra_paths = set()
+        for t in self.build.get_targets().values():
+            cross_built = not self.environment.machines.matches_build_machine(t.for_machine)
+            can_run = not cross_built or not self.environment.need_exe_wrapper()
+            in_bindir = t.should_install() and not t.get_install_dir(self.environment)[1]
+            if isinstance(t, build.Executable) and can_run and in_bindir:
+                # Add binaries that are going to be installed in bindir into PATH
+                # so they get used by default instead of searching on system when
+                # in developer environment.
+                extra_paths.add(os.path.join(self.environment.get_build_dir(), self.get_target_dir(t)))
+                if mesonlib.is_windows() or mesonlib.is_cygwin():
+                    # On windows we cannot rely on rpath to run executables from build
+                    # directory. We have to add in PATH the location of every DLL needed.
+                    extra_paths.update(self.determine_windows_extra_paths(t, []))
+        env.prepend('PATH', list(extra_paths))
+        return env
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -1223,10 +1223,10 @@ class Backend:
     def get_run_target_env(self, target: build.RunTarget) -> build.EnvironmentVariables:
         env = target.env if target.env else build.EnvironmentVariables()
         introspect_cmd = join_args(self.environment.get_build_command() + ['introspect'])
-        env.add_var(env.set, 'MESON_SOURCE_ROOT', [self.environment.get_source_dir()], {})
-        env.add_var(env.set, 'MESON_BUILD_ROOT', [self.environment.get_build_dir()], {})
-        env.add_var(env.set, 'MESON_SUBDIR', [target.subdir], {})
-        env.add_var(env.set, 'MESONINTROSPECT', [introspect_cmd], {})
+        env.set('MESON_SOURCE_ROOT', [self.environment.get_source_dir()])
+        env.set('MESON_BUILD_ROOT', [self.environment.get_build_dir()])
+        env.set('MESON_SUBDIR', [target.subdir])
+        env.set('MESONINTROSPECT', [introspect_cmd])
         return env
 
     def run_postconf_scripts(self) -> None:
--- a/mesonbuild/mdevenv.py
+++ b/mesonbuild/mdevenv.py
@@ -1,0 +1,79 @@
+import os, subprocess
+import argparse
+import tempfile
+
+from pathlib import Path
+from . import build
+from .mesonlib import MesonException, is_windows
+
+import typing as T
+
+def add_arguments(parser: argparse.ArgumentParser) -> None:
+    parser.add_argument('-C', default='.', dest='wd',
+                        help='directory to cd into before running')
+    parser.add_argument('command', nargs=argparse.REMAINDER,
+                        help='Command to run in developer environment (default: interactive shell)')
+
+def get_windows_shell() -> str:
+    mesonbuild = Path(__file__).parent
+    script = mesonbuild / 'scripts' / 'cmd_or_ps.ps1'
+    command = ['powershell.exe', '-noprofile', '-executionpolicy', 'bypass', '-file', str(script)]
+    result = subprocess.check_output(command)
+    return result.decode().strip()
+
+def get_env(b: build.Build, build_dir: str) -> T.Dict[str, str]:
+    env = os.environ.copy()
+    for i in b.devenv:
+        env = i.get_env(env)
+
+    extra_env = build.EnvironmentVariables()
+    extra_env.set('MESON_DEVENV', ['1'])
+    extra_env.set('MESON_PROJECT_NAME', [b.project_name])
+
+    meson_uninstalled = Path(build_dir) / 'meson-uninstalled'
+    if meson_uninstalled.is_dir():
+        extra_env.prepend('PKG_CONFIG_PATH', [str(meson_uninstalled)])
+
+    return extra_env.get_env(env)
+
+def run(options: argparse.Namespace) -> int:
+    options.wd = os.path.abspath(options.wd)
+    buildfile = Path(options.wd) / 'meson-private' / 'build.dat'
+    if not buildfile.is_file():
+        raise MesonException(f'Directory {options.wd!r} does not seem to be a Meson build directory.')
+    b = build.load(options.wd)
+
+    devenv = get_env(b, options.wd)
+
+    args = options.command
+    if not args:
+        prompt_prefix = f'[{b.project_name}]'
+        if is_windows():
+            shell = get_windows_shell()
+            if shell == 'powershell.exe':
+                args = ['powershell.exe']
+                args += ['-NoLogo', '-NoExit']
+                prompt = f'function global:prompt {{  "{prompt_prefix} PS " + $PWD + "> "}}'
+                args += ['-Command', prompt]
+            else:
+                args = [os.environ.get("COMSPEC", r"C:\WINDOWS\system32\cmd.exe")]
+                args += ['/k', f'prompt {prompt_prefix} $P$G']
+        else:
+            args = [os.environ.get("SHELL", os.path.realpath("/bin/sh"))]
+        if "bash" in args[0] and not os.environ.get("MESON_DISABLE_PS1_OVERRIDE"):
+            tmprc = tempfile.NamedTemporaryFile(mode='w')
+            bashrc = os.path.expanduser('~/.bashrc')
+            if os.path.exists(bashrc):
+                tmprc.write(f'. {bashrc}\n')
+            tmprc.write(f'export PS1="{prompt_prefix} $PS1"')
+            tmprc.flush()
+            # Let the GC remove the tmp file
+            args.append("--rcfile")
+            args.append(tmprc.name)
+
+    try:
+        return subprocess.call(args, close_fds=False,
+                               env=devenv,
+                               cwd=options.wd)
+    except subprocess.CalledProcessError as e:
+        return e.returncode
--- a/docs/markdown/Commands.md
+++ b/docs/markdown/Commands.md
@@ -263,3 +263,37 @@ An utility to manage WrapDB dependencies.
 {{ wrap_arguments.inc }}
 
 See [the WrapDB tool documentation](Using-wraptool.md) for more info.
+
+### devenv
+
+*(since 0.58.0)*
+
+{{ devenv_usage.inc }}
+
+Runs a command, or open interactive shell if no command is provided, with
+environment setup to run project from the build directory, without installation.
+
+We automatically handle `bash` and set `$PS1` accordingly. If the automatic `$PS1`
+override is not desired (maybe you have a fancy custom prompt), set the
+`$MESON_DISABLE_PS1_OVERRIDE` environment variable and use `$MESON_PROJECT_NAME`
+when setting the custom prompt, for example with a snippet like the following:
+
+```bash
+...
+if [[ -n "${MESON_PROJECT_NAME-}" ]];
+then
+  PS1+="[ ${MESON_PROJECT_NAME} ]"
+fi
+...
+```
+
+These variables are set in environment in addition to those set using `meson.add_devenv()`:
+- `MESON_DEVENV` is defined to `'1'`.
+- `MESON_PROJECT_NAME` is defined to the main project's name.
+- `PKG_CONFIG_PATH` includes the directory where Meson generates `-uninstalled.pc`
+  files.
+- `PATH` includes every directory where there is an executable that would be
+  installed into `bindir`. On windows it also includes every directory where there
+  is a DLL needed to run those executables.
+
+{{ devenv_arguments.inc }}
--- a/docs/markdown/Reference-manual.md
+++ b/docs/markdown/Reference-manual.md
@@ -2055,6 +2055,19 @@ the following methods.
 
 - `version()`: return a string with the version of Meson.
 
+- `add_devenv()`: *(Since 0.58.0)* add an [`environment()`](#environment) object
+  to the list of environments that will be applied when using [`meson devenv`](Commands.md#devenv)
+  command line. This is useful for developpers who wish to use the project without
+  installing it, it is often needed to set for example the path to plugins
+  directory, etc. Alternatively, a list or dictionary can be passed as first
+  argument.
+  ``` meson
+  devenv = environment()
+  devenv.set('PLUGINS_PATH', meson.current_build_dir())
+  ...
+  meson.add_devenv(devenv)
+  ```
+
 ### `build_machine` object
 
 Provides information about the build machine — the machine that is
--- a/docs/markdown/snippets/devenv.md
+++ b/docs/markdown/snippets/devenv.md
@@ -1,0 +1,29 @@
+## Developer environment
+
+New method `meson.add_devenv()` adds an [`environment()`](#environment) object
+to the list of environments that will be applied when using `meson devenv`
+command line. This is useful for developpers who wish to use the project without
+installing it, it is often needed to set for example the path to plugins
+directory, etc. Alternatively, a list or dictionary can be passed as first
+argument.
+
+``` meson
+devenv = environment()
+devenv.set('PLUGINS_PATH', meson.current_build_dir())
+...
+meson.add_devenv(devenv)
+```
+
+New command line has been added: `meson devenv -C builddir [<command>]`.
+It runs a command, or open interactive shell if no command is provided, with
+environment setup to run project from the build directory, without installation.
+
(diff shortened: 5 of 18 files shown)


mesonbuild/backend/backends.py
```diff
@@ -1484,3 +1484,17 @@ def get_introspection_data(self, target_id: str, target: build.Target) -> T.List
...
+        env = build.EnvironmentVariables()
+        if mesonlib.is_windows() or mesonlib.is_cygwin():
+            # On windows we cannot rely on rpath to run executables from build
+            # directory. We have to add in PATH the location of every DLL needed.
+            extra_paths = set()
+            for t in self.build.get_targets().values():
+                cross_built = not self.environment.machines.matches_build_machine(t.for_machine)
+                can_run = not cross_built or not self.environment.need_exe_wrapper()
+                if isinstance(t, build.Executable) and can_run:
+                    extra_paths.update(self.determine_windows_extra_paths(t, []))
```
tp-m, line 1498: Please forgive my ignorance about Meson implementation details, but what exactly counts as `build.Executable` in Meson terms?

I wonder if there should be a way to opt out of this automatism, e.g. I don't know if we would want every little test/utility/example added to `PATH`.

Does it include all `library()`? If so, it would again be good to have an opt-out mechanism for plugins (which I don't think need to be in `PATH`?).

Reason this matters is that on Windows there's a hard limit on the number of characters for e.g. `PATH` of 4096 or somesuch and it's very easy to break that in GStreamer, for example, if your build directory is not underneath `C:\` but say `C:\Users\Jane Doe\Sources` (the current script is far from ideal in that respect).

xclaesse: What that code does is look for every executable(), for each of them check every DLL it links from build directory and add those directories into PATH. It could be too much indeed... Probably should limit to installed exe.

Note that most of this logic is actually copied from gst-uninstalled, I'll double check what it does there for Windows path.

xclaesse: @tp-m I checked gst-env.py, and as far as I can tell, it does the same thing. I guess if we hit Windows limitation there is not much we can do...

Now modified that code to also add in PATH the location of the executable itself, so they can be run from command line without specifying the path in build dir. gst-env.py does that for GStreamer tools.

xclaesse: @tp-m FYI, I gave it a try in gst-build: https://gitlab.freedesktop.org/xclaesse/gst-build/-/commit/b829abdb1230dd96fb73b62989c8254bb01d4b76. With that I can run gst-inspect-1.0 and it list all plugins from my build directory, just like gst-env.py does.

There is still tones of hacks in gst-env.py that I did not understand yet, I'm sure we can copy what makes sense to improve Meson later.

tp-m: Yeah, just saying that there are problems with gst-env.py and things we could probably do better there.

It's cool to move this into Meson imho, makes a lot of sense to me.

xclaesse: Do you have suggestions to make it better? Are there alternatives to setting huge PATH on Windows?

tristan957:
Pretty funny. I just added a `shell` `run_target()` to my project which has `PATH` set to include the parent directories of all the `executable`s in my project. As well as `PYTHONPATH` for my Python bindings subprojects.

xclaesse:
> Pretty funny. I just added a `shell` `run_target()` to my project which has `PATH` set to include the parent directories of all the `executable`s in my project. As well as `PYTHONPATH` for my Python bindings subprojects.

Nice, let me know if this PR works in your case then. I could set PYTHONPATH too, does meson know where there are modules?

tristan957:
I will give it a test! One thing that I think will make this harder to pickup is that I think you talked about only including installed executables on the PATH, but I definitely need access to uninstalled executables as well at least at the moment. I am sure other people could also come in and want access to those executables as well because I and others wouldn't be building these executables if they weren't important to begin with. They may just not be important enough for installation environments.

Meson could have the ability to know about my Python modules with the pymod.exetension_module() function from the Python module, but Meson really can't know at what level the root of the Python package is for instance, and I am not sure that it should.

I think once you get into the realm of specialized environment variables like PYTHONPATH, you are just asking for trouble. I think setting PYTHONPATH correctly is best left as an exercise for the user. PKG_CONFIG_PATH is good though since that comes directly from the build system configuration itself. Is LD_LIBRARY_PATH worth being magically setup by Meson? I don't know. One thing that could enable setting things like PYTHONPATH correctly is having functions on subproject objects to retrieve the source root and the build root of said subprojects. Right now I am kind of hacking it in by creating a variable called `project_build_root = meson.project_build_root()` in my Python bindings project, and then calling subproject.get_variable('project_build_root') in order to properly setup the PYTHONPATH. Feels kind of hackish to me, and would rather switch to something more official.

I think this is one those features that is best left being as manual as possible otherwise you're going to be fielding feature requests for various environment variables to be setup by Meson for the rest of your life. Enable people with the functions they need to be successful, and everyone involved will have a better time.

mesonbuild/backend/backends.py
tp-m: My suggestion was to add a kwarg to `executable()` (or `subdir()`? but `executable()` makes more sense to me) to opt out of being added to the path automagically for the devenv. If this becomes part of Meson we can actually do this.

xclaesse:
> I will give it a test! One thing that I think will make this harder to pickup is that I think you talked about only including installed executables on the PATH, but I definitely need access to uninstalled executables as well at least at the moment. I am sure other people could also come in and want access to those executables as well because I and others wouldn't be building these executables if they weren't important to begin with. They may just not be important enough for installation environments.

On linux you can always run them as `./builddir/... ` but on Windows you could miss some DLL because it won't set PATH for them automatically. Note that you can also append to PATH manually:
```meson
e = environment()
e.prepend('PATH', meson.current_build_dir())
meson.add_devenv(e)
```

That said, thinking about it... we could also allow:
```meson
exe = executable(...)
meson.add_devenv(exe)
```

> Meson could have the ability to know about my Python modules with the pymod.exetension_module() function from the Python module, but Meson really can't know at what level the root of the Python package is for instance, and I am not sure that it should.

You're right, better leave PYTHONPATH handling to the user.

> Is LD_LIBRARY_PATH worth being magically setup by Meson? I don't know.

We don't need to set it because meson sets rpath on Linux/MacOS, no? On Windows this PR does it automatically using PATH.

However `gnome.generate_gir()` could set `GI_TYPELIB_PATH`, just saw that `gst-env.py` does that.

mesonbuild/backend/backends.py
xclaesse: `executable(..., devenv: false)` that default to the value of `install` kwarg? That would make sense indeed.

Another idea I had is `meson devenv` could first create `<builddir>/meson-sysroot` directory, copy all installed dll/exe there, and add just that directory into PATH. If we only copy when mtime changed (like `meson install` does) it should be fast enough. That would solve the problem completely, what do you think?

xclaesse: Could do both actually, as @tristan957 mentioned below we could want to include some not installed executables into PATH too. In that case `executable(..., devenv: true)` could be handy.

mesonbuild/mdevenv.py
```diff
@@ -0,0 +1,78 @@
...
+                args += ['-NoLogo', '-NoExit']
+                prompt = 'function global:prompt {{  "{} PS " + $PWD + "> "}}'.format(prompt_prefix)
+                args += ['-Command', prompt]
+            else:
+                args = [os.environ.get("COMSPEC", r"C:\WINDOWS\system32\cmd.exe")]
+                args += ['/k', 'prompt {} $P$G'.format(prompt_prefix)]
+        else:
+            args = [os.environ.get("SHELL", os.path.realpath("/bin/sh"))]
+        if "bash" in args[0] and not os.environ.get("MESON_DISABLE_PS1_OVERRIDE"):
+            tmprc = tempfile.NamedTemporaryFile(mode='w')
```
dcbaker, line 62: It looks like we're going to leak this, right?

mesonbuild/mdevenv.py
```diff
@@ -0,0 +1,78 @@
+import os, subprocess
+import argparse
+import tempfile
+import shutil
+
+from pathlib import Path
+from . import build
+from .mesonlib import MesonException, is_windows
+
+def add_arguments(parser):
```
dcbaker, line 10:
```suggestion
def add_arguments(parser: argparse.ArgumentParser) -> None:
```

mesonbuild/mdevenv.py
```diff
@@ -0,0 +1,78 @@
...
+    parser.add_argument('-C', default='.', dest='wd',
+                        help='directory to cd into before running')
+    parser.add_argument('command', nargs=argparse.REMAINDER,
+                        help='Command to run in developer environment (default: interactive shell)')
+
+def get_windows_shell():
```
dcbaker, line 16:
```suggestion
def get_windows_shell() -> str:
```

mesonbuild/mdevenv.py
```diff
@@ -0,0 +1,78 @@
...
+    mesonbuild = Path(__file__).parent
+    script = mesonbuild / 'scripts' / 'cmd_or_ps.ps1'
+    command = ['powershell.exe', '-noprofile', '-executionpolicy', 'bypass', '-file', str(script)]
+    result = subprocess.check_output(command)
+    return result.decode().strip()
+
+def get_env(b: build.Build, build_dir: str):
```
dcbaker, line 23: what does this return?

mesonbuild/mdevenv.py
```diff
@@ -0,0 +1,78 @@
...
+    extra_env = build.EnvironmentVariables()
+    extra_env.add_var(extra_env.set, 'MESON_DEVENV', ['1'], {})
+
+    meson_uninstalled = Path(build_dir) / 'meson-uninstalled'
+    if meson_uninstalled.is_dir():
+        extra_env.add_var(extra_env.prepend, 'PKG_CONFIG_PATH', [str(meson_uninstalled)], {})
+
+    return extra_env.get_env(env)
+
+def run(options):
```
dcbaker, line 37:
```suggestion
def run(options: argparse.Namespace) -> int:
```

dcbaker (changes requested):
With the rest of the type annotations added, can we add this to the `run_mypy` script? It would be nice to not add new modules without type annotations if possible.

I don't really have a problem with this. This may be useful for scipy as well, as we can probably expand this to emulate setup.py's editable installs.

mesonbuild/mdevenv.py
xclaesse: No, they are deleted automatically by GC: `If delete is true (the default), the file is deleted as soon as it is closed`.

mesonbuild/mdevenv.py
xclaesse: `T.Dict[str, str]`

dcbaker (approved):
For my part, LGTM.

jpakkane:
What happens if the master and subprojects do different things? Such as one setting `foo` to `bar` wheas the other one sets it to `baz`?

tristan957:
My project has integrations with other projects. One of them is Mongo. If I go to my project and run `meson devenv`. Would be nice if Mongo could pick up a non-installed `libhse` on `LD_LIBRARY_PATH`. It is easy enough for me to setup manually that if it doesn't get implemented, it doesn't matter to me. I think setting it is a good way to avoid having to install my project on the system.

xclaesse:
> What happens if the master and subprojects do different things? Such as one setting `foo` to `bar` wheas the other one sets it to `baz`?

If they use `environment.set()` they override each other so last one wins. If they use `environment.append/prepend()` they get applied in the order in which they have been added using `meson.add_env()`.

The expectation is that it's either a variable reserved for your project, like we have `MESON_*` and it won't clash, or it's stuff like PATH where it should be additive.

xclaesse:
> For my part, LGTM.

@dcbaker I sneaked a couple other commits, because running mypy forced me to get EnvironmentVariables annotated.

mesonbuild/mdevenv.py
```diff
@@ -0,0 +1,80 @@
...
+                args = ['powershell.exe']
+                args += ['-NoLogo', '-NoExit']
+                prompt = 'function global:prompt {{  "{} PS " + $PWD + "> "}}'.format(prompt_prefix)
+                args += ['-Command', prompt]
+            else:
+                args = [os.environ.get("COMSPEC", r"C:\WINDOWS\system32\cmd.exe")]
+                args += ['/k', 'prompt {} $P$G'.format(prompt_prefix)]
+        else:
+            args = [os.environ.get("SHELL", os.path.realpath("/bin/sh"))]
+        if "bash" in args[0] and not os.environ.get("MESON_DISABLE_PS1_OVERRIDE"):
```
tristan957, line 63: I feel like this specialty stuff for UNIX-land is pointless and asking for pain. I can very easily change my prompt in my bashrc to do something special when `MESON_DEVENV` is defined. I already do something similar to Python virtual environments. As long as it is documented, developers can customize on their own. Plus it is going to become really ugly in this file when everyone in the world wants to add their random shell logic here.

tristan957:
I would like to see named `devenv`s.

```meson
meson.add_devenv('default', devenv)
```

I would also like to see the ability to change workdir.

xclaesse:
> My project has integrations with other projects. One of them is Mongo. If I go to my project and run `meson devenv`. Would be nice if Mongo could pick up a non-installed `libhse` on `LD_LIBRARY_PATH`. It is easy enough for me to setup manually that if it doesn't get implemented, it doesn't matter to me. I think setting it is a good way to avoid having to install my project on the system.

So in your case Mongo is installed, not a subproject, and you want to run it with uninstalled libhse? Not sure this kind of stuff should be in scope.

tristan957:
I would be operating Mongo out of its build tree in this case. I am just giving an example for LD_LIBRARY_PATH being auto-set, but like I said, I can just set it myself.

xclaesse:
> I would like to see named `devenv`s.
> 
> ```meson
> meson.add_devenv('default', devenv)
> ```
> 
> I would also like to see the ability to change workdir.

To be able to do `meson devenv --name default` ? I guess that could be added later as a new kwarg, would like to land initial version first.

tristan957:
Mongo would not be a subproject because Mongo hates the world and chooses to be a scons project.

mesonbuild/mdevenv.py
xclaesse: It's harder to do a different PS1 for each project in which you could be using devenv. IMHO it's useful to have that information that you're currently inside a devenv.

eli-schwartz: The current copying of bashrc is completely painful, just create a new bashrc that does:

```
. ~/.bashrc
PS1="..."
```

I'm not terribly upset with the idea of doing this at all, though... it's basically like `venv/bin/activate` in python land.

mesonbuild/backend/backends.py
tristan957: Now that I have thought about this more. I almost think that anything magical happening in the PR is just going to be difficult to maintain and that something like this could be much better:

```
meson.devenv(devenv, exectuables: [...])
```

mesonbuild/mdevenv.py
tristan957: Then this PR should support a way to disable the default prompt behavior like Python virtual environments do.

```
[0 0 15:34:58] [tpartin@campise] [hse] [updates/NFSE-4592-followup3 $%=] [venv]
● $
```

This is my prompt and here `venv` is the name of my Python virtual environment which is very different than the default behavior.

tristan957: `export VIRTUAL_ENV_DISABLE_PROMPT=1` is what I do for Python.

xclaesse: @eli-schwartz tbh this is a blind copy paste from gst-env.py, but I think you're right.

@tristan957 you can disable it with MESON_DISABLE_PS1_OVERRIDE, but that trick is missing from doc I think.

tristan957: oh sweet. I missed that in my ocde readthrough

xclaesse: @eli-schwartz changed the rcfile as you suggested.
@tristan957 added  in the doc a mention about MESON_DISABLE_PS1_OVERRIDE.

mesonbuild/backend/backends.py
xclaesse: @tp-m let's start with including everything by default in the initial PR, we can extend later by adding a way to opt-out some exe, or workaround by copying into a sysroot dir, or any other fancy solution. OK?

tp-m: Sure :+1:

xclaesse:
Actually just got a case where setting LD_LIBRARY_PATH on Linux would be useful: I was chasing a bug in GTK for which I had to run gedit (from system) using my build of GTK. Would have been nice to be able to run gedit inside `meson devenv`.

> For my part, LGTM.

Let's merge this then, we can always extend it in follow up PRs.

tristan957:
Why can't you run gedit inside the devenv

xclaesse:
> Why can't you run gedit inside the devenv

You can, but currently that will use system gtk because LD_LIBRARY_PATH is not set. gtk project coud of course set LD_LIBRARY_PATH themself in their meson.build, but Meson could be smart and do it for us. That's for a future PR.

tristan957:
Ok. I just misunderstood your previous question. I'll try to give the feature a whirl later this week

lantw44:
> > Is LD_LIBRARY_PATH worth being magically setup by Meson? I don't know.
> 
> We don't need to set it because meson sets rpath on Linux/MacOS, no? On Windows this PR does it automatically using PATH.

No, RPATH doesn't work if it can be overriden by `LD_LIBRARY_PATH`: https://github.com/mesonbuild/meson/issues/1635. As far as I know, autotools (libtool) work because it uses a wrapper script to set environment variables, but meson and cmake don't.

xclaesse:
@lantw44 not sure what you mean.

I opened https://github.com/mesonbuild/meson/pull/8548 for setting GI_TYPELIB_PATH and LD_LIBRARY_PATH.

lb90:
Hi @xclaesse, thanks for this, looks great!

May I ask a question?

During the `meson compile` many projects have to first compile and then run some tools (generators). Those tools are placed in the build directory (i.e. they are uninstalled). Does the devenv apply in this case during `meson compile`?

xclaesse:
@lb90 no, that env is really only used for the `meson devenv` command atm. We could easily add `env` kwarg to generator() just like we already have it for `run_target()` and `custom_target()`.

Wondering if we could have a special value for `env:` kwarg to mean "pass the devenv"... maybe... do you have a specific use-case in mind?

tristan957:
@xclaesse I do. We have internal tools we use for testing that wouldn't be distributed with the rest of our software. It would be nice if there was a way to add executables into the devenv either through a kwarg or a method on the devenv object itself.

xclaesse:
@tristan957 You mean running those executable from a `custom_target()` or `run_target()` or manually from within `meson devenv` shell? Not that you can always add executables into devenv PATH manually:
```
env = environment()
env.append('PATH', fd.dirname(exe.full_path()))
meson.add_devenv(env)
```

tristan957:
Manual is good enough for me then.

PR #8398, opened 2021-02-22
install_man locale support

jaos:
Rather than having to manually build the locale aware man paths with

`install_data('foo.fr.1', install_dir: join_paths(get_option('mandir'), 'fr', 'man1'), rename: 'foo.1')`

Support doing

`install_man('foo.fr.1', locale: 'fr')`

### DIFF
--- a/docs/markdown/Reference-manual.md
+++ b/docs/markdown/Reference-manual.md
@@ -1130,6 +1130,11 @@ Accepts the following keywords:
   format and optionally the owner/uid and group/gid for the installed files.
   An example value could be `['rwxr-sr-x', 'root', 'root']`.
 
+- `locale` *(since 0.58.0)*: can be used to specify the locale into which the
+  man page will be installed within the manual page directory tree.
+  An example manual might be `foo.fr.1` with a locale of `fr`, such
+  that `{mandir}/{locale}/man{num}/foo.1` becomes the installed file.
+
 *(since 0.49.0)* [manpages are no longer compressed
  implicitly][install_man_49].
 
--- a/docs/markdown/snippets/install_man_locale.md
+++ b/docs/markdown/snippets/install_man_locale.md
@@ -1,0 +1,9 @@
+## Specify man page locale during installation
+
+Locale directories can now be passed to `install_man`:
+    
+```meson
+# instead of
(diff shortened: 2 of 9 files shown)


jpakkane:
Is the file name always of the type `foo.fr.number`? That is, could we just detect that and always do the right thing without the user needing to add kwargs by hand?

jaos:
> Is the file name always of the type `foo.fr.number`? That is, could we just detect that and always do the right thing without the user needing to add kwargs by hand?

I'm not sure. As far as I know there is not an obvious, universal convention. I've had translations given to me with locale name in the file but I have seen other translation projects that keep them in a source directory structure with the locale represented that way. 

Auto detection seems haphazard but we might be able to check against a list. The locale py module does not seem to have an exhaustive list via locale_alias, however.

dcbaker:
Could we do something like "if you don't set locale and your file ends in `.[str][str].section` we'll guess for you", but an explicit `locale` always overrides?

dcbaker:
I'm going to be doing some work here, so I could implement the guess mechanism myself if you think it's reasonable, and we could just merge this as-is?

jaos:
Auto detecting locals might work if you are able to account for all possible formats. 

https://www.gnu.org/software/gettext/manual/html_node/Locale-Names.html

It would be nice if there was something to leverage here so we don't get it wrong. Otherwise requiring it be specified seems the most deterministic approach.

jaos:
> I'm going to be doing some work here, so I could implement the guess mechanism myself if you think it's reasonable, and we could just merge this as-is?

Feel free to incorporate this or a similar approach as you see fit. I really only meant to get the conversation started, not necessarily propose a specific solution.

dcbaker:
Hmm, we could do something naïve, but I'm okay with starting with an explicit mechanism and adding an implicit one later.

dcbaker (approved):
The only thing that this needs is a replease note in docs/markdown/snippets

dcbaker (changes requested):
Oops, clicked the wrong box, sorry.

jaos:
> The only thing that this needs is a replease note in docs/markdown/snippets

Ok, added docs/markdown/snippets/install_man_locale.md.

dcbaker (approved):

dcbaker:
thanks looks good to me.

PR #8401, opened 2021-02-23
mcompile: make sure arguments are passed in the correct order

eli-schwartz:
meson compile itself doesn't permit GNU-style argument permutation, i.e. TARGET to precede options, so why should we expect ninja to?

And indeed, ninja doesn't document support for this -- but it does accept it anyway, which is confusing and results in people thinking it's "supposed to" work.
    
However, if `NINJA=samu` then this is in fact enforced. samu does not permit GNU-style argument permutation. As a result, the arguments passed to mcompile are actively re-ordered before being passed to the subprocess, and samu dies with a fatal error.
    
Fix ordering in mcompile.py and add a comment to warn future readers that the order does, in fact, matter.

### DIFF
--- a/mesonbuild/mcompile.py
+++ b/mesonbuild/mcompile.py
@@ -144,13 +144,6 @@ def get_parsed_args_ninja(options: 'argparse.Namespace', builddir: Path) -> T.Tu
 
     cmd = runner + ['-C', builddir.as_posix()]
 
-    if options.targets:
-        intro_data = parse_introspect_data(builddir)
-        for t in options.targets:
-            cmd.extend(generate_target_names_ninja(ParsedTargetName(t), builddir, intro_data))
-    if options.clean:
-        cmd.append('clean')
-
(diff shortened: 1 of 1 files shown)


eli-schwartz:
```
Download action repository 'actions/checkout@v2'
Warning: Failed to download action 'https://api.github.com/repos/actions/checkout/tarball/5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f'. Error: The operation was canceled.
Warning: Back off 15.265 seconds before retry.
Warning: Failed to download action 'https://api.github.com/repos/actions/checkout/tarball/5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f'. Error: The operation was canceled.
Warning: Back off 19.459 seconds before retry.
Error: The operation was canceled.
```

CI just looooves to find new ways to hilariously break.

dcbaker (approved):

PR #8402, opened 2021-02-23
Add compiler arguments to linker arguments when the compiler acts as a linker driver

dcbaker:
The original approach we used was completely and totally wrong, didn't work, shouldn't have worked, and was just bad. I've deleted that, done a bit of refactoring, and put the logic in the Compiler class, which makes more sense. 

Fixes: #8345

### DIFF
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -658,12 +658,6 @@ class CoreData:
             raise type(e)(('Validation failed for option %s: ' % option_name) + str(e)) \
                 .with_traceback(sys.exc_info()[2])
 
-    def get_external_args(self, for_machine: MachineChoice, lang: str) -> T.Union[str, T.List[str]]:
-        return self.options[OptionKey('args', machine=for_machine, lang=lang)].value
-
-    def get_external_link_args(self, for_machine: MachineChoice, lang: str) -> T.Union[str, T.List[str]]:
-        return self.options[OptionKey('link_args', machine=for_machine, lang=lang)].value
-
     def update_project_options(self, options: 'KeyedOptionDictType') -> None:
         for key, value in options.items():
             if not key.is_project():
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -776,7 +776,7 @@ class CoreData:
                       for_machine: MachineChoice, env: 'Environment') -> None:
         """Add global language arguments that are needed before compiler/linker detection."""
         from .compilers import compilers
-        options = compilers.get_global_options(lang, comp, for_machine, env)
+        options = compilers.get_global_options(lang, for_machine, env)
         self.add_compiler_options(options, lang, for_machine, env)
 
     def process_new_compiler(self, lang: str, comp: 'Compiler', env: 'Environment') -> None:
--- a/run_unittests.py
+++ b/run_unittests.py
@@ -5627,6 +5627,22 @@ class AllPlatformTests(BasePlatformTests):
         self.assertEqual(0, output.count('File reformatted:'))
         self.build('clang-format-check')
 
+    def test_compiler_args_added_to_linker_args(self) -> None:
+        # If the linker is invoked by the compiler then we need to add compiler
+        # arguments to the linker arguments, otherwise we should not. This is
+        # all meant to be transparent to the user, they should never see that
+        # this is happening from the Meosn DSL
+        env = get_fake_env()
+        comp = env.detect_compiler_for('c', MachineChoice.HOST)
+        env.coredata.options[OptionKey('args', machine=MachineChoice.HOST, lang='c')].set_value(['compile'])
+        env.coredata.options[OptionKey('link_args', machine=MachineChoice.HOST, lang='c')].set_value(['link'])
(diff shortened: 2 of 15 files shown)


jpakkane:
```
test cases/common/172 initial c_args/meson.build:6:0: ERROR: Assert failed: Incorrect value for c_link_args option.
```

This one is clearly wrong. Running it by hand shows that this causes `c_args` to be put in c_link_args as well.

```
Message: ['-Dtest_harmless_but_useless_link_arg', '-funroll-loops']
```

dcbaker:
Wait, I've always been under the impression `-Dc_args` and `CFLAGS` do exactly the same thing, you're telling me that `-Dc_args` shouldn't be applied to the linker as well?

jpakkane:
That is correct. In Meson compile args are compile only and link args are link only. We just have to do some song&dance to make CPPFLAGS and CFLAGS do the thing distro packagers expect it to.

dcbaker:
Except that we also have code specifically to make GCC use compiler args for when it's used as the linker. I don't think it's actually possible to implement the way you describe things right now in meson, ie, we would need two sets of compiler args, one that can be applied only to the compiler and one that is applied to both, and we've never had that.

jpakkane:
The original design was very explicitly how I described it above. I remember that I had to battle with people who wanted to make c_args apply also to linking (possibly also to add something like CPPFLAGS) but eventually people agreed that the strict split between compile and link was the right thing to do. Apparently this has changed silently in some refactoring. Which is bad, because that is not how it was ever supposed to work and there is probably code around that depends on that not being the case and which will break when least expected.

The reason I chose to do it this way was that it was _maddening_ on other build systems to find out which flags were used where. In contrast Meson was straightforward: c_args -> compile only, c_link_args -> link only.

dcbaker:
My concern here is that CFLAGS and Dc_args behaving differently is also maddening, and you generally need all of the cargs for compilers acting as link drivers. I don't really care which behavior we go for here, but I'm strongly of the opinion that the two args should behave the same. At this point though whichever we pick it's going to break someone's project

dcbaker:
Since we have LDFLAGS is he in favor of not mixing them, but I'm sure someone will be shipped l annoyed either way

jpakkane:
Unrelated to this issue, the test failure is quite clear. It passes -Dc_args=-FOO and -Dc_link_args=-BAR but when queried c_link_args is `['-FOO', '-BAR']`. That is just plain wrong. Whether or not c_args is used during linking, they _must not_ jump to `c_link_args`.

mesonbuild/coredata.py
```diff
@@ -776,8 +776,7 @@ def add_lang_args(self, lang: str, comp: T.Type['Compiler'],
                       for_machine: MachineChoice, env: 'Environment') -> None:
         """Add global language arguments that are needed before compiler/linker detection."""
         from .compilers import compilers
-        options = compilers.get_global_options(lang, comp, for_machine, env)
-        self.add_compiler_options(options, lang, for_machine, env)
+        self.options.update(compilers.get_global_options(lang, comp, for_machine, env))
```
xclaesse, line 779: Are you sure it should use update() here and not setdefault()? Is add_lang_args() is called only once per language so we are sure those options does not exist yet?

dcbaker: This is when the language options are created for the first time, and it's called once per language, per machine.

dcbaker:
I have a completely different approach for this, I'm just running tests locally to make sure everything works before I re-push it.

dcbaker:
@jpakkane, @xclaesse, waht about this approach?

run_unittests.py
```diff
@@ -5627,6 +5627,22 @@ def test_clang_format(self):
         self.assertEqual(0, output.count('File reformatted:'))
         self.build('clang-format-check')
 
+    def test_compiler_args_added_to_linker_args(self) -> None:
+        # If the linker is invoked by the compiler then we need to add compiler
+        # arguments to the linker arguments, otherwise we should not. This is
```
jpakkane, line 5632: I still do not agree with this and it is not how things were originally implemented and documented which is:

- c_args is used for compilation only
- c_link_args is used for linking only
- when parsing system envvars, CPPFLAGS is put only to c_args, but CFLAGS is put to both c_args and c_link_args

This is compatible with the "classic" disto envvar spamming approach but still keeps a strict separation between the two internally. It should also be noted that there are a many compile args that make no sense to add to the linker command, like -DFOO or -O2.

dcbaker: Sure, but for gcc if i have, for example: `-DFOO -flto` in my c_args, passing `-DFOO` to gcc as a linker argument doesn't hurt anything, and not passing `-flto` breaks everything.

And having `$CFLAGS` and `-Dc_args=` have different behavior (which I cannot find documented anywhere) is bad UI, it's the sort of subtlety that an end user is likely to be unaware of until things stop working, and they can't figure out why.

dcbaker:
Okay, I'll go back to the drawing board.

run_unittests.py
jpakkane: It should be noted that `CPPFLAGS` should only be used during compilation. If we were to always pass `c_args`, then this would be impossible to support.

PR #8415, opened 2021-02-25
meson: add .has_external_property() method

tp-m:
Useful in case of boolean values to distinguish between a boolean
value having been set in the native/cross file and not having been
provided, which can't be achieved by passing a fallback parameter
to .get_external_property().

### DIFF
--- a/docs/markdown/snippets/has_external_property.md
+++ b/docs/markdown/snippets/has_external_property.md
@@ -1,0 +1,7 @@
+## Check if native or cross-file properties exist
+
+It is now possible to check whether a native property or a cross-file property
+exists with `meson.has_external_property('foo')`. This is useful if the
+property in question is a boolean and one wants to distinguish between
+"set" and "not provided" which can't be done the usual way by passing a
+fallback parameter to `meson.get_external_property()` in this particular case.
--- a/mesonbuild/interpreter.py
+++ b/mesonbuild/interpreter.py
@@ -32,6 +32,7 @@ from .interpreterbase import InterpreterObject, MutableInterpreterObject, Disabl
 from .interpreterbase import FeatureNew, FeatureDeprecated, FeatureNewKwargs, FeatureDeprecatedKwargs
 from .interpreterbase import ObjectHolder, MesonVersionString
 from .interpreterbase import TYPE_var, TYPE_nkwargs
+from .interpreterbase import typed_pos_args
 from .modules import ModuleReturnValue, ExtensionModule
 from .cmake import CMakeInterpreter
 from .backend.backends import TestProtocol, Backend, ExecutableSerialisation
@@ -1947,6 +1948,7 @@ class MesonMain(InterpreterObject):
                              'project_name': self.project_name_method,
                              'get_cross_property': self.get_cross_property_method,
                              'get_external_property': self.get_external_property_method,
+                             'has_external_property': self.has_external_property_method,
                              'backend': self.backend_method,
                              })
 
@@ -2284,6 +2286,14 @@ class MesonMain(InterpreterObject):
                 return self.get_cross_property_method(args, kwargs)
             else:
                 return _get_native()
+
+    @permittedKwargs({'native'})
+    @FeatureNew('meson.has_external_property', '0.58.0')
+    @typed_pos_args('meson.has_external_property', str)
+    def has_external_property_method(self, args: T.Tuple[str], kwargs: T.Dict[str, T.Any]) -> str:
+        prop_name = args[0]
+        for_machine = self.interpreter.machine_from_native_kwarg(kwargs)
+        return prop_name in self.interpreter.environment.properties[for_machine]
 
 known_library_kwargs = (
     build.known_shlib_kwargs |
--- a/test cases/common/223 native prop/meson.build
+++ b/test cases/common/223 native prop/meson.build
@@ -23,3 +23,27 @@ assert(x=='fallback', 'fallback native:false did not work')
 
 x = meson.get_external_property('anarray')
 assert(x==['one', 'two'], 'array did not work')
+
+assert(meson.has_external_property('anarray'), 'expected property "anarray" to exist')
+assert(meson.has_external_property('astring'), 'expected property "astring" to exist')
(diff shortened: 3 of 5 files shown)


mesonbuild/interpreter.py
```diff
@@ -2253,6 +2255,21 @@ def get_cross_property_method(self, args, kwargs) -> str:
                 return args[1]
             raise InterpreterException('Unknown cross property: %s.' % propname)
 
+    @noArgsFlattening
```
xclaesse, line 2258: Not needed since there is only one arg anyway. We now also have @typed_pos_args() to validate the number and  type of args. It's not used yet in interpreter but should be used for new API.

mesonbuild/interpreter.py
```diff
@@ -2253,6 +2255,21 @@ def get_cross_property_method(self, args, kwargs) -> str:
...
+        if len(args) != 1:
+            raise InterpreterException('Must have exactly one argument.')
+        propname = args[0]
+        if not isinstance(propname, str):
+            raise InterpreterException('Property name must be string.')
+        try:
+            props = self.interpreter.environment.properties.host
+            return propname in props
+        except Exception:
+            return False
```
xclaesse, line 2271: Why try? What specific exception could it raise?

mesonbuild/interpreter.py
```diff
@@ -2282,6 +2299,33 @@ def _get_native() -> str:
...
+        propname = args[0]
+        if not isinstance(propname, str):
+            raise InterpreterException('Property name must be string.')
+
+        def _get_native() -> str:
+            try:
+                props = self.interpreter.environment.properties.build
+                return propname in props
+            except Exception:
+                return False
```
xclaesse, line 2317: Same here, why try?

test cases/common/223 native prop/meson.build
```diff
@@ -22,4 +22,14 @@ assert(x=='fallback', 'fallback native:false did not work')
...
+
+assert(meson.has_external_property('anarray'))
+assert(meson.has_external_property('astring'))
+assert(not meson.has_external_property('abool'))
+
+if meson.is_cross_build()
+  assert(meson.has_cross_property('anarray'))
+  assert(meson.has_cross_property('astring'))
+  assert(not meson.has_cross_property('abool'))
+endif
```
xclaesse, line 35: Could add else case where you check they all return false.

dcbaker:
I'm fine with `has_external_property`, but I thought that `get_cross_property` was deprecated since you can do `get_external_property(native : bool)`?

mesonbuild/interpreter.py
```diff
@@ -2282,6 +2299,33 @@ def _get_native() -> str:
...
+        if 'native' in kwargs:
+            if kwargs['native']:
+                return _get_native()
+            else:
+                return self.has_cross_property_method(args, {})
+        else:  # native: not specified
+            if self.build.environment.is_cross_build():
+                return self.has_cross_property_method(args, kwargs)
+            else:
+                return _get_native()
```
dcbaker, line 2327: I think you could simplify all of this to:
```python
machine = self.interpreter.machine_from_native_kwarg(kwargs)
return prop in self.interpreter.properties[machine]
```

xclaesse:
> I'm fine with `has_external_property`, but I thought that `get_cross_property` was deprecated since you can do `get_external_property(native : bool)`?

It's not deprecated, but I agree it should.

dcbaker:
@xclaesse: https://github.com/mesonbuild/meson/pull/8419

mesonbuild/interpreter.py
tp-m: Because I blindly cargo-culted/copied it without thinking :)

Fixed now by using Dylan's massively simplified version.

test cases/common/223 native prop/meson.build
tp-m: Done.

mesonbuild/interpreter.py
tp-m: Hrm yes, I thought it was all a bit convoluted :laughing:

Thanks, fixed.

mesonbuild/interpreter.py
tp-m: Added `@types_pos_args` and removed manual type/num checking, thanks.

tp-m:
Removed `has_cross_property()` variant now that `get_cross_property()` has been deprecated.

mesonbuild/interpreter.py
```diff
@@ -2284,6 +2286,14 @@ def _get_native() -> str:
             else:
                 return _get_native()
 
+    @permittedKwargs({'native'})
+    @FeatureNew('meson.has_external_property', '0.58.0')
+    @typed_pos_args('meson.has_external_property', str)
+    def has_external_property_method(self, args: T.Sequence[str], kwargs: dict) -> str:
```
dcbaker, line 2292:
```suggestion
    def has_external_property_method(self, args: T.Tuple[str], kwargs: T.Dict[str, T.Any]) -> str:
```

dcbaker (approved):
The type signature is slightly off, but otherwise this all looks good to me at this point.

mesonbuild/interpreter.py
tp-m: Updated as suggested, thanks.

mesonbuild/interpreter.py
tp-m: See above, it's gone now.

docs/markdown/snippets/has_external_property.md
```diff
@@ -0,0 +1,7 @@
+## Check if native or cross-file properties exist
+
+It is now possible to check whether a native property or a cross-file property
+exists with `meson.has_external_property('foo')`. This is useful if the
+property in question is a boolean and one wants to distinguish between
+"set" and "not provided" which can't be done the usual way by passing a
+fallback parameter to `meson.get_external_property()` in this particular case.
```
xclaesse, line 7: That's the release snippet, but missing the reference manual doc.

tp-m: True. Added now, thanks!

xclaesse:
LGTM, thanks.

Issue #8405, opened 2021-02-24
meson test --repeat error


schutzekatze:
**Describe the bug**
Using `meson test --repeat=X` consistently raises an exception in meson:
```
Exception in callback TestHarness._run_tests.<locals>.test_done(<Task finishe...fa86d3f5100>)>) at /usr/lib/python3.9/site-packages/mesonbuild/mtest.py:1807
handle: <Handle TestHarness._run_tests.<locals>.test_done(<Task finishe...fa86d3f5100>)>) at /usr/lib/python3.9/site-packages/mesonbuild/mtest.py:1807>
Traceback (most recent call last):
  File "/usr/lib/python3.9/asyncio/events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
  File "/usr/lib/python3.9/site-packages/mesonbuild/mtest.py", line 1809, in test_done
    f.result()
  File "/usr/lib/python3.9/site-packages/mesonbuild/mtest.py", line 1805, in run_test
    self.process_test_result(res)
  File "/usr/lib/python3.9/site-packages/mesonbuild/mtest.py", line 1558, in process_test_result
    l.log(self, result)
  File "/usr/lib/python3.9/site-packages/mesonbuild/mtest.py", line 657, in log
    self.running_tests.remove(result)
  File "/usr/lib/python3.9/_collections_abc.py", line 600, in remove
    raise KeyError(value)
KeyError: <mesonbuild.mtest.TestRunExitCode object at 0x7fa86d3f5100>
```

**To Reproduce**
A `meson.build` as simple as:
```
project('simple', 'cpp')
test('test1', executable('test1', 'test.cpp'))
```

And `test.cpp`:

```
#include <iostream>

int main() {
  std::cerr << "this is a test" << std::endl;
  return 0;
}
```
Reproduces this behavior if you do: `meson build`, `cd build`, `meson test --repeat=10`.

**system parameters**
OS: Arch Linux
Python: 3.9.1
Meson: 0.57.1
Ninja: 1.10.2

This doesn't seem to be happening on Arch Linux alone, as I was getting the same error on my CI builds which use Ubuntu and MacOS.

schutzekatze:
Also, this issue is somewhat new. I've been using `meson test --repeat=X` successfully before. A month ago this wasn't happening.

eli-schwartz:
Bisected to commit 72e96a470a13928526a292cc8066e230df612d19, part of #8029

@bonzini

[...]

PR #8416, opened 2021-02-25
mtest: create separate runners for multiple repeats

bonzini:
Reusing the runners for multiple repeats of the test run gets in the way of the progress report, which stores runners in an OrderedSet. Instead, create a separate SingleTestRunner object for each repeat.

While at it, fix the "duplicate suite" assertion as it can fire with TAP tests and --repeat=N.

Fixes: #8405


jpakkane:
This should have a test.

Issue #8414, opened 2021-02-25
error() should be able to take multiple arguments just like message() and warning()


tp-m:
Not sure if this is on purpose or an oversight, but I think `error()` should also be able to take multiple arguments, so I can write e.g.
```
  error('Unexpected value:', somevariable)
```
without having to go through `.format()`.

PR #8417, opened 2021-02-25
Support multiple args in error()

xclaesse:
Seems it got forgotten when that was added to warnings() and message().

Fixes: #8414.


dcbaker (approved):
It would be nice if there was a test for this, probably just expand the existing error test?

xclaesse:
> It would be nice if there was a test for this, probably just expand the existing error test?

Right, done.

PR #8418, opened 2021-02-25
Use `typed_pos_args` in fs module

dcbaker:
This is just a single commit from my larger series to use `typed_pos_args` in all of meson, plus a couple of easy optimizations I saw.

### DIFF
--- a/mesonbuild/modules/fs.py
+++ b/mesonbuild/modules/fs.py
@@ -26,9 +26,8 @@ from ..mesonlib import (
     MesonException,
     path_is_in_root,
 )
-from ..interpreterbase import FeatureNew
+from ..interpreterbase import FeatureNew, typed_pos_args, noKwargs, permittedKwargs
 
-from ..interpreterbase import stringArgs, noKwargs
 if T.TYPE_CHECKING:
     from ..interpreter import Interpreter, ModuleState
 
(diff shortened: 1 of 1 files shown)


mesonbuild/modules/fs.py
```diff
@@ -60,69 +59,57 @@ def _resolve_dir(self, state: 'ModuleState', arg: str) -> Path:
...
-    @stringArgs
     @noKwargs
     @FeatureNew('fs.expanduser', '0.54.0')
-    def expanduser(self, state: 'ModuleState', args: T.Sequence[str], kwargs: dict) -> ModuleReturnValue:
-        if len(args) != 1:
-            raise MesonException('fs.expanduser takes exactly one argument.')
-        return ModuleReturnValue(str(Path(args[0]).expanduser()), [])
+    @typed_pos_args('fs.expanduser', str)
+    def expanduser(self, state: 'ModuleState', args: T.Tuple[str], kwargs: T.Dict[str, T.Any]) -> ModuleReturnValue:
+        return ModuleReturnValue(os.path.expanduser(args[0]), [])
```
xclaesse, line 72: I thought we wanted to use pathlib everywhere?

xclaesse: Ok forget me, you explain in the commit message. I don't think performance matter in this case, but it's indeed not worth doing conversion from/to str.

xclaesse:
Looks good to me, but CI is red.

dcbaker:
Yup, I think it's that cleanup pass. Apparently `Path.something` is different than `os.path.something`.

xclaesse:
> Yup, I think it's that cleanup patch. Apparently `Path.something` is different than `os.path.something`.

Yep, it's full of subtle differences.

PR #8419, opened 2021-02-25
deprecated `meson.get_cross_property`

dcbaker:
It's a pure subset of `get_external_property`, and has odd behavior in
host == build configurations. `get_external_property` is clear, and uses
the standard `native : bool` syntax to control host vs build properties


xclaesse:
Lgtm

PR #8422, opened 2021-02-25
Simplify/optimize `get_external_property` and `get_cross_property`

dcbaker:
The implementations of these are way more complicated than they need to be, I've simplified them with a shared implementation, which both makes them easier to understand and presumably a little faster.


xclaesse:
This pr contains the one that deprecate the function. But +1 for both PRs.

dcbaker:
I'll just rebase it out. I apparently also need to fix up a couple of the "failing" tests

mesonbuild/interpreter.py
```diff
@@ -2239,51 +2239,32 @@ def version_method(self, args, kwargs):
     def project_name_method(self, args, kwargs):
         return self.interpreter.active_projectname
 
+    def __get_external_property_impl(self, propname: str, fallback: T.Optional[object], machine: MachineChoice) -> object:
+        """Shared implementaiton for get_cross_property and get_external_property."""
```
tp-m, line 2243: Typo: implementation

Issue #8421, opened 2021-02-25
Adding a subproject with library() makes meson Error


viric:
**Describe the bug**
If you have a configured build directory for a project, and you add a subproject with a library inside it, meson fails with this error:
    ERROR: Tried to get unknown builtin option newproject:default_library

The only solution is to delete the build directory and start over.

**To Reproduce**
Configure a build directory and add subproject with a library() or get_option('default_library'). @xclaesse is aware, from an irc conversation, saying it was a regression from 0.56

**Expected behavior**
Reconfigure fine.

**system parameters**
* native build
* windows 10 64-bit, linux-x86_64
* python 3.8.5
* meson 0.57.0 and 0.57.1
* ninja 1.10.2

dcbaker:
I have a pretty good idea what's going wrong.

dcbaker:
I have a test, and I know what's wrong. I'm just not sure what the best way to fix it is.

dcbaker:
I think I have a fix too

PR #8424, opened 2021-02-25
interpreter: correctly track whether a subproject is initialized

dcbaker:
The way the tracking is currently done it works if no new subprojects
are added to a configured build directory. For cases where we want to
add a new subproject, it fails because we don't initialize builtins for
that subproject. This corrects that by checking to see if the subproject
already exists, and if it doesn't initializes the bultins for it.

Fixes: #8421

### DIFF
--- a/mesonbuild/interpreter.py
+++ b/mesonbuild/interpreter.py
@@ -3171,7 +3171,11 @@ external dependencies (including libraries) must go to "dependencies".''')
         # have any effect.
         self.project_default_options = mesonlib.stringlistify(kwargs.get('default_options', []))
         self.project_default_options = coredata.create_options_dict(self.project_default_options, self.subproject)
-        if self.environment.first_invocation:
+
+        # If this is the first invocation we alway sneed to initialize
+        # builtins, if this is a subproject that is new in a re-invocation we
+        # need to initialize builtins for that
+        if self.environment.first_invocation or (self.subproject != '' and self.subproject not in self.subprojects):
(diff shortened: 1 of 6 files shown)


mesonbuild/interpreter.py
```diff
@@ -3171,7 +3171,10 @@ def func_project(self, node, args, kwargs):
         # have any effect.
         self.project_default_options = mesonlib.stringlistify(kwargs.get('default_options', []))
         self.project_default_options = coredata.create_options_dict(self.project_default_options, self.subproject)
-        if self.environment.first_invocation:
+
+        # For the main project we use first_invocation to know whether to init builtins,
+        # for subprojects see if the surpbojrect is in the known subprojects
+        if (self.subproject == '' and not self.environment.first_invocation) or self.subproject not in self.subprojects:
```
xclaesse, line 3177: I don't think you got that condition right. `self.subproject not in self.subprojects` that's always going to be true when subproject is `''`

xclaesse:
One day I should find the time to remove first_invocation completely. We can reevaluate all options on reconfigure thanks to cmd_line.txt that tells what has been set by user and is not default value.

mesonbuild/interpreter.py
dcbaker: Gah, yeah. It's annoying how many cases of "we need the same information about supprojects and the main project, but we encode them differently

xclaesse:
Lgtm

PR #8426, opened 2021-02-26
Allow printing UserOptions in the summary

The-King-of-Toasters:
Currently, if a user wants to print the status of a build option in the `summary()`, they would have do do something like:

```meson
summary({
	'feature': get_option('feature').enabled()
})
```

which prints the feature status as a boolean to string:

```
feature: True
```

This patch allows any `UserOption` to be printed in the summary, so:

```meson
summary({
	'feature': get_option('feature')
})
```

gives:

```
feature: enabled
```

Another enhancement would be to check for an instance of `UserFeatureOption` and highlight enabled/disabled as green/red like for normal bools.

### DIFF
--- a/docs/markdown/Reference-manual.md
+++ b/docs/markdown/Reference-manual.md
@@ -1356,6 +1356,7 @@ dictionaries does not guarantee ordering. `key` must be string,
 
 - an integer, boolean or string
 - *since 0.57.0* an external program or a dependency
+- *since 0.58.0* a feature option
 - a list of those.
 
 `summary()` can be called multiple times as long as the same
--- a/mesonbuild/interpreter.py
+++ b/mesonbuild/interpreter.py
@@ -1877,6 +1877,9 @@ class Summary:
                 elif isinstance(i, (ExternalProgram, Dependency)):
                     FeatureNew.single_use('dependency or external program in summary', '0.57.0', subproject)
                     formatted_values.append(i.summary_value())
+                elif isinstance(i, coredata.UserOption):
+                    FeatureNew.single_use('feature option in summary', '0.58.0', subproject)
+                    formatted_values.append(i.printable_value())
                 else:
                     m = 'Summary value in section {!r}, key {!r}, must be string, integer, boolean, dependency or external program'
                     raise InterpreterException(m.format(section, k))
--- a/test cases/unit/73 summary/meson.build
+++ b/test cases/unit/73 summary/meson.build
@@ -8,6 +8,7 @@ summary({'Some boolean': false,
          'Some string': 'Hello World',
          'A list': ['string', 1, true],
          'empty list': [],
+         'enabled_opt': get_option('enabled_opt'),
         }, section: 'Configuration')
 summary({'missing prog': find_program('xyzzy', required: false),
          'existing prog': import('python').find_installation(),
--- a/docs/markdown/snippets/summary-accepts-features.md
+++ b/docs/markdown/snippets/summary-accepts-features.md
@@ -1,0 +1,3 @@
+## `summary()` accepts features
+
+Build feature options can be passed to `summary()` as the value to be printed.
--- a/run_unittests.py
+++ b/run_unittests.py
@@ -4996,6 +4996,7 @@ class AllPlatformTests(BasePlatformTests):
                                  1
                                  True
                 empty list     :
+                enabled_opt    : enabled
                 A number       : 1
                 yes            : YES
                 no             : NO
--- a/test cases/unit/73 summary/meson_options.txt
+++ b/test cases/unit/73 summary/meson_options.txt
@@ -1,0 +1,1 @@
+option('enabled_opt', type: 'feature', value: 'enabled')


mesonbuild/interpreter.py
```diff
@@ -1877,6 +1877,8 @@ def add_section(self, section, values, kwargs, subproject):
                 elif isinstance(i, (ExternalProgram, Dependency)):
                     FeatureNew.single_use('dependency or external program in summary', '0.57.0', subproject)
                     formatted_values.append(i.summary_value())
+                elif isinstance(i, coredata.UserOption):
+                   formatted_values.append(i.printable_value())
```
xclaesse, line 1881: It's missing featurenew just like a couple lines above, to mark it new in 0.58.0

test cases/unit/73 summary/meson.build
```diff
@@ -8,6 +8,7 @@ summary({'Some boolean': false,
          'Some string': 'Hello World',
          'A list': ['string', 1, true],
          'empty list': [],
+         'enabled_opt': get_option('enabled_opt'),
```
xclaesse, line 11: There is a unit test that verify the output, you'll see that CI fails now. See test_summary in run_unittests.py

xclaesse:
Thanks for this contribution, I like the idea. Could you please also add doc in reference manual, and release notes snippet.

test cases/unit/73 summary/meson.build
The-King-of-Toasters: What do you think reads better:

- `feature in summary`
- `UserOption` in summary

mesonbuild/interpreter.py
The-King-of-Toasters: What do you think reads better:

- `feature in summary`
- `UserOption` in summary

xclaesse: UserOption is an internal object name, not something we should print to the user. Also this is not limited to feature options, or is it? What happens for other option types?

The-King-of-Toasters:
Done. LGTM?

docs/markdown/Reference-manual.md
```diff
@@ -1356,6 +1356,7 @@ dictionaries does not guarantee ordering. `key` must be string,
 
 - an integer, boolean or string
 - *since 0.57.0* an external program or a dependency
+- *since 0.58.0* a feature
```
xclaesse, line 1359:
```suggestion
- *since 0.58.0* a feature option 
```

docs/markdown/Release-notes-for-0.58.0.md
```diff
@@ -0,0 +1,10 @@
+---
+title: Release 0.58.0
+short-description: Release notes for 0.58.0
+...
+
+# New Features
+
+## `summary()` accepts features
+
+Build feature options can be passed to `summary()` as the value to be printed.
```
xclaesse, line 10: You should create a new file in that folder for your snippet. Files are concat for the release.

docs/sitemap.txt
```diff
@@ -81,6 +81,7 @@ index.md
 		Shipping-prebuilt-binaries-as-wraps.md
 		fallback-wraptool.md
 	Release-notes.md
+		Release-notes-for-0.58.0.md
```
xclaesse, line 84: Not yet, will be done when making the release.

mesonbuild/interpreter.py
```diff
@@ -1877,6 +1877,9 @@ def add_section(self, section, values, kwargs, subproject):
                 elif isinstance(i, (ExternalProgram, Dependency)):
                     FeatureNew.single_use('dependency or external program in summary', '0.57.0', subproject)
                     formatted_values.append(i.summary_value())
+                elif isinstance(i, coredata.UserOption):
+                    FeatureNew.single_use('feature in summary', '0.58.0', subproject)
```
xclaesse, line 1881:
```suggestion
                    FeatureNew.single_use('feature option in summary', '0.58.0', subproject)
```

xclaesse (commented):
Small nitpicking, looks good otherwise.

docs/markdown/Release-notes-for-0.58.0.md
The-King-of-Toasters: I'm unsure what you mean here. What file should I make?

xclaesse: Sorry, that's indeed impossible to discover for new contributors... Our process is to create a new file in `docs/markdown/snippets/` all those files there will be concat to make `docs/markdown/Release-notes-for-0.58.0.md` when doing the release.

xclaesse:
LGTM, thanks.

bonzini:
I am not sure about this, because in practice what you want is to say yes/no, not enabled/auto/disabled: auto should be replaced with the outcome of the test done at `meson setup` time.

Of course you are not forced to use it; however, I am not sure it's a good practice to pass feature options to summary, and therefore I'm not sure it should be supported.

eli-schwartz:
I observed the same thing at https://github.com/mesonbuild/meson/issues/8430#issuecomment-787412174

@The-King-of-Toasters mentioned there:
> I took another look at #8426, and I'm not sure it's worth keeping. I remember there was some global UserOption that could be printed, but I can't remember it.

So I think we need to come to a decision on whether or not to revert this before it lands in an actual release.

PR #8428, opened 2021-02-26
pkg-config: dump all PKG_CONFIG environment variables to simplify debugging

jmoguillansky-gpsw:
The output of pkg-config is affected by environment variables.
Currently, the meson log output doesn't provide sufficient information to debug pkg-config issues.
With this change, all the pkg-config environment variables are outputted, including the system env variables, to simplify debugging.

### DIFF
--- a/mesonbuild/dependencies/base.py
+++ b/mesonbuild/dependencies/base.py
@@ -676,14 +676,16 @@ class PkgConfigDependency(ExternalDependency):
         if sysroot:
             env['PKG_CONFIG_SYSROOT_DIR'] = sysroot
         new_pkg_config_path = ':'.join([p for p in extra_paths])
-        mlog.debug('PKG_CONFIG_PATH: ' + new_pkg_config_path)
         env['PKG_CONFIG_PATH'] = new_pkg_config_path
 
         pkg_config_libdir_prop = environment.properties[for_machine].get_pkg_config_libdir()
         if pkg_config_libdir_prop:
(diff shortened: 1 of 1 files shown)


mesonbuild/dependencies/base.py
```diff
@@ -676,14 +676,16 @@ def setup_env(env: T.MutableMapping[str, str], environment: 'Environment', for_m
...
 
         pkg_config_libdir_prop = environment.properties[for_machine].get_pkg_config_libdir()
         if pkg_config_libdir_prop:
             new_pkg_config_libdir = ':'.join([p for p in pkg_config_libdir_prop])
             env['PKG_CONFIG_LIBDIR'] = new_pkg_config_libdir
-            mlog.debug('PKG_CONFIG_LIBDIR: ' + new_pkg_config_libdir)
+        # Dump all PKG_CONFIG environment variables
+        for key in env:
+            if key.startswith('PKG_'):
+                mlog.debug('env[' + key + ']: ' + env[key])
```
dcbaker, line 688:
```suggestion
        for key, value in env.items():
            if key.startswith('PKG_'):
                mlog.debug(f'env[{key}]: {value}')
```

I could live without using `.items()` but the f-strings are so much more readable.

dcbaker (changes requested):

mesonbuild/dependencies/base.py
jmoguillansky-gpsw: Thanks, I applied your suggestion

dcbaker (approved):

PR #8429, opened 2021-02-26
rust: correctly handle -l link args

dcbaker:
Previously it was assumed (because of the way meson handles pkg-config
dependencies) that `-l` option would be static. This is not true, `-l`
could be a shared lib as well. So, let's respect that.

Co-Authored-by: Dylan Baker <dylan@pnwbakers.com>

### DIFF
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -1624,8 +1624,8 @@ int dummy;
                 elif a.startswith('-L'):
                     args.append(a)
                 elif a.startswith('-l'):
-                    # This should always be a static lib, I think
-                    args.extend(['-l', f'static={a[2:]}'])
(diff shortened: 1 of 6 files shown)


jljusten:
With #8404 plus this PR, it looks like I can now link without any `rust_args` hacks.

Why not two patches? Add your fix, then my test case patch? I think as one patch, it implies I had something to do with the fix. :)

I wonder why this also fixes the `-L` library path hack that I needed before. ?

dcbaker:
I think problem was that Meson was telling rustc to use a static library, even when one didn't exist. The unfortunate thing about Meson's static is that it really means "prefer static if possible" internally we have support for both "prefer-X" and "only-X" but we don't expose that.

dcbaker:
Okay, I split the commits.

jljusten:
Looks good. :+1: 

Thanks.

Issue #7140, opened 2020-05-14
[Dlang] d_module_versions is not applied to all needed files if generated source used


denizzzka:
**Describe the bug**
If generated source used `d_module_versions` is not applied to all needed files.

**To Reproduce**
meson.build
```Meson
project('meso_vers_d_bug', ['c', 'd'],
  version : '0.1',
  default_options : ['warning_level=3'])

cp = find_program('cp', native: true)

# dumb generator, just copies file
copy_gen = generator(
    cp,
    arguments : [
        '@INPUT@',
        '@BUILD_DIR@',
    ],
    output: '@PLAINNAME@',
)

# copied file
func_file = copy_gen.process(
  files('func.d'),
)

exe = executable(
  'meso_vers_d_bug',
  sources: [
    'main.d',
    func_file,
  ],
  d_module_versions: [
    'SomeVer',
  ],
)
```

main.d
```D
import std.stdio;
import func;

void main()
{
    version(SomeVer)
        writeln("Main: version SomeVer is defined");
    else
        writeln("Main: version SomeVer is not defined");

    somefunc();
}
```

func.d
```D
module func;

import std.stdio;

void somefunc()
{
    version(SomeVer)
        writeln("Func: version SomeVer is defined");
    else
        writeln("Func: version SomeVer is not defined");
}
```

After build and run these binary:
```
./meso_vers_d_bug
Main: version SomeVer is not defined
Func: version SomeVer is defined
```

**Expected behavior**
```
./meso_vers_d_bug
Main: version SomeVer is defined
Func: version SomeVer is defined
```

It can be achieved now by adding source file directly into executable():
```Meson
  sources: [
    'main.d',
    'func.d',
#    func_file,
  ],
```

**system parameters**
* Is this a [cross build](https://mesonbuild.com/Cross-compilation.html) or just a plain native build (for the same computer)?
Both affected

* what operating system (e.g. MacOS Catalina, Windows 10, CentOS 8.0, Ubuntu 18.04, etc.)
Debian

* what Python version are you using e.g. 3.8.0
3.8.2

* what `meson --version`
Current ~master 57e55b1a9ddce282f0b5d0508139f0f06ce952e5

* what `ninja --version` if it's a Ninja build
1.10.0

PR #8435, opened 2021-02-28
fix missing D versions with generated source

rtbo:
I've notice that when I add a generated source (with `custom_target`) to a target (e.g; an `executable` or `library`), the D versions specified with `d_module_versions` are not passed anymore to the compiler.
Inspecting the Ninja rules, I see that the D versions are passed to the compiler only for the generated source (which is first in the rules file), but not for the other sources.
I've tracked this down in the Meson code to the `get_feature_args` implementation that pops from `target.d_features`.

Tested to work in my own project.

Fixes: #7140

### DIFF
--- a/mesonbuild/compilers/d.py
+++ b/mesonbuild/compilers/d.py
@@ -153,6 +153,9 @@ class DmdLikeCompilerMixin(CompilerMixinBase):
     def get_feature_args(self, kwargs: T.Dict[str, T.Any], build_to_src: str) -> T.List[str]:
         # TODO: using a TypeDict here would improve this
         res = []
+        # get_feature_args can be called multiple times for the same target when there is generated source
+        # so we have to copy the kwargs (target.d_features) dict before popping from it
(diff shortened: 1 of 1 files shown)


dcbaker:
Instead of making a copy, can we just avoid using `pop()` so we're not mutating the keywords at all?

rtbo:
> Instead of making a copy, can we just avoid using pop() so we're not mutating the keywords at all?

That was my initial intention, but the function ends like this:
```python
        if kwargs:
            raise EnvironmentException('Unknown D compiler feature(s) selected: %s' % ', '.join(kwargs.keys()))
```
If we don't pop it won't be that easy to report the error.

rtbo:
This PR fixes #7140

dcbaker (approved):
Okay, we can fix this later whenever we get some kind of kwarg validator (I have a proposal for one, maybe someone will provide a better one), and can do this for now.

PR #8438, opened 2021-02-28
d_import_dirs add search path in the build directory

rtbo:
I'd like to have the possibility to import files from the build directory (resource files generated with `custom_target`).
The following works:
```meson
executable('my_exe', my_src,
    include_directories: include_directories('source'),
    d_import_dirs: [
        include_directories('views'),
        meson.current_build_dir() / 'views',
    ],
)
```
but is an anti-pattern, and rightfully warned:
```
WARNING: Building a path to the source dir is not supported. Use a relative path instead.
This will become a hard error in the future.
```
Instead I am proposing that `include_directories` have the same behavior in the context of `d_import_dirs` than in the context of adding header (or modules for D) search path: 
 - add the path to the source directory (as currently done)
 - add the path to the corresponding build directory 

This is already the documented behavior of `include_directories`, so this PR is more a fix than a new feature IMO.

### DIFF
--- a/mesonbuild/compilers/d.py
+++ b/mesonbuild/compilers/d.py
@@ -220,13 +220,15 @@ class DmdLikeCompilerMixin(CompilerMixinBase):
             for idir_obj in import_dirs:
                 basedir = idir_obj.get_curdir()
                 for idir in idir_obj.get_incdirs():
+                    bldtreedir = os.path.join(basedir, idir)
                     # Avoid superfluous '/.' at the end of paths when d is '.'
                     if idir not in ('', '.'):
-                        expdir = os.path.join(basedir, idir)
+                        expdir = bldtreedir
                     else:
                         expdir = basedir
                     srctreedir = os.path.join(build_to_src, expdir)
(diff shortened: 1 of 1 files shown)


dcbaker:
For C like languages the correct pattern is to do something like:
```meson
my_custom_target = custom_target('foo_h', output : ['foo.h'])
my_compiled_target = executable('mct', ['main.c', my_custom_target])
```

Meson will automatically add the `-I` flag automatically.  Does it make sense that when a generated target is added to D that we generate both a -I and a -J flag?

But this looks fine for fixing the current behavior.

dcbaker (approved):

rtbo:
>  Does it make sense that when a generated target is added to D that we generate both a -I and a -J flag?

Yes something like this I guess. The PR adds the -J switch to the build directory in a systematic way. I could have checked for a generated source as a condition to emit the additional switch, but I'm not that clever with meson code base.
Regarding the -I switch, AFAICT it is already done correctly.

Issue #8440, opened 2021-03-01
ERROR: add_dist_script may not be used in a subproject.


bog-dan-ro:
According to https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/3242#note_1047340 it seems that `add_dist_script` should work also for subprojects. 

Is it a leftover, or is a real restriction?

PR #8444, opened 2021-03-02
Allow add_dist_script() in subprojects

xclaesse:
Fixes: #8440.

### DIFF
--- a/docs/markdown/Reference-manual.md
+++ b/docs/markdown/Reference-manual.md
@@ -1907,16 +1907,25 @@ the following methods.
   archived. Note that this runs the script file that is in the
   _staging_ directory, not the one in the source directory. If the
   script file can not be found in the staging directory, it is a hard
-  error. This command can only invoked from the main project, calling
-  it from a subproject is a hard error. *(since 0.49.0)* Accepts multiple arguments
-  for the script. *(since 0.54.0)* The `MESON_SOURCE_ROOT` and `MESON_BUILD_ROOT`
-  environment variables are set when dist scripts are run.
-
+  error. The `MESON_DIST_ROOT` environment variables is set when dist scripts is
+  run.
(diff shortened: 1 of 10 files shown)


xclaesse:
@jpakkane @dcbaker any objection?

dcbaker:
I didn't look over the implementation too closely, but I'm fine with idea. Consider this an acked-by

jpakkane:
The thing that concerns me is that this may cause problems when people start disting subprojects that have not been updated to work with this. Dunno if we can do anything about it, though...

xclaesse:
They certainly won't work out of the box, they have to be ported to use MESON_PROJECT_ variables otherwise all paths are going to be wrong. But since it's currently not allowed to call add_dist_script from a subproject, it can't create regressions.

docs/markdown/Reference-manual.md
```diff
@@ -1877,16 +1877,20 @@ the following methods.
...
   *(since 0.55.0)* The output of `configure_file`, `files`, and `find_program`
   as well as strings.
-
   *(since 0.57.0)* `file` objects and the output of `configure_file` may be
   used as the `script_name` parameter.
+  *(since 0.58.0)* This command can be invoked from a subproject, it was a hard
+  error in earlier versions. Subproject dist scripts will only be executed
+  when running `meson dist --include-subprojects`. `MESON_PROJECT_SOURCE_ROOT`
+  and `MESON_PROJECT_BUILD_ROOT` and `MESON_PROJECT_DIST_ROOT` environment
+  variables are set when dist scripts are run.
```
jpakkane, line 1893: This should specify what they are set to when running the master project. Perhaps also add some examples of the values one would expect from them.

xclaesse: Done. Also added the release notes file that I forgot to git add.

PR #8449, opened 2021-03-02
external_project: Fix default cross compile parameters

leiflm:
The variable `{host}` cannot be successfully expanded.
Using the `@HOST@` parameter, as suggested by the documentation, works.


xclaesse:
Thanks!

Issue #8434, opened 2021-02-28
find_program('cpp') incorrectly returns path to clang++


jeremyhu:
When doing:

    cpp = find_program('cpp')
    message(cpp.full_path())

I see the following results:

Program clang++ found: YES
Message: /usr/bin/clang++

We should not be using clang++ (C++ compiler) for cpp (C Pre-Processor).

eli-schwartz:
```
Program cpp found: YES (/usr/bin/cpp)
Message: /usr/bin/cpp
```

Works for me...

jeremyhu:
[1024] ~/tmp/sample $ cat meson.build 
project('find preprocessor')
cpp = find_program('cpp')
message(cpp.full_path())

[1025] ~/tmp/sample $ CXX="clang++" meson build.cxx.set   
The Meson build system
Version: 0.57.1
Source dir: /Users/jeremy/tmp/sample
Build dir: /Users/jeremy/tmp/sample/build.cxx.set
Build type: native build
Project name: find preprocessor
Project version: undefined
Host machine cpu family: aarch64
Host machine cpu: arm64
Program clang++ found: YES
Message: /usr/bin/clang++
Build targets in project: 0

Found ninja-1.10.2 at /opt/buildX11/bin/ninja

[1026] ~/tmp/sample $ meson build.cxx.unset            
The Meson build system
Version: 0.57.1
Source dir: /Users/jeremy/tmp/sample
Build dir: /Users/jeremy/tmp/sample/build.cxx.unset
Build type: native build
Project name: find preprocessor
Project version: undefined
Host machine cpu family: aarch64
Host machine cpu: arm64
Program cpp found: YES (/usr/bin/cpp)
Message: /usr/bin/cpp
Build targets in project: 0

Found ninja-1.10.2 at /opt/buildX11/bin/ninja

eli-schwartz:
Apparently envconfig sets up the keys for internal programs using the identifier 'cpp' as the C plus plus compiler.

Similar problem for `find_program('c')` given a native file like:

```
[binaries]
c = ['gcc']
```

jon-turney:
In an email discussion on this issue, @dcbaker has suggested that this is working as designed.

If that's the case, I think:
- [the manual](https://mesonbuild.com/Reference-manual.html#find_program) needs to mention that using a meson language id with `find_program` has a special meaning
- this needs to work consistently, not only when the compiler for that language is set via env var or machine file

I have some doubts about if this is a good design, as:
[...]

PR #8452, opened 2021-03-02
Add representation of the C (or other language) preprocessor

dcbaker:
There are cases where users want/need to use the C Pre Processor directly. They might attempt this with code like:
```meson
cpp = find_program('cpp')
custom_target(..., command : [cpp, ...])
```
Of course, that runs into problems when they use a machine file, because meson uses `cpp` to mean `C++`, and they get their compiler instead of `cpp`. Of course, there's another problem, `cpp` isn't portable.

This MR attempts make that better by adding an object representation it can be gotten like this:
```meson
cc = meson.get_compiler('c')  # or cpp, objc, objc++, or any other compiler with a pre-processor
pp = cc.get_preprocessor()
if pp.has_argument('-foo')
   ...
endif
```

It provides a number of methods like the compiler object:
- cmd_array
- has_argument
- has_multi_argument
- get_supported_arguments
- first_supported_argument

But is also an `ExternalProgram`, like those returned by `find_program()` and can be used like one:
```meson
custom_target(..., command : [pp, ...])
```

This has the advantage of being portable, and has the ability to be tested like a compiler would be.

This is obviously still a WIP, it needs tests, docs, and such. I'm just trying to gauge if this is a useful solution or not.

Related: #8434

### DIFF
--- a/docs/markdown/snippets/pre_processor.md
+++ b/docs/markdown/snippets/pre_processor.md
@@ -1,0 +1,3 @@
+## A pre-processor object
+
+A new pre-processor object has been added.
--- a/mesonbuild/compilers/compilers.py
+++ b/mesonbuild/compilers/compilers.py
@@ -1031,7 +1031,7 @@ class Compiler(HoldableObject, metaclass=abc.ABCMeta):
         return []
 
     def get_preprocess_only_args(self) -> T.List[str]:
-        raise EnvironmentException('This compiler does not have a preprocessor')
+        raise EnvironmentException(f'The {self.id} compiler for {self.language} does not have a preprocessor')
 
     def get_default_include_dirs(self) -> T.List[str]:
         # TODO: This is a candidate for returning an immutable list
--- a/mesonbuild/interpreter/compiler.py
+++ b/mesonbuild/interpreter/compiler.py
@@ -2,9 +2,14 @@
 # Copyright 2012-2021 The Meson development team
 # Copyright © 2021 Intel Corporation
 
+from __future__ import annotations
 import enum
 import functools
+import itertools
+import os
 import typing as T
+
+from mesonbuild.interpreter.interpreter import SourceInputs
 
 from .. import build
 from .. import coredata
@@ -13,6 +18,7 @@ from .. import mesonlib
 from .. import mlog
 from ..compilers import SUFFIX_TO_LANG
 from ..compilers.compilers import CompileCheckMode
+from .interpreterobjects import ExternalProgramHolder
 from ..interpreterbase import (ObjectHolder, noPosargs, noKwargs,
                                FeatureNew, disablerIfNotFound,
                                InterpreterException)
@@ -21,9 +27,10 @@ from .interpreterobjects import (extract_required_kwarg, extract_search_dirs)
 from .type_checking import REQUIRED_KW, in_set_validator, NoneType
 
 if T.TYPE_CHECKING:
-    from ..interpreter import Interpreter
     from ..compilers import Compiler, RunResult
+    from ..interpreter import Interpreter
     from ..interpreterbase import TYPE_var, TYPE_kwargs
+    from ..programs import ExternalProgram
     from .kwargs import ExtractRequired, ExtractSearchDirs
 
     from typing_extensions import TypedDict, Literal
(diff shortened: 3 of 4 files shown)


dcbaker:
@jeremyhu, I know this affects you (and probably other xorg people)

jpakkane:
How portable is this as a concept? Does MSVC provide a standalone preprocessor executable?

eli-schwartz:
The visualstudio compiler mixin returns `['compiler', '/EP']`.

The gcc mixin doesn't return cpp either, it returns `['gcc', '-E', '-P']`

dcbaker:
The point isn't to provide `cpp`, it's to provide a portable mechanism to get a thing that does C (or other language with a pre-processor, like D) and check to see what arguments it supports, just like we would with a compiler.

e820:
Truly lovely addition - just gotta make sure to remember passing `-P` / `--no-line-commands` to clang-like preprocessors to avoid funky output.

jon-turney:
> Truly lovely addition - just gotta make sure to remember passing `-P` / `--no-line-commands` to clang-like preprocessors to avoid funky output.

Not sure if this means we should assume `-P` is wanted, but that seems unsafe, as we can't know if the preprocessor output is going to be used as is, or fed into some sort of tool which can report errors (in which case #line directives are wanted to enable those errors to be reported against the source location)

e820:
> > Truly lovely addition - just gotta make sure to remember passing `-P` / `--no-line-commands` to clang-like preprocessors to avoid funky output.
> 
> Not sure if this means we should assume `-P` is wanted, but that seems unsafe, as we can't know if the preprocessor output is going to be used as is, or fed into some sort of tool which can report errors (in which case #line directives are wanted to enable those errors to be reported against the source location)

That certainly is a justified concern. Do you a few specific ones in mind? I absolutely agree that the preprocessor representation should be generic enough but I'd also like to avoid having to perform manual compiler id checks to determine the required preprocessor flags if possible.

dcbaker:
@e820 you shouldn't have to, the design is to something use `has_argument` or `get_supported_arguments` instead of doing ID checks, just like you would with a compiler

dcbaker:
I've rebased this on master, but nothing more ATM. I'll try to find time to add some tests and documentation

jon-turney:
> There are cases where users want/need to use the C Pre Processor directly. They might attempt this with code like:
> 
> ```meson
> cpp = find_program('cpp')
> custom_target(..., command : [cpp, ...])
> ```
> 
> Of course, that runs into problems when they use a machine file, because meson uses `cpp` to mean `C++`, and they get their compiler instead of `cpp`. Of course, there's another problem, `cpp` isn't portable.

Cross-referencing #8434 for that.

xclaesse:
Oh, I did not know you were working on this, that's nice. There is a little bit of overlap with what I've been working on there: https://github.com/mesonbuild/meson/pull/10149. You worked mostly on interpreter side of preprocessor while I worked mostly on the backend side. Not sure there is much we can do in common.

Another thing I've been wanting is for example `file = cc.preprocess('foo.c')` that create a target that preprocess but not compile the file. For something like that: https://gitlab.freedesktop.org/gstreamer/meson-ports/libffi/-/blob/meson/src/meson.build#L73.

dcbaker:
I mostly need this because xorg had some interesting legacy code that uses the c pre processor as a code generator, but yeah, if we can make it with together that would be nice

dcbaker:
@xclaesse I've added a completely untested `preprocess` method to the `Preprocessor` object, but I think to make it reliable I'll need your series

xclaesse:
@dcbaker I think ideally it should somehow go through the backend to get c_args and get -I from all deps, etc. But maybe that's going too far?

xclaesse:
let's keep that out of scope for this PR, we can think about something later.

e820:
Any updates on this on PR?

nirbheek:
```
<dcbaker@dcbaker:matrix.org> I think 8452 should be closed
<dcbaker@dcbaker:matrix.org> I’m on my phone though, so not logged into GitHub
```

Context: we now have `cc.preprocess()`, as pointed out by @tristan957 on matrix.

jon-turney:
> There are cases where users want/need to use the C Pre Processor directly. They might attempt this with code like:
> 
> ```meson
> cpp = find_program('cpp')
> custom_target(..., command : [cpp, ...])
> ```
> 
> Of course, that runs into problems when they use a machine file, because meson uses `cpp` to mean `C++`, and they get their compiler instead of `cpp`. Of course, there's another problem, `cpp` isn't portable.

I think there's maybe still a small issue here that if you had a project which just used the pre-processor, and was expecting to be built in an environment where only the pre-processor (and no compiler) is available, you can't write `find_program('cpp', 'mcpp')` and have it always work, due to this confusion of identifiers.

PR #8453, opened 2021-03-02
Windows Subsystem for Linux can run .exe without mono interpreter

xclaesse:


mesonbuild/mesonlib/universal.py
```diff
@@ -597,6 +598,8 @@ def is_windows() -> bool:
     platname = platform.system().lower()
     return platname == 'windows'
 
+def is_wsl() -> bool:
+    return is_linux() and 'microsoft' in platform.version().lower()
```
dcbaker, line 602: Shouldn't we have one of these in the `MachineInfo` as well? We really should stop using the `mesonlib` versions of these functions.

xclaesse: Hm, I did not know MachineInfo duplicates all those methods. This should be refactored indeed, but I think it's a bit off topic for this PR, no real point in adding that in MachineInfo if nobody uses it yet... I would like to keep this PR simple because it's a regression so should be for 0.57.2.

xclaesse:
Ok, tested on my Windows machine and it works :)

PR #5209, opened 2019-04-03
Add include() function

TheQwertiest:
Based on https://github.com/mesonbuild/meson/pull/3557.

Proposal: global `include(path_to_file)` function that works similar to `subdir(dir)`, but has the following restrictions:
 * The path can be either relative or absolute and can point to the arbitrary file (if it satisfies restrictions listed below).
 * Included file must be a valid meson file.
 * Same file can be included only once.
 * Can't include `meson.build` file.
 * Can't include file from outside of source directory.
 * Can't include file from subprojects directory (unless invoked from inside subprojects directory).
 * File must have a specific extension (currently `*.meson`).
 
Reasoning:
The main purpose of `include()` is to avoid overly long `meson.build` by splitting it in multiple files and to avoid needless clutter (currently you have to create a separate folder with `meson.build` inside to achieve the same functionality).

PR #8455, opened 2021-03-02
Refactor subdir visitation to track files instead of dirs.

jpakkane:
For #5209, it's better to have only one place where processed buildfiles are stored.

Added an explicit test for this, since one did not seem to already exist.

### DIFF
--- a/mesonbuild/ast/interpreter.py
+++ b/mesonbuild/ast/interpreter.py
@@ -70,10 +70,10 @@ class AstInterpreter(interpreterbase.InterpreterBase):
     def __init__(self, source_root: str, subdir: str, subproject: str, visitors: T.Optional[T.List[AstVisitor]] = None):
         super().__init__(source_root, subdir, subproject)
         self.visitors = visitors if visitors is not None else []
-        self.visited_subdirs = {}     # type: T.Dict[str, bool]
-        self.assignments = {}         # type: T.Dict[str, BaseNode]
-        self.assign_vals = {}         # type: T.Dict[str, T.Any]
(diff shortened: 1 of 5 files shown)


mesonbuild/ast/interpreter.py
```diff
@@ -70,10 +70,10 @@ class AstInterpreter(interpreterbase.InterpreterBase):
     def __init__(self, source_root: str, subdir: str, subproject: str, visitors: T.Optional[T.List[AstVisitor]] = None):
         super().__init__(source_root, subdir, subproject)
         self.visitors = visitors if visitors is not None else []
-        self.visited_subdirs = {}     # type: T.Dict[str, bool]
-        self.assignments = {}         # type: T.Dict[str, BaseNode]
-        self.assign_vals = {}         # type: T.Dict[str, T.Any]
-        self.reverse_assignment = {}  # type: T.Dict[str, BaseNode]
+        self.processed_buildfiles = {} # type: T.Dict[str, bool]
```
dcbaker, line 73: why isn't this a `set()`, we never use the value AFAICT?

dcbaker (changes requested):

mesonbuild/interpreter.py
```diff
@@ -2431,7 +2431,7 @@ def __init__(
             self.sanity_check_ast()
         self.builtin.update({'meson': MesonMain(build, self)})
         self.generators = []
-        self.visited_subdirs = {}
+        self.processed_buildfiles = {}
```
dcbaker, line 2434: Same here, also, let's add an annotation here.

PR #8457, opened 2021-03-02
Duplicate pkg config path

tristan957:
Previously builds would *potentially* get spammed with messaging at
configure time that duplicate entries in an array would be an error in
the future, and the cause was because the same entries were getting
added over and over to pkg_config_path.


tristan957:
No I am still an idiot fk.

PR #8458, opened 2021-03-02
Fix duplicate pkg_config_path entries

tristan957:
Previously builds would *potentially* get sammed with messaging at
configure time that duplicate entries in an array would be an error in
the future, and the cause was because the same entries were getting
added over and over to pkg_config_path.p

### DIFF
--- a/mesonbuild/dependencies/base.py
+++ b/mesonbuild/dependencies/base.py
@@ -669,8 +669,8 @@ class PkgConfigDependency(ExternalDependency):
     @staticmethod
     def setup_env(env: T.MutableMapping[str, str], environment: 'Environment', for_machine: MachineChoice,
                   extra_path: T.Optional[str] = None) -> None:
-        extra_paths: T.List[str] = environment.coredata.options[OptionKey('pkg_config_path', machine=for_machine)].value
-        if extra_path:
+        extra_paths: T.List[str] = environment.coredata.options[OptionKey('pkg_config_path', machine=for_machine)].value[:]
+        if extra_path and extra_path not in extra_paths:
             extra_paths.append(extra_path)
         sysroot = environment.properties[for_machine].get_sys_root()
         if sysroot:
--- a/run_unittests.py
+++ b/run_unittests.py
@@ -7318,6 +7318,18 @@ class LinuxlikeTests(BasePlatformTests):
         self.assertEqual(relative_path_dep.get_link_args(), link_args)
 
     @skipIfNoPkgconfig
+    def test_pkgconfig_duplicate_path_entries(self):
+        testdir = os.path.join(self.unit_test_dir, '111 pkgconfig duplicate path entries')
+        pkg_dir = os.path.join(testdir, 'pkgconfig')
+
+        env = get_fake_env(testdir, self.builddir, self.prefix)
+        env.coredata.set_options({OptionKey('pkg_config_path'): pkg_dir}, subproject='')
+
+        PkgConfigDependency.setup_env({}, env, MachineChoice.HOST, pkg_dir)
+        pkg_config_path = env.coredata.options[OptionKey('pkg_config_path')].value
+        self.assertTrue(len(pkg_config_path) == 1)
+
+    @skipIfNoPkgconfig
     def test_pkgconfig_internal_libraries(self):
         '''
         '''


tristan957:
Sorry for the spam but my git abilities are utter shite today

xclaesse:
Hmm, are you sure this fix your issue? The only place where `setup_env()` is called with `extra_path` is from `unstable_external_project.py`. It is true that it could add a duplicate path, but only in the env passed to the external project's configure.

xclaesse:
Still LGTM since it cannot hurt, but I would be surprised this is the bug you were looking for.

tristan957:
@xclaesse I think this is the issue. I am fairly certain this issue only crops up (at least in my case) when I have external project subprojects. I did just try a rebuild on the project, I see this often on and it didn't show back up after this change.

jpakkane:
There should be a test to avoid future regressions. It would also verify whether this fixes the issue or not.

tristan957:
Where would be a good place to add a regression test like this

jpakkane:
`run_unittests.py`, probably under `LinuxlikeTests`.

xclaesse:
if you've got difficulties to reproduce please give me a stripped down meson.build I can use to reproduce locally, that would help. Just by reading the code I don't see how it could lead to a warning.

tristan957:
@xclaesse here are the logs that might help understand why the issue gets hit
```
|Executing subproject curl method meson 
|
|DEPRECATION: Duplicated values in array option is deprecated. This will become a hard error in the future.
|DEPRECATION: Duplicated values in array option is deprecated. This will become a hard error in the future.
|Project name: curl
|Project version: 7.61.1
|C compiler for the host machine: cc (gcc 10.2.1 "cc (GCC) 10.2.1 20201125 (Red Hat 10.2.1-9)")
|C linker for the host machine: cc ld.bfd 2.34-6
|DEPRECATION: Duplicated values in array option is deprecated. This will become a hard error in the future.
|DEPRECATION: Duplicated values in array option is deprecated. This will become a hard error in the future.
|DEPRECATION: Duplicated values in array option is deprecated. This will become a hard error in the future.
|DEPRECATION: Duplicated values in array option is deprecated. This will become a hard error in the future.
```
Curl is an ExternalProject in this case. I haven't had time to come back and add a test yet. Apologies

tristan957:
Got a unit test up for it! Not super difficult.

tristan957:
I think the way the deprecation message hits me is through UserArrayOption.validate_value()

tristan957:
@xclaesse anything I need to do to get this PR accepted?

tristan957:
Thanks for confirming the fix.

xclaesse:
Of course it fix the unit test because it calls PkgConfigDependency.setup_env() directly. But that case cannot be reached otherwise.

There are only 2 places that calls PkgConfigDependency.setup_env():
- From PkgConfigDependency._call_pkgbin() where extra_path is None.
- From ExternalProject where it's used to add meson-uninstalled directory into the env when calling external project's configure script. That will never reach the warning `Duplicated values in array option is deprecated`, unless you're configuring a meson project using externalproject which would make no sense...

Not saying this MR is wrong, it cannot hurt to check for dups, but that's really unlikely to fix the root issue. Unless there is something I totally missed.

tristan957:
```
$ meson build -Dforce_fallback_for=curl,libyaml
The Meson build system
Version: 0.57.1
Source dir: /home/tpartin/Projects/hse
Build dir: /home/tpartin/Projects/hse/build
Build type: native build
Project name: hse
Project version: 1.10.0
C compiler for the host machine: ccache cc (gcc 10.2.1 "cc (GCC) 10.2.1 20201125 (Red Hat 10.2.1-9)")
C linker for the host machine: cc ld.bfd 2.34-6
Picked up JAVA_TOOL_OPTIONS: -DproxySet=true -Dhttp.proxyHost=sbuproxy -Dhttp.proxyPort=8080 -Dhttps.proxyHost=sbuproxy -Dhttps.proxyPort=8080
Picked up _JAVA_OPTIONS: -DproxySet=true -Dhttp.proxyHost=sbuproxy -Dhttp.proxyPort=8080 -Dhttps.proxyHost=sbuproxy -Dhttps.proxyPort=8080
Picked up JAVA_TOOL_OPTIONS: -DproxySet=true -Dhttp.proxyHost=sbuproxy -Dhttp.proxyPort=8080 -Dhttps.proxyHost=sbuproxy -Dhttps.proxyPort=8080
Picked up _JAVA_OPTIONS: -DproxySet=true -Dhttp.proxyHost=sbuproxy -Dhttp.proxyPort=8080 -Dhttps.proxyHost=sbuproxy -Dhttps.proxyPort=8080
Java compiler for the host machine: javac (unknown 11.0.9.1)
Picked up JAVA_TOOL_OPTIONS: -DproxySet=true -Dhttp.proxyHost=sbuproxy -Dhttp.proxyPort=8080 -Dhttps.proxyHost=sbuproxy -Dhttps.proxyPort=8080
Picked up _JAVA_OPTIONS: -DproxySet=true -Dhttp.proxyHost=sbuproxy -Dhttp.proxyPort=8080 -Dhttps.proxyHost=sbuproxy -Dhttps.proxyPort=8080
Picked up JAVA_TOOL_OPTIONS: -DproxySet=true -Dhttp.proxyHost=sbuproxy -Dhttp.proxyPort=8080 -Dhttps.proxyHost=sbuproxy -Dhttps.proxyPort=8080
Picked up _JAVA_OPTIONS: -DproxySet=true -Dhttp.proxyHost=sbuproxy -Dhttp.proxyPort=8080 -Dhttps.proxyHost=sbuproxy -Dhttps.proxyPort=8080
Host machine cpu: x86_64
Program python3 found: YES (/usr/bin/python3)
Program git found: YES (/usr/bin/git)
Looking for a fallback subproject for the dependency libcurl because:
Use of fallback was forced for that specific subproject

|Executing subproject curl method meson 
|
|Project name: curl
|Project version: 7.61.1
|C compiler for the host machine: ccache cc (gcc 10.2.1 "cc (GCC) 10.2.1 20201125 (Red Hat 10.2.1-9)")
|C linker for the host machine: cc ld.bfd 2.34-6
|Program make found: YES (/usr/bin/make)
|Program /home/tpartin/Projects/hse/subprojects/curl-curl-7_61_1/configure found: YES (/home/tpartin/Projects/hse/subprojects/curl-curl-7_61_1/configure)
|External project curl-curl-7_61_1: configure
|Build targets in project: 1
|Subproject curl finished.

Dependency libcurl from subproject subprojects/curl-curl-7_61_1 found: YES 7.61.1
Looking for a fallback subproject for the dependency yaml-0.1 because:
Use of fallback was forced for that specific subproject

|Executing subproject libyaml method meson 
|
|Project name: libyaml
|Project version: 0.1.7
|C compiler for the host machine: ccache cc (gcc 10.2.1 "cc (GCC) 10.2.1 20201125 (Red Hat 10.2.1-9)")
|C linker for the host machine: cc ld.bfd 2.34-6
|Program make found: YES (/usr/bin/make)
|Program /home/tpartin/Projects/hse/subprojects/libyaml-0.1.7/configure found: YES (/home/tpartin/Projects/hse/subprojects/libyaml-0.1.7/configure)
|External project libyaml-0.1.7: configure
|Build targets in project: 2
|Subproject libyaml finished.

Dependency yaml-0.1 from subproject subprojects/libyaml-0.1.7 found: YES 0.1.7
Found pkg-config: /usr/bin/pkg-config (1.6.3)
Run-time dependency uuid found: YES 2.35.2
Run-time dependency liburcu-bp found: YES 0.11.1
Run-time dependency libcjson found: YES 1.7.14
Run-time dependency libbsd found: YES 0.10.0

|Executing subproject lz4 method meson 
|
|DEPRECATION: Duplicated values in array option is deprecated. This will become a hard error in the future.
|DEPRECATION: Duplicated values in array option is deprecated. This will become a hard error in the future.
|Project name: lz4
|Project version: 1.9.2
|C compiler for the host machine: ccache cc (gcc 10.2.1 "cc (GCC) 10.2.1 20201125 (Red Hat 10.2.1-9)")
|C linker for the host machine: cc ld.bfd 2.34-6
|DEPRECATION: Duplicated values in array option is deprecated. This will become a hard error in the future.
|DEPRECATION: Duplicated values in array option is deprecated. This will become a hard error in the future.
|DEPRECATION: Duplicated values in array option is deprecated. This will become a hard error in the future.
|DEPRECATION: Duplicated values in array option is deprecated. This will become a hard error in the future.
|Build targets in project: 3
|Subproject lz4 finished.


|Executing subproject xxhash method meson 
|
|DEPRECATION: Duplicated values in array option is deprecated. This will become a hard error in the future.
|DEPRECATION: Duplicated values in array option is deprecated. This will become a hard error in the future.
|Project name: xxHash
|Project version: 0.8.0
|C compiler for the host machine: ccache cc (gcc 10.2.1 "cc (GCC) 10.2.1 20201125 (Red Hat 10.2.1-9)")
|C linker for the host machine: cc ld.bfd 2.34-6
|DEPRECATION: Duplicated values in array option is deprecated. This will become a hard error in the future.
|DEPRECATION: Duplicated values in array option is deprecated. This will become a hard error in the future.
|DEPRECATION: Duplicated values in array option is deprecated. This will become a hard error in the future.
|DEPRECATION: Duplicated values in array option is deprecated. This will become a hard error in the future.
|Configuring libxxhash.pc using configuration
|Build targets in project: 4
|Subproject xxhash finished.
...
```

A better log. `curl` and `libyaml` are ExternalProjects here and `lz4` and `xxhash` are Meson subprojects. The issue seems to be when you have ExternalProjects getting configured before Meson subprojects because pkg_config_path is a global setting and not per subproject?

xclaesse:
After configuring a project that has an external project, the `pkg_config_path` option has the directory `builddir/meson-uninstalled` that's not normal. It should only be added into the env when running the ep configure not being added to meson options.

On reconfigure it gets added a 2nd time which leads to the warning.

xclaesse:
Ok, found the root cause. In `setup_env()` it does `extra_paths: T.List[str] = environment.coredata.options[OptionKey('pkg_config_path', machine=for_machine)].value` so extra_paths is a reference to the list from coredata, then it append() to it which modifies the value in coredata! We must do a shallow copy of the list.

tristan957:
I will add this to my pr

tristan957:
@xclaesse thanks for reviewing the PR so thoroughly :). Always good to find root causes, which I should have done much earlier.

PR #8460, opened 2021-03-03
Simplify module API

xclaesse:
- ModuleState is now a real class that will have methods in the future
      for actions modules needs, instead of using interpreter internal API.
    - New ModuleObject base class, similar to InterpreterObject, that should
      be used by all objects returned by modules. Its methods gets the
      ModuleState passed as first argument. It has a `methods` dictionary to
      define what is public API that can be called from build definition.
    - Method return value is not required to be a ModuleReturnValue any
      more, it can be any type that interpreter can holderify, including
      ModuleObject.
    - Legacy module API is maintained until we port all modules.
    
    In the future modules should be updated:
    - Use methods dict.
    - Remove snippets.
    - Custom objects returned by modules should all be subclass of
      ModuleObject to get the state iface in their methods.
    - Modules should never call into interpreter directly and instead state
      object should have wrapper API.
    - Stop using ModuleReturnValue in methods that just return simple
      objects like strings. Possibly remove ModuleReturnValue completely
      since all objects that needs to be processed by interpreter (e.g.
      CustomTarget) should be created through ModuleState API.

### DIFF
--- a/mesonbuild/interpreter.py
+++ b/mesonbuild/interpreter.py
@@ -33,7 +33,7 @@ from .interpreterbase import FeatureNew, FeatureDeprecated, FeatureNewKwargs, Fe
 from .interpreterbase import ObjectHolder, MesonVersionString
 from .interpreterbase import TYPE_var, TYPE_nkwargs
 from .interpreterbase import typed_pos_args
-from .modules import ModuleReturnValue, ExtensionModule
+from .modules import ModuleReturnValue, ModuleObject, ModuleState
 from .cmake import CMakeInterpreter
 from .backend.backends import TestProtocol, Backend, ExecutableSerialisation
 
@@ -55,7 +55,6 @@ if T.TYPE_CHECKING:
     from .compilers import Compiler
     from .envconfig import MachineInfo
     from .environment import Environment
-    from .modules import ExtensionModule
 
 permitted_method_kwargs = {
     'partial_dependency': {'compile_args', 'link_args', 'links', 'includes',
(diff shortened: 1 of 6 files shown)


xclaesse:
This is a rebase of https://github.com/mesonbuild/meson/pull/6766 that also keeps legacy module API working to make it easier to port them one by one in the future.

obilaniu:
Could you please demonstrate how a port of the unstable CUDA module would be done? Currently, it's got 3 functions all of which are `str` or `List[str]` manipulations returning `str` or `List[str]`, defines no custom object, and uses the `state` in only one place to determine whether the host system is Linux or Windows.

I am unofficially its maintainer, but I don't quite understand what the new contract for modules will be.

xclaesse:
@obilaniu With this change you can return str and list[str] directly without using ModuleReturnValue, really simple change for cuda module. The other change cuda module should do is adding its public methods into the `methods` dict in __init__ just like other InterpreterObject does in interpreter.py.

mesonbuild/interpreter.py
```diff
@@ -2589,28 +2540,20 @@ def process_new_values(self, invalues):
             elif isinstance(v, build.Data):
                 self.build.data.append(v)
             elif isinstance(v, dependencies.ExternalProgram):
-                return ExternalProgramHolder(v, self.subproject)
+                pass
```
dcbaker, line 2543: Could this case be combined with the final `elif`?

mesonbuild/modules/__init__.py
```diff
@@ -18,21 +18,58 @@
...
+        self.methods = {}  # type: T.Dict[str, T.Callable[[T.List[TYPE_nvar], TYPE_nkwargs], TYPE_var]]
+        # FIXME: Port all modules to stop using self.interpreter and use API on
+        # ModuleState instead.
         self.interpreter = interpreter
-        self.snippets = set()  # type: T.Set[str] # List of methods that operate only on the interpreter.
-
-    def is_snippet(self, funcname: str) -> bool:
-        return funcname in self.snippets
+        # FIXME: Port all modules to remove snippets methods.
+        self.snippets = set()
```
dcbaker, line 68: while you're here, I thinks this could be annotated as `self.snippets: T.Set[str] = set()`

mesonbuild/modules/__init__.py
```diff
@@ -18,21 +18,58 @@
...
+        self.headers = interpreter.build.get_headers()
+        self.man = interpreter.build.get_man()
+        self.global_args = interpreter.build.global_args.host
+        self.project_args = interpreter.build.projects_args.host.get(interpreter.subproject, {})
+        self.build_machine = interpreter.builtin['build_machine'].held_object
+        self.host_machine = interpreter.builtin['host_machine'].held_object
+        self.target_machine = interpreter.builtin['target_machine'].held_object
+        self.current_node = interpreter.current_node
+
+class ModuleObject:
```
dcbaker, line 59: Doesn't this need to be a subclass of `InterpeterObject`?

dcbaker:
I've got a couple of small nits, but in general this looks like a nice cleanup.

mesonbuild/interpreter.py
xclaesse: Could be a nice little cleanup indeed.

mesonbuild/modules/__init__.py
xclaesse: That's going away in the next PR, but I'll do that in the meantime, yes.

xclaesse:
@dcbaker the failing CI seems unrelated, could it be https://github.com/mesonbuild/meson/pull/8468 ?

xclaesse:
> @dcbaker the failing CI seems unrelated, could it be #8468 ?

Sorry, meant @mensinda ^

mensinda:
@xclaesse yeah, looks like it.

PR #8463, opened 2021-03-03
minstall: Correctly set uid/gid of installed files

Saur2000:
In commit caab4d3d, the uid and gid arguments passed to os.chown() by
set_chown() were accidentally swapped, causing files to end up with
incorrect owner/group if the owner and group are not the same.

Also update the documentation to better indicate which argument to
install_mode is which.

Signed-off-by: Peter Kjellerstedt <peter.kjellerstedt@axis.com>

### DIFF
--- a/docs/markdown/Reference-manual.md
+++ b/docs/markdown/Reference-manual.md
@@ -1065,7 +1065,7 @@ arguments. The following keyword arguments are supported:
 
   `install_mode: 'rw-r--r--'` for just the file mode
 
-  `install_mode: ['rw-r--r--', 'nobody', 'nobody']` for the file mode and the user/group
+  `install_mode: ['rw-r--r--', 'nobody', 'nogroup']` for the file mode and the user/group
 
   `install_mode: ['rw-r-----', 0, 0]` for the file mode and uid/gid
 
--- a/mesonbuild/minstall.py
+++ b/mesonbuild/minstall.py
@@ -148,7 +148,7 @@ def set_chown(path: str, user: T.Optional[str] = None, group: T.Optional[str] = 
         Use a real function rather than a lambda to help mypy out. Also real
         functions are faster.
         """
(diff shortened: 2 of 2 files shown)


xclaesse:
Nice catch, LGTM when CI is done.

Saur2000:
I believe there should be some test that actually validates this, but I have no idea how to write that, so feel free to complement this commit with that. AFAICT, the tests in "test cases/common/191 install_mode" only validates that the files can be created, not that they actually have the right mode and owner/group.

xclaesse:
A unit test would be ideal of course. You can write one in run_unittests.py that init, build and install that 191 install_mode project and then check the mode of installed files.

Saur2000:
Turns out there is a unit test already that validates the modes of the installed files (LinuxlikeTests.test_installed_modes_extended). However, there is nothing that validates the owner/group of the installed files. Doing that would probably require mocking os.chown() somehow, but I have no idea how to go about doing that as it is the `meson.py` process started by `ninja install` that would need to be mocked and not `run_unitests.py`...

eli-schwartz:
Mocking os.chown is needed because when running unittests as non-root you can't really do install uids/gids at all.

Except, you actually can. There is a tool called fakeroot which mocks out the system call for you, maybe this will help. You'd probably need to run the unittest `meson install` and the verification checks under the same fakeroot subprocess though, which could introduce different kinds of awkwardness... But it should not need to be mocking out meson.py

There are pros and cons to both. Hmm...

Saur2000:
Well, since no other test is using fakeroot, adding a test that does is out of my league.

jpakkane:
Based on the comments above the code, this piece of functionality is unlikely to change much until Python upstream is fixed (if it ever does). Since the test would seem to get complicated and the fix is straightforward, I'd say we can merge this as is. Thanks.

eli-schwartz:
Yeah, I dunno how to actually test this either. :/

PR #8467, opened 2021-03-04
mintro: include the correct target filenames for flat layout

bonzini:
`meson introspect --targets` does not take flat layout into account, and this also breaks `meson test` starting with 0.57. This pull request adds a fix and testcase.

This should be included in 0.57.2.

@mensinda, please take a look. Thanks!

### DIFF
--- a/mesonbuild/mintro.py
+++ b/mesonbuild/mintro.py
@@ -119,6 +119,12 @@ def list_installed(installdata: backends.InstallData) -> T.Dict[str, str]:
             res[i.path] = os.path.join(installdata.prefix, i.install_path)
     return res
 
+def get_target_dir(coredata: cdata.CoreData, subdir: str) -> str:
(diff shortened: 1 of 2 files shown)


mesonbuild/mintro.py
```diff
@@ -120,9 +119,16 @@ def list_installed(installdata: backends.InstallData) -> T.Dict[str, str]:
             res[i.path] = os.path.join(installdata.prefix, i.install_path)
     return res
 
+def get_target_dir(coredata: cdata.CoreData, subdir: str) -> bool:
```
mensinda, line 122: This function returns a string and not a `bool`.

mensinda (commented):
LGTM, besides the one typing issue.

mensinda:
Looks like the Mac CI error is caused by the brew update to qt6. Should be fixed with #8468

mensinda (approved):
LGTM

Merging since the CI error is unrelated.

PR #8472, opened 2021-03-04
Port CUDA module to new API

obilaniu:
Consequent to #8460 and [comments](https://github.com/mesonbuild/meson/pull/8460#issuecomment-789697043). @xclaesse?


obilaniu:
@xclaesse I think this is ready for review. I hope I did the port correctly. I also took the liberty of adding type annotations.

xclaesse:
Yes that's right. Another thing you can modernize is to use @typed_pos_args on all your public methods to validate args. But LGTM already.

PR #8474, opened 2021-03-04
Update python 3.5 string format to python 3.6 in Wrap logic

michaelbrockus:
This PR is mainly focused on updating the format strings. See if we go green.


eli-schwartz:
This could be done in a much more consistent manner. #8475

xclaesse:
+1 for doing it on the whole code base.

PR #8475, opened 2021-03-04
Pyupgrade all the things

eli-schwartz:
Fixes:
- some invalid escape sequences that will become errors in future versions of python
- f-strings everywhere!

Closes #8474


dcbaker (approved):
I don't know if you want to land https://github.com/mesonbuild/meson/pull/8476 or this first, but either way this looks good.

test cases/unit/39 python extmodule/blaster.py
```diff
@@ -10,5 +10,5 @@
     sys.exit(1)
 
 if result != 1:
-    print('Returned result {} is not 1.'.format(result))
+    print(f'Returned result {result} is not 1.')
```
eli-schwartz, line 13: The shebang LIES, this is running on python2 on macOS and failing unittests.

eli-schwartz:
Yay, the naive mostly-automated conversion flagged something that was incredibly wrong anyway. Let's merge this one first, then that one. Should be an easy(ier) rebase.

xclaesse:
YAY! totally +1 for using fstring everywhere! 

Is it possible to have a commit that only change there string format? I'm a bit scared that in the middle of that diff there could be changes that we want to review.

xclaesse:
Oh, it's already like that because you do a commit with --keep-percent-format before. Neat.

eli-schwartz:
Indeed, I had the same concern. The f-strings were the bulkiest change.

xclaesse:
Yay, thanks!

PR #8476, opened 2021-03-04
compilers: Use EnvironmentException not EnvironmentError

dcbaker:
The latter is a python built-in exception, the former is a meson
exception.


xclaesse:
LGTM. I think things needs to go before the upgrade to py3.6.

xclaesse:
Well, can go after as long as we don't forget to do it. Was just a bit afraid that once renamed to OSError we won't be able too grep then easily.

eli-schwartz:
Note to self: rebasing the PR does not mean accepting all merge values of "theirs" :roll_eyes:

eli-schwartz:
> Was just a bit afraid that once renamed to OSError we won't be able too grep then easily.

It's technically easy to do :p you just cherry-pick the PR onto the other merged PR and resolve the differences.

PR #8477, opened 2021-03-05
dependency: Add JDK system dependency

tristan957:
The JDK system dependency is important for detecting JDK include paths
that may be useful when developing a JNI interface.


tristan957:
Sorry about the brand new PR. My git skills were trash the other day.

CC: @xclaesse @dcbaker since you all were on the last PR. This time I have a full e2e test case :).

I am not sure of the best way to go about the `find_program('java')`, so that the java that is found is in the same directory as the `javac`. Grounds for a Java module? Or do something stupid like `find_program('java', dirs: [fs.parent(javac.cmd_array()[0]))`?

tristan957:
```
______________________ NativeFileTests.test_java_compiler ______________________
[gw0] linux -- Python 3.8.7 /usr/bin/python3

self = <run_unittests.NativeFileTests testMethod=test_java_compiler>

    @skip_if_not_language('java')
    def test_java_compiler(self):
>       self._single_implementation_compiler(
            'java', 'javac', 'javac 9.99.77', '9.99.77')

run_unittests.py:8716: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
run_unittests.py:8700: in _single_implementation_compiler
    self.assertEqual(compiler.version, version)
E   AssertionError: '15.0.2' != '9.99.77'
E   - 15.0.2
E   + 9.99.77
=========================== short test summary info ============================
FAILED run_unittests.py::NativeFileTests::test_java_compiler - AssertionError...
```

tristan957:
Not super sure what is up with the test. Obviously the wrapper binary didn't get caught and detect function picked out the javac from the environment. Good news is that the `jdk` test passed :).

jpakkane:
Did some debugging and the issue seems to be here.

```python
    def detect_java_compiler(self, for_machine):
        exelist = self.lookup_binary_entry(for_machine, 'javac')
        if exelist is None:
            java_home = self.properties[for_machine].get_java_home()
            if java_home is not None:
                exelist = [str(pathlib.Path(java_home) / 'bin' / 'javac')]
        info = self.machines[for_machine]
        if exelist is None:
            exelist = [str(pathlib.Path(shutil.which(self.default_java[0])))]
```

Both `lookup_binary_entry` and `self.properties` return `None` and then on the last line `exelist` is looked up with the value of `default_java[0]`. Dunno why that is.

mesonbuild/dependencies/dev.py
```diff
@@ -505,6 +508,59 @@ def get_methods():
...
+        When inspecting a JDK release tarball or $JAVA_HOME, inside the `include/` directory is a
+        platform dependent folder that must be on the target's include path in addition to the
+        parent `include/` directory.
+        """
+        if m.is_linux():
+            return 'linux'
+        elif m.is_windows():
+            return 'win32'
+        elif m.is_darwin():
+            return 'darwin'
```
dcbaker, line 559: we should have something like:
```python
mlog.log("Could not find a JDK platform include directory for your OS, please open an issue or provide a merge request.")
```
if we hit the return None case.

tristan957: great idea.

tristan957: Gonna use mlog.error though since in theory this is a hard error.

mesonbuild/envconfig.py
```diff
@@ -210,6 +211,14 @@ def get_cmake_use_exe_wrapper(self) -> bool:
         assert isinstance(res, bool)
         return res
 
+    def get_java_home(self) -> T.Optional[Path]:
+        value = self.properties.get('java_home')
+        if value is None:
+            value = os.environ.get('JAVA_HOME')
+            return Path(value) if value else None
```
tristan957, line 218: Is this uncessary? It looks like from the comment that java_home is already set from the environment before this.

test cases/java/9 jdk/meson.build
```diff
@@ -0,0 +1,10 @@
+project('jdkjava', ['c', 'java'])
+
+fs = import('fs')
+
+java = find_program('java', required: true)
```
tristan957, line 5: Not really sure what to about this since the Java executable that will be found won't necessarily be the one in JAVA_HOME/bin

tristan957:
I also don't understand the test failures

test cases/java/9 jdk/src/meson.build
```diff
@@ -0,0 +1,17 @@
...
+  main_class : 'com.mesonbuild.JdkTest',
+)
+
+test(
+  'jdktest',
+  java,
+  args: [
+    '-Djava.library.path=@0@'.format(fs.parent(jdkjava.full_path())),
+    '-jar',
+    jdkjar,
```
tristan957, line 13: Does this jdkjar not create an implict dependency on the target and that is why the test is failing?

dcbaker: It looks like your building a 32bit .dll and running witha  64bit java

tristan957: Can I force a 64 bit build of the .dll? I don't do Windows development.

dcbaker: That image is a 32bit image, so we should be able to get a 32bit java or we need to skip running the test as it's not going to work.

dcbaker:
I haven't tested this, but try adding:
```yaml
  - task: JavaToolInstaler@0
    inputs:
        jdkArchitectureOption: $(arch)
```
right after the `-task: UsePythonVersion` in the `azure-pipelines.yml` and see if that helps.

tristan957:
@dcbaker thanks I will try this. Hopefully works first try and this can be merged

test cases/java/9 jdk/meson.build
tristan957: I think that this is best served by being fixed in a java module which I have been working on

tristan957:
Python can suck it. Typing should have been a part of the language from the beginning. Mypy....................

tristan957:
The error makes no sense because the build images say they come with Java 1.8, 11, and 13

tristan957:
They literally could not have made Azure Pipelines any crappier

tristan957:
@jpakkane do we have an `@skipIfWindows` decorator?

dcbaker:
```python
@unittest.skipIf(mesonlib.is_windows())
```
IIRC

docs/markdown/snippets/jdk-system-dependency.md
```diff
@@ -0,0 +1,16 @@
...
+
+When building projects such as those interacting with the JNI, you need access
+to a few header files located in a Java installation. This system dependency
+will add the correct include paths to your target. It assumes that either
+`JAVA_HOME` will be set to a valid Java installation, or the default `javac` on
+your system is a located in the `bin` directory of a Java installation. Note:
+symlinks are resolved.
+
+```meson
+jdk = dependency('jdk', version : '>=1.8', required : true)
```
jpakkane, line 11: In case this needs any further changes, take out `required: true`. It's not needed as it is the default. However if this would be the only thing that would need changing then it's fine as is.

tristan957: I will push it up real quick.

tristan957:
@jpakkane should I be using host_machine or build_machine in my test case?

test cases/java/9 jdk/meson.build
```diff
@@ -0,0 +1,14 @@
+project('jdkjava', ['c', 'java'])
+
+if build_machine.system() == 'windows' and build_machine.cpu_family() == 'x86'
+  error('MESON_SKIP_TEST: failing builds on 32bit Windows because a 32bit JDK isn't available in the Azure Pipelines Windows images')
```
dcbaker, line 4: I think you need to escape the `'` in `isn't`

dcbaker:
I think `build_machine` is correct in this case. It's hard to say since you're not cross compiling.

mesonbuild/dependencies/dev.py
dcbaker: Looks like this isn't resolved yet?

test cases/java/9 jdk/meson.build
tristan957: Fixed.

mesonbuild/dependencies/dev.py
tristan957: Look at line 533.

jpakkane:
Skipping Cygwin is also ok, sadly a lot of things on it are broken.

tristan957:
The macos issues seem related to homebrew. Would like to get this merged now.

dcbaker:
I kicked them, see if they work this time. If not I'm happy to ignore them

Issue #8480, opened 2021-03-06
UnicodeDecodeError after "Running custom install script..."


amezin:
**Describe the bug**
On Windows, with non-English locale:

```
> git clone https://gitlab.gnome.org/GNOME/gtk/
> cd gtk
> mkdir build
> cd build
> meson
> ninja
> ninja install
...
Running custom install script 'c:\\users\\aleksandr\\appdata\\local\\programs\\python\\python39\\python.exe C:\\Users\\Aleksandr\\gtk\\build-aux/meson/post-install.py 4.0 4.0.0 C:/Users/Aleksandr/gtk/build/install/lib C:/Users/Aleksandr/gtk/build/install/share C:/Users/Aleksandr/gtk/build/install/bin'
--- stdout ---
Compiling GSettings schemas...

--- stderr ---
Traceback (most recent call last):
  File "c:\users\aleksandr\appdata\local\programs\python\python39\lib\site-packages\mesonbuild\mesonmain.py", line 132, in run
    return options.run_func(options)
  File "c:\users\aleksandr\appdata\local\programs\python\python39\lib\site-packages\mesonbuild\minstall.py", line 702, in run
    installer.do_install(datafilename)
  File "c:\users\aleksandr\appdata\local\programs\python\python39\lib\site-packages\mesonbuild\minstall.py", line 508, in do_install
    self.run_install_script(d, destdir, fullprefix)
  File "c:\users\aleksandr\appdata\local\programs\python\python39\lib\site-packages\mesonbuild\minstall.py", line 580, in run_install_script
    rc = self.run_exe(i, env)
  File "c:\users\aleksandr\appdata\local\programs\python\python39\lib\site-packages\mesonbuild\minstall.py", line 352, in run_exe
    return run_exe(*args, **kwargs)
  File "c:\users\aleksandr\appdata\local\programs\python\python39\lib\site-packages\mesonbuild\scripts\meson_exe.py", line 77, in run_exe
    print(stderr.decode())
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xcd in position 914: invalid continuation byte
FAILED: meson-install
"C:\Users\Aleksandr\AppData\Local\Programs\Python\Python39\Scripts\meson" "install" "--no-rebuild"
ninja: build stopped: subcommand failed.
```

**Expected behavior**

I expected to see output of the failed script, not stack trace from Meson itself

```
[...]

PR #8481, opened 2021-03-06
scripts/meson_exe: fix stdout and stderr decoding

amezin:
1. use `locale.getpreferredencoding()` to get encoding name.

`bytes.decode()` assumes `encoding='utf-8'` by default. It is incorrect on my
Windows setup, and causes `UnicodeDecodeError`.

2. use `errors='replace'`.

`bytes.decode()` assumes `errors='strict'` by default. Meson shouldn't crash
if subprocess outputs some garbage that can't be decoded.

`surrogateescape` doesn't work as expected on Windows. On Linux, default
`errors` for `sys.stdout` is `strict`, so `surrogateescape` can't be used there
too (at least until `sys.stdout` is reconfigured).

Fixes https://github.com/mesonbuild/meson/issues/8480

### DIFF
--- a/mesonbuild/scripts/meson_exe.py
+++ b/mesonbuild/scripts/meson_exe.py
@@ -18,6 +18,7 @@ import argparse
 import pickle
 import subprocess
 import typing as T
+import locale
 
 from .. import mesonlib
 from ..backend.backends import ExecutableSerialisation
@@ -70,11 +71,12 @@ def run_exe(exe: ExecutableSerialisation, extra_env: T.Optional[dict] = None) ->
             print(f'while executing {cmd_args!r}')
         if exe.verbose:
             return p.returncode
+        encoding = locale.getpreferredencoding()
         if not exe.capture:
             print('--- stdout ---')
-            print(stdout.decode())
+            print(stdout.decode(encoding=encoding, errors='replace'))
         print('--- stderr ---')
-        print(stderr.decode())
+        print(stderr.decode(encoding=encoding, errors='replace'))
         return p.returncode
 
     if exe.capture:


jpakkane:
I looked up the docs for surrogate escape and it was not immediately obvious. What is the benefit of using that instead of plain replace?

xclaesse:
Shouldn't we use `Popen_safe()` and print with `mlog.log()` instead, we deal with encoding already there.

amezin:
> I looked up the docs for surrogate escape and it was not immediately obvious. What is the benefit of using that instead of plain replace?

```
>>> data = b'\xfa\xff'
>>> decoded = data.decode(errors='surrogateescape')
>>> encoded = decoded.encode(errors='surrogateescape')
>>> encoded == data
True
```

```
>>> data = b'\xfa\xff'
>>> decoded = data.decode(errors='replace')
>>> encoded = decoded.encode(errors='replace')
>>> encoded == data
False
>>> encoded
b'\xef\xbf\xbd\xef\xbf\xbd'
```

So if output from subprocess contains some invalid text, and it is decoded with `errors='surrogateescape'`, printing it to stdout that is configured with `errors='surrogateescape'` should output the same byte sequence. As far as I understand.

However, `meson_exe` doesn't set `errors` for stdout. Oops. And even after I've added `sys.stdout.reconfigure(errors='surrogateescape')`, it still doesn't work the way I expected...

jpakkane:
Maybe a better solution would be to not even try to decode anything, but instead print raw bytes [as shown on SO](https://stackoverflow.com/questions/908331/how-to-write-binary-data-to-stdout-in-python-3)?

amezin:
> Maybe a better solution would be to not even try to decode anything, but instead print raw bytes [as shown on SO](https://stackoverflow.com/questions/908331/how-to-write-binary-data-to-stdout-in-python-3)?

Tested it and it doesn't show localized error messages correctly on Windows. So that's why `surrogateescape` didn't work too - because passing bytes as is doesn't work. No idea why

amezin:
Changed to `errors='replace'`

amezin:
No idea why one CI config is failing

https://dev.azure.com/jussi0947/jussi/_build/results?buildId=9614&view=logs&j=5015ea34-1c35-577b-b206-c8bea8465a58&t=d96bae21-2af5-55ef-6693-add22ffd7f07

```
Total passed tests: 432
Total failed tests: 0
Total skipped tests: 198
##[error]PowerShell exited with code '1'.
```

jpakkane:
Probably this; https://twitter.com/AzureSupport/status/1371544292502564866?s=19

PR #8485, opened 2021-03-07
Add support for basic format strings

e820:
This is a little patch to add support for Python-esque format strings to the Meson language. This might be a controversial addition so I don't expect this to be merged as is, but I've purposefully limited this implementation to keep the lexer & parser simple.

The current version allows for the following (as indicated by the test case):
```meson
n = 10
s = 'a string'

message(f'n = {n}, string = {s}')
```

As I said the current version **is restricted** to identity-expressions inside of format strings (`{...}`) but if we decide that we want format strings and that they should support arbitrary expressions I'll lift that restriction.

I'm in favor of this addition as
- Format strings are intuitive, especially for people coming from Python and other languages offering them
- The current syntax is unnecessarily noisy and cumbersome to write

### DIFF
--- a/docs/markdown/Syntax.md
+++ b/docs/markdown/Syntax.md
@@ -170,6 +170,8 @@ functionality described below.
 
 ### String formatting
 
+#### .format()
+
 Strings can be built using the string formatting functionality.
 
 ```meson
@@ -181,6 +183,7 @@ res = template.format('text', 1, true)
 As can be seen, the formatting works by replacing placeholders of type
 `@number@` with the corresponding argument.
 
+#### Format strings
 *(Added 0.58)*
 
 Format strings can be used as a non-positional alternative to the
--- a/docs/markdown/Syntax.md
+++ b/docs/markdown/Syntax.md
@@ -198,7 +198,15 @@ s = f'int: @n@, string: @m@'
 ```
 
 Currently only identity-expressions are supported inside of format
-strings.
+strings, meaning you cannot use arbitrary Meson expressions inside of them.
+
+```meson
+n = 10
+m = 5
+
+# The following is not a valid format string
+s = f'result: @n + m@'
+```
 
 ### String methods
 
--- a/docs/markdown/Syntax.md
+++ b/docs/markdown/Syntax.md
@@ -181,6 +181,22 @@ res = template.format('text', 1, true)
 As can be seen, the formatting works by replacing placeholders of type
 `@number@` with the corresponding argument.
 
+*(Added 0.58)*
+
+Format strings can be used as a non-positional alternative to the
+string formatting functionality described above.
+
+```meson
+n = 10
+m = 'hi'
+
+s = f'int: @n@, string: @m@'
+# s now has the value 'int: 10, string: hi'
+```
+
+Currently only identity-expressions are supported inside of format
+strings.
+
 ### String methods
 
 Strings also support a number of other methods that return transformed
--- a/docs/markdown/snippets/fstrings.md
+++ b/docs/markdown/snippets/fstrings.md
@@ -1,0 +1,7 @@
+## Introducing format strings to the Meson language
+
+In addition to the conventional `'A string @0@ to be formatted @1@'.format(n, m)`
+method of formatting strings in the Meson DSL, there's now the additional
+`f'A string @n@ to be formatted @m@'` notation that provides a non-positional
+and clearer alternative. Meson's format strings are currently restricted to
+identity-expressions, meaning `f'format {'m' + 'e'}'` will not parse.
--- a/docs/markdown/snippets/fstrings.md
+++ b/docs/markdown/snippets/fstrings.md
@@ -1,7 +1,7 @@
 ## Introducing format strings to the Meson language
 
 In addition to the conventional `'A string @0@ to be formatted @1@'.format(n, m)`
-method of formatting strings in the Meson DSL, there's now the additional
+method of formatting strings in the Meson language, there's now the additional
 `f'A string @n@ to be formatted @m@'` notation that provides a non-positional
 and clearer alternative. Meson's format strings are currently restricted to
 identity-expressions, meaning `f'format @'m' + 'e'@'` will not parse.
--- a/docs/markdown/snippets/fstrings.md
+++ b/docs/markdown/snippets/fstrings.md
@@ -4,4 +4,4 @@ In addition to the conventional `'A string @0@ to be formatted @1@'.format(n, m)
 method of formatting strings in the Meson DSL, there's now the additional
 `f'A string @n@ to be formatted @m@'` notation that provides a non-positional
 and clearer alternative. Meson's format strings are currently restricted to
-identity-expressions, meaning `f'format {'m' + 'e'}'` will not parse.
+identity-expressions, meaning `f'format @'m' + 'e'@'` will not parse.
--- a/mesonbuild/interpreterbase.py
+++ b/mesonbuild/interpreterbase.py
@@ -928,6 +928,7 @@ The result of this is undefined and will become a hard error in a future Meson r
         else:
             return self.evaluate_statement(node.falseblock)
 
+    @FeatureNew('format strings', '0.58.0')
     def evaluate_fstring(self, node: mparser.FormatStringNode) -> TYPE_var:
         assert(isinstance(node, mparser.FormatStringNode))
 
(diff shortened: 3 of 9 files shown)


test cases/common/238 fstrings/meson.build
```diff
@@ -0,0 +1,9 @@
+project('meson-test', 'c')
+
+n = 10
+m = 'bar'
+s = f'test {n} string ({n}): {m}'
+
+if s != 'test 10 string (10): bar'
+  error('Incorrect string formatting')
+endif
```
xclaesse, line 9: You can use assert() instead.

mesonbuild/interpreterbase.py
```diff
@@ -926,6 +928,18 @@ def evaluate_ternary(self, node: mparser.TernaryNode) -> TYPE_var:
...
+        assert(isinstance(node, mparser.FormatStringNode))
+
+        def replace(match: T.Match[str]) -> str:
+            var = str(match.group(1))
+            try:
+                return str(self.variables[var])
+            except KeyError:
+                raise mesonlib.MesonException(f'Identifier "{var}" does not name a variable.')
+
+        return re.sub(r'{([_a-zA-Z][_0-9a-zA-Z]*)}', replace, node.value)
```
xclaesse, line 941: In Python we can use {{ to escape, can we do it here too? Also I'm meson format was using @0@, I think we should continue to use that instead of curly brackets for consistency.

mesonbuild/interpreterbase.py
```diff
@@ -926,6 +928,18 @@ def evaluate_ternary(self, node: mparser.TernaryNode) -> TYPE_var:
         else:
             return self.evaluate_statement(node.falseblock)
 
+    def evaluate_fstring(self, node: mparser.FormatStringNode) -> TYPE_var:
+        assert(isinstance(node, mparser.FormatStringNode))
+
+        def replace(match: T.Match[str]) -> str:
+            var = str(match.group(1))
+            try:
+                return str(self.variables[var])
```
xclaesse, line 937: We should check the type of variable, probable limit it to str, into and bool.

xclaesse:
I personally like a lot the idea, I suggested it myself on IRC not long ago. I was expecting the implementation to be much more complex, happy to see it's actually pretty simple.

test cases/common/238 fstrings/meson.build
e820: Oh yea, I keep forgetting we have assert().

mesonbuild/interpreterbase.py
e820: Alright, I'll add type checks.

mesonbuild/interpreterbase.py
e820: This regex will replace {{x}} with {value} if x is currently defined. I don't think explicit escaping is required as using characters that do not constitute an identifier (such as whitespace) within the curlies will prevent substitution.
As for the consistency point, I deliberately went for the more familiar curly brackets instead of Meson's current notation because it looks clearer to me - are we sure we want to keep the old syntax?

jpakkane: I chose not to use curly brackecs on purpose because they are used by so many other text replacement libraries as well as being a semantically meaningful character in most programming languages. The strings Meson expands often end up as inputs to other string replacements which means you have to do escaping and that gets nasty quite quickly.

e820: Alright, I'll adjust the syntax then.

eli-schwartz: If accepted, it should behave the same as the existing `@0@` style placeholders, which I'll note do not do escaping:

```
message('@@0@@.format('hi'))
```

returns `Message: @hi@`

eli-schwartz:
This should probably have tests to cover invalid output too, which you should be able to do via https://mesonbuild.com/Contributing.html#configuring-project-tests

e820:
> This should probably have tests to cover invalid output too, which you should be able to do via https://mesonbuild.com/Contributing.html#configuring-project-tests

I've never used the test configuration support, could you elaborate on what you mean? If I'm not mistaken this is merely to configure the test environment, how would I setup/capture build script variables?

mensinda:
+1 for adding this feature.

However, this PR still lacks a release note snippet and documentation. Also, at least for now, I would keep the functionality as it currently is. Adding new features when needed will be easier than maintaining a complex system that is barely used.

eli-schwartz:
test.json can parse stdout for error messages, and probably check multiple hard errors if needed by using a matrix to toggle one error on at a time via options.

mensinda:
Also, the `evaluate_fstring` should have a `FetureNew` check.

docs/markdown/Syntax.md
```diff
@@ -181,6 +183,23 @@ res = template.format('text', 1, true)
...
+n = 10
+m = 'hi'
+
+s = f'int: @n@, string: @m@'
+# s now has the value 'int: 10, string: hi'
+```
+
+Currently only identity-expressions are supported inside of format
+strings.
+
```
xclaesse, line 202: nitpicking: I personally find "non-positional" and "identity-expressions" a bit expert language. Maybe it can be reworded in a more accessible wording? Dunno, maybe it's just me :P

docs/markdown/snippets/fstrings.md
```diff
@@ -0,0 +1,7 @@
+## Introducing format strings to the Meson language
+
+In addition to the conventional `'A string @0@ to be formatted @1@'.format(n, m)`
+method of formatting strings in the Meson DSL, there's now the additional
```
xclaesse, line 4: likewise here, when you're not into programming languages, "DSL" could be a bit cryptic?

mesonbuild/interpreterbase.py
```diff
@@ -926,6 +928,22 @@ def evaluate_ternary(self, node: mparser.TernaryNode) -> TYPE_var:
...
 
+    def evaluate_fstring(self, node: mparser.FormatStringNode) -> TYPE_var:
+        assert(isinstance(node, mparser.FormatStringNode))
+
+        def replace(match: T.Match[str]) -> str:
+            var = str(match.group(1))
+            try:
+                val = self.variables[var]
+                if not isinstance(val, (str, int, float, bool)):
+                    raise InvalidCode(f'Identifier "{var}" does not name a formattable variable.')
```
xclaesse, line 940: Maybe add here that it must be one of string, integer, float or boolean, to give a hint to user what they can do.

xclaesse (commented):
Just small nitpicking, but I like it a lot already! IMHO that's a super nice addition to the language.

xclaesse:
Awesome, as far as I'm concerned this is ready to merge. @jpakkane any objection?

@e820 Thanks for your work!

eli-schwartz:
Why are there two test cases? Can't we do a test matrix for this and run both in one meson.build?

e820:
> Why are there two test cases? Can't we do a test matrix for this and run both in one meson.build?

I was not able to find an example that does this, I tried to switch using a project option defined in the matrix but `stdout` doesn't have a way to associate option values with output expectations as far as I can see.

eli-schwartz:
Hmph, that seems slightly awkward. Oh well.

test cases/common/238 fstrings/meson.build
```diff
@@ -0,0 +1,7 @@
+project('meson-test', 'c')
+
+n = 10
+m = 'bar'
+s = f'test @n@ string (@@n@@): @m@'
+
+assert(s == 'test 10 string (@10@): bar', 'Incorrect string formatting')
```
jpakkane, line 7: This seems a bit strange. Typically doubling of the quote character prints a single quote character and nothing else. For example in Python:

```python
>>> x = 3
>>> f'{{x}}'
'{x}'
```

eli-schwartz: I agree that's wrong, but as per previous comments in this PR, it is already the case for .format() too.

e820: That's true but this behavior is consistent with `.format()`.

e820: If we choose to adjust this behavior I'd say we should update both fstrings and `.format()`. I'm not sure if that's an acceptable amount of breakage though.

xclaesse: Doing it properly is probably harder than a simple regex if I'm not mistaken. Let's not block this PR on that, if it ever becomes an issue we should fix both fstring and .format() anyway.

PR #8487, opened 2021-03-08
Add full_path() method to files objects

### LINKED ISSUE #8433: Ability to get full path of File object
when you have this
```meson
my_files = files(['myfile1.scm', 'myfile2.scm'])
```

you should be able to do this
```meson
foreach f : my_files
  message (f.full_path())
endforeach
```

so you can reuse those paths elsewhere

tristan957:
I am missing some places where mesonlib.File is probably assumed. This is a work in progress, but it worked in my very minimal test case.

The holder pattern is quite nice. Wasn't super hard to contribute this at all.

Fixes #8433

### DIFF
--- a/docs/markdown/Reference-manual.md
+++ b/docs/markdown/Reference-manual.md
@@ -860,6 +860,8 @@ Then you can use it in `bar2` like this:
 
 Meson will then do the right thing.
 
+The returned object is a array of [`file objects`](#File-object).
+
 ### generator()
 
 ``` meson
@@ -2713,6 +2715,15 @@ library. This object has the following methods:
   type name, and methods as the object that called it. This new
   object will only inherit other attributes from its parent as
   controlled by keyword arguments.
+
+### File object
+
+*(since 0.58.0)*
+
+This object is an entry in the file array returned by [`files()`](#files). This
+object contains the following methods:
+
+- `full_path()`: returns a string of the full path to the file
 
 ### Feature option object
 
--- a/mesonbuild/interpreter.py
+++ b/mesonbuild/interpreter.py
@@ -2341,6 +2341,23 @@ permitted_kwargs = {'add_global_arguments': {'language', 'native'},
                     }
 
 
+class FileHolder(InterpreterObject, ObjectHolder[mesonlib.File]):
+    def __init__(self, file: mesonlib.File, source_dir: str, build_dir: str):
+        InterpreterObject.__init__(self)
+        ObjectHolder.__init__(self, file)
+        self.source_dir = source_dir
+        self.build_dir = build_dir
+        self.methods.update({
+            'full_path': self.full_path_method,
+        })
+
+    @FeatureNew('file.full_path', '0.58.0')
+    @noPosargs
+    @noKwargs
+    def full_path_method(self, args, kwargs):
+        return self.held_object.absolute_path(self.source_dir, self.build_dir)
+
+
 class Interpreter(InterpreterBase):
 
     def __init__(
@@ -2645,7 +2662,7 @@ class Interpreter(InterpreterBase):
     @stringArgs
     @noKwargs
     def func_files(self, node, args, kwargs):
-        return [mesonlib.File.from_source_file(self.environment.source_dir, self.subdir, fname) for fname in args]
+        return [FileHolder(mesonlib.File.from_source_file(self.environment.source_dir, self.subdir, fname), self.environment.source_dir, self.environment.build_dir) for fname in args]
 
     # Used by declare_dependency() and pkgconfig.generate()
     def extract_variables(self, kwargs, argname='variables', list_new=False, dict_new=False):
@@ -4790,7 +4807,7 @@ Try setting b_lundef to false instead.'''.format(self.coredata.options[OptionKey
             sources = [sources]
         results: T.List[mesonlib.File] = []
         for s in sources:
-            if isinstance(s, (mesonlib.File, GeneratedListHolder,
+            if isinstance(s, (mesonlib.File, FileHolder, GeneratedListHolder,
                               TargetHolder, CustomTargetIndexHolder,
                               GeneratedObjectsHolder)):
                 pass
--- a/docs/markdown/snippets/file_full_path.md
+++ b/docs/markdown/snippets/file_full_path.md
@@ -1,0 +1,11 @@
+## file.full_path()
+
+File objects now support a method called `full_path()`. It returns the full path
+of the file.
+
+```meson
+srcs = files('x.c', 'y.c')
+foreach s : srcs
+  message(s.full_path()) # prints the full path of each file
+endforeach
+```


tristan957:
Besides failing tests, known missing parts of PR:

* FeatureNew decorators
* snippets contribution
* docs update for new method and new type since I don't think file object are currently documented

tristan957:
@jpakkane thanks for your input. I think you are definitely correct about the documentation being super explicit about this method having very specific use-cases.

xclaesse:
> This should probably be `absolute_path` instead as it is unambiguous. It is also what the Python function is called, probably for this exact reason.

I understand your concern but unfortunately all other meson objects already have `.full_path()` (e.g. BuildTarget, CustomTarget, ExternalProgram). I think in this case consistency is better, or we should deprecate them all and replace by `.absolute_path()` but IMHO that would annoy all our users for little gain.

xclaesse:
@tristan957 I think this is the wrong approach, why not add support for File object in `message()` instead? Wrapping files into a FileHolder, while making general sense, is going to cause countless regressions because we have isinstance() all over the place.

jpakkane:
If that is the only thing this would be used for, then definitely.

tristan957:
I was thinking that it could make sense to model files as FileHolders from an interpreter perspective, and this could be a good PR to possibly match how other interpreter object are also modeled as Holders.

I am honestly struggling to come up with use cases for this method, and people who care about this should propose their use cases. I thought I had one while redoing the build system on my project, but I seem to have gotten away without needing it.

Re full_path() vs absolute_path(): I think coming up with internal consistency should be prio #1. Once that is done, if absolute_path() is the way to go, then I don't think the churn is that big a deal because Meson is extremely consistent in its usage of full_path(), so people could just s/full_path/absolute_path/ in a single commit.

xclaesse:
FileHolder makes a lot of sense, yes. But the question is whether it's worth the pain. The only use case I see so far can trivially be done without it.

tristan957:
I am just going to put the PR on hold until people come up with good use cases.

jpakkane:
>  I think this is the wrong approach, why not add support for File object in message() instead?

If this is something you need, then it should be added regardless of what happens to this PR. I

tristan957:
That is not something that I need currently. If other people do, then I can add it.

jpakkane:
What _did_ you need this for, then?

In Meson we purposefully avoid adding features just because they seem useful. Every feature must have an actual use case or requirement behind it.

tristan957:
There was an issue for it and I wanted to fix it for other people.

randy408:
I have a project where files can be reloaded at runtime when debugging, the files are also part of the build so it just makes sense to generate the asset/source<->fullpath associations from meson for change monitoring and reloading.

tristan957:
That seems like a reasonable request pending maintainer approval.

@jpakkane

mesonbuild/interpreter.py
```diff
@@ -2645,7 +2662,7 @@ def func_import(self, node, args, kwargs):
     @stringArgs
     @noKwargs
     def func_files(self, node, args, kwargs):
-        return [mesonlib.File.from_source_file(self.environment.source_dir, self.subdir, fname) for fname in args]
+        return [FileHolder(mesonlib.File.from_source_file(self.environment.source_dir, self.subdir, fname), self.environment.source_dir, self.environment.build_dir) for fname in args]
```
xclaesse, line 2665: Unfortunately it is not that easy, there are many places where we do `isinstance(f, mesonlib.File)`, and they won't work any more because the type is now `FileHolder`.

xclaesse: So basically, in every API that takes file objects you now have to unholder() them first in interpreter.

tristan957: @xclaesse I am willing to put in the work to create `FileHolder` if people think it is an improvement in the interpreter code.

jpakkane: `interpreterbase.py` has a `method_call` method that deals with method calls to non-interpreterobjects.

docs/markdown/Reference-manual.md
```diff
@@ -2714,6 +2716,15 @@ library. This object has the following methods:
...
   controlled by keyword arguments.
 
+### File object
+
+*(since 0.58.0)*
+
+This object is an entry in the file array returned by [`files()`](#files). This
+object contains the following methods:
+
+- `full_path()`: returns a string of the full path to the file
```
jpakkane, line 2726: This should have a warning of something like: "Use this method with caution. The return value is a string and carries no dependency information. Thus if you use it as an argument to a command, Meson does not know that it refers to an actual file and can not set up dependencies properly. Whenever possible you should use the file object directly in the command line. It will be autoexpanded to point to the file and all necessary dependencies are set up automatically."

gerion0:
> There was an issue for it and I wanted to fix it for other people.

Thank you for the effort!

A small example that I need that function for currently:
I have a resource file (`res.txt`) and a script that uses that file as part of a test case (`test.py`).
Both files are _not_ in the same folder. I want to communicate the path of `res.txt` via an environment variable to `test.py` (by argument is not simply possible due to various reasons).

This results in these meson files:
somewhere/meson.build:
```
resource_file = files('res.txt')[0]
```

elsewhere/meson.build:
```
py_mod = import('python')
py_inst = py_mod.find_installation('python3')

# does not work currently
res_file = 'RESOURCE_FILE=' + resource_file.full_path()

test('python-test',
     py3_inst,
     args: [files('test.py')],
     env: [res_file]
)
```

Note, that I was not fully able to find a workaround. This here is probably the best:
```
res_file = 'RESOURCE_FILE=@0@/@1@'.format(meson.source_root(), resource_file)
```
Putting a file into the format function seems to result in the path relative to the source_root but I'm not sure at all about this. How it is in subprojects?
Also, using `/` as a path separator is not OS agnostic (although I'm not sure if using the environment is possible on all systems).

tristan957:
You can set the workdir for the test. Does that help you?

gerion0:
Hmm, nothing other in the script depends on the working directory so it would be a possible workaround, yes.
However, I think that a `full_path()` function is cleaner since it gives Meson the full control over the file location.

gerion0:
BTW, in the above code the dependency to the resource file is gone for the test target (like @jpakkane already mentioned). It does not seem simple, though, to add it back. At least, the `depends` kwarg of `test()` only take targets, not files.

anarazel:
@jpakkane wrote:
> What _did_ you need this for, then?
> 
> In Meson we purposefully avoid adding features just because they seem useful. Every feature must have an actual use case or requirement behind it.

@tristan957 wrote:
> There was an issue for it and I wanted to fix it for other people.

FWIW, I've needed this in postgres' meson conversion, for an error check during configure. We need to make sure that there are no in-tree files left over from configure, as they'll sometimes get "preferred" (due to `#include "path"` always looking first in the source file's directory). I got away with '@0@'.format(configure_file_return), but that's just this PR in worse.  Before the check there were repeated issues due to outdated files being left over and at some point later failing in weird ways.

tristan957:
I am happy to work on this again if it would be accepted by @jpakkane. Don't want to put in effort for it to languish again.

bruchar1:
Fixed by #12408.

PR #8488, opened 2021-03-08
Remove redundant lib prefix if the target name already includes it

jaos:
Remove the surprise of

libfoo = library('libfoo', lib_sources, dependencies : deps, install: true, version: version, soversion: 0)

yielding 'liblibfoo.so' instead of 'libfoo.so'.

### DIFF
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -1796,7 +1796,7 @@ class StaticLibrary(BuildTarget):
         # the import library. Using libfoo.a is ok because people using MSVC
         # always pass the library filename while linking anyway.
         if not hasattr(self, 'prefix'):
-            self.prefix = 'lib'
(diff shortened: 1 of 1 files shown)


jpakkane:
This would mean that you can't have a Windows dll named `libsomething.dll`. Unfortunately it is quite likely that someone out there requires that.

eli-schwartz:
What is the purpose of this PR? If it is redundant, remove it from meson.build; in cases where it's not redundant, why take that choice away from users?

jon-turney:
This also makes in impossible to make a library called e.g `liblibreoffice`?

jaos:
Haha fair enough.

jpakkane:
Since this can't be done reliably, I'm going to close it. Thanks.

PR #8489, opened 2021-03-08
Add str.replace() method

### LINKED ISSUE #8459: no string.replace method
Currently the only way to perform an operation similar to pythons `str.replace(old, new)` is with `new.join(str.split(old))` which feels clunky. A `str.replace(old, new)` method would be a welcome addition.

tristan957:
Fixes #8459

### DIFF
--- a/docs/markdown/Reference-manual.md
+++ b/docs/markdown/Reference-manual.md
@@ -2134,6 +2134,9 @@ are immutable, all operations return their results as a new string.
 - `join(list_of_strings)`: the opposite of split, for example
   `'.'.join(['a', 'b', 'c']` yields `'a.b.c'`.
 
+- `replace('old_substr', 'new_str')`: replaces instances of `old_substr` in the
+  string with `new_str` and returns a new string
+
 - `split(split_character)`: splits the string at the specified
   character (or whitespace if not set) and returns the parts in an
   array.
--- a/docs/markdown/Syntax.md
+++ b/docs/markdown/Syntax.md
@@ -186,6 +186,17 @@ As can be seen, the formatting works by replacing placeholders of type
 Strings also support a number of other methods that return transformed
 copies.
 
+#### .replace()
+
+Since 0.58.0, you can replace a substring from a string.
+
+```meson
+# Replaces all instances of one substring with another
+s = 'semicolons;as;separators'
+s = s.replace('as', 'are')
+# 's' now has the value of 'semicolons;are;separators'
+```
+
 #### .strip()
 
 ```meson
--- a/docs/markdown/snippets/string_replace.md
+++ b/docs/markdown/snippets/string_replace.md
@@ -1,0 +1,10 @@
+## String `.replace()`
+
+String objects now have a method called replace for replacing all instances of a
+substring in a string with another.
+
+```meson
+s = 'aaabbb'
+s = s.replace('aaa', 'bbb')
+# 's' is now 'bbbbbb'
+```
--- a/mesonbuild/interpreterbase.py
+++ b/mesonbuild/interpreterbase.py
@@ -1192,6 +1192,13 @@ The result of this is undefined and will become a hard error in a future Meson r
                     raise InterpreterException('substring() argument must be an int')
                 end = posargs[1]
             return obj[start:end]
+        elif method_name == 'replace':
+            FeatureNew.single_use('str.replace', '0.58.0', self.subproject)
+            if len(posargs) != 2:
+                raise InterpreterException('replace() takes exactly two arguments.')
+            if not isinstance(posargs[0], str) or not isinstance(posargs[1], str):
+                raise InterpreterException('replace() requires that both arguments be strings')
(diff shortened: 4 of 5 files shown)


mensinda:
The python [`replace`](https://docs.python.org/3.6/library/stdtypes.html#str.replace) method works with substrings, not chars. So the current documentation is misleading (unless the intention was to only allow chars). There should also be an explicit test for replacing substrings.

This also needs a `FetureNew` check.

tristan957:
Ah. You are right. I will fix that.

docs/markdown/Reference-manual.md
```diff
@@ -2134,6 +2134,9 @@ are immutable, all operations return their results as a new string.
 - `join(list_of_strings)`: the opposite of split, for example
   `'.'.join(['a', 'b', 'c']` yields `'a.b.c'`.
 
+- `replace(old_char, new_char)`: replaces instances of `old_char` in the string
+  with `new_char`
```
xclaesse, line 2138: We should make it clear that it's not an inplace replace, but return a new string.

docs/markdown/Syntax.md
```diff
@@ -186,6 +186,17 @@ As can be seen, the formatting works by replacing placeholders of type
 Strings also support a number of other methods that return transformed
 copies.
 
+#### .replace()
+
+Since 0.58.0, you can now replace characters in a string
+
+```meson
+# Similar to the Python str.replace(). Replaces instances of one character with another
```
xclaesse, line 194: I think we voluntarily avoid documenting stuff as reference to python, because it's really an implementation detail.

docs/markdown/Reference-manual.md
xclaesse: I would also specify "replaces **all** instances".

docs/markdown/Syntax.md
xclaesse: I would also specify "replaces **all** instances". (I hate how this doc is duplicated)

docs/markdown/Syntax.md
```diff
@@ -186,6 +186,17 @@ As can be seen, the formatting works by replacing placeholders of type
 Strings also support a number of other methods that return transformed
 copies.
 
+#### .replace()
+
+Since 0.58.0, you can now replace characters in a string.
```
xclaesse, line 191:
```suggestion
Since 0.58.0, you can replace a substring from a string.
```
To make it clear it's not replacing characters, but substring.

docs/markdown/Syntax.md
```diff
@@ -186,6 +186,17 @@ As can be seen, the formatting works by replacing placeholders of type
...
+
+```meson
+# Replaces all instances of one substring with another
+s = 'semicolons;as;separators'
+s = s.replace(';', ' ')
```
xclaesse, line 196: I would rather use an example that makes it cleared it's replacing a substring and not characters.

docs/markdown/snippets/string_replace.md
```diff
@@ -0,0 +1,10 @@
+## String `.replace()`
+
+String objects now have a method called replace for replacing all instances of a
+substring in a string with another.
+
+```meson
+s = 'aaa'
+s = s.replace('a', 'b')
```
xclaesse, line 8: Same here.

test cases/common/36 string operations/meson.build
```diff
@@ -19,6 +19,9 @@ long = 'abcde'
 prefix = 'abc'
 suffix = 'cde'
 
+assert(long.replace('b', 'd') == 'adcde')
+assert(long.replace('z', 'x') == long)
```
xclaesse, line 23: Add a test with more than one letter replacement?

xclaesse (commented):
Just really small nitpicking about emphasing it's about substring and not just characters.

Other than that, looks really good, thanks.

tristan957:
I just missed a couple spots rewording the docs and a new test should add a substring replace.

xclaesse:
LGTM when CI is green

mesonbuild/interpreterbase.py
```diff
@@ -1192,6 +1192,12 @@ def string_method_call(self, obj: str, method_name: str, posargs: T.List[TYPE_nv
                     raise InterpreterException('substring() argument must be an int')
                 end = posargs[1]
             return obj[start:end]
+        elif method_name == 'replace':
+            if len(posargs) != 2:
+                raise InterpreterException('replace() takes exactly two arguments.')
```
xclaesse, line 1197: Oh wait, missing FeatureNew here.

tristan957: How do you add it when it's not a function decorator? I didn't know

eli-schwartz: `FeatureNew.single_use()`

tristan957: Thanks guys. Got it in. Seeing if tests are green :green_circle:

PR #8491, opened 2021-03-08
Fix D library search path translation

rtbo:
Library search paths issued by pkg-config on Windows not properly supported.
A search path that pkg-config issues like `-LC:\SomeDir\lib` should be translated as `-L=-LC:\SomeDir\lib`, but in this case the `-L` is not doubled and the linker complains about `C:\SomeDir\lib.obj` not existing.

This PR attemps to fix the logic behind the search path translation.

Example of pkg-config file that can trigger the issue:
```pc
prefix=C:\SomeDir
includedir=${prefix}/include
libdir=${prefix}/lib

Name: testlib
Version: 1.0.0
Description: Test library
Cflags: -I${includedir}/d
Libs: -L${libdir} -ltestlib
```


mesonbuild/compilers/d.py
```diff
@@ -332,7 +332,7 @@ def _translate_args_to_nongnu(self, args: T.List[str]) -> T.List[str]:
                 else:
                     dcargs.append('-I' + arg[10:])
                 continue
-            elif arg.startswith('-L/') or arg.startswith('-L./'):
+            elif arg.startswith('-L'):
```
rtbo, line 335: A search path can start with anything, not only `/` and `./`. On Windows, it cannot start with `/`.

mesonbuild/compilers/d.py
```diff
@@ -341,14 +341,25 @@ def _translate_args_to_nongnu(self, args: T.List[str]) -> T.List[str]:
...
+
+                # The user has passed a linker flag directly for the D compiler
+                if suffix.startswith('-L'):
+                    dcargs.append(arg)
+                    continue
+
+                # linker flag such as -L=/DEBUG must pass through
+                if self.linker.id == 'link' and self.info.is_windows() and suffix.startswith('/'):
+                    dcargs.append(arg)
+                    continue
```
rtbo, line 357: This function actually receives flag such as `-L=/DEBUG` `-L=/SUBSYSTEM:CONSOLE` etc.
They are correctly passed to the linker only because of the '=' sign, not because the logic is correct.

mesonbuild/compilers/d.py
```diff
@@ -341,14 +341,25 @@ def _translate_args_to_nongnu(self, args: T.List[str]) -> T.List[str]:
...
 
+                if arg.startswith('-L='):
+                    suffix = arg[3:]
+                else:
+                    suffix = arg[2:]
+
+                # The user has passed a linker flag directly for the D compiler
+                if suffix.startswith('-L'):
+                    dcargs.append(arg)
+                    continue
```
rtbo, line 352: A user providing library search path to Meson should pass them in the form `-L=-L/some/path`, right?

dcbaker:
Can I be super honest that I really dislike the fact that we keep gcc style arguments and try to translate them into other things? I'd really like to have an abstraction for this and then have each compiler reduce the abstraction to it's concrete arguments in the backend, specifically because of these sort of cases. I'm not expecting you to fix this, but it would be so much nicer to have:
```python
[LinkSearchPath('/foo/bar'), Link('foo', type=LinkType.SHARED)]
```
and then have the Compiler lower that... I might just go work on that now.

mesonbuild/compilers/d.py
dcbaker: For D? yes.

rtbo:
The test failure seems unrelated to this change

jpakkane:
@dcbaker what do you want to do with this then?

dcbaker:
@jpakkane my though tis to have a set of abstract objects like `LinkerPath('/some/path')`, `Link('name_or_path'), `Include('some path')`, (maybe a few more), and `RawArg('some_arg')`, which we convert specific arguments to when we receive internally (like from cmake or pkg-config), then each compiler would have a `to_raw_args(args : T.List[AbstractArgument]) -> T.List[str]`, internally we always pass around lists of these `AbstractArguments`, then the backend calles the compiler method to lower them into the raw arguments right before writing them out.

I started implementing this, I can push a branch if you want to look at it, though it's still very much a WIP. There's a bunch more work to do, especially since I realized I need to move some stuff around. When I started it looked like we could probably simplify or remove the `CompilerArgs` class, since seeing patterns like `[Link('foo.so'), 'Link('foo.so')]` are the same is easy, and taking things like `[LinkPath('/usr/lib/foo/'), Link('bar', static=True)]` is easy to convert to `[Link('/usr/lib/foo/bar.a')]`, which is something we currently do in the pkg-config dependency, but not in other cases.

PR #8495, opened 2021-03-08
Split program representations out of dependencies

dcbaker:
This was all originally part of my find_program refactor work. That's kinda stalled and I think if I get back to it it'll be a start from scratch situation. Therefor, I've rebased this and split it out as it seems to be a nice cleanup and someone (mensidna?) suggested that it was worth landing in and of itself.


dcbaker:
That warning is silly, it's always existed, I've just moved it from one pace to another.

xclaesse:
I like the idea of splitting large files, we should do that more. LGTM.

PR #8496, opened 2021-03-09
remove unused environment function

eli-schwartz:
The script dir is never really used since meson --internal handles this. The last remaining use of the raw script dir got removed in commit 522392e7553823e6b3ac38cadc4fbee72eae9540.


xclaesse:
LGTM when CI is green.

PR #8497, opened 2021-03-09
setuptools: don't hardcode list of modules to install, use find_packages

eli-schwartz:
And don't run a pointless test to verify that the hardcoded list has been manually maintained correctly. The same test rules used there can translate directly to find_packages pattern rules.

...

Why do we need this either? find_packages() on its own behaves completely correctly, passing options to it is more explicit but does not actually change anything.

### DIFF
--- a/run_meson_command_tests.py
+++ b/run_meson_command_tests.py
@@ -136,21 +136,6 @@ class CommandTests(unittest.TestCase):
         # Check that all the files were installed correctly
         self.assertTrue(bindir.is_dir())
         self.assertTrue(pylibdir.is_dir())
-        from setup import packages
-        # Extract list of expected python module files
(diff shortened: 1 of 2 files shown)


eli-schwartz:
"used packages"?

As I understand it, setuptools doesn't check for "use", but checks every directory to see if it:
- is in the same directory as setup.py
- matches the include/exclude filters
- contains a `__init__.py`

If you remove the `include=['mesonbuild', 'mesonbuild.*']` then nothing changes.
But if you additionally `touch tools/__init__.py` it goes ahead and packages up the
```
from tools import ac_converter, boost_names, build_website, cmake2meson, copy_files, dircondenser, gen_data, regenerate_docs
```
modules.

PR #8499, opened 2021-03-09
stdout content is printed to console again if capture is not enabled

### LINKED ISSUE #8498: Script output is not printed to console
**Describe the bug**
The stdout content is captured to a file if `capture:true` is set in a custom_target. Otherwise it won't be outputed anymore.

**To Reproduce**
Upgrading meson from 0.56.2 to 0.57.0 or newer introduced this bug/feature? with the commit 0626465ea8aa65b10776d5c4064e881fe0d6fa25 of fixing #8187 

The behaviour seems not to happen if I do not depend on other meson objects like e.g. an executable() which compiles, in my case 45 C objects beforehand.

```
python3 = find_program('python3', required : false)
my_custom_script = custom_target('Target_custom',
    command : [python3 , '@INPUT@',],
    # capture : true, # required since 0.57.0 as output is not printed to stdout, capture to file
    output : 'custom.log',
    input : 'random.py',
    depends: [binary_executable],
    build_by_default    : false
)
alias_target('Target_custom', my_custom_script)
```

random.py file looks similar to this.
```#!/usr/bin/env python
import logging
import sys

logging.basicConfig(level=logging.DEBUG, stream=sys.stdout)
logging.debug('This will get logged')
```

Output is:
`[45/45] Generating Target_custom with a custom command (wrapped by meson to set PATH)`

**Expected behavior**
`[45/45] Generating TargetA_custom with a custom command (wrapped by meson to set PATH)
DEBUG:root:This will get logged`

**system parameters**
* Is this a cross build
* Windows 10
* Python 3.8.5
* Meson 0.57.1
* Ninja 1.10.0.git.kitware.jobserver-1

PR will be created right after this ticket

brainelectronics:
stdout content is printed to the console again if capturing to a file is not enabled

### DIFF
--- a/mesonbuild/scripts/meson_exe.py
+++ b/mesonbuild/scripts/meson_exe.py
@@ -87,6 +87,9 @@ def run_exe(exe: ExecutableSerialisation, extra_env: T.Optional[dict] = None) ->
         if not skip_write:
             with open(exe.capture, 'wb') as output:
                 output.write(stdout)
+    else:
+        if stdout:
+            print(stdout.decode())
 
     return 0
 


eli-schwartz:
CI failure is relevant:

```
[0/1] Running external command check_exists (wrapped by meson to set env)
  File "/__w/meson/meson/meson.py", line 29, in <module>
    sys.exit(mesonmain.main())
  File "/__w/meson/meson/mesonbuild/mesonmain.py", line 229, in main
  File "/__w/meson/meson/mesonbuild/mesonmain.py", line 218, in run
  File "/__w/meson/meson/mesonbuild/mesonmain.py", line 166, in run_script_command
  File "/__w/meson/meson/mesonbuild/scripts/meson_exe.py", line 114, in run
    return run_exe(exe)
  File "/__w/meson/meson/mesonbuild/scripts/meson_exe.py", line 91, in run_exe
    print(stdout.decode())
AttributeError: 'NoneType' object has no attribute 'decode'
FAILED: meson-check_exists 
```

bonzini:
This change was done on purpose, since `generate_custom_target` uses `as_meson_exe_cmdline` with verbose set to False. @xclaesse, what do you think?

brainelectronics:
@xclaesse it's fixed now with our pipeline. I'm not sure about the two failing jobs here.

We use python logger and output the logged content to sys.stdout on the console. After your mentioned change all output got piped into the pickle file. We need that content in our console. Enabling capture solved the issue, but showed another one. 
We (intentionally) set the output to a never existing file (with capture:false) to force a re-run of that custom target. This target is doing some pre- another one post processing of the binary output, which is always required.
That tasks/runs will be skipped if capture is set to true, as there is no change detected after the first run. So second and any further run pass simply with "nothing to do" on ninja.

TheChangeloggers:
@xclaesse any updates or feedback from your side, would like to see this change merged as well. We also lack the stdout content in our build

eli-schwartz:
Can you fix the commit message as I mentioned above?

Also, this should not be two commits. Please next time fix the first commit using `git commit --amend && git push -f`.

Since both commits now exist, you can combine them together by using `git rebase -i origin/master` and changing the word "pick" to "fixup" in the line for the second commit.

eli-schwartz:
If if p.returncode != 0 then I think this will now print *twice*.

bonzini:
> is your commit requirement now satisfied

You can remove the merge commit. It will be handled when the pull request is merged.

xclaesse:
I think it would be better to set `exe.verbose` to True in the cases we want, it has the advantage of not capturing potentially long output into memory before dumping it all.

xclaesse:
I think you can add `verbose=True` here: https://github.com/mesonbuild/meson/blob/master/mesonbuild/interpreter/mesonmain.py#L103. That will make postconf/dist/install scripts output to stdout directly without buffering.

kevr:
What's the blocker on this one; did code coverage fall short? The codecov report doesn't show a diff at the current time; no base commit is associated.

bruchar1:
Can be closed. Fixed by #11368.

PR #8500, opened 2021-03-09
change RSP quote style decision logic

### LINKED ISSUE #8494: LDC can't parse RSP file issued by Meson
**Describe the bug**
When the command line length hits a threshold, it is converted to a RSP file. In this file, some of the arguments are quoted using python `shlex.quote`.
`shlex.quote` happens to quote using single quotes `'` (at least on Windows).

In my case, the RSP file contains among many other `'-L=/PDB:lib\\doptest.pdb'` and LDC fails to parse with `Error: unrecognized file extension pdb'`. (note the `'` in the file extension).

LDC forwards expansion of RSP file directly to LLVM ([implementation](https://llvm.org/doxygen/CommandLine_8cpp_source.html#l00930))

So who's to blame here? `shlex` ? LLVM? Meson (for using shlex)?

**To Reproduce**
you may set `MESON_RSP_THRESHOLD` to a low value and compile a D project with LDC.

**Expected behavior**
No failure

**system parameters**
* plain native build 
* Windows 10
* Python 3.9.0
* Meson 0.57.1
* Ninja 1.10.1

rtbo:
I took a conservative approach, fixing only what I have observed wrong.

Fixes #8494

### DIFF
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -33,7 +33,9 @@ from ..arglist import CompilerArgs
 from ..compilers import (
     Compiler, CCompiler,
     DmdDCompiler,
-    FortranCompiler, PGICCompiler,
+    FortranCompiler,
+    LLVMDCompiler,
+    PGICCompiler,
     VisualStudioCsCompiler,
     VisualStudioLikeCompiler,
 )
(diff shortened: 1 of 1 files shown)


dcbaker (approved):
This looks fine, although I'd really prefer to add a method to the compiler class to declare this. But I can do that as a follow up.

PR #8501, opened 2021-03-09
Add address sanitizer support for Visual Studio.

jpakkane:
https://devblogs.microsoft.com/cppblog/address-sanitizer-for-msvc-now-generally-available/

### DIFF
--- a/mesonbuild/compilers/cpp.py
+++ b/mesonbuild/compilers/cpp.py
@@ -663,7 +663,6 @@ class VisualStudioCPPCompiler(CPP11AsCPP14Mixin, VisualStudioLikeCPPCompilerMixi
         CPPCompiler.__init__(self, exelist, version, for_machine, is_cross,
                              info, exe_wrapper, linker=linker, full_version=full_version)
         MSVCCompiler.__init__(self, target)
-        self.base_options = {OptionKey(o) for o in ['b_pch', 'b_vscrt', 'b_ndebug']} # FIXME add lto, pgo and the like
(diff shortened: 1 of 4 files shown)


mesonbuild/compilers/cpp.py
```diff
@@ -663,7 +663,6 @@ def __init__(self, exelist: T.List[str], version: str, for_machine: MachineChoic
         CPPCompiler.__init__(self, exelist, version, for_machine, is_cross,
                              info, exe_wrapper, linker=linker, full_version=full_version)
         MSVCCompiler.__init__(self, target)
-        self.base_options = {OptionKey(o) for o in ['b_pch', 'b_vscrt', 'b_ndebug']} # FIXME add lto, pgo and the like
```
dcbaker, line 666: Why was this deleted, is it duplicated further down?

dcbaker (changes requested):
Presumably this deserves a release snippet, and possibly something in the docs?

mesonbuild/compilers/cpp.py
jpakkane: No, further up. :)

It is set in the base class. I had to spend a fair bit of time head scratching why the changes I added there were not having any effect. For comparison the C version uses `base_options.update` so as not to clobber the base code.

dcbaker: I'm sure I meant up, though who knows... Anyway, yes that should be using update.

dcbaker (approved):
LGTM thanks

PR #8503, opened 2021-03-09
clangformat: Add include and ignore files

xclaesse:


xclaesse:
In my current case it's because my source tree has basically 3 top folders: `src/` and `plugins/` is code that we wrote and want to format, and `sdk/` which is a huge 3G code drop from a vendor that I don't want to format, just walking that tree takes noticeable time. `sdk/` contains pre-built binaries, in addition to source code, so it's not even built by Meson.

I usually rely on [run-clang-format.py](https://github.com/Sarcasm/run-clang-format) but it feels like something Meson could do itself, to not have to copy that script in every project I touch...

Maybe in the future we could even extend more this feature:
- Make it a meson subcommand instead of relying on ninja.
- Add support for other formatters like GNU indent that GStreamer currently use (but is causing issues).
- Add `clang-format-check` target that would return with error code as soon as a file gets modified, to be used in CI instead of grepping for `File reformatted` string (I should do that in this PR tbh).

jpakkane:
>  just walking that tree takes noticeable time. 

We could detect the "do nothing" clangformat files (they are basically just two lines) and not recurse there. This would make our behaviour match what upstream clang-format does.

xclaesse:
@jpakkane feels easier to have include/ignore lists at toplevel tbh. Note the reason I opted for `.clang-fornat-ignore` file is because it's used by `run-clang-format.py` script that is pretty popular already, so I expected some people could just switch to use Meson native support transparently.

jpakkane:
> Note the reason I opted for .clang-fornat-ignore file is because it's used by run-clang-format.py script that is pretty popular already,

I tried to look up those files, but my google-fu failed. If they already exist and we use the same format for them then it's fine.

jpakkane:
The docs should explicitly mention that these files have the same format etc as run-clang-format, thoubh.

xclaesse:
> The docs should explicitly mention that these files have the same format etc as run-clang-format, thoubh.

ok, mentioned that in the doc now. tbh I don't know if it's super popular tool, I found it from stackoverflow a while ago...

tristan957:
Putting my 2 cents in, but I find the built-in clang-format target kind of pointless, and it should be users' jobs to hook up in a run_target(), otherwise people are going to request alternative formatter targets like `uncrustify` for instance.

jpakkane:
That is how it already works, if you define your own `clang-format` run target, the default one is not created.

jpakkane:
Is there any harm in it? Sure it's there, but if you never call it it does not really matter?

tristan957:
Developers could get frustrated that it doesn't work as intended because only certain folders/files should be formatted.

xclaesse:
> Putting my 2 cents in, but I find the built-in clang-format target kind of pointless, and it should be users job to hook it up in a run_target(), otherwise people are going to request alternative formatter targets like `uncrustify` for instance.

I see no reason why we can't also add support for other formatters if someone comes with a PR. We can probably share most of the code, in the end it's just the command to execute that change.

> Right, but I am saying there shouldn't even be this default behavior to begin with. I have chosen to call my format target of choice `format` because I have black and clang-format for formatting files. Still the default `clang-format` target persists.

I'm personally not super fan of this being a ninja target neither, tbh. But since it was already there, I just extended a bit its capabilities to make it more useful. IMHO this could become a meson command in the future that does not rely on the backend: `meson format -C builddir` ? But that can still be done later if we want.

tristan957:
I like format as a command much more that this ninja target, but I could still live without it. Like I said just wanted to get my 2 cents in as a casual observer since this PR was tangentially related to something I have been thinking about.

xclaesse:
@jpakkane any objection on this now?

jpakkane:
Needs conflict resolution but other than that, no.

PR #8403, opened 2021-02-23
Submit/rust fix shared internal linking

dcbaker:
Found through inspection, even in 0.57 we don't correctly generate linking arguments for shared c linkage libraries that Meson itself builds. This is a bug in 0.57 and should be applied to 0.57.2

PR #8504, opened 2021-03-09
Add test to build a mixed C/Rust shared library.

jpakkane:
This will probably fail on Windows and macOS. But those platforms probably don't care about this use case so disabling the test there is an option.


dcbaker:
This should work on Windows I think, but dynamic linking on macOS is broken... So is linking with frameworks. I just can't be bothered to get out my MacBook and get it working, though that will probably bite me later...

dcbaker:
This does overlap with #8403, which I should fix

dcbaker (approved):

PR #8505, opened 2021-03-09
Disable broken asm test on Cygwin as nobody knows how to fix it.

jpakkane:


dcbaker (approved):

jon-turney:
Logs are no longer available, so I can only observe that test now passes.

If I wasn't nobody, you could ping me here or on IRC and I could perhaps take a look in a more timely fashion...

PR #8507, opened 2021-03-09
Add a method to the Compiler and Linker objects for rsp syntax

### LINKED ISSUE #8502: compiler's rsp syntax should be a Compiler method
This is mostly just to help me remember to do this.

See: #8500

dcbaker:
Since this really is something about the compilers and linkers themselves, the information should be encoded there.

Fixes  #8502

### DIFF
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -32,14 +32,11 @@ from .. import compilers
 from ..arglist import CompilerArgs
 from ..compilers import (
     Compiler, CCompiler,
-    DmdDCompiler,
     FortranCompiler,
-    LLVMDCompiler,
     PGICCompiler,
-    VisualStudioCsCompiler,
     VisualStudioLikeCompiler,
 )
-from ..linkers import ArLinker, VisualStudioLinker
+from ..linkers import ArLinker, RSPFileSyntax
 from ..mesonlib import (
     File, LibType, MachineChoice, MesonException, OrderedSet, PerMachine,
     ProgressBar, quote_arg, unholder,
@@ -51,6 +48,8 @@ from ..interpreter import Interpreter
 
 if T.TYPE_CHECKING:
     from ..linkers import StaticLinker
+    from ..compilers.cs import CsCompiler
+
 
 FORTRAN_INCLUDE_PAT = r"^\s*#?include\s*['\"](\w+\.\w+)['\"]"
 FORTRAN_MODULE_PAT = r"^\s*\bmodule\b\s+(\w+)\s*(?:!+.*)*$"
@@ -90,12 +89,6 @@ else:
     quote_func = quote_arg
     execute_wrapper = []
     rmfile_prefix = ['rm', '-f', '{}', '&&']
-
-
-def dlang_cl_rsp(compiler: Compiler) -> bool:
-    '''Return whether the D compiler accepts cl style RSP quote'''
-    return (isinstance(compiler, DmdDCompiler) or
-        mesonlib.is_windows() and isinstance(compiler, LLVMDCompiler))
 
 
 def get_rsp_threshold():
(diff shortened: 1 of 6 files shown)


dcbaker:
@rtbo, I think this is the same logic as you had, but I don't have a windows machine handy to test on, could you make sure this doesn't immediately regress the D compilers on Windows?

jon-turney:
Thanks.  I'd noted this as future work after fbacf87af525f7b69e4f1a310f6ef38c5853406c, but never got around to it.

I think `can_linker_accept_rsp()` and `rsp_file_syntax()` are only being used in the `NinjaRule()` constructor calls, so they could perhaps be collapsed to a single method?

jon-turney (approved):

rtbo:
@dcbaker I have a failure:
```
  File "C:\sdk\perso\meson\mesonbuild\mesonmain.py", line 132, in run
  File "C:\sdk\perso\meson\mesonbuild\msetup.py", line 275, in run
    app.generate()
  File "C:\sdk\perso\meson\mesonbuild\msetup.py", line 183, in generate
    self._generate(env)
  File "C:\sdk\perso\meson\mesonbuild\msetup.py", line 245, in _generate
    intr.backend.generate()
  File "C:\sdk\perso\meson\mesonbuild\backend\ninjabackend.py", line 529, in generate
    self.generate_rules()
  File "C:\sdk\perso\meson\mesonbuild\backend\ninjabackend.py", line 1115, in generate_rules
    self.generate_static_link_rules()
  File "C:\sdk\perso\meson\mesonbuild\backend\ninjabackend.py", line 1868, in generate_static_link_rules
    rspfile_quote_style=static_linker.rsp_file_syntax(),
  File "C:\sdk\perso\meson\mesonbuild\linkers.py", line 114, in rsp_file_syntax
    assert not self.can_linker_accept_rsp(), 'This linker accepts RSP, but doesn\' provide a supported format, this is a bug'
AssertionError: This linker accepts RSP, but doesn' provide a supported format, this is a bug
```

rtbo:
The missing implementation is in the `DLinker` class. 

For DMD, here is a comment in the [response parsing code](https://github.com/dlang/dmd/blob/master/src/dmd/root/response.d)
```d
/* 
 * ...
 * Arguments are separated by spaces, tabs, or newlines. These can be
 * imbedded within arguments by enclosing the argument in "".
 * Backslashes can be used to escape a ".
 * A line comment can be started with #.
 * Recursively expands nested response files.
 * ...
 */
```
I suppose MSVC quote will work fine.

For LDC there is a one implementation for Windows and one for other platforms (GNU compatible)
Using `CL` on windows and `GCC` everywhere else should work (there won't be much RSP anywhere else though).
(The same applies to Clang BTW)

For GDC I suppose only GNU... (But I don't use this compiler)

dcbaker:
@rtbo since the DLinker just uses the D compiler, I'm just going to proxy that through, does that seem reasonable?

rtbo:
> since the DLinker just uses the D compiler, I'm just going to proxy that through, does that seem reasonable?

Yes.

> Is that clang with the gcc-style interface, or clang-cl with the msvc interface that uses cl style rsp files?

`clang-cl` expects CL RSP syntax. GNU is accepted everywhere else. (see [clang source code](https://github.com/llvm/llvm-project/blob/155c49e0878de667b8021b9ba685390151dee11e/clang/tools/driver/driver.cpp#L393))
You can also force one of the two modes with the special switches `--rsp-quoting=posix` or `--rsp-quoting=windows`.
I don't see any choice made based on the running OS.

For the other LLVM based tools such as LDC or LLD, the choice is made following the running OS (if windows then CL mode else GNU mode)

dcbaker:
Cool, then I think clang is correct in this series now, (assuming that you don't pass the `--rsp-quoting` argument.

jpakkane (approved):

rtbo:
I have tested 73b434060f39d671f6d2e80a045cb6857a41a66b successfully on Windows.

PR #5206, opened 2019-04-02
RFC: A way to instantiate custom "feature" objects

Akaricchi:
It is often useful to override the values of some build options. Sometimes you have options that depend on one another, or are incompatible with the target platform, etc. In Meson, this is usually accomplished with the use of variables:
```meson
opt_foobar = get_option('foobar')

if foobar_makes_no_sense
  opt_foobar = foobar_fallback
endif

# rest of the code only uses opt_foobar
```

However, there is one case where this pattern breaks: feature options. Every type of object that `get_option` may return can be declared in `meson.build`, except for these. That is rather unfortunate, because it is not trivial to replicate their semantics — in fact, it is impossible to do so accurately. Some fairly trivial cases can be worked around:
```meson
opt_video = get_option('use_video')

opt_video_x11 = get_option('use_video_x11')
opt_video_wayland = get_option('use_video_wayland')
opt_video_opengl = get_option('use_video_opengl')
opt_video_openglesv2 = get_option('use_video_openglesv2')
opt_video_vulkan = get_option('use_video_vulkan')

if opt_video.disabled()
    opt_video_x11 = opt_video
    opt_video_wayland = opt_video
    opt_video_opengl = opt_video
    opt_video_openglesv2 = opt_video
    opt_video_vulkan = opt_video
    opt_render = opt_video
endif
```
[(snippet source: my SDL2 meson port/wrap)](https://github.com/taisei-project/SDL-mirror/blob/2eb019d2b655c19e12b9d6967859f5305eae25ff/meson.build#L77)
In this case, I can get away with using `opt_video` when what I really mean is "a feature that is disabled".

Some hacks are uglier than others:
```meson
egl_dep = dependency('egl', required : opt_video_openglesv2)
if not egl_dep.found()
    egl_dep = dependency('egl', required : opt_video_wayland)
endif
```
[(snippet source: my SDL2 meson port/wrap)](https://github.com/taisei-project/SDL-mirror/blob/2eb019d2b655c19e12b9d6967859f5305eae25ff/meson.build#L134)
This one in particular isn't just inelegant, it can potentially degrade performance in case the `egl` dependency is not installed. Yet it is the most concise way to express that multiple features require the same dependency. The double look-up also makes the configuration log slightly more confusing.

There are also cases with no easy workarounds. Here's a paraphrased snippet from the Taisei build system that I'm currently refactoring:

```meson

opt_zip = get_option('zip_packages')
opt_package_data = get_option('package_data')

if opt_zip.disabled()
[...]

PR #8512, opened 2021-03-10
Utility methods for feature objects

bonzini:
This pull request adds utility methods to operate on feature objects. These methods should cover most needs for manual creation of feature objects (as in #5206). For example:

```
if opt_video.disabled()
    opt_video_x11 = opt_video
    opt_video_wayland = opt_video
    ...
endif
```

can be changed to this:

```
opt_video_x11 = opt_video_x11.require(opt_video)
opt_video_wayland = opt_video_wayland.require(opt_video)
...
```

The other method I am adding, `may_disable_if`, is useful to skip possibly expensive checks when it's known that a library will not be used.


docs/markdown/Reference-manual.md
```diff
@@ -2724,6 +2724,24 @@ The following methods are defined for all [`feature` options](Build-options.md#f
 - `enabled()`: returns whether the feature was set to `'enabled'`
 - `disabled()`: returns whether the feature was set to `'disabled'`
 - `auto()`: returns whether the feature was set to `'auto'`
+- `allowed()` *(since 0.58.0)*: returns whether the feature was set to `'enabled'` or `'auto'`
+- `may_disable_if(value)` *(since 0.58.0)*: returns the feature, with
+  `'auto'` converted to `'disabled'` if value is true.
```
xclaesse, line 2729: If I understand correctly, the use case is something like:
`directx_opt = get_option('directx').may_disable_if(host_machine.os != 'windows')`

Could be just me, but I find the name a bit weird. What about shorter `.disabled_if()`?

docs/markdown/Reference-manual.md
```diff
@@ -2724,6 +2724,24 @@ The following methods are defined for all [`feature` options](Build-options.md#f
...
+- `require(value, error_message: '')` *(since 0.58.0)*: returns
+  the object itself if the value is true; an error if the object is
+  `'enabled'` and the value is false; a disabled feature if the object
+  is `'auto'` or `'disabled'` and the value is false.
```
xclaesse, line 2733: So this is the same as `may_disable_if()` except that it is an error if the user set that feature to enabled when it cannot be supported, right?

This makes me wonder if we even need `may_disable_if()`, the way I see feature options is that if user enable one it's should be a guarantee that if for any reason that cannot be satisfied they get an error and not a silently disabled feature.

docs/markdown/Reference-manual.md
```diff
@@ -2724,6 +2724,24 @@ The following methods are defined for all [`feature` options](Build-options.md#f
...
+
+`require` is useful to restrict the applicability of `'auto'` features,
+for example based on other features or on properties of the host machine:
+
+```
+if get_option('directx').require(host_machine.os == 'windows',
+       error_message: 'DirectX only available on Windows').allowed() then
+  src += ['directx.c']
+  config.set10('HAVE_DIRECTX', 1)
+endif
```
xclaesse, line 2743: I find it more readable when not packing everything in the condition:
```meson
directx_opt = get_option('directx').require(host_machine.os == 'windows',
  error_message: 'DirectX only available on Windows'
)
if directx_opt.allowed()
  ...
endif
```

mesonbuild/interpreter.py
```diff
@@ -82,29 +82,71 @@ class FeatureOptionHolder(InterpreterObject, ObjectHolder[coredata.UserFeatureOp
...
     def auto_method(self, args, kwargs):
-        return self.held_object.is_auto()
+        return self.value == 'auto'
+
+    @permittedKwargs({'error_message'})
+    def require_method(self, args, kwargs):
+        if len(args) != 1:
+            raise InvalidArguments('Expected 1 argument, got %d.' % (len(args), ))
+        if not isinstance(args[0], bool):
+            raise InvalidArguments('boolean argument expected.')
```
xclaesse, line 129: Please use `@typed_pos_args`

mesonbuild/interpreter.py
```diff
@@ -82,29 +82,71 @@ class FeatureOptionHolder(InterpreterObject, ObjectHolder[coredata.UserFeatureOp
...
+        error_message = kwargs.pop('error_message', '')
+        if error_message and not isinstance(error_message, str):
```
xclaesse, line 131: nitpicking

```suggestion
        if not isinstance(error_message, str):
```

mesonbuild/interpreter.py
```diff
@@ -82,29 +82,71 @@ class FeatureOptionHolder(InterpreterObject, ObjectHolder[coredata.UserFeatureOp
...
 
     @noPosargs
     @permittedKwargs({})
     def disabled_method(self, args, kwargs):
-        return self.held_object.is_disabled()
+        return self.value == 'disabled'
+
+    @noPosargs
+    @permittedKwargs({})
+    def allowed_method(self, args, kwargs):
```
xclaesse, line 116: `@FeatureNew`

docs/markdown/Reference-manual.md
bonzini: `may_disable_if()` does not degrade `enabled` to `disabled`, only `auto`. This is why it has a slightly more complicated name than `disabled_if()`.

The design should cover exactly what you mention here. The idea is that you use

```
    libone = dependency('one', requires: get_option('one'))
    libtwo = dependency('two',
       requires: get_option('two').may_disable_if(not libone.found())
```

This way:
* with `-Dtwo=auto` libtwo won't even be searched unless libone was found
* with `-Dtwo=enabled` it will be an error if libtwo is absent, even if the build will end up not using it because libone is absent.

mesonbuild/interpreter.py
```diff
@@ -82,29 +82,71 @@ class FeatureOptionHolder(InterpreterObject, ObjectHolder[coredata.UserFeatureOp
...
+            raise InvalidArguments('boolean argument expected.')
+        error_message = kwargs.pop('error_message', '')
+        if error_message and not isinstance(error_message, str):
+            raise InterpreterException("Error message must be a string.")
+        if args[0]:
+            return self
+
+        if self.value == 'enabled':
+            prefix = 'Feature {} cannot be enabled'.format(self.name)
+            prefix = prefix + ': ' if error_message else ''
```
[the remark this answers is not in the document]
bonzini: I thought about it, but it's the opposite, in this case `host_machine.os != 'windows'`. It's also much more user-oriented than `assert()`, so an expression might not be very friendly.

mesonbuild/interpreter.py
```diff
@@ -82,29 +82,71 @@ class FeatureOptionHolder(InterpreterObject, ObjectHolder[coredata.UserFeatureOp
...
+                             'require': self.require_method,
+                             'may_disable_if': self.may_disable_if_method,
                              })
 
+    @property
+    def value(self):
+        return 'disabled' if not self.held_object else self.held_object.value
+
+    def as_disabled(self):
+        return FeatureOptionHolder(None, self.name, None)
```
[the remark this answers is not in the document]
dcbaker: Further, you're breaking the type annotations, they don't allow this to be `None`. So you either need to fix the type annotations, or use a `UserFeatureOption()`. I'd also prefer the latter.

mesonbuild/interpreter.py
```diff
@@ -82,29 +82,71 @@ class FeatureOptionHolder(InterpreterObject, ObjectHolder[coredata.UserFeatureOp
...
-        return self.held_object.is_auto()
+        return self.value == 'auto'
+
+    @permittedKwargs({'error_message'})
+    def require_method(self, args, kwargs):
+        if len(args) != 1:
+            raise InvalidArguments('Expected 1 argument, got %d.' % (len(args), ))
+        if not isinstance(args[0], bool):
+            raise InvalidArguments('boolean argument expected.')
+        error_message = kwargs.pop('error_message', '')
```
dcbaker, line 130: Please don't `pop` here, use `get`, Let's avoid mutation if possible.

jpakkane:
> opt_video_x11 = opt_video_x11.require(opt_video)

I don't really grasp why this is the way it is. You have to have had looked up `opt_video_x11` already and overwriting a variable with itself is always a bit suspect. For example, if you have already used it by accident somewhere, then that old value persists. Could this be put in `get_option` somehow:

```meson
opt_video_x11 = get_option('video_x11', ..., if_enabled: opt_video)
```

bonzini:
> Could this be put in get_option somehow.

I like it, what about:

```
# if condition is false and no conflict_message: auto->disabled, enabled->enabled
# if condition is false and conflict_message is present: auto->disabled, enabled->error
get_option('video_x11',
      condition: **condition**,
      conflict_message: **string**)
```

docs/markdown/Reference-manual.md
xclaesse: Hmm, I'm not really convinced by `may_disable_if()` tbh, do you have real use-case in qemu for that? It feels like `.require()` is easy and intuitive, but `may_disable_if()` is for subtle corner case that I'm not even sure to really grasp.

xclaesse:
The condition can be put on get_option() indeed, that would make sense.

I'm personally still not convinced by the enabled->enabled case, IMHO if a feature can only be used under certain condition and you set that feature enabled, you always should get an error if the condition is not met.

That way we could even support boolean options, if condition is false true->error?

dcbaker:
Yeah, the semantics of `enabled` right now is "Use this or error", and `disabled` is "don't use this or error", so anytime `enabled` or `disabled` effectively degrades to `auto` we're breaking the semantics of the type.

bonzini:
> IMHO if a feature can only be used under certain condition and you set that feature enabled, you always should get an error if the condition is not met.

The enabled->enabled case is not for feature that can only be used under certain conditions; it's for dependencies that `meson.build` decides not to use for unrelated reasons that are essentially implementation details. For example, assume that dbus is for some internal reason used only on Linux but not on other POSIX systems&mdash;but it's implementation detail, therefore `-Ddbus=enabled` should not fail compilation on non-Linux systems.

But this is very different from degrading `enabled` to `auto`, to some extent it's the opposite. Having `enabled`->`enabled` will _ensure_ that all the tests are performed even in cases where the dependencies is not used, but it will skip a pointless test and speed up meson if the feature is `auto`. In the example above, dbus will be tested for even on non-Linux systems.

xclaesse:
@bonzini OOC do you have an actual use-case for may_disable_if() in qemu? I would prefer to leave that out of scope for now if you did that just because you _think_ it could be useful.

Do I understand it correctly that it would be for example if an optional feature is not recommended on certain platforms. For example on Windows you could prefer directx over opengl, unless user explicitly enable opengl?
```
# On Windows disable opengl by default, unless user explicitly asked for it.
gl_opt = get_option('opengl').may_disable_if(platform == 'windows')
gl_dep = dependency('opengl', required: gl_opt)
```

xclaesse:
Could be as well `gl_opt = get_option('opengl').may_disable_if(directx_dep.found())` to mean that if we already build the directx backend better disable the opengl backend, unless user really wants both. Right?

bonzini:
> do you have an actual use-case for may_disable_if() in qemu? I would prefer to leave that out of scope for now if you did that just because you think it could be useful.

Yes, I have a bunch of

```
if get_option('xkbcommon').auto() and not have_system
  xkbcommon = not_found
else
  xkbcommon = dependency('xkbcommon', required: get_option('xkbcommon'),
                         method: 'pkg-config', kwargs: static_kwargs)
endif
```

which would become `required: get_option('xkbcommon').may_disable_if(not have_system)`. The logic is: if system emulators are not enabled, xkbcommon is not used so degrade `auto` to `disabled`; but if `-Dxkbcommon=enabled`, check whether the dependency is present even if it will not be used.

> Do I understand it correctly that it would be for example if an optional feature is not recommended on certain platforms.

That's another usecase that I hadn't thought of.

docs/markdown/Reference-manual.md
xclaesse: Now that I have a better understanding of the use-case for this, I like it... But I think we should include example(s) in the documentation to help understanding the intended usage. Since you're using directx story for `.require()`, would make sense to include my example from https://github.com/mesonbuild/meson/pull/8512#issuecomment-799612560.

xclaesse: What about also renaming `value` to `condition` here for clarity?

docs/markdown/Reference-manual.md
xclaesse: What about renaming `value` to `condition` here for clarity?

Could also change the error_message kwarg to be simply an optional 2nd positional argument, just like `assert(condition, 'optional message')`?

Maybe add working like "This describes a condition required to allow enabling this feature" ?

xclaesse:
On 2nd thought, I think I prefer having those methods on FeatureOptionHolder rather than in get_option() because it is more flexible. Doing it in get_option() restricts us to kwargs that only apply to certain type of options, would make no sense on string options... Doing it on FeatureOptionHolder it's easier to add as many methods and arguments we want in the future.

Doing it in get_option() does not really fix @jpakkane concern anyway, because you could still do `get_option('dirextx')` in one place and `get_option('directx', condition: platform == 'windows')` somewhere else.

We have been discussing ways to combine feature options together since feature options have been introduced. This is the best proposal I've seen so far. The biggest difference with previous attemps is here we combine feature with boolean, instead of feature with feature, that makes a lot more sense IMHO.

docs/markdown/Reference-manual.md
xclaesse: What about even adding a table in the doc to make it totally clear:

| Feature / Condition  | True | False |
| -------------------------- | ------ | ------- |
| Enabled                   | Enabled | Enabled |
| Disabled                  | Disabled | Disabled |
| Auto                         | Disabled | Auto |

bonzini:
> I also have a real world use case today that I found that this would be awesome for

That one pretty much matches mine.

> Also, on the color of the bikeshed, I would disable_if or disable_auto_if to may_disable_if

I added the "may" because it doesn't turn enabled into disabled, so disable_if is misleading. But disable_auto_if works for me too. I won't get back to this too soon, but it'll remain on my todo list.

xclaesse:
I like `disable_auto_if` because it is the most descriptive and unambigous, even if it sounds a bit long. Maybe `opt.auto_to_disabled(condition)` ? I don't immediately see a use-case for `auto_to_enabled()` but it would be a nice parallel indeed.

dcbaker:
```meson
opt_foo = get_option('foo')
opt_bar = get_option('bar').enable_auto_if(opt_foo.enabled())
```
maybe? Although that case having something like:
```meson
opt_foo = get_option('foo')
opt_bar = get_option('bar').auto_to(opt_foo)
```
might be more useful. Though that doesn't need to happen in this PR, just throwing ideas out.

tp-m:
I'd love to make use of this for what it's worth.

Use case 1: https://gitlab.freedesktop.org/cairo/cairo/-/merge_requests/159/diffs

Use case 2: flipping auto defaults for GStreamer plugins based on some kind of "group" options (e.g. an option to enable/disable gpl plugins, or video plugins, or all external plugins).

mensinda:
OK, looks like `vs2019x64vs` is broken. The error is in `AllPlatformTests.test_meson_compile`, which seems to have stopped working randomly. It also fails for me in #8817.

I can also reproduce this error on master locally in my Windows 10 VM. Seems like a VS update broke something...

mensinda:
Looks like this is definitely an MSBuild bug...

```ps1
msbuild -target:mylib_sha '.\build\shared library linking test.sln'      # Works
msbuild -target:mycpplib_sha '.\build\shared library linking test.sln'   # Fails
```

Part of the `.sln` file:

```sln
Microsoft Visual Studio Solution File, Format Version 11.00
# Visual Studio 2019
Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "mylib@sha", "mylib@sha.vcxproj", "{83B210AF-6664-4E8D-9B48-900E158BA88B}"
EndProject
Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "prog@exe", "prog@exe.vcxproj", "{88597C1C-B700-472D-BD78-DA98CF318CBC}"
EndProject
Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "mycpplib@sha", "mycpplib@sha.vcxproj", "{A6B275FC-5B0C-4B4E-BBA9-B1314796B91B}"
EndProject
```

Here, `mylib@sha` is defined first, followed by `prog@exe` and then `mycpplib@sha`. If I remove the first two `Project()` calls, the `msbuild -target:mycpplib_sha` command works. I guess that means that only the first target in the `.sln` file can be manually specified...

bonzini:
Ping @jpakkane for review.

docs/markdown/Reference-manual.md
```diff
@@ -2818,6 +2818,31 @@ The following methods are defined for all [`feature` options](Build-options.md#f
...
+- `require(value, error_message: '')` *(since 0.59.0)*: returns
+  the object itself if the value is true; an error if the object is
+  `'enabled'` and the value is false; a disabled feature if the object
+  is `'auto'` or `'disabled'` and the value is false.
+
+`require` is useful to restrict the applicability of `'auto'` features,
+for example based on other features or on properties of the host machine:
+
+```
+if get_option('directx').require(host_machine.os == 'windows',
```
jpakkane, line 2840: That should probably be `host_machine.system()`.

Issue #2173, opened 2017-08-11
Rust: Support external crates


zeenix:
I was told that support of Rust was experimental so I can expect hickups and that's understandable but currently it seems external crates are not supported. Since you can't really do much in Rust without external crates, it basically means that Meson doesn't really support Rust. IMO until this issue is addressed in some way, support for Rust shouldn't be advertised at all.

TingPing:
Correct me if I am wrong but Cargo simply just builds all dependencies and there is no concept of external dependencies (ala pkgconfig) right? Does Rust even have any standard directory to install shared libraries? If so we should support `rust.find_library()`.

tp-m:
I think what's needed is more something that does what Cargo does - downloads crates as per `Cargo.toml` and `Cargo.lock` and builds them / links them in as needed. This would be really powerful, esp. for mixed C/Rust projects.

TingPing:
So we either:

- Re-implement Cargo
- Simply call Cargo making the Meson integration worthless

zeenix:
@TingPing typically they just do static linking in Rust currently. `cargo` will simply download, build all deps and then take care of linking your binaries.

zeenix:
@TingPing yeah but please keep in mind that this is useful for projects that will be a mix of Rust and C (or some other language for which meson is a good option).

TingPing:
I realize its valuable but its not a realistic thing we can support. It is like wanting Meson to support building autotools projects. Cargo is its own independent and incompatible build system.

zeenix:
@TingPing I would completely understand if you decide not to support this in meson but then either Rust support shouldn't be advertised at all or this limitation should be explicitly specified. As I mentioned, without this, Rust support is mostly useless.

TingPing:
We have Rust support, we don't have Cargo support. It is sad that the Rust community only cares about Cargo and does not installed shared libraries or have any tooling to collect information about installed libraries.

tp-m:
While it is certainly "independent", It's not clear to me that it really is "incompatible". I see crates more like a special kind of subproject.

zeenix:
@TingPing yeah, i know but fact remains that the support is useless for most uses so really nothing to advertise. I fully agree about Rust folks not doing things the standard way.

TingPing:
I wasn't aware Meson was 'advertising' rust support other than it being documented. It is certainly basic.

zeenix:
[...]

PR #8514, opened 2021-03-10
Rust module : Add support for cargo subprojects

dcbaker:
I want to preface this that this is still a work in progress, there's a *lot* left to do, but I have been able to compile a couple of simple demo projects with this, so we've gotten to the point that "something works". I'm sharing this mostly to get a little visibility into the work.

The end goal is to be able to build Rust crates as fully native Meson subprojects, there's still some work to do on a lot of fronts, but it's getting there.

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -7,6 +7,7 @@ Cargo to meson.build:
         - target cfg() paring
 
     - TODO:
+        - links
         - targets
             - no support for the triple style
         - tests being triggered when they shouldn't?
@@ -40,4 +41,3 @@ Overall:
     - Figure out what to do about multiple versions of the same crate
     - Add a mechanism to test.json to verify that the correct tests are being run
     - Come up with a better solution for subprojects affecting each-others options
-    - how to handle recursive subprojects
--- a/TODO.md
+++ b/TODO.md
@@ -1,0 +1,43 @@
+Cargo to meson.build:
+    - Completed:
+        - dependencies
+        - binaries
+        - libraries
+        - features
+        - target cfg() paring
+
+    - TODO:
+        - targets
+            - no support for the triple style
+        - tests being triggered when they shouldn't?
+        - build-dependencies
+        - example
+        - bench
+        - profile
+        - patch
+        - test for default_options being set correctly
+        - binaries:
+            - test for auto bin and manual bin collision
+            - targets required features
+            - doctest?
+            - bench
+        - libraries
+            - targets required features
+            - proc-macro
+            - doctest?
+            - bench
+        - Add test for invalid feature names
+        - extend test 15 to cover test, exampeles, and benches
+        - is the use of a disabler() for dev-dependencies smart?
+
+Wrap
+    - TODO:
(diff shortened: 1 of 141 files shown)


jpakkane:
Does someone who actually knows how this should work want to review this?

bauen1:
Related to #2173

For fetching crates, there is [cargo-download](https://docs.rs/crate/cargo-download/0.1.2).

I was looking into setting up an embedded rust project with meson (to make it easy to integrate into C/C++ projects), and the lack of crates support is annoying. I was hacking up a way to build the crates I need with a combination of `cargo download` and `cargo build`, but this would be much better.

dcbaker:
I might look at cargo download. My preference would be to not have *any* dependency on cargo at all, but we'll just have to see once this is actually in a state to land.

DemiMarie:
@dcbaker I don’t think that avoiding a Cargo dependency is realistic.  Cargo does a massive amount of work related to stuff like dependency resolving, some of which is even (if I understand it correctly) NP-complete (so you won’t be able to practically reimplement it in pure Python without massive performance hits).

What I *do* consider realistic is building Rust code that vendors its dependencies (via `cargo vendor`).  That avoids having to interact with `crates.io`, which is in and of itself subtle (consider stuff like HTTP proxies, git fetching with CLI vs libgit2, etc).  In short, I recommend letting Cargo do all of the network I/O; Meson can then do the actual building.

eli-schwartz:
Handling rust dependencies as subprojects means that none of this matters. Subprojects are a form of vendoring which uses pinned versions, so the only dependency resolution involved is "use it? Yes or no?" And if yes, "here is exactly one url which may be used to acquire it".

dcbaker:
Right, and my plan initially is to use cargo vendor to do that vendoring. Long term i would like meson wrap to be able to handle it, which is not run so build or compile time, so the cost is paid once and then checked into git, making the cost relatively small

Ericson2314:
FYI c.f. https://github.com/kolloch/crate2nix and https://github.com/NixOS/nixpkgs/tree/master/pkgs/build-support/rust/build-rust-crate, which build Rust crates from the lockfile with little use of Cargo itself. `crate2nix` uses some libraries from Cargo just to be able to share data structures and parsing. `buildRustCrate` seems to just use `cargo metadata`.

bonzini:
Hi @dcbaker, what is the state of this? I might be able to spend some time on it.

It seems plausible to rely solely on `cargo update` to build Cargo.lock + `cargo vendor subprojects` for the downloading, possibly mapping `meson subprojects` commands to cargo commands too.

dcbaker:
This should really be closed in favor of #10151

The basic problem I've run into is that so many of the core types are really tied up in the interpreter, so I wrote all of the code to fix the build.py problem, and that's stalled. If you wanna help me figure out why #10714 all works fine for me but fails utterly in CI that would be super helpful :)

bonzini:
Ok in the past I tried bisecting the failure, I can try doing that here too.

PR #8517, opened 2021-03-10
Remove -pipe from default list of args. Closes #8508.

### LINKED ISSUE #8508: How to get meson to not add the -pipe option to my cpp compilation steps
I have declared my project settings as such.
`project('MyProject', 'cpp', 'c',
    default_options: ['buildtype=plain', 'b_staticpic=false', 'b_asneeded=false', 'b_lundef=false']`

The project requires cross-compilation and neither my c _args or my cpp_args that come under the properties section of my crossfile include the '-pipe' option.

However, the build.ninja that meson generates for my build directories shows that the '-pipe' has been added as a compiler argument in all my c/cpp compilation steps. Not sure if I missed it, but I'm unable to find a way to not enable meson to do this from the documentation. Is there a default option that I can set to not allow meson to insert compilation arguments other than the ones listed in my crossfile?

System parameters:
OS: Debian stretch
Python version: 3.5.3
Meson version: 0.56.0
Ninja version; 1.7.2

jpakkane:
Let's check the time it takes to run this through CI. If there are any performance slowdowns we have to consider more deeply what to do.

### DIFF
--- a/docs/markdown/snippets/nopipe.md
+++ b/docs/markdown/snippets/nopipe.md
@@ -1,0 +1,10 @@
+## `-pipe` no longer used by default
+
+Meson used to add the `-pipe` command line argument to all compilers
+that supported it, but no longer does. If you need this, then you can
+add it manually. However note that you should not do this unless you
+have actually measured that it provides performance improvements. In
+our tests we could not find a case where adding `-pipe` made
+compilation faster and using `-pipe` [can cause sporadic build
+failures in certain
+cases](https://github.com/mesonbuild/meson/issues/8508).
--- a/mesonbuild/compilers/fortran.py
+++ b/mesonbuild/compilers/fortran.py
@@ -335,12 +335,6 @@ class IntelFortranCompiler(IntelGnuLikeCompiler, FortranCompiler):
     def get_preprocess_only_args(self) -> T.List[str]:
         return ['-cpp', '-EP']
 
-    def get_always_args(self) -> T.List[str]:
-        """Ifort doesn't have -pipe."""
-        val = super().get_always_args()
-        val.remove('-pipe')
-        return val
-
     def language_stdlib_only_link_flags(self) -> T.List[str]:
         return ['-lifcore', '-limf']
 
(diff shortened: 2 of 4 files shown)


dcbaker:
I'm not sure this is a good idea. We just ran into a case the other day in one of the projects that autotools doesn't add -pipe and it broke stuff that worked in meson because it automatically added -pipe

dcbaker:
I know you're concerned about option bloat, but I feel like this might still be a good case for an option because for most people in most cases they want `-pipe`, even if distros tend to add it, as a user I don't.

jpakkane:
>  We just ran into a case the other day in one of the projects that autotools doesn't add -pipe and it broke stuff that worked in meson because it automatically added -pipe

This seems like a toolchain bug (as is the original reporter's bug with pipe). One of the reasons why build systems have been so terrible for such a long time is that all these hacks and workarounds are shoved into the build system rather than getting fixed at the core.

Also, is there a case where `-pipe` yields a performance benefit? Or any other benefit? I have not been able to find such a case.

dcbaker:
It probably makes a huge difference if you're on spinning rust (like most of the build machine at my work, often on slow spinning rust). It also avoids writes to my SSD, which is a lifetime of my SSD enhancement :)

jpakkane:
>  It also avoids writes to my SSD

Temp files that are immediately deleted won't get written to disk if there is enough RAM. And if there isn't, then _something_ is getting swapped out in any case.

dcbaker:
I guess in the worst case we can always revert this. There is one more thing that should be changed off the top of my head, one of the fortran compilers (I think IFORT) explicitly removes -pipe from its arguments because it inherits this, we should remove that too.

jpakkane:
Done.

nirbheek:
A linked issue on the original issue talks about faster build times on network filesystems: https://github.com/spack/spack/pull/23162

They ended up adding it manually to CFLAGS. I think an option is probably worth adding here.

jpakkane:
Reading the comments it turns out that even then it might not be needed. In any case this is getting way too niche to warrant a top level Meson option.

PR #8520, opened 2021-03-11
compilers: clang-cl: Also accept .s files

scott-ph:
clang-cl has support for gas-compatible assembly files.

### DIFF
--- a/mesonbuild/compilers/mixins/visualstudio.py
+++ b/mesonbuild/compilers/mixins/visualstudio.py
@@ -412,6 +412,9 @@ class ClangClCompiler(VisualStudioLikeCompiler):
         super().__init__(target)
         self.id = 'clang-cl'
 
+        # Assembly
+        self.can_compile_suffixes.add('s')
+
     def has_arguments(self, args: T.List[str], env: 'Environment', code: str, mode: str) -> T.Tuple[bool, bool]:
         if mode != 'link':
             args = args + ['-Werror=unknown-argument']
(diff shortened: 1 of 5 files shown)


dcbaker (changes requested):
Hey Scott! can you enable clang-cl in `tests/common/128` so we can get some test coverage for this? Otherwise this looks good.

scott-ph:
> Hey Scott! can you enable clang-cl in `tests/common/128` so we can get some test coverage for this? Otherwise this looks good.

yo @dcbaker! it's been a while. I took a crack at enabling the test and it wound up being a bit ugly. Let me know if you have any suggestions

dcbaker (approved):
LGTM, thanks!

dcbaker:
I kicked the CI, so lets see if it'll turn green.

jpakkane:
This ... makes no sense. Two tests test_dist_git and test_command_line fail with this:

    ../../../../c/Users/VSSADM~1/AppData/Local/Temp/tmpo9fl_peu/meson.build:1:0: ERROR: Unable to detect linker for compiler "cc -Wl,--version"

But in an environment where clang-cl is not even available (as far as I know). The test does pass for other MRs so this does seem relevant.

jpakkane:
Maybe it was just Cygwin being sporadic. Please rebase to run the tests again.

jpakkane:
Cygwin failure is unrelated so merging. Thanks.

PR #8521, opened 2021-03-11
Allow path separators in output kwargs

### LINKED ISSUE #2320: Allow configure_file, custom_target and generator output file to subdirectories under builddir
I've encountered two cases that would require configurating / generating file into subdirectories.

1. Project configuration header that would be included from other headers in the same project

Say I have a library project named foo and the project layout is like this:
```
include/
    foo/
        bar/
            foobar/
                foobar.h # includes config.h
            bar.h        # includes config.h
        foo.h            # includes config.h
        config.h.in # configured by meson
src/
    foo/
        foo.cpp
...
```
We would want our client code use `#include <foo/config.h>` to include our `config.h` since we don't install `config.h` directly to `includedir`. But for our own code this would be impossible since generated `config.h` is directly under `<builddir>`, not `<builddir/foo>`.

Of course, for `foo.cpp` the work-around is easy: just `#include <config.h>`. But it becomes complicated we need to include `config.h` from `bar.h` and `foobar.h`.

In fact, there is no way to correctly include `config.h` that both work in project directory and after installation. If we want `bar.h` and `foobar.h` to be correct after installation we must `#include <foo/config.h>` or use relative path in them. Including `<foo/config.h>` won't compile, and although relative path happens to work for `bar.h`, it won't for `foobar.h`. Neither will it work if we reside in a subdir. Using `#include <config.h>` allows us to compile, but `bar.h` and `foobar.h` breaks up after installation.

The correctly way to configure and include config.h should be just generate it under `builddir/foo` and `#include <foo/config.h>` everywhere, just as if it were in the position of `config.h.in`. Currently meson does not allow this, and forces a flat include directory, or hacking into builddir.

2. Protobuf generation

This one is simpler. If we have some protos like
```
protos/
    foo/
        foo.proto
    bar.proto
````
it would be nice if we could use one generator for both protos and and get under build directory something like this:
```
protos/
    foo/
        foo.pb.h
        foo.pb.cc
    bar.pb.h
    bar.pb.cc
````
Althouth this would also require something like @DIRNAME@ substitution if we are to pass generated sources to build targets properly.  

PS: I suggest putting every generated file from `configure_file()` and `custom_target()` (including ones in `subdir`s) into one directory tree named something like `<builddir>/sources@gen`, add it to include directory, and allow `output`ing to subdirectories under it. This should avoid potential name clashes and give developers control over generated folder layout.

tristan957:
Path separators in output kwargs is very important for allowing Meson to wrap other tooling/compilers. Take for instance Meson wrapping Maven. This may sound kind of crazy, but imagine a C library and you are writing JNI bindings for it. There are essentially 2 build artifacts that you need to generate: a JNI library written in C and a JAR written in Java. In my case, errors in the C library exist as what are essentially errno values. I decided to pull in a dependency from Maven Central which provides errno values in Java-land. Manual management of the class path can be quite difficult and tedious among other things when reaching for `javac` directly. I decided to use Maven to compile the Java sources and create a JAR. What I essentially want is a one step process to build both my JNI library and the JAR (`ninja -C build`). Maven produces many artifacts in many different folders. Just off the top of my head, if you didn't care about any of the other build artifacts, Maven produces the JAR in the root of its build directory, and *.class files in the a subdir called `classes`. In order to properly model Maven in Meson, I need access to path separators in output kwargs. I have tested this patch locally on my system, and it works great.

Not entirely sure what the best way to add a test is.

Ninja: 1.10
Meson: master

Fixes #2320

### DIFF
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -1536,8 +1536,11 @@ class Generator:
             depfile = kwargs['depfile']
             if not isinstance(depfile, str):
                 raise InvalidArguments('Depfile must be a string.')
-            if os.path.basename(depfile) != depfile:
-                raise InvalidArguments('Depfile must be a plain filename without a subdirectory.')
+            deppath = pathlib.PurePath(depfile)
+            if deppath.is_absolute():
+                raise InvalidArguments('Depfile must not be an absolute path')
+            if '..' in deppath.parts:
+                raise InvalidArguments('Depfile must not contain ".."')
             self.depfile = depfile
         if 'capture' in kwargs:
             capture = kwargs['capture']
@@ -2286,8 +2289,11 @@ class CustomTarget(Target, CommandBase):
                 raise InvalidArguments('Output must not be empty.')
             if i.strip() == '':
                 raise InvalidArguments('Output must not consist only of whitespace.')
-            if has_path_sep(i):
-                raise InvalidArguments(f'Output {i!r} must not contain a path segment.')
+            ipath = pathlib.PurePath(i)
+            if ipath.is_absolute():
+                raise InvalidArguments('Output must not be an absolute path')
+            if '..' in ipath.parts:
+                raise InvalidArguments('Output path must not contain ".."')
             if '@INPUT@' in i or '@INPUT0@' in i:
                 m = 'Output cannot contain @INPUT@ or @INPUT0@, did you ' \
                     'mean @PLAINNAME@ or @BASENAME@?'


tristan957:
I see @nirbheek did this quite a while back: https://github.com/mesonbuild/meson/commit/7c15837bd25d37672bfddcb6ef9402792b3d3c23.
Added some extra protections as well, which this patch should since it is pretty much the exact same thing, and I agree with his restrictions. Would love to generate some discussion around this.

tristan957:
I understand why this restriction existed, but if Meson is to grow to something larger, this is a necessary update. Not every compiler/generator/tool has the ability to be told where artifacts should go, unlike C compilers for instance.

jpakkane:
> I understand why this restriction existed, but if Meson is to grow to something larger, this is a necessary update. Not every compiler/generator/tool has the ability to be told where artifacts should go, unlike C compilers for instance.

Having this limitation is a good thing and it makes things easier and simpler for the vast majority of people and use cases while making things slightly inconvenient for those who use broken tools. Switching it makes things worse for the majority who do it right for the benefit of a minority that does things wrong. This will not be accepted. A different approach would be to make a higher level interface (like how `unstable_external_project` works) that does what people want rather than break invariants so that people need to then fix things by hand in their build files. That is both fragile and tedious.

tristan957:
You can't properly model dependency graphs without path separators in so many tools. I don't understand what would break here. Previously path separators were not allowed, and now they are in this PR. What would break? The people that need the feature will use it, the people that don't won't.

eli-schwartz:
Removing a design restriction requires more rationale than merely coding the implementation.

This PR is inherently unmergeable on the grounds that the discussion in the linked issue does not include the lead meson developer (@jpakkane) making the philosophical call to change his mind about this.

Closing this PR to prevent discussion about "why" from ending up split across two different github tickets.

PR #8522, opened 2021-03-11
externalproject: misc stuff

xclaesse:
commit 85011f01be75546718fbf8f11d755e449edf2ecf

    externalproject: Flatten configure_options kwarg

commit 65b323e4c053fbc258f127ee900ba8df47807914

    externalproject: Do not add --includedir by default
    
    Some projects (e.g. OpenSSL) does not support setting include directory
    at all.

commit 1174973bd2db35fd13931a802b4ba23fe57bf027

    externalproject: Fix error when used from main project


xclaesse:
Remove commit adding `option_arg()` method because following IRC discussion we could want to do it a bit differently.

Issue #8516, opened 2021-03-10
Meson inconsistency error when trying to build project with a subproject and subdir using the same subproject declared dependency


ihilt:
Starting with e00df904 meson fails to complete on [this repo](https://github.com/ihilt/meson_subproject_error.git) with this error.
```
bar/meson.build:1:0: ERROR: Inconsistency: Subproject has overridden the dependency with another variable than 'baz_dep'
```
This can be observed by doing the following.
```bash
# install meson from e00df904 or after. easiest path would be pip3 install meson==0.56
$ git clone https://github.com/ihilt/meson_subproject_error.git
$ cd meson_subproject_error
$ meson build
```

**Expected behavior**
I expect the meson build to complete.

**system parameters**
* Is this a [cross build](https://mesonbuild.com/Cross-compilation.html) or just a plain native build (for the same computer)?
Native.
* what operating system (e.g. MacOS Catalina, Windows 10, CentOS 8.0, Ubuntu 18.04, etc.)
Fedora 33 and Ubuntu 20.04
* what Python version are you using e.g. 3.8.0
3.9.2 on Fedora 33 and 3.8.5 on Ubuntu 20.04
* what `meson --version`
0.56.0
* what `ninja --version` if it's a Ninja build
It's not a ninja build.

xclaesse:
Thanks for the test case, it helps a lot. I included it as unit test with the fix.

ihilt:
On 03/12/21 at 10:09:12 AM, Xavier Claessens ***@***.***> wrote:
>Thanks for the test case, it helps a lot. I included it as unit test with the fix.

No problem.  Do you know if the fix will be back-ported to 0.56.0-0.57.1?

xclaesse:
We only do fix releases in the latest branch, so maybe 0.57.2 or 0.58.0.

ihilt:
On 03/12/21 at 11:58:19 AM, Xavier Claessens ***@***.***> wrote:
>We only do fix releases in the latest branch, so maybe 0.57.2 or 0.58.0.

Ok, got it. Really appreciate how quickly you fixed this, thanks!

PR #8525, opened 2021-03-12
interpreter: Cache found dependency before converting for include_type

xclaesse:
Fixes: #8516.


xclaesse:
CC @mensinda since that code comes from you.

mensinda (approved):
LGTM

PR #8529, opened 2021-03-13
Replace custom program lookups with find_program

jpakkane:
These are all standalone commits so this should go in with a rebase.

### DIFF
--- a/mesonbuild/interpreter.py
+++ b/mesonbuild/interpreter.py
@@ -4010,6 +4010,9 @@ external dependencies (including libraries) must go to "dependencies".''')
                 mlog.warning('''Custom target input \'%s\' can\'t be converted to File object(s).
 This will become a hard error in the future.''' % kwargs['input'], location=self.current_node)
         kwargs['env'] = self.unpack_env_kwarg(kwargs)
+        if 'command' in kwargs and isinstance(kwargs['command'], list) and kwargs['command']:
+            if isinstance(kwargs['command'][0], str):
+                kwargs['command'][0] = self.func_find_program(node, kwargs['command'][0], {})
         tg = CustomTargetHolder(build.CustomTarget(name, self.subdir, self.subproject, kwargs, backend=self.backend), self)
         self.add_target(name, tg.held_object)
         return tg
--- a/mesonbuild/interpreter.py
+++ b/mesonbuild/interpreter.py
@@ -1884,7 +1884,6 @@ class MesonMain(InterpreterObject):
         InterpreterObject.__init__(self)
         self.build = build
         self.interpreter = interpreter
-        self._found_source_scripts = {}
         self.methods.update({'get_compiler': self.get_compiler_method,
                              'is_cross_build': self.is_cross_build_method,
                              'has_exe_wrapper': self.has_exe_wrapper_method,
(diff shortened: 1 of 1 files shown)


mesonbuild/interpreter.py
```diff
@@ -1908,25 +1907,10 @@ def __init__(self, build: 'build.Build', interpreter: 'Interpreter'):
...
-            found = dependencies.ExternalProgram(prog, search_dir=self.interpreter.environment.build_dir)
-        else:
-            found = dependencies.ExternalProgram(prog, search_dir=search_dir)
-
-        if found.found():
-            self._found_source_scripts[key] = found
-        else:
-            m = 'Script or command {!r} not found or not executable'
-            raise InterpreterException(m.format(prog))
+        found = self.interpreter.func_find_program({}, prog, {}).held_object
```
xclaesse, line 1913: nitpicking: I think you should use find_program_impl() instead.

mesonbuild/interpreter.py
```diff
@@ -3983,6 +3983,9 @@ def _func_custom_target_impl(self, node, args, kwargs):
                 mlog.warning('''Custom target input \'%s\' can\'t be converted to File object(s).
 This will become a hard error in the future.''' % kwargs['input'], location=self.current_node)
         kwargs['env'] = self.unpack_env_kwarg(kwargs)
+        if 'command' in kwargs and isinstance(kwargs['command'], list) and kwargs['command']:
```
xclaesse, line 3970: I think you should do `command = extract_as_list(kwargs, 'command')` because you also want to replace the command if it's a single string.

mesonbuild/interpreter.py
```diff
@@ -3983,6 +3983,9 @@ def _func_custom_target_impl(self, node, args, kwargs):
...
+            if isinstance(kwargs['command'][0], str):
```
xclaesse, line 3971: Also if it's a mesonlib.File.

xclaesse (commented):
While at it, could do it in `run_command_impl()` too for consistency.

mesonbuild/interpreter.py
jpakkane: I tried that and it does not work, because a File can point to a thing that will only be generated during build, so looking it up will fail.

xclaesse: Hmm, I thought File was only for configure time generated files. OOC which API returns File for a build time file?

jpakkane: I don't remember offhand, but if you change that and run project tests, you'll find out fairly quickly.

xclaesse: Hmm, I tried with
```patch
diff --git a/mesonbuild/interpreter.py b/mesonbuild/interpreter.py
index 6a51f3d99..23a26896f 100644
--- a/mesonbuild/interpreter.py
+++ b/mesonbuild/interpreter.py
@@ -4013,6 +4013,10 @@ external dependencies (including libraries) must go to "dependencies".''')
                 mlog.warning('''Custom target input \'%s\' can\'t be converted to File object(s).
 This will become a hard error in the future.''' % kwargs['input'], location=self.current_node)
         kwargs['env'] = self.unpack_env_kwarg(kwargs)
+        command = mesonlib.extract_as_list(kwargs, 'command')
+        if command and isinstance(command[0], (str, mesonlib.File)):
+            command[0] = self.find_program_impl(command[0])
+            kwargs['command'] = command
         tg = CustomTargetHolder(build.CustomTarget(name, self.subdir, self.subproject, kwargs, backend=self.backend), self)
         self.add_target(name, tg.held_object)
         return tg
```

And project tests passed locally. Did you see it failing on the CI? Maybe there is another root cause?

mesonbuild/interpreter.py
xclaesse: Note, a difference if you use find_program_impl() is it's silent by default. I'm not sure we want to log the this program lookup, it could be spamming a bit too much IMHO.

jpakkane:
I don't have the time or energy to keep polishing this. Sorry. However since it already improves things, I'll merge it. Further improvements can go in their own MRs.

PR #8531, opened 2021-03-15
Warn when grabbing internals of subprojects with include_directories.

jpakkane:
And again we find that if something is possible, no matter how wrong, someone will do it.

### DIFF
--- a/mesonbuild/interpreter.py
+++ b/mesonbuild/interpreter.py
@@ -4595,6 +4595,28 @@ put in the include directories by default so you only need to do
 include_directories('.') if you intend to use the result in a
 different subdirectory.
 ''')
+            else:
+                try:
+                    self.validate_within_subproject(a, '')
+                except InterpreterException:
+                    mlog.warning('include_directories sandbox violation!')
+                    print(f'''The project is trying to access the directory {a} which belongs to a different
+subproject. This is a problem as it hardcodes the relative paths of these two projeccts.
+This makes it impossible to compile the project in any other directory layout and also
+prevents the subproject from changing its own directory layout.
+
+Instead of poking directly at the internals the subproject should be executed and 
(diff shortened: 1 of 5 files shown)


mesonbuild/interpreter.py
```diff
@@ -4572,6 +4572,28 @@ def build_incdir_object(self, incdir_strings, is_system=False):
...
+
+Instead of poking directly at the internals the subproject should be executed and 
+it should set a variable that the caller can then use. Something like:
+
+# In subproject
+my_inc = include_directories('include')
+
+# In parent project
+dep_sp = subproject('sub')
+dep_inc = dep_sp.get_variable('my_inc')
```
xclaesse, line 4594: I would rather recommend using declare_dependency() and use the subproject as dependency() fallback. IMHO we should even deprecate or at least no recommend subp.get_variable(), it's also grabbing a subproject internals.

tp-m: For this case `declare_dependency()` would be better, but I think there are legitimate use cases for `subproject.get_variable()` too where it would be awkward to make a dummy dependency to export some variables/values.

So not convinced we should deprecate `subproject.get_variable()`.

xclaesse: Deprecate is indeed probably too much, but I would not recommend it in this warning at least.

xclaesse:
I'm hitting this warning a lot when configuring GStreamer. It happens when doing `include_directories('.')` at the root of a subproject.
- `self.validate_within_subproject(a, '')` I think that should be instead `self.validate_within_subproject(self.subdir, a)`.
- `self.validate_within_subproject()` is actually not expecting getting the root directory of a subproject, it expects a file in the root directory. It's missing a check for `norm == project_root`.

PR #8532, opened 2021-03-15
Delete leftover files from some tests.

jpakkane:


run_project_tests.py
```diff
@@ -1306,6 +1306,15 @@ def get_version(t: dict) -> str:
         print('{0:<{2}}: {1}'.format(tool['tool'], get_version(tool), max_width))
     print()
 
+def clear_transitive_files():
+    a = Path('test cases/common')
+    for d in a.glob('*subproject subdir/subprojects/subsubsub*'):
+        if d.is_dir():
+            shutil.rmtree(d)
```
dcbaker, line 1313: probably should `windows_proof_rmdir` and and the one for files too, right?

run_unittests.py
```diff
@@ -5793,12 +5793,16 @@ def test_subproject_variables(self):
            contain required keys.
         '''
         tdir = os.path.join(self.unit_test_dir, '20 subproj dep variables')
+        stray_file = os.path.join(tdir, 'subprojects/subsubproject.wrap')
+        if os.path.exists(stray_file):
+            os.unlink(stray_file)
```
dcbaker, line 5798: One more, with that it looks good to me.

dcbaker (approved):

PR #8536, opened 2021-03-16
msetup: do not print bogus "Option ... is:" messages

bonzini:
`get_non_matching_default_options` is checking a string from `project_default_options` against a validated value from `coredata.options`.

Passing the string to validate_value ensures that the comparison is sound; otherwise, `'false'` might be compared against `False`
and a bogus difference is printed.

Would be nice to have in 0.57.2 (QEMU is seeing this now that we can set `b_staticpic` to false in the default options).


xclaesse:
Makes sense. We could also maybe move those into summary(). What it could also do is print all options user has set, Meson knows it from cmd_line.txt file. But those could be done later, this is a fix that could go in 0.57.2 when you get CI green.

PR #8537, opened 2021-03-16
coredata: be more robust on unpickling errors

bonzini:
When reverting from 0.57 to 0.56, one can see an error like this:

```
  File /Users/pm215/src/qemu-for-merges/meson/mesonbuild/coredata.py,
line 1016, in load
    obj = pickle.load(f)
ModuleNotFoundError: No module named 'mesonbuild.mesonlib.universal';
'mesonbuild.mesonlib' is not a package
FAILED: build.ninja
```

The reason is that the old version fails to resolve mesonbuild.mesonlib, which is a similar situation to the existing AttributeError check.  Raise a MesonException for ModuleNotFoundError as well, so that reconfiguration proceeds using cmd_line.txt.

### DIFF
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -943,7 +943,7 @@ def load(build_dir: str) -> CoreData:
             obj = pickle.load(f)
     except (pickle.UnpicklingError, EOFError):
         raise MesonException(load_fail_msg)
-    except AttributeError:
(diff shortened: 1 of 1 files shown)


bonzini:
Also proposing this for 0.57.2.

xclaesse:
LGTM, when you fix the syntax error :)

Issue #8533, opened 2021-03-15
`meson test` hangs if the write side of the stdout/stderr pipes is left open


bonzini:
If a test spawns a runaway subprocess and therefore the stdout or stderr pipe is left open, the test should timeout. Instead, Meson notices that the test has finished and just waits for the `stdo_task` and `stde_task` to finish, causing a hang. Worse, the hang cannot be cured with Ctrl-C.

hadess:
(This happened, coupled with a bug in the tests, with a version of [this MR](https://gitlab.freedesktop.org/hadess/power-profiles-daemon/-/merge_requests/46) which I've kept locally in a branch called `wip/hadess/meson-hang-bug`, let me know if it's needed for debugging in the future)

PR #8546, opened 2021-03-17
mtest: timeout tests if they have runaway child processes

bonzini:
mtest: timeout if the write side of pipes does not close

If a test program forks a child, the pipes might remain open and `await stdo_task`/`await stde_task` will never complete in `SingleTestRunner._run_cmd()`.

Instead, catch them in TestSubprocess.wait() so that the whole process group is killed and the test times out.

Fixes: #8533


bonzini:
Regression in 0.57. There is still a change in behavior from before 0.57, where the timeout was not reported at all (the tests passed and the runaway child process was never killed), but I think the remaining change counts as a bugfix. @hadess is that okay with you?

PR #8549, opened 2021-03-17
interpreter: Add varname as positional arg in dep.get_variable()

xclaesse:


mesonbuild/interpreter.py
```diff
@@ -512,10 +512,16 @@ def partial_dependency_method(self, args, kwargs):
...
     @permittedKwargs({'cmake', 'pkgconfig', 'configtool', 'internal', 'default_value', 'pkgconfig_define'})
     @FeatureNewKwargs('dep.get_variable', '0.54.0', ['internal'])
-    def variable_method(self, args, kwargs):
+    def variable_method(self, args: T.Tuple[T.Optional[str]], kwargs: T.Dict[str, T.Any]) -> str:
+        default_varname = args[0]
+        if default_varname is not None:
+            FeatureNew('0.58.0', 'Positional argument to dep.get_variable()').use(self.subproject)
+            for k in ['cmake', 'pkgconfig', 'configtool', 'internal']:
+                if k not in kwargs:
+                    kwargs[k] = default_varname
```
dcbaker, line 524: you could simplify this to
I think you can simplify this to:
```suggestion
            for k in ['cmake', 'pkgconfig', 'configtool', 'internal']:
                kwargs.setdefault(k, default_varname)
```

test cases/common/212 dependency get_variable method/meson.build
```diff
@@ -62,3 +62,4 @@ assert(idep.get_variable(pkgconfig : 'foo', cmake : 'foo', configtool : 'foo',
 
 idep = declare_dependency(variables : ['foo=value'])
 assert(idep.get_variable(internal: 'foo') == 'value')
+assert(idep.get_variable('foo') == 'value')
```
dcbaker, line 65: I would however like some tests that cover a case like `dep.get_variable('foo', cmake : 'FOO')`, (or pkg-config, same idea) where the specialized variable both is and isn't found, to ensure that the specialized variable has precedence.

dcbaker (changes requested):
I'm still not convinced this is a good idea, it's making something that was explicit implicit, and gives a false sense of "if i just write dep.get_variable('foo') everything will work!" until it silently doesn't.

However, if others like the idea I wont hold it up.

eli-schwartz:
> gives a false sense of "if i just write dep.get_variable('foo') everything will work!"

The alternative is giving the same people the false sense of "if i just write `dep.get_variable(pkgconfig: 'foo')` everything will work!"

It already takes manual work to support every dependency method.

I like this idea, so +1 from me...

test cases/common/212 dependency get_variable method/meson.build
eli-schwartz: On the previous line we could change to:
```
assert(idep.get_variable('FOO', internal: 'foo') == 'value')
```

xclaesse:
> > gives a false sense of "if i just write dep.get_variable('foo') everything will work!"
> 
> The alternative is giving the same people the false sense of "if i just write `dep.get_variable(pkgconfig: 'foo')` everything will work!"

Exactly, I had to fix that anti-pattern countless times. It's in the top list of reasons why fallback to subproject does not work. In practice the different variable name is really just a CMAKE thing, every project that ship a pkgconfig file will use the same variable name for their internal and pc dep. But often projects will ship pkgconfig OR cmake, not both, which means you rarely actually need to pass different variable names.

test cases/common/212 dependency get_variable method/meson.build
xclaesse: @eli-schwartz Added that test too, good point.

eli-schwartz (approved):

test cases/common/212 dependency get_variable method/meson.build
dcbaker: Please do leave the just `get_variable(internal : 'foo')` we should have a test with and without the default.

xclaesse: Sure, I now test the 3 cases.

PR #8551, opened 2021-03-17
Check for, and prefer, pkgconf when looking for pkg-config

dcbaker:
Some distros (debian and nix) provide pkg-config and pkgconf, but don't link pkgconf to pkg-config. Others do link pkg-config to pkgconf (arch). pkgconf is just better than pkg-config is pretty much every way, so we should prefer pkgconf if it's installed. Right now we don't even search for pkgconf.

This series fixes that, and also centralizes the names to look for in a single place.


docs/markdown/snippets/prefer_pkgconf.md
```diff
@@ -0,0 +1,5 @@
+## Pkgconf is now the prefer pkg-config implementation
```
xclaesse, line 1:
```suggestion
## Pkgconf is now the prefered pkg-config implementation
```

docs/markdown/snippets/prefer_pkgconf.md
```diff
@@ -0,0 +1,5 @@
...
+
+Previous versions of meson will not look for pkgconf unless it has `pkg-config`
+alias. Meson will not search for `pkgconf` first, as it is a superior
```
xclaesse, line 4:
```suggestion
alias. Meson will now search for `pkgconf` first, as it is a superior
```

xclaesse (commented):
Small nitpicking. Other than that, LGTM.

xclaesse:
@dcbaker At least on Ubuntu pkgconf package conflicts with pkg-config, so you can install only one of them. And if you install pkgconf you get a symlink `/usr/bin/pkg-config -> pkgconf`

dcbaker:
Nix at least allows you to install both simultaneously and doesn't create a symlink, I'm sure Gentoo does the same.

docs/markdown/snippets/prefer_pkgconf.md
QuLogic:
```suggestion
## Pkgconf is now the preferred pkg-config implementation
```

jpakkane:
That test failure is suspicious as it only happens on this MR.

docs/markdown/snippets/prefer_pkgconf.md
```diff
@@ -0,0 +1,6 @@
+## Pkgconf is now the preferred pkg-config implementation
+
+Previous versions of meson would not look for `pkgconf` unless it had a
+`pkg-config` alias. Meson will not search for `pkgconf` first, as it is a
```
eli-schwartz, line 4:
```suggestion
`pkg-config` alias. Meson will now search for `pkgconf` first, as it is a
```

unittests/failuretests.py
```diff
@@ -150,10 +150,9 @@ def test_extraframework_dependency_method(self):
 
     def test_sdl2_notfound_dependency(self):
         # Want to test failure, so skip if available
-        if shutil.which('sdl2-config'):
-            raise unittest.SkipTest('sdl2-config found')
-        self.assertMesonRaises("dependency('sdl2', method : 'sdlconfig')", self.dnf)
-        if shutil.which('pkg-config'):
+        if not shutil.which('sdl2-config'):
+            self.assertMesonRaises("dependency('sdl2', method : 'sdlconfig')", self.dnf)
+        if not shutil.which('pkg-config'):
```
eli-schwartz, line 155: So... Apparently this was always broken? This deserves a call-out in the commit message.

docs/markdown/snippets/prefer_pkgconf.md
```diff
@@ -0,0 +1,7 @@
+## Pkgconf is now the preferred pkg-config implementation
+
+Previous versions of meson would not look for `pkgconf` unless it had a
+`pkg-config` alias. Meson will now search for `pkgconf` first, as it is a
+superior implementation, and then for `pkg-config` second, if `pkgconf` is
+not found. For some OSes this will not result in any changes, as `pkg-config` is
```
eli-schwartz, line 6: "For most OSes this will not result in any changes, as either pkgconf is not installed, or pkg-config is an alias to pkgconf."

eli-schwartz (changes requested):

unittests/failuretests.py
dcbaker: Does the revised commit message look better?

eli-schwartz: Ah, I was thinking about the fact that this test only passes if sdl2-config skips... because if it gets to the pkg-config check, it only runs when pkg-config *is* found, and then it asserts that the dependency must not be found, which seems... confusing.

PR #8553, opened 2021-03-18
Tests: add (broken) override_options test case

MathieuDuponchelle:
This reproduces a case I encountered in a project, where a pkg-config
dependency (aws-cpp-sdk-s3) has -std=c++11 in its cflags, but the
project uses c++17 features and passes -std=c++17 through
override_options.

The final compile command has c++17 before c++11, causing compilation
to fail.

### DIFF
--- a/run_unittests.py
+++ b/run_unittests.py
@@ -7915,6 +7915,18 @@ class LinuxlikeTests(BasePlatformTests):
         self.assertEqual(len(obj_files), 1)
         self.assertTrue(obj_files[0].endswith('-prelink.o'))
 
+    @skipIfNoPkgconfig
+    def test_override_options_cpp_std(self):
+        '''
+        Tests that a cpp_std passed through override_options takes
+        precedence over a different -std retrieved from a pkg-config
+        dependency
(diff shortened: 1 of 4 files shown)


jpakkane:
I would call this a bug in the pc file. It should not define the version to be used. The C++ standard libraries guarantee (as far as I can tell) that old and new code can be compiled and linked together and the result will work. If the code can not be used with an older version of C++ then its headers should have an `#ifdef` check for the version and error out if it is too old.

Still, that being said, we should probably detect this and either remove all `-std` flags coming from pc files or, at the very least, print an error.

MathieuDuponchelle:
> I would call this a bug in the pc file. It should not define the version to be used. The C++ standard libraries guarantee (as far as I can tell) that old and new code can be compiled and linked together and the result will work. If the code can not be used with an older version of C++ then its headers should have an `#ifdef` check for the version and error out if it is too old.
> 
> Still, that being said, we should probably detect this and either remove all `-std` flags coming from pc files or, at the very least, print an error.

Sounds fair to me

PR #8555, opened 2021-03-19
minstall: always track meson-created directories

eli-schwartz:
If a custom_target output is a directory, we install it as a directory, not as a file. And, we try to track subdirectories which are created so uninstalling works. But one directory creation did not go through DirMaker, in the case where the output directory does not have any further subdirectories.

Consolidate on makedirs, since I don't see much point in using os.mkdir right here.

Discovered while investigating https://github.com/DISTRHO/DISTRHO-Ports/issues/83

### DIFF
--- a/mesonbuild/minstall.py
+++ b/mesonbuild/minstall.py
@@ -292,10 +292,6 @@ class Installer:
         # ['sub1', ...] means skip only those.
         self.skip_subprojects = [i.strip() for i in options.skip_subprojects.split(',')]
 
-    def mkdir(self, *args: T.Any, **kwargs: T.Any) -> None:
-        if not self.dry_run:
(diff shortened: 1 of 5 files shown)


xclaesse:
A unit test would be great. Other than that LGTM.

eli-schwartz:
`common/8 install/` and run_unittests.py:test_uninstall seems like a great place to put the test. Will do.

eli-schwartz:
There we go. `./run_unittests.py -k uninstall` locally passes.

eli-schwartz:
Test failure is irrelevant; cygwin once again complains: `test cases/frameworks/13 yelp/meson.build:1:0: ERROR: Compiler cc can not compile programs.`

dcbaker (approved):

PR #8556, opened 2021-03-19
run wrapped-due-to-env commands on unix via the env program

eli-schwartz:
First, check if the env program exists. If it does, it is faster than doing it via a python script `basically-env.py` that maybe imports all of mesonbuild.* as a side effect of project structure.

We do not, however, use env for setting up PATH additions, since env can override an environment variable but not extend it. So in that case we still need to wrap the command via python.

By default, all run_targets (at least) are wrapped and now wrap via the `env` program as they export e.g. `MESONINTROSPECT='/usr/bin/meson introspect'`

/cc @xclaesse

### DIFF
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -23,6 +23,7 @@ import json
 import os
 import pickle
 import re
+import shutil
 import typing as T
 import hashlib
 
@@ -584,15 +585,31 @@ class Backend:
         if any('\n' in c for c in es.cmd_args):
             reasons.append('because command contains newlines')
 
-        if es.env and es.env.varnames:
+        if env and env.varnames:
             reasons.append('to set env')
 
+        # force_serialize passed to this function means that the VS backend has
+        # decided it absolutely cannot use real commands. This is "always",
+        # because it's not clear what will work (other than compilers) and so
+        # we don't bother to handle a variety of common cases that probably do
+        # work.
+        #
+        # It's also overridden for a few conditions that can't be handled
+        # inside a command line
+
+        can_use_env = not force_serialize
         force_serialize = force_serialize or bool(reasons)
 
         if capture:
             reasons.append('to capture output')
         if feed:
             reasons.append('to feed input')
+
+        if can_use_env and reasons == ['to set env'] and shutil.which('env'):
+            envlist = []
+            for k, v in env.get_env({}).items():
+                envlist.append(f'{k}={v}')
+            return ['env'] + envlist + es.cmd_args, ', '.join(reasons)
 
         if not force_serialize:
             if not capture and not feed:


xclaesse:
Minor nitpicking, but I don't like that much setting `serialize_with_env` in every condition. I prefer keeping it contained in the `if es.env and es.env.varnames:` condition, what do you think? I think it should work in the capture case too, no?

I'm wondering if we should also add a `shutil.which('env')` to be sure it exists, I'm not sure all non-windows platforms always has 'env' program? Also that way we could even use env under msys2/cygwin on Windows I guess.

```patch
diff --git a/mesonbuild/backend/backends.py b/mesonbuild/backend/backends.py
index e2297a30c..c846f566a 100644
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -475,20 +475,23 @@ class Backend:
         if any('\n' in c for c in es.cmd_args):
             reasons.append('because command contains newlines')
 
+        env_cmd = []
         if es.env and es.env.varnames:
-            reasons.append('to set env')
+            if not mesonlib.is_windows():
+                env_cmd = ['env']
+                for k, v in es.env.get_env(dict()).items():
+                    env_cmd.append(f'{k}={v}')
+            else:
+                reasons.append('to set env')
 
         force_serialize = force_serialize or bool(reasons)
 
-        if capture:
-            reasons.append('to capture output')
-
         if not force_serialize:
             if not capture:
-                return es.cmd_args, ''
+                return env_cmd + es.cmd_args, ''
             return ((self.environment.get_build_command() +
-                    ['--internal', 'exe', '--capture', capture, '--'] + es.cmd_args),
-                    ', '.join(reasons))
+                    ['--internal', 'exe', '--capture', capture, '--'] + env_cmd + es.cmd_args),
+                    'to capture output')
 
         if isinstance(exe, (dependencies.ExternalProgram,
                             build.BuildTarget, build.CustomTarget)):
```

eli-schwartz:
When I first wrote it, I didn't know how many conditions it would need to cover... Then I realized it would only cover one of them.

Your suggestion is awkward as is, because if it needs to be serialized with python anyway then there's no point running env too.

shutil.which is a very good point.

xclaesse:
> Your suggestion is awkward as is, because if it needs to be serialized with python anyway then there's no point running env too.

I think the --capture case is because it saves the time of pickle from/to a file, and also it makes the command line appear in ninja instead of being hidden, that's often easier to debug.

eli-schwartz:
> also it makes the command line appear in ninja instead of being hidden, that's often easier to debug.

Not sure what this means? env doesn't let you capture anything, so it would still be `/usr/bin/meson --internal exe --capture`. But current versions of meson then pass the command line at the end, after `--`.

e.g.
```
$ cat meson.build 
project('foo', 'c')

custom_target('foo',
  output: ['foo.txt'],
  command: ['echo', 'hi'],
  capture: true
)
$ meson setup builddir/
[...]
$ ninja -v -C builddir/
ninja: Entering directory `builddir'
[1/1] /usr/bin/meson --internal exe --capture foo.txt -- echo hi
```

jpakkane:
> it is guaranteed to exist on any POSIX platform

And we must support Windows, which is not a POSIX platform. No functionality in Meson is allowed to use Unix shell tools, precisely because of this reason.

mesonbuild/backend/backends.py
```diff
@@ -572,17 +573,22 @@ def as_meson_exe_cmdline(self, exe: T.Union[str, mesonlib.File, build.BuildTarge
         cmd.extend(cmd_args)
         es = self.get_executable_serialisation(cmd, workdir, extra_bdeps, capture, feed, env, verbose=verbose)
         reasons: T.List[str] = []
+        serialize_with_env = shutil.which('env') is not None
         if es.extra_paths:
             reasons.append('to set PATH')
+            serialize_with_env = False
```
xclaesse, line 579: Can't we use `env` to set `PATH` too?

eli-schwartz: PATH is special, because we don't set it as an override but rather enhance the one that already exists. So running the ExecutableSerialisation will actually set the PATH to `extra_path:$PATH`.

This would either need to be baked in at configure time, or do what we currently do, which is calculate `os.environ['PATH']` at build time and append it.
Unfortunately, the env program isn't specialized enough to have an option that *prepends* (or appends) an environment variable to whatever its current value is. Usually people use shells for that.

Meson *could* use `sh -c 'env FOO=bar "PATH=/extra/path:$PATH" mycmd'` but at that point just use sh instead of env... Recursive sh invocations is really not a good rabbit hole to dive into (we already safe-quote arguments because ninja uses a shell to run every build edge).

xclaesse: Fair enough.

xclaesse:
LGTM, but I'm not sure if CI failures are relevant.

eli-schwartz:
I think I hit the jackpot for flaky issues in msys/macOS/wine but I don't understand the VS issue at all, especially as this change should not affect that at all.

xclaesse:
> The debug failure is because apparently the VS backend finds env.exe, which is from `C:\Program Files\Git\usr\bin\env.EXE`.

That's weird, maybe a special config of GitHub CI, because on my Windows machine that's not in PATH, which is why I did https://github.com/mesonbuild/meson/pull/10849.

> I neglected to handle force_serialize.

I never understood why we had that force_serialize on vs backend, but that explains it. While at it, please add a comment explaining this.

eli-schwartz:
> That's weird, maybe a special config of GitHub CI

Yes, github's stock actions set up a bunch of things for you but I'm slightly surprised that that includes a partial unix compatibility layer.



> I never understood why we had that force_serialize on vs backend, but that explains it. While at it, please add a comment explaining this.

Indeed, that's the plan (so that next time I don't accidentally ignore this either).

xclaesse:
> I'm slightly surprised that that includes a partial unix compatibility layer.

It even has 2, there is also the perl stuff. Those Windows images are a total mess!

eli-schwartz:
That's not really quite so surprising. Perl and git both have compat layers, and neither one is made globally available. Both have entry points just for the relevant programs, and a private directory with support software.

Perl's public directory contains one specific assassination tool. Git is well behaved.

GitHub seems to add git's private directory to the public environment.

eli-schwartz:
CI finally passes for this...

grimmy:
I just wanted to mention that this fixes a bug that was introduced in ca77adea89b3f102ca3e62465f651811516071f5 that caused the stderr output of g-ir-scanner to be eaten so you _might_ want to mention this in the 0.64.0 release notes?

PR #8569, opened 2021-03-23
Fix a file permissions issues in depfixer

trhd:
Installing a file with read-only permissions is broken when installing to an arbitrary directory.

This change makes depfixer to temporarily change file permissions of the installed file to work around this.

To reproduce the issue:

meson.build:
```
project('meson_bug', 'c',
  version : '0.1',

exe = executable('meson_bug', 'meson_bug.c',
  install : true)

touch = find_program('touch')

c = custom_target('foobar target', command : [touch, 'foobar'],
  output : 'foobar', install : true, install_mode : 'r--r--r--',
  install_dir : get_option('includedir'))

test('basic', exe)
```

meson_bug.c
```
#include <stdio.h>

#define PROJECT_NAME "meson_bug"

    if(argc != 1) {
        printf("%s takes no arguments.\n", argv[0]);
        return 1;
    }
    printf("This is project %s.\n", PROJECT_NAME);
    return 0;
}
```

Then run:
```
+ meson b
The Meson build system
Version: 0.56.1
Source dir: /home/hemmo/meson_bug
Build dir: /home/hemmo/meson_bug/b
Build type: native build
Project name: meson_bug
Project version: 0.1
C compiler for the host machine: cc (gcc 10.2.1 "cc (Debian 10.2.1-3) 10.2.1 20201224")
C linker for the host machine: cc ld.bfd 2.35.1
Host machine cpu: x86_64
Program touch found: YES (/usr/bin/touch)
Build targets in project: 2

Found ninja-1.10.1 at /usr/bin/ninja
+ DESTDIR=/home/hemmo/meson_bug/dest ninja -C b install
ninja: Entering directory `b'
[3/4] Installing files.
Installing meson_bug to /home/hemmo/meson_bug/dest/usr/local/bin
Installing foobar to /home/hemmo/meson_bug/dest/usr/local/include
  File "/usr/lib/python3/dist-packages/mesonbuild/mesonmain.py", line 140, in run
  File "/usr/lib/python3/dist-packages/mesonbuild/minstall.py", line 554, in run
    installer.do_install(datafilename)
  File "/usr/lib/python3/dist-packages/mesonbuild/minstall.py", line 372, in do_install
    self.install_targets(d)
  File "/usr/lib/python3/dist-packages/mesonbuild/minstall.py", line 527, in install_targets
    depfixer.fix_rpath(outname, t.rpath_dirs_to_remove, install_rpath, final_path,
  File "/usr/lib/python3/dist-packages/mesonbuild/scripts/depfixer.py", line 468, in fix_rpath
    fix_elf(fname, rpath_dirs_to_remove, new_rpath, verbose)
  File "/usr/lib/python3/dist-packages/mesonbuild/scripts/depfixer.py", line 369, in fix_elf
    with Elf(fname, verbose) as e:
  File "/usr/lib/python3/dist-packages/mesonbuild/scripts/depfixer.py", line 123, in __init__
    self.bf = open(bfile, 'r+b')
PermissionError: [Errno 13] Permission denied: '/home/hemmo/meson_bug/dest/usr/local/include/foobar'
Exception ignored in: <function Elf.__del__ at 0x7f204e5f0550>
  File "/usr/lib/python3/dist-packages/mesonbuild/scripts/depfixer.py", line 140, in __del__
    if self.bf:
AttributeError: 'Elf' object has no attribute 'bf'
FAILED: meson-install 
/usr/bin/meson install --no-rebuild
```

The problem can also be seen when using `--prefix` instead of `DESTDIR=`.

### DIFF
--- a/mesonbuild/scripts/depfixer.py
+++ b/mesonbuild/scripts/depfixer.py
@@ -13,8 +13,12 @@
 # limitations under the License.
 
 
-import sys, struct
-import shutil, subprocess
+import sys
+import os
+import stat
+import struct
+import shutil
+import subprocess
 import typing as T
 
 from ..mesonlib import OrderedSet
@@ -120,9 +124,9 @@ class Elf(DataSizes):
     def __init__(self, bfile: str, verbose: bool = True) -> None:
         self.bfile = bfile
         self.verbose = verbose
-        self.bf = open(bfile, 'r+b')
         self.sections = []  # type: T.List[SectionHeader]
         self.dynamic = []   # type: T.List[DynamicEntry]
+        self.open_bf(bfile)
         try:
             (self.ptrsize, self.is_le) = self.detect_elf_type()
             super().__init__(self.ptrsize, self.is_le)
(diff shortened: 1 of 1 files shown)


eli-schwartz:
I'm slightly more curious why it is triggering on a custom target that probably has no deps to fix.

Anyway, the file without write permissions is not an exe file and installing binaries with read-only permissions is probably never actually a good idea. It's very particularly a huge pain in the neck for distro packaging and at least Arch Linux will completely skip stripping or generating debuginfo packages for such binaries on the grounds that we're not working around upstream's bizarre insistence on declaring their precious files to be untouchable.

So I ready hope no one under any circumstances ever is relying on such functionality. :(

trhd:
Perhaps it should also be mentioned that in a normal case (without a custom `--prefix` or a `DESTDIR`) running `install` as a normal user would typically end in a understandable permission denied error and when run as root (e.g. via sudo) the read-only permissions don't matter and the operation will succeed.

mesonbuild/scripts/depfixer.py
```diff
@@ -120,29 +124,49 @@ class Elf(DataSizes):
...
+            try:
+                self.bf = open(bfile, 'r+b')
+            except Exception:
+                os.chmod(bfile, self.bf_perms)
+                self.bf_perms = None
+                raise e
+
+    def close_bf(self) -> None:
+        if self.bf:
+            if self.bf_perms:
```
jpakkane, line 157: For absolute correctness, this should be `if self.bf_perms is not None:`.

PR #8571, opened 2021-03-23
Fix spurious sandbox violation warning

xclaesse:
commit 366fa781fd95db80908856f6b75c3214d67b29f5

    interpreter: Fix spurious warning in include_directories()
    
    When doing include_directories('.') at the root of the subproject we
    should not warn about sandboxing violation.

commit 00020a34de5ef6b0c4d32e60a0b65a75e800ff0c

    test.json: Add support for not matching stdout lines
    
    By default expected line must be matched in order. When an expected line
    is matched it does not matter if it's matched again later or not.
    
    When defining "count", it means that line must be matched exactly that
    many times before matching the next expected line. Once all occurences
    have been matched for an expected line, it not must appear any more in
    all next lines.

commit 29f700e6b3ff0bb6d8de2142437834c9de291b55

    run_project_tests.py: Allow "--only common/240" syntax

### DIFF
--- a/mesonbuild/interpreter.py
+++ b/mesonbuild/interpreter.py
@@ -4601,7 +4601,7 @@ different subdirectory.
 ''')
             else:
                 try:
-                    self.validate_within_subproject(a, '')
+                    self.validate_within_subproject(self.subdir, a)
                 except InterpreterException:
                     mlog.warning('include_directories sandbox violation!')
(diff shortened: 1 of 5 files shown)


xclaesse:
@jpakkane this is a regression introduced by https://github.com/mesonbuild/meson/pull/8531.

Issue #8511, opened 2021-03-10
Expose `Environment::get_build_command()` to meson.build


anarazel:
Hi,

As part of the build generation I would like to generate tests / scripts that in invoke meson. One example is generating a script that runs `DESTDIR=/sometemp meson install` and then runs tests against that temporary installation. I can't just use `find_program('meson')` because meson may not be on `PATH` or the meson on `PATH` isn't the same as used in a build directory.

I obviously can just invoke `ninja install ` in such scripts / tests, but that will not work in case a different backend is used. Hence the desire to expose `get_build_command()`.

Would a patch adding that feature have a decent chance of getting accepted?

A second, less crucial, use-case for this is that I would like to generate a script that uses meson introspect + knowledge of the source-tree to compares - and optionally resyncs -  the list of tests existing in the source tree and the tests that the corresponding meson.build knows about.

`get_build_command()` effectively already is exposed via `MESONINTROSPECT` (including inside `run_command` and `add_postconf_script`, which doesn't seem to be documented). Exposing just the meson invocation as an array to meson.build seems a lot cleaner however than needing to do shell-splitting inside commands.

Regards,

Andres

jpakkane:
Seems like a reasonable addition. For some reason or another I seemed to remember that we already expose that info. Maybe we did in the past and it was removed for some reason.

behlec:
To clarify what you are asking for: Do you want to have a `MESONCOMMAND` variable like `MESONINTROSPECT` in `run_command` or do you want a command in `meson.build` that returns an array with the meson command?

anarazel:
Hi,

On Fri, Apr 2, 2021, at 08:43, behlec wrote:
> To clarify what you are asking for: Do you want to have a 
> `MESONCOMMAND` variable like `MESONINTROSPECT` in `run_command` or do 
> you want a command in `meson.build` that returns an array with the 
> meson command?

The latter.  That way no shell string parsing has to be done.

Regards,

Andres

jpakkane:
> That way no shell string parsing has to be done.

Well there is still need for some parsing logic even if it is passed as an argument. The array may contain one or two entries (I guess theoretically it could have even more than that). Making it work reliably requires code in every script that uses this.

anarazel:
> > That way no shell string parsing has to be done.
> 
[...]

PR #8572, opened 2021-03-23
Expose get_build_command() as requested in issue #8511

behlec:
Expose `get_build_command()` in meson as requested in #8511

First draft. I am not sure if the return type is as intended. Currently it is the same as of the internal function `get_build_command()`.

### DIFF
--- a/docs/markdown/Reference-manual.md
+++ b/docs/markdown/Reference-manual.md
@@ -2094,7 +2094,7 @@ the following methods.
   returns the "cross" compiler if we're currently cross-compiling and
   the "native" compiler if we're not.
 
-- `get_command()`:
+- `meson_cmd_array()`:
   *(since 0.60.0)*: returns an array which allows to run meson using run_command.
   The first element is an executable. The following elements are possible
   arguments to the executable. The result is platform specific.
--- a/docs/markdown/Reference-manual.md
+++ b/docs/markdown/Reference-manual.md
@@ -2094,6 +2094,11 @@ the following methods.
   returns the "cross" compiler if we're currently cross-compiling and
   the "native" compiler if we're not.
 
+- `get_command()`:
+  *(since 0.60.0)*: returns an array which allows to run meson using run_command.
+  The first element is an executable. The following elements are possible
+  arguments to the executable. The result is platform specific.
+
 - `get_cross_property(propname, fallback_value)`:
   *Deprecated since 0.58.0, use `get_external_property()` instead*.
   Returns the given property from a cross file, the optional fallback_value
--- a/docs/markdown/snippets/meson_cmd_array.md
+++ b/docs/markdown/snippets/meson_cmd_array.md
@@ -1,4 +1,4 @@
-## `meson.get_command()`
+## `meson.meson_cmd_array()`
 
 Returns an array with the information to call meson.
 It contains an executable as first element followed command line parameters if necessary.
--- a/docs/markdown/snippets/meson_get_command.md
+++ b/docs/markdown/snippets/meson_get_command.md
@@ -1,0 +1,5 @@
+## `meson.get_command()`
+
+Returns an array with the information to call meson.
+It contains an executable as first element followed command line parameters if necessary.
+The result is platform specific.
--- a/mesonbuild/interpreter/mesonmain.py
+++ b/mesonbuild/interpreter/mesonmain.py
@@ -59,7 +59,7 @@ class MesonMain(MesonInterpreterObject):
                              'has_external_property': self.has_external_property_method,
                              'backend': self.backend_method,
                              'add_devenv': self.add_devenv_method,
-                             'get_command': self.get_command_method,
+                             'meson_cmd_array': self.meson_cmd_array_method,
                              })
 
     def _find_source_script(self, prog: T.Union[str, mesonlib.File, build.Executable, ExternalProgram], args):
@@ -424,8 +424,8 @@ class MesonMain(MesonInterpreterObject):
             raise build.InvalidArguments(f'"add_devenv": {msg}')
         self.build.devenv.append(ENV_KW.convertor(env))
 
-    @FeatureNew('meson.get_command', '0.60.0')
+    @FeatureNew('meson.meson_cmd_array', '0.60.0')
     @noPosargs
     @noKwargs
-    def get_command_method(self, args, kwargs):
+    def meson_cmd_array_method(self, args, kwargs):
         return self.interpreter.environment.get_build_command()
--- a/mesonbuild/interpreter/mesonmain.py
+++ b/mesonbuild/interpreter/mesonmain.py
@@ -59,6 +59,7 @@ class MesonMain(MesonInterpreterObject):
                              'has_external_property': self.has_external_property_method,
                              'backend': self.backend_method,
                              'add_devenv': self.add_devenv_method,
+                             'get_command': self.get_command_method,
                              })
 
     def _find_source_script(self, prog: T.Union[str, mesonlib.File, build.Executable, ExternalProgram], args):
@@ -422,3 +423,9 @@ class MesonMain(MesonInterpreterObject):
         if msg:
             raise build.InvalidArguments(f'"add_devenv": {msg}')
         self.build.devenv.append(ENV_KW.convertor(env))
+
+    @FeatureNew('meson.get_command', '0.60.0')
+    @noPosargs
+    @noKwargs
+    def get_command_method(self, args, kwargs):
+        return self.interpreter.environment.get_build_command()
--- a/test cases/common/245 meson getcommand/meson.build
+++ b/test cases/common/245 meson getcommand/meson.build
@@ -1,0 +1,7 @@
+project('check_get_command', 'c',
+  version : '0.1',
+  default_options : ['warning_level=3'])
+
+a = meson.get_command()
+res = run_command(a, '--version', check: true)
+assert(res.stdout().strip() == '@0@'.format(meson.version()))
--- a/test cases/common/245 meson meson_cmd_array/meson.build
+++ b/test cases/common/245 meson meson_cmd_array/meson.build
@@ -1,7 +1,7 @@
-project('check_get_command', 'c',
+project('check_meson_cmd_array', 'c',
   version : '0.1',
   default_options : ['warning_level=3'])
 
-a = meson.get_command()
+a = meson.meson_cmd_array()
 res = run_command(a, '--version', check: true)
 assert(res.stdout().strip() == '@0@'.format(meson.version()))


xclaesse:
It seems a bit too niche to have a top level function IMHO. What about making this a method on the meson object? `meson.get_command()` ?

mesonbuild/interpreter.py
```diff
@@ -1911,6 +1911,7 @@ def __init__(self, build: 'build.Build', interpreter: 'Interpreter'):
                              'has_external_property': self.has_external_property_method,
                              'backend': self.backend_method,
                              'add_devenv': self.add_devenv_method,
+                             'get_command': self.get_command,
```
xclaesse, line 1914: The convention is to have _method suffix.

mesonbuild/interpreter.py
```diff
@@ -2232,6 +2233,14 @@ def add_devenv_method(self, args: T.Union[str, list, dict, EnvironmentVariablesH
             env = EnvironmentVariablesHolder(env)
         self.build.devenv.append(env.held_object)
 
+    @noPosargs
+    @permittedKwargs({'unbuffered'})
+    def get_command(self, args, kwargs):
+        unbuffered = kwargs.get('check', False)
```
xclaesse, line 2239:
```suggestion
        unbuffered = kwargs.get('unbuffered', False)
```

I'm not sure we need to expose that, what does it do?

xclaesse (commented):
Just a couple nitpickings in the implementation. To be complet, would be nice to add:
- reference manual doc
- release notes snippet
- unit test

mesonbuild/interpreter.py
```diff
@@ -2232,6 +2233,14 @@ def add_devenv_method(self, args: T.Union[str, list, dict, EnvironmentVariablesH
             env = EnvironmentVariablesHolder(env)
         self.build.devenv.append(env.held_object)
 
+    @noPosargs
+    @permittedKwargs({'unbuffered'})
+    def get_command(self, args, kwargs):
```
xclaesse, line 2238: Missing @FeatureNew('meson.get_command', '0.58.0')

mesonbuild/interpreter.py
behlec: It adds the `-u` flag to the call. I.e. `[ '/usr/bin/python3', '/foo/meson.py' ]` becomes `['/usr/bin/python3', '-u', '/foo/meson.py']`. Not sure if that is necessary. I included it, since the original method provides it. I am happy to remove that part.

mesonbuild/interpreter.py
behlec: I will add the requested parts if we are clear on whether we want to support unbuffered or not.
> * reference manual doc
> * release notes snippet
> * unit test

For the final pull request I will squash the branch, this is a small, on commit change.

behlec:
I have removed the `unbuffered` kwarg. Added documentation and a simple test.
Trying to write the test I stumbled across the problem, that I cannot compare the output of `run_command.stdout()` with `meson.version()`, since they have different types.

mesonbuild/interpreter.py
```diff
@@ -2221,6 +2222,11 @@ def add_devenv_method(self, args: T.Union[str, list, dict, EnvironmentVariablesH
             env = EnvironmentVariablesHolder(env)
         self.build.devenv.append(env.held_object)
 
+    @FeatureNew('meson.get_command', '0.58.0')
+    @noPosargs
+    def get_command_method(self, args, kwargs):
```
xclaesse, line 2227: `@noKwargs` too.

xclaesse (commented):
> Trying to write the test I stumbled across the problem, that I cannot compare the output of `run_command.stdout()` with `meson.version()`, since they have different types.

Hmmm, that seems to be a regression in the interpreter that I made when adding some special casing for version string.
```
class MesonVersionString(str):
    pass
```

As a work around for now, that works: `assert('0.57.999' == '@0@'.format(meson.version()))`.

xclaesse:
Looks good to me, could you squash please? Thanks.

jpakkane:
This is slightly different than what I got from the original bug report. If we expose something like 'MESONCOMMAND` in an env var like `MESONINTROSPECT` then would that provide for everything needed? It seems that the most common usage of this would be to pass it to external scripts so they can invoke Meson. Putting it in a common envvar would make usage simpler.

behlec:
I see what you mean. I will start working on that. However, maybe it would be a good idea if OP could comment, what he had exactly in mind?
Until we are sure, I will keep the current commit and add MESONCOMMAND to envar in a separate commit.

behlec:
First draft. If I got the idea right I will add test, documentation, and update the release snippet.

behlec:
So according to the requester of the issue he wants an `meson.build` command that returns an array.

behlec:
I have rebased the changes on master. For now I have only added `meson.get_command()`. This is what the requester had in mind. If you want I will also add `MESONCOMMAND` again.

anarazel:
> This is slightly different than what I got from the original bug report. If we expose something like 'MESONCOMMAND`in an env var like`MESONINTROSPECT` then would that provide for everything needed? It seems that the most common usage of this would be to pass it to external scripts so they can invoke Meson. Putting it in a common envvar would make usage simpler.

FWIW, I was thinking of a function like meson.get_build_command() that returns an array (like `['path/to/python', '/path/to/meson.py']`). Although `MESONCOMMAND` could also be useful. The advantage of the function would be that it's trivial to write a run_command() executing meson - whereas  `MESONCOMMAND` wouldn't be as easy to execute in a portable fashion.

anarazel:
@behlec  This is great - sorry that I missed you opening this PR.

The only substantive comment about the changes is that it might make sense to change the function name. The compiler objects have `cmd_array(): returns an array containing the command(s) for the compiler.`, so maybe it'd make sense to name this `meson.meson_cmd_array()`?

behlec:
I have renamed the command as wished.
What version will this be in? I have set it to 0.60.0 for now.

As a reminder, I do have patches to add the environment variables if wished. Let me know if I should add them.

xclaesse:
Saying that because in report https://github.com/mesonbuild/meson/issues/8511 @anarazel's first reflex was to try that, so presumably it's the most discoverable way of doing it?

eli-schwartz:
That sounds like an excellent idea, much preferred.

behlec:
Let me know what the final decision is and I will make the changes.

xclaesse:
In interpreter.program_lookup() we already special case 'python3', IMHO it makes sense to handle 'meson' there the same way.

anarazel:
@xclaesse 
> I just had an idea: what about instead special case `find_program('meson')` ?
> Saying that because in report #8511 @anarazel's first reflex was to try that, so presumably it's the most discoverable way of doing it?

It'd work for my current needs. I'm mildly worried that there could be cases where the fact that the path will likely consist out of two components would cause problems. E.g. because it's embedded in a generated script or such.

behlec:
> In interpreter.program_lookup() we already special case 'python3', IMHO it makes sense to handle 'meson' there the same way.

So `find_program('meson')` returns always the current meson location, even if it is not in PATH?
If I read the case for python3 correctly, meson searches for python3 and only returns the currently used location, if it cannot find python3 in the regular way.
@anzarel mentioned, he could not use it, because  the meson instance used might not be in PATH.

If we go the `find_program` route we need to carefully decide how to handle all the options. What happens if native is set to false? What happens if override_find_program is set?

anarazel:
I personally see the `meson_cmd_array()` approach as superior (better multi-component handling, less magic). While @xclaesse is right that I looked at find_program() for a solution, I don't think I would have if `meson_cmd_array()` existed...

eli-schwartz:
@behlec we could raise an error message in both cases.

@anarazel Hmm, how so? find_program already sometimes is an array command, e.g. smart shebang handling, required interpreters for scripts on windows where you can't run anything from cmd, that isn't an executable


It is true that this would be a bit magic. :( Ok the other hand, it folds into existing API rather than adding new API. Dunno which is better.

eli-schwartz:
Also extending find_program means that people currently using
```meson
run_target('meson', ...)
```

Will be automatically upgraded.

xclaesse:
I have to admit, I'm undecided, both solutions have pros/cons. We already have `compiler.cmd_array()` so it's try that it's not a new type of API neither. One thing is I don't like the repetition of "meson" in `meson.meson_cmd_array()`, but I think that was done on purpose to avoid confusion.

@jpakkane What's your opinion? `find_program('meson')`, `meson.cmd_array()` or `meson.meson_cmd_array()`?

Akaricchi:
I think `find_program('meson')` on its own is a bad idea. The external program object does not have a method to return the command as an array of args, it only has `full_path()` which returns a string. It's not clear how to handle commands with args there. Extending `find_program` for consistency and use cases like `run_target('meson', ...)` still seems worthwhile, however.

We can add `cmd_array()` to the external program object, and then we don't need `meson.cmd_array()`. This may be the best solution overall: most general and consistent with the compiler object.

tristan957:
@jpakkane seems like this PR stalled pending your input.

tristan957:
> I think `find_program('meson')` on its own is a bad idea. The external program object does not have a method to return the command as an array of args, it only has `full_path()` which returns a string. It's not clear how to handle commands with args there. Extending `find_program` for consistency and use cases like `run_target('meson', ...)` still seems worthwhile, however.
> 
> We can add `cmd_array()` to the external program object, and then we don't need `meson.cmd_array()`. This may be the best solution overall: most general and consistent with the compiler object.

I like this idea quite a bit. Is anyone against it?

anarazel:
> > We can add `cmd_array()` to the external program object, and then we don't need `meson.cmd_array()`. This may be the best solution overall: most general and consistent with the compiler object.
> 
> I like this idea quite a bit. Is anyone against it?

That seems like a good idea, regardless of the this issue's resolution.


> I think find_program('meson') on its own is a bad idea. The external program object does not have a method to return the command as an array of args, it only has full_path() which returns a string. It's not clear how to handle commands with args there. Extending find_program for consistency and use cases like run_target('meson', ...) still seems worthwhile, however.

I don't love magic in  `find_program()`, it feels hard to discover. What about `meson.get_meson_program()` or such? The proposed `Program.cmd_array()` could then be used to execute commands?

xclaesse:
What I like with `find_program('meson')` is it will automatically improve existing code that already do it, or do `run_target('meson', ...)`. IIRC custom_target(), run_target(), etc, already pass argv[0] through find_program() and that will already work when `meson` is in PATH. Adding the special casing will make it work in more situations. The best feature is a feature you don't even have to discover because it just works.

xclaesse:
> The external program object does not have a method to return the command as an array of args, it only has `full_path()` which returns a string. It's not clear how to handle commands with args there.

Why would you need the command array at all? Meson will already automatically expand any program returned by `find_program()` to a string array when used inside a command arguments, even when it's not used as argv[0]. I'm not sure what's the use-case here.

tristan957:
I decided to put up a PR for `find_program('meson')` at https://github.com/mesonbuild/meson/pull/11849.

tristan957:
@Akaricchi I think you could just switch to that now and get the same behavior as you have now, but also get the upgraded behavior when the PR merges.

xclaesse:
> You really should be asking why does `.full_path()` need to exist, because that's essentially a broken version of the proposed `.cmd_array()`.

That's a fair point, `.full_path()` should probably be deprecated, question is does it need to be replaced by `.cmd_array()`.

tristan957:
I don't really understand why it would need to be deprecated.

xclaesse:
Let's close, we have find_program('meson') now.

PR #8576, opened 2021-03-25
windows_proof_rmtree: Also retry os.chmod() part

xclaesse:
It looks like when Windows media scanner holds files we can't change
their permission neither.


xclaesse:
This is an attempt to fix a backtrace posted on IRC: https://paste.gnome.org/ppyuncm3q.

PR #8577, opened 2021-03-25
docs: correct log upload paths.

Flameeyes:
SInce the build directory is `builddir/` it should be
`builddir/meson-logs/testlog.txt` to be uploaded.


eli-schwartz:
Apparently I broke this in commit 1db800bf67fc80abee313381aac0528ee33103c9. :p

Thanks for noticing and correcting it!

PR #8578, opened 2021-03-25
Fix getting native dependency when doing a native build

xclaesse:
When doing a native build, `meson.override_dependency('foo', d)` is
caching the dependency for HOST machine and `dependency('foo', native:
true)` is searching into the cache for BUILD machine. It should make no
difference if `native` kwarg is omitted and when it's set to true when
doing a native build.

### DIFF
--- a/mesonbuild/interpreter.py
+++ b/mesonbuild/interpreter.py
@@ -2099,7 +2099,7 @@ class MesonMain(InterpreterObject):
         if len(args) != 1:
             raise InterpreterException('get_compiler_method must have one and only one argument.')
         cname = args[0]
-        for_machine = Interpreter.machine_from_native_kwarg(kwargs)
+        for_machine = self.interpreter.machine_from_native_kwarg(kwargs)
         clist = self.interpreter.coredata.compilers[for_machine]
         if cname in clist:
             return CompilerHolder(clist[cname], self.build.environment, self.interpreter.subproject)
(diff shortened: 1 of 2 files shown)


xclaesse:
@dcbaker I hope this makes sense. I'm a bit worried because it modifies machine_from_native_kwarg() that is used in many places, but I think it actually fix the same bug in all APIs that takes a native kwarg. What do you think?

xclaesse:
Hmm, seems I need to do the opposite, `native: true` in a native build should return HOST machine.

dcbaker:
This is definately wrong.
```
native : true == cross_compile ? build : host
native: false == host
```

There's a big mismatch between the native and internal MachineChoice, and it's unfortuante because the internal structs are much nicer to work with and reason about (internally in a host == build configuration host === build, so you can freely grab either and get what you wanted). I think it might be worth the pain of deprecating `native` and having a new `machine : "host" | "build"` keyword argument.

dcbaker:
However, the bug you're seeing might be related to something I spotted yesterday, let me send a patch and see if that helps

xclaesse:
@dcbaker changed the fix, I think we can just plainly ignore the native kwarg when doing native build, and standardize on using the HOST machine for native build.

Let's see if CI disagrees..

dcbaker:
What is the actual bug you're hitting?

xclaesse:
> What is the actual bug you're hitting?

That code fails because `foo` is not found when doing a native build:
```
meson.override_dependency('foo', declare_dependency())
dependency('foo', native: true)
```

More specifically, I noticed that when using `gnome.generate_gir()` when gobject-introspection is a subproject, because the gnome module search for `gobject-introspection-1.0` dependency with `native:true` kwarg but gobject-introspection subproject uses override_dependency() without native kwarg.

dcbaker:
subprojects are always asumed to be `native : false` right now, there's no way to have a native subproject, are we hitting a collisoin of a subproject for the host machine trying to override something for the build machine?

xclaesse:
It has nothing to do with subprojects. Just those 2 lines in the main project and configure with no cross file and you'll get the error.

dcbaker:
I know what the problem is. hold on.

xclaesse:
Ok let's close this PR, the problem is even bigger than I thought and @dcbaker has a more complete fix: https://github.com/mesonbuild/meson/pull/8582

PR #8580, opened 2021-03-25
coredata: Fix dependency cache

dcbaker:
The host cache had BUILD, which is wrong.


xclaesse:
LGTM, when CI is green.

dcbaker:
Are we expecting cygwin to fail? that doesn't look at all related.

jpakkane:
Rerunning it got green so merging.

PR #8582, opened 2021-03-25
Per-machine caches that are user accessible should use PerMachineDefaultable

dcbaker:
For structures that are not exposed to the user it is fine to have completely separate build and host structures, but for anything exposed to the user we need to use PerMachineDefaultable. This is because PerMachine is a host/build separated structure, but meson as a language doesn't expose this, it exposes a native/not native interface.

The implication of this is if a user says `native : true`, that has a different meaning in a cross compile than in a !cross compile, whereas build and host do not. So `depencency('zlib', native : true)` in a cross compile (host != build) means find zlib for the build machine, but in a native build (host == build) then it means get zlib for the host machine. To deal with this, data structures exposed to the user use a Defaultable, which in host == build configurations makes the host and build attributes references to the same object, but in a host != build configuration they are references to separate objects.

### DIFF
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -206,10 +206,21 @@ class Build:
         self.projects = {}
         self.targets: T.MutableMapping[str, 'Target'] = OrderedDict()
         self.run_target_names: T.Set[T.Tuple[str, str]] = set()
-        self.global_args: PerMachine[T.Dict[str, T.List[str]]] = PerMachine({}, {})
-        self.projects_args: PerMachine[T.Dict[str, T.List[str]]] = PerMachine({}, {})
-        self.global_link_args: PerMachine[T.Dict[str, T.List[str]]] = PerMachine({}, {})
-        self.projects_link_args: PerMachine[T.Dict[str, T.List[str]]] = PerMachine({}, {})
+
+        global_args: PerMachineDefaultable[T.Dict[str, T.List[str]]] = PerMachineDefaultable({})
+        global_link_args: PerMachineDefaultable[T.Dict[str, T.List[str]]] = PerMachineDefaultable({})
+        project_args: PerMachineDefaultable[T.Dict[str, T.List[str]]] = PerMachineDefaultable({})
+        project_link_args: PerMachineDefaultable[T.Dict[str, T.List[str]]] = PerMachineDefaultable({})
+        if environment.is_cross_build():
+            global_args.host = {}
+            global_link_args.host = {}
+            project_args.host = {}
+            project_link_args.host = {}
+        self.global_args = global_args.default_missing()
+        self.projects_args = project_args.default_missing()
+        self.global_link_args = global_link_args.default_missing()
+        self.projects_link_args = project_link_args.default_missing()
+
         self.tests: T.List['Test'] = []
         self.benchmarks: T.List['Test'] = []
         self.headers: T.List[Headers] = []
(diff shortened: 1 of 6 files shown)


dcbaker:
LGTM is kinda lying. It's still used in a bunch of type annotations, and I don't really want to rewrite all of those

test cases/common/99 subproject subdir/meson.build
```diff
@@ -30,6 +30,11 @@ assert(not d.found(), 'Dependency should be not-found')
 d = dependency('sub_implicit')
 assert(d.found(), 'Should implicitly fallback')
 
+# `native: true` should not make a difference when doing a native build.
+if not meson.is_cross_build()
+  dependency('sub_implicit', native: true)
+endif
```
xclaesse, line 36: I sneaked this here because I was too lazy to open a new test case. But now that you created one better move this there. You don't need a subproject to test override_dependency().

mesonbuild/build.py
```diff
@@ -229,7 +233,13 @@ def __init__(self, environment: environment.Environment):
         self.test_setup_default_name = None
         self.find_overrides = {}
         self.searched_programs = set() # The list of all programs that have been searched for.
-        self.dependency_overrides = PerMachine({}, {})
+
+        # If we are doing a cross build we need two caches, if we're doing a
+        # build == host compilation the both caches should point to the same place.
+        dependency_overrides: PerMachineDefaultable[T.Dict[T.Tuple, DependencyOverride]] = PerMachineDefaultable({})
```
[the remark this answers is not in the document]
dcbaker: yup, done.

test cases/common/99 subproject subdir/meson.build
dcbaker: okay, moved. Hopefully the dependency I picked works, lol

xclaesse:
Looks good to me, thanks. For the CI I think that test case needs to be skipped in cross build scenarios and in platforms that don't have zlib.

test cases/common/241 dependency native host == build/meson.build
```diff
@@ -0,0 +1,10 @@
+project('foo')
+
+dependency('zlib')
+dependency('zlib', native : true)
+dependency('zlib', native : false)
+
+# `native: true` should not make a difference when doing a native build.
+if not meson.is_cross_build()
+  dependency('expat', native: true)
```
xclaesse, line 9: oh missed that, why are you checking for expat?

```suggestion
  meson.override_dependency('something', declare_dependency())
  dependency('something', native: true)
```

mesonbuild/coredata.py
```diff
@@ -394,11 +394,16 @@ def __init__(self, options: argparse.Namespace, scratch_dir: str, meson_command:
         self.version = version
         self.options: 'KeyedOptionDictType' = {}
         self.cross_files = self.__load_config_files(options, scratch_dir, 'cross')
-        self.compilers = PerMachine(OrderedDict(), OrderedDict())  # type: PerMachine[T.Dict[str, Compiler]]
 
-        build_cache = DependencyCache(self.options, MachineChoice.BUILD)
-        host_cache = DependencyCache(self.options, MachineChoice.BUILD)
-        self.deps = PerMachine(build_cache, host_cache)  # type: PerMachine[DependencyCache]
+        self.compilers: PerMachine['T.OrderedDict[str, Compiler]'] = PerMachineDefaultable.default(
+            self.is_cross_build(), {}, {})
```
xclaesse, line 399:
```suggestion
            self.is_cross_build(), OrderedDict(), OrderedDict())
```

no?

dcbaker:
@xclaesse, @jpakkane I think I have all the bugs sorted out now. How do you feel about this for 0.58, I think it might be worth it since it's purely bug fixes.

xclaesse:
LGTM.

PR #8584, opened 2021-03-26
mtest: terminate test with SIGQUIT on timeout

tbeloqui:
This gives us the chance to generate a core dump when a test timeout

### DIFF
--- a/mesonbuild/mtest.py
+++ b/mesonbuild/mtest.py
@@ -1213,7 +1213,7 @@ class TestSubprocess:
 
         return self.stdo_task, self.stde_task
 
-    async def _kill(self) -> T.Optional[str]:
+    async def _kill(self, result: TestResult) -> T.Optional[str]:
         # Python does not provide multiplatform support for
         # killing a process and all its children so we need
         # to roll our own.
@@ -1224,7 +1224,7 @@ class TestSubprocess:
             else:
                 # Send a termination signal to the process group that setsid()
                 # created - giving it a chance to perform any cleanup.
-                os.killpg(p.pid, signal.SIGTERM)
+                os.killpg(p.pid, signal.SIGQUIT if result is TestResult.TIMEOUT else signal.SIGTERM)
 
                 # Make sure the termination signal actually kills the process
                 # group, otherwise retry with a SIGKILL.
@@ -1267,12 +1267,12 @@ class TestSubprocess:
         try:
             await complete_all(self.all_futures, timeout=timeout)
         except asyncio.TimeoutError:
-            additional_error = await self._kill()
             result = TestResult.TIMEOUT
+            additional_error = await self._kill(result)
         except asyncio.CancelledError:
             # The main loop must have seen Ctrl-C.
-            additional_error = await self._kill()
             result = TestResult.INTERRUPT
+            additional_error = await self._kill(result)
         finally:
             if self.postwait_fn:
                 self.postwait_fn()


jpakkane:
We can't really change this quickly like this, because there are people depending on the old behaviour. It is also not clear why SIGQUIT is preferable, you can catch SIGINT as far as I know.

If this is to be changed it would also require properly documenting it in the test pages. Specifically what signals are sent when and whether there are any platform specific (i.e. Windows) behavioural changes.

@bonzini you seem to be the person who changed this last. Do you have an opinion on how this should behave?

bonzini:
The difference between SIGQUIT and SIGTERM is that the former causes a core dump to be generated.

All I did was move code around; Meson has used SIGTERM since forever. Anyway, since my opinion was requested :) I don't think it should be the default. Maybe a command line option (`--kill=SIGQUIT|SIGTERM|SIGKILL|SIGINT`) would be useful, but let's take a step back and see what you're trying to achieve which is better debuggability of timed-out tests.  There is already `--gdb` for that, would it work for you?

tbeloqui:
Thanks for taking time into looking at this.

So we found out that our CI was not generating core dumps for timed-out tests and the only way we could generate them was to use SIGQUIT instead of SIGTERM. Unfortunately the `--gdb` option is not suitable for our automated environment as we want to get a core dump but not involve the debugger. 

From the libc documentation (https://www.gnu.org/software/libc/manual/html_node/Termination-Signals.html):
```The SIGQUIT signal is similar to SIGINT, except that it’s controlled by a different key—the QUIT character, usually C-\—and produces a core dump when it terminates the process, just like a program error signal. You can think of this as a program error condition “detected” by the user.```

So I think that  a timeout falls into the group of "condition detected by the user", would you agree?
Anyhow, we think the `--kill` option would be a better approach as that is less disruptive to others.

bonzini:
> So I think that a timeout falls into the group of "condition detected by the user", would you agree?

It does, but on the other hand it is generally better to let the test cleanup after itself if possible (deleting temporary files, etc.).

> So we found out that our CI was not generating core dumps for timed-out tests

I guess your tests are generating a core-dump via SIGABRT for failures?

Anyway I'll happily review a patch that implements `--kill`. It should be enough to store the signal in a global variable and modify `TestSubprocess._kill`.

The case of `--kill=SIGINT` is an interesting one. Thinking more about it, when this should not be an option for `--kill`; rather, it should be the default when `--gdb` is in use, and it should also suppress the subsequent `SIGKILL`: this will cause the program to go to gdb, where the user can decide how to handle the issue.

tbeloqui:
> It does, but on the other hand it is generally better to let the test cleanup after itself if possible (deleting temporary files, etc.).
> 
> >

Agree, I was mainly thinking on timeouts by deadlock but that is not always the case.



> I guess your tests are generating a core-dump via SIGABRT for failures?

Yes, that is working fine for us :).

> Anyway I'll happily review a patch that implements `--kill`. It should be enough to store the signal in a global variable and modify `TestSubprocess._kill`.
> 
> The case of `--kill=SIGINT` is an interesting one. Thinking more about it, when this should not be an option for `--kill`; rather, it should be the default when `--gdb` is in use, and it should also suppress the subsequent `SIGKILL`: this will cause the program to go to gdb, where the user can decide how to handle the issue.

Brilliant, I will then close this PR and consider implementing the `--kill` argument whenever we get time for it.

Thanks once again for your time and feedback.

Issue #7644, opened 2020-08-26
Instructions to run scan-build are broken


mcatanzaro:
**Describe the bug**
https://mesonbuild.com/howtox.html#use-clang-static-analyzer says that this should work:

```
$ meson compile scan-build
```

But it doesn't work. When running that command in a build directory, meson 0.55.0 prints:

```
$ meson compile scan-build
Found runner: /usr/bin/ninja

ERROR: Can't invoke target `scan-build`: target not found
```

Now, meson really has generated a usable scan-build target. Executing `ninja scan-build` works perfectly fine. I have no clue why `meson compile` doesn't see the target.

**To Reproduce**
I think this bug occurs for every meson project. I've tested epiphany and glib-networking.

**Expected behavior**
`meson compile` should be able to do a scan-build.

**system parameters**
* native build
* Fedora 32
* Python 3.8.5
* meson 0.55.0
* ninja 1.10.0

Volker-Weissmann:
Same here.

PR #8590, opened 2021-03-27
docs: Provide working instructions for scan-build

dreamer:
Replace `meson compile scan-build` with `ninja -C dir scan-build`,
because scan-build target does not work with `meson compile`.

Note about SCANBUILD env variable was not precise enough to describe how
to pass arguments to scan-build - provide an example to make it clear.

Fixes: #7644.

### DIFF
--- a/docs/markdown/howtox.md
+++ b/docs/markdown/howtox.md
@@ -211,19 +211,33 @@ test failures.
 
 ## Use Clang static analyzer
 
-Install scan-build and configure your project. Then do this:
+Install scan-build program, then do this:
 
 ```console
-$ meson compile scan-build
+$ meson setup builddir
+$ ninja -C builddir scan-build
 ```
 
 You can use the `SCANBUILD` environment variable to choose the
 scan-build executable.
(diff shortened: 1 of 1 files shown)


jpakkane:
Thanks. Note in the future that doc only changes should have `[skip ci]` at the end of their commit messages so CI is not run.

aleinon:
I'm new to meson and in windows I came across a problem using `ninja -C dir scan-build`. I get the following error:

..\meson.build:12:0: ERROR: Unknown compiler(s): [['C:/Projects/llvm-project/clang/tools/scan-build/bin/../libexec/ccc-analyzer']]
The following exception(s) were encountered:
Running "C:/Projects/llvm-project/clang/tools/scan-build/bin/../libexec/ccc-analyzer --version" gave "[WinError 193] %1 is not a valid Win32 application"

In windows, ccc-analyzer is a .bat file with contents `perl -S ccc-analyzer %*`, so it calls a script. Is there a way to make this work in windows?

PR #8597, opened 2021-03-29
wrap: Tell which wrap-redirect file is missing

seungha-yang:
Otherwise user cannot understand which wrap file is missing or wrong

### DIFF
--- a/mesonbuild/wrap/wrap.py
+++ b/mesonbuild/wrap/wrap.py
@@ -136,7 +136,7 @@ class PackageDefinition:
                 raise WrapException('wrap-redirect filename must be a .wrap file')
             fname = dirname / fname
             if not fname.is_file():
-                raise WrapException('wrap-redirect filename does not exist')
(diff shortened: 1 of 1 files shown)


mesonbuild/wrap/wrap.py
```diff
@@ -136,7 +136,7 @@ def parse_wrap(self) -> None:
                 raise WrapException('wrap-redirect filename must be a .wrap file')
             fname = dirname / fname
             if not fname.is_file():
-                raise WrapException('wrap-redirect filename does not exist')
+                raise WrapException('wrap-redirect ' + str(fname) + ' filename does not exist')
```
xclaesse, line 139: Nitpicking:

```suggestion
                raise WrapException(f'wrap-redirect {fname} filename does not exist')
```

seungha-yang: thanks for the suggestion! done

PR #8606, opened 2021-03-30
Fix _FOR_BUILD env variables

dcbaker:
How this was all working before is a mystery to me. we had
1) a test that was wrong
2) no test for the right behavior
3) then broke the right behavior to fix the test that was implementing the wrong behavior.

For reference, the correct behavior is documented here: https://mesonbuild.com/Reference-tables.html#environment-variables-per-machine

### DIFF
--- a/run_unittests.py
+++ b/run_unittests.py
@@ -9348,6 +9348,26 @@ class CrossFileTests(BasePlatformTests):
                 break
         self.assertEqual(found, expected, 'Did not find all sections.')
 
+    def test_for_build_env_vars(self) -> None:
+        testcase = os.path.join(self.common_test_dir, '2 cpp')
+        config = self.helper_create_cross_file({'built-in options': {}})
+        cross = self.helper_create_cross_file({'built-in options': {}})
+
+        self.init(testcase, extra_args=['--native-file', config, '--cross-file', cross],
+                  override_envvars={'PKG_CONFIG_PATH': '/bar', 'PKG_CONFIG_PATH_FOR_BUILD': '/dir'})
+        configuration = self.introspect('--buildoptions')
+        found = 0
+        for each in configuration:
+            if each['name'] == 'pkg_config_path':
+                self.assertEqual(each['value'], ['/bar'])
+                found += 1
+            elif each['name'] == 'build.pkg_config_path':
+                self.assertEqual(each['value'], ['/dir'])
+                found += 1
+            if found == 2:
+                break
+        self.assertEqual(found, 2, 'Did not find all sections.')
+
 
 class TAPParserTests(unittest.TestCase):
     def assert_test(self, events, **kwargs):
--- a/run_unittests.py
+++ b/run_unittests.py
@@ -9325,13 +9325,15 @@ class CrossFileTests(BasePlatformTests):
 
     def test_builtin_options_conf_overrides_env(self):
         testcase = os.path.join(self.common_test_dir, '2 cpp')
-        config = self.helper_create_cross_file({'built-in options': {'pkg_config_path': '/native'}})
-        cross = self.helper_create_cross_file({'built-in options': {'pkg_config_path': '/cross'}})
+        config = self.helper_create_cross_file({'built-in options': {'pkg_config_path': '/native', 'cpp_args': '-DFILE'}})
+        cross = self.helper_create_cross_file({'built-in options': {'pkg_config_path': '/cross', 'cpp_args': '-DFILE'}})
 
         self.init(testcase, extra_args=['--native-file', config, '--cross-file', cross],
-                  override_envvars={'PKG_CONFIG_PATH': '/bar', 'PKG_CONFIG_PATH_FOR_BUILD': '/dir'})
+                  override_envvars={'PKG_CONFIG_PATH': '/bar', 'PKG_CONFIG_PATH_FOR_BUILD': '/dir',
+                                    'CXXFLAGS': '-DENV', 'CXXFLAGS_FOR_BUILD': '-DENV'})
         configuration = self.introspect('--buildoptions')
         found = 0
+        expected = 4
         for each in configuration:
             if each['name'] == 'pkg_config_path':
                 self.assertEqual(each['value'], ['/cross'])
@@ -9339,9 +9341,12 @@ class CrossFileTests(BasePlatformTests):
             elif each['name'] == 'build.pkg_config_path':
                 self.assertEqual(each['value'], ['/native'])
                 found += 1
-            if found == 2:
+            elif each['name'].endswith('cpp_args'):
+                self.assertEqual(each['value'], ['-DFILE'])
+                found += 1
+            if found == expected:
                 break
-        self.assertEqual(found, 2, 'Did not find all sections.')
+        self.assertEqual(found, expected, 'Did not find all sections.')
 
 
 class TAPParserTests(unittest.TestCase):
--- a/mesonbuild/compilers/compilers.py
+++ b/mesonbuild/compilers/compilers.py
@@ -1236,19 +1236,26 @@ def get_global_options(lang: str,
     largkey = argkey.evolve('link_args')
     envkey = argkey.evolve('env_args')
 
+    comp_key = argkey if argkey in env.options else envkey
+
+    comp_options = env.options.get(comp_key, [])
+    link_options = env.options.get(largkey, [])
+
     cargs = coredata.UserArrayOption(
         description + ' compiler',
-        env.options.get(argkey, []), split_args=True, user_input=True, allow_dups=True)
-    # the compiler args always gets the environemtn variable arguments
-    cargs.extend_value(env.options.get(envkey, []))
+        comp_options, split_args=True, user_input=True, allow_dups=True)
 
     largs = coredata.UserArrayOption(
         description + ' linker',
-        env.options.get(largkey, []), split_args=True, user_input=True, allow_dups=True)
-    # The linker gets the compiler environment variable only if the comiler
-    # acts as the linker
-    if comp.INVOKES_LINKER:
-        largs.extend_value(env.options.get(envkey, []))
+        link_options, split_args=True, user_input=True, allow_dups=True)
+
+    if comp.INVOKES_LINKER and comp_key == envkey:
+        # If the compiler acts as a linker driver, and we're using the
+        # environment variable flags for both the compiler and linker
+        # arguments, then put the compiler flags in the linker flags as well.
+        # This is how autotools works, and the env vars freature is for
+        # autotools compatibility.
+        largs.extend_value(comp_options)
 
     opts: 'KeyedOptionDictType' = {argkey: cargs, largkey: largs}
 
--- a/mesonbuild/environment.py
+++ b/mesonbuild/environment.py
@@ -798,11 +798,7 @@ class Environment:
 
         env_opts: T.DefaultDict[OptionKey, T.List[str]] = collections.defaultdict(list)
 
-        if self.is_cross_build():
-            for_machine = MachineChoice.BUILD
-        else:
-            for_machine = MachineChoice.HOST
-        for evar, keyname in opts:
+        for (evar, keyname), for_machine in itertools.product(opts, MachineChoice):
             p_env = _get_env_var(for_machine, self.is_cross_build(), evar)
             if p_env is not None:
                 # these may contain duplicates, which must be removed, else
@@ -832,7 +828,7 @@ class Environment:
                             key = key.evolve(lang=lang)
                             env_opts[key].extend(p_list)
                     elif keyname == 'cppflags':
-                        key = OptionKey('args', machine=for_machine, lang='c')
+                        key = OptionKey('env_args', machine=for_machine, lang='c')
                         for lang in compilers.compilers.LANGUAGES_USING_CPPFLAGS:
                             key = key.evolve(lang=lang)
                             env_opts[key].extend(p_list)


jpakkane:
Let's try something different as an experiment. I'll write here how I _think_ this should have worked without looking at the  current implementation or what changes this MR has. Then you can tell me whether that is what happens here or something different.

When compiling natively:

- CFLAGS et al go to the corresponding c_args et al
- _FOR_BUILD is ignored

When cross compiling:

- if _FOR_BUILD envvars are defined, they go to `build.c_args` et al and the regular ones are ignored
- if they are not defined, then `CFLAGS` go to `build.c_args` et al

dcbaker:
Which is the same behavior that autotools and a host of other build systems have.

jpakkane:
> CFLAGS always goes to c_args regardless of whether CFLAGS_FOR_BUILD is defined

This is definitely not how it was originally. The design point of the cross file was that everything dealing with cross compilation _must_ come from that. Envvars were not put into it. Every now and then people came and complained about this but then agreed with it once the rationale was explained properly.

dcbaker:
The commit that changed this (back in 0.54) was 3a4388e51dee5e7e58f1e5ad4e60fb73b4aacf08, the request for which came from distro people (nixos in this case).

The rational was specifically "we have CFLAGS and CC for compatibility with autotools/cmake/premake/etc/etc, but our CFLAGS and CC work differently than all of those do, and it breaks distros scripting."

jpakkane:
The commit says:

> Previously in Meson, variables like `CC` effected both the host and build platforms for native builds, but the just the build platform for cross builds. Now `CC_FOR_BUILD` is used for the build platform in cross builds

Sadly there are two ways of interpreting this. It does definitely say that `CC_FOR_BUILD` is used for the native compiler but it does not say that `CC` is used for the cross compiler. It might or might not, depending on how you interpret that. The original intention always was that envvars are never used for cross, everything comes from the cross file.

As a specific example, the `debcrossgen` tool that ships with Debian parses the environment variables and puts them in the cross file. If they were also taken by Meson, then all the flags would end up in the result twice (because the envvars are not unset).

All of this is not really nice at all. If I hade dictatorial powers over the entire world, I'd burn the envvar support down to the ground and jump the resulting ashes. Sadly I can't do that so we have to make something more reasonable instead.

> The rational was specifically "we have CFLAGS and CC for compatibility with autotools/cmake/premake/etc/etc, but our CFLAGS and CC work differently than all of those do, and it breaks distros scripting."

And as far as I remember the fix we agreed on was that if the  `_FOR_BUILD` envvars are used then we use them, but nothing else. For cross compilation distros already have to create a cross file, so the flags should go there (as they  do on Debian). Having two different ways of injecting basic flags via environment variables seems fragile.

dcbaker:
Certianly, I'd agree that *not* having CFLAGS at all would be best, but worse than not having them is if they behave differently than autotools does. And the test case I wrote works in the 0.56.x series, so that's the behavior we've had. Changing it now *is* certain to break something for someone. And, I don't see any difference between CFLAGS, -Dc_args, and `[built-in options]c_args` in the cross file and CFLAGS_FOR_BUILD, -Dbuild.c_args and `[built-in options]c_args` in the native file. The load order is clear, well defined, well tested, and consistant.

My arguments for keeping the current behavior then is:
1) changing behavior in a bug fix release that isn't explicitly a bug is annoying
2) distros are relying on this (at least nixos is)
3) having consistancy a cross vs native build is less surprising than not
4) we added env vars for autotools compatibility, having them behave differently than autotools is worse than not having them

run_unittests.py
```diff
@@ -7971,17 +7971,6 @@ def setUp(self):
...
-    def test_cflags_cross_environment_pollution(self):
-        '''
-        Test that the CFLAGS environment variable does not pollute the cross
-        environment. This can't be an ordinary test case because we need to
-        inspect the compiler database.
-        '''
-        testdir = os.path.join(self.common_test_dir, '3 static')
-        self.init(testdir, override_envvars={'CFLAGS': '-DBUILD_ENVIRONMENT_ONLY'})
-        compdb = self.get_compdb()
-        self.assertNotIn('-DBUILD_ENVIRONMENT_ONLY', compdb[0]['command'])
```
jpakkane, line 7983: Waitaniminute. This test actually proves that `CFLAGS` were _not_  put in cross compilation flags earlier. If they had been then this test would have failed. This test has been around since 2017 so that should be established behaviour that people probably depend on.

dcbaker:
```console
git checkout 0.56.2
cd test\ cases/common/02\ cpp
CXXFLAGS="-m32" ../../../meson.py setup build --cross-file cross.ini
../../../meson.py configure build | grep cpp_args
```

cross.ini
```ini
[binaries]
cpp = 'g++'

[host_machine]
system = 'linux'
endian = 'little'
cpu_family = 'x86'
cpu = 'i686'
```

```
  Compiler options        Current Value                                                                                                                                                                                                                                                                                      >
  ----------------        -------------                                                                                                                                                                                                                                                                                      >
  cpp_args                [-m32]                                                                                                                                                                                                                                                                                             >
  cpp_debugstl            false                                                                                                                                                                                                                                                                                              >
  cpp_eh                  default                                                                                                                                                                                                                                                                                            >
  cpp_link_args           [-m32]                                                                                                                                                                                                                                                                                             >
  cpp_rtti                true                                                                                                                                                                                                                                                                                               >
  cpp_std                 none                                                                                                                                                                                                                                                                                               >
                                                                                                                                                                                                                                                                                                                             >
  build.cpp_args          []                                                                                                                                                                                                                                                                                                 >
  build.cpp_debugstl      false                                                                                                                                                                                                                                                                                              >
  build.cpp_eh            default                                                                                                                                                                                                                                                                                            >
  build.cpp_link_args     []                                                                                                                                                                                                                                                                                                 >
  build.cpp_rtti          true                                                                                                                                                                                                                                                                                               >
  build.cpp_std           none   
```

dcbaker: I *think* that test was succeeding before because the cross file has precedence, and was overwriting the CFLAGS values.

dcbaker: No, I think that's a bad idea. If we're not going to maintain autotools compatibility (ie, CFLAGS is always for the HOST and CFLAGS_FOR_BUILD is always for the BUILD machine) then we should delete them and not have env var support, full stop. Having them behave differently is terrible UI, and just asking for a ton of bug reports. Otherwise, distros are just going to say "this is annoying", and patch meson to behave the way they expect, and users are going to complain bitterly that "this doesn't work", or "this works fine for me on distro X, don't know distro Y must just be broken".

@eli-schwartz, @Ericson2314, as distro people: opinions?

dcbaker: I just realized laying here trying to fall asleep the test i right. There's another big in my implementation... I'll fix it tomorrow.

Ericson2314:
Catching up on this thread, but IIRC the behavior we have is mostly right (in my eyes), but we got there somewhat indirectly via bugfixes and things.

Ericson2314:
Yes I agree extremely strongly with what @dcbaker is saying.

Additionally I can confirm/add

 - We do in fact rely on this stuff for NixOS / Nixpkgs. A lot -- in that a number of issues were stuck open until CC and CC_FOR_BUILD worked independently. This is one for example: https://github.com/NixOS/nixpkgs/issues/58831.
 - Since the variables' usage changed, I don't recall any complaints from users in the issue tracker here.

run_unittests.py
jpakkane: Testing this we find that the current master behaves differently and CFLAGS are not used in cross compilation. Bisection points, as expected, to  2cd0723c42ae2076c7ba2.

Weirdly if I run the test by hand and have things in the cross file's `c_args` the envvar is _still_ put there, even for old versions. I don't understand how the test could pass previously.

jpakkane:
> The bottom line, @jpakkane, is that if we do what you propose, Meson will go from being the best build system for packages to work distro-wide cross compilation in Nixpkgs to being pretty broken.

Can you specify which of the steps I listed above for the behaviour is the one that causes you problems? (Note that this is specifically for CFLAGS, not for CC.)

Ericson2314:
> (Note that this is specifically for CFLAGS, not for CC.)

I'm a bit confused by this, because aren't we talking about the policy for env vars and config files in general? Currently `_FOR_BUILD` or no suffix has a consistent meaning whether it's `CC`, `CFLAGS`, `PKG_CONFIG_PATH` or anything else.

> Can you specify which of the steps I listed above for the behaviour is the one that causes you problems?

>   if _FOR_BUILD envvars are defined, they go to build.c_args et al and the regular ones are ignored
>   if they are not defined, then CFLAGS go to build.c_args et al

So we don't actually use `CFLAGS` that much, because we've gotten to the habit of sneaking flags to the compiler behind buiuld system's back unfortunately. (I would like to get away from that for well behaved build systems like Meson, but that's another topic :).) but we do use `PKG_CONFIG_PATH` and `PKG_CONFIG_PATH_FOR_BUILD`.

We never try to put a pkg config path in native or cross files, but
https://github.com/NixOS/nixpkgs/blob/d600f006643e074c2ef1d72e462e218b647a096c/pkgs/build-support/pkg-config-wrapper/setup-hook.sh#L16 this line here will append to `PKG_CONFIG_PATH` or `PKG_CONFIG_PATH_FOR_BUILD` (based on `role_post`) the pkg-config dirs of every dependency (each of with is installed to its own prefix) if it exists. Our dependencies are sorted into build and host, so we very much expect all the build ones to end up in `PKG_CONFIG_PATH_FOR_BUILD` and then `build.pkg_config_path`, and all the host ones to end up in `PKG_CONFIG_PATH` and then `pkg_config_path`.

In fact we have something called `strictDeps` that will enforce this build vs host separation in native builds too. This is like opting into "always cross compiling" even when build = host. We like to use this where possible to ensure that we have all the proper native and foreign deps *correctly categorized* so the cross build is less likely to regress while the native still works.

run_unittests.py
eli-schwartz: I'm not completely positive what's being suggested here. Remove $CFLAGS support in cross builds? Remove $CFLAGS support in native builds "because it doesn't work in cross builds"?

jpakkane: Removing CFLAGS in cross builds because those should come from the cross file. This is how it was designed and even had tests (the one removed in this MR) to ensure that CFLAGS do not go to cross compilations. However it seems that it was broken somehow and at some point some change made the flags bleed to cross compilations.

This is bad, because for example on Debian the cross file is created on the fly when package building starts. With this behaviour every flag in CFLAGS et al is passed to the compiler twice: once from the cross file and once from the envvar.

jpakkane:
What is still unclear is that above you say this

> The bottom line, @jpakkane, is that if we do what you propose, 

but then you don't say which one of the changes would break things and how. Could you be more specific?

dcbaker:
So, just to clear things up. That test was fine, there was a bug in my patches. Here's what was happening, and what my bug was:
1) (before my patch) if the cross file specifics `c_args`, this has precedent and $CFLAGS would be ignored (as would -Dc_args)
2) my patch caused `c_args` (either from the machine file or from the command line) to be *combined*, which is why the test suddenly started failing, which is a legitimate bug.

So, the behavior change was from my patches (the previously merged ones) and was a new bug.

Ericson2314:
@jpakkane 

Sorry let me spell it out. With your changes, in a cross build I understand:

 - `PKG_CONFIG_PATH_FOR_BUILD` would effect `build.pkg_config_path`
 - `PKG_CONFIG_PATH` would do nothing or also effect `build.pkg_config_path` (based on whether `PKG_CONFIG_PATH_FOR_BUILD` was defined)

The results of this would be

 - run-time / host dependencies don't end up in `pkg_config_path` where they are needed
 - run-time / host dependencies might incorrectly end up in `build.pkg_config_path` where they could conflict with the proper build-time deps

Either of these will probably break the build.

run_unittests.py
dcbaker: No, it's not. CFLAGS and machine files are exclusive. If you pass a Machine file with `[built-in options]:c_args` then $CFLAGS will be ignored. *That* was broken, and was why the test was failing. I've put the test back and fixed the implentation such that you only get one or the other.

jpakkane: Weird, because I'm fairly sure I tried that with old releases and it did not work. Let me recheck.

jpakkane: I could not reproduce it. However upon verification I did discover one discrepancy. If `c_link_args` is not defined in the cross file but `c_args` is, then `CFLAGS` are not added to `c_link_args` whereas it used to be added. This is a good example of why having multiple ways of defining something is not great, because there are so many ways they can be combined that inconsistent behaviour like this appears.

So I'm not really sure what should be done here. Breaking visible behaviour in a point release should be avoided, but on the other hand this is starting to get way too complicated altogether. Suggestions welcome.

jpakkane:
Can you test if this mr would work in your cases or not (ignoring the one flag issue mentioned above)? This change seems to be fairly high risk and it would be nice to get some outside confirmation.

nirbheek:
FWIW, this PR as-is looks fine to me for the stable branch since it's purely a bugfix (`_FOR_BUILD` flags that were supposed to be picked up aren't, and there is no case in which someone might reasonably rely on that behaviour).

If we want to stop picking up env flags in some cases, that is something that should happen in the next release. I am inclined to pull this into the stable branch and make the release this weekend. Please holler if you think that's a bad idea.

dcbaker:
I'm in favor of Nirbheek's plan. Let's at the very least keep the behavior in the 0.57.x branch consistent with the .56, .55, and .54 branches, and discuss changes to the behavior for 0.58.0

jpakkane:
Let's go with that then.

Ericson2314:
@jpakkane I could test this PR if you like, but since it is a fairly small change with respect to the status quo --- a bug fix, as everyone seems to agree now --- that I am quite confident it will be no problem for Nixpkgs.

The bug fix if anything, by making the `_FOR_BUILD` ones more reliability picked up, make Nix work better. (If a `_FOR_BUILD` wasn't getting passed, Meson would probably just autodetect the same thing anyways.)

PR #8607, opened 2021-03-31
wrap: Use Linux path separator even on Windows for wrap-redirect

seungha-yang:
A single meson project can be shared between Windows and Linux.
It might not be a rare case for WSL based development environment.
So, to cover both Linux and Windows, we can use Linux path separator
since it's valid on Windows as well.


mesonbuild/wrap/wrap.py
```diff
@@ -294,10 +294,19 @@ def resolve(self, packagename: str, method: str, current_subproject: str = '') -
...
+                    # on Windows can be directly used on Linux as well,
+                    # WSL for example.
+                    if mesonlib.is_windows():
+                        filename = filename.replace('\\', '/')
+
                     f.write(textwrap.dedent('''\
                         [wrap-redirect]
                         filename = {}
-                        '''.format(os.path.relpath(self.wrap.filename, self.subdir_root))))
+                        '''.format(filename)))
```
xclaesse, line 309: You lose the path being relative to `self.subdir_root`. I would use `Path(filename).relative_to(self.subdir_root).as_posix()`.

seungha-yang: done!

mesonbuild/wrap/wrap.py
```diff
@@ -29,6 +29,7 @@
 import textwrap
 
 from pathlib import Path
+from pathlib import PurePath
```
xclaesse, line 32: Any reason for using PurePath instead of just Path?

mesonbuild/wrap/wrap.py
```diff
@@ -294,10 +295,11 @@ def resolve(self, packagename: str, method: str, current_subproject: str = '') -
                 # Write a dummy wrap file in main project that redirect to the
                 # wrap we picked.
                 with open(main_fname, 'w') as f:
+                    filename = PurePath(os.path.relpath(self.wrap.filename, self.subdir_root)).as_posix()
```
xclaesse, line 298: While using pathlib, you can also replace os.path.relpath() by Path().relative_to().

PR #8610, opened 2021-04-01
dependencies: Check for pthread.h

xclaesse:


dcbaker:
Pretty sure that's going to break, since threads could be pthreads or winthreads

xclaesse:
I also does not work because it fallback to check using cmake that does not check for headers.

jon-turney:
See also #553 et al

PR #8611, opened 2021-04-02
modules/fs: support FileOrString arguments

### LINKED ISSUE #8608: Support fs module methods for File objects
**Describe the bug**

Nearly all functions provided by the fs module don't support `File` objects created with the builtin `files()` they all operate only on strings.
I have the following use case where is would be great if the fs methods would also accept `File` as argument type:

We would like to structure our test cases in a one directory per submodule tree rather than in a single flat directory.

	tests/ - for generic tests
	tests/io/ - tests for the io subsystem
	tests/utils/ - tests for utility code

A test would be defined as a dictionary and appended to a global list.
This list would be used to build the actual test objects.
All source files used to build a test would be an array of `File` objects created with `files()` in the test dictionary.

Furthermore we have the convention that a test's name equals its main source file without a suffix.

To prevent duplication of names for all test cases like:

```meson
tests = [{'sources': files('foo.cpp'), 'name': 'foo'}]

foreach test_dict: tests
	sources = test_dict['sources']
	name = test_dict['name']
	exe = executable(test_name, sources, ...)
	test(test_name, exe, ...)
endforeach
```

I would like to calculate the test name from its sources:

```meson

tests = [{'sources': files('foo.cpp')]

foreach test_dict: tests
	sources = test_dict['sources']
	name = fs.stem(sources[0])
	exe = executable(test_name, sources, ...)
	test(name, exe, ...)
endforeach
```

Unfortunately `fs.stem` expects a `str` argument and I don't know of a way to convert a `File` object into a `str` in meson.

To solve this we could either introduce a way to convert a `File` object to `str`.
Or we could extend the fs methods to take `FileOrString` arguments and transparently convert a `File` argument to `str`.

**To Reproduce**

foo.cpp:

```c++
int main() {
    return 0;
}
```

meson.build

```meson
project('fs_files_reproducer', 'cpp')                                           

fs = import('fs')

f = files('foo.cpp')
executable(fs.stem(f[0]), f)
```

**Expected behavior**

Either

	executable(fs.stem(f[0]), f)

or

	executable(fs.stem(f[0]).to_string(), f)

should work.

I am willing to implement both approaches if this is something meson wants
and there is no obvious solution I am unaware of.

**system parameters**
* Arch Linux
* meson 0.57.1

fischerling:
With this change File objects created with the builtin files() function
can be used with the fs submodule like normal strings.

All methods that seem reasonable support FileOrSting arguments.
For example fs.exists() still only takes str arguments because meson
already ensures that File objects do exist when creating them with files().

Te test cases for fs are extended appropriately with tests for File objects.

Fixes #8608.

### DIFF
--- a/mesonbuild/modules/fs.py
+++ b/mesonbuild/modules/fs.py
@@ -39,13 +39,15 @@ class FSModule(ExtensionModule):
         super().__init__(interpreter)
         self.snippets.add('generate_dub_file')
 
-    def _absolute_dir(self, state: 'ModuleState', arg: str) -> Path:
+    def _absolute_dir(self, state: 'ModuleState', arg: 'FileOrString') -> Path:
         """
         make an absolute path from a relative path, WITHOUT resolving symlinks
         """
+        if isinstance(arg, File):
+            return Path(arg.absolute_path(state.source_root, self.interpreter.environment.get_build_dir()))
         return Path(state.source_root) / Path(state.subdir) / Path(arg).expanduser()
 
-    def _resolve_dir(self, state: 'ModuleState', arg: str) -> Path:
+    def _resolve_dir(self, state: 'ModuleState', arg: 'FileOrString') -> Path:
         """
         resolves symlinks and makes absolute a directory relative to calling meson.build,
         if not already absolute
@@ -59,7 +61,7 @@ class FSModule(ExtensionModule):
             pass
         return path
 
-    def _check(self, check: str, state: 'ModuleState', args: T.Sequence[str]) -> ModuleReturnValue:
+    def _check(self, check: str, state: 'ModuleState', args: T.Sequence['FileOrString']) -> ModuleReturnValue:
         test_file = self._resolve_dir(state, args[0])
         val = getattr(test_file, check)()
         if isinstance(val, Path):
@@ -74,9 +76,11 @@ class FSModule(ExtensionModule):
 
     @noKwargs
     @FeatureNew('fs.is_absolute', '0.54.0')
-    @typed_pos_args('fs.is_absolute', str)
-    def is_absolute(self, state: 'ModuleState', args: T.Tuple[str], kwargs: T.Dict[str, T.Any]) -> ModuleReturnValue:
-        return ModuleReturnValue(PurePath(args[0]).is_absolute(), [])
+    @typed_pos_args('fs.is_absolute', (str, File))
+    def is_absolute(self, state: 'ModuleState', args: T.Tuple['FileOrString'], kwargs: T.Dict[str, T.Any]) -> ModuleReturnValue:
+        if isinstance(args[0], File):
+            FeatureNew('fs.is_absolute_file', '0.59.0').use(state.subproject)
+        return ModuleReturnValue(PurePath(str(args[0])).is_absolute(), [])
 
     @noKwargs
     @FeatureNew('fs.as_posix', '0.54.0')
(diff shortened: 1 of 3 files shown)


mesonbuild/modules/fs.py
```diff
@@ -39,13 +39,15 @@ def __init__(self, interpreter: 'Interpreter') -> None:
         super().__init__(interpreter)
         self.snippets.add('generate_dub_file')
 
-    def _absolute_dir(self, state: 'ModuleState', arg: str) -> Path:
+    def _absolute_dir(self, state: 'ModuleState', arg: 'FileOrString') -> Path:
```
dcbaker, line 42: you need to have a `FeatureNew()` check for files as well, it needs to be on each entry point so the name printed is correct, not on the helper.

dcbaker (changes requested):
I have one set of changes that I think need to be made here before this could land. I haven't thought through whether this is a good idea. Maybe someone else will do that as I'm exhuasted today :)

mesonbuild/modules/fs.py
fischerling: I am sorry but I do not fully understand what you mean this may be caused by my lack of knowledge of the meson code base.

So please correct me if I am wrong!

You would like a `@FeatureNew` check for each fs method newly accepting a File argument?

Something like:

```python
@noKwargs
@typed_pos_args('fs.is_symlink_file', File)
@FeatureNew('fs.is_symlink_file', '0.58.0')
def is_symlink_file(self, state: 'ModuleState', args: T.Tuple[File], kwargs: T.Dict[str, T.Any]) -> ModuleReturnValue:
    p = Path(arg.absolute_path(state.source_root, self.interpreter.environment.get_build_dir()))
    return ModuleReturnValue(p.is_symlink(), [])

@noKwargs
@typed_pos_args('fs.is_symlink_str', str)
def is_symlink_str(self, state: 'ModuleState', args: T.Tuple[str], kwargs: T.Dict[str, T.Any]) -> ModuleReturnValue:
    return ModuleReturnValue(self._absolute_dir(state, args[0]).is_symlink(), [])

@noKwargs
@typed_pos_args('fs.is_symlink', (str, File))
def is_symlink(self, state: 'ModuleState', args: T.Tuple['FileOrString'], kwargs: T.Dict[str, T.Any]) -> ModuleReturnValue:
    if isistance(args[0], File):
        return self.is_symlink_file(args[0])
    return self.is_symlink_str(args[0])
```

Without modifying `_absolute_dir` at all ?

But this seems really ugly and it would introduce new user visible functions.
Is there a way easier to use the `@FeatureNew` decorator for only a subset of accepted argument types?

And how I understand the code the helper has no check at all except the mypy hints and relies
on the `@typed_pos_args` checks before each entry function, which we have.

mesonbuild/modules/fs.py
```diff
@@ -74,9 +76,9 @@ def expanduser(self, state: 'ModuleState', args: T.Tuple[str], kwargs: T.Dict[st
 
     @noKwargs
     @FeatureNew('fs.is_absolute', '0.54.0')
-    @typed_pos_args('fs.is_absolute', str)
-    def is_absolute(self, state: 'ModuleState', args: T.Tuple[str], kwargs: T.Dict[str, T.Any]) -> ModuleReturnValue:
-        return ModuleReturnValue(PurePath(args[0]).is_absolute(), [])
+    @typed_pos_args('fs.is_absolute', (str, File))
```
fischerling, line 79: I am not sure if it is reasonable for this method to take a File object.
This should be rechecked before a possible merge.

mesonbuild/modules/fs.py
xclaesse: You don't need to make 2 separate methods:
```meson
@noKwargs
@typed_pos_args('fs.is_symlink', (str, File))
def is_symlink(self, state: 'ModuleState', args: T.Tuple['FileOrString'], kwargs: T.Dict[str, T.Any]) -> ModuleReturnValue:
    if isistance(args[0], File):
        FeatureNew('fs.is_symlink_file', '0.58.0').use(state.subproject)
    return self._absolute_dir(state, args[0]).is_symlink()
```

fischerling: Thanks @xclaesse.
My python foo was not good enough to fully understand the FeatureNew code.
I thought of it only as a function wrapper.
Now there are FeatureNew checks for each user facing function taking a File object.

fischerling:
I everyone is ok with this changes I would squash the commits.

jpakkane:
@dcbaker what do you think now?

PR #8617, opened 2021-04-05
Fix using a cross and native file at the same time

dcbaker:
Currently, there are two bugs (one reported and one I found by inspection):

1) If a native file contains a `[project options]` section, and is used in a cross compile, then it's options will be loaded. This is a bug, only the cross file's `[project options]` should be used
2) If a native file is used in a cross compile, if it defines builtin options that are not allowed for the build machine, then those options will cause an uncaught exception in subprojects.

@nirbheek, I think these are probably worth fixing for 0.57.2 as they're regressions in the 0.57 branch, but I also don't want to hold up the release forever.


jpakkane:
LGTM.

PR #8619, opened 2021-04-06
docs: document bool to_int and to_string methods

emersion:
This documents the bool methods to_int and to_string implemented in
[1].

[1]: https://github.com/mesonbuild/meson/blob/a9e9b7c7501a3c8a5984a93879d1f309bf8c72aa/mesonbuild/interpreterbase.py#L1109

### DIFF
--- a/docs/markdown/Syntax.md
+++ b/docs/markdown/Syntax.md
@@ -93,6 +93,14 @@ A boolean is either `true` or `false`.
 truth = true
 ```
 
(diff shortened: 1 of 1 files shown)


xclaesse:
Note that it was already documented here: https://mesonbuild.com/Reference-manual.html#boolean-object. But I agree it needs to be mentioned in both places. Thanks for the patch!

PR #2611, opened 2017-11-11
Several fixes to pkg-config dependencies and the pkg-config module (try 3)

nirbheek:
`dependency(..., static: true)` had little effect when using pkg-config dependencies because we still passed `-lfoo -lbaz` to the compiler which with gcc meant that you would use shared libraries, and with MSVC meant that we would convert that to `foo.lib baz.lib` which meant that we would use import libraries (when linking to projects built with Meson) or possibly static libraries (those that were not built with Meson, because we name them `libfoo.a`).

With these changes, when `static: true` is set, we parse the pkg-config output, and if it contains `-L` fields, we use `cc.find_library`'s manual library searching feature to find the actual location of the library and use it directly.

These changes also improve `cc.find_library`'s manual searching to only search for platform-specific libraries. Earlier, it would find `libfoo.dylib` as a valid file on non-macOS platforms.

You can now also specify what type of library you want to find. For instance, you can specify that you only want static libraries or only shared libraries or prefer static over shared, etc. This functionality is currently not exposed to build files because it only works when `extra_dirs` is specified. However, I can see it being useful in the future because this is an oft-asked-for feature among autotools and libtool users.

We now also use `shlex.split` to parse `Libs:` and `Cflags:` in pkg-config files so that we properly parse them when `libdir`, `prefix`, etc contain spaces. I also found that our pkg-config generator does not properly escape paths with spaces (`prefix`, `libdir`, etc), which is [required by all implementations](https://github.com/pkgconf/pkgconf/issues/153), so I also fixed that.

Overall, pkg-config usage on Windows should be much better with these commits.

PR #8623, opened 2021-04-07
pkgconfig: Do not escape custom variables

xclaesse:
We need to escape space in variables that gets into cflags or libs
because otherwise we cannot split compiler args when paths contains
spaces. But custom variables are unlikely to be path that gets used in
cflags/libs, and escaping them cause regression in GStreamer that use
space as separator in a list variable.

### DIFF
--- a/docs/markdown/Pkgconfig-module.md
+++ b/docs/markdown/Pkgconfig-module.md
@@ -55,12 +55,20 @@ keyword arguments.
   `includedir` are reserved and may not be used. *Since 0.56.0* it can also be a
   dictionary but ordering of Meson dictionaries are not guaranteed, which could
   cause issues when some variables reference other variables.
+  Spaces in values are escaped with `\`, this is required in the case the value is
+  a path that and is used in `cflags` or `libs` arguments. *Since 0.59.0* if
+  escaping is not desired (e.g. space separate list of values) `unescaped_variables`
+  keyword argument should be used instead.
+- `uninstalled_variables` used instead of the `variables` keyword argument, when
+  generating the uninstalled pkg-config file. Since *0.54.0*
+  Spaces in values are escaped with `\`, this is required in the case the value is
+  a path that and is used in `cflags` or `libs` arguments. *Since 0.59.0* if
+  escaping is not desired (e.g. space separate list of values)
+  `unescaped_uninstalled_variables` keyword argument should be used instead.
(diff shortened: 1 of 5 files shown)


xclaesse:
CC @nirbheek who did that change in https://github.com/mesonbuild/meson/pull/2611.

xclaesse:
For reference, here is GStreamer bug report: https://gitlab.freedesktop.org/gstreamer/gst-plugins-base/-/issues/884.

tp-m:
Question is if this will then break other users that put directory/file paths into custom variables and expect those to be escaped? (Who does the unescaping then?)

We seem to have paths in custom variables too, but they're ultimately prefix-based
```
toolsdir=${exec_prefix}/bin
pluginsdir=${prefix}/lib/x86_64-linux-gnu/gstreamer-1.0
pluginscannerdir=${libexecdir}/gstreamer-1.0
```
(in our case anyway)

xclaesse:
Spaces in variables is only a problem when you want to split them on space using `shlex.split()`. For example if you do:
```
mypath=/home/Xavier Claessens
Cflags: -I${mypath}/include -pthread
```

`pkg-config --cflags` would print `-I/home/Xavier Claessens/include -pthread`, and meson would then use `shlex.split()` on that and it gives `['-I/home/Xavier', 'Claessens/include', '-pthread']` which is wrong. In that case the variable must be escaped so `pkg-config --cflags` would print `-I/home/Xavier\ Claessens/include -pthread` which gets split as `['-I/home/Xavier Claessens/include', '-pthread']`.

The common case is to have custom variables for path to a binary, like `myapp=/home/Xavier Claessens/app`, in that case it's not a problem because even if someone is expecting that to be escaped, they would have code that replace `\ ` back to ` ` which will be noop now.

So this is not backward compatible for projects who use custom variable inside `cflags:` or `libs:` fields like above... I hope...

nirbheek:
We could add a new kwarg that specifies whether the value is a path or not?

tp-m:
> We could add a new kwarg that specifies whether the value is a path or not?

or a kwarg that provides a list of variables that should not be escaped?

xclaesse:
> or a kwarg that provides a list of variables that should not be escaped?

Done that: `unescaped_variables` and `unescaped_uninstalled_variables`.

jpakkane:
LGTM, test failure is unrelated.

PR #8625, opened 2021-04-07
Add missing rpath when using link_whole

xclaesse:
A static library we link_whole could itself depend on shared libraries
that needs to be added to our rpath.


xclaesse:
This needs a unit test, and as far as I can tell there are other places we recurse only on `self.link_targets` and could be missing `self.link_whole_targets`.

jpakkane:
With a test this can be merged. It will make things better than currently, other issues can be fixed if and when they occur.

tristan957:
@xclaesse got time for a test here?

PR #8627, opened 2021-04-08
Revert meson.build_root() and meson.source_root() deprecation

emersion:
There are valid use-cases for these functions.

References: https://github.com/mesonbuild/meson/pull/7772

### DIFF
--- a/docs/markdown/Reference-manual.md
+++ b/docs/markdown/Reference-manual.md
@@ -1986,17 +1986,17 @@ the following methods.
   or `xcode`.
 
 - `build_root()`: returns a string with the absolute path to the build
-  root directory. *(deprecated since 0.56.0)*: this function will return the
-  build root of the parent project if called from a subproject, which is usually
-  not what you want. Try using `current_build_dir()` or `project_build_root()`.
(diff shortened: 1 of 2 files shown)


dcbaker:
I'm still in favor of not un-deprecating and replacing them with the `top_(source|build)_root` functions. These functions have been the source of a *ton* of bugs because they don't do what people think they do.

jpakkane:
Yes. Maybe `project_source_root` and `global_source_root`?

xclaesse:
I like global_source_root.

tp-m:
Random idea: How about keeping the existing names but adding kwargs for global/project, with no kwarg defaulting to current behaviour. We could then warn if no kwarg is specified or the 'wrong' kwarg is specified. That provide a nicer upgrade path.

New prefixed names would be more consistent with other funcs though I guess.

dcbaker:
I definitely prefer global and project, then the semantics match the other global and project methods/functions.

xclaesse:
Opened https://github.com/mesonbuild/meson/pull/8629.

emersion:
Thanks, looks good! Closing since this PR has been superseded.

PR #8629, opened 2021-04-09
Add global_source/build_root()

xclaesse:


emersion:
LGTM!

eli-schwartz:
Should the FeatureDeprecated for `meson.source_root()` etc. be updated to also mention this in an either/or?

eli-schwartz (approved):

xclaesse:
> Should the FeatureDeprecated for `meson.source_root()` etc. be updated to also mention this in an either/or?

Good point, I updated to `use meson.project_source_root() or meson.global_source_root() instead.`.

dcbaker (approved):
LGTM, thanks for doing this

PR #8643, opened 2021-04-12
Add MONO_PATH when running C# executables under Linux or MacOS

steffen-kiess:
C# executables do not have an rpath. Therefore, when building custom
targets or running tests, MONO_PATH has to be set on Linux or MacOS to
make sure that libraries are found. Under windows this is achieved by
setting PATH (which is used for binaries, native libraries and C#
libraries).

### DIFF
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -137,7 +137,7 @@ class SubdirInstallData(InstallDataBase):
 
 class ExecutableSerialisation:
     def __init__(self, cmd_args, env: T.Optional[build.EnvironmentVariables] = None, exe_wrapper=None,
-                 workdir=None, extra_paths=None, capture=None) -> None:
+                 workdir=None, extra_paths=None, extra_mono_paths=None, capture=None) -> None:
         self.cmd_args = cmd_args
         self.env = env
         if exe_wrapper is not None:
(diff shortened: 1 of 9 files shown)


steffen-kiess:
I've pushed an updated version which should also work on Windows. Note that on Windows it sets the `DEVPATH` environment variable and will only work with an application config file like this:

```
<configuration>
  <runtime>
    <developmentMode developerInstallation="true"/>
  </runtime>
</configuration>
```
(See <https://docs.microsoft.com/en-us/dotnet/framework/configure-apps/file-schema/runtime/developmentmode-element>.)

I'm not sure whether there is any better alternative on windows. Other possibilities would be to automatically create a application config with either:
```
<configuration>  
   <runtime>  
      <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">  
         <probing privatePath="some/path"/>  
      </assemblyBinding>  
   </runtime>  
</configuration>  
```
or:
```
<configuration>
  <runtime>
    <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
      <dependentAssembly>
        <assemblyIdentity name="libb" />
        <codeBase href="some/path/libb.dll" />
      </dependentAssembly>
    </assemblyBinding>
  </runtime>
</configuration>
```
but both solutions will only work if the libraries are in some subdirectory of the directory the application is in.

DEVPATH works even if the application is in some other directory, but requires `developmentMode` to be set. When `developmentMode` is set but DEVPATH is not set, the application does not seem to start at all (terminates without any error message).

PR #8654, opened 2021-04-13
OpenMP: do a compile test before declaring the dependency found

dcbaker:
Our dependency is pretty fragile, it assumes that if it can find a header and there are arguments in the compiler everything is good to go. That may or may not be the case. For GCC and Clang, `-fopenmp` actually instructs them to add linking arguments, such as `-lgomp` or `-lomp`, if those aren't available then the dependency will still return found, but fail to work. Let's just do a simple compile to make sure that the dependency actually works.


jpakkane:
:crossed_swords: conflicts.

PR #8656, opened 2021-04-14
Fix issue where multiple dist files were not being produced for comma separated formats value

### LINKED ISSUE #8655: Multiple dist formats doesn't work as stated in help documentation
Seems that `meson dist -C build --formats=xztar,gztar` no longer works. Help documentation for `--formats` says: `Comma separated list of archive types to create.`

tristan957:
Fixes #8655

### DIFF
--- a/mesonbuild/mdist.py
+++ b/mesonbuild/mdist.py
@@ -29,6 +29,7 @@ from mesonbuild import mlog, build
 from .scripts.meson_exe import run_exe
 
 archive_choices = ['gztar', 'xztar', 'zip']
+
 archive_extension = {'gztar': '.tar.gz',
                      'xztar': '.tar.xz',
                      'zip': '.zip'}
@@ -36,8 +37,8 @@ archive_extension = {'gztar': '.tar.gz',
 def add_arguments(parser):
     parser.add_argument('-C', default='.', dest='wd',
                         help='directory to cd into before running')
-    parser.add_argument('--formats', default='xztar', choices=archive_choices,
-                        help='Comma separated list of archive types to create.')
+    parser.add_argument('--formats', default='xztar',
+                        help='Comma separated list of archive types to create. Supports xztar (default), gztar, and zip.')
     parser.add_argument('--include-subprojects', action='store_true',
                         help='Include source code of subprojects that have been used for the build.')
     parser.add_argument('--no-tests', action='store_true',
--- a/run_unittests.py
+++ b/run_unittests.py
@@ -3249,8 +3249,20 @@ class AllPlatformTests(BasePlatformTests):
                       workdir=self.builddir)
             self.assertPathExists(zip_distfile)
             self.assertPathExists(zip_checksumfile)
+            os.remove(xz_distfile)
+            os.remove(xz_checksumfile)
+            os.remove(gz_distfile)
+            os.remove(gz_checksumfile)
+            os.remove(zip_distfile)
+            os.remove(zip_checksumfile)
             self._run(self.meson_command + ['dist', '--formats', 'xztar,gztar,zip'],
                       workdir=self.builddir)
+            self.assertPathExists(xz_distfile)
+            self.assertPathExists(xz_checksumfile)
+            self.assertPathExists(gz_distfile)
+            self.assertPathExists(gz_checksumfile)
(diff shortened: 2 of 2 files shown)


tristan957:
Took away the choices because the help text looks nasty listing all the permutations. This restores the behavior from what looks like a year ago if Git blame is right.

mesonbuild/mdist.py
```diff
@@ -29,14 +31,25 @@
...
+def __archive_permutations() -> T.Tuple[str]:
+    permutations: T.List[str] = []
+    for i in range(len(archive_choices) + 1):
+        for subset in itertools.permutations(archive_choices, i):
+            if len(subset) > 0:
+                permutations.append(','.join(subset))
+
+    return tuple(permutations)
+
+archive_permutations = __archive_permutations()
```
xclaesse, line 44: `archive_permutations` does not seems to be used anywhere.

mesonbuild/mdist.py
```diff
@@ -29,14 +31,25 @@
...
 archive_extension = {'gztar': '.tar.gz',
                      'xztar': '.tar.xz',
                      'zip': '.zip'}
 
 def add_arguments(parser):
     parser.add_argument('-C', default='.', dest='wd',
                         help='directory to cd into before running')
-    parser.add_argument('--formats', default='xztar', choices=archive_choices,
+    parser.add_argument('--formats', default='xztar',
                         help='Comma separated list of archive types to create.')
```
xclaesse, line 53: You can add list of supported archives in the help message.

run_unittests.py
```diff
@@ -3232,15 +3235,52 @@ def dist_impl(self, vcs_init, vcs_add_all=None, include_subprojects=True):
...
-            self.assertPathDoesNotExist(zip_checksumfile)
-            self._run(self.meson_command + ['dist', '--formats', 'zip'],
-                      workdir=self.builddir)
-            self.assertPathExists(zip_distfile)
-            self.assertPathExists(zip_checksumfile)
+            self.build()
+
+            for i in range(len(archive_choices) + 1):
+                permutations = tuple(itertools.permutations(archive_choices, i))
+                for i, subset in enumerate(permutations):
```
xclaesse, line 3242: Is it really worth testing all possible permutations? I would just test the case where you have all archives with `','.join(archive_choices)`.

mesonbuild/mdist.py
tristan957: Meant to do that.

run_unittests.py
tristan957: Good point. I think will do as you have said.

run_unittests.py
```diff
@@ -3232,15 +3235,52 @@ def dist_impl(self, vcs_init, vcs_add_all=None, include_subprojects=True):
...
+                    # skip empty subset
+                    if len(subset) == 0:
```
dcbaker, line 3244: using `len()` to test for container emptiness is a python anti-pattern, just:
```python
if not subset:
```

run_unittests.py
dcbaker: At least one test that isn't all of the options is probably useful. I've been bitten enough times by "but I tested x + y + z and it worked, why doesn't just x wor‽"

run_unittests.py
tristan957: Good to know.

run_unittests.py
```diff
@@ -3235,12 +3237,20 @@ def dist_impl(self, vcs_init, vcs_add_all=None, include_subprojects=True):
...
+            self._run(self.meson_command + ['dist', '--formats', 'gztar'],
+                      workdir=self.builddir)
+            self.assertPathExists(gz_distfile)
+            self.assertPathExists(gz_checksumfile)
             self._run(self.meson_command + ['dist', '--formats', 'zip'],
                       workdir=self.builddir)
             self.assertPathExists(zip_distfile)
             self.assertPathExists(zip_checksumfile)
+            self._run(self.meson_command + ['dist', '--formats', 'xztar,gztar,zip'],
+                      workdir=self.builddir)
```
jpakkane, line 3253: This should check that the output files did get generated.

jpakkane:
Test failure is not related.

PR #8658, opened 2021-04-14
mconf: add --error-on-unkown-options/-e

Flowdalic:

I am not sure why meason decided that this should only be a
WARNING. But users should at least be able to opt-in for a fail-safe
behavior. This is now possible via the --error-on-unknown-options
switch.

### DIFF
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -709,7 +709,7 @@ class CoreData:
                 except KeyError:
                     continue
 
-    def set_options(self, options: T.Dict[OptionKey, T.Any], subproject: str = '', warn_unknown: bool = True) -> None:
+    def set_options(self, options: T.Dict[OptionKey, T.Any], subproject: str = '', warn_unknown: bool = True, error_unknown: bool = False) -> None:
         if not self.is_cross_build():
             options = {k: v for k, v in options.items() if k.machine is not MachineChoice.BUILD}
         # Set prefix first because it's needed to sanitize other options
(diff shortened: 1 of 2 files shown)


eli-schwartz:
Isn't this a duplicate of --fatal-meson-warnings?

Flowdalic:
> Isn't this a duplicate of --fatal-meson-warnings?

No, ``-fatal-meson-warnings` is currently only used in msetup. But I could imagine that one moves `--fatal-meson-warnings`  up, making it a top level argument, so that it can be used in msetup and mconf.

eli-schwartz:
Oh hmm, right.

I wonder if it makes sense to at least sync the names...

Flowdalic:
> I wonder if it makes sense to at least sync the names...

Probably. I wasn't aware of `--fatal-meson-warnings` when I wrote this. Happy to sync it, but I first want to await further feedback from others. But thanks for yours so far :)

Flowdalic:
And I want to add that, if it was for me, I would directly upgrade every "unknown option" warning to an error. I do not see a point why this should be just a warning, as the command clearly was not able to execute the users intention. We where already hurt by this behavior of meson, treating those just as warnings, and I assume others are too. If there is a valid reason to have those as warnings, then this should be explicitly requested (by a command line switch).

I just went with the approach in this PR, as I assumed it is the path of the least resistance. If, however, there is consensus that "unknown option" should be errors by default, then I am happy to provide an alternative PR with that change.

xclaesse:
IMHO this should be the default, no need for a new command line argument. Or maybe do the opposite `--allow-unknown-options` ?

I would like to also make it fatal by default during configuration, but it's a bit more tricky because you could set option for subprojects that are not going to be used, or compiler specific options...

xclaesse:
I think the reason this is a warning is because *years* ago meson was not even checking for unknown option at all, so making it fatal was breaking too many scripts... but it has been warning for a long time now, if making it fatal breaks you script you had your chance...

pinkflames:
Bump. Having --error-on-unknown-options (or even better turned on by default with an opt-out instead) [would have saved me from public embarrassment](https://github.com/gentoo/gentoo/pull/20388#discussion_r635060218). ;)

jpakkane:
I'm fine with making it the default, however in that case this should be merged at the beginning of a dev cycle rather at the end, because my spider sense tingles that this will cause regressions (not our fault, but still).

In either case this needs a test and a release note snippet.

pinkflames:
@jpakkane thanks for taking a look at it. I'm hopeful that this can be sorted out in the next dev cycle then. ;)

Flowdalic:
> I'm fine with making it the default, however in that case this should be merged at the beginning of a dev cycle rather at the end, because my spider sense tingles that this will cause regressions (not our fault, but still).

Would it be ok to merge the current state, given tests and a release note snipped are added, and switch the default to "error on unknown options" at the beginning of the next dev cycle?

tristan957:
This has been implemented for a while.

PR #8661, opened 2021-04-15
docs: document default version for declare_dependency()

emersion:
By default, the project version is picked.


xclaesse:
Thanks!

PR #8663, opened 2021-04-15
Visual Studio Address Sanitizer updates

peterh:
Two commits in this merge request. I'd like feedback on both:

1. Reduce minimum compiler version to 16.9. The version check for 19.29.29917 added in d76345b4beec48484e31110b225c3256d8ba264b refers to an unreleased version (16.10 beta, perhaps?). Was that intentional (ie. is there some fix in the beta that I'm not aware of)? Seems to work in the current "released" version 16.9.4 / 19.28.29914
2. Covert /fstanitize=address to project setting. When I tested this, /fsanitize=address is still being added to "AdditionalOptions". It seems to work as-is, but it would be cleaner if I could remove it from "AdditionalOptions". I'm not quite sure how or where it is being added; pointers would be appreciated.

### DIFF
--- a/mesonbuild/compilers/mixins/visualstudio.py
+++ b/mesonbuild/compilers/mixins/visualstudio.py
@@ -125,7 +125,7 @@ class VisualStudioLikeCompiler(Compiler, metaclass=abc.ABCMeta):
             self.machine = 'arm'
         else:
             self.machine = target
-        if mesonlib.version_compare(self.version, '>=19.29.29917'):
+        if mesonlib.version_compare(self.version, '>=19.28.29910'): # VS 16.9.0 includes cl 19.28.29910
             self.base_options.add(mesonlib.OptionKey('b_sanitize'))
         assert self.linker is not None
         self.linker.machine = self.machine
--- a/mesonbuild/backend/vs2010backend.py
+++ b/mesonbuild/backend/vs2010backend.py
@@ -181,6 +181,10 @@ class Vs2010Backend(backends.Backend):
         self.buildtype = self.environment.coredata.get_option(OptionKey('buildtype'))
         self.optimization = self.environment.coredata.get_option(OptionKey('optimization'))
         self.debug = self.environment.coredata.get_option(OptionKey('debug'))
+        try:
+            self.sanitize = self.environment.coredata.get_option(OptionKey('b_sanitize'))
+        except MesonException:
+            self.sanitize = 'none'
         sln_filename = os.path.join(self.environment.get_build_dir(), self.build.project_name + '.sln')
         projlist = self.generate_projects()
         self.gen_testproj('RUN_TESTS', os.path.join(self.environment.get_build_dir(), 'RUN_TESTS.vcxproj'))
@@ -613,6 +617,8 @@ class Vs2010Backend(backends.Backend):
         # Remove arguments that have a top level XML entry so
         # they are not used twice.
         # FIXME add args as needed.
+        if entry[1:].startswith('fsanitize'):
+            return True
         return entry[1:].startswith('M')
 
     def add_additional_options(self, lang, parent_node, file_args):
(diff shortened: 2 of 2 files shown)


jpakkane:
Asan only became usable in a recent version (which I think is the release number in the code). Before that you had to add a bunch of compiler flags and needed environment setup so it was not really supportable.

peterh:
Non-Windows CI failures appear to be mostly "503 Service Unavailable" in docker.

Windows CI failures appear to be "ERROR: Tried to get unknown builtin option b_sanitize". Unless you have a better suggestion, the next time I have a time slice to spend on this I'll probably change things so that b_sanitize always exists, and will be ignored older MSVC.

peterh:
> Asan only became usable in a recent version (which I think is the release number in the code). Before that you had to add a bunch of compiler flags and needed environment setup so it was not really supportable.

Yes, ASAN only became usable in the most recent minor¹ version (released ~6 weeks ago). But the release number in the code (before my change) is for a future version.

¹ assuming the version components are labelled major.minor.patch

jpakkane:
That's because I originally developed it with the preview version as it was not available in the release version. If it now does and the version is that then the change is ok.

There seems to be some sort of a sporadic failure in CI. Please rebase so the tests get rerun and hopefully go green.

peterh:
"ERROR: Tried to get unknown builtin option b_sanitize". Interesting. I couldn't reproduce this failure after my most recent changes.

mesonbuild/compilers/mixins/visualstudio.py
```diff
@@ -125,8 +125,7 @@ def __init__(self, target: str):
             self.machine = 'arm'
         else:
             self.machine = target
-        if mesonlib.version_compare(self.version, '>=19.29.29917'):
-            self.base_options.add(mesonlib.OptionKey('b_sanitize'))
+        self.base_options.add(mesonlib.OptionKey('b_sanitize'))
```
jpakkane, line 128: This is a behavioural change. The option should not be added at all if the compiler does not support it. It should not handle it if it exists because it conceivably might come from a different compiler. So the version check and warning below is correct but this one is not.

jpakkane:
I tried to replicate this on Windows. It does not seem to happen if you run `run_unittests.py` but does happen when `run_tests.py` is run. Don't know why.

jpakkane:
It turns out that the reason is that it's using the VS backend as `run_tests.py` sets the `MESON_UNIT_TEST_BACKEND` envvar to `vs`.

jpakkane:
This fixes it for me:

```diff
diff --git a/mesonbuild/backend/vs2010backend.py b/mesonbuild/backend/vs2010backend.py
index 5baeff6d2..7ee9fb194 100644
--- a/mesonbuild/backend/vs2010backend.py
+++ b/mesonbuild/backend/vs2010backend.py
@@ -181,7 +181,11 @@ class Vs2010Backend(backends.Backend):
         self.buildtype = self.environment.coredata.get_option(OptionKey('buildtype'))
         self.optimization = self.environment.coredata.get_option(OptionKey('optimization'))
         self.debug = self.environment.coredata.get_option(OptionKey('debug'))
-        self.sanitize = self.environment.coredata.get_option(OptionKey('b_sanitize'))
+        try:
+            self.sanitize = self.environment.coredata.get_option(OptionKey('b_sanitize'))
+        except MesonException:
+            self.sanitize = False
         sln_filename = os.path.join(self.environment.get_build_dir(), self.build.project_name + '.sln')
         projlist = self.generate_projects()
         self.gen_testproj('RUN_TESTS', os.path.join(self.environment.get_build_dir(), 'RUN_TESTS.vcxproj'))
```

jpakkane:
Ping. It would be nice to get this in the next release.

jpakkane:
Have you given up on this pull request?

peterh:
No, I haven't given up. Sorry for the delay. It was due to a combination of gmail not notifying for a week, me not having time, and me forgetting when I did have time (mostly me forgetting, sorry again).

I just updated per your suggestion. I'm going to go test on my other machine (with the older pre-asan vs2019) and then come back and look at the CI results.

jpakkane:
It's all green now. Do you still have things to change or can this be merged?

peterh:
This can be merged, as far as I'm concerned.

Thanks for your patience with me.

PR #8664, opened 2021-04-15
Make `meson.override_find_program` more cross compile friendly

dcbaker:
Currently, Meson doesn't track the difference between a `find_program` override for the build machine vs the host machine. So if you run some code like:
```meson
exe = executable(..., native : true)
meson.overide_find_program('exe', exe)
```
then run
```meson
exe = find_program('exe', native : false)
```
you'll get the exe your overwrote, even though that probably isn't what you meant. With this PR that is actually correctly tracked. The best part is that you don't need to pass a `native` argument to `override_find_program` in most cases, as Meson already knows what the built binaries are for. However, when using a script (like a python script) you do need to set the machine, otherwise it's assumed to be for the HOST machine.

### DIFF
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -268,18 +268,18 @@ class Build:
         self.test_setup_default_name = None
 
         # All overwritten programs, per-machine
-        find_overrides: PerMachineDefaultable[T.Dict[str, T.Union[ExternalProgram, Executable, programs.OverrideProgram, File]]] = PerMachineDefaultable()
-        find_overrides.build = {}
+        program_overrides: PerMachineDefaultable[T.Dict[str, T.Union[ExternalProgram, Executable, programs.OverrideProgram, File]]] = PerMachineDefaultable()
+        program_overrides.build = {}
 
         # All programs that have been searched for, per machine
         searched_programs: PerMachineDefaultable[T.Set[str]] = PerMachineDefaultable()
         searched_programs.build = set()
 
         if environment.is_cross_build():
-            find_overrides.host = {}
+            program_overrides.host = {}
             searched_programs.host = set()
 
-        self.find_overrides = find_overrides.default_missing()
+        self.program_overrides = program_overrides.default_missing()
         self.searched_programs = searched_programs.default_missing()
 
         # If we are doing a cross build we need two caches, if we're doing a
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -266,15 +266,20 @@ class Build:
         self.stdlibs = PerMachine({}, {})
         self.test_setups: T.Dict[str, TestSetup] = {}
         self.test_setup_default_name = None
-        self.find_overrides: T.Dict[str, T.Union['Executable', programs.ExternalProgram, programs.OverrideProgram]] = {}
+
+        # All overwritten programs, per-machine
+        find_overrides: PerMachineDefaultable[T.Dict[str, T.Union[ExternalProgram, Executable, programs.OverrideProgram, File]]] = PerMachineDefaultable()
+        find_overrides.build = {}
 
         # All programs that have been searched for, per machine
         searched_programs: PerMachineDefaultable[T.Set[str]] = PerMachineDefaultable()
         searched_programs.build = set()
 
         if environment.is_cross_build():
+            find_overrides.host = {}
             searched_programs.host = set()
 
+        self.find_overrides = find_overrides.default_missing()
         self.searched_programs = searched_programs.default_missing()
 
         # If we are doing a cross build we need two caches, if we're doing a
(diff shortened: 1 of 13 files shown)


mesonbuild/build.py
```diff
@@ -227,8 +228,22 @@ def __init__(self, environment: environment.Environment):
         self.stdlibs = PerMachine({}, {})
         self.test_setups: T.Dict[str, TestSetup] = {}
         self.test_setup_default_name = None
-        self.find_overrides = {}
-        self.searched_programs = set() # The list of all programs that have been searched for.
+
+        # All overwritten programs, per-machine
+        find_overrides: PerMachineDefaultable[T.Dict[str, T.Union['ExternalProgram', 'Executable', 'File']]] = PerMachineDefaultable()
```
xclaesse, line 233: Could we take this opportunity to rename it to `program_overrides` because `find_overrides` means nothing, especially since we also have `dependency_overrides`.

mesonbuild/interpreter/mesonmain.py
```diff
@@ -277,23 +280,30 @@ def install_dependency_manifest_method(self, args, kwargs):
...
+
         if isinstance(exe, mesonlib.File):
             abspath = exe.absolute_path(self.interpreter.environment.source_dir,
                                         self.interpreter.environment.build_dir)
             if not os.path.exists(abspath):
                 raise InterpreterException('Tried to override %s with a file that does not exist.' % name)
-            exe = OverrideProgram(name, abspath)
+            exe = OverrideProgram(name, abspath, for_machine=for_machine)
         if not isinstance(exe, (ExternalProgram, build.Executable)):
             raise InterpreterException('Second argument must be an external program or executable.')
```
xclaesse, line 305: nitpicking: this check is now useless, it's guaranteed to be one those 2 types. Removing those is the point of typed_pos_args.

mesonbuild/interpreter/mesonmain.py
```diff
@@ -277,23 +280,30 @@ def install_dependency_manifest_method(self, args, kwargs):
...
+        name = args[0]
+        exe: T.Union[build.Executable, mesonlib.File, ExternalProgram] = unholder(args[1])
+
+        # We need native when using a file, and someone might want to override
+        # something for a specific machine, otherwise use the for_machien from
+        # the program itself.
+        if isinstance(exe, mesonlib.File) or 'native' in kwargs:
+            for_machine = self.interpreter.machine_from_native_kwarg(kwargs)
+        else:
+            for_machine = exe.for_machine
```
xclaesse, line 296: I think we should error out if `native` kwarg does not match `exe.for_machine`.

dcbaker: I thought about that too, but sometimes things for the host machine are "runs on the host machine" and sometimes they mean "runs for the host machine", compilers, cmake, pkg-config, and some other tools are all set as "host machine binaries" but they actually run on the build machine and produce output for the host machine. Not sure if that's a worthwile thing to handle or not?

dcbaker: Hmmm, the more I think about it the more I think starting with the stricter behavior and relaxing it after we find a usecase might be better.

xclaesse: Exactly what I was going to say.

xclaesse:
As far as I remember, it was intentional to not have the `native` kwarg on `meson.override_find_program()`. The reason is a program is always used on the BUILD machine, with an exe wrapper if needed.

In your example:
```meson
exe_for_host = executable(..., native : false)
exe_for_build = executable(..., native : true)
meson.override_find_program('exe', exe_for_host)
meson.override_find_program('exe', exe_for_build)
```

I think erroring is actually the right thing, but the message could be better. There is no point in doing `meson.override_find_program('exe', exe_for_host)` because overridden programs are only useful to be run on the build machine.

On a similar topic, I had that PR: https://github.com/mesonbuild/meson/pull/10899. The idea is it only makes sense to override a program with something we can run. Could be a python script, an exe for build machine, or an exe for host machine and an exe wrapper is available.

dcbaker:
Here's the problem, it's not really clear if being for BUILD means "runs natively" or "outputs for build". Certainly some HOST machine binaries are built for the HOST machine and need an exe wrapper, but others are very much not. Think of things like your compilers, linkers, strip, etc. There are also cases where programs generate different code depending on the platform/architecture they're generating for. flex and bison come to mind, where the windows versions do some trickery to make code that MSVC is happy with, but the *nix versions don't.

xclaesse:
> Think of things like your compilers, linkers, strip, etc

I see, that makes more sense indeed. It can get useful together with https://github.com/mesonbuild/meson/pull/10938 too.

dcbaker:
I'm closing this in favor of the version in my native subproject series

PR #8665, opened 2021-04-16
Improve error messages for wrap-redirect file

benjamin-otte:
While discussing error messages on IRC with xclaesse, we found 448b11cb7fb41199880cc60d9fce040ea0cf5a3b which improves one of the error messages, but leaves out the others. This commit updates them, too.

### DIFF
--- a/mesonbuild/wrap/wrap.py
+++ b/mesonbuild/wrap/wrap.py
@@ -253,15 +253,15 @@ class PackageDefinition:
             for i, p in enumerate(fname.parts):
                 if i % 2 == 0:
                     if p == '..':
-                        raise WrapException('wrap-redirect filename cannot contain ".."')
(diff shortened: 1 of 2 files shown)


xclaesse:
You have to update the expected error message in unit tests. See `test_wrap_redirect()` in run_unittests.py.

eli-schwartz (approved):
lgtm once CI passes

benjamin-otte:
Looks like this fails on Windows because the separator is `\` instead of `/`. Is there a better comparison for filenames/paths or do I just use a wildcard to get around the separator mismatch?

xclaesse:
fname is a Path() object already, so you can use fname.as_posix() in the error message to always have /.

jpakkane:
Ping. Please fix the last issue so this can be merged.

eli-schwartz:
⚔️ conflicts

And another ping...

Issue #8144, opened 2020-12-31
dist failure with relative submodule paths


scop:
**Describe the bug**

dist fails with git submodules using relative paths.

**To Reproduce**

The entire tree is at https://gitlab.freedesktop.org/xdg/shared-mime-info/-/tree/40db4c7adba8a5cc586e24f5b4b5aff0b2a81bcc, see .gitlab-ci.yml there for how it's built in CI. The xdgmime submodule is defined as `../xdgmime.git`

CI failure log is at https://gitlab.freedesktop.org/xdg/shared-mime-info/-/jobs/6403481 -- essentially it seems meson is trying to checkout the submodule from a directory relative to the main project working dir, whereas it should treat relative submodule path as relative to the superproject's default remote repo; https://git-scm.com/docs/git-submodule#Documentation/git-submodule.txt-add-bltbranchgt-f--force--nameltnamegt--referenceltrepositorygt--depthltdepthgt--ltrepositorygtltpathgt

**Expected behavior**

Working submodule checkout during dist.

**system parameters**
* Is this a [cross build](https://mesonbuild.com/Cross-compilation.html) or just a plain native build (for the same computer)?
* what operating system (e.g. MacOS Catalina, Windows 10, CentOS 8.0, Ubuntu 18.04, etc.)
* what Python version are you using e.g. 3.8.0
* what `meson --version`
* what `ninja --version` if it's a Ninja build

eli-schwartz:
I have linked a PR which should fix this.

PR #8667, opened 2021-04-16
Fix meson dist with relative submodule URLs

drobilla:
See the last commit message for details.

### DIFF
--- a/mesonbuild/mdist.py
+++ b/mesonbuild/mdist.py
@@ -77,13 +77,40 @@ def parse_gitmodules(path):
 
 
 def prepare_submodules(src_root, module_file):
-    """Read .gitmodules and return a list of all submodule paths."""
+    """Read .gitmodules and rewrite it if necessary to support relative URLs.
+
+    Returns a list of all submodule paths.
+    """
 
     submodules = parse_gitmodules(module_file)
 
+    # Gather paths while checking if there are any relative submodules
     paths = []
+    has_relative_submodules = False
     for path, submodule in submodules.items():
         paths += [path]
+        url = submodule['url']
+        if url.startswith('./') or url.startswith('../'):
+            has_relative_submodules = True
+
+    # If all submodules have absolute URLs, just return the paths
+    if not has_relative_submodules:
+        return paths
+
+    # Otherwise, write a new .gitmodules file with URLs set to the original
+    # paths.  This way, the submodules will be cloned from the submodules in
+    # the source root, just like the top-level project is.  It would also be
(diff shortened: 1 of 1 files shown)


drobilla:
I'm not sure what makes sense with submodules in the grand scheme of things, this just seemed like an outright bug that could be fixed, so I tried to make it work.

From a user perspective one thing that still seems odd about this is that, if you have subprojects as submodules, they will be included in the archive unconditionally.  One might want to exclude them and rely on wrap files to fetch things for users if they need them, but `meson dist` has no option for this.  Though `--include-subprojects` is somehow in a similar vein... submodules aren't necessarily subprojects.  🤷

eli-schwartz:
Great work! The overall approach seems sane here, and the first two commits especially seem good to have anyway so thank you for factoring out each discrete change into its own commit.

> From a user perspective one thing that still seems odd about this is that, if you have subprojects as submodules, they will be included in the archive unconditionally. One might want to exclude them and rely on wrap files to fetch things for users if they need them, but `meson dist` has no option for this. Though `--include-subprojects` is somehow in a similar vein... submodules aren't necessarily subprojects. shrug

Shower thoughts here (literally, actually) as I don't have much time today:

Maybe we could compare each submodule and, if it is in the subprojects directory and has an associated wrap file, ignore it by default, unless `--include-subprojects` is used?

This seems like an additional change, above and beyond "fix relative urls", so it's okay to not fix it right now. :D

drobilla:
> Maybe we could compare each submodule and, if it is in the subprojects directory and has an associated wrap file, ignore it by default, unless --include-subprojects is used?

Hmm, good idea.  Seems like that would cover most of what people might want to do, without really breaking anything or adding new options.  I'll have a go.

eli-schwartz:
I completely forgot all about this PR. :D Meanwhile, I ended up taking a completely different approach to completely rewriting `meson dist` in #9210 

My rewrite uses git archive and assumes you have inited the submodule yourself in the actual project tree, so it doesn't need to rewrite .gitmodules -- it is also marked as fixing #8144 which this PR would have fixed too, I guess.

I'm going to close this PR since the thing it is solving is solved a different way, but 

> Maybe we could compare each submodule and, if it is in the subprojects directory and has an associated wrap file, ignore it by default, unless `--include-subprojects` is used?

This is welcome to receive a new PR.

drobilla:
Heh, no worries, me too (or at least it was on a back burner).  Not needing to rewrite .gitmodules is nice.  I'll revisit this on the node with my project structure soonish.

PR #8679, opened 2021-04-20
mconf: print core options for subprojects

dcbaker:
Looking in the right dictionary is pretty useful, actually.

@jenatali


xclaesse:
Lgtm

PR #8680, opened 2021-04-20
Wrap mconf output

### LINKED ISSUE #6965: `meson configure` output formatting unreadable
In meson version 0.54.0 the output of `meson configure` (in build dir) to the terminal became unreadable due to seemingly broken spacing between columns and description text does not staying in it's column if more than one line is required to display it.

Here are a screenshots of how it looks for me currently:
image
image

I believe that in earlier versions there was no such problem, and the formatting started to change since 0.53, but was still readable until now. So if nothing else, it would be nice if it could be rolled back to pre-0.53 state.

dcbaker:
This clamps the maximum line length of `meson configure` output to 160 characters, with each column taking multiple rows if necessary, but in a nice order. I picked 160, because shorter than that started to make the output hard to read, descriptions either got wrapped basically all the time (because they'd only be able to be between 20 and 40 characters wide), or names and values got clamped, even at 120 it was pretty tight.

Fixes #6965


eli-schwartz:
I don't see how this helps anything. We should try to use the entire interactive terminal width, not clamp to some hardcoded-in-source maximum (maybe if the output is not a tty, hardcoded widths could be used).

If someone is using an 80-char terminal for some reason, then the output should not be wrapped to 160, since I've just tried it and it became even harder to read than before this change. If we detect the current terminal width then there's no downside IMO to making it "pretty tight", because for them, it is still better than the alternative...

eli-schwartz:
Also is there some way to avoid printing whitespace at the end of a rendered line but before the newline char? If you really exercise that 80-char display wrapping, you sometimes end up with random empty lines interspersed through the output, because the terminal is wrapping lines on whitespace only.

eli-schwartz:
The project I tested this on is wrapping a project option at $COLUMNS == 168, such that 3 characters flow onto a new line :D so respecting those extra 8 chars I have would be quite handy...

dcbaker:
I've updated it to check your terminal width and honor that (and $COLUMNS), but frankly there's just too much data for 80 characters, it's going to look like crap no matter what we do here.

dcbaker:
Yes, actually we can avoid printing whitespace at the end. We just need to wrap to that whitespace limit. That was just me blindly copying what the old code did :/

eli-schwartz:
Right, practically speaking if your terminal is 80 chars you might want to pipe it to cat, in order to suppress wrapping, and use methods like highlighting a line by triple-clicking in order to follow each option.

Output: [mconf-out.txt](https://github.com/mesonbuild/meson/files/6347180/mconf-out.txt)

The makepkg-template-dir option is still being split across lines, even though the first column could really be widened by 3 chars I'd say.

dcbaker:
Here's another go with a slightly tweaked huerisitc for splitting up the screen.

eli-schwartz:
[mconf-out.txt](https://github.com/mesonbuild/meson/files/6347847/mconf-out.txt)

Looks much better this time!

xclaesse:
While looking at output, I noticed that `<inherited from main project>` is not printed any more for yielding options value.

xclaesse:
Nitpicking: I also notice that the `subproject:` prefix is not repeated any more for every option in that subproject. I guess that's intentional change, but I'm not sure that's a good idea. Having the prefix has the advantage of making that syntax more discoverable (just like for `build.` prefix) and I can just copy/paste the option from the list when I do `-D` instead of having to compose the name manually. OTOH, I agree it makes things a bit more readable... I personally still would prefer keeping the prefix.

dcbaker:
I guess my concern here is that it's not bad when you have something like: `sub1:option`, but something like `company-special-format-reader:option` or `company-special-format-reader:really-long-and-complicated-option` make the wrapping problem worse (yes, I personally know companies that really wants projects named things that are purely descriptive and not abbreviated like that.)

xclaesse:
I have to admit `gst-plugins-good:` prefix is a bit border line already for readability. But since I have an alias that pipe the output into a pager that don't wrap lines, it's not that bad.

jpakkane:
Have the nits been sufficiently picked or does someone still want changes to the current layout?

eli-schwartz (approved):
As far as I'm concerned this is ready to go.

ppaalanen:
Thanks for this, I just got notified via closing an issue!

I wonder, I looked at the example output in the PR description and it seems to me that if you have to choose between wrapping an option value and description, then wrap the description. Descriptions are human-readable text, so wrapping them does not hurt, while option values can be white-space sensitive so wrapping them may make them ambiguous to the reader.

Anyway, much better than before.

PR #8687, opened 2021-04-22
store the list of initializes subprojects in the coredata structure

dcbaker:
We need to konw on rconfigure which options have already bee set not
just for the super project, but also for the subproject. However, using
first_invocation is not sufficient, as a reconfigure could add a new
subpproject that wasn't present before, and we need to initialize that
project's builtins.


jenatali:
Just chiming in, this fixes the issue for me.

PR #8688, opened 2021-04-23
Allow C# projects to contain multiple file types

tschoonj:
This allows for linking a C# executable against a C++ .NET assembly that was generated in the same project.

### DIFF
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -818,8 +818,9 @@ class BuildTarget(Target):
                     m = '{0} targets can only contain {0} files:\n'.format(lang.capitalize())
                     m += '\n'.join([repr(c) for c in check_sources])
                     raise InvalidArguments(m)
-                # CSharp and Java targets can't contain any other file types
-                assert(len(self.compilers) == 1)
+                # Java targets can't contain any other file types
+                if lang == 'java':
+                    assert(len(self.compilers) == 1)
                 return
 
     def process_link_depends(self, sources, environment):


tschoonj:
This fix is probably too simple, even though it actually fixes my problem...

Any thoughts how this could be improved upon? Basically, linking a C# executable against a DLL should not add a C++ compiler to `self.compilers`.

dcbaker:
At the very least we need a test for this, but I'm not even sure how you output .net assembly from c++ and whether what we really need is a `donet()` target like the `jar()` target, or what.

tschoonj:
With the patch in this PR I managed to compile the C++ .NET assembly and use it successfully from C#:

https://github.com/tschoonj/xraylib/blob/8c082b576da211ff915222c64e3de2b47c7872d2/csharp/meson.build

tschoonj:
I should be able to come up with a test for this, I will do that when I have some time later.

tschoonj:
It looks to me as if the C++ compiler generates a 32-bit dll, but `csc` produces a 64-bit executable.

Is there anything I can do to ensure that both produce output for the same architecture?

Thanks!

tschoonj:
@dcbaker any thoughts on how to fix the build?

Thanks!

dcbaker:
I'll have to think about it some more, i've been putting off getting out my windows machine for a while now, I may have to go ahead and do that.

tschoonj:
Something that surprises me is that vc2017x86ninja appears to be the only build where the C# tests are run, as they are skipped on the other vc build.

PR #8692, opened 2021-04-25
Generic Posix complier linker

TAAPArthur:
There exists more compilers than gcc/clang, like tcc for example, and
the interface is kinda the same. If a gcc/clang compiler can't be found
then default to this generic one instead of just giving up.

## Background:
I'm trying to use tcc as my main compiler because it is significantly faster than gcc. So projects I use using meson and meson not supporting tcc is non ideal. This patch here works and I've been using it for a few months now. It can probably be cleaned up; alot of it was pattern matching how the gcc compiler worked.

## Things that may be controversial: 
I also imagine that someone else would come along complaining that their alternative POSIX compiler doesn't work. So I made this patch more generic. It matches any compiler ending in cc (tcc, scc, lacc, cc etc) that isn't already matched by the specific compilers. If, for some reason, there was a compiler than wasn't aptly name as such, they could also link/rename it to /bin/cc as a quick workaround and alternative to adding support for their compiler.


jpakkane:
The relevant question for this is "how do we know that the generic linker is working a) now and b) in the future". The only correct answer is that we must have CI in place. If you do that then this is no longer a "general compiler" but specific to the compiler that is tested (and any code not tested is broken). At that point it might well be called by its real name rather than a generic name. 

Furthermore, just making one compiler backend work reliably among its different release versions so supporting many different ones with a single backend seems fraught with peril.

All of that is a very complicated way of saying that a "general compiler" like this is unlikely to be accepted, but if you change it so that it is specific to tcc (including fixing the detection based on output) then that could be merged.

TAAPArthur:
> The relevant question for this is "how do we know that the generic linker is working a) now and b) in the future". The only correct answer is that we must have CI in place. If you do that then this is no longer a "general compiler" but specific to the compiler that is tested (and any code not tested is broken). At that point it might well be called by its real name rather than a generic name.

I don't following this reasoning. My ideal is that it is tested according to the some known [spec](https://pubs.opengroup.org/onlinepubs/7908799/xcu/cc.html). And that any such compiler could be used to test it to that end.

> Furthermore, just making one compiler backend work reliably among its different release versions so supporting many different ones with a single backend seems fraught with peril.

Isn't this problem specific to the bloated compilers that are already supported. I believe there won't be as much issue since these compilers aren't as fragile as gcc. Also the actually compiler commit is like 70 lines with few options. Would making the patch smaller resolve this concern.

> All of that is a very complicated way of saying that a "general compiler" like this is unlikely to be accepted, but if you change it so that it is specific to tcc (including fixing the detection based on output) then that could be merged.

So the part that I don't get is why its better to just fail without trying rather than attempting the compilation? Especially given that this enables a fair amount of niche compilers. Even if that was changed to be specific to tcc, there is some chance that tcc doesn't support some feature and the compilation fails midway. Personally, when trying to build projects on a system without gcc, I'd prefer for the code to "just work" rather than being forced to install some heavy compiler when it can be avoided. 

Also the code to support tcc would be the exact same as say scc. Think this would generally be true for many compilers. 

Is it viable to have a list of "supported" that are explicitly tested but still keep the code and detection generic?

dcbaker:
The problem is with specs (as someone who is paid to implement one for a day job) is that no one really implements the spec, they implement what they *think* the spec is, plus other stuff that seems useful to them. I doubt that many of these compilers actually implement exactly the posix spec, but either something very much like the posix spec, or a superset of the posix spec (either on accident, or because some part of the posix spec is actually quite annoying). I agree with @jpakkane here, having a "generic" posix compiler and linker sounds great on paper, but is much harder in practice. Especially if it just tires to blindly match anything that ends in "cc".

That said, as an abstraction I think having a generic Posix compiler and linker interface within Meson makes a ton of sense (I think I added one for the linker already?), even if the end result is that the implementation of several compilers becomes:
```python
class TCCompiler(PosixCompiler, CCompiler):
    id: str = 'tcc'

class OtherCCompiler(PosixCompiler, CCompiler):
    id: str = 'occ'
```

Another complication here is that we like to be able to present information about the name and version of the compiler you're using. There isn't a spec on what should be printed by the `-v` or `--version` flag, so you could get a GNU style printout (way too much information), or you could get `1.2.3`, with no other information at all.

And TCC is even more complicated because there are effectively two forks of it. The stable ancient release, and the "mob" branch that has a lot of fixes, but is also scary because *anyone* can commit *anything* to that branch without any kind of review. I think it's worth differentiating those from each other.

My opinion (and just mine) is that failing gracefully with "Sorry, I don't support Foo C Compiler, patches welcome" is much nicer than partially configuring and then failing to work in strange and mysterious ways, including hitting tracebacks somewhere down in the depths of Meson because some assumption somewhere was broken, or fully configuring and failing to compile, or worse "succeeding" to compile, but with a pile of warnings about "I ignored option --important because I don't know what it is" (I'm looking at you icc)

dcbaker:
Among the challenges off the top of my head:
1) Meson will automatically configure the compiler to emit a depfile if possible, for GCC and Clang we inject `-MD -MQ $target -MF $file`, for this is not part of the posix spec, but users would absolutely want this to make rebuilding reliable. To do this wtih tcc you'd add `-MD -MF $file `(note no -MQ option)
2) rsp files are not part of the posix spec, but tcc supports them, and Meson will automatically use them to escape command line limts if it knows the compiler can use them.
3) tcc needs the -enable-new-dtags to use DT_RUNPATH instead of DT_RPATH, you really want the former and not the later (DT_RUNPATH isn't "new" and hasn't been for a long time.)
4) tcc doesn't have -pedantic or -Wextra, just -Wall and -Werror, Meson has builtins for setting reasonble argument configurations based on some predefined warning levels (they're not perfect, but for many people they're "good enough")

I'm sure there's more, but I think that gives you an idea. Meson is "smart" with compilers, it knows something about supported compilers and allows users to abstract away implementation details for common operations, or simply automate things you want but may not know how to do like using depfiles so you don't have to list headers in your dependencies, which you will inevitably 
get wrong.

> This can't really be helped?. tcc doesn't support everything gcc does like 128bit numbers or some assembly instructions. I can't imagine all these cases can be detected during configuring.

Sure it can:
```meson
if not cc.compiles('#include <stdint.h>\n int128_t foo = 0; \n int main(void) { return (int)foo; }')
  error('128 bit integer support required, but not detected!')
endif
```
and if you're using something exotic like 128 bit ints I sure hope you're writing some kind of check for them.

But meson also has things like
```meson
project(
  'foo',
  'c',
  default_options : ['c_std=c11'],
)
```
If I have a version of GCC that doesn't support C11, Meson will figure that out as soon as it tries to detect gcc and error out. But with this generic compiler you can't do that.

eli-schwartz:
- I don't see the problem in having a compiler type that doesn't support rsp files because it's not in the POSIX spec and needs to be implemented in a KnownCompilerFoo class.
- I don't see what dtags has to do with it, meson certainly doesn't specify or try to handle them at any point in time... is this a red herring? Quote from Allan McRae (Arch toolchain guy, former release manager for one version of glibc, blah blah blah) who commented in March of this year on the topic of applications having *inconsistent behavior* on Ubuntu vs. Arch:
  > I know Fedora does not enable new dtags in their binutils either.  Not sure why Ubuntu has this early.

  "Early." In 2021. new-dtags isn't some godlike panacea. There's a reason Debian/Ubuntu needs to explicitly enable it because it doesn't default to on, and I surely don't see how it is meson's place to override toolchain policy, which, then again, meson doesn't anyway.
- A PosixCCompiler doesn't need to set non-POSIX warnings, and if users intentionally use some niche "simple POSIX" $CC then I humbly suggest they know best about their non-fatal warnings which aren't even perfect anyway.

Depfiles/std are the problematic things I see here, but maybe we can simply print a warning for depfiles, at least. If given explicit std=, we could either error out "unsupported, your PosixCCompiler sucks" or, maybe test out -std=... since we know the argument calling convention uses dashes, not slashes.

TAAPArthur:
Giving up on this or anything similar being merged in favor of [muon](https://git.sr.ht/~lattis/muon)

PR #8697, opened 2021-04-26
Update Syntax.md

mooskagh:
"Stored by value" is more correct way to explain that example.

Mutable vs immutable means that you cannot mutate the value (e.g. list vs tuple in Python), and the example shows that `var2` is actually mutable.

Copying/storing a reference vs value is what what matters in the assignment, in Python `a=b` means `a` and `b` are references to the same list, while in meson `a=b` copies the value of `b` into `a`.

### DIFF
--- a/docs/markdown/Syntax.md
+++ b/docs/markdown/Syntax.md
@@ -36,8 +36,10 @@ var2 = 102
 ```
 
 One important difference in how variables work in Meson is that all
-objects are stored by value. This is different from, for example, how
-Python works.
+objects are immutable. When you see an operation which appears like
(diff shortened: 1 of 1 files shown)


mooskagh:
I've read the document further and realized what exactly was meant by immutability.

I've updated it to be less confusing, but now it's a bit too wordy.

dcbaker (approved):

eli-schwartz:
Should be amended and force pushed, not made into 2 commits. :/

eli-schwartz:
Of course github is allergic to good formatting, lol, so the resulting squashed commit message is rather run-on...

PR #8699, opened 2021-04-26
interpreter: Add dependency_fallbacks()

xclaesse:
The main goal of the new dependency_fallbacks() function call is to be
able to express the common pattern: Check pkg-config first with
multiple names, then try with find_library with multiple names, and
finally fallback to a subproject. An extra niche use-case is to try with
has_function() too in the case the needed API is part of libc builtin in
the compiler (e.g. iconv in glib).

### DIFF
--- a/mesonbuild/interpreter/dependencyfallbacks.py
+++ b/mesonbuild/interpreter/dependencyfallbacks.py
@@ -1,13 +1,15 @@
 from .interpreterobjects import DependencyHolder, SubprojectHolder, extract_required_kwarg
+from .compiler import CompilerHolder, find_library_permitted_kwargs, has_function_permitted_kwargs
 
 from .. import mlog
 from .. import dependencies
 from .. import build
 from ..wrap import WrapMode
-from ..mesonlib import OptionKey, extract_as_list, stringlistify, version_compare_many
-from ..dependencies import DependencyException, NotFoundDependency
-from ..interpreterbase import (InterpreterObject, FeatureNew,
+from ..mesonlib import OptionKey, extract_as_list, stringlistify, version_compare_many, unholder
+from ..dependencies import DependencyException, NotFoundDependency, InternalDependency
+from ..interpreterbase import (InterpreterObject, ObjectHolder, FeatureNew,
                                InterpreterException, InvalidArguments,
+                               permittedKwargs, typed_pos_args,
                                TYPE_nkwargs, TYPE_nvar)
 
 import typing as T
@@ -15,9 +17,8 @@ if T.TYPE_CHECKING:
     from .interpreter import Interpreter
 
 
-class DependencyFallbacksHolder(InterpreterObject):
+class DependencyFallbacks:
     def __init__(self, interpreter: 'Interpreter', names: T.List[str], allow_fallback: T.Optional[bool] = None) -> None:
-        super().__init__()
         self.interpreter = interpreter
         self.subproject = interpreter.subproject
         self.coredata = interpreter.coredata
@@ -28,6 +29,7 @@ class DependencyFallbacksHolder(InterpreterObject):
         self.subproject_name = None
         self.subproject_varname = None
         self.subproject_kwargs = None
+        self.candidates = []
         self.names: T.List[str] = []
         for name in names:
             if not name:
@@ -38,6 +40,17 @@ class DependencyFallbacksHolder(InterpreterObject):
             if name in self.names:
                 raise InterpreterException('dependency_fallbacks name {name!r} is duplicated')
             self.names.append(name)
+
+    def add_find_library(self, args, kwargs):
+        self.candidates.append((self._do_find_library, args, kwargs))
+
+    def add_has_function(self, args, kwargs):
+        self.candidates.append((self._do_has_function, args, kwargs))
+
+    def set_subproject(self, args, kwargs):
+        if self.subproject_name:
+            raise InterpreterException('dependency_fallbacks.subproject is only allowed once')
+        self._subproject_impl(args[0], None, kwargs)
 
     def set_fallback(self, fbinfo: T.Optional[T.Union[T.List[str], str]], default_options: T.Optional[T.List[str]] = None) -> None:
         # Legacy: This converts dependency()'s fallback and default_options kwargs.
@@ -71,6 +84,12 @@ class DependencyFallbacksHolder(InterpreterObject):
         self.subproject_varname = varname
         self.subproject_kwargs = kwargs
 
+    def _make_internal_dep(self, ext_deps=None, incdirs=None):
+        # Wrap ext_deps (e.g. ExternalLibrary) into an InternalDependency
+        # because dependency() function garantees the return type is DependencyHolder.
+        dep = InternalDependency('undefined', incdirs or [], [], [], [], [], [], ext_deps or [], {})
+        return DependencyHolder(dep, self.subproject)
+
     def _do_dependency_cache(self, kwargs: TYPE_nkwargs, func_args: TYPE_nvar, func_kwargs: TYPE_nkwargs) -> T.Optional[DependencyHolder]:
         name = func_args[0]
         cached_dep = self._get_cached_dep(name, kwargs)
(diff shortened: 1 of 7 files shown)


xclaesse:
Since this is a huge refactoring that moves lots of code into a new file, I'm wondering if we should merge that first (excluding the new API) and then discuss on the actual API? IMHO the refactoring is valuable on its own.

mensinda:
I would also agree that we should do the code refactoring separately to keep the two topics as much separated as possible. All changes that reduce the size of `interpreter.py` are good changes.

xclaesse:
Made the needed refactoring, without adding new API, in https://github.com/mesonbuild/meson/pull/8861.

xclaesse:
Ok, this is now ready for review, at least getting opinions on the proposed API.

mesonbuild/interpreter/dependencyfallbacks.py
```diff
@@ -18,6 +20,7 @@
 class DependencyFallbacksHolder(InterpreterObject):
```
mensinda, line 20: This is not an `ObjectHolder` since it doesn't actually hold anything. Especially when looking at #8885 (which I hope will be merged very soon), it would be nice if this was renamed to something like `DependencyFallbacksObject`.

Even better would be if you could refactor this class to into a raw `DependencyFallbacksObject` with just the logic and *all* the data, and a `DependencyFallbacksHolder` with all the user-facing functions and only the `self.held_object`.

mesonbuild/interpreter/dependencyfallbacks.py
```diff
@@ -38,6 +42,30 @@ def __init__(self, interpreter: 'Interpreter', names: T.List[str], allow_fallbac
             if name in self.names:
                 raise InterpreterException('dependency_fallbacks name {name!r} is duplicated')
             self.names.append(name)
+        self.methods.update({'find_library': self.find_library_method,
+                             'has_function': self.has_function_method,
+                             'subproject': self.subproject_method,
+                             })
+
+    @permittedKwargs(find_library_permitted_kwargs - {'required', 'static'})
```
mensinda, line 50: If you are using `typed_pos_args` you might as well also use `typed_kwargs`.

mesonbuild/interpreter/dependencyfallbacks.py
mensinda: Also, `DependencyFallbacksHolder` must derive from `MutableInterpreterObject` since it is mutable

mensinda (changes requested):
I haven't looked at the new syntax yet, but I have some technical suggestions.

Also, I am aware that you don't like type annotations, but please use full type annotations and enable `dependencyfallbacks.py` in `run_mypy.py`, since this is new code and we will never get full type checking coverage otherwise..

mesonbuild/interpreter/dependencyfallbacks.py
xclaesse: > Even better would be if you could refactor this class to into a raw `DependencyFallbacksObject` with just the logic and _all_ the data, and a `DependencyFallbacksHolder` with all the user-facing functions and only the `self.held_object`.

That makes a lot of sense, good idea. Made that change.

> Also, `DependencyFallbacksHolder` must derive from `MutableInterpreterObject` since it is mutable

It's not that easy because interpreterbase does deepcopy MutableInterpreterObject, which causes copy of self.interpreter.

mesonbuild/interpreter/dependencyfallbacks.py
xclaesse: Because find_library() in Compiler object has not been ported yet, that should be done in another PR.

xclaesse:
That would make sense, yes. Easier syntax for the common case.

tristan957:
I think the concept behind the PR makes a lot of sense. Hope you can get back to it eventually.

xclaesse:
@tristan957 since that use-case was already fully implemented internally, I made a PR with just that. https://github.com/mesonbuild/meson/pull/9330

dcbaker:
This needs a rebase something awful :)

I'm not entirely convinced by this approach, but one aboslutely glaring problem is see is that you should be passing the language as a string, not the compiler object, because cross compiling
```meson
cc = meson.get_compiler('c', native : false)
cc_build = meson.get_compiler('c', native : true)

df = dependency_fallback('a', 'b', 'c)
df.find_library(cc, 'foo')
dependency(df, native : false)
```

Sure, you *could* have some logic to check that you pass the right `cc`, but that might not get noticed if it only affects a host != build configuration. passing the language as a string though lets meson just get the right compiler every time.

Here's my main concern with the approach: meson tries *really* hard not to have mutable objects, but this is a mutable object. So we should think really hard about whether this is a good idea or not. Your example doesn't give my any case where something like
```meson
dependency('foo', 'bar', find_library : ['foo'], has_function : 'bar', had_header : 'foo.h', language : 'c')
```
wouldn't already do exactly what we want without introducing more types and more biolerplate, as well as a mutable object. or, alternatively could we construct another immutable object to use instead?
```meson
lf = library_fallback('foo', 'bar', has_function : 'bar', has_header : 'foo.h', prefix : '#define _GNU_SOURCE', langauge : c')
dependency('foo', 'bar', find_library : lf, fallback : 'foo')
```
(or `cc.library_fallback`, though again, you have the target machine mismatch issue). The other thing about this appoarch, is it keeps the existing `fallback` keyword argument, and provides a similar approach with a `find_library` keyword

jpakkane:
> this is a mutable object. So we should think really hard about whether this is a good idea or not.

I don't like those either. At the very least it should be made more explicit, perhaps by calling it a fallback_builder or something and then doing the usual thing where you need to call `fb.get_fallback()` to get the actual object (and make the builder so that calling anything else on it is a hard failure).

Alternatively we could try something a bit more immutable like:

```
arr = [dep_lookup(type:'library', has_headers: 'foo.h'),
   dep_lookup(type: 'normal', name: 'foobar'),
   ...
]

final_dep = multistage_dep_lookup(arr)
```

The syntax is a bit wonky but you get the idea.

tristan957:
> dependency('foo', 'bar', find_library : ['foo'], has_function : 'bar', had_header : 'foo.h', language : 'c')

Seems not bad

xclaesse:
> Sure, you could have some logic to check that you pass the right cc, but that might not get noticed in testing if it only affects a host != build configuration and you don't have one of those in your CI (or the right options set). passing the language as a string though lets meson just get the right compiler every time.

Very good point, I like that.

> ```meson
> dependency('foo', 'bar', find_library : ['foo'], has_function : 'bar', had_header : 'foo.h', language : 'c')
> ```

Finding the right API is the difficult part indeed, we had many other proposals in the past. 

Your proposal has a few issues too:
- it clusters kwargs on dependency function that already has many.
- we would have to hardcore an order in which we do the lookup. Probably not too bad?
- when you pass arrays to find_library, does it mean AND or OR ? 
- what if different lib would need to check for different header?
- does it mean you need the lib with that func, or func and fallback to lib?
- what if you need to pass include_directories, or any other kwargs of find_library or has_function?

One thing I like about my proposal is each method has the same API as the corresponding function. But I agree mutable is not great. Maybe we should introduce real builder syntax:

```
df = dependency_fallbacks().find_library().has_function()
# df is now immutable
```

Another idea I vaguely had is using a module:
```
df = import('dependency-fallbacks')
dependency('foo', df.find_library(), df.has_function(), ...)
```

eli-schwartz:
IMHO we need to support the case where a dependency can be found by two or more mutually exclusive find_library fallbacks, each with a different library name and header include. There are actual real world cases where that is needed.

dcbaker:
> it clusters kwargs on dependency function that already has many.

Indeed, that's one of the reasons I suggested that having a new, immutable, object like `cc.library_fallback()` or `dependency_fallback()` instead of putting all of the arguments in dependency

>    when you pass arrays to find_library, does it mean AND or OR ?

I would expect that `dependency(a, b, c)` and `dependency(..., find_library : ['a', 'b', 'c'])` would have the same behavior, try `a`, if that fails try `b`, etc.

>   what if different lib would need to check for different header?

There must be at some point that the environment you're checking is so non related that just doing the `required : false` dance becomes the right solution. synactic sugare should make the common case easy, not try to make *every* case easy. I think solving the "library X has many different names, and we want to try them all" is the common case we should be solving for, ie, is is "libz", "libz-1", "z", "z-1", "libz-1.0", "libz-in-rust-1.0", etc.

The only other case that I can think of, is that of "there are seven different libraries that provide a function that does X, and I can use any of them. Which is probably best solved with a for loop anyway:
```meson
for args : [['ssl', {has_function : 'ssl_sha1', has_header : 'ssl.h'}],
                ['openssl', {has_header : 'openssl.h']}],
                ['nettle', {'has_function : 'nettle_sha1', has_header : 'nettle.h'}]]
    lib = cc.find_library(args[0], kwargs : args[1], required : false)
    if lib.found()
        conf.set('HAVE_@0@'.format(args[0].upper()))
        break
    endif
endfor
```

really isn't that big of a difference against
```meson
df = dependency_fallbacks()
df.find_library('ssl', has_function : 'ssl_sha1')
df.find_library('openssl', ...)
df.find_library('nettle', ...')
d = dependency(df)
conf.set('HAVE_@0@'.format(args[0].upper()))
```
And, it works with any version of meson already

> df = dependency_fallbacks().find_library().has_function()

How does meson know that you're done with it though? that's introducing significant to assignment, which is something we don't do generally (except with values the copy-on-assign instead of reference-on-assign). From meson's perspective
```meson
def = dependency_fallbacks()
def = def.find_library()
```
is equivalent to
```meson
def = dependency_fallbacks().find_library()
```

xclaesse:
> How does meson know that you're done with it though? that's introducing significant to assignment, which is something we don't do generally (except with values that copy-on-assign instead of reference-on-assign)

It's easy for meson to set a immutable flag in assign. That would be new concept, but simple to implement. But to be fair I'm not seriously proposing doing it, I prefer mutable than that.

> There must be at some point that the dependeices you're checking are so non related that just doing the `required : false` dance becomes the right solution. synactic sugare should make the common case easy, not try to make _every_ case easy.

That's true but it's still worth investigating if we can find a solution that solves as many cases as possible, and not stop thinking too soon. I think the most important is having something flexible that we can extend. Trying to fit everything into a single function kwargs is going to be hard to extend. That's why I like having a mutable object, or even a whole module, we can always add more methods.

Now that I think again about this proposal, I think I like the most the module API.

PR #8713, opened 2021-04-29
Do not accidentally format files when only checking if they are forma…

jpakkane:
…tted.

### DIFF
--- a/mesonbuild/scripts/clangformat.py
+++ b/mesonbuild/scripts/clangformat.py
@@ -36,12 +36,17 @@ def parse_pattern_file(fname: Path) -> T.List[str]:
     return patterns
 
 def run_clang_format(exelist: T.List[str], fname: Path, check: bool) -> subprocess.CompletedProcess:
+    if check:
+        original = fname.read_bytes()
     before = fname.stat().st_mtime
-    ret = subprocess.run(exelist + ['-style=file', '-i', str(fname)])
+    args = ['-style=file', '-i', str(fname)]
+    ret = subprocess.run(exelist + args)
     after = fname.stat().st_mtime
     if before != after:
         print('File reformatted: ', fname)
         if check:
+            # Restore the original if only checking.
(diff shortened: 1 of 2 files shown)


mesonbuild/scripts/clangformat.py
```diff
@@ -37,7 +37,10 @@ def parse_pattern_file(fname: Path) -> T.List[str]:
 
 def run_clang_format(exelist: T.List[str], fname: Path, check: bool) -> subprocess.CompletedProcess:
     before = fname.stat().st_mtime
-    ret = subprocess.run(exelist + ['-style=file', '-i', str(fname)])
+    args = ['-style=file', str(fname)]
+    if check:
+        args = ['-i'] + args
+    ret = subprocess.run(exelist + args)
```
xclaesse, line 43: You inverted the logic here, should be `if not check`.

It's a bit more complicated than that. When not using `-i` it will print the formatted file on stdout, you have to compare that output with the original file to know if it was modified because the mtime check below won't work since it did not modify the file.

In retrospect relying on mtime was maybe a cheap trick, we should probably never use `-i` and always do the diff between original file and stdout. Then `if not check and there_is_a_diff: rewrite_the_file(stdout)`.

run_unittests.py
```diff
@@ -5658,9 +5658,10 @@ def test_clang_format(self):
             output = self.build('clang-format-check')
             self.assertEqual(1, output.count('File reformatted:'))
 
-        # All code has been reformatted already, so it should be no-op now.
+        # The check format should not touch any files. Thus
+        # running format again has some work to do.
         output = self.build('clang-format')
-        self.assertEqual(0, output.count('File reformatted:'))
+        self.assertEqual(1, output.count('File reformatted:'))
         self.build('clang-format-check')
```
xclaesse, line 5665: nitpicking, could keep the comment `# All code has been reformatted already, so it should be no-op now.` and move it here.

Issue #7344, opened 2020-06-17
Spam in output when checking for ObjC compiler that doesn't exist


tp-m:
**Describe the bug**
When checking whether an ObjC compiler is available Meson prints this if none is found:
```
|C++ linker for the host machine: c++ ld.bfd 2.34
cc: fatal error: cannot execute ‘cc1obj’: execvp: No such file or directory
compilation terminated.
|Compiler for language objc for the host machine not found.
```

That will look pretty disconcerting to users even though there's no problem here.

Is it supposed to be outputting this or is that just a forgotten redirect to `/dev/null` somewhere?

**To Reproduce**
```
project('foo', 'c')
add_languages('objc', native: true, required: false)
```
on a system that does not have an ObjC compiler installed.

**Expected behavior**
It should output just this:
```
|C++ linker for the host machine: c++ ld.bfd 2.34
|Compiler for language objc for the host machine not found.
```

**system parameters**
* Meson 0.54.2 (but also happens with latest from git)

xclaesse:
`ObjCCompiler.sanity_check()` uses `subprocess.Popen` without redirecting stdout/stderr. From a quick grep in compilers that same mistake has been copy/paster in many places.

PR #8717, opened 2021-05-01
Use sanity_check_impl for objc/objpp

rgommers:
This gets rid of compile warnings when running the test suite (see gh-7344), and simplifies the code. Note that `work_dir` in `sanity_check_impl` was incorrect, it was used both to prepend to file names _and_ as `cwd=work_dir` argument to `Popen`. This is fixed here.

Closes gh-7344

Also adds validation for `--only`. This now gives a clear error rather than silently passes for unrecognized categories, like:
```
python run_project_tests.py --only objc  # should be 'objective c'
```
That's just a tiny 2-line usability improvement.

### DIFF
--- a/mesonbuild/compilers/mixins/clike.py
+++ b/mesonbuild/compilers/mixins/clike.py
@@ -313,7 +313,7 @@ class CLikeCompiler(Compiler):
         # Compile sanity check
         # NOTE: extra_flags must be added at the end. On MSVC, it might contain a '/link' argument
         # after which all further arguments will be passed directly to the linker
-        cmdlist = self.exelist + [source_name] + self.get_output_args(binary_name) + extra_flags
+        cmdlist = self.exelist + [sname] + self.get_output_args(binname) + extra_flags
         pc, stdo, stde = mesonlib.Popen_safe(cmdlist, cwd=work_dir)
         mlog.debug('Sanity check compiler command line:', ' '.join(cmdlist))
         mlog.debug('Sanity check compile stdout:')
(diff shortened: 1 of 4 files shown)


rgommers:
There are multiple languages using `sanity_check_impl`, and breaking something there may silently skip more tests. So I compared with another recent PR that the number of skipped tests didn't change, for `vs2019 vc2019x64ninja`:
```
Total passed tests: 433
Total failed tests: 0
Total skipped tests: 201
```

dcbaker (approved):
This is a very nice cleanup, can you fix the unused imports from LGTM though? Then I think this is good to merge.

rgommers:
Thanks for the review. LGTM warnings fixed now.

I do have a workflow question. Say I'm running:
```
python run_project_tests.py --only python
```
and then something inside `_sanity_check_impl` isn't quite working and I want to view the `mlog.debug` output, what's the best way to do that?

What I was seeing is that it goes into a log file into `b e5wqe/meson-private/....`, and that gets cleaned up immediately when the test suite run ends. So what I was doing was change `mlog.debug` to `mlog.warning` to see the command-line output, but there must be a better way.

dcbaker:
I have a patch that skips the cleanup step and leaves the `b ...` directory in place. There's also the (I'm not on my development machine, so the name might not be quite right) run_single_test.py, which can be helpful for looking at one test at a time.

rgommers:
Thanks @dcbaker. Skipping the cleanup step makes sense, I'll do that next time.

The one CI failure is unrelated (a Homebrew download failure for openssl).

Issue #6469, opened 2020-01-15
python get_install_dir returns the wrong directory for a conda env


Zaharid:
**Describe the bug**

When targeting a python installation in a conda environment, meson returns an install directory that has nothing to do with the installation that was found.

Something like

```
message(python_inst.path())
message(python_inst.get_install_dir())
```

returns

```
Message: /tmp/py/bin/python3
Message: /usr/local/lib/python3.8/site-packages/
```

**To Reproduce**

Create a meson.build file like
```
project('Wronginst', 'c')



py3_mod = import('python')
python_inst = py3_mod.find_installation('python3')
message(python_inst.path())
message(python_inst.get_install_dir())
```

Create and activate a conda environment like

```
conda create -p /tmp/testenv meson ninja python pkgconfig

...
The following NEW packages will be INSTALLED:

  _libgcc_mutex      pkgs/main/linux-64::_libgcc_mutex-0.1-main
  ca-certificates    pkgs/main/linux-64::ca-certificates-2019.11.27-0
  certifi            pkgs/main/linux-64::certifi-2019.11.28-py38_0
  ld_impl_linux-64   pkgs/main/linux-64::ld_impl_linux-64-2.33.1-h53a641e_7
  libedit            pkgs/main/linux-64::libedit-3.1.20181209-hc058e9b_0
  libffi             pkgs/main/linux-64::libffi-3.2.1-hd88cf55_4
  libgcc-ng          pkgs/main/linux-64::libgcc-ng-9.1.0-hdf63c60_0
  libstdcxx-ng       pkgs/main/linux-64::libstdcxx-ng-9.1.0-hdf63c60_0
  meson              pkgs/main/noarch::meson-0.52.0-py_0
  ncurses            pkgs/main/linux-64::ncurses-6.1-he6710b0_1
  ninja              pkgs/main/linux-64::ninja-1.9.0-py38hfd86e86_0
  openssl            pkgs/main/linux-64::openssl-1.1.1d-h7b6447c_3
  pip                pkgs/main/linux-64::pip-19.3.1-py38_0
  pkg-config         pkgs/main/linux-64::pkg-config-0.29.2-h1bed415_8
  pkgconfig          pkgs/main/linux-64::pkgconfig-1.4.0-py38_0
  python             pkgs/main/linux-64::python-3.8.1-h0371630_1
  readline           pkgs/main/linux-64::readline-7.0-h7b6447c_5
[...]

PR #5700, opened 2019-07-24
Add a 'use_venv_prefix' option to python module functions

thiblahute:
Making it simpler to work cleanly in virtual environment for pyton projects

Issue #3995, opened 2018-08-08
install_sources installs python sources into wrong directory on windows


fpn:
Standard MSVC python:
```
Python 3.6.6 (v3.6.6:4cf1f54eb7, Jun 27 2018, 03:37:03) [MSC v.1900 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> from distutils import sysconfig
>>> sysconfig.get_python_lib()
'C:\\Python36\\Lib\\site-packages'
```

modules/python.py  - INSTALL_PATHS_COMMAND - sets posix_prefix (on windows)

```
print (json.dumps(sysconfig.get_paths(scheme='posix_prefix', vars={'base': '', 'platbase': '', 'installed_base': ''})))
{"stdlib": "\\lib\\python3.6", "platstdlib": "\\lib\\python3.6", "purelib": "\\lib\\python3.6\\site-packages", "platlib": "\\lib\\python3.6\\site-packages", "include": "\\include\\python3.6", "platinclude": "C:\\Python36\\include\\python3.6", "scripts": "\\bin", "data": "."}
```

syconfig does the right thing when not forced into posix_prefix
```
print (json.dumps(sysconfig.get_paths( vars={'base': '', 'platbase': '', 'installed_base': ''})))
>>> print (json.dumps(sysconfig.get_paths( vars={'base': '', 'platbase': '', 'installed_base': ''})))
{"stdlib": "\\Lib", "platstdlib": "\\Lib", "purelib": "\\Lib\\site-packages", "platlib": "\\Lib\\site-packages", "include": "\\Include", "platinclude": "\\Include", "scripts": "\\Scripts", "data": "."}
```

nirbheek:
I'm not sure I understand the issue here, what are you writing in your build file, what do you expect, and what actually happens?

CCing @MathieuDuponchelle

MathieuDuponchelle:
We probably want to use scheme='nt' on Windows, in any case meson's prefix will be prepended to these, so afaiu if you want the module to do "the right thing" you will probably need to pass a `C:\python36` prefix in that case as well, which is slightly weird :)

fpn:
Hi @MathieuDuponchelle ,

Sorry - I guess there is to much information,

The install should go to:
```
'C:\\Python36\\Lib\\site-packages'
```

Instead it goes to:
```
'C:\\Python36\\lib\\python3.6\\site-packages`
```

[...]

PR #8719, opened 2021-05-02
Add test for py.get_install_dir that fails with default prefix

rgommers:
Shows the issues described in gh-6469:

    message(python_inst.path())
    message(python_inst.get_install_dir())

gives:

    Message: /home/username/anaconda3/envs/meson-dev/bin/python
    Message: /usr/lib/python3.9/site-packages/

The root cause seems to be that the `prefix` variable retrieved at https://github.com/mesonbuild/meson/blob/1ca8fa31ea22df14bf78c90af5ca0b881165ac41/mesonbuild/modules/python.py#L290 gives `/usr`, which is clearly wrong. I'm just not sure what the best way to fix it is.

Submitting as a PR to see if any tests fail (it does fail for me locally).

Any hints on how to debug inside `python.py` would be much appreciated. Trying to set a breakpoint or embed IPython and then running the tests just gives `Reason: Generating the build system failed.` and doesn't actually get me into a debugger.

### DIFF
--- a/test cases/python/1 basic/meson.build
+++ b/test cases/python/1 basic/meson.build
@@ -20,6 +20,12 @@ if not py_platlib.endswith('site-packages')
   error('Python3 platlib path seems invalid? ' + py_platlib)
 endif
 
+py_installdir = py.get_install_dir()
+if not py_installdir.startswith(py_purelib)
+  error('Python3 install dir seems invalid? ' + py_installdir)
+endif
+
+
 main = files('prog.py')
 
 test('toplevel', py, args : main)


rgommers:
If an explicit prefix is given,
```
meson setup build --prefix=/home/user/some/path
cd build
meson introspect --install
```
shows that things do work as expected. So it's just the default prefix that is wrong. 

There'll probably be another bug lurking there for include paths, because of how paths are queried from `sysconfig` through [INTROSPECT_COMMAND](https://github.com/mesonbuild/meson/blob/1ca8fa31ea22df14bf78c90af5ca0b881165ac41/mesonbuild/modules/python.py#L261):
```
In [11]: sysconfig.get_paths(scheme='posix_prefix', vars={'base': '', 'platbase': '', 'installed_base': ''})
Out[11]: 
{'stdlib': '/lib/python3.9',
 'platstdlib': '/lib/python3.9',
 'purelib': '/lib/python3.9/site-packages',
 'platlib': '/lib/python3.9/site-packages',
 'include': '/include/python3.9',
 'platinclude': '/home/rgommers/anaconda3/envs/scipy-meson/include/python3.9',
 'scripts': '/bin',
 'data': '.'}
```

rgommers:
Failing in Azure CI as well, so it's not specific to conda:
```
Program python3 found: YES (C:\hostedtoolcache\windows\Python\3.6.8\x64\python3.EXE)
Message: Python purelib path: C:\hostedtoolcache\windows\Python\3.6.8\x64\Lib\site-packages

test cases\python\1 basic\meson.build:25:2: ERROR: Problem encountered: Python3 install dir seems invalid? x:/usr\lib\python3.6\site-packages\
```

naveen521kk:
I think the test case failing in azure is windows only https://github.com/mesonbuild/meson/issues/3995

rgommers:
> I think the test case failing in azure is windows only #3995 (where it explicitly uses posix scheme idk why)

It's not, it's failing for me on Linux too. 

I'll see if I can push this forward soon, now that I understand a bit more about Meson.

naveen521kk:
> It's not, it's failing for me on Linux too.

Hmm, I think I understand what meson is doing on Linux(also on Windows, but it is wrong). Meson wants to install everything under the prefix passed(`--prefix`) so that it is portable(in the case of Linux it is `/usr` by default). That's because Meson wants to respect the prefix passed by the user. And you can never expect it to install on the virtual environement.

You can simply pass `--prefix` as `sys.prefix` to make the test pass on Linux atleast.

rgommers:
> so that it is portable(in the case of Linux it is /usr by default). ... And you can never expect it to install on the virtual environement.

This simply assumes that the Python install you're using is the system install. And if it's not, then the assumption breaks. You may be installing a package that you're building with Python 3.9, while the system install is Python 3.6 for example.

Each Python interpreter will have its own default location, and that's what should be used. Always using `/usr` is a bug.

naveen521kk:
> This simply assumes that the Python install you're using is the system install. 

No, I don't think so. If you use `py_mod.find_installation('python')` you would get the python which meson is running if you use something like `py_mod.find_installation('python3')` you would get system python though because it will use `pkg-config` for finding python in that case.

> You may be installing a package that you're building with Python 3.9, while the system install is Python 3.6 for example.

I don't think so they would match `/usr/lib/python3.9/site-packages` vs `/usr/lib/python3.6/site-packages`.

> Each Python interpreter will have its own default location, and that's what should be used. Always using /usr is a bug.

Hmm, I feel that everything should be installed to the same `--prefix`; nothing should outside of that. If it uses the python interpreter default location I think It is simply propblematic.

eli-schwartz:
Huh, since commit 0063eb251e836e777b427cbe59b43ab937ac1924 it fails on Debian-like systems simply because https://mesonbuild.com/Python-module.html#get_path does not contain the path which we special-case as an install path due to sysconfig lying on those OSes.

rgommers:
> EDIT: sorry, my bad -- it is fixed by #9156 which is not yet merged.

Thanks, it's merged now. So I'll close this PR.



> There is still the general question of what to do about the current expectation that the install --prefix is also the python sys.prefix -- #5700 had an interesting idea about that.



> Hmm, I feel that everything should be installed to the same `--prefix`; nothing should outside of that.

In that case, it'd be better to error out than write to an obviously incorrect default prefix.

PR #8724, opened 2021-05-03
Add non-mutable container types to mypy

dcbaker:
Meson makes extensive use of caches to speed things up, which can be very helpful. One of the problems with caches in python is the mix of "pass by reference" semantics, and "everything is mutable". This leads to cases where a caller is unaware that it has a cached value, and mutates that value. To help with this I've introduced some new typing protocols. Protocols are a very neat feature of the typing module, they allow for implicit fulfillment, similar to golang's interfaces. This means that if a protocol says "a class must have the following methods", any class implementing at least those methods is a valid candidate. Therefore, I've written a `set` and `list` class that implement only methods that are non-mutative, but we can still return a normal list or set (or ordered set) where an immutable version is expected, and mypy will enforce that we don't call any of the mutative methods of those containers. This gives an extra level of type safety for cached values, but without the runtime overhead of converting them to a tuple or custom class, and without breaking `isinstance` checks.´

They can be used like:

```python

_cache: T.Dict[str, T.List[str]] = {}

def do_something(v: str) -> T.List[str]: ...

def calculate_value(v: str) -> ImmutableList[str]:
    if v in _cache:
       return _cache[v]
    o = do_something(v)
    _cache[v] = o
    return o

a = calculate_value(foo)
a.append('foo')  # mypy error here, ImmutableList doesn't have an `append` method
assert type(a) == list  # true, a is still a plain list
```


mesonbuild/cmake/interpreter.py
```diff
@@ -609,7 +610,7 @@ def _all_lang_stds(self, lang: str) -> T.List[str]:
         for i in res:
             assert isinstance(i, str)
 
-        return res
+        return T.cast(T.List[str], res)
```
mensinda, line 613: This `T.cast` looks unnecessary. The previous asserts should already be enough for mypy.

mesonbuild/_typing.py
```diff
@@ -0,0 +1,124 @@
...
+    def __iter__(self) -> typing.Iterator[T]: ...
+
+    @typing.overload
+    def __getitem__(self, index: int) -> T:...
+    @typing.overload
+    def __getitem__(self, index: slice) -> typing.List[T]: ...
+
+    def __contains__(self, item: T) -> bool: ...
+
+    def __reversed__(self) -> T.Iterator[T]: ...
```
mensinda, line 68: Is the `T.Iterator[T]` intentional or should this be `typing.Iterator[T]`?

mensinda (commented):
I have some minor questions, otherwise, this looks good.

mesonbuild/_typing.py
dcbaker: That should be typing

mesonbuild/cmake/interpreter.py
dcbaker: I'll have to check, but mypy usually treats unnecessary casts as errors, so it must need it for some reason. Maybe i can just declare the type of the variable?

mensinda: I tested it on my system and mypy doesn't complain if I remove the cast.

`reveal_type(res)` looks a bit interesting though:
```
mesonbuild/cmake/interpreter.py:613:21: note: Revealed type is 'Union[builtins.list[Any], builtins.list[builtins.str]]'
```

Specifically, I don't know why the `List[Any]` is still there after the asserts...

dcbaker: yeah, pylance sees the same thing `List | List | List[str]`. I'm not seeing any issues now either, so I'll drop this.

mensinda (approved):
LGTM, assuming all but one CIs pass.

PR #8732, opened 2021-05-06
Small fix to dependency() kwargs

xclaesse:
commit b44b9437f21398fe6af0767860a75a1e12716a3d
Date:   Thu May 6 09:38:35 2021 -0400

    interpreter: Skip not_found_message and include_type in cache id

commit 3d2cbc5c031d0577723c5f783bb704d8cdbe6392
Date:   Thu May 6 09:31:19 2021 -0400

    interpreter: 'embed' kwarg is only valid in python module

### DIFF
--- a/mesonbuild/dependencies/base.py
+++ b/mesonbuild/dependencies/base.py
@@ -2061,6 +2061,9 @@ class DependencyFactory:
 
 def get_dep_identifier(name, kwargs) -> T.Tuple:
     identifier = (name, )
+    from ..interpreter import permitted_dependency_kwargs
+    assert len(permitted_dependency_kwargs) == 19, \
+           'Extra kwargs have been added to dependency(), please review if it makes sense to handle it here'
     for key, value in kwargs.items():
         # 'version' is irrelevant for caching; the caller must check version matches
         # 'native' is handled above with `for_machine`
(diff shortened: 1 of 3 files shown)


mesonbuild/dependencies/base.py
```diff
@@ -2069,7 +2072,10 @@ def get_dep_identifier(name, kwargs) -> T.Tuple:
         #     once a dependency has been found through a fallback, it should
         #     be used for the rest of the Meson run.
         # 'default_options' is only used in fallback case
-        if key in ('version', 'native', 'required', 'fallback', 'allow_fallback', 'default_options'):
+        # 'not_found_message' has no impact on the dependency lookup
+        # 'include_type' is handled after the dependency lookup
+        if key in ('version', 'native', 'required', 'fallback', 'allow_fallback', 'default_options',
+                   'not_found_message', 'include_type'):
```
dcbaker, line 2078: This should really be a set not a tuple.

mesonbuild/dependencies/base.py
```diff
@@ -2061,6 +2061,9 @@ def __call__(self, env: Environment, for_machine: MachineChoice,
 
 def get_dep_identifier(name, kwargs) -> T.Tuple:
     identifier = (name, )
+    from ..interpreter import permitted_dependency_kwargs
+    assert len(permitted_dependency_kwargs) == 19, \
```
dcbaker, line 2065: I'm not entirely sure this is useful, since it wouldn't catch something like `embed`, which is probably pretty important to differentiate.

xclaesse: It's just to be catched when someone adds new kwargs. Maybe just the comment in func_dependency() is enough?

dcbaker: I don't know? I guess it's not hurting anything, I'm just worried that it gives a false sense of security

PR #8735, opened 2021-05-06
A bunch or random assorted fixes pulled out of various branches

dcbaker:
The only thing here that I think would be remotely contraversial is that Vala now doesn't require C in it's languages, it adds that automatically. The actual commit has the justification


jon-turney:
> [Automatically add 'c' to languages when 'vala' is used](/mesonbuild/meson/pull/8735/commits/0368b8b587236f33ce50cc7410e911011c393a2b)

It seems like there should be some documentation change with this, since it is user visible.  But I can't see where, so perhaps a snippet?

dcbaker:
Yeah, I've added a snippet for the Vala/C change.

jpakkane:
```
       # let's pick lld, which isn't the default for either (currently)
>       self._check_ld('ld.lld', 'lld', 'd', 'ld.lld')

run_unittests.py:7925: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
run_unittests.py:7890: in _check_ld
    self.assertEqual(comp.linker.id, expected)
E   AssertionError: 'ld64' != 'ld.lld'
E   - ld64
E   + ld.lld
```

Seems like a real bug but maybe not related to this?

dcbaker:
That doesn't make any sense, I didn't touch any of that code.

dcbaker:
It appears that we have the same problem on master as well… I'm going to go eat lunch then I'll look at ldc on macos I guess.

dcbaker:
So yes, looks like a real bug, but not realted to this PR

test cases/failing/93 vala without c/meson.build
```diff
@@ -1,2 +0,0 @@
-project('vala without c')
-add_languages('vala')
```
xclaesse, line 2: This should be moved to non-failing tests instead of deleting.

xclaesse:
@dcbaker sorry I'm late to the party :(

test cases/failing/93 vala without c/meson.build
dcbaker: Let's just remove c from an existing test, to save on test time. I'll send a patch in s bit

PR #8747, opened 2021-05-08
.C files are now treated as C++ code

Volker-Weissmann:
@jpakkane said in #8239

```
    print a deprecation warning when a project tries to use .C files ("this will change in the future, rename your files, it is already broken on gcc and clang, etc")
    have at least one release with this, see if anyone files bugs
    make the actual switch in a different release
```
Two releases passed and afaik no one filed a bug report.

So please merge.

See also:
#8249


mesonbuild/compilers/compilers.py
```diff
@@ -510,7 +510,7 @@ def __repr__(self) -> str:
     def can_compile(self, src: 'mesonlib.FileOrString') -> bool:
         if isinstance(src, mesonlib.File):
             src = src.fname
-        suffix = os.path.splitext(src)[1].lower()
+        suffix = os.path.splitext(src)[1]
```
jpakkane, line 513: I have a nagging feeling that this will break things for someone. It also requires adding things in many different places. How about just special casing the upper case C? Something like:

```python
suffix = os.path.splitext(src)[1]
if suffix != 'C':
    suffix = suffix.lower()

**Comment by jpakkane** [comment]
Also, one of the C++ test cases should be changed to have a `something.C` file to ensure this keeps working.

File: mesonbuild/compilers/compilers.py (hunk 1 above)
Comment (Volker-Weissmann) [reply mesonbuild/compilers/compilers.py:513]: Ok, I did that.

**Comment by Volker-Weissmann** [comment]
> Also, one of the C++ test cases should be changed to have a `something.C` file to ensure this keeps working.

I added a test

**Comment by jpakkane** [comment]
Which seems to be failing because MSVC treats `.C` as C, not C++. This is why we can't have nice things. Anyone know of a better fix than adding the compile-as-c++ -flag to VS?

Also, we try to limit the number of test directories as each one takes time to run, slowing down running the test suite. Please move that to test dir 2 by adding a new exe target with the `.C` file.

File: mesonbuild/mesonlib/universal.py [hunk 2]
```diff
@@ -350,9 +350,8 @@ def perms_s_to_bits(cls, perms_s: T.Optional[str]) -> int:
         return perms
 
 dot_C_dot_H_warning = """You are using .C or .H files in your project. This is deprecated.
-         Currently, Meson treats this as C code, but this
-            might change in the future, breaking your build.
-         You code also might be already broken on gcc and clang.
+         Currently, Meson treats this as C++ code, but they
+            used to be treated as C code.
```
jpakkane, line 354: This should be extended with something like "The Visual Studio compiler will always treat files with a .C extension as plain C, whereas GCC and Clang will compile them as C++. The compiler flag /TP can be used to manually force Visual Studio to compile all target sources as C++ instead. This is unreliable, so you should consider changing the file extension to an unambiguous one instead."

dcbaker:
I'll defer to @jpakkane here, but I still think this is a bad idea. I would *really* prefer this to be guarded in some way, a `project()` option, or a command line option or something, I'm fairly sure that we're going to get a bug report of "my really ancient C program that has dos FOOBAR.C is suddenly being compiled as C++, this is a regression."

Volker-Weissmann:
> I'll defer to @jpakkane here, but I still think this is a bad idea. I would _really_ prefer this to be guarded in some way, a `project()` option, or a command line option or something, I'm fairly sure that we're going to get a bug report of "my really ancient C program that has dos FOOBAR.C is suddenly being compiled as C++, this is a regression."

A project option would be ok for me. A command line option would not be ok, because I want to be able to tell my users: You can build it using `meson setup build && ninja -C build`.

jpakkane:
> my really ancient C program that has dos FOOBAR.C is suddenly being compiled as C++, this is a regression

That's actually not the case. because we don't change the way it is compiled (because we can't without compiler flags):

- If there were using MSVC, the source has been compiled as C and will be compiled as C
- If they were using gcc/clang, then the source file has already been compiled as C++

The only real difference is that for projects on gcc/clang have to specify `cpp` as a language when building their project (because of the extension detection).

I _really_ don't want to add an option for this. Unfortunately no matter what we do, someone is going to end up unhappy because they have to do something. As much as I I sympathise, the true correct answer to this problem is "rename your source files to have an extension that actually works".

Volker-Weissmann:
> 
> I _really_ don't want to add an option for this. 

Then you should merge this PR.

> As much as I I sympathise, the true correct answer to this problem is "rename your source files to have an extension that actually works".

That's why there is a warning saying that you should not use .C files.

jpakkane:
The warning message text still needs to be updated.

mesonbuild/mesonlib/universal.py
Volker-Weissmann: Done

Volker-Weissmann:
> The warning message text still needs to be updated.

Done

eli-schwartz:
> I have a nagging feeling that this will break things for someone. It also requires adding things in many different places. How about just special casing the upper case C? Something like:

So I take it the desired solution is, anywhere we want to handle C++ `.C` files and check whether file extensions are present in lang_suffixes, we want to:
- check if `C` is there
- check if the extension, lowercased, is there

and we do NOT want to drop the .lower() as a general rule.

Okay, I've got some more cases that need to handle that, see #10629, yay...

PR #8748, opened 2021-05-08
Set up VS environment automatically when run.

jpakkane:


dcbaker:
But that means for me msvc will *never* autoconfigure, since I have clang-cl in my path. Is clang-cl tried before msvc?

jpakkane:
I really don't know. How about we merge this as it is conservative and does not break anything, we can enable things in later MRs as needed?

jon-turney:
> it is conservative and does not break anything

Heh :-)

This appears to be breaking the UnusedMissingReturn (test cases warning flags) workflow on windows
e.g. see https://github.com/mesonbuild/meson/runs/2650696616#step:5:13

I'm not sure if we just want to also check for `gcc` in the path

Sahnvour:
This is awesome, thanks.

> I really don't know. How about we merge this as it is conservative and does not break anything, we can enable things in later MRs as needed?

I really wish we could use it with `clang-cl` out of the box.
Also, is there a way to pass this information along to `ninja` for example ?

Should we open issues about this ?

xclaesse:
>  Also, is there a way to pass this information along to `ninja` for example ?

You should use `meson compile` as a wrapper around ninja that will setup the env. I added a message at the end of meson configuration to tell user about it: https://github.com/mesonbuild/meson/pull/8809.

PR #8750, opened 2021-05-08
little punctuation and add a phrase

mbaykara:
Add the `second` to improve readability.

### DIFF
--- a/docs/markdown/Running-Meson.md
+++ b/docs/markdown/Running-Meson.md
@@ -6,7 +6,7 @@ short-description: Building a project with Meson
 
 There are two different ways of invoking Meson. First, you can run it
 directly from the source tree with the command
-`/path/to/source/meson.py`. Meson may also be installed in which case
+`/path/to/source/meson.py`. Second, Meson may also be installed in which case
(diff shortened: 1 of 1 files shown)


docs/markdown/Running-Meson.md
```diff
@@ -6,7 +6,7 @@ short-description: Building a project with Meson
 
 There are two different ways of invoking Meson. First, you can run it
 directly from the source tree with the command
-`/path/to/source/meson.py`. Meson may also be installed in which case
+`/path/to/source/meson.py`. Second, Meson may also be installed in which case
```
eli-schwartz, line 9: When changing from "Meson may also be installed"

shouldn't it become

"Second, Meson may be installed"?

mbaykara: I would say no, but I am not sure 100%.

dcbaker: I don't think "second" is really the right word here. Maybe "Alternatively, Meson may be installed, in which case…"

dcbaker:
Otherwise this looks like nice changes to me.

PR #8753, opened 2021-05-09
Add general override functionality. Closes #3001. [WIP]

### LINKED ISSUE #3001: Generic overrider functionality

jpakkane:
This is the very beginning of #3001. Currently only allows overriding an option value for all subprojects (not wired to the backend so does not actually work yet). Should permit at least:

- overriding an option value (`c_std`) per-subproject
- overriding options per target (i.e. optimization value for one target)
- whether a target gets built as a shared or static lib

I currently put the definitions inside the native file. This is probably bad. Maybe we should create its own file type and argument for this? Definitely editing the override file should cause a reconfiguration and rebuild as needed.

Ideas and design requirements welcome.

### DIFF
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -374,6 +374,22 @@ class DependencyCache:
     def clear(self) -> None:
         self.__cache.clear()
 
+
+class OptionOverrides:
+    def __init__(self):
+        self.subproject_overrides = {} # Overrides option value in all subprojects but not the main project.
+        # FIXME add per-project overrides here
+
+    def add_subprojects_override(self, key, value):
+        self.subproject_overrides[key] = value
+
+    def value_if_overridden(self, key, subproject): # FIXME add target as an argument?
+        if subproject == '':
+            return None
+        if key not in self.subproject_overrides:
+            return None
+        return self.subproject_overrides[key]
+
 # Can't bind this near the class method it seems, sadly.
 _V = T.TypeVar('_V')
 
@@ -400,6 +416,7 @@ class CoreData:
         self.options: 'KeyedOptionDictType' = {}
         self.cross_files = self.__load_config_files(options, scratch_dir, 'cross')
         self.compilers = PerMachine(OrderedDict(), OrderedDict())  # type: PerMachine[T.Dict[str, Compiler]]
+        self.overrides = OptionOverrides()
 
         # Set of subprojects that have already been initialized once, this is
         # required to be stored and reloaded with the coredata, as we don't
@@ -415,6 +432,7 @@ class CoreData:
         self.config_files = self.__load_config_files(options, scratch_dir, 'native')
         self.builtin_options_libdir_cross_fixup()
         self.init_builtins('')
+        self.init_overrides()
 
     @staticmethod
     def __load_config_files(options: argparse.Namespace, scratch_dir: str, ftype: str) -> T.List[str]:
@@ -540,6 +558,11 @@ class CoreData:
         for for_machine in iter(MachineChoice):
             for key, opt in BUILTIN_OPTIONS_PER_MACHINE.items():
                 self.add_builtin_option(self.options, key.evolve(subproject=subproject, machine=for_machine), opt)
+
+    def init_overrides(self) -> None:
+        sections = MachineFileParser(self.config_files).sections
+        for k, v in sections.get('override_subprojects', {}).items():
+            self.overrides.add_subprojects_override(k, v)
 
     @staticmethod
     def add_builtin_option(opts_map: 'KeyedOptionDictType', key: OptionKey,
--- a/mesonbuild/interpreter/interpreter.py
+++ b/mesonbuild/interpreter/interpreter.py
@@ -858,6 +858,10 @@ external dependencies (including libraries) must go to "dependencies".''')
     def get_option_internal(self, optname: str):
         key = OptionKey.from_string(optname).evolve(subproject=self.subproject)
 
+        maybe_v = self.coredata.overrides.value_if_overridden(optname, self.subproject)
+        if maybe_v is not None:
+            return maybe_v
+
         if not key.is_project():
             for opts in [self.coredata.options, compilers.base_options]:
                 v = opts.get(key)
--- a/run_unittests.py
+++ b/run_unittests.py
@@ -5702,6 +5702,11 @@ class AllPlatformTests(BasePlatformTests):
                 link_args = env.coredata.get_external_link_args(cc.for_machine, cc.language)
                 self.assertEqual(sorted(link_args), sorted(['-flto']))
 
+    def test_overrides(self):
+        testdir = os.path.join(self.unit_test_dir, '96 overriding')
+        native_file = os.path.join(testdir, 'native_overrides.txt')
+        self.init(testdir, extra_args=['--native-file', native_file])
+
 class FailureTests(BasePlatformTests):
     '''
     Tests that test failure conditions. Build files here should be dynamically
--- a/test cases/unit/96 overriding/meson.build
+++ b/test cases/unit/96 overriding/meson.build
@@ -1,0 +1,4 @@
+project('mainprog', 'c')
+
+assert(get_option('buildtype') == 'debug')
+subproject('somesub')
--- a/test cases/unit/96 overriding/native_overrides.txt
+++ b/test cases/unit/96 overriding/native_overrides.txt
@@ -1,0 +1,2 @@
+[override_subprojects]
+buildtype = 'debugoptimized'
--- a/test cases/unit/96 overriding/subprojects/somesub/meson.build
+++ b/test cases/unit/96 overriding/subprojects/somesub/meson.build
@@ -1,0 +1,3 @@
+project('somesub', 'c')
+
+assert(get_option('buildtype') == 'debugoptimized')


jpakkane:
Some rethinking. Override file is now an option so it can be set from the command line allowing you to enable it to an existing build dir or switch between override files. Also added a new option type that point to files and validates that they exist.

eli-schwartz:
It would be extremely nice to be able to override dependency lookups too. In the style of autotools having `$NAME_CFLAGS` and `$NAME_LIBS` except, you know, not an environment variable ;)

jpakkane:
That can be added too. It's mostly a question of syntax.

jpakkane:
Added functionality to override an option for a single subproject rather than all of them.

dcbaker:
@jpakkane I had thought about dependency overrides as well. I think that probably should be in the machine file, since they're a per-machine option, my thought had bee something like:

```ini
[dependencies.zlib]
include_directories = ['/usr/local/include']
c_link_args = ['-L/usr/local/lib', '-lz']
c_args = []
; Additional language args could be placed here if you had both a C and C++ ABI
```

dcbaker:
Also, while we're talking about it, I'd really like to have a single "machine" file, instead of separate cross and native files, an imaginary syntax could be:
```ini
[build.machine]
  system = 'linux'
[build.built-in options]
  c_args = ['-foo']
[host.machine]
  system = 'windows'
[host.dependencies.zlib]
  ...
```
Obviously my sort-of-toml-in-ini isn't a very nice format, but we don't have a ton of options without external dependencies unless you want to use json (I don't advice json for humans to write by hand), or write something by hand. With our already available layering then it's easy to compose them together.

Also, I have a branch that makes *_std a per-subproject option, I didn't quite get it finished, but it's probably useful since currently the standard is part of the ninja compile rule template.

jpakkane:
> Also, I have a branch that makes *_std a per-subproject option

Actually, one of the outcomes of this PR is that we would no longer need per-subproject options. Instead they could be written in the override file. The proliferation of per-subproject options is not a nice thing and makes the number of options explode and complicate things.

dcbaker:
And, having two files sucks. Let's not do that. Let's have one file. just one, with one syntax, one argument, and not have a proliferation of files with slightly different formats.

jpakkane:
This is something I have thought about for a lot and unfortunately there does not seem to be a simple solution. First of all if you put this and cross and native files in a single file then it must be JSON or something else that has multiple hierarchical nested levels. A single flat ini-style thing becomes unwieldy if it has tens of entries. Writing JSON by hand is doable, but not that pleasant.

Further, there are cases where you'd want to switch between different override sets but keep all other info the same. As an example you'd typically want to use the setup described in the doc where the top project is debug but all others are debugoptimized. Then you might want to switch to a different set with additional debugs or warnings. Or build some individual projects as shared rather than static for build speedups. If everything is in the same file, you need to copy all the other definitions as well. If you specify multiple conf files (like with cross files) then specifying `"-Doverride_file=['foo/bar', 'baz']"` by hand gets tedious.

If someone comes up with a nicer UI for this, then we could and should definitely do that.

dcbaker:
Realistically we have three choices:
1. JSON
2. YAML
3. TOML

All of them suck in some way.

Ini stinks because it's flat, underdefined, and there's a ton of variants all called "ini".

JSON is really annoying to write by hand, since it has obnoxious rules about commas. It was designed as a human inspectable language. It's great for machine -> machine situations. I've done a ton of work with JSON, I really don't recomend it for human use. There's a python builtin.

YAML is really big and complicated, it has some odd whitespace interaction between tabs and spaces. It is, on the other hand, very flexible and thanks to github and gitlab quite popular and thus well known. There's several good implementaitons for python. It's also a strict superset of JSON.

TOML is the current darling of the world, it's nice in that it's basically a speced ini format with heiarchy. It's got some odd syntax, it can be a little verbose, but it's generally not bad to write. There's one good implementaiton for python.

The thing they all have in common is that they are structured. You can represent *exactly* the same data in all three formats:
```json
{
  "a": 6,
  "b": [
    1,
    2
  ],
  "c": {
    "x": "foo"
  }
}
```
```yaml
a: 6
b:
  - 1
  - 2
 c:
  x: 'foo'
```
```toml
a = 6
b = [1, 2]

[c]
x = "foo"
```
And, pyyaml, json, and toml all will give you *exactly* the same data structure. So, here's a crazy idea. We could use all of them. JSON always, and one or both of the other two *if* you have the necessary dependencies. That gives us a format we can always accept, even if it's very annoying to write by hand *and* one or two formats that are actually designed to be written by humans that we can accept in most cases (really other than boostrapping).

eli-schwartz:
Toml is flat too. Nesting is defined by encoding the full hierarchy, dot-separated, into each section name. This is not hierarchical, this is flat but with some easy-for-humans-to-mess-up convention for inferring hierarchies. (You can arbitrarily decide the same rule for your project's ini files, leading to an identical experience and no better or worse wieldiness.) It quickly becomes unwieldy.

yaml solves that problem, but does add different problems.

I'm not aware of a single truly good file format. At least configparser is well defined as being configparser, and is in the stdlib!

dcbaker:
> Toml is flat too. Nesting is defined by encoding the full hierarchy, dot-separated, into each section name. This is not hierarchical, this is flat but with some easy-for-humans-to-mess-up convention for inferring hierarchies. (You can arbitrarily decide the same rule for your project's ini files, leading to an identical experience and no better or worse wieldiness.) It quickly becomes unwieldy.

Whether it is unwieldy, it is *speced* that a `.` in a table name is hierarchical. I understand you don't like it, but please, the spec is super clear on this: https://github.com/toml-lang/toml/blob/master/toml.md#user-content-table

The reality is that we need a hierarchical format if we're going to have "one file to rule them all". I'd gladly pick an imperfect format (since, as you say, there is no perfect config format) to have one file with one format than have two files with different formats.

xclaesse:
Currently machine files are ConfigParser but with values interpreted for basic types such as strings, lists, additions. Note that the "ini" format leaves values undefined, so our machine files are still valid ini, just with extra formatting.

I worked a long time ago on bringing the same syntax to wrap files: https://github.com/xclaesse/meson/commit/ab1caf10cf92d87352068009ff6671daee455f55. I think it's important to have the same syntax to all files formats exposed to users.

I personally would like to keep using that format if possible because that's what our users know already. With a bit of structured section names, it shouldn't be that bad, we do that already for options in machine files. If we start using json or something else maybe we should switch wrap files too for consistency.

dcbaker:
I coded up a . separated syntax using our config parser, it's not bad. I need to write a few real tests and I'll push it for review.

dcbaker:
https://github.com/dcbaker/meson/pull/new/wip/2021-05/config-parser-sections

dcbaker:
fully backwards compatible, but allows . separate section headers like toml.

jpakkane:
For reference, here is a copypaste of the syntax:

```ini
[host.machine]
system = 'linux'
endian = 'little'
cpu_family = 'aarch64'
cpu = 'aarch64'

[host.built-in options]
foo = 'bar'

[host.built-in options.zlib]
default_library = 'static'
```

I'm not a fan of the `host.` prefix. Needing to write that everywhere gets tedious. Maybe make it so that the names are `foo` and 'cross.foo`. Thinking further, that only makes sense for machine definitions because there are, for example, no separate built-in options for cross building.

One major question here is what should be done if override and cross files are combined? I'd be inclined to say that that should be prohibited. If you specify an override file then:

- specifying cross files is an error
- per-subproject options can not be defined via command line flags, only by editing the override file
- eventually we could deprecate and remove setting per-subproject options via the command line

The latter is a bit controversial, however I have found that when you have a project with something like 5 subprojects then the output of `meson configure` gets close to unreadable because there is just so much stuff in it. A plain text file that you can modify with an editor is more convenient, not to mention once you have written the file you can use it as a build configuration template later.

As an idea that came to mind: if the user does not specify an override file then we could write an empty one to the build dir and use that. It could have some comments at the top to tell people how to use it.

dcbaker:
I would prefer to have a single file that replaces the separate files. Combing cross/native files with the new file should be prohibited.

I wonder if we could do something like:
```
[Host]
[@machine]
....
```
Where @ means "under the previous section"?, Ie that would be equvalent to [host.machine]

xclaesse:
I personally don't see the problem in having `host.` prefix, we don't have that many sections anyway. But I still totally hate that we have useless split in option sections, a single `[options]` section can do the job.

xclaesse:
```ini
[host.built-in options.zlib]
default_library = 'static'
```

That section name is starting to be ridiculous, the same can be achieved by:

```ini
[host.options]
zlib:default_library = 'static'
```

ntonnaett:
This looks great. The only thing I'm missing is the coherence between subprojects and differently obtained dependencies.
This should work for both:
```
[host.options]
zlib:default_library = 'static'
```
Or there should be the possibility to declare the same for dependencies:
```
[host.options]
zlib:default_library = 'static'
zlib_dep:static = true
```

Maybe it's better to leave them separated because the behavior can differ between pkgconfig and subprojects.

annacrombie:
Why not have the config file be in the meson dsl?  You can already evalutate some meson expressions / assign and retrieve variables.

```meson
# like json, but with expressions, variables, no weird comma rules
{
  'a': 6,
  'b': [1, 2],
  'c': {
    'x': "foo"
  }
}

# or, define with functions

set_options(
  for: 'host',
  machine: machine_options(
    system: 'linux'
    endian: 'little'
    cpu_family: 'aarch64'
    cpu: 'aarch64'
  ),
  built_in: {
    'foo': 'bar',
    'zlib:default_library': 'static',
  },
)

# or anything in between!
```

Pros:
- no extra dependency to parse
- familiar syntax by default
- it is clear that expressions / variables can be used
- flexible ways to compose / express hierarchies

Cons:
- too flexible?
- difficult / impossible for non-meson programs to parse (is this a problem?)

Issue #8744, opened 2021-05-08
0.58.0: install target fails in gtkdoc helper script


kloczek:
After upgrade from 0.57.2 to 0.58.0 I've started noticing strange issues in few projects like grillo and vte.
`install` targe fails with:
```console
+ cd grilo-0.3.13
+ DESTDIR=/home/tkloczko/rpmbuild/BUILDROOT/grilo-0.3.13-4.fc35.x86_64
+ /usr/bin/meson install -C x86_64-redhat-linux-gnu --no-rebuild
Installing src/libgrilo-0.3.so.0.313.1 to /home/tkloczko/rpmbuild/BUILDROOT/grilo-0.3.13-4.fc35.x86_64/usr/lib64
Installing src/Grl-0.3.gir to /home/tkloczko/rpmbuild/BUILDROOT/grilo-0.3.13-4.fc35.x86_64/usr/share/gir-1.0
Installing src/Grl-0.3.typelib to /home/tkloczko/rpmbuild/BUILDROOT/grilo-0.3.13-4.fc35.x86_64/usr/lib64/girepository-1.0
Installing libs/net/libgrlnet-0.3.so.0.313.0 to /home/tkloczko/rpmbuild/BUILDROOT/grilo-0.3.13-4.fc35.x86_64/usr/lib64
Installing libs/net/GrlNet-0.3.gir to /home/tkloczko/rpmbuild/BUILDROOT/grilo-0.3.13-4.fc35.x86_64/usr/share/gir-1.0
Installing libs/net/GrlNet-0.3.typelib to /home/tkloczko/rpmbuild/BUILDROOT/grilo-0.3.13-4.fc35.x86_64/usr/lib64/girepository-1.0
Installing libs/pls/libgrlpls-0.3.so.0.313.0 to /home/tkloczko/rpmbuild/BUILDROOT/grilo-0.3.13-4.fc35.x86_64/usr/lib64
Installing libs/pls/GrlPls-0.3.gir to /home/tkloczko/rpmbuild/BUILDROOT/grilo-0.3.13-4.fc35.x86_64/usr/share/gir-1.0
Installing libs/pls/GrlPls-0.3.typelib to /home/tkloczko/rpmbuild/BUILDROOT/grilo-0.3.13-4.fc35.x86_64/usr/lib64/girepository-1.0
[...]

PR #8757, opened 2021-05-11
gnome: Fix gtkdoc generation

xclaesse:
install_scripts used to replace @BUILD_ROOT@ and @SOURCE_ROOT@ but it
    was not documented and got removed in Meson 0.58.0. gnome.gtkdoc() was
    relying on that behaviour, but it has always been broken in the case the
    source or build directory contains spaces.
    
    Fix this by changing get_include_args() to substitue paths directly
    which will then get escaped correctly.
    
    Add a unit test that builds GObject documentation which is where this
    issue has been spotted.
    
    Fixes: #8744


mesonbuild/backend/backends.py
```diff
@@ -1429,7 +1429,19 @@ def generate_target_install(self, d: InstallData) -> None:
...
+        result: T.List[ExecutableSerialisation] = []
+        srcdir = self.environment.get_source_dir()
+        builddir = self.environment.get_build_dir()
+        for i in self.build.install_scripts:
+            fixed_args = []
+            for a in i.cmd_args:
+                a = a.replace('@SOURCE_ROOT@', srcdir)
+                a = a.replace('@BUILD_ROOT@', builddir)
+                fixed_args.append(a)
+            es = copy.copy(i)
```
heftig, line 1441: The `import copy` needs to be restored, as well.

xclaesse: You're right. Changed my implementation now.

xclaesse:
@heftig can you give this new path a try please?

heftig:
Seems to work fine with glib and gnome-builder. Thanks!

jon-turney:
I guess the test this adds supersedes #7274

xclaesse:
@jon-turney oh, I didn't know you made a PR for that already. I think that PR can be closed now, yes.

Issue #8758, opened 2021-05-11
need an example(sample code) on how to use meson's devenv commad


nitinosiris:
I don't know how to use meson's devenv , so i need one example :)

nitinosiris:
I want to use meson devenv for setting PATH, GI_TYPELIB_DIR, and LD_LIBRARY_PATH.

nitinosiris:
don't know how to code for that :( thats why i need an example

xclaesse:
You probably don't have anything to code:
- `GI_TYPELIB_DIR` is set automatically when using `gnome.generate_gir()`.
- `LD_LIBRARY_PATH` is set automatically to all places where you have an installed shared library.
- `PATH` is set automatically to all places where you have an installed executable.

Just type `meson devenv -C builddir`, and see the env that has been set for you.

nitinosiris:
Ohh, okay .I will try it :) thanks!

PR #8760, opened 2021-05-12
doc: Add example how to use devenv

xclaesse:
Fixes: #8758


xclaesse:
@nitinosiris let me know if that helps.

nitinosiris:
> @nitinosiris let me know if that helps.

Thanks

PR #8764, opened 2021-05-13
intel-cl fortran: recognize oneAPI LLVM and Classic compilers

scivision:
Intel oneAPI was officially released January 2021. The Linux / Mac just works with Meson so far as I see in casual tests.
For Windows oneAPI is broken because [xilink is deprecated and removed](https://software.intel.com/content/www/us/en/develop/articles/porting-guide-for-icc-users-to-dpcpp-or-icx.html?wapkw=xilink).
Meson still works on Windows and Mac because simply "ifort" is invoked for linking. That also works for Windows when done by hand. 

```sh
ifort simple.exe.p\simple.f90.obj
```
works for Windows Intel-Cl like on Linux.

@dcbaker do you have thoughts about this?

I think we might detect oneAPI for Windows, and treat it separately from the pre-oneAPI Intel compiler, which is soon to be deprecated and replaced by oneAPI. Instead of the Xilink functions we can make oneAPI on Windows link very similar to Linux/Mac I think.

### DIFF
--- a/mesonbuild/environment.py
+++ b/mesonbuild/environment.py
@@ -1587,7 +1587,8 @@ class Environment:
                         compiler, version, for_machine, is_cross, info,
                         exe_wrap, full_version=full_version, linker=linker)
 
-                if 'Intel(R) Visual Fortran' in err:
+                if any(s in err for s in ['Intel(R) Visual Fortran', 'Intel(R) Fortran Compiler',
+                                          'Intel(R) Fortran Intel(R) 64 Compiler Classic']):
                     version = search_version(err)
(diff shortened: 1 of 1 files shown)


scivision:
Without this PR, ifort simply fails to be found on Windows. With this PR, code compiles but fails to link since xilink.exe is deprecated.

dcbaker:
I've honestly been putting off dealing with the oneAPI compilers :)

I know that they're LLVM based, but I'm not sure what the interface is like (I know ifort is still pretty much the same, not sure about C and C++). We probably need to just run the `guess_x_Compiler` logic and then fall back to 'then try xilink approach I guess.

Issue #8770, opened 2021-05-17
Missing CPU Family: csky


yatescr:
**Describe the bug**
Building an image under yocto produces an error:

randyy@Lubuntu-25316:build-csky-toolchain-kernel_5.12$ bitbake core-image-minimal -c populate_sdk
Loading cache: 100% |###############################################################################################################################################################################################################################################################################################| Time: 0:00:00
Loaded 1357 entries from dependency cache.
NOTE: Resolving any missing task queue dependencies

Build Configuration:
BB_VERSION           = "1.48.0"
BUILD_SYS            = "x86_64-linux"
NATIVELSBSTRING      = "universal"
TARGET_SYS           = "csky-poky-linux"
MACHINE              = "qemucsky"
DISTRO               = "poky"
DISTRO_VERSION       = "3.2.2"
TUNE_FEATURES        = " littleendian"
meta                 
meta-poky            
meta-yocto-bsp       
meta-csky            = "<unknown>:<unknown>"

Initialising tasks: 100% |##########################################################################################################################################################################################################################################################################################| Time: 0:00:01
Sstate summary: Wanted 730 Found 0 Missed 730 Current 463 (0% match, 38% complete)
NOTE: Executing Tasks
ERROR: xorgproto-2020.1-r0 do_configure: meson failed
ERROR: xorgproto-2020.1-r0 do_configure: Execution of '/edg/build-csky-toolchain-kernel_5.12/tmp/work/csky-poky-linux/xorgproto/2020.1-r0/temp/run.do_configure.18580' failed with exit code 1:

ERROR: Unknown CPU family csky, see https://wiki.yoctoproject.org/wiki/Meson/UnknownCPU for directions.
WARNING: exit code 1 from a shell command.

ERROR: Logfile of failure stored in: /edg/build-csky-toolchain-kernel_5.12/tmp/work/csky-poky-linux/xorgproto/2020.1-r0/temp/log.do_configure.18580
Log data follows:
| DEBUG: Executing python function extend_recipe_sysroot
[...]

PR #8771, opened 2021-05-17
add cpu family csky

yatescr:
This is to support the relatively new CPU architecture CSky. See https://c-sky.github.io/

### DIFF
--- a/docs/markdown/Reference-tables.md
+++ b/docs/markdown/Reference-tables.md
@@ -86,6 +86,7 @@ set in the cross file.
 | arm                 | 32 bit ARM processor     |
 | avr                 | Atmel AVR processor      |
 | c2000               | 32 bit C2000 processor   |
+| csky                | 32 bit CSky processor    |
 | dspic               | 16 bit Microchip dsPIC   |
 | e2k                 | MCST Elbrus processor    |
 | ia64                | Itanium processor        |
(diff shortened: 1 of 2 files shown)


dcbaker:
Can c-sky host a meson build natively, or is only cross compiling currently supported?

jpakkane:
Also, does csky come in both 32 and 64 bit variants and if yes, how should they be named in Meson?

yatescr:
Dylan,

I would eventually hope that it could host a build natively. I am at
the state where I am creating a new image for the processor/board in
yocto and have not yet tried to include meson in the rootfs and have
not been able to try a build with it on the actual hardware (or qemu).
My first step now is to just get the yocto toolchain to build, which
requires a simple image to be built (core-image-minimal).

--Randy

yatescr:
Jussi,

No, it is only available in 32-bit.

However, there are several flavors of csky processor: the 810, 803, etc.
See the cpu-csky.c file in
https://sourceware.org/git/?p=binutils-gdb.git;a=blob_plain;f=bfd/cpu-csky.c;hb=HEAD

--Randy

dcbaker:
For my question then, if it's not at the point of self-hosting then I think this is fine as-is. If it can self host we want to make sure that we return the string `'csky'` when auto-detecting.

jpakkane:
Thanks.

eli-schwartz:
Fixes #8770, please mention this in the future so we can ensure issue tickets get closed when implemented. :D If you link it in the commit message/first comment then github will autolink it as `closed by ticket <whatever>` which is helpful...

yatescr:
Will do, sorry for the hassle! Thanks for handling this so quickly!

--Randy

PR #8773, opened 2021-05-17
More enhancements for Rust + clippy support

dcbaker:
This pull request fixes a few bugs for Rust, plus implements a number of missing compiler abstractions like werror, warn arguments, and pic. Finally it adds support for using clippy, which is a rust linter that wraps rustc, but doesn't provide the necessary information in it's output for Meson to figure that out.

### DIFF
--- a/mesonbuild/compilers/rust.py
+++ b/mesonbuild/compilers/rust.py
@@ -196,3 +196,20 @@ class RustCompiler(Compiler):
         # Rustc currently has no way to toggle this, it's controlled by whether
         # pic is on by rustc
         return []
+
+
+class ClippyRustCompiler(RustCompiler):
+
+    """Clippy is a linter that wraps Rustc.
+
+    This just provides us a different id
+    """
+
+    def __init__(self, exelist: T.List[str], version: str, for_machine: MachineChoice,
+                 is_cross: bool, info: 'MachineInfo',
+                 exe_wrapper: T.Optional['ExternalProgram'] = None,
+                 full_version: T.Optional[str] = None,
+                 linker: T.Optional['DynamicLinker'] = None):
+        super().__init__(exelist, version, for_machine, is_cross, info,
+                         exe_wrapper, full_version, linker)
+        self.id = 'clippy-driver rustc'
--- a/mesonbuild/compilers/rust.py
+++ b/mesonbuild/compilers/rust.py
@@ -168,3 +168,8 @@ class RustCompiler(Compiler):
         for a in super().get_linker_always_args():
             args.extend(['-C', f'link-arg={a}'])
         return args
+
+    def get_werror_args(self) -> T.List[str]:
+        # Use -D warnings, which makes every warning not explicitly allowed an
+        # error
+        return ['-D', 'warnings']
--- a/mesonbuild/compilers/rust.py
+++ b/mesonbuild/compilers/rust.py
@@ -187,3 +187,12 @@ class RustCompiler(Compiler):
 
     def get_no_warn_args(self) -> T.List[str]:
         return self._WARNING_LEVELS["0"]
+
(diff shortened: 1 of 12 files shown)


mesonbuild/compilers/rust.py
```diff
@@ -191,6 +192,18 @@ def get_pie_args(self) -> T.List[str]:
         # pic is on by rustc
         return []
 
+    def get_lto_compile_args(self, *, threads: int = 0, mode: str = 'default') -> T.List[str]:
+        # Enable cross langauge lto
+        args = ['-C', 'linker-plugin-lto=yes']
```
evverx, line 197: I forgot to mention that `-C linker-plugin-lto=yes` doesn't seem to be compatible with `gcc`: https://github.com/systemd/systemd/pull/19598#discussion_r633071853

In a script compiling systemd on GitHub Actions `-C linker-plugin-lto=yes` is passed only when `clang` is used:
```
if [[ "$COMPILER" == "clang" ]]; then
      ARGS+=("--optimization=3 -Db_lto=true -Drust_args='--deny warnings -Copt-level=3 -Clinker-plugin-lto'")
else
      ARGS+=("--optimization=3 -Db_lto=true -Drust_args='--deny warnings -Copt-level=3'")
fi
```

dcbaker: Yeah, I saw that too. I couldn't come up with any incantation that would do any kind of lto with gcc support, and I thought that gcc and clang's lto implementations were not generally compatible anyway? I could be wrong thought that basically gcc acted on gimpl, and llvm acted on LLVM-ir, that they embed in the .o files.

evverx: I have to say I haven't figured out how `lto`, `linker`, `linker-flavor` and `linker-plugin-lto` interact with each other and why, for example,
```
meson -Drust_args="-C lto=yes"  -Db_lto=true -Dbuild-rust=true build
```
works with both `gcc` and `clang` without `-Clinker-plugin-lto`. Would it be possible to put the "lto" commits on hold (so that it was possible for the systemd CI to keep building systemd with `-Db_lto=true` with both `gcc` and `clang`?

dcbaker: Okay, I finally found the actual documentation. without -C linker-plugin-lto rustc performs LTO on the .o file it generates, but then does no interprocdedural optimization at link time (calling it IPO makes more sense than LTO, I guess ICC was smarter than GCC and Clang on this). Defering will result in better optimization, but requires that all of the compielrs be LLVM based and use LTO.

What a pain.

dcbaker: I've dropped the LTO commit, I'll try to sort that out later. Now I get the fun of starting up Windows.

evverx: I can imagine a scenario where I need both "IPO" and the IPO deferred to the link time in C/Rust projects. Now I'm puzzled. I'm not sure what `-Db_lto=true` should do.

dcbaker: You can always manually set -D${lang}_args and -D${lang}_link_args and ignore the builtins, but yeah, I think the goal of `-Db_lto` is to provide a mechanism that works as much as possible, and provides the thing most people want. It sounds like from the docs though that IPO at object generation time (or really, when rust takes the .rs files and outputs a .a, .so, etc) is strictly a subset of what happens with link-time IPO.

dcbaker: So, I don't think you actually need both, use link-time IPO if possible, but .o time is better than no IPO. Rustc does the .o IPO by default.

jpakkane:
Test failure seems relevant:

```
================================== FAILURES ===================================
______________________ AllPlatformTests.test_rust_clippy ______________________
```

PR #8775, opened 2021-05-18
Structured inputs for Rust (and more languages in the future, probably)

### LINKED ISSUE #8725: Filesystem significant languages and build directory implementation details
Meson has a strict rule that the layout of the build directory as an implementation detail, this makes a lot of sense as it gives us (Meson) flexibility to change the layout when it makes sense (we've done this a number of times, including with things like shortening the names of generated directories, or adding the private dir). This is also completely fine for languages like C in which the filesystem layout is an implementation detail of the language, and can easily be fixed with `include_directories()`. There are however, a growing number of languages supported by Meson (Rust, Python), or with pending support in Meson (Cython, Zig), in which the filesystem layout *is* the import layout. This introduces real problems when trying to mix generated and static sources, as the compilers generally expect a single, unified tree.

I'd like to propose then a way to provide sources as a tree, but in an abstract way such that Meson can make configure time decisions about what it wants to do to fulfill these requirements. Specifically, I'm proposing:

```meson
executable(
  'foo',
  {
    '' : 'main.rs',
    'foo' : [generated_rs],
    'foo/bar' : ['other.rs'],
  }
)
```
This specifically would tell Meson that when compiling a file main.rs, it must be in a directory such that there is a subdir foo, which contains the output of `generated_rs` (whatever that happens to be), and foo must have a subdir bar, with `other.rs`. Meson could at configure time decide that all of the inputs are static (non-generated) files, and do nothing, or it could decide that since `generated_rs` is a `custom_target`, to copy/link all of the files into a new directory and compile them there.

This solves the problem that Rust, Zig, and Cython have of needing a single tree of filres, but also creates an abstraction such that Meson can continue to hide the actual layout of the build dir, as this tree could exist anywhere within the build dir.

dcbaker:
The fixed issue has most of the background on this, so I'm not going to repeat it, go look at #8725 for more info.

The goal of this series is to make it possible to use generated files with languages that do all of their compilation in one go (Rust, Cython, Zig), and rely on the layout of those files on disk for imports (Rust, Cython, Zig). For those languages we don't generate a single .o for each source file, the compilers and/or languages simply don't make that possible. Instead, we generate a singlle compilation target for all of those sources.

To that end this series introduces a new concept, the "structured input". This input is a dictionary mapping an arbitrary "root" to a directory tree (at the moment the target's private directory is used as the root), like so:
```meson
executable(
  'exe',
  {
    '': 'main.rs',
    'foo' : 'mod.rs',
    'bar': generated_mod.rs,
  }
)
```
Meson will ensure that at the root of the tree is main.rs, then a root/foo/mod.rs exists, as does root/bar/generated_mod.rs, it will do this before compiling mian.rs If necessary it will copy files (at build time) into the build tree. It does this at build time to avoid reconfiguring on file changes, and to ensure that the tree is always accurate. This is unnecessary if there are no generated sources, and simply passing `executable('exe', 'main.rs')` has not been changed.

Fixes #8725


jpakkane:
```
executable(
  'exe',
  {
    '': 'main.rs',
    'foo' : 'mod.rs',
    'bar': generated_mod.rs,
  }
)
```

This is _unacceptable_ as a syntax. It is completely unreadable with the magic dictionaries and all.

dcbaker:
Then do you have another idea on what the syntax should be?

dcbaker:
@jpakkane, how would you feel about a new function instead?:

```meson
sources = structured_source(
  foo_gen_pyx, 'foo_static.pyx',
  sub : bar_gen_pyx, 'bar_static.pyx',
  ...
)

target = py.module('foo', sources)
```

jpakkane:
Sure, something like that makes a lot more sense as it is its own isolated thing.

dcbaker:
This is going to need enough rework that I'm just giong to open a new PR for this.

PR #8781, opened 2021-05-20
Better detection of Visual Studio compiler

Jacob-Nielsen:
setup_vsenv() failed for me as:
1. I have VS 2021 and 2017 installed and both in non-standard locations (paths without spaces in them)
2. Meson defaulted to VS 2017 but I use VS 2012 and had called setvars in VS 2012 before calling Meson.

I implemented two additional checks:
- one for the VSINSTALLDIR environment variable
- one that checks for "cl.exe" in PATH when the environment variable OS is Windows_NT

### DIFF
--- a/mesonbuild/mesonmain.py
+++ b/mesonbuild/mesonmain.py
@@ -54,6 +54,16 @@ def setup_vsenv():
         return
     if 'Visual Studio' in os.environ['PATH']:
         return
+    # VSINSTALL is set when running setvars from a Visual Studio installation
+    # Tested with Visual Studio 2012 and 2017
+    if os.environ.get('VSINSTALLDIR', bat_placeholder) != '%VSINSTALLDIR%':
+        return
+    # Check explicitly for cl when on Windows
+    # Windows XP, 2000 and 10 reports 'Windows_NT'. I assume that all other Windows versions in between do the same.
+    if (os.environ.get('OS', bat_placeholder) == 'Windows_NT'):
+        if shutil.which('cl.exe'):
+            return
+
     bat_locator_bin = r'c:\Program Files (x86)\Microsoft Visual Studio\Installer\vswhere.exe'
     if not os.path.exists(bat_locator_bin):
         return
--- a/mesonbuild/mesonmain.py
+++ b/mesonbuild/mesonmain.py
@@ -56,7 +56,7 @@ def setup_vsenv():
         return
     # VSINSTALL is set when running setvars from a Visual Studio installation
     # Tested with Visual Studio 2012 and 2017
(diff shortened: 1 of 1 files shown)


mesonbuild/mesonmain.py
```diff
@@ -54,6 +54,16 @@ def setup_vsenv():
         return
     if 'Visual Studio' in os.environ['PATH']:
         return
+    # VSINSTALL is set when running setvars from a Visual Studio installation
+    # Tested with Visual Studio 2012 and 2017
+    if os.environ.get('VSINSTALLDIR', bat_placeholder) != '%VSINSTALLDIR%':
```
jpakkane, line 59: Shouldn't this be `if 'VSINSTALLDIR' in os.environ`?

xclaesse: While at it, I noticed that bat_placeholder is useless, `os.environ.get('VSINSTALLDIR')` with no default returns `None` which is fine to compare with a string.

jpakkane: It was not leftover. [It was intentional}(https://youtu.be/kK4H-LkrQjQ?t=127).

Jacob-Nielsen: changed the check to "in os.environ"

mesonbuild/mesonmain.py
```diff
@@ -54,6 +54,16 @@ def setup_vsenv():
         return
     if 'Visual Studio' in os.environ['PATH']:
         return
+    # VSINSTALL is set when running setvars from a Visual Studio installation
+    # Tested with Visual Studio 2012 and 2017
+    if 'VSINSTALLDIR' in os.environ:
+        return
+    # Check explicitly for cl when on Windows
+    # Windows XP, 2000 and 10 reports 'Windows_NT'. I assume that all other Windows versions in between do the same.
+    if (os.environ.get('OS', bat_placeholder) == 'Windows_NT'):
```
Jacob-Nielsen, line 63: There is no need to check for Windows (and not in this way) when there already is a "mesonlib.is_windows()" check in the start of the function. I will remove that in yet another commit

mesonbuild/mesonmain.py
Jacob-Nielsen: I wonder why the shutil.which checks at the beginning does not include all the entries in "default_c" from environment.py but only 'cc', 'clang' and 'clang-cl'

xclaesse: > It was not leftover. [It was intentional](https://youtu.be/kK4H-LkrQjQ?t=127).

ahah, ok :)

jpakkane:
As per the documentation, [pull requests should never have merge commits](https://mesonbuild.com/Contributing.html#keeping-pull-requests-up-to-date). I can get rid of them this time, just be aware of this in the future.

PR #8782, opened 2021-05-20
Fail upon encountering unknown options, add --allow-unknown-options

### LINKED ISSUE #7288: Instead of just printing a warning, Meson should abort and exit with an error code if an unknown build option is used
**Describe the bug**

**To Reproduce**

```
$ meson -Dfoo=bar build-foo/
The Meson build system
Version: 0.52.1
Build type: native build
Project name: barproject
Project version: 0.1
Host machine cpu: x86_64
Build targets in project: 9
```

**Expected behavior**

Flowdalic:

Meson now fails if an unknown option is encountered. The previous
behavior can be restored by using the --allow-unknown-options switch.

Fixes #7288.

This is an alternative to https://github.com/mesonbuild/meson/pull/8658.

### DIFF
--- a/mesonbuild/mconf.py
+++ b/mesonbuild/mconf.py
@@ -36,6 +36,9 @@ def add_arguments(parser: 'argparse.ArgumentParser') -> None:
     parser.add_argument('builddir', nargs='?', default='.')
     parser.add_argument('--clearcache', action='store_true', default=False,
                         help='Clear cached state (e.g. found dependencies)')
+    parser.add_argument('--allow-unknown-options', action='store_false',
+                        dest='error_unknown',
+                        help='Do *not* fail if unknown build options are encountered')
 
 def make_lower_case(val: T.Any) -> T.Union[str, T.List[T.Any]]:  # T.Any because of recursion...
     if isinstance(val, bool):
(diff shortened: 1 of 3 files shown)


Flowdalic:
Meh, some tests appear to rely on unknown options being ignored. :/

mesonbuild/mconf.py
```diff
@@ -36,6 +36,9 @@ def add_arguments(parser: 'argparse.ArgumentParser') -> None:
     parser.add_argument('builddir', nargs='?', default='.')
     parser.add_argument('--clearcache', action='store_true', default=False,
                         help='Clear cached state (e.g. found dependencies)')
+    parser.add_argument('--allow-unknown-options', action='store_false',
+                        dest='error_unknown',
+                        help='Do *not* fail if unknown build options are encountered')
```
xclaesse, line 41: I think in mconf.py it should always be fatal, you set options after configuration, the exact list of available options is known at this point.

xclaesse (commented):
There are other cases we could also make fatal:
- `meson setup -Dnonexistant=true`, this PR only deal with `meson configure`. In this case we probably need `--allow-unknown-options`.
- project(..., default_options: ['nonexistant=true']). But probably not for backend (`b_*`) and compiler options (`<lang>_*`).

But those are not really blocker for this PR, could also be done later.

In any case, also miss doc and release notes snippet.

bonzini:
#7288 has been closed and unknown options are fatal, the PR is not needed.

PR #8783, opened 2021-05-20
vala: Test that adding C manually isn't required

dcbaker:
I deleted the test that enforced that this failed, but having a test that enforces it passes would be good too.


xclaesse:
LGTM, not sure what's going on with macos CI :/

dcbaker:
Apparently ldc has changed the way it handles linkers. I need to get out my macbook and look, but it's slow and old and it's hard to find motivation

PR #8788, opened 2021-05-22
MacOS fixes

jpakkane:
These are needed to make the unit test suite run on the Apple laptop I have.


jpakkane:
Windows failures are unrelated.

PR #8793, opened 2021-05-23
Fix double negative in 'No CMake binary not found'

jon-turney:


jon-turney:
At the moment this messge says `No CMake binary for machine 1 not found. Giving up.`.

This PR fixes the double negative, but `self.for_machine` should probably be converted from an enum to text, as well.

Issue #8754, opened 2021-05-10
[RFC] Simplified wrapdb


xclaesse:
I have been thinking and experimenting a bit with ways to improve our wrapdb workflow.

Current issues:
- The worflow to contribute new wraps requires manual steps that only @jpakkane can do (create new git repo, create releases).
- The namespace under github.com/mesonbuild/ is a mess with over a hundred projects.
- It requires running a (complex) python script server. It has to be hosted somewhere which has been an issue in the past. Even if it seems pretty stable now.
- For projects that has meson build system upstream we still have to create a git repo containing a single upstream.wrap file. Currently wrapdb force generating a patch tarball that contains erronous license file.

My proposal: https://github.com/xclaesse/wrapdb
- Move everything into a single git repository.
- All commits are preserved, I wrote a script that import git history from all individual repos we had.
- Contributing can be done with a simple PR, the CI takes care of creating release tarballs once merged. No more scripts manually run.
- All previous release tarballs have been imported, sha256 are still the same.
- No need of hosting a server, everything is just static files on github. However, we'll need to setup redirects to keep using https://wrapdb.mesonbuild.com domain.
- All wraps and their files are in `subprojects/` directory, so we can actually build them for testing.

Example workflow:
- Here is an example PR to add gtk4.wrap: https://github.com/xclaesse/wrapdb/pull/2
- When that PR has been merged, CI has run: https://github.com/xclaesse/wrapdb/runs/2546577289?check_suite_focus=true
- Release tag has been created automatically: https://github.com/xclaesse/wrapdb/releases/gtk_4.3.0-1
- Note that it does not contain a patch zip because upstream has meson build system already.

TODO:
- Change `meson wrap` command to use new system. We'll have keep old wrapdb server running for a while for people running older meson version.
- Add CI that actually build wrap subprojects, at least for those we can easily reproduce. It can be used for Meson RC validation maybe.

Needed redirects:
- Old patch URL must keep working, e.g. https://wrapdb.mesonbuild.com/v1/projects/zlib/1.2.11/5/get_zip -> https://github.com/xclaesse/wrapdb/releases/download/zlib_1.2.11-5/zlib_1.2.11-5_patch.zip
- We don't want to hardcode github.com URLs in meson. To access release list, we should add a redirect: https://wrapdb.mesonbuild.com/v2/releases.txt -> https://raw.githubusercontent.com/xclaesse/wrapdb/master/releases.txt.
[...]

PR #8796, opened 2021-05-25
wrap: Port to v2 protocol

xclaesse:
Fixes: #8754.

### DIFF
--- a/mesonbuild/wrap/wraptool.py
+++ b/mesonbuild/wrap/wraptool.py
@@ -19,8 +19,9 @@ import shutil
 import typing as T
 
 from glob import glob
-
-from .wrap import API_ROOT, open_wrapdburl
+from urllib.parse import urlparse
+from urllib.request import urlopen
+from .wrap import WrapException
 
 from .. import mesonlib
 
@@ -57,38 +58,30 @@ def add_arguments(parser: 'argparse.ArgumentParser') -> None:
     p.add_argument('project_path')
     p.set_defaults(wrap_func=promote)
 
-def get_result(urlstring: str) -> T.Dict[str, T.Any]:
-    u = open_wrapdburl(urlstring)
-    data = u.read().decode('utf-8')
-    jd = json.loads(data)
-    if jd['output'] != 'ok':
-        print('Got bad output from server.', file=sys.stderr)
-        raise SystemExit(data)
-    assert isinstance(jd, dict)
-    return jd
-
-def get_projectlist() -> T.List[str]:
-    jd = get_result(API_ROOT + 'projects')
-    projects = jd['projects']
-    assert isinstance(projects, list)
-    return projects
+def get_releases() -> T.Dict[str, T.Any]:
+    url = urlopen('https://wrapdb.mesonbuild.com/v2/releases.json')
+    return T.cast(T.Dict[str, T.Any], json.loads(url.read().decode()))
 
 def list_projects(options: 'argparse.Namespace') -> None:
-    projects = get_projectlist()
-    for p in projects:
+    releases = get_releases()
+    for p in releases.keys():
         print(p)
 
 def search(options: 'argparse.Namespace') -> None:
     name = options.name
-    jd = get_result(API_ROOT + 'query/byname/' + name)
-    for p in jd['projects']:
-        print(p)
+    releases = get_releases()
+    for p in releases.keys():
+        if p.startswith(name):
+            print(p)
 
 def get_latest_version(name: str) -> tuple:
-    jd = get_result(API_ROOT + 'query/get_latest/' + name)
-    branch = jd['branch']
-    revision = jd['revision']
-    return branch, revision
+    releases = get_releases()
+    info = releases.get(name)
+    if not info:
+        raise WrapException(f'Wrap {name} not found in wrapdb')
+    latest_version = info['versions'][0]
+    version, revision = latest_version.rsplit('-', 1)
+    return version, revision
 
 def install(options: 'argparse.Namespace') -> None:
     name = options.name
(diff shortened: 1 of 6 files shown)


jpakkane:
Could we do an option flag for this instead? That is, if you do, `meson wrap foo` it will use v1 and `meson wrap --v2 foo` would use version 2? In this way we could merge this faster and can finish the server bits and once everything is done and working delete the v1 code?

xclaesse:
@jpakkane we could, but I don't think it's really needed, the server side is already done. We'll have to keep the v1 server for a while to keep older meson version working.

xclaesse:
The plan I suggest:
1. Rename https://github.com/mesonbuild/wrapdb_v2test to https://github.com/mesonbuild/wrapdb (that repo already exists but is basically empty).
2. Update the 301 redirect rules on the server for the new github URL, that's trivial, I have the access.
3. Merge this PR to make "meson wrap" commands use the new v2 URLs. Older versions of Meson will keep using v1 server. Older .wrap files still will have v1 patch URL that use the same server.
4. (Optional) Update server to use 301 redirect for v1 patch URLs as well.
5. (Optional, after a while, at least a couple meson releases) Decommission the v1 server which will make "meson wrap" commands fail on older meson versions, but patch URL will keep working with simple redirect rules.

xclaesse:
Actually, the only missing piece atm is porting the website https://wrapdb.mesonbuild.com. My plan is to generate static github pages from releases.json in the CI.

jpakkane:
Note that there have been updates to wrapdb since the test repo was created.

xclaesse:
> Note that there have been updates to wrapdb since the test repo was created.

I have a script that import releases, I'll just run it again, not a problem.

xclaesse:
For https://wrapdb.mesonbuild.com my plan is to generate some hotdoc markdown that could be merged directly into the main mesonbuild.com website, but I know nothing about hotdoc... Maybe @MathieuDuponchelle could help.

mesonbuild/wrap/wraptool.py
```diff
@@ -57,38 +58,30 @@ def add_arguments(parser: 'argparse.ArgumentParser') -> None:
...
-        raise SystemExit(data)
-    assert isinstance(jd, dict)
-    return jd
-
-def get_projectlist() -> T.List[str]:
-    jd = get_result(API_ROOT + 'projects')
-    projects = jd['projects']
-    assert isinstance(projects, list)
-    return projects
+def get_releases() -> T.Dict[str, T.Any]:
```
jpakkane, line 61: Eventually this should be cached locally for, say, 1 hour.

jpakkane:
[Something's not quite right there.](https://github.com/mesonbuild/wrapdb_v2test/pull/1/checks?check_run_id=2712001711)

xclaesse:
> [Something's not quite right there.](https://github.com/mesonbuild/wrapdb_v2test/pull/1/checks?check_run_id=2712001711)

I know, was running my import script at the same time, and it screwed things. Long story short, my script cannot import new updates from wrapdb v1 into the existing wrapdb_v2test repo. We have to first decide v1 is RO then run the import script from scratch into a clean new wrapdb repo.

mesonbuild/wrap/wraptool.py
xclaesse: We could do that, yes. I would like to keep it simple for now and get it merged first, we don't have a cache for v1 neither so it's not a regression anyway.

OOC, is there a way in python std to get platform specific user cache dir?

jpakkane: [I don't think there is](https://stackoverflow.com/questions/13184414/how-can-i-get-the-path-to-the-appdata-directory-in-python).

xclaesse:
All wraps have been imported into https://github.com/mesonbuild/wrapdb. I think this PR is ready to be merged.

xclaesse:
@jpakkane I added the script that generate the wrapdb table here. That way when we push new releases into wrapdb we can just trigger Meson CI to regenerate the website. We'll probably just have to redirect wrapdb.mesonbuild.com to that page.

jpakkane:
The version list in the  output looks a bit cluttered. Maybe the latest version should be emphasized somehow and separated from the rest?

Screenshot 2021-06-05 013130

jpakkane:
Or maybe only have the latest version on that page? Old versions are probably not that interesting. And the page title and name should be something more descriptive like "WrapDB packages".

xclaesse:
> Or maybe only have the latest version on that page? Old versions are probably not that interesting.

I think older upstream versions could be useful, but we could keep only latest revision for each upstream version?

xclaesse:
What about that

image

It keeps only the latest revision per upstream release, and highlight the first (latest) one.

xclaesse:
TBF, I'm not against keeping only the latest version, anyone has an opinion?

We can always tweak that page later, I mostly wanted to infrastructure in place atm.

jpakkane:
```
tools/regenerate_docs.py:120:1: error: Function is missing a type annotation
tools/regenerate_docs.py:136:24: error: Incompatible types in assignment (expression has type "str", variable has type "List[str]")
tools/regenerate_docs.py:154:5: error: Call to untyped function "generate_wrapdb_table" in typed context
```

tp-m:
I think old versions **are** interesting and we should continue to make them accessible in some way. I can think of multiple reasons, e.g.
1. we don't track API/ABI versions, different versions can have different APIs/ABIs. It will depend on the lib if that's the case or not.
2. the latest version(s) might have certain known bugs/regressions (implementation-wise, or build-wise) that one might want to avoid for now, thus sticking to an older version.

Having said that, I don't think they need to be exposed by default. There could be a global "show older versions" thing, or an expander arrow for each package so that only the latest version is shown and older ones are shown only if the expander is clicked.

jpakkane:
Having the newest one in bold is good enough for me.

xclaesse:
>     2\. the latest version(s) might have certain known bugs/regressions

That's true, and it means we probably should display older revisions too. I tried to make an expander (we can add <details> HTML in the markdown), but not great results. Let's go with just highlight the first in the list, that can be improved later if people have great ideas.

xclaesse:
@jpakkane mypy issues fixed.

Issue #7582, opened 2020-08-12
gnome.mkenums_simple() doesn't work if implicit_include_directories : false


ystreet:
The crux of the problem is that meson generates includes assuming that the source directory is part of the include list like so:

```
#include "gstvkapi.h"
```

with this code:

```
for hdr in sources:
    fhead += '#include "%s"\n' % os.path.basename(str(hdr))
```

One potential fix is to generate a build to source directory path and prepend that to each included file or even determine that mapping at runtime for built sources passed to `mkenums_simple()`.

QuLogic:
This also breaks the case of source files in subdirectories of the target. Those would normally not have the subdirectory in the include path, regardless of where `mkenums_simple` was called.  Meson usually uses relative paths to the source directory, so using a plain base name is unusal.

matthiasclasen:
Also see https://github.com/mesonbuild/meson/issues/2969

xclaesse:
Made a PR: https://github.com/mesonbuild/meson/pull/10855, but you still need to have the subdir where you call mkenum_simple() in your include directories. But it does not have to be the same directory than where you compile the generated .c file.

PR #8798, opened 2021-05-26
Use relative paths to headers in mkenums_simple.

QuLogic:
This fixes `#include` when the source headers are in a different directory to the final target, while not having `include_directories` for the other directory (which otherwise works by having a subdirectory in `#include` lines). It also fixes the case of `implicit_include_directories: false`.

### DIFF
--- a/mesonbuild/modules/gnome.py
+++ b/mesonbuild/modules/gnome.py
@@ -1452,12 +1452,7 @@ class GnomeModule(ExtensionModule):
         c_file_kwargs = copy.deepcopy(mkenums_kwargs)
         if 'sources' not in kwargs:
             raise MesonException('Missing keyword argument "sources".')
-        sources = kwargs['sources']
-        if isinstance(sources, str):
-            sources = [sources]
-        elif not isinstance(sources, list):
-            raise MesonException(
-                'Sources keyword argument must be a string or array.')
+        sources = self.interpreter.source_strings_to_files(kwargs['sources'])
 
         # The `install_header` argument will be used by mkenums() when
         # not using template files, so we need to forcibly unset it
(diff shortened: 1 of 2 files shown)


mesonbuild/modules/gnome.py
```diff
@@ -1477,7 +1472,8 @@ def mkenums_simple(self, state, args, kwargs):
             fhead += '%s\n' % body_prefix
         fhead += '#include "%s"\n' % hdr_filename
         for hdr in sources:
-            fhead += '#include "%s"\n' % os.path.basename(str(hdr))
+            relpath = hdr.rel_to_builddir(os.path.join(os.path.relpath('.', state.subdir), state.build_to_src))
```
xclaesse, line 1475: This does not work `hdr.rel_to_builddir(state.build_to_src)`? Everywhere else `rel_to_builddir` is used that's what is done.

xclaesse:
A unit test would be ideal ;-)

mesonbuild/modules/gnome.py
QuLogic: That produces a path relative to the build _root_. There's no guarantee the produced `enums` source will be in the root, and in fact, the `mkenums` test case does not have it there.

QuLogic:
The existing test does exercise that the path is correct, but does not exercise that the path must be expanded like this. I think I can toggle `implicit_include_directories` on one of them to trigger #7582.

mesonbuild/modules/gnome.py
xclaesse: Oh, I see, wouldn't be more clear as `os.path.join(hdr.rel_to_builddir(state.build_to_src), state.subdir)`? Or that's not the same?

xclaesse:
> I think I can toggle `implicit_include_directories` on one of them to trigger #7582.

Yes, can you check please that it would reproduce the bug, and that the PR fix it?

xclaesse:
This issue resurfaced recently, I made a new PR: https://github.com/mesonbuild/meson/pull/10855.

PR #8804, opened 2021-05-26
Only try to get RSP syntax if RSP is supported

### LINKED ISSUE #8800: ERROR: xxxxx does not implement rsp format ?
I get this error on two of my builds for xc16 and c2000 targets, I'm not sure why?

"ERROR: ar2000 does no implemnt rsp format, this shouldn't be called"
"ERROR: xc16-ar does no implemnt rsp format, this shouldn't be called"

Both projects had been working on Meson 0.57.1. I have just upgraded to meson 0.58.0 and its the first error I get

I seem to have found this spot in the code, mentioning it's not implemented? Is that the case?
`
If `self.can_linker_accept_rsp()` returns True, then this needs to        
be implemented        
"""        
assert not self.can_linker_accept_rsp(), f'{self.id} linker accepts RSP, but doesn\' provide a supported format, this is a bug'     
raise mesonlib.EnvironmentException(f'{self.id} does no implemnt rsp format, this shouldn\'t be called')

`

dcbaker:
Otherwise we get an exception as the method is not implemented.

Fixes #8800

### DIFF
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -1831,6 +1831,17 @@ int dummy;
         # Introspection information
         self.create_target_source_introspection(target, swiftc, compile_args + header_imports + module_includes, relsrc, rel_generated)
 
+    def _rsp_options(self, tool: T.Union['Compiler', 'StaticLinker', 'DynamicLinker']) -> T.Dict[str, T.Union[bool, RSPFileSyntax]]:
+        """Helper method to get rsp options.
+
+        rsp_file_syntax() is only guaranteed to be implemented if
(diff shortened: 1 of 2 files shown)


jon-turney:
I think these NinjaRule() constructors in ninjabackend.py are the only places that `rsp_file_syntax()` and `can_?_accept_rsp()` are called from, so combining these by adding NONE to the RspFileSyntax enum would be a way to solve this making things simpler rather than more complex...

dcbaker:
@jon-turney i think your suggestion is a good one, aka i started to implement it. What I'm having trouble with is verifying that I'm not breaking all the compilers i can't check again

PR #8807, opened 2021-05-28
Vala: Add missing FeatureNew() when C is missing

xclaesse:
It was previously an hard error, only permitted since 0.59.0.


dcbaker (approved):
LGTM

Issue #2622, opened 2017-11-14
"cl" is required to be in PATH even for --backend vs2017


mojca:
After some successful usage of meson on "unix", I tried to use it on Windows. I have Visual Studio 2017 installed, but getting this to work is quite problematic.

The `vs2017` backend creates a project file, so it's not entirely clear to me why meson needs to have access to `cl` from `PATH`. And even if it does: couldn't it somehow find it itself given that it knows it should be looking for the binary inside VS2017 installation? In conrast `cmake` never asks for `cl` to be present when generating Visual Studio projects.

The most weird thing is that generating the project files for Visual Studio worked just fine on Mac. But I guess it didn't hardcode `clang` compiler into the project? An additional (independent) problem is that I didn't install all SDKs an the one that `meson` required was not available (it was an older Windows 8 SDK), so I had to manually retarget all projects. And for some reason I was unable to even run the build of `REGEN` afterwards due to `Access to the path 'z:\path\to\project\build\msvs\meson-private\' is denied.` (this was generated on Mac).

I'll try to figure out where `cl` lives and add it to path to see if that works any better.

nirbheek:
The way to use Meson on Windows with MSVC currently is to run it from inside the Visual Studio toolchain console. That calls `vcvarsall.bat` for you with the correct arguments and sets the environment needed to use MSVC.

cmake does not require this because it handles all that for you. We should indeed add something similar to cmake's `-G` argument so you don't have to do this. Note, however, that `--backend=vs*` is required to generate vcxproj files, but you can also use the ninja backend with MSVC.

gulgi:
Adding a copy of https://github.com/Microsoft/vswhere to the windows version would solve all that, right?

nirbheek:
> Or maybe even better would be to have a generic --backend=vs that creates the correct visual studio project based on the environment.

fwiw, this already exists.

> Adding a copy of https://github.com/Microsoft/vswhere to the windows version would solve all that, right?

We can't use that code directly, but thanks for pointing it out, we should be able to look at it to figure out the best way to find Visual Studio.

nirbheek:
Jonathan Blow published a C++ file (MIT licensed) that does the same job as vswhere with easier to understand code: 

https://pastebin.com/3YvWQa5c

See `find_visual_studio_by_fighting_through_microsoft_craziness()` in that paste. We may need to do some COM API calls through Python.

nirbheek:
[...]

PR #8808, opened 2021-05-28
Find Visual Studio Build Tools 2019 also

naveen521kk:
Got the idea from [setuptools](https://github.com/pypa/setuptools/blob/a5131f0b82e098da6c07a03a47f36f3a52f73fb6/setuptools/msvc.py#L176)

https://github.com/mesonbuild/meson/issues/2622#issuecomment-850486068

Works locally, I have only `Visual Studio Build Tools 2019` installed.

### DIFF
--- a/mesonbuild/mesonmain.py
+++ b/mesonbuild/mesonmain.py
@@ -62,21 +62,31 @@ def setup_vsenv():
     if shutil.which('cl.exe'):
         return
 
-    bat_locator_bin = r'c:\Program Files (x86)\Microsoft Visual Studio\Installer\vswhere.exe'
-    if not os.path.exists(bat_locator_bin):
+    root = os.environ.get("ProgramFiles(x86)") or os.environ.get("ProgramFiles")
+    bat_locator_bin = pathlib.Path(root, 'Microsoft Visual Studio/Installer/vswhere.exe')
+    if not bat_locator_bin.exists():
         return
-    bat_json = subprocess.check_output([bat_locator_bin, '-latest', '-format', 'json'])
+    bat_json = subprocess.check_output(
+        [
+            str(bat_locator_bin),
+            '-latest',
+            '-prerelease',
+            '-requiresAny',
+            '-requires', 'Microsoft.VisualStudio.Component.VC.Tools.x86.x64',
+            '-requires', 'Microsoft.VisualStudio.Workload.WDExpress',
+            '-products', '*',
+            '-format',
+            'json'
+        ]
+    )
     bat_info = json.loads(bat_json)
-    if not bat_info:
-        bat_json = subprocess.check_output([bat_locator_bin, '-prerelease', '-format', 'json'])
-        bat_info = json.loads(bat_json)
     if not bat_info:
         # VS installer instelled but not VS itself maybe?
         return
     print('Activating VS', bat_info[0]['catalog']['productDisplayVersion'])
-    bat_root = bat_info[0]['installationPath']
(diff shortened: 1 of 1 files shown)


mesonbuild/mesonmain.py
```diff
@@ -62,21 +62,31 @@ def setup_vsenv():
...
         return
-    bat_json = subprocess.check_output([bat_locator_bin, '-latest', '-format', 'json'])
+    bat_json = subprocess.check_output(
+        [
+            str(bat_locator_bin),
+            '-latest',
+            '-prerelease',
+            '-requiresAny',
+            '-requires', 'Microsoft.VisualStudio.Component.VC.Tools.x86.x64',
+            '-requires', 'Microsoft.VisualStudio.Workload.WDExpress',
```
xclaesse, line 76: I think we only require `Microsoft.VisualStudio.Component.VC.Tools.x86.x64`, what's `Microsoft.VisualStudio.Workload.WDExpress` ?

If I understand correctly, with `-requiresAny` it will accept any of those 2, so that probably still works even if I don't have WDExpress installed, but if I have WDExpress installed but not Tools.x86.x64 it will work but won't have the C++ toolchain. If I read the [doc](https://docs.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-express?view=vs-2017) correctly WDExpress only contains ARM and ARM64 toolchains, really weird...

Disclaimer: I'm not a VisualStudio expert by any mean, I could be totally wrong :D

mesonbuild/mesonmain.py
```diff
@@ -62,21 +62,31 @@ def setup_vsenv():
...
+            '-products', '*',
+            '-format',
+            'json'
```
xclaesse, line 79: While at it, according to some random stackoverflow I found, we don't need to parse the whole json, just add `-property installationPath` and it will print only the path.

naveen521kk: Ok.

mesonbuild/mesonmain.py
naveen521kk: I think I should remove `Microsoft.VisualStudio.Workload.WDExpress`.

> If I understand correctly, with -requiresAny it will accept any of those 2, so that probably still works even if I don't have WDExpress installed, but if I have WDExpress installed but not Tools.x86.x64 it will work but won't have the C++ toolchain.

I think this is how it is going to work.

mesonbuild/mesonmain.py
naveen521kk: We are printing the version also
https://github.com/mesonbuild/meson/blob/dc5d3277ce599436b8763f3affb5f32671ed06d3/mesonbuild/mesonmain.py#L86
so I think it is fine to parse the json.

xclaesse:
Looks good to me, LGTM.

edsky:
@naveen521kk @xclaesse

It’s not friendly for non-English, there are errors,
Need to add -utf8 parameter

as follows
```
    bat_json = subprocess.check_output(
        [
            str(bat_locator_bin),
            '-utf8', # add this
            '-latest',
            '-prerelease',
            '-requiresAny',
            '-requires', 'Microsoft.VisualStudio.Component.VC.Tools.x86.x64',
            '-products', '*',
            '-format',
            'json'
        ]
```
Otherwise there will be the following error
`UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc3 in position 674: invalid continuation byte`

xclaesse:
@edsky could you open a PR please?

PR #8809, opened 2021-05-28
vsenv: Recommend using "meson compile" wrapper

xclaesse:
When meson has setup the VS environment, running ninja to build won't
work, user should use meson wrapper to compile.


xclaesse:
@naveen521kk Since you use vsenv, what do you think about this?

naveen521kk:
LGTM

Issue #3551, opened 2018-05-09
Add a way to check for the presence of a Python module


ebassi:
Some projects depend on the availability of a Python module for things like generating documentation or running tests. Typically, with Autotools, you'd use `AX_PYTHON_MODULE`. It would be useful to have something like this for Meson as well.

I'm not entirely sure about the shape this would take; either a method on the `python|python3` modules, e.g.

```meson
  python3 = import('python3')
  python3.find_module('dbusmock', version: '>= 0.10')

  python2 = import('python2')
  mako = python2.find_module('mako', required: false)
  build_docs = mako.found()
```

Or have a dependency object:

```meson
dbusmock = dependency('dbusmock', version: '>= 0.10', modules: 'python3')
mako = dependency('mako', modules: 'python', required: false)
```

nirbheek:
CCing @MathieuDuponchelle

nirbheek:
Personally, I like the `has_module()` approach, which would return a `bool`. Not sure if returning a dependency object is useful or meaningful.

ebassi:
@nirbheek Yeah, a dependency object is kind of wasted on this; it's not like it's going to be used like any other dependency objects. Unless, of course, we think we should rebuild the documentation if the installed version of mako changes, which would be slightly weird but not completely unexpected.

MathieuDuponchelle:
I think going through dependency() would make more sense, as far as I understand that is the purpose of the module argument, and while the dependency object would obviously not bring in any compile-time arguments, it could indeed be passed through depends to build targets that need to be rebuilt when the module changes or is removed.

nbenitez:
Adding here two ways to detect python modules, found on `meson.build` files of current projects. So we can choose the preferred method to implement this feature:

https://gitlab.gnome.org/World/PasswordSafe/blob/master/meson.build#L30
https://gitlab.gnome.org/GNOME/gtk-doc/blob/master/meson.build#L40

MathieuDuponchelle:
@nbenitez , note that the problem here isn't the implementation of the feature, but rather the exposed interface :)

a-andre:
The check if a python module is available was [added to version 0.51.0](https://github.com/mesonbuild/meson/blob/master/docs/markdown/Release-notes-for-0.51.0.md#new-modules-kwarg-for-pythonfind_installation). The only thing missing is a way to specify the required version of the dependency.

MathieuDuponchelle:
[...]

PR #8811, opened 2021-05-28
New method `find_package` for python module

naveen521kk:
This introduces a method `.find_package()` for object returned by `import('python').find_installation`. `.find_package()` accepts 
a position argument for project name two keyword argument `version` and `required`. Default value or `required` is `True`. The object it returns has two methods `.found()` and `.version()`.

For example,
```meson
project('python package sample test')

py_mod = import('python')
py = import('python').find_installation('python')
a = py.find_package('setuptools', version:'>=1.0', required: true)

if a.found()
    message('setuptools is found version: '+a.version())
endif
```
would give output
```
The Meson build system
Version: 0.58.999
Source dir: H:\mesonstuff\meson\venvtest
Build dir: H:\mesonstuff\meson\venvtest\_b
Build type: native build
Project name: python package sample test
Project version: undefined
Host machine cpu: x86_64
Program python found: YES (H:\mesonstuff\meson\.venv\Scripts\python.EXE)
Python Package setuptools(>=1.0) found: YES (47.1.0)
Message: setuptools is found version: 47.1.0
Build targets in project: 0

Found ninja-1.10.0.git.kitware.jobserver-1 at "C:\Users\User\AppData\Roaming\Python\Python37\Scripts\ninja.EXE"
```
Fixes https://github.com/mesonbuild/meson/issues/3551

### DIFF
--- a/mesonbuild/modules/python.py
+++ b/mesonbuild/modules/python.py
@@ -15,6 +15,7 @@
 import os
 import json
 import shutil
+import textwrap
 import typing as T
 
 from pathlib import Path
@@ -24,7 +25,8 @@ from . import ExtensionModule
 from ..interpreterbase import (
     noPosargs, noKwargs, permittedKwargs,
     InvalidArguments,
-    FeatureNew, FeatureNewKwargs, disablerIfNotFound
+    FeatureNew, FeatureNewKwargs, disablerIfNotFound,
+    InterpreterObject, ObjectHolder
 )
 from ..interpreter import ExternalProgramHolder, extract_required_kwarg, permitted_dependency_kwargs
 from ..build import known_shmod_kwargs
@@ -32,7 +34,8 @@ from .. import mlog
 from ..environment import detect_cpu_family
 from ..dependencies.base import (
     DependencyMethods, ExternalDependency,
-    PkgConfigDependency, NotFoundDependency
+    PkgConfigDependency, NotFoundDependency,
+    DependencyException
 )
 from ..programs import ExternalProgram, NonExistingExternalProgram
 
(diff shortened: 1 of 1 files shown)


mesonbuild/modules/python.py
```diff
@@ -461,6 +478,65 @@ def get_variable_method(self, args, kwargs):
...
+            raise MesonException(f"Expected a string instead found {type(want_package)}")
+
+        command_importlib = textwrap.dedent("""\
+            try:
+                # only 3.8+
+                from importlib.metadata import version
+            except (ImportError,ModuleNotFoundError):
+                # try using backport
+                from importlib_metadata import version
+                # fail if it is not available
```
eli-schwartz, line 502: Deferred until meson requires python 3.8, meson.build features cannot depend on thirdparty modules.

eli-schwartz (changes requested):

mesonbuild/modules/python.py
naveen521kk: I think even when meson requires python 3.8+, meson would be supporting installing on old python version (say py2), do you think this wait is needed?

eli-schwartz: Python 3.7 is EOL on 2023-06-27 so likely not until around then...

naveen521kk: Hmm, then can `meson.build` feature depend on the python version it is on? What I am saying is make `.find_package()` work only when host python `>=3.8` or else it will simply fail.

naveen521kk: also, if this pr is going to wait till `2023-06-27`, I would be closing it(It would be a lot better to implement it that time rather than having this).

PR #8817, opened 2021-06-01
Split dependencies.base

mensinda:
The first commit moves the CMake, Config-tool, Pkg-Config, etc. classes out of `dependencies.base` into separate files. Additionally, the factory and detection logic is also moved. With these changes, the size of `dependencies.base` is drastically reduced and it only contains the base classes.

~The second commit moves the custom dependency classes into a new subfolder to more clearly separate them from the more generic code.~

Finally, the remaining code in `dependencies.base` is fully annotated.


mensinda:
OK, looks like `vc2019x64vs` is broken. See [my comment](https://github.com/mesonbuild/meson/pull/8512#issuecomment-852189987) in #8512

xclaesse:
I like a lot the split into separate files.

> The second commit moves the custom dependency classes into a new subfolder to more clearly separate them from the more generic code.

Not sure that part is really needed, that seems arbitrary separation making some 2nd class citizens, no? I personally wouldn't mind having all files into the dependencies/ directory, but as you wish :)

mensinda:
> Not sure that part is really needed, that seems arbitrary separation making some 2nd class citizens, no? I personally wouldn't mind having all files into the dependencies/ directory, but as you wish :)

It was primarily to reduce the clutter from the new files introduced by the split. The idea is that more general modules go into `dependencies` and the more specialiced ones (boost, etc.) custom.

dcbaker:
I agree with @xclaesse, this change overall is nice, but I don't like custom split. BoostDependency isn't really any different than PkgConfigDependency, they're both specializations of ExternalDependency.

mensinda:
I guess I am outnumbered then :) I have removed the commit.

mesonbuild/dependencies/base.py
```diff
@@ -347,32 +339,32 @@ def get_partial_dependency(self, *, compile_args: bool = False,
         if not sources:
             new.sources = []
         if not includes:
-            new.include_directories = []
+            pass # TODO maybe filter compile_args?
```
dcbaker, line 342: I have a patch that fixes this for pkg-config, it's really cmake that I'm not sure how to fix... maybe you know how :)

dcbaker (approved):
This looks good to me. I only skimmed the more mechanical parts, but the type annotations look good. Thanks for doing this.

dcbaker:
I went ahead and merged this, as the chances of having conflicts with this is pretty high.

mesonbuild/dependencies/base.py
mensinda: Just filtering for `self.compile_args` for anything starting with `-I` (or `-isystem`) would also work for CMake (not sure how you are implementing it for pkg-config). The CMake dependency class does detect the include directories separately from the other compiler arguments. However, it then just adds them to `compile_args` and forgets about them.

Ideally, we should separate the include directories out from the normal `compile_args`, but this would be another big refactoring...

PR #8819, opened 2021-06-01
ci: Automatically update website when pushing to master

xclaesse:


xclaesse:
@jpakkane something like this should work. You need to add your personal access token into meson repository's secrets and name it `ACCESS_TOKEN`.

xclaesse:
I think if we add `on: workflow_dispatch` we can then trigger that workflow from wrapdb repository with a curl command:

```
curl \
  -X POST \
  -H "Accept: application/vnd.github.v3+json" \
  -H "Authorization: token ${{ secrets.ACCESS_TOKEN }}" \
  https://api.github.com/repos/mesonbuild/meson/actions/workflows/website.yml/dispatches \
  -d '{"ref":"master"}'
```

xclaesse:
@jpakkane as suggested on IRC, the safest way would be to create a new dummy github user (e.g. `meson-bot`) and create an access token for that user. That way if it gets compromised at least we don't compromise our personal github accounts.

jpakkane:
> @jpakkane as suggested on IRC, the safest way would be to create a new dummy github user (e.g. meson-bot) and create an access token for that user. That way if it gets compromised at least we don't compromise our personal github accounts.

I was under the impression that you can create an access token that can do that one thing. Need to dig in the docs...

jpakkane:
Access tokens, no. But it seeems [SSH deploy keys can be used](https://www.webfactory.de/blog/use-ssh-key-for-private-repositories-in-github-actions).

jpakkane:
I have created a deployment key with the secret id `WEBSITE_PRIV_KEY`.

xclaesse:
@jpakkane Updated the PR accordingly, will see how it goes.

jpakkane:
Well I got the page build warning email I always get when updating the web site so I'm guessing it's working then.

xclaesse:
Had to sneak a typo fix: https://github.com/mesonbuild/meson/commit/4516b7a8a93bb1e6d4683359aafedb3917d9e998. It worked, yes: https://github.com/mesonbuild/meson/runs/2728982237.

PR #8822, opened 2021-06-01
Rewrite the Qt module for type safety!

dcbaker:
This basically rewrites the implementation of the qt module using the type checking decorators, then creates three new, separate functions and deprecates the `preprocess` method.

All I really wanted to do was type check generators, but here we are…

I know the history is a mess, I can clean that up if people like.


dcbaker:
and further down the rabbit hole I go...

docs/markdown/Qt5-module.md
```diff
@@ -3,27 +3,97 @@
...
+## compile_moc
+
+*New in 0.59.0*
+
+Compiles Qt's moc files (.moc) into header and/or source files. At least one of
+the keyword arguments `headers` and `sources` must be provided.
+
+It takes no positional arguments, and the following keyword arguments:
+  - `sources` (File | string)[]: A list of sources to be transpiled int .cpp files
+  - `headers` (File | string)[]: A list of sources to be transpiled int .h files
```
deepbluev7, line 44:
```suggestion
  - `sources` (File | string)[]: A list of sources to be transpiled into moc_*.cpp files
  - `headers` (File | string)[]: A list of headers to be transpiled into moc_*.cpp files
```

Moc always generates cpp files, that then need to be compiled.

I think those are both compiled into .cpp files though?

docs/markdown/Qt5-module.md
```diff
@@ -3,27 +3,97 @@
 The Qt5 module provides tools to automatically deal with the various
 tools and steps required for Qt. The module has two methods.
 
+## compile_resources
+
+*New in 0.59.0*
+
+Compiles Qt's resources files (.qpa) into c++ files for compilation.
```
deepbluev7, line 10:
```suggestion
Compiles Qt's resource collection files (.qrc) into c++ files for compilation.
```

See https://doc.qt.io/qt-5/resources.html

docs/markdown/Qt5-module.md
dcbaker: The existing code transforms the headers .moc files, and the sources into `.cpp` files

deepbluev7: Interesting, I would have maybe expected both, not just a moc file. Especially since if I actually run it on a project of mine, it does generate a moc_*.cpp file?

dcbaker: Wierd.... the headers are transpiled into `.cpp` files and the sources into `.moc` files... Does that seem right?

dcbaker: Okay, I've changed this to:
```markdown
  - `sources` (File | string)[]: A list of sources to be transpiled int .moc
    files for manual inclusion.
  - `headers` (File | string)[]: A list of headers to be transpiled int .cpp files
```

deepbluev7: Okay, to summarize the matrix discussion, source files seem to be compiled to .moc, headers to .cpp. The former you need to manually include in your sources, the latter should be added to another target to be compiled and linked.

deepbluev7: Why "int" instead of "into" btw?

dcbaker: lol, typo + copy paste fail

test cases/frameworks/4 qt/meson.build
```diff
@@ -66,7 +67,11 @@ foreach qt : ['qt4', 'qt5', 'qt6']
     endif
 
     # Test that setting a unique name with a positional argument works
-    qtmodule.preprocess(qt + 'teststuff', qresources : files(['stuff.qrc', 'stuff2.qrc']), method : get_option('method'))
```
mensinda, line 69: Since `preprocess` is still there (deprecated but still supported) we should keep testing it.

mensinda (commented):
I have only skimmed over the changes for now, but most of it looks good. Also, could you please split the `build.py` and interpreter changes out into a new PR, similar to #8834 (I know, I am also not good myself at keeping the size of my PRs in check :) )?

jpakkane:
The reason there was originally only the preprocess function is that most Qt projects (that I were dealing with at the time at least) were using those together. That is, you'd use the output of `preprocess` in one, and only one, target. If this is still the case then having the three individual functions makes for a worse UX, because people need to write method calls instead of only one. It would be nice if we could get actual usage info from existing projects. If people are using those things separately then the change might make sense. I suspect that is not the case, but that is only a guess.

test cases/frameworks/4 qt/meson.build
dcbaker: I'm pretty sure there were two calls to prepocess in that test and I only changed one of them, though I could be wrong.

dcbaker:
@deepbluev7 is using the qt module I think

deepbluev7:
I actually prefer having separate functions compared to doing everything with preprocess. Usually you have such a big list of files, that adding them all to one preprocess call becomes messy quickly. Consider the following:

```meson
moc = qt.compile_moc(headers: ['a.h', 'b.h', 'c.h', 'd.h', 'e.h', 'f.h', 'g.h', 'h.h', 'i.h', 'j.h', 'k.h', 'l.h'], 
                            extra_arguments: ['-DMAKES_MY_MOC_HEADER_COMPILE'],
                            include_directories: inc,
                            dependencies: [qt5_dep, otherlib])
res = qt.compile_resources(sources: ['resources/res.qrc'])
translations = qt.compile_translations(ts_files: ['translations/app_de.ts', 'translations/app_en.ts'])
```

Most of the overhead is just the files you are listing. With the old preprocess, you are not really saving a lot of code, but you make the dependencies and what is happening less clear:


```meson
qt_preprocessed = qt.preprocess(moc_headers: ['a.h', 'b.h', 'c.h', 'd.h', 'e.h', 'f.h', 'g.h', 'h.h', 'i.h', 'j.h', 'k.h', 'l.h'], 
                            moc_extra_arguments: ['-DMAKES_MY_MOC_HEADER_COMPILE'],
                            include_directories: inc,
                            dependencies: [qt5_dep, otherlib],
                            qresources: ['resources/res.qrc'])
translations = qt.compile_translations(ts_files: ['translations/app_de.ts', 'translations/app_en.ts'])
```

So in general I'd say the split stuff is cleaner, easier to understand and raises less questions. Maybe other devs see that differently though.

mesonbuild/build.py
```diff
@@ -1499,140 +1500,96 @@ def check_module_linking(self):
...
+    @staticmethod
+    def is_parent_path(parent: str, trial: str) -> bool:
         relpath = pathlib.PurePath(trial).relative_to(parent)
         return relpath.parts[0] != '..' # For subdirs we can only go "down".
 
-    def process_files(self, name, files, state, preserve_path_from=None, extra_args=None):
-        new = False
+    def process_files(self, files: T.Iterable[T.Union[str, File, 'CustomTarget', 'CustomTargetIndex', 'GeneratedList']],
+                      state: T.Union['Interpreter', 'ModuleState'], preserve_path_from: T.Optional[str] = None,
+                      extra_args: T.Optional[T.List[str]] = None) -> 'GeneratedList':
```
mensinda, line 1548: This is hard to read... Please use one line for each argument.

mesonbuild/build.py
```diff
@@ -1499,140 +1500,96 @@ def check_module_linking(self):
...
-        if len(args) != 1:
-            raise InvalidArguments('Generator requires exactly one positional argument: the executable')
-        exe = unholder(args[0])
-        if not isinstance(exe, (Executable, programs.ExternalProgram)):
-            raise InvalidArguments('First generator argument must be an executable.')
+    def __init__(self, exe: T.Union['Executable', programs.ExternalProgram],
+                 arguments: T.List[str], output: T.List[str], *,
+                 depfile: T.Optional[str] = None, capture: bool = False,
+                 depends: T.Optional[T.List[T.Union[BuildTarget, 'CustomTarget']]] = None,
+                 name: str = 'Generator'):
```
mensinda, line 1507: This would be a good case for using one line for each parameter to improve readability...

mesonbuild/modules/qt.py
```diff
@@ -118,21 +185,24 @@ def _detect_tools(self, state, method, required=True):
...
+        # FIXME: what error are we actually tring to check here?
         try:
             tree = ET.parse(abspath)
             root = tree.getroot()
-            result = []
+            result: T.List[str] = []
             for child in root[0]:
                 if child.tag != 'file':
-                    mlog.warning("malformed rcc file: ", os.path.join(state.subdir, rcc_file))
+                    mlog.warning("malformed rcc file: ", os.path.join(state.subdir, str(rcc_file)))
```
mensinda, line 205: This should no longer be required

mesonbuild/modules/qt.py
```diff
@@ -170,12 +240,19 @@ def parse_qrc_deps(self, state, rcc_file):
...
+        KwargInfo('required', (bool, FeatureOptionHolder), default=False),
+        KwargInfo('method', str, default='auto'),
+    )
+    def has_tools(self, state: 'ModuleState', args: T.Tuple, kwargs: 'HasToolKwArgs') -> bool:
         method = kwargs.get('method', 'auto')
-        disabled, required, feature = extract_required_kwarg(kwargs, state.subproject, default=False)
+        # We have to cast here because TypedDicts are invariant, even though
+        # ExtractRequiredKwArgs is a subset of HasToolKwArgs, type checkers
+        # will insist this is wrong
+        disabled, required, feature = extract_required_kwarg(T.cast('ExtractRequiredKwArgs', kwargs), state.subproject, default=False)
```
mensinda, line 255: Haven't tested it, but can't you fix this with something like this:

```py
class HasToolKwArgs(ExtractRequiredKwArgs):
    method: str
```

mensinda (commented):
I have mostly looked over the typing-related changes, I am assuming that the QT logic is OK.

Most of the comments are style/formating related, and are more suggestions. There are also some typing-related questions.

mesonbuild/modules/qt.py
dcbaker: I'm not sure, possibly? This is a case I might have to open an issue with upstream, because it seems like that should be valid, though the reality is I find python's `TypedDict` to fall far short of TypeScript's object annotations in general, where this would Just Work™

mensinda: OK, I have tested it and we can get rid of the `T.cast` if `HasToolKwArgs` inherits from `ExtractRequiredKwArgs`

mesonbuild/modules/qt.py
dcbaker: I checked, it is because `os.path.join` doesn't understand File like objects.

mensinda: Oh, I was assuming this was a `Path`. Makes sense.

mesonbuild/modules/qt.py
dcbaker: yay, that's nice.

mesonbuild/build.py
dcbaker: Have I ever mentioned how much I miss [attrs](https://www.attrs.org/en/stable/)?

mensinda (approved):
LGTM code-wise, however, I am not a Qt expert so I can't comment much on the Qt-specific logic.

dcbaker:
Okay, I've undeprecated `qt.process` in this version, and replaced the doc notice with a `consider using the individual functions` instead. I've also cleaned up the qt module documentation a bit in this latest version, but otherwise it's the same.

jpakkane:
Either the line numbers are off or that error message does not make any sort of sense...

jon-turney:
I'm guessing this is the cause of a corpus-test failure for aqemu https://github.com/jon-turney/meson-corpus-test/runs/2841610352 ?

```
  File "/usr/local/lib/python3.8/dist-packages/mesonbuild/mesonmain.py", line 223, in run
  File "/usr/local/lib/python3.8/dist-packages/mesonbuild/msetup.py", line 281, in run
    app.generate()
  File "/usr/local/lib/python3.8/dist-packages/mesonbuild/msetup.py", line 184, in generate
    self._generate(env)
  File "/usr/local/lib/python3.8/dist-packages/mesonbuild/msetup.py", line 223, in _generate
    intr.run()
  File "/usr/local/lib/python3.8/dist-packages/mesonbuild/interpreter/interpreter.py", line 2664, in run
    super().run()
  File "/usr/local/lib/python3.8/dist-packages/mesonbuild/interpreterbase/interpreterbase.py", line 112, in run
  File "/usr/local/lib/python3.8/dist-packages/mesonbuild/interpreterbase/interpreterbase.py", line 137, in evaluate_codeblock
    raise e
  File "/usr/local/lib/python3.8/dist-packages/mesonbuild/interpreterbase/interpreterbase.py", line 130, in evaluate_codeblock
    self.evaluate_statement(cur)
  File "/usr/local/lib/python3.8/dist-packages/mesonbuild/interpreterbase/interpreterbase.py", line 145, in evaluate_statement
    self.assignment(cur)
  File "/usr/local/lib/python3.8/dist-packages/mesonbuild/interpreterbase/interpreterbase.py", line 839, in assignment
    value = self.evaluate_statement(node.value)
  File "/usr/local/lib/python3.8/dist-packages/mesonbuild/interpreterbase/interpreterbase.py", line 147, in evaluate_statement
    return self.method_call(cur)
  File "/usr/local/lib/python3.8/dist-packages/mesonbuild/interpreterbase/interpreterbase.py", line 564, in method_call
    return obj.method_call(method_name, args, kwargs)
  File "/usr/local/lib/python3.8/dist-packages/mesonbuild/interpreter/interpreterobjects.py", line 826, in method_call
  File "/usr/local/lib/python3.8/dist-packages/mesonbuild/interpreterbase/decorators.py", line 417, in wrapper
  File "/usr/local/lib/python3.8/dist-packages/mesonbuild/modules/qt.py", line 435, in preprocess
    sources.extend(self.compile_ui(state, tuple(), ui_kwargs).return_value)
TypeError: 'GeneratedListHolder' object is not iterable
```

textshell:
This used to work:
````
some_generated = qt5.preprocess(moc_headers: some_moc_headers, include_directories: '../src')
````
It now fails with:
> ERROR: qt.preprocess keyword argument "include_directories" contained a value of type "str" but should have been "IncludeDirsHolder"

dcbaker:
@textshell I'll send a patch

Issue #8707, opened 2021-04-27
Cannot debug scan-build failure, scanbuild log is deleted


mcatanzaro:
**Describe the bug**

Epiphany's CI has started failing due to this strange scan-build issue:

```
$ flatpak build flatpak_app ninja -C _build scan-build
ninja: Entering directory `_build'
[0/1] /usr/bin/meson --internal scanbuild /home/mcatanzar...ize=4 -Dwarning_level=2 -Dwerror=true -Dwrap_mode=default
scan-build: Using '/usr/bin/clang-10' for static analysis
WARNING: Recommend using either -Dbuildtype or -Doptimization + -Ddebug. Using both is redundant since they override each other. See: https://mesonbuild.com/Builtin-options.html#build-type-options
The Meson build system
Version: 0.57.1
Source dir: /home/mcatanzaro/Projects/epiphany
Build dir: /home/mcatanzaro/Projects/epiphany/_build/meson-private/tmp8phqp91m
Build type: native build
Project name: epiphany
Project version: 40.0

../meson.build:1:0: ERROR: Compiler /usr/bin/../libexec/ccc-analyzer can not compile programs.

A full log can be found at /home/mcatanzaro/Projects/epiphany/_build/meson-private/tmp8phqp91m/meson-logs/meson-log.txt
scan-build: Removing directory '/tmp/scan-build-2021-04-27-150438-6-1' because it contains no reports.
scan-build: No bugs found.
FAILED: meson-scan-build 
[...]

Issue #8818, opened 2021-06-01
'meson scan-build' is broken


mcatanzaro:
**Describe the bug**
From #8707, `meson scan-build` is broken since https://github.com/mesonbuild/meson/commit/e81acbd6069e8c1ae8e3be7bb83ddc239009d42d. ~~This commit should be reverted.~~ Sorry, I see the regression from that commit was already fixed, and the real commit that introduced this bug is hidden by that one....

**To Reproduce**
See #8707

**Expected behavior**
See #8707

**system parameters**
See #8707

mcatanzaro:
[freedesktop-sdk tracking issue here](https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/issues/1258)

jpakkane:
Can you test #8823?

mcatanzaro:
Yup, thanks!

mcatanzaro:
I tested #8823 and found it fixed #8707, making it finally possible to debug this issue. The error is:

```
Build started at 2021-06-02T12:16:03.129061
Main binary: /usr/bin/python3
[...]

PR #8823, opened 2021-06-02
Do not delete workdir in case scan-build fails.

jpakkane:


mcatanzaro:
I can confirm this fixes #8707 (but of course not #8818).

PR #8825, opened 2021-06-02
test cases/common/103 has header symbol: set c++ standard

dcbaker:
On mac this appears to default to c++98, which in turn falls over with
boost that requires at least c++11


xclaesse:
LGTM

mensinda (approved):
LGTM, and the CI is as green as is currently possible.

PR #8826, opened 2021-06-02
Split compiler detection from Environment 

mensinda:
Another primary code-moving-around-PR to split up some large modules.

The first commit moves all the compiler detection logic into the new `compilers.detect` module. This dramatically reduces the size and complexity of `Environment`. This also puts more of the compiler-specific logic into the `compilers` module (which is where it arguably belongs to).

The second commit adds type annotations to the new module. It also fixes some issues that were caught by mypy.


mesonbuild/build.py
```diff
@@ -34,7 +34,7 @@
 )
 from .compilers import (
     Compiler, is_object, clink_langs, sort_clink, lang_suffixes,
-    is_known_suffix
+    is_known_suffix, detect_static_linker
```
dcbaker, line 37: I think this probably belongs in linkers, but since I wanted to split the dynamic and static linkers anyway I'll do that after this lands.

mensinda: Yeah, but `linkers.py` is already pretty big so the detection logic shouldn't go there. So it would probably be a good idea to make a new `mesonbuild.linkers` module instead. But then we have to maintain another `__init__.py` for just two files, so I don't really know.

mensinda: OK, I have moved *most* of the linker stuff from `compilers.detect` to `linkers.detect`. However, the `detect_static_linker` function is too tightly coupled with the compilers logic to sensibly move it somewhere else. I guess I will leave refactoring this to you then :)

dcbaker: That's fine, I seem to have a thing for doing awful refactors :)

mensinda:
Rebased. There are some slight changes to the Cython compiler detection logic.

mensinda:
Rebased again.

mesonbuild/compilers/detect.py
```diff
@@ -0,0 +1,1220 @@
...
+
+def _handle_exceptions(
+        exceptions: T.Mapping[str, T.Union[Exception, str]],
+        binaries: T.List[T.List[str]],
+        bintype: str = 'compiler'
+    ) -> T.NoReturn:
+    errmsg = f'Unknown {bintype}(s): {binaries}'
+    if exceptions:
+        errmsg += '\nThe following exception(s) were encountered:'
+        for (c, e) in exceptions.items():
```
dcbaker, line 259: style nit, you don't need the parens here. You only need them when you're doing complex unpacking, such as a nested tuple

mesonbuild/compilers/detect.py
```diff
@@ -0,0 +1,1220 @@
...
+defaults['swift'] = ['swiftc']
+defaults['vala'] = ['valac']
+defaults['cython'] = ['cython']
+defaults['static_linker'] = ['ar', 'gar']
+defaults['strip'] = ['strip']
+defaults['vs_static_linker'] = ['lib']
+defaults['clang_cl_static_linker'] = ['llvm-lib']
+defaults['cuda_static_linker'] = ['nvlink']
+defaults['gcc_static_linker'] = ['gcc-ar']
+defaults['clang_static_linker'] = ['llvm-ar']
```
dcbaker, line 196: I konw this is kinda syle-ish, buut maybe:
```python
defaults = {
  'd': ['ldc2', 'ldc', 'gdc', 'dmd'],
  'java': ['javac'],
  ...
}
if mesonlib.is_windows():
  defaults.update({
    'c': ['icl', 'cl', ...],
   ...
  })
```

dcbaker (approved):
I've got a couple of stylish things, but you can take them or leave them. Feel free to merge when you're ready.

mesonbuild/compilers/detect.py
mensinda: Fixed

mesonbuild/compilers/detect.py
mensinda: I will leave it like it currently is, I am split on which option is better with the current amount of branches...

PR #8829, opened 2021-06-03
Move Cygwin tests to a GitHub workflow

jon-turney:
Unfortunately still slow, but a minute or two faster than azure, probably because I've added caching of wheels built by pip.


jpakkane:
LGTM, what does this still need to no longer be WIP?

jon-turney:
> LGTM, what does this still need to no longer be WIP?

Just wanted to double-check a few things.  This should be good to merge now.

PR #8833, opened 2021-06-04
install: apply ldconfig at the end of the install

dabrain34:
On linux system ldconfig needs to be called to update
the ld.so.cache to be able to load libraries from
/usr/local/lib/x86_64-linux-gnu on debian based
distributions for example.

### DIFF
--- a/mesonbuild/minstall.py
+++ b/mesonbuild/minstall.py
@@ -240,12 +240,25 @@ def restore_selinux_contexts() -> None:
         # If the list of files is empty, do not try to call restorecon.
         return
 
-    proc, out, err = Popen_safe(['restorecon', '-F', '-f-', '-0'], (b'\0'.join(os.fsencode(f) for f in selinux_updates) + b'\0').decode())
+    proc, out, err = Popen_safe(['restorecon', '-F', '-f-', '-0'], ('\0'.join(f for f in selinux_updates) + '\0'))
     if proc.returncode != 0 :
         print('Failed to restore SELinux context of installed files...',
               'Standard output:', out,
               'Standard error:', err, sep='\n')
 
+def apply_ldconfig() -> None:
+    '''
+    Apply ldconfig to update the ld.so.cache.
+    '''
+    if not shutil.which('ldconfig'):
+        # If we don't have ldconfig, failure is ignored quietly.
+        return
+
+    proc, out, err = Popen_safe(['ldconfig'])
+    if proc.returncode != 0:
+        print('Failed to apply ldconfig ...',
+              'Standard output:', out,
+              'Standard error:', err, sep='\n')
 
 def get_destdir_path(destdir: str, fullprefix: str, path: str) -> str:
     if os.path.isabs(path):
@@ -345,6 +358,10 @@ class Installer:
     def restore_selinux_contexts(self) -> None:
         if not self.dry_run:
             restore_selinux_contexts()
+
+    def apply_ldconfig(self, destdir: str) -> None:
+        if not self.dry_run and not destdir:
+            apply_ldconfig()
 
     def Popen_safe(self, *args: T.Any, **kwargs: T.Any) -> T.Tuple[int, str, str]:
         if not self.dry_run:
@@ -520,6 +537,7 @@ class Installer:
                 self.install_man(d, dm, destdir, fullprefix)
                 self.install_data(d, dm, destdir, fullprefix)
                 self.restore_selinux_contexts()
+                self.apply_ldconfig(destdir)
                 self.run_install_script(d, destdir, fullprefix)
                 if not self.did_install_something:
                     self.log('Nothing to install.')
--- a/mesonbuild/minstall.py
+++ b/mesonbuild/minstall.py
@@ -229,9 +229,6 @@ def restore_selinux_contexts() -> None:
         # is ignored quietly.
         return
 
-    if os.environ.get('DESTDIR'):
-        return
-
     if not shutil.which('restorecon'):
         # If we don't have restorecon, failure is ignored quietly.
         return
@@ -355,8 +352,8 @@ class Installer:
         if not self.dry_run:
             set_mode(*args, **kwargs)
 
-    def restore_selinux_contexts(self) -> None:
-        if not self.dry_run:
+    def restore_selinux_contexts(self, destdir: str) -> None:
+        if not self.dry_run and not destdir:
             restore_selinux_contexts()
 
     def apply_ldconfig(self, destdir: str) -> None:
@@ -536,7 +533,7 @@ class Installer:
                 self.install_headers(d, dm, destdir, fullprefix)
                 self.install_man(d, dm, destdir, fullprefix)
                 self.install_data(d, dm, destdir, fullprefix)
-                self.restore_selinux_contexts()
+                self.restore_selinux_contexts(destdir)
                 self.apply_ldconfig(destdir)
                 self.run_install_script(d, destdir, fullprefix)
                 if not self.did_install_something:
--- a/mesonbuild/minstall.py
+++ b/mesonbuild/minstall.py
@@ -229,6 +229,9 @@ def restore_selinux_contexts() -> None:
         # is ignored quietly.
         return
 
+    if os.environ.get('DESTDIR'):
+        return
+
     if not shutil.which('restorecon'):
         # If we don't have restorecon, failure is ignored quietly.
         return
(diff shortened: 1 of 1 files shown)


mesonbuild/minstall.py
```diff
@@ -244,6 +244,27 @@ def restore_selinux_contexts() -> None:
...
+        # If we don't have ldconfig, failure is ignored quietly.
+        return
+
+    if os.environ.get('DESTDIR'):
+        print('Do not execute ldconfig if DESTDIR is set')
+        return
+
+    with subprocess.Popen(['ldconfig'],
+                          stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE) as proc:
+        out, err = proc.communicate(input=b'\0'.join(os.fsencode(f) for f in selinux_updates) + b'\0')
```
xclaesse, line 263: You should use `Popen_safe()` which does all kind of hacks to get output encoding right. Also you use `selinux_updates` here, seems copy/paste error.

xclaesse: While at it, I'm wondering if we should fix the selinux part too:
- Skip it completely when DESTDIR is set, instead of just ignoring error.
- Use `proc, out, err = Popen_safe(b'\0'.join(os.fsencode(f) for f in selinux_updates) + b'\0')`

dabrain34: Same PR but different commit !

mesonbuild/minstall.py
```diff
@@ -236,14 +239,31 @@ def restore_selinux_contexts() -> None:
...
+    Apply ldconfig to update the ld.so.cache.
+
+    If $DESTDIR is set, do not execute ldconfig.
+    '''
+    if not shutil.which('ldconfig'):
+        # If we don't have ldconfig, failure is ignored quietly.
+        return
+
+    if os.environ.get('DESTDIR'):
+        print('Do not execute ldconfig if DESTDIR is set')
```
eli-schwartz, line 259: It seems odd to print a warning if we don't do anything due to DESTDIR, but not if we *can't* do anything because we cannot find ldconfig, and not a status message if we *do* do something.

Elsewhere, we do both, but we do them via `self.log()` which respects the "quiet" option.

Wording could be better, see e.g.:

```
self.log(f'Skipping custom install script because DESTDIR is set {name!r}')
self.log(f'Running custom install script {name!r}')
```

apply_ldconfig() should accept the destdir variable available in do_install, and, similar to self.run_install_script, check `if destdir:` to skip running ldconfig.

eli-schwartz (changes requested):

mesonbuild/minstall.py
```diff
@@ -236,14 +239,31 @@ def restore_selinux_contexts() -> None:
...
         return
 
-    with subprocess.Popen(['restorecon', '-F', '-f-', '-0'],
-                          stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE) as proc:
-        out, err = proc.communicate(input=b'\0'.join(os.fsencode(f) for f in selinux_updates) + b'\0')
-        if proc.returncode != 0 and not os.environ.get('DESTDIR'):
-            print('Failed to restore SELinux context of installed files...',
-                  'Standard output:', out.decode(),
-                  'Standard error:', err.decode(), sep='\n')
+    proc, out, err = Popen_safe(['restorecon', '-F', '-f-', '-0'], (b'\0'.join(os.fsencode(f) for f in selinux_updates) + b'\0').decode())
```
xclaesse, line 242: Why `.decode()` here?

dabrain34: he is complaining about a Byte String otherwise

xclaesse: I think it is the type annotation of `Popen_safe()` that is wrong, I think it should be `write: T.Optional[T.Union[str, T.ByteString]]`. @dcbaker is that right?

dabrain34: yeah any advise is welcome :)

mesonbuild/minstall.py
xclaesse: > It seems odd to print a warning if we don't do anything due to DESTDIR, but not if we _can't_ do anything because we cannot find ldconfig, and not a status message if we _do_ do something.

We could be silent about DESTDIR too IMHO. If we don't have ldconfig it's probably a platform that does not use it, like Windows. I think it's fine to be silent about it.

> Elsewhere, we do both, but we do them via `self.log()` which respects the "quiet" option.

Would be nice to use `self.log()` but that function is outside the class, so would require a bit more of refactoring. It is already very unconsistent, there are tones of `print()`. IMHO this is not blocker for this PR, we are diverting already a lot from original goal.

> apply_ldconfig() should accept the destdir variable available in do_install, and, similar to self.run_install_script, check `if destdir:` to skip running ldconfig.

It's nicer to not depend on the env indeed. Note that it works fine because we have `os.environ['DESTDIR'] = self.options.destdir` but I agree better not count on it.

eli-schwartz: A better check might be if islinux or isbsd, and then allow ldconfig to raise "command not found" errors.

The reason I suggested passing destdir around is indeed specifically because we read the env var once in a consistent manner and then pass that python variable around for reuse.

dabrain34:
Related PR https://github.com/mesonbuild/meson/pull/8846

mesonbuild/minstall.py
eli-schwartz: On the self.log thing... I realized after your comment that this PR adds the ldconfig function to DirMaker which seems wrong. There's a reason restoring selinux is done in DirMaker because it's about low level file handling. But the ldconfig should he defined at the same level as running install scripts. So it should then have access to self.log()

dabrain34: is it better this way ?

I dont think it can happen to have a linux/BSD without ldconfig installed

mesonbuild/minstall.py
jpakkane: `popen_safe` should always return text because it uses `universal_newlines=True` and the legacy version decodes by hand.

mesonbuild/minstall.py
```diff
@@ -236,14 +239,25 @@ def restore_selinux_contexts() -> None:
...
 
+def apply_ldconfig() -> None:
+    '''
+    Apply ldconfig to update the ld.so.cache.
+    '''
+    if not shutil.which('ldconfig'):
+        # If we don't have ldconfig, failure is ignored quietly.
+        return
+
+    proc, out, err = Popen_safe(['ldconfig'])
```
jpakkane, line 256: Should this not first check if you are root? On my machine running `ldconfig` as a regular user always errors out due to missing permissions.

dabrain34: As the install process, if a permission is raised during the call of apply_ldconfig, the whole install process will be called using pkexec.

mensinda (approved):
LGTM. Are there any objections to merging this?

mesonbuild/minstall.py
xclaesse: @jpakkane but here it calls `decode()` on the input, not output. I think the input should remain a bytestring.

mesonbuild/minstall.py
```diff
@@ -228,6 +228,9 @@ def restore_selinux_contexts() -> None:
         # is ignored quietly.
         return
 
+    if os.environ.get('DESTDIR'):
+        return
```
xclaesse, line 232: nitpicking: I don't like getting it back from environ, it's nicer to do like you did for self.apply_ldconfig() and pass destdir in args, no?

xclaesse:
> LGTM. Are there any objections to merging this?

There was still an open review comment, but it's almost ready for merge, yes.

mesonbuild/minstall.py
dabrain34: I'd like to clarify this thread, do you want to change the arg of popen_safe to be bytestring ?

eli-schwartz (changes requested):
Repeating my overlooked comment from a resolved conversation...

> On the self.log thing... I realized after your comment that this PR adds the ldconfig function to the toplevel which seems wrong. IMO the ldconfig should he defined at the same level as running install scripts. So it should then have access to self.log()

mesonbuild/minstall.py
xclaesse: Hmm, actually forget me, the doc says input should be string too when in text mode, which our Popen_safe does:

> If streams were opened in text mode, input must be a string. Otherwise, it must be bytes.

xclaesse: But then I think `(b'\0'.join(os.fsencode(f) for f in selinux_updates) + b'\0').decode()` can be simplified to `'\0'.join(f for f in selinux_updates)`, no need to build a bytestring then decode it.

xclaesse:
> Repeating my overlooked comment from a resolved conversation...
> 
> > On the self.log thing... I realized after your comment that this PR adds the ldconfig function to the toplevel which seems wrong. IMO the ldconfig should he defined at the same level as running install scripts. So it should then have access to self.log()

Hmm, to be fair, it's like that for all other functions too. I think it's fine to not use self.log() there because it's printing error message, we want to print them even in quiet mode, no?

xclaesse:
LGTM, when CI is done. Note that a few jobs are currently failing for unrelated reasons, CI is currently in a poor state :/

PR #8834, opened 2021-06-04
A few of the basic cleanups and enhancements from the Qt series

dcbaker:
This is a few of the simpler side cleanup sort of patches from #8822, most of these are type annotations, but there is also a change to the `typed_kwargs` decorator to allow safely passing mutable containers as default values in the Kwargs, which is useful for cases were you want the default to be an empty list.


mensinda (approved):
LGTM

PR #8836, opened 2021-06-04
typing: Fully annotate dependencies.coarrays and dependencies.configtool

mensinda:


mesonbuild/dependencies/coarrays.py
```diff
@@ -21,15 +21,15 @@
...
     from ..environment import Environment, MachineChoice
 
 
 @factory_methods({DependencyMethods.PKGCONFIG, DependencyMethods.CMAKE, DependencyMethods.SYSTEM})
 def coarray_factory(env: 'Environment', for_machine: 'MachineChoice',
-                    kwargs: T.Dict[str, T.Any], methods: T.List[DependencyMethods]) -> T.List['DependencyType']:
+                    kwargs: T.Dict[str, T.Any], methods: T.List[DependencyMethods]) -> T.List[T.Callable[[], 'Dependency']]:
     fcid = detect_compiler('coarray', env, for_machine, 'fortran').get_id()
-    candidates = []  # type: T.List[DependencyType]
+    candidates: T.List[T.Callable[[], 'Dependency']] = []
```
dcbaker, line 32: I think this should be `ExternalDependency`

mesonbuild/dependencies/configtool.py
```diff
@@ -63,14 +72,12 @@ def _sanitize_version(self, version):
             return m.group(0).rstrip('.')
         return version
 
-    def find_config(self, versions: T.Optional[T.List[str]] = None, returncode: int = 0) \
-            -> T.Tuple[T.Optional[str], T.Optional[str]]:
+    def find_config(self, versions: T.List[str] = None, returncode: int = 0) \
+            -> T.Tuple[T.Optional[T.List[str]], T.Optional[str]]:
```
dcbaker, line 76: Did you look through the ConfigToolDependency subclasses to make sure that they're all going to be okay with this change?

dcbaker (approved):
I have one little nit, and one question, but In general I think this is fine, and I trust your judgement on both of the things I pointed out whichever way

dcbaker:
BTW, just so you're aware, my plan with `typed_kwargs` is to pull all of the `process_kwargs` stuff out of the `build` and `dependencies` modules and move it all into the interpreter, with the lower level code getting fully validated typed code from the get go.

mesonbuild/dependencies/coarrays.py
mensinda: It would make sense, but `factory.py` is also using `Dependency` instead of `ExternalDependency`. And changing it to `ExternalDependency` for just `coarrays.py` makes mypy unhappy :(

Since the plan of this PR series is to "just" got the dependencies fully annotated, I would say that we should refactor this after all PRs are merged.

mesonbuild/dependencies/configtool.py
mensinda: Yes. `find_config` is only called once in `ConfigToolDependency` and only overwritten once in `GnuStepDependency`.

However, I did forget to remove the `= None` for the `versions` kwarg...

PR #8837, opened 2021-06-04
typing: Fully annotate dependencies.cuda

mensinda:


mesonbuild/dependencies/base.py
```diff
@@ -364,7 +367,7 @@ def _check_version(self) -> None:
             if not self.version:
                 found_msg: mlog.TV_LoggableList = []
                 found_msg += ['Dependency', mlog.bold(self.name), 'found:']
-                found_msg += [mlog.red('NO'), 'unknown version, but need:', self.version_reqs]
+                found_msg += [mlog.red('NO'), 'unknown version, but need:', str(self.version_reqs)]
```
dcbaker, line 370: now that we have typing extensions, we could actually fix this to make it accurate. mlog doesn't really need a string, it needs something that meets this protocol:
```python
class StringProto(Protocol):
    def __str__(self) -> str: ...
```

Or we can leave it, whatever you prefer.

dcbaker (approved):
However you feel about the protocol suggestion, this is r-b me.

mesonbuild/dependencies/base.py
mensinda: Changed it, `StringProtocol` is much nicer.

PR #8838, opened 2021-06-05
add extra_path argument to dependency()

Garcia6l20:
Allows customization of `PkgConfigDependency`'s PKG_CONFIG_PATH.

### DIFF
--- a/mesonbuild/dependencies/base.py
+++ b/mesonbuild/dependencies/base.py
@@ -314,6 +314,7 @@ class ExternalDependency(Dependency, HasNativeKwarg):
         self.is_found = False
         self.language = language
         self.version_reqs = kwargs.get('version', None)
+        self.extra_path = kwargs.get('extra_path', None)
         if isinstance(self.version_reqs, str):
             self.version_reqs = [self.version_reqs]
         self.required = kwargs.get('required', True)
--- a/mesonbuild/dependencies/detect.py
+++ b/mesonbuild/dependencies/detect.py
@@ -35,7 +35,7 @@ _packages_accept_language = set()
 def get_dep_identifier(name, kwargs) -> T.Tuple:
     identifier = (name, )
     from ..interpreter import permitted_dependency_kwargs
-    assert len(permitted_dependency_kwargs) == 19, \
+    assert len(permitted_dependency_kwargs) == 20, \
            'Extra kwargs have been added to dependency(), please review if it makes sense to handle it here'
     for key, value in kwargs.items():
         # 'version' is irrelevant for caching; the caller must check version matches
--- a/mesonbuild/dependencies/pkgconfig.py
+++ b/mesonbuild/dependencies/pkgconfig.py
@@ -142,7 +142,7 @@ class PkgConfigDependency(ExternalDependency):
         else:
             env = env.copy()
 
-        PkgConfigDependency.setup_env(env, self.env, self.for_machine)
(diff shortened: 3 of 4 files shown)


Garcia6l20:
Hello !

I've been stuck for a while using meson with conan.
This allow us to do:
```meson
dependency('catch2', extra_path : meson.current_build_dir())
```
in the same spirit that we can do:
```meson
dependency('catch2', cmake_module_path : meson.current_build_dir())
```
but using conan's pkg_config generator.

mensinda:
Changes that modify the functions, etc. always also need a documentation update. In this case, a release note snippet is also required since this is new functionality. Finally new functionality also always requires a test case. See https://mesonbuild.com/Contributing.html.

Also, isn't it already possible to do something like this with the `build.pkg_config_path` option? So instead of adding `extra_path : meson.current_build_dir()` to each dependency call you should already (haven't tested it) do something like this:

```meson
project('foo', default_options: ['build.pkg_config_path=@0@'.format(meson.current_build_dir())])
```

While we appreciate new contributors, we also want to add new functionality only if it has a wide enough use case, and can't be reasonably done already with other mechanisms. Otherwise, we would end up with 10 badly maintained different methods for doing essentially the same thing.

mensinda:
Also, assuming we go forward with the `extra_path` method, I would suggest renaming it to something like `pkgconfig_extra_paths`, since it is pkg-config specific. Additionally, it should also support a list of paths (instead of only one).

Garcia6l20:
Thanks, for this quick response.

I didn't found the way to do this by myself in the documentation, so, I assumed conan should be used with cmake (witch adds extra requirements for end-users).
The default-option way works great so, I'll close this PR since there is no requirements.

Thanks again for reactivity

Garcia6l20:
Also, I'd like to show you my meson 'wrapper' (actually more an 'unwrapper') that is just removing the scripting part of meson: [nosem](https://github.com/Garcia6l20/nosem).

The idea is that (IMO) most people (including me) doesn't actually want new scripting languages, meson is in python, so, why not keeping python for build scripts ?

What do you think about this ?
And, could it be a part of meson ?

mensinda:
The main idea behind defining a new scripting language was to be independent of the programming language Meson is written in (maybe Python decides to have another breaking change, etc.). Link to the docs: https://mesonbuild.com/Use-of-Python.html

Also, this allows us to explicitly **not** to be turning complete. With nosem, nothing is stopping you from implementing Pacman in your build system, since it is just python with some special functions.

So, while I think that this is an interesting project (and I am still surprised that this works), I don't see any chance for something like this to be included in Meson itself.

Also, everything in the `mesonbuild` package is **not** a public API and subject to change. We don't break anything on purpose but some refactorings and changes are unavoidable.

Garcia6l20:
Well, it actually seems to be a problem with subprojects, I'm investingating on it and marking the PR as draft.

mensinda:
You can also pass `default_options` to subprojects: https://mesonbuild.com/Reference-manual.html#subproject. There is also the `pkg_config_path` option without the `build.` prefix you could try.

Garcia6l20:
Thats is !
Subprojects should install their dependencies to the `build_root()` instead of `current_build_dir()`, it looks ok for public dependencies, but Im wondering if it could not be a problem for private dependencies, eg.: lib1 needs zlib-1.0.0 while lib2 needs zlib-2.0.0 (internally and not exposed to the public API).

PR #8843, opened 2021-06-06
interpreter: Suggest to install wrap from wrapdb when not found

xclaesse:


mesonbuild/wrap/wraptool.py
```diff
@@ -58,10 +59,24 @@ def add_arguments(parser: 'argparse.ArgumentParser') -> None:
     p.add_argument('project_path')
     p.set_defaults(wrap_func=promote)
 
+@lru_cache(maxsize=None)
```
jpakkane, line 62: I'm not a fan of causing an http query on every configuration that encounters a missing dependency. This lookup should be done using only local data, meaning caching the release file in the file system. Alternatively it could download the file if wrap mode is not `nodownload` or something similar. Also `maxsize=None` is meaningless here since the function takes no arguments.

tp-m:
Doing this automatically without explicit opt-in might raise some privacy concerns..

xclaesse:
> Doing this automatically without explicit opt-in might raise some privacy concerns..

Making it opt-out with --wrap-mode=nodownload wouldn't be enough in your opinion?

mesonbuild/wrap/wraptool.py
xclaesse: `--wrap-mode=nodownload` should indeed skip this. You think we should not download it implicitely, but force user to run a command (e.g. `meson wrap update-db`) that store the DB locally and use that?

> Also `maxsize=None` is meaningless here since the function takes no arguments.

Good point, that's blind copy/pasted from elsewhere we used it.

eli-schwartz:
The idea behind "privacy concerns" is that the people who might get upset at this will get upset at the idea that it is opt out rather than opt in. So pointing out that they can, in fact, opt out, will not actually address the concern.

(I am not one of those people. But I've heard them before...)

tp-m:
> Making it opt-out with --wrap-mode=nodownload wouldn't be enough in your opinion?

Not sure. There's a reason even "Allow VLC to check if it's the lastest version" is opt-in.

Mind you, I'm not voicing an opinion with regard to what's right or reasonable here, just saying this this is the kind of thing that can easily land you on the front-page news, that's all.

Arguably downloading provided subproject wraps by default without any opt-in is not much different and hasn't been a problem so far, so perhaps I'm overthinking this and we needn't worry.

xclaesse:
Ok, I think I'll add a command `meson wrap update-db` that will download and store the json file in user files. Then we can use that file offline and it's user's repsonsability to update the file times to times?

xclaesse:
Closing this, superseded by https://github.com/mesonbuild/meson/pull/8941.

PR #8844, opened 2021-06-06
more work on f-strings

eli-schwartz:
More work is not the same as complete, but it is at least more work.

Most of it is stuff that the semi-automated #8475 didn't catch, though some things seem to just be new instances introduced since then.


dcbaker (approved):
LGTM

eli-schwartz:
Just rebased against master again before merging, no changes. :)

Issue #7345, opened 2020-06-17
'ninja install' attempts to gain elevated privileges


dankegel:
It is disturbing for 'ninja install' to prompt for a sudo password with polkit.

This breaks user expectations, and introduces a security risk.

This has evidently been meson's behavior for a very long time, but it's distressing, and it's surprising because meson is usually so keen on simplicity; the last thing I expect when invoking a build tool is for it to take over the screen and request a password.

(It would be equally distressing for 'sudo ninja install' to clutter the local directory with root-owned files.)

michaelbadcrumble:
What system are you using? Asking because it helps use determine how to fix this. Also would help to say what version of Meson and Ninja you’re using.

dankegel:
This is on Linux (ubuntu 20.04), and using meson from either the system or up-to-date git.

It's not a bug; it's a feature.  It just happens to be a feature that gives me the shivers.

Here's an example meson.build to reproduce with:
```
project('blah', 'c')
configure_file(
   input: 'foo.h.in',
   output: 'foo.h',
   configuration : {},
   install_dir: 'include',
   install: true)
```
Create a foo.h.in with anything in it, then do
```
$ meson . btmp
$ cd btmp
$ ninja install
```

This will trigger the prompt for a sudo password.

michaelbadcrumble:
The `sudo` command is used to temporarily give members of the sudo group privileges similar to the one root users have. When setting up Ubuntu you are asked to provide a password to proceed with installation. That is the password needed here.

Note it won't be seen on the terminal when typing as this prevents any one around from looking over your shoulder and seeing it.

That is the password that will be asked of you whenever you want to carry out any tasks that required elevated user privileges, like in this particular case.

It’s possible to disable sudo but I don’t recommend doing it.

dankegel:
That is all true, and I knew that already.

The bug I am reporting is a design bug; build utilities have traditionally
never attempted to gain elevated privileges, and it is highly surprising that meson does so.

eli-schwartz:
@michaelbadcrumble This issue was opened with the rationale "asking for sudo is a disturbing design and security violation because build tools should not ask for elevated privileges". It's very out of place to try explaining that sudo is a tool to gain elevated privileges, because that was never in doubt...

...

This was added in #3567 with the rationale:
- No more accidentally touching a file before running `sudo ninja install`, causing the outputs to become owned by root.

dankegel:
[...]

PR #3567, opened 2018-05-14
Support installation via polkit

kirbyfan64:
This basically means that, if you run `ninja install` without root access, `pkexec` will be used to gain it within the installation script. This has a couple of niceties that come along with it:

- No more forgetting to put `sudo` in front of `ninja`.
- No more accidentally touching a file before running `sudo ninja install`, causing the outputs to become owned by root.

PR #8846, opened 2021-06-07
uninstall: use polkit on PermissionError

dabrain34:
As for an install, the uninstall should be able
to use the polkit to gain privileges.


mesonbuild/scripts/uninstall.py
```diff
@@ -32,17 +40,26 @@ def do_uninstall(log: str) -> None:
             print('Deleted:', fname)
             successes += 1
         except Exception as e:
-            print(f'Could not delete {fname}: {e}.')
-            failures += 1
+            if e.__class__ == PermissionError  and shutil.which('pkexec') is not None and 'PKEXEC_UID' not in os.environ:
```
xclaesse, line 43: IMHO it's cleaner to have multiple except:
```
try:
  ...
except PermissionError:
  ...
except Exception as e:
  ...

xclaesse (commented):
Small nitpicking, but other than that looks OK.

eli-schwartz:
We already have #7345 asking for the install logic to ***stop*** using polkit/pkexec with the rationale "this is a bad idea and extremely disturbing".

Now we want to add more polkit/pkexec logic?

xclaesse:
TBF, #7345 is just 1 person asking to stop using pkexec a year ago. Surely if it was serious issue a PR would have been submitted by now. Personally I think it makes sense to use pkexec, there are other CLI that does that like systemctl, so it's not totally surprising behaviour IMHO.

xclaesse:
Also we can uninstall only via ninja ATM, so the rational from https://github.com/mesonbuild/meson/pull/3567 still apply to uninstall I think?

eli-schwartz:
It's not just one person, and no patch was submitted because it is a design issue, not a bugfix, so presumably people don't want to go to the effort of writing a patch if it's going to be rejected.

The rationale from #3567 explicitly does NOT apply. The install target depends on the "all" target and there was a concern that people would "accidentally touching a file before running sudo ninja install, causing the outputs to become owned by root".

This is NOT applicable to `ninja uninstall`, because it is a PHONY target with no dependencies and will never, ever, ever cause targets to be rebuilt as root.

rmader:
I created https://github.com/mesonbuild/meson/pull/13948 now, copying the behavior of https://github.com/mesonbuild/meson/pull/11366

dabrain34:
closing in favor of #13948

PR #8847, opened 2021-06-07
ci: Fix CI image builder

mensinda:
This also moves installing most python packages to one location.


jon-turney (approved):

jon-turney:
I think  it would be clearer for review if 'fix' and 'improve' were separate commits, but I guess that's just me...

mensinda:
Yeah, I agree, but building these containers (even locally) takes forever, so I did everything in one go...

PR #8848, opened 2021-06-07
typing: Fully annotate dependencies

mensinda:
With this PR, the dependencies module is completely annotated!


mesonbuild/dependencies/dev.py
```diff
@@ -255,19 +254,20 @@ def __fix_bogus_link_args(self, args):
...
         """
-        cpp = self.env.coredata.compilers[self.for_machine]['cpp']
 
         new_args = []
         for arg in args:
             if arg.startswith('-l') and arg.endswith('.so'):
                 new_args.append(arg.lstrip('-l'))
             elif arg.startswith('-LIBPATH:'):
+                cpp = self.env.coredata.compilers[self.for_machine]['cpp']
+                assert isinstance(cpp, CLikeCompiler)
```
dcbaker, line 264: why is this assert needed?

mesonbuild/dependencies/misc.py
```diff
@@ -43,7 +43,7 @@ def netcdf_factory(env: 'Environment', for_machine: 'MachineChoice',
     if language not in ('c', 'cpp', 'fortran'):
         raise DependencyException(f'Language {language} is not supported with NetCDF.')
 
-    candidates = []  # type: T.List['DependencyType']
+    candidates: 'TV_DepGenerators' = []
```
dcbaker, line 46: I find these `TV_*` arguments really not helpful. It's not at all obvious what is allowed here, as opposed to `T.List[T.Callable[[], 'ExternalDependency']` which makes it super obvious what this is a list of. I'd really prefer that an alias have a descriptive name like `T.List[CallbleReturnsDependency]`. These are as much for me the author of the code to understand what's going on as it is for the type checker.

dcbaker (changes requested):
I've got one concrete change, and then one comment I'm just more being a grumpy old man than anything else.

mesonbuild/dependencies/dev.py
mensinda: Because `get_linker_search_args` is only defined in `CLikeCompiler` and not in `Compiler` (which is what is stored in `coredata.compilers`).

mesonbuild/dependencies/misc.py
mensinda: I primarily changed it so that the type is defined in one place, and the `TV_` prefix is because it is used in multiple locations already. Though I agree that something like `T.List[DependencyGenerator]` is a lot nicer in this case.

mensinda: Changed it to `T.List['DependencyGenerator']`

mesonbuild/dependencies/dev.py
dcbaker: We should fix that, can you add a stub definition to `Compiler`? I've been trying to fix this whenever I find cases like this. Even if the stub implementation is `raise MesonException(f{self.id} doesn't implement this'}` that would be great.

mesonbuild/dependencies/misc.py
dcbaker: sounds good, thanks

mesonbuild/dependencies/dev.py
dcbaker: No, it's sort of a historical problem. Originally each compiler sublcass just added methods randomly, as we've added type annotations and fixed a ton of those issues I've added them to the base class. Thanks for fixing that!

dcbaker (approved):

PR #8849, opened 2021-06-07
Type checking and annotations for the `add_*_arguments functions and helpers

dcbaker:
This adds two additional keywords to the `KwargInfo` used by `typed_kwargs` (which are the last two I'd planned to add): "validator" and "converter" the first is used to set additional validation constraints (say a function accepts a string that is really an enum and only accepts 3 values). The converter can then be used to convert the Meson DSL value into whatever our implementation actually wants (like an Enum type instead of a string, or bool).

I'm trying to just pick small pieces to add type checking to so that they're easier to review than doing everything all at once.

### DIFF
--- a/mesonbuild/interpreter/interpreter.py
+++ b/mesonbuild/interpreter/interpreter.py
@@ -25,7 +25,7 @@ from ..mesonlib import FileMode, MachineChoice, OptionKey, listify, extract_as_l
 from ..programs import ExternalProgram, NonExistingExternalProgram
 from ..dependencies import Dependency, NotFoundDependency, DependencyException
 from ..depfile import DepFile
-from ..interpreterbase import InterpreterBase, typed_pos_args
+from ..interpreterbase import ContainerTypeInfo, InterpreterBase, KwargInfo, typed_kwargs, typed_pos_args
 from ..interpreterbase import noPosargs, noKwargs, stringArgs, permittedKwargs, noArgsFlattening
 from ..interpreterbase import InterpreterException, InvalidArguments, InvalidCode, SubdirDoneRequest
 from ..interpreterbase import InterpreterObject, Disabler, disablerIfNotFound
@@ -60,9 +60,7 @@ import typing as T
 import importlib
 
 if T.TYPE_CHECKING:
-    from ..compilers import Compiler
-    from ..envconfig import MachineInfo
-    from ..environment import Environment
+    from . import kwargs
 
     # Input source types passed to Targets
     SourceInputs = T.Union[mesonlib.File, GeneratedListHolder, TargetHolder,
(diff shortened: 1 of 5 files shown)


mesonbuild/interpreterbase.py
```diff
@@ -424,18 +424,30 @@ class KwargInfo(T.Generic[_T]):
...
+        string, but the implementation using an Enum. This should not do
+        validation, just converstion.
     """
 
     def __init__(self, name: str, types: T.Union[T.Type[_T], T.Tuple[T.Type[_T], ...], ContainerTypeInfo],
                  required: bool = False, listify: bool = False, default: T.Optional[_T] = None,
-                 since: T.Optional[str] = None, deprecated: T.Optional[str] = None):
+                 since: T.Optional[str] = None, deprecated: T.Optional[str] = None,
+                 validator: T.Optional[T.Callable[[_T], T.Optional[str]]] = None,
+                 convertor: T.Optional[T.Callable[[_T], T.Any]] = None):
```
mensinda, line 441: Maybe use a second `T.TypeVar` instead of `T.Any` or does this not work?

mensinda (commented):
LGTM, besides the one question.

mesonbuild/interpreter/interpreter.py
```diff
@@ -72,6 +70,36 @@
                             build.BuildTarget, build.CustomTargetIndex,
                             build.GeneratedList]
 
+    class AddProjectArgsKwArgs(TypedDict):
```
mensinda, line 73: I don't like the name of this class... I know why you want the `KwArgs` suffix but still. I don't have a better suggestion though, so...

Also, this has a different naming convention compared to `FuncGeneratorKwArgs` from your QT PR.

mesonbuild/interpreterbase.py
dcbaker: I wanted to use object there, but it doesn't fit the type annotations. I guess maybe we could do this with an overload? let met try that.

mesonbuild/interpreter/interpreter.py
dcbaker: You're right, it is. I like the KwArgs because it makes it clear that this is the Keyword Arguments for a function, but I could be convinced to use a different convention

mensinda: I am not sure if this the best idea, but we could make a new `typing` package in `interpreter` with a `kwargs` module. Then we could use `kwargs.AddProjectArgs`. This would also have the benefit of reducing the size of the `interpreter.py`, which is still too large for my taste.

mesonbuild/interpreterbase.py
dcbaker: Okay, I couldn't get object working, but it seems happy with `TYPE_nvar`, so I guess we can got with that for now.

mesonbuild/interpreter/interpreter.py
dcbaker: I like that idea, I'll do that.

mensinda (approved):
LGTM

Issue #8412, opened 2021-02-25
Meson passing '/NOSTDLIB' instead of '/NODEFAULTLIB' to MS' link.exe


ManuelAtWork:
When using Meson v0.57.1 and the MSVC 8 (2005) toolchain for a WinCE cross-build, '/NOSTDLIB' is passed to link.exe.
The behaviour produces the following warning:

    LINK : warning LNK4044: unrecognized option '/nostdlib'; ignored

Similarly, Meson is passing '-nostdinc++' to cl.exe, which is not understood by the compiler, giving the warning:

    cl : Command line warning D9002 : ignoring unknown option '-nostdinc++'

**Reproduce**
The cross file to reproduce the issue is attached to this ticket. It should work with the MSVC 2005 Express edition, as well as other, more recent versions of MSVC. ([wince.zip](https://github.com/mesonbuild/meson/files/6041160/wince.zip))

**Expected behavior**
[Instead of '/NOSTDLIB', '/NODEFAULTLIB' should be passed to the MS' link.exe](https://stackoverflow.com/a/23533692/4792058).
Instead of '-nostdinc++', '/X' should be passed to the MS' cl.exe.

**system parameters**
* build machine: desktop computer running Win10
* Meson installed from the pre-built, downloadable MSI package
* Meson being executed in VS2008 command prompt (referencing the VS2005 toolchain in the cross file)
* `meson --version` gives `0.57.1`
* `ninja --version` gives `1.10.2`

PR #8851, opened 2021-06-08
'-nostdinc'/'-nostdinc++' become '/X' and '-nostdlib' becomes '/NODEFAULTLIB' in MSVC

ManuelAtWork:
Hopefully fixed issue #8412. 

Could not test on my local machine. ... How can I build an MSI for testing on Windows?

### DIFF
--- a/mesonbuild/compilers/c.py
+++ b/mesonbuild/compilers/c.py
@@ -73,7 +73,10 @@ class CCompiler(CLikeCompiler, Compiler):
         CLikeCompiler.__init__(self, exe_wrapper)
 
     def get_no_stdinc_args(self) -> T.List[str]:
-        return ['-nostdinc']
+        if self.get_id() == 'msvc':
+            return ['/X']
+        else:
+            return ['-nostdinc']
 
     def sanity_check(self, work_dir: str, environment: 'Environment') -> None:
         code = 'int main(void) { int class=0; return class; }\n'
--- a/mesonbuild/compilers/mixins/clike.py
+++ b/mesonbuild/compilers/mixins/clike.py
@@ -159,10 +159,16 @@ class CLikeCompiler(Compiler):
         return self.get_largefile_args()
 
     def get_no_stdinc_args(self) -> T.List[str]:
-        return ['-nostdinc']
+        if self.get_id() == 'msvc':
+            return ['/X']
+        else:
+            return ['-nostdinc']
 
     def get_no_stdlib_link_args(self) -> T.List[str]:
-        return ['-nostdlib']
+        if self.get_id() == 'msvc':
+            return ['/NODEFAULTLIB']
+        else:
+            return ['-nostdlib']
 
     def get_warn_args(self, level: str) -> T.List[str]:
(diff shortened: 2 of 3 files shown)


mesonbuild/compilers/c.py
```diff
@@ -73,7 +73,10 @@ def __init__(self, exelist: T.List[str], version: str, for_machine: MachineChoic
         CLikeCompiler.__init__(self, exe_wrapper)
 
     def get_no_stdinc_args(self) -> T.List[str]:
-        return ['-nostdinc']
+        if get_id() == 'msvc':
```
dcbaker, line 76: In python you need to explicitly reference self, unlike C++ and Ruby, so this should be:
```python
if self.get_id() == 'msvc':
   return ['/X']
else:
    return ['-nostdinc']
```

However, that said, I'd prefer to implement the correct value in the subclasses rather than having an if here. Since I don't have access to a number of the compilers we support to test with, the safest way to do this would be to just add an override method into the `VisualStudioLikeCCompilerMixin` class (I've checked that ClangCL, and I can check the IntelCL compiler later), do the same for C++.

mesonbuild/compilers/mixins/clike.py
```diff
@@ -159,10 +159,16 @@ def get_always_args(self) -> T.List[str]:
...
+            return ['/X']
+        else:
+            return ['-nostdinc']
 
     def get_no_stdlib_link_args(self) -> T.List[str]:
-        return ['-nostdlib']
+        if get_id() == 'msvc':
+            return ['/NODEFAULTLIB']
+        else:
+            return ['-nostdlib']
```
dcbaker, line 171: I actually think we can delete this out of the clike abstraction, since we set it explicitly in the C and C++ layers. I'll send a patch for that.

dcbaker (changes requested):
I checked the arguments, and I agree with you those are correct, and we should be using those. I think we should handle this at a different layer though.

mesonbuild/compilers/c.py
dcbaker: To be evern more clear, we've had a long layering violation, really the `-nostdinc` and `-nostdlib` should be in the `GnuLikeCompiler` layer, but I don't know which other compilers might be affected by that change.

mesonbuild/compilers/mixins/clike.py
ManuelAtWork: Yeah, I was wondering if this is used at all. Thank's for the patch! :-)

ManuelAtWork:
> I checked the arguments, and I agree with you those are correct, and we should be using those. I think we should handle this at a different layer though.

Absolutely agree. But that would be a different ticket, as this one is about fixing functionality, not refactoring.

dcbaker:
We already have the proper abstraction, your code just isn't in it. There's `VisualStudioLike${lang}Compiler` classes already, which is the right place for the MSVC-like compiler arguments to be handled.

bonzini:
Dup of #15585.

PR #8852, opened 2021-06-08
Some run_project_tests.py improvements

mensinda:
Includes:

- full type annotation
- now runs *all* tests at once (not only all tests per section)
- optional tqdm progress bar

### DIFF
--- a/run_project_tests.py
+++ b/run_project_tests.py
@@ -70,6 +70,7 @@ if T.TYPE_CHECKING:
 
         extra_args: T.List[str]
         backend: str
+        num_workers: int
         failfast: bool
         no_unittests: bool
         only: T.List[str]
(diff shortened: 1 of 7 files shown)


run_project_tests.py
```diff
@@ -1045,61 +1074,103 @@ def __init__(self, category: str, subdir: str, skip: bool = False, stdout_mandat
...
-        return _run_tests(all_tests, log_name_base, failfast, extra_args, use_tmp)
+        return _run_tests(all_tests, log_name_base, failfast, extra_args, use_tmp, num_workers, lf)
+
+class TestStatus(Enum):
+    OK       = normal_green(' [SUCCESS] ')
+    SKIP     =       yellow(' [SKIPPED] ')
+    ERROR    =          red('  [ERROR]  ')
+    CANCELED =         cyan('[CANCELED] ')
+    RUNNING  =         blue(' [RUNNING] ')  # Should never be actually printed
+    LOG      =         bold('   [LOG]   ')  # Should never be actually printed
```
dcbaker, line 1092: this is incredibly hard to read because nothing is lined up except the = signs. There's litteraly *no* python style guid that recomends this, and this is *not* tablular data, can we just write this
```python
class TestStatus(Enum):
    OK = normal_green(' [SUCCESS] ')
    SKIP = yellow(' [SKIPPED] '`)
    ...
```

run_project_tests.py
```diff
@@ -1045,61 +1074,103 @@ def __init__(self, category: str, subdir: str, skip: bool = False, stdout_mandat
...
+        self.testdef = testdef
+        self.future = future
+        self.status = TestStatus.RUNNING
+
+    @property
+    def result(self) -> T.Optional[TestResult]:
+        return self.future.result()
+
+    def log(self) -> None:
+        without_install = "" if len(install_commands) > 0 else "(without install)"
```
dcbaker, line 1112: this should be just `'' if install_commands else "(without install)"`, using `len()` to check container emptiness is a python anti-pattern.

dcbaker (changes requested):
Mostly this looks fine to me, though I'm not an expert on how `run_project_tests` works anymore. I do have a couple of small nits.

run_project_tests.py
mensinda: Yeah, missed that when I copy-pasted code around. Is fixed now.

mensinda:
Additionally, this PR also allows now to specify the number of workers with `-j`.

Performance-wise, it decreases the runtime from 5:30 to 4:59 on my machine.

run_project_tests.py
dcbaker: For truely tabluar data this makes sense to me (things like the giant table of data in compilers/cuda.py), but really I don't care what the values of the enums are, the keys are much more intresting.

dcbaker (approved):

PR #8853, opened 2021-06-08
docs: fix issues with the cython docs

dcbaker:
As noticed by @dnicolodi

### DIFF
--- a/docs/markdown/Cython.md
+++ b/docs/markdown/Cython.md
@@ -1,7 +1,7 @@
 ---
 title: Cython
 short-description: Support for Cython in Meson
-...
+---
 
 # Cython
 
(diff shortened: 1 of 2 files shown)


docs/markdown/Cython.md
```diff
@@ -23,7 +23,7 @@ extension_module method:
 project('my project', 'cython')
 
 py = import('python')
-dep_py3 = py.dependency()
+dep_py = py.dependency()
```
dnicolodi, line 26: This is still wrong. If you want to write it like this, you need to do
```
py = import('python').find_installation()
```
in the previous line. Or something like:
```
py = import('python')
python = py.find_installation()
python_dep = python.dependency()
```
or something similar. Bottom line: the `python` module does not have a `dependency()` function.
```

dcbaker: oops, you're right. I always forget that the `python` module is basically just a holder for `find_installation()`

eli-schwartz (approved):

PR #8855, opened 2021-06-08
Use permitted_*args and add annotations for the test and benchmark functions

dcbaker:
In addition, this adds two new arguments to the `KwargInfo` class, for marking values as added or deprecated. In this case we use them with the `protocol` keyword argument.


mesonbuild/interpreter/interpreter.py
```diff
@@ -188,17 +188,24 @@ def dump_value(self, arr, list_sep, indent):
...
-    'timeout',
-    'workdir',
-}
+TEST_KWARGS: T.List[KwargInfo] = [
+    KwargInfo('args', ContainerTypeInfo(list, (str, mesonlib.File, TargetHolder)),
+              listify=True, default=[]),
+    KwargInfo('should_fail', bool, default=False),
+    KwargInfo('timeout', int, default=60),
+    KwargInfo('workdir', (str, None), default=None,
+              validator=lambda x: 'workdir must be an absolute path' if not os.path.abspath(x) else None),
```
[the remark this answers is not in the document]
dcbaker: I'm not 100% sure we can actually get None here except as the default (which doesn't go through the validator, I'm assuming you're going to write a valid default), but I've updated it anyway.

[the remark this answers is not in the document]
dcbaker: Actually, the more I thought about this the `(str, None)` is wrong, it should just be `str`, None is only possible to get as a default value, and the validator doesn't test default values, `typed_kwargs` assumes that the author set a valid default value.

mesonbuild/interpreterbase/decorators.py
```diff
@@ -290,22 +290,30 @@ class KwargInfo(T.Generic[_T]):
         different type. This is intended for cases such as the meson DSL using a
         string, but the implementation using an Enum. This should not do
         validation, just converstion.
+    :param deprecated_values: a dictionary mapping a value to the version of
+        meson it was added in.
```
jon-turney, line 294: should this comment read 'deprecate in' rather than 'added in'?

mesonbuild/mtest.py
```diff
@@ -1528,6 +1528,7 @@ def get_test_setup(self, test: T.Optional[TestSerialisation]) -> build.TestSetup
             return self.build_data.test_setups[self.options.setup]
         else:
             full_name = test.project_name + ":" + self.options.setup
+            print(self.build_data.test_setups)
```
mensinda, line 1531: Is this a leftover debug print or intentional?

mesonbuild/interpreterbase/decorators.py
dcbaker: yup, it should. fixed the in the latest vresion

mesonbuild/mtest.py
dcbaker: I think debug print, I've removed it.

mensinda (approved):

PR #8857, opened 2021-06-09
cuda module: fully buy into new modules API

eli-schwartz:
In commit 3340284805b96a4b9b62405f626020276341944c the new ModuleObject API got further updated to hide self.interpreter; at the time, the CUDA module got transferred over to the wrapper which does provide it.

But it works fine without self.interpreter, so let's just use the best base class.


xclaesse:
LGTM.

PR #8860, opened 2021-06-09
interpreter: Add missing FeatureNewKwarg for allow_fallback

xclaesse:


mensinda:
Shouldn't the `FeatureNew.single_use('"allow_fallback" keyword argument for dependency', '0.56.0', self.subproject)` check be removed at the same time?

xclaesse:
@mensinda oh good catch! Removed it, better use the decorator.

mensinda (approved):
LGTM

PR #8861, opened 2021-06-09
interpreter: Extract dependency() logic into its own helper class

xclaesse:
The dependency lookup is a lot of complex code. This refactor it all
into a single file/class outside of interpreter main class. This new
design allows adding more fallbacks candidates in the future (e.g. using
cc.find_library()) but does not yet add any extra API.


xclaesse:
I have a few cases I'm not sure what's expected:
- `dependency('foo', fallback: ['foo', 'var_does_not_exist'], required: false)`: Currently it returns not-found dependency, but my refactor changed that to raising InvalidCode() exception (abort the whole thing, even if from a subproject that is not required).
- `dependency('foo', fallback: ['foo', 'var_exist_but_is_not_dep_object'], required: false)`: Currently raise InvalidCode(), which is inconsistent with the case above IMHO. That's why I made both cases raise InvalidCode() im my refactor.
- `dependency('foo', fallback: ['foo', 'var_is_another_dep_than_overriden'], required: false)`: Currently raises DependencyException() which abort the current subproject but continue if that subproject was not required. My refactor change that case to InvalidCode() too like above.

Anyone has opinion on what's the best behaviour here?

dcbaker:
I agree, we shouldn't abort in `required : false`, if you have a fallback we shoud fallback, otherwise we should return a `NotFoundDependency`.

dcbaker:
And since that's a loosening of requirements, we can just do that without backwards compatibility concerns

mesonbuild/interpreter/interpreter.py
```diff
@@ -1583,18 +1430,22 @@ def _handle_featurenew_dependencies(self, name):
...
+        fallback = kwargs.get('fallback')
+        default_options = kwargs.get('default_options')
+        df = DependencyFallbacksHolder(self, names, allow_fallback)
+        df.set_fallback(fallback, default_options)
         not_found_message = kwargs.get('not_found_message', '')
         if not isinstance(not_found_message, str):
             raise InvalidArguments('The not_found_message must be a string.')
         try:
-            d = self.dependency_impl(name, display_name, kwargs)
+            d = df.lookup(kwargs)
```
jpakkane, line 1448: Why is `DependencyFallback` an `InterpreterObject`? It seems to be only used here to do a lookup and never actually returned itself.

xclaesse: Because I'm planning on returning that object and add public methods, but wanted to have the refactoring done first before adding new API. This refactoring is extracted from https://github.com/mesonbuild/meson/pull/8699.

xclaesse:
pfiou, finally got all tests to pass. I had to move a few tests from failing/ to warning/ because cases like `dependency('foo', fallback: ['foo', 'var_does_not_exist'])` will now success (but warn) as long as the subproject overrides `foo` dependency.

xclaesse:
@mensinda @dcbaker @jpakkane ready for review :)

mesonbuild/interpreter/dependencyfallbacks.py
```diff
@@ -0,0 +1,346 @@
...
+        self.nofallback = wrap_mode == WrapMode.nofallback
+        self.forcefallback = (force_fallback or
+                              wrap_mode == WrapMode.forcefallback or
+                              any(name in force_fallback_for for name in self.names) or
+                              self.subproject_name in force_fallback_for)
+
+        # Add an implicit subproject fallback if none has been set explicitly,
+        # unless implicit fallback is not allowed.
+        # Legacy: self.allow_fallback can be None when that kwarg is not defined
+        # in dependency('name'). In that case we want don't want to use implicit
```
mensinda, line 303: `we want don't want` --> `we don't want` probably

mesonbuild/interpreter/dependencyfallbacks.py
```diff
@@ -0,0 +1,346 @@
...
+            FeatureNew.single_use('Pcap Dependency', '0.42.0', self.subproject)
+        elif name == 'vulkan':
+            FeatureNew.single_use('Vulkan Dependency', '0.42.0', self.subproject)
+        elif name == 'libwmf':
+            FeatureNew.single_use('LibWMF Dependency', '0.44.0', self.subproject)
+        elif name == 'openmp':
+            FeatureNew.single_use('OpenMP Dependency', '0.46.0', self.subproject)
+
+    def _notfound_dependency(self) -> DependencyHolder:
+        return DependencyHolder(NotFoundDependency(self.environment), self.subproject)
```
mensinda, line 253: A function for `DependencyHolder(NotFoundDependency(self.environment), self.subproject)` is a bit overkill IMHO.

Especially, since the `DependencyHolder` part will be gone after #8885 is merged.

mensinda (commented):
I only got a few nitpicks, otherwise this LGTM.

Using `typed_posargs` and `typed_kwargs` would be nice to get rid of the `TYPE_nvar` mess, but this is not required IMHO. Besides, @dcbaker should know more here.

mesonbuild/interpreter/dependencyfallbacks.py
xclaesse: Fixed.

mesonbuild/interpreter/dependencyfallbacks.py
xclaesse: It was already there in previous code, I think it's nicer tbh. Will be easier for your refactoring, just one place to edit instead of everywhere it's called.

mensinda: Sure, it is just a very minor style choice anyway.

xclaesse:
@mensinda Thanks for the quick review!

mensinda:
@xclaesse Seems like I missed that in the review (maybe it was lost in a rebase), but please actually add `dependencyfallbacks.py` to `run_mypy.py`.

xclaesse:
@mensinda it does tones of errors, I gave up. Nothing in the interpreter is checked ATM.

mensinda:
@xclaesse Looking at them now, and most of the errors are missing annotations in `dependencyfallbacks.py` itself (for instance with `self.subproject_name = None`, mypy can't figure out automatically that this should be `T.Optional[str]` and gives an error in line 70. The `Cannot determine type of "nofallback"` error is due to `self.nofallback` not being set in `self.__init__`, etc.

90% of the remaining errors should be easily fixable with `typed_posargs` and `typed_kwargs`.

jon-turney:
There's an interesting regression on rizin in the corpus test which bisects to this commit:

https://github.com/jon-turney/meson-corpus-test/runs/2849780430?check_suite_focus=true

```
shlr/meson.build:42:0: ERROR: Fallback info must have one or two items.
```

the [line](https://github.com/rizinorg/rizin/blob/6fabb6d16afadd734cd7c443f04bbb0ab57e493c/shlr/meson.build#L42) complained of 

```
tree_sitter_dep = dependency('tree-sitter', required: get_option('use_sys_tree_sitter'), static: is_static_build, fallback: []))]
```

So it's a bit unclear to me if ` fallback: []` should ever have been accepted or not. (I guess if it's invalid, that would complicate conditionally setting a fallback using a variable, if someone ever wanted to do that...)

xclaesse:
@jon-turney Hmm, never thought about that case, there is no unit test. Wondering if it should have the same effect than `allow_fallback: false`, so preventing from using fallback from wrap [provide] section? Or just same as if fallback kwarg was not provided at all?

xclaesse:
@jon-turney https://github.com/mesonbuild/meson/pull/8897

jon-turney:
> Wondering if it should have the same effect than `allow_fallback: false`, so preventing from using fallback from wrap [provide] section? Or just same as if fallback kwarg was not provided at all?

I guess the question I would ask is how it currently behaves?

xclaesse:
> I guess the question I would ask is how it currently behaves?

From quick testing it seems to act like allow_fallback=false, that's what I implemented.

jon-turney:
> > I guess the question I would ask is how it currently behaves?
> 
> From quick testing it seems to act like allow_fallback=false, that's what I implemented.

Ok, that seems fine, then.

PR #8863, opened 2021-06-10
Split interpreterbase.py

mensinda:
And here is one more PR for splitting a large module that does multiple things into multiple, more manageable pieces :)

I left the code itself mostly untouched (with one exception being refactoring the disabler functions). This is purely a code-moving-around PR.


mesonbuild/interpreterbase/decorators.py
```diff
@@ -0,0 +1,561 @@
...
+                self.feature_check_class.single_use(
+                        name, self.feature_version, subproject, self.extra_message)
+            return f(*wrapped_args, **wrapped_kwargs)
+        return T.cast(TV_func, wrapped)
+
+class FeatureNewKwargs(FeatureCheckKwargsBase):
+    feature_check_class = FeatureNew
+
+class FeatureDeprecatedKwargs(FeatureCheckKwargsBase):
+    feature_check_class = FeatureDeprecated
```
dcbaker, line 561: I wonder if this file doesn't actually belong in the interpreter module, as I don't think they're used anywhere else?

dcbaker (approved):
I wonder if the deocrators don't actually belong in the interpreter package, but otherwise this LGTM.

mesonbuild/interpreterbase/decorators.py
mensinda: Maybe, but it will be a pain if we ever need them in `interpreterbase` in the future. I am also planning a bit of a refactoring of `InterpreterBase` itself and I might end up using some of them there...

mensinda:
I am going ahead and merging this since moving the decorators somewhere else should be fairly trivial if we ever want to do that.

PR #8864, opened 2021-06-10
Make workflow steps requiring secrets conditional

jon-turney:
Make workflow steps requiring secrets conditional on the presence of those secrets.

This stops the 'build images' workflow from running and failing on it's weekly schedule in a forked repository, which results in a mail, unless you manually disable that workflow.

(The test has to be written rather indirectly as secrets aren't available to `if`, see actions/runner#520)

### DIFF
--- a/.github/workflows/images.yml
+++ b/.github/workflows/images.yml
@@ -21,6 +21,8 @@ on:
 
 jobs:
   build:
+    env:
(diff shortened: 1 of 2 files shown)


mensinda (approved):
LGTM. The build errors will be fixed by #8847

Issue #8858, opened 2021-06-09
Adding "--wrap-mode=forcefallback"  has no effect on a reconfigure


dabrain34:
**Describe the bug**
The project has been configured **without** `--wrap-mode=forcefallback` and after calling `meson build_dir --reconfigure --wrap-mode=forcefallback`, the fallback mode is not taken into account.

**To Reproduce**
in gst-build, start with the vanilla build:
```
$ meson build_dir
```
Then try to regenerate the build with 
```
meson build_dir --reconfigure --wrap-mode=forcefallback
````
The fallback are not used by the configure.

**Expected behavior**
To use the fallback instead of the system.

**system parameters**
* Is this a [cross build](https://mesonbuild.com/Cross-compilation.html) or just a plain native build (for the same computer)?
  * no
* what operating system (e.g. MacOS Catalina, Windows 10, CentOS 8.0, Ubuntu 18.04, etc.)
  * Ubuntu
* what Python version are you using e.g.
  *  3.6.9
* what `meson --version`
  * 0.58.0
* what `ninja --version` if it's a Ninja build
  * 1.8.2

**Comment by xclaesse** [comment]
From a quick test it worked for me. Do you have more specific steps to reproduce?

**Comment by dabrain34** [comment]
I took a clean gst-build/master on docker Ubuntu 18.04, configure and build it, then I got this [issue](https://gitlab.freedesktop.org/gstreamer/gst-build/-/issues/166#note_951565) so I tried to reconfigure with `--wrap-mode=forcefallback`

I just gave it a try again with gst-build and I can reproduce with only the configure step:

Where there is much less subprojects to build with this steps:

```
$ meson build_dir
$ meson build_dir --reconfigure --wrap-mode=forcefallback
```
or

```
$ meson build_dir --wrap-mode=forcefallback
```

Same with master

**Comment by xclaesse** [comment]
Can you please attach logs in both cases so we can compare what differ?

**Comment by dabrain34** [comment]
```
$ meson build_dir --wrap-mode=forcefallback
```
[meson-log.txt](https://github.com/mesonbuild/meson/files/6630018/meson-log.txt)

```
$ meson build_dir 
$ meson build_dir --reconfigure --wrap-mode=forcefallback
```
[meson-log-reconfigure.txt](https://github.com/mesonbuild/meson/files/6630031/meson-log-reconfigure.txt)

xclaesse:
[...]

PR #8865, opened 2021-06-10
coredata: Invalidate deps cache when changing wrap_mode option

xclaesse:
Fixes: #8858

### DIFF
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -634,6 +634,15 @@ class CoreData:
 
         if key.name == 'buildtype':
             self._set_others_from_buildtype(value)
+        elif key.name in {'wrap_mode', 'force_fallback_for'}:
+            # We could have the system dependency cached for a dependency that
+            # is now forced to use subproject fallback. We probably could have
+            # more fine grained cache invalidation, but better be safe.
+            self.clear_deps_cache()
+
+    def clear_deps_cache(self):
+        self.deps.host.clear()
+        self.deps.build.clear()
 
     def get_nondefault_buildtype_args(self):
         result= []
--- a/mesonbuild/mconf.py
+++ b/mesonbuild/mconf.py
@@ -85,8 +85,7 @@ class Conf:
             raise ConfException(f'Directory {build_dir} is neither a Meson build directory nor a project source directory.')
 
     def clear_cache(self):
-        self.coredata.deps.host.clear()
-        self.coredata.deps.build.clear()
+        self.coredata.clear_deps_cache()
 
     def set_options(self, options):
         self.coredata.set_options(options)


mesonbuild/coredata.py
```diff
@@ -634,6 +634,12 @@ def set_option(self, key: OptionKey, value) -> None:
 
         if key.name == 'buildtype':
             self._set_others_from_buildtype(value)
+        elif key.name in ('wrap_mode', 'force_fallback_for'):
+            # We could have the system dependency cached for a dependency that
+            # is now forced to use subproject fallback. We probably could have
+            # more fine grained cache invalidation, but better be safe.
+            self.deps.host.clear()
+            self.deps.build.clear()
```
mensinda, line 642: Instead of clearing the cache, I would prefer if the `wrap_mode` value was stored in the key for the dependency cache in [`get_dep_identifier`](https://github.com/mesonbuild/meson/blob/master/mesonbuild/dependencies/detect.py#L42-L68).

xclaesse: I thought about that but:
- it's more work.
- it has the same effect, the whole cache is useless.
- it leaks all previous keys that are not unreachable instead of just freeing them.

xclaesse: I would actually be more confortable if we invalidate the whole cache each time any option is touched. This cache is really just a small performance improvements at the cost of so many invalidation issues we already got so far...

mensinda: My main fear is that we end up with 10 different places where the cache is invalidated at some point. However, always invalidating the cache when any option changes sounds good to me.

In this case, could you also clear all the caches just in case?

mesonbuild/coredata.py
```diff
@@ -634,6 +634,12 @@ def set_option(self, key: OptionKey, value) -> None:
 
         if key.name == 'buildtype':
             self._set_others_from_buildtype(value)
+        elif key.name in ('wrap_mode', 'force_fallback_for'):
```
dcbaker, line 637: Use a set for this `if z in {x, y}`

mesonbuild/coredata.py
dcbaker: The cache is a *huge * performance improvement for mesa. Absolutely huge. invalidating the cache because *any* option changed would be a considerable regression for us.

xclaesse: You often change options and rebuild? If so, then I think this PR is the right thing to do. We can always add more options into the list if we have use-cases.

mesonbuild/coredata.py
xclaesse: OOC, does it make any difference in python interpreter (e.g. performance), or is it just semantically better?

In any case, changed now.

mesonbuild/coredata.py
dcbaker: I might do something like `-Doptimization=debug` or suddenly decide "hmm, I need to build i965 as well as iris", `-Ddri-drivers=i965`. So yeah, I would be pretty annoyed if changing an option not related to dependencies suddenly caused the dependency cache to reset

mesonbuild/coredata.py
jpakkane: If anything it should be a performance degradation because you need to compute hashes to the two strings to build the set and a third hash for the lookup (unless the interpreter has an optimization that skips that for small sets).

dcbaker: In Python string comparison is O(n), and tuple/list membership tests are O(n), but a hash comparison is a single int comparison, and the lookup is log(n),

but, as they say, for proof:
```python
import timeit

# A pessimal list case
def a():
    return '71' in ['a', 'b', 'c', 'd', '71']

# Optimal list
def b():
    return 'a' in ['a', 'b', 'c', 'd', '71']
    
# set
def c():
    return '71' in {'a', 'b', 'c', 'd', '71'}

print('pessimal list: ', timeit.timeit(a))
print('optimal list: ', timeit.timeit(b))
print('set: ', timeit.timeit(c))
```
(timeit.timeit, for the unfamiliar, runs a function 10,000 times in a loop)

```
pessimal list:  0.08077858299657237
optimal list:  0.054498803001479246
set:  0.054652670994983055
```
So, the set lookup is *slightly* slower than the optimal case, but *much* faster than the pessimal case, and obviously that gets worse the more cases you add, at 10,000 members the list/tuple takes so long I gave up and canceld it, the set takes 0.0728 seconds. Unless your data is unhashable, you want a set for membership lookups, not a list or tuple. (There's no meaningful difference between a tuple and list for iteration speed)

xclaesse: Thanks for the info, seems python sets are well implemented. In any case, it's sementically more right to use {} here, so I agree with that change.

My 2 cents on the off-topic:
For small sets, those numbers are pretty much meaningless. I don't know how python hash tables are implemented, but it can often be slower on small sets than a list/array (it famously is in GLib's GHashTable vs GPtrArray). After all, O(log(n)) < O(n) only for big (infinity) value of n.

mesonbuild/coredata.py
xclaesse: @mensinda, @dcbaker, any objection on this PR then? I refactored the cache cleaning in a method, so at least it's done from one place.

mensinda (approved):

PR #8866, opened 2021-06-10
add example for defining curses preprocessor symbols

andy5995:


jon-turney:
"File docs\markdown\Dependencies.md contains a tab indent on line 676. Only spaces are permitted."

PR #8869, opened 2021-06-11
CI: Split out file format test as a separate GitHub workflow

jon-turney:
Split out checking of file format as a separate GitHub workflow, rather than running it as part of the project tests for every platform and toolchain combination in CI, so that this test is not effected by the changed paths constraints which are applied to the project tests.

### DIFF
--- a/.github/workflows/file_format.yml
+++ b/.github/workflows/file_format.yml
@@ -1,0 +1,13 @@
+name: File format check
+
+on: [push, pull_request]
+
+jobs:
+  format:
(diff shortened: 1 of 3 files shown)


jon-turney:
This should avoid mess-ups like #8866, where this test didn't run because that PR didn't change any code or test cases.

If we think we now have all the workflows written with appropriate path constraints, we might consider removing from Contributing.md the advice to use 'skip ci' for documentation only PRs, so this workflow always runs.

mensinda (approved):

PR #8870, opened 2021-06-11
run_mypy: Check mypy version to avoid spurious errors

xclaesse:


run_mypy.py
```diff
@@ -55,6 +57,10 @@ def check_mypy() -> None:
     except ImportError:
         print('Failed import mypy')
         sys.exit(1)
+    from mypy.version import __version__ as mypy_version
+    if not version_compare(mypy_version, '>=0.902'):
+        print('mypy >=0.902 is required, older versions causes spurious errors')
```
jon-turney, line 62: 'causes' should be singular, although perhaps 'report' would be clearer...

xclaesse: report is better indeed.

mensinda (approved):

PR #8871, opened 2021-06-12
Fix issue with generated Cython code in a subdir

rgommers:
This is a follow-up to gh-8706 (Cc @dcbaker), which contained the initial fix to `ninjabackend.py` but somehow lost it. This re-applies the fix and adds a test for it.

Without the fix, the error is:
```
  ninja: error: 'ct2.pyx', needed by 'libdir/ct2.cpython-39-x86_64-linux-gnu.so.p/ct2.pyx.c',
  missing and no known rule to make it
```


dcbaker:
I could have sworn I had that exact code in my branch... I must hae rebased it out :(

PR #8873, opened 2021-06-13
minit: do not misdetect files suffixed as "" or named "." or "c" or "s" as c#

### LINKED ISSUE #8872: default language is 'cs' on almost empty project
**Describe the bug**
Meson seems to think that a project with just a COPYING and a README.md file is a C# project

**To Reproduce**
create a directory in `/tmp/aa` with the following files only:

README.md
```
# title

Paragraph
```

COPYING
```
test
```

then do `meson init .` and the language meson uses is `'cs'`:
```
Using "aa" (name of current directory) as project name.
Using "aa" (project name) as name of executable to build.
Detected language: cs
Generated meson.build file:

project('aa', 'cs',
  version : '0.1',

executable('aa',
           '.',
           install : true)
```

**Expected behavior**
Since no language is currently used in the project use the default (`'c'`)

**system parameters**
* Is this a [cross build](https://mesonbuild.com/Cross-compilation.html) or just a plain native build (for the same computer)? N/A
* what operating system (e.g. MacOS Catalina, Windows 10, CentOS 8.0, Ubuntu 18.04, etc.) Parabola (Arch Linux based)
* what Python version are you using Python 3.9.5
* what `meson --version`   0.58.1
* what `ninja --version` if it's a Ninja build 1.10.2

eli-schwartz:
You cannot `str() in str()` and expect it to act like `str() in list()`.

Fixes regression in commit bbc2745dccc40761989a3e1efbe5a69eea0bc77e
Unbreaks #6573
Fixes #8872

### DIFF
--- a/mesonbuild/minit.py
+++ b/mesonbuild/minit.py
@@ -105,7 +105,7 @@ def autodetect_options(options: 'argparse.Namespace', sample: bool = False) -> N
             if f.suffix in ('.cc', '.cpp'):
                 options.language = 'cpp'
                 break
-            if f.suffix in '.cs':
+            if f.suffix == '.cs':
                 options.language = 'cs'
                 break
             if f.suffix == '.cu':


dcbaker (approved):
Yup, this is a good fix

Issue #4468, opened 2018-11-04
Automatically find dependencies on standard locations on BSD systems


nomis:
Meson currently includes in `/usr/local/lib` in the `get_library_dirs()` function but doesn't use this path when trying to find dependencies. It is noted that this is "probably Debian/Ubuntu specific". There's no checking of `/usr/local/include` anywhere.

For Boost on Windows it will attempt to search various paths in `C:\` (this is probably wrong because the system may not be installed on `C:\`), but on BSD systems I have to set `BOOST_ROOT`.

I think Meson should include the standard locations of packages on BSD systems when searching for dependencies:
  * FreeBSD: `/usr/local`
  * OpenBSD: `/usr/local`
  * NetBSD: `/usr/pkg`
  * DragonFlyBSD: `/usr/local`

These are not simply the default locations when manually installing dependencies (that would be an exercise for the user in specifying the correct location), they're the standard locations when using the OS-provided package manager to install packages from a predefined list. Similar to how Debian/Ubuntu systems install packages from a predefined list to `/usr/lib`.

andy5995:
Thanks @nomis, I used your patch from https://github.com/nomis/dtee/blob/f18e8a9a60cc23d0d9144da3fa78a31ddf9c781b/i18n/meson.build earlier tonight

andy5995:
Apparently related to or a duplicate of #3929

andy5995:
@nomis @emersion You may want to test [this branch](https://github.com/eli-schwartz/meson/tree/intl-dep) that @eli-schwartz is working on. I think it's supposed to take care of this.

nomis:
Please stop trying to consider this to be a duplicate of difficulty finding `libintl`. It is not.

The `boost` dependency has `/usr/local` included but not `/usr/pkg`; are we to wait until every dependency has special code to handle all operating systems?

emersion:
My projects don't need libintl. Instead, they need `/usr/local/lib` to be included to the default search paths on FreeBSD, or else the required dependencies (e.g. iniparser) cannot be found.

Also `/usr/local/include` should be included in the isystem paths, otherwise the compiler will produce warnings for headers coming from third-party dependencies.

eli-schwartz:
My suggested code for handling libintl will use the default search path of /usr/lib because it will act like cc.find_library().

It is indeed completely unrelated to (and will not fix) any /usr/local/lib problems on BSD systems.

mqudsi:
[...]

PR #8875, opened 2021-06-13
localisation doc: add example for finding libintl

andy5995:
Related to issue #4468, the included example is part of a patch I found
by @nomis

@eli-schwartz Do you  have any other suggestions for this? You mentioned you had written some libintl support code somewhere...

### DIFF
--- a/docs/markdown/Localisation.md
+++ b/docs/markdown/Localisation.md
@@ -84,3 +84,26 @@ from your build folder.
 ```console
 $ meson compile intltest-update-po
 ```
+
+### libintl issues
+
+In [GNU libc][gnu_libc_gettext], the gettext symbols are defined. On
+some systems, such as BSD or MacOS, you may get linker errors regarding
+`intl` or `_libintl_gettext`. This example should adequately work
+around that problem for most cases:
+
+```meson
+cc = meson.get_compiler('c')
+
+if host_machine.system() == 'linux' or host_machine.system() == 'gnu'
+  gettext_deps = []
+elif host_machine.system() == 'freebsd' or host_machine.system() == 'openbsd' or host_machine.system() == 'dragonfly'
+  gettext_deps = [cc.find_library('intl', has_headers: ['libintl.h'], dirs: ['/usr/local/lib'], header_args: c_args)]
+elif host_machine.system() == 'netbsd'
+  gettext_deps = [cc.find_library('intl', has_headers: ['libintl.h'], dirs: ['/usr/pkg/lib'], header_args: c_args)]
+else # darwin, cygwin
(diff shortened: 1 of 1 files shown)


eli-schwartz:
```
libintl = dependency('', required: false)
if get_option('i18n')
  if not cc.has_function('ngettext')
    libintl = cc.find_library('intl', required : false, static: get_option('buildstatic'))
    if not libintl.found()
      error('ngettext not found but NLS support requested')
    endif
  endif
  conf.set('ENABLE_NLS', 1)
endif
```

I'm finding it based on a build option, and don't bother with trying to hardcode arbitrary paths on some OSes. I don't get the point of the arbitrary paths, unless the suggestion is that meson is inherently broken on FreeBSD and *every* meson.build ever needs to hardcode those paths.

nomis:
Yes, Meson does not support BSDs properly: https://github.com/mesonbuild/meson/issues/4468

(The comments referring to this issue that I have in my build file have been removed before being copied into this PR.)

xclaesse:
For the record, GLib does that: https://gitlab.gnome.org/GNOME/glib/-/blob/main/meson.build#L2050.

xclaesse:
IMHO, instead of documenting how to do it, we should make `dependency('libintl')` just work.

nomis:
> IMHO, instead of documenting how to do it, we should make `dependency('libintl')` just work.

Part of the problem is that there is no `libintl` on Linux to find as a dependency. The library _may be required_ for gettext to work so the optional `libintl` dependency should be provided by the i18n module.

eli-schwartz:
This is handled by my WIP 'intl' dependency with custom lookup, like any custom lookup style dependency it intercepts `dependency('...')` and applies its own lookup rules which may or may not include pkg-config or cmake.

In this case, the idea is to proxy `dependency('intl')` and have it:
- first check if libc defines ngettext, and if so return a "built-in" dependency type defining no compiler or linker arguments
- then check if the `intl` library can be found via `find_library()`, if so return a "system" dependency type defining to link to `-lintl` (really /path/to/libintl.dylib or whatever)
- don't check for a `pkg-config --libs libintl` or whatever, because gettext when installing the intl standalone library on systems without a libc ngettext, does not install a pkg-config file referencing it... So trying to use pkg-config here is a dead end.


Nowhere does a custom dependency require being done via meson modules rather than dependency() and in fact it's wrong to go through `import('i18n') for this.

xclaesse:
@eli-schwartz that's what I had in mind indeed. Would have to see if we can retrofit the glib stuff into that, their handling of iconv ([which itself could have custom lookup](https://gitlab.gnome.org/GNOME/glib/-/blob/main/meson.build#L1953)) in static case should probably be included into your plan.

PR #8878, opened 2021-06-14
Fix System dependencies setting type_name instead of name

### LINKED ISSUE #8877: custom lookup dependencies have broken .type_name()
As discussed on IRC, I was trying to create a new custom lookup dependency and in the process (message() debugging in meson.build) discovered this is a giant pile of fail. Ordinarily, we set e.g. pkg-config lookups as pkg-config types etc. but mesonbuild.dependencies.* is handling this... bizarrely.

https://github.com/mesonbuild/meson/blob/15fb6a407f06eae35b9d1ab72953bfdf7d9aa5ed/mesonbuild/dependencies/base.py#L309-L313

But Dependency treats type_name as type_name. And everywhere we create an ExternalDependency, we give it the .name() of the dependency in the type_name position.

dcbaker:
This makes two changes. First it introduces a new SystemDependency intermediate class, this works much like PkgConfigDependency or ConfigToolDependency, but for system dependencies, this resolves the actual issue. Then I've added a `typeing.NewType`, which allows mypy to find cases where we're accidently writing the dependeny name to the type_name. It found a few cases of that.

Fixes #8877

### DIFF
--- a/mesonbuild/dependencies/base.py
+++ b/mesonbuild/dependencies/base.py
@@ -60,6 +60,9 @@ class DependencyMethods(Enum):
     DUB = 'dub'
 
 
+DependencyTypeName = T.NewType('DependencyTypeName', str)
+
+
 class Dependency:
 
     @classmethod
@@ -72,7 +75,7 @@ class Dependency:
             raise DependencyException("include_type may only be one of ['preserve', 'system', 'non-system']")
         return kwargs['include_type']
 
-    def __init__(self, type_name: str, kwargs: T.Dict[str, T.Any]) -> None:
+    def __init__(self, type_name: DependencyTypeName, kwargs: T.Dict[str, T.Any]) -> None:
         self.name = "null"
         self.version:  T.Optional[str] = None
         self.language: T.Optional[str] = None # None means C-like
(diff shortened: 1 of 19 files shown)


mensinda:
Seems like boost is failing. Otherwise LGTM.

eli-schwartz:
boost seems to be failing, because `test cases/failing/68 wrong boost module/test.json` is matching a line of stdout which *now* includes ", tried system" at the end.

Overall, this approach does seem good. I noticed `self.type_name = 'vulkan_sdk'` myself while grepping... it weirded me out and I feel like if we do want to list this as something other than a system dependency (for "found on the system, but not with pkg-config or config-tool or cmake") then we should have a generic name for it. Which I don't think we do need.

eli-schwartz:
`test cases/common/164 dependency factory/meson.build` should really be testing this much better than it is now.

eli-schwartz (approved):
LGTM overall, but I'd still like to actually try some dependencies of type "system" and check their .type_name().

Also is the SystemDependency really complex or distinct enough to stick it in its own file? Maybe it would be fine in base.py, idk.

dcbaker:
I don't know either (with system in it's own file), I saw that other fairly simple dependencies were in their own files, so I copied that. @mensinda, you did the splitting opinions?

I'll add a test case.

mensinda:
I am not really sure. I guess if you rename `SystemDependency` to `SystemDependencyBase` it could also easily fit into `base.py`. Since it is (at least currently) really small, I would gravitate to putting it into `base.py`

jpakkane:
No need to overpolish this one. That can be fixed later.

PR #8882, opened 2021-06-14
msubprojects: Run action on all subprojects in parallel

xclaesse:
This moves all the code into a class and call its run() method in a
thread. The class queues all logs to print them at the end to avoid
mixing output of multiple actions.

### DIFF
--- a/docs/markdown/snippets/subprojects_command_parallel.md
+++ b/docs/markdown/snippets/subprojects_command_parallel.md
@@ -1,0 +1,7 @@
+## Parallelized `meson subprojects` commands
+
+All `meson subprojects` commands are now run on each subproject in parallel by
+default. The number of processes can be controlled with `--num-processes`
+argument.
+
+This speeds up considerably IO-bound operations such as downloads and git fetch.
--- a/mesonbuild/msubprojects.py
+++ b/mesonbuild/msubprojects.py
@@ -1,380 +1,414 @@
 import os, subprocess
 import argparse
+import asyncio
+import threading
+import copy
+from concurrent.futures.thread import ThreadPoolExecutor
 from pathlib import Path
 
 from . import mlog
-from .mesonlib import quiet_git, verbose_git, GitException, Popen_safe, MesonException, windows_proof_rmtree
+from .mesonlib import quiet_git, GitException, Popen_safe, MesonException, windows_proof_rmtree
 from .wrap.wrap import PackageDefinition, Resolver, WrapException, ALL_TYPES
 from .wrap import wraptool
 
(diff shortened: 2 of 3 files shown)


xclaesse:
This speed up "meson subprojects update" in gst-build from 25s to 3s in my case, when there is no changes to fetch.

xclaesse:
The huge diff is because everything gets re-indended to be part of the class and `self.` is prepended on all method calls and to access attributes.

docs/markdown/snippets/subprojects_command_parallel.md
```diff
@@ -0,0 +1,7 @@
+## Parallelized `meson subprojects` commands
+
+All `meson subprojects` commands are now run on each subproject in parallel by
+default. The number of processes can be controled with `--num-processes`
```
bonzini, line 4: controled -> controlled.

mesonbuild/msubprojects.py
```diff
@@ -1,380 +1,414 @@
...
+            return True
         return False
-    return True
 
-def foreach(r, wrap, repo_dir, options):
-    mlog.log(f'Executing command in {repo_dir}')
-    if not os.path.isdir(repo_dir):
-        mlog.log('  -> Not downloaded yet')
+    def download(self):
+        self.log(f'Download {self.wrap.name}...')
```
bonzini, line 318: Some of these can take a long time and they will not be printed until the end of the run. So one possibility could be to change all these log messages to start with `f'{self.wrap.name}: '`. Then there's no need to introduce the log queue, only the lock is needed.

If desired, you could have all messages after the first start with `'  -> '` instead of `f'{self.wrap.name}: '` if execution is not parallel.

bonzini (changes requested):

docs/markdown/snippets/subprojects_command_parallel.md
xclaesse: Thanks, fixed.

mesonbuild/msubprojects.py
xclaesse: The output would still be a mess if logs are all interleaved, even when prefixed with the subproject name, IMHO.

I added a commit that shows a progress line: `Progress: 56 / 69 (win-nasm, glib-networking, vpx, win-flex-bison-binaries, libepoxy, tinyalsa, ca...)`

xclaesse: This is similar to running parellel tests.

jpakkane:
Featurewise this is nice. I don't have much to comment on the implementation, though. Except maybe use process executor instead of threads, as stopping processes in case of problems is easier as you can just kill them.

PR #8883, opened 2021-06-14
clang: Add support for prelinking

kennylevinsen:
This rather useful feature was previously only available for GCC. Make
it available for clang as well on some "safe" linkers, inspired from
restrictions elsewhere in the clang mixin.


kennylevinsen:
clang doesn't quite like this yet:

```
clang -v -r -o libwlroots.a.p/wlroots-prelink.o [ ... SNIP ... ] libwlroots.a.p/xwayland_xwm.c.o
clang version 12.0.0
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
Found candidate GCC installation: /usr/bin/../lib/gcc/x86_64-pc-linux-gnu/11.1.0
Found candidate GCC installation: /usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/11.1.0
Found candidate GCC installation: /usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0
Found candidate GCC installation: /usr/lib64/gcc/x86_64-pc-linux-gnu/11.1.0
Selected GCC installation: /usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/11.1.0
Candidate multilib: .;@m64
Candidate multilib: 32;@m32
Selected multilib: .;@m64
 "/usr/bin/ld" --eh-frame-hdr -m elf_x86_64 -dynamic-linker /lib64/ld-linux-x86-64.so.2 -o libwlroots.a.p/wlroots-prelink.o /usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../lib64/crt1.o /usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../lib64/crti.o /usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/11.1.0/crtbegin.o -L/usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/11.1.0 -L/usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../lib64 -L/usr/bin/../lib64 -L/lib/../lib64 -L/usr/lib/../lib64 -L/usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/11.1.0/../../.. -L/usr/bin/../lib -L/lib -L/usr/lib -r  [ ... SNIP ... ] libwlroots.a.p/xwayland_xwm.c.o -lgcc --as-needed -lgcc_s --no-as-needed -lc -lgcc --as-needed -lgcc_s --no-as-needed /usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/11.1.0/crtend.o /usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../lib64/crtn.o
/usr/bin/ld: cannot find -lgcc_s
/usr/bin/ld: cannot find -lgcc_s
clang-12: error: linker command failed with exit code 1 (use -v to see invocation)
```

kennylevinsen:
The problem is that the clang linker wrapper trips completely over itself when we try to use it for this purpose, while all the underlying linkers support the feature. While Xcode supports prelinking, it appears to work by calling the linker directly (https://xcodebuildsettings.com/#generate_master_object_file).

The current version of my branch implements a very naïve way of calling the linker directly for clang prelinking. I assume we need to integrate with some binary lookup mechanism like is used for the compiler itself, but a few hints would be nice on how to proceed here.

jpakkane:
If it is only a problem on macOS, then it is fine to detect the case and exit with an error.

kennylevinsen:
It is unfortunately not macOS-specific - my use-case is wlroots on Linux and FreeBSD, where we wish to prelink and strip local symbols. The xcode reference was just to indicate that the "official"-ish solution was direct linker instantiation.

PR #8884, opened 2021-06-14
Add annotations for the various install_* functions

dcbaker:
I've gone to much greater lengths to validate the inputs than was done originally, especially for the `install_mode` argument, this may catch some invalid cases that were being silently ignored before. I've also removed using `None | FileMode` from these, functions, and instead always return `FileMode`, as it's simpler and having a `FileMode` with no values has always been valid.


jpakkane:
⚔️ conflicts.

PR #8885, opened 2021-06-15
Refactor the complete ObjectHolder logic!

mensinda:
OK, first of all, this PR ended up being **a lot** more involved (read bigger) than anticipated. At the same time, I don't think that it would be a good idea to split this PR, since doing this refactoring half-baked can get really complicated really fast. With this current, admittedly huge, PR, there is a clear before and after state.

Still, sorry for this big PR...

----

This PR is a complete overhaul of the `ObjectHolder` / `holderify` / `unholder` mechanism:

diag

Now all objects that **can** be held by an `ObjectHolder` **must** be derived from `InterpreterHeldObject`. Additionally, `ObjectHolder` now directly inherits from `InterpreterObject`. Non-Object holders can be differentiated by inheriting from `MesonInterpreterObject`. They **must not** directly inherit from `InterpreterObject`.

This setup ensures that it can always be checked whether an object is present in its "held" or "raw" state. The dummy `MesonInterpreterObject` class ensures that non-ObjectHolders (like `MesonMain`) are not affected by this change.

----

The next big change in this PR is a clear separation where `ObjectHolder`s and the plain objects are used. Additionally, `InterpreterBase` is now the **sole** instance in charge of holderifying and unholderifying objects. It is also the **only** class that should now deal with `ObjectHolder`s. Called functions and methods will now **always** receive the "raw" objects instead of `ObjectHolder`s. Additionally, interpreter functions and methods now **must not** return `ObjectHolder`s (as in, doing so is now a hard error).

Code example before:
```py
class Interpreter:
    # stuff ...

    def func_do_stuff(self, node, args, kwargs):
        exe1 = args[0]
        dep1 = args[1]
        if isinstance(exe1, ExecutableHolder):
            exe1 = exe1.held_object
        dep1 = getattr(dep1, 'held_object', dep1)  # We really had code like this...
        assert isinstance(exe1, build.Executable)
        assert isinstance(dep1, Dependency)
        # Do some more stuff...
        lib = build.SharedLibrary(exe1, dep1)
        return SharedLibraryHolder(lib)  # Returning an ObjectHolder is now an error
```

Code example after:
```py
class Interpreter:
    def __init__(self, ...) -> None:
        self.holder_map.update({
            build.SharedLibrary: SharedLibraryHolder
        }

    def func_do_stuff(self, node, args, kwargs):
        exe1 = args[0]
        dep1 = args[1]
        assert isinstance(exe1, build.Executable)
        assert isinstance(dep1, Dependency)
        return build.SharedLibrary(exe1, dep1)  # let InterpreterBase take care of holderifying this
```


mesonbuild/build.py
```diff
@@ -17,6 +17,7 @@
 import copy
 import hashlib
 import itertools, pathlib
+from mesonbuild.interpreterbase.baseobjects import MesonInterpreterObject
```
dcbaker: I know that vscode did this automatically for you, but we usually use relative imports rather than absolute ones. This is also in the wrong section, it should be with the other meson imports not the system ones.

mesonbuild/build.py
```diff
@@ -123,13 +125,13 @@ def get_target_macos_dylib_install_name(ld) -> str:
 class InvalidArguments(MesonException):
     pass
 
-class DependencyOverride:
+class DependencyOverride(InterpreterHeldObject):
```
dcbaker, line 128: I think this is backwards, we should have a `BuildObject` base class defined in the build module, and the interpreter should import that rather tha build being aware of Interpreter. The problem with this approach is that the build module now has to be aware of the interpreter (a layering violation), and one of my big goals in the type work I'm doing in the interpreter is to fix that violation

mesonbuild/interpreterbase/decorators.py
```diff
@@ -290,10 +290,15 @@ class KwargInfo(T.Generic[_T]):
...
+<<<<<<< HEAD
     :param deprecated_values: a dictionary mapping a value to the version of
         meson it was deprecated in.
-    :param since: a dictionary mapping a value to the version of meson it was
+    :param since_values: a dictionary mapping a value to the version of meson it was
         added in.
+=======
+    :param not_set_warning: A warning messsage that is logged if the kwarg is not
+        set by the user.
+>>>>>>> 3412c3776 (typing: Add not_set_warning to typed_kwargs)
```
dcbaker: rebase conflicts?

mesonbuild/build.py
```diff
@@ -39,7 +38,10 @@
     is_known_suffix
 )
 from .linkers import StaticLinker
-from .interpreterbase import FeatureNew
+from .interpreterbase import FeatureNew, TYPE_nkwargs, TYPE_nvar
+import mesonbuild
+
+from mesonbuild import mesonlib
```
dcbaker: this looks odd too

mesonbuild/build.py
```diff
@@ -1500,7 +1502,7 @@ def check_module_linking(self):
                 return
 
 class Generator(InterpreterHeldObject):
-    def __init__(self, args, kwargs):
+    def __init__(self, args: T.List[TYPE_nvar], kwargs: TYPE_nkwargs) -> None:
```
dcbaker, line 1496: the Generator and Generated List stuff is going to conflict massively when the qt module rework lands.

run_project_tests.py
```diff
@@ -1280,6 +1280,15 @@ def tqdm_print(*args: mlog.TV_Loggable, sep: str = ' ') -> None:
             safe_print(bold('During:'), result.step.name)
             safe_print(bold('Reason:'), result.msg)
             failing_tests += 1
+            # Append a visual seperator for the different test cases
+            cols = shutil.get_terminal_size((100, 20)).columns
+            name_str = ' '.join([str(x) for x in f.testdef.display_name()])
+            name_len = len(re.sub(r'\x1B[^m]+m', '', name_str))  # Do not count escape sequences
+            left_w = (cols / 2) - (name_len / 2) - 1
```
dcbaker, line 1290: instead of doing the int case below you probably want to use floor division (`//`) here

dcbaker (changes requested):
I like the ideas, and in general I'm in favor for this, but I do think that `InterpreterHeldObject` is looking at the problem backwards. The Build layer should be unaware of details of the Interpreter, just like the various `*.Holder` classes are supposed to abstract the Build layer in the interpreter. I think it would be better to have a `BuildObject` base class in the `build` module, and have the interpreter holders act on that instead of the other way around.

There's also a bunch of conflicts that are going to pop up from my qt rework, which fixes most of the Generator and GeneratedList cludge that we have right now.

jpakkane:
What dcbaker said. I always get confused which way around the ownership/inheritance arrows go in UML diagrams but `build` objects should not even know that there is such a thing as an interpreter. That is what the `Holder` in the variable names originally mean. They are "interpreter things" that "hold" "build-level things".

mesonbuild/build.py
mensinda: Good catch, I missed that. I probably should turn the auto-import off since it never produces good results.

mensinda:
The main reason why I want something like `InterpreterHeldObject` is to tell mypy that only certain objects (generally all Meson objects that are not interpreter related) can be held by an `ObjectHolder`. That's why `InterpreterHeldObject` is also in `mesonlib` and not any interpreter package. Would be renaming it to `MesonBaseObject` or `MesonBase` be acceptable (it is currently an empty class anyway in `mesonlib` anyway)?

> There's also a bunch of conflicts that are going to pop up from my qt rework, which fixes most of the Generator and GeneratedList cludge that we have right now.

Yeah, I know... But with such a big PR, conflicts will be unavoidable.

mesonbuild/build.py
mensinda: vscode auto-import probably. Anyway, it is gone now.

mesonbuild/build.py
mensinda: `InterpreterHeldObject` is actually defined in `mesonlib`, so it's technically only the name. See my comment below.

mesonbuild/build.py
mensinda: Yeah, but resolving this should be relatively easy (compared to actually creating this PR and making all the tests run again).

I am well aware of the size of this PR and what this means for the review time :)

mesonbuild/interpreterbase/decorators.py
mensinda: This is gone in the final diff... Is this from a specific commit?

run_project_tests.py
mensinda: Thanks. Still learning new things in Python :)

mensinda:
Anyway, looks like I have some bad commits in my history. I will clean this up once the Qt module changes have landed.

jpakkane:
> The main reason why I want something like InterpreterHeldObject is to tell mypy that only certain objects (generally all Meson objects that are not interpreter related) can be held by an ObjectHolder

On the other hand this is too loose. The real limitation is that an ExecutableHolder can only hold an Executable. Dunno if that is doable with Mypy, though.

Alternatively can't you specify that type to be an union of Executable, SharedLibrary etc?

dcbaker:
@jpakkane yeah, you can do that with mypy. We actually already do that, you use a Generic with specializations:
```python
T = typing.TypeVar('T')

class Holder(typing.Generic[T]):
   pass

class ExecutableHolder(Holder[build.Executable]):
  pass
```

But you can constrain `TypeVar` as well, which might be another way to handle this "what can a Holder hold?" problem
```python
Holdable = typing.TypeVar('Holdable', BuildObject, str)

class Holder(typing.Generic[T]):
   pass

class ExecutableHolder(Holder[build.Executable]):  # this is fine
  pass

class BytesHolder(Holder[bytes]):  # mypy error
  pass
```

mesonbuild/interpreterbase/decorators.py
mensinda: I will fix the commits once your Qt rewrite lands.

mensinda:
The other problem is that `ObjectHolder` is not limited to `build.Target` objects. For instance, `mesonlib.File` can now also be held by an `ObjectHolder`. With this change, basically all objects (except `str`, `int`, ...) returned by the interpreter need a corresponding `ObjectHolder`.

The current constraint of `ObjectHolder` (see [`interpreterbase.baseobjects`](https://github.com/mensinda/meson/blob/builtinRefactor/mesonbuild/interpreterbase/baseobjects.py#L69-L80)) is `T.TypeVar('InterpreterObjectTypeVar', bound=InterpreterHeldObject)`. `InterpreterHeldObject` is defined in [`mesonlib.universal`](https://github.com/mensinda/meson/blob/builtinRefactor/mesonbuild/mesonlib/universal.py#L270). I don't see which other existing class we could use instead to cover all possible held objects...

dcbaker:
You wouldn't use `bound=`, you'd have to do something like `TypeVar(T, str, int, list, dict, build.Target, ..., covariant=True)` (I think, I always get covariance and contravariance confused)

mensinda:
@dcbaker So, we should list all possible base classes in the `T.TypeVar` directly instead of introducing a new one? I can test if this works.

dcbaker:
Yes, that should work. Though having a shared base class would make that easier, and I like the idea of all `build` objects having a shared base. There's another option as well, though I haven't tested it. the `abc.ABCMeta` class can have "virtual subclasses" so you could (potentially do something like):
```python
class Holdable(metaclass=abc.ABCMeta):
     pass

@Holdable.register
class Target
    pass
```

jpakkane:
> Though having a shared base class would make that easier, and I like the idea of all build objects having a shared base.

I'm not a fan of overuse of inheritance, which this thing is approaching. A union type better describes the property of "this can be any of these unrelated types A, B and C". Sadly Python's syntax for that is not the greatest because it is basically a duck typed language.

mensinda:
Well, the shared base class is what I was trying to achieve with `InterpreterHeldObject` (though I admit that the name is misleading and should just have been something like `HoldableObject`).

From the current options, I would say I like the `abc.ABCMeta` approach the most (assuming that it works). Honestly, I am not a fan of adding all the types to the `TypeVar`, even if it is just base types. Just search for `InterpreterHeldObject)` in vscode and you will get around 30 hits (not including `isinstance` checks)... Maintaining this centrally in a `TypeVar` will be a pain. So I would prefer a decentralized solution (where the best one looks to be `abc.ABCMeta`).

dcbaker:
I'm fine with a base class, but that should be part of build and not part of the mesonlib or the interpreter, at least, that's my take. Maybe others have other opinions.

mensinda:
Also, I forgot to mention that this PR turns off the type annotations for the `ast` module, since what the `AstInterpreter` does is fundamentally incompatible with `InterpreterBase`...

dcbaker:
Would it help if we had holders for everything, btw? I've really wanted to add a ListHolder, StringHolder, NumberHolder, BoolHolder, and DictHolder for a long time now...

mensinda:
This was also my plan add the beginning of the refactoring, but for this we need the fundamentals from this PR.

This way we can also extract the custom logic for these basic types from InterpreterBase.


Also on where to put the base class: It can basically only be mesonlib because mesonlib.File is also a held object. And mesonlib cant depend on build.
-- 

dcbaker:
Honest question, is there any reason `File` couldn't move to `build`?

mensinda:
OK, the `abc.ABCMeta` approach won't work because of https://github.com/python/mypy/issues/2922 (although actually running the code still works),

Also, maybe I am doing something wrong, but using `TypeVar('T', build.Target, mesonlib.File, ..., covariant=True)` does also not work (also tried `contravariant=True`). Thus, I am going forward with a new abstract base class in `build.py`, unless we can come up with a better alternative.

jpakkane:
It's not really nice that one needs to add code and inheritance levels just to work around the limitations of some tool like this. :(

mensinda:
Renamed `InterpreterHeldObject` --> `HoldableObject`.

Sorry @dcbaker, moving `File` and/or `HoldableObject` to `build.py` is just impossible. Here is just one problem: `build.py` depends on `dependencies`, however, `dependencies` would now also depend on `build.py` for `HoldableObject`. Same problem with `programs.py`.

I am going to leave `HoldableObject` in `mesonlib` for now, however, we could make a new module for it if you really don't want to have it in `mesonlib`.

mensinda:
@jpakkane we also need a way in `_holderify` and `_unholder` to check what objects can/must be held at *runtime*. So, purely typing only changes wouldn't cut it anyway...

mensinda:
Rebased. I did my best to split the commits, but it is far from perfect. At least there shouldn't be any rebase artifacts left.

mensinda:
Rebased for #8861 (ping @xclaesse just in case).

mesonbuild/interpreterbase/decorators.py
```diff
@@ -67,13 +68,20 @@ def noArgsFlattening(f: TV_func) -> TV_func:
     setattr(f, 'no-args-flattening', True)  # noqa: B010
     return f
 
+def unholderReturn(f: TV_func) -> T.Callable[..., TYPE_var]:
```
dcbaker, line 71: `unolder_return`

dcbaker (changes requested):
I have one small nit about functions names, but otherwise this looks good to me now. @jpakkane?

mesonbuild/interpreterbase/decorators.py
mensinda: Done

dcbaker (approved):
This looks good enough to me at this point to land (I'm going to look at moving File and FileMode into build, but that can be follow on work. I think the unreachable code is real (and both mypy and pylint can find reaching code that should be unreachable)

dcbaker:
@jpakkane, I'd like to make a decision to land this or send it back for more reworks, a lot of the work I've been doing will conflict with this.

jpakkane:
⚔️ conflicts. But apart from that I don't really have much to say (and did not go through it in detail as there is a lot of stuff which I'm not really an expert on). If people are happy with this then let's go for it.

dcbaker:
@mensinda I guess then let's go ahead and rebase and merge this.

mensinda:
Yeah, let's hope that the CI stays green.

PR #8888, opened 2021-06-16
cleanup dependency code and introduce intl dependency

### LINKED ISSUE #3929: New dependency type: libintl
The functions provided by libintl can be either a compiler built-in (on Linux and macOS) or provided by an external library (on Windows). GLib does this:

```meson
if cc.has_function('ngettext')
  libintl = []
else
  libintl = cc.find_library('intl', required : false)
  if not libintl.found()
    libintl = subproject('proxy-libintl').get_variable('intl_dep')
  endif
endif
```

This simplifies to:

```meson
if cc.has_function('ngettext')
  libintl = []
else
  libintl = cc.find_library('intl')
endif
```

There is no way to specify a 'found dependency with no arguments' in build files, so this complicates the logic quite a bit since you can't call `libintl.found()` without first checking the type. Having a special dependency type would solve this.

There are probably more edge-cases ([glib-gettext.m4](https://gitlab.gnome.org/GNOME/glib/blob/4e1488eebb2129c6e39162b56275d311f24216ab/m4macros/glib-gettext.m4)), and we can slowly grow those in Meson as needed.

eli-schwartz:
The intl dependency needs better testing, probably, because as noted in #3929

> There are probably more edge-cases (glib-gettext.m4), and we can slowly grow those in Meson as needed.

The initial implementation here should make for a good start, covers the libc builtin case, and probably covers at least the simple case of external libintl.dylib (it links to a simple brew install gettext).

Fixes #3929, obsoletes #8875

/cc @andy5995 
(whose project repo did some macOS sanity checks on this)


docs/markdown/snippets/intl-dependency.md
```diff
@@ -0,0 +1,35 @@
...
+if cc.has_function('ngettext')
+  intl_found = true
+else
+  intl_dep = cc.find_library('intl', required: false)
+  intl_found = intl_dep.found()
+endif
+
+if intl_found
+  # build options that need gettext
+  # conf.set('ENABLE_NLS', 1)
```
jon-turney, line 22: Why should this line be commented out?

jon-turney (approved):
Also needs some documentation change in Dependencies.md ?

docs/markdown/snippets/intl-dependency.md
eli-schwartz: Because it made perfect sense at 1AM!

... You're right. That looks odd.

mesonbuild/dependencies/base.py
```diff
@@ -42,6 +42,8 @@ class DependencyMethods(Enum):
     AUTO = 'auto'
     PKGCONFIG = 'pkg-config'
     CMAKE = 'cmake'
+    # The dependency is provided by the standard library and does not need to be linked
+    BUILTIN = 'builtin'
```
jon-turney, line 46: Perhaps this distinction needs to be memorialized in the discussion of `method:` in Dependencies.md?

docs/markdown/snippets/intl-dependency.md
eli-schwartz: Fixed

mesonbuild/dependencies/base.py
eli-schwartz: Fixed, I think.

docs/markdown/Dependencies.md
```diff
@@ -282,6 +282,15 @@ in the build system DSL or with a script, likely calling
...
 
+## Builtin
+
+Some dependencies provide no valid methods for discovery on some systems,
+because they are provided internally by the language. One example of this is
+intl, which is built into GNU or musl libc but otherwise comes as a `system`
+dependency.
+
+In these cases meson provides convenience wrappers for the `system` dependency,
+but first checks if the functionality is usable by default.
```
jon-turney, line 293: NIce.

This, and the previous section 'system' are clearly misplaced here (in a list of dependency names which get special handling), and should be up in the list of detection methods.

jon-turney (approved):

docs/markdown/Dependencies.md
eli-schwartz: And config-tool is half one and half the other :/

PR #8894, opened 2021-06-17
coverage: Enable coverage reports

mensinda:
Coverage reports seem to be broken/not enabled/forgotten about for a long time now. This PR aims to fix that.


mensinda:
This took a bit longer than expected, but everything is now being uploaded. Sorry for the CI spam.

PR #8898, opened 2021-06-17
interpreter: Add checked kwarg to compiler.get_supported_arguments

e820:
To avoid manual compiler support checks `add_project_arguments` and
`add_global_arguments` receive a new keyword argument to perform
them automatically.

The current compiler interface does not allow for hard requirements which
means users will likely end up with a construct something like the following:
```meson
required_args = [..]
foreach arg : required_args
  if not cc.has_argument(arg)
    error(f'C compiler does not support "@arg@"')
  endif
endforeach
add_project_arguments(required_args, language : ..)
```
This patch simply allows projects to avoid repetition here:
```meson
add_project_arguments(.., language : .., required : true)
```
The goal is to supplement the `get_supported(_link)_arguments` interface
to express both filters and hard requirements.

### DIFF
--- a/mesonbuild/interpreter/compiler.py
+++ b/mesonbuild/interpreter/compiler.py
@@ -1,4 +1,5 @@
 import functools
+from mesonbuild.interpreterbase.decorators import typed_kwargs, KwargInfo
 
 from .interpreterobjects import (extract_required_kwarg, extract_search_dirs)
 
@@ -684,12 +685,24 @@ class CompilerHolder(ObjectHolder['Compiler']):
         return result
 
     @FeatureNew('compiler.get_supported_arguments', '0.43.0')
-    @permittedKwargs({})
-    def get_supported_arguments_method(self, args, kwargs):
+    @FeatureNewKwargs('compiler.get_supported_arguments', '0.59.0', ['checked'])
+    @typed_kwargs('compiler.get_supported_arguments', KwargInfo('checked', str, default='off'))
+    def get_supported_arguments_method(self, args: T.Sequence[str], kwargs: T.Dict[str, T.Any]):
         args = mesonlib.stringlistify(args)
         supported_args = []
+        checked = kwargs.pop('checked')
+
+        if checked not in ['warn', 'require', 'off']:
+            raise mesonlib.MesonException('"checked" kwarg must be one of "warn", "require" or "off"')
+
         for arg in args:
-            if self.has_argument_method(arg, kwargs):
+            if not self.has_argument_method(arg, kwargs):
+                msg = f'Compiler for {self.compiler.get_display_language()} does not support "{arg}"'
+                if checked == 'warn':
+                    mlog.warning(msg)
+                elif checked == 'require':
+                    raise mesonlib.MesonException(msg)
+            else:
                 supported_args.append(arg)
         return supported_args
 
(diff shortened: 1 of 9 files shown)


mensinda:
Would there be a theoretical use case for `add_global_arguments` to still check whether the arguments are supported, but not strictly require them (so check with `require: false`)? In that case, there should be a separate toggle for enabling the argument checks. Just having `require` be responsible for essentially two things (enable checks *and* explicitly say that the flags are required) could easily lead to problems further down the road.

dcbaker:
This doesn't belong in `add_*_argument`, what if you want to use that somewhere else, like in an `exectuable`? I would be better to add a new interface to the compiler for this. we already have `compiler.get_supported_arguments()` which, provides the `required : false` behavior, i think that's the right place to put the `required` keyword argument.

e820:
> This doesn't belong in `add_*_argument`, IMO, what if you want that behavior somewhere else, like in an `exectuable`? We already have `compiler.get_supported_arguments()` which, provides the `required : false` behavior, i think that's the right place to put the `required` keyword argument.
> 
> for example, I do this a lot:
> 
> ```meson
> add_project_arguments(
>   cc.get_supported_arguments(
>     'list',
>     'of',
>     'optional',
>     'arguments
>   ),
>   language : 'c',
> )
> ```
> 
> simply adding `required : true` there seems fine (or adding a new method, either way)

I'm happy to implement it this way, this was the original consideration but I found it to be fairly tedious overall.

mesonbuild/interpreter/compiler.py
```diff
@@ -684,12 +685,24 @@ def has_multi_arguments_method(self, args: T.Sequence[str], kwargs: dict):
...
-    def get_supported_arguments_method(self, args, kwargs):
+    @FeatureNewKwargs('compiler.get_supported_arguments', '0.59.0', ['checked'])
+    @typed_kwargs('compiler.get_supported_arguments', KwargInfo('checked', str, default='off'))
+    def get_supported_arguments_method(self, args: T.Sequence[str], kwargs: T.Dict[str, T.Any]):
         args = mesonlib.stringlistify(args)
         supported_args = []
+        checked = kwargs.pop('checked')
+
+        if checked not in ['warn', 'require', 'off']:
+            raise mesonlib.MesonException('"checked" kwarg must be one of "warn", "require" or "off"')
```
dcbaker, line 696: `typed_kwargs` actually has you covered here:
```python
@typed_kwargs(
  'compiler.get_supported_args')
  KwargInro('checked', str, default='off', since='0.59.0',
             validator=lambda x: 'must be one of "warn", "require", or "off"' if x not in {'warn', 'require', 'off'} else None),
)
```
Thne you don't need `FeatureNewKwargs` or this check here, `typed_kwargs` will do that for you.

mesonbuild/interpreter/compiler.py
```diff
@@ -1,4 +1,5 @@
 import functools
+from mesonbuild.interpreterbase.decorators import typed_kwargs, KwargInfo
```
dcbaker, line 2: we usually use relative imports, so `from ..interpreterbase.decoartors`

dcbaker (changes requested):
I've got a couple of small change requests, but otherwise this looks good to me. Thanks for moving this to the compiler instead of doing it in `add_*_arguments`.

mesonbuild/interpreter/compiler.py
e820: Yeah my language server snuck that in via auto-import, will adjust.

mesonbuild/interpreter/compiler.py
e820: Alright neat, I was not aware.

dcbaker (approved):
Thanks!

PR #8899, opened 2021-06-18
build: fix object path for vs backend

DFOVIT:
Using objects in library with the vs backend will generate a project with the wrong path to the object.
This merge request adds more tests to this specific test case and fixes the issue with vs backend.
Warning: the test case will leave object files in the source directory.

### DIFF
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -412,7 +412,14 @@ class Backend:
                                  self.build_to_src, target.get_subdir(), obj)
                 obj_list.append(o)
             elif isinstance(obj, mesonlib.File):
-                obj_list.append(obj.rel_to_builddir(self.build_to_src))
+                if obj.is_built:
+                    o = os.path.join(proj_dir_to_build_root,
+                                     obj.rel_to_builddir(self.build_to_src))
+                    obj_list.append(o)
+                else:
+                    o = os.path.join(proj_dir_to_build_root,
+                                     self.build_to_src)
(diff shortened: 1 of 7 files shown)


jpakkane:
> Warning: the test case will leave object files in the source directory.

Well that's not good. Why is that happening?

DFOVIT:
The current test case uses a configure_file to build a 'prebuilt' object in current_build_dir(). I think the intended usage of 'objects:' is to link with files that are stored in a 'source' directory. So i used configure_file to build an object in source to make sure that the path to objects (build dir to source dir) is correctly computed. The test case should increase coverage of the 'objects' option by providing an object as a str, a built=False File, and built=True File.

DFOVIT:
I withdraw the contentious tests. The warning does not apply anymore.

jpakkane:
There is already a test for exactly this [in the unit test suite](https://github.com/mesonbuild/meson/tree/master/test%20cases/unit/15%20prebuilt%20object). Why is it not triggering the bug? Is it because it is generated in the root dir instead of a subdir? If yes, then would moving the logic in that test to a subdir be enough to test this properly?

DFOVIT:
I modified the tests to output objects in the build dir. This is not matching the use case of prebuilt objects but it still triggers the bug for the vs backend so i guess it's ok. When in a subdir 'objdir', the vcxproj is generated in build/objdir but the object path is relative to the root builddir ie. objdir/source4.obj. VS obviously fails to find it, while for ninja it is correct since ninja.build resides in the root builddir.

jpakkane:
Great, thanks.

PR #8900, opened 2021-06-18
extract_objects: fixes, tests and documentation for using the result in a custom_target

bonzini:
QEMU would like to use the result of `extract_objects` in a custom target; examples are using `objcopy`, or using the object files as the key to look up command line arguments in `compile_commands.json`.  This is slightly peculiar and not covered by the test suite, but it mostly works; in order to avoid regressions, add a test case and document it.

One case that fails is when the custom_target inputs include header files, either by way of direct mention (which is common when using sourcesets) or simply because `extract_all_objects` is being used. In both cases ignore the headers and only use the source files, as done already in the case of unity builds.

### DIFF
--- a/docs/markdown/Reference-manual.md
+++ b/docs/markdown/Reference-manual.md
@@ -2552,7 +2552,9 @@ module](#shared_module).
   object files generated for those source files. This is typically used
   to take single object files and link them to unit tests or to compile
   some source files with custom flags. To use the object file(s)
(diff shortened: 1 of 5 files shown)


bonzini:
Would be nice to have in the next 0.58.x too...

jpakkane:
I was sure I had already merged this. Apparently not.

PR #8905, opened 2021-06-19
fix: Fix set_variable not holderifying (fixes #8904)

### LINKED ISSUE #8904: ERROR: Argument gobject/glib-mkenums of type File is not held by an ObjectHolder.
**Describe the bug**
master branch fails to build glib 2,64.1
```
gobject/meson.build:99:0: ERROR: Argument gobject/glib-mkenums of type File is not held by an ObjectHolder.

```

**To Reproduce**
1. get the source code of glib 2.64.1 (or switch to git tag 2.64.1)
2. run `meson build` using latest meson from master branch
3. above error

**Expected behavior**
no error

**system parameters**
* Is this a [cross build](https://mesonbuild.com/Cross-compilation.html) or just a plain native build (for the same computer)? native
* what operating system (e.g. MacOS Catalina, Windows 10, CentOS 8.0, Ubuntu 18.04, etc.) arch linux (parabola)
* what Python version are you using e.g. 3.8.0: 3.9.5
* what `meson --version` latest master branch
* what `ninja --version` if it's a Ninja build N/A

mensinda:


jpakkane:
@xclaesse can you confirm if this fixes the issue for you?

mensinda:
@jpakkane: All three PRs (so also #8912) is required to build gst-build with `-Ddefault_library=both`.

jpakkane:
Both with all of those, it has been confirmed to work?

PR #8908, opened 2021-06-20
fix: Ensure that build targets have all methods from ExternalProgram

mensinda:
As a side-effect from #8885 `find_program()` returns now `Executable`
objects when `meson.override_find_program` is called with an
executable target. To resolve this conflict the missing methods
from `ExternalProgram` are added to `BuildTarget`.

### DIFF
--- a/docs/markdown/snippets/build-target-found.md
+++ b/docs/markdown/snippets/build-target-found.md
@@ -1,0 +1,16 @@
+## New `build target` methods
+
+The [`build target` object](Reference-manual.md#build-target-object) now supports
+the following two functions, to ensure feature compatebility with
+[`external program` objects](Reference-manual.html#external-program-object):
+
+- `found()`: Always returns `true`. This function is meant
+  to make executables objects feature compatible with
+  `external program` objects. This simplifies
+  use-cases where an executable is used instead of an external program.
+
+- `path()`: **(deprecated)** does the exact same as `full_path()`.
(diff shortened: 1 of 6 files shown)


docs/markdown/snippets/build-target-found.md
```diff
@@ -0,0 +1,16 @@
...
+
+- `found()`: Always returns `true`. This function is meant
+  to make executables objects feature compatible with
+  `external program` objects. This simplifies
+  use-cases where an executable is used instead of an external program.
+
+- `path()`: **(deprecated)** does the exact same as `full_path()`.
+  **NOTE:** This function is solely kept for compatebility
+  with `external program` objects. It will be
+  removed once the, also deprecated, corresponding `path()` function is
```
jpakkane, line 15: So `path` will be removed when  `path` is removed? That seems a bit weird. What does this actually refer to?

mensinda: It refers to the already deprecated `path()` method in `external program` object. The idea here is that the `path` method in `build target` objects is only required because `external program` objects also have a `path()` method.

A new user should never actually use this new `path` method intentionally.

jpakkane: Yes, but the text should say "corresponding `path()` function in `external_program` is removed".

mensinda: Yeah, updated.

PR #8909, opened 2021-06-20
CI: Port Windows tests from Azure pipelines to GitHub workflow

jon-turney:


mensinda:
Is this actually going to be faster/beneficial? We only get [20 concurrent jobs](https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration) currently and we are already above that...

jon-turney:
> Is this actually going to be faster/beneficial?

Yeah, I was in two minds about whether this was worthwhile or not.

I doubt that it's any faster, since it's probably running on the same cloud, but the main benefit is that these tests get run in forks without any special effort...

PR #8911, opened 2021-06-20
Fix for Issue 8910 (Meson filters CMake asm files)

nanolith:
This patch adds a variable, assembler_suffixes, which is added to the list of supported suffixes in the CMake interpreter.

### DIFF
--- a/mesonbuild/compilers/compilers.py
+++ b/mesonbuild/compilers/compilers.py
@@ -74,8 +74,7 @@ c_suffixes = lang_suffixes['c'] + ('h',)  # type: T.Tuple[str, ...]
 # C ABI; these can generally be used interchangeably
 clib_langs = ('objcpp', 'cpp', 'objc', 'c', 'fortran',)  # type: T.Tuple[str, ...]
 # List of assembler suffixes that can be linked with C code directly by the linker
-assembler_suffixes = tuple() # type: T.Tuple[str, ...]
-assembler_suffixes += ('s', 'S')
+assembler_suffixes: T.Tuple[str, ...] = ('s', 'S')
 # List of languages that can be linked with C code directly by the linker
 # used in build.py:process_compilers() and build.py:get_dynamic_linker()
 clink_langs = ('d', 'cuda') + clib_langs  # type: T.Tuple[str, ...]
--- a/mesonbuild/compilers/compilers.py
+++ b/mesonbuild/compilers/compilers.py
@@ -73,6 +73,9 @@ c_suffixes = lang_suffixes['c'] + ('h',)  # type: T.Tuple[str, ...]
 # List of languages that by default consume and output libraries following the
 # C ABI; these can generally be used interchangeably
 clib_langs = ('objcpp', 'cpp', 'objc', 'c', 'fortran',)  # type: T.Tuple[str, ...]
+# List of assembler suffixes that can be linked with C code directly by the linker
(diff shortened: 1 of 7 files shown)


mesonbuild/compilers/compilers.py
```diff
@@ -73,6 +73,9 @@
 # List of languages that by default consume and output libraries following the
 # C ABI; these can generally be used interchangeably
 clib_langs = ('objcpp', 'cpp', 'objc', 'c', 'fortran',)  # type: T.Tuple[str, ...]
+# List of assembler suffixes that can be linked with C code directly by the linker
+assembler_suffixes = tuple() # type: T.Tuple[str, ...]
+assembler_suffixes += ('s', 'S')
```
mensinda, line 78: The first line is not neccessary. You can just do `assembler_suffixes: T.Tuple[str, ...] = ('s', 'S')`

mensinda (changes requested):
This really needs a test case (to ensure that it does not break in the future). Could you add a new minimal test project in https://github.com/mesonbuild/meson/tree/master/test%20cases/cmake?

nanolith:
@mensinda - Thanks for the prompt reply.  I will fix both issues momentarily.

mesonbuild/compilers/compilers.py
nanolith: Issue resolved in latest commit.

nanolith:
As requested, a test case was added to protect this change.  It fails on master, but it works with this patch.

The test case will work on Linux, OpenBSD, FreeBSD, and NetBSD systems running on x86, x86_64, arm, and aarch64 systems.  It is set to be skipped on any other systems.  Let me know if this is sufficient.  These are all I have to test locally.

mensinda (approved):
LGTM, lets see if the CI also agrees.

Thanks for fixing this.

test cases/cmake/25 assembler/meson.build
```diff
@@ -0,0 +1,9 @@
+project('assembler test', ['c'])
```
mensinda, line 1: You also need 'cpp' here, so that meson can tell CMake the correct C++ compiler.

Alternatively, only enable `c` in CMake by explicitly setting the languages in the CMake `project()` call.

mensinda (commented):
Besides the points in the comment below, you need to fix some warnings in your C code.

jpakkane:
Test failure seems relevant:

```
../test cases/cmake/25 assembler/main.c:4:1: error: function declaration isn’t a prototype [-Werror=strict-prototypes]
    4 | int32_t cmTestFunc();
      | ^~~~~~~
../test cases/cmake/25 assembler/main.c: In function ‘main’:
../test cases/cmake/25 assembler/main.c:6:14: error: unused parameter ‘argc’ [-Werror=unused-parameter]
    6 | int main(int argc, char* argv[])
      |          ~~~~^~~~
../test cases/cmake/25 assembler/main.c:6:26: error: unused parameter ‘argv’ [-Werror=unused-parameter]
    6 | int main(int argc, char* argv[])
      |                    ~~~~~~^~~~~~
cc1: some warnings being treated as errors
```

nanolith:
@jpakkane and @mensinda -

I enabled the warnings in these tests so I could see them when running locally.  I fixed the warnings and the meson.build test file as requested.

nanolith:
Please let me know if there is anything else needed to get this merged and to resolve the linked issue.  It looks like all automated tests are passing.

Issue #8907, opened 2021-06-19
Regression: Error in GStreamer configuration


xclaesse:
I get this error during gst-build configuration, I believe it's ObjectHolder refactoring: `subprojects/gstreamer/gst/meson.build:275:2: ERROR: Gir target must be an executable or library`.

xclaesse:
Bisected, the first that broke gst-build is `8f997efc8 decorators: Add not_set_warning to KwargInfo`, with that commit I get meson crash because of typing issues.

mensinda:
I have looked at the commit and this doesn't make any sense *at all*. Then again, I didn't make the PR particularly easy to bisect, since I didn't actually check the commits :(

The problem is caused by `orcc` using `override_find_program` and `gst-plugins-base` using `find_program('orcc').found()`. Thus `find_program('orcc')` now returns an executable which doesn't have the `.found()` method.

mensinda:
@xclaesse Are you sure that you have used the latest changes from my branch? I am unable to reproduce this bug locally (after I have fixed the unrelated `orcc` bug).

xclaesse:
Still happening with both PR #8908 and #8905 applied.

Here is the backtrace I get with `MESON_FORCE_BACKTRACE=1`:
```
subprojects/gstreamer/gst/meson.build:275:2: ERROR: Gir target must be an executable or library

A full log can be found at /home/xclaesse/programmation/gst-build/builddir/meson-logs/meson-log.txt
Traceback (most recent call last):
  File "/home/xclaesse/.local/bin/meson", line 29, in <module>
    sys.exit(mesonmain.main())
  File "/home/xclaesse/programmation/meson/mesonbuild/mesonmain.py", line 321, in main
    return run(sys.argv[1:], launcher)
  File "/home/xclaesse/programmation/meson/mesonbuild/mesonmain.py", line 311, in run
    return CommandLineParser().run(args)
  File "/home/xclaesse/programmation/meson/mesonbuild/mesonmain.py", line 223, in run
    return options.run_func(options)
  File "/home/xclaesse/programmation/meson/mesonbuild/msetup.py", line 281, in run
    app.generate()
  File "/home/xclaesse/programmation/meson/mesonbuild/msetup.py", line 184, in generate
    self._generate(env)
  File "/home/xclaesse/programmation/meson/mesonbuild/msetup.py", line 223, in _generate
    intr.run()
[...]

PR #8912, opened 2021-06-21
Fix `BothLibraries` processing

mensinda:


xclaesse:
I can confirm this fixes https://github.com/mesonbuild/meson/issues/8907, however I hit another bug later where it cannot find `found()` method on orcc, but I believe you had a fix for that already, right?

Sorry I don't have much time for doing review these days, got work stuff to close before going on vacations :(

mensinda:
@xclaesse You have to combine all 3 fixes PRs to to build gst-build :)

PR #8524, opened 2021-03-12
Add 'subprojects purge' command

tristan957:
This will help facilitate cache busting in certain situations, and
replaces hand-rolled solutions of writing a length command to remove
various files/folders within the subprojects directory.

PR #8914, opened 2021-06-21
Delete redirected wrap files in subprojects purge

tristan957:
We need to store the original filename as well as whether the wrap was
redirected in order to properly purge the redirected wrap.

Follow on to https://github.com/mesonbuild/meson/pull/8524. Maybe it is worth backporting.

### DIFF
--- a/mesonbuild/msubprojects.py
+++ b/mesonbuild/msubprojects.py
@@ -386,6 +386,12 @@ class Runner:
         if not self.wrap.type:
             return True
 
+        if self.wrap.redirected:
+            redirect_file = Path(self.wrap.original_filename).resolve()
+            if self.options.confirm:
+                redirect_file.unlink()
+            mlog.log(f'Deleting {redirect_file}')
+
         if self.wrap.type == 'redirect':
             redirect_file = Path(self.wrap.filename).resolve()
             if self.options.confirm:
@@ -416,7 +422,7 @@ class Runner:
             # parallelized, another thread could have deleted it already.
             try:
                 if not any(packagecache.iterdir()):
-                    packagecache.rmdir()
+                    windows_proof_rmtree(str(packagecache))
             except FileNotFoundError:
                 pass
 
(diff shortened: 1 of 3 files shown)


jpakkane:
It would be nice if one of the existing purge tests could be updated to test that the deletion works.

tristan957:
Ok I will see how easy it is and report back.

xclaesse:
Wondering if we should write into the redirect file itself that it has been generated by Meson, and delete only files that have that flag.

e.g.
```
[wrap-redirect]
...
generated = true
```

Thinking about this because in https://github.com/mesonbuild/meson/pull/8941 I download wraps from WrapDB but we could want to add that `generated = true` flag to purge them too?

tristan957:
I like that idea if it would help you out!

xclaesse:
Does not have to be done in this PR actually.

jpakkane:
⚔️ conflicts.

tristan957:
GGGGGGGGGGGGGGGGGGGGGGGGRRRRRRRRRRRRRRRRR. Fixing now

jpakkane:
```
mesonbuild/interpreter/interpreterobjects.py:730:6: error: Name "permissive_unholder_return" is not defined
mesonbuild/interpreter/interpreterobjects.py:730:6: error: Untyped decorator makes function "get_variable_method" untyped
```

jpakkane:
:crossed_swords: again, probably because it changed the same parts as the other MR that just got to master.

jpakkane:
```
  File "D:\a\meson\meson\mesonbuild\msubprojects.py", line 425, in purge
    packagecache.rmdir()
  File "D:\a\_temp\msys\msys64\mingw64\lib\python3.8\pathlib.py", line 1337, in rmdir
    self._accessor.rmdir(self)
PermissionError: [WinError 5] Access is denied: 'D:\\a\\meson\\meson\\tmp9l400ep6\\src\\subprojects\\packagecache'
```

Some virus scanner or the like is holding on to some files. If you change that line to use Meson's `windows_proof_rmtree` instead it should pass.

tristan957:
Wow. Thanks for the tip. I should have remembered this considering I've used this function.

PR #8917, opened 2021-06-22
Fix project tests category name used in skippable() and XML output

jon-turney:
Since 25df6e7d split the iteration over tests to start them from the iteration to collect their results, the variable `name` is only being
set in the first iteratiorn, so all tests are treated as being in the last test category read (probably `wasm`) for `skipppable()` and in the
XML output.


jon-turney:
I'm going to claim this is a bug introduced in #8852, so Cc: @mensinda

mensinda (approved):
LGTM. I really wish Python had more strict scoping to catch this kind of bug...

Issue #7295, opened 2020-06-10
Issue under MSYS2 on Windows when using ImDisk RAM disk


brechtsanders:
To speed up my Windows builds under MSYS2 I always use an ImDisk RAM disk.

Meson 0.54.0 and higher always seemed to be broken in my environment, but then I came across this Python issue: https://bugs.python.org/issue31842 which described exactly what I was experiencing.

So I tried building something that failed before - p11-kit v0.23.20 - on a regular Windows drive and apparently Meson 0.54.2 was working just fine in this case. On the RAM drive there were always Python errors.

After some digging and trial and error I came up with this solution which fixes the problem:
```
patch -ulbf mesonbuild/interpreter.py << EOF
@@ -2426,3 +2426,6 @@
             builddir = Path(self.environment.get_build_dir())
-            f = Path(f).resolve()
+            if mesonlib.is_windows():
+                f = Path(os.path.normpath(f))
+            else:
+                f = Path(f).resolve()
             if builddir in f.parents:
EOF
```

I'm not sure if this as a generally acceptible solution, but it may also fix issues on Windows with other types of non-local drives. It does however skip symlink resolution stuff, but I don't think that will be a problem on Windows.

I hope you will consider my solution, so I start using the next Meson release without modifications.

scivision:
Thank you, that's a nifty idea to use the RAM disk for builds.
I wonder if we tried catching the OSError exception since this isn't fixed in Python itself yet as per the bug report you cited.
@brechtsanders is `OSError` the exception that's being thrown for your situation?

scivision:
Oh, I should have looked at master branch first--it appears this idea was already implemented: https://github.com/mesonbuild/meson/blob/master/mesonbuild/interpreter.py#L2516
@brechtsanders I wonder if Meson master branch works for you? This was done about 12 days ago.

brechtsanders:
Just trued building GSSDP again with Meson master branch and it still fails on the RAM drive (on a local volume it works fine). Here's the output:
```
The Meson build system
Version: 0.54.999
Source dir: R:\winlibs64-10.1.0\gssdp-1.2.2
Build dir: R:\winlibs64-10.1.0\gssdp-1.2.2\build_both
Build type: native build
Traceback (most recent call last):
  File "D:\Prog\winlibs64-10.1.0\custombuilt\lib\mesonbuild\mesonmain.py", line
131, in run
    return options.run_func(options)
[...]

PR #8918, opened 2021-06-22
pathlib: Patch pathlib to work around some bugs (fixes #8263 #7295)

### LINKED ISSUE #8263: Codepage issue on Windows 8.1
**Describe the bug**
I receive a build failure when building `harfbuzz` through `vcpkg`. `vcpkg` maintainers believe it is a `meson` upstream bug with codepages. The downstream bug is microsoft/vcpkg#15916. I have no ability to file a sane bug report, since I know nothing. However, I have logs.

**To Reproduce**
I can only reproduce this when building harfbuzz with vcpkg, using `./vcpkg install harfbuzz`. I don't receive this error when building harfbuzz with the project's native meson instructions without vcpkg. These builds are on the same system, but my native build uses mingw-w64's gcc, while the vcpkg build uses Visual Studio. The compiler isn't yet invoked when the build error occurs, so I don't believe this difference of compilers is the cause.

**Expected behavior**
I expected it to build cleanly. This is the log:

```
[3/66] "C:\Users\adm\Downloads\vcpkg\downloads\tools\python\python-3.9.0-x64\python.exe" "C:\Users\adm\Downloads\vcpkg\downloads\tools\meson\meson-f5871f434a5b768ad9fcafa797a6db0286421842\meson.py" "--internal" "depscan" "src/harfbuzz.dll.p/harfbuzz.dat" src/harfbuzz.dll.p/depscan.dd ../src/f4d3aa1127-a820a32874.clean/src/hb-aat-layout.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-aat-map.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-blob.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-buffer-serialize.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-buffer.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-common.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-draw.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-face.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-fallback-shape.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-font.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-map.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-number.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-cff1-table.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-cff2-table.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-color.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-face.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-font.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-layout.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-map.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-math.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-meta.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-metrics.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-name.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-shape-complex-arabic.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-shape-complex-default.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-shape-complex-hangul.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-shape-complex-hebrew.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-shape-complex-indic-table.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-shape-complex-indic.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-shape-complex-khmer.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-shape-complex-myanmar.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-shape-complex-thai.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-shape-complex-use-table.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-shape-complex-use.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-shape-complex-vowel-constraints.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-shape-fallback.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-shape-normalize.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-shape.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-tag.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-var.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-set.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-shape-plan.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-shape.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-shaper.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-static.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-style.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ucd.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-unicode.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ft.cc
FAILED: src/harfbuzz.dll.p/depscan.dd 
"C:\Users\adm\Downloads\vcpkg\downloads\tools\python\python-3.9.0-x64\python.exe" "C:\Users\adm\Downloads\vcpkg\downloads\tools\meson\meson-f5871f434a5b768ad9fcafa797a6db0286421842\meson.py" "--internal" "depscan" "src/harfbuzz.dll.p/harfbuzz.dat" src/harfbuzz.dll.p/depscan.dd ../src/f4d3aa1127-a820a32874.clean/src/hb-aat-layout.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-aat-map.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-blob.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-buffer-serialize.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-buffer.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-common.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-draw.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-face.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-fallback-shape.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-font.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-map.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-number.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-cff1-table.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-cff2-table.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-color.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-face.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-font.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-layout.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-map.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-math.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-meta.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-metrics.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-name.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-shape-complex-arabic.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-shape-complex-default.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-shape-complex-hangul.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-shape-complex-hebrew.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-shape-complex-indic-table.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-shape-complex-indic.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-shape-complex-khmer.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-shape-complex-myanmar.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-shape-complex-thai.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-shape-complex-use-table.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-shape-complex-use.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-shape-complex-vowel-constraints.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-shape-fallback.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-shape-normalize.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-shape.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-tag.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ot-var.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-set.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-shape-plan.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-shape.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-shaper.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-static.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-style.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ucd.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-unicode.cc ../src/f4d3aa1127-a820a32874.clean/src/hb-ft.cc
  File "C:\Users\adm\Downloads\vcpkg\downloads\tools\meson\meson-f5871f434a5b768ad9fcafa797a6db0286421842\meson.py", line 29, in <module>
    sys.exit(mesonmain.main())
  File "C:\Users\adm\Downloads\vcpkg\downloads\tools\meson\meson-f5871f434a5b768ad9fcafa797a6db0286421842\mesonbuild\mesonmain.py", line 229, in main
  File "C:\Users\adm\Downloads\vcpkg\downloads\tools\meson\meson-f5871f434a5b768ad9fcafa797a6db0286421842\mesonbuild\mesonmain.py", line 218, in run
  File "C:\Users\adm\Downloads\vcpkg\downloads\tools\meson\meson-f5871f434a5b768ad9fcafa797a6db0286421842\mesonbuild\mesonmain.py", line 166, in run_script_command
  File "C:\Users\adm\Downloads\vcpkg\downloads\tools\meson\meson-f5871f434a5b768ad9fcafa797a6db0286421842\mesonbuild\scripts\depscan.py", line 91, in run
    return scanner.scan()
  File "C:\Users\adm\Downloads\vcpkg\downloads\tools\meson\meson-f5871f434a5b768ad9fcafa797a6db0286421842\mesonbuild\scripts\depscan.py", line 64, in scan
    self.scan_file(s)
  File "C:\Users\adm\Downloads\vcpkg\downloads\tools\meson\meson-f5871f434a5b768ad9fcafa797a6db0286421842\mesonbuild\scripts\depscan.py", line 37, in scan_file
    for line in pathlib.Path(fname).read_text().split('\n'):
  File "pathlib.py", line 1256, in read_text
  File "encodings\cp1252.py", line 23, in decode
UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 36961: character maps to <undefined>
```

Attached: [src\harfbuzz.dll.p\harfbuzz.dat](https://github.com/mesonbuild/meson/files/5883593/harfbuzz.txt)
Attached: [package-x64-windows-rel-out.log](https://github.com/mesonbuild/meson/files/5883605/package-x64-windows-rel-out.log), probably no need to look at this

**system parameters**
* native build
* Windows 8.1
* Python 3.9.1
* meson 0.56.2
* ninja 1.10.2

mensinda:


eli-schwartz:
Test failures seem very relevant:
```
_pickle.PicklingError: Can't pickle <class 'pathlib.PosixPath'>: attribute lookup PosixPath on pathlib failed
```

I don't understand the PR anyway, per @jpakkane 
> Rather than monkeypatching, it would be better to change the call sites. The thing about monkeypatching is that it always comes back to bite you in the ass.

But here we are, monkeypatching `sys.modules['pathlib']`.

Last time we were hacking around pathlib, we did this everywhere:

```
-from pathlib import Path
+from mesonbuild._pathlib import Path
```

eli-schwartz:
Given this (#8263) is a "feature" of python, not a bug, it feels wrong to modify pathlib as we'd need to keep this code around "literally forever". Why is passing explicit encoding= so unpalatable to you?

jpakkane:
> we have no good way of ensuring that we actually always pass encoding='utf-8'.

One can write a unit that that regexes all `.py` files for `read_text` that is not followed by `encoding`. It's not 100% reliable but for this 95% should be good enough.

mensinda:
> One can write a unit that that regexes all .py files for read_text that is not followed by encoding. It's not 100% reliable but for this 95% should be good enough.

Yeah, this might be the better solution. Let's see if I can get something to work.

mensinda:
OK, changed **all** the `open()` and `{read,write}_text()` to always use `encoding=` (and have a script check it).

jpakkane:
The `_pathlib` file should be removed then.

eli-schwartz:
> NEW  The module 'mesonbuild._pathlib' imports itself.

Erm, what. This check looks quite broken, lol, _pathlib is not pathlib

> NEW   Import of 'T' is not used.

Accurate analysis...

Given the overridden _pathlib.py is now only here to fix these unusual ImDisk issues, we might want to defer it to upstream python?

mensinda:
@jpakkane The `_pathlib` module is still required to work around the actual python bug with `.resolve()`.

https://bugs.python.org/issue31842
https://github.com/mesonbuild/meson/issues/7295

mensinda:
Fixed, the unused typing issue.

mesonbuild/backend/ninjabackend.py
```diff
@@ -462,10 +462,8 @@ def detect_vs_dep_prefix(self, tempfilename):
             return open(tempfilename, 'a', encoding='utf-8')
         filename = os.path.join(self.environment.get_scratch_dir(),
                                 'incdetect.c')
-        with open(filename, 'w') as f:
-            f.write('''#include<stdio.h>
-int dummy;
-''')
+        with open(filename, 'w', encoding='utf-8') as f:
+            f.write('#include<stdio.h>\nint dummy;\n')
```
eli-schwartz, line 466: Unrelated string change.

Also I think the change looks ugly!

run_custom_lint.py
```diff
@@ -0,0 +1,78 @@
...
+def check_missing_encoding(lines: T.List[str], path: str) -> int:
+    errors = 0
+    functions = ['read_text', 'write_text', 'open']
+    for num, line in enumerate(lines):
+        for func in functions:
+            l = line
+
+            # Skip ignored lines
+            if '[ignore encoding]' in l:
+                continue
```
eli-schwartz, line 20: Why do we need magic comments specific to this script, in order to flag allowed uses of encoding='something other than utf-8'?

If something intentionally other than utf-8 is being used, then it's obviously intentional and therefore obviously fine (at least, ***it will behave the same on Linux and Windows***). So we should just permit it. No need to later on check that the encoding is specifically either 'ascii' or 'utf-8'.

eli-schwartz (changes requested):

mesonbuild/backend/ninjabackend.py
dcbaker: textwrap.dedent is best for this I think

run_custom_lint.py
mensinda: Fine, I removed the special cases for `'ascii'` and `'utf-8'`, but this is still needed for stuff like `tarfile.open(`.

mesonbuild/backend/ninjabackend.py
mensinda: switched to `dedent`.

mesonbuild/_pathlib.py
```diff
@@ -0,0 +1,43 @@
...
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pathlib
+import os
+
```
jpakkane, line 17: It does not seem possible to do this without monkey patching, so that's what we need to do then. However the implementation needs some fixes:

- the name should not be `_pathlib` as leading underscores in names are confusing and can lead to problems (as can be seen with the CI), instead it should be something more dangerous sounding like `pybugXXXXworkaround`
- The file should have a comment at the top explaining what it is, why it's needed, have links to the relevant bug reports, a notice that no new code or functionality should be added to it and also when the workaround can be removed.

mensinda: Done.

mensinda: Well, I guess renaming the file did not fix the `1 for Module imports itself` warning...

jpakkane: That seems like an LGTM bug then. Let's just rename it `pathlibworkaround` and call it a day.

mensinda: So, do you want me to rename it again, or does the current name work?

jpakkane: According to the web ui the module name is `zzz_pathlibworkaround` which is not a great name ...

mensinda: But, then it looks like a regular meson module and blends in with the rest of them, which goes against the point of making it look scary. Also, we already have the `_typing` module...

Would you be ok with renaming it to `zzzpathlibworkaround`? Also which web ui is complaining about the name?

eli-schwartz: I believe LGTM is complaining that whatever the name of your module, it's being inserted into `sys.modules['pathlib']` while at the same time importing `pathlib`.

Which I still think is a bad idea, when we have _typing we import from _typing rather than inserting it into `sys.modules['typing']`.

jpakkane: > Also which web ui is complaining about the name?

No system complains, this is 100% my own esthetical eye thing. A name like `zzz_anything` reads like "a placeholder name thrown in to test something which will be replaced with the real name later". I don't really have strong opinions about the naming, it just seemed strange to me.

mensinda: The difference to `_typing` is that it is *never* actually imported in production. It is only used in `if T.TYPE_CHECKING:` blocks.

If we do the same with `pathlib` and `_pathlib` we have to replace all `import pathlib` and `from pathlib import *`, remember to always to use `_pathlib` (which can result in non-obvious errors otherwise). And once we remove the workaround we have to revert all these changes again... IMHO, at some point one dirty hack is more convenient and developer-friendly than renaming 50 different strings and enforcing that.

@jpakkane Would you then be OK, to revert to renaming it to `_pathlib` since LGTM is just broken in this regard? I really don't want it to be sorted in the middle of the other meson modules and have it blend in and look more legitimate than it is.

jpakkane: Fine by me.

jpakkane: Actually, would `import _pathlib as pathlib` work better?

mensinda: If we do `import _pathlib as pathlib` or `from _pathlib import Path`, then we have to do it *everywhere* and *consitently*, which is why I wanted to do the `sys.modules['typing']` hack in the first place. This definetely is not the most elegant solution, but the soultion with the smallest git diff (not that this would make a difference after we went for fixing `encoding` on the call site...) and the least maintenance burden.

jpakkane: Let's go with the current one then, it can be always fixed later. One thing to note, though, is that since this issue only occurs on Windows, the monkeypatching should probably also be done on Windows.

mensinda: *If* we do `import _pathlib` we need *another* custom linter for a semi temporary fix/hack to ensure that we don't mix `_pathlib` and `pathlib`

mensinda: Done.

PR #8923, opened 2021-06-23
Add feed arg to custom_target()

emersion:
An example of a tool not accepting an input file param is [scdoc](https://git.sr.ht/~sircmpwn/scdoc/).

### DIFF
--- a/docs/markdown/Reference-manual.md
+++ b/docs/markdown/Reference-manual.md
@@ -383,6 +383,10 @@ following.
   `{'NAME1': 'value1', 'NAME2': 'value2'}` or `['NAME1=value1', 'NAME2=value2']`,
   or an [`environment()` object](#environment-object) which allows more
   sophisticated environment juggling.
+- `feed` *(since 0.59.0)*: there are some compilers that can't be told to read
+  their input from a file and instead read it from standard input. When this
+  argument is set to true, Meson feeds the input file to `stdin`. Note that
+  your argument list may not contain `@INPUT@` when feed mode is active.
 
 The list of strings passed to the `command` keyword argument accept
 the following special string substitutions:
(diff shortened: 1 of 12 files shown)


docs/markdown/Reference-manual.md
```diff
@@ -383,6 +383,10 @@ following.
   `{'NAME1': 'value1', 'NAME2': 'value2'}` or `['NAME1=value1', 'NAME2=value2']`,
   or an [`environment()` object](#environment-object) which allows more
   sophisticated environment juggling.
+- `feed` *(since 0.59.0)*: there are some compilers that can't be told to read
```
jon-turney, line 386: Perhaps want to write 'programs' rather than 'compilers'?

test cases/common/243 custom target feed/meson.build
```diff
@@ -0,0 +1,24 @@
+project('custom target feed', 'c')
+
+python3 = import('python3').find_python()
+
+# Note that this will not add a dependency to the compiler executable.
+# Code will not be rebuilt if it changes.
+comp = '@0@/@1@'.format(meson.current_source_dir(), 'my_compiler.py')
```
jon-turney, line 7: Um... why are we doing this?

jon-turney (changes requested):

test cases/common/243 custom target feed/meson.build
emersion: This is copy-pasta from `109 custom target capture`.

docs/markdown/Reference-manual.md
emersion: Right. The `capture` description mentions "compilers", so I just went with that. Can change if "programs" is better.

test cases/common/243 custom target feed/meson.build
jon-turney: Yeah, that seems to be that way since it was added in 70d94a555004563178878b73274e345adff5b4e8.  I have no idea why :)

jpakkane:
This requires a release note snippet.

emersion:
Done.

emersion:
Thanks for the review!

PR #8924, opened 2021-06-23
Various fixes to test.json schema

jon-turney:


jon-turney:
> Fix test.json for boost test to match schema

No idea what the intent was here, so Cc: @mensinda

eli-schwartz:
> Is it right that an array option gets turned into a ';'-separated string for cmake?  Is that something we need a test for?

One of a number of reasons why meson is infinitely better designed than cmake. The CMakeLists.txt build description format does not have an array type, and functions which know they're supposed to be using arrays do so by internally storing them as `;`-separated strings, then writing functions that know they *only* ever operate on lists, recognize the semicolon and do string splitting before operating. (The builtin foreach, for example.) Otherwise I suppose you just guess.

The command line interface to that is, of course, strings too. ;) So in order for meson to pass a list into cmake, it was literally running `cmake -DCMAKE_PREFIX_PATH="val1;val2"`.

I do not know why CMAKE_PREFIX_PATH was chosen here, but I suppose it might just be testing that meson does this correctly. Even though CMAKE_PREFIX_PATH is not supposed to be given a list, it may just have been the most convenient hack to test said codepath.

test cases/cmake/7 cmake options/subprojects/cmOpts/CMakeLists.txt
```diff
@@ -5,6 +5,6 @@ if(NOT "${SOME_CMAKE_VAR}" STREQUAL "something")
   message(FATAL_ERROR "Setting the CMake var failed")
 endif()
 
-if(NOT "${CMAKE_PREFIX_PATH}" STREQUAL "val1;val2")
+if(NOT "${CMAKE_PREFIX_PATH}" STREQUAL "/some/prefix/path")
```
mensinda, line 8: We really should keep testing for list conversion/handling, so please revert that change.

mensinda (changes requested):
I can't really comment on the JSON schema itself, since I didn't create it originally. It was added after I introduced the single `test.json` format.

However, the changes to the CMake array handling should be reverted.

test cases/cmake/7 cmake options/subprojects/cmOpts/CMakeLists.txt
jon-turney: I dropped that change, and adjusted the schema to allow `val:` and `exclude:` to contain arrays of strings as well.

jon-turney:
> However, the changes to (testing) the CMake array handling should be reverted

Fair enough.  

(If there was a comment in 0332d7e35006afa78afb5e4b4f4d2ecf9a7ded27 or #7254 that we're also testing that, I wouldn't have had to ask :) )

mensinda (approved):

PR #8925, opened 2021-06-23
interpreter: Empty list used to be allowed in install_headers/man

xclaesse:
That change introduced when porting to @typed_pos_args breaks gtk4. We
could decide to deprecate/warn but should not error for backward
compatibility.


dcbaker (approved):
It seems odd to me, but if it's what we've been doing...

xclaesse:
I agree it's odd, in the gtk case it's really just a placeholder: https://gitlab.gnome.org/GNOME/gtk/-/blob/master/gtk/deprecated/meson.build. But I could also imagine cases where the list of installed headers depends on options and when everything is disabled that list end up empty... Granted it's super niche.

Issue #8772, opened 2021-05-17
Compiler.check_header fails if the checked header includes header configured using configure_file


woju:
**Describe the bug**
When performing compiler checks, including header checks, meson doesn't add build directories when adding `-I` using `include_directories`. Only directories from source directories are added. This deviates from usual behaviour of `include_directories()`.

**To Reproduce**
```meson
project('project', 'c')
configure_file(
    input: 'config.h.in',
    output: 'config.h',
    configuration: configuration_data({'MACRO': true}))
meson.get_compiler('c').check_header('header.h', include_directories: include_directories('.'))
```
`header.h`:
```c
#include <config.h>
```
`config.h.in`:
```c
#mesondefine MACRO
```
**Expected behavior**

I expect this check passes.

**Actual behaviour**

```
The Meson build system
Version: 0.53.2
Source dir: /home/woju/meson-conf-header
Build dir: /home/woju/meson-conf-header/build
Build type: native build
Project name: project
Project version: undefined
C compiler for the host machine: cc (gcc 9.3.0 "cc (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0")
C linker for the host machine: cc ld.bfd 2.34
Host machine cpu family: x86_64
Host machine cpu: x86_64
Configuring config.h using configuration
Check usable header "header.h" : NO 
Build targets in project: 0

Found ninja-1.10.0 at /usr/bin/ninja
```

Workaround:
```
meson.get_compiler('c').check_header('header.h', include_directories: include_directories('.'),
    args: '-I@0@'.format(meson.current_build_dir()))
```

**system parameters**
* Is this a [cross build](https://mesonbuild.com/Cross-compilation.html) or just a plain native build (for the same computer)? no
* what operating system (e.g. MacOS Catalina, Windows 10, CentOS 8.0, Ubuntu 18.04, etc.) Ubuntu 20.04
* what Python version are you using e.g. 3.8.0: 3.8
* what `meson --version` 0.53.2, but IIUC this is also on current master
* what `ninja --version` if it's a Ninja build 1.10, I think this doesn't matter

LessiKai:
Seems reasonable.
I would try to fix this issue if nobody else is currently working on it.

dcbaker:
I don't think anyone else is

JamBalaya56562:
[...]

PR #8931, opened 2021-06-25
Automatically added the build dir to include directories in the compiler.check_header routine in case include_directories are passed.

LessiKai:
This prevents generated config header files from not being recognized when checking for them, because they only occur in the build dir and not in the source tree.
Link to issue: [https://github.com/mesonbuild/meson/issues/8772](url)

### DIFF
--- a/mesonbuild/compilers/mixins/clike.py
+++ b/mesonbuild/compilers/mixins/clike.py
@@ -347,6 +347,18 @@ class CLikeCompiler(Compiler):
                      dependencies: T.Optional[T.List['Dependency']] = None) -> T.Tuple[bool, bool]:
         code = f'''{prefix}
         #include <{hname}>'''
+
+        # Change extra_args to list of strings
+        if callable(extra_args):
+            extra_args = extra_args("compile")
+
+        if extra_args is not None:
+            builddir = Path(env.build_dir)
+
+            # Append builddir to include directories in case config_header is used before check header
+            if any(arg.startswith("-I") and not builddir.samefile(arg[2:]) for arg in extra_args):
+                extra_args.append(f"-I{builddir.as_posix()}")
+
         return self.compiles(code, env, extra_args=extra_args,
                              dependencies=dependencies)
 


mesonbuild/compilers/mixins/clike.py
```diff
@@ -347,6 +347,18 @@ def check_header(self, hname: str, prefix: str, env: 'Environment', *,
                      dependencies: T.Optional[T.List['Dependency']] = None) -> T.Tuple[bool, bool]:
         code = f'''{prefix}
         #include <{hname}>'''
+
+        # Change extra_args to list of strings
+        if callable(extra_args):
+            extra_args = extra_args("compile")
+
+        if extra_args is not None:
```
mensinda, line 355: This causes the build dir not to be added in all circumstances. Shouldn't the build dir be added unconditionally?

mensinda (changes requested):
This really needs a test case. Please either update an existing test case or add a new one in https://github.com/mesonbuild/meson/tree/master/test%20cases/common

mesonbuild/compilers/mixins/clike.py
LessiKai: In this case the build dir is only added if the include_directories parameter is used and does not already include the build dir.  I thought it would be a more natural usage like this. In addition the include_directories parameter has to be used anyways as far my understanding goes.

mensinda: Well, this currently checks if *any* args are passed (so technically adding `-O2` should also trigger this). Also, it might be better to move this code to https://github.com/mesonbuild/meson/blob/master/mesonbuild/interpreter/compiler.py. This should also give you more options.

Additionally, if adding the build dir is conditional (which I am still not sure that it is a good thing), it also needs to be documented.

LessiKai:
When calling the check_header method the build_dir is now always added. 
The logic was moved from the previous clike compilers to the more generic compiler class.
A specific test case was added too.
A previous test case had to be removed, because it contradicted the fact that the build_dir is always added.

mensinda:
This still needs a release note snippet and must be signed off by @jpakkane since this is a change in behavior. Otherwise LGTM.

PR #8933, opened 2021-06-25
refactor: Refactor BothLibraries logic

mensinda:
This commit introduces a new type of `HoldableObject`: The
`SecondLevelHolder`. The primary purpose of this class is
to handle cases where two (or more) `HoldableObject`s are
stored at the same time (with one default object). The
best (and currently only) example here is the `BothLibraries`
class.

### DIFF
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -28,7 +28,7 @@ from . import dependencies
 from . import mlog
 from . import programs
 from .mesonlib import (
-    HoldableObject,
+    HoldableObject, SecondLevelHolder,
     File, MesonException, MachineChoice, PerMachine, OrderedSet, listify,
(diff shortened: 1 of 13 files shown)


mesonbuild/build.py
```diff
@@ -1264,9 +1262,6 @@ def link(self, target):
 
     def link_whole(self, target):
         for t in listify(target):
-            # Always use the static library from BothLibraries, since shared libs aren't supported anyway
-            if isinstance(t, BothLibraries):
-                t = t.static
```
xclaesse, line 1269: In this case you still need to pick the static one somehow, no?

xclaesse: nevermind, that was not supported in 0.58, it's fine to leave it broken.

xclaesse:
Tested and it works for me. LGTM when CI is green.

PR #8934, opened 2021-06-26
Generate Users.md

jon-turney:


jpakkane:
Note that `yaml` is not in the Python standard library, it is an external dependency. Let's use JSON instead, that is what we use in all other places as well, no point in adding the need for a brand new parser.

mensinda:
I don't think that it makes any difference here, since this does not affect the meson core and hotdoc already [depends on YAML](https://github.com/hotdoc/hotdoc/blob/master/hotdoc/core/tree.py#L32). Also, YAML is much nicer to write and read as a user than JSON. Though I might be a bit biased here because of #8960 :)

However, I would recommend the use of [strictyaml](https://github.com/crdoconnor/strictyaml) instead of YAML, because of the built-in shema support.

jon-turney:
Rebased and updated for changes in master

> Also, YAML is much nicer to write and read as a user than JSON.

Yeah, the choice of YAML was deliberate here, since JSON is fine for programs to read and write, not so much for people.

TODO:
- [ ] Use strictyaml

Issue #8739, opened 2021-05-07
On Debian and derivatives, `python` and `python3` modules are wrong about installation paths


woju:
**Describe the bug**
On Debian and derivatives, Python paths returned by `import('python3').sysconfig_path()` and `import('python').find_installation().get_path()` are not in `sys.path` (for `--prefix=/usr/local` and `--prefix=/usr`), because they have slight differences (`dist-packages` vs `site-packages`, `pythonX` vs `pythonX.Y`).

**To Reproduce**
```meson
python3_pkgdir = join_paths(import('python3').sysconfig_path('platlib'), 'packagename')
install_data('__init__.py', install_dir: python3_pkgdir)

import('python').find_installation().install_sources('module.py')
```

**Expected behavior**
I expect that if using `--prefix=/usr` or `--prefix=/usr/local` the package will be installed so that it can be imported without explicitly setting `$PYTHONPATH`.

**system parameters**
* Is this a [cross build](https://mesonbuild.com/Cross-compilation.html) or just a plain native build (for the same computer)?
  * native (didn't check for cross-compilation)

what operating system (e.g. MacOS Catalina, Windows 10, CentOS 8.0, Ubuntu 18.04, etc.) | Ubuntu 18.04 | Debian 10 | Ubuntu 20.04 | &nbsp;
--- | --- | --- | --- | ---
what Python version are you using e.g. 3.8.0 | 3.6 | 3.7 | 3.8 | i.e. system python
what `meson --version` | 0.45.1 | 0.49.2 | 0.53.2 | but I believe this is reproducible on current master
what `ninja --version` if it's a Ninja build | 1.8.2 | 1.8.2 | 1.10.0 | N/A I think

**further information**

There is a post https://discuss.python.org/t/pep-632-deprecate-distutils-module/5134/122 (the discussion is revolving around PEP 632, which deprecates `distutils` package) which describes the status quo, but the explanation is incomplete. What really happens is that Debian patches only `distutils` module and not `sysconfig`, and furthermore those changes only really work for `--prefix=/usr`, because with `--prefix=/usr/local` the paths are invalid because of the `pythonX` vs `pythonX.Y` difference.

There are 4 approaches to this problem using only standard library:
- `sysconfig` (doesn't work, because on debian and derivatives it returns wrong things)
- `distutils.sysconfig` (only works for `--prefix=/usr`)
- `site` (AFAIK there is no clear way to infer the correct data from any function there)
[...]

PR #8939, opened 2021-06-28
Add a note about Debian's dist-packages to Python-module.md

woju:
This is a result of discussion in #8739.


docs/markdown/Python-module.md
```diff
@@ -238,6 +238,63 @@ module.
 
 It exposes the same methods as its parent class.
 
+## A note about Debian installation paths
+
+Debian and derivatives (Ubuntu, Linux Mint etc.) [change default installation
+and import paths][debian-dist-packages] in system interpreter. The problem is,
```
jon-turney, line 244: This needs an article. "the system interpreter"?  But I'm not sure what this means?  "the system Python interpreter"?

docs/markdown/Python-module.md
```diff
@@ -238,6 +238,63 @@ module.
...
+the way they do this is inconsistent with documented ways of discovering those
+paths inside Python. Specifically, the [`sysconfig` module][python-sysconfig]
+returns values that are not in [`sys.path`][python-sys-path]. This module is
+used internally by Meson and therefore the installation paths . But it is
```
jon-turney, line 248: This sentence is confusing. "therefore in the installation paths"?  Or "therefore the installation paths are _something_"?

docs/markdown/Python-module.md
```diff
@@ -238,6 +238,63 @@ module.
...
+generally expected that if you configure build with `--prefix=/usr` or
```
jon-turney, line 249: 'configure a build'

docs/markdown/Python-module.md
```diff
@@ -238,6 +238,63 @@ module.
...
+`--prefix=/usr/local` (the latter is the default), Python modules should be
+available without adjusting [`$PYTHONPATH`][python-pythonpath]. Therefore,
+a&nbsp;workaround is needed. There are two, non mutually exclusive options:
```
jon-turney, line 252: why `&nbsp;` here?
'non-mutually' should be hypenated. 
putting it in a postive form ("two solutions, which can be used in combination") might be clearer?

docs/markdown/Python-module.md
```diff
@@ -238,6 +238,63 @@ module.
...
+
+### option to override
+
+An option in `meson_options.txt` might allow overriding the value obtained from
+[][`get_install_dir()`], and the files are installed using `install_data()`.
```
jon-turney, line 257: Is this a missing a reference-link name?

docs/markdown/Python-module.md
woju: Fixed.

docs/markdown/Python-module.md
woju: Oops, this was misdrafted, the second part of the sentence was cut and pasted elsewhere. Sorry for this.

docs/markdown/Python-module.md
woju: Fixed.

docs/markdown/Python-module.md
woju: I don't understand. Do you mean get_install_dir()? It works, I copied the approach from elsewhere in this module.

woju (commented):
Sorry for the delay. I pushed the updated commit.

docs/markdown/Python-module.md
eli-schwartz: https://hotdoc.github.io/syntax-extensions.html#referencing-a-symbol

> The label can be empty, in which case it will be set to the name of the symbol at format time

docs/markdown/Python-module.md
eli-schwartz: This is still not fixed, please fix the formatting.

xclaesse:
I've been hit by this today, wasted half a day of work until I found the reason why my python modules were not found.

IMHO documenting this is not enough, we must fix it.
- Did we raise the issue with Debian? That seems pretty critical issue in their distro that they MUST fix.
- In python module, `INTROSPECT_COMMAND` can return `sys.path` too. If none of self.platlib_install_path are in sys.path, assume it's a broken debian and hardcode `/usr/lib/python3/dist-packages`. Possibly also check mesonlib.is_debianlike().

eli-schwartz:
Did we communicate with debian? Debian knows, it's been a source of argument between the pip maintainers and the python for debian maintainers, on the pip bug tracker.

This is a completely ecosystem wide problem.

As for saving this for real in meson, I mentioned on the linked ticket:
> We could also continue to discuss special casing this in meson just for Debian's sake so workarounds are not needed (we even do this for --libdir defaults although there we don't have an official API like sysconfig to return what the OS is supposed to use), but so far I'm the only core maintainer who responded to the ticket so I dunno what everyone thinks here. I'd really like to know what the general view of the team is on handling Debian here...

eli-schwartz:
is_debianlike seems not unreasonable.

xclaesse:
I opened https://github.com/mesonbuild/meson/pull/9147 for a workaround and warn when we can detect the situation. I guess we should still document the problem too, so I don't think that PR replace this PR, they complete each other?

dcbaker:
I still think the right way to handle this is with a site native config file that could be shipped in the Debian package. That way we don't have to have a bunch of Debian work arounds and they're maintainers can just add a a config file and be done with it

eli-schwartz:
That is not a viable solution for making meson functional on a platform, users who pip install meson won't have Debian's site native config file.

docs/markdown/Python-module.md
```diff
@@ -238,6 +238,64 @@ module.
...
+
+    if is_debian and is_usr_local:
+        return distutils.util.subst_vars(
+            distutils.command.install.INSTALL_SCHEMES['unix_local']['platlib'],
+            {
+                'platbase': '/usr',
+                'py_version_short': '.'.join(map(str, sys.version_info[:2])),
+            })
+
+    return distutils.sysconfig.get_python_lib(plat_specific=True, prefix=prefix)
```
FFY00, line 290: `distutils` is deprecated in Python 3.10, I recommend avoiding this usage.

Since this solution is still tied to the Debian patching, I think it is almost as fragile as simply hardcoding the value, and adding a sanity check to make sure it is in `sys.path`.

So, I would do
```suggestion
    try:
        sysconfig.get_paths('deb_system')
    except KeyError:
        path = '/usr/local/lib/python{}.{}/dist-packages'.format(sys.version_info[0], sys.version_info[1])
        if path not in sys.path:
            raise KeyError('Expected Debian Python platlib `{}` not in sys.path! Please open an issue.')
        return platlib
    return sysconfig.get_path('platlib')
```

xclaesse: In master we already hard code the path for Debian derivatives. And we warn if it's not in sys.path. note that it will warn on Windows too. What remains to be done is to explain more in details the reason of that warning and how to avoid it by using an option to override the path.

PR #8941, opened 2021-06-28
Automatically use WrapDB fallback

xclaesse:
New command `meson wrap update-db` downloads the DB into `subprojects/wrapdb.json`. If that file exists, dependency() will automatically download missing wrap files from WrapDB.

### DIFF
--- a/docs/markdown/Using-wraptool.md
+++ b/docs/markdown/Using-wraptool.md
@@ -82,3 +82,17 @@ straightforward:
 Wraptool can do other things besides these. Documentation for these
 can be found in the command line help, which can be accessed by
 `meson wrap --help`.
+
+## Automatic dependency fallback
+
+Since *0.64.0* Meson can use WrapDB to automatically find missing dependencies.
+
+The user simply needs to download latest database, the following command stores
+it in `subprojects/wrapdb.json`:
+    $ meson wrap update-db
+
+Once the database is available locally, any dependency not found on the system
+but available in WrapDB will automatically be downloaded.
+
+Automatic fetch of WrapDB subprojects can be disabled by removing the file
+`subprojects/wrapdb.json`, or by using `--wrap-mode=nodownload`.
--- a/docs/markdown/snippets/wrapdb.md
+++ b/docs/markdown/snippets/wrapdb.md
@@ -1,0 +1,6 @@
+## Automatic fallback using WrapDB
+
+A new command has been added: `meson wrap update-db`. It downloads the list of
+wraps available in [WrapDB](wrapdb.mesonbuild.com) and stores it locally in
+`subprojects/wrapdb.json`. When that file exists and a dependency is not found
+on the system but is available in WrapDB, Meson will automatically download it.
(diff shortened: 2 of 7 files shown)


tp-m:
Not that it's not a cool feature, but will there be an option for project maintainers to disable this or provide a positive-list of 'approved wraps' or at least make users jump through hoops?

e.g. as a gst-build maintainer I will then have to deal with everyone who's got any problem with any wrap that's in the db...

xclaesse:
@tp-m It should not really affect GStreamer since we have wrap files already. It won't use WrapDB for deps already provided by a wrap. Users can already do `meson wrap install <foo>` for each dep manually, it just automate the process.

xclaesse:
Note that you could commit your own wrapdb.json file into your project. It won't prevent users from overriding it with latest one from WrapDB (by running `meson wrap update-db`), but it could make sense for a project to commit a subset of the json with specific versions available in WrapDB, for example. When all your deps are available in WrapDB (which is a long term goal of the v2 simplification) then that could be easier to ship your wrapdb.json than each individual wrap file.

xclaesse:
@dabrain34 the wrapdb json file is not downloaded by default, the user needs to run `meson wrap update-db` to fetch the json first. That makes this feature opt-in and not enabled by default.

If the json file is there when meson configure the project, for any missing dependency that it can find in the json, it will download the wrap file from wrapdb and use it. It is then up to the user to commit/push those downloaded .wrap files into their project if they want.

xclaesse:
Rebased this PR, any one interested in doing a review please?

jpakkane:
Functionality-wise this is something we do want, it makes a lot of things easier. Opt-in is probably the correct default here. There should be a way of removing the local db so that automatic resolution no longer happens.

docs/markdown/Using-wraptool.md
```diff
@@ -82,3 +82,13 @@ straightforward:
...
+
+## Automatic dependency fallback
+
+Since *0.64.0* Meson can use WrapDB to automatically find missing dependencies.
+
+The user simply needs to download latest database:
+    $ meson wrap update-db
+
+Once the database is available locally, any dependency not found on the system
+but available in WrapDB will automatically be downloaded.
```
jpakkane, line 94: The path `subprojects/wradb.json` file should be mentioned here, too. Having it only in release notes has poor discoverability.

xclaesse: done

xclaesse:
> There should be a way of removing the local db so that automatic resolution no longer happens

Not sure we need a specific subcommand just to `rm subprojects/wrapdb.json`. Note that `--wrap-mode=nodownload` also prevents downloading new wraps.

jpakkane:
> Not sure we need a specific subcommand

I'm not saying that we do, just that there needs to be documentation on how to do it manually.

xclaesse:
> I'm not saying that we do, just that there needs to be documentation on how to do it manually.

Ok added a sentence to make that explicit.

PR #8943, opened 2021-06-29
C2000: linker uses now binary setting from cross-file

miebka:
When using the C2000 toolchain under a non Windows-OS meson can't setup or compile the project since the linker binary is hardcoded within linkers.py. With this PR the setting from the cross-file will be used.

I'm not sure if
`super().__init__(exelist, for_machine, '', [],`
should be prefered instead of
`super().__init__(exelist or ['cl2000.exe'], for_machine, '', [],`

### DIFF
--- a/.github/workflows/lint_mypy.yml
+++ b/.github/workflows/lint_mypy.yml
@@ -23,6 +23,15 @@ jobs:
     - run: python -m pip install pylint==2.4.4
     - run: pylint mesonbuild
 
+  custom_lint:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
+    - run: python ./run_custom_lint.py
+
   mypy:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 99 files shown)


mesonbuild/linkers.py
```diff
@@ -982,9 +982,9 @@ class C2000DynamicLinker(DynamicLinker):
 
     id = 'cl2000'
 
-    def __init__(self, for_machine: mesonlib.MachineChoice,
+    def __init__(self, exelist: T.List[str], for_machine: mesonlib.MachineChoice,
                  *, version: str = 'unknown version'):
-        super().__init__(['cl2000.exe'], for_machine, '', [],
+        super().__init__(exelist or ['cl2000.exe'], for_machine, '', [],
```
dcbaker, line 987: Is there any reason this shouldn't be `exelist or ['cl2000']`, since that should work for both windows and !windows?

Otherwise this looks good to me, if you could rebase this I'd be happy to merge it..

miebka: Sure `exelist or ['cl2000']` should work for windows and non-windows OS but since this is a cross-compile setup, the compiler must be set within the crossfile and `exelist` will always be not empty, so the hard coded `cl2000.exe` isn't needed at all(?).

miebka:
Mhh, I did something wrong when trying to rebase... 
Since I am not that in in using git (still svn...) It would be easier for me to give my 2 lines of changes a new change with a fresh clean pull request.

Issue #8541, opened 2021-03-16
meson dist not including hidden folders


Tachi107:
**Describe the bug**
`meson dist` seems to exclude hidden directories from the generated tarball. In particular, it doesn't include the `.github` folder.

**To Reproduce**
- `git clone https://github.com/Tachi107/pistache.git --branch=meson`
- `cd pistache`
- `meson setup build --buildtype=release --wrap-mode=nodownload`
- `meson dist -C build --no-test`

**Expected behavior**
All source code should be included in the tarball.

**system parameters**
* Build type: native
* Operating system: Debian Testing
* Python version: 3.9.1
* Meson version: 0.57.1
* Ninja version: 1.10.1

eli-schwartz:
It's not "hidden folders".

It's "anything in the repository root that starts with `.git`". Like .gitignore, .gitmodules, .github/

This is by design, since you don't need those for dist tarballs.

Tachi107:
Thanks for the response, I thought it was some kind of bug since it wasn't documented. I agree with the fact that you don't need git files, but `.github` is not a directory created by git itself, but gets manually created by the owner of the repository, and is indeed needed for dist tarballs. For example, take a look at [Debian's packaging repo of simdjson](https://salsa.debian.org/debian/simdjson); `meson dist` would create an invalid tarball since it would exclude the `.github` directory, part of the upstream source.

eli-schwartz:
Why is this an invalid dist tarball? The contents of that directory are not needed, they only provide info to github.com -- generally, one would be excluding files like this via .gitattributes `export-ignore` (and also export-ignore'ing .circleci, .appveyor.yml, .cirrus.yml, .travis.yml, .drone.yml), but IMO the fact that this glob pattern (`.git*`) matches the .github/ directory is a happy coincidence. There are open issues for meson to support export-ignore.

Is this actually a problem for Debian packaging? Why do the salsa.debian.org sources need the github.com-specific issue templates and stuff?

Tachi107:
Yes, it is a problem because in Debian packaging the source tarball has to be identical to the upstream repository, even if not all the files are needed; if the tarball differs from the upstream source `debhelper` interprets the difference as a modification by the Debian maintainer... In short, a mess.
I agree with you with the fact that those files are useless, but stripping them automatically can cause issues, and should not be the default behavior.

eli-schwartz:
[...]

PR #8951, opened 2021-07-01
dist: only exclude actual git files

Tachi107:
`meson dist` currently excludes all files starting with `.git`. The actual special files used by Git are `.git/`, `.gitattributes`, `.gitignore` and `.gitmodules`; other files starting with `.git` are not part of revision control metadata. The current behavior can leave users surprised, mainly because it is undocumented (happened to me in #8541).

I've never touched Python, so my solution could be extremely inefficient, and maybe there is a method to avoid duplicating code as I did :)

### DIFF
--- a/mesonbuild/mdist.py
+++ b/mesonbuild/mdist.py
@@ -56,11 +56,13 @@ def create_hash(fname):
 
 
 def del_gitfiles(dirname):
+    gitfiles = ('.git', '.gitattributes', '.gitignore', '.gitmodules')
     for f in glob(os.path.join(dirname, '.git*')):
-        if os.path.isdir(f) and not os.path.islink(f):
-            windows_proof_rmtree(f)
-        else:
-            os.unlink(f)
+        if os.path.split(f)[1] in gitfiles:
+            if os.path.isdir(f) and not os.path.islink(f):
+                windows_proof_rmtree(f)
+            else:
+                os.unlink(f)
 
 def process_submodules(dirname):
     module_file = os.path.join(dirname, '.gitmodules')


mesonbuild/mdist.py
```diff
@@ -56,11 +56,12 @@ def create_hash(fname):
 
 
 def del_gitfiles(dirname):
-    for f in glob(os.path.join(dirname, '.git*')):
-        if os.path.isdir(f) and not os.path.islink(f):
-            windows_proof_rmtree(f)
-        else:
-            os.unlink(f)
+    for f in [os.path.join(dirname, '.git'), os.path.join(dirname, '.gitattributes'), os.path.join(dirname, '.gitignore'), os.path.join(dirname, '.gitmodules')]:
```
jpakkane, line 59: Rather than joining every time you should use `os.path.split()` to get the last component and then comparing that against the options.

Tachi107: You mean like:

```python
for f in glob(os.path.join(dirname, 'git*'))
  if os.path.split(f).tail in ['.git', '.gitattribute', '.gitignore', '.gitmodules']
    ...
```

(sorry I'm on my phone, not sure if `os.path.split().tail` is actually doable)

jpakkane: Basically `os.path.split(f)[1]`. You should build the array before the loop and store it in a variable, otherwise Python will recreate the object on each iteration. Since it is immutable you should use a tuple (`('foo', 'bar')`) rather than an array (`['foo', 'bar']`).

Tachi107:
Oops, messed up commit message. You can squash and merge, right?

jpakkane:
You have to fix the syntax error (your `if` and `for` statements are missing colons at the end) so you should do the squashing thing also.

Tachi107:
Ok, I'll fix it as soon as I get home. Writing patches on the go is not ideal 😅

PR #8952, opened 2021-07-01
backends/ninja: only pass project specific arguments to scan-build

dcbaker:
Currently all arguments are being passed to scan-build as part of the
refactoring of how Meson internally handles arguments, but that's wrong,
only project specific arguments are supposed to be passed.

Fixes: #8818


mcatanzaro:
I tested this and confirmed it fixes #8818. Thanks!

dcbaker:
No problem! I really should have looked at this sooner, since the regressions were in code that I'm probably the most familiar with

PR #8956, opened 2021-07-03
Fix image link names.

jpakkane:


mensinda:
Could you also add `fatal_warnings: true,` to the `meson.build`, so we catch this in the future?

mensinda:
This will also make broken links hard errors.

jpakkane:
Let's do that on a separate PR, I'm guessing it will uncover bugs so better to fix them separately.

mensinda (approved):
OK, sure.

mensinda (changes requested):
Actually, the `osx_xcode.png` image is still missing...

jpakkane:
Fixed.

mensinda (approved):

PR #8957, opened 2021-07-03
Improvements to dependency doc

jon-turney:
Some reordering to sections in dependency doc, as suggested in https://github.com/mesonbuild/meson/pull/8888#pullrequestreview-686752028, and some slight polishing while I have it open in an editor.

### DIFF
--- a/docs/markdown/Dependencies.md
+++ b/docs/markdown/Dependencies.md
@@ -355,6 +355,41 @@ additional toolkit libraries that need to be explicitly linked to.
 
 `method` may be `auto`, `config-tool`, `pkg-config`, `cmake` or `extraframework`.
 
+## Curses
+
+*(Since 0.54.0)*
+
+Curses (and ncurses) are a cross platform pain in the butt. Meson
+wraps up these dependencies in the `curses` dependency. This covers
+both `ncurses` (preferred) and other curses implementations.
+
+`method` may be `auto`, `pkg-config`, `config-tool`, or `system`.
+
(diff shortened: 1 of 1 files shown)


jon-turney:
Interestingly, looking at the code `dependency('generic', method: 'system')` fails without doing anything, rather than telling you it's nonsense (and likewise for other methods not meaningful for a generic dependency).

docs/markdown/Dependencies.md
```diff
@@ -136,23 +136,49 @@ Since they can be used interchangeably, the rest of the build
...
+and cmake, except when it is part of the base OS image (such as in FreeBSD
+and macOS); OpenGL which has pkg-config on Unices from glvnd or mesa, but has
+no pkg-config on macOS and Windows.
+
+In these cases meson provides convenience wrappers in the form of `system`
+dependencies. Internally these dependencies do exactly what a user would do
+in the build system DSL or with a script, likely calling
+`compiler.find_library()`, setting `link_with` and `include_directories`. By
+putting these in meson upstream the barrier of using them is lowered, as
+projects using meson don't have to re-implement the logic.
```
jpakkane, line 171: "Meson" is a proper name so it should be written properly capitalizes, not as "meson". There are multiple lower case spellings in this section.

jon-turney: Done

eli-schwartz (approved):
Seems a lot cleaner. :)

PR #8958, opened 2021-07-03
Enable fatal warnings in HotDoc when building website.

jpakkane:


jpakkane:
This passed even though it should not have (there is an error in wrap docs). Presumably we don't have a CI step that runs a Hotdoc check on merge requests.

eli-schwartz:
Should probably be covered by https://github.com/mesonbuild/meson/blob/746f00e69e19c712107da78424ceb84f60195564/.github/workflows/website.yml#L3-L8

Is env.HAS_SSH_KEY true in a PR? I'd assume not.

mensinda:
> Is env.HAS_SSH_KEY true in a PR?

I guess it shouldn't be. Now that I think about it anyone could just leak it by pushing a PR with `echo secrets.WEBSITE_PRIV_KEY`.

So, building the website on PRs should just involve duplicating the `push:` section for `pull_request:`.

jon-turney:
> I guess it shouldn't be. Now that I think about it anyone could just leak it by pushing a PR with `echo secrets.WEBSITE_PRIV_KEY`.

You could always make the condition

```
        if: github.event_name == 'push' && env.HAS_SSH_KEY == 'true'
```

to be sure

> So, building the website on PRs should just involve duplicating the `push:` section for `pull_request:`.

Yeah, after #8864 it probably makes sense to test building the website on every push that changes doc files.

jpakkane:
The key should most definitely not be in the container _at all_ when doing PR tests.

eli-schwartz:
I trivially rebased this.

Documentation CI is a lot better these days. Verified by reverting a commit that fixed a couple hyperlink issues, proof that this fails as it should: https://github.com/mesonbuild/meson/actions/runs/4931104057/jobs/8812791491?pr=8958

In git master this is clean.

PR #8959, opened 2021-07-03
dependencies: Deterministic LLVM compile and link arg ordering

kira78:
In LLVMDependencyConfigTool, the members compile_args and required_modules are
either converted to or stored as sets, which do not have a stable ordering. This
results in nondeterministic builds, particularly with required_modules causing
the order in which the LLVM libraries are linked in to the output binaries to
change across independent builds. As any guarantee about ordering for
compile_args is lost by being converted from a list to a set and back, and the
modules added to required_modules was even already sorted once, sort both when
converting them to lists.

### DIFF
--- a/mesonbuild/dependencies/dev.py
+++ b/mesonbuild/dependencies/dev.py
@@ -231,7 +231,7 @@ class LLVMDependencyConfigTool(ConfigToolDependency):
         self.check_components(opt_modules, required=False)
 
         cargs = set(self.get_config_value(['--cppflags'], 'compile_args'))
-        self.compile_args = list(cargs.difference(self.__cpp_blacklist))
+        self.compile_args = sorted(cargs.difference(self.__cpp_blacklist))
 
         if version_compare(self.version, '>= 3.9'):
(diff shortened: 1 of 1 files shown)


mesonbuild/dependencies/dev.py
```diff
@@ -334,7 +334,7 @@ def _set_new_link_args(self, environment: 'Environment') -> None:
 
         link_args = ['--link-static', '--system-libs'] if self.static else ['--link-shared']
         self.link_args = self.get_config_value(
-            ['--libs', '--ldflags'] + link_args + list(self.required_modules),
+            ['--libs', '--ldflags'] + link_args + sorted(self.required_modules),
```
dcbaker, line 337: It's probably easier to just store the modules using mesonlib.OrderedSet instead of pain sets, then we won't run into other issues like this elsewhere

kira78: Thank you for the suggestion! That sounds like a much better suggestion, since mesonlib includes an OrderedSet.

dcbaker (approved):

PR #8154, opened 2021-01-05
Hotdoc: use template for Commands.md instead of generating the entire file

eli-schwartz:
Fixes #8061 and is all-around cleaner IMHO.

PR #8960, opened 2021-07-04
Reference Manual 2.0

mensinda:
The main goal of this PR is to have a *machine-readable* representation of our Reference manual (in the form of multiple YAML documents) that can be used as an input for multiple generators.

Advantages:
- Everything is sorted
- Universal layout for all functions/methods
- Structural changes are (relatively) easy to do
- Different generators can be added in the future (man, LaTeX, JSON (for IDE integration), etc.)

Disadvantages:
- We now need documentation for writing documentation (which I haven't written yet)
- Less flexibility

See [my `gh-pages` website](https://mensinda.github.io/meson/RefMan.html) for the current preview of the new Reference manual. I did not convert the complete docs yet since I want to wait for that until the design is finalized.

---

Also, to build the docs [strictyaml](https://pypi.org/project/strictyaml/) and [chevron](https://pypi.org/project/chevron/) are now required. Additionally, the current master version of hotdoc is required because of https://github.com/hotdoc/hotdoc/pull/228 and https://github.com/hotdoc/hotdoc/pull/229.


tristan957:
On the `Functions` page, not a huge fan of `Function` being prepended to every function in the right-hand sidebar

tristan957:
It might make sense to combine the positional, keyword, and variadic argument sections into one arguments section.

mensinda:
Thanks for the feedback.

> On the Functions page, not a huge fan of Function being prepended to every function in the right-hand sidebar

I will look into it, but this is auto-generated by hotdoc. So I will also likely need to drop the function prefix in the heading as well.

> It might make sense to combine the positional, keyword, and variadic argument sections into one arguments section.

Makes sense.

tristan957:
I'll come back with more feedback, the more I explore. If that is an artifact of hotdoc, maybe hotdoc needs to be changed then

mensinda:
I have updated the PR with your feedback and tweaked a few things. I have also updated my gh-pages site for easy access.

See https://github.com/hotdoc/hotdoc/issues/231for the `Function` prefix issue. For now, I have decided to drop this prefix.

tristan957:
The subpages section here looks misplaced: https://mensinda.github.io/meson/RefMan.html#modules

tristan957:
The descriptions for functions are a little too large for my taste and I don't think they need the blockquote styling. https://mensinda.github.io/meson/RefMan_builtin_build_machine.html#build_machinecpu

tristan957:
The 'since $version" seems out of line: https://mensinda.github.io/meson/RefMan_elementary_str.html#signature1

tristan957:
The titles of the pages for the built-in types are awkward with the backticks. String (str) looks much better over String (\`str\`)

mensinda:
> The titles of the pages for the built-in types are awkward with the backticks.

Done.

> I am guessing this is another hotdoc issue with the intended code styling not taking effect.

More a suboptimal hotdoc documentation issue: https://github.com/hotdoc/hotdoc/pull/233

> The subpages section here looks misplaced: https://mensinda.github.io/meson/RefMan.html#modules

Fixed with a bit of CSS

> The descriptions for functions are a little too large for my taste and I don't think they need the blockquote styling. 

I have reduced the size to the default, but, to me, it looks even more awkward without the extra styling... I am happy to use another construct if someone has a better idea.

> The 'since $version" seems out of line

Fixed. Now I know remember why I don't want to be a web frontend developer...

tristan957:
> I have reduced the size to the default, but, to me, it looks even more awkward without the extra styling... I am happy to use another construct if someone has a better idea.

Do you have control over whether or not it is a `<blockquote>`

mensinda:
> Do you have control over whether or not it is a `<blockquote>`

Yes, I have full control over it [here](https://github.com/mesonbuild/meson/pull/8960/files#diff-8831c9a9d8fc49e1eef1d1edb64da074b488022bcbe6e367a312ac5425965e6bR3-R7)

tristan957:
I think a paragraph would be a much better element to use there than a blockquote. I am also not sure how blockquotes are interpreted by screen readers, so that is also something to think about.

mensinda:
Paragraphs don't render any different though (just text with a bit of spacing and that's it)

tristan957:
With blockquote I see a gray border on left hand side which makes not much sense to me given the context of the information.

mensinda:
I have mostly added more documentation (see [here](https://mensinda.github.io/meson/RefMan.html)). My goal now is to get this to a reviewable state for the complete doc.

mensinda:
I have added all the documentation from the current Reference Manual. I am considering this now as done and ready for review.

Preview: https://mensinda.github.io/meson/RefMan.html

mensinda:
One downside is that it currently requires https://github.com/hotdoc/hotdoc/pull/236, but that should resolve itself hopefully soon enough.

I am not 100% happy with how the arguments are rendered in tables, but my HTML / CSS skills are currently not good enough to come up with something better and this can now easily be changed in a follow-up PR, due to the nature of the new documentation.

jpakkane:
Why is the CMake module the only one that is shown?

I like the hierarchical layout, it's definitely more accessible that way.

mensinda:
Because it is the only module I have started to document. It is currently basically just there as a stub to show that modules can and should now also be documented there.

I didn't want to document all the modules in this PR (which is again verry big) and do that with follow up PRs since the modules are not in the old refman.

On 23 August 2021 22:39:45 CEST, Jussi Pakkanen ***@***.***> wrote:
>Why is the CMake module the only one that is shown?
>
>I like the hierarchical layout, it's definitely more accessible that way.
>
>-- 
>You are receiving this because you authored the thread.
>https://github.com/mesonbuild/meson/pull/8960#issuecomment-904108900
-- 

mensinda:
https://github.com/hotdoc/hotdoc/pull/236 is now merged and included 0.13.5, so all the new hotdoc features required for this PR are now officially released.

mensinda:
Ping :)

This PR is mostly separate from the main Meson code base and the old Reference manual is not updated frequently but it would still be nice to get this either accepted or rejected at some point.

jpakkane:
As said I like the new look. Does anyone have objections or things they'd like to see changed before merging?

xclaesse:
I haven't read the PR, only the rendered pages, but I would say go for it and we can always improve over time. The kind of big change that cannot be perfect on the first attempt.

eli-schwartz:
This has broken the "Edit on github" links in a similar way to the problem that #8154 fixed. Each page has "_build/" in the link to the sources.

I guess this will be a lot harder to fix than just "use includes rather than adding built files". The good news is, not all files have the bad links, only the generated ones which really should point to their version-controlled yaml files.

Is this something that could be handled by the extension? Do we need new hotdoc features entirely for this?

/cc @MathieuDuponchelle

eli-schwartz:
Also it feels rather awkward that the page https://mesonbuild.com/Reference-manual.html is now a dead link. Any particular reason you chose to name the new version RefMan.html instead?

mensinda:
I chose RefMan so that it would not conflict with `Reference-manual.html`. Forgot to rename it after I removed the old doc.

Regarding the `"Edit on github"`, I don't know how to fix this. I would basically need to tell hotdoc where the original YAML files are that are used to generate the `*.md` files.

eli-schwartz:
Hopefully the hotdoc developers have some advice for how to make that link work for files which are completely 100% generated/transformed from non-markdown files.

MathieuDuponchelle:
I remember back then when the same problem occurred, I mentioned I would also be opened to adding a new yaml key to pages, to allow pointing to the original source. We fixed that particular problem more elegantly IMO at the time, but here the solution (includes) will obviously not work.

The question I would have here is does it even make sense to redirect the user to those yaml files? Are they meant to be editable? If not, perhaps the correct solution is simply adding a new yaml key to pages (from the edit on github extension), that allows disabling it on a particular page.

eli-schwartz:
Currently those yaml files are hand edited and checked into git, and any person interested in correcting that part of the documentation will need to edit the yaml. It is fortunately, relatively easy to edit. Example: https://github.com/mesonbuild/meson/blob/master/docs/yaml/functions/project.yaml

My initial feeling is we do want to offer to show people where to edit this.

So I think we want the inline-yaml-frontmatter to be pointing at the yaml document sources. But I'm sure it couldn't hurt to add both options, on the theory that one day someone will probably find that very useful.

mensinda:
What could complicate things though is that multiple YAML files can (and are for functions) be merged into a single MD file.

On 5 October 2021 02:00:28 CEST, Eli Schwartz ***@***.***> wrote:
>Currently those yaml files are hand edited and checked into git, and any person interested in correcting that part of the documentation will need to edit the yaml. It is fortunately, relatively easy to edit. Example: https://github.com/mesonbuild/meson/blob/master/docs/yaml/functions/project.yaml
>
>My initial feeling is we do want to offer to show people where to edit this.
>
>So I think we want the inline-yaml-frontmatter to be pointing at the yaml document sources. But I'm sure it couldn't hurt to add both options, on the theory that one day someone will probably find that very useful.
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/mesonbuild/meson/pull/8960#issuecomment-933945717
-- 

MathieuDuponchelle:
@mensinda fwiw the route you picked means hotdoc has fewer opportunities to inject itself into the formatting of the final page as it would if the extension you wrote had created hotdoc symbols: one potential feature I thought of at some point is the addition of "edit on github" buttons next to individual symbols, for instance for GStreamer this would link to the location of the symbol's comment in the GStreamer source code, and here it could link to the exact yaml file for the function.

As you have chosen to simply generate markdown, I'm not sure how you could achieve a similar result, I'm open to suggestions though.

mensinda:
The main reason I chose a seperate step is that I want to support multiple documentation backends. For instance it would no be fairly trivial to add a JSON backend for IDE integration.

Additionally, I don't rally know how I could achieve the current results wuth hotdoc (again lack of dicumentation), so it was much easier to write a MD backend.

It would stil be possible to handle all this with a hotdoc plugin, sine there already is a [pickle dump backend](https://github.com/mesonbuild/meson/blob/master/docs/refman/main.py). So writing a plugin that uses this data should also work.

However, trying to get this to work is not very high on my todo.

On 5 October 2021 16:26:16 CEST, Mathieu Duponchelle ***@***.***> wrote:
***@***.*** fwiw the route you picked means hotdoc has fewer opportunities to inject itself into the formatting of the final page as it would if the extension you wrote had created hotdoc symbols: one potential feature I thought of at some point is the addition of "edit on github" buttons next to individual symbols, for instance for GStreamer this would link to the location of the symbol's comment in the GStreamer source code, and here it could link to the exact yaml file for the function.
>
>As you have chosen to simply generate markdown, I'm not sure how you could achieve a similar result, I'm open to suggestions though.
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/mesonbuild/meson/pull/8960#issuecomment-934462866
-- 

MathieuDuponchelle:
@mensinda I understand, just commenting, not judging, :+1: for working on this at all :)

vcunat:
The description in `docs/yaml/objects/compiler.yaml` about `static` parameter of `compiler.find_library()` clashes with what's been in release announcements (notably [0.51.0](https://mesonbuild.com/Release-notes-for-0-51-0.html#add-keyword-static-to-find_library)).  The difference is around the default behavior and `false`.

PR #8962, opened 2021-07-04
Annotate framework tests with where they are expected to skip in CI

jon-turney:
Replace the hard-coded list of 'may be skipped' framework tests in `skippable()` with annotations in `test.json` which record 'will be skipped in these specific CI jobs'.

The motivation here is that, at the moment, if something is dropped from any VM image our CI runs on, if it's permitted by `skippable()`, we'll just silently stop testing it.  This is particularly relevant where we don't control the VM image (e.g. on Windows), and has actually happened with boost on Windows (see the comments for commit 5c0cb0547bd52230faff322a33bfeac0eaaeea0d).

### DIFF
--- a/docs/markdown/Contributing.md
+++ b/docs/markdown/Contributing.md
@@ -380,6 +380,16 @@ matched:
 | `literal` | Literal match (default) |
 | `re`      | regex match             |
 
+#### skip_on_jobname
+
+The `skip_on_jobname` key contains a list of strings. If the `MESON_CI_JOBNAME`
+environment variable is set, and any of them are a sub-string of it, the test is
+expected to be skipped (that is, it is expected that the test will output
+`MESON_SKIP_TEST`, because the CI environment is not one in which it can run,
+for whatever reason).
+
+The test is failed if it skips or runs unexpectedly.
+
 ### Skipping integration tests
 
 Meson uses several continuous integration testing systems that have
--- a/docs/markdown/Contributing.md
+++ b/docs/markdown/Contributing.md
@@ -390,6 +390,14 @@ for whatever reason).
 
 The test is failed if it skips or runs unexpectedly.
 
+#### skip_on_os
+
+The `skip_on_os` key can be used to specify a list of OS names (or their
+negations, prefixed with a `!`).  If at least one item in the `skip_on_os` list
+is matched, the test is expected to be skipped.
+
+The test is failed if it skips or runs unexpectedly.
+
 ### Skipping integration tests
 
 Meson uses several continuous integration testing systems that have
--- a/docs/markdown/Contributing.md
+++ b/docs/markdown/Contributing.md
@@ -305,17 +305,15 @@ project tests with different Meson options.
 
 In the `options` dict, all possible options and their values are
 specified. Each key in the `options` dict is a Meson option. It stores
-a list of all potential values in a dict format, which allows to skip
-specific values based on the current environment.
+a list of all potential values in a dict format.
 
 Each value must contain the `val` key for the value of the option.
 `null` can be used for adding matrix entries without the current
 option.
 
-Additionally, the `skip_on_env` key can be used to specify a list of
-environment variables. If at least one environment variable in
-`skip_on_env` is present, all matrix entries containing this value are
-skipped.
+The `skip_on_env`, `skip_on_jobname` and `skip_on_os` keys (as described below)
+may be used in the value to skip that matrix entry, based on the current
+environment.
 
 Similarly, the `compilers` key can be used to define a mapping of
 compilers to languages that are required for this value.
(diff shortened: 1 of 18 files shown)


jon-turney:
As it stands, this PR results in two different meanings of 'skip' (which could therefore use some better naming):

1. Test isn't run, because we 'know' it can't work in this enviroment
2. Test is run, and is expected to output 'MESON_SKIP_TEST'

Possible future work:

- [ ] Set MESON_CI_JOBNAME in all CI jobs
- [ ] Add test.json skip annotations for those jobs
- [ ] Make it an error if CI is set, but MESON_CI_JOBNAME isn't
- [ ] Extend to test categories other than framework?

docs/markdown/Contributing.md
```diff
@@ -380,6 +378,30 @@ matched:
...
+#### skip_on_env
+
+The `skip_on_env` key can be used to specify a list of environment variables. If
+at least one environment variable in the `skip_on_env` list is present, the test
+is skipped.
+
+#### skip_on_jobname
+
+The `skip_on_jobname` key contains a list of strings. If the `MESON_CI_JOBNAME`
+environment variable is set, and any of them are contained in it, the test is
```
mensinda, line 390: This should probably make it clearer that only a substring has to be matched, because I missed that part initially.

run_project_tests.py
```diff
@@ -264,22 +264,22 @@ def __lt__(self, other: object) -> bool:
...
+compile_commands:   T.List[str] = []
+clean_commands:     T.List[str] = []
+test_commands:      T.List[str] = []
+install_commands:   T.List[str] = []
+uninstall_commands: T.List[str] = []
 
-backend:      'Backend'
-backend_flags: T.List[str]
+backend:       Backend = Backend.ninja
+backend_flags: T.List[str] = []
```
mensinda, line 282: Are these changes really required or automatically generated?

run_project_tests.py
```diff
@@ -731,7 +730,37 @@ def force_regenerate() -> None:
...
+
+    # Test is expected to skip if os matches
+    if 'skip_on_os' in test_def:
+        for skip_os in test_def['skip_on_os']:
+            if skip_os.startswith('!'):
+                if mesonenv.machines.host.system != skip_os[1:]:
+                    skip_expected = True
+            else:
+                if mesonenv.machines.host.system == skip_os:
+                    skip_expected = True
```
mensinda, line 751: Isn't using a full `Environment` way too overkill here? Alternatively, wouldn't it also be possible to create a "fake" Environment without a build dir, thus eliminating the need of an `TemporaryDirectoryWinProof`, directly here?

mensinda (commented):
Overall, I really like this change. However, I have some minor questions.

run_project_tests.py
jon-turney: The initializers are needed to prevent flake8 from reporting `F821 undefined name` (which on reflection, I guess is a bug in flake8, since the name should be defined by type-annotated name, even without an initial value)

I'm not sure the current annotation for the `backend` global is actually right:  the string `backend` isn't the type `backend`?

Maybe I should just drop the flake8 fixes commit (since we don't seem to be running flake8/sider anymore)

docs/markdown/Contributing.md
jon-turney: Yeah, I'll adjust that text.

I did initally think of making the jobnames a bit more structured (e.g. a series of '-' separated strings, and exactly matching one of those, so e.g. foo and nofoo were matchable), but that didn't seem to be needed yet.

run_project_tests.py
jon-turney: There's only a single Environment object made per detect_tests_to_run(), so it's not particularly heavyweight, and I think it's possible in future we might want to have some other tests in load_tests_json which require one.

(There's some refactoring which could be done to pass an Environment object around and down to stuff which detect_tests_to_run calls which needs an Environment, e.g. have_objc_compiler etc., which each make one at the moment.
The `with TemporaryDirectoryWinProof` then construct a `environment.Environment` stanza appears 4 times in this file now)

run_project_tests.py
mensinda: I would say that we drop the changes then.

> I'm not sure the current annotation for the backend global is actually right: the string 'backend' isn't the type backend?

The `'Backend'` is a valid type annotation. It is a string so that `Backend` does not have to be imported outside the `if T.TYPE_CHECKING:` block.

run_project_tests.py
mensinda: My main problem is that it is passed through a lot of functions just that don't do anything with it, just to be used here...

So, could you try if `environment.Environment(None, None, get_fake_options('/'))` (without a temp dir) also works, so that it can be directly constructed here? If it doesn't then I guess it is fine the way it currently is.

mensinda (approved):
LGTM and the OSX CI fails because of clock skew. Thanks for fixing/implementing this :)

PR #8969, opened 2021-07-06
Fix unsupported linker error message

Tachi107:
An `f` was dropped in bd6f46e723813dfadaba1f7c87d3b0b751d05219


eli-schwartz:
... oops?

Apparently our CI doesn't test whether error messages print the *correct* error. Who knew? :D

Issue #8765, opened 2021-05-13
`meson test -C <symlink>` broke


fiesh:
**Describe the bug**
Running `meson setup <symlink>` works properly.  Running `meson test -C <symlink>` used to work properly but was broken by 79e2c52a15e896e46ff3cfa3ec16fbf3f132ee01

**To Reproduce**

`i.c`:
```c
int main() {}
```

`meson.build`:
```meson
project('reproduce_sym_link_test_issue', 'c')
files = files('i.c')
main = executable('i', files)
test('i', main)
```

Then:
```zsh
% mkdir build
% ln -s build sym_link
% meson setup sym_link
<...>
% meson test -C sym_link
ninja: Entering directory `/tmp/r/sym_link'
ninja: error: unknown target '../build/i'
Could not rebuild /tmp/r/sym_link
% meson test -C build
ninja: Entering directory `/tmp/r/build'
[2/2] Linking target i
1/1 i        OK              0.01s


Ok:                 1   
Expected Fail:      0   
Fail:               0   
Unexpected Pass:    0   
Skipped:            0   
Timeout:            0   

Full log written to /tmp/r/build/meson-logs/testlog.txt
```

**Expected behavior**
No regression to before.

**system parameters**
* No cross build
* Gentoo Linux
* 3.8.9
* Any version after the regression commit
* ninja 1.10.1

PR #8970, opened 2021-07-07
Patches pending for 0.58.2 milestone

bonzini:
Since I would like to use a 0.58 release with #8900, maybe I can help by doing the backports in the [0.58.2 milestone](https://github.com/mesonbuild/meson/milestone/63?closed=1) myself. :)

PR #8972, opened 2021-07-07
resolve symlinks passed to -C

bonzini:
`meson setup` is resolving symlinks for the build directory in
`validate_core_dirs`.  For consistency with it, do the same when
the build directory is passed via -C to devenv, dist, init, install
and test.
    
This ensures for example that the path to test dependencies is
computed correctly in `meson test`.
    
Fixes: #8765


mensinda:
This should come with a test case, otherwise LTGM.

bonzini:
All failures seem independent (mpi, SubprojectsCommandTests.test_purge, and coverage should have gone up rather than down.

mensinda (approved):
LGTM, see the discussion in #8970 for merging.

nirbheek:
FTR: This is also in the 0.58.2 release, but the milestone only shows 0.59.1 because it was merged after 0.59.0 and will be in that stable release too.

PR #8973, opened 2021-07-08
mconf: add --fatal-meson-warnings (just as msetup)

### LINKED ISSUE #7288: Instead of just printing a warning, Meson should abort and exit with an error code if an unknown build option is used

Flowdalic:

I am not sure why meason decided that this should only be a
WARNING. But users should at least be able to opt-in for a fail-safe
behavior. This is now possible via the --fatal-meson-warnings
switch for the 'configure' subcommand.

Partially fixes #7288.

### DIFF
--- a/docs/markdown/snippets/fatal-meson-warnings.md
+++ b/docs/markdown/snippets/fatal-meson-warnings.md
@@ -1,0 +1,8 @@
+## `--fatal-meson-warnings` added to 'configure' task
+
+There was previously no way to enable fatal Meson warnings for the
+'configure' task. Hence it was easy to miss that an unknown option,
+e.g. caused by a typo, was encountered, since only a warning was
+emitted while Meson would exit without an error code. If
+`--fatal-meson-warnings` is now given, the 'configure' task will exit
+with an error code if an unknown Meson option was encountered.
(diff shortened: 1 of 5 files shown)


Flowdalic:
This supersedes #8658. And I would love to write a test for, but I am not sure how to write a test that calls meson with --fatal-meson-warnings.

pinkflames:
Does --fatal-meson-warnings only affect configure or would it affect other stages as well? I'm unsure how safe it is to make all warnings fatal for configure but I really can't imagine distros being able to make them fatal for other stages, too - that seems like asking for a lot of maintenance pain, since I can make a build system output a warning in configure as easily as calling warning(), which means it's not a question of if but when something will fail for a silly reason requiring either patching the build system or disabling the --fatal-meson-warnings option, so I'd really prefer the original idea of --error-on-unknown-options or as suggested by @jpakkane - make it a default with a way to opt-out of the unknown options error.

bonzini:
#7288 has been closed and unknown options are fatal, the PR is not needed.

PR #8974, opened 2021-07-08
coredata: throw a MesonException on unknown options

### LINKED ISSUE #7288: Instead of just printing a warning, Meson should abort and exit with an error code if an unknown build option is used

Flowdalic:
Fixes #7288.

### DIFF
--- a/docs/markdown/snippets/unknown-options-always-fatal.md
+++ b/docs/markdown/snippets/unknown-options-always-fatal.md
@@ -1,0 +1,5 @@
+## Unknown options are now always fatal
+
+Passing unknown options to "meson setup" or "meson configure" is now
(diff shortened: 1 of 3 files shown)


Flowdalic:
I am not sure why this change has such an impact on the code coverage. It is from side ready for a review.

dcbaker:
This seems really strange to me. Why is it only throwing of the machine isn't the build machine

PR #8975, opened 2021-07-08
coverage: prefer gcovr over lcov for html

williamspatrick:


mensinda:
This should come with a unit test since this is (at least a partial) bug fix. LGTM otherwise.

PR #8979, opened 2021-07-09
azure: propagate run_tests.py exit code

jon-turney:
Propagate the run_tests.py exit code, rather than overwriting it with the codecov exit code.


jon-turney:
Since #8894, it seems we ignoring if tests fail on azure, which is probably not what's wanted... :)

mensinda:
Yeah, that's definitely not what I intended to do :)

PR #8982, opened 2021-07-09
Windows: setup vsenv when clang-cl is present too

Sahnvour:


jpakkane:
Originally we did this but it was changed because it reportedly caused problems for some.

Sahnvour:
Do you know which problems ? I remember you mentionning CI issues in #8748, I wanted to try again and it seems OK now.

Sahnvour:
@jpakkane I saw that unfortunately this didn't make it for 0.59.

Could you expand on the issues encountered ? I would argue having `clang-cl` in path should be considered a default development environment, and that maybe it's their setup that is problematic.
I used a similar trick (only in bash) for a long time over multiple VS/msvc versions with clang-cl without issues.

jrobeson:
is there some way to make it work if gcc is present as well?

jrobeson:
This seems to be closer to the acceptable solution: https://github.com/mesonbuild/meson/pull/9306

PR #8989, opened 2021-07-13
intl dep: support static kwarg

eli-schwartz:
/cc @lazka who asked for this

### DIFF
--- a/mesonbuild/dependencies/base.py
+++ b/mesonbuild/dependencies/base.py
@@ -22,7 +22,7 @@ from enum import Enum
 
 from .. import mlog
 from ..compilers import clib_langs
-from ..mesonlib import MachineChoice, MesonException, HoldableObject
+from ..mesonlib import LibType, MachineChoice, MesonException, HoldableObject
 from ..mesonlib import version_compare_many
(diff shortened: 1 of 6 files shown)


lazka:
I can confirm that this works for my project. And the changes also lgtm, for what it's worth.

dcbaker (approved):
This all looks good to me

jpakkane:
The first commit should probably not be there...

dcbaker:
That's already merged and should rebase out, right?

jpakkane:
FWICT yes.

jpakkane:
Also this :crossed_swords: conflicts now.

eli-schwartz:
Right, the first commit was there because I rebased this on top of the codecov.yml PR in order to test out the latter. Rebased to fix conflicts...

eli-schwartz:
Oops, I was too hasty to merge this. It passed CI before the rebase... but now there is `[UNEXSKIP]   frameworks: 6 gettext    (static=true)` on macOS, which I guess isn't shocking since I had investigated the static case for msys2. But I do wonder why it reports skipped instead of failed.

dcbaker:
cvan't find intl maybe?

PR #8990, opened 2021-07-13
Clang: Apply `-O0`

### LINKED ISSUE #8986: Seems like `optimization=0` does nothing
**Describe the bug**

It seems that `optimization=0` does not have any effect.
i.e. `-O0` does not appear in the compiling command when using `optimization=0`.

**To Reproduce**

With the following code:
```C
// meson.build
project('hello', 'c')
executable('hello', 'hello.c')

// hello.c
#include <stdio.h>
int main(void) {
	puts("Hello world");
	return 0;
}
```

And using the following commands:
```C
meson setup buildir -Doptimization=0
cd buildir
meson compile -v
```

I obtain:
```
ninja: Entering directory `.'
[1/2] gcc -Ihello.p -I. -I.. -fdiagnostics-color=always -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -g -MD -MQ hello.p/hello.c.o -MF hello.p/hello.c.o.d -o hello.p/hello.c.o -c ../hello.c
[2/2] gcc  -o hello hello.p/hello.c.o -Wl,--as-needed -Wl,--no-undefined
```

**Expected behavior**

I expected to see `-O0` to be passed as an option to GCC.

**system parameters**
* Is this a [cross build](https://mesonbuild.com/Cross-compilation.html) or just a plain native build (for the same computer)?: **native**
* what operating system (e.g. MacOS Catalina, Windows 10, CentOS 8.0, Ubuntu 18.04, etc.): **NixOS**
* what `meson --version`: **0.57.1**
* what `ninja --version` if it's a Ninja build: **1.10.2**

Pamplemousse:
Fix #8986

**Motivation for this change**

Calling `clang` without any `-O` option is **NOT** equivalent to calling it with `-O0`.
One important difference is that when debugging, without `-O0` (and even with `-Og`), certain parameter values are marked as `<optimized out>` in GDB;
On the contrary, using `-O0` allows to get the value of any variable.

### DIFF
--- a/mesonbuild/compilers/mixins/clang.py
+++ b/mesonbuild/compilers/mixins/clang.py
@@ -35,7 +35,7 @@ clang_color_args = {
 }  # type: T.Dict[str, T.List[str]]
 
 clang_optimization_args = {
-    '0': [],
+    '0': ['-O0'],
     'g': ['-Og'],
     '1': ['-O1'],
     '2': ['-O2'],


dcbaker (approved):
LGTM

dcbaker:
Not a big deal, but we don't use DSO for Meson, so please don't add a `signed-off-by`, as it has no meaning for the project. But, if in the future we did begin using `signed-off-by` for something (whether DSO or something else), you're `signed-off-by` might me misinterpreted as something it's not. This is true for any project that doesn't have a defined use for `signed-off-by`s

jpakkane:
This was improperly merged. Trunk is currently frozen for release and only regression fixes are allowed in and this does not qualify as one. Please revert from master and recreate.

dcbaker:
This is a bug fix, even if it's not a regression. I thought bug fixes were allowed even during trunk freezes?

jpakkane:
Well it's arguable whether it is a bug fix or a behavioural change (or both). I have gotten paranoid about changes in behaviour close to the release since they seem to end up breaking someone no matter how innocent they seem. I guess we can keep this but from now on only notable bug fixes should go in trunk Hopefully we'll have a release on Sunday and can merge all the outstanding stuff.

dcbaker:
Okay, I thought this was more important since clang and gcc have different default behavior, gcc defaults (IIRC) to `-O0`, while clang defaults to something that is not any of the `-O*` options, so setting `optimization=0` would, without this, have different behavior between clang and gcc. I'll refrain from merging anything else that isn't a regression fix until 0.59 is out

PR #8992, opened 2021-07-13
Cleanup the python module

dcbaker:
I wanted to get us down to one set of python dependency classes, but that's still not achieved here. What is achieved is a complete set of type annotations for the python module (there's still some places where things fail because we need more typing in the interpreter), the use of the `typed_*` decorators where applicable, and the conversion of the mega `PythonDependency` to a dependency factory. The use of a factory allows us to drop a bunch of legacy cruft that had been mostly purged, but was sitll around because of this one dependency.

@xclaesse I dropped the test that checked the pkg_config.pcdep test, but I'm not sure if you were specifically trying to test `.pcdep` or if that was just the method you used to test the module itself. If the latter I can rewrite the test instead.

### DIFF
--- a/mesonbuild/modules/python.py
+++ b/mesonbuild/modules/python.py
@@ -13,8 +13,9 @@
 # limitations under the License.
 
 from pathlib import Path
-import os
+import functools
 import json
+import os
 import shutil
 import typing as T
 
@@ -23,6 +24,7 @@ from .. import mesonlib
 from .. import mlog
 from ..build import known_shmod_kwargs
 from ..dependencies import DependencyMethods, PkgConfigDependency, NotFoundDependency, SystemDependency
+from ..dependencies.base import process_method_kw
 from ..environment import detect_cpu_family
 from ..interpreter import ExternalProgramHolder, extract_required_kwarg, permitted_dependency_kwargs
 from ..interpreterbase import (
@@ -30,13 +32,14 @@ from ..interpreterbase import (
     InvalidArguments, typed_pos_args, typed_kwargs, KwargInfo,
     FeatureNew, FeatureNewKwargs, disablerIfNotFound
 )
-from ..mesonlib import MachineChoice, MesonException
+from ..mesonlib import MachineChoice
 from ..programs import ExternalProgram, NonExistingExternalProgram
 
 if T.TYPE_CHECKING:
     from . import ModuleState
     from ..build import SharedModule, Data
-    from ..dependencies import ExternalDependency
+    from ..dependencies import ExternalDependency, Dependency
+    from ..dependencies.factory import DependencyGenerator
     from ..environment import Environment
     from ..interpreter import Interpreter
     from ..interpreterbase.interpreterbase import TYPE_var, TYPE_kwargs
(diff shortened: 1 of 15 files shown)


mesonbuild/modules/python.py
```diff
@@ -498,32 +529,41 @@ def _get_win_pythonpath(self, name_or_path):
...
 
-    @FeatureNewKwargs('python.find_installation', '0.49.0', ['disabler'])
-    @FeatureNewKwargs('python.find_installation', '0.51.0', ['modules'])
     @disablerIfNotFound
-    @permittedKwargs({'required', 'modules'})
-    def find_installation(self, state, args, kwargs):
+    @typed_pos_args('python.find_installation', optargs=[str])
+    @typed_kwargs(
+        'python.find_installation',
+        KwargInfo('required', bool, default=True),
```
eli-schwartz, line 544:
```
test cases/unit/39 python extmodule/meson.build:25:0: ERROR: python.find_installation keyword argument "required"" was of type "UserFeatureOption" but should have been "bool"
```

mesonbuild/modules/python.py
```diff
@@ -354,136 +409,113 @@ def extension_module_method(self, args, kwargs):
...
-            raise InvalidArguments('get_path must have one or two arguments.')
-        path_name = args[0]
-        if not isinstance(path_name, str):
-            raise InvalidArguments('get_path argument must be a string.')
-
+    def get_path_method(self, args: T.Tuple[str, T.Optional['TYPE_var']], kwargs: 'TYPE_kwargs') -> 'TYPE_var':
+        path_name, fallback = args
         try:
-            path = self.paths[path_name]
+           return self.paths[path_name]
```
eli-schwartz, line 469: Strangely indented with 3 spaces, not four :o

dcbaker: I blame mesa and it's 3 space indent :)

mesonbuild/modules/python.py
```diff
@@ -12,41 +12,59 @@
...
-        self.name = 'python'
-        self.static = kwargs.get('static', False)
-        self.embed = kwargs.get('embed', False)
-        self.version = python_holder.version
+if T.TYPE_CHECKING:
+    _Base = ExternalDependency
+else:
+    _Base = object
+
+class _PythonDependencyBase(_Base):
```
mensinda, line 61: Is there a specific reason for doing this (cyclic imports?) or is this just for improving imports?

mensinda (commented):
Is there a specific reason why we still need `class DependencyMethods(Enum)` in `base.py` or can we get rid of it completely?

I really can't comment on the pcdep stuff since I have no idea what it is/was used for...

mesonbuild/modules/python.py
dcbaker: It helps mypy, which doesn't understand these kind of mixing classes at all. So we get two choices, forward declare a bunch of methods and attributes in the mixing class and try to keep them in sync, or this, which has mypy understand this as an ExternalProgram, but at runtime avoids the two bases problem.

dcbaker: I use the same trick in the compilers class.

Issue #1181, opened 2016-12-10
gnome: g-ir-scanner scan build error


tp-m:
I have a feeling this has been discussed somewhere already, but I can't find it right now.

So here's what happened when I did `git pull && ninja -C build` on a git checkout of gstreamer with a meson build from a few months ago:

```
[1/346] 'Generating Gst-1.0.gir with a custom command.'
FAILED: gst/Gst-1.0.gir 
'/usr/bin/g-ir-scanner' ../gst/lots-of-c-and-h-files '-pthread' '-I/usr/include/gobject-introspection-1.0' '-I/usr/include/glib-2.0' '-I/usr/lib/x86_64-linux-gnu/glib-2.0/include' '--no-libtool' '--namespace=Gst' '--nsversion=1.0' '--warn-all' '--output' 'gst/Gst-1.0.gir' '--add-init-section=extern void gst_init(gint*,gchar**);g_setenv("GST_REGISTRY_1.0", "/no/way/this/exists.reg", TRUE);g_setenv("GST_PLUGIN_PATH_1_0", "", TRUE);g_setenv("GST_PLUGIN_SYSTEM_PATH_1_0", "", TRUE);gst_init(NULL,NULL);' '--c-include=gst/gst.h' '-I/home/tpm/gst-meson/gstreamer/gst' '-I/home/tpm/gst-meson/gstreamer/build/gst' '-I/home/tpm/gst-meson/gstreamer/build/.' '-I/home/tpm/gst-meson/gstreamer/.' '-I/home/tpm/gst-meson/gstreamer/build/gst/parse' '-I/home/tpm/gst-meson/gstreamer/gst/parse' '--include=GLib-2.0' '--include=GObject-2.0' '--include=GModule-2.0' '--symbol-prefix=gst' '--identifier-prefix=Gst' '--pkg-export=gstreamer-1.0' '.' '-pthread' '-I/usr/include/glib-2.0' '-I/usr/lib/x86_64-linux-gnu/glib-2.0/include' 'parse' '-L/home/tpm/gst-meson/gstreamer/build/gst' '-lunwind' '-lgmodule-2.0' '-pthread' '-lgobject-2.0' '-lm' '-lglib-2.0' '--library' 'gstreamer-1.0'
[...]

PR #8993, opened 2021-07-14
gnome: Always pass absolute -L paths to g-ir-scanner

nirbheek:
g-ir-scanner does not convert relative -L paths to runtime paths which
are added to `-Wl,-rpath` and `LD_LIBRARY_PATH` / `DYLD_LIBRARY_PATH`
/ `PATH`. This means that the local library will either not be found at
runtime (while building introspection data), or the system-wide
library will be picked instead.

See: `giscanner/ccompiler.py:get_internal_link_flags()` in
gobject-introspection for more details.

This is a regression likely from the time when we moved `@BUILD_ROOT@` from an absolute path to a relative path; the oldest known working version of Meson is 0.55. Both 0.58.1 and 0.59.0rc2 are broken.

### DIFF
--- a/mesonbuild/modules/gnome.py
+++ b/mesonbuild/modules/gnome.py
@@ -640,14 +640,18 @@ class GnomeModule(ExtensionModule):
                 # Because of https://gitlab.gnome.org/GNOME/gobject-introspection/merge_requests/72
                 # we can't use the full path until this is merged.
                 libpath = os.path.join(girtarget.get_subdir(), girtarget.get_filename())
+                # Must use absolute paths here because g-ir-scanner will not
+                # add them to the runtime path list if they're relative. This
+                # means we cannot use @BUILD_ROOT@
+                build_root = state.environment.get_build_dir()
                 if isinstance(girtarget, build.SharedLibrary):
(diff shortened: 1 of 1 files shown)


nirbheek:
The codecov numbers look... wrong?

nirbheek:
This PR is yet another whackamole fix for https://github.com/mesonbuild/meson/issues/1181

dcbaker (approved):
This looks like the right solution to me.

dcbaker:
@jpakkane I think this could go in before the release.

dcbaker:
@nirbheek It would be nice ot have a test for this so it doesn't silently get broken again

nirbheek:
> @nirbheek It would be nice ot have a test for this so it doesn't silently get broken again

I tried to make a test case for this, but it was surprisingly difficult. I think it's some combination of ld.so and ld rules for system libraries. You need to be building a gir for a library that is also available on the system + some other factors that I am not sure of.

nirbheek:
FTR: This is also in the 0.58.2 release, but the milestone only shows 0.59.1 because it was merged after 0.59.0 and will be in that stable release too.

PR #8994, opened 2021-07-14
Cuda fixes

SoapGentoo:
The `-isystem` fix:

### Before:
```sh
ninja: Entering directory `build/'
[1/2] ccache nvcc -Iprog.p -Xcompiler=-Wall,-Winvalid-pch,-Wnon-virtual-dtor --std=c++14 -g -G -g -Xcompiler=-isystem/usr/include/ModemManager -Xcompiler=-isystem/usr/include -DBOOST_ALL_NO_LIB -I.. -I. -Iprog.p -o prog.p/main.cu.o -c ../main.cu
FAILED: prog.p/main.cu.o 
ccache nvcc -Iprog.p -Xcompiler=-Wall,-Winvalid-pch,-Wnon-virtual-dtor --std=c++14 -g -G -g -Xcompiler=-isystem/usr/include/ModemManager -Xcompiler=-isystem/usr/include -DBOOST_ALL_NO_LIB -I.. -I. -Iprog.p -o prog.p/main.cu.o -c ../main.cu
In file included from /opt/cuda/bin/../targets/x86_64-linux/include/crt/math_functions.h:8958,
                 from /opt/cuda/bin/../targets/x86_64-linux/include/crt/common_functions.h:303,
                 from /opt/cuda/bin/../targets/x86_64-linux/include/cuda_runtime.h:115,
                 from <command-line>:
/usr/lib/gcc/x86_64-pc-linux-gnu/10.3.0/include/g++-v10/cmath:45:15: fatal error: math.h: No such file or directory
   45 | #include_next <math.h>
      |               ^~~~~~~~
compilation terminated.
```

### After:
```sh
ninja: Entering directory `build/'
[1/2] ccache nvcc -Iprog.p -Xcompiler=-Wall,-Winvalid-pch,-Wnon-virtual-dtor --std=c++14 -g -G -g -isystem /usr/include/ModemManager -DBOOST_ALL_NO_LIB -I.. -I. -Iprog.p -o prog.p/main.cu.o -c ../main.cu
[2/2] nvcc  -o prog prog.p/main.cu.o -lstdc++
```

### DIFF
--- a/mesonbuild/compilers/cuda.py
+++ b/mesonbuild/compilers/cuda.py
@@ -759,3 +759,6 @@ class CudaCompiler(Compiler):
             return [self._shield_nvcc_list_arg('-ccbin='+ccbindir, False)]
         else:
             return []
+
+    def get_disable_assert_args(self) -> T.List[str]:
+        return self.host_compiler.get_disable_assert_args()
--- a/mesonbuild/compilers/cuda.py
+++ b/mesonbuild/compilers/cuda.py
@@ -767,5 +767,11 @@ class CudaCompiler(Compiler):
         else:
             return []
 
+    def get_profile_generate_args(self) -> T.List[str]:
+        return ['-Xcompiler=' + x for x in self.host_compiler.get_profile_generate_args()]
+
+    def get_profile_use_args(self) -> T.List[str]:
+        return ['-Xcompiler=' + x for x in self.host_compiler.get_profile_use_args()]
+
     def get_disable_assert_args(self) -> T.List[str]:
         return self.host_compiler.get_disable_assert_args()
--- a/mesonbuild/compilers/cuda.py
+++ b/mesonbuild/compilers/cuda.py
@@ -290,8 +290,7 @@ class CudaCompiler(Compiler):
                 raise ValueError("-Xcompiler flag merging failed, unknown argument form!")
         return xflags
 
-    @classmethod
-    def _to_host_flags(cls, flags: T.List[str], phase: _Phase = _Phase.COMPILER) -> T.List[str]:
+    def _to_host_flags(self, flags: T.List[str], phase: _Phase = _Phase.COMPILER) -> T.List[str]:
         """
         Translate generic "GCC-speak" plus particular "NVCC-speak" flags to NVCC flags.
 
@@ -353,7 +352,7 @@ class CudaCompiler(Compiler):
             # an exception for -D (where this would be value-changing) and -U (because
             # it isn't possible to define a macro with a comma in the name).
 
-            if flag in cls._FLAG_PASSTHRU_NOARGS:
+            if flag in self._FLAG_PASSTHRU_NOARGS:
                 xflags.append(flag)
                 continue
 
@@ -384,19 +383,23 @@ class CudaCompiler(Compiler):
                 else:                            # -Isomething
                     val = flag[2:]
                 flag = flag[:2]                  # -I
-            elif flag in cls._FLAG_LONG2SHORT_WITHARGS or \
-                 flag in cls._FLAG_SHORT2LONG_WITHARGS:
+            elif flag in self._FLAG_LONG2SHORT_WITHARGS or \
+                 flag in self._FLAG_SHORT2LONG_WITHARGS:
                 # This is either -o or a multi-letter flag, and it is receiving its
                 # value isolated.
                 try:
                     val = next(flagit)           # -o something
                 except StopIteration:
                     pass
-            elif flag.split('=',1)[0] in cls._FLAG_LONG2SHORT_WITHARGS or \
-                 flag.split('=',1)[0] in cls._FLAG_SHORT2LONG_WITHARGS:
+            elif flag.split('=',1)[0] in self._FLAG_LONG2SHORT_WITHARGS or \
+                 flag.split('=',1)[0] in self._FLAG_SHORT2LONG_WITHARGS:
                 # This is either -o or a multi-letter flag, and it is receiving its
                 # value after an = sign.
                 flag, val = flag.split('=',1)    # -o=something
+            # Some dependencies (e.g., BoostDependency) add unspaced "-isystem/usr/include" arguments
+            elif flag.startswith('-isystem'):
+                val = flag[8:].strip()
+                flag = flag[:8]
             else:
                 # This is a flag, and it's foreign to NVCC.
                 #
@@ -418,7 +421,7 @@ class CudaCompiler(Compiler):
                     xflags.append('-prec-div=true')
                     xflags.append('-Xcompiler='+flag)
                 else:
-                    xflags.append('-Xcompiler='+cls._shield_nvcc_list_arg(flag))
+                    xflags.append('-Xcompiler='+self._shield_nvcc_list_arg(flag))
                     # The above should securely handle GCC's -Wl, -Wa, -Wp, arguments.
                 continue
 
@@ -427,7 +430,7 @@ class CudaCompiler(Compiler):
 
 
             # Take care of the various NVCC-supported flags that need special handling.
-            flag = cls._FLAG_LONG2SHORT_WITHARGS.get(flag,flag)
+            flag = self._FLAG_LONG2SHORT_WITHARGS.get(flag,flag)
 
             if   flag in {'-include','-isystem','-I','-L','-l'}:
                 # These flags are known to GCC, but list-valued in NVCC. They potentially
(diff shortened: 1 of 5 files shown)


mesonbuild/compilers/cuda.py
```diff
@@ -384,16 +383,19 @@ def _to_host_flags(cls, flags: T.List[str], phase: _Phase = _Phase.COMPILER) ->
                 else:                            # -Isomething
                     val = flag[2:]
                 flag = flag[:2]                  # -I
-            elif flag in cls._FLAG_LONG2SHORT_WITHARGS or \
-                 flag in cls._FLAG_SHORT2LONG_WITHARGS:
+            elif flag.startswith('-include') or flag.startswith('-isystem'):
+                val = flag[8:].strip()
+                flag = flag[:8]                  # -include/-isystem
```
eli-schwartz, line 388: Trying to read this, I found it hard to realize you're special casing this because `-isystem /usr/include` is handled by "normal option parsing", but the quite odd variant `-isystem/usr/include` wouldn't be detected properly.

Could you comment this? The current comment is really quite useless, the startswith check tells us the "what" already, but an actual description of "why" would be nice.

SoapGentoo: both `-isystem /usr/include` and `-isystem/usr/include` get caught by this code.

So you want me to comment here instead of later? I have a comment for treating `-isystem` and catching `/usr/include` later when the actual processing/filtering happens

obilaniu: @SoapGentoo @eli-schwartz I am the author of this file and of the arguments parsing, for good or ill.

I believe I understand what my screw-up was here. The difference appears to be that in NVCC, `-isystem` is a list option and is documented as requiring that it be separated from its value by either a space or an `=` sign. Therefore, `['-isystemfoobar']` (perhaps constructed from string concatenation) is an illegal argument. However, GCC appears to accept it just fine (!) and interprets it the same as `-isystem foobar`. The relative path `foobar` doesn't even need to exist apparently. I falsely assumed GCC would reject it as well for similar reasons due to potential ambiguity (in its `man`page, the two following flags are `-iwithprefix dir` and `-iwithprefixbefore dir`, which exhibit precisely that).

I think there is something subtly wrong in the patch as proposed. Because this special-case is found *before* the `elif flag in self._FLAG_LONG2SHORT_WITHARGS or flag in self._FLAG_SHORT2LONG_WITHARGS:` test, it will capture `-isystem` and `-include` with value concatenated (`-isystemfoobar`), separated (`-isystem=foobar`) *and* isolated (`-isystem foobar`), but in the latter case it will compute `'-isystem'[8:] == ''` as the value and then consider `foobar` a standalone argument.

If this could be fixed and a comment explaining the foregoing added, I'd be very happy.

obilaniu: Ah yes, and to make things even more complicated, it turns out that GCC considers that if the first character of the directory argument of `-I -iquote -isystem -dirafter` is `=`, then the path is assumed to be relative to the sysroot prefix. That introduces an ambiguity I did not heretofore think about when writing this code, as it is impossible for me to guess what `-I=foobar` should be understood as (GCC-style sysroot include, or NVCC-style regular include?). How painful.

```
       -I dir
       -iquote dir
       -isystem dir
       -idirafter dir
           Add the directory dir to the list of directories to be searched for header files during preprocessing.  If dir
           begins with = or $SYSROOT, then the = or $SYSROOT is replaced by the sysroot prefix; see --sysroot and -isysroot.
```

eli-schwartz: I'd assume if nvcc successfully parses it as an option at all, it reinterprets it... so err on the side of NVCC style parsing.

And yes, I don't know why GCC supports it without spaces but if you look through the meson codebase you'll see we use both forms.

(e.g. the generic clike mixin defines get_include_args as `['-isystem', '/path']` while gnu defines it to `['-isystem/path']`)

obilaniu:
I think it would perhaps be worth it here to begin a discussion on what precisely the Meson interface to NVCC, and the function [`_to_host_flags()`](https://github.com/mesonbuild/meson/blob/e8c901048508370ee602e91cc413feb7de7a706a/mesonbuild/compilers/cuda.py#L293-L299), ought to be.

The output of this function must definitely be flags acceptable to NVCC. However, at present there is fuzziness in what its _input_ is supposed to be.

As my code comments hint at, NVCC introduces new flags, renames others and subtly redefines the semantics of yet other flags relative to GCC. My argument-parsing attempts to accept both NVCC-speak and GCC-speak and translate them correctly, but there's ambiguity in many cases. For instance, [these two lines](https://github.com/mesonbuild/meson/blob/e8c901048508370ee602e91cc413feb7de7a706a/mesonbuild/compilers/cuda.py#L443-L444) currently make Meson take the GCC interpretation of `-I` flags (`['-Ihave,commas']`) by suitably shielding the argument for NVCC. But if Meson did not do this, then it would have been accepting the NVCC interpretation (which is equivalent to `['-Ihave', '-Icommas']`). I believe I chose the GCC interpretation because it allowed making use of forwarded `self.host_compiler.*_args()` methods, but I nevertheless accept a wide range of flags totally unique to NVCC because they lack equivalents in GCC. There may be surprises lying in wait for the user.

How closely should Meson tack to one side or the other? Should there be a defined set of arguments (`-I`, `-l`, `-L`, `-isystem`, ...) for which we declare upfront we will always take the GCC interpretation? But then, how much effort can we spend papering over NVCC's idiosyncrasies in order to pretend it is a GCC with extra GPU flags? (The dep-generation flags `-M*` are notoriously limited in NVCC, for instance).

mesonbuild/compilers/cuda.py
SoapGentoo: @obilaniu are you on IRC? discussing this on IRC seems a lot more promising than here.

obilaniu: @SoapGentoo Ay, no I have no account there. Anything in particular you wanted to ask about?

SoapGentoo:
I'd go with the GCC syntax if in doubt. What specifically do you want me to do? Can you describe the input algorithm as you'd like to have it?

obilaniu:
@SoapGentoo The only change I'm arguing for in your patch is that the checks should proceed in this order:
```
            elif flag in cls._FLAG_LONG2SHORT_WITHARGS or \
                 flag in cls._FLAG_SHORT2LONG_WITHARGS:
                ...
            elif flag.startswith('-include') or flag.startswith('-isystem'):
                ...
```
and not the reverse. This allows the first check to handle the `-isystem isolated` and `-isystem=separated` cases, which is already done correctly, and have the new `elif` handle only `-isystemconcatenated`, which (as currently written) will *not* handle `-isystem isolated` correctly.

SoapGentoo:
@obilaniu changed, good now?

obilaniu:
@SoapGentoo Looks OK! Nitpick: Further down, I find
```python
                elif flag != '-isystem' or val not in self.host_compiler.get_default_include_dirs():
                    xflags.append(flag)
                    xflags.append(self._shield_nvcc_list_arg(val))
```
I find this condition hard to interpret and prone to bugs should the `elif` be extended with an `else` or more `elif`s. I'd rather phrase it in the positive, redundant sense:
```python
                elif flag == '-isystem' and val in self.host_compiler.get_default_include_dirs():
                    pass
                else:
                    xflags.append(flag)
                    xflags.append(self._shield_nvcc_list_arg(val))
```
Otherwise, great!

mesonbuild/compilers/cuda.py
```diff
@@ -384,19 +383,23 @@ def _to_host_flags(cls, flags: T.List[str], phase: _Phase = _Phase.COMPILER) ->
...
                     pass
-            elif flag.split('=',1)[0] in cls._FLAG_LONG2SHORT_WITHARGS or \
-                 flag.split('=',1)[0] in cls._FLAG_SHORT2LONG_WITHARGS:
+            elif flag.split('=',1)[0] in self._FLAG_LONG2SHORT_WITHARGS or \
+                 flag.split('=',1)[0] in self._FLAG_SHORT2LONG_WITHARGS:
                 # This is either -o or a multi-letter flag, and it is receiving its
                 # value after an = sign.
                 flag, val = flag.split('=',1)    # -o=something
+            # Some dependencies (e.g., BoostDependency) add unspaced "-isystem/usr/include" arguments
+            elif flag.startswith('-isystem'):
```
SoapGentoo, line 400: @nirbheek this finally fixes the issue from https://github.com/mesonbuild/meson/pull/7418#discussion_r452087151 properly

jpakkane:
Is everyone happy with this now?

Issue #8746, opened 2021-05-08
shared_module() builds a module with SONAME field while it is not a shared_library()


MichaelJTokarev:
For an ELF targets (eg gcc on linux), shared_module() meson function builds a module with SONAME field (using -Wl,-soname argument). This is wrong, - only the shared_library() needs SONAME, while shared_module() does not. More, some tools - e.g. debian's dpkg-shlibdeps - uses presence of SONAME field as an indicator that this is shared _library_ (as opposed to shared module - for the module it is okay to have unresolved symbols which are imported from the executable which loads the module, while a library should have all symbols resolved).

It'd be very nice if meson will not use -soname when building shared_module(), -soname should be used only for shared_library().

dcbaker:
We should only be setting soname if an explicit version is set, at the very least. So I'm calling this a bug

PR #8996, opened 2021-07-14
Do not add SONAME to shared modules

bonzini:
For an ELF targets, shared_module() builds a module with SONAME field (using -Wl,-soname argument). This is wrong: only the shared_library() needs SONAME, while shared_module() does not. Moreover, tools such as debian's dpkg-shlibdeps use presence of SONAME field as an indicator that this is shared library as opposed to shared module (e.g., for the module it is okay to have unresolved symbols which are imported from the executable which loads the module, while a library should have all symbols resolved).

This was in fact already the behavior on Darwin.

Fixes: #8746


dcbaker (approved):
This makes sense to me, and the code looks correct.

Issue #9000, opened 2021-07-18
Meson not correctly process with `-l:xxx.a` link arguments in pkgconfig .pc file.


slayercat:
# Bug
**Describe the bug**
Meson not correctly process with `-l:xxx.a` link arguments in pkgconfig .pc file.

When links with `DPDK 21.05.0` staticly, Meson will complain could not find librarys startswith '-l:'. 

Also, generated `build.ninja` will not faithful Faithful presentation with `-Wl,--whole-archive` and `-Wl,--no-whole-archive`.

[DPDK](https://github.com/DPDK/dpdk) provides a `libdpdk.pc` file (pasted below).

After a few dig, I've found there's two issues to be fixed:

1. identify `-l:xxx.a` in `mesonbuild/dependencies/base.py` ([Line 814](https://github.com/mesonbuild/meson/blob/aeda7f249c4a5dbbecc52e44f382246a2377b5b0/mesonbuild/dependencies/base.py#L814))
2. some items could not be searched by `clib_compiler.find_library`, may be check `libpaths` for the filename? ([Line 819](https://github.com/mesonbuild/meson/blob/aeda7f249c4a5dbbecc52e44f382246a2377b5b0/mesonbuild/dependencies/base.py#L819))

A Sample meson configure outputs snippet attached.

A sample `build.ninja` generated snippet attached.

**To Reproduce**
A sample project with two files pasted below. run `meson` and check the result.

**Expected behavior**
1. meson will not complain about `Static library ':xxx.a' not found `
2. build.ninja will correctly sent link arguments inside `-Wl,--whole-archive` and `-Wl,--no-whole-archive`

**system parameters**

- Build Type: *plain native build*
- OS: *Centos 7.8* (pkg-config upgraded)
- Meson Version : *0.58.1*
- Ninja Version: *1.10.0.git.kitware.jobserver-1*
- pkg-config: *0.29.2*

## Sample meson configure output
```
WARNING: Static library ':librte_common_cpt.a' not found for dependency 'libdpdk', may not be statically linked
WARNING: Static library ':librte_common_dpaax.a' not found for dependency 'libdpdk', may not be statically linked
WARNING: Static library ':librte_common_iavf.a' not found for dependency 'libdpdk', may not be statically linked
WARNING: Static library ':librte_common_octeontx.a' not found for dependency 'libdpdk', may not be statically linked
WARNING: Static library ':librte_common_octeontx2.a' not found for dependency 'libdpdk', may not be statically linked
[...]

PR #9002, opened 2021-07-19
pkg-config: support for `-l:libfoo.a`

slayercat:
fixs: #9000 Meson not correctly process with -l:xxx.a link arguments in pkgconfig .pc file.

see also: https://stackoverflow.com/questions/48532868/gcc-library-option-with-a-colon-llibevent-a

### DIFF
--- a/mesonbuild/dependencies/pkgconfig.py
+++ b/mesonbuild/dependencies/pkgconfig.py
@@ -276,6 +276,30 @@ class PkgConfigDependency(ExternalDependency):
             elif lib.startswith('-L'):
                 # We already handled library paths above
                 continue
+            elif lib.startswith('-l:'):
+                # see: https://stackoverflow.com/questions/48532868/gcc-library-option-with-a-colon-llibevent-a
+                # also : See the documentation of -lnamespec | --library=namespec in the linker manual  
+                #                     https://sourceware.org/binutils/docs-2.18/ld/Options.html
+                
+                # Don't resolve the same -l:libfoo.a argument again
+                if lib in libs_found:
+                    continue
+                libfilename = lib[3:]
+                foundname = None
+                for libdir in libpaths:
+                    target = os.path.join(libdir, libfilename)
+                    if os.path.exists(target):
+                        foundname = target
+                        break
+                if foundname is None:
+                    if lib in libs_notfound:
+                        continue
+                    else:
+                        mlog.warning('Library {!r} not found for dependency {!r}, may '
+                                    'not be successfully linked'.format(libfilename, self.name))
+                    libs_notfound.append(lib)
+                else:
+                    lib = foundname
             elif lib.startswith('-l'):
                 # Don't resolve the same -lfoo argument again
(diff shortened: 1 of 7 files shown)


mesonbuild/dependencies/pkgconfig.py
```diff
@@ -277,6 +277,30 @@ def _search_libs(self, out: str, out_raw: str) -> T.Tuple[T.List[str], T.List[st
...
+                    target = os.path.join(libdir, libfilename)
+                    if os.path.exists(target):
+                        foundname = target
+                        break
+                if foundname is None:
+                    if lib in libs_notfound:
+                        continue
+                    if self.static:
+                        mlog.warning('Static library {!r} not found for dependency {!r}, may '
+                                    'not be statically linked'.format(libfilename, self.name))
```
mensinda, line 300: Do libraries specified with `-l:filename` *have* to be static, or does a `-l:foobar.so.0.0.1` also work here? If this is the case, this has to be adjusted.

mensinda (commented):
This looks mostly good to me, but a test case for this would be good to have. We already have some `*.pc` tests you could use as a template: https://github.com/mesonbuild/meson/blob/master/unittests/linuxliketests.py#L1088

mesonbuild/dependencies/pkgconfig.py
eli-schwartz: Good question! It does not, it's just treated as a filename literal rather than performing heuristic lookup while preferring the heuristics for shared or static libraries, depending.

slayercat:
Thanks for review, made some improvements:

1. subjected a test case for this issue.
2. changed the message when a library cannot be found.

test cases/unit/97 link full name/proguser/receiver.c
```diff
@@ -0,0 +1,19 @@
...
+int  __attribute__((weak)) get_checked(void) {
+    return -1;
+}
+
+
+#define CHECK_VALUE (100)
+#define TEST_SUCCESS (0)
+#define TEST_FAILTURE (-1)
+
+int main() {
```
mensinda, line 11: This needs to be changed to `int main(void)`, otherwise one of our CIs will fail (combination of all the warnings and `-Werror`)

slayercat: Fixed in commit 90a5f0ad1077eb7ea26af464c5d8dfc5f3724fe6

mensinda (approved):
LGTM, ~and the CI errors are unrelated.~

Thanks for fixing this.

mensinda (changes requested):
Actually, double-checking revealed that https://github.com/mesonbuild/meson/pull/9002/checks?check_run_id=3273735562 *is* related to this.

mensinda:
This time all CI failures are *really* unrelated. Merging.

@nirbheek I am adding this to 0.59.1 since I would count this as a bugfix, but feel free to remove it again if you think it doesn't fit.

PR #9005, opened 2021-07-19
Add expected skip annotations for non-linux CI runs to framework tests

jon-turney:
This is a follow-up to #8962, and adds the same expected skip annotations to framework tests for non-linux CI runs


jon-turney:
Possible future work:

- [ ] Review where framework tests are getting skipped (esp. on Windows, where some which used to run are now skipped)
- [ ] Extend to test categories other than framework?

Issue #9009, opened 2021-07-21
Error building glib with meson 0.59


jjardon:
Logs here: https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/jobs/1439678151

dcbaker:
Quickly, it looks like there's some caller of make_test not providing a timeout when it should. Probably party of the strict type checking changes

xclaesse:
@dcbaker That's because the gnome module uses interpreter internal method directly instead of going through typed_kwargs decorator that sets defaults. We'll have to be more careful for that in the future.

dcbaker:
I didn't realize that the gnome module wrapped the tests. It really should go through the internal methods, but without full type checking it's hard to make sure that everything is set correctly. :/

PR #9010, opened 2021-07-21
modules: Add API to add test

xclaesse:
This fix kwargs not going through typed_kwargs() decorator that sets
defaults.

Fixes: #9009


xclaesse:
@dcbaker I think this is ready to be merged.

PR #9011, opened 2021-07-21
Regression: Fix version check for MesonMain methods

xclaesse:


xclaesse:
Ideally we should make the subproject argument mandatory in `InterpreterObject.__init__()`. I'm not doing that here because this PR should be backported to 0.59.1, so better keep it safe.

mesonbuild/interpreterbase/baseobjects.py
```diff
@@ -82,7 +82,7 @@ def __repr__(self) -> str:
         return f'<[{type(self).__name__}] holds [{type(self.held_object).__name__}]: {self.held_object!r}>'
 
 class RangeHolder(MesonInterpreterObject):
-    def __init__(self, start: int, stop: int, step: int, *, subproject: T.Optional[str] = None) -> None:
+    def __init__(self, start: int, stop: int, step: int, subproject: str) -> None:
```
mensinda, line 85: `subproject` should stay kwarg only (to be consistent with the other calls) but be non-optional

xclaesse: right, done.

mensinda (approved):
LGTM if the CI also agrees,

Issue #9007, opened 2021-07-20
Regression: CustomTarget no longer allowed for qt.preprocess in 0.59


ximion:
Hi!

A project I am working on currently fails to build with Meson 0.59 because output of the Qt DBus generator / Qt Remote Object compiler is passed to `qt.preprocess`, resulting in this new error:

```
ERROR: qt.preprocess keyword argument "moc_headers" contained a value of type "CustomTarget" but should have been one of: "File", "str"
```

This has worked fine before (Meson <= 0.58), and is probably a sideeffect of the Qt refactoring that has happened - maybe it's just a matter of allowing `CustomTarget` as a parameter, I will have a look at the code later :-)

dcbaker:
It's probably pretty simple to fix, I'll have a look later today. So many of these interfaces are unfortunately badly documented as to what types they accept :/

ximion:
Nice, thank you very much for the really quick fix! :-)

PR #9012, opened 2021-07-21
Qt module generated sources

dcbaker:
This has two things in it, a fix for #9007, in which CustomTargets were erroneously excluded as inputs to the various `qt.preprocess` arguments, and then a series of patches to make the various `compile_*` arguments accept all forms of generated inputs, not just `CustomTargets` (including CustomTargetIndex and GeneratedList inputs`.

@nirbheek: the first patch to this series should be pulled for 0.59.1, but the rest should not.

@ximion: can you see if the first patch fixes your issue?

### DIFF
--- a/docs/markdown/_include_qt_base.md
+++ b/docs/markdown/_include_qt_base.md
@@ -21,8 +21,9 @@ It takes no positional arguments, and the following keyword arguments:
 Compiles Qt's ui files (.ui) into header files.
 
 It takes no positional arguments, and the following keyword arguments:
-  - `sources` (File | string)[]: A list of sources to be transpiled. Required,
-    must have at least one source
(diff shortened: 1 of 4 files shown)


nirbheek:
@dcbaker do you think you could put the first commit into a separate PR? It looks trivially correct, so we can merge it immediately, and that's also clearer for documenting what went into a stable release.

dcbaker:
Sure, i can split it out.

jpakkane:
:crossed_swords: conflicts now.

dcbaker:
github is so strange, that rebased without conflicts

jpakkane:
⚔️ conflicts.

dcbaker:
@jpakkane, rebased

Issue #7703, opened 2020-09-06
"links" test with file argument should use the appropriate compiler for the source file


bonzini:
**Describe the bug**
I would like to test that the C++ driver (e.g. `g++`) can be used to link C objects. One usecase is that the C compiler's libsanitizer might not be compatible with the one included by the C++ driver..

However, something like `meson.get_compiler('cpp').links(files('main.c'))` results in the following log:

```
Command line:  c++ /home/pbonzini/ff/main.c -o /home/pbonzini/ff/build/meson-private/tmpa0bmxj_m/output.exe -pipe -D_FILE_OFFSET_BITS=64 -O0 -fpermissive
```

instead of for example:

```
Command line:  cc /home/pbonzini/ff/main.c -o /home/pbonzini/ff/build/meson-private/tmpa0bmxj_m/output.o -pipe -D_FILE_OFFSET_BITS=64 -O0
Command line:  c++ /home/pbonzini/ff/build/meson-private/tmpa0bmxj_m/output.o -o /home/pbonzini/ff/build/meson-private/tmpa0bmxj_m/output.exe -pipe -O0
```

**To Reproduce**
* main.c: `int main() {}`
* meson.build:
```
project('a', ['c', 'cpp'])
meson.get_compiler('cpp').links(files('main.c'))
```

bonzini:
This is theoretically backwards-compatible, but it should be treated as a bugfix in my opinion. There's no way in Meson to compile a `.c` file with the C++ driver as part of a build target, therefore there would be no reason to want this configuration.

Another possibility is to add a `can_link_language` method to the compiler object. The two changes can be done independently.

PR #9014, opened 2021-07-22
Use appropriate compiler for the source file for "links" tests with file argument

bonzini:
These patches makes it possible to test that the C++ driver (e.g. g++) can be used to link C objects, for example to check that the C compiler's `libsanitizer` is compatible with the one included by the C++ driver.

For example `meson.get_compiler('cpp').links(files('main.c'))` results in the following log:

```
Command line:  cc /home/pbonzini/ff/main.c -o /home/pbonzini/ff/build/meson-private/tmpa0bmxj_m/output.obj -pipe -D_FILE_OFFSET_BITS=64 -O0
Command line:  c++ /home/pbonzini/ff/build/meson-private/tmpa0bmxj_m/output.obj -o /home/pbonzini/ff/build/meson-private/tmpa0bmxj_m/output.exe -O0
```
Fixes: #7703

### DIFF
--- a/docs/markdown/Reference-manual.md
+++ b/docs/markdown/Reference-manual.md
@@ -2513,7 +2513,10 @@ the following methods:
   positional argument compiles and links, you can specify external
   dependencies to use with `dependencies` keyword argument, `code` can
   be either a string containing source code or a `file` object
-  pointing to the source code.
+  pointing to the source code.  *Since 0.60.0*, if the `file` object's
+  suffix does not match the compiler object's language, the compiler
+  corresponding to the suffix is used to compile the source, while the
+  target of the `links` method is used to link the resulting object file.
 
 - `run(code)`: attempts to compile and execute the given code fragment,
   returns a run result object, you can specify external dependencies
--- a/mesonbuild/compilers/compilers.py
+++ b/mesonbuild/compilers/compilers.py
@@ -775,8 +775,11 @@ class Compiler(HoldableObject, metaclass=abc.ABCMeta):
                 contents = code
             else:
                 srcname = code.fname
-                with open(code.fname, encoding='utf-8') as f:
-                    contents = f.read()
+                if not is_object(code.fname):
+                    with open(code.fname, encoding='utf-8') as f:
+                        contents = f.read()
+                else:
+                    contents = '<binary>'
 
             # Construct the compiler command-line
             commands = self.compiler_args()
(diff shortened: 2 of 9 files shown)


dcbaker:
The first three cleanup patches look good, but I'm on my phone so thinking about behavioral changes is too hard

mesonbuild/compilers/compilers.py
```diff
@@ -83,6 +83,8 @@
     clink_suffixes += lang_suffixes[_l]
 clink_suffixes += ('h', 'll', 's')
 all_suffixes = set(itertools.chain(*lang_suffixes.values(), clink_suffixes))  # type: T.Set[str]
+suffix_to_lang = dict(itertools.chain(*(
+    [(suffix, lang) for suffix in v] for lang, v in lang_suffixes.items()))) # type: T.Dict[str, str]
```
dcbaker, line 87: this is a module constant, so let's treat it like one (the existing code is old and hasn't been updated). Convention in python is that constants are all caps with underscores:
```python
SUFFIX_TO_LANG: T.Dict[str, str] = ...
```

mesonbuild/interpreter/compiler.py
```diff
@@ -502,17 +503,31 @@ def links_method(self, args, kwargs):
         if len(args) != 1:
             raise InterpreterException('links method takes exactly one argument.')
         code = args[0]
+        compiler = None
         if isinstance(code, mesonlib.File):
             code = mesonlib.File.from_absolute_file(
                 code.rel_to_builddir(self.environment.source_dir))
+            suffix = code.suffix
+            if suffix not in self.compiler.file_suffixes:
+                for_machine = self.compiler.for_machine
```
dcbaker, line 512: unused?

run_unittests.py
```diff
@@ -4125,6 +4125,21 @@ def test_check_module_linking(self):
...
 
+    def test_mixed_language_linker_check(self):
+        testdir = os.path.join(self.unit_test_dir, '97 compiler.links file arg')
+        self.init(testdir)
+        cmds = self.get_meson_log_compiler_checks()
+        self.assertEqual(len(cmds), 5)
+        cc = cmds[0][0]
+        cxx = cmds[1][0]
+        if cc == cxx:
+            raise unittest.SkipTest(f'Using a single compiler driver for both C and C++')
```
dcbaker, line 4136: is there really any reason not to test the cc == cxx path?

dcbaker (changes requested):
I've got a couple of style things and a question, but I think otherwise this is good.

mesonbuild/compilers/compilers.py
bonzini: I didn't use the "a: b = c" form for consistency with the other lines, but uppercase is better indeed.

run_unittests.py
bonzini: I was thinking that in principle you do not need the separate compiler/linker execution if cc == cxx. But in that case the test would fail because len(cmds) == 4; so I just removed the if.

dcbaker (approved):
Looks good, thanks

docs/markdown/Reference-manual.md
```diff
@@ -2463,7 +2463,10 @@ the following methods:
   positional argument compiles and links, you can specify external
   dependencies to use with `dependencies` keyword argument, `code` can
   be either a string containing source code or a `file` object
-  pointing to the source code.
+  pointing to the source code.  If the `file` object's suffix does
+  not match the compiler object's language, the compiler corresponding
+  to the suffix is used to compile the source, while the target of
+  the `links` method is used to link the resulting object file.
```
jpakkane, line 2469: This should mention that this is only with 0.59 onwards. And should this say "while the compiler of the `links` method" rather than "while the target of the `links` method"?

dcbaker: 0.60.0, right? 😃

bonzini: I used "target" as in "what's to the left of the dot". Added the "since" now.

bonzini:
Ping.

bonzini:
Ping again...

PR #9016, opened 2021-07-22
qt: Allow CustomTargets for qt.preprocess source arguments

### LINKED ISSUE #9007: Regression: CustomTarget no longer allowed for qt.preprocess in 0.59

dcbaker:
This works for `moc_*` and `ui_files`, but it never could have worked
for `qresources` due to the implementation assuming a `str` or `File`.
To restore previous compatibility I've added `CustomTarget` where it
would have worked, but not where it would have failed, the former would
raised an exception along the lines anyway.

Fixes #9007

### DIFF
--- a/docs/markdown/_include_qt_base.md
+++ b/docs/markdown/_include_qt_base.md
@@ -62,10 +62,10 @@ sources += qt.preprocess(qresources : ['resources'])
 ```
 
 This method takes the following keyword arguments:
- - `qresources` (string | File | CustomTarget | BuildTarget)[]: Passed to the RCC compiler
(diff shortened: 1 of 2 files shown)


dcbaker:
@nirbheek as requested, this is the first patch of #9012 split out

PR #9019, opened 2021-07-22
Fix #8274

### LINKED ISSUE #8274: Non-actionable error message "ERROR: No host machine compiler for ..."
**Describe the bug**
I started a meson project with a single C file.  My meson command does not work, it ends up with this obscure error message:

```
../meson.build:5:0: ERROR: No host machine compiler for "get-size.c"

A full log can be found at /var/tmp/tt/build/meson-logs/meson-log.txt
```
There is no further information in the log file.

**To Reproduce**

Here is my full `meson.build` file:
```
project('myprog', version : '0.1')
nbd_dep = dependency('libnbd')
executable('get-size', 'get-size.c', dependencies : [nbd_dep])
```
There is a single source file (`get-size.c`) which I am not including but it's not important (except that it exists) because we never get as far as compiling.

**Expected behavior**
The error message should be actionable - it should tell me what's wrong and how to correct it.

**system parameters**
* Is this a [cross build](https://mesonbuild.com/Cross-compilation.html) or just a plain native build (for the same computer)? **No**
* what operating system (e.g. MacOS Catalina, Windows 10, CentOS 8.0, Ubuntu 18.04, etc.) **Fedora**
* what Python version are you using **python3-3.9.1-1.fc33.x86_64**
* what `meson --version` **meson-0.55.3-1.fc33.noarch**
* what `ninja --version` if it's a Ninja build **1.10.2**

IsaccBarker:
Fix #8274: Add more detailed error on no machine compiler message.

### DIFF
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -810,7 +810,7 @@ class BuildTarget(Target):
                         break
                 else:
                     if is_known_suffix(s):
-                        raise MesonException('No {} machine compiler for "{}"'.
+                        raise MesonException('No {} machine compiler for "{}". Check the documentation of `project` for how to specify required compilers!'.
                                              format(self.for_machine.get_lower_case_name(), s))
 
             # Re-sort according to clink_langs


jpakkane:
The test failure seems relevant:

```
  [ERROR]    failing: 100 no lang
During: configure
Reason: Unexpected output while configuring.
```

mesonbuild/build.py
```diff
@@ -810,7 +810,7 @@ def process_compilers(self):
                         break
                 else:
                     if is_known_suffix(s):
-                        raise MesonException('No {} machine compiler for "{}"'.
+                        raise MesonException('No {} machine compiler for "{}". Check the documentation of `project` for how to specify required compilers!'.
```
jon-turney, line 813: This definitely needs improving, but I'm not sure this is how:

* The documentation for `project` doesn't mention compilers.  It mentions languages (which have compilers)
* Languages can also be added with `add_language`
* You can probably also get this error in cross-situations where the language has been specified, but the compiler isn't required (this is the 'The build machine compilers for the specified languages are not required' sentence in the documentation for `project`) and isn't present, but you still try to use it. (see https://github.com/mesonbuild/meson/pull/6532#issuecomment-583859162 for the gruesome details)

IsaccBarker (commented):
- Hv s i. Uuuuh ei u. S []()r te rrdud 


Eg image

PR #9021, opened 2021-07-22
WIP: pkg-config dependency consistency

xclaesse:
commit 67f8caaa08194d484eeff9671f15b50da9d45316

    dependencies: Make sure we cannot mix pkg-config and subproject deps
    
    When using a dependency from pkg-config we have to make sure that any
    transitive dependencies it pulls are also always taken from pkg-config
    and cannot come from subproject override.

commit 15cf5c691899ff7ce82ccb31ec199c010edb470b

    Add internal pkg-config implementation
    
    This adds a python implementation of pkg-config parser. It also adds
    `meson pkg-config` internal command (hidden from meson --help output),
    similar CLI to original pkg-config, for testing purpose only.


mesonbuild/environment.py
```diff
@@ -572,6 +573,106 @@ def __init__(self, source_dir: T.Optional[str], build_dir: T.Optional[str], opti
...
+
+    @staticmethod
+    def create_pkgconfig_repo(for_machine: MachineChoice = MachineChoice.BUILD,
+                              is_cross: bool = False,
+                              system_paths: T.Optional[T.List[str]] = None,
+                              extra_paths: T.Optional[T.List[str]] = None,
+                              sysroot_dir: T.Optional[str] = None)  -> pkgconfig.Repository:
+        # Get some default paths. This is normally defined by pkg-config packager,
+        # it could need per distro/platform tweaking in the future.
+        default_prefixes = [Path('/usr/local'), Path('/usr')]
```
tristan957, line 586: Would it make sense to look at ~/.local as well?

xclaesse: I don't think any distro/platform configure pkg-config/pkgconf to include that path, is there? I'm not against it if there is use-case for it? In any case, it can already be done by setting PKG_CONFIG_PATH.

tristan957: If you are only going for distro config, then this is good as is. ~/.local would be for extra convenience, but mimicking distros sounds like a good idea to me.

xclaesse:
@dcbaker Since you worked on pkg-config implementation for meson++, would appreciate your review here :)

eli-schwartz:
I continue to be skeptical about implementing yet another version of pkg-config. There's too many of them already. There's even a python one at https://github.com/gbiggs/pykg-config. They all behave differently -- the perl one is known-bad and Meson already rejects it unconditionally.

Among the issues which a vendored implementation cannot ever solve, we need to respect `$PKG_CONFIG` and use it to allow selecting toolchain-specific layouts compiled into the binary. We cannot reliably detect the OS layout, which is normally compiled into pkg-config using `--with-pkg-config-dir`, when nothing is specified for pkg_config_path / libdir.

Dependency consistency feels like something that should be solved another way. @kaniini suggested on IRC that pkgconf could add a `--poison` flag, so we could do e.g.

```
pkgconf --poison "glib,zlib" --modversion "gtk4"
```
and it would detect that gt4.pc from the system could normally be returned successfully, but in this case cannot because it's not allowed to depend on zlib. So we'd just pass all our overridden subprojects as `--poison`ed names.

This would solve dependency consistency without having our own lightly tested version of pkg-config that resolves different things depending on whether Meson has been invoked with a specified `$PKG_CONFIG` or not, and which is only tested by exactly one project in the entire world.

dcbaker:
I agree with @eli-schwartz on this. There’s so many corners and the test coverage is not there. I would expect at least all of pkgconf’s tests to be imported. Pkg-conf is an awful format

xclaesse:
> Among the issues which a vendored implementation cannot ever solve, we need to respect `$PKG_CONFIG` and use it to allow selecting toolchain-specific layouts compiled into the binary.

Note that this PR only use the internal implementation to get the chain of `requires`, it still uses pkg-config CLI for getting cflags, etc. In the future I would like to at least optionally use the internal implementation in the case pkg-config CLI is not available such as on Windows, to avoid things like https://gitlab.freedesktop.org/gstreamer/gstreamer/-/merge_requests/1584/diffs?commit_id=824ecac7a31c8dd7881db98c08fa88c7c08ee51a where they download prebuilt pkg-config.exe.

If PKG_CONFIG is set, of course we can always keep using the CLI instead of internal implementation.

> We cannot reliably detect the OS layout, which is normally compiled into pkg-config using `--with-pkg-config-dir`, when nothing is specified for pkg_config_path / libdir.

That's true, and that's in my opinion the only real issue with dropping CLI completely. I think the current detection is good enough for optional consistency checks this PR does, worst case, pc files are not found and no checks are performed.

> Dependency consistency feels like something that should be solved another way. @kaniini suggested on IRC that pkgconf could add a `--poison` flag, so we could do e.g.

That could work, but means it would work only with pkgconf, and solves only this specific issue. I think having our own internal implementation opens the door for more things, like dropping the on-disk cache needed because CLI is slow and causes reconfiguration issues, support platforms that does not have pkg-config installed.

xclaesse:
> I agree with @eli-schwartz on this. There’s so many corners and the test coverage is not there. I would expect at least all of pkgconf’s tests to be imported. Pkg-conf is an awful format

It's true that the format is awful, but it's not that complicated neither. I wrote a pretty decent test already and ran it on all pc files I can find on my system. More tests is of course always a good thing, I could see if we can import pkgconf tests. But again, this PR keeps using the CLI for everything important, it uses the internal implementation only for this specific dependency consistency check. The idea is to see how we can mature it and extend its usage over time.

xclaesse:
> `pkgconf --poison "glib,zlib" --modversion "gtk4"`

That actually only covers half the issue here, we also want to detect the case where we `meson.override_dependency()` with a dependency name that has been looked before. That means that every time we do a pkg-config lookup we need the transitive list of requires. We can of course add a new CLI for that, but it means doing another call which slows down configure time.

kaniini:
While I think that a python implementation of pkg-config makes a lot of sense for Meson (the story about downloading a pre-built `pkg-config.exe` is compelling, though they should use pkgconf anyway), I don't think it is time for that yet.

At present, we are trying to improve upon the worst aspects of the pkg-config format, and formally specify behavior of the solver, including the creation of high quality test vectors.  Having to sync with another implementation of pkg-config which lacks these improvements will just hold back forward progress.

Now is not the time.  Maybe by this time next year, we can have an actual spec for pkg-config implementations, including all of the new things people are asking for (moving away from CFLAGS/LDFLAGS fragments to something with better semantics, copyright and license tracking in .pc files for SBOMs, etc.), and then Meson can benefit from having a pkg-config implementation that is provably equivalent to pkgconf.

kaniini:
Based on IRC conversations, at least my current concerns are largely assuaged by this work, so I think it should be merged, actually :)

japm48:
@kaniini @xclaesse Is there any progress regarding this?
On Windows with MSVC, having to manually install `pkg-config` is a bit of a nuisance. But at least I can use `winget install bloodrock.pkg-config-lite`.

PR #9027, opened 2021-07-24
interperter: Fix list contains for Holders (fixes #9020)

### LINKED ISSUE #9020: contains() no longer works with lists of file objects
**Describe the bug**

With 0.59.0, `[].contains()` no longer works as expected with file objects

**To Reproduce**

foo.c:

```
```
meson.build:

```
project('contains-bug', 'c',
  meson_version: '>= 0.59.0',
)

l = [files('foo.c')]

message(l.contains(files('foo.c')))
```

**Expected behavior**

This should print `Message: True`

**system parameters**

* plain native build
* Fedora 33
* python 3.9.2
* meson 0.59.0
* ninja 1.10.2

mensinda:


dcbaker:
I'm not sure what do here, this changed behavior. The old code would not have matched something like `[['a'], 'b'].contains('a')`, right? I'm just not sure what the correct behavior is in this case

mensinda:
Not quite, your example would have matched since the old code does have effective list flattening with the recursion. What wouldn't have matched would be something like`['a', 'b'].contains([[['a']]])`.

I guess that this would be a change in behavior, but without the flattening the argument list, the `.contains(files('foo.txt'))` breaks...

mensinda:
@MathieuDuponchelle I checked with 0.58.2, and the test `[files('a.txt', 'b.txt')].contains(files('a.txt'))` fails with 0.58.2 and succeeds with this PR.

I am not sure what this means for FFmpeg...

MathieuDuponchelle:
@mensinda it shouldn't make a difference for the FFmpeg build. Anyway, I can't comment on whether the flattening is desirable or not, but if my second test case passes with this then the FFmpeg build should be fixed :)

MathieuDuponchelle:
@mensinda just confirming that this fixes the FFmpeg build, thanks again for that. I'll let others more familiar with the code comment on the MR now.

tp-m:
Anyone who could review this?

mesonbuild/interpreterbase/interpreterbase.py
```diff
@@ -783,19 +783,11 @@ def array_method_call(self,
...
-                        if found:
-                            return True
-                    if element == item:
-                        return True
-                return False
-            return check_contains(obj)
+            elements = mesonlib.listify(_unholder(obj), flatten=True)
+            posargs = mesonlib.listify(posargs, flatten=True)
+            if len(posargs) != 1:
+                raise InterpreterException('Contains method takes exactly one argument.')
```
nirbheek, line 789: This actually breaks this:

```meson
list = [['a', 'b'], ['c', 'd']]
message(list.contains(['a', 'b']))
```

Returns `true` in 0.58 and `true` in master, but causes a hard error with this PR.

This change also breaks this:

```meson
flist = [files('a.c', 'b.c'), files('foo.c', 'bar.c')]
message(flist.contains(files('a.c', 'b.c')))
```

Returns `true` in 0.58 and `false` in master (incorrectly), but with this change it causes a hard error.

I think the correct fix is not to mangle `posargs` but to check for both `obj` and `_unholder(obj)`. That does actually work for me:

```diff
@@ -787,13 +787,14 @@ The result of this is undefined and will become a hard error in a future Meson r
                 if len(posargs) != 1:
                     raise InterpreterException('Contains method takes exactly one argument.')
                 item = posargs[0]
-                for element in el:
-                    if isinstance(element, list):
-                        found = check_contains(element)
-                        if found:
+                for each in (el, _unholder(el)):
+                    for element in each:
+                        if isinstance(element, list):
+                            found = check_contains(element)
+                            if found:
+                                return True
+                        if element == item:
                             return True
-                    if element == item:
-                        return True
                 return False
             return check_contains(obj)
         elif method_name == 'length':
```

test cases/common/56 array methods/meson.build
```diff
@@ -5,6 +5,30 @@ one = ['abc']
...
+  error('Files are not equal')
+endif
+
+if not file_list.contains(file_a[0])
+  error('Contains with ObjectHolder lists does not work')
+endif
+
+if not file_list.contains(file_a)
+  error('Contains with ObjectHolder lists does not work -- flattening edition')
+endif
```
nirbheek, line 22: This doesn't work in 0.58.

test cases/common/56 array methods/meson.build
```diff
@@ -5,6 +5,30 @@ one = ['abc']
...
+
+if not [file_c, [file_list]].contains([file_a])
+  error('Contains with ObjectHolder lists does not work -- flattening edition 2')
+endif
```
nirbheek, line 26: This also doesn't work in 0.58

mesonbuild/interpreterbase/interpreterbase.py
nirbheek: Note that this doesn't fix the two cases below that didn't work in 0.58, so it likely needs more changes.

nirbheek: I think `listify` is wrong for generic array operations. It's only correct when passing args to things that expect a flattened list, like files list for a target or a command-line argument list.

mensinda: I dropped the `listify` refactoring and reverted back to one `_unholder` pass at the start.

FelixSchwarz:
just a nitpick but you might want to fix the commit message ("interperter" -> "interpreter").

mensinda:
@MathieuDuponchelle could you please check again? I made some (rather significant) changes since your test.

xclaesse:
Tested ffmpeg again it this PR fix it. LGTM.

PR #9029, opened 2021-07-24
Split run_unittests.py file

xclaesse:


xclaesse:
I expect this is going to take many try-and-error until I get it to pass on CI.

dcbaker (approved):
If you can get the CI to pass you have my acked-by on this. I didn't look at it closely, but it I did look at the run_unittests.py changes, at those seem reasonable.

jon-turney:
Nice.  this also needs to change the path filters in the CI configuration files so CI which runs unittests runs on changes just in unittests/

mensinda (approved):
This also gets my stamp of approval (haven't looked through at all the changes though :D)

xclaesse:
> This also needs to change the path filters in the CI configuration files so CI which runs unittests runs on changes just in unittests/

oh good point, forgot that, done.

PR #9030, opened 2021-07-25
find_program: fix search order according to doc

kynzie:
`find_program()` allows to give extra search directories. The documentation says the following about the search order:

> 1. Program overrides set via meson.override_find_program()
> 2. [provide] sections in subproject wrap files, if wrap_mode is set to forcefallback
> 3. [binaries] section in your machine files
> 4. Directories provided using the dirs: kwarg (see below)
> 5. Project's source tree relative to the current subdir
>     - If you use the return value of configure_file(), the current subdir inside the build tree is used instead
> 6. PATH environment variable
> 7. [provide] sections in subproject wrap files, if wrap_mode is set to anything other than nofallback

If the given program is installed system wild, the search is never done in extra search directories, even if there are some defined.

Because, in [program.py, search() function](https://github.com/mesonbuild/meson/blob/2e30b5a1e2ca21555bb3aa50f91e87f4d2821bb6/mesonbuild/programs.py#L290), if `search_dir` function fails, the search looks directly in PATH instead of extra directories as said in the documentation.


mensinda:
This looks good code-wise, but this needs a test case (so we don't accidentally break it in the future).

You should be able to write a project test (new one or use an existing in `test cases/common`) for this by setting/patching `PATH` in a [`test.json`](https://mesonbuild.com/Contributing.html#configuring-project-tests).

kynzie:
While writing my test, I came across a problem I do not understand.

I added a `test_path.py` script in a `path` directory in `test\ cases/common/26 find program/`.

Here is the end of my `meson.build`:

```python
prog = find_program('test_path.py')
assert(prog.found(), 'Program should be found')
`
[test-run.log](https://github.com/mesonbuild/meson/files/6873839/test-run.log)
``

Here is my `test.json`:
```json
{
  "env": {
    "PATH": "@ROOT@/path:@PATH@"
  }
}
```

When I run:
``` shell
$ ./run_project_tests.py --only "common/26 find program"
```

Meson is run twice:
* First time, my `test_path.py` script is found
* The second time, it is not found: it seems `ninja` is reconfiguring `meson` but without `PATH` override.

Can someone explain me what is happening in the second part ? @mensinda maybe ?

mensinda:
@kynzie Could you check if this patch fixes the error?

```diff
diff --git a/run_project_tests.py b/run_project_tests.py
index 648ec47b5..68bc1c8a1 100755
--- a/run_project_tests.py
+++ b/run_project_tests.py
@@ -661,7 +661,7 @@ def _run_test(test: TestDef,
     # Build with subprocess
     def build_step() -> None:
         build_start = time.time()
-        pc, o, e = Popen_safe(compile_commands + dir_args, cwd=test_build_dir)
+        pc, o, e = Popen_safe(compile_commands + dir_args, cwd=test_build_dir, env=test.env)
         testresult.add_step(BuildStep.build, o, e, '', time.time() - build_start)
         if should_fail == 'build':
             if pc.returncode != 0:
```

kynzie:
Yes ! It does !

Do I need to add this patch in my PR ? or is it already in a merging process ?

mensinda:
No, it isn't. Please include it.

mesonbuild/programs.py
```diff
@@ -287,14 +284,15 @@ def _search_windows_special_cases(self, name: str, command: str) -> T.List[T.Opt
                 return commands
         return [None]
 
-    def _search(self, name: str, search_dir: T.Optional[str]) -> T.List[T.Optional[str]]:
+    def _search(self, name: str, search_dirs: T.List[T.Optional[str]]) -> T.List[T.Optional[str]]:
```
dcbaker, line 287: I think this should be `Optional[List[str]]`

dcbaker (changes requested):
I think we also need to ensure that the extra_dies argument is actually a list and not a string or none, which means changes in the interpreter

mensinda:
This test case also doesn't seem to actually check that `dirs` is searched before PATH. Also, the test error looks related.

mesonbuild/programs.py
kynzie: Done in 7eb1997

kynzie:
Yes, you are right @mensinda. I just change the test and duplicate it in order to ensure the order on windows and linux like platforms.

mesonbuild/programs.py
mensinda: @dcbaker I would disagree here... Looking at the code, it should either be `T.List[T.Optional[str]]` or `T.List[str]`. Passing `None` to this function would result in an error (iterating over `None`).

Though, I would say that `T.List[str]` would make the most sense in this case.

kynzie: regarding the call to `_search` done  in [`ExternalProgram constructor`](https://github.com/mesonbuild/meson/blob/46380b5b1e82a5c2b49299eb35ac185d4fa02e18/mesonbuild/programs.py#L62), I would agree on `T.List[T.Optional[str]]`.

kynzie:
@mensinda @dcbaker Can you tell me if the new tests I added are ok for you?

> I think we also need to ensure that the extra_dies argument is actually a list and not a string or none, which means changes in the interpreter

@dcbaker I do not understand why we need to ensure this. Because in [`ExternalProgram` constructor](https://github.com/mesonbuild/meson/blob/46380b5b1e82a5c2b49299eb35ac185d4fa02e18/mesonbuild/programs.py#L62), `all_search_dirs` is a list whatever the value of `extra_search_dirs` is : a list of `None` can happen. So, I do not see the point :confused:

Anyway, after diving in `interpreter/interpreter.py` code, I can say:
- Can be `None`: [case 1](https://github.com/mesonbuild/meson/blob/46380b5b1e82a5c2b49299eb35ac185d4fa02e18/mesonbuild/interpreter/interpreter.py#L793), [case 2](https://github.com/mesonbuild/meson/blob/46380b5b1e82a5c2b49299eb35ac185d4fa02e18/mesonbuild/interpreter/interpreter.py#L1491), [case 3 (depending on the value of `search_dirs` which can be `None`)](https://github.com/mesonbuild/meson/blob/46380b5b1e82a5c2b49299eb35ac185d4fa02e18/mesonbuild/interpreter/interpreter.py#L1395)
- Can be an empty list: [case](https://github.com/mesonbuild/meson/blob/46380b5b1e82a5c2b49299eb35ac185d4fa02e18/mesonbuild/interpreter/interpreter.py#L1389)

mesonbuild/programs.py
dcbaker: I don't think we should be passing `None` here at all, looking at the code. But I guess we can leave it for now.

dcbaker (approved):
This looks fine to me, though there should probably be a release snippet, and something the docs mentioning that the behavior changed in 0.60.0

kynzie:
@dcbaker I am an idiot. I was looking for the problem more complex than an iteration over a list of `None`. I go it. Let me suggest a patch in the constructor, please. I will do it by the end of the day.

kynzie:
@dcbaker You were right about modifying the interpreter.

So, I removed `extra_search_dirs` variable and replace it by `search_dirs` which merge  `extra_search_dirs` and `search` variables and keeps the right order for the search. Furthermore, `search_dirs` is always a list of strings : empty or not.

kynzie:
@dcbaker You were right about modifying the interpreter.

So, I removed `extra_search_dirs` variable and replace it by `search_dirs` which merge  `extra_search_dirs` and `search` variables and keeps the right order for the search. Furthermore, `search_dirs` is always a list of strings : empty or not.

kynzie:
@dcbaker or @mensinda may I have a review ?

mesonbuild/interpreter/interpreter.py
```diff
@@ -1427,7 +1424,7 @@ def notfound_program(self, args):
     # TODO update modules to always pass `for_machine`. It is bad-form to assume
     # the host machine.
     def find_program_impl(self, args, for_machine: MachineChoice = MachineChoice.HOST,
-                          required=True, silent=True, wanted='', search_dirs=None,
+                          required=True, silent=True, wanted='', search_dirs=[],
```
mensinda, line 1427: You should in general *never* pass `[]` (or any other mutable type) as a default parameter in python, since this exact same list gets *reused* for each function call. Not sure who thought that this is a good idea but here we are...

So, instead please keep the `search_dirs=None` and add a `search_dirs = search_dirs or []` at the start of the function.

The following example illustrates the problem:

```py
#!/usr/bin/env python3

def foo(data=[]):
    data += ['foo']
    return data

print(foo())
print(foo())
print(foo())
```

Result:
```
['foo']
['foo', 'foo']
['foo', 'foo', 'foo']
```

mensinda (changes requested):
This is almost good to go, but the CI *must* be green for any PR to get merged. Having a green (or mostly green) CI also helps for getting a PR reviewed since those usually get priority and are less likely to be forgotten :)

Also please rebase onto master again just in case.

PR #9031, opened 2021-07-26
build: learn to take CustomTargetIndex as custom_target command argument

elmarco:
Fix ERROR: Argument <CustomTargetIndex:...>[0]> in "command" is invalid.

### DIFF
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -2186,6 +2186,10 @@ class CommandBase:
             elif isinstance(c, (BuildTarget, CustomTarget)):
                 self.dependencies.append(c)
                 final_cmd.append(c)
+            elif isinstance(c, CustomTargetIndex):
+                FeatureNew.single_use('CustomTargetIndex for command argument', '0.60', self.subproject)
+                self.dependencies.append(c.target)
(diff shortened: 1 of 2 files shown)


mesonbuild/build.py
```diff
@@ -2186,6 +2186,8 @@ def flatten_command(self, cmd):
             elif isinstance(c, (BuildTarget, CustomTarget)):
                 self.dependencies.append(c)
                 final_cmd.append(c)
+            elif isinstance(c, CustomTargetIndex):
```
dcbaker, line 2189: you also need `self.dependencies.append(c.target)` here, or you won't create a proper dependency on the target.

You'll also need to emit a `FeatureNew.single_use()` here, as this is a new feature for 0.60

dcbaker (changes requested):
This also needs a release snippet, and possibly an update to the reference manual.

mesonbuild/build.py
elmarco: done

elmarco:
> This also needs a release snippet, and possibly an update to the reference manual.

Are you sure it's worth it? When I wrote the test I realize where my "mistake" was. I don't think people really can reach that condition easily.

dcbaker (approved):
I guess not? At least with the FeatureNew we have some form of documentation.

PR #9033, opened 2021-07-26
doc: mention CUDA_PATH in dependency('cuda') [skip ci]

SoapGentoo:
Let's document this, I always forget the variable.
@obilaniu @jpakkane


jpakkane:
We have updated the pipeline stuff so `[skip ci]` is no longer needed.

PR #9035, opened 2021-07-26
Add type checking for the Compiler and RunResult objects

dcbaker:
This uses a full set of `typed_pos_args` and `typed_kwargs` to the `interpreter/compiler` module. In doing this I discovered several deficiencies in the `compiler` annotations, as well as one case of undesirable behavior for the type checkers.

I did remove the `unittest_args` method, which has been marked in the docstring as deprecated for some time, as well as not being documented. I feel like given those two features, it's fine to remove it.

### DIFF
--- a/docs/markdown/snippets/remove_unittest_args_method.md
+++ b/docs/markdown/snippets/remove_unittest_args_method.md
@@ -1,0 +1,4 @@
+## Compiler.unittest_args has been removed
+
+It's never been documented, and it's been marked deprecated for a long time, so
+let's remove it.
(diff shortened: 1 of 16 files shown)


dcbaker:
I don't understand why this test fails on some instances, and passes on others, and always passes on my machine.

dcbaker:
Anyone object to merging this?

mensinda:
I only had time to skim through it, but looks good to me from what I have seen.

mensinda (approved):
I hade a propper look now, and this should be good to go after a rebase to check that the CI is green.

Issue #8876, opened 2021-06-13
gnome.generate_gir incorrectly filters clang cflags like -W , passed to g-ir-scanner, and gcc fails


bootchk:
**Describe the bug**

The symptom is that, using clang, meson calls g-ir-scanner which calls gcc which fails with something like

```
 "gcc: error: unrecognized command-line option ... -Wsometimes-unitialized"
 ```

 It seems that gnome.generate_gir is passing clang specific c flags to generate warnings, but g-ir-scanner calls gcc,
 and gcc fails, not recognizing flags specific to clang compiler.

**To Reproduce**

The case is a meson build of the GIMP app v2.99, using clang, on Linux, Alpine or Ubuntu.

**Analysis and suggested fix**

In the code for gnome.generate_gir there is this:

```
def _get_scanner_cflags(cflags):
        'g-ir-scanner only accepts -I/-D/-U; must ignore all other flags'
        for f in cflags:
            if f.startswith(('-D', '-U', '-I')):
                yield f
```

So it seems that the intent is to filter out all the -W flags.

But later:

```
        cflags, internal_ldflags, external_ldflags = self._get_langs_compilers_flags(state, langs_compilers)
        ...omitted...
        dep_cflags, dep_internal_ldflags, dep_external_ldflags, gi_includes = \
            self._get_dependencies_flags(deps, state, depends, use_gir_args=True)
        cflags += list(self._get_scanner_cflags(dep_cflags))
```

In the above, cflags is initially an unfiltered list of cflags, and then the code appends more cflags
that are filtered by calls to _get_scanner_cflags().

I believe a fix is to add one line like so:

```
cflags, internal_ldflags, external_ldflags = self._get_langs_compilers_flags(state, langs_compilers)
cflags = list(self._get_scanner_cflags(cflags))
```

That is, immediately filter the initial contents of the variable cflag.

I tested this fix minimally, only to build GIMP.
The meson build of GIMP is "experimental", not used for releases.

dcbaker:
That seems like a bug, it also seems like a bug that we're letting g-ir-scanner call GCC when Meson is using Clang for compilation. That might work out with GCC and clang (though I can come up with case that it wouldn't work)

bootchk:
[...]

PR #9041, opened 2021-07-29
gnome: Filter all compiler and linker flags

ebassi:
When sending them to the introspection scanner, we must always filter
all compiler and linker flags, because g-ir-scanner does not accept any
random argument but only a specific subset.

Fixes: #8876

### DIFF
--- a/mesonbuild/modules/gnome.py
+++ b/mesonbuild/modules/gnome.py
@@ -867,10 +867,16 @@ class GnomeModule(ExtensionModule):
         # are not used here.
         dep_cflags, dep_internal_ldflags, dep_external_ldflags, gi_includes = \
             self._get_dependencies_flags(deps, state, depends, use_gir_args=True)
-        cflags += list(self._get_scanner_cflags(dep_cflags))
-        cflags += list(self._get_scanner_cflags(self._get_external_args_for_langs(state, [lc[0] for lc in langs_compilers])))
-        internal_ldflags += list(self._get_scanner_ldflags(dep_internal_ldflags))
-        external_ldflags += list(self._get_scanner_ldflags(dep_external_ldflags))
+        scan_cflags = []
+        scan_cflags += list(self._get_scanner_cflags(cflags))
+        scan_cflags += list(self._get_scanner_cflags(dep_cflags))
+        scan_cflags += list(self._get_scanner_cflags(self._get_external_args_for_langs(state, [lc[0] for lc in langs_compilers])))
+        scan_internal_ldflags = []
+        scan_internal_ldflags += list(self._get_scanner_ldflags(internal_ldflags))
+        scan_internal_ldflags += list(self._get_scanner_ldflags(dep_internal_ldflags))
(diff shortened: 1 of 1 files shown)


xclaesse:
Looks good, assuming you tested that it does the right thing. CI does not seems to like it however.

xclaesse:
On a side note, I really wish g-ir-scanner stop compiling/running binaries. It should only generate the .c and let meson/ninja build and run it. That would massively simplify this thing because would wouldn't have to pass flags Meson's internal flags to an external script

mesonbuild/modules/gnome.py
```diff
@@ -867,10 +867,16 @@ def generate_gir(self, state, args, kwargs: T.Dict[str, T.Any]):
...
-        cflags += list(self._get_scanner_cflags(dep_cflags))
-        cflags += list(self._get_scanner_cflags(self._get_external_args_for_langs(state, [lc[0] for lc in langs_compilers])))
-        internal_ldflags += list(self._get_scanner_ldflags(dep_internal_ldflags))
-        external_ldflags += list(self._get_scanner_ldflags(dep_external_ldflags))
+        scan_cflags = []
+        scan_cflags += list(self._get_scanner_cflags(cflags))
+        scan_cflags += list(self._get_scanner_cflags(dep_cflags))
+        scan_cflags += list(self._get_scanner_cflags(self._get_external_args_for_langs(state, [lc[0] for lc in langs_compilers])))
+        scan_internal_ldflags = []
+        scan_internal_ldflags += list(self._get_scanner_ldflags(internal_ldflags)
```
dcbaker, line 875: missing brace here

ebassi: Indeed.

ebassi:
> On a side note, I really wish g-ir-scanner stop compiling/running binaries. It should only generate the .c and let meson/ninja build and run it. That would massively simplify this thing because would wouldn't have to pass flags Meson's internal flags to an external script

It's not entirely trivial to decouple the `dumper.c` generation, build, and compilation. Actually: I have no idea how to even begin doing that.

xclaesse:
> It's not entirely trivial to decouple the `dumper.c` generation, build, and compilation. Actually: I have no idea how to even begin doing that.

It's not that complicated, I had a WIP implementation for gtk-doc that I never finished.

You can just add a `g-ir-scanner --generate-only` option, and in `dumper.py` you `exit()` (or propertly return) after writing the file and before compilation. Then meson can create a proper `executable()``target to compile/link that source file, and a custom_target() that run it to generate `dump.xml`. And finally, meson creates a last custom_target() that runs `g-ir-scanner --with-dump=dump.txt`  and it would skip everything and go directly into the step where that file is parsed in `gdumpparser.py`,  `_execute_binary_get_tree()`.

xclaesse:
LGTM.

@ebassi how confident are you about tagging this for 0.59.1? This is not a regression, is it?

ebassi:
It's not a regression *per se*, so I think it can wait until the next release.

PR #9042, opened 2021-07-29
interenaltests: Add some version strings with ansi escapes in them

dcbaker:
As we've run into a few programs using ansi escapes in their version
outputs, we'd like to test them as well.


textshell (approved):
lgtm

PR #9044, opened 2021-07-30
Fix native targets for vs backend cross compilation

GustavoLCR:
This fixes the `.vcxproj` output for native targets during cross compilation using vs backends.
Tested by using the `vcvarsall.bat amd64_arm64` cross compilation toolchain and the `cross/arm64cl.txt` file.

### DIFF
--- a/mesonbuild/backend/vs2010backend.py
+++ b/mesonbuild/backend/vs2010backend.py
@@ -27,7 +27,7 @@ from .. import mlog
 from .. import compilers
 from ..interpreter import Interpreter
 from ..mesonlib import (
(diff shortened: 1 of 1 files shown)


dcbaker:
Is this a regression, of just something we've always done wrong?

GustavoLCR:
> Is this a regression, of just something we've always done wrong?

I believe it was always wrong

dcbaker:
The windows and cygwin failures look relevant

GustavoLCR:
> The windows and cygwin failures look relevant

@dcbaker I fixed those failures, the others seem to be in master as well

dcbaker:
Great, looks good, thanks.

PR #9045, opened 2021-07-30
Escape path in exclude filter passed to gcovr

alois31:
Gcovr interprets exclude filters, as passed to the -e option, as
regexes. Since we want to exclude a raw path, the argument must be
escaped.

### DIFF
--- a/mesonbuild/scripts/coverage.py
+++ b/mesonbuild/scripts/coverage.py
@@ -14,7 +14,7 @@
 
 from mesonbuild import environment, mesonlib
 
-import argparse, sys, os, subprocess, pathlib, stat
+import argparse, re, sys, os, subprocess, pathlib, stat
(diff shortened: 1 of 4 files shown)


mensinda:
Could you add a test case for this (probably somewhere [unittests/allplatformstests.py](https://github.com/mesonbuild/meson/blob/master/unittests/allplatformstests.py#L3331) in would be a good idea).

mensinda (approved):
LGTM, let's see if our CI also agrees.

mensinda:
@alois31 This error is related to this PR: https://github.com/mesonbuild/meson/pull/9045/checks?check_run_id=3233594244

The other errors are unrelated and also happening on master.

mensinda:
Thanks for fixing this.

PR #9048, opened 2021-07-30
compilers: Add default search path stdlib_only_link_flags

dcbaker:
This should be done in all cases of language_stdlib_only_link_flags, but
I don't have access to all of the compilers to test it.

This is required in cases where object files created by gfortran are
linked using another compiler with a different default search path, such
as gfortran and clang together. And also on nixos, where no compilers
implementation libs are in another compiler's path.

### DIFF
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -1425,7 +1425,7 @@ You probably should put it in link_with instead.''')
         # If the user set the link_language, just return that.
         if self.link_language:
             comp = all_compilers[self.link_language]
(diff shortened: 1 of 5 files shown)


dcbaker:
@scivision: I don't have flang (or an easy way to install it) or the PGI compiler or ifort setup at the moment, could you at least check flang for me? I'd be super grateful if you could help with the PGI/nvidia compiler :)

matthew-brett:
Anything I can do here to help with this PR?

PR #9051, opened 2021-07-31
Fix python module leaving PKG_CONFIG_LIBDIR set in os.environ.

xclaesse:


xclaesse:
@dcbaker this is a regression introduced by 4d67dd19e5b7dcec6716840d30984fa41eef55c6, if old_pkg_libdir is None you don't pop it from env anymore.

mesonbuild/modules/python.py
```diff
@@ -247,6 +247,7 @@ def wrap_in_pythons_pc_dir(name: str, env: 'Environment', kwargs: T.Dict[str, T.
             try:
                 return PythonPkgConfigDependency(name, env, kwargs, installation)
             finally:
+                os.environ.pop('PKG_CONFIG_LIBDIR', None)
```
dcbaker, line 250: I really dislike using pop here since we're not taking the valuie, and pop is really more about move than deleting I guess we could write as:
```python
if old_pkg_libdir is not None:
  os.environ['PKG_CONFIG_LIBDIR'] = old_pkg_libdir
elif pkg_libdir:
  del os.environ['PKG_CONFIG_LIBDIR']
```

Really though this manual cleanup is really a pain, and I wish os.environ just has a "with os.environ.temp({'foo': 'bar'}:` that would just do the cleanup automatically and not really that nice.

xclaesse: ok, removed usage of pop here.

dcbaker (approved):
LGTM, thanks

nirbheek:
https://github.com/mesonbuild/meson/commit/4d67dd19e5b7dcec6716840d30984fa41eef55c6 never made it into 0.59.0, so this is not a release regression afaict. Removing milestone.

xclaesse:
@nirbheek ah you're right, I got tricked by git log because it seems we branched 0.59 before final release.

Issue #8140, opened 2020-12-29
Can't define cpp_std to latest version with multi-platform support


jmoguillansky-gpsw:
**Describe the bug**
Try to set cpp_std to c++latest.  The problem is that msvc toolchain recognizes "c++latest" while gcc toolchain recognizes "c++2a".  Also, can't define project default_options using conditional logic.  
Meson complains that the first statement needs to be project statement.  

**To Reproduce**
cpp_std = ''
host_system = host_machine.system()
cpp = meson.get_compiler('cpp')
if cpp.get_id() == 'msvc'
  cpp_std = 'c++latest'
else
  cpp_std = 'c++2a'
endif

project(
'testproj',
[ 'c', 'cpp' ],
default_options: 'cpp_std='+cpp_std
...
)

**Expected behavior**
Meson should support defining cpp_std to latest version in a way that's platform agnostic.  
Also, meson should support setting cpp_std using conditional logic.

**system parameters**
Cross build : Windows + MSVC, Linux + GCC

willwray:
Still an issue on v0.59.1 - there should be a portable way to select c++20 for a project, in 2021...

I hit it a month ago, independently came up with the same conditional platform-specific hack,
and visited the meson IRC to raise it there too.

HolyWu:
According to https://devblogs.microsoft.com/cppblog/msvc-cpp20-and-the-std-cpp20-switch/, MSVC has added the /std:c++20 switch in 16.11. Currently I still get the following error using Meson v0.59.1.

`meson.build:1:0: ERROR: Value "c++20" (of type "string") for combo option "C++ language standard to use" is not one of the choices. Possible choices are (as string): "none", "c++11", "vc++11", "c++14", "c++latest", "vc++latest", "vc++14", "c++17", "vc++17".`

Looking at https://github.com/mesonbuild/meson/blob/173a40be330c1aedec9f7f99ebca67a5edd25c3d/mesonbuild/compilers/cpp.py#L566 and https://github.com/mesonbuild/meson/blob/173a40be330c1aedec9f7f99ebca67a5edd25c3d/mesonbuild/compilers/cpp.py#L668, the code is not updated to support c++20 for MSVC yet.

seishun:
Meson now supports C++20 for Visual Studio, but this is still an issue for C++23 and C++26.

dcbaker:
With more recent versions of Meson you can specify an array of choices, and Meson will use the first supported one:
```meson
[...]

PR #9053, opened 2021-07-31
Add support for c++2b and gnu++2b cpp_stds

Joshua-Ashton:
Add support for the work in progress C++23 support in GCC and Clang

Signed-off-by: Joshua Ashton <joshua@froggi.es>

### DIFF
--- a/mesonbuild/compilers/cpp.py
+++ b/mesonbuild/compilers/cpp.py
@@ -209,8 +209,8 @@ class ClangCPPCompiler(ClangCompiler, CPPCompiler):
         })
         opts[key.evolve('std')].choices = [
             'none', 'c++98', 'c++03', 'c++11', 'c++14', 'c++17', 'c++1z',
-            'c++2a', 'c++20', 'gnu++11', 'gnu++14', 'gnu++17', 'gnu++1z',
-            'gnu++2a', 'gnu++20',
+            'c++2a', 'c++20', 'c++2b', 'gnu++11', 'gnu++14', 'gnu++17', 'gnu++1z',
+            'gnu++2a', 'gnu++20', 'gnu++2b',
         ]
         if self.info.is_windows() or self.info.is_cygwin():
             opts.update({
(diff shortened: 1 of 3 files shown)


dcbaker:
This generally looks good to me, can you rebase so that CI can rerun and turn green?

dcbaker (approved):

Joshua-Ashton:
Done.

Joshua-Ashton:
Seems to fail because the CI's packages are too outdated to have those C++stds in the compilers...

Not super sure what to do about that given it shouldn't really matter.

dcbaker:
The unittest itself has a set of "c++11 needs at least this version of gcc, c++14 needs this one". It's in unittests/linuxliketest.py, in a method called `_test_stds_impl` you just need to update that for the new standards.

Joshua-Ashton:
Thanks

willwray:
#8140
Still lacking a portable, cross platform, way to specify a cpp_std to select c++20 for a project.

chriselrod:
Any updates on this?

mesonbuild/compilers/cpp.py
```diff
@@ -354,8 +354,8 @@ def get_options(self) -> 'KeyedOptionDictType':
         })
         opts[key].choices = [
             'none', 'c++98', 'c++03', 'c++11', 'c++14', 'c++17', 'c++1z',
-            'c++2a', 'c++20', 'gnu++03', 'gnu++11', 'gnu++14', 'gnu++17',
-            'gnu++1z', 'gnu++2a', 'gnu++20',
+            'c++2a', 'c++20', 'c++2b', 'gnu++03', 'gnu++11', 'gnu++14', 'gnu++17',
+            'gnu++1z', 'gnu++2a', 'gnu++20', 'gnu++2b',
```
eli-schwartz, line 358: TBH I wonder if we should do what we do for example with e.g. VisualStudioCPPCompiler and IntelCPPCompiler, and do a bit of `if version_compare(self.version, '{added_version}'):` to decide whether to add support for these stds.

chriselrod: It would be nice to get this in for meson v1.0 if it's not too late. 

Would that allow automatically supporting new standards the compiler supports?

theoparis:
> Any updates on this?

Bumping this because I currently have to use this PR branch instead of upstream meson, which causes gtk to fail to build because the version is too old 😕

PR #9054, opened 2021-08-01
fix bat_info UTF-8 string error

xth:
" bat_info = json.loads(bat_json) " may produce error
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc3 ...
Because the vswhere.exe's output is not utf-8 by default
Use UTF-8 encoding for vswhere.exe can fixing it .


jpakkane:
Thanks.

Issue #9036, opened 2021-07-27
Dependency Cache Regression from 0.58.2 to 0.59.0


nosracd:
**Describe the bug**
In version `0.59.0` `dependency` results are cached improperly. On a clean `meson build` it passes with no errors. On a reconfigure, the configure fails.

**To Reproduce**
On Ubuntu, have the `googletest` package installed but not `libgtest-dev` (`googletest` gets installed when `libgtest-dev` in installed but does not automatically get removed when `libgtest-dev` is removed).

Using the following `meson.build`:
```
project('foo', 'cpp')

gtest_dep = dependency('gtest', version: '>=1.8.0')
```

with the following wrap file at `subprojects/gtest.wrap`:
```
[wrap-file]
directory = googletest-release-1.10.0

source_url = https://github.com/google/googletest/archive/release-1.10.0.zip
source_filename = gtest-1.10.0.zip
source_hash = 94c634d499558a76fa649edb13721dce6e98fb1e7018dfaeba3cd7a083945e91

patch_url = https://wrapdb.mesonbuild.com/v1/projects/gtest/1.10.0/1/get_zip
patch_filename = gtest-1.10.0-1-wrap.zip
patch_hash = 04ff14e8880e4e465f6260221e9dfd56fea6bc7cce4c4aff0dc528e4a2c8f514
```

Then configure a build directory using 
```
meson build
```
Then reconfigure using
```
meson build --reconfigure
```

**Expected behavior**
Using `0.58.2` results in expected behavior where `meson build` and `meson build --reconfigure` yield the same result. 

`meson build` results in
```
Dependency gtest found: NO unknown version, but need: ['>=1.8.0']
Run-time dependency GTest found: YES (building self)
```
and `meson build --reconfigure` results in the same output:
```
Dependency gtest found: NO unknown version, but need: ['>=1.8.0']
Run-time dependency GTest found: YES (building self)
```

**Actual behavior**
On `0.59.0`, the first `meson build` results in:
```
Dependency gtest found: NO unknown version, but need: ['>=1.8.0']
Run-time dependency GTest found: YES (building self)
```
but the `meson build --reconfigure` results in
```
Dependency gtest found: NO found unknown but need: '>=1.8.0' (cached)

meson.build:3:0: ERROR: Dependency 'gtest' is required but not found.
```

**system parameters**
* Is this a [cross build](https://mesonbuild.com/Cross-compilation.html) or just a plain native build (for the same computer)?
[...]

PR #9056, opened 2021-08-02
Fix when gtest dep is found but does not pass version check

xclaesse:
GTestDependencySystem (and other similar dep classes) sets
self.is_found=True, but the version check could still fail. In the case
the dependency is not required `ExternalDependency._check_version()`
won't raise an exception and thus the dependency is accepted.

_check_version() sets self.is_found() in the case self.version is not
empty, we should do it too when self.version is empty.

Fixes: #9036.


xclaesse:
Note that with 0.58, without gtest wrap, `dependency('gtest', required: false, version: '>=1.8')` will use gtest from `/usr/src/googletest/googletest/src` but `dependency('gtest', version: '>=1.8')` aborts with not-found dependency, so it was already unconsistent.

PR #9059, opened 2021-08-03
Add type checking and annotations for run_target

dcbaker:
This includes some basic cleanups, and a conversion of `func_run_target` to use `typoed_kwargs`, and the `build.RunTarget` to stop processing keyword arguments itself. This is my first step in untagling the build and interpreter target creation from each other.


dcbaker:
These are wrong, so I'm going to close this an try again later.

PR #9060, opened 2021-08-03
destdir: Allow relative to build directory

xclaesse:
Meson already works like that, except in do_copydir() that requires
absolute destdir. Better explicitly support that instead of leaving it
undefined and unconsistent.


mensinda (approved):
LGTM and the CI errors seem unrelated

Issue #8050, opened 2020-11-29
override_dependency and dependency(..., static: true) broken


textshell:
When using `override_dependency` and a `dependency(...)` call that uses `static: true` the overridden dependency seems to be completely ignored.

Reproducer:

main project:

```
project('parent', 'c')
dependency('sub', static: true)
```

sub project:
```
project('sub', 'c')
dep = declare_dependency()
meson.override_dependency('sub', dep)
```

This yields the following output:
```
Run-time dependency sub found: NO (tried pkgconfig and cmake)
Looking for a fallback subproject for the dependency sub

|Executing subproject sub method meson 
|
|Project name: sub
|Project version: undefined
|C compiler for the host machine: ccache cc (gcc 8.3.0 "cc (Debian 8.3.0-6) 8.3.0")
|C linker for the host machine: cc ld.bfd 2.31.1
|Build targets in project: 0
|Subproject sub finished.

Dependency sub from subproject subprojects/sub found: NO

meson.build:4:0: ERROR: Subproject subprojects/sub did not override dependency sub
```

This should at least give a hint, that and how the overriding was found to be incompatible. I had to resort to adding debug prints to meson to understand why my override seemed to be completely ignored, so think trying to understand this will be frustrating to other users of meson as well.

But i think if the overriden dependency contains a static lib it should just work.

Tested using meson master from 24.11.2020 (0fa808f7066d)

/CC @xclaesse

ntonnaett:
This is still a problem. Is the underlying problem that there is default_library for subprojects and static: bool for dependencies?
Maybe default_library shouldn't be used at all for this. I would like to have a way to specify easily what part of my dependency chain is linked statically or shared. Right now there is no way to have a dependency that should be linked statically but all dependencies of that should be linked shared. Or I am not aware of that.

What do you think about automatically adding dependency options?

xclaesse:
There are 2 issues here:
- `dependency('sub', static: true)` is not going to pass `default_library=static` to the fallback. So your subproject is going to build shared libraries, unless you also add `default_options:['default_library=static']`. IMHO, `static: true` should implicitly set the `default_library=static` option on fallback subproject.
[...]

PR #9061, opened 2021-08-04
interpreter: Fix dependency(..., static: true) fallback

xclaesse:
It should build the fallback subprject with default_library=static and
override the dependency for both static=True and static kwarg not given.

Fixes: #8050.

### DIFF
--- a/mesonbuild/interpreter/mesonmain.py
+++ b/mesonbuild/interpreter/mesonmain.py
@@ -14,11 +14,16 @@ from ..interpreterbase import (MesonInterpreterObject, FeatureNew, FeatureDeprec
 from .interpreterobjects import (ExecutableHolder, ExternalProgramHolder,
                                  CustomTargetHolder, CustomTargetIndexHolder,
                                  EnvironmentVariablesObject)
+from .type_checking import NATIVE_KW
 
 import typing as T
 
 if T.TYPE_CHECKING:
     from .interpreter import Interpreter
+    from typing_extensions import TypedDict
+    class FuncOverrideDependency(TypedDict):
+        native: mesonlib.MachineChoice
+        static: T.Optional[bool]
 
 class MesonMain(MesonInterpreterObject):
     def __init__(self, build: 'build.Build', interpreter: 'Interpreter'):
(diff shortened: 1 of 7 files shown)


jpakkane:
⚔️ conflicts.

xclaesse:
> crossed_swords conflicts.

fixed.

mesonbuild/interpreter/mesonmain.py
```diff
@@ -294,21 +295,52 @@ def override_find_program_method(self, args, kwargs):
             raise InterpreterException('Second argument must be an external program or executable.')
         self.interpreter.add_find_program_override(name, exe)
 
+    @typed_kwargs('meson.override_dependency', NATIVE_KW,
+                  KwargInfo('static', bool, since='0.60.0'))
+    @typed_pos_args('meson.override_dependency', str, dependencies.Dependency)
     @FeatureNew('meson.override_dependency', '0.54.0')
-    @permittedKwargs({'native'})
     def override_dependency_method(self, args, kwargs):
```
dcbaker, line 302:
```suggestion
    def override_dependency_method(self, args: T.Tuple[str, dependencies.Dependency], kwargs: 'FuncOverrideDependency') -> None:
```

Plus add:
```python
from typing_extensions import TypedDict

class FuncOverrideDependency(TypedDict):
    native: mesonlib.MachineChoice
    static: T.Optional[bool]
```
inside the `if T.TYPE_CHECKING:` block.

mesonbuild/interpreter/mesonmain.py
```diff
@@ -294,21 +295,52 @@ def override_find_program_method(self, args, kwargs):
...
-        dep = args[1]
-        if not isinstance(name, str) or not name:
+        name, dep = args
+        if not name:
             raise InterpreterException('First argument must be a string and cannot be empty')
-        if not isinstance(dep, dependencies.Dependency):
-            raise InterpreterException('Second argument must be a dependency object')
+
+        optkey = OptionKey('default_library', subproject=self.interpreter.subproject)
+        default_library = self.interpreter.coredata.get_option(optkey)
```
dcbaker, line 308: for mypy can you add:
```python
assert isinstance(default_library, str), 'for mypy'
```

mesonbuild/interpreter/mesonmain.py
```diff
@@ -294,21 +295,52 @@ def override_find_program_method(self, args, kwargs):
...
+            # override regardless of the static value here. But do not raise error
+            # if it has already been overridden, which would happend when overriding
+            # static and shared separately:
+            # meson.override_dependency('foo', shared_dep, static: false)
+            # meson.override_dependency('foo', static_dep, static: true)
+            # In that case dependency('foo') would return the first override.
+            self._override_dependency_impl(name, dep, kwargs, static=None, permissive=True)
+            self._override_dependency_impl(name, dep, kwargs, static=static)
+
+    def _override_dependency_impl(self, name, dep, kwargs, static, permissive=False):
```
dcbaker, line 333:
```suggestion
    def _override_dependency_impl(self, name: str, dep: dependencies.Dependency, kwargs: 'FuncOverrideDependency', static: T.Optional[bool], permissive: bool = False) -> None:
```

mesonbuild/interpreter/mesonmain.py
```diff
@@ -294,21 +295,52 @@ def override_find_program_method(self, args, kwargs):
...
+        if static is None:
+            kwargs.pop('static', None)
+        else:
+            kwargs['static'] = static
```
dcbaker, line 337: This seems sketchy, you're mutating the state that's going to be re-used in the calling context. I feel like we should be copying here before mutating. Also, we know static is going to be in here, so just use `del`

dcbaker (changes requested):
This seems like a nice change. I've suggested some type annotations. There's also some state mutation going on that seems dangerous to me.

mesonbuild/interpreter/mesonmain.py
xclaesse: done.

mesonbuild/interpreter/mesonmain.py
xclaesse: done.

mesonbuild/interpreter/mesonmain.py
xclaesse: static is not always there because we could have removed it in a previous _override_dependency_impl() call. I changed now to always do a copy, in that case we are sure the `static` kwarg is always there and we can use del.

PR #9063, opened 2021-08-04
Split keyword definitions from interpreter into their own module

dcbaker:
This also fixes the type checking of these keyword definitions.

@xclaesse, this should fix the issues you're seeing if you rebase on this

### DIFF
--- a/mesonbuild/interpreterbase/decorators.py
+++ b/mesonbuild/interpreterbase/decorators.py
@@ -323,7 +323,7 @@ class KwargInfo(T.Generic[_T]):
                  deprecated: T.Optional[str] = None,
                  deprecated_values: T.Optional[T.Dict[str, str]] = None,
                  validator: T.Optional[T.Callable[[_T], T.Optional[str]]] = None,
-                 convertor: T.Optional[T.Callable[[_T], TYPE_var]] = None,
+                 convertor: T.Optional[T.Callable[[_T], object]] = None,
                  not_set_warning: T.Optional[str] = None):
         self.name = name
         self.types = types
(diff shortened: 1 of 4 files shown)


mesonbuild/interpreterbase/decorators.py
```diff
@@ -323,7 +323,7 @@ def __init__(self, name: str, types: T.Union[T.Type[_T], T.Tuple[T.Type[_T], ...
                  deprecated: T.Optional[str] = None,
                  deprecated_values: T.Optional[T.Dict[str, str]] = None,
                  validator: T.Optional[T.Callable[[_T], T.Optional[str]]] = None,
-                 convertor: T.Optional[T.Callable[[_T], TYPE_var]] = None,
+                 convertor: T.Optional[T.Callable[[_T], object]] = None,
```
mensinda, line 326: Is there a specific reason for why this is required or is this is a debugging leftover?

xclaesse:
LGTM when CI is done. Thanks!

mesonbuild/interpreterbase/decorators.py
dcbaker: No, it's needed. The convertor function is in no way bound to return a TYPE_var, in fact of the three uses 2 don't return a TYPE_var. One of them converts the `bool` of the `native` keyword argument into a `MachineChoice`, the second converts the `install_mode` keyword argument into a `FileMode` object.

mensinda: Ok, fair enough.

mensinda (approved):
LGTM, when the CI is green.

Issue #9065, opened 2021-08-06
`--force-fallback-for` not affecting `--wrap-mode=nofallback`


branchvincent:
**Describe the bug**

The [docs](https://mesonbuild.com/Subprojects.html#commandline-options) state:

> --force-fallback-for=list,of,dependencies
>
> This option takes precedence over --wrap-mode=nofallback

This has regressed in 0.59.

**To Reproduce**
```console
$ cat <<EOF >meson.build
project('a')
b = dependency('b', fallback: 'b')
EOF
$ mkdir -p subprojects/b
$ echo "project('b')" >subprojects/b/meson.build
$ meson setup ./build --wrap-mode=nofallback --force-fallback-for=b
...
Not looking for a fallback subproject for the dependency b because:
Use of fallback dependencies is disabled.
```

**Expected behavior**
As observed with 0.58:

```console
Looking for a fallback subproject for the dependency b because:
Use of fallback was forced for that specific subproject
```

**system parameters**
* Is this a [cross build](https://mesonbuild.com/Cross-compilation.html) or just a plain native build (for the same computer)? native
* what operating system (e.g. MacOS Catalina, Windows 10, CentOS 8.0, Ubuntu 18.04, etc.) MacOS Big Sur
* what Python version are you using? 3.9.6
* what `meson --version` 0.59.0
* what `ninja --version` 1.10.2

Presumably, the issue is here: https://github.com/mesonbuild/meson/blob/5381ec89afdbf61e575cb0ac5c25e339279a5af2/mesonbuild/interpreter/dependencyfallbacks.py#L108-L114

xclaesse:
Thanks for detailed bug report, I indeed inverted the conditions while refactoring that part of the code during 0.59 cycle. Fixed in https://github.com/mesonbuild/meson/pull/9071.

PR #9071, opened 2021-08-06
Fix --force-fallack-for with --wrap-mode=nofallback

xclaesse:
Fixes: #9065


mensinda (approved):
LGTM, and the CI failures are unrelated

PR #9074, opened 2021-08-08
i18n: use real build/install targets for gmo files

eli-schwartz:
Don't just create a .PHONY target which runs a script that magically generates files ninja doesn't know about. It results in untracked files, and `meson install` has to run additional commands instead of copying over files, and then cannot track them to uninstall them later.

I'm not even really sure why it was originally done via a proxy script, most likely bad legacy design. This is after all one of the oldest modules...

One side effect of this is that meson doesn't know how to rename build.CustomTarget files on install (only data files are supported?), and every file needs to be installed as "domainname.mo" so it must be named that in-tree too. To prevent clashes, every locale gets its own locale-specific subdirectory.

Once we are doing that anyway, we can output them to the actual structure required by the gettext family of functions, and bindtextdomain() can therefore point to this location if desired. This might be useful for running localized programs from the build tree.

### DIFF
--- a/mesonbuild/modules/i18n.py
+++ b/mesonbuild/modules/i18n.py
@@ -17,6 +17,7 @@ import shutil
 from os import path
 from .. import coredata, mesonlib, build, mlog
 from ..mesonlib import MesonException
+from ..scripts.gettext import read_linguas
 from . import ModuleReturnValue
 from . import ExtensionModule
 from ..interpreterbase import permittedKwargs, FeatureNew, FeatureNewKwargs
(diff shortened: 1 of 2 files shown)


eli-schwartz:
(It's a very nice feeling to improve the code by dropping 2 dozen lines.)

eli-schwartz:
It's not clear to me why it used to run do_install, then manually return 0 instead of returning do_install in that one branch. :/

eli-schwartz:
CI failures are all irrelevant (and occur on git master too).

dcbaker (approved):
LGTM. thanks for doing this, I've been meaning to do it myself (I'd like to have these as build targets rather than run targets in Mesa).

eli-schwartz:
After rebasing, there is another irrelevant CI failure:

```
test cases/osx/4 framework/meson.build:24:4: ERROR: No host machine compiler for "stat.c"
```

PR #9075, opened 2021-08-08
compilers/cpp: Set the MSVC macro to make the __cplusplus define

dcbaker:
accurate

Otherwise it always returns the value for c++98.

TODO: needs test
TODO: needs docs

### DIFF
--- a/docs/markdown/snippets/msvc_cplusplus_define.md
+++ b/docs/markdown/snippets/msvc_cplusplus_define.md
@@ -1,0 +1,7 @@
+## MSVC now sets the __cplusplus #define accurately
+
+For reasons, MSVC will always return `199711L` for `__cplusplus`, even when a
+newer c++ standard is explicitly requested, unless you pass a specific option to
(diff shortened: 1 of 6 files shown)


jenatali:
Do you support older versions of VS? That link says:

> The /Zc:__cplusplus option is available starting in Visual Studio 2017 version 15.7

So you might need a version check rather than an always.

dcbaker:
Yeah, we do support older versions, so we'll need to check the version. I missed that

jenatali (approved):
LGTM

jpakkane:
This has a bug where the command line option does not persist over a regeneration and causes the noop CI failure (it was previously masked by the Rust failure).

To reproduce:

- configure `6 linkshared` in the VS shell
- verify that `/Zc:__cplusplus` is in the Ninja file
- `ninja`
- touch the main `meson.build`
- `ninja reconfigure`
- verify that `/Zc:__cplusplus` is no longer on the command line

jpakkane:
This had to be reverted from master to bring CI to green. Please reopen.

Issue #7638, opened 2020-08-25
WAF build system compatibility


pwnorbitals:
As a follow-up to CMake / make subprojects compatibility, I would like to issue a feature request to be able to use WAF projects as subprojects (https://waf.io/). 

For example in my case, I need to use libCSP (https://github.com/libcsp/libcsp) and sadly I'm stuck. Afaik, my only option would be to re-create the entire build using `meson.build` files

mensinda:
#6920 could be made to work with WAF (cc @xclaesse)

xclaesse:
What are the commands to configure and build? Especially to pass prefix, libdir, etc.

mensinda:
From the looks of https://github.com/libcsp/libcsp/blob/master/wscript and https://waf.io/book/ there is only `--prefix` and building is done by `waf configure <opts>` and `waf build`.

Note: I didn't know WAF existed before reading this issue and I only spent a few minutes glancing over the docs and trying to understand the libcsp wafscript.

xclaesse:
What's not clear is whether is supports out of tree build.

rubyFeedback:
> What's not clear is whether is supports out of tree build.

I think many waf users do not know either. ;-)

The more momentum meson/ninja has, the better, IMO, because for me as a user
it is simpler to work with meson/ninja than with e. g. waf or scons (mpv is still using
waf; right now they seem to consider contemplating to moving a custom build 
system since waf is not so ideal. Custom build systems are also not great, so you
sort of trade-in and trade-out different disadvantages. So the smoother it may be
to transition into meson/ninja, the better IMO for people who rather prefer to 
work with meson. There are not that many waf-based projects, so knowledge 
about it and online documentation is more limited than compared to, say, 
GNU configure, cmake or meson).

xclaesse:
The minimum requirement to use the approach from #6920 are:
- Out of tree build.
- Separate configure/build steps (currently build must be `make` but that can be easily extended)
- Take common directories in configure command line: prefix/libdir/etc
- Read basic env as CFLAGS, CC (if needed I can extend to pass them in command line).
- Remember those env passed at configure step when doing build step (if needed I can extend to set that env at build time too).
- Build step should be able to reconfigure itself when needed.

Any half sane build system would do that, no idea if waf does. I'm not really interested in digging myself, but if someone who knows waf can give me example command lines to configure/build out of tree, I can give it a try.

averne:
[...]

PR #9076, opened 2021-08-09
external-project: Add support for WAF build system

xclaesse:
Fixes: #7638

### DIFF
--- a/docs/markdown/External-Project-module.md
+++ b/docs/markdown/External-Project-module.md
@@ -63,6 +63,10 @@ directory and executable. Note that if a bootstrap script is required
 (e.g. `autogen.sh` when building from git instead of tarball), it can
 be done using `run_command()` before calling `add_project()` method.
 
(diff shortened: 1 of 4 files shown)


tristan957:
Not that I am at all familiar with WAF, but does it make sense to have a `type` kwarg for external project so someone could be explicit about what alternative build system they are using?

`type: 'autotools'`
`type: 'waf'`

xclaesse:
> Not that I am at all familiar with WAF, but does it make sense to have a `type` kwarg for external project so someone could be explicit about what alternative build system they are using?

Hmm, I haven't seen the need for that yet. I think using the positional argument for that works well atm, don't you think? Note that we could still just change the API later if needed, it's unstable module after all.

tristan957:
Yeah can always be changed later. Just being forward thinking that maybe another build system could exist that has the same command as WAF or Autotools. Because you only get a chance to design a good API once!

Issue #9022, opened 2021-07-22
i18n.merge_file fails to substitute @BASENAME@ when generating target if input is an object


AdrianVovk:
**Describe the bug**
`i18n.merge_file` creates targets with the name `@BASENAME@_merge` when the input is an object. If `i18n.merge_file` is called twice with `output: '@BASENAME@'` set and `input: <some object>`, meson complains that there are two targets with the same name. This was originally fixed in #3946, but a special case has since been added for when `input` is an object.

I'm guessing something about 34c28dc92cb4f28e32e633297e4b8a737228aae5 broke this again. The comment [there](https://github.com/mesonbuild/meson/commit/34c28dc92cb4f28e32e633297e4b8a737228aae5#diff-e9d99e08d876e79e48b22f23823c3fa4b1ba48d852daa2186982a480ffa4e8c3R111) even says:
`# I have no idea why/how this if isinstance(inputfile, mesonlib.HoldableObject) works / used to work...`

**To Reproduce**
[Here's](https://gitlab.gnome.org/GNOME/gcr/-/blob/master/ui/meson.build) the file that triggered the discovery of this bug. Here's the meson config snippet that breaks it:
```
i18n.merge_file(
  input: configure_file(
    input: 'gcr-viewer.desktop.in.in',
    output: '@BASENAME@',
    configuration: { 'bindir': gcr_prefix / get_option('bindir') },
  ),
  output: '@BASENAME@',
  type: 'desktop',
  po_dir: podir,
  install: true,
  install_dir: get_option('datadir') / 'applications',
)

i18n.merge_file(
  input: configure_file(
    input: 'gcr-prompter.desktop.in.in',
    output: '@BASENAME@',
    configuration: { 'libexecdir': gcr_prefix / get_option('libexecdir') },
  ),
  output: '@BASENAME@',
  type: 'desktop',
  po_dir: podir,
  install: true,
  install_dir: get_option('datadir') / 'applications',
)
```

**Expected behavior**
The build doesn't fail, and the target name is generated after variable substitution

**system parameters**
* Plain native build
* OS: carbonOS 2021.1
* Python version: 3.9.6
* Meson version: 0.59.0-rc2
* Ninja version: 1.10.2

xclaesse:
@AdrianVovk can you please give a try to https://github.com/mesonbuild/meson/pull/9078?

PR #9078, opened 2021-08-09
Fix i18n target name when using @BASENAME@ and configure_file() input

xclaesse:
Fixes: #9022


test cases/frameworks/6 gettext/data/meson.build
```diff
@@ -27,4 +27,31 @@ i18n.merge_file(
   install_dir: join_paths(get_option('datadir'), 'applications')
 )
 
+# Regression test: https://github.com/mesonbuild/meson/issues/9022
```
dcbaker, line 30: Could you add specifically "passing a File object as input", so that the next person doesn't have to open github to figure this out?

dcbaker (approved):
I have one very tiny request to clarify a comment, but otherwise this looks fine to me

test cases/frameworks/6 gettext/data/meson.build
xclaesse: ok, done.

PR #9083, opened 2021-08-09
interpreter: Fix spurious warning in configure_data()

xclaesse:
This is a regression when porting to typed_pos_args().


eli-schwartz:
Thanks.

The commit this is fixing up isn't present in 0.59 though, nor is it a current candidate for the maintenance release, so this should not be milestoned for 0.59.1

eli-schwartz (approved):

xclaesse:
> The commit this is fixing up isn't present in 0.59 though

oh, I thought it was released already with that PR. My bad.

PR #9090, opened 2021-08-10
Type annotations and fixes for backend/backends

dcbaker:
My goal was to add type annotations to backend/backends.py, as it seemed like a good starting point for looking into the backend. This ended up requiring a lot of fixes for the build module as well, and revealed several issues in already typed modules, like mesonlib. I've split each change into a single patch, as this is good practice, and makes bisecting easier for regressions.

### DIFF
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -500,7 +500,7 @@ class Backend:
                                        exe_wrapper, workdir,
                                        extra_paths, capture, feed, tag)
 
-    def as_meson_exe_cmdline(self, tname, exe, cmd_args, workdir=None,
+    def as_meson_exe_cmdline(self, exe, cmd_args, workdir=None,
                              extra_bdeps=None, capture=None, feed=None,
                              force_serialize=False,
                              env: T.Optional[build.EnvironmentVariables] = None,
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -149,9 +149,18 @@ class SubdirInstallData(InstallDataBase):
         self.exclude = exclude
 
 class ExecutableSerialisation:
-    def __init__(self, cmd_args, env: T.Optional[build.EnvironmentVariables] = None, exe_wrapper=None,
-                 workdir=None, extra_paths=None, capture=None, feed=None,
-                 tag: T.Optional[str] = None) -> None:
+
+    # XXX: should capture and feed default to False, instead of None?
+
+    def __init__(self, cmd_args: T.List[str],
+                 env: T.Optional[build.EnvironmentVariables] = None,
+                 exe_wrapper: T.Optional['programs.ExternalProgram'] = None,
+                 workdir: T.Optional[str] = None,
+                 extra_paths: T.Optional[T.List] = None,
+                 capture: T.Optional[bool] = None,
+                 feed: T.Optional[bool] = None,
+                 tag: T.Optional[str] = None,
+                 ) -> None:
         self.cmd_args = cmd_args
         self.env = env
         if exe_wrapper is not None:
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -48,7 +48,7 @@ if T.TYPE_CHECKING:
 LANGS_CANT_UNITY = ('d', 'fortran', 'vala')
 
 class RegenInfo:
-    def __init__(self, source_dir, build_dir, depfiles):
+    def __init__(self, source_dir: str, build_dir: str, depfiles: T.List[str]):
         self.source_dir = source_dir
         self.build_dir = build_dir
         self.depfiles = depfiles
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -45,6 +45,16 @@ if T.TYPE_CHECKING:
     from ..mesonlib import FileMode, FileOrString
     from ..wrap import WrapMode
 
+    from typing_extensions import TypedDict
+
+    class TargetIntrospectionData(TypedDict):
+
+        language: str
+        compiler : T.List[str]
+        parameters: T.List[str]
+        sources: T.List[str]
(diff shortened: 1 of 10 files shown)


mesonbuild/backend/backends.py
```diff
@@ -135,13 +153,24 @@ def __init__(self, path: str, install_path: str, install_mode: 'FileMode', subpr
...
         self.exclude = exclude
 
 class ExecutableSerialisation:
-    def __init__(self, cmd_args, env: T.Optional[build.EnvironmentVariables] = None, exe_wrapper=None,
-                 workdir=None, extra_paths=None, capture=None, feed=None) -> None:
+
+    # XXX: should capture and feed default to False, instead of None?
+
+    def __init__(self, cmd_args: T.List[str],
+                env: T.Optional[build.EnvironmentVariables] = None,
```
mensinda, line 166: bad indentation (and `cmd_args` could also go on a newline).

mesonbuild/backend/backends.py
```diff
@@ -267,13 +299,17 @@ def get_compiler_options_for_target(self, target: build.BuildTarget) -> OptionOv
...
         if option_name in target.option_overrides_base:
             override = target.option_overrides_base[option_name]
-            return self.environment.coredata.validate_option_value(option_name, override)
-        return self.environment.coredata.get_option(option_name.evolve(subproject=target.subproject))
+            v = self.environment.coredata.validate_option_value(option_name, override)
+        else:
+            v = self.environment.coredata.get_option(option_name.evolve(subproject=target.subproject))
+        # We don't actually have wrapmode here to do an assert, so just do a
+        # cast, we know what's in coredata anyway.
+        return T.cast(T.Union[str, int, bool, 'WrapMode'], v)
```
mensinda, line 310: This should be more a more noticeable todo to remove this cast once we have annotated coredata.

mesonbuild/backend/backends.py
```diff
@@ -291,7 +327,7 @@ def get_build_dir_include_args(self, target, compiler, *, absolute_path=False):
                 curdir = '.'
         return compiler.get_include_args(curdir, False)
 
-    def get_target_filename_for_linking(self, target):
+    def get_target_filename_for_linking(self, target: T.Union[build.Target, build.CustomTargetIndex]) -> T.Optional[str]:
```
mensinda, line 330: Is it really out of the question that not a `build.CustomTarget` can be passed here? There is an `isinstance` check for it below.

Same with the other functions below.

mesonbuild/backend/backends.py
```diff
@@ -820,9 +905,9 @@ def generate_basic_compiler_args(self, target: build.BuildTarget, compiler: 'Com
         # command-line or default_options inside project().
         commands += compiler.get_option_compile_args(copt_proxy)
         # Add buildtype args: optimization level, debugging, etc.
-        commands += compiler.get_buildtype_args(self.get_option_for_target(OptionKey('buildtype'), target))
-        commands += compiler.get_optimization_args(self.get_option_for_target(OptionKey('optimization'), target))
-        commands += compiler.get_debug_args(self.get_option_for_target(OptionKey('debug'), target))
+        commands += compiler.get_buildtype_args(T.cast(str, self.get_option_for_target(OptionKey('buildtype'), target)))
+        commands += compiler.get_optimization_args(T.cast(str, self.get_option_for_target(OptionKey('optimization'), target)))
+        commands += compiler.get_debug_args(T.cast(bool, self.get_option_for_target(OptionKey('debug'), target)))
```
mensinda, line 910: I would prefer asserts here. There is no reason that the types should ever accidentally change, but they do for some reason we will be glad that the asserts tell us early where the problem is.

mesonbuild/backend/backends.py
```diff
@@ -910,46 +995,49 @@ def get_mingw_extra_paths(self, target):
                 paths.update(cc.get_library_dirs(self.environment))
         return list(paths)
 
-    def determine_windows_extra_paths(self, target: T.Union[build.BuildTarget, str], extra_bdeps):
-        '''On Windows there is no such thing as an rpath.
+    def determine_windows_extra_paths(
+            self, target: T.Union[build.BuildTarget, build.CustomTarget, programs.ExternalProgram, mesonlib.File, str],
```
mensinda, line 999: I don't really know the code but from the current annotations this looks like a `build.CustomTargetIndex` or `build.GeneratedList` might also be passed here.

mesonbuild/backend/backends.py
```diff
@@ -1337,7 +1427,8 @@ def create_install_data(self) -> InstallData:
                         self.environment.get_build_dir(),
                         self.environment.get_prefix(),
                         strip_bin,
-                        self.environment.coredata.get_option(OptionKey('install_umask')),
+                        # This will only be a string or number, but mypy doesn't know that.
+                        T.cast(T.Union[str, int], self.environment.coredata.get_option(OptionKey('install_umask'))),
```
mensinda, line 1431: Again, an assert will be more code, but we really want this to blow up if something goes wrong.

mensinda (commented):
I only got a few issues, but this LGTM otherwise.

mesonbuild/backend/backends.py
dcbaker: Custom target is a Target (though not a BuildTarget), but CustomTargetIndex is its own thing.

mensinda: Oh, right. Makes sense.

mesonbuild/backend/backends.py
dcbaker: I don't think that we can properly annotate this, get_option returns  different type depending on the exact input, and I don't think there's a good way to annotate that, but I can add a better comment.

mesonbuild/backend/backends.py
dcbaker: I don't see any path where that could happen

mensinda (approved):
LGTM

PR #9092, opened 2021-08-10
Add unset_variable()

tristan957:
This should be useful for helping to control variable scope within
Meson. CMake has something similar for controlling scope.

### DIFF
--- a/mesonbuild/interpreter/interpreter.py
+++ b/mesonbuild/interpreter/interpreter.py
@@ -364,6 +364,7 @@ class Interpreter(InterpreterBase, HoldableObject):
                            'static_library': self.func_static_lib,
                            'both_libraries': self.func_both_lib,
                            'test': self.func_test,
+                           'unset_variable': self.func_unset_variable,
                            'vcs_tag': self.func_vcs_tag,
                            'range': self.func_range,
                            })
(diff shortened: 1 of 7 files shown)


tristan957:
Obviously amenable to any changes y'all might have.

xclaesse:
TIL we have `is_variable()`, seems logical addition and implementation looks good to me.

@jpakkane any objection?

mesonbuild/interpreter/interpreter.py
```diff
@@ -2638,6 +2639,16 @@ def func_get_variable(self, node: mparser.BaseNode, args: T.Tuple[T.Union[str, D
     def func_is_variable(self, node: mparser.BaseNode, args: T.Tuple[str], kwargs: 'TYPE_kwargs') -> bool:
         return args[0] in self.variables
 
+    @typed_pos_args('unset_variable', str)
+    @noKwargs
+    @noArgsFlattening
```
dcbaker, line 2644: This doesn't seem necessary

mesonbuild/interpreter/interpreter.py
```diff
@@ -2638,6 +2639,16 @@ def func_get_variable(self, node: mparser.BaseNode, args: T.Tuple[T.Union[str, D
...
+    def func_unset_variable(self, node: mparser.BaseNode, args: T.Tuple[str], kwargs: 'TYPE_kwargs') -> None:
+        varname = args[0]
+        if varname not in self.variables:
+            raise InterpreterException(f'Tried to unset unknown variable "{varname}".')
+
+        del self.variables[varname]
+
```
dcbaker, line 2651: I think it would be better to do this with a try/except. That will make the success path faster, and we don't care in the failure case, because that's fatal

dcbaker (changes requested):

mesonbuild/interpreter/interpreter.py
```diff
@@ -2638,6 +2639,15 @@ def func_get_variable(self, node: mparser.BaseNode, args: T.Tuple[T.Union[str, D
     def func_is_variable(self, node: mparser.BaseNode, args: T.Tuple[str], kwargs: 'TYPE_kwargs') -> bool:
         return args[0] in self.variables
 
+    @typed_pos_args('unset_variable', str)
+    @noKwargs
```
dcbaker, line 2643:
```python
@FeatureNew('unset_variable', '0.60.0')
```

tristan957: Whoops totally forgot about that. Will add it right now!

dcbaker (approved):

xclaesse:
Lgtm

jpakkane:
Please rebase against current master so the CI is rerun. It should be all green now.

PR #8527, opened 2021-03-12
[CUDA] Bugfix: Improve NVCC cuda_std flags handling when it differs from cpp_std

obilaniu:
Closes #8523 , plus minor cleanups.

@dcbaker, can this also be marked for backport to 0.57.2?

Issue #8523, opened 2021-03-11
Combining default_options : ['cpp_std=c++2a', 'cuda_std=c++17'] breaks nvcc


drewrobson:
I'm trying to compile a project where the host code (*.cpp) uses c++20, but the device code (*.cu) uses c++17, since that is the latest version currently supported by nvcc.

**Describe the bug**
In `meson.build`, I tried this:
```
project('C++ std', 'cpp', 'cuda', version : '1.0.0', default_options : ['cpp_std=c++2a', 'cuda_std=c++17'])
```
but `nvcc` fails because it was called with `--std=c++17 -std=c++2a`:

```
meson compile -v
ninja: Entering directory `.'
[1/2] nvcc -Xcompiler=-Wall -Xcompiler=-Winvalid-pch -Xcompiler=-Wnon-virtual-dtor --std=c++17 -std=c++2a -g -G -g -I.. -I. -Iprog.p -Iprog.p -o prog.p/main.cu.o -c ../main.cu
FAILED: prog.p/main.cu.o 
nvcc -Xcompiler=-Wall -Xcompiler=-Winvalid-pch -Xcompiler=-Wnon-virtual-dtor --std=c++17 -std=c++2a -g -G -g -I.. -I. -Iprog.p -Iprog.p -o prog.p/main.cu.o -c ../main.cu
nvcc fatal   : Value 'c++2a' is not defined for option 'std'
ninja: build stopped: subcommand failed.
```

**To Reproduce**
Start with cuda example [6 std](https://github.com/mesonbuild/meson/tree/master/test%20cases/cuda/6%20std), but use this:
```project('C++ std', 'cpp', 'cuda', version : '1.0.0', default_options : ['cpp_std=c++2a', 'cuda_std=c++17'])```

**Expected behavior**
I expected that `gcc` would be invoked with c++2a and `nvcc` would be invoked with c++17.

If I manually remove the spurious `-std=c++2a` then `nvcc` succeeds.

**system parameters**
This is a native build on 5.4.101-1-MANJARO with Meson 0.57.1.

obilaniu:
@drewrobson Can you have a try at my solution? It solves it on my end (except with c++17 and 14; I don't have c++2a)

drewrobson:
Yep, that solves it on my end as well, with c++2a and c++17. Thank you for fixing this so quickly!

PR #8396, opened 2021-02-21
[CUDA] Bugfix: Forward sanitizer_*_args() methods to host compiler.

obilaniu:
Enables `-Db_sanitize=undefined` and company.

Also serves as a testcase for NVCC comma-shielding: Because the test-case declares `b_sanitize=address,undefined`, the host GCC compiler needs `-fsanitize=address,undefined`, but this stands a danger of being split by NVCC when wrapped with `-Xcompiler=args,args`. Special, already-existing comma-shielding codepaths activate to prevent this splitting.

Closes #8394.

PR #9098, opened 2021-08-11
Disable asan memory leak checks on a Cuda test.

jpakkane:
Should fix Arch Cuda CI failure.

### DIFF
--- a/test cases/cuda/16 multistd/meson.build
+++ b/test cases/cuda/16 multistd/meson.build
@@ -1,4 +1,7 @@
-project('C++-CUDA multi-std', 'cpp', 'cuda', version : '1.0.0', default_options : ['cpp_std=c++17', 'cuda_std=c++14'])
+project('C++-CUDA multi-std', 'cpp', 'cuda',
+        version : '1.0.0',
+        default_options : ['cpp_std=c++17', 'cuda_std=c++14'])
 
 exe = executable('prog', 'main.cu')
-test('cudatest', exe)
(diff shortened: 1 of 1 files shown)


eli-schwartz:
@obilaniu since you've touched the cuda module a lot, perhaps you know why this started to fail to begin with?

obilaniu:
@eli-schwartz Hard to say! I *did* touch #8396 concerning the sanitizers, but that was CUDA test case 15. This is test case 16 concerning support for multiple CUDA standards, reported #8523 and fixed in #8527 / #8596.

That was 3 months ago. At the time, I seem to recall the CUDA Arch runner was using GCC 10 and CUDA 11.2; Now I see it is using GCC 11 and CUDA 11.3. Perhaps mixed C++ standards and sanitizers play badly, or the new GCC or CUDA Toolkit expose a problem?

eli-schwartz:
When I was trying to figure out why it failed, I tried running multiple cuda test cases locally using meson.py setup builddir. They all produced asan erroring executables, but only one test case actually executes the program in a testsuite.

So it's either a general problem with the cuda toolkit itself, or a problem with all our example program sources. I really don't think it's about any PR specifically.

I was hoping maybe you knew some answer like "yeah, current versions of cuda are just buggy".

obilaniu:
@eli-schwartz It looks like `libcuda` does weird things that play poorly with ASAN.

- https://github.com/google/sanitizers/issues/629
- https://github.com/NVIDIA/DALI/pull/362

PR #9099, opened 2021-08-11
Test boost-python on macOS

kjmeagher:
Previously the meson test case would only test boost-python on linux.
With the #7909 it is now possible to use boost-python on macOS/homebrew.
This enables the boost-python test on both linux and macOS.
It also uses python.extension_module() instead of shared_library to make the
python extension module.


jpakkane:
CI has been fixed. Please rebase so that tests are rerun.

kjmeagher:
I rebased and all tests are green

PR #9106, opened 2021-08-12
Use `typed_kwargs` and `typed_pos_args` for CustomTarget

### LINKED ISSUE #9096: IndexError with `custom_target(…, output: [], build_by_default: true)` in a `subdir()`
**Describe the bug**

`custom_target(…, output: [], build_by_default: true)` in a `subdir()` causes a `IndexError`.

```
  File "/usr/lib/python3.9/site-packages/mesonbuild/mesonmain.py", line 227, in run
  File "/usr/lib/python3.9/site-packages/mesonbuild/msetup.py", line 281, in run
    app.generate()
  File "/usr/lib/python3.9/site-packages/mesonbuild/msetup.py", line 184, in generate
    self._generate(env)
  File "/usr/lib/python3.9/site-packages/mesonbuild/msetup.py", line 246, in _generate
    intr.backend.generate()
  File "/usr/lib/python3.9/site-packages/mesonbuild/backend/ninjabackend.py", line 559, in generate
    self.generate_ending()
  File "/usr/lib/python3.9/site-packages/mesonbuild/backend/ninjabackend.py", line 3209, in generate_ending
    targetlist.append(os.path.join(self.get_target_dir(t), t.get_outputs()[0]))
IndexError: list index out of range
```

Omitting `build_by_default: true` or doing this in the project root does not cause a `IndexError`. However `ninja.build` seems to have a syntax error.

```
ninja: error: build.ninja:31: expected path
build : CUSTOM_COMMAND ../sub1/infile | /bin/true
      ^ near here
WARNING: Could not create compilation database.
ninja: Entering directory `_builddir'
ninja: error: build.ninja:31: expected path
build : CUSTOM_COMMAND ../sub1/infile | /bin/true
      ^ near here
```

**To Reproduce**

```console
$ mkdir -p test01/sub1
$ cd test01
$ touch sub1/infile
$ cat meson.build
project('test01')

subdir('sub1')
$ cat sub1/meson.build
custom_target('index-error', command: ['/bin/true'], input: 'infile', output: [], build_by_default: true)
$ meson setup _builddir && meson compile -C _builddir
```

**Expected behavior**

If `output` must not be empty a error message like `src/meson.build:42:0: ERROR: output must not be empty`.

**system parameters**
  native
  Fedora 34
  3.9.6
* what `meson --version`
  0.59.0
  1.10.2

dcbaker:
This is a rather lengthy series on it's own, but it's built on top of #9090. I tried to rebase it directly on master, and gave up.

This was originally suppsoed to be part of a larger series to fix the layering violations between the build and interpreter layers, but that turns out to be a *lot* of work, for a lot of reasons. To simplify things, and because this fixes an actual bug, I split some of the work off into it's own series.

Fixes #9096

### DIFF
--- a/mesonbuild/modules/qt.py
+++ b/mesonbuild/modules/qt.py
@@ -556,14 +556,14 @@ class QtBaseModule(ExtensionModule):
             else:
                 outdir = state.subdir
             cmd = [self.tools['lrelease'], '@INPUT@', '-qm', '@OUTPUT@']
-            lrelease_kwargs = {'output': '@BASENAME@.qm',
-                               'input': ts,
-                               'install': kwargs['install'],
-                               'install_tag': 'i18n',
-                               'build_by_default': kwargs['build_by_default'],
-                               'command': cmd}
-            if install_dir is not None:
-                lrelease_kwargs['install_dir'] = install_dir
+            lrelease_kwargs: T.Dict[str, T.Any] = {
+                'output': '@BASENAME@.qm',
+                'input': ts,
+                'install': kwargs['install'],
+                'install_dir': install_dir or [],
+                'install_tag': 'i18n',
+                'build_by_default': kwargs['build_by_default'],
+                'command': cmd}
             lrelease_target = build.CustomTarget(f'qt{self.qt_version}-compile-{ts}', outdir, state.subproject, lrelease_kwargs)
             translations.append(lrelease_target)
         if qresource:
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -2226,6 +2226,11 @@ class BothLibraries(SecondLevelHolder):
         raise MesonBugException(f'self._preferred_library == "{self._preferred_library}" is neither "shared" nor "static".')
 
 class CommandBase:
+
+    depend_files: T.List[File]
+    dependencies: T.List[T.Union[BuildTarget, 'CustomTarget']]
+    subproject: str
+
     def flatten_command(self, cmd):
         cmd = listify(cmd)
         final_cmd = []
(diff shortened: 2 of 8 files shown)


mensinda:
More typing is always good! I will have look at this once #9090 is in and the CI is green.

jpakkane:
⚔️ conflicts.

dcbaker:
@mensinda, I think this is still green after rebase :D

mesonbuild/modules/qt.py
```diff
@@ -556,14 +556,14 @@ def compile_translations(self, state: 'ModuleState', args: T.Tuple, kwargs: 'Com
...
             cmd = [self.tools['lrelease'], '@INPUT@', '-qm', '@OUTPUT@']
-            lrelease_kwargs = {'output': '@BASENAME@.qm',
-                               'input': ts,
-                               'install': kwargs['install'],
-                               'install_tag': 'i18n',
-                               'build_by_default': kwargs['build_by_default'],
-                               'command': cmd}
-            if install_dir is not None:
-                lrelease_kwargs['install_dir'] = install_dir
+            lrelease_kwargs: T.Dict[str, T.Any] = {
```
mensinda, line 559: I think this could be `T.Dict[str, TYPE_var]` (or even better a typed dict)

mensinda (approved):
This is basically LGTM, besides the one minor point below.

mesonbuild/modules/qt.py
dcbaker: CustomTarget currently takes `T.Dict[str, T.Any]` for its initializer, so we'd have to change that too. I've got another series built on this that is moving towards not passing a dictionary into CustomTarget, but keyword arguments, since all the argument parsing should be happening in the interpreter.

mensinda: Fine by me then.

PR #9107, opened 2021-08-12
trivial lint fixes

eli-schwartz:
The trailing whitespace one is iffy, but since we can actually make this be standardized...


mensinda (approved):

eli-schwartz:
This passed in CI, then failed after being merged via a fast-forward merge. The same commit hash has two test runs for "Ubuntu Rolling (gcc, g++)" and the second CI run has "LinuxCrossMingwTests::test_exe_wrapper_behaviour" failing due to:

```
E           RuntimeError: There is no current event loop in thread 'Dummy-1'.

/usr/lib/python3.9/asyncio/events.py:642: RuntimeError
```

(╯°□°)╯︵ ┻━┻

Issue #7007, opened 2020-04-21
[RFC] Install-level tags


lb90:
It would be nice if installed files could be tagged in order to support selective installing. For example, installed headers could be tagged with `tags: '#dev'`, installed doc-related files tagged with `tags: '#doc'`, . Then the user could select the install level with `meson install --tags base,i18n`.

Why is that useful?

The use case I'm thinking of here is deployment of applications, where development files are not needed. But that would also be useful for distro packagers to create split packages.

eli-schwartz:
This is something I've pondered would be nice too, but I'd instead suggest letting the "install" kwarg which appears in various places, instead of only being boolean true/false, also be a string, and if it's a string e.g. `install: 'foo'`, then it would be added to a target `install-foo` which is also installed by the toplevel install.

meson install --tags seems like it should be a non-starter, that just forwards to ninja install and meson cannot arbitrarily add a --tags flag to ninja.

jpakkane:
> meson install --tags seems like it should be a non-starter, that just forwards to ninja install 

It's actually the other way around. `ninja install` calls `meson install` with default arguments. That is the same as `ninja test` etc. The default target does the default action, fine tuning is done by calling the Meson command and passing whatever extra args you need.

lb90:
Is it ok if I implement it?

jpakkane:
Thinking out loud some:

- No hashtags, these should be plain strings.
- `tags` is not a very good kwarg name, as it does not specify much, maybe something like `install_groups`?
- There should probably be a way to not install some groups, such as `meson install --no-groups devtools`

xclaesse:
I made a similar proposal there: https://github.com/mesonbuild/meson/issues/7246. It can be seen as a special-case of this tag/group proposal. What I like about special casing a "runtime" group is that for small projects it will often do the right thing by default without having to tag manually every target.

dcbaker:
@xclaesse, that could be made to work by default by making the default install_group/tag/whatever `runtime` for `install_header`, `install_man`, `pkgconfig.generate`, `cmake.generate`, etc, right?

eli-schwartz:
We could go even further and automatically generate targets similar to the ones autotools would classically do.

install-binPROGRAMS
install-binSCRIPTS
install-dist_binSCRIPTS
install-includeHEADERS
install-libLTLIBRARIES
install-nodist_includeHEADERS
install-pkgconfigDATA

Admittedly we can probably choose better names than the ones autotools chose.

[...]

Issue #7863, opened 2020-10-17
Expose a PEP 517 backend


FFY00:
Python packaging is shifting away from setuptools as the only real build backend option. [PEP 517](https://www.python.org/dev/peps/pep-0517/) introduces a standard interface to insteract with build backends, [PEP 518](https://www.python.org/dev/peps/pep-0518/) introduces a standard way to specify which backend the project uses.

I think it would be interesting to see if meson could expose a Python build system backend, so that it could be used in Python packaging.

Would this be something you might consider?

jpakkane:
I have not looked at the details, but in theory yes.

dcbaker:
Id like to do it. The one trick is that we'd probably need a toml parser, which isn't in the standard library (yet).

lazka:
There is https://gitlab.com/thiblahute/mesonpep517

FFY00:
> Id like to do it. The one trick is that we'd probably need a toml parser, which isn't in the standard library (yet).

Great, I would also like to help out.

The toml parser would be the easy part, we can pull `toml` as an optional dependency.

PEP 517 mandates two hooks, `build_sdist` and `build_wheel`.

`build_sdist` build a source distribution. There isn't a standard for this yet, so we essentially just have to produce a tarball with the source code, which is pretty easy.

`build_wheel` is the one that requires work. It builds a Python wheel file, which is described in [PEP 427](https://www.python.org/dev/peps/pep-0427/).

A quick overview, a Python wheel is a zip file and its contents look something like this:
```
python_module1
python_module2
my_package.dist-info
my_package.data
```

`my_package.dist-info` is the metadata directory and `my_package.data` the data directory. The data directory can contain headers, scripts and general data, each in its own subdirectory.

Before starting working on this, everyone should read the whole PEP.

> There is [gitlab.com/thiblahute/mesonpep517](https://gitlab.com/thiblahute/mesonpep517)

I am aware, but it does not give me much confidence.

![mesonpep517-releases](https://user-images.githubusercontent.com/11718923/96369084-d34c1780-114f-11eb-8711-b613b95b157c.png)

lazka:
> I am aware, but it does not give me much confidence.

Why is that?

FFY00:
> Why is that?

Did you see the picture I attached? That does not seem like a reasonable release model, IMO.

The author is also inactive, at least in the recent months. Here's an issue I'd really like to see addressed.
[...]

PR #9110, opened 2021-08-13
Add install tags

xclaesse:
Fixes: #7007.

### DIFF
--- a/docs/markdown/Installing.md
+++ b/docs/markdown/Installing.md
@@ -74,7 +74,7 @@ giving an absolute install path.
 install_data(sources : 'foo.dat', install_dir : '/etc') # -> /etc/foo.dat
 ```
 
-## Custom install behavior
+## Custom install script
 
 Sometimes you need to do more than just install basic targets. Meson
 makes this easy by allowing you to specify a custom script to execute
@@ -133,23 +133,57 @@ command in the build tree:
 $ meson install --no-rebuild --only-changed
 ```
 
-## Finer control over install locations
+## Installation tags
 
-Sometimes it is necessary to only install a subsection of output files
-or install them in different directories. This can be done by
-specifying `install_dir` as an array rather than a single string. The
-array must have as many items as there are outputs and each entry
-specifies how the corresponding output file should be installed. For
-example:
+*Since 0.60.0*
 
-```meson
-custom_target(...
-    output: ['file1', 'file2', 'file3'],
-    install_dir: ['path1', false, 'path3'],
-    ...
-)
-```
+It is possible to install only a subset of the installable files using
+`meson install --tags tag1,tag2` command line. When `--tags` is specified, only
+files that have been tagged with one of the tags are going to be installed.
 
-In this case `file1` would be installed to `/prefix/path1/file1`,
-`file2` would not be installed at all and `file3` would be installed
-to `/prefix/path3/file3'.
+This is intended to be used by packagers (e.g. distributions) who typically
+want to split `libfoo`, `libfoo-dev` and `libfoo-doc` packages. Instead of
+duplicating the list of installed files per category in each packaging system,
+it can be maintained in a single place, directly in upstream `meson.build` files.
+
+Meson sets predefined tags on some files. More tags are likely to be added over
+time, please help extending the list of well known categories.
+- `devel`:
+  * `static_library()`,
+  * `install_headers()`,
+  * `pkgconfig.generate()`,
+  * `gnome.generate_gir()` - `.gir` file,
+  * Files installed into `libdir` and with `.a` or `.pc` extension,
+  * File installed into `includedir`.
+- `runtime`:
+  * `executable()`,
+  * `shared_library()`,
+  * `shared_module()`,
+  * `jar()`,
+  * Files installed into `bindir`.
+  * Files installed into `libdir` and with `.so` or `.dll` extension.
+- `python-runtime`:
+  * `python.install_sources()`.
+- `man`:
+  * `install_man()`.
+- `doc`:
+  * `gnome.gtkdoc()`,
+  * `hotdoc.generate_doc()`.
+- `i18n`:
+  * `i18n.gettext()`,
+  * `qt.compile_translations()`,
+  * Files installed into `localedir`.
+- `typelib`:
+  * `gnome.generate_gir()` - `.typelib` file.
+
+Custom installation tag can be set using the `install_tag` keyword argument
+on various functions such as `custom_target()`, `configure_file()`,
+`install_subdir()` and `install_data()`. See their respective documentation
+in the reference manual for details. It is recommended to use one of the
+predefined tags above when possible.
+
+Installable files that have not been tagged either automatically by Meson, or
+manually using `install_tag` keyword argument won't be installed when `--tags`
+is used. They are reported at the end of `<builddir>/meson-logs/meson-log.txt`,
+it is recommended to add missing `install_tag` to have a tag on each installable
+files.
--- a/docs/markdown/Reference-manual.md
+++ b/docs/markdown/Reference-manual.md
@@ -304,6 +304,9 @@ false otherwise.
   string, the file is not installed.
 - `install_mode` *(since 0.47.0)*: specify the file mode in symbolic format
   and optionally the owner/uid and group/gid for the installed files.
+- `install_tag` *(since 0.60.0)*: A string used by `meson install --tags` command
+  to install only a subset of the files. By default the file has no install
+  tag which means it is not being installed when `--tags` argument is specified.
 - `output`: the output file name. *(since 0.41.0)* may contain
   `@PLAINNAME@` or `@BASENAME@` substitutions. In configuration mode,
   the permissions of the input file (if it is specified) are copied to
@@ -378,7 +381,7 @@ following.
 - `install_dir`: If only one install_dir is provided, all outputs are installed there.
    *Since 0.40.0* Allows you to specify the installation directory for each
     corresponding output. For example:
-    ```
+    ```meson
     custom_target('different-install-dirs',
       output : ['first.file', 'second.file'],
       install : true,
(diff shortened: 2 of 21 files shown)


tristan957:
Should `jar()` have a default tag?

docs/markdown/Reference-manual.md
```diff
@@ -1964,6 +1976,9 @@ the following methods.
   can be specified. If `true` the script will not be run if DESTDIR is set during
   installation. This is useful in the case the script updates system wide
   cache that is only needed when copying files into final destination.
+  *(since 0.60.0)* `install_tag` string keyword argument can be specified.
```
tristan957, line 1979: Could make sense to mention the default tag value in these types of comments

tristan957:
Can you add multiple tags to `install_tag`?

xclaesse:
> Should `jar()` have a default tag?

Currently it's tagged `runtime` like all other buildtarget (except static lib), forgot it in the doc. Is that correct?

> Can you add multiple tags to `install_tag`?

No, but custom_target() can have one tag per output. The idea is you would put multiple tags at installation, like `--tags runtime,man` because man page often goes with the binary.

anarazel:
It would be nice if the set of tags that are installed by default could be adjusted. In a lot of projects doc generation isn't something all that many people need to do and it's often a very slow step. Perhaps a project level option specifying the default install tags?

xclaesse:
> It would be nice if the set of tags that are installed by default could be adjusted. In a lot of projects doc generation isn't something all that many people need to do and it's often a very slow step. Perhaps a project level option specifying the default install tags?

Install tags only affect the install step, not configure/build steps. For example most GNOME projects have `gtk_doc` option to enable/disable build of documentation because that's slow. I think adding builtin options is out of scope here.

docs/markdown/Reference-manual.md
xclaesse: The next line is:
> By default the script has no install tag which means it is not being run when `meson install --tags` argument is specified.

I think I specify the default everywhere, no?

anarazel:
> Install tags only affect the install step, not configure/build steps.

So if I do meson install --tags=fast-stuff it'll also (re-)build things that aren't dependencies?

> For example most GNOME projects have gtk_doc option to enable/disable build of documentation because that's slow. I think adding builtin options is out of scope here.

I think having build-time options like gtk_doc is a proper solution to reduce dependencies. However using options to control whether docs are built or not regardless of dependencies seems like a workaround for not having more fine grained targets for things like install.

xclaesse:
> So if I do meson install --tags=fast-stuff it'll also (re-)build things that aren't dependencies?

Yes. `meson install` always rebuilds everything (unless `--no-rebuild`), it actually has no way to map an installable file to a ninja target, it does not have that information. We could/should do that, but I think it's complicated and out of scope here.

The intended use-case is to split installed files during packaging. You would still need to build and install all files, just separately so you can make `-dev`, `-doc`, etc packages.

xclaesse:
That being said, `gnome.gtkdoc` is actually special, it builds the doc in an install script. That script will be run only if you give `--tags doc` (or no tags). So in this specific case, yes you would save the time of building the doc. But that's really a weird implementation detail of gtkdoc.

anarazel:
>Yes. meson install always rebuilds everything (unless --no-rebuild), it actually has no way to map an installable file to a ninja target, it does not have that information. We could/should do that, but I think it's complicated and out of scope here.

Ah, right. I forgot about that (it's been a few months since I hacked on my meson port). Consider my comments withdrawn / wishes for future features.

tristan957:
Great question on `jar()`. I don't really feel like I am the right person to answer the question. Intuitively I think JARs can be both compile time (devel) or runtime.

eli-schwartz:
> That being said, `gnome.gtkdoc` is actually special, it builds the doc in an install script. That script will be run only if you give `--tags doc` (or no tags). So in this specific case, yes you would save the time of building the doc. But that's really a weird implementation detail of gtkdoc.

By "special", you mean broken, right? :p 

Nothing should ever be getting built outside of ninja, and the fact that gtkdoc does so causes extremely weird edge cases, like building it as root, or failing to build target dependencies. causing gtkdoc itself to fail.

xclaesse:
> > Should `jar()` have a default tag?
> 
> Currently it's tagged `runtime` like all other buildtarget (except static lib), forgot it in the doc. Is that correct?

Added it explicitly in the doc as `runtime`, but I'm not a java developer myself, I'll change if someone think it's wrong.

mesonbuild/backend/backends.py
```diff
@@ -1354,6 +1362,27 @@ def create_install_data_files(self):
         with open(install_data_file, 'wb') as ofile:
             pickle.dump(self.create_install_data(), ofile)
 
+    def guess_install_tag(self, fname, outdir = None):
```
dcbaker, line 1365: I'm trying to have this fully typed, so can we add full annotations here?

I think that should be:
```python
def guess_install_tag(self, fname: str, outdir: T.Optional[str] = None) -> T.Optional[str]:
```

mesonbuild/backend/backends.py
```diff
@@ -1354,6 +1362,27 @@ def create_install_data_files(self):
...
+        prefix = self.environment.get_prefix()
+        bindir = Path(prefix, self.environment.get_bindir())
+        libdir = Path(prefix, self.environment.get_libdir())
+        incdir = Path(prefix, self.environment.get_includedir())
+        localedir = Path(prefix, self.environment.coredata.get_option(mesonlib.OptionKey('localedir')))
+        dest_path = Path(prefix, outdir, Path(fname).name) if outdir else Path(prefix, fname)
+        if bindir in dest_path.parents:
```
dcbaker, line 1372: I think this should be:
```python
if dest_path.is_relative_to(bindir):
```

mesonbuild/build.py
```diff
@@ -2339,10 +2344,25 @@ def process_kwargs(self, kwargs, backend):
...
+                self.install_tag = kwargs.get('install_tag', [])
+                if isinstance(self.install_tag, str):
+                    self.install_tag = [self.install_tag] * len(self.outputs)
+                else:
+                    self.install_tag = stringlistify(self.install_tag)
+                    missing = len(self.outputs) - len(self.install_tag)
+                    if missing < 0:
+                        raise InvalidArguments('install_tag list cannot have more items than outputs')
+                    if missing > 0:
+                        self.install_tag += [None] * missing
```
dcbaker, line 2360: Please no, this is going to make used `typed_kwargs` for this miserable, since `foo` and `[foo]` will have different meanings, which Meson generally does not do. Can we instead do something like:
```python
install_tag = extract_as_list(kwargs, 'install_tag')
if len(install_tag) == 1:
   self.install_tag = [install_tag] * len(install_tag)
elif len(install_tag) != len(self.outputs):
   # raise an error message
else:
  self.install_tag = install_tag
```

If you want something to be untagged you could pass `false` like we do for `install_dir`, so at least those are symaterical, but we're also not providing different behavior for `['devel']` and `'devel'`. I think this is important because Meson generally does not distinguish between an scalar and an array of length 1.

dcbaker (changes requested):
In general this looks good, but I have a real concern that you're making `install_tag` of a single string different than an array with one string. Meson doesn't do that anywhere that I'm aware except for the `message` family of functions, and that's a discrepancy that I think we should avoid creating.

mesonbuild/backend/backends.py
xclaesse: You live too far into the future: `New in version 3.9.` ;-)

mesonbuild/backend/backends.py
xclaesse: done.

mesonbuild/build.py
xclaesse: You're right, better stick to the same behaviour as `install_dir` which was my initial intention. I changed that code, what do you think?

mesonbuild/build.py
```diff
@@ -2339,10 +2344,19 @@ def process_kwargs(self, kwargs, backend):
                 # the list index of that item will not be installed
                 self.install_dir = typeslistify(kwargs['install_dir'], (str, bool))
                 self.install_mode = kwargs.get('install_mode', None)
+                # If only one tag is provided, assume all outputs have the same tag.
+                # Otherwise, we must have as much tags as outputs.
+                self.install_tag = typeslistify(kwargs.get('install_tag', [None]), (str, bool))
+                if len(self.install_tag) == 1:
+                    self.install_tag = self.install_tag * len(self.outputs)
+                if len(self.install_tag) != len(self.outputs):
```
dcbaker, line 2352: this could safely be elif, to avoid running the check, right?

dcbaker (approved):
I've got one teensy nit, but otherwise this looks good. Thanks for making changes.

mesonbuild/build.py
xclaesse: Yep, fixed.

docs/markdown/Installing.md
```diff
@@ -133,23 +133,55 @@ command in the build tree:
...
+*Since 0.60.0*
+
+It is possible to install only a subset of the installable files using
+`meson install --tags tag1,tag2` command line. When `--tags` is specified, only
+files that have been tagged with one of the tags are going to be installed.
+
+This is intended to be used by packagers (e.g. distributions) who typically
+want to split `libfoo`, `libfoo-dev` and `libfoo-doc` packages. Instead of
+duplicating the list of installed files per category in each packaging system,
+it can be maintained in a single place, directly in upstream `meson.build` files.
```
jon-turney, line 147: This seems to assume that all distros will agree about what package a file ends up in.  I'm not sure that's true.
 
I'm not sure I understand how this is supposed to be used by a distro packaging tool? Run the install step with each of a set of tags into a clean DESTDIR, and package the files which end up there?  How is the tool supposed to learn the set of tags needed to span all installed files?

xclaesse: Note that here I mean packagers in a larger sense than just distro. For example in my current project we deploy the runtime as a tarball we extract on devices, that "packaging" is easier to do if I can just tell meson to install only runtime files in my DESTDIR, then I can just tar that dir and copy on devices.

I'm not expecting distro like debian to adopt this system any time soon, they already have all their tooling based on maintaining list of files for each binary package. But if they do, yes they would run install into different DESTDIRs for runtime, devel, etc.

I defined a set of well known tags in the doc, I think most projects should stick that that list. But I also saw use-cases for more specific projects (not typically packaged in distro) that would want to define their own split.

mesonbuild/modules/python.py
```diff
@@ -442,8 +442,7 @@ def install_sources_method(self, args: T.Tuple[T.List[T.Union[str, mesonlib.File
         return self.interpreter.install_data_impl(
             self.interpreter.source_strings_to_files(args[0]),
             self._get_install_dir_impl(kwargs['pure'], kwargs['subdir']),
-            mesonlib.FileMode(),
-            None)
+            mesonlib.FileMode(), rename=None, tag='runtime')
```
eli-schwartz, line 445: Noticed this in the latest push... shouldn't this allow specifying a user-set tag? Maybe the tag should be python-related, this would also be convenient for marking installable files for consumption by mesonpep517, cross-reference #7863

xclaesse: I should add `install_tag` kwarg there too, indeed. Maybe I should change the default to `python-runtime` to make them installable separately? I'm not sure, because my understanding is you could install a python executalbe to `/usr/bin` (that would be tagged as `runtime`) and it requires its modules to be installed by `install_sources()` ?

xclaesse: Added `install_tag` kwarg to `python.install_sources()`.

eli-schwartz: In that case, one would need to manually tag the python executable as `install_tag: 'python-runtime'`

xclaesse: Changed the default to `python-runtime` after discussion on IRC.

jpakkane:
I was just going to say that this is ok featurewise but apparently was too slow. :P

PR #9094, opened 2021-08-11
Make the flex test case work

Kekun:
This adds the noyywrap option so flex doesn't wait for more input once
we reached EOF. This also adds the nounput and noinput options to fix
compilation warnings.

We can now run the test as expected.

PR #9113, opened 2021-08-14
Force unity builds off in Flex test.

jpakkane:


eli-schwartz:
Context: flex and unity builds aren't playing well together as discovered here: https://github.com/mesonbuild/meson/pull/9094#issuecomment-898036660

PR #9116, opened 2021-08-15
Refresh Ninja cache files on regeneration.

jpakkane:


mesonbuild/backend/ninjabackend.py
```diff
@@ -567,6 +567,9 @@ def generate(self):
         # fully created.
         os.replace(tempfilename, outfilename)
         mlog.cmd_ci_include(outfilename)  # For CI debugging
+        # Refresh Ninja's caches. https://github.com/ninja-build/ninja/pull/1685
+        if os.path.exists('.ninja_deps'):
+            subprocess.call(self.ninja_command + ['-t', 'restat'])
```
mensinda, line 572: From the linked PR, it looks like this requires ninja 1.10. As far as I understand we only require ninja 1.8.2 atm, so this should be behind a version check.

Issue #2370, opened 2017-09-24
Warning color is unreadable on white background


adrianjohnson:
When printing warnings, the word 'WARNING' is printed in bright yellow. This is unreadable on a white background. I suggest orange would be a better color to use.

[meson-color.txt](https://github.com/mesonbuild/meson/files/1327221/meson-color.txt)

nirbheek:
Perhaps it's better to not assume what the background colour of the user is, and supply our own for warnings and errors?

adrianjohnson:
Yes, that would be better.
[meson-color.txt](https://github.com/mesonbuild/meson/files/1332846/meson-color.txt)

shevegen:
I myself usually use KDE konsole, black background, white text. (Same for other terminals).

I do however also completely understand the people who use the reverse or similar.

Ideally I guess there may be different colour profiles and people could cherry-pick whichever variant they prefer, but this is perhaps not a minor issue compared to other situations - to me, bright yellow versus orange, on a black background, is almost the same anyway so I would not mind a change.

You could however had also ask why warnings are not in ... red colour! :D But I think to some extent it's arbitrary anyway whatever colour is picked.

PR #9117, opened 2021-08-15
Make warning color readable on a white background

adrianjohnson:
Use standard yellow instead of bright yellow. Standard yellow is readable on black, white, light yellow and solarized light backgrounds.

Fixes: #2370

### DIFF
--- a/mesonbuild/mlog.py
+++ b/mesonbuild/mlog.py
@@ -166,7 +166,7 @@ def green(text: str) -> AnsiDecorator:
     return AnsiDecorator(text, "\033[1;32m")
 
 def yellow(text: str) -> AnsiDecorator:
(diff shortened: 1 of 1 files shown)


mensinda:
I would say that this is more an issue with your color theme, since all combinations of standard color codes should be clearly readable. Furthermore, I don't think that it is a good idea to drop the bold font for warnings, since it also effects the font and not only the color.

A quick fix for you could be to change your theme so that the bold yellow color equals your normal yellow color.

bonzini:
I agree this is a problem with the color theme having different colors for normal yellow and bold yellow; if anything, bold should make the yellow text slightly _more_ visible if they were the same. Closing.

PR #9121, opened 2021-08-15
Document new rules for green CI.

jpakkane:


mensinda (approved):
This would also give everyone more motivation (me included) to actually fix the CI instead of hoping that it will resolve itself over time :)

Issue #9038, opened 2021-07-28
Regression in Subproject Exception Handling from 0.58.2 to 0.59.0


nosracd:
**Describe the bug**
Subprojects failing to configure cause errors later on.

**To Reproduce**
Using the following meson.build:

```
project('foo', 'cpp')

dependency('bar',
    required: false,
    disabler: true,
    fallback: ['bar', ''])

python = import('python').find_installation('python3').dependency()
```
with the following purposely failing subproject at `subprojects/bar/meson.build`
```
project('bar', 'cpp')

python = import('python').find_installation('python3')

dependency('nonexistant-dependency')
```

attempt to configure using
```
meson build
```

**Expected behavior**
Using `0.58.2` results in the subproject failing but the rest of the project configuring successful and the command completing with a zero exit code 

**Expected behavior**
Using `0.59.0` results in the subproject failing as expected, but the rest of the project completing with nonzero exit code and the following error
```
meson.build:8:0: ERROR: Unknown method "dependency" in object <[ExternalProgramHolder] holds [PythonExternalProgram]: <PythonExternalProgram 'python3' -> ['/usr/bin/python3']>> of type ExternalProgramHolder.
```

**system parameters**
* Is this a [cross build](https://mesonbuild.com/Cross-compilation.html) or just a plain native build (for the same computer)?
    * This is a plain bare-metal build
* what operating system (e.g. MacOS Catalina, Windows 10, CentOS 8.0, Ubuntu 18.04, etc.)
    * Ubuntu 20.04
* what Python version are you using e.g. 3.8.0
    * Python 3.8.10
* what `meson --version`
    * `0.59.0`
* what `ninja --version` if it's a Ninja build
    * N/A

Potentially related to #9036

dcbaker:
This sure looks like `find_installation` isn't getting holdered correctly, it should be using a special holder type, but is instaed getting the generic ExternalProgramHolder. @mensinda, ideas?

mensinda:
Seems like an entry in [this list](https://github.com/mesonbuild/meson/blob/master/mesonbuild/interpreter/interpreter.py#L468) is missing for `PythonExternalProgram`. However, it is added to the interpreter [here](https://github.com/mesonbuild/meson/blob/master/mesonbuild/modules/python.py#L651).

So, in short, I currently have no idea until I have time to debug this.

xclaesse:
[...]

PR #9122, opened 2021-08-16
ci: Use 32-bit Rust and Python on vs2017 x86

nirbheek:
See also: https://github.com/mesonbuild/meson/issues/9038#issuecomment-898960933


nirbheek:
Hm, this seems to have magically fixed the `vs2017 clangclx64ninja` job too.

dcbaker (approved):
I could have sworn i had written this exact patch already...

xclaesse:
I suspect this should fix CI for https://github.com/mesonbuild/meson/pull/9123 too.

PR #9125, opened 2021-08-16
compilers/c++: Add MSVC option to make the __cplusplus define accurate

dcbaker:
Otherwise the define will always have the C++98 value, even when using a
newer version. This is only relavent for versions starting at 15.7,
before that the #define would be correct.

### DIFF
--- a/docs/markdown/snippets/msvc_cplusplus_define.md
+++ b/docs/markdown/snippets/msvc_cplusplus_define.md
@@ -1,0 +1,7 @@
+## MSVC now sets the __cplusplus #define accurately
+
+For reasons, MSVC will always return `199711L` for `__cplusplus`, even when a
+newer c++ standard is explicitly requested, unless you pass a specific option to
+the compiler for MSVC 2017 15.7 and newer. Older versions are unaffected by
+this. Meson now always sets the option if it is available, as it is unlikley
+that users want the default behavior.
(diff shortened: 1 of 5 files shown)


jpakkane:
Test failure seems relevant.

dcbaker:
Yeah, i just have no idea why. @jenatali do your have any idea why this would cause rebuilds?

jenatali:
No, no idea. I'd need to debug it, I'm not familiar enough with meson's internals to guess why new `always_args` only apply to first-generation.

docs/markdown/snippets/msvc_cplusplus_define.md
```diff
@@ -0,0 +1,7 @@
+## MSVC now sets the __cplusplus #define accurately
+
+For reasons, MSVC will always return `199711L` for `__cplusplus`, even when a
+newer c++ standard is explicitly requested, unless you pass a specific option to
+the compiler for MSVC 2017 15.7 and newer. Older versions are unaffected by
+this. Meson now always sets the option if it is available, as it is unlikley
```
eli-schwartz, line 6:
```suggestion
this. Meson now always sets the option if it is available, as it is unlikely
```

eli-schwartz:
This was a reapply of #9075 which was reverted due to breaking the CI. It got reapplied with working CI in #10323

PR #9128, opened 2021-08-17
Add PEP 517 backend

### LINKED ISSUE #7863: Expose a PEP 517 backend

FFY00:
Fixes  #7863


FFY00:
@jpakkane I have started implementing the PEP 517 backend and would like initial feedback for the approach I am going for. I have only implemented the `build_sdist` hook, but it should give a fairly decent overview of how the code would look like. What do you think?

To test you can just use the following `pyproject.toml` in a meson project:
```
[build-system]
requires = ['meson']
build-backend = 'mesonbuild.python'
```

Example run: https://asciinema.org/a/HCVWtTRW8i65nNZTTwUWyWGmx

eli-schwartz:
> and would like initial feedback for the approach I am going for. I have only implemented the `build_sdist` hook, but it should give a fairly decent overview of how the code would look like.

Coding style aside, it seems like the main discussion around how to do this would inherently take place over "how to build a wheel"... also "what's included in a wheel"...

As brought up in the linked ticket, we probably need a toml module, and "The toml parser would be the easy part, we can pull toml as an optional dependency" is a strict violation of meson's zero-dependencies policy.

What's the status of getting a toml module into the stdlib? IIRC the PyPA ecosystem has now gone through *three* toml modules, they keep on getting abandoned and it's starting to look like a trend.

FFY00:
> Coding style aside, it seems like the main discussion around how to do this would inherently take place over "how to build a wheel"... also "what's included in a wheel"...

It's not as it does not add a dependency to meson itself, but rather to build system of the package we are building.

The `pyproject.toml` a project that uses the meson python backend would look something like:
```toml
[build-system]
requires = ['meson', 'tomli']
build-backend = 'mesonbuild.python'
```

Anyway, I am not even sure it makes sense to put the python specific config in `pyproject.toml`. We, as the build system, don't really need to read `pyproject.toml`, we only would if we had some config there. We could provide a meson directive to set the python metadata of project.

eli-schwartz:
I don't believe it makes any sense to implement a build backend for meson, which cannot be used in the frontend when people run meson themselves for enhanced control. If people just want to use meson as a hidden detail of a pep517 build backend, then mesonpep517 exists to factor that dependency out.

Since meson is a multi-language, general-purpose build system, it should be able to install a python distribution alongside other languages, e.g. to add bindings to an existing project. When used via a pep517 builder, it could build a statically linked bindings-only wheel and hide the fact that meson is used.

So the code to support building python distributions would be a first-class citizen of meson and not just something to say "oh, it can depend on anything it wants because it's only ever used by pip or pyproject-build".

Again, if people want the latter then mesonpep517 is a perfectly reasonable solution and there's no reason to include it directly in meson.

eli-schwartz:
> That's why I created the issue first, and the feedback was that it was reasonable.

It is confusing to have this discussion both here and in the linked issue. And my comment was primarily directed at the linked issue, which stated that we would probably need a toml parser, a statement no one disagreed with.

> Eli, please re-read my replies, and my code. Which part depends on anything external? We don't need to depend on a toml parser, or anything else.

Given this is still a draft that has only implemented adding one file to `meson dist`, I have no way of foreseeing what the not-yet-implemented part would do.

But:

> We, as the build system, don't really need to read `pyproject.toml`, we only would if we had some config there. We could provide a meson directive to set the python metadata of project.

This seems against the spirit of PEP 621, at least.

>  Personally, I don't think meson needs to support exporting Python distributions via the CLI. I think having first-class integration with the Python ecosystem is reasoning enough to merge such functionality, even if it is not exposed by the frontend.

That's precisely what mesonpep517 provides: first class integration with the python ecosystem to delegate the build process to meson. What use case is enabled by a single-directional integration directly in the meson source code?

PR #9131, opened 2021-08-17
backends/ninja: write depscan input files to json

### LINKED ISSUE #9129: Response file needed for Fortran project on Linux
Inspired by #7212.

A project with lots of Fortran source:

```sh
#!/bin/sh
set -eu
work_dir=meson_toolong
build_dir=b
n=12000
mkdir -p ${work_dir}
cd ${work_dir}
rm -rf ${build_dir}
{
  echo "project('toolong', 'fortran')"
  echo "srcs = ["
} > meson.build
for i in $(seq -w ${n})
do
  src_file=f${i}.f90
  {
    echo "Subroutine s${i}"
    echo "End"
  } > "${src_file}"
  echo " '${src_file}'," >> meson.build
done
{
  echo "]"
  echo "library('toolong', srcs)"
} >> meson.build
meson . ${build_dir}
cd ${build_dir}
ninja -v
```

gives

```
The Meson build system
Version: 0.59.0
Source dir: /home/mat_cross/documents/Meson/resp/meson_toolong
Build dir: /home/mat_cross/documents/Meson/resp/meson_toolong/b
Build type: native build
Project name: toolong
Project version: undefined
Fortran compiler for the host machine: gfortran (gcc 11.2.1 "GNU Fortran (GCC) 11.2.1 20210728 (Red Hat 11.2.1-1)")
Fortran linker for the host machine: gfortran ld.bfd 2.35.2-4
Host machine cpu: x86_64
Build targets in project: 1

ninja: fatal: posix_spawn: Argument list too long
```

(Python is 3.9.6).

dcbaker:
Currently, we write each file to the command line, but this can result in
situations where the number of files passed exceeds OS imposed command
line limits. For compilers, we solve this with response files. For
depscan I've chosen to use a JSON list instead. JSON has several
advantages in that it's standardized, there's a built-in python module
for it, and it's familiar. I've also chosen to always use the JSON file
instead of having a heuristic to decide between JSON and not JSON,
while there may be a small performance trade off here, keeping the
implementation simple with only one path is wort it.

Fixes #9129

### DIFF
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -11,18 +11,20 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import typing as T
-import os
-import re
-import pickle
-import shlex
-import subprocess
+
 from collections import OrderedDict
 from enum import Enum, unique
-import itertools
-from textwrap import dedent
-from pathlib import PurePath, Path
 from functools import lru_cache
+from pathlib import PurePath, Path
+from textwrap import dedent
+import itertools
+import json
+import os
+import pickle
+import re
+import shlex
+import subprocess
+import typing as T
 
 from . import backends
 from .. import modules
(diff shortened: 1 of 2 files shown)


eli-schwartz (approved):

rgommers:
This PR caused a regression for us, see gh-9258.

PR #9134, opened 2021-08-18
make python extension_module install to the documented location

### LINKED ISSUE #6331: Python extension_module does not install in site_packages
In the *python* module documentation is claimed that `extension_module` install per default in `site_packages`:
> `subdir`: By default, meson will install the extension module in the relevant top-level location for the python installation, eg `/usr/lib/site-packages`. When subdir is passed to this method, it will be appended to that location. This keyword argument is mutually exclusive with `install_dir`

However, this does not work. When set `install: true` the modules are installed to `$prefix/lib64` on my system.

One workaround is to specify the target like:
```
py3_mod = import('python')
py3_inst = py3_mod.find_installation('python3')

foo = py3_inst.extension_module('foo',
    <source_target>,
    install: true,
    subdir: get_option('prefix') + py3_inst.get_path('purelib'),
    dependencies: [py3_inst.dependency()]
)
```
It would be nice, if the subdir is set autotically to the `site-packages` folder.

This bug is btw similar to #2859. However, this talks about the now deprecated *python3* module.

eli-schwartz:
Currently it installs to libdir instead.

While I am at it, I had to extend the testsuite to support configuring via test.json, if a file is a python file and needs to be configured with python-specific path expansions.

Fixes #6331

### DIFF
--- a/docs/markdown/Contributing.md
+++ b/docs/markdown/Contributing.md
@@ -264,15 +264,18 @@ current platform. The following values are currently supported:
 | type          | Description                                                                                             |
 | ------------- | ------------------------------------------------------------------------------------------------------- |
 | `file`        | No postprocessing, just use the provided path                                                           |
+| `python_file` | Use the provided path while replacing the python directory.                                             |
 | `dir`         | To include all files inside the directory (for generated docs, etc). The path must be a valid directory |
 | `exe`         | For executables. On Windows the `.exe` suffix is added to the path in `file`                            |
 | `shared_lib`  | For shared libraries, always written as `name`. The appropriate suffix and prefix are added by platform |
+| `python_lib`  | For python libraries, while replacing the python directory. The appropriate suffix is added by platform |
 | `pdb`         | For Windows PDB files. PDB entries are ignored on non Windows platforms                                 |
 | `implib`      | For Windows import libraries. These entries are ignored on non Windows platforms                        |
+| `py_implib`   | For Windows import libraries. These entries are ignored on non Windows platforms                        |
 | `implibempty` | Like `implib`, but no symbols are exported in the library                                               |
 | `expr`        | `file` is an expression. This type should be avoided and removed if possible                            |
 
-Except for the `file` and `expr` types, all paths should be provided *without* a suffix.
+Except for the `file`, `python_file` and `expr` types, all paths should be provided *without* a suffix.
 
 | Argument   | Applies to                 | Description                                                                   |
 | -----------|----------------------------|-------------------------------------------------------------------------------|
@@ -283,6 +286,11 @@ The `shared_lib` and `pdb` types takes an optional additional
 parameter, `version`, this is us a string in `X.Y.Z` format that will
 be applied to the library. Each version to be tested must have a
 single version. The harness will apply this correctly per platform:
+
+The `python_file`, `python_lib`, and `py_implib` types have basic support for configuring the string with the `@<VAR>@` syntax:
+
+- `@PYTHON_PLATLIB@`: python `get_install_dir` directory relative to prefix
+- `@PYTHON_PURELIB@`: python `get_install_dir(pure: true)` directory relative to prefix
 
 `pdb` takes an optional `language` argument. This determines which
 compiler/linker should generate the pdb file. Because it's possible to
--- a/mesonbuild/modules/python.py
+++ b/mesonbuild/modules/python.py
@@ -402,9 +402,9 @@ class PythonInstallation(ExternalProgramHolder):
         suffix = self.variables.get('EXT_SUFFIX') or self.variables.get('SO') or self.variables.get('.so')
 
         # msys2's python3 has "-cpython-36m.dll", we have to be clever
-        split = suffix.rsplit('.', 1)
-        suffix = split.pop(-1)
-        args[0] += ''.join(s for s in split)
+        # FIXME: explain what the specific cleverness is here
+        split, suffix = suffix.rsplit('.', 1)
+        args[0] += split
 
         kwargs['name_prefix'] = ''
         kwargs['name_suffix'] = suffix
--- a/mesonbuild/modules/python.py
+++ b/mesonbuild/modules/python.py
@@ -409,10 +409,10 @@ class PythonInstallation(ExternalProgramHolder):
 
     @permittedKwargs(mod_kwargs)
     def extension_module_method(self, args: T.List['TYPE_var'], kwargs: 'TYPE_kwargs') -> 'SharedModule':
-        if 'subdir' in kwargs and 'install_dir' in kwargs:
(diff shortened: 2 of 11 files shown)


eli-schwartz:
(Why does Windows install import libraries for a python extension_module.)

run_project_tests.py
```diff
@@ -282,6 +300,8 @@ def __lt__(self, other: object) -> bool:
 backend:      'Backend'
 backend_flags: T.List[str]
 
+python_info: 'PythonIntrospectionDict'
```
eli-schwartz, line 303: https://bugs.python.org/issue34939

cross-only-armhf fails, but not in my fork due to never running :/ and this is a lovely python bug in 3.7

docs/markdown/Contributing.md
```diff
@@ -264,15 +264,18 @@ current platform. The following values are currently supported:
 | type          | Description                                                                                             |
 | ------------- | ------------------------------------------------------------------------------------------------------- |
 | `file`        | No postprocessing, just use the provided path                                                           |
+| `python_file` | Use the provided path while replacing the python directory.                                             |
 | `dir`         | To include all files inside the directory (for generated docs, etc). The path must be a valid directory |
 | `exe`         | For executables. On Windows the `.exe` suffix is added to the path in `file`                            |
 | `shared_lib`  | For shared libraries, always written as `name`. The appropriate suffix and prefix are added by platform |
+| `python_lib`  | For python libraries, always written as `python_file`. The appropriate suffix is added by platform      |
```
dcbaker, line 271: shoudl'nt this say `python_lib` in the Description?

mesonbuild/modules/python.py
```diff
@@ -334,6 +335,28 @@ def __init__(self, name: str, command: T.Optional[T.List[str]] = None,
             'version': '0.0',
         }
 
+    def sanity(self) -> 'PythonExternalProgram':
```
dcbaker, line 338: this seems wrong, it can return a `NonExistingExternalProgram`, which is a sibling class, right? So it either needs to be `ExternalProgram`, or it needs to be `T.Union[PythonExternalProgram, NonExistingExternalProgram]`

mesonbuild/modules/python.py
```diff
@@ -334,6 +335,28 @@ def __init__(self, name: str, command: T.Optional[T.List[str]] = None,
...
+            info = json.loads(stdout)
+        except json.JSONDecodeError:
+            info = None
+            mlog.debug('Could not introspect Python (%s): exit code %d' % (str(p.args), p.returncode))
+            mlog.debug('Program stdout:\n')
+            mlog.debug(stdout)
+            mlog.debug('Program stderr:\n')
+            mlog.debug(stderr)
+
+        if isinstance(info, dict) and 'version' in info:
```
dcbaker, line 352: I really feel like we should be checking separately whether the info is None or whether it's a dictionary. `None` is completely expected, but if we got a list her that would be a programming error that we'd be silently hiding.

mesonbuild/modules/python.py
```diff
@@ -334,6 +335,28 @@ def __init__(self, name: str, command: T.Optional[T.List[str]] = None,
...
+            variables = info['variables']
+            info['suffix'] = variables.get('EXT_SUFFIX') or variables.get('SO') or variables.get('.so')
+            self.info = T.cast('PythonIntrospectionDict', info)
```
dcbaker, line 355: this all feels kinda sketchy, a `PythonIntrospectionDict` needs to have all the keys in it, maybe we should be checking for all of them?

run_project_tests.py
```diff
@@ -122,6 +123,8 @@ def add_step(self, step: BuildStep, stdo: str, stde: str, mlog: str = '', time:
     def fail(self, msg: str) -> None:
         self.msg = msg
 
+python_info: 'PythonIntrospectionDict'
```
dcbaker, line 126: why can't this just be assigned here at module time, instead of with the global hack in the InstallFile initializer?

eli-schwartz: Hmm... I wasn't sure whether it should run at script load or only once used. In retrospect, it is still used every single time test.json is parsed, not just for python, so... I guess we could do this globally.

mesonbuild/modules/python.py
eli-schwartz: This is a straight refactor from the previous location, which included:

```python
if isinstance(info, dict) and 'version' in info and self._check_version(name_or_path, info['version'])
```

But I think the intention was to just not catch KeyError, not sure why it didn't use info.get().

mesonbuild/modules/python.py
eli-schwartz: The original check (which I simply refactored), didn't try to "check for programming errors", only expected errors, i.e. it has to either be None or a json.loads which should always be a dict.

mesonbuild/modules/python.py
eli-schwartz: True. I slapped this on at the end trying desperately to quell mypy, it's not my area of expertise.

docs/markdown/Contributing.md
eli-schwartz: Well, I meant it follows the same format as python_file. idk, describing things here almost takes more time than writing the actual code...

mesonbuild/modules/python.py
eli-schwartz: The introspect command cannot, IMO, fail here. It has to either error out, or return a very precise json.dumps...

mesonbuild/modules/python.py
eli-schwartz: `if info is not None` -> access it while assuming it is a dict. Programming errors will result in tracebacks now.

mesonbuild/modules/python.py
eli-schwartz: Now we check for none of them, but assume 'version' exists.

eli-schwartz:
Didn't realize pure: true is the default... having some mechanism to demonstrate that in practice feels important for clarity...

jpakkane:
Failure seems relevant.

```
test cases/unit/47 native file binary/meson.build:18:2: ERROR: <PythonExternalProgram 'python3' -> ['/__w/meson/meson/tmpkmods5fy/binary_wrapper0.py']> is not a valid python or it is missing setuptools
```

eli-schwartz:
I have been trying to figure that one out for a while now, turns out it is because I wanted to get something useful out of PythonExternalProgram.name

dcbaker (approved):

nirbheek:
> @eli-schwartz the patches in this PR don't apply cleanly to 0.59, and the changes involved are large enough that I don't feel comfortable porting it over. Lots of changes happened in master to this file after 0.59 was released. For instance, type checking was added.

This has been ported at https://github.com/mesonbuild/meson/pull/9295 and merged, re-adding milestone.

PR #9135, opened 2021-08-18
docs: Remove the deleted function find_library

mensinda:


xclaesse:
about time we remove that, yes. LGTM.

PR #9137, opened 2021-08-19
Lfortran

HaoZeke:
This needs a custom linker before it is ready.


dcbaker:
Just an fyi, we don't do back merges, you'll need to rebase on master instead

HaoZeke:
Sorry about that @dcbaker, I was swayed by the shiny new buttons on the Github web interface; I'll rebase before changing the draft status.

dcbaker:
No worries. Just thought I'd warn you while it was still easy to rebase :)

mesonbuild/compilers/fortran.py
```diff
@@ -464,6 +464,27 @@ def __init__(self, exelist: T.List[str], version: str, for_machine: MachineChoic
     def language_stdlib_only_link_flags(self) -> T.List[str]:
         return ['-lflang', '-lpgmath']
 
+class LFortranFortranCompiler(FortranCompiler):
```
eli-schwartz, line 467: `gfortran` -> `GnuFortranCompiler`
`lfortran` -> `LFortranFortranCompiler` ?

I believe a better class name for this (so that people don't wonder what a FortranFortran compiler is) would be `LLVMFortranCompiler`

HaoZeke: I'd like that but it isn't unambiguous. Flang is also an LLVM compiler (and part of LLVM core nowadays)

dcbaker: Which we already support, and I'm pretty sure it's called either FlangFortran or ClangFortran

eli-schwartz: Well, that seems annoying. :/ Dunno what to suggest for a better name.

dcbaker: How about just `LFortranCompiler`?

HaoZeke: I think `LFortranCompiler` sounds good; thanks ^_^

mesonbuild/compilers/fortran.py
```diff
@@ -72,7 +72,7 @@ def sanity_check(self, work_dir_: str, environment: 'Environment') -> None:
         if binary_name.is_file():
             binary_name.unlink()
 
-        source_name.write_text('print *, "Fortran compilation is working."; end', encoding='utf-8')
+        source_name.write_text('print *, "Fortran compilation is working."', encoding='utf-8')
```
eli-schwartz, line 75: I know nothing about fortran. What's the meaning of this change?

eli-schwartz (commented):
Hmm, the tests are failing with `ERROR: Compiler gfortran can not compile programs.` so something here has caused general regressions.

There seems to only be one change outside of the Lfortran scaffolding itself.

mesonbuild/compilers/fortran.py
HaoZeke: Ah, sorry; actually this one is an incomplete fix for fortran compilers which was more fully fleshed out in #9248; I shouldn't have put this here.

HaoZeke: Essentially a complete fortran program to test with is:

```fortran
program main
print *, "Hello"
end program
```

Which can be written as `program main; print*, "Fortran compilation is working."; end program`

Depending on the standards; some compilers accept not having `program` in the beginning; and others accept not explicitly requiring `end program`.

eli-schwartz: The other PR is merged now, so you can "put this here" by rebasing. :D

mesonbuild/compilers/fortran.py
```diff
@@ -464,6 +464,27 @@ def __init__(self, exelist: T.List[str], version: str, for_machine: MachineChoic
...
+                                 full_version=full_version)
+        self.id = 'lfortran'
+        default_warn_args = []
+        self.warn_args = {'0': [],
+                          '1': default_warn_args,
+                          '2': default_warn_args,
+                          '3': default_warn_args}
+
+    def language_stdlib_only_link_flags(self) -> T.List[str]:
+        return ['-llfortran_runtime_static', '-llfortran_runtime']
```
dcbaker, line 486: I just got code merged that also add the default search path here, which is necessary when combining different compiler family's (GCC and lfortran,  for example).

eli-schwartz:
The lfortran topic came up again on conda-forge w.r.t. building Windows versions of SciPy (see linked issue).

@HaoZeke any further progress in implementing this?

PR #9141, opened 2021-08-20
Again some Xcode fixes

jpakkane:
A few regressions and minor tweaks. Can now build [Godot](https://github.com/godotengine/godot/pull/51153).

We should probably consider adding some sort of a CI for the Xcode backend. Sadly it is incredibly slow and not all of the test suite works with it yet (and there does not seem to be many volunteers around to fix it).

### DIFF
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -1269,13 +1269,17 @@ class Backend:
             elif isinstance(i, build.GeneratedList):
                 fname = [os.path.join(self.get_target_private_dir(target), p) for p in i.get_outputs()]
             elif isinstance(i, build.ExtractedObjects):
-                fname = [os.path.join(self.get_target_private_dir(i.target), p) for p in i.get_outputs(self)]
+                outputs = i.get_outputs(self)
+                fname = self.get_extracted_obj_paths(i.target, outputs)
             else:
                 fname = [i.rel_to_builddir(self.build_to_src)]
             if target.absolute_paths:
                 fname = [os.path.join(self.environment.get_build_dir(), f) for f in fname]
             srcs += fname
         return srcs
+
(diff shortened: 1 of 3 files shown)


dcbaker:
I'm planning to start adding type annotations to the backends now that the core backend annotations are in. If we were serious about getting a CI build going i can help fix bugs.

mesonbuild/backend/xcodebackend.py
```diff
@@ -136,6 +136,9 @@ def add_item(self, key, value, comment=''):
         self.keys.add(key)
         self.items.append(item)
 
+    def has_item(self, key):
+        return key in self.keys
```
dcbaker, line 140: This isn't very idiomatic python. Why not implement a `__contains__` dunder method?

mesonbuild/backend/backends.py
```diff
@@ -1171,7 +1171,12 @@ def get_custom_target_sources(self, target):
             elif isinstance(i, build.GeneratedList):
                 fname = [os.path.join(self.get_target_private_dir(target), p) for p in i.get_outputs()]
             elif isinstance(i, build.ExtractedObjects):
-                fname = [os.path.join(self.get_target_private_dir(i.target), p) for p in i.get_outputs(self)]
+                outputs = i.get_outputs(self)
+                if self.name == 'xcode':
+                    # Xcode is stupid and requires a completely different naming scheme.
+                    fname = outputs
+                else:
+                    fname = [os.path.join(self.get_target_private_dir(i.target), p) for p in outputs]
```
dcbaker, line 1179: This really feels like a layering violation. Maybe we need a separate method that the xcode backend can override?

dcbaker (changes requested):
Overall looks reasonable, i have a few implementation suggestions, but otherwise I'm good with this.

dcbaker (approved):

jpakkane:
Test failure is unrelated and did pass on all the other test runs.

PR #9143, opened 2021-08-20
Update instructions on installing with Pip. Closes #9034.

### LINKED ISSUE #9034: DOC Avoiding to install as root on Linux
The first line of the [installation docs](https://mesonbuild.com/Getting-meson.html#installing-meson-with-pip) says,

>  Meson is available in the Python Package Index and can be installed with `sudo pip3 install meson` which requires root

Installing packages as root from PyPi is practice that's very much frowned upon, and the solution is often to use virtualenvs rather that using sudo. Sources:
 - The [Python Packaging Guide](https://packaging.python.org/tutorials/installing-packages/#id12) 
   > Due to the way most Linux distributions are handling the Python 3 migration, Linux users using the system Python without creating a virtual environment first should replace the python command in this tutorial with python3 and the python -m pip command with python3 -m pip --user. **Do not run any of the commands in this tutorial with sudo**: if you get a permissions error, come back to the section on creating virtual environments, set one up, and then continue with the tutorial as written.
 - Various most voted Stack Overflow posts: https://stackoverflow.com/a/21056000/1791279, https://stackoverflow.com/questions/15028648/is-it-acceptable-and-safe-to-run-pip-install-under-sudo

It's more or less arbitrary code execution as root. So IMO it might be better not to have this as the recommended installation option for unaware users.

In addition from that sentence it's not clear if it's only `sudo` that requires root (and user is free to install it as non root in a virtualenv) or if meson install additionally does something that requires root, and it needs to be installed as root in any case. This question is answered a few paragraphs below  but still.

I can make a PR to improve the documentation if useful.

jpakkane:


docs/markdown/Getting-meson.md
```diff
@@ -24,19 +24,21 @@ several platforms.
 ## Installing Meson with pip
 
 Meson is available in the [Python Package Index] and can be installed
-with `sudo pip3 install meson` which requires root and will install it
-system-wide.
-
-If you have downloaded a copy of the Meson sources already, you can
-install it with `sudo pip3 install path/to/source/root/`.
```
eli-schwartz, line 31: The description on installing from a source checkout got dropped entirely.

eli-schwartz (changes requested):

docs/markdown/Getting-meson.md
jpakkane: That is intentional. This is a beginner document and that is an advanced concept. Simpler is better.

rth:
Thanks for improving the wording!

PR #9145, opened 2021-08-20
There's a problem with compiler arguments subcommand parser

evilint:
There's was a problem with the argument list passed for target object as LANGUAGE_args: ['ARGS'],
that all user arguments passed in the form of a single text (without backspaces)
like: "-Cprefer-dynamic=yes" works without problems, but "-C prefer-dynamic=yes"
does not works and lead in various errors!

The meson.build file:
Screenshot from 2021-08-20 19-53-38

The problem:
Screenshot from 2021-08-20 19-11-23
Screenshot from 2021-08-20 20-37-18

My fix:
Screenshot from 2021-08-20 19-53-57

### DIFF
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -1350,7 +1350,18 @@ You probably should put it in link_with instead.''')
         if language in self.extra_args:
             self.extra_args[language] += args
         else:
-            self.extra_args[language] = args
+            arglist: T.List[FileOrString] = None
+            string = ''.join(args)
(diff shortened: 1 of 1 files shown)


eli-schwartz:
meson by design has an actual type system, you should use an array here.

```
rust_args: ['-C', 'prefer-dynamic=yes']
```

We will not guess whether or not your whitespace is significant, this:
- is bad (unpredictable) style
- will break existing code
- has a better, official way to do it

PR #9147, opened 2021-08-21
python: Workaround broken install path

xclaesse:


xclaesse:
@eli-schwartz @woju what do you think?

eli-schwartz:
Looks reasonable, slightly odd the warning mentions platlib even for purelib though.

Test failure is significant, my recent PR added tests to this module's installation capabilities which are now failing to find installed files in `info['install_paths']['platlib']`and suchlike.

xclaesse:
> slightly odd the warning mentions platlib even for purelib though

Reformulated the warning a bit and print it only once.

> Test failure is significant, my recent PR added tests to this module's installation capabilities which are now failing to find installed files in `info['install_paths']['platlib']`and suchlike.

Fixed CI by moving the logic down to `PythonExternalProgram()` instead of in `PythonInstallation()`.

I also changed to not warn on Debian derivatives because there is no need to warn since we workaround the problem.

eli-schwartz:
We can still warm people on debian too :p maybe a debian maintainer will notice and fix sysconfig. Also we do not actually know that the path were using actually works, we're relying on a dirty heuristic which debian might change out from under us. So it makes sense to warm the user to double check that the results actually work as expected.

xclaesse:
Hm, ok, I guess you're right, projects can get ride of the warning by adding a project option as we suggest in the other doc PR.

eli-schwartz (approved):

xclaesse:
Hold on, tested this on Windows and it prints the warning too.
- `rel_path = self.info['install_paths'][key][1:]` that gives: `lib\python3.8\site-packages`
- `sys.path` gives `['', 'C:\\Python38\\python38.zip', 'C:\\Python38\\DLLs', 'C:\\Python38\\lib', 'C:\\Python38', 'C:\\Python38\\lib\\site-packages']`

This is with default installation of python 3.8 on Windows with official installer.

eli-schwartz:
The python module's python_installation.install_dir() method is being seeded by the sysconfig module get_paths() configured to output the "posix_prefix" scheme rather than the "nt" scheme. I suspect this is the problem.

It exists prior to this PR, of course. Good job exposing it. :D

nirbheek:
@xclaesse the patch in this PR doesn't apply cleanly to 0.59, and the changes involved are large enough that I don't feel comfortable porting it over. Lots of changes happened in master to this file after 0.59 was released. For instance, type checking was added. Same problem as https://github.com/mesonbuild/meson/pull/9134.

nirbheek:
Thanks to the backporting in https://github.com/mesonbuild/meson/pull/9295, porting this to 0.59 was trivial again, so re-added milestone.

Issue #3426, opened 2018-04-18
Detect when people try to run meson from a subdir in the source tree and print a friendly error


xyproto:
I think it's hard to understand what "First statement must be a call to project" means. I question if the sentence is proper English, but I'm not a native speaker, so it would be nice if someone could confirm this.

Whom are you calling upon to project something, and what will you project? Or, if it's not in the context of calling upon someone or projecting something, which function are you calling? `project` is not a function and it's not obvious how `project` should be called.

Could the error message be reworded?

nirbheek:
> project is not a function and it's not obvious how project should be called.

I don't understand what you mean. `project` is a function: http://mesonbuild.com/Reference-manual.html#project

We should definitely make the error message clearer though. Perhaps we should point to that link.

xyproto:
I was trying to compile just `swaybar` which is part of the [Sway](https://github.com/swaywm/sway) window manager.

Being new to Meson, I naively checked out the `sway` sources, entered the `swaybar` directory, typed in `meson build` and got:

    ERROR: First statement must be a call to project

In the context of the commandline, there is no `project` function immediately available, so this was incomprehensible, to me.

I believe the real problem was that I needed to build the entire project (`cd ..; meson build; ninja -C build`) and not anything that had to do with the `project` function.

This issue may mainly be with me, since I'm new to meson, but perhaps another error message for this situation could also be helpful for others. Consider it a feature request in the pile of feature requests.

Thanks for developing meson, appart from the above issue I think it looks really promising and I believe it's an "up and coming" build system. :)

NickeZ:
Can't we do like git an walk the directories upwards until we find a `meson.build` file with `project` in it?

nirbheek:
Yes, that's what I was thinking too. We probably want to look upwards till we find a directory without a `meson.build`, and the previous dir was probably the source root. Then we tell the user to run meson from that directory.

In some cases there will indeed not be a meson.build with `project()` anywhere in the parent directories (f.ex., for git submodules that are dropped in as subdirs and not subprojects), and in that case a more useful error message than `use project()` would be appropriate.

rubyFeedback:
I found this issue via Google just now. I am using meson on windows.

On Linux it is all fine; on meson I get this error "First statement must be a call to project".

I was doing this line:

    meson --prefix=/usr --libdir=/usr/lib BUILD_DIRECTORY

This works fine on Linux. On windows, when I try to compile atk (from the gtk stack),
[...]

PR #9148, opened 2021-08-21
interpreter: Improve incorrect source root error

nirbheek:
Keep it simple instead of using a complicated git walk or recursive
parent listdir, which would likely be wrong. For instance:

1. We would have to parse the topmost meson.build file to
   check if it does indeed contain a project()
2. It is not required that each subdir with a meson.build file has
   a meson.build file in its parent, since you can call
   subdir('foo/bar')
3. Anytime we do recursive searches, we can fall prey to loops due to
   symlinks or mounts.

Fixes https://github.com/mesonbuild/meson/issues/3426

### DIFF
--- a/mesonbuild/interpreterbase/interpreterbase.py
+++ b/mesonbuild/interpreterbase/interpreterbase.py
@@ -45,7 +45,7 @@ from .disabler import Disabler, is_disabled
 from .helpers import check_stringlist, default_resolve_key, flatten, resolve_second_level_holders
 from ._unholder import _unholder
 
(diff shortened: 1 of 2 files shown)


eli-schwartz:
Alright, based on discussion in my PR #9152 and on IRC we're sticking with my alternative approach but I will follow up on it by adding a simple check for project( in the first line of a heuristically found meson.build

3 is not a problem because the implementation doesn't perform symlink traversal at all, so we can't get caught in an endless loop.

2 will just fall back to not detecting a project.

PR #9152, opened 2021-08-22
use a more informative error message for invoking meson in a subdir

### LINKED ISSUE #3426: Detect when people try to run meson from a subdir in the source tree and print a friendly error

eli-schwartz:
Explicitly mention that the project definition is invalid, and clarify that project is `project()` -- a function.

Also try to walk the directory tree upward, and if there are parent meson.build files, just say this isn't the project root, and "maybe you meant to run meson there instead?"

This won't catch calls to subdir('foo/bar') but we can't be perfect, only better than before and catch the *majority* of such cases, and hopefully it's a lot more clear if meson protests that the project is "invalid, there is no project() function", where the user should look for a potential solution.

Fixes #3426

### DIFF
--- a/mesonbuild/interpreterbase/interpreterbase.py
+++ b/mesonbuild/interpreterbase/interpreterbase.py
@@ -45,7 +45,7 @@ from .disabler import Disabler, is_disabled
 from .helpers import check_stringlist, default_resolve_key, flatten, resolve_second_level_holders
 from ._unholder import _unholder
 
-import os, copy, re
+import os, copy, re, pathlib
 import typing as T
 
 if T.TYPE_CHECKING:
@@ -123,7 +123,19 @@ class InterpreterBase:
             raise InvalidCode('No statements in code.')
         first = self.ast.lines[0]
         if not isinstance(first, mparser.FunctionNode) or first.func_name != 'project':
-            raise InvalidCode('First statement must be a call to project')
+            p = pathlib.Path(self.source_root).resolve()
+            found = p
+            for parent in p.parents:
+                if (parent / 'meson.build').is_file():
+                    found = parent
+                else:
+                    break
+
+            error = 'first statement must be a call to project()'
+            if found != p:
+                raise InvalidCode(f'Not the project root: {error}\n\nDid you mean to run meson from the directory: "{found}"?')
+            else:
+                raise InvalidCode(f'Invalid source tree: {error}')
 
     def run(self) -> None:
         # Evaluate everything after the first line, which is project() because
--- a/test cases/failing/1 project not first/test.json
+++ b/test cases/failing/1 project not first/test.json
@@ -1,7 +1,7 @@
 {
   "stdout": [
     {
-      "line": "ERROR: First statement must be a call to project"
+      "line": "ERROR: Invalid source tree: first statement must be a call to project()"
     }
   ]
 }


mesonbuild/interpreterbase/interpreterbase.py
```diff
@@ -123,7 +123,19 @@ def sanity_check_ast(self) -> None:
             raise InvalidCode('No statements in code.')
         first = self.ast.lines[0]
         if not isinstance(first, mparser.FunctionNode) or first.func_name != 'project':
-            raise InvalidCode('First statement must be a call to project')
+            p = pathlib.Path(self.source_root).resolve()
+            found = p
+            for parent in p.parents:
```
nirbheek, line 128: This can cause a loop.

mesonbuild/interpreterbase/interpreterbase.py
```diff
@@ -123,7 +123,19 @@ def sanity_check_ast(self) -> None:
...
+                if (parent / 'meson.build').is_file():
+                    found = parent
+                else:
+                    break
```
nirbheek, line 132: All subdirs don't necessarily contain `meson.build` because you can do `subdir('foo/bar')`.

mesonbuild/interpreterbase/interpreterbase.py
```diff
@@ -123,7 +123,19 @@ def sanity_check_ast(self) -> None:
...
+
+            error = 'first statement must be a call to project()'
+            if found != p:
+                raise InvalidCode(f'Not the project root: {error}\n\nDid you mean to run meson from the directory: "{found}"?')
```
nirbheek, line 136: We should not recommend an alternate path to run meson from without first checking whether that file also has a `project()` in it, otherwise it will just confuse users.

Even if that file has `project()` in it, it may not be the project that the user wanted to build.

nirbheek (commented):
It's because of exactly these three issues that I opened this simpler PR: https://github.com/mesonbuild/meson/pull/9148#issuecomment-903133675

mesonbuild/interpreterbase/interpreterbase.py
nirbheek: Also, the message implies that you must run meson from the project's root directory, which is not true. Some projects have the meson build files in a subdir of the project root (such as VLC).

nirbheek:
I've updated https://github.com/mesonbuild/meson/pull/9148 to also contain a revert commit for this just to fix its state, but if you don't agree with my concerns, please feel free to close that PR. It's no big deal.

mesonbuild/interpreterbase/interpreterbase.py
eli-schwartz: What is the loop? pathlib returns an iterator over the parent elements as pathlib components, terminating in the filesystem anchor (`/` on Linux, `C:\\` or some other drive on Windows).

It doesn't even resolve symlinks or anything. In fact, it doesn't actually resolve their on disk locations.  It just treats everything as a PurePath string, you get the same (successful) results for a path location that doesn't exist on disk.

mesonbuild/interpreterbase/interpreterbase.py
eli-schwartz: I noted this in the commit message, meson will try to provide a guess but isn't meant to be perfect.

mesonbuild/interpreterbase/interpreterbase.py
eli-schwartz: I'd be open to checking that one for a project() function, not sure of the best way to do it though.

Since meson doesn't tell you to build that one, merely asks you if you "meant" to, I don't believe it's a concern that it might be the wrong but valid project.

eli-schwartz:
I didn't notice your PR -- I actually wrote the code for this on Friday but didn't have a chance to update the tests or submit it.

That being said, I only agree with one of your three concerns, and that one could be fixed as an addition to this PR.

mesonbuild/interpreterbase/interpreterbase.py
nirbheek: Ah, I didn't know that. Fair enough.

mesonbuild/interpreterbase/interpreterbase.py
nirbheek: If we're doing heuristics, just opening the file and checking that the first line starts with `project(` might be enough.

We expect users to be "smart enough" but if you tell them that they can try something, they will try it, and then get even more confused when it doesn't work. Sometimes more information is worse. If they're not smart enough to figure out the topmost meson.build file themselves, giving them the wrong location will make things worse.

eli-schwartz: Makes sense. That would be a nice and simple improvement and continue to catch the vast majority of cases.

eli-schwartz:
Follow-up in #9153

PR #9156, opened 2021-08-23
python module: produce the correct install path on every OS

eli-schwartz:
The sysconfig paths are, by default, correct for every OS -- they are supposed to follow the scheme that python knows about per default.

For some reason, this overrode the scheme to posix_prefix, which is the default for posix OSes like linux and macOS, but wrong on Windows. Simply deleting this entirely makes everything that used to work, still work, and a couple new things start working.

@MathieuDuponchelle I'm seeking insight into why this originally got implemented the way it did. Do you happen to remember?

### DIFF
--- a/mesonbuild/modules/python.py
+++ b/mesonbuild/modules/python.py
@@ -279,7 +279,7 @@ import sysconfig
 import json
 import sys
 
-install_paths = sysconfig.get_paths(scheme='posix_prefix', vars={'base': '', 'platbase': '', 'installed_base': ''})
+install_paths = sysconfig.get_paths(vars={'base': '', 'platbase': '', 'installed_base': ''})
 
 def links_against_libpython():
     from distutils.core import Distribution, Extension


xclaesse:
I'm wondering if we should go one step further on Windows. By default `sys.path` has `C:\Python38\lib\site-packages` but with this PR meson is going to install into `<prefix>\lib\site-packages` which by default is `C:\lib\site-packages`. Note sure what should be the best heuristic here, if meson prefix is the root of the same drive where python is installed, then we use the platlib without overriding `'base': ''`?

xclaesse:
otoh, this is always going to be at best a guess. Maybe what we really need is adding an option, we discussed recently we should allow modules to add their own options. Currently every project that install python modules pretty much have to add their own option which makes it inconsistent.

MathieuDuponchelle:
@eli-schwartz I remember writing this specific line, but I really don't remember the reason, should have added an inline comment probably :(

PR #9159, opened 2021-08-23
java module: fix FeatureNew version

eli-schwartz:
It should apply to the *next* stable release.

Reported-by: Tristan Partin <tristan@partin.io>


tristan957:
This is what happens when your PR spans releases :). Thanks for cleanup once again.

PR #9160, opened 2021-08-23
Delete old outputs that are no longer in the Ninja file.

jpakkane:


mesonbuild/backend/ninjabackend.py
```diff
@@ -572,6 +572,7 @@ def generate(self):
         # Refresh Ninja's caches. https://github.com/ninja-build/ninja/pull/1685
         if mesonlib.version_compare(self.ninja_version, '>=1.10.2') and os.path.exists('.ninja_deps'):
             subprocess.call(self.ninja_command + ['-t', 'restat'])
+            subprocess.call(self.ninja_command + ['-t', 'cleandead'])
```
dcbaker, line 575: I don't suppose you can call two -t options at the same time?

dcbaker (approved):

PR #9163, opened 2021-08-24
compile: Add MESONFLAGS equivalent of MAKEFLAGS

xclaesse:
When meson is used indirectly within some build scripts, it is not
always easy to pass options to control number of jobs or enable verbose
compilation.

Ninja famously does not want to support NINJAFLAGS despite being a very
popular request (https://github.com/ninja-build/ninja/issues/1482).
Since Meson wraps ninja already, we could do it at meson level.


eli-schwartz:
Hmm, but why don't people just use samurai instead? That is what I use for all my wrapped build scripts (which do not use meson compile either).

For me, this PR would mean that MESONFLAGS which I don't use would override SAMUFLAGS which I do use, to no real benefit.

Aside for that, MESONFLAGS is a confusing name since it isn't flags for meson, but flags for meson compile.

xclaesse:
Advantage is it works regardless of the backend, which is the whole point of `meson compile` command line. If you use samurai, then just don't use MESONFLAGS... otoh, is there even a reason you're using samurai other than SAMUFLAGS support?

Regarding the name, it's always the hardest part... I wanted something that looks the same as existing SAMUFLAGS and MAKEFLAGS since it has the same purpose. Also it works for vs/xcode backends so I didn't want NINJA in the name...

xclaesse:
Note that currently most scripts probably use ninja command directly, making this useless... But I think we have been encouraging people to use `meson compile` wrapper now, doc has been updated, and this could be a new incentive for that. I personally use `meson compile` in the build/ci scripts I write nowadays.

eli-schwartz:
Samurai also reduces the bootstrap graph by not requiring cmake, C++, or re2c.

It supports SAMUFLAGS, which is pretty important for scripting builds, and also doesn't bizarrely delete colored output requested by GCC `CFLAGS=-fcolor-diagnostics=always`.

meson compile meanwhile doesn't support the majority of ninja or samurai options, and requires you to still know all about the backend in order to pass --ninja-args. It doesn't work consistently on all versions of meson. It doesn't support all *targets* of build.ninja. it used to be broken if you had samurai installed instead of ninja (e.g. you are on alpine) until I fixed it because I'm weirdly invested in it not being downright broken even though I hate it.

I've been recommending everywhere that I don't know what earthly reason it exists for, other than to document in a README.md ***purely for the sake of people who don't know what meson backends are*** because you don't want to have to teach Windows users how to tell if they are using VS, and when to run msbuild or xcodebuild or ninja. But I strongly recommend people with even a 1% clue to just use ninja directly.

jpakkane:
The stated design goals of Meson have always been:

- There should only be one, and only one, way of configuring something
- Never use environment variables for anything if you can help it

This violates both of these.

In addition, this is only useful for a case where you don't invoke Meson directly (because then you could use `-j` directly) but instead via some wrapper thingy W. If W does not provide you a way to control the number of parallel jobs, then that is _a flaw in the wrapper program_. It should be fixed there. Using environment variables like this to bypass things is an abstraction layer violation.

Further, this thing unfortunately has a slippery slope written all over it. People _will_ start demanding all other sorts of things to set via envvars. If we refuse they will complain loudly and bitterly that "well _I_ don't see why we can't _just_ add add an envvar for this, since you can already do A, B, C and D with envvars". I'd imagine that eventually they will request ways to set command line arguments differently for different subcommands because "it's _just_ a simple if/else, what is the problem".

So, in conclusion, very much not a fan.

xclaesse:
>     * There should only be one, and only one, way of configuring something
>     * Never use environment variables for anything if you can help it

You lost already on both of those goals on day 1, let's be honest. Meson uses env for millions things and we added options to override them constantly over time.

> If W does not provide you a way to control the number of parallel jobs, then that is _a flaw in the wrapper program_

A flaw in a build script? Let's be honest, that's the norm not the exception.

> Further, this thing unfortunately has a slippery slope written all over it

We have been on that slope since day 1, that's the world of build systems, like it or not.

eli-schwartz:
I guess @jpakkane is making the argument:

Just because we're already on the slippery slope isn't a reason to shout "Geronimo" and take a running jump off of a steep incline.

And it seems like a poor reason to add an environment variable that must be manually set up in order to use it, to work around the fact that build environments don't currently manually set up the existing method for controlling the same thing.

Given people need to change something anyway, and it only works for o
1 of 3 different code paths, why not recommend the use of samurai instead -- that works for 2 code paths without any changes to meson, and works on any previous version of meson too, and does it at a semantically appropriate layer even.

No need to add yet another environment variable for people to remember, which is distinct to a third tool

xclaesse:
> Just because we're already on the slippery slope isn't a reason to shout "Geronimo" and take a running jump off of a steep incline.

True, but it's not really anything crazy here, number of jobs is already something configurable, I stated a pretty reasonable use-case, and most (to exception of ninja) does it too. It's even one of the key selling point of samurai. I'm not at all for adding env for everything, but on a case-by-case basis when it makes sense, I don't see the problem.

> And it seems like a poor reason to add an environment variable that must be manually set up in order to use it, to work around the fact that build environments don't currently manually set up the existing method for controlling the same thing.
> 
> Given people need to change something anyway, and it only works for o
> 1 of 3 different code paths, why not recommend the use of samurai instead -- that works for 2 code paths without any changes to meson, and works on any previous version of meson too, and does it at a semantically appropriate layer even.

I don't understand what you mean here. One of they key point of using `meson compile` is it works with samurai, ninja, vs and xcode. MESONARGS would work with them all too, unlike SAMUFLAGS that works with only one of the 4.

Granted, it's pretty useless right now because most scripts uses ninja directly, but we have to start somewhere. In the future it could become more common to use `meson compile` wrapper.

jhasse:
> Samurai also reduces the bootstrap graph by not requiring cmake, C++, or re2c.

Neither CMake nor re2c are required to build Ninja.

> [...] doesn't bizarrely delete colored output requested by GCC `CFLAGS=-fcolor-diagnostics=always`.

Ninja won't do that when invoked from a terminal or when `CLICOLOR_FORCE=1` is set.

xclaesse:
> > [...] doesn't bizarrely delete colored output requested by GCC `CFLAGS=-fcolor-diagnostics=always`.
> 
> Ninja won't do that when invoked from a terminal or when `CLICOLOR_FORCE=1` is set.

Not sure where that comes from, maybe an old version of ninja? At least here I get colored warnings by default with meson/ninja, no need to pass any CFLAGS.

PR #9164, opened 2021-08-25
dependency: fallback and allow_fallback are mutually exclusive

xclaesse:
That used to abort in previous Meson versions but 0.59 stopped
forbidding that by mistake.


xclaesse:
Inspired from https://gitlab.gnome.org/GNOME/libsoup/-/merge_requests/249, I actually think we should allow `dependency('foo', fallback: ['foo', 'foo_dep'], allow_fallback: host_system != 'windows')` to use the fallback subproject but not on Windows.

But first let's restore previous behavior in 0.59.2, we can add new behavior for 0.60 afterward.

PR #9170, opened 2021-08-26
optinterpreter: Normalize options and check for dups

xclaesse:
Meson always favoured underscore instead of dash, but in practice it's
an unconsistent mess in most projects. Instead of having to remember
which option is named foo-bar or foo_bar, better treat them the same way
so -Dfoo-bar=val and -Dfoo_bar=val works the same.

While at it, also make option name case unsensitive.


xclaesse:
I still think this PR would be nice, but since everyone else disagree, I opened https://github.com/mesonbuild/meson/pull/10296 to be at least able to rename an option.

PR #9171, opened 2021-08-26
Add typing to msubprojects.py

tristan957:


tristan957:
@xclaesse are you ok if I change `parse_patch_url()` to return a tuple[str, int]? Every where else in wraptool.py uses int. I don't find fixing the typing fallout that you worked with in your last commit.

xclaesse:
Where does it need to be int?

mesonbuild/msubprojects.py
```diff
@@ -148,23 +148,23 @@ def git_rebase(self, revision):
...
         except GitException as e:
-            self.log('  -> Could not reset', mlog.bold(repo_dir), 'to', mlog.bold(revision))
+            self.log('  -> Could not reset', mlog.bold(self.repo_dir), 'to', mlog.bold(revision))
             self.log(mlog.red(e.output))
             self.log(mlog.red(str(e)))
             return False
         return True
 
-    def git_checkout(self, revision, create=False):
+    def git_checkout(self, revision: str, create: bool=False) -> bool:
```
dcbaker, line 164: PEP8 style is no spaces with no anotations, spaces with annotations:
```python
def foo(
  x=None,
  y: Optional[str] = None
)
```

mesonbuild/msubprojects.py
```diff
@@ -520,8 +520,8 @@ def add_arguments(parser):
     p.set_defaults(subprojects_func=Runner.purge)
     p.set_defaults(post_func=Runner.post_purge)
 
-def run(options):
-    src_dir = os.path.relpath(os.path.realpath(options.sourcedir))
+def run(options: argparse.Namespace) -> int:
+    src_dir = os.path.relpath(os.path.realpath(T.cast(str, options.sourcedir)))
```
dcbaker, line 524: There's a better option for this, you can write something like:
```python
if T.TYPE_CHECKING:
    from typing_extensions import Protocol

    class Arguments(Protocol):
         sourcedir: str
         types: str
         num_processes: bool

def run(options: 'Arguments):
    ...
```

dcbaker: You may have to cast to the 'Arguments' somewhere, but it's still better in that once you do the single cast everything is correct.

dcbaker (changes requested):
I've got a few small suggestions, otherwise this looks like a nice cleanup.

tristan957:
@xclaesse referring to the return type on that function.

xclaesse:
> @xclaesse referring to the return type on that function. [38db460#diff-59bb6925ad442d3aeb1948303aa6ae52a92997dd6b58f47d4c33486b16aa2187R101](https://github.com/mesonbuild/meson/commit/38db4602d8f80c3d2aec80ed527e3eed3c41b08c#diff-59bb6925ad442d3aeb1948303aa6ae52a92997dd6b58f47d4c33486b16aa2187R101)

I changed it to str otherwise it compares str and int there: https://github.com/mesonbuild/meson/blob/master/mesonbuild/wrap/wraptool.py#L213. I don't think we ever need the revision as an integer, do we?

tristan957:
I don't really care. It just needs to be consistent. I don't trust greater than or less than comparisons with strings however.

xclaesse:
> I don't trust greater than or less than comparisons with strings however

Sure but here it is `==` which is fine. I'm fine with changing to int, but then you need to be careful of that comparison there. But mypy would warn you I guess.

run_mypy.py
```diff
@@ -45,6 +46,7 @@
     'mesonbuild/mtest.py',
     'mesonbuild/optinterpreter.py',
     'mesonbuild/programs.py',
+    'mesonbuild/wrap/wraptool.py',
```
eli-schwartz, line 49: Don't list this in "specific files", it is already listed above in "fully typed modules".

This is the cause of the CI failure: 
```
mesonbuild/wrap/wraptool.py: error: Duplicate module named "mesonbuild.wrap.wraptool" (also at "mesonbuild/wrap/wraptool.py")
```

eli-schwartz (changes requested):

mesonbuild/msubprojects.py
```diff
@@ -381,12 +414,14 @@ def download(self):
             return False
         return True
 
-    def foreach(self):
+    def foreach(self) -> bool:
+        options = T.cast(ForeachArguments, self.options)
```
eli-schwartz, line 418:
```suggestion
        options = T.cast('ForeachArguments', self.options)
```

This is only defined during type checking, so quoting it will ensure it is doing deferred evaluation.

mesonbuild/msubprojects.py
```diff
@@ -398,28 +433,30 @@ def foreach(self):
         return True
 
     def purge(self) -> bool:
+        options = T.cast(PurgeArguments, self.options)
```
eli-schwartz, line 436:
```suggestion
        options = T.cast('PurgeArguments', self.options)
```

mesonbuild/msubprojects.py
```diff
@@ -89,7 +116,9 @@ def update_wrapdb_file(self):
             wraptool.update_wrap_file(self.wrap.filename, self.wrap.name, new_branch, new_revision)
             self.log('  -> New wrap file downloaded.')
 
-    def update_file(self):
+    def update_file(self) -> bool:
+        options = T.cast(UpdateArguments, self.options)
```
eli-schwartz, line 120:
```suggestion
        options = T.cast('UpdateArguments', self.options)
```

eli-schwartz (dismissed):
Although I have no particular idea how this works and am unsure why it can't just get the arguments correct the first time :/

tristan957:
I will clean up the git history after adding @eli-schwartz suggestions later. Hopefully that was enough to get the pipeline passing although I didn't investigate the failures originally.

mesonbuild/msubprojects.py
```diff
@@ -74,14 +105,14 @@ def __init__(self, logger: Logger, r: Resolver, wrap: PackageDefinition, repo_di
         self.wrap = self.wrap_resolver.wrap = wrap
```
tristan957, line 105: What should I do about the line above?

Seems to be an abomination where self.wrap_resolver is some copy of `Resolver`, but has an extra field.

mesonbuild/msubprojects.py
```diff
@@ -65,23 +96,23 @@ def done(self, wrap_name: str, log_queue: T.List[T.Tuple[mlog.TV_LoggableList, T
 
 
 class Runner:
-    def __init__(self, logger: Logger, r: Resolver, wrap: PackageDefinition, repo_dir: str, options: argparse.Namespace) -> None:
+    def __init__(self, logger: Logger, r: Resolver, wrap: PackageDefinition, repo_dir: str, options: 'Arguments') -> None:
         # FIXME: Do a copy because Resolver.resolve() is stateful method that
         # cannot be called from multiple threads.
         self.wrap_resolver = copy.copy(r)
-        self.wrap_resolver.current_subproject = 'meson'
```
tristan957, line 72: Is this right?

mesonbuild/msubprojects.py
```diff
@@ -15,6 +15,37 @@
...
+if T.TYPE_CHECKING:
+    from typing_extensions import Protocol
+
+    class Arguments(Protocol):
+        sourcedir: str
+        num_processes: int
+        subprojects: T.List[str]
+        types: str
+        # FIXME: Add narrowed type; descriptor protocol...
+        subprojects_func: T.Any
```
tristan957, line 27: Anyone know a better type here?

mesonbuild/msubprojects.py
eli-schwartz: It was correct until I removed the code which accessed `.current_subproject` a week later in commit f52a5a7cd3f30645c1b746488ec921451545d9b6.

Now it is just unneeded...

mesonbuild/msubprojects.py
dcbaker: The only place I see this explicitly set is a function of type `() =? bool`, so you could write `T.Callable[[], bool]`, not sure if that's valid in all cases though.

tristan957: I'll give it a try in my next iteration, while I also try to fix the tests.

tristan957: https://github.com/mesonbuild/meson/pull/9171/files#diff-cb814916ae23be3d4674fc49efca2710d088dc27ac3513e45d4a9cba1fa9bb3bR107

Can you review this if you get a chance?

mesonbuild/msubprojects.py
```diff
@@ -15,23 +15,54 @@
...
 
+if T.TYPE_CHECKING:
+    from typing_extensions import Protocol
+
+    class Arguments(Protocol):
+        sourcedir: str
+        num_processes: int
+        subprojects: T.List[str]
+        types: str
+        # FIXME: Add narrowed type; descriptor protocol...
```
dcbaker, line 26: Is this fixme still relavent?

tristan957: Kind of. This is reference to https://github.com/mesonbuild/meson/pull/9171#discussion_r753945340

tristan957: Notice the `type: ignore` in the PR.

mesonbuild/msubprojects.py
```diff
@@ -65,23 +96,22 @@ def done(self, wrap_name: str, log_queue: T.List[T.Tuple[mlog.TV_LoggableList, T
...
         # cannot be called from multiple threads.
         self.wrap_resolver = copy.copy(r)
-        self.wrap_resolver.current_subproject = 'meson'
         self.wrap_resolver.dirname = os.path.join(r.subdir_root, wrap.directory)
         self.wrap = self.wrap_resolver.wrap = wrap
         self.repo_dir = repo_dir
         self.options = options
-        self.run_method = options.subprojects_func.__get__(self)
-        self.log_queue = []
+        self.run_method: T.Callable[[], bool] = options.subprojects_func.__get__(self) # type: ignore
```
dcbaker, line 107: Wait, what? why are we assigning the `descriptor get to this? I know you don't write this, but WAT.

tristan957: Yea, I don't quite understand it or how it works.

dcbaker (approved):
Other than the fixme that looks like it might not be relevant anymore, this looks good to me.

I'm still wondering why were' assigning the `__get__` method, but that's not your fault, so…

tristan957:
If there is something I can fix regarding to `__get__`, let me know. Otherwise I will just leave it as is for now. cc @xclaesse maybe when you rewrote this file?

PR #9173, opened 2021-08-26
docs: Use a custom hotdoc theme

mensinda:
The main changes are a different base bootstrap theme, a bigger font size, and better scaling for bigger monitors.

Preview: https://mensinda.github.io/meson/index.html

This currently also includes the new RefMan changes since I only have one gh-pages setup...

---

Edit: We should probably move https://github.com/mensinda/meson_bootstrap_theme to the mesonbuild org on GitHub if this gets accepted.


tristan957:
Not sure I am much of a fan of the new theme. Seems like website could stay the same but change the accent color from green to Meson's purple. Change the background color to increase contrast.

mensinda:
My main problem with the current website is that the font size is too small for my taste on my 1440p monitor. Additionally, there is a lot of wasted space on the sides. Additionally, the contrast is not great.

These were the main points I tried to fix. As for the color theme, I pretty much just downloaded [this one](https://bootswatch.com/3/darkly/) and called it a day.

PR #9174, opened 2021-08-26
I18n return

### LINKED ISSUE #6227: Get the i18n targets inside the meson files
I need to generate a file from the i18n *.gmo files. This `custom_target` needs to depend on the `<project_id>-gmo` target documented [here](https://mesonbuild.com/i18n-module.html).

This target should be retrievable through a method like :
```
gettext_targets = i18n.gettext('mypackage')
custom_target(
    depends: gettext_targets.target_gmo()
)
```

For now, I think the only way to achieve that is something like that :
```
ninja_gmo = custom_target(command: [ 'ninja', '-C' meson.build_root(), 'mypackage-gmo'])
```

eli-schwartz:
Return something useful from i18n.gettext()

Fixes #6227

### DIFF
--- a/docs/markdown/i18n-module.md
+++ b/docs/markdown/i18n-module.md
@@ -38,6 +38,11 @@ This function also defines targets for maintainers to use:
 * `<project_id>-update-po`: regenerates the `.po` files from current `.pot` file
 * `<project_id>-gmo`: builds the translations without installing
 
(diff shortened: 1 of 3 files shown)


dcbaker (approved):

eli-schwartz:
Rebased + release notes. This technically depends on #9202 in order to safely consume the returned run_targets.

PR #9177, opened 2021-08-27
interpreter: fix IndexError when specifying dependency 'include_type'

rhssk:
Exception is thrown when dependency name is empty and when its `include_type` differs from the default one.

Regression from b6d754a40c.


eli-schwartz:
@nirbheek This could go in 0.59.2 as well...

PR #9181, opened 2021-08-27
mcompile: treat load-average as a float

floppym:
`ninja -l` accepts a double. We should do the same.

Bug: https://bugs.gentoo.org/810655


xclaesse:
@eli-schwartz I like meson compile because it makes ninja a simple implementation detail of meson. It could be replaced by another backend transparently. It is true that currently ninja/samu are the only real backends, but who knows in the future.

PR #9183, opened 2021-08-27
Validate default values for KwargInfo

dcbaker:
Currently we're allowing default values to be values that we don't allow as inputs, that creates some real, confusion as `ContainerTypeinfo` doesn't assume that's going to happen. This series adds assertions to make sure we're doing that right thing, and fixes all the cases of doing the wrong thing.


dcbaker:
Of course mypy doesn't like the NoneType... I'll fix it on Monday

PR #9184, opened 2021-08-27
optinterpreter: Add deprecated kwarg

xclaesse:


xclaesse:
@dcbaker I have seriously no idea how to get mypy happy on this PR, could you try to fix it for me please?

mesonbuild/optinterpreter.py
```diff
@@ -217,34 +151,82 @@ def evaluate_statement(self, node: mparser.BaseNode) -> None:
...
+            raise OptionException(f'Unknown type {opt_type}.')
+        description = kwargs['description'] or opt_name
 
-        description = kwargs.pop('description', '')
-        if not isinstance(description, str):
-            raise OptionException('Option descriptions must be strings.')
+        # Only keep in kwargs arguments that are used by option type's parser
+        # because they use @permittedKwargs().
+        known_parser_kwargs = {'value', 'choices', 'yield', 'min', 'max'}
+        parser_kwargs = {k: v for k, v in kwargs.items() if k in known_parser_kwargs and v is not None}
```
dcbaker, line 184: you can resolve the last mypy error either by forcing the type (I think) or casting (for sure). I think:
```python
parser_kwargs: 'TypeDictUsedByParser' = {k: v for ...}
```
will work, but you might hav to do:
```python
opt = parser(description, T.cast('TypeDictUsedByParser', parser_kwargs))
```

dcbaker:
The first two are more annoying (and we're really pushing what python's type system can do here arleady). I *think* those two can be fixed by changing the initializer for KWargInfo to:
```python
    def __init__(self, name: str,
                 types: T.Union[T.Type[_T], T.Tuple[T.Union[T.Type[_T], ContainerTypeInfo], ...], ContainerTypeInfo],
                 *, required: bool = False, listify: bool = False,
                 default: T.Optional[_T] = None,
                 since: T.Optional[str] = None,
                 since_values: T.Optional[T.Dict[str, str]] = None,
                 deprecated: T.Optional[str] = None,
                 deprecated_values: T.Optional[T.Dict[str, str]] = None,
                 validator: T.Optional[T.Callable[[T.Any], T.Optional[str]]] = None,
                 convertor: T.Optional[T.Callable[[_T], object]] = None,
                 not_set_warning: T.Optional[str] = None):
``

mesonbuild/optinterpreter.py
xclaesse: > parser_kwargs: 'TypeDictUsedByParser' = {k: v for ...}

That did not work, but the cast worked. Thanks.

xclaesse:
>  I _think_ those two can be fixed by changing the initializer for KWargInfo

Wow that worked on first try, great. I was suspecting something like that, but nothing I tried before did it :D

mensinda (approved):
LGTM

eli-schwartz:
Ah, so the gst-editing-services one is essentially identical except that the one that won didn't have a description.

Well okay, as long as we're actually doing the same thing we used to do I guess.

PR #9185, opened 2021-08-27
Make EnvironmentVariablesObject a  holder

dcbaker:
EnvironmentVariablesObject is a somewhat odd class, it's effectively a holder for `build.EnvironmentVariables`, but it it's not. This series converts it into one, and adds some additional type checking and validation.

My motivaiton here is that doing this simplifies add the type checking decorators to the various target interfaces, as we know we'll have an EnviornmentVariables object. This also is nice, in that it means that all cases where an and EnvironmentVariables is expected can get one, instead of each caller having to go through the dance of "is this a list? a dict? a str? an EnviornmentVariablesObject?" Instead that's all done in the inerpreter, and the lower layers always get a ciearn, EnvironmentVaraibles object.

### DIFF
--- a/mesonbuild/interpreter/interpreterobjects.py
+++ b/mesonbuild/interpreter/interpreterobjects.py
@@ -241,7 +241,9 @@ class EnvironmentVariablesObject(MutableInterpreterObject, MesonInterpreterObjec
                              'append': self.append_method,
                              'prepend': self.prepend_method,
                              })
-        if isinstance(initial_values, dict):
+        if isinstance(initial_values, build.EnvironmentVariables):
+            self.vars = initial_values
+        elif isinstance(initial_values, dict):
             for k, v in initial_values.items():
                 self.set_method([k, v], {})
         elif initial_values is not None:
--- a/mesonbuild/interpreter/interpreterobjects.py
+++ b/mesonbuild/interpreter/interpreterobjects.py
@@ -17,7 +17,7 @@ from ..interpreterbase import (
                                ContainerTypeInfo, KwargInfo,
                                InterpreterObject, MesonInterpreterObject, ObjectHolder, MutableInterpreterObject,
                                FeatureCheckBase, FeatureNewKwargs, FeatureNew, FeatureDeprecated,
-                               typed_pos_args, typed_kwargs, stringArgs, permittedKwargs,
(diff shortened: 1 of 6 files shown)


mesonbuild/interpreter/interpreterobjects.py
```diff
@@ -229,76 +234,52 @@ def stdout_method(self, args: T.List[TYPE_var], kwargs: TYPE_kwargs) -> str:
...
-class EnvironmentVariablesObject(MutableInterpreterObject, MesonInterpreterObject):
-    # TODO: Move the type cheking for initial_values out of this class and replace T.Any
-    def __init__(self, initial_values: T.Optional[T.Any] = None, subproject: str = ''):
-        super().__init__(subproject=subproject)
-        self.vars = build.EnvironmentVariables()
+
+_ENV_SEPARATOR_KW = KwargInfo('separator', str, default=os.pathsep)
+
+
+class EnvironmentVariablesObject(ObjectHolder[build.EnvironmentVariables], MutableInterpreterObject):
```
mensinda, line 241: For consistency, this should be called `EnvironmentVariablesHolder`

mesonbuild/interpreter/type_checking.py
```diff
@@ -126,3 +127,52 @@ def _lower_strlist(input: T.List[str]) -> T.List[str]:
     default=True,
     # TODO: extract_required_kwarg could be converted to a convertor
 )
+
+def _env_validator(value: T.Union[EnvironmentVariables, T.List, T.Dict, str, None]) -> T.Optional[str]:
```
mensinda, line 131: This could be more strict/explicit with `T.List[TYPE_var], T.Dict[str, TYPE_var]` (or possibly replace `TYPE_var` with `T.Any`)

mensinda (commented):
I only got a few minor style issues.

mesonbuild/interpreter/interpreterobjects.py
dcbaker: yeah, that makes sense.

mesonbuild/interpreter/type_checking.py
dcbaker: done.

PR #9186, opened 2021-08-27
Simplify get_callee_args

xclaesse:


xclaesse:
@dcbaker @jpakkane while revisiting those decorators, I added a commit that turn unknown kwarg fatal. It has been saying `This will become a hard error in the future.` for years...

mesonbuild/interpreter/mesonmain.py
```diff
@@ -235,7 +235,7 @@ def can_run_host_binaries_method(self, args: T.Tuple[object, ...], kwargs: T.Dic
         return self.can_run_host_binaries_impl(args, kwargs)
 
     def can_run_host_binaries_impl(self, args, kwargs):
-        if (self.is_cross_build_method(None, None) and
+        if (self.is_cross_build_method([], {}) and
```
dcbaker, line 238: Can we just avoid calling the method? It's literally a wrapper around self.environment.is_cross_build

xclaesse: You're right, added a commit that simplify that condition. `if (a) if (b) {}` is same as `if (a and b) {}`

xclaesse: Also the "Need to revisit" part is outdated, it has already been revisited, that's why we have `can_run_host_binaries()` where that condition is not confusing at all.

dcbaker (approved):

PR #9188, opened 2021-08-28
Make assignment a statement in Syntax.md

jimman2003:
force to change line 709 to something kind of non sensical

### DIFF
--- a/docs/markdown/Syntax.md
+++ b/docs/markdown/Syntax.md
@@ -695,7 +695,7 @@ additive_expression: multiplicative_expression | (additive_expression additive_o
 additive_operator: "+" | "-"
 argument_list: positional_arguments ["," keyword_arguments] | keyword_arguments
 array_literal: "[" [expression_list] "]"
-assignment_statement: conditional_expression | (logical_or_expression assignment_operator assignment_expression)
+assignment_statement: expression asssignment_operator expression
 assignment_operator: "=" | "*=" | "/=" | "%=" | "+=" | "-="
 boolean_literal: "true" | "false"
 build_definition: (NEWLINE | statement)*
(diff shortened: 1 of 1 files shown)


docs/markdown/Syntax.md
```diff
@@ -706,7 +706,7 @@ DECIMAL_NUMBER: /[1-9][0-9]*/
 dictionary_literal: "{" [key_value_list] "}"
 equality_expression: relational_expression | (equality_expression equality_operator relational_expression)
 equality_operator: "==" | "!="
-expression: assignment_expression
+expression: assignment_statement
 expression_list: expression ("," expression)*
```
dcbaker, line 710: This still doesn't seem right. Assignment is a statement, but an expression isn't assignment

dcbaker (changes requested):

docs/markdown/Syntax.md
jimman2003: I know ... could really think of a way to resolve this

docs/markdown/Syntax.md
```diff
@@ -695,7 +695,7 @@ additive_expression: multiplicative_expression | (additive_expression additive_o
 additive_operator: "+" | "-"
 argument_list: positional_arguments ["," keyword_arguments] | keyword_arguments
 array_literal: "[" [expression_list] "]"
-assignment_expression: conditional_expression | (logical_or_expression assignment_operator assignment_expression)
+assignment_statement: conditional_expression | (logical_or_expression assignment_operator assignment_expression)
```
dcbaker, line 698: I think this should be `assignment_statement: IDENTIFIER assignment_operator expression`

docs/markdown/Syntax.md
dcbaker: And this should be:
`expression: conditional_expression | (logical_or_expression assignment_operator assignment_expression)`

docs/markdown/Syntax.md
dcbaker: Actually, meson++ has this https://github.com/dcbaker/meson-plus-plus/blob/b7099bd3e8b3d1a2a7c8fea63821993ccb7da7ff/src/frontend/parser.yy#L106,
so I think this should be
`assignment_statement: expression asssignment_operator expression`

dcbaker (approved):

PR #9190, opened 2021-08-29
interpreter: Add IntegerHolder

mensinda:
The goal of this PR is to introduce a foundation for holderifying all the primitive types that currently need special logic in `InterpreterBase`. The end goal is to just have `InterpreterBase` deal with method and operator calls to `InterpreterObjects`.

I have only added an `IntegerHolder` to keep (for once) the diff on the small side. The downside here is that some extra logic in `InterpreterBase` is required to support a mix of primitive holders and old-style primitive handling while the conversion process is in progress.

Finally, the long deprecated/undefined behavior comparisons with `==` and `!=` for objects with different types has been turned into a hard error.

### DIFF
--- a/mesonbuild/interpreterbase/baseobjects.py
+++ b/mesonbuild/interpreterbase/baseobjects.py
@@ -111,6 +111,9 @@ class InterpreterObject:
         ))
 
     def op_equals(self, other: TYPE_var) -> bool:
+        # We use `type(...) == type(...)` here to enforce an *exact* match for comparison. We
+        # don't want comparisons to be possible where `isinstance(derived_obj, type(base_obj))`
+        # would pass because this comparison must never be true: `derived_obj == base_obj`
         if type(self) != type(other):
             self._throw_comp_exception(other, '==')
         return self == other
@@ -148,6 +151,7 @@ class ObjectHolder(InterpreterObject, T.Generic[InterpreterObjectTypeVar]):
 
     # Override default comparison operators for the held object
     def op_equals(self, other: TYPE_var) -> bool:
+        # See the comment from InterpreterObject why we are using `type()` here.
         if type(self.held_object) != type(other):
             self._throw_comp_exception(other, '==')
         return self.held_object == other
--- a/mesonbuild/interpreterbase/baseobjects.py
+++ b/mesonbuild/interpreterbase/baseobjects.py
@@ -13,9 +13,11 @@
 # limitations under the License.
 
 from .. import mparser
-from .exceptions import InvalidCode
+from .exceptions import InvalidCode, InvalidArguments
 from .helpers import flatten, resolve_second_level_holders
-from ..mesonlib import HoldableObject
+from .operator import MesonOperator
+from ..mesonlib import HoldableObject, MesonBugException
+import textwrap
 
 import typing as T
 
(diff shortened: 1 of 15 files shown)


mesonbuild/interpreterbase/baseobjects.py
```diff
@@ -62,13 +88,47 @@ def method_call(
...
+            '''
+        ))
+
+    def op_equals(self, other: TYPE_var) -> bool:
+        if type(self) != type(other):
+            self._throw_comp_exception(other, '==')
+        return self == other
+
+    def op_not_equals(self, other: TYPE_var) -> bool:
+        if type(self) != type(other):
```
dcbaker, line 119: is there a reason not to use:
```python
isisnstance(other, type(self)):
```
here? If there is then please add a comment so this doesn't get refactored later

mesonbuild/interpreterbase/decorators.py
```diff
@@ -110,6 +111,49 @@ def wrapped(*wrapped_args: T.Any, **wrapped_kwargs: T.Any) -> T.Any:
...
+        @wraps(f)
+        def wrapper(self: 'InterpreterObject', other: TYPE_var) -> TYPE_var:
+            if not isinstance(other, types):
+                raise InvalidArguments(f'The `{operator.value}` of {self.display_name()} does not accept objects of type {type(other).__name__} ({other})')
+            return f(self, other)
+        return T.cast('_TV_FN_Operator', wrapper)
+    return inner
+
+def unary_operator(operator: MesonOperator) -> T.Callable[['_TV_FN_Operator'], '_TV_FN_Operator']:
+    """Decorator that does type checking for operator calls.
```
dcbaker, line 142: shouldn't this say "unary operator calls"?

mesonbuild/interpreterbase/baseobjects.py
```diff
@@ -36,17 +38,41 @@
...
 
+        # Some default operators supported by all objects
+        self.operators.update({
+            MesonOperator.EQUALS: self.op_equals,
+            MesonOperator.NOT_EQUALS: self.op_not_equals,
+        })
+
+    # The type of the object that can be printed to the user
+    def display_name(self) -> str:
+        return type(self).__name__
```
dcbaker, line 74: Oh, thanks for this. I've been meaning to do this myself.

dcbaker (changes requested):
I have a few small nits, but otherwise this looks good

mesonbuild/interpreterbase/decorators.py
mensinda: Thanks, fixed.

mesonbuild/interpreterbase/baseobjects.py
mensinda: No, because `isinstance(derived, type(base))` would be true, but not the other way around. Additionally, we really want an exact type match here since `derived_obj` must never equal `base_obj`.

dcbaker: Okay, then please add a comment explaining that, so someone well-meaning doesn't come along and change this to an isinstance check.

mensinda: I already did that.

mensinda: Well ok, in the function 5 lines above but this should still count :)

dcbaker (approved):

PR #9192, opened 2021-08-30
msvc: Assume UTF8 source by default

xclaesse:
Currently every project that uses UTF8 for its source files must add
'/utf-8' argument otherwise they don't work non-English locale MSVC.

Since meson.build itself is assumed to be UTF8 by default, seems better
to assume it for source files by default too.

For example:
- https://gitlab.freedesktop.org/gstreamer/gst-build/-/blob/master/meson.build#L62
- https://gitlab.gnome.org/GNOME/glib/-/blob/main/meson.build#L29

### DIFF
--- a/test cases/windows/18 msvc charset/iso-8859-1.c
+++ b/test cases/windows/18 msvc charset/iso-8859-1.c
@@ -1,0 +1,7 @@
+#include <stdio.h>
+
+int main(int argc, char *argcv[])
+{
+  printf("This is ISO-8859-1 encoded é\n");
+  return 0;
+}
--- a/docs/markdown/snippets/msvc_utf8.md
+++ b/docs/markdown/snippets/msvc_utf8.md
@@ -1,0 +1,16 @@
+## MSVC compiler now assumes UTF-8 source code by default
+
+Every project that uses UTF-8 source files had to add manually `/utf-8` C/C++
+compiler argument for MSVC otherwise they wouldn't work on non-English locale.
+Meson now switched the default to UTF-8 to be more consistent with all other
+compilers.
+
+This can be overridden but using `/source-charset`:
+```meson
+if cc.get_id() == 'msvc'
+  add_project_arguments('/source-charset:.XYZ', language: ['c', 'cpp'])
+endif
+```
+
+See Microsoft documentation for details:
+https://docs.microsoft.com/en-us/cpp/build/reference/source-charset-set-source-character-set.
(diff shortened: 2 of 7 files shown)


xclaesse:
CC @nirbheek 

This is probably not a backward compatible change, I guess it could break some Windows-only projects, but it feels wrong to force the majority to work around a not well known issue, causing endless issues in every meson projects.

nirbheek:
Note: `clang-cl` also supports `/utf8` but it is enabled by default (unlike `cl.exe`). Maybe we want to pass it explicitly anyway in case the default changes?

xclaesse:
I could move it up in the VisualStudioLikeCompiler base class, but I see there is intel-cl too that uses it. Any idea of that one supports the same arg? Btw, is it `/utf8` or `/utf-8`? Please don't tell me it's not consistent between cl and clang-cl.

xclaesse:
Note that if it's the default with clang-cl, that's even more important to make the switch for msvc, better consistency even for pure VisualStudio projects.

xclaesse:
Hmm, if we do this unconditionally, then there is no way for the project to override it: `cl : Command line error D8016 : '/utf-8' and '/source-charset:utf8' command-line options are incompatible`.

As suggested by @jpakkane we could skip adding /utf-8 in the case we find `/source-charset` in c_args already.

xclaesse:
Added release snippet and a unit test to verify user can still override with `/source-charset` and `/execution-charset`. I think this is ready to go if CI pass.

nirbheek:
> Btw, is it `/utf8` or `/utf-8`? Please don't tell me it's not consistent between cl and clang-cl.

It's `/utf-8`, my bad. So it's consistent. :)

test cases/windows/18 msvc charset/iso-8859-1.c
```diff
@@ -0,0 +1,7 @@
+#include <stdio.h>
+
+int main(int argc, char *argcv[])
+{
+  printf("This is ISO-8859-1 encoded �\n");
```
nirbheek, line 5: Can we also have a unit test that this file fails to compile with `/utf-8`? Just in case something changes in the toolchain in the future that breaks everything.

xclaesse: ok, added.

xclaesse:
let's just skip the test if it's not ninja.

PR #9199, opened 2021-08-31
replace `an_unpicklable_object` with proper python dunder methods

dcbaker:
Instead of using an object that happens to be unpicklable to ensure the Interpreter isn't pickled, we can just make it an error to call the dunder methods used for pickling. I've also added a test for this, to ensure it remains working in the future.


xclaesse:
LGTM.

eli-schwartz (approved):

PR #9202, opened 2021-08-31
run_target depends

### LINKED ISSUE #9175: alias_target that depends on a run_target is quite broken
While testing #9174, I tried to create the target I alluded to, there.

In lib/libalpm/po/meson.build I added:

```
lib_mo = i18n.gettext(...) # this used to not be saved

alias_target('regenerate-all-pot', lib_mo[1])
```

In my build.ninja, I ended up with:

```
build regenerate-all-pot: phony  | lib/libalpm/po/libalpm-pot
```

This is plainly wrong!

It's reproducible with a simple meson.build

```
$ cat meson.build
project('foo', 'c')

subdir('in_subdir')
$ cat in_subdir/meson.build
r = run_target('run_a_target', command: ['echo', 'ran a target'])
alias_target('aliased_run_a_target', r)
```

However, it is not reproducible in released meson at all, because instead I get:

```
in_subdir/meson.build:2:0: ERROR: Depends items must be build targets.
```

This seems like a regression from commit 982973f303e3d0431d9b79f89cdf2c8f1ee67bc0 which removed that error message and made it accept any build.Target -- which I then assumed should work while testing my PR...
/cc @dcbaker 

I would actually sort of kind of like this to work properly as I'd like an alias_target that lets me run the i18n.gettext() pseudo-targets for multiple gettext domains in a single ninja invocation. Currently we have a 40-line shell script which spends most of its time trying to find the builddir.

eli-schwartz:
run_target: do not yield broken names with subdirs

A run_target object created in a subdir/meson.build always has a ninja rule name of "name", not "subdir/name".

Fixes #9175

### DIFF
--- a/unittests/allplatformstests.py
+++ b/unittests/allplatformstests.py
@@ -3076,6 +3076,8 @@ class AllPlatformTests(BasePlatformTests):
         self.run_target('build-all')
         self.assertPathExists(os.path.join(self.builddir, 'prog' + exe_suffix))
         self.assertPathExists(os.path.join(self.builddir, 'hello.txt'))
+        out = self.run_target('aliased-run')
+        self.assertIn('a run target was here', out)
 
     def test_configure(self):
         testdir = os.path.join(self.common_test_dir, '2 cpp')
--- a/docs/markdown/snippets/run_target-depends.md
+++ b/docs/markdown/snippets/run_target-depends.md
@@ -1,0 +1,20 @@
+## `run_target` can now be used as a dependency
+
+A `run_target()` can now be saved in a variable and reused as a dependency in
+an `alias_target()`. This can be used to create custom alias rules that ensure
+multiple other targets are run, even if those targets don't produce output
+files.
+
+For example:
+
+```
+i18n = import('i18n')
+
+all_pot_targets = []
+
+foo_i18n = i18n.gettext('foo')
+
+all_pot_targets += foo_i18n[1]
+
+alias_target('all-pot', all_pot_targets)
+```
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -358,7 +358,10 @@ class Backend:
 
     @lru_cache(maxsize=None)
     def get_target_dir(self, target: T.Union[build.Target, build.CustomTargetIndex]) -> str:
-        if self.environment.coredata.get_option(OptionKey('layout')) == 'mirror':
+        if isinstance(target, build.RunTarget):
+            # this produces no output, only a dummy top-level name
+            dirname = ''
+        elif self.environment.coredata.get_option(OptionKey('layout')) == 'mirror':
             dirname = target.get_subdir()
         else:
             dirname = 'meson-out'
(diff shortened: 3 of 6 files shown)


dcbaker (approved):

xclaesse:
This definitely need a unit test. Great that this can be made to work, it has been a recurrent request. Does it mean that we now can do things like that too?

```meson
rt = run_target()
run_target(..., depends: rt)
custom_target(..., depends: rt)
```

xclaesse:
Also, what happens on rebuild? Does it execute the run_target() again every time one of the inputs of the custom_target() changed? Or does it consider the run_target() to be always stale and it will rebuild the custom_target() again on every rebuild? All those questions should be answered by a unit test :D

eli-schwartz:
This only works for adding a dep to an alias_target, not a custom_target.

The latter doesn't accept all build.Target subclasses, and since both alias_target and run_target are targets of type "phony" they
- produce no output
- are not run in the "all" default target (build_by_default: true)
- when run manually via `ninja <target>` they are always considered out of date

Adding one to a custom_target would make that custom target need to be rebuilt on every single ninja invocation which is weird. :D

...

On the test front, I can test with a simple unit test that it successfully configured, but I don't think I can test that the target runs and does something? Is there prior art in testing a run_target?

xclaesse:
> This only works for adding a dep to an alias_target, not a custom_target.

It's sad and good at the same time. I would like to see this possible to have run_target() and custom_target() depend on another run_target(), but better leave it not allowed if unsure.

> Is there prior art in testing a run_target?

Yes of course, even alias_target() has a unit test: AllPlatformTests.test_alias_target()

... while checking that test, I see vs backend is broken already... oh well, as long as nobody complains...

eli-schwartz:
OK, let's see if this works. Extended that very test to check for an alias_target successfully running a run_target inside a subdirectory.

unittests/allplatformstests.py
```diff
@@ -3076,6 +3076,11 @@ def test_alias_target(self):
         self.run_target('build-all')
         self.assertPathExists(os.path.join(self.builddir, 'prog' + exe_suffix))
         self.assertPathExists(os.path.join(self.builddir, 'hello.txt'))
+        expected = 'a run target was here'
+        out = self.run_target('aliased-run')
+        out_start = out.find(expected)
+        out_line = out[out_start:].rstrip('\n')
+        self.assertEqual(expected, out_line)
```
xclaesse, line 3083: `self.assertIn(b'a run target was here', out)` should be enough.

xclaesse: Could also self.assertNotIn() when building `build-all`.

xclaesse: That one is particular, the summary text is made with dictionaries that does not guarantee ordering with python 3.6. So some lines of that output can be reordered depending on you python version.

eli-schwartz: Ah, I was wondering if maybe it was trying to be strict about like, whole-line searches or something.

LAK132:
These particular targets generate folders of stuff, and the second one generates stuff in the same folder. cunstom_targets can't handle commands that generate folders

PR #9203, opened 2021-09-01
interpreter: Add summary of all user defined options

xclaesse:
It is a commonly needed information to help debugging build issues. We
    already were printing options with non-default value at the end of the
    configure but outside of the summary.
    
    Keeping the list of user defined options in the interpreter will also in
    the future be useful to use new default value on reconfigure.


xclaesse:
Sans titre

mesonbuild/mesonlib/universal.py
```diff
@@ -2104,13 +2104,18 @@ def __eq__(self, other: object) -> bool:
...
+        elif self.subproject != other.subproject:
+            return self.subproject < other.subproject
+        elif self.type != other.type:
+            return self.type < other.type
+        elif self.lang != other.lang:
+            return self.lang < other.lang
+        elif self.machine != other.machine:
+            return self.machine < other.machine
+        else:
+            return self.name < other.name
```
mensinda, line 2118: I think you can use `(self.subproject, self.type, self.lang, self.machine, self.name) < (other.subproject, other.type, other.lang, other.machine, other.name)` here.

mesonbuild/interpreter/interpreter.py
```diff
@@ -70,6 +70,7 @@
 import typing as T
 import textwrap
 import importlib
+import argparse
```
mensinda, line 73: Since this only affects typing stuff this should go into the `if T.TYPE_CHECKING:` block. You will have to use `'argparse.Namespace'` then but it reduces the number of actual imports.

However, this is only a very minor style nitpick.

mensinda (commented):
Mostly LGTM besides:

mesonbuild/mesonlib/universal.py
xclaesse: oh nice, didn't know that syntax, thanks. Updated.

mesonbuild/interpreter/interpreter.py
xclaesse: done.

PR #9205, opened 2021-09-01
new custom dependency lookup for iconv

eli-schwartz:
Also internally needed by intl, so add that as a proxied dependency instead of coding it manually.

### DIFF
--- a/mesonbuild/dependencies/misc.py
+++ b/mesonbuild/dependencies/misc.py
@@ -448,6 +448,25 @@ class CursesSystemDependency(SystemDependency):
                 break
 
 
+class IconvBuiltinDependency(BuiltinDependency):
+    def __init__(self, name: str, env: 'Environment', kwargs: T.Dict[str, T.Any]):
+        super().__init__(name, env, kwargs)
+
+        if self.clib_compiler.has_function('iconv_open', '', env)[0]:
+            self.is_found = True
+
+
+class IconvSystemDependency(SystemDependency):
+    def __init__(self, name: str, env: 'Environment', kwargs: T.Dict[str, T.Any]):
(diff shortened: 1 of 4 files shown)


dcbaker (changes requested):
This needs to be added to the lookup table, right?

eli-schwartz:
Oh darn, knew I was forgetting something. :D

dcbaker (approved):

dcbaker (changes requested):
Oh, wait, there's another docs page for built-in dependencies

dcbaker (approved):
Looks great. Thanks for bearing with me as i remember all of the things to do.

mesonbuild/dependencies/misc.py
```diff
@@ -467,7 +486,9 @@ def __init__(self, name: str, env: 'Environment', kwargs: T.Dict[str, T.Any]):
             self.is_found = True
 
             if self.static:
-                self.link_args += self.clib_compiler.find_library('iconv', env, [], self.libtype)
+                if not self._add_sub_dependency(iconv_factory(env, self.for_machine, {})):
```
lazka, line 489: Could it be that this makes it no longer forward "static" to iconv?

dcbaker: Yes, that would be the case

eli-schwartz: I... think that was largely copy/paste from other uses of add_sub_dependency, which almost universally don't pass the libtype in.

dcbaker: Likely, i think all the other uses are threads, which is kinda special

lazka: -> #9722

PR #9206, opened 2021-09-01
make the display name of a gettext translation build, look prettier

eli-schwartz:
And more accurate too, TBH. Currently it says it is building "lang.mo", even though it is actually building "domain.mo" inside lang/LC_MESSAGES/.

Since meson loudly complains if I try to name the display name "lang/domain.mo", name it with a dash instead of a slash. The actual name isn't a priority here IMO, and this is nicely readable.

### DIFF
--- a/mesonbuild/modules/i18n.py
+++ b/mesonbuild/modules/i18n.py
@@ -184,7 +184,7 @@ class I18nModule(ExtensionModule):
                           'install_dir': path.join(install_dir, l, 'LC_MESSAGES'),
                           'install_tag': 'i18n',
                           }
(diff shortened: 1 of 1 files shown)


xclaesse:
LGTM

I'm currious why meson complains when there are / in the name. Last time I checked custom_target() name is litterally used only in the ninja rule description. We could even make it optional to have a name at all and default to outputs[0].

eli-schwartz:
I dunno, it confused me too for the same reason.

eli-schwartz:
I tried kicking the CI twice and the same azure pipeline keeps on timing out after 2 hours. This PR passes everywhere else and can't possibly affect the CI here anyway, this is obviously a transient failure.

PR #9207, opened 2021-09-01
interpreter: Introduce StringHolder

mensinda:
Another commit in my quest to rid InterpreterBase from all higher
level object processing logic.

Additionally, there is a a logic change here, since `str.join` now
uses varargs and can now accept more than one argument (and supports
list flattening).

### DIFF
--- a/mesonbuild/interpreter/primitives/string.py
+++ b/mesonbuild/interpreter/primitives/string.py
@@ -21,6 +21,11 @@ from ...interpreterbase import (
 
     InvalidArguments,
 )
+from ...mparser import (
+    MethodNode,
+    StringNode,
+    ArrayNode,
+)
 
 
 if T.TYPE_CHECKING:
@@ -105,6 +110,13 @@ class StringHolder(ObjectHolder[str]):
     @noKwargs
     @typed_pos_args('str.join', varargs=str)
     def join_method(self, args: T.Tuple[T.List[str]], kwargs: TYPE_kwargs) -> str:
+        # Implement some basic FeatureNew check on the AST level
+        assert isinstance(self.current_node, MethodNode)
+        n_args = self.current_node.args.arguments
+        if len(n_args) != 1 or not isinstance(n_args[0], ArrayNode) or not all(isinstance(x, StringNode) for x in n_args[0].args.arguments):
+            FeatureNew.single_use('str.join (varargs)', '0.60.0', self.subproject, 'List-flattening and variadic arguments')
+
+        # Actual implementation
         return self.held_object.join(args[0])
 
     @noKwargs
--- a/mesonbuild/interpreter/primitives/string.py
+++ b/mesonbuild/interpreter/primitives/string.py
@@ -64,6 +64,7 @@ class StringHolder(ObjectHolder[str]):
         # Use actual methods for functions that require additional checks
         self.operators.update({
             MesonOperator.DIV: self.op_div,
+            MesonOperator.INDEX: self.op_index,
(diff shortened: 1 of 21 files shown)


eli-schwartz:
> Additionally, there is a a logic change here, since `str.join` now uses varargs and can now accept more than one argument (and supports list flattening).

I don't see a FeatureNew for this.

mensinda:
> I don't see a FeatureNew for this.

That's because there is no way to check for this after the refactoring. `@typed_pos_args('str.join', varargs=str)` does the conversion and there is no way to get to the original value. Additionally, methods don't know about the original Function node, so there is also no way to do that check on the AST level...

mensinda:
Nevermind, I forgot that the `InterpreterObject`s also store the current node. It's not pretty, but it works.

mensinda:
ping

mesonbuild/interpreter/primitives/string.py
```diff
@@ -0,0 +1,193 @@
...
+from ...mparser import (
+    MethodNode,
+    StringNode,
+    ArrayNode,
+)
+
+import re
+from pathlib import PurePath
+
+import typing as T
```
dcbaker, line 28: minor nit, in PEP8 imports should be:
```python
import python_builtins

import thrid_party

import local
```

dcbaker (approved):
I have one small style nit, otherwise this looks great to me. I'm happy to not see the v2, my reviewed-by stands.

jpakkane:
:crossed_swords: conflicts.

mesonbuild/interpreter/primitives/string.py
```diff
@@ -0,0 +1,194 @@
...
+
+        return re.sub(r'@(\d+)@', arg_replace, self.held_object)
+
+    @noKwargs
+    @typed_pos_args('str.join', varargs=str)
+    def join_method(self, args: T.Tuple[T.List[str]], kwargs: TYPE_kwargs) -> str:
+        # Implement some basic FeatureNew check on the AST level
+        assert isinstance(self.current_node, MethodNode)
+        n_args = self.current_node.args.arguments
+        if len(n_args) != 1 or not isinstance(n_args[0], ArrayNode) or not all(isinstance(x, StringNode) for x in n_args[0].args.arguments):
```
eli-schwartz, line 116: I've just accidentally created a case where this is triggering:

https://github.com/mesonbuild/wrapdb/pull/176/files#diff-8c713a7313f3f36bd6f36590315691da5d03b76273f019ad09cd8dfcbbb8a3f1R6

Apparently this is an IndexNode instead of a StringNode.

dcbaker: Since this is in we need an issue we can attach to the 0.60 milestone so we don't accidentally release with this.

eli-schwartz: @mensinda Thoughts on fixing this?

mensinda: Oh... This looks like it's going to be a huge pain. The only real way of fixing I can think of is to call `evaluate_statement` again on the nodes to check the type. But then this will break: `', '.join([run_command('./must_only_be_called_once.sh').stdout()])`.

I really don't know how to properly implement this check then... Would it be ok to completely remove it or has someone a better idea?

eli-schwartz: Since we are right on top of rc1 and no one has proposed a better idea, let's remove it. #9382

PR #9210, opened 2021-09-02
mdist: use git archive with git repositories for correctness

### LINKED ISSUE #2287: ninja dist should respect gitattributes export-ignore
Quoting from the [manpage](https://manpages.debian.org/unstable/git-man/gitattributes.5.en.html#Creating_an_archive) for gitattribute

`Files and directories with the attribute `export-ignore` won’t be added to archive files.`

It would be great if `ninja dist` would respect that and not include files with that attribute in the tarball it generates.

### LINKED ISSUE #3081: Exclude files from ninja dist?
Looking into meson bu I could not find anything in the docs how to exclude files from ``ninja dist``. I don't want to have my transiflex or travis configuration files included in the tarball.

### LINKED ISSUE #8144: dist failure with relative submodule paths

eli-schwartz:
Fixes various inconsistencies:
- gitattributes is respected
  - export-subst
  - export-ignore
- submodules with relative paths are not checked out relative to the
  local clone (which does not work anyway)
- no need to manually remove gitfiles with inaccurate heuristics

Fixes #2287
Fixes #3081
Fixes #8144

### DIFF
--- a/mesonbuild/mdist.py
+++ b/mesonbuild/mdist.py
@@ -18,6 +18,8 @@ import os
 import sys
 import shutil
 import subprocess
+import tarfile
+import tempfile
 import hashlib
 import json
 from glob import glob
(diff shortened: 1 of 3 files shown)


xclaesse:
I've always wondered why we were not doing that and have been assuming @jpakkane had a reason.

eli-schwartz:
The lint_mypy failure is not actually my fault. @mensinda is there some way to automatically tell that open() doesn't take an encoding parameter because it's from the tarfile module, not the built-in function or io.open?

eli-schwartz:
@xclaesse the reason is "because submodules looked hard". :)

mensinda:
Yes, there is. You can turn it of with a comment. We already use tarfile somewhere and it should be there on the same line.

Can't be more specific because I can't look it up right now.

On 2 September 2021 14:12:06 CEST, Eli Schwartz ***@***.***> wrote:
>The lint_mypy failure is not actually my fault. @mensinda is there some way to automatically tell that open() doesn't take an encoding parameter because it's from the tarfile module, not the built-in function or io.open?
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/mesonbuild/meson/pull/9210#issuecomment-911604924
-- 

eli-schwartz:
I was hoping for something slightly more "automatic", but I guess the manually manual way works too...

xclaesse:
@eli-schwartz fair enough. Note that we now also support dist for a subproject that is in the same git repo of its parent project. We're going to use that feature for GStreamer. IIRC I made a unit test for that, so CI should tell if that still works.

eli-schwartz:
Maybe that's the reason why CI fails in the "samerepo" setup...

xclaesse:
> The lint_mypy failure is not actually my fault. @mensinda is there some way to automatically tell that open() doesn't take an encoding parameter because it's from the tarfile module, not the built-in function or io.open?

Probably that syntax: `tar = tarfile.open(xz_distfile, "r:xz")  # [ignore encoding]`

xclaesse:
> Maybe that's the reason why CI fails in the "samerepo" setup...

It's failing for the `--include-subprojects` test case, that's another pitfall that I forgot.

xclaesse:
subprojects could themself have submodule... recursively... not sure that case ever worked tbh.

eli-schwartz:
- free function rename now that it doesn't actually clone anything
- the case of project root not being repo root, is now corrected to actually pass in the temporary distdir for copying (oops) and pass the subdir as `subdir=` instead of `revision=` (also oops)

And with that, `./run_unittests.py test_dist_git` as flagged by the CI as a relevant test to verify this PR... now passes locally.

eli-schwartz:
(I lied, I have local gitconfig which borks things up. Might as well add the changes I needed here too.)

xclaesse:
LGTM, seems to be a safer way indeed.

PR #9213, opened 2021-09-02
Make custom_target() name argument optional

xclaesse:


mesonbuild/interpreter/interpreter.py
```diff
@@ -1620,7 +1620,7 @@ def func_subdir_done(self, node, args, kwargs):
                       'build_always', 'capture', 'depends', 'depend_files', 'depfile',
                       'build_by_default', 'build_always_stale', 'console', 'env',
                       'feed', 'install_tag'})
-    @typed_pos_args('custom_target', str)
+    @typed_pos_args('custom_target', optargs=[str])
     def func_custom_target(self, node: mparser.FunctionNode, args: T.Tuple[str], kwargs: 'TYPE_kwargs') -> build.CustomTarget:
```
dcbaker, line 1624: I'm not sure how I feel about making the name optional, it effectively means we can't ever add any required positional arguments to custom target without breaking backwards compatibility. However, you haven't updated the signature here, it needs to be Optional[str]

xclaesse: We could add other args, just that if you want to pass more than one arg then the name becomes required. You could still do `custom_target('', other_arg)`. A bit ugly but would work in the very unlikely event this ever happen.

xclaesse: > you haven't updated the signature here, it needs to be Optional[str]

done now.

mesonbuild/interpreter/interpreter.py
```diff
@@ -1620,15 +1620,24 @@ def func_subdir_done(self, node, args, kwargs):
...
         name = args[0]
+        if name is None:
+            # name will default to first output, but we cannot do that yet because
+            # they could need substitutions (e.g. @BASENAME@) first. CustomTarget()
+            # will take care of setting a proper default but name must be an empty
+            # string in the meantime.
+            FeatureNew('custom_target() with no name argument', '0.60.0').use(self.subproject)
+            name = ''
+        elif has_path_sep(name):
+            FeatureNew('Path separator in custom_target() name argument', '0.60.0').use(self.subproject)
```
eli-schwartz, line 1640: The only difference between versions is whether or not we print a warning message about it being unsupported.

Since it works fine on old versions of meson, warning aside, I don't believe it merits a FeatureNew.

eli-schwartz (changes requested):
In general I love this idea. One tiny nit.

mesonbuild/interpreter/interpreter.py
xclaesse: Fair enough, removed.

eli-schwartz (approved):

eli-schwartz:
Test failure is relevant:

`C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\IDE\VC\VCTargets\Microsoft.CppBuild.targets(321,5): error MSB3491: Could not write lines to file "bdf6c15@@subdir@file.txt3@cus\subdir/file.txt3.tlog\subdir/file.txt3.lastbuildstate". Could not find a part of the path 'D:\a\1\s\tmph75mip7h\subdir\bdf6c15@@subdir@file.txt3@cus\subdir\file.txt3.tlog\subdir\file.txt3.lastbuildstate'. [D:\a\1\s\tmph75mip7h\subdir\bdf6c15@@subdir@file.txt3@cus.vcxproj]`

xclaesse:
Interesting, ninja backend does not care if the command actually creates the outputs, but not vs backends. Also only ninja backend actually print the target name. Fixed.

eli-schwartz:
Well, in ninja if you don't create the outputs it will just keep trying to rebuild them since they are still out of date.

eli-schwartz:
Test failure looks relevant?

```
builddir = 'D:\\a\\1\\s\\tmpuqf3ozgu', target = 'check_exists'

    def find_vcxproj_with_target(builddir, target):
        import re, fnmatch
        t, ext = os.path.splitext(target)
        if ext:
            p = fr'<TargetName>{t}</TargetName>\s*<TargetExt>\{ext}</TargetExt>'
        else:
            p = fr'<TargetName>{t}</TargetName>'
        for _, _, files in os.walk(builddir):
            for f in fnmatch.filter(files, '*.vcxproj'):
                f = os.path.join(builddir, f)
                with open(f, encoding='utf-8') as o:
                    if re.search(p, o.read(), flags=re.MULTILINE):
                        return f
>       raise RuntimeError(f'No vcxproj matching {p!r} in {builddir!r}')
E       RuntimeError: No vcxproj matching '<TargetName>check_exists</TargetName>' in 'D:\\a\\1\\s\\tmpuqf3ozgu'

run_tests.py:201: RuntimeError
```

xclaesse:
Yes, I've been trying to understand what's going on, seems vs backend actually use the name and does not like having /

keszybz:
Nice feature!

xclaesse:
After more digging, it is indeed a bad idea to have path separators in the custom target name, because it is used in the introspection data, it breaks `meson compile <target>` command.

However, we could still use the first output basename as default name, and we could still prepend the target subdir in the ninja description.

keszybz:
FWIW, the target name doesn't matter too much. It could a randomly-generated string for all I care.

xclaesse:
@eli-schwartz what do you think about this now?

docs/markdown/Reference-manual.md
```diff
@@ -326,8 +326,20 @@ false otherwise.
...
 
 Create a custom top level build target. The only positional argument
-is the name of this target and the keyword arguments are the
-following.
+is the name of this target and cannot contain path separators (`/` or `\`).
+The name of custom target might not be used by every backends, for instance with
+the Ninja backend, `subdir/meson.build` containing the example below,
+`ninja -C builddir foo` or `ninja -C builddir subdir/foo` won't work,
+it is instead `ninja -C builddir subdir/file.txt`. Howerver, `meson compile subdir/foo`
+is accepted.
```
eli-schwartz, line 334: > After more digging, it is indeed a bad idea to have path separators in the custom target name, because it is used in the introspection data, it breaks meson compile <target> command.

I'm *very* surprised that `meson compile` accepts nonsense targets that `ninja` doesn't accept, and I hate it.

Apparently, `meson compile` only *usually* does more or less the same thing as ninja because target names are usually manually coded as the output basename based on superstition. e.g.

```
ninja doc/pacman.8
meson compile doc/pacman.8

ninja scripts/po/tr/LC_MESSAGES/pacman-scripts.mo
```

all work, but 
```
meson compile ninja scripts/po/tr/LC_MESSAGES/pacman-scripts.mo
``` 
reports an error and only using the display name works.

I am now firmly convinced that `meson compile` is completely evil and must be immediately removed.

xclaesse: Each backend has a different way to identify targets, ninja uses exclusively output files path. It makes `meson compile` even more important to have common target names across backends. But in any case, that's offtopic.

As far as this PR is concerned, I think it does make sense to forbid path separator in custom_target() name, because that's consistent with all other targets. Prepending the subdir in ninja description automatically also makes a lot of sense because that's exactly what it does for other targets already: `executable('app', 'foo.c')` in subdir `foo` produce
```
[1/2] Compiling C object foo/app.p/foo.c.o
[2/2] Linking target foo/app
```

xclaesse: @eli-schwartz any objection on merging this now?

eli-schwartz: The PR itself seems fine, I'm just being my usual grumpish self about the existence of mcompile. :)

Offtopic: if we *actually* wanted to support this sort of "common target name", the appropriate solution would have been to add .PHONY targets to the meson.build, but apparently the time for that has passed...

test cases/unit/99 custom target name/subdir/meson.build
```diff
@@ -0,0 +1,9 @@
+# Name argument is optional and should default to 'file.txt', but should be
+# displayed as 'subdir/file.txt' by ninja.
+custom_target(
+  input: '../file.txt.in',
+  output: ['@BASENAME@'],
+  command: [python, '--version'],
+  build_by_default: true,
+  capture: true,
+)
```
eli-schwartz, line 9: trivial nit: missing newline.

Hopefully with the editorconfig improvement in 173a40be330c1aedec9f7f99ebca67a5edd25c3d this sort of thing will never happen again and no one will need to think about the topic again.

For now we might as well fix it. :)

eli-schwartz (approved):
lgtm but see below for style

test cases/unit/99 custom target name/subdir/meson.build
xclaesse: fixed.

Issue #9208, opened 2021-09-02
is_wsl returns false negative when using WSL2


mon:
**Describe the bug**
 #8445 is hitting me again, where Windows EXEs attempt to be wrapped using `mono`. This is caused by `is_wsl` using the check:  
`'microsoft' in platform.version().lower()`

Which fails on WSL2, as `platform.version()` in my case is `#1 SMP Wed Oct 28 23:40:43 UTC 2020`.

**To Reproduce**
```meson
project('test', 'c')

run_target('example',
    command : [find_program('/mnt/c/WINDOWS/system32/help.exe')]
)
```

**Expected behavior**
help.exe's default output is printed as meson runs the command, but instead it fails to find `mono` (which I do not have installed)

**system parameters**
Meson: 0.59.1
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.3 LTS
Release:        20.04
Codename:       focal

xclaesse:
Any suggestion how we can detect wsl2 then?

mon:
Good question, it seems we're in luck!

I just checked, on another machine, running `platform.uname()`:

WSL 1:  
`uname_result(system='Linux', node='smol', release='4.4.0-19041-Microsoft', version='#1151-Microsoft Thu Jul 22 21:05:00 PST 2021', machine='x86_64', processor='x86_64')`

WSL 2:  
`uname_result(system='Linux', node='smol', release='4.19.104-microsoft-standard', version='#1 SMP Wed Feb 19 06:37:35 UTC 2020', machine='x86_64', processor='x86_64')`

So it seems that `'microsoft' in platform.release().lower()` will work for both WSL 1 and WSL2, and *hopefully* into the future. Should I make a PR, or would you be faster?

xclaesse:
They even changed to remove the capital letter, to be sure they break as many script as possible, well done.

Here they suggest reading `/proc/version`: https://github.com/microsoft/WSL/issues/4071, they even have a meson.build example :D

> So it seems that `'microsoft' in platform.release().lower()` will work for both WSL 1 and WSL2, and _hopefully_ into the future. Should I make a PR, or would you be faster?

Yes, I think that should be fine until they change it again... I haven't found a formal documented way of doing it, so we can only hope. PR welcome :)

eli-schwartz:
Fixed by linked PR.

PR #9214, opened 2021-09-03
Improve WSL detection

mon:
WSL 2 removes the "Microsoft" from `platform.version` but leaves it inside `platform.release`. This lets us detect both types of WSL without issue.

Followup from #9208


xclaesse:
> Followup from #9208

pro tip: If you write `Fixes: #9208` in commit msg, github closes the issue automatically when merged.

xclaesse:
LGTM when CI is green.

Issue #6314, opened 2019-12-08
Visual Studio 2019 backend doesn't pick up c++ standard version


DaOnlyOwner:
Bug description:
I use ```project('Hyra','cpp', default_options : ['cpp_std=c++17'], version : '0.01')``` in my meson build file to specify that c++17 should be used. When configuring the project (and also regenerating it) however, VS doesn't seem to care about this setting, it doesn't get applied, as you can see here (the last entry):
![image](https://user-images.githubusercontent.com/11295581/70388549-d6b13700-19b3-11ea-8126-a430cbf19435.png)
I have to manually change the standard version in the UI in order to use c++ 2017 (e.g. std::variant).

System Parameters
* Native build
* Win 10
* Python version: 3.7.0
* Meson version: 0.52.0

scivision:
The /std:c++17 arg is passed into vs2010backend.py:gen_vcxproj but I didn't trace it farther. Is there a anyone familiar with Meson's VS backend to know if setting cpp_std used to work for MSVC? It looks like it's intended to work but maybe there's a glitch somewhere in vs2010backend.py?

DaOnlyOwner:
That might be the case.
I found out that when I manually set the language standard, the option appears in the corresponding vcxproj file: 
![image](https://user-images.githubusercontent.com/11295581/70430150-6e795880-1a7a-11ea-9a75-93f754650472.png)

A newly setup VS2019 project doesn't have that option.
Indeed vs2010backend.py doesn't specify "<LanguageStandard>...</LanguageStandard>" anywhere. I think that's the problem. 
So I created a method which defines "<LanguageStandard>...</LanguageStandard>" as if I had specified the option manually in the UI.
```python
    def set_cpp_std_version(self, clconf, target):
        std_standard = self.environment.coredata.compiler_options[target.for_machine]['cpp_std'].printable_value()
        std_standard_xml = ET.SubElement(clconf,"LanguageStandard")
        std_standard_xml.text = "std" + std_standard.replace("+","p")
        std_standard_xml.attrib["Condition"] = "'$(Configuration)|$(Platform)'=='" + self.buildtype + "|" + self.platform + "'" 
        # END
```
It should work as a temporary workaround I guess.

scivision:
here's a minimal example meson.build

```meson
project('demo','cpp', default_options: ['cpp_std=c++17'])

executable('main', 'main.cpp')
```
[...]

PR #9216, opened 2021-09-03
Encode Language standard in VS XML. Fixes: #6314

qmor:


qmor:
> Please amend this PR to use a commit message of the form:
> 
> ```
> do foo when bar in order to baz
> 
> Fixes #bugid
> ```
> 
> The description of the commit needs to tell the reader what it is trying to accomplish without just saying "look at this bug report", and bug report links should be used in the postscript.
> 
> Also the exact syntax "Fixes "#bugid" will let github generate a close action when the PR is merged.

done

mesonbuild/backend/vs2010backend.py
```diff
@@ -1099,6 +1099,11 @@ def gen_vcxproj(self, target, ofname, guid):
         else:
             ET.SubElement(clconf, 'FavorSizeOrSpeed').text = 'Speed'
         # Note: SuppressStartupBanner is /NOLOGO and is 'true' by default
+        if self.name in('vs2017', 'vs2019'):
```
dcbaker, line 1102: `in` is a keyword, there should be a space after `in`.

mesonbuild/backend/vs2010backend.py
```diff
@@ -1099,6 +1099,11 @@ def gen_vcxproj(self, target, ofname, guid):
...
+            if 'cpp' in file_args:
+                optargs = [x for x in file_args['cpp'] if x.startswith('/std:c++')]
+                if len(optargs)>0:
```
dcbaker, line 1105: using `len()` to check container emptiness is a python anti-pattern, as it o(N):
```python
if optargs:
    ...
```

dcbaker (changes requested):
I've got a couple of style nits, but also a question. Why is this only beind done for C++ and not for C as well?

qmor:
> I've got a couple of style nits, but also a question. Why is this only beind done for C++ and not for C as well?

unfortunatelly i have no idea how to take right c_std option passed from meson.build file. 
i've wrote in my meson.build:

           project('test_project', ['c'], default_options : ['c_std=c11'])

and then in vs2010backend.py:

            if 'c' in file_args:
                opts = self.get_base_options_for_target(target)
                if OptionKey('std',lang='c') in opts.options:
                    optcstd = opts.options[OptionKey('std',lang='c')].value
                    if optcstd in ('c11','c17'):
                        ET.SubElement(clconf, 'LanguageStandard_C').text = 'std'+optcstd

but in some reason optcstd is always 'c99'

dcbaker:
IIRC you need a fairly recent verison of MSVC to get c11 support, 2019 I think.

dcbaker:
I can poke at this more tomorrow if you need some help with the C part

qmor:
> I can poke at this more tomorrow if you need some help with the C part

Looks like it works after clean configure. Has pushed the same changes as for c++

mesonbuild/backend/vs2010backend.py
qmor: done

mesonbuild/backend/vs2010backend.py
qmor: done

qmor:
@dcbaker, looks like it's works now with c and c++

jpakkane:
This branch has many commits that are merges from master. These are not permitted and need to be removed. For details see:

https://mesonbuild.com/Contributing.html#keeping-pull-requests-up-to-date

qmor:
Will open new pull request. More clear

PR #9217, opened 2021-09-03
ci: Use GitHub concurrency

mensinda:
This should significantly speed up the CI by canceling all previous runs of a PR on a force push. I tested that this works in https://github.com/mensinda/meson/pull/5.


.github/workflows/website.yml
```diff
@@ -1,5 +1,8 @@
 name: Update website
 
+concurrency:
+  group: mesonbuild.com
```
xclaesse, line 4: I'm a bit worried what happens if it gets stopped during the upload of the updated website? I think the upload it a `git push`, is that atomic operation?

xclaesse:
It is the concurrency per PR?

.github/workflows/website.yml
mensinda: It should *not* cancel anything since the flag is not to set to `true` here. The way it is currently set up is that only one website job can run at a time. All other jobs are paused/pending until the previous one is done.

See https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency

mensinda:
> It is the concurrency per PR?

Yes. PRs don't interfere with each other (at least in my fork).

.github/workflows/website.yml
xclaesse: Oh, I see, you don't have `cancel-in-progress` here, so that's going to queue every website updates. That's actually super important fix, nice catch!

xclaesse:
To be honnest, I've always assumed this behaviour of cancelling previous pipeline when pushing into a PR was the default. I'm shocked to learn it was not.... Definitively +1 LGTM.

mensinda:
Let's hope that the CI passes this time. I didn't even realize that our image builders were broken...

.github/workflows/website.yml
mensinda: I have also added a comment so it doesn't get added by accident.

mensinda:
Azure seems to also [support something like this](https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema%2Cparameter-schema#pr-trigger), and it is enabled by default, but apparently broken...

dcbaker (approved):

PR #9218, opened 2021-09-03
Allow as_link_whole() on external dependencies

### LINKED ISSUE #5797: Custom target linking with `link_whole` in a shared library is broken
On using a `custom_target` to create a static library, then trying to link that static library with a shared library using `link_whole`, the meson process quits with a stack trace ending with:
```
  File "/usr/lib/python3.7/site-packages/mesonbuild/build.py", line 1081, in link_whole
    if isinstance(self, SharedLibrary) and not t.pic:
AttributeError: 'CustomTarget' object has no attribute 'pic'
```

xclaesse:
When a static library link whole on an external static library we need
to extract the archive and use object files it contains instead.

Fixes: #5797

### DIFF
--- a/mesonbuild/dependencies/base.py
+++ b/mesonbuild/dependencies/base.py
@@ -14,6 +14,7 @@ import itertools
 import typing as T
 import uuid
 from enum import Enum
+from dataclasses import dataclass
 
 from .. import mlog, mesonlib
 from ..compilers import clib_langs
@@ -129,6 +130,12 @@ class DependencyMethods(Enum):
 
 
 DependencyTypeName = T.NewType('DependencyTypeName', str)
+
+
+@dataclass
+class ExtractArchives:
+    archive_filename: str
+    objects: T.List[str]
 
 
 class Dependency(HoldableObject):
@@ -303,6 +310,26 @@ class Dependency(HoldableObject):
         """Used as base case for internal_dependency"""
         return self
 
+    def generate_link_whole_dependency(self) -> T.Tuple[Dependency, T.List[ExtractArchives]]:
+        ''' Return a copy of this dependency suited for link-whole
+        Removes all external static libraries (*.a, *.lib) this dependency links
+        with, and return them as a list of ExtractArchives.
+        '''
+        from ..utils import ar
+
+        new_dep = copy.copy(self)
+        new_dep.link_args = []
+        archives: T.List[ExtractArchives] = []
+        for arg in self.link_args:
+            if arg.endswith('.a'):
+                objects = ar.extract(arg, dry_run=True)
+                if objects:
+                    t = ExtractArchives(arg, objects)
+                    archives.append(t)
+                    continue
+            new_dep.link_args.append(arg)
+        return new_dep, archives
+
 class InternalDependency(Dependency):
 
     type_name = DependencyTypeName('internal')
@@ -389,22 +416,21 @@ class InternalDependency(Dependency):
             return val
         raise DependencyException(f'Could not get an internal variable and no default provided for {self!r}')
 
-    def generate_link_whole_dependency(self) -> Dependency:
+    def generate_link_whole_dependency(self) -> T.Tuple[Dependency, T.List[ExtractArchives]]:
         from ..build import SharedLibrary, CustomTarget, CustomTargetIndex
-        new_dep = copy.deepcopy(self)
-        for x in new_dep.libraries:
+        new_dep, archives = super().generate_link_whole_dependency()
+        assert isinstance(new_dep, InternalDependency) # For mypy
+        new_dep.libraries = []
+        new_dep.whole_libraries = self.whole_libraries.copy()
+        for x in self.libraries:
             if isinstance(x, SharedLibrary):
                 raise MesonException('Cannot convert a dependency to link_whole when it contains a '
                                      'SharedLibrary')
             elif isinstance(x, (CustomTarget, CustomTargetIndex)) and x.links_dynamically():
                 raise MesonException('Cannot convert a dependency to link_whole when it contains a '
                                      'CustomTarget or CustomTargetIndex which is a shared library')
-
-        # Mypy doesn't understand that the above is a TypeGuard
-        new_dep.whole_libraries += T.cast('T.List[T.Union[StaticLibrary, CustomTarget, CustomTargetIndex]]',
-                                          new_dep.libraries)
-        new_dep.libraries = []
-        return new_dep
+            new_dep.whole_libraries.append(x)
+        return new_dep, archives
 
     def get_as_static(self, recursive: bool) -> InternalDependency:
         new_dep = copy.copy(self)
--- a/mesonbuild/dependencies/base.py
+++ b/mesonbuild/dependencies/base.py
@@ -29,7 +29,8 @@ if T.TYPE_CHECKING:
     from ..interpreterbase import FeatureCheckBase
     from ..build import (
         CustomTarget, IncludeDirs, CustomTargetIndex, LibTypes,
-        StaticLibrary, StructuredSources, ExtractedObjects, GeneratedTypes
+        StaticLibrary, StructuredSources, ExtractedObjects, GeneratedTypes,
+        GeneratedList
     )
     from ..interpreter.type_checking import PkgConfigDefineType
 
@@ -155,6 +156,7 @@ class Dependency(HoldableObject):
         self.featurechecks: T.List['FeatureCheckBase'] = []
         self.feature_since: T.Optional[T.Tuple[str, str]] = None
         self.meson_variables: T.List[str] = []
+        self.objects: T.List[T.Union[str, mesonlib.File, ExtractedObjects, CustomTarget, CustomTargetIndex, GeneratedList]] = []
 
     def __eq__(self, other: object) -> bool:
         if not isinstance(other, Dependency):
@@ -327,7 +329,7 @@ class InternalDependency(Dependency):
         self.extra_files = list(extra_files)
         self.ext_deps = ext_deps
         self.variables = variables
-        self.objects = objects
+        self.objects.extend(objects)
         if d_module_versions:
             self.d_features['versions'] = d_module_versions
         if d_import_dirs:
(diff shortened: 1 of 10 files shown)


xclaesse:
@mikezackles Ok, it's starting to get useful result now:
```meson
glib_dep = dependency('glib-2.0', static: true)
static_library('foo', 'foo.c', dependencies: glib_dep.as_link_whole())
```
This creates libfoo.a that contains glib from system. Let me know if it works for you, especially on OSX.

xclaesse:
@nirbheek You could be interested by this too, I remember you suggested this kind of solution a loooooong time ago.

mikezackles:
I'm getting `ValueError: embedded null byte` from `with output.open('wb') as ofile` on OS X. Maybe a null byte in the filename? Might be able to dig some more later.

xclaesse:
Can you attach the .a file to help debug that? Probably will have to add it into a unit test anyway.

mikezackles:
Sorry I was trying to do just that, but this OS X machine is graying out the file for upload for some reason.

mikezackles:
Seems it's a github thing? I emailed it to you.

dcbaker:
This seems like a lot of work to avoid just creating an extracted objects object from a bunch of .o files, or am i missing something?

xclaesse:
> This seems like a lot of work to avoid just creating an extracted objects object from a bunch of .o files, or am i missing something?

This is for the case where the staticb library comes from the system, you don't have .o files until you manually extract the .a, unless there is something I'm missing?

xclaesse:
On second thought, I think it's better to extract at configure time. The cons is it means reconfiguring when system deps change which should be pretty rare. But doing it at build time means using a Python wrapper for every link which has significant performance impact for each rebuild. Also from an implementation pov it is easier to extract at configure time because it is backend agnostic. I think keeping implementation self contained is important too for a niche use case, we cannot justify the maintenance burden otherwise.

xclaesse:
Also extracting archives will explode the number of args passed to the linker, that will easily exceed the limit on Windows. Doing it at configure time means the backend generator can switch to using a rsp file AFAIK.

jpakkane:
As an alternative thing, maybe we could consider adding a top level function like `merge_static_libs('newname', lib1, lib2)`. No other build system does static library merging so maybe it would be better to be explicit about it?

xclaesse:
@jpakkane Hmm, you could also want to link_whole an external library into a shared_library(), that's also a case we don't currently support, do we? But in that case I don't think we need to extract the library, the compiler can do it for us, right? I think building on top of existing `as_link_whole()` API would be more elegant, but your solution is easier and more self contained indeed.

@mikezackles I updated the PR to fix the OSX archive case, included the .a you sent me in the unit test.

mikezackles:
> you could also want to link_whole an external library into a shared_library(), that's also a case we don't currently support, do we?

I don't know if it's intentional, but I think this does work currently even without using `as_link_whole` as I am able to ship shared libraries that link against my deps statically. (I haven't looked into why.) Otherwise this would be a much bigger problem for me.

> @mikezackles I updated the PR to fix the OSX archive case, included the .a you sent me in the unit test.

Awesome, thank you!

xclaesse:
> I am able to ship shared libraries that link against my deps statically

I _think_ in that case the shared library does not expose the full ABI of its static dependencies. The linker only picks what it needs from static deps and drop the rest.

mikezackles:
> does not expose the full ABI

Ah I see, that would make sense. I see the distinction now. (The dynamic libs I ship do not attempt to do this.)

nirbheek:
> On second thought, I think it's better to extract at configure time. The cons is it means reconfiguring when system deps change which should be pretty rare. But doing it at build time means using a Python wrapper for every link which has significant performance impact for each rebuild. Also from an implementation pov it is easier to extract at configure time because it is backend agnostic. I think keeping implementation self contained is important too for a niche use case, we cannot justify the maintenance burden otherwise.

Instead of doing it at configure time, why not make a _new_ target that does the job? It will only run once per build.

xclaesse:
> Instead of doing it at configure time, why not make a _new_ target that does the job? It will only run once per build.

If I don't extract the archive at configure time then I don't know the list of object files to pass to `ar` command when generating the link target. The way I used in a previous version of this patch is to wrap ar command completely with a meson script (see old commit https://github.com/mesonbuild/meson/commit/0a186abbadc6189188d7ec869969534fe504e994).

nirbheek:
> If I don't extract the archive at configure time then I don't know the list of object files to pass to `ar` command when generating the link target

Why not list the contents of the static archive with `ar t path/to/libfoo.a`? In my testing that's 10x faster than extracting the static library.

xclaesse:
> Why not list the contents of the static archive with `ar t path/to/libfoo.a` and `lib /list drive:\path\to\foo.lib`?

Does it deal with duplicate names? Even if we only list the content at configure it still means we need to reconfigure when the lib changes, so that does not make much difference to just extract directly at configure time. Since we need an ar parser anyway to extract, better use the same code to list the content too, pretty sure if I add an option to not extract it's going to be faster than using extrnal command for that.

nirbheek:
> Does it deal with duplicate names? Even if we only list the content at configure it still means we need to reconfigure when the lib changes, so that does not make much difference to just extract directly at configure time.

A 100MB static library takes 200ms to extract for me and 20ms to list contents; that's going to add up very quickly I think. Semantically as well, I think it's wrong to do I/O-heavy tasks in `setup`.

> Since we need an ar parser anyway to extract, better use the same code to list the content too, pretty sure if I add an option to not extract it's going to be faster than using extrnal command for that.

I agree that your parser is probably the best option for listing files in the archive, and not just extracting. Cross-platform support will probably be easier.

xclaesse:
> I think it's wrong to do I/O-heavy tasks in `setup`.

Ok, that's actually a good point, you're right.

nirbheek:
Just for completeness (I don't think this is worth spending time on), I'll mention that we can also eliminate reconfigure when the static library changes by doing the same thing we do for avoiding relinks to shared libraries: have another target that checks whether the list of objects inside the static library is the same, and if it is, then we don't need to reconfigure.

xclaesse:
@nirbheek Updated my ar.py module to support listing without extracting. I'm curious how it performs compared to `ar t`.

xclaesse:
@nirbheek I revisited this PR, it now does as you suggested: list object files contained in a static lib at configure but only extract them at build time with a custom target.

mesonbuild/dependencies/base.py
```diff
@@ -248,6 +258,20 @@ def generate_system_dependency(self, include_type: str) -> 'Dependency':
         new_dep.include_type = self._process_include_type_kw({'include_type': include_type})
         return new_dep
 
+    def generate_link_whole_dependency(self) -> T.Tuple[Dependency, T.List[ExtractArchives]]:
+        new_dep = copy.copy(self)
+        new_dep.link_args = []
+        archives: T.List[ExtractArchives] = []
+        for arg in self.link_args:
+            if any(arg.endswith(i) for i in {'.a', '.lib'}):
```
nirbheek: Import libraries can also be extracted as static libraries, but we probably don't want to use them like that, right? What do we do about that? One option is if the symbols in the library start with `__imp_` we should treat it as an import library. Or we could skip `.lib` entirely for now.

mesonbuild/interpreter/interpreterobjects.py
```diff
@@ -555,8 +555,25 @@ def as_system_method(self, args: T.Tuple[T.Optional[str]], kwargs: TYPE_kwargs)
     @noPosargs
     def as_link_whole_method(self, args: T.List[TYPE_var], kwargs: TYPE_kwargs) -> Dependency:
         if not isinstance(self.held_object, InternalDependency):
-            raise InterpreterException('as_link_whole method is only supported on declare_dependency() objects')
-        new_dep = self.held_object.generate_link_whole_dependency()
+            FeatureNew('as_link_whole on external dependency', '1.1.0').use(self.subproject)
```
nirbheek: FeatureNew needs updating (just leaving it as a comment here to check before it's merged)

stsp:
Hi!

Just wondering, would it be feasible to
submit `Allow as_link_whole() on external dependencies`
separately from an archive extractor?
For my problem that would be enough,
and maybe it can come easier than the
whole thing together?

eli-schwartz:
I think the problem is the other way around. We need the archive extractor *first*, because implementing as_link_whole() depends on having been able to extract the archive.

as_link_whole() is based on using the individual `*.o` files.

That being said, it would be great to land this PR...

stsp:
Then maybe I don't need `as_link_whole()`?
I definitely don't need an archive extractor.
All I need, is this:
```
libdjs = dependency('dj64static', static: true)
elf = custom_target(TARGET + '.elf',
  input: [ some_inputs, libdjs],
  command: [ just call linker -static here, and it deals with static libs on its own],
...
```

So what do I need for this?
Since I can call the linker directly, archive
extractor is not what I really need.
Maybe it would be better to answer in
#12709 if this appears unrelated to this PR.

nsaxena16:
Hi,

I also came across this functionlity where external dependency needs to be added with "as_link_shole()". What is the recommened method for doing it?

nirbheek (approved):
FIVE YEARS LATER. I am finally submitting these pending review comments. I thought I had done it already...

bonzini:
Missing release notes

mesonbuild/dependencies/base.py
xclaesse: Ok, only including *.a for now. Static lib naming on windows is a hot debate...

mesonbuild/interpreter/interpreterobjects.py
xclaesse: done

xclaesse:
> Missing release notes

snippet and doc done

nirbheek:
Tests are failing spectacularly

bonzini:
PlatformAgnosticTests.test_setup_loaded_modules due to the new mesonbuild.utils.ar module.

xclaesse:
> PlatformAgnosticTests.test_setup_loaded_modules due to the new mesonbuild.utils.ar module.

Fixed.

xclaesse:
Remaining CI failures are because some platforms does not have static glib installed:

> WARNING: Static library 'glib-2.0' not found for dependency 'glib-2.0', may not be statically linked

I also notice it makes as_link_whole recursive. Basically any -lfoo from pkgconfig that meson resolved to a `.a` will be extracted. Not sure that's desired or how to solve it.

nirbheek:
> I also notice it makes as_link_whole recursive. Basically any -lfoo from pkgconfig that meson resolved to a `.a` will be extracted. Not sure that's desired or how to solve it.

I think that is desirable. The external dependency is represented by all the static libs that it depends on.

Issue #9197, opened 2021-08-31
CMake detecting dependency assumes PACKAGE_LIBRARIES are libraries for one configuration


gerard-ryan-immersaview:
**Describe the bug**
[/mesonbuild/dependencies/cmake.py:451](https://github.com/mesonbuild/meson/blob/0.59.1/mesonbuild/dependencies/cmake.py#L451) assumes that PACKAGE_LIBRARIES are for only one configuration, but [FindOpenSSL.cmake](https://github.com/Kitware/CMake/blob/master/Modules/FindOpenSSL.cmake#L359-L360) for example uses [`select_library_configurations()`](https://cmake.org/cmake/help/latest/module/SelectLibraryConfigurations.html) which creates a list with both debug and release libraries that requires further parsing.

**To Reproduce**
```meson.build
project('hello-world', 'c')
openssl_dep = dependency('openssl')
executable('hello-world', 'main.c', dependencies: openssl_dep)
```
```cmd
meson setup --buildtype=debug build
meson.py compile -C build
```

**Expected behaviour**
The linker will attempt to link with only the debug OpenSSL libraries. 
AKA not both with erroneous keywords like "optimized".
```
"optimized" "C:/Program Files/OpenSSL-Win64/lib/VC/libssl64MD.lib" "debug" "C:/Program Files/OpenSSL-Win64/lib/VC/libssl64MDd.lib" "optimized" "C:/Program Files/OpenSSL-Win64/lib/VC/libcrypto64MD.lib" "debug" "C:/Program Files/OpenSSL-Win64/lib/VC/libcrypto64MDd.lib"
```

**system parameters**
* Is this a [cross build](https://mesonbuild.com/Cross-compilation.html) or just a plain native build (for the same computer)? **plain native**
* what operating system (e.g. MacOS Catalina, Windows 10, CentOS 8.0, Ubuntu 18.04, etc.) **Windows 10**
* what Python version are you using e.g. 3.8.0 **3.9.2**
* what `meson --version` **0.59.1**
* what `ninja --version` if it's a Ninja build **1.10.0.git.kitware.jobserver-1**
* what `cmake --version` **3.20.21032501-MSVC_2**
* what `cl` version **Microsoft (R) C/C++ Optimizing Compiler Version 19.29.30133 for x64**

This seems to be the cause of https://gitlab.freedesktop.org/libnice/libnice/-/issues/140

mensinda:
[...]

PR #9221, opened 2021-09-04
CMake: Implement support for keywords in *_LIBRARIES variables.

mensinda:
CMakes `target_link_libraries()` supports certain keywords to only enable specific libraries for specific CMake configurations. We now try our best to replicate this for Meson dependencies.

Additionally, a warning is printed if the old fallback variables are used instead of an imported target.

This is not a regression but a bug fix. However, it should still be good for 0.59.2. Feel free to remove the milestone if there are any concerns.

Fixes:
- https://github.com/mesonbuild/meson/issues/9197
- https://gitlab.freedesktop.org/libnice/libnice/-/issues/140


mensinda:
Does anyone have any objections to merging this?

mensinda:
Merging then

PR #9223, opened 2021-09-05
Add relative path support for native/crossfiles

### LINKED ISSUE #3439: cross file cannot find binaries with a relative path
When building numerous embedded projects it is necessary to have the toolchain binaries part of the project so that multiple vendor version can be installed without collisions.  Being able to specify the binaries relative to the cross file allows portable toolchains.

miebka:
As indicated in #3439 some project setups have the toolchain included within the repository. To reduce the initial starting effort when a developer clones/checkouts the repository this PR adds support for binaries with a relative path within native/crossfile.

Fixes #3439

### DIFF
--- a/docs/markdown/snippets/relative_path_within_native_and_crossfiles.md
+++ b/docs/markdown/snippets/relative_path_within_native_and_crossfiles.md
@@ -1,0 +1,5 @@
+## Relative paths are supported within native and cross files
+
+Support for relative paths within native and cross files has been added. 
+This reduces the initial effort when projects include the toolchain 
+within the repository.
--- a/docs/markdown/snippets/relative_path_within_native_and_crossfiles.md
+++ b/docs/markdown/snippets/relative_path_within_native_and_crossfiles.md
@@ -1,5 +1,11 @@
 ## Relative paths are supported within native and cross files
 
-Support for relative paths within native and cross files has been added. 
-This reduces the initial effort when projects include the toolchain 
-within the repository.
+The binaries specified in the machine file can now be defined
+by using their relative path in relation to the machine file
+If the following machine file is located under 
+`/my_project/toolchain` the entry for `c` will be extended 
+to `/my_project/toolchain/bin/gcc`.
+
+```ini
+c = 'bin/gcc'
+```
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -24,10 +24,11 @@ from .mesonlib import (
     HoldableObject,
     MesonException, EnvironmentException, MachineChoice, PerMachine,
     PerMachineDefaultable, default_libdir, default_libexecdir,
+
     default_prefix, default_datadir, default_includedir, default_infodir,
     default_localedir, default_mandir, default_sbindir, default_sysconfdir,
     split_args, OptionKey, OptionType, stringlistify,
-    pickle_load, replace_if_different
+    pickle_load, replace_if_different, listify
 )
 from .wrap import WrapMode
 import ast
@@ -932,18 +933,20 @@ class MachineFileParser():
         self.constants = {'True': True, 'False': False}
         self.sections = {}
 
-        self.parser.read(filenames)
+        for fname in filenames:
+            self.parser.read(fname)
 
-        # Parse [constants] first so they can be used in other sections
-        if self.parser.has_section('constants'):
-            self.constants.update(self._parse_section('constants'))
+            # Parse [constants] first so they can be used in other sections
+            if self.parser.has_section('constants'):
+                self.constants.update(self._parse_section('constants', fname))
 
-        for s in self.parser.sections():
-            if s == 'constants':
-                continue
-            self.sections[s] = self._parse_section(s)
+            for s in self.parser.sections():
+                if s == 'constants':
+                    continue
+                self.sections[s] = self._parse_section(s, fname)
+            
 
-    def _parse_section(self, s):
+    def _parse_section(self, s, fname):
         self.scope = self.constants.copy()
         section = {}
         for entry, value in self.parser.items(s):
@@ -958,6 +961,15 @@ class MachineFileParser():
                 raise EnvironmentException(f'Malformed value in machine file variable {entry!r}.')
             except KeyError as e:
                 raise EnvironmentException(f'Undefined constant {e.args[0]!r} in machine file variable {entry!r}.')
+                        
+            if s == 'binaries':
+                if not isinstance(res, (list, str)):
+                    raise MesonException(
+                        f'Invalid type {res!r} for entry {entry!r} in machine file')
+                res = listify(res)
+                if self._is_binary_with_relative_path(res):
+                    res[0] = os.path.join(os.path.dirname(fname), res[0])            
+            
             section[entry] = res
             self.scope[entry] = res
         return section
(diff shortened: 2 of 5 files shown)


mesonbuild/envconfig.py
```diff
@@ -375,8 +376,25 @@ def __init__(
...
                         f'Invalid type {command!r} for entry {name!r} in cross file')
+                command = self._rel2abs_path(command)
                 self.binaries[name] = mesonlib.listify(command)
 
+    def _rel2abs_path(self, command: T.Union[str, T.List[str]]) -> T.Union[str, T.List[str]]:
+        if command:
+            if isinstance(command, list):
+                binary = command[0]
+            else:
+                binary = command
```
xclaesse, line 387: Call `mesonlib.listify` first so you know it's always a list, we listify anyway just after this call.

docs/markdown/snippets/relative_path_within_native_and_crossfiles.md
```diff
@@ -0,0 +1,5 @@
+## Relative paths are supported within native and cross files
+
+Support for relative paths within native and cross files has been added. 
+This reduces the initial effort when projects include the toolchain 
+within the repository.
```
xclaesse, line 5: It should specify relative to what? Users might assume it's relative to the location of the file instead of relative to top source dir.

xclaesse: Also should specify that it's not assumed to be relative in the case it's just a filename with no path. I guess we could suggest `./foo` in that case.

xclaesse: Should be documented in `Machine-files.md` too, snippets/ is for release notes only. You can just copy/paste the text.

xclaesse (commented):
Would need a unit test too.

mesonbuild/envconfig.py
miebka: Thanks, first calling `mesonlib.listify` makes this a lot easier.

docs/markdown/snippets/relative_path_within_native_and_crossfiles.md
miebka: I'll work your comments in.

miebka:
Ok, the `relative to what` question showed me that my first implementation was not quite what the intent of #3439 was.
At least for embedded projects it is common to have the toolchain within the project repository or have a seperate toolchain repo.
With the following project structure and the `crossfile.ini` within the toolchain repository the entry `c` of the binary should
be expanded to `/projects/project_A/toolchain/bin/gcc`.

```bash
projects/
|
+- project_A/
|  |
|  +- toolchain/
|  |  |─ bin/
|  |  |  +- gcc
|  |  |─ include/
|  |  |─ lib/
|  |  |─ crossfile.ini
|  |
|  +- source/
|  |  |─ (...)
|  +- meson.build
```

```ini
[binaries]
c = 'bin/gcc'
```

With this PR the toolchain folder including the machinefile could be placed at any place. As long as the binary entries are referenced relative to the machinefile, they will be found.

mesonbuild/coredata.py
```diff
@@ -876,18 +877,20 @@ def __init__(self, filenames: T.List[str]) -> None:
         self.constants = {'True': True, 'False': False}
         self.sections = {}
 
-        self.parser.read(filenames)
+        for fname in filenames:
+            self.parser.read(fname)
```
xclaesse, line 881: I don't think you should change this, configparser has special treatment when reading multiple files.

xclaesse: Ah, I understand now why you did that, you want paths to be relative to the machine file, and not relative to the source dir... It actually make sense... But this changes slightly the behaviour of the parser I think, because now you're evaluating constants for each file instead of after merging all files together. It is possible for example to have one machine file with only `[constants]` and another file that uses those constants, the idea being you can have common parts in one file and only switch the constants file for each arch.

I'm not sure how to make this work though....

mesonbuild/coredata.py
```diff
@@ -902,6 +905,19 @@ def _parse_section(self, s):
...
             except KeyError as e:
                 raise EnvironmentException(f'Undefined constant {e.args[0]!r} in machine file variable {entry!r}.')
+
+            if s == 'binaries':
+                if isinstance(res, list):
+                    for element in res:
+                        if not isinstance(element, str):
+                            raise EnvironmentException(f'Invalid type {res!r} for entry {entry!r} in machine file')        
+                elif not isinstance(res, str):
+                    raise EnvironmentException(f'Invalid type {res!r} for entry {entry!r} in machine file')
```
xclaesse, line 915: You can use `stringlistify()` that also checks all items are strings.

mesonbuild/coredata.py
miebka: First I agreed with you, that the handling of constants has changed. So I added a small unittest containing three machinfiles where the constant `arch` is overwritten in the last crossfile.
```python
def test_cross_file_constants_merge(self):
        with temp_filename() as crossfile1, temp_filename() as crossfile2, temp_filename() as crossfile3:
            with open(crossfile1, 'w', encoding='utf-8') as f:
                f.write(textwrap.dedent(
                    '''
                    [constants]
                    arch = 'win64'
                    '''))
            with open(crossfile2, 'w', encoding='utf-8') as f:
                f.write(textwrap.dedent(
                    '''
                    [binaries]
                    c = arch + '-gcc'
                    '''))
            with open(crossfile3, 'w', encoding='utf-8') as f:
                f.write(textwrap.dedent(
                    '''
                    [constants]
                    arch = 'arm64'

                    [binaries]
                    g++ = arch + '-g++'
                    '''))
            config = mesonbuild.coredata.parse_machine_files([crossfile1, crossfile2, crossfile3])
            bt = mesonbuild.envconfig.BinaryTable(config.get('binaries', {}))
            self.assertEqual(bt.binaries['c'], ['arm64-gcc'])
            self.assertEqual(bt.binaries['g++'], ['arm64-g++'])
```
My expectation was that with my change the binary `c` equals to `win64-gcc` instead of `arm64-gcc`, unfortunately the unittest passed... 

So after digging into the code it was clear: Every time the configparser is called by `self.parser.read(fname)` the new machinefile is merged into the already parsed data, so at the end the behaviour is identical.

BUT this also means that if there are e.g. 2 machinefiles with binary entries with a relative path, the evaluation of the last machinefile overwrites the path from the first one.

=> still needs some rework

mesonbuild/coredata.py
miebka: `stringlistify()` would result in a more general exception message, while the code above points the developer to his error.
With regard to clean code, `stringlistify()` should definitely be preferred.

miebka:
IMHO the current implementation should now respect all use-cases of using machinefiles. To pre-evaluate the constants, first the list with all machinefiles is passed to the configparser. Then each machinefile is evaluated individually and a binary entry with a relative path (related to the machinfile) is converted to one with an absolute path.

xclaesse:
Thinking about this again, instead of allowing relative paths, maybe we should just add predefined constants with the path to sourcedir/builddir. This is similar to pkg-config files that have `${pcfiledir}` for example.

e.g.
```ini
[binaries]
c = meson_source_dir / 'toolchain' / 'gcc'
```

That also solves having to use ugly `./gcc` prefix to mean relative path instead of search program in system PATH.

xclaesse:
Implementation is also simpler, and it's easier to extend later with other predefined constants: https://github.com/mesonbuild/meson/pull/12083.

miebka:
Became redundant due to x https://github.com/mesonbuild/meson/pull/12083 :+1:

Issue #6526, opened 2020-01-27
Design for generator improvements


jpakkane:
This is basically a fleshed out #3342. Generators seem to not fulfill requirements that people have for them, so let's see if we can make them better. One possible way of making them better is that if you do this:

```meson
g = generator(...)
l = g.process(..., unique_id: 'somename')
```

It would output the files under `${builddir}/meson-gen/somename` (and subprojects under the respective subproject dir) instead of the target's private directory. Then when you use the output in multiple targets, they all use the same generated files.

Why like this? Mainly for backwards compatibility. People have projects that use the target private dir for header lookups and suddenly generating the headers somewhere else would break things. The user must specify a unique id for each such generator (using the same name multiple times would be a hard error). The reason for this is that I could not come up with a way to generate a reliable set of state to hash to create a unique id automatically.

germandiagogomez:
FWIW I myself started using `custom_target` in what should be a `generator` in my opinion.

1. It looks weird to me that a generator generates the same files again and again. I ended up using `custom_target` but what I want is a compiler for idl files (capnproto in my case).

2. Since this is related to generating code improvements, I also had a related issue when generating the file output (but this is in a `custom_target`, not sure it affects to `generator`). The issue is here:  https://github.com/mesonbuild/meson/issues/6385

bonzini:
In QEMU I think we have one use case for reusing the unique name. We generate files from many directories, and we would like to have them available from source files in the easiest way. With Makefiles we use basically `-I$(@D)`, so that can have:

* chardev/trace.h generated from chardev/meson.build, included as `#include "trace.h" within chardev/
* hw/trace.h from hw/meson.build, included as `#include "trace.h" within hw/
* and so on (there are about 60 of these)

It would be nice if you could have this as something like

    trace_h = generator(...)
    include_dir = g.private_dir_include(for_source_dir: true)

and `trace_h.process('trace.h', unique_name: 'trace_h')` in each directory.

Another possibility is, instead of having the unique_name, to have something like

bonzini:
Going back to your original proposal, I wonder if all that is needed is a "custom_target template" (suggesting that it could be created with `custom_target_template`). This would accept the same arguments as `custom_target` except `input` and the initial named argument; in addition, `output` would also support substitution of `@BASENAME@`, `@PLAINNAME@`, and possibly `@PRIVATE_DIR@`. It would then return a factory object similar to generators, so that later on you could do:

[...]

PR #9224, opened 2021-09-05
Support import() of out-of-tree modules

cmcneish:
This pull-request adds a 'dirs' kwarg to import(), and updates tests / documentation to include this parameter.

`import('foo', dirs: '/bar')` will search for /bar/foo.py, and attempt to load it as an external module, failing if it's not found.

---

I have a couple modules that I'm not able to merge upstream. With this change, I'm able to manage these as part of the projects they're related to, rather than by shipping my own package of Meson with them included.

### DIFF
--- a/docs/markdown/Modules.md
+++ b/docs/markdown/Modules.md
@@ -23,3 +23,12 @@ Meson has a selection of modules to make common requirements easy to
 use. Modules can be thought of like the standard library of a
 programming language. Currently Meson provides the modules listed on
 subpages.
+
+In addition to modules that come with Meson, it is possible to write
+your own. These can then be imported using the `dirs` kwarg:
+
+```meson
+mymod = import('somemodule', dirs : meson.project_source_root() / 'modules')
+```
+
+This will import the file `modules/somemodule.py` as a Meson module.
--- a/docs/markdown/Reference-manual.md
+++ b/docs/markdown/Reference-manual.md
@@ -1001,8 +1001,17 @@ a non-existing variable will cause a fatal error.
 ### import()
 
 ```
-  module_object import(string, required : bool | feature, disabler : bool)
+  module_object import(string, ...)
 ```
+
+Keyword arguments are the following:
+
+- `required` *(since 0.59.0)*: By default, `required` is `true` and
+  Meson will abort if the module could not be found.
+- `disabler` *(since 0.59.0)*: if `true` and the module couldn't be found,
+  returns a [disabler object](#disabler-object) instead of a not-found module.
+- `dirs` *(since 0.60.0)*: extra list of absolute paths where to look for program
+  names.
 
 Imports the given extension module. Returns an object that can be used to call
 the methods of the module. Here's an example for a hypothetical `testmod`
@@ -1012,8 +1021,6 @@ module.
     tmod = import('testmod')
     tmod.do_something()
 ```
-
-*Since 0.59.0* the required and disabler keyword arguments
 
 ### include_directories()
 
--- a/mesonbuild/interpreter/interpreter.py
+++ b/mesonbuild/interpreter/interpreter.py
@@ -70,6 +70,7 @@ import collections
 import typing as T
 import textwrap
 import importlib
+import importlib.util
 
 if T.TYPE_CHECKING:
     import argparse
@@ -518,9 +519,24 @@ class Interpreter(InterpreterBase, HoldableObject):
                 dep = df.lookup(kwargs, force_fallback=True)
                 self.build.stdlibs[for_machine][l] = dep
 
-    def _import_module(self, modname: str, required: bool) -> T.Union[ExtensionModule, NewExtensionModule, NotFoundExtensionModule]:
+    def _import_module(self, modname: str, required: bool, search_dirs: T.List[str]) -> T.Union[ExtensionModule, NewExtensionModule, NotFoundExtensionModule]:
         if modname in self.modules:
             return self.modules[modname]
+
+        for search_dir in search_dirs:
+            try:
+                module_file = os.path.join(search_dir, modname + '.py')
+                spec = importlib.util.spec_from_file_location(modname, module_file)
+                module = importlib.util.module_from_spec(spec)
+                spec.loader.exec_module(module)
+            except FileNotFoundError:
+                pass
+            else:
+                ext_module = module.initialize(self)
+                assert isinstance(ext_module, (ExtensionModule, NewExtensionModule))
+                self.modules[modname] = ext_module
+                return ext_module
+
         try:
             module = importlib.import_module('mesonbuild.modules.' + modname)
         except ImportError:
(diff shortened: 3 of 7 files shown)


eli-schwartz:
https://mesonbuild.com/FAQ.html#why-is-meson-not-just-a-python-module-so-i-could-code-my-build-setup-in-python

This is explicitly not a design goal of meson -- meson wants problems to be solved once, for everyone, and for meson.build files to be easily understood by reading the official meson manual / documentation.

Allowing people to code (large?) parts of their build systems in python and then invoke that as a custom module violates that design principle.

Furthermore, meson's `import()` module interface is 100% an internal implementation detail and commits to zero stability. We can and will change it whenever we please in order to make make it work better, or more consistently, or add missing safety or features to it, etc... additionally, actually doing something meaningful with a module means calling into the interpreter, instantiating build.Target and its subclasses, etc. Code refactoring happens, and is currently completely safe because meson itself is the only consumer and all code is updated to match the new interfaces.

tl;dr the only way third-party modules could work at all is if they were bound to a specific bugfix release of meson.

Finally, there are multiple projects currently working on re-implementing meson in C or C++, and currently they have the ability to do so because they only need to implement the official interface of any given module as documented, per the meson version compatibility the reimplementation announces support for.

All this goes out the window if we add an offiical python interface for extending the build system with custom modules, because none of that will EVER work in a C/C++ rewrite.

> I have a couple modules that I'm not able to merge upstream. With this change, I'm able to manage these as part of the projects they're related to, rather than by shipping my own package of Meson with them included.

Why not? Are the open source? Can they be made generic enough to be usable for more than one project?


...


Is this something that could be solved, instead, by https://github.com/mesonbuild/meson/issues/6526#issuecomment-579541292 (e.g. something like python `functools.partial()` but for meson `custom_target()`)? Or #5209 ?

cmcneish:
Thanks for the quick response. I sort of expected this to get closed, but I was hoping for some discussion.

For reference, my usage of these external modules is essentially:

```
mm = import('mymodule')
foo = mm.add_executable('foo')
mm.test('foo', foo)
```

All the other builtin functions (library(), dependency(), ...) work fine, but there's special configuration required to build or test.

---

I'll start from 
>Why not? Are the open source? Can they be made generic enough to be usable for more than one project?

No, and no.

The problem is essentially covered by [this documentation on proprietary toolchains](https://mesonbuild.com/FAQ.html#i-have-proprietary-compiler-toolchain-x-that-does-not-work-with-meson-how-can-i-make-it-work), but driving the toolchain (names and arguments and such) is also private. There's no simple way to hack into the `mesonbuild/environment.py` file to add common steps, from Meson. But, I can sidestep that with `import()`, and just duplicate a bit of functionality.

---

> Is this something that could be solved, instead, by #6526 (comment) (e.g. something like python functools.partial() but for meson custom_target())? Or #5209 ?

Future improvements versus present need?

Everything I want implementable currently, with overuse of generators and a subdirectory of meson.build files to organize the build tree. When I had it that way, it was larger and harder to manage / synchronize across projects than my solution using a module.

#5209 sure, a bit. `include('something.meson)` would flatten and simplify my subdirectory of build files, but would need extension to `generator` to tidy up intermediate build outputs.

#6526, not really. My two problems with generators/custom_targets are the syntax and the argument passing contract. The commented proposal helps a little bit with the syntax, but all parameters are still passed as strings (this is sort of also an issue with `include()`).

---

> Furthermore, meson's import() module interface is 100% an internal implementation detail and commits to zero stability

> tl;dr the only way third-party modules could work at all is if they were bound to a specific bugfix release of meson.

This isn't really any different from the unstable-modules situation, from a user's standpoint.

---

> Finally, there are multiple projects currently working on re-implementing meson in C or C++, and currently they have the ability to do so because they only need to implement the official interface of any given module as documented, per the meson version compatibility the reimplementation announces support for.

None of them are very far along in implementing modules, that I know of:

https://github.com/dcbaker/meson-plus-plus#modules
>  I simply do not have the time and energy to re-implement every one of them to feature parity.

https://git.sr.ht/~lattis/muon/tree/master/item/src/functions/modules.c#L28
```
static const struct func_impl_name *module_func_tbl[module_count] = {
	[module_fs] = impl_tbl_module_fs,
};
```

https://git.sr.ht/~bl4ckb0ne/boson/tree/master/item/src/builtin.c#L329
```
	{"import", todo},
```

I know the forks are all fairly new, but they only seem interested in implementing core Meson, at least for now.

---

> All this goes out the window if we add an official python interface for extending the build system with custom modules, because none of that will EVER work in a C/C++ rewrite.

Well, what are my alternatives? I appreciate your stance, but I really dislike my generator/custom_target setup

jpakkane:
> All the other builtin functions (library(), dependency(), ...) work fine, but there's special configuration required to build or test.

The relevant question here then is what is this "special configuration" and whether there is a better way to specify it. For example when running tests you can specify a wrapper executable that can be used to invoke the real test. See `meson test --help` for details.

Similarly if your toolchain has been designed poorly and requires setting some envvars or something to function you can make it work with a wrapper script like this:

```
#1/bin/sh

FOO=bar
/path/to/real/program "$*"
```

(The actual syntax might be different, that is just from memory to illustrate the point.)

eli-schwartz:
> There's no simple way to hack into the `mesonbuild/environment.py` file to add common steps, from Meson.

Hmm, perhaps it *should* be possible to stub out a compiler, and let people describe theirs using an ini file. I wonder if anyone has tried seeing how easy that would be? Food for thought. (I have no idea.)

eli-schwartz:
> This isn't really any different from the unstable-modules situation, from a user's standpoint.

Sure it is different. unstable-modules are tested to work, and any changes to them should always have a way forward. We do reserve the right to change their final API, but only in the sense that they are provisional while we figure out what people need and how they want to use the module. We don't change things willy nilly.

Mostly, though, a change to an unstable module should never EVER even in the worst case, result in a python traceback. And, it will never change in a bugfix release because those have a compatibility contract and changing the meson.build API as a side effect of a bugfix is itself a new bug and must be fixed.

Changes to the python backend which break a user module coded in python will result in python tracebacks, and there is no reason to think that a bugfix release will be immune to such changes.

> None of them are very far along in implementing modules, that I know of:
> 
> https://github.com/dcbaker/meson-plus-plus#modules
> 
> > I simply do not have the time and energy to re-implement every one of them to feature parity.

This seems like an unfair analysis, since you excluded "I will make an effort to keep a few of the very common and important ones working, but help is very much welcome for the rest."

> I know the forks are all fairly new, but they only seem interested in implementing core Meson, at least for now.

This is reasonable for something fairly new where the MVP is to get the core working, so people actually have something to look at. One presumes that having reached that point, people will be happy to help complete them -- just like people helped out to create the original modules in python.

It's little different from saying "they haven't implemented it yet, so too bad for them".

cmcneish:
Thank you both for the comments, I appreciate the insight

---

> The relevant question here then is what is this "special configuration" and whether there is a better way to specify it

> Similarly if your toolchain has been designed poorly and requires setting some envvars or something to function you can make it work with a wrapper script like this:

Sort of...

Some peseudocode:
```
for test in tests:
    testlist.write(some_formatter(test))
custom_target('test') = command(some_program, testlist)
```

So, thinking about this again, I think it's actually doable. I can have the stub programs write to the testlist, and then make the test execution happen last. Reading through the docs, I think `is_parallel` and `priority` can be juggled to make this happen. I could also just sidestep `test()` and make a 'custom-test' custom_target do the test execution.

For the rest of the "special configuration", I think it would be covered by Eli's suggestion, plus a finalization step for my applications:

> Hmm, perhaps it should be possible to stub out a compiler, and let people describe theirs using an ini file. I wonder if anyone has tried seeing how easy that would be? Food for thought. (I have no idea.)

This could work, I'll have to think about it some more.

> unstable-modules

Makes sense, thanks for the clarification. Sorry, I came off a little harsh towards the reimplementations.

cmcneish:
> ini file for new toolchains

One potential sticking point: this would effectively create an alternate way of cross compiling, in addition to the machine file setup.

I'm thinking it would go something like, I don't have a "cross compiler", I have a C+++ compiler which happens to behave exactly like a C++ compiler. And, like C++, and D can coexist in the same meson invocation, so can C++ and C+++.

I don't think that's something ~any~ many people will do, but it makes it possible

PR #9228, opened 2021-09-07
Patch dir names independent from subprojects dir names

### LINKED ISSUE #9085: [enhancement] patch overlaying independent from directories names
If we want to maintain Meson overlay repository for any external package it is important that patch directory must be named as package directory.

But this is too hard to maintain because usually names of patch directories produced by 3rd party code (for example, github creates .zip files for each release which contain repository directory name in form `repo_name-1.2.3`)

And if we try to equalize directories by wrap `revision` directive then we broke versions.

In other side, it must be strictly one root directory of package and one root directory for overlaying patch. So, It will be handy if overlaying files will be added independenly from source and destination directory names.

Side effect: Meson will not create subprojects directories accidentally (or will not patch another subpackage which not described in current .wrap file :-) ) if bad names was used. It is make sense if we need to replace some files by patching instead of add - sometimes is too hard to find accidentally unpatched project.

xggrnx:
Fixes #9085

### DIFF
--- a/mesonbuild/wrap/wrap.py
+++ b/mesonbuild/wrap/wrap.py
@@ -14,6 +14,7 @@
 
 from .. import mlog
 import contextlib
+import glob
 import urllib.request
 import urllib.error
 import urllib.parse
@@ -87,6 +88,12 @@ class WrapException(MesonException):
     pass
 
 class WrapNotFoundException(WrapException):
+    pass
+
+class WrapPatchFormatException(WrapException):
+    pass
+
+class PatchUnpackException(WrapException):
     pass
 
 class PackageDefinition:
@@ -566,6 +573,14 @@ class Resolver:
 
             return path.as_posix()
 
+    def unpack_archive(self, path):
+        with tempfile.TemporaryDirectory() as tmpdir:
+            shutil.unpack_archive(path, tmpdir)
+            src = glob.glob(f'{tmpdir}{os.path.sep}*{os.path.sep}')
+            if len(src) != 1:
+                raise WrapPatchFormatException("Error patch archive format.")
+            self.copy_tree(src[0], os.path.join(self.subdir_root, self.wrap.name))
+
     def apply_patch(self) -> None:
         if 'patch_filename' in self.wrap.values and 'patch_directory' in self.wrap.values:
             m = 'Wrap file {!r} must not have both "patch_filename" and "patch_directory"'
@@ -573,11 +588,9 @@ class Resolver:
         if 'patch_filename' in self.wrap.values:
             path = self.get_file_internal('patch')
             try:
-                shutil.unpack_archive(path, self.subdir_root)
+                self.unpack_archive(path)
             except Exception:
-                with tempfile.TemporaryDirectory() as workdir:
-                    shutil.unpack_archive(path, workdir)
-                    self.copy_tree(workdir, self.subdir_root)
+                raise PatchUnpackException(f"Can't unpack patch {self.wrap.name}")
         elif 'patch_directory' in self.wrap.values:
             from ..interpreterbase import FeatureNew
             FeatureNew('patch_directory', '0.55.0').use(self.current_subproject)


mesonbuild/wrap/wrap.py
```diff
@@ -89,6 +90,12 @@ class WrapException(MesonException):
 class WrapNotFoundException(WrapException):
     pass
 
+class WrapPatchFormatException(WrapException):
+    pass
+
+class PatchUnpackException(WrapException):
+    pass
+
```
xclaesse, line 98: I don't think it's worth adding specialised exceptions if you don't catch them anywhere. WrapNotFoundException above is because we need to catch that specific case for special treatment in interpreter IIRC.

mesonbuild/wrap/wrap.py
```diff
@@ -566,18 +573,24 @@ def get_file_internal(self, what: str) -> str:
 
             return path.as_posix()
 
+    def unpack_archive(self, path):
+        with tempfile.TemporaryDirectory() as tmpdir:
+            shutil.unpack_archive(path, tmpdir)
+            src = glob.glob(f'{tmpdir}{os.path.sep}*{os.path.sep}')
+            if len(src) != 1:
+                raise WrapPatchFormatException("Error patch archive format.")
```
xclaesse, line 581: If I understand correctly, this assumes the tarballs contains a single toplevel directory and we need to copy what it contains, right? `os.listdir(tmpdir)` seems more straightforward.

But that's not the format we were assuming before, won't it break the case where the tarball does not have a toplevel directory and must be extracted directly onto the subproject?

denizzzka: > But that's not the format we were assuming before

Probably, no one actually uses or used this behaviour: current `~master` unpacks patch files from these archives without directories into `subprojects/` directly.

(Manual about patch archives extracting also can be understood in two ways, as for me)

xclaesse: oh we used to extract in `subprojects/` and not `subprojects/foo`, ok. Wondering why we were catching exceptions and then extract into a tempdir, do you now what kind of exceptions we could have there?

denizzzka: I don't know Python :-)

mesonbuild/wrap/wrap.py
```diff
@@ -89,6 +89,9 @@ class WrapException(MesonException):
 class WrapNotFoundException(WrapException):
     pass
 
+class WrapPatchFormatException(WrapException):
+    pass
+
```
xclaesse, line 94: You can remove this one too, I think.

mesonbuild/wrap/wrap.py
xclaesse: You can raise WrapException() here, ideally with a more specific error message, saying we expected the archive to only contain a single directory.

mesonbuild/wrap/wrap.py
```diff
@@ -566,18 +569,24 @@ def get_file_internal(self, what: str) -> str:
...
     def apply_patch(self) -> None:
         if 'patch_filename' in self.wrap.values and 'patch_directory' in self.wrap.values:
             m = 'Wrap file {!r} must not have both "patch_filename" and "patch_directory"'
             raise WrapException(m.format(self.wrap.basename))
         if 'patch_filename' in self.wrap.values:
             path = self.get_file_internal('patch')
             try:
-                shutil.unpack_archive(path, self.subdir_root)
+                self.unpack_archive(path)
             except Exception:
```
xclaesse, line 588: This try/except block is useless now, just call self.unpack_archive() without try.

mesonbuild/wrap/wrap.py
eli-schwartz: The exception causes it to unpack into a tempdir, and then invoke self.copy_tree which is an "enhanced copy" that also tries to chmod existing read-only files before copying a new version of the file. shutil.unpack_archive would just error out by default.

This has nothing to do with changing the extraction directory name.

denizzzka:
Seems like codecov isn't accessible during CI

eli-schwartz:
The CI failure has nothing to do with codecov, it failed much earlier on at e.g. https://dev.azure.com/jussi0947/jussi/_build/results?buildId=11364&view=logs&j=678c5357-36f0-50cc-b96e-1a6f7149eed2&t=fb842c1e-cd65-5c1e-ab4f-d7ba55deb214&l=1997

eli-schwartz:
As far as I can tell, this PR doesn't really have anything to do with the linked ticket, which wanted patch archives to not have a base directory component at all, while this PR still requires one.

denizzzka:
> As far as I can tell, this PR doesn't really have anything to do with the linked ticket

If you about  #9085 then in my opinion it does exactly what is meant in this ticket
_-- #9085 author_

mesonbuild/wrap/wrap.py
```diff
@@ -566,18 +566,21 @@ def get_file_internal(self, what: str) -> str:
 
             return path.as_posix()
 
+    def unpack_archive(self, path):
+        with tempfile.TemporaryDirectory() as tmpdir:
+            shutil.unpack_archive(path, tmpdir)
+            src = os.listdir(tmpdir)
+            if len(src) != 1:
+                raise WrapException("Error patch archive format.")
+            self.copy_tree(src[0], os.path.join(self.subdir_root, self.wrap.name))
```
denizzzka, line 575: Probably, `self.wrap.name` always contains name obtainted from `.wrap` filename without version suffix like `foo-1.0` and this causes CI fail in case whan wrap filename isn't same as directory of downloaded subproject

eli-schwartz:
I don't think accepting `hahathisisafunnyname/meson.build` and extracting it to `zlib-1.2.11/meson.build` fits the stated goal of that ticket.

Moreover, I'm not entirely convinced it should be a goal.


It was, however, also mentioned in that ticket that a malformed patch is not stopped on a technical level from patching the wrong thing... I would approve of a patch that continued to require the directory names to match, but added a check that there is only one directory and it has the right name

mesonbuild/wrap/wrap.py
eli-schwartz: Correct, the `{name}.wrap` is probably never (or rarely) the same as the `directory=`

denizzzka:
I don't think as you about this PR: I think wrap functionality is initially was weak by dealing with all these top-level directories. This PR makes wraps better as possible in this situation and prevents some possible errors.

mesonbuild/wrap/wrap.py
```diff
@@ -566,18 +566,21 @@ def get_file_internal(self, what: str) -> str:
 
             return path.as_posix()
 
+    def unpack_archive(self, path):
+        with tempfile.TemporaryDirectory() as tmpdir:
+            shutil.unpack_archive(path, tmpdir)
+            src = os.listdir(tmpdir)
+            if len(src) != 1:
+                raise WrapException("Error patch archive format.")
```
xclaesse, line 574: Last small nitpicking: The error message could be a bit more informative. Should tell we expect the archive to contain only 1 top directory.

xggrnx: @xclaesse  error message == "Archive must contain only one top directory" ok ?

eli-schwartz:
I still disagree with this change.

denizzzka:
>  or do not have a directory name at all 

> Note that actually changing to store files as meson.build instead of fooproject-1.0/meson.build would break current wraps and probably require a config key... It's not clear if this is sufficiently useful.

Earlier we found out that this functionality is broken long time ago and, since no one wrote an issue ticket, it means that no one needs it and not uses it. Also, top-level directory inside of archive it is a old good practice.

> I strongly object to having data, but ignoring it.

... Directory name of patch equal to subproject name is _redundant_ data - we already know what name of subproject directory, here isn't any sense to duplicate it

denizzzka:
Please don't let this PR gone out from sight and rot!

denizzzka:
Bump

xclaesse:
Do I understand correctly that the desired use-case here is to have a patch tarball that contains a single directory `foo` and apply it on `subprojects/foo-1.2.3`, that way you can keep the tarball unchanged when bumping the version of foo to e.g. `foo-1.2.4`?

IMHO that makes sense, and it's something that has annoyed me in the past too. But on the other hand, we probably want to make sure we don't apply a totally unrelated patch, like applying `bar.tar.gz` onto `foo-1.2.3`. Should we check that the single directory in the patch tarball must be a prefix of the directory it applies on? Copy `foo/` to `foo-1.2.3/` is fine, but not copy `foo/` to `bar-1.2.3`?

Should we do that only if we set a special key in the wrap file? This sounds similar to `patch -p1` case, and also similar to the [`lead_directory_missing`](https://mesonbuild.com/Wrap-dependency-system-manual.html#specific-to-wrapfile) we already have. Maybe add `patch_skip_lead_directory=1`?

denizzzka:
>Do I understand correctly that the desired use-case here is to have a patch tarball that contains a single directory foo and apply it on subprojects/foo-1.2.3, that way you can keep the tarball unchanged when bumping the version of foo to e.g. foo-1.2.4?

Yes

>IMHO that makes sense, and it's something that has annoyed me in the past too. But on the other hand, we probably want to make sure we don't apply a totally unrelated patch, like applying bar.tar.gz onto foo-1.2.3.

Patch is specified by user manually, there is no reason for doubt (and we don't doubt now)

>Should we check that the single directory in the patch tarball must be a prefix of the directory it applies on? Copy foo/ to foo-1.2.3/ is fine, but not copy foo/ to bar-1.2.3?

My opinion - no. This is an unnecessary complication that will only forces to maintain redundant information about subproject name. It is enough to check that there is only one directory provided, it does not really matter what its name is.

>Should we do that only if we set a special key in the wrap file? This sounds similar to patch -p1 case, and also similar to the [lead_directory_missing](https://mesonbuild.com/Wrap-dependency-system-manual.html#specific-to-wrapfile) we already have. Maybe add patch_skip_lead_directory=1?

In this case we won't fix the current dangerous behavior: malformed patch archive silently can not patch anything (due to dirs name mismatch), accidentally patch few packages at once (few dirs in one archive)

denizzzka:
@xclaesse 

I see what only one case then this patch can break build: an **initially incorrect** directory name inside of archive - mismatch between dir name in user's wrap archive and needed name as described in current Meson documentation.

After Meson update it will correctly apply such patch and build will go not usually as before.

eli-schwartz:
> In this case we won't fix current dangerous behavior: malformed patch archive silently can't patch anything (due to dirs name mismatch) or accidentally can patch few packages at once (few dirs in one patch archive)

So the solution to this is pretty simple: no changes are needed to meson above and beyond making it not-silent.

denizzzka:
@eli-schwartz We can get more functionality with less effort here. Brainfuck with these names is really tired.

xclaesse:
My suggestion would be to add `patch_lead_directory_missing=1` in wrap files, that's nicely symetrical with what we already support for source archives. Then you can create patch overlays with no root directory inside. It has the bonus benefit of making simple overlays much easier to create, at least in GNOME you can rght-click a `meson.build` file -> compress, and you get a zip file containing only the meson.build file and no lead directory.

denizzzka:
@xclaesse it is also ok for me personally

And, in any case, we also should add check for:

> fix current dangerous behavior: malformed patch archive silently can't patch anything (due to dirs name mismatch) or accidentally can patch few packages at once (few dirs in one patch archive)

denizzzka:
@xclaesse 

> My suggestion would be to add patch_lead_directory_missing=1 in wrap files, that's nicely symetrical with what we already support for source archives

This does not solve one problem (which I desire as important) well described in initial issue here: https://github.com/mesonbuild/meson/issues/9085

_usually names of patch directories produced by 3rd party code (for example, github creates .zip files for each release which contain repository directory name in form repo_name-1.2.3)_

denizzzka:
> Owners of broken(!) archives will get a problem

But it is possible to display warning if directory name check failed and specified in meson.build Meson version is old or ommited

xclaesse:
@denizzzka not sure to understand what it would not support. If it extracted source tarball into subprojects/foo-1.0 and patch overlay does not have a root directory, meson can extract the overlay directly into subprojects/foo-1.0.

denizzzka:
@xclaesse if patch archive maintained as GitHub (or analogous) repository (what is natural and comfortable) it is impossible to control name of root directory or its absense or use arbitrary archive format. Releases archives provided by GitHub itself in its own format and can't be changed

xclaesse:
Oh, you want to maintain your overlay as a github repo? I guess that makes sense but we did not expect people doing that...

Maybe we need both `patch_lead_directory_missing=true` and `[patch,source]_ignore_lead_directory=true`, maybe the latter should imply the former.

denizzzka:
I just don't understand why Meson wants to hold on matching internal directory names inside of archives? Currently (and from very beginning) names aren't checked by Meson code. You can easily lower requirements for these names (and in documentation too). After all, these requirements are initially redundant.

eli-schwartz:
It is trivial to re-tag an existing commit with a second tag.

Adding a dozen options to the documentation to opt in to disabling sanity checks in various ways feels very wrong. Heuristically assuming that sanity checks should be disabled feels wrong.


The use case still feels non-obvious to me, especially because wrap files typically change in order to update the version number in `project()` / `declare_dependency()`, which is necessary to make `dependency('foobar', version: '>=1.2')` work without

```
Dependency foobar found: NO found undefined but need: '`>=1.2' (overridden)

ERROR: Dependency 'foobar' is required but not found.
```

denizzzka:
> I guess that makes sense but we did not expect people doing that...

No other real choice

Yes, it is possible to install my own _properly repacking server_ for Meson wrap archives, which from time to time will checks GitHub (etc) repositories and host correctly made "repacks" extractly as Meson wants. But this looks like a bureaucratic crutch.

xclaesse:
> Currently (and from very beginning) names aren't checked by Meson code

Indeed, that's really bad and we should do check that. IMHO a fix should go toward safer code and not "we were not checking before, so let's just document that we won't check".

eli-schwartz:
There's a reason I said "typically" instead of "always".

denizzzka:
@xclaesse 
> Indeed, that's really bad and we should do check that. IMHO a fix should go toward safer code and not "we were not checking before, so let's just document that we won't check".

But also this is a chanse to change current behaviour to less strict without any issues. It's actually not convenient (not just on mine case) to have to exactly matched directory name inside of the archive. I am absolutely sure that anyone who tried to create such archives have encountered problems in this place.

PR #9229, opened 2021-09-07
gnome: Fix expand-content-files in gtkdoc generation

inigomartinez:
Unlike `content-files` parameter, `expand-content-files` parameter doesn't allow to use generated target files, even though they are similar.

The same code used for `content-files` has been extracted and now is shared with `expand-content-files` so it can also used generated target files.

### DIFF
--- a/mesonbuild/modules/gnome.py
+++ b/mesonbuild/modules/gnome.py
@@ -1025,6 +1025,37 @@ class GnomeModule(ExtensionModule):
         rv = [inscript, pottarget, potarget]
         return ModuleReturnValue(None, rv)
 
+    def _get_content_files_args(self, kwargs, state, expand_content_files):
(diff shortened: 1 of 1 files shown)


eli-schwartz:
Would be nice to have a test that uses generated files, or that even uses expand_content_files at all.

Also this parameter to gnome.gtkdoc() is completely undocumented. :(

inigomartinez:
> Would be nice to have a test that uses generated files, or that even uses expand_content_files at all.
> 
> Also this parameter to gnome.gtkdoc() is completely undocumented. :(

I don't have enough knowledge to add a test using `expand_content_files` at the moment. I have used it only to port application/libraries using it.

However, I've added it to the documentation so it will be present for anyone looking for the available parameters.

PR #9230, opened 2021-09-07
summary: fix dependencies

bonzini:
Dependencies are currently printed as

   [<mesonbuild.mlog.AnsiDecorator object at 0x7faa85aeac70>, ' ', <mesonbuild.mlog.AnsiDecorator object at 0x7faa85aeab50>]

This was introduced in commit adb1b2f3f6ad54b346348ec6e5b8d96f2f7ba0a6, due to
an incorrect type annotation on the AnsiText constructor.  Fix both the
annotation and the usage.

This is a regression in 0.59.

### DIFF
--- a/mesonbuild/_typing.py
+++ b/mesonbuild/_typing.py
@@ -36,6 +36,9 @@ T = typing.TypeVar('T')
 class StringProtocol(Protocol):
     def __str__(self) -> str: ...
 
+class SizedStringProtocol(Protocol, StringProtocol, typing.Sized):
+    pass
+
 class ImmutableListProtocol(Protocol[T]):
 
     """A protocol used in cases where a list is returned, but should not be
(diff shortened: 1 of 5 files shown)


xclaesse:
What I don't understand is https://github.com/mesonbuild/meson/commit/adb1b2f3f6ad54b346348ec6e5b8d96f2f7ba0a6 added a unit test, why has this not been catched? Could you please add a test that fails before this PR and gets fixed by it?

bonzini:
The unit test didn't catch it because it only happens with external dependencies. It is a bit messy to test external dependencies in the existing test_summary, but I've added it now.

bonzini:
@nirbheek can you add this to the 0.59.2 milestone?

xclaesse:
Thanks, merged and tagged.

PR #9238, opened 2021-09-10
objc/objcpp compiler: accept all gnu stds corresponding to c/c++ stds

### LINKED ISSUE #9237: Value "gnu89" (of type "string") for combo option "C language standard to use" is not one of the choices. Possible choices are (as string): "none", "c89", "c99", "c11", "c17", "gnu99", "gnu11".
Building glib2 2.62.6 with meson 0.59.1 fails:

```
Value "gnu89" (of type "string") for combo option "C language standard to use" is not one of the choices. Possible choices are (as string): "none", "c89", "c99", "c11", "c17", "gnu99", "gnu11".
```

With meson 0.58.1 there was no such error.

See https://trac.macports.org/ticket/63462

In #8766 gnu99 and gnu11 were added to the list of possible choices but for some reason gnu89 was not added? Could you add it, please?

eli-schwartz:
The clang compiler now reimplements and re-checks the c_std and cpp_std options in order to use them for objc as well, but it didn't consistently support the same options. First it completely excluded all the gnu ones, and then it added a handful of them but not for C++.

Be fully consistent -- or at least as consistent as we can be, given a minimally working fix. (The C/C++ compiler mixin actually gates different stds depending on detected clang version, we do not do that here.)

Fixes regression in c54dd63547b030e3d9feee694ec6f49c434f0df8
Fixes incomplete fix from #8766 (which didn't fix objcpp at all)

Fixes #9237

### DIFF
--- a/mesonbuild/compilers/objc.py
+++ b/mesonbuild/compilers/objc.py
@@ -90,7 +90,7 @@ class ClangObjCCompiler(ClangCompiler, ObjCCompiler):
         opts.update({
             OptionKey('std', machine=self.for_machine, lang='c'): coredata.UserComboOption(
                 'C language standard to use',
-                ['none', 'c89', 'c99', 'c11', 'c17', 'gnu99', 'gnu11'],
+                ['none', 'c89', 'c99', 'c11', 'c17', 'gnu89', 'gnu99', 'gnu11', 'gnu17'],
                 'none',
             )
         })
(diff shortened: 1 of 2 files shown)


dcbaker (approved):
lookgs good to me. We should probably refactor this so that the C/ObjC and C++/ObjC++ compilers can share code, but this seems like a good improvement.

PR #9243, opened 2021-09-12
Die, layout=flat, die

### LINKED ISSUE #996: vala: Building with flat layout does not work
It does not seem to resolve the generated header path properly:

```
[5/7] Compiling c object 'meson-out/valaprog@exe/meson-out_valaprog@exe_prog.c.o'
FAILED: meson-out/valaprog@exe/meson-out_valaprog@exe_prog.c.o 
cc  '-Imeson-out/valaprog@exe' '-fdiagnostics-color=always' '-I../prog' '-Iprog' '-Ilib' '-I../lib' '-pipe' '-w' '-O0' '-g' '-I/usr/include/glib-2.0' '-I/usr/lib64/glib-2.0/include' '-MMD' '-MQ' 'meson-out/valaprog@exe/meson-out_valaprog@exe_prog.c.o' '-MF' 'meson-out/valaprog@exe/meson-out_valaprog@exe_prog.c.o.d' -o 'meson-out/valaprog@exe/meson-out_valaprog@exe_prog.c.o' -c 'meson-out/valaprog@exe/prog.c'
meson-out/valaprog@exe/prog.c:9:21: erreur fatale : valalib.h : No such file or directory
 #include "valalib.h"
                     ^
compilation terminée.
[6/7] Generating symbol file 'meson-out/valalib@sha/meson-out/libvalalib.so.symbols'.
```

Also, there should be some CI support to run all the test cases using the flat layout.

I'll probably look into this.

### LINKED ISSUE #1521: Missing shared library links in flat layout
I have a project with a few libraries built with `version` (`0.0.0` in this case, but it doesn’t matter) and plugins. I use the flat layout so I can set the build dir as a plugin dir and load them all for tests.
Recently I added some API in one of the library, and my (manual) test suddenly failed due to a missing symbol.

It appears that Meson, with the flat layout, doesn’t create the `liblib.so.0 -> liblib.so.0.0.0` symlink in the build dir. Removing my system installed package confirms this is the issue.

### LINKED ISSUE #1908: Executable names clash between the main project and subprojects
I am not in a position to be able to build a simple test case for this right now but I think what I am experiencing is that a test() from the main project ends up running an executable that is built by a subproject, because its name happens to be the same as another executable in the main project.

I'm not sure what's the best solution for this, probably an error should be reported, the build aborted?

### LINKED ISSUE #7133: Crash in gnome module when using -Dlayout=flat

### LINKED ISSUE #7480: The ’layout’ option needs some work for Rust projects
I did some investigation on the phenomenon  where any attempt to building a `rust` project ends with a missing chunk of likable code. So I ran some experiments and found that nothing works until I remove the usage of `layout` option assigned to `flat` that things started to work again. 

So my conclusion is that `layout` option needs some work for Rust projects with the value `flat`.

### LINKED ISSUE #8378: configure_file not working with vs and layout=flat
**Describe the bug**
On a toy project following [the manual](https://mesonbuild.com/Configuration.html#a-full-example), using layout=flat on vs backend fails because the include path is incorrect.

**To Reproduce**
[meson_layout_flat.zip](https://github.com/mesonbuild/meson/files/5996320/meson_layout_flat.zip)
meson build --backend=vs --layout=flat

**Expected behavior**
compilation will fail with 'projconfig.h' : No such file or directory
while it compiles with meson build --backend=vs

**system parameters**
* native build
* windows 7 64bits
* Python 3.8.7
* `meson --version` = 0.56.2
* `ninja --version` = 1.10.2

eli-schwartz:
The option for layout=flat is horribly broken beyond repair in various circumstances which no one really keeps track of, no one seems to actually use it in practice for any good reason, CI doesn't test it, no one is committed to maintaining it, etc. etc. etc.

How is it broken? Well:
- projects doing trivially reasonable things, such as generating "foo/util.py" and "bar/util.py", create clashing output names. This will never, ever, ever, ever work with layout=flat. This is by design, because they are fundamentally incompatible goals.
- Generally, meson assumes all over the place that files go in their subdir output directories, and since it is never tested in CI, these assumptions go unnoticed...
- Sometimes meson itself generates files in subdir-based layouts, for example my recent changes to the i18n module produce dozens of files all named "project.mo" but in directories distinguished by unique locales. This is because locale files are all named the same thing, and gettext chooses the right file by searching a locale directory... there is nothing we can do here! There are probably also other generated cases, but who even cares to try to figure out what they are???

I propose that layout=flat is a failed experiment. It only ever existed in the first place, because... Windows. Windows is technically  broken. It relies on PATH to find dlls which executables depend on, so programs don't run while uninstalled if subdirs are in play. Enter `layout=flat` just stick everything all in one directory, surely nothing can possibly go wrong... except it does.

There is a much better way to handle this, and recent versions of meson include it: `meson devenv` will let you enter an environment where your PATH is set up properly and you can run uninstalled programs. Let's deprecate and remove layout=flat in favor of `devenv`. Everyone will thank us.

Closes #996
Closes #1521
Closes #1908
Closes #7133
Closes #7135
Closes #7480
Closes #8378

(and probably more, who knows)

### DIFF
--- a/mesonbuild/msetup.py
+++ b/mesonbuild/msetup.py
@@ -177,8 +177,6 @@ class MesonApp:
         mlog.initialize(env.get_log_dir(), self.options.fatal_warnings)
         if self.options.profile:
             mlog.set_timestamp_start(time.monotonic())
-        if env.coredata.options[mesonlib.OptionKey('backend')].value == 'xcode':
-            mlog.warning('xcode backend is currently unmaintained, patches welcome')
         with mesonlib.BuildDirLock(self.build_dir):
             self._generate(env)
 
@@ -264,6 +262,11 @@ class MesonApp:
 
             # Post-conf scripts must be run after writing coredata or else introspection fails.
             intr.backend.run_postconf_scripts()
+
+            # collect warnings about unsupported build configurations; must be done after full arg processing
+            # by Interpreter() init, but this is most visible at the end
+            if env.coredata.options[mesonlib.OptionKey('backend')].value == 'xcode':
+                mlog.warning('xcode backend is currently unmaintained, patches welcome')
         except Exception as e:
             mintro.write_meson_info_file(b, [e])
             if 'cdf' in locals():
--- a/mesonbuild/msetup.py
+++ b/mesonbuild/msetup.py
@@ -267,6 +267,13 @@ class MesonApp:
             # by Interpreter() init, but this is most visible at the end
             if env.coredata.options[mesonlib.OptionKey('backend')].value == 'xcode':
                 mlog.warning('xcode backend is currently unmaintained, patches welcome')
+            if env.coredata.options[mesonlib.OptionKey('layout')].value == 'flat':
+                mlog.warning('-Dlayout=flat is unsupported and probably broken. It was a failed experiment at '
+                             'making Windows build artifacts runnable while uninstalled, due to PATH considerations, '
+                             'but was untested by CI and anyways breaks reasonable use of conflicting targets in different subdirs. '
+                             'Please consider using `meson devenv` instead. See https://github.com/mesonbuild/meson/pull/9243 '
+                             'for details.')
+
         except Exception as e:
             mintro.write_meson_info_file(b, [e])
             if 'cdf' in locals():


dcbaker (approved):
This looks good to me. I agree, unless someone wants to step up and get this working and in CI, removing it seems like the right decision to me.

xclaesse:
@eli-schwartz Seems worth adding a release snippet.

tabedzki:
I only recently started using meson and thought this would be nice to have. However, since this is a pain to maintain and has been deprecated, let's remove it from https://mesonbuild.com/Builtin-options.html so new users don't accidentally stumble upon it.

eli-schwartz:
One thing some people do is use `subdir()` to collect files into a variable, and then define the actual target in the toplevel meson.build using a much simpler one-line declaration that doesn't use bulky filelists.

tabedzki:
Pardon my inexperience, but how do you do that? I have a gpu_files=['file1', 'file2'] within the `src/meson.build`. Do I just need to append a 'src/` in front of all the files for that?

eli-schwartz:
```meson
gpu_files = files('file1', 'file2')
```

You can now use the `gpu_files` variable from ***anywhere***, any subdir at all, and it is guaranteed to always correctly refer to the right files.

tabedzki:
Ahh ok; thanks!

PR #9244, opened 2021-09-12
 Fixes .vcxproj for vs2017 vs2019 fixes #6314

### LINKED ISSUE #6314: Visual Studio 2019 backend doesn't pick up c++ standard version

qmor:
…h stdcpp version and LanguageStandard_C tag with stdc version in .vcxproj file

### DIFF
--- a/mesonbuild/backend/vs2010backend.py
+++ b/mesonbuild/backend/vs2010backend.py
@@ -1119,15 +1119,7 @@ class Vs2010Backend(backends.Backend):
         else:
             ET.SubElement(clconf, 'FavorSizeOrSpeed').text = 'Speed'
         # Note: SuppressStartupBanner is /NOLOGO and is 'true' by default
-        if self.name in ('vs2017', 'vs2019'):
-            if 'cpp' in file_args:
-                optargs = [x for x in file_args['cpp'] if x.startswith('/std:c++')]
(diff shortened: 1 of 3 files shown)


jpakkane:
I don't know enough about VS to make reasonable comments. Does someone else know, maybe @jenatali?

jenatali (approved):

eli-schwartz:
Could you please fix the commit message?
- the summary line is very long and overly detailed, and some of that should be moved to the extended description paragraph
- `fixes 6314` does not include a `#` character in the ticket number, so github does not automatically linkify this and mark the ticket as closed once this PR is merged

qmor:
I suggest to move this PR to milestone 0.60.2 because https://github.com/mesonbuild/meson/pull/9390 depend on this PR

PR #9248, opened 2021-09-13
Fix fortran

HaoZeke:
This PR ensures that the test programs can be compiled with standards conforming compilers like conda's `gfortran` and others.

- Fixes compiler check
- Updates tests
- ~Changes `.gitignore` to ignore generated `fortran`~

### DIFF
--- a/mesonbuild/compilers/fortran.py
+++ b/mesonbuild/compilers/fortran.py
@@ -72,7 +72,7 @@ class FortranCompiler(CLikeCompiler, Compiler):
         if binary_name.is_file():
             binary_name.unlink()
 
-        source_name.write_text('print *, "Fortran compilation is working."; end', encoding='utf-8')
+        source_name.write_text('program main; print *, "Fortran compilation is working."; end program', encoding='utf-8')
 
         extra_flags: T.List[str] = []
         extra_flags += environment.coredata.get_external_args(self.for_machine, self.language)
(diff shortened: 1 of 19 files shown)


eli-schwartz (changes requested):
> Changes `.gitignore` to ignore generated `fortran`

It is ignoring a lot more than just generated fortran, but... I don't understand the point of this, it should only ever be generated inside a configured builddir in "test cases/".

But meson already creates a gitignore inside every configured builddir which ignores "*" for precisely the reason of ignoring ALL generated anything whatsoever (and not just fortran).

HaoZeke:
> > Changes `.gitignore` to ignore generated `fortran`
> 
> It is ignoring a lot more than just generated fortran, but... I don't understand the point of this, it should only ever be generated inside a configured builddir in "test cases/".
> 
> But meson already creates a gitignore inside every configured builddir which ignores "*" for precisely the reason of ignoring ALL generated anything whatsoever (and not just fortran).

Ah, true, somehow I had some `.mod` files sneaking around but they weren't part of `meson`.  Sorry, removed that commit.

eli-schwartz:
Alright, let's see what CI thinks.

(From a general meson sense, this PR looks fine to me. However, I have no knowledge of fortran programming, so I can't comment on what constitutes a standard fortran program.)

Issue #9247, opened 2021-09-12
vs backend doesn't handle alias targets


anarazel:
Hi,

Using meson 0.59.0-318-g214d03568 I tried to build my WIP postgres-meson branch with visual studio. Despites lots of other issues that failed because of the presence of an alias_target().

A simple repro follows:

meson.build:
```
project('repro', ['c'],
	version : '0.1',
	meson_version: '>=0.59.99')

sometarget = custom_target('bleat', command: ['cmd', '/s', '/c', 'echo', 'foo'], capture: true, output: ['bleat'])
alias_target('noisy', [sometarget])
```

invocation and failure:
```
C:\Users\anfreund\src\meson-vs-repro>python C:\Users\anfreund\src\meson\meson.py setup --backend vs2019 --wipe build-x64
The Meson build system
Version: 0.59.99
Source dir: C:\Users\anfreund\src\meson-vs-repro
Build dir: C:\Users\anfreund\src\meson-vs-repro\build-x64
Build type: native build
Project name: repro
Project version: 0.1
C compiler for the host machine: cl (msvc 19.29.30133 "Microsoft (R) C/C++ Optimizing Compiler Version 19.29.30133 for x64")
C linker for the host machine: link link 14.29.30133.0
Host machine cpu family: x86_64
Host machine cpu: x86_64
Program cmd found: YES (C:\Windows\system32\cmd.EXE)
Build targets in project: 2

repro 0.1

  User defined options
    backend: vs2019

Traceback (most recent call last):
  File "C:\Users\anfreund\src\meson\mesonbuild\mesonmain.py", line 228, in run
    return options.run_func(options)
  File "C:\Users\anfreund\src\meson\mesonbuild\msetup.py", line 290, in run
    app.generate()
  File "C:\Users\anfreund\src\meson\mesonbuild\msetup.py", line 181, in generate
    self._generate(env)
  File "C:\Users\anfreund\src\meson\mesonbuild\msetup.py", line 243, in _generate
    intr.backend.generate()
  File "C:\Users\anfreund\src\meson\mesonbuild\backend\vs2010backend.py", line 212, in generate
    projlist = self.generate_projects()
  File "C:\Users\anfreund\src\meson\mesonbuild\backend\vs2010backend.py", line 447, in generate_projects
    self.gen_vcxproj(target, str(projfile_path), proj_uuid)
[...]

PR #9253, opened 2021-09-13
backends/vs: Do not emit dummy command for alias_command().

anarazel:
Alias commands did not work with the vs backend, due to trying to access
target.command[0] with an empty command. Fix this by just not emitting a
CustomBuild node for alias targets - the project references are enough to
trigger the necessary actions.

Fixes: #9247

### DIFF
--- a/mesonbuild/backend/vs2010backend.py
+++ b/mesonbuild/backend/vs2010backend.py
@@ -27,7 +27,7 @@ from .. import mlog
 from .. import compilers
 from ..interpreter import Interpreter
 from ..mesonlib import (
-    File, MesonException, python_command, replace_if_different, OptionKey, version_compare, MachineChoice
+    File, MesonException, replace_if_different, OptionKey, version_compare, MachineChoice
 )
 from ..environment import Environment, build_filename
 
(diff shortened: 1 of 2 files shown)


mesonbuild/backend/vs2010backend.py
```diff
@@ -550,19 +550,26 @@ def create_basic_crap(self, target, guid):
...
+            # enough to emit the references to the other projects for them to
+            # be built/run/..., if necessary.
+            assert(isinstance(target, build.AliasTarget))
+            assert(len(depend_files) == 0)
         else:
+            assert(not isinstance(target, build.AliasTarget))
+
             _, _, cmd_raw = self.eval_custom_target_command(target)
-        depend_files = self.get_custom_target_depend_files(target)
-        target_env = self.get_run_target_env(target)
```
xclaesse, line 560: I think you removed that line by mistake, but CI will hopefully tell :)

anarazel: Yea, sorry for that. I accidentally fixed it in 0.59 because I was bisecting and forgot a bisect reset. And then when rebasing I broke it :(

xclaesse:
Great, you even unskipped the unit test, good job! LGTM.

PR #9255, opened 2021-09-13
Argument against source file list specified as wild cards does not stand

johan-boule:
I'm sorry, but the argument that says that dynamically looking up for 10k files is not achievable in a fraction of a second doesn't stand for a second.
First, looking up for directory changes means you only need to stat the directories, i.e. you don't need to read the directories!
Second, since you are already stating the 10k files for changes anyway, it already proves that this amount of stats is not problematic.
I've done an implementation in a python-based build system circa 2009 and it took about 0.6s for 20k files.

### DIFF
--- a/docs/markdown/FAQ.md
+++ b/docs/markdown/FAQ.md
@@ -61,23 +61,16 @@ Instead of specifying files explicitly, people seem to want to do this:
 executable('myprog', sources : '*.cpp') # This does NOT work!
 ```
 
-Meson does not support this syntax and the reason for this is simple.
-This can not be made both reliable and fast. By reliable we mean that
-if the user adds a new source file to the subdirectory, Meson should
-detect that and make it part of the build automatically.
+Meson does not support this syntax.
 
 One of the main requirements of Meson is that it must be fast. This
 means that a no-op build in a tree of 10 000 source files must take no
-more than a fraction of a second. This is only possible because Meson
-knows the exact list of files to check. If any target is specified as
-a wildcard glob, this is no longer possible. Meson would need to
+more than a fraction of a second. Meson would need to
 re-evaluate the glob every time and compare the list of files produced
-against the previous list. This means inspecting the entire source
-tree (because the glob pattern could be `src/\*/\*/\*/\*.cpp` or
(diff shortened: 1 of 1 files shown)


eli-schwartz:
You aren't even talking about the same concept, you are trying to compare stat'ing known files to check if they are out of date (the absolute minimum which any build system must, ultimately, do no matter what, and an unavoidable cost) to doing a full tree glob match.

Stat'ing directories can only help in the optimistic case to determine whether you can completely skip checking for new files matching the glob. Many things that aren't new source files can change those mtimes, and then you are back to where you began, having to list the contents of lots of directories and match very broad patterns against them.

It also completely discounts the work done during calculation. Performing filesystem access is not the only thing that can take time. "I've done an implementation" is quite sparse on details...

It also completely discounts the fact that trying to check at all, even just to stat directories and discover that they haven't changed, would mean leaving ninja and entering meson.py in order to load the list of glob patterns and perform all these checks. Loading up a python program on every ninja invocation just to check if the build system is out of date, is most certainly actual work that is comparatively slow. A normal build does not invoke meson at all, because meson is written in python and relatively slow but that doesn't matter if it is only run once during configuration.

Also, fun fact: it means ninja cannot report "nothing to do" since it instead reports one build edge occurred: the buildsystem out-of-date check.

I don't see any reason to change the FAQ at this time.

johan-boule:
Same concept,
A directory is merely a file that contains a list of names and their associated inodes.
So say many things would possibly change a directory timestamp, but what are those ?
Checking whether a tree glob needs to be redone only involves stating directories and comparing their new timestamps against the old timestamps.
You can record its timestamp in a persistent db accross build runs.
If the timestamp has not changed, it means there's no new, renamed or deleted files.
The implementation i was talking about is there: https://github.com/psycledelics/wonderbuild/blob/master/wonderbuild/filesystem.py.
Benchmarks are there: https://psycledelics.github.io/wonderbuild/benchmarks/time.xml
IMO the presentation of meson's design shouldn't be tied to its current implementation with a Ninja backend. The average user doesn't care that meson uses Ninja. And in ten years you may just use something else.

PR #9256, opened 2021-09-14
flake8 fixes

eli-schwartz:
Apply flake8 fixes for unused imports and missing imports, mostly.

### DIFF
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -34,14 +34,14 @@ import typing as T
 
 if T.TYPE_CHECKING:
     from . import dependencies
-    from .compilers.compilers import Compiler
+    from .compilers.compilers import Compiler, CompileResult
     from .environment import Environment
     from .mesonlib import OptionOverrideProxy, FileOrString
     from .cmake.traceparser import CMakeCacheEntry
 
     OptionDictType = T.Union[T.Dict[str, 'UserOption[T.Any]'], OptionOverrideProxy]
     KeyedOptionDictType = T.Union[T.Dict['OptionKey', 'UserOption[T.Any]'], OptionOverrideProxy]
-    CompilerCheckCacheKey = T.Tuple[T.Tuple[str, ...], str, 'FileOrString', T.Tuple[str, ...], str]
+    CompilerCheckCacheKey = T.Tuple[T.Tuple[str, ...], str, FileOrString, T.Tuple[str, ...], str]
 
 version = '0.59.99'
 backendlist = ['ninja', 'vs', 'vs2010', 'vs2012', 'vs2013', 'vs2015', 'vs2017', 'vs2019', 'xcode']
@@ -267,7 +267,7 @@ class UserFeatureOption(UserComboOption):
         return self.value == 'auto'
 
 if T.TYPE_CHECKING:
-    from .dependencies.detect import TV_DepIDEntry, TV_DepID
+    from .dependencies.detect import TV_DepID
 
 
 class DependencyCacheType(enum.Enum):
@@ -442,7 +442,7 @@ class CoreData:
             DependencyCache(self.options, MachineChoice.BUILD),
             DependencyCache(self.options, MachineChoice.HOST))
 
-        self.compiler_check_cache = OrderedDict()  # type: T.Dict[CompilerCheckCacheKey, compiler.CompileResult]
+        self.compiler_check_cache: T.Dict['CompilerCheckCacheKey', 'CompileResult'] = OrderedDict()
 
         # CMake cache
         self.cmake_cache: PerMachine[CMakeStateCache] = PerMachine(CMakeStateCache(), CMakeStateCache())
--- a/mesonbuild/interpreterbase/interpreterbase.py
+++ b/mesonbuild/interpreterbase/interpreterbase.py
@@ -54,7 +54,9 @@ import textwrap
 from functools import wraps
 
 if T.TYPE_CHECKING:
-    from ..interpreter import Interpreter
+    # T.cast is not handled by flake8 to detect quoted annotation use
+    # see https://github.com/PyCQA/pyflakes/pull/632
+    from ..interpreter import Interpreter # noqa
 
 HolderMapType = T.Dict[
     T.Union[
--- a/mesonbuild/mlog.py
+++ b/mesonbuild/mlog.py
@@ -22,7 +22,9 @@ from contextlib import contextmanager
 from pathlib import Path
 
 if T.TYPE_CHECKING:
-    from ._typing import StringProtocol, SizedStringProtocol
+    # T.Union is not handled by flake8 to detect quoted annotation use (StringProtocol)
+    # see https://github.com/PyCQA/pyflakes/pull/632
+    from ._typing import StringProtocol, SizedStringProtocol # noqa: F401
 
 """This is (mostly) a standalone module used to write logging
 information about Meson runs. Some output goes to screen,
--- a/.flake8
+++ b/.flake8
@@ -28,4 +28,6 @@ ignore =
     W504,
     # A003: builtin class attribute
     A003
+per-file-ignores =
+    mesonbuild/mesonlib/__init__.py:F401
 max-line-length = 120
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -53,7 +53,7 @@ from ..mesonmain import need_setup_vsenv
 
 if T.TYPE_CHECKING:
     from .._typing import ImmutableListProtocol
-    from ..linkers import StaticLinker
+    from ..linkers import DynamicLinker, StaticLinker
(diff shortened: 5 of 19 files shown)


mesonbuild/coredata.py
```diff
@@ -442,7 +442,7 @@ def __init__(self, options: argparse.Namespace, scratch_dir: str, meson_command:
             DependencyCache(self.options, MachineChoice.BUILD),
             DependencyCache(self.options, MachineChoice.HOST))
 
-        self.compiler_check_cache = OrderedDict()  # type: T.Dict[CompilerCheckCacheKey, compiler.CompileResult]
+        self.compiler_check_cache: T.Dict['CompilerCheckCacheKey', 'CompileResult'] = OrderedDict()
```
eli-schwartz, line 445: This file goes down from 85 mypy errors to 84. Why didn't I just delete the broken stuff, lol...

mesonbuild/interpreterbase/interpreterbase.py
```diff
@@ -50,6 +50,9 @@
 
 import os, copy, re, pathlib
 import typing as T
+# T.cast is not handled by flake8 to detect quoted annotation use
+# see https://github.com/PyCQA/pyflakes/pull/632
+import typing
```
mensinda, line 55: Would it be possible to ignore the errors resulting from this instead until the linked PR is merged? I don't think that it is worth it to have two `import typing` statements just to make flake8 happy.

mesonbuild/mlog.py
```diff
@@ -18,6 +18,9 @@
 import time
 import platform
 import typing as T
+# T.Union is not handled by flake8 to detect quoted annotation use
+# see https://github.com/PyCQA/pyflakes/pull/632
+import typing
```
mensinda, line 23: Same point as above...

mensinda (commented):
Overall really good fixes, with only one minor issue.

Do we actually run flake8 currently in the CI? If not this should also be changed in this PR.

mesonbuild/interpreterbase/interpreterbase.py
eli-schwartz: It *could* be ignored, although then if the T.cast changed flake8 would not say anything.

eli-schwartz:
I have no idea what the flake8 story is or when it would ever have been run. This is the kind of thing you want to do via comments like the lgtm run, though, and only fail if the PR adds additional flake8 issues. So I don't think we just want to run `flake8` in a lint.yml

mesonbuild/interpreterbase/interpreterbase.py
mensinda: I was more thinking about something [like this](https://flake8.pycqa.org/en/3.1.1/user/ignoring-errors.html#in-line-ignoring-errors).

eli-schwartz: That's what I meant too. If the T.cast stopped using the noqa import, flake8 would not notice and say "hey, Interpreter is unused".

mensinda:
I am also not sure how the flake8 stuff used to be handled but fair enough. However, if we don't check it we have to constantly make PRs like this. I get what you mean though after running `flake8` myself...

mesonbuild/interpreterbase/interpreterbase.py
mensinda: Even if you explicitly specify the one error we want to get rid of with `# noqa: E<xyz>` ?

dcbaker: for the record, pylint has the same issue: https://github.com/PyCQA/pylint/issues/3846. using `T` was a compromise between those of us who wanted the typing stuff in a namespace, and those who wanted it in not a namespace. But maybe the right solution is move to PEP585 annotations as forward annoations, then you could write:
```python
my_dict: `'dict[str, list[int]'`, instead of `typing.Dict[str, typing.List[int]]`, yes it means using strings, but at least we can vastly reduce the need for the typing module. Heck, we 3.9 we get a literal union syntax, `int | str` instead of `Union[int, str]`, which is pretty nice

Comment (eli-schwartz) [reply mesonbuild/interpreterbase/interpreterbase.py:55]: flake8 currently thinks Interpreter is unused. You can use noqa to say "no, darn it, always pretend it is used". But then if you actually do stop using it, the error is still specified to be gotten rid of.

Currently:

```python
import typing as T

if T.TYPE_CHECKING:
    from ..interpreter import Interpreter # used, but flake8 thinks it is unused and warns you

T.cast('Interpreter', self) # flake8 does not know this is the typing module and doesn't interpret type annotations
```

Proposed:
```python
import typing as T
import typing

if T.TYPE_CHECKING:
    from ..interpreter import Interpreter # used, and flake8 agrees

typing.cast('Interpreter', self) # flake8 sees a type annotation here
```

You want:

```python
from blah import BetterInterpreter
import typing as T

if T.TYPE_CHECKING:
    from ..interpreter import Interpreter # noqa: F401

T.cast('BetterInterpreter', self)

class Thing(BetterInterpreter):
    ...
```

Due to misc rewrites, BetterInterpreter is used both as an annotation and at runtime, and Interpreter isn't used at all. But the noqa is still there...

tl;dr I have a mild dislike of ignoring errors.

dcbaker (commented):
Apart from the typing changes this looks good.

mesonbuild/coredata.py
```diff
@@ -34,14 +34,14 @@
...
-    from .compilers.compilers import Compiler
+    from .compilers.compilers import Compiler, CompileResult
     from .environment import Environment
     from .mesonlib import OptionOverrideProxy, FileOrString
     from .cmake.traceparser import CMakeCacheEntry
 
     OptionDictType = T.Union[T.Dict[str, 'UserOption[T.Any]'], OptionOverrideProxy]
     KeyedOptionDictType = T.Union[T.Dict['OptionKey', 'UserOption[T.Any]'], OptionOverrideProxy]
-    CompilerCheckCacheKey = T.Tuple[T.Tuple[str, ...], str, 'FileOrString', T.Tuple[str, ...], str]
+    CompilerCheckCacheKey = T.Tuple[T.Tuple[str, ...], str, FileOrString, T.Tuple[str, ...], str]
```
eli-schwartz, line 44: Incidentally, this FileOrString is only imported in T.TYPE_CHECKING, and it is also only *used* in T.TYPE_CHECKING. So I figured there was zero point in quoting it, and then flake8 detected it as used, which it is. The other cases were a bit harder since they weren't already inside T.TYPE_CHECKING...

mesonbuild/interpreterbase/interpreterbase.py
mensinda: @eli-schwartz I get your point, but I don't think that the risk is that great here, especially since all the code inside the `if T.TYPE_CHECKING:` is never actually run and just there for mypy.

@dcbaker I would wait with the `dict[str, list[str]]` syntax until we can use `from __future__ import annotations` in 3.7, then we don't have to use strings anymore.

mensinda (approved):
LGTM

mesonbuild/interpreterbase/interpreterbase.py
eli-schwartz: I don't think the risk of using `typing` is great either, but I concede and have went and used noqa.

And yes, I'm eager to use `__future__.annotations` too. :D

Sadly it has been delayed as the default mode (which it was supposed to be in 3.10) until at least 3.11, and it's not clear when it may become the default at all... :(

eli-schwartz:
```
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\concurrent\futures\process.py", line 239, in _process_worker
    r = call_item.fn(*call_item.args, **call_item.kwargs)
  File "D:\a\1\s\run_project_tests.py", line 628, in run_test
  File "run_project_tests.py", line 1567, in <module>
    res = run_tests(all_tests, 'meson-test-run', options.failfast, options.extra_args, options.use_tmpdir, options.num_workers)
  File "run_project_tests.py", line 1114, in run_tests
    return _run_tests(all_tests, log_name_base, failfast, extra_args, use_tmp, num_workers, lf)
  File "run_project_tests.py", line 1251, in _run_tests
    result = f.result
  File "run_project_tests.py", line 1141, in result
    return self.future.result() if self.future else None
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\concurrent\futures\_base.py", line 435, in result
    return self.__get_result()
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\concurrent\futures\_base.py", line 384, in __get_result
    raise self._exception
PermissionError: [Errno 13] The process cannot access the file because it is being used by another process: 'D:\\a\\1\\s\\b f7fdaed927\\prog2.exe'
```

I'm calling this a successful run as it's clearly transient and passed before...

eli-schwartz:
(Sure enough, it passed once merged, and then the cygwin job failed instead with "device busy". How do people even use Windows???)

PR #9257, opened 2021-09-14
setup, configure: Add -C option for consistency

xclaesse:
compile and install commands take the builddir with -C option, for
consistency it is better to also allow it for setup and configure
commands.

It is also less ambigous for the setup command that was allowing to swap
builddir and sourcedir when passed as positional arguments. When using
-C it must be the builddir and only one positional argument is allowed
and must be the sourcedir.


eli-schwartz:
One could reasonably argue that since setup fundamentally operates on the sourcedir, the -C argument which by convention across many tools simply means "chdir into this directory and operate from there" should be the source directory, not the build directory.

It is also a completely bizarre functionality from that perspective, to have "chdir into the builddir but it doesn't actually exist yet" as an option. It feels discordant to my sensibilities.

Anyway given that both of these commands fundamentally operate on directories (unlike test or compile which operate on targets) I'm wondering what particular value a -C option is.

I don't acknowledge "it is less ambiguous" because there is nothing ambiguous to begin with. A directory with a meson.build is only a sourcedir, and an empty directory or one with meson-private/* is only a builddir.

We will just end up now with two ways to do the same thing. Maybe three ways depending on how you count it.

xclaesse:
@eli-schwartz for "meson configure" it makes perfect sense, it operates only on the build directory, just like "install" and "compile". That leaves only "setup" being the weird one, I agree. But it also feels wrong that every commands accepts -C for build directory except that one...

xclaesse:
"dist", "test" and "devenv" also accept -C. But not "introspect" which should be converted too because that one also operate only on build dir.

xclaesse:
> I don't acknowledge "it is less ambiguous" because there is nothing ambiguous to begin with. A directory with a meson.build is only a sourcedir, and an empty directory or one with meson-private/* is only a builddir.

It is not ambiguous from Meson's POV, but it is from user's POV. I don't think it's even documented that you can swap them, --help says it's `[builddir] [sourcedir]`.

eli-schwartz:
It's actually historically pretty consistent. For all commands `meson $cmd` that are an enhanced version of `ninja $cmd`, that command is a meson implementation of some concept which classically, build systems consider part of the "make" concept, not the "./configure" concept.

For all such commands, there is also `ninja -C builddir/ $cmd`, and thus also `meson $cmd -C builddir/`.

For the rest, I guess... devenv is inconsistent in not mapping to ninja? But it also ***would*** be a classical concept associated with "make", if other build systems ever implemented a simpler concept. I think it's fine to not have a `ninja devenv` just for pedantry.

setup and configure are firmly in the realm of a "./configure" concept, which means they fundamentally operate on the unit of a directory and currently associate that with the top-level concept of importance also known as "positional arguments".

xclaesse:
IMHO we should not think of -C only as chdir, we actually do chdir only in install command and I'm not sure it's even needed. We can just simply think of -C as being the build dir, consistently in all our commands. That makes everything much simpler.

eli-schwartz:
How do we not do chdir in test? It:
- runs ninja
- executes a bunch of test binaries the same way `ninja -C builddir/ test` does, which I hope includes a consistent workdir

As for compile, literally the only thing this (useless subcommand) does is immediately invoke ninja with all arguments other than `--*-args` forwarded to ninja. This command more than any other is treating -C as chdir.

And the -C argument is directly ripped off of make/ninja as an imitation of ninja, for no reason more or less than it corresponds to running the corresponding ninja targets with ninja -C.

I really believe these semantics are important, and dislike the idea of `meson setup -C builddir/`. I am -1 on this part of the PR.

I am -0 on adding it to `meson configure` as I don't see much point, but I don't see much harm either.

eli-schwartz:
> We can just simply think of -C as being the build dir, consistently in all our commands. 

What should

```
cd ~
meson configure git/myorg/myproject/
```
do? Say that -C denotes a builddir and throw an error because you used it on a sourcedir?

xclaesse:
> Say that -C denotes a builddir and throw an error because you used it on a sourcedir?

I forgot that `meson configure sourcedir` is allowed, is it even documented? --help message clearly says positional argument must be the build directory... IMHO yes, `-C sourcedir` should be an error.

eli-schwartz:
`meson configure` on an unconfigured source directory being both allowed and showing you the available core+project options, is such an important feature to me that it's even more important than "has a sensible DSL" on my personal list of reasons why meson is better than than cmake. I'm not willing to go back to the dark ages of build systems by removing this intentional feature. :D

xclaesse:
Of course not removing it, `meson configure sourcedir` should print available options, `meson configure -C builddir` should print configured options. Makes it less ambiguous again ;-)

xclaesse:
> IMHO yes, `-C sourcedir` should be an error.

Note this PR makes it an error for `meson setup -C sourcedir` too, that would be consistent.

>     * executes a bunch of test binaries the same way `ninja -C builddir/ test` does, which I hope includes a consistent workdir

ninja does not run tests, meson does. Indeed mtest.py weirdly play with os.chdir() probably because it wants to shoot in its feets, but in the end it sets `cwd=self.test.workdir` on the subprocess anyway. So the -C option is really just a way to give the builddir, nothing else.

xclaesse:
Updated to port introspection command too, and update `--help` messages to indicate positional arguments can be either build or source directories.

mesonbuild/coredata.py
```diff
@@ -1069,6 +1069,33 @@ def parse_cmd_line_options(args: argparse.Namespace) -> None:
             args.cmd_line_options[key] = value
             delattr(args, name)
 
+def validate_dirs(builddir: T.Optional[str], builddir_or_sourcedir: T.Optional[str]) -> T.Tuple[T.Optional[Path], T.Optional[Path]]:
+    if builddir is not None:
+        if builddir_or_sourcedir is not None:
+            raise MesonException('builddir defined with -C argument and positional argument')
```
bonzini, line 1075: Perhaps "directory specified with both -C and a positional argument".

mesonbuild/msetup.py
```diff
@@ -111,7 +116,11 @@ def has_build_file(self, dirname: str) -> bool:
         fname = os.path.join(dirname, environment.build_filename)
         return os.path.exists(fname)
 
-    def validate_core_dirs(self, dir1: str, dir2: str) -> T.Tuple[str, str]:
+    def validate_core_dirs(self, dir1: str, dir2: str, builddir: str) -> T.Tuple[str, str]:
+        if builddir is not None:
+            if dir2 is not None:
+                raise MesonException('builddir defined with -C argument and positional argument')
```
bonzini, line 122: Simpler: "when -C is specified, only one positional argument is specified".

mesonbuild/coredata.py
```diff
@@ -1069,6 +1069,33 @@ def parse_cmd_line_options(args: argparse.Namespace) -> None:
...
+        build_or_source = '.'
+    else:
+        build_or_source = builddir_or_sourcedir
+
+    build_or_source = Path(build_or_source).resolve()
+    if Path(build_or_source, 'meson-private').exists():
+        builddir = build_or_source
+        sourcedir = None
+    elif builddir is not None:
+        raise MesonException(f'{builddir} is not a Meson build directory.')
```
bonzini, line 1087: It's a pity that "Please specify a valid build dir or change the working directory to it." is lost.

bonzini (changes requested):

mesonbuild/coredata.py
xclaesse: ok, updated.

mesonbuild/msetup.py
xclaesse: ok, updated.

mesonbuild/coredata.py
xclaesse: That message was also slightly wrong because you could point to a source directory too. I updated the message now.

keszybz:
`We can just simply think of -C as being the build dir, consistently in all our commands.`

+1. I generally don't write long commands from scratch, but usually recall something like `meson compile -C build && meson test -C build && sudo meson install -C build --quiet --no-rebuild` from shell history. But when I have to edit this to add a `meson setup` step somewhere, half of the times I'll forget to remove `-C` from the command. It doesn't matter what `-C` was originally, now it's just "the build directory".

eli-schwartz:
This is functionally identical in the success case (rather than the pointless error case) to adding `-C` as an option, not an option-argument, and continuing to accept builddir and sourcedir as operands.

"Ignore this flag for the sake of muscle memory compatibility" is not quite what is being proposed here...

Also I don't understand the difficulty, but to be fair I do usually cd into the builddir, or run `ninja -C` rather than meson. I object to the `meson compile` subcommand anyway...

eli-schwartz:
> Renaming `-C` to `-B` everywhere would make the most sense semantically, but I think practically it is much easier to keep -C.

I'm not sure how much more emphatic I can get, beyond "the CLI interface for cmake, but most especially the use of -B and other nonsense, is the single thing I hate most about cmake, ***even more than the string-based CMakeLists.txt syntax".

Arguing to change the meson CLI calling convention because "cmake does it" is the very best thing you can do if you want me to disagree, because I absolutely disagree.

xclaesse:
I'm not doing it because of cmake, I'm doing it because of you. I strongly prefer accepting -C everywhere.

xclaesse:
> -C is poorly defined?

Yes, it is poorly defined. "cd into this directory first"... what does that even mean? reading file there? writing files there? workdir for
subprocesses? The answer is no to all of that. It is so poorly defined that we have "meson init" that uses it for source dir, the opposite of every other commands.

eli-schwartz:
> Yes, it is poorly defined. "cd into this directory first"... what does that even mean? reading file there? writing files there? workdir for subprocesses? The answer is no to all of that.

It means literally `os.chdir(opts.wd)` first thing or else `subprocess.run(the_thing, cwd=opts.wd)`:
- minit
- mcompile
- mdevenv
- minstall
- mtest

It means "load a couple of files from that directory, parse/modify them, and exit":
- mconf
- mintro

It means "don't literally change directories, but in all cases where actions would be taken based on the current directory, build the path relative to this other directory instead":
- msetup
- mdist (actually constructs even more private directories inside the -C directory, and then runs dist steps in there)

Keeping in mind that msetup actually operates on a *recursive* source tree, so most actions it performs (like running commands in run_command) will operate on the current source `subdir()`

> It is so poorly defined that we have "meson init" that uses it for source dir, the opposite of every other commands.

No, we have `meson init` that uses it for "operate as though it were run in that directory", which is exactly consistent with every other command. Because "meson init" is a source-creating command, and other commands are build-queueing commands.

xclaesse:
See, different meaning for all commands, thanks for proving my point.

Issue #1799, opened 2017-05-14
VS and Xcode backends ignore link_depends


jon-turney:
VS backend doesn't generate a dependency on the targets in link_depends.

See #1798 for a test

(This problem is not detected by the tests added in #1172 as they are only run on linuxlike platforms)

nirbheek:
This is because `link_depends:` has always been used for static files. You must change this in order to implement #1792.

jon-turney:
I should have referenced the PR #1651 rather than bug #1172.

> This is because link_depends: has always been used for static files. You must change this in order to implement #1792.

So no, currently meson claims that link_depends can take a custom_target, and that doesn't work correctly with a VS backend.

Sorry, but I will not be contributing fixes to the VS backend.

nirbheek:
Ordinarily, we require that changes must work across both the Ninja and VS backends, but in this case it's a preexisting bug, and it's better to fix it on just the Ninja backend than not at all.

jon-turney:
I don't actually see any sign that link_depends is used inthe xcode backend, either.  I'm not sure if that's tested in CI, currently.

drsog:
Should have label macos added?

TheQwertiest:
IMO, the meson reference documentation should be amended to indicate that `link_depends` currently works *only* with ninja back-end, since this bug is really old and there is no indication that it will be fixed any time soon.

TheQwertiest:
PS: I don't know how and why my comment *unassigned* nioncode :\

nirbheek:
> IMO, the meson reference documentation should be amended to indicate that `link_depends` currently works _only_ with ninja back-end,

This is a terrible idea. Then no one will come around to fix it. Documenting a bug as expected behaviour is the easiest way to make sure that it will never get fixed.

TheQwertiest:
Hm.. Then what about adding smth like `the implementation is currently buggy on VS and XCode (link to issue #1799)`?

nirbheek:
How about we just fix it? ;)

TheQwertiest:
That would be the ideal solution, but this is a two year old bug, so forgive me for being doubtful... :\ (not that I'm blaming anybody, there are priorities and all that)

jon-turney:
Perhaps we should just document the reality that these backends are unmaintained and liable to be removed if someone doesn't step up...

nioncode:
I don't know about the XCode backend, but I implemented many things for the VS backends just a few weeks ago.

[...]

PR #9259, opened 2021-09-15
backend/vs: process link dependencies.

anarazel:
Needs to be split in separate commits.  Might make sense to refactor get_target_deps() first.

Fixes: #1799

### DIFF
--- a/mesonbuild/backend/vs2010backend.py
+++ b/mesonbuild/backend/vs2010backend.py
@@ -280,6 +280,16 @@ class Vs2010Backend(backends.Backend):
                         all_deps[ldep.get_id()] = ldep.target
                     else:
                         all_deps[ldep.get_id()] = ldep
+
+                for ldep in target.link_depends:
+                    if isinstance(ldep, build.CustomTargetIndex):
+                        all_deps[ldep.get_id()] = ldep.target
+                    elif isinstance(ldep, File):
+                        # Already built, no target references needed
+                        pass
+                    else:
+                        all_deps[ldep.get_id()] = ldep
+
                 for obj_id, objdep in self.get_obj_target_deps(target.objects):
                     all_deps[obj_id] = objdep
             else:
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -870,8 +870,7 @@ class BuildTarget(Target):
                 self.link_depends.append(
                     File.from_source_file(environment.source_dir, self.subdir, s))
             elif hasattr(s, 'get_outputs'):
-                self.link_depends.extend(
-                    [File.from_built_file(s.get_subdir(), p) for p in s.get_outputs()])
+                self.link_depends.append(s)
             else:
                 raise InvalidArguments(
                     'Link_depends arguments must be strings, Files, '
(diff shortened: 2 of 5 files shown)


mesonbuild/backend/vs2010backend.py
```diff
@@ -280,6 +280,16 @@ def get_target_deps(self, t, recursive=False):
                         all_deps[ldep.get_id()] = ldep.target
                     else:
                         all_deps[ldep.get_id()] = ldep
+
+                for ldep in target.link_depends:
+                    if isinstance(ldep, build.CustomTargetIndex):
+                        all_deps[ldep.get_id()] = ldep.target
+                    elif isinstance(ldep, File):
+                        assert not ldep.is_built
+                        pass
```
dcbaker, line 289: You don't need `pass` here, pass is only necessary for cases where a body would be empty otherwise.

But more to the point, why are we asserting not built here?

anarazel: I unfortunately haven't gotten the test pass on windows (nor all on linux, w/ debian unstable), even without any modification. Not quite sure what the issue is. Hence doing part of the debugging via CI :(.

> Also, why are we asserting not built here?

Well, with msbuild we can't (or at least we don't know how yet, it's probably possible *somehow*) have dependencies on files, we need the targets.  And before the changes to build.py to process_link_depends() and the vs_module_defs handling link_depends always contained files, even when targets were passed, because that's sufficient for ninja. So I tried to change it so that only static files are in link_depends, not ones we might need a dependency for. Hence the assert.

But doing this in the vs backend is a bad idea, not sure what I was thinking, it should of course be somewhere in build.py.

That might have showed me more quickly that we can't make that assertion, because of configure_file(). As the test failing in `test cases/windows/9 vs module defs generated/subdir/meson.build` shows.

I don't yet know how dependencies on configure_file are handled.

anarazel: > I don't yet know how dependencies on configure_file are handled.

Since they presumably are static once meson generated, they shouldn't be needed to handle project dependencies I assume. But depending on what kind of target it's the input to, it might still be needed to trigger rebuilds. I guess msbuild's dependency tracking would do the trick for compile / link tasks, and custom targets shouldn't have link_depends. 

Brr, msbuild, brr.

eli-schwartz: configure_file is processed exclusively during `meson setup` and the input file is added as a reconfigure dependency so edits to the source.in will trigger a full buildsystem reconfigure.

This is usually more or less okay, since the expected use of a configure_file is to configure that file with with a configuration_data() object which only exists in meson.

Is that what you mean?

anarazel:
> configure_file is processed exclusively during meson setup and the input file is added as a reconfigure dependency so edits to the source.in will trigger a full buildsystem reconfigure.

> This is usually more or less okay, since the expected use of a configure_file is to configure that file with with a configuration_data() object which only exists in meson.

> Is that what you mean?

Yes, pretty much. I think as-is they won't reliably trigger rebuilds. But that's a smaller issue than not building at all :). And I think we can relatively add them as additional dependencies, but that might be separate patch.

mesonbuild/build.py
```diff
@@ -2066,20 +2065,18 @@ def process_kwargs(self, kwargs, environment):
...
                 # When passing output of a Custom Target
-                path = File.from_built_file(path.subdir, path.get_filename())
-                self.vs_module_defs = path
-                self.link_depends.append(path)
+                self.vs_module_defs = File.from_built_file(path.subdir, path.get_filename())
             else:
                 raise InvalidArguments(
                     'Shared library vs_module_defs must be either a string, '
                     'a file object or a Custom Target')
+            self.process_link_depends(path, environment)
```
anarazel, line 2078: I was wondering if this instead should be a shared method used by both the vs_module_defs and the link_depends kwargs handling.

anarazel:
The failure in the cross test seems to be a transient one independent of this commit: `wine: a wine server seems to be running, but I cannot connect to it.`

PR #9264, opened 2021-09-15
during executable lookup, do not search PATH if a directory component is given

### LINKED ISSUE #9262: When `add_install_script` is used with `./` (curr.dir), the script may fail to run
**Describe the bug**

When using `meson.add_install_script('./test.sh')`, depending on the directory tree, it may happen that the script could not be run upon running later `ninja install` or `meson install`.

**To Reproduce**

```
 λ cat test.sh
#!/bin/bash
echo hello
 λ ./test.sh               # NOTE: the script itself runs fine
hello
 λ cat meson.build
project(
  'test',
  ['c'],
)
subdir('dir1/dir2/')
 λ cat dir1/dir2/meson.build
meson.add_install_script('./test.sh')
 λ meson build
The Meson build system
Version: 0.59.1
Source dir: /tmp/foo
Build dir: /tmp/foo/build
Build type: native build
Project name: test
Project version: undefined
C compiler for the host machine: ccache cc (gcc 11.1.0 "cc (GCC) 11.1.0")
C linker for the host machine: cc ld.bfd 2.36.1
Host machine cpu: x86_64
Program ./test.sh found: YES (./test.sh)
Build targets in project: 0

 λ DESTDIR=/tmp/ ninja -C build install
ninja: Entering directory `build'
[0/1] Installing files.
Running custom install script './test.sh'
FAILED: install script './test.sh' could not be run, stopped
FAILED: meson-install
/usr/bin/meson install --no-rebuild
```

**Expected behavior**

The last command would succeed

**Known Workarounds**

With the directory tree here, if script is used as `meson.add_install_script('../../test.sh')`, then there will be no errors upon installation.

**system parameters**
* Is this a [cross build](https://mesonbuild.com/Cross-compilation.html) or just a plain native build ? Native
* OS: Archlinux
* Python version: 3.9.6
* `meson --version`: 0.59.1
* `ninja --version`: 1.10.2

eli-schwartz:
This will always be wrong, because when a directory component is provided we need to match an exact filename on a manual search path, for example find_program with dirs: or the current meson.build subdir.

If we ever get this far, shutil.which will do the same "is there a dirname, if so just check whether the filename exists relative to
cwd"... except that the documented meson lookup path is that we check relative to meson.build subdir, not relative to the cwd, and the cwd could be anything, but is probably the root sourcedir.

Since internally, meson does not actually os.chdir into the sourcedir, it could be absolutely anything at all, though.

...

The actual returned name for shutil.which(name) given a literal pathname with a directory component is "return name" without adding the absolute path, which means that this is double-broken. Not only does it find things we didn't expect, the resulting ExternalProgram object doesn't have the correct path to the program, so it will report "found" and then fail to actually run when the current directory is changed, for example by ninja -C.

Fixes #9262

### DIFF
--- a/mesonbuild/programs.py
+++ b/mesonbuild/programs.py
@@ -295,6 +295,9 @@ class ExternalProgram(mesonlib.HoldableObject):
         commands = self._search_dir(name, search_dir)
         if commands:
             return commands
+        # If there is a directory component, do not look in PATH
+        if os.path.dirname(name) and not os.path.isabs(name):
+            return [None]
         # Do a standard search in PATH
         path = os.environ.get('PATH', None)
         if mesonlib.is_windows() and path:


eli-schwartz:
Sorry, red herring... I thought maybe it would still need to check windows_special_cases but the isabs check is already sufficient there. Reset back to the original commit.

xclaesse:
A unit test would be nice. In allplatformtests.py you can do `self.init(testdir, workdir=testdir)`, put foo.sh at the root of your unit test, and do find_program('./foo.sh') in a subdir, and assert it's not found.

xclaesse:
While at it, it could also check that `foo.sh` (without directory component) is not found in cwd too. shutil.which() seems to work in this case already.

eli-schwartz:
TBH this isn't really something prone to regressing, and also -- this doesn't actually fix a bug, it just makes an error condition move from failing during build to failing during `meson setup`.

I don't think we need to add another test project which has to fork and run meson rather than being, say, folded into `test cases/common/26 find program/` which doesn't work because the cwd you need to test is going to end up being the root of the repository.

PR #9267, opened 2021-09-15
C2000: Added depfile generation for incremental builds

miebka:
While working with meson and the `C2000` toolchain, I noticed that changes in the header file did not result in a recompilation of the corresponding source files. Since the function `get_dependency_gen_args()` was not implemented, this is not a big surprise.


dcbaker (approved):
I can't test it, but it is what the manual seems to say to do, so this looks good to me.

Issue #9258, opened 2021-09-14
Regression for Fortran extension using `custom_target` from depscan input to json


rgommers:
**Describe the bug**

gh-9131 broke the SciPy build. Traceback:
```
[225/1400] Module scanner.
FAILED: scipy/linalg/_fblas.cpython-39-x86_64-linux-gnu.so.p/depscan.dd 
/home/rgommers/anaconda3/envs/scipy-meson/bin/meson --internal depscan scipy/linalg/_fblas.cpython-39-x86_64-linux-gnu.so.p/_fblas.cpython-39-x86_64-linux-gnu.dat scipy/linalg/_fblas.cpython-39-x86_64-linux-gnu.so.p/depscan.dd /home/rgommers/code/bldscipy/build/scipy/linalg/_fblas.cpython-39-x86_64-linux-gnu.so.p/_fblas.cpython-39-x86_64-linux-gnu-deps.json
Traceback (most recent call last):
  File "/home/rgommers/anaconda3/envs/scipy-meson/bin/meson", line 8, in <module>
    sys.exit(main())
  File "/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/mesonbuild/mesonmain.py", line 326, in main
    return run(sys.argv[1:], launcher)
  File "/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/mesonbuild/mesonmain.py", line 314, in run
    return run_script_command(args[1], args[2:])
  File "/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/mesonbuild/mesonmain.py", line 262, in run_script_command
    return module.run(script_args)
  File "/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/mesonbuild/scripts/depscan.py", line 203, in run
    return scanner.scan()
  File "/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/mesonbuild/scripts/depscan.py", line 153, in scan
    self.scan_file(s)
  File "/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/mesonbuild/scripts/depscan.py", line 52, in scan_file
    self.scan_fortran_file(fname)
  File "/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/mesonbuild/scripts/depscan.py", line 61, in scan_fortran_file
[...]

PR #9269, opened 2021-09-16
fix depscan json depends

dcbaker:
After we changed to the depscan tool consuming a json file, we lost the correct ordering dependencies, so we need to add those back.

Fixes: #9258

### DIFF
--- a/test cases/fortran/7 generated/meson.build
+++ b/test cases/fortran/7 generated/meson.build
@@ -9,10 +9,8 @@ conf_data.set('ONE', 1)
 conf_data.set('TWO', 2)
 conf_data.set('THREE', 3)
 
-configure_file(input : 'mod3.fpp', output : 'mod3.f90', configuration : conf_data)
-# Manually build absolute path to source file to test
-# https://github.com/mesonbuild/meson/issues/7265
-three = library('mod3', meson.current_build_dir() / 'mod3.f90')
+mod3_f = configure_file(input : 'mod3.fpp', output : 'mod3.f90', configuration : conf_data)
+three = library('mod3', mod3_f)
 
 templates_basenames = ['mod2', 'mod1']
 generated_sources = []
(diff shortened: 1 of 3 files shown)


dcbaker:
@rgommers, Can you test and see if this fixes your issue?

rgommers:
> @rgommers, Can you test and see if this fixes your issue?

Works like a charm, thanks @dcbaker!

eli-schwartz:
So compared to before 035df5369e34fdb7bda954601800cd00b8d8ef8e, we are demoting the order deps from scan_sources to generated_source_files. Which I *think* means it excludes unity files and files which are sourcedir files? So that should be fine for depscan/dyndep.

test cases/fortran/7 generated/meson.build
```diff
@@ -7,12 +7,22 @@ project('generated', 'fortran',
 conf_data = configuration_data()
 conf_data.set('ONE', 1)
 conf_data.set('TWO', 2)
-conf_data.set('THREE', 3)
 
-configure_file(input : 'mod3.fpp', output : 'mod3.f90', configuration : conf_data)
-# Manually build absolute path to source file to test
-# https://github.com/mesonbuild/meson/issues/7265
-three = library('mod3', meson.current_build_dir() / 'mod3.f90')
+mod3_f = custom_target(
```
eli-schwartz, line 11: Changing this from a configure_file to a custom_target is fine, I'm sure, but... I don't think custom_target does absolute paths, so are we not dropping the coverage for the issue referenced there?

Why are you also changing the library() invocation to use mod3_f instead of `meson.current_build_dir() / 'mod3.f90'`?

dcbaker: `meson.corrent_build_dir()` was a workaround for the mentioned bug, not the bug itself. So the test wasn't actually doing anything useful in this case. I had changed it to do the correct/idiomatic thing in the first commit since that bug was fixed but the test was never updated. I guess we could build two copies of mod3.f90 or something?

eli-schwartz: Well, as long as the original issue got fixed in the meantime I guess. :)

PR #9270, opened 2021-09-16
Pickle binaries from build environment and use them in symbol extractor

ggabriel96:
### Description

This PR is a follow-up from https://github.com/mesonbuild/meson/pull/9222, from which we talked on IRC and decided to pickle the binaries table in a dedicated file (so it's not too big) and re-use that in other scripts. The first one is `symbolextractor.py`. The problem this solves is that, today, these scripts do not respect native or cross files specifications of binaries, only environment variables.

### Changes

1. pickle the binaries table:
    - added a `pickle_save` function to `mesonlib/universal.py`
        I need this piece of code and saw that some places have that exact same snippet, so I created a function that could eventually be reused in other places too
    - save the `binaries` object from the `Environment` in `Build` via the `pickle_save` function mentioned above in a file `meson-private/binaries.dat`, right after pickling the whole `Build` class

1. use the binaries table in `symbolextractor.py`:
    - I decided to implement this with the minimum possible changes to the script, so I added a global `BINARIES` variable that will hold the binaries and is accessible to all functions. I can make another PR later dedicated to a refactor of this script, if that's desired
    - load the binaries once on script "startup" and reuse that via `get_tool`
    - the `get_tool` function still needs to look in `os.environ` because not all of the binaries that `symbolextractor.py` uses exist in the pickled binaries table **if they are specified via environment variables**, e.g. `ar`, `nm`, `readelf`, etc. I don't know if this is temporary or not, but it seems reasonable that Meson doesn't know all the binaries that will eventually be needed
    - I was going to add an `@lru_cache` to `get_tool`, but apparently it is not currently being called more than once per tool per invocation of the script, so that would save nothing

### Notes

1. should the `load_binaries` function be somewhere else so other places can re-use it? Or is it too simple for that? My concern is about not duplicating the `binaries.dat` name everywhere
1. I'm not sure how to solve [this](https://github.com/ggabriel96/meson/blob/a6fd6439da2e7904d7bafa3a635bc1ed6a52ca96/mesonbuild/scripts/symbolextractor.py#L288-L292):

    ```python
    if cross_host is not None:
        # In case of cross builds just always relink. In theory we could
        # determine the correct toolset, but we would need to use the correct
        # `nm`, `readelf`, etc, from the cross info which requires refactoring.
        dummy_syms(outfilename)
    ```

   But I also think this is out of the scope of _this particular PR_.

1. I'm not familiar with Meson tests, so I don't know where to add tests for these changes :sweat_smile:


mesonbuild/msetup.py
```diff
@@ -243,6 +243,7 @@ def _generate(self, env: environment.Environment) -> None:
                 intr.backend.generate()
             b.devenv.append(intr.backend.get_devenv())
             build.save(b, dumpfile)
+            universal.pickle_save(b.environment.binaries, os.path.join(env.get_scratch_dir(), 'binaries.dat'))
```
eli-schwartz, line 246: b.environment.binaries is a meson object. Unpickling this reveals that a bunch of modules are imported in order to restore this state.

I was thinking, if one wanted to avoid loading everything the usual coredata loads, that you'd probably want to save this:

```
binaries_dict = {'host': b.environment.binaries.host.binaries, 'build': b.environment.binaries.build.binaries}
```

eli-schwartz: Although hmm, to be fair there are 70 modules loaded by binaries.dat, and 200 loaded by coredata.dat

ggabriel96: Oh I see... I wasn't aware of that. This is the first time I mess around with the pickle module. Would you mind telling me how you get to count those imports?

If I understand correctly, your suggestion is to basically convert the `b.environment.binaries` object into a plain dictionary, right? So there will be no need for any imports when unpickling it.

eli-schwartz:
```python
>>> import pickle, sys
>>> old = set(sys.modules.keys())
>>> with open('meson-private/binaries.dat', 'rb') as f:
...     data = pickle.load(f)
... 
>>> new = set(sys.modules.keys())
>>> new - old
```

Yes, what I suggested would effectively convert it into a plain dictionary.

ggabriel96: Oh, ok, thanks.

So, I thought about adding a `to_dict` method (or `__iter__` or something) to `PerMachine` so that the keys of the resulting dictionary are standardized and controlled by only one place. But `PerMachine` is a generic class, so that would only work for the things in `build` and `host` that also somehow support that operation. Is that ok? Or I should implement a specific solution for this case?

ggabriel96: @eli-schwartz refactored it to pickle plain `dict`s instead of the `PerMachine` object. I went for an `as_plain_dict` method instead of `__iter__` because I would personally like to use `MachineChoice`s as keys, but then that would not be a plain dictionary and would require `import`s.

eli-schwartz: That seems fine.

ggabriel96:
I see mypy is failing with the following errors:

```
mesonbuild/mesonlib/universal.py:511:22: error: "_T" has no attribute "as_plain_dict"
mesonbuild/mesonlib/universal.py:512:21: error: "_T" has no attribute "as_plain_dict"
```

What would be a good way to solve them? I thought about maybe checking the existence of `as_plain_dict` and reporting a meaningful error if that fails.

I am also still in doubt about how to test this new code. Would these tests be in `unittests/machinefiletests.py`?

eli-schwartz:
The Cygwin failure can be fixed by rebasing onto master.

No idea about the mypy error, perhaps  @dcbaker knows.

ggabriel96:
> No idea about the mypy error, perhaps @dcbaker knows.

The `PerMachine` class is generic, using `_T` as type parameter for its `build` and `host` attributes. Since `_T` can be _any type_, mypy knows it doesn't have the method `as_plain_dict`. So we basically need to handle types that don't have that, because `PerMachine` can be used with them. It's just `as_plain_dict` that cannot. I just wanted to know if Meson already has some rule or guideline to deal with this situation that I should follow.

ggabriel96:
Okay, I _think_ the rebase worked :sweat_smile: 
Will wait for CI :eyes:

eli-schwartz:
> (sorry if I explained something you already know :sweat_smile:, I might have misinterpreted what you said)

I simply mean that I am not the typing expert around here, so I'll defer to the person who both knows a lot about it and actually cares a whole lot more than I do. :D

Issue #9272, opened 2021-09-17
setup for Ninja with Virtual Studio fails if buildpath on different drive than cwd


rkuester:
When running `setup` using the Ninja backend and Visual Studio compiler, an unhandled exception is raised if the build directory is on a different drive than the working directory from which `setup` is run (often the source directory).

This occurs merely in service of formatting a relative path for a log message; the build works fine if the formatting is avoided. When `setup` creates a Visual Studio environment, a message is logged which contains a path to the build directory. Typically, this path is converted to a relative path prior to printing. If the path cannot be converted to a relative path (e.g., because buildpath is on a different drive from the cwd), it fails with an unhandled ValueError exception.

## To reproduce
hello.cpp
```
#include <iostream>

int main(int argc, char* argv[]) {
  std::cout << "Hello, world!\n";
  return 0;
}
```

meson.build
```
project('hello', 'cpp')
executable('hello', 'hello.cpp')
```

 command
```
PS C:\Users\rkues> cd p:\testsrc
PS p:\testsrc> meson c:\Users\rkues\build
Activating VS 16.10.4
The Meson build system
Version: 0.59.99
Source dir: \\callisto\share\testsrc
Build dir: C:\Users\rkues\build
Build type: native build
Project name: hello
Project version: undefined
C++ compiler for the host machine: cl (msvc 19.29.30040 "Microsoft (R) C/C++ Optimizing Compiler Version 19.29.30040 for x64")
C++ linker for the host machine: link link 14.29.30040.0
Host machine cpu family: x86_64
Host machine cpu: x86_64
Build targets in project: 1

Found ninja-1.10.2 at C:\ProgramData\chocolatey\bin\ninja.EXE
Traceback (most recent call last):
  File "C:\Users\rkues\wrk\meson\mesonbuild\mesonmain.py", line 228, in run
    return options.run_func(options)
  File "C:\Users\rkues\wrk\meson\mesonbuild\msetup.py", line 290, in run
    app.generate()
  File "C:\Users\rkues\wrk\meson\mesonbuild\msetup.py", line 181, in generate
    self._generate(env)
  File "C:\Users\rkues\wrk\meson\mesonbuild\msetup.py", line 243, in _generate
    intr.backend.generate()
  File "C:\Users\rkues\wrk\meson\mesonbuild\backend\ninjabackend.py", line 518, in generate
    builddir = builddir.relative_to(Path.cwd())
[...]

PR #9273, opened 2021-09-17
ninjabackend/vs: handle builddir on different drive than cwd

rkuester:
When setup creates a Visual Studio environment, a message is logged
which contains a path to the build directory. Typically, this path is
converted to a relative path prior to printing. If the path cannot be
converted to a relative path (e.g., because buildpath is on a different
drive from the cwd), print out the full path instead of failing with an
unhandled exception.

### DIFF
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -515,7 +515,13 @@ class NinjaBackend(backends.Backend):
         ninja = environment.detect_ninja_command_and_version(log=True)
         if need_setup_vsenv:
             builddir = Path(self.environment.get_build_dir())
-            builddir = builddir.relative_to(Path.cwd())
+            try:
+                # For prettier printing, reduce to a relative path. If
+                # impossible (e.g., because builddir and cwd are on
+                # different Windows drives), skip and use the full path.
+                builddir = builddir.relative_to(Path.cwd())
+            except ValueError:
+                pass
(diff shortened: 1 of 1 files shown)


rkuester:
This fixes #9272.

I'm not sure of the best way (or whether there is a good way) to create a unit test for this issue. Any suggestions?

mesonbuild/backend/ninjabackend.py
```diff
@@ -515,7 +515,10 @@ def generate(self):
         ninja = environment.detect_ninja_command_and_version(log=True)
         if need_setup_vsenv:
             builddir = Path(self.environment.get_build_dir())
-            builddir = builddir.relative_to(Path.cwd())
+            try:
```
dcbaker, line 518: This really deserves a comment, otherwise this looks good to me

dcbaker (approved):

mesonbuild/backend/ninjabackend.py
eli-schwartz: @rkuester What's the status of getting a comment here?

rkuester: Okay, took a stab at a comment. Thanks for the review.

dcbaker: Thanks, I was mostly just concerned about someone well meaning removeing this try/except because "Why are we doing this here?" This looks good to me now, thanks.

dcbaker (approved):

eli-schwartz (approved):

Issue #9275, opened 2021-09-18
Building python packages: adding C extensions with subdir removes sources installed using install_sources


daknuett:
**Describe the bug**
I try to build a python package using `meson` and `mesonpep517`. This is probably the way packages should be built in the future as by PEP517 and PEP518. I have a couple of source files (python modules) that I install using `python_installation.install_sources`. This works fine by itself.
When I add C extension (using `python_installation.extension_module`) either the compiled C extensions or the python modules are missing. This problem might be related to  #6331.

**To Reproduce**
I have built a [toy project showing three cases](https://github.com/daknuett/python_module_with_meson) you can use it to reproduce the problem.

- on branch `master` I set `subdir:''` (because of the comments in #6331). The compiled extension is placed in the wheel's root directory as expected. This is not how `setuptools` would do it; but most importantly the python modules are missing.
- on branch `subdir_as_by_setuptools` the `subdir` is set to match the layout from `setuptools`. Again, the python modules are missing.
- on branch `extension_no_subdir` I removed the `subdir` from `extension_module`. Then the extension is missing in the `whl` but the python modules are present.

In all three cases you can build the wheels using `python3 -m build .` and inspect the resulting `whl` in the `dist/` directory. Installing the wheel or source dists using pip does not yield different results.

**Expected behavior**

Both python modules (`install_sources`) and C extensions (`extension_module`) should be present. Ideally the `subdir` could be chosen automatically to match the layout from `setuptools` but that is just a bonus.

**system parameters**
* Native build. (via `python3 -m build`)
* Fedora 34 Workstation
* Python 3.9.7
* meson 0.59.1
* ninja 1.10.2

eli-schwartz:
I don't see any bug here.

- `master` is behaving as expected, meson installs `site-packages/mymath/test.py` and `site-packages/mymath.mymath.cpython-39-x86_64-linux-gnu.so`
- `extension_no_subdir` is not a useful case to demo, it is merely demonstrating that the issue you linked exists which is not news
- `subdir_as_by_setuptools` is behaving as expected, meson installs `site-packages/mymath/test.py` and `site-packages/mymath/mymath.mymath.cpython-39-x86_64-linux-gnu.so`


meson is behaving perfectly correctly, except in the `extension_no_subdir` case which is an acknowledged and fixed bug that you linked to.

[...]

PR #9276, opened 2021-09-19
Module:Python:: Added a bit documentation for quirky behaviour when chained with mesonpep517

daknuett:
I found this quirk which I thought was a bug(#9275). It should probably be fixed in `mesonpep517` (although I am not sure if it can be fixed there). For now it seems like it should be noted in the docs.


eli-schwartz:
It would be better to submit an issue at https://gitlab.com/thiblahute/mesonpep517/-/issues to fix this (or document it there).

PR #9277, opened 2021-09-19
msubprojects: gracefully handle being offline

eli-schwartz:
This command is useful to e.g. update a cloned subproject which does not have its packagefiles merged, or which has updated packagefiles. This does not strictly require internet, so if we can satisfy the ref, simply log a warning and continue.

This enables a workflow where for network-isolated builds, the subproject is cloned and moved into place and later on the build orchestrator runs `meson subprojects update` without a network interface to initialize the subproject.

While we are at it, if the revision is a commit hash we can immediately tell if we don't even need to *try* finding updates. So short-circuit that.

This ameliorates a workflow concern raised in https://github.com/Plagman/gamescope/issues/61

### DIFF
--- a/mesonbuild/msubprojects.py
+++ b/mesonbuild/msubprojects.py
@@ -248,7 +248,13 @@ class Runner:
             self.log(mlog.red(e.output))
             self.log(mlog.red(str(e)))
             return False
-        if True:
+        if self.wrap_resolver.is_git_full_commit_id(revision) and \
+                quiet_git(['rev-parse', '--verify', revision + '^{commit}'], self.repo_dir)[0]:
+            # The revision we need is both a commit and available. So we do not
+            # need to fetch it because it cannot be updated.  Instead, trick
+            # git into setting FETCH_HEAD just in case, from the local commit.
+            self.git_output(['fetch', '.', revision])
+        else:
             try:
                 # Fetch only the revision we need, this avoids fetching useless branches.
                 # revision can be either a branch, tag or commit id. In all cases we want
--- a/mesonbuild/msubprojects.py
+++ b/mesonbuild/msubprojects.py
@@ -261,9 +261,14 @@ class Runner:
                 self.git_output(['fetch', '--refmap', heads_refmap, '--refmap', tags_refmap, 'origin', revision])
             except GitException as e:
                 self.log('  -> Could not fetch revision', mlog.bold(revision), 'in', mlog.bold(self.repo_dir))
-                self.log(mlog.red(e.output))
-                self.log(mlog.red(str(e)))
-                return False
+                if quiet_git(['rev-parse', revision + '^{commit}'], self.repo_dir)[0]:
+                    self.log(mlog.yellow('WARNING:'), 'Proceeding with locally available copy')
+                    # Trick git into setting FETCH_HEAD from the local revision.
+                    quiet_git(['fetch', '.', revision], self.repo_dir)
+                else:
+                    self.log(mlog.red(e.output))
+                    self.log(mlog.red(str(e)))
+                    return False
 
         if branch == '':
             # We are currently in detached mode
--- a/mesonbuild/msubprojects.py
+++ b/mesonbuild/msubprojects.py
@@ -248,21 +248,22 @@ class Runner:
             self.log(mlog.red(e.output))
             self.log(mlog.red(str(e)))
             return False
-        try:
-            # Fetch only the revision we need, this avoids fetching useless branches.
-            # revision can be either a branch, tag or commit id. In all cases we want
-            # FETCH_HEAD to be set to the desired commit and "git checkout <revision>"
-            # to to either switch to existing/new branch, or detach to tag/commit.
-            # It is more complicated than it first appear, see discussion there:
-            # https://github.com/mesonbuild/meson/pull/7723#discussion_r488816189.
-            heads_refmap = '+refs/heads/*:refs/remotes/origin/*'
-            tags_refmap = '+refs/tags/*:refs/tags/*'
-            self.git_output(['fetch', '--refmap', heads_refmap, '--refmap', tags_refmap, 'origin', revision])
-        except GitException as e:
-            self.log('  -> Could not fetch revision', mlog.bold(revision), 'in', mlog.bold(self.repo_dir))
-            self.log(mlog.red(e.output))
-            self.log(mlog.red(str(e)))
-            return False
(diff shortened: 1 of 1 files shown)


dcbaker (approved):
LGTM

mesonbuild/msubprojects.py
```diff
@@ -248,21 +248,33 @@ def update_git(self):
...
+                # https://github.com/mesonbuild/meson/pull/7723#discussion_r488816189.
+                heads_refmap = '+refs/heads/*:refs/remotes/origin/*'
+                tags_refmap = '+refs/tags/*:refs/tags/*'
+                self.git_output(['fetch', '--refmap', heads_refmap, '--refmap', tags_refmap, 'origin', revision])
+            except GitException as e:
+                self.log('  -> Could not fetch revision', mlog.bold(revision), 'in', mlog.bold(self.repo_dir))
+                if quiet_git(['rev-parse', revision + '^{commit}'], self.repo_dir)[0]:
+                    self.log(mlog.yellow('WARNING:'), 'Proceeding with locally available copy')
+                    # Trick git into setting FETCH_HEAD from the local revision.
+                    quiet_git(['fetch', '.', revision], self.repo_dir)
```
xclaesse, line 273: Does this means that if the branch exists but it fails to fetch latest revision, that will still return success? I'm not sure this is a good idea, because you generally do want your CI script to fail on network issues. One particular case I have in mind is the current trend on renaming `master` branch to `main`, if one of my subproject does that, I want my CI to fail to update `master` branch unless of silently keep using old revision.

I don't really understand the point of calling `meson subprojects update` when you're offline at all, when does it make any sense?

xclaesse:
> This enables a workflow where for network-isolated builds, the subproject is cloned and moved into place and later on the build orchestrator runs `meson subprojects update` without a network interface to initialize the subproject.

Then just call `meson subprojects update` (and/or download) before moving to a network isolated build. Or just ignore the return value of the command instead of failing your CI.

eli-schwartz:
You cannot do that before moving to a network isolated build VM and assembling the source code from a git-level caching layer, e.g. the OBS.

You cannot do that in flathub's security model, because apparently they don't allow running random commands with network access (but do support a manifest of source urls which will be git cloned or retrieved as files).

You cannot do that in various distro ecosystems which act like flathub and may or may not have methods for disabling network isolation but if they do then that enables network access for the entire build, not just for source preparation.

You can, however, do that in other distros where network isolation is not enforced.

This was discussed extensively in the linked issue. The previous behavior was problematic for as many people as it was okay for.

In contrast I don't understand in the slightest why this is a problem for CI, so as far as I'm concerned this PR has no downsides.

mesonbuild/msubprojects.py
eli-schwartz: It makes sense in exactly the situation I described in the first place!

I don't understand why this is problematic for CI, it only succeeds if you side load a copy of the git repo so that the revision is satisfiable.

- It doesn't succeed on the first download of the subproject.
- It doesn't succeed if you eject the cache using subprojects purge.

Why does CI have a partial cache anyway? Its not available when your CI does `git clone` of the superproject sources.

Even if it was there (how???) and the CI has worse internet than my terrible home connection, the next git run will pick up the problem.

Or you could use pinned sources.

eli-schwartz: If someone renames a long-lived branch then there will be lots of problems and nowhere near limited to meson. It will break git pull too, in mysterious and hard to fix ways.

But fully deleting the repo and re-cloning it will at least get ***you*** back to the renamed branch, and... if you use `revision = head` instead of naming the branch, it will fix meson too. So CI won't be affected, and users manually running these commands will have the log messages they need.

xclaesse: @eli-schwartz GStreamer for example cache all subprojects in their docker image to only do an update when running the job. But IMHO it's just plain wrong to return success when it clearly was not a success and even printed a warning that it failed.

xclaesse:
I don't understand why you call `meson subprojects update` at all when you know you have no network access. Do they also expect wget to return success?

xclaesse:
The more I think about this, the more I think it should be reverted. I understand offline build is important, and that's actually the reason I even added `meson subprojects download` in the first place. Y can download subprojects during a preparation stage with network, then do the build completely offline. You're simply using the wrong command.

eli-schwartz:
> I don't understand why you call `meson subprojects update` at all when you know you have no network access. Do they also expect wget to return success?

You do it because you expect the `--reset` operation to return success.

eli-schwartz:
Ugh. After we both re-read the code it seems at no point did it ever actually apply the patch overlay??? It only does that for wrap-file because it deletes everything and re-extracts *and* re-patches, and apparently if no .git/ directory exists in which case it does the same thing...

I don't understand this because I tried this and I could swear it resulted in the correct meson.build file being put in place. :/ None of this makes sense.

If that is the case then commit 2 was just erroneous to begin with.

xclaesse:
What I would suggest: In the case it failed to fetch you can continue and do checkout/rebase operations, also re-apply overlay, but in the end of update_git() function still return False. That indicates meson did its best but the repository might not be up to date. flatpak users could then simply ignore the return value of that command.

The reason I want the return code to be respected is because I know some projects keep in their CI cache old clone of subprojects and then run `meson subprojects update` (to reduce bandwidth consumption, and speedup runner time) and wants it to guarantee that it's building the latest code.

Alternatively, we could add a `--ignore-network-errors` argument?

Also, `meson subprojects download` command has been added specifically for this use-case of a preparation stage to allow doing the build offline. If flatpak refuse to allow running that command, it's their problem...

eli-schwartz:
`--ignore-network-errors` sounds reasonable, but for consistency with wrap-git the update command should really be refreshing the patch overlay no matter what else we decide. I'm actually quite surprised it doesn't...

doraskayo:
If we're already talking about adding command arguments to better support this use case, how about avoiding the update attempt in the first place with an `--offline` argument instead?

It seems to better suit the use case, and doesn't require returning/ignoring failures, showing warnings or performing operations that are expected to fail anyway when the command is executed.

xclaesse:
It's true that if the only thing you need is apply patch overlay, that seems worth adding a specific command for that. Seems even useful for everyone when developing a wrap.

eli-schwartz:
I have reverted this, and followed up by adding a new subcommand dedicated to this purpose, in #9354

Issue #9271, opened 2021-09-17
Required but not found dependencies shouldn't be cached


Plagman:
**Describe the bug**
Some dependencies are required - the typical flow when a user is confronted with them is to install the missing dependency, then run the build again. Without user intervention such as forcing reconfigure, it will keep failing as the missing dependency stays cached.

**To Reproduce**
On a system with wayland-protocols 1.21, check out wlroots prior to commit https://github.com/swaywm/wlroots/commit/c67e3fe3b7e189ed7a617cac412a16116d77cf21 and build it naively with eg. `mkdir build && cd build && meson ..`, then `ninja`. Then update it with eg. `git pull --rebase && ninja` from the same folder. Observe required dependency error. Acquire the missing dependency (through eg. `pacman -Syu`), then run `ninja` again. Observe error:

```
wlroots| Dependency wayland-protocols found: NO found 1.21 but need: '>=1.22' (cached)

../subprojects/wlroots/protocol/meson.build:1:0: ERROR: Dependency 'wayland-protocols' is required but not found.
```

**Expected behavior**
meson appears to know that the dependency is missing, required, and not found. It seems like it has all the information to bypass its own cache here and make this process easier on the user and not require a manual meson reconfigure step requiring more familiarity with its internals.

**system parameters**
Arch Linux, meson 0.59.1, ninja 1.10.2, Python 3.9.7

xclaesse:
The problem here is the dependency is not missing, meson found version 1.21 and thus kept that version into the cache. I don't know if there is a way to fix it, other than removing the dependency cache completely... When changing system dependencies, you often have to rebuild from scratch using `--wipe` option. We discussed about changing `--reconfigure` to also drop the cache, but I don't think we actually did it yet.

I'm personally actually in favor of removing Meson's on-disk cache because invalidation caused so many issues, it's pretty much impossible to get it right. From a performance POV, using a python pkg-config implementation make it so much faster than using the CLI that it's probably not even needed to use a cache at all. See https://github.com/mesonbuild/meson/pull/9021 for a pkg-config implementation in meson.

dcbaker:
I know we disagree on the cache, it makes projects like mesa that have a *lot* of dependencies much faster to reconfigure, especially when cmake is involved. However, this does look like a legitimate bug, if we have a dependency that doesn't meet the version requirements it shouldn't be being pulled from the cache as valid, we should try to find it again. I think the version is already part of the cache key anyway, so we should be able to check this pretty easily.

[...]

PR #9281, opened 2021-09-20
dependency: If cached version is outdated, look on system again

xclaesse:
Fixes: #9271

### DIFF
--- a/mesonbuild/interpreter/dependencyfallbacks.py
+++ b/mesonbuild/interpreter/dependencyfallbacks.py
@@ -223,6 +223,10 @@ class DependencyFallbacksHolder(MesonInterpreterObject):
         if cached_dep:
             found_vers = cached_dep.get_version()
             if not self._check_version(wanted_vers, found_vers):
+                if not override:
+                    # We cached this dependency on disk from a previous run,
+                    # but it could got updated on the system in the meantime.
+                    return None
                 mlog.log('Dependency', mlog.bold(name),
                          'found:', mlog.red('NO'),
                          'found', mlog.normal_cyan(found_vers), 'but need:',


xclaesse:
Ok, was actually much easier than I thought. @plagman could you give this a try please?

mesonbuild/interpreter/dependencyfallbacks.py
```diff
@@ -219,10 +220,15 @@ def _get_cached_dep(self, name: str, kwargs: TYPE_nkwargs) -> T.Optional[Depende
...
             cached_dep = self.coredata.deps[for_machine].get(identifier)
+            from_disk_cache = True
 
         if cached_dep:
             found_vers = cached_dep.get_version()
             if not self._check_version(wanted_vers, found_vers):
+                if from_disk_cache:
+                    # We cached this dependency on disk from a previous run,
+                    # but it could got updated on the system in the meantime.
+                    return None
```
dcbaker, line 231: I don't think we even need this flag, right? we're already inside `cached_dep`, so we basically just need to check if `override.explicit`, and error for that case, otherwise return None, or did I miss something?

xclaesse: It could be written as `if not override:` but it felt better to have a proper variable name to describe that condition.

eli-schwartz: Isn't that what comments are for?

dcbaker: I'm concerned about having two variables to represent the same state, it's really easy for them to get out of sync and end up with subtly broken code.

xclaesse: ok, updated :)

dcbaker (approved):

dcbaker:
Thanks for making that change, this looks good to me as soon as CI is green.

PR #9282, opened 2021-09-20
Add platform agnostic tests

xclaesse:
This adds a new category of tests that does not need to run on all
platforms during CI. For now only run them on Linux runners because they
are not the bottleneck.


xclaesse:
CC @eli-schwartz

xclaesse:
Hmm, wondering why LGTM never complained about those unused imports in the past: `from unittests.platformagnostictests import PlatformAgnosticTests`. The problem is they are actually used, indirectly by unittest module.

eli-schwartz (approved):
This looks fantastic, thanks!

Issue #7176, opened 2020-05-20
Dlang: -Dbuildtype=minsize invokes -Os instead of -Oz for LDC


denizzzka:
`-Dbuildtype=minsize` invokes `-Os` instead of `--Oz` for LDC

As result compiler produces binaries about 10% larger.

**system parameters**
* Is this a [cross build](https://mesonbuild.com/Cross-compilation.html) or just a plain native build (for the same computer)?
Cross build
* what `meson --version`
Current ~master

RossSmyth:
This also occurs in GCC.

eli-schwartz:
Fallout from 4761e4cad946b42daa8723672957e9c1eaeaa5be I guess?

PR #9286, opened 2021-09-21
Use `-Oz` when `optimization=s` in Clang

Tachi107:
`-Oz` is the appropriate flag to use when you want to produce the smallest possible binary, and is one would expect when setting `optimization` to `s` or using the `minsize` build type.

More info [here](https://android.googlesource.com/platform/ndk/+/master/docs/ClangMigration.md#versus) 

Related to #7176


dcbaker (approved):
I can see references to the `-Oz` option for clang back into 3.8 pretty quickly, so I think it's safe to implement `Oz` without any type checking guards

PR #3489, opened 2018-04-27
Convert buildtype to optimization and debug options

jpakkane:
The current setup of a couple of different build types is a bit limiting. Ideally you'd want to toggle the optimization levels and "debugness" independent of each other. Some people might want to do releases with `-O2` and others with `-O3` for example.

This is a very rudimentary fix showing how a small fraction of this would be done. The full setup would go something like this:

 - optimization gets its own option, with values as shown in this MR
 - debug gets a boolean switch
 - buildtype gets a new value of `custom`
 - setting `buildtype` sets also the value of the other two to match current behaviour
 - optimization and debug can be set to any value combinations
 - if the combination exactly matches some current build type, the value of `builtype` is set to that, otherwise it is set to `custom`

Comments welcome.

PR #9287, opened 2021-09-21
compilers/gnu: set level 0 optimization to '-O0'

dcbaker:
GCC with optimization set to 0 does not actually result in no
optimizations, which can be annoying when trying to use a debugger like
gdb, and finding that your variable has been optimized out. We already
do this with clang, so gcc is a bit of an outlier here.

### DIFF
--- a/mesonbuild/compilers/mixins/gnu.py
+++ b/mesonbuild/compilers/mixins/gnu.py
@@ -55,7 +55,7 @@ gnulike_buildtype_args = {
 }  # type: T.Dict[str, T.List[str]]
 
 gnu_optimization_args = {
-    '0': [],
+    '0': ['-O0'],
     'g': ['-Og'],
     '1': ['-O1'],
     '2': ['-O2'],


eli-schwartz:
So which optimization level is the "don't do anything" option now?

This breaks --buildtype=plain. NACK.

dcbaker:
I see a difference with gcc 10.3 on nixos, possibly they're patching things?

dcbaker:
Still, it seems better to explicitly set the value than depend on it being implicitly set, especially when in (at least one case) not setting results in not -O0

jpakkane:
If memory serves we did use to use `-O0` but then we stopped doing that because it caused some problems.

dcbaker:
I can't find that, all I could find is that when we split the buildtypes into debug and optimization this silently disapeared. The combined options had `['-O0', '-g']`, and the split options only put `-g` in the debug options. There's no comments saying that there were problems with '-O0'. At one point we set '-Og', which *did* have problems because it was new in gcc 4.8.

https://github.com/mesonbuild/meson/pull/3489

Issue #9284, opened 2021-09-21
Incorrect behavior in Python module when the sysconfig paths aren't relative to the install prefix


FFY00:
**Describe the bug**
When calculating the install paths in the Python module, if a sysconfig path is not relative to install prefix, Meson will have the wrong path.

**To Reproduce**
Change the sysconfig paths on the Python interpreter to not be relative to the prefix, and try to install a Python module with meson.

Eg.
```
$ python -m sysconfig
Platform: "linux-x86_64"
Python version: "3.11"
Current installation scheme: "posix_prefix"

Paths:
	data = "/home/anubis/.virtualenvs/meson-cpython"
	include = "/usr/local/include/python3.11d"
	platinclude = "/usr/local/include/python3.11d"
	platlib = "/opt/custom/lib/python3.11/site-packages"
	platstdlib = "/home/anubis/.virtualenvs/meson-cpython/lib/python3.11"
	purelib = "/opt/custom/lib/python3.11/site-packages"
	scripts = "/home/anubis/.virtualenvs/meson-cpython/bin"
	stdlib = "/usr/local/lib/python3.11"

...
```

```meson
project(
    'pure',
    version: '1.0.0',
)

py_mod = import('python')
py = py_mod.find_installation()

py.install_sources('pure.py')
```

```
$ DESTDIR=destdir ninja install
[0/1] Installing files.
Installing /home/anubis/git/mesonpy/tests/packages/pure/pure.py to /home/anubis/git/mesonpy/tests/packages/pure/build/destdir/usr/local/opt/custom/lib/python3.11/site-packages
```

The file would get installed to `/usr/local/opt/custom/lib/python3.11/site-packages`, which is incorrect.

**Expected behavior**
As Meson cannot correctly calculate the correct path, it should error out. We could optionally bypass this for a set of known paths, like `/opt`, but I am a bit unsure about that.

**system parameters**
* Is this a [cross build](https://mesonbuild.com/Cross-compilation.html) or just a plain native build (for the same computer)? No
* what operating system (e.g. MacOS Catalina, Windows 10, CentOS 8.0, Ubuntu 18.04, etc.)
* what Python version are you using e.g. 3.8.0
* what `meson --version`? `master`
* what `ninja --version` if it's a Ninja build

xclaesse:
Meson master should at least warn when this happens. I think the generate concensus was that it's impossible to get it always right, and we should add a Meson built-in option to set the python install path.

[...]

Issue #4267, opened 2018-09-26
meson 0.48.0 quietly introduces dependency on setuptools package


decathorpe:
The official stance is (or used to be?) that meson isn't allowed to depend on anything outside the python standard library: http://mesonbuild.com/Use-of-Python.html#dependency-hell

These commits introduce a run-time dependency on setuptools, because it adds imports from `pkg_resources`:

https://github.com/mesonbuild/meson/commit/2ee28029f945d74d01e63d6efd8802c9df03b28f
https://github.com/mesonbuild/meson/commit/86298f2109d215ad6b26d3462af7b685d52d0dd7

Probably the documentation on mesonbuild.com needs to be updated, and a mention of the new dependency in the release notes for 0.48.0 would have been nice.

decathorpe:
I'm on fedora, where setuptools is not part of a minimal buildroot, and this caused initial issues when meson was updated to 0.48.0 (it was fixed by @ignatenkobrain already, by adding setuptools to runtime dependencies of meson).

I was just surprised by the quietly added dependency, since I assumed the "nothing except stdlib" thing was still valid.

I understand the reasoning for adding the setuptools requirement - I'd just have liked it to be documented in the release notes at least, since a new dependency is noteworthy in my opinion. And updating the documentation needs to be updated to state "nothing except stdlib and setuptools".

hroncok:
Note that python3 in Fedora 29+ recommends python3-setuptools, so most *users* won't be affected by this as they will get setuptools anyway. On the other hand Fedora's build system does not install recommended dependencies and hence other Fedora packages failed to use meson at build time.

lazka:
I wrote a distutils command once which only uses setuptools on Windows as a build dep: https://github.com/quodlibet/quodlibet/blob/605c54c3270db46f9e1532e3f50a3ee349176f0d/quodlibet/gdist/launchers.py

A bit extreme, but that gets your dependencies to a minimum.

jpakkane:
The Debian packaging [has had a fix applied for this for a while now](https://tracker.debian.org/news/989630/accepted-meson-0480-2-source-into-unstable/).

decathorpe:
(Just cleaning up after myself, since this is obviously going nowhere.)

PaulBone:
This is still broken:

```
paul@fluorine /tmp $ pip3 install --user meson
Collecting meson
[...]

PR #9288, opened 2021-09-21
modules: python: better handling of the Python paths for Debian

FFY00:
Hardcoding the name is fragile, and enabling it based on the existence of
/etc/debian_version (as the is_debianlike helper does) will result in
incorrect paths if the Python binary is not provided by Debian.

Using the deb_system distutils scheme instead makes sure we use the
install path from the current interpreter, which Debian could change
between releases, and gives us the correct value on Python installations
that are not provided by Debian (eg. deadsnakes, Github Action Python,
etc.)

Do notice, though, that there is still no guarantee that these are the
correct paths, as they assume all schemes paths have the install prefix
as a base, see #9284.

### DIFF
--- a/mesonbuild/modules/python.py
+++ b/mesonbuild/modules/python.py
@@ -276,11 +276,43 @@ def python_factory(env: 'Environment', for_machine: 'MachineChoice',
 
 
 INTROSPECT_COMMAND = '''\
+import os.path
 import sysconfig
 import json
 import sys
+import distutils.command.install
 
-install_paths = sysconfig.get_paths(vars={'base': '', 'platbase': '', 'installed_base': ''})
+def get_distutils_paths(scheme=None, prefix=None):
+    import distutils.dist
+    distribution = distutils.dist.Distribution()
+    install_cmd = distribution.get_command_obj('install')
+    if prefix is not None:
+        install_cmd.prefix = prefix
+    if scheme:
+        install_cmd.select_scheme(scheme)
+    install_cmd.finalize_options()
+    return {
+        'data': install_cmd.install_data,
+        'include': os.path.dirname(install_cmd.install_headers),
+        'platlib': install_cmd.install_platlib,
+        'purelib': install_cmd.install_purelib,
+        'scripts': install_cmd.install_scripts,
+    }
+
+# On Debian derivatives, the Python interpreter shipped by the distribution uses
+# a custom install scheme, deb_system, for the system install, and changes the
+# default scheme to a custom one pointing to /usr/local and replacing
+# site-packages with dist-packages.
+# See https://github.com/mesonbuild/meson/issues/8739.
+# XXX: We should be using sysconfig, but Debian only patches distutils.
+
+if 'deb_system' in distutils.command.install.INSTALL_SCHEMES:
+    paths = get_distutils_paths(scheme='deb_system')
+    install_paths = get_distutils_paths(scheme='deb_system', prefix='')
+else:
+    paths = sysconfig.get_paths()
+    empty_vars = {'base': '', 'platbase': '', 'installed_base': ''}
+    install_paths = sysconfig.get_paths(vars=empty_vars)
 
 def links_against_libpython():
     from distutils.core import Distribution, Extension
@@ -290,7 +322,7 @@ def links_against_libpython():
 
 print(json.dumps({
   'variables': sysconfig.get_config_vars(),
-  'paths': sysconfig.get_paths(),
+  'paths': paths,
   'install_paths': install_paths,
   'sys_paths': sys.path,
   'version': sysconfig.get_python_version(),
@@ -373,14 +405,9 @@ class PythonExternalProgram(ExternalProgram):
         sys_paths = self.info['sys_paths']
         rel_path = self.info['install_paths'][key][1:]
         if not any(p.endswith(rel_path) for p in sys_paths if not p.startswith(user_dir)):
-            # On Debian derivatives sysconfig install path is broken and is not
-            # included in the locations python actually lookup.
-            # See https://github.com/mesonbuild/meson/issues/8739.
             mlog.warning('Broken python installation detected. Python files',
                          'installed by Meson might not be found by python interpreter.',
                          once=True)
-            if mesonlib.is_debianlike():
-                rel_path = 'lib/python3/dist-packages'
         return rel_path
 
 
--- a/test cases/python/1 basic/meson.build
+++ b/test cases/python/1 basic/meson.build
@@ -9,14 +9,14 @@ if py_version.version_compare('< 3.2')
 endif
 
 py_purelib = py.get_path('purelib')
-if not py_purelib.endswith('site-packages')
+if not (py_purelib.endswith('site-packages') or py_purelib.endswith('dist-packages'))
   error('Python3 purelib path seems invalid? ' + py_purelib)
 endif
 message('Python purelib path:', py_purelib)
 
 # could be 'lib64' or 'Lib' on some systems
 py_platlib = py.get_path('platlib')
-if not py_platlib.endswith('site-packages')
+if not (py_platlib.endswith('site-packages') or py_platlib.endswith('dist-packages'))
   error('Python3 platlib path seems invalid? ' + py_platlib)
 endif
 
--- a/test cases/python3/1 basic/meson.build
+++ b/test cases/python3/1 basic/meson.build
@@ -9,13 +9,13 @@ if py3_version.version_compare('< 3.2')
 endif
 
 py3_purelib = py3_mod.sysconfig_path('purelib')
-if not py3_purelib.to_lower().startswith('lib') or not py3_purelib.endswith('site-packages')
+if not py3_purelib.to_lower().startswith('lib') or not (py3_purelib.endswith('site-packages') or py3_purelib.endswith('dist-packages'))
   error('Python3 purelib path seems invalid?')
 endif
 
 # could be 'lib64' or 'Lib' on some systems
 py3_platlib = py3_mod.sysconfig_path('platlib')
-if not py3_platlib.to_lower().startswith('lib') or not py3_platlib.endswith('site-packages')
+if not py3_platlib.to_lower().startswith('lib') or not (py3_platlib.endswith('site-packages') or py3_platlib.endswith('dist-packages'))
   error('Python3 platlib path seems invalid?')
 endif
 


mesonbuild/modules/python.py
```diff
@@ -280,7 +280,18 @@ def set_env(name, value):
...
 import sys
 
-install_paths = sysconfig.get_paths(vars={'base': '', 'platbase': '', 'installed_base': ''})
+# On Debian derivatives, the Python interpreter shipped by the distribution uses
+# a custom install scheme, deb_system, for the system install, and changes the
+# default scheme to a custom one pointing to /usr/local and replacing
+# site-packages with dist-packages.
+# See https://github.com/mesonbuild/meson/issues/8739.
+empty_vars = {'base': '', 'platbase': '', 'installed_base': ''}
+if 'deb_system' in sysconfig.get_scheme_names():
```
xclaesse, line 289: On my Ubuntu 20.04:
```
>>> sysconfig.get_scheme_names()
('nt', 'nt_user', 'osx_framework_user', 'posix_home', 'posix_prefix', 'posix_user')
```

When has deb_system been added? Seems we still need the hack at least for older versions.

FFY00: It has existed for a long time, but on older versions we may need to check `distutils.commands.install.INSTALL_SCHEMES` instead.

FFY00: Which Python version is that btw?

FFY00: I just checked, and Debian started patching `syconfig` in 3.6, which is the minimum supported version by Meson. Maybe they did not patch it in all minor versions, and you happen to be running one of those?

xclaesse: I tried Ubuntu 21.10 and Debian 11, both does not have that scheme. Where did you find that trick? Is that super new in Debian experimental?

xclaesse: Maybe it comes from non std modules then. I tried by spawning a docker and installed python3 package.

FFY00: I did the same thing and confirmed sysconfig is not patched.

As I understood, they were doing it since 3.6, maybe I am mistaken.
https://salsa.debian.org/cpython-team/python3/-/blob/python3.6/debian/patches/sysconfig-debian-schemes.diff

Anyway, `distutils.command.install.INSTALL_SCHEMES` is patched, so we can use that.
```
>>> import distutils.command.install
>>> distutils.command.install.INSTALL_SCHEMES.keys()
dict_keys(['unix_prefix', 'unix_local', 'deb_system', 'unix_home', 'nt', 'nt_user', 'unix_user'])
```

The expansion there is less straight-forward, but we can make do. My only concern is distutils being deprecated and scheduled for removal in 3.12, we need to check how Debian/Ubuntu is planning to patch this.

FFY00: Ah, I remember now. The patch for sysconfig was never applied, *sigh*.

See https://github.com/pypa/pip/issues/9617.

We can either get the paths from distutils, or just use it as a way of flagging if the Python installation is patched by Debian.
I think the patch will now be applied in the next Debian version.

FFY00: @xclaesse sorry for the confusion, can you check the logic again and confirm it now works as expected on your system(s)?

mesonbuild/modules/python.py
```diff
@@ -289,8 +316,9 @@ def links_against_libpython():
     return bool(cmd.get_libraries(Extension('dummy', [])))
 
 print(json.dumps({
+  'base_prefix': sys.base_prefix,
   'variables': sysconfig.get_config_vars(),
-  'paths': sysconfig.get_paths(),
+  'paths': paths,
   'install_paths': install_paths,
```
xclaesse, line 322: install_paths is not defined any more, and base_prefix seems unused. Other than that, it seems to give the right values.

I am not an expert in python echosystem, but I thought distutils was about to be deprecated. Maybe that should be used only as fallback if sysconfig gives a value that is not in sys_paths ?

lazka: The problem is that sysconfig was never used to install packages and distros only patched distutils. This was recently uncovered/rectified by pip which compared both at runtime and warned the user to file a bug with their distro.

So the safest way to handle this would be to use distutils and fall back to sysconfig.

FFY00: > install_paths is not defined any more, and base_prefix seems unused. Other than that, it seems to give the right values.

Sorry for the bad cleanup, I will fix this later.

> I am not an expert in python echosystem, but I thought distutils was about to be deprecated. Maybe that should be used only as fallback if sysconfig gives a value that is not in sys_paths ?

It is, I didn't keep that logic here to make it easier for you to verify the disutils bit.

> The problem is that sysconfig was never used to install packages and distros only patched distutils. This was recently uncovered/rectified by pip which compared both at runtime and warned the user to file a bug with their distro (but that doesn't fix older distros)

I think by distros you mean Debian, and Debian-based, which just inherit the package. Fedora for eg. currently patches sysconfig. Recent-ish pip versions should use sysconfig to install IIRC.

> So the safest way to handle this would be to use distutils and fall back to sysconfig once distutils is really gone (not much visible upstream movement there atm...)

IMO we should use sysconfig on Python >= 3.12, and distutils on older if it is patched by Debian.

I think Debian will drop their patches in favor of https://www.python.org/dev/peps/pep-0668/ in 3.12 but I am not sure, and we won't know until they release it. But considering that there won't be distutils anymore, if they keep the patches, they will probably move them to sysconfig.

xclaesse:
Gave this a try on my Windows machine too:
```
>>> import distutils.command.install
>>> distribution = distutils.dist.Distribution()
>>> install_cmd = distribution.get_command_obj('install')
>>> install_cmd.prefix = ''
>>> install_cmd.finalize_options()
  File "<stdin>", line 1, in <module>
  File "C:\Python38\lib\distutils\command\install.py", line 344, in finalize_options
    self.convert_paths('lib', 'purelib', 'platlib',
  File "C:\Python38\lib\distutils\command\install.py", line 487, in convert_paths
    setattr(self, attr, convert_path(getattr(self, attr)))
  File "C:\Python38\lib\distutils\util.py", line 122, in convert_path
    raise ValueError("path '%s' cannot be absolute" % pathname)
ValueError: path '/Lib/site-packages' cannot be absolute
```

FFY00:
Okay, we can use a dummy prefix and then remove it.

xclaesse:
OOC, that's the plan with distutils upstream? They could remove the module completely, so we should guard the import behind try/except and fallback to sysconfig? Also, other than deb_system, are there any other case where sysconfig is not enough?

FFY00:
Yes, 1sec. I will push my code to handle all that.

xclaesse:
I'm also wondering if we should actually allow installing python modules outside of Meson's prefix. Thinking about Windows where they should go into `C:\Python38\Lib\site-packages`. But that's a bridge we could cross later.

mesonbuild/modules/python.py
```diff
@@ -276,11 +276,47 @@ def set_env(name, value):
...
+
+# On Debian derivatives, the Python interpreter shipped by the distribution uses
+# a custom install scheme, deb_system, for the system install, and changes the
+# default scheme to a custom one pointing to /usr/local and replacing
+# site-packages with dist-packages.
+# See https://github.com/mesonbuild/meson/issues/8739.
+# XXX: We should be using sysconfig, but Debian only patches distutils.
+
+debian_distutils = False
+if sys.version_info < (3, 12):
```
xclaesse, line 309: That's doing a gamble long into the future, we are 3.9 currently... anything can still happen, no? What about instead:
```python
try:
  debian_distutils = 'deb_system' in distutils.command.install.INSTALL_SCHEMES
except:
  pass
```
As long as `deb_system` remains in distutils, it's pretty safe to assume they kept the patch. That would keep working if they postpone migration to >3.12.

mesonbuild/modules/python.py
```diff
@@ -289,8 +325,9 @@ def links_against_libpython():
     return bool(cmd.get_libraries(Extension('dummy', [])))
 
 print(json.dumps({
+  'base_prefix': sys.base_prefix,
```
xclaesse, line 328: Still unused AFAIK.

FFY00: Yup, sorry, forgot to remove.

mesonbuild/modules/python.py
FFY00: `distutils.command.install` is not installed on Debian by default, it is provided by `python3-distutils`, so we want to actually import and error out, perhaps with a nicer message.

> As long as `deb_system` remains in distutils, it's pretty safe to assume they kept the patch. That would keep working if they postpone migration to >3.12.

They won't, that would require the Steering Council to revert the PEP (https://www.python.org/dev/peps/pep-0632/), which is very very unlikely. `distutils` has been frozen in CPython and extracted to https://github.com/pypa/distutils/. It will then be merged into setuptools at some point in the future.

xclaesse: > `distutils.command.install` is not installed on Debian by default, it is provided by `python3-distutils`, so we want to actually import and error out, perhaps with a nicer message.

Yes, that's the whole problem. Meson must not depend on extra modules, so we cannot fail here, we have to fallback with a warning (hinting at install distutils). Pretty sure there are tons of CI out there that only install `python3` package in their docker image and we cannot break them.

FFY00: This script already depends on distutils for `links_against_libpython`.

> Pretty sure there are tons of CI out there that only install `python3` package in their docker image and we cannot break them.

I'd guess almost everyone has `python3-pip`, which pull `python3-distutils`.

xclaesse: tbh, this whole story is pretty much going full circle back to just hardcode the path on debian systems. If python3-distutils is not installed by default we still have to hardcode the path... The tricky part is to correctly detect when debian will stop needing that hack, and I have sub-zero confidence with that being 3.12, that's the kind of break everything plan that gets postponed for years.

xclaesse: > This script already depends on distutils for `links_against_libpython`.

Hmm, that's actually a good point... probably make that conditional too. After double checking, that will make the script fail, but Meson then continue with some default values. IMHO we should not make that script abort, but return sysconfig values instead.

xclaesse: > I'd guess almost everyone has `python3-pip`, which pull `python3-distutils`.

On developer's laptop of course, but here I'm more thinking about build bots, ci docker, etc...

FFY00: distutils is not an external module, it's part of the standard library, Debian just splits it into a different package against the upstream wishes.

The Meson package in Debian should depend on `python3-disutils`, and if people installing Meson via pip, well, they have pip so they should have distutils.

IMO it's reasonable to ask people to install `python3-distutils` if it is not there for some reason, given it should already be there anyway. This should not affect almost anyone.

xclaesse: Maybe I'm being too cautious... `import('python').find_installation() aborts already in a debian/ubuntu docker where I install only meson package:
```
meson.build:3:17: ERROR: <ExternalProgram 'python3' -> ['/usr/bin/python3']> is not a valid python or it is missing setuptools
```

Been like that for 3 years, no complains... ok, I guess we could do it then...

FFY00: I think you should report a bug to the distro, the Meson package should depend on `python3-distutils`.

xclaesse: Actually, people do complain, but we just put the blame on Debian for their mess, it seems: https://github.com/mesonbuild/meson/issues/4267.

FFY00: Anyway, just let me know if you want to make any change to the logic.

xclaesse: ok, ok, sorry for the long conversation, I learned a lot today :D

Still, I'm not sure we should make this `<3.12` check. What happens if 3.12 gets released and debian still haven't patched sysconfig instead of distutils? Why isn't it enough to check if `deb_system` is there regardless of the version?

FFY00: No worries, this is a mess :sweat_smile: 

Considering the script will fail on `links_against_libpython` if distutils is missing anyway, we can do that. The worry was that the import would fail if distutils is missing and we would mistakenly think that we are in a new version of Python without it. But yeah, we are gonna fail anyway later.

FFY00: No worries, this is a mess :sweat_smile: 

Considering the script will fail on `links_against_libpython` if distutils is missing anyway, we can do that. The worry was that the import would fail if distutils is missing and we would mistakenly think that we are in a new version of Python without it. But yeah, we are gonna fail anyway later.

FFY00: I added a heuristic to keep the helpful message.

FFY00: But one worry about this is that after 3.12, people will be able to provide a `distutils` module via pip. So yeah, I would prefer the version check, but we can go with the try block if you feel it is better.

mesonbuild/modules/python.py
```diff
@@ -276,11 +276,57 @@ def set_env(name, value):
...
+    # not installed. Let's try detecting that and error out with an helpful
+    # error message.
+    import distutils
+    try:
+        import distutils.command.install
+    except ModuleNotFoundError:
+        raise ModuleNotFoundError('Unable to import distutils, please install python3-distutils')
+    debian_distutils = 'deb_system' in distutils.command.install.INSTALL_SCHEMES
+except ModuleNotFoundError:
+    debian_distutils = False
```
xclaesse, line 321: I don't think we need a try inside a try, `import distutils` should always work on all platforms, right?

FFY00: Not after 3.12.

FFY00: Unless someone is naughty enough to create a Python package that provides the same module.

FFY00: Unless someone is naughty enough to create a Python package that provides the same module.

FFY00: > In that case we'll need to fix links_against_libpython() too, let's cross that bridge when/if we get there? OTOH, this is getting petty nitpicking I admit ;P

I'd prefer doing in a different PR as it is not actually an issue right now, while this is. We can fix the current handling of the Debian install paths and then explore how to fase out distutils entirely.

> OOC, how confident are we that it's going to be removed in 3.12?

Very.

> It has been formally announced already? When that happens, what's the likely outcome for Debian, they'll patch sysconfig instead?

Yes, my best guess is either PEP 668 or that, or both.
Also related: https://bugs.python.org/issue43976 (maybe skip down to the latest discussion)

> Sorry again for all my questions, I wasn't familiar with this topic, so I like to learned it a bit :)

No worries, I have been dealing with these issues for a little while and still struggle to keep track of everything.

mesonbuild/modules/python.py
eli-schwartz: Yeah, indeed meson has always hard errored if:
- the python installation is required
- distutils is not available

and with an error message that mentioned setuptools :( but that got fixed in #9196

Since INTROSPECT_COMMAND already fails with a traceback when distutils is installed, and this traceback is used to `raise mesonlib.MesonException(f'{python} is not a valid python or it is missing distutils')`, I don't see any point in adding a bunch of complicated try/excepts here. Please just import `distutils.command.install` on one line and check for 'deb_system' in INSTALL_SCHEMES on another line, and defer to the existing handling to raise a suitable error message.

FFY00: Well, this identifies if the origin of the issue is indeed python3-distutils missing on Debian, and points that out in the traceback, so I think there is a point in keeping it. But I can remove if you want.

eli-schwartz: Maybe the existing error message can be improved to provide a debian-specific recommendation?

Note that the traceback is run in a subprocess and then logged as mlog.debug in meson itself, and so is your ModuleNotFoundError. However the MesonException is logged as a fatal error. :)

FFY00: Can we do that in a separate PR? I really want to move this forward as it is blocking my work. With the patch as-is, user will get the same message they were getting previously, which the added extra that they now get a better debug message.

eli-schwartz: It should not be that complicated:

```diff
diff --git a/mesonbuild/modules/python.py b/mesonbuild/modules/python.py
index bae5f9f05..f34d56d94 100644
--- a/mesonbuild/modules/python.py
+++ b/mesonbuild/modules/python.py
@@ -281,6 +281,8 @@ import sysconfig
 import json
 import sys
 
+import distutils.command.install
+
 def get_distutils_paths(*, scheme=None, prefix=None):
     import distutils.dist
     distribution = distutils.dist.Distribution()
@@ -305,22 +307,7 @@ def get_distutils_paths(*, scheme=None, prefix=None):
 # See https://github.com/mesonbuild/meson/issues/8739.
 # XXX: We should be using sysconfig, but Debian only patches distutils.
 
-try:
-    # Debian patches distutils, keeping only distutils.version, so
-    # `import distutils` will always work, but
-    # `import distutils.command.install` will fail if python3-distutils is
-    # not installed. Let's try detecting that and error out with an helpful
-    # error message.
-    import distutils
-    try:
-        import distutils.command.install
-    except ModuleNotFoundError:
-        raise ModuleNotFoundError('Unable to import distutils, please install python3-distutils')
-    debian_distutils = 'deb_system' in distutils.command.install.INSTALL_SCHEMES
-except ModuleNotFoundError:
-    debian_distutils = False
-
-if debian_distutils:
+if 'deb_system' in distutils.command.install.INSTALL_SCHEMES
     paths = get_distutils_paths(scheme='deb_system')
     install_paths = get_distutils_paths(scheme='deb_system', prefix='')
 else:
@@ -719,7 +706,11 @@ class PythonModule(ExtensionModule):
                 return python
             else:
                 if required:
-                    raise mesonlib.MesonException(f'{python} is not a valid python or it is missing distutils')
+                    msg = f'{python} is not a valid python or it is missing distutils.'
+                    if mesonlib.is_debianlike():
+                        msg += ' Try installing python3-distutils.'
+                    raise mesonlib.MesonException(msg)
+
                 return NonExistingExternalProgram()
 
         raise mesonlib.MesonBugException('Unreachable code was reached (PythonModule.find_installation).')
```


The final diff hunk is complete overkill, the lack of distutils in versions of python released before the end of 2023 is already a sign that is_debianlike() is true... but why not.

FFY00: > the lack of distutils in versions of python released before the end of 2023 is already a sign that is_debianlike() is true

But the opposite is not true. Your patch will fail with the `Try installing python3-distutils` message if the introspection script fails on Python installations that are not patched by Debian, as long we are in a Debian based system. This is the case of the ubuntu runner in Github Actions for example, where they have their own Python.

It's not a big deal in this case in particular, but it is precisely this kind of inaccuracies, well it was exactly this inaccuracy, that lead to bugs like the one that is being fixed in this PR. For that reason I always like to avoid such inaccurate checks, even if they do not have significant consequences -- they often end up leaking to places that do have consequences.

Anyway, your patch completely removes the accurate pinpoint of the issue. The approach is basically "the script failed? well, maybe install pyhon3-distutils idk".

I don't really want to keep arguing with you, so just fix up my commits with the code you deem to be better 🙃

FFY00:
@xclaesse can you have a look? I think this is essentially ready.

mesonbuild/modules/python.py
eli-schwartz: Both messages are "the script failed? well maybe install python3-distutils idk".

And the only other failure mode is a `python2` executable reporting it is >=3.0 (or vice versa).

The difference between the two messages is that one of them actually prints output which the user will see, and one of them logs debug output which will *only ever* be seen by people who already know what it means.

> well it was exactly this inaccuracy, that lead to bugs like the one that is being fixed in this PR

No? The inaccuracy which this PR is intending to fix is:

meson uses incorrect sources of information for one vendor's custom-patched runtime, instead let us detect and special case this to handle it correctly.

The inaccuracy I am proposing is:

there is currently a vague 'idk' message, and the only way to improve it is to improve *that* message, which you can do by implementing subprocess state communication or, if you "really want to move this forward as it is blocking my work", you can at least improve the message to mention the package name.

FFY00: > meson uses incorrect sources of information for one vendor's custom-patched runtime, instead let us detect and special case this to handle it correctly.

The incorrect source of information is checking for the presence of `/etc/debian-version` and deciding that that means the Python installation is patched by Debian.

Like I said, I don't want to argue, so just fix up my commit, or push a new commit with your patch.

eli-schwartz: An alternative way to phrase my review comment on this import nest:

This is logged at mlog.debug which is the wrong place to log informative guidance. No one will ever see it, it only exists so we can tell people "post this really long logfile for us to interpret". We can just get rid of it entirely as it effectively serves no purpose other than to make the code complex.

Aside: feel free to improve this other line of code which is user-visible, now or in the future.

FFY00: Yes, and I agree. I asked above if we could do this in a different PR so that we can move this one, I expect to be some discussion on how exactly it is reported.

FFY00:
I pushed a fix hopefully dealing with the failing tests due to old Python versions.

@eli-schwartz I will submit the improvements to the message in a different pull request, as we probably want to have a different discussion and review for that. Hopefully, we can merge this ASAP.

mesonbuild/modules/python.py
eli-schwartz: Maybe we could just remove the try/except stuff (diff hunks 1 and 2 which I pasted above, but not diff hunk 3), not touch the error message or try to do any error reporting ***yet***, and open a draft PR with the `ModuleNotFoundError` added back and where the pros and cons of different forms of error reporting could be discussed.

FFY00:
Sorry for the delay, I have been busy. I removed the try .. except block for the Debian stuff from this PR, I'll see if I have time to open a new PR next week.

xclaesse:
Looks good to me, thanks!

eli-schwartz:
Thanks! Indeed this looks good to me too.

PR #9290, opened 2021-09-22
Permit buildtargets in generator process

JWolvers:
Hi!
This change permits a generator to consume outputs by any target instead of failing.
This is helpfull to create objcopy generators. I want to use objcopy and an python script to generate a header to embed an executable inside another executable for an dual core embedded target.

### DIFF
--- a/docs/markdown/snippets/permit_generator_any_target.md
+++ b/docs/markdown/snippets/permit_generator_any_target.md
@@ -1,0 +1,3 @@
+## Allow using generator with any Build Target.
+
+Calling `generator.process()` with a BuildTarget is now permitted.
--- a/mesonbuild/interpreter/interpreterobjects.py
+++ b/mesonbuild/interpreter/interpreterobjects.py
@@ -960,14 +960,14 @@ class GeneratorHolder(ObjectHolder[build.Generator]):
         super().__init__(gen, interpreter)
         self.methods.update({'process': self.process_method})
 
-    @typed_pos_args('generator.process', min_varargs=1, varargs=(str, mesonlib.File, build.CustomTarget, build.CustomTargetIndex, build.GeneratedList))
+    @typed_pos_args('generator.process', min_varargs=1, varargs=(str, mesonlib.File, build.BuildTarget, build.CustomTarget, build.CustomTargetIndex, build.GeneratedList))
     @typed_kwargs(
         'generator.process',
         KwargInfo('preserve_path_from', (str, NoneType), since='0.45.0'),
         KwargInfo('extra_args', ContainerTypeInfo(list, str), listify=True, default=[]),
     )
     def process_method(self,
-                       args: T.Tuple[T.List[T.Union[str, mesonlib.File, build.CustomTarget, build.CustomTargetIndex, build.GeneratedList]]],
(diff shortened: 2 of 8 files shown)


mesonbuild/interpreter/interpreterobjects.py
```diff
@@ -974,6 +974,10 @@ def process_method(self,
             FeatureNew.single_use(
                 f'Calling generator.process with CustomTaget or Index of CustomTarget.',
                 '0.57.0', self.interpreter.subproject)
+        elif any(isinstance(a, (build.Target)) for a in args[0]):
```
dcbaker, line 977: CustomTarget is Target, so you'll end up with a false positive here

docs/markdown/snippets/permit_generator_any_target.md
```diff
@@ -0,0 +1,3 @@
+## Allow using generator with any Taget.
```
eli-schwartz, line 1:
```suggestion
## Allow using generator with any Target.
```

mesonbuild/interpreter/interpreterobjects.py
```diff
@@ -974,6 +974,10 @@ def process_method(self,
...
+            FeatureNew.single_use(
+                f'Calling generator.process with any Taget.',
```
eli-schwartz, line 979:
```suggestion
                f'Calling generator.process with any Target.',
```

mesonbuild/interpreter/interpreterobjects.py
JWolvers: The false positive is caught by the if statement above it, but I see now that there is a possible false negative if there is also a CustomTarget, CustomTargetIndex or GeneratedList in the  argument list. I will fix it in my next commit.

mesonbuild/interpreter/interpreterobjects.py
```diff
@@ -974,6 +974,10 @@ def process_method(self,
             FeatureNew.single_use(
                 f'Calling generator.process with CustomTaget or Index of CustomTarget.',
                 '0.57.0', self.interpreter.subproject)
+        if any((isinstance(a, (build.Target)) and not isinstance(a, (build.CustomTarget))) for a in args[0]):
```
dcbaker, line 977: I think this should just be is BuildTarget, which is IIRC every target except CustomTarget

JWolvers: No there also is RunTarget which is also a Target, so using BuildTarget is no option.

dcbaker: I don't think generators can depend on RunTargets. I think RunTargets are only allowed to depend on each other.

jpakkane: Yeah, run targets are conceptually "do something" like upload a generated file somewhere. Build targets are "create something with a specific output name" which is what this is for, since said output is meant to be used as input to the generator.

JWolvers:
Okay I refactored the changes to allow BuildTargets, CustomTargets and CustomTargetIndexes in the Generator.Process function.

bonzini:
Closing because the original repository disappeared and it's not possible to rebase the PR.

PR #9291, opened 2021-09-23
backend/vs: Generate dependencies for CustomTargetIndex for a CustomTarget

anarazel:
Test & fix.


eli-schwartz:
As you are also trying to fix #9259 it would be nice to fold this test into test `common/226 link depends indexed custom target`

anarazel:
ISTM testing the handling of link depends of executables/libraries is reasonably separate from input sources to custom targets...

keszybz:
LGTM.

PR #9294, opened 2021-09-23
ar linker: generate thin archives for uninstalled static libraries

### LINKED ISSUE #2129: Support thin archives for static libraries that are not installed
I have a project which generates several executables which (partially) share the same source files.
Since each executable is first compiled in its own subdirectory, all these shared source files are recompiled for every executable. This is undesirable if I know that identical object files will be produced (that is, the compiler options are identical).

Is there a way to specify the name of the build-subdirectory to avoid this repeated redundant compilation?

I am aware that one could use static libraries to work around this problem, but besides being overkill this also has several disadvantages: It also has the overhead of building the library, and moreover, to support e.g. proper support for LTO it requires linker plugins etc.

### LINKED ISSUE #8057: Cannot create thin archives
I'm doing cross-platform builds and using the ar archiver from our SDK for creating static libraries. The issue I am facing is that I am unable to find a way to override/append the 'csrD' options that mesonbuild/linker.py inherently adds to the ar archival, to 'csrDT' to create thin archives. 

I'm using a crossfile to provide the paths to the compilers, linker and archiver in our SDK. I have tried the following:
1.  Adding the required argument while listing to path to the ar binary:
```
[binaries]
cpp = '/opt/sdk_dir/g++'
ar = ['/opt/sdk_dir/ar', 'csrDT'] 
```

2. Adding the required argument to ar_args under the properties section of the crossfile:
```
[binaries]
cpp = '/opt/sdk_dir/g++'
ar = ['/opt/sdk_dir/ar', 'csrDT'] 

[properties]
cpp_args = '-fno-exceptions'
ar_args  = 'csrDT'
```
3. Adding the required argument to ar_link_args under the properties section of the crossfile:
```
[binaries]
cpp = '/opt/sdk_dir/g++'
ar = ['/opt/sdk_dir/ar', 'csrDT'] 

[properties]
cpp_args = '-fno-exceptions'
ar_link_args  = 'csrDT'
```
Neither of the three approaches worked for me. I'm not really sure on what else I could try to override the default options. There doesn't seem to be a built-in option that I could toggle either, to make this happen. Looking for some help here.

I'm using meson 0.51.2

### LINKED ISSUE #9292: Overhead for uninstalled, static_library targets too great
**Describe the bug**
In order to create groupings of files with different C or C++ compiler flags, IIUC, one must use `static_library` in Meson and simply not install them. Those libraries can be linked into targets that are.

However, those libraries are not thin and therefore have a high overhead: they occupy as much disk space as the .o files that are included and they take time to create (disk I/O). Ideally, there would be a way to create a "phantom" or "convenience" library (Automake's term is "convenience") that cannot get installed and Meson figures out on its own how to deal with it. Maybe it simply passes the listing of .o files to the target that uses them, if it's linked `link_whole`.

Or it creates thin static libraries otherwise. See #8057 for a more general case for thin static libraries.

**System parameters**
* what operating system (e.g. MacOS Catalina, Windows 10, CentOS 8.0, Ubuntu 18.04, etc.) Linux
* what Python version are you using? 3.6 to 3.9
* what `meson --version`: 0.59.1
* what `ninja --version` if it's a Ninja build: 1.10.2

eli-schwartz:
Since they will never be used outside of the build directory, they do not need to literally contain the .o files, and references will besufficient.

This covers a major use of object libraries, which is that the static library would potentially take up a lot of space by including another copy of every .o file.

Fixes #9292
Fixes #8057
Fixes #2129


mesonbuild/linkers/linkers.py
```diff
@@ -174,17 +174,26 @@ def __init__(self, exelist: T.List[str]):
...
         self.can_rsp = '@<' in stdo
 
     def can_linker_accept_rsp(self) -> bool:
         return self.can_rsp
 
-    def get_std_link_args(self) -> T.List[str]:
-        return self.std_args
+    def get_std_link_args(self, is_thin: bool) -> T.List[str]:
+        # not supported by armar
+        # FIXME: osx ld rejects this: "file built for unknown-unsupported file format"
```
eli-schwartz, line 192: I don't understand this OSX failure.

From https://github.com/eli-schwartz/meson/runs/3693121858

```
cc  -o prog prog.p/prog.c.o -Wl,-dead_strip_dylibs -Wl,-headerpad_max_install_names -Wl,-undefined,error libhelper.a
ld: warning: ignoring file libhelper.a, building for macOS-x86_64 but attempting to link with file built for unknown-unsupported file format ( 0x21 0x3C 0x74 0x68 0x69 0x6E 0x3E 0x0A 0x2F 0x20 0x20 0x20 0x20 0x20 0x20 0x20 )
Undefined symbols for architecture x86_64:
  "_func", referenced from:
      _main in prog.c.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

jpakkane:
Featurewise this is good, just have to ensure that it does not cause regressions for some use cases. (No, I don't know what those would be, just thinking out loud.)

dcbaker (approved):
This looks good to me to.

mesonbuild/linkers/linkers.py
```diff
@@ -174,17 +174,26 @@ def __init__(self, exelist: T.List[str]):
...
         # Enable deterministic builds if they are available.
+        stdargs = 'csr'
+        thinargs = ''
         if '[D]' in stdo:
-            self.std_args = ['csrD']
-        else:
-            self.std_args = ['csr']
+            stdargs += 'D'
+        if '[T]' in stdo:
+            thinargs = 'T'
```
keszybz, line 182: `thinargs += 'T'` (for consistency)

mesonbuild/linkers/linkers.py
```diff
@@ -174,17 +174,26 @@ def __init__(self, exelist: T.List[str]):
...
+        self.std_args = [stdargs]
+        self.std_thin_args = [stdargs + thinargs]
```
keszybz, line 184: Maybe write this as
```py
self.std_args = ('csr' +
                         ('D' if '[D]' in stdo else ''))
self.std_thin_args = (self.std_args +
                                 ('T' if '[T]' in stdo else ''))
```
I think it's easier to read.

keszybz (commented):
We use a lot of convenience libraries for systemd, and this PR makes great improvements:
```
$ ls -l build{,-meson}/src/core/*.a              
-rw-r--r-- 1 zbyszek zbyszek 10208766 Oct  1 14:47 build/src/core/libcore.a
-rw-r--r-- 1 zbyszek zbyszek    29218 Oct  3 14:13 build-meson/src/core/libcore.a
$ ls -l build{,-meson}/src/basic/libbasic.a       
-rw-r--r-- 1 zbyszek zbyszek 3120020 Oct  1 14:46 build/src/basic/libbasic.a
-rw-r--r-- 1 zbyszek zbyszek   35110 Oct  3 14:12 build-meson/src/basic/libbasic.a
```

Me likes ;)

eli-schwartz:
Not a new alert, lol, just a moved one. Still, why not fix that too...

PR #9295, opened 2021-09-24
0.59 backport 9134

eli-schwartz:
Let's make sure I actually got these cherry-pick conflicts correct...

/cc @nirbheek

### DIFF
--- a/data/test.schema.json
+++ b/data/test.schema.json
@@ -21,11 +21,14 @@
             "type": "string",
             "enum": [
               "file",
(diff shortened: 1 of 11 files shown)


eli-schwartz:
(the difference between args being `[]` or `['']` depending on whether typed_pos_args is in play)

eli-schwartz:
@nirbheek okay, this now passes all tests other than the 2 unexpected-run tests on Ubuntu. \o/

PR #9299, opened 2021-09-26
interpreter: Holderify arrays and dicts

mensinda:
This is the final refactoring for extracting the builtin object logic out of Interpreterbase. I decided to do both arrays and dicts in one go since splitting it would have been a lot more confusing.

Additionally, the `RangeHolder` is moved to the interpreter.


dcbaker (approved):
Looks good, thanks

PR #9302, opened 2021-09-26
pkgconfig module: correctly generate Libs search path with absolute install_dir

eli-schwartz:
For example the OpenRC build files install libraries to install_dir: '/lib' and this works, but causes the generated pkg-config to say:

```
prefix=/usr

Libs: -L${prefix}//lib
```

which is both ugly (double //) and resolves to /usr/lib which is exactlywhat does not work.


eli-schwartz:
Found while reading the meson.build files of random projects :D and wondering ***why*** doesn't openrc use the pkgconfig module.

I suppose @williamh can confirm if my guess is correct, that the pkgconfig module was attempted and rejected because it gave incorrect results here.

mesonbuild/modules/pkgconfig.py
```diff
@@ -319,9 +319,13 @@ def _make_relative(self, prefix, subdir):
...
         try:
-            return subdir.relative_to(prefix).as_posix()
+            libdir = subdir.relative_to(prefix)
         except ValueError:
-            return subdir.as_posix()
+            libdir = subdir
+        if libdir.is_absolute():
+            return libdir.as_posix()
+        else:
+            return f'${{prefix}}/{libdir.as_posix()}'
```
keszybz, line 328: This should be equivalent to `Path(prefix).joinpath(libdir).as_posix()`

keszybz (commented):
Looks reasonable.

mesonbuild/modules/pkgconfig.py
eli-schwartz: No, because this is a pkg-config file so `${prefix}` is a string literal and not a python variable.

keszybz: Oh, right. So `return Path('${prefix}').joinpath(libdir).as_posix()` ?

eli-schwartz: What's the advantage of that over the current method?

xclaesse: I find it dangerous and useless to use a f-string in this case, `'${prefix}/' + libdir.as_posix()` does the same. Less confusing than the double brackets.

xclaesse: @keszybz IMHO it's not worth going through a Path() object just to literaly put a '/' between 2 strings. In this case we explicitly don't want a platform specific path.

eli-schwartz: Fair enough about the f-string confusion.

keszybz: Please note the comment is attached to 4 lines. `.joinpath` does what the conditional does.

xclaesse: ah, I see, makes more sense indeed. I'm fine either way, for small nitpicking like that I leave it to the author to pick their preferred way :)

eli-schwartz: I have been trying to decide whether using pathlib on a string literal `${prefix}` feels better or worse than having more lines of code...

keszybz:
LGTM.

PR #9303, opened 2021-09-27
fix message function accepting bools and casting to string

eli-schwartz:
This was allowed by accident despite what meson said would work, because in python a bool counts as a subclass of int.

### DIFF
--- a/mesonbuild/interpreter/interpreter.py
+++ b/mesonbuild/interpreter/interpreter.py
@@ -91,6 +91,8 @@ def stringifyUserArguments(args, quote=False):
         return '[%s]' % ', '.join([stringifyUserArguments(x, True) for x in args])
     elif isinstance(args, dict):
         return '{%s}' % ', '.join(['{} : {}'.format(stringifyUserArguments(k, True), stringifyUserArguments(v, True)) for k, v in args.items()])
+    elif isinstance(args, bool):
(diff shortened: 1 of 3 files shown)


mesonbuild/interpreter/interpreter.py
```diff
@@ -91,7 +91,7 @@ def stringifyUserArguments(args, quote=False):
         return '[%s]' % ', '.join([stringifyUserArguments(x, True) for x in args])
     elif isinstance(args, dict):
         return '{%s}' % ', '.join(['{} : {}'.format(stringifyUserArguments(k, True), stringifyUserArguments(v, True)) for k, v in args.items()])
-    elif isinstance(args, int):
+    elif type(args) == int: # isinstance will accept bools too
```
dcbaker, line 94: you should use `is` not `==` here.

But it would be more idiomatic to do something like:
```python
elif isinstance(args, bool):
    pass  # we want this to fallthrough to the error case
elif isinstance(args, int):
    return str(args)
```

dcbaker (changes requested):

mesonbuild/interpreter/interpreter.py
eli-schwartz: Fair enough.

dcbaker (approved):
LGTM

Issue #9300, opened 2021-09-26
The `check` argument of `run_command` should be `True` by default


Volker-Weissmann:
The `run_command` argument takes a `check` bool-argument. If `check: true`, is given and the command fails, meson will error out.

May I ask, who thought that it was a good idea to let `check` default to `false`? Ignoring errors by default goes against everything I learned.

So essentially, I want something like this, (plus documentation updates)
```
diff --git a/mesonbuild/interpreter/interpreterobjects.py b/mesonbuild/interpreter/interpreterobjects.py
index fca371cbc..3b3218c54 100644
--- a/mesonbuild/interpreter/interpreterobjects.py
+++ b/mesonbuild/interpreter/interpreterobjects.py
@@ -163,7 +163,7 @@ class RunProcess(MesonInterpreterObject):
                  subdir: str,
                  mesonintrospect: T.List[str],
                  in_builddir: bool = False,
-                 check: bool = False,
+                 check: bool = True,
                  capture: bool = True) -> None:
         super().__init__()
         if not isinstance(cmd, ExternalProgram):
@@ -184,7 +184,7 @@ class RunProcess(MesonInterpreterObject):
                     subdir: str,
                     mesonintrospect: T.List[str],
                     in_builddir: bool,
-                    check: bool = False) -> T.Tuple[int, str, str]:
+                    check: bool = True) -> T.Tuple[int, str, str]:
         command_array = cmd.get_command() + args
         menv = {'MESON_SOURCE_ROOT': source_dir,
                 'MESON_BUILD_ROOT': build_dir,
```

xclaesse:
I think it's like that for historical reasons, changing this would not be backward compatible. But I agree it's not ideal, we should warn when check argument is not set for a few releases, then so the flip eventually.

Volker-Weissmann:
> I think it's like that for historical reasons, changing this would not be backward compatible.

I can understand that.

> we should warn when check argument is not set

Do you want me to write the PR?

eli-schwartz:
I agree that this change would be nice and yes, if you would write the PR that would be great. ;)

eli-schwartz:
Fixed via linked PR.

xclaesse:
FWIW, I checked GStreamer and they have plenty of places where it assumes that check is false by default. I still agree the default should be true, but we'll probably need a long deprecation period because that seems a common thing.

eli-schwartz:
Sure, baby steps for now though. :D

[...]

PR #9304, opened 2021-09-27
Added warning if `check` kwarg of `run_command` is not given.

Volker-Weissmann:
See https://github.com/mesonbuild/meson/issues/9300


test cases/unit/59 introspect buildoptions/meson.build
```diff
@@ -9,7 +9,7 @@ foo = false
 
 executable(target_name, target_src, build_by_default : foo)
 
-r = run_command(find_program('c_compiler.py'))
+r = run_command(find_program('c_compiler.py'), check: false)
 if r.returncode() != 0
   error('FAILED')
```
eli-schwartz, line 14: This one should should probably be check: true, lol

xclaesse: Make sure we still have at least one place where we test with `check: false` and a command that actually fails.

eli-schwartz: `33 run program` does that several times. ` assert(ret.returncode() == 0, 'failed to run python3: ' + ret.stderr())`

If you're looking for one that is expected to error out and raises a testsuite failure if `ret.returncode() == 0`, this doesn't do that either.

Volker-Weissmann: If you run `rg -F "check: false" test\ cases/`, you'll find this pattern at a couple of places:
```
r = run_command(..., check: false)
if r.returncode() != 0
    error('some error')
endif
```
(sometimes with an if clause, sometimes with an assert). This can be removed if we set `check` to `true`, but this would change the error message.
In some places (like `error('FAILED')`), this error message isn't better than the default, but in some other places it is.
I think you should decide which places should keep their custom error messages and which places can use the default error message.

test cases/common/220 fs module/meson.build
```diff
@@ -15,7 +15,7 @@ if not is_windows and build_machine.system() != 'cygwin'
   # * Windows user permissions to create symlinks, and/or Windows in Developer mode
   # so at this time the symlink test is skipped for Windows.
   symlink = meson.current_build_dir() / 'a_symlink'
-  run_command('ln', '-s', meson.current_source_dir() / 'meson.build', symlink)
+  run_command('ln', '-s', meson.current_source_dir() / 'meson.build', symlink, check: true)
```
eli-schwartz, line 18: This is a great error right here!

The test runner configures this test case correctly, and it succeeds in generating, running, and in theory passing. But the test runner ***also*** checks that the backend will correctly force regeneration for build/test, by updating the utime of meson.build.

On this second run, the `ln` program fails because
```
--- stderr ---
/usr/bin/ln: failed to create symbolic link '/tmp/b d19af95f3b/a_symlink': File exists
```

eli-schwartz: (after your latest push)

Note that the additional `-f` flag to ln would make it overwrite the existing symlink rather than erroring.

Volker-Weissmann: Done

Volker-Weissmann:
> Meson PRs cannot include merge commits. Please rebase instead.
> 
> Also the "simple bug fix" commit should be squashed into the commit which it is fixing.

I fought with git for a while and ended up saving a code-diff elsewhere, deleting my local repo, and force-pushing a commit created from the code-diff.

Relevant xkcd: https://xkcd.com/1597/

eli-schwartz:
That sounds complicated... I'd just `git branch newbranch origin/master` and apply the code diff to that, and then force push. No need to delete the repo.

Anyway, changes seem okay, let's see what CI thinks.

Volker-Weissmann:
> That sounds complicated... I'd just `git branch newbranch origin/master` and apply the code diff to that, and then force push. No need to delete the repo.
> 

Then I would have to create a new PR, because this PR wants to merge check_false_warning and not newbranch

eli-schwartz:
Oh! Almost forgot... sorry...

This should get a release notes snippet in `docs/markdown/snippets/`.

eli-schwartz:
Nah, you can push branches like this:

```
git push myfork <currentbranch>:<remotebranch>
```

Or actually, it's a good point that you could/should also set `--set-upstream-to=myfork/check_false_warning`.

Volker-Weissmann:
> Oh! Almost forgot... sorry...
> 
> This should get a release notes snippet in `docs/markdown/snippets/`.

Done

Volker-Weissmann:
> ```
> if result.stdout() == '' and result.stderr() == ''
>   error('No output in stdout and stderr. Did the compiler run at all?')
> endif
> ```
> 
> We _do_ care about this, please don't remove it. :)

Oops, my fault, fixed.

eli-schwartz:
Thanks!

PR #9306, opened 2021-09-28
Setup vsenv even if gcc/clang is in PATH when MESON_FORCE_VSENV is set

xclaesse:
Also setup vsenv only for subcommands that needs it.


jpakkane:
Doing this only for commands that need it is good. The envvar thing, not so much. It's so unfortunate that everything has to be pointlessly broken. :(

xclaesse:
@dcbaker @jpakkane what do you think about this new version that uses CC env instead?

xclaesse:
@dcbaker @jpakkane ok, changed my mind completely, and when with adding `--vsenv` command line argument instead. This also force vs activation when using vs backend. What do you think?

eli-schwartz:
What was wrong with using the environment variable (which is already used for all the other compilers)?

xclaesse:
@eli-schwartz for small details:
- I generally prefer command line over env.
- The whole discussion started because @jpakkane couldn't find how set env var conditionally in github action. Advantage of --vsenv is you can add it unconditionally.
- Feels weird to set CC to a binary that is not in the PATH yet. Feels like hidden trick.
- More extensible, we could some day accept `--vsenv=arm`, etc.

eli-schwartz:
Thanks for clarifying. :)

jrobeson:
The doc snippet doesn't actually describe what happens other than the vsenv being setup. Does it automatically set the compiler to msvc?  I'm purposely not reading the code to see if it does, so i can see if the documentation is complete by itself.

mesonbuild/mesonlib/vsenv.py
```diff
@@ -0,0 +1,100 @@
...
+    if not force:
+        if shutil.which('cc'):
+            return False
+        if shutil.which('gcc'):
+            return False
+        if shutil.which('clang'):
+            return False
+        if shutil.which('clang-cl'):
+            return False
+
```
jrobeson, line 46: are these `which`es still necessary by the time we get this far?

xclaesse: Yes because we still try to setup VS environment even when --vsenv is not specified, but in that case we don't want to activate it when other compilers are available in your PATH.

xclaesse:
> The doc snippet doesn't actually describe what happens other than the vsenv being setup. Does it automatically set the compiler to msvc?

Yes it will uses msvc, unless you also set CC=gcc.exe in env of course.

jrobeson:
@xclaesse  I"m saying that the documentation should be more explicit in saying that. Something like "and will use it as your active compiler"

PR #9232, opened 2021-09-07
mtest: fix test output issues (in console)

trhd:
I was annoyed by e.g. the meson's test subcommand's output's usage of the scissors emoji that never displayed properly in my development environments and set out to improve the look and feel of the test subcommand's output. 

In the commit message I write:

> This change set aims to fix various "issues" seen with the current
> implementation. The changes can be summarized with the following list:
> 
> * Replace emojis and spinners with multiline status displaying the name
>   and running time of each currently running test.
> * The test output (especially in verbose mode or when multiple failing
>   tests' output gets printed out) can get confusing. Try to make the
>   output easier to read and grasp. Most notable change here is the
>   addition of the test number to the beginning of each printed line.
> * Print exit details (i.e. exit code) of the test in verbose mode.
> * Try to make the verbose "live" output from tests to match the look and
>   feel of otherwise produced (verbose) test output.

which quite well describes the changes I believe (hope). A "picture" is however worth a thousand words:

  https://asciinema.org/a/dGOULiN7z9t0OJXrAno570W7T

In my opinion the end result is better that the original. This is however quite a subjective matter and you might disagree...

In the case that people agree with me on the above I still feel that it would be good if other people tried this out as well before merging. AFAIK the "test output" is not really tested via meson's own tests and as I can check only so many corner cases on my own there are bound to be issues that got overlooked. Especially if someone has strong opinions on how the subtests should be reported I'd love to hear back from you.

PR #9307, opened 2021-09-29
Revert "mtest: fix test output issues (in console)"

bonzini:
This reverts commit 5fcb0e6525e2044e0f82bda488a51350e0f7f29f.
The commit is a massive change that should have been split in
separate pieces, and it also removes a few features:

* in verbose mode, subtests are not printed as they happen

* in non-verbose mode the progress report does not include the number of subtests that have been run

* in non-parallel mode, output is batched rather than printed as it happens

Furthermore, none of these changes are not documented in the release
notes.  Revert so that each proposal can be tested, evaluated and
documented individually.


bonzini:
(The revert is #9232; adding @trhd)

PR #9308, opened 2021-09-29
CODEOWNERS: add myself as "meson test" owner

bonzini:
"git shortlog" lists myself as the main contributor to `meson test`. Add myself to CODEOWNERS so that I automatically get subscribed to pull requests and can help with review.


xclaesse:
Thanks, we should do that more often!

eli-schwartz:
Unfortunately this does not actually work:

> The people you choose as code owners must have write permissions for the repository.

A third-party bot surely exists somewhere on GitHub that can implement this properly since GitHub doesn't wanna. I can't fathom that no one else has had this exact problem before.

PR #9313, opened 2021-09-29
clangformat: Only format files tracked by git by default

xclaesse:


jpakkane:
One thing which I have been thinking about is to have two different targets for clang-format:

- `clang-format` which formats all files that have uncommitted changes
- `clang-format-all` which formats all files

Always formatting all files gets slow for large projects.

Dunno if that is people expect or want, though.

xclaesse:
@jpakkane clang-format can check a diff: https://github.com/llvm-mirror/clang/blob/master/tools/clang-format/clang-format-diff.py.

At some point I would like to move that script as a top level meson command to support more arguments. It's a bit weird to be a ninja target, especially since you're not supposed to need a builddir to reformat all your source...

PR #9319, opened 2021-09-30
Vim: match octal and hexadecimal number literals

dkearns:


eli-schwartz:
Can you just use one PR with four commits, rather than opening one PR for each facet of the changes you want to make?

PR #9323, opened 2021-09-30
Vim: improve syntax and ftplugin support

dkearns:
- Highlight octal and hexadecimal numbers
- Use the Boolean highlight group for booleans
- Set b:undo_ftplugin. See :help undo_ftplugin for details.
- Improve the ftplugin with comment formatting, meson_recommended_style config, matchit and browsefilter support.

### DIFF
--- a/data/syntax-highlighting/vim/syntax/meson.vim
+++ b/data/syntax-highlighting/vim/syntax/meson.vim
@@ -136,13 +136,13 @@ endif
 " The default highlight links.  Can be overridden later.
 hi def link mesonStatement	Statement
 hi def link mesonConditional	Conditional
-hi def link mesonRepeat	Repeat
+hi def link mesonRepeat		Repeat
 hi def link mesonOperator	Operator
 hi def link mesonComment	Comment
 hi def link mesonTodo		Todo
-hi def link mesonString	String
-hi def link mesonEscape	Special
-hi def link mesonNumber	Number
+hi def link mesonString		String
+hi def link mesonEscape		Special
+hi def link mesonNumber		Number
 hi def link mesonBuiltin	Function
 hi def link mesonBoolean	Boolean
 if exists("meson_space_error_higlight")


data/syntax-highlighting/vim/ftplugin/meson.vim
```diff
@@ -12,9 +12,28 @@ set cpo&vim
 
 setlocal commentstring=#\ %s
 setlocal comments=:#
+setlocal formatoptions+=croql formatoptions-=t
 
-setlocal shiftwidth=2
-setlocal softtabstop=2
+let b:undo_ftplugin = "setl com< cms< fo<"
+
+if get(g:, "meson_recommended_style", 1)
```
dkearns, line 19: This should probably be documented at :help ft-meson-plugin

tristan957:
What is the best way to install the Meson Vim plugin assuming I am using a vim package manager.

data/syntax-highlighting/vim/ftplugin/meson.vim
```diff
@@ -12,9 +12,30 @@ set cpo&vim
...
+  let b:undo_ftplugin .= "| unlet! b:match_words"
+endif
+
+if (has("gui_win32") || has("gui_gtk")) && !exists("b:browsefilter")
+  let b:browsefilter = "Meson Build Files (meson.build)\tmeson.build\n" .
+	\	       "All Files (*.*)\t*.*\n"
+  let b:undo_ftplugin .= "| unlet! b:browsefilter"
+endif
+
+let b:undo_ftplugin = "setl com< cms< sts< sw<"
```
eli-schwartz, line 38: Seems like this is overwriting the previous contents you added.

data/syntax-highlighting/vim/ftplugin/meson.vim
```diff
@@ -12,9 +12,30 @@ set cpo&vim
...
 
-setlocal shiftwidth=2
-setlocal softtabstop=2
+let b:undo_ftplugin = "setl com< cms< fo<"
+
+if get(g:, "meson_recommended_style", 1)
+  setlocal expandtab
+  setlocal shiftwidth=2
+  setlocal softtabstop=2
+  let b:undo_ftplugin .= " | setl et< sts< sw<"
```
eli-schwartz, line 23: In theory this could probably append to the previous setl, rather than starting a new one... would require actually making sure undo_ settings are defined in the right order, not sure it is worth it.

dkearns: Yes, it becomes more fragile in larger and more complicated plugins.  I updated over a hundred so I was trying to follow a consistent strategy.

eli-schwartz: That's a reasonable approach. :)

data/syntax-highlighting/vim/ftplugin/meson.vim
eli-schwartz: Sure, I think it makes sense to squash it into the commit which obsoleted it as mentioned: "This was retained in error in 695e320"

Issue #9241, opened 2021-09-10
non-root install into non-default prefix shows ldconfig errors


anarazel:
Hi,

I often install programs into separate prefixes, particularly when developing. I'm not using root for that (and the user often won't have the right to sudo). After #8833 that shows errors:
```
Failed to apply ldconfig ...
Standard output:

Standard error:
ldconfig: Can't create temporary cache file /etc/ld.so.cache~: Permission denied
```
The failure to run ldconfig doesn't make the install fail. But it still seems confusing / noisy. And the user can't do anything about not being able to run ldconfig.

@dabrain34, maybe this needs to be refined further? Conceivably permission errors should just be swallowed? It may also be worth to only run ldconfig if any libraries installed - the repro below doesn't and yet ldconfig is being called...

Reproduce with the following:
```
andres@awork3:~$ lsb_release -a
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux bookworm/sid
Release:	unstable
Codename:	sid

andres@awork3:~$ git -C src/meson describe
0.59.0-303-g0fcd696b1

andres@awork3:/tmp/repro-meson-build$ cat /tmp/repro-meson-src/meson.build
project('repro', ['c'],
	version : '0.1',
	meson_version: '>=0.59.99')

andres@awork3:~$ ~/src/meson/meson.py /tmp/repro-meson-src/ /tmp/repro-meson-build/ --prefix /tmp/repro-meson-install
The Meson build system
Version: 0.59.99
Source dir: /tmp/repro-meson-src
Build dir: /tmp/repro-meson-build
Build type: native build
Project name: repro
Project version: 0.1
C compiler for the host machine: ccache cc (gcc 10.3.0 "cc (Debian 10.3.0-9) 10.3.0")
C linker for the host machine: cc ld.bfd 2.37
Host machine cpu family: x86_64
Host machine cpu: x86_64
Build targets in project: 0

repro 0.1

  User defined options
    prefix: /tmp/repro-meson-install

Found ninja-1.10.1 at /usr/bin/ninja

andres@awork3:~$ ~/src/meson/meson.py install -C /tmp/repro-meson-build/ && echo install command succeeded
ninja: Entering directory `/tmp/repro-meson-build'
ninja: no work to do.
Failed to apply ldconfig ...
Standard output:

Standard error:
ldconfig: Can't create temporary cache file /etc/ld.so.cache~: Permission denied

Nothing to install.
install command succeeded

```

[...]

PR #9325, opened 2021-09-30
minstall: Skip ldconfig if not needed

xclaesse:
Fixes: #9241

### DIFF
--- a/mesonbuild/minstall.py
+++ b/mesonbuild/minstall.py
@@ -89,10 +89,12 @@ class DirMaker:
     def __init__(self, lf: T.TextIO, makedirs: T.Callable[..., None]):
         self.lf = lf
         self.dirs: T.List[str] = []
+        self.all_dirs: T.Set[str] = set()
         self.makedirs_impl = makedirs
 
     def makedirs(self, path: str, exist_ok: bool = False) -> None:
         dirname = os.path.normpath(path)
+        self.all_dirs.add(dirname)
         dirs = []
         while dirname != os.path.dirname(dirname):
             if dirname in self.dirs:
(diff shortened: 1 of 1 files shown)


xclaesse:
CC @anarazel @dabrain34

mesonbuild/minstall.py
```diff
@@ -246,14 +248,36 @@ def restore_selinux_contexts() -> None:
...
+def apply_ldconfig(dm: DirMaker) -> None:
     '''
     Apply ldconfig to update the ld.so.cache.
     '''
     if not shutil.which('ldconfig'):
         # If we don't have ldconfig, failure is ignored quietly.
         return
 
+    # Get the list of directories where ldconfig looks for its cache
+    proc, out, err = Popen_safe(['ldconfig', '-v', '-N', '-X'])
```
eli-schwartz, line 260: This is broken on non-Linux at least because -N and -X aren't valid options on e.g. openbsd, freebsd, netbsd. They all do have `-v`, but I haven't checked their output and I'm nervous that this human-readable output is going to have significant formatting differences...

anarazel:
Hm. Perhaps we instead could just check if running as root? That'd get around needing to handle ldconfig - and it's not like ldconfig is likely to work without a problem for a non-root user, even if one configured it for some per-user install directories.

mesonbuild/minstall.py
xclaesse: Pfff, sad :(

What I liked with this PR is it also prevent wasting time when the project does not install shared libraries.

I guess for now we could just check `os.geteuid() == 0`, or do you have a smarter idea?

eli-schwartz: I'm pretty sure at the time of the original ldconfig implementation I tried and failed to think of anything.

Checking for root is also "maybe technically wrong -- what if the user has ldconfig set up to index non-root directories" but that is getting pretty niche so idk.

If you want to avoid wasting time when the project does not install shared libraries, you could make ldconfig be an attribute set by `library(..., install: true)`?

xclaesse:
Ok, changed this PR to instead only use '-v' option, so it will always try to update cache but if it fails it checks if the error was harmless and can be ignored.

PR #9326, opened 2021-09-30
Improve wraptool search

danielcjacobs:
[Wrapdb issue #148](https://github.com/mesonbuild/wrapdb/issues/148) explains that the wraptool search isn't very helpful, because it only checks if a wrap starts with the search substring.  
This PR searches the entire wrap name for the substring.


xclaesse:
Bonus point if you can also search in all `dependency_names`. For example, searching `glib-2.0` should return the `glib` wrap.

eli-schwartz:
In that case it should print something which describes how the dependency got resolved so there is visibility into how "meson wrap search foo" returned a hit for "bar".

mesonbuild/wrap/wraptool.py
```diff
@@ -71,8 +71,11 @@ def search(options: 'argparse.Namespace') -> None:
     name = options.name
     releases = get_releases()
     for p in releases.keys():
-        if p.startswith(name):
+        if p.find(name) != -1:
             print(p)
+        for dep in releases.get(p).get('dependency_names'):
```
xclaesse, line 76: Better do `for p, info in releases.items()` and then `for dep in info.get('dependency_names', [])`.

mesonbuild/wrap/wraptool.py
```diff
@@ -71,8 +71,11 @@ def search(options: 'argparse.Namespace') -> None:
...
+            if dep.find(name) != -1:
+                print("Dependency {} found in wrap {}" % (dep, p))
```
xclaesse, line 78:
```suggestion
                print(f'Dependency {dep} found in wrap {p}')
```

mesonbuild/wrap/wraptool.py
xclaesse: Maybe should not look in dependencies if the name matched already, which probably would happen most of the time.

danielcjacobs:
Can't figure out what to edit to fix [this error](https://github.com/mesonbuild/meson/pull/9326/commits/4613c90c589505d7e5cee74dc53901960dfd7cc6)

xclaesse:
@danielcjacobs you mean the Codecov message? It means you should write a unit test for that code path. Would be nice to add `unittests/platformagnostictests.py`, but I wouldn't consider it blocker tbh.

mesonbuild/wrap/wraptool.py
```diff
@@ -70,9 +70,13 @@ def list_projects(options: 'argparse.Namespace') -> None:
...
     releases = get_releases()
-    for p in releases.keys():
-        if p.startswith(name):
+    for p, info in releases.items():
+        if p.find(name) != -1:
             print(p)
+        else:
+            for dep in info.get('dependency_names', []):
+                if dep.find(name) != -1:
+                    print("Dependency {dep} found in wrap {p}")
```
eli-schwartz, line 79: I don't need a unittest to review this -- but it's not an f-string so this will not do variable replacement.

eli-schwartz (changes requested):

PR #9329, opened 2021-09-30
Add type checking and annotations to the i18n module

dcbaker:
The main reason I'm doing this is that these are CustomTarget wrappers, and as such having full type checking is going to be important to getting the layering violations between the interpreter and build modules sorted out


mensinda (approved):
LGTM when the CI is green

PR #9330, opened 2021-09-30
dependency: Allow searching for multiple names

xclaesse:


tristan957:
Nice. Super easy solution for this problem.

dcbaker (approved):

eli-schwartz:
This calls out the example of listing two dependency names, one for pkg-config and one for cmake. I've hit this before and had pondered the idea of imitating `get_variable(foo, cmake: 'FOO')` by having `dependency('foo', cmake: 'FOO')`, but this seems a lot more elegant, since it covers both that use case and the use case of checking for multiple independent providers (e.g. checking for crypto support via either `dependency('libcrypto') or `dependency('nettle')`).

eli-schwartz (approved):

PR #9331, opened 2021-09-30
Add type annotations and typed_kwargs for external_project

dcbaker:
~~This is based on the i18n series, so that should land first.~~ i18n changes are in.

This is more work toward being able to refactor the CustomTarget argument handling out of the build module and into the interpreter.


xclaesse:
Lgtm

mensinda (approved):
LGTM, from what I can tell, though @xclaesse would know more here.

Issue #7702, opened 2020-09-06
PEP 632: Proposed deprectation of `distutils`


FRidh:
There is a [PEP proposed](https://discuss.python.org/t/pep-632-deprecate-distutils-module/) for deprecating `distutils`. Meson uses `distutils` in the `python` module for building extension modules, so this PEP may be of interest to you.

lazka:
distutils is only used to decide if extensions should link against libpython to follow the platform defaults (Debian patched to to not link). With Python 3.8 upstream Python default changed to not link to libpython as default, so if distutils ever gets removed meson can just default to not linking: https://docs.python.org/3/whatsnew/3.8.html#debug-build-uses-the-same-abi-as-release-build

eli-schwartz:
Note that in #9288 we added an additional reliance on distutils (in the same exact introspection command that already imports distutils). In addition to detecting whether to link against libpython, we now also use distutils to detect whether sysconfig is completely crippled by the vendor and unable to tell the truth about install paths.

Handling libpython linking could be done by a try/except that reacts to an ImportError by saying "clearly this must be `sys.version_info >= (3, 8)` so the answer is don't link".

Handling debian paths cannot be resolved other than by waiting on debian to fix things.

As we get closer to the final removal of distutils in 3.12 we will need to both consider that try/except for libpython, and keep an eye on the debian situation.

tacaswell:
`distutils` has now been pulled out of the main branch of CPython and meson is failing to detect it as a valid Python:


```
$ meson --version
0.64.0
$ python --version
Python 3.12.0a1+
```

adapted from numpy's meson branch:

```
project(
  'test',
  'c',
  version: '0',
  license: 'BSD-3',
  meson_version: '>= 0.64.0',
)

cc = meson.get_compiler('c')

# https://mesonbuild.com/Python-module.html
py_mod = import('python')
py = py_mod.find_installation(pure: false)
py_dep = py.dependency()

```

```
$ meson setup build --prefix=/home/tcaswell/.virtualenvs/bleeding/lib/python3.12/site-packages
The Meson build system
Version: 0.64.0
Source dir: /home/tcaswell/source/p/numpy/numpy
Build dir: /home/tcaswell/source/p/numpy/numpy/build
Build type: native build
Project name: test
Project version: 0
C compiler for the host machine: ccache cc (gcc 12.2.0 "cc (GCC) 12.2.0")
C linker for the host machine: cc ld.bfd 2.39.0
[...]

PR #9335, opened 2021-10-01
modules/python: detect debian distutils missing and show an error

FFY00:


FFY00:
cc @eli-schwartz

mesonbuild/modules/python.py
```diff
@@ -282,6 +282,21 @@ def set_env(name, value):
 import sys
 import distutils.command.install
```
xclaesse, line 283: You need to move that I think.

FFY00: Ah, yes.

mesonbuild/modules/python.py
```diff
@@ -390,13 +414,18 @@ def sanity(self) -> bool:
...
-            variables = info['variables']
-            info['suffix'] = variables.get('EXT_SUFFIX') or variables.get('SO') or variables.get('.so')
-            self.info = T.cast('PythonIntrospectionDict', info)
-            self.platlib = self._get_path('platlib')
-            self.purelib = self._get_path('purelib')
-            return True
+        if info is not None:
+            if 'error' in info:
+                assert isinstance(info['error'], str)
+                mlog.error('Python interpreter introspection failed: %s' % info['error'])
```
eli-schwartz, line 420: Please use f-strings (or, when the complexity penalty is too great, the `str.format()` method).

The assert doesn't seem to be performing any debug sanity checks, so it can probably be dropped.

eli-schwartz (commented):
In general I like the look of this. But I'm not sure we should be logging an error directly in the Python ExternalProgram constructor, since it may be requested with `required: false` and hence not be an error after all.

It should probably either:
- be logged as mlog.log and rely on the later checks for sanity() return value to promote to an error
- set an attribute on self and look that up later to raise the error message as mlog.error

mesonbuild/modules/python.py
FFY00: The assert is to make sure the nobody screw up when editing the introspection script, but I can remove it if you want.

eli-schwartz: Actually that may speak to a greater need: adding a test case that actually tries to trigger this code path and testing that it produces the correct error message.

On the other hand -- if it's not a string what will it be? It is loaded as json there are only so many options...

eli-schwartz: I don't have strong opinions on the assert, but as a project standard we have purged percent formatting in favor of f-strings wherever it doesn't have major readability costs (now that we require a minimum python which supports f-strings) and the .format() method otherwise.

FFY00: > Actually that may speak to a greater need: adding a test case that actually tries to trigger this code path and testing that it produces the correct error message.

We would have to keep the tests in updated to make sure they cover all code paths in the introspection command.

> On the other hand -- if it's not a string what will it be? It is loaded as json there are only so many options...

Could be anything. I just put the assert there as a sanity check to avoid obscure error. Yes, this is me being preventive, but it's cheap, so...

> I don't have strong opinions on the assert, but as a project standard we have purged percent formatting in favor of f-strings wherever it doesn't have major readability costs (now that we require a minimum python which supports f-strings) and the .format() method otherwise.

Of course. I don't like nesting quotes in f-strings, so I went with `%` as that's what the neighboring code did, but I can definitely change.

FFY00:
> be logged as mlog.log and rely on the later checks for sanity() return value to promote to an error

I would prefer this. If my Python interpreter is bad and consequently the Python specific bits of the project gets disabled, I want to know why.

mesonbuild/modules/python.py
eli-schwartz: Nesting quotes? Do you mean like `f'{some string!r}'`  to guarantee that the printed content itself has quite characters?

FFY00:
```python
f'something: {info["error"]}'
```

eli-schwartz: Also it is true we do have some percent formatted strings still... I would like to get rid of all of them but I mainly drove the upgrade using pyupgrade which is pretty good but still heuristics based and thus not perfect. The vast majority were changed though. :D

FFY00: Anyway, I change to `str.format`, let me know if you'd like me to change to an f-string.

eli-schwartz:
That's a reasonable perspective and I think I agree.

cclauss:
Python 3.10 (scheduled to be released on Monday) deprecates ditsutils as discussed in https://www.python.org/dev/peps/pep-0632/

FFY00:
Yes, we are aware, but unfortunately we don't have a migration plan right now due to Debian patching the interpreter, but not sysconfig. As distutils will only be removed in 3.12 we still have some time to coordinate and update the code.

eli-schwartz:
The deprecation of distutils is unrelated to this PR as we already depended on distutils for entirely unrelated reasons. For a long time, meson has considered the lack of distutils to be a fatal error. This PR merely changes the formatting of the error message that is used when meson fatally exits.

Please take all discussion of distutils deprecation to #7702.

And distutils is not deprecated right now! It is only deprecated in python 3.10! This is important, because it means we have an awful long time before it is actually removed, which means that ***Debian's failure to have a legitimate python*** is not a valid reason for us to make meson ***not*** work on Debian in order to punish ordinary Debian users for the fact that Debian developers do not comply with the PEP.

It's not a bug to rely on deprecated things. It is a concern, yes, but that just means we need to balance the pros and cons of need vs. consequences.

mesonbuild/modules/python.py
```diff
@@ -280,7 +280,21 @@ def set_env(name, value):
...
 import sys
-import distutils.command.install
+
+def debian_distutils_missing():
+    # Debian partially splits the distutils module from the python3, they
+    # keep distutils and distutils.version in python3 and put the rest of
+    # the submodules in python3-distutils
+    try:
+        import distutils
+        import distutils.version
```
eli-schwartz, line 290: The second import includes the first import, we only need the second one.

eli-schwartz: The check could be even simpler, I think.
```python
try:
    import distutils.version
    import distutils.core
except ModuleNotFoundError as e:
    if e.name == 'distutils.core':
        return True # distutils was found, and distutils.version successfully imported, but core did not
return False
```

Also ModuleNotFoundError > ImportError here, since we specifically want to detect "a module could not be located" rather than "the import statement has troubles trying to load a module"

unittests/pythontests.py
```diff
@@ -80,3 +82,11 @@ def test_versions(self):
         with self.assertRaises(unittest.SkipTest):
             self.init(testdir, extra_args=['-Dpython=dir'])
         self.wipe()
+
+    @unittest.mock.patch('mesonbuild.mesonlib.Popen_safe', return_value=(None, '{"error": "hello!"}', None))
+    @unittest.mock.patch('mesonbuild.mlog.log')
+    def test_introspection_error(self, mlog_log, Popen_safe):
+        python = PythonExternalProgram('some-python', ['some-python'])
+
+        assert not python.sanity()
```
eli-schwartz, line 91: Hmm, shouldn't this be a unittest assertFalse, not a regular assert?

cclauss: When tests are [run with `pytest`](https://github.com/mesonbuild/meson/search?q=pytest), either is acceptable.  In fact, this simplicity without loss of debugging information is [pytest's number one feature](https://docs.pytest.org/en/latest/#id1].

eli-schwartz: Well yes, but if you don't have both pytest and pytest-xdist then the tests get run without pytest, just unittest, and that is a supported configuration...

eli-schwartz:
Due to the previous PR which I just merged, this now :crossed_swords: conflicts.

jpakkane:
:crossed_swords: conflicts.

eli-schwartz:
Seems like the rc is being delayed until tomorrow anyway due to time constraints. :) Looking forward to the update for this PR which should still make it in.

FFY00:
@eli-schwartz rebased on `master`.

unittests/pythontests.py
FFY00: It's the same thing in this case, these functions just exist to provide a nicer error output. pytest rewrites the bytecode to replace the asserts with something similar in order to achieve that. It makes sense when you are comparing lists for eg, but as we are just checking a bool, is pretty much the same.

eli-schwartz: If it's "pretty much the same thing" to you, it shouldn't be a problem to use unittest asserts for the sake of those odd people using unittest instead of pytest to run this testsuite :)

eli-schwartz:
What happened?

PR #9336, opened 2021-10-01
Disable codecov comment

xclaesse:


dcbaker:
At that point, let's just delete codecov. The reports are so hilariously wedding they're not useful

xclaesse:
To be fair, it's actually more customizable: https://docs.codecov.com/docs/pull-request-comments maybe there is a middle ground to find, removing it completely could be a bit extreme. Does someone actually look at those reports? (I confess I don't)

eli-schwartz:
IIRC @mensinda was the one who implemented this.

I certainly don't look at these reports, they do bother me but not enough to argue against it as long as someone else does find it useful. My priority was, and remains, making it not cause CI to fail, so that is enough to make me happy.

xclaesse:
@eli-schwartz @dcbaker I'm getting really bored by those comments and annotation that nobody look at, let's disable that. The report can still be see manually in the "Checks" tab.

eli-schwartz:
AFAICT from the codecov documentation, this change will only disable the "added lines not covered by tests" boxes in the "Files" view, and doesn't touch the posted comment at all. I would have assumed you'd want the exact opposite.

xclaesse:
@eli-schwartz There are 2 changes, I think `comment: false` should be disabling the comment. Seems to work since I don't see a comment on this PR...

xclaesse:
Actually, according to https://docs.codecov.com/docs/pull-request-comments it seems it should be toplevel key. Changed.

xclaesse:
Since nobody is interested in those reports, let's merge this.

eli-schwartz:
+1

PR #9339, opened 2021-10-01
Structured Sources

dcbaker:
This is a reworked version of https://github.com/mesonbuild/meson/pull/8775, with the main complaints resolved.

The purpose of this series is to provide a mechanism for languages in which filesystem layout is relevant to ensure that generated sources end up in the right place. To do this, one describes the layout they want to a `structrued_sources` object, which the backend converts (if necessary) into a series of copy commands, to build the correct directory structure in the build directory. This structure is placed in the private directory of the target, to eliminate collisions. Then the actual compile is done on the copied sources rather than on the original sources. This is done at build time to avoid reconfiguration when any of the structured sources are changed.

### DIFF
--- a/docs/markdown/Rust.md
+++ b/docs/markdown/Rust.md
@@ -1,0 +1,56 @@
+---
+title: Rust
+short-description: Working with Rust in Meson
+---
+
+# Using Rust with Meson
+
+## Mixing Rust and non-Rust sources
+
+Meson currently does not support creating a single target with Rust and non Rust
+sources mixed together, therefore one must compile multiple libraries and link
+them.
+
+```meson
+rust_lib = static_library(
+    'rust_lib',
+    sources : 'lib.rs',
+    ...
+)
+
+c_lib = static_library(
+    'c_lib',
+    sources : 'lib.c',
+    link_with : rust_lib,
+)
+```
+This is an implementation detail of Meson, and is subject to change in the future.
+
+## Mixing Generated and Static sources
+
+*Note* This feature was added in 0.62
+
+You can use a [[structured_source]] for this. Structured sources are a dictionary
+mapping a string of the directory, to a source or list of sources.
+When using a structured source all inputs *must* be listed, as Meson may copy
+the sources from the source tree to the build tree.
+
+Structured inputs are generally not needed when not using generated sources.
+
+As an implementation detail, Meson will attempt to determine if it needs to copy
+files at configure time and will skip copying if it can. Copying is done at
+build time (when necessary), to avoid reconfiguring when sources change.
+
+```meson
+executable(
+    'rust_exe',
+    structured_sources(
+        'main.rs',
+        {
+            'foo' : ['bar.rs', 'foo/lib.rs', generated_rs],
+            'foo/bar' : [...],
+            'other' : [...],
+        }
+    )
+)
+```
--- a/docs/markdown/snippets/structured_sources.md
+++ b/docs/markdown/snippets/structured_sources.md
@@ -1,0 +1,26 @@
+## structured_sources()
+
+A new function, `structured_sources()` has been added. This function allows
+languages like Rust which depend on the filesystem layout at compile time to mix
+generated and static sources.
+
+```meson
+executable(
+  'main',
+  structured_sources(
+    'main.rs,
+    {'mod' : generated_mod_rs},
+  )
+)
+```
+
+Meson will then at build time copy the files into the build directory (if
+necessary), so that the desired file structure is laid out, and compile that. In
+this case:
+
+```
+root/
+  main.rs
+  mod/
+    mod.rs
+```
--- a/docs/yaml/functions/structured_sources.yaml
+++ b/docs/yaml/functions/structured_sources.yaml
@@ -1,0 +1,21 @@
+name: structured_sources
+returns: structured_src
+since: 0.62.0
+description: |
+  Create a StructuredSource object, which is opaque and may be passed as a source
+  to any build_target (including static_library, shared_library, executable,
+  etc.). This is useful for languages like Rust, which use the filesystem layout
+  to determine import names. This is only allowed in Rust targets, and cannot be
+  mixed with non structured inputs.
+
+posargs:
+  root:
+    type: list[str | file | custom_tgt | custom_idx | generated_list]
+    description: Sources to put at the root of the generated structure
+
+optargs:
+  additional:
+    type: dict[str | file | custom_tgt | custom_idx | generated_list]
+    description: |
+      Additional sources, where the key is the directory under the root to place
+      the values
--- a/docs/yaml/objects/structured_src.yaml
+++ b/docs/yaml/objects/structured_src.yaml
@@ -1,0 +1,3 @@
+name: structured_src
+long_name: Structured Source
+description: Opaque object returned by [[structured_sources]].
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -1176,6 +1176,8 @@ class NinjaBackend(backends.Backend):
         self.add_rule(NinjaRule('CUSTOM_COMMAND_DEP', ['$COMMAND'], [], '$DESC',
                                 deps='gcc', depfile='$DEPFILE',
                                 extra='restat = 1'))
+        self.add_rule(NinjaRule('COPY_FILE', self.environment.get_build_command() + ['--internal', 'copy'],
+                                ['$in', '$out'], 'Copying $in to $out'))
 
         c = self.environment.get_build_command() + \
             ['--internal',
@@ -1649,6 +1651,16 @@ class NinjaBackend(backends.Backend):
                     generated_sources[ssrc] = mesonlib.File.from_built_file(gen.get_subdir(), ssrc)
 
         return static_sources, generated_sources, cython_sources
+
+    def _generate_copy_target(self, src: 'mesonlib.FileOrString', output: Path) -> None:
+        """Create a target to copy a source file from one location to another."""
+        if isinstance(src, File):
+            instr = src.absolute_path(self.environment.source_dir, self.environment.build_dir)
+        else:
+            instr = src
+        elem = NinjaBuildElement(self.all_outputs, [str(output)], 'COPY_FILE', [instr])
+        elem.add_orderdep(instr)
+        self.add_build(elem)
 
     def generate_rust_target(self, target: build.BuildTarget) -> None:
         rustc = target.compilers['rust']
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -54,6 +54,7 @@ if T.TYPE_CHECKING:
     from .._typing import ImmutableListProtocol
     from ..linkers import DynamicLinker, StaticLinker
     from ..compilers.cs import CsCompiler
+    from ..interpreter.interpreter import SourceOutputs
 
 
 FORTRAN_INCLUDE_PAT = r"^\s*#?include\s*['\"](\w+\.\w+)['\"]"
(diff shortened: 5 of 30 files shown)


dcbaker:
@rgommers I'd like to have this work for cython too before it's merged, but I'm struggling to write a test case so that I can get started. Do you think you could at least help point me in the right direction?

rgommers:
> @rgommers I'd like to have this work for cython too before it's merged, but I'm struggling to write a test case so that I can get started. Do you think you could at least help point me in the right direction?

Yes of course. There are three file types that matter: `__init__.py`'s and `.pxi`/`.pxd` files. Here is a relevant subset of the SciPy code with comments:

`scipy/meson.build`:
```python
# Needed to trick Cython, it won't do a relative import outside a package.
# Every Cython extension built from a generated `.pyx` file needs to rely on this,
# as well as on the `__init__.py` files in subdirs between the root one we're copying
# here and the subdir it's in.
_cython_tree = custom_target('_cython_tree',
  output : [
    '__init__.py',
    'linalg.pxd',
    'special.pxd',
  ],
  input : [
    '__init__.py',
    'linalg.pxd',
    'special.pxd',
  ],
  command : ['cp', '@INPUT@', '@OUTDIR@'],
)
```

Codegen in `scipy/special/meson.build`:
```python
cython_special = custom_target('cython_special',
  output : [...
            '_ufuncs.pyi',
            'cython_special.pyx',
            'cython_special.pxd'
           ],
  input : ['_generate_pyx.py', 'functions.json', 'add_newdocs.py'],
  command : [py3, '@INPUT0@', '-o', '@OUTDIR@'],
)
```

`scipy/stats/meson.build` (uses `cython_special`):
```python
# Relies on cython_special, which ends up in the build dir. Hence we need ` '-I', '@BUILD_ROOT@'` here.
# (note: I need to double check if this `custom_target` is still needed, may have to do with the now
#        solved bug about `cython_special` not being indexable when passed to `extension_module`)
_stats_c = custom_target('_stats_c',
  output: '_stats.c',
  input: [
    '_stats.pyx',
    _cython_tree,
    cython_special,
  ],
  command: [cython, '-3', '--fast-fail',
            '@INPUT0@', '-o', '@OUTPUT@', '-I', '@BUILD_ROOT@'],
)

_stats = py3.extension_module('_stats',
  _stats_c,
  dependencies : [py3_dep],
  install : true,
  subdir : 'scipy/stats'
)
```

The `__init__.py` files should always be copied over I believe (without listing them in structured sources), and that logic is easy to implement. The `.pxd`/`.pxi` files need to be explicitly declared I'd think.

I believe the above covers most/all of what is relevant. Please let me know if it's not clear or if I should try and turn this into a self-contained test case.

dcbaker:
@rgommers: I've added a cython test, and it compiles but doesn't run. Maybe you can help me, I'm bashing my head against the wall, the merge window closes Sunday, and I'd *really* like to get this in before that.

eli-schwartz:
You have created:
```
builddir
├── lib.cpython-310-x86_64-linux-gnu.so
└── lib.cpython-310-x86_64-linux-gnu.so.p
    ├── meson-generated_structured_gen_gen.pyx.c.o
    ├── meson-generated_structured_lib.pyx.c.o
    └── structured
        ├── gen
        │   └── gen.pyx
        ├── lib.pyx
        └── lib.pyx.c
```

Then you add builddir to the PYTHONPATH, run a script that imports `lib`, which works, and then lib itself tries to do ` from gen import generated` however gen.cpython-310-x86_64-linux-gnu.so is never created. The object file built from gen.pyx is being merged into the `lib` extension_module.

It seems like the problem is that the structured source is being treated as a file layout for compiling `lib`, when what you actually want is a file layout for compiling both `lib` and `gen` as well as syncing an entire tree of pure python source code into a directory not named `*.p/` suitable for public access.

dcbaker:
So, for the moment I've dropped cython support, and I'll go at that again in a follow up, as it's going to require some thought on how it's going to work, and I'm not sure what that needs to look like.

rgommers:
> So, for the moment I've dropped cython support, and I'll go at that again in a follow up, as it's going to require some thought on how it's going to work, and I'm not sure what that needs to look like.

That makes sense. I think we're good with the workaround we merged in https://github.com/scipy/scipy/pull/15407. Maybe we should add that as a Meson test and example in the docs?

dcbaker:
> That makes sense. I think we're good with the workaround we merged in https://github.com/scipy/scipy/pull/15407. Maybe we should add that as a Meson test and example in the docs?

That is an issue I hadn't thought of before. The issue is that you have dependencies that only the pyx -> c step depends on, but not the c -> obj step? That sounds like a bug we should fix.

eli-schwartz:
That issue has been discussed before. We need depfile support in the cython tool, which was recently contributed to cython by a scipy contributor but is only in cython git master.

blink1073:
>  We need depfile support in the cython tool

I believe that is available in Cython [0.29.27](https://github.com/cython/cython/blob/0.29.x/CHANGES.rst#02927-2022-01-28)

eli-schwartz:
Oh nice, didn't realize it was released already. I'll get to work on integrating support for that.

dcbaker:
It's only in cythonize unfortunately, so not useful for us :(

dcbaker:
@eli-schwartz: it sounds like what they need though isn't depfile, but something like `transpile_depends : ` (like `depends` but for the transpile step. Is that right @rgommers?

dcbaker:
@jpakkane, since this is a language change feature, I assume you'd like to look at it

mesonbuild/backend/ninjabackend.py
```diff
@@ -1670,6 +1706,13 @@ def generate_rust_target(self, target: build.BuildTarget) -> None:
         orderdeps: T.List[str] = []
 
         main_rust_file = None
+        if target.structured_sources:
+            if target.structured_sources.needs_copy(target):
+                _ods, main_rust_file = self.__generate_compile_structure(target)
+                orderdeps.extend(_ods)
+            else:
+                main_rust_file = target.structured_sources[0][''][0]
```
tristan957, line 1714: Would be nice if the magical `''` was saved as a constant somewhere.

mesonbuild/build.py
```diff
@@ -452,10 +452,51 @@ def get_outputs(self, backend: 'Backend') -> T.List[str]:
             for source in self.get_sources(self.srclist, self.genlist)
         ]
 
+
+@dataclass(eq=False, order=False, )
```
tristan957, line 456: Drop the trailing comma and space? Not sure what the style is here.

mesonbuild/build.py
```diff
@@ -1571,6 +1629,16 @@ def uses_rust(self) -> bool:
...
                 return True
+        elif self.structured_sources:
+            for v in self.structured_sources.sources.values():
+                for s in v:
+                    if isinstance(s, (str, File)):
+                        if s.endswith('.rs'):
+                            return True
+                    else:
+                        for ss in s.get_outputs():
+                            if ss.endswith('.rs'):
```
tristan957, line 1640: Should all the outputs end in .rs?

mesonbuild/interpreter/interpreter.py
```diff
@@ -2076,6 +2078,29 @@ def func_install_symlink(self, node: mparser.BaseNode,
...
+            self, node: mparser.BaseNode,
+            args: T.Tuple[object, T.Optional[T.Dict[str, object]]],
+            kwargs: 'TYPE_kwargs') -> build.StructuredSources:
+        valid_types = (str, mesonlib.File, build.GeneratedList, build.CustomTarget, build.CustomTargetIndex, build.GeneratedList)
+        sources: T.Dict[str, T.List[T.Union['mesonlib.FileOrString', 'build.GeneratedTypes']]] = collections.defaultdict(list)
+
+        for arg in mesonlib.listify(args[0]):
+            if not isinstance(arg, valid_types):
+                raise InvalidArguments(f'structured_sources: type "{type(arg)}" is not valid')
+            sources[''].append(arg)
```
tristan957, line 2095: Good place for the constant suggested above.

mesonbuild/build.py
dcbaker: it is, "there used to be another value here, but it got removed :)

test cases/rust/17 structured sources/meson.build
```diff
@@ -0,0 +1,28 @@
...
+  input : 'src/foo.rs.in',
+  output : 'mod.rs',
+  configuration : {'message' : 'Hello, World!'},
+)
+
+conf_file = executable(
+  'main_conf_file',
+  structured_sources(
+    'src/main.rs',
+    {'foo' : [foo_mod_rs]},
```
tristan957, line 13: Maybe useful to add a usage of `''` in the test.

docs/markdown/snippets/structured_sources.md
```diff
@@ -0,0 +1,25 @@
+## structured_sources()
+
+A new function, `structured_sources()` has been added. This function allows
+languages like Rust and Cython which depend on the filesystem layout at compile
```
tristan957, line 4: May want to remove any usage of Cython from this PR since I think you took it out :(.

mesonbuild/build.py
dcbaker: No, this just checks that *any* file is a rust file. We have a sort of odd restriction that you can't mix rust and non-rust targets. Which doesn't really make sense. You can compile .c files to .o files, and link them with rust targets, we'd just need to generate a separate archiver step with Rust. It's on my list of things to do.

test cases/rust/17 structured sources/meson.build
dcbaker: oh, that's an interesting point. I didn't meant to expose the option of '' in meson DSL. Adding an error for that.

tristan957: Oh whoops. I misunderstood that then :).

dcbaker: That's a good catch. I've added an error and a test for said error.

mesonbuild/interpreter/interpreter.py
dcbaker: I did away with one magical use with a method, so this is the only use left. I'm disinclined to add a constant for a single use, what do you think?

docs/markdown/snippets/structured_sources.md
```diff
@@ -0,0 +1,25 @@
...
+A new function, `structured_sources()` has been added. This function allows
+languages like Rust which depend on the filesystem layout at compile time to mix
+generated and static sources.
+
+```meson
+executable(
+  'main',
+  structured_sources(
+    'main.rs,
+    mod : generated_mod_rs,
```
tristan957, line 12: Might be useful to put `mod` in quotes.

docs/yaml/functions/structured_sources.yaml
```diff
@@ -0,0 +1,21 @@
+name: structured_source
+returns: structured_src
+since: 0.62.0
+description: |
+  Create a StructuredSource object, which is opaque and may be passed as a source
+  to any build_target (including static_library, shared_library, executable,
+  etc.). This is useful for languages like Rust, which use the filesystem layout
+  to determine import names. This is only allowed in Rust targets, and cannot be
+  mixed non structured inputs
```
tristan957, line 9: ending period

docs/markdown/snippets/structured_sources.md
dcbaker: Oh, even better. That's still got the old *non working* syntax :(

docs/yaml/objects/structured_src.yaml
```diff
@@ -0,0 +1,3 @@
+name: structured_src
+long_name: Structured Source
+description: Opaque object returned by [[structured_source]].
```
tristan957, line 3: structured_sources?

mesonbuild/interpreter/interpreter.py
tristan957: I'll leave it up to others :).

I was envisioning something like

```python
class StructuredSource:
    ROOT_MODULE = ''

    def first_file(self) -> T.Union[str, File, CustomTarget, CustomTargetIndex, GeneratedList]:
        """Get the first source in the root
        :return: The first source in the root
        """
        return self.sources[ROOT_MODULE][0]
```

tristan957:
Think that is all I could find. Implementation looks pretty good to me :).

jpakkane:
As this is a notable feature that has been long underway, I'll grant an exception so that this can be merged after the actual DL in case it can't get reviewed in time.

docs/markdown/Rust.md
```diff
@@ -0,0 +1,56 @@
...
+)
+```
+This is an implementation detail of Meson, and is subject to change in the future.
+
+## Mixing Generated and Static sources
+
+*Note* This feature was added in 0.60
+
+You can use a "structured source" for this. Structured sources are a dictionary
+mapping a string of the directory ('' means root), to a source or list of sources.
```
jpakkane, line 34: Please explain what "root" means here, specifically. It is a term with many potential meanings.

jpakkane:
Featurewise this seems like a useful addition so approval on that front.

I have not looked too deeply in the implementation yet, but it seems that there should be a failing test case when using two different structured sources that define the same file in two different structured sources. Something like:

```meson
executable(...,
    structured_sources({'subdir': 'clashing.rs'}),
    structured_sources({'subdir': 'other/clashing.rs'})
)
```

Either there can only be one structured sources per target (which seems a bit limiting) or it should validate all the files upfront.

dcbaker:
Only one structured_source is allowed to be passed, but I can add a test to ensure that passing two structured_sources fails.

dcbaker:
@jpakkane I've added the validation and tests you wanted

PR #9342, opened 2021-10-03
install_man: add support for building man pages via internal custom_target

### LINKED ISSUE #1550: install_man does not accept target output

eli-schwartz:
This is completely terrible, because really you should use a custom_target and then feed it to install_man. However, that has been decided to never happen. So, reluctantly, if install_man won't accept a custom_target, then let install_man copy the entire interface and internally make one.

This solves a massive usability problem where custom targets with install_dir are strictly inferior to install_man() which automatically handles sectioned or localized subdirectories of mandir, and results in nontrivial code to painfully work around this. One alternative approach was to hack in @MANDIR@ support to install_dir itself (which is itself an inconsistent interface), but after adding a locale kwarg to install_man it is no longer possible to infer all information from the filename alone, so that's a total wash.

Fixes: #1550

### DIFF
--- a/mesonbuild/interpreter/interpreter.py
+++ b/mesonbuild/interpreter/interpreter.py
@@ -231,6 +231,21 @@ permitted_dependency_kwargs = {
     'version',
 }
 
+CT_MAN_COMMON_KWARGS = [
+    CT_INPUT_KW,
+    DEPENDS_KW,
+    DEPEND_FILES_KW,
+    DEPFILE_KW,
+    ENV_KW.evolve(since='0.57.0'),
+    INSTALL_MODE_KW.evolve(since='0.47.0'),
+    OVERRIDE_OPTIONS_KW,
+    KwargInfo('build_always', (bool, type(None)), deprecated='0.47.0'),
+    KwargInfo('build_always_stale', (bool, type(None)), since='0.47.0'),
+    KwargInfo('feed', bool, default=False, since='0.59.0'),
+    KwargInfo('capture', bool, default=False),
+    KwargInfo('console', bool, default=False, since='0.48.0'),
+]
+
 class Interpreter(InterpreterBase, HoldableObject):
 
     def __init__(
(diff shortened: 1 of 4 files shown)


jpakkane:
An alternative would be to have a `custom_man_target` that takes the same arguments as `custom_target` (except `install_dir`), validates the output name, computes the mandir and then calls `custom_target` internally.

eli-schwartz:
That is almost the same thing as this PR, except for complaining when people add useless arguments but don't specify a command.

On the other hand it adds an entirely new function which is rather annoying on its own. Would you prefer this?

jpakkane:
Well that's the thing: I don't really know. OTOH having generation functionality in `install_man` is nonstandard because none of the other ones have it. On the other hand additional functions are not nice either, for exactly the reasons you mention.

Do other people have opinions? Or better suggestions for exposing this functionality?

jpakkane:
Well how about this then:

```meson
custom_target(...,
    install_true,
    is_manpage: true) # Defining both this and an install_dir would be an error
```

dcbaker:
I prefer the `is_man` or my previous `install_dir : '@MANDIR@'` approach, personally. either that or we could do something like add a `unix` module with a `man_target` method. 🤷🏻

eli-schwartz:
@dcbaker this despite the fact that my commit message references your own decision that `@MANDIR@` cannot work?

@jpakkane an `is_manpage` kwarg would share the exact same problem as `@MANDIR@` since it would not come with an accompanying `manpage_locale` kwarg.

eli-schwartz:
Does anyone have further thoughts on this? I don't want to merge a solution that doesn't actually have feature compatibility for the existing uses of install_man() without built outputs.

jpakkane:
Thinking about this I'm leaning towards `install_dir : '@MANDIR@' `, because it simplifies code from:

    install_dir: get_option('libdir') / 'foobar'

to

    install_dir: '@LIBDIR@/foobar'

Technically this is an API change, but I've never heard of anyone installing things to a directory with a `@` in it. Has anyone?

eli-schwartz:
Once again, I do not understand how this solves the feature compatibility with `install_man(..., locale: 'fr')`.

jpakkane:
In the only test case we have the filename (`foo.fr.1`) does have all the information needed. It even removes the `.fr` part from the middle. Could we say that if you want the automatic locale based thing to work, you must name your generated files as `bob.langname.number`?

jpakkane:
Or possibly `@FR_MANDIR@`?

Tachi107:
> Could we say that if you want the automatic locale based thing to work, you must name your generated files as `bob.langname.number`?

This is a sane requirement IMO. I can't think of a reason why you would want to name localized manpages differently anyway

PR #9343, opened 2021-10-03
modules/i18n: fix gettext use of --datadirs param

ddstreet:
The previous commit bd2fcb268b9ff48797bebb6a2ef94d2741234191
accidentally used the wrong var so the param name was missing,
leading to an error of "unrecognized arguments" for the
datadirs parameter value.


eli-schwartz (approved):
makes sense, thanks for catching this!

PR #9344, opened 2021-10-03
add install_emptydir function

eli-schwartz:
This replaces the absolute hack of using

```
install_subdir('nonexisting', install_dir: 'share')
```

which requires you to make sure you don't accidentally or deliberately have a completely different directory with the same name in your source tree that is full of files you don't want installed. It also avoids splitting the name in two and listing them in the wrong order.

You can also set the install mode of each directory component by listing them one at a time in order, and in fact create nested structures at all.


test cases/common/246 install_emptydir/meson.build
```diff
@@ -0,0 +1,4 @@
+project('install_emptydir')
+
+install_emptydir(get_option('datadir')/'new_directory', install_mode: 'rwx------')
+install_emptydir(get_option('datadir')/'new_directory/subdir', install_mode: 'rwxr-----')
```
jpakkane, line 4: There should be an `install_emptydir(get_option('datadir')/'foo/bar/baz')` to test that it creates the intermediary path segments also.

eli-schwartz: It already creates datadir (`usr/local/share/`).

keszybz:
```diff
diff --git src/journal-remote/meson.build src/journal-remote/meson.build
index 4debab0d84..612a25bb49 100644
--- src/journal-remote/meson.build
+++ src/journal-remote/meson.build
@@ -65,11 +65,8 @@ if conf.get('ENABLE_REMOTE') == 1 and conf.get('HAVE_MICROHTTPD') == 1
                      install_dir : pkgdatadir / 'gatewayd')
 
         if get_option('create-log-dirs')
-                meson.add_install_script(sh, '-c',
-                                         mkdir_p.format('/var/log/journal/remote'))
-                meson.add_install_script(sh, '-c',
-                                         '''chown 0:0 $DESTDIR/var/log/journal/remote &&
-                                            chmod 755 $DESTDIR/var/log/journal/remote || :''')
+                install_emptydir('/var/log/journal/remote',
+                                 install_mode : ['rwxr-x---', 0, 0])
         endif
 endif
```

The first install succeeds, but the second one fails:
```console
$ DESTDIR=/var/tmp/inst1 ~/src/meson/meson/meson.py install -C build-meson
...
Installing new directory /var/tmp/inst1/var/log/journal/remote
  File "/home/zbyszek/src/meson/meson/mesonbuild/mesonmain.py", line 228, in run
  File "/home/zbyszek/src/meson/meson/mesonbuild/minstall.py", line 761, in run
    installer.do_install(datafilename)
  File "/home/zbyszek/src/meson/meson/mesonbuild/minstall.py", line 541, in do_install
    self.install_emptydir(d, dm, destdir, fullprefix)
  File "/home/zbyszek/src/meson/meson/mesonbuild/minstall.py", line 602, in install_emptydir
    dm.makedirs(full_dst_dir)
  File "/home/zbyszek/src/meson/meson/mesonbuild/minstall.py", line 105, in makedirs
    self.makedirs_impl(path, exist_ok=exist_ok)
  File "/home/zbyszek/src/meson/meson/mesonbuild/minstall.py", line 321, in makedirs
    os.makedirs(*args, **kwargs)
  File "/usr/lib64/python3.10/os.py", line 225, in makedirs
    mkdir(name, mode)
FileExistsError: [Errno 17] File exists: '/var/tmp/inst1/var/log/journal/remote'
```

Two considerations:
- installs must be idempotent, so this should not fail
- but more broadly, it must be OK for the directory to already exist and contain files. This will match behaviour in other cases, where install replaces existing files to the extent necessary to perform the installation, but does not touch anything else.

I think it might be better to drop "empty" from the name, because the command should only create a directory without any promises about it being empty.

eli-schwartz:
Hmm, the idempotence worked the first time I tried this locally, but I think I might have dropped the exist_ok parameter by accident while refactoring it...

Yikes!

eli-schwartz:
When I named the function, I considered that it would install an empty dir, potentially overlapping an existing directory with content. Hence the name.

I was concerned that the name should signify a distinction between it and install_subdir, which copies contents, and I didn't like "install_create_dir" because "install" and "create" have duplicated meaning.

Of course, it is intended that it work as you say, even if it is not named as you say... if you have a better name I'm willing to listen, but I don't think "install_dir" or "install_create_dir" is the answer so I will not change it to that unless @jpakkane weighs in as a tiebreaker.

keszybz:
I even said that naming is hard in my initial bug report. :)

I'm completely fine with `install_emptydir()`. But please mention that existing contents will not be removed in the docs: it might not be obvious, esp. with "empty" in the name.

eli-schwartz:
Ha, of course :D

I will clarify that in the docs, probably at the same time I add a release notes snippet.

xclaesse:
Hmmm, I would have expected a kwarg on install_subdir(..., empty: true) tbh. But I've no strong opinion. At least add something I'm the install_subdir() doc that points to this new function.

keszybz:
> install_subdir(..., empty: true)

Hmm, indeed. That would be nicer from the users' point of view: less entities to read about.

eli-schwartz:
install_subdir already "works" but is unfortunately crippled by the fact that it takes a distinct install_dir kwarg which requires splitting the name into two parts.

Are you suggesting that install_subdir should have two function signatures, one for empty dirs and one for recursive subdirs, and that for backwards compatibility purposes (because we have documented it already) the latter *also* supports empty dirs using a different syntax and without support for install_mode?

docs/yaml/functions/install_emptydir.yaml
```diff
@@ -0,0 +1,27 @@
+name: install_emptydir
+returns: void
+since: 0.60.0
+description: |
+  Installs a new directory entry to the location specified by the positional
+  argument. If the directory exists and is not empty, the contents are left in
+  place.
```
eli-schwartz, line 7: docs updated

mesonbuild/backend/backends.py
```diff
@@ -147,6 +148,13 @@ def __init__(self, fname: str, outdir: str, outdir_name: str, aliases: T.Dict[st
         self.optional = optional
         self.tag = tag
 
+class InstallEmptyDir:
+    def __init__(self, path: str, install_mode: 'FileMode', subproject: str, tag: T.Optional[str] =None):
```
dcbaker, line 152:
```suggestion
    def __init__(self, path: str, install_mode: 'FileMode', subproject: str, tag: T.Optional[str] = None):
```

Missing a space after the =

docs/yaml/functions/install_subdir.yaml
```diff
@@ -5,9 +5,12 @@ description: |
...
 
-  If the subdirectory does not exist in the source tree, an empty directory is
-  created in the specified location. *(since 0.45.0)* A newly created
-  subdirectory may only be created in the keyword argument `install_dir`.
+  *(since 0.45.0, deprecated since 0.60.0)* If the subdirectory does not exist
+  in the source tree, an empty directory is created in the specified location.
+  A newly created subdirectory may only be created in the keyword argument
+  `install_dir`. There are a number of flaws with this method, and it was never
+  intentionally designed to work this way, please use [[install_emptydir]]
+  instead.
```
eli-schwartz, line 13: And here too.

mesonbuild/interpreter/interpreter.py
```diff
@@ -1894,6 +1896,15 @@ def func_install_man(self, node: mparser.BaseNode,
...
         return m
 
+    @FeatureNew('install_emptydir', '0.60.0')
+    @typed_kwargs(
+        'install_emptydir',
+        INSTALL_MODE_KW
+    )
+    def func_install_emptydir(self, node: mparser.BaseNode, args: T.Tuple[str], kwargs) -> None:
+        d = build.EmptyDir(args[0], kwargs['install_mode'], self.subproject)
+        self.build.emptydir.append(d)
```
dcbaker, line 1906: I think this should return d, right? Otherwise there really isn't any reason for build.EmptyDir to be Holdable, or for the EmptyDirHolder to exist

dcbaker (commented):
Why Insta

mesonbuild/interpreter/interpreter.py
eli-schwartz: Oops, yeah.

dcbaker (approved):

PR #9347, opened 2021-10-04
wrap clone: be less noisy when doing automated code checkouts

### LINKED ISSUE #9318: wrap-git installs can show too much detail when the clone operation completes as expected
**Describe the bug**
When using a wrapped git subproject (and especially so with a revision), the clone command use the user configured git settings for hints, which show when first installing the wrap. 

**To Reproduce**
Use this `meson.build` file (generated with `meson init`):

```meson
[wrap-git]
url=https://github.com/GNOME/glib
directory = glib
revision=a0d2efdcb043af10c11e0d44a3faa4fd5b7f76bb
depth = 1
```
With a wrap file like this:
```ini
[wrap-git]
url=https://github.com/GNOME/glib
directory = glib
revision=a0d2efdcb043af10c11e0d44a3faa4fd5b7f76bb
depth = 1
```

Run `meson setup builddir`

See this snipped output:

```sh
hint: Using 'master' as the name for the initial branch. This default branch name
hint: is subject to change. To configure the initial branch name to use in all
hint: of your new repositories, which will suppress this warning, call:
hint: 
hint: 	git config --global init.defaultBranch <name>
hint: 
hint: Names commonly chosen instead of 'master' are 'main', 'trunk' and
hint: 'development'. The just-created branch can be renamed via this command:
hint: 
hint: 	git branch -m <name>
Initialized empty Git repository in /path/workspaces/meson_test/subprojects/glib/.git/
remote: Enumerating objects: 2085, done.
remote: Counting objects: 100% (2085/2085), done.
remote: Compressing objects: 100% (1772/1772), done.
remote: Total 2085 (delta 493), reused 909 (delta 186), pack-reused 0
Receiving objects: 100% (2085/2085), 7.55 MiB | 8.70 MiB/s, done.
Resolving deltas: 100% (493/493), done.
From https://github.com/GNOME/glib
 * branch            a0d2efdcb043af10c11e0d44a3faa4fd5b7f76bb -> FETCH_HEAD
Note: switching to 'a0d2efdcb043af10c11e0d44a3faa4fd5b7f76bb'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false

HEAD is now at a0d2efd Merge branch '2498-io-module-annotations' into 'main'
```
**Expected behavior**
I would expect not see the git hints, and also to simply get output like "Subproject glib was successfully checked out at revision X" 

**system parameters**
* what operating system (e.g. MacOS Catalina, Windows 10, CentOS 8.0, Ubuntu 18.04, etc.) Fedora 35 Beta
* what Python version are you using e.g. 3.8.0 3.10
* what `meson --version` 0.59.1
* what `ninja --version` if it's a Ninja build 1.10.1

eli-schwartz:
There are two possible issues, both of which emit very long messages from git:
- when shallow cloning via depth + a pinned commit we locally init rather than cloning; use an almost-certainly not conflicting dummy branch name
- any time a checkout of a revision is performed, it might not be a branch name, in which case it will be a detached HEAD. By default git is very noisy about this -- it wants you to know what happened and how not to mess up. But wraps aren't per default intended for user editing and development, it's just part of the internal transport which meson uses. So, temporarily squelch this advice. Do not permanently configure the repo like this though, because the user *might* cd in and start developing on the subproject; in this case, any additional git advice they trigger is their responsibility (and they probably do want it).

Fixes #9318

### DIFF
--- a/mesonbuild/wrap/wrap.py
+++ b/mesonbuild/wrap/wrap.py
@@ -401,6 +401,7 @@ class Resolver:
         if not GIT:
             raise WrapException(f'Git program not found, cannot download {self.packagename}.wrap via git.')
         revno = self.wrap.get('revision')
+        checkout_cmd = ['-c', 'advice.detachedHead=false', 'checkout', revno, '--']
         is_shallow = False
         depth_option = []    # type: T.List[str]
         if self.wrap.values.get('depth', '') != '':
@@ -410,11 +411,11 @@ class Resolver:
         if is_shallow and self.is_git_full_commit_id(revno):
             # git doesn't support directly cloning shallowly for commits,
             # so we follow https://stackoverflow.com/a/43136160
-            verbose_git(['init', self.directory], self.subdir_root, check=True)
+            verbose_git(['init', '-b', 'meson-dummy-branch', self.directory], self.subdir_root, check=True)
             verbose_git(['remote', 'add', 'origin', self.wrap.get('url')], self.dirname, check=True)
             revno = self.wrap.get('revision')
             verbose_git(['fetch', *depth_option, 'origin', revno], self.dirname, check=True)
-            verbose_git(['checkout', revno, '--'], self.dirname, check=True)
+            verbose_git(checkout_cmd, self.dirname, check=True)
             if self.wrap.values.get('clone-recursive', '').lower() == 'true':
                 verbose_git(['submodule', 'update', '--init', '--checkout',
                              '--recursive', *depth_option], self.dirname, check=True)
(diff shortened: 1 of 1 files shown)


mesonbuild/wrap/wrap.py
```diff
@@ -413,11 +414,11 @@ def get_git(self) -> None:
         if is_shallow and self.is_git_full_commit_id(revno):
             # git doesn't support directly cloning shallowly for commits,
             # so we follow https://stackoverflow.com/a/43136160
-            verbose_git(['init', self.directory], self.subdir_root, check=True)
+            verbose_git(['init', '-b', 'meson-dummy-branch', self.directory], self.subdir_root, check=True)
```
xclaesse, line 417: Why do we need to create a branch here, can't it left detached? If that's only to avoid printing `Initialized empty Git repository in ...` then just use `quiet_git()` instead?

xclaesse: Oh, did not know it would fallback the `master`, I thought it would fallback to detached mode. We actually checkout the commit later so we end up in the detached mode anyway.

Nitpicking: Maybe remove the dummy branch after the checkout?

xclaesse:
Just a small nitpicking, feel free to ignore. Also CI failed it seems. Other than that, LGTM.

eli-schwartz:
The CI failed for hysterical raisins, but it passed immediately before I rebased on master and did a fast-forward merge. So it should pass momentarily...

PR #9349, opened 2021-10-04
interpreter: improve the error message about install_mode

keszybz:
We wrote:
 ERROR: install_emptydir keyword argument "install_mode" permissions string must
        be exactly 9 characters, got "4" in the form rwxr-xr-x

Let's change this around to be easier to read.

Also, 1-based numbering was used (for components) and 0-based for "bits".
And actually the "bits" are not bits, but octal digits. So say "permissions
character 1", "permissions character 2".

And finally change "must be … if provided" to "can only be". (If it isn't
provided, it "is not", so the sentence is still valid. The user will only
get this error if they provide something, so we don't need to be super precise
and say "if provided". And then we avoid confusing the reader whether
it's "if provided" attaches to the the "False" or to the whole sentence.)

### DIFF
--- a/mesonbuild/interpreter/type_checking.py
+++ b/mesonbuild/interpreter/type_checking.py
@@ -44,40 +44,39 @@ def _install_mode_validator(mode: T.List[T.Union[str, bool, int]]) -> T.Optional
 
     This is a rather odd thing, it's a scalar, or an array of 3 values in the form:
     [(str | False), (str | int | False) = False, (str | int | False) = False]
-    Where the second and third arguments are not required, and are considered to
-    default to False.
+    where the second and third components are not required and default to False.
     """
     if not mode:
         return None
     if True in mode:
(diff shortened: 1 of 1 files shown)


mesonbuild/interpreter/type_checking.py
```diff
@@ -56,28 +55,28 @@ def _install_mode_validator(mode: T.List[T.Union[str, bool, int]]) -> T.Optional
 
     perms = mode[0]
     if not isinstance(perms, (str, bool)):
-        return 'permissions part must be a string or false'
+        return 'permissions string must be a string or false'
```
eli-schwartz, line 58: Permissions ***string*** must be either the aforementioned string, or else violate that rule and be something which is not a string?

keszybz: Argh, I'll reword.

PR #9351, opened 2021-10-04
python: Add install_platlib and install_purelib options

xclaesse:
More generally this adds support for module options. It use the format `<module>.<option name>` format because the dot is already forbidden in project options and we already use that format for `build.`.

Currently python module prints a warning that cannot be avoided when prefix option is different from the python site-package's prefix, which is always the case on Windows and is the case on Linux when not installing to `/usr` which is the case by default. However, I'm actually wondering if we should instead have an option that force installing into python's site-package even when that's outside of meson's install prefix? That seems to be a more user friendly solution, but less flexible.


eli-schwartz:
Nitpick: the option name sounds like the kind of option you'd use as a boolean "should platlib files be installed or not". Maybe `python.platlib_dir` would be a better style name?

eli-schwartz:
```
  File "run_project_tests.py", line 128, in <module>
    python.sanity()
TypeError: sanity() missing 1 required positional argument: 'state'
```

xclaesse:
> Nitpick: the option name sounds like the kind of option you'd use as a boolean "should platlib files be installed or not". Maybe `python.platlib_dir` would be a better style name?

Good point. I named them `platlibdir` and `purelibdir` because we don't have underscore in other dir options like `datadir`, `libdir`, etc.

xclaesse:
@eli-schwartz CI is passing now.

xclaesse:
Lol that's completely unexpected side effect of renaming the option. Proving one again that even simple stuff always needs a unit test...

eli-schwartz:
This seems a bit magical compared to a mutually exclusive option e.g. `-Dpython.defaultdirs=[env|prefix]`

Env would be the python environment, prefix would make it relative to the prefix.

Or something like that which involves another option and doesn't cause people who for bizarre reasons want to install to `f'{prefix}/default'` to get surprising results.

xclaesse:
@eli-schwartz ok, I removed the commit, we can always add that later.

mesonbuild/modules/__init__.py
```diff
@@ -102,6 +103,12 @@ def test(self, args: T.Tuple[str, T.Union[build.Executable, build.Jar, 'External
         # TODO: Use interpreter internal API, but we need to go through @typed_kwargs
         self._interpreter.func_test(self.current_node, args, kwargs)
 
+    def get_option(self, name: str, subproject: str = '',
+                 machine: MachineChoice = MachineChoice.HOST,
+                 lang: T.Optional[str] = None,
+                 module: T.Optional[str] = None) -> T.Union[str, int, bool, 'WrapMode']:
```
dcbaker, line 109: indent looks off here.

unittests/datatests.py
```diff
@@ -145,8 +146,8 @@ def test_builtin_options_documented(self):
             found_entries |= options
 
         self.assertEqual(found_entries, {
-            *(str(k) for k in mesonbuild.coredata.BUILTIN_OPTIONS),
-            *(str(k) for k in mesonbuild.coredata.BUILTIN_OPTIONS_PER_MACHINE),
+            *(str(k.evolve(module=None)) for k in mesonbuild.coredata.BUILTIN_OPTIONS),
+            *(str(k.evolve(module=None)) for k in mesonbuild.coredata.BUILTIN_OPTIONS_PER_MACHINE),
```
dcbaker, line 150: This seems weird, why do you need to set the module to None here?

PR #9353, opened 2021-10-04
docs: Temporarily disable modules and move RefMan --> Reference-manual

mensinda:


xclaesse:
Assuming you tested locally, lgtm

mensinda:
Yes, tested locally. I am going to wait for the CI though since you can never truly know :)

PR #9354, opened 2021-10-05
msubprojects: add new subcommand "packagefiles"

eli-schwartz:
This will re-apply the meson.build patch overlay, ensuring it is up to date. Also take the opportunity offered by this infrastructure to repatch when performing `update --reset` since internally this will run `git stash` and thus cause the (possibly locally modified) meson.build files to disappear.

/cc @xclaesse

### DIFF
--- a/docs/markdown/snippets/subprojects_packagefiles.md
+++ b/docs/markdown/snippets/subprojects_packagefiles.md
@@ -1,0 +1,11 @@
+## New `subprojects packagefiles` subcommand
+
+It is now possible to re-apply `meson.build` overlays (`patch_filename` or
+`patch_directory` in the wrap ini file) after a subproject was downloaded and
+set up, via `meson subprojects packagefiles --apply <wrap-name>`.
+
+It is also possible for `patch_directory` overlays in a `[wrap-file]`, to copy
+the packagefiles out of the subproject and back into `packagefiles/<patch_directory>/`
+via `meson subprojects packagefiles --save <wrap-name>`. This is useful for
+testing an edit in the subproject and then saving it back to the overlay which
+is checked into git.
--- a/mesonbuild/msubprojects.py
+++ b/mesonbuild/msubprojects.py
@@ -58,7 +58,9 @@ class Runner:
         # FIXME: Do a copy because Resolver.resolve() is stateful method that
         # cannot be called from multiple threads.
         self.wrap_resolver = copy.copy(r)
-        self.wrap = wrap
+        self.wrap_resolver.current_subproject = 'meson'
+        self.wrap_resolver.dirname = os.path.join(r.subdir_root, wrap.directory)
+        self.wrap = self.wrap_resolver.wrap = wrap
         self.repo_dir = repo_dir
         self.options = options
         self.run_method = options.subprojects_func.__get__(self)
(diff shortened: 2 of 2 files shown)


eli-schwartz:
It feels kind of meh that the name once more hardcodes the term "patch" even though it is not a "patch". I recognize this is ironic since I rewrote the docs to emphasize:

> For historic reasons this is called a "patch", however, it serves as an overlay to add or replace files rather than modifying them. The file must be an archive; it is downloaded and automatically extracted into the subproject. The extracted files will include a Meson build definition for the given subproject.

I can't think of a better name at the moment though...

eli-schwartz:
Are there other cases we should be re-applying the patch? In #9277 I said I was surprised we don't generally do this, but thinking over it again, given we explicitly stash during the `update --reset` case, I am no longer sure we should generally apply it as we might overwrite local modifications.

It is completely reasonable to overwrite local modifications when manually running `subprojects repatch`, or when all changes got stashed. Not so sure about when merely pulling.

xclaesse:
Thanks, I forgot about this stuff.

Just a quick thought, I'm wondering if we could move the code I wrote in wrapdb to update packagefiles here too. That's the reverse operation of repatch. Maybe name them: meson subprojects packagefiles --apply and --update ?

doraskayo:
I'm not familiar with the implementation details and design around wraps, but I think it's important for the implementation to be able to discern in which state a subproject is at any given point without any external input:
1. Subproject source wasn't obtained yet
2. Subproject source was obtained but not patched/prepared yet
3. Subproject is patched/prepared and ready for use on builds

Meson should be able to attempt to progress from (1) to (3) and (2) to (3) given the state of the subproject on its own, even as part of `meson build`.

Specifically, if a subproject gir repository is in state (2) because the source was already obtained through some means before `meson build` was invoked (`git submodule update --init` or otherwise), Meson should be able to figure out that it needs to attempt to checkout the relevant branch or tag, patch the source repository and do whatever else it need to do to get to (3).

Having a dedicated command to help us get from (2) to (3) is indeed nice to have, but perhaps not a must.

My guess is that determining we are at (1) should be fairly easy: just check if the subproject source directory exists.

The more difficult part is distinguishing between (2) and (3). The first idea that comes to mind is to drop a hidden file inside the source directory (`.meson-packagefiles-applied`? Maybe something shorter) when the patching/preparation is done. Checking for its existence would determine if the subproject is in (2) or (3). Of course, manual fiddling with the subproject source outside Meson commands would not supported.

This approach should allow better versatility around the method by which the subproject source is obtained (Meson no longer cares), and should provide a more graceful solution to the use case of sandboxed build environments since no additional command would be required to fully build the code.

I'm not sure how sensible this idea is given the current capabilities, but I thought it would be a good idea to raise it for discussion.

eli-schwartz:
> Just a quick thought, I'm wondering if we could move the code I wrote in wrapdb to update packagefiles here too. That's the reverse operation of repatch. Maybe name them: meson subprojects packagefiles --apply and --update ?

Hmm, what's the general use case of --update outside of the wrapdb, (and is there a name that doesn't sound like it does the same thing as --apply)?

eli-schwartz:
@xclaesse command updated, added --save.

xclaesse:
@eli-schwartz implementation looks good. Missing release snippet. A unit test would be great too.

jpakkane:
Is the description still correct? The code changes do not have the word `repatch` anywhere...

eli-schwartz:
The description is heinously outdated, sorry.

Originally it was a copy of the central commit message, which I updated without getting around to fixing GitHub.

eli-schwartz:
I'll fix it in a bit, when I add a release snippet.

PR #9355, opened 2021-10-05
modules/gnome: deprecate yelp variadic sources

dcbaker:
Yelp currently can take sources two different ways, the first is via
variadic arguments, the second is by a keyword argument. If the keyword
is passed then the variadic arguments are silently ignored, which is
obviously not ideal. Fortunately the variadic form was never documented,
and is likely not in wide use.

This patch fixes it by deprecating the variadic form, and warning if
both are passed. It does not change behavior as someone may be relying
on it.

### DIFF
--- a/docs/markdown/Gnome-module.md
+++ b/docs/markdown/Gnome-module.md
@@ -298,8 +298,22 @@ VAPI or Vala binaries.
 
 ### gnome.yelp()
 
+```meson
+  gnome.yelp(id: string, sources: ...string, sources: []string, media: []string,
+             languages: []string, symlink_media: bool = true): void
+```
+
 Installs help documentation using Yelp. The first argument is the
 project id.
+
+Additionally, sources can be passed as additional positional arguments. This
+was, however, undocumented and never officially supported. Due to a longstanding
+bug, passing sources as a keyword argument will result in the positional
+argument sources to be ignored. *since 0.60.0* A warning is raised in this case.
+
+*Since 0.60.0* the use of the positional argument sources has been deprecated,
+and the "sources" keyword argument should be used instead. The passing of
+sources as positional arguments will be removed in the future.
 
 This also creates two targets for translations
 `help-$project-update-po` and `help-$project-pot`.
--- a/docs/markdown/snippets/gnome_yelp_sources.md
+++ b/docs/markdown/snippets/gnome_yelp_sources.md
@@ -1,0 +1,6 @@
+## gnome.yelp variadic argument deprecation
+
+`gnome.yelp` previously allowed sources to be passed either as variadic
+arguments or as a keyword argument. If the keyword argument was given the
+variadic arguments would be silently ignored. This has changed in 0.60.0, the
+variadic form has been deprecated, and a warning is printed if both are given.
--- a/mesonbuild/modules/gnome.py
+++ b/mesonbuild/modules/gnome.py
@@ -33,7 +33,7 @@ from ..mesonlib import (
     join_args, HoldableObject
 )
 from ..dependencies import Dependency, PkgConfigDependency, InternalDependency
-from ..interpreterbase import noPosargs, noKwargs, permittedKwargs, FeatureNew, FeatureNewKwargs, FeatureDeprecatedKwargs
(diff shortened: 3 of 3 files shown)


docs/markdown/Gnome-module.md
```diff
@@ -298,8 +298,20 @@ VAPI or Vala binaries.
...
 Installs help documentation using Yelp. The first argument is the
-project id.
+project id. Additionally it may take sources as variadic arguments.
+
+Due to a longstanding bug, passing sources as a keyword argument will result in
+the variadic argument sources to be ignored. A warning is raised in this case
+*since 0.60.0*.
+
+*Since 0.60.0* the use of the variadic argument sources has been deprecated, and
+the "sources" keyword argumetn should be used instead.
```
TingPing, line 314:
```suggestion
the "sources" keyword argument should be used instead.
```

docs/markdown/snippets/gnome_yelp_sources.md
```diff
@@ -0,0 +1,6 @@
+## gnome.yelp variadic argument deprecation
+
+`gnome.yelp` previously allowed sources to be passed either as variadic
+argumetns, or as a keyword argument. If the keyword argument was given the
```
TingPing, line 4:
```suggestion
arguments, or as a keyword argument. If the keyword argument was given the
```

mesonbuild/modules/gnome.py
```diff
@@ -969,12 +969,18 @@ def yelp(self, state, args, kwargs):
             raise MesonException('Yelp requires a project id')
 
         project_id = args[0]
+        if len(args) > 1:
+            FeatureDeprecated.single_use('gnome.yelp variadic source args', '0.60.0', 'use the "sources" keyword argument instead.')
```
xclaesse, line 973: IMHO "variadic" is too technical, most user won't know what it is about. We already have tones of APIs where we say something along "more than one positional argument".

docs/markdown/Gnome-module.md
```diff
@@ -298,8 +298,20 @@ VAPI or Vala binaries.
...
 ### gnome.yelp()
 
+```meson
+  gnome.yelp(id: string, sources: ...string, sources: []string, media: []string,
+             languages: []string, symlink_media: bool = true): void
+```
+
 Installs help documentation using Yelp. The first argument is the
-project id.
+project id. Additionally it may take sources as variadic arguments.
```
eli-schwartz, line 307: I would kinda prefer something like:

> Additionally, there was undocumented support for sources as positional arguments rather than as a kwarg. This should never be used, and if you did use it in combination with proper kwargs, then the positional arguments would be silently ignored.
>
> Since 0.60.0, this raises a warning and may be removed at any time.

And then just... don't remove it for a few releases.

eli-schwartz: I just feel like documenting it as "valid but deprecated" is giving it too much credit and people might for some strange reason decide to use it because it works for earlier versions of meson. Emphasizing that it was never officially allowed and that we make zero commitment to keeping support for it (whereas deprecated things are usually kept around for a long time and are thus somehow "safe") will give people the additional impetus to stop using this really fast, if they happened to be doing so.

dcbaker: I tried to update the comments to point out that this used to work, but we don't support it, and it's slated for removal.

docs/markdown/Gnome-module.md
```diff
@@ -298,9 +298,24 @@ VAPI or Vala binaries.
...
+```meson
+  gnome.yelp(id: string, sources: ...string, sources: []string, media: []string,
+             languages: []string, symlink_media: bool = true): void
+```
+
 Installs help documentation using Yelp. The first argument is the
 project id.
 
+Addiitionally, sources can be passed as additional positional
+arguments. This was, however undcoumented, and never officially supported. Due
```
eli-schwartz, line 310:
```suggestion
arguments. This was, however undocumented, and never officially supported. Due
```

docs/markdown/Gnome-module.md
eli-schwartz: Nice, wording seems good.

PR #9356, opened 2021-10-05
Zipapp docs

eli-schwartz:
Make this actually visible to users. We have previously documented this but only in README.md... it got removed "temporarily" because it was incompatible with multiple entry points, which we no longer use, but the docs never got restored at that time.

Motivation: this is helpful to e.g. packagers such as https://github.com/openwrt/openwrt/pull/4635 but people simply don't know it exists until I mention it directly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,14 +47,7 @@ Meson can be run as a [Python zip
 app](https://docs.python.org/3/library/zipapp.html). To generate the
 executable run the following command:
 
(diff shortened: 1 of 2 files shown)


dhewg:
For the record:
Until the linked PR OpenWrt basically used ´cp $gitclonedir/* $targetdir`, just because that ability was/is documented:
https://mesonbuild.com/Quick-guide.html#installation-from-source
https://mesonbuild.com/Running-Meson.html#running-meson

Issue #5364, opened 2019-05-05
feature option can't be set to auto when auto_features is set


lazka:
meson.build:
```meson
project('foo', 'c', version : '1.16.0')
if get_option('opencv').auto()
  message('opencv: auto')
elif get_option('opencv').enabled()
  message('opencv: enabled')
else
  message('opencv: disabled')
endif
```

meson_options.txt:
```
option('opencv', type : 'feature', value : 'auto')
```

```
meson -Dauto_features=enabled -Dopencv=auto _build
```

* Expected: opencv is set to auto
* Actual: opencv is set to enabled

Context: I'm using auto_features=enabled for packaging and after that changing each option as needed.

dcbaker:
That is the intended behavior, the idea is that packagers can use that option to ensure that there is no automagic going on behind their backs. We could (and maybe should) give a warning/error when auto_features is set and a features is explicitly set to auto.

PR #9357, opened 2021-10-06
WIP: interpreter: don't flip user-specified 'auto' with 'auto_features'

havardgraff:
Would like some feedback on this one.

The problem it is solving is that for GStreamer, you might want to be very picky about what gets built, and in that case
setting 'auto_features' to disabled is the perfect solution.

However, you might want to build some platform dependant plugins, like v4l2 or applemedia,  and the best way to build these it to use the 'auto' feature, where they will be built dependent on their dependencies existing on your system.

So this patch proposes to keep track of all "user defined" options, and then *not* flip the 'auto' ones that are specifically set to 'auto' by the user.

### DIFF
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -72,6 +72,7 @@ class UserOption(T.Generic[_T], HoldableObject):
         if not isinstance(yielding, bool):
             raise MesonException('Value of "yielding" must be a boolean.')
         self.yielding = yielding
+        self.user_input = False
 
     def printable_value(self) -> T.Union[str, int, bool, T.List[T.Union[str, int, bool]]]:
         assert isinstance(self.value, (str, int, bool, list))
@@ -83,8 +84,12 @@ class UserOption(T.Generic[_T], HoldableObject):
     def validate_value(self, value: T.Any) -> _T:
         raise RuntimeError('Derived option class did not override validate_value.')
 
-    def set_value(self, newvalue: T.Any) -> None:
+    def set_value(self, newvalue: T.Any, user_input: T.Optional[bool] = False) -> None:
         self.value = self.validate_value(newvalue)
+        self.user_input = user_input
+
+    def is_user_input(self) -> bool:
+        return self.user_input
 
 class UserStringOption(UserOption[str]):
     def __init__(self, description: str, value: T.Any, yielding: T.Optional[bool] = None):
@@ -629,7 +634,7 @@ class CoreData:
                 value = self.sanitize_dir_option_value(prefix, key, value)
 
         try:
-            self.options[key].set_value(value)
+            self.options[key].set_value(value, True)
         except KeyError:
             raise MesonException(f'Tried to set unknown builtin option {str(key)}')
 
--- a/mesonbuild/interpreter/interpreterobjects.py
+++ b/mesonbuild/interpreter/interpreterobjects.py
@@ -83,7 +83,7 @@ def extract_search_dirs(kwargs: 'kwargs.ExtractSearchDirs') -> T.List[str]:
 class FeatureOptionHolder(ObjectHolder[coredata.UserFeatureOption]):
     def __init__(self, option: coredata.UserFeatureOption, interpreter: 'Interpreter'):
         super().__init__(option, interpreter)
-        if option and option.is_auto():
+        if option and option.is_auto() and not option.is_user_input():
             # TODO: we need to case here because options is not a TypedDict
             self.held_object = T.cast(coredata.UserFeatureOption, self.env.coredata.options[OptionKey('auto_features')])
             self.held_object.name = option.name


nirbheek:
The concept makes sense to me. I'm not too sure about the architecture used by the solution since I have lost track of how options are passed around these days.

mesonbuild/coredata.py
```diff
@@ -629,7 +634,7 @@ def set_option(self, key: OptionKey, value) -> None:
                 value = self.sanitize_dir_option_value(prefix, key, value)
 
         try:
-            self.options[key].set_value(value)
+            self.options[key].set_value(value, True)
```
xclaesse, line 637: AFAIK `set_option()` is called for `project(..., default_options: ...)` too, and in that case you still want to use auto_feature.

mesonbuild/interpreter/interpreterobjects.py
```diff
@@ -83,7 +83,7 @@ def extract_search_dirs(kwargs: 'kwargs.ExtractSearchDirs') -> T.List[str]:
 class FeatureOptionHolder(ObjectHolder[coredata.UserFeatureOption]):
     def __init__(self, option: coredata.UserFeatureOption, interpreter: 'Interpreter'):
         super().__init__(option, interpreter)
-        if option and option.is_auto():
+        if option and option.is_auto() and not option.is_user_input():
```
xclaesse, line 86: I think you can instead do `option.name in interpreter.user_defined_options.cmd_line_options`. However, I'm wondering what we should do for options that have been set to `auto` in the cross/native file, those should be considered set by user too afaik. tbf, I'm not entirely happy with `interpreter.user_defined_options` it should be refactored as a boolean on the UserOption object as you tried to do indeed.

bonzini:
No objection on the concept, but please make it happen only when a new-enough version of Meson is specified in `project()`. In particular, this would break QEMU (which still needs a configure script that wraps meson, unfortunately).

eli-schwartz:
The meson_version in project() is intended to raise deprecations and FeatureNew and error if you have a known incompatible version of meson.

It's not supposed to actually change the behavior of meson, and crippling this feature in the majority of projects which don't yet specify meson 0.60.0 is a big problem.

It would be better to not merge it at all than to have it act differently depending on your project(meson_version: ...)

eli-schwartz:
> In particular, this would break QEMU (which still needs a configure script that wraps meson, unfortunately).

Sorry, I don't understand why this should be the case. And if you're saying it will eventually not be the case, what exactly is going to change in QEMU between now and then?

eli-schwartz:
That configure script: https://github.com/qemu/qemu/blob/master/configure

This honestly feels like it is not really using meson. Is it possible to run meson on its own? It seems like there is a --skip-meson flag which should let you specify your own option there, no?

And more generally, I think this is a good case for checking a *maximum* version of meson. Seems like you have such fiddly expectations about how meson behaves that you need to very carefully keep this... script... in sync with meson itself. You are literally reimplementing the *entire* command line of meson, or should I say, "only parts of it and completely preventing users from passing any options not special cased in the configure script".

It's bad design, tbh. You're not checking any of these settings, only translating them into -D options. You should completely delete them from your script, document that -D* options are passed unmodified to meson, and leave be.

Given how closely this is tied to the version of meson this configure script is developed against, I see no reason to reject this PR on the grounds that qemu doesn't support it.

Also -- there are not even any real consequences of qemu's command line reimplementation being incorrect here! All that happens is, the auto_features option fails to provide enhanced value to this one project. So, don't take advantage of it then. The actual build system works fine, and you can configure it the traditional way -- yourself.

eli-schwartz:
TBH, users will be allowed to set those if/when you fully switch over to meson without wrappers, and, as a general rule of thumb people do not set the _std themselves but might in a superproject.

And, unity is in fact intended for being set by users, not by projects. In theory this is supposed to work fine either way. In practice, unity builds are dumb and don't work half the time, but even for projects which do actually work correctly with them, it is supposed to be a user choice, not a project choice. Admittedly, it might make sense to have a project choice `is_unity_compatible`...

And layout is technically the same kind of user choice, except it is so broken that we've finally deprecated it for removal.

Anyway, the point stands -- as an advanced use case you want to allow this escape hatch with the acknowledgement that people who use it to break their build get to keep both pieces.

... ooooor you could add those to the option parsing case statement so they get explicitly checked for in configure itself, and raise a suitable error.

> there are still a few options that require processing in both configure and meson

That's why I said for compatibility purposes, you can do automatic pattern based mapping in pure POSIX sh. :D

You could handle the manual cases explicitly for compatibility too -- just set a suitable variable for processing in configure, in *addition* to appending to `$meson_remap_opts` rather than as a replacement (don't use those variables for processing the `meson setup` command, just rely on them being set in `$meson_remap_opts`)

eli-schwartz:
Oh, I forgot to mention earlier...

> In fact, I chose meson because I was expecting any incompatibilities to be guarded by `meson_version`.

Usually we don't guard them by meson_version, because:
- fixing bugs is generally not regarded as incompatibilities
- rather than changing how things work depending on version of meson (in the `meson_version` declaration or otherwise), we typically opt to not change how they work at all, but instead deprecate bad ideas, or add kwargs to toggle behavior and preserve existing behavior when the kwarg is not set but also deprecate *not* passing the kwarg. An example I recently stumbled upon (because I wanted to fix deprecation warnings in GLib) is `static_library().extract_all_objects()`, which takes a `recursive: false` parameter and loudly warns you that "you probably meant recursive: true and we will change to this by default in the future, but right now you didn't set anything at all, so here, have a warning and a free backwards-compat default of false."

However, auto_features is not a meson.build option, it is a user flag and I somehow doubt that anyone other than qemu is unintentionally passing useless and unwanted -Dfoo=auto at the same time, ***and*** the workaround is for the user who specified -Dauto_features=disabled to manually unroll options  and specify them all, as used to be the case before meson.

Since this is fundamentally a user interactivity impacting PR, it is both incorrect to depend on the project itself, and reasonable to expect that users doing the wrong thing anyway will read the release notes when upgrading their use of meson (in contrast to projects, which cannot just magically upgrade old build files on a whim).

That being said.

This currently :crossed_swords: conflicts, needs release notes, and I think @xclaesse may want this to handle options specified in the native/cross file first.... and is really close to rc1, so it may not land in time for 0.60, which would mean qemu has all the way until 0.61 to fix its configure script to not trample on individual options, and release that fix in a stable tag.

To be honest, I'm fine with that resolution too -- nothing is broken either way, users wanting to take advantage of the **nifty less-typing bulk option setter otherwise known as "auto_features"** by way of simply upgrading qemu, and users of other projects get the granularity of `-Dauto_features=disabled -Dfoobar=auto` eventually, if not as fast as they could have otherwise.

I am still convinced that whenever this PR lands, it should be immediately functional rather than requiring projects to update their meson.build before users are permitted to specify the user option.

eli-schwartz:
This PR implements #5364.

Can we get a release notes snippet and resolve the conflicts?

rossburton:
I'm interested in this too.  I'm adding `--auto-features=enabled` by default to all Meson invocations, but pixman is quite pathological:

```
option('loongson-mmi', type: 'feature', description: 'Use Loongson MMI intrinsic optimized paths')
option('mmx', type: 'feature', description: 'Use X86 MMX intrinsic optimized paths')
option('sse2', type: 'feature', description: 'Use X86 SSE2 intrinsic optimized paths')
option('ssse3', type: 'feature', description: 'Use X86 SSSE3 intrinsic optimized paths')
option('vmx', type: 'feature', description: 'Use PPC VMX/Altivec intrinsic optimized paths')
option('arm-simd', type: 'feature', description: 'Use ARMv6 SIMD intrinsic optimized paths')
option('neon', type: 'feature', description: 'Use ARM NEON intrinsic optimized paths')
option('a64-neon', type: 'feature', description: 'Use ARM A64 NEON intrinsic optimized paths')
option('iwmmxt', type: 'feature', description: 'Use ARM IWMMXT intrinsic optimized paths')
option('iwmmxt2', type: 'boolean', value: true,  description: 'Use ARM IWMMXT2 intrinsic instead of IWMMXT')
option('mips-dspr2', type: 'feature', description: 'Use MIPS32 DSPr2 intrinsic optimized paths')
```

We're building for x86/arm/ppc/mips so having to set those specifically for each platform isn't fun. I'd like to be able to override the auto-features=enabled for those options specifically, and not the rest of the auto options in pixmant too.

eli-schwartz:
@havardgraff ping? I'd like to get this in soon.

PR #9364, opened 2021-10-07
interpreter: extract_objects provides a valid source

bonzini:
This ensures that there is no warnings when running meson on `test cases/common/22 object extraction`.

### DIFF
--- a/mesonbuild/modules/unstable_rust.py
+++ b/mesonbuild/modules/unstable_rust.py
@@ -17,7 +17,7 @@ import typing as T
 
 from . import ExtensionModule, ModuleReturnValue
 from .. import mlog
-from ..build import BothLibraries, BuildTarget, CustomTargetIndex, Executable, GeneratedList, IncludeDirs, CustomTarget
+from ..build import BothLibraries, BuildTarget, CustomTargetIndex, Executable, ExtractedObjects, GeneratedList, IncludeDirs, CustomTarget
 from ..dependencies import Dependency, ExternalLibrary
 from ..interpreter.interpreter import TEST_KWARGS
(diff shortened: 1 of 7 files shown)


bonzini:
Ping (for 0.60.1 even?)

mesonbuild/modules/unstable_rust.py
```diff
@@ -203,6 +203,8 @@ def bindgen(self, state: 'ModuleState', args: T.List, kwargs: 'FuncBindgen') ->
         name: str
         if isinstance(header, File):
             name = header.fname
+        elif isinstance(header, (BuildTarget, BothLibraries, ExtractedObjects)):
+            raise InterpreterException('bindgen source file must be a C header, not an object or build target')
```
nirbheek, line 207: Why was this added? It seems unsafe to put this into a stable release?

nirbheek: Actually, this whole commit seems to be unrelated to the PR?

bonzini: It's in the commit message (not your fault, GitHub review is horrible):

> This also fixes a mypy failure from the next patch, since ExtractedObjects does not have a zero-argument get_outputs method.

nirbheek: Okay. So I'm wondering, is it possible that someone is doing this nonsensical thing in their build files and it accidentally works somehow right now, and it will break if this goes into a stable release?

bonzini: In other words, ExtractedObjects as in the new testcase would have caused a Python backtrace, while BuildTarget and BothLibraries simply do not make sense in this position (I tried quickly `bindgen /bin/ls` and bindgen panics very soon).

nirbheek: Okay, thanks :)

nirbheek:
LGTM

PR #9372, opened 2021-10-08
Add  a module of xorg helpers

dcbaker:
Xorg traditionally has it's own set of autoconf helper macros. Many of these are rendered basically useless by Meson, but there are still some that are useful, and require a great deal of copy/pasting to get right. This module is meant to solve that problem.

The first thing it addds is a helper to configure Xorg manpages, which use a common schema and sed. This doesn't make use of sed (which isn't portable), but instead uses a small python helper that does regex replace on files. I have a port of libxrandr here that uses this module here: https://gitlab.freedesktop.org/dbaker/libxrandr/-/tree/meson


eli-schwartz:
This doesn't need to use sed, string replacement would probably work fine too... That being said, sed is probably simpler than using python at all, and... it doesn't matter if sed is portable to Windows, since sed is portable to every OS that xorg runs on.

So would a configure_file be, though. Either triggered by this module or triggered by manual code in meson.build which tbh doesn't look all that complicated?

Even less complicated if the 4 replacements that are constant data were replaced in source, hmm...

mesonbuild/modules/unstable_xorg.py
```diff
@@ -0,0 +1,140 @@
...
+        super().__init__()
+        self.methods.update({
+            'format_man': self.format_man_method,
+        })
+
+    def _uses_sysv_man_sections(self) -> bool:
+        """Returns true if using sysv man sections."""
+        # XXX: This is only correct for build == host
+        # Copied from xorg-macros
+        return os.path.exists('/usr/share/man/man7/attributes.7')
```
eli-schwartz, line 69: Is this really the only way to detect the condition? Seems risky. A project option would be better.

mesonbuild/modules/unstable_xorg.py
```diff
@@ -0,0 +1,140 @@
...
+        'file': '4',
+    }
+}
+
+
+class XorgModule(NewExtensionModule):
+
+    """Module for upstream xorg development.
+
+    This module contains helper and abstractions fo upstream xorg projects to
```
eli-schwartz, line 54:
```suggestion
    This module contains helper and abstractions for upstream xorg projects to
```

mesonbuild/modules/unstable_xorg.py
```diff
@@ -0,0 +1,140 @@
...
+        """Returns true if using sysv man sections."""
+        # XXX: This is only correct for build == host
+        # Copied from xorg-macros
+        return os.path.exists('/usr/share/man/man7/attributes.7')
+
+    @typed_pos_args('xorg.format_man', (str, mesonlib.File), str)
+    @typed_kwargs('xorg.format_man', KwargInfo('apploaddir', str, default=''))
+    def format_man_method(self, state: 'ModuleState', args: T.Tuple['mesonlib.FileOrString', str],
+                          kwargs: 'FormatManKW') -> ModuleReturnValue:
+        """Create a custom target that format's a man file.
```
eli-schwartz, line 75:
```suggestion
        """Create a custom target that formats a man file.
```

mesonbuild/modules/unstable_xorg.py
```diff
@@ -0,0 +1,140 @@
...
+        infile, section = args
+        if section not in sections:
+            raise InvalidArguments(
+                f'xorg-devel.format_man() second argument must be one of "{", ".join(sorted(sections))}", not "{section}"')
+
+        if isinstance(infile, str):
+            # In case this is passed in the form `man/foo.man`
+            name = os.path.basename(infile)
+        else:
+            name = os.path.join(state.subdir, infile.fname)
```
eli-schwartz, line 113: This is only used for the custom target description. Why add the current subdir to this, but only if passed as a File object?

mesonbuild/modules/unstable_xorg.py
```diff
@@ -0,0 +1,140 @@
...
+            ('__vendorversion__', f'"{state.project_version}" "X Version 11"'),
+            ('__xconfigfile__', 'xorg.conf'),
+            ('__xorgversion__', f'"{state.project_name} {state.project_version}" "X Version 11"'),
+            ('__xservername__', 'Xorg'),
+        ]
+
+        infile, section = args
+        if section not in sections:
+            raise InvalidArguments(
+                f'xorg-devel.format_man() second argument must be one of "{", ".join(sorted(sections))}", not "{section}"')
```
eli-schwartz, line 107: This is unstable_xorg, what is "xorg-devel"?

mesonbuild/modules/unstable_xorg.py
dcbaker: I don't know if it's even really useful to do this, @alanc would know I think. you really don't want a project optinon because in 99% of cases you don't want the old solaris sections.

mesonbuild/modules/unstable_xorg.py
dcbaker: Ah, originally I called this `xorg-devel` because it's not for users of xorg, it's for actual xorg upstream. I dropped that because I didn't think it was useful, and I must have missed this usage of `xorg-devel`

xclaesse:
> That being said, sed is probably simpler than using python at all, and... it doesn't matter if sed is portable to Windows, since sed is portable to every OS that xorg runs on

I think xorg is larger than just X. Since @dcbaker works on Mesa I assume it's used there too, and mesa does build on Windows these days, there is a lot of effort put by Microsoft themself into that direction AFAIK. But @dcbaker certainly knows better.

mesonbuild/modules/unstable_xorg.py
Toasterson: In theory what @alanc says is true, but in practice most OSS have not bothered in a while so we have it mixed now. FreeIPMI misc man page for example is in section 7 same as on Linux. So just update it and we will fix our own pages eventually.

For Uname checks that need to identify illumos use `uname -o`

bonzini:
Is this is a problem that is common to all users of Meson, and therefore we want to support textual strings for manpage sections even in install_man etc.?

alanc:
The specific problem originally targeted here (which XFree86 originally solved by using the C Pre-Processor on man pages, much as it did to generate Makefiles via Imake), was universal at the time (the end of the previous millennium) - System V had adopted different man page sections than those used by BSD & Linux systems, and would affect any software delivering man pages.  Since [Solaris adopted the common section numbers in the 11.4 release](https://blogs.oracle.com/solaris/post/normalizing-man-page-section-numbers-in-solaris-114) and [illumos is discussing the same change](https://github.com/illumos/ipd/blob/master/ipd/0004/README.md), I don't know if this will be a problem remaining for any platform Meson supports now.

Of course, once the preprocessing was in place (either the original cpp or later sed style), it got used for other things, such as substituting version numbers and installation paths from the build configuration, or dropping sections for features disabled in the build configuration, and those still seem like universal problems, not limited to Xorg.  Of course, this can also be solved via universal mechanisms, as the xorg-server package has done by treating the man pages as configuration output in Meson:
- https://gitlab.freedesktop.org/xorg/xserver/-/blob/master/meson.build#L663
- https://gitlab.freedesktop.org/xorg/xserver/-/blob/master/meson.build#L809
- https://gitlab.freedesktop.org/xorg/xserver/-/commit/2e497bf887aca832dc0dd30d071c5288ab5c1e15

dcbaker:
Right, at this point I don't know if the method here is useful. I do have some patches to add an option for xtrans to this module that might be useful.

PR #9377, opened 2021-10-09
docs: Add a JSON documentation backend

mensinda:
This PR adds a new JSON backend to the RefMan generator. The produced JSON document is designed for IDE integration to provide a convenient-to-use representation of the Meson reference manual. The final JSON file can be previewed [here](https://github.com/mensinda/meson/releases/tag/0.0.4-json-docs). It will be automatically attached to every new Meson release with a GitHub action.

Ping @dcbaker, since this could also be interesting for the VS Code Meson plugin.

Since this PR doesn't touch the main code-base it *should* be save to merge it even the first RC release for 0.60.0 is missed.

### DIFF
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -12,6 +12,9 @@ on:
     paths:
       - docs/**
   workflow_dispatch:
+  release:
+    types:
+      - published
 
 # This job is copy/paster into wrapdb CI, please update it there when doing any
 # change here.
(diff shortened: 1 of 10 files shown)


mensinda:
Ping, since I would like to get this into 0.60.0, so that the docs can be published with it.

jpakkane:
Since this is a doc-only change it can be added after rc2 if needed.

Issue #9374, opened 2021-10-08
SDL2 game generated with meson won't compile correctly


Veinhelm:
OS: Windows 10
Python: 3.10.0
Meson: 0.59.2
Ninja: 1.10.2

Whenever I try to compile my SDL2 game, Meson gives me an error saying:
LINK : fatal error LNK1181: cannot open input file 'libvendor\SDL2\lib\x64\SDL2.lib.a'
ninja: build stopped: subcommand failed.

Here is my meson.build file
```
project('sdltest', 'cpp')

src = [
'src/main.cpp',
'src/RenderWindow.cpp',
'src/RenderWindow.h',
'src/Entity.cpp',
'src/Entity.h',
'src/Math.h'
]

SDL2 = static_library('vendor/SDL2/lib/x64/SDL2.lib')
SDL2main = static_library('vendor/SDL2/lib/x64/SDL2main.lib')
SDL2_image = static_library('vendor/SDL2/lib/x64/SDL2_image.lib')
FMOD = static_library('vendor/FMOD/lib/x64/fmod_vc.lib')

deps = [SDL2, SDL2main, SDL2_image, FMOD]
inc = include_directories('vendor/SDL2/include', 'vendor/FMOD/inc')

executable('sdltest', sources : src, include_directories : inc, link_with : deps)
```
I'm pretty sure im doing everything right..?
Note: Im using the Visual Studio backend, I tested it with Ninja and it gives the same error

PR #9378, opened 2021-10-09
Be more strict about target names with slashes.

jpakkane:


eli-schwartz:
Presumably a response to my comment at https://github.com/mesonbuild/meson/issues/9374#issuecomment-939142581

I don't have a problem with this change, but...

So now we have dead code that warns if you do this that it's bad, probably won't work, and will become an error in the future... But that warning can't be reached anymore?

jpakkane:
> But that warning can't be reached anymore?

It can, if the segment points to a dir that does not exist in the source tree. Which kinda sorta almost works and is used under some weird circumstances (i.e. picolibc uses it to build something like 50-100 targets all called `libc`). Fixing that properly would require new functionality, I think.

> Seems like the correct place for this fix would be 

That is where I originally put it, but it won't work. You need source root directory to determine whether the path exists or not.

PR #9379, opened 2021-10-09
Add NASM compiler

xclaesse:
Currently Meson can only compile `.s` assembly files when the C compiler supports it. To compile `.asm` files with NASM we need a generator, see for example: https://code.videolan.org/videolan/dav1d/-/blob/master/meson.build#L442.

This adds a new language `asm` to build `.asm` files using NASM.

Note for example that OpenSSL generates .s files on Linux that build with gcc, but .asm files on Windows that build with nasm, so this is going to be useful there too.

### DIFF
--- a/docs/markdown/snippets/asm.md
+++ b/docs/markdown/snippets/asm.md
@@ -1,0 +1,18 @@
+## New language `nasm`
+
+When the `nasm` language is added to the project, `.asm` files are
+automatically compiled with NASM. This is only supported for x86 and x86_64 CPU
+family.
+
+Support for other compilers compatible with NASM language, such as YASM, could
+be added in the future.
+
+Note that GNU Assembly files usually have `.s` extension and were already built
+using C compiler such as GCC or CLANG.
+
+```meson
+project('test', 'nasm')
+
+exe = executable('hello', 'hello.asm')
+test('hello', exe)
+```
--- a/docs/markdown/snippets/asm.md
+++ b/docs/markdown/snippets/asm.md
@@ -2,10 +2,7 @@
 
 When the `nasm` language is added to the project, `.asm` files are
 automatically compiled with NASM. This is only supported for x86 and x86_64 CPU
-family.
-
-Support for other compilers compatible with NASM language, such as YASM, could
-be added in the future.
+family. `yasm` is used as fallback if `nasm` command is not found.
 
 Note that GNU Assembly files usually have `.s` extension and were already built
 using C compiler such as GCC or CLANG.
--- a/mesonbuild/compilers/asm.py
+++ b/mesonbuild/compilers/asm.py
@@ -1,7 +1,7 @@
 import os
 import typing as T
 
-from ..mesonlib import EnvironmentException
+from ..mesonlib import EnvironmentException, get_meson_command
 from .compilers import Compiler
 
 if T.TYPE_CHECKING:
@@ -75,3 +75,21 @@ class NasmCompiler(Compiler):
             if i[:2] == '-I':
                 parameter_list[idx] = i[:2] + os.path.normpath(os.path.join(build_dir, i[2:]))
         return parameter_list
+
+class YasmCompiler(NasmCompiler):
+    id = 'yasm'
+
+    def get_exelist(self) -> T.List[str]:
+        # Wrap yasm executable with an internal script that will write depfile.
+        exelist = super().get_exelist()
+        return get_meson_command() + ['--internal', 'yasm'] + exelist
+
+    def get_debug_args(self, is_debug: bool) -> T.List[str]:
+        if is_debug:
+            if self.info.is_windows():
+                return ['-g', 'null']
+            return ['-g', 'dwarf2']
+        return []
+
+    def get_dependency_gen_args(self, outtarget: str, outfile: str) -> T.List[str]:
+        return ['--depfile', outfile]
--- a/mesonbuild/compilers/asm.py
+++ b/mesonbuild/compilers/asm.py
@@ -1,0 +1,77 @@
+import os
+import typing as T
+
+from ..mesonlib import EnvironmentException
+from .compilers import Compiler
+
+if T.TYPE_CHECKING:
+    from ..environment import Environment
+
+
+class NasmCompiler(Compiler):
+    language = 'nasm'
+    id = 'nasm'
+
+    def needs_static_linker(self) -> bool:
+        return True
+
+    def get_always_args(self) -> T.List[str]:
+        cpu = '64' if self.info.is_64_bit else '32'
+        if self.info.is_windows() or self.info.is_cygwin():
+            plat = 'win'
+            define = f'WIN{cpu}'
+        elif self.info.is_darwin():
+            plat = 'macho'
+            define = 'MACHO'
+        else:
+            plat = 'elf'
+            define = 'ELF'
+        args = ['-f', f'{plat}{cpu}', f'-D{define}']
+        if self.info.is_64_bit:
+            args.append('-D__x86_64__')
+        return args
+
+    def get_werror_args(self) -> T.List[str]:
+        return ['-Werror']
+
+    def get_output_args(self, outputname: str) -> T.List[str]:
+        return ['-o', outputname]
+
+    def get_optimization_args(self, optimization_level: str) -> T.List[str]:
+        return [f'-O{optimization_level}']
+
+    def get_debug_args(self, is_debug: bool) -> T.List[str]:
+        if is_debug:
+            if self.info.is_windows():
+                return []
+            return ['-g', '-F', 'dwarf']
+        return []
+
+    def get_depfile_suffix(self) -> str:
+        return 'd'
+
+    def get_dependency_gen_args(self, outtarget: str, outfile: str) -> T.List[str]:
+        return ['-MD', '-MQ', outtarget, '-MF', outfile]
+
+    def sanity_check(self, work_dir: str, environment: 'Environment') -> None:
+        if self.info.cpu_family not in {'x86', 'x86_64'}:
+            raise EnvironmentException(f'ASM compiler {self.id!r} does not support {self.info.cpu_family} CPU family')
+
+    def get_buildtype_args(self, buildtype: str) -> T.List[str]:
+        # FIXME: Not implemented
+        return []
+
+    def get_pic_args(self) -> T.List[str]:
+        return []
+
+    def get_include_args(self, path: str, is_system: bool) -> T.List[str]:
+        if not path:
+            path = '.'
+        return ['-I' + path]
+
+    def compute_parameters_with_absolute_paths(self, parameter_list: T.List[str],
+                                               build_dir: str) -> T.List[str]:
+        for idx, i in enumerate(parameter_list):
+            if i[:2] == '-I':
+                parameter_list[idx] = i[:2] + os.path.normpath(os.path.join(build_dir, i[2:]))
+        return parameter_list
--- a/mesonbuild/compilers/detect.py
+++ b/mesonbuild/compilers/detect.py
@@ -88,6 +88,7 @@ defaults['clang_cl_static_linker'] = ['llvm-lib']
 defaults['cuda_static_linker'] = ['nvlink']
 defaults['gcc_static_linker'] = ['gcc-ar']
 defaults['clang_static_linker'] = ['llvm-ar']
+defaults['nasm'] = ['nasm']
 
 
 def compiler_from_language(env: 'Environment', lang: str, for_machine: MachineChoice) -> T.Optional[Compiler]:
@@ -105,6 +106,7 @@ def compiler_from_language(env: 'Environment', lang: str, for_machine: MachineCh
         'fortran': detect_fortran_compiler,
         'swift': detect_swift_compiler,
         'cython': detect_cython_compiler,
+        'nasm': detect_nasm_compiler,
     }
     return lang_map[lang](env, for_machine) if lang in lang_map else None
 
@@ -1133,6 +1135,35 @@ def detect_swift_compiler(env: 'Environment', for_machine: MachineChoice) -> Com
             exelist, version, for_machine, is_cross, info, linker=linker)
 
     raise EnvironmentException('Unknown compiler: ' + join_args(exelist))
+
+def detect_nasm_compiler(env: 'Environment', for_machine: MachineChoice) -> Compiler:
+    from .asm import NasmCompiler
+    compilers, _, _ = _get_compilers(env, 'nasm', for_machine)
+    is_cross = env.is_cross_build(for_machine)
+
+    # We need a C compiler to properly detect the machine info and linker
+    cc = detect_c_compiler(env, for_machine)
+    if not is_cross:
+        from ..environment import detect_machine_info
+        info = detect_machine_info({'c': cc})
+    else:
+        info = env.machines[for_machine]
+
+    popen_exceptions: T.Dict[str, Exception] = {}
+    for comp in compilers:
+        try:
+            output = Popen_safe(comp + ['--version'])[1]
+        except OSError as e:
+            popen_exceptions[' '.join(comp + ['--version'])] = e
+            continue
+
+        version = search_version(output)
+        if 'NASM' in output:
+            comp_class = NasmCompiler
+            env.coredata.add_lang_args(comp_class.language, comp_class, for_machine, env)
+            return comp_class(comp, version, for_machine, info, cc.linker, is_cross=is_cross)
+    _handle_exceptions(popen_exceptions, compilers)
+    raise EnvironmentException('Unreachable code (exception to make mypy happy)')
 
 
 # GNU/Clang defines and version
--- a/mesonbuild/compilers/detect.py
+++ b/mesonbuild/compilers/detect.py
@@ -88,7 +88,7 @@ defaults['clang_cl_static_linker'] = ['llvm-lib']
 defaults['cuda_static_linker'] = ['nvlink']
 defaults['gcc_static_linker'] = ['gcc-ar']
 defaults['clang_static_linker'] = ['llvm-ar']
-defaults['nasm'] = ['nasm']
+defaults['nasm'] = ['nasm', 'yasm']
 
 
 def compiler_from_language(env: 'Environment', lang: str, for_machine: MachineChoice) -> T.Optional[Compiler]:
@@ -1137,7 +1137,7 @@ def detect_swift_compiler(env: 'Environment', for_machine: MachineChoice) -> Com
     raise EnvironmentException('Unknown compiler: ' + join_args(exelist))
 
 def detect_nasm_compiler(env: 'Environment', for_machine: MachineChoice) -> Compiler:
-    from .asm import NasmCompiler
+    from .asm import NasmCompiler, YasmCompiler
     compilers, _, _ = _get_compilers(env, 'nasm', for_machine)
     is_cross = env.is_cross_build(for_machine)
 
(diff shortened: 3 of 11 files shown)


mesonbuild/compilers/asm.py
```diff
@@ -0,0 +1,83 @@
...
+    language = 'asm'
+    id = 'nasm'
+
+    def needs_static_linker(self) -> bool:
+        return False
+
+    def get_always_args(self) -> T.List[str]:
+        if self.info.is_windows():
+            plat = 'win'
+        elif self.info.is_linux() or self.info.is_android():
```
dcbaker, line 21: This is going to fail on the BSDs, Haiku, Solaris, and AIX, all of which we vaguely support, and all of which use elf.

mesonbuild/compilers/asm.py
```diff
@@ -0,0 +1,83 @@
...
+    def get_optimization_args(self, optimization_level: str) -> T.List[str]:
+        return [f'-O{optimization_level}']
+
+    def get_depfile_suffix(self) -> str:
+        return 'd'
+
+    def get_dependency_gen_args(self, outtarget: str, outfile: str) -> T.List[str]:
+        return ['-MD', '-MQ', outtarget, '-MF', outfile]
+
+    def sanity_check(self, work_dir: str, environment: 'Environment') -> None:
```
dcbaker, line 45: since nasm only works on x86 (I think), it might be worth short cirucuiting here since we have the environment and can check if we're on x86 or x86_64

mesonbuild/compilers/detect.py
```diff
@@ -1164,6 +1167,31 @@ def detect_swift_compiler(env: 'Environment', for_machine: MachineChoice) -> Com
...
+            popen_exceptions[' '.join(comp + ['-v'])] = e
+            continue
+
+        version = search_version(output)
+        if 'NASM' in output:
+            comp_class = NasmCompiler
+            env.coredata.add_lang_args(comp_class.language, comp_class, for_machine, env)
+            return comp_class(comp, version, for_machine, info, cc.linker, is_cross=is_cross)
+    _handle_exceptions(popen_exceptions, compilers)
+    raise EnvironmentException('Unreachable code (exception to make mypy happy)')
```
dcbaker, line 1193: this shouldn't be required, mypy konws that `_handle_exceptions` is `NoReturn` and should be happy with that.

dcbaker (changes requested):
I'm not sure about this, it kinda makes sense, but on the other hand there are so many assembly formats, does it really make sense to hard code support in the language like this? I almost wonder if a module would be more appropriate.

eli-schwartz:
(Gah, accidentally clicked the screen on mobile without meaning to.)

Since some C compilers, specifically on Linux, this works fine OOTB... shouldn't we detect that as a valid asm compiler too?

dcbaker:
To add to my concerns, there isn't really a standard for what ".asm" means. A lot of assembler with incompatible syntax use that extension, not just nasm. This is differen than say .c, where every C compiler is more or less compatible.

tp-m:
> This adds a new language `asm` to build `.asm` files using NASM.

What makes NASM so special? I mean, some projects want NASM, others YASM, others yet something else.

mesonbuild/compilers/asm.py
xclaesse: Good point. That code was just place holder I copied form nasm tutorial to see if it works.

xclaesse:
> A lot of assembler with incompatible syntax use that extension, not just nasm.

> What makes NASM so special?

I used nasm because it's the one used by both projects where I saw they need to build asm: OpenSSL and vlc. I could be wrong but it seems to most popular?

But you're right, the language is not "asm" but "nasm", the compiler is it's owns language as I understand. I'll change that so we explicitly add nasm as language. If in the future we add yasm language we'll have to forbid adding both languages in the same project. Does that sound good?

xclaesse:
Actually, after reading a bit stackoverflow, it seems nasm, yasm, tasm, etc all support the same Intel asm language, is that right? In that case it's fine to name the language "asm" and we can always add other compilers later, in the same way we have gcc and clang that both can do .c files.

As far as I can tell, for other arch than Intel, you have no choice but using GAS, the assembly from gnu/gcc. Right?

mesonbuild/compilers/detect.py
xclaesse: done.

mesonbuild/compilers/asm.py
xclaesse: done.

mesonbuild/compilers/detect.py
xclaesse: Sorry, no, mypy is not that smart: `mesonbuild/compilers/detect.py:666:1: error: Missing return statement`

xclaesse:
Added a unit test and a release snippet. Do we have a list of supported languages somewhere in our doc??

I gave a try to YASM, but its support for depfile is weird. It has -M argument that make it output a Makefile-style depfile to stdout and does not compile the file in that case. So we have to invoke yasm twice, I don't know if we have any support for that kind of thing...

bonzini:
I think the common cases are nasm/yasm and gas/llvm. OpenSSL's preprocessing mechanism is quite unique. Using masm on Windows is also less common for projects that use Meson, due to lack of cross compilation.

All in all, "either nasm or gas on x86, and gas on other architectures" is probably the common case. Either way, 'nasm'  and 's' should both be their own languages.

eli-schwartz:
> The assembly language itself is almost the same (though you have `BYTE` in nasm/yasm vs. `BYTE PTR` in Intel syntax), the macros and the directives are completely different

Software is pain. :( Okay, I guess people really really need specifically nasm and nothing else then.

xclaesse:
To summary the different assembly use-cases we have so far:
- `.asm` -> nasm - covered by this PR.
- `.asm` -> [run_yasm.py](https://chromium.googlesource.com/chromium/src/third_party/yasm/+/530de22e88b9b90d6e0711ea6cfcd13e6eae7e7f/run_yasm.py) - trivial python wrapper for yasm that we could copy into meson to generate depfile, as fallback to nasm.
- `.s` -> gcc - meson already do this.
- `.s` -> [gas-preprocessor.pl](https://github.com/libav/gas-preprocessor/blob/master/gas-preprocessor.pl) -> clang - could translate that perl script to python and import im Meson.
- `.s` -> `cl /P` -> `ml`, `ml64`, `armasm`, or `armasm64` - suggested in https://github.com/mesonbuild/meson/issues/9889.

xclaesse:
This now supports yasm as fallback, just like FFmpeg does.

docs/markdown/snippets/asm.md
```diff
@@ -0,0 +1,17 @@
+## New language `nasm`
+
+When the `nasm` language is added to the project, `.asm` files are automatically
```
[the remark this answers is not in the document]
xclaesse: I've never seen a project using `.nasm` extension, do you have examples in mind? At least vlc, ffmpeg and openssl all use `.asm` for nasm compiler.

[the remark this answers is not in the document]
eli-schwartz: The goal here isn't to introduce a convention for people to add asm to new projects.

The goal is to fit into the convention that already exists.

The use of `.asm` is intentional and shall not change.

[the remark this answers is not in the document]
xclaesse: True, I would like to follow what people actually do in real world projects out there, certainly not invent something new. If we have examples of people using `.nasm` then we can of course also associate that extension too.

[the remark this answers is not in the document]
bonzini: Here is one: https://github.com/tianocore/edk2/blob/master/OvmfPkg/Sec/X64/SecEntry.nasm. The problem is twofold:

* `.asm` can mean multiple things, it can mean both Microsoft and nasm syntaxes.  edk2 started to use `.nasm` because they wanted to use separate build rules for Microsoft and nasm syntaxes. Identifying build rules by the extension is something that Meson already does (e.g. anything ending with `*.c` _must_ be compiled with the C compiler).

* it's not even sure that assembly language it would require a separate `add_languages` on any architecture but x86.

There are other cases in which Meson says my way or the highway (e.g. no include files named `*.inc.c`), and projects that switch to Meson have to adapt (hello QEMU). I think in this case using `.nasm` is safer until OS-portable support for assemblers is fleshed out on other architectures. Adding stuff later is much easier than removing it.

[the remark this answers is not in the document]
xclaesse: Ok, I could definitely map `.nasm ` to nasm language, that's unambiguous. I personally like the convention `.S` files goes through the C compiler, which can be either gcc/as or cl/ml, that seems unambiguous too.

But it's true that plain `.asm` can be ambiguous, ultimately it will have to be configurable I guess. How does edk2 compile their `.asm` currently?

> In the future we might want to allow `.asm` with MSVC just like we allow `.s` with GCC

Do you have examples of projects that does that? edk2? Maybe if we user does `add_language('nasm')` then `.asm` goes to nasm, otherwise if C compiler is MSVC then they go to cl/ml? Another solution that I had in mind for other non conventional C/C++ file extensions is adding an explicit API to map extension to language, such as `add_language('c', suffixes: ['.asm'])`.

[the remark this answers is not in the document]
xclaesse: In the end here we only want to help the "common" case, it's fine to keep using custom_target()/generator() for more "exotic" conventions, that's already what they have to do right now anyway.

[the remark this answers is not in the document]
bonzini: > Do you have examples of projects that does that? edk2

Yes. Though by now I think they have converted everything to just nasm, they used to have some files as .S/.asm (synced by hand :)) and some as .nasm.

My suggestion is simply to not touch .asm with a ten foot pole for x86, until you're satisfied with the handling of .asm in other architectures. Assigning extensions in `add_languages` makes sense but again I would be wary of adding stuff that later will be unnecessary, or worse a poor design.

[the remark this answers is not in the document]
bonzini: >  I personally like the convention .S files goes through the C compiler, 

The problem is that .S on MSVC is unheard of. They use .asm. :(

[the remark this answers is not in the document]
xclaesse: > The problem is that .S on MSVC is unheard of. They use .asm. :(

At least libffi does it:
```
[1/9] "C:\BuildTools\VC\Tools\MSVC\14.29.30037\bin\HostX64\x64\cl.EXE" "/nologo" "/EP" "/P" "/Fisrc/x86_win64_intel_S.i" "/DTARGET=X86_WIN64" "/DFFI_BUILDING" "../src/x86/win64_intel.S" "/IC:/Users/Collabora/xclaesse/libffi/src/.." "/IC:/Users/Collabora/xclaesse/libffi/build/src/.." "/IC:/Users/Collabora/xclaesse/libffi/src/../include" "/IC:/Users/Collabora/xclaesse/libffi/build/src/../include"
win64_intel.S
[7/9] "C:\BuildTools\VC\Tools\MSVC\14.29.30037\bin\HostX64\x64\ml64.EXE" "/Fo" "src/x86_win64_intel_S.obj" "/c" "src/x86_win64_intel_S.i" "/Zi" "/Zd"
Microsoft (R) Macro Assembler (x64) Version 14.29.30038.1
Copyright (C) Microsoft Corporation.  All rights reserved.

 Assembling: src/x86_win64_intel_S.i
```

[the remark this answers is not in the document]
ePirat: I too would be in favor of using `.nasm` to make it explicit.

ePirat:
Regarding the yasm fallback, is it guaranteed you can always use that for nasm files? IIRC recently nasm gained some features that yasm did not, so not sure automatically falling back to yasm works in general…

docs/markdown/snippets/asm.md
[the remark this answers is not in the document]
xclaesse: The question is not what we prefer, but what projects actually do. So far every projects I've seen need to build .asm with nasm. If a project has .asm files that are built with another compiler they just have to not add nasm language.

xclaesse:
> Regarding the yasm fallback, is it guaranteed you can always use that for nasm files

I'm sure they have differences, a project that only support nasm can always do `find_program('nasm')` to be sure it won't fallback.

ePirat:
Yes of course, but how would I deal with the differences of the tools? As the flags they accept are not 100% the same? And what would I do if I need yasm instead of nasm?

xclaesse:
> As the flags they accept are not 100% the same

You can use `meson.get_compiler('nasm').get_id() == 'yasm'` condition to add different flags.

> And what would I do if I need yasm instead of nasm?

Are there real use case for that? You can force yasm with NASM=yasm just like you can do CC=clang to prefer clang over gcc.

ePirat:
@xclaesse I guess thats good enough initially.

eli-schwartz:
It's not like there aren't already projects that require either clang or gcc and fail if you use the other one.

xclaesse:
Right the problem is not new and could need a common solution. Just on the top of my head we could add `add_language('c', id: 'clang')`

xclaesse:
This is ready to be reviewed.

docs/markdown/snippets/asm.md
```diff
@@ -0,0 +1,18 @@
+## New language `nasm`
+
+When the `nasm` language is added to the project, `.asm` and `.nasm` files are
```
bonzini, line 3: I really would prefer to have `.nasm` only for NASM, at least for the time being. It's always safer to be strict first and relax the requirements later.

If we want to add MASM support later, we would have the same issue where preexisting projects use `.asm` for MASM sources. And then introduce the special case that MASM conflicts with NASM, or having to add `suffixes:` to `add_language` or other horrible things.

Let's keep Meson opinionated. :)

xclaesse: > I really would prefer to have `.nasm` only for NASM

That would make this PR basically useless, I never seen any project using that extension. They would have top copy all their .asm files to the builddir and rename them. I'm trying to follow what projects actually do, and `.asm` for NASM is by far the most common.

> If we want to add MASM support later, we would have the same issue where preexisting projects use `.asm` for MASM sources. And then introduce the special case that MASM conflicts with NASM, or having to add `suffixes:` to `add_language` or other horrible things.

I think that's a non-issue actually. MASM is itself much more rare than NASM, and I would be very surprised any project mix nasm and masm at the same time, for the same platform. We could make both of them compile `.asm` files and just make adding those languages mutually exclusive.

ePirat: I agree that it would be better to just use `.nasm` for NASM, we can always relax it later if really needed but the other way around not…

ePirat: > They would have top copy all their .asm files to the builddir and rename them.

Why would they need to copy their source files instead of just rename them once they adopt native meson nasm support?

IMHO ideally we would support `.asm` for which meson will pick either nasm or yasm, `.nasm` for which meson will only use nasm and `.yasm` for which meson will only use yasm, but I am not sure how hard it would be to fit that into how meson currently handles languages/extensions…

bonzini: @xclaesse I don't care about what projects do *now*, I only care if projects can adapt. It wouldn't be the first time Meson forces you to follow Meson conventions, and that's a feature, not a bug.

xclaesse: > Why would they need to copy their source files instead of just rename them once they adopt native meson nasm support?

Yes but only for projects that actually adopt meson, not wrapdb ports for example. To be fair, projects I would like to use it are all non upstream meson ports: libjpeg-turbo, ffmpeg, libffi.

> IMHO ideally we would support `.asm` for which meson will pick either nasm or yasm, `.nasm` for which meson will only use nasm and `.yasm` for which meson will only use yasm

It's true that currently it would fallback to use yasm to compile .nasm file which could be surprising. I would like to keep yasm and nasm as the same language at the meson level, because it's pretty much like gcc/clang, 2 compilers for the same language even if not 100% compatible. To be honest, I'm getting tempted to actually remove `.nasm` completely, I don't want projects to start renaming their files to `.nasm` if they can actually be built with yasm too. If a project really supports only yasm or only nasm, they should rather do:
```
asm_id = meson.get_compiler('nasm').get_id()
if asm_id != 'yasm'
  error('YASM compiler is required but found' + asm_id)
endif
```
That's basically the same situation for a project that requires GCC and not CLANG, or vice versa.

That being said, assigning .nasm to only NASM and .yasm to only YASM is easy to do from an implementation POV. I'm just not sure there is an actual use-case for it...

ePirat: I think you overestimate how compatible yasm/nasm nowadays are, it's I do not think it is uncommon that projects work just with one or the other. While this is frequent for yasm/nasm it is rare for C compilers.

So either we should:

- Make `.nasm` and `.yasm` be limited to nasm/yasm respectively
- Drop `.nasm` and `.yasm` extensions altogether

Everything else will just be way too confusing…

xclaesse: > I think you overestimate how compatible yasm/nasm nowadays are, I do not think it is uncommon that projects work just with one or the other. While this is frequent for yasm/nasm it is rare for C compilers.

gcc/clang was a bad example, they are mostly compatible. gcc/msvc is a better example.

xclaesse: @ePirat @bonzini Updated the PR to only accept `.asm` suffix and build them with either nasm or yasm. We can easily add `.nasm` and `.yasm` suffixes later.

test cases/nasm/2 asm language/meson.build
```diff
@@ -0,0 +1,26 @@
+project('test', 'c')
+
+nasm = find_program('nasm', required: false)
```
nirbheek, line 3: Does this check work with a binary subproject providing nasm like we do in gstreamer?

https://gitlab.freedesktop.org/gstreamer/gstreamer/-/blob/main/subprojects/win-nasm/meson.build#L24

docs/markdown/snippets/asm.md
nirbheek: fwiw, I agree that we shouldn't invent a new extension, and using nasm to handle `.asm` files is fine. If we want to change this later, we could add a new kwarg to `add_language()` which specifies which file extensions we want that compiler to handle or something like that. 

Picking a default doesn't mean it will be the only option going forward. Just means we can't change the default without a long deprecation process.

mesonbuild/compilers/asm.py
```diff
@@ -0,0 +1,95 @@
...
+        elif self.info.is_darwin():
+            plat = 'macho'
+            define = 'MACHO'
+        else:
+            plat = 'elf'
+            define = 'ELF'
+        args = ['-f', f'{plat}{cpu}', f'-D{define}']
+        if self.info.is_64_bit:
+            args.append('-D__x86_64__')
+        return args
```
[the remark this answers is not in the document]
nirbheek: I think this is a good idea.

test cases/nasm/2 asm language/meson.build
xclaesse: That check if for the unit test only. I don't think meson actually look for compilers using overridden programs, so it probably won't work for gstreamer indeed. The path needs to come from a machine file, or from NASM=/path/nasm.

xclaesse: @nirbheek why does GStreamer install nasm like that instead of requiring user to install it using choco for example? To make things easier, I added a commit in this PR to lookup nasm into its default location because it is not in PATH by default on Windows.

xclaesse: This is fixed with https://github.com/mesonbuild/meson/pull/10938, but I don't think it should block this PR.

nirbheek (approved):

PR #9382, opened 2021-10-10
Revert "interpreter: Add FeatureNew check"

eli-schwartz:
This reverts commit c0efa7ab22f8900f6fa1dadf0d306ec375569c8d.

This was a nice idea, or a beautiful hack depending on your perspective. Unfortunately, it turns out to be a lot harder than we originally thought. By operating on bare nodes, we end up triggering a FeatureNew on anything that isn't a string literal, rather than anything that isn't a string.

Since no one else has come up with a better idea for implementing a FeatureNew, let's just revert it. Better to not have a warning, than have it trigger way too often.

### DIFF
--- a/mesonbuild/interpreter/primitives/string.py
+++ b/mesonbuild/interpreter/primitives/string.py
@@ -21,11 +21,6 @@ from ...interpreterbase import (
 
     InvalidArguments,
 )
-from ...mparser import (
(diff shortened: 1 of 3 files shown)


mensinda (approved):
LGTM, thanks for rewerting, forgot about it myself :)

PR #9384, opened 2021-10-10
pkg-config: do not ever successfully detect Strawberry Perl.

eli-schwartz:
This is broken and terrible and thus completely unusable. Don't torture users by finding pkg-config on Windows, thus permitting the pkg-config lookup of several dependencies that do not actually work -- which then fails at build time.

This also breaks CI for the wrapdb, because Strawberry Perl is provided as part of the base image for the OS (yes, even though it is terribly broken!!!) and anything that depends on e.g. zlib will "find" zlib because of this broken disaster, even though it should use the wrapdb subproject of zlib.

It is assumed no one actually wants to mix Strawberry Perl and meson. In fact, some projects, such as gst-build, already unconditionally error out if Strawberry Perl is detected in PATH:

error('You have Strawberry Perl in PATH which is known to cause build issues with gst-build. Please remove it from PATH or uninstall it.')

Other projects (postgresql) actually do want to build perl extensions, and link to the perl dlls, but absolutely under no circumstances ever want to use its pkg-config implementation. ;)

Let's solve this problem by just considering this to not be a valid pkg-config, let the user find another or not have one at all.

This change "solves" https://github.com/StrawberryPerl/Perl-Dist-Strawberry/issues/11


anarazel:
Checking for "Strawberry/" in the path doesn't strike me as particularly reliable - yes, that's the default, but e.g. I installed it under c:/perl/version.of.perl/ because I need to switch between multiple versions.

A bit more complicated approach would be to check if (dirname pkg-config.bat)/../../DISTRIBUTIONS.txt starts with "Strawberry Perl". But maybe that's too complicated...

mesonbuild/dependencies/pkgconfig.py
```diff
@@ -423,18 +423,20 @@ def check_pkgconfig(self, pkgbin: ExternalProgram) -> T.Optional[str]:
         if not pkgbin.found():
             mlog.log(f'Did not find pkg-config by name {pkgbin.name!r}')
             return None
+        command_as_string = ' '.join(pkgbin.get_command())
+        if r'Strawberry\perl\bin' in pkgbin.get_path():
```
xclaesse, line 427: I'm slightly worried that this check is too strict. Like get_path() could return POSIX path, or it could be in another subdir. I would rather go with `'strawberry' in pkgbin.get_path().lower()`. But in any case, yes we should black list that crap, and probably do that for any binary such as gcc.exe, etc. I just now realise it's the cause of msvc env not being activated by default on GitHub Windows CI, meson find strawberry gcc...

eli-schwartz:
```python
if seems_to_be_strawberry and self.name != 'perl':
    return [None]
```

*dies*

Because we do of course want to make sure that actual perl.exe is detected, people might want to actually use that. As a script interpreter at the very least.

anarazel:
FWIW I compiled a bunch of things successfully with strawberry perl's gcc. And used several binaries from it. Mostly because it made the CI process faster. So refusing anything from strawberry seems like a good of overkill? 
-- 

xclaesse:
> A bit more complicated approach would be to check if (dirname pkg-config.bat)/../../DISTRIBUTIONS.txt starts with "Strawberry Perl". But maybe that's too complicated...

I think that's the most reliable way indeed.

> FWIW I compiled a bunch of things successfully with strawberry perl's gcc. And used several binaries from it. Mostly because it made the CI process faster. So refusing anything from strawberry seems like a good of overkill?

I agree, let's blacklist only pkg-config for now because that one cannot be useful in any possible way.

PR #9388, opened 2021-10-11
modules/python: fix path introspection on virtual environments on Debian

FFY00:

### DIFF
--- a/mesonbuild/modules/python.py
+++ b/mesonbuild/modules/python.py
@@ -292,6 +292,17 @@ def debian_distutils_missing():
             return True
     return False
 
+def debian_distutils_deb_system():
+    # The easiest way to see if the deb_system scheme is in use is by checking
+    # the value of get_python_lib.
+    # https://salsa.debian.org/cpython-team/python3/-/blob/python3.6/debian/patches/distutils-install-layout.diff#L140
+    import distutils.command.install
+    import distutils.sysconfig
+    return (
+        'deb_system' in distutils.command.install.INSTALL_SCHEMES
+        and distutils.sysconfig.get_python_lib().endswith('dist-packages')
+    )
+
 def get_distutils_paths(scheme=None, prefix=None):
     import distutils.dist
     distribution = distutils.dist.Distribution()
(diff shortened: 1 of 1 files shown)


ev-br:
Confirm that it fixes https://github.com/ev-br/mc_lib/issues/45.

The warning is still there,
``WARNING: Broken python installation detected. Python files installed by Meson might not be found by python interpreter.
 This warning can be avoided by setting "python.platlibdir" option.
WARNING: Broken python installation detected. Python files installed by Meson might not be found by python interpreter.
```

but the installed files are in `.../python3.8/site-packages`.

eli-schwartz:
But the python module will still install to `/usr(/local)?/lib/python3.7/site-packages` rather than `/test/lib/python3.7/site-packages`, right?

And, even *with* the PR, the warning message is still appearing which states that `self.info['install_paths']['platlib']` doesn't match any location in `sys.path`. If that warning is appearing, it is supposed to mean that something is actually truly mismatched. So my assumption is the warning message means that there is a problem somewhere, not that it is a heuristic which should be ignored.

What am I missing here?

FFY00:
No, it will not. You will just be much more likely to run into #9284, which is its own separate issue. Meson will mangle the installation path if it does not match the installation prefix, but that has nothing to with the introspection.

This PR makes the introspection script return the correct path. Currently, the introspection script will use the `deb_system` scheme, which is incorrect.

In the future please be more mindful with how you reply. Stating that the "PR doesn't actually fix anything" without any properly looking into the issue, and following that with "What's the goal here?", is not very constructive. Meson makes various wrong assumptions about Python and install paths. This is a complex issue, made even more complex by Meson's inherently incompatible approach with Python's install path design. I have been trying to make it slightly more correct/reliable, but it's a frustrating getting feedback that my small fixes are wrong when the actual issue is the whole approach Meson takes to installing Python modules.	I don't know, maybe I am being overly sensitive but, well, that is the current reality of my mental health. This whole ordeal (bad downstream patching of Python locations) has already taken a pretty big toll, being the person that gets to go to the different parties and brings up a bunch of issues and tries to get them to have coherent understandings of how install locations in Python work is not very fun. Anyway, sorry for the small rant, just please don't dismiss things unless you are 100% sure, I am tired from trying to put out fires all over the place.

PR #9394, opened 2021-10-13
per-subproject compiler options... again...

xclaesse:


mesonbuild/coredata.py
```diff
@@ -40,7 +40,8 @@
     from .cmake.traceparser import CMakeCacheEntry
 
     OptionDictType = T.Union[T.Dict[str, 'UserOption[T.Any]'], OptionOverrideProxy]
-    KeyedOptionDictType = T.Union[T.Dict['OptionKey', 'UserOption[T.Any]'], OptionOverrideProxy]
+    MutableKeyedOptionDictType = T.Dict['OptionKey', 'UserOption[T.Any]']
+    KeyedOptionDictType = T.Union[MutableKeyedOptionDictType, OptionOverrideProxy]
```
dcbaker, line 44: This naming is pretty unclear, I'd assume that 'Mutable' type can be mutated, and the other is immutable.

dcbaker (approved):
The whole option dict typing stuff is kinda a mess anyway, and this doesn't make it any worse. So that comment from before doesn't seem relevant.

I don't see anything wrong with the approach or the code.

This is pretty important for Rust, mixing editions/standards is intentionally allowed, so we really *need* to be able to compile rust 2021 and 2018 code and link them together, like what I'm going to be doing with the cargo subproject support.

PR #9396, opened 2021-10-13
depfixer: drop unused and unnecessary printing code

keszybz:


mesonbuild/scripts/depfixer.py
```diff
@@ -319,17 +270,17 @@ def fix_deps(self, prefix: bytes) -> None:
...
+        self.fix_rpathtype_entry(fname, rpath_dirs_to_remove, new_rpath, DT_RPATH)
+        self.fix_rpathtype_entry(fname, rpath_dirs_to_remove, new_rpath, DT_RUNPATH)
 
-    def fix_rpathtype_entry(self, rpath_dirs_to_remove: T.Set[bytes], new_rpath: bytes, entrynum: int) -> None:
+    def fix_rpathtype_entry(self, fname: str, rpath_dirs_to_remove: T.Set[bytes], new_rpath: bytes, entrynum: int) -> None:
         rp_off = self.get_entry_offset(entrynum)
         if rp_off is None:
             if self.verbose:
-                print('File does not have rpath. It should be a fully static executable.')
+                print('File "%s" does not have rpath. It should be a fully static executable.' % fname)
```
xclaesse, line 283: We prefer using f-string

```suggestion
                print(f'File {fname!r} does not have rpath. It should be a fully static executable.')
```

mesonbuild/scripts/depfixer.py
```diff
@@ -391,12 +342,9 @@ def remove_rpath_entry(self, entrynum: int) -> None:
         return None
 
 def fix_elf(fname: str, rpath_dirs_to_remove: T.Set[bytes], new_rpath: T.Optional[bytes], verbose: bool = True) -> None:
-    with Elf(fname, verbose) as e:
-        if new_rpath is None:
-            e.print_rpath()
-            e.print_runpath()
```
xclaesse, line 397: I guess those print have been added for debugging, that could still be useful to have if we find bugs in that script. Maybe just comment out those instead?

mesonbuild/scripts/depfixer.py
```diff
@@ -391,12 +385,10 @@ def remove_rpath_entry(self, entrynum: int) -> None:
...
-        else:
-            e.fix_rpath(rpath_dirs_to_remove, new_rpath)
+    if new_rpath is not None:
+        with Elf(fname, verbose) as e:
+            # note: e.get_rpath() and e.get_runpath() may be useful
```
eli-schwartz, line 390: So for debug purposes you'd edit the script and use these functions? Maybe it should be printed `if verbose`?

keszybz: Just insert `breakpoint()` in the appropriate place and explore from there… There are 5 functions, 3 were unused, 2 were only used here. As mentioned in the commit message, I don't think it's worth to print anything here, even with `verbose`, because it's not particularly interesting when it works. And when it *doesn't* work for some reason, just the rpath/runpath are not going to be enough to diagnose the issue.

keszybz:
Rebased.

jpakkane:
Test failure is not relevant, it happens with other MRs too.

PR #9397, opened 2021-10-13
Fix cases where options are looked by string instead of OptionKey

xclaesse:
This fix 2 places where strings are used to lookup a dictionary where keys are OptionKey.

### DIFF
--- a/mesonbuild/compilers/cuda.py
+++ b/mesonbuild/compilers/cuda.py
@@ -627,12 +627,9 @@ class CudaCompiler(Compiler):
 
         # We must strip the -std option from the host compiler option set, as NVCC has
         # its own -std flag that may not agree with the host compiler's.
-        overrides = {name: opt.value for name, opt in options.items()}
-        overrides.pop(OptionKey('std', machine=self.for_machine,
-                                       lang=self.host_compiler.language), None)
-        host_options = self.host_compiler.get_options().copy()
-        if 'std' in host_options:
-            del host_options['std'] # type: ignore
+        host_options = {key: options.get(key, opt) for key, opt in self.host_compiler.get_options().items()}
+        std_key = OptionKey('std', machine=self.for_machine, lang=self.host_compiler.language)
+        overrides = {std_key: 'none'}
         return OptionOverrideProxy(overrides, host_options)
 
     def get_option_compile_args(self, options: 'KeyedOptionDictType') -> T.List[str]:


mesonbuild/optinterpreter.py
```diff
@@ -48,7 +48,7 @@
 reserved_prefixes = ['cross_']
 
 def is_invalid_name(name: str, *, log: bool = True) -> bool:
-    if name in forbidden_option_names:
+    if mesonlib.OptionKey(name) in forbidden_option_names:
         return True
     pref = name.split('_')[0] + '_'
     if pref in forbidden_prefixes:
```
dcbaker, line 54: You need to use `from_string` here, not just `OptionKey`. But actually, this code is all pretty silly given the OptionKey helpers, it could be simplified to:
```python
key = mesonlib.OptionKey.from_string(name)
if not key.is_project():
    return True
elif key.name.startswith('cross_') and log:
    mlog.deprecation(...)
return False
```

dcbaker (changes requested):
Mypy really should be catching this, it's interesting that it isn't, since we have the non-overlapping type check on...

Anyway, this function really should have been refactored when we added the OptionKey class anyway, since that class has a method that does everything this function does.

mesonbuild/optinterpreter.py
dcbaker: As we discussed on IRC, this whole function is actually useless as the deprecation is invalid anyway. Instead the caller could be replaced with:
```python
key = mesonlib.OptionKey.from_string(name).evolve(subproject=self.subproject)
if not key.is_project():
    raise OptionError(...)
```

xclaesse: Indeed! And `cross_` does not need to be reserved any more, we went with `build.` prefix a long time ago.

xclaesse:
@dcbaker I'm not sure what `_to_host_compiler_options()` is supposed to do, judging from the CI failure that should never have worked...

xclaesse:
@jpakkane Can you check the commit about cuda here, git blame says it's you wrote that code in https://github.com/mesonbuild/meson/pull/8596 and I'm not sure what was the intention.

obilaniu:
@xclaesse I was indeed the author of the related PR. I was trying as much as possible to convey the idea of "Forward CUDA options to C++ host compiler" (while overriding where needed) but I definitely was not, and still am not, familiar with the nuts and bolts of Options.

Another thing I don't think is handled too well is the distinction between `native: true` and `native: false` (and thus the `build.<x>` and `<x>` options), as I don't even know if a given CUDA install even allows for toggling. What's in there now seems to work but maybe only holding up by a thread. Sorry...

PR #9400, opened 2021-10-14
Allow custom targets in extract_objects and test

### LINKED ISSUE #7585: test doesn't take custom_target index objects
This does not work

    qapi_doc = custom_target(...)
    test('name, exec, args: [ qapi_doc[0] ])

but this works:

    qapi_doc = custom_target(...)
    test('name, exec, args: [ qapi_doc ])

The error is

    ERROR: Command line arguments must be strings, files or targets.

but target index objects are close enough to targets that they should be accepted.

### LINKED ISSUE #8333: Can't extract objects form custom_target()
While trying to add a custom_target() to a plugin in qemu, I reached the following error:

ERROR: Object extraction arguments must be strings or Files.

Here is a small example:

```
project('object extraction', 'c')

ct = custom_target('custom target', output: ['test.c'], command: ['echo', ''], capture: true)
lib1 = shared_library('somelib', ct)
obj1 = lib1.extract_objects(ct)
```

help welcome!

bonzini:
This fixes two pending issues related to custom targets, allowing them to be used in extract_objects and in test. 

Fixes #7585
Fixes #8333

### DIFF
--- a/mesonbuild/interpreter/interpreter.py
+++ b/mesonbuild/interpreter/interpreter.py
@@ -199,7 +199,7 @@ known_build_target_kwargs = (
 )
 
 TEST_KWARGS: T.List[KwargInfo] = [
-    KwargInfo('args', ContainerTypeInfo(list, (str, mesonlib.File, build.BuildTarget, build.CustomTarget)),
+    KwargInfo('args', ContainerTypeInfo(list, (str, mesonlib.File, build.BuildTarget, build.CustomTarget, build.CustomTargetIndex)),
               listify=True, default=[]),
     KwargInfo('should_fail', bool, default=False),
     KwargInfo('timeout', int, default=30),
--- a/docs/markdown/snippets/extract-more-objects.md
+++ b/docs/markdown/snippets/extract-more-objects.md
@@ -1,0 +1,4 @@
+## ``extract_objects()`` supports generated sources
+
+Custom targets or generated files (returned by ``generator.process()``)
+can now be passed to ``extract_objects()``.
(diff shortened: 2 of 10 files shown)


mesonbuild/interpreter/interpreter.py
```diff
@@ -192,7 +193,7 @@ def dump_value(self, arr, list_sep, indent):
 )
 
 TEST_KWARGS: T.List[KwargInfo] = [
-    KwargInfo('args', ContainerTypeInfo(list, (str, mesonlib.File, build.Target)),
+    KwargInfo('args', ContainerTypeInfo(list, (str, mesonlib.File, build.Target, build.CustomTargetIndex)),
```
dcbaker, line 196: I don't think it's your bug but mine, but this should *not* allow `Target`, ti should be be `BuildTarget | CustomTarget`, otherwise a `RunTarget` would be allowed, which we don't want.

dcbaker (approved):
Your code all looks good. I did notice a separate bug looking at your code. Would you like me to patch that separately, or would you like to patch it?

mesonbuild/interpreter/interpreter.py
bonzini: Good catch; I think it should be fixed on top, possibly by creating a superclass of BuildTarget/CustomTarget. In fact, Target has methods such as `get_outputs()` that make very little sense for a RunTarget, and probably most of the type annotations that say "Target" actually mean `T.Union[BuildTarget, CustomTarget]`.

bonzini: Ok, I'll take this&mdash;there's possibly another instance for `add_install_script`, I have to write tests and then I'll send a separate pull request.

bonzini:
Ping @jpakkane, the failures seem extraneous.

bonzini:
Conflicts fixed and CI is green, hint hint :)

PR #9401, opened 2021-10-14
linkers: detect LLD when built with PACKAGE_VENDOR

bonzini:
This should make CI green again. Homebrew's "lld -v" output now starts with "Homebrew LLD", and Meson no longer detects it,

### DIFF
--- a/mesonbuild/linkers/detect.py
+++ b/mesonbuild/linkers/detect.py
@@ -79,7 +79,7 @@ def guess_win_linker(env: 'Environment', compiler: T.List[str], comp_class: T.Ty
         check_args.extend(extra_args)
 
     p, o, _ = Popen_safe(compiler + check_args)
-    if o.startswith('LLD'):
+    if 'LLD' in o.split('\n')[0]:
         if '(compatible with GNU linkers)' in o:
             return LLVMDynamicLinker(
                 compiler, for_machine, comp_class.LINKER_PREFIX,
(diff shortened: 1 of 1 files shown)


mesonbuild/linkers/detect.py
```diff
@@ -151,7 +151,7 @@ def guess_nix_linker(env: 'Environment', compiler: T.List[str], comp_class: T.Ty
     _, o, e = Popen_safe(compiler + check_args)
     v = search_version(o + e)
     linker: DynamicLinker
-    if o.startswith('LLD'):
+    if re.search(r'^(\S+\s+)*LLD', o):
```
eli-schwartz, line 154: This seems equivalent to `if 'LLD' in o`, which is how we handle GNU below.

bonzini: The idea was to only match the first line but yeah I am not sure it will work. Should have been `.*\bLLD` since period does not match newline.

I won't be able to adjust the PR over the weekend though, sorry.

dcbaker: If you really want to just match the first line, I still suspect `'LLD' in o.split()[0]` will still be faster than a regex

mesonbuild/linkers/detect.py
```diff
@@ -151,7 +151,7 @@ def guess_nix_linker(env: 'Environment', compiler: T.List[str], comp_class: T.Ty
     _, o, e = Popen_safe(compiler + check_args)
     v = search_version(o + e)
     linker: DynamicLinker
-    if o.startswith('LLD'):
+    if re.search(r'^.*\bLLD\b', o):
```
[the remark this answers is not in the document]
bonzini: `partition` also takes time and memory proportional to the size of the string. Sure `--help` is not going to be 1G large, but it seems just wrong if we know the first line is the only one that is involved; same for `'LLD' in o`.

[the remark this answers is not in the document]
eli-schwartz: Maybe it seems wrong, but I don't think it actually is. A really smart search algorithm will figure out what it's searching and then select the methodology that is best suited to the input. In this case it might either check the string size and switch to regex after a certain cutoff point, or simply assume that program --help text can't be megabytes in size without it being broken in unusually bizarre ways.

PR #9402, opened 2021-10-15
wrap: move FeatureNew checks to a more natural place

### LINKED ISSUE #9118: subprojects which depend on other subprojects trigger warnings for patch_directory
From the wrapdb, try to build the openh264 project.

```
$ meson setup --wipe builddir -Dwraps=openh264 --fatal-meson-warnings
The Meson build system
Version: 0.59.0
Source dir: /home/eschwartz/git/mesonbuild/wrapdb
Build dir: /home/eschwartz/git/mesonbuild/wrapdb/builddir
Build type: native build
Project name: wrapdb
Project version: undefined
Host machine cpu: x86_64

Executing subproject openh264 

openh264| Project name: openh264
openh264| Project version: 2.1.0
openh264| C compiler for the host machine: ccache cc (gcc 11.1.0 "cc (GCC) 11.1.0")
openh264| C linker for the host machine: cc ld.bfd 2.36.1
openh264| C++ compiler for the host machine: ccache c++ (gcc 11.1.0 "c++ (GCC) 11.1.0")
openh264| C++ linker for the host machine: c++ ld.bfd 2.36.1
openh264| Compiler for C++ supports arguments -Wno-non-virtual-dtor: YES
openh264| Compiler for C++ supports arguments -Wno-class-memaccess: YES
openh264| Compiler for C++ supports arguments -Wno-strict-aliasing: YES
openh264| Run-time dependency threads found: YES
openh264| Library pthread found: YES
openh264| Library libm found: YES
openh264| Program nasm found: YES (/usr/bin/nasm)
openh264| Found pkg-config: /usr/bin/pkg-config (1.7.3)
openh264| Run-time dependency GTest found: NO (tried pkgconfig and system)
openh264| Looking for a fallback subproject for the dependency gtest
openh264| Using gtest source from cache.
openh264| WARNING: Project targeting '>= 0.50' but tried to use feature introduced in '0.55.0': patch_directory.

subprojects/openh264-2.1.1/test/meson.build:4:2: ERROR: Fatal warnings enabled, aborting

A full log can be found at /home/eschwartz/git/mesonbuild/wrapdb/builddir/meson-logs/meson-log.txt
```

Several other warnings exist but don't have a chance to be reached, however those are fixed in upstream openh264 git... but there is nothing they can do about the first warning. The openh264 project does include their own gtest.wrap which scrupulously uses patch_url, but obviously this does not undergo wrap promotion if the superproject has an existing wrap.

This unfortunately causes some awkwardness for the wrapdb, since the wrapdb tooling uses --fatal-meson-warnings to catch certain types of errors.

Perhaps warnings coming from .wrap files should always obey the superproject meson_version, regardless of which meson.build they got invoked by?
Perhaps the wrap files should be tracked down to their original locations and obey that project's meson_version?

eli-schwartz:
Now, warnings are unconditionally raised when parsing the wrap file, whether they are used or not. That being said, these warnings literally just check for a couple of keys used in the .wrap ini file.

Moving these checks from the time of use to the time of loading, means that we no longer report warnings only when originally downloading or extracting the file or VCS repo.

It also means we no longer report warnings in one subproject, when a wrap file is picked up from a different subproject because the first subproject actually does a dependency lookup. This caused issues for the WrapDB tooling, which uses patch_directory everywhere and the superproject requires a suitable minimum version of meson for this... but individual wraps might use a much lower version, and would then raise a warning (in strict mode, converted to an error) when it resolved a dependency from another WrapDB project.

Fixes #9118


mesonbuild/wrap/wrap.py
```diff
@@ -554,7 +563,6 @@ def get_file_internal(self, what: str) -> str:
             self.download(what, cache_path)
             return cache_path
         else:
-            from ..interpreterbase import FeatureNew
             FeatureNew(f'Local wrap patch files without {what}_url', '0.55.0').use(self.current_subproject)
```
xclaesse, line 566: You forgot to remove that line. You can also remove self.current_subproject that should be unused now.

Issue #8114, opened 2020-12-20
Incorrect name of optional dependency if it was not found


radioactiveman:
**Describe the bug**
If an optional dependency is not found, the method name() returns "not-found".
But the [documentation](
https://mesonbuild.com/Reference-manual.html#dependency-object) states something different.
> returns the name of the dependency that was searched.

This is obviously wrong, I was not searching for a dependency named "not-found".
See the example below.

**To Reproduce**
```meson
foo_dep = dependency('foo', required: false)

if not foo_dep.found()
  message('Plugin foo disabled due to missing dependency: ' + foo_dep.name())
endif
```

**Actual behavior**
> Plugin foo disabled due to missing dependency: not-found

**Expected behavior**
> Plugin foo disabled due to missing dependency: foo

**System parameters**
* Meson 0.56.0

eli-schwartz:
This was fixed in the above linked PR, but the cross reference didn't contain the correct closing keywords.

PR #9404, opened 2021-10-15
dep.name(): return dependency name even if dependency is not found

heijligen:
The dep.name() function schould always return the name of the
dependency as documented. No matter if it was found or not.
https://mesonbuild.com/Reference-manual_returned_dep.html#depfound

Currently it returns 'not-found'

https://github.com/mesonbuild/meson/issues/8114

### DIFF
--- a/mesonbuild/interpreter/dependencyfallbacks.py
+++ b/mesonbuild/interpreter/dependencyfallbacks.py
@@ -270,7 +270,7 @@ class DependencyFallbacksHolder(MesonInterpreterObject):
             FeatureNew.single_use('OpenMP Dependency', '0.46.0', self.subproject)
 
     def _notfound_dependency(self) -> NotFoundDependency:
-        return NotFoundDependency(self.environment)
+        return NotFoundDependency(self.names[0] if self.names else '', self.environment)
 
     @staticmethod
     def _check_version(wanted: T.Optional[str], found: str) -> bool:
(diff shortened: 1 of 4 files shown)


mesonbuild/modules/python.py
```diff
@@ -497,7 +497,7 @@ def dependency_method(self, args: T.List['TYPE_var'], kwargs: 'TYPE_kwargs') ->
 
         # it's theoretically (though not practically) possible for the else clse
         # to not bind dep, let's ensure it is.
-        dep: 'Dependency' = NotFoundDependency(self.interpreter.environment)
+        dep: 'Dependency' = NotFoundDependency('python', self.interpreter.environment)
```
xclaesse, line 500: We could be more precise here, in python_factory() it builds the full pkg-config name:
```
        pkg_embed = '-embed' if embed and mesonlib.version_compare(installation.version, '>=3.8') else ''
        pkg_name = f'python-{pkg_version}{pkg_embed}'
```

mesonbuild/interpreter/dependencyfallbacks.py
```diff
@@ -270,7 +270,7 @@ def _handle_featurenew_dependencies(self, name: str) -> None:
             FeatureNew.single_use('OpenMP Dependency', '0.46.0', self.subproject)
 
     def _notfound_dependency(self) -> NotFoundDependency:
-        return NotFoundDependency(self.environment)
+        return NotFoundDependency(' '.join(self.names), self.environment)
```
xclaesse, line 273: Just take `self.names[0]`, that list is fallbacks when a dependency could have multiple names.

heijligen: Done

mesonbuild/modules/python.py
heijligen:
```
pymod = import('python')
py_inst = pymod.find_installation('python3.9')
py_dep = py_inst.dependency()
message(py_dep.name())
```
This will print ```python```. So the name should be the same if the dependency was not found

I reverted this to the first version.

PR #8312, opened 2021-02-07
scripts: Make sure we load as little python modules as possible

xclaesse:
Fixes: #8310.

PR #9407, opened 2021-10-17
Improve speed of meson wrapped command execution (for capture etc).

anarazel:
This improves meson wrapped command execution ~215ms to ~66ms on average on my machine. There's more changes to come, but I thought this is a good enough piece to get some feedback.

Any thoughts on mesonbuild/serialization.py as the location for {Executable,Test}Serialisation?

I'm not sure that the deferred import in mesonmain.py is the best approach. Might be better to just not build a global argparse with all the sub-parsers at all, and just import the relevant subcommand when the relevant top-level command is specified?

### DIFF
--- a/mesonbuild/serialisation.py
+++ b/mesonbuild/serialisation.py
@@ -1,0 +1,79 @@
+# Copyright 2013-2021 The Meson development team
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import typing as T
+from . import programs
+
+if T.TYPE_CHECKING:
+    from . import build
+    from .backend.backends import TestProtocol
+
+class ExecutableSerialisation:
+
+    # XXX: should capture and feed default to False, instead of None?
+
+    def __init__(self, cmd_args: T.List[str],
+                 env: T.Optional['build.EnvironmentVariables'] = None,
+                 exe_wrapper: T.Optional[programs.ExternalProgram] = None,
+                 workdir: T.Optional[str] = None,
+                 extra_paths: T.Optional[T.List] = None,
+                 capture: T.Optional[bool] = None,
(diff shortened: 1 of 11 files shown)


mesonbuild/serialisation.py
```diff
@@ -0,0 +1,84 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import typing as T
+
+if T.TYPE_CHECKING:
+    from . import build
+    from . import programs
+    from .backend.backends import TestProtocol
```
anarazel, line 20: This runs into danger of being a circular import, not great... Isn't an issue for type checking but...

mesonbuild/serialisation.py
```diff
@@ -0,0 +1,84 @@
...
+                 feed: T.Optional[bool] = None,
+                 tag: T.Optional[str] = None,
+                 verbose: bool = False,
+                 ) -> None:
+        self.cmd_args = cmd_args
+        self.env = env
+        if exe_wrapper is not None:
+            # XXX: import is a bit slow - but perhaps mypy type checking obsoleted this
+            # anyway?
+            from . import programs
```
anarazel, line 41: ^

xclaesse:
I prefer your way of delaying the imports instead of special moving special casing of --internal. But the most important is to verify what are the actual imports done, please copy the unit test I wrote in https://github.com/mesonbuild/meson/pull/8312 so we can compare which approach loads the less modules.

mesonbuild/serialisation.py
anarazel: The next line does assert isinstance(exe_wrapper, programs.ExternalProgram), which is the only reason ExternalProgram is needed at that point. I was wondering if that assert isn't really needed anymore, due to the mypy type checking.

eli-schwartz: Yes, I believe mypy renders this unneeded.

xclaesse: `./run_mypy.py` pass locally without that assert, so I think you can drop it.

eli-schwartz: Asserts were added in 416a00308f5b0f228af3c93eb597eca8529fdbb0, and the file got properly typed in f6f35aa906c8ebac21af07c17449df65d08aeaeb.

xclaesse:
@anarazel I made another attempt there: https://github.com/mesonbuild/meson/pull/10809. While I like your idea of deferring imports in CommandLineParser, it still imports the whole mesonlib and many other modules that can be avoided. Defferring everything gets complicated, so I think it's better to move the CLI part into its own module.

PR #9408, opened 2021-10-18
Fix linker detection when linker path is absolute

mon:
When using an absolute compiler path, linker detection incorrectly compares the full path with any linker detection strings, returning false for all of them. In the case of `ar2000`, this causes linker detection to fail as it is not given the correct version arg.

This PR uses the basename for these checks, similar to how `_detect_c_or_cpp_compiler` already works.

This also allows some simplification of the detection logic by stripping exe suffixes from some checks.

### DIFF
--- a/mesonbuild/compilers/detect.py
+++ b/mesonbuild/compilers/detect.py
@@ -292,9 +292,11 @@ def detect_static_linker(env: 'Environment', compiler: Compiler) -> StaticLinker
             linkers = default_linkers
     popen_exceptions = {}
     for linker in linkers:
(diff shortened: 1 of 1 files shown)


mon:
I'm really not sure how I broke the linker detection for dynamic linkers by modifying static linkers - as far as I can see the logic shouldn't really be any different. Anything obvious I've missed?

eli-schwartz:
Homebrew LLD was not broken by you, it fails for every PR right now because they changed the format of the output. :)

See #9401

jpakkane:
Please rebase and squash to a single commit. If CI gives green (it should) then this can be merged.

mon:
I've let this PR lapse because it's superseded/included in #9419 - if you'd prefer the commits be split, I'll rebase.

jpakkane:
If everything needed is in that one, then just close this MR.

PR #9410, opened 2021-10-18
mtest: limit "magic" CTRL+C behavior to process group leaders

bonzini:
If meson is not a process group leader, a SIGINT will be delivered also to its parent process (and possibly other processes).  The parent process then will probably exit and mtest will continue running in the background, without any way to interrupt the run completely.

To fix this, treat SIGINT and SIGTERM the same way unless mtest is a process group leader.

This can be reproduced when a parallel make invokes "meson test".

### DIFF
--- a/mesonbuild/mtest.py
+++ b/mesonbuild/mtest.py
@@ -1893,7 +1893,10 @@ class TestHarness:
             l.start(self)
 
         if sys.platform != 'win32':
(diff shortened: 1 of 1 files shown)


xclaesse:
Seems to make sense. OOC, when is meson not the group leader? You wrote a scrip that does that specifically?

bonzini:
(In fact you can see from the coverage report that Meson is *never* a group leader during a CI run).

xclaesse:
Ok, LGTM.

Issue #9412, opened 2021-10-19
Visual Studio environment settings are failing, because named temporary file can't be opened twice by name on Windows


gaborkertesz-linaro:
**Describe the bug**
Visual Studio environment setting is failing by exception as temporary bat file is created, but can't be opened later in subprocess.
https://github.com/mesonbuild/meson/blob/master/mesonbuild/mesonlib/vsenv.py#L76
https://github.com/mesonbuild/meson/blob/master/mesonbuild/mesonlib/vsenv.py#L79

https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile
> Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows NT or later).

Moreover "str(bat_file)" doesn't return the string of the temporary file, but "bat_file.name" makes the job.

**To Reproduce**
I use Meson for FreeType build:
https://gitlab.freedesktop.org/freetype/freetype/-/blob/master/meson.build

> meson setup out --buildtype release -Dtests=enabled --wipe

Any setup would want to activate VS on Windows will fail.

**Expected behavior**
Visual Studio environment should be activated properly by the created .bat temporary file.
0.59 version is working as expected.

**system parameters**
* native x64 and native win-arm64
* Windows 10
* Python 3.10
* `meson --version`: 0.60.0.rc1
* `ninja --version`: 1.10.2

PR #9413, opened 2021-10-19
Fix NamedTemporaryFile file reopen issue on Win #9412

gaborkertesz-linaro:
NamedTemporaryFile can't be opened by name on Windows.
For Windows the created temporary bat file is now closed before
passing to a subprocess, prevented from removal automatically upon
close and deleted explicitly upon finish.


xclaesse:
You're right, why did it work when I tested that before... Anyway, LGTM before 0.60.0 release.

PR #9419, opened 2021-10-21
Generalise C2000 to "Texas Instruments compiler", add MSP430 compiler support

mon:
The existing C2000 compiler in meson is actually just one of many flavours of TI's compilers for microprocessors. It can be trivially generalised because the compilers all share the same argument format.

To verify this, I checked:
- [C2000](https://www.ti.com/lit/ug/spru514r/spru514r.pdf)
- [C2000 from 2015](https://downloads.ti.com/docs/esd/SPRU514J/index.html#SPRU514J_HTML/using_the_c_c_compiler.html#2.3.2_Frequently_Used_Options%3FTocPath%3D2%2520Using%2520the%2520C%252FC%252B%252B%2520Compiler%7C2.3%2520Changing%2520the%2520Compiler's%2520Behavior%2520with%2520Options%7C_____2)
- [MSP430](https://www.ti.com/lit/ug/slau132r/slau132r.pdf)
- [ARM (CC3200)](https://www.ti.com/lit/ug/spnu151r/spnu151r.pdf)

I didn't do a full generalisation (CMake does) because that would require running `cl.exe ?`  to correctly detect the compiler regardless of name. For now, it's just easier to add the binary names in detect.py.

I spent a lot of time wondering about some of the flags which didn't seem to match up at all, but it seems the original PR that added the C2000 mistakenly copied them over from Renesas' compiler definition.

The linker uses `--include-path` but the compiler uses `--include`. Both have the short form `-I` so no transform is required from the unix format.
For defines, however, the compiler has `-D` but the linker only supports `--define=`, so this transform remains.

This PR builds off #9408 so there's a handful of extra lines in the diff.


docs/markdown/Reference-tables.md
```diff
@@ -54,7 +54,7 @@ These are return values of the `get_linker_id` method in a compiler object.
 | optlink    | optlink (used with DMD)                     |
 | rlink      | The Renesas linker, used with CCrx only     |
 | xc16-ar    | The Microchip linker, used with XC16 only   |
-| ar2000     | The Texas Instruments linker, used with C2000 only |
+| ti-ar      | The Texas Instruments linker, used with C2000/MSP430 only |
```
jpakkane, line 57: This is a problematic change because it breaks existing projects that use `ar2000` (and `c2000`) strings in their build files.

dcbaker: Indeed, we likely need to keep the ar2000 name around for some time, it's probably going to have to be a case of:
```python
class TiCompiler:
   id = 'it-ar'

   < rest of implementation >

class ArCompiler(TiCompiler):
    is = 'ar200'
```

How to pass a deprecation message for ar2000 name is a more interesting question...

dcbaker:
I haven't read the rest of this super closely, but other than the naming thing this seems like a fine change.

docs/markdown/Reference-tables.md
mon: Ah, I forgot about the ID backwards compat issue. I think for now I'll just modify the detection methods to assign that suggested wrapper class to the c2000 toolchain.

This would stop new users to the c2000 toolchain from being able to use 'ti' as their compiler ID, but it's not a huge userbase so I think it's ok.

One solution I can think of is allowing a compiler to have 2 IDs, but I think that generalisation is not worth the effort considering how niche this compatibility issue is.

mon:
Resolved conflicts and added the detection for the ARM compiler as well - needed a little reshuffle so the `'ARM' in output` check didn't mis-identify it.

I also am not very comfortable with how generically TI has named their compiler ("armcl") so I've reverted to exact-match checking for the compiler name with/without exe suffix.

Let me know if there's anything else!

jpakkane:
This branch has merge commits. They are not permitted [as per the documentation](https://mesonbuild.com/Contributing.html#keeping-pull-requests-up-to-date) and must be removed.

mon:
Ah, yep. Fixed, and added the release snippet that I also didn't do. The failing tests appear to be unrelated.

mon:
Rebased on master again because of some merge conflicts.

Anything else I need to do? Thanks for your time reviewing so far.

mon:
Similar to my other PR, should I squash all the commits to make the history cleaner?

PR #9429, opened 2021-10-22
Fix to find Python files for Windows virtualenvs

matthew-brett:
Virtualenvs do not have their Python DLLs etc in the `sys.prefix`
directory, but in the `sys.base_prefix` directory.  This directory is
the same as `sys.prefix` if not in a virtualenv, so is safe for either
case:

https://docs.python.org/3/library/sys.html#sys.base_prefix

### DIFF
--- a/mesonbuild/modules/python.py
+++ b/mesonbuild/modules/python.py
@@ -161,7 +161,8 @@ class PythonSystemDependency(SystemDependency, _PythonDependencyBase):
                     libpath = Path(f'python{vernum}.dll')
                 else:
                     libpath = Path('libs') / f'python{vernum}.lib'
-            lib = Path(self.variables.get('base')) / libpath
+            # base_prefix to allow for virtualenvs.
+            lib = Path(self.variables.get('base_prefix')) / libpath
         elif self.platform == 'mingw':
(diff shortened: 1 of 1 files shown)


jpakkane:
Since no-one has objected I'm going to merge this.

matthew-brett:
Great - thanks!

mesonbuild/modules/python.py
```diff
@@ -318,8 +319,11 @@ def links_against_libpython():
     cmd.ensure_finalized()
     return bool(cmd.get_libraries(Extension('dummy', [])))
 
+variables = sysconfig.get_config_vars()
+variables.update({'base_prefix': getattr(sys, 'base_prefix', sys.prefix)})
```
FFY00, line 323: This is incorrect, it assumes the target Python is the same as the Python Meson is running under. Why is it needed?

eli-schwartz: I don't think it does? This is all happening inside the INTROSPECT_COMMAND which is run by the target Python.

FFY00: Ah, yes, sorry, my bad.

PR #9453, opened 2021-10-26
Restore the possibility of forcing "fat" static library links on a per-target basis

obilaniu:
With 0.60.0, non-installed `static_library()` targets are built by default as thin archives where supported, such as with `ar csrT`.  This broke a flow where I was statically linking code into an archive, then passing the archive to a `custom_target()` that was expecting a "fat" archive and wasn't handling "thin" archives _(specifically, it's `nvcc` being driven with custom flags)_.

Thin archives are handy, but "fattening" an archive where needed is quite involved and annoying.

Add a `thin:` kwarg to `static_library()` and default it to `true`, continuing the new behaviour of using thin archives in 0.60.0+. Passing the non-default `thin: false` restores the known-good behaviour of Meson prior to 0.60.0.

### DIFF
--- a/docs/yaml/functions/static_library.yaml
+++ b/docs/yaml/functions/static_library.yaml
@@ -23,3 +23,15 @@ kwargs:
       If `true` the object files in the target will be prelinked,
       meaning that it will contain only one prelinked
       object file rather than the individual object files.
+
+  thin:
+    type: bool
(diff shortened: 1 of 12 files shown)


bonzini:
This seems more like a task for a built-in option, than a keyword argument.

Issue #9414, opened 2021-10-19
After reconfiguring old object files are still linked in


lb90:
**Describe the bug**

After removing a source file in the Meson build definition, Meson reconfigures and then that source file is not compiled, but the corresponding object file (available from previous compilations) is still passed as input to the linker.

**To Reproduce**
I have created a very simple branch from the master branch that adds a source file to [GdkWin32](https://gitlab.gnome.org/lb90/gtk/-/commits/new-feature). That source file uses a function from `cabinet.dll` so it adds a new dependency on a system DLL.

So I can run Meson and build that branch just fine. But then if I checkout the master branch and try a rebuild I get:

```
Build targets in project: 114
gtk 4.5.0
Option buildtype is: debug [default: debugoptimized]
Found ninja-1.10.2 at D:\msys64\mingw64\bin/ninja.EXE
[1/41] Generating demo-header with a custom command (wrapped by meson to capture output)
[2/36] Compiling C object gdk/win32/libgdk-win32.a.p/gdkdisplay-win32.c.obj
[3/36] Linking static target gdk/win32/libgdk-win32.a
[4/36] Linking static target gdk/libgdk.a
[5/36] Linking target tools/gtk4-update-icon-cache.exe
[6/36] Linking target gtk/libgtk-4-1.dll
FAILED: gtk/libgtk-4-1.dll
[...]

PR #9468, opened 2021-10-27
Delete static library archives also on Windows

lb90:
Fixes https://github.com/mesonbuild/meson/issues/9414

### DIFF
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -74,12 +74,8 @@ def cmd_quote(arg: str) -> str:
 # (see https://ninja-build.org/manual.html#ref_rule_command)
 if mesonlib.is_windows():
     quote_func = cmd_quote
-    execute_wrapper = ['cmd', '/c']  # unused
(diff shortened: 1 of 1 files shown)


jpakkane:
I'm not a fan of complicated shell shenanigans like this. It would be better to have an `internal` command for deleting files that is then called from this. It would get rid of the `if windows` branch too.

eli-schwartz:
The if windows branch has the advantage of not requiring to go through mesonbuild.mesonmain and the python overhead, for the fairly common case that you're building on unix and don't need `rm.py` because you have `rm`.

Maybe we should just use pathlib though ;)

lb90:
So the consensus is that we should use a Meson --internal command to delete files on Windows? I can look into that i wanted :) Let me know

lb90:
ping :)

lb90:
Hi @xclaesse, I can work on that! how would I go about adding an internal command?

xclaesse:
You can add scripts/rm.py and use it had meson --internal rm file

lb90:
Ok, updated to use an internal script. Let me know if anything is missing!

We could use powershell in alternative to python, but I don't know if startup times are any better

PR #9475, opened 2021-10-27
Use typed_pos args in gnome module (and other typing cleanups)

dcbaker:
This series is the starting work of a larger gnome module modernization/cleanup I've been working on. The series is quite large, and the keyword argument conversion is much more involved, and likely to require a deeper review. I've shaved off what are (hopefully) smaller, easier to verify changes.

The first part of this series is groundwork, adding or fixing type annotations, using helpers instead of open coding, fixing polymorphism in union types, and that sort of thing.

Then i've added most of the "easy" type annotations to the gnome module, namely those that don't involve using typed_pos_args or typed_kwargs.

Finally, the last par of the series makes use of `typed_pos_args` or `noPosargs`, as appropriate. I'll follow this up with work on using typed_kwargs, some of which is done, some of which is in progress, but I felt this was already big enough of a series without that.

### DIFF
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -1003,7 +1003,7 @@ class BuildTarget(Target):
         return ExtractedObjects(self, self.sources, self.generated, self.objects,
                                 recursive)
 
-    def get_all_link_deps(self):
+    def get_all_link_deps(self) -> 'ImmutableListProtocol[Target]':
         return self.get_transitive_link_deps()
 
     @lru_cache(maxsize=None)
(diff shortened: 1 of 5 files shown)


jpakkane:
⚔️ conflicts.

dcbaker:
@jpakkane rebased

mensinda (approved):
LGTM. Not sure what is up with Cygwin though.

dcbaker:
I think I just need to rebase to fix cygwin

PR #9483, opened 2021-10-28
python: Restore logging whether or not the dependency has been found

xclaesse:


dcbaker:
I'm happy with patch 1 r-b on that, patch two not so much. We already have the type information in the form of the class, and i really don't like that python dependencies suddenly have different behavior than every other kind of dependency

xclaesse:
> We already have the type information in the form of the class

That's true, changed to have method string as class member instead of passing to contructor everywhere.

> i really don't like that python dependencies suddenly have different behavior than every other kind of dependency

I think it's useful information to know the method that found python dependency. This just restore the previous behavior that was lost, it's not new.

dcbaker:
But it's still specific to the python dependency, that should be handled for all dependencies in a single centralized place, not one off for just python.

xclaesse:
@dcbaker we should use factory infrastructure indeed, but that's a bigger refactoring that should not go into 0.60.1. We can always rethink this for 0.61.

dcbaker:
Then patch two shouidn't go into 0.60.1 either, since it's not a bug fix?

xclaesse:
@dcbaker it is a regression, it's not reported in the logs any more if it found python dep or not. Can we stop useless bikeshedding here please?

nirbheek:
I think Dylan's concern is valid, so it sounds to me like we should:

1. Split patch 2 out of this PR
2. Merge this PR immediately and release 0.60.1
3. Do a proper fix for logging what method found the python dependency, and merge it into master
4. Open a PR targeting the 0.60 branch with patch 2, and merge it into 0.60.2

Does this sound ok?

xclaesse:
That sounds like an awfully complicated plan for something that could have been merged already instead of discussing non-issues. Do as you wish, I don't care.

eli-schwartz:
Since everyone seems to agree that the first patch is good and should be merged ASAP and added to 0.60.1, I merged it manually as c8d125653d0535624741c591397a2b9b4ae3bb43.

We can now proceed to discuss whether or not adding more info back to the log message is important enough to merge a disputed change for urgent backporting into 0.60.1 -- without holding up the first, undisputed, patch.

nirbheek:
> Since everyone seems to agree that the first patch is good and should be merged ASAP and added to 0.60.1, I merged it manually as [c8d1256](https://github.com/mesonbuild/meson/commit/c8d125653d0535624741c591397a2b9b4ae3bb43).

I've cherry-picked this into 0.60 and moved the milestone on this PR to 0.60.2

test cases/unit/101 python without pkgconfig/meson.build
```diff
@@ -0,0 +1,4 @@
+project('python wihtout pkgconfig', 'c')
```
Conan-Kudo, line 1: `wihtout` -> `without`

Conan-Kudo (changes requested):

test cases/unit/101 python without pkgconfig/meson.build
nirbheek: This was already merged into master at the time of this comment.

nirbheek:
I've rebased this on top of master and updated the PR title to reflect the remaining issue. @xclaesse do you want to update this PR and get it in for 0.60.2?

nirbheek:
Removed from 0.60.2 milestone.

tristan957:
Anything holding this back guys?

xclaesse:
> Anything holding this back guys?

Rebased in case someone wants to approve it.

tristan957 (approved):

eli-schwartz:
A year and a half ago this was largely ad-hoc code for finding the dependency. After a lot of refactoring including my uplifting the code to mesonbuild.dependencies.* this actually behaves normally when used as `dependency('python3')`.

We should just use `dep.log_tried()`.

xclaesse:
Ah, no I see, the trick is passing `installation` param to `python_factory()`

xclaesse:
@eli-schwartz Updated to use `detect.find_external_dependency()`. A bit hackish but it makes consistent logs. I hesitated to instead add the factory into `packages` dict with a special key, but that felt worse.

eli-schwartz:
Aha, that turned out to be surprisingly easy. :)

The commit message could do with a bit of updating now, though.

xclaesse:
> The commit message could do with a bit of updating now, though.

updated

dcbaker (approved):
Ah, yeah. I see how that happened. I'm going to blame this on "dependencies used to be really complicated, now they're only kinda complicated" :)

dcbaker:
but you since @eli-schwartz has been looking at this, probably should have at least an ack from him too.

I do like that you're using `detect.find_external_dependency`. It seems you and Eli have finally made my dream of killing the remains of the old multi-dependency system come true :)

eli-schwartz:
Yes, this is excellent.

PR #1469, opened 2017-03-14
Support multiple install dirs for built/custom targets

nirbheek:
You can now pass a list of strings to the `install_dir:` kwarg to `build_target` and `custom_target`.

Custom Targets:
===============
Allows you to specify the installation directory for each corresponding output. For example:

```meson
custom_target('different-install-dirs',
  output : ['first.file', 'second.file'],
  ...
  install : true,
  install_dir : ['somedir', 'otherdir])
```

This would install `first.file` to `somedir` and `second.file` to `otherdir`.

If only one `install_dir` is provided, all outputs are installed there (same behaviour as before).

To only install some outputs, pass `false` for the outputs that you don't want installed. For example:

```meson
custom_target('only-install-second',
  output : ['first.file', 'second.file'],
  ...
  install : true,
  install_dir : [false, 'otherdir])
```

This would install `second.file` to `otherdir` and not install `first.file`.

Build Targets:
==============
With `build_target()` (which includes `executable()`, `library()`, etc), usually there is only one primary output. However some types of targets have multiple outputs.

For example, while generating Vala libraries, `valac` also generates a header and a `.vapi` file both of which often need to be installed. This allows you to specify installation directories for those too.

```meson
# This will only install the library (same as before)
shared_library('somevalalib', 'somesource.vala',
  ...
  install : true)

# This will install the library, the header, and the vapi into the
# respective directories
shared_library('somevalalib', 'somesource.vala',
  ...
  install : true,
  install_dir : ['libdir', 'incdir', 'vapidir'])

# This will install the library into the default libdir and
# everything else into the specified directories
shared_library('somevalalib', 'somesource.vala',
  ...
  install : true,
  install_dir : [true, 'incdir', 'vapidir'])

# This will NOT install the library, and will install everything
# else into the specified directories
shared_library('somevalalib', 'somesource.vala',
  ...
  install : true,
  install_dir : [false, 'incdir', 'vapidir'])
```

`true`/`false` can also be used for secondary outputs in the same way.

Includes tests for all these.

Closes https://github.com/mesonbuild/meson/issues/705
Closes https://github.com/mesonbuild/meson/issues/891
Closes https://github.com/mesonbuild/meson/issues/892
Closes https://github.com/mesonbuild/meson/issues/1178
Closes https://github.com/mesonbuild/meson/issues/1193

Issue #1014, opened 2016-11-08
Add generate_typelib in the GNOME module


arteymix:
This would cover Vala and any tool that produce custom GIR.

We should deprecate `generate_gir` and split it into two distinct tools. This will have a few advantages like not returning two targets or passing redundant arguments.

ptomato:
This would also be useful for projects that generate private typelibs and don't want to install the GIR because it's not public API.

You can kind of fake this with `install_dir_gir: '/dev/null', install_dir_typelib: typelibdir`, but that breaks `DESTDIR` so it's not a really workable solution.

ptomato:
It seems that `install_dir_gir: false` works for that purpose, no idea if that's intentional / liable to break or not.

arteymix:
I believe it's intended. We have the same behaviour for Vala `install_dir` with a special `true`-implies-default-location behaviour.

esodan:
But the issue here is to generate the typelib. It is generated by a different command and depends on the library compilation.

albfan:
Based on existing use case for vala:

```meson
foo_lib = shared_library('foo', 'foo.vala',
                  vala_gir: 'Foo-1.0.gir',
                  dependencies: [glib_dep, gobject_dep],
                  install: true,
                  install_dir: [true, true, true, true])

g_ir_compiler = find_program('g-ir-compiler')
typelib_input = meson.current_build_dir() / 'Foo-1.0.gir'
custom_target('foo typelib', command: [g_ir_compiler, '--output', '@OUTPUT@', typelib_input],
              output: 'Foo-1.0.typelib',
              depends: foo_lib,
              install: true,
              install_dir: get_option('libdir') / 'girepository-1.0')
```

shared_library is a general object, but it knows about `vala_gir`, as vala needs to create gir file with its own compiler

Can `gnome.generate_gir` know about a shared_library result is built with vala, so it uses its gir file and only creates the typelib part?

This back to standard how people use `gnome.generate_gir` and avoid developer to point to generated files in build directory.

If that's not possible, this function might be:

```meson
foo_lib = shared_library('foo', 'foo.vala',
                  vala_gir: 'Foo-1.0.gir',
                  dependencies: [glib_dep, gobject_dep],
                  install: true,
                  install_dir: [true, true, true, true]
)

gnome = import('gnome')

gnome.generate_typelib(foo_lib, 
  gir:  meson.current_build_dir() / 'Foo-1.0.gir',
[...]

PR #9484, opened 2021-10-28
modules/gnome: fix missing install_dir, again, harder

### LINKED ISSUE #9350: Throws exception instead of parsing error
meson git c6d74ac7e0890c323bd1190d5f5d3d938fc6d59a

When building this tree, meson throws an exception instead of complaining about the parsing error and where it occurred.

[grilo-wip-hadess-grlnet-disable-fix.zip](https://github.com/mesonbuild/meson/files/7278069/grilo-wip-hadess-grlnet-disable-fix.zip)

```sh
$ ~/Projects/jhbuild/meson/meson.py --prefix /home/hadess/Projects/gnome-install --libdir lib  --buildtype=debugoptimized /home/hadess/Downloads/grilo-wip-hadess-grlnet-disable-fix
The Meson build system
Version: 0.59.99
Source dir: /home/hadess/Downloads/grilo-wip-hadess-grlnet-disable-fix
Build dir: /tmp/bug-repro
Build type: native build
Project name: grilo
Project version: 0.3.14
C compiler for the host machine: ccache cc (gcc 11.2.1 "cc (GCC) 11.2.1 20210728 (Red Hat 11.2.1-1)")
C linker for the host machine: cc ld.bfd 2.37-10
Host machine cpu: x86_64
Found pkg-config: /usr/bin/pkg-config (1.8.0)
Run-time dependency gio-2.0 found: YES 2.70.0
Run-time dependency glib-2.0 found: YES 2.70.0
Run-time dependency gmodule-2.0 found: YES 2.70.0
Run-time dependency gobject-2.0 found: YES 2.70.0
Run-time dependency libxml-2.0 found: YES 2.9.12
Run-time dependency libsoup-2.4 found: YES 2.74.0
Run-time dependency totem-plparser found: YES 3.26.6
Program g-ir-scanner found: YES (/usr/bin/g-ir-scanner)
Program vapigen found: YES (/usr/bin/vapigen)
Run-time dependency gtk+-3.0 found: YES 3.24.30
Run-time dependency oauth found: YES 1.0.3
Run-time dependency gobject-introspection-1.0 found: YES 1.70.0
Run-time dependency vapigen found: YES 0.54.1
Found pkg-config: /usr/bin/pkg-config (1.8.0)
Program glib-genmarshal found: YES (/usr/bin/glib-genmarshal)
  File "/home/hadess/Projects/jhbuild/meson/mesonbuild/mesonmain.py", line 228, in run
  File "/home/hadess/Projects/jhbuild/meson/mesonbuild/msetup.py", line 290, in run
    app.generate()
  File "/home/hadess/Projects/jhbuild/meson/mesonbuild/msetup.py", line 181, in generate
    self._generate(env)
  File "/home/hadess/Projects/jhbuild/meson/mesonbuild/msetup.py", line 225, in _generate
    intr.run()
  File "/home/hadess/Projects/jhbuild/meson/mesonbuild/interpreter/interpreter.py", line 2456, in run
    super().run()
  File "/home/hadess/Projects/jhbuild/meson/mesonbuild/interpreterbase/interpreterbase.py", line 165, in run
  File "/home/hadess/Projects/jhbuild/meson/mesonbuild/interpreterbase/interpreterbase.py", line 190, in evaluate_codeblock
    raise e
  File "/home/hadess/Projects/jhbuild/meson/mesonbuild/interpreterbase/interpreterbase.py", line 183, in evaluate_codeblock
    self.evaluate_statement(cur)
  File "/home/hadess/Projects/jhbuild/meson/mesonbuild/interpreterbase/interpreterbase.py", line 196, in evaluate_statement
  File "/home/hadess/Projects/jhbuild/meson/mesonbuild/interpreterbase/interpreterbase.py", line 82, in wrapper
    res = f(self, node)
  File "/home/hadess/Projects/jhbuild/meson/mesonbuild/interpreterbase/interpreterbase.py", line 629, in function_call
    return func(node, func_args, kwargs)
  File "/home/hadess/Projects/jhbuild/meson/mesonbuild/interpreterbase/decorators.py", line 697, in wrapped
  File "/home/hadess/Projects/jhbuild/meson/mesonbuild/interpreterbase/decorators.py", line 114, in wrapped
  File "/home/hadess/Projects/jhbuild/meson/mesonbuild/interpreterbase/decorators.py", line 275, in wrapper
    return f(*nargs, **wrapped_kwargs)
  File "/home/hadess/Projects/jhbuild/meson/mesonbuild/interpreter/interpreter.py", line 1941, in func_subdir
    self.evaluate_codeblock(codeblock)
  File "/home/hadess/Projects/jhbuild/meson/mesonbuild/interpreterbase/interpreterbase.py", line 190, in evaluate_codeblock
    raise e
  File "/home/hadess/Projects/jhbuild/meson/mesonbuild/interpreterbase/interpreterbase.py", line 183, in evaluate_codeblock
    self.evaluate_statement(cur)
  File "/home/hadess/Projects/jhbuild/meson/mesonbuild/interpreterbase/interpreterbase.py", line 198, in evaluate_statement
    self.assignment(cur)
  File "/home/hadess/Projects/jhbuild/meson/mesonbuild/interpreterbase/interpreterbase.py", line 848, in assignment
    value = self.evaluate_statement(node.value)
  File "/home/hadess/Projects/jhbuild/meson/mesonbuild/interpreterbase/interpreterbase.py", line 200, in evaluate_statement
    return self.method_call(cur)
  File "/home/hadess/Projects/jhbuild/meson/mesonbuild/interpreterbase/interpreterbase.py", line 666, in method_call
    return self._holderify(obj.method_call(method_name, args, kwargs))
  File "/home/hadess/Projects/jhbuild/meson/mesonbuild/interpreter/interpreterobjects.py", line 751, in method_call
  File "/home/hadess/Projects/jhbuild/meson/mesonbuild/interpreterbase/decorators.py", line 114, in wrapped
  File "/home/hadess/Projects/jhbuild/meson/mesonbuild/modules/gnome.py", line 1669, in genmarshal
    header = build.CustomTarget(output + '_h', state.subdir, state.subproject, custom_kwargs)
  File "/home/hadess/Projects/jhbuild/meson/mesonbuild/build.py", line 2317, in __init__
    self.process_kwargs(kwargs, backend)
  File "/home/hadess/Projects/jhbuild/meson/mesonbuild/build.py", line 2426, in process_kwargs
    if isinstance(kwargs['install_dir'], list):
KeyError: 'install_dir'
```

dcbaker:
It turns out this could be missing in GResource*Target as well, due
mostly to the same problem, side effects of mutating a shared
dictionary; though it could also happen with a specific set of keywords
given and others omitted.

Fixes #9350, again

### DIFF
--- a/mesonbuild/modules/gnome.py
+++ b/mesonbuild/modules/gnome.py
@@ -1458,7 +1458,7 @@ class GnomeModule(ExtensionModule):
             c_kwargs = custom_kwargs.copy()
             # Never install the C file. Complain on bug tracker if you need it.
             c_kwargs['install'] = False
-            c_kwargs['install_dir'] = False
+            c_kwargs['install_dir'] = []
             if h_template is not None:
                 if 'depends' in custom_kwargs:
                     c_kwargs['depends'] += [h_target]


justinkb:
with these changes (and the backports to 0.60.x), gjs 1.70.0 build no longer works properly

it complains about install_dir needing to be specified here

```
meson.build:563:0: ERROR: "install_dir" must be specified when installing a target
```

```
gjs_private_gir = gnome.generate_gir(libgjs,
    includes: ['GObject-2.0', 'Gio-2.0'], sources: libgjs_private_sources,
    namespace: 'GjsPrivate', nsversion: '1.0', identifier_prefix: 'Gjs',
    symbol_prefix: 'gjs_', extra_args: '--warn-error', install: true,
    install_dir_gir: false, install_dir_typelib: pkglibdir / 'girepository-1.0')
```

eli-schwartz:
Since #9520,

```
meson.build:580:0: ERROR: gnome.generate_gir keyword argument 'install_dir_gir' was of type 'bool' but should have been one of: str, NoneType
```

This was never legal. What was the point of passing "false" there?

justinkb:
I have just finished doing some digging, it is an issue with gjs meson.build. (Present since the first version where the switch to meson was made). the .gir file GjsPrivate-1.0.gir is missing on every package on every distribution that's using meson to build gjs). I have made a patch, and will submit it to gjs gitlab

ptomato:
> This was never legal, I think. What was the point of passing "false" there?

I was told that passing `false` here was the intended way of installing the typelib without installing the GIR: https://github.com/mesonbuild/meson/issues/1014#issuecomment-521852908

Is there any other recommended way to install the typelib without installing the GIR? If that is no longer possible, then I think this is a regression in Meson's `gnome` module. GJS is not the only project that generates a typelib with private functions and therefore declines to install a GIR that other programs could consume.

ptomato:
[Someone figured out](https://gitlab.gnome.org/GNOME/gjs/-/merge_requests/694) that `install_dir_gir: [false]` still works for this purpose. Could this (or any other way of achieving it that you prefer) be documented and a regression test added for it?

eli-schwartz:
That won't work either, at least on meson-git master:

```
meson.build:580:0: ERROR: gnome.generate_gir keyword argument 'install_dir_gir' was of type 'list' but should have been one of: str, NoneType
```

Fiddling with the types in order to trick the parser isn't a robust solution. Anything that *should* work, should work without hiding it inside an array -- and anything that shouldn't work, can't be relied on to work in the future just because the internal implementation currently swallows an array of boolean better than a boolean.

However, I guess documented or not, it does seem that the code allowed it due to the use of kwargs.get() and ***actually it does seem like there is a good reason to allow this***. Specifically, the rationale of #1469 implies that it should be explicitly possible, even though that came *after* install_dir_gir's implicit behavior.

The real problem I guess is that this wasn't documented (even though that PR documented it for vala), so the fact that it "should" work wasn't obvious... and then this got fundamentally broken by commit be1d013453e3df3b83da0c91f5211c822d4da4d7 which adds typechecking to this function but assumed it only accepted strings as implied by the docs.

@dcbaker sorry, but this is broken a third time...

The good news is that every function that gets properly typechecked then has a pure code description of what it should ***and does*** accept, so once we get this non-broken it should remain that way.

eli-schwartz:
Also per the comment on the gjs MR:

> given that it looks like that 4 year old proposal to split generate_gir into two never gained traction.

I think the main problem is no one ever submitted a PR for it. :( It seems like a basically good idea.

dcbaker:
The fix should be trivial, I'll have a look at it tomorrow or Monday

eli-schwartz:
> That won't work either, at least on meson-git master:
> 
> ```
> meson.build:580:0: ERROR: gnome.generate_gir keyword argument 'install_dir_gir' was of type 'list' but should have been one of: str, NoneType
> ```
> 
> Fiddling with the types in order to trick the parser isn't a robust solution. Anything that _should_ work, should work without hiding it inside an array -- and anything that shouldn't work, can't be relied on to work in the future just because the internal implementation currently swallows an array of boolean better than a boolean.

Apparently several projects did in fact use that broken hack anyway. @jbeich

https://github.com/linuxmint/cinnamon/pull/10596
https://github.com/linuxmint/cjs/pull/100
https://gitlab.gnome.org/GNOME/gjs/-/merge_requests/705


No idea if there are more.

ptomato:
Yep, I was aware that we would have to revert that. I was just waiting to see what the resolution was going to be from this issue.

eli-schwartz:
The resolution is, fixed in 0.60.3 (released December 22) and deprecated in 0.61.

In 0.61.0, gjs will print the following informational message at the end:

```
NOTICE: Future-deprecated features used:
 * 0.55.0: {'gnome.generate_gir argument --warn-error'}
 * 0.56.0: {'meson.build_root', 'Dependency.get_pkgconfig_variable'}
 * 0.61.0: {'"gnome.generate_gir" keyword argument "install_dir_gir" value "False"'}
```

That last one wants you to switch:
```diff
-install_dir_gir: false
+install_gir: false
```

As soon as you bump the minimum version of meson to 0.61

PR #9494, opened 2021-10-29
WIP: Paralellize some compiler checks

xclaesse:


xclaesse:
GLib can trivially be changed to use cc.has_headers() and cc.has_functions().
Linux: 14s to 10s
Windows: 94s to 124s

I was hoping for more impressive results, but it's still something...

anarazel:
Hm, why is this using a threaded executor instead of starting multiple subprocesses with subprocess / multiprocessing / ...? I'd assume that the python side will only be a very small fraction of the execution time.

nirbheek:
You don't need an Executor at all. Popen is already asynchronous. You just need to keep the handles around and wait on all of them.

nirbheek:
We did benchmarking for all this back in 2018: https://github.com/mesonbuild/meson/issues/3635#issuecomment-393280805

Asynchronous Popen (without async / await) is the fastest of them all.

anarazel:
On 2021-11-05 08:32:20 -0700, Nirbheek Chauhan wrote:
> We did benchmarking for all this back in 2018: https://github.com/mesonbuild/meson/issues/3635#issuecomment-393280805
> 
> Asynchronous Popen (without async / await) is the fastest of them all.

It pretty much has to be, because all of the work that Popen does is also
required for other methods of execution. The cost of executing lots of
subprocesses is fork() and exec(). A popen has a bit of file descriptor
overhead over that nothing more. E.g. a threaded approach also needs to create
threads etc.


I wonder is if this could be reframed a bit more fundamentally than just doing
the same thing in parallel to get to larger speedups. Imagine translating
has_headers(["<first/include>", "<SECOND>", ...]) to something like

#if __has_include("<first/include>")
#define MESON_HAS_INCLUDE_FIRST_INCLUDE
#endif

#if __has_include("<SECOND>")
#define MESON_HAS_INCLUDE_SECOND
#endif

....

then meson could preprocess the file and know all of the includes after one
invocation of the compiler. Of course that would only work if the compiler has
__has_include(), but that's not hard to test for separately.

xclaesse:
@nirbheek I did that a long time ago: https://github.com/mesonbuild/meson/pull/6664 but it makes the code a lot more complicated. The executor solution has the advantage of being trivial to implement and still have most of the speedup you can hope from parallelization.

PR #9502, opened 2021-10-30
Update Installing.md#custom-install-script

rusty-snake:
- Use `$MESON_INSTALL_DESTDIR_PREFIX` as this is the preferred way (https://mesonbuild.com/Reference-manual_builtin_meson.html#mesonadd_install_script)
- Drop `DESTDIR` note as it is no longer correct (https://mesonbuild.com/Release-notes-for-0-57-0.html#specify-destdir-on-command-line)

### DIFF
--- a/docs/markdown/Installing.md
+++ b/docs/markdown/Installing.md
@@ -84,13 +84,12 @@ empty file in a custom directory.
 ```bash
 #!/bin/sh
 
-mkdir "${DESTDIR}/${MESON_INSTALL_PREFIX}/mydir"
-touch "${DESTDIR}/${MESON_INSTALL_PREFIX}/mydir/file.dat"
+mkdir "${MESON_INSTALL_DESTDIR_PREFIX}/mydir"
+touch "${MESON_INSTALL_DESTDIR_PREFIX}/mydir/file.dat"
 ```
 
 As you can see, Meson sets up some environment variables to help you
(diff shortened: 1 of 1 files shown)


eli-schwartz:
This is factually incorrect, the --destdir argument to `meson install` is only used to modify the environment (`setenv` / `os.environ['...']=...`) regardless of the previous value of the environment. So, $DESTDIR is still always correct.

And in fact, it's absolutely, positively, utterly forbidden to stop providing $DESTDIR as that would be a breaking change in existing scripts, many of which aren't written specifically for meson or which need to modify things that aren't in $MESON_INSTALL_PREFIX, such as `$DESTDIR/var/lib` or `$DESTDIR/etc` which is fully impossible to do using MESON_  prefixed variables.

While it is true one *could* use either `$MESON_INSTALL_DESTDIR_PREFIX` or `$DESTDIR/$MESON_INSTALL_PREFIX`, the only actual difference between the two is that the former correctly handles Windows drive-prefixed paths for the prefix, which is... odd to hear as being a problem for DESTDIR as Windows doesn't really need such concepts -- there's no real concept of installing to a staging directory for packaging purposes there (in part because everything must be relocatable as nothing has a canonical installation path).

So I don't see much point of using this, especially given it only solves anything for prefix and not for sysconfdir or localstatedir.

PR #9507, opened 2021-10-31
minstall: Add more safety checks when unpickling installdata

xclaesse:
When need to catch exceptions just like we do in coredata.load() to
print proper error message instead of backtrace when user mix meson
versions.

This happens frequently when user has a newer version of meson installed
in their HOME and then "sudo meson install" uses the system version of
meson.

### DIFF
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -23,6 +23,7 @@ from .mesonlib import (
     MesonException, EnvironmentException, MachineChoice, PerMachine,
     PerMachineDefaultable, default_libdir, default_libexecdir,
     default_prefix, split_args, OptionKey, OptionType, stringlistify,
+    pickle_load
 )
 from .wrap import WrapMode
 import ast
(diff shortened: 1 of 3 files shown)


eli-schwartz:
Since all this code is copied from coredata.load() it would be nice to have a single shared implementation.

xclaesse:
> Since all this code is copied from coredata.load() it would be nice to have a single shared implementation.

Indeed, done now.

Note that won't fix the issue until this fix gets into the distro version of meson of course...

eli-schwartz:
Of course. :p

mesonbuild/coredata.py
```diff
@@ -1025,22 +1026,7 @@ def major_versions_differ(v1: str, v2: str) -> bool:
...
-        raise MesonException(
-            f"Coredata file {filename!r} references functions or classes that don't "
-            "exist. This probably means that it was generated with an old "
-            "version of meson.")
-    if not isinstance(obj, CoreData):
-        raise MesonException(load_fail_msg)
-    if major_versions_differ(obj.version, version):
-        raise MesonVersionMismatchException(obj.version, version)
-    return obj
+    pickle_load(filename, 'Coredata', CoreData)
```
eli-schwartz, line 1029: This needs to return the loaded coredata...

xclaesse: argh, of course, how did I miss that! Fixed.

mesonbuild/minstall.py
```diff
@@ -26,9 +25,8 @@
 
 from . import environment
 from .backend.backends import InstallData, InstallDataBase, InstallEmptyDir, TargetInstallData, ExecutableSerialisation
-from .coredata import major_versions_differ, MesonVersionMismatchException
 from .coredata import version as coredata_version
```
eli-schwartz, line 28: This is still lingering around and making lgtm unhappy.

xclaesse:
@dcbaker do you know if there is a better way to annotate `def pickle_load(filename: str, object_name: str, object_type: T.Type) -> T.Any:`? The return type of that function should be `object_type`.

eli-schwartz (approved):

PR #2237, opened 2017-08-22
Check if Watcom version of cl exists in the path and avoid using it.

cr1901:
This pull request fixes a bug where meson would attempt to run any `cl.exe` found on the path, regardless of whether it was Microsoft's version or not.

The OpenWatcom toolchain, used for vintage projects, provides its own `cl.exe` implementation that has a paged help text, expect user intervention to page through the help text. `meson` idles waiting for the compiler to finish running, which will never occur as the compiler is waiting for input.

My solution is to never attempt to use the watcom `cl.exe` (or POSIX `owcc`) wrapper if they are found on the path; they are buggy and can interfere with other vendors.

With this fix, I was able to generate a ninja file with `cc`, `gcc`, and `msvc` from the visual studio command prompt without error.

PR #9509, opened 2021-10-31
Add 64-bit paths to check for unsupported Watcom cl.exe clones.

cr1901:
This is an addition to #2237. I also added logging for when unsupported `cl.exe` binaries are detected :).


eli-schwartz:
LGTM, seems logical.

dcbaker (approved):
Lgtm

PR #9510, opened 2021-11-01
RFC: bump minimum version of python to python 3.7

eli-schwartz:
Bumping to a new version of python essentially has two facets to it.

- Is it still supported by the PSF?
  
  python 3.6 is the oldest still-supported version of python. And it is reaching EOL in less than 2 months, on December 23.

- Is it still required by supported OSes we care about?

  According to https://repology.org/project/python/versions I don't believe we have any issues here. You can even get python 3.7 in Ubuntu 18.04, apparently. It's just painful since the testsuite assumes `/usr/bin/env python3`, plus you need to install python3.6 to get pip for 3.7 -- brilliant!

  Due to the byzantine nature of some distros, python > 3.8, 3.9, 3.10 is available on OpenSUSE Tumbleweed, but the default python is 3.6? Maybe? And their website is aggressively uncooperative with the idea of people exploring availability for stuff. I am actually not sure how to navigate this. UPDATE: Currently this actually works, python3 now resolves to python38.

The main benefits which I've already discovered:
- the use of importlib.resources to reliably get bundled data as data files. No more need to freeze them as inline strings and test whether they are up to date. Sadly, to do this truly nicely we need a 3.9 API...
- future annotations means that typing is both easy and has little/no performance impact


eli-schwartz:
>     * You can even get python 3.7 in Ubuntu 18.04, apparently.

... but without pip, of course, yet the python3.7-venv patched thing without ensurepip.

anarazel:
RHEL7 / Centos7 are still widely used and only have python 3.6 available in the default repositories. So I'd prefer not to bump the minimum version unless it has quite substantial benefits.

xclaesse:
Latest RHEL8 also has 3.6 it seems and I don't even find a release schedule for RHEL9 :/

eli-schwartz:
CentOS 8 has a python38 and a python39 package according to repology.

We're not supporting python 3.6 in meson until 2024, when CentOS 7 is EOL. I'm pretty skeptical that we should continue supporting it past December of this year, when the PSF marks python 3.6 as EOL. Personally, I've been waiting for importlib.resources ever since #7461

Given the average release timeline of meson, we will probably be releasing 0.61 around or shortly after the python EOL, so now is a good time to start thinking about this.

xclaesse:
> CentOS 8 has a python38 and a python39 package according to repology.

oh I see, it has 3.6 by default but backports are available.

Personally I've got nothing against bumping the requirement. I know it will upset some users, it always does... OOC, besides `importlib.resources` what do we get from 3.7?

eli-schwartz:
- @bonzini may be able to simplify the mtest code using asyncio.run()
- PEP 563: Postponed Evaluation of Annotations, I should probably add `from __future__ import annotations` everywhere now which will make typing annotations close to zero runtime cost and no need to always quote them

bonzini:
For asyncio, create_task is nice to have but mostly cosmetic. 3.7 is a much smaller bump than 3.6.

eli-schwartz:
I mentioned future annotations. :D I'll probably go add those to this PR soon. Migrating to data classes is something hard to do without analyzing the code we have and figuring out *where* it's currently applicable, and I haven't done that kind of analysis (yet?)

dcbaker:
I'm find with not dealing with dataclasses in this PR, I'm just offering it as a nice feature of moving to 3.7

jpakkane:
We'll do this immediately after the next major release.

mesonbuild/backend/backends.py
```diff
@@ -61,11 +62,11 @@ class TargetIntrospectionData(TypedDict):
...
 
+@dataclass(eq=False)
 class RegenInfo:
-    def __init__(self, source_dir: str, build_dir: str, depfiles: T.List[str]):
-        self.source_dir = source_dir
-        self.build_dir = build_dir
-        self.depfiles = depfiles
+    source_dir: str
+    build_dir: str
+    depfiles: T.List[str]
```
scivision, line 69: with Python >= 3.7, if the file starts with

```python
from __future__ import annotations
```

and then annotation like:

```python
depfiles: list[str]
```

scivision:
Once Python 3.7 is the minimum based on research above and what I see, Python 3.8 minimum has little user downsides and lets further simplications/improvement of Meson internals.
https://packages.ubuntu.com/bionic/python/python3.8  Ubuntu 18.04 has Python 3.8 in default repos

mesonbuild/backend/backends.py
eli-schwartz: We could do that on 3.6 too. Future annotations just means we don't have to quote it.

It would then... fail with mypy on older versions of python before list became a generic type? I dunno. Does mypy explicitly handle that or just rely on python to evaluate the types. I'd assume the latter...

I guess it's not too terrible if lint_mypy requires newer python than meson itself.

eli-schwartz:
Debian Buster (oldstable) has 3.7, and then the next version of Debian jumps straight to 3.9.

As python 3.7 is not end of lifed (and won't be until 2023) I think it is likely premature to drop support for it when it is supported by both a distro and upstream python.org.

I do agree it would be pretty neat -- for example importlib.metadata implies a feature that could be added to the python module.

jpakkane:
⚔️ conflicts. Also should add the inverse of the deprecation notice MR.

eli-schwartz:
Yeah, I've been waiting for the release to rebase, because I knew it was inevitable and didn't want to keep doing it. :D

eli-schwartz:
The big fat rebase is in.

eli-schwartz:
Followup addition of release notes.

jpakkane:
Should #9802 go in first?

eli-schwartz:
Yes, because it will let the CI stay green and help keep track of what is actually failing due to there being an issue, and what doesn't fail except due to unrelated updates to mypy or whatever.

mesonbuild/backend/backends.py
dcbaker: It requires a pretty up to date mypy, which may or may not be nice. I'd rather make that change as a separate, explicit change anyway

eli-schwartz: On the other hand, this PR doesn't actually make use of future annotations. :D

I do have some preliminary work on slogging through that... https://github.com/eli-schwartz/meson/commits/python37-additions

dcbaker (approved):
I only gave it a quick glance this time (although we've talked about a number of the changes in this PR before), and it looks fine to me.

eli-schwartz:
The CI all passes, except for one. That one is `linux / Ubuntu Bionic` and it is expected to fail, because this PR includes critical updates to `CI image builder / Ubuntu Bionic`, which does run and passes, and will update the base CI images once merged to master.

PR #9511, opened 2021-11-01
MSVC: Support x86 compilation and xp compilation

deadash:
Two environment variables have been added so that the generated arch can be set, and the toolset can also be set.

If I need to generate x86 programs, I just need to simply set VS_ARCH to x86. Similarly, if I need to generate xp support, then I can set VS_TOOLSET to v141_xp.

Before that, it was very difficult to modify to x86 to compile or generate xp support, and it was necessary to modify the source code inside.

If you use conan, you can simply modify the code as follows
```
command = 'meson "%s" "%s" %s' % (source_dir, self.build_dir, arg_list)
with environment_append({"PKG_CONFIG_PATH": pc_paths,
    "VS_ARCH": self._settings.get_safe('arch'),
    "VS_TOOLSET": self._settings.get_safe('compiler.toolset')}):
    self._run(command)
```

### DIFF
--- a/mesonbuild/backend/vs2015backend.py
+++ b/mesonbuild/backend/vs2015backend.py
@@ -35,4 +35,4 @@ class Vs2015Backend(Vs2010Backend):
                     # We don't have support for versions older than 2019 right now.
                     raise MesonException('There is currently no support for ICL before 19, patches welcome.')
             if self.platform_toolset is None:
-                self.platform_toolset = 'v140'
+                self.platform_toolset = os.environ['VS_TOOLSET'] or 'v140'
--- a/mesonbuild/backend/vs2017backend.py
+++ b/mesonbuild/backend/vs2017backend.py
@@ -41,7 +41,7 @@ class Vs2017Backend(Vs2010Backend):
                         # We don't have support for versions older than 2019 right now.
                         raise MesonException('There is currently no support for ICL before 19, patches welcome.')
         if self.platform_toolset is None:
-            self.platform_toolset = 'v141'
+            self.platform_toolset = os.environ['VS_TOOLSET'] or 'v141'
         # WindowsSDKVersion should be set by command prompt.
         sdk_version = os.environ.get('WindowsSDKVersion', None)
         if sdk_version:
--- a/mesonbuild/backend/vs2019backend.py
+++ b/mesonbuild/backend/vs2019backend.py
@@ -35,7 +35,7 @@ class Vs2019Backend(Vs2010Backend):
                     self.platform_toolset = 'Intel C++ Compiler 19.0'
                 # We don't have support for versions older than 2019 right now.
             if not self.platform_toolset:
-                self.platform_toolset = 'v142'
+                self.platform_toolset = os.environ['VS_TOOLSET'] or 'v142'
             self.vs_version = '2019'
         # WindowsSDKVersion should be set by command prompt.
         sdk_version = os.environ.get('WindowsSDKVersion', None)
--- a/mesonbuild/mesonlib/vsenv.py
+++ b/mesonbuild/mesonlib/vsenv.py
@@ -12,7 +12,7 @@ from .universal import MesonException, is_windows
 
 bat_template = '''@ECHO OFF
 
-call "{}"
+call "{}" {}
 
 ECHO {}
 SET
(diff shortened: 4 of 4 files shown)


eli-schwartz:
> Before that, it was very difficult to modify to x86 to compile or generate xp support, and it was necessary to modify the source code inside.

The VS_ARCH change certainly doesn't help you with that -- simply do not use `meson setup --vsenv`, but activate the environment first yourself. It is also incredibly inconsistent, it won't remember what it did last time.

Also why does meson need code to work around support for Windows XP, which has been dead for 7 and a half years?

deadash:
> > Before that, it was very difficult to modify to x86 to compile or generate xp support, and it was necessary to modify the source code inside.
> 
> The VS_ARCH change certainly doesn't help you with that -- simply do not use `meson setup --vsenv`, but activate the environment first yourself. It is also incredibly inconsistent, it won't remember what it did last time.
> 
> Also why does meson need code to work around support for Windows XP, which has been dead for 7 and a half years?

Xp is one of the toolset, for example, I can also compile v141, or v140 on vs2019, and these operations will work.

Sorry, I don't know how to compile a 32-bit program, if backend is vs.

If you need to manually execute vcvars.bat, I think setup_venv is useless, if the backend is ninja, you can indeed generate 32-bit programs directly.

jpakkane:
There are two different issues here. The first one is setting the toolset version. From what I can tell that does not come from VS but is instead something new. This is not really good for two different reasons, first of all it is a design point of Meson to never do configuration via environment variables (of our own) and the second is that this detection should be automatic.

That is the toolset version should be detected from the state set by vcvarsall or equivalent. If you run the configuration in an x86 command prompt then it should generate 32 bit solution files.

> If you need to manually execute vcvars.bat, I think setup_venv is useless,

Yes, and that is the point. The heuristic is fairly simple:

- if a VS environment is already set, use that
- if it is not set and there are no toolchains available (like gcc or clang) try to activate the default toolset (x86_64)

If the toolset version is not detectable via the things VS sets up then a better solution would be to add a backend-specific option for setting it manually.

deadash:
According to what you said, I think I can add options to modify the toolset, similar to b_vscrt. As for the heuristic method, because the toolset must be manual, the default vs2022 corresponds to v143, vs2019 corresponds to v142, and vs2017 corresponds to v141, so there is no setting. The following are the default values, if set, you can also change them.

For 32-bit programs, I need to manually call setup_venv to find the vc32 environment, and then the generated 32-bits will follow. This can also be changed by configuration, because for crossfile, the vc32-bit is not properly configured. For ninja, It is normal to generate 32 bits.

Similarly, the toolset does not work at all in ninja, and it needs to be modified here.

xclaesse:
> if it is not set and there are no toolchains available (like gcc or clang) try to activate the default toolset (x86_64)

Note that we now have command line arguments `--vsenv` or `--backend vs` to override that. Meson will activate and use VisualStudio even if gcc is found in that case (e.g. Windows GitHub action). One thing I had in mind is to extend the command line to `--vsenv ARCH` to be able to build for x86.

deadash:
@xclaesse 
Yes, it is feasible to modify the --vsenv parameter to arch

It would be even better if it can support multiple arches, such as --vsenv x86,x64, and then the generated vs solution contains win32 and x64.

In bazel, this step can be achieved by converting ET, cmake comes with two arches, it will be even better if meson supports it.

deadash:
I checked ninja and found that ninja relies on environment variables, and vcvars no longer supports xp, so it becomes very difficult to modify ninja to support toolset.

So using vs as the backend to support toolset is very important. I also need to introduce an option to set toolset instead of the default vs2019 corresponding to v142, because v141 can also be used in vs2019, or llvm_v142 can be used instead of clang directly. Just as easy as CMAKE using -T parameter.

deadash:
@eli-schwartz 
I have changed it now, add a parameter named b_vstoolset to set the toolset, do I reopen the pr or submit a new pr

deadash:
Add support options as follows
```
    OptionKey('b_vstoolset'): coredata.UserComboOption('VS toolset to use.',
                                                   ['none', 'v90', 'v100', 'v110', 'v110_xp', 'v120', 'v120_xp',
                                                   'v140', 'v140_xp', 'v140_clang_c2', 'LLVM-vs2012', 'llvm-VS2012_xp',
                                                   'LLVM-vs2013', 'LLVM-vs2013_xp', 'LLVM-vs2014', 'LLVM-vs2014_xp',
                                                   'LLVM-vs2017', 'LLVM-vs2017_xp', 'v141', 'v141_xp', 'v141_clang_c2',
                                                   'LLVM', 'LLVM_v141', 'LLVM_v141_xp', 'LLVM_v142'
                                                   'v142', 'v143'],
                                                   'none'),
```

edsky:
can you reopen it? jpakkane

eli-schwartz:
No, in order to re-open the PR the original submitter must ensure the state of the fork matches what it was when the PR got closed.

It looks like github is claiming the repository got deleted??? I'm not sure what is up with that.

Anyway, given github simply doesn't allow it at this point, I would say just open a new PR. But please, next time do not close a PR instead of updating it.

deadash:
Okay, I will never open the repeated pr next time.

First of all, I only submit questions about vs toolset. Arch issues can be solved using the command line. I have submitted pr to conan to support vs arch.

PR #9512, opened 2021-11-01
interpreter: void holder

mensinda:
The goal of this refactoring is to slightly simplify the logic in `InterpreterBase` by ensuring that even `None` now has a corresponding holder. This way we no longer have to check for it and can always assume that we are dealing with `InterpreterObject`s.


mesonbuild/interpreterbase/baseobjects.py
```diff
@@ -75,6 +75,11 @@ def __init__(self, *, subproject: T.Optional[str] = None) -> None:
     def display_name(self) -> str:
         return type(self).__name__
 
+    @property
+    def is_assignable(self) -> bool:
+        ''' Property used to indicate whether an object can be used at all '''
+        return True
```
dcbaker, line 81: properties are really slow, you generally want to avoid using them if at all possible, I'd make this a class variable instead, set the base class to `True`, and have the `VoidHolder` class override it in its class definition.

mesonbuild/interpreterbase/void.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2021 The Meson development team
+# SPDX-license-identifier: Apache-2.0
```
dcbaker, line 2:
```suggestion
# SPDX-License-Identifier: Apache-2.0
```

dcbaker (changes requested):
A couple of small nits, but otherwise this is a nice change.

PR #9513, opened 2021-11-01
MSVC: Support x86 compilation and xp compilation

deadash:
Two environment variables have been added so that the generated arch can be set, and the toolset can also be set.

If I need to generate x86 programs, I just need to simply set VS_ARCH to x86. Similarly, if I need to generate xp support, then I can set VS_TOOLSET to v141_xp.

Before that, it was very difficult to modify to x86 to compile or generate xp support, and it was necessary to modify the source code inside.

If you use conan, you can simply modify the code as follows

```
command = 'meson "%s" "%s" %s' % (source_dir, self.build_dir, arg_list)
with environment_append({"PKG_CONFIG_PATH": pc_paths,
    "VS_ARCH": self._settings.get_safe('arch'),
    "VS_TOOLSET": self._settings.get_safe('compiler.toolset')}):
    self._run(command)
```

@eli-schwartz 

> The VS_ARCH change certainly doesn't help you with that -- simply do not use meson setup --vsenv, but activate the environment first yourself. It is also incredibly inconsistent, it won't remember what it did last time.

> Also why does meson need code to work around support for Windows XP, which has been dead for 7 and a half years?

Xp is one of the toolset, for example, I can also compile v141, or v140 on vs2019, and these operations will work.

Sorry, I don't know how to compile a 32-bit program, if backend is vs.

If you need to manually execute vcvars.bat, I think setup_venv is useless, if the backend is ninja, you can indeed generate 32-bit programs directly.

### DIFF
--- a/mesonbuild/backend/vs2015backend.py
+++ b/mesonbuild/backend/vs2015backend.py
@@ -17,6 +17,7 @@ from ..mesonlib import MesonException
 from ..interpreter import Interpreter
 from ..build import Build
 import typing as T
+import os
 
 
 class Vs2015Backend(Vs2010Backend):
@@ -35,4 +36,4 @@ class Vs2015Backend(Vs2010Backend):
                     # We don't have support for versions older than 2019 right now.
                     raise MesonException('There is currently no support for ICL before 19, patches welcome.')
             if self.platform_toolset is None:
-                self.platform_toolset = 'v140'
+                self.platform_toolset = os.environ.get('VS_TOOLSET', 'v140')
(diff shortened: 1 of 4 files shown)


eli-schwartz:
Duplicate of #9511

deadash:
There is a problem with that, I modified it. @eli-schwartz

deadash:
@eli-schwartz Can you reopen it? There was an error in the previous one, I modified it

eli-schwartz:
Can you please not open duplicate PRs instead of modifying the one you've already opened?

PR #9514, opened 2021-11-01
Add tutorial for building an SDL app from scratch.

jpakkane:


docs/markdown/GuiTutorial.md
```diff
@@ -0,0 +1,194 @@
...
+
+```meson
+project('sdldemo', 'c')
+
+executable('sdlprog', 'sdlprog.c')
+```
+
+With this done we can start the build with the following command:
+
+    meson build
```
eli-schwartz, line 66: It's somewhat customary in our docs to refer to `builddir` to make it more obvious that it is not a subcommand, but a directory name.

eli-schwartz (changes requested):

docs/markdown/GuiTutorial.md
jpakkane: Updated.

PR #3413, opened 2018-04-15
run_command: Add new kwarg 'capture'

nirbheek:
`capture: false` means we won't try to read the output at all.

Closes https://github.com/mesonbuild/meson/issues/3364

Will add docs and a release snippet if this is approved.

PR #9515, opened 2021-11-01
use typed_* args for the `run_command` function

dcbaker:
There is one strange thing about this function, we allow `build.Executable` to be passed through the type guards, even though it's actually an error to use it. We do this so that we can generate a better error message.

This also adds support for passing `Compiler` objects to any of the positional arguments of `run_command`, not just the first one.

### DIFF
--- a/mesonbuild/interpreter/interpreter.py
+++ b/mesonbuild/interpreter/interpreter.py
@@ -32,7 +32,7 @@ from ..interpreterbase import InterpreterException, InvalidArguments, InvalidCod
 from ..interpreterbase import Disabler, disablerIfNotFound
 from ..interpreterbase import FeatureNew, FeatureDeprecated, FeatureNewKwargs, FeatureDeprecatedKwargs
 from ..interpreterbase import ObjectHolder
-from ..interpreterbase.baseobjects import TYPE_nkwargs, TYPE_var, TYPE_kwargs
+from ..interpreterbase.baseobjects import TYPE_var, TYPE_kwargs
 from ..modules import ExtensionModule, ModuleObject, MutableModuleObject, NewExtensionModule, NotFoundExtensionModule
 from ..cmake import CMakeInterpreter
 from ..backend.backends import Backend, ExecutableSerialisation
@@ -721,12 +721,13 @@ external dependencies (including libraries) must go to "dependencies".''')
 
         m = 'must be a string, or the output of find_program(), files() '\
             'or configure_file(), or a compiler object; not {!r}'
+        overridden_msg = ('Program {!r} was overridden with the compiled '
+                          'executable {!r} and therefore cannot be used during '
+                          'configuration')
         expanded_args: T.List[str] = []
         if isinstance(cmd, build.Executable):
             progname = node.args.arguments[0].value
-            msg = 'Program {!r} was overridden with the compiled executable {!r}'\
-                    ' and therefore cannot be used during configuration'
-            raise InterpreterException(msg.format(progname, cmd.description()))
+            raise InterpreterException(overridden_msg.format(progname, cmd.description()))
         if isinstance(cmd, ExternalProgram):
             if not cmd.found():
                 raise InterpreterException(f'command {cmd.get_name()!r} not found or not executable')
@@ -750,15 +751,22 @@ external dependencies (including libraries) must go to "dependencies".''')
             if not prog.found():
                 raise InterpreterException(f'Program or command {cmd!r} not found or not executable')
             cmd = prog
-        for a in listify(cargs):
+        for a in cargs:
             if isinstance(a, str):
                 expanded_args.append(a)
             elif isinstance(a, mesonlib.File):
                 expanded_args.append(a.absolute_path(srcdir, builddir))
             elif isinstance(a, ExternalProgram):
                 expanded_args.append(a.get_path())
+            elif isinstance(a, compilers.Compiler):
+                FeatureNew.single_use('Compiler object as a variadic argument to `run_command`', '0.61.0', self.subproject, location=node)
+                prog = ExternalProgram(a.exelist[0], silent=True)
+                if not prog.found():
+                    raise InterpreterException(f'Program {cmd!r} not found or not executable')
+                expanded_args.append(prog.get_path())
             else:
-                raise InterpreterException('Arguments ' + m.format(a))
+                raise InterpreterException(overridden_msg.format(a.name, cmd.description()))
+
         # If any file that was used as an argument to the command
         # changes, we must re-run the configuration step.
         self.add_build_def_file(cmd.get_path())
@@ -766,6 +774,7 @@ external dependencies (including libraries) must go to "dependencies".''')
             if not os.path.isabs(a):
                 a = os.path.join(builddir if in_builddir else srcdir, self.subdir, a)
             self.add_build_def_file(a)
+
         return RunProcess(cmd, expanded_args, env, srcdir, builddir, self.subdir,
                           self.environment.get_build_command() + ['introspect'],
                           in_builddir=in_builddir, check=check, capture=capture)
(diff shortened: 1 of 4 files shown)


mesonbuild/interpreter/interpreter.py
```diff
@@ -736,15 +746,21 @@ def run_command_impl(self,
...
-        for a in listify(cargs):
+        for a in cargs:
             if isinstance(a, str):
                 expanded_args.append(a)
             elif isinstance(a, mesonlib.File):
                 expanded_args.append(a.absolute_path(srcdir, builddir))
             elif isinstance(a, ExternalProgram):
                 expanded_args.append(a.get_path())
+            elif isinstance(a, compilers.Compiler):
+                FeatureNew('Compiler boject as a variadic argument to `run_command`', '0.61.0')
```
eli-schwartz, line 757:
```suggestion
                FeatureNew('Compiler object as a variadic argument to `run_command`', '0.61.0')
```

mesonbuild/interpreter/interpreter.py
```diff
@@ -678,42 +678,52 @@ def validate_arguments(self, args, argcount, arg_types):
...
             check = False
 
-        if not isinstance(check, bool):
-            raise InterpreterException('Check must be boolean.')
-
-        env = self.unpack_env_kwarg(kwargs)
 
         m = 'must be a string, or the output of find_program(), files() '\
             'or configure_file(), or a compiler object; not {!r}'
+        overriden_msg = ('Program {!r} was overridden with the compiled '
```
eli-schwartz, line 719: overridden has two "d"s in it, so this looks odd -- although of course it's an internal implementation detail and a perfectly valid variable reference string.

dcbaker: I reworded this slightly, and used the `overrode` instead of `overridden`.

dcbaker:
And mypy is catching all the bugs :)

eli-schwartz:
Please do not delete codecov comments :D codecov will repeatedly update the existing comment, but if you delete the comment, it will repeatedly post new comments...

mesonbuild/interpreter/interpreter.py
eli-schwartz: in the prose message at least, it *needs* to be overridden.

Present: Program X is override  with the compiled"
Past: Program X is overrode with the compiled

dcbaker:
@eli-schwartz wow, that's bad behavior.

mesonbuild/interpreter/interpreter.py
dcbaker: back to overridden it is.

dcbaker:
@eli-schwartz, I've reverted the message back to the original message. How do you feel about this version?

mesonbuild/interpreter/interpreter.py
```diff
@@ -677,42 +677,51 @@ def validate_arguments(self, args, argcount, arg_types):
                 if not isinstance(actual, wanted):
                     raise InvalidArguments('Incorrect argument type.')
 
-    @FeatureNewKwargs('run_command', '0.50.0', ['env'])
-    @FeatureNewKwargs('run_command', '0.47.0', ['check', 'capture'])
-    @permittedKwargs({'check', 'capture', 'env'})
-    def func_run_command(self, node, args, kwargs):
+    # Executables aren'ta actually accepted, but we allow them here to allow for
```
eli-schwartz, line 680:
```suggestion
    # Executables aren't actually accepted, but we allow them here to allow for
```

mesonbuild/interpreter/interpreter.py
```diff
@@ -677,42 +677,51 @@ def validate_arguments(self, args, argcount, arg_types):
...
+    # better error messages when overriden
```
eli-schwartz, line 681:
```suggestion
    # better error messages when overridden
```

eli-schwartz (commented):
Typos are creeping invaders from the Dungeon Dimensions, a couple more have slipped their way in.

User-facing message looks fine I guess.

dcbaker:
It's super annoying that my spell checker insists that code files cannot be spell-checked, only "text" flies.

dcbaker:
I don't understand the msys failure, apparently some test ran that wasn't expected? but I didn't modify anything to do with that test?

eli-schwartz:
That test has been unexpectedrun on other commits as well, it's definitely not unique to this PR.

dcbaker:
Well that's good to know

eli-schwartz:
After rebasing on master the CI should stop mysteriously failing, and I think otherwise this looked fine?

mesonbuild/interpreter/interpreter.py
eli-schwartz: For the record, this seems to have been corrected in the wrong commit which makes it look a bit odd when following the commit difflog.

mesonbuild/interpreter/interpreter.py
```diff
@@ -677,42 +677,41 @@ def validate_arguments(self, args, argcount, arg_types):
...
+    # Executables aren't actually accepted, but we allow them here to allow for
+    # better error messages when overridden
     @typed_pos_args(
         'run_command',
         (build.Executable, ExternalProgram, compilers.Compiler, mesonlib.File, str),
         varargs=(build.Executable, ExternalProgram, compilers.Compiler, mesonlib.File, str))
+    @typed_kwargs(
+        'run_command',
+        KwargInfo('check', (bool, NoneType), since='0.47.0'),
+        KwargInfo('capture', bool, default=True, since='0.47.0'),
```
eli-schwartz, line 689: This was added in #3413 without reference manual documentation. :(

mesonbuild/interpreter/interpreter.py
```diff
@@ -742,15 +743,21 @@ def run_command_impl(self,
...
-        for a in listify(cargs):
+        for a in cargs:
             if isinstance(a, str):
                 expanded_args.append(a)
             elif isinstance(a, mesonlib.File):
                 expanded_args.append(a.absolute_path(srcdir, builddir))
             elif isinstance(a, ExternalProgram):
                 expanded_args.append(a.get_path())
+            elif isinstance(a, compilers.Compiler):
+                FeatureNew('Compiler object as a variadic argument to `run_command`', '0.61.0')
```
eli-schwartz, line 754: This can accept a `location=node` now, to provide line numbers for the exact place in meson.build this feature got triggered.

mesonbuild/interpreter/interpreter.py
```diff
@@ -742,15 +743,21 @@ def run_command_impl(self,
...
+                prog = ExternalProgram(a.exelist[0], silent=True)
+                if not prog.found():
+                    raise InterpreterException(f'Program {cmd!r} not found or not executable')
```
eli-schwartz, line 757: Is this actually possible? Can users get this some way other than meson.get_compiler() which only works for languages that got detected and sanity checked?

dcbaker: not sure, it may have just been me programming defensively, but I can't remember exactly why now.

mesonbuild/interpreter/interpreter.py
dcbaker: At least it'll be fixed now :D

mesonbuild/interpreter/interpreter.py
dcbaker: cool, done. Also apparently missing the `single_use` call. derp

mesonbuild/interpreter/interpreter.py
eli-schwartz: err that is to say I mean these: https://github.com/mesonbuild/meson/pull/9515#pullrequestreview-796720479

```diff
-    def func_run_command(self, node, args, kwargs):
+    # Executables aren'ta actually accepted, but we allow them here to allow for
+    # better error messages when overriden
+    @typed_pos_args(
+        'run_command',
```

Followed by:
```diff
-    @permittedKwargs({'check', 'capture', 'env'})
-    # Executables aren'ta actually accepted, but we allow them here to allow for
-    # better error messages when overriden
+    # Executables aren't actually accepted, but we allow them here to allow for
+    # better error messages when overridden
```

eli-schwartz (approved):

PR #9520, opened 2021-11-02
Add typed_kwargs to some of the gnome module functions

dcbaker:
This is not complete for the module, but it is a start. I've broken this up as these are fairly complex changes:
- A fair number of helper functions are ripped out as they are not helpful once we konw that the key in the dictionary exists
- avoiding kwargs mutation
- moving invalid argument handling to the typed_kwarg helper

Since this is already around 750 LOC change,  I thought it would be easier to review in smaller pieces like this.

### DIFF
--- a/mesonbuild/modules/gnome.py
+++ b/mesonbuild/modules/gnome.py
@@ -32,7 +32,7 @@ from .. import mlog
 from ..build import CustomTarget, CustomTargetIndex, GeneratedList
 from ..dependencies import Dependency, PkgConfigDependency, InternalDependency
 from ..interpreter.type_checking import DEPEND_FILES_KW
-from ..interpreterbase import noPosargs, noKwargs, permittedKwargs, FeatureNew, FeatureNewKwargs, FeatureDeprecatedKwargs, FeatureDeprecated
+from ..interpreterbase import noPosargs, noKwargs, permittedKwargs, FeatureNew, FeatureNewKwargs, FeatureDeprecatedKwargs
 from ..interpreterbase import typed_kwargs, KwargInfo, ContainerTypeInfo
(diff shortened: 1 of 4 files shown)


dcbaker:
the LGTM error is silly, I've already documented the assumptions of the function, and it's called after a validator that enforces that the assumptions are true.

mesonbuild/modules/gnome.py
```diff
@@ -979,36 +1013,38 @@ def generate_gir(self, state: 'ModuleState', args: T.Tuple[T.List[T.Union[build.
...
+        'gnome.yelp',
+        KwargInfo('languages', ContainerTypeInfo(list, str), listify=True, default=[]),
+        KwargInfo('media', ContainerTypeInfo(list, str), listify=True, default=[]),
+        KwargInfo('sources', ContainerTypeInfo(list, str), listify=True, default=[]),
+        KwargInfo('symlink_media', bool, default=True),
+    )
+    def yelp(self, state: 'ModuleState', args: T.Tuple[str, T.List[str]], kwargs: 'Yelp') -> ModuleReturnValue:
         project_id = args[0]
-        if len(args) > 1:
-            FeatureDeprecated.single_use('gnome.yelp more than one positional argument', '0.60.0', 'use the "sources" keyword argument instead.')
```
eli-schwartz, line 1009: While looking at a conflict with a PR of mine after this got merged, I noticed that this FeatureDeprecated went missing. However, as far as I can tell this vararg is still used, and isn't raising a warning any other way. (It doesn't seem like typed_pos_args handles that either way).

dcbaker:
I'm not sure what happened, I'll send another PR

PR #9521, opened 2021-11-02
wrap: Fix concurrent os.mkdir()

xclaesse:
Since 0.59.0 Meson downloads multiple wraps in parallel, so the
packagecache directory could be created by one then the 2nd would hit
error when calling os.mkdir() because it already exists.


xclaesse:
This fix sporadic CI failure in libnice such as https://gitlab.freedesktop.org/libnice/libnice/-/jobs/15364684.

eli-schwartz:
What, you mean we can fix bugs by using simpler, more concise code???

:D lgtm

PR #9526, opened 2021-11-04
vs: Add b_vstoolset parameter to support setting vs toolset.

deadash:
#9511 Add a new parameter to modify the vs toolset.

I don't know whether to write a test or not, because I am not sure whether the specified toolset is installed in the test CI, such as installing vs2017 compilation on vs2019.

I added `-Db_vstoolset=v141_xp` to make sure it is effective. This is only effective for backend to vs. Ninja itself does not support modifying the toolset. CMake does support it. It is not clear whether it is necessary to support cmake.


mesonbuild/compilers/compilers.py
```diff
@@ -296,6 +296,14 @@ class CompileCheckMode(enum.Enum):
...
                                                    ['none', 'md', 'mdd', 'mt', 'mtd', 'from_buildtype', 'static_from_buildtype'],
                                                    'from_buildtype'),
+    OptionKey('b_vstoolset'): coredata.UserComboOption('VS toolset to use.',
+                                                   ['none', 'v90', 'v100', 'v110', 'v110_xp', 'v120', 'v120_xp',
+                                                   'v140', 'v140_xp', 'v140_clang_c2', 'LLVM-vs2012', 'llvm-VS2012_xp',
+                                                   'LLVM-vs2013', 'LLVM-vs2013_xp', 'LLVM-vs2014', 'LLVM-vs2014_xp',
+                                                   'LLVM-vs2017', 'LLVM-vs2017_xp', 'v141', 'v141_xp', 'v141_clang_c2',
+                                                   'LLVM', 'LLVM_v141', 'LLVM_v141_xp', 'LLVM_v142',
+                                                   'v142', 'v143'],
+                                                   'none'),
```
nirbheek, line 306: The naming scheme here has inconsistent capitalization, underscore vs hyphen, etc. At the very least it should be all lowercase and use either hyphens or underscores. Not both interchangeably.

deadash: This is the name of vs own toolset. I didn’t name it. I try to collect it comprehensively.

deadash: I have removed clang now, so that the naming scheme is consistent, and most of what we want to use is the remaining vstoolset.

PR #9529, opened 2021-11-04
Fix a typo

rusty-snake:
old: `No such build data file as`
new: `No such build data file at`

### DIFF
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -2826,7 +2826,7 @@ def get_sources_string_names(sources, backend):
 def load(build_dir: str) -> Build:
     filename = os.path.join(build_dir, 'meson-private', 'build.dat')
     load_fail_msg = f'Build data file {filename!r} is corrupted. Try with a fresh build tree.'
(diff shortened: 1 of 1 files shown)


eli-schwartz:
"no such thing as XXXX" is not the same meaning as "no such thing at location XXXX". I guess it's probably subjective which one is a better message, but I'm pretty sure that it's not actually a typo. Just a different expository style.

I'm undecided about this change, but if we are going to make it then I think the word "such" should be removed at the same time...

xclaesse:
I would reformulate to "build data not found at" for example.

dcbaker:
Using `at` here feels awkward, as a native speaker (with that intuative "this feels wierd but I can't explain why" sense going)

Maybe `expected file named "{filename}" does not exist`?

QuLogic:
Just remove the preposition? 'No such build data file "xyz.ext"' is fine.

Also, the `!r` format adds quotes to a string already, so the explicit quotes in `nonexisting_fail_msg` should be removed (just like `load_fail_msg` doesn't have any).

QuLogic:
`filename` is a `str`; see the type annotations.

PR #9531, opened 2021-11-04
More use of typed_* in the interpreter module

dcbaker:
This covers many of the non target uses remaining, `add_test_setup`, `project`, `summary`, and `find_program`. Some of these already used `typed_pos_args`, but not `typed_kwargs`

### DIFF
--- a/mesonbuild/interpreter/interpreter.py
+++ b/mesonbuild/interpreter/interpreter.py
@@ -1489,23 +1489,27 @@ external dependencies (including libraries) must go to "dependencies".''')
         self.do_subproject(fallback, 'meson', sp_kwargs)
         return self.program_from_overrides(args, extra_info)
 
-    @FeatureNewKwargs('find_program', '0.53.0', ['dirs'])
-    @FeatureNewKwargs('find_program', '0.52.0', ['version'])
-    @FeatureNewKwargs('find_program', '0.49.0', ['disabler'])
-    @disablerIfNotFound
-    @permittedKwargs({'required', 'native', 'version', 'dirs'})
     @typed_pos_args('find_program', varargs=(str, mesonlib.File), min_varargs=1)
-    def func_find_program(self, node: mparser.BaseNode, args: T.Tuple[T.List[mesonlib.FileOrString]], kwargs) -> T.Union['build.Executable', ExternalProgram, 'OverrideProgram']:
+    @typed_kwargs(
+        'find_program',
+        DISABLER_KW.evolve(since='0.49.0'),
+        NATIVE_KW,
+        REQUIRED_KW,
+        KwargInfo('dirs', ContainerTypeInfo(list, str), default=[], listify=True, since='0.53.0'),
+        KwargInfo('version', ContainerTypeInfo(list, str), default=[], listify=True, since='0.52.0'),
+    )
+    @disablerIfNotFound
+    def func_find_program(self, node: mparser.BaseNode, args: T.Tuple[T.List[mesonlib.FileOrString]],
+                          kwargs: 'kwargs.FindProgram',
+                          ) -> T.Union['build.Executable', ExternalProgram, 'OverrideProgram']:
         disabled, required, feature = extract_required_kwarg(kwargs, self.subproject)
         if disabled:
             mlog.log('Program', mlog.bold(' '.join(args[0])), 'skipped: feature', mlog.bold(feature), 'disabled')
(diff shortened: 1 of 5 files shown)


mesonbuild/interpreter/interpreter.py
```diff
@@ -1493,19 +1493,17 @@ def find_program_fallback(self, fallback, args, required, extra_info):
     @FeatureNewKwargs('find_program', '0.49.0', ['disabler'])
     @disablerIfNotFound
     @permittedKwargs({'required', 'native', 'version', 'dirs'})
-    def func_find_program(self, node, args, kwargs) -> T.Union['build.Executable', ExternalProgram, 'OverrideProgram']:
-        if not args:
-            raise InterpreterException('No program name specified.')
-
+    @typed_pos_args('find_program', varargs=(str, mesonlib.File), min_varargs=1)
```
eli-schwartz, line 1492: Why do we allow mesonlib.File here? It was needed before all callers of self.func_find_program got fixed to use the impl, but, we do not document that this should work:

```meson
prog = files('foo.py')
prog = find_program(prog)
```

And it's super pointless to allow it. Anywhere that should transparently convert a files object into a found program, which I think literally just means configure_file, can upgrade that behind the scenes at time-of-use.

mesonbuild/interpreter/interpreter.py
```diff
@@ -1019,16 +1029,29 @@ def set_backend(self):
...
+    @typed_kwargs(
+        'project',
+        DEFAULT_OPTIONS,
+        KwargInfo('meson_version', (str, NoneType)),
+        KwargInfo(
+            'version',
+            (str, mesonlib.File, NoneType, list),
+            default='undefined',
+            validator=_project_version_validator,
+            convertor=lambda x: x[0] if isinstance(x, list) else x,
```
eli-schwartz, line 1038: What all is this even doing? We don't document this accepts lists at all.

> when passed as array must have a length of 1

Why does this warning message exist like that?

Is there some perfectly dreadful reason we need to keep supporting single-element list literals here for backwards compat, even though we never allowed it and anyone doing it was completely wrong?



```python
        version = kwargs.get('version', 'undefined')
        if isinstance(version, list):
            if len(version) != 1:
                raise InvalidCode('Version argument is an array with more than one entry.')
            version = version[0]
```

This smells to me like badly designed error checking, because what we actually want is: 3f0a0c1582f55e0da950a7a321d25ffcf240ec6f

We can't directly accept a mesonlib.File seemingly because files() returns a list of them (but people can technically do `files('VERSION')[0]`, it just looks odd, so maybe we *should* keep accepting that type), but we should be checking, if it is a list, that it has exactly one element and that element is a File object, *not* a string.

mesonbuild/interpreter/interpreter.py
eli-schwartz: Although I guess the correct solution here is to only allow configure_file outputs, which could be done by simply subclassing mesonlib.File with no changes? Maybe?

mesonbuild/interpreter/interpreter.py
dcbaker: I tried to get rid of it, but found some case where we really did need a list, but I can't remember now what it was

eli-schwartz: The original "support" for lists was only added in the same commit that added files() support -- again, files() returns an `array[file]`, so what we actually want to accept here is `str | array[file]`.

dcbaker: That's it. And that's a lie of work just avoid taking an array of string

dcbaker: Gah, typing in a moving car is a pain

eli-schwartz: It's not a lot of work, it's just removing `str` from the validator.

dcbaker: By work, it seems like saying `project(version : [files('foo')])` Okay! but `project(version : ['0.1.2'])`, nope! is a lot of work, somewhat confusing, and just not that useful? :shrug:

eli-schwartz: Actually `[files('foo')]` becomes `array[array[file]]` and was illegal either way. :D

dcbaker: But it get's flattened to `[]file`, right?

PR #9532, opened 2021-11-04
fix regression that broke string.format with list objects

### LINKED ISSUE #9530: [regression] String formatting has regressed since 0.60. ERROR: Format placeholder @9@ out of range.
**Describe the bug**
There is a regression of as yet unknown cause in Meson's string formatting, as discovered by @alex3kov and diagnosed by myself at libsixel/libsixel#50. This regression causes previously building package to no longer build.

**To Reproduce**
https://github.com/libsixel/libsixel/blob/v1.10.3/meson.build

**Expected behavior**
Good build

**system parameters**
    Plain build.
    Arch Linux
    Python 3.9.7
* what `meson --version`
    Versions above 0.59.2 I confirm are affected.
    We do not get to Ninja step, Meson is the cause.

eli-schwartz:
String formatting should validly assume that printing a list means printing the list itself. Instead, something like this broke:

```
'one is: @0@ and two is: @1@'.format(['foo',  'bar'], ['baz'])
```

which would evaluate as:

```
'one is: foo and two is: bar'
```

or:

```
'the value of array option foobar is: @0@'.format(get_option('foobar'))
```

which should evaluate with '-Dfoobar=[]' as

```
'the value of array option foobar is: []'
```

But instead produced:

```
meson.build:7:0: ERROR: Format placeholder @0@ out of range.
```

Fixes #9530

### DIFF
--- a/mesonbuild/interpreter/primitives/string.py
+++ b/mesonbuild/interpreter/primitives/string.py
@@ -13,6 +13,7 @@ from ...interpreterbase import (
     MesonOperator,
     FeatureNew,
     typed_operator,
+    noArgsFlattening,
     noKwargs,
     noPosargs,
     typed_pos_args,
@@ -85,6 +86,7 @@ class StringHolder(ObjectHolder[str]):
     def endswith_method(self, args: T.Tuple[str], kwargs: TYPE_kwargs) -> bool:
         return self.held_object.endswith(args[0])
 
+    @noArgsFlattening
     @noKwargs
     @typed_pos_args('str.format', varargs=object)
     def format_method(self, args: T.Tuple[T.List[object]], kwargs: TYPE_kwargs) -> str:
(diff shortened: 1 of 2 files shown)


dcbaker (approved):

eli-schwartz:
@lazka

```
 [UNEXRUN]   frameworks: 15 llvm    (method=cmake link-static=True)
```

This is configured as:

```
        { "val": "config-tool", "skip_on_jobname": ["msys2-gcc"]},
```

but apparently it now works? Do you know why, or if it is expected to stay that way? (I genuinely have no idea when tests are/were expected to be skipped, or why, to begin with.)

lazka:
I guess you mean `{ "val": "cmake", "skip_on_jobname": ["msys2"] }`?

Why was it skipped in the first place?

lazka:
ah, I'd guess this is https://github.com/msys2/MINGW-packages/commit/345fe436e8dfdc3527791ca262fa2e77855edda1, so likely fixed and can be unskipped.

lazka:
I've created #9530 9534

Issue #5793, opened 2019-08-13
ninja install build gtk-doc as root


grimmy:
While working on many of my meson projects, if I run `ninja -C build install` on a tree thats using a default prefix (`/usr/local`) I'll get prompted by policykit for the root password once the build is completed.  However, since `gtk-doc` targets aren't run until after the policykit prompt, and thus privilege escalation, they get built as root.  This then makes it impossible to `ninja -C build foo-doc` as a user because the files are owned by root.

I'm running meson 0.51.1 on debian unstable.  I can provide repos for reference or an asciinema as well, but making a small test case is kind of involved so I haven't done that yet.

QuLogic:
You can reproduce with one of the test cases:
```
$ ./meson.py test cases/frameworks/10\ gtk-doc/ work\ area/
$ ninja -C work\ area/ install
$ ls -l work\ area/doc/*
```
and verify that all these files are owned by `root:root`, and then things fail if you try to build the docs yourself:
```
$ touch test\ cases/frameworks/10\ gtk-doc/doc/foobar1/foobar-docs.sgml 
 
$ ninja -C work\ area/ foobar-doc
ninja: Entering directory `work area/'
[1/1] Generating foobar-doc with a custom command.
FAILED: doc/foobar1/foobar-decl.txt 
/usr/bin/python3 /home/elliott/code/meson/meson.py --internal gtkdoc '--sourcedir=/home/elliott/code/meson/test cases/frameworks/10 gtk-doc' '--builddir=/home/elliott/code/meson/work area' --subdir=doc/foobar1 '--headerdirs=/home/elliott/code/meson/test cases/frameworks/10 gtk-doc/include@@/home/elliott/code/meson/work area/include' --mainfile=foobar-docs.sgml --modulename=foobar --moduleversion= --mode=auto --gtkdoc-scan=/usr/bin/gtkdoc-scan --gtkdoc-scangobj=/usr/bin/gtkdoc-scangobj --gtkdoc-mkdb=/usr/bin/gtkdoc-mkdb --gtkdoc-mkhtml=/usr/bin/gtkdoc-mkhtml --gtkdoc-fixxref=/usr/bin/gtkdoc-fixxref '--content-files=/home/elliott/code/meson/work area/include/bar.xml@@/home/elliott/code/meson/work area/doc/version.xml' '--cc=ccache cc' '--ld=ccache cc'
Building documentation for foobar
Traceback (most recent call last):
[...]

PR #9539, opened 2021-11-06
modules/gnome: check for dependencies before running yelphelper

### LINKED ISSUE #7653: Doesn't check for required itstool for gnome.yelp
Using meson-0.55.1-1.fc34 trying to build malcontent, which uses gnome.yelp would fail with:
```
  File "/usr/bin/meson", line 33, in <module>
    sys.exit(load_entry_point('meson==0.55.1', 'console_scripts', 'meson')())
  File "/usr/lib/python3.9/site-packages/mesonbuild/mesonmain.py", line 228, in main
  File "/usr/lib/python3.9/site-packages/mesonbuild/mesonmain.py", line 217, in run
  File "/usr/lib/python3.9/site-packages/mesonbuild/mesonmain.py", line 165, in run_script_command
  File "/usr/lib/python3.9/site-packages/mesonbuild/scripts/yelphelper.py", line 130, in run
    merge_translations(build_subdir, abs_sources, langs)
  File "/usr/lib/python3.9/site-packages/mesonbuild/scripts/yelphelper.py", line 57, in merge_translations
    subprocess.call([
  File "/usr/lib64/python3.9/subprocess.py", line 349, in call
  File "/usr/lib64/python3.9/subprocess.py", line 947, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib64/python3.9/subprocess.py", line 1819, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'itstool'
```
if `itstool` is missing.

pabloyoyoista:
The original reason for this PR is that I was trying to minimize the set of build dependencies for a project in alpine linux and was surprised that the programs required for `gnome.yelp` were `itstool` and `gettext` instead of `yelp`. It took me a while reading the helper code and understanding how `yelp`, `itstool` and `gettext` interact to actually realize that `gnome.yelp` is building for `yelp`, but using the other tools to build the translations. Therefore, I thought that changing the wording and mentioning the tools used in the documentation could help others in the future. Additionally, as there exists an open ticket,  I added some checks to fail early if `gettext` or `itstool` are missing.

Closes #7653

### DIFF
--- a/docs/markdown/Gnome-module.md
+++ b/docs/markdown/Gnome-module.md
@@ -303,8 +303,8 @@ VAPI or Vala binaries.
              languages: []string, symlink_media: bool = true): void
 ```
 
-Installs help documentation using Yelp. The first argument is the
-project id.
+Installs help documentation for Yelp using itstool and gettext. The first
+argument is the project id.
 
 Additionally, sources can be passed as additional positional arguments. This
 was, however, undocumented and never officially supported. Due to a longstanding
--- a/mesonbuild/modules/gnome.py
+++ b/mesonbuild/modules/gnome.py
@@ -18,6 +18,7 @@ functionality such as gobject-introspection, gresources and gtk-doc'''
 import copy
 import functools
 import os
+import shutil
 import subprocess
 import textwrap
 import typing as T
@@ -1004,6 +1005,13 @@ class GnomeModule(ExtensionModule):
                              'Use a LINGUAS file in the source directory instead')
     @typed_pos_args('gnome.yelp', str, varargs=str)
     def yelp(self, state: 'ModuleState', args: T.Tuple[str, T.List[str]], kwargs) -> ModuleReturnValue:
+        if not shutil.which('itstool'):
+            raise MesonException('itstool not found, required for gnome.yelp target')
+        if not shutil.which('msgmerge'):
+            raise MesonException('gettext not found, required for gnome.yelp target')
+        if not shutil.which('msgfmt'):
+            raise MesonException('gettext not found, required for gnome.yelp target')
+
         project_id = args[0]
         if len(args) > 1:
             FeatureDeprecated.single_use('gnome.yelp more than one positional argument', '0.60.0', 'use the "sources" keyword argument instead.')


eli-schwartz:
This seems like a useful improvement. The only thing I can really object against, is that it doesn't do *enough* -- because really, we shouldn't be offloading all this work to external tools that `build.ninja` doesn't know about. See for example what I did in #9074

Of course, pointing out the i18n module reminds me of something superbly unsatisfying that meson does there:
https://github.com/mesonbuild/meson/blob/5af556e2b85ce264bb7742261e68ef4347b316aa/mesonbuild/modules/i18n.py#L120-L122

Needless to say, if we fail to find gettext in the gnome.yelp() method, we don't do this... thing. That was probably an oversight, since it results in inconsistency... but I don't want to fix the inconsistency, lol, unless that is to make i18n return errors instead of skipping targets.

test cases/frameworks/13 yelp/meson.build
```diff
@@ -1,8 +1,5 @@
 project('yelp', 'c')
 
 itstool = find_program('itstool', required: false)
-if not itstool.found()
-  error('MESON_SKIP_TEST itstool not found.')
-endif
```
eli-schwartz, line 6: This should not be removed, I think, because its purpose is to skip this test when itstool is not installed, not error out.

If we don't skip it then before your PR `ninja` would error out when building, and after your PR, `meson setup` will fail when configuring, but both result in testsuite errors.

dcbaker:
Additionally for find_program, it makes overrides from the machine files or meson.override_find_program. that really should work

test cases/frameworks/13 yelp/meson.build
pablofsf: Added this back, sorry for my ignorance. I thought removing this would be a good way of testing that the bug fix works.

pablofsf:
> This seems like a useful improvement. The only thing I can really object against, is that it doesn't do _enough_ 

That makes sense. The problem I have is that I am really not good with python and a lot of the language features just look like black magic to me. I already tried working around #5793 twice and I never even managed to get anywhere. That being said, I would be happy to do an attempt to remove (or trim) the `yelphelper` script following the issue you point to. However, I would likely need some support and some patience for the possible stupid mistakes I make on the way. If you are able to provide that, then during next week I could possibly do a MR that replaces this one :)


> Of course, pointing out the i18n module reminds me of something superbly unsatisfying that meson does there:
> 
> https://github.com/mesonbuild/meson/blob/5af556e2b85ce264bb7742261e68ef4347b316aa/mesonbuild/modules/i18n.py#L120-L122
> 
> Needless to say, if we fail to find gettext in the gnome.yelp() method, we don't do this... thing. That was probably an oversight, since it results in inconsistency... but I don't want to fix the inconsistency, lol, unless that is to make i18n return errors instead of skipping targets.

From my software packager perspective this is a huge problem. Ideally everyone should be looking at the build log in detail when they package any software, but my experience is more that of: if it builds, it's good to go. Throwing an error instead of a warning would be much more helpful, as the dependency would not get be overseen :)

eli-schwartz:
I would be happy to mentor you through the process. Feel free to join the #mesonbuild IRC channel on the OFTC network.

> From my software packager perspective this is a huge problem.

I have the same perspective and the same conclusion too. :D

pablofsf:
> I would be happy to mentor you through the process. Feel free to join the #mesonbuild IRC channel on the OFTC network.

Ok wonderful! You'll hear back from me next week then :)

jpakkane:
:crossed_swords: conflicts.

Also, if it seems like getting the extra functionality done is taking too much time, then it is fine to just merge this as is. It improves things already and perfect is the enemy of good and all that.

pablofsf:
If you are interested in having this merged fast, then I could fix the conflicts, no problem. But the problem tackled here should be fixed much more elegantly in #9557. I'll be working again on it this week to lead with the symlink support, although I would anticipate the whole process to still take a couple of weeks. I guess it  is up to you if you prefer this first merged and then reverted  or for the bug to stay there for a little longer and get fixed properly only once with less noise :)

PR #9546, opened 2021-11-08
Recursively look for rpath dependencies

### LINKED ISSUE #9542: Meson 0.60 missing RPATH for ExternalLibrary after static_library dependency propagation
**Describe the bug**
Build targets linking a `static_library` that itself depends on an external library found by `compiler.find_library(dirs: ...)` are missing the requisite RPATH arguments in the final link line. If the library provides an `SONAME` the resulting binaries will be unusable (without setting `LD_LIBRARY_PATH`).

This error is observable in 0.60.0 and the current master (5af556e2b85ce264bb7742261e68ef4347b316aa). It does not appear in 0.59.4.

**To Reproduce**
```meson
# meson.build
project('test', 'c')
foo_lib = meson.get_compiler('c').find_library('foo', dirs: meson.project_source_root())
main_lib = static_library('main', 'main.c', dependencies: foo_lib)
executable('main', link_with: main_lib)
```
```c
// foo.c
#include <stdio.h>
void foo() { printf("Foo\n"); }

// main.c
extern void foo();
int main() { foo(); return 0; }
```
```console
$ cc foo.c -o libfoo.so -shared -fPIC -Wl,-soname,libfoo.so
$ meson _builddir
The Meson build system
Version: 0.60.99
Source dir: /tmp/zzz
Build dir: /tmp/zzz/_builddir
Build type: native build
Project name: test
Project version: undefined
C compiler for the host machine: ccache cc (gcc 11.2.0 "cc (Debian 11.2.0-10) 11.2.0")
C linker for the host machine: cc ld.bfd 2.37
Host machine cpu: x86_64
Library foo found: YES
Build targets in project: 2

Found ninja-1.10.1 at /usr/bin/ninja
$ meson compile -C _builddir -v
ninja: Entering directory `/tmp/zzz/_builddir'
[1/3] ccache cc -Ilibmain.a.p -I. -I.. -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -g -fPIC -MD -MQ libmain.a.p/main.c.o -MF libmain.a.p/main.c.o.d -o libmain.a.p/main.c.o -c ../main.c
[2/3] rm -f libmain.a && gcc-ar csrDT libmain.a libmain.a.p/main.c.o
[3/3] cc  -o main  -Wl,--as-needed -Wl,--no-undefined -Wl,--start-group libmain.a /tmp/zzz/libfoo.so -Wl,--end-group
$ ./_builddir/main
./_builddir/main: error while loading shared libraries: libfoo.so: cannot open shared object file: No such file or directory
```

**Expected behavior**
RPATH arguments for the external library's directory would be present in the final link, as they are in 0.59.4:
```console
$ meson _builddir
The Meson build system
Version: 0.59.4
...
[3/3] cc  -o main  -Wl,--as-needed -Wl,--no-undefined '-Wl,-rpath,$ORIGIN/..' -Wl,-rpath-link,/tmp/zzz/_builddir/../ -Wl,--start-group libmain.a /tmp/zzz/libfoo.so -Wl,--end-group
$ ./_builddir/main
Foo
```

**system parameters**
* what operating system (e.g. MacOS Catalina, Windows 10, CentOS 8.0, Ubuntu 18.04, etc.): Debian Unstable/Sid
* what Python version are you using e.g. 3.8.0: Python 3.9.7
* what `meson --version`: Meson 0.60.99
* what `ninja --version` if it's a Ninja build: Ninja 1.10.1

dcbaker:
In 0.60 a bug was fixed where a cached variable was being mutated. Unfortunately, we were relying on that mutation to get the behavior of recursively looking up the dependency web of a target. This series does a bit of cleanup, then fixes the bug by explicitly doing the up chain lookup.

Fixes #9542

### DIFF
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -768,7 +768,7 @@ class Backend:
     def determine_rpath_dirs(self, target: build.BuildTarget) -> T.Tuple[str, ...]:
         result: OrderedSet[str]
         if self.environment.coredata.get_option(OptionKey('layout')) == 'mirror':
-            # NEed a copy here
+            # Need a copy here
             result = OrderedSet(target.get_link_dep_subdirs())
         else:
             result = OrderedSet()
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -395,15 +395,15 @@ class Backend:
             return os.path.join(self.build_to_src, target_dir)
         return self.build_to_src
 
-    def get_target_private_dir(self, target: build.Target) -> str:
+    def get_target_private_dir(self, target: T.Union[build.BuildTarget, build.CustomTarget, build.CustomTargetIndex]) -> str:
         return os.path.join(self.get_target_filename(target, warn_multi_output=False) + '.p')
 
-    def get_target_private_dir_abs(self, target: build.Target) -> str:
+    def get_target_private_dir_abs(self, target: T.Union[build.BuildTarget, build.CustomTarget, build.CustomTargetIndex]) -> str:
         return os.path.join(self.environment.get_build_dir(), self.get_target_private_dir(target))
 
     @lru_cache(maxsize=None)
     def get_target_generated_dir(
-            self, target: build.Target,
+            self, target: T.Union[build.BuildTarget, build.CustomTarget, build.CustomTargetIndex],
             gensrc: T.Union[build.CustomTarget, build.CustomTargetIndex, build.GeneratedList],
             src: str) -> str:
         """
@@ -418,7 +418,8 @@ class Backend:
         # target that the GeneratedList is used in
         return os.path.join(self.get_target_private_dir(target), src)
 
-    def get_unity_source_file(self, target: build.Target, suffix: str, number: int) -> mesonlib.File:
+    def get_unity_source_file(self, target: T.Union[build.BuildTarget, build.CustomTarget, build.CustomTargetIndex],
+                              suffix: str, number: int) -> mesonlib.File:
         # There is a potential conflict here, but it is unlikely that
         # anyone both enables unity builds and has a file called foo-unity.cpp.
         osrc = f'{target.name}-unity{number}.{suffix}'
@@ -765,7 +766,9 @@ class Backend:
                 paths.append(libdir)
         return paths
 
-    def determine_rpath_dirs(self, target: build.BuildTarget) -> T.Tuple[str, ...]:
+    # This may take other types
+    def determine_rpath_dirs(self, target: T.Union[build.BuildTarget, build.CustomTarget, build.CustomTargetIndex]
+                             ) -> T.Tuple[str, ...]:
         result: OrderedSet[str]
         if self.environment.coredata.get_option(OptionKey('layout')) == 'mirror':
             # Need a copy here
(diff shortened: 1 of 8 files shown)


unittests/allplatformstests.py
```diff
@@ -1499,19 +1499,19 @@ def test_prebuilt_shared_lib(self):
...
-        # Run the test
-        try:
-            self.init(tdir)
-            self.build()
-            self.run_tests()
-        finally:
-            os.unlink(shlibfile)
-            if is_windows():
-                # Clean up all the garbage MSVC writes in the
-                # source tree.
```
eli-schwartz, line 1511: This comment is probably safe to re-add.

eli-schwartz (commented):
I wonder if we could add a test for this?

jpakkane:
Test failures seem relevant:

```
__________________ AllPlatformTests.test_prebuilt_shared_lib __________________
[gw0] win32 -- Python 3.7.9 C:\hostedtoolcache\windows\Python\3.7.9\x64\python.exe

self = <unittest.case._Outcome object at 0x00000257D977AB88>
test_case = <unittests.allplatformstests.AllPlatformTests testMethod=test_prebuilt_shared_lib>
isTest = False

    @contextlib.contextmanager
    def testPartExecutor(self, test_case, isTest=False):
>               function(*args, **kwargs)
E               FileNotFoundError: [WinError 2] The system cannot find the file specified: 'D:\\a\\1\\s\\test cases/unit\\17 prebuilt shared\\alexandria.dll'

C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\unittest\case.py:669: FileNotFoundError
```

mesonbuild/backend/backends.py
```diff
@@ -763,9 +764,14 @@ def rpaths_for_bundled_shared_libraries(self, target: build.BuildTarget, exclude
                 paths.append(os.path.join(self.build_to_src, rel_to_src))
             else:
                 paths.append(libdir)
+        for i in chain(target.link_targets, target.link_whole_targets):
+            if isinstance(i, build.BuildTarget):
+                paths.extend(self.rpaths_for_bundled_shared_libraries(i, exclude_system))
```
nirbheek, line 769: Looks like this hunk should be in the next commit?

mesonbuild/backend/backends.py
```diff
@@ -779,8 +779,9 @@ def determine_rpath_dirs(self, target: T.Union[build.BuildTarget, build.CustomTa
         else:
             result = OrderedSet()
             result.add('meson-out')
-        result.update(self.rpaths_for_bundled_shared_libraries(target))
-        target.rpath_dirs_to_remove.update([d.encode('utf-8') for d in result])
+        if isinstance(target, build.BuildTarget):
+            result.update(self.rpaths_for_bundled_shared_libraries(target))
+            target.rpath_dirs_to_remove.update([d.encode('utf-8') for d in result])
```
nirbheek, line 784: And this hunk should be in the previous commit?

mesonbuild/backend/backends.py
dcbaker: yup, you're right, i split these wrong. New version incoming.

nirbheek:
I dove into this code recently for a different PR, so I can endorse these changes. LGTM!

PR #9547, opened 2021-11-08
test cases: enable cmake llvm on msys2

dcbaker:
This seems to work now, though we don't expect it to. I don't know why
it works, I'm assuming a change to msys2 itself, and not something Meson
did.


eli-schwartz:
You deleted the line entirely, so now we don't run that part of the matrix at all.

@lazka tried to fix this in #9534 but it was still skipping on the combination of cmake and link-static=true

PR #9556, opened 2021-11-10
mtest: add an ASCII-only version of the spinner

bonzini:
While the horizontal line and the other pictograms in mtest have an ASCII-only
version, the spinner does not.  This causes mtest to fail with a
UnicodeEncodeError exception on non-Unicode locales.


anarazel:
Thanks!

Issue #9105, opened 2021-08-12
"Cannot open XML file" with gnome.yelp targets


rmnvgr:
**Describe the bug**

When using the targets from [`gnome.yelp`](https://mesonbuild.com/Gnome-module.html#gnomeyelp), it only gives this error:

```
[0/1] Running external command help-project-pot (wrapped by meson to set env)
Error: cannot open XML file C/index.page
```

I checked that the file is at the right path, and that the permissions are correct. It not only happens on my project but also on other projects using `gnome.yelp`. I also tried in a clean VM to make sure it was not a problem with my setup. If I manually run `itstool`, it works fine.

**To Reproduce**

`meson.build`:

```meson
project('test')

import('gnome').yelp(
  'test',
  sources: ['index.page']
)
```

`C/index.page`:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<page xmlns="http://projectmallard.org/1.0/" xmlns:its="http://www.w3.org/2005/11/its" type="guide">
	<title>Test</title>
</page>
```

Empty `LINGUAS` file.

Run `meson build && meson compile -C build help-test-pot`.

**Expected behavior**

The targets execute correctly and generate the translation files.

**System parameters**

* OS: Fedora 34
* Python version: 3.9.6
* `meson --version`: 0.59
* `ninja --version`: 1.10.2

eli-schwartz:
This is reproducible in `test cases/frameworks/13 yelp/`, and I can verify that @pabloyoyoista's PR fixes it.

Issue #1601, opened 2017-04-10
failed installed with i18n.gettext are not removed on uninstall


keszybz:
After `ninja-build install`, `ninja-build uninstall` I see 
```
$ find /var/tmp/inst7 -type f
/var/tmp/inst7/usr/share/locale/be/LC_MESSAGES/systemd.mo
/var/tmp/inst7/usr/share/locale/be@latin/LC_MESSAGES/systemd.mo
...
/var/tmp/inst7/usr/share/locale/zh_TW/LC_MESSAGES/systemd.mo
```
All other installed files seem to be removed, so it seems only `i18n` does not support that.

inigomartinez:
Files installed by `gnome.yelp` aren't also installed. Neither files installed by `i18n.gettext` nor `gnome.yelp` are present in the `meson-logs/install-log.txt` file.

TingPing:
Yes any custom installation provided by modules is not currently tracked, we need to add some helper function to do so.

raveit65:
I like to push the report again.
We see that problem with several packages, currently we are switching MATE desktop to use meson build system.
https://github.com/mate-desktop/mate-backgrounds/pull/22#issuecomment-489966620
and
https://github.com/mate-desktop/mate-desktop/pull/333#issuecomment-483708111
Would be nice if this can be fixed.
Thanks.

ericonr:
@TingPing what sort of help is necessary for this to happen? I can try to add the feature.

dcbaker:
There's several lists in the `Build` class (`mesonbuild/build.py`), which is in the `ModuleState`, `man`, `data`, and `headers` which modules need to update. There's code in the `mesonbuild/interpreter.py` that does a bunch of the front end parsing and updates these. I'm not sure exactly what it would look like, but that's where I'd start.

andy5995:
Anyone have suggestions as to user-friendly ways to uninstall the mo files? I was thinking of generating an `uninstall_langs` shell script, based on an `.in` file that contains variables which get replaced by the paths to the mo files.

andy5995:
> Anyone have suggestions as to user-friendly ways to uninstall the mo files? I was thinking of generating an `uninstall_langs` shell script, based on an `.in` file that contains variables which get replaced by the paths to the mo files.

```meson
if get_option('nls')
  ul = configuration_data()
  ul.set('LOCALEDIR', localedir)
  ul.set('MESON_SOURCE_ROOT', meson.source_root())
[...]

PR #9557, opened 2021-11-10
modules/gnome: replace yelphelper with custom_targets

### LINKED ISSUE #1052: 0.36.0 test regression test cases/frameworks/13 yelp
Hi, after trying to package the new 0.36.0 version i have noticed a test regression in ```test cases/frameworks/13 yelp```. I have used ```MESON_PRINT_TEST_OUTPUT=1``` but not quote sure what the problem is:

```
Failed test: test cases/frameworks/13 yelp
Reason: Running install failed.
The Meson build system
Version: 0.36.0
Source dir: /build/meson/src/meson-0.36.0/test cases/frameworks/13 yelp
Build dir: /build/meson/src/meson-0.36.0/b 9cu54r21
Build type: native build
Project name: yelp
Native c compiler: cc (gcc 6.2.1)
Appending CFLAGS from environment: '-march=i686 -mtune=generic -O2 -pipe -fstack-protector-strong'
Appending LDFLAGS from environment: '-Wl,-O1,--sort-common,--as-needed,-z,relro'
Appending CPPFLAGS from environment: '-D_FORTIFY_SOURCE=2'
Build machine cpu family: x86
Build machine cpu: i686
Build targets in project: 2
ninja: no work to do.
No tests defined.
No benchmarks defined.
[0/1] Installing files.
Running custom install script /usr/bin/python3
Failed to run install script: /usr/bin/python3 /build/meson/src/meson-0.36.0/meson.py --internal yelphelper install --subdir=help --id=meson --installdir=share/help --sources=index.page --media=media/test.txt --langs=de@@es
FAILED: install 
'/usr/bin/python3' '/build/meson/src/meson-0.36.0/meson.py' '--internal' 'install' '/build/meson/src/meson-0.36.0/b 9cu54r21/meson-private/install.dat'
```

### LINKED ISSUE #2775: gnome.yelp() doesn't compile help pages at compile time
Currently, `gnome.yelp()` does not build Yelp help pages at compile time, instead it both builds and installs the help pages using (what I assume is) an install-time script hook.

Geary has historically built its help pages as part of the build process so we know it's working, to be able to proofread the manual using the app from the source directory, to catch tool errors, and so on. It would be good to be able to continue to do this.

PR to follow.

### LINKED ISSUE #6916: Internal yelp helper should error instead of warn
meson 0.54.0

[gnome-desktop](https://gitlab.gnome.org/GNOME/gnome-desktop/-/merge_requests/68) recently had a broken translation which is supposed to be merged using yelp.

This is the warning it produced:
```
Running custom install script '/home/hadess/Projects/gnome-install/bin/meson --internal yelphelper install --subdir=desktop-docs/lgpl --id=lgpl --installdir=share/help --sources=index.docbook --symlinks=true'
Warning: Could not merge gl translation for msgid:
You must give prominent notice with each copy of the work that the Library is used in it and that the Library and its use are covered by this License. You must supply a copy of this License. If the work during execution displays copyright notices, you must include the copyright notice for the Library among them, as well as a reference directing the user to the copy of this License. Also, you must do one of these things: <_:orderedlist-1/>
```

The fact that it doesn't error out is a problem for CIs as the warnings won't be reported unless there's a manual check. I think meson should error out in those cases, or at least make it possible to error out instead of skipping broken translations.

### LINKED ISSUE #7034: gnome.yelp symlinks are absolute rather than relative to the prefix which breaks 'ninja dist'
**Describe the bug**
When doing 'ninja dist' the check stage tries to change the permissions of each installed file, doing so on a symlink for each locale of the gnome.yelp help output. The installed prefix is whithn the build/ directory but the symlink points to /usr/local so the chmod operation throws an exception.

**To Reproduce**
Run ninja dist on a module that outputs docs using gnome.yelp

**Expected behavior**
For the dist to succeed and symlinks to be relative to the install prefix or the docs directory even

**system parameters**
This is fedora running on x86_64 doing a standard meson build && ninja -C build/ dist
```
[1030][aruiz@localhost ~]$ meson --version
0.54.0
[1031][aruiz@localhost ~]$ python3 --version
Python 3.8.2
[1032][aruiz@localhost ~]$ ninja --version
1.10.0
[1033][aruiz@localhost ~]$ yelp --version
yelp 3.36.0
```

### LINKED ISSUE #7653: Doesn't check for required itstool for gnome.yelp

pabloyoyoista:
This is basically a rewrite of the `gnome.yelp` target to remove the ad-hoc script, which has many issues, as mentioned in #9539. 

Closes #7653
Closes #9539
Closes #6916 (I believe so, but I do not have a reproducer)
Closes #2775
Closes #7034
Closes #1052

Related #9105
Related #1601 

TODO:
 - [x] Install C files
 - [x] Install media files
 - [x] Take care of `symlink_media` option

@eli-schwartz 

I am really not that good with meson, so there might be a lot of issues in this PR. Things that I struggled with:
- The targets are basically all dependent on each other in a way that the output from one target is the input of the next one. I have solved this by using `depfile` instead of `input` in `gmotarget` and `mergetarget` and by reading the `.pot` file with `mesonlib.File` in `potarget`. I am not sure at all this is the right way to do it.
- In the original help script, the `update-po` now equivalent to `potarget` would execute both `pottarget` and `potarget`. I thought that the added `depends` would take care of that. Is it enough the way it is done now?
- Target naming is a bit of a mess. But I wasn't sure if changing a lot the naming would make sense or not.
- Reusing code from `i18n.py` might be nice, but I have no clue how to do it.

Additionally, I really do not know how to solve the TODOs marked above:
- For the C files, they can be directly installed from the inputs folder. How does one take care of that? Is there some copy target?
- The media files have the problem that they have to be copied to each language subdirectory (or symlinked). Similar to the previous one, is there some simple way to take care of this? I really could not find one :(


eli-schwartz (commented):
Thanks for taking a look at this! Some quick preliminary comments below.

mesonbuild/modules/gnome.py
```diff
@@ -1016,55 +1017,73 @@ def yelp(self, state: 'ModuleState', args: T.Tuple[str, T.List[str]], kwargs) ->
...
-        source_str = '@@'.join(sources)
+        sources_files = [mesonlib.File.from_source_file(state.environment.source_dir,
+                                                        os.path.join(state.subdir, 'C'),
+                                                        s) for s in sources]
 
         langs = mesonlib.stringlistify(kwargs.pop('languages', []))
+        if not langs:
+            langs = read_linguas(state.subdir)
+        # I am currently ignoring media and symlinks
+        # How could they be added?
```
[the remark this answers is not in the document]
pablofsf: Ok, I'll look into this and see if I have the skills to tackle it. If I feel I could pull it through, then I would rather take care of it now. That it will definitely delay the landing of this PR, but also save work on the long-term. Given how long this helper has been around, I guess there's no real rush to pull this is only half-way cleaned?

mesonbuild/modules/gnome.py
```diff
@@ -1016,55 +1017,73 @@ def yelp(self, state: 'ModuleState', args: T.Tuple[str, T.List[str]], kwargs) ->
...
+        pot_kwargs = {'command': ['itstool', '-o', '@OUTPUT@', '@INPUT@'],
+                      'input': sources,
+                      'output': pot_name,
+        }
+        pottarget = build.CustomTarget('help-' + project_id + '-pot',
+                                       os.path.join(state.subdir, 'C'),
+                                       state.subproject, pot_kwargs)
+        targets.append(pottarget)
+        pot_file = mesonlib.File.from_built_file(os.path.join(state.subdir, 'C'),
+                                                 pot_name)
```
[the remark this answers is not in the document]
pablofsf: Mmm, yes, I saw that in your changes and also something similar in `gtkdochelper.py`, but I couldn't really understand why. Now you mentioning a ".PHONY" target, I understand that, as there exists no `.pot` file to begin with, this is not a rule associated with a file and should be ".PHONY". However, some questions:
- Is there some way to limit the execution of this target to only when the `input` files change? It is true the `.pot` file is not a source file, but what's the point of rebuilding documentation every time if the documentation files did not change? I think this might be a clear difference between the `i18n.gettext` and `gnome.yelp`, as the former defines the source files, whereas the first one basically looks at the whole tree (unless I'm getting something wrong...)
- Which is the recommended way to handle temporary non-source files like the `.pot` and `.po`? I honestly very much liked that the targets are more clear, short and readable than the helper script :)

mesonbuild/modules/gnome.py
```diff
@@ -1016,55 +1017,73 @@ def yelp(self, state: 'ModuleState', args: T.Tuple[str, T.List[str]], kwargs) ->
...
+                          'depends': potarget,
+            }
+            gmotarget = build.CustomTarget('help-' + project_id + '-' + l + '-gmo',
+                                           l_subdir, state.subproject, gmo_kwargs)
+            targets.append(gmotarget)
+
+            merge_kwargs = {'command': ['itstool', '-m', '@DEPFILE@', '-o', '@OUTDIR@', '@INPUT@'],
+                            'input': sources_files,
+                            'output': sources,
+                            'depfile': l + '.gmo',
```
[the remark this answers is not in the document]
pablofsf: Thank you for the explanation, it is changed now. Waiting to update it in the po target for the other comments.

mesonbuild/modules/gnome.py
```diff
@@ -1016,55 +1017,73 @@ def yelp(self, state: 'ModuleState', args: T.Tuple[str, T.List[str]], kwargs) ->
...
+            potarget = build.CustomTarget('help-' + project_id + '-' + l + '-update-po',
+                                          l_subdir, state.subproject, po_kwargs)
+            targets.append(potarget)
+
+            gmo_kwargs = {'command': ['msgfmt', '@DEPFILE@', '-o', '@OUTPUT@'],
+                          'depfile': po_file,
+                          'output': l + '.gmo',
+                          'depends': potarget,
+            }
+            gmotarget = build.CustomTarget('help-' + project_id + '-' + l + '-gmo',
```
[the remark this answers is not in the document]
pablofsf: Looks much nicer, thank you!

mesonbuild/modules/gnome.py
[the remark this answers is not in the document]
eli-schwartz: They aren't temporary :D they traditionally get checked into git.

The primary concern here about having them get built by input/output rules is that source files are used to build pot files, and pot files are used to build po files, and po files are used to build mo files.

But, you only want pot/po to be triggered by maintainers, as po files have *two* inputs. The pot file, and the contents pulled from e.g. transifex. And the mo files only use the transifex updates, not the ones merged from the pot file...

It is probably ok to have the pot file generated via a non-phony target, and just not built by default. However, po files should still be recreated every time the maintainer target is run... because we can't know whether the po files were updated from transifex and are more recent than the pot file, yet still need to merge pot updates too.

The pot -> po rules can still use direct gettext utility commands rather than going through a helper script, but they should still be .PHONY rules (RunTarget) I think.

mesonbuild/modules/gnome.py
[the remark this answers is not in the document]
eli-schwartz: A more important consideration is that this should only land for meson 0.61, which isn't imminent, so we have time to polish it before release. :)

mesonbuild/modules/gnome.py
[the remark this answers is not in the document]
pablofsf: First of all, thank you very much for your time and patience, I really appreciate it!

I think I have gotten some concepts totally wrong here, based on my little and biased experience with translatable programs and lack of understanding of previous code. My understanding was that during build we would have:

  `source` + `.po` files in VCS -> generate `.pot` from `source` -> new `.po` files with `.pot` and VCS'ed `.po` -> generate `.mo` (or here `.gmo` + merge) -> install `.mo` (or merged documents here in yelp).
 
It seems I was only considering the packaging process and totally dismissing the maintainer process. Then after your comments, my understanding is:

- **Maintainer** process: (updated) `source` + `.po` files in VCS -> generate `.pot` from `source` -> new `.po` files with `.pot` and (updated) VCS'ed `.po` -> maintainer commits updates.
- **Packaging** process: `source` + `.po` files in VCS -> `.mo` (or here `.gmo` + merge) -> install `.mo` (or merged documents here in yelp)

If my understanding now is correct, it makes total sense to have both the `.pot` and the `.po` targets as RunTargets, as they are not needed for packaging, but might be used by developers when updating translations.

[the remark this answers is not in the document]
pablofsf: Hopefully last push has fixed this. At least I have verified that running `meson compile help-{package_id}-pot` and `meson compile help-{package_id}-update-po` seem to work as expected.

mesonbuild/modules/gnome.py
```diff
@@ -1016,55 +1017,71 @@ def yelp(self, state: 'ModuleState', args: T.Tuple[str, T.List[str]], kwargs) ->
...
+            mergetarget = build.CustomTarget('', l_subdir, state.subproject, merge_kwargs)
+            targets.append(mergetarget)
+
+        allpotarget = build.AliasTarget(f'help-{project_id}-update-po', potargets,
+                                        state.subdir, state.subproject)
+        targets.append(allpotarget)
+
+        c_data = self.interpreter.install_data_impl(sources_files, os.path.join(install_dir, 'C', project_id),
+                                                    mesonlib.FileMode(), [], [])
+        targets.append(c_data)
```
pablofsf, line 1082: I guess I have misunderstood something from your comment, but this is not working (neither if `c_data` is not added to the targets). Any clue of what might I be doing wrong? I've spent a fair amount of time trying to understand why is this not installing the data, but still no clue..,

eli-schwartz: Sorry, I meant to take a look at how to use a `build.Data()` -- you cannot invoke interpreter functions directly from a module, it's complicated.

This works:

```
-        c_data = self.interpreter.install_data_impl(sources_files, os.path.join(install_dir, 'C', project_id),
-                                                    mesonlib.FileMode(), [], [])
+        final_installdir = os.path.join(install_dir, 'C', project_id)
+        c_data = build.Data(sources_files, final_installdir, final_installdir ,mesonlib.FileMode(), state.subproject)
```

pablofsf: Oh, sorry my confusion, that works perfectly!

mesonbuild/modules/gnome.py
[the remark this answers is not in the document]
pablofsf: There is a temptative implementation in #9608. I have rebased this PR on top of it just for testing purposes. But seems to work alright on this scenario :)

pablofsf:
Now that there are symlinks in master, this is ready to be properly reviewed. I have added the functionality that I had missing in different commits, but can squash on request. 

There seems to be some CI failure, as I am getting bombarded by emails complaining that CI could not start because of problems with the definition. So I guess this will need a rebase once CI is fixed.

mesonbuild/modules/gnome.py
```diff
@@ -1055,47 +1056,92 @@ def yelp(self, state: 'ModuleState', args: T.Tuple[str, T.List[str]], kwargs: 'Y
...
+                            mesonlib.FileMode(), state.subproject)
+        targets.append(c_data)
+
+        media_files = []
+        for m in media:
+            f = mesonlib.File.from_source_file(state.environment.source_dir,
+                                               os.path.join(state.subdir, 'C'), m)
+            media_files.append(f)
+            m_install_dir = os.path.join(c_install_dir, os.path.dirname(m))
+            m_data = build.Data(media_files, m_install_dir, m_install_dir,
```
eli-schwartz, line 1084: Why do we build a list of media_files and on each pass, create a new build.Data object with all media_files found so far?

Shouldn't we just use `build.Data(f, ...)` here?

mensinda:
I have no idea why GitHub suddenly has problems here... It seems to still work on master, so maybe try triggering the CI again with a rebase / force push?

The only other thing I could think of is that `github.head_ref` may be broken in this PR. Could you have done some merges, rebases to other branches than `master` that could potentially maybe confuse GitHub?

mesonbuild/modules/gnome.py
pablofsf: Yes, that's a bug, thanks! The list is still need for the case when a certain locale does not have media files and they have to be installed from the "C" locale.

pablofsf:
Now with the latest push it seems to be fixed. Roughly at the same time when I reported this another PR also had the same issue, so maybe it was some temporary GitHub timeout. Thanks for looking into it!

mesonbuild/modules/gnome.py
```diff
@@ -1055,47 +1056,92 @@ def yelp(self, state: 'ModuleState', args: T.Tuple[str, T.List[str]], kwargs: 'Y
...
+            potarget = build.RunTarget(f'help-{project_id}-{l}-update-po',
+                                       po_args, [pottarget], l_subdir, state.subproject)
+            targets.append(potarget)
+            potargets.append(potarget)
+
+            gmo_kwargs = {'command': ['msgfmt', '@INPUT@', '-o', '@OUTPUT@'],
+                          'input': po_file,
+                          'output': l + '.gmo',
+            }
+            gmotarget = build.CustomTarget(f'help-{project_id}-{l}-gmo', l_subdir, state.subproject, gmo_kwargs)
```
pablofsf, line 1126: I am missing something here. The error in the CI for yelp is: `ERROR: Multiple producers for Ninja target "help/de/de.gmo". Please rename your targets.` However, where those are actually generated, there is a unique target name :(

mesonbuild/modules/gnome.py
```diff
@@ -1055,47 +1056,92 @@ def yelp(self, state: 'ModuleState', args: T.Tuple[str, T.List[str]], kwargs: 'Y
...
+
+            merge_kwargs = {'command': ['itstool', '-m', os.path.join(l_subdir, l + '.gmo'),
+                                        '-o', '@OUTDIR@', '@INPUT@'],
+                            'input': sources_files,
+                            'output': sources,
+                            'depends': gmotarget,
+                            'install': True,
+                            'install_dir': l_install_dir,
+            }
+            mergetarget = build.CustomTarget(f'help-{project_id}-{l}', l_subdir, state.subproject, merge_kwargs)
```
pablofsf, line 1137: Also used here, but that is exactly the same case :(

mesonbuild/modules/gnome.py
dcbaker: The name isn't the output, it's the name you can call work ninja $name that had to be unique

mesonbuild/modules/gnome.py
```diff
@@ -1055,47 +1056,92 @@ def yelp(self, state: 'ModuleState', args: T.Tuple[str, T.List[str]], kwargs: 'Y
...
         langs = kwargs['languages']
+        if not langs:
+            langs = read_linguas(state.subdir)
+
+        media = kwargs['media']
+        symlinks = kwargs['symlink_media']
+        targets: T.List['Target'] = []
+        potargets: T.List['build.RunTarget'] = []
+
+        install_dir = kwargs.get('install_dir', os.path.join(state.environment.get_datadir(), 'help'))
```
dcbaker, line 1072: This isn't going to do what you expect, typed_kwargs ensures that there's always a value. You'll need kwargs or fallback

mesonbuild/modules/gnome.py
pablofsf: This is fixed in the last commit. Seems to have been a test-design issue.

mesonbuild/modules/gnome.py
pablofsf: Thank you for this comment. Originally this module hard-coded the install_dir to `os.path.join(state.environment.get_datadir(), 'help')`. I do not know why did I change the behavior without documenting it and I cannot see any reason for kwargs. So I have reverted to the original behavior.

pablofsf:
Finally all related tests passed and all functionality is present. The failure is the unrelated rust problem. Let me know if I should squash the commits.

mesonbuild/modules/gnome.py
```diff
@@ -1055,47 +1056,93 @@ def yelp(self, state: 'ModuleState', args: T.Tuple[str, T.List[str]], kwargs: 'Y
...
+                                                        os.path.join(state.subdir, 'C'),
+                                                        s) for s in sources]
 
         langs = kwargs['languages']
+        if not langs:
+            langs = read_linguas(os.path.join(state.environment.source_dir, state.subdir))
+
+        media = kwargs['media']
+        symlinks = kwargs['symlink_media']
+        targets: T.List['Target'] = []
```
dcbaker, line 1069: This isn't right, you put a Data in here just a little down the way, at the very least this should be `T.List[T.Union['Target', build.Data]]`

mesonbuild/modules/gnome.py
```diff
@@ -1055,47 +1056,93 @@ def yelp(self, state: 'ModuleState', args: T.Tuple[str, T.List[str]], kwargs: 'Y
...
+        potargets: T.List['build.RunTarget'] = []
+
+        install_dir = os.path.join(state.environment.get_datadir(), 'help')
+        c_install_dir = os.path.join(install_dir, 'C', project_id)
+        c_data = build.Data(sources_files, c_install_dir, c_install_dir,
+                            mesonlib.FileMode(), state.subproject)
+        targets.append(c_data)
+
+        media_files = []
```
dcbaker, line 1078:
```suggestion
        media_files: T.List[mesonlib.File] = []
```

mesonbuild/modules/gnome.py
```diff
@@ -1055,47 +1056,93 @@ def yelp(self, state: 'ModuleState', args: T.Tuple[str, T.List[str]], kwargs: 'Y
...
+            f = mesonlib.File.from_source_file(state.environment.source_dir,
+                                               os.path.join(state.subdir, 'C'), m)
+            media_files.append(f)
+            m_install_dir = os.path.join(c_install_dir, os.path.dirname(m))
+            m_data = build.Data([f], m_install_dir, m_install_dir,
+                                mesonlib.FileMode(), state.subproject)
+            targets.append(m_data)
+
+        pot_file = os.path.join('@SOURCE_ROOT@', state.subdir, 'C', project_id + '.pot')
+        pot_args = ['itstool', '-o', pot_file] + [os.path.join('@SOURCE_ROOT@', state.subdir, 'C', s) for s in sources]
```
dcbaker, line 1089: I think we should be using `state.find_program(`itstool')` here, not just a raw string. Among other things, that makes overriding the program in the machine file work

mesonbuild/modules/gnome.py
```diff
@@ -1055,47 +1056,93 @@ def yelp(self, state: 'ModuleState', args: T.Tuple[str, T.List[str]], kwargs: 'Y
...
+                    try:
+                        m_file = mesonlib.File.from_source_file(state.environment.source_dir, l_subdir, m)
+                    except MesonException:
+                        m_file = media_files[i]
+                    l_data = build.Data([m_file], m_install_dir, m_install_dir,
+                                        mesonlib.FileMode(), state.subproject)
+                targets.append(l_data)
+
+            po_file = l + '.po'
+            po_args = ['msgmerge', '-q', '-o', os.path.join('@SOURCE_ROOT@', l_subdir, po_file),
```
dcbaker, line 1115: same here with `state.find_program`

dcbaker (changes requested):
I see a few things looking through this, but otherwise I think this looks good and I really like this change.

mesonbuild/modules/gnome.py
pablofsf: Thanks, I also added `build.SymlinkData`

mesonbuild/modules/gnome.py
pablofsf: Yes, fixed. Actually, this was originally the reason for me implementing this functionality :)

I have only used it for the `RunTarget`s. Is it also needed in the `command` arg or `CustomTarget`s? Or will the logic in `CustomTarget` take care of it?

pablofsf: I got that that is the case in matrix. So I also added them for CustomTargets.

dcbaker (approved):
looks great, thanks!

eli-schwartz:
0.61 rc1 is scheduled for tomorrow and I realized that this was still pending.

I've just quickly rebased on master and re-ordered the commits, because the test cleanup should come first I think (for bisectability).

pabloyoyoista:
Thanks a lot for looking into it! I assumed it was just waiting merge :)

eli-schwartz:
I am pretty sure this was looking quite good several weeks ago :) merged.

PR #9559, opened 2021-11-10
Share common code between clang tidy and format

xclaesse:


xclaesse:
This adds support for .clang-tidy-include and .clang-tidy-exclude files just like for clang-format.

PR #9560, opened 2021-11-11
Remove incorrect arguments for C2000 C++ compiler.

nathanaelg:
This comes from a similar place to https://github.com/mesonbuild/meson/pull/9419 but is a bit more limited in scope. 

The original C2000 C++ support was never tested and possibly copied from the CCRX implementation. It actually just needs to be very similar to the C implementation (with the exception of compiler standards where c++03 is the only supported type).

### DIFF
--- a/mesonbuild/compilers/cpp.py
+++ b/mesonbuild/compilers/cpp.py
@@ -838,20 +838,21 @@ class C2000CPPCompiler(C2000Compiler, CPPCompiler):
         opts[key].choices = ['none', 'c++03']
         return opts
 
-    def get_always_args(self) -> T.List[str]:
-        return ['-nologo', '-lang=cpp']
-
     def get_option_compile_args(self, options: 'KeyedOptionDictType') -> T.List[str]:
-        return []
+        args = []
+        key = OptionKey('std', machine=self.for_machine, lang=self.language)
+        std = options[key]
+        if std.value != 'none':
+            args.append('--' + std.value)
+        return args
 
-    def get_compile_only_args(self) -> T.List[str]:
-        return []
+    def get_no_optimization_args(self) -> T.List[str]:
+        return ['-Ooff']
 
     def get_output_args(self, target: str) -> T.List[str]:
-        return [f'-output=obj={target}']
+        return [f'--output_file={target}']
 
-    def get_option_link_args(self, options: 'KeyedOptionDictType') -> T.List[str]:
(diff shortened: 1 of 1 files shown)


mesonbuild/compilers/cpp.py
```diff
@@ -824,8 +824,9 @@ def get_compiler_check_args(self, mode: CompileCheckMode) -> T.List[str]:
         return []
 
 class C2000CPPCompiler(C2000Compiler, CPPCompiler):
-    def __init__(self, exelist: T.List[str], version: str, for_machine: MachineChoice, is_cross: bool,
-                 info: 'MachineInfo', exe_wrapper: T.Optional['ExternalProgram'] = None,
+    def __init__(self, exelist: T.List[str], version: str, for_machine: MachineChoice,
+                 is_cross: bool, info: 'MachineInfo',
+                 exe_wrapper: T.Optional['ExternalProgram'] = None,
```
eli-schwartz, line 829: Why is this just being reflowed without changes?

nathanaelg: Good point, both the inherited `__init__` methods will be called anyway won't they?

nathanaelg: I think it is basically to set exe_wrapper, linker and full_version to None?

nathanaelg: Without this I end up with `TypeError: __init__() got an unexpected keyword argument 'full_version'`

eli-schwartz: The `__init__` method belongs there.

My question is why is this formatting change included in the PR, you are not adding any code changes to this init method?

nathanaelg: Ah I see, I think I was comparing with the `class C2000CCompiler(C2000Compiler, CCompiler):` which had it formatted slightly differently. I will revert the whitespace change.

eli-schwartz: Thanks! The diff is much clearer now.

nathanaelg:
I can't see much in the way of detail of the test failure (or if it is related to this change) but let me know if I can look at anything.

eli-schwartz:
msys2 is known to have an unexpected-run in git master for statically linked llvm + cmake. Don't worry about it.

dcbaker (approved):

PR #9561, opened 2021-11-11
hdf5 dependency: check for alternative config-tool names

### LINKED ISSUE #9555: Require parallel tools when depending on HDF5
I have an application that specifically requires the availability of `h5pcc` instead of `h5cc`: https://github.com/mesonbuild/meson/blob/39ceb677744e1f959a0db6406d51a26d78f0ab68/mesonbuild/dependencies/hdf5.py#L97-L99

In my case HDF5 is only build with parallel compatibility and therefore `h5cc` is not available. 
My current workaround is to simple edit this line and specify the other tool directly. 

It there a more appropriate workaround for this issue?
Alternatively: Is it possible to add an optional parameter to `hdf5_dep = dependency('hdf5')` e.g. `dependency('hdf5', parallel: true)`

eli-schwartz:
Depending on whether hdf5 is compiled with parallel support, the same config-tool program may be installed with a mysterious "p" in the name. In this case, dependency lookup will totally fail, unless of course you use the superior pkg-config interface in which case you get a predictable name.

Work around this insanity by checking for both types of config-tool name.

Fixes #9555

### DIFF
--- a/mesonbuild/dependencies/hdf5.py
+++ b/mesonbuild/dependencies/hdf5.py
@@ -96,13 +96,13 @@ class HDF5ConfigToolDependency(ConfigToolDependency):
 
         if language == 'c':
             cenv = 'CC'
-            tools = ['h5cc']
+            tools = ['h5cc', 'h5pcc']
         elif language == 'cpp':
             cenv = 'CXX'
-            tools = ['h5c++']
+            tools = ['h5c++', 'h5pc++']
         elif language == 'fortran':
             cenv = 'FC'
-            tools = ['h5fc']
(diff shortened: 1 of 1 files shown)


jpakkane:
Is there ever a case where the user specifically would want to find one name rather than the other?

eli-schwartz:
I guess if you have two builds in different installation prefixes, and you only want the one that happens to have been configured with --enable-parallel?

But then if you have pkg-config files the dependency name is the same pkg-config dependency either way... it's the same cmake dependency name too, but you can check if the cmake dependency defines `HDF5_ENABLE_PARALLEL=ON` to do conditional logic.

PR #9564, opened 2021-11-14
qt module: add depfile support to moc as well

eli-schwartz:
We currently enable this only for rcc (where this really really matters) but it can often matter for moc as well, and is just generally more correct.

Really, this should have been added in #7451 too, but I neglected it since the module warned about inaccurate dependencies only for rcc...

### DIFF
--- a/mesonbuild/modules/qt.py
+++ b/mesonbuild/modules/qt.py
@@ -98,6 +98,7 @@ if T.TYPE_CHECKING:
 class QtBaseModule(ExtensionModule):
     _tools_detected = False
     _rcc_supports_depfiles = False
+    _moc_supports_depfiles = False
 
     def __init__(self, interpreter: 'Interpreter', qt_version: int = 5):
         ExtensionModule.__init__(self, interpreter)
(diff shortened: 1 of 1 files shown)


mesonbuild/modules/qt.py
```diff
@@ -169,6 +170,10 @@ def _detect_tools(self, state: 'ModuleState', method: str, required: bool = True
         if qt.found():
             # Get all tools and then make sure that they are the right version
             self.compilers_detect(state, qt)
+            if version_compare(qt.version, '>=5.14.0'):
+                self._moc_supports_depfiles = True
+            else:
+                mlog.warning('rcc dependencies will not work properly until you move to Qt >= 5.15', fatal=False)
```
dcbaker, line 176: That should be moc, not RCC, right?

dcbaker (approved):
Other than the one cut and paste this looks good to me

mesonbuild/modules/qt.py
eli-schwartz: copy-paste typo...

PR #9565, opened 2021-11-14
reject run_target in test or install script arguments

bonzini:
Found by @dcbaker  during the review of #9400 ~~(and based on it to avoid conflicts, since it is approved already).~~


dcbaker (approved):
LGTM, thanks

jpakkane:
The Cygwin test failure seems relevant.

```
 FileNotFoundError: [Errno 2] No such file or directory: '/cygdrive/d/a/meson/meson/test cases/common/22 object extraction/create-source.py'
```

bonzini:
Fixed now, thanks (and also reordered the patches so that this can be in 0.60.2 as per milestone).

PR #9568, opened 2021-11-14
Fix more code which appears copied from CcrxCompiler.

nathanaelg:
I missed a few more places of `-include` and `-define` in the last PR.


eli-schwartz:
The CI failure should be fixed by rebasing on git master.

Issue #9543, opened 2021-11-08
Uninstalled pc files generated by pkgconfig module do not set RPATH


nirbheek:
You can successfully link to shared libraries provided by the uninstalled pkgconfig files, but you cannot run your executable(s) without setting `LD_LIBRARY_PATH` or `DYLD_LIBRARY_PATH`. This is usually fine, but on macOS this is strongly discouraged and can break things. For instance, using the libjpeg-turbo wrap on macOS and setting `DYLD_LIBRARY_PATH` causes _all_ programs run to fail with:

```
$ vim
dyld: Symbol not found: __cg_jpeg_resync_to_restart
  Referenced from: /System/Library/Frameworks/ImageIO.framework/Versions/A/Resources/libTIFF.dylib
  Expected in: /Users/nirbheek/projects/repos/gst-build/_build_macos/subprojects/libjpeg-turbo-2.1.0/libJPEG.dylib
 in /System/Library/Frameworks/ImageIO.framework/Versions/A/Resources/libTIFF.dylib
Abort trap: 6
```

The solution is to use RPATHs instead, but for that the uninstalled pkgconfig files should set them; something like this hacky patch to test that it works:

```diff
--- a/mesonbuild/modules/pkgconfig.py
+++ b/mesonbuild/modules/pkgconfig.py
@@ -394,7 +394,10 @@ class PkgConfigModule(ExtensionModule):
                                 Lflag = '-r${libdir}/%s' % l.filename
                         else:
                             if isinstance(install_dir, str):
-                                Lflag = '-L{}'.format(self._escape(self._make_relative(prefix, install_dir)))
+                                libdir = self._escape(self._make_relative(prefix, install_dir))
+                                Lflag = '-L{}'.format(libdir)
+                                if uninstalled:
+                                    Lflag += ' -Wl,-rpath,' + libdir
                             else:  # install_dir is True
                                 Lflag = '-L${libdir}'
                         if Lflag not in Lflags:
```

I think we should add RPATH entries to `Libs:` for uninstalled pc files on all UNIX platforms. It will eliminate the need for setting env vars, and will also make `meson devenv` more useful.

CC: @xclaesse

nirbheek:
This turned out to be a bug in how we consume non-system pkgconfig dependencies: https://github.com/mesonbuild/meson/pull/9573. The tests show that this use-case is supposed to work, and is intended to work.

Issue #4372, opened 2018-10-15
Adds build-rpath for .pc library in non-standard path, but only if Requires: is not set


wjt:
I have a C application which loads out-of-tree plugins. It provides a helper library that these plugins link against, which contains a GObject interface that plugins must implement (and some related helper APIs). The helper library is in a private subdirectory of `$libdir` so as not to pollute the global namespace; plugins are loaded from another subdirectory, whose path is included in the pkg-config file.

The app is free but the plugin I'm developing is not; I've extracted the important bits to https://github.com/wjt/meson-plugin-rpath/ to demonstrate the problem outlined below.

The external plugin finds the helper library with pkg-config and builds a `shared_library` for the plugin itself, and a non-installed executable test case. However, the test case fails to link:

```
[...]
[3/6] cc  -o libplugin.so 'plugin@sha/plugin.c.o' -Wl,--no-undefined -Wl,--as-needed -shared -fPIC -Wl,--start-group -Wl,-soname,libplugin.so /tmp/throwaway/lib/x86_64-linux-gnu/app/libhelper.so /usr/lib/x86_64-linux-gnu/libglib-2.0.so -Wl,--end-group
[...]  
[5/6] cc  -o plugin-test 'plugin-test@exe/test.c.o' -Wl,--no-undefined -Wl,--as-needed -Wl,--start-group libplugin.so -Wl,--end-group '-Wl,-rpath,$ORIGIN/' -Wl,-rpath-link,/sysroot/home/wjt/src/wjt/meson-plugin-rpath/plugin/build/  
FAILED: plugin-test 
cc  -o plugin-test 'plugin-test@exe/test.c.o' -Wl,--no-undefined -Wl,--as-needed -Wl,--start-group libplugin.so -Wl,--end-group '-Wl,-rpath,$ORIGIN/' -Wl,-rpath-link,/sysroot/home/wjt/src/wjt/meson-plugin-rpath/plugin/build/  
/usr/bin/ld: warning: libhelper.so, needed by libplugin.so, not found (try using -rpath or -rpath-link)
libplugin.so: undefined reference to `helper_get_three'
collect2: error: ld returned 1 exit status
ninja: build stopped: subcommand failed.
```

The `.pc` file looks like this:

```
prefix=/tmp/throwaway
libdir=${prefix}/lib/x86_64-linux-gnu
includedir=${prefix}/include

pluginsdir=/tmp/throwaway/lib/x86_64-linux-gnu/app/plugins

[...]

PR #2397, opened 2017-09-30
Better support for prebuilt shared libs

jpakkane:
The first few commits move prebuild tests under unit tests where they belong.

Then better usability is added to prebuilt shared libraries. This is done by adding an rpath entry for all shared libraries that are inside the source tree. This makes them work transparently.

This might require some tweaking on Windows, but the basic principle should be solid.

PR #9573, opened 2021-11-15
Set RPATH for all non-system libs with absolute paths

nirbheek:
If a pkg-config dependency has multiple libraries in it, which is the most common case when it has a `Requires:` directive, or when it has multiple `-l` args in `Libs:` (rare), then we don't add `-Wl,-rpath` directives to it when linking.

The existing test wasn't catching it because it was linking to a pkgconfig file with a single library in it. Update the test to demonstrate this.

This function was originally added for shared libraries in the source directory, which explains the name: https://github.com/mesonbuild/meson/pull/2397

However, now it is also used for linking to *all* non-system shared libraries that we link to with absolute paths: https://github.com/mesonbuild/meson/pull/3092

But that PR is incomplete / wrong, because it only adds RPATHs for dependencies that specify a single library, which is simply inconsistent. Things will work for some dependencies and not work for others, with no logical reason for it.

We should add RPATHs for *all* libraries. There are no special length limits for RPATHs that I can find.

For ELF, `DT_RPATH` or `DT_RUNPATH` are used, which have are just stored in a string table `DT_STRTAB`. The maximum length is only a problem when editing pre-existing tags.

For Mach-O, each RPATH is stored in a separate `LC_RPATH` entry so there are no length issues there either.

Fixes https://github.com/mesonbuild/meson/issues/9543

Fixes https://github.com/mesonbuild/meson/issues/4372

### DIFF
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -708,7 +708,6 @@ class Backend:
         return False
 
     def get_external_rpath_dirs(self, target: build.BuildTarget) -> T.Set[str]:
-        dirs: T.Set[str] = set()
         args: T.List[str] = []
         for lang in LANGUAGES_USING_LDFLAGS:
             try:
@@ -719,6 +718,11 @@ class Backend:
                     args.extend(e)
             except Exception:
                 pass
+        return self.get_rpath_dirs_from_link_args(args)
+
+    @staticmethod
+    def get_rpath_dirs_from_link_args(args: T.List[str]) -> T.Set[str]:
+        dirs: T.Set[str] = set()
         # Match rpath formats:
         # -Wl,-rpath=
         # -Wl,-rpath,
@@ -777,6 +781,8 @@ class Backend:
                     paths.add(os.path.join(self.build_to_src, rel_to_src))
                 else:
                     paths.add(libdir)
+            # Don't remove rpaths specified by the dependency
+            paths.difference_update(self.get_rpath_dirs_from_link_args(dep.link_args))
         for i in chain(target.link_targets, target.link_whole_targets):
             if isinstance(i, build.BuildTarget):
                 paths.update(self.rpaths_for_non_system_absolute_shared_libraries(i, exclude_system))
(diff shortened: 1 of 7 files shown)


nirbheek:
This will also fix https://github.com/mesonbuild/meson/issues/4372.

nirbheek:
The only remaining test failure is the llvm cmake static-libs=true UNEXRUN that is broken on master.

nirbheek:
The diff is smaller than it looks, check it out in whitespace-ignored mode: https://github.com/mesonbuild/meson/pull/9573/files?diff=unified&w=1 :wink:

nirbheek:
Needs to be rebased on top of https://github.com/mesonbuild/meson/pull/9546 once it's merged

nirbheek:
Rebased and ready for review / merging :grin:

mesonbuild/backend/backends.py
```diff
@@ -737,37 +741,40 @@ def get_external_rpath_dirs(self, target: build.BuildTarget) -> T.Set[str]:
...
+                if libdir.startswith(self.environment.get_source_dir()):
+                    rel_to_src = libdir[len(self.environment.get_source_dir()) + 1:]
+                    assert not os.path.isabs(rel_to_src), f'rel_to_src: {rel_to_src} is absolute'
+                    paths.add(os.path.join(self.build_to_src, rel_to_src))
+                else:
+                    paths.add(libdir)
+            # Don't remove rpaths specified by the dependency
+            for d in self.get_rpath_dirs_from_link_args(dep.link_args):
+                if d in paths:
+                    paths.remove(d)
```
dcbaker, line 773:
```suggestion
            paths.update_difference(self.get_rpath_dirs_from_link_args(dep.link_args))
```

I think?

nirbheek: `difference_update()` does indeed do what I want it to, thanks!

nirbheek: Hah, it wasn't implemented for our implementation of OrderedSet, so I added it.

nirbheek:
I'm inclined to merge this tomorrow if no one has any objections :slightly_smiling_face:

PR #6421, opened 2020-01-07
Add a "system" dependency for zlib

dcbaker:
This adds a system depenedency for zlib, so that platforms that ship zlib as part of the base os (macOS, FreeBSD, and probably other BSDs but I don't have VMs for all of them).  This allows zlib to be found without the use of cmake of pkg-config, which is especially helpful on macOS.

This has been updated on top of #6443

Fixes #2654

PR #9576, opened 2021-11-15
dependencies/zlib: Add system zlib method for android

Dudemanguy:
I noticed that the zlib system method currently doesn't work on android and it seems it wasn't included with #6421. The same method that the BSDs use should also work for android, and with some very light testing on my end with cross-compiling to android it behaves as expected.

### DIFF
--- a/docs/markdown/Dependencies.md
+++ b/docs/markdown/Dependencies.md
@@ -711,7 +711,7 @@ $ wx-config --libs std stc
 ## Zlib
 
 Zlib ships with pkg-config and cmake support, but on some operating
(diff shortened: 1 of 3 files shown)


Dudemanguy:
Hmm, that's strange. The check that failed passes locally on my machine. I'm not sure what the discrepancy is.

eli-schwartz:
There's a known unexpected-run on msys2, it's not related to this PR.

Dudemanguy:
>There's a known unexpected-run on msys2, it's not related to this PR.

Ah cool. Thanks!

PR #9579, opened 2021-11-16
dependencies/zlib: Allow for generic OS checks

Dudemanguy:
Contrary to most system method checks, zlib currently functions as a
whitelist of OSes. This isn't really needed however. The first special
case for OSes that provide zlib as part of the base OS is worth keeping.
However, the elif for windows is more than generic enough to allow any
other potential OSes to try. Just make it a simplie if/else instead.


eli-schwartz:
Can you rebase on master to catch the CI fixes?

Dudemanguy:
Done.

PR #9586, opened 2021-11-18
Fix vs backend cross compilation regression

GustavoLCR:
It broke after 1fb2818


eli-schwartz:
Can you rebase on master to catch the CI fixes?

PR #9588, opened 2021-11-18
dependencies: add a system dependency for OpenSSL

eli-schwartz:
On some platforms, this is provided by the base system, which nevertheless refrains from providing the .pc files that go along with it. As a result, it's impossible to sensibly find these dependencies.

I'm looking at you, FreeBSD...

Upstream tracking bug: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=257659

This is also a problem on macOS, but that's unfixable (and I believe they provide some truly ancient and out-of-support version anyway) and you probably want to use the homebrew OpenSSL. Still, meson can now detect these old versions and even set their version number.

### DIFF
--- a/mesonbuild/dependencies/__init__.py
+++ b/mesonbuild/dependencies/__init__.py
@@ -36,7 +36,7 @@ from .misc import (
     BlocksDependency, OpenMPDependency, cups_factory, curses_factory, gpgme_factory,
     libgcrypt_factory, libwmf_factory, netcdf_factory, pcap_factory, python3_factory,
     shaderc_factory, threads_factory, ThreadDependency, iconv_factory, intl_factory,
-    dl_factory
+    dl_factory, openssl_factory, libcrypto_factory, libssl_factory,
 )
 from .platform import AppleFrameworks
 from .qt import qt4_factory, qt5_factory, qt6_factory
@@ -256,6 +256,9 @@ packages.update({
     'iconv': iconv_factory,
     'intl': intl_factory,
     'dl': dl_factory,
+    'openssl': openssl_factory,
+    'libcrypto': libcrypto_factory,
+    'libssl': libssl_factory,
 
     # From platform:
     'appleframeworks': AppleFrameworks,
--- a/mesonbuild/dependencies/factory.py
+++ b/mesonbuild/dependencies/factory.py
@@ -49,6 +49,10 @@ if T.TYPE_CHECKING:
         T.List[DependencyGenerator]
     ]
 
+    # This should be str, Environment, T.Dict[str, T.Any], T.Optional[str]
+    # But if you try that, you get error: Cannot infer type of lambda
+    CmakeDependencyFunc = T.Callable[..., CMakeDependency]
+
 class DependencyFactory:
 
     """Factory to get dependencies from multiple sources.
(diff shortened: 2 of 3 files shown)


dcbaker:
can these be found with cmake? we presumably shouldnt break that if it currently works

eli-schwartz:
If you try it you first hit the issue where "openssl" isn't a valid dependency, and when you try messing with the case and making it into "OpenSSL" you get this:

```
WARNING: Could not find and exact match for the CMake dependency OpenSSL.

However, Meson found the following partial matches:

    ['OpenSSL::Crypto', 'OpenSSL::SSL', 'OpenSSL::applink']

Using imported is recommended, since this approach is less error prone
and better supported by Meson. Consider explicitly specifying one of
these in the dependency call with:

    dependency('OpenSSL', modules: ['OpenSSL::<name>', ...])

Meson will now continue to use the old-style OpenSSL_LIBRARIES CMake
variables to extract the dependency information since no explicit
target is currently specified.


Run-time dependency openssl found: YES 1.1.1l
Run-time dependency libssl found: NO (tried cmake)

t/meson.build:4:0: ERROR: Dependency "libssl" not found, tried cmake
```

So it's impossible to support robustly across pkg-config and cmake, however, this PR does mean you cannot do `dependency('OpenSSL', modules: ['OpenSSL::Crypto'])` as a fallback for `dependency('libcrypto', required: false)`.

The actual "openssl.pc" is just a meta target anyway...

eli-schwartz:
In theory I suppose we could make cmake work for the first time ever? by plugging it into the "correct" cmake analogue names. I'm not entirely sure how to do that since e.g. `DependencyFactory(..., cmake_name='XXX')` doesn't provide an interface for the modules.

eli-schwartz:
> Command line:  cc -L/usr/local/lib -I/usr/local/include /Users/runner/work/meson/meson/b 240c1a102d/meson-private/tmpds8waqvp/testfile.c -o /Users/runner/work/meson/meson/b 240c1a102d/meson-private/tmpds8waqvp/output.exe -O0 -Werror=implicit-function-declaration -F/Applications/Xcode_13.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.0.sdk/System/Library/Frameworks -framework python 
>  ld: cannot link directly with dylib/framework, your binary is not an allowed client of /Applications/Xcode_13.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.0.sdk/System/Library/Frameworks/python.framework/python.tbd for architecture x86_64

I don't understand this error...

dcbaker:
That is Apple's completely intuitive deprecation message. As you can clearly tell, you're not supposed to link with that anymore.

eli-schwartz:
@dcbaker @mensinda

I'm not really sure what to do about cmake here. Advice? Decisions on whether to just ignore it? I would like to move forward with this somehow.

dcbaker:
@eli-schwartz I think the cmake_class signature needs to be replaced with:
```python
cmake_class: T.Union[Type[CmakeDependency], T.Callable[[str, Environment, T.Dict[str, T.Any], T.Optional[str]], CmakeDependency]
```
Because of course `Type[T]` does not satisfy `Callable[..., T]`

eli-schwartz:
Luckily, mypy is excellent at analyzing lambdas and doesn't at all error out and require us to use `T.Callable[..., CMakeDependency]` instead.

dcbaker (approved):
LGTM, thanks for dealing with cmake

PR #9589, opened 2021-11-18
Check correct directory for wrap install

piotrrak:
wrap install checks if its subproject directory already exists
and fails in this case.
Current wrapdb wraps specify this directory as {name}-{version},
while subprojects/{name} directory was checked

Signed-off-by: Piotr Rak <piotr.rak@gmail.com>

### DIFF
--- a/mesonbuild/wrap/wraptool.py
+++ b/mesonbuild/wrap/wraptool.py
@@ -91,12 +91,12 @@ def install(options: 'argparse.Namespace') -> None:
     name = options.name
     if not os.path.isdir('subprojects'):
         raise SystemExit('Subprojects dir not found. Run this script in your source root directory.')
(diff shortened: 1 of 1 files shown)


eli-schwartz:
I think the purpose of the check is because you'd end up with conflicts if you have a pre-existing `subproject('foo') that uses `subprojects/foo/`, and then also try to add a foo.wrap.

That specific subproject() invocation would never try to configure `subprojects/foo-1.0/` though. So I don't see a reason to restrict that here.

piotrrak:
Sure, what it *prevented* me to do seems impossible anyway right now.
My project has `.gitmodules` I'd prefer to keep and possibly use with other scripts, but I'd like to use as many wrapdb wraps as possible with meson.

PR #9593, opened 2021-11-18
tests: remove extraframework test for python

dcbaker:
This tries to link the system provided python, which is deprecated and
will result in an ambiguous error like "your binary is not an allowed
client of .../Library/Frameworks/python.framework/python.tbd for
architecture x86_64". Just remove it, we already have coverage of
extraframeworks anyway.


eli-schwartz:
Following up on https://github.com/mesonbuild/meson/pull/9588#issuecomment-972504616 I actually tried rerunning the CI workflow several times, each time a different combination of jobs failed (and every single one passed at least once, on the same commit). Baffling.

...

There's still another python extraframeworks lookup in test_extraframework_dependency_method.

dcbaker:
Alright, I'll flail a bit and try metal :) hopefully that's the name of the framework.

PR #9594, opened 2021-11-18
More typed_kwargs for the gnome module

dcbaker:
This is a smaller follow up to the last set of cleanups to the gnome module with conversions to `typed_kwargs`. There are only two functions covered here, genmarshal and generate_vapi. Again, this allows a number of nice cleanups. In addition I noticed that genmarshal still exposes `build_always` without deprecation. I've made that deprecated, and added the `build_always_stale` and `build_by_default` keyword arguments matching the behavior of `custom_target`.

### DIFF
--- a/mesonbuild/modules/gnome.py
+++ b/mesonbuild/modules/gnome.py
@@ -1086,12 +1086,15 @@ class GnomeModule(ExtensionModule):
         self._devenv_prepend('GSETTINGS_SCHEMA_DIR', os.path.join(state.environment.get_build_dir(), state.subdir))
         return ModuleReturnValue(target_g, [target_g])
 
-    @FeatureDeprecatedKwargs('gnome.yelp', '0.43.0', ['languages'],
-                             'Use a LINGUAS file in the source directory instead')
     @typed_pos_args('gnome.yelp', str, varargs=str)
     @typed_kwargs(
         'gnome.yelp',
-        KwargInfo('languages', ContainerTypeInfo(list, str), listify=True, default=[]),
+        KwargInfo(
+            'languages', ContainerTypeInfo(list, str),
+            listify=True, default=[],
+            deprecated='0.43.0',
+            deprecated_message='Use a LINGUAS file in the source directory instead',
+        ),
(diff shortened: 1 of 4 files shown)


dcbaker:
The `lgtm` warning makes no sense, it thinks a function returning a tuple of 5 elements returns 4, then complains about it.

mesonbuild/modules/gnome.py
```diff
@@ -1664,49 +1679,47 @@ def _make_mkenum_custom_target(
...
-    @permittedKwargs({'sources', 'prefix', 'install_header', 'install_dir', 'stdinc',
-                      'nostdinc', 'internal', 'skip_source', 'valist_marshallers',
-                      'extra_args'})
     @typed_pos_args('gnome.genmarshal', str)
-    def genmarshal(self, state: 'ModuleState', args: T.Tuple[str], kwargs) -> ModuleReturnValue:
+    @typed_kwargs(
+        'gnome.genmarshal',
+        DEPEND_FILES_KW,
+        INSTALL_KW.evolve(name='install_header'),
+        KwargInfo('build_always', bool, default=False),
```
eli-schwartz: There's a bunch of mysterious kwargs here, that are not in the documentation *and* are not in permittedKwargs. Attempting to use them raises a fatal error, e.g. 

```
genmarshal/meson.build:1:0: ERROR: Got unknown keyword arguments "build_always"
```

In commit 80d665e8decd644d737930d450f0bc66d6b5b02c, we added that error checking in, by following the documentation. At the time, the error message would be:

```
Meson encountered an error in file genmarshal/meson.build, line 1, column 0:
Invalid keyword argument build_always.
```

We also have some terrible manual argument checking in known_kwargs and known_custom_target_kwargs, but they've been entirely overridden for 4.5 years now, and no one has ever complained that they need these undocumented kwargs, which, again, were undocumented.

I see no grounds to re-add support for them half a decade later, especially just to mark at least one of them as deprecated in a later commit in the same PR... and if any of them do get added, I want them to be documented.

eli-schwartz (changes requested):

eli-schwartz (approved):

PR #9595, opened 2021-11-19
Mention  in error message that ExtractedObjects are usable objects for BuildTarget

piotrrak:


mesonbuild/build.py
```diff
@@ -765,8 +765,9 @@ def process_objectlist(self, objects):
             elif isinstance(s, (GeneratedList, CustomTarget)):
                 msg = 'Generated files are not allowed in the \'objects\' kwarg ' + \
                     f'for target {self.name!r}.\nIt is meant only for ' + \
-                    'pre-built object files that are shipped with the\nsource ' + \
-                    'tree. Try adding it in the list of sources.'
+                    'pre-built shipped with\nthe source tree ' + \
```
eli-schwartz, line 768: Wording feels a bit clumsy.

Hmm, what about:

> It is meant only for object files that are pre-built and shipped in the source tree, or extracted from a target object.

bonzini: Or "extracted from a build target". Otherwise there's confusion between "object files" and "objects of the Meson language".

bonzini:
Obsoleted by 8d29400.

PR #9596, opened 2021-11-19
Checkout subproject submodules even when nested

mon:
I'm using submodules for my subprojects because I prefer how they work over wraps. I also like how meson will automatically checkout any submodules it finds in subprojects.

However, my subproject itself has subprojects, which are also git submodules. For these, meson's current check fails because it is essentially:  
`cd project; git submodule status project/subprojects/top_dep/subprojects/bottom_dep`  
Because the command is run in the project directory, git only checks for the submodule there instead of inside `top_dep`:  
`error: pathspec 'project/subprojects/top_dep/subprojects/bottom_dep' did not match any file(s) known to git`

Uninitialised but present submodules always get an empty directory checked out, so it's safe to just use the submodule directory for the working dir, changing the equivalent command to:  
`cd project/subprojects/top_dep/subprojects/bottom_dep; git submodule status .`  
... which works fine.

### DIFF
--- a/mesonbuild/wrap/wrap.py
+++ b/mesonbuild/wrap/wrap.py
@@ -372,12 +372,16 @@ class Resolver:
         # definitely cannot try to conveniently set up a submodule.
         if not GIT:
             return False
+        # Does the directory exist? Even uninitialised submodules checkout an
+        # empty directory to work in
+        if not os.path.isdir(self.dirname):
+            return False
         # Are we in a git repository?
-        ret, out = quiet_git(['rev-parse'], self.subdir_root)
+        ret, out = quiet_git(['rev-parse'], Path(self.dirname).parent)
         if not ret:
             return False
         # Is `dirname` a submodule?
-        ret, out = quiet_git(['submodule', 'status', self.dirname], self.subdir_root)
+        ret, out = quiet_git(['submodule', 'status', '.'], self.dirname)
         if not ret:
             return False
         # Submodule has not been added, add it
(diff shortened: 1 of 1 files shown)


eli-schwartz:
This looks like a reasonable change. Can you rebase on git master to see if that resolves the CI error? I think it should :)

eli-schwartz:
... also can you please squash the commits together as I don't see a reason for them to be two distinct commits. It seems that the second commit fixes an issue with the first commit.

mon:
Good point, should be good now!

mesonbuild/wrap/wrap.py
```diff
@@ -372,12 +372,16 @@ def resolve_git_submodule(self) -> bool:
         # definitely cannot try to conveniently set up a submodule.
         if not GIT:
             return False
+        # Does the directory exist? Even uninitialised submodules checkout an
+        # empty directory to work in
+        if not os.path.isdir(self.dirname):
+            return False
         # Are we in a git repository?
-        ret, out = quiet_git(['rev-parse'], self.subdir_root)
+        ret, out = quiet_git(['rev-parse'], str(Path(self.dirname).parent))
```
eli-schwartz, line 380: Hmm, I just realized. You are converting a Path to a str, because mesonlib.quiet_git requires a str there... but it is directly passed to subprocess.Popen as `cwd=`, and that accepts os.PathLike.

So, if we just fix that typing issue...

eli-schwartz: #9888

mon: There appears to be a lot of code calling the `*git` functions - is mypy happy with passing a string to a signature accepting a PathLike? Otherwise it's a much bigger change.

Also is the expectation that I make the change? I'm more than happy to, just don't want to duplicate work if you're planning on it :)

eli-schwartz: My PR declares support for anything that subprocess.Popen itself takes, i.e. `T.Union[str, bytes, os.PathLike]`, so mypy should be happy. :)

eli-schwartz: TBH, in the long run we kind of want to use fewer raw strings anyway, PathLike objects like pathlib.Path are really nice for this exact case, so... fixing `*git()` is just a good idea no matter what.

mon: Oh lovely, of course just making them take a union works.

I'll wait for that to merge, then rebase again without the cast!

eli-schwartz: Merged, please rebase. :)

eli-schwartz (approved):

PR #9599, opened 2021-11-19
Improve error messages for typed_kwarg type mismatches in containers

dcbaker:
Currently if an argument is `T.List[str]`, but you pass `T.List[int]`, then you get a message like `should ahve been List[str], but was list`. That's not helpful.

This series makes a few small tweaks to make this all much more useful. First, it replaces that wit "but was" message with a more useful `but was list[int]`. Then it replaces "list" with "array", since meson has array's, not lists. Finally, it moves to the more standard `|` as a representation for union types, ie `array[int | str]`, which both python and typescript use, so it should be clearer.


eli-schwartz:
And you even have a free fix for run_single_test being useless for tests that invert the error code and test that certain things should produce errors! That one's been bugging me for a while...

Issue #9587, opened 2021-11-18
Feature: build_target for objects `build_target(target_type:'objects')`


piotrrak:
At this moment there is no straightforward way to create library/executable with `<lang>_flags` defined per compilation unit.

The idea to do that is to provide functionality that is close to cmake per source properties, but bit more powerful and convenient.

Example usage could be:
```
avx_objects = build_target(
  target_type:'objects',
  sources: ['my_super_fast_kernel.cc'],
  cpp_args: my_fast_libflags + avx_flags)`,
  name_prefix='avx_'
)

my_fast_lib = library('fast', ['fast.cc'], 
  cpp_args: my_fastlib_flags,
  objects: avx_objects.extract_all_objects()
)
```

I am aware that can be achieved by using `build_target(target_type:'static_library')` but considering it simply might be single object creating ar archive to achieve that is bit suboptimal.

That also could be helpful to improve/fix cmake.subproject() support which currently gets it wrong.
During compilation of OpenCV for entry from `cmake_trace.txt`:

`{"args":["SOURCE","/home/prak/DALI_deps/builddir/third_party/opencv/__CMake_build/modules/imgproc/sumpixels.avx2.cpp","PROPERTY","COMPILE_DEFINITIONS","CV_CPU_DISPATCH_MODE=AVX2"],"cmd":"set_property","file":"/home/prak/DALI_deps/builddir/meson-private/data/preload.cmake","frame":6,"line":69,"time":1637193286.2726841}`

Library with union of all source files `COMPILE_DEFINITIONS` / flags is created in `meson.build`
This results in:

```
build third_party/opencv/libopencv_imgproc.a.p/_home_prak_DALI_deps_builddir_third_party_opencv___CMake_build_modules_imgproc_sumpixels.avx2.cpp.o: cpp_COMPILER /home/prak/DALI_deps/builddir/third_party/opencv/__CMake_build/modules/imgproc/sumpixels.avx2.cpp || third_party/opencv/opencl_kernels_imgproc.hpp
 DEPFILE = third_party/opencv/libopencv_imgproc.a.p/_home_prak_DALI_deps_builddir_third_party_opencv___CMake_build_modules_imgproc_sumpixels.avx2.cpp.o.d
 DEPFILE_UNQUOTED = third_party/opencv/libopencv_imgproc.a.p/_home_prak_DALI_deps_builddir_third_party_opencv___CMake_build_modules_imgproc_sumpixels.avx2.cpp.o.d
[...]

PR #9600, opened 2021-11-19
WIP: compiled_objects() with ninja backend support

piotrrak:
Some discussion in #9587

### DIFF
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -118,6 +118,7 @@ known_shlib_kwargs = known_build_target_kwargs | {'version', 'soversion', 'vs_mo
 known_shmod_kwargs = known_build_target_kwargs | {'vs_module_defs'}
 known_stlib_kwargs = known_build_target_kwargs | {'pic', 'prelink'}
 known_jar_kwargs = known_exe_kwargs | {'main_class'}
+known_objs_kwargs = known_build_target_kwargs | {'pic'}
 
 @lru_cache(maxsize=None)
 def get_target_macos_dylib_install_name(ld) -> str:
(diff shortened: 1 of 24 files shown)


mesonbuild/build.py
```diff
@@ -2262,6 +2265,41 @@ def __init__(self, name, subdir, subproject, for_machine: MachineChoice, sources
     def get_default_install_dir(self, environment) -> T.Tuple[str, str]:
         return environment.get_shared_module_dir(), '{moduledir_shared}'
 
+class CompiledObjects(BuildTarget):
+    known_kwargs = known_objs_kwargs
+
+    def __init__(self, name, subdir, subproject, for_machine: MachineChoice, sources, objects, environment, kwargs):
+        self.typename = 'compiled objects'
+        if 'version' in kwargs:
+            raise MesonException('Built objects must not specify the version kwarg.')
```
dcbaker, line 2274: `InvalidArguments` is usually the specific exception used for this kind of problem

mesonbuild/interpreter/interpreter.py
```diff
@@ -1594,8 +1603,13 @@ def func_build_target(self, node, args, kwargs):
         target_type = kwargs.pop('target_type')
         if target_type == 'executable':
             return self.build_target(node, args, kwargs, build.Executable)
+        elif target_type == 'objects':
+            return self.build_target(node, args, kwargs, build.CompiledObjects)
         elif target_type == 'shared_library':
             return self.build_target(node, args, kwargs, build.SharedLibrary)
+            # TODO: adjust to release
+            FeatureNew('build_target(target_type: \'objects\')',
+                       '0.61.0').use(self.subproject)
```
dcbaker, line 1612: We generally use the `single_use` factory method here. I think you can also pass the `node` argument to `FeatureNew` to get line speicifc error information.

piotrrak: Could use bit of advice on how to implement compiled_objects install:

Here is my attempt:
https://github.com/piotrrak/meson/tree/compiled_objects_install

Got bit stuck here.

Cleanest way I see is to be able to create list of objects from sources in `CompiledObjects` and specify it as `target.outputs`. That would mean moving logic of object list generation from `NinjaBackend::generate_target` to `BuildTarget`. This feels ackward and ninja backend specific;

It is more involved than I would expected due to:
- vala
- cython
- generated sources

Second option is to lift it Backend and special case for CompiledObjects in Backend::generate_target install.

Anyone has insights about how this should be solved?

PR #9602, opened 2021-11-21
Make the generated reproducible .cmake files reproducible.

lamby:
Whilst working on the [Reproducible Builds effort](https://reproducible-builds.org/), I noticed that meson did not generate reproducible `.cmake` files: they include the full path name. This commit not only makes the build reproducible, but it also matches CMake's own behaviour. Specifically, CMakePackageConfigHelpers.cmake does the equivalent transformation using `get_filename_component(inputFileName "${_inputFile}" NAME)`.

I originally filed this in Debian as [bug #1000327](https://bugs.debian.org/1000327).

### DIFF
--- a/mesonbuild/modules/cmake.py
+++ b/mesonbuild/modules/cmake.py
@@ -299,7 +299,7 @@ class CmakeModule(ExtensionModule):
 
     def create_package_file(self, infile, outfile, PACKAGE_RELATIVE_PATH, extra, confdata):
         package_init = PACKAGE_INIT_BASE.replace('@PACKAGE_RELATIVE_PATH@', PACKAGE_RELATIVE_PATH)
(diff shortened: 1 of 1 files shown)


eli-schwartz (approved):
Huh, that makes a lot of sense... I wonder why this used the whole path in the first place, the cmake module does try to explicitly copy cmake's output.

lamby:
Perhaps it doesn't usually matter in that `infile` is usually relative? Either way, its pretty harmless as it's commented anyway. :)

eli-schwartz:
We literally mention `Any changes to this file will be overwritten by the next CMake run` which is untrue as it will be overwritten by the next *meson* run... that is an irrelevant comment too. :D

Over in my neck of the woods this sort of variation isn't actually tested as build paths are predictable. :/ Thanks for the catch.

PR #6109, opened 2019-10-26
Prefer ext static libs when --default-library=static

rc-matthew-l-weber:
This patch adds a case in the library pattern logic to prefer static
libraries when the Meson Core option for "default_library" is set to
solely static.

The existing library search order makes sense for cases of shared and
shared / static mixed. However if using a prebuilt cross-toolchain,
they usually provide both a static and shared version of sysroot
libraries. This presents a problem in a complete static build where
there won't be shared libraries at runtime and during build time there
are failures like "ld: attempted static link of dynamic object".

Bug:
https://github.com/mesonbuild/meson/issues/6108

Signed-off-by: Matthew Weber <matthew.weber@rockwellcollins.com>

PR #2816, opened 2017-12-21
PkgConfigDependency: Add option to choose between shared or static

xclaesse:
This adds 2 global project options:
- default_link: Tells that we prefer static link when possible. This
option could be used in more places where both static and shared
libraries are available (e.g. future both_library()).
- static_paths: List of path prefixes where it is allowed to use static
libraries. By default it has "/" which means all paths are allowed.
Sometimes it is useful to have there the path to an SDK that we want to
static link and keep dynamic link for everything else (e.g. plateform
libraries).

Closes #2765.

PR #9603, opened 2021-11-21
[RFC] add prefer_static built-in option

### LINKED ISSUE #7621: Feature wish: Option to force static:true for all dependency() calls
I don't have much experience with build systems etc. I've currently been confronted with meson for the first time, trying to create a static binary of imagemagick. Many dependencies like glib and pango seem to have switched to meson. So far, the experience is great and much better than with other systems (Makefile, CMake etc.) which I often found way more confusing and less readable. I found, though, that to successfully compile some libraries as static, I needed to patch their meson.build files and add a "static: true" to their dependencies. For example to compile a static pango with a static cairo, I needed to add static:true to the dependency('cairo') call. Surprisingly, simply specifiying "-Ddefault_library=static" wasn't sufficient, but probably this option is about sth. different. Nevertheless, I think it would be very nice to have an option to always prefer the static version for all dependencies if it is available. Sth. like "-Dforce_static=true". Hope it makes sense ;-)

Dudemanguy:
By default, meson will try to look for shared libraries first before
static ones. In the meson.build itself, one can use the static keyword
to control if a static library will be tried first but there's no simple
way for an end user performing a build to switch back and forth at will.
Let's cover this usecase by adding an option that allows a user to
specify if they want dependency lookups to try static or shared
libraries first. The writer of the meson.build can manually specify the
static keyword where appropriate which will override the value of this
option.

I suppose this needs test cases to go with it? I was a bit lost on where exactly tests should be made/created so that was left out for now (guidance would be appreciated).

Fixes #2816
Fixes #4276
Fixes #6109
Fixes #6629
Fixes #7621

### DIFF
--- a/mesonbuild/dependencies/__init__.py
+++ b/mesonbuild/dependencies/__init__.py
@@ -140,8 +140,8 @@ There are a couple of things about this that still aren't ideal. For one, we
 don't want to be reading random environment variables at this point. Those
 should actually be added to `envconfig.Properties` and read in
 `environment.Environment._set_default_properties_from_env` (see how
-`BOOST_ROOT` is handled). We can also handle the `static` keyword. So
-now that becomes:
+`BOOST_ROOT` is handled). We can also handle the `static` keyword and the
+`prefer_static` built-in option. So now that becomes:
 
 ```python
 class FooSystemDependency(ExternalDependency):
@@ -154,7 +154,9 @@ class FooSystemDependency(ExternalDependency):
             self.is_found = False
             return
 
-        static = Mesonlib.LibType.STATIC if kwargs.get('static', False) else Mesonlib.LibType.SHARED
+        get_option = environment.coredata.get_option
+        static_opt = kwargs.get('static', get_option(Mesonlib.OptionKey('prefer_static'))
+        static = Mesonlib.LibType.STATIC if static_opt else Mesonlib.LibType.SHARED
         lib = self.clib_compiler.find_library(
             'foo', environment, [os.path.join(root, 'lib')], libtype=static)
         if lib is None:
--- a/mesonbuild/dependencies/hdf5.py
+++ b/mesonbuild/dependencies/hdf5.py
@@ -51,7 +51,7 @@ class HDF5PkgConfigDependency(PkgConfigDependency):
         newinc = []  # type: T.List[str]
         for arg in self.compile_args:
             if arg.startswith('-I'):
-                stem = 'static' if kwargs.get('static', False) else 'shared'
+                stem = 'static' if self.static else 'shared'
                 if (Path(arg[2:]) / stem).is_dir():
                     newinc.append('-I' + str(Path(arg[2:]) / stem))
         self.compile_args += newinc
@@ -129,7 +129,7 @@ class HDF5ConfigToolDependency(ConfigToolDependency):
         # We first need to call the tool with -c to get the compile arguments
         # and then without -c to get the link arguments.
         args = self.get_config_value(['-show', '-c'], 'args')[1:]
-        args += self.get_config_value(['-show', '-noshlib' if kwargs.get('static', False) else '-shlib'], 'args')[1:]
+        args += self.get_config_value(['-show', '-noshlib' if self.static else '-shlib'], 'args')[1:]
         for arg in args:
             if arg.startswith(('-I', '-f', '-D')) or arg == '-pthread':
                 self.compile_args.append(arg)
--- a/mesonbuild/dependencies/misc.py
+++ b/mesonbuild/dependencies/misc.py
@@ -537,7 +537,7 @@ def shaderc_factory(env: 'Environment',
         shared_libs = ['shaderc']
         static_libs = ['shaderc_combined', 'shaderc_static']
 
-        if kwargs.get('static', False):
+        if kwargs.get('static', env.coredata.get_option(mesonlib.OptionKey('prefer_static'))):
             c = [functools.partial(PkgConfigDependency, name, env, kwargs)
                  for name in static_libs + shared_libs]
         else:
--- a/mesonbuild/dependencies/misc.py
+++ b/mesonbuild/dependencies/misc.py
@@ -175,7 +175,6 @@ class Python3DependencySystem(SystemDependency):
             return
 
         self.name = 'python3'
-        self.static = kwargs.get('static', False)
         # We can only be sure that it is Python 3 at this point
         self.version = '3'
         self._find_libpy3_windows(environment)
--- a/docs/markdown/Builtin-options.md
+++ b/docs/markdown/Builtin-options.md
@@ -79,6 +79,7 @@ machine](#specifying-options-per-machine) section for details.
 | layout {mirror,flat}                 | mirror        | Build directory layout                                         | no             | no                |
 | optimization {0, g, 1, 2, 3, s}      | 0             | Optimization level                                             | no             | no                |
 | pkg_config_path {OS separated path}  | ''            | Additional paths for pkg-config to search before builtin paths | yes            | no                |
+| prefer_static                        | false         | Whether to try static linking before shared linking            | no             | no                |
 | cmake_prefix_path                    | []            | Additional prefixes for cmake to search before builtin paths   | yes            | no                |
 | stdsplit                             | true          | Split stdout and stderr in test logs                           | no             | no                |
 | strip                                | false         | Strip targets on install                                       | no             | no                |
--- a/docs/markdown/Configuring-a-build-directory.md
+++ b/docs/markdown/Configuring-a-build-directory.md
@@ -31,6 +31,7 @@ a sample output for a simple project.
       install_umask   0022          [preserve, 0000-0777]                                      Default umask to apply on permissions of installed files
       layout          mirror        [mirror, flat]                                             Build directory layout
       optimization    3             [0, g, 1, 2, 3, s]                                         Optimization level
+      prefer_static   false         [true, false]                                              Whether to try static linking before shared linking
       strip           false         [true, false]                                              Strip targets on install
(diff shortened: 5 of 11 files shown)


eli-schwartz:
This would probably solve:

#2816
#4276
#6109
#6629
#7621

Dudemanguy:
Fixed the tests.

mesonbuild/dependencies/hdf5.py
```diff
@@ -51,7 +51,8 @@ def __init__(self, name: str, environment: 'Environment', kwargs: T.Dict[str, T.
         newinc = []  # type: T.List[str]
         for arg in self.compile_args:
             if arg.startswith('-I'):
-                stem = 'static' if kwargs.get('static', False) else 'shared'
+                static_opt =  environment.coredata.get_option(OptionKey('prefer_static'))
+                stem = 'static' if kwargs.get('static', static_opt) else 'shared'
```
eli-schwartz, line 55: `kwargs.get('static', False)` seems sub-optimal and as a general cleanup we can probably use `self.static` here (as anywhere that inherits from ExternalDependency).

mesonbuild/dependencies/hdf5.py
```diff
@@ -128,8 +129,9 @@ def __init__(self, name: str, environment: 'Environment', kwargs: T.Dict[str, T.
 
         # We first need to call the tool with -c to get the compile arguments
         # and then without -c to get the link arguments.
+        static_opt =  environment.coredata.get_option(OptionKey('prefer_static'))
         args = self.get_config_value(['-show', '-c'], 'args')[1:]
-        args += self.get_config_value(['-show', '-noshlib' if kwargs.get('static', False) else '-shlib'], 'args')[1:]
+        args += self.get_config_value(['-show', '-noshlib' if kwargs.get('static', static_opt) else '-shlib'], 'args')[1:]
```
eli-schwartz, line 134: Likewise here.

mesonbuild/dependencies/misc.py
```diff
@@ -175,7 +175,7 @@ def __init__(self, name: str, environment: 'Environment', kwargs: T.Dict[str, T.
             return
 
         self.name = 'python3'
-        self.static = kwargs.get('static', False)
+        self.static = kwargs.get('static', environment.coredata.get_option(mesonlib.OptionKey('prefer_static')))
```
eli-schwartz, line 178: This seems odd! Why does this line exist, it seems to be literally overriding the existing self.static attribute...

mesonbuild/dependencies/misc.py
```diff
@@ -537,7 +537,7 @@ def shaderc_factory(env: 'Environment',
         shared_libs = ['shaderc']
         static_libs = ['shaderc_combined', 'shaderc_static']
 
-        if kwargs.get('static', False):
+        if kwargs.get('static', env.coredata.get_option(mesonlib.OptionKey('prefer_static'))):
```
eli-schwartz, line 540: This, however, is a function, not a class, so kwargs.get() is correct as is your modification.

eli-schwartz (commented):
Overall looks like a nice change to make.

While we are here, though, there are a couple things that should be cleaned up on general principle, and if you do clean them up, it means actually implementing this new feature changes fewer lines. With that in mind...

mesonbuild/dependencies/misc.py
Dudemanguy: Do you mean that this is attribute is already inherited from `base.py`? Yeah, I can just delete the line in that case.

eli-schwartz: That is exactly what I mean. :)

mesonbuild/dependencies/__init__.py
```diff
@@ -154,7 +154,9 @@ def __init__(self, name: str, environment: 'Environment', kwargs: T.Dict[str, T.
             self.is_found = False
             return
 
-        static = Mesonlib.LibType.STATIC if kwargs.get('static', False) else Mesonlib.LibType.SHARED
+        get_option = environment.coredata.get_option
+        static_opt = kwargs.get('static', get_option(Mesonlib.OptionKey('prefer_static'))
+        static = Mesonlib.LibType.STATIC if static_opt else Mesonlib.LibType.SHARED
```
Dudemanguy, line 159: Should this change still be here? All of this is comments so I guess it doesn't matter too much, but I don't want to add wrong things in here.

eli-schwartz: Yeah, this documentation is, in part, showing how it all fits together. So I believe it is correct to demonstrate the underlying implementation. IIRC a major motivation in adding this comment section was so that we could move from one core dev understanding how the `mesonbuild/dependencies/` directory works, to multiple core devs understanding.

After all, there is also a `self.libtype` which can technically be used too, but that doesn't obsolete the comment tutorial either.

Dudemanguy:
Split this into two commits now and rebased.

xclaesse:
I haven't looked at code, does it cover both dependency() and find_library()? This new option is better than nothing but unfortunately it's a all-or-nothing option. Ideally we should have more fine grained way of telling which deps should be static and which should be shared.

A long time ago I made that attempt: https://github.com/mesonbuild/meson/pull/4276, and a few others, but we never agreed on a good solution.

The use-case I still have is when you build an app (e.g. GStreamer) for a platform that provide shared libs for some basic deps (e.g. zlib). In that case I want to use the shared libs for stuff provided by the device's base platform and static link everything else into my app.

xclaesse:
That said, since we had that use-case for years and we never found an elegant enough solution, I wouldn't block this PR that already covers a good part of the use-case.

Dudemanguy:
>I haven't looked at code, does it cover both dependency() and find_library()?

That's a really good point. This does not cover find_library() and it really should. I'll fix that.

Dudemanguy:
I added support to `find_library` now. I noticed that `static` works different in `find_library` than it does in `dependency`. In `dependency`, it's merely just a preference and always tries the fallback if it fails. However in `find_library`, explicitly setting the `static` kwarg makes it a hard value (i.e. it will either only try static or only try shared). If you just don't set anything (i.e. the default), then it acts as a prefer shared. So the only thing that was missing was a prefer static (but fallback to shared) setting. That was just another `elif` statement.

ePirat:
This would be an incredibly useful to have feature for VLC's WIP meson port, currently we clutter every dependency and find_library with the static kwarg to get this ability.

xclaesse:
Throwing just an idea: what I often want is use shared libraries from the system, and if not found build a static fallback subproject. Not sure how we could try to express that use-case.

eli-schwartz:
Perhaps the prefer_static option should cause all subprojects to default to static lookup? IIRC we already special case `dependency('foo', static: true)` to pass `default_library=static` as a subproject default option. Perhaps we should check this option there too in the case that the static kwarg is None.

xclaesse:
> IIRC we already special case dependency('foo', static: true) to pass default_library=static as a subproject default option

Yes we do that, so if I understand correctly prefer_static as proposed would already build static subprojects. However my use-case is a little bit different: I build GStreamer against a platform SDK that contains a few basic libraries (e.g. zlib, openssl) and I want to use the shared libraries provided by the platform for them. But everything else that is not part of the SDK is built as subproject and those would be nice to be static.

Dudemanguy:
>Yes we do that, so if I understand correctly prefer_static as proposed would already build static subprojects. 

Well this PR currently only touches code related to `dependency` and `find_library`. I'm not sure if that's all that is required and the subproject case automatically inherits the method or if more changes are required.

>So dependency('zlib', fallback: 'zlib') with -Dprefer_static=subprojects would return the shared lib from system, or static lib from fallback subproject.

If such an option is added, would it make sense to turn the `static` kwarg to a combo option like that as well? The main intent of this PR was simply mimic the static kwarg but on a global level, so I would think any new options should also propagate downwards.

Dudemanguy:
@xclaesse: Just revisiting this one, but if I understand correctly, would your usecase be covered if `prefer_static` is allowed to be set per-subproject like `default_library`? That way you could just set per subproject options for those to be static and everything else would do shared lookups. I can add a commit for that. It's not complicated.

As a sidenote, I finally got around to confirming that yes, subprojects do inherit the `prefer_static` argument so they all do static lookups as well (as expected).

Tachi107:
> Throwing just an idea: what I often want is use shared libraries from the system, and if not found build a static fallback subproject. Not sure how we could try to express that use-case, `-Dprefer_static=subprojects` ?

Doesn't `dependency('name', default_options: 'default_library=static')` already cover this use case?

xclaesse:
Yes, but it means editing entry dependency() calls. I would like instead a global switch for that.

xclaesse:
> Doesn't `dependency('name', default_options: 'default_library=static')` already cover this use case?

Re-thinking about this, subprojects inherit `default_library=static` from main project by default, so as long as it's fine to also have the main project built with `default_library=static` my use-case is already covered. More fine grained would still be nice, but probably out of scope for this PR.

Ultimately, we could add a list of libs/deps in native/cross file that tells exactly which one should be static.

Dudemanguy:
Anything else that this PR needs?

eli-schwartz:
No, I don't think so. And it seems we're all in agreement that this option is useful as it stands.

Dudemanguy:
Thanks!

sertonix:
Could this be added to the documentation? The reference manual doesn't mention `prefer_static` at all: https://mesonbuild.com/Reference-manual_functions.html#dependency

Dudemanguy:
It's a built-in option. You can see it here. https://mesonbuild.com/Builtin-options.html

eli-schwartz:
I have no objection to also adding a reference to the builtin option, in the documentation for `dependency()`. ;)

@sertonix would you like to submit a patch for this?

PR #9604, opened 2021-11-21
make sure files arguments to compiler.compiles and friends, performs rebuild

### LINKED ISSUE #1656: compiler object compiles() does not track changes to the files being compiled
The compiler object compiles() method does not take into account changes to the file object being compiled. For example,

```
project('foo', 'c')
cc = meson.get_compiler('c')
test = cc.compiles(files('foo.c'))
```

won't get re-evaluated if `foo.c` changes.

I presume the same applies to link() too, but did not try.

eli-schwartz:
If the compiler check is updated as a string in meson.build, we force rebuild, which is a good thing since the outcome of that check changes the configuration context and can enable or disable parts of the build.

If the compiler check came from a files() object then we didn't add a regen rule on those files.

Fixes #1656

### DIFF
--- a/mesonbuild/interpreter/compiler.py
+++ b/mesonbuild/interpreter/compiler.py
@@ -275,6 +275,7 @@ class CompilerHolder(ObjectHolder['Compiler']):
     def run_method(self, args: T.Tuple['mesonlib.FileOrString'], kwargs: 'CompileKW') -> 'RunResult':
         code = args[0]
         if isinstance(code, mesonlib.File):
+            self.interpreter.add_build_def_file(code)
             code = mesonlib.File.from_absolute_file(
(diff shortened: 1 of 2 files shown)


dcbaker (approved):

eli-schwartz:
CI fails because assertReconfiguredBuildIsNoop claims to support the VS backend, but it looks for "ClCompile:\n  All outputs are up-to-date." which apparently is not present when no compile outputs are defined... ninja, sensibly, has a status message that talks about project build artifacts being up to date rather than singling out GCC?

I don't know how to fix this, and it also seems broken. What happens if a unittest build has both MSVC compile commands and custom_target commands, does assertReconfiguredBuildIsNoop agree that the build is a noop even if the custom_targets actually get rebuilt?

PR #9607, opened 2021-11-22
Use -Oz when optimization=s in ldc2

### LINKED ISSUE #7176: Dlang: -Dbuildtype=minsize invokes -Os instead of -Oz for LDC

denizzzka:
Fix #7176


dcbaker (approved):
LGTM, thanks.

Just for reference, I've seen issues with -Oz going back to 2013, so I think it's reasonable to just always return Oz

PR #9608, opened 2021-11-22
Add symlink_function

### LINKED ISSUE #1602: RFE: add new `install_symlink` command
I have to create a bunch of symlinks, e.g. for alternate binary names. I do by with something like this:
```meson
meson.add_install_script('sh', '-c',
                         'ln -s systemd-mount $DESTDIR@0@/systemd-umount'.format(bindir))
```
This has the following shortcomings:
- requires explicit manipulation of paths and $DESTDIR
- the symlink is not removed on `uninstall` (#371 is about that part too).

I'd be great to have something like
```meson
systemd_mount = executable(..., install : true)
install_symlink(systemd_mount, bindir + '/alternate-name')
# or
install_symlink('systemd-mount', bindir + '/alternate-name')
```
where a symlink from arg2 to arg1 would be created, underneath $DESTDIR. In case an argument like `systemd_mount` which holds a reference to an executable, the install path of that executable would be used.

`install_symlink` should take a boolean keyword arg `relative`, `true` by default, to create relative symlinks (like `ln -s --relative`), even if an absolute path is specified.

The symlink should be removed on uninstallation.

pablofsf:
Allows installing symlinks directly from meson, which can
become useful in multiple scenarios. Current main use is to
help moving forward #9557 

I have implemented this as a totally unrelated to regular `install_data`. However, I saw multiple hints in the code which could possibly recommend somewhat integrating the link part into `install_data`. I guess it might fix some issues and reduce code duplication. That could be easily done by reusing `sources` and `rename` arguments. However, from the user perspective that might be totally confusing, so I decided to just do it as independent and let reviews do the rest. 

There are still some small tasks for which I need some help, hence the Draft:
 * https://github.com/mesonbuild/meson/compare/master...pablofsf:install_symlink?expand=1#diff-e7d2c475161a598f52d91577d6c100c65b455297addc5d6c35d5424a3ba2bce0R461
 * ~~https://github.com/mesonbuild/meson/compare/master...pablofsf:install_symlink?expand=1#diff-abffc4d23a79ba9faffdc902d5347cccf4d0eef69ee7503b7ad3c6f86a6fc78eR1946~~
 * Tests are quite limited, but given the code checks for the existence of the pointed file, maybe they are enough?

@eli-schwartz 

Resolves #1602


jpakkane:
One of the things that always trips me up with symlinks is working out what points where. Especially if the symlink points to a file with a different name. So something like this would be more clear:

```meson
install_symlink('foo',
    pointing_to: '../baz/fob',
    install_dir: '/something/something')
``´

This should also check that there are no path separators in the actual name. The dir part should be in `install_dir`.

**Comment by pablofsf** [comment]
> One of the things that always trips me up with symlinks is working out what points where. Especially if the symlink points to a file with a different name.
> 
> ```meson
> install_symlink('foo',
>     pointing_to: '../baz/fob',
>     install_dir: '/something/something')
> ```

Yes, that's always an issue for me too. However, I believe using the link name as a positional argument can become quite annoying in multiple circumstances. I am assuming in most of the cases people do not want/need to rename the links, but are fine with having the links installed with the same name as the basename of what they are pointing to. For example, in order to install multiple symlinks using `pointing_to`, one would have to do:

```meson
install_symlink(['foo', 'bar', 'baz'],
     pointing_to: ['../bin/foo', '../bin/bar', '../bin/baz'],
     install_dir: 'sbin')
```
Inverting the positional arguments and `pointing_to` would simplify the meson file, specially in the (I believe) more common case where the links are not renamed:

```meson
install_symlink(['../bin/foo', '../bin/bar', '../bin/baz'],
     install_dir: 'sbin')
```

So instead, what do you think about:
 * Remove the additional `targets` option and force users to send a list as a positional argument if they want multiple symlinks
 * Edit the documentation to replace instances of `targets` to `pointed files` (any better suggestion?)
 * Maybe also find a better name for `rename`?

> This should also check that there are no path separators in the actual name. The dir part should be in `install_dir`.

Makes sense, fixed!

jpakkane:
`install_symlink` should only install one symlink and trying to specify more than one should be a hard error. 

```
install_symlink(['../bin/foo', '../bin/bar', '../bin/baz'],
     install_dir: 'sbin')
```

This is confusting. I would not be able to understand what that does without reading the documentation and then the implementation code.

Whenever you have the case where doing a fairly simple thing requires noticeable brain power, the design is probably flawed in some way.

pablofsf:
> This is confusting. I would not be able to understand what that does without reading the documentation and then the implementation code.
> 
> Whenever you have the case where doing a fairly simple thing requires noticeable brain power, the design is probably flawed in some way.

That conviced me :D I changed the interface and the documentation now :)

pablofsf:
I have no clue of why the errors in windows. It looks like in linux is passing (also tested locally), but do not have a windows environment to reproduce. I guess I am doing something which is not portable?

pablofsf:
Fixed tests by skipping symlinks in windows with ninja as suggested in matrix. I still wonder if the test coverage is enough and the handling when a link already exists in place but is pointing to a different location, but I think this is ready for review :)

mesonbuild/interpreter/interpreter.py
```diff
@@ -1930,6 +1934,23 @@ def func_install_emptydir(self, node: mparser.BaseNode, args: T.Tuple[str], kwar
...
+    @typed_kwargs(
+        'install_symlink',
+        KwargInfo('pointing_to', str),
+        KwargInfo('install_dir', (str, NoneType)),
+        KwargInfo('install_tag', (str, NoneType)),
+    )
+    def func_install_symlink(self, node: mparser.BaseNode,
+                             args: T.Tuple[T.List[str]],
+                             kwargs) -> build.SymlinkData:
+        name = args[0]
```
jpakkane, line 1947: This should check that `len(args) == 1` and that `args[0]` is a string and does not contain `/` or `\`.

There should also be a `failing` test that tries to install a symlink called `foo/bar` which then fails.

mesonbuild/build.py
```diff
@@ -2795,6 +2799,18 @@ def __init__(self, sources: T.List[File], install_dir: str, install_dir_name: st
...
 
+class SymlinkData(HoldableObject):
+    def __init__(self, target: str, name: str, install_dir: str,
+                 subproject: str, install_tag: T.Optional[str] = None):
+        self.target = target
+        if name != os.path.basename(name):
+            raise InvalidArguments(f'Link name is "{n}", but link names cannot contain path separators. '
+                                   'The dir part should be in install_dir.')
+        self.name = name
+        self.install_dir = install_dir
```
tristan957, line 2810: I would rather this be a pathlib.Path object.

mesonbuild/minstall.py
```diff
@@ -441,6 +446,25 @@ def do_copyfile(self, from_file: str, to_file: str,
         append_to_log(self.lf, to_file)
         return True
 
+    def do_symlink(self, target: str, link: str, full_dst_dir: str) -> bool:
```
tristan957, line 449: Make the arguments pathlib.Path objects instead of strings where appropriate.

mesonbuild/minstall.py
```diff
@@ -441,6 +446,25 @@ def do_copyfile(self, from_file: str, to_file: str,
...
+        abs_target = target
+        if not os.path.isabs(target):
```
tristan957, line 451: Use pathlib.Path functions instead of os.path once args are converted.

mesonbuild/minstall.py
```diff
@@ -441,6 +446,25 @@ def do_copyfile(self, from_file: str, to_file: str,
...
+        if os.path.exists(link):
+            link_target = os.readlink(link)
+            if link_target == target:
+                self.log(f'Symlink {link} to {target} already exists, skipping')
+                return True
+            else:
+                # What should we do in this case? Should we consider it?
+                return False
+        self.log(f'Installing symlink {link} to {target}')
+        os.symlink(target, link, target_is_directory=os.path.isdir(abs_target))
```
tristan957, line 464: pathlib.Path has a function for creating symlinks.

mesonbuild/interpreter/interpreter.py
eli-schwartz: We can use `typed_pos_args('install_symlink', str)` for this. It won't validate that it doesn't contain path separators though.

mesonbuild/minstall.py
eli-schwartz: The entire file uses strings and os.path methods, so I would hardly consider this a PR blocker.

eli-schwartz:
Sorry for the lateness of the reply. In general the functionality looks great! I didn't spot any real issues on my first look.

> One of the things that always trips me up with symlinks is working out what points where. Especially if the symlink points to a file with a different name. So something like this would be more clear:

Meson is highly consistent in that for all such functions, the first posarg is the name of the meson object being installed (except in cases where build outputs are created via the output kwarg), so it naturally follows that `install_symlink('/path/to/foo.txt', '/path/to/bar.txt')` must install a symlink named foo.txt that points to bar.txt

That being said, I have no particular objection to implementing it as a kwarg.

I agree that multiple symlinks should require multiple invocations of the function.

mesonbuild/minstall.py
```diff
@@ -584,6 +609,16 @@ def install_data(self, d: InstallData, dm: DirMaker, destdir: str, fullprefix: s
...
             self.set_mode(outfilename, i.install_mode, d.install_umask)
 
+    def install_symlinks(self, d: InstallData, dm: DirMaker, destdir: str, fullprefix: str) -> None:
+        for s in d.symlinks:
+            if not self.should_install(s):
+                continue
+            full_dst_dir = get_destdir_path(destdir, fullprefix, s.install_path)
+            full_link_name = get_destdir_path(destdir, fullprefix, s.name)
+            dm.makedirs(full_dst_dir, exist_ok=True)
+            if self.do_symlink(s.target, full_link_name, full_dst_dir):
```
eli-schwartz, line 619: Part of the functional effects here are done in a do_symlink function, part in install_symlink. I wonder whether we actually need two functions at all? The only other secondary functions we have are do_copyfile and do_copydir, but in both cases we need to use them multiple times, so they are factored out for practical reasons.

mesonbuild/minstall.py
eli-schwartz: And we shouldn't use that!

Instead, we should use self.symlink, which handles --dry-run, and may choose to do pathlib stuff if it wants to.

mesonbuild/minstall.py
```diff
@@ -441,6 +446,25 @@ def do_copyfile(self, from_file: str, to_file: str,
...
+    def do_symlink(self, target: str, link: str, full_dst_dir: str) -> bool:
+        abs_target = target
+        if not os.path.isabs(target):
+            abs_target = os.path.join(full_dst_dir, target)
+        if not os.path.exists(abs_target):
+            raise RuntimeError(f'Tried to install symlink to missing file {abs_target}')
+        if os.path.exists(link):
+            link_target = os.readlink(link)
+            if link_target == target:
+                self.log(f'Symlink {link} to {target} already exists, skipping')
```
eli-schwartz, line 458: I don't think it's necessary to skip it, since if it exists and points to the same location then deleting and recreating it doesn't cause any harm. This is a matter of semantics, except for the next bit...

In the case where they are not the same, I'd say we can still overwrite it with the target destination we want. Rationale being, that we would overwrite a regular file, so why not overwrite a symlink too? :D

mesonbuild/minstall.py
eli-schwartz: We do in fact have some code for installing targets that handles symlinks too. Specifically shared library soname aliases. So I think it makes some sense to factor out that code into a do_symlink method, and take advantage of a single-source handling of remove + symlink + append_to_log + warn when symlink creation doesn't work on a given plaform.

docs/yaml/functions/install_symlink.yaml
```diff
@@ -0,0 +1,33 @@
...
+      Target to point the link to.
+      Can be absolute or relative and that will be respected when creating the link.
+
+  install_dir:
+    type: str
+    description: |
+      The absolute or relative path to the installation directory for the links.
+      If this is a relative path, it is assumed to be relative to the prefix.
+
+      If omitted, the directory defaults to `{datadir}/{projectname}`.
```
eli-schwartz, line 26: I'm trying to imagine how useful this default would actually be... There is nothing particular about symlinks that I can imagine, which would actually benefit from this that I can think of.

I think it would be clearer to just make the install_dir required. If someone comes up with a convincing use case for adding a default, we can always change that later, but it's difficult to drop support once we add it.

mesonbuild/interpreter/interpreter.py
pablofsf: Thank for the suggestions! Adding the test found a bug, so that was definitely a requirement... 

I used the positional argument. The path separator verification is done inside `build.SymlinkData`. I added a comment so that it is more obvious.

mesonbuild/minstall.py
pablofsf: Thanks! Moved to self.symlink.

mesonbuild/minstall.py
pablofsf: I added a last commit so that install_targets also uses the `do_symlink` function. Initially I just created it for completion. Good to know it was actually useful and not just stupid :D

mesonbuild/minstall.py
pablofsf: These two comments really simplified the function, thanks!

pablofsf:
Thank you very much to all for the comments. I added all the new changes in different commits so that they are easier to review, but I could squash them on request. Changes:
 - Use as positional argument for the symlink_name
 - Add a failing test to verify that symlink names cannot have path separators.
 - Mark `install_dir` and `pointing_to` as required parameters.
 - Rework symlink installation to replace links if exist and let `--dry-run` work by using `self.symlink`
 - Let `install_targets` use the new `do_symlinks` option and move the check for non-existent platform to `symlink` function.


@tristan957 I have not addressed your comments related to using path.Pathlib instead of plain strings. The problem is that I am really not good at python and not that acquainted either with meson, as this is barely my first contribution. I have not problem changing them, but I need a reference implementation. Unfortunately, there does not seem to be any use of Pathlib objects in neither `build.py` nor `minstall.py` which I could use. If you or anybody else can at least fix that in one or two of the other functions, then it would not be a problem for me.

mesonbuild/minstall.py
```diff
@@ -333,7 +337,13 @@ def remove(self, *args: T.Any, **kwargs: T.Any) -> None:
 
     def symlink(self, *args: T.Any, **kwargs: T.Any) -> None:
         if not self.dry_run:
-            os.symlink(*args, **kwargs)
+            try:
+                os.symlink(*args, **kwargs)
+            except (NotImplementedError, OSError):
```
eli-schwartz, line 342: Since the symlink() function, like other os module wrappers, currently just directly wraps the real function, I think maybe it's a better idea to do this try/except in do_symlink instead.

Then, you could return False if symlinking fails, skip appending to the log, and skip setting self.did_install_something

pablofsf: Makes sense. Now it looks quite obvious, sorry for the stupid mistake.

eli-schwartz: No need to apologize :) a general rule in life and in computing is that most things make sense when you're blessed with hindsight. ;)

eli-schwartz:
Alright, this looks good to me.

> I added all the new changes in different commits so that they are easier to review, but I could squash them on request.

Thanks -- reviewing was pretty easy, I see why and how you changed from the initial implementation etc.

Now that the review is complete, can you do a final squash? There should be two commits in the end -- the first commit to sort functions, and the second commit adding install_symlink together with its tests and docs.

pablofsf:
> Thanks -- reviewing was pretty easy, I see why and how you changed from the initial implementation etc.

Good to know, thanks!

> Now that the review is complete, can you do a final squash?

Squashed and hopefully quite close to being merged.

mesonbuild/minstall.py
```diff
@@ -452,6 +458,29 @@ def do_copyfile(self, from_file: str, to_file: str,
...
+        if not os.path.isabs(target):
+            abs_target = os.path.join(full_dst_dir, target)
+        if not os.path.exists(abs_target):
+            raise RuntimeError(f'Tried to install symlink to missing file {abs_target}')
+        if os.path.exists(link):
+            if not os.path.islink(link):
+                raise RuntimeError(f'Destination {link!r} already exists and is not a symlink')
+            self.remove(link)
+        if not self.printed_symlink_error:
+            self.log(f'Installing symlink {link} to {target}')
```
eli-schwartz, line 472: I upgraded a local project to use the new function, and gave it a test run. Noticed one teensy tiny little UI issue:

```
[...]
Installing doc/alpm-hooks.5 to /tmp/builddir/pkg/usr/share/man/man5
[...]
Installing doc/repo-add.8 to /tmp/builddir/pkg/usr/share/man/man8
[...]
Installing symlink /tmp/builddir/pkg/usr/share/man/man8/repo-remove.8 to repo-add.8
Installing symlink /tmp/builddir/pkg/usr/share/man/man5/pacman-hooks.5 to alpm-hooks.5
[...]
```

The link and target are swapped, IMO. The last position in each log line, "to XXX", is traditionally the on-disk node that is being created.

pablofsf: I think here again the issue raised by @jpakkane hits again, I really do not get from my own message what is the link and what is the target.

Maybe `Installing symlink {os.path.basename(link)} pointing to {target} to {full_dst_dir}`?

eli-schwartz: Maybe `Installing symlink with target {target} to {link}`?

pablofsf: I still think it makes sense to keep the `pointing to` syntax because it is more obvious and used in the function. Should `Installing symlink pointing to {target} to {link}` work?

pablofsf: Changed in last commit. Let me know if it would need further changes.

docs/yaml/functions/install_symlink.yaml
```diff
@@ -0,0 +1,33 @@
+name: install_symlink
+returns: void
+description: |
+  Installs a symbolic link to `pointing_to` target under install_dir.
```
eli-schwartz, line 4: Also add

```yaml
since: 0.61.0
```

pablofsf: Thanks, fixed now.

PR #9610, opened 2021-11-22
qt module: fix stupid copy-paste error

eli-schwartz:
As evidenced by the warning message immediately below this, I meant to write "5.15" here. As is, this will enable depfile support on too-old versions of moc.

@textshell
cf. https://github.com/mesonbuild/meson/pull/7451#issuecomment-975694152


dcbaker (approved):

textshell:
Thanks, this PR fixed the problem i was seeing.

PR #3989, opened 2018-08-07
Make python module more robust

SoapGentoo:
* Lookup is now performed according to the following order:
  1. use `pkg-config` with `PKG_CONFIG_LIBDIR=LIBPC` from python
  2. use plain `PKG_CONFIG_{LIBDIR,PATH}` from the environment
  3. try to extract the information from SYSCONFIG

The second step is necessary for relocated python installations
and cross compilation scenarios, where the value of `LIBPC` will
be wrong.

#### standard installation:
```
Found "python-2.7" via pkg-config lookup in LIBPC (/usr/lib64/pkgconfig)
```

#### If installation has been relocated:
```
"python-2.7" cannot be found in LIBPC (BROKEN/usr/lib64/pkgconfig), this is likely due to a relocated python installation
Found "python-2.7" via fallback pkg-config lookup in PKG_CONFIG_LIBDIR/PKG_CONFIG_PATH
```

#### if installation has been relocated AND pc files are hosed:
```
"python-2.7" cannot be found in LIBPC (BROKEN/usr/lib64/pkgconfig), this is likely due to a relocated python installation
"python-2.7" cannot be found via fallback pkg-config lookup in PKG_CONFIG_LIBDIR/PKG_CONFIG_PATH
Found "python-2.7" via SYSCONFIG module
```

PR #9613, opened 2021-11-22
python module: only find a pkg-config dependency from the found python

### LINKED ISSUE #9612: Python configuration via pkg-config may differ from that of found Python on path.
It is fairly easy to have two Python installations on a system, only one of which reports to `pkg-config`.  If another Python is earlier on the path, then `mod = import('python')` will find one Python, but `mod.find_installation().dependency()` will find another, because it first uses `pkg_config`, regardless of whether that matches the found Python.

The consequence is that the code then compiles and links against a different Python to the one found.

**To Reproduce**

Install Python.org Python on Windows (no pkg-config files), to `c:\Python39` (e.g. with `choco install python --version=3.9`).  Install Msys2 gcc, that comes, by default, with a version of Python, along with `pkg-config` binary, and a `python3.pc` pkg-config configuration.   Make sure `c:\Python3.9` is earlier in the PATH.

In that case, with this in `meson.build`:

```
py_mod = import('python')
py_inst = py_mod.find_installation()
py_dep = py_inst.dependency()
```

the `py_inst` will be the `c:\Python39\Python.exe` Python, but the `py_dep` variables will refer to the Msys2 Python.

**Expected behavior**

I think what should happen is that the `dependency()` call should confirm that the `pkg-config` variables do in fact refer to the `py_inst` Python, and if not, then `dependency` should use the `system` method to get dependency information.

**system parameters**
* Native build
* Windows 10
* Python 3.9
* `meson --version` : 0.60.99
* `ninja --version` : 1.10.2.git.kitware.jobserver-1

eli-schwartz:
If the found python returns None from sysconfig.get_config_var('LIBPC') then we cannot (and don't) set PKG_CONFIG_LIBDIR from it. In fact, we can virtually guarantee we won't find a PkgConfigDependency either, because any python that doesn't have a LIBPC is presumably not installed to the system pkg-config directory (maybe it's an isolated relocatable install, maybe it just doesn't have pkg-config support for who knows what reason).

Trying to find one anyway using pkg-config's builtin search paths can unexpectedly succeed, though, by finding a completely unrelated python installation installed to a system location, which isn't the one we are actually building for.

Instead, return early so that we use the system dependency class fallback.

While we are at it, add back the debug messages from #3989 which got removed.

Fixes #9612

### DIFF
--- a/mesonbuild/modules/python.py
+++ b/mesonbuild/modules/python.py
@@ -79,9 +79,18 @@ class _PythonDependencyBase(_Base):
 class PythonPkgConfigDependency(PkgConfigDependency, _PythonDependencyBase):
 
     def __init__(self, name: str, environment: 'Environment',
-                 kwargs: T.Dict[str, T.Any], installation: 'PythonInstallation'):
+                 kwargs: T.Dict[str, T.Any], installation: 'PythonInstallation',
+                 libpc: bool = False):
+        if libpc:
+            mlog.debug(f'Searching for {name!r} via pkgconfig lookup in LIBPC')
+        else:
+            mlog.debug(f'Searching for {name!r} via fallback pkgconfig lookup in default paths')
+
         PkgConfigDependency.__init__(self, name, environment, kwargs)
         _PythonDependencyBase.__init__(self, installation, kwargs.get('embed', False))
+
+        if libpc and not self.is_found:
+            mlog.debug(f'"python-{self.version}" could not be found in LIBPC, this is likely due to a relocated python installation')
 
 
 class PythonFrameworkDependency(ExtraFrameworkDependency, _PythonDependencyBase):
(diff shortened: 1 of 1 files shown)


eli-schwartz:
@matthew-brett can you confirm this fixes your python.org / MSYS2 confusion?

@SoapGentoo While trying to understand what all the module is doing in the dependency lookup, I got the impression quite a bit of it is needed for your cross compilation use case, so I'd like some confirmation that it still fits the intention of what you need, and that you still need these mlog.debug prints.

matthew-brett:
@eli-schwartz - thanks - yes - that fixes my problem.

SoapGentoo:
@eli-schwartz what if the python installation gets moved around? this is cheesy, because python isn't relocatable once the prefix has been set, but isn't extremely uncommon. By "just" using pkg-config without getting it from python itself, you can avoid a broken `LIBPC` path.

eli-schwartz:
Right, so in that case LIBPC should be something other than `None` and this PR will still handle it... I hope. Right? This is what I wanted you to test. :)

matthew-brett:
I ran into this one again today.  I hate to say it - but what do you think about merging this and seeing what happens?

eli-schwartz:
It should first be reviewed by a core maintainer that isn't the submitter... I'll try to get that rolling.

mesonbuild/modules/python.py
```diff
@@ -79,10 +79,19 @@ def __init__(self, python_holder: 'PythonInstallation', embed: bool):
 class PythonPkgConfigDependency(PkgConfigDependency, _PythonDependencyBase):
 
     def __init__(self, name: str, environment: 'Environment',
-                 kwargs: T.Dict[str, T.Any], installation: 'PythonInstallation'):
+                 kwargs: T.Dict[str, T.Any], installation: 'PythonInstallation',
+                 libpc = False):
```
dcbaker, line 83:
```suggestion
                 libpc: bool = False):
```

dcbaker (approved):
I've been kind a downer on the whole python dependency thing (it's a mess, IMO), but this isn't making it worse. Can you add the suggested type annotation and consider this r-b

PR #9615, opened 2021-11-23
install modes should not apply sticky bit to files

### LINKED ISSUE #5902: installed mode tests fail for non-root user on Solaris
```
======================================================================
FAIL: test_installed_modes (__main__.LinuxlikeTests)
----------------------------------------------------------------------
  File "run_unittests.py", line 4579, in test_installed_modes
    self.assertEqual(want_mode, found_mode[1:])
AssertionError: 'rw------T' != 'rw-------'
- rw------T
?         ^
+ rw-------
?         ^
```

```
======================================================================
FAIL: test_installed_modes_extended (__main__.LinuxlikeTests)
----------------------------------------------------------------------
  File "run_unittests.py", line 4650, in test_installed_modes_extended
    (fsobj, want_mode, found_mode)))
AssertionError: '-r--r--r-T' != '-r--r--r--'
- -r--r--r-T
?          ^
+ -r--r--r--
?          ^
 : Expected file include/rootdir.h to have mode -r--r--r-T but found -r--r--r-- instead.

```

This appears to be because [Solaris chmod(2)](https://docs.oracle.com/cd/E88353_01/html/E37841/chmod-2.html) has an additional restriction:

>        If the process is not a privileged process and the file is not a direc-
>        tory, mode bit 01000 (S_ISVTX, the sticky bit) is cleared.

eli-schwartz:
This is generally a bad idea, e.g. it causes OSError on freebsd.

It also gets ignored by solaris and thus causes unittest failures.

The proper solution is to simply reject any attempt to set this, and log a warning.

The install_emptydir function does apply the mode as well, and since it is a directory it actually does something. This is the only place where we don't reset the mode.

Although install_subdir also installs directories, and in theory it could set the mode as well, that would be a new feature. Also it doesn't provide much granularity and has mixed semantics with files. Better to let people use install_emptydir + install_subdir.

Fixes #5902

### DIFF
--- a/docs/markdown/snippets/deprecated_install_mode_sticky.md
+++ b/docs/markdown/snippets/deprecated_install_mode_sticky.md
@@ -1,0 +1,14 @@
+## various `install_*` functions no longer handle the sticky bit
+
+It is not possible to portably grant the sticky bit to a file, and where
+possible, it doesn't do anything. It is not expected that any users are using
+this functionality.
+
+Variously:
+- on Linux, it has no meaningful effect
+- on Solaris, attempting to set the permission bit is silently ignored by the OS
+- on FreeBSD, attempting to set the permission bit is an error
+
+Attempting to set this permission bit in the `install_mode:` kwarg to any
+function other than [[install_emptydir]] will now result in a warning, and the
+permission bit being ignored.
--- a/mesonbuild/interpreter/interpreter.py
+++ b/mesonbuild/interpreter/interpreter.py
@@ -1894,6 +1894,7 @@ class Interpreter(InterpreterBase, HoldableObject):
                            kwargs: 'kwargs.CustomTarget') -> build.CustomTarget:
         if kwargs['depfile'] and ('@BASENAME@' in kwargs['depfile'] or '@PLAINNAME@' in kwargs['depfile']):
             FeatureNew.single_use('substitutions in custom_target depfile', '0.47.0', self.subproject, location=node)
+        install_mode = self._warn_kwarg_install_mode_sticky(kwargs['install_mode'])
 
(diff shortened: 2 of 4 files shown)


jpakkane:
I don't know enough about non-Linux unixes to say whether this is technically a good idea or not but it is a behavioural change so there should be a release note snippet for it.

eli-schwartz:
Well, I also want to actually raise a deprecation warning when it's used anywhere but in install_emptydir, but I'm not sure the best way to do it.

Basically, this is probably not ready to merge yet.

eli-schwartz:
https://en.m.wikipedia.org/wiki/Sticky_bit

> There are two definitions: one for files, one for directories.
> 
> For files, particularly executables, superuser could tag these as to be retained in main memory, even when their need ends, to minimize swapping that would occur when another need arises, and the file now has to be reloaded from relatively slow secondary memory.[1] This function has become obsolete due to swapping optimization.

eli-schwartz:
Alrighty, time for review probably.

@alanc if you'd like to confirm the testsuite passes with this commit?

jpakkane:
Assuming the test suite passes this looks good to me.

alanc:
On Solaris 11.4, with this patch:
```
run_unittests.py::LinuxlikeTests::test_installed_modes <- unittests/linuxliketests.py PASSED [ 83%]
run_unittests.py::LinuxlikeTests::test_installed_modes_extended <- unittests/linuxliketests.py PASSED [ 84%]
```
where previously both failed.   Thanks!

eli-schwartz:
And it fixes things on FreeBSD in the CI I have pending on a local branch too. Yay!

PR #9617, opened 2021-11-23
Stop backend_startup_project from erasing the last project in sln.

lukester1975:
As mentioned here https://github.com/mesonbuild/meson/pull/9611#discussion_r754671772

### DIFF
--- a/mesonbuild/backend/vs2010backend.py
+++ b/mesonbuild/backend/vs2010backend.py
@@ -467,7 +467,7 @@ class Vs2010Backend(backends.Backend):
 
         # Put the startup project first in the project list
         if startup_idx:
-            projlist = [projlist[startup_idx]] + projlist[0:startup_idx] + projlist[startup_idx + 1:-1]
+            projlist.insert(0, projlist.pop(startup_idx))
 
         return projlist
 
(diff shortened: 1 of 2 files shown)


mesonbuild/backend/vs2010backend.py
```diff
@@ -467,7 +467,7 @@ def generate_projects(self):
 
         # Put the startup project first in the project list
         if startup_idx:
-            projlist = [projlist[startup_idx]] + projlist[0:startup_idx] + projlist[startup_idx + 1:-1]
+            projlist = [projlist[startup_idx]] + projlist[0:startup_idx] + projlist[startup_idx + 1:]
```
dcbaker, line 470: This is effectively performing the following right:
```python
first = projectlist.pop(startup_idx)
protlist.insert(0, first)
```
right? That seems so much clearer to me, and less likely to reintroduce the same problems with slicing.

dcbaker (changes requested):

mesonbuild/backend/vs2010backend.py
lukester1975: Sure, done. I was just going for the minimal change and thought 2 chars was tough to beat :)

dcbaker: thanks, looks good

dcbaker (approved):

PR #9622, opened 2021-11-23
Cleanup the dlang module and add annotations

dcbaker:
Initially I just wanted to quickly add annotations, but discovered that there was some serious duplication of code (plus some confusion on python class vs instance attributes). This starts by de-duplicating the code to find dub between the the `DubDependency` class and the `DlangModule` class. It does this using a constructor (yes, a constructor, not an initializer), which isn't a common pattern in python, but I think is pretty elegant in this situation. Then the standard type cleanups and the addition to the run mypy script.

This does not use `typed_kwargs` for the one method in the module, as the keyword arguments seem to be more freeform than structured inputs, that are just passed directly into a json file.

### DIFF
--- a/mesonbuild/modules/dlang.py
+++ b/mesonbuild/modules/dlang.py
@@ -48,8 +48,7 @@ class DlangModule(ExtensionModule):
             self.dubbin = DlangModule.class_dubbin
 
         if not self.dubbin:
-            if not self.dubbin:
-                raise MesonException('DUB not found.')
+            raise MesonException('DUB not found.')
 
     def generate_dub_file(self, state, args, kwargs):
         if not DlangModule.init_dub:
--- a/mesonbuild/modules/dlang.py
+++ b/mesonbuild/modules/dlang.py
@@ -22,11 +22,13 @@ from . import ExtensionModule
 from .. import dependencies
 from .. import mlog
 from ..mesonlib import Popen_safe, MesonException
-from ..programs import ExternalProgram
+
 
 class DlangModule(ExtensionModule):
-    class_dubbin = None
-    init_dub = False
+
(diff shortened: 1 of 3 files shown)


mesonbuild/modules/dlang.py
```diff
@@ -16,49 +16,38 @@
...
+
 
 class DlangModule(ExtensionModule):
-    class_dubbin = None
-    init_dub = False
 
-    def __init__(self, interpreter):
+    # This will always be ready since the DubDependency constructor will have
+    # been called by importing the module
+    dubbin = dependencies.DubDependency.dubbin
```
eli-schwartz, line 37:
```python
  File "/__w/meson/meson/mesonbuild/modules/dlang.py", line 37, in DlangModule
    dubbin = dependencies.DubDependency.dubbin
AttributeError: type object 'DubDependency' has no attribute 'dubbin'
```

The `__new__()` method is invoked on every instance creation, to construct a "new" copy of the class. The `__init__()` method is invoked after, to initialize the new copy of the class.

eli-schwartz: As far as I can tell, this is only ever invoked by `dependency('foo', method: 'dub')` fully invoking `functools.partial(DubDependency, name, env, kwargs)`, which we have no guarantee comes before `import('dlang')` and certainly doesn't in our example templates. It certainly doesn't happen "by importing" anything.

eli-schwartz:
I apparently didn't realize this PR existed when I submitted #10096 which added type-checking for the module but didn't touch annotations at all.

Anyway, this currently conflicts quite a bit, but could also be re-scoped to just the cleanup half of the PR.

tristan957:
Is the cleanup portion of this PR still valid?

Issue #8117, opened 2020-12-22
gnome.gtkdoc dependencies are not built during install


elmarco:
**Describe the bug**

gtk4 gtk-doc has dependencies on custom targets to generate XML from md. Those are not built, and installation fails because gtk-doc fails to find the generated files 

**To Reproduce**

Build gtk4 with gtk_doc=true, and install.

**Expected behavior**

Dependencies from gtkdoc targets should be generated when needed.

**system parameters**
* Is this a [cross build](https://mesonbuild.com/Cross-compilation.html) or just a plain native build (for the same computer)?
* No
* what operating system (e.g. MacOS Catalina, Windows 10, CentOS 8.0, Ubuntu 18.04, etc.)
* Fedora 33
* what Python version are you using e.g. 3.8.0
* 3.9.1
* what `meson --version`
* 0.55.3
* what `ninja --version` if it's a Ninja build
* 1.10.1

The workaround I found is to build the gtk4-doc manually before the installation.

elmarco:
See also https://src.fedoraproject.org/fork/elmarco/rpms/gtk4/c/698045a5fb40932f6c8c6caa9a2f696bce52a7ac

Issue #6939, opened 2020-04-10
gnome.gtk-doc() gives permission denied error in Yocto (do_install)


vishnumotghare:
Following code gives permission denied error in yocto while do_install operation.

> gnome.gtkdoc('mydoc',
   main_xml : 'main-doc.xml',
    src_dir : [
    join_paths(meson.source_root())
   ],
 fixxref_args: [
      '--html-dir=' + (gnome.gtkdoc_html_dir(meson.project_name())),
    ],
  content_files: content_files,
  install : true,
)

**Error message**

> ERROR: ['gtkdoc-mkhtml', '--path=/home/user/yocto/build-zeus/tmp/work/cortexa7t2hf-neon-fslc-linux-gnueabi/mydoc/1.0.0-r0/git/:/home/user/yocto/build-zeus/tmp/work/cortexa7t2hf-neon-fslc-linux-gnueabi/mydoc/1.0.0-r0/build/', 'mydoc', '../main-doc.xml'] failed with status 1

> Traceback (most recent call last):

> File "/home/user/yocto/build-zeus/tmp/work/cortexa7t2hf-neon-fslc-linux-gnueabi/mydoc/1.0.0-r0/recipe-sysroot-native/usr/bin/gtkdoc-mkhtml", line 51, in sys.exit(mkhtml.run(options)) File "/home/user/yocto/build-zeus/tmp/work/cortexa7t2hf-neon-fslc-linux-gnueabi/mydoc/1.0.0-r0/recipe-sysroot-native/usr/share/gtk-doc/python/gtkdoc/mkhtml.py", line 93, in run quiet] + remaining_args + [gtkdocdir + '/gtk-doc.xsl', document])

> File "/home/user/yocto/build-zeus/tmp/work/cortexa7t2hf-neon-fslc-linux-gnueabi/mydoc/1.0.0-r0/recipe-sysroot-native/usr/share/gtk-doc/python/gtkdoc/mkhtml.py", line 42, in run_xsltproc return subprocess.call(command + args)

> File "/home/user/yocto/build-zeus/tmp/work/cortexa7t2hf-neon-fslc-linux-gnueabi/mydoc/1.0.0-r0/recipe-sysroot-native/usr/lib/python3.7/subprocess.py", line 339, in call with Popen(*popenargs, **kwargs) as p:

[...]

PR #5735, opened 2019-07-29
WIP: Rewrite gtkdoc() to use custom_target() instead of script

xclaesse:
This has a few adventages:
- Easier to debug because command lines are writted in ninja file
- Source file generated is compiled as a regular executable which ensure
  exact build arguments are passed to the compiler. Passing args to
  gtkdoc-scan was error prone in the past and probably still have
  issues.
- gtkdoc binaries are found using find_program() which allows overriding
  them when using gtkdoc as subproject.

PR #9625, opened 2021-11-23
Allow building gtkdoc during build

pablofsf:
Currently `gnome.gtkdoc` builds documentation during the installation step instead of during compilation. This has led to a pile of issues related this topic (#5793, #5599, #4770, #8117, #6939, #1844 at least) and has some further consequences. 

The original reason why I looked into this was that postmarketos' build tool cannot build programs that use `gnome.gktdoc`. That is because the cross-compilation environment it sets up is not designed to compile things during install (which is what one would expect!). Obviously, this is not more important than any of the other issues, just a further example of the problems that such a behavior causes.

Additionally,  there has been one previous failed attempt (that I know of) to fix it in #5735, but  that took a more aggressive approach by removing the helper, which is blocked by other issues.

Therefore, my idea until the helper can be replaced, was to simply set `build_by_default: True`  in the `CustomTarget`. However, as discussed in #2242 and in some of the other issues, most people do not want to run the documentation build most of the time and some possibly rely on the bug to do that for them. In consequence, to be able to satisfy the long-term goal while not breaking users without prior notice, I thought of the following long-term solution, where this PR takes care of the first step:

 * In next release, add boolean `build_during_build` with `False` as default
   to `gnome.gtkdoc` and add a warning. This should not disrupt current users
   of `gtk_doc`-like options, while warning users that are relying in the bug.
 * In a future meson release, change `build_during_build` default to `True`
   and deprecate it. This might break those users that have not done the
   transition, but given there was enough time and notice, it should not be an
   issue.
 * In a further meson release, remove `build_during_build` option and warranty
   that gtk documentation is always build during `meson compile` step.

The documentation suggests a recommended way for users that rely on the bug to deal with the transition:

```meson
gnome = import('gnome')
if get_option('gtk_doc')
  gnome.gtkdoc(...)
endif
```

I have personally seen this construct in multiple projects which I have packaged, and it is also the way [gtk](https://gitlab.gnome.org/GNOME/gtk/-/blob/master/docs/reference/gsk/meson.build) and [glib](https://gitlab.gnome.org/GNOME/glib/-/blob/main/docs/reference/glib/meson.build) build their docs. Therefore, I thought it would be a reasonable default, but I have no personal preferences as long as it allows a smooth transition.

Of course, if the maintainers believe that changing the default without a transition is ok, then I  can rework this PR for that. I just thought it would have much more chances of getting merged by proposing a roadmap and avoid directly breaking some users.


mesonbuild/modules/gnome.py
```diff
@@ -1140,6 +1142,11 @@ def gtkdoc(self, state: 'ModuleState', args: T.Tuple[str], kwargs: 'GtkDoc') ->
                 raise InvalidArguments('gnome.gtkdoc: main_xml and main_xgml are exclusive arguments')
             main_file = main_xml
         moduleversion = kwargs['module_version']
+        if not kwargs['build_during_build']:
+            mlog.warning('"gnome.gtkdoc" documentation will be built during install instead of during build. '
+                         'This is discouraged and will be changed in a future meson version. '
+                         'To avoid building docs by default, wrap "gnome.gtkdoc" around a meson_option. '
+                         'If you are already doing this, this warning is safe to ignore')
```
dcbaker, line 1149: I think this is a place for `FeatureDeprecated.single_use()`, that way we give a deprecation warning, pluse then you only get the warning if you can actually fix it.

test cases/frameworks/10 gtk-doc/doc/foobar1/meson.build
```diff
@@ -6,4 +6,5 @@ gnome.gtkdoc('foobar',
   # Manually written types file for regression test:
   # https://github.com/mesonbuild/meson/issues/8744
   gobject_typesfile: 'foobar.types',
-  install : true)
+  install : true,
+  build_during_build: true)
```
dcbaker, line 10:
```suggestion
  build_during_build : true)
```

test cases/frameworks/10 gtk-doc/doc/foobar3/meson.build
```diff
@@ -3,4 +3,5 @@ gnome.gtkdoc('foobar',
   src_dir : inc,
   main_sgml : 'foobar-docs.sgml',
   content_files : [docbook, version_xml],
-  install : true)
+  install : true,
+  build_during_build: true)
```
dcbaker, line 7:
```suggestion
  build_during_build : true)
```

docs/markdown/snippets/gtkdoc_build_during_install.md
```diff
@@ -0,0 +1,31 @@
...
+reduce build time by skipping documentation builds). The right way for users
+to do this is to create a `gtk_doc` boolean meson option which conditionally
+enables building the documentation:
+
+```meson
+gnome = import('gnome')
+
+if get_option('gtk_doc')
+  gnome.gtkdoc(...)
+endif
```
dcbaker, line 18: If you wanna get really fancy, you could actually use a `feature` type option and do this (I think):
```meson
gnome = import('gnome', required : get_option('gtk_doc'), disabler : true)
gnome.gtkdoc(...)
```

dcbaker (changes requested):
I really like this change, and I really don't like the current behavior. So I'm for this.

@tp-m, @xclaesse , @nirbheek any of you want to look at this or recommend someone from gnome would would?

nirbheek:
Sounds good to me in principle, haven't looked at the code yet.

docs/markdown/snippets/gtkdoc_build_during_install.md
pablofsf: Thanks for the suggestion! I included it, but kept the previous example for people that might do more things than just running `gnome.gtkdoc`.

mesonbuild/modules/gnome.py
pablofsf: > pluse then you only get the warning if you can actually fix it.

Sorry for my ignorance about meson, but how would that actually work? The way the new option is now implemented, its default will trigger the warning/deprecation message. In consequence, **all** current users of `gnome.gtkdoc` will get the warning. I feel this is totally suboptimal, but if the goal is to deprecate the feature in the long-run, I guess it is not nice to have maintainers which are doing things right, first add and then remove `build_during_build` flag. Does that make sense? Do you have any better suggestions on how to approach this?

jpakkane:
From memory people did not like building stuff during install because when you do `sudo ninja install` the builds are done as root and write their output in the build dir with the uid of root.

eli-schwartz:
In principle I like this a lot. In practice due to build_always_stale this forces rebuilding the docs at install time.

Also I see no reason to default to false, or provide a warning. Quite frankly, making things be owned by root after install because of a *project* decision feels terribly wrong. If we want to support (in any way, shape or form) not building by default, then that should be by making the `install` target depend on additional targets other than `all`.

mesonbuild/modules/gnome.py
eli-schwartz: The Feature checks framework will consult the minimum version of meson specified in `project()`, if any, and if you currently declare a minimum version greater than the FeatureDeprecated, it will print a warning. Otherwise, it will remain silent on the theory that you currently support older versions of meson that don't support this new kwarg, so you cannot actually add it, and there is no use telling you that you should.

eli-schwartz: Anyway, the current implementation of this warning actually warns you even if you do `build_during_build: false`. You'd need to have it default to None, check if `is None`, and then print the warning + update the kwargs to specify `False`.

pablofsf:
> In principle I like this a lot. In practice due to build_always_stale this forces rebuilding the docs at install time.

I had not realized because the default in alpine is to run install with `--no-rebuild` option, but I guess there are plenty of cases where that is not the way to go. Is there any way to sensibly fix this then? Or are you suggesting to remove `build_always_stale`?

pablofsf:
Given the comments in the first review, I guess the way to implement the default and warning needs some discussion. The options I can think of:

 - Keep current default to `False` and print warning, annoying all users
 - Do not set a default and interpret that case as `False`. The warning can then be skipped by doing `build_during_build: false`. `build_during_build` will be deprecated in a future release, forcing users to do it correctly.
 - Do not set a default and interpret that case as `True`. Additionally directly deprecate `build_during_build`. This might annoy some users, as it will change the default. At least they'll have a fast way to temporary go back to previous behavior (although by using a deprecated feature).
 - Do not even create a `build_during_build` option and directly enforce doing the right thing. This might break some users, that might be annoyed. The release notes should be clear enough so that users can solve the problem without much effort.

I am not a meson developer and I do not have any specific preference on how to solve this, but I would just need an agreement from devs on how should be best solved. Of course, also let me know if I should add any other options.

pablofsf:
Rebased and removed one of the cleanup commits after #9627 has been merged

pabloyoyoista:
Rebased on top of master. Given that already somebody else tried to do the same, that #9669 isn't moving forward and some of the comments in this PR, I would suggest reconsidering this. 

I  would be willing to nuke the transition process presented in the description and simply build during build. Rationale: current behavior is clearly a bug and the more I look around and also from comments in #9669, most projects already use some sort of `doc` flag. So the transition  should be mostly unnecessary for most projects. It is still likely that the behavior change will annoy some users. But given that it is not much  likely that those users read meson Release Notes, how would a transition period help?

> In principle I like this a lot. In practice due to build_always_stale this forces rebuilding the docs at install time.

Still, I would need some suggestion on how to fix this

Issue #9492, opened 2021-10-29
meson 0.60 does not set SONAME for shared_module() anymore


Apteryks:
**Describe the bug**
After upgrading meson from 0.59.1 to 0.60.0, the network-manager build was failing on GNU Guix like so:
```
starting phase `validate-runpath'
validating RUNPATH of 9 binaries in "/gnu/store/mq8ggp0wybipvnxpwhqw71rgw1k3pkll-network-manager-1.32.12/lib"...
/gnu/store/mq8ggp0wybipvnxpwhqw71rgw1k3pkll-network-manager-1.32.12/lib/NetworkManager/1.32.12/libnm-device-plugin-bluetooth.so: error: depends on 'src/core/devices/wwan/libnm-wwan.so', which cannot be found in RUNPATH ("/gnu/store/2fk1gz2s7ppdicynscra9b19byrrr866-glibc-2.33/lib" "/gnu/store/6p74cpb86qp6h559ck96593bn2x6iq76-glib-2.70.0/lib")
/gnu/store/mq8ggp0wybipvnxpwhqw71rgw1k3pkll-network-manager-1.32.12/lib/NetworkManager/1.32.12/libnm-device-plugin-wwan.so: error: depends on 'src/core/devices/wwan/libnm-wwan.so', which cannot be found in RUNPATH ("/gnu/store/2fk1gz2s7ppdicynscra9b19byrrr866-glibc-2.33/lib" "/gnu/store/6p74cpb86qp6h559ck96593bn2x6iq76-glib-2.70.0/lib")
validating RUNPATH of 5 binaries in "/gnu/store/mq8ggp0wybipvnxpwhqw71rgw1k3pkll-network-manager-1.32.12/libexec"...
validating RUNPATH of 3 binaries in "/gnu/store/mq8ggp0wybipvnxpwhqw71rgw1k3pkll-network-manager-1.32.12/bin"...
validating RUNPATH of 1 binaries in "/gnu/store/mq8ggp0wybipvnxpwhqw71rgw1k3pkll-network-manager-1.32.12/sbin"...
error: in phase 'validate-runpath': uncaught exception:
misc-error #f "RUNPATH validation failed" () #f 
phase `validate-runpath' failed after 0.0 seconds
Backtrace:
           8 (primitive-load "/gnu/store/21szgh3rd49a9dzldpd7wr5v5w9…")
In guix/build/gnu-build-system.scm:
[...]

PR #9629, opened 2021-11-24
shared_module: Add soname when used as a link target

nirbheek:
Emit a detailed deprecation warning that explains what to do instead.
Also add a unittest.

```
DEPRECATION: target main1 links against shared module linked_module1, which is incorrect.
             This will be an error in the future, so please use shared_library() for linked_module1 instead.
             If the target requires undefined symbols, you can use `override_options: ['b_lundef=false']`.
```

Fixes https://github.com/mesonbuild/meson/issues/9492

### DIFF
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -1591,11 +1591,16 @@ You probably should put it in link_with instead.''')
             if isinstance(link_target, SharedModule):
                 if self.environment.machines[self.for_machine].is_darwin():
                     raise MesonException(
-                        'target links against shared modules. This is not permitted on OSX')
+                        f'target {self.name} links against shared module {link_target.name}. This is not permitted on OSX')
                 else:
-                    mlog.warning('target links against shared modules. This '
-                                 'is not recommended as it is not supported on some '
-                                 'platforms')
+                    mlog.deprecation(f'target {self.name} links against shared module {link_target.name}, which is incorrect.'
+                            '\n             '
+                            f'This will be an error in the future, so please use shared_library() for {link_target.name} instead.'
+                            '\n             '
+                            f'If shared_module() was used for {link_target.name} because it has references to undefined symbols,'
+                            '\n             '
+                            'use shared_libary() with `override_options: [\'b_lundef=false\']` instead.')
+                    link_target.backwards_compat_want_soname = True
                 return
 
 class Generator(HoldableObject):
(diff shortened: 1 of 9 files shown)


bonzini (approved):
Looks good to me.

eli-schwartz (approved):

jon-turney:
This also needs a documentation change to the note about this:

https://github.com/mesonbuild/meson/blob/bd9d9818711eaba18b774e17c9ce405c5020c6f9/docs/yaml/functions/shared_library.yaml#L5-L10

(which seems to have slipped from shared_module to shared_library in refman 2.0)

jon-turney:
> DEPRECATION: target main1 links against shared module linked_module1, which is incorrect.

I feel "incorrect" is  a bit vague here, but not sure what else to say...

>              This will be an error in the future, so please use shared_library() for linked_module1 instead.
>              If the target requires undefined symbols, you can use `override_options: ['b_lundef=false']`.

Perhaps something like "If `shared_module()` was used because the linked_module1 has undefined symbols, use `shared_libary()` with `override_options: ['b_lundef=false']` instead" would be clearer about the problem and solutions.

nirbheek:
> I feel "incorrect" is a bit vague here, but not sure what else to say...

Yeah I thought that keeping it brief was probably best. I remember NetworkManager uses `shared_module()` instead of `shared_library()` because `override_options:` did not exist back then, so all such uses of `shared_module()` are hacks that were always incorrect and are no longer necessary.

I'll update based on the other feedback you've given.

mesonbuild/build.py
```diff
@@ -1591,11 +1591,16 @@ def check_module_linking(self):
...
-                                 'platforms')
+                    mlog.deprecation(f'target {self.name} links against shared module {link_target.name}, which is incorrect.'
+                            '\n             '
+                            f'This will be an error in the future, so please use shared_library() for {link_target.name} instead.'
+                            '\n             '
+                            f'If shared_module() was used for {link_target.name} because it has references to undefined symbols,'
+                            '\n             '
+                            'use shared_libary() with `override_options: [\'b_lundef=false\']` instead.')
+                    link_target.backwards_compat_want_soname = True
                 return
```
floppym, line 1604: This return statement causes the workaround to fail if a target is linked against multiple shared_module targets. I think it should be removed.

floppym: As an example, see gthumb.
https://bugs.gentoo.org/827143
https://gitlab.gnome.org/GNOME/gthumb/-/blob/3.12.0/extensions/photo_importer/meson.build#L23

floppym: https://github.com/mesonbuild/meson/pull/9640

PR #9630, opened 2021-11-24
interpreter: fix `feature.require` handling of error message

benoit-pierre:
Don't show a blank error when no `error_message` was passed as argument.

### DIFF
--- a/mesonbuild/interpreter/interpreterobjects.py
+++ b/mesonbuild/interpreter/interpreterobjects.py
@@ -139,7 +139,8 @@ class FeatureOptionHolder(ObjectHolder[coredata.UserFeatureOption]):
         assert isinstance(error_message, str)
         if self.value == 'enabled':
             prefix = f'Feature {self.held_object.name} cannot be enabled'
-            prefix = prefix + ': ' if error_message else ''
+            if error_message:
+                prefix += ': '
             raise InterpreterException(prefix + error_message)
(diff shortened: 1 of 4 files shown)


mesonbuild/interpreter/interpreterobjects.py
```diff
@@ -139,7 +139,7 @@ def require_method(self, args: T.List[TYPE_var], kwargs: TYPE_kwargs) -> coredat
         assert isinstance(error_message, str)
         if self.value == 'enabled':
             prefix = f'Feature {self.held_object.name} cannot be enabled'
-            prefix = prefix + ': ' if error_message else ''
+            prefix += ': ' if error_message else ''
```
eli-schwartz, line 142:
```python
if error_message:
    prefix += ': '
```

maybe?

This code looks a bit confusing, and it seems like the first person here thought that
```python
             prefix = prefix + ': ' if error_message else ''
```

was actually 

```python
             prefix = prefix + (': ' if error_message else '')
```

I'm wondering what we can do to make it more obvious. :)

dcbaker: add the parens? :D

benoit-pierre: Amended to use:
```python
if error_message:
    prefix += ': '
```

PR #9631, opened 2021-11-25
fix BSD ldconfig handling

### LINKED ISSUE #9592: ldconfig on install corrupts freebsd cache
**Describe the bug**
2c3a68bc30cf and 6b7d43ebe0d7 make it so `ldconfig` is invoked after install. But it seems like thos corrupts the cache on FreeBSD (maybe other BSDs as well), and results in errors similar to:

    ld-elf.so.1: Shared object "libreadline.so.8" not found, required by "bash"

when running executables after `ninja install`.

Example full log: https://builds.sr.ht/~emersion/job/630186#task-tinywl

Downstream discussion: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3349#note_1162574

cc @jbeich 

**To Reproduce**
Run `ninja install` on freebsd.

**Expected behavior**
`ldconfig` cache isn't corrupted.

**system parameters**
* what operating system: FreeBSD
* what Python version are you using: N/A
* what `meson --version`: 0.60.1
* what `ninja --version`: 1.10.2

eli-schwartz:
This fix comes in two parts:
- prevent ldconfig from running in a broken fashion
- add a new feature to use ldconfig -m and add the configured libdir

Fixes #9592

/cc @jbeich, thoughts?

At least the first commit should probably go into 0.60.2

### DIFF
--- a/mesonbuild/minstall.py
+++ b/mesonbuild/minstall.py
@@ -244,7 +244,7 @@ def restore_selinux_contexts() -> None:
               'Standard output:', out,
               'Standard error:', err, sep='\n')
 
-def apply_ldconfig(dm: DirMaker) -> None:
+def apply_ldconfig(dm: DirMaker, libdir: str) -> None:
     '''
     Apply ldconfig to update the ld.so.cache.
     '''
@@ -253,6 +253,12 @@ def apply_ldconfig(dm: DirMaker) -> None:
         return
 
     if 'bsd' in platform.system().lower():
+        if libdir in dm.all_dirs:
+            proc, out, err = Popen_safe(['ldconfig', '-m', libdir])
+            if proc.returncode != 0:
+                print('Failed to apply ldconfig ...',
+                      'Standard output:', out,
+                      'Standard error:', err, sep='\n')
         return
 
     # Try to update ld cache, it could fail if we don't have permission.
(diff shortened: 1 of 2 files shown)


nirbheek:
Hum, why do we run ldconfig on libdir when there might not be anything installed in there? Shouldn't we check that one of the targets actually does install a shared library into libdir first?

nirbheek:
After discussion on Matrix, split the first commit out for the stable release: https://github.com/mesonbuild/meson/pull/9636

mesonbuild/minstall.py
```diff
@@ -243,14 +244,22 @@ def restore_selinux_contexts() -> None:
...
-def apply_ldconfig(dm: DirMaker) -> None:
+def apply_ldconfig(dm: DirMaker, libdir: str) -> None:
     '''
     Apply ldconfig to update the ld.so.cache.
     '''
     if not shutil.which('ldconfig'):
         # If we don't have ldconfig, failure is ignored quietly.
         return
 
+    if 'bsd' in platform.system().lower():
```
jbeich, line 255: DragonFly BSD doesn't have `BSD` in `uname -s` output, so this `dragonfly` here would fail to match despite inheriting `ldconfig` from FreeBSD.

eli-schwartz: Argh! And mesonlib has a different function for every BSD, but nothing to check for "is in the BSD family".

eli-schwartz: Unfortunately, this commit already landed in 0.60.2 as is. I had really wanted to wait for more confirmation, but the release was imminent...

jbeich:
`ldconfig -m $libdir` seems to work OK. Tested via sway build/install inside FreeBSD 12.2 amd64 jail.

PR #9632, opened 2021-11-25
iconv dependency: include header when checking for libc builtin

eli-schwartz:
This header is required anyway. And the compile test for linking to libc with the iconv_open symbol, can succeed when we try to use the literal symbol name without includes, but fail later during project build, because actually including iconv.h might redefine the function to match a forked symbol. This happens when GNU iconv is installed as a standalone library on systems that have a less fully-featured iconv implementation.

So, by including the header in has_function, we ensure that we test against the default resolved header. In the event that the symbol which is #define'd by the header is 'libiconv_open', linking will fail against libc even when a builtin iconv does exist, and we will fall back to the iconv dependency that provides -liconv (and which is needed to properly use the default header).

Fixes compiling against the iconv dependency on e.g. FreeBSD when the libiconv port is installed.

### DIFF
--- a/mesonbuild/dependencies/misc.py
+++ b/mesonbuild/dependencies/misc.py
@@ -452,7 +452,7 @@ class IconvBuiltinDependency(BuiltinDependency):
     def __init__(self, name: str, env: 'Environment', kwargs: T.Dict[str, T.Any]):
         super().__init__(name, env, kwargs)
 
-        if self.clib_compiler.has_function('iconv_open', '', env)[0]:
+        if self.clib_compiler.has_function('iconv_open', '#include <iconv.h>', env)[0]:
             self.is_found = True
 
 


eli-schwartz:
@Dudemanguy This should solve the mpv edge case, I think.

Dudemanguy:
>@Dudemanguy This should solve the mpv edge case, I think.

I can't really test it but less edge case handling is always welcome.

dcbaker (approved):
Your explanation seems reasonable to me.

Dudemanguy:
Hmm, it appears the additional header check somehow causes `builtin` to be found when cross compiling to windows when previously it used `system` and causes a linking error.

eli-schwartz:
@dcbaker my explanation *should* be perfectly reasonable. But it isn't. Because this:

https://github.com/mesonbuild/meson/blob/007c4659c2154755fc1f57d415afc8a736f81af2/mesonbuild/compilers/mixins/clike.py#L852-L871

As far as I can tell, on compilers without `__has_builtin` this goes straight to "is the function defined in the header? Okay, immediately return success". So if you include a header in a has_function check it always falls back to successfully answering that it's a valid builtin.

That being said, the mpv failure is occurring with `i686-w64-mingw32-gcc-posix`, which I can only assume does have `__has_builtin` but nevertheless returns that `__has_builtin(__builtin_iconv_open)` is... true? I don't understand this.

eli-schwartz:
@Dudemanguy Could you try testing this one? It really should be as simple as temporarily modifying the mpv CI to use

```
pip install 'git+https://github.com/eli-schwartz/meson@mingw-iconv'
```

Dudemanguy:
@eli-schwartz: I can confirm that one actually works. Sorry for not testing it earlier!

PR #9636, opened 2021-11-25
minstall: do not run ldconfig on the *BSDs

### LINKED ISSUE #9592: ldconfig on install corrupts freebsd cache

nirbheek:
They don't utilize a soname cache, so running ldconfig without arguments
is basically pointless -- and at least some of them are buggy: running
ldconfig with the verbose flag switches on "delete all directories from
the hints file, and recreate it using 0 command-line positional
arguments", which would soft-brick the system.

Fixes #9592

### DIFF
--- a/mesonbuild/minstall.py
+++ b/mesonbuild/minstall.py
@@ -18,6 +18,7 @@ import argparse
 import errno
 import os
 import pickle
+import platform
 import shlex
 import shutil
 import subprocess
@@ -249,6 +250,9 @@ def apply_ldconfig(dm: DirMaker) -> None:
     '''
     if not shutil.which('ldconfig'):
         # If we don't have ldconfig, failure is ignored quietly.
+        return
+
+    if 'bsd' in platform.system().lower():
         return
 
     # Try to update ld cache, it could fail if we don't have permission.


nirbheek:
Split this out from https://github.com/mesonbuild/meson/pull/9631 so we can merge it for the stable release.

PR #9640, opened 2021-11-26
Remove premature return in BuildTarget.check_module_linking()

floppym:
We want to loop over all link_targets to update
backwards_compat_want_soname if necessary.

Fixes: ec9bdc6edb17d1d9da5df2d6525025242c119f3a

### DIFF
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -1588,7 +1588,7 @@ You probably should put it in link_with instead.''')
         Warn if shared modules are linked with target: (link_with) #2865
         '''
         for link_target in self.link_targets:
-            if isinstance(link_target, SharedModule):
+            if isinstance(link_target, SharedModule) and not link_target.backwards_compat_want_soname:
                 if self.environment.machines[self.for_machine].is_darwin():
                     raise MesonException(
                         f'target {self.name} links against shared module {link_target.name}. This is not permitted on OSX')
@@ -1601,7 +1601,6 @@ You probably should put it in link_with instead.''')
                             '\n             '
                             'use shared_libary() with `override_options: [\'b_lundef=false\']` instead.')
                     link_target.backwards_compat_want_soname = True
-                return
 
 class Generator(HoldableObject):
     def __init__(self, exe: T.Union['Executable', programs.ExternalProgram],


eli-schwartz:
Makes sense, but I think in this case the repeated mlog.deprecations could get very confusing.

I think it would be easy to optimize this into a list and then warn once after leaving the for loop.

floppym:
I don't really see how repeated deprecation warnings would be _confusing_. Maybe a bit _annoying_, but isn't that the point?

Also, I don't think it makes much sense to optimize a deprecation message for a scenario that probably affects relatively few projects.

eli-schwartz:
It could be confusing to read, and figure out which ones you actually need to fix. `More lines == more annoying` isn't, IMO, an actual benefit here, not when people actually trying to fix the problem with multiple shared modules then have to fix it several times in a row to catch all instances.

"It probably doesn't matter except to relatively few projects" is a valid point...

floppym:
Taking [gthumb](https://gitlab.gnome.org/GNOME/gthumb) as an example, this deprecation warning already appears around a dozen times because they have many different targets that link against shared_module targets. This PR just adds a few more warnings to an already long list of them.

If we really wanted to print this deprecation warning once, we would need to refactor this so that the warning is emitted once per project instead of once per target. I don't think it's worth it.

eli-schwartz:
That is why I said, that reporting it once per target instead of once per dependency per target is "easy to optimize". I figured that it couldn't hurt to target some low-hanging fruit.

Now that I think about it, another possibility while we are at it is to only report once per target dependency, if multiple targets link against the same shared_module then we can change `if isinstance(link_target, SharedModule):` to `if isinstance(link_target, SharedModule) and not link_target.backwards_compat_want_soname:`

Yes, obviously the priority here is to ensure the project builds and runs successfully.

floppym:
> Now that I think about it, another possibility while we are at it is to only report once per target dependency, if multiple targets link against the same shared_module then we can change if isinstance(link_target, SharedModule): to if isinstance(link_target, SharedModule) and not link_target.backwards_compat_want_soname:

Yeah, that's a trivial change that doesn't involve messing with the message itself. I have updated the PR with this change.

eli-schwartz (approved):

PR #9645, opened 2021-11-26
Allow 'backend_startup_project' to refer to a variable.

lukester1975:
There's one last issue stopping `backend_startup_project` working for me: when a target's name is derived and not a trivial constant string, it can't be set in a project's default options.

I wondered if you'd consider allowing `backend_startup_project` to refer to a variable assigned to a target? E.g.

```
project('startup_test', ['cpp'], default_options : ['backend_startup_project=b'])
my_suffix = '_some_suffix_derived_from_environment'
a = executable('a' + my_suffix, 'main.cpp')
b = executable('b' + my_suffix, 'main.cpp')
```

Pretty simple mod attempted. I guess behaviour might change if there happened to be a variable with the same name as a later target's actual name, the earlier seen target would be selected as the default project. Pretty unlikely I'd have thought and not like it breaks anything "real"!

I haven't done a test as a) you might reject this out of hand and b) I can't see any existing ones for `backend_startup_project`. Plus c) it would have to parse a sln...

Thanks

### DIFF
--- a/docs/markdown/snippets/allow_variable_backend_startup_project.md
+++ b/docs/markdown/snippets/allow_variable_backend_startup_project.md
@@ -1,0 +1,13 @@
+## backend_startup_project can reference a target variable
+
+If the value of `backend_startup_project` is not the name of a target, it
+can be the name of a variable assigned to a target.
+
+This allows referencing a target that has an unknown name at the point
+default project options are specified:
+```
+project('startup_test', ['cpp'], default_options : ['backend_startup_project=b'])
+my_suffix = '_some_suffix_based_off_environment'
+a = executable('a' + my_suffix, 'main.cpp')
+b = executable('b' + my_suffix, 'main.cpp')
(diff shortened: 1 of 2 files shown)


jpakkane:
> a) you might reject this out of hand a

This is unfortunately the case. In Meson we aim to make things point directly to things rather than make strings that point to things that might point to things. You can still set the project you want to be the startup one from the outside with `meson configure` for every build dir separately.

lukester1975:
OK, shame. Also I'm not sure that really applies in the `backend_startup_project` case - it's not pointing at a thing (the target), but at a property of the thing. Seems like the same amount of indirection to me, just maybe in a different direction (property target ->target vs. variable -> target) :)

Sure if can be set, but it would be nice to avoid that step. Project with 10 dependencies of a single exe, say - everyone using it is going to want the exe as the startup project.

Thanks

PR #9664, opened 2021-11-26
report the context, if possible, on python tracebacks

eli-schwartz:
The interpreter tries to catch any exception and add the latest node information to it, but currently we only used that to print better formatted error messages on MesonException.

Since we should theoretically have that property for most/all exceptions, let's percolate that upward, and message the user that an unexpected traceback was encountered, that it should be reported as a bug, and the helpful information of "how far into parsing this meson.build did we get before erroring out, anyway?"

### DIFF
--- a/mesonbuild/mesonmain.py
+++ b/mesonbuild/mesonmain.py
@@ -27,7 +27,7 @@ import shutil
 from . import mesonlib
 from . import mlog
 from . import mconf, mdist, minit, minstall, mintro, msetup, mtest, rewriter, msubprojects, munstable_coredata, mcompile, mdevenv
-from .mesonlib import MesonException
+from .mesonlib import MesonException, MesonBugException
 from .environment import detect_msys2_arch
 from .wrap import wraptool
 
@@ -144,10 +144,16 @@ class CommandLineParser:
             if os.environ.get('MESON_FORCE_BACKTRACE'):
                 raise
             return 1
-        except Exception:
+        except Exception as e:
             if os.environ.get('MESON_FORCE_BACKTRACE'):
                 raise
             traceback.print_exc()
+            msg = 'Unhandled python exception'
+            if all(getattr(e, a, None) is not None for a in ['file', 'lineno', 'colno']):
+                e = MesonBugException(msg, e.file, e.lineno, e.colno) # type: ignore
+            else:
+                e = MesonBugException(msg)
+            mlog.exception(e)
             return 2
         finally:
             mlog.shutdown()


mesonbuild/mesonmain.py
```diff
@@ -144,10 +144,14 @@ def run(self, args):
             if os.environ.get('MESON_FORCE_BACKTRACE'):
                 raise
             return 1
-        except Exception:
+        except Exception as e:
             if os.environ.get('MESON_FORCE_BACKTRACE'):
                 raise
             traceback.print_exc()
+            args = ()
```
dcbaker, line 151: The typing for this is
```python
Tuple[Union[int, str], ...]
```

eli-schwartz: This file is totally untyped anyway...

But, it's actually a tuple of the form `(string, integer, integer)`.

mesonbuild/mesonmain.py
```diff
@@ -144,10 +144,14 @@ def run(self, args):
...
+            if all(getattr(e, a, None) is not None for a in ['file', 'lineno', 'colno']):
+                args = (e.file, e.lineno, e.colno) # type ignore
+            mlog.exception(MesonBugException('Unhandled python exception', *args))
```
tristan957, line 154: Capitalized `python` perhaps?

mesonbuild/mesonmain.py
dcbaker: But it's not really, it's either that or an empty tuple. But to satisfy Moog.exception what i wrote is fine and correct. Alternatively you could use a list, which is really the "correct" type here

eli-schwartz: Why is a list the correct type here? Lists are homogeneous, while each element of this variable has specific and different meaning. Changing it to a list would just lead to using unverifiable `List[Union[int, str]]` which, again, doesn't know that `[integer, string, string]` is invalid and not permitted.

The correct typing here is:

```
Union[Tuple[str, int, int], Tuple[()]]
```

And all of this seems quite irrelevant since the file isn't typed to begin with and isn't run under mypy.

dcbaker: Well, you've added `# type: ignore` below, so we might as well just do it right the first time. But yes, you could write this as `Tuple[str, int, int] | Tuple[()]`, but the splat (`*`) operator works on iterables, so any iterable is fair game, and `List[str | int]` is technically correct and conceptually simpler. maybe though it would make more sense to write as
```python
msg = 'Unhandled python exception'
if al(...)
    e = MesonBugException(msg, e.file, e.lineno, e.colno)
else:
    e = MesonBugException(msg)
mlog.exception(e)

eli-schwartz: I added that as copypasta from mlog.py :D 

if/else seems like a nice way to sidestep the problem though, and it does indeed seem clearer to read.

dcbaker (approved):
LGTM. thanks for slogging through the typing stuff with me.

PR #9666, opened 2021-11-28
gnome module: document and fix install_dir x3, by allowing false *_gir and *_typelib

eli-schwartz:
generate_gir forces building both the typelib and gir, and some people only want one or the other (probably only the typelib?) which means flagging the other as install_dir: false in the same way custom_target supports.

As this always worked, albeit undocumented, make sure it keeps working. It's pretty reasonable to allow, anyway.

Fixes https://github.com/mesonbuild/meson/pull/9484#issuecomment-980131791

### DIFF
--- a/mesonbuild/modules/gnome.py
+++ b/mesonbuild/modules/gnome.py
@@ -801,13 +801,21 @@ class GnomeModule(ExtensionModule):
     def _make_gir_target(self, state: 'ModuleState', girfile: str, scan_command: T.List[str],
                          generated_files: T.Sequence[T.Union[str, mesonlib.File, build.CustomTarget, build.CustomTargetIndex, build.GeneratedList]],
                          depends: T.List[build.Target], kwargs: T.Dict[str, T.Any]) -> GirTarget:
+        install = kwargs['install']
+
+        install_dir = kwargs['install_dir_gir']
+        if install_dir is None:
+            install_dir = os.path.join(state.environment.get_datadir(), 'gir-1.0')
+        elif install_dir is False:
+            install = False
+
         scankwargs = {
             'input': generated_files,
             'output': girfile,
             'command': scan_command,
             'depends': depends,
-            'install': kwargs['install'],
-            'install_dir': kwargs['install_dir_gir'] or os.path.join(state.environment.get_datadir(), 'gir-1.0'),
+            'install': install,
+            'install_dir': install_dir,
             'install_tag': 'devel',
             'build_by_default': kwargs['build_by_default'],
         }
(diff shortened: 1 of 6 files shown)


dcbaker:
This used to be a really common way to do this, but we stopped in favor of plain install keywords, we should follow suit here and deprecate False and add install keywords

eli-schwartz:
I added the additional kwargs, but I am a bit concerned since there is a feature request #1014 asking to back out of this entire design and use two different methods, one for gir and one for typelib.

mesonbuild/modules/gnome.py
```diff
@@ -900,8 +916,10 @@ def _get_scanner_ldflags(ldflags: T.Iterable[str]) -> T.Iterable[str]:
         KwargInfo('identifier_prefix', ContainerTypeInfo(list, str), default=[], listify=True),
         KwargInfo('include_directories', ContainerTypeInfo(list, (str, build.IncludeDirs)), default=[], listify=True),
         KwargInfo('includes', ContainerTypeInfo(list, (str, GirTarget)), default=[], listify=True),
+        KwargInfo('install_gir', (bool, NoneType), since='0.61.0'),
         KwargInfo('install_dir_gir', (str, bool, NoneType),
             validator=lambda x: 'as boolean can only be false' if x is True else None),
```
dcbaker, line 921: I think there's a deprecated_values keyword argument you can use for bool here

eli-schwartz: Nice, it works.

Sad and unfortunate, its type annotations claim it only supports `T.Dict[str, str]` which is a bit bogus as by definition `KwargInfo.deprecated_values` should accept the same keys as `KwargInfo.types`.

Using this suggestion reports the following warning:

```
meson.build:584: WARNING: Project targeting '>= 0.61.0' but tried to use feature deprecated since '0.61.0': "gnome.generate_gir" keyword argument "install_dir_gir" value "False".

WARNING: Deprecated features used:
 * 0.61.0: {'"gnome.generate_gir" keyword argument "install_dir_gir" value "False"'}
```

Using the custom FeatureNew I wrote locally...

```
meson.build:584: WARNING: Project targeting '>= 0.61.0' but tried to use feature deprecated since '0.61.0': install_dir_gir false to disable installation. Use install_gir instead

WARNING: Deprecated features used:
 * 0.61.0: {'install_dir_gir false to disable installation'}
```

What I like about that is the extra_message telling people what they *should* do.

eli-schwartz: Will probably rebase on #9687

eli-schwartz: Rebased, looks very pretty now. Thanks!

dcbaker (approved):

PR #9667, opened 2021-11-28
gnome module: fix install_dir x3, by allowing false *_gir and *_typelib

eli-schwartz:
generate_gir forces building both the typelib and gir, and some people only want one or the other (probably only the typelib?) which means flagging the other as install_dir: false in the same way custom_target supports.

As this always worked, albeit undocumented, make sure it keeps working. It's pretty reasonable to allow, anyway.

Fixes https://github.com/mesonbuild/meson/pull/9484#issuecomment-980131791

### DIFF
--- a/mesonbuild/modules/gnome.py
+++ b/mesonbuild/modules/gnome.py
@@ -746,10 +746,12 @@ class GnomeModule(ExtensionModule):
                       'depends': depends}
 
         if 'install' in kwargs:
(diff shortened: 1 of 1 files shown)


dcbaker (approved):

dcbaker:
This seems fine for 0.60, for the main branch we should add install_* keywords and deprecate this value like we've done elsewhere

PR #9668, opened 2021-11-28
properly error out when project version is an array other than files()

eli-schwartz:
Due to the support for specifying version as files('VERSION'), we need to internally accept an array, since that is what files() returns. Before that, we didn't accept arrays, and after that, we don't intend to accept generic arrays, only arrays as a side effect of files(). So tighten the typechecking to ensure that that is what we actually get.

See https://github.com/mesonbuild/meson/pull/9531#discussion_r757838111


mensinda (approved):
LGTM

PR #9669, opened 2021-11-29
gnome module: add global option to choose when to build gtkdoc HTML docs

eli-schwartz:
It is fairly common for projects to want to make installation of HTML docs optionkal. Many projects have custom options that use

if get_option('something')
    indented_gtkdoc_invocation
endif

which is messy, hard to read, and above all else, inconsistent with different names for each project. Known options include
- gtk_doc
- gtk-doc
- doc
- docs

and possibly more. This is especially problematic for yielding options to a superproject, which is an exercise in frustration when it turns out each project has a different name.

Solve all these problems by resolving the option in the module itself.

This uses the new module options infrastructure, so the option name is backwards compatible. Any existing uses of get_option will simply add an extra layer of checks, but not break the option namespace.

FIXME: linkify ticket references

Fixes:

1811 8315

Related: 7444


xclaesse:
Makes a lot of sense, yes.

I'm a bit unsure about returning Disabler object thought, all other APIs return not-found object and have an explicit `disabler: true` kwarg. But I agree it does not really match this case neither... I think it's fine as you did, because it is extremely rare to use the return value of gtkdoc() method, and in practice it's only useful as order dependency when building other docs which would be disabled too anyway. Maybe make it more explicit in meson doc that it returns a disabler object?

Ideally, but not blocker for this PR, we should also have a migration plan for projects that have `gtk_doc` option already. We already have a mechanism for deprecating project options, we should extend it for example with `option('gtk_doc', deprecated_for: 'gnome.gtkdoc')`. Some project have a boolean for that option so we could have to even support that syntax: `option('gtk_doc', deprecated_for: 'gnome.gtkdoc', deprecated: {'true': 'enabled', 'false': 'disabled'})`.

nielsdg:
I don't know how this impacts this PR, but I know some GNOME modules (like GTK) have migrated to gi-docgen, but have reused the `gtk_doc` name of the option to just build docs. 

Example in GTK: https://gitlab.gnome.org/GNOME/gtk/-/blob/main/docs/reference/meson.build#L4

mesonbuild/modules/gnome.py
```diff
@@ -1170,11 +1171,26 @@ def abs_filenames(files: T.Iterable['FileOrString']) -> T.Iterator[str]:
                   '--modulename=' + modulename,
                   '--moduleversion=' + moduleversion,
                   '--mode=' + kwargs['mode']]
+        opt = mesonlib.OptionKey('gtkdoc', module='gnome')
+        feature = state.environment.coredata.options[opt]
+        if feature.is_disabled():
+            mlog.log(mlog.bold('gnome.gtkdoc()'), 'skipped: feature', mlog.bold('gnome.gtkdoc'), 'disabled')
+            return ModuleReturnValue(Disabler(), [])
```
xclaesse, line 1178: You can return Disabler() directly, ModuleReturnValue is only needed in the case you have build targets.

docs/markdown/Builtin-options.md
```diff
@@ -267,6 +267,16 @@ Since 0.56.0 `warning_level` can also be defined per subproject.
 Some Meson modules have built-in options. They can be set by prefixing the option
 name with the module name: `-D<module>.<option>=<value>` (e.g. `-Dpython.platlibdir=/foo`).
 
+### Gnome module
+
+| Option           | Default value | Possible values          | Description |
+| ------           | ------------- | ------------------------ | ----------- |
+| gtkdoc           | enabled       | enabled, disabled, auto  | Whether to build the Gtk API html docs (Since 0.61.0) |
```
xclaesse, line 274: I think it should be `auto` by default, so it can be controlled by `-Dauto_features=disabled`.

xclaesse (commented):
Thinking out loud, and I know I contradict what I said previously, but wondering if we should instead of have generic built-in option `doc` that can be used by hotdoc and gi-docgen too. Projects could also use it with their custom doxygen targets for example. It could cause issues with projects that already define that option, but could be generally useful. What do you think?

docs/markdown/Builtin-options.md
eli-schwartz: I don't really have a strong opinion on that, but my rationale was:

Currently there is no option, and it is required.
Making it auto by default is a behavior change.

Is this behavior change actually problematic in any way? You tell me... :D

eli-schwartz:
@nielsdg on that note... is gi-docgen something that the gnome community wants to add to meson's gnome module?

Either way it will remain trivial to do:

```diff
-build_docs = get_option('gtk_doc')
+build_docs = get_option('gnome.gtkdoc')
```

Assuming one bumps the minimum version of meson to require that builtin option.

nielsdg:
> @nielsdg on that note... is gi-docgen something that the gnome community wants to add to meson's gnome module?

No idea. Currently it works well enough for me with just a `custom_target()`, but it might be good to ask someonel like @ebassi .

About the change of option name: I'm going to ping the GNOME release team on this to get their feedback, since they would be impacted the first (and most) on such a change. Personally, I don't really care for a global option for building gtkdoc

matthiasclasen:
I'm a member of a the release team, but this is just my personal opinion: I don't like this option sharing business (I guess I'm just not a fan of submodules in general).

ebassi:
> No idea. Currently it works well enough for me with just a `custom_target()`, but it might be good to ask someonel like @ebassi .

I was toying with the idea of writing a `gnome.gidocgen()`, but to be honest I haven't had time. 

Changing the name of the option is exceedingly annoying, and in any case the update to Meson will be staggered across multiple releases, because not everyone depends on the bleeding edge version of Meson. The only problem is if we bump up the version of Meson in the GNOME SDK and we suddenly start getting warnings/errors on projects using `option('gtk_doc')`.

eli-schwartz:
My understanding is that currently the name of the option is already inconsistent across a number of different naming conventions. And as I mentioned earlier, the two options (builtin and project) can coexist.

nielsdg:
I think the problem is that this PR is too late for being useful:

* most projects have either already been ported to gtkdoc (with their own name for the specific option). If they change it again, they'll end up with yet another change that needs to be propagated through several builds systems (like jhbuild, gnome-build-meta, and every Meson project that uses it as a subproject, or builds as a git submodule)
* Some projects are already switching away from gtk-doc to gi-docgen, at which point they probably lose the gtkdoc option, or _don't_ want to change it all, given the previous point

mcatanzaro:
> Thinking out loud, and I know I contradict what I said previously, but wondering if we should instead of have generic built-in option `doc` that can be used by hotdoc and gi-docgen too. Projects could also use it with their custom doxygen targets for example. It could cause issues with projects that already define that option, but could be generally useful. What do you think?

My $0.02: call the option `docs` (or `doc`). That's very simple, and agnostic to the actual documentation tool in use. I wouldn't call it `gtkdoc` because this tool has become disfavored and its use is likely to decline over the next few years.

> I'm a member of a the release team, but this is just my personal opinion: I don't like this option sharing business (I guess I'm just not a fan of submodules in general).

My $0.02: I think having a few standard options is useful to promote consistency, if used judiciously only where it makes sense to do so. A `docs` option makes sense to me.

eli-schwartz:
Refactored the preliminary cleanups into #9735

tristan957:
I feel like this isn't going to be good. Say I am contributing to GTK 3 and using a GLib subproject. I probably don't want to build the gtk-docs for GLib as well. I just want to build the docs for GTK3. Would be a huge waste of time to build the GLib docs as well.

PR #9675, opened 2021-11-30
fix broken FeatureNew checks that never printed

eli-schwartz:
They passed the arguments in the wrong order, so the version parsed as the message and the message parsed as a version.

While we are at it, pass the location node in too.


eli-schwartz:
f551de4aea1b826dd7efab02ee41207f35ef02ce is the fun commit where I somehow got mypy to tell me where we messed these up.

dcbaker (approved):
nice catch. Mypy to the rescue again :)

eli-schwartz:
(For backporting to 0.60.3 we need to drop location= since that is new API added in master only.)

Issue #9676, opened 2021-11-30
"File name too long" error when use CMake subproject


smarttowel:
I want to use [libappimageupdate](https://github.com/AppImage/AppImageUpdate) as CMake subproject, so I add next lines to my `meson.build`:
```
cmake = import('cmake')
appimage_dep = cmake.subproject('appimageupdate').dependency('libappimageupdate')
```
Wrap file:
```
[wrap-git]
revision = main
url = https://github.com/AppImage/AppImageUpdate.git
clone-recursive=true
depth=1
```
But in configuration stage next error occurred:
```
OSError: [Errno 36] File name too long: '/home/pavel/myproject/client/build/subprojects/appimageupdate/__CMake_build/lib/libappimage/libarchive-EXTERNAL-prefix/src/libarchive-EXTERNAL/configure --with-pic --disable-shared --enable-static --disable-bsdtar --disable-bsdcat --disable-bsdcpio --with-zlib --without-bz2lib --without-iconv --without-lz4 --without-lzma --without-lzo2 --without-nettle --without-openssl --without-xml2 --without-expat'
```

[full.log](https://github.com/mesonbuild/meson/files/7626096/full.log)

Meson 0.60.2

eli-schwartz:
@mensinda

But I'm going to predict this is irreparably broken. That project is attempting to download, configure, and build a ton of projects, in this case libarchive, which is a problem even without mangling the command from a cmake string into a real programming language array.

mensinda:
Please try again with a newer CMake version. The `_guess_files` code path is only relevant for old CMake (<3.17) versions where we have to do some fancy guesswork regarding spaces in file paths.

smarttowel:
Works with CMake 3.21.1, thanks.

Issue #7832, opened 2020-10-11
CMake: Dropping support for old CMake versions


mensinda:
The goal of this issue is to track when support for old CMake versions can be deprecated and dropped. Due to changes in the CMake APIs over time meson needs to maintain multiple code paths to (at least partially) support all functionality over multiple CMake versions.

The current minimum version is: **3.7**

|        API         |                             CMake version                                        |                                                        CMake status                                                     |
| :----------------- | :------------------------------------------------------------------------------: | :---------------------------------------------------------------------------------------------------------------------: |
| `--trace-expand`   | [**3.4**](https://cmake.org/cmake/help/latest/release/3.4.html#other)            | *supported* :heavy_check_mark:                                                                                          |
| Server API         | [**3.7**](https://cmake.org/cmake/help/latest/release/3.7.html#other)            | [*deprecated in 3.15*](https://cmake.org/cmake/help/latest/release/3.15.html#deprecated-and-removed-features) :warning: |
| File API           | [**3.14**](https://cmake.org/cmake/help/latest/release/3.14.html#file-based-api) | *supported* :heavy_check_mark:                                                                                          |
| `--trace-redirect` | [**3.16**](https://cmake.org/cmake/help/latest/release/3.16.html#command-line)   | *supported* :heavy_check_mark:                                                                                          |
| JSON trace         | [**3.17**](https://cmake.org/cmake/help/latest/release/3.17.html#command-line)   | *supported* :heavy_check_mark:                                                                                          |

Dropping support for the CMake Server API would mean a switch to CMake **3.14** (only for CMake subprojects, dependencies would not be affected) and would at least eliminate around 400 lines of Meson code. It might be possible to reduce the code size even more with a bit of refactoring.

Requiering the new JSON trace format (CMake **3.17**) would affect both subprojects and dependencies. The benefit would be that `traceparser.py` could be cleaned up since there would no longer be a need to guess file paths with spaces. This would also automatically give us `--trace-redirect` from **3.16** which no longer requires parsing stderr.

| Distro | CMake version |
| :- | :-: |
| Ubuntu 20.03 | 3.16 |
| Ubuntu 18.04 | 3.10 |

scivision:
I require at least CMake 3.14 in all my projects using CMake, and CMake 3.15 where Python is used. I don't get any complaints at all about that. People can `{brew,pip,snap} install cmake`.
[...]

PR #9677, opened 2021-11-30
cmake: Deprecate CMake <3.14 and warn for <3.17

mensinda:
See:

- #7832
- #9676

I did *not* deprecate CMake <3.14 support for CMake dependencies, since it makes no difference because the CMake server API is not used. The only meaningful minimum version bump for CMake deps would be to 3.17.0

### DIFF
--- a/mesonbuild/cmake/interpreter.py
+++ b/mesonbuild/cmake/interpreter.py
@@ -32,6 +32,7 @@ from functools import lru_cache
 from pathlib import Path
 import typing as T
 import re
+import textwrap
 from os import environ
 
 from ..mparser import (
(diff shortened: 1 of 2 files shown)


mesonbuild/cmake/interpreter.py
```diff
@@ -903,6 +904,19 @@ def configure(self, extra_cmake_options: T.List[str]) -> CMakeExecutor:
         if version_compare(cmake_exe.version(), '>=3.14'):
             self.cmake_api = CMakeAPI.FILE
             self.fileapi.setup_request()
+        else:
+            mlog.warning(f'Support for CMake <3.14 (Meson found {cmake_exe.version()}) is deprecated since Meson 0.61.0')
```
eli-schwartz, line 908: Shouldn't we use `mlog.deprecation` here?

mensinda: Yeah, good catch.

PR #9685, opened 2021-12-02
i18n module: correctly annotate the availability of data_dirs

eli-schwartz:
It was originally forgotten in merge_file() and later added for parity with gettext(). This means that the FeatureNew for each function is different.


dcbaker (approved):

eli-schwartz:
`vs2017 vc2017x86ninja` continues to fail, it's not a flaky test. It fails on multiple PRs so it's not a regression here (it was unlikely...) and does need to be urgently fixed separately.

PR #9687, opened 2021-12-02
Add deprecation messages to KwargInfo deprecation and since

dcbaker:
TypedKwarg currently doesn't allow an extra message to be passed when it uses FeatureDeprecated or FeatureNew, which is not commonly needed, but is needed sometimes. This series adds support for that.

### DIFF
--- a/ci/ciimage/ubuntu-rolling/install.sh
+++ b/ci/ciimage/ubuntu-rolling/install.sh
@@ -12,7 +12,6 @@ pkgs=(
   python3-pip libxml2-dev libxslt1-dev libyaml-dev libjson-glib-dev
   wget unzip
   qt5-qmake qtbase5-dev qtchooser qtbase5-dev-tools clang
-  pkg-config-arm-linux-gnueabihf
   libomp-dev
   llvm lcov
   dub ldc
--- a/.github/workflows/cygwin.yml
+++ b/.github/workflows/cygwin.yml
@@ -74,7 +74,7 @@ jobs:
       - name: Run pip
         run: |
           export PATH=/usr/bin:/usr/local/bin:$(cygpath ${SYSTEMROOT})/system32
-          python3 -m pip --disable-pip-version-check install gcovr jsonschema pefile pytest pytest-xdist coverage codecov
+          python3 -m pip --disable-pip-version-check install gcovr jsonschema pefile pytest pytest-subtests pytest-xdist coverage codecov
         shell: C:\cygwin\bin\bash.exe --noprofile --norc -o igncr -eo pipefail '{0}'
 
       - name: Run tests
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -30,7 +30,7 @@ jobs:
         python-version: '3.x'
     - run: |
         python -m pip install --upgrade pip
-        python -m pip install pytest pytest-xdist jsonschema coverage codecov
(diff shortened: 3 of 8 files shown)


eli-schwartz (approved):
I like this functionality and want to use it.

LGTM once CI passes.

dcbaker:
I love test failures that I can't reproduce locally 😞

ci/ciimage/ubuntu-rolling/install.sh
```diff
@@ -12,7 +12,6 @@ pkgs=(
   python3-pip libxml2-dev libxslt1-dev libyaml-dev libjson-glib-dev
   wget unzip
   qt5-qmake qtbase5-dev qtchooser qtbase5-dev-tools clang
-  pkg-config-arm-linux-gnueabihf
```
eli-schwartz, line 15: If this is removed in impish, won't that mean that actually trying to use it in the cross test will now fail?

```
gcc-defaults (1.193) experimental; urgency=medium

  * Bump GCC 10 version to 10.3.0.
  * Bump GCC 11 version to 11.2.0.
  * Also don't build the gccgo-doc package when not building gccgo.
    LP: #1562598.
  * d/rules: Factor-out mips* architectures into macros (Yunqiang Su).
  * Stop building the cross pkg-config packages for Ubuntu.
  * Drop references to powerpcspe.

 -- Matthias Klose <doko@debian.org>  Thu, 05 Aug 2021 09:21:17 +0200
```

Hmm...

dcbaker: it looks like some other pkg-config-cross packages are still available in impish, just not armv7

dcbaker: But yeah, that means that some of our tests are now skipping instead of running.

dcbaker: This is exactly why I don't like running multiple distros in CI, it just creates a ton of fragile moving parts

dcbaker: It doesn't seem to have actually broken anything... Which is odd

eli-schwartz: I inspected pkg-config_0.29.2-1ubuntu1.diff.gz, the modifications to Ubuntu's pkg-config package. It installs the original `/usr/share/pkg-config-crosswrapper`, and apparently also something called /usr/share/pkg-config-dpkghook which I think somehow automatically creates these wrappers for every output of `dpkg --print-foreign-architectures` in a post-install hook.

eli-schwartz: I really hate the hidden nature of these, BTW. :/

dcbaker: So does that mean that there is in fact a pkg-config anyway?

eli-schwartz: I... think so? The most reliable way to actually test this would be to install the entire environment somewhere with shell access and then actually physically check by hand. :(

dcbaker: I actually came up with another idea, I'll look at it tomorrow. Just need to compare the output of main and this PR, if the skips are the same then it's fine

dcbaker: So the answer is, there are no changes in the number of passes vs skips between this and the main branch on the ubuntu rolling cross build, so I think we're good.

PR #9689, opened 2021-12-03
don't spew pathlib internal state if $HOME is not readable

eli-schwartz:
For some reason, pathlib raises tracebacks when exists() fails due to permission errors. This is obviously wrong, and worse, unfriendly; use the sane os.path.exists() behavior and return "doesn't exist".

### DIFF
--- a/mesonbuild/dependencies/cmake.py
+++ b/mesonbuild/dependencies/cmake.py
@@ -349,8 +349,11 @@ class CMakeDependency(ExternalDependency):
         # Check the Linux CMake registry
         linux_reg = Path.home() / '.cmake' / 'packages'
         for p in [linux_reg / name, linux_reg / lname]:
-            if p.exists():
-                return True
+            try:
+                if p.exists():
+                    return True
+            except PermissionError:
+                continue
 
         return False
 


eli-schwartz:
@mensinda

mensinda:
This should really be reported upstream since I couldn't find any pathlib documentation mentioning this behavior.

Also, instead of fixing this on every call site, I would rather fix this in our [_pathlib.py](https://github.com/mesonbuild/meson/blob/master/mesonbuild/_pathlib.py) module.

eli-schwartz:
As stated, we do not need to fix it on every call site, just the ones that use $HOME but aren't under our control.

Also I morally object to our [_pathlib.py](https://github.com/mesonbuild/meson/blob/master/mesonbuild/_pathlib.py) module.

mensinda:
> As stated, we do not need to fix it on every call site, just the ones that use $HOME but aren't under our control.

As of today, that we know of.

> Also I morally object to our _pathlib.py module

Well, and I really don't want workarounds for `pathlib` bugs spread all around the Meson code base. Even if it is just this one now, with the next bug (`from a quick glance over the `pathlib` code, it looks like the `is_*()` functions should also all be affected) we will already have two places we need to track.

> since it has long since outlived its stated purpose of "backporting fixes from newer versions of CPython".

Are all the bugs fixed in *all* the Python versions we still support (3.6)?

As a compromise, could you add a BIG FAT comment describing why this code workaround exists, when the workaround should be deleted and why it is not in `_pathlib.py`?

And then if any other cases are found, can we revisit the idea of adding another workaround to `_pathlib.py`?

eli-schwartz:
> Are all the bugs fixed in all the Python versions we still support (3.6)?

It's progressed from "backporting" to "monkeypatching because the upstream code sucks", apparently. :p

The thing is, pathlib is nice because it automatically handles os.Pathlike objects transparently and nicely. It is considerably less nice in the divergences from `os.path.*` replacement functions, and this isn't the first or the second time that that bit us.

eli-schwartz:
> since I couldn't find any pathlib documentation mentioning this behavior

pathlib formally documents this using the language " Many of these methods can raise an OSError if a system call fails (for example because the path doesn’t exist)".

Of course, exists() is a function whose sole purpose is to check whether a system call succeeds or raises ENOENT, and the implementation silently permits some other system call errors, but doesn't document which ones, or what the criteria are.

Basically, what I'm getting at here is that all concrete path methods provided by pathlib may mysteriously fail at any time, and users are expected to except OSError and enumerate the system call values they are and are not comfortable with, in order to get any control whatsoever here. There is NO method of checking "is this file available to me for use", like os.path.exists did.

It's intentional design, there is a bug report from 2005 bikeshedding over whether exists() "should" return false when it errors out and cannot detect a path, and when pathlib came along, they declared that it isn't obligated to have the same semantics as os.path (hence, I argue the concrete path methods aren't a suitable replacement for os.path).

PR #9691, opened 2021-12-03
cuda: allow dependency generation on recent cuda toolkits

smibarber:
CUDA toolkit >= 10.1.168 (aka 10.1 update 1) has an nvcc that supports
dependency generation.

### DIFF
--- a/mesonbuild/compilers/cuda.py
+++ b/mesonbuild/compilers/cuda.py
@@ -21,7 +21,7 @@ from .. import coredata
 from .. import mlog
 from ..mesonlib import (
     EnvironmentException, MachineChoice, Popen_safe, OptionOverrideProxy,
-    is_windows, LibType, OptionKey,
+    is_windows, LibType, OptionKey, version_compare
 )
 from .compilers import (Compiler, cuda_buildtype_args, cuda_optimization_args,
                         cuda_debug_args, CompileCheckMode)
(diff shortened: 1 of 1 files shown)


eli-schwartz:
@obilaniu

obilaniu:
I once attempted this. I just want to highlight here that `nvcc` doesn't do `-MQ quoted_target` (only `-MT unquoted_target`, which is what's being proposed here again), and that was the main blocker for me because I didn't see an obvious workaround that didn't involve a full-scale wrapper around `nvcc` invocations.

@eli-schwartz I don't know how Meson determines target names, but if they are allowed to contain Make-special characters likes spaces or punctuation, this will produce a `.d` depfile that will fail to parse.

eli-schwartz:
One simple solution for that would be for get_dependency_gen_args to check for such issues, and if they are present refrain from emitting depfiles. Perhaps with a warning!

I can't remember if such things are allowed, I never tried them before ;) so I'd have to try with a test meson.build -- my gut feeling is that meson doesn't explicitly check for any such thing.

obilaniu:
@eli-schwartz If the proposed target name happens to be, or include, the path from the root, and along this path there is a directory name with unacceptable characters, then _everything below that directory_ will (silently?) fail to generate a depfile.

I would have loved some way of executing under Ninja more than just one command per rule in order to switch `-MT original_value` for `-MT <placeholder>` and then sub out the placeholder for the desired name, but Ninja is what it is.

smibarber:
Thanks for the context - I didn't realize there was an earlier attempt and that this is still fundamentally broken without `-MQ` :(

Luckily my `nvcc` exposure is minimal so I'll just live with `ninja clean && ninja`.

PR #9696, opened 2021-12-04
Fix performance regression in build file generation

### LINKED ISSUE #9695: Generation of build files takes very long time
**Describe the bug**
The generation of build files in our project takes 2 minutes in meson 0.60.2 compared to 3 seconds in meson 0.60.1. Using git bisect, the the performance regression was introduced in commit 73bac14c8fe7279037433b15b6c15bc93e9156a1.

**To Reproduce**
Cannot include any meson.build files since it is proprietary project. However, by debugging the involved meson code in the Backend class, the following characteristics was found:
* Number of calls to Backend.determine_rpath_dirs with a build.BuildTarget: 206
* Number of calls to Backend.rpaths_for_bundled_shared_libraries with a build.BuildTarget: 2434034

**Expected behavior**
The time to generate build files takes about the same as in meson 0.60.1.

**system parameters**
* what operating system (e.g. MacOS Catalina, Windows 10, CentOS 8.0, Ubuntu 18.04, etc.): Arch Linux (updated 2021-12-02)
* what Python version are you using e.g. 3.8.0: Python 3.9.9
* what `meson --version`: Meson 0.60.2
* what `ninja --version` if it's a Ninja build; Ninja 1.10.2

complexoctonion:
Re-use any already determined rpaths for a target.

Fixes #9695

### DIFF
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -748,7 +748,8 @@ class Backend:
                         raise MesonException(f'Invalid arg for --just-symbols, {dir} is a directory.')
         return dirs
 
-    def rpaths_for_bundled_shared_libraries(self, target: build.BuildTarget, exclude_system: bool = True) -> T.List[str]:
+    @lru_cache(maxsize=None)
+    def rpaths_for_bundled_shared_libraries(self, target: build.BuildTarget, exclude_system: bool = True) -> 'ImmutableListProtocol[str]':
         paths: T.List[str] = []
         for dep in target.external_deps:
             if not isinstance(dep, (dependencies.ExternalLibrary, dependencies.PkgConfigDependency)):


dcbaker:
Have you considered using functools.lru_cache? That's what we usually do

complexoctonion:
Thank you for the tip about lru_cache, I will change to that. I have very little experience with python, so it was really just a naive attempt at a fix.

What about the failing "mesonbuild.meson (vs2017 vc2017x86ninja)"? Did the caching cause that?

dcbaker:
No, i think that's caused by rustc being updated to 1.57, but I haven't had s chance to investigate yet

eli-schwartz:
Indeed, that Windows test has been failing across all PRs.

The lint_mypy test however is correctly complaining that your PR adds new code which is lacking type annotations. If you switch to https://docs.python.org/3/library/functools.html#functools.lru_cache then you don't need to create your own handmade memoize caching dict, and you don't need to type-annotate its expected contents. :D

complexoctonion:
Just want to check one thing that I wasn't 100% sure about. I made the assumption that a build.BuildTarget object is immutable, at least with respect to the rpath information, after construction. Otherwise the caching in my proposed fix is not valid. Also, even if that is the case now, what are the guarantees for that not changing in the future? Is it simply that it would not make sense to make BuildTarget objects mutable after construction?

eli-schwartz:
I do not believe the aspects which this function uses should ever logically change.

dcbaker:
Targets are supposed to be immutable after creation, in practice that doesn't always happen, as you'be noted.

mesonbuild/backend/backends.py
```diff
@@ -748,6 +748,7 @@ def get_external_rpath_dirs(self, target: build.BuildTarget) -> T.Set[str]:
                         raise MesonException(f'Invalid arg for --just-symbols, {dir} is a directory.')
         return dirs
 
+    @lru_cache(maxsize=None)
     def rpaths_for_bundled_shared_libraries(self, target: build.BuildTarget, exclude_system: bool = True) -> T.List[str]:
```
dcbaker, line 752: Can you change the return type of this to `ImmutableListProto[str]`? This tells the static type checker that the return of this shouldn't be mutated

eli-schwartz:
Right, e.g. in theory `self.rpath_dirs_to_remove` is tracked data added by the backend to an existing target, in practice that cannot affect the calculation of which directories are rpaths to *add*.

mesonbuild/backend/backends.py
eli-schwartz: Basically the same thing done by the `get_custom_target_provided_by_generated_source` and `get_custom_target_provided_libraries` functions.

Note: it is ImmutableListProtocol with a "col" at the end. :p

PR #9700, opened 2021-12-05
Clean up gnome mkenums, and use typed_kwargs

dcbaker:
this makes two sets of distinct changes
1. it cleans up mkenums and mkenums_simple so that they don't call into each other
2. it uses typed_kwargs for both of them

This is currently based on #9594, and that should be reviewed and merged first.

This does not fix all the type issues in the gnome module, I'll follow this up with another series to address that separately, but this does get us complete typed_pos_args and typed_kwargs coverage

### DIFF
--- a/mesonbuild/modules/gnome.py
+++ b/mesonbuild/modules/gnome.py
@@ -754,7 +754,7 @@ class GnomeModule(ExtensionModule):
 
         return ret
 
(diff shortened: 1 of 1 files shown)


dcbaker:
Anyone planning to look at this, or should I just merge it?

PR #9703, opened 2021-12-07
docs: Add the `configuration` keyeword argument to `configure_File`

dcbaker:
We were missing the most important keyword argument of them all!

### DIFF
--- a/docs/yaml/functions/configure_file.yaml
+++ b/docs/yaml/functions/configure_file.yaml
@@ -40,6 +40,13 @@ kwargs:
       argument, see [[custom_target]] for details about string
       substitutions.
 
+  configuration:
+    type: "cfg_data | dict[str | int | bool]"
+    description: |
+      As explained above, when passed this will provide the replacement
+      data for the input file (if provided) or key value pairs to be
+      written to the output.
+
   copy:
     type: bool
     default: false


docs/yaml/functions/configure_file.yaml
```diff
@@ -39,6 +39,13 @@ kwargs:
       file objects and more than one file can be passed to the `input` keyword
       argument, see [[custom_target]] for details about string
       substitutions.
+      
+  configuration:
+    type: "[[@cfg_data]] | Dict"
```
eli-schwartz, line 44: types don't take markdown `[[@obj]]` references, also there is no such type as `Dict` (but there is `dict`).

eli-schwartz (changes requested):

dcbaker:
Edited online, so we'll need to squash commit this

docs/yaml/functions/configure_file.yaml
eli-schwartz: The type itself is cfg_data, not configuration_data, it just needs to lose the markdown parts (`[[@` and `]]`) because the yaml is already strongly typed so it takes `obj`.

mensinda: Not sure how I missed this, but I guess here we are now... It seems like I even missed that `cfg_data` is returned by the `configuration_data` function, and not a dict...

Also, this should only be `dict[str | bool | int]`.

eli-schwartz:
Squashed and corrected commit message spelling: `keyeword` -> `keyword`.

eli-schwartz:
Humorously enough, it seems that since this PR was created from a branch in the repo the CI successfully updated the website from the state of the open PR.

PR #9704, opened 2021-12-07
Fully type check the gnome module

dcbaker:
Finally, with these changes (and the ones in #9700), the gnome module has become type safe, and with it all callers of CustomTarget, so I can finally get to the work I actually wanted to, fixing the layering violation between the build and interpreter modules!

### DIFF
--- a/mesonbuild/modules/gnome.py
+++ b/mesonbuild/modules/gnome.py
@@ -30,7 +30,7 @@ from .. import build
 from .. import interpreter
 from .. import mesonlib
 from .. import mlog
-from ..build import BuildTarget, CustomTarget, CustomTargetIndex, GeneratedList, InvalidArguments
+from ..build import BuildTarget, CustomTarget, CustomTargetIndex, Executable, GeneratedList, InvalidArguments
 from ..dependencies import Dependency, PkgConfigDependency, InternalDependency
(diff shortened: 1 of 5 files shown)


nirbheek:
Let's merge this immediately after 0.61.1?

mesonbuild/modules/gnome.py
```diff
@@ -1048,7 +1051,7 @@ def generate_gir(self, state: 'ModuleState', args: T.Tuple[T.List[T.Union[build.
         srcdir = os.path.join(state.environment.get_source_dir(), state.subdir)
         builddir = os.path.join(state.environment.get_build_dir(), state.subdir)
 
-        depends: T.List[T.Union['FileOrString', build.GeneratedTypes, build.Executable, build.SharedLibrary, build.StaticLibrary]] = []
+        depends: T.List[T.Union['FileOrString', 'build.GeneratedTypes', build.BuildTarget]] = []
```
eli-schwartz, line 1081: How did this ever work? Weird. build.GeneratedTypes is a typechecking-only variable but it never produced undefined variable errors when running live?

eli-schwartz: It seems that the answer to this is: https://www.python.org/dev/peps/pep-0526/#runtime-effects-of-type-annotations

A variable inside a function doesn't get stored in `__annotations__` of the function, unlike class or module variables, and in fact isn't even evaluated...

PR #9705, opened 2021-12-08
Finish adding typing to mesonbuild.{programs,minit}

tristan957:
It is unclear to me how this ever passed mypy. The lack of typing was
pretty abundant. Maybe it is just pylance acting different than mypy

### DIFF
--- a/mesonbuild/minit.py
+++ b/mesonbuild/minit.py
@@ -85,7 +85,7 @@ def autodetect_options(options: 'argparse.Namespace', sample: bool = False) -> N
         # The rest of the autodetection is not applicable to generating sample projects.
         return
     if not options.srcfiles:
-        srcfiles = []
+        srcfiles: T.List[Path] = []
         for f in (f for f in Path().iterdir() if f.is_file()):
             if f.suffix in LANG_SUFFIXES:
                 srcfiles.append(f)
@@ -176,7 +176,7 @@ def run(options: 'argparse.Namespace') -> int:
             print('Build directory already exists, deleting it.')
             shutil.rmtree(options.builddir)
         print('Building...')
-        cmd = mesonlib.get_meson_command() + [options.builddir]
+        cmd = T.cast(T.List[str], mesonlib.get_meson_command()) + [options.builddir]
         ret = subprocess.run(cmd)
         if ret.returncode:
             raise SystemExit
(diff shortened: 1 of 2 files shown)


tristan957:
cc @dcbaker

mesonbuild/minit.py
```diff
@@ -176,7 +176,7 @@ def run(options: 'argparse.Namespace') -> int:
             print('Build directory already exists, deleting it.')
             shutil.rmtree(options.builddir)
         print('Building...')
-        cmd = mesonlib.get_meson_command() + [options.builddir]
+        cmd = T.cast(T.List[str], mesonlib.get_meson_command()) + [options.builddir]
```
dcbaker, line 179: I wouldn't cast here, it's unlikely this is null, but it could be. I'd probably do
```python
cmd = mesonlib.get_meson_command()
assert cmd is not None, 'for mypy'
cmd = cmd + [options.builddir]  # to avoid mutating cmd
```

mesonbuild/programs.py
```diff
@@ -46,7 +46,7 @@ def __init__(self, name: str, command: T.Optional[T.List[str]] = None,
         self.path: T.Optional[str] = None
         self.cached_version: T.Optional[str] = None
         if command is not None:
-            self.command = mesonlib.listify(command)
+            self.command: T.List[str] = mesonlib.listify(command)
```
dcbaker, line 49: This shouldn't need an annotation, mypy should know that command `List[str]` at this point

mesonbuild/programs.py
```diff
@@ -114,6 +115,7 @@ def get_version(self, interpreter: 'Interpreter') -> str:
             if not match:
                 raise mesonlib.MesonException(f'Could not find a version number in output of {raw_cmd!r}')
             self.cached_version = match.group(1)
+        assert self.cached_version
```
dcbaker, line 118: are you asserting this isn't an empty string, or that it isn't None?

dcbaker (changes requested):
Mostly looks good, there's a couple of things that look off to me though.

dcbaker:
Yeah, pylance and mypy are different, In my experience it's not uncommon for one to figure out what's going on, and the other to not.

mesonbuild/programs.py
```diff
@@ -266,7 +268,7 @@ def _search_windows_special_cases(self, name: str, command: str) -> T.List[T.Opt
             commands = self._shebang_to_cmd(command)
             if commands:
                 return commands
-            return [None]
```
dcbaker, line 269: You may find yourself biting off more than you want to chew with this one, unfortunately

PR #9706, opened 2021-12-08
mesonlib: make get_meson_command return an ImmutableList

dcbaker:
We really don't want to be mutating global state, so let mypy yell at us
when we try.


dcbaker:
@tristan957, I noticed this looking at your PR

tristan957:
Thanks for fixing this.

dcbaker:
Any objections to me merging this?

PR #9717, opened 2021-12-10
wrap: Add support for applying a list of patch files

pwmarcz:
As discussed in #4570, I'd like to submit an updated patch. Since I'm building on the work by @xclaesse, I added him in `Co-authored-by`.

I made the following changes compared to the original PR:

* hardcode `-p1` instead of configuring strip level: this should be good for most workflows (e.g. patches generated from Git), and if the user has an incompatible patch, they can always convert or re-generate it
* instead of a separate section, use a single property, `diff_files`:
  * "diff" instead of "patch" since "Meson build patches" unfortunately already mean something else
  * I proposed `diff_filename` previously, for consistency with `patch_filename`; but it looks awkward when used for multiple patches (and `diff_filenames` could be an annoying source of typos)
* parse a comma-separated list (instead of Python list literal): this is a common convention for INI files, and already used in wrap files (`dependency_names`, `program_names`)
* prefer using `patch` to Git: `patch` is the purpose-specific tool, and one usually used in build scripts; Git requires extra workarounds (`--work-tree`)
* don't use `git am` for `wrap-git`, but apply patches in the same way as everything else: this is simpler, less prone to failure (no special rarely-taken case in code), and less surprising for the user (who probably doesn't expect Meson to apply patches as commits in this one case; and might reasonably be expected to provide a "bare" patch without a commit message)
* don't apply patches with `--ignore-whitespace`: I think it's better if we fail when the patch doesn't apply exactly, as it could indicate problems (e.g. patch for a different version of code); since the user controls the patch, they can re-generate it if necessary

Please let me know if I should change anything. The above choices seem good to me as a potential user of the feature, but I'm pretty new to the Meson project.

Co-authored-by: Xavier Claessens <xavier.claessens@collabora.com>

### DIFF
--- a/docs/markdown/Wrap-dependency-system-manual.md
+++ b/docs/markdown/Wrap-dependency-system-manual.md
@@ -84,6 +84,8 @@ previously reserved to `wrap-file`:
 - `patch_directory` - *Since 0.55.0* Overlay directory, alternative to `patch_filename` in the case
   files are local instead of a downloaded archive. The directory must be placed in
   `subprojects/packagefiles`.
+- `diff_files` - *Since 0.63.0* Comma-separated list of local diff files (see
+  [Diff files](#diff-files) below).
 
 ### Specific to wrap-file
 - `source_url` - download url to retrieve the wrap-file source archive
@@ -146,6 +148,37 @@ Prior to *0.55.0* Meson build patches were only supported for
 wrap-file mode. When using wrap-git, the repository must contain all
 Meson build definitions. Since *0.55.0* Meson build patches are
 supported for any wrap modes, including wrap-git.
+
+## Diff files
+
+*Since: 0.63.0*
+
+You can also provide local patch files in `diff` format. For historic reasons,
+they are referred to as "diff files", since the "patch" name is already used for
+overlay archives.
+
+The diff files are described by the `diff_files` property (a comma-separated
+list), and must be available locally in the `subprojects/packagefiles`
+directory.
+
+Meson will apply the diff files after extracting or cloning the project, and
+after applying the overlay archive (`patch_*`). For this feature, the `patch` or
+`git` command-line tool must be available.
+
+The diff files will be applied with `-p1`, i.e. treating the first path
+component as a prefix to be stripped. This is the default for diffs produced by
+Git.
+
+```ini
+[wrap-file]
+directory = libfoobar-1.0
+
+source_url = https://example.com/foobar-1.0.tar.gz
+source_filename = foobar-1.0.tar.gz
+source_hash = 5ebeea0dfb75d090ea0e7ff84799b2a7a1550db3fe61eb5f6f61c2e971e57663
+
+diff_files = libfoobar-1.0/0001.patch, libfoobar-1.0/0002.patch
+```
 
 ## `provide` section
 
--- a/mesonbuild/wrap/wrap.py
+++ b/mesonbuild/wrap/wrap.py
@@ -33,7 +33,7 @@ import textwrap
 from pathlib import Path
 from . import WrapMode
 from .. import coredata
-from ..mesonlib import quiet_git, GIT, ProgressBar, MesonException, windows_proof_rmtree
+from ..mesonlib import quiet_git, GIT, ProgressBar, MesonException, windows_proof_rmtree, Popen_safe
 from ..interpreterbase import FeatureNew
 from ..interpreterbase import SubProject
 from .. import mesonlib
@@ -54,6 +54,8 @@ SSL_WARNING_PRINTED = False
 WHITELIST_SUBDOMAIN = 'wrapdb.mesonbuild.com'
 
 ALL_TYPES = ['file', 'git', 'hg', 'svn']
+
+PATCH = shutil.which('patch')
 
 def whitelist_wrapdb(urlstr: str) -> urllib.parse.ParseResult:
     """ raises WrapException if not whitelisted subdomain """
@@ -101,6 +103,7 @@ class PackageDefinition:
         self.values = {} # type: T.Dict[str, str]
         self.provided_deps = {} # type: T.Dict[str, T.Optional[str]]
         self.provided_programs = [] # type: T.List[str]
+        self.diff_files = [] # type: T.List[Path]
         self.basename = os.path.basename(fname)
         self.has_wrap = self.basename.endswith('.wrap')
         self.name = self.basename[:-5] if self.has_wrap else self.basename
@@ -162,6 +165,15 @@ class PackageDefinition:
             raise WrapException(f'{self.wrap_section!r} is not a valid first section in {self.basename}')
         self.type = self.wrap_section[5:]
         self.values = dict(config[self.wrap_section])
+        if 'diff_files' in self.values:
+            FeatureNew('Wrap files with diff_files', '0.63.0').use(self.subproject)
+            for s in self.values['diff_files'].split(','):
+                path = Path(s.strip())
+                if path.is_absolute():
+                    raise WrapException('diff_files paths cannot be absolute')
+                if '..' in path.parts:
+                    raise WrapException('diff_files paths cannot contain ".."')
+                self.diff_files.append(path)
 
     def parse_provide_section(self, config: configparser.ConfigParser) -> None:
         if config.has_section('provide'):
@@ -349,6 +361,7 @@ class Resolver:
                     raise WrapException(f'Unknown wrap type {self.wrap.type!r}')
             try:
                 self.apply_patch()
+                self.apply_diff_files()
             except Exception:
                 windows_proof_rmtree(self.dirname)
                 raise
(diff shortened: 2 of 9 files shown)


mesonbuild/wrap/wrap.py
```diff
@@ -587,6 +595,29 @@ def apply_patch(self) -> None:
                 raise WrapException(f'patch directory does not exist: {patch_dir}')
             self.copy_tree(src_dir, self.dirname)
 
+    def apply_diff_files(self) -> None:
+        for filename in self.wrap.diff_files:
+            mlog.log(f'Applying diff file "{filename}"')
+            path = Path(self.wrap.filesdir) / filename
+            if not path.exists():
+                raise WrapException(f'Diff file "{path}" does not exist')
+            if False and PATCH:
```
xclaesse, line 604: `False and PATCH` ??

docs/markdown/Wrap-dependency-system-manual.md
```diff
@@ -147,6 +149,37 @@ wrap-file mode. When using wrap-git, the repository must contain all
 Meson build definitions. Since *0.55.0* Meson build patches are
 supported for any wrap modes, including wrap-git.
 
+## Diff files
+
+*Since: 0.61.0*
+
+You can also provide local patch files in `diff` format. For historic reasons,
+they are referred to as "diff files", since the "patch" name is already used for
+Meson build patches.
```
eli-schwartz, line 158: Due to precisely this confusion I prefer to avoid the name "build patch" and use "overlay archive".

docs/markdown/Wrap-dependency-system-manual.md
```diff
@@ -147,6 +149,37 @@ wrap-file mode. When using wrap-git, the repository must contain all
...
+
+The diff files are described by the `diff_files` property (a comma-separated
+list), and must be available locally in the `subprojects/packagefiles`
+directory.
```
eli-schwartz, line 162: They are searched for at the top level? If applying more than one, I guess paths with directory components would be used for grouping?

docs/markdown/Wrap-dependency-system-manual.md
```diff
@@ -147,6 +149,37 @@ wrap-file mode. When using wrap-git, the repository must contain all
...
+
+Meson will apply the diff files after extracting or cloning the project, and
+after applying the build patches (`patch_*`). For this feature, `patch` or `git`
```
eli-schwartz, line 165: "overlay archive" :p

Missing "the" before "patch or git".

docs/markdown/Wrap-dependency-system-manual.md
```diff
@@ -147,6 +149,37 @@ wrap-file mode. When using wrap-git, the repository must contain all
...
+command-line tool must be available.
+
+The diff files will be applied with `-p1`, i.e. treating the first path
+component as prefix to be stripped. This is the default for diffs produced by
```
eli-schwartz, line 169: "a prefix"

mesonbuild/wrap/wrap.py
```diff
@@ -587,6 +595,29 @@ def apply_patch(self) -> None:
...
+                cmd = [PATCH, '-f', '-p1', '-i', str(path)]
+            elif GIT:
+                # If the `patch` command is not available, fall back to `git
+                # apply`. The `--work-tree` is necessary in case we're inside a
+                # Git repository: by default, Git will try to apply the patch to
+                # the repository root.
+                cmd = [GIT, '--work-tree', '.', 'apply', '-p1', str(path)]
```
xclaesse, line 611: In the case it's a wrap-git, I'm wondering if we should use `git am` instead?

mesonbuild/wrap/wrap.py
eli-schwartz: Huh?

pwmarcz: Leftover from debugging, sorry. Fixed.

eli-schwartz (changes requested):
I'm a bit nervous about adding support to modify upstream code, as bugs should generally be fixed upstream or by committing to forking, not by vendoring.

(For example, if this is added to meson it will be completely forbidden in the WrapDB.)

Leaving that aside for now and just looking at the implementation...

mesonbuild/wrap/wrap.py
pwmarcz: As I said above, I believe it's better not to have it:
* This is surprising for the user: in this one case, the feature would do a different thing (not just apply patches to worktree, but also create commits); it also introduces an extra requirement for the patches to be in `git format-patch` format
* Special cases like this one are more likely to break (since they're less used).

If you disagree, I can add it back.

mesonbuild/wrap/wrap.py
eli-schwartz: Oh, apparently this was already fixed.

pwmarcz: Leftover from debugging, sorry. Fixed.

mesonbuild/wrap/wrap.py
eli-schwartz: I agree we should NOT use `git am`.

docs/markdown/Wrap-dependency-system-manual.md
pwmarcz: Changed to "overlay archives".

docs/markdown/Wrap-dependency-system-manual.md
pwmarcz: >They are searched for at the top level?

Yes. The original PR used the top level, and I think that's a good choice because it's consistent with behavior of `source_filename`, `patch_filename`, and `patch_directory` (which also do not expect any extra subdirectory).

>If applying more than one, I guess paths with directory components would be used for grouping?

Yes, I think so. Actually, I'll edit the example to reflect that.

docs/markdown/Wrap-dependency-system-manual.md
pwmarcz: Done and done.

docs/markdown/Wrap-dependency-system-manual.md
pwmarcz: Done.

mesonbuild/wrap/wrap.py
xclaesse: ok

xclaesse:
Missing release notes (add a file in docs/markdown/snippets/) but other than that it looks good to me. Thanks for finishing it.

pwmarcz:
Release notes added.

pwmarcz:
Hi, is there a chance of getting this merged?

mesonbuild/wrap/wrap.py
```diff
@@ -159,6 +162,10 @@ def parse_wrap_section(self, config: configparser.ConfigParser) -> None:
             raise WrapException(f'{self.wrap_section!r} is not a valid first section in {self.basename}')
         self.type = self.wrap_section[5:]
         self.values = dict(config[self.wrap_section])
+        if 'diff_files' in self.values:
+            FeatureNew('Wrap files with diff_files', '0.61.0').use(self.subproject)
+            diff_files = [s.strip() for s in self.values['diff_files'].split(',')]
+            self.diff_files += diff_files
```
jpakkane, line 168: This should validate that the patch file does not a) have an absolute path and b) have a `..` segment in it to ensure that only files inside the proper directory are considered. (people have a tendency to do nasty workarounds like these if you let them)

pwmarcz: Done, I added a check for both cases.

tristan957:
Anything missing for this PR to go in 0.62 @eli-schwartz. Seems like you had marked "changes requested", but they seem to have resolved.

mesonbuild/wrap/wrap.py
```diff
@@ -610,6 +623,29 @@ def apply_patch(self) -> None:
...
+                cmd = [PATCH, '-f', '-p1', '-i', str(path)]
+            elif GIT:
+                # If the `patch` command is not available, fall back to `git
+                # apply`. The `--work-tree` is necessary in case we're inside a
+                # Git repository: by default, Git will try to apply the patch to
+                # the repository root.
+                cmd = [GIT, '--work-tree', '.', 'apply', '-p1', str(path)]
+            else:
+                raise WrapException(
+                    'Missing "patch" or "git" commands to apply diff files')
```
eli-schwartz, line 642: Nit: this doesn't need to be split across two lines. We have other lines of similar length in this file and exceeding the 80 character line limit IMO is better as long as it's not *too* long and completely avoids multi-line invocations.

test cases/common/153 wrap file should not failed/meson.build
```diff
@@ -14,3 +14,8 @@ executable('grabprog2', files('src/subprojects/foo/prog2.c'))
 subdir('src')
 
 subproject('patchdir')
+
+if find_program('patch', required : false).found() or find_program('git', required : false).found()
+  exe = subproject('patchfile').get_variable('foo_exe')
```
eli-schwartz, line 19: Feels like this should maybe be a MESON_SKIP_TEST?

docs/markdown/Wrap-dependency-system-manual.md
xclaesse: The path is relative to `subprojects/packagefiles` but user can decide to add subdirectories of course.

xclaesse:
I think you also need to re-apply patches when doing "meson subprojects update --reset".

pwmarcz:
>I think you also need to re-apply patches when doing "meson subprojects update --reset".

Done: I assume you mean the `git_reset()` method in `msubprojects.py`, since that's where we call `apply_patch()` already.

test cases/common/153 wrap file should not failed/meson.build
pwmarcz: You mean, we should skip this whole test case when there's no `patch` / `git`? OK, done.

mesonbuild/wrap/wrap.py
pwmarcz: OK, reformatted to a single line.

mkow:
What's the status of this PR? It seems that all discussions were resolved and this patch would really make our lives easier.

tristan957:
This should probably be marked for 0.63

mesonbuild/wrap/wrap.py
```diff
@@ -162,6 +165,15 @@ def parse_wrap_section(self, config: configparser.ConfigParser) -> None:
             raise WrapException(f'{self.wrap_section!r} is not a valid first section in {self.basename}')
         self.type = self.wrap_section[5:]
         self.values = dict(config[self.wrap_section])
+        if 'diff_files' in self.values:
+            FeatureNew('Wrap files with diff_files', '0.62.0').use(self.subproject)
```
xclaesse, line 169:
```suggestion
            FeatureNew('Wrap files with diff_files', '0.63.0').use(self.subproject)
```

xclaesse (commented):
I thought this was merged already, LGTM.

xclaesse:
@pwmarcz Could you do the quick fix above so this can be merged please?

pwmarcz:
Done.

xclaesse:
LGTM when CI is green.

mkow:
> LGTM when CI is green.

@xclaesse: CI looks green :)

PR #9719, opened 2021-12-11
rewriter: create {add,rm}_extra_files commands

ElementW:
Add ability to mutate a target's `extra_files` list through the rewriter.

The logic is copied from sources add/rm, but changes the `extra_files` kwarg instead of the sources positional argument.
Has additional logic to handle creating the `extra_files` list if it doesn't exist.

### DIFF
--- a/docs/markdown/Rewriter.md
+++ b/docs/markdown/Rewriter.md
@@ -77,6 +77,17 @@ and print an error message.
 For more information see the help output of the rewriter target
 command.
 
+### Adding and removing `extra_files`
+
+*Since 0.61.0*
+
+In the same way you can add and remove source files from a target, you can modify a target's
+`extra_files` list:
+
+```bash
+meson rewrite target <target name/id> {add_extra_files/rm_extra_files} [list of extra files]
+```
+
 ### Setting the project version
 
 It is also possible to set kwargs of specific functions with the
@@ -191,7 +202,7 @@ The format for the type `target` is defined as follows:
 {
   "type": "target",
   "target": "target ID/name/assignment variable",
-  "operation": "one of ['src_add', 'src_rm', 'target_rm', 'target_add', 'info']",
+  "operation": "one of ['src_add', 'src_rm', 'target_rm', 'target_add', 'extra_files_add', 'extra_files_rm', 'info']",
   "sources": ["list", "of", "source", "files", "to", "add, remove"],
   "subdir": "subdir where the new target should be added (only has an effect for operation 'tgt_add')",
   "target_type": "function name of the new target -- same as in the CLI (only has an effect for operation 'tgt_add')"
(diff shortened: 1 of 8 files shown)


mesonbuild/rewriter.py
```diff
@@ -709,6 +709,87 @@ def find_node(src):
                 if root not in self.modified_nodes:
                     self.modified_nodes += [root]
 
+        if cmd['operation'] == 'extra_files_add':
```
mensinda, line 712: This doesn't really matter, but for consistency, this should also be an `elif`.

mensinda (commented):
Thanks for adding this. The PR looks mostly good, but needs tests and a [release note snippet](https://github.com/mesonbuild/meson/tree/master/docs/markdown/snippets).

For the existing tests:

- https://github.com/mesonbuild/meson/blob/master/unittests/rewritetests.py
- https://github.com/mesonbuild/meson/tree/master/test%20cases/rewrite

mesonbuild/rewriter.py
ElementW: Great catch, didn't notice when I copy-pasted the sources add/rm logic. Will fix.

ElementW:
> The PR looks mostly good, but needs tests and a [release note snippet](https://github.com/mesonbuild/meson/tree/master/docs/markdown/snippets).

I did mention that:

> Unit test, doc and release note snippets will come, until then this is a draft PR.

mensinda:
> I did mention that:

I was working with the E-Mail I got from GitHub and didn't notice your edit :)

docs/markdown/Rewriter.md
```diff
@@ -77,6 +77,17 @@ and print an error message.
 For more information see the help output of the rewriter target
 command.
 
+### Adding and removing `extra_files`
+
+*Since 0.61.0*
```
ElementW, line 82: Not sure about the version I should specify here, if I should specify at all.

ElementW:
Added unit tests, documentation, as well as release snipped.
The contribution guide mentions

> Must be registered with the FeatureChecks framework

but I don't think this is applicable to the feature I'm adding as this is not a feature usable in build files.

docs/markdown/Rewriter.md
mensinda: Always use the next Meson release version, so this is correct. If a PR doesn't make it the version must be updated.

mensinda:
> but I don't think this is applicable to the feature I'm adding as this is not a feature usable in build files.

Yes, this is correct because it is a command-line only change.

ElementW:
Fixed bits of english and renamed `rewriter_extra_sources.md` to `rewriter_extra_files.md`

ElementW:
Fixed the `extra_files argument must be a list` error being triggered on targets with `extra_files` being a variable whose value flattens to an empty array. Added to the `test_target_add_extra_files` test case.

ElementW:
Fixed the project language in the test case Meson script, it should resolve the "*Could not get a dynamic linker for build target*" issue on msys2 targets

mensinda (approved):
LGTM

PR #9722, opened 2021-12-12
intl dep: forward static to iconv if needed

lazka:
In 1fb6c939473ca7cdab2 the intl dep gained support for static linking
which also forwarded this property to the iconv sub dependency.

The refactoring in 214d03568f75 lost this change, which results in
iconv getting linked dynamically again.

Forward static again to fix this.


lazka:
Thanks!

PR #9723, opened 2021-12-13
add ft32 cpu family

### LINKED ISSUE #9670: Unknown CPU family ft32
I'm currently compiling a project for an embedded device usint a bridgetek FT902 (FT32 architecture in GCC), as I expected this is not a supported target on meson, I'm not sure what this target variable implies, but compiling with the unsupported
```
cpu_family = 'ft32'
cpu = 'ft900'
```
doesnt seem to produce any issues for me.

If there is something I can do about this let me know I'm open to contribute.

perigoso:
Is there anything one needs to to 'support' a new family? this is only use for quirks correct?

closes #9670

### DIFF
--- a/docs/markdown/Reference-tables.md
+++ b/docs/markdown/Reference-tables.md
@@ -90,6 +90,7 @@ set in the cross file.
 | csky                | 32 bit CSky processor    |
 | dspic               | 16 bit Microchip dsPIC   |
 | e2k                 | MCST Elbrus processor    |
+| ft32                | 32 bit Bridgetek MCU     |
 | ia64                | Itanium processor        |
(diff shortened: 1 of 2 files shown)


perigoso:
closes #9670

jpakkane:
If this cpu is known by many names (like 32 bit arm etc) then the cpu family detection code would need to be updated. If these are cross compile only chips (as seems to be the case) then that's not needed as the code in question is only run when detecting the CPU of the current build machine.

perigoso:
It's just an MCU, and a niche one at that, so no

perigoso:
Anything I need to do to get this merged?

Issue #8490, opened 2021-03-08
how to get qt5 pkg-config variable?


goulf-3m:
In `/usr/lib64/pkgconfig/Qt5.pc`, there are variables like:

```
libexecdir=/usr/lib64/qt5/libexec
moc=/usr/lib64/qt5/bin/moc
plugindir=/usr/lib64/qt5/plugins
qmake=/usr/lib64/qt5/bin/qmake
```

I need to get the value of `plugindir` in the build system.
But `qt5_dep.get_variable(pkgconfig: 'plugindir')` does not work:

```
../source/gather/meson.build:35:0: ERROR: No default provided for dependency <Qt5Dependency qt5: True>, which is not pkg-config, cmake, or config-tool based.
```

dcbaker:
Unfortunately qt has not been updated to use the new factory mechanism for dependencies with multiple ways to be found, which means there's no way to get the pkg-config variables. I've looked at it and tried a few times to convert it, but it's a dense bit of code and I've never been able to get it successfully converted.

goulf-3m:
Very quick response! Thanks!

Here is some samples for this variable:

```
# Fedora
pkg-config --variable=plugindir Qt5
qmake-qt5 -query QT_INSTALL_PLUGINS
# Fedora/MinGW64
x86_64-w64-mingw32-qmake-qt5 -query QT_INSTALL_PLUGINS
# MSYS2/MinGW64
qmake.exe -query QT_INSTALL_PLUGINS
```

I've read some of meson's code.
Can overloading `get_variable` for `QtBaseDependency` solve this?
The interface will become `qt5_dep.get_variable(pkgconfig: 'plugindir', qmake: 'QT_INSTALL_PLUGINS')`.

PR #8568, opened 2021-03-23
QT: use a proper dependency factory

dcbaker:
Currently our Qt dependency is still using the old "combined" mechanism for dependencies with multiple ways to be found. This means that `get_variable` and `get_pkg_config_variable` don't work on `Qt`.

After this change they're either a `PkgConfigDependency`, or a `ConfigToolDependency`, and you can use the normal methods to interact with them.

Fixes #8490
Fixes #7864

PR #9728, opened 2021-12-14
`dep.get_variable()` for qmake

bb010g:
### factor out `ConfigToolDependency.get_configtool_args_for_variable()` from `ConfigToolDependency.get_configtool_variable()`

This facilitates getting variables from config tools with non-standard argument syntax, both in Python and in Meson's specification language (via `dep.get_variable()` or `dep.get_configtool_variable()`), avoiding duplication of error handling & further parsing of tool results that would come from overriding the whole `ConfigToolDependency.get_configtool_variable()` function.

## dependencies/qt: `dep.get_variable()` for qmake

To get config tool variables with qmake, `qmake --{variable_name}` cannot be used; instead, `qmake -query {variable_name}` *must* be used.

This customizes `ConfigToolDependency.get_configtool_args_for_variable()` via `QmakeQtDependency.get_configtool_args_for_variable()` to support `dep.get_variable()` and `dep.get_configtool_variable()` for qmake dependencies.

Addresses <https://github.com/mesonbuild/meson/issues/8490>. For example, `qt5_dep.get_variable(configtool : 'QT_INSTALL_PLUGINS)` as suggested by @goulf-3m actually works now.

Follow up to <https://github.com/mesonbuild/meson/pull/8568>.

Refactoring `QmakeQtDependency.__init__()` to use `self.get_configtool_variable()` or `self.get_configtool_args_for_variable()` was considered, but given that it's a special use of `self.get_config_value()` that eagerly queries all variables at once, which `self.get_configtool_variable()` doesn't currently support, the function was left alone.


mesonbuild/dependencies/qt.py
```diff
@@ -319,6 +319,9 @@ def __init__(self, name: str, env: 'Environment', kwargs: T.Dict[str, T.Any]):
             if not self._link_with_qtmain(is_debug, libdir):
                 self.is_found = False
 
+    def get_configtool_args_for_variable(self, variable_name: str) -> T.List[str]:
+        return self.config + ['-query', variable_name]
```
dcbaker, line 323: This either needs to be a protected method (ie, `_get_configtool_args_for_variable(...)`), or it needs to have a stub version that raises `NotImplementedError` in the `Dependency` base class. The only other thing this probably deserves is a release snippet,. Thanks for working on this!

dcbaker (changes requested):

eli-schwartz:
The changes themselves look reasonable enough, but the commit *messages* are... very information-dense.

Describing the exact implementation classes being modified is generally not necessary, and may detract from readability in `git log`. This information is already available in the change diff.

`Addresses <https://github.com/mesonbuild/meson/issues/8490>` in the middle of a paragraph should be removed, and replaced by `Fixes #8490` in its *own* paragraph at the end of the commit message. This avoids stray `<>` in the github UI, because raw links are expected by the renderer, and matches the expected pattern for github's auto-linkification keywords (this ensures that merging the PR marks the issue as closed).

Neumann-A:
does this correctly support `qmake --qtconf '<somefullpath' --query <variablename>` ?

bb010g:
> The changes themselves look reasonable enough, but the commit _messages_ are... very information-dense.
> 
> Describing the exact implementation classes being modified is generally not necessary, and may detract from readability in `git log`. This information is already available in the change diff, and doesn't aid the reader in the meta-knowledge of what effective difference the change makes, and why, nor does it add contextual information lacking in the diff itself.
> 
> `Addresses <https://github.com/mesonbuild/meson/issues/8490>` in the middle of a paragraph should be removed, and replaced by `Fixes #8490` in its _own_ paragraph at the end of the commit message. This avoids stray `<>` in the github UI, because raw links are expected by the renderer, and matches the expected pattern for github's auto-linkification keywords (this ensures that merging the PR marks the issue as closed).

@eli-schwartz I'll shorten those up; figured I'd lean generous on commit messages rather than possibly have to write up comprehensive ones later. :)

> does this correctly support `qmake --qtconf '<somefullpath' --query <variablename>` ?

@Neumann-A I've tested only on Qt 5.15.7; `--query` is an invalid option for `qmake`. `~/Qt/5.15.7/clang_64/bin/qmake -query QT_INSTALL_LIBS -qtconf ~/Qt/5.15.7/clang_64/bin/qt.conf` works as expected; `~/Qt/5.15.7/clang_64/bin/qmake -qtconf ~/Qt/5.15.7/clang_64/bin/qt.conf -query QT_INSTALL_LIBS` exits with code 1, prints `***Unknown option -query` to stdout, and prints help to stderr; `~/Qt/5.15.7/clang_64/bin/qmake --query QT_INSTALL_LIBS --qtconf ~/Qt/5.15.7/clang_64/bin/qt.conf` exits with code 1, prints `***Unknown option --query` to stdout, and prints help to stderr.

`-qtconf <file>` is currently not used by `QmakeQtDependency`; support for this option can be implemented in a separate patch set.

If any of this is preferable in Qt 6 or required in Qt 6, patches are welcome here.

eli-schwartz:
I think we're still just waiting on @dcbaker's review here (and ideally a tweak to the commit message wording). Any news on an update?

tristan957:
@bb010g still interested in finishing this PR? Seems rather close to being done.

PR #9729, opened 2021-12-15
Full typing and typed_* for the interperterobjects module

dcbaker:
This series adds full type annotations and use of the typed_* decorators for interpreter objects. as part of that work I've changed the `ConfigurationDataObject` into a proper `ConfigurationDataHolder`, that is a holder and acts and works like a holder, with all of the core logic pushed into `build.ConfigurationData`, and the interpreter logic in the interpreter. This has a nice cleaning effect throughout the code base.

There is one contentious change here, and that's dropping support for setting and getting values that are not `str | int | bool`. We never meant to accept other kinds of values, but we did, and we only deprecated that support at the start of the year, so it may well be too soon to drop that. Nevertheless, I've included that in the series as it allows more cleanups, and allows for full type checking, which, I think is pretty valuable.

### DIFF
--- a/docs/yaml/objects/cfg_data.yaml
+++ b/docs/yaml/objects/cfg_data.yaml
@@ -41,8 +41,17 @@ methods:
       type: str
       description: The name of the variable to set
     value:
-      type: bool
-      description: The value to set as either `1` or `0`
+      type: bool | int
+      description: |
+        The value to set as either `1` or `0`
+
+        Passing numbers was never intended to work, and since 0.62 it has been
+        deprecated. It will be removed in a future version of Meson. If you
+        need to pass numbers use the `.set` method.
+  warnings:
+  - numeric values < 0 have the surprising behavior of being converted to
+    [[true]], values > 1 have the more expected but unintentional behavior of
+    being interpretered as [[true]].
 
   kwargs_inherit: cfg_data.set
 
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -2767,17 +2767,16 @@ class CustomTargetIndex(HoldableObject):
         return self.target.get_custom_install_dir()
 
 class ConfigurationData(HoldableObject):
-    def __init__(self) -> None:
+    def __init__(self, initial_values: T.Optional[T.Union[
+                T.Dict[str, T.Tuple[T.Union[str, int, bool], T.Optional[str]]],
+                T.Dict[str, T.Union[str, int, bool]]]
+            ] = None):
         super().__init__()
-        self.values: T.Dict[
-            str,
-            T.Tuple[
-                T.Union[str, int, bool],
-                T.Optional[str]
-            ]
-        ] = {}
+        self.values: T.Dict[str, T.Tuple[T.Union[str, int, bool], T.Optional[str]]] = \
+            {k: v if isinstance(v, tuple) else (v, None) for k, v in initial_values.items()} if initial_values else {}
+        self.used: bool = False
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         return repr(self.values)
 
     def __contains__(self, value: str) -> bool:
--- a/mesonbuild/dependencies/pkgconfig.py
+++ b/mesonbuild/dependencies/pkgconfig.py
@@ -24,6 +24,7 @@ import typing as T
 
 if T.TYPE_CHECKING:
     from ..environment import Environment
+    from .._typing import ImmutableListProtocol
 
 class PkgConfigDependency(ExternalDependency):
     # The class's copy of the pkg-config path. Avoids having to search for it
@@ -380,18 +381,13 @@ class PkgConfigDependency(ExternalDependency):
             raise DependencyException(f'Could not generate libs for {self.name}:\n\n{out_raw}')
         self.link_args, self.raw_link_args = self._search_libs(out, out_raw)
 
-    def get_pkgconfig_variable(self, variable_name: str, kwargs: T.Dict[str, T.Union[str, T.List[str]]]) -> str:
+    def get_pkgconfig_variable(self, variable_name: str,
+                               define_variable: 'ImmutableListProtocol[str]',
+                               default: T.Optional[str]) -> str:
         options = ['--variable=' + variable_name, self.name]
 
-        if 'define_variable' in kwargs:
-            definition = kwargs.get('define_variable', [])
-            if not isinstance(definition, list):
-                raise DependencyException('define_variable takes a list')
-
-            if len(definition) != 2 or not all(isinstance(i, str) for i in definition):
-                raise DependencyException('define_variable must be made up of 2 strings for VARIABLENAME and VARIABLEVALUE')
-
-            options = ['--define-variable=' + '='.join(definition)] + options
+        if define_variable:
+            options = ['--define-variable=' + '='.join(define_variable)] + options
 
         ret, out, err = self._call_pkgbin(options)
         variable = ''
@@ -406,9 +402,8 @@ class PkgConfigDependency(ExternalDependency):
             if not variable:
                 ret, out, _ = self._call_pkgbin(['--print-variables', self.name])
                 if not re.search(r'^' + variable_name + r'$', out, re.MULTILINE):
-                    if 'default' in kwargs:
-                        assert isinstance(kwargs['default'], str)
-                        variable = kwargs['default']
+                    if default:
+                        variable = default
                     else:
                         mlog.warning(f"pkgconfig variable '{variable_name}' not defined for dependency {self.name}.")
 
@@ -483,13 +478,8 @@ class PkgConfigDependency(ExternalDependency):
                      default_value: T.Optional[str] = None,
                      pkgconfig_define: T.Optional[T.List[str]] = None) -> T.Union[str, T.List[str]]:
         if pkgconfig:
-            kwargs: T.Dict[str, T.Union[str, T.List[str]]] = {}
-            if default_value is not None:
-                kwargs['default'] = default_value
-            if pkgconfig_define is not None:
-                kwargs['define_variable'] = pkgconfig_define
             try:
-                return self.get_pkgconfig_variable(pkgconfig, kwargs)
+                return self.get_pkgconfig_variable(pkgconfig, pkgconfig_define or [], default_value)
             except DependencyException:
                 pass
         if default_value is not None:
--- a/mesonbuild/interpreter/interpreterobjects.py
+++ b/mesonbuild/interpreter/interpreterobjects.py
@@ -477,14 +477,9 @@ class DependencyHolder(ObjectHolder[Dependency]):
     @FeatureDeprecated('dependency.get_configtool_variable', '0.56.0',
                        'use dependency.get_variable(configtool : ...) instead')
     @noKwargs
-    def configtool_method(self, args: T.List[TYPE_var], kwargs: TYPE_kwargs) -> str:
-        args = listify(args)
-        if len(args) != 1:
-            raise InterpreterException('get_configtool_variable takes exactly one argument.')
-        varname = args[0]
-        if not isinstance(varname, str):
-            raise InterpreterException('Variable name must be a string.')
-        return self.held_object.get_configtool_variable(varname)
+    @typed_pos_args('dependency.get_config_tool_variable', str)
+    def configtool_method(self, args: T.Tuple[str], kwargs: TYPE_kwargs) -> str:
+        return self.held_object.get_configtool_variable(args[0])
 
     @FeatureNew('dependency.partial_dependency', '0.46.0')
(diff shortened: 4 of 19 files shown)


tristan957:
Quick all the checks are green. Merge fast :)

mesonbuild/build.py
```diff
@@ -2773,17 +2773,13 @@ def get_custom_install_dir(self) -> T.List[T.Union[str, bool]]:
...
         super().__init__()
-        self.values: T.Dict[
-            str,
-            T.Tuple[
-                T.Union[str, int, bool],
-                T.Optional[str]
-            ]
-        ] = {}
+        self.values: T.Dict[str, T.Tuple[T.Union[str, int, bool], T.Optional[str]]] = \
+            {k: (v, None) for k, v in initial_values.items()} if initial_values else {}
```
dcbaker, line 2779: gah, this is wrong. and mypy should be complaining about this

mesonbuild/interpreter/interpreterobjects.py
```diff
@@ -393,15 +393,13 @@ def keys_method(self, args: T.List[TYPE_var], kwargs: TYPE_kwargs) -> T.List[str
...
-            raise InterpreterException('Merge_from takes one positional argument.')
-        from_object_holder = args[0]
-        if not isinstance(from_object_holder, ConfigurationDataObject):
+        from_object = args[0]
+        if not isinstance(from_object, ConfigurationDataObject):
             raise InterpreterException('Merge_from argument must be a configuration data object.')
-        from_object = from_object_holder.conf_data
-        for k, v in from_object.values.items():
-            self.conf_data.values[k] = v
+        self.conf_data.values.update(from_object.conf_data.values)
```
eli-schwartz: This is not a typed_* update, this is just a general dict cleanup. :p

mesonbuild/interpreter/interpreterobjects.py
```diff
@@ -276,8 +276,13 @@ def prepend_method(self, args: T.Tuple[str, T.List[str]], kwargs: 'EnvironmentSe
         self.held_object.prepend(name, values, kwargs['separator'])
 
 
+_CONF_DATA_SET_KWS: T.List[KwargInfo] = [
+     KwargInfo('description', (str, NoneType)),
+]
```
eli-schwartz, line 281: This doesn't need to allocate a list.

mesonbuild/interpreter/interpreterobjects.py
```diff
@@ -302,51 +307,28 @@ def is_used(self) -> bool:
...
+        self.conf_data.values[args[0]] = (args[1], kwargs['description'])
+
+    @typed_pos_args('configuration_data.set_quoted', str, str)
+    @typed_kwargs('configuration_data.set_quoted', *_CONF_DATA_SET_KWS)
+    def set_quoted_method(self, args: T.Tuple[str, str], kwargs: 'kwargs.ConfigurationDataSet') -> None:
+        self.__check_used()
+        escaped_val = '\\"'.join(args[1].split('"'))
+        self.conf_data.values[args[0]] = (f'"{escaped_val}"', kwargs['description'])
+
+    @typed_pos_args('configuration_data.set10', str, (int, bool))
```
eli-schwartz, line 321: This only takes bool, not int.

eli-schwartz (commented):
Overall looks like a really nice change. Just a couple minor nits.

mesonbuild/interpreter/interpreterobjects.py
```diff
@@ -297,97 +294,63 @@ def __init__(self, subproject: str, initial_values: T.Optional[T.Dict[str, T.Any
...
+        self.__check_used()
+        self.held_object.values[args[0]] = (int(args[1]), kwargs['description'])
+
+    @typed_pos_args('configuration_data.has', (str, int, bool))
+    @noKwargs
+    def has_method(self, args: T.Tuple[T.Union[str, int, bool]], kwargs: TYPE_kwargs) -> bool:
+        return args[0] in self.held_object.values
 
     @FeatureNew('configuration_data.get()', '0.38.0')
     @noArgsFlattening
```
eli-schwartz, line 333: I think there is nothing to flatten-or-dont-flatten now?

mesonbuild/interpreter/interpreterobjects.py
```diff
@@ -396,25 +359,23 @@ def get_unquoted_method(self, args: T.List[TYPE_var], kwargs: TYPE_kwargs) -> T.
         return val
 
     def get(self, name: str) -> T.Tuple[T.Union[str, int, bool], T.Optional[str]]:
-        return self.conf_data.values[name]
+        return self.held_object.values[name]
 
     @FeatureNew('configuration_data.keys()', '0.57.0')
     @noPosargs
```
eli-schwartz, line 365: We can ban kwargs here too.

mesonbuild/interpreter/interpreterobjects.py
dcbaker: we document that we do: docs/yaml/objects/feature.yaml

mesonbuild/interpreter/interpreterobjects.py
dcbaker: I think that has to do with the fallback argument, but I guess we can get rid of that now?

mesonbuild/dependencies/pkgconfig.py
```diff
@@ -380,18 +381,13 @@ def _set_libs(self) -> None:
...
+                               default: T.Optional[str]) -> str:
         options = ['--variable=' + variable_name, self.name]
 
-        if 'define_variable' in kwargs:
-            definition = kwargs.get('define_variable', [])
-            if not isinstance(definition, list):
-                raise DependencyException('define_variable takes a list')
-
-            if len(definition) != 2 or not all(isinstance(i, str) for i in definition):
-                raise DependencyException('define_variable must be made up of 2 strings for VARIABLENAME and VARIABLEVALUE')
```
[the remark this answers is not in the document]
dcbaker: We have a pairs check, but I've added a check to ensure that it's length 0 or 2 in the interpreter.

mesonbuild/interpreter/interpreterobjects.py
eli-schwartz: What do feature objects have to do with cfg_data? :p And nowhere in that file is `int` mentioned.

cfg_data does mention int, but only in the context of set / set_quoted, not set10. That is because those actually support defining e.g. `3`.

mesonbuild/interpreter/interpreterobjects.py
eli-schwartz: Yeah, that's what I meant. We no longer accept fallbacks which can be of type list, which "need" to be preserved as is.

mesonbuild/interpreter/interpreterobjects.py
dcbaker: from said place:
```meson
     if get_option('directx').require(host_machine.system() == 'windows',
           error_message: 'DirectX only available on Windows').allowed() then
       src += ['directx.c']
       config.set10('HAVE_DIRECTX', 1)
     endif
```

I'm not opossed to remove bool support, but I suspect since we've documented it, and we've allowed it, someone is using it; and unlike passing say `'1'` there isn't any reason that passing a literal number isn't valid, since meson does inherit a bit of the bool is int madness from C.

dcbaker:
Oh, I noticed that there is code that could be made better here since we're using a holder!

mesonbuild/interpreter/interpreterobjects.py
eli-schwartz: There is absolutely a reason to not allow it. If we allow it, people can pass `set10('VAR', 5)` and that is interpreted as "true" same as "1". In fact, the reason it is interpreted as true is because we currently do no validation at all, you can do `config.set10('HAVE_DIRECTX', "the little brown fox jumped over the lazy dog")` and meson does a python

```python
        if val:
            self.conf_data.values[name] = (1, desc)
        else:
            self.conf_data.values[name] = (0, desc)
```

Since *any* non-empty string, and any non-zero integer, and anything that generally has contents, is true in python... this sets to 1.

In fact, this works too:

```meson
config.set10('HAVE_DIRECTX', ' ')
```
A string of length one, filled with space characters. Equates to true.

This also evaluates to true:

```meson
config.set10('HAVE_DIRECTX', -1)
```
Meson has negative integers. Python considers negative integers to be truthy, not falsy.


This was added quite recently in 2f2d99e1d88c9f14fa139d450cb6f55173d73c9a and it isn't even the set10 documentation...

I'm *strongly* opposed to documenting it as allowed, but would begrudgingly compromise on raising a deprecation warning declaring that it is an invalid construct and scheduling it for removal in 0.62.0

But as far as I'm concerned, the documentation is buggy and makes no sense. :/

If we are going to keep it and mark it as deprecated, I desperately hope no one is accidentally passing negative numbers and getting truthy results!!!

eli-schwartz: I approve this PR once a deprecation is added for passing integers here.

mesonbuild/interpreter/interpreterobjects.py
```diff
@@ -320,6 +320,14 @@ def set_quoted_method(self, args: T.Tuple[str, str], kwargs: 'kwargs.Configurati
     @typed_kwargs('configuration_data.set10', _CONF_DATA_SET_KWS)
     def set10_method(self, args: T.Tuple[str, T.Union[int, bool]], kwargs: 'kwargs.ConfigurationDataSet') -> None:
         self.__check_used()
+        if isinstance(args[1], int):
+            FeatureDeprecated.single_use('configuration_data.set10 with number',
```
eli-schwartz, line 324: Don't use FeatureDeprecated for this. It should be unconditionally deprecated, not only when users set their `meson_version: '>=0.61.0'`

Just use 
```python
if isinstance(args[1], int):
    mlog.deprecation('configuration_data.set10 received an invalid numeric argument, when only bools '
                     'are permitted. This has unexpected results for numbers other than 0 or 1. This '
                     'will become a hard error in a future version of meson.',
                     location=self.interpreter.current_node)
```

eli-schwartz (changes requested):

docs/yaml/objects/cfg_data.yaml
```diff
@@ -41,8 +41,15 @@ methods:
...
-      description: The value to set as either `1` or `0`
+      type: bool | int
+      description: |
+        The value to set as either `1` or `0`
+
+        Since 0.61.0 passing a number is deprecated, use only booleans or the
+        `.set` method instead.
+
+        *NOTE:* numeric values < 0 have the surprising behavior of being
+        converted to `true`.
```
eli-schwartz, line 52: It is not deprecated "since 0.61". And, the `NOTE:` can be described in yaml as

```yaml
notes:
  - numeric values < 0 have [...]
```

But better yet we can make it a `warnings:`

Actually I'd like the entire thing to be a warnings:

```
warnings:
  - Historically, meson would accidentally allow passing a number instead of a boolean. This has the
    surprising behavior of converting to `true` for both numbers < 0 and numbers > 1. It was never
    supposed to be accepted and should never be used. Instead, use only booleans or the
    [[cfg_data.set]] method.
```

mesonbuild/interpreter/interpreterobjects.py
dcbaker: I thought we were trying to stop using `mlog.deprecation` in favor of `FeatureDeprecated`?

eli-schwartz: Are we?

We use mlog.warning all the time for things that are always warnings, maybe that would be preferable?

In my mind, the purpose of mlog is for things that always need to be logged, whatever the reason, and Features are for things that depend on what version of meson you're trying to support.

I guess an argument could be made that "deprecation" implies versions where it is okay to use, and that therefore this *should* be an mlog.warning either way.

docs/yaml/objects/cfg_data.yaml
```diff
@@ -41,8 +41,16 @@ methods:
...
       description: The name of the variable to set
     value:
-      type: bool
-      description: The value to set as either `1` or `0`
+      type: bool | int
+      description: |
+        The value to set as either `1` or `0`
+
+        Since 0.62.0 passing a number is deprecated, use only booleans or the
+        `.set` method instead.
```
eli-schwartz, line 49: I would still like this to not claim that it is only deprecated since 0.62

```suggestion
        Passing a number was never intended to work, and is deprecated and will be
        removed in a future version of meson. Use only booleans or the
        `.set` method instead.
```

dcbaker: reworked again. I think having 0.62 in there is important so that users know why they are suddenly getting this warning

eli-schwartz (approved):

Issue #3287, opened 2018-03-21
Allow excluding directories from coverage report


ptomato:
`gcovr` has an `--exclude-directories` option. It would be nice to either expose this option in Meson, or make Meson automatically exclude paths outside the source and build roots. Otherwise, you get lots of entries inside `/usr/include` in your coverage report (to be precise, any `.h` file which was included in your code and defines an `inline` function.) Generally one wouldn't want to measure these external libraries' code as part of a coverage report.

NickeZ:
This is already implemented for lcov. So I guess it is simply a matter of doing a PR. I would however wait until https://github.com/mesonbuild/meson/pull/3242 is merged.

thejk:
It should already be included in #3242, I added --exclude for subprojects/ and --rootdir option should already remove references outside project root in gcovr 3.1+. Might need something else for older versions of gcovr tho.

kjerstadius:
Is anyone working on this feature? I'm willing to try to implement it if not. Before diving into the specifics of the implementation, what would be a good way to control the exclude option? I'm thinking a function like "coverage_exclude_directory('test-dir')" would perhaps be in line with how Meson works in general? I'm just mindful of the number of different functions exposed as one of the things I really like about Meson is the relatively small amount of supported functions.

kjerstadius:
So to be clear, what I would be after is to exclude parts of my own project. For the project I'm working on we publish (and potentially gate on) coverage numbers on Jenkins. However, while we're interested in coverage for the code that we ship, we're not really interested in coverage of the test code itself. So I'd like to exclude the test code from the coverage report so we can publish the XML report directly, rather than having to go through some xml filtering step. Or am I missing something here?

kjerstadius:
While I believe you're probably right, I managed to put a working solution for ninja + gcovr together. My implementation adds a new project option that accepts a string of colon separated exclude patterns that are each passed as -e arguments to gcovr, essentially identical to how the subproject root is being excluded already. There's no lcov support yet, but it should be possible to do the same thing using --remove as a post process step, again similar to how subproject root is excluded.

Though the solution works for my case, there's definitely more work needed to make a complete change (like making the list of patterns work regardless file and path separators, lcov support). Before I proceed with that though, I'd like to ask whether a pull request with a solution as described would stand any chance of getting accepted. If it goes against any fundamental design principles I'd rather know upfront :)

perror:
[...]

PR #9732, opened 2021-12-15
Add headless support to exclude paths from coverage reports

perror:
This code simply add the feature to the `mesonbuild/scripts/coverage.py` and has no way to trigger it yet.
Yet, if everything is okay, we should try to get a way to trigger this from a `meson.build` file. For example:
```
coverage_set_exclude_pattern('*tests/*')
```

Related to issue #3287


perror:
Okay, I need to make it support lists or regex and write some test for code coverage... I will do it asap.

PR #9737, opened 2021-12-16
cmake: Use new CMakeExecutor instead of old logic (fixes #9733)

### LINKED ISSUE #9733: [0.60.2] import('cmake') cannot handle cmake not being on path.
Meson just says it cannot find cmake although it has been passed via a native/cross file as: 
```
[binaries]
...
cmake = 'E:/vcpkg_cache/downloads/tools/cmake-3.21.1-windows/cmake-3.21.1-windows-i386/bin/cmake.exe'
....
```

CI run (https://dev.azure.com/vcpkg/public/_build/results?buildId=64620&view=logs&jobId=7922e5c4-0103-5f8f-ad17-45ce9bb98e80&j=878666d5-db33-5b27-9e7d-b0c7ee352005)

The fact that meson assumes stuff to be on path although full paths are passed seems to be a consistent error...... (same goes for ninja)

mensinda:
No test case, since this would require removing CMake from path

### DIFF
--- a/mesonbuild/modules/cmake.py
+++ b/mesonbuild/modules/cmake.py
@@ -19,7 +19,7 @@ import typing as T
 from . import ExtensionModule, ModuleReturnValue, ModuleObject
 
 from .. import build, mesonlib, mlog, dependencies
-from ..cmake import SingleTargetOptions, TargetOptions, cmake_defines_to_args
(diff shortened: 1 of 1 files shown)


mensinda:
Not sure why mypy is suddenly failing, I did not change any of these files.

eli-schwartz:
It fails literally everywhere. mypy released an updated version which caused the changes.

mensinda:
Closing, since #9751 got in.

PR #9738, opened 2021-12-16
Fix mypy errors

mensinda:


mesonbuild/interpreter/interpreterobjects.py
```diff
@@ -786,7 +786,13 @@ def __init__(self, target: _BuildTarget, interp: 'Interpreter'):
...
     def __repr__(self) -> str:
         r = '<{} {}: {}>'
-        h = self.held_object
+        h: build.BuildTarget
+        if isinstance(self, BothLibrariesHolder) and self is not None:
+            h = self.held_object.get_default_object()
+        elif self is not None and isinstance(self.held_object, build.BuildTarget):
+            h = self.held_object
+        else:
+            raise mesonlib.MesonBugException(f'It should be impossible to reach this point: {self}')
```
dcbaker, line 795: Ahh, if you pull BothLibraries out of the type union this goes away. We can't even actually hit this, because BothLibrariesHolder has its own `__repr__` method. I think the right solution is to split the `BuildTargetHolder` class so that it doesn't hold both build.BuildTarget and build.BothLibraries

dcbaker: Alternatively... since BothLibrariesHolder has it's own repr and doesn't call super(), you could probably just `assert not isinstance(self, BothLibrariesHolder)` at the top and be done with it

PR #9742, opened 2021-12-18
i18n: Add support for joining XML localization via itstool

ximion:
Hi!

This PR adds support for the "join" feature of `itstool` to Meson. While (x)gettext can merge XML, when working with AppStream metadata I found out that it is pretty unpredictable in how data is merged, especially when inline tags are involved. At the very least, inline markup gets XML-escaped, and I found no way at all to work around this particular issue.
[Itstool](http://itstool.org/) however always generates correct results and obeys the rules defined in ITS files properly. This tool is also very widespread and available in every Linux distribution I checked. It also is used by GNOME's help.

Using `itstool` from Meson is pretty cumbersome at the moment, as you need to scrape through the output generated from `i18n.gettext()` and rename mo files so itstool can infer locale names.
This is admittedly a bit dumb, but there's also no real way around this (even if itstool gained support to specify per-file locale on the command line, it would take ages before the changes are widespread enough to be used unconditionally).
The new code comes with a pretty realistic example of how this feature will likely be used in Meson as unit test.

If `itstool_join` is called we fail if the tool could not be found (as opposed to `merge_file`). I chose this behavior because a silent failure would mean crucial data may not get installed, and even if it was installed would be untranslated which explicitly was not the intent of the developer when the join command is used.
Besides that, I don't think there's anything else particularly noteworthy about this change.

Thanks for considering!

### DIFF
--- a/docs/markdown/i18n-module.md
+++ b/docs/markdown/i18n-module.md
@@ -56,3 +56,15 @@ for normal keywords. In addition it accepts these keywords:
 * `args`: (*Added 0.51.0*) list of extra arguments to pass to `msgfmt`
 
 *Added 0.37.0*
+
+### i18n.itstool_join()
+
+This joins translations into a XML file using `itstool`. See
+[[@custom_tgt]]
+for normal keywords. In addition it accepts these keywords:
+
+* `its_files`: filenames of ITS files that should be used explicitly
+  (XML translation rules are autodetected otherwise).
+* `mo_targets` *required*: mo file generation targets as returned by `i18n.gettext()`.
+
+*Added 0.61.0*
--- a/mesonbuild/modules/i18n.py
+++ b/mesonbuild/modules/i18n.py
@@ -59,6 +59,20 @@ if T.TYPE_CHECKING:
         languages: T.List[str]
         preset: T.Optional[str]
 
+    class ItsJoinFile(TypedDict):
+
+        input: T.List[T.Union[
+            str, build.BuildTarget, build.CustomTarget, build.CustomTargetIndex,
+            build.ExtractedObjects, build.GeneratedList, ExternalProgram,
+            mesonlib.File]]
+        output: T.List[str]
+        build_by_default: bool
+        install: bool
+        install_dir: T.List[T.Union[str, bool]]
+        install_tag: T.List[str]
+        its_files: T.List[str]
+        mo_targets: T.List[T.Union[build.BuildTarget, build.CustomTarget, build.CustomTargetIndex]]
+
 
 _ARGS: KwargInfo[T.List[str]] = KwargInfo(
     'args',
@@ -115,11 +129,16 @@ class I18nModule(ExtensionModule):
         self.methods.update({
             'merge_file': self.merge_file,
             'gettext': self.gettext,
+            'itstool_join': self.itstool_join,
         })
 
     @staticmethod
     def nogettext_warning() -> None:
         mlog.warning('Gettext not found, all translation targets will be ignored.', once=True)
+
+    @staticmethod
+    def noitstool_error() -> T.NoReturn:
+        raise mesonlib.MesonException('Did not find itstool. Please install it to continue.')
 
     @staticmethod
     def _get_data_dirs(state: 'ModuleState', dirs: T.Iterable[str]) -> T.List[str]:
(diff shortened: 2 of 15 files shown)


ximion:
Looks like on bionic, itstool does not generate the expected output - I'll try to reproduce that locally to see what's going on there (and either fix the issue or add a smarter test).

ximion:
That bionic build failure was caused by an older itstool version which did not read custom ITS files correctly while joining. I adjusted the test to pass even if we have an older itstool which has that bug.

test cases/frameworks/6 gettext/data3/meson.build
```diff
@@ -0,0 +1,36 @@
...
+if itstool.found()
+
+
+mi_translated = i18n.itstool_join(
+  input:  'com.mesonbuild.test.intlprog.metainfo.xml',
+  output: 'com.mesonbuild.test.intlprog.metainfo.xml',
+  mo_targets: mo_targets,
+  its_files: ['metainfo.its'],
+  install: true,
+  install_dir: join_paths(get_option('datadir'), 'metainfo'),
```
jpakkane, line 11: This is old-style code, the modern equivalent should be `get_option('datadir') / 'metainfo'`.

test cases/frameworks/6 gettext/data3/meson.build
```diff
@@ -0,0 +1,36 @@
...
+)
+
+# older versions of itstool have a bug where ITS rules specified on the command-line
+# are not read when joining files. Since we don't install appstream in the Meson CI
+# environment, the to-be-tested entry will be untranslated and the test would fail, so
+# we just skip verification if the installed itstool is too old.
+r = run_command(itstool, '-v', check: true)
+itstool_v = r.stdout().strip().split()
+if itstool_v[1].version_compare('>=2.0.6')
+  grep_exe = find_program('grep')
```
jpakkane, line 21: This is _not_ acceptable. Any use of shell tools is absolutely prohibited as all tests must run on a bog standard Windows `cmd.exe` without any unixy stuff. This should be a standalone Python script that reads the result file and validates it.

test cases/frameworks/6 gettext/data3/meson.build
```diff
@@ -0,0 +1,36 @@
...
+            mi_translated]
+  )
+else
+  message('Skipping translation verification: Itstool too old.')
+endif
+
+
+else
+install_data('com.mesonbuild.test.intlprog.metainfo.xml',
+    install_dir: join_paths(get_option('datadir'), 'metainfo'))
```
jpakkane, line 34: This should use `/` as well.

jpakkane (changes requested):

eli-schwartz:
> Using `itstool` from Meson is pretty cumbersome at the moment, as you need to scrape through the output generated from `i18n.gettext()` and rename mo files so itstool can infer locale names.
This is admittedly a bit dumb, but there's also no real way around this (even if itstool gained support to specify per-file locale on the command line, it would take ages before the changes are widespread enough to be used unconditionally).

It will take even longer if no one ever pushes to have this changed, so it never gets implemented. If we are going to handle this I'd vastly, vastly prefer to have two codepaths, and choose which one to use depending on the detected version of itstool:
- one where this is done properly
- one where strange file copying happens invisibly via private scripts that hide itstool from ninja and circumvent find_program_impl and repeatedly create tempdirs to repeatedly do file IO

eli-schwartz:
Hmmm... When I was fixing shared-mime-info use of itstool that ran a custom shellscriot which invoked ninja to build mo files, my first implementation actually ran cp to copy those files as ninja build rules, and then just ran itstool directly from ninja. This is more portably done as a meson internal copy.py script... Honestly I like that thought a lot better than layering it in a script with a tempdir.

However I never committed that code because I realized it could just use the better project via i18.merge_file; maybe gettext itself could be fixed to not have the bugs you say are there?

It is a bit inconvenient that itstool requires running msgfmt to generate .mo files which are never installed, when gettext can directly merge .po files into your xml.

ximion:
@eli-schwartz 
> It will take even longer if no one ever pushes to have this changed, so it never gets implemented. If we are going to handle this I'd vastly, vastly prefer to have two codepaths, and choose which one to use depending on the detected version of itstool

Sure, if/when `itstool` has gained support for this, we can do it! Annoyingly, a mo-file does not contain the locale string that it belongs to, so there is nothing clever we can do here, so I am not even sure how a proper solution that does not depend on file names would look like (it would always be some kind of heuristic for semi-clever locale name autodetection).

> It is a bit inconvenient that itstool requires running msgfmt to generate .mo files which are never installed, when gettext can directly merge .po files into your xml.

True, both tools are surprisingly inconvenient in respective corner cases... But at least `itstool` with its more limited scope does the job properly...

@jpakkane I addressed your style comments, and the resulting data is also now verified by a small Python script instead of using `grep` & Co.

test cases/frameworks/6 gettext/data3/meson.build
```diff
@@ -0,0 +1,36 @@
+
+if itstool.found()
+
+
+mi_translated = i18n.itstool_join(
```
jpakkane, line 5: These should be indented.

ximion: Done :-)

eli-schwartz:
> Sure, if/when `itstool` has gained support for this, we can do it!

Are you saying you've submitted a bug report to itstool for this? Great. Please link it here though.

ximion:
> > Sure, if/when `itstool` has gained support for this, we can do it!
> 
> Are you saying you've submitted a bug report to itstool for this? Great. Please link it here though.

There is https://github.com/itstool/itstool/issues/32 and https://github.com/itstool/itstool/issues/44 on this issue (with the latter being the specific issue that we work around here).

docs/markdown/i18n-module.md
```diff
@@ -56,3 +56,15 @@ for normal keywords. In addition it accepts these keywords:
...
+
+This joins translations into a XML file using `itstool`. See
+[[@custom_tgt]]
+for normal keywords. In addition it accepts these keywords:
+
+* `its_files`: filenames of ITS files that should be used explicitly
+  (XML translation rules are autodetected otherwise).
+* `mo_targets` *required*: mo file generation targets as returned by `i18n.gettext()`.
+
+*Added 0.61.0*
```
eli-schwartz, line 70: This is out of date.

mesonbuild/modules/i18n.py
```diff
@@ -269,5 +288,65 @@ def gettext(self, state: 'ModuleState', args: T.Tuple[str], kwargs: 'Gettext') -
 
         return ModuleReturnValue([gmotargets, pottarget, updatepotarget], targets)
 
+    @FeatureNew('i18n.itstool_join', '0.61.0')
```
eli-schwartz, line 291: This is out of date.

mesonbuild/modules/i18n.py
```diff
@@ -269,5 +288,65 @@ def gettext(self, state: 'ModuleState', args: T.Tuple[str], kwargs: 'Gettext') -
...
+        CT_INSTALL_TAG_KW,
+        CT_OUTPUT_KW,
+        INSTALL_KW,
+        _ARGS.evolve(),
+        KwargInfo('its_files', ContainerTypeInfo(list, str)),
+        KwargInfo('mo_targets', ContainerTypeInfo(list, build.CustomTarget), required=True),
+    )
+    def itstool_join(self, state: 'ModuleState', args: T.List['TYPE_var'], kwargs: 'ItsJoinFile') -> ModuleReturnValue:
+        if not shutil.which('itstool'):
+            self.noitstool_error()
```
eli-schwartz, line 307: This seems like a weird thing to delegate to a dispatch function as it is only used once, and we should really just use the find_program implementation, which respects tools defined in cross files, then pass it as an argument to the helper script.

docs/markdown/i18n-module.md
eli-schwartz: #9867

PR #9745, opened 2021-12-19
docs: correct the shaderc documentation

Dudemanguy:
The docs don't mention this, but the shaderc system method also checks
shaderc_static if shaderc_combined is not found.

### DIFF
--- a/docs/markdown/Dependencies.md
+++ b/docs/markdown/Dependencies.md
@@ -653,11 +653,17 @@ or as an OSX framework.
 
 *(added 0.51.0)*
 
-Shaderc currently does not ship with any means of detection.
-Nevertheless, Meson can try to detect it using `pkg-config`, but will
-default to looking for the appropriate library manually. If the
-`static` keyword argument is `true`, `shaderc_combined` is preferred.
-Otherwise, `shaderc_shared` is preferred. Note that it is not possible
+Meson will first attempt to find shaderc using `pkg-config`. Upstream
+currently ships three different `pkg-config` files and by default will
+check them in this order: `shaderc`, `shaderc_combined`, and
+`shaderc_static`. If the `static` keyword argument is `true`, then
+Meson instead checks in this order: `shaderc_combined`, `shaderc_static`,
+and `shaderc`.
+
+If no `pkg-config` file is found, then Meson will try to detect the
+library manually. In this case, it will try to link against either
+`-lshaderc_shared` or `-lshaderc_combined`, preferring the latter
+if the static keyword argument is true. Note that it is not possible
 to obtain the shaderc version using this method.
 
 `method` may be `auto`, `pkg-config` or `system`.


eli-schwartz:
I think the docs are even more outdated than that, because since b510bc12ac55ffd074b32136802f59f599e14b9e we should be preferring pkg-config.

The full resolution here is:

Try to find pkg-config, as:
- prefer shared? shaderc.pc > shaderc_combined.pc > shaderc_static.pc
- prefer static? shaderc_combined.pc > shaderc_static.pc > shaderc.pc

Try to find_library (no versions available), as:
- prefer shared? `-lshaderc_shared` > `-lshaderc_combined`
- prefer static? `-lshaderc_combined` > `-lshaderc_shared`


The current state of the docs only describes how it performs raw library lookups. You're adding a reference to a pkg-config lookup.

This is doubly, triply, and quadruply confusing because of how naming works (or rather, doesn't work) upstream:
- shaderc_shared ---> is lib name
- shaderc_static ---> is pkg-config
- shaderc ---> is pkg-config, refers to shared
- shaderc_combined ---> is both, refers to static

eli-schwartz:
Would you be able to do a more comprehensive rework of the docs here? :)

Dudemanguy:
Woops, I forgot I opened this. I'll get to it.

Dudemanguy:
I think goofed up my rebase there which triggered a bunch of code review requests (sorry for the pings). I updated the commit now.

docs/markdown/Dependencies.md
```diff
@@ -653,12 +653,19 @@ or as an OSX framework.
...
-Nevertheless, Meson can try to detect it using `pkg-config`, but will
-default to looking for the appropriate library manually. If the
-`static` keyword argument is `true`, `shaderc_combined` is preferred.
-Otherwise, `shaderc_shared` is preferred. Note that it is not possible
-to obtain the shaderc version using this method.
+Meson will first attempt to find shaderc using `pkg-config`. Upstream
+currently ships three different `pkg-config` files and by default will
+check them in this order: `shaderc.pc`, `shaderc_combined.pc`, and
+`shaderc_static.pc`. If the `static` keyword argument is `true`, then
+Meson instead checks in this order: `shaderc_combined`, `shaderc_static`,
```
eli-schwartz, line 660: These two are the only places in the paragraph that don't include `.pc`, which is inconsistent.

(Although maybe they are redundant everywhere else.)

docs/markdown/Dependencies.md
```diff
@@ -653,12 +653,19 @@ or as an OSX framework.
...
+and `shaderc.pc`.
+
+If no `pkg-config` file is found, then Meson will try to detect the
+library manually. In this case, it will try to link against `-lshaderc_shared`
+first and then try `-lshaderc_combined`. If the `static` keyword argument
+is `true`, then `-lshaderc_combined` is preferred followed by checking
+`-lshaderc_shared`. Note that it is not possible to obtain the shaderc
```
eli-schwartz, line 667: I wonder if it might be clearer to emulate the previous docs, e.g.

> In this case, it will try to link against either `-lshaderc_shared` or `-lshaderc_combined`, preferring the latter if the `static` keyword argument is `true`.

Just trying to find some wording that involves less "enumerate the same things again, but in reverse order". Well-worded documentation is hard...

docs/markdown/Dependencies.md
Dudemanguy: Yup, I agree it's redundant.

docs/markdown/Dependencies.md
Dudemanguy: That sounds better to me too.

eli-schwartz: I believe you still want to mention "Note that it is not possible to obtain the shaderc version with this method."

Dudemanguy:
Updated the wording.

docs/markdown/Dependencies.md
Dudemanguy: crap

eli-schwartz: I mean, it could be worse. You could have done a long and involved rebase of a heavy refactoring series which accidentally dropped actual features. We have been there and done that.

Dudemanguy: The sentence is now undeleted.

eli-schwartz:
(Sorry for forgetting about this.)

PR #9751, opened 2021-12-21
modules: use find_program implementation to find programs

### LINKED ISSUE #9733: [0.60.2] import('cmake') cannot handle cmake not being on path.

eli-schwartz:
Do not use ExternalProgram as that is too low-level and doesn't handle e.g. machine file overrides.

Fixes #9733

### DIFF
--- a/mesonbuild/modules/cmake.py
+++ b/mesonbuild/modules/cmake.py
@@ -34,7 +34,6 @@ from ..interpreterbase import (
     InvalidArguments,
     InterpreterException,
 )
-from ..programs import ExternalProgram
 
 
 COMPATIBILITIES = ['AnyNewerVersion', 'SameMajorVersion', 'SameMinorVersion', 'ExactVersion']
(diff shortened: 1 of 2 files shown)


eli-schwartz:
/cc @mensinda 

Simple alternative to #9737

mensinda (approved):
Yeah, this is probably the better solution.

PR #9752, opened 2021-12-22
azure ci: Bump windows image and VS versions

nirbheek:
The Windows-2016 image is being deprecated [and will be removed in 2022](https://github.com/actions/virtual-environments/issues/4312). There is no image that contains Visual Studio 2017 anymore, and installing it manually will be too flaky and slow.

Testing of VS 2022 is more important than VS 2017 anyway, so bump VS 2019 jobs to VS 2022, and VS 2017 jobs to VS 2019.


nirbheek:
Hm, Visual Studio Enterprise is not added to PATH by default in the windows-2022 image, which is a good thing I guess, but needs us to change our jobs.

nirbheek:
... there's a LOT of test failures with VS 2022. Sigh.

nirbheek:
Let's try running everything on VS 2019 only for now. We're getting regular brownouts in the VS 2017 jobs (CI getting cancelled before it can even run). We can look at VS 2022 later.

jpakkane:
I tried to re-run the failed job and it would not do it. This probably needs a rebase.

nirbheek:
Rebased.

eli-schwartz:
The image was only removed for Github Actions. The Azure removal is currently tracked for removal on June 30 via https://github.com/actions/virtual-environments/issues/5403

Any update on getting things to play nicely on the updated images?

azure-pipelines.yml
```diff
@@ -38,55 +38,33 @@ variables:
...
+          compiler: msvc2019
+          backend: ninja
         vc2019x64ninja:
           arch: x64
           compiler: msvc2019
           backend: ninja
+        vc2019x86vs:
+          arch: x86
+          compiler: msvc2019
+          backend: ninja
```
eli-schwartz, line 59: Is this a typo and it's supposed to use the vs2019 backend?

nirbheek: Yes, this is a typo

eli-schwartz:
From googling around it seems like the clang-cl errors with lld-link not finding "kernel32.lib" "user32.lib" "gdi32.lib" "winspool.lib" "shell32.lib" "ole32.lib" "oleaut32.lib" "uuid.lib" "comdlg32.lib" "advapi32.lib" and suchlike, is because clang-cl can either build for the mingw ABI or the MSVC ABI, and it may be targeting the wrong ABI while not having the VC directory added manually to its LIB path.

eli-schwartz:
We're down to 2 weeks before final removal of the windows-2016 image.

Any luck in figuring this out?

eli-schwartz:
The final removal is in 3 days. Any luck?

eli-schwartz:
The real final removal finally happened after it stopped getting pushed off. It's over and done with.

I've opened #10653 to simply remove the vs2017 jobs for now.

cclauss:
What is the status of this?

PR #9757, opened 2021-12-26
Fix some broken type annotations

eli-schwartz:
And also a broken variable reference.

I discovered these problems while bored and trying to get cython to convert all of meson into compiled C modules. Cython actually uses type annotations at runtime, and also has a much better undefined variables scanner.


eli-schwartz:
Note: mypy 0.930 fails on git master.

PR #9759, opened 2021-12-26
unpin pylint in CI -- the mentioned issue is long since fixed

eli-schwartz:
And the outdated pin means some error messages that are a lot better in newer versions aren't available.


jpakkane:
The test failure seems relevant. Might make sense to fix it at the same time to keep master green.

eli-schwartz:
The test failure was in mypy, not pylint, and as such is irrelevant.

I rebased on git master, including commit 269337ceb26c8fd1ca2680ee7db9e82fdbf6b25c which fixed that check. So it should pass now.

PR #9761, opened 2021-12-27
Add the function exclude_from_coverage()

perror:
The function exclude_from_coverage() allow to pass a list of patterns to exclude them from the code coverage computation.

I am not sure that this code is optimal, so feel free to come with remarks and things to fix in order to make it better.

This patch is intended to fix issue #3287.

### DIFF
--- a/mesonbuild/scripts/coverage.py
+++ b/mesonbuild/scripts/coverage.py
@@ -17,7 +17,9 @@ from mesonbuild import environment, mesonlib
 import argparse, re, sys, os, subprocess, pathlib, stat
 import typing as T
 
-def coverage(outputs: T.List[str], source_root: str, subproject_root: str, build_root: str, log_dir: str, use_llvm_cov: bool) -> int:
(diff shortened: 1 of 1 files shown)


jpakkane:
One of the design goals of Meson is to keep the build files focused on building and keep all other configuration information away from them. It would be better if this would be the case here, too. Something like #9732 was better in this regard.

Even better would be if there were a standalone file like `.clang-format` et al that you could put in a standard location and that would work generally. I don't think there is a standard for that, though. :(

perror:
@jpakkane In fact, #9732 is not far, it is just the diff on `mesonbuild/scripts/coverage.py`, I just extended the code to accept a list of patterns.

But, if I understand well the philosophy of meson, the code of `mesonbuild/scripts/coverage.py` should also look for `.lcovrc` or `gcovr.cfg` file at `self.environment.source_dir` and take it in consideration.

Is it correct ?

perror:
@jpakkane I think I got your point, I will come back with a more global and headless solution. Thanks for your helpful message.

PR #9773, opened 2022-01-03
Add a man page backend to refman

annacrombie:
With this pr you can pass `-g man` to genrefman.py and get one big man page with all information that refman knows about.  I haven't integrated this into the build yet since I'm not sure what would be best.

### DIFF
--- a/docs/markdown/snippets/refman_man_page_backend.md
+++ b/docs/markdown/snippets/refman_man_page_backend.md
@@ -1,0 +1,4 @@
+## Add a man page backend to refman
+
+The refman docs (function and object reference) can now be generated as a man
+page.
--- a/docs/refman/generatorman.py
+++ b/docs/refman/generatorman.py
@@ -1,0 +1,382 @@
+import re
+from pathlib import Path
+
+from .generatorbase import GeneratorBase
+from .model import (
+    ReferenceManual,
+    Function,
+    Object,
+    PosArg,
+    VarArgs,
+    Kwarg,
+)
+
+import typing as T
+
+
+class ManPage:
+    def __init__(self, path: Path):
+        self.path = path
+        self.text = ""
+
+    def reset_font(self) -> None:
+        self.text += ".P\n"
+
+    def title(self, name: str, section: int) -> None:
+        import datetime
+
+        date = datetime.date.today()
+        self.reset_font()
+        self.text += f'.TH "{name}" "{section}" "{date}"\n'
+
+    def section(self, name: str) -> None:
+        self.reset_font()
+        self.text += f".SH {name}\n"
+
+    def subsection(self, name: str) -> None:
+        self.reset_font()
+        self.text += f".SS {name}\n"
+
+    def par(self, text: str) -> None:
+        self.reset_font()
+        self.text += f"{text}\n"
+
+    def indent(self, amount: int = 4) -> None:
+        self.text += f".RS {amount}\n"
+
+    def unindent(self) -> None:
+        self.text += ".RE\n"
+
+    def br(self) -> None:
+        self.text += ".br\n"
+
+    def nl(self) -> None:
+        self.text += "\n"
+
+    def line(self, text: str) -> None:
+        if text and text[0] in [".", "'"]:
+            self.text += "\\"
+        self.text += f"{text}\n"
+
+    def inline(self, text: str) -> None:
+        self.text += f"{text}"
+
+    def write(self) -> None:
+        self.path.write_text(self.text, encoding="utf-8")
+
+    @staticmethod
+    def bold(text: str) -> str:
+        return f"\\fB{text}\\fR"
+
+    @staticmethod
+    def italic(text: str) -> str:
+        return f"\\fI{text}\\fR"
+
+
+class GeneratorMan(GeneratorBase):
+    def __init__(
+        self, manual: ReferenceManual, out: Path, enable_modules: bool
+    ) -> None:
+        super().__init__(manual)
+        self.out = out
+        self.enable_modules = enable_modules
+        self.links: T.List[str] = []
+
+    def generate_description(self, page: ManPage, desc: str) -> None:
+        def italicise(match: T.Match[str]) -> str:
+            v = match.group(1)
+            if v[0] == "@":
+                v = v[1:]
+
+            return ManPage.italic(v)
+
+        desc = re.sub(re.compile(r"\[\[(.*?)\]\]", re.DOTALL), italicise, desc)
+
+        def linkify(match: T.Match[str]) -> str:
+            replacement = ManPage.italic(match.group(1))
+
+            if match.group(2)[0] != "#":
+                if match.group(2) in self.links:
+                    num = self.links.index(match.group(2))
+                else:
+                    self.links.append(match.group(2))
+                    num = len(self.links)
+
+                replacement += f"[{num}]"
+
+            return replacement
+
+        desc = re.sub(re.compile(r"\[(.*?)\]\((.*?)\)", re.DOTALL), linkify, desc)
+
+        def bold(match: T.Match[str]) -> str:
+            return ManPage.bold(match.group(1))
+
+        desc = re.sub(re.compile(r"\*(.*?)\*"), bold, desc)
+
+        isCode = False
+        for chunk in desc.split("```"):
+            if isCode:
+                page.indent()
+                lines = chunk.strip().split("\n")
+                if lines[0] == "meson":
+                    lines = lines[1:]
+
+                for line in lines:
+                    page.line(line)
+                    page.br()
+                page.unindent()
+            else:
+                inList = False
+                for line in chunk.strip().split("\n"):
+                    if len(line) == 0:
+                        page.nl()
+                        if inList:
+                            page.nl()
+                            inList = False
+                    elif line[0:2] in ["- ", "* "]:
+                        if inList:
+                            page.nl()
+                            page.br()
+                        else:
+                            inList = True
+
+                        page.inline(line.strip() + " ")
+                    elif inList and line[0] == " ":
+                        page.inline(line.strip() + " ")
+                    else:
+                        inList = False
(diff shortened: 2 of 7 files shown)


eli-schwartz:
That looks pretty neat!

I tested it out locally and built the reference manual as https://eschwartz.srht.site/meson-reference.3.html

I notice that function headers look a bit unusual, e.g.

```
add_languages(Language..., native:, required:) -> bool
```

which might make it a teensy bit hard to track down a particular function. e.g. the hotdoc generator produces section headers with just the name, followed by a function description naming its argument calling convention. Not sure how to provide both. :/

For some reason, that particular part is missing the type value, even though it includes the `:` part. :p

annacrombie:
> I notice that function headers look a bit unusual, e.g.
> ```
> add_languages(Language..., native:, required:) -> bool
> ```

I modeled this after the python documentation style ([example](https://docs.python.org/3/library/functions.html)) where the function signature is given in a concise way as the header for each function.  I agree it looks bad in the html version, but I think it looks pretty good as a man page.  I don't see how it makes looking for the function any more difficult, all you go is grep for the functions name.

> For some reason, that particular part is missing the type value

I intentionally omitted types from arguments in the signature, because otherwise it would be way too long.

eli-schwartz:
> I modeled this after the python documentation style ([example](https://docs.python.org/3/library/functions.html)) where the function signature is given in a concise way as the header for each function. I agree it looks bad in the html version, but I think it looks pretty good as a man page. I don't see how it makes looking for the function any more difficult, all you go is grep for the functions name.

I would argue that both_libraries() looks quite bad in `man ./meson-reference.3`. :p

Using grep (or rather, `/<token>`) helps but visually it is still a bit awkward to decipher once you do find it. In the end I basically just glossed over everything following the `(`

> I intentionally omitted types from arguments in the signature, because otherwise it would be way too long.

Speaking of the python docs, the python docs don't provide the type info in function signatures, but they do provide the default value.

tristan957:
Any possible way to indent the arguments past the name of the function if it takes more than one line? Think that would help with readability.

tristan957:
Other than that not entirely sure how best to clean it up unless you moved the arguments out of the function name into a table similar to how the online docs have it.

annacrombie:
The function signature isn't really necessary, all the information is included in the argument sections below.  I think it is useful for the simpler functions like assert(), but not for things like the build target definition functions.  Maybe if there are more than N kwargs I should just have something like `kwargs...`?

docs/refman/generatorman.py
```diff
@@ -0,0 +1,375 @@
...
+        self.out = out
+        self.enable_modules = enable_modules
+        self.links: T.List[str] = []
+
+    def _extract_meson_version(self) -> str:
+        # Hack around python relative imports to get to the Meson version
+        import sys
+
+        sys.path.append(Path(__file__).resolve().parents[2].as_posix())
+        from mesonbuild.coredata import version
```
eli-schwartz, line 91: In setuptools we handle this via:

```ini
[metadata]
version = attr: mesonbuild.coredata.version
```

Which (relies on the import system to find a filename and) reads the contents of a given file to extract an attribute without actually importing. See `from setuptools.config import StaticModule`.

Extracting the actual important bits from their importlib entanglement, you can get the version like this:

```python
import ast

with open(Path(__file__).resolve().parents[2] / 'mesonbuild/coredata.py', 'r') as f:
    src = f.read()

astdata = ast.parse(src)
version = next(
    ast.literal_eval(statement.value)
    for statement in astdata.body
    if isinstance(statement, ast.Assign)
    for target in statement.targets
    if isinstance(target, ast.Name) and target.id == 'version'
)
```

annacrombie: I shamelessly copied that function from `generatorjson.py` :). https://github.com/mesonbuild/meson/blob/09f52a3817126bc1dc9d1b9d93d531505801a59e/docs/refman/generatorjson.py#L95

I wonder if there was any particular reason @mensinda decided to write it that way.  Your version seems just a little bit hacky as well. In any case, the duplication is probably bad and should be consolidated!

eli-schwartz: This hack is an elegant methodology for parsing and iterating over an AST. :p 

No need to mess with sys.path, avoid assuming that anything is importable (although it should be) -- it's generally a nice clean way to avoid runtime requirements to retrieve version info, which is why setuptools implements it.

Yes, it should probably be moved to a more accessible location.

annacrombie: So what should be done for this PR?  Should I implement your AST method in `generatorman.py` and be done with it?  I'm not really sure where an accessible version getter function should go, and it probably wouldn't be in scope for this PR anyway.

eli-schwartz: You could put it as a staticmethod in generatorbase.py I guess.

Actually I'm a bit confused by the fact that it already imports mesonbuild.mlog a bunch of times, apparently from a previous system install of meson.

mensinda: For the `mlog` import, there is another `path` hack [here](https://github.com/mesonbuild/meson/blob/master/docs/genrefman.py#L18-L21). So no existing meson install should be required. As for the `_extract_meson_version` the additional path hack is completely unnecessary and just importing `mesonbuild.coredata` works. Seems like I forgot about the initial `sys.path` hack when I wrote the JSON generator...

Since everything from `mesonbuild` is importable already, I don't see a reason for using the `ast` workaround, when we can just import the version. In general, this hack is just a workaround to tell python: Sorry, I didn't put the root script in the git root directory because it doesn't belong there, but trust me this is still our project code and this import is fine.

To clarify, I was using `mlog` so that I could use the existing Meson printing helpers (color, nested, etc.) without duplicating any code or introducing another external dependency just for that.

mensinda: So basically, this should IMHO be fine in `GeneratorBase`, since `mlog` is already being used.

```python
    @staticmethod
    def _extract_meson_version() -> str:
        from mesonbuild.coredata import version
        return version
```

@eli-schwartz If you really don't like the `sys.path` hack, then we can try to find another solution.

eli-schwartz: Ah, I only looked in the refman/ directory...

Using mlog is sufficient rationale for importing mesonbuild for real. The sys.path hack is fine. As such, we might as well use it for the version too.

If we were *only* doing it for the version, I'd probably opt for the ast hack instead.

mensinda (changes requested):
Overall, this looks really good and thanks for working on this. However, the function signature logic needs some work though IMHO.

In general, I am not sure if it is a good idea to have single line function signatures, since they can get really long, especially for build targets.

Are there any plans to also include the method documentation for the objects?

mensinda:
As for integrating this into the current CI, just add a step for the man generator around [here](https://github.com/mesonbuild/meson/blob/master/docs/meson.build#L65). It also shouldn't hurt to automatically release it similar to the [JSON manual](https://github.com/mesonbuild/meson/blob/master/.github/workflows/website.yml#L57-L63). I also don't know what else to do with it currently.

This PR should also have a release note snippet.

annacrombie:
Thanks for the feedback!  I applied all your suggestions to the function signatures, as well as putting them under a SYNOPSIS sub-header, using just the function's name as the section header. 

As far as what to do with this, I really want to have it be installed along with [muon](https://muon.build), and potentially `meson` too if that seems desirable.  In meson's case it should be easy to integrate with the existing way `meson.1` gets installed.  For `muon` I'm not sure what to do at this point other than using the entire meson repo as a subproject and patching in a `meson.build` to pull out the man page.  It would also be nice to put the hotdoc docs behind an option.

> This PR should also have a release note snippet.

Where would this go?  Should I just make a section at the bottom of `markdown/Release-notes-for-0.60.0.md`?

eli-schwartz:
We use a script genrelnotes.py to produce release notes at the time of release. Release notes snipptlets are added, one file per feature, in markdown/snippets/ so they don't clash with other PRs also trying to edit a single file.

(There are a number of implementations of this general concept, e.g. https://towncrier.readthedocs.io/en/actual-freaking-docs/)

tristan957:
@annacrombie create a markdown file in this folder: https://github.com/mesonbuild/meson/tree/master/docs/markdown/snippets as a part of your PR.

tristan957:
Might be nice if functions were suffixed with `()`.

`infinity times` could maybe just be `N`. I feel like `0..N` is pretty well understood.

I think if the synopsis spans multiple lines, everything after the first should be indented (if possible).

I like the indenting idea @eli-schwartz mentioned, but it would need to be consistent with `kwargs`. I think the extra indentation would go a long way if someone is reading in quick glances.

docs/markdown/snippets/refman_man_page_backend.md
```diff
@@ -0,0 +1,4 @@
+## Add a man page backend to refman
+
+The refman docs (funciton and object reference) can now be generated as a man
```
eli-schwartz, line 3:
```suggestion
The refman docs (function and object reference) can now be generated as a man
```

annacrombie: :sweat_smile:

eli-schwartz:
Updated again. I think this looks pretty nice, now.

tristan957:
It does look good! I still think `library()` looks a little jank, but if that is all that is possible, then I can live with it. :)

mensinda (approved):
This LGTM code wise and is ready to go from my side when everyone else is happy with the generated man page.

@jpakkane Since this is a docs only change, this should be good to go for 0.61.0 IMHO.

annacrombie:
, but other than that I agree that it is looking pretty good.

Are you referring to @eli-schwartz's hosted version?  If so, I have made one more small change that breaks up long signatures into multiple lines that isn't on there yet.

tristan957:
@annacrombie I was referring to that version. Hope to see the latest and greatest soon!

eli-schwartz:
I updated it yet again, this time for the final version merged to git master.

PR #9785, opened 2022-01-07
Detect visual studio express edition

DFOVIT:
vswhere.exe does not find vs express when using only the workload 'Microsoft.VisualStudio.Component.VC.Tools.x86.x64'.
This adds the 'Microsoft.VisualStudio.Workload.WDExpress' worlkload, like in [setuptools](https://github.com/pypa/setuptools/blob/40725756391163a7557113b9922d2086506b9db0/setuptools/msvc.py#L176).
Verified ok on my system.


nirbheek:
Could you please fix the commit message? It does not describe the changes at all. At the very least your PR description should also be in the commit message.

PR #9788, opened 2022-01-07
python module: add option to specify a python environment to install to

eli-schwartz:
The default behavior of installing relative to prefix may be unexpected, and is definitely wrong in many cases.

Give users control in order to specify that yes, they actually want to install to a venv.

This is particularly useful for projects that use meson as a build system for a python module, where *all* files shall be installed into the python site-packages.

### DIFF
--- a/docs/markdown/Builtin-options.md
+++ b/docs/markdown/Builtin-options.md
@@ -271,10 +271,11 @@ name with the module name: `-D<module>.<option>=<value>` (e.g. `-Dpython.platlib
 
 ### Python module
 
-| Option           | Default value | Possible values | Description |
-| ------           | ------------- | --------------- | ----------- |
-| platlibdir       |               | Directory path  | Directory for site-specific, platform-specific files (Since 0.60.0) |
-| purelibdir       |               | Directory path  | Directory for site-specific, non-platform-specific files  (Since 0.60.0) |
+| Option           | Default value | Possible values             | Description |
+| ------           | ------------- | -----------------           | ----------- |
+| install_env      | prefix        | {auto,prefix,system,venv}   | Which python environment to install to (Since 0.62.0) |
+| platlibdir       |               | Directory path              | Directory for site-specific, platform-specific files (Since 0.60.0) |
+| purelibdir       |               | Directory path              | Directory for site-specific, non-platform-specific files  (Since 0.60.0) |
 
 *Since 0.60.0* `python.platlibdir` and `python.purelibdir` options are used by
 python module methods `python.install_sources()` and `python.get_install_dir()`.
@@ -283,3 +284,14 @@ relative to the installation `prefix`, which will often result in installed pyth
 modules to not be found by the interpreter unless `prefix` is `/usr` on Linux,
 or for example `C:\Python39` on Windows. These options can be absolute paths
 outside of `prefix`.
+
+*Since 0.62.0* The `python.install_env` option is used to detect the correct
+installation path. Setting to `system` will avoid making the paths relative to
+`prefix` and instead use the global site-packages of the selected python
+interpreter directly, even if it is a venv. Setting to `venv` will instead use
+the paths for the virtualenv the python found installation comes from (or fail
+if it is not a virtualenv).  Setting to `auto` will check if the found
+installation is a virtualenv, and use `venv` or `system` as appropriate (but
+never `prefix`). This option is mutually exclusive with the `platlibdir`/`purelibdir`.
+
+For backwards compatibility purposes, the default `install_env` is `prefix`.
--- a/docs/markdown/snippets/python_module_env.md
+++ b/docs/markdown/snippets/python_module_env.md
@@ -1,0 +1,9 @@
+## New option to choose python installation environment
+
+It is now possible to specify `-Dpython.install_env` and choose how python modules are installed.
+
+- `venv`: assume that a virtualenv is active and install to that
+- `system`: install to the global site-packages of the selected interpreter
+  (the one that the venv module calls --system-site-packages)
+- `prefix`: preserve existing behavior
+- `auto`: autodetect whether to use venv or system
--- a/mesonbuild/modules/python.py
+++ b/mesonbuild/modules/python.py
@@ -338,11 +338,13 @@ variables.update({'base_prefix': getattr(sys, 'base_prefix', sys.prefix)})
 print(json.dumps({
   'variables': variables,
   'paths': paths,
+  'sysconfig_paths': sysconfig.get_paths(),
   'install_paths': install_paths,
   'sys_paths': sys.path,
   'version': sysconfig.get_python_version(),
   'platform': sysconfig.get_platform(),
   'is_pypy': '__pypy__' in sys.builtin_module_names,
+  'is_venv': sys.prefix != variables['base_prefix'],
   'link_libpython': links_against_libpython(),
 }))
 '''
@@ -352,7 +354,9 @@ if T.TYPE_CHECKING:
 
         install_paths: T.Dict[str, str]
         is_pypy: bool
+        is_venv: bool
         link_libpython: bool
+        sysconfig_paths: T.Dict[str, str]
         paths: T.Dict[str, str]
         platform: str
         suffix: str
@@ -377,7 +381,9 @@ class PythonExternalProgram(ExternalProgram):
         self.info: 'PythonIntrospectionDict' = {
             'install_paths': {},
             'is_pypy': False,
+            'is_venv': False,
             'link_libpython': False,
+            'sysconfig_paths': {},
             'paths': {},
             'platform': 'sentinal',
             'variables': {},
@@ -422,7 +428,22 @@ class PythonExternalProgram(ExternalProgram):
             return rel_path
         value = state.get_option(f'{key}dir', module='python')
         if value:
+            if state.is_user_defined_option('install_env', module='python'):
+                raise mesonlib.MesonException(f'python.{key}dir and python.install_env are mutually exclusive')
             return value
+
+        install_env = state.get_option('install_env', module='python')
+        if install_env == 'auto':
+            install_env = 'venv' if self.info['is_venv'] else 'system'
+
+        if install_env == 'system':
+            rel_path = os.path.join(self.info['variables']['prefix'], rel_path)
+        elif install_env == 'venv':
+            if not self.info['is_venv']:
+                raise mesonlib.MesonException('python.install_env cannot be set to "venv" unless you are in a venv!')
+            # inside a venv, deb_system is *never* active hence info['paths'] may be wrong
+            rel_path = self.info['sysconfig_paths'][key]
+
         # Use python's path relative to prefix, and warn if that's not a location
         # python will lookup for modules.
         abs_path = Path(state.get_option('prefix'), rel_path)
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -1218,6 +1218,8 @@ BUILTIN_CORE_OPTIONS: 'KeyedOptionDictType' = OrderedDict([
     (OptionKey('force_fallback_for'), BuiltinOption(UserArrayOption, 'Force fallback for those subprojects', [])),
 
     # Python module
+    (OptionKey('install_env', module='python'),
+     BuiltinOption(UserComboOption, 'Which python environment to install to', 'prefix', choices=['auto', 'prefix', 'system', 'venv'])),
     (OptionKey('platlibdir', module='python'),
      BuiltinOption(UserStringOption, 'Directory for site-specific, platform-specific files.', '')),
     (OptionKey('purelibdir', module='python'),
--- a/mesonbuild/modules/__init__.py
+++ b/mesonbuild/modules/__init__.py
@@ -103,6 +103,13 @@ class ModuleState:
                    module: T.Optional[str] = None) -> T.Union[str, int, bool, 'WrapMode']:
         return self.environment.coredata.get_option(mesonlib.OptionKey(name, subproject, machine, lang, module))
 
+    def is_user_defined_option(self, name: str, subproject: str = '',
+                               machine: MachineChoice = MachineChoice.HOST,
+                               lang: T.Optional[str] = None,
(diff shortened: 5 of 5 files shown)


eli-schwartz:
@jpakkane what are your thoughts on merging this for 0.61?

Although it's past rc1, I forgot to submit this first... :( and it is behavior-preserving but should make things a lot more comfortable for python modules adopting meson, e.g. script.

I believe that this change should not cause surprises or breakage as it's very narrowly scoped. But it should definitely go in an rc2 first.

docs/markdown/snippets/python_module_env.md
```diff
@@ -0,0 +1,8 @@
+## New option to choose python installation environment
+
+It is now possible to specify `-Dpython.install_env` and choose how python modules are installed.
+
+- `venv`: assume that a virtualenv is active and install to that
+- `system`: install to the global site-packages
```
rgommers, line 6: This may be hard to understand (it could be the conda/homebrew/etc one or the actual system provided one - you mean the former here). How about something like _"install to the site-packages of the active Python interpreter (if a virtualenv, this means the Python it is derived from)"_?

docs/markdown/Builtin-options.md
```diff
@@ -281,3 +282,12 @@ relative to the installation `prefix`, which will often result in installed pyth
 modules to not be found by the interpreter unless `prefix` is `/usr` on Linux,
 or for example `C:\Python39` on Windows. These options can be absolute paths
 outside of `prefix`.
+
+*Since 0.61.0* The `python.install_env` option is used to detect the correct
+installation path. Setting to `system` will avoid making the paths relative to
+`prefix`. Setting to `venv` will instead use the paths for the virtualenv the
```
rgommers, line 288: Also needs explanation of what `system` actually does do (see my comment on line 6 of `python_module_env.md`).

rgommers (approved):
Thanks @eli-schwartz! This works like a charm. It'd be great to get this in `61.0`, because it makes it possible to install a Python package into site-packages correctly without manually setting the prefix to the active Python interpreter's `lib` directory.

I've tested all four options within a `conda` environment on Arch Linux, and it all works as expected.

It also takes care of avoiding these warnings in my development workflow:
```
WARNING: Python files installed by Meson might not be found by python interpreter.
 This warning can be avoided by setting "python.platlibdir" option.
WARNING: Python files installed by Meson might not be found by python interpreter.
 This warning can be avoided by setting "python.purelibdir" option.
```

My only code review comment (not large/blocking) is to make the description of `system` somewhat clearer. 

Two other thoughts/questions:

What do you think about this superceding the separate `purelibdir`/`platlibdir` options? I'm not sure there's a use for them after this feature lands, and they're badly thought out and Linux-specific concepts to begin with.

Regarding the default, it's understandable (but a little unfortunate) that it has to be `prefix` because of backwards compat. For any Python package that's always wrong (hence `auto` will never select `prefix`). But given that most Python devs will not be using the `meson` CLI but `pip install .` or some such thing, it's hopefully not too much of a problem in practice.

docs/markdown/snippets/python_module_env.md
eli-schwartz: I think maybe I was thinking of that when I used the word "global", which has the connotation, to me, of "non-venv" but still within the interpreter.

What about this change?

```diff
-- `system`: install to the global site-packages
+- `system`: install to the global site-packages of the selected interpreter
+  (the one that the venv module calls --system-site-packages)
```

eli-schwartz:
> What do you think about this superceding the separate purelibdir/platlibdir options? I'm not sure there's a use for them after this feature lands, and they're badly thought out and Linux-specific concepts to begin with.

I'm not entirely sure what the precise use case for them would be either. Although I guess in theory they serve a similar role as --prefix and friends if you want to control install paths with absolute precision.

> Regarding the default, it's understandable (but a little unfortunate) that it has to be `prefix` because of backwards compat. For any Python package that's always wrong (hence `auto` will never select `prefix`). But given that most Python devs will not be using the `meson` CLI but `pip install .` or some such thing, it's hopefully not too much of a problem in practice.

I'd be willing to discuss changing the default in the future. But I do think that we need the option in the end, because users may use `--prefix=$HOME/.local/` to install as non-root, and this will do the correct thing (unless perhaps Debian makes a hash of dist-packages again? I'm not sure) by installing to the `posix_user` installation scheme in `~/.local/lib/pythonX.Y/site-packages/`.

rgommers:
> But I do think that we need the option in the end

Yes I agree that `prefix` has use cases. I'm still using it to install a development build in-tree right now (it escapes me for a minute why it works better than `destdir`, but it does at the moment).

docs/markdown/snippets/python_module_env.md
rgommers: That looks good, it's quite clear.

jpakkane:
Policy says that after rc1 only bug fixes are allowed, no new features. So sadly this will have to wait until the next release.

mesonbuild/modules/python.py
```diff
@@ -410,6 +416,19 @@ def _get_path(self, state: T.Optional['ModuleState'], key: str) -> None:
         value = state.get_option(f'{key}dir', module='python')
         if value:
             return value
+
+        install_env = state.get_option('install_env', module='python')
+        if install_env == 'auto':
+            install_env = 'venv' if self.info['is_venv'] else 'system'
+
+        if install_env == 'system':
+            rel_path = os.path.join(self.info['variables']['prefix'], rel_path)
```
xclaesse, line 425: Do I understand correctly that in the `system` and `venv` cases, `rel_path` is actually an absolute path? Otherwise it's going to prepend prefix, right? Maybe change the variable name to reflect that.

docs/markdown/Builtin-options.md
```diff
@@ -281,3 +282,14 @@ relative to the installation `prefix`, which will often result in installed pyth
 modules to not be found by the interpreter unless `prefix` is `/usr` on Linux,
 or for example `C:\Python39` on Windows. These options can be absolute paths
 outside of `prefix`.
+
+*Since 0.61.0* The `python.install_env` option is used to detect the correct
```
xclaesse, line 286: It's will have to wait for 0.62.0 sadly, right?

xclaesse:
> What do you think about this superceding the separate purelibdir/platlibdir options? I'm not sure there's a use for them after this feature lands, and they're badly thought out and Linux-specific concepts to begin with.

I think they are still needed for the case you install into prefix, which is still what you want on Linux. It is not linux-only neither, as long as you know to set PYTHONPATH when running your app, it's fine to install into your app's prefix even on Windows. I personally would even recommend doing that TBH.

rgommers:
> I think they are still needed for the case you install into prefix,

I'm talking about `-Dpython.platlibdir` given on the command line as an option, see https://mesonbuild.com/Builtin-options.html#module-options. That's an absolute path, typically outside the prefix. It was only added in `0.60.0` as a workaround, and really doesn't make too much sense - any use case we can think of should be superceded by this PR. 

I think you are talking here about the regular purelib/platlib specified as `pure: true/false` in `meson.build`.

xclaesse:
> That's an absolute path, typically outside the prefix

They could be relative to prefix too, IIRC. But seems that's not actually documented that way...

eli-schwartz:
install_dir can always be absolute or relative, and in the latter case it is relative to prefix. And I am pretty sure all path options other than prefix either *can* be relative or *must* be relative, and do so in relation to prefix.


As I said above,
> Although I guess in theory they serve a similar role as --prefix and friends if you want to control install paths with absolute precision.

Perhaps users are baking an installer package whose launcher sets up PYTHONPATH, idk. I haven't given much thought to it yet.

rgommers:
> Perhaps users are baking an installer package whose launcher sets up PYTHONPATH, idk. I haven't given much thought to it yet.

I use `PYTHONPATH` all the time, it's a very useful env var - but it has very little to do with whether the package as a whole is pure Python or not. The way Python packages work 99.x% of the time is that you have a single tree of files which get installed to `site-packages/package_name/` - or some arbitrary destdir, but that is then still not dependent on whether you have an extension module inside your package or not. 

Okay, since it's not directly related to this PR, I'll shut up about it now. I just know I'm going to have to talk about Meson to a ton of people over the next year, so 2 out of 3 Python specific CLI options (after this PR is merged) being irrelevant to all of those people is not too helpful.

tristan957:
This whole situation and the band-aids on top of it are irrelevant in my opinion because Meson shouldn't be in the business of warning people about where they are installing files. Meson going out of its way to make bad Python distributors happy is only hurting Meson. Python developers/users are well aware of PYTHONPATH and can use it appropriately when installing outside of places that the Python interpreter can't, by default, see. Is there anyone outside of Debian who has even complained about this? I still don't even understand the necessity. Is there a bug report I can read?

mesonbuild/modules/python.py
eli-schwartz: It is still treated like a relative path, as its purpose is basically to be joined with `get_option('prefix')` and assigned as abs_path further down.


What do you recommend instead?

`can_be_rel_path`?
`ans`?

mesonbuild/modules/python.py
```diff
@@ -410,6 +416,19 @@ def _get_path(self, state: T.Optional['ModuleState'], key: str) -> None:
...
+        elif install_env == 'venv':
+            if not self.info['is_venv']:
+                raise mesonlib.MesonException('python.install_env cannot be set to "venv" unless you are in a venv!')
+            # inside a venv, deb_system is *never* active hence info['paths'] may be wrong
+            rel_path = self.info['sysconfig_paths'][key]
+
         # Use python's path relative to prefix, and warn if that's not a location
         # python will lookup for modules.
         abs_path = Path(state.get_option('prefix'), rel_path)
```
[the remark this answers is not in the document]
eli-schwartz: install_env cannot suppress the warning though, it merely gives meson a better hint about which heuristic path to choose.

Given the warning is really just about Debian's brokenness, it's not actually unlikely that Debian's bad patching results in actual sysconfig paths that aren't on the pythonpath.

So I believe it is inappropriate to update the warning below, unless setting `python.install_env` actually returns early, skipping the warning check altogether.

docs/markdown/Builtin-options.md
```diff
@@ -281,3 +282,14 @@ relative to the installation `prefix`, which will often result in installed pyth
...
+installation path. Setting to `system` will avoid making the paths relative to
+`prefix` and instead use the global site-packages of the selected python
+interpreter directly, even if it is a venv. Setting to `venv` will instead use
+the paths for the virtualenv the python found installation comes from (or fail
+if it is not a virtualenv).  Setting to `auto` will check if the found
+installation is a virtualenv, and use `venv` or `system` as appropriate (but
+never `prefix`).
```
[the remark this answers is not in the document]
eli-schwartz: In the sense that it's invalid to manually define both options at the same time, since they conceptually conflict?

The "correct" way to do that would be to check whether `python.install_env` has been set, rather than checking if its value is `prefix`, but of course we can't actually do that, right? We don't know whether an option has been defined on the command line. :/

docs/markdown/Builtin-options.md
eli-schwartz: Yeah, just never rebased it yet.

eli-schwartz:
> Is there anyone outside of Debian who has even complained about this? I still don't even understand the necessity. Is there a bug report I can read?

- No
- https://github.com/mesonbuild/meson/issues/8739

NGL, I really am incredibly tempted to remove the warning altogether. Especially given that ever since commit 05b5a1e56fe8f5400b65d0d69680cc6531fe74f8 we should not be getting this wrong (or not any more so on Debian than any environment in which install_env=prefix is problematic). So it really feels like the warning has run its course.

tristan957:
We should just force Debian to keep a Meson patch to fix this. Let them live with their decisions.

eli-schwartz:
@FRidh could confirm it, but my understanding is that NixOS also needs the status quo behavior or explicit `-Dpython.install_env=prefix` to return to the status quo, since they have different per-package envs and sysconfig doesn't communicate that information.

FRidh:
> @FRidh could confirm it, but my understanding is that NixOS also needs the status quo behavior or explicit -Dpython.install_env=prefix to return to the status quo, since they have different per-package envs and sysconfig doesn't communicate that information.

That seems correct to me. sysconfig does not contain the correct prefix for us so we need to replace it with a prefix whenever we install something into a Nix store path. See e.g. the following solution https://github.com/NixOS/nixpkgs/pull/105714/files#diff-6845adc675c162d94bec555a24f0486cb2688af7fdd7f7627a25dd53a9a7a23fR75.

mesonbuild/modules/python.py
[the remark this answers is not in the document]
xclaesse: The warning happens on Windows too, IIRC. Setting `python.install_env` should avoid the warning IMHO.

docs/markdown/Builtin-options.md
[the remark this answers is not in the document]
xclaesse: We have `interpreter.user_defined_options` for that.

mesonbuild/modules/python.py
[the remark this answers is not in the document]
eli-schwartz: Should, but isn't guaranteed...

I'm worried about printing "you can avoid this warning by setting install_env" and then people actually setting this and still getting warnings because they have a bizarre patched sysconfig, which we *think* is only the case on Debian.

Conditionally mentioning to use install_env depending on whether we guess that platform is bizarre, feels broken and implies we should conditionally change our detection logic instead, and maybe get rid of the warning. :p

docs/markdown/Builtin-options.md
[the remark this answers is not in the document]
eli-schwartz: ... I did not realize that was a thing. That being said, it feels like a rather inelegant way of approaching the get_option problem, and also... do we want to add more poking at the interpreter from inside a module?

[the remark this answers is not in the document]
eli-schwartz: Hopefully implemented.

eli-schwartz:
xclaesse gave a lgtm via IRC.

PR #9789, opened 2022-01-07
wraptool: status: Include filename in error message if wrap is not on wrapdb.

### LINKED ISSUE #9787: `meson wrap status` with non-wrapdb wrap should have better message
**Describe the bug**
When you call `meson wrap status` and you have one or more wrap files which are not from wrapdb, the following line will print

`Wrap file not from wrapdb`

It would be better if it described which wrap file it is referring to.

**To Reproduce**
Omitting because it's obvious

**Expected behavior**
Print something akin to `Wrap file 'userwrap.wrap' is not from wrapdb`

sielicki:
Would close #9787

Signed-off-by: Nicky Sielicki <sielicki@yandex.com>

### DIFF
--- a/mesonbuild/wrap/wraptool.py
+++ b/mesonbuild/wrap/wraptool.py
@@ -212,7 +212,7 @@ def status(options: 'argparse.Namespace') -> None:
         try:
             (current_branch, current_revision, _, _, _) = get_current_version(w)
         except Exception:
-            print('Wrap file not from wrapdb.', file=sys.stderr)
+            print('Wrap file "{}" not from wrapdb.'.format(name), file=sys.stderr)
             continue
         if current_branch == latest_branch and current_revision == latest_revision:
             print('', name, f'up to date. Branch {current_branch}, revision {current_revision}.')


mesonbuild/wrap/wraptool.py
```diff
@@ -212,7 +212,7 @@ def status(options: 'argparse.Namespace') -> None:
         try:
             (current_branch, current_revision, _, _, _) = get_current_version(w)
         except Exception:
-            print('Wrap file not from wrapdb.', file=sys.stderr)
+            print('Wrap file "{}" not from wrapdb.'.format(name), file=sys.stderr)
```
eli-schwartz, line 215:
```suggestion
            print(f'Wrap file {name!r} not from wrapdb.', file=sys.stderr)
```

f-strings are faster and often more concise than format functions.

eli-schwartz (changes requested):

sielicki:
fixed in fc5dfb852b

Closing.

eli-schwartz:
Oh huh, so it was. :o

Issue #9791, opened 2022-01-08
Qt 6 host binaries have moved to QT_HOST_LIBEXECS in Qt 6.1+


jlindgren90:
**Describe the bug**
meson fails to find the `moc`, `rcc`, and `uic` binaries from Qt 6.1+ because they have moved from `QT_HOST_BINS` to `QT_HOST_LIBEXECS`.

**To Reproduce**
`meson.build`:
```
project('test', 'c')
qt6 = import('qt6')
qt6.preprocess(qresources: 'test.qrc')
```

Run `QMAKE=qmake6 meson build`.
Partial output:
```
Run-time dependency qt6 (modules: Core) found: YES 6.2.2 (qmake)
Program /usr/lib/qt6/bin/moc found: NO
Program moc-qt6 found: NO
Program moc found: NO found 5.15.2 but need: '== 6.2.2' (/usr/bin/moc)
Program /usr/lib/qt6/bin/uic found: NO
Program uic-qt6 found: NO
Program uic found: NO found 5.15.2 but need: '== 6.2.2' (/usr/bin/uic)
Program /usr/lib/qt6/bin/rcc found: NO
Program rcc-qt6 found: NO
Program rcc found: NO found 5.15.2 but need: '== 6.2.2' (/usr/bin/rcc)
Program /usr/lib/qt6/bin/lrelease found: YES 6.2.2 (/usr/lib/qt6/bin/lrelease)

meson.build:3:4: ERROR: RCC sources specified and couldn't find rcc-qt6, please check your qt6 installation
```

**Expected behavior**
meson should find the `rcc` binary which is located at `/usr/lib/qt6/rcc`.

**system parameters**
* native build
* Arch Linux
* Python 3.10.1
* meson 0.60.3
* ninja 1.10.2

Proof-of-concept patch to add support for QT_HOST_LIBEXECS:
```
From 3dcdd69637ffba416e2e99fd812772b028dd0b83 Mon Sep 17 00:00:00 2001
From: John Lindgren <john@jlindgren.net>
Date: Sat, 8 Jan 2022 16:21:27 -0500
Subject: [PATCH] Add support for Qt 6.1+

Qt 6.1 moved the location of some binaries from QT_HOST_BINS to
QT_HOST_LIBEXECS as noted in the changelog:

c515ee178f Move build tools to libexec instead of the bin dir
- Tools that are called by the build system and are unlikely to be
called by the user are now installed to the libexec directory.

[...]

PR #9792, opened 2022-01-08
Add support for Qt 6.1+

jlindgren90:
Qt 6.1 moved the location of some binaries from QT_HOST_BINS to
QT_HOST_LIBEXECS as noted in the changelog:

c515ee178f Move build tools to libexec instead of the bin dir
- Tools that are called by the build system and are unlikely to be
called by the user are now installed to the libexec directory.

https://code.qt.io/cgit/qt/qtreleasenotes.git/tree/qt/6.1.0/release-note.txt

To compensate, meson now needs to look for moc, rcc, uic, etc. in
QT_HOST_LIBEXECS as well as QT_HOST_BINS.

Fixes: https://github.com/mesonbuild/meson/issues/9791

### DIFF
--- a/mesonbuild/dependencies/qt.py
+++ b/mesonbuild/dependencies/qt.py
@@ -69,6 +69,12 @@ def get_qmake_host_bins(qvars: T.Dict[str, str]) -> str:
     return qvars['QT_INSTALL_BINS']
 
 
+def get_qmake_host_libexecs(qvars: T.Dict[str, str]) -> str:
+    if 'QT_HOST_LIBEXECS' in qvars:
+        return qvars['QT_HOST_LIBEXECS']
+    return qvars['QT_INSTALL_LIBEXECS']
+
+
 def _get_modules_lib_suffix(version: str, info: 'MachineInfo', is_debug: bool) -> str:
     """Get the module suffix based on platform and debug type."""
     suffix = ''
@@ -277,6 +283,7 @@ class QmakeQtDependency(_QtBase, ConfigToolDependency, metaclass=abc.ABCMeta):
         libdir = qvars['QT_INSTALL_LIBS']
         # Used by qt.compilers_detect()
         self.bindir = get_qmake_host_bins(qvars)
+        self.libexecdir = get_qmake_host_libexecs(qvars)
 
         # Use the buildtype by default, but look at the b_vscrt option if the
         # compiler supports it.
@@ -353,6 +360,7 @@ class QmakeQtDependency(_QtBase, ConfigToolDependency, metaclass=abc.ABCMeta):
             self.is_found = True
             # Used by self.compilers_detect()
             self.bindir = get_qmake_host_bins(qvars)
+            self.libexecdir = get_qmake_host_libexecs(qvars)
 
     def log_info(self) -> str:
         return 'qmake'
--- a/mesonbuild/modules/qt.py
+++ b/mesonbuild/modules/qt.py
@@ -128,6 +128,8 @@ class QtBaseModule(ExtensionModule):
             for b in self.tools:
                 if qt_dep.bindir:
                     yield os.path.join(qt_dep.bindir, b), b
+                if qt_dep.libexecdir:
+                    yield os.path.join(qt_dep.libexecdir, b), b
                 # prefer the <tool>-qt<version> of the tool to the plain one, as we
                 # don't know what the unsuffixed one points to without calling it.
                 yield f'{b}-qt{qt_dep.qtver}', b


nirbheek:
It would be nice to get a test for this in one of the CI jobs so we know that it's actually working.

jlindgren90:
> It would be nice to get a test for this in one of the CI jobs so we know that it's actually working.

No disagreement here, however I don't have the time to learn your test framework and write a new test. I can tell you that the patch worked for me in a real build of Audacious that failed before.

nirbheek:
Test failures are also relevant. This patch is incomplete.

eli-schwartz:
In theory `test cases/frameworks/4 qt/meson.build` accomplishes precisely this. Qt6 is installed on the Arch CI images already.

jlindgren90:
> In theory `test cases/frameworks/4 qt/meson.build` accomplishes precisely this. Qt6 is installed on the Arch CI images already.

Please try the test case included in #9791 , it was failing on an up-to-date Arch Linux system as of yesterday.

jlindgren90:
> Test failures are also relevant. This patch is incomplete.

Please feel free to use my patch as a starting point and complete it however you see fit 👍

radioactiveman:
@jlindgren90: I have tried your patch to compile Audacious on Arch Linux. But the qt6 dependency is still not found with qmake. How did you get this to work?

The following patch helps by also looking for qmake6. There is no symlink named qmake-qt6 like with qmake-qt5, at least on Arch.

```diff
diff --git a/mesonbuild/dependencies/qt.py b/mesonbuild/dependencies/qt.py
index a004688ab..34df0a0a3 100644
--- a/mesonbuild/dependencies/qt.py
+++ b/mesonbuild/dependencies/qt.py
@@ -237,7 +237,7 @@ class QmakeQtDependency(_QtBase, ConfigToolDependency, metaclass=abc.ABCMeta):
 
     def __init__(self, name: str, env: 'Environment', kwargs: T.Dict[str, T.Any]):
         _QtBase.__init__(self, name, kwargs)
-        self.tools = [f'qmake-{self.qtname}', 'qmake']
+        self.tools = [f'qmake-{self.qtname}', f'qmake{self.qtver}', 'qmake']
 
         # Add additional constraints that the Qt version is met, but preserve
         # any version requrements the user has set as well. For example, if Qt5

```

Result (extract of build/meson-log.txt):

```
Determining dependency 'Qt6Core' with pkg-config executable '/usr/bin/pkg-config'
env[PKG_CONFIG_PATH]: 
qmake binary missing from cross or native file, or env var undefined.
Trying a default qmake fallback at qmake-Qt6
Trying a default qmake fallback at qmake6
Trying a default qmake fallback at qmake
qmake found: YES (/usr/bin/qmake6) 6.2.3
Run-time dependency qt6 (modules: Core, Gui, Widgets) found: YES 6.2.3 (qmake)
```

jlindgren90:
> @jlindgren90: I have tried your patch to compile Audacious on Arch Linux. But the `qt6` dependency is still not found with `qmake`. How did you get this to work?

I believe I just did `export QMAKE=qmake6` as a workaround.  But your patch looks like a better solution :+1:

radioactiveman:
@jlindgren90: With the patch your test case from #9791 works with Qt 6, but not with Qt 5.

meson.build:
```
project('test', 'cpp')

qt5 = import('qt5')
qt5.preprocess(qresources: 'test.qrc')

#qt6 = import('qt6')
#qt6.preprocess(qresources: 'test.qrc')
```

test.qrc:
```
<RCC>
  <qresource />
</RCC>
```

Same error as in the CI: https://github.com/mesonbuild/meson/runs/4754571933?check_suite_focus=true
```
  File "/usr/lib/python3.10/site-packages/mesonbuild/modules/qt.py", line 174, in _detect_tools
    self.compilers_detect(state, qt)
  File "/usr/lib/python3.10/site-packages/mesonbuild/modules/qt.py", line 138, in compilers_detect
    for b, name in gen_bins():
  File "/usr/lib/python3.10/site-packages/mesonbuild/modules/qt.py", line 131, in gen_bins
    if qt_dep.libexecdir:
AttributeError: 'Qt5PkgConfigDependency' object has no attribute 'libexecdir'

meson.build:4:4: ERROR: Unhandled python exception
```

radioactiveman:
> In theory `test cases/frameworks/4 qt/meson.build` accomplishes precisely this. Qt6 is installed on the Arch CI images already.

@eli-schwartz: Any idea why this Qt 6 test is skipped in the CI build on Arch Linux? I assume it would fail even on master if it was actually executed.

```
run_unittests.py::LinuxlikeTests::test_qt6dependency_qmake_detection <- unittests/linuxliketests.py 
[gw1] [ 68%] SKIPPED run_unittests.py::LinuxlikeTests::test_qt6dependency_qmake_detection <- unittests/linuxliketests.py
```

eli-schwartz:
Because it searches for `qmake-qt6` and `qmake`, according to the patterns followed by qt5. Qt 6 created actual official symlinks for that, but distributors of Qt have to manually opt in to create that, by iterating over a file dumped in the build directory and running `ln -s`. Fascinating.

At least it's official now. Sadly, now it is called `qmake6`, because I guess the Qt Company wasn't very impressed with the prior art that distributors did with appending `-qtX` instead of `X`.

Arch does have a qmake6, but long after the initial Meson implementation of the test case, and once Qt made an official solution, migrated to the official names. We never adapted, the test just began to skip.

eli-schwartz:
I attempted to minimally adapt to the new name: https://github.com/eli-schwartz/meson/commit/d143d872e48794a1f1f77e062e7c417e936b22a6

In local attempts to build, I then got this fascinating error message:
```python
qmake found: YES (/usr/bin/qmake6) 6.2.3
  File "/home/eschwartz/git/meson/mesonbuild/mesonmain.py", line 149, in run
  File "/home/eschwartz/git/meson/mesonbuild/msetup.py", line 294, in run
    app.generate()
  File "/home/eschwartz/git/meson/mesonbuild/msetup.py", line 185, in generate
    self._generate(env)
  File "/home/eschwartz/git/meson/mesonbuild/msetup.py", line 229, in _generate
    intr.run()
  File "/home/eschwartz/git/meson/mesonbuild/interpreter/interpreter.py", line 2632, in run
    super().run()
  File "/home/eschwartz/git/meson/mesonbuild/interpreterbase/interpreterbase.py", line 148, in run
  File "/home/eschwartz/git/meson/mesonbuild/interpreterbase/interpreterbase.py", line 173, in evaluate_codeblock
    raise e
  File "/home/eschwartz/git/meson/mesonbuild/interpreterbase/interpreterbase.py", line 166, in evaluate_codeblock
    self.evaluate_statement(cur)
  File "/home/eschwartz/git/meson/mesonbuild/interpreterbase/interpreterbase.py", line 183, in evaluate_statement
    return self.method_call(cur)
  File "/home/eschwartz/git/meson/mesonbuild/interpreterbase/interpreterbase.py", line 482, in method_call
    res = obj.method_call(method_name, args, kwargs)
  File "/home/eschwartz/git/meson/mesonbuild/interpreter/interpreterobjects.py", line 731, in method_call
  File "/home/eschwartz/git/meson/mesonbuild/interpreterbase/decorators.py", line 667, in wrapped
  File "/home/eschwartz/git/meson/mesonbuild/interpreterbase/decorators.py", line 52, in wrapped
  File "/home/eschwartz/git/meson/mesonbuild/interpreterbase/decorators.py", line 573, in wrapper
  File "/home/eschwartz/git/meson/mesonbuild/modules/qt.py", line 271, in has_tools
    self._detect_tools(state, method, required=False)
  File "/home/eschwartz/git/meson/mesonbuild/modules/qt.py", line 172, in _detect_tools
    qt = T.cast('QtPkgConfigDependency', find_external_dependency(f'qt{self.qt_version}', state.environment, kwargs))
  File "/home/eschwartz/git/meson/mesonbuild/dependencies/detect.py", line 114, in find_external_dependency
    d = c()
  File "/home/eschwartz/git/meson/mesonbuild/dependencies/factory.py", line 110, in <lambda>
    self.classes[DependencyMethods.CONFIG_TOOL] = lambda env, kwargs: configtool_class(name, env, kwargs)
  File "/home/eschwartz/git/meson/mesonbuild/dependencies/qt.py", line 305, in __init__
    libfiles = self.clib_compiler.find_library(
AttributeError: 'NoneType' object has no attribute 'find_library'

meson.build:6:3: ERROR: Unhandled python exception

```

Which I think somehow has to be @dcbaker's fault because who else should I blame `mesonbuild.dependencies.*` on? :D

But also because it seems like #8568 is just large and confusing enough to cause the problem.

dcbaker:
Ugggg. You do have c or c++ as a found language, right? I think that’s the only case clib compiler is supposed to be none

eli-schwartz:
Uhm, actually no I didn't (this was a very small dummy `/tmp/meson.build` which I dump random one-liners I want to test into), which I guess explains quite a bit. But it's still a pretty silly way to fail, lol.

Indeed, that fails with a deeply sorcerous traceback if I try to do `dependency('any-custom-dependency-lookup')`. But pkg-config does of course work fine.

dcbaker:
Probably just worth changing all custom dependency classes to require a compiler and fail otherwise

eli-schwartz:
> @jlindgren90: With the patch your test case from #9791 works with Qt 6, but not with Qt 5.

This should somehow handle the case where QT_HOST_LIBEXECS / QT_INSTALL_LIBEXECS does not exist, which it will not for qmake-qt5, and additionally the QtPkgConfigDependency needs to handle it the same way -- because currently pkg-config is exclusive to qt5.

Probably by setting it to `None` and checking in `mesonbuild/modules/qt.py` whether `qt_dep.libexecdir` is None before attempting to use it as a path.

eli-schwartz:
Replaced by #10259

Issue #9097, opened 2021-08-11
Adding `env` to a custom_target causes unexpected side-effects (empty stdout)


rusty-snake:
**Describe the bug**

Adding `env` to a custom_target causes unexpected side-effects (stdout is empty).

**To Reproduce**

meson.build:
```meson
project('test02')

custom_target('without-env',
  command: ['echo', 'hello meson (1)'],
  output: ['nothing01'],
  build_by_default: true,
)
custom_target('env',
  command: ['echo', 'hello meson (2)'],
  output: ['nothing02'],
  build_by_default: true,
  env: {'QUARK': 'STRANGE'},
)
```

```console
$ meson setup _builddir && meson compile -C _builddir
The Meson build system
Version: 0.59.0
Source dir: /home/rusty-snake/test02
Build dir: /home/rusty-snake/test02/_builddir
Build type: native build
Project name: test02
Project version: undefined
Host machine cpu family: x86_64
Host machine cpu: x86_64
Program echo found: YES (/usr/bin/echo)
Program echo found: YES (/usr/bin/echo)
Build targets in project: 2

Found ninja-1.10.2 at /usr/bin/ninja
ninja: Entering directory `_builddir'
[1/2] Generating without-env with a custom command
hello meson (1)
[2/2] Generating env with a custom command (wrapped by meson to set env)
$ 
```

**Expected behavior**

Adding `env` doesn't have any other effects than setting a environment variable.

**system parameters**
* Is this a [cross build](https://mesonbuild.com/Cross-compilation.html) or just a plain native build (for the same computer)?
  plain native
* what operating system (e.g. MacOS Catalina, Windows 10, CentOS 8.0, Ubuntu 18.04, etc.)
  Fedora 34
* what Python version are you using e.g. 3.8.0
  3.9.6
* what `meson --version`
  0.59.0
* what `ninja --version` if it's a Ninja build
  1.10.2

xclaesse:
Hmm, now the question is which one is desired? For compiler targets I think it prints stderr but print stdout only on error?

xclaesse:
IMHO we should hide output, except when `console: true`. The doc says: `which has special properties such as not buffering stdout`.

declanqian:
hmm, just had the same issue.

After adding karg env to a custom_target which also declared `console: true`.
Command's stdout got hidden when run `meson compile` after adding `env`

declanqian:
> IMHO we should hide output, except when console: true.

@xclaesse 
[...]

PR #9794, opened 2022-01-09
ninja backend: Fix custom_target() console: kwarg when using env

declanqian:
See context from here #9097


nirbheek:
Could you please fix the commit message to be a more descriptive? Something like:

```
ninja backend: Fix custom_target() console: kwarg when using env

When a custom_target() uses an env, meson uses a wrapper 
script to run the executable. This breaks the console: kwarg 
because the wrapper script buffers the output. Fix it by setting 
the verbose flag which will not buffer output.
```

declanqian:
Right. I just git commit amend with the nice commit message you putted together. @nirbheek

PR #9798, opened 2022-01-09
subproject: fix version validation on lookup

oyvindaakre:
Fixes a bug where the subproject version was not validated
when the subproject had already been processed.

The bug would cause inconsistent build results if the subproject was
referenced more than once (diamond) with conflicting version requirements.

### DIFF
--- a/mesonbuild/interpreter/interpreter.py
+++ b/mesonbuild/interpreter/interpreter.py
@@ -829,6 +829,11 @@ external dependencies (including libraries) must go to "dependencies".''')
             subproject = self.subprojects[subp_name]
             if required and not subproject.found():
                 raise InterpreterException(f'Subproject "{subproject.subdir}" required but not found.')
+            if 'version' in kwargs:
+                pv = self.build.subprojects[subp_name]
+                wanted = kwargs['version']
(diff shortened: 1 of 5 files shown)


jpakkane:
There should be a test for this in the `failing`  subdir.

eli-schwartz:
Changes look reasonable, but the test case should probably be squashed into the first commit.

jpakkane:
```
 test cases/failing/120 subproject version conflict/meson.build:4:0: 
ERROR: Subproject B version is 100 but 1 required.
expected "test cases/failing/120 subproject version/meson.build:4:0: 
```

The latter message is missing "conflict" in the directory name.

oyvindaakre:
> Changes look reasonable, but the test case should probably be squashed into the first commit.

Upon merge or do you prefer I push a squashed commit first?

eli-schwartz:
Preferably pushed by the PR author. Github's squash thingy has unreliable behavior IMO.

oyvindaakre:
> Preferably pushed by the PR author. Github's squash thingy has unreliable behavior IMO.

OK, done

eli-schwartz:
Thanks!

PR #9763, opened 2021-12-27
fix(rpm): remove unnecessary index in loop

drwatsno:
Discovered this error while attempting to build swaync

```
Traceback (most recent call last):
  File "/usr/lib/python3.10/site-packages/mesonbuild/mesonmain.py", line 228, in run
    return options.run_func(options)
  File "/usr/lib/python3.10/site-packages/mesonbuild/msetup.py", line 281, in run
    app.generate()
  File "/usr/lib/python3.10/site-packages/mesonbuild/msetup.py", line 184, in generate
    self._generate(env)
  File "/usr/lib/python3.10/site-packages/mesonbuild/msetup.py", line 223, in _generate
    intr.run()
  File "/usr/lib/python3.10/site-packages/mesonbuild/interpreter/interpreter.py", line 2476, in run
    super().run()
  File "/usr/lib/python3.10/site-packages/mesonbuild/interpreterbase/interpreterbase.py", line 132, in run
    self.evaluate_codeblock(self.ast, start=1)
  File "/usr/lib/python3.10/site-packages/mesonbuild/interpreterbase/interpreterbase.py", line 157, in evaluate_codeblock
    raise e
  File "/usr/lib/python3.10/site-packages/mesonbuild/interpreterbase/interpreterbase.py", line 150, in evaluate_codeblock
    self.evaluate_statement(cur)
  File "/usr/lib/python3.10/site-packages/mesonbuild/interpreterbase/interpreterbase.py", line 167, in evaluate_statement
    return self.method_call(cur)
  File "/usr/lib/python3.10/site-packages/mesonbuild/interpreterbase/interpreterbase.py", line 598, in method_call
    return self._holderify(obj.method_call(method_name, args, kwargs))
  File "/usr/lib/python3.10/site-packages/mesonbuild/interpreter/interpreterobjects.py", line 766, in method_call
    ret = method(state, args, kwargs)
  File "/usr/lib/python3.10/site-packages/mesonbuild/interpreterbase/decorators.py", line 55, in wrapped
    return f(*wrapped_args, **wrapped_kwargs)
  File "/usr/lib/python3.10/site-packages/mesonbuild/modules/rpm.py", line 93, in generate_spec_template
    fn.write('BuildRequires: pkgconfig(%s)\n' % dep[0])
TypeError: not all arguments converted during string formatting
```
looks like some time ago it was broken (when .host was added to .deps)

PR #9799, opened 2022-01-10
remove the RPM module

### LINKED ISSUE #9764: RFC: delete the rpm module
Quite some time ago, @ignatenkobrain added an RPM module. This is no longer actively supported, doesn't seem to be tested, and on at least one occasion has broken for 3 years straight ***and no one even noticed***.

From taking a quick look at it, it does a lot of fragile sorcery by poking at internal coredata details, so it will inevitably break, be broken, and keep breaking if it isn't actively maintained.

I propose that no one uses it, no one should be using it, and we shouldn't commit to providing it.

For context, the purpose of this module is apparently *NOT* that your meson.build should actually use this module... The purpose of it is to temporarily add this module to meson.build, configure meson ***once*** in order to generate an RPM spec file, and then delete it from meson.build again. The developer can then manually edit the spec file (some things are missing from it), commit it to git, and maintain it purely manually from then on.

This dramatically decreases its value, because it is never actively used.

@dcbaker suggested on IRC that this would be better off as an independent script that reads the introspection data and creates a spec file based on that.

Opinions?

eli-schwartz:
It is unmaintained, broken (frequently for long periods of time) and not really required for any meson functionality. Its purpose is to be used as a one-shot tool for creating a distro package recipe, and then deleted from your meson.build files.

Due to its fragile dependency on coredata implementation details, we cannot assume it will reliably work, or continue to work, without someone who is actively willing to take responsibility for it.

Even if that were to happen, this might be better off as an external script that parses introspection data.

Closes #9764
Closes #9763


dcbaker (approved):
I'm fine with this, it's an odd module anyway.

PR #9800, opened 2022-01-10
Add a fastyaml loader for refman docs

annacrombie:
`genrefman.py -l yaml` currently takes about 80 seconds on my machine.  This is almost entirely due to the usage of `strictyaml`, which [does not consider speed to be a priority](https://hitchdev.com/strictyaml/why/speed-not-a-priority/).  This PR adds a `fastyaml` loader which uses the `CLoader` from `pyyaml`.  On my machine this reduces the runtime to 0.95 seconds.  I'm not entirely sure if this is of interest to meson developers, since you can already use `-l pickle` for a similar speedup (after you do the initial load), but I thought I would put it here just in case :).

### DIFF
--- a/docs/refman/loaderyaml.py
+++ b/docs/refman/loaderyaml.py
@@ -295,9 +295,9 @@ class LoaderYAML(LoaderBase):
         return [module, *objs]
 
     def load_impl(self) -> ReferenceManual:
-        mlog.log('Loading YAML refererence manual')
+        mlog.log('Loading YAML reference manual')
         with mlog.nested():
-            return ReferenceManual(
+            manual = ReferenceManual(
                 functions=[self._load_function(x) for x in self.func_dir.iterdir()],
                 objects=mesonlib.listify([
                     [self._load_object(ObjectType.ELEMENTARY, x) for x in self.elem_dir.iterdir()],
@@ -306,3 +306,8 @@ class LoaderYAML(LoaderBase):
                     [self._load_module(x) for x in self.modules_dir.iterdir()]
                 ], flatten=True)
             )
+
+            if not self.strict:
+                mlog.warning('YAML reference manual loaded using the best-effort fastyaml loader.  Results are not guaranteed to be stable or correct.')
+
+            return manual
--- a/docs/refman/loaderyaml.py
+++ b/docs/refman/loaderyaml.py
@@ -28,71 +28,130 @@ from .model import (
 from mesonbuild import mlog
 from mesonbuild import mesonlib
 
-from strictyaml import Map, MapPattern, Optional, Str, Seq, Int, Bool, load, EmptyList, OrValidator
 from pathlib import Path
 import typing as T
 
-d_named_object = {
-    'name': Str(),
-    'description': Str(),
-}
+class Template:
+    d_feature_check: T.Dict[str, T.Any] = {}
+    s_posarg: T.Dict[str, T.Any] = {}
+    s_varargs: T.Dict[str, T.Any] = {}
+    s_kwarg: T.Dict[str, T.Any] = {}
+    s_function: T.Dict[str, T.Any] = {}
+    s_object: T.Dict[str, T.Any] = {}
 
-d_feture_check = {
-    Optional('since', default=''): Str(),
-    Optional('deprecated', default=''): Str(),
-}
+class StrictTemplate(Template):
+    def __init__(self) -> None:
+        from strictyaml import Map, MapPattern, Optional, Str, Seq, Int, Bool, EmptyList, OrValidator
 
-s_posarg = Map({
-    **d_feture_check,
-    'description': Str(),
-    'type': Str(),
-    Optional('default', default=''): Str(),
-})
+        d_named_object = {
+            'name': Str(),
+            'description': Str(),
+        }
 
-s_varargs = Map({
-    **d_named_object, **d_feture_check,
-    'type': Str(),
-    Optional('min_varargs', default=-1): Int(),
-    Optional('max_varargs', default=-1): Int(),
-})
+        d_feture_check = {
+            Optional('since', default=''): Str(),
+            Optional('deprecated', default=''): Str(),
+        }
 
-s_kwarg = Map({
-    **d_feture_check,
-    'type': Str(),
-    'description': Str(),
-    Optional('required', default=False): Bool(),
-    Optional('default', default=''): Str(),
-})
+        self.s_posarg = Map({
+            **d_feture_check,
+            'description': Str(),
+            'type': Str(),
+            Optional('default', default=''): Str(),
+        })
 
-s_function = Map({
-    **d_named_object, **d_feture_check,
-    'returns': Str(),
-    Optional('notes', default=[]): OrValidator(Seq(Str()), EmptyList()),
-    Optional('warnings', default=[]): OrValidator(Seq(Str()), EmptyList()),
-    Optional('example', default=''): Str(),
-    Optional('posargs'): MapPattern(Str(), s_posarg),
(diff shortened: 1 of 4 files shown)


eli-schwartz:
Is this guaranteed to interpret the file the same way strictyaml does? i.e. none of the YAML spec's insane built-in types?

annacrombie:
diffing the `yaml` generated man page with the `fastyaml` one, it appears that booleans are getting mangled slightly in both versions.  with `yaml` we get `True` and `False`, and with `fastyaml` we get `true`, and `false` gets dropped entirely :(.  I guess I still have some work to do.

Other than that, everything is identical.  I can't offer any guarantees going forward though, and I definitely understand the importance of strictyaml, since yaml is such a ridiculous format.  I just can't bear to have the generation of 1 man page take 6x longer than the entire setup+build of muon.

docs/refman/loaderyaml.py
```diff
@@ -28,80 +28,169 @@
...
+            from yaml import CLoader
+            def loader(file: str, template: T.Union[T.Dict, strictyaml.validators.Validator], label: str) -> T.Dict:
+                return {**template, **yaml.load(file, Loader=CLoader)}
+
+            self._load = loader
+            template = FastTemplate()
+
+        self.template = template
+
+    def _fix_default(self, v: T.Dict) -> None:
```
mensinda, line 185: I am not sure we should even attempt to fix this. For instance, this wouldn't catch `default: ON`, `default: NO`, `default: TRUE`, etc.

mensinda (commented):
I decided to go with strictyaml specifically because it applies the schema while parsing and thus avoids all the conversion issues with YAMLs built in types. Because of these differences I don't think that an attempt to have both parsers be 100% compatible is a good idea. If we attempt this and have a CI for it, we risk having to fix the fastyaml loader for every doc change.

That being said, I think the fastyaml loader has a place for development when the end result does not need to be 100% correct.

annacrombie:
I've fixed it so that both generated man pages are identical by hacking around yaml's type conversions for the default: keyword.  I sort of agree with @mensinda that having the `fastyaml` loader tested in CI might get annoying.

docs/refman/loaderyaml.py
annacrombie: This fix is needed to make the generated man pages from both yaml loaders identical.  It is the only case where yaml type coercion was causing issues, and mostly that was because I used `if arg.default:` to determine whether an argument had a default, which caused problems for `False` and `0`.

mensinda: My point is that we shouldn't even attempt to achieve parity between the two loaders. Instead clearly state that only the `strictyaml` loader is meant for production and the `fastyaml` loader is for development only.

If we try to achieve parity between the two loaders with this fix we are implicitly saying that both can be used and also implicitly commit to fixing issues as they come up (which will be hard and result in confusion if we don't have CI to test it, which again is annoying).

IMHO this is the wrong approach here. I would rather provide both loaders and clearly state that one is for production and the other one can be used for development since it is faster but does **not** always produce the correct results.

mensinda: In that case, how about improving the performance by throwing more threads / [processes](https://docs.python.org/3.8/library/multiprocessing.html) at the Problem?

Of course this will never be as fast as `fastyaml`, but should still provide a major speed improvement.

> and I am perfectly happy to maintain it in a fork

Let's hope it doesn't come to that. We should be able to come to a compromise here.

annacrombie: Sorry, I didn't mean to suggest a full-on fork.  Rather what I meant was to maintain a small patch that replaced the yaml loader only.  I don't think it would be necessarily a bad thing, just more work for me.  And, if I ever get tired of it, I can always just go back to the `strictyaml`

As far as trying to use threading to improve strictyaml, I'm not sure I would be willing to go that far.

mensinda:
I had some time to think and IMHO, this is fine to include in Meson (also with the fix for default values) **if** there is a warning printed that the `fastyaml` frontend is only a best-effort implementation and the Meson project is not obligated to fix bugs (basically something similar to [this](https://mesonbuild.com/Mixing-build-systems.html#what-does-this-mean-for-support-and-compatibility)).

Additionally, this should also be documented [here](https://mesonbuild.com/Yaml-RefMan.html).

annacrombie:
> this is fine to include in Meson (also with the fix for default values) **if** there is a warning printed that the `fastyaml` frontend is only a best-effort implementation and the Meson project is not obligated to fix bugs

I think this sounds reasonable.  Is it sufficient to include the warning in the `--help` + docs, or would you like a banner message in the output of regular invocations?  I think describing the different loaders briefly in `--help` could be useful anyways.

mensinda:
> Is it sufficient to include the warning in the --help + docs, or would you like a banner message in the output of regular invocations?

I wasn't thinking about a banner message, but a single line stating that the significantly slower `strictyaml` loader is recommended since it is the reference implementation + a link to the docs. Adding it to the help message would also work, but I think I would still prefer always printing a small notice (ideally after everything is parsed, so the message is not missed at the start).

> I think describing the different loaders briefly in --help could be useful anyways.

Same here. If you manage to get this working (without replacing `argparse`) then that would be great.

mensinda (approved):
LGTM once mypy is happy.

mensinda:
From the looks of it, adding `types-PyYAML` [here](https://github.com/mesonbuild/meson/blob/master/.github/workflows/lint.yml#L45) should do the trick (hopefully).

docs/refman/loaderyaml.py
```diff
@@ -306,3 +306,8 @@ def load_impl(self) -> ReferenceManual:
                     [self._load_module(x) for x in self.modules_dir.iterdir()]
                 ], flatten=True)
             )
+
+            if not self.strict:
+                mlog.warning('YAML reference manual loaded using the best-effort fastyaml loader.  Results are not guaranteed to be stable or correct.')
+
+            return manual
```
eli-schwartz, line 313: Hmm, why isn't the mlog.warning just moved up a drop?

eli-schwartz: Oh... maybe this is actually set by the `ReferenceManual()` invocation?

annacrombie: Are you saying the `if not self.strict` should be unnecessary?

eli-schwartz: No, I just don't understand why you assign `manual`, *then* do `if not self.strict: mlog.warning`, then return `manual`.

Just do `if not self.strict: mlog.warning` and then `return ReferenceManual(...)`.

eli-schwartz: Ah, I see. Makes sense.

eli-schwartz:
The website is suspiciously incredibly broken now. 10k lines of documentation have disappeared as soon as the website was rebuilt by the github action running on this PR's merge.

docs/refman/loaderyaml.py
```diff
@@ -174,7 +273,13 @@ def as_methods(mlist: T.List[Function]) -> T.List[Method]:
 
         methods = raw.pop('methods', [])
         obj = Object(methods=[], obj_type=obj_type, **raw)
-        obj.methods = as_methods([self._process_function_base(x, obj) for x in methods])
+
+        methods = []
+        for x in methods:
```
eli-schwartz, line 278: Ummm.

This deleted `methods`, replaced it with an empty list, and then iterated over the empty list, mutated values, and appended to the thing it was iterating over.

eli-schwartz: Fixed via 8d648532295d131898b392032d13cd1b13e639af.

annacrombie: :cry:

PR #9802, opened 2022-01-10
compilers: push the compiler id to a class variable

dcbaker:
It really is a per class value, and shouldn't be set per instance. It
also allows us to get rid of useless constructors, including those
breaking mypy

### DIFF
--- a/mesonbuild/compilers/cs.py
+++ b/mesonbuild/compilers/cs.py
@@ -40,9 +40,8 @@ class CsCompiler(BasicLinkerIsCompilerMixin, Compiler):
     language = 'cs'
 
     def __init__(self, exelist: T.List[str], version: str, for_machine: MachineChoice,
-                 info: 'MachineInfo', comp_id: str, runner: T.Optional[str] = None):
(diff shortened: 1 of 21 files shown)


mesonbuild/compilers/cs.py
```diff
@@ -40,9 +40,8 @@ class CsCompiler(BasicLinkerIsCompilerMixin, Compiler):
     language = 'cs'
 
     def __init__(self, exelist: T.List[str], version: str, for_machine: MachineChoice,
-                 info: 'MachineInfo', comp_id: str, runner: T.Optional[str] = None):
+                 info: 'MachineInfo', runner: T.Optional[str] = None):
         super().__init__(exelist, version, for_machine, info)
-        self.id = comp_id
```
eli-schwartz, line 45: Lol, why were we passing the subclass id around as a parameter to `__init__`. Even worse than the garden variety use of `self.id = '...'` in extraneous init functions. :D

eli-schwartz (approved):
This makes sense, and looks like the same basic concept as the reorganization I did in 253ff71e6b801786290b6ecb23aea61dcfeebeca.

Issue #9795, opened 2022-01-09
rust bindgen code examples are outdated on the website


declanqian:
rust bindgen code examples are outdated on the [website](https://mesonbuild.com/Rust-module.html#bindgen-input-string-buildtarget-string-buildtarget-output-string-include_directories-include_directories-c_args-string-args-string)

Shouldn't be consistent with code from the [test](https://github.com/mesonbuild/meson/blob/2c079d855ed87488bdcc6c5c06f59abdb9b85b6c/test%20cases/rust/12%20bindgen/meson.build)?

```
input: 'myheader.h',
output: 'generated.rs',
```
```
input:  [h1, h2],  # h1 includes h2,
output: 'out.rs',
```

PR #9803, opened 2022-01-10
Fix rust module bindgen docs

dcbaker:
Fixes: #9795

### DIFF
--- a/docs/markdown/Rust-module.md
+++ b/docs/markdown/Rust-module.md
@@ -34,7 +34,7 @@ that automatically.
 Additional, test only dependencies may be passed via the dependencies
 argument.
 
-### bindgen(*, input: string | BuildTarget | []string | []BuildTarget, output: string, include_directories: []include_directories, c_args: []string, args: []string)
+### bindgen(*, input: string | BuildTarget | [](string | BuildTarget), output: string, include_directories: []include_directories, c_args: []string, args: []string)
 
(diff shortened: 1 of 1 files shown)


docs/markdown/Rust-module.md
```diff
@@ -34,7 +34,7 @@ that automatically.
 Additional, test only dependencies may be passed via the dependencies
 argument.
 
-### bindgen(*, input: string | BuildTarget | []string | []BuildTarget, output: string, include_directories: []include_directories, c_args: []string, args: []string)
+### bindgen(*, input: string | BuildTarget | [](string | BuildTarget), output: string, include_directories: []include_directories, c_args: []string, args: []string)
```
eli-schwartz, line 37: This is not actually solving any kind of confusion, the href anchors and section headers are the wrong place to use ad-hoc, complex syntax.

Still, this is hardly worse than before.

And the other two changes seem correct.

eli-schwartz (approved):

docs/markdown/Rust-module.md
dcbaker: This isn't really ad-hoc, it's typescript syntax, and I was using this long before we added the python style syntax of the yaml.  Like many things in meson this has all grown rather organically. (I really prefer this to the `List[str]` syntax anyway, since Meson only has arrays and dicts anyway, and it's less wordy.

PR #9806, opened 2022-01-10
gnome: fix typo in vapigen option metadatadir

paper42:
This typo was introduced in 9ef36fa80bf4483bcd837ea3985a51ab533c7972 - #9594 and can cause a build failure, because --metadatdir is not a valid vapigen option. This can be seen for example in the [fontmanager](https://github.com/FontManager/font-manager) or [AppStream](https://github.com/ximion/appstream) builds, so this fix should ideally be part of 0.60.1.

@eli-schwartz @dcbaker


eli-schwartz (approved):
Ooof, thanks for the catch.

dcbaker (approved):
oops. Clearly our testing of the gnome module leaves something to be desired...

dcbaker:
that mypy error will rebase out, it's fixed in the main branch and not related to this PR.

PR #9816, opened 2022-01-12
Fix default install tag for shared lib symlinks

xclaesse:
Versioned shared libraries should have .so file in devel, .so.1 and
.so.1.2.3 in runtime.

### DIFF
--- a/mesonbuild/minstall.py
+++ b/mesonbuild/minstall.py
@@ -428,11 +428,11 @@ class Installer:
         append_to_log(self.lf, to_file)
         return True
 
-    def do_symlink(self, target: str, link: str, full_dst_dir: str) -> bool:
+    def do_symlink(self, target: str, link: str, full_dst_dir: str, allow_missing: bool) -> bool:
         abs_target = target
         if not os.path.isabs(target):
             abs_target = os.path.join(full_dst_dir, target)
-        if not os.path.exists(abs_target):
+        if not os.path.exists(abs_target) and not allow_missing:
             raise MesonException(f'Tried to install symlink to missing file {abs_target}')
         if os.path.exists(link):
             if not os.path.islink(link):
@@ -592,7 +592,7 @@ class Installer:
             full_dst_dir = get_destdir_path(destdir, fullprefix, s.install_path)
             full_link_name = get_destdir_path(destdir, fullprefix, s.name)
             dm.makedirs(full_dst_dir, exist_ok=True)
-            if self.do_symlink(s.target, full_link_name, full_dst_dir):
+            if self.do_symlink(s.target, full_link_name, full_dst_dir, s.allow_missing):
                 self.did_install_something = True
 
     def install_man(self, d: InstallData, dm: DirMaker, destdir: str, fullprefix: str) -> None:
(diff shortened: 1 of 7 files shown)


mesonbuild/minstall.py
```diff
@@ -463,8 +463,8 @@ def do_symlink(self, target: str, link: str, full_dst_dir: str) -> bool:
         abs_target = target
         if not os.path.isabs(target):
             abs_target = os.path.join(full_dst_dir, target)
-        if not os.path.exists(abs_target):
-            raise RuntimeError(f'Tried to install symlink to missing file {abs_target}')
+        #if not os.path.exists(abs_target):
+        #    raise RuntimeError(f'Tried to install symlink to missing file {abs_target}')
```
xclaesse, line 467: This is a TODO: when installing only `devel` the target does not exist, not sure we should always allow installing symlink to a target that does not exist??

mesonbuild/mintro.py
```diff
@@ -107,9 +107,6 @@ def list_installed(installdata: backends.InstallData) -> T.Dict[str, str]:
         for t in installdata.targets:
             res[os.path.join(installdata.build_dir, t.fname)] = \
                 os.path.join(installdata.prefix, t.outdir, os.path.basename(t.fname))
-            for alias in t.aliases.keys():
-                res[os.path.join(installdata.build_dir, alias)] = \
-                    os.path.join(installdata.prefix, t.outdir, os.path.basename(alias))
```
xclaesse, line 112: TODO: aliases are now parts of `installdata.symlinks` but those are not part of introspection data.

mesonbuild/minstall.py
eli-schwartz, line 467: Huh, why does this need to be commented out? Don't we install destinations before symlinks already?

mesonbuild/mintro.py
```diff
@@ -217,7 +214,6 @@ def list_targets(builddata: build.Build, installdata: backends.InstallData, back
     install_lookuptable = {}
     for i in installdata.targets:
         out = [os.path.join(installdata.prefix, i.outdir, os.path.basename(i.fname))]
-        out += [os.path.join(installdata.prefix, i.outdir, os.path.basename(x)) for x in i.aliases]
```
xclaesse, line 220: TODO: should symlinks be part of targets?

mesonbuild/minstall.py
xclaesse: Opened a TODO review comment myself about this, see above.

eli-schwartz: Ah, good point!

xclaesse: Allowed it only for library symlinks for now. Better be conservative.

mesonbuild/mintro.py
xclaesse: Side effect of this PR, installdata.symlinks are now part of introspection data.

mesonbuild/mintro.py
xclaesse: yes, now they are.

xclaesse:
@eli-schwartz ok, finally got the CI pass here, ready for review \o/

PR #9817, opened 2022-01-13
gome.gdbus_codegen: fix annotations argument for multiple annotations, harder

eli-schwartz:
The original attempted fix only allowed configuration to succeed, but not building. It was modeled based on the gdbus-codegen documentation, which states:

```
  --annotate WHAT KEY VALUE WHAT KEY VALUE WHAT KEY VALUE
                        Add annotation (may be used several times)
```

which clearly indicates that gdbus-codegen accepts an --annotate flag that is followed by a multiple of 3 values, despite this not actually working.

The manpage actually contradicts the --help text:

```
   --annotate ELEMENT KEY VALUE
       Used to inject D-Bus annotations into the given XML files. []
```

... and gives examples that use multiple --annotate flags each with 3 arguments.

To better understand what meson is supposed to do here, we should look at ef52e60936665c982cd17a4a17c2045b445d8e6d, which ported to typed_kwargs. There is actually a big chunk of code to handle annotations that got completely dropped, leading with a comment (that did not get dropped): "they are a list of lists of strings..."

Reimplement this logic inside a validator/converter for the annotations kwarg container:

- do not listify, we don't accept `annotations: ''` and listify is  supposed to be for when either x or list[x] is valid

- go back to checking for a list of exactly 3 values

- allow a list of the aforementioned, in the traditionally expected form:
  ```
  [
    ['foo1', 'foo2', 'foo3'],
    ['bar1', 'bar2', 'bar3'],
  ]
  ```

- pass one --annotate flag per 3-value-list

And add some better error reporting for the cause of errors when processing lists of lists.

### DIFF
--- a/mesonbuild/modules/gnome.py
+++ b/mesonbuild/modules/gnome.py
@@ -134,7 +134,7 @@ if T.TYPE_CHECKING:
         namespace: T.Optional[str]
         object_manager: bool
         build_by_default: bool
-        annotations: T.List[str]
+        annotations: T.List[T.List[str]]
         install_header: bool
         install_dir: T.Optional[str]
         docbook: T.Optional[str]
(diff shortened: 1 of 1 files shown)


mesonbuild/modules/gnome.py
```diff
@@ -222,6 +222,27 @@ class MkEnums(_MkEnumsCommon):
     KwargInfo('symbol_prefix', (str, NoneType)),
 ]
```
dcbaker, line 223: Dang github won't let me put the comment in the right place, but on line 127, you need to replace `annotations: List[str]` with `annotations: List[List[str]]`

dcbaker (changes requested):

mesonbuild/modules/gnome.py
eli-schwartz: Oops, fixed.

dcbaker (approved):

Issue #9822, opened 2022-01-13
Meson complaining about use of -fsanitize-recover=alignment


tristan957:
**Describe the bug**
When using `-fsanitize-recover=alignment`, Meson thinks I am using the `-fsanitize` argument I supposed, so it gives the following warning:

```
meson.build:64: WARNING: Consider using the built-in option for sanitizers instead of using "-fsanitize-recover=alignment".
```

**To Reproduce**
Add the argument to your project.

**Expected behavior**
Not warn, since this flag is about adjusting behavior of the sanitizer instead of enabling it.

**system parameters**
* Is this a [cross build](https://mesonbuild.com/Cross-compilation.html) or just a plain native build (for the same computer)? No
* what operating system (e.g. MacOS Catalina, Windows 10, CentOS 8.0, Ubuntu 18.04, etc.) Fedora 35
* what Python version are you using e.g. 3.8.0 3.10.1
* what `meson --version` 0.59,.5
* what `ninja --version` if it's a Ninja build 1.10.2

Something tells me the fix is probably just fixing the `-fsanitize` detection logic.

kcgen:
Still hitting this in 1.1.0

tristan957:
See the linked PR. I am sure it'll get picked up again in the future.

kcgen:
> See the linked PR. I am sure it'll get picked up again in the future.

Indeed! Thanks for the heads up,  @tristan957.

Issue #7761, opened 2020-09-19
Missing support for CFLAGS="-fsanitize=fuzzer"? 


lmcjoma:
I am trying to build a system with meson where fuzzing is activated.
The code I am trying to build is here:
https://gitlab.freedesktop.org/elmarco/libslirp/-/commit/9fba8af484ec6bc10b22e3f49d9e34d95c28b086

However, when I try to build it according to the second method in the commit log I get the following:
--------------------------------------------------------------------------------------------------------------------------------
FLAGS="-fsanitize=fuzzer" CC=clang CXX=clang++ meson build-clang -Db_lundef=false 
The Meson build system
Version: 0.55.3
Source dir: /home/jmaloy/fuzzing/lureau/libslirp
Build dir: /home/jmaloy/fuzzing/lureau/libslirp/build-clang
Build type: native build
Project name: slirp
Project version: 4.0.0
Using 'CC' from environment with value: 'clang'
Using 'CFLAGS' from environment with value: '-fsanitize=fuzzer'

meson.build:1:0: ERROR: Compiler clang can not compile programs.

A full log file can be found at /home/jmaloy/fuzzing/lureau/libslirp/build-clang/meson-logs/meson-log.txt
[jmaloy@f31 libslirp]$ 

The log referred to above looks as follows:
----------------------------------------------------------------
Build started at 2020-09-19T15:11:11.217041
Main binary: /usr/bin/python3
Build Options: -Db_lundef=false
Python system: Linux
The Meson build system
Version: 0.55.3
Source dir: /home/jmaloy/fuzzing/lureau/libslirp
Build dir: /home/jmaloy/fuzzing/lureau/libslirp/build-clang
Build type: native build
None of 'PKG_CONFIG_PATH' are defined in the environment, not changing global flags.
None of 'PKG_CONFIG_PATH' are defined in the environment, not changing global flags.
Project name: slirp
Project version: 4.0.0
Using 'CC' from environment with value: 'clang'
Using 'CFLAGS' from environment with value: '-fsanitize=fuzzer'
None of 'LDFLAGS' are defined in the environment, not changing global flags.
None of 'CPPFLAGS' are defined in the environment, not changing global flags.
None of 'CC_LD' are defined in the environment, not changing global flags.
Sanity testing C compiler: clang
Is cross compiler: False.
None of 'CC_LD' are defined in the environment, not changing global flags.
[...]

Issue #7192, opened 2020-05-23
Misleading warning about -fsanitize-blacklist


christianrauch:
**Describe the bug**
Setting `-fsanitize-blacklist` withi meson gives a misleading warning:
```WARNING: Consider using the built-in option for sanitizers instead of using "-fsanitize-blacklist=asan_blacklist.txt"```
that suggests that meson supports blacklisting as a built-in function. However, https://mesonbuild.com/Builtin-options.html does not mention any sanitizer blacklisting functionality.

**To Reproduce**
Enable sanitizer blacklisting in meson:
```python
cc = meson.get_compiler('c')
if cc.get_id() == 'clang'
  add_global_arguments('-fsanitize-blacklist=asan_blacklist.txt', language : 'c')
endif
```

**Expected behavior**
Meson should not suggest that `sanitize-blacklist` is a built-in option or print the built-in option to use in place of `sanitize-blacklist`.

**system parameters**
* native build
* Ubuntu 20.04
* Python 3.8.2
* meson 0.53.2
* ninja 1.10.0

PR #9825, opened 2022-01-13
Sanitizer enhancements

dcbaker:
This little series makes a serious overhaul of the way b_sanitize works.

First, some cleanup around spurious warnings for sanitizer options.

Then, instead of being a combo option, it becomes a free form array, and we do an internal compiler.has_arguments check to see if the compiler actually supports the arguments that we've asked for. Using an array, rather than a string, has some nice properties for writing meson build files. With a string, it's tempting to write things like:
```meson
if get_option('b_sanitize') == 'address,undefined'
  ...
endif
```

But this isn't actually safe anymore, since it well could be `'undefined,address'`. It also heads off the possibility of a compiler that doesn't use a `,` as a separator, say something that takes `-Osanitize=address -Osanitize=undefined`.

But, in order to preserve backwards compatibility, the `get_option('b_sanitize')` has been specialized to return a sorted string value if the meson compatibility version is < 0.62 (< 0.61.99 at the moment, because it's impossible to test otherwise) and a list if it is >= 0.62. This means that if a user with 0.62 sets `-Db_sanitize=['undefined', 'address']` then `get_option()` will still return `address,undefined`. This doesn't fix cases of adding new, previously unsupported sanitizers into the mix, `['undefined', 'address', 'fuzzer']` will still trip things up.

Fixes: #9822 
Fixes: #7192
Fixes: #8283
Fixes: #7761
Fixes: #5154
Fixes: #1582


dcbaker:
@bonzini, You also took a stab at this in the past, so I thought you might be interested

eli-schwartz:
> specialized to return a sorted string value if the meson compatibility version is < 0.62 (< 0.61.99 at the moment, because it's impossible to test otherwise) and a list if it is >= 0.62.

This makes me so incredibly nervous, it's basically cmake policies.

eli-schwartz:
Both string and array *do* have .contains().

Only the latter has the `in` operator.

Splitting it because you expect it to be a string and want to convert it to an array would indeed break and cause existing meson.build files to break strangely.

...

You are correct that none of the options really seem to inspire tremendous enthusiasm.

Shower thought: could we somehow warn any time get_option('b_sanitize') is checked with `==` instead of `.contains()`? `WARNING: b_sanitize will change to an array when minimum meson_version >=0.62. Use .contains() to ensure it works either way.` Probably not...

dcbaker:
That would actually be fairly trivial to do in meson++ with it's IR, lol. I think that would be quite challenging for an interpreter though, there isn't a clean, easy way to declare that you're comparing the output of `get_option('b_sanitize')`. maybe @mensinda knows how to do it?

eli-schwartz:
Actually, we already do this for `class MesonVersionStringHolder(StringHolder)`, to override its `str.version_compare()` method (so that those suppress FeatureNew warnings in conditional blocks).

dcbaker:
for MesonVersionStringHolder that's a small amount of technical debt for a huge payoff. I wonder if it's worth the additional technical debt. Maybe it would be easier to add `in` to strings?

tristan957:
I think I am going to be targeting Meson 0.62 as the next bump for my project. Would be sweet to get this in for that release assuming it is feasible.

bonzini:
There might be another way to do it, which is to name the new option b_sanitizers and return the string `get_option('b_sanitize')`. Support for `-Db_sanitize` can be dropped.

bonzini:
> Maybe it would be easier to add `in` to strings?

I prefer not to have it. Looking for substrings tends to be an antipattern. I'm worried that people would write `in` instead of the more correct `startswith`/`endswith` just because it's more compact.

dcbaker:
I've rebased this and done the 0.61 -> 0.62 and 0.62 -> 0.63 dance. I think there's probably still a few CI failures to sort out.

But, let's talk about the options we have not breaking everything.

1) continue to do what I did, of returning a sorted, comma joined string for < 0.63, an array otherwise
2) return an array, and point out that `.contains()` will work in both cases, accept some old projects will break
3) implement a special container type that does the coercion of `==`, `in`, etc, and deprecate those comparisons
4) replace b_sanitize with b_sanitize2
5 some other option I'm not thinking of

tristan957:
I only like options 1 and 2. Option 2 is probably my favorite one.

bonzini:
My favorite is 4/2/1, but I am worried that 2 is not a good idea. For option 4, I would not bother with supporting -Db_sanitize, unless it turns out to be easy.

dcbaker:
My thought for `b_sanitize2` (or `b_sanitize_array`?) is that it would return an array, while `b_sanitize` would continue to return a string which would always match the values we guarantee, so if you use a new value like "return" then it would just be ignored by `b_sanitize`, since you wouldn't be able to pass that anyway.

bonzini:
My idea was:

*Name it b_sanitizers which is better anyway than b_sanitize
*Do not allow setting b_sanitize anymore on the command line to simplify the implementation
*Make get_option('b_sanitize') return the join of the elements of get_option('b_sanitizers')

dcbaker:
yup, that makes sense to me. I'll do that.

eli-schwartz:
Oh by the way, since commit e19e9ce6f196f7c127a2668b5df0ada1d50806df the return value of `get_option('b_sanitize')` is now a `P_OBJ.OptionString` (subclass of `str`). So, we can actually check that now.

eli-schwartz:
Resubmitted as #12648 and partially merged. Resubmitted again as #14302 and the final bits were merged.

All done.

PR #9826, opened 2022-01-14
Fix system include arguments for clang-cl

Jairard:
Currently the include directories with `is_system: true` produce `-I` arguments instead of `/clang:-isystem` ones.
This fixes the issue and adds a test case.


test cases/common/250 system include dir/meson.build
```diff
@@ -0,0 +1,11 @@
+project('system_include_dir', 'cpp',
+    version : '0.1',
+    default_options : 'werror=true',
+)
+
+lib_include_directories = include_directories('lib', is_system: true)
+
+if meson.get_compiler('cpp').get_id() != 'msvc'
+    add_project_arguments('-Wsign-conversion', language: 'cpp')
+endif
```
dcbaker, line 10: This test is really fragile, since it assumes every compiler that isn't msvc has -Wsign-conversion (it doesn't). It would probably be better to check this as:
```meson
if not ['gcc', 'clang',' 'clang-cl'].contains(meson.get_compiler('cpp').get_id())
  error('MESON_SKIP_TEST: compiler @0@ either doesn\'t support is_system includes or needs to have support for this test aded'.format(meson.get_compiler('cpp').get_id())
endif
```

Which means that no compilers get her unexpectedly, an dyou don't need the fragile !msvc test (which will fail on lots of compilers because for example, icc needs a different argument).

dcbaker (changes requested):
The code change itself looks good, but the test looks fragile, and can be improved.

test cases/common/250 system include dir/meson.build
Jairard: Is this enough to statisfy CI or should I also configure the test to be skipped when compiler does not have system includes ?

dcbaker (approved):
LGTM, thanks!

Jairard:
Great :)
Do you have any idea when 0.61.1 will be released ?

nirbheek:
> Do you have any idea when 0.61.1 will be released ?

Within a week.

Jairard:
>> Do you have any idea when 0.61.1 will be released ?
>
> Within a week.

Wonderful, thanks ! 😃

PR #9828, opened 2022-01-14
gnome.genmarshal: restore the ability to pass sources as Files objects

### LINKED ISSUE #9827: gnome.marshal() accept files in addition to strings for the sources
Hello!
Since meson > 0.61 the method [gnome.marshal()](https://mesonbuild.com/Gnome-module.html#gnomegenmarshal) doesn't accept anymore a list of files for the parameter `sources`. According to the documentation this is correct:

```
sources []str required: List of string sources to consume
sources: the list of sources to use as inputs
```

I assume a more strict and therefore better type checking. The maintainers of gnome-terminal [prefer](https://gitlab.gnome.org/GNOME/gnome-terminal/-/issues/7886#note_1355184) to keep passing files. Is it possible to extend the API?

eli-schwartz:
It used to support:
- a single string
- an array of anything

And as long as CustomTarget supported it too, everything worked fine. So, a `files('foo')` worked but a `files('foo')[0]` did not, which is silly... and it's not exactly terrible to use files() here, the input is literally a list of source files.

Fixes building gnome-terminal
Fixes #9827


dcbaker (changes requested):
It looks like we actually do document the type in the gnome module documentation, so we should fix that too.

docs/markdown/Gnome-module.md
```diff
@@ -133,7 +133,6 @@ argument is the basename of the output files.
 * `nostdinc`: if true, don't include the standard marshallers from glib
 * `prefix`: the prefix to use for symbols
 * `skip_source`: if true, skip source location comments
-* `sources` []str *required*: List of string sources to consume
 * `sources`: the list of sources to use as inputs
```
dcbaker, line 136: We should at least kee the `*required*` here I think.

nirbheek:
Failing test on Arch Linux is `17 mpi` segfaulting, which happens from time to time.

dcbaker (approved):

nirbheek:
New test is failing because glib-genmarshal actually can't handle multiple source arguments correctly. Fixing...

Issue #9774, opened 2022-01-04
Fail to find ninja when `meson install` with VS15 environment


elmarco:
**Describe the bug**
`meson install` doesn't find ninja (but `meson build` does)

**To Reproduce**
Run meson with VS15
```
Found ninja-1.8.2 at C:\BuildTools\Common7\IDE\CommonExtensions\Microsoft\CMake\Ninja\ninja.EXE
Visual Studio environment is needed to run Ninja. It is recommended to use Meson wrapper:
C:\Python310\Scripts\meson compile -C builddir
```

meson build works
meson install fails with: Can't find ninja, can't rebuild test.

(I suspect the environment is modified with coredata or build.dat loading, so meson `build` works)

nirbheek:
I am not sure how to reproduce this. Could you paste / attach `builddir/meson-logs/meson-log.txt`?

nirbheek:
Okay, I can reproduce it.

elmarco:
thanks! sorry for the bad bug report ;)

PR #9833, opened 2022-01-15
install: Setup VS env if we did that during setup

nirbheek:
Otherwise we might not find a ninja that was picked up from the Visual
Studio installation.

```
$ meson setup _build
...
Activating VS 15.9.40
...
Found ninja-1.8.2 at "C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\Common7\IDE\CommonExtensions\Microsoft\CMake\Ninja\ninja.EXE"
$ meson compile -C _build
Activating VS 15.9.40
...
$ meson install -C _build
Can't find ninja, can't rebuild test.
```

Fixes https://github.com/mesonbuild/meson/issues/9774

### DIFF
--- a/mesonbuild/minstall.py
+++ b/mesonbuild/minstall.py
@@ -25,6 +25,7 @@ import subprocess
 import sys
 import typing as T
 
+from . import build
 from . import environment
 from .backend.backends import (
     InstallData, InstallDataBase, InstallEmptyDir, InstallSymlinkData,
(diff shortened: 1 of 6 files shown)


mesonbuild/minstall.py
```diff
@@ -792,6 +793,8 @@ def run(opts: 'ArgumentType') -> int:
     log_dir = os.path.join(private_dir, '../meson-logs')
     if not os.path.exists(os.path.join(opts.wd, datafilename)):
         sys.exit('Install data not found. Run this command in build directory root.')
+    b = build.load(opts.wd)
+    setup_vsenv(b.need_vsenv)
```
xclaesse, line 797: Shouldn't you do that inside the not no_rebuild condition below?

nirbheek: Good point!

nirbheek:
I've added a test that had to do a bunch of hacky stuff to function. Review would be appreciated.

nirbheek:
> I've added a test that had to do a bunch of hacky stuff to function. Review would be appreciated.

I slightly reworked the test to not be hacky. Will merge when the CI passes.

PR #9834, opened 2022-01-15
New keyword argument `verbose` for tests and benchmarks

bonzini:
The new keyword argument `verbose` can be used to mark tests and benchmarks
that must always be logged verbosely on the console.  This is particularly
useful for long-running tests, or when a single Meson test() is wrapping
an external test harness.  In this case, TAP should be used by the external
harness so that Meson will log each subtest as it runs, instead of including
the log at the end of the run.


docs/yaml/functions/benchmark.yaml
```diff
@@ -104,3 +104,13 @@ kwargs:
       The starting order of tests with identical priorities is
       implementation-defined. The default priority is 0, negative numbers are
       permitted.
+
+  verbose:
```
andy5995, line 108: Does adding the info here mean that https://mesonbuild.com/Reference-manual_functions.html#test gets updated as well?

bonzini: Yes, see [here](https://github.com/mesonbuild/meson/blob/master/docs/yaml/functions/test.yaml):

```yaml
posargs_inherit: benchmark
kwargs_inherit: benchmark
```

bonzini:
Ping.

docs/markdown/snippets/test-verbose.md
```diff
@@ -0,0 +1,8 @@
+## New keyword argument `verbose` for tests and benchmarks
+
+The new keyword argument `verbose` can be used to mark tests and benchmarks
+that must always be logged verbosely on the console.  This is particularly
+useful for long-running tests, or when a single Meson test() is wrapping
+an external test harness.  In this case, TAP should be used by the external
+harness so that Meson will log each subtest as it runs, instead of including
+the log at the end of the run.
```
jpakkane, line 8: The last sentence does not make sense to me. What is "in this case"? The thing it refers to does not even appear anywhere else in the snippet.

bonzini: It refers to "Meson test() is wrapping an external test harness".

jpakkane: But why does it talk about TAP? This is not directly tied to TAP as far as I can see.

bonzini: Without TAP, the output is logged all at once at the end of the run. It is better if the external harness uses TAP, because then `meson test` will separately log each `ok`/`not ok` line that the external harness emits.

jpakkane: This explanation should be in the snippet itself so people reading it can understand what is happening.

bonzini: I am not sure what was missing in the snippet, so I moved the whole explanation to the reference manual instead. This makes it possible to be less concise.

bonzini:
The cygwin failures seem unrelated, and boil down to `cannot find -lpython3.8`.

PR #9835, opened 2022-01-16
Subprojects.md:add extra detail

andy5995:
This adds a link to
https://mesonbuild.com/Builtin-options.html#specifying-options-per-subproject
in two different places and provides an example for changing the default
options.

This info is partially based on a [recent
discussion](https://github.com/mesonbuild/meson/discussions/9830)
between me and @eli-schwartz

### DIFF
--- a/docs/markdown/Subprojects.md
+++ b/docs/markdown/Subprojects.md
@@ -84,12 +84,13 @@ return variables as strings.
 ### Build options in subproject
 
 All Meson features of the subproject, such as project options keep
-working and can be set in the master project. There are a few
-limitations, the most important being that global compiler arguments
-must be set in the main project before calling subproject. Subprojects
-must not set global arguments because there is no way to do that
-reliably over multiple subprojects. To check whether you are running
-as a subproject, use [[meson.is_subproject]].
+working and [can be set in the master
+project](Builtin-options.md#specifying-options-per-subproject). There
+are a few limitations, the most important being that global compiler
+arguments must be set in the main project before calling subproject.
+Subprojects must not set global arguments because there is no way to
+do that reliably over multiple subprojects. To check whether you are
+running as a subproject, use [[meson.is_subproject]].
 
 ## Using a subproject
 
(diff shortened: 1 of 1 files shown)


docs/markdown/Subprojects.md
```diff
@@ -84,12 +84,13 @@ return variables as strings.
...
 
 All Meson features of the subproject, such as project options keep
-working and can be set in the master project. There are a few
-limitations, the most important being that global compiler arguments
-must be set in the main project before calling subproject. Subprojects
-must not set global arguments because there is no way to do that
-reliably over multiple subprojects. To check whether you are running
-as a subproject, use the `is_subproject` function.
+working and [can be set in the master
+project](https://mesonbuild.com/Builtin-options.html#specifying-options-per-subproject).
```
eli-schwartz, line 88:
```suggestion
project](Builtin-options.md#specifying-options-per-subproject).
```

The markdown files don't "know" that they are converted to HTML, and they definitely shouldn't hardcode the domain name. Elsewhere in the docs, we always do inter-doc links by referring to the markdown files.

docs/markdown/Subprojects.md
```diff
@@ -175,9 +176,22 @@ executable('my_project',
...
+libsimple_dep = dependency(
+  'libsimple',
+  fallback : ['libsimple', 'libsimple_dep'],
+  default_options: ['default_library=static']
+)
+```
+
+With this setup when libsimple is provided by the system, we use it (if
+'static' is specified for the subproject, the shared system library
+will be used). When that is not the case the fallback will be used.
```
eli-schwartz, line 194: This description feels a bit confusing. What about "With this setup when libsimple is provided by the system we use it (and completely ignore subproject options, i.e. we link to a shared system library)."

eli-schwartz (changes requested):

docs/markdown/Subprojects.md
andy5995: @eli-schwartz , thanks, I used your suggestion with a slight modification.

This PR is ready for re-review.

docs/markdown/Subprojects.md
```diff
@@ -175,9 +176,22 @@ executable('my_project',
   install : true)
 ```
 
-With this setup when libsimple is provided by the system, we use it.
-When that is not the case we use the embedded version (the one from
-subprojects).
```
eli-schwartz, line 180: It's not clear to me why you dropped the last sentence here.

eli-schwartz (commented):
These commits should really be squashed...

docs/markdown/Subprojects.md
andy5995: It was unintentional, I was working too late at night again. ;) Fixed (afaik)

eli-schwartz: Thanks, I was wondering if I had missed something!

PR #9838, opened 2022-01-18
Wrap download recovery

### LINKED ISSUE #9688: Make Meson recover from bad download
**Describe the bug**
When meson encounters an error while downloading a dependency, it might get into inconsistent state with following error:
```
meson.build:147:0: ERROR: Subproject exists but has no meson.build file
```

**To Reproduce**
```meson
# meson.build
project('test', 'c')

executable('main', 'main.c',
  dependencies: [
    subproject('libuv').get_variable('libuv_dep')
  ]
)
```
```c
// main.c
    return 0;
}
```
```ini
# subprojects/libuv.wrap
[wrap-file]
directory = libuv-v1.42.0
source_url = https://dist.libuv.org/dist/v1.42.0/libuv-v1.42.0.tar.gz
source_filename = libuv-v1.42.0.tar.gz
source_hash = 43129625155a8aed796ebe90b8d4c990a73985ec717de2b2d5d3a23cfe4deb72
patch_filename = libuv_1.42.0-1_patch.zip
patch_url = https://bad-wrapdb.mesonbuild.com/v2/libuv_1.42.0-1/get_patch
patch_hash = 891698d781bc68379b799e47bc7e100698f03985b6db5d1ddc4eaa9be42a8f35

[provide]
libuv = libuv_dep
```

`bad-wrapdb` above will simulate a network error (in practice error might be caused by network timeout, temporary DNS resolution issue, etc.).

Now do usual `meson setup build` and you'll get an obvious error.

If you replace `bad-wrapdb` with `wrapdb` to simulate the end of intermittent network issues you'll get:
```
meson.build:3:0: ERROR: Subproject exists but has no meson.build file
```

**Expected behavior**
Meson shoudn't keep subprojects in inconsistent state when such issues happen and thus should be able to recover without requiring user to go and manually remove directories under `subprojects`.

Retries would also be nice to have I think, like 3 attempts with exponential backoff.

**system parameters**
* Plain native build
* Ubuntu 21.04
* Python 3.9.5
* Meson 0.60.2
* Ninja 1.10.2.git.kitware.jobserver-1

eli-schwartz:
- recover from bad downloads by deleting broken contents we just created
- use exponential backoff to try downloading files on error

Fixes #9688

### DIFF
--- a/mesonbuild/wrap/wrap.py
+++ b/mesonbuild/wrap/wrap.py
@@ -26,6 +26,7 @@ import stat
 import subprocess
 import sys
 import configparser
+import time
 import typing as T
 import textwrap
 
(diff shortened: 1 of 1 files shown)


eli-schwartz:
One quick update to make it play better with --fatal-meson-warnings...

nazar-pc (approved):
Makes sense to me and works nicely as well :+1:

PR #9839, opened 2022-01-18
implement 'dist --allow-dirty' flag

### LINKED ISSUE #9824: emit error message and exit if the repo has uncommitted changes
I just wanted to suggest putting this warning at the bottom of the `meson dist` output

>WARNING: Repository has uncommitted changes that will not be included in the dist tarball

I had a bit of trouble because I hadn't read the paragraph in the docs about [how `meson dist` works](https://mesonbuild.com/Creating-releases.html#autotools-dist-vs-meson-dist), and I feel that it's an important warning that could be easily missed (especially by nubs like myself) after it's scrolled away.

andy5995:
closes #9824

### DIFF
--- a/docs/markdown/Creating-releases.md
+++ b/docs/markdown/Creating-releases.md
@@ -60,6 +60,14 @@ for example when done in CI that already does its own testing.
 So with `--no-tests` you can tell Meson "Do not build and test generated
 packages.".
 
+## Use `--allow-dirty` to override error when git repository contains uncommitted changes
+
+*Since 0.62.0* Instead of emitting a warning when a repository contains
+uncommitted changes, Meson will produce an error message notifying the
+user and immediately exit. If `--allow-dirty` is given as an option to
+`meson dist`, a warning will be emitted instead and Meson will proceed
+as usual.
+
 ## Release a subproject separately
 
 *Since 0.57.0* the `meson dist` command can now create a distribution tarball
--- a/mesonbuild/mdist.py
+++ b/mesonbuild/mdist.py
@@ -40,6 +40,8 @@ archive_extension = {'gztar': '.tar.gz',
 def add_arguments(parser):
     parser.add_argument('-C', dest='wd', action=RealPathAction,
                         help='directory to cd into before running')
+    parser.add_argument('--allow-dirty', action='store_true',
+                        help='Allow even when repository contains uncommitted changes.')
     parser.add_argument('--formats', default='xztar',
                         help='Comma separated list of archive types to create. Supports xztar (default), gztar, and zip.')
     parser.add_argument('--include-subprojects', action='store_true',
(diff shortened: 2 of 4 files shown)


docs/markdown/Creating-releases.md
```diff
@@ -60,6 +60,16 @@ for example when done in CI that already does its own testing.
 So with `--no-tests` you can tell Meson "Do not build and test generated
 packages.".
 
+## Use `--allow-dirty` to override error when git repository contains uncommitted changes
+
+*Since 0.61.2* Instead of emitting a warning when a repository contains
```
andy5995, line 65: Obviously this number might need changing depending on the next release and when this PR is merged, but for now I put it there as a placeholder

eli-schwartz: It will always be scheduled for the next major.minor release (a feature release, in this case probably 0.62.0), because we do not backport features into patch-level releases.

docs/markdown/Creating-releases.md
```diff
@@ -60,6 +60,16 @@ for example when done in CI that already does its own testing.
...
 
+## Use `--allow-dirty` to override error when git repository contains uncommitted changes
+
+*Since 0.62.0* Instead of emitting a warning when a repository contains
+uncommitted changes, Meson will produce an error message notifying the
+user and immediately exit. If `--allow-dirty` is given as an option to
+`meson dist`, a warning will be emitted instead and Meson will proceed
+as usual.
+
+...
```
eli-schwartz, line 71: Hmm, what is this doing here?

mesonbuild/mdist.py
```diff
@@ -70,7 +72,18 @@ def copy_git(src, distdir, revision='HEAD', prefix=None, subdir=None):
         t = tarfile.open(fileobj=f) # [ignore encoding]
         t.extractall(path=distdir)
 
-def process_submodules(src, distdir):
+msg_uncommitted_changes = 'Repository has uncommitted changes that will not be included in the dist tarball'
+msg_info_dirty_override = 'Use --allow-dirty to ignore the warning and proceed anyway'
```
eli-schwartz, line 76: The second string is only used once, it should be inlined into the `mlog.error` down below.

The first string used to be duplicated in two places. I have no strong feelings about whether or not to move it into a variable.

eli-schwartz (changes requested):

docs/markdown/Creating-releases.md
eli-schwartz: To clarify: the 3 dots over there seem to not have a purpose.

eli-schwartz (approved):

docs/markdown/Creating-releases.md
```diff
@@ -60,6 +60,14 @@ for example when done in CI that already does its own testing.
 So with `--no-tests` you can tell Meson "Do not build and test generated
 packages.".
 
+## Use `--allow-dirty` to override error when git repository contains uncommitted changes
+
+*Since 0.62.0* Instead of emitting a warning when a repository contains
+uncommitted changes, Meson will produce an error message notifying the
+user and immediately exit. If `--allow-dirty` is given as an option to
```
tristan957, line 67: Should no warning be given in this case if you are explicitly asking Meson to allow this?

andy5995: @tristan957 That's a good point and I can't argue with your logic. xD

I don't have strong feelings either way, I pretty much do whatever Eli tells me to do. :)

PR #9847, opened 2022-01-19
Enable wrapped dependencies to link back to the project's website if one is available.

paulherman:


tools/regenerate_docs.py
```diff
@@ -131,7 +131,8 @@ def generate_wrapdb_table(output_dir: Path) -> None:
             dependency_names_str = ', '.join(dependency_names)
             program_names = info.get('program_names', [])
             program_names_str = ', '.join(program_names)
-            f.write(f'| {name} | {versions_str} | {dependency_names_str} | {program_names_str} |\n')
+            website = info.get('website', '#')
+            f.write(f'| [{name}]({website}) | {versions_str} | {dependency_names_str} | {program_names_str} |\n')
```
mensinda, line 135: Markdown links should only be generated if there is an actual website to link to. Otherwise, dummy links are generated which are visually indistinguishable from *actual* links.

With the current setup, hotdoc also complains about a lot of `Trying to link to non-existing identifier ‘#’`.

eli-schwartz:
This is missing context, such as an explanation or a feature request.

So, to rephrase:

Hi, I'm @paulherman and I feel there is a need that could be fulfilled in the https://github.com/mesonbuild/wrapdb/ project. Currently, wrapdb projects can have a:
- name identifier
- version
- dependency provide

I would like them to also provide:
- a marketing website

Here is a PR that doesn't add a marketing website to wrapdb projects, but it does turn the mesonbuild list of known wraps into hyperlinks pointing to the marketing website (that isn't known).

...

So, there is some discussion that could be had here. First, does https://github.com/mesonbuild/wrapdb/ want to add information about marketing websites? Currently, all the information there is about programmatically interacting with those projects.

If we do want it, we should probably begin by adding that to all wraps.

Do we want it to be required info? If so, this should be checked in `tools/sanity_checks.py`, and third-party consumers can safely assume it exists. If not, why not?

...

And on the topic of meson, rather than the wrapdb, should `meson wrap info` report the marketing url?

paulherman:
Fair enough. Apologies for sending this straight away without starting a discussion! I've done so in https://github.com/mesonbuild/wrapdb/issues/292.

tools/regenerate_docs.py
jpakkane: Regardless if there is no website defined the entry should be empty, otherwise something like `[website](https://url.to.upstream)`.

paulherman: Should be done. Is there any particular place where the JSON schema is documented? I believe it should be mentioned that the "website" field is optional.

xclaesse: @paulherman the json format is loosely defined in wrapdb's README. But again, I think it makes sense to have that info in the .wrap file instead of releases.json, it can be generally useful even outside of wrapdb.

paulherman: That makes sense. I think this PR stands alone since passing this through the .wrap file is in a different repo and I'll do the following:
1. change meson/mesonbuild/wrap/wraptool.py, specifically `info` to parse the website
2. change wrapdb/tools/import-wraps.py to extract the website from the `meson wrap info <name>` command

I think these can also be individual PRs. WDYT @xclaesse?

xclaesse: You don't need to change `wrapdb/tools/import-wraps.py` that's a script that I used for the initial import into wrapdb, it's not used any more. Could delete it actually.

Hmm, actually I did not realize that we don't have the wrap file when generating this doc page. Downloading each wrap file from wrapdb individually is not going to scale... You're probably right that we need that website into releases.json then.

paulherman:
@jpakkane could you please take another look at this? I believe your explicit approval is required since this is my first PR.

Issue #8113, opened 2020-12-19
meson 0.56.0 errors out with Intel(R) oneAPI DPC++ Compiler 2021.1 (2020.10.0.1113)


ms178:
I am on Manjaro 20.2 and installed the new Intel(R) oneAPI DPC++ Compiler 2021.1 (2020.10.0.1113) which is now called "icx" for C and "icpx" for C++ (formerly known as: "icc" and "icpc").

I tried to compile some projects which use meson with makepkg from AUR and albeit having the compiler setup correctly, meson errors out with:

Using 'PKG_CONFIG_PATH' from environment with value: '/opt/intel/oneapi/mkl/latest/tools/pkgconfig'
Using 'PKG_CONFIG_PATH' from environment with value: '/opt/intel/oneapi/mkl/latest/tools/pkgconfig'
The Meson build system
Version: 0.56.0
Source dir: /home/marcus/Downloads/mesa-git/src/mesa
Build dir: /home/marcus/Downloads/mesa-git/src/_build
Build type: native build
Program python found: YES (/usr/bin/python)
Project name: mesa
Project version: 21.0.0-devel
Using 'CC' from environment with value: 'icx'

mesa/meson.build:21:0: ERROR: Unknown compiler(s): [['icx']]

I've used both methods described here (altered for the new 2021 edition) to setup the compiler with the same results as seen above: https://wiki.archlinux.org/index.php/Intel_C%2B%2B

I've also tried to modify the PKGBUILD file with adding CC, CXX and CC_LD options as described here: https://mesonbuild.com/howtox.html

The issue persists though. 

I've found a pull request which seems to be related to the underlying issue, but I am not an expert: https://github.com/mesonbuild/meson/pull/6827

dcbaker:
This is probably due to changes in the output of icx compared to icc (the the C++ variants). I'll see if I can get icx installed this week and look at what's changed.

ms178:
Thanks for taking a look. A lot has changed with the new release. They now use a modified Clang under the hood, icc/icpc lives on as the classic compiler (which this issue is not about, I haven't tried that newer release yet but icc/icpc did work with an older release some weeks ago). There is also another c++ compiler bundled with the new release, which is called "dpcpp" - which brings SYCL support.

dcbaker:
[...]

PR #9850, opened 2022-01-20
Add compiler detection for Intel oneAPI DPC++/C++ Compiler.

vinsonlee:
Reported-by: Marcus Seyfarth <m.seyfarth@gmail.com>
Suggested-by: Harry van Haaren <harry.van.haaren@intel.com>
Tested-by: Vinson Lee <vlee@freedesktop.org>
Tested-by: Marcus Seyfarth <m.seyfarth@gmail.com>
Signed-off-by: Vinson Lee <vlee@freedesktop.org>

### DIFF
--- a/mesonbuild/compilers/detect.py
+++ b/mesonbuild/compilers/detect.py
@@ -518,7 +518,7 @@ def _detect_c_or_cpp_compiler(env: 'Environment', lang: str, for_machine: Machin
             return cls(
                 compiler, version, for_machine, is_cross, info, target,
                 exe_wrap, linker=linker)
-        if 'clang' in out or 'Clang' in out:
+        if 'clang' in out or 'Clang' in out or 'oneAPI DPC++/C++' in out:
             linker = None
 
             defines = _get_clang_compiler_defines(compiler)


harry-van-haaren:
Just a note that this PR is the output of the longer discussion on issue https://github.com/mesonbuild/meson/issues/8113, and the code-change workings is described in https://github.com/mesonbuild/meson/issues/8113#issuecomment-961945940.

Merging this PR will enable Meson to correctly detect and use the OneAPI compiler.

jpakkane:
As this is new functionality it should have a release note snippet.

dcbaker:
Isn't this going to result in using th eIntel compiler with an Id as clang? Even if it uses clang under the hood, there are still some differences right, enough to justify having at least a unique name like "oneAPI" or "Intel DPC" or something (just not `intel` or `intel-cl`)?

mathstuf:
I think a new ID is called for (CMake gives it `IntelLLVM`, but `oneapi` or `intel-oneapi` is probably fine). As a concrete use case, the Intel OneAPI compiler has some…interesting decisions around some flag defaults that projects probably want to be able to detect it specifically (of note, `-fp-model=precise` is needed to remove `-ffast-math`-like behaviors on `-O1` and above).

dcbaker:
That's... interesting. And means that the implementation in Meson needs to override the optimization levels as well, so that we add -fp-mode=precise to the arguments by default

rgommers:
> I think a new ID is called for (CMake gives it `IntelLLVM`,

This is what gh-10909 does. So that probably supercedes this PR.

eli-schwartz:
It adds a new class, but doesn't override anything such as the ID.

It is the right approach, just not all the way there. :)

aaronsuydam:
Just noting that this hasn't been touched in 2 years, and it'd be really nice to have!

dcbaker:
@aaronsuydam #10909 implements this, and has been landed for almost 2 years. The only think I see here that we don't have upstream is the note about the `-fp-model=precise`. Since the intention of this is already upstream, I'm going to close this.

aaronsuydam:
@dcbaker ah ok great. There's nothing in the web docs about this having been implemented (that I saw), what's the meson.build option I need to use? Haven't been able to get it to work.

eli-schwartz:
@aaronsuydam there isn't a meson.build option. It is supporting a new compiler and compilers are defined in the toolchain ("machine") file or inferred by setting the environment variables CC and CXX.

The meson docs include a reference table listing supported compilers alongside the ID return values for checking inside meson.build "which compiler is this build directory + toolchain configured to use".

> intel-llvm --> Intel oneAPI LLVM-based compiler

It's also mentioned in the release notes for meson 0.64.0

PR #9853, opened 2022-01-20
Type checking for the CustomTarget initializer

dcbaker:
Since we now do all of our user input type checking in the interpreter, the `CustomTarget` initializer can act much more as a dumb initializer, and (for the most part) assigning values to attributes. This makes the whole thing much cleaner and easier to understand, and has allowed mypy to find at least one bug!

### DIFF
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import annotations
 from collections import OrderedDict
 from dataclasses import dataclass, field
 from functools import lru_cache
@@ -23,6 +24,7 @@ import pickle
 import re
 import textwrap
 import typing as T
+
 
 from . import environment
 from . import dependencies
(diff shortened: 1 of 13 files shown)


mesonbuild/interpreter/interpreter.py
```diff
@@ -1662,23 +1662,28 @@ def func_build_target(self, node, args, kwargs):
         else:
             raise InterpreterException('Unknown target_type.')
 
-    @permittedKwargs({'input', 'output', 'fallback', 'command', 'replace_string'})
-    @FeatureDeprecatedKwargs('custom_target', '0.47.0', ['build_always'],
-                             'combine build_by_default and build_always_stale instead.')
```
eli-schwartz, line 1667: This could be removed as its own distinct thing. I suspect rebase errors, but in commit 06481666f4e74ecef01e59351fc345ab0962d998 this line got removed from build.py and added into the interpreter, but to the wrong function.

Since then, build_always became a typed kwarg that carries that info around natively, but it never had the warning. It should.

I've done a more specific fix in #9887

mesonbuild/interpreter/interpreter.py
```diff
@@ -1776,10 +1785,8 @@ def _func_custom_target_impl(self, node, args, kwargs):
             # string in the meantime.
             FeatureNew('custom_target() with no name argument', '0.60.0', location=node).use(self.subproject)
             name = ''
-        kwargs['install_mode'] = self._get_kwarg_install_mode(kwargs)
         if 'input' in kwargs:
             kwargs['input'] = self.source_strings_to_files(extract_as_list(kwargs, 'input'), strict=False)
-        kwargs['env'] = self.unpack_env_kwarg(kwargs)
```
eli-schwartz, line 1782: Hmm, I'm guessing these could be removed independent of the typed_kwargs conversion? Neither is supported by vcs_tag, so they would always end up as None, and any other entry point into `_func_custom_target_impl` either has them as None also, or converts them correctly?

mesonbuild/interpreter/interpreter.py
```diff
@@ -1706,12 +1709,13 @@ def func_vcs_tag(self, node: mparser.BaseNode, args: T.List['TYPE_var'], kwargs:
                 source_dir,
                 replace_string,
                 regex_selector] + vcs_cmd,
-            'input': kwargs['input'],
-            'output': kwargs['output'],
-            'build_by_default': True,
-            'build_always_stale':True,
-        }
-        return self._func_custom_target_impl(node, [kwargs['output']], cmd_kwargs)
+            self.source_strings_to_files(kwargs['input'], strict=False),
```
eli-schwartz, line 1712: This `strict=False` hack is unique to `custom_target()` because we document passing a typelib file to custom_target as a manually crafted string, see commit 1e5d7f212246e967f3473a55f632f6f5eda87b65 for details.

This seems like an excellent opportunity to reduce the surface area of that hack, so that functions such as vcs_tag which happen to create a `build.CustomTarget` no longer get this silliness as a side effect of using `_func_custom_target_impl`.

mesonbuild/build.py
```diff
@@ -2405,119 +2438,6 @@ def get_transitive_build_target_deps(self) -> T.Set[T.Union[BuildTarget, 'Custom
...
-            if '@INPUT@' in i or '@INPUT0@' in i:
-                m = 'Output cannot contain @INPUT@ or @INPUT0@, did you ' \
-                    'mean @PLAINNAME@ or @BASENAME@?'
-                raise InvalidArguments(m)
-            # We already check this during substitution, but the error message
-            # will be unclear/confusing, so check it here.
-            if len(inputs) != 1 and ('@PLAINNAME@' in i or '@BASENAME@' in i):
-                m = "Output cannot contain @PLAINNAME@ or @BASENAME@ when " \
-                    "there is more than one input (we can't know which to use)"
-                raise InvalidArguments(m)
```
eli-schwartz, line 2435: AFAICT we are totally dropping this messaging entirely?

mesonbuild/build.py
```diff
@@ -2405,119 +2438,6 @@ def get_transitive_build_target_deps(self) -> T.Set[T.Union[BuildTarget, 'Custom
...
-            self.install = kwargs['install']
-            if not isinstance(self.install, bool):
-                raise InvalidArguments('"install" must be boolean.')
-            if self.install:
-                if not kwargs.get('install_dir', False):
-                    raise InvalidArguments('"install_dir" must be specified '
-                                           'when installing a target')
-
-            if isinstance(kwargs['install_dir'], list):
-                FeatureNew.single_use('multiple install_dir for custom_target', '0.40.0', self.subproject)
```
eli-schwartz, line 2475: This warning is being totally dropped?

mesonbuild/build.py
```diff
@@ -2405,119 +2438,6 @@ def get_transitive_build_target_deps(self) -> T.Set[T.Union[BuildTarget, 'Custom
...
-            # If only one tag is provided, assume all outputs have the same tag.
-            # Otherwise, we must have as much tags as outputs.
-            install_tag: T.List[T.Union[str, bool, None]] = typeslistify(kwargs.get('install_tag', []), (str, bool, type(None)))
-            if not install_tag:
-                self.install_tag = [None] * len(self.outputs)
-            elif len(install_tag) == 1:
-                self.install_tag = install_tag * len(self.outputs)
-            elif install_tag and len(install_tag) != len(self.outputs):
-                m = f'Target {self.name!r} has {len(self.outputs)} outputs but {len(install_tag)} "install_tag"s were found.'
-                raise InvalidArguments(m)
```
eli-schwartz, line 2489: The initializer just assumes anything that isn't length 1 is correct, without verifying that they are the same length. And this is not verified in the interpreter either.

This check is being dropped.

eli-schwartz (changes requested):

mesonbuild/build.py
dcbaker: And we can even add location now since we're handling this in the right place :)

dcbaker:
All fixed up locally (I hope), I'm just running tests locally before I push the udpate

mesonbuild/build.py
eli-schwartz: Indeed. :D

mesonbuild/interpreter/interpreter.py
```diff
@@ -1662,21 +1662,28 @@ def func_build_target(self, node, args, kwargs):
...
-    @permittedKwargs({'input', 'output', 'fallback', 'command', 'replace_string'})
     @noPosargs
-    def func_vcs_tag(self, node, args, kwargs):
-        if 'input' not in kwargs or 'output' not in kwargs:
-            raise InterpreterException('Keyword arguments input and output must exist')
-        if 'fallback' not in kwargs:
+    @typed_kwargs(
+        'vcs_tag',
+        CT_INPUT_KW,
+        CT_OUTPUT_KW,
```
eli-schwartz, line 1671: We no longer manually code that this is required, because `KwargInfo(..., required=True)` yay.

But, while CT_OUTPUT_KW is required, custom_target has optional input, so CT_INPUT_KW needs to evolve to required.

mesonbuild/interpreter/interpreter.py
```diff
@@ -1706,19 +1714,34 @@ def func_vcs_tag(self, node: mparser.BaseNode, args: T.List['TYPE_var'], kwargs:
...
+        """Checks for additional invalid values in a custom_target output.
+
+        This cannot be done with typed_kwargs because it requires the number of
+        inputs.
+        """
+        for out in outputs:
+            if has_multi_in and ('@PLAINNAME@' in out or '@BASENAME@' in out):
+                raise InvalidArguments(f'{name}: output cannot containe "@PLAINNAME@" or "@BASENAME@" '
+                                       'when there is more than one input (we can\'t know which to use)',
+                                       file=node.filename, lineno=node.lineno, colno=node.colno)
```
eli-schwartz, line 1743: We generally do not pass node information to `InvalidArguments()`, because the interpreter loop already catches exceptions and modifies them to add the current_node information to them.

We do need to pass node information around for Feature checks, but that is because the interpreter doesn't stop everything when it sees a Feature check in order to modify that. :D

mesonbuild/interpreter/interpreter.py
```diff
@@ -1747,49 +1770,88 @@ def func_custom_target(self, node: mparser.FunctionNode, args: T.Tuple[str],
...
+        if kwargs['install'] and not kwargs['install_dir']:
+            raise InvalidArguments('custom_target: "install_dir" keyword argument must be set when "install" is true.')
+        if len(kwargs['install_dir']) > 1:
+            FeatureNew.single_use('multiple install_dir for custom_target', '0.40.0', self.subproject, location=node)
+        if len(kwargs['install_tag']) not in {0, 1, len(kwargs['output'])}:
+            raise InvalidArguments('custom_target: install_tag argument must have 0 or 1 outputs, '
+                                   'or the same number of elements as the output keyword argument. '
+                                   f'(there are {len(kwargs["install_tag"])} install_tags, '
+                                   f'and {len(kwargs["output"])} outputs)',
+                                   file=node.filename, lineno=node.lineno, colno=node.colno)
```
eli-schwartz, line 1829: Probably-extraneous node information here too.

eli-schwartz (commented):
Just a few last nits.

mesonbuild/interpreter/interpreter.py
dcbaker: It's still not 100% correct (though as correct as it was previously), since we really should be setting `allow_empty=False`, but we don't have a way to `evolve` that in. Probably we should change the semantics of `required=True` to mean `allow_empty=False`, but that's a separate change, and this does give us parity with the original implementation

eli-schwartz: Ah, indeed, that does seem a bit unfortunate...

Really, once again that would be a case where we should not allow empty ever, but instead have your future thoughts for a `default()` object.

Oh well,. more typed_kwargs means we get closer to the day when that can be sanely implemented, right?

Is it worth putting in a comment here?

eli-schwartz (approved):

PR #9854, opened 2022-01-20
Fully type check the sourceset module

dcbaker:


bonzini:
The typed*Args-based code is much nicer, and it should be safe enough since the module is well covered by the testsuite. Thanks!

eli-schwartz:
Indeed, and that testsuite coverage helpfully errors out many times because the typechecking is wrong and doesn't allow dicts. :)

dcbaker:
I'm 90% of the way to making ConfigurationData just a wrapper around dict which would make this all work transparently...

mesonbuild/modules/sourceset.py
```diff
@@ -125,7 +177,7 @@ def add_all_method(self, state: ModuleState, args: T.Tuple[T.List[SourceSetImpl]
         for s in if_true:
             if not isinstance(s, SourceSetImpl):
                 raise InvalidCode('Arguments to \'add_all\' after the first must be source sets')
-            s.frozen = True
+        s.frozen = True
```
eli-schwartz, line 189: This seems broken, we need to mark each sourceset as frozen once we add its contents into our new sourceset. After this change, we only mark the last one as frozen.

eli-schwartz (commented):
I neglected to review this earlier so it got merged without review... sorry for coming in post-merge with complaints.

PR #9855, opened 2022-01-21
Add meson.command() as a way to find out how to run Meson

Akaricchi:
## Rationale
It's sometimes useful to invoke Meson from a `run_target()` or even from the built application, but currently it's not possible to do this reliably, e.g. we have to assume that `meson` is in `$PATH` and it's the right one. Exposing `Environment.get_build_command()` seems like a good solution, since this is what gets hardcoded into `build.ninja`/other backend-specific files.

I currently have two use-cases for this:
* Our project has `run_target()`s that generate binary release packages. They work by calling a python script that installs the project into a temporary directory (calling `meson install`), then doing some additional processing and generating an archive/windows installer/etc.
* I'd like to implement a simple live reloading system for our game. The idea is to have it monitor source files for changes and call `meson compile` to rebuild a `shared_module()`. The meson command and path to sources could be passed through devenv, or directly compiled into the executable.

## ...and naming things
I'm not sure whether the method should be called `command()`, `get_command()`, or something else. The compiler object has `cmd_array()` for equivalent functionality, which I'm not a big fan of, but I'm ok with using that name for consistency's sake.

### DIFF
--- a/docs/yaml/builtins/meson.yaml
+++ b/docs/yaml/builtins/meson.yaml
@@ -446,3 +446,8 @@ methods:
       env:
         type: env
         description: The [[@env]] object to add.
+
+  - name: command
+    returns: list[str]
+    since: 0.62.0
+    description: Returns the command used to run Meson as an array, with arguments.
--- a/mesonbuild/interpreter/mesonmain.py
+++ b/mesonbuild/interpreter/mesonmain.py
@@ -76,6 +76,7 @@ class MesonMain(MesonInterpreterObject):
                              'has_external_property': self.has_external_property_method,
                              'backend': self.backend_method,
                              'add_devenv': self.add_devenv_method,
+                             'command': self.command_method,
                              })
 
     def _find_source_script(
@@ -448,3 +449,9 @@ class MesonMain(MesonInterpreterObject):
         converted = ENV_KW.convertor(env)
         assert isinstance(converted, build.EnvironmentVariables)
         self.build.devenv.append(converted)
+
+    @FeatureNew('command', '0.62.0')
+    @noPosargs
+    @noKwargs
+    def command_method(self, args: T.List['TYPE_var'], kwargs: 'TYPE_kwargs') -> T.List[str]:
+        return self.interpreter.environment.get_build_command()


jpakkane:
When Meson runs external commands it sets some envvars for exactly this use case, [see here](https://mesonbuild.com/Reference-tables.html#script-environment-variables). I'm almost certain that at some point we added the Meson command itself to that but grepping does not find anything. However the introspection data should have the things you need and if not, then either adding what you need there or possibly adding Meson command itself to the envvars should be preferred to adding new functions to Meson.

eli-schwartz:
There is another PR attempting to do the exact same thing (with lots of existing discussion and an associated issue): https://github.com/mesonbuild/meson/pull/8572

Akaricchi:
@jpakkane 

> When Meson runs external commands it sets some envvars for exactly this use case, see here. I'm almost certain that at some point we added the Meson command itself to that but grepping does not find anything.

There is `MESONINTROSPECT`, which is close. I suppose I could extract the meson command from that by removing the `introspect` argument, but that doesn't seem very… clean. It also does not solve my second use-case, where I want to pass the meson command to a built target in some way, which is not ran by Meson. Encoding the command line in a plain string also does not seem like a good idea… are we expected to `shlex.split` it? I would rather just have access to the array.

> However the introspection data should have the things you need and if not, then either adding what you need there or possibly adding Meson command itself to the envvars should be preferred to adding new functions to Meson.

I have checked the json files under `meson-info`, there's no meson command in there. Even if there was, I'd prefer not having to parse it out of json in my C program.

@eli-schwartz 

> There is another PR attempting to do the exact same thing (with lots of existing discussion and an associated issue): #8572

Well, oops, should've checked. Good to know I'm not the only one who wants this feature.

bonzini:
> are we expected to shlex.split it? I would rather just have access to the array.

True, I think that's the best justification of this feature as opposed to an environment variable.

eli-schwartz:
Either the universe implodes, or alternative implementations error out with an implementation of this method that returns

```meson
error('the meson CLI is not portable between implementations and this meson.build hardcodes implementation details of of the reference python-based one')
```

bonzini:
Does it have to be identical? I think mesonbuild should define the basic level of CLI support for all meson implementations, it could be as simple as `meson {setup  [-Cxxx] [dir],compile [-Cxxx],dist [-Cxxx] [--no-tests] [--include-subprojects],install [--no-rebuild]}`.

eli-schwartz:
Exposing the ninja command might be easier, though, which would solve the live reloading case if not the one that runs `meson install` in a packaging script.

Muon doesn't write out an install rule because it's assumed one would just run `muon -C builddir install` manually (note even that already has CLI interface differences!)

Perhaps @annacrombie would be tempted to add a build.ninja `install` rule for the sake of portability in packaging scripts.

Or perhaps the packaging script could be reframed by use of `meson.add_postinstall_script()` and then running `meson install` and expecting it to inherently produce binary installer packages.

eli-schwartz:
@bonzini we cross-posted...

As I mentioned above, there's already a fairly big difference in that muon expects -C to be an argument to muon, not to the subcommand, and quite frankly it makes sense and I wish meson could reasonably do the same but it would probably be very annoying to implement.

bonzini:
> I wish meson could reasonably do the same but it would probably be very annoying to implement.

Either very annoying or very hackish. Perhaps a hackish solution is fine.

dcbaker:
and Muon and Meson++ both implement a command as a combined setup + configure, I think (I know meson++ does and I expect that Muon does).

annacrombie:
> Perhaps @annacrombie would be tempted to add a build.ninja install rule for the sake of portability in packaging scripts.

I would be okay with this for compatibility.  I don't like having install a rule of build.ninja because of the mess with permissions that you have to worry about (install was invoked as root, but causes rebuilds).

As far as general CLI compatibility goes, I feel like the CLI of meson.py was one of my biggest pain points using meson, and I don't really want to re-create that for muon.

eli-schwartz:
`muon install` doesn't check that `ninja all` is up to date (but `meson install` without `--no-rebuild` does). Perhaps `ninja install` could dispatch to `muon install` without actually depending on any targets.

This is actually hugely inconsistent across the ecosystem. Especially handwritten Makefiles often don't have their install rules depend on the build targets, but assume that people first run a build, then run an install immediately after. In the current state of affairs, muon makes the same assumption.

It's not a terrible rationale, as you say it avoids implicit building as root. The only  alternative I can imagine would be having `muon install` itself fork and run ninja after dropping privileges `runuser`-style.

Akaricchi:
> what happens if the command is muon or meson++, neither of which implements an identical cli to meson?

I think it's reasonable to expect alternative implementations to be at least somewhat CLI-compatible for scripting purposes. If they aren't, then I just can't support them if my use-case requires calling meson, whether I find out the correct command properly or blindly assume it's `meson` in `PATH`.

> Exposing the ninja command might be easier, though, which would solve the live reloading case if not the one that runs meson install in a packaging script.

That would obviously break on any backends that aren't ninja. That's the idea behind `meson compile` etc.; it's backend-agnostic.

> Or perhaps the packaging script could be reframed by use of meson.add_install_script() and then running meson install and expecting it to inherently produce binary installer packages.

No, that's not acceptable. I don't always want to produce a release package when I install, and there may be multiple kinds of release packages to produce (e.g. a plain `.zip` and an NSIS installer for windows) that I want on separate build targets.

dcbaker:
I can't speak for @annacrombie, but I don't plan to implement the meson cli exactly in meson++.

meson++ has `configure` which is a combined `setup` and `configure`, but the flags are different. I'm not going to split them, I think it makes more sense to have a single command than two.

Likely I'd have to implement it in the way @eli-schwartz suggests. Unless we also add something like `meson.implementation()` and expect people to switch on it.

Akaricchi:
I can see a few ways to resolve the cross-implementation compatibility issue:

1. Ignore it. If your CLI is incompatible with Meson's, state that in your documentation and maybe indeed error out or warn on the `meson.command()` function. You'll have to accept that your implementation will be necessarily incompatible with some projects, but that's a valid choice, and 100% compatibility is probably not your goal anyway.
2. We don't have to require `meson.command()` to return the actual meson command intended for user interaction. We can just spec it to have an interface similar to @bonzini's suggestion. In that case Meson can return the user-facing command since its interface is compatible, and your implementation can return something like `/usr/bin/meson++ internalcommand mesoncompat`. You'll still have to implement a meson-like CLI for that subcommand, but:
    * You have a clear (and probably simpler) specification to follow rather than "do whatever meson does"
    * You don't compromise your user-facing CLI
    * You retain compatibility with projects that need to call meson
3. Device some crazy API to construct meson command lines, with an abstraction for most useful things you could do with the meson command. For example:
    ```meson
    cmd = meson.command('compile', target : 'foobar')
    # ['meson', 'compile', '-C', '/home/yukari/myproject/build', 'foobar']
    ```
    ```meson
    cmd = meson.command('install',
        rebuild : true,
        tags : ['reimu', 'marisa'])
    # ['meson', 'install', '-C', '/home/yukari/myproject/build', '--tags', 'reimu,marisa']
    ```
    I don't really like this solution because it seems rather overcomplicated, but it might have some advantages, like being able to automatically pass the build directory and translate target names when running in a subproject, etc.

Akaricchi:
In any case, I want to stress that any solution would be strictly better than the current situation. Not being able to know the right meson command won't stop me from trying to guess it and relying on meson implementation details, which obviously rules out any kind of compatibility with alternative meson implementations.

eli-schwartz:
> No, that's not acceptable. I don't always want to produce a release package when I install, and there may be multiple kinds of release packages to produce (e.g. a plain .zip and an NSIS installer for windows) that I want on separate build targets.

FWIW in that case you'd probably add a meson_options.txt option to toggle whether to build packages and/or which ones to create.

You'd probably also want to perform a DESTDIR install anyway, so you could also gate the package building logic in a sort of reverse `gnome.postinstall()` kind of thing (all of which are only supposed to run when DESTDIR is *not* set).

That being said, having a method to return the implementation name might not be a bad idea in general. It should be used with extreme caution since the point of having multiple implementations is that they all work the same to build a project, but indeed for niche cases such as running scripts using the implementation command having all implementations behave the same may be out of scope.

Akaricchi:
> FWIW in that case you'd probably add a meson_options.txt option to toggle whether to build packages and/or which ones to create.
>
> You'd probably also want to perform a DESTDIR install anyway, so you could also gate the package building logic in a sort of reverse gnome.postinstall() kind of thing (all of which are only supposed to run when DESTDIR is not set).

That would be a an exceptionally awful interface. Generating a package this way can't even be done in a single command, requires a re-configure step per package built, requires me to manually specify a temporarily DESTDIR that I'm going to discard immediately after packaging, and pollutes the already miserable output of `meson configure` even further. I honestly can't think of a worse way to implement this. This is a total abuse of `meson.add_install_script()`.

With my approach, generating e.g. windows release packages is as simple as `ninja zip nsis`. Other than having to guess the correct meson command inside the `run_command()` script, it works just perfectly.

annacrombie:
I kind of like option 3.  Maybe you could have multiple functions like `meson.install_command(tags: ['reimu', 'marisa'])`, `meson.compile_command(target: 'foobar')`

Option 2 is probably how `muon` would work if a simple `meson.command()` was implemented.

dcbaker:
That’s where I’m leaning as well

eli-schwartz:
Alright so, we're going to spec out a simple stable interface for scripting then?

dcbaker:
I think so, I’d prefer if we kept a fairly narrow set of things in that list, say install, compile, dist?

Akaricchi:
> I kind of like option 3. Maybe you could have multiple functions like meson.install_command(tags: ['reimu', 'marisa']), meson.compile_command(target: 'foobar')

Agreed; if we're doing option 3 that's probably the better way to do it. For cleaner documentation of the kwargs if nothing else.

I'm personally more in favor of option 2, but no strong preference here.

annacrombie:
Either 2 or 3 should have a specced, stable interface.  It's just a matter of stability of a meson-dsl api, or stability of a part of the CLI.

annacrombie:
muon now puts MUON_PATH into the environment as a workaround until this is decided on.

xclaesse:
My 2 cents:
- I think it should be named `meson.cmd_array()` to be consistent with `compiler.cmd_array()`.
- I don't think muon compat is a big issue, the CLI is part of Meson documentation and stability guarantees, muon can provide a compatibility layer. Muon could implement that method by returning `['muon', '--meson-compat']` or provide a separate `muon-compat` binary.
- My use-case is to run commands inside devenv, something like `run_target('run', command: meson.cmd_array() + ['devenv', exe_target])`. I think it could be useful to also be able to run custom_target() or test() inside devenv this way too.

eli-schwartz:
> the CLI is part of Meson documentation and stability guarantees

It's certainly part of the documentation, which I'm not sure proves anything (what, just because the DSL is documented on the same site as the CLI?)

Why should Meson condone having the CLI be part of the DSL spec?

> My use-case is to run commands inside devenv

So muon cannot claim to implement the build language unless it also provides devenv?

I'm not really sure why custom targets or executables need to run inside devenv anyway. They should already have environment variables set up where it can be logically inferred. And the correct solution is to add an env object to that command anyway.

xclaesse:
> Why should Meson condone having the CLI be part of the DSL spec?

It just means that muon **can** implement meson CLI if it wants. If muon wants to implement meson.cmd_array() properly, there is nothing preventing it to provide a CLI compat layer. But muon could also just decide it's up to the user to deal with CLI differences, or just accept to leave it unimplemented. I don't think such layer needs to be in meson itself.

I'm saying that because I don't like the proposition of duplicating an already well defined CLI as part of the DSL with API like `meson.command('compile', target : 'foobar')`.

> I'm not really sure why custom targets or executables need to run inside devenv anyway. They should already have environment variables set up where it can be logically inferred. And the correct solution is to add an env object to that command anyway.

In most cases it's true that custom_target can set its own env variable, but it has limitations:
- env is duplicated with what is already set using meson.add_devenv().
- devenv contains many things automatically, e.g. `GI_TYPELIB_PATH`.
- running a command could need env a dependency subproject has defined with meson.add_devenv().

bonzini:
Perhaps you need a meson.project_devenv() method that returns an environment object and prevents further calls to add_devenv() from the current project and any subprojects of the current project?

annacrombie:
muon now has a cli compatibility layer with meson (`muon meson`) which translates a good deal of meson [test|install|setup] into the equivalent muon cli syntax.

xclaesse:
We decided to special case `find_program('meson')`. I think this PR can be closed now.

Issue #9858, opened 2022-01-21
Spurious "configuration_data.set10 with number" warnings


anarazel:
In current git (acef5a976 at the time of writing) uses of configuration_data.set10 warn about parameters that are not booleans - even if the parameters actually are. As far as I can tell it's not possible to use set10 *without* triggering the warning.

The problem is that the code checks for the value being an integer using `isinstance(arg, int)` - unfortunately that is true for boolean values as well:

```python
    @typed_pos_args('configuration_data.set10', str, (int, bool))
    @typed_kwargs('configuration_data.set10', _CONF_DATA_SET_KWS)
    def set10_method(self, args: T.Tuple[str, T.Union[int, bool]], kwargs: 'kwargs.ConfigurationDataSet') -> None:
        self.__check_used()
        if isinstance(args[1], int):
            mlog.deprecation('configuration_data.set10 with number. the `set10` '
                             'method should only be used with booleans',
                             location=self.interpreter.current_node)
```

```python
>>> isinstance(False, bool)
True
>>> isinstance(False, int)
True
```

@dcbaker Looks like this comes from 9c7ddf59f

dcbaker:
Sigh, why did python have to inherit that disaster from C?

PR #9860, opened 2022-01-21
interpreterobjects: Don't warn on set10(bool)

dcbaker:
Python inherited a really bad design from C, booleans happen to be ints.
Which is awful, but it's true.

Fixes: #9858


anarazel:
Seems like a test might be worth it?

PR #9865, opened 2022-01-23
tests: Add test if compiler checks ignore args

ePirat:
As states in the meson documentation, compiler checks ignore compiler
arguments added with add_*_arguments, via the *_args options on the
command line or via the  respective environment variables like CLFAGS.

This test ensures this is actually the case by testing it for the
compiler.get_define and compiler.compiles checks.


ePirat:
No idea why this decreases coverage. Failure is expected, see #9866

ePirat:
Depends on #9936

ePirat:
The cross compilation failure is because meson apparently does not honor CFLAGS for the host compiler when cross-compiling. For now I could probably just disable the test for the cross case until this is fixed, as I dont think its something that should be done in this PR…

test cases/common/251 ignoring args for checks/meson.build
```diff
@@ -0,0 +1,43 @@
...
+FAIL FAIL FAIL
+#endif
+'''
+
+add_project_arguments('-DMESON_CC_CHECKS_SHOULDNT_SEE_ME__FUNC=1', language: 'c')
+
+cc = meson.get_compiler('c')
+
+define_to_check = get_option('define_to_check')
+define_expectet = get_option('define_expected')
```
eli-schwartz, line 18: Curious why this variable is misspelled everywhere, is it intentional?

Why does it need to be a get_option? It's never actually set, can't it just be initialized at the top of meson.build as true?

ePirat: Indeed, fixed typo and removed the option.

ePirat:
So after more investigation the test only failed because the `ubuntu-armhf.txt` cross file set `c_args` which causes meson to ignore the `CFLAGS` from the env. It was overwriting it as it was needed for a specific test only,, so I've patched the test runner to properly add the test-local crossfile.ini only if actually running cross-tests and made it local for the test that needs it.

This fixes my test, but uncovered a unit test that now fails, because the test is actually incorrect… It tests that CFLAGS are ignored when cross-compiling, which is not the case and not how things actually work. The test only succeeded before because the cross-file indadvertedly caused this behaviour.

PR #9867, opened 2022-01-23
fix broken version annotations for the new itstool_join feature

eli-schwartz:
It was not added in 0.61.0 as that was already released.


ximion:
LGTM :-)
Thanks for updating this!

dcbaker (approved):

PR #9871, opened 2022-01-23
Document argument flattening in Meson

mensinda:


docs/markdown/Syntax.md
```diff
@@ -527,6 +527,54 @@ executable('progname', 'prog.c',
 
 Attempting to do this causes Meson to immediately exit with an error.
 
+### Argument flattening
+
+Argument flattening is a Meson feature that aims to simplify using
+methods and functions. For functiosn where this feature is active,
```
Tachi107, line 533: "For functiosn" -> "For functions"

docs/yaml/objects/cfg_data.yaml
```diff
@@ -50,8 +50,8 @@ methods:
         need to pass numbers use the `.set` method.
   warnings:
   - numeric values < 0 have the surprising behavior of being converted to
-    [[true]], values > 1 have the more expected but unintentional behavior of
-    being interpretered as [[true]].
+    `true`, values > 1 have the more expected but unintentional behavior of
+    being interpretered as `true`.
```
Tachi107, line 54:
```suggestion
    being interpreted as `true`.
```

docs/markdown/Syntax.md
```diff
@@ -527,6 +527,54 @@ executable('progname', 'prog.c',
...
+# A more contrived example that also works but certainly
+# isn't good Meson code:
+l1 = ['bar.c']
+executable('exe1', [[['foo.c', l1]], ['foobar.c']])
+
+# How meson will treat all the previous calls internally:
+executable('exe1', 'foo.c', 'bar.c', 'foobar.c')
+```
+
+Because of an internal implementation detail, the follwing syntax
```
Tachi107, line 553:
```suggestion
Because of an internal implementation detail, the following syntax
```

docs/markdown/Syntax.md
```diff
@@ -527,6 +527,54 @@ executable('progname', 'prog.c',
...
+```meson
+# WARNING: This example is only valid because of an internal
+#          implementation detail and not because it is intended
+#
+#          PLEASE DO NOT DO SOMETHING LIKE THIS!
+#
+executable(['exe1', 'foo.c'], 'bar.c', 'foobar.c')
+```
+
+This code is currently accepted because arguement *currently*
```
Tachi107, line 566:
```suggestion
This code is currently accepted because argument *currently*
```

docs/refman/jsonschema.py
```diff
@@ -43,13 +43,14 @@ class Function(BaseObject):
...
     '''
-    returns:     T.List[Type]  # A non-empty list of types that are supported.
-    returns_str: str           # Formated version of `returns`. Is guranteed to not contain any whitespaces.
-    example:     T.Optional[str]
-    posargs:     T.Dict[str, Argument]
-    optargs:     T.Dict[str, Argument]
-    kwargs:      T.Dict[str, Argument]
-    varargs:     T.Optional[Argument]
+    returns:        T.List[Type]  # A non-empty list of types that are supported.
+    returns_str:    str           # Formated version of `returns`. Is guranteed to not contain any whitespaces.
```
Tachi107, line 47:
```suggestion
    returns_str:    str           # Formatted version of `returns`. Is guaranteed to not contain any whitespaces.
```

docs/markdown/Syntax.md
mensinda: Thanks for finding those typos. I really should have used a spellchecker before opening this PR :)

docs/markdown/Syntax.md
```diff
@@ -527,6 +527,54 @@ executable('progname', 'prog.c',
...
+# A more contrived example that also works but certainly
+# isn't good Meson code:
+l1 = ['bar.c']
+executable('exe1', [[['foo.c', l1]], ['foobar.c']])
+
+# How meson will treat all the previous calls internally:
+executable('exe1', 'foo.c', 'bar.c', 'foobar.c')
+```
+
+Because of an internal implementation detail, the following syntax
```
nirbheek, line 553: Hm, I didn't know about this implementation detail. Why are we documenting it? Shouldn't we detect it and print a warning on usage instead?

eli-schwartz: I think the problem is that the current interpreter implementation also makes it effectively very hard to know that that is happening?

docs/markdown/Syntax.md
```diff
@@ -527,6 +527,54 @@ executable('progname', 'prog.c',
...
+# WARNING: This example is only valid because of an internal
+#          implementation detail and not because it is intended
+#
+#          PLEASE DO NOT DO SOMETHING LIKE THIS!
+#
+executable(['exe1', 'foo.c'], 'bar.c', 'foobar.c')
+```
+
+This code is currently accepted because argument *currently*
+happens before the parameters are evaluated. "Support" for
```
eli-schwartz, line 567: "because argument" what?

docs/markdown/Yaml-RefMan.md
```diff
@@ -115,6 +120,9 @@ optargs_inherit: _build_target_base  # Use the optargs definition of `_build_tar
 varargs_inherit: _build_target_base  # Use the varargs definition of `_build_target_base` here
 kwargs_inherit: _build_target_base   # Add all kwargs of `_build_target_base` to this function
 
+# Wether argument flattening (see Syntax.md) is enabled
```
eli-schwartz, line 123:
```suggestion
# Whether argument flattening (see Syntax.md) is enabled
```

docs/markdown/Syntax.md
mensinda: I have already tried to detect / properly deprecate this, but it is nearly impossible with the current interpreter setup. It will be possible once we have every function (that supports argument flattening) type-checked with `@typed_pos_args` and friends. In that case, we could (and should) move the logic to there.

My point for documenting this is that we can't really deprecate it because we currently can't feasibly check it. So being transparent about this fact (and telling people that this code *will* break in the future) in the docs is the best solution I could come up with.

nirbheek: I don't see the point of documenting it in such detail either way. We don't generally put bugs / unintentional behaviour in documentation unless users are very likely to encounter it. What am I missing?

mensinda: IMHO, it is not unreasonable to unintentionally run into this if you do something like this:

```meson
src = ['main', 'main.cpp']

executable(src)
```

So my point is that this is the best solution I could currently come up with to warn users **not** to do this.

mensinda:
ping :)

eli-schwartz:
Shouldn't the last commit's changes in docs/markdown/Syntax.md be folded into the earlier commit where they originally got introduced? The remaining typos are pre-existing, but could use a better commit message IMHO.

eli-schwartz:
I'm not entirely sure what to think about making every single function declare that args flattening is or isn't supported, especially since it requires diving through links to find the information that maybe you shouldn't be relying on it at all.

mensinda:
IMHO it should be documented somewhere which functions support argument flattening and which don't. Otherwise, it can be pretty confusing. Maybe we should just add the note when argument flattening is not supported?

eli-schwartz (changes requested):
> Shouldn't the last commit's changes in docs/markdown/Syntax.md be folded into the earlier commit where they originally got introduced? The remaining typos are pre-existing, but could use a better commit message IMHO.

I would still like this change to be made.

mensinda:
> > Shouldn't the last commit's changes in docs/markdown/Syntax.md be folded into the earlier commit where they originally got introduced? The remaining typos are pre-existing, but could use a better commit message IMHO.
> 
> I would still like this change to be made.

Done

eli-schwartz (approved):

PR #9876, opened 2022-01-25
compilers/c_function_attributes: add retain + typo correction

ArsenArsen:
retain is a relatively young attribute which has proven itself useful
for working with --gc-sections -z start-stop-gc.

### DIFF
--- a/docs/markdown/Reference-tables.md
+++ b/docs/markdown/Reference-tables.md
@@ -240,6 +240,7 @@ which are supported by GCC, Clang, and other compilers.
 | optimize                 |
 | packed                   |
 | pure                     |
+| retain⁴                  |
 | returns_nonnull          |
 | unused                   |
 | used                     |
@@ -260,6 +261,8 @@ which are supported by GCC, Clang, and other compilers.
 "visibility" as they provide narrower checks.
 
 ³ *New in 0.55.0*
+
+⁴ *New in 0.62.0*
 
 ### MSVC __declspec
 
--- a/docs/markdown/Reference-tables.md
+++ b/docs/markdown/Reference-tables.md
@@ -252,7 +252,7 @@ which are supported by GCC, Clang, and other compilers.
 | warning                  |
 | warn_unused_result       |
 | weak                     |
-| weakreaf                 |
+| weakref                  |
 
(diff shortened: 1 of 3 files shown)


docs/markdown/Reference-tables.md
```diff
@@ -251,7 +251,8 @@ which are supported by GCC, Clang, and other compilers.
 | warning                  |
 | warn_unused_result       |
 | weak                     |
-| weakreaf                 |
+| weakref                  |
+| retain                   |
```
dcbaker, line 255: We should mark this as new in 0.62.0, I think there's other examples of this. I think this table is alphabetically sorted as well.

dcbaker (changes requested):
I've got a few doc nits, otherwise this looks good. Thanks for taking the time to add this!

docs/markdown/Reference-tables.md
ArsenArsen: Changes made. Let me know if it's up-to-spec and I'll rebase.

dcbaker (approved):

docs/markdown/Reference-tables.md
dcbaker: looks good. I can just squash when I commit, so don't worry about rebasing (saves CI time)

ArsenArsen: Sure thing.

test cases/common/197 function attributes/meson.build
```diff
@@ -53,6 +53,7 @@ attributes = [
   'used',
   'warn_unused_result',
   'weak',
+  'retain',
```
eli-schwartz, line 56: Unconditional check is probably a bad idea, see e.g. later on in the file we do:

```meson
  if c.get_id() == 'gcc' and c.version().version_compare('>= 7.0.0')
    attributes += 'fallthrough'
  endif
```

CI fails because this attribute is lacking and triggers an assert on too-old GCC.

eli-schwartz (changes requested):

test cases/common/197 function attributes/meson.build
ArsenArsen: Scoped to clang 13+, I'll add a comment about why when rebasing

PR #9877, opened 2022-01-26
add location data to various Feature checks

eli-schwartz:
Incremental progress on getting all Feature warnings to tell the user where the warning came from.

### DIFF
--- a/mesonbuild/interpreter/interpreter.py
+++ b/mesonbuild/interpreter/interpreter.py
@@ -546,7 +546,7 @@ class Interpreter(InterpreterBase, HoldableObject):
                 except KeyError:
                     continue
                 if len(di) == 1:
-                    FeatureNew.single_use('stdlib without variable name', '0.56.0', self.subproject)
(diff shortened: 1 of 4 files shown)


mesonbuild/interpreter/interpreter.py
```diff
@@ -1331,7 +1331,7 @@ def add_languages_for(self, args: T.List[str], required: bool, for_machine: Mach
         # compilers we don't add anything for cython here, and instead do it
         # When the first cython target using a particular language is used.
         if 'vala' in langs and 'c' not in langs:
-            FeatureNew.single_use('Adding Vala language without C', '0.59.0', self.subproject)
+            FeatureNew.single_use('Adding Vala language without C', '0.59.0', self.subproject, location=node)
```
dcbaker, line 1334: I don't see where the `node` is coming from here, or in the next use?

eli-schwartz: This one was broken, that particular function doesn't have a node kwarg -- the next use does though...

eli-schwartz:
```
10:13 PM <@dcbaker> I’m on my phone so I can’t give an official r-b, but yes
```

PR #9879, opened 2022-01-26
Rename JDK system dep to JNI

tristan957:
JNI is a more apt name because it currently only supports the JNI. I
also believe that CMake uses the terminology JNI here as well.

JNI is currently the only way to interact with the JVM through native
code, but there is a project called "Project Panama" which aims to be
another way for native code to interact with the JVM.


tristan957:
Potentially we could keep `jdk`, and depending on how Project Panama works, we could add support for the `method` kwarg in dependency if we didn't feel like breaking backwards compat.

I feel like I am probably the only consumer of this, so I don't really care if it is broken imo. Happy to hear other suggestions though. I also don't mind breaking compat because this was actually never documented to begin with. See https://github.com/mesonbuild/meson/pull/9878

tristan957:
cc @dcbaker

dcbaker:
The code looks fine to me, but I'm concerned about the rename. Can we alias it and deprecated the JDK dependency, even if just for a couple of versions? We've run into a few cases where we don't think anyone is using a feature, only to find out that *someone* is, in fact, using that feature.

tristan957:
Yea I can do that. Works for me :). I will update the pr

eli-schwartz:
Dependencies don't get a handle on the interpreter and cannot process Feature checks. Same issue as 39f61795ca1ae7de3c4c1fc08f88b3d864502cec.

tristan957:
I see. That is unfortunate. I will remove the Feature checks then here.

eli-schwartz:
Actually I have an idea for fixing all of these...

tristan957:
Exciting. Ping me when you have something.

eli-schwartz:
My PR got merged, and meanwhile this conflicts.

tristan957:
working on it...

tristan957:
I took an opportunity to fixup the docs to make it more clear how the lookup works and a common pitfall which may occur that I myself have run into.

mesonbuild/dependencies/dev.py
```diff
@@ -549,6 +552,18 @@ def __machine_info_to_platform_include_dir(m: 'MachineInfo') -> T.Optional[str]:
         return None
 
 
+class JDKSystemDependency(JNISystemDependency):
+    def __init__(self, environment: 'Environment', kwargs: T.Dict[str, T.Any]):
+        super().__init__(environment, kwargs)
```
eli-schwartz, line 557: This probably results in the jdk dependency having a `dep.name()` of `jni` even when you look it up with `jdk`.

This is... probably fine? It is a deprecated alias and all that.

eli-schwartz (approved):

mesonbuild/dependencies/dev.py
tristan957: I figured it would probably be a little painful in that regard, but since it is deprecated, I don't think anyone will Notice. I could do a refactor however and move the actual lookups to a common function instead.

eli-schwartz: I agree that it may be entirely too painful to bother for a deprecated thing.

tristan957:
Think I am going to adjust the documentation one more time. I didn't denote the usage of `JAVA_HOME` in the lookup.

eli-schwartz:
CI passed before this last push, which only changed the documentation text. So the rest of the CI can be skipped.

eli-schwartz:
> Update website / update_website

succeeded

PR #9880, opened 2022-01-26
fix a bunch of flake8 issues

eli-schwartz:


tristan957:
Is flake8 a part of the CI currently?

eli-schwartz:
No, it is not.

Probably mostly due to the fact that there are around 300 remaining flake8 issues, and this PR only fixes 60 of them. So if we added flake8 to CI, it would mean master was always red.

I hope to add flake8 to CI soon... this PR is part of that.

tristan957:
Cool. Thanks for your work.

dcbaker (approved):

PR #9881, opened 2022-01-26
Deprecate java.generate_native_header() in favor of java.generate_native_headers()

tristan957:
After implementing a much more extensive Java native module than what
currently exists in the tests, I found shortcomings.

1. You need to be able to pass multiple Java files.
2. Meson needs more information to better track the generated native
   headers.
3. Meson wasn't tracking the header files generated from inner classes.

This new function should fix all the issues the old function had with
room to grow should more functionality need to be added. What I
implemented here in this new function is essentially what I have done in
the Heterogeneous-Memory Storage Engine's Java bindings.

### DIFF
--- a/docs/markdown/Java-module.md
+++ b/docs/markdown/Java-module.md
@@ -1,10 +1,12 @@
 # Java Module
 
-*Added 0.60.0*
+*(added in 0.60.0)*
 
 ## Functions
 
 ### `generate_native_header()`
+
+*(deprecated in 0.62.0, use `generate_native_headers()`)*
 
 This function will generate a header file for use in Java native module
 development by reading the supplied Java file for `native` method declarations.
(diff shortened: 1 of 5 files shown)


tristan957:
CC @dcbaker

dcbaker:
This looks okay to me, but you need to rebase because the `CustomTarget` interface has changed.

tristan957:
Cool I will check that out. Thank! Maybe will look into it tonight.

tristan957:
Did you have any suggestions on what to use for the name of the custom-target? `native-headers` is pretty generic.

dcbaker:
Hmmm, not really? Naming is hard. It just needs to be unique enough that one could call the function twice and not get the same target name.

tristan957:
Updated. It isn't clear how unique a custom target name has to be.

dcbaker (approved):

tristan957:
Just pushed a fix for mypy.

PR #9884, opened 2022-01-26
gnome: Fix gtk4 pkgconfig name

bilelmoussaoui:
Otherwise it shows up during my builds with
```
Build-time dependency gtk-4.0 found: NO (tried pkgconfig and cmake)
```
which is extremely confusing


eli-schwartz:
Huh, https://github.com/GNOME/gtk/commit/d080be3e5091c98d5171063a95d55c01170881f3 renamed it from `gtk+-4.0` to `gtk4`, meanwhile we were missing the plus (which the commit message mentions removing) for the former while having an extraneous dash and `.0` for the latter (which the commit message does not mention removing).

So, this would always have been wrong...

eli-schwartz (approved):

PR #9887, opened 2022-01-27
remove incorrect deprecated feature for vcs_tag

eli-schwartz:
In commit 06481666f4e74ecef01e59351fc345ab0962d998 this warning got moved from build.py to the interpreter. Unfortunately it got added to the wrong function... it is supposed to be part of custom_target and even mentions this as the feature_name.

Since then, build_always became a KwargInfo and has the deprecated-since attribute baked into it. But it didn't have the additional message which it really should have.

Add that message at the same time we remove it from vcs_tag.


dcbaker (approved):

eli-schwartz:
No-change rebase on master in the hope of getting cygwin to actually run instead of hanging.

dcbaker:
Strange, I kicked it and it was running tests, I was watching it run

eli-schwartz:
Okay, maybe I just didn't notice lol.

PR #9890, opened 2022-01-27
do not report context on python traceback, for PermissionError

eli-schwartz:
It's not a MesonBug which needs to be reported, and the existing error already adequately points out the problematic file.

It is impossible to get a PermissionError for files created by meson itself, once the build directory has been created, anyway.


dcbaker (approved):
I saw the conversation about this on IRC, and this seems like a reasonable solution to me

Issue #9883, opened 2022-01-26
Errors in `gdbus_codegen` annotations support


inigomartinez:
**Describe the bug**
There are changes in `gdbus_codegen` that make the build to fail because it now doesn't accept empty lists as possible input.

**To Reproduce**
This happens with [ModemManager](https://gitlab.freedesktop.org/mobile-broadband/ModemManager/-/commit/86496bd326f2ec139101049fb19a10b255ee8ad5). An example would be as follows:
```
  gnome.gdbus_codegen(
    'example',
    sources: 'example.c',
    interface_prefix: 'org.example.',
    namespace: 'Example',
    annotations: [],
    object_manager: false,
    autocleanup: 'objects',
    build_by_default: true,
  )
```

**Expected behavior**
The annotation parameter should be ignored if empty list is provided.

**system parameters**
* Is this a [cross build](https://mesonbuild.com/Cross-compilation.html) or just a plain native build (for the same computer)?
* what operating system (e.g. MacOS Catalina, Windows 10, CentOS 8.0, Ubuntu 18.04, etc.)
* what Python version are you using e.g. 3.8.0
* what `meson --version`
* what `ninja --version` if it's a Ninja build
This issue starts at meson 0.61.

eli-schwartz:
Why should this be categorized as a regression instead of "better input sanity-checking"?

inigomartinez:
I have categorized it as a regression because it made [already working builds to fail](https://gitlab.freedesktop.org/mobile-broadband/ModemManager/-/issues/499#note_1232380).

However, I can change the description if you think it fits better.

eli-schwartz:
"made already working builds to fail" is not the definition of a regression, though.

A regression is when already working builds ***that could reasonably be expected to work*** start failing.

More specifically, a regression is a subclass of a bug. If something is not a bug, then it doesn't matter whether previously working builds now fail -- it becomes the responsibility of the people with failing builds, to fix *their* definitions.

For example, when i18n.merge_file started erroring out with "function does not take positional arguments", that wasn't a regression, even though it caused already working builds to fail, because the builds were always wrong. That's why meson 0.61.0 will not back out the change, even though we did for 0.60 and very few people actually took advantage of the deprecation warning, because those builds deserve to be broken and we aren't carrying a long-term workaround.

So again I ask my original question: why did you need to pass an empty list, when semantically it is supposed to accept a list of 3 elements. e.g. can't you just... Not specify the kwarg at all?

[...]

PR #9893, opened 2022-01-27
modules/gnome: Allow gdbus_annotations to take an empty list again

dcbaker:
This was allows up to 0.61.0 (including with the initial type
annotations), but was accidentally broken by fixes for other bugs in
0.61.1.

Fixes: #9883

### DIFF
--- a/mesonbuild/modules/gnome.py
+++ b/mesonbuild/modules/gnome.py
@@ -224,12 +224,13 @@ _MK_ENUMS_COMMON_KWS: T.List[KwargInfo] = [
 ]
 
 def annotations_validator(annotations: T.List[T.Union[str, T.List[str]]]) -> T.Optional[str]:
-
     """Validate gdbus-codegen annotations argument"""
 
     badlist = 'must be made up of 3 strings for ELEMENT, KEY, and VALUE'
 
-    if all(isinstance(annot, str) for annot in annotations):
+    if not annotations:
+        return None
+    elif all(isinstance(annot, str) for annot in annotations):
         if len(annotations) == 3:
             return None
         else:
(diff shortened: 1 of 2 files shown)


mesonbuild/modules/gnome.py
```diff
@@ -224,12 +224,13 @@ class MkEnums(_MkEnumsCommon):
...
 def annotations_validator(annotations: T.List[T.Union[str, T.List[str]]]) -> T.Optional[str]:
-
     """Validate gdbus-codegen annotations argument"""
 
     badlist = 'must be made up of 3 strings for ELEMENT, KEY, and VALUE'
 
-    if all(isinstance(annot, str) for annot in annotations):
+    if not annotations:
+        return None
+    elif all(isinstance(annot, str) for annot in annotations):
```
eli-schwartz, line 233: This line doesn't need to be modified into an elif (instead of `-1 +3` it could just be `+2`) because it should always run, except when we returned early.

eli-schwartz (approved):
This seems fine to me, I guess... it is once again meh that we need to code special support for "placeholder that represents not passing an argument in a foreach loop" but I guess there isn't currently another choice...

mesonbuild/modules/gnome.py
dcbaker: I'm really not suer what the concern here is? I'm adding another condition to a series of if/elif conditions, and it happens to be the first one so the second becomes an elif, this seems straightforward to me.

dcbaker:
I have a branch that adds a `default()` object, but to make it work cleanly and reliably we need to use typed_* everywhere, so there's still some work to do

mesonbuild/modules/gnome.py
eli-schwartz: There's no actual concern, just a random comment. It doesn't make a difference, but stylistically I would have done it different.

PR #9896, opened 2022-01-28
ninja backend: Fix usage of same constants file for native and cross

nirbheek:
For example:
```
meson builddir \
        --native-file vs2019-paths.txt \
        --native-file vs2019-win-x64.txt \
        --cross-file vs2019-paths.txt \
        --cross-file vs2019-win-arm64.txt
```
This was causing the error:

> ERROR: Multiple producers for Ninja target "/path/to/vs2019-paths.txt". Please rename your targets.

Fix it by using a set() when generating the list of regen files, and add a test for it too.

Also fix a warning about distutils deprecation.

### DIFF
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -1191,14 +1191,14 @@ class Backend:
     def get_regen_filelist(self) -> T.List[str]:
         '''List of all files whose alteration means that the build
         definition needs to be regenerated.'''
-        deps = [str(Path(self.build_to_src) / df)
-                for df in self.interpreter.get_build_def_files()]
+        deps = OrderedSet([str(Path(self.build_to_src) / df)
+                for df in self.interpreter.get_build_def_files()])
         if self.environment.is_cross_build():
-            deps.extend(self.environment.coredata.cross_files)
-        deps.extend(self.environment.coredata.config_files)
-        deps.append('meson-private/coredata.dat')
+            deps.update(self.environment.coredata.cross_files)
+        deps.update(self.environment.coredata.config_files)
+        deps.add('meson-private/coredata.dat')
         self.check_clock_skew(deps)
-        return deps
+        return list(deps)
 
     def generate_regen_info(self) -> None:
         deps = self.get_regen_filelist()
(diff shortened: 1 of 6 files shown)


mesonbuild/backend/backends.py
```diff
@@ -1191,14 +1191,14 @@ def generate_depmf_install(self, d: InstallData) -> None:
...
         if self.environment.is_cross_build():
-            deps.extend(self.environment.coredata.cross_files)
-        deps.extend(self.environment.coredata.config_files)
-        deps.append('meson-private/coredata.dat')
+            deps.update(self.environment.coredata.cross_files)
+        deps.update(self.environment.coredata.config_files)
+        deps.add('meson-private/coredata.dat')
         self.check_clock_skew(deps)
-        return deps
+        return list(deps)
```
dcbaker, line 1201: do we need to `sort()` here (or use an `OrderedSet`) to get reproducible ninja output?

unittests/rewritetests.py
```diff
@@ -26,7 +26,9 @@ def setUp(self):
         self.maxDiff = None
 
     def prime(self, dirname):
-        copy_tree(os.path.join(self.rewrite_test_dir, dirname), self.builddir)
+        if os.path.exists(self.builddir):
+            shutil.rmtree(self.builddir)
```
dcbaker, line 30: We probably should use `mesonlib.windows_proof_rm` here, right?

dcbaker (changes requested):
A couple of small nits, otherwise this looks good to me.

mesonbuild/backend/backends.py
eli-schwartz: Do we generally need to have byte-identical `build.ninja` outputs? Do you mean something else?

nirbheek: Yeah you're right, `OrderedSet` is good here. I somehow thought sets worked like dicts now and were iterated in insertion order.

nirbheek: > Do we generally need to have byte-identical build.ninja outputs? Do you mean something else?

No, but if the same user regenerates build.ninja again with the same parameters and environment, the contents should be identical. Unordered iterables will break that and can cause an unnecessary regen / rebuild.

eli-schwartz: Ah, makes sense. :)

PR #9897, opened 2022-01-28
Type checking for configure file (plus some cleanups)

dcbaker:
This is currently based on #9853.

This adds type checking for the `configure_file` interface. The helpers in `typed_kwargs` allow getting rid of a lot of hand rolled stuff, especially the use of the `in_set_validator`. This value narrowing allows us to use tighter constraints in the mesonlib module, which allows removing asserts, because mypy has already allowed us to guarantee much of what those asserts do.


mesonbuild/interpreter/interpreter.py
```diff
@@ -2211,67 +2208,68 @@ def func_install_subdir(self, node: mparser.BaseNode, args: T.Tuple[str],
...
+        'configure_file',
+        KwargInfo('output', str, required=True),
+        KwargInfo('capture', bool, default=False, since='0.41.0'),
+        KwargInfo('format', str, default='meson', since='0.46.0',
+                  validator=in_set_validator({'meson', 'cmake', 'cmake@'})),
+        KwargInfo('output_format', str, default='c', since='0.47.0',
+                  validator=in_set_validator({'c', 'nasm'})),
+        DEPFILE_KW.evolve(since='0.52.0'),
+        # Cannot use shared implementation until None backwards compat is dropped
+        KwargInfo('install', (bool, NoneType), since='0.50.0'),
```
tristan957, line 2222: Seems like install could become a constant KwargInfo like DEPFILE_KW above,

tristan957: Same with install_dir

eli-schwartz: There is a comment describing why we cannot.

tristan957: I am blind

dcbaker: There is a const version, but it only allows a bool, and configure_file has a legacy behavior when `install` isn't passed.

mesonbuild/interpreter/interpreter.py
```diff
@@ -2212,67 +2213,68 @@ def func_install_subdir(self, node: mparser.BaseNode, args: T.Tuple[str],
...
+        ),
+    )
+    def func_configure_file(self, node: mparser.BaseNode, args: T.List[TYPE_var],
+                            kwargs: kwargs.ConfigureFile):
+        actions = sorted(x for x in {'configuration', 'command', 'copy'} if bool(kwargs[x]))
+        num_actions = len(actions)
+        if num_actions == 0:
             raise InterpreterException('Must specify an action with one of these '
                                        'keyword arguments: \'configuration\', '
                                        '\'command\', or \'copy\'.')
```
eli-schwartz, line 2254: `configuration: {}` triggers this error, which is confusing.

mesonbuild/interpreter/interpreter.py
```diff
@@ -2212,67 +2213,68 @@ def func_install_subdir(self, node: mparser.BaseNode, args: T.Tuple[str],
...
+            'input',
+            ContainerTypeInfo(list, (mesonlib.File, str)),
+            listify=True,
+            default=[],
+        ),
+        KwargInfo(
+            'configuration',
+            (ContainerTypeInfo(dict, (str, int, bool)), build.ConfigurationData),
+            default={},
+        ),
```
eli-schwartz, line 2245: Alright, if we want to error out on empty configurations this should really have a validator that rejects it directly rather than just letting it evaluate to `bool(False)`. Fun fact: `configuration: configuration_data()` has different truthy status than `{}`.

mesonbuild/interpreter/interpreter.py
```diff
@@ -2340,7 +2340,7 @@ def func_configure_file(self, node, args, kwargs):
             else:
                 mesonlib.dump_conf_header(ofile_abs, conf, output_format)
             conf.used = True
-        elif 'command' in kwargs:
+        elif kwargs['command']:
```
eli-schwartz, line 2338: And here we also don't know whether the user specified literally `command: []` but if they do, we apparently report it as "must specify one of these 3, no action provided".

eli-schwartz (commented):
Mostly LGTM, but I'm a bit concerned about how we determine which mode to use.

Also the ordering of the KwargInfos seems odd. :D

mesonbuild/interpreter/interpreter.py
dcbaker: That's easy to fix :)

mesonbuild/interpreter/interpreter.py
eli-schwartz: We can now change this too to `is not None`, like we did for configuration above.

mesonbuild/interpreter/interpreter.py
```diff
@@ -2215,67 +2214,68 @@ def func_install_subdir(self, node: mparser.BaseNode, args: T.Tuple[str],
...
+                  validator=lambda x: 'must be `false` if boolean' if x is True else None),
+        KwargInfo('output', str, required=True),
+        KwargInfo('output_format', str, default='c', since='0.47.0',
+                  validator=in_set_validator({'c', 'nasm'})),
+    )
+    def func_configure_file(self, node: mparser.BaseNode, args: T.List[TYPE_var],
+                            kwargs: kwargs.ConfigureFile):
+        actions = sorted(x for x in {'configuration', 'command'} if kwargs[x] is not None)
+        if kwargs['copy']:
+            actions.append('copy')
```
eli-schwartz, line 2255: Hmm, perhaps `if kwargs[x] not in {None, False}`?

mesonbuild/interpreter/interpreter.py
eli-schwartz: Indeed, seems reasonable. :)

mesonbuild/interpreter/interpreter.py
dcbaker: I guess yeah, we could do that. I don't really know why we'd care to check `is None` in the `elif` case above, we can never get a falsy value that isn't None since we don't allow `command` to be empty

eli-schwartz (approved):

PR #9898, opened 2022-01-28
ninjabackend: treat link_whole_targets like link_targets for Rust targets

zachreizner:
For static library crates that depend on other internal static library
crates, all link_with targets get promoted to link_whole targets. Due to
a bug, only link_with targets are considered when generating a Rust
target for the ninja backend. This made it impossible to link a Rust
static library with another internal Rust static library.

This change fixes that issue by chaining link_whole_targets with
link_targets, just like many other languages within the ninja backend.

### DIFF
--- a/test cases/rust/17 staticlib link staticlib/meson.build
+++ b/test cases/rust/17 staticlib link staticlib/meson.build
@@ -1,0 +1,8 @@
+project('staticlib link staticlib', 'c', 'rust')
+
+leaf = static_library('leaf', 'leaf.rs', rust_crate_type : 'rlib')
+# Even though leaf is linked using link_with, this gets implicitly promoted to link_whole because
+# it is an internal Rust project.
+branch = static_library('branch', 'branch.rs', link_with: leaf, rust_crate_type : 'staticlib', install : true)
+e = executable('prog', 'prog.c', link_with : branch, install : true)
+test('linktest', e)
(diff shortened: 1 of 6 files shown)


dcbaker:
The change looks correct. It would be nice to have a test for this though

zachreizner:
I have added a test case. Please take a look.

dcbaker (approved):
Looks correct to me. I think we need To kick the azure ci again, those failures look spurious to me

test cases/rust/17 staticlib link staticlib/meson.build
```diff
@@ -0,0 +1,6 @@
+project('staticlib link staticlib', 'c', 'rust')
+
+leaf = static_library('leaf', 'leaf.rs', rust_crate_type : 'rlib')
+branch = static_library('branch', 'branch.rs', link_with: leaf, rust_crate_type : 'staticlib', install : true)
+e = executable('prog', 'prog.c', link_with : branch, install : true)
```
jpakkane, line 5: The description talks of link_whole but neither of these two is defined as link_whole. Is that intentional? If yes, there should be comment explaining what is happening here.

zachreizner: It was intentional. I've added a comment explaining what is going on there. (For those who might be reading this issue PR later, the link_with gets promoted to link_whole, which wasn't handled properly in the ninja backend for Rust targets).

Issue #9250, opened 2021-09-13
[d] empty include directory when using include_directories('.') in meson.source_root()


Panke:
**Describe the bug**
In my D project I need to always have the root source dir as an include-directory.
Thus I did
```
 project_dep = declare_dependency(
    include_directories: ['.'],
)
```

and use it like this: 
```
bix_lib = library(
	'bix',
	['bix.d'],
    dependencies: [ project_dep]
)
```

But I get the following error using ldc:
```
ldc2 -I=ex/mod/libbx.so.p -I=ex/mod -I=../ex/mod -I= -I=.. -I=subprojects -I=../subprojects -enable-color -wi -g -d-debug -relocation-model=pic -makedeps=ex/libbx.so.p/bx.d.o.deps -of=ex/libbx.so.p/bx.d.o -c ../ex/bx.d
Error: Expected argument to '-I'
```

**To Reproduce**
Attached a tar ball reproducing the issue for me.
[meson_issue.tar.gz](https://github.com/mesonbuild/meson/files/7153466/meson_issue.tar.gz)
**Expected behavior**
-I=. instead of -I=

**system parameters**
* Is this a [cross build](https://mesonbuild.com/Cross-compilation.html) or just a plain native build (for the same computer)?
plain native
* what operating system (e.g. MacOS Catalina, Windows 10, CentOS 8.0, Ubuntu 18.04, etc.)
 linux (arch)
* what Python version are you using e.g. 3.8.0
python 3.9.6
* what `meson --version`
 meson --version
0.58.1

PR #9899, opened 2022-01-29
dlang: add test case for root include_directory #9250

Panke:


Panke:
This PR shows to possible fixes to #9250: one changes the D compiler infrastructure to change '-I=' to '-I=.', the other one tries to tackle it more at the root by avoiding to replace '.' with '' before passing it to get_include_args in the first place.

Panke:
The DLL entry point was missng. I added one know, let's see if it fixes the issues on windows.

dcbaker (approved):

PR #9903, opened 2022-01-29
meson: Allow directory options outside of prefix

jtojnar:
This bring us in line with Autotools and CMake and it is useful
for platforms like Nix, which install projects
into multiple independent prefixes.

As a consequence, `get_option` might return absolute paths for some
directory options, if a directory outside of prefix is passed.

This is technically a backwards incompatible change but its effect
should be minimal, thanks to widespread use of `join_paths`/`/` operator
and pkg-config generator module. It should only cause an issue when
a path were constructed by concatenating the value of directory path option.

Fixes: https://github.com/mesonbuild/meson/issues/2561


docs/markdown/snippets/dir_options_outside_prefix.md
```diff
@@ -0,0 +1,9 @@
+It is now allowed to set directory options to paths outside of prefix.
+This bring us in line with Autotools and CMake and it is useful for
+platforms like Nix, which install projects into multiple independent prefixes.
+
+As a consequence, `get_option` might return absolute paths for some
+directory options, if a directory outside of prefix is passed. This
+is technically a backwards incompatible change but its effect
```
eli-schwartz, line 7: Arguably it is not backwards-incompatible, because it depends on the user configuration, right? And so Meson actually does the same thing in all cases, with the same return values of `get_option()`, ***except*** in cases where currently it would report an error and refuse to continue.

docs/yaml/functions/_build_target_base.yaml
```diff
@@ -161,10 +161,10 @@ kwargs:
...
     type: str
     description: |
-      override install directory for this file. The value is
-      relative to the `prefix` specified. F.ex, if you want to install
-      plugins into a subdir, you'd use something like this: `install_dir :
-      get_option('libdir') / 'projectname-1.0'`.
+      override install directory for this file. If the value is a relative path,
+      it will be joined to the value of the `prefix` option.
+      F.ex, if you want to install plugins into a subdir, you'd use something
+      like this: `install_dir : get_option('libdir') / 'projectname-1.0'`.
```
eli-schwartz, line 167: This change isn't strictly necessary in the context of this PR, although an argument could be made that the rewording is useful regardless? idk

Point being, `install_dir` currently takes arbitrary locations, including the return value of `get_option('sysconfdir')` (typically the absolute value `/etc`) and also random hardcoded strings (why one would write a meson.build like that I do not know...)

So the existing wording "the value is relative to the prefix" applies to `/etc` being relative to prefix, but since it is an absolute value it cannot actually be made relative... and now more paths are *like* sysconfdir.

Eh, I guess the wording change cannot hurt even just to clarify the /etc case.

Maybe instead:
> If the value is a relative path, it is relative to the `prefix` specified.

mesonbuild/coredata.py
```diff
@@ -576,7 +575,8 @@ def sanitize_dir_option_value(self, prefix: str, option: OptionKey, value: T.Any
             try:
                 value = value.relative_to(prefix)
             except ValueError:
-                raise MesonException(msg.format(option, value, prefix))
+                # Path is not relative, let’s return it as is.
+                pass
```
eli-schwartz, line 579: We still need `msg` for the case where `..` is present. However, it should be reworded to be more specific, since now if `..` is present in an option, Meson errors out by telling you... that anything outside of the prefix is not permitted, when that is incorrect, you just need to do it without a very specific substring designed to cause trouble.

eli-schwartz (changes requested):

docs/markdown/snippets/dir_options_outside_prefix.md
jtojnar: Yes, I meant incompatibility in a sense that previously correct Meson build scripts may become incorrect because of the changed assumptions.

docs/yaml/functions/_build_target_base.yaml
jtojnar: Yeah, I felt that “the value is relative” was wrong and now it was wrong in even more cases so I decided to improve the wording but I am not completely happy with the new wording.

I do not like “is relative” in your suggestion for the same reason. Maybe something like, “it will be considered relative”?

mesonbuild/coredata.py
jtojnar: I wonder if we should even care about `..` now that we relaxed the “everything is inside prefix”. It appears to originate from https://github.com/mesonbuild/meson/pull/6398 without much context. @scivision, do you remember if it was just to prevent escaping prefix or were there another purpose.

docs/yaml/functions/_build_target_base.yaml
```diff
@@ -161,10 +161,10 @@ kwargs:
   install_dir:
     type: str
     description: |
-      override install directory for this file. The value is
-      relative to the `prefix` specified. F.ex, if you want to install
-      plugins into a subdir, you'd use something like this: `install_dir :
-      get_option('libdir') / 'projectname-1.0'`.
+      override install directory for this file. If the value is a relative path,
+      it will be considered relative the `prefix` option.
+      F.ex, if you want to install plugins into a subdir, you'd use something
```
bonzini, line 166: Spell out "For example", please (the short form _e.g._ doesn't apply here).

mesonbuild/coredata.py
bonzini: I think going outside the prefix with ".." makes little sense. It should be either an absolute path or a path inside the prefix.

bonzini (changes requested):
I think this feature is a good idea. I left a couple style comments as well as my opinion on the handling of `..`. Thanks!

mesonbuild/coredata.py
jtojnar: I do not see an issue with `..` per-se – the produced denormalized paths might be somewhat unsightly but there are countless other ways to introduce de-normalization. Nevertheless, we do not really need `..` so I re-introduced the error message (with more specific wording).

docs/yaml/functions/get_option.yaml
```diff
@@ -5,16 +5,14 @@ description: |
...
-  [`install_dir` arguments](Installing.md) handles that as expected, but
-  if you need the absolute path to one of these e.g. to use in a define
-  etc., you should use `get_option('prefix') /
-  get_option('localstatedir')`
+  such as `bindir` and `libdir` is usually a path relative to (and
+  inside) the `prefix` but you should not rely on that, as it can also
+  be an absolute path [in some cases](Builtin-options.md#Universal options).
+  [`install_dir` arguments](Installing.md) handle that as expected
+  but if you need an absolute path, e.g. to use in a define etc.,
+  you should use path concatenation operator like this:
```
eli-schwartz, line 13:
```suggestion
  you should use the path concatenation operator like this:
```

eli-schwartz (changes requested):
I have one minor grammar nit, but other than that this looks good.

docs/yaml/functions/get_option.yaml
jtojnar: Done, thanks.

Ericson2314:
Yay! All of us Nixpkg users of Meson are very happy and grateful this could go through. Thanks!!

gbeatty:
Us Windows users are happy too. Our dependencies are not always in a convenient location either.

Tachi107:
How can we now reliably embed install paths in e.g. configuration files?

For example, I used to fill in the bindir path in systemd unit files with `configure_file({'bindir': get_option('prefix')/get_option('bindir')}` (see [here](https://github.com/Tachi107/cloudflare-ddns/blob/fd8dbaae51b4fe270bc8be7fa3f6e93511700c59/src/meson.build#L45-L51)), but this could now lead to incorrect paths, right?

jtojnar:
@Tachi107  The `join_paths` function and `/` operator will behave correctly here – when `bindir` contains an absolute path, the `prefix` will be ignored. The issue only occurs when joining them as a string (e.g. <del>`'@0@/@1@'.format(get_option('prefix'), get_option('bindir'))`</del>).

Tachi107:
Oh ok, thanks for the clarification :D

PR #9906, opened 2022-01-30
CMake: Remove (<3.14 for subrpojects) and deprecate (<3.17 globally) support

mensinda:
See #7832.


mesonbuild/cmake/traceparser.py
```diff
@@ -138,6 +138,17 @@ def __init__(self, cmake_version: str, build_dir: Path, permissive: bool = True)
             'meson_ps_disabled_function': self._meson_ps_disabled_function,
         }  # type: T.Dict[str, T.Callable[[CMakeTraceLine], None]]
 
+        if version_compare(self.cmake_version, '<3.17.0'):
+            mlog.deprecation(textwrap.dedent(f'''\
+                CMake support for versions <3.17 is deprecated since Meson 0.62.0.
```
eli-schwartz, line 143: Why is this moving to the traceparser from its previous location?

mensinda: Because this also affects CMake dependencies, not only CMake subprojects and the traceparser is used in both code paths.

mensinda: Ideally, this should have been in the traceparser from the start, but I didn't think that far ahead.

PR #9908, opened 2022-01-30
pkgconfig: set libdir only if used

Tachi107:
This should address some concerns outlined in https://github.com/yhirose/cpp-httplib/pull/1182

### DIFF
--- a/mesonbuild/modules/pkgconfig.py
+++ b/mesonbuild/modules/pkgconfig.py
@@ -351,7 +351,8 @@ class PkgConfigModule(ExtensionModule):
                 ofile.write('prefix={}\n'.format(self._escape(prefix)))
                 if uninstalled:
                     ofile.write('srcdir={}\n'.format(self._escape(srcdir)))
-                ofile.write('libdir={}\n'.format(self._escape('${prefix}' / libdir)))
+                if deps.pub_libs or deps.priv_libs:
+                    ofile.write('libdir={}\n'.format(self._escape('${prefix}' / libdir)))
                 ofile.write('includedir={}\n'.format(self._escape('${prefix}' / incdir)))
             if variables or unescaped_variables:
                 ofile.write('\n')


mesonbuild/modules/pkgconfig.py
```diff
@@ -351,7 +351,8 @@ def _generate_pkgconfig_file(self, state, deps, subdirs, name, description,
                 ofile.write('prefix={}\n'.format(self._escape(prefix)))
                 if uninstalled:
                     ofile.write('srcdir={}\n'.format(self._escape(srcdir)))
-                ofile.write('libdir={}\n'.format(self._escape('${prefix}' / libdir)))
+                if len(deps.pub_libs) > 0 or len(deps.priv_libs) > 0:
```
eli-schwartz, line 354: These are initialized as empty lists and if they are length 0 then they are going to be `False`'y.

```suggestion
                if deps.pub_libs or deps.priv_libs:
```

Tachi107: Much better, thanks :)

eli-schwartz (approved):
LGTM, tested locally and it both generates the correct pkgconfig files in both cases, and passes `pkgconf --validate`.

I noticed in the process that -uninstalled.pc files do NOT get any Cflags pointing to the uninstalled includedir, and that happens because we currently only support defining uninstalled includedirs by introspecting any compiled libraries associated with the pkg-config file and adding any halfway-relevant directories from there. But this won't work if there are no `deps.pub_libs` or `deps.priv_libs`.

So the -uninstalled.pc file is broken and I'm not completely sure how to fix it, but on the other hand -- it was broken before your PR, and it is after all documented as
> This is an experimental feature provided on a best-effort basis, it might not work in all use-cases.

thierryreding:
Something like this fixes the issue with gst-plugins-base. It's a bit ugly, but if nobody has any better ideas I can send a PR for this.

```
diff --git a/mesonbuild/modules/pkgconfig.py b/mesonbuild/modules/pkgconfig.py
index 2e985cf0364d..172f74d28e92 100644
--- a/mesonbuild/modules/pkgconfig.py
+++ b/mesonbuild/modules/pkgconfig.py
@@ -342,12 +342,21 @@ class PkgConfigModule(ExtensionModule):
         libdir = PurePath(coredata.get_option(mesonlib.OptionKey('libdir')))
         incdir = PurePath(coredata.get_option(mesonlib.OptionKey('includedir')))
         fname = os.path.join(outdir, pcfile)
+
+        # check if any variables reference libdir, so make sure we include it
+        # when its needed
+        need_libdir = False
+
+        for k, v in [ *variables, *unescaped_variables ]:
+            if '${libdir}' in v:
+                need_libdir = True
+
         with open(fname, 'w', encoding='utf-8') as ofile:
             if not dataonly:
                 ofile.write('prefix={}\n'.format(self._escape(prefix)))
                 if uninstalled:
                     ofile.write('srcdir={}\n'.format(self._escape(srcdir)))
-                if deps.pub_libs or deps.priv_libs:
+                if deps.pub_libs or deps.priv_libs or need_libdir:
                     ofile.write('libdir={}\n'.format(self._escape('${prefix}' / libdir)))
                 ofile.write('includedir={}\n'.format(self._escape('${prefix}' / incdir)))
             if variables or unescaped_variables:
```

thierryreding:
I think your change is fine for the problem that it solves, so I don't think we need to revert it. Instead the above works on top of it, so it should restore the old behavior where necessary and at the same time keep the libdir from being included when it's really not needed.

eli-schwartz:
In this particular case you probably don't especially want the includedir defined either (it's not a library, and it's not a *header-only* library either), so `dataonly` might be more fitting.

IIRC you should be able to define libdir as a custom variable in conjunction with dataonly.

PR #9910, opened 2022-01-30
Remove unnecessary check during VS activation

nazar-pc:
As discussed on Matrix, this doesn't seem to be a true indicator of Visual Studio environment activation


eli-schwartz:
Just to confirm -- I guess `VSINSTALLDIR` is enough on its own, and prevents meson from activating a nested vsenv?

nazar-pc:
As far as I understand it, activation just means setting a bunch of environment variables. So even if nested (or rather sequential) activation happens, it shouldn't cause any issues.

Here is an example of environment variables set during activation: https://stackoverflow.com/a/59679967

nazar-pc:
The error doesn't seem to be related to the changed made here (flaky test?)

xclaesse:
lgtm

PR #9916, opened 2022-01-31
cmake: add arch_independent kwarg

Tachi107:
CMake's `write_basic_package_version_file` has supported an `ARCH_INDEPENDENT` option since version 3.14 that make it skip its architecture check in the Version file.

With this patch Meson now supports it as well, and the change is also compatible with older CMake versions, as they will simply ignore the option.

This also slightly changes the contents of the generated Version file when `arch_independent` is not set: previously, the `if()` needed to skip the arch check was always filled with an empty string, while CMake puts `"FALSE"` (or `"TRUE"`) in it. Now, that `if()` will always be filled with either `"False"` or `"True"`, better matching CMake's behaviour.

I still have to add this to the documentation and mark the new kwarg as introduced in Meson 0.62.0 (I don't know how, though), so I'm marking this as a draft

### DIFF
--- a/docs/markdown/CMake-module.md
+++ b/docs/markdown/CMake-module.md
@@ -226,7 +226,7 @@ specification](Machine-files.md).
 ### cmake.write_basic_package_version_file()
 
 This function is the equivalent of the corresponding [CMake
-function](https://cmake.org/cmake/help/v3.11/module/CMakePackageConfigHelpers.html#generating-a-package-version-file),
+function](https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html#command:write_basic_package_version_file),
 it generates a `name` package version file.
 
 * `name`: the name of the package.
@@ -235,6 +235,8 @@ it generates a `name` package version file.
 `AnyNewerVersion`, `SameMajorVersion`, `SameMinorVersion` or `ExactVersion`.
 It defaults to `AnyNewerVersion`. Depending on your cmake installation some kind of
 compatibility may not be available.
+* `arch_independent`: *new in 0.62.0*, if true the generated package file
+  will skip architecture checks. Useful for header-only libraries.
 * `install_dir`: optional installation directory, it defaults to `$(libdir)/cmake/$(name)`
 
 
--- a/mesonbuild/modules/cmake.py
+++ b/mesonbuild/modules/cmake.py
@@ -42,6 +42,7 @@ from ..interpreterbase import (
 if T.TYPE_CHECKING:
     class WriteBasicPackageVersionFile(T.TypedDict):
 
+        arch_independent: bool
         compatibility: str
         install_dir: T.Optional[str]
         name: str
@@ -270,12 +271,14 @@ class CmakeModule(ExtensionModule):
     @noPosargs
     @typed_kwargs(
         'cmake.write_basic_package_version_file',
+        KwargInfo('arch_independent', bool, default=False, since='0.62.0'),
         KwargInfo('compatibility', str, default='AnyNewerVersion', validator=in_set_validator(set(COMPATIBILITIES))),
         KwargInfo('install_dir', (str, NoneType), default=None),
         KwargInfo('name', str, required=True),
         KwargInfo('version', str, required=True),
     )
     def write_basic_package_version_file(self, state, args, kwargs: 'WriteBasicPackageVersionFile'):
+        arch_independent = kwargs['arch_independent']
         compatibility = kwargs['compatibility']
         name = kwargs['name']
         version = kwargs['version']
@@ -296,7 +299,8 @@ class CmakeModule(ExtensionModule):
 
         conf = {
             'CVF_VERSION': (version, ''),
-            'CMAKE_SIZEOF_VOID_P': (str(self.detect_voidp_size(state.environment)), '')
+            'CMAKE_SIZEOF_VOID_P': (str(self.detect_voidp_size(state.environment)), ''),
+            'CVF_ARCH_INDEPENDENT': (arch_independent, ''),
         }
         mesonlib.do_conf_file(template_file, version_file, conf, 'meson')
 
--- a/mesonbuild/modules/cmake.py
+++ b/mesonbuild/modules/cmake.py
@@ -21,6 +21,7 @@ from . import ExtensionModule, ModuleReturnValue, ModuleObject
 from .. import build, mesonlib, mlog, dependencies
 from ..cmake import SingleTargetOptions, TargetOptions, cmake_defines_to_args
 from ..interpreter import SubprojectHolder
+from ..interpreter.type_checking import NoneType, in_set_validator
 from ..interpreterbase import (
     FeatureNew,
     FeatureNewKwargs,
@@ -33,8 +34,18 @@ from ..interpreterbase import (
 
     InvalidArguments,
     InterpreterException,
+
+    typed_kwargs,
+    KwargInfo,
 )
 
+if T.TYPE_CHECKING:
+    class WriteBasicPackageVersionFile(T.TypedDict):
+
+        compatibility: str
+        install_dir: T.Optional[str]
+        name: str
+        version: str
 
 COMPATIBILITIES = ['AnyNewerVersion', 'SameMajorVersion', 'SameMinorVersion', 'ExactVersion']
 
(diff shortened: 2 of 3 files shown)


mesonbuild/modules/cmake.py
```diff
@@ -256,7 +256,8 @@ def detect_cmake(self, state):
         self.cmake_detected = True
         return True
 
-    @permittedKwargs({'version', 'name', 'compatibility', 'install_dir'})
+    # TODO: mark arch_independent as introduced in 0.62, add release snippet
```
eli-schwartz, line 259: See the use of `@FeatureNewKwargs` used elsewhere in this file.

mesonbuild/modules/cmake.py
```diff
@@ -272,6 +273,10 @@ def write_basic_package_version_file(self, state, _args, kwargs):
         if compatibility not in COMPATIBILITIES:
             raise mesonlib.MesonException('compatibility must be either AnyNewerVersion, SameMajorVersion or ExactVersion.')
 
+        arch_independent = kwargs.get('arch_independent', False)
+        if not isinstance(arch_independent, bool):
+            raise mesonlib.MesonException('arch_independent is not a bool.')
```
eli-schwartz, line 278: Ideally we'd also revamp the function to use `@typed_kwargs` instead of `@permittedKwargs`, and then each kwarg would be a KwargInfo instead of a string, and the `KwargInfo` would contain attributes such as the allowed type (bool) and the default (False) and *also* the available since (0.62.0).

However, that is not your job (unless you want to do it for bonus points).

eli-schwartz (commented):
This change makes sense, thanks!

Commented below on how you can mark the feature availability.

mesonbuild/modules/cmake.py
Tachi107: I'll take this as a personal challenge :)

I'll either get a bit more familiar with the codebase or come back here crying for help

eli-schwartz:
> like why do I have to define the type and the list of kwargs both in `@typed_kwargs()` an in an external class? What

You may or may not want that, but doing it would be for the sake of getting [mypy](http://mypy-lang.org/) coverage. You don't have to do that too. :)

`typed_kwargs` is not python language type annotations only processed during static analysis linting, it is a runtime meson.build parser/interpreter validator that checks meson.build types. :D

Sadly this often means doing similar stuff twice. But the interpreter validator (`typed_kwargs`) is much more powerful.

> What's `(str, NoneType)`, a kind of optional type?

KwargInfo takes a class or tuple of classes that are allowed, so `KwargInfo('compatibility', str, ...)` only accepts strings, `KwargInfo('install_dir', (str, NoneType), ...)` accepts either a string or None, and `KwargInfo('foo', object)` is very naughty and accepts any object imaginable, which it should not.

mesonbuild/modules/cmake.py
```diff
@@ -256,31 +268,35 @@ def detect_cmake(self, state):
...
-        if not isinstance(name, str):
-            raise mesonlib.MesonException('Name not specified.')
+        name = kwargs['name']
 
-        compatibility = kwargs.get('compatibility', 'AnyNewerVersion')
-        if not isinstance(compatibility, str):
-            raise mesonlib.MesonException('compatibility is not string.')
+        compatibility = kwargs['compatibility']
         if compatibility not in COMPATIBILITIES:
             raise mesonlib.MesonException('compatibility must be either AnyNewerVersion, SameMajorVersion or ExactVersion.')
```
eli-schwartz, line 287: We can move this into the KwargInfo too.

mesonbuild/modules/cmake.py
```diff
@@ -256,31 +268,35 @@ def detect_cmake(self, state):
...
-    @permittedKwargs({'version', 'name', 'compatibility', 'install_dir'})
-    def write_basic_package_version_file(self, state, _args, kwargs):
-        version = kwargs.get('version', None)
-        if not isinstance(version, str):
-            raise mesonlib.MesonException('Version must be specified.')
+    # TODO: add release snippet
+    @typed_kwargs(
+        'cmake.write_basic_package_version_file',
+        KwargInfo('arch_independent', bool, default=False, since='0.62.0'),
+        KwargInfo('compatibility', str, default='AnyNewerVersion'),
```
eli-schwartz, line 275:
```suggestion
        KwargInfo('compatibility', str, default='AnyNewerVersion', validator=somefunc),
```

For the validator function `somefunc`, many parts of meson use one-off lambdas to define a completely custom function. But "must be one of a specific set of strings" is a common thing to need, so here we can use:
```python
validator=in_set_validator(set(COMPATIBILITIES))
```
(This function comes from `mesonbuild.interpreter.type_checking` and returns an inner function for you.)

eli-schwartz (commented):
I think it would also be nice to reorder the commits. Add the typed_kwargs first, then add arch_independent, rather than adding arch_independent with custom checks, and then migrating it to the typed_kwargs.

Overall though, it looks really nice.

Tachi107:
Unless I screwed up something with the small refactor, I think this is ready :)

eli-schwartz:
Hmm, what happened to the commit messages? :p There seems to be some mysterious "feat()" function being applied to the word "cmake".

mesonbuild/modules/cmake.py
```diff
@@ -256,31 +266,25 @@ def detect_cmake(self, state):
...
-        if compatibility not in COMPATIBILITIES:
-            raise mesonlib.MesonException('compatibility must be either AnyNewerVersion, SameMajorVersion or ExactVersion.')
+    @typed_kwargs(
+        'cmake.write_basic_package_version_file',
+        KwargInfo('compatibility', str, default='AnyNewerVersion', validator=in_set_validator(set(COMPATIBILITIES))),
+        KwargInfo('install_dir', (str, NoneType), default=None),
+        KwargInfo('name', str, required=True),
+        KwargInfo('version', str, required=True),
+    )
+    def write_basic_package_version_file(self, state, _args, kwargs: 'WriteBasicPackageVersionFile'):
```
eli-schwartz, line 278: `args` got renamed to a leading underscore, this is not the usual Meson style even for unused arguments.

On the other hand, this does conveniently point out that there are no allowed args, so we can raise an error message if people try to pass args anyway. See the `@noPosargs` decorator (from the same place `@typed_kwargs` is imported from).

docs/markdown/CMake-module.md
```diff
@@ -134,37 +133,37 @@ together with the `options` kwarg.
 This object is returned by the `subproject` function described above
 and supports the following methods:
 
- - `dependency(target)` returns a dependency object for any CMake target. The
-   `include_type` kwarg *(new in 0.56.0)* controls the include type of the
```
eli-schwartz, line 138: This file has been reformatted a lot, which makes it harder to review and should probably be in its own commit if you want to do it.

eli-schwartz (changes requested):

mesonbuild/modules/cmake.py
Tachi107: I didn't even realize, my IDE probably did it without even asking. But thanks I'll add the `@noPosargs` decorator

Tachi107:
> Hmm, what happened to the commit messages? :p There seems to be some mysterious "feat()" function being applied to the word "cmake".

Sorry, I'm used to follow the [Conventional Commits](https://www.conventionalcommits.org) specification, but that you don't so they look weird :/

I find them useful, so you might want to try them out in the future, but this time I'll keep them "normal"

mesonbuild/modules/cmake.py
eli-schwartz: Oh my gosh, lol. I wonder why... but it isn't your fault then. :D 

`@noPosargs` is still useful either way.

docs/markdown/CMake-module.md
Tachi107: Undid all the formatting changes, kept only the actual changes in content

docs/markdown/CMake-module.md
```diff
@@ -273,15 +275,15 @@ conf = configuration_data()
 conf.set_quoted('VAR', 'variable value')
 
 cmake.configure_package_config_file(
-    name: 'myProject',
-    input: 'myProject.cmake.in',
-    configuration: conf
+  name: 'myProject',
+  input: 'myProject.cmake.in',
+  configuration: conf
```
eli-schwartz, line 280: All the other changes in this file are now:
- the doc updates
- correcting a link
- fixing a syntax highlight issue
- using the correct capitalization for CMake

But this here is a style change  for differing tab length conventions (and changing from the style I personally like to one I don't like). Can we skip this diff hunk?

mesonbuild/modules/cmake.py
```diff
@@ -290,7 +298,8 @@ def write_basic_package_version_file(self, state, _args, kwargs):
 
         conf = {
             'CVF_VERSION': (version, ''),
-            'CMAKE_SIZEOF_VOID_P': (str(self.detect_voidp_size(state.environment)), '')
+            'CMAKE_SIZEOF_VOID_P': (str(self.detect_voidp_size(state.environment)), ''),
+            'CVF_ARCH_INDEPENDENT': (arch_independent, '')
```
eli-schwartz, line 302: For future proofing in case we ever need to add another line here, we can leave a trailing comma on this line too.

This is not some terrible language like json, where trailing commas are a syntax error. :D

docs/markdown/CMake-module.md
Tachi107: Oke, I only kept the link update because it pointed to CMake 3.11, where ARCH_INDEPENDENT isn't defined

mesonbuild/modules/cmake.py
```diff
@@ -33,8 +34,18 @@
 
     InvalidArguments,
     InterpreterException,
+
+    typed_kwargs,
+    KwargInfo,
 )
 
+class WriteBasicPackageVersionFile(T.TypedDict):
```
eli-schwartz, line 42: This should only be defined when `if T.TYPE_CHECKING:`, see CI failures.

eli-schwartz: In other words, this works on python 3.8 but errors out on python 3.7, and once moved into a mypy-only code block we can simply not care that it doesn't support using mypy which itself runs on python 3.7

Tachi107: Done, thanks :)
I wasn't really sure how to fix that

PR #9917, opened 2022-01-31
Fix dep.get_variable() with empty string default_value

xclaesse:


xclaesse:
This fix regression introduced by https://github.com/mesonbuild/meson/pull/9729.

xclaesse:
CC @dcbaker 

I haven't reviewed the rest of that PR, maybe there are other similar cases?

dcbaker (approved):
LGTM, thanks for fixing this.

nirbheek:
https://github.com/mesonbuild/meson/pull/9729 is not in 0.61, so this doesn't need cherry-picking. However, I have backported the unit test so we know that this is still working.

PR #9918, opened 2022-01-31
mesonlib: Fix Popen_safe_legacy() stderr assumption

oleavr:
It may be `None`. This was encountered with radare's build system on
Windows, where symbolextractor.py crashes without any output displayed.

### DIFF
--- a/mesonbuild/mesonlib/universal.py
+++ b/mesonbuild/mesonlib/universal.py
@@ -1416,7 +1416,7 @@ def Popen_safe_legacy(args: T.List[str], write: T.Optional[str] = None,
         else:
             o = o.decode(errors='replace').replace('\r\n', '\n')
     if e is not None:
-        if sys.stderr.encoding:
+        if sys.stderr is not None and sys.stderr.encoding:
             e = e.decode(encoding=sys.stderr.encoding, errors='replace').replace('\r\n', '\n')
         else:
             e = e.decode(errors='replace').replace('\r\n', '\n')


eli-schwartz:
Why is it None?

And why is sys.stdout not at risk the same way?

oleavr:
> Why is it None?

I don't know yet, but one theory is that it's because Ninja [passes](https://github.com/ninja-build/ninja/blob/41156cd9c97a4e49473c473b7952f8cb7c43c56c/src/subprocess-win32.cc#L98-L99) the same named pipe handle for stdout and stderr.

If I run the exact same command from cmd.exe, then sys.stderr is fine. It's only when Ninja runs it that it's None.

> And why is sys.stdout not at risk the same way?

I can in theory also be None, but I wasn't sure if it's desirable to deal with that in the same PR.

trufae:
The offending build log is here https://github.com/radareorg/radare2/actions/runs/1775047277

err

In my case, i can build on windows properly with any version of meson (0.59.0, 0.59.1, 0.61.1) and ninja (tried 1.9.0 and 1.10.2), and latest VS2019Community(16.11.9). But for some reason in the CI (github actions) is failing (without changing any line of code).

So I assume an update in the GHA images surfaced this bug, but for some reason, some people like @oleavr can repro in local.

The reason for stdout or stderr to be None, should be (according to cpython code) because the fd is of an unknown type or invalid.

```c
hfile = (HANDLE)_get_osfhandle(fd);
    _Py_END_SUPPRESS_IPH
    return (hfile != INVALID_HANDLE_VALUE
            && GetFileType(hfile) != FILE_TYPE_UNKNOWN);
```

PD: I agree that the same check should be done for `sys.stdout` in the line `1414`, just in case.
PD2: meson is hiding the exception and no error code or program output is printed

trufae:
I can confirm the fix https://github.com/radareorg/radare2/actions/runs/1775286569

`pip install git+https://github.com/frida/meson.git@f7f25b19a8d71cebf8e2934733eb041eb6862eee`

eli-schwartz:
> I don't know yet, but one theory is that it's because Ninja [passes](https://github.com/ninja-build/ninja/blob/41156cd9c97a4e49473c473b7952f8cb7c43c56c/src/subprocess-win32.cc#L98-L99) the same named pipe handle for stdout and stderr.

This seems odd, because redirecting stderr to stdout, or the reverse, is a common practice that should not break.

> PD2: meson is hiding the exception and no error code or program output is printed

If stderr is so broken that python thinks it doesn't exist, then it would seem to make sense that it won't print anything. This PR attempts to fix that by preventing it from crashing on attribute access for something that doesn't exist, but the underlying problem that stderr doesn't exist remains in place, and I think that's a real issue which should not be swept under the rug.

By the way: ninja works fine on Windows everywhere else, AFAIK. And the symbolextractor is part of every project build which a) uses ninja, b) compiles a shared library DLL.

oleavr:
> > I don't know yet, but one theory is that it's because Ninja [passes](https://github.com/ninja-build/ninja/blob/41156cd9c97a4e49473c473b7952f8cb7c43c56c/src/subprocess-win32.cc#L98-L99) the same named pipe handle for stdout and stderr.
> 
> This seems odd, because redirecting stderr to stdout, or the reverse, is a common practice that should not break.

Yeah I know. I was thinking about whether this could be some kind of named pipe idiosyncrasy where the file handles have to be distinct, from a Windows process creation API point-of-view. But nah that really can't be it, it's too weird.

> > PD2: meson is hiding the exception and no error code or program output is printed
> 
> If stderr is so broken that python thinks it doesn't exist, then it would seem to make sense that it won't print anything. This PR attempts to fix that by preventing it from crashing on attribute access for something that doesn't exist, but the underlying problem that stderr doesn't exist remains in place, and I think that's a real issue which should not be swept under the rug.

Totally agree. I honestly don't know if it makes sense to handle it in a Meson context.

> By the way: ninja works fine on Windows everywhere else, AFAIK. And the symbolextractor is part of every project build which a) uses ninja, b) compiles a shared library DLL.

It's strange though that GitHub's CI exhibits this behavior, and the Windows 10 + VS2019 setup in my basement CI also does. Something is fishy with this setup, just haven't been able to figure out exactly what yet. (Though I haven't yet found enough time to really dig into this rabbit-hole.)

trufae:
I spent 2 days digging in that hole without any real insights, mainly because i can't reproduce this problem in my local machine, and debugging in the CI is not really a pleasant experience.

Despite not finding the root cause of this problem, meson fails when running the internal tool to dump the library symbols into a text file, which seems not to be used for anything, and also, it doesn't have a way to disable this, so i can't find any workaround. This tool (is from visual studio) and it runs fine with no error code or

The reasons of having `stderr` as `None` can be different ones from what I understand (invalid codepage, being a pipe, invalid fd or directly not existing), so it seems to me like meson should avoid failing in those situations, and the patch proposed by @oleavr works for me.

According to the documentation `sys.__stderr__` can be also used and it keeps the original value, so maybe it's worth checking when sys.stderr is none. Is meson able to log into a file? This way I could see the real error, because I can't see even the return code from the child process that fails.

jpakkane:
Since this fixes an actual issue, seems to be reasonable and does not break anything that I can tell, I'm going to merge this.

carlkl:
The issue described here seems to be related to this pip issue: [stderr duplication failed in console scripts ](https://github.com/pypa/pip/issues/10875). `sys.stderr` will be swallowed if a package is started by the console launchers located in `Scripts` _and_ if this launcher was build with pip>=22.0.
This PR may help with meson, but the problem with `sys.stderr is None` may hit you later (I've seen this with `tqdm`).

The correct mitigation right now is to install pip with `python -m pip install -U pip==21.3.1` and also reinstall meson to ensure the meson launcher will be build again.

eli-schwartz:
@carlkl thank you for the link.

This is a very good demonstration of why I disagree with @jpakkane's analysis of "seems to be reasonable", and why I would appreciate if this PR could be reverted.

It adds undocumented magic hacks that likely result in incorrect processing of information in the event that the issue comes up to begin with, and it does it for a very niche situation that only manifests when some other project has a bug.

This is why I wanted to defer merging a change that is not going to be *released* right now anyway, until we actually understand the issue and can better evaluate whether the PR is a good idea. There was no practical benefit in prematurely merging this PR, and it turned out to be a bug in pip rather than Meson...

jpakkane:
On the other hand:

- that pip issue is probably not the only reason why `stderr` might be `None` 
- there are people who can't or won't update their Pip versions (typically because they need to use distro packaged stuff as is and they are not allowed to update them)

Being robust against bad input is thus beneficial in these cases. The best solution would be if Python had a stricter type system that would make these kinds of bugs impossible but sadly that is not an option.

carlkl:
Maybe this patch should be augmented with a warning to the user that this is an intermediate patch to circumvent a problem with pip that is not solved yet. And this patch should be removed as soon as pip doesn't produce faulty windows launchers anymore.

> that pip issue is probably not the only reason why `stderr` might be `None`

What kind of reasons do you think about?

jpakkane:
> What kind of reasons do you think about?

I don't know any. But experience has shown that if an error is made once, it will most likely done multiple times.

PR #9919, opened 2022-01-31
unittests: fine-tune the check for whether we are in CI

eli-schwartz:
The $CI environment variable may be generally set by Github or Gitlab actions, and is not a reliable indicator of whether we are running "CI". It could also, for an absolutely random example that didn't *just happen*, be Alpine Linux's attempt to enable the Meson testsuite in their packaging, which... uses Gitlab CI.

In this case, we do want to perform normal skipping on not-found requirements. Instead of checking for $CI, check for $MESON_CI_JOBNAME as we already use that in all of our own CI jobs for various reasons.

This makes it easier for linux distros to package Meson without accumulating hacks like "run the testsuite using `env -u CI`".

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -35,6 +35,7 @@ jobs:
     - env:
         CPPFLAGS: "-I/usr/local/include"
         LDFLAGS: "-L/usr/local/lib"
+        MESON_CI_JOBNAME: unittests-appleclang
         MESON_UNIT_TEST_BACKEND: ninja
         # These cannot evaluate anything, so we cannot set PATH or SDKROOT here
       run: |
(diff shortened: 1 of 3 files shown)


nirbheek:
Glad we made that into a helper function, eh? ;) LGTM

eli-schwartz:
Oh G-d, the nightmare of changing this string in a hundred places... :D

eli-schwartz:
Added a free extra commit to enable pytest more often in the unittests. Alpine could not make use of pytest in their packaging for better output, because while pytest itself is in `main`, pytest-xdist is in `community` so meson cannot checkdepends on it.

This seems silly to me, because xdist isn't really important enough to revert back to unittest.

nirbheek:
lgtm, we should merge this.

Issue #9707, opened 2021-12-08
ldconfig during install causes exception during cross build


Princemachiavelli:
**Describe the bug**
Since https://github.com/mesonbuild/meson/commit/2c3a68bc30cf6e08c2e3b77b64718ef3a0e5d78c Meson attempts to update the ld.so cache during `meson install`. In https://github.com/mesonbuild/meson/commit/6b7d43ebe0d77ac85fa2c01e0e358efc64418126 errors from ldconfig are ignored but if the ldconfig binary is not compatible with the host during a cross build then Popen throws an exception that is not caught:

`OSError: [Errno 8] Exec format error: 'ldconfig'`

The issue *can* be avoided by configuring binfmt on the host system but that wasn't necessary prior to this change (https://github.com/mesonbuild/meson/commit/2c3a68bc30cf6e08c2e3b77b64718ef3a0e5d78c).

The issue is also being tracked in the NixOS project: https://github.com/NixOS/nixpkgs/pull/144779

I'm not a meson expert or even a NixOS expert so creating a simple test project was a bit rough. Let me know if you need more information. Let me know if you need more information:

**To Reproduce**

On a Linux machine during an cross build.
Also the machine's binmft is either not configured for the target system architecture or is disabled via:
`echo 0 | sudo tee /proc/sys/fs/binfmt_misc/status`.

Then the error will occur during the install action:
`meson install --no-rebuild`

src: nixos.c
```
#include <stdio.h>

int main(int argc, char **argv) {
    printf("Hello World");
}
```

meson.build
```
project('nixos', 'c')

cc = meson.get_compiler('c')
dirs = []
executable('nixos', 'nixos.c')
```

cross-file:
```
[properties]
needs_exe_wrapper = true

[host_machine]
system = 'linux'
cpu_family = 'aarch64'
cpu = 'aarch64'
endian = 'little'

[binaries]
llvm-config = 'llvm-config-native'
```
Meson/NixOS Log:

```
[...]

PR #9926, opened 2022-02-01
minstall: stop running ldconfig for the user

### LINKED ISSUE #9721: Stop calling ldconfig on `meson install`
We've now had three different bugs around this, and it's slow when we don't need to call it, and it doesn't always yield the desired results anyway. I'd like to call for its removal in 0.61. I'm opening this, so we can generate some consensus in a single place.

eli-schwartz:
This was a nice idea in theory, but in practice it had various problems:

- On the only platform where ldconfig is expected to be run, it is
 really slow, even when the user uses a non-default prefix and ldconfig doesn't even have permission to run
- On FreeBSD, it bricked the system: #9592
- On cross builds, it should not be used and broke installing, because ldconfig may not be runnable without binfmt + qemu: #9707
- it prints weird and confusing errors in the common "custom prefix" layout: #9241

Some of these problems can be or have been fixed. But it's a constant source of footguns and complaints and for something that was originally supposed to be just "it's the right thing to do anyway, so just do it automatically" it is entirely too risky.

Ultimately I do not think there is justification for keeping this feature in since it doesn't actually make everyone happy. Better for
users to decide whether they need this themselves.

This is anyways the case for cmake and autotools and generally any other build system, so it should not be too intimidating...

Fixes #9721

### DIFF
--- a/docs/markdown/snippets/no_ldconfig.md
+++ b/docs/markdown/snippets/no_ldconfig.md
@@ -1,0 +1,5 @@
+## ldconfig is no longer run on install
+
+Due to various issues of fragility and concern that it doesn't predictably do
+the right thing, meson no longer runs ldconfig during `meson install`, and
+users who need it run should run it themselves, instead.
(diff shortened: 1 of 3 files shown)


dcbaker (changes requested):
The d.is_cross_build is now  completely unused, it was just for fixing the cross build bug. You can rip that out too :)

dcbaker (approved):
Looks great. Thanks for doing this

PR #9927, opened 2022-02-01
new custom dependency lookup for libdl

noiseless:
The background to this patch can be seen here: https://github.com/mpv-player/mpv/pull/9793

In a nutshell:
Different systems/implementations of libc have different policies about having the 'libdl' library in the system. On some systems (*BSD, glibc >= 2.34) the dynamic linker interface is part of libc, on others (mostly glibc < 2.34) a separate 'libdl' library is supplied which contains the necessary functions.

In order to simplify the handling of having libdl in meson, @eli-schwartz suggested that I implement separate logic for this in meson, similar to mesonbuild/meson@2c6ccfe and mesonbuild/meson@214d035.

That's exactly what I ended up doing.
Tested the mpv build with meson with my changes on OpenBSD (libdl is missing) and on Archlinux (glib 2.33, libdl comes as a separate library).
I used this patch for mpv instead of the one now presented in https://github.com/mpv-player/mpv/pull/9793 :
```
--- a/meson.build
+++ b/meson.build
@@ -336,7 +336,7 @@ if get_option('ta-leak-report')
     features += 'ta-leak-report'
 endif
 
-libdl = cc.find_library('libdl', required: false)
+libdl = dependency('dl', required: false)
 if libdl.found()
     dependencies += libdl
     features += 'libdl'
```

The build was successful on both OpenBSD and Linux.

I'd be happy to get any comments, objections or other criticism you might have.

### DIFF
--- a/docs/markdown/Dependencies.md
+++ b/docs/markdown/Dependencies.md
@@ -393,6 +393,17 @@ foreach h : check_headers
 endforeach
 ```
 
+## dl (libdl)
+
+*(added 0.62.0)*
+
+Provides access to the dynamic link interface (functions: dlopen,
+dlclose, dlsym and others). On systems where this is not built
+into libc (mostly glibc < 2.34), tries to find an external library
+providing them instead.
+
+`method` may be `auto`, `builtin` or `system`.
+
 ## Fortran Coarrays
 
 *(added 0.50.0)*
--- a/mesonbuild/dependencies/misc.py
+++ b/mesonbuild/dependencies/misc.py
@@ -60,6 +60,25 @@ def netcdf_factory(env: 'Environment',
     return candidates
 
 
+class DlBuiltinDependency(BuiltinDependency):
+    def __init__(self, name: str, env: 'Environment', kwargs: T.Dict[str, T.Any]):
+        super().__init__(name, env, kwargs)
+
+        if self.clib_compiler.has_function('dlopen', '#include <dlfcn.h>', env)[0]:
+            self.is_found = True
+
+
+class DlSystemDependency(SystemDependency):
+    def __init__(self, name: str, env: 'Environment', kwargs: T.Dict[str, T.Any]):
+        super().__init__(name, env, kwargs)
+
+        h = self.clib_compiler.has_header('dlfcn.h', '', env)
+        self.link_args = self.clib_compiler.find_library('dl', env, [], self.libtype)
+
+        if h[0] and self.link_args:
+            self.is_found = True
+
+
 class OpenMPDependency(SystemDependency):
     # Map date of specification release (which is the macro value) to a version.
     VERSIONS = {
@@ -557,6 +576,13 @@ cups_factory = DependencyFactory(
     [DependencyMethods.PKGCONFIG, DependencyMethods.CONFIG_TOOL, DependencyMethods.EXTRAFRAMEWORK, DependencyMethods.CMAKE],
     configtool_class=CupsDependencyConfigTool,
     cmake_name='Cups',
+)
+
+dl_factory = DependencyFactory(
+    'dl',
+    [DependencyMethods.BUILTIN, DependencyMethods.SYSTEM],
+    builtin_class=DlBuiltinDependency,
+    system_class=DlSystemDependency,
 )
 
 gpgme_factory = DependencyFactory(
--- a/docs/markdown/snippets/libdl-dependency.md
+++ b/docs/markdown/snippets/libdl-dependency.md
@@ -1,0 +1,8 @@
+## New custom dependency for libdl
+
+```
+dependency('dl')
+```
+
+will now check for the functionality of libdl.so, but first check if it is
+provided in the libc (for example in libc on OpenBSD or in musl libc on linux).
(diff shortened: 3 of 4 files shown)


mesonbuild/dependencies/misc.py
```diff
@@ -60,6 +60,25 @@ def netcdf_factory(env: 'Environment',
     return candidates
 
 
+class DlBuiltinDependency(BuiltinDependency):
+    def __init__(self, name: str, env: 'Environment', kwargs: T.Dict[str, T.Any]):
+        super().__init__(name, env, kwargs)
+
+        if self.clib_compiler.has_function('dlopen', '#include <dlfcn.h>', env)[0]:
```
eli-schwartz, line 67: including dlfcn.h isn't strictly always necessary, although there is nothing wrong with doing so. There is a dummy/stub declaration that is used instead, if you don't include a header.

https://github.com/mesonbuild/meson/blob/5d932f31b985e46a4995d3c02b021cc94bef901d/mesonbuild/compilers/mixins/clike.py#L816-L823

Feel free to ignore this comment entirely. Arguably adding the include means you do a free check that the header is available (it should never be missing, but who knows).

docs/markdown/Dependencies.md
```diff
@@ -393,6 +393,15 @@ foreach h : check_headers
 endforeach
 ```
 
+## dl (libdl)
+
```
eli-schwartz, line 397: Should mention:

```
 *(added 0.62.0)*
```

eli-schwartz (changes requested):
Thank you, this looks great!

Please include a release notes snippet in `docs/markdown/snippets/*.md`.

docs/markdown/Dependencies.md
noiseless: Thank you, I'll do a force-push with the fix.

mesonbuild/dependencies/misc.py
noiseless: I'll think about it now, thank you.

noiseless:
```
-        if self.clib_compiler.has_function('dlopen', '#include <dlfcn.h>', env)[0]:
+        if self.clib_compiler.has_function('dlopen', '', env)[0]:
```

With this patch, I had a successful mpv build on Archlinux. However, this is not a sign of success, because I have glibc 2.33, i.e. libdl is present in the system.
I can check on OpenBSD, which doesn't have this library, only later. I'll be sure to write here about the results as soon as I do it.

Intuitively it seems to me that the variant with an explicit specification of the title '#include <dlfcn.h>' is more correct, but I do not have a good reason for this position.
If an OpenBSD build with these modifications succeeds, I'm willing to add them to my patch.

noiseless: I think that if the OpenBSD build will succeed with this patch, there is no reason to do otherwise. Your advice is very valuable - thank you very much.
I'll be able to test the patch later tonight (MSK), then I'll post the final version.

eli-schwartz:
Not sure if you forgot to add a release notes snippet in `docs/markdown/snippets/*.md`, but that is still outstanding.

See the other files in there for details, but basically just include a `## XXX` markdown header followed by the paragraph(s) of release notes text.

noiseless:
> Not sure if you forgot to add a release notes snippet in `docs/markdown/snippets/*.md`, but that is still outstanding.

Oops, sorry, I didn't do what you asked. I will now redo it correctly.

noiseless:
Now it's [done](https://github.com/mesonbuild/meson/pull/9927/files#diff-9388c0a900ff1332e662d267f133cf11ca1e0fc52b80e181d49829c8af97f60c).

noiseless:
Oh, you already merged my pull-request with the master branch.

Thank you very much!

I promised to check under OpenBSD a variant of the patch without explicitly specifying a header file. Keeping my promise: the build succeeds. Relevant part of the log:
```
Running compile:
Working directory:  /usr/ports/pobj/mpv-0.34.999/build-amd64/meson-private/tmpc9kpadhm
Command line:  cc -L/usr/local/lib -I/usr/local/include /usr/ports/pobj/mpv-0.34.999/build-amd64/meson-private/tmpc9kpadhm/testfile.c -o /usr/ports/pobj/mpv-0.34.999/buil
d-amd64/meson-private/tmpc9kpadhm/output.exe -O2 -pipe -g -D_FILE_OFFSET_BITS=64 -O0 -Werror=implicit-function-declaration 

Code:
 
        #define dlopen meson_disable_define_of_dlopen
        
        #include <limits.h>
        #undef dlopen
        
        #ifdef __cplusplus
        extern "C"
        #endif
        char dlopen (void);
        
        #if defined __stub_dlopen || defined __stub___dlopen
        fail fail fail this function is not going to work
        #endif
        
        int main(void) {
          return dlopen ();
        }
Compiler stdout:
 
Compiler stderr:
 
Run-time dependency dl found: YES
```
Diff:
```
-        if self.clib_compiler.has_function('dlopen', '#include <dlfcn.h>', env)[0]:
+        if self.clib_compiler.has_function('dlopen', '', env)[0]:
```

While I was going home, I came to the conclusion that this option, as you suggest, is still more correct - if in some exotic system the required header will be called differently and still be identical in everything else - your code will work, while mine will not. From my point of view, this is a good reason to do as you suggested to me originally.
Shall we leave it as it is or shall we change it?

And in any case, thank you very much again! Very impressed with the speed with which you provide help and the usefulness of the solutions you offer!

eli-schwartz:
It's probably not worth bothering, unless someone knows of a system where `dlopen()` is provided by a differently-named header.

noiseless:
Yes, I agree. Thank you.

PR #9930, opened 2022-02-01
mtest: test output changes

trhd:
This is the second iteration of #9232.

I've done my best to better split the changes into smaller commits although while doing that I ended writing even more changes.

The changes can now be summarized:

> ## Changes in the textual test log output
> 
> Test specific metadata was split to multiple lines to improve readability.
> Visual hints were added to group togehter all the lines related to a
> specifc test.
> 
> Textual test log (too) now respects the  and 
> options.
> 
> ## Changes in the test output in the console
> 
> The way test details (when being verbose or printing details for a failed
> test) get printed was updated. Most notable changes are:
> 
> * each line is prefixed with the relevant test number
> * "headers" and indentation is used to group relevant lines together
> * the style in which subtest results are listed changed
> 
> Results of individual subtests are displaed only in verbose mode or when
> printing details of a failed test.
> 
> The "live" output (when being verbose and using only one process) now
> follows the style of the normal verbose output more closely.
> 
> The "live" output is now automatically used when running only one test in
> verbose mode.
> 
> The statusline was updated to a multi line one.
> 
> The status line in quiet mode was simplified.

I made a recording where I try to demo the changes: https://asciinema.org/a/dE9DsiPzUXJcW5X0q0XyO4vQ9

I have some concers w.r.t these changes:

- Multiple status lines tend to "flicker" a bit when lots of lines are being outputted. Some people might find this annoying.
- There is currently no way to disable "live" output when using only one process (though this is true even withtout these changes). I thought about adding an option for that but too may options for the command is not that great either.
- The first iteration received feedback that the statusline is not nice on some windows terminal. I failed to reproduce this but then again my Windows skills are lacking. I tried these changes with the "command prompt for VS 2019" but didn't notice any issues there. I feel the statusline should just be disabled when using an incapable terminal (think `meson test | cat`) but without being able to reproduce the issue I don't quite know what to do.
- I've only tried this out in Sourcehut's and GitLab's CI systems. Other CI sytems (Travis, Azures, ...) have not been tested at all.
- Do other people like the change?

I'll mark thie PR as WIP until the above concerns get clarified and we can get a blessing from the relevant people. I'll mention @bonzini and @anarazel here by name since they joined the discussion during the first iteration.


anarazel:
> There is currently no way to disable "live" output when using only one process (though this is true even without these changes). I thought about adding an option for that but too may options for the command is not that great either.

That seems fine to me.

The output doesn't look good when the terminal isn't high. The RUNNING lines take up all the space. Occasionally one of the "OK" lines flickers in.  Part of that is just due to the number of running tests shown. It's exacerbated by the existing problem of the line-length often being very long and wrapping around.

jpakkane:
How will this tie in with #9834?

trhd:
> How will this tie in with #9834?

Needs to be resolved by hand for sure but shouldn't be that big of a problem (guessing after looking at the commits).

trhd:
> The output doesn't look good when the terminal isn't high. The RUNNING lines take up all the space. Occasionally one of the "OK" lines flickers in. Part of that is just due to the number of running tests shown. It's exacerbated by the existing problem of the line-length often being very long and wrapping around.

I tried improving handling of small terminals by pushing an additional commit in. Would you say it worked?

trhd:
> This pull request **introduces 2 alerts** when merging [13fd52b](https://github.com/mesonbuild/meson/commit/13fd52badeb6f9d9dd226d3c2473bd82b383b920) into [6c1bb02](https://github.com/mesonbuild/meson/commit/6c1bb02f138a8c956a217c278dc8d82bfbc6d849) - [view on LGTM.com](https://lgtm.com/projects/g/mesonbuild/meson/rev/pr-b5b62fcd3157f4d3731fb55cbc03a10f41d3676f)
> 
> **new alerts:**
> 
>     * 2 for Unused local variable

False positives.

anarazel:
> I tried improving handling of small terminals by pushing an additional commit in. Would you say it worked?

It's a bit better. Previously the progress was useful without using as much space, even in "lower" terminals. But that's perhaps OK to give up.


The run counter seems confusing to me:
```
205/209 postgresql:tap+ssl / ssl/t/001_ssltests.pl                                         OK                6.48s   110 subtests passed
  2/209 postgresql:pg_regress+main / main/pg_regress                                       OK               29.53s
158/209 postgresql:tap+recovery / recovery/t/018_wal_optimize.pl                           OK               14.42s   38 subtests passed
189/209 postgresql:tap+subscription / subscription/t/022_twophase_cascade.pl               OK                9.81s   41 subtests passed
194/209 postgresql:tap+subscription / subscription/t/027_nosuperuser.pl                    OK                8.45s   14 subtests passed
195/209 postgresql:tap+subscription / subscription/t/100_bugs.pl                           OK               13.08s   7 subtests passed
  3/209 postgresql:isolation+main / main/isolation                                         OK               35.68s
```
I guess that's the number in which tasks started. But that's not useful information next to the number of tests in total...

bonzini:
Thanks, I'll review this early next week.

mesonbuild/mtest.py
```diff
@@ -903,8 +903,7 @@ def num(self) -> int:
             self._num = TestRun.TEST_NUM
         return self._num
 
-    @property
-    def detail(self) -> str:
```
bonzini, line 903: Please explain in the commit message why this is an improvement or why it is required for the rest of the changes.

mesonbuild/mtest.py
```diff
@@ -1124,15 +1124,13 @@ async def read_decode(reader: asyncio.StreamReader, console_mode: ConsoleUser) -
...
-                            console_mode: ConsoleUser) -> str:
+async def read_decode_lines(reader: asyncio.StreamReader,
+                            q: 'asyncio.Queue[T.Optional[str]]') -> str:
     stdo_lines = []
     try:
         while not reader.at_eof():
             line = decode(await reader.readline())
             stdo_lines.append(line)
-            if console_mode is ConsoleUser.STDOUT:
-                print(line, end='', flush=True)
```
bonzini, line 1125: This is needed to flush the output on every line with `--verbose` mode, why are you removing it? Please use commit messages fully, not just the subject.

mesonbuild/mtest.py
```diff
@@ -520,17 +519,20 @@ def __init__(self) -> None:
             self.sub = '| '
             self.spinner = self.ASCII_SPINNER
 
-    def flush(self) -> None:
-        if self.should_erase_line:
-            print(self.should_erase_line, end='')
-            self.should_erase_line = ''
+    @classmethod
```
bonzini, line 512: This is a bad idea. The ConsoleLogger has a single instance, so there's no need to move things to the class side. If you need to access it from elsewhere keep the methods in the instance and add a `get_instance()` class method to retrieve the singleton.

mesonbuild/mtest.py
bonzini: Ah, I see that this is `read_decode_lines()` so the subject is incorrect; and because it is used with `self.runobj.needs_parsing == True`, the `console_mode` cannot be `ConsoleUser.STDOUT`.

mesonbuild/mtest.py
```diff
@@ -1120,9 +1121,10 @@ async def read_decode(reader: asyncio.StreamReader, console_mode: ConsoleUser) -
             except asyncio.LimitOverrunError as e:
                 line_bytes = await reader.readexactly(e.consumed)
             line = decode(line_bytes)
-            stdo_lines.append(line)
-            if console_mode is ConsoleUser.STDOUT:
-                print(line, end='', flush=True)
+            if line:
```
bonzini, line 1205: Why the "if line:" addition? Please explain in the commit message (but it seems wrong).

mesonbuild/mtest.py
```diff
@@ -1482,7 +1482,10 @@ async def _run_cmd(self, harness: 'TestHarness', cmd: T.List[str]) -> None:
             parse_task = asyncio.ensure_future(parse_coro)
             stdo_task = stde_task = None
         else:
-            stdo_task, stde_task = p.communicate(self.runobj, self.console_mode)
+            def printer(line: str) -> None:
+                print(line, end='')
```
bonzini: `flush=True` missing.

mesonbuild/mtest.py
```diff
@@ -1032,25 +1022,40 @@ def complete(self) -> None:
...
+                    harness: 'TestHarness',
+                    lines: T.AsyncIterator[str],
+                    printer: T.Optional[T.Callable[[str], bool]] = None) -> None:
         res = None
 
         async for i in TAPParser().parse_async(lines):
             if isinstance(i, TAPParser.Bailout):
-                res = TestResult.ERROR
-                harness.log_subtest(self, i.message, res)
+                msg = harness.format_subtest(self, i.message, TestResult.ERROR, ConsoleLogger.SUB)
```
bonzini: The formatting code should not be part of TestRun subclasses. The printer callback should receive the parameters that are needed to call `harness.format_subtest`.

mesonbuild/mtest.py
```diff
@@ -1616,6 +1616,9 @@ def __init__(self, options: argparse.Namespace):
             for s in t.suite:
                 ss.add(s)
         self.suites = list(ss)
+        console_busy = any(runner.console_mode == ConsoleUser.GDB
+                           for runner in self.get_runners())
```
bonzini, line 1663: I think it's better to only create the runners once.

I think what you want is to move this code:

```python
        if self.options.setup:
            env = self.merge_setup_options(options, test)
        else:
            env = os.environ.copy()
```

to `TestHarness.__init__`.

mesonbuild/mtest.py
```diff
@@ -1422,8 +1423,12 @@ def __init__(self, test: TestSerialisation, env: T.Dict[str, str], name: str,
         else:
             timeout = self.test.timeout * self.options.timeout_multiplier
 
+        if self.options.quiet:
+            verbose = 0
+        else:
+            verbose = max(test.verbose, self.options.verbose)
```
bonzini, line 1446: I think `test.verbose` is a bool, not an int?

trhd:
Thanks for the good comments. I'll try to make an update soon.

I did try these changes both in Gitlab and Sourcehut and didn't see any issues there. Looked to me as if the whole status line was just disabled there (along with colors and all other "TTY features").  This is actually something that seems to have changed since the first iteration of these changes as then a change was required in Meson to achieve similar result. Not sure if the change has been in Meson or on the CI side but was pleased to see that those changes were not needed in Meson any more.

mesonbuild/mtest.py
trhd: You are right. That is a better approach.

mesonbuild/mtest.py
trhd: Indeed. The commit message was wrong.

mesonbuild/mtest.py
trhd: Hopefully this is better now. I tried to improve the commit messages all around...

mesonbuild/mtest.py
trhd: Hopefully the end result is even better with no class side functions or even `get_instance()`.

mesonbuild/mtest.py
trhd: Explained.

mesonbuild/mtest.py
trhd: Added.

mesonbuild/mtest.py
trhd: It was changed to int to support specifying more than one verbose flag.

mesonbuild/mtest.py
```diff
@@ -482,6 +482,9 @@ def __init__(self, use_statusline: bool) -> None:
         self.stop = False
         self.test_count = 0
         self.started_tests = 0
+        self.spinner = ['   ', '.  ', '.. ', '...', ' ..', '  .']
```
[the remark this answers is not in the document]
trhd: Only one line remains from the previous spinner code and the code is so heavily dependent of the changes in between the commits that I feel moving those changes around is not worth the trouble.

trhd:
Lots of changes...

> Hi, as a start please submit the following commits as they can be merged immediately:
>     * "mtest: show "timed out" message with --print-errorlogs option"
>     * "mtest: fix a couple of minor whitespace / typing issues"
>     * "mtest: split TestRun's get_details() to multiple functions" (even if `.detail` is kept as a property)
>     * "mtest: introduce get_test_num_prefix() helper function"
>     * "mtest: remove unused printing arguments from read_decode()" (with fixed and enlarged commit message)
>     * "mtest: print "live" test output when verbose with only 1 test"
>     * "mtest: differentiate stdout and stderr in test log text files" (I think, but please clarify what's the change in the commit message)
>     * "mtest: improve text test log formatting"

These are now the first commits in the change set and can be easily merged via nother RR if seen ready for merging.

> Please reorganize the other commits so that they do not use class methods on ConsoleLogger. 

No more class side methods...

> Each commit message that changes the output should include a before/after comparison, or a description of what is being fixed. 

I tried improving the commit messages all around. Hopefully better now.

> For example:
>     * "mtest: update look and feel of test output" looks like an improvement in general, but I'd like the commit message to explain the organization of the output. In particular it's not clear why you have removed the dashes that delimit the sections. It seems to me that _all_ the sections could use the dashed delimiters.

I thought it didn't need lines in between but decided to try it out. Turns out it was better with some lines between tests. Not between every section though as that would be too noisy and the lines would lose their meaning.

>    * in "mtest: print test details only when needed (not redundant)", there is a complex conditional. 

Added a comment.

> What makes the details redundant in each of the cases?

Hopefully this is now explained better in the commit message (short version: the same info is already shown in the verbose ouput).

>    * in "mtest: use bold with test statuses in verbose mode" I cannot understand if this is useful in general, or mostly because of the earlier patch that removes the dashes

Even with the lines added/restored I still liked those there so we now have both.

>    * "mtest: remove unused arguments from harness' format functions" should mention which commit made the arguments unused
>    * in "mtest: align printed subtest statuses of individiual tests" it's not clear whether this issue is present in current releases of Meson, or it is introduced earlier in the series

Commit messages improved.

> Regarding the big change in the status line, it looks nice in big terminals, but I wonder what it looks like with "ANSI to HTML" converters such as the ones found in Travis or GitLab CI logs. 

Is I said earlier I haven seen any issues in Sourcehut or Gitlab. Don't know how else I could try to verify this? IMO the status line should be just disabled whenever we are not outputting to a proper terminal and this seems to be what happens now (like when doing `meson test | cat`).

> On a small terminal it should degrade to the one used in "quiet" mode, which in turn I see no reason to change from the current one.

The quiet version of the status line is now used with small terminals. I prefer that one over the current one most of all because I think the current one has too may moving parts and is too difficult to read when running many short tests. Also, still not a fan of the emojis but I suppose these are a matter of taste most of all.

Did my best not to break any changes done to mtest lately. Stderr should be printed even when parsing the output and event the stdout can be printed with two verbose options.

trhd:
> > * in "mtest: use bold with test statuses in verbose mode" I cannot understand if this is useful in general, or mostly because of the earlier patch that removes the dashes
> 
> Even with the lines added/restored I still liked those there so we now have both.

Actually...maybe the use of bold with the lines is now a bit excessive after all... Think I'll remove that.

trhd:
I added a couple of changes to improve the output with small terminals (e.g. cut too long lines shorter).

Here is an updated demo:

 - 80x24 terminal: https://asciinema.org/a/th7EMdaePxQeiBOSZQiyMqUu8
 - bigger terminal: https://asciinema.org/a/F99Bf6QxR2BGYPAWGwgNa5ZQi

The tests are green and the three new LGTM warnings are false positives. Overall I'm quite happy with this now.

jpakkane:
Seems fine by me. Anyone else have ideas on the visual changes?

jpakkane:
The unused variable warnings seem to be false positives.

bonzini:
I will check it out tomorrow

jpakkane:
Ping. :)

anarazel:
Haven't debugged it yet, but I just noticed that with these changes cancelling tests with ctrl-c doesn't reliably work anymore.

trhd:
Can you elaborate? I did a quick comparison of ctrl-c beaviour with and without these changes but couldn't notice any differenece.

I did however notice that when when tests are run via `ninja test` (in comparison to `meson test`) a single ctrl-c kills all the tests but this too seemed to be unrelated to these changes.

anarazel:
@trhd 
> Can you elaborate? I did a quick comparison of ctrl-c beaviour with and without these changes but couldn't notice any differenece.
> 
> I did however notice that when when tests are run via `ninja test` (in comparison to `meson test`) a single ctrl-c kills all the tests but this too seemed to be unrelated to these changes.

I unfortunately cannot reproduce it really reliably. It seems to happen most when I cancel a testrun shortly after it started. I just see
```
        postgresql:tap+recovery / recovery/t/023_pitr_prepared_xact.pl                     INTERRUPT        killed by signal 15 SIGTERM
```
at the last line, there's no further output or CPU usage. I need to kill it with -QUIT/KILL, more ctrl-C doesn't do anything.

Using gdb + py-bt:
```
Traceback (most recent call first):
  File "/usr/lib/python3.10/selectors.py", line 469, in select
    fd_event_list = self._selector.poll(timeout, max_ev)
  File "/usr/lib/python3.10/asyncio/base_events.py", line 1860, in _run_once
    event_list = self._selector.select(timeout)
  File "/usr/lib/python3.10/asyncio/base_events.py", line 600, in run_forever
    self._run_once()
  File "/usr/lib/python3.10/asyncio/base_events.py", line 633, in run_until_complete
    self.run_forever()
  File "/home/andres/src/meson/mesonbuild/mtest.py", line 2041, in run_tests
    loop.run_until_complete(self._run_tests(runners))
  File "/home/andres/src/meson/mesonbuild/mtest.py", line 1902, in doit
    self.run_tests(runners)
  File "/home/andres/src/meson/mesonbuild/mtest.py", line 2221, in run
    return th.doit()
  File "/home/andres/src/meson/mesonbuild/mesonmain.py", line 151, in run
  File "/home/andres/src/meson/mesonbuild/mesonmain.py", line 260, in run
    return CommandLineParser().run(args)
  File "/home/andres/src/meson/mesonbuild/mesonmain.py", line 269, in main
  File "/home/andres/src/meson/meson.py", line 29, in <module>
    sys.exit(mesonmain.main())
```

It might be noteworthy that starting with a debug python yields the following deprecation warnings:
```
/home/andres/src/meson/mesonbuild/mtest.py:2040: DeprecationWarning: There is no current event loop
  loop = asyncio.get_event_loop()
  1/234 postgresql:setup / tmp_install                                                                 OK      0.34s
  2/234 postgresql:tap+pg_upgrade / pg_upgrade/t/001_basic.pl                                          OK      0.35s   8 subtests passed
^CWARNING: CTRL-C detected, interrupting postgresql:pg_regress+main / main/pg_regress
/home/andres/src/meson/mesonbuild/mtest.py:1245: DeprecationWarning: The explicit passing of coroutine objects to asyncio.wait() is deprecated since Python 3.8, and scheduled for removal in Python 3.11.
  await asyncio.wait(awaitables,
  3/234 postgresql:pg_regress+main / main/pg_regress                                            INTERRUPT      0.67s   killed by signal 15 SIGTERM
```

The first warning I see in master as well, but the :1245 one I don't.

trhd:
Managed to reproduce the ctrl-c issue using the same tests. Will look into it.

bonzini:
> I did however notice that when when tests are run via ninja test (in comparison to meson test) a single ctrl-c kills all the tests but this too seemed to be unrelated to these changes.

This is unfortunately intentional (#9410 talks about make but ninja is the same).

trhd:
The ctrl-c issue should be fixed now. If you @anarazel can take a look I'd appreciate.

I also noticed, while looking into this, that when running postgres' tests some `postgres` and `krb5kdc` processes gets left behind after the tests get interrupted (escapes the process group or something similar I suppose). This didn't however seem to be related to these chagnes but happened even with vanilla meson.

anarazel:
> The ctrl-c issue should be fixed now. If you @anarazel can take a look I'd appreciate.

Will do, once I'm back at work. Out of curiosity: What was the problem?


> I also noticed, while looking into this, that when running postgres' tests some postgres and krb5kdc processes gets left behind after the tests get interrupted (escapes the process group or something similar I suppose). This didn't however seem to be related to these chagnes but happened even with vanilla meson.

Hm. Odd. Thanks for noticing. Clearly something not quite right yet...  The krb5kdc thing is from some testcase, it's not too surprising that the test doesn't handle interruptions perfectly. But postgres itself should handle it...

trhd:
> > The ctrl-c issue should be fixed now. If you @anarazel can take a look I'd appreciate.

> Will do, once I'm back at work. Out of curiosity: What was the problem?

Essentially the parse_task would not get cancelled properly if the interrupt was received at the right/wrong time.

Something like this:
```
--- a/mesonbuild/mtest.py
+++ b/mesonbuild/mtest.py
@@ -1379,8 +1379,9 @@ class TestSubprocess:
             if self.stde_task:
                 self.stde_task.cancel()
 
     async def wait(self, test: 'TestRun') -> None:
         p = self._process
+        rv = False
 
         self.all_futures.append(asyncio.ensure_future(p.wait()))
         try:
@@ -1388,16 +1389,20 @@ class TestSubprocess:
         except asyncio.TimeoutError:
             test.additional_error += await self._kill() or ''
             test.res = TestResult.TIMEOUT
+            rv = True
         except asyncio.CancelledError:
             # The main loop must have seen Ctrl-C.
             test.additional_error += await self._kill() or ''
             test.res = TestResult.INTERRUPT
+            rv = True
         finally:
             if self.postwait_fn:
                 self.postwait_fn()
 
         test.returncode = p.returncode or 0
 
+        return rv
+
 class SingleTestRunner:
 
     def __init__(self, test: TestSerialisation, env: T.Dict[str, str], name: str,
@@ -1605,10 +1610,13 @@ class SingleTestRunner:
             console_log.print_horizontal_line(harness)
             console_log.print_command_details(get_prefix(), self.runobj)
 
-        await p.wait(self.runobj)
+        err = await p.wait(self.runobj)
 
         if parse_task:
-            await parse_task
+            if err:
+                parse_task.cancel()
+            else:
+                await parse_task
         if stdo_task:
             await stdo_task
         if stde_task:
```

jpakkane:
The lgtm messages are clearly false positives, it does not know how to handle `nonlocal`.

jpakkane:
FWICT all issues have been dealt with. If someone has opposing views, speak up quickly.

bonzini:
In particular, commit 294a33557657c9c523ced1cd6414c431a8c64a12 breaks `--verbose` completely before the new format is introduced. I am okay with pushing up to its parent, i.e. commit 1ede7160e.

jpakkane:
I put those in #10471 to get a clean CI run first.

jpakkane:
And indeed there is a typing failure:

```
run_project_tests.py:1226:120: error: Argument "state" to "submit" of "Executor" has incompatible type "Tuple[List[str], List[str], List[str], List[str], List[str], Backend, List[str], Optional[str]]"; expected "GlobalState"
```

trhd:
Rebased.

There does indeed seem to be 3 commits where the verbose option is broken for parsed tests. Squashing the breaking and fixing commits would fix it but the fixing commit is already quite large... I'll try to have a think about it during the weekend.

eli-schwartz:
(It's more important that #10471 be rebased.)

eli-schwartz:
Please rebase again now that some of the commits were separately merged.

anarazel:
> My main concern is that in verbose mode, subtests are not printed until the whole command has finished running.

Agreed, that's bad. I think there's too much buffering happening with --verbose already, so adding even more... 


> I understand that the new output is easier to read, but it provides less information. If one wants ordered output, meson-logs/test-log.txt and the JUnit XML are the right place to look at.

I think it's *sometimes* a useful output to look at directly in the console. I can see benefit in making this part configurable.


> There are indeed some good things about the new format (though I'm not sure if having double the code for two status lines is any better than the current rotating output), but reporting subtests as they pass is an important feature for projects with very long test suites. Sorry about the delay in review, but this is still a NACK from me. @anarazel what do you think?

Similar. But that might be fixable?

trhd:
I did some updates.

The `verbose` option should now work better for parsed tests also between [32ceb4e](https://github.com/mesonbuild/meson/pull/9930/commits/32ceb4e1c055ab0d4bfb3364536dbf178dc268ba) and [e923319](https://github.com/mesonbuild/meson/pull/9930/commits/e923319d2ae5f62d0c696409f817bb3835425525).

I still don't like mixing/interleaving output from multiple tests but I gave it a shot. Specifying the `verbose` option twice will now make subtest results be printed out in real time while still allowing the default output style to be nice and tidy.

There a demo available here: https://asciinema.org/a/APrfk2l8vTvSo807cTntoUCRQ
Skip to 3:45 to see real time subtest results in action.

Let me know what you think.

jpakkane:
The test failures are due to the current Mac CI image brokenness.

How ready is this otherwise? Are there known issues or is it ready to merge?

jpakkane:
If someone has issues that are still pending, speak up quickly.

bonzini:
I will test it today.

eli-schwartz:
+1, please split this into individually reviewable PRs.

I'm weary of the frustration that this PR has already caused, such as when we accepted a bunch of changes that turned out to be broken, and which was only broken because it assumed that later commits would fix that breakage, except those commits weren't part of the list that got merged.

Quite frankly, I've just been ignoring this PR altogether. I don't trust it to be correct, I have no will to review it, and I live in terror of the day it gets merged but feel guilty at the idea of being a downer so I've been avoiding saying anything. But I think all those problems could be fixed if this was split into easily digestible chunks, so I'm finally going to say something.

PR #9936, opened 2022-02-02
test runner: fix excluding null options

ePirat:
It was impossible to specify null arguments in the exclude array,
while they were passed properly in the code, they were stringified
while the entries in the options array were just completely removed.

To fix this, do not early stringify the argument name and option
together as that would make it impossible to properly match it to
None without introducing a new special case for the string "None".
So instead they are now kept as tuples and only stringified when
its actually needed.

### DIFF
--- a/run_project_tests.py
+++ b/run_project_tests.py
@@ -832,15 +832,15 @@ def load_test_json(t: TestDef, stdout_mandatory: bool) -> T.List[TestDef]:
         t.stdout = stdout
         return [t]
 
-    new_opt_list: T.List[T.List[T.Tuple[str, bool, bool]]]
+    new_opt_list: T.List[T.List[T.Tuple[str, str, bool, bool]]]
 
     # 'matrix; entry is present, so build multiple tests from matrix definition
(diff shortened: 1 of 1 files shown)


jpakkane:
This should have a test to verify that the change is correct.

ePirat:
@jpakkane How do I write a test for the test runner itself?

eli-schwartz:
It would need to be a python unittest, not a project test.

ePirat:
Are there already unit tests for the test runner I can look at as an example?

eli-schwartz:
I don't know why I responded as I did above. I guess I didn't realize that this is about fixing/enhancing the Meson testsuite, not fixing/enhancing the end user program `meson test`.

I don't really think we need to write unittests for our unittests. But we should have at least one unittest that depends on and uses this functionality.

ePirat:
Ping on this as my other MR depends on this and it seems nothing really is blocking this?

eli-schwartz (approved):
Indeed.

PR #9938, opened 2022-02-03
cmake: configure_package_config_file refactor

Tachi107:
I liked so much adding a new kwarg to `write_basic_package_version_file` I thought I would try to also add something to `configure_package_config_file`. Functions taking a `cfg_data` object can generally also take a `dict`, but cmake's `configure_package_config_file` was an exception, so I added the possibility to also pass a `dict` to it.

Following @eli-schwartz's suggestion in https://github.com/mesonbuild/meson/pull/9916#discussion_r795921708 I also refactored a bit the function to use `@typed_kwargs`, but in doing so I encountered a couple of issues I don't know how to solve... The first is a regression, as now the `input` kwarg complains when passing the result of `files()`, saying that it expects a `File` or a `str` while an `array[File]` has been passed in. I think this could be worked around by saying that the function can take a list of `File`s, but it's not really a solution. The second one is less important, and is the fact that I don't know how to put the default value of `install_dir` in `@typed_kwargs`, so I have to still keep an `if isinstance` in the function.

If someone would like to help me a bit understand what's going on it would be great :)

### DIFF
--- a/mesonbuild/modules/cmake.py
+++ b/mesonbuild/modules/cmake.py
@@ -37,16 +37,26 @@ from ..interpreterbase import (
 
     typed_kwargs,
     KwargInfo,
+    ContainerTypeInfo,
 )
 
 if T.TYPE_CHECKING:
-    class WriteBasicPackageVersionFile(T.TypedDict):
+    from typing_extensions import TypedDict
+
+    class WriteBasicPackageVersionFile(TypedDict):
 
         arch_independent: bool
         compatibility: str
         install_dir: T.Optional[str]
         name: str
         version: str
+
+    class ConfigurePackageConfigFile(TypedDict):
+
+        configuration: build.ConfigurationData
+        input: T.Union[str, mesonlib.File]
+        install_dir: T.Optional[str]
+        name: str
 
 COMPATIBILITIES = ['AnyNewerVersion', 'SameMajorVersion', 'SameMinorVersion', 'ExactVersion']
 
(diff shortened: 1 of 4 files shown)


eli-schwartz:
> now the `input` kwarg complains when passing the result of `files()`, saying that it expects a `File` or a `str` while an `array[File]` has been passed in. I think this could be worked around by saying that the function can take a list of `File`s, but it's not really a solution.

You need a custom `convertor=` and `validator=`, the validator should check that if it is a list, it is of length 1 and that `x[0]` is a File. The convertor should return `x[0] if isinstance(x, list) else x`.

> The second one is less important, and is the fact that I don't know how to put the default value of `install_dir` in `@typed_kwargs`, so I have to still keep an `if isinstance` in the function.

That's completely okay and what we typically do elsewhere in the same situation.

mesonbuild/modules/cmake.py
```diff
@@ -334,42 +341,31 @@ def create_package_file(self, infile, outfile, PACKAGE_RELATIVE_PATH, extra, con
...
         name = kwargs['name']
 
         (ofile_path, ofile_fname) = os.path.split(os.path.join(state.subdir, f'{name}Config.cmake'))
         ofile_abs = os.path.join(state.environment.build_dir, ofile_path, ofile_fname)
 
-        install_dir = kwargs.get('install_dir', os.path.join(state.environment.coredata.get_option(mesonlib.OptionKey('libdir')), 'cmake', name))
-        if not isinstance(install_dir, str):
-            raise mesonlib.MesonException('"install_dir" must be a string.')
+        install_dir = kwargs['install_dir']
+        if isinstance(install_dir, NoneType):
```
eli-schwartz, line 365: `if install_dir is None`

eli-schwartz (changes requested):

mesonbuild/modules/cmake.py
```diff
@@ -48,6 +49,13 @@ class WriteBasicPackageVersionFile(T.TypedDict):
         name: str
         version: str
 
+    class ConfigurePackageConfigFile(T.TypedDict):
```
dcbaker, line 52: `TypedDict` isn't in typing until python 3.8, so I think we still need to use `typing_extensions.TypedDict`

dcbaker (changes requested):
I have one small nit, otherwise this looks good to me.

mesonbuild/modules/cmake.py
Tachi107: Oh ok, I thought that the `if T.TYPE_CHECKING` was enough, as the tests are passing in the Ubuntu Bionic. Should I change that also in the `WriteBasicPackageVersionFile` class introduced in #9916?

dcbaker: It will work if you run with mypy based on python 3.8 or later. I don’t think we’ve ever had a clear policy on whether we support all versions with mypy we do with in general or not.

Tachi107: Done :)

dcbaker (approved):

dcbaker:
Looks good to me, thanks. @eli-schwartz, look good to you?

mesonbuild/modules/cmake.py
```diff
@@ -334,42 +344,34 @@ def create_package_file(self, infile, outfile, PACKAGE_RELATIVE_PATH, extra, con
...
-            raise mesonlib.MesonException('configure_package_config_file takes only keyword arguments.')
-
-        if 'input' not in kwargs:
-            raise mesonlib.MesonException('configure_package_config_file requires "input" keyword.')
+    @noPosargs
+    @typed_kwargs(
+        'cmake.configure_package_config_file',
+        KwargInfo('configuration', build.ConfigurationData, required=True),
+        KwargInfo('input',
+                  (str, mesonlib.File, ContainerTypeInfo(list, (str, mesonlib.File))), required=True,
```
eli-schwartz, line 352:
```suggestion
                  (str, mesonlib.File, ContainerTypeInfo(list, mesonlib.File)), required=True,
```

I think there's probably no value in letting people pass `input: ['string']` simply because as an internal implementation detail we allow `input: files('string')` and don't force the user to de-arrayize it with `input: files('string')[0]`

eli-schwartz (changes requested):
We currently test this function twice:
- `test cases/cmake/20 cmake file/`
- `test cases/cmake/22 cmake module/`

We should change one of those to use a dict as configuration.

Tachi107:
@eli-schwartz done :)

eli-schwartz:
Did you read my mind -- you made that last push just as I was trying to submit a review comment about it. :p

Tachi107:
@eli-schwartz sorry for the ping, but I suspect you forgot about this PR (if not, sorry!)

eli-schwartz:
Thanks for the ping, I did indeed forget. :(

This looks great now, but I think maybe we want a release notes snippet in `docs/markdown/snippets/` in addition to the main docs update.

Tachi107:
I didn't include it as this change is really small, but I can add it if you want

jpakkane:
As this is new functionality it needs a release note snippet.

Tachi107:
Snippet added :)

dcbaker:
@eli-schwartz are you happy with this?

eli-schwartz (approved):

PR #9488, opened 2021-10-29
More porting over manual warnings to FeatureDeprecated

eli-schwartz:

PR #9940, opened 2022-02-03
interpreter: deprecate using an array for project(license:)

dcbaker:
It's ambiguous as to whether the array means AND or OR, and ambiguity in
software licenses is *not* a good thing. Further, SPDX (which we
generally recommend) is an ISO standard and well understood outside of
Meson (including in SBOM software). Deprecate passing and array and
suggest using an SDPX expression instead.


xclaesse:
LGTM

docs/markdown/snippets/license_as_array.md
```diff
@@ -0,0 +1,15 @@
...
+
+Meson has traditionally supported using either a string or an array of strings
+as the `license` argument to `project`. More recently it has been strongly
+suggested to use an SPDX license expression string instead. SPDX is now an ISO
+standard, and is well understood outside of Meson. When used in a Software Bill
+of Materials it is natively understood by many tools. An array of licenses is
+ambiguous, does `['Apache-2.0', 'GPL-2.0-only']` mean `Apache-2.0 AND GPL-2.0-only`
+or `Apache-2.0 OR GPL-2.0-only`? These mean very different things, as the first
+is widely believed to be incompatible, and the latter is widely understood to
+mean "Apache-2.0 in general, but for you may use it as GPL-2.0 for projects
```
eli-schwartz, line 11:
```suggestion
mean "Apache-2.0 in general, but you may use it as GPL-2.0 for projects
```

mesonbuild/interpreter/interpreter.py
```diff
@@ -1135,6 +1135,9 @@ def func_project(self, node: mparser.FunctionNode, args: T.Tuple[str, T.List[str
         if self.build.project_version is None:
             self.build.project_version = self.project_version
         proj_license = kwargs['license']
+        if len(proj_license) > 1:
+            FeatureDeprecated.single_use('license as an array of strings', '0.62.0', self.subproject,
```
eli-schwartz, line 1139: I wonder if we actually need a version-gated feature? We recommend against it purely because it is "a bad idea" and the alternative is to do something that has also worked since this kwarg was added. So there's no reason to only recommend against it once you bump your minimum meson.build high enough.

dcbaker: You're right, the way this is written is bad. I've changed it so that the function takes a `str | list[str]`, and that passing `list[str]` is deprecated, and the warning is gated on that.

eli-schwartz: While this does fine tune the check to also trigger on `license: ['GPL-2.0-only']` it doesn't really answer my question, which is whether we should use `mlog.deprecation` and print warnings even for projects that support Meson 0.45, instead of using FeatureDeprecated and only warning people away from this once they bump `meson_version: '>=0.62'`.

Again, the point of a FeatureDeprecated is typically because we change how things work and add new features which replace old and badly designed features, so as a result a FeatureDeprecated does not warn you if you cannot migrate to the new method due to your requirement to support old versions of Meson.

dcbaker: I've made a change to add a warning in the even that you're using a version < 0.62 in addition to the deprecation warning (either the FeatureDeprecated or the separate warning). The FeatureDeprecated is still useful as passing an array should go away, but you're right, encouraging the use of an SPDX license expression even in older versions is desirable

eli-schwartz: This seems like a very complicated way to add two distinct warnings for one thing, when passing as an array is *still* not something that needs to version-gated the warning.

Deprecating a feature with a replacement that has been available for as long as the feature itself, doesn't need a version gate, whether that feature is "multiple license strings or "passing as an array".

We can just tell people using Meson 0.45 "you know, it's kind of bad to not use SPDX, and since you're going to use SPDX, it's kind of silly to pass a single string wrapped in an array, and also this still works on 0.45 *anyway* (and quite frankly it works on like 0.25 also)".

dcbaker: I'm strongly against removing the `FeatureDeprecated`, because it provides a clean way for projects receiving bug reports about the deprecation to know exactly what's happening, and when it was deprecated. Having been on the end of CMake policy bug reports/PRs that don't include that information, I think this makes for good user experience. It's immaterial that the replacement has been available since the beginning, 0.62 is when we declared "this is going away".

The other problem we have, is that we don't (and can't easily without external deps) verify that the project license is an SPDX expression, only that it's an array or a single string. That string could easily be `LOLZ!1!`, and we can't say with certainty "this is not SPDX!". But we can say that `['GPL', 'MIT']` is ambiguous, knock that off.

eli-schwartz: That would imply we should extend FeatureCheck with a `.single_use(..., always_warn=True)` option, and just emit the FeatureDeprecated but ensure it is always logged as a full warning.

mesonbuild/interpreter/interpreter.py
```diff
@@ -1135,6 +1135,17 @@ def func_project(self, node: mparser.FunctionNode, args: T.Tuple[str, T.List[str
...
             self.build.project_version = self.project_version
         proj_license = kwargs['license']
+        if isinstance(proj_license, list):
+            triggered = FeatureDeprecated.single_use(
+                'project license as an array of strings', '0.62.0', self.subproject,
+                'Use an SPDX license expression with AND and/or OR instead', node)
+            if not triggered and len(proj_license) > 1:
+                mlog.warning(
+                    'Passing an array of licenses to project() is ambiguous, '
+                    'consider using an SPDX license expression string instead.',
```
xclaesse, line 1145: I don't like the warning message being repeated, why not passing an new arg to `FeatureDeprecated.single_use()` like `always_warn=True`? Note that Meson already print a NOTICE in the summary about future deprecations, that could be enough?

eli-schwartz: I already had this same discussion in the previous review on this PR, although my opinion there was that we should only use mlog.warning and not use FeatureDeprecated at all.

xclaesse: It's in the review comment above indeed. Closing this discussion.

mesonbuild/interpreter/interpreter.py
xclaesse: @dcbaker could you update your PR to add FeatureDeprecated.single_use(..., always_warn=True) please?

dcbaker:
@xclaesse, @eli-schwartz: I've rebased this, added the requested `always_warn` flag, added a test, and removed the patch that makes `.use` return a boolean, as it no longer makes sense. I also noticed a stale FIXME comment that it looks like @eli-schwartz resolved, so I removed that too.

mesonbuild/interpreterbase/decorators.py
```diff
@@ -726,16 +741,26 @@ def get_warning_str_prefix(tv: str) -> str:
...
+                f"'{self.feature_version}',",
+                'and can safely be replaced without causing a regression.'
+            ]
+        else:
+            args = [
+                'Project targets', f"'{tv}'",
+                'but uses feature deprecated since',
+                f"'{self.feature_version}':",
+                f'{self.feature_name}.',
+            ]
```
xclaesse, line 760: Not sure why we need 2 different messages here, what's wrong with the original one? Also it probably should be fatal when targeting Meson >=0.63 but since that's inside `project()` I don't think the target meson version is set already, is it?

Sorry for nitpicking, maybe I'm just overthinking this simple PR tbh.

dcbaker: I don't think the normal message makes sense in this case, since you'd get a warning like: `Project is targing >= 0.52 but using a feature deprecated since 0.63.` That seems confusing to me.

eli-schwartz: I agree that the dedicated warning message makes sense.

However I also agree with @xclaesse that it seems weird to mark it as non-fatal, because if we always warn about it and can always use the better non-deprecated version of it, then we also do want it to trigger --fatal-meson-warnings.

mesonbuild/interpreterbase/decorators.py
```diff
@@ -631,8 +631,6 @@ def use(self, subproject: 'SubProject', location: T.Optional['mparser.BaseNode']
         feature_key = (self.feature_name, location)
         if feature_key in register[self.feature_version]:
             # Don't warn about the same feature multiple times
-            # FIXME: This is needed to prevent duplicate warnings, but also
-            # means we won't warn about a feature used in multiple places.
```
eli-schwartz, line 622: Good catch, but the actual fix was in 7a033c78870e55b00d10a6499b40ad98b88a0509.

eli-schwartz: e.g. the discussion at https://github.com/mesonbuild/meson/pull/9488#discussion_r739301985

mesonbuild/interpreterbase/decorators.py
```diff
@@ -702,7 +719,7 @@ def log_usage_warning(self, tv: str, location: T.Optional['mparser.BaseNode']) -
         ]
         if self.extra_message:
             args.append(self.extra_message)
-        mlog.warning(*args, location=location)
+        mlog.warning(*args, location=location, fatal=not always)
```
eli-schwartz, line 720: This should just be dropped. The default is True, but we're setting it to False *iff* it is an `always_warn`.

This means that a deprecation for something that is always deprecated and always wrong and always should be replaced... does NOT trigger --fatal-meson-warnings, even though it really should.

eli-schwartz (changes requested):

PR #9947, opened 2022-02-04
Drop emacs meson-mode

elmarco:
The one on MELPA is way more advanced: https://melpa.org/#/meson-mode
https://github.com/wentasah/meson-mode


eli-schwartz:
Would it be worth merging improvements instead?

elmarco:
> Would it be worth merging improvements instead?

Merging isn't really the appropriate word here, it's a very different best.

@wentasah would you like to move the emacs mode maintainance back to meson project?

dcbaker:
I think this is the right thing to do even if we were to move the meson-mode code into the meson project. If we move it back it should probably be treated like the vscode extension and live in its own repo under the project.

dcbaker (approved):

wentasah:
> @wentasah would you like to move the emacs mode maintainance back to meson project?

Hi, I wouldn't mind moving meson-mode under the mesonbuild organization. In the best case, it will get more visibility and attract more contributions. In the worst case, it will stay the same as now. Is there anything that must be changed for this move to happen, e.g. licensing, maintenance process, etc.?

dcbaker:
Up to @jpakkane  on that, but I’ve done the transfer before and can help with that process if we decide to do it

PR #9949, opened 2022-02-04
devenv: support bash under MSYS2 by default

lazka:
Currently it tries to run "cmd" by default in a MSYS2 bash.
Passing "bash" doesn't work since that defaults to WSL and
one has to pass an absolute path to bash instead, and even then
one misses out on the PS1 override.

In case $SHELL is set and the contained path exists prefer it even if
we are on Windows.

To make the PS1 override work we can't use Unix paths in Python since
we might be on Windows, so move the .bashrc check into the temporary
bash script itself.

This makes "meson devenv" work the same under MSYS2 as on Linux.


lazka:
Thanks!

PR #9950, opened 2022-02-04
Use typed_*_args for build_targets

dcbaker:
This is still a Work in Progress, but I'm putting it here in case others want to add things, or make comments (and to see the work underway). This is *mostly* working for the main interpreter, but there are issues with the introspection code that isn't working yet.


mesonbuild/mesonlib/universal.py
```diff
@@ -2232,3 +2233,17 @@ def is_project(self) -> bool:
...
+
+    :param iter: The iterable to search
+    :param predicate: A finding function that takes an element from the iterable
+        and returns True if found, otherwise False
+    :return: The first found element, or None if it is not found
+    """
+    for i in iter:
+        if predicate(i):
+            return i
+    return None
```
dnicolodi, line 2249: `next(i for i in iter if predicate(i))` or `next(filter(predicate, iter))` ?

dcbaker: If there is no element though you get a `StopIteration` exception and have to catch it, so you actually end up with
```python
try:
    return next(i for i in iter if predicate(i))
except StopIteration:
    return None
````
Which doesn't seem any cleaner to me

dnicolodi: That's why `next()` has a `default` argument: `return next(i for i in iter if predicate(i), None)` 😃

eli-schwartz: Is the goal here to do things in the least number of lines, or to be readable?

dnicolodi: My brain takes less time to parse the one liner than the for loop, but I guess it is subjective.

eli-schwartz: The loop or the one-liner is the entirety of the function body, and it's a function to give the whole thing an interface that peoples' brains can parse in the least amount of time ("`first(foo, func)` gets the first thing in foo where func is true").

Accepting the usefulness of such a helper function, its main purpose is that people don't look at the loop at all, merely rely on its interface. That being the case, IMHO we should rely on whichever version of the function body guarantees understanding... and for me at least, one-liners are easier to read, but harder to be confident that I've read *correctly*.

eli-schwartz:
A bunch of these commits now landed separately, time to rebase.

Since the current PR is 47 commits, covering 37 files and 4,000 lines of changes -- and the rebase will only make it a little better -- I would *really* appreciate if some of these changes could be split out into their own series to make review easier. :)

dcbaker:
Yeah, I’m working on splitting out the first 10 or so patches that cover most of the interpreter changes for a first batch

dcbaker:
This is very out of date, most of this has landed in other forms

PR #9958, opened 2022-02-07
add FeatureNew check for compiler.* methods with dependency on internal dep

### LINKED ISSUE #9957: Missing backward compat warning when passing declare_dependency() to compiler
Hi,

Starting in 0.57 one can pass declare_dependency() to compiler.has_function() etc:
```
[](https://mesonbuild.com/Release-notes-for-0-57-0.html#passing-internal-dependencies-to-the-compiler-object)Passing internal dependencies to the compiler object

Methods on the compiler object (such as compiles, links, has_header) can be passed dependencies returned by declare_dependency, as long as they only specify compiler/linker arguments or other dependencies that satisfy the same requirements.
```
Which is great!

However, there is no warning about doing so even when `meson_version: '>=0.54',` or such. It's pretty easy to accidentally break even just slightly older operating systems like Debian Bullseye. 

Regards,

Andres

eli-schwartz:
In commit 0deab2ee9efc2ffe9e43f2787611e34656e6a304 we added the ability to pass a declare_dependency() to any compiler method that accepts "dependencies", but we never marked the version it is available since.

Fixes #9957

### DIFF
--- a/mesonbuild/interpreter/compiler.py
+++ b/mesonbuild/interpreter/compiler.py
@@ -189,8 +189,7 @@ class CompilerHolder(ObjectHolder['Compiler']):
     def compiler(self) -> 'Compiler':
         return self.held_object
 
-    @staticmethod
-    def _dep_msg(deps: T.List['dependencies.Dependency'], endl: str) -> str:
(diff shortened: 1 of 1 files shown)


dcbaker (approved):

anarazel:
Thanks!

PR #9960, opened 2022-02-07
gnome: Build documentation by default

benzea:
Without this, we have the problem that the documentation is not build by
the "all" target. This in turn means that it will not have been build
during "install".

Sometimes, this works fine, i.e. meson will run the gtk-doc command just
fine in the installation step. However, the result is that any if there
are generated input files for gtk-doc, then these are not build in
advance.

We do have the PROJECT-doc target, which is sufficient when run. As
such, this patch changes things so that we have a dependncy such as

  install -> all -> PROJECT-doc -> generated input for gnome.gtkdoc

an alternative to this could be doing a more direct

  install -> PROJECT-doc -> generated input for gnome.gtkdoc

dependency chain. Doing that would fix the problem without also
"regressing" by causing the documentation to be re-build every single
time the binaries are build.

### DIFF
--- a/mesonbuild/modules/gnome.py
+++ b/mesonbuild/modules/gnome.py
@@ -1461,6 +1461,7 @@ class GnomeModule(ExtensionModule):
             [],
             [f'{modulename}-decl.txt'],
             build_always_stale=True,
+            build_by_default=True,
             extra_depends=depends,
         )
         alias_target = build.AliasTarget(targetname, [custom_target], state.subdir, state.subproject)


benzea:
So, yes, I would fully understand if this is *not* acceptable. But, I do think this is a problem, and I feel the patch shows nicely why this is problematic.

xclaesse:
IIRC the reason it's not built by default is because it's super slow and you don't want to rebuild doc each time you compile.

benzea:
Yeah, and there is a point in that (and actually plenty of duplicates for this issue).

But with the current situation we get the worse of both worlds.
 1. The `%s-doc` target is never build automatically
 2. The `install` target still *builds* the documentation

I think we have the following possibilities:
 1. Throw an informative error and force users to run `%s-doc` manually before doing `install`
 2. Keep trying to build it during install, but if it fails, show an informative warning that building `%s-doc` will fix it
 3. Add `%s-doc` to the `install` target (ugly)
 4. Add `%s-doc` to the `all` target (slow)

eli-schwartz:
Duplicate of #9625

eli-schwartz:
In the linked ticket there is some discussion about the issues with just marking it as built by default. The other PR also has the distinction of avoiding to rebuild at install, IIRC.

PR #9967, opened 2022-02-08
ninjabackend: fix rust program names with dashes

alyssais:
This substitution matches the [behaviour of rustc][1] when inferring
crate name based on file name.

[1]: https://github.com/rust-lang/rust/tree/4e8fb743ccbec27344b2dd42de7057f41d4ebfdd/compiler/rustc_session/src/output.rs#L88

### DIFF
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -1709,7 +1709,8 @@ class NinjaBackend(backends.Backend):
             args.extend(rustc.get_linker_always_args())
 
         args += self.generate_basic_compiler_args(target, rustc, False)
-        args += ['--crate-name', target.name]
+        # This matches rustc's default behavior.
+        args += ['--crate-name', target.name.replace('-', '_')]
         depfile = os.path.join(target.subdir, target.name + '.d')
         args += ['--emit', f'dep-info={depfile}', '--emit', 'link']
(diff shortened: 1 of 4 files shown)


test cases/rust/1 basic/meson.build
```diff
@@ -1,6 +1,6 @@
 project('rustprog', 'rust')
 
-e = executable('program', 'prog.rs',
+e = executable('rust-program', 'prog.rs',
```
dcbaker, line 3: I haven’t looked for sure, but you probably need to update the test.json’s installed files list

mesonbuild/backend/ninjabackend.py
```diff
@@ -1709,7 +1709,7 @@ def generate_rust_target(self, target: build.BuildTarget) -> None:
             args.extend(rustc.get_linker_always_args())
 
         args += self.generate_basic_compiler_args(target, rustc, False)
-        args += ['--crate-name', target.name]
+        args += ['--crate-name', target.name.replace('-', '_')]
```
dcbaker, line 1712: This deserves a comment that we’re matching rustic behavior here. Otherwise this looks good to me, thinks!

dcbaker (changes requested):

dcbaker (approved):
Looks good, thanks

PR #9969, opened 2022-02-09
Properly error out when cross file is missing a compiler.

jpakkane:


mesonbuild/compilers/detect.py
```diff
@@ -240,8 +240,9 @@ def _get_compilers(env: 'Environment', lang: str, for_machine: MachineChoice) ->
         # Return value has to be a list of compiler 'choices'
         compilers = [comp]
     else:
-        if not env.machines.matches_build_machine(for_machine):
-            raise EnvironmentException(f'{lang!r} compiler binary not defined in cross or native file')
+        # Cross compilers must NEVER EVER be taken from envvars.
```
eli-schwartz, line 243: @Ericson2314 @ePirat seems like we're about to regress again, sigh.

eli-schwartz (changes requested):
I completely disagree with this change, it literally exists just to make absolutely no difference to anyone, except for the people whose lives are thus made harder with zero rationale.

dcbaker:
Hard NAK, distros rely on this, and every other major build system implants this, they’ll just end up reverting it because that’s easier than work around it.

mesonbuild/compilers/detect.py
Ericson2314: Yeah thanks for letting me know. We have been using env vars with Meson quite successfully and necessarily with NixOS/Nixpkgs for many releases of Meson now, and I am sure others are too.

Ericson2314:
The *only* people who might conceivable benefit are those the start with their default native build enviroment (as a black box they don't know quite what it contains), and then add in some cross compilers (maybe installed in an ad-hoc way) and don't want to worry about whatever environment variables might have been set by the default native build environment.

I get it; this is perhaps everyone's *first* time experience trying to cross compile something. But anyone managing builds seriously quickly learns you shouldn't just "throw more stuff on your native environment" and call it finished, but rather construct the environment you need for the platform you are building for from first principles, native or cross.

It is much better to meet the needs of these serious cross users, than try to cushion the blow for the cross beginners --- beginners who are going to be beset with build failures anyways before long if they try to hastily throw together a build environment that way.

jpakkane:
Since the very beginning the design has been that cross executables ONLY come from the cross file. Never, ever, ever, ever from envvars. The fact that they did was a bug. The code I changed very explicitly tried to do this and got it wrong.

Ericson2314:
@jpakkane I recall we all talked about this many times. It's on the docs too https://mesonbuild.com/Reference-tables.html#environment-variables-per-machine

Your example that is supposed to be so egregious doesn't really rattle me. In general, we assume native, and your empty cross file doesn't override any assumptions. Why should it do anything different?

If is anything confusing, it's the basic flow (at least I checked) of both guessing compilers based on tentative platform description, and guessing platform description based on compiler choice.

In your example, because there there is no `[machine]`, it assumes both are equal, and then (no env vars needed!) it is happy to use the default logic to find the compiler for the build and host machine the same way. If, say, we could specify a default tool prefix in the cross file, then we could at least force ourselves only to look up so-prefixed tools if there is no env vars or `[binaries]`. That would allow making meson to fail to find a host compiler as desired.

jpakkane:
> it is happy to use the default logic to find the compiler for the build and host machine the same way

It did not use to do that. When I originally wrote that I was _very_ strict about it never accepting cross properties from anything except the cross file. Sadly this got broken by some refactoring.

dcbaker:
This was not "broken by some refactoring" It was an explicit decision to make our usage of CC, CFLAGS, etc match what everyone else does, because implementing a standard interface in a non-standard way is worse than useless.

jpakkane:
>  recall we all talked about this many times. It's on the docs too https://mesonbuild.com/Reference-tables.html#environment-variables-per-machine

There are two different cases here.

The first one is that if `CC_FOR_BUILD` is defined, then that is used for the native compiler. This is totally fine.

The second one is that then `CC` is used for cross compilation. This is **not** fine. It is wrong. I never approved this, only the first one of the two. 

Also:

> distros rely on this

Debian (and by extension any distro based on it) does not. It has used cross files exclusively since day one. One of the main reasons I became the maintainer of the Meson package in Debian was to ensure that they would actually do the right thing and not write a ton of custom monstrosity patches to make it pretend to be autotools. 

There has been some desire to get the `debcrossgen` tool to Meson upstream so everyone can do something like `meson env2cross` to convert their env var mishmash soup into a cross file.

ePirat:
@jpakkane I actually have a similar script for VLCs contrib system too that produces a crossfile from env vars, so having something like that upstream would definitely be useful

Ericson2314:
If you must kill env vars, yes let's upstream that tool. Let's also just remove env vars from Meson *entirely*. `env2cross` should be called `env2platform` (or similar) so it works with native and cross files alike, which have the same structure. The debian packaging becomes simpler too, because it's the same code paths in both the native and cross case.

jpakkane:
Voice your comments in #10018. Thanks.

PR #9970, opened 2022-02-09
pypi: include a ninja extra

henryiii:
This will really help with pipx run, as `pipx run meson[ninja]` will now work on any system with pipx 0.17 or newer, no dependencies required. This now includes GitHub Actions, which just updated to pipx 1.0.0. Pipx run lets you use recent (always <1 week old) versions of anything on PyPI.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -40,6 +40,8 @@ console_scripts =
   meson = mesonbuild.mesonmain:main
 
 [options.extras_require]
+ninja =
+  ninja>=1.8.2
 progress =
   tqdm
 typing =


eli-schwartz:
Hmm, I think the original rationale of leaving this out is because `pip install meson ninja` and `pip install meson[ninja]` are basically the same and don't really win you anything by including the extra.

The fact that the latter apparently integrates with pipx is pretty neat, and a new fact to me. I'd say this is good grounds for adding that extra in after all.

henryiii:
This also documents the minimum requirement nicely (which I took from the error message). It's also nice for the future if you want to move the default support to a new packaging of ninja[^1] or add something - `setuptools_scm[toml]` was identical to `setuptools_scm toml` until it moved to `tomli`, then it wasn't, and only the people using `setuptools_scm[toml]` still got the right thing.

[^1]: I am a maintainer for the ninja PyPI package, so you shouldn't ever need to do this. :)

PR #9972, opened 2022-02-09
setup.cfg: update classifier, drop deprecated

henryiii:
This adds 3.10 (didn't check to see if it was tested on 3.10, that should be done, but I'm using it on 3.10).

setup_requires is deprecated and should be producing warnings on modern setuptools, and never worked for setuptools anyway - setuptools can't update itself, and setuptools is what is reading this anyway!

FYI, you can keep you setup.cfg classifiers up to date with setup-cfg-fmt and pre-commit, see https://scikit-hep.org/developer for that and other recommendations for modern Python packaging.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,14 +26,13 @@ classifiers =
   Programming Language :: Python :: 3.7
   Programming Language :: Python :: 3.8
   Programming Language :: Python :: 3.9
+  Programming Language :: Python :: 3.10
   Topic :: Software Development :: Build Tools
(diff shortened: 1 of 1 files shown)


eli-schwartz:
Indeed, the Arch workflow tests against the system python 3.10, so it's known to work, we just forgot to tag that in the PyPI classifiers.

The setup_requires on setuptools is... very confusing to me, it seems to have come in with cd79ec2839ddef766b42e0653f8be2199b723e77 which made the argument that we need it at runtime, so we should document we need it at build time *and then not say anything about runtime* because... Debian pkg_resources isn't part of setuptools.

I don't get it :D but either way we don't depend on pkg_resources now either.

Issue #7632, opened 2020-08-24
[meson|cmake] Generated cmake configs are broken


Neumann-A:
- do not work on windows `_lib_suffix` is not set correctly 
- are not separated by configuration making integration into vcpkg impossible ...... 
- this means general setup must be separated and put into `-config/-target.cmake` while the configuration aware targets belong into `-target-<config>.cmake`. CMake will select the configuration depending on `CMAKE_BUILD_TYPE` or a set mapping

cc @mensinda

mensinda:
Is there some official CMake specification on how these files should be generated? I was looking through https://cmake.org/cmake/help/latest/command/find_package.html and there doesn't seem to be anything regarding configurations in there.

What is the exact format/layout vcpkg expects?

Neumann-A:
Easy layout: (GTest as an example)
The first two are generally generated by https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html
[GTestConfig.cmake.txt](https://github.com/mesonbuild/meson/files/5329648/GTestConfig.cmake.txt)
Contains required information about the package. Normally this is generated by the project itself and pulls in required imported/external targets via `find_dependency`. It needs to include the generated `<package>Targets.cmake`
[GTestConfigVersion.cmake.txt](https://github.com/mesonbuild/meson/files/5329649/GTestConfigVersion.cmake.txt)
Probably generated by `write_basic_package_version_file`. contains required version/architecture information for cmake

The next three files are generated by the `install(EXPORT)` command (https://cmake.org/cmake/help/latest/command/install.html#export) and are the really important ones
[GTestTargets.cmake.txt](https://github.com/mesonbuild/meson/files/5329650/GTestTargets.cmake.txt)
this contains all exported target as import libraries and generally setups the imported target. In general it setups:
`INTERFACE_INCLUDE_DIRECTORIES`, `INTERFACE_LINK_LIBRARIES` etc.... (important those properties will have configuration dependent values using generator expressions). It does not contain the information where the libraries are located!
[GTestTargets-debug.cmake.txt](https://github.com/mesonbuild/meson/files/5329651/GTestTargets-debug.cmake.txt)
[GTestTargets-release.cmake.txt](https://github.com/mesonbuild/meson/files/5329652/GTestTargets-release.cmake.txt)
these files actually contain the required information locate the library and link against it in a configuration dependent manner. 

[...]

PR #9978, opened 2022-02-09
cmake: add generate_export()

Tachi107:
This is a wip branch where I'm trying to implement what I described in #7632, i.e. a function that generates files comparable to CMake's `install(EXPORT)`. 

When I say that this is a work in progress I really mean it. The code is as bad as my Python skills and currently only works for for very simple cases (like generating exports of a shared library), but I'm sharing it here so that I can get some feedback :)

One major thing I have to implement is extracting private dependencies of the libraries passed to the function (I need to do something similar to what is done in `pkgconfig.generate()` for `Libs.private`), so I guess I'll have to figure out how the pkgconfig module does it.

Also, the generated files are currently written to the build dir with `open().write()` simply because I don't know what's the proper way to do it, and the same goes for installing the files, but these are minor issues that I'll be able to solve by looking at existing code that already generates and installs files.


mesonbuild/modules/cmake.py
```diff
@@ -416,5 +542,91 @@ def subproject(self, state, args, kwargs):
...
+            name = deps[0].libraries[0].name
+
+        includedir = coredata.get_option(mesonlib.OptionKey('includedir'))
+        include_dirs = f'"${{_IMPORT_PREFIX}}/{includedir}" '
+        for subdir in kwargs['subdirs']:
+            include_dirs += f'"${{_IMPORT_PREFIX}}/{includedir}/{subdir}" '
+        include_dirs = include_dirs.strip()
+
+        targets_file = TARGETS_INIT.replace('@expected_targets@', expected_targets)
+        targets_file = targets_file.replace('@prefix@', coredata.get_option(mesonlib.OptionKey('prefix')))
```
Tachi107, line 600: This returns the install prefix with the "correct" path separators depending on the platform (`\` on Windows), but CMake uses `/` everywhere. It is possible to always get the path with unix file separators? Or am I forced to convert `\` to `/` when building on Windows?

mesonbuild/modules/cmake.py
```diff
@@ -416,5 +542,91 @@ def subproject(self, state, args, kwargs):
...
+            targets_file = targets_file.replace('@compile_options@', compile_options[i])
+            targets_file = targets_file.replace('@include_dirs@', include_dirs)
+            #targets_file = targets_file.replace('@libraries@', '')
+            targets_file = targets_file.replace('@link_options@', link_options[i])
+
+            targets_config_file += TARGETS_CONFIG_TARGET.replace('@namespace@', namespace)
+            targets_config_file = targets_config_file.replace('@target_name@', target.name)
+            targets_config_file = targets_config_file.replace('@buildtype@', buildtype)
+            targets_config_file = targets_config_file.replace('@libdir@', coredata.get_option(mesonlib.OptionKey('libdir')))
+            targets_config_file = targets_config_file.replace('@lib_file_name@', target.filename)
```
Tachi107, line 621: When building a shared library, this will generate an imported location pointing to the versioned library (e.g. `libname.so.1.2`). This matches CMake's behaviour, but it is wrong as users should link against the unversioned symlink. Is there a way to get the unversioned file name from the library?

eli-schwartz: Yeah, you can generate `-lfoo` link arguments just like pkg-config does. :)

Tachi107: I can't, I need the actual file name of the library because CMake needs the full path to the compiled library that it needs to link against to be set in the `IMPORTED_PATH_<config>` property
https://github.com/mesonbuild/meson/blob/8c5e76eee00bb6dddc5429a49ce8ca6aefd29563/mesonbuild/modules/cmake.py#L201

Tachi107:
Yeah maybe doing something like that would be easier, but I don't think that the current approach is too complicated either, it is just a matter of extracting link flags and compile args and putting them in the appropriate place.

It is gets a bit complicated because obviously CMake's way of doing so is extremely convoluted (multiple libraries in a single export set, a set is composed of several files, you can have n files for n different build configurations...), but the worst thing is that we have to account for them in order to generate .cmake files useful for the existing CMake ecosystem (as I really want to fix #7632 - what would be the point otherwise?)

A lot of stuff is copy-pasted because it was a quick attempt at getting something working, but it doesn't have to be the final solution. Some stuff would need to be copied over from classic .cmake files though, as those files must always contain a lot of static boilerplate code that is the same regardless of the type of generated export (check the CMake version, push a new policy, prevent multiple inclusion, load the various `-<config>` files, check that targets actually exist, clean up temporary variables).

eli-schwartz:
I'm fairly confused by #7632 because different components should probably not really care how each are built. e.g. you may want to build your main project using a debug buildtype, but you probably DON'T want to link to debug versions of every third-party component you link to. So cmake's design seems questionable.

Regardless, meson doesn't build the entire project twice, once with release flags and once with debug flags (???) so there doesn't seem to be any actual target-BUILDTYPE.cmake files to create, you either get (are happy with) the one you originally configured meson with, or you error out. And what do you do with "custom"? Worse, what do you do with "plain"?

Generating `find_package` compatible configs seems like a worthy goal in its own respect. I'm not actually sure "adding such and such frill that cmake does" is a worthy goal...

Tachi107:
I don't know how generating different exports for different build configurations is helpful in practice, and I've never encountered someone who actually ships multiple `Targets-<config>.cmake` files. As an example, Debian only ships `Targets-none.cmake` files, as when they build packages they set the `CMAKE_BUILD_TYPE` to None. It doesn't really matter what config you use to generate or consume the export, as long as there is only one you'll use that regardless of your current build type.

Don't ask me why things work this way, I have no idea... But people rely on having different files even if they don't care about the actual build type.

Meson could even generate `Targets-meson.cmake` files, it really doesn't matter.

eli-schwartz:
> To be clear, CMake doesn't either. If you build your project with `CMAKE_BUILD_TYPE` set to "Unicorn" you'll end up only with a `Targets-unicorn.cmake` file, and your users will happily consume your export without even noticing.

This is not convincing me that we should include `-unicorn` (or `-release`) in the filename...

:D

I still don't understand whose workflow is harmed by naming the file `Targets.cmake` instead of `Targets-producedbymeson.cmake`.

Tachi107:
You must have both. It's just the way CMake exports work.

Quoting the issue I linked before:

> - are not separated by configuration making integration into vcpkg impossible ...... 
> - this means general setup must be separated and put into `-config/-target.cmake` while the configuration aware targets belong into `-target-<config>.cmake`. CMake will select the configuration depending on `CMAKE_BUILD_TYPE` or a set mapping

eli-schwartz:
I fail to see why "you must have both" is true for `find_package` to work.

Maybe you need both for "vcpkg integration" to work. But that seems to be doomed anyway because you'll only have one buildtype.

Tachi107:
> I fail to see why "you must have both" is true for `find_package` to work.

`find_package` doesn't require both. It doesn't even require a `Targets.cmake` file. The goal of this PR is to replicate CMake's `install(EXPORT)` ecosystem. 

> Maybe you need both for "vcpkg integration" to work. But that seems to be doomed anyway because you'll only have one buildtype.

If you build your project once even CMake produces a single buildtype, to have more than one `Targets-config.cmake` file you have to build the project multiple times.

eli-schwartz:
> `find_package` doesn't require both. It doesn't even require a `Targets.cmake` file.

That sounds fascinating. I'd love to hear more! How do we do this, while playing nice with the templates we use for config mode (FindFoo modules are supposedly deprecated / considered terribad practice)?

> The goal of this PR is to replicate CMake's `install(EXPORT)` ecosystem

I see.

Personally I thought the goal of meson, and the cmake module, was to make cmake projects more accessible to meson projects, and vice versa. So it makes sense that we'd... not *want* to generate cmake config files, but be willing to. We have users who would like to use meson, but have projects depending on them using cmake and which want cmake config files so they can actually do so. On windows. Without installing pkg-config.

Now granted, I am not the most knowledgeable about how cmake works under the hood. So I wasn't sure if, as implemented here, certain aspects were needed. They might very well have been needed! I don't know, and I reviewed the PR and asked: this makes me a bit nervous, and I'm not sure why we need it. Can you elaborate?

So originally I thought the goal was to also actually provide the targets so that this worked. Now I realize I was mistaken, and inadvertently led you on. Apparently the goal of this PR is to "reimplement cmake, doing things the way cmake does it, because cmake does it", and we're getting scope creep that isn't backed up by explanations of how it helps achieve meson's goals, a confusing design for files we create which ***aren't*** just copied verbatim from cmake's installation directory, and lots of poking at and from-scratch reimplementation of undocumented cmake internals.

As such, it seems this PR is in direct contradiction to my vision, and I can't rely on it to avoid adding unnecessary complexity. It's going to add that complexity simply to "Follow The ~~Leader~~ Cmake".

As such I NACK this PR but defer final judgment to @mensinda.

Tachi107:
> > `find_package` doesn't require both. It doesn't even require a `Targets.cmake` file.
> 
> That sounds fascinating. I'd love to hear more! How do we do this, while playing nice with the templates we use for config mode (FindFoo modules are supposedly deprecated / considered terribad practice)?

As far as I understand `find_package()` simply runs a CMake script named `pkgnameConfig.cmake` or `pkgname-config.cmake`. That script is supposed to define some targets and some variables representing the dependency, but there are no hard requirements. We have `Targets.cmake` and `Targets-config.cmake` because exports work that way, but they are not required by `find_package()` itself. What a lot of projects do is to create a `Config.cmake` file that simply includes the `Targets.cmake` file, and this one includes the `Targets-*.cmake` files, so `find_package()` uses them indirectly.

> > The goal of this PR is to replicate CMake's `install(EXPORT)` ecosystem
> Personally I thought the goal of meson, and the cmake module, was to make cmake projects more accessible to meson projects, and vice versa. So it makes sense that we'd... not _want_ to generate cmake config files, but be willing to. We have users who would like to use meson, but have projects depending on them using cmake and which want cmake config files so they can actually do so. 

My goal is not to replicate CMake's `install(EXPORT)` ecosystem for the sake of it, but because just like there are users who want to use Meson and consume non-pkg-config dependencies there would like to use Meson but can't because they currently provide import files and dropping them would be break stuff for their users. And also to improve interoperability between Meson and CMake, because with this generator Meson users would be able to also offer first class CMake support.

> Now granted, I am not the most knowledgeable about how cmake works under the hood. So I wasn't sure if, as implemented here, certain aspects were needed. They might very well have been needed! I don't know, and I reviewed the PR and asked: this makes me a bit nervous, and I'm not sure why we need it. Can you elaborate?

Currently the CMake module offers the `configure_package_config_file()` function, and it perfectly replicates CMake's functionality, but that function on its own is not that useful because it simply pastes some CMake magic in a text file that is unrelated to the particular project that is using it.

> So originally I thought the goal was to also actually provide the targets so that this worked. Now I realize I was mistaken, and inadvertently led you on.

Actually I don't have a fixed objective. I know I want to make Meson able to generate files consumable by `find_package()`, and that requires doing something similar to what is already done in the pkgconfig module.
As that functionality is where 80% of the difficulty lies, I thought I could replicate more closely CMake's EXPORT/import ecosystem by simply splitting the information needed to find the dependency in two different files (`Targets.cmake and `Targets-config.cmake`).

Doing things this way is in my opinion worth it, because it would:

1. Allow CMake users to switch to Meson while not breaking anything
2. Work with tools that make use of those files (e.g. vcpkg)

Let me provide an extra example. The first one is an hypothetical `dependencyConfig.cmake` file generated by Meson with the "simple" approach, while the other is composed by `dependencyConfig.cmake`, `dependencyTargets.cmake` and `dependencyTargets-config.cmake`, and follows the "CMake-like" approach.

Simple `dependencyConfig.cmake`

```cmake
add_library(project::name SHARED IMPORTED)

target_compile_options(project::name INTERFACE "-DSHARED_LIB")
target_include_directories(project::name INTERFACE "${_IMPORT_PREFIX}/include/subdir")
target_link_libraries(project::name INTERFACE "pthread")

set_target_properties(project::name PROPERTIES IMPORTED_LOCATION "${_IMPORT_PREFIX}/lib/x86_64-linux-gnu/libname.so)"
```
CMake-like `dependencyConfig.cmake`

```cmake
include(${CMAKE_CURRENT_LIST_DIR}/dependencyTargets.cmake)
```

CMake-like `dependencyTargets.cmake`

```cmake
add_library(project::name SHARED IMPORTED)

target_compile_options(project::name INTERFACE "-DSHARED_LIB")
target_include_directories(project::name INTERFACE "${_IMPORT_PREFIX}/include/subdir")
target_link_libraries(project::name INTERFACE "pthread")

file(GLOB CONFIG_FILES "${CMAKE_CURRENT_LIST_DIR}/dependencyTargets-*.cmake")
foreach(f ${CONFIG_FILES})
  include(${f})
endforeach()
```

CMake-like `dependencyTargets-release.cmake`

```cmake
set_property(TARGET project::name APPEND PROPERTY IMPORTED_CONFIGURATIONS RELEASE)
set_target_properties(project::name PROPERTIES
  IMPORTED_LOCATION_RELEASE "${_IMPORT_PREFIX}/lib/x86_64-linux-gnu/libname.so"
)
```

If we leave out all of the CMake boilerplate that I copy-pasted in my _draft_ for testing purposes and a lot of boilerplate that the two approaches would both need, we can see that there aren't a lot of differences.

mesonbuild/modules/cmake.py
Febbe: CMake itself supports unix seperators. Also, windows does since a specific version.

Tachi107: Yes, but python/meson converts them to backslashes on Windows, while CMake uses Posix separators everywhere. How can I tell python/meson to unconditionally return Posix separator?

Febbe: From [stackoverflow](https://stackoverflow.com/questions/54671385/convert-windowspath-to-posixpath) use: `pathlib.path(r'foo\bar').as_posix()`

Febbe:
# TLDR:

 - find_package only requires one file: `<targetname>Config.cmake` or `<targetname>-config.cmake`
 - find_package accepts an optional `<targetname>ConfigVersion.cmake` or `<targetname>-config-version.cmake` file
 - other files are meant to be called by the config file, they are not required to be called:
     - calling the `<XXX>Targets.cmake` file is just the way how cmake does this, meson is not required to, when writing cmake-config files.
     - cmake makes no decision, whether to import Debug or Release libs, it does what the project owner defined or what he generated via the cmake export helpers
        - this is usually guarded by am option or cache variable whether to load a library with Debug or Release. For example, set(<LIBNAME>_BUILD_TYPE Release CACHE STRING "Choose the link type of the library")
        - They (cmake) use different files, because it's then not a problem, when for example the debug build is not there, the file can't be loaded then.
        - The general config or targets file is then the same for all cmake-configurations/cmake-build-types (meson-buildtypes)
        - This file checks for the availability of the configuration and loads them if desired. 

# Answers, and feedback:

I think, the reason why cmake chose the approach to split config and target files is to give the library designer a maximum of flexibility, but without increasing the effort. The targets file can be created by cmake, but is not required to. It just dumps "Public and interface attributes" which are required to link against the lib. A programmer can't do it better. But for special cases, wishes and exceptions, he/she can modify the config file.

The reason, why cmake may split the Target files into `XXXTarget-<buildtype>.cmake` is, that they don't require to parse possible existing files and add a new build-type to the target config. All possible build-type-targets are already defined and are just added as extra optional file.

Meson could choose any other way. 

> 
```
cmake.generate_export(
  dep target...
  name: str
  namespace: str
  subdirs: list[str]
)
```

I would mimic `pkg.generate`

--

@eli-schwartz since you already mimic [configure_package_config_file()](https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html#command:configure_package_config_file) you should also generate those Target Files like cmake does this. If you don't like that, you should remove it and define your own way to generate a default config file.

Those generated Configures are useless without the targets file, the other way around, not.

# I tried to search for definitions in the cmake doc, for the most of the discussions:

> There may be additional files providing CMake commands or [Imported Targets](https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#imported-targets) for you to use. CMake does not enforce any naming convention for these files. They are related to the primary <PackageName>Config.cmake file by use of the CMake [include()](https://cmake.org/cmake/help/latest/command/include.html#command:include) command. The <PackageName>Config.cmake file would typically include these for you, so they won't usually require any additional step other than the call to [find_package()](https://cmake.org/cmake/help/latest/command/find_package.html#command:find_package). 

Also, Version files are optional: 
 > A separate optional file named <PackageName>ConfigVersion.cmake or <LowercasePackageName>-config-version.cmake may also exist in the same directory.

1: [naming-convention, optional version-file](https://cmake.org/cmake/help/latest/guide/using-dependencies/index.html#config-file-packages)

mesonbuild/modules/cmake.py
Tachi107: Oh ok, simpler than expected :D

Tachi107:
@Febbe thanks for your deep feedback!

> The reason, why cmake may split the Target files into `XXXTarget-<buildtype>.cmake` is, that they don't require to parse possible existing files and add a new build-type to the target config. All possible build-type-targets are already defined and are just added as extra optional file.

I don't see how splitting the files per build type is an issue. It is not difficult, and I would just do that to mimic what CMake does, so that users could switch from CMake to Meson without ever worrying about breaking stuff for their users. Whether splitting stuff makes sense from a technical standpoint is not relevant, as unfortunately we can't change the way CMake works. After all, I use Meson for a reason ;)

> Also, Version files are optional:

Yep, but Meson already generates them so this is not an issue.

mesonbuild/modules/cmake.py
eli-schwartz: The pkg-config file generator also uses as_posix() for the same reason.

eli-schwartz:
> I don't see how splitting the files per build type is an issue. It is not difficult

Difficult or not, it seems to me that it is wrong...

>  and I would just do that to mimic what CMake does, so that users could switch from CMake to Meson without ever worrying about breaking stuff for their users.

I don't see why it would break anything, to stop referencing files that don't exist in the first place.

Febbe:
@eli-schwartz :
> I don't see why it would break anything, to stop referencing files that don't exist in the first place.

I think, @Tachi107 means, that it should be possible to just copy the old `<XXX>Config.cmake.in` template file from the old cmake build system into the new meson system and having it worked out of the box. This would actually be nice, but may be too much of the good. Most maintainers just want to provide a valid configuration in the simplest possible way. Like just calling `cmake.export(my_meson_lib, MyNs::my_meson_lib);` And it generates automatically a fully working XXXConfig.cmake script.

>> I don't see how splitting the files per build type is an issue. It is not difficult

> Difficult or not, it seems to me that it is wrong...

It's definitely not wrong, here in Germany we say `Names are sound and smoke`, the naming actually doesn't matter here, so basically a **valid** cmake config could import `import(apples_and_pears)`. Also, it can import an arbitrary amount of other cmake scripts. This is actually common practice, since some libraries depend on other cmake projects. So this cmake config will issue a call to find_package. The reason for this is, that a cmake config file is itself just a script which is executed after an import in find_package.

This means, that you (meson) can do whatever you want, and it is also possible to just write everything into the `XXXConfig.cmake` file; no target files required.
**But** everyone who wrote custom code in the old CMake(old tooling) Config file, has to write again a new file
If it's desired to also ship other build types than release, for `example` `plain`. 
1. You also have to either write a general config file which tries to detect if a build-type has been build (search for debug headers, search for debug libs etc.)
2. Or you have to reread the file, and you must detect if there is already a build done, then override it, else add it.

Both versions have huge drawbacks:
1. : All flags for all desired buildtypes must be known in each build-type run. Also you would have to write extremly much cmake boilerplate code for the detection.
2. : You would have to parse and interpret the content, this requires either cmake as dep, or you must an own cmake interpreter.

Therefore, I recommend to just use one extra file for an additionally build-type-target, this will be both more robust and reduces code size.

Alternatively, you could just support one Target to be exported, like only Release or the current build type. This will enable you to just use one cmake-config file.

Long story short: if you want to support multiple export sets, it will be the best to split it.
If you don't, just export everything into the cmake-config file or one single targets file.

Tachi107:
Yes, from a technical point of view I agree, it doesn't make that much sense (pkg-config files are way simpler, and work just as well as CMake exports, if not better).

But the goal of the CMake module is to replicate CMake functionality so that users switching to Meson can keep providing CMake support to their users. The write_basic_package_version_file() function has questionable reasoning behind it (i.e. it implements the actual logic that checks for version compatibility, and also performs architecture checks for some reason, while imo it should've simply declared the version of the library), but Meson supports it anyway.

> I don't see why it would break anything, to stop referencing files that don't exist in the first place.

I think I haven't been clear. Of course it wouldn't break anything for existing Meson users, but doing things differently would create issues for users switching from CMake to Meson.

Febbe:
> (pkg-config files are way simpler, and work just as well as CMake exports, if not better).

Not better for me on Windows with msvc as compiler in the tool chain at least :/. If pkg-config worked, I would not even bother if there is a cmake config.

PR #9984, opened 2022-02-10
wraptool: be forgiving of local wrap files

eli-schwartz:
Do not traceback when trying to update a wrap that isn't a [wrap-file], just report the problem.

Do not traceback on perfectly valid WrapDB wraps that don't have a patch_url because they have upstream meson.build, instead try to parse the version from the source tarball filename.


xclaesse:
LGTM. On the longer term we should see what's still useful in wraptool, I think most features are in `meson subprojects` now.

eli-schwartz:
I think there is probably a useful distinction between `meson wrap ...` (interacts with the wrapdb to acquire .wrap files from an official index) and `meson subprojects ...` (interacts with locally present wraps but doesn't know which authority provided them).

PR #9985, opened 2022-02-10
Add unstable_wayland module

markbolhuis:
This module can look for wayland xml files from the wayland-protocols package, and can generate .c and .h files using wayland-scanner.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -3,6 +3,7 @@
 /mesonbuild/modules/cmake.py @mensinda
 /mesonbuild/modules/unstable_external_project.py @xclaesse
 /mesonbuild/modules/unstable_rust.py @dcbaker
+/mesonbuild/modules/unstable_wayland.py @markbolhuis
 /mesonbuild/ast/ @mensinda
 /mesonbuild/cmake/ @mensinda
 /mesonbuild/compilers/ @dcbaker
--- a/docs/markdown/Wayland-module.md
+++ b/docs/markdown/Wayland-module.md
@@ -1,0 +1,64 @@
+# Unstable Wayland Module
+
+This module is available since version 0.62.0.
+
+This module provides helper functions to find wayland protocol
+xmls and to generate .c and .h files using wayland-scanner
+
+**Note**: this module is unstable. It is only provided as a technology
+preview. Its API may change in arbitrary ways between releases or it
+might be removed from Meson altogether.
+
+## Quick Usage
+
+```meson
+project('hello-wayland', 'c')
+
+wl_dep = dependency('wayland-client')
+wl_mod = import('unstable-wayland')
+
+xml = wl_mod.find_protocol('xdg-shell')
+xdg_shell = wl_mod.scan_xml(xml)
+
+executable('hw', 'main.c', xdg_shell, dependencies : wl_dep)
+```
+
+## Methods
+
+### find_protocol
+
+```meson
+xml = wl_mod.find_protocol(
+  'xdg-decoration',
+  state : 'unstable',
+  version : 1,
+)
+```
+This function requires one positional argument: the protocol base name.
+- `state` Optional arg that specifies the current state of the protocol.
+Either stable, staging, or unstable.
+The default is stable.
+- `version` The backwards incompatible version number.
+Required for staging or unstable. An error is raised for stable.
+
+### scan_xml
+```meson
+generated = wl_mod.scan_xml(
+  'my-protocol.xml',
+  side : 'client',
+  scope : 'private',
+)
+```
+This function accepts one or more arguments of either string or file type.
+
+- `side` Optional arg that specifies if client or server side code is generated.
+The default is client side.
+- `scope` Optional arg that specifies the scope of the generated code.
+Either public or private.
+The default is private.
+
+
+## Links
+- [Official Wayland Documentation](https://wayland.freedesktop.org/docs/html/)
+- [Wayland GitLab](https://gitlab.freedesktop.org/wayland)
+- [Wayland Book](https://wayland-book.com/)
--- a/mesonbuild/modules/unstable_wayland.py
+++ b/mesonbuild/modules/unstable_wayland.py
@@ -1,0 +1,120 @@
+# Copyright 2022 Mark Bolhuis <mark@bolhuis.dev>
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+
+from . import ExtensionModule, ModuleReturnValue
+from ..build import CustomTarget
+from ..interpreter.type_checking import NoneType, in_set_validator
+from ..interpreterbase import FeatureNew, typed_pos_args, typed_kwargs, KwargInfo
+from ..mesonlib import File, MesonException, MachineChoice
+
+
+class WaylandModule(ExtensionModule):
+
+    @FeatureNew('wayland module', '0.62.0')
+    def __init__(self, interpreter):
+        super().__init__(interpreter)
+
+        self.protocols_dep = None
+        self.pkgdatadir = None
+        self.scanner_bin = None
+
+        self.methods.update({
+            'scan_xml': self.scan_xml,
+            'find_protocol': self.find_protocol,
+        })
+
+    @typed_pos_args('wayland.scan_xml', varargs=(str, File), min_varargs=1)
+    @typed_kwargs(
+        'wayland.scan_xml',
+        KwargInfo('side', str, default='client', validator=in_set_validator({'client', 'server'})),
+        KwargInfo('scope', str, default='private', validator=in_set_validator({'private', 'public'})),
+    )
+    def scan_xml(self, state, args, kwargs):
+        if self.scanner_bin is None:
+            self.scanner_bin = state.find_program('wayland-scanner', for_machine=MachineChoice.BUILD)
+
+        scope = kwargs['scope']
+        side = kwargs['side']
+
+        xml_files = self.interpreter.source_strings_to_files(args[0])
+        targets = []
+        for xml_file in xml_files:
+            name = os.path.splitext(os.path.basename(xml_file.fname))[0]
+
+            code = CustomTarget(
+                f'{name}-protocol',
+                state.subdir,
+                state.subproject,
+                [self.scanner_bin, f'{scope}-code', '@INPUT@', '@OUTPUT@'],
+                [xml_file],
+                [f'{name}-protocol.c'],
+                backend=state.backend,
+            )
+            targets.append(code)
+
+            header = CustomTarget(
+                f'{name}-{side}-protocol',
+                state.subdir,
+                state.subproject,
+                [self.scanner_bin, f'{side}-header', '@INPUT@', '@OUTPUT@'],
+                [xml_file],
+                [f'{name}-{side}-protocol.h'],
+                backend=state.backend,
+            )
+            targets.append(header)
+
+        return ModuleReturnValue(targets, targets)
+
+    @typed_pos_args('wayland.find_protocol', str)
+    @typed_kwargs(
+        'wayland.find_protocol',
+        KwargInfo('state', str, default='stable', validator=in_set_validator({'stable', 'staging', 'unstable'})),
+        KwargInfo('version', (int, NoneType)),
+    )
+    def find_protocol(self, state, args, kwargs):
+        base_name = args[0]
+        xml_state = kwargs['state']
+        version = kwargs['version']
+
+        if xml_state != 'stable' and version is None:
+            raise MesonException(f'{xml_state} protocols require a version number.')
+
+        if xml_state == 'stable' and version is not None:
+            raise MesonException('stable protocols do not require a version number.')
+
+        if self.protocols_dep is None:
+            self.protocols_dep = self.interpreter.func_dependency(state.current_node, ['wayland-protocols'], {})
+
+        if self.pkgdatadir is None:
+            self.pkgdatadir = self.protocols_dep.get_variable(pkgconfig='pkgdatadir', internal='pkgdatadir')
+
+        if xml_state == 'stable':
+            xml_name = f'{base_name}.xml'
+        elif xml_state == 'staging':
+            xml_name = f'{base_name}-v{version}.xml'
+        else:
+            xml_name = f'{base_name}-unstable-v{version}.xml'
+
+        path = os.path.join(self.pkgdatadir, xml_state, base_name, xml_name)
+
+        if not os.path.exists(path):
+            raise MesonException(f'The file {path} does not exist.')
+
+        return File.from_absolute_file(path)
+
+
+def initialize(interpreter):
+    return WaylandModule(interpreter)
--- a/ci/ciimage/arch/install.sh
+++ b/ci/ciimage/arch/install.sh
@@ -14,7 +14,7 @@ pkgs=(
   itstool gtk3 java-environment=8 gtk-doc llvm clang sdl2 graphviz
   doxygen vulkan-validation-layers openssh mercurial gtk-sharp-2 qt5-tools
   libwmf valgrind cmake netcdf-fortran openmpi nasm gnustep-base gettext
-  python-lxml hotdoc rust-bindgen qt6-base qt6-tools
+  python-lxml hotdoc rust-bindgen qt6-base qt6-tools wayland wayland-protocols
   # cuda
 )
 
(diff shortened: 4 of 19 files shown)


xclaesse:
I haven't looked at this PR yet, does it allow having wayland-protocols as subproject and generate C files?
- https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/137
- https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/4482

tristan957:
> I haven't looked at this PR yet, does it allow having wayland-protocols as subproject and generate C files?
> 
> * https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/137
> * https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/4482

```
        self.protocols_pc = PkgConfigDependency('wayland-protocols', interpreter.environment, {'required': False})
        if self.protocols_pc.found():
            self.pkgdatadir = self.protocols_pc.get_pkgconfig_variable('pkgdatadir', [], None)
        else:
            self.pkgdatadir = None
```

Does not seem like it. Seems like it could though. Just needs to not use the pkg-config specific stuff. Seems like the PR needs to get the dependency through `Dependency('xxx', type: 'auto')` and then call the `get_variable()` function on the dependency and then it would probably work (?)

xclaesse:
@markbolhuis are you still planning on working on this PR? I would like to get this in, if we can get it into shape.

markbolhuis:
> @markbolhuis are you still planning on working on this PR? I would like to get this in, if we can get it into shape.

Just pushed an update that should detect sub-projects.

Also do we want the module to detect the wayland-protocols sub-project automatically or do we want to just allow the user to pass in a path string to `find_protocol` to override `pkgdatadir`?
Like so:
```
proj = subproject('wayland-protocols')
dir = proj.get_variable(pkgconfig: 'pkgdatadir')
wl_mod.find_protocol('xdg-shell', search_dir : dir)
```

mesonbuild/modules/unstable_wayland.py
```diff
@@ -0,0 +1,133 @@
...
+                state.subdir,
+                state.subproject,
+                [self.scanner_bin, f'{side}-header', '@INPUT@', '@OUTPUT@'],
+                [xml_path],
+                [f'{name}-{side}-protocol.h'],
+                backend=state.backend,
+            )
+            targets.append(header)
+
+        return ModuleReturnValue(targets, [targets])
```
xclaesse, line 94: nitpicking: I don't think you need to build a list of list for the 2nd argument

mesonbuild/modules/unstable_wayland.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def find_protocol(self, state, args, kwargs):
```
xclaesse, line 96: Should use @typed_pos_args and @typed_kwargs here. They replace the manual type checks you do below.

mesonbuild/modules/unstable_wayland.py
```diff
@@ -0,0 +1,132 @@
...
+            xml_name = f'{base_name}-unstable-v{version}.xml'
+        else:
+            raise MesonException('stage must be either stable, staging, or unstable.')
+
+        path = os.path.join(self.pkgdatadir, xml_state, base_name, xml_name)
+
+        if not os.path.exists(path):
+            raise MesonException(f'The file {path} does not exist.')
+
+        return ModuleReturnValue(path, [path])
```
xclaesse, line 128: I think you should return `File.from_absolute_file(path)`, better return a file object than a string. Also ModuleReturnValue is only needed when returning build targets.

xclaesse (changes requested):
I like the logic, mostly small comments to make it more modern meson code.

I would like to see a unit test for this, we'll probably have to install wayland-protocols on some CI runners thought.

xclaesse:
> Also do we want the module to detect the wayland-protocols sub-project automatically

Yes, but I think it's going to work already, when you call dependency() or find_program() meson will fallback to the subproject as long as the main project has the proper subprojects/wayland-protocols.wrap file. But I think we should add `program_names=wayland-scanner` in https://github.com/mesonbuild/wrapdb/blob/master/subprojects/wayland.wrap for this to work.

docs/markdown/Wayland-module.md
```diff
@@ -0,0 +1,57 @@
...
+  state : 'unstable',
+  version : 1,
+)
+```
+This function requires one positional argument: the protocol base name.
+- `state` Optional arg that specifies the current state of the protocol.
+Either stable, staging, or unstable.
+The default is stable.
+- `version` The backwards incompatible version number.
+Required for staging or unstable. Ignored for stable.
```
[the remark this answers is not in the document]
markbolhuis: A warning is probably enough IMO.

markbolhuis:
> I would like to see a unit test for this, we'll probably have to install wayland-protocols on some CI runners thought.

I'm ready to push a new commit that addresses most your points, and I've added some tests, but before I do I'm not sure how to detect the absence of wayland-protocols to skip them. 
What I have right now, but this isn't enough:
```python
def should_skip_wayland():
    if mesonlib.is_windows() or mesonlib.is_osx():
        return True
    if not shutil.which('wayland-scanner'):
        return True
    return False
```

tristan957:
Could you check for the existence of the .pc file?

xclaesse:
I don't think you need to deal with absence of wayland-protocols at all. It's up to the user to know if they need wayland or not for the given platform. As soon as the user does `wl_mod.scan_xml()` and the dependency is missing that should error out. I would expect applications to do for example:

```meson
wl_client_dep = dependency('wayland-client', required: false)
if wl_client_dep.found()
  wl_mod.scan_xml(...)
endif
```

xclaesse:
Ahh, sorry, you meant for the unittest. We probably need to add the package in the docker image first. Then in your unit test meson.build:

```meson
wl_client_dep = dependency('wayland-protocols', required: false)
if not wl_client_dep.found()
  error('MESON_SKIP_TEST: wayland-protocols is not installed')
endif```

**Comment by markbolhuis** [comment]
> Would it make more sense to skip if not Linux instead of checking windows or osx? Can't remember if any BSDs use Wayland.

Wayland can be compiled for FreeBSD at least:
https://gitlab.freedesktop.org/wayland/wayland/-/blob/main/meson.build#L72

File: docs/markdown/Wayland-module.md (hunk 4 above)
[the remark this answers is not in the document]
Comment (xclaesse) [reply docs/markdown/Wayland-module.md:42]: fair enough.

File: mesonbuild/modules/unstable_wayland.py [hunk 5]
```diff
@@ -0,0 +1,128 @@
...
+        self.methods.update({
+            'scan_xml': self.scan_xml,
+            'find_protocol': self.find_protocol,
+        })
+
+    @typed_pos_args('wayland.scan_xml', varargs=(str, File), min_varargs=1)
+    @typed_kwargs(
+        'wayland.scan_xml',
+        KwargInfo('side', str, default='client'),
+        KwargInfo('scope', str, default='private'),
```
xclaesse, line 44: Can do even better: `KwargInfo('side', str, default='client', validator=in_set_validator({'client', 'server'})),` likewise for scope. With that you can remove the check below.

mesonbuild/modules/unstable_wayland.py
```diff
@@ -0,0 +1,128 @@
...
+                backend=state.backend,
+            )
+            targets.append(header)
+
+        return ModuleReturnValue(targets, targets)
+
+    @typed_pos_args('wayland.find_protocol', str)
+    @typed_kwargs(
+        'wayland.find_protocol',
+        KwargInfo('state', str, default='stable'),
```
xclaesse, line 91: You can use validator here too to restrict possible values.

xclaesse (commented):
Small nitpicking and I think it's ready to go. Waiting to see how CI goes however.

mesonbuild/modules/unstable_wayland.py
```diff
@@ -0,0 +1,128 @@
...
+
+    @typed_pos_args('wayland.scan_xml', varargs=(str, File), min_varargs=1)
+    @typed_kwargs(
+        'wayland.scan_xml',
+        KwargInfo('side', str, default='client'),
+        KwargInfo('scope', str, default='private'),
+    )
+    def scan_xml(self, state, args, kwargs):
+        if self.scanner_bin is None:
+            self.scanner_bin = state.find_program('wayland-scanner')
```
xclaesse, line 48: Oh, here you actually should search for the build machine, add `for_machine=MachineChoice.BUILD`. You'll have to add that argument in `modules/__init__.py` file where it calls `find_program_impl()` that has that arg already. We probably also could be smarter and lookup in the pkg-config file, IIRC they define the path to wayland-scanner there too. But maybe we can leave that to later, as I mentioned earlier we should fix wayland-protocols project to use `meson.override_find_program()` there too.

xclaesse:
run_project_tests.py:1036:1: error: Function is missing a return type annotation
run_project_tests.py:1099:44: error: Call to untyped function "should_skip_wayland" in typed context
Found 2 errors in 1 file (checked 16[8](https://github.com/mesonbuild/meson/runs/5344466215?check_suite_focus=true#step:5:8) source files)

xclaesse:
Now the question is does it actually run your test on any of our ci images? You probably need to add needed package in a few Linux distro images: https://github.com/mesonbuild/meson/tree/master/ci/ciimage

markbolhuis:
> Now the question is does it actually run your test on any of our ci images? You probably need to add needed package in a few Linux distro images: https://github.com/mesonbuild/meson/tree/master/ci/ciimage

Seems it:
https://github.com/mesonbuild/meson/runs/5344466344?check_suite_focus=true#step:4:1788

xclaesse:
Looks all good to me, just waiting for ci to complete

xclaesse:
Hmmm fails to build opensuse docker image, but seems unrelated. Also it seems arch and Fedora ci are not using the updates images. I think that needs to be a separate pr that we merge first.

eli-schwartz:
Updates to the CI files trigger the CI builder set of checks, which both build an image and test it -- and once merged to master, also upload it.

Until that image is uploaded, the regular workflows, including the ones in this PR, will use the old image... Which is perfectly fine, we don't care. If the CI builder check passes, we can ignore the status of the regular workflow corresponding to the same check. Maybe master isn't green, but it is not-green in the same way any known transient failure is (for example, canceled checks, macOS clock skew, mpi/cuda flakiness) and the green will resume with the next push to master.

eli-schwartz:
> ../test cases/vala/14 target glib version and gresources/gres/test-resources.xml: Failed to close file descriptor for child process (Operation not permitted).

Let's try kicking it.

CODEOWNERS
```diff
@@ -3,6 +3,7 @@
 /mesonbuild/modules/cmake.py @mensinda
 /mesonbuild/modules/unstable_external_project.py @xclaesse
 /mesonbuild/modules/unstable_rust.py @dcbaker
+/mesonbuild/modules/unstable_wayland.py @markbolhuis
```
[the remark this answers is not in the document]
eli-schwartz: What is Github supposed to do, error out with "error: not allowed to push updates to a file named CODEOWNERS because it has content that github doesn't like"?

In fact, github does at least do this:

> This CODEOWNERS file contains errors

This error shows up in git master too.

The file is useless, no one looks at it other than github and github makes it useless by not allowing component ownership by people who lack global push access.

docs/markdown/Wayland-module.md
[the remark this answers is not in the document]
eli-schwartz: If we can get this API right from the beginning, why warn but allow bad usage instead?

mesonbuild/modules/unstable_wayland.py
```diff
@@ -0,0 +1,121 @@
...
+            self.scanner_bin = state.find_program('wayland-scanner', for_machine=MachineChoice.BUILD)
+
+        scope = kwargs['scope']
+        side = kwargs['side']
+
+        xml_files = self.interpreter.source_strings_to_files(args[0])
+        targets = []
+        for xml_file in xml_files:
+            base_name = os.path.basename(xml_file.fname)
+            name, _ = os.path.splitext(base_name)
```
eli-schwartz, line 57: This could probably just be `os.path.splitext(os.path.basename(xml_file.fname))[0]`

mesonbuild/modules/unstable_wayland.py
```diff
@@ -0,0 +1,121 @@
...
+        'wayland.find_protocol',
+        KwargInfo('state', str, default='stable', validator=in_set_validator({'stable', 'staging', 'unstable'})),
+        KwargInfo('version', (int, NoneType)),
+    )
+    def find_protocol(self, state, args, kwargs):
+        base_name = args[0]
+        xml_state = kwargs['state']
+        version = kwargs['version']
+
+        if (xml_state == 'staging' or xml_state == 'unstable') and version is None:
```
eli-schwartz, line 94:
```suggestion
        if xml_state in {'staging', 'unstable'} and version is None:
```

Or

```suggestion
        if xml_state != 'stable' and version is None:
```

Either one currently gives technically correct results, the semantics depend on whether you'd expect potential future types to require a version or not. I'd actually expect the latter to be correct, but your current check implies the former.

mesonbuild/modules/unstable_wayland.py
```diff
@@ -0,0 +1,121 @@
...
+
+        if self.protocols_dep is None:
+            self.protocols_dep = self.interpreter.func_dependency(state.current_node, ['wayland-protocols'], {})
+
+        if self.pkgdatadir is None:
+            self.pkgdatadir = self.protocols_dep.get_variable(pkgconfig='pkgdatadir', internal='pkgdatadir')
+
+        if xml_state == 'stable':
+            if version is not None:
+                mlog.warning('stable protocols do not require a version number.', location=self.interpreter.current_node, once=True)
```
eli-schwartz, line 105: I would prefer a hard error.

eli-schwartz: Also this validation check could be moved up above to where we check and error out on a missing one for unstable/staging, as it's better to error out or warn on invalid arguments earlier rather than later, and currently it happens after checking the environment state (dependency lookup).

eli-schwartz (changes requested):

xclaesse:
About CI failure, I think you need this change:
```diff
--- a/test cases/vala/14 target glib version and gresources/test.vala   
+++ b/test cases/vala/14 target glib version and gresources/test.vala   
@@ -9,7 +9,7 @@ public class TestWidget : Box {
   }
 
   [GtkChild]
-  private Entry entry;
+  private unowned Entry entry;
 
   public TestWidget (string text) {
     this.text = text;
```

markbolhuis:
> About CI failure, I think you need this change:

Should be a separate PR, right?

xclaesse:
I think you can just add a commit here

eli-schwartz:
That's not going to fix anything, it is a non-fatal warning several lines down from the fatal error I mentioned above.

xclaesse:
@markbolhuis FYI, created a wayland MR: https://gitlab.freedesktop.org/xclaesse/wayland/-/merge_requests/1

markbolhuis:
Sorry for being a bit slow here, but I'm bit confused on what exactly I have to do. 
Probably be far easier if one of you could just submit a change ahead of merging this, and I'll just revert my last commit.


> @markbolhuis FYI, created a wayland MR: https://gitlab.freedesktop.org/xclaesse/wayland/-/merge_requests/1

Getting 404

xclaesse:
@markbolhuis sorry, not sure what happened, gitlab created a private fork for wayland repository... Here it is now: https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/212

xclaesse:
> That's not going to fix anything, it is a non-fatal warning several lines down from the fatal error I mentioned above.

Ah right, misread the logs. Unfortunately the actual error reproduces every time.

eli-schwartz:
And yet for some reason only reproduces "when using an updated set of packages"?

xclaesse:
> And yet for some reason only reproduces "when using an updated set of packages"?

yes... let's see how it goes in a separate PR: https://github.com/mesonbuild/meson/pull/10037

eli-schwartz:
Well, I mean if you go look at https://github.com/mesonbuild/meson/actions/workflows/images.yml?query=event%3Aschedule

For the past 3 weeks that image has repeatedly failed with this precise error.

xclaesse:
ah... ok thanks, let's merge then.

eli-schwartz:
Actually I wanted to remove the CODEOWNERS change since it is totally useless and that's not how github works. :/

xclaesse:
Does not hurt to have a person of reference who knows about wayland, even if they don't have write access. Does it cause an issue with github?

eli-schwartz:
Github shows the file in red and says "This CODEOWNERS file contains errors". Then it totally ignores the person without write access.

Not the hugest issue in the world. It doesn't exactly break the feature so we get *no* reviewers.

...

So the only way to actually request that person for review is to have the PR author manually look at this "automatic reviewers" file, and manually ping interested parties.

Github dumb. It is what it is.

And no one is actually doing this manual stuff at all.

tristan957:
I wouldn't mind being marked as code owner of java module, but doesn't really matter.

xclaesse:
> Then it totally ignores the person without write access.

oh, was expecting github to still send notification... bah, not a big deal, conceptually it is right to have him in CODEOWNER and maybe some day github will fix that.

eli-schwartz:
GitHub doesn't think it's a bug. :( One of the side effects of this is the UX that CODEOWNERS actually displays. It only notifies people as a side effect of marking them for review. The purpose is tied to the ability to configure a repository to only allow merging after a certain number of required reviews.

Only people with write access can submit a require review.

fooishbar:
I'm also very happy to be CCed on Wayland changes.

PR #9988, opened 2022-02-11
Qt: fix broken tool names

eli-schwartz:
There is broken behavior w.r.t. looking up tool names:
- we tried to find `foo-Qt5` instead of `foo-qt5`, resulting in two wrongly skipped jobs
- we don't look for `foo6` on Qt6, but the Qt Company finally standardized a name and it's not the one the community gathered around

Okay, now that we actually look for the right name, we actually find qt6 at all, ever. So now it turns out our support is broken. Yay! This goes together with #9792

### DIFF
--- a/mesonbuild/dependencies/qt.py
+++ b/mesonbuild/dependencies/qt.py
@@ -245,7 +245,7 @@ class QmakeQtDependency(_QtBase, ConfigToolDependency, metaclass=abc.ABCMeta):
 
     def __init__(self, name: str, env: 'Environment', kwargs: T.Dict[str, T.Any]):
         _QtBase.__init__(self, name, kwargs)
-        self.tools = [f'qmake-{self.name}', 'qmake']
+        self.tools = [f'qmake{self.qtver}', f'qmake-{self.name}', 'qmake']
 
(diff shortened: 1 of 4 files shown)


radioactiveman:
@jpakkane, @eli-schwartz: No updates yet? This PR is still waiting for a review.

jpakkane:
LGTM but this is still marked as draft and the test failure is relevant.

eli-schwartz:
The test failure was fixed by the PR referenced above, and after rebasing my original changes on top of that, everything passes.

radioactiveman:
@eli-schwartz, @jlindgren90
I can confirm Meson works now with Qt 6 on Arch Linux. Thanks for taking care of this. :+1: 
#9791 and #9792 can be closed now. Do you plan to backport the fixes for 0.62.1 as well?

eli-schwartz:
> Do you plan to backport the fixes for 0.62.1 as well?

These seem to be stable enough to backport, so I've marked them for backporting. Ultimately it's up to @nirbheek, of course.

nirbheek:
LGTM.

PR #9989, opened 2022-02-11
Fix underscore detection

### LINKED ISSUE #5482: Underscore prefix detection broken if -g specified
This is not an issue currently, due to #5481 hiding this bug but once that is solved, it will become an issue again.

If the cflags contain `-g`, the test for the underscore symbol would incorrectly return as soon as it finds the symbol with or without underscore, but when debug symbols are enabled using `-g`, this will yield the false result. This can easily be illustrated by a small testcase:

```python
symbol_name = b'meson_uscore_prefix'
with open('out.obj', 'rb') as o:
    for line in o:
        # Check if the underscore form of the symbol is somewhere
        # in the output file.
        if b'_' + symbol_name in line:
            print('True')
        # Else, check if the non-underscored form is present
        elif symbol_name in line:
            print('False')
```

Would print:

```
False
True
```

as it first finds the symbol without the underscore, only later finds the symbol with underscore.

Of course this is not an issue when using the meson buildtype option, as meson will not use the `-g` argument from that for the test, but is an issue when `-g` is contained in externally provided flags by env variables or crossfile.

ePirat:
Add two new ways to check if functions have underscore prefix:

- Query the compiler-provided `__USER_LABEL_PREFIX__` define
- Check a hardcoded list of cases where we know the outcome

Fix #5482

### DIFF
--- a/mesonbuild/compilers/mixins/clike.py
+++ b/mesonbuild/compilers/mixins/clike.py
@@ -918,7 +918,7 @@ class CLikeCompiler(Compiler):
         n = '_symbols_have_underscore_prefix_searchbin'
         with self._build_wrapper(code, env, extra_args=args, mode='compile', want_output=True, temp_dir=env.scratch_dir) as p:
             if p.returncode != 0:
-                raise RuntimeError(f'BUG: Unable to compile {n!r} check: {p.stdout}')
+                raise RuntimeError(f'BUG: Unable to compile {n!r} check: {p.stderr}')
             if not os.path.isfile(p.output_name):
                 raise RuntimeError(f'BUG: Can\'t find compiled test code for {n!r} check')
             with open(p.output_name, 'rb') as o:
--- a/mesonbuild/compilers/mixins/clike.py
+++ b/mesonbuild/compilers/mixins/clike.py
@@ -900,9 +900,10 @@ class CLikeCompiler(Compiler):
         return self.compiles(t, env, extra_args=extra_args,
                              dependencies=dependencies)
 
-    def symbols_have_underscore_prefix(self, env: 'Environment') -> bool:
+    def _symbols_have_underscore_prefix_searchbin(self, env: 'Environment') -> bool:
         '''
-        Check if the compiler prefixes an underscore to global C symbols
+        Check if symbols have underscore prefix by compiling a small test binary
+        and then searching the binary for the string,
         '''
         symbol_name = b'meson_uscore_prefix'
         code = '''#ifdef __cplusplus
@@ -914,7 +915,7 @@ class CLikeCompiler(Compiler):
         #endif
         '''
         args = self.get_compiler_check_args(CompileCheckMode.COMPILE)
-        n = 'symbols_have_underscore_prefix'
+        n = '_symbols_have_underscore_prefix_searchbin'
         with self._build_wrapper(code, env, extra_args=args, mode='compile', want_output=True, temp_dir=env.scratch_dir) as p:
             if p.returncode != 0:
                 raise RuntimeError(f'BUG: Unable to compile {n!r} check: {p.stdout}')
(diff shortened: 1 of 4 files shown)


mesonbuild/compilers/mixins/clike.py
```diff
@@ -925,13 +926,83 @@ def symbols_have_underscore_prefix(self, env: 'Environment') -> bool:
...
-                        mlog.debug("Symbols have underscore prefix: YES")
+                        mlog.debug("Underscore prefix check found prefixed function in binary")
                         return True
                     # Else, check if the non-underscored form is present
                     elif symbol_name in line:
-                        mlog.debug("Symbols have underscore prefix: NO")
+                        mlog.debug("Underscore prefix check found non-prefixed function in binary")
                         return False
-        raise RuntimeError(f'BUG: {n!r} check failed unexpectedly')
+        return None
```
eli-schwartz, line 935: This is always the last check, and if I understand correctly, it's a Meson bug if this doesn't have one or the other, right?

So I don't think it should return None.

mesonbuild/compilers/mixins/clike.py
```diff
@@ -925,13 +926,83 @@ def symbols_have_underscore_prefix(self, env: 'Environment') -> bool:
...
+        if result is not None:
+            return result
+
+        # As a last resort, try search in a compiled binary, which is the
+        # most unreliable way of checking this, see #5482
+        result = self._symbols_have_underscore_prefix_searchbin(env)
+        if result is not None:
+            return result
+
+        raise RuntimeError('BUG: symbols_have_underscore_prefix check failed unexpectedly')
```
eli-schwartz, line 1004: This should probably assume one of the checks "must" return something useful, and in this last check, just `return self._symbols_have_underscore_prefix_searchbin(env)`.

If `_symbols_have_underscore_prefix_searchbin` is so broken it returns neither True nor False, it would be its own bug anyway, right?

mesonbuild/compilers/mixins/clike.py
```diff
@@ -925,13 +926,83 @@ def symbols_have_underscore_prefix(self, env: 'Environment') -> bool:
...
                         return True
                     # Else, check if the non-underscored form is present
                     elif symbol_name in line:
-                        mlog.debug("Symbols have underscore prefix: NO")
+                        mlog.debug("Underscore prefix check found non-prefixed function in binary")
                         return False
-        raise RuntimeError(f'BUG: {n!r} check failed unexpectedly')
+        return None
+
+    def _symbols_have_underscore_prefix_define(self, env: 'Environment') -> T.Union[None,bool]:
```
eli-schwartz, line 937:
```suggestion
    def _symbols_have_underscore_prefix_define(self, env: 'Environment') -> T.Optional[bool]:
```

mesonbuild/compilers/mixins/clike.py
```diff
@@ -925,13 +926,83 @@ def symbols_have_underscore_prefix(self, env: 'Environment') -> bool:
...
+        '''
+        Check if symbols have underscore prefix by querying the
+        __USER_LABEL_PREFIX__ define that most compilers provide
+        for this. Return if funtions have underscore prefix or None
```
eli-schwartz, line 941: typo
```suggestion
        for this. Return if functions have underscore prefix or None
```

mesonbuild/compilers/mixins/clike.py
```diff
@@ -925,13 +926,83 @@ def symbols_have_underscore_prefix(self, env: 'Environment') -> bool:
...
+        Check if symbols have underscore prefix by consulting a hardcoded
+        list of cases where we know the results.
+        Return if functions have underscore prefix or None if unknown.
+        '''
+        m = env.machines[self.for_machine]
+        # Darwin always uses the underscore prefix, not matter what
+        if m.is_darwin():
+            return True
+        # Windows uses the underscore prefix on x86 (32bit) only
+        if m.is_windows() or m.is_cygwin():
```
eli-schwartz, line 978: I genuinely have no clue here, but...

Given cygwin is "generally like Linux instead of Windows", and it does not use the Windows libc -- the cygwin DLL is based on newlib -- is this still true?

Also curious what MSYS2 does here, although MSYS2 anyways patches meson for support (you cannot just pip install it).

mesonbuild/compilers/mixins/clike.py
ePirat: Indeed, will re-add this.

mesonbuild/compilers/mixins/clike.py
ePirat: IMO its easier to read this way and makes it easier to quickly debug this by commenting out if blocks. However if you prefer I can change it to return the last function instead.

mesonbuild/compilers/mixins/clike.py
ePirat: Given that this is about the __cdecl calling convention, I would be surprised if it would make a difference.  However I do not have any cygwin env to test it, I can try to set one up tomorrow though and check.

nirbheek: We could write a test for it that involves using a third method of detecting symbols prefixes?

Generate a module with a `foo()` and `_foo()` and `__foo()` and have each of them print a different number when called. Write an executable that loads `foo` at runtime and see which one gets called when you run it?

This is a decent test but a bad configure test because it needs an exe wrapper when cross-compiling.

nirbheek: The module would need to be generated with assembly or something that doesn't go through the linker's symbol mangling.

nirbheek: Another option for the test is to run the `symbolextractor.py` script, which will give the mangled symbol list on each platform.

mesonbuild/compilers/mixins/clike.py
```diff
@@ -925,13 +926,79 @@ def symbols_have_underscore_prefix(self, env: 'Environment') -> bool:
...
+    def _symbols_have_underscore_prefix_list(self, env: 'Environment') -> T.Optional[bool]:
+        '''
+        Check if symbols have underscore prefix by consulting a hardcoded
+        list of cases where we know the results.
+        Return if functions have underscore prefix or None if unknown.
+        '''
+        m = env.machines[self.for_machine]
+        # Darwin always uses the underscore prefix, not matter what
+        if m.is_darwin():
+            return True
```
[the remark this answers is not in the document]
ePirat: I doubt Darwin will ever change this convention and anyway this is only a fallback for the first detection.

mesonbuild/compilers/mixins/clike.py
```diff
@@ -925,13 +926,79 @@ def symbols_have_underscore_prefix(self, env: 'Environment') -> bool:
...
+        # First, try to query the compiler directly
+        result = self._symbols_have_underscore_prefix_define(env)
+        if result is not None:
+            return result
+
+        # Else, try to consult a hardcoded list of cases we know
+        # absolutely have an underscore prefix
+        result = self._symbols_have_underscore_prefix_list(env)
+        if result is not None:
+            return result
```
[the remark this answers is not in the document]
ePirat: Yes because if the compiler macro detection works, thats the most reliable info we can get in any case. As you already noted above we never know what might happen in the future so I prefer only falling back to this, even if it probably will always be correct for the current list entries.

ePirat:
> We could write a test for it that involves using a third method of detecting symbols prefixes?

For some reason Github wont let me answer inline to that one… So yeah that could work if there is a way to tell the compiler to not use a underscore prefix… Not sure if thats possible easily for all compilers we target, and anyway the first check is the most reliable so not sure this hassle is really needed…

I do have another idea to improve this last resort detection, however I did not want to have even more things in this MR that are not strictly needed to fix the issue, as that would just make it harder to review and merge quickly… It's not like after this PR we can't ever improve these things again.

mesonbuild/compilers/mixins/clike.py
ePirat: Oh I see, Github confusingly split this thread up not showing the follow up comments before… 🤯 

Indeed for a unit test in meson we could just check the actual binary using a tool like `nm` or so to verify the list in a test. Or even simpler just rely on the current logic and make sure in the test we have no arguments that can mess with it. (Like -g and similar)

nirbheek: Yeah, that's what I meant. We could write a unit test that uses methods from `symbolextractor.py` which already has a lot of platform-specific code. With that we won't need to worry about testing on cygwin by hand because the unit test will do that for us.

jon-turney: This comment about the C ABI used in 32-bit Cygwin is correct.

I'd be surprised if we ever end up there anyhow, as any Cygwin-targetting compiler anyone might be using should be reporting __USER_LABEL_PREFIX__.

ePirat: I've added unittests now, please let me know if anything else is missing to get this merged.

ePirat:
Test failure is wrong because the test is wrong. The mock environment does not has the right values so `cpu_family`  is incorrect and the wrong value is picked from the list.

If anyone has an idea how to get a proper Environment for a unit test, please let me know.

ePirat:
Should be resolved now, by re-detecting machine info with the proper compiler.

mesonbuild/compilers/mixins/clike.py
```diff
@@ -914,24 +915,90 @@ def symbols_have_underscore_prefix(self, env: 'Environment') -> bool:
...
-                        mlog.debug("Symbols have underscore prefix: YES")
+                        mlog.debug("Underscore prefix check found prefixed function in binary")
                         return True
                     # Else, check if the non-underscored form is present
                     elif symbol_name in line:
-                        mlog.debug("Symbols have underscore prefix: NO")
+                        mlog.debug("Underscore prefix check found non-prefixed function in binary")
                         return False
-        raise RuntimeError(f'BUG: {n!r} check failed unexpectedly')
+        raise RuntimeError('BUG: {n!r} check did not find symbol string in binary')
```
nirbheek, line 935: Why did this remove the f-string?

nirbheek (commented):
LGTM other than the f-string comment.

mesonbuild/compilers/mixins/clike.py
ePirat: Oh, nicely spotted, must have been an accident when changing the message…

PR #9990, opened 2022-02-11
FeatureNew: add mypy type annotations for subproject arg

eli-schwartz:
Use a derived type when passing `subproject` around, so that mypy knows it's actually a SubProject, not a str. This means that passing anything other than a handle to the interpreter state's subproject attribute becomes a type violation, specifically when the order of the *four* different str arguments is typoed.

### DIFF
--- a/mesonbuild/dependencies/base.py
+++ b/mesonbuild/dependencies/base.py
@@ -24,7 +24,7 @@ from .. import mlog
 from ..compilers import clib_langs
 from ..mesonlib import LibType, MachineChoice, MesonException, HoldableObject
 from ..mesonlib import version_compare_many
-from ..interpreterbase import FeatureDeprecated
+#from ..interpreterbase import FeatureDeprecated, FeatureNew
 
 if T.TYPE_CHECKING:
     from .._typing import ImmutableListProtocol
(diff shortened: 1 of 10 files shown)


mesonbuild/dependencies/base.py
```diff
@@ -24,7 +24,7 @@
 from ..compilers import clib_langs
 from ..mesonlib import LibType, MachineChoice, MesonException, HoldableObject
 from ..mesonlib import version_compare_many
-from ..interpreterbase import FeatureDeprecated
+#from ..interpreterbase import FeatureDeprecated, FeatureNew
```
dcbaker, line 27: this looks like something half finished?

mesonbuild/interpreterbase/interpreterbase.py
```diff
@@ -54,6 +54,8 @@
 if T.TYPE_CHECKING:
     from ..interpreter import Interpreter
 
+    SubProject = T.NewType('SubProject', str)
```
dcbaker, line 57: You gnerallly don't put these inside a TYPE_CHECKING block, because you can do `SubProject('subproject_name')` otherwise.

mesonbuild/dependencies/base.py
eli-schwartz: It's part of the FIXME down below, we want to do Feature checks in this file but we cannot.

mesonbuild/interpreterbase/interpreterbase.py
eli-schwartz: Interesting, I didn't realize NewType returned something useful outside of type checking. Apparently `X(y)` does essentially the same as `T.cast(X, y)`, although with the technical downside of actually evaluating a `T.NewType.__init__()` at runtime so you can use it at runtime.

Even as a micro-optimization it's debatable if that is worth caring about. Might as well do as you say.

dcbaker (approved):

eli-schwartz:
Runtime circular imports, lol. I tried.

Well okay, I tried harder.

eli-schwartz:
More cyclic imports. Clearly I'll have to actually take an actual look at this.

But really this just shows why I wasn't really sure about making type checking types available at runtime, lol.

PR #9992, opened 2022-02-11
pkgconfig: Include libdir in .pc files if referenced

thierryreding:
Some packages define variables in their .pc files and those variables
may depend on the value of the libdir variable even if no public or
private libraries are referenced in the .pc file.

Walk the list of variables and check if any of them reference ${libdir}
and if so, make sure to include the libdir definition as well.

### DIFF
--- a/mesonbuild/modules/pkgconfig.py
+++ b/mesonbuild/modules/pkgconfig.py
@@ -342,12 +342,21 @@ class PkgConfigModule(ExtensionModule):
         libdir = PurePath(coredata.get_option(mesonlib.OptionKey('libdir')))
         incdir = PurePath(coredata.get_option(mesonlib.OptionKey('includedir')))
         fname = os.path.join(outdir, pcfile)
+
+        # check if any variables reference libdir, to make sure we include it
+        # when its needed
+        need_libdir = False
+
+        for k, v in variables + unescaped_variables:
+            if '${libdir}' in v:
+                need_libdir = True
+
         with open(fname, 'w', encoding='utf-8') as ofile:
             if not dataonly:
                 ofile.write('prefix={}\n'.format(self._escape(prefix)))
                 if uninstalled:
                     ofile.write('srcdir={}\n'.format(self._escape(srcdir)))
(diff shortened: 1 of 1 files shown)


thierryreding:
@Tachi107 here's a slightly cleaned up version. I couldn't find any projects that use both regular and unescaped variables, but hacking up gst-plugins-base to include a dummy unescaped variable seems to have worked. I'm not sure if it's worth adding a test to meson for this behavior.

xclaesse:
LGTM, thanks.

Tachi107:
Nice, thanks again for noticing the issue :)

eli-schwartz:
Can you take a look at #9994 instead?

This is something I have wanted to do for quite some time, but never got around to. Since it actually would fix a regression now, I decided it is high time I actually sat down and implemented it, so I quickly whipped something up.

Basically, your PR is way too limited in scope. It fixes the regression for gst, but the underlying issue highlights that we should be adding lots of variables! Any time someone references a variable in their variables, we should autogenerate that variable for them so they don't have to.

Just adding back libdir alone is missing so much potential for improvement.

mesonbuild/modules/pkgconfig.py
```diff
@@ -342,12 +342,21 @@ def _generate_pkgconfig_file(self, state, deps, subdirs, name, description,
         libdir = PurePath(coredata.get_option(mesonlib.OptionKey('libdir')))
         incdir = PurePath(coredata.get_option(mesonlib.OptionKey('includedir')))
         fname = os.path.join(outdir, pcfile)
+
+        # check if any variables reference libdir, to make sure we include it
+        # when its needed
+        need_libdir = False
+
+        for k, v in variables + unescaped_variables:
+            if '${libdir}' in v:
```
eli-schwartz, line 351: As mentioned in a review comment to my PR, this will match if the value is `${prefix}/${libdir}`.

My response to that was I *intentionally* only allow it at the beginning, because it's not allowed to appear in the middle -- that would produce broken paths, and thus is a bug in the user's `meson.build` which Meson should not cater to.

PR #9993, opened 2022-02-11
clike compilers: Add default sysroot include directory

thierryreding:
When building against a sysroot, meson currently fails to detect that
the /usr/include directory within that sysroot is a default search
directory. This causes the directory to be added with -isystem to the
compiler command-line which in turn causes include lookup failures.

A minimal test case to reproduce this behaviour is:

  $ echo -e "#include <iostream>\nint main() { return 0; }" > foo.cpp
  $ g++ --sysroot / -isystem /usr/include -o foo foo.cpp
  In file included from /usr/include/c++/11.1.0/ext/string_conversions.h:41,
                   from /usr/include/c++/11.1.0/bits/basic_string.h:6594,
                   from /usr/include/c++/11.1.0/string:55,
                   from /usr/include/c++/11.1.0/bits/locale_classes.h:40,
                   from /usr/include/c++/11.1.0/bits/ios_base.h:41,
                   from /usr/include/c++/11.1.0/ios:42,
                   from /usr/include/c++/11.1.0/ostream:38,
                   from /usr/include/c++/11.1.0/iostream:39,
                   from foo.cpp:1:
  /usr/include/c++/11.1.0/cstdlib:75:15: fatal error: stdlib.h: No such file or directory
     75 | #include_next <stdlib.h>
      |               ^~~~~~~~~~
  compilation terminated.

Omitting the -isystem /usr/include option avoids that error because it
doesn't mess up the include directory search order.

Fix this by adding the sysroot's /usr/include directory to the list of
default search directories when a sysroot is detected. This ensures that
any occurrences of it are filtered out and the include directory search
order is left alone.


thierryreding:
@dcbaker any comments on this?

thierryreding:
@dcbaker I'm still running into this issue with sysrooted cross builds, can you take a look at this?

mesonbuild/compilers/mixins/clike.py
```diff
@@ -100,6 +100,21 @@ def to_native(self, copy: bool = False) -> T.List[str]:
                 new.insert(group_start, '-Wl,--start-group')
         # Remove system/default include paths added with -isystem
         default_dirs = self.compiler.get_default_include_dirs()
+        sys_root = None
+
+        for i, each in enumerate(new):
+            if each.startswith('--sysroot'):
+                if each == '--sysroot' and i + 1 < len(new):
+                    sys_root = new[i + 1]
+                    break
```
dcbaker, line 109: I'd prefer to optimize for the common case, which is that `--sysroot` has a proper argument passed to it:
```python
if each == '--sysroot':
    try:
        sys_root = new[i + 1]
    except IndexError:
        mlog.warn('--sysroot passed to', self.language, 'compiler without an argument')
    break
```
Plus this will warn someone that they're probably doing something they didn't mean to.

Otherwise, this looks good to me.

dcbaker (changes requested):

dcbaker:
Also, I'm sorry I missed this before. I haven't been doing a good job of keeping my github notifications manageable :/

mesonbuild/compilers/mixins/clike.py
thierryreding: Sorry, forgot to actually push this after incorporating your suggestions and then forgot about it. I've rebased this and pushed out the updated commit.

mering:
One should probably also add the library search paths within the sysroot automatically.

thierryreding:
I don't think adding the library search paths is necessary. When --sysroot is specified, the library search paths returned by the compiler will typically be modified to include the sysroot. meson should then automatically pick up the right search paths.

I've uploaded a rebased version and fixed an issue that was flagged by mypy. There's a remaining failing check, but that looks unrelated.

@dcbaker any chance we can get this in? There are a few references to this, so this might actually be helpful for use-cases outside the one I originally encountered.

PR #9994, opened 2022-02-11
pkgconfig module: allow custom variables to reference builtin directories

eli-schwartz:
Automatically generate additional variables and write them into the generated pkg-config file.

This means projects no longer need to manually define the ones they use, which is annoying for dataonly usages (it used to forbid setting the base library-relevant "reserved" ones, and now allows it only for dataonly. But it's bloat to manualy list them anyway).

It also fixes a regression in commit 248e6cf4736ef9ec636228da66c28f9be03aa74f which caused libdir to not be set, and to be unsettable, if the pkg-config file has no libraries but uses the ${libdir} expansion in a custom variable. This could be considered likely a case for dataonly, but it's not guaranteed.

Obsoletes / Closes #9992

### DIFF
--- a/mesonbuild/modules/pkgconfig.py
+++ b/mesonbuild/modules/pkgconfig.py
@@ -22,8 +22,9 @@ from .. import build
 from .. import dependencies
 from .. import mesonlib
 from .. import mlog
+from ..coredata import BUILTIN_DIR_OPTIONS
 from ..dependencies import ThreadDependency
-from ..interpreterbase import permittedKwargs, FeatureNew, FeatureNewKwargs
+from ..interpreterbase import permittedKwargs, FeatureNew, FeatureDeprecated, FeatureNewKwargs
 
 if T.TYPE_CHECKING:
     from . import ModuleState
@@ -329,6 +330,41 @@ class PkgConfigModule(ExtensionModule):
                                  url, version, pcfile, conflicts, variables,
                                  unescaped_variables, uninstalled=False, dataonly=False):
         coredata = state.environment.get_coredata()
+        referenced_vars = set()
+        optnames = [x.name for x in BUILTIN_DIR_OPTIONS.keys()]
+
+        if not dataonly:
+            # includedir is always implied, although libdir may not be
+            # needed for header-only libraries
+            referenced_vars |= {'prefix', 'includedir'}
+            if deps.pub_libs or deps.priv_libs:
+                referenced_vars |= {'libdir'}
+        # also automatically infer variables referenced in other variables
+        implicit_vars_warning = False
+        redundant_vars_warning = False
+        varnames = set()
+        varstrings = set()
+        for k, v in variables + unescaped_variables:
+            varnames |= {k}
+            varstrings |= {v}
+        for optname in optnames:
+            optvar = f'${{{optname}}}'
+            if any(x.startswith(optvar) for x in varstrings):
+                if optname in varnames:
+                    redundant_vars_warning = True
+                else:
+                    # these 3 vars were always "implicit"
+                    if dataonly or optname not in {'prefix', 'includedir', 'libdir'}:
+                        implicit_vars_warning = True
+                    referenced_vars |= {'prefix', optname}
+        if redundant_vars_warning:
+            FeatureDeprecated.single_use('pkgconfig.generate variable for builtin directories', '0.62.0',
+                                         state.subproject, 'They will be automatically included when referenced',
+                                         state.current_node)
+        if implicit_vars_warning:
+            FeatureNew.single_use('pkgconfig.generate implicit variable for builtin directories', '0.62.0',
+                                  state.subproject, location=state.current_node)
+
         if uninstalled:
             outdir = os.path.join(state.environment.build_dir, 'meson-uninstalled')
             if not os.path.exists(outdir):
@@ -338,18 +374,17 @@ class PkgConfigModule(ExtensionModule):
         else:
             outdir = state.environment.scratch_dir
             prefix = PurePath(coredata.get_option(mesonlib.OptionKey('prefix')))
-        # These always return paths relative to prefix
-        libdir = PurePath(coredata.get_option(mesonlib.OptionKey('libdir')))
-        incdir = PurePath(coredata.get_option(mesonlib.OptionKey('includedir')))
         fname = os.path.join(outdir, pcfile)
         with open(fname, 'w', encoding='utf-8') as ofile:
-            if not dataonly:
-                ofile.write('prefix={}\n'.format(self._escape(prefix)))
-                if uninstalled:
-                    ofile.write('srcdir={}\n'.format(self._escape(srcdir)))
-                if deps.pub_libs or deps.priv_libs:
-                    ofile.write('libdir={}\n'.format(self._escape('${prefix}' / libdir)))
-                ofile.write('includedir={}\n'.format(self._escape('${prefix}' / incdir)))
+            for optname in optnames:
+                if optname in referenced_vars - varnames:
+                    if optname == 'prefix':
+                        ofile.write('prefix={}\n'.format(self._escape(prefix)))
+                    else:
+                        dirpath = PurePath(coredata.get_option(mesonlib.OptionKey(optname)))
+                        ofile.write('{}={}\n'.format(optname, self._escape('${prefix}' / dirpath)))
+            if uninstalled and not dataonly:
+                ofile.write('srcdir={}\n'.format(self._escape(srcdir)))
             if variables or unescaped_variables:
                 ofile.write('\n')
             for k, v in variables:
--- a/test cases/common/44 pkgconfig-gen/meson.build
+++ b/test cases/common/44 pkgconfig-gen/meson.build
@@ -1,4 +1,4 @@
-project('pkgconfig-gen', 'c')
+project('pkgconfig-gen', 'c', meson_version: '>=0.60.0')
 
 # Some CI runners does not have zlib, just skip them as we need some common
 # external dependency.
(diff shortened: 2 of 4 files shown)


eli-schwartz:
Needs release notes and a FeatureNew check, will get to that soon.

mesonbuild/modules/pkgconfig.py
```diff
@@ -329,6 +330,27 @@ def _generate_pkgconfig_file(self, state, deps, subdirs, name, description,
...
+        # also automatically infer variables referenced in other variables
+        varnames = []
+        varstrings = []
+        for k, v in variables + unescaped_variables:
+            varnames.append(k)
+            varstrings.append(v)
+        for optname in optnames:
+            if optname not in varnames:
+                optvar = f'${{{optname}}}'
+                if any(x.startswith(optvar) for x in varstrings):
```
xclaesse, line 351: Shouldn't it be `optvar in x` instead? The variable could be in the middle of the value, like `foo=${prefix}/${libdir}`.

mesonbuild/modules/pkgconfig.py
```diff
@@ -338,18 +360,17 @@ def _generate_pkgconfig_file(self, state, deps, subdirs, name, description,
...
         with open(fname, 'w', encoding='utf-8') as ofile:
-            if not dataonly:
-                ofile.write('prefix={}\n'.format(self._escape(prefix)))
-                if uninstalled:
-                    ofile.write('srcdir={}\n'.format(self._escape(srcdir)))
-                if deps.pub_libs or deps.priv_libs:
-                    ofile.write('libdir={}\n'.format(self._escape('${prefix}' / libdir)))
-                ofile.write('includedir={}\n'.format(self._escape('${prefix}' / incdir)))
+            for optname in optnames:
+                if optname in referenced_vars:
```
xclaesse, line 366: can't this be simplified to `for optname in referenced_vars` ?

mesonbuild/modules/pkgconfig.py
```diff
@@ -329,6 +330,27 @@ def _generate_pkgconfig_file(self, state, deps, subdirs, name, description,
                                  url, version, pcfile, conflicts, variables,
                                  unescaped_variables, uninstalled=False, dataonly=False):
         coredata = state.environment.get_coredata()
+        referenced_vars = {'prefix'}
+        optnames = [x.name for x in BUILTIN_DIR_OPTIONS.keys()]
+
+        if not dataonly:
+            # includedir is always implied, although libdir may not be
+            # needed for header-only libraries
+            referenced_vars |= {'includedir'}
```
xclaesse, line 339: We can always add `prefix` in `referenced_vars` too.

mesonbuild/modules/pkgconfig.py
```diff
@@ -338,18 +360,17 @@ def _generate_pkgconfig_file(self, state, deps, subdirs, name, description,
...
-                if deps.pub_libs or deps.priv_libs:
-                    ofile.write('libdir={}\n'.format(self._escape('${prefix}' / libdir)))
-                ofile.write('includedir={}\n'.format(self._escape('${prefix}' / incdir)))
+            for optname in optnames:
+                if optname in referenced_vars:
+                    if optname == 'prefix':
+                        ofile.write('prefix={}\n'.format(self._escape(prefix)))
+                    else:
+                        dirpath = PurePath(coredata.get_option(mesonlib.OptionKey(optname)))
+                        ofile.write('{}={}\n'.format(optname, self._escape('${prefix}' / dirpath)))
```
xclaesse, line 371: I think it needs to be `dirpath.as_posix()`

xclaesse (commented):
I like the idea, it's more generic indeed.

mesonbuild/modules/pkgconfig.py
xclaesse: I think that's equivalent only if we add `prefix` in `referenced_vars` as suggested below.

mesonbuild/modules/pkgconfig.py
eli-schwartz: That would be broken, because you'd end up with for example:

 ```
prefix=/usr
libdir=${prefix}/lib
foo=${prefix}/${libdir}/foo
```

And then foo would expand to `/usr/usr/lib/foo`.

Any user doing that is better served by the pkgconfig module generating a broken pkg-config file that errors out, than by a pkg-config file that subtly returns the wrong answer.

File: mesonbuild/modules/pkgconfig.py (hunk 2 above)
Comment (eli-schwartz) [reply mesonbuild/modules/pkgconfig.py:366]: No, because sets are not ordered and I want to do this in the order that BUILTIN_DIR_OPTIONS returns. For the sake of neatness, and also because it is either that or `sorted(referenced_vars)` to ensure that the output is stable.

File: mesonbuild/modules/pkgconfig.py (hunk 1 above)
Comment (xclaesse) [reply mesonbuild/modules/pkgconfig.py:351]: Good point, nevermind.

File: mesonbuild/modules/pkgconfig.py (hunk 3 above)
Comment (eli-schwartz) [reply mesonbuild/modules/pkgconfig.py:339]: It's initialized with that, because we almost always need to include it. Any other variable is bound to refer to it, and explicitly including it might get a teensy bit hairy.

File: mesonbuild/modules/pkgconfig.py (hunk 4 above)
Comment (eli-schwartz) [reply mesonbuild/modules/pkgconfig.py:371]: Nope, that's explicitly handled by `self._escape()` and the argument *to* `self._escape` needs to use the overloaded `/` to join itself.

Take a look at the old version of this code, it didn't use `.as_posix()` either.

Comment (xclaesse) [reply mesonbuild/modules/pkgconfig.py:371]: nevermind, was already like that so I gave it a try and it converts to posix already.

File: mesonbuild/modules/pkgconfig.py (hunk 2 above)
Comment (dcbaker) [reply mesonbuild/modules/pkgconfig.py]: mesonlib.OrderedSet?

File: mesonbuild/modules/pkgconfig.py (hunk 3 above)
Comment (eli-schwartz) [reply mesonbuild/modules/pkgconfig.py]: Oh what the heck, I included it. :)

File: mesonbuild/modules/pkgconfig.py (hunk 2 above)
Comment (eli-schwartz) [reply mesonbuild/modules/pkgconfig.py]: TBH that seems a solution in search of a problem.

Comment (eli-schwartz) [reply mesonbuild/modules/pkgconfig.py]: Actually mesonlib.OrderedSet would probably result in ugly looking files. It would be stable, but depending on the generate() invocation always have libdir/includedir out of order compared to `coredata.BUILTIN_DIR_OPTIONS`.

File: mesonbuild/modules/pkgconfig.py [hunk 5]
```diff
@@ -329,6 +330,39 @@ def _generate_pkgconfig_file(self, state, deps, subdirs, name, description,
...
+                    implicit_vars_warning = True
+                    referenced_vars |= {'prefix', optname}
+        if redundant_vars_warning:
+            FeatureDeprecated.single_use('pkgconfig.generate variable for builtin directories', '0.62.0',
+                                         state.subproject, 'They will be automatically included when referenced',
+                                         state.current_node)
+        if implicit_vars_warning:
+            FeatureNew.single_use('pkgconfig.generate implicit variable for builtin directories', '0.62.0',
+                                  state.subproject, location=state.current_node)
+
```
thierryreding, line 365: This isn't entirely accurate if libdir is the only implicit variable. For example, the build of the gst-plugins-base package that I mentioned in #9992 now complains loudly multiple times, yet prior to the breaking commit this was just fine.

That said, this works fine and is a nice fix and given that this is only a warning I don't think it's worth being all that nitpicky about it, especially since the extra check needed would complicate things.

eli-schwartz: Oh right, that is an annoying confluence of edge cases indeed.

mesonbuild/modules/pkgconfig.py
xclaesse: ok, my comment was just a nitpicking to be honest. Since there is a reason for doing this that way, I'm fine with it.

mesonbuild/modules/pkgconfig.py
xclaesse: I think this is a valid point and not really an edge case. prefix, libdir and includedir were always guaranteed to be part of the generated file, so using them implicitly is not a new behaviour.

eli-schwartz: Alright, I added a suppression for this, and a test case that checks both that the common case is warned about, and that your case is silent.

test cases/common/44 pkgconfig-gen/meson.build
```diff
@@ -149,3 +149,17 @@ ct = custom_target('stat3',
...
+  name : 'libvartest',
+  description : 'Check that implicit vars are created',
+  version : libver,
+  variables: ['datadir=${prefix}/data', 'foo=${datadir}/foo', 'bar=${bindir}/bar']
+)
+pkgg.generate(
+  name : 'libvartest2',
+  description : 'Check that libdir is not an implicit var',
+  version : libver,
+  variables: ['bar=${libdir}/bar']
```
eli-schwartz, line 164: This should not generate a warning!

test cases/common/44 pkgconfig-gen/test.json
```diff
@@ -7,11 +7,26 @@
...
     {"type": "file", "file": "usr/lib/pkgconfig/ct.pc"},
     {"type": "file", "file": "usr/lib/pkgconfig/ct0.pc"}
+  ],
+  "stdout": [
+    {
+      "line": "test cases/common/44 pkgconfig-gen/meson.build:158: WARNING: Project targeting '>=0.60.0' but tried to use feature introduced in '0.62.0': pkgconfig.generate implicit variable for builtin directories."
+    },
+    {
+      "line": "test cases/common/44 pkgconfig-gen/meson.build:164: WARNING: Project targeting '>=0.60.0' but tried to use feature introduced in '0.62.0': pkgconfig.generate implicit variable for builtin directories.",
+      "count": 0
```
eli-schwartz, line 25: This errors out if it is, somehow, still a warning.

mesonbuild/modules/pkgconfig.py
eli-schwartz: Also yes it is absolutely an edge case. ***Edge cases should be handled, not left for people to cut themselves on***, which is why I was always going to fix this and just needed to get home and work on it... my comment simply meant that "hmm, I can see why I overlooked that".

xclaesse: oh ok, I understood your comment as it's rare enough that we could leave it like that.

xclaesse:
Lgtm

PR #9997, opened 2022-02-13
validate the literal directory "subprojects" when checking sandbox violations

eli-schwartz:
We do not want anyone touching this entire directory tree, but due to the way it was implemented, we only checked if its direct parent was a subproject violation. This generally worked, unless people tried to add `subprojects/` as an include directory.

Patch this hole. It now provides the same warning any sandbox violation does (but is not currently an error, just a "will become an error in the future").


dcbaker (approved):
Nice catch

PR #10003, opened 2022-02-14
Support finding toolchain relative to source root

gaudetja:
Adds in a rudimentary replacement for `meson.source_root()' in the processed machine files. This was a "least intrusive" way to enable relative path definitions for toolchains in machine files. Another approach might be to pass additional information from environment.py to coredata.py when the machine files are read in. To test the change out, a cross file similar to below can be used:
```
[constants]
repo = 'meson.source_root()'
path = repo / 'relative/path/bin'
arch = 'arm-none-eabi-'

[binaries]
c = [path / arch + 'gcc']
ar = [path / arch + 'ar']
strip = [path / arch + 'strip']

[host_machine]
system = 'baremetal'
cpu_family = 'arm'
cpu = 'armv7'
endian = 'little'
```
Draft status will be removed if there is sufficient interest in adding this feature. I will be happy to update documentation and tests at that point.

### DIFF
--- a/docs/markdown/Machine-files.md
+++ b/docs/markdown/Machine-files.md
@@ -59,11 +59,13 @@ String and list concatenation is supported using the `+` operator,
 joining paths is supported using the `/` operator. Entries defined in
 the `[constants]` section can be used in any other section (they are
 always parsed first), entries in any other section can be used only
-within that same section and only after it has been defined.
+within that same section and only after it has been defined. The root
+of the source tree is supported using `@SOURCE_ROOT@`.
 
 ```ini
 [constants]
 toolchain = '/toolchain'
+repo_toolchain = '@SOURCE_ROOT@/tools'
 common_flags = ['--sysroot=' + toolchain / 'sysroot']
 
 [properties]
@@ -72,6 +74,7 @@ cpp_args = c_args + ['-DSOMETHING_ELSE']
 
 [binaries]
 c = toolchain / 'gcc'
+somebin = repo_toolchain / 'somebin'
 ```
 
 This can be useful with cross file composition as well. A generic
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -924,6 +924,8 @@ class MachineFileParser():
                 raise EnvironmentException(f'Malformed variable name {entry!r} in machine file.')
             # Windows paths...
             value = value.replace('\\', '\\\\')
+            # replace 'function call'
+            value = value.replace('meson.source_root()', os.getcwd())
             try:
                 ast = mparser.Parser(value, 'machinefile').parse()
                 res = self._evaluate_statement(ast.lines[0])
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -924,8 +924,8 @@ class MachineFileParser():
                 raise EnvironmentException(f'Malformed variable name {entry!r} in machine file.')
             # Windows paths...
             value = value.replace('\\', '\\\\')
-            # replace 'function call'
-            value = value.replace('meson.source_root()', os.getcwd())
+            # Fill in path to source root here
+            value = value.replace('@SOURCE_ROOT@', os.getcwd())
             try:
                 ast = mparser.Parser(value, 'machinefile').parse()
                 res = self._evaluate_statement(ast.lines[0])


tristan957:
I think this is a good change, but I am not sure that `meson.project_source_root()` is the best idea. I think `@SOURCE_ROOT@`  is better which is similar to the `custom_target()` stuff.

xclaesse:
See also https://github.com/mesonbuild/meson/pull/9223.

xclaesse:
`@SOURCE_ROOT@` looks much better indeed. And this PR could actually be a much better alternative than what has been proposed in https://github.com/mesonbuild/meson/pull/9223. Would have to check if it covers the same use-cases: https://github.com/mesonbuild/meson/issues/3439.

CC @miebka

miebka:
In #9223, the binaries are resolved relative to the machine file location. 
I also started #9223 with the requirement to resolve the binaries relative to the meson root directory, but after thinking about it more than twice, I concluded that the benefit would be greater if the binaries were resolved relative to the machine file, since the toolchain, including the machine file, could then be located anywhere. 
If I want to switch between 2 different compiler versions, I just need to run meson setup with --cross-file pointing to the other installation.

docs/markdown/Machine-files.md
```diff
@@ -59,11 +59,13 @@ String and list concatenation is supported using the `+` operator,
 joining paths is supported using the `/` operator. Entries defined in
 the `[constants]` section can be used in any other section (they are
 always parsed first), entries in any other section can be used only
-within that same section and only after it has been defined.
+within that same section and only after it has been defined. The root
+of the source tree is supported using `@SOURCE_ROOT@`.
```
xclaesse, line 63: This should say "Since 0.62.0 ..."

mesonbuild/coredata.py
```diff
@@ -924,6 +924,8 @@ def _parse_section(self, s):
                 raise EnvironmentException(f'Malformed variable name {entry!r} in machine file.')
             # Windows paths...
             value = value.replace('\\', '\\\\')
+            # Fill in path to source root here
+            value = value.replace('@SOURCE_ROOT@', os.getcwd())
```
xclaesse, line 928: Nothing guarantees that `os.getcwd()` is the source root. You should pass `self.source_dir` from environment.py to `parse_machine_files()` down to MachineFileParser.

xclaesse:
While at it, I think it would make sense to add `@CURRENT_FILE_DIR@` to make relative to the location of the cross file? Or if someone has a better name to suggest? I think the cross file often goes together with a toolchain, so it makes a lot of sense to have path relative to cross file too.

tristan957:
@xclaesse `@CURRENT_SOURCE_DIR@`?

xclaesse:
> @xclaesse `@CURRENT_SOURCE_DIR@`? What does `custom_target()` use?

But the cross file is not always even in source dir. You can do `meson --cross-file /path/to/installed/toolchain/crossfile.txt`.

xclaesse:
oh wait, won't work with composition of multiple cross files in different locations... :/

tristan957:
@DIRNAME@ as an ode to the command line tool

gaudetja:
Given the number of pull requests attempting to bring in a feature similar to each other, is there a good place to have a meta discussion about the end goal? I'll admit that the way I use meson may be on the edge of the project's ideals, so I do not have a good vision of a solution that expands to other use cases.

A possible user story to contribute would be:
"As a developer targeting embedded platforms, I need to be able to lock down the version of compiler tools used to generate a release binary. Additionally I would prefer if this could be done without the need for external tools (e.g. sed) running before a `meson setup`.

tristan957:
You could create a discussion here on github. I think if we want to have `@DIRNAME@` to include the use case from @xclaesse and the Android NDK, the `@SOURCE_ROOT@` also talked about in this PR makes a lot of sense. The previous PR seems like a worse solution to the same problem. Although you could have both PRs.

tristan957:
Would be nice if there was a common function to perform this text substitution instead of having this random replace.

`mesonlib.substitute(my_str, {'SOURCE_ROOT': source_root} )`

tristan957:
Why close?

Nomura-RH:
> Why close?

Maybe it's because it isn't finished :c

PR #10004, opened 2022-02-15
mtest: print stderr of TAP/Rust tests in verbose/non-parallel mode

bonzini:
Verbose, non-parallel tests generally print their output as they run, rather than
after they finish.  This however is not the case if stdout of the test is parsed
as is the case for TAP and Rust tests.  In this case, the output during the run
is the list of the subtests, but stderr still has to be printed after the test
finishes.

### DIFF
--- a/mesonbuild/mtest.py
+++ b/mesonbuild/mtest.py
@@ -606,11 +606,9 @@ class ConsoleLogger(TestLogger):
                                  max_left_width=self.max_left_width,
                                  right=test.res.get_text(mlog.colorize_console())))
             print(test.res.get_command_marker() + test.cmdline)
-            if test.needs_parsing:
-                pass
-            elif not test.is_parallel:
+            if test.direct_stdout:
                 print(self.output_start, flush=True)
-            else:
+            elif not test.needs_parsing:
                 print(flush=True)
 
         self.started_tests += 1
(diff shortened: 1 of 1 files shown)


mesonbuild/mtest.py
```diff
@@ -665,9 +665,8 @@ def log(self, harness: 'TestHarness', result: 'TestRun') -> None:
 
         if not harness.options.quiet or not result.res.is_ok():
             self.flush()
-            if result.verbose and not result.is_parallel and result.cmdline:
-                if not result.needs_parsing:
-                    print(self.output_end)
+            if result.verbose and not result.is_parallel and result.cmdline and not result.needs_parsing:
```
jpakkane, line 668: This condition is getting unwieldy. It would be better to extract it to its own function with a proper descriptive name instead.

bonzini: Good point, I actually did that on top of this commit (because with the old condition there's no common code to extract).

bonzini:
@nirbheek I see that a 0.61.3 milestone exists, can this be included?

nirbheek:
This seems to depend on https://github.com/mesonbuild/meson/pull/9834, removing from milestone.

bonzini:
It doesn't depend, but it doesn't apply cleanly. I created #10112 with the backport.

PR #10007, opened 2022-02-15
Check if test executable exists

### LINKED ISSUE #10006: Unhandled Python exception on meson test --no-rebuild when test executable not built
**Describe the bug**
Unhandled Python exception when running `meson test --no-rebuild` when the test executable is not up to date.

**To Reproduce**

```
echo "project('x','c')"                > meson.build
echo "test('t', executable('missing'))" >> meson.build
meson build
meson test -C build --no-rebuild
....long stack trace....
FileNotFoundError: [Errno 2] No such file or directory: '..../build/missing'

```

**Expected behavior**
Either skip the test with the missing executable, or abort with a tidier error message.

**system parameters**
* Ubuntu 21.10
* Python 3.9.10
* `meson --version 0.61.2`

mwm126:
Resolve #10006

### DIFF
--- a/mesonbuild/mtest.py
+++ b/mesonbuild/mtest.py
@@ -1388,6 +1388,10 @@ class SingleTestRunner:
         if self.cmd is None:
             skip_stdout = 'Not run because can not execute cross compiled binaries.'
             harness.log_start_test(self.runobj)
+            self.runobj.complete_skip(skip_stdout)        
+        elif not (Path(self.cmd[0])).is_file():
+            skip_stdout = 'Not run because test binary does not exist.'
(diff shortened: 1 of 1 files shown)


jpakkane:
Would it be better to check the existance of all needed binaries and hard error out if any of them are missing? Ignoring missing ones can lead people to thinking that their test suite has passed when in reality is hasn't.

eli-schwartz:
I implemented a fix that causes the testsuite to immediately abort when test programs are missing. IMHO it's not very useful to check this in the middle of the test run, and it's definitely not right to flag this as a `SKIP`ped test, though in theory it would be okay to flag it as an `ERROR`ing test.

PR #10008, opened 2022-02-15
mtest: raise informative error message when test program doesn't exist

### LINKED ISSUE #10006: Unhandled Python exception on meson test --no-rebuild when test executable not built

eli-schwartz:
If --no-rebuild is used, the test program might not exist, spawning a FileNotFoundError inside a long traceback rooted in subprocess.Popen trying to run that test program. Current versions of Meson even say it's an "unhandled python exception".

But we can do one better and actually tell the user what is wrong, why, and what to do to fix it. And we can do so before getting waist deep in partially running tests.

Fixes #10006
Closes #10007

### DIFF
--- a/mesonbuild/mtest.py
+++ b/mesonbuild/mtest.py
@@ -1347,9 +1347,12 @@ class SingleTestRunner:
             self.console_mode = ConsoleUser.LOGGER
 
     def _get_test_cmd(self) -> T.Optional[T.List[str]]:
-        if self.test.fname[0].endswith('.jar'):
+        testentry = self.test.fname[0]
+        if self.options.no_rebuild and not os.path.isfile(testentry):
+            raise TestException(f'The test program {testentry!r} does not exist. Cannot run tests before building them.')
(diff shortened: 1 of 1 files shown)


eli-schwartz:
Can't be backported, see linked ticket -- solution to the linked ticket changes the mtest format which means it isn't a candidate for a point release.

eli-schwartz:
I think we accidentally forgot to remove this commit from the backport branch. :(

PR #10009, opened 2022-02-16
Add public kwarg to set_variable()

xclaesse:
This deprecate subp.get_variable() unless the variable has been
explicitly set public by the subproject. This avoids poking at random
internal variables of a subproject and instead define what is their
public build API.

### DIFF
--- a/mesonbuild/interpreter/interpreter.py
+++ b/mesonbuild/interpreter/interpreter.py
@@ -295,6 +295,7 @@ class Interpreter(InterpreterBase, HoldableObject):
         self.build_func_dict()
         self.build_holder_map()
         self.user_defined_options = user_defined_options
+        self.public_variables: T.Set[str] = set()
 
         # build_def_files needs to be defined before parse_project is called
         #
@@ -2897,12 +2898,15 @@ This will become a hard error in the future.''', location=self.current_node)
         return self.subproject != ''
 
     @typed_pos_args('set_variable', str, object)
-    @noKwargs
+    @typed_kwargs('set_variable',
+        KwargInfo('public', bool, default=False, since='0.62.0'))
     @noArgsFlattening
     @noSecondLevelHolderResolving
     def func_set_variable(self, node: mparser.BaseNode, args: T.Tuple[str, object], kwargs: 'TYPE_kwargs') -> None:
         varname, value = args
         self.set_variable(varname, value, holderify=True)
+        if kwargs['public']:
+            self.public_variables.add(varname)
 
     @typed_pos_args('get_variable', (str, Disabler), optargs=[object])
     @noKwargs
--- a/mesonbuild/interpreter/interpreterobjects.py
+++ b/mesonbuild/interpreter/interpreterobjects.py
@@ -706,7 +706,13 @@ class SubprojectHolder(MesonInterpreterObject):
         if not isinstance(varname, str):
             raise InterpreterException('Get_variable first argument must be a string.')
         try:
-            return self.held_object.variables[varname]
+            value = self.held_object.variables[varname]
+            if varname not in self.held_object.public_variables:
+                FeatureDeprecated.single_use(
+                    'subproject.get_variable() for variable not marked public',
+                    '0.62.0', self.subproject,
+                    f'The subproject should mark public variables using set_variable({varname}, value, public: true)',self.current_node)
+            return value
         except KeyError:
             pass
 
--- a/test cases/common/241 set and get variable/meson.build
+++ b/test cases/common/241 set and get variable/meson.build
@@ -69,3 +69,7 @@ assert(not is_disabler(list_t1))
 assert(not is_disabler(list_t2))
 assert(not is_disabler(list_t3))
 assert(not is_disabler(list_t4))
+
+sub = subproject('sub')
+assert(sub.get_variable('private') == 'private')
+assert(sub.get_variable('public') == 'public')
--- a/test cases/common/241 set and get variable/subprojects/sub/meson.build
+++ b/test cases/common/241 set and get variable/subprojects/sub/meson.build
@@ -1,0 +1,5 @@
+project('sub')
+
+private = 'private'
+public = 'public'
+set_variable('public', public, public: true)


eli-schwartz:
This is a terrible idea for a number of reasons, including several that were mentioned on IRC.

The biggest issue (aside for my wholehearted belief that there is no valid reason to have some variables be "not public") is the fact that this doesn't even do a good job of doing this!

No one wants to declare a complicated variable using ethical syntax like `foo = complicated_function(15_kwargs: ........)`, then declare it again using something both obscure and cursed like set_variable. It is literally designed to make me want to gouge out my eyes if I ever read a meson.build that does this. :)

tp-m:
I think there's something to be said for making variable export explicit and warning if it wasn't exported explicitly (with a looong transition period before even thinking about making that fail).

I'm not a fan of exporting it via `set_variable()` though, that's just weird IMHO.

As mentioned on IRC, I'd do it explicitly via a new `export_variable('')` then if we want to do this. (Not that I feel it's particularly pressing)

xclaesse:
The more I think about this, the more I want to deprecate `sub.get_variable()` completely with no way to even mark some variables as public/exported.

For dependencies we used to need to poke at variables, but that got fixed by using `meson.override_dependency()`.

The other use-case I see is to take files from source tree. Meson forbid access them by path string (sandboxing violation) so we want to take them via a `var = files()` which feels like just working around a limitation.

A valid point made is that we should allow accessing files that the subproject is going to install, because that's not really a sandbox violation, those files are part of the public interface of that subproject. However the source tree layout is not always the same as the installed layout, and in my opinion having to rely on the source tree layout is a sandbox violation. Ideally we should be able to access to those files the same way regardless if that dependency comes from a subproject or installed on the system, avoiding those ugly `if dep.type_name() == 'internal'`.

It feels like we should be able to add API on the dependency object to retrieve a `files()` object for installed files that would transparently translate to the path inside source/build tree in the case of an InternalDependency object. I'm not sure exactly how that API would look like, and if that could cover all use-cases, but what do you think?

Akaricchi:
I am strongly against limiting access to installed files only, unless some mechanism is added to prevent actual installation of subprojects and/or certain tags via `meson install` *by default*. `--skip-subprojects` is a nonstarter for me. I have over a dozen subprojects that I use for static linking purposes only; none of their files should be installed ever. I use the install feature to generate binary packages for multiple platforms, where a bunch of headers, static libraries, and other intermediate build artifacts obviously do not belong. Since I maintain the meson build definitions for all of those subprojects, my solution so far has been to disable all installation there.

Unfortunately, I also have to rely on `subproject.get_variable()` for some things. In particular, [SPIRV-Headers](https://github.com/taisei-project/SPIRV-Headers/blob/meson-1.5.4/meson.build) has to export a bunch of json files for other projects to generate some code from. In your world, I would be forced to install those files together with my main project by default. Which is stupid, because they are just an implementation detail of some of my dependencies that is in no way required at runtime.

Honestly, I always thought it dumb that the main project can't control what gets installed by subprojects by default. It's one of the reasons why I don't feel comfortable using wrapdb. `--tags` and `--skip-subprojects` are nice for package maintainers, but being unable to provide reasonable defaults still sucks. Perhaps adding some built-in options to act as default values for those parameters would be a good start; I could then set them in `project(default_options : …)`. There are a few minor problems with that too, but it's better than nothing.

Akaricchi:
Also, what about `declare_dependency(variables : …)` and `dep.get_variable()`? That's another potential "sandbox violation" vector. Do you want to deprecate that as well? It's probably even worse since it's limited to string values.

> For dependencies we used to need to poke at variables, but that got fixed by using meson.override_dependency().

Sometimes you want to force using a specific dependency from a subproject. Maybe you have a private fork of something like Lua that's incompatible with the upstream version, and you intend to link it statically. You don't want to have an option to accidentally link to the system version, because that won't work. You also may not want to override the dependency globally — what if one of your dependencies expects to use the standard version for some unrelated purpose? `sub.get_variable()` seems like the best way to express this right now. The only alternative I see is renaming the incompatible fork, and hoping the name you've chosen is unique enough to not clash with some random system thing.

Akaricchi:
> I think skipping install of subprojects was an anti-feature. The real use-case is to install only runtime files which is now possible with meson install --targs runtime. For subprojects that only generate static libraries, that won't install anything.

The problem is that a bare `meson install` will install a lot of junk, and the user should not have to qualify it with --lots --of --arguments for that not to happen. A `meson install` should install my application, whatever is needed to run it, its documentation, and its public headers (if applicable). It should *not* install its internal build-time dependencies, that makes no sense. Whether I use a subproject or a system-installed library should not affect the behavior of `meson install` (with the exception of building a shared library from a subproject and linking to it).

> How does it work when that project is not used as a subproject then? You have to install those files. Again that feels like a use-case for installing only runtime files.

Yes, I'd have to install those files — when installing that dependency system-wide. Not together with the project that uses it.

eli-schwartz:
>  A `meson install` should install my application, whatever is needed to run it, its documentation, and its public headers (if applicable). It should not install its internal build-time dependencies, that makes no sense. Whether I use a subproject or a system-installed library should not affect the behavior of `meson install` (with the exception of building a shared library from a subproject and linking to it).

While I do 100% agree with you, gnome made early, heavy use of subprojects to ship large ecosystem bundles in what would be better termed a "metaproject" rather than a project with dependency fallbacks.

Despite my objections, they really want to preserve the current semantics ;) and the compromise solution was letting people opt in to only installing the main project, or using install tags (which don't let you install the main project's devel files while sticking to the subproject runtime files).

It's... better than before.

Some other people just code their entire subproject with `if not meson.is_subproject()`, which is its own type of exciting.

xclaesse:
Maybe in your specific use-case you think you want to skip install of subprojects completely, but that's not a general solution. Subprojects could be installing data files that are needed at runtime even if you static link them, icons, config file, translation files, etc... It is at best a "it works for me" but that's never going to be a general solution.

Akaricchi:
> That's purely a personal opinion. Someone else will say it should not install doc/headers by default, other will say they want static libs by default etc.... The only generic solution is installation tags.

Yeah, which is why I said:

> Perhaps adding some built-in options to act as default values for those parameters would be a good start; I could then set them in `project(default_options : …)`

I don't want to impose my view on every project. I just want to be in control of what is installed by default.

> But anyway, this whole conversation is completely offtopic, whatever is the solution we come with, it's never going to be to set install: false on targets. So as far as getting files from subprojects is concerned, we can still always count on the fact that only installable files make sense to be taken from parent project.

This conversation is relevant because as-is, your idea breaks my use-case. There either must be a "blessed" way to retrieve non-installable files from subprojects, or (much more preferably) a way for the parent project to control what's installed by default. I believe this must be addressed before doing anything like deprecating `sub.get_variable()`.

PR #10010, opened 2022-02-16
tests: various improvements to make alpine packaging work

eli-schwartz:
See https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/29489

changes:
- make ninja detection faster and more robust. Actually checks, even on CI
- remove pointless dependency on bash in one test
- a couple of changes to how we check whether we are in CI. Distinguishes between "CI environment, so do UX and accessibility quirks" and "Meson CI environment, don't allow tests to be surprisingly skipped"
- fix the testsuite on musl libc

and finally:
- runtime fix for detecting `dependency('intl')`

### DIFF
--- a/run_tests.py
+++ b/run_tests.py
@@ -46,13 +46,13 @@ from mesonbuild.mesonlib import OptionKey, setup_vsenv
 
 NINJA_1_9_OR_NEWER = False
 NINJA_CMD = None
-# If we're on CI, just assume we have ninja in PATH and it's new enough because
-# we provide that. This avoids having to detect ninja for every subprocess unit
-# test that we run.
-if 'CI' in os.environ:
-    NINJA_1_9_OR_NEWER = True
-    NINJA_CMD = ['ninja']
-else:
+# If we're on CI, detecting ninja for every subprocess unit test that we run is slow
+# Optimize this by respecting $NINJA and skipping detection, then exporting it on
+# first run.
+try:
+    NINJA_1_9_OR_NEWER = bool(int(os.environ['NINJA_1_9_OR_NEWER']))
+    NINJA_CMD = [os.environ['NINJA']]
+except (KeyError, ValueError):
     # Look for 1.9 to see if https://github.com/ninja-build/ninja/issues/1219
     # is fixed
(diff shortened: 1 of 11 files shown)


eli-schwartz:
@psycholance

run_tests.py
```diff
@@ -46,13 +46,13 @@
...
-if 'CI' in os.environ:
-    NINJA_1_9_OR_NEWER = True
-    NINJA_CMD = ['ninja']
-else:
+# If we're on CI, detecting ninja for every subprocess unit test that we run is slow
+# Optimize this by respecting $NINJA and skipping detection, then exporting it on
+# first run.
+try:
+    NINJA_1_9_OR_NEWER = bool(int(os.environ['NINJA_1_9_OR_NEWER']))
+    NINJA_CMD = [os.environ['NINJA']]
```
jpakkane, line 54: Shouldn't these be the other way around? If the first is not set then it throws an exception and the `NINJA` envvar is never read.

eli-schwartz: That's fine, if we throw an exception then we ignore NINJA_CMD anyway so the assignment is meaningless. In the except block we read the environment variable again, using `detect_ninja()`.

jpakkane:
LGTM, if someone from Alpine can confirm that this fixes things for them, feel free to merge.

psycholance:
yes- these fixes fix the majority of the test issues on our CI, for the testsuites that run (some more niche languages and roughly ~20% of the main unittests are skipped). the remainder of minor fixes are just 4 individual test disables (see https://gitlab.alpinelinux.org/alpine/aports/-/blob/7648ff3d04795c8fc205d1fd6554dec2cfca0570/main/meson/skip-broken-tests.patch), and the only remaining thing is that the cmake (13 linuxlike cmake dependency) test fails, but that is mostly separate to this

eli-schwartz:
Alright, merged. \o/

Marking this for potential backporting since some or all of this should be suitable for a .3 point release, given it largely just affects running the tests, and it would be nice to allow reducing the patch delta in alpine sooner rather than later.

PR #10011, opened 2022-02-17
flake8 fixes

eli-schwartz:
Down from 242 warnings to 189.

There is one behavioral change I discovered while investigating a bizarre textwrap.dedent alignment issue: df1642f15770a40f27e6635f4d212317cbd411fc
> fix malformed warning to print the way it was meant to print

Other than that it's mostly just whitespace juggling.

### DIFF
--- a/mesonbuild/compilers/detect.py
+++ b/mesonbuild/compilers/detect.py
@@ -613,7 +613,7 @@ def _detect_c_or_cpp_compiler(env: 'Environment', lang: str, for_machine: Machin
                 ccache + compiler, version, for_machine, is_cross, info,
                 exe_wrap, full_version=full_version, linker=l)
         if 'TMS320C2000 C/C++' in out or 'MSP430 C/C++' in out or 'TI ARM C/C++ Compiler' in out:
-            lnk : T.Union[T.Type[C2000DynamicLinker], T.Type[TIDynamicLinker]]
+            lnk: T.Union[T.Type[C2000DynamicLinker], T.Type[TIDynamicLinker]]
             if 'TMS320C2000 C/C++' in out:
                 cls = C2000CCompiler if lang == 'c' else C2000CPPCompiler
                 lnk = C2000DynamicLinker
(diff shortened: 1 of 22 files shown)


mesonbuild/mparser.py
```diff
@@ -194,14 +193,11 @@ def lex(self, filename: str) -> T.Generator[Token, None, None]:
...
-                                """),
-                                self.getline(line_start),
-                                str(lineno),
-                                str(col)
-                            )
+                            msg = ParseException("Newline character in a string detected, use ''' (three single quotes) "
+                                                 "for multiline strings instead.\n"
+                                                 "This will become a hard error in a future Meson release.",
+                                                 self.getline(line_start), lineno, col)
+                            mlog.warning(msg, location=BaseNode(lineno, col, filename))
```
dcbaker, line 200: I find the triple quoted, demented version much easier to read, since it looks in the editor exactly his it will be printed

dcbaker (changes requested):
A couple of small nits, but otherwise this looks good to me. Only the return type dedent is a hard ask, the other is a soft ask

mesonbuild/compilers/detect.py
```diff
@@ -256,7 +256,7 @@ def _handle_exceptions(
         exceptions: T.Mapping[str, T.Union[Exception, str]],
         binaries: T.List[T.List[str]],
         bintype: str = 'compiler'
-    ) -> T.NoReturn:
+) -> T.NoReturn:
```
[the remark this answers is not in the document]
eli-schwartz: I hate it too, but apparently that is "the thing" or something...

The thing is, I don't actually know what would look good here. I'm very fond of single-line function signatures, but sadly that isn't always possible. :(

mesonbuild/mparser.py
eli-schwartz: And yet it results in flake8 indentation warnings, junk whitespace at the end, and accomplishes nothing because it isn't being printed any which way in particular, the exact formatted layout is not meaningful.

I'm basically pretty meh on that.

dcbaker (approved):

eli-schwartz:
While I'm at it, one of the problems with textwrap.dedent here was just misuse... the same misuse that is present in a couple other cases (not all) in the codebase. I fixed those too.

mesonbuild/mparser.py
eli-schwartz: (To be clear: flake8 complains because it cannot figure out a good way to indent the line immediately after, because apparently `func1(func2('string'), otherarg)` cannot be happy when string is multiline. Go figure.)

dcbaker (approved):

PR #10012, opened 2022-02-17
Add all supported arguments in docs for summary()

annacrombie:
The docs for summary don't include all valid keywords.

### DIFF
--- a/docs/yaml/functions/summary.yaml
+++ b/docs/yaml/functions/summary.yaml
@@ -10,7 +10,7 @@ description: |
   the section keyword argument is omitted, those key/value pairs are
   implicitly grouped into a section with no title. key/value pairs can
   optionally be grouped into a dictionary, but keep in mind that
-  dictionaries does not guarantee ordering. `key` must be string,
+  dictionaries do not guarantee ordering. `key` must be string,
   `value` can be:
 
   - an integer, boolean or string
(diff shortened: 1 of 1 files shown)


docs/yaml/functions/summary.yaml
```diff
@@ -62,10 +62,30 @@ example: |
...
+
+kwargs:
+  bool_yn:
+    type: bool
+    default: false
+    description: Convert bool values to yes and no
+  section:
+    type: str
+    default: "''"
+    description: The section to put this summary information under.
```
eli-schwartz, line 85: This feels a bit awkward to describe. Maybe it should t declare the default as `''` but say that if unspecified, this summary information goes into an unnamed leading section?

docs/yaml/functions/summary.yaml
```diff
@@ -62,10 +62,30 @@ example: |
...
 
 posargs:
-  key:
-    type: str
-    description: The name of the new entry
+  key_or_dict:
+    type: str | dict
+    description: |
+      The name of the new entry, or a dictionary with values of the same type
+      as the `value` optarg.
```
eli-schwartz, line 69: This is confusing and makes it sound like the value optarg is actually passed in this case. Also it's a bit indirect.

All dicts need a string key and the reference manual usually specifies the allowed values alone. So this could be `str | dict[str | bool | int | dep | external_program | list[str | bool | int | dep | external_program]]`

PR #3830, opened 2018-07-03
java: install jar files to share/java by default

noverby:

PR #10013, opened 2022-02-17
Change jar() default install dir

tristan957:
The previous install dir seemed incorrect when looking at various Linux distributions.

Closes #3830

### DIFF
--- a/docs/markdown/snippets/jar_default_install_dir.md
+++ b/docs/markdown/snippets/jar_default_install_dir.md
@@ -1,0 +1,5 @@
+## JAR default install dir
+
+The previous default for `jar()` was `libdir`. With this release, it has been
+changed to `datadir/java`. Please open an issue if this is not a sane default
+for your system.
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -2627,6 +2627,9 @@ class Jar(BuildTarget):
             return ['-cp', os.pathsep.join(cp_paths)]
         return []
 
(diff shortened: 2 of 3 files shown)


eli-schwartz:
Apparently none of our current tests in `test cases/java/` perform installation of jars?

tristan957:
That, or they don't check to make sure something got installed correctly. Is there any prior art that I can copy?

eli-schwartz:
It should suffice to:
- add `install: true` where relevant
- create test.json with the `installed` list as documented here: https://mesonbuild.com/Contributing.html#installed

jpakkane:
The only thing to note is that we will probably need to add an option for this at some point (like `libdir`). Though since this is rare, it should probably go in the Java namespace and only be enabled if Java is enabled. So something like `java_jardir`. Though we might also need it for Kotlin projects if we ever enable them.

But probably we can do this first and only add that if and when people actually need it.

tristan957:
I am not super familiar with Kotlin, but I think `jardir` would be sufficient. JARs can be used across JVM languages. I would be happy to add that option in this PR @jpakkane.

eli-schwartz:
Turns out someone opened a similar PR in the past: #3830 

That one no longer applies to git master, and included a top level build option that was apparently not wanted, hence its state of limbo.

tristan957:
Sorry I haven't come back to add testing. Plan is to get this in for 0.62

jpakkane:
Test failure is unrelated. Let's merge this now and we can add an option later if it is truly needed.

Issue #7560, opened 2020-08-06
Dlang dub dependency cannot be found


vladimmi:
**Describe the bug**
Meson is unable to find dependencies already fetched and built by dub.

**To Reproduce**
`meson.build`
```meson
project('example', 'd')
dparse_dep = dependency('libdparse', method: 'dub')
executable('bin', ['app.d'], dependencies: [dparse_dep])
```
`app.d`
```d
import dparse;
void main() {}
```
Build process:
```
> dub fetch libdparse
> dub build libdparse --compiler=ldc2 --build=release
> dub list
Packages present in the system and known to dub:
  libdparse 0.15.4: C:\Users\vladi\AppData\Local\dub\packages\libdparse-0.15.4\libdparse\
  stdx-allocator 2.77.5: C:\Users\vladi\AppData\Local\dub\packages\stdx-allocator-2.77.5\stdx-allocator\
```
Actual build dir is `C:\Users\vladi\AppData\Local\dub\packages\libdparse-0.15.4\libdparse\.dub\build\library-release-windows-x86_64-ldc_2092-294ACC73B9429440621F081FE84A254D`
```
> meson builddir --buildtype=release
The Meson build system
Version: 0.55.0
Source dir: D:\Workspace\d\example
Build dir: D:\Workspace\d\example\builddir
Build type: native build
Project name: example
Project version: undefined
D compiler for the host machine: ldc2 (llvm 1.22.0 "LDC - the LLVM D compiler (1.22.0):")
D linker for the host machine: ldc2 link 14.26.28806.0
Host machine cpu family: x86_64
Host machine cpu: x86_64
Found DUB: D:\Portable\dub.EXE (DUB version 1.22.0, built on Jul  9 2020)
ERROR: libdparse found but it wasn't compiled with ldc
Run-time dependency libdparse found: NO

meson.build:2:0: ERROR: Dependency "libdparse" not found
```

**Expected behavior**
Project configured and dependency found.

**system parameters**
* OS: Windows 10
* Python 3.8.3
* Meson 0.55.0

vladimmi:
I've digged a bit Meson source code and found some possible issues (though I can miss something):

1) https://github.com/mesonbuild/meson/blob/e5fabce1f32ffe918a96acd026c0e258f6d682b8/mesonbuild/dependencies/base.py#L1727-L1739
code here attempts to determine host **D compiler** version but for some reason calls `dub --version`?..

[...]

PR #10016, opened 2022-02-18
Log that DUB dependency is not found because of missing lib

rtbo:
Dub is a bit messy at times and won't necessarily build all dependencies of a package.
So when one runs `dub build vibe-d:http`, DUB will build ONLY this package, and not all the packages it depends on.
(I think this is possible because of compiling static libraries).

This situation can create confusion, which I think is the main reason of #7560.

This problem is not Meson's to solve, but a hint would help users of where to turn to.

### DIFF
--- a/mesonbuild/dependencies/dub.py
+++ b/mesonbuild/dependencies/dub.py
@@ -148,6 +148,10 @@ class DubDependency(ExternalDependency):
                     if lib_path:
                         self.link_args.append(lib_path)
                     else:
+                        if os.path.exists(file):
+                            mlog.debug("Cannot use " + file + " as a library")
+                        else:
+                            mlog.debug("Cannot find " + file)
                         self.is_found = False
 
     def _find_right_lib_path(self,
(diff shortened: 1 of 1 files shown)


mesonbuild/dependencies/dub.py
```diff
@@ -148,6 +148,10 @@ def add_lib_args(field_name: str, target: T.Dict[str, T.Dict[str, str]]) -> None
                     if lib_path:
                         self.link_args.append(lib_path)
                     else:
+                        if os.path.exists(file):
+                            mlog.debug("Cannot use " + file + " as a library")
```
tristan957, line 152: I would use an f-string

mesonbuild/dependencies/dub.py
```diff
@@ -148,6 +148,10 @@ def add_lib_args(field_name: str, target: T.Dict[str, T.Dict[str, str]]) -> None
...
+                        else:
+                            mlog.debug("Cannot find " + file)
```
dcbaker, line 154: This should be if/elif/else not a nested else I think

dcbaker (approved):

rtbo:
superseded by #10019

PR #10018, opened 2022-02-19
Add new env2cross command.

jpakkane:
Let's get this ball rolling. This is very much not complete, just the basis. This is basically a port of `debcrossgen` that is used for Debian package cross-compilation. Needs among other things:

- final API design
- implementation for envvars
- also do the same for native files somehow?
- doc fixes

Add your comments and desires below.

### DIFF
--- a/mesonbuild/scripts/env2mfile.py
+++ b/mesonbuild/scripts/env2mfile.py
@@ -1,0 +1,368 @@
+# Copyright 2022 The Meson development team
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sys, os, subprocess, shutil
+import shlex
+
+import argparse
+from typing import TextIO, Dict, List, Union, Tuple, Any, Optional
+
+from .. import mlog
+
+UNIXY_ENVVARS_COMPILER = {'c': 'CC',
+                          'cpp': 'CXX',
+                          'objc': 'OBJCC',
+                          'objcpp': 'OBJCXX',
+                          'fortran': 'FC',
+                          'rust': 'RUSTC',
+                          'vala': 'VALAC',
+                          'cs': 'CSC',
+                          }
+
+UNIXY_ENVVARS_TOOLS = {'ar': 'AR',
+                       'strip': 'STRIP',
+                       'windres': 'WINDRES',
+                       'pkgconfig': 'PKG_CONFIG',
+                       'vapigen': 'VAPIGEN',
+                       'cmake': 'CMAKE',
+                       'qmake': 'QMAKE',
+                       }
+
+UNIXY_ENVVARS_FLAGS = {'c': 'CFLAGS',
+                       'cpp': 'CXXFLAGS',
+                       'objc': 'OBJCFLAGS',
+                       'objcpp': 'OBJCXXFLAGS',
+                       'fortran': 'FFLAGS',
+                       'rust': 'RUSTFLAGS',
+                       'vala': 'VALAFLAGS',
+                       'cs': 'CSFLAGS', # This one might not be standard.
+                       }
+
+TYPICAL_UNIXY_COMPILER_NAMES = {'c': ['cc', 'gcc', 'clang'],
+                                'cpp': ['c++', 'g++', 'clang++'],
+                                'objc': ['objc', 'clang'],
+                                'objcpp': ['objcpp', 'clang++'],
+                                'fortran': ['gfortran'],
+                                }
+
+LANGS_USING_CPPFLAGS = {'c', 'cpp', 'objc', 'objcxx'}
+
+def has_for_build() -> bool:
+    for cenv in UNIXY_ENVVARS_COMPILER.values():
+        if os.environ.get(cenv + '_FOR_BUILD'):
+            return True
+    return False
+
+def add_arguments(parser: 'argparse.ArgumentParser') -> None:
+    parser.add_argument('--debarch', default=None,
+                        help='The dpkg architecture to generate.')
+    parser.add_argument('--gccsuffix', default="",
+                        help='A particular gcc version suffix if necessary.')
+    parser.add_argument('-o', required=True, dest='outfile',
+                        help='The output file.')
+    parser.add_argument('--cross', default=False, action='store_true',
+                        help='Generate a cross compilation file.')
+    parser.add_argument('--native', default=False, action='store_true',
+                        help='Generate a native compilation file.')
+    parser.add_argument('--system', default=None,
+                        help='Define system for cross compilation.')
+    parser.add_argument('--cpu', default=None,
+                        help='Define cpu for cross compilation.')
+    parser.add_argument('--cpu-family', default=None,
+                        help='Define cpu family for cross compilation.')
+    parser.add_argument('--endian', default='little', choices=['big', 'little'],
+                        help='Define endianness for cross compilation.')
+
+class MachineInfo:
+    def __init__(self) -> None:
+        self.compilers: Dict[str, List[str]] = {}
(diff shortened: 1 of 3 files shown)


Ericson2314:
Moving over comments to here, as requested by @jpakkane 

CC @dcbaker 

##  This should not be specific to cross files in the slightest.

Cross and native files have the same structure, and the command should work equally well for both of them.

The name of the command therefore probably shouldn't have "cross" in it, as that would mask the fact that it works equally well for native files.

##  We shouldn't duplicate the env var logic

Originally, the treatment of env vars in Meson was completely ad-hoc. But through @dcbaker's and my refactors, it is now centralized: see how few places `os.environ[` occurs now and how the names of environment env vars are looked up from a map.

All of that logic should be reused as much as possible, so we don't have two separate implementations of populating configs from env vars.

Ideally, we should go env further, so that *regardless* of the configuration, we turn env vars into config the same way. We don't quite do that today because we still read some env vars on demand based on the choice of compiler, even though the routing of env vars to config fields is static. Put another way, we *lazilly load env vars on demand*, but we probably should eagerly load  them.

## Future work

Due to the incentivisation principles layed out in https://nibblestew.blogspot.com/2020/03/its-not-what-programming-languages-do.html, cross compilation users depend on cross being as "analogous / not a special case" as possible. Otherwise native users were just do whatever works for them, and it will unintentionally break stuff for cross users, and if we try to fix it and tell them they keep on doing the wrong thing, they will cast us out as a burden.

Concretely, to make NixOS be a regular (non-cross-specific distro like Yocto) that nonetheless supports cross compilation everywhere, I have to make it so that by default "if it works on native, it only works on cross", so that native-only users *unintentionally* do the right thing for everyone.

That means if env vars work in native, they must also work for cross, and contrapositively, in env vars don't work for cross, they must also not work for native.

Therefore, if we wish to remove cross support for env vars in Meson, and don't wish to through us cross users under the bus *socially* (as opposed to technically), we must also remove native support for env vars in from Meson.

That means, `meson env-to-machine-file` should be the *only* way to use env vars to parameterize the build. No `os.environ` in the rest of Meson. No exceptions.

## Between then and now

I am not sure whether we want to go all in and make `meson env-to-machine-file` the only way to ingest env vars. And as long as cross and native get the same treatment, I don't really care.

But further centralizing the access of `os.environment`, and making the ingesting of env vars eager (and solely in `environment,py`) rather than lazy (and perhaps also in `compilers/`) is an extremely virtuous refactor that we should do no matter what.

If nothing else, our conversations here and in #9969 seem to indicate that @jpakkane was not aware of the full ramifications of @dcbaker's and my refactor cleanups. I therefore recommend @jpakkane spearhead any further cleanups in service of deduplication and the static (configuration agnostic) handling of env vars so going forward we are all on the same page.

jpakkane:
This currently does only cross because you can't really do both native and cross with the same command. If you are looking up native compilers you need to use, for example, `CC_FOR_BUILD` but when you look up cross compilers you mustn't use it. Maybe there should also be a `env2native` command or maybe this command should have a command line argument specifying which one it should create. Comments welcome.

Ericson2314:
I was thinking `env-to-machine-file [--native] [--cross]` but the important thing is not the exact CLI, but that the implementation is capable of doing both. If we dedup with `environment.py`, we have all the `for_machine: MachineChoice, is_cross: Bool` logic at our disposal to make doing both ways easy.

xclaesse:
FWIW, Cerbero has similar code too: https://gitlab.freedesktop.org/gstreamer/cerbero/-/blob/main/cerbero/build/build.py#L833. I guess you're doing this for the same reason: a meta build system needs to setup everything in environment for other build tools (e.g. autotools) and when a package is using meson they need to write the cross file for it.

Maybe not realistic on the short term, but I was wondering if we could do this backward: have a tool that setup env from a cross file. So what debian/cerbero would write is a cross file first, and autotools/etc packages would use that tool to setup the env. Maybe we could achieve something if the cross file format could be more than just Meson. Anyway... just dreaming out loud :P

jpakkane:
> Maybe we could achieve something if the cross file format could be more than just Meson.

It's designed the way it is (it has no Meson internals) just so other build systems could use it.

Ericson2314:
The cross file format or the implementation of this new command is designed that way?

jpakkane:
Updated to do envvar based detection as well as clarify in the docs that this is still experimental for now.

Detecting native files is still missing, doing that next.

mesonbuild/scripts/env2mfile.py
```diff
@@ -0,0 +1,281 @@
...
+                       'cpp': 'CXXFLAGS',
+                       'objc': 'OBJCFLAGS',
+                       'objcpp': 'OBJCXXFLAGS',
+                       'fortran': 'FFLAGS'}
+
+LANGS_USING_CPPFLAGS = {'c', 'cpp', 'objc', 'objcxx'}
+
+def add_arguments(parser: 'argparse.ArgumentParser') -> None:
+    parser.add_argument('--debarch', default=None,
+                        help='The dpkg architecture to generate.')
```
Ericson2314, line 39: It would be good to separate out the debian-specific env vars, in case non-debian happens to use them differently.

jpakkane:
> The cross file format or the implementation of this new command is designed that way?

The file format.

> I see much functionality that is duplicated from the rest of Meson and I wonder whether that is intentional or not.

The current envvar handling is spread out and hard to make out. I specifically wanted to have everything concentrated in this one file. Eventually we might even run this transparently upon first initialisation to create an internal cross/native file if one is not specified and then use that. 

> Also, don't forget _FOR_BUILD

I'll add that to the native file generation bit once I get around to implementing it.

jpakkane:
> It would be good to separate out the debian-specific env vars

There should be none. Debian package building uses CC et al just the same as all others. `--debarch` only uses the `dpkg-architecture` command to detect the compilers and platform details, compiler flags (should) come from the same envvars.

Ericson2314:
> Eventually we might even run this transparently upon first initialisation to create an internal cross/native file if one is not specified and then use that.

I would highly support that.

> The current envvar handling is spread out and hard to make out.

It is a *lot* less bad than it used to be. Most of it is in `environment.py`.

> I specifically wanted to have everything concentrated in this one file. Eventually we might even run this transparently upon first initialisation to create an internal cross/native file if one is not specified and then use that.

I guess what I am saying is that I am very uncomfortable landing two sets of env var implementations. Until we deduplicate as you say (which sounds great!) that is a huge recipe for bugs, almost like making it super spread out again as it was before.

How about instead of making this depend on the other stuff, you make the other stuff depend on this? (or both depend on a third module, so this just does the main / arg pars parts.) That would make me feel a lot more confortable, and is total in line with the goal of consolidation?

mesonbuild/scripts/env2mfile.py
```diff
@@ -0,0 +1,281 @@
...
+            continue
+        k, v = line.split('=', 1)
+        data[k] = v
+    host_arch = data['DEB_HOST_GNU_TYPE']
+    host_os = data['DEB_HOST_ARCH_OS']
+    host_cpu_family = cpu_family_map.get(data['DEB_HOST_GNU_CPU'],
+                                         data['DEB_HOST_GNU_CPU'])
+    host_cpu = cpu_map.get(data['DEB_HOST_ARCH'],
+                           data['DEB_HOST_ARCH'])
+    host_endian = data['DEB_HOST_ARCH_ENDIAN']
```
Ericson2314, line 167: fair enough, I misread those as env vars at first.

Ericson2314:
https://github.com/mesonbuild/meson/blob/5e4ea9f0f444120e566e364aba0b12f2520bd9ac/mesonbuild/envconfig.py#L92-L134 more env vars in a nice map like the ones you use here.

jpakkane:
As this MR:

- is explicitly marked as experimental
- needs to be put out to get actual usage feedback
- is fully additive, no change in functionality
- does most of the things raised in review
- the test failures are unrelated

I'm going to merge this. This also allows people to contribute to this by submitting actual patches rather than via comments.

eli-schwartz:
> the test failures are unrelated

Before this PR, the documentation build passed just fine. In the CI for this PR, the documentation build and unittests both have an infinite hang and were canceled, and the merge to git master infinitely hangs too so most of those were canceled and I just canceled the documentation build too.

It reproducibly bisects to this commit locally as well, as soon as you try to run:
- `tools/regenerate_docs.py` in the docs build
- unittests: `test_commands_documented`

I believe it was premature to merge this PR with failing CI and I believe it was entirely wrong to take a pathologically freezing build and say that the failures are "unrelated".

Issue #6581, opened 2020-02-04
Using dub for dependencies doesn't work if they're of type "sourceLibrary"


Cogitri:
**Describe the bug**
When specifying a dub dependency on a package that is of the `targetType` `sourceLibrary` meson can't find it via `dependency('openssl', method: 'dub')` and throws the following error: `ERROR: openssl found but it wasn't compiled with $DC`. Since nothing is compiled for packages of type `sourceLibrary` (they're just bindings after all), there's no way that the compilers don't match for this.

**To Reproduce**
Any D project that depends on a `sourceLibrary` package (e.g. openssl) will do.

**Expected behavior**
Meson should recognise `sourceLibrary` packages and don't check if the compiler for that was the same as is currently being used for compilation.

**system parameters**
I've used Meson 0.53.1

PR #10019, opened 2022-02-19
rewrite Dub dependencies

rtbo:
This is nearly a complete rewrite of Dub dependencies that were broken except for the most simple cases.

Biggest change on user standpoint: Meson will actually fetch and build libraries that are not found for the correct compiler.

Other fixes:
 - fix the research of target built by DUB
 - explicitely state that DUB dynamic libraries and source libraries are not supported (yet) (#6581)
 - fix the build settings of recipes having sub-dependencies (#7560)
 - sanitization, comments, explanations...

My test case is a simple executable depending on `vibe-d:http`. Vibe-d is complex and has a many dependencies.
I can easily add it in automated tests ~~, but I'm not sure that it can be skipped on CI that do not have DUB~~.
I tested with success on Linux and Windows (to figure out the right order of static libraries took some time but is now OK).

The non-compatibility with Dub package having sub-dependencies is especially annoying and is the main motivation for me to write this PR.
My meson file before:
```meson
vibed_deps = [
    dependency('stdx-allocator',
        version: '2.77.5',
        method: 'dub',
    ),
    dependency('eventcore',
        version: '0.9.20',
        method: 'dub',
    ),
    dependency('vibe-core',
        version: '1.22.0',
        method: 'dub',
    ),
    dependency('diet-ng',
        version: '1.8.0',
        method: 'dub',
    ),
    dependency('taggedalgebraic',
        version: '0.11.22',
        method: 'dub',
    ),
    dependency('vibe-d:utils',
        version: '0.9.4',
        method: 'dub',
    ),
    dependency('vibe-d:crypto',
        version: '0.9.4',
        method: 'dub',
    ),
    dependency('vibe-d:stream',
        version: '0.9.4',
        method: 'dub',
    ),
    dependency('vibe-d:textfilter',
        version: '0.9.4',
        method: 'dub',
    ),
    dependency('vibe-d:data',
        version: '0.9.4',
        method: 'dub',
    ),
    dependency('vibe-d:tls',
        version: '0.9.4',
        method: 'dub',
    ),
    dependency('vibe-d:inet',
        version: '0.9.4',
        method: 'dub',
    ),
    dependency('vibe-d:http',
        version: '0.9.4',
        method: 'dub',
    ),
]
if host_machine.system() == 'windows'
    vibed_deps += declare_dependency(
        link_args: [
            join_paths(meson.source_root(), 'libs/windows-x64/libcrypto.lib'),
            join_paths(meson.source_root(), 'libs/windows-x64/libssl.lib'),
            'user32.lib'
        ],
    )
endif
```
My meson file now:
```meson
vibed_deps = [
    dependency('vibe-d:http',
        version: '0.9.4',
        method: 'dub',
    ),
]
if host_machine.system() == 'windows'
    vibed_deps += declare_dependency(
        link_args: [
            'user32.lib'
        ],
    )
endif
```
I still have to explicitely add `user32.lib`. I don't think it has anything to do with DUB though.


rtbo:
I added my test case, as well as a test case that currently fails.
The one that fails is due to the library order on the link command line and the fact that these are apparently de-duped.
```meson
vibed_deps = [
  dependency('vibe-core', method: 'dub'),
  dependency('vibe-d:http', method: 'dub'),
]
```
`vibe-d:http` depends on `vibe-core`, but the latter comes first on the link command and the linker don't find some symbols in `vibe-core`.
If `vibe-core` is placed after `vibe-d:http`, then the build works fine.
I don't know if this is something that must be fixed. This dependency pattern didn't work before anyway.
Surprisingly, the one with the complete dependencies list of my first post is working although it is not listed in the right order.

rtbo:
> Surprisingly, the one with the complete dependencies list of my first post is working although it is not listed in the right order.

In fact not, it only works on Windows, so does the version with two dependencies.
The reason is that the linker is probably more flexible on Windows and can find symbols regardless on the command line order.

So I don't think that I am breaking anything in fact.

rtbo:
BTW, on Windows I have the following warnings from the linker:
```
LINK : warning LNK4044: unrecognized option '/lws2_32'; ignored
LINK : warning LNK4044: unrecognized option '/luser32'; ignored
LINK : warning LNK4044: unrecognized option '/ladvapi32'; ignored
```
which is the reason of linker not finding `user32.lib`. It can be specific to D flags translation. I can work on this on a future PR.

mesonbuild/coredata.py
```diff
@@ -1216,6 +1216,7 @@ def add_to_argparse(self, name: str, parser: argparse.ArgumentParser, help_suffi
     (OptionKey('werror'),          BuiltinOption(UserBooleanOption, 'Treat warnings as errors', False, yielding=False)),
     (OptionKey('wrap_mode'),       BuiltinOption(UserComboOption, 'Wrap mode', 'default', choices=['default', 'nofallback', 'nodownload', 'forcefallback', 'nopromote'])),
     (OptionKey('force_fallback_for'), BuiltinOption(UserArrayOption, 'Force fallback for those subprojects', [])),
+    (OptionKey('dub_fetch_build'), BuiltinOption(UserBooleanOption, 'Whether to fetch and build the missing DUB dependencies', True)),
```
tristan957, line 1219: Maybe there should be a dub module? Would hate to put dub_ in the top-level options

tristan957: More of a complaint really.

eli-schwartz: There is a `dlang` module.

rtbo: There is already a dlang module that generate dub.json files. But the dub dependencies do not depend on it.
You mean that the module would provide a function to download and fetch missing libraries?

tristan957: I am not super familiar with how module options work, but I would think you could read that option in the regular dependency() call like you are doing right now. @eli-schwartz knows better about module-specific options.

eli-schwartz: module-specific options are only shown in configure, if that module has been `import()`ed. But otherwise they exist just fine and can be in theory used for any purpose, though one might debate whether it "should" be used for code which is not resident inside said module.

tristan957: Probably fine as-is

dcbaker:
This causes dub to build dependencies durring configure time, right?

rtbo:
> This causes dub to build dependencies durring configure time, right?

Yes correct. This is the only relevant thing to do IMO (otherwise than not to do it all).

rtbo:
The problem about `user32.lib` is because DMD receives `-L=-luser32` and passes it directly to `link.exe`.
I can fix this easily here with the following commit. 
I think the logic belongs more to a general function that does unix to windows argument translation. 
Do you want me to include the fix now, which can be reverted later when fixed in the appropriate place?

```diff
diff --git a/mesonbuild/dependencies/dub.py b/mesonbuild/dependencies/dub.py
index b78f5c3cb..d2aae7b32 100644
--- a/mesonbuild/dependencies/dub.py
+++ b/mesonbuild/dependencies/dub.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from mesonbuild.mesonlib.universal import MachineChoice
 from .base import ExternalDependency, DependencyException, DependencyTypeName
 from .pkgconfig import PkgConfigDependency
 from ..mesonlib import (Popen_safe, OptionKey)
@@ -240,6 +241,12 @@ class DubDependency(ExternalDependency):
         for flag in bs['lflags']:
             self.link_args.append(flag)
 
+        is_windows = self.env.machines[MachineChoice.HOST].is_windows()
+        if is_windows:
+            winlibs = ['kernel32', 'user32', 'gdi32', 'winspool', 'shell32', 'ole32',
+                'oleaut32', 'uuid', 'comdlg32', 'advapi32', 'ws2_32']
+
+
         for lib in bs['libs']:
             if os.name != 'nt':
                 # trying to add system libraries by pkg-config
@@ -252,6 +259,11 @@ class DubDependency(ExternalDependency):
                     for arg in pkgdep.get_link_args(raw=True):
                         self.raw_link_args.append(arg)
                     continue
+
+            if is_windows and lib in winlibs:
+                self.link_args.append(lib + '.lib')
+                continue
+
             # fallback
             self.link_args.append('-l'+lib)
```

dcbaker:
> Yes correct. This is the only relevant thing to do IMO (otherwise than not to do it at all).

We should not be building at compile time, only at build time. Otherwise, we make configure take even longer. Besides that, one option we have in the future would be to actually do the building ourselves, since what `dub describe` returns would be sufficient for us to generate meson AST and normal build rules, like we do with CMake, and I've done with Cargo.

rtbo:
> Besides that, one option we have in the future would be to actually do the building ourselves

OK I see. This is a lot more complex. I think it will not be possible for all Dub packages. Many have pre-build steps commands that rely on environment variables provided by Dub and will require substancial maintenance effort to support this by Meson. 

If you don't like driving Dub at configure time, I will remove that part of the code. The most important is to get the correct build settings. 
Where driving Dub build is interesting though is that Dub do not build the sub-dependencies. So the user currently has to run `dub build` for every dependency. `vibe-d` has a dozen of them.

rtbo:
The build of DUB dependencies is removed.
Instead I write in the log suggestions of commands to run. For example:
```txt
ERROR: No suitable target found for several DUB packages
You may run the following commands to build them:
dub build vibe-d:http@0.9.4 --compiler=ldc2 --build=debug --arch=x86_64 --config=library
dub build diet-ng@1.8.0 --compiler=ldc2 --build=debug --arch=x86_64 --config=library
dub build vibe-d:crypto@0.9.4 --compiler=ldc2 --build=debug --arch=x86_64 --config=library
dub build mir-linux-kernel@1.0.1 --compiler=ldc2 --build=debug --arch=x86_64 --config=library
dub build vibe-d:inet@0.9.4 --compiler=ldc2 --build=debug --arch=x86_64 --config=library
dub build vibe-d:textfilter@0.9.4 --compiler=ldc2 --build=debug --arch=x86_64 --config=library
dub build vibe-d:textfilter@0.9.4 --compiler=ldc2 --build=debug --arch=x86_64 --config=library
dub build vibe-d:tls@0.9.4 --compiler=ldc2 --build=debug --arch=x86_64 --config=openssl
dub build vibe-d:stream@0.9.4 --compiler=ldc2 --build=debug --arch=x86_64 --config=library
dub build vibe-core@1.22.0 --compiler=ldc2 --build=debug --arch=x86_64 --config=epoll
dub build eventcore@0.9.20 --compiler=ldc2 --build=debug --arch=x86_64 --config=epoll
dub build taggedalgebraic@0.11.22 --compiler=ldc2 --build=debug --arch=x86_64 --config=library
```
The user can copy and paste the commands directly to the terminal.
In the end most users will want this, so running the commands directly by Meson would not be a loss of configure time.

tristan957:
@rtbo I think what @dcbaker was saying is that you should add something like these commands to the build.ninja file. I am not familiar with D at all, so I don't know what the output of `dub describe` is.

rtbo:
> I think what @dcbaker was saying is that you should add something like these commands to the build.ninja file.

Not really an option. Dub will always write in its own directory structure, and we don't even know exactly where in advance. It is not possible to tell it to write into the meson directory. 
The proposal was more to write build rules to perform the build instead of Dub (directly invoke the D compiler to build the Dub packages into the meson dir). `dub describe` provide all the needed data to do it. 

@dcbaker  what exactly do you have in mind? I'm thinking of an additional function in the `dlang` module such as `target = dlang.dub_library(...)`

For this PR though I prefer to fix what is broken rather than introduce such advanced new feature.

rtbo:
Anything else I can do to promote this PR? Is the feature set acceptable as is?
The test failures seem irrelevant (codecov down?)

dcbaker:
Sorry I've not been attentive to this.

I'm a little concerned with configure and build having side effects outside of our build directory. 

> Not really an option. Dub will always write in its own directory structure, and we don't even know exactly where in advance. It is not possible to tell it to write into the meson directory.

Isn't that what `dub build --root=...` is for? Even if we can't know exactly what it's going to write to, having it write into our build directory, and not modify external state would be best.

In the worst case (I'd really like this to be automatic if possible) it would be possible to us a `RunTarget` for this, where the result would be something like:
```sh
meson setup builddir
ninja -C builddir meson-dub
ninja -C builddir
```
Where the meson-dub target calls dub to build things.

I'm still uncomfortable with us thrashing with significant state outside the meson build directory though, as we promise in general we won't do that.

rtbo:
> Isn't that what dub build --root=... is for? 

`dub --root` is similar to the `ninja -C` switch. So it would be possible to have dub build into the meson dir only if the dub source package is located into the meson dir. 
I don't think having dub build into the meson dir is feasible at all without modifying dub.

> In the worst case it would be possible to us a RunTarget for this

Hmm, it looks difficult to put all the logic of fetching/building subdependencies in a single target. Or this target has to be a python script executed by ninja. 

> I'd really like this to be automatic if possible

For this, the targets depending on the dub dependency must also depend on the RunTarget, but won't this run the script more often than necessary?

I have an alternative proposal, which is to do the fetch and build during configuration only if the `MESON_DUB_BUILD` environment variable is set.
This has several benefits:
 - The user explicitly asks for the build and therefore expects it to happen.
 - No need for additional option like `dub_fetch_build`.
 - Can be self documented by reporting adequate message when the dependency is not found.
 - A LOT more simple to code, test and maintain. This is located only inside `dub.py`.

What I've experienced so far is that building the dozen of `vibe-d` dependencies only takes less than a minute and saves a great deal of time. (Today, meson reports `Dependency not found` and we don't even know why).
I've made the following test with a project of mine:
```txt
$ rm -rf ~/.dub && rm -rf build
$ time $MESON setup build
|...]
Found DUB: /usr/bin/dub (DUB version 1.27.0, built on Jan 22 2022)
vibe-d:data is not present locally. Attempting to fetch on Dub registry.
Building DUB package vibe-d:data@0.9.4
Building DUB package vibe-d:utils@0.9.4
Building DUB package stdx-allocator@2.77.5
Run-time dependency vibe-d:data found: YES 0.9.4
Building DUB package vibe-d:http@0.9.4
Building DUB package diet-ng@1.8.0
Building DUB package vibe-d:crypto@0.9.4
Building DUB package mir-linux-kernel@1.0.1
Building DUB package vibe-d:inet@0.9.4
Building DUB package vibe-d:textfilter@0.9.4
Building DUB package vibe-d:tls@0.9.4
Building DUB package vibe-d:stream@0.9.4
Building DUB package vibe-core@1.22.0
Building DUB package eventcore@0.9.20
Building DUB package taggedalgebraic@0.11.22
Run-time dependency vibe-d:http found: YES 0.9.4
unit-threaded:assertions is not present locally. Attempting to fetch on Dub registry.
Building DUB package unit-threaded:assertions@2.0.3
Building DUB package unit-threaded:exception@2.0.3
Building DUB package unit-threaded:from@2.0.3
Run-time dependency unit-threaded:assertions found: YES 2.0.3
[...]
real    0m42,566s
user    0m20,231s
sys     0m4,642s

$ rm -rf build
$ time $MESON setup build
[...]
Found DUB: /usr/bin/dub (DUB version 1.27.0, built on Jan 22 2022)
Run-time dependency vibe-d:data found: YES 0.9.4
Run-time dependency vibe-d:http found: YES 0.9.4
Run-time dependency unit-threaded:assertions found: YES 2.0.3
[...]
real    0m3,200s
user    0m2,494s
sys     0m0,652s
```
These times are the worst as it can get, and the user knows exactly what is going on.

tristan957:
Any proposal with an environment variable is likely to be a non-starter in Meson fyi.

rtbo:
Ok, we can think of something else then.

I'm not  sure a RunTarget is the best approach though. Injecting a target implicitely from a dependency looks more like a hack than a designed solution. (or I'm not getting the full picture)

And regarding this PR, I don't intend to bring deep modifications to Meson. My goal is:
 1. to fix what is broken (mainly the `compile_args` and `link_args` with sub-dependencies)
 2. to provide a simple boost to the user (build missing packages)

The `dub build` during configure time is really simple and robust.
If 2. can't be made that way, I will only keep 1. and come back to what I've proposed in #10016.
IMO 2. should be implemented into Dub in the first place anyway.

tristan957:
I am not dcbaker, but I think getting your PR to just fix the bug, and figuring out something else for 2 later is for the best. 0.62 is closely approaching.

eli-schwartz:
I'm not dcbaker either, but I do have a Meson hat of my own. :D

I concur with dcbaker that this should not be automatically building things outside the build directory during configure time. Either factor is on its own something Meson doesn't, as a rule, consider correct. Taken together it makes me reeeeeeeally nervous.

So I would anyways advise that yes, please retarget this PR to fix the first goal, which we can all agree is good (I haven't even taken a single glance at the implementation of this PR, sorry -- I know nothing about dlang or dub, just here to observe on the general meta nature of feature scope).

The second goal is controversial, may be better solved in dub itself, and will 100% drag things down and prevent the first part of the PR from being merged during this development cycle, which would be a huge shame. So, if you want it merged in time for a nice shiny 0.62.0 release, you are strongly advised to split this into two PRs so they can be considered separately.

...

Hmm, alternatively given the discussion in this PR has largely focused on the second goal, it may make sense to open a second PR which extracts the changes for the first goal.

rtbo:
> it may make sense to open a second PR which extracts the changes for the first goal.

Yes I'll do just that.
I have reset to 058f2eaff9d4c8c9c27aa4d309806d551309fd59 which is the version of the code that actually calls `dub build` for future references.

PR #10023, opened 2022-02-22
devenv: Backport more features from gst-env.py

xclaesse:
`meson devenv` idea comes from GStreamer project' `gst-env.py` script. They have more goodies we could make everyone benefit.

- Allow setting method/separator in environment() and meson.add_devenv()
- devenv: Do not prepend empty list to PATH and LD_LIBRARY_PATH
- devenv: Add --dump option
- devenv: Setup GDB auto-load scripts
- devenv: Source bash completion scripts
- devenv: Set PYTHONPATH where we install python modules
- Add API for modules that wants to define their devenv
- modules: Make ExtensionModule inherit from NewExtensionModule
- backends: Cache creation of install data
- Cache the result of python.find_installation()

### DIFF
--- a/docs/yaml/builtins/meson.yaml
+++ b/docs/yaml/builtins/meson.yaml
@@ -444,5 +444,25 @@ methods:
 
     posargs:
       env:
-        type: env
-        description: The [[@env]] object to add.
+        type: env | str | list[str] | dict[str] | dict[list[str]]
+        description: |
+          The [[@env]] object to add.
+          Since *0.62.0* list of strings is allowed in dictionnary values. In that
+          case values are joined using the separator.
+    kwargs:
+      separator:
+        type: str
+        since: 0.62.0
+        description: |
+          The separator to use for the initial values defined in
+          the first positional argument. If not explicitly specified, the default
+          path separator for the host operating system will be used, i.e. ';' for
+          Windows and ':' for UNIX/POSIX systems.
+      method:
+        type: str
+        since: 0.62.0
+        description: |
+          Must be one of 'set', 'prepend', or 'append'
+          (defaults to 'set'). Controls if initial values defined in the first
+          positional argument are prepended, appended or repace the current value
+          of the environment variable.
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -476,6 +476,9 @@ class EnvironmentVariables(HoldableObject):
     def has_name(self, name: str) -> bool:
         return name in self.varnames
 
+    def get_names(self) -> T.Set[str]:
+        return self.varnames
+
     def set(self, name: str, values: T.List[str], separator: str = os.pathsep) -> None:
         self.varnames.add(name)
         self.envvars.append((self._set, name, values, separator))
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -46,6 +46,7 @@ from .linkers import StaticLinker
 from .interpreterbase import FeatureNew, FeatureDeprecated
 
 if T.TYPE_CHECKING:
+    from typing_extensions import Literal
     from ._typing import ImmutableListProtocol, ImmutableSetProtocol
     from .backend.backends import Backend, ExecutableSerialisation
     from .interpreter.interpreter import Test, SourceOutputs, Interpreter
@@ -451,15 +452,19 @@ class ExtractedObjects(HoldableObject):
             for source in self.get_sources(self.srclist, self.genlist)
         ]
 
+EnvInitValueType = T.Dict[str, T.Union[str, T.List[str]]]
+
 class EnvironmentVariables(HoldableObject):
-    def __init__(self, values: T.Optional[T.Dict[str, str]] = None) -> None:
+    def __init__(self, values: T.Optional[EnvValueType] = None,
+                 init_method: Literal['set', 'prepend', 'append'] = 'set', separator: str = os.pathsep) -> None:
         self.envvars: T.List[T.Tuple[T.Callable[[T.Dict[str, str], str, T.List[str], str], str], str, T.List[str], str]] = []
         # The set of all env vars we have operations for. Only used for self.has_name()
         self.varnames: T.Set[str] = set()
 
         if values:
+            init_func = getattr(self, init_method)
             for name, value in values.items():
-                self.set(name, [value])
+                init_func(name, listify(value), separator)
 
     def __repr__(self) -> str:
         repr_str = "<{0}: {1}>"
--- a/mesonbuild/interpreter/type_checking.py
+++ b/mesonbuild/interpreter/type_checking.py
@@ -3,10 +3,13 @@
 
 """Helpers for strict type checking."""
 
+from __future__ import annotations
+import os
 import typing as T
 
 from .. import compilers
-from ..build import EnvironmentVariables, CustomTarget, BuildTarget, CustomTargetIndex, ExtractedObjects, GeneratedList, IncludeDirs
+from ..build import (EnvironmentVariables, EnvInitValueType, CustomTarget, BuildTarget,
+                     CustomTargetIndex, ExtractedObjects, GeneratedList, IncludeDirs)
 from ..coredata import UserFeatureOption
 from ..interpreterbase import TYPE_var
 from ..interpreterbase.decorators import KwargInfo, ContainerTypeInfo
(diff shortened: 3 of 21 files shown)


mesonbuild/modules/python.py
```diff
@@ -620,6 +622,50 @@ def __init__(self, interpreter: 'Interpreter') -> None:
             'find_installation': self.find_installation,
         })
 
+    def get_devenv(self, state: 'ModuleState', install_data: 'InstallData') -> T.Optional[EnvironmentVariables]:
+        python = PythonExternalProgram('python3', mesonlib.python_command)
+        if not python.found() or not python.sanity(state):
+            return None
```
eli-schwartz, line 628: This will be wrong as a general rule whenever find_installation has a real argument.

More generally it seems entirely backwards as a matter of course -- we have this information in the found installation, we know inside the found installation whether we ever invoke installation rules, and we're guaranteed to know exactly where it is installed without using *very* gnarly code to guess at it.

eli-schwartz (changes requested):
There are a couple implementation concerns I have, and a couple meta concerns about whether we should do this at all.

Particularly, this cannot be done at all, period, for mixed `install_sources()` and `extension_module()`, because PYTHONPATH is an override, not an overlay.

You can have `import foo` and `import bar` from two different directories on the PYTHONPATH. What you ***cannot*** do is have `import foo.bar` and `import foo.baz` from different directories on the PYTHONPATH (technically there are namespace packages, but people are not going to be using those, because they shouldn't be -- having `__init__.py` is sometimes necessary, and the rest of the time it is good, because it makes the interpreter faster).

This is a problem because sourcedir and builddir have different sets of files whenever `configure_file` or `custom_target` or `py_installation.extension_module()` are used, and they need to be *merged* so that python even allows loading both.

IMO the solution is going to have to be devenv copying over files into `meson-private/pythonpath/`. Alternatively, it needs to operate via wheel collection routines (so, waiting on Meson's ability to create dist-info and generate wheels, which is blocked on the ecosystem, though I do have some ideas).

mesonbuild/modules/python.py
xclaesse: I thought about doing this in `install_sources()` and `extension_module()` which would be nicer of course. However I was afraid that people would be using a lot `install_subdir(..., install_dir: python.get_install_dir())` which would be missed.

That said, I could mix both and use install_data only for the subdir, which is a much smaller list to iterate. Would that be better?

xclaesse:
> What you cannot do is have import foo.bar and import foo.baz from different directories on the PYTHONPATH

But how common would that be? I think documenting the limitation could be enough here?

tristan957:
> But how common would that be? I think documenting the limitation could be enough here?

I have a similar issue (I think) with https://github.com/hse-project/hse-python. I have a source dir called `hse2`, and I copy various files and generate the extension modules in the mirrored build dir.

Here is a run_target() where I set the PYTHONPATH, so that someone could enter a Python REPL successfully. I am sure that it only works because the current working directory in the REPL is the build root.

https://github.com/hse-project/hse-python/blob/master/meson.build#L62

VSCode gets confused with this setup however. It is very annoying because it resolves hse2 to the source root instead of the build root.

mesonbuild/modules/python.py
xclaesse: Actually, I changed my mind. I'll setup PYTHONPATH when calling `install_sources()` and `extension_module()` and if user installs python modules manually using `install_data()` or `install_subdir()` they can always setup PYTHONPATH themself as well:
```meson
env = environment()
env.prepend('PYTHONPATH', meson.current_source_dir())
meson.add_devenv(env)
```

@eli-schwartz would that be better?

xclaesse:
Hmm, why does it fail to link python moduel on Windows CIs?
```
[5/7] Linking target src/mymod2/mod2.pyd
FAILED: src/mymod2/mod2.pyd src/mymod2/mod2.pdb 
"link" @src/mymod2/mod2.pyd.rsp
LINK : fatal error LNK1104: cannot open file 'python37_d.lib'
```

xclaesse:
Nevermind, found in other unit tests that we need `default_options : ['buildtype=release']`.

mesonbuild/modules/python.py
xclaesse: Made that change now.

docs/yaml/builtins/meson.yaml
```diff
@@ -444,5 +444,23 @@ methods:
...
-        type: env
-        description: The [[@env]] object to add.
+        type: env | str | list[str] | dict[str] | dict[list[str]]
+        description: |
+          The [[@env]] object to add.
+          Since *0.62.0* list of strings is allowed in dictionnary values. In that
+          case values are joined using the separator.
+    kwargs:
+      separator:
+        type: str
```
dcbaker, line 454: pretty sure there's a `since` key

mesonbuild/build.py
```diff
@@ -451,15 +451,22 @@ def get_outputs(self, backend: 'Backend') -> T.List[str]:
...
         self.varnames: T.Set[str] = set()
 
         if values:
+            methods = {'set': self.set,
+                       'prepend': self.prepend,
+                       'append': self.append}
+            init = methods[init_method]
             for name, value in values.items():
-                self.set(name, [value])
+                init(name, listify(value), separator)
```
dcbaker, line 469:
```suggestion
                getattr(self, 'init_method')(name, listify(value), separator)
```

Then  you can get rid of the `methods` dict`

mesonbuild/build.py
```diff
@@ -451,15 +451,22 @@ def get_outputs(self, backend: 'Backend') -> T.List[str]:
             for source in self.get_sources(self.srclist, self.genlist)
         ]
 
+EnvInitValueType = T.Dict[str, T.Union[str, T.List[str]]]
+
 class EnvironmentVariables(HoldableObject):
-    def __init__(self, values: T.Optional[T.Dict[str, str]] = None) -> None:
+    def __init__(self, values: T.Optional[EnvValueType] = None,
+                 init_method: str = 'set', separator: str = os.pathsep) -> None:
```
dcbaker, line 458:
```suggestion
                 init_method: Literal['set', 'prepend', 'append'] = 'set', separator: str = os.pathsep) -> None:
```

Which either needs to be string declared, or in a `__future__.annotations` module, since it's still from `typing_extensions`. This will get mypy to catch any values that are not in the expected values.

mesonbuild/interpreter/type_checking.py
```diff
@@ -167,18 +176,25 @@ def split_equal_string(input: str) -> T.Tuple[str, str]:
     a, b = input.split('=', 1)
     return (a, b)
 
+_FullEnvInitValueType = T.Union[EnvironmentVariables, T.List[str], T.List[T.List[str]], EnvInitValueType, str, None]
 
-def _env_convertor(value: T.Union[EnvironmentVariables, T.List[str], T.List[T.List[str]], T.Dict[str, str], str, None]) -> EnvironmentVariables:
+# Split _env_convertor() and env_convertor_with_method() to make mypy happy.
+# It does not want extra arguments in KwargInfo convertor callable.
+def env_convertor_with_method(value: _FullEnvInitValueType,
+                              init_method: str = 'set',
```
dcbaker, line 184: this can probably get the same `Literal` treatement

dcbaker (changes requested):
I like the ideas, and these generally seem useful. There's a few implementation details we can tighten up, I think.

docs/yaml/builtins/meson.yaml
xclaesse: Done.

mesonbuild/build.py
xclaesse: oh nice trick using getattr(). However, I changed to only use it once before the loop.

mesonbuild/build.py
xclaesse: done.

eli-schwartz:
I would just like to reiterate that although I didn't review the whole PR, I did review the PYTHONPATH part and I do NOT think it should have been merged.

We cannot do it reliably, and doing it unreliably and confusing people with bizarre errors when a straightforward "module not found" error would tell them it's unsupported, is actively bad. We should remove the pythonpath parts and expect users to manually add that via explicit meson.build calls if and only if they know it will work correctly.

xclaesse:
@eli-schwartz Sorry, did not understand you were not happy with that, I thought your comment was just about doing it install_sources() and extension_module() instead of checking installed files afterward, which I agreed and changed.

PR #10024, opened 2022-02-23
Dependency lookups with custom functionality can now trigger FeatureNew

eli-schwartz:
Typically needed for stuff with completely custom logic due to being:
- method `builtin` on some systems
- a base system library on other systems or simply not providing pkg-config

Plus jdk which is, um, Java. /cc @tristan957 and cross-reference #9879

(Yes, the `self.featurechecks` list is not currently used by my PR. It *will* be used by jdk -> jni, because I have a fast path for new dependencies which I think is potentially quite common, but I don't have a fast path for deprecated dependencies which I don't foresee being at all common.)

### DIFF
--- a/mesonbuild/dependencies/base.py
+++ b/mesonbuild/dependencies/base.py
@@ -30,6 +30,7 @@ if T.TYPE_CHECKING:
     from .._typing import ImmutableListProtocol
     from ..compilers.compilers import Compiler
     from ..environment import Environment
+    from ..interpreterbase import FeatureCheckBase
     from ..build import BuildTarget, CustomTarget, IncludeDirs
     from ..mesonlib import FileOrString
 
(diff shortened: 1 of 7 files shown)


tristan957:
Maybe fix these FIXMEs while you are here: https://github.com/mesonbuild/meson/commit/39f61795ca1ae7de3c4c1fc08f88b3d864502cec

dcbaker:
I like the idea, I agree with @tristan957, if possible can we take care of the FIXME's?

eli-schwartz:
I intend to fix that too, yes. Does it have to be as part of this PR? It is not strictly related, and I realized while implementing this that I don't think there is a common thread allowing me to reuse any of this work.

So that would be a standalone improvement that can be done in a later PR.

Meanwhile this PR is a soft blocker for the jdk work in another PR.

dcbaker (approved):

PR #10026, opened 2022-02-23
Add missing install_tag kwarg to install_emptydir()

cwe-skov-dk:


mesonbuild/interpreter/interpreter.py
```diff
@@ -2036,10 +2036,11 @@ def func_install_man(self, node: mparser.BaseNode,
     @FeatureNew('install_emptydir', '0.60.0')
     @typed_kwargs(
         'install_emptydir',
-        INSTALL_MODE_KW
+        INSTALL_MODE_KW,
+        KwargInfo('install_tag', (str, NoneType))
```
eli-schwartz, line 2040: Since this only got added in a later version, it will need to be marked `since` the version we are currently developing for.

eli-schwartz (changes requested):

xclaesse:
Could be added into the unit test too: `test cases/unit/99 install all targets/meson.build` and `unittests/allplatformstests.py` method `test_install_tag`.

mesonbuild/interpreter/interpreter.py
cwe-skov-dk: Which is the correct version to use?

eli-schwartz:
I was actually gonna merge this last night but I had to approve the CI again. :)

PR #10028, opened 2022-02-24
Map gnu++ to vc++ for Visual Studio

### LINKED ISSUE #10015: msvc chokes on `cpp_std=gnu++14` setting in meson.build
**Describe the bug**
I have a cross-platform C++ code base that needs language extensions to compile correctly. Unfortunately, there does not appear to be a way to say "C++ with extensions"; meson only supports "standard C++", "gnu C++", and "vc C++".

Right now, I either need to set the default version to gnu++14 and manually specify vc++14 on the command line when building on Windows, or set the default version to vc++14 and manually specify gnu++14 on the command line when building elsewhere.

#7611 resolved this issue for C by making "gnu99" mean "c99 with local extensions" and allowing that to compile on Windows. A similar change for C++ would resolve this issue.

**To Reproduce**
Complete meson.build file:
```
project('plusplus', 'cpp', default_options:['cpp_std=gnu++14'])
```

Output:
```
C:\src\tmp>meson build
The Meson build system
Version: 0.61.2
Source dir: C:\src\tmp
Build dir: C:\src\tmp\build
Build type: native build
Project name: plusplus
Project version: undefined

meson.build:1:0: ERROR: Value "gnu++14" (of type "string") for combo option "C++ language standard to use" is not one of the choices. Possible choices are (as string): "none", "c++11", "vc++11", "c++14", "c++latest", "vc++latest", "vc++14", "c++17", "vc++17", "c++20", "vc++20".
```
(`meson build -Dcpp_std=vc++14` allows configuration to complete, but should not be necessary)

**Expected behavior**
When cpp_std is set to gnu++*ver*, the msvc compiler target should treat it as if it was set to vc++*ver*

(Alternatively, there could be an c++*ver*ext set of targets that means "C++*ver*, but without strict conformance settings" and can map to gnu++ or vc++ or whatever other compiler-specific C++ version as applicable)

**system parameters**
* Plain native build
* Windows/MSVC only
* Python version: whatever is bundled with the meson 0.61.2 MSI installer.
* `meson --version`: 0.61.2
* `ninja --version`: 1.10.2 (also happens with vs backend)

peterh:
Cross platform applications that rely on compiler extensions cannot
compile with c++, they must use either gnu++ or vc++. Unfortunately,
those two settings are mutually exclusive.

Simplify the selection of a default project cpp_std by allowing
VisualStudioCPPCompiler to recognize gnu++ as vc++.

Fixes #10015

### DIFF
--- a/mesonbuild/compilers/cpp.py
+++ b/mesonbuild/compilers/cpp.py
@@ -612,6 +612,13 @@ class VisualStudioLikeCPPCompilerMixin(CompilerMixinBase):
         'vc++17': (True, 17),
         'vc++20': (True, 20),
         'vc++latest': (True, "latest"),
+        'gnu++11' : (True, 11),
+        'gnu++14' : (True, 14),
+        'gnu++17' : (True, 17),
+        'gnu++1z' : (True, 17),
+        'gnu++2a' : (True, 20),
+        'gnu++20' : (True, 20),
+        'gnu++latest' : (True, "latest"),
         'c++11': (False, 11),
         'c++14': (False, 14),
         'c++17': (False, 17),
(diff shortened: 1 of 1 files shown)


peterh:
The unsuccessful Windows tests are because "This is a scheduled windows-2016 brownout. The windows-2016 environment is deprecated and will be removed on March 15, 2022. For more details, see https://github.com/actions/virtual-environments/issues/4312 "

jpakkane:
This PR changes things so that if you configure the a build dir with VS,  `meson configure` shows the various gnu dialects as valid options. This is quite confusing because VS does not really support those (and conversely the `vs++17` etc values do not show up with other compilers)

eli-schwartz:
I pointed the same thing out in the linked ticket.

> > #7611 resolved this issue for C by making "gnu99" mean "c99 with local extensions" and allowing that to compile on Windows. A similar change for C++ would resolve this issue.
> 
> This feels extremely wrong. We don't _actually_ support it for gnu99 either, even though we demote it. But the reason we allow that is because it accidentally used to work (the c_std was ignored for MSVC!) so we allowed and ignored it for compatibility with existing projects that regressed their builds.
> 
> There is no regression here if it never worked with MSVC. So the same rationale does not apply here and #10028 is wrong.
> 
> > (Alternatively, there could be an c++_ver_ext set of targets that means "C++_ver_, but without strict conformance settings" and can map to gnu++ or vc++ or whatever other compiler-specific C++ version as applicable)
> 
> This seems like a reasonable approach to take.

peterh:
Closing. If/when I get time, I'll open a new pull request with the alternative approach.

PR #10030, opened 2022-02-24
Add support for EnvironmentVariables in Rust targets

dcbaker:
Idiomatic usage of Rust requires the use of environment variables at a compile time. This can be seen in the common idiom:
```rust
include!(concat!(env!("OUT_DIR"), "/generated.rs"))
```
Which is a very common thing to do. While I think the structured_sources proposal is more elegant and generally a better idea, both for the purpose of building cargo based projects with meson, and for the purpose of including incomplete snippets, we should support environment variables for Rust based targets.

This series adds that support by adding the `env` keyword to build_targets (usage in non-rust targets is considered an error), and generates two different rust rules, one for env wrapping and one for non wrapping.

### DIFF
--- a/mesonbuild/scripts/meson_exe.py
+++ b/mesonbuild/scripts/meson_exe.py
@@ -3,7 +3,6 @@
 
 from __future__ import annotations
 
-from __future__ import annotations
 import os
 import sys
 import argparse
@@ -12,7 +11,7 @@ import subprocess
 import typing as T
 import locale
 
-from ..utils.core import ExecutableSerialisation
+from ..utils.core import ExecutableSerialisation, EnvironmentVariables
 
 if T.TYPE_CHECKING:
     from typing_extensions import Protocol
@@ -22,6 +21,7 @@ if T.TYPE_CHECKING:
         unpickle: bool
         capture: bool
         feed: bool
+        env: T.List[str]
 
 
 def buildparser() -> argparse.ArgumentParser:
@@ -29,6 +29,7 @@ def buildparser() -> argparse.ArgumentParser:
     parser.add_argument('--unpickle')
     parser.add_argument('--capture')
     parser.add_argument('--feed')
+    parser.add_argument('--env', action='append', default=[])
     return parser
 
 def run_exe(exe: ExecutableSerialisation, extra_env: T.Optional[T.Dict[str, str]] = None) -> int:
@@ -119,7 +120,12 @@ def run(args: T.List[str]) -> int:
             exe = pickle.load(f)
             exe.pickled = True
     else:
-        exe = ExecutableSerialisation(cmd_args, capture=options.capture, feed=options.feed)
+        env = EnvironmentVariables()
+        for e in options.env:
+            name, value = e.split('=')
+            # This is internal, we can pick whatever separator we want!
+            env.append(name, value.split(':'))
+        exe = ExecutableSerialisation(cmd_args, capture=options.capture, feed=options.feed, env=env)
 
     return run_exe(exe)
 
--- a/mesonbuild/scripts/meson_exe.py
+++ b/mesonbuild/scripts/meson_exe.py
@@ -3,6 +3,7 @@
 
 from __future__ import annotations
 
+from __future__ import annotations
 import os
 import sys
 import argparse
@@ -12,6 +13,16 @@ import typing as T
 import locale
 
 from ..utils.core import ExecutableSerialisation
+
+if T.TYPE_CHECKING:
+    from typing_extensions import Protocol
+
+    class Args(Protocol):
+
+        unpickle: bool
+        capture: bool
+        feed: bool
+
 
 def buildparser() -> argparse.ArgumentParser:
     parser = argparse.ArgumentParser(description='Custom executable wrapper for Meson. Do not run on your own, mmm\'kay?')
@@ -93,6 +104,7 @@ def run_exe(exe: ExecutableSerialisation, extra_env: T.Optional[T.Dict[str, str]
 
 def run(args: T.List[str]) -> int:
     parser = buildparser()
+    options: Args
     options, cmd_args = parser.parse_known_args(args)
     # argparse supports double dash to separate options and positional arguments,
     # but the user has to remove it manually.
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -44,6 +44,7 @@ if T.TYPE_CHECKING:
     from ..linkers.linkers import DynamicLinker, StaticLinker
     from ..compilers.cs import CsCompiler
     from ..compilers.fortran import FortranCompiler
+    from ..compilers.rust import RustCompiler
 
     CommandArgOrStr = T.List[T.Union['NinjaCommandArg', str]]
     RUST_EDITIONS = Literal['2015', '2018', '2021']
@@ -1846,7 +1847,7 @@ class NinjaBackend(backends.Backend):
         return target.rust_dependency_map.get(dependency.name, dependency.name).replace('-', '_')
 
     def generate_rust_target(self, target: build.BuildTarget) -> None:
-        rustc = target.compilers['rust']
+        rustc = T.cast('RustCompiler', target.compilers['rust'])
         # Rust compiler takes only the main file as input and
         # figures out what other files are needed via import
         # statements and magic.
@@ -2063,8 +2064,9 @@ class NinjaBackend(backends.Backend):
             for rpath_arg in rpath_args:
                 args += ['-C', 'link-arg=' + rpath_arg + ':' + os.path.join(rustc.get_sysroot(), 'lib')]
 
-        if target.env:
-            args.extend(rustc.get_env_args(target.env))
+        compiler_name = self.compiler_to_rule_name(rustc)
+        if rustc.needs_env_wrapper:
+            compiler_name = self.get_compiler_rule_name(rustc.get_language(), rustc.for_machine, 'ENV')
 
         proc_macro_dylib_path = None
         if getattr(target, 'rust_crate_type', '') == 'proc-macro':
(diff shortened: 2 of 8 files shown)


tristan957:
This is also necessary for `GResource`s in Rust. Thanks for adding it.

eli-schwartz (changes requested):
> Idiomatic usage of Rust requires the use of environment variables at a compile time. This can be seen in the common idiom:
> 
> ```
> include!(concat!(env!("OUT_DIR"), "/generated.rs"))
> ```

This topic has come up before, and it might be "idiomatic" usage of Rust but it's not "idiomatic compiler design". Previous discussion on the topic I'm pretty sure was targeting getting a rustc fix for this.

It's basically terrible to do this for including generated files, and it's outright horrible to do this for any other reason -- what are people doing, using env variables to select target features? To define constants used for switching out runtime logic?

What's the use case for this other than describing the `current_build_dir()` and why do we need a generic solution to do frankly bad stuff? What's the solution to not requiring a "wrapped due to env" design for all sources involved in this target, rather than just the one source that cats out a generated file into the current file by absolute path underneath a pile of macros?

Given that `structured_sources` is such a better more elegant solution to this problem, why merge a worse solution at all?

w.r.t mention of crate subprojects, I'd consider that to be best solved by handling this inside crate conversion to meson rules, not something that requires an extension to the Meson DSL (which crate subprojects wouldn't use... right?)

tristan957:
This is the documentation: https://doc.rust-lang.org/std/macro.env.html. @eli-schwartz, I think you are taking the example posted too literally. While including a generated file is a solved problem in Meson, the `env!()` macro is useful for other things too.

eli-schwartz:
Including a generated file in rust is a solved problem using meson?

...

Thanks for the rust docs link, but it doesn't tell me anything I didn't know and acknowledge in my original comment. I was never in doubt that rust lets you do this, I simply challenge the notion that it's a good "compiler design" (i.e. rust was wrong to allow it) and that people should use it when it is allowed.

The original PR asserted that it is useful, but didn't say why. Now you assert that it is useful too, but also don't say why. I, meanwhile, am asking why this thing that seems to be a terrible idea is in fact not only good, but so good that we should extend the meson DSL in a way that no other compiler supports and which slows down compilation by wrapping rust compile rules in a python script for something that is, of necessity, hardcoded into meson.build and always known at buildsystem generation time.

So that ***source code files*** can process environment variables.

Environment variables. Which meson loves so much the lead developer of meson would prefer to ban them entirely.

I think it's not unreasonable of me to try to understand why this new feature is needed and what the practical use cases are.

I'm perfectly okay with meson "forbidding someone from using a feature of their language effectively" when that feature is *exclusively* of use to move build system logic into source code files because cargo isn't a real build system.

tristan957:
I mean it really is just an alternative to `configure_file()`. It is just crappier because the Rust community decided that Cargo is an amazing build system, which it isn't. Because Cargo is so bad, Rust has `env!()`. That is pretty much all there is to it.

I thought this comment was interesting:

> w.r.t mention of crate subprojects, I'd consider that to be best solved by handling this inside crate conversion to meson rules, not something that requires an extension to the Meson DSL (which crate subprojects wouldn't use... right?)

Maybe the problem could be solved by using `configure_file()` for a `.env` file, and then after `meson setup` someone could source the env file, and then hopefully `rustc` just picks up the environment variables that way?

dcbaker:
> w.r.t mention of crate subprojects, I'd consider that to be best solved by handling this inside crate conversion to meson rules, not something that requires an extension to the Meson DSL (which crate subprojects wouldn't use... right?)

What are you proposing? that I write a Rust parser in Meson and process every Rust file with a heuristic to determine when we could replace it with `use` or `mod` instead? 

> Maybe the problem could be solved by using configure_file() for a .env file, and then after meson setup someone could source the env file, and then hopefully rustc just picks up the environment variables that way?

That requires altering the source files, which isn't allowed in a wrap.

Among other things though, bindgen recommends this as *the* way to handle generated sources: https://rust-lang.github.io/rust-bindgen/tutorial-4.html

This may not be the way I'd design a compiler, but it *is* how Rust works, and refusing to make Rust work because it's not ideal isn't practical, we might as well just remove Rust support if we're going to refuse to implement common use cases because we don't find them pretty enough.

eli-schwartz:
> What are you proposing? that I write a Rust parser in Meson and process every Rust file with a heuristic to determine when we could replace it with `use` or `mod` instead?

No, I just want people writing meson.build for their own rust code to do it properly. :p 

Do we actually need it for wraps? I thought the idea was to inspect a crate and convert cargo's output to build.ninja rules somehow. That doesn't require having people literally add an `env` kwarg to an actual meson.build for the crate.

dcbaker:
No, but it likely requires the converted crate to add the keyword arguments because the rust files in the crate use `include!(env!(...))` because that's idiomatically how you include generated sources in Cargo, and thus how you generate them in Rust

tristan957:
Was not thinking of the wrap use case when I made my comment. Thanks for reminding me!

eli-schwartz:
Again, I thought the idea for crates was *not* to just add meson.build patches and add a bajillion things to the wrapdb, but to actually introspect cargo's output and use it to generate build.ninja rules similar to how cmake.subproject works from a high-level perspective, no meson.build involved.

dcbaker:
> Again, I thought the idea for crates was not to just add meson.build patches and add a bajillion things to the wrapdb, but to actually introspect cargo's output and use it to generate build.ninja rules similar to how cmake.subproject works from a high-level perspective, no meson.build involved.

Those rust files may already be using `env!()`, which we need to support, otherwise we have to patch the .rs files. There's no other solutions than:
1) support a feature of cargo + rust, (ie, this PR) so that meson works seamlessly
2) patch Rust sources to remove any use of `env!()` (assuming that's always possible, since Rust doesn't have a pre-processor)

mesonbuild/build.py
```diff
@@ -1050,6 +1055,8 @@ def process_kwargs(self, kwargs, environment):
         self.process_kwargs_base(kwargs)
         self.copy_kwargs(kwargs)
         kwargs.get('modules', [])
+        if kwargs.get('env'):
+            self.env = EnvironmentVariables(kwargs['env'])
```
xclaesse, line 1059: You should use type_checking._env_convertor()

mesonbuild/interpreter/interpreter.py
```diff
@@ -1608,26 +1608,31 @@ def func_disabler(self, node, args, kwargs):
 
     @FeatureNewKwargs('executable', '0.42.0', ['implib'])
     @FeatureNewKwargs('executable', '0.56.0', ['win_subsystem'])
+    @FeatureNewKwargs('executable', '0.62.0', ['env'])
```
xclaesse, line 1611: 0.63

mesonbuild/scripts/meson_exe.py
```diff
@@ -117,7 +129,12 @@ def run(args: T.List[str]) -> int:
             exe = pickle.load(f)
             exe.pickled = True
     else:
-        exe = ExecutableSerialisation(cmd_args, capture=options.capture, feed=options.feed)
+        env = build.EnvironmentVariables()
+        for e in options.env:
+            name, value = e.split('=')
+            # This is internal, we can pick whatever separator we want!
+            env.append(name, value.split(':'))
```
xclaesse, line 136: I'm not so sure, that's going to split `C:\foo` on Windows. Also you should not assume user wants to append to current env, they could want to set() or prepend(). I think for that you should pickle an ExecutableSerialisation() that contains the original EnvironmentVariables() object, instead of adding `--env` in the command line.

dcbaker:
This has been updated to use the nightly --set-env option, though the fallback to use a wrapper is currently not implemented. That will need to be reimplemented

xclaesse:
@dcbaker Wondering if we should instead pretend Rust use -DFOO=BAR like every sane compilers, and make `CompilerArgs.to_native()` translate them to `--env-set FOO=BAR`.

docs/markdown/snippets/rust_env_variables.md
```diff
@@ -0,0 +1,22 @@
...
+idiomatic Rust usage. This may now be done with the `env` keyword argument,
+which works like the one to `custom_target`:
+
+```meson
+ct = custom_target('foo', command : ['foo', '--outdir', '@OUTDIR@'])
+
+executable(
+  'main'
+  'main.rs',
+  env : {'OUT_DIR' : meson.current_build_dir()}
```
xclaesse, line 20: As far as I know, only Rust is crazy enough to use env at compile time, this should at least be renamed to `rust_env`.

tristan957:
> @dcbaker Wondering if we should instead pretend Rust uses `-DFOO=BAR` like every sane compilers, and make `CompilerArgs.to_native()` translate them to `--env-set FOO=BAR`. That would make stuff like `add_project_arguments()` useful for Rust as well.

I don't quite understand what you are saying, but it doesn't sound great.

```meson
add_project_arguments('--set-env=Hello=World', language: 'rust')
```

xclaesse:
> the fallback to use a wrapper is currently not implemented

Not sure we need to implement it. We could just make env usage requiring recent rustc version. Rust devs are used to always use latest version, that's not a problem.

eli-schwartz:
Yes, I think it's extremely reasonable to say that we support this but only with a recent enough rustc version that it can be done sanely.

I've objected all along to adding a python wrapper which juggles os.environ -- the --set-env argument to rustc brings this more in line with -D from C/C++, which is fantastic news and makes this sane to implement. We should leave it at that.

xclaesse:
> I don't quite understand what you are saying, but it doesn't sound great.
> 
> ```meson
> add_project_arguments('--set-env=Hello=World', language: 'rust')
> ```

That's fair enough. In that case we don't need `env` kwarg at all, we can do:
```meson
executable(
  'main'
  'main.rs',
  rust_args : ['--env-set=OUT_DIR=@0@'.format(meson.current_build_dir()],
)
```

dcbaker:
> I've objected all along to adding a python wrapper which juggles os.environ -- the --set-env argument to rustc brings this more in line with -D from C/C++, which is fantastic news and makes this sane to implement. We should leave it at that.

Unfortunately it's not yet stable, so we're basically telling people "you have to use an unstable feature in a nightly compiler, sorry. That's not great user experience. I was going to propose that the env wrapper be added with an explicit deprecation and remove in 1.6 or 1.7, with the argument that if it's stabilized soon, that there would be a least a few rustc versions with it stabilized and we're no longer telling people "hey this thing that you really need to do idiomatic rust (ie, crates) isn't available, sorry", but we're also being clear that this is a stop gap and we're not going to provide long term fallback support since it's slow and bad. It also doesn't change anything from a language point of view so the cost from our point of view is fairly small. I might be convinced to do away with it entirely, but that assumes that it's stabilized in the next rust release. 

We certainly could drop the `env` keyword argument, I had mostly used it because originally this was only implemnted as a wrapper (it's obviously been in the works a long time). I think my preference would be to drop `env` entirely. I'm nervious about the `-D` thing, since what do we do if rust adds support for `-D` to mean somethign else in the future?

dcbaker:
This push is ugly, it needs some patches split, doc changes, etc. However, it does away with the `env` keyword, and just reads the `--set-env` command directly in the backend to decide to use the wrapper. The result is that we have barely any code required for the wrapper (and I can probably tuck it all up in one comment for easy removal later), to get a useful feature, and it's still only used if the feature is enabled.

bonzini:
Closing in favor of #15837.

PR #10038, opened 2022-02-27
sourceset module: fix minor regression in making sourcesets immutable

eli-schwartz:
In commit c0be7e05b070d85b38c79088df882970a5cd0279 the setting of merged sourcesets as immutable in a loop accidentally got dedented, and only applied to the last loop iteration.

Fixes regression in #9854


dcbaker (approved):
Thanks for fixing this! I think I probably misread s as self… it’ss Shane we don’t have tests to cover this

PR #10039, opened 2022-02-28
dependencies: allow get_variable to expose files from subprojects

eli-schwartz:
There are somewhat common, reasonable and legitimate use cases for a dependency to provide data files installed to /usr which are used as command inputs. When getting a dependency from a subproject, however, the attempt to directly construct an input file from a subproject results in a sandbox violation. This means not all dependencies can be wrapped as a subproject.

One example is wayland-protocols XML files which get scanned and used to produce C source files.

Teach Meson to recognize when a string path is the result of fetching a dep.get_variable(), and special case this to be exempt from subproject violations.

### DIFF
--- a/mesonbuild/dependencies/cmake.py
+++ b/mesonbuild/dependencies/cmake.py
@@ -627,17 +627,23 @@ class CMakeDependency(ExternalDependency):
     def get_variable(self, *, cmake: T.Optional[str] = None, pkgconfig: T.Optional[str] = None,
                      configtool: T.Optional[str] = None, internal: T.Optional[str] = None,
                      default_value: T.Optional[str] = None,
-                     pkgconfig_define: T.Optional[T.List[str]] = None) -> T.Union[str, T.List[str]]:
+                     pkgconfig_define: T.Optional[T.List[str]] = None) -> str:
         if cmake and self.traceparser is not None:
             try:
                 v = self.traceparser.vars[cmake]
             except KeyError:
                 pass
             else:
-                if len(v) == 1:
-                    return v[0]
-                elif v:
-                    return v
+                # CMake does NOT have a list datatype. We have no idea whether
+                # anything is a string or a string-separated-by-; Internally,
+                # we treat them as the latter and represent everything as a
+                # list, because it is convenient when we are mostly handling
+                # imported targets, which have various properties that are
+                # actually lists.
+                #
+                # As a result we need to convert them back to strings when grabbing
+                # raw variables the user requested.
+                return ';'.join(v)
         if default_value is not None:
             return default_value
         raise DependencyException(f'Could not get cmake variable and no default provided for {self!r}')
--- a/mesonbuild/interpreter/interpreter.py
+++ b/mesonbuild/interpreter/interpreter.py
@@ -670,6 +670,18 @@ class Interpreter(InterpreterBase, HoldableObject):
         d_module_versions = extract_as_list(kwargs, 'd_module_versions')
         d_import_dirs = self.extract_incdirs(kwargs, 'd_import_dirs')
         final_deps = []
+        srcdir = Path(self.environment.source_dir)
+        # convert variables which refer to an -uninstalled.pc style datadir
+        for k, v in variables.items():
+            try:
+                p = Path(v)
+            except ValueError:
+                continue
+            else:
+                if not self.is_subproject() and srcdir / self.subproject_dir in p.parents:
+                    continue
+                if p.is_absolute() and p.is_dir() and srcdir / self.root_subdir in p.resolve().parents:
+                    variables[k] = P_OBJ.DependencyVariableString(v)
         for d in deps:
             if not isinstance(d, (dependencies.Dependency, dependencies.ExternalLibrary, dependencies.InternalDependency)):
                 raise InterpreterException('Dependencies must be external deps')
@@ -2718,8 +2730,9 @@ external dependencies (including libraries) must go to "dependencies".''')
     @noKwargs
     def func_join_paths(self, node: mparser.BaseNode, args: T.Tuple[T.List[str]], kwargs: 'TYPE_kwargs') -> str:
         parts = args[0]
+        other = os.path.join('', *parts[1:]).replace('\\', '/')
         ret = os.path.join(*parts).replace('\\', '/')
-        if isinstance(parts[0], P_OBJ.DependencyVariableString):
+        if isinstance(parts[0], P_OBJ.DependencyVariableString) and '..' not in other:
             return P_OBJ.DependencyVariableString(ret)
         else:
             return ret
@@ -2783,7 +2796,7 @@ Try setting b_lundef to false instead.'''.format(self.coredata.options[OptionKey
             norm = Path(fname)
             # variables built from a dep.get_variable are allowed to refer to
             # subproject files, as long as they are scheduled to be installed.
-            if norm.is_absolute() and '..' not in norm.parts and validate_installable_file(norm):
+            if validate_installable_file(norm):
                 return
         norm = Path(srcdir, subdir, fname).resolve()
         if os.path.isdir(norm):
--- a/mesonbuild/interpreter/interpreter.py
+++ b/mesonbuild/interpreter/interpreter.py
@@ -420,6 +420,7 @@ class Interpreter(InterpreterBase, HoldableObject):
             bool: P_OBJ.BooleanHolder,
             str: P_OBJ.StringHolder,
             P_OBJ.MesonVersionString: P_OBJ.MesonVersionStringHolder,
+            P_OBJ.DependencyVariableString: P_OBJ.DependencyVariableStringHolder,
 
             # Meson types
             mesonlib.File: OBJ.FileHolder,
@@ -2716,7 +2717,12 @@ external dependencies (including libraries) must go to "dependencies".''')
     @typed_pos_args('join_paths', varargs=str, min_varargs=1)
     @noKwargs
     def func_join_paths(self, node: mparser.BaseNode, args: T.Tuple[T.List[str]], kwargs: 'TYPE_kwargs') -> str:
-        return os.path.join(*args[0]).replace('\\', '/')
+        parts = args[0]
+        ret = os.path.join(*parts).replace('\\', '/')
+        if isinstance(parts[0], P_OBJ.DependencyVariableString):
+            return P_OBJ.DependencyVariableString(ret)
+        else:
+            return ret
 
     def run(self) -> None:
         super().run()
@@ -2759,6 +2765,26 @@ Try setting b_lundef to false instead.'''.format(self.coredata.options[OptionKey
     # declare_dependency).
     def validate_within_subproject(self, subdir, fname):
         srcdir = Path(self.environment.source_dir)
+        builddir = Path(self.environment.build_dir)
+        if isinstance(fname, P_OBJ.DependencyVariableString):
+            def validate_installable_file(fpath: Path) -> bool:
+                installablefiles: T.Set[Path] = set()
+                for d in self.build.data:
+                    for s in d.sources:
+                        installablefiles.add(Path(s.absolute_path(srcdir, builddir)))
+                installabledirs = [str(Path(srcdir, s.source_subdir)) for s in self.build.install_dirs]
+                if fpath in installablefiles:
+                    return True
+                for d in installabledirs:
+                    if str(fpath).startswith(d):
+                        return True
+                return False
+
+            norm = Path(fname)
+            # variables built from a dep.get_variable are allowed to refer to
+            # subproject files, as long as they are scheduled to be installed.
+            if norm.is_absolute() and '..' not in norm.parts and validate_installable_file(norm):
+                return
         norm = Path(srcdir, subdir, fname).resolve()
         if os.path.isdir(norm):
             inputtype = 'directory'
--- a/test cases/common/251 subproject dependency variables/subprojects/subfiles/meson.build
+++ b/test cases/common/251 subproject dependency variables/subprojects/subfiles/meson.build
@@ -1,0 +1,26 @@
+project('dependency variable resource')
+
+files_dep = declare_dependency(
+    variables: [
+      'pkgdatadir=@0@/subdir'.format(meson.current_source_dir()),
+      'pkgdatadir2=@0@/subdir2'.format(meson.current_source_dir()),
+    ]
+)
+
+install_data('subdir/foo.c', install_dir: get_option('datadir') / 'subdir')
+install_subdir('subdir2', install_dir: get_option('datadir'))
+
+import('pkgconfig').generate(
+    name: 'depvar_resource',
+    description: 'Get a resource file from pkgconfig or a subproject',
+    version: '0.1',
+    variables: [
+      'pkgdatadir=${datadir}/subdir',
+      'pkgdatadir2=${datadir}/subdir2',
+    ],
+    uninstalled_variables: [
+      'pkgdatadir=@0@/subdir'.format(meson.current_source_dir()),
+      'pkgdatadir2=@0@/subdir2'.format(meson.current_source_dir()),
+    ],
+    dataonly: true,
+)
--- a/test cases/failing/123 subproject sandbox violation/meson.build
+++ b/test cases/failing/123 subproject sandbox violation/meson.build
@@ -6,10 +6,19 @@ sub1_mustfail = sub1_d.get_variable('dir') / '..' / 'file.txt'
 sub2_d = subproject('subproj2').get_variable('d')
 sub2_mustfail = sub2_d.get_variable('dir') / 'file.txt'
 
+main_d = declare_dependency(
+    variables: [
+        'dir=@0@'.format(meson.current_source_dir()),
+    ]
+)
+main_mustfail = main_d.get_variable('dir') / 'subprojects/subproj3/file.txt'
+
 if get_option('failmode') == 'parent-dir'
     mustfail = sub1_mustfail
 elif get_option('failmode') == 'not-installed'
     mustfail = sub2_mustfail
+elif get_option('failmode') == 'root-subdir'
+    mustfail = main_mustfail
 endif
 
 custom_target(
--- a/test cases/failing/123 subproject sandbox violation/meson.build
+++ b/test cases/failing/123 subproject sandbox violation/meson.build
@@ -1,0 +1,25 @@
+project('subproject-sandbox-violation')
+
+sub1_d = subproject('subproj1').get_variable('d')
+sub1_mustfail = sub1_d.get_variable('dir') / '..' / 'file.txt'
+
+sub2_d = subproject('subproj2').get_variable('d')
+sub2_mustfail = sub2_d.get_variable('dir') / 'file.txt'
+
+if get_option('failmode') == 'parent-dir'
+    mustfail = sub1_mustfail
+elif get_option('failmode') == 'not-installed'
+    mustfail = sub2_mustfail
+endif
+
+custom_target(
+    'mustfail',
+    input: mustfail,
+    output: 'file.txt',
+    command: [
+        'python3', '-c',
+        'import os; shutil.copy(sys.argv[1], sys.argv[2])',
+        '@INPUT@',
+        '@OUTPUT@'
+    ],
+)
--- a/docs/markdown/Dependencies.md
+++ b/docs/markdown/Dependencies.md
@@ -90,6 +90,28 @@ following will happen: If 'default_value' was provided that value will
 be returned, if 'default_value' was not provided then an error will be
 raised.
 
+## Dependencies that provide resource files
+
+Sometimes a dependency provides installable files which other projects then
+need to use. For example, wayland-protocols XML files.
+
+```meson
+foo_dep = dependency('foo')
+foo_datadir = foo_dep.get_variable('pkgdatadir')
+custom_target(
+    'foo-generated.c',
+    input: foo_datadir / 'prototype.xml',
+    output: 'foo-generated.c',
+    command: [generator, '@INPUT@', '@OUTPUT@']
+)
+```
+
+*Since 0.63.0* these actually work as expected, even when they come from a
+(well-formed) internal dependency. This only works when treating the files to
+be obtained as interchangeable with a system dependency -- e.g. only public
+files may be used, and leaving the directory pointed to by the dependency is
+not allowed.
+
 # Declaring your own
 
 You can declare your own dependency objects that can be used
--- a/mesonbuild/dependencies/base.py
+++ b/mesonbuild/dependencies/base.py
@@ -216,7 +216,7 @@ class Dependency(HoldableObject):
     def get_variable(self, *, cmake: T.Optional[str] = None, pkgconfig: T.Optional[str] = None,
                      configtool: T.Optional[str] = None, internal: T.Optional[str] = None,
                      default_value: T.Optional[str] = None,
-                     pkgconfig_define: T.Optional[T.List[str]] = None) -> T.Union[str, T.List[str]]:
+                     pkgconfig_define: T.Optional[T.List[str]] = None) -> str:
         if default_value is not None:
             return default_value
         raise DependencyException(f'No default provided for dependency {self!r}, which is not pkg-config, cmake, or config-tool based.')
@@ -232,7 +232,7 @@ class InternalDependency(Dependency):
                  libraries: T.List[T.Union['BuildTarget', 'CustomTarget']],
                  whole_libraries: T.List[T.Union['BuildTarget', 'CustomTarget']],
                  sources: T.Sequence[T.Union['FileOrString', 'CustomTarget', StructuredSources]],
-                 ext_deps: T.List[Dependency], variables: T.Dict[str, T.Any],
+                 ext_deps: T.List[Dependency], variables: T.Dict[str, str],
                  d_module_versions: T.List[str], d_import_dirs: T.List['IncludeDirs']):
         super().__init__(DependencyTypeName('internal'), {})
         self.version = version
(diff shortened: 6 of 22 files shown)


eli-schwartz:
@emersion I have implemented this in the end.

mesonbuild/dependencies/cmake.py
```diff
@@ -637,7 +637,7 @@ def get_variable(self, *, cmake: T.Optional[str] = None, pkgconfig: T.Optional[s
                 if len(v) == 1:
                     return v[0]
                 elif v:
-                    return v
+                    raise DependencyException(f'Got cmake variable (always a list) for {self!r} with more than one value, violating design constraints')
```
eli-schwartz, line 640: @mensinda I don't understand what cmake is doing here. Please advise. Can we restrict this?

mesonbuild/interpreter/interpreter.py
```diff
@@ -2668,6 +2675,12 @@ def check_clang_asan_lundef(self) -> None:
     # subproject than it is defined in (due to e.g. a
     # declare_dependency).
     def validate_within_subproject(self, subdir, fname):
+        if isinstance(fname, P_OBJ.DependencyVariableString):
+            norm = Path(fname)
+            # variables built from a dep.get_variable are allowed to refer to
+            # subproject files. FIXME: check if they are installed?
```
eli-schwartz, line 2681: Note: I'm wondering what to do for this FIXME. I'm also wondering whether I *should* do this FIXME.

eli-schwartz (commented):
There are a couple of unclear spots here.

tristan957:
Thanks for submitting this. It is useful for my Python bindings I think.

xclaesse:
I noticed that `files('subprojects/foo/private-file')` actually bypass completely the sandbox check, gtk started doing that work around: https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/4465. We should forbid/deprecate that... but it's going to make everyone scream :P

Regarding this PR, similar to my MesonVersionString hack, not sure if it's genious or evil... both probably :D

eli-schwartz:
So the files() function never passes through the sandbox validator at all?

Luckily if it does pass through the validator it will still accept a DependencyVariableString and nothing breaks.

mesonbuild/dependencies/cmake.py
mensinda: In CMake, everything is a string and lists are strings with `;` separators. Thus the trace parser automatically splits the strings in the CMake variables to make processing easier.

So storing lists with more than one element in a variable is possible in CMake. If you we want to access them with `get_variable()` we either have to support lists as a return type or  `';'.join(v)` them and let the user in the `meson.build` deal with it.

eli-schwartz: It's fine if the traceparser wants to split them internally so that it can reference splitted strings when handling dependency targets.

I do not understand however, the use case of using `dep.get_variable()` on a splitted string.

Under what circumstances would a user actually try to do that from a meson.build?

eli-schwartz: I think in that case trying to use `${MY_FILES}` would be a sandbox violation anyway.

eli-schwartz:
I added an additional check to see whether the grabbed file has been installed by the subproject.

eli-schwartz:
Rebased on master.

eli-schwartz:
@mensinda okay, I'm not super motivated to argue about this so I return `';'.join(v)` now.

mesonbuild/dependencies/cmake.py
eli-schwartz: Fine, whatever :) we can return a joined string and let the user cope with it.

mesonbuild/interpreter/interpreter.py
```diff
@@ -2701,6 +2707,26 @@ def check_clang_asan_lundef(self) -> None:
...
+                for d in self.build.data:
+                    for s in d.sources:
+                        installablefiles.add(s.absolute_path(srcdir, builddir))
+                installabledirs = [str(Path(srcdir, s.source_subdir)) for s in self.build.install_dirs]
+                if fname in installablefiles:
+                    return True
+                for d in installabledirs:
+                    if fname.startswith(d):
+                        return True
+                return False
```
xclaesse, line 2723: While this is technically correct, I'm a little bit worried that it's a lot of loops every time we create a File object. Can't tell if that's going to be a significant performance hit without profiling real world projects though.

xclaesse: ah no sorry, only when picking files from subprojects... ok guess that's fine actually. Nevermind.

eli-schwartz: Right, so I hope the impact is at least minimal. When you do pick files from a subproject, we loop over every installable file "to date" but I'm not sure how else to do it.

Ultimately I really want this done on time for the merge window, so if a clever performance boost is thought up after the fact we can merge that as a bugfix.

xclaesse:
LGTM

xclaesse:
Oh, wait, would need a release notes, and find a place where we can document that magic trick. Probably a FeatureNew but not sure how to handle that.

eli-schwartz:
As a FeatureNew it feels a bit wrong, because this is supposed to be transparent to the superproject with regard to whether subproject fallback is done.

If a project depends on a subproject providing such files, then that project *normally* gets it from /usr via `dep.get_variable()` as a pkg-config dependency.

If someone adds a subproject wrap, it can *satisfy* this via the subproject, which on old versions of Meson fail due to sandbox issues, and with this PR work.

But the main project shouldn't add a minimum meson version of 0.62.0 just because it does `dependency('wayland-protocols')` and someone *might* run `meson wrap install wayland-protocols && meson setup builddir --wrap-mode=forcefallback`.

eli-schwartz:
I'm not sure what kind of documentation is correct to have, here. The ideal use case for this relaxation of permissions is that people use `dependency('foo')`, and we make subprojects "just work".

It's not a feature, per se, and the trick shouldn't be documented since it's not really a trick -- we're just making get_variable() fulfill its contract of equivalency with `'internal' == 'pkg-config'`.

eli-schwartz:
msys doesn't seem to like that. :/

jpakkane:
Rc1 is out and we want to minimize the number of changes made after that. However I have been told that this is a major issue for some. If this is really the case please confirm it here so and we'll add this to the next release.

emersion:
Hi! Yes, not having Meson subproject support for wayland-protocols would be quite a deal-breaker for me. The reasons are:

- Writing a new Wayland protocol extension, or extending an existing one, involves writing compositor and client patches for it. Usually the XML is written in parallel with implementations. Going back-and-forth between the XML and the implementations is a pain without subproject support, because one needs to setup a separate prefix, tell the implementation to use that prefix, don't forget to run `meson install` on every change, etc.
- Compositors and clients often times require a fresh wayland-protocols installation, because conditional compilation is not very workable. A wayland-protocols subproject allows our users to build the projects on non-rolling-release distros.

Up until now we've been using subprojects just fine, and I'm starting to see patches with a `files()` invocation to workaround the sandbox (abusing a Meson bug…). So it would be really great to see this PR merged! :)

test cases/common/251 subproject dependency variables/subprojects/subfiles/meson.build
```diff
@@ -0,0 +1,26 @@
+project('dependency variable resource')
+
+files_dep = declare_dependency(
+    variables: [
+      'pkgdatadir=@0@/subdir'.format(meson.current_source_dir()),
+      'pkgdatadir2=@0@/subdir2'.format(meson.current_source_dir()),
+    ]
+)
```
jpakkane, line 8: This is ... less than ideal.This is doing a complete end run around sandboxing and further the master project needs to know how the subproject has arranged its files. By corollary if you update the subproject so that it changes its file locations (which is its implementation detail) then things will fail to configure. The proper syntax for this is the regular:

```

files1_dep = declare_dependency(
    sources: files('subdir/foo.c'
)
```

This already works today, is less code for end users to write and fully obeys sandboxing.

eli-schwartz: No? It's not an end run around sandboxing at all. The master project needs to know *nothing* about how the subproject arranges its files. Rather, the master project needs to know that the subproject has documented:
- use `pkg-config --variable pkgdatadir` to retrieve a directory such as /usr/share/subdir/
- this installed directory has important data files, namely foo.xml or foo.c

So, the master project looks up a dependency, and asks that dependency where the installed datadir is. However, subprojects may specify that variable too, in order to provide the same *interface* as the installed project.

In order to pedantically guarantee that no one is *allowed* to violate the sandbox, we ensure two things:
- you are not permitted to use .. in this workflow
- the only files you are allowed to retrieve from the sandbox are ones that are marked for public installation

You cannot use a declare_dependency "sources" for data files like XML files to be processed by the wayland-protocols scanner, so that has never worked and never will work. My test case doesn't depend on wayland-scanner, so I figured using executable() and source files was the simplest way of testing the sandbox handling.

Also, even if this were only about getting source files for an executable, that declare_dependency requires special casing the difference between a subproject and a system dependency. In real use cases, you don't do `dependency('subfiles')` and expect an *installed* pkg-config dependency to provide source files as cflags. So you cannot use that declare_dependency as a wrap `[provide]` fallback for a system dependency.

jpakkane: > You cannot use a declare_dependency "sources" for data files like XML files to be processed by the wayland-protocols scanner, so that has never worked and never will work.

Then that is the actual feature we need to provide, a way for declared dependencies to carry arbitrary extra files. Maybe something like:

```
declare_dependency(extra_files: files('bob.xml'))
```

Alternatively there has been talk of adding a `directories` object in addition to a `files` one for cases like these. In that case you could more properly validate that when it is used in a different subproject you are not allowed to add any `..` segments to it (possibly any directory path segments).

Also based on a quick reading I already found a potential way to use this to escape the sandbox (I did not test if it actually works):

```
d = declare_dependency(variables: ['nastydir=path/to/a/different/subproject/'])
h4cker_file = d.get_variable('nastydir') / 'internal_file.c'
```

In any case storing information like this in plain strings is *bad* and must *never* be done. The innards should be hidden in an opaque object. Have we learned nothing from CMake?

In addition, new code should not use `join_paths`. We should probably deprecate that if we have not already.

emersion: > declare_dependency(extra_files: files('bob.xml'))

This sounds useful, but how would one make use of this? Do we need to add `dep.get_extra_file()` as well, which would look at pkgconfig variables or internal `extra_files`? We probably want to make `extra_files` a dict instead of a list?

One last thing which would be useful is to allow targets in `extra_files` as well, to allow executables and generated files to be passed from a subproject to its parent project.

eli-schwartz: > Then that is the actual feature we need to provide, a way for declared dependencies to carry arbitrary extra files. Maybe something like:
> 
> ```
> declare_dependency(extra_files: files('bob.xml'))
> ```

Please suggest how this can work with a system installed pkg-config dependency, if it doesn't work with a system-installed pkg-config dependency then projects will, rightfully, refuse to use it because they want to write code that works with system-installed dependencies but have subproject fallbacks "do the right thing".

> Alternatively there has been talk of adding a `directories` object in addition to a `files` one for cases like these. In that case you could more properly validate that when it is used in a different subproject you are not allowed to add any `..` segments to it (possibly any directory path segments).

This is LITERALLY a directories object, except that it doesn't require:
- adding a new function
- editing all existing meson.build which currently provides `declare_dependency(variables: ...)` as strings, to provide `variables: dir('.')` instead
- forcing all projects that want to make use of this functionality to bump the minimum version of Meson to the very latest bleeding edge version of Meson just so they can describe a string, which is used as a string and joined as a string, via a special function that returns `class DependencyVariableString(str)` except sed'ed out to `class DirectoryString(str)`, at which point we are exactly back to where I was in this PR, but with one more meson function to document.

> Also based on a quick reading I already found a potential way to use this to escape the sandbox (I did not test if it actually works):

It will not work, unless `path/to/a/different/subproject/` declares `internal_file.c` as an `install_data()` in which case I guess it is not actually an internal file. I considered this possibility and guarded against it.

In fact, my successfully figuring out how to ban this so that it doesn't work, is what made me confident enough to take this PR out of draft mode and consider it a viable PR.

> In any case storing information like this in plain strings is _bad_ and must _never_ be done. The innards should be hidden in an opaque object. Have we learned nothing from CMake?

Again, the ground source of truth is `/usr/share/wayland-protocols/stable/xdg-shell/xdg-shell.xml` and that *must* be handled as a string, and passed as a string as `input: ...` to a custom_target.

This PR allows `dependency('wayland-protocols').get_variable('pkgdatadir') / 'stable/xdg-shell/xdg-shell.xml'`, which successfully works ***and is the correct source specification methodology*** if get_variable() evaluates to /usr/share/wayland-protocols, to *also* work as a subproject where get_variable() evaluates to some install_data or install_subdir provided by the subproject.

> In addition, new code should not use `join_paths`. We should probably deprecate that if we have not already.

This is the wrong PR to discuss deprecating join_paths, and since this PR is backwards-compatible to meson.build files specifying a minimum version of meson from before any proposed deprecation, this PR *must* work with `join_paths()`.

eli-schwartz: > to allow targets in `extra_files` as well, to allow executables and generated files to be passed from a subproject to its parent project.

No, if you want a program installed on the system you want to use find_program() and we already handle that via program overrides.

And per my consistent theme in this discussion, I want subprojects to transparently act like system installed dependencies, which means we cannot have people doing if/else and using find_program if dep.type_name() evaluates to pkgconfig, but doing some declare_dependency thing if it evaluates to 'internal'.

The point of a subproject is to be transparent, so the code originally written for a system dependency should, as a Meson development goal, continue to work without modification if a dependency is of type 'internal'.

That's all I want here. This PR is in aid of that. No new API.

nirbheek: Eli, I actually agree with Jussi, as you know from our discussion on Matrix.

> This is LITERALLY a directories object, except that it doesn't require:
>  *  adding a new function

I think when we find ourselves painted into a corner because of our older API decisions, adding new API based on our learnings from that is a good thing.

>  *  editing all existing meson.build which currently provides declare_dependency(variables: ...) as strings, to provide variables: dir('.') instead

We've had to do this in the past a lot for subprojects (with gstreamer and GNOME), and we did it anyway because it led us to the right API (`[provide]` sections in wraps).

>  *  forcing all projects that want to make use of this functionality to bump the minimum version of Meson to the very latest bleeding edge version of Meson

They don't have to require the latest version, the code can be conditional on the current meson version. You need the latest meson version to get things to work anyway.

>  just so they can describe a string, which is used as a string and joined as a string, 

Except that it's not a string. It's a path. Strings are not paths. If someone is using strings as paths, that's because there are no APIs for path manipulation. A good example is `os.path` vs `pathlib`. I don't think we should continue down that path. I said the same thing on Matrix, and you seemed to agree (edited for clarity):

```
<nirbheek> I was just thinking that anytime we're using a string for a path, it's because we don't have the correct abstraction in place
<nirbheek> And if we're adding an abstraction here, we might as well make it the right one?
...
<nirbheek> I just think that `DependencyVariableString` is the wrong abstraction? Maybe we should have `dirs()` and the `includedir` is a `dirs()` for `declare_dependency` and you can get `files()` in the subdirs of that dir with `/`
<nirbheek> i.e., make the object that is in `variables:` rich instead of making the internal implementation of `variables:` rich?
<elibrokeit> so you want `declare_dependency` to provide some object type that isn't a string?
<elibrokeit> like `declare_dependency(variables: {'foodir': dirs('.')})`
<nirbheek> Yes
<elibrokeit> adding a `dirs()` function to the meson API for this seems a bit too niche, and I guess it would need to return a dedicated meson object type that subclasses a stringholder anyway
<nirbheek> I don't know if it's niche -- I think this is just the first compelling use-case for it
```

> via a special function that returns class DependencyVariableString(str) except sed'ed out to class DirectoryString(str), at which point we are exactly back to where I was in this PR, but with one more meson function to document.

Not quite. I suggested `dirs()` above, and so did Jussi, and it would work differently than what you say. On further thinking, I think `path()` makes more sense.

Producer:
```
proj_dep = declare_dependency(...
    # We should eventually deprecate the 'key=value' syntax, it's too stringy
    variables: {'datadir': path('.')})
```

Consumer:
```
proj_dep = dependency('proj-name')
datadir = proj_dep.get_variable('datadir')
protocol_xml = datadir / 'protocol.xml'
```

When `proj_dep` is provided by a system dep (pkgconfig, cmake, etc), `datadir` would be a string, because we have no way to know what the variable actually is.

When `proj_dep` is provided by a subproject, `datadir` would be a `path()` object, which can also be used as a string implicitly.

In both cases, `protocol_xml` should be a `path()` object, i.e., using the `/` operator on a `str` or a `path()` object should yield a `path()` object (which is usable interchangeably as an `str`).

We can either have a `path()` object be converted to a `files()` object explicitly with `.to_files()`, or implicitly by having all API that accepts `files()` also accept `path()`. I am not sure which one is better here, but usually explicit is better.

eli-schwartz: I apologize if my attempt to gently cast doubt on your proposal by pointing out that it adds niche API which has no practical use, sounded like I was agreeing with you.

That was never my intention. I continue to maintain that this entire line of thought is 100% wrong.

I simply cannot fathom your objection here. And I once more want to reiterate, that you're saying "not quite" and implying that `dirs()` or `path()` is somehow different from `DependencyVariableString(str)` while in fact they are the same thing, because as you pointed out they can be used interchangeably.

I would like projects to just work the expected way, and have meson intelligently detect when projects are doing the right thing and penalizing them for that right thing involving a subproject. You want to add new API that forces every project to edit their build files, adding additional cognitive overhead to the build language, and the ***only*** thing it does is produce a string which has been marked at the time of declaration as "not a sandbox violation" rather than at time of variable retrieval. Yes, that's right. You already explained that a path object would in all cases be usable as a string, and of course strings must always be usable as paths so that we can join them and use them as arguments to functions that expect strings describing a path (and which years of existing build files in thousands of projects are reasonably and sensibly using).

The actual defense against sandbox violations is my check that you can only use installed data files, and you need that check no matter what niche and cognitively strange API you propose I add to declare_dependency.

I am saddened that my attempt to make things better for Wayland is not going to be accepted, but oh well. Life isn't perfect and even though meson can't sensibly use subprojects in this case, it's still a better build system than cmake.

jpakkane: >  dirs() or path() is somehow different from DependencyVariableString(str) while in fact they are the same thing,

There is a major semantic difference. The return value of dirs could only point to an existing directory on the file system. It would be verified on creation and thus would be safer to use. A `DependencyVariableString` has no such guarantee, it would need to be validated on use.

Another thing is that Directory (or Path) objects could be extended to relative paths in command invocations. DependencyVariableStrings would have to be absolute. This is actually important for deterministic builds. I had to change the implementation a fair bit ages ago to use relative paths whenever possible because absolute paths do not work (I do not know why, just that this is what I had to to on request by the reproducible builds folks).

eli-schwartz: Anyway, I implemented this check anyway, even though it is pointless.

I also closed the door on
```meson
d = declare_dependency(variables: ['nastydir=path/to/a/different/subproject/'])
```

Note: ***even though*** this cannot ever work unless it refers to installed files.

However, with the latest push, now it won't work even to poke at another project's installed files, unless that project does the declare_dependency itself. Because it is tracked during declare_dependency, so we actually know which subproject it is part of.

eli-schwartz:
Previously this allowed a measured sandbox bypass under the following conditions:
- the accessed file must be installed via install_data or install_subdir
- the accessed file must not have a `..` anywhere in the path
- the accessed file must start with a declare_dependency() object's get_variable() return value

With the latest push:
- the accessed file must come from the same subproject that created the declare_dependency() object
- the accessed file can now have a `..` as long as it was part of the original `declare_dependency(variables: ...)`
- the accessed file cannot be created by a user with a puckish sense of humor experimenting with TOCTOU by:
  - creating the declare_dependency() directory and
  - creating the file processed by install_subdir or install_data

  after the declare_dependency() itself is created but before the install_data/install_subdir is performed by that same subproject (and of course before leaving the subproject and evaluating dep.get_variable() by the superproject). I do not know if this TOCTOU thing matters.

All three of these are checked during `declare_dependency()`, the status of this is then saved for later retrieval.

jpakkane:
I was thinking about this and finally understood what was bugging me about it. In addition to exposing directories, projects might want to expose files in them. If the file in that is generated, then the data stored in the internal variable may _not_ be a string. It must carry with it the dependency information to the target so that when you use the file in your generator or whatever, Meson sets up a dependency to the generated file. Thus you'd need to be able to put the output of, say, `custom_target` in the internal dependency's variable definition. Plain strings just won't cut it.

This is a separate issue of whether there should be a `dirs` object or not.

eli-schwartz:
> It must carry with it the dependency information to the target so that when you use the file in your generator or whatever, Meson sets up a dependency to the generated file. Thus you'd need to be able to put the output of, say, `custom_target` in the internal dependency's variable definition. Plain strings just won't cut it.

The thing is, that this is wrong.

ninja already does that. We need a dependency on the input file to make rebuilds work. As long as there is a separate output rule that creates the file, that output rule acts as a provider for the file, and the two depend on each other.

...

But also this doesn't matter, because that won't be interchangeable with `dependency('wayland-protocols').get_variable('pkgdatadir')`.

And also because wayland-protocols doesn't use built files, it uses install_data and install_subdir.

And this PR only supports those, because that is what can be done reliably (and also because: that is what projects need).

eli-schwartz:
So what's the conclusion here? Sounds like:
- a separate dirs() object doesn't actually make you/nirbheek happy
- actually using ***generated*** files from a subproject or from the installed pkg-config interface will require complicated hackery that:
  - hasn't been specified
  - I'm not willing to code
  - no one has come forward to say they need it
- this PR solves the use case of wayland-protocols
- this PR cannot be abused to do anything that we don't think is reliable. So in theory it doesn't help with certain use cases. It doesn't hurt either


"The perfect is the enemy of the good." We should merge this PR because it reliably helps projects like wayland-protocols without allowing bad things to happen.

jpakkane:
Going through this there are at least the following issues that came to mind (regardless of how this will be eventually implemented).

Removing support for array of strings might break some existing code. There should be a snippet mentioning this behavioural change.

There need to be failing tests that try to break the sandbox to verify the code is working as it should. Personally I would be fine with being even so strict that you can only access files under the subtree in question, not any parent subdir. Effectively this would mean that the joining function should check that any string that is to be joined to a dependencystring does not contain `..`.

Assuming a top level project and two subprojects, the following should be tested:

A `declare_dependency` in `subproj1` that tries to point to a directory in `subproj2`.

In the top level project do:

```
d = declare_dependency(
    variables: [
      'dir=@0@'.format(meson.current_source_dir()),
    ]
)
mustfail = d.get_variable('dir') / 'subprojects/subproj1'
```

There should be a comment above `DependencyVariableString` saying that if a proper directory object is added to Meson, then `DependencyVariableString` should be deprecated and removed.

eli-schwartz:
> Removing support for array of strings might break some existing code.

Currently, no support for anything is being removed.

> There need to be failing tests that try to break the sandbox to verify the code is working as it should.

Good point. Currently I have a succeeding test that checks that something works which did not used to, but not a failing test that checks that bad things are still forbidden.

> Personally I would be fine with being even so strict that you can only access files under the subtree in question, not any parent subdir. Effectively this would mean that the joining function should check that any string that is to be joined to a dependencystring does not contain `..`.

Note: the current implementation already forbids the use of `..` for exactly this reason.

eli-schwartz:
@jpakkane done

eli-schwartz:
@jpakkane ping?

test cases/failing/123 subproject sandbox violation/meson.build
```diff
@@ -0,0 +1,34 @@
...
+elif get_option('failmode') == 'root-subdir'
+    mustfail = main_mustfail
+endif
+
+custom_target(
+    'mustfail',
+    input: mustfail,
+    output: 'file.txt',
+    command: [
+        'python3', '-c',
```
jpakkane, line 29: This should use the Python module because `python3` is not available in plain `cmd.exe` (which we sadly don't have in our CI setup). See #10305 for a similar change for the `echo` command.

jpakkane:
We don't seem to have any documentation on how to properly use this. It should be added to  `dependencies.md`. It does not need to be thorough, just a brief mention emphasizing that you are not allowed to poke outside the given directory.

test cases/failing/123 subproject sandbox violation/meson.build
eli-schwartz: `custom_target(..., command: ['string'])` internally runs `self.find_program_impl(['string'])`, and find_program_impl already does fallback for `python3` -> `mesonlib.python_command`.

I prefer to not use the python module as that costs a module import, then does obscure things to set up extension_module introspection data.

jpakkane:
```
run_project_tests.py:1234:120: error: Argument "state" to "submit" of "Executor" has incompatible type "Tuple[List[str], List[str], List[str], List[str], List[str], Backend, List[str], Optional[str]]"; expected "GlobalState"
```

Seems relevant.

eli-schwartz:
Sadly, it is not relevant. It happens on all PRs because of a mypy update, and I submitted a PR for it here: https://github.com/mesonbuild/meson/pull/10324

PR #10041, opened 2022-02-28
wayland: Also lookup scanner in pkgconfig

xclaesse:
This moves generally useful logic from GNOME module's
_get_native_binary() into find_program() implementation. We could decide
later to expose it as public API.

### DIFF
--- a/mesonbuild/interpreter/interpreter.py
+++ b/mesonbuild/interpreter/interpreter.py
@@ -1472,12 +1472,12 @@ external dependencies (including libraries) must go to "dependencies".''')
                           required: bool = True, silent: bool = True,
                           wanted: T.Union[str, T.List[str]] = '',
                           search_dirs: T.Optional[T.List[str]] = None,
-                          version_func: T.Optional[T.Callable[[T.Union['ExternalProgram', 'build.Executable', 'OverrideProgram']], str]] = None
-                          ) -> T.Union['ExternalProgram', 'build.Executable', 'OverrideProgram']:
+                          version_func: T.Optional[T.Callable[[T.Union['ExternalProgram', 'build.Executable', 'OverrideProgram']], str]] = None,
+                          depname: T.Optional[str] = None, varname: T.Optional[str] = None) -> T.Union['ExternalProgram', 'build.Executable', 'OverrideProgram']:
         args = mesonlib.listify(args)
 
         extra_info: T.List[mlog.TV_Loggable] = []
-        progobj = self.program_lookup(args, for_machine, required, search_dirs, extra_info)
+        progobj = self.program_lookup(args, for_machine, required, search_dirs, depname, varname, extra_info)
         if progobj is None:
             progobj = self.notfound_program(args)
 
@@ -1521,7 +1521,9 @@ external dependencies (including libraries) must go to "dependencies".''')
         return progobj
 
     def program_lookup(self, args: T.List[mesonlib.FileOrString], for_machine: MachineChoice,
-                       required: bool, search_dirs: T.List[str], extra_info: T.List[mlog.TV_Loggable]
+                       required: bool, search_dirs: T.List[str],
+                       depname: T.Optional[str], varname: T.Optional[str],
+                       extra_info: T.List[mlog.TV_Loggable]
                        ) -> T.Optional[T.Union[ExternalProgram, build.Executable, OverrideProgram]]:
         progobj = self.program_from_overrides(args, extra_info)
         if progobj:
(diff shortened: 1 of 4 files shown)


xclaesse:
CC @markbolhuis and @eli-schwartz

fooishbar:
Yep, this looks reasonable to me, and matches existing use of wayland-scanner.

eli-schwartz:
I think it makes sense to look up tools based on pkg-config name for wayland-scanner.

I'm undecided on whether it belongs in find_program or perhaps moved from gnome to `state.get_native_binary()`. It feels a bit niche to expose as public API, but if it is not public API I'm not sure there's any real advantage to having it in find_program_impl?

xclaesse:
@eli-schwartz the reason I prefer having it there is to keep the find_program logic in a single place. `_get_native_binary()` was repeating the steps that goes before checking pkgconfig (checking overrides, machine files), easier like that if we ever want to change those steps, or add other steps in the future.

Also, I think eventually we should expose that into `find_program()` public API, we can't always get away with this by wrapping every tool in a Meson module.

xclaesse:
> crossed_swords conflicts.

Fixed.

xclaesse:
@jpakkane yes, but I'm not even sure it needs to be limited to pkgconfig, but then we open the can of worms where the name of variable could be different depending on the dep type, like we do in `dep.get_variable(pkgconfig: 'foo', cmake: 'FOO')`. Since those API details could need discussion I'm not doing that atm.

eli-schwartz (changes requested):
In #10140 it was pointed out that incorrect usage of pkg-config dependency checks in cases where the pkg-config file doesn't actually provide the referenced variable, does not emit a warning.

This seems wrong as I figure we do want a warning so that we can e.g. fix the code to not look up pkg-config at all.

Also if this was publicly exposed as a kwarg to find_program we would definitely want to warn people when their meson.build tries to look up invalid variables.

xclaesse:
This is done on purpose because those variables are typically added over time and you want it to work even if you have an older version that does not have the variable yet.

xclaesse:
For the same reason that finding the dependency in pkg-config at all is not required in that code.

xclaesse:
@eli-schwartz @jpakkane ping?

xclaesse:
Kindly reminder to get a review please.

eli-schwartz:
> For the same reason that finding the dependency in pkg-config at all is not required in that code.

No, that's a completely different reason. It makes perfect sense to not require a pkg-config dependency, because there are many cases e.g. split packages in distros where the pkg-config files may not be installed even though tools are installed.

We still do log that it wasn't found though. As we should.

Completely failing to log when we don't find something we expected to find, is absolutely not something we should be doing, it's 100% a bug in this PR regardless of anything else.

> This is done on purpose because those variables are typically added over time and you want it to work even if you have an older version that does not have the variable yet.

They are typically not added at all (projects which consider this a logical tool-finding approach are rare, I'm pretty sure glib cares because they want to make sure all tools are found for the same version of the glib ecosystem even if you have a system version and a dev copy) and I'm strongly opposed to making that fact be silently ignored.

eli-schwartz:
> Kindly reminder to get a review please.

My stance hasn't changed, because your response to my existing review is the same response you gave in the linked ticket and I didn't agree with your reasoning then either.

eli-schwartz:
Sure. And my point is that this is rare enough that I don't think we need special language in find_program to do this. In fact, instead of new language to look up both a program on PATH and a program from a pkg-config dependency, we could just do this:

```
# don't use:
find_program('foo-prog', dependency: 'fooproject', variable: 'foo_prog')

# do use
find_program(dependency('fooproject').get)_variable('foo_prog'), 'fooprog')
```

It's backwards compatible to basically any version of Meson. It finds both, using the same lookup order. It respects override_find_program. It lets you control important aspects of the dependency, such as `version: `. I don't see a single case where sticking the *same* dependency info into kwargs instead of a positional parameter, is better.

But it's also a moot point if per wayland upstream, it is only possible to use the pkg-config dependency, *never* find_program. So in the end, "look up both" is still only needed... by gtk?

xclaesse:
> find_program(dependency('fooproject').get_variable('foo_prog'), 'fooprog')

That won't work if `fooproject` dependency is not found, you have to do it like that:
```
dep = dependency('fooproject', required: false, native: true)
if dep.found()
  prog = find_program(dep.get_variable('foo_prog'))
else
  prog = find_program('fooprog')
endif
```

> But it's also a moot point if per wayland upstream, it is only possible to use the pkg-config dependency, _never_ find_program.

They say that because they think that relying on pkgconfig exclusively guarantees that they get the same version for wayland-scanner and libwayland-client. But that's not true when cross compiling, wayland-scanner pkgconfig file comes from the BUILD machine which can have completely different version as wayland-client that comes from HOST machine.

But the good news is find_program() actually can check the version and that works regardless if the program is found from pkgconfig, PATH or overridden by subproject.

> So in the end, "look up both" is still only needed... by gtk?

glib, gobject-introspection, gtk, basically everything that we cover already in the gnome module. Wayland will ultimately need it too, I think one of their dev now understood. I think Wayland module will just have to first lookup wayland-client or wayland-server dependency to get their version, then add the version requirement when doing find_program('wayland-scanner').

I agree given the current use-cases we have, dealing with this only in Meson's ModuleState.find_program() instead of doing it in Interpreter level is good enough. I still have a preference on doing this in Interpreter.program_lookup() to keep the logic of find_program() in a single place (as this PR does), but that's not a hill I'm ready to die on if you have strong opinion against it.

PR #10042, opened 2022-02-28
Deprecate configure_file(copy : true), and replace it with fs.copyfile

dcbaker:
This is based on #9897

It's a bit of a historical accident that `configure_file` is used for copying, mainly because it was able to happen because of loose argument handling, and that we didn't yet have an `fs` module to be a clear, obvious, place to put such a helper. The `copyfile` method is basically synonymous, but with the added value of not requiring an `output` keyword argument.

### DIFF
--- a/docs/markdown/Fs-module.md
+++ b/docs/markdown/Fs-module.md
@@ -215,3 +215,30 @@ fs.stem('foo/bar/baz.dll.a')  # baz.dll
    specified by `path` changes, this will trigger Meson to reconfigure the
    project. If the file specified by `path` is a `files()` object it
    cannot refer to a built file.
+
+
+### copyfile
+
+*Since 0.64.0*
+
+Copy a file from the source directory to the build directory at build time
+
+Has the following positional arguments:
+   - src `File | str`: the file to copy
+
+Has the following optional arguments:
+   - dest `str`: the name of the output file. If unset will be the basename of
+     the src argument
+
+Has the following keyword arguments:
+   - install `bool`: Whether to install the copied file, defaults to false
+   - install_dir `str`: Where to install the file to
+   - install_tag: `str`: the install tag to assign to this target
+   - install_mode `array[str | int]`: the mode to install the file with
+
+returns:
+   - a [[custom_target]] object
+
+```meson
+copy = fs.copyfile('input-file', 'output-file')
+```
--- a/mesonbuild/modules/fs.py
+++ b/mesonbuild/modules/fs.py
@@ -12,24 +12,28 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import typing as T
+from __future__ import annotations
+from pathlib import Path, PurePath, PureWindowsPath
 import hashlib
 import os
-from pathlib import Path, PurePath, PureWindowsPath
+import typing as T
 
+from . import ExtensionModule, ModuleReturnValue, ModuleInfo
 from .. import mlog
-from . import ExtensionModule, ModuleInfo
+from ..build import CustomTarget, InvalidArguments
+from ..interpreter.type_checking import INSTALL_KW, INSTALL_MODE_KW, INSTALL_TAG_KW, NoneType
+from ..interpreterbase import FeatureNew, KwargInfo, typed_kwargs, typed_pos_args, noKwargs
 from ..mesonlib import (
     File,
     MesonException,
+    has_path_sep,
     path_is_in_root,
 )
-from ..interpreterbase import FeatureNew, KwargInfo, typed_kwargs, typed_pos_args, noKwargs
 
 if T.TYPE_CHECKING:
     from . import ModuleState
     from ..interpreter import Interpreter
-    from ..mesonlib import FileOrString
+    from ..mesonlib import FileOrString, FileMode
 
     from typing_extensions import TypedDict
 
@@ -37,6 +41,15 @@ if T.TYPE_CHECKING:
         """Keyword Arguments for fs.read."""
 
         encoding: str
+
+    class CopyKw(TypedDict):
+
+        """Kwargs for fs.copy"""
+
+        install: bool
+        install_dir: T.Optional[str]
+        install_mode: FileMode
+        install_tag: T.Optional[str]
 
 
 class FSModule(ExtensionModule):
@@ -61,6 +74,7 @@ class FSModule(ExtensionModule):
             'name': self.name,
             'stem': self.stem,
             'read': self.read,
+            'copyfile': self.copyfile,
         })
 
     def _absolute_dir(self, state: 'ModuleState', arg: 'FileOrString') -> Path:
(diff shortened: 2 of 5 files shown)


eli-schwartz:
I like the idea of this, but I hate the implementation. Actually, I hate the configure_file implementation which this doesn't fix.

configure_file originally assumed that the input file needs to be modified by a configuration_data, for obvious reasons. Also for obvious reasons, this needs to be done at configure time, and any changes to the file template almost certainly entail changes to meson.build logic too. None of this applies to `copy: true`, it just inherited a weird design.

Could we have this generate a build.ninja rule to do the copying? I really dislike doing file copying at configure time and then having edits to the copied file force a full build system reconfigure. And I'm pretty sure I've seen people just make custom_target with Unix `cp` as the command because they didn't like  this either.

docs/markdown/Fs-module.md
```diff
@@ -215,3 +215,22 @@ fs.stem('foo/bar/baz.dll.a')  # baz.dll
...
+
+*Since 0.62.0*
+
+Copy a file from the source directory to the build directory
+
+Has the following positional arguments:
+   - src `File | str`: the file to copy
+
+Has the following optional arguments:
+   - dest `str`: the destination in the build dir to copy to. If unset will be the basename of the src
```
jpakkane, line 230: This is wrong. It should say filename and explicitly state that you can't set the output path and trying to do that is a fatal error.

jpakkane:
The design reason for `configure_file` was that the output can be used in any target without needing to add it to the list of sources manually. This is how `config.h` works and is generally what people expect.

Maybe this could have a kwarg specifying whether it should be done at build time or configure time (if the former is something we really want to support)?

tristan957:
I like `fs.copy()` better than `fs.copyfile()` since it is a more futureproof name in my opinion although on UNIX everything is a file :).

eli-schwartz:
> The design reason for `configure_file` was that the output can be used in any target without needing to add it to the list of sources manually. This is how `config.h` works and is generally what people expect.

Right, so `copy: true` inherited that design limitation because it would be somewhat unfortunate to have configure_file sometimes return a File and sometimes return a CustomTarget.

> Maybe this could have a kwarg specifying whether it should be done at build time or configure time (if the former is something we really want to support)?

... I don't really think it makes a lot of sense to make this optional? We can just make it return a CustomTarget. It is trivially added as a source to any build target, or to a declare_dependency, so we can guarantee it is up to date. Considering a fairly common use case of `copy: true` is "fix cmake projects which have headers in the wrong directory layout" I think people will always want the latter, never the former.

xclaesse:
If we copy at configure time it means we have to reconfigure the whole project each time the file is touched, instead of just rebuilding some targets.

Also I think it would make sense to take a list of files that all goes to the same destination. Useful for the headers use case.

xclaesse:
Note that if we copy at build time, maybe we could keep configure_file() not deprecated for doing it at configure time.

tristan957:
@ilayn instead of configuration kwarg, use the copy kwarg in your original code block

dcbaker:
@eli-schwartz, the only case I can come up with for when configure_time is necessary, is if you want to use the output for a `run_command`

dcbaker:
I've reworked the `fs.copyfile` to be a CustomTarget as requested by @eli-schwartz, and addressed @jpakkane's concerns about outputs having path seperators (both in the docs and in the code). I'm still not sure what to do about configure_time copying, because I did run into a case where you want that, and it's `run_commnad`.

mesonbuild/modules/fs.py
```diff
@@ -254,6 +268,47 @@ def read(self, state: 'ModuleState', args: T.Tuple['FileOrString'], kwargs: 'Rea
...
+        INSTALL_TAG_KW,
+        KwargInfo('install_dir', (str, NoneType)),
+    )
+    def copyfile(self, state: ModuleState, args: T.Tuple[FileOrString, T.Optional[str]],
+                 kwargs: CopyKw) -> ModuleReturnValue:
+        """Copy a file into the build directory at build time."""
+        if kwargs['install'] and not kwargs['install_dir']:
+            raise InvalidArguments('"install_dir" must be specified when "install" is true')
+
+        src = File.from_source_file(state.source_root, state.subdir, args[0]) if isinstance(args[0], str) else args[0]
```
eli-schwartz, line 286: We should use source_strings_to_files here.

eli-schwartz (changes requested):

docs/markdown/Fs-module.md
eli-schwartz: I don't see how this conversation is resolved, it still has that doc issue.

That being said, this is basically a fancy sort of custom_target and we just document the output kwarg as "list of output files", so I do not think we need to explicitly talk about how output paths are wrong.

I *do* think this should say "output file" not "destination".

docs/markdown/Fs-module.md
```diff
@@ -215,3 +215,29 @@ fs.stem('foo/bar/baz.dll.a')  # baz.dll
...
+
+Has the following keyword arguments:
+   - install `bool`: Whether to install the copied file, defaults to false
+   - install_dir `str`: Where to install the file to
+   - install_tag: `str`: the install tag to assign to this target
+   - install_mode `array[str | int]`: the mode to install the file with
+
+returns:
+   - a `custom_target` object
```
eli-schwartz, line 239:
```suggestion
   - a [[custom_target]]
```

mesonbuild/modules/fs.py
```diff
@@ -254,6 +268,47 @@ def read(self, state: 'ModuleState', args: T.Tuple['FileOrString'], kwargs: 'Rea
...
+        INSTALL_TAG_KW,
+        KwargInfo('install_dir', (str, NoneType)),
+    )
+    def copyfile(self, state: ModuleState, args: T.Tuple[FileOrString, T.Optional[str]],
+                 kwargs: CopyKw) -> ModuleReturnValue:
+        """Copy a file into the build directory at build time."""
+        if kwargs['install'] and not kwargs['install_dir']:
+            raise InvalidArguments('"install_dir" must be specified when "install" is true')
+
+        src = self.interpreter.source_strings_to_files([args[0]], strict=True)[0]
```
eli-schwartz, line 286: strict is already the default?

mesonbuild/modules/fs.py
```diff
@@ -254,6 +268,47 @@ def read(self, state: 'ModuleState', args: T.Tuple['FileOrString'], kwargs: 'Rea
...
+
+        # The input is allowed to have path separators, but the output may not,
+        # so use the basename for the default case
+        dest = args[1] if args[1] else os.path.basename(src.fname)
+        if has_path_sep(dest):
+            raise InvalidArguments('Destination path may not have path seperators')
```
eli-schwartz, line 292: Shouldn't this be a validator in the KwargInfo?

eli-schwartz: Also "seperators" -> "separators".

dcbaker: It's a positional argument, so no validators. I've thought about adding a `PosArgInfo` class to allow some of this to be handled int `typed_pos_args` because of this exact situation

dcbaker: Actually, what I'd really like to do is replace both `typed_kwargs` and `typed_pos_args` with an `interface` decorator that also handles added/deprecated all in one decorator, but I have enough big projects going on atm...

docs/markdown/Fs-module.md
eli-schwartz: Ping?

eli-schwartz (changes requested):
:crossed_swords: conflicts

eli-schwartz (approved):

PR #10044, opened 2022-03-01
compilers/d: fix mangling of rpath-link in DMD-like compilers

eli-schwartz:
We didn't consider that it has arguments following it, so the resulting compiler command line ended up with stuff like:

```
-L=-rpath-link -L=-L=/path/to/directory -L=more-args
```

and the directory for rpath-link got eaten up as a regular -L path to the compiler rather than being passed as -Xlinker to the linker.

Then the -rpath-link would consume the next -Xlinker argument, end up with the wrong rpath-link (may or may not cause link errors) and then disappear arguments we need.

As an example failure mode, if the next argument is -soname this treats the soname text as an input file, which probably does not exist if it was generated in a subdirectory, and also because it can never be successfully built in the first place -- though if it did, it would link to itself which is very wrong.

/cc @Panke

### DIFF
--- a/mesonbuild/compilers/d.py
+++ b/mesonbuild/compilers/d.py
@@ -286,7 +286,7 @@ class DmdLikeCompilerMixin(CompilerMixinBase):
         # see the comment in the "-L" section
         link_expect_arg = False
         link_flags_with_arg = [
-            '-rpath', '-soname', '-compatibility_version', '-current_version',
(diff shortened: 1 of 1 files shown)


eli-schwartz:
I debugged this in IRC by analyzing compiler commands and noticing an inconsistency, then asking @Panke to test this change locally to see if it works.

I don't actually know anything about D myself! But it seems correct, and it works. Testsuite adaptations to prove that it works, may be contributed by @Panke.

PR #10045, opened 2022-03-01
Fix Dub dependencies

rtbo:
This is nearly a complete rewrite of Dub dependencies that were broken except for the most simple cases.
 
 - fix the research of target built by DUB
 - explicitely state that DUB dynamic libraries and source libraries are not supported (yet) (#6581)
 - fix the build settings of recipes having sub-dependencies (#7560)
 - fix the link to windows library coming from dub
 - sanitization, comments, explanations...

My test case is a simple executable depending on `vibe-d:http`. Vibe-d is complex and has a many dependencies.
I tested with success on Linux and Windows (to figure out the right order of static libraries took some time but is now OK).

The non-compatibility with Dub package having sub-dependencies is especially annoying and is the main motivation for me to write this PR.
My meson file before:
```meson
vibed_deps = [
    dependency('stdx-allocator',
        version: '2.77.5',
        method: 'dub',
    ),
    dependency('eventcore',
        version: '0.9.20',
        method: 'dub',
    ),
    dependency('vibe-core',
        version: '1.22.0',
        method: 'dub',
    ),
    dependency('diet-ng',
        version: '1.8.0',
        method: 'dub',
    ),
    dependency('taggedalgebraic',
        version: '0.11.22',
        method: 'dub',
    ),
    dependency('vibe-d:utils',
        version: '0.9.4',
        method: 'dub',
    ),
    dependency('vibe-d:crypto',
        version: '0.9.4',
        method: 'dub',
    ),
    dependency('vibe-d:stream',
        version: '0.9.4',
        method: 'dub',
    ),
    dependency('vibe-d:textfilter',
        version: '0.9.4',
        method: 'dub',
    ),
    dependency('vibe-d:data',
        version: '0.9.4',
        method: 'dub',
    ),
    dependency('vibe-d:tls',
        version: '0.9.4',
        method: 'dub',
    ),
    dependency('vibe-d:inet',
        version: '0.9.4',
        method: 'dub',
    ),
    dependency('vibe-d:http',
        version: '0.9.4',
        method: 'dub',
    ),
]
if host_machine.system() == 'windows'
    vibed_deps += declare_dependency(
        link_args: [
            join_paths(meson.source_root(), 'libs/windows-x64/libcrypto.lib'),
            join_paths(meson.source_root(), 'libs/windows-x64/libssl.lib'),
            'user32.lib'
        ],
    )
endif
```
My meson file now:
```meson
vibed_deps = [
    dependency('vibe-d:http',
        version: '0.9.4',
        method: 'dub',
    ),
]
```

### DIFF
--- a/mesonbuild/dependencies/dub.py
+++ b/mesonbuild/dependencies/dub.py
@@ -14,15 +14,15 @@
 
 from .base import ExternalDependency, DependencyException, DependencyTypeName
 from .pkgconfig import PkgConfigDependency
-from ..mesonlib import Popen_safe
+from ..mesonlib import (Popen_safe, OptionKey)
+from ..mesonlib.universal import join_args
 from ..programs import ExternalProgram
 from ..compilers import DCompiler
+from ..compilers.d import d_feature_args
 from .. import mlog
 import re
 import os
-import copy
 import json
-import platform
 import typing as T
 
 if T.TYPE_CHECKING:
@@ -34,7 +34,6 @@ class DubDependency(ExternalDependency):
     def __init__(self, name: str, environment: 'Environment', kwargs: T.Dict[str, T.Any]):
         super().__init__(DependencyTypeName('dub'), environment, kwargs, language='d')
         self.name = name
-        self.module_path: T.Optional[str] = None
 
         _temp_comp = super().get_compiler()
         assert isinstance(_temp_comp, DCompiler)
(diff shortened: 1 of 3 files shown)


mesonbuild/dependencies/dub.py
```diff
@@ -59,158 +58,316 @@ def __init__(self, name: str, environment: 'Environment', kwargs: T.Dict[str, T.
...
             self.is_found = False
             return
 
-        comp = self.compiler.get_id().replace('llvm', 'ldc').replace('gcc', 'gdc')
-        packages = []
+        # A command that might be useful in case of missing DUB package
+        def dub_build_deep_command() -> str:
+            return f'dub run dub-build-deep --yes -- "{main_pack_spec}" ' + \
+                f'--arch={arch} --compiler={self.compiler.get_exelist()[-1]} ' + \
+                f'--build={buildtype}'
```
rtbo, line 98: `dub-build-deep` is a dub package I wrote to build the missing dependencies of a package.
https://github.com/rtbo/dub-build-deep

eli-schwartz:
Commits 3 and 4 should probably be rolled as fixups into the first commit.

mesonbuild/dependencies/dub.py
eli-schwartz: I assume that it's impossible to end up with hilarious errors like main_pack_spec having a double quote char, but compiler.get_exelist can theoretically have whitespace, probably? Depends where it is installed?

You can let python do the quoting for both by using `{expr!r}` assuming they are guaranteed strings. Alternatively, define the command as a list, then use `shlex.join()` as the string.

rtbo:
> Commits 3 and 4 should probably be rolled as fixups into the first commit.

I agree but I'm not sure how to do this with commit 2 sitting between them. In general I'd expect that you would squash everything into a single commit before merge.

mesonbuild/dependencies/dub.py
rtbo: none of `{expr!r}' and `shlex.join()` are portable on Windows because the escape with single quote.
I went for the following (adhoc double quote escape):
```py
        # A command that might be useful in case of missing DUB package
        def dub_build_deep_command() -> str:
            assert '"' not in main_pack_spec, 'package name cannot contain double quote'
            return f'dub run dub-build-deep --yes -- "{main_pack_spec}" ' + \
                f'--arch={arch} "--compiler={self.compiler.get_exelist()[-1]}" ' + \
                f'--build={buildtype}'
```
The double quote should work everywhere as long as there is no double quote in the string (which should not be possible).
This command is only shown to the user, so it is not highly critical if it can't run as is.

eli-schwartz: ... right, I forgot about that.

`mesonlib.quote_arg(str)` or `mesonlib.join_args(list)` should work on either Windows or Linux, though. It's used in calculating command lines for build.ninja so it must always be implemented correctly for the current OS or the resulting generated build would be broken.

mesonbuild/dependencies/dub.py
```diff
@@ -59,158 +58,318 @@ def __init__(self, name: str, environment: 'Environment', kwargs: T.Dict[str, T.
         mlog.debug('Determining dependency {!r} with DUB executable '
                    '{!r}'.format(name, self.dubbin.get_path()))
 
+        # if an explicit version spec was stated, use this when querying Dub
+        main_pack_spec = name
+        if 'version' in kwargs:
+            main_pack_spec = f'{name}@{kwargs["version"]}'
```
rtbo, line 64: For the `version` kwarg, currently we can only specify an exact version like `0.9.4`, otherwise Meson complains about version mismatch.
However it would be possible to use whatever DUB normally accepts such as `~>0.9.3`.

dcbaker: I think I'm reading the documentation right that `~>0.9.3` would be equivalent to the meson version expression `['>= 0.9.3', '< 0.10']`?

rtbo:
Each time I bring modification to `ci/ciimage/opensuse/install.sh`, there are weird errors in the Arch and Opensuse test images.
It is systematically the same failure in `vala/14 target glib version and gresources`:

```txt
[1/6] /usr/bin/glib-compile-resources '../test cases/vala/14 target glib version and gresources/gres/test-resources.xml' --sourcedir '../test cases/vala/14 target glib version and gresources/gres/.' --sourcedir '../test cases/vala/14 target glib version and gresources/gres' --internal --generate --target gres/testui.c --dependency-file gres/testui.c.d
FAILED: gres/testui.c 
/usr/bin/glib-compile-resources '../test cases/vala/14 target glib version and gresources/gres/test-resources.xml' --sourcedir '../test cases/vala/14 target glib version and gresources/gres/.' --sourcedir '../test cases/vala/14 target glib version and gresources/gres' --internal --generate --target gres/testui.c --dependency-file gres/testui.c.d
../test cases/vala/14 target glib version and gresources/gres/test-resources.xml: Failed to close file descriptor for child process (Operation not permitted).
[2/6] /usr/bin/glib-compile-resources '../test cases/vala/14 target glib version and gresources/gres/test-resources.xml' --sourcedir '../test cases/vala/14 target glib version and gresources/gres/.' --sourcedir '../test cases/vala/14 target glib version and gresources/gres' --internal --generate --target gres/testui.h
FAILED: gres/testui.h 
/usr/bin/glib-compile-resources '../test cases/vala/14 target glib version and gresources/gres/test-resources.xml' --sourcedir '../test cases/vala/14 target glib version and gresources/gres/.' --sourcedir '../test cases/vala/14 target glib version and gresources/gres' --internal --generate --target gres/testui.h
../test cases/vala/14 target glib version and gresources/gres/test-resources.xml: Failed to close file descriptor for child process (Operation not permitted).
```

Only the modification of `install.sh` triggers the issue. Any idea on this?
I don't know how to make the automated test otherwise.

eli-schwartz:
This happens in any PR that tries to modify the CI images, not just yours. It also causes the weekly cron image rebuilds to fail. This is not your problem, in other words.

The issue has been discovered elsewhere too, particularly https://gitlab.gnome.org/Infrastructure/GitLab/-/issues/545 which analyses it quite a bit. Ultimately I think we're at the mercy of Github's infrastructure, which I'm not sure we can do anything about and which has a timeframe I possess no knowledge of.

On other PRs which did CI updates on images other than the ones which currently fail to rebuild, I merged the CI image update separately, expected a bunch of jobs to fail, and just wanted to get the Ubuntu image to rebuild. See #10070

Unfortunately, that doesn't help if you actually want to update the Arch or OpenSUSE images.

rtbo:
I've changed to only modify the Ubuntu images, but it still leads to failure the Arch and Opensuse image builders.
I believe that "linux / Ubuntu Rolling" are failing due to being based on a previous image, is that correct?

rtbo:
There is a bug in the Windows 32bits environment (`vc2017x86ninja`), but IMO not from my code:

 - At least for the 2 failing tests, Meson is configured to build for `x86_64` host machine.
   - I don't know if it is on purpose to do this cross compilation, but I don't think so.
   - The trace shows that all machines (build, host and target) are `x86_64`. Is it a Meson bug?
  
 - My code looks for dependencies for the host machine.
   - For `d/11 dub`: it fails to find the dependency because only the `x86` version of `urld` is installed
   - For `d/14 dub with deps`: it finds the dependencies because I install the right ones with `run_command` in the `meson.build` script.
  
 - `d/14 dub with deps` fail to link because the compiler generates 32bit code, despite host-machine (and dependencies) being `x86_64`.  (The arch flags are based on `compiler.arch`, not `host_machine.cpu_family()`). This is also a bug (I guess). 

For the time being I can skip the tests on Windows, but my code might break some situations that are working today (maybe shouldn't) on 32bits Windows. (arguablly not a large audience).

rtbo:
I thought `vc2017x86ninja `was 32bit, but it is 64bit Windows with 32bit toolchain.
The current behavior of Meson is to have `x86_64` `host_machine` and `build_machine`, but to generate 32bits binaries.
If this is intended and by design, my current code is correct and reflects this behavior.

Please let me know if anything must be changed.

rtbo:
Anything I can do to help the review?
On my side I think the code is ready for merge.

dcbaker (approved):
LGTM, thanks for persisting with this

Issue #10040, opened 2022-02-28
Wayland module: Generate both client and server sides


xclaesse:
CC @markbolhuis 

Some projects (e.g. Mesa I get told) uses both client and server side at the same time. In that case the .c file generated is identical, but the header is different. Currently trying to generate both cause that error:

```
meson.build:13:0: ERROR: Tried to create target "xdg-shell-protocol", but a target of that name already exists.
```

What about allowing that syntax: `wl_mod.scan_xml(xml, side: ['client', 'server'])` that would return 3 files: one c file and 2 headers?

fooishbar:
Most Wayland compositors also need this.

ppaalanen:
What if your project has some executables that are pure servers, some are pure clients, and some are both?

Hmm... what if `side: ['client', 'server']` was the _only_ option?

The difference is in headers, and the `.c` file is identical to both. You pick which header you use by the `#include` in your program.

xclaesse:
> Hmm... what if `side: ['client', 'server']` was the _only_ option?

To be honest, why then wayland-scanner itself does not gen 3 files in 1 call?

xclaesse:
Oh, and now I also realize we need to document that the header file name is `{name}-{side}-protocol.h` otherwise user cannot know how to `#include` them.

ppaalanen:
> To be honest, why then wayland-scanner itself does not gen 3 files in 1 call?

That's a good question. Historical reasons? Not needing a `.c` file at all would be nice too, somehow.

I have got the feeling that in the early days of Wayland there was a lot of premature optimization going on.

markbolhuis:
I was aware of this but didn't consider it a problem since you can run scan_xml from different subdirs and it works no problem.
Do projects that require both client and server headers typically place client and server code in the same directory and thus the same meson.build file?

Also I don't agree that 3 files should be generated.
Either two boolean flags like so, where either one or the other must be true, or both:
```
gen = wl_mod.scan_xml('protocol.xml', client : true, server : true)
``
as suggested in #9895 

Either that or as @xclaesse suggests.

xclaesse:
> `gen = wl_mod.scan_xml('protocol.xml', client : true, server : true)`

yes, that's even better I think.

fooishbar:
> To be honest, why then wayland-scanner itself does not gen 3 files in 1 call?

[...]

PR #10047, opened 2022-03-01
wayland module: Allow building both client and server sides

xclaesse:
- Change `scope` kwarg to `public` boolean default to false.
- Change `side`  kwarg to `client` and `server` booleans.
- Document returned values
- Aggregate in a single unit test because have lots of small tests
  increases CI time.

Fixes: #10040.


xclaesse:
CC @markbolhuis

markbolhuis:
Do we want to throw an error on both client and server set to false?

xclaesse:
> Do we want to throw an error on both client and server set to false?

That would still generate c file, is there a use case for that? I guess I should error on that, yes.

xclaesse:
> Do we want to throw an error on both client and server set to false?

@markbolhuis Changed to make that an error. Better be more strict, we can always relax later in the unlikely event someone as a use-case I didn't think about.

evelikov-work:
Big fan of the defaults - private + both headers. In particular:

Very few project will need public symbols. For the header side - more commonly projects will need client size only, yet the overhead of the server header is irrelevant in practise.

Small question: how does this module handle dependency of the client/server headers, where source is not needed?
Consider the following example, project X with separate internal "sub-modules", let's consider them static libraries for now:
 - sub-module A: pulls the client header, does not need the .C
 - sub-module B: pulls the server header + .C file
 - final shared library C: A + B + misc other

If I'm reading the docs correctly, to avoid multiple symbols, while ensuring proper dependency tracking one should:
 - A: declare dependency of the client header aka [1], alone
 - B: add the [0] and [2] elements as sources

Is that correct? In addition - is it possible to add symbolic names, instead of a numerical index?

xclaesse:
@evelikov-work That's what GTK does, I made a draft port to use the module there: https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/4541

tl;dr: You can split the return value to keep sources and headers separately:
```
gen = wl_mod.scan_xml(xml)
sources += gen[0]
headers += gen[1]
```

markbolhuis:
I'd like to avoid indexing the list.
What's the point in scanning it all in one go if it's going to be split up?
If we want to keep both headers and code separate, then I suggest the following:
```
headers += wlmod.scan_headers(xmls, client : true, server : true)
src += wlmod.scan_code(xmls, public : false)
```
Thoughts?

xclaesse:
I would e expect most of the time we don't need to split headers and source, is that more common than I thought?

dcbaker:
I think we split them in mesa because we propagate the header, but I’d rather keep it as one call not two

dcbaker (approved):

PR #10048, opened 2022-03-01
Jni dep libraries

### LINKED ISSUE #9601: JDK system dependency: link against libjvm
**Describe the bug**
`jni.h` does not only contain structures to run c(++) code from java, but also java from c(++). For this to work the resulting binary has to be linked against `libjvm.so` located in `$JAVA_HOME/lib/server/`. Otherwise if your code tries to call these, the linker will fail:
```
FAILED: mjvmtest 
cc  -o mjvmtest mjvmtest.p/jvm.c.o -Wl,--as-needed -Wl,--no-undefined
/usr/lib64/gcc/x86_64-suse-linux/11/../../../../x86_64-suse-linux/bin/ld: mjvmtest.p/jvm.c.o: in function `main':
/home/admin/workspace/cli/meson/mjvmtest/_build/../jvm.c:16: undefined reference to `JNI_CreateJavaVM'
```

**To Reproduce**
I've taken the JNI Invokation Example: https://docs.oracle.com/en/java/javase/11/docs/specs/jni/invocation.html, ported it to C and made it actually do something.
```c
#include <jni.h>	/* where everything is defined */
#include <stdbool.h>

int main(int, char**) {
	JavaVM *jvm;	/* denotes a Java VM */
	JNIEnv *env;	/* pointer to native method interface */
	JavaVMInitArgs vm_args;	/* JDK/JRE 10 VM initialization arguments */
	JavaVMOption options[2];
	options[0].optionString = "-Djava.class.path=/usr/lib64/java";
	options[1].optionString = "-Xcheck:jni";
	vm_args.version = JNI_VERSION_10;
	vm_args.nOptions = 1;
	vm_args.options = options;
	vm_args.ignoreUnrecognized = false;
	/* load and initialize a Java VM, return a JNI interface
	* pointer in env */
	JNI_CreateJavaVM(&jvm, (void**)&env, &vm_args);
	/* get stdout PrintStream */
	jclass system = (*env)->FindClass(env, "java/lang/System");
	jfieldID jstdout_id = (*env)->GetStaticFieldID(env, system, "out", "Ljava/io/PrintStream;");
	jobject jstdout = (*env)->GetStaticObjectField(env, system, jstdout_id);
	/* invoke println method of PrintStream */
	jclass print_stream = (*env)->FindClass(env, "java/io/PrintStream");
	jmethodID mid = (*env)->GetMethodID(env, print_stream, "println", "(Ljava/lang/String;)V");
	jstring to_print = (*env)->NewStringUTF(env, "Hello World from C");
	(*env)->CallVoidMethod(env, jstdout, mid, to_print);
	/* We are done. */
	(*jvm)->DestroyJavaVM(jvm);
}
```

```meson
project('mjvmtest', 'c',
	version : '0.1',
	default_options : ['warning_level=3', 'c_std=gnu2x'])

app = executable('mjvmtest', [
		'jvm.c'
	],
	dependencies: [
		#dependency('jdk', version: ['>=11', '<12'])
		dependency('jdk', version: '>=11')
	],
	install : true
)
```

**Expected behavior**
I expect it to link against libjvm.so.

Adding ` self.link_args = [f'{self.java_home}/lib/server/libjvm.so']`to the JDKSystemDependency __init__ method in `mesonbuild/depenencies/dev.py` makes it work.

What needs to be done is to extend it that it uses the correct file extension for each OS and maybe link it in a way that libjvm.so does not have to be in the library search path.

**system parameters**
* OS: openSUSE Tumbleweed 20211117
* Python 3.8.12
* Meson 0.59.4 & cc80187
* Ninja 1.10.2

tristan957:
This allows someone to link against libjvm.so and libjawt.so.

Fixes #9601 

cc @eli-schwartz @sp1ritCS

### DIFF
--- a/mesonbuild/dependencies/dev.py
+++ b/mesonbuild/dependencies/dev.py
@@ -15,6 +15,8 @@
 # This file contains the detection logic for external dependencies useful for
 # development purposes, such as testing, debugging, etc..
 
+from __future__ import annotations
+
 import glob
 import os
 import re
@@ -38,6 +40,13 @@ from .pkgconfig import PkgConfigDependency
 if T.TYPE_CHECKING:
     from ..envconfig import MachineInfo
     from .. environment import Environment
+    from typing_extensions import TypedDict
+
+    class JNISystemDependencyKW(TypedDict):
+        modules: T.List[str]
+        # FIXME: When dependency() moves to typed Kwargs, this should inherit
+        # from its TypedDict type.
+        version: T.Optional[str]
 
 
 def get_shared_library_suffix(environment: 'Environment', for_machine: MachineChoice) -> str:
@@ -500,8 +509,8 @@ class ZlibSystemDependency(SystemDependency):
 
 
 class JNISystemDependency(SystemDependency):
-    def __init__(self, environment: 'Environment', kwargs: T.Dict[str, T.Any]):
-        super().__init__('jni', environment, kwargs)
+    def __init__(self, environment: 'Environment', kwargs: JNISystemDependencyKW):
+        super().__init__('jni', environment, T.cast(T.Dict[str, T.Any], kwargs))
 
         self.feature_since = ('0.62.0', '')
 
@@ -511,6 +520,14 @@ class JNISystemDependency(SystemDependency):
             detect_compiler_for(environment, 'java', self.for_machine)
         self.javac = environment.coredata.compilers[self.for_machine]['java']
         self.version = self.javac.version
+
+        modules: T.List[str] = mesonlib.listify(kwargs.get('modules', []))
+        for module in modules:
+            if module not in {'jvm', 'awt'}:
+                log = mlog.error if self.required else mlog.debug
+                log(f'Unknown JNI module ({module})')
+                self.is_found = False
+                return
 
         if 'version' in kwargs and not version_compare(self.version, kwargs['version']):
             mlog.error(f'Incorrect JDK version found ({self.version}), wanted {kwargs["version"]}')
@@ -530,6 +547,44 @@ class JNISystemDependency(SystemDependency):
         java_home_include = self.java_home / 'include'
         self.compile_args.append(f'-I{java_home_include}')
         self.compile_args.append(f'-I{java_home_include / platform_include_dir}')
+
+        if modules:
+            if m.is_windows():
+                java_home_lib = self.java_home / 'lib'
+                java_home_lib_server = java_home_lib
+            else:
+                if version_compare(self.version, '<= 1.8.0'):
+                    # The JDK and Meson have a disagreement here, so translate it
+                    # over. In the event more translation needs to be done, add to
+                    # following dict.
+                    def cpu_translate(cpu: str) -> str:
+                        java_cpus = {
+                            'x86_64': 'amd64',
+                        }
+
+                        return java_cpus.get(cpu, cpu)
+
+                    java_home_lib = self.java_home / 'jre' / 'lib' / cpu_translate(m.cpu_family)
+                    java_home_lib_server = java_home_lib / "server"
+                else:
+                    java_home_lib = self.java_home / 'lib'
+                    java_home_lib_server = java_home_lib / "server"
+
+            if 'jvm' in modules:
+                jvm = self.clib_compiler.find_library('jvm', environment, extra_dirs=[str(java_home_lib_server)])
+                if jvm is None:
+                    mlog.debug('jvm library not found.')
+                    self.is_found = False
+                else:
+                    self.link_args.extend(jvm)
+            if 'awt' in modules:
+                jawt = self.clib_compiler.find_library('jawt', environment, extra_dirs=[str(java_home_lib)])
+                if jawt is None:
+                    mlog.debug('jawt library not found.')
+                    self.is_found = False
+                else:
+                    self.link_args.extend(jawt)
+
         self.is_found = True
 
     @staticmethod
@@ -553,7 +608,7 @@ class JNISystemDependency(SystemDependency):
 
 
 class JDKSystemDependency(JNISystemDependency):
-    def __init__(self, environment: 'Environment', kwargs: T.Dict[str, T.Any]):
+    def __init__(self, environment: 'Environment', kwargs: JNISystemDependencyKW):
         super().__init__(environment, kwargs)
 
         self.feature_since = ('0.59.0', '')
--- a/docs/markdown/Dependencies.md
+++ b/docs/markdown/Dependencies.md
@@ -494,22 +494,33 @@ Deprecated name for JNI. `dependency('jdk')` instead of `dependency('jni')`.
 
 *(added 0.62.0)*
 
+`modules` is an optional list of strings containing any of `jvm` and `awt`.
+
 Provides access to compiling with the Java Native Interface (JNI). The lookup
 will first check if `JAVA_HOME` is set in the environment, and if not will use
 the resolved path of `javac`. Systems will usually link your preferred JDK to
 well known paths like `/usr/bin/javac` on Linux for instance. Using the path
 from `JAVA_HOME` or the resolved `javac`, this dependency will place the JDK
 installation's `include` directory and its platform-dependent subdirectory on
-the compiler's include path.
+the compiler's include path. If `modules` is non-empty, then the proper linker
+arguments will also be added.
 
 ```meson
-dep = dependency('jni', version: '>= 1.8.0')
+dep = dependency('jni', version: '>= 1.8.0', modules: ['jvm'])
 ```
 
 **Note**: Due to usage of a resolved path, upgrading the JDK may cause the
-various headers to not be found. In that case, please reconfigure the build
+various paths to not be found. In that case, please reconfigure the build
 directory. One workaround is to explicitly set `JAVA_HOME` instead of relying on
 the fallback `javac` resolved path behavior.
+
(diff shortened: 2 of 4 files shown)


tristan957:
Don't review until other PR goes in.

mesonbuild/dependencies/dev.py
```diff
@@ -500,18 +509,26 @@ def __init__(self, name: str, environment: 'Environment', kwargs: T.Dict[str, T.
...
 
         m = self.env.machines[self.for_machine]
+        assert m
 
         if 'java' not in environment.coredata.compilers[self.for_machine]:
             detect_compiler_for(environment, 'java', self.for_machine)
         self.javac = environment.coredata.compilers[self.for_machine]['java']
         self.version = self.javac.version
 
+        modules: T.Set[str] = set(mesonlib.listify(kwargs.get('modules', [])))
```
eli-schwartz, line 525: Why does this need to be converted to a set? It does not seem that you are using any of the properties of a set compared to a list.

mesonbuild/dependencies/dev.py
```diff
@@ -530,6 +547,40 @@ def __init__(self, environment: 'Environment', kwargs: T.Dict[str, T.Any]):
...
+                if 'awt' in modules:
+                    self.link_args.append(f'-ljawt')
+            else:
+                if 'jvm' in modules:
+                    self.link_args.append(f'-L{java_home_lib / "server"}')
+                    self.link_args.append(f'-ljvm')
+
+                if 'awt' in modules:
+                    self.link_args.append(f'-L{java_home_lib}')
+                    self.link_args.append(f'-ljawt')
```
eli-schwartz, line 582: Maybe these should be `extend([f'-L{java_home_lib}', '-ljawt'])` etc.?

Also you do not need f-strings for string literals so please remove those. It's a lint error (but we don't run flake8 in CI yet... one of these days I will get there...)

mesonbuild/dependencies/dev.py
```diff
@@ -530,6 +547,40 @@ def __init__(self, environment: 'Environment', kwargs: T.Dict[str, T.Any]):
...
+                # following dict.
+                def cpu_translate(cpu: str) -> str:
+                    java_cpus = {
+                        'x86_64': 'amd64'
+                    }
+
+                    return java_cpus.get(cpu, cpu)
+
+                if len(modules) > 0:
+                    self.link_args.append(f'-L{java_home_lib / "jre" / cpu_translate(m.cpu_family)}')
```
eli-schwartz, line 568: This is kind of a really complicated expression to put in an f-string. :(

Maybe cpu_translate could be redesigned to calculate the correct directory and 

```python
return java_home_lib / 'jre' / java_cpus.get(cpu, cpu)
```

eli-schwartz (changes requested):

mesonbuild/dependencies/dev.py
tristan957: Mistake! Extend API sounds good.

mesonbuild/dependencies/dev.py
tristan957: Wanted to deduplicate in the case someone passed `modules: ['jvm', 'jvm']`

mesonbuild/dependencies/dev.py
tristan957: I'll  just create a local variable then.

mesonbuild/dependencies/dev.py
eli-schwartz: But you only ever check `if 'jvm' in modules` and so on, so that doesn't care if it appears multiple times.

tristan957: You win :)

mesonbuild/dependencies/dev.py
```diff
@@ -530,6 +547,38 @@ def __init__(self, environment: 'Environment', kwargs: T.Dict[str, T.Any]):
...
+                    java_cpus = {
+                        'x86_64': 'amd64'
+                    }
+
+                    return java_cpus.get(cpu, cpu)
+
+                translated_cpu = cpu_translate(m.cpu_family)
+
+                if 'jvm' in modules:
+                    java_home_jre_lib_cpu_server = self.java_home / "jre" / "lib" / translated_cpu / 'server'
```
eli-schwartz, line 569: In the previous version, `'server'` was only part of the JDK 9 `else` clause. Was this an intentional fix compared to the previous patch revision?

tristan957: Yes. This fixed the CI failure which I also noticed on my local machine to confirm after switching javac to 1.8.0

tristan957: If you have a Java 8 on your system, it is easy to confirm. I keep `latest`, `11`, and `1.8` installed on my system.

`find /usr/lib/jvm/` -type f -name "libjvm.so"`. Switch `jvm` from `jawt` to view the other library.

tristan957: I have 11 as my default, so I didn't notice the 1.8 failure initially although I obviously planned for it :)

eli-schwartz: Ah, neat. So what this means then is that we can simplify a lot of this. All we need to do is define `java_home_lib` depending on the java version, e.g. for < 1.8.0 we define `java_home_lib = self.java_home / "jre" / "lib" / cpu_translate(m.cpu_family)`.

Then there is no version difference for the rest, so we leave the version checking if/else and, for all versions of java, extend the link args with:

```python
if 'jvm' in modules:
    self.link_args.extend([f'-L{java_home_lib / "server"}', '-ljvm'])
if 'awt' in modules:
    self.link_args.extend([f'-L{java_home_lib}', '-ljawt'])
```

tristan957: Good point. Let me go fix this up.

tristan957:
Applied the changes suggested in previous comment. All that seems to be left is the CI.

eli-schwartz:
```
[6/6] "lld-link"  /MACHINE:x64 /OUT:lib/jnijava.dll lib/jnijava.dll.p/native.c.obj lib/jnijava.dll.p/com_mesonbuild_JniTest.c.obj "/nologo" "/DEBUG" "/PDB:lib\jnijava.pdb" "/DLL" "/IMPLIB:lib\jnijava.lib" "/LIBPATH:C:\hostedtoolcache\windows\Java_Temurin-Hotspot_jdk\8.0.322-6\x64\jre\lib\amd64\server" "jvm.lib" "/LIBPATH:C:\hostedtoolcache\windows\Java_Temurin-Hotspot_jdk\8.0.322-6\x64\jre\lib\amd64" "jawt.lib" "kernel32.lib" "user32.lib" "gdi32.lib" "winspool.lib" "shell32.lib" "ole32.lib" "oleaut32.lib" "uuid.lib" "comdlg32.lib" "advapi32.lib"
FAILED: lib/jnijava.dll lib/jnijava.pdb 
"lld-link"  /MACHINE:x64 /OUT:lib/jnijava.dll lib/jnijava.dll.p/native.c.obj lib/jnijava.dll.p/com_mesonbuild_JniTest.c.obj "/nologo" "/DEBUG" "/PDB:lib\jnijava.pdb" "/DLL" "/IMPLIB:lib\jnijava.lib" "/LIBPATH:C:\hostedtoolcache\windows\Java_Temurin-Hotspot_jdk\8.0.322-6\x64\jre\lib\amd64\server" "jvm.lib" "/LIBPATH:C:\hostedtoolcache\windows\Java_Temurin-Hotspot_jdk\8.0.322-6\x64\jre\lib\amd64" "jawt.lib" "kernel32.lib" "user32.lib" "gdi32.lib" "winspool.lib" "shell32.lib" "ole32.lib" "oleaut32.lib" "uuid.lib" "comdlg32.lib" "advapi32.lib"
lld-link: error: could not open 'jvm.lib': no such file or directory
lld-link: error: could not open 'jawt.lib': no such file or directory
```

This looks vaguely relevant? https://bugs.openjdk.java.net/browse/JDK-8183166

tristan957:
Seems like the Windows builds are actually sane and distribute the lib in `$JAVA_HOME/lib`. I will update appropriately.

eli-schwartz:
> Seems like the Windows builds are actually sane

That sounds tragic!

mesonbuild/dependencies/dev.py
```diff
@@ -530,6 +546,34 @@ def __init__(self, environment: 'Environment', kwargs: T.Dict[str, T.Any]):
...
+                    java_home_lib = self.java_home / 'jre' / 'lib' / cpu_translate(m.cpu_family)
+                    java_home_lib_server = java_home_lib / "server"
+                else:
+                    java_home_lib = self.java_home / 'lib'
+                    java_home_lib_server = java_home_lib / "server"
+
+            if 'jvm' in modules:
+                self.link_args.extend([f'-L{java_home_lib_server}', '-ljvm'])
+            if 'awt' in modules:
+                self.link_args.extend([f'-L{java_home_lib}', '-ljawt'])
```
eli-schwartz, line 575: I just realized these should use `self.clib_compiler.find_library()`, not just because it "consistently" duplicates paths (by using filenames) but also because as we've seen, things might not necessarily be installed e.g. on Ubuntu package splitting.

So using find_library means we can actually do a configure check that these libraries exist, rather than failing at build time.

eli-schwartz (changes requested):

eli-schwartz (approved):

PR #10049, opened 2022-03-01
document and raise an error for disallowed combination of install_headers args

### LINKED ISSUE #10046: install_headers ignores 'subdir:' if 'install_dir:' is set
**Describe the bug**
The install_headers function provides both `install_dir` and `subdir` parameters and documents that the `subdir` parameter is added to any `install_dir` parameter. However, the code ignores `subdir` when `install_dir` is specified.

**To Reproduce**
[install-header.zip](https://github.com/mesonbuild/meson/files/8165005/install-header.zip)

**Expected behavior**
The header file should be installed in install_dir / subdir

**system parameters**
* plain native build
* Debian unstable
* Python 3.9.10
* meson version 0.61.2
* ninja version 1.10.1

**Suggested Fix**
Because existing meson will ignore subdir, I suggest that the documentation be changed to indicate that you cannot use both options at the same time and then make the code raise an error if they are both specified.

eli-schwartz:
It makes no sense to specify both:
- `install_dir`, which overrides the `-Dincludedir=` builtin option
- `subdir`, which suffixes the `-Dincludedir=` builtin option

We've always silently ignored the `subdir` in this case, which is really surprising if someone actually passed it and expected it to do something. We also confusingly didn't say anything in the documentation about it.

Document that the options are incompatible, and explicitly check to see if they are both passed -- if so, raise an error message pointing out that only `install_dir` should be used.

Fixes #10046

### DIFF
--- a/mesonbuild/interpreter/interpreter.py
+++ b/mesonbuild/interpreter/interpreter.py
@@ -2004,8 +2004,11 @@ external dependencies (including libraries) must go to "dependencies".''')
                              kwargs: 'kwargs.FuncInstallHeaders') -> build.Headers:
         source_files = self.source_strings_to_files(args[0])
         install_subdir = kwargs['subdir']
-        if install_subdir is not None and os.path.isabs(install_subdir):
-            mlog.deprecation('Subdir keyword must not be an absolute path. This will be a hard error in the next release.')
+        if install_subdir is not None:
+            if kwargs['install_dir'] is not None:
+                raise InterpreterException('install_headers: cannot specify both "install_dir" and "subdir". Use only "install_dir".')
+            if os.path.isabs(install_subdir):
+                mlog.deprecation('Subdir keyword must not be an absolute path. This will be a hard error in the next release.')
 
         h = build.Headers(source_files, install_subdir, kwargs['install_dir'],
                           kwargs['install_mode'], self.subproject)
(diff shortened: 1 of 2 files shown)


mesonbuild/interpreter/interpreter.py
```diff
@@ -2004,8 +2004,11 @@ def func_install_headers(self, node: mparser.BaseNode,
                              kwargs: 'kwargs.FuncInstallHeaders') -> build.Headers:
         source_files = self.source_strings_to_files(args[0])
         install_subdir = kwargs['subdir']
-        if install_subdir is not None and os.path.isabs(install_subdir):
-            mlog.deprecation('Subdir keyword must not be an absolute path. This will be a hard error in the next release.')
+        if install_subdir is not None:
+            if kwargs['install_dir'] is not None:
+                raise InterpreterException('install_headers: cannot specify both "install_dir" and "subdir". Use only "install_dir".')
+            if os.path.isabs(install_subdir):
+                mlog.deprecation('Subdir keyword must not be an absolute path. This will be a hard error in the next release.')
```
xclaesse, line 2011: Should it recommend using install_dir in that case?

eli-schwartz: The isabs deprecation I haven't touched, only changed the indentation. For the code I added, I do already say:
> Use only "install_dir".

xclaesse:
I think you're right, and that's consistent with python.install_sources() that does the same iirc.

mesonbuild/interpreter/interpreter.py
xclaesse: Ok

xclaesse:
Lgtm

PR #10056, opened 2022-03-02
mpi: fix usage of intel-mpi installed via spack

dguibert:
See https://github.com/NorESMhub/BLOM/issues/145 for details

### DIFF
--- a/mesonbuild/dependencies/mpi.py
+++ b/mesonbuild/dependencies/mpi.py
@@ -65,11 +65,11 @@ def mpi_factory(env: 'Environment',
                 nwargs['returncode_value'] = 3
 
             if language == 'c':
-                tool_names = [os.environ.get('I_MPI_CC'), 'mpiicc']
+                tool_names = [os.environ.get('MPICC'), 'mpiicc']
             elif language == 'cpp':
-                tool_names = [os.environ.get('I_MPI_CXX'), 'mpiicpc']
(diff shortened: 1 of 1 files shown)


mesonbuild/dependencies/mpi.py
```diff
@@ -157,6 +157,10 @@ class IntelMPIConfigToolDependency(_MPIConfigToolDependency):
     """Wrapper around Intel's mpiicc and friends."""
 
     version_arg = '-v'  # --version is not the same as -v
+    # ifort.cfg with: -Xlinker -rpath=... breaks -v (as ifort -v return 1 as exitcode)
+    os.environ["ICCCFG"] = "/dev/null"
```
tristan957, line 161: Wouldn't this not work on Windows for instance?

dcbaker: the ConfigToolDependency class has some code for handling for tools that return non 0 on success, which... it looks like we actually do but set it to 3?

dcbaker:
@scivision, would really appreciate it if you could have a look at this.

mesonbuild/dependencies/mpi.py
eli-schwartz: We should probably not be setting global environment variables, but perhaps teaching the ConfigToolDependency base class to recognize a class attribute dict that can be passed into tool invocations as a custom Popen environment.

PR #10058, opened 2022-03-02
add GitHub URL for PyPi

andriyor:
Warehouse now uses the project_urls provided to display links in the sidebar on [this screen](https://pypi.org/project/requests/), as well as including them in API responses to help the automation tool find the source code for Requests.

Example: [pypi.org/pypi/requests/json](https://pypi.org/pypi/requests/json)
Docs: [packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -5,6 +5,8 @@ description = A high performance build system
 author = Jussi Pakkanen
 author_email = jpakkane@gmail.com
 url = https://mesonbuild.com
+project_urls =
+    Source = https://github.com/mesonbuild/meson
 keywords =
   meson
   mesonbuild


setup.cfg
```diff
@@ -2,9 +2,13 @@
 name = meson
 version = attr: mesonbuild.coredata.version
 description = A high performance build system
+long_description = file: README.md
+long_description_content_type = text/markdown
```
eli-schwartz, line 6: This is not mentioned in the PR description or the commit message. Also, we have an existing long_description, and with your PR we have two of them and setuptools emits a hard error, so this could never build.

Please remove the long_description change.

eli-schwartz (changes requested):
I don't mind adding the repository source url directly to PyPI -- even though the existing link is to the Meson homepage, which in turn directly links to the repository source url.

However, there are other changes here and those other changes cause the CI to fail.

eli-schwartz:
Could you please squash the commits together into one? :)

andriyor:
I think you also can do this just right on GitHub
https://github.blog/2016-04-01-squash-your-commits/
https://docs.github.com/en/enterprise-server@3.1/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-squashing-for-pull-requests

eli-schwartz:
Unfortunately, github makes a complete disaster of a wreck of a commit message and committer name, and I will never trust Github to get it right.

...

Unfortunately, instead of squashing we now seem to have gone from two commits to three commits plus a merge commit.

PR #10060, opened 2022-03-02
Add support for rust proc-macro crates

dcbaker:
Proc macros are the only type of rust crate that we don't support, and support is rather trivial. Plus, I've had a request to add support for them.

### DIFF
--- a/docs/markdown/snippets/rust_proc_macro_crates.md
+++ b/docs/markdown/snippets/rust_proc_macro_crates.md
@@ -1,0 +1,16 @@
+## Rust proc-macro crates
+
+Rust has these handy things called proc-macro crates, which are a bit like a
+compiler plugin. We can now support them, simply build a [[shared_library]] with
+the `rust_crate_type` set to `proc-macro`.
+
+```meson
+proc = shared_library(
+  'proc',
+  'proc.rs',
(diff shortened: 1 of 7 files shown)


jpakkane:
The reference manual does not seem to have `rust_crate_type` documented.

dcbaker:
@jpakkane added

docs/markdown/snippets/rust_proc_macro_crates.md
```diff
@@ -0,0 +1,16 @@
+## Rust proc-macro crates
+
+Rust has these handy things called proc-macro crates, which are a bit like a
+compiler plugin. We can now support them, simply build a *shared_library* with
```
eli-schwartz, line 4: We should be able to use `[[shared_library]]` here too.

dcbaker: Can we? I thought that only worked from yank?

eli-schwartz: No, it's not handled by the refman generator. It's handled by:
```meson
    extra_extension: meson.current_source_dir() / 'extensions' / 'refman_links.py',
```

eli-schwartz (approved):

PR #10061, opened 2022-03-02
add D features to declare_dependency

rtbo:
The goal is to allow such construct:
```meson
my_dep = declare_dependency(
    d_module_versions: ['MyVersion'],
    d_import_dirs: include_directories('...'),
)
```


mesonbuild/build.py
```diff
@@ -1301,6 +1301,16 @@ def add_deps(self, deps):
...
                 continue
+
+            dep_d_features = dep.get_d_features()
+
+            for feat in ('versions', 'import_dirs'):
+                if feat in dep_d_features:
+                    if feat in self.d_features:
+                        self.d_features[feat].extend(dep_d_features[feat])
+                    else:
+                        self.d_features[feat] = dep_d_features[feat]
```
dcbaker, line 1312: We can simplify this if we use a `collections.defaultdict` here:
```suggestion
            for feat in ('versions', 'import_dirs'):
                if feat in dep_d_features:
                    self.d_features[feat].extend(dep_d_features[feat])
```

mesonbuild/dependencies/base.py
```diff
@@ -93,6 +93,7 @@ def __init__(self, type_name: DependencyTypeName, kwargs: T.Dict[str, T.Any]) ->
         self.sources: T.List[T.Union['FileOrString', 'CustomTarget']] = []
         self.include_type = self._process_include_type_kw(kwargs)
         self.ext_deps: T.List[Dependency] = []
+        self.d_features: T.Dict[str, T.Union[T.List[str], T.List['IncludeDirs']]] = {}
```
dcbaker, line 96:
```suggestion
        self.d_features: T.DefaultDict[str, T.Union[T.List[str], T.List['IncludeDirs']]] = collections.defaultdict(list)
```

mesonbuild/dependencies/base.py
```diff
@@ -221,13 +222,17 @@ def generate_system_dependency(self, include_type: str) -> 'Dependency':
         new_dep.include_type = self._process_include_type_kw({'include_type': include_type})
         return new_dep
 
+    def get_d_features(self) -> T.Dict[str, T.Union[T.List[str], T.List['IncludeDirs']]]:
+        return self.d_features
+
```
dcbaker, line 227: This isn't particularly useful, it returns a reference, but it creates the overhead of adding a function call. We can drop it and just use `self.d_features` directly

mesonbuild/interpreter/interpreter.py
```diff
@@ -2404,8 +2407,8 @@ def func_configure_file(self, node: mparser.BaseNode, args: T.List[TYPE_var],
                                               install_tag=install_tag, data_type='configure'))
         return mesonlib.File.from_built_file(self.subdir, output)
 
-    def extract_incdirs(self, kwargs):
-        prospectives = extract_as_list(kwargs, 'include_directories')
+    def extract_incdirs(self, kwargs, key = 'include_directories'):
```
dcbaker, line 2410:
```suggestion
    def extract_incdirs(self, kwargs, key: str = 'include_directories'):
```
 
Let's just add the type annotation while we're here please :)

dcbaker (changes requested):
Looks reasonable, I have a few implementation nits though.

docs/markdown/D.md
```diff
@@ -35,6 +35,23 @@ project('myapp', 'd')
 executable('myapp', 'app.d', d_debug: [3, 'DebugFeatureA'])
 ```
 
+## In `declare_dependency`
+
+When declaring your own dependency using `declare_dependency`, it is
+possible to add parameters for D specific features, e.g. to propagate
+conditional compilation versions:
```
dcbaker, line 42: Probably should note `*since 0.62.0*`

dcbaker:
This also deserves a release snippet in docs/markdown/snippets/

mesonbuild/build.py
eli-schwartz, line 1312: Offtopic: I have an irrational dislike of the token `feat`, because it makes me think this: https://www.dictionary.com/browse/feat

> a noteworthy or extraordinary act or achievement, usually displaying boldness, skill, etc.

I wonder if maybe we could use `for feature in ('versions', 'import_dirs'):` instead?

rtbo: sure :smile: I didn't know this word!

test cases/d/11 dub/.gitignore
```diff
@@ -0,0 +1 @@
+dub.json
```
rtbo, line 1: this one tend to annoy me (file written by the test)

dcbaker (approved):

PR #10063, opened 2022-03-03
install: Add --strip option

xclaesse:


eli-schwartz:
Maybe we should deprecate this as a configure time option?

dcbaker:
Can I be the annoying one? Can we keep the strip option, but just have it set the —strip flag for the ninja install target? That’s very close to 100% backwards compatible, and provides a way for people who don’t want to invoke Meson install directly to continue getting the strip behavior

xclaesse:
I want to keep it 100% backwards compatible, just deprecated. But maybe I should revert this pr to initial version where I don't deprecated if we are in doubt?

xclaesse:
Ok, misunderstood what you meant. We could add --strip to the ninja run target, instead of storing that info for each TargetInstallData, but if we do that we lose per-target strip value: `executable(..., override_options: 'strip=true')`. I'm not sure that's used in practice, but I'm a bit afraid of changing that, I could imagine that when you build a proprietary target you could want to force that target to be always stripped, for example.

dcbaker:
Okay, I'd rather not deprecate `-Dstrip` then, since if we deprecate it then we also need to remove the `ninja install` target, since it can't be configured usefully (ie, you can never strip)

eli-schwartz:
We need to what now?????

dcbaker:
If we deprecate the -Dstrip option, then `ninja install` cannot strip installed binaries, at which point it's useless and we should remove it

xclaesse:
> If we deprecate the -Dstrip option, then `ninja install` cannot strip installed binaries, at which point it's useless and we should remove it

Not remove it, but deprecate it too. I reverted this PR to not deprecate strip option, we can always do that later.

dcbaker:
Deprecation is just a warning something is going to be removed in the future, right?

PR #10064, opened 2022-03-03
WIP: Add meson install --dbg

xclaesse:
It split debug symbols into a separate file. This is similar to Debian -dbg packages.


xclaesse:
@jpakkane I have no idea how that works on Windows to be honest. Are debug files splitted by compiler, or extracted at install like on Linux?

Personally my use-case for this is a project we basically pack the install destdir into a tarball and extract it on devices. By default all binaries are stripped, but I would like to easily create a dbg tarball we could extract on top for developers, without having to ship all new unstripped binaries.

PR #10070, opened 2022-03-03
CI images: add the JRE to Ubuntu

eli-schwartz:
This is needed in order to test a pending improvement to the jni dependency in #10048.

/cc @tristan957

The purpose of this PR is to rebuild the images with the needed change. Some images are expected to fail to rebuild due to preexisting issues and we can ignore that -- they fail the weekly scheduled rebuilds too.

This lets the CI pass on Ubuntu for the other PR, because Ubuntu *does* rebuild once this is merged to master.

### DIFF
--- a/ci/ciimage/bionic/install.sh
+++ b/ci/ciimage/bionic/install.sh
@@ -27,6 +27,7 @@ pkgs=(
   libblocksruntime-dev
   libperl-dev libscalapack-mpi-dev libncurses-dev
   itstool
+  openjdk-11-jre
 )
 
 boost_pkgs=(atomic chrono date-time filesystem log regex serialization system test thread)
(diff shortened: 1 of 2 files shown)


tristan957:
Thanks for pulling this out.

jpakkane:
Should this use the `-jdk` package instead of the `-jre` one? The Java compiler and/or headers are in the former package if memory serves.

eli-schwartz:
Alright, the changes extracted from that other PR have passed:
- CI image builder / Ubuntu Bionic
- CI image builder / Ubuntu Rolling

Which are the two affected images. The entire workflow does run, and fails on two images which aren't being modified here.

I'm calling this a success.

PR #10071, opened 2022-03-03
Convert java module to new API

tristan957:


mesonbuild/modules/java.py
```diff
@@ -17,60 +17,61 @@
...
+        super().__init__()
         self.methods.update({
             'generate_native_header': self.generate_native_header,
             'generate_native_headers': self.generate_native_headers,
         })
 
-        if 'java' not in interpreter.environment.coredata.compilers[MachineChoice.BUILD]:
-            detect_compiler_for(interpreter.environment, 'java', MachineChoice.BUILD)
-        self.javac = interpreter.environment.coredata.compilers[MachineChoice.BUILD]['java']
+    def __get_java_compiler(self, state: ModuleState) -> Compiler:
```
tristan957, line 35: Is there some helper here that I am no aware of?

mesonbuild/modules/java.py
```diff
@@ -17,60 +17,64 @@
...
-from . import ExtensionModule, ModuleReturnValue, ModuleState
-from ..interpreter import Interpreter
+from . import NewExtensionModule, ModuleReturnValue, ModuleState
 
-class JavaModule(ExtensionModule):
+class JavaModule(NewExtensionModule):
     @FeatureNew('Java Module', '0.60.0')
-    def __init__(self, interpreter: Interpreter):
-        super().__init__(interpreter)
+    def __init__(self, *args: T.Any, **kwargs: T.Any):
```
dcbaker, line 28: just `def __init__(self) -> None:` right? (need the None to tell mypy to type check the function

mesonbuild/modules/java.py
```diff
@@ -17,60 +17,64 @@
...
+        if 'java' not in state.environment.coredata.compilers[MachineChoice.BUILD]:
+            detect_compiler_for(state.environment, 'java', MachineChoice.BUILD)
+        return state.environment.coredata.compilers[MachineChoice.BUILD]['java']
 
     @FeatureDeprecated('java.generate_native_header', '0.62.0', 'Use java.generate_native_headers instead')
-    @typed_pos_args('generate_native_header', (str, FileHolder))
+    @typed_pos_args('generate_native_header', (str, FileHolder), min_varargs=1, max_varargs=1)
     @typed_kwargs('java.generate_native_header', KwargInfo('package', str, default=None))
     def generate_native_header(self, state: ModuleState, args: T.Tuple[T.Union[str, FileHolder]],
                                kwargs: T.Dict[str, T.Optional[str]]) -> ModuleReturnValue:
```
dcbaker, line 44: There are two things that look funny here, the varargs (I don't see any varargs, just one positional argument. and the `str | FileHolder`. I think that should be `str | File`, the holder should already be removed by that point.

dcbaker (changes requested):

mesonbuild/modules/java.py
tristan957: Oh shit.... Thanks for your review.

mesonbuild/modules/java.py
tristan957: I believe we still receive the interpreter here. Otherwise Python would error out, right?

dcbaker: We shouldn't be using `T.Any` if we can avoid it, its a step in the wrong direction either way. Either leave it as-is, or remove the interpreter argument if we don't need it.

dcbaker (approved):

PR #10072, opened 2022-03-03
typing annotation refactor and avoid runtime imports

eli-schwartz:
- Fix a few inconsistencies in TYPE_CHECKING blocks
- add a couple future annotations
- move imports into TYPE_CHECKING where appropriate

### DIFF
--- a/mesonbuild/ast/interpreter.py
+++ b/mesonbuild/ast/interpreter.py
@@ -31,7 +31,6 @@ from ..interpreterbase import (
 )
 
 from ..interpreter import (
-    Interpreter,
     StringHolder,
     BooleanHolder,
     IntegerHolder,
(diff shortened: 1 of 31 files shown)


mesonbuild/ast/interpreter.py
```diff
@@ -378,15 +378,15 @@ def quick_resolve(n: BaseNode, loop_detect: T.Optional[T.List[str]] = None) -> T
             mkwargs = {} # type: T.Dict[str, TYPE_nvar]
             try:
                 if isinstance(src, str):
-                    result = StringHolder(src, T.cast(Interpreter, self)).method_call(node.name, margs, mkwargs)
+                    result = StringHolder(src, T.cast('Interpreter', self)).method_call(node.name, margs, mkwargs)
```
tristan957, line 381: I think you can save the hassle here and add `from __future__ import annotations`.

eli-schwartz: The commit message for this individual change says:
> This is not possible to do with T.cast though, because it is a function argument instead of an annotation.

tristan957: Whoopssss

eli-schwartz:
Rebased to fix the merge conflict.

PR #10073, opened 2022-03-04
Take into account coredata-based args for external projects

tristan957:
Previously when setting things like "warning_level=0", external projects
wouldn't know what that meant, so parent projects would have to
translate it appropriately.

By looking at all coredata-based args, we can properly forward certain
options into the external project.

### DIFF
--- a/mesonbuild/modules/unstable_external_project.py
+++ b/mesonbuild/modules/unstable_external_project.py
@@ -140,13 +140,30 @@ class ExternalProject(NewExtensionModule):
                 continue
             cargs = self.env.coredata.get_external_args(MachineChoice.HOST, lang)
             assert isinstance(cargs, list), 'for mypy'
+
+            cargs += compiler.get_always_args()
+            cargs += compiler.get_warn_args(T.cast(str, self.env.coredata.get_option(
+                OptionKey('warning_level', state.subproject))))
+            cargs += compiler.get_optimization_args(T.cast(
+                str, self.env.coredata.get_option(OptionKey('optimization', state.subproject))))
+            if T.cast(bool, self.env.coredata.get_option(OptionKey('werror', state.subproject))):
+                cargs += compiler.get_werror_args()
+            cargs += compiler.get_debug_args(T.cast(bool, self.env.coredata.get_option(
+                OptionKey('debug', state.subproject))))
+            cargs += compiler.get_buildtype_args(T.cast(str,
+                self.env.coredata.get_option(OptionKey('buildtype', state.subproject))))
+
             self.run_env[ENV_VAR_PROG_MAP[lang]] = self._quote_and_join(compiler.get_exelist())
             self.run_env[CFLAGS_MAPPING[lang]] = self._quote_and_join(cargs)
             if not link_exelist:
                 link_exelist = compiler.get_linker_exelist()
                 _l = self.env.coredata.get_external_link_args(MachineChoice.HOST, lang)
                 assert isinstance(_l, list), 'for mypy'
+
                 link_args = _l
+                link_args += compiler.get_linker_always_args()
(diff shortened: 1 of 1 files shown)


tristan957:
Are there any compile/link args that I am missing?

xclaesse:
> Are there any compile/link args that I am missing?

I'm sure there are... but non that I can think of now :P

> What do you think is a good way to test this?

You can check the value of CFLAGS in `test cases/common/230 external project/libfoo/configure`.

mesonbuild/modules/unstable_external_project.py
```diff
@@ -140,13 +140,30 @@ def _configure(self, state: 'ModuleState') -> None:
...
+            cargs += compiler.get_warn_args(T.cast(str, self.env.coredata.get_option(
+                OptionKey('warning_level', state.subproject))))
+            cargs += compiler.get_optimization_args(T.cast(
+                str, self.env.coredata.get_option(OptionKey('optimization', state.subproject))))
+            if T.cast(bool, self.env.coredata.get_option(OptionKey('werror', state.subproject))):
+                cargs += compiler.get_werror_args()
+            cargs += compiler.get_debug_args(T.cast(bool, self.env.coredata.get_option(
+                OptionKey('debug', state.subproject))))
+            cargs += compiler.get_buildtype_args(T.cast(str,
+                self.env.coredata.get_option(OptionKey('buildtype', state.subproject))))
```
eli-schwartz, line 154: The ordering seems wrong for this. While many of these are going to be non-conflicting groups of arguments, the external args should come last in order to override anything else.

See `generate_basic_compiler_args()` in mesonbuild/backend/backends.py for the logical ordering we use there. It may also be useful to refactor some of that logic into something shareable between backends and unstable_external_project.

eli-schwartz (changes requested):

mesonbuild/modules/unstable_external_project.py
tristan957: Thanks for the feedback. I'll look into this tonight. Hopefully we can get it in in the upcoming release.

tristan957: So far I am thinking that the Backend class is doing too much. I think generate_basic_compiler_args() should essentially be moved to the Compiler class.

```python
def get_basic_compiler_args(self,
            env: T.Optional[Environment]=None,
            target: T.Optional[Target]=None,
            subproject: T.Optional[SubProject]=None) -> T.List[str]:
    ...
```
Requires mirroring get_option_for_target() in the Compiler class. Any thoughts on this potential approach?

Generally it feels like the Backend class is doing too much to get compiler args and should just be able to ask the Compiler class for most of it. get_source_dir_include_args() and get_build_dir_include_args() for instance.

tristan957: Maybe get_option_for_target() could be moved to the Environment class?

eli-schwartz: @dcbaker thoughts?

xclaesse:
@tristan957 I thought I commented on this PR already, but seems I forgot. I remember why I haven't done this initially, most of the time those warning/debug extra args are handled with args like `--enable-debug`. The idea is you're supposed to map Meson options to configure options yourself.

xclaesse:
- compiler.get_always_args() - The external build system is supposed to have them already.
- compiler.get_warn_args() - This one could make sense because I don't think it is typically handled by other build system.
- compiler.get_optimization_args() - Same, that one is probably fine.
- compiler.get_debug_args() - Probably better add a method in the module to translate `get_option('debug')` to `--enable-debug`.
- compiler.get_buildtype_args() - Why do we even have that in Meson itself, it is supposed to be a combination of the 2 above.

PR #10076, opened 2022-03-06
lots of fixes to i18n script tool handling

### LINKED ISSUE #6165: i18n.merge_file results in a confusing error with gettext missing
Where gettext is missing, use of `i18n.merge_file` results in
```
ERROR: Tried to assign an invalid value to variable.
```
This is a bit confusing, doubly so because it doesn't have a reference to why this is happening.

Turns out there is a "gettext is missing" warning but that warning may be a few miles off screen by the time the error is hit. I didn't actually notice the warning until I wrote the reduced meson.build for the bug report.

Minimal reproducer in a container:
```
$ podman run --rm -i -t  fedora:latest /bin/bash
$ dnf install -y --quiet meson
$ mkdir test && cd test
$ echo > meson.build <<EOF
project('test')

i18n = import('i18n')
foo = i18n.merge_file(input: 'input.in', output: 'output', type: 'desktop',
		      po_dir: 'po', install: true)
EOF
$ meson build
[...]
WARNING: Gettext not found, all translation targets will be ignored.
[...]
meson.build:4:0: ERROR: Tried to assign an invalid value to variable.
```
It'd be useful to have the error complain that the target was disabled or something.

### LINKED ISSUE #8436: i18n module only checks for gettext and misses xgettext
**Describe the bug**
We have a configure file in our build that uses `i18n.merge_file`.  In the users log we saw

```
Program gettext found: YES (/usr/bin/gettext)
```

But when Meson reached the `i18n.merge_file` this error was spit out.

```
WARNING: Gettext not found, all translation targets will be ignored.
```

**To Reproduce**

It took some time to figure out, but the user is on Debian stable, but the issue was that they had `gettext-base` installed that contains `/usr/bin/gettext` but `xgettext` that `i18n.merge_file` requires is actually in the `gettext` package.

**Expected behavior**

It would have been awesome if when detecting the gettext binary, meson would check for xgettext as well if `i18n.merge_file` is in use.  If it couldn't find it, it would be much nicer if it would mention the executable it was looking for as I had to dig through the meson source to figure out what it was complaining about as the user already verified that they had `/usr/bin/gettext` installed.

**system parameters**

The user was on Debian stable with meson 0.56.2, but their system isn't relevant here out side of the `gettext-base` was installed but `gettext` was not installed scenario.

eli-schwartz:
By the way, argparse sucks and I hate it.

Some things this fixes:
- argparse sucks and mishandles arguments, but only sometimes
- check for tools which we need at configure time and respect machine file overrides, rather than using `shutil.which('xgettext')` as a universal sentinel. Log what we found, because state.find_program has decent logging and shutil.which does not because shutil.which is a low level python function and not something that belongs in the high level parts of the Meson interpreter
- Stop disabling really important, downright crucial parts of peoples' builds.

Fixes #6165
Fixes #8436

### DIFF
--- a/mesonbuild/modules/i18n.py
+++ b/mesonbuild/modules/i18n.py
@@ -167,10 +167,7 @@ class I18nModule(ExtensionModule):
     )
     def merge_file(self, state: 'ModuleState', args: T.List['TYPE_var'], kwargs: 'MergeFile') -> ModuleReturnValue:
         if self.tools['msgfmt'] is None:
-            self.tools['msgfmt'] = state.find_program('msgfmt', required=False, for_machine=mesonlib.MachineChoice.BUILD)
-        if not self.tools['msgfmt'].found():
-            self.nogettext_warning(state.current_node)
-            return ModuleReturnValue(None, [])
+            self.tools['msgfmt'] = state.find_program('msgfmt', for_machine=mesonlib.MachineChoice.BUILD)
         podir = path.join(state.build_to_src, state.subdir, kwargs['po_dir'])
 
         ddirs = self._get_data_dirs(state, kwargs['data_dirs'])
--- a/mesonbuild/modules/i18n.py
+++ b/mesonbuild/modules/i18n.py
@@ -32,6 +32,7 @@ if T.TYPE_CHECKING:
     from ..build import Target
     from ..interpreter import Interpreter
     from ..interpreterbase import TYPE_var
+    from ..mparser import BaseNode
     from ..programs import ExternalProgram
 
     class MergeFile(TypedDict):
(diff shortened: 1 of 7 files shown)


mesonbuild/modules/i18n.py
```diff
@@ -219,9 +223,13 @@ def merge_file(self, state: 'ModuleState', args: T.List['TYPE_var'], kwargs: 'Me
...
-        if not shutil.which('xgettext'):
-            self.nogettext_warning()
-            return ModuleReturnValue(None, [])
+        for tool in ['msgfmt', 'msginit', 'msgmerge', 'xgettext']:
+            if self.tools[tool] is None:
+                self.tools[tool] = state.find_program(tool, required=False)
+            # still not found?
+            if not self.tools[tool].found():
+                self.nogettext_warning(state.current_node)
+                return ModuleReturnValue(None, [])
```
xclaesse, line 232: Nitpicking: you don't need modulereturnvalue when there are no targets. Just return nothing it none.

xclaesse:
OOC are those tools also defined in pkg-config like gnome tools?

Should you also pass for_machine=BUILD ?

Other than that, lgtm

eli-schwartz:
> OOC are those tools also defined in pkg-config like gnome tools?

No they are not. (This is a GNU project, which means half the time they respect standards they didn't start themselves and half the time they utterly refuse to while citing "portability" as an objection to a simple optional text file.)

> Should you also pass for_machine=BUILD ?

Ah... It's supposed to use the OS native tools which produce the same output as the cross compiled versions of the tools, yeah.

PR #10078, opened 2022-03-06
modules/wayland: Add version functions

markbolhuis:
Add protocols_version to get the version of wayland-protocols.
Add scanner_version to get the version of wayland-scanner.

CC @fooishbar 
CC @xclaesse

### DIFF
--- a/docs/markdown/Wayland-module.md
+++ b/docs/markdown/Wayland-module.md
@@ -66,6 +66,22 @@ It takes the following keyword arguments:
 server side header. Generated header files have the name
 `<name>-<client|server>-protocol.h`.
 
+### protocols_version
+```meson
+version = wl_mod.protocols_version()
+```
+This function does not take any positional or keyword arguments.
+
+**Returns**: The version [[@str]].
+
+### scanner_version
+```meson
+version = ml_mod.scanner_version()
+```
+This functions does not take any positional or keyboard arguments.
+
+**Returns**: The version [[@str]].
+
 ## Links
 - [Official Wayland Documentation](https://wayland.freedesktop.org/docs/html/)
 - [Wayland GitLab](https://gitlab.freedesktop.org/wayland)
--- a/mesonbuild/modules/unstable_wayland.py
+++ b/mesonbuild/modules/unstable_wayland.py
@@ -17,8 +17,8 @@ import os
 from . import ExtensionModule, ModuleReturnValue
 from ..build import CustomTarget
 from ..interpreter.type_checking import NoneType, in_set_validator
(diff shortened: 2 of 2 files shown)


xclaesse:
Why is this needed? You can already do `dependency('wayland-protocols').get_version()` and `find_program('wayland-scanner').get_version()`, no? But it's true that won't give you the scanner from pkgconfig... What about lookup for scanner program when you import the module, that way you know that any find_program will give you the same scanner.

eli-schwartz:
The module doesn't use the scanner from pkg-config either, so that's perfectly fine. If it did, then I'd expect you would still not want a function that returns the string version, but rather a function that wraps find_program to *find* the tool and set it in the module, which has a `version: '>=XXX'` kwarg.

markbolhuis:
It's redundant to find both the scanner and protocols package twice, and I'd rather not require users to use their own logic when this module already handles it. Also `find_program` from a meson.build file uses `MachineChoice.HOST`, whereas the module uses `MachineChoice.BUILD`. I'm not too familiar with cross compilation but that seems reason enough to me.

> `find_program('wayland-scanner').get_version()`, no?

There is no `get_version` function on an `ExternalProgramHolder`, or am I misinterpreting what you are saying?

eli-schwartz:
> Also `find_program` from a meson.build file uses `MachineChoice.HOST`, whereas the module uses `MachineChoice.BUILD`. I'm not too familiar with cross compilation but that seems reason enough to me.

The module uses the backend implementation of `find_program()` but passing the correct option which is exposed in meson.build as the `native: true` kwarg.

> There is no `get_version` function on an `ExternalProgramHolder`, or am I misinterpreting what you are saying?

It is correct that there is no such method. However, during original lookup you can add a required version such that lookup fails if it would not satisfy the version you want. IMHO if you need a specific version of wayland-scanner, you should not fetch the version as a string and then check whether that string is good enough and manually error out if it is not: you should just do:

```meson
find_program('wayland-scanner', native: true, version: '>=XXXX')
```

xclaesse:
> The module doesn't use the scanner from pkg-config either

Not yet, not it should: https://github.com/mesonbuild/meson/pull/10041

xclaesse:
> There is no `get_version` function on an `ExternalProgramHolder`, or am I misinterpreting what you are saying?

Hmmm, we have it internally, I was sure that was exposed... We should expose it IMHO.

eli-schwartz:
Done: #10079

xclaesse:
> `prog.version()` is useful for the case where you want to do different things depending on which version of prog you got, not for when you want to guarantee you got a specific minimum version of wayland-scanner

Agreed. Note that wayland-scanner has its own pkg-config file too, so `dependency('wayland-scanner').version()` would work too.

PR #10079, opened 2022-03-07
find_program: add a version() method to match the one for dependencies

eli-schwartz:
It is often useful to check the found version of a program without checking whether you can successfully find `find_program('foo', required: false, version: '>=XXX')`

/cc @xclaesse

### DIFF
--- a/docs/markdown/snippets/find_program-version.md
+++ b/docs/markdown/snippets/find_program-version.md
@@ -1,0 +1,5 @@
+## found programs now have a version method
+
+The return value of [[find_program]] can now check the exact version of the
+found program, independent of the minimum version requirement. This can be used
+e.g. to perform different actions depending on the exact version detected.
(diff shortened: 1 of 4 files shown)


docs/markdown/snippets/find_program-version.md
```diff
@@ -0,0 +1,5 @@
+## found programs now have a version method
+
+The return value of [[find_program]] can now check the exact version of the
+found program, independent of the minimum version requirement. This can be used
+e.g. to perform different actions depending on the exact version detected.
```
xclaesse, line 5: nitpicking: could tell the new method is `prog.version()`.

xclaesse:
LGTM

docs/markdown/snippets/find_program-version.md
eli-schwartz: I mentioned the "version" method in the markdown title, which I'm not sure can (or should) do anything complicated with formatting, and TBH it's exactly comparable to `dependency()`. Too lazy to change it. :D

xclaesse: ok fair enough.

PR #10081, opened 2022-03-07
Fix mypy CI.

jpakkane:


dcbaker (approved):
I was just writing this patch myself.

eli-schwartz:
Huh...

Commit ade6e3a19ed48c058138473a2f543031a1386393 had an invisible conflict with 0f2f87a003477bd79487a47d5f60f67751e2f44e that caused the mypy CI to fail, but only once merged. Instead of returning a version of 'unknown', Meson would crash due to a missing import.

PR #10082, opened 2022-03-07
Fix @typed_pos_args on java.generate_native_headers

tristan957:


tristan957:
Originally, I wanted to expand the test, and I am glad I did, so I could catch this bug prior to release.

PR #10085, opened 2022-03-08
fix correctly detecting whether python is found if link args are empty

### LINKED ISSUE #8570: Broken `python.dependency()` with PyPy, maybe
**Describe the bug**

Meson is erroring out when trying to detect `python.dependency()` when the interpreter is PyPy:

```
Compiler for C supports arguments -fvisibility=hidden: YES 
Configuring pycairo-1.20.0.egg-info using configuration
Dependency python found: NO

cairo/meson.build:35:0: ERROR: Python dependency not found
```

That line in `cairo/meson.build` is calling `python.dependency()`.

**To Reproduce**

Encountered trying to build [pycairo 1.20.0](https://github.com/pygobject/pycairo) in [conda-forge](https://github.com/conda-forge/pycairo-feedstock); specifically the pull request conda-forge/pycairo-feedstock#23. Can be reproduced locally by checking out the PR branch on a Linux machine and running `./build-locally.py linux_64_python3.6.____73_pypy`.

**Expected behavior**

No surprising Python-related failures.

**system parameters**
* what operating system? CentOS 6.10
* what Python version are you using? 3.6 with PyPy 7.3.3
* what `meson --version`: 0.57.1
* what `ninja --version` if it's a Ninja build: 1.10.2

**More Information**

The root cause of this issue seems to be the same as initially reported in #4841. Meson expects the PyPy library to be found in `${base}/bin`, which is not the case for conda-forge.

I see that the logic in this key line:

https://github.com/mesonbuild/meson/blob/1be13710adba2126f5731fccdf698a9d405a9d67/mesonbuild/modules/python.py#L149

was inverted in a later pull request #5873 by @jpakkane. The original logic seems correct to me, and it's not clear to me what that change had to do with the rest of the pull request — maybe something unintended snuck in?

**edit**: I did a quick check and `link_libpython` is false for conda-forge PyPy, so I think that the original logic would avoid the problem in this particular case.

eli-schwartz:
There are two cases where we can assume we found the python dependency with its requisite libraries using sysconfig:

- we found the library with find_library and are prepared to link to it
- the library is not actually part of the dependency, so its presence or absence is irrelevant

In the latter case, we should consider it found if link_libpython is False. Originally we did this, but the logic was inverted in commit 5b422fce87826beff3bca85e9c9081f22b3f45b7 in an unrelated change and without explanation, likely by accident.

Normally this doesn't much matter, since a python invariably comes with a predictably located libpython and the first condition evaluates true. But that is not true for pypy, and in fact that is the reason the link_libpython check was originally added in commit 1bd14b52b26b0f4ec207bf7e38813b82bfc15f86.

Restore that original logic.

Fixes #8570

### DIFF
--- a/mesonbuild/modules/python.py
+++ b/mesonbuild/modules/python.py
@@ -158,7 +158,7 @@ class PythonSystemDependency(SystemDependency, _PythonDependencyBase):
         if largs is not None:
             self.link_args = largs
 
-        self.is_found = largs is not None or self.link_libpython
+        self.is_found = largs is not None or not self.link_libpython
 
         inc_paths = mesonlib.OrderedSet([
             self.variables.get('INCLUDEPY'),


dcbaker (approved):
looking at the git history, your explanation seems likely. LGTM

Issue #9072, opened 2021-08-07
Support custom linkers.


abdullah-saal:
Getting this error when trying to use a different linker 
`Unsupported linker, only bfd, gold, and lld are supported, not mold.`  

I am trying to use this https://github.com/rui314/mold.  
Is it possible to add support for custom linkers?

dcbaker:
Meson needs to know about the linkers it invokes, as we have abstractions that site on top of them.

Sahnvour:
I was looking to add support for `mold` as it's a drop-in replacement for GNU `ld`, and didn't seem to require much modification in meson.<br/>
However, since meson uses `-fuse-ld=` pervasively, it will not work with GCC as is because it is reported as an incorrect argument (it only allows `ld`, `gold` and `lld` as parameters I guess). Do you have recommendations on how to handle this cleanly ?

dcbaker:
You'll probably have to do the -B trick that the mold documentation explains passed as LDFLAGS. We should see about adding detection logic for mold though

Sahnvour:
This is getting official support in [GCC 12](https://gcc.gnu.org/git/?p=gcc.git;a=commitdiff;h=ad964f7eaef9c03ce68a01cfdd7fde9d56524868) :)

PR #10086, opened 2022-03-08
linkers: Add support for mold linker

Finii:
**[why]**
Support for the relatively new mold linker is missing. If someone wants
to use mold as linker `LDFLAGS="-B/path/to/mold"` has to be added instead
of the usual `CC_LD=mold meson ...` or `CXX_LD=mold meson ...`.

**[how]**
Allow `mold' as linker for clang and newer GCC versions (that versions
that have support).

The error message can be a bit off, because it is generic for all GNU
like compilers, but I guess that is ok. (i.e. 'mold' is not listed as
possible linker, even if it would be possible for the given compiler.)

**[note]**
GCC Version 12.0.1 is not sufficient to say `mold` is supported. The
expected release with support will be 12.1.0.
On the other hand people that use the un-released 12.0.1 will probably
have built it from trunk. Allowing 12.0.1 is helping bleeding edge
developers to use mold in Meson already now.

Fixes: #9072

_Edit: Adapt to force push_

### DIFF
--- a/docs/markdown/Reference-tables.md
+++ b/docs/markdown/Reference-tables.md
@@ -46,6 +46,7 @@ These are return values of the `get_linker_id` method in a compiler object.
 | ld.bfd     | The GNU linker                              |
 | ld.gold    | The GNU gold linker                         |
 | ld.lld     | The LLVM linker, with the GNU interface     |
+| ld.mold    | The fast MOLD linker                        |
 | ld.solaris | Solaris and illumos                         |
 | ld.wasm    | emscripten's wasm-ld linker                 |
 | ld64       | Apple ld64                                  |
--- a/mesonbuild/compilers/mixins/gnu.py
+++ b/mesonbuild/compilers/mixins/gnu.py
@@ -1,4 +1,4 @@
-# Copyright 2019 The meson development team
+# Copyright 2019-2022 The meson development team
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -313,7 +313,7 @@ class GnuLikeCompiler(Compiler, metaclass=abc.ABCMeta):
         return ['-I' + path]
 
     @classmethod
-    def use_linker_args(cls, linker: str) -> T.List[str]:
+    def use_linker_args(cls, linker: str, version: str) -> T.List[str]:
         if linker not in {'gold', 'bfd', 'lld'}:
             raise mesonlib.MesonException(
                 f'Unsupported linker, only bfd, gold, and lld are supported, not {linker}.')
@@ -396,3 +396,9 @@ class GnuCompiler(GnuLikeCompiler):
         elif threads > 0:
             return [f'-flto={threads}']
         return super().get_lto_compile_args(threads=threads)
+
+    @classmethod
+    def use_linker_args(cls, linker: str, version: str) -> T.List[str]:
+        if linker == 'mold' and mesonlib.version_compare(version, '>=12.0.1'):
+            return ['-fuse-ld=mold']
+        return super().use_linker_args(linker, version)
--- a/mesonbuild/linkers/linkers.py
+++ b/mesonbuild/linkers/linkers.py
@@ -1,4 +1,4 @@
-# Copyright 2012-2017 The Meson development team
+# Copyright 2012-2022 The Meson development team
 
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -804,6 +804,11 @@ class GnuGoldDynamicLinker(GnuDynamicLinker):
 class GnuBFDDynamicLinker(GnuDynamicLinker):
 
     id = 'ld.bfd'
+
+
+class MoldDynamicLinker(GnuDynamicLinker):
+
+    id = 'ld.mold'
 
 
 class LLVMDynamicLinker(GnuLikeDynamicLinkerMixin, PosixDynamicLinkerMixin, DynamicLinker):
--- a/docs/markdown/snippets/mold-support.md
+++ b/docs/markdown/snippets/mold-support.md
@@ -1,0 +1,4 @@
+## Support for mold linker added
+
+The high performance linker mold can be selected via `CC_LD` or `CXX_LD` for
+Clang and GCC >= 12.0.1.
--- a/mesonbuild/compilers/compilers.py
+++ b/mesonbuild/compilers/compilers.py
@@ -1,4 +1,4 @@
-# Copyright 2012-2019 The Meson development team
+# Copyright 2012-2022 The Meson development team
 
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -1007,7 +1007,7 @@ class Compiler(HoldableObject, metaclass=abc.ABCMeta):
         return dep.get_link_args()
 
     @classmethod
-    def use_linker_args(cls, linker: str) -> T.List[str]:
+    def use_linker_args(cls, linker: str, version: str) -> T.List[str]:
         """Get a list of arguments to pass to the compiler to set the linker.
         """
         return []
--- a/mesonbuild/compilers/d.py
+++ b/mesonbuild/compilers/d.py
@@ -1,4 +1,4 @@
-# Copyright 2012-2017 The Meson development team
+# Copyright 2012-2022 The Meson development team
 
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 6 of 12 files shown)


Finii:
Linking to the Issue with the commit message/PR description did not work ... `#` was missing :grimacing: 

Force push.

mesonbuild/compilers/mixins/gnu.py
```diff
@@ -396,3 +396,11 @@ def get_lto_compile_args(self, *, threads: int = 0, mode: str = 'default') -> T.
         elif threads > 0:
             return [f'-flto={threads}']
         return super().get_lto_compile_args(threads=threads)
+
+    @classmethod
+    def use_linker_args(cls, linker: str) -> T.List[str]:
+        if hasattr(cls, 'version') and linker == 'mold':
+            # In detect.py cls is in fact self
+            if mesonlib.version_compare(cls.version, '>=12.1.0'):
+                return ['-fuse-ld=mold']
```
dcbaker, line 405: `cls` is the idiomatic name for referring to the `Type` object passed in a `classmethod`, while `self` is the idiomatic name of the `object` object passed to an instance method, so this comment isn't super useful.

Also, we shouldn't return something else here, we should error.

mesonbuild/linkers/linkers.py
```diff
@@ -790,7 +790,7 @@ def build_rpath_args(self, env: 'Environment', build_dir: str, from_dir: str,
 
 class GnuDynamicLinker(GnuLikeDynamicLinkerMixin, PosixDynamicLinkerMixin, DynamicLinker):
 
-    """Representation of GNU ld.bfd and ld.gold."""
+    """Representation of GNU ld.bfd, ld.gold, and mold."""
```
dcbaker, line 793: I'd rather not add mold here, the fact that mold acts so much like bfd and gold it can inherit that class directly is more of an implementation detail. I'm not entirely sure that it's safe to do that though, there may be cases where we use `isinstance(ld, GnuDynamicLinker)` that we don't want to find mold. It might makes more sense to copy the `LLVMDynaicLinker` hierarchy here, though we don't need to override anything it seems.

docs/markdown/Reference-tables.md
```diff
@@ -46,6 +46,7 @@ These are return values of the `get_linker_id` method in a compiler object.
 | ld.bfd     | The GNU linker                              |
 | ld.gold    | The GNU gold linker                         |
 | ld.lld     | The LLVM linker, with the GNU interface     |
+| ld.mold    | A Modern Linker                             |
```
dcbaker, line 49: I know `"A Modern Linker"` is the slogan for mold, but most of these are more descriptive. Maybe a "High performance GNU like linker"?

dcbaker (changes requested):
Thanks for working on this! I'd planned to get to it at some point but there's been so many other things on my plate. I have a few inline comments, but mostly this looks like the right direction.

This also needs a release snippet to be added to docs/markdown/snippets/

mesonbuild/compilers/mixins/gnu.py
```diff
@@ -396,3 +396,11 @@ def get_lto_compile_args(self, *, threads: int = 0, mode: str = 'default') -> T.
         elif threads > 0:
             return [f'-flto={threads}']
         return super().get_lto_compile_args(threads=threads)
+
+    @classmethod
+    def use_linker_args(cls, linker: str) -> T.List[str]:
+        if hasattr(cls, 'version') and linker == 'mold':
```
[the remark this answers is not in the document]
Finii: Removed/redesigned.

mesonbuild/compilers/mixins/gnu.py
Finii: Of course you are right, I was kind of blind here.
Removed/redesigned.

mesonbuild/linkers/linkers.py
Finii: Reverted.

docs/markdown/Reference-tables.md
Finii: Replaced by `The fast MOLD linker`.

I thought 'GNU like' is not saying anything specific (well it has the same interface, but unlike LLVM's linker it has just this one interface, so why mention it?

High performance was a bit too adverting speak for me, so downgraded to fast ;-)

I used the 'slogan' because that is the name/abbreviation expanded.

Finii:
This needed a deeper redesign if we want to allow different linkers based on compiler version. Started afresh.

**Why**

The new linker `mold` is supported by some compiler, but only with recent versions of that compilers. That means knowledge about the compiler version is needed in `add_linker_args()`.

This is not possible with the current design in Meson: While preparing the instantiation of the compiler object we want to collect the linker arguments. But we do not have a finished compiler object in that moment, so we can not reference the version easily via `self.version` within the compiler object - we are in an un-initialized compiler class.

See approx https://github.com/mesonbuild/meson/blob/0.61/mesonbuild/compilers/detect.py#L525: While detecting the C compiler the function `guess_nix_compiler()` is called. Now that function gets/has no knowledge of the compiler version, but would need it to determine the linker arguments (that are allowed of that version of the compiler class `cls`).

On the other hand, the version information has already been determined at that point in time. It is just not possible to access it from within the compiler mixin that determines allowed linker arguments.

**How**

Hand down the version information as new/additional argument. The interface has been changed for all compilers and platforms to keep the interface unified (i.e. `guess_windows_compiler()`, ...) This is the reason why so many files are touched. In all instances we just hand down the exact same `version` string to the linker-finding-mechanism that is later used to construct the compiler object.

Afterwards the version number is trivially available in `add_linker_args()` and we can decide if `mold` is accepted or not.

Finii:
Here a concise overview how the signature of existing functions were changed/expanded:
```diff
     @classmethod
-    def use_linker_args(cls, linker: str) -> T.List[str]:
+    def use_linker_args(cls, linker: str, version: str) -> T.List[str]:
```
```diff
 def guess_win_linker(env: 'Environment', compiler: T.List[str], comp_class: T.Type['Compiler'],
-                     for_machine: MachineChoice, *,
+                     comp_version: str, for_machine: MachineChoice, *,
                      use_linker_prefix: bool = True, invoked_directly: bool = True,
                      extra_args: T.Optional[T.List[str]] = None) -> 'DynamicLinker':
```
```diff
 def guess_nix_linker(env: 'Environment', compiler: T.List[str], comp_class: T.Type['Compiler'],
-                     for_machine: MachineChoice, *,
+                     comp_version: str, for_machine: MachineChoice, *,
                      extra_args: T.Optional[T.List[str]] = None) -> 'DynamicLinker':

```

dcbaker (approved):
LGTM, thanks for making those changes. I'll leave it up to @jpakkane whether this lands  for 0.62 or after the branchpoint

Finii:
Further information:

It would be possible to enable `mold` also on older GCCs, we just need to set the correct path, like so:
```diff
@@ -399,6 +400,10 @@ class GnuCompiler(GnuLikeCompiler):
 
     @classmethod
     def use_linker_args(cls, linker: str, version: str) -> T.List[str]:
-        if linker == 'mold' and mesonlib.version_compare(version, '>=12.0.1'):
-            return ['-fuse-ld=mold']
+        if linker == 'mold':
+            if mesonlib.version_compare(version, '>=12.0.1'):
+                return ['-fuse-ld=mold']
+            linker_exe = shutil.which(linker)
+            if linker_exe:
+                return [f'-B/usr/local/libexec/mold']
         return super().use_linker_args(linker, version)
```
The problem is, that we need the path to the symlink to `mold` that is called `ld`. And that can be anything - at least I do not know any limits to search this or if there are already similar cases in Meson. Typically it will be either `/usr/libexec/mold` or `/usr/local/libexec/mold`, but I do not dare to hardcode these paths.

```
$ ll /usr/local/libexec/mold
total 0
lrwxrwxrwx 1 root root 19 Mär  8 18:18 ld -> /usr/local/bin/mold
```

And then I realized that Meson does not honor the `LDFLAGS` when it searches for the linker. That is some kind of **bug** I would say. If the user specifies a different linker via `LDFLAGS`, it is not used/found when Meson starts and prints the 'used' linker, but it is later when really building it is honored.

Of course it is possible to use `LDFLAGS` of the environment like we use `CC_LD` in the linker search process, but that seems to contradict the other changes that were implemented in the linker search related code. So ... I will not propose anything in this direction.

Finii:
Another thing I found along the way, where I believe the code has an error...

image

```python
        if shutil.which(linker):
            if not shutil.which(linker):
                raise mesonlib.MesonException(
                    f'Cannot find linker {linker}.')
            return [f'-fuse-ld={linker}']
        return super().use_linker_args(linker, version)
```

Is it really intended to do these two `if`s? @dcbaker, at least for my understanding it can not be `true` and in the next line `None`. But it is in the same one commit.

dcbaker:
I noticed that too looking at this. It's probably a botched rebase that was missed, I'd say it's safe to delete the `if not...` case altogether.

dcbaker:
In the -B case, I'm happy to let users add `-Dc_link_args=-B...` for themselves and let the automatic detection solve it for them. There's simply too many variables to control for that to work reliably on our end.

Finii:
How it looks "on my machine" with GCC build from `trunk` and also selfbuild `mold`:

image

dcbaker:
That looks good, the failures on MSVC are expected, so I think once we get 0.62 out we'll get this merged

xclaesse:
> That looks good, the failures on MSVC are expected, so I think once we get 0.62 out we'll get this merged

It has been released already, LGTM.

PR #5775, opened 2019-08-06
Introduce dataonly for the pkgconfig module

shadeslayer:
This allows users to disable writing out the inbuilt variables to
the pkg-config file as they might actualy not be required.

https://gitlab.freedesktop.org/wayland/weston/issues/269 is a
great example of this.

Fixes #4011

PR #10089, opened 2022-03-08
Move dataonly pkgconfig file to datadir/pkgconfig

### LINKED ISSUE #9902: pkgconfig: `dataonly` files should be installed in `datadir`/pkgconfig
**Describe the bug**
Since `dataonly` pkg-config files are architecture-independent they [should](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s11.html) get installed in `datadir`/pkgconfig, and not `libdir`/pkgconfig

**To Reproduce**

```meson
project('dataonly.pc')

import('pkgconfig').generate(
	name: 'dataonly',
	dataonly: true,
	description: 'Should be in $prefix/share/pkgconfig'
)
```

```sh
$ tree install_prefix
install_prefix
└── lib
   └── x86_64-linux-gnu
      └── pkgconfig
         └── dataonly.pc
```

**Expected behavior**
The .pc file should get installed in $prefix/share/pkgconfig

**system parameters**
* Build type: native
* OS: Debian Testing
* Python version: 3.9.10
* Meson version: 0.61.1
* Ninja version: 1.10.1

tristan957:
dataonly files are architecture independent (lib vs lib64 for example).

Fixes #9902

CC @Tachi107 @eli-schwartz

### DIFF
--- a/docs/markdown/snippets/dataonly-pkgconfig-default-install-path.md
+++ b/docs/markdown/snippets/dataonly-pkgconfig-default-install-path.md
@@ -1,0 +1,4 @@
+## `dataonly` Pkgconfig Default Install Path
+
+The default install path for `dataonly` pkgconfig files has changed from
+`${libdir}/pkgconfig` to `${datadir}/pkgconfig`.
--- a/mesonbuild/modules/pkgconfig.py
+++ b/mesonbuild/modules/pkgconfig.py
@@ -524,6 +524,7 @@ class PkgConfigModule(ExtensionModule):
             blocked_vars = ['libraries', 'libraries_private', 'require_private', 'extra_cflags', 'subdirs']
             if any(k in kwargs for k in blocked_vars):
                 raise mesonlib.MesonException(f'Cannot combine dataonly with any of {blocked_vars}')
+            default_install_dir = os.path.join(state.environment.get_datadir(), 'pkgconfig')
 
         subdirs = mesonlib.stringlistify(kwargs.get('subdirs', default_subdirs))
         version = kwargs.get('version', default_version)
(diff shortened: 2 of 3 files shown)


Tachi107:
I believe this should be changed to `usr/share/pkgconfig/libhello_nolib.pc`

https://github.com/mesonbuild/meson/blob/afa47c90b02c4265518cc9773d00ad62486a8dd9/test%20cases/common/44%20pkgconfig-gen/test.json#L9

xclaesse:
I'm surprised this was not always the case. I'm a little bit worried that it could be seen as non backward compatible change, but to be honest if you put a dataonly pc file in libdir that's most probably a bug anyway. LGTM.

tristan957:
@xclaesse does this deserve a release snippet?

eli-schwartz:
I think probably yes.

@xclaesse from the linked issue, I commented:
> It probably should, but no one thought to mention that when #5775 was reviewed.

docs/markdown/snippets/dataonly-pkgconfig-default-install-path.md
```diff
@@ -0,0 +1,4 @@
+## `dataonly` Pkgconfig Default Install Path
+
+The default install path for `dataonly` pkgconfig files has changed from
+`${prefix}/${libdir}/pkgconfig` to `${prefix}/${datadir}/pkgconfig`.
```
eli-schwartz, line 4: Nit: it's just libdir/datadir, and prefix is inferred for relative paths -- which actually changed just recently to allow absolute paths outside of prefix.

We don't need to spell out the prefix here. Also I'm not sure we need to join it to the datadir in the code either... Relative install_dir paths are a first class feature.

tristan957: Ok sounds good. I will make the changes in both places.

eli-schwartz (approved):

PR #10093, opened 2022-03-08
Add a few easy module typing annotations

dcbaker:
This covers the test module and the icestorm module.

### DIFF
--- a/mesonbuild/modules/unstable_icestorm.py
+++ b/mesonbuild/modules/unstable_icestorm.py
@@ -13,83 +13,119 @@
 # limitations under the License.
 
 from __future__ import annotations
-
+import itertools
 import typing as T
 
-from . import ExtensionModule
+from . import ExtensionModule, ModuleReturnValue
+from .. import build
 from .. import mesonlib
+from ..interpreter.type_checking import CT_INPUT_KW
 from ..interpreterbase import FeatureNew
-from ..interpreterbase import flatten
+from ..interpreterbase.decorators import KwargInfo, typed_kwargs, typed_pos_args
 
 if T.TYPE_CHECKING:
+    from typing_extensions import TypedDict
+
+    from . import ModuleState
+    from ..interpreter import Interpreter
     from ..programs import ExternalProgram
+
+    class ProjectKwargs(TypedDict):
+
+        sources: T.List[T.Union[mesonlib.FileOrString, build.GeneratedTypes]]
+        constraint_file: T.Union[mesonlib.FileOrString, build.GeneratedTypes]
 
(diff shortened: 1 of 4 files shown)


mesonbuild/modules/unstable_icestorm.py
```diff
@@ -12,78 +12,119 @@
...
     @FeatureNew('FPGA/Icestorm Module', '0.45.0')
-    def __init__(self, interpreter):
+    def __init__(self, interpreter: Interpreter) -> None:
         super().__init__(interpreter)
-        self.yosys_bin = None
+        self.yosys_bin: T.Optional[ExternalProgram] = None
+        self.arachne_bin: T.Optional[ExternalProgram] = None
+        self.icepack_bin: T.Optional[ExternalProgram] = None
+        self.iceprog_bin: T.Optional[ExternalProgram] = None
+        self.icetime: T.Optional[ExternalProgram] = None
```
eli-schwartz, line 47: In other modules we turned this into a `self.tools: T.Dict`.

dcbaker: Yeah, that seems better.

mesonbuild/modules/unstable_icestorm.py
```diff
@@ -12,78 +12,115 @@
...
+        proj_name, arg_sources = args
+        all_sources = self.interpreter.source_strings_to_files(
+            list(itertools.chain(arg_sources, kwargs['sources'])))
+
+        blif_target = build.CustomTarget(
+            f'{proj_name}_blif',
+            state.subdir,
+            state.subproject,
+            [self.tools['yosys'], '-q', '-p', 'synth_ice40 -blif @OUTPUT@', '@INPUT@'],
+            all_sources,
```
eli-schwartz, line 81: mypy thinks that all_sources might contain StructuredSources here.

eli-schwartz (changes requested):
Could we re-order the 3rd commit to the beginning, so we don't add a bunch of self.X attributes and then immediately delete them?

mesonbuild/modules/unstable_icestorm.py
dcbaker: Yeah, I was unmotivated to tackle it

dcbaker: I’ll deal with it Monday

eli-schwartz: ping?

eli-schwartz: Ping?

eli-schwartz (approved):

PR #10096, opened 2022-03-09
upgrade some modules to type-checked arguments

eli-schwartz:
posargs only because it's not clear if/how to do kwargs.


mesonbuild/modules/dlang.py
```diff
@@ -52,13 +52,11 @@ def _init_dub(self, state):
             if not self.dubbin:
                 raise MesonException('DUB not found.')
 
+    @typed_pos_args('dlang.generate_dub_file', str, str)
     def generate_dub_file(self, state, args, kwargs):
```
xclaesse, line 56: @noKwargs ?

eli-schwartz: It actually does take kwargs, as mentioned in the commit message those kwargs are fully arbitrary and get added as json keys.

This is not the only place where we have an API of "you may add any kwarg you like and it gets treated as opaque data strings communicated to the file being generated".

xclaesse: hm, there are lots of kwargs, I guess that can be done later.

xclaesse:
LGTM

dcbaker (approved):

PR #10098, opened 2022-03-09
Replace freezing regex with plain text operations.

jpakkane:
The regex in question causes Python's regex parser to freeze
indefinitely in certain Python versions. That might be due
to a bug or because the re does infinite backtracking and it just
happened to work on earlier implementations.


jpakkane:
And, as lgtm says, it was the latter.

PR #10100, opened 2022-03-09
Add new env2mfile command.

jpakkane:
Attempt number two now that the regex backtracking issue has been fixed.

### DIFF
--- a/docs/markdown/Commands.md
+++ b/docs/markdown/Commands.md
@@ -150,6 +150,35 @@ Create a project in `sourcedir`:
 meson init -C sourcedir
 ```
 
+### env2mfile
+
+*This command is experimental and subject to change.*
+
+*{Since 0.62.0}*
+
+{{ env2mfile_usage.inc }}
+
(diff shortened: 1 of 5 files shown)


docs/markdown/snippets/env2cross.md
```diff
@@ -0,0 +1,40 @@
...
+
+Creating a native file is done in the same way:
+
+    meson env2mfile --native -o current_system.txt
+
+This system will detect if the `_FOR_BUILD` environment variables are
+enabled and then uses them as needed.
+
+With this you should be able to convert any envvar-based cross build
+setup to cross and native files and then use those. Thit means, among
```
Tachi107, line 30:
```suggestion
setup to cross and native files and then use those. This means, among
```

jpakkane:
The macOS failure is due to clock skew.

jpakkane:
The cancelled ones are due to the vs image brownout.

eli-schwartz:
Yay, the CI all passes now, so we can reinstate the code.

PR #10103, opened 2022-03-10
use shared implementation to convert files() strings to File objects

eli-schwartz:
This handles various edge cases.

### DIFF
--- a/mesonbuild/interpreter/interpreter.py
+++ b/mesonbuild/interpreter/interpreter.py
@@ -619,7 +619,7 @@ class Interpreter(InterpreterBase, HoldableObject):
     @typed_pos_args('files', varargs=str)
     @noKwargs
     def func_files(self, node: mparser.FunctionNode, args: T.Tuple[T.List[str]], kwargs: 'TYPE_kwargs') -> T.List[mesonlib.File]:
-        return [mesonlib.File.from_source_file(self.environment.source_dir, self.subdir, fname) for fname in args[0]]
+        return self.source_strings_to_files(args[0])
 
     # Used by declare_dependency() and pkgconfig.generate()
     def extract_variables(self, kwargs, argname='variables', list_new=False, dict_new=False):


dcbaker (approved):

xclaesse:
Does that mean it will refuse to create a File object for "subprojects/foo/bar" string? Does it abort or warn? Projects started doing that because we didn't have a better solution until recently.

eli-schwartz:
This was never supposed to work, because `files('sandbox/violation/path')` is illegal and fuzzing the parser to circumvent the check isn't something I care about at all.

It's illegal, it is warned about, and people's response to that is to say "oh, found a function that forgot to check, now all is well". :p

The correct fix is #10039 which actually works in both cases, whether you use `files(fname)` or `fname`. I think that PR is even ready to land, now.

eli-schwartz:
Can I please reiterate?

This forbids nothing. We only *intend* to allow dep.get_variable(), and this *continues* to work, with or without files().

This change means that `files()` now performs the same validation that anywhere else would perform, which includes not just sandbox violations, but also "you may not refer to files in the builddir, unless that is vala files and may G-d have mercy on your soul". And also warning about utf-8 filenames causing issues if your locale doesn't support that. etc.

xclaesse:
@eli-schwartz you merged despite I clearly said it breaks projects, and it does. I find that behavior frustrating especially after you reverted one of my commits for much less than that and you still haven't explained why despite I asked many times over past weeks.

I still think this deserve being deprecated first. Projects might not go through subp.get_variable() yet, and even if they do, it seems broken somehow because I get that error in gtk: `gdk/wayland/meson.build:81:4: ERROR: Sandbox violation: Tried to grab file primary-selection-unstable-v1.xml from a nested subproject.`. I think it is supposed to work because they do `files(join_paths(wlproto_dir, '@0@/@1@/@2@.xml'.format(proto_stability, proto_name, output_base)))` and `wlproto_dir = wlprotocolsdep.get_variable('pkgdatadir')`.

eli-schwartz:
I merged because I got a non-author approving review.

Also, I honestly believed that I satisfactorily answered your concern. Which, for the record, I still do. It conceptually doesn't break projects, and if people don't go through subp.get_variable() then I don't see how that is our problem, because in that case the sandbox explicitly exists to prevent that.

@jpakkane can help us decide whether making the sandbox ***actually work*** and stop people from doing forbidden things, is okay.

As for the gtk error, the answer there is pretty simple. My test case covered dependencies dealing with a  subdirectory of `meson.project_source_root()` but not a project which is `meson.project_source_root()` itself as one collection of installed data files.

This is indeed supposed to work, and it is a trivially fixable oversight that it didn't. In fact, the compromises I made in #10039 actually broke it. I had originally tested gtk with wayland-protocols as a subproject, but didn't test the revised modifications I made in response to reviews on that PR.

Thank you for testing this before release!

xclaesse:
> I merged because I got a non-author approving review.

That was before my very clear objection. I find this behavior really hard to understand when you're the the one who unilaterally reverted a perfectly fine commit and then ignored me for a month before giving a reason that is at best a nitpicking. Should I also push a revert of this PR?

> if people don't go through subp.get_variable() then I don't see how that is our problem, because in that case the sandbox explicitly exists to prevent that.

That's a new trick we didn't thought about before. It is perfectly fair for projects to still be doing `files('subprojects/foo')`. Luckily the few projects I checked does indeed use `subp.get_variable()` already, so we could get away with this, we'll see.

PR #10107, opened 2022-03-10
Remove a spurious debug print in Python module

rgommers:
This prints many lines of unwanted `"done /absolute/path"`, I noticed this when testing 0.62.0rc1 with SciPy.

Introduced in commit 79c6075b56, Cc @xclaesse. Can this be backported for 0.62.0?

[ci skip]

### DIFF
--- a/mesonbuild/modules/python.py
+++ b/mesonbuild/modules/python.py
@@ -414,7 +414,6 @@ class PythonExternalProgram(ExternalProgram):
         if subdir_parts[-len(install_subdir_parts):] == install_subdir_parts:
             pypath = os.path.join(basedir, *subdir_parts[:-len(install_subdir_parts)])
             self.devenv_pythonpath.add(pypath)
-            print('done', pypath)
 
     def _check_version(self, version: str) -> bool:
         if self.name == 'python2':


eli-schwartz:
Also: this is part of the code I didn't think should be merged, because I don't believe it's possible to reliably add PYTHONPATH for built artifacts in an out of tree build system. So we should probably just remove all devenv handling for the python module, and in the future maybe we can do editable installs by constructing something like structured sources or a staging wheel and use that.

rgommers:
> and in the future maybe we can do editable installs by constructing something like structured sources or a staging wheel and use that.

I don't know about this devenv PR, but indeed I think something like a .pth file pointing to a set of symlinks (structured sources / whatever) which point back to in-tree and builddir files sounds about right.

xclaesse:
Oops, LGTM

> I don't believe it's possible to reliably add PYTHONPATH

As far as I know the only case it won't work is when you mix a extension_module() with install_sources() in the same directory. I think it's still useful in all other case. devenv has plenty of limitations, it's mostly a best effort to help developers when possible.

rgommers:
> As far as I know the only case it won't work is when you mix a extension_module() with install_sources() in the same directory.

Which is going to be the case for pretty much every Python package that wants to adopt Meson, right? You always need `install_sources` (`__init__.py` at least), and you'll always have `extension_module` whenever you use compiled code.

xclaesse:
Yeah, I guess, unless it's only py files and no `extension_module`, maybe less common that I thought. My use-case when I wrote this is a python app that has a C library that gets used by python with gobject-introspection.

rgommers:
> Yeah, I guess, unless it's only py files and no `extension_module`, maybe less common that I thought. My use-case when I wrote this is a python app that has a C library that gets used by python with gobject-introspection.

Yes, I'd say that's the <1% use case. Manual use of `ctypes` is similar. Every regular Python package will use Cython, the CPython C API, pybind11, f2py or similar if they have compiled code - and those all won't work.

PR #10108, opened 2022-03-10
Add type annotations to the Pkgconfig Module, and fix lots of bugs

dcbaker:
~This is based on #10093.~

This began as a way to add type annotations and typed_*args to the pkgconfig module, but the typing annotations ended up leading up leading me to notice a number of other issues, largely around using CustomTarget and CustomTargetIndex as linkable targets, and around around link_whole, and then found some places where our annotations weren't strictly correct. The end result is 24 commits stepping all over the tree. I could probably shave off the last few commits and separate them into a different PR, if there was desire to land them separately.

### DIFF
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -58,6 +58,7 @@ if T.TYPE_CHECKING:
 
     GeneratedTypes = T.Union['CustomTarget', 'CustomTargetIndex', 'GeneratedList']
     LibTypes = T.Union['SharedLibrary', 'StaticLibrary', 'CustomTarget', 'CustomTargetIndex']
+    BuildTargetTypes = T.Union['BuildTarget', 'CustomTarget', 'CustomTargetIndex']
 
 pch_kwargs = {'c_pch', 'cpp_pch'}
 
@@ -1066,11 +1067,11 @@ class BuildTarget(Target):
         return ExtractedObjects(self, self.sources, self.generated, self.objects,
                                 recursive)
 
-    def get_all_link_deps(self) -> 'ImmutableListProtocol[T.Union[BuildTarget, CustomTarget, CustomTargetIndex]]':
+    def get_all_link_deps(self) -> ImmutableListProtocol[BuildTargetTypes]:
         return self.get_transitive_link_deps()
 
     @lru_cache(maxsize=None)
-    def get_transitive_link_deps(self) -> 'ImmutableListProtocol[T.Union[BuildTarget, CustomTarget, CustomTargetIndex]]':
+    def get_transitive_link_deps(self) -> ImmutableListProtocol[BuildTargetTypes]:
         result: T.List[Target] = []
         for i in self.link_targets:
             result += i.get_all_link_deps()
@@ -2402,7 +2403,7 @@ class CommandBase:
     dependencies: T.List[T.Union[BuildTarget, 'CustomTarget']]
     subproject: str
 
-    def flatten_command(self, cmd: T.Sequence[T.Union[str, File, programs.ExternalProgram, 'BuildTarget', 'CustomTarget', 'CustomTargetIndex']]) -> \
+    def flatten_command(self, cmd: T.Sequence[T.Union[str, File, programs.ExternalProgram, BuildTargetTypes]]) -> \
             T.List[T.Union[str, File, BuildTarget, 'CustomTarget']]:
         cmd = listify(cmd)
         final_cmd: T.List[T.Union[str, File, BuildTarget, 'CustomTarget']] = []
@@ -2444,10 +2445,11 @@ class CustomTarget(Target, CommandBase):
                  subproject: str,
                  environment: environment.Environment,
                  command: T.Sequence[T.Union[
-                     str, BuildTarget, CustomTarget, CustomTargetIndex, GeneratedList, programs.ExternalProgram, File]],
+                     str, BuildTargetTypes, GeneratedList,
+                     programs.ExternalProgram, File]],
                  sources: T.Sequence[T.Union[
-                     str, File, BuildTarget, CustomTarget, CustomTargetIndex,
-                     ExtractedObjects, GeneratedList, programs.ExternalProgram]],
+                     str, File, BuildTargetTypes, ExtractedObjects,
+                     GeneratedList, programs.ExternalProgram]],
                  outputs: T.List[str],
                  *,
                  build_always_stale: bool = False,
@@ -2633,7 +2635,7 @@ class RunTarget(Target, CommandBase):
     typename = 'run'
 
     def __init__(self, name: str,
-                 command: T.Sequence[T.Union[str, File, BuildTarget, 'CustomTarget', 'CustomTargetIndex', programs.ExternalProgram]],
+                 command: T.Sequence[T.Union[str, File, BuildTargetTypes, programs.ExternalProgram]],
                  dependencies: T.Sequence[Target],
                  subdir: str,
                  subproject: str,
--- a/mesonbuild/modules/pkgconfig.py
+++ b/mesonbuild/modules/pkgconfig.py
@@ -538,7 +538,7 @@ class PkgConfigModule(ExtensionModule):
             raise mesonlib.MesonException('dataonly must be boolean.')
         if dataonly:
             default_subdirs = []
-            blocked_vars = ['libraries', 'libraries_private', 'require_private', 'extra_cflags', 'subdirs']
+            blocked_vars = ['libraries', 'libraries_private', 'requires_private', 'extra_cflags', 'subdirs']
             if any(k in kwargs for k in blocked_vars):
                 raise mesonlib.MesonException(f'Cannot combine dataonly with any of {blocked_vars}')
             default_install_dir = os.path.join(state.environment.get_datadir(), 'pkgconfig')
--- a/mesonbuild/modules/pkgconfig.py
+++ b/mesonbuild/modules/pkgconfig.py
@@ -1,4 +1,4 @@
-# Copyright 2015 The Meson development team
+# Copyright 2015-2022 The Meson development team
 
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -14,6 +14,7 @@
 
 from __future__ import annotations
 from collections import defaultdict
+from dataclasses import dataclass
 from pathlib import PurePath
 import os
 import typing as T
@@ -34,6 +35,7 @@ if T.TYPE_CHECKING:
     from typing_extensions import TypedDict
 
     from . import ModuleState
+    from .. import mparser
     from ..interpreter import Interpreter
 
     ANY_DEP = T.Union[dependencies.Dependency, build.BuildTargetTypes, str]
@@ -62,8 +64,6 @@ if T.TYPE_CHECKING:
         unescaped_uninstalled_variables: T.Dict[str, str]
 
 
-already_warned_objs = set()
-
 _PKG_LIBRARIES: KwargInfo[T.List[T.Union[str, dependencies.Dependency, build.SharedLibrary, build.StaticLibrary, build.CustomTarget, build.CustomTargetIndex]]] = KwargInfo(
     'libraries',
     ContainerTypeInfo(list, (str, dependencies.Dependency,
@@ -86,8 +86,17 @@ def _as_str(obj: object) -> str:
     return obj
 
 
+@dataclass
+class MetaData:
+
+    filebase: str
+    display_name: str
+    location: mparser.BaseNode
+    warned: bool = False
+
+
 class DependenciesHelper:
-    def __init__(self, state: ModuleState, name: str) -> None:
+    def __init__(self, state: ModuleState, name: str, metadata: T.Dict[str, MetaData]) -> None:
         self.state = state
         self.name = name
         self.pub_libs: T.List[LIBS] = []
@@ -97,6 +106,7 @@ class DependenciesHelper:
         self.cflags: T.List[str] = []
         self.version_reqs: T.DefaultDict[str, T.Set[str]] = defaultdict(set)
         self.link_whole_targets: T.List[T.Union[build.CustomTarget, build.CustomTargetIndex, build.StaticLibrary]] = []
+        self.metadata = metadata
 
     def add_pub_libs(self, libs: T.Sequence[ANY_DEP]) -> None:
         p_libs, reqs, cflags = self._process_libs(libs, True)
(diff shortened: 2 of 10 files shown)


mesonbuild/build.py
```diff
@@ -2512,6 +2512,16 @@ def is_linkable_target(self) -> bool:
         suf = os.path.splitext(self.outputs[0])[-1]
         return suf in {'.a', '.dll', '.lib', '.so', '.dylib'}
 
+    def can_static_link(self) -> bool:
+        """Whether this target can be linked statically.
+
+        Does not assert the target is linkable, just that it is not shared
+
+        :return: If this target is stacally linkable
```
tristan957, line 2520: s/stacally/statically

tristan957: Same mistake in both functions. CustomTargetIndex is so strange...

unittests/linuxliketests.py
```diff
@@ -54,6 +56,18 @@
 from .baseplatformtests import BasePlatformTests
 from .helpers import *
 
+def _prepend_pkg_config_path(path: str) -> str:
```
tristan957, line 59: Feels like there must be a helper function for this in the source code already that this guy could make use of

dcbaker: Not that I'm aware of. Usually this is just `os.path.pathsep.join(list_of_paths)`

eli-schwartz:
> I could probably shave off the last few commits and separate them into a different PR, if there was desire to land them separately.

Yes please. (I like smaller, more digestible PRs to be honest.)

eli-schwartz:
⚔️ conflicts.

mesonbuild/modules/pkgconfig.py
```diff
@@ -25,15 +25,65 @@
...
+
     from . import ModuleState
     from ..interpreter import Interpreter
 
+    ANY_DEP = T.Union[
+        dependencies.Dependency, build.SharedLibrary, build.StaticLibrary,
+        build.CustomTarget, build.CustomTargetIndex, str]
+    LIBS = T.Union[
+        str, build.SharedLibrary, build.StaticLibrary, build.CustomTarget,
+        build.CustomTargetIndex]
```
eli-schwartz, line 46: This honestly feels like something the build / dependencies subpackages should be providing, which would be useful for more than just pkgconfig.

mesonbuild/modules/pkgconfig.py
```diff
@@ -587,32 +639,30 @@ def parse_variable_list(vardict):
                 variables.append((name, value))
             return variables
 
-        variables = self.interpreter.extract_variables(kwargs, dict_new=True)
-        variables = parse_variable_list(variables)
-        unescaped_variables = self.interpreter.extract_variables(kwargs, argname='unescaped_variables')
-        unescaped_variables = parse_variable_list(unescaped_variables)
+        if isinstance(kwargs['variables'], dict):
+            FeatureNew.single_use('variables as dictionary', '0.56.0', state.subproject, location=state.current_node)
```
eli-schwartz, line 694: For declare_dependency I handled this via 
```python
VARIABLES_KW.evolve(since='0.54.0', since_values={list: '0.56.0'}),
```
after making that work in c1aeb2f7e4ef81a046ca28d5dad89bb8fdeefafd.

(It will also need to override the `default={}` if we are making it default to list and use dict as since? Maybe?)

mesonbuild/modules/pkgconfig.py
```diff
@@ -499,23 +549,37 @@ def generate_uninstalled_cflags(libs):
...
+        KwargInfo('dataonly', bool, default=False, since='0.54.0'),
+        KwargInfo('description', str, default=''),
+        KwargInfo('extra_cflags', ContainerTypeInfo(list, str), default=[], listify=True, since='0.42.0'),
+        KwargInfo('filebase', str, default=''),
+        KwargInfo('name', str, default=''),
+        KwargInfo('subdirs', ContainerTypeInfo(list, str), default=[], listify=True),
+        KwargInfo('url', str, default=''),
+        KwargInfo('version', str, default=''),
+        VARIABLES_KW.evolve(name="unescaped_uninstalled_variables", since='0.54.0'),
+        VARIABLES_KW.evolve(name="unescaped_variables", since='0.54.0'),
```
eli-schwartz, line 611: These are new since 0.59.0, only the regular uninstalled variables are new since 0.54.0.

mesonbuild/modules/pkgconfig.py
```diff
@@ -499,23 +549,37 @@ def generate_uninstalled_cflags(libs):
...
-                      'unescaped_variables', 'unescaped_uninstalled_variables'})
     @typed_pos_args('pkgconfig.generate', optargs=[(build.SharedLibrary, build.StaticLibrary)])
-    def generate(self, state: ModuleState, args: T.Tuple[T.Optional[T.Union[build.SharedLibrary, build.StaticLibrary]]], kwargs):
+    @typed_kwargs(
+        'pkgconfig.generate',
+        D_MODULE_VERSIONS_KW.evolve(since='0.43.0'),
+        INSTALL_DIR_KW,
+        KwargInfo('conflicts', ContainerTypeInfo(list, str), default=[], listify=True),
+        KwargInfo('dataonly', bool, default=False, since='0.54.0'),
+        KwargInfo('description', str, default=''),
```
eli-schwartz, line 603: Currently, if someone explicitly passes `description: ''` it will override the description from the default. After making it default to an empty string and checking `bool(kwargs['description'])` that no longer works, we should probably instead check if it is not the sentinel value None.

mesonbuild/modules/pkgconfig.py
```diff
@@ -499,23 +549,37 @@ def generate_uninstalled_cflags(libs):
...
+        KwargInfo('extra_cflags', ContainerTypeInfo(list, str), default=[], listify=True, since='0.42.0'),
+        KwargInfo('filebase', str, default=''),
+        KwargInfo('name', str, default=''),
```
eli-schwartz, line 606: In a similar vein, these would allow passing `''` as the value but then trip over a logic error I guess (because they cannot be empty!) -- perhaps we should actually check that in a validator?

mesonbuild/modules/pkgconfig.py
```diff
@@ -525,60 +589,48 @@ def generate(self, state: ModuleState, args: T.Tuple[T.Optional[T.Union[build.Sh
...
 
-        dataonly = kwargs.get('dataonly', False)
-        if not isinstance(dataonly, bool):
-            raise mesonlib.MesonException('dataonly must be boolean.')
+        dataonly = kwargs['dataonly']
         if dataonly:
             default_subdirs = []
-            blocked_vars = ['libraries', 'libraries_private', 'require_private', 'extra_cflags', 'subdirs']
-            if any(k in kwargs for k in blocked_vars):
+            blocked_vars = ['libraries', 'libraries_private', 'requires_private', 'extra_cflags', 'subdirs']
```
eli-schwartz, line 648: Fixing this typo probably deserves a distinct commit.

mesonbuild/modules/pkgconfig.py
eli-schwartz: Same with the version, I guess.

eli-schwartz (changes requested):
commit `modules/pkgconfig: link_with can contain an Executable` does a couple of things that should be folded into previous commits, not all of which are mentioned in the commit message.

Why do we add `_T = T.TypeVar` in one commit, then remove it in another?

mesonbuild/modules/pkgconfig.py
dcbaker: This is actually super annoying to deal with, for lots of reasons. Blarg.

dcbaker: Actually, it turns out not to be. using since_values actually does exatly what we want, because it doesn't check in the case that the value is the default, so we can set the default value to `{}` and it will not emit a warning at all, unlike the current code

mesonbuild/modules/pkgconfig.py
```diff
@@ -496,89 +596,95 @@ def generate_uninstalled_cflags(libs):
...
-        dataonly = kwargs.get('dataonly', False)
-        if not isinstance(dataonly, bool):
-            raise mesonlib.MesonException('dataonly must be boolean.')
+        else:
+            if kwargs['version'] is None:
+                FeatureNew.single_use('pkgconfig.generate implicit version keyword', '0.46.0', state.subproject)
+            if kwargs['name'] is None:
+                raise build.InvalidArguments(
+                    'pkgconfig.generate: if a library is not passed as a '
+                    'positional arugment, the name keyword argument is '
```
eli-schwartz, line 645:
```suggestion
                    'positional argument, the name keyword argument is '
```

mesonbuild/modules/pkgconfig.py
```diff
@@ -250,7 +330,17 @@ def _add_exclude(x):
         for t in self.link_whole_targets:
             _add_exclude(t)
 
-        def _fn(xs, libs=False):
+        # Mypy thinks these overlap, but since List is invaraint they don't,
```
eli-schwartz, line 333:
```suggestion
        # Mypy thinks these overlap, but since List is invariant they don't,
```

eli-schwartz (commented):
minor typos

eli-schwartz:
I think the only remaining review comment I have is about what the best place for ANY_DEP is.

dcbaker:
I'm just not sure about that, since the ability to pass `str` is, AFAICT, unique to the `pkgconfig` module, I don't think any other interfaces accept it.

eli-schwartz:
So? It's a useful abstraction for the non-str case that probably makes sense to use a bit more tree-wide.

For the pkgconfig module we can just use `T.Union[str, ANY_DEP]`.

eli-schwartz:
Ping?

eli-schwartz:
Alright, this can now be simplified to:

```meson
ANY_DEP = T.Union[str, build.LibTypes, dependencies.Dependency]
LIBS = T.Union[str, build.LibTypes]
```

or some such. Put in that context, I'm not sure how generally useful ANY_DEP is. LIBS could be inlined where used or not, to taste...

mesonbuild/modules/pkgconfig.py
```diff
@@ -587,32 +689,28 @@ def parse_variable_list(vardict):
...
                                       version, pcfile, conflicts, variables,
                                       unescaped_variables, False, dataonly,
                                       pkgroot=pkgroot if relocatable else None)
         res = build.Data([mesonlib.File(True, state.environment.get_scratch_dir(), pcfile)], pkgroot, pkgroot_name, None, state.subproject, install_tag='devel')
-        variables = self.interpreter.extract_variables(kwargs, argname='uninstalled_variables', dict_new=True)
-        variables = parse_variable_list(variables)
-        unescaped_variables = self.interpreter.extract_variables(kwargs, argname='unescaped_uninstalled_variables')
-        unescaped_variables = parse_variable_list(unescaped_variables)
+        if isinstance(kwargs['uninstalled_variables'], dict):
+            FeatureNew.single_use('uninstalled_variables as dictionary', '0.56.0', state.subproject, location=state.current_node)
```
anarazel, line 711: This seems to trigger spuriously in postgres (which uses a meson minimum of 0.54), likely because uninstalled_variables defaults to being a dict.

dcbaker: This... shouldn't exist. This is all *supposed* to be handled in `typed_kwargs`. I've deleted this check.

mesonbuild/modules/pkgconfig.py
```diff
@@ -78,33 +80,39 @@ class GenerateKw(TypedDict):
...
-    def add_priv_libs(self, libs) -> None:
-        libs, reqs, _ = self._process_libs(libs, False)
-        self.priv_libs = libs + self.priv_libs
+    def add_priv_libs(self, libs: T.Sequence[ANY_DEP]) -> None:
+        p_libs, reqs, _ = self._process_libs(libs, False)
+        self.priv_libs = p_libs + self.priv_libs
         self.priv_reqs += reqs
 
-    def add_pub_reqs(self, reqs) -> None:
+    def add_pub_reqs(self, reqs: T.Sequence[T.Union[str, dependencies.Dependency]]) -> None:
```
eli-schwartz: This does not rely on covariance, and can be T.List; important, because of the T.Sequence[str] trap.

mesonbuild/modules/pkgconfig.py
```diff
@@ -78,33 +80,39 @@ class GenerateKw(TypedDict):
...
         self.pub_reqs += self._process_reqs(reqs)
 
-    def add_priv_reqs(self, reqs) -> None:
+    def add_priv_reqs(self, reqs: T.Sequence[T.Union[str, dependencies.Dependency]]) -> None:
```
eli-schwartz: ... and here

mesonbuild/modules/pkgconfig.py
```diff
@@ -124,9 +132,9 @@ def _check_generated_pc_deprecation(self, obj) -> None:
                          location=obj.generated_pc_warn[1])
         already_warned_objs.add((name, obj.name))
 
-    def _process_reqs(self, reqs: T.List[T.Union[str, dependencies.Dependency]]):
+    def _process_reqs(self, reqs: T.Sequence[T.Union[str, dependencies.Dependency]]) -> T.List[str]:
```
eli-schwartz: ... and here, the good news is that just not changing anything, works. :)

mesonbuild/modules/pkgconfig.py
```diff
@@ -230,8 +230,8 @@ def _process_libs(
         return processed_libs, processed_reqs, processed_cflags
 
     def _add_lib_dependencies(
-            self, link_targets: T.List[T.Union[build.StaticLibrary, build.SharedLibrary, build.CustomTarget, build.CustomTargetIndex]],
-            link_whole_targets: T.List[T.Union[build.StaticLibrary, build.CustomTarget, build.CustomTargetIndex]],
+            self, link_targets: T.Sequence[build.BuildTargetTypes],
+            link_whole_targets: T.Sequence[T.Union[build.StaticLibrary, build.CustomTarget, build.CustomTargetIndex]],
```
eli-schwartz, line 234: ... and here, link_whole doesn't need to be covariant, but it doesn't support str anyway so that's not a big deal.

mesonbuild/modules/pkgconfig.py
```diff
@@ -151,14 +159,16 @@ def _process_reqs(self, reqs: T.List[T.Union[str, dependencies.Dependency]]):
                                               f'or pkgconfig-dependency object, got {obj!r}')
         return processed_reqs
 
-    def add_cflags(self, cflags) -> None:
+    def add_cflags(self, cflags: T.List[str]) -> None:
         self.cflags += mesonlib.stringlistify(cflags)
 
-    def _process_libs(self, libs, public: bool) -> T.Tuple[T.List, T.List, T.List]:
+    def _process_libs(
+            self, libs: T.Sequence[ANY_DEP], public: bool
```
eli-schwartz, line 177: This could avoid being a Sequence, if...

mesonbuild/modules/pkgconfig.py
```diff
@@ -78,33 +80,39 @@ class GenerateKw(TypedDict):
...
+        self.pub_libs: T.List[LIBS] = []
+        self.pub_reqs: T.List[str] = []
+        self.priv_libs: T.List[LIBS] = []
+        self.priv_reqs: T.List[str] = []
+        self.cflags: T.List[str] = []
+        self.version_reqs: T.DefaultDict[str, T.Set[str]] = defaultdict(set)
+        self.link_whole_targets: T.List[T.Union[build.CustomTarget, build.CustomTargetIndex, build.StaticLibrary]] = []
+
+    def add_pub_libs(self, libs: T.Sequence[ANY_DEP]) -> None:
+        p_libs, reqs, cflags = self._process_libs(libs, True)
```
eli-schwartz, line 112: ... right here we add
```python
assert isinstance(libs, list), 'for mypy'
````

Which we should do anyway, because process_libs should not conceptually take `self._process_libs('foo', True)`, that will break.

File: mesonbuild/modules/pkgconfig.py (hunk 12 above)
Comment (dcbaker) [reply mesonbuild/modules/pkgconfig.py]: Yeah, actually it does. it gets called in `self._add_lib_dependencies` with `List[Dependency]`

Comment (eli-schwartz) [reply mesonbuild/modules/pkgconfig.py]: Nope, that calls add_pub_libs / add_priv_libs.

It passes mypy as List.

Comment (dcbaker) [reply mesonbuild/modules/pkgconfig.py]:
```
mesonbuild/modules/pkgconfig.py:254:32: error: Argument 1 to "add_priv_libs" of "DependenciesHelper" has incompatible type "List[Dependency]"; expected "List[Union[Dependency, Union[BuildTarget, CustomTarget, CustomTargetIndex], str]]"
mesonbuild/modules/pkgconfig.py:254:32: note: "List" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
mesonbuild/modules/pkgconfig.py:254:32: note: Consider using "Sequence" instead, which is covariant
```

eli-schwartz (approved):

PR #10109, opened 2022-03-10
boost: preserve compiler/linker argument order

trhd:
Looks like boost dependency mixes up it's compiler and linker argument
order when it is removing duplicates (?) from those. This causes
unnecessary recompilations of everything depending on those components.

Use OrderedSet to remove the duplicates while also maintaining
consistent order for them.


dcbaker (approved):

jpakkane:
Test failure is unrelated.

PR #10110, opened 2022-03-11
Revert "devenv: Set PYTHONPATH where we install python modules"

eli-schwartz:
This reverts commit 79c6075b560dbf1c3e4e0b30f1c472dc2086421e.

```
# Conflicts:
#	docs/markdown/snippets/devenv.md
#	mesonbuild/modules/python.py
#	test cases/unit/91 devenv/test-devenv.py
```

PYTHONPATH cannot be reliably determined. The standard use case for installing python modules with Meson is mixed pure sources (at least `__init__.py`) and compiled extension_modules or configured files. Unfortunately that doesn't actually work because python will not load the same package hierarchy from two different directories, one a source directory and one a (mandatory) out of tree build directory.

(It kind of can, but you need to do what this test case accidentally stumbled upon, which is namespace packages. Namespace packages are a very specific use case and you are NOT SUPPOSED to use them outside that use case, so people are not going to use them just to circumvent Meson devenv stuff as that would have negative install-time effects.)

Adding PYTHONPATH anyway will just lead to documentation commitments which we cannot actually uphold, and confusing issues at time of use because some imports *will* work... and some will *not*. The end result will be a half-created tree of modules which just doesn't work together at all, but because it partially works, users attempting to debug it will spend time wondering why parts of it do import.

For any case where the automatic devenv would work correctly, it will also work correctly to use `meson.add_devenv()` a single time, which is very easy to manually get correct and doesn't provide any significant value to automate.

In the long run, an uninstalled python package environment will require "editable installs" support.

/cc @xclaesse @rgommers


eli-schwartz:
@jpakkane I would like to have this considered for 0.62.0rc2 as it reverts a feature I'm not sure we should release in the first place.

docs/markdown/Commands.md
```diff
@@ -340,15 +340,6 @@ These variables are set in environment in addition to those set using `meson.add
...
   more than one directory.
-- `PYTHONPATH` *Since 0.62.0* includes every directory where a python module is being
-  installed using [`python.install_sources()`](Python-module.md#install_sources)
-  and [`python.extension_module()`](Python-module.md#extension_module). Python
-  modules installed by other means, such as `install_data()` or `install_subdir()`,
-  will not be included and should be added to `PYTHONPATH` manually using
-  [`meson.add_devenv()`](Reference-manual_builtin_meson.md#mesonadd_devenv).
-  Note that when modules are installed into subdirectories the source tree
-  layout must match the installed tree layout otherwise `import subdir.mod`
-  cannot work.
```
rgommers, line 351: Agreed, this should be removed. Use of PYTHONPATH is not a habit that tools should teach to users even if it would work reliably, which it definitely won't for out of tree builds.

Long-term we can indeed make editable installs (`pip install . -e` and the like) work.

xclaesse: > Use of PYTHONPATH is not a habit that tools should teach to users even if it would work reliably

Why?

I think it's still useful even if it has limitations. devenv has plenty of limitations, it's mostly a best effort to help developers when possible. Can of course always be extended by symlinking everything in a common location, can be done later IMHO.

My use-case when I wrote this is a python app that has a C library that gets used by python with gobject-introspection. devenv already sets GI_TYPELIB_PATH so python bindings for glib based libraries does work.

xclaesse:
I personally object on reverting this, not because something is not perfect that it is not useful. Will leave it to others to judge.

eli-schwartz:
> not because something is not perfect that it is not useful

Again, the stated goal of this PR is because I believe it to be a *combination* of:
- not useful in the 99% of cases
- documented as reliable
- actually has legit negative side effects

I could live with something that is often useful, but occasionally does nothing.

This is not that case. This is rarely useful, and most of the time does the *wrong* thing. We should not ever be doing the ***wrong*** thing, and we certainly shouldn't do the wrong thing in the majority case.

docs/markdown/Commands.md
eli-schwartz: I disagree with the idea that we should expect limitations and implement broken stuff that actively makes things worse, because "it's just a best effort".

Instead, we should say that devenv doesn't implement everything, so yes it is limited, but what it does implement is always better than not implementing it.

Using highly loaded language like "it's just a best effort" is not needed and has no effect here other than to play a sympathy card which I don't think is topical. We can evaluate each component of devenv based on whether it is individually productive, removing PYTHONPATH isn't an indictment on the rest of devenv, and we don't need to make excuses for why a destructive component should stay in regardless because "it has good intentions".

eli-schwartz: In fact, as mentioned in the commit message, devenv really is super useful! It provides an easy, helpful framework for extending itself via `meson.add_devenv()` for when you as the developer of a specific project know that PYTHONPATH will actually work correctly in that project.

I am very much in support of devenv! Used properly. Where we only implement stuff expected to be correct, and provide people easy tools to build on that in project-specific ways.

rgommers: @xclaesse "why?" can have a long or a very short answer. The long one involves years of history on PYTHONPATH, how `setuptools` didn't treat it right, env vars not being best practice, etc. But the short one is probably enough: the consensus in the Python packaging crowd is that it's not the thing to teach or use. For valid use cases there should be better designed alternatives - which in this case is editable installs.

Note that I personally do use PYTHONPATH on occasion, it can be handy. But it's not something that should be taught or made the default in a widely used tool like Meson.

jpakkane: > But the short one is probably enough: the consensus in the Python packaging crowd is that it's not the thing to teach or use. 

Based on this it seems like a reasonable thing to revert this.

PR #10111, opened 2022-03-11
mtest: fix logic bug that broke tests where the cmd is a Windows found program

### LINKED ISSUE #10027: Spurious "The test program '...' does not exist. Cannot run tests before building them."
Hi,

Starting with c85f7a39a80ccbadb9ffb76102da30c05e3db66f I'm seeing postgres' tests (in my WIP branch using meson) fail on windows with $subject. I think the problem is the os.path.isfile() test introduced in the patch. Previously one could use a shebang that worked across operating systems (i.e. no .exe, with the binary being found via PATH), but now that'll fail because of course the file only exists with the file extension.

Of course I can work around that by not relying on the extracted executable. But I bet this will bite not just us.

Shouldn't this be shutil.which() instead of `os.path.isfile(testentry)`? And probably also gated by self.test.cmd_is_built?

```diff
@@ -1347,9 +1347,12 @@ class SingleTestRunner:
             self.console_mode = ConsoleUser.LOGGER
 
     def _get_test_cmd(self) -> T.Optional[T.List[str]]:
-        if self.test.fname[0].endswith('.jar'):
+        testentry = self.test.fname[0]
+        if self.options.no_rebuild and not os.path.isfile(testentry):
+            raise TestException(f'The test program {testentry!r} does not exist. Cannot run tests before building them.')
+        if testentry.endswith('.jar'):
             return ['java', '-jar'] + self.test.fname
-        elif not self.test.is_cross_built and run_with_mono(self.test.fname[0]):
```

CC: @eli-schwartz 

- Andres

eli-schwartz:
In this case, the test fname might have an implicit extension and cannot be found by `os.path.isfile()` -- intead use `shutil.which()` since it is guaranteed to find anything that subprocess.Popen or asyncio.create_subprocess_exec can successfully run.

Fixes #10027

### DIFF
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -224,6 +224,7 @@ class TestSerialisation:
     protocol: TestProtocol
     priority: int
     cmd_is_built: bool
+    cmd_is_exe: bool
     depends: T.List[str]
     version: str
     verbose: bool
@@ -1147,6 +1148,7 @@ class Backend:
                                    t.is_parallel, cmd_args, t.env,
                                    t.should_fail, t.timeout, t.workdir,
                                    extra_paths, t.protocol, t.priority,
+                                   isinstance(exe, build.Target),
                                    isinstance(exe, build.Executable),
                                    [x.get_id() for x in depends],
                                    self.environment.coredata.version,
--- a/mesonbuild/mtest.py
+++ b/mesonbuild/mtest.py
@@ -1349,18 +1349,18 @@ class SingleTestRunner:
 
     def _get_test_cmd(self) -> T.Optional[T.List[str]]:
         testentry = self.test.fname[0]
-        if self.options.no_rebuild and not os.path.isfile(testentry):
+        if self.options.no_rebuild and self.test.cmd_is_built and not os.path.isfile(testentry):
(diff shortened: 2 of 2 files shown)


eli-schwartz:
/cc @anarazel

@nirbheek this is a prerequisite for #10008 being backported to 0.61.3

jpakkane (approved):
LGTM once someone verifies that this fixes the issue.

eli-schwartz:
It fails on Linux because shutil.which cannot find jar files, I was afraid of that @anarazel.

anarazel:
@eli-schwartz Just tested, and can confirm that this PR fixes the bug (although I needed to rebuild from scratch to avoid "AttributeError: 'TestSerialisation' object has no attribute 'cmd_is_exe'" - but that's afaict ok for the development branch)

eli-schwartz:
Yep. The attribute error just means that the internal build data representation has changed structure, normally we'd just throw an error if you use a different version of meson than the one that originally generated the builddir. "Error: different versions of meson used for generating the builddir and for trimming to run this command right now, you can't do that, please wipe and regenerate" or thereabouts.

eli-schwartz:
... aaaand since it changes the format, we cannot backport it to 0.61.3, yeah.

PR #10114, opened 2022-03-11
fix regression in vcs_tag when the VCS program is not installed

eli-schwartz:
We are supposed to fallback on the fallback when running the vcstagger, but instead we errored out during configure.

Fixes regression in commit b402817fb6f0392812bfa272bdbc05c9c30139fa. Before this, we used shutil.which || relative paths, and in the latter case if it could not be found we still wrote out that path but it failed to run in vcstagger. Now, we use find_program under the hood, so it needs to be run in non-fatal mode, and if it is not found, we simply keep the original command string. It's a VCS command, so if we magically end up finding it at runtime because it was installed after running configure, that is *fine*.


eli-schwartz:
Thanks to @helmutg for running test builds against the Debian archive, and uncovering this issue.

dcbaker (approved):

dcbaker (approved):

PR #10115, opened 2022-03-11
D compiler checks

rtbo:
Implement and test a few compiler checks for D:
 - run
 - sizeof
 - has_header
 - alignment

May be a function to test built-in D versions would be useful too, but that would be specific to D.
Something like:
```meson
meson.get_compiler('d').has_version('LittleEndian')
```

### DIFF
--- a/docs/markdown/snippets/d_compiler_checks.md
+++ b/docs/markdown/snippets/d_compiler_checks.md
@@ -1,0 +1,13 @@
+## D compiler checks
+
+Some compiler checks are implemented for D:
+ - `run`
+ - `sizeof`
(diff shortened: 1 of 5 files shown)


dcbaker (changes requested):
This probably deserves a release snippet, otherwise this looks good to me.

dcbaker (approved):

dcbaker (approved):

eli-schwartz:
Thinking more about this... I'm not sure what the best way to do it is, or even if we *can* do it at that level, but ideally we'd want a FeatureNew for this somehow? Can't think of a feasible way to return one from the individual compiler though.

PR #10117, opened 2022-03-13
give subproject interpreters a view of user defined options

eli-schwartz:
We need this for the python module, as implemented in commit e8375d20a9aeb8c3b0ad58f299ded0e5e978b447, but that then crashed in subprojects because those options were never forwarded to the subproject interpreter.


eli-schwartz:
Fixes a regression that broke the wrapdb CI for e.g. glib.

PR #10119, opened 2022-03-14
gnome module: fix crash due to misused function while generating gir command

eli-schwartz:
In commit 68e684d51f1e469e0d9f4b499ffda15146cad98a the _get_link_args function was modified from returning a list[str] of arguments, to a tuple of both that and a modified copy of the entire target's current/enhanced dependencies (why not just the new ones? I don't know).

However, the existing use of the function was not adapted to this change, and tried to turn this entire tuple into a node of the command line. Tuples cannot flatten to lists, and mesonlib.File or HoldableObjects don't make good command line arguments.

As a result we errored out with:

```
ERROR: Argument (['-L/path/to/builddir/', '--extra-library=foo'], [<SharedLibrary 25a6634@@foo@sha: foo>, <SharedLibrary 25a6634@@foo@sha: foo>, <SharedLibrary 25a6634@@foo@sha: foo>]) in "command" is invalid
```

Split out the flags and the dependencies and update the former while replacing the latter.

### DIFF
--- a/mesonbuild/modules/gnome.py
+++ b/mesonbuild/modules/gnome.py
@@ -1132,9 +1132,9 @@ class GnomeModule(ExtensionModule):
         scan_command += ['-I' + srcdir, '-I' + builddir]
         scan_command += state.get_include_args(girtargets_inc_dirs)
         scan_command += ['--filelist=' + self._make_gir_filelist(state, srcdir, ns, nsversion, girtargets, libsources)]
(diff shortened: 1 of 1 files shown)


eli-schwartz:
Thanks to @helmutg for running test builds against the Debian archive, and uncovering this issue.

dcbaker (approved):
Ugggh, the gnome module has so many side effects that make changes so hard to reason about

PR #10122, opened 2022-03-14
Store include and compile arguments separately in non-internal Dependencies

dcbaker:
Currently, `InternalDependency` stores include directories separately from other compile specific flags, but all other dependency types combine them. This is problematic for any code wanting to get only the include paths, specifically `dependency.partial_dependency()`, which when used with `dependency.partial_dependency(includes : true)` only works with a dependency declared with `declare_dependency`.

This series attempts to change the other dependencies to work the same way. I say attempts because there's a lot of code that doesn't make any effort to separate the flags conceptually. This *should* work for pkg-config, which is the majority of the dependency usage. I think this works for cmake as well, which makes up the second largest chunk of dependencies. System and Builtin Dependency should work as well, and I think that `ConfigTool` dependencies work in general, though some of them may not work. Boost is the biggest black hole to me, as none of the tests work on nixos, and I haven't been properly motivated to sort them out.


dcbaker:
Rebased and passing all tests locally (Only on Linux, and on nix some of the tests fail regardless, so there still might be some issues here)

docs/markdown/snippets/dependency_include_dirs.md
```diff
@@ -0,0 +1,8 @@
+## All dependencies now separate include directory and compile args
+
+Prior to 1.2.0, dependencies declared with `declare_dependency` would separate
+-I/-isystem into separate fields, but external dependencies would combine them
+as raw arguments. This meant that using `dependency.partial_dependency(includes
+: True)` would not work as expected with external dependencies, and made the
```
tristan957, line 6: this should be `true`

docs/yaml/objects/dep.yaml
```diff
@@ -136,6 +136,9 @@ methods:
     warnings:
       - A bug present until 0.50.1 results in the above behavior
         not working correctly.
+      - Due to implementation details, using "includes" with external
+        dependencies would not work prior to 1.2.0, use compile_args
+        to get includes for meson < 1.2.0
```
tristan957, line 141: Perhaps capitalize meson

mesonbuild/backend/ninjabackend.py
```diff
@@ -2734,10 +2734,14 @@ def generate_llvm_ir_compile(self, target, src):
...
-    def generate_inc_dir(self, compiler: 'Compiler', d: str, basedir: str, is_system: bool) -> \
+    def generate_inc_dir(self, compiler: 'Compiler', d: str, basedir: T.Optional[str], is_system: bool) -> \
             T.Tuple['ImmutableListProtocol[str]', 'ImmutableListProtocol[str]']:
+        # If the basedir is None, this is a system provided path
+        if basedir is None:
+            return (compiler.get_include_args(d, is_system), [])
+
         # Avoid superfluous '/.' at the end of paths when d is '.'
-        if d not in ('', '.'):
+        if d not in {'', '.'}:
```
tristan957, line 2744: Is this a candidate for frozenset?

mesonbuild/build/include_dirs.py
```diff
@@ -0,0 +1,73 @@
...
+import os
+import typing as T
+
+from ..mesonlib import HoldableObject
+
+
+@dataclass(eq=False)
+class IncludeDirs(HoldableObject):
+
+    """Internal representation of an paths to be treated as compiler include
```
tristan957, line 16: Remove `an`

mesonbuild/build/include_dirs.py
```diff
@@ -0,0 +1,73 @@
...
+        if self.curdir is None:
+            return self.incdirs
+        for idir in self.incdirs:
+            strlist.append(os.path.join(sourcedir, self.curdir, idir))
+            if self.curdir and builddir:
+                strlist.append(os.path.join(builddir, self.curdir, idir))
+        return strlist
+
+    def to_system(self) -> IncludeDirs:
+        """Create a shallow copy of this IncludDirs as a system dependency."""
```
tristan957, line 68: IncludeDirs. Also in the next doc comment

test cases/common/183 partial dependency/declare_dependency/meson.build
```diff
@@ -30,3 +30,5 @@ dec_exe = executable(
 )
 
 test('Declare Dependency', dec_exe)
+
```
tristan957, line 33: Probably should remove these changes.

tristan957 (commented):
Here is a pretty crappy review that catches simple mistakes.

.gitignore
```diff
@@ -24,6 +24,7 @@ __pycache__
 packagecache
 /MANIFEST
 /build
+!mesonbuild/build
```
tristan957, line 27: I don't think is actually necessary if my gitignore knowledge is right.

mesonbuild/backend/ninjabackend.py
```diff
@@ -2734,10 +2734,14 @@ def generate_llvm_ir_compile(self, target, src):
         return (rel_obj, rel_src)
 
     @lru_cache(maxsize=None)
-    def generate_inc_dir(self, compiler: 'Compiler', d: str, basedir: str, is_system: bool) -> \
+    def generate_inc_dir(self, compiler: 'Compiler', d: str, basedir: T.Optional[str], is_system: bool) -> \
```
tristan957, line 2737: If you're here, might as well `from __future__ import annotations`.

mesonbuild/build/include_dirs.py
```diff
@@ -40,9 +40,14 @@ def to_string_list(self, sourcedir: str, builddir: T.Optional[str] = None) -> T.
         :returns: A list of strings (without compiler argument)
         """
         strlist: T.List[str] = []
+        # if curdir is `None`, then we are dealing with some kind of external
```
tristan957, line 56: Capitalize if

mesonbuild/build/include_dirs.py
```diff
@@ -40,9 +40,14 @@ def to_string_list(self, sourcedir: str, builddir: T.Optional[str] = None) -> T.
...
+        # dependency include paths, and we don't want to make them relative to
+        # the source or builddir, otherwise we do.
```
tristan957, line 58: `builddir. Otherwise, we do.`

mesonbuild/build/include_dirs.py
```diff
@@ -13,7 +13,20 @@
 @dataclass(eq=False)
 class IncludeDirs(HoldableObject):
 
-    """Internal representation of an include_directories call."""
+    """Internal representation of an paths to be treated as compiler include
+    paths.
+
+    the `include_directories` function is lowered into this, and it is used
```
tristan957, line 19: Capitalize The

mesonbuild/build/include_dirs.py
```diff
@@ -13,7 +13,20 @@
...
+    internally.
+
+    :param curdir: The current working directory, set to none for system
+        dependencies
+    :param incdirs: a list of paths, either relative to the source dir, or
+        absolute
+    :param is_system: is a system path. Use this if you want these to be used
+        with -`isystem` (or equivalent) when possible
```
tristan957, line 27: The dash should probably be within the backticks

mesonbuild/dependencies/configtool.py
```diff
@@ -78,6 +79,32 @@ def _sanitize_version(self, version: str) -> str:
             return m.group(0).rstrip('.')
         return version
 
+    @staticmethod
+    def _split_include_dirs(args: T.Iterable[str]) -> T.Tuple[T.List[IncludeDirs], T.List[str]]:
+        """Split include and other compile args
+
+        And convert the include arguments into an IncludeDirs object
+
+        :param args: An iterable of strings
```
tristan957, line 88: Perhaps describe it as `include directories`? Otherwise you are just describing the type annotation :)

mesonbuild/dependencies/base.py
```diff
@@ -155,6 +155,15 @@ def get_all_compile_args(self) -> T.List[str]:
         return list(itertools.chain(self.get_compile_args(),
                                     *(d.get_all_compile_args() for d in self.ext_deps)))
 
+    def get_all_include_dirs(self) -> T.Iterable[IncludeDirs]:
+        """Recursively get all include directories from this dependency and it's dependencies
```
tristan957, line 143: its

mesonbuild/dependencies/pkgconfig.py
```diff
@@ -218,11 +219,17 @@ def _set_cargs(self) -> None:
...
             env['PKG_CONFIG_ALLOW_SYSTEM_CFLAGS'] = '1'
-        ret, out, err = self._call_pkgbin(['--cflags', self.name], env=env)
+        ret, out, err = self._call_pkgbin(['--cflags-only-other', self.name], env=env)
         if ret != 0:
             raise DependencyException(f'Could not generate cargs for {self.name}:\n{err}\n')
         self.compile_args = self._convert_mingw_paths(self._split_args(out))
 
+        ret, out, err = self._call_pkgbin(['--cflags-only-I', self.name], env=env)
+        if ret != 0:
+            raise DependencyException(f'Could not generate cargs for {self.name}:\n{err}\n')
```
tristan957, line 229: Should cargs be changed for this exception or left as is? I could see both ways.

mesonbuild/dependencies/scalapack.py
```diff
@@ -147,10 +148,19 @@ def _set_cargs(self) -> None:
             env = os.environ.copy()
             env['PKG_CONFIG_ALLOW_SYSTEM_CFLAGS'] = '1'
         ret, out, err = self._call_pkgbin([
-            '--cflags', self.name,
+            '--cflags-only-other', self.name,
             '--define-variable=prefix=' + self.__mklroot.as_posix()],
             env=env)
         if ret != 0:
-            raise DependencyException('Could not generate cargs for %s:\n%s\n' %
+            raise DependencyException('Could not generate compile arguments for %s:\n%s\n' %
```
tristan957, line 155: Perhaps the same vocabulary should be used above?

mesonbuild/interpreter/interpreterobjects.py
```diff
@@ -511,6 +513,9 @@ def configtool_method(self, args: T.Tuple[str], kwargs: TYPE_kwargs) -> str:
     @noPosargs
     @typed_kwargs('dependency.partial_dependency', *_PARTIAL_DEP_KWARGS)
     def partial_dependency_method(self, args: T.List[TYPE_nvar], kwargs: 'kwargs.DependencyMethodPartialDependency') -> Dependency:
+        if kwargs['includes'] and not isinstance(self.held_object, dependencies.InternalDependency):
```
tristan957, line 516: Seems weird that this is the only thing in this commit. Perhaps you were going to do some squashing?

tristan957 (commented):
This is a really good change.

mesonbuild/backend/ninjabackend.py
dcbaker: Python will optimize an anonymous set like this into a frozenset when it byte compiles the .py files

dcbaker:
So, umm. these msys failures are really intresting. We're getting something like `-I/D/some/path\`, and `shlex` is choking on that `\`, which it sees as an escape with no character.... So, is this a bug in pkgconf, msys, or should Meson be looking for this and fixing it?

tristan957:
I don't use msys2 otherwise I would lend hand.

dcbaker:
It really helps to test the right branch in mingw... Sigh

I can reproduce the error now

dcbaker:
The bug itself seems rather trivial: https://gitea.treehouse.systems/ariadne/pkgconf/pulls/249

Working around it in Meson seems somewhat harder... Sigh

mesonbuild/dependencies/pkgconfig.py
```diff
@@ -210,6 +210,18 @@ def _convert_mingw_paths(self, args: T.List[str]) -> T.List[str]:
     def _split_args(self, cmd: str) -> T.List[str]:
         # pkg-config paths follow Unix conventions, even on Windows; split the
         # output using shlex.split rather than mesonlib.split_args
+
+        # Due to https://gitea.treehouse.systems/ariadne/pkgconf/issues/238
+        # it's possible for us to end up with a escape character that escapes
+        # nothing. This means that a space hasn't been properly escaped, and
+        # things are probably going to fail. However, we can at least not die in
+        # a pile
```
tristan957, line 218: A pile of what?! :)

mesonbuild/dependencies/pkgconfig.py
```diff
@@ -210,6 +210,18 @@ def _convert_mingw_paths(self, args: T.List[str]) -> T.List[str]:
...
+        if 'MSYSTEM' in os.environ and cmd.endswith("\\"):
+            # TODO: possibly pending fix: https://gitea.treehouse.systems/ariadne/pkgconf/pulls/249
+            mlog.warning("pkgconf on msys/mingw doesn't handl escaped spaces correctly, "
+                         "it looks like that has happened. Compilation may fail")
```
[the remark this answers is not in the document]
dcbaker: took me a minute to figure this out, apparently I had fixed the handl -> handle, but not committed it. lol

mesonbuild/build/include_dirs.py
tristan957: Looks like this issue still exists fyi

mesonbuild/build/include_dirs.py
```diff
@@ -13,7 +13,20 @@
...
-    """Internal representation of an include_directories call."""
+    """Internal representation of paths to be treated as compiler include
+    paths.
+
+    The `include_directories` function is lowered into this, and it is used
+    internally.
+
+    :param curdir: The current working directory, set to none for system
+        dependencies
+    :param incdirs: a list of paths, either relative to the source dir, or
```
tristan957, line 24: Probably want to be consistent about capitalization of the first word.

mesonbuild/build/include_dirs.py
```diff
@@ -65,9 +65,9 @@ def to_string_list(self, sourcedir: str, builddir: T.Optional[str] = None) -> T.
         return strlist
 
     def to_system(self) -> IncludeDirs:
-        """Create a shallow copy of this IncludDirs as a system dependency."""
+        """Create a shallow copy of this IncludeDirs as a system dependency."""
```
tristan957, line 68: Oh I see. This fix got folded into the wrong commit.

mesonbuild/backend/ninjabackend.py
eli-schwartz: Also there is no speed difference between sets and frozensets AFAIK -- the difference is that the former has mutating methods and the latter can be hashed / used as dictionary keys.

There is, generally, no real purpose in using frozenset() instead of set() except inasmuch as you care about asserting immutability.

Using set *literals* is useful because it can be compiled in bytecode, but that isn't conditional on its being optimized into specifically a frozenset. It just happens to be an internal implementation detail of python.

.gitignore
eli-schwartz: It's an artifact of setuptools -- `python setup.py build --build-base=build/` is the default build base, it's how setuptools handles build directories.

I do not understand why we need to override an anchored path either, though.

mesonbuild/backend/ninjabackend.py
eli-schwartz: Hmm? This was added last year in 8ae2bf5a9e652d49027191709cda5353446af487.

docs/markdown/snippets/dependency_include_dirs.md
```diff
@@ -0,0 +1,8 @@
+## All dependencies now separate include directory and compile args
+
+Prior to 1.2.0, dependencies declared with `declare_dependency` would separate
+-I/-isystem into separate fields, but external dependencies would combine them
+as raw arguments. This meant that using `dependency.partial_dependency(includes
+: true)` would not work as expected with external dependencies, and made the
+implementation more complicated. As of 1.2.0 this has been changed to separate
```
eli-schwartz, line 7: I don't think we need to mention the implementation complexity in the release notes.

We should just highlight the fact that this change is about making the behavior consistent, so that `dependency()` and "dependency via subproject fallback" don't do different things.

unittests/allplatformstests.py
```diff
@@ -1777,6 +1777,7 @@ def test_underscore_prefix_detection_define(self) -> None:
             raise SkipTest('Did not find the underscore prefix define __USER_LABEL_PREFIX__')
 
     @skipIfNoPkgconfig
+    @skip_if_env_set("MSYSTEM")  # msys/mingw doesn't handle \ escaped correctly
```
eli-schwartz, line 1780: I'm curious what this currently does in CI.

.gitignore
dcbaker: I don't know either, but if you remove this line and run `touch mesonbuild/build/foo.txt` git will not add it and makes you use --force

tristan957:
Ping me when you get the tests passing. I'll give another review.

PR #10128, opened 2022-03-16
python: Avoid annoying warning when prefix is the default /usr/local

xclaesse:
Python does not support installing modules to /usr/local. If user
install their python modules there, it should be expected that they
won't be found by their python interpreter unless they set PYTHONPATH
manually.

The warning was mostly intended for the case where user installs into
/usr and for some reason the proper install path was not correctly
detected, which was the case on Debian in the past.


eli-schwartz:
I disagree with this PR completely.

The warning's rationale is to tell you if you have usage issues which in this case you absolutely do. Either we believe such usage issues matter and we should warn even for /usr/local, or we should never warn under any circumstances, because Debian is fixed and is now properly detected.

xclaesse:
Ok nevernind, let's remove the warning completely.

PR #10129, opened 2022-03-16
python: Remove warning about invalid install path

xclaesse:
It was originally added because proper detection was not working on
Debian, but that has been fixed since. It was causing annoying warning
by default when prefix is /usr/local that can only be avoided by setting
options.


eli-schwartz:
https://github.com/mesonbuild/meson/pull/9788#issuecomment-1019582550

I'd be okay with this change. :)

/cc @tristan957 @rgommers

tristan957:
Yes, please remove this. It's horrible! 🙂

rgommers:
Thank you @xclaesse and @eli-schwartz! It'd be great to remove this warning indeed.

PR #10130, opened 2022-03-16
coredata: be even more robust on unpickling errors

bonzini:
When reverting from 0.62 to 0.59, one can see an error like this:
    
    line 1003, in load
        obj = pickle.load(f)
      File "/Users/pm215/src/qemu-for-merges/meson/mesonbuild/mesonlib/universal.py",
    line 2076, in __setstate__
        self.__init__(**state)  # type: ignore
    TypeError: __init__() got an unexpected keyword argument 'module'
    FAILED: build.ninja
    
Raise a MesonException for TypeError as well, so that reconfiguration
proceeds using cmd_line.txt.

### DIFF
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -1036,7 +1036,7 @@ def load(build_dir: str) -> CoreData:
             obj = pickle.load(f)
     except (pickle.UnpicklingError, EOFError):
         raise MesonException(load_fail_msg)
-    except (ModuleNotFoundError, AttributeError):
+    except (TypeError, ModuleNotFoundError, AttributeError):
         raise MesonException(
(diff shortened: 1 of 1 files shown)


xclaesse:
@bonzini this PR has many unrelated commits, is that a mistake?

bonzini:
Yes, looks like my previous PR was rebased-and-applied. Fixed now.

xclaesse:
@bonzini Thanks, much better. LGTM.

@jpakkane I think this can go for 0.62.0, it's a bugfix IMHO. Ok?

eli-schwartz (approved):
This is absolutely a stable-worthy bugfix, LGTM and if 0.61.3 wasn't already out I'd say backport it there too.

PR #10132, opened 2022-03-16
fix detection of language standard library paths

bonzini:
The code in the C++ and Fortran compilers' `language_stdlib_only_link_flags` method is broken and cannot possibly have ever worked.  Instead of splitting by line, it splits by whitespace and therefore, instead of the last line of the compiler output:

 programs: =/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
libraries: =/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/12.0.0

it is only the last field that has its first 11 characters removed. Instead of reinventing the wheel with a new and brittle pattern, reuse get_compiler_dirs.

Fixes: 64c267c49 ("compilers: Add default search path stdlib_only_link_flags", 2021-09-25)
Signed-off-by: Paolo Bonzini <pbonzini@redhat.com>

### DIFF
--- a/mesonbuild/compilers/cpp.py
+++ b/mesonbuild/compilers/cpp.py
@@ -251,11 +251,9 @@ class ClangCPPCompiler(ClangCompiler, CPPCompiler):
         # be passed to a different compiler with a different set of default
         # search paths, such as when using Clang for C/C++ and gfortran for
         # fortran,
-        search_dir = self._get_search_dirs(env)
         search_dirs: T.List[str] = []
-        if search_dir is not None:
-            for d in search_dir.split()[-1][len('libraries: ='):].split(':'):
-                search_dirs.append(f'-L{d}')
+        for d in self.get_compiler_dirs(env, 'libraries'):
+            search_dirs.append(f'-L{d}')
         return search_dirs + ['-lstdc++']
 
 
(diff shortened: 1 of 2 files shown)


dcbaker (approved):
It absolutely works because without it you can't mix Fortran and C or C++ on nix, the `fortran/9 cpp`, for example. So it *does* solve the issue it was created for. This is, however, a better approach.

bonzini:
> It absolutely works because without it you can't mix Fortran and C or C++ on nix, the fortran/9 cpp, for example. So it does solve the issue it was created for.

It works if the search directory is not the first in the output of -print-search-dirs. And also every "Linking" step on macos produces a warning like:

ld: warning: directory not found for option
'-Lns/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/12.0.0'

So it should be considered for stable releases in 0.61.4 will be a thing.

dcbaker:
Yeah, nix injects dozens and dozens, and the first directory is useless. But it does make mixing languages on nix possible, which was not without it.

eli-schwartz:
If the commit message is so wrong, shouldn't that have been fixed before merging?

dcbaker:
I mean, it's wrong and it isn't wrong. It's wrong in saying "this never could have possibly worked" but it's right in saying "there are bugs". It solved the specific issue I had, ie, that nix separates the default include dirs of gcc and gfortran. On the other hand, it doesn't help MacOS

Issue #10080, opened 2022-03-07
Building scipy with PyPy fails on windows


mattip:
The [error](https://github.com/mattip/scipy/runs/5450598923?check_suite_focus=true#step:9:57) is 
> Could not find Python3 library 'c:\\hostedtoolcache\\windows\\pypy\\3.8.12\\x86\\python38.dll'

The correct dll name is `libpypy3-c.dll`, but shouldn't the compiler be MSVC and the library it is looking for be the MSVC import library which is `c:\\hostedtoolcache\\windows\\pypy\\3.8.12\\x86\\libs\python3.8.lib` (like CPython)?

tristan957:
Is this fixed with the PR that got merged?

mattip:
It seems not. I [reset the PR](https://github.com/mattip/scipy/runs/5461997995?check_suite_focus=true#step:7:83) to use meson HEAD, and am still getting the same error (scroll down to the bottom of that log).

I am willing to submit a patch to fix this, but would like to reproduce locally. Is there a guide to running tests? I tried on windows 10 with
```
git clone https://github.com/mesonbuild/meson
cd meson
python run_tests.py
```

Where `python` is either CPython 3.8 or PyPy 3.8, and in both cases I get many dialog boxes with the error message 
> Entry Point Not Found : The procedure entry point _ZNSt13runtime_erroraSEOS_ could not be located in the dynamic link library C:\ProgramData\chocolatey\lib\mingw\tools\install\mingw64\bin\ccache.exe

even though that mingw tools set is not on my path.

eli-schwartz:
If CC/CXX are not set to override the compiler, the automatic detection logic will try to run `ccache --version` to see if it succeeds, then prepend it to the real compiler if it did.

So I'd suspect that for each unittest it would perform that check and fail if it found a broken ccache.exe (possibly with a Windows error dialog box explaining exactly what is broken about it) but I do not know how that could happen when that directory is not in your PATH...

mattip:
right, sorry for the noise. That ccache.exe is broken.

eli-schwartz:
> Is this fixed with the PR that got merged?

As mentioned in #8570 the two issues are different, because that was originally reported on Linux and is about link_libpython being false on Linux, but still leading to detection failure.

On Windows, symbols need to be resolved in built extensions, so link_libpython is true, which means that the PyPy link library does need to be found there. Hence detection failure is actually correct right now.

eli-schwartz:
[...]

PR #10134, opened 2022-03-17
unittests: convert python tests to project tests

eli-schwartz:
Perhaps when this test case was originally created, project tests could not use a matrix of options? This is certainly possible today, so don't write special unittest handling for this instead.

This adds proper visibility into what gets run and what doesn't. Now we know which python executables got tested and which got skipped.


eli-schwartz:
Motivated by #10080 since it seems to be quite hard to tell whether pypy actually got checked, which means we spent entirely too much time being very confused.

/cc @mattip

dcbaker (approved):

PR #10136, opened 2022-03-17
wrap: use shared function to connect to wrapdb with better errors

eli-schwartz:
We currently inconsistently handle connection, `has_ssl`, and printing errors on urlopen failure between `meson subprojects` and `meson wrap`.

Make the latter work more like the former.

### DIFF
--- a/mesonbuild/msubprojects.py
+++ b/mesonbuild/msubprojects.py
@@ -132,9 +132,9 @@ class Runner:
             branch, revision = wraptool.parse_patch_url(patch_url)
         except WrapException:
             return
-        new_branch, new_revision = wraptool.get_latest_version(self.wrap.name)
+        new_branch, new_revision = wraptool.get_latest_version(self.wrap.name, self.options.allow_insecure)
         if new_branch != branch or new_revision != revision:
-            wraptool.update_wrap_file(self.wrap.filename, self.wrap.name, new_branch, new_revision)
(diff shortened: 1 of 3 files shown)


xclaesse:
Should we really allow unauthenticated downloads? When downloading tarballs it's not too bad because the wrap contains a checksum, but when downloading wraps from wrapdb we must be sure we really are talking to wrapdb.mesonbuild.com otherwise we could receive a hacked .wrap file that itself contains altered checksums.

I personally think it's 2022, if you don't have SSL then it's your problem.

eli-schwartz:
I have no clue why we ever try to handle that TBH. But it only happens when your actual *python installation* is built without the `import ssl` module.

My original motivation for this change was actually about catching URLError and raising this instead:
```
WrapException(f'WrapDB connection failed to {urlstring} with error {excp}')
```

Currently it fails with a 31-line traceback that doesn't even mention the url...

xclaesse:
Maybe we should have `meson wrap install foo --insecure` like curl, or `--no-check-certificate` like wget to make sure it is user's choice? I can see a need to ignore SSL when it's done in harmless CI job that might not have proper CA database, etc.

eli-schwartz:
I do understand the caution about allowing non-SSL connections.

Your suggestion about a new option makes sense, I'll add one. :)

xclaesse:
But totally +1 on improving error message, I'm just not confortable at all with the fallback to plain http we have in open_wrapdburl().

PR #10137, opened 2022-03-17
tests: don't skip python version tests if python is found but its dep is broken [merged but github broke right then]

eli-schwartz:
If a version of python is installed for testing against, we should assume it's actually important to test against it.

### DIFF
--- a/test cases/python/8 different python versions/meson.build
+++ b/test cases/python/8 different python versions/meson.build
@@ -6,18 +6,14 @@ py_mod = import('python')
 py = py_mod.find_installation(get_option('python'), required : false)
 
 if py.found()
-  py_dep = py.dependency(required : false)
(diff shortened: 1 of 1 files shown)


mattip:
Thanks. Using this branch I now see the error message.

xclaesse:
If this pass CI then sure LGTM.

eli-schwartz:
Well, I should say that I *did* merge it. But github is not actually marking this PR as such or closing it with the "merged" label (or even closing it as "closed").

This would go along with the 500 internal server errors when trying to push branches, and "you can't comment at this time" when trying to post comments.

But I must say, refusing to close a PR is a new kind of error for me.

eli-schwartz:
And it turns out one of the brownouts actually does fail, see #10144

PR #10138, opened 2022-03-17
python module: windows dll name for pypy needs special casing

mattip:
Fix one source of problems when using meson with PyPy3: the name of the DLL. 

Along the way, fix a file descriptor being left open.

### DIFF
--- a/.github/workflows/msys2.yml
+++ b/.github/workflows/msys2.yml
@@ -65,6 +65,8 @@ jobs:
             git
             mercurial
             lcov
+            wget
+            unzip
             mingw-w64-${{ matrix.MSYS2_ARCH }}-cmake
             mingw-w64-${{ matrix.MSYS2_ARCH }}-glib2
             mingw-w64-${{ matrix.MSYS2_ARCH }}-libxml2
@@ -81,8 +83,26 @@ jobs:
         run: |
           python3 -m pip --disable-pip-version-check install gcovr jsonschema pefile pytest pytest-subtests pytest-xdist coverage codecov
 
+      - name: Install pypy3 on x86_64
+        run: |
+          mkdir pypy3local
+          pushd pypy3local
+          wget -nv https://downloads.python.org/pypy/pypy3.8-v7.3.9-win64.zip
+          unzip pypy3.8-v7.3.9-win64.zip
+          mv pypy3.8-v7.3.9-win64/* ./
+          popd
+        if: ${{ matrix.MSYS2_ARCH == 'x86_64' }}
+
       - name: Run Tests
         run: |
+          if [[ "${{ matrix.MSYS2_ARCH }}" == "x86_64" ]]; then
+            # There apparently is no clean way to add to the PATH in the 
+            # previous step?
+            # See for instance https://github.com/msys2/setup-msys2/issues/171
+            export PATH=$PATH:$PWD/pypy3local
+            # Make sure it is on the PATH
+            pypy3 -c "import sys; print(sys.version)"
+          fi
           export BOOST_ROOT=
           export PATHEXT="$PATHEXT;.py"
 
--- a/ci/run.ps1
+++ b/ci/run.ps1
@@ -54,18 +54,26 @@ echo "Extracting ci_data.zip"
 Expand-Archive $env:AGENT_WORKFOLDER\ci_data.zip -DestinationPath $env:AGENT_WORKFOLDER\ci_data
 & "$env:AGENT_WORKFOLDER\ci_data\install.ps1" -Arch $env:arch -Compiler $env:compiler -Boost $true -DMD $dmd
 
+if ($env:arch -eq 'x64') {
+    DownloadFile -Source https://downloads.python.org/pypy/pypy3.8-v7.3.9-win64.zip -Destination $env:AGENT_WORKFOLDER\pypy38.zip
+    Expand-Archive $env:AGENT_WORKFOLDER\pypy38.zip -DestinationPath $env:AGENT_WORKFOLDER\pypy38
+    $ENV:Path = $ENV:Path + ";$ENV:AGENT_WORKFOLDER\pypy38\pypy3.8-v7.3.9-win64;$ENV:AGENT_WORKFOLDER\pypy38\pypy3.8-v7.3.9-win64\Scripts"
+    pypy3 -m ensurepip
+}
+
 
 echo "=== PATH BEGIN ==="
 echo ($env:Path).Replace(';',"`n")
 echo "=== PATH END ==="
 echo ""
 
-$progs = @("python","ninja","pkg-config","cl","rc","link")
+$progs = @("python","ninja","pkg-config","cl","rc","link","pypy3")
 foreach ($prog in $progs) {
   echo ""
   echo "Locating ${prog}:"
   where.exe $prog
 }
+
 
 echo ""
 echo "Ninja / MSBuld version:"
--- a/mesonbuild/modules/python.py
+++ b/mesonbuild/modules/python.py
@@ -186,12 +186,20 @@ class PythonSystemDependency(SystemDependency, _PythonDependencyBase):
     def _get_windows_link_args(self) -> T.Optional[T.List[str]]:
         if self.platform.startswith('win'):
             vernum = self.variables.get('py_version_nodot')
+            verdot = self.variables.get('py_version_short')
+            imp_lower = self.variables.get('implementation_lower', 'python')
             if self.static:
                 libpath = Path('libs') / f'libpython{vernum}.a'
             else:
                 comp = self.get_compiler()
                 if comp.id == "gcc":
-                    libpath = Path(f'python{vernum}.dll')
+                    if imp_lower == 'pypy' and verdot == '3.8':
+                        # The naming changed between 3.8 and 3.9
+                        libpath = Path(f'libpypy3-c.dll')
+                    elif imp_lower == 'pypy':
+                        libpath = Path(f'libpypy{verdot}-c.dll')
+                    else:
+                        libpath = Path(f'python{vernum}.dll')
                 else:
                     libpath = Path('libs') / f'python{vernum}.lib'
             # base_prefix to allow for virtualenvs.
--- a/run_project_tests.py
+++ b/run_project_tests.py
@@ -651,7 +651,8 @@ def _run_test(test: TestDef,
     (returncode, stdo, stde) = run_configure(gen_args, env=test.env, catch_exception=True)
     try:
         logfile = Path(test_build_dir, 'meson-logs', 'meson-log.txt')
-        mesonlog = logfile.open(errors='ignore', encoding='utf-8').read()
+        with logfile.open(errors='ignore', encoding='utf-8') as fid:
+            mesonlog = fid.read()
     except Exception:
         mesonlog = no_meson_log_msg
     cicmds = run_ci_commands(mesonlog)


mattip:
Related to #10080

eli-schwartz:
Thanks. I'd probably prefix the commit message with `python module: ` though. :)

mattip:
It is a shame there is no sysconfig.get_config_var('LDLIBRARY') like on linux, macOS

mattip:
I changed the commit and the PR title

mattip:
I am running the scipy build with this now at https://github.com/mattip/scipy/actions

mattip:
While this is a step forward and the compilation seems like it succeeds, the actual artifacts [cannot be imported to python](https://github.com/mattip/scipy/runs/5585859996?check_suite_focus=true#step:15:42). Do you have an intuition what might be going on?

mattip:
The error is
```
    from . import _ufuncs
E   ImportError: %1 is not a valid Win32 application
```

I think that means it requires a dependent `dll` that cannot be found

eli-schwartz:
One possible debug method would be to run the compilation with `ninja -v` instead of `ninja`, and then compare the compilation commands used to the commands which successfully compile a working extension, e.g. using distutils.

Might be a bit easier to spot when using a stripped down example project, rather than all of scipy...

mattip:
Agreed that it would be nice to have CI testing PyPy. The meson build exposed an issue with the way PyPy defined `PyTypeObject` (for all platforms): it is not PEP-3123 (strict-aliasing) compliant. It turns out this is a bit tricky to solve in the cpyext C-API emulation layer: the forward references confuse the simple C parser we are using. I opened an [issue](https://foss.heptapod.net/pypy/pypy/-/issues/3706) and will come back to this once that is solved.

eli-schwartz:
Without knowing all that much about PyPy... the linked issue was closed as implemented: "the py3-pep-3123 branch was merged".

Does that mean that we can now go forward with this ticket? I guess it can be tested against PyPy nightly, right?

mattip:
A scipy build does succeed with this PR but ...

I still can't figure out what is wrong with the `pyd` file produced by the build, as in [the comment](https://github.com/mesonbuild/meson/pull/10138#issuecomment-1070950555) above. I have uploaded the bad pyd to my onedrive [here](https://1drv.ms/u/s!AjItDERGqi8Ajh-PPJzirVCi3ujy?e=wJvtl9), and the parallel good one produced by CPython [here](https://1drv.ms/u/s!AjItDERGqi8Ajh6NymZ0402PpyTf?e=SvxtnC). I don't see anything significant when looking at `depends.exe`: just the normal missing support dlls. Even opening the `pyd` with `ctypes.CDLL(<fullpath>)` shows the problem: the CPython one gives a normal `FileNotFound` error (as expected since the support DLLs are missing), where the PyPy one errors out with a `not a valid Win32 application`.

mattip:
Of the 111 `pyd` files built for scipy, only these 4 give this strange `WindowsError [193]`, all in `scipy\special`: 
```
scipy\special\cython_special.pypy38-pp73-win_amd64.pyd
scipy\special\_test_round.pypy38-pp73-win_amd64.pyd
scipy\special\_ufuncs.pypy38-pp73-win_amd64.pyd
scipy\special\_ufuncs_cxx.pypy38-pp73-win_amd64.pyd
```
The other `pyd` files in that directory seem to work normally: 
```
_comb.pypy38-pp73-win_amd64.pyd
_ellip_harm_2.pypy38-pp73-win_amd64.pyd
_specfun.pypy38-pp73-win_amd64.pyd
```

mattip:
I would propose to merge this as-is. I think it would be better to persue the scipy/special problems there, maybe this is some interaction between cython, windows, and PyPy.

eli-schwartz:
Alright, that sounds reasonable enough. As discussed above, can we get PyPy added to the CI images to make sure this doesn't regress in the future?

mattip:
> get PyPy added to the CI images ...

If I understand the CI system, this would mean adding a ubuntu-pypy image to `ci/ciimage` with an `install.sh` that would install pypy3 and make sure it is called on the path as `python3`. Is that correct?

eli-schwartz:
We usually use images that run on different OSes or compilers or options that make deep changes to the entire testsuite system, like running all tests with unity builds. That already has us at 35 images. :o

If the intention is to make sure that Meson can be run using PyPy, then we would want to add a new image and check that meson generally works using PyPy, including hopefully full test coverage.

If the intention is to make sure that Meson (running on the default python) can build projects with PyPy as the detected python dependency, then just installing it on one or more existing images as an ordinary testsuite dependency would be the route to take.

This PR is really about enabling the latter use case. That being said, if you're interested in the former I would not be averse to adding a new image. (I really have no idea under what circumstances it's important to test both CPython and PyPy. If you believe it's important then I guess you're the expert here.)

mattip:
> installing it on one or more existing images as an ordinary testsuite dependency

This seems more sustainable. I will try to add it here

mattip:
I can figure out how I can add PyPy as `/usr/bin/pypy3` in the [bionic image install script](https://github.com/mesonbuild/meson/blob/master/ci/ciimage/bionic/install.sh), but I am a bit lost how to get from there to running unit tests with it. Any hints? For instance, in that script I see that both python3.6 and python3.7 are installed, but I don't see where that shows up in the [test run](https://github.com/mesonbuild/meson/runs/5585782859?check_suite_focus=true) of Ubuntu bionic.

eli-schwartz:
The test run is only supposed to test `python3`.

Python 3.7 is the target python, and but bionic still ships the EOL 3.6 so we hack around that. In theory there is no reason meson cannot build extension modules for python 3.6 but you cannot run meson itself like that.

We cannot actually get rid of 3.6 from bionic since the OS uses it, but we don't attempt to test it.

mattip:
> The test run is only supposed to test python3.

OK. How can I "install PyPy on one or more existing images as an ordinary testsuite dependency" and have it run the tests?

eli-schwartz:
Sorry, that was confusing of me. I meant that as far as CPython is concerned we aren't testing versioned releases of python.

The tests are defined here: https://github.com/mesonbuild/meson/blob/master/test%20cases/python/8%20different%20python%20versions/test.json

It's a matrix of versions, we don't test "python3.7" just "python3" and "python2".

We also test "pypy3" and "pypy" but those are currently raising `SkipTest` since they aren't installed. My assumption is just adding it to the image will cause that to switch from status SKIPPED to status PASSED (or FAILED) but I would eyeball the PR's CI logs to make sure of it.

Note that changes to the CI images trigger the "CI image builder" workflow to build and test the updated images, which currently fails on opensuse and archlinux IIRC due to gnome bugs inside old versions of docker used in GitHub infrastructure. As known failures, we will simply ignore those. Sorry about that.

eli-schwartz:
Sorry, that was confusing of me. I meant that as far as CPython is concerned we aren't testing versioned releases of python.

The tests are defined here: https://github.com/mesonbuild/meson/blob/master/test%20cases/python/8%20different%20python%20versions/test.json

It's a matrix of versions, we don't test "python3.7" just "python3" and "python2".

We also test "pypy3" and "pypy" but those are currently raising `SkipTest` since they aren't installed. My assumption is just adding it to the image will cause that to switch from status SKIPPED to status PASSED (or FAILED) but I would eyeball the PR's CI logs to make sure of it.

Note that changes to the CI images trigger the "CI image builder" workflow to build and test the updated images, which currently fails on opensuse and archlinux IIRC due to gnome bugs inside old versions of docker used in GitHub infrastructure. As known failures, we will simply ignore those. Sorry about that.

eli-schwartz:
BTW: Would it also be possible to add it to azure-pipelines.yml so we can get some Windows coverage?

ci/ciimage/bionic/install.sh
```diff
@@ -58,6 +58,22 @@ update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.7 2 \
 
 python3 -m pip install -U "${base_python_pkgs[@]}" "${python_pkgs[@]}"
 
+pushd /opt
+# Download and install PyPy3.8, and link it to /usr/bin/pypy3
+# At some point it would be more robust to download and parse
+# https://downloads.python.org/pypy/versions.json
+wget https://downloads.python.org/pypy/pypy3.8-v7.3.9-linux64.tar.bz2
```
eli-schwartz, line 65: Is there e.g. some sort of "latest" style url download available? That might save parsing a json file I suppose.

mattip: Adding a symlink to a https://downloads.python.org/pypy/pypy3.8-latest-linux64.tar.bz2 is a nice idea. Maybe if we move to signed tarballs, then you could verify the signature.

ci/ciimage/bionic/install.sh
```diff
@@ -58,6 +58,22 @@ update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.7 2 \
...
+pypy_sha256="08be25ec82fc5d23b78563eda144923517daba481a90af0ace7a047c9c9a3c34"
+if [ $pypy_sha256 != $(sha256sum pypy3.8-v7.3.9-linux64.tar.bz2 | cut -f1 -d" ") ]; then
+  echo bad sha256 for PyPy
+  exit -1
+fi
+tar -xf pypy3.8-v7.3.9-linux64.tar.bz2
+pypy3.8-v7.3.9-linux64/bin/pypy3 -m ensurepip
+popd
+ln -s /usr/bin/pypy3 /opt/pypy3.8-v7.3.9-linux64/bin/pypy3
```
eli-schwartz, line 74: I am prone to forgetting how this works too. Here's what I came up with to remember:

The `ln` command is similar to the `cp` command. First you list the thing you want to install, in this case a string pointing to a destination, and then you list the location you want to install it *to* (by writing the first string as a symlink object).

So `ln -s foo bar` creates a symlink which *finds* foo, and the symlink is installed to bar.

` ln -s /usr/bin/pypy3 /opt/pypy3.8-v7.3.9-linux64/bin/pypy3` creates a symlink which *finds* /usr/bin/pypy3, and it installs the symlink to /opt/pypy3.8-v7.3.9-linux64/bin/pypy3

mattip: Thanks. I added 5 lines and needed 4 commits to get them going :(. Sorry for the wasted time.

eli-schwartz: No big deal, I've needed far more than just 4 commits for far fewer than 5 lines. Also, you did several of those commits while I wasn't looking so I didn't actually end up wasting my time (I guess github wasted their donated CI time but they waste much more than that on a regular basis).

ci/run.ps1
```diff
@@ -54,6 +54,11 @@ echo "Extracting ci_data.zip"
 Expand-Archive $env:AGENT_WORKFOLDER\ci_data.zip -DestinationPath $env:AGENT_WORKFOLDER\ci_data
 & "$env:AGENT_WORKFOLDER\ci_data\install.ps1" -Arch $env:arch -Compiler $env:compiler -Boost $true -DMD $dmd
 
+DownloadFile -Source https://downloads.python.org/pypy/pypy3.8-v7.3.9-win64.zip -Destination $env:AGENT_WORKFOLDER\pypy38.zip
+Expand-Archive $env:AGENT_WORKFOLDER\pypy38.zip -DestinationPath $env:AGENT_WORKFOLDER\pypy38
+$ENV:Path = $ENV:Path + ";$ENV:AGENT_WORKFOLDER\pypy38\pypy3.8-v7.3.9;$ENV:AGENT_WORKFOLDER\pypy38\pypy3.8-v7.3.9\Scripts"
+pypy3 -m ensurepip
```
eli-schwartz, line 60: It seems like azure has a hostedtoolscache version of PyPy already:
https://github.com/actions/virtual-environments/blob/main/images/win/Windows2022-Readme.md#pypy
https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md#pypy

Should we use this instead?

mattip: That should work. We need to add it to the path. Do you know where it is installed?

mattip:
Something is still off with the bionic image, the [pypy tests are skipped](https://github.com/mesonbuild/meson/runs/5876697129?check_suite_focus=true#step:4:1808) on the Ubuntu Bionic run
```
[SKIPPED]   python: 8 different python versions    (python=pypy3)
```

eli-schwartz:
Since that OS comp test needs updates to the CI image itself, we ignore its status and look only at the CI image builder for bionic.

```
[SKIPPED]   python: 8 different python versions    (python=pypy)
Reason: Python not found, skipping test.
 [SUCCESS]   python: 8 different python versions    (python=pypy3)
 [SUCCESS]   python: 8 different python versions    (python=python2)
 [SUCCESS]   python: 8 different python versions    (python=python3)
```

I'm happy with that commit to add pypy3 to the bionic image.

ci/run.ps1
eli-schwartz: Based on:

https://github.com/mesonbuild/meson/blob/e5aa47d8af917282794a8da3956ce444b8d8378d/azure-pipelines.yml#L62-L67

That key for "addToPath" seems like this is handled automatically for us, if you use the Microsoft task for adding a tool.

I assume either there is a similar task type for PyPy, or some version string instead of "3.7" that denotes to use PyPy.

Unfortunately I know very little about the azure pipelines ecosystem.

mattip: right, but that will get us only pypy3 or only python3.7.

mattip: maybe we can add both steps?

One caveat: in the PyPy directory there are a few *.exe files, including python.exe, so we need to make sure the PyPy directory is last on the PATH so it does not override CPython
```
>dir *.exe /b
pypy.exe
pypy3.9.exe
pypy3.9w.exe
pypyw.exe
python.exe
pythonw.exe
```

eli-schwartz: Is it not possible to use this twice and get two versions of a tool into the path? Do they just assume no one ever does that other than once per action in a matrix?

mattip: Looking at the [source of the task](https://github.com/microsoft/azure-pipelines-tasks/blob/master/Tasks/UsePythonVersionV0/usepythonversion.ts#L125), the name seems to suggest it will prepend. So I will try adding the PyPy step before the CPython one, in the hopes that does the RightThing

eli-schwartz:
So this works fine except for the fact that the vs2017 CI runners error out with this:

```
Program pypy3 found: YES (C:\hostedtoolcache\windows\PyPy\3.6.12\x86\pypy3.EXE)
Need python for 64-bit, but found 32-bit

test cases\python\8 different python versions\meson.build:17:2: ERROR: Python dependency not found
```

Meanwhile we specifically install:
```yaml

  - task: UsePythonVersion@0
    # Put this first since addToPath prepends, and
    # we want CPython3.7 to "win"
    inputs:
      versionSpec: 'pypy3'
      addToPath: true
      architecture: 'x64'
```

Install an x64 pypy, get the x86 version on the path. What the heck. Mind-boggling.

eli-schwartz:
The "good" news is that vs2017 is deprecated and will be removed on April 1, 2022. Yes, that's last week. No, it's still there so the precise meaning of "will be removed" is either an April Fools joke that someone forgot to take out of the pipeline logging, or else confusingly there and who knows when it will just disappear suddenly.

Unfortunately updating it is blocked on #9752

eli-schwartz:
Maybe installing pypy manually is the best option.

mattip:
It looks like pypy3 is installed on windows and tests are reporting [it passes](https://dev.azure.com/jussi0947/jussi/_build/results?buildId=13822&view=logs&j=447df341-d5a1-5d0e-9651-c9f66402368a&t=780da750-c959-5d10-38a1-0ae8a9e9cdaf&l=2142)
```
 [SUCCESS]  (without install) python: 8 different python versions    (python=pypy3)
```

eli-schwartz:
AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA

```
Program pypy found: YES (D:\a\pypy38\pypy3.8-v7.3.9-win64\pypy.EXE)
Need python for 32-bit, but found 64-bit

test cases\python\8 different python versions\meson.build:17:2: ERROR: Python dependency not found
```

Now it wants the 32-bit version.

...

Okay. I think I'm beginning to understand. The vs2017 job looks like this:

```yaml
  strategy:
    matrix:
        vc2017x86ninja:
          arch: x86
          compiler: msvc2017
          backend: ninja
          MESON_RSP_THRESHOLD: 0
        vc2017x64vs:
          arch: x64
          compiler: msvc2017
          backend: vs2017
        clangclx64ninja:
          arch: x64
          compiler: clang-cl
          backend: ninja
```

Only two of those jobs are 64-bit, the third is 32-bit.

Notice how the ci script has a block of code doing: `($env:arch -eq 'x64')` checks. Apparently to hack around rust.

It's not clear to me, though, if pypy3 has prebuilt binaries for 32-bit at all, these days? Looks like maybe Azure's hostedtoolscache is always 32-bit for vs2017, and always 64-bit on newer versions of the Windows image. Seemingly corresponding to whichever version pypy3 offers for download.

So it seems like the correct conditional logic is either in the powershell script OR as an azure task, to request pypy3 only when the available version corresponds to the right architecture.
- This is probably a lot easier in the powershell script, since you can guarantee the latest version of pypy3, guarantee that it is 64-bit, and check that you only download and install it on 64-bit tests.
- The alternative seems to be to look up whether any given image has 32-bit or 64-bit pypy3, hope that this remains the same, and hardcode, according to [the docs](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/conditions?view=azure-devops&tabs=yaml), `condition: and(succeeded(), eq(variables.arch, '<known arch value>'))`

mattip:
> This is probably a lot easier in the powershell script

This seems reasonable. Azure has only 32-bit PyPy3.6 for windows, which is not supported (neither 3.6 nor 32-bit is supported)

mattip:
Windows builds are passing, could you approve the others?

eli-schwartz:
Done.

eli-schwartz:
So I wanted to test how this plays out, and therefore I made a branch here with just the testsuite additions: https://github.com/mesonbuild/meson/commits/0.pypy-test

Weirdly, the tests actually pass even without the dependency changes.

eli-schwartz:
I forced debug logs by adding a `raise SystemExit('dump logs')` meson.build test, and checked what it is doing on Windows.

https://dev.azure.com/jussi0947/jussi/_build/results?buildId=13845&view=logs&j=678c5357-36f0-50cc-b96e-1a6f7149eed2&t=fb842c1e-cd65-5c1e-ab4f-d7ba55deb214&l=2692

This apparently works:
```
[2/2] "lld-link"  /MACHINE:x64 /OUT:ext/tachyon.pypy38-pp73-win_amd64.pyd ext/tachyon.pypy38-pp73-win_amd64.pyd.p/tachyon_module.c.obj "/nologo" "/OPT:REF" "/DLL" "/IMPLIB:ext\tachyon.pypy38-pp73-win_amd64.lib" "D:\a\pypy38\pypy3.8-v7.3.9-win64\libs\python38.lib" "kernel32.lib" "user32.lib" "gdi32.lib" "winspool.lib" "shell32.lib" "ole32.lib" "oleaut32.lib" "uuid.lib" "comdlg32.lib" "advapi32.lib"
```

```ninja
build ext/tachyon.pypy38-pp73-win_amd64.pyd.p/tachyon_module.c.obj: c_COMPILER ../test$ cases/python/8$ different$ python$ versions/ext/tachyon_module.c
 DEPFILE = "ext\tachyon.pypy38-pp73-win_amd64.pyd.p\tachyon_module.c.obj.d"
 DEPFILE_UNQUOTED = ext\tachyon.pypy38-pp73-win_amd64.pyd.p\tachyon_module.c.obj.d
 ARGS = "-Iext\tachyon.pypy38-pp73-win_amd64.pyd.p" "-Iext" "-I..\test$ cases\python\8$ different$ python$ versions\ext" "-ID:\a\pypy38\pypy3.8-v7.3.9-win64\Include" "-DNDEBUG" "/MD" "/nologo" "/showIncludes" "/utf-8" "/W2" "/O2" "/Fdext\tachyon.pypy38-pp73-win_amd64.pyd.p\tachyon_module.c.pdb"

build ext/tachyon.pypy38-pp73-win_amd64.pyd.p/tachyon.pypy38-pp73-win_amd64.pyd.symbols: SHSYM ext/tachyon.pypy38-pp73-win_amd64.pyd
 IMPLIB = "ext\tachyon.pypy38-pp73-win_amd64.lib"

build ext/tachyon.pypy38-pp73-win_amd64.pyd: c_LINKER ext/tachyon.pypy38-pp73-win_amd64.pyd.p/tachyon_module.c.obj | D$:/a/pypy38/pypy3.8-v7.3.9-win64/libs/python38.lib
 LINK_ARGS = "/nologo" "/OPT:REF" "/DLL" "/IMPLIB:ext\tachyon.pypy38-pp73-win_amd64.lib" "D:\a\pypy38\pypy3.8-v7.3.9-win64\libs\python38.lib" "kernel32.lib" "user32.lib" "gdi32.lib" "winspool.lib" "shell32.lib" "ole32.lib" "oleaut32.lib" "uuid.lib" "comdlg32.lib" "advapi32.lib"
```

eli-schwartz:
Oh wait -- I think I understand the difference. We are testing MSVC there, and this PR only changes the link naming for gcc on Windows semantics. I guess maybe we need to test in MSYS2? Now that I come to think of it, I believe SciPy was using MinGW anyway because of some kind of twisty thing where if you do compile with MSVC, you need a fortran compiler and the only decently available OSS one is gfortran, and combining MSVC + gfortran is "complicated". So presumably the issues building SciPy with pypy were also using MinGW.

So perhaps we should add pypy to .github/workflows/msys2.yml as well, to make sure that that does not regress either.

.github/workflows/msys2.yml
```diff
@@ -81,6 +82,18 @@ jobs:
...
+        run: |
+          mkdir pypy3local
+          pushd pypy3local
+          wget https://downloads.python.org/pypy/pypy3.8-v7.3.9-win64.zip
+          unzip pypy3.8-v7.3.9-win64.zip
+          mv pypy3.8-v7.3.9-win64/* ./
+          popd 
+          echo $PWD\pypy3local | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+          pypy3 -m ensurepip
+        if ${{ env.MSYS2_ARCH == 'x86_64' }}
```
eli-schwartz, line 95: https://github.com/mesonbuild/meson/actions/runs/2151064770/workflow#L95

```suggestion
        if: ${{ env.MSYS2_ARCH == 'x86_64' }}
```

mattip: thanks

mattip:
If adding msys2 looks complicated, maybe we should merge without it and add it in a second PR. That way I will already be a second-time contributor and won't need you to approve running workflows.

eli-schwartz:
I went ahead and cherry-picked the Ubuntu bionic addition to git master. This has two advantages in one:
- you are now a contributor to this repository, and github no longer refuses to run your CI until someone clicks "approve" on each push
- within the next hour, the bionic image will be rebuilt and pushed, so the Arch/OpenSUSE image builds no longer falsely report existing breakage

eli-schwartz:
The msys2 x86_64 jobs are skipping PyPy installation. I can never remember off the top of my head which type of variables you can use, at what times, and on which OCI services... Is it possible that this is silently null at the point you're trying to use it?

eli-schwartz:
... or it could be a typo and github's workflow validator simply doesn't care if you reference things that don't exist, lol.

BTW: you could rebase on master and drop the commit I cherry-picked, and the PR will stop trying to run the CI image builder workflow (minus 6 jobs taking up slots). And the bionic image is uploaded now.

eli-schwartz:
For log readability purposes -- the wget call should be wget -nv to make the progress bar not be mis-detected in CI and print a billion lines instead of constantly updating a single line.

mattip:
The msys2/clangx64ninja run [found and tested PyPy3](https://github.com/mesonbuild/meson/runs/5980965739?check_suite_focus=true#step:6:1918), as did the [msys2/gcc64ninja run](https://github.com/mesonbuild/meson/runs/5980965667?check_suite_focus=true#step:6:1915)

eli-schwartz:
Can I trouble you to fold the last two commits into a single commit before this gets merged?

mattip:
Done

mattip:
Thanks!

eli-schwartz:
Thank *you* for being so patient and championing this bug fix and bravely facing the Dread Beast that is CI! :p

PR #10140, opened 2022-03-17
gnome module: fix incorrect lookup of nonexistent dependencies in post_install

### LINKED ISSUE #10139: Installed components (gtk4, desktop-file-utils) not detected by Meson
**Describe the bug**
Installed components (gtk4, desktop-file-utils) not detected by Meson
```
$ dnf -q --installed list gtk4 desktop-file-utils | sed 1d
desktop-file-utils.x86_64                0.26-5.fc35                 @System
gtk4.x86_64                              4.4.2-1.fc35                @System
```

**To Reproduce**
```
$ meson --reconfigure builddir
The Meson build system
Version: 0.61.3
Source dir: /home/user/Downloads/Flatseal
Build dir: /home/user/Downloads/Flatseal/builddir
Build type: native build
Project name: flatseal
Project version: 1.7.5
Host machine cpu: x86_64
Program desktop-file-validate found: YES (/usr/bin/desktop-file-validate)
Program appstream-util found: YES (/usr/bin/appstream-util)
Program glib-compile-schemas found: YES (/usr/bin/glib-compile-schemas)
Found pkg-config: /usr/bin/pkg-config (1.8.0)
Program glib-compile-resources found: YES (/usr/bin/glib-compile-resources)
Program glib-compile-resources found: YES (/usr/bin/glib-compile-resources)
Program gjs found: YES (/usr/bin/gjs)
Configuring com.github.tchx84.Flatseal using configuration
Program jasmine found: NO
Dependency gio-2.0 found: YES 2.70.4 (cached)
Program glib-compile-schemas found: YES (/usr/bin/glib-compile-schemas)
Found CMake: /usr/bin/cmake (3.22.2)
Build-time dependency gtk4 found: NO (tried pkgconfig and cmake)
Program gtk4-update-icon-cache found: YES (/usr/bin/gtk4-update-icon-cache)
Build-time dependency desktop-file-utils found: NO (tried pkgconfig and cmake)
Program update-desktop-database found: YES (/usr/bin/update-desktop-database)
Build targets in project: 22

```

**Expected behavior**
Installed components to be detected by Meson

**system parameters**
cross build: ?; plain native build: ?
(in plain English) build via 
```
$ git clone https://github.com/mesonbuild/meson.git`
$ meson builddir
```
**OS:** Fedora
```
$ pip3 -V
pip 21.2.3 from /usr/lib/python3.10/site-packages/pip (python 3.10)
```

eli-schwartz:
While gtk+-3.0 / gtk4 do exist, they have never provided the location of the gtk-update-icon-cache program as a pkgconfig variable. Trying to find one anyway, resulted in two things happening:
- a useless dep lookup
- a fatal-meson-warnings error and build failure because the get_pkgconfig_variable() in question never existed

The desktop-file-utils package is a package solely providing some command line programs, and has never provided a pkg-config file in the first place, so this always logged that the dependency was not found and fell back to normal find_program_impl(), although without fatal-meson-warnings build errors.

Fixes #10139

### DIFF
--- a/mesonbuild/modules/gnome.py
+++ b/mesonbuild/modules/gnome.py
@@ -368,17 +368,17 @@ class GnomeModule(ExtensionModule):
                 rv.append(script)
         if kwargs['gtk_update_icon_cache'] and not self.install_gtk_update_icon_cache:
             self.install_gtk_update_icon_cache = True
-            prog = self._get_native_binary(state, 'gtk4-update-icon-cache', 'gtk4', 'gtk4_update_icon_cache', required=False)
+            prog = state.find_program('gtk4-update-icon-cache', required=False)
             found = isinstance(prog, build.Executable) or prog.found()
             if not found:
-                prog = self._get_native_binary(state, 'gtk-update-icon-cache', 'gtk+-3.0', 'gtk_update_icon_cache')
+                prog = state.find_program('gtk4-update-icon-cache')
             icondir = os.path.join(datadir_abs, 'icons', 'hicolor')
             script = state.backend.get_executable_serialisation([prog, '-q', '-t', '-f', icondir])
             script.skip_if_destdir = True
             rv.append(script)
         if kwargs['update_desktop_database'] and not self.install_update_desktop_database:
             self.install_update_desktop_database = True
-            prog = self._get_native_binary(state, 'update-desktop-database', 'desktop-file-utils', 'update_desktop_database')
+            prog = state.find_program('update-desktop-database')
             appdir = os.path.join(datadir_abs, 'applications')
             script = state.backend.get_executable_serialisation([prog, '-q', appdir])
             script.skip_if_destdir = True


jpakkane:
```
====================== python: 8 different python versions    (python=python2) =====================
 

Build started at 2022-03-18T01:32:57.846263
    ret = f(*wrapped_args, **wrapped_kwargs)
  File "D:\a\1\s\mesonbuild\interpreterbase\decorators.py", line 127, in wrapped
  File "D:\a\1\s\mesonbuild\interpreterbase\decorators.py", line 768, in wrapped
  File "D:\a\1\s\mesonbuild\interpreterbase\decorators.py", line 64, in wrapped
  File "D:\a\1\s\mesonbuild\modules\python.py", line 547, in dependency_method
    raise mesonlib.MesonException('Python dependency not found')
mesonbuild.mesonlib.universal.MesonException: Python dependency not found
 
```

Seems unrelated but worrisome. Why is it trying to look up `python2` on Windows where it clearly could not exist.

eli-schwartz:
Because it does exist! I tried to fix it in #10144 but perhaps there's a better way.

xclaesse:
> a fatal-meson-warnings error and build failure because the get_pkgconfig_variable() in question never existed

Fun fact, that would have been fixed as side effect of https://github.com/mesonbuild/meson/pull/10041.

eli-schwartz:
Then please fix the bug in #10041 that causes it to not emit warnings for incorrect usage.

xclaesse:
LGTM, I see no rational for using pkg-config in those cases indeed. I'm surprised because IIRC it is somethingBSD that likes to put those tools out of PATH and rely on pkg-config to find them which actually makes sense IMHO. I'm a little bit worried they could be patching pc files downstream for that, but that's their problem if they do. If they care they would patch gtk4 upstream and then we can add this back.

xclaesse:
> Then please fix the bug in #10041 that causes it to not emit warnings for incorrect usage.

That's not a bug, I made it on purpose because it's the kind of variable that gets added over time (I fully expect gtk4 to add it some day), so I wanted this feature to use the variable only if available.

eli-schwartz:
> I'm surprised because IIRC it is somethingBSD that likes to put those tools out of PATH and rely on pkg-config to find them which actually makes sense IMHO.

Actually it is all the BSDs that refuse to include pkg-config files in the base system for the dependencies that they natively provide.

They install all programs, libraries, and anything else that isn't part of the base system, as part of "ports" which means /usr/local, and I believe this is in fact on the default $PATH. They do have other quirks though, related to the compiler path.

eli-schwartz:
This is a bugfix and a candidate for backporting in a point release, so I've merged it during the rc freeze.

PR #10143, opened 2022-03-18
Rename one of the duplicate 18 test cases

tristan957:


eli-schwartz:
As mentioned in IRC, `./tools/dircondenser.py` will do lots of this right before the final release (plus fixups for the directory string in various files), so we agreed to close this PR.

PR #10144, opened 2022-03-18
tests: make python2 dependency on 32-bit windows non-fatal

eli-schwartz:
Apparently Azure provides 64-bit python2 when we try to test 32-bit, and that breaks everything on the 32-bit test runner.

I don't understand the environment setup, and that runner is disappearing soon anyway. Hopefully this shuts up the known breakage.


eli-schwartz:
Comment from another PR:

> Why is it trying to look up python2 on Windows where it clearly could not exist.

@jpakkane perhaps there is a way to *make* it not exist? There is definitely a python2 installed on that windows runner though.

Issue #9849, opened 2022-01-19
depfiles not working for incbin asm statements


kusma:
**Describe the bug**

The dependency-tracking method Meson's Ninja backend uses doesn't keep track of the `.incbin <source>` assembly directive. The same issue probably also exists for the `.include <source>` directive, but that can be worked around using the C pre-processor on top.

Arguably the problem lies in GCC, but Meson doesn't seem to provide any escape-hatch that can be used to add back the missing dependencies.

**To Reproduce**

Create the following project:

`foo.S`:
```asm
.section .data

.global foo_bin
foo_bin:
.incbin "foo.bin"
```
`foo.bin`:
```
abcdef
```
`main.c`:
```c
#include <stdio.h>
extern char foo_bin[];
int main()
{
       printf("foo_bin: \"%.*s\"\n", 4, foo_bin);
       return 0;
}
```
`meson.build`:
```
project('meson-incbin-test', 'c')
executable('incbin-test', 'main.c', 'foo.S')
```

Now, run
1. `meson setup build`
2. `ninja -C build/`
3. `echo "foobar" > foo.bin`
4. `ninja -C build/`
5. `build/incbin-test`
6.  Observe how `foo_bin: "abcd"` is printed.

**Expected behavior**
The output `foo_bin: "foob"` should be printed on screen 

**system parameters**

Happens both for cross and native builds.
Ubuntu 21.10
Python 3.9.7
Meson 0.61.0
Ninja 1.10.1

eli-schwartz:
Add `foo.bin` as a source file, meson considers it an unknown filetype and treats it the same way it treats `*.h` files -- by making it an order-only dependency.

kusma:
@eli-schwartz:
That gives the following error:
```
ERROR: No specified compiler can handle file foo.bin
```

Can you please reopen?

eli-schwartz:
Actually I'm not sure whether that will recompile the asm file itself... Hmm.

eli-schwartz:
Huh, i know this works for stuff like .inc so I'm not sure why it fails for .bin

kusma:
Dunno, doesn't seem to work for .inc for me either:

> ERROR: No specified compiler can handle file foo.inc

kusma:
Seems [CMake has had similar issues](https://gitlab.kitware.com/cmake/cmake/-/issues/20426), which I guess explains things a bit:

Here's what I generate if I run the compile-step from the example:

```
[...]

Issue #9889, opened 2022-01-27
Built-in support for pre-processing + compiling assembly with MSVC


nirbheek:
With GCC, Clang, ICC and other GNU-compatible C compilers, you can pass a `.S` file as a source file alongside C/C++ source files to pre-process and compile.

However, with MSVC these files need to be pre-processed with `cl /P` and then compiled with one of `ml` `ml64` `armasm` `armasm64` depending on the target architecture, [see example where `custom_target()` is used to do this manually](https://gitlab.freedesktop.org/gstreamer/meson-ports/libffi/-/blob/5330111d527dd5671a31e586d579c29f12c345e4/src/meson.build#L23-69).

Projects have been implementing their own versions of the `custom_target()` workaround shown above. We should just support it all out of the box, f.ex. libpng: https://github.com/mesonbuild/wrapdb/pull/205

Somewhat related to https://github.com/mesonbuild/meson/pull/9379 -- will share base classes perhaps.

dcbaker:
We also found out recently that gcc/clang don't handle depfiles for .S correctly, so we actually need to do the same tihng with gcc and clang.

xclaesse:
@nirbheek @bonzini I was wondering what are the typical file extension used for this use-case. I think .S -> cl -> MASM is pretty unambiguous, but does Windows projects typically use `.asm` extension as well? Do they also expect them to be preprocessed? Are there cases where people expects files to be compiled by MASM without preprocess?

bonzini:
.asm is common for MASM projects (which is why I suggested .nasm for NASM to avoid ambiguity) but they are not preprocessed and that's painful.

Based on other reading such as https://blog.mozilla.org/nfroyd/2019/04/25/an-unexpected-benefit-of-standardizing-on-clang-cl/, I am wondering if Meson should completely ignore MASM, and focus on the gas (including clang-cl) and nasm (including yasm) syntaxes. In that case, my objection on using `.asm` for nasm is rescinded.

nirbheek:
> Ok, so we can easily add `add_language('masm')` just like I did for NASM. Adding both masm and nasm language in the same project would be an error because they compete for the same file extension. That seems fair enough to me.

Out of interest, does that work correctly for subprojects right now? i.e., can one use masm and nasm in separate subprojects?

xclaesse:
> IIRC compiler list is global instead of per subproject

[...]

Issue #8961, opened 2021-07-04
Cython does not respect dependency for generating `.pyx.c` file


rgommers:
**Describe the bug**

Building a Python extension from a generated Cython file (`_ufuncs.pyx`) fails with:
```
[1/539] Compiling Cython source scipy/special/_ufuncs.pyx
FAILED: scipy/special/_ufuncs.cpython-39-x86_64-linux-gnu.so.p/scipy/special/_ufuncs.pyx.c 
cython --fast-fail -3 -o scipy/special/_ufuncs.cpython-39-x86_64-linux-gnu.so.p/scipy/special/_ufuncs.pyx.c scipy/special/_ufuncs.pyx

Error compiling Cython file:
------------------------------------------------------------
...
# This file is automatically generated by _generate_pyx.py.
# Do not edit manually!

include "_ufuncs_extra_code_common.pxi"
^
------------------------------------------------------------
```

The relevant `meson.build` content is:
```python
# Generates the `_ufuncs.pyx` file that fails to build later
_generate_cy = custom_target('_generate_cy',
  output : ['_ufuncs.pyx',
            '_ufuncs_defs.h',
            '_ufuncs_cxx.pyx',
            '_ufuncs_cxx.pxd',
            '_ufuncs_cxx_defs.h',
            '_ufuncs.pyi',
            'cython_special.pyx',
            'cython_special.pxd'
           ],
  input : ['_generate_pyx.py', 'functions.json', 'add_newdocs.py'],
  command : [py3, '@INPUT0@', '-o', '@OUTDIR@'],
)

# Simply copy over the files we need to the build dir
_ufuncs_pxi_pxd_sources = custom_target('_ufuncs_pxi_pxd',
  output : [
    '__init__.py',
    ...
    '_ufuncs_extra_code.pxi',
    '_ufuncs_extra_code_common.pxi',
  ],
  input : [
    '__init__.py',
    ...
    '_ufuncs_extra_code.pxi',
    '_ufuncs_extra_code_common.pxi',
  ],
  command : ['cp', '@INPUT@', '-t', '@OUTDIR@'],
)

# FIXME: the generated ninja.build file is missing the dependency between
#       copying these files into the build dir and the `.pyx -> .c` stage.
#       For the `.c -> .so` stage the dependency (see `||` in the `build.ninja`
#       file) is picked up correctly. This only shows up when running the build
#       with `-j 1` (when the steps are run in parallel, the copying of sources
[...]

PR #10149, opened 2022-03-19
ninjabackend: Split preprocessing .S files and compiling .s files

xclaesse:
commit d8737c5930e6e93c292903514a5028abbd0b57f4

    ninjabackend: Split preprocessing .S files and compiling .s files
    
    GCC internally compiles .S files in 2 steps. It first preprocess them
    like any C file which generate a depfile for macros like `#include`. It
    then hand the preprocessed file to GAS that generates its own depfile
    for assembly instructions like `.incbin`. Since Ninja does not support
    targets that generate multiple depfiles we have to split into 2 separate
    Ninja targets.
    
    This has the bonus side effect of opening the door to preprocess
    assembly files using any C compiler, then compile them with a different
    tool such as MASM on Windows.
    
    This is implemented with a new CompilerMode class that is basically a
    small override for a given Compiler object that represents a different
    Ninja rule with different fixed arguments. All other compiler arguments
    (e.g. c_args='-I..') are identical for all modes, so they are not a new
    language at Meson level.

commit 40b618e736cd82136fbf6f9363d92e5f1b049d15

    ninjabackend: Refactor how vala/cython transpilers works
    
    This makes the code path for transpilers more generic to make it easy to
    add new transpilers in the future (e.g. .S -> .s preprocessor). Any
    Compiler class can implement get_output_suffix() method to return the
    file suffix it can transpile to (e.g. cython -> '.c') and their output
    will be handed to another compiler, until we get a final object file.

commit d3d420707f891cddf76fc071dcbcab3ff8040f25

    ninjabackend: Fix get_target_generated_sources() return type
    
    Type annotation, documentation string, and implementation were doing 3
    different things. Change implementation to match type annotation which
    makes the most sense because it match what get_target_sources() does.
    
    All callers only use keys from the returned dictionary any way, but
    that's going to change in next commits.

commit 72a19964046ddb0db1f04713c1edb719b6fe5f0e

    compilers: Make sure to not use ccache in compiler checks
    
    ccache was used in all command lines but disabled using CCACHE_DISABLE
    in Compiler.compile() method. Wrapping invokations still has a cost,
    especially on Windows.
    
    With sccache things are even worse because CCACHE_DISABLE was not
    respected at all, making configure *extremely* slow on Windows when
    sccache is installed.

commit ec2cfa4954d2fcd7fabe9ab97577b572fd65d3b8

    compilers: Cleanup a bit languages/suffixes lists
    
    Use set where order does not matter, fix is_source() to really mean only
    source suffixes.

### DIFF
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -820,7 +820,7 @@ class Backend:
         # Filter out headers and all non-source files
         sources: T.List['FileOrString'] = []
         for s in raw_sources:
-            if self.environment.is_source(s) and not self.environment.is_header(s):
+            if self.environment.is_source(s):
                 sources.append(s)
             elif self.environment.is_object(s):
                 result.append(s.relative_name())
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -774,7 +774,7 @@ class Backend:
             fname = fname.replace(ch, '_')
         return fname
 
-    def object_filename_from_source(self, target: build.BuildTarget, source: 'FileOrString') -> str:
+    def object_filename_from_source(self, target: build.BuildTarget, source: 'FileOrString', obj_suffix: T.Optional[str] = None) -> str:
         assert isinstance(source, mesonlib.File)
         build_dir = self.environment.get_build_dir()
         rel_src = source.rel_to_builddir(self.build_to_src)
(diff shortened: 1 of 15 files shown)


xclaesse:
This fixes https://github.com/mesonbuild/meson/issues/9849, https://github.com/mesonbuild/meson/issues/8961, and is a good start at https://github.com/mesonbuild/meson/issues/9889.

xclaesse:
Known issues so far:
- clang fails to pass -Wa,-MD down to GAS, works with gcc. We might have to invoke as itself instead of relying on C front-end afterall.
- extract_all_objects, used by both_libraries, is broken because it assume .S sources have .S.o output instead of .S.s.o. I think this issue is not new and happens for cython too.

xclaesse:
@dcbaker @eli-schwartz Ok, this is ready for review now.

xclaesse:
Anyone for a review please?

xclaesse:
Let's close this PR for now, I started working on another approach.

PR #10151, opened 2022-03-19
Cargo subprojects, take two

dcbaker:
This is still *very* WIP, but I wanted to be able to track the status of it publically, since I'm trying to get this ready to solve real issues in Mesa, ASAP.

This is the second attempt at cargo-as-a-subproject. It differs from the previous attempt in a couple of major ways:
 - uses `tomli` instead of `toml`. `tomli` is on track to be in the standard library of python 3.11, so it will only be an external dependency for older versions of python
 - uses a different AST builder approach
 - uses a bewtter cfg parser
 - is structured better, due to me understanding the problem better

TODO:

- [x] trivial example
- [x] nested crates
- [ ] virtual workspaces
- [ ] non-virtual workspaces
- [ ] option parsing
- [ ] proc-macros
- [ ] tests
- [ ] benchmarks
- [ ] examples?
- [ ] combining features from multiple subprojects
- [x] replace pytest tests with unittests
- [ ] support [system-deps](https://crates.io/crates/system-deps) thanks for pointing this out @xclaesse 
- [ ] toml2json as an alternative to tomllib/tomli
- [ ] Support cargo-c
- [ ] Figure out something better than cargo vendor, which now requires a non-trival Cargo.toml
- [ ] Write a cargo plugin to allow moving rustc version checks out of build.rs and into cargo.toml
- [ ] add suggested metadata (and support) to tell meson it's okay to ignore the build.rs

### DIFF
--- a/mesonbuild/cargo/builder.py
+++ b/mesonbuild/cargo/builder.py
@@ -1,0 +1,284 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright © 2022-2023 Intel Corporation
+
+"""Provides helpers for building AST
+
+This is meant to make building Meson AST from foreign (largely declarative)
+build descriptions easier.
+"""
+
+from __future__ import annotations
+import builtins
+import dataclasses
+import typing as T
+
+from .. import mparser
+
+
+def _token(tid: str, filename: str, value: mparser.TV_TokenTypes) -> mparser.Token[mparser.TV_TokenTypes]:
+    """Create a Token object, but with the line numbers stubbed out.
+
+    :param tid: the token id (such as string, number, etc)
+    :param filename: the filename that the token was generated from
+    :param value: the value of the token
+    :return: A Token object
+    """
+    return mparser.Token(tid, filename, -1, -1, -1, (-1, -1), value)
+
+
+def string(value: str, filename: str) -> mparser.StringNode:
+    """Build A StringNode
+
+    :param value: the value of the string
+    :param filename: the file that the value came from
+    :return: A StringNode
+    """
+    return mparser.StringNode(_token('string', filename, value))
+
+
+def number(value: int, filename: str) -> mparser.NumberNode:
+    """Build A NumberNode
+
+    :param value: the value of the number
+    :param filename: the file that the value came from
+    :return: A NumberNode
+    """
+    return mparser.NumberNode(_token('number', filename, value))
+
+
+def bool(value: builtins.bool, filename: str) -> mparser.BooleanNode:
+    """Build A BooleanNode
+
+    :param value: the value of the boolean
+    :param filename: the file that the value came from
+    :return: A BooleanNode
+    """
+    return mparser.BooleanNode(_token('bool', filename, value))
+
+
+def array(value: T.List[mparser.BaseNode], filename: str) -> mparser.ArrayNode:
+    """Build an Array Node
+
+    :param value: A list of nodes to insert into the array
+    :param filename: The file the array is from
+    :return: An ArrayNode built from the arguments
+    """
+    args = mparser.ArgumentNode(_token('array', filename, 'unused'))
+    args.arguments = value
+    return mparser.ArrayNode(args, -1, -1, -1, -1)
+
+
+def identifier(value: str, filename: str) -> mparser.IdNode:
+    """Build A IdNode
+
+    :param value: the value of the boolean
+    :param filename: the file that the value came from
+    :return: A BooleanNode
+    """
+    return mparser.IdNode(_token('id', filename, value))
+
+
+def method(name: str, id_: mparser.IdNode,
+           pos: T.Optional[T.List[mparser.BaseNode]] = None,
+           kw: T.Optional[T.Mapping[str, mparser.BaseNode]] = None,
+           ) -> mparser.MethodNode:
+    """Create a method call.
+
+    :param name: the name of the method
+    :param id_: the object to call the method of
+    :param pos: a list of positional arguments, defaults to None
+    :param kw: a dictionary of keyword arguments, defaults to None
+    :return: a method call object
+    """
+    args = mparser.ArgumentNode(_token('array', id_.filename, 'unused'))
+    if pos is not None:
+        args.arguments = pos
+    if kw is not None:
+        args.kwargs = {identifier(k, id_.filename): v for k, v in kw.items()}
+    return mparser.MethodNode(id_.filename, -1, -1, id_, name, args)
+
+
+def function(name: str, filename: str,
+             pos: T.Optional[T.List[mparser.BaseNode]] = None,
+             kw: T.Optional[T.Mapping[str, mparser.BaseNode]] = None,
+             ) -> mparser.FunctionNode:
+    """Create a function call.
+
+    :param name: the name of the function
+    :param filename: The name of the current file being evaluated
+    :param pos: a list of positional arguments, defaults to None
+    :param kw: a dictionary of keyword arguments, defaults to None
+    :return: a method call object
+    """
+    args = mparser.ArgumentNode(_token('array', filename, 'unused'))
+    if pos is not None:
+        args.arguments = pos
+    if kw is not None:
+        args.kwargs = {identifier(k, filename): v for k, v in kw.items()}
+    return mparser.FunctionNode(filename, -1, -1, -1, -1, name, args)
+
+
+def equal(lhs: mparser.BaseNode, rhs: mparser.BaseNode) -> mparser.ComparisonNode:
+    """Create an equality operation
+
+    :param lhs: The left hand side of the equal
+    :param rhs: the right hand side of the equal
+    :return: A compraison node
+    """
+    return mparser.ComparisonNode('==', lhs, rhs)
+
+
+def or_(lhs: mparser.BaseNode, rhs: mparser.BaseNode) -> mparser.OrNode:
+    """Create and OrNode
+
+    :param lhs: The Left of the Node
+    :param rhs: The Right of the Node
+    :return: The OrNode
+    """
+    return mparser.OrNode(lhs, rhs)
+
+
+def and_(lhs: mparser.BaseNode, rhs: mparser.BaseNode) -> mparser.AndNode:
+    """Create an AndNode
+
+    :param lhs: The left of the And
+    :param rhs: The right of the And
+    :return: The AndNode
+    """
+    return mparser.AndNode(lhs, rhs)
+
+
+def not_(value: mparser.BaseNode, filename: str) -> mparser.NotNode:
+    """Create a not node
+
+    :param value: The value to negate
+    :param filename: the string filename
+    :return: The NotNode
+    """
+    return mparser.NotNode(_token('not', filename, ''), value)
+
+
+def assign(value: mparser.BaseNode, varname: str, filename: str) -> mparser.AssignmentNode:
+    """Create an AssignmentNode
+
+    :param value: The rvalue
+    :param varname: The lvalue
+    :param filename: The filename
+    :return: An AssignmentNode
+    """
+    return mparser.AssignmentNode(filename, -1, -1, varname, value)
+
+
+def block(filename: str) -> mparser.CodeBlockNode:
+    return mparser.CodeBlockNode(_token('node', filename, ''))
+
+
+@dataclasses.dataclass
+class Builder:
+
+    filename: str
+
+    def assign(self, value: mparser.BaseNode, varname: str) -> mparser.AssignmentNode:
+        return assign(value, varname, self.filename)
+
+    def string(self, value: str) -> mparser.StringNode:
+        """Build A StringNode
+
+        :param value: the value of the string
+        :return: A StringNode
+        """
+        return string(value, self.filename)
+
+    def number(self, value: int) -> mparser.NumberNode:
+        """Build A NumberNode
+
+        :param value: the value of the number
+        :return: A NumberNode
+        """
+        return number(value, self.filename)
+
+    def bool(self, value: builtins.bool) -> mparser.BooleanNode:
+        """Build A BooleanNode
+
+        :param value: the value of the boolean
+        :return: A BooleanNode
+        """
+        return bool(value, self.filename)
+
+    def array(self, value: T.List[mparser.BaseNode]) -> mparser.ArrayNode:
+        """Build an Array Node
+
+        :param value: A list of nodes to insert into the array
+        :return: An ArrayNode built from the arguments
+        """
+        return array(value, self.filename)
+
+    def identifier(self, value: str) -> mparser.IdNode:
+        """Build A IdNode
+
+        :param value: the value of the boolean
+        :return: A BooleanNode
+        """
+        return identifier(value, self.filename)
+
+    def method(self, name: str, id_: mparser.IdNode,
+               pos: T.Optional[T.List[mparser.BaseNode]] = None,
+               kw: T.Optional[T.Mapping[str, mparser.BaseNode]] = None,
+               ) -> mparser.MethodNode:
+        """Create a method call.
+
+        :param name: the name of the method
+        :param id_: the object to call the method of
+        :param pos: a list of positional arguments, defaults to None
+        :param kw: a dictionary of keyword arguments, defaults to None
+        :return: a method call object
+        """
+        return method(name, id_, pos or [], kw or {})
+
+    def function(self, name: str,
+                 pos: T.Optional[T.List[mparser.BaseNode]] = None,
+                 kw: T.Optional[T.Mapping[str, mparser.BaseNode]] = None,
+                 ) -> mparser.FunctionNode:
+        """Create a function call.
+
+        :param name: the name of the function
+        :param pos: a list of positional arguments, defaults to None
+        :param kw: a dictionary of keyword arguments, defaults to None
+        :return: a method call object
+        """
+        return function(name, self.filename, pos or [], kw or {})
+
+    def equal(self, lhs: mparser.BaseNode, rhs: mparser.BaseNode) -> mparser.ComparisonNode:
+        """Create an equality operation
+
+        :param lhs: The left hand side of the equal
+        :param rhs: the right hand side of the equal
+        :return: A compraison node
+        """
+        return equal(lhs, rhs)
+
+    def or_(self, lhs: mparser.BaseNode, rhs: mparser.BaseNode) -> mparser.OrNode:
+        """Create and OrNode
+
+        :param lhs: The Left of the Node
+        :param rhs: The Right of the Node
+        :return: The OrNode
+        """
+        return or_(lhs, rhs)
+
+    def and_(self, lhs: mparser.BaseNode, rhs: mparser.BaseNode) -> mparser.AndNode:
+        """Create an AndNode
+
+        :param lhs: The left of the And
+        :param rhs: The right of the And
+        :return: The AndNode
+        """
+        return and_(lhs, rhs)
+
+    def not_(self, value: mparser.BaseNode, filename: str) -> mparser.NotNode:
+        """Create a not node
+
+        :param value: The value to negate
+        :return: The NotNode
+        """
+        return not_(value, self.filename)
--- a/mesonbuild/cargo/interpreter.py
+++ b/mesonbuild/cargo/interpreter.py
@@ -1,0 +1,451 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright © 2022-2023 Intel Corporation
+
+"""Interpreter for converting Cargo Toml definitions to Meson AST
+
+There are some notable limits here. We don't even try to convert something with
+a build.rs: there's so few limits on what Cargo allows a build.rs (basically
+none), and no good way for us to convert them. In that case, an actual meson
+port will be required.
+"""
+
+from __future__ import annotations
+import dataclasses
+import glob
+import importlib
+import itertools
+import json
+import os
+import shutil
+import typing as T
+
+from . import builder
+from . import version
+from .. import mparser
+from .._pathlib import Path
+from ..mesonlib import MesonException, Popen_safe
+
+if T.TYPE_CHECKING:
+    from types import ModuleType
+
+    from . import manifest
+    from ..environment import Environment
+
+# tomllib is present in python 3.11, before that it is a pypi module called tomli,
+# we try to import tomllib, then tomli,
+# TODO: add a fallback to toml2json?
+tomllib: T.Optional[ModuleType] = None
+toml2json: T.Optional[str] = None
+for t in ['tomllib', 'tomli']:
+    try:
+        tomllib = importlib.import_module(t)
+        break
+    except ImportError:
+        pass
+else:
+    # TODO: it would be better to use an Executable here, which could be looked
+    # up in the cross file or provided by a wrap. However, that will have to be
+    # passed in externally, since we don't have (and I don't think we should),
+    # have access to the `Environment` for that in this module.
+    toml2json = shutil.which('toml2json')
+
+
+def load_toml(filename: str) -> T.Dict[object, object]:
+    if tomllib:
+        with open(filename, 'rb') as f:
+            raw = tomllib.load(f)
+    else:
+        if toml2json is None:
+            raise MesonException('Could not find an implementation of tomllib, nor toml2json')
+
+        p, out, err = Popen_safe([toml2json, filename])
+        if p.returncode != 0:
+            raise MesonException('toml2json failed to decode output\n', err)
+
+        raw = json.loads(out)
+
+    if not isinstance(raw, dict):
+        raise MesonException("Cargo.toml isn't a dictionary? How did that happen?")
+
+    return raw
+
+
+def fixup_meson_varname(name: str) -> str:
+    """Fixup a meson variable name
+
+    :param name: The name to fix
+    :return: the fixed name
+    """
+    return name.replace('-', '_')
+
+# Pylance can figure out that these do not, in fact, overlap, but mypy can't
+@T.overload
+def _fixup_raw_mappings(d: manifest.BuildTarget) -> manifest.FixedBuildTarget: ...  # type: ignore
+
+@T.overload
+def _fixup_raw_mappings(d: manifest.LibTarget) -> manifest.FixedLibTarget: ...  # type: ignore
+
+@T.overload
+def _fixup_raw_mappings(d: manifest.Dependency) -> manifest.FixedDependency: ...
+
+def _fixup_raw_mappings(d: T.Union[manifest.BuildTarget, manifest.LibTarget, manifest.Dependency]
+                        ) -> T.Union[manifest.FixedBuildTarget, manifest.FixedLibTarget,
+                                     manifest.FixedDependency]:
+    """Fixup raw cargo mappings to ones more suitable for python to consume.
+
+    This does the following:
+    * replaces any `-` with `_`, cargo likes the former, but python dicts make
+      keys with `-` in them awkward to work with
+    * Convert Dependndency versions from the cargo format to something meson
+      understands
+
+    :param d: The mapping to fix
+    :return: the fixed string
+    """
+    raw = {fixup_meson_varname(k): v for k, v in d.items()}
+    if 'version' in raw:
+        assert isinstance(raw['version'], str), 'for mypy'
+        raw['version'] = version.convert(raw['version'])
+    return T.cast('T.Union[manifest.FixedBuildTarget, manifest.FixedLibTarget, manifest.FixedDependency]', raw)
+
+
+@dataclasses.dataclass
+class Package:
+
+    """Representation of a Cargo Package entry, with defaults filled in."""
+
+    name: str
+    version: str
+    description: str
+    resolver: T.Optional[str] = None
+    authors: T.List[str] = dataclasses.field(default_factory=list)
+    edition: manifest.EDITION = '2015'
+    rust_version: T.Optional[str] = None
+    documentation: T.Optional[str] = None
+    readme: T.Optional[str] = None
+    homepage: T.Optional[str] = None
+    repository: T.Optional[str] = None
+    license: T.Optional[str] = None
+    license_file: T.Optional[str] = None
+    keywords: T.List[str] = dataclasses.field(default_factory=list)
+    categories: T.List[str] = dataclasses.field(default_factory=list)
+    workspace: T.Optional[str] = None
+    build: T.Optional[str] = None
+    links: T.Optional[str] = None
+    exclude: T.List[str] = dataclasses.field(default_factory=list)
+    include: T.List[str] = dataclasses.field(default_factory=list)
+    publish: bool = True
+    metadata: T.Dict[str, T.Dict[str, str]] = dataclasses.field(default_factory=dict)
+    default_run: T.Optional[str] = None
+    autobins: bool = True
+    autoexamples: bool = True
+    autotests: bool = True
+    autobenches: bool = True
+
+
+@dataclasses.dataclass
+class Dependency:
+
+    """Representation of a Cargo Dependency Entry."""
+
+    version: T.List[str]
+    registry: T.Optional[str] = None
+    git: T.Optional[str] = None
+    branch: T.Optional[str] = None
+    rev: T.Optional[str] = None
+    path: T.Optional[str] = None
+    optional: bool = False
+    package: T.Optional[str] = None
+    default_features: bool = False
+    features: T.List[str] = dataclasses.field(default_factory=list)
+
+    @classmethod
+    def from_raw(cls, raw: manifest.DependencyV) -> Dependency:
+        """Create a dependency from a raw cargo dictionary"""
+        if isinstance(raw, str):
+            return cls(version.convert(raw))
+        return cls(**_fixup_raw_mappings(raw))
+
+
+@dataclasses.dataclass
+class BuildTarget:
+
+    name: str
+    crate_type: T.List[manifest.CRATE_TYPE] = dataclasses.field(default_factory=lambda: ['lib'])
+    path: dataclasses.InitVar[T.Optional[str]] = None
+
+    # https://doc.rust-lang.org/cargo/reference/cargo-targets.html#the-test-field
+    # True for lib, bin, test
+    test: bool = True
+
+    # https://doc.rust-lang.org/cargo/reference/cargo-targets.html#the-doctest-field
+    # True for lib
+    doctest: bool = False
+
+    # https://doc.rust-lang.org/cargo/reference/cargo-targets.html#the-bench-field
+    # True for lib, bin, benchmark
+    bench: bool = True
+
+    # https://doc.rust-lang.org/cargo/reference/cargo-targets.html#the-doc-field
+    # True for libraries and binaries
+    doc: bool = False
+
+    harness: bool = True
+    edition: manifest.EDITION = '2015'
+    required_features: T.List[str] = dataclasses.field(default_factory=list)
+    plugin: bool = False
+
+
+@dataclasses.dataclass
+class Library(BuildTarget):
+
+    """Representation of a Cargo Library Entry."""
+
+    doctest: bool = True
+    doc: bool = True
+    proc_macro: bool = False
+    crate_type: T.List[manifest.CRATE_TYPE] = dataclasses.field(default_factory=lambda: ['lib'])
+    doc_scrape_examples: bool = True
+
+
+@dataclasses.dataclass
+class Binary(BuildTarget):
+
+    """Representation of a Cargo Bin Entry."""
+
+    doc: bool = True
+
+
+@dataclasses.dataclass
+class Test(BuildTarget):
+
+    """Representation of a Cargo Test Entry."""
+
+    bench: bool = True
+
+
+@dataclasses.dataclass
+class Benchmark(BuildTarget):
+
+    """Representation of a Cargo Benchmark Entry."""
+
+    test: bool = True
+
+
+@dataclasses.dataclass
+class Example(BuildTarget):
+
+    """Representation of a Cargo Example Entry."""
+
+    crate_type: T.List[manifest.CRATE_TYPE] = dataclasses.field(default_factory=lambda: ['bin'])
+
+
+@dataclasses.dataclass
+class Manifest:
+
+    """Cargo Manifest definition.
+
+    Most of these values map up to the Cargo Manifest, but with default values
+    if not provided.
+
+    Cargo subprojects can contain what Meson wants to treat as multiple,
+    interdependent, subprojects.
+
+    :param subdir: the subdirectory that this cargo project is in
+    :param path: the path within the cargo subproject.
+    """
+
+    package: Package
+    dependencies: T.Dict[str, Dependency]
+    dev_dependencies: T.Dict[str, Dependency]
+    build_dependencies: T.Dict[str, Dependency]
+    lib: Library
+    bin: T.List[Binary]
+    test: T.List[Test]
+    bench: T.List[Benchmark]
+    example: T.List[Example]
+    features: T.Dict[str, T.List[str]]
+    target: T.Dict[str, T.Dict[str, Dependency]]
+    subdir: str
+    path: str = ''
+
+
+def _create_project(package: Package, build: builder.Builder, env: Environment) -> mparser.FunctionNode:
+    """Create a function call
+
+    :param package: The Cargo package to generate from
+    :param filename: The full path to the file
+    :param meson_version: The generating meson version
+    :return: a FunctionNode
+    """
+    args: T.List[mparser.BaseNode] = []
+    args.extend([
+        build.string(package.name),
+        build.string('rust'),
+    ])
+    kwargs: T.Dict[str, mparser.BaseNode] = {
+        'version': build.string(package.version),
+        # Always assume that the generated meson is using the latest features
+        # This will warn when when we generate deprecated code, which is helpful
+        # for the upkeep of the module
+        'meson_version': build.string(f'>= {env.coredata.version}'),
+        'default_options': build.array([build.string(f'rust_std={package.edition}')]),
+    }
+    if package.license:
+        kwargs['license'] = build.string(package.license)
+    elif package.license_file:
+        kwargs['license_files'] = build.string(package.license_file)
+
+    return build.function('project', args, kwargs)
+
+
+def _convert_manifest(raw_manifest: manifest.Manifest, subdir: str, path: str = '') -> Manifest:
+    # This cast is a bit of a hack to deal with proc-macro
+    lib = _fixup_raw_mappings(raw_manifest.get('lib', {}))
+
+    # We need to set the name field if it's not set manually,
+    # including if other fields are set in the lib section
+    lib.setdefault('name', raw_manifest['package']['name'])
+
+    pkg = T.cast('manifest.FixedPackage',
+                 {fixup_meson_varname(k): v for k, v in raw_manifest['package'].items()})
+
+    return Manifest(
+        Package(**pkg),
+        {k: Dependency.from_raw(v) for k, v in raw_manifest.get('dependencies', {}).items()},
+        {k: Dependency.from_raw(v) for k, v in raw_manifest.get('dev-dependencies', {}).items()},
+        {k: Dependency.from_raw(v) for k, v in raw_manifest.get('build-dependencies', {}).items()},
+        Library(**lib),
+        [Binary(**_fixup_raw_mappings(b)) for b in raw_manifest.get('bin', {})],
+        [Test(**_fixup_raw_mappings(b)) for b in raw_manifest.get('test', {})],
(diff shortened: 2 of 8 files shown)


eli-schwartz:
> uses `tomli` instead of `toml`. `tomli` is on track to be in the standard library of python 3.11, so it will only be an external dependency for older versions of python

toml is already in the standard library of CPython from git, but using the name `tomllib`, not `tomli`.

We can, of course, start depending on that once our minimum supported version of python is python 3.11 (which isn't even released today!), this will probably happen in 2026 once python 3.10 is end of lifed.

Incidentally, that's why once again, it is *really* disappointing that CPython chose to standardize a (bad) configuration file format in 2016, then take until 2022 to put a parser for it into the stdlib, including some pretty silly bikeshedding back before the 3.10 release (we could have gotten it a year earlier, and relied on it being ubiquitous in 2025).

eli-schwartz:
@jpakkane please clarify if we can depend on a third-party module and raise a

```python
MesonException(f'Could not build cargo subproject {subp_name} because tomli is not available')
```

if it is not installed.

My vote is "no"...

dcbaker:
I think it would be a pretty bad decision to block cargo support over a module that will be optional for some versions of python, but core in others. Getting cargo support is a *must have* for Mesa, ASAP. We have two Rust subsystems pending, both of which need this feature. And Cargo is, whether we like it or not, pretty much a hard requirement for Rust use, there is basically nothing in the standard library and the language expects you to use crates. But this is in a module, and no one is required to use it. I've gone to great lengths to ensure that nothing except the cargo feature fails if you don't have tomli (or tomllib as it's going to be called in 3.11).

eli-schwartz:
pip bundles tomli precisely because they too have a zero-dependencies policy, but a different idea about how to solve that.

So having pip does NOT mean you have tomli.

> I think it would be a pretty bad decision to block cargo support over a module that will be optional for some versions of python, but core in others.

Keeping in mind *also* that this will only be a core module in an *unreleased* version of python, I could hear an argument that it would be acceptable to merge this PR, with a dependency on a third-party PyPI module, only after the October release of Python 3.11.

Before October, it is in fact a decision to block cargo support over a module that is required for cargo support in ALL versions of python.

However, thanks for updating the PR so that users of the python 3.11 alpha7 (due next month) will no longer need to install a third-party module. :)

dcbaker:
I've added support for yj as a fallback to tomli/tomllib, which converts the toml to json, then we load that with the json module. It all works, it's fine. Users of python 3.11 don't need it, and if you have tomli installed you don't need it either, but it provides a (slower) binary alternative.

dcbaker:
Also, the interface for `Popen_safe` is attrotious :(

xclaesse:
I did not follow the conversation too much, but we could bundle tomli in meson as fallback of importing the module from std failed?

xclaesse:
Also, any reason Mesa cannot just use cargo in a custom_target() or using external-project module? To be honest I find the cmake module barely usable and a *giant* maintaining burden for meson, I fear this cargo module is going to be the same. It is an immensely complicated solution that works only in very limited use cases.

dcbaker:
Because we’d have to treat cargo as always stale which makes incremental rebuilds really slow, for one. For two we have distro grumpiness about it, it adds another dependency, and cargo doesn’t respect no download.

cargo has a pretty big advantage over cmake in that it’s declarative (mostly) do you can actually translate it reliably. But most importantly, calling one build system from another is super unreliable and cumbersome. My last go at this showed me that Meson could compile Rust create chains faster than cargo could due to better parallelism. It means that we are sure that cargo different build a second copy of zlib being our backs, that it h on honors our Rust flags, and I suspect we’ll be support gcc-rs before cargo does, which matters for lot.

xclaesse:
> Because we’d have to treat cargo as always stale which makes incremental rebuilds really slow

I don't think that's true, it would invoke cargo which would be no-op, unless they are completely incompetent at doing a build system?

> we have distro grumpiness about it

That's distro's problem, I don't think that's a reason to add a HUGE maintaining burden on us.

> it adds another dependency, and cargo doesn’t respect no download.

First google result seems to say they can? https://www.ncameron.org/blog/cargo-offline/

dcbaker:
I’m in a mood right now, so I’ll wait till tomorrow to respond.

xclaesse:
> But most importantly, calling one build system from another is super unreliable and cumbersome

I don't think so, GStreamer-rs has been doing that with 90 lines script https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/blob/main/cargo_wrapper.py. Compaired to this PR that adds 1600 lines...

> My last go at this showed me that Meson could compile Rust create chains faster than cargo could due to better parallelism

That's bugs in cargo, fix it in cargo?

> It means that we are sure that cargo different build a second copy of zlib being our backs

OOC, how do you avoid that with this PR? you can translate c-cargo doing pkg-config to a meson dependency() call? Note that with external-project that's already solved too by setting PKG_CONFIG_PATH to meson-uninstalled directory where it will find zlib-uninstalled.pc if it was built as subproject.

xclaesse:
Also, how do you deal with build.rs, my understanding was that we cannot support crates that use that, can we? Are those crate rare enough?

xclaesse:
OOC, for mesa specifically, which crates are they going to depend on? If it's just a few selected crates, they could have native meson support just like we do with wrapdb.

Sorry to seems negative here, but I would like to be sure there are no easy alternative before committing to something like this. Cargo will evolve and we'll be playing catch-up all the time, we should not take this lightly neither. I understand that there are reasons to believe it won't be as a giant mess as cmake module, but I got really sceptical.

eli-schwartz:
> That's bugs in cargo, fix it in cargo?

Do you know how long people have been begging cargo to fix it? It probably constitutes a hopeless design limitation.



> Also, how do you deal with build.rs, my understanding was that we cannot support crates that use that, can we? Are those crate rare enough?

This PR explicitly states that it does not handle them and such crates must be manually converted to a subproject by writing a meson.build by hand.

Fortunately, the crates that don't need that are common enough that I do understand the need/desire for allowing Meson to support the sane, rational crates natively. I think that's what @dcbaker's use case in Mesa is, anyway.

mesonbuild/cargo/version.py
```diff
@@ -0,0 +1,63 @@
...
+                out.append(f'>= {".".join(v)}')
+            if len(v) == 2:
+                out.append(f'< {v[0]}.{int(v[1]) + 1}')
+            elif len(v) == 1:
+                out.append(f'< {int(v[0]) + 1}')
+
+        else:
+            # If there is no qualifier, then it means this or the next major version
+            # https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio
+            out.extend([f'>= {ver}', f'< {int(ver.split(".", maxsplit=1)[0]) + 1}'])
```
bonzini, line 60: If ver starts with zero and has a dot, then the next major version is identified by the second item of the version string. Something like:

```
out.append(f'>= {ver}')
v = ver.split('.')
if len(v) > 1 and v[0] == 0:
    out.append(f'< 0.{int(v[1]) + 1}')
else:
    out.append(f'< {int(v[0]) + 1}')
```

DemiMarie:
> Because we’d have to treat cargo as always stale which makes incremental rebuilds really slow, for one.

Cargo can generate a dependency file that Ninja can load.  I use this in [my Wayland compositor for Qubes OS](https://github.com/DemiMarie/qubes-wayland).

sdroege:
> I don't think so, GStreamer-rs has been doing that with 90 lines script https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/blob/main/cargo_wrapper.py. Compaired to this PR that adds 1600 lines...

Now that I'm reading through this again, I have to comment something about this specific part for posterity.

We're now at 300+ lines for that, it still has a lot of problems and solving many of them requires parsing Cargo.toml or duplicating information from Cargo.toml elsewhere. While it *works* it's fragile, slow and generally annoying to deal with. As maintainer of that project, I'm myself not using it because it just gets in my way all the time and I'm just using cargo directly. We're also not using it for building the plugins for the GStreamer binaries.

Experience with similar approaches in librsvg and the GNOME apps written in Rust are similarly bad.

Running cargo from meson works as a stopgap solution but is not useful in the long term, and something along the line of this PR is the way forward.

AFAIU this directly handles Cargo.toml internal, but I think for debuggability and also for getting a "template" to modify in case of complicated crates (-> build.rs) it would be useful to be able to also write out a meson.build + wrap file.

sdroege:
Also personally I think a tool in the shape of `cargo add` / `meson wrap install` (plus the corresponding `update`) to explicitly add dependencies from crates.io (or some other crates registry) would give for more smooth workflows and would be useful to add in addition (and integrated into meson, maybe even as part of the `meson wrap` subcommands`).

This here would still require to manually handle crate dependencies and add their code into your project without an easy way of updating.

xclaesse:
@sdroege I changed my mind since that old comment and I agree with you. Especially since your recent experiment, I totally support this PR, at least in the concept. I haven't read implementation thought.

bonzini:
My main worry (not an opposition to the PR - just a worry) is the impossibility of supporting `build.rs`. I would expect a lot of projects using C+Rust to use the libc crate or something else that ultimately uses a `build.rs` file.

sdroege:
@bonzini Those could get a proper meson build system, especially foundational crates like libc. Either as part of the meson wrapdb, or maybe even upstream if the crate maintainers want to deal with a second build system. Everything you can do in build.rs can be translated to meson, some patterns might be possible to support automatically (e.g. if it just uses the `system-deps` crate), and it might even be possible to actually re-use build.rs (with all the disadvantages that brings).

thiblahute:
I just rebased that branch, played with it and added support for a few things in [this branch](https://github.com/thiblahute/meson/commits/rust-cargo-subproject).  It adds `system-deps` support, and it is now possible to have meson fallback to pkg-config,/subprojects automatically. I fixed a few other things and I started porting the [gstreamer-plugin-meson-test](https://gitlab.freedesktop.org/slomo/gst-plugin-meson-test/)  project by @sdroege  in my [cargo_subprojects](https://gitlab.freedesktop.org/thiblahute/gst-plugin-meson-test/-/commits/cargo_subprojects) branch. Got `glib-sys-rs`, `gstreamer-sys-rs`,, `gstreamer-rs` and ~10 more deps using it now and enabling more is very simple but we have to handle features better as next step. I believe this MR is a solid basis to have proper support of Cargo subprojects!

thiblahute:
> AFAIU this directly handles Cargo.toml internal, but I think for debuggability and also for getting a "template" to modify in case of complicated crates (-> build.rs) it would be useful to be able to also write out a meson.build + wrap file.

It already outputs very readable meson.build files in the corresponding build folder. We should find a way to make writing .wrap files simpler still.

sdroege:
> handle features better as next step

This is also interesting. In cargo the features are (if we ignore executables) part of the dependency declaration. Conceptually, this would map to the `dependency()` function in meson, and maybe `default_options`.

Did you (or anybody else) have some thoughts how to map them nicely to meson?

> We should find a way to make writing .wrap files simpler still

"writing" seems like the wrong approach IMHO :) It should be as simple as a `meson wrap install --cargo some_crate_name`, unless customization is needed. If customization is needed then it would still be useful to have the basics be autogenerated though.

xclaesse:
> "writing" seems like the wrong approach IMHO :) It should be as simple as a meson wrap install --cargo some_crate_name

I experimented a bit that this weekend, stay tuned :)

thiblahute:
> > handle features better as next step
> 
> This is also interesting. In cargo the features are (if we ignore executables) part of the dependency declaration. Conceptually, this would map to the `dependency()` function in meson, and maybe `default_options`.
> 
> Did you (or anybody else) have some thoughts how to map them nicely to meson?

My idea is to have meson feature options for each cargo feature, and for optional dependency have them "auto" by default. Then in the call to `dependency()` pass them as `default_options`. Then the problem is that if you have 2 crates that use the same dep with different options it will fail, I am not even sure what happens in that case with Cargo?

> > We should find a way to make writing .wrap files simpler still
>
> "writing" seems like the wrong approach IMHO :) It should be as simple as a `meson wrap install --cargo some_crate_name`, unless customization is needed. If customization is needed then it would still be useful to have the basics be autogenerated though.

Right, bad  choice of word, generating was what I thinking of :-)

sdroege:
> Then the problem is that if you have 2 crates that use the same dep with different options it will fail, I am not even sure what happens in that case with Cargo?

Features in cargo are additive. The final build will have the union of all enabled features. Something like that is not possible with meson?

eli-schwartz:
No, because subprojects are only configured once, the first time you ask for it.

Regardless of what mechanism you use, at the time the subproject is first configured, it needs to know everything there is to know about setting it up. Once a build target is created, it can't be modified after the fact.

eli-schwartz:
Some mechanism to hoist options into the superproject's default_options.may be apropos.

xclaesse:
My general feeling after a first read:
- I did not read patch-by-patch because this needs squashing a lot.
- The whole mesonbuild/cargo/ directory is super clean and self contained, I like it a lot. I'm sure there are things that can be improved, but since it does not affect the rest of meson, I'm in the opinion to merge that directory as-is in a single commit. That will allow to get started on smaller incremental improvements in different PRs.
- The part that adds the `cargo` wrap method can be left out of the initial PR and either done after merging https://github.com/mesonbuild/meson/pull/11730, or even as part of that PR.
- mesonbuild/mparser.py has a few typing fixed/improvements that seems unrelated and should be moved to its own PR. That can be merged quickly.
- The part that adds `[cargo]` section in wrap files (afaik only in @thiblahute branch) definitely should be its own PR after the initial merge. I think there is more to discuss around that.

sdroege:
I agree that getting in a minimal version "early" and then iterating on that seems like the best way forward here. There are a lot of different aspects to the whole story and if we wait for a complete implementation in a single PR, that PR is going to take forever and be huge.

dcbaker:
Another thought I've had that's minimally invasive for upstream is to have some custom meson metadata that can be put in the `package.metadata` section, something like:
```toml
[package.metadata.meson]
ignore_build_rs = true
```

dcbaker:
> Features in cargo are additive. The final build will have the union of all enabled features. Something like that is not possible with meson?

I have thought a *lot* about the features mismatch problem. The only solution I could come up with was basically to have a promise like object that from the DSL looks like a normal subproject/dependency/whatever and then do the actual resolution right before handing off to the backend.

dcbaker:
If we're okay with going ahead and merging just the cargo module, I've pulled that out into a clean series with hopefully sane commit messages. I've made the following changes:

- Merged @thiblahute's fixes for various patches back into them
- Fixed a few issues I spotted
- Addressed @bonzini's review of the version function (and added a bunch of test cases)
- Fixed all issues spotted by mypy, pylint, and flake8
- added support for project(license_files)
- converted pure pylint tests into UnitTests and hooked them into the run_unittests script
- dropped everything that was not restricted purely to the cargo module

xclaesse (commented):
```
 cargo/cfg: Add a parser for the rust/cargo cfg() functions

TODO: the tests need to be ported away from pytest and put in the
unittest package
```

The TODO can be removed from commit message as it seems to be done already, right?

xclaesse:
CI failure seems relevant:
```
  File "/__w/meson/meson/mesonbuild/mparser.py", line 363, in set_kwarg_no_check
    self.kwargs[name] = value
TypeError: unhashable type: 'ArithmeticNode'
```


Other than that, good job, LGTM!

mesonbuild/cargo/interpreter.py
```diff
@@ -0,0 +1,416 @@
...
+
+if T.TYPE_CHECKING:
+    from types import ModuleType
+
+    from . import manifest
+    from ..environment import Environment
+
+# tomllib is present in python 3.11, before that it is a pypi module called tomli,
+# we try to import tomllib, then tomli,
+# TODO: add a fallback to toml2json?
```
eli-schwartz, line 33: I would go ahead and add the fallback path, and also add a custom error if tomllib is still none and we can't find the fallback executable.

eli-schwartz (changes requested):

mesonbuild/cargo/interpreter.py
xclaesse: IIRC in the original PR it checks if the module is none when doing a cargo subproject, and it's not fatal if required=false.

dcbaker: Yeah, my goal was originally that we'd basically do a check in the cargo subproject call that fails if required == true. but implementing a path for toml2json before landing this probably make sense

eli-schwartz: Ah okay I see, the part that catches the error is not included in this PR because it's been pared down to the mesonbuild/cargo/ directory only and that was part of mesonbuild/modules/.

xclaesse: What value does it add to have toml2json? One dependency or another does not make much difference, it just have to be installed with pip or as distro dependency. better not have too many code paths IMHO.

eli-schwartz: "installed with pip" would be the main thing I think. As a command-line tool there's a lot of ways to install it -- I know of at least 5 projects that all create an executable named `toml2json`, so you can take your pick of C++, rust, golang, or even python again. But the C++ one by @dcbaker is trivially bootstrapped.

xclaesse: I did not realize that's a CLI option. I guess that could be added, but I don't think that should be blocker to the initial merge.

xclaesse:
@dcbaker Some nodes are not hashable because `eq=True` and `frozen=False` in `@dataclass`. We can't set `frozen=True` because in that case even `__init__` cannot mutate the data class. I think in our case we can set `unsafe_hash=True` because it is immutable after init. That would ensure that `__hash__` match `__eq__`, using `id(self)`is not a great solution because it means that 2 nodes could have `__eq__` return True while `__hash__` is different.

What about keeping eq=False and implement `__eq__` and `__hash__` in BaseNode using id() ?

xclaesse:
@dcbaker OTOH, why do you even need to compare nodes at all?

mesonbuild/mparser.py
```diff
@@ -266,11 +266,19 @@ def accept(self, visitor: 'AstVisitor') -> None:
             if callable(func):
                 func(self)
 
+@dataclass(init=False)
```
eli-schwartz, line 269: init=False is not strictly needed since these classes have their own init methods; dataclasses is documented to ignore the value of init= and refrain from overwriting an existing method.

xclaesse: I gave that a try, but it's needed otherwise I get `TypeError: non-default argument 'value' follows default argument`. That's because BaseNode has default value for end_lineno and end_colno. This is arguably a python bug.

I would suggest that fix: https://github.com/mesonbuild/meson/commit/2f05c20a74074ff2080c83d9b21c7fd2d3f3e918

dcbaker:
I've updated it to generate simple hashes only where needed. the lint fails, because I need to tell it to ignore more stuff. yay mypy bugs. I just figured I'd put up a version that works and we can see how the approach is

mesonbuild/cargo/builder.py
```diff
@@ -0,0 +1,280 @@
...
+    """Create a method call.
+
+    :param name: the name of the method
+    :param id_: the object to call the method of
+    :param pos: a list of positional arguments, defaults to None
+    :param kw: a dictionary of keyword arguments, defaults to None
+    :return: a method call object
+    """
+    args = mparser.ArgumentNode(_token('array', id_.filename, 'what goes here?'))
+    args.arguments = pos
```
xclaesse, line 93: Wondering why mypy accepted that, but args.arguments cannot be None.

mesonbuild/cargo/builder.py
```diff
@@ -0,0 +1,280 @@
...
+             ) -> mparser.FunctionNode:
+    """Create a function call.
+
+    :param name: the name of the function
+    :param pos: a list of positional arguments, defaults to None
+    :param kw: a dictionary of keyword arguments, defaults to None
+    :return: a method call object
+    """
+    args = mparser.ArgumentNode(_token('array', filename, 'what goes here?'))
+    args.arguments = pos
```
xclaesse, line 111: Same here

dcbaker: good question. I've fixed it

xclaesse:
> I've updated it to generate simple hashes only where needed. the lint fails, because I need to tell it to ignore more stuff. yay mypy bugs. I just figured I'd put up a version that works and we can see how the approach is

I think we can go a bit further, what about this commit based on your latest changes: https://github.com/mesonbuild/meson/commit/765c66a77f7d4d78d297376db94532f58b97dde5

xclaesse:
> - Make ArgumentNode hashable because it is used for feature_key in
>   FeatureCheckBase.use(). Not sure how that was working since location
>   is always different?

Tracked that down to https://github.com/mesonbuild/meson/pull/9488 from @eli-schwartz. I think adding location to feature_key was wrong because it defeat the whole point of that code which is to not warn twice for the same deprecated/new feature in different locations.

eli-schwartz:
> Tracked that down to #9488 ([7a033c7](https://github.com/mesonbuild/meson/commit/7a033c78870e55b00d10a6499b40ad98b88a0509)) from @eli-schwartz. I think adding location to feature_key was wrong because it defeat the whole point of that code which is to not warn twice for the same deprecated/new feature in different locations.

No, I'm pretty sure we DO want to warn twice if you use a deprecated feature multiple times in a single project. We don't want *duplicate* warnings though, e.g. if a list of inputs is provided for a function and multiple elements trigger the same warning, we'd end up printing the same warning multiple times, identical line numbers and all.

xclaesse:
@eli-schwartz I see, I think your commit actualled fixed the FIXME part in the comment there, and that can be removed. But that cause a problem with @dcbaker changes here, because we made lineno/colno excluded from BaseNode hash/eq.

xclaesse:
@dcbaker now that mparser part got merged separately, LGTM after one last rebase :)

dcbaker:
@eli-schwartz I added initial support for using `toml2json`, should work with both my C++ implementation and with the toml2json Rust implementation, they both have the same interface. Didn't check the go or python implementations. This probably should be updated to somehow pass in an ExternalProgram, but as a first pass I think this is sufficient.

Rebased on top of @xclaesse's mparser cleanups

mesonbuild/cargo/builder.py
```diff
@@ -0,0 +1,283 @@
...
+
+def function(name: str, filename: str,
+             pos: T.Optional[T.List[mparser.BaseNode]] = None,
+             kw: T.Optional[T.Mapping[str, mparser.BaseNode]] = None,
+             ) -> mparser.FunctionNode:
+    """Create a function call.
+
+    :param name: the name of the function
+    :param pos: a list of positional arguments, defaults to None
+    :param kw: a dictionary of keyword arguments, defaults to None
```
eli-schwartz, line 109: Documentation for most params, but filename is missing. :D

This is of course better than the majority of functions throughout the codebase, which have no documentation.

mesonbuild/cargo/version.py
```diff
@@ -0,0 +1,90 @@
...
+                out.append(f'< {v[0]}.{int(v[1]) + 1}')
+            else:
+                out.append(f'< {int(v[0]) + 1}')
+
+        elif ver.startswith('^'):
+            # https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements
+            out.append(f'== {ver[1:]}')
+
+        elif '*' in ver:
+            # Rust has astrisk requirements,, which are like 1.* == ~1
```
eli-schwartz, line 47: Seems like this has to end with an asterisk, not just have it in the middle (which I guess would be entirely invalid).

mesonbuild/cargo/version.py
```diff
@@ -0,0 +1,90 @@
...
+            if v:
+                out.append(f'>= {".".join(v)}')
+            if len(v) == 2:
+                out.append(f'< {v[0]}.{int(v[1]) + 1}')
+            elif len(v) == 1:
+                out.append(f'< {int(v[0]) + 1}')
+
+        else:
+            # If there is no qualifier, then it means this or the next non-zero version
+            # That means that if this is `1.1.0``, then we need `>= 1.1.0` && `< 2.0.0`
```
eli-schwartz, line 59: This sounds superficially like a tilde but I don't understand rust versioning well enough. A comment explaining the difference (if any) would be great to have.

mesonbuild/cargo/interpreter.py
```diff
@@ -0,0 +1,446 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright © 2022-2023 Intel Corporation
+
+"""Interpreter for converting Cargo Toml definitions to Meson AST
+
+There are some notable limits here. We don't even try to convert something with
+a build.rs, there's so few limits on what Cargo alows a build.rs (basically
+none), and no good way for us to convert time. In that case, an actual meson port
```
eli-schwartz, line 8: typos:
- alow -> allow
- time -> them?

mesonbuild/cargo/interpreter.py
```diff
@@ -0,0 +1,446 @@
...
+    :return: a FunctionNode
+    """
+    args: T.List[mparser.BaseNode] = []
+    args.extend([
+        build.string(package.name),
+        build.string('rust'),
+    ])
+    kwargs: T.Dict[str, mparser.BaseNode] = {
+        'version': build.string(package.version),
+        # Always assume that the generated meson is using the lastest features
```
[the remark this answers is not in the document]
eli-schwartz: Also yes, wise move to future-proof this with setting an explicit meson_version for when we start emitting deprecation warnings by default. :D

mesonbuild/cargo/interpreter.py
```diff
@@ -0,0 +1,446 @@
...
+                        kw,
+                    ),
+                    f'dep_{fixup_meson_varname(name)}',
+                ),
+            ])
+
+    # Libs are always auto-discovered and there's no other way to handle them,
+    # which is unfortunate for reproducability
+    if os.path.exists(os.path.join(env.source_dir, cargo.subdir, cargo.path, 'src', 'lib.rs')):
+        ast.extend(_create_lib(cargo, build))
```
eli-schwartz, line 444: This isn't such a big deal, I hope, because at least wrapped crates coming from crates.io have versions and key off of the versions to do a complete meson reconfigure -- because the directory and all its contents have changed.

Maybe for the case of wrap-git we could warn here? Dunno if this is something to reliably detect.

Also, typo: should be reproducibility.

eli-schwartz (commented):
Overall looks decent to me. I've largely skimmed over exact implementation details because it takes domain specific knowledge I don't have to assert the correctness of the translation, but from a meson and style perspective I don't have anything to critique other than what I'm commenting below, and otherwise LGTM.

tl;dr a bunch of prose typos and one style nit regarding TypedDict.

mesonbuild/cargo/manifest.py
```diff
@@ -0,0 +1,229 @@
...
+from __future__ import annotations
+import typing as T
+
+from typing_extensions import Literal, TypedDict
+
+EDITION = Literal['2015', '2018', '2021']
+CRATE_TYPE = Literal['bin', 'lib', 'dylib', 'staticlib', 'cdylib', 'rlib', 'proc-macro']
+
+Package = TypedDict(
+    'Package',
```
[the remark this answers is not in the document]
dcbaker: they have keys with `-` in them (ie, `rust-version`). That' can't be represented with the class form because Pytho will interpret the `-` as the "minus operator" not as part of the string name.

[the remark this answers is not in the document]
eli-schwartz: Oh, wonderful...

What about FixedPackage, FixedDependency, FixedBuildTarget though, which seems to exist to fix that? :D

I do see on second look that you're using the class form as well, later down. I'm guessing these 3 are holdovers?

mesonbuild/cargo/version.py
dcbaker: There is no written requirement, but I just tested and `*.0` was rejected by cargo with a warning to the effect of "nothing can come after the asterisk"

mesonbuild/cargo/version.py
dcbaker: Sure, the basic difference is that tilde requirements start from the patch number and work up: `~1.2.3` == `>= 1.2.3, < 1.3.0` in Meson. Bare versions start at the major number and work down `1.2.3` == `>= 1.2.3, < 2.0.0`. I can add that to the inline comment.

eli-schwartz: Great, so, what nodejs calls caret versions. :roll_eyes:

mesonbuild/cargo/manifest.py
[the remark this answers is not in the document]
dcbaker: Would you prefer to maximize the class form, or to maximize the non-class form?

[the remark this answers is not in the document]
eli-schwartz: I think we should use the class form as much as we can.

mesonbuild/cargo/version.py
```diff
@@ -0,0 +1,90 @@
...
+            if len(v) == 3:
+                out.append(f'< {v[0]}.{int(v[1]) + 1}.0')
+            elif len(v) == 2:
+                out.append(f'< {v[0]}.{int(v[1]) + 1}')
+            else:
+                out.append(f'< {int(v[0]) + 1}')
+
+        elif ver.startswith('^'):
+            # https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements
+            out.append(f'== {ver[1:]}')
```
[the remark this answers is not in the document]
LingMan: There seems to be a misunderstanding here. A requirement of `^1.2.3` does not mean `= 1.2.3`, it means `>=1.2.3, <2.0.0`.

mesonbuild/cargo/manifest.py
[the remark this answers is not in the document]
eli-schwartz: Oh, now I saw your edit... sad.

mesonbuild/cargo/version.py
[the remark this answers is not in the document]
eli-schwartz: Yeah, I was just about to comment that I discovered the same thing when talking to someone else.

The current logic is therefore in error.

mesonbuild/cargo/manifest.py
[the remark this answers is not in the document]
dcbaker: I've updated with class form as much as possible. There's also some wierdness with inheritance and total=False apparently, since I tried to create some extra shared classes in the middle of stuff to save on things common between the fixed and raw varieties.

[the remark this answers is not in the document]
dcbaker: I also folded the PEP 655 stuff into the other patches, it didn't seem like anyone objected and it really did belong in the other patches.

mesonbuild/cargo/version.py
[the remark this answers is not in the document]
LingMan: Eh, looks pretty well documented to me. Almost everybody specifies requirements as `1.2.3` in the Rust world. I think I've seen a single instance of someone using `^1.2.3`.

mesonbuild/cargo/interpreter.py
```diff
@@ -0,0 +1,449 @@
...
+            raw = load_toml(filename)
+
+            raw_manifest = T.cast('manifest.Manifest', raw)
+            man = _convert_manifest(raw_manifest, subdir, m)
+            manifests[man.package.name] = man
+
+    return manifests
+
+
+def load_all_manifests(subproject_dir: str) -> T.Dict[str, Manifest]:
```
xclaesse, line 372: Why would we need to load all manifests? My intention was to have wrap files for Cargo subprojects with `method=cargo`, the same way I'm doing it for cmake there: https://github.com/mesonbuild/meson/pull/11730.

mesonbuild/cargo/interpreter.py
```diff
@@ -0,0 +1,449 @@
...
+    * Convert Dependndency versions from the cargo format to something meson
+      understands
+
+    :param d: The mapping to fix
+    :return: the fixed string
+    """
+    raw = {fixup_meson_varname(k): v for k, v in d.items()}
+    if 'version' in raw:
+        assert isinstance(raw['version'], str), 'for mypy'
+        raw['version'] = version.convert(raw['version'])
```
xclaesse, line 111: I started testing that code, and found a bug here. That changes the type of version from `str` to `List[str]`. It is later passed to `Package` constructor in `_convert_manifest()`. Mypy did not detect that issue, but that causes down the road to create a `StringNode` with a list value instead of str.

xclaesse: Actually, the actual problem is `project( version: )` kwarg cannot be a list. We should not call `version.convert()` in that case.

xclaesse: Suggestion that fix the issue:

```diff
diff --git a/mesonbuild/cargo/interpreter.py b/mesonbuild/cargo/interpreter.py
index a79cbf882..56e465b02 100644
--- a/mesonbuild/cargo/interpreter.py
+++ b/mesonbuild/cargo/interpreter.py
@@ -310,8 +310,10 @@ def _convert_manifest(raw_manifest: manifest.Manifest, subdir: str, path: str =
     # including if other fields are set in the lib section
     lib.setdefault('name', raw_manifest['package']['name'])
 
+    pkg = {fixup_meson_varname(k): v for k, v in raw_manifest['package'].items()}
+
     return Manifest(
-        Package(**_fixup_raw_mappings(raw_manifest['package'])),
+        Package(**pkg),
         {k: Dependency.from_raw(v) for k, v in raw_manifest.get('dependencies', {}).items()},
         {k: Dependency.from_raw(v) for k, v in raw_manifest.get('dev-dependencies', {}).items()},
         {k: Dependency.from_raw(v) for k, v in raw_manifest.get('build-dependencies', {}).items()},
```

mesonbuild/cargo/interpreter.py
```diff
@@ -0,0 +1,449 @@
...
+    kw['link_with'] = build.identifier('lib')
+
+    # FIXME: currently assuming that an rlib is being generated, which is
+    # the most common.
+    return [
+        build.assign(
+            build.function(
+                'static_library',
+                [
+                    build.string(fixup_meson_varname(cargo.package.name)),
```
xclaesse, line 400: I don't think we need to replace `-` in library name here.

mesonbuild/cargo/builder.py
```diff
@@ -0,0 +1,284 @@
...
+    """Build A NumberNode
+
+    :param value: the value of the number
+    :param filename: the file that the value came from
+    :return: A NumberNode
+    """
+    return mparser.NumberNode(_token('number', filename, value))
+
+
+def bool(value: builtins.bool, filename: str) -> mparser.BooleanNode:
```
xclaesse, line 49: Isn't this dangerous to have a function name that is a python keyword?  You appended an underscore to other such functions: `def or_()`.

xclaesse: Likewise for Builder method.

mesonbuild/cargo/interpreter.py
```diff
@@ -0,0 +1,449 @@
...
+    return manifests
+
+
+def _create_lib(cargo: Manifest, build: builder.Builder) -> T.List[mparser.BaseNode]:
+    kw: T.Dict[str, mparser.BaseNode] = {}
+    if cargo.dependencies:
+        ids = [build.identifier(f'dep_{n}') for n in cargo.dependencies]
+        kw['dependencies'] = build.array(
+            [build.method('get_variable', i, [build.string('dep')]) for i in ids])
+    kw['link_with'] = build.identifier('lib')
```
xclaesse, line 391: I think this makes the library link to itself. I think this is more correct: https://github.com/mesonbuild/meson/commit/efbb9ff31709865fa5869e91831ba225cc360d21.

xclaesse:
@dcbaker I finished my review based, just small issues found while playing with that code. I have more improvements locally, but they are additions that should be done after initial merge.

mesonbuild/cargo/interpreter.py
dcbaker: Pretty sure rust targets can't have `-` in their outputs

mesonbuild/cargo/builder.py
dcbaker: it's dangerous in as much that it would alias the builtin bool. But the intention is that you *don't* import as `from builder import ...` built just `import builder`, and then `builder.bool()` is perfectly safe to call because it's not aliasing the builtin. this is why the `builtins` module (as seen in the value) exists. I think the ergonomics of not having the `_` outweights the very small cost of having to use the `builtin` module inside of the `builder` module.

There is no cost to having an instance method called `bool`, since that has to be called with either `self` or a reference to the instance, ie `builder.bool()`

`or` is different because it's a keyword, and you *cannot* alias keywords. just like `assert` or `for`

mesonbuild/cargo/interpreter.py
dcbaker: Ah, you're right. Originally I was using the dict `|` operator inside the arugments, ie:
```python
return [
   static_library(...)
   declare_dependency(..., kw={link_with: ...} | kw
]
```
but that of course only works in python 3.10, and I screwed up pulling that out.

mesonbuild/cargo/version.py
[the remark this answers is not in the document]
dcbaker: yeah, that documentation is terribly worded "^1.2.3 is exactly equivalent to 1.2.3" sure sounds like `^1.2.3.` means `exactly 1.2.3`

mesonbuild/cargo/interpreter.py
dcbaker: I'm not sure how we warn about this here, since we don't (and I don't think should) know whether this is a wrap-git or not at this level

mesonbuild/cargo/interpreter.py
dcbaker: The code that uses this isn't in the tree currently, obviously, but it's to deal with the fact that you may need the same cargo project multiple times, this allows us to load all of the manifests the first time a cargo subproject is used, then refer back to those manifests later, either because you have some kind of parallel dependency or because you have two separate rust dependencies with the same dependency, ie:

```
C   C
|   |
B   D
 \ /
  A
```
or something like:
```
B        B
|        |
A        C
```

dcbaker: Obviously when we get to the point of being able to use cargo wraps then we'll need to update the approach somehow, but I think for the current solution this is fine, and it can be fixed when cargo wraps land

mesonbuild/cargo/interpreter.py
dcbaker: That isn't sufficient, we still have to cast to get pkg to be `FixedManifest` for mypy

mesonbuild/cargo/interpreter.py
xclaesse: I have cargo wraps working locally on top of this PR (and on top of https://github.com/mesonbuild/meson/pull/11730), that's why I noticed this. But I'm fine leaving that there for now and see later how that fits.

mesonbuild/cargo/interpreter.py
xclaesse: I thought Meson was already replacing dash to underscore when needed, but I was wrong: https://github.com/mesonbuild/meson/blob/master/mesonbuild/build.py#L2038. Note that dash is allowed for cdylib and staticlib, but we don't support those yet anyway. Let's ignore this then, you're right.

mesonbuild/cargo/builder.py
xclaesse: vscode highlight it wrong, that's how I noticed. But you're right, `bool()` is not a keyword, unlike `or`, that makes sense.

mesonbuild/cargo/manifest.py
```diff
@@ -0,0 +1,227 @@
...
+        'metadata': T.Dict[str, T.Dict[str, str]],
+        'default-run': str,
+        'autobins': bool,
+        'autoexamples': bool,
+        'autotests': bool,
+        'autobenches': bool,
+    },
+    total=False,
+)
+"""A description of the Package Dictionary."""
```
xclaesse, line 43: That docstring seems misplaced.

xclaesse: Or maybe not, when using the `Package = TypedDict()` syntax?

dcbaker: this is a wierdness of vscode. If you put it here vscode will attach the docstring if you hover over it, like if you open cargo/interpreter.py and hover over it. If you put the comment above it, then it doesn't. Maybe that's too clever

xclaesse: Ok, whatever works best... :)

mesonbuild/cargo/interpreter.py
xclaesse: Seems very unlikely that a project add or remove its lib, and even less likely to do so without touching Cargo.toml. I guess ninja cannot trigger a reconfigure based on the existence of a file? Or some kind of custom_target that runs meson reconfigure?

mesonbuild/cargo/manifest.py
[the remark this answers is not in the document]
xclaesse: I think this thread is solved, right?

mesonbuild/cargo/interpreter.py
xclaesse: Btw, shouldn't we have a `FixedManifest` class that have fields like `'package': FixedPackage``?

xclaesse: Nevermind, I understood that's what interpreter.Manifest is, not the same thing as manifest.Manifest.

mypy is happy with `pkg = T.cast('manifest.FixedPackage', {fixup_meson_varname(k): v for k, v in raw_manifest['package'].items()})` indeed.

mesonbuild/cargo/version.py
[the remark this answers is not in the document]
xclaesse: > Almost everybody specifies requirements as 1.2.3 in the Rust world. I think I've seen a single instance of someone using ^1.2.3.

That's probably true for Cargo.toml files, but crates.io uses it everywhere: https://crates.io/crates/gstreamer/0.20.5/dependencies.

Anyway, the fix should be trivial, drop first `^` char and fallthrought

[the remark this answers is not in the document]
xclaesse: ah, fixed already.

mesonbuild/cargo/version.py
```diff
@@ -0,0 +1,96 @@
...
+                out.append(f'< {v[0]}.{int(v[1]) + 1}.0')
+            elif len(v) == 2:
+                out.append(f'< {v[0]}.{int(v[1]) + 1}')
+            else:
+                out.append(f'< {int(v[0]) + 1}')
+
+        elif '*' in ver:
+            # Rust has astrisk requirements,, which are like 1.* == ~1
+            # https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#wildcard-requirements
+            v = ver.split('.')[:-1]
```
xclaesse, line 45: Should we do some sanity checks? This assumes that `*` is last char but you did `elif '*' in ver:`. Also assumes there are only at most 3 components. What happens with 1.2.3.4? Should we error out?

xclaesse: Also all `int()` will make Meson backtrace if version has letters like a.b.c.

dcbaker: I mean, cargo would have already rejected those things. Do we need to validate here, or just assume that cargo already has and move on?

mesonbuild/cargo/version.py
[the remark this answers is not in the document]
LingMan: > yeah, that documentation is terribly worded "^1.2.3 is exactly equivalent to 1.2.3" sure sounds like `^1.2.3.` means `exactly 1.2.3`

Ah, I see where you're coming from. If you didn't read the section above and assume that `1.2.3` means `= 1.2.3`, it can be confusing with the current wording.
What about this?:
> [...]. `^1.2.3` is exactly equivalent to `1.2.3`, which is documented above.

Would that resolve the ambiguity from your perspective?

> That's probably true for Cargo.toml files, but crates.io uses it everywhere: https://crates.io/crates/gstreamer/0.20.5/dependencies.

Fair. Forgot about that.

mesonbuild/cargo/version.py
xclaesse: Ok, let's assume it's a valid file for now at least.

dcbaker:
Ah, now we have the python dist bug

mesonbuild/cargo/interpreter.py
xclaesse: @dcbaker AFAIK this is the last blocker, right?

dcbaker: I guess? This is fixed in the latest version

xclaesse: Oh missed that you pushed a fix already. Thanks

eli-schwartz:
The code looks clean to me, thanks for adding the explanatory comments I requested. :) This should be fine to merge I guess. Just rebase on master to catch the CI fix.

PR #10152, opened 2022-03-20
project tests: log the reason why a test is skipped

eli-schwartz:
We expose a reason after the string 'MESON_SKIP_TEST', but it is actually ignored when running the test, so it is only useful as documentation and really might as well be a comment.

Make it even more useful by actually printing that string after the '[SKIPPED]' message.

Also, sometimes a test can be skipped for multiple reasons, and it would be useful to know which one occurred.

### DIFF
--- a/run_project_tests.py
+++ b/run_project_tests.py
@@ -1173,6 +1173,12 @@ class LogRunFuture:
 
 RunFutureUnion = T.Union[TestRunFuture, LogRunFuture]
 
+def test_emits_skip_msg(line: str) -> bool:
+    for prefix in {'Problem encountered', 'Assert failed', 'Failed to configure the CMake subproject'}:
+        if f'{prefix}: MESON_SKIP_TEST' in line:
+            return True
+    return False
+
 def _run_tests(all_tests: T.List[T.Tuple[str, T.List[TestDef], bool]],
                log_name_base: str,
                failfast: bool,
@@ -1281,10 +1287,19 @@ def _run_tests(all_tests: T.List[T.Tuple[str, T.List[TestDef], bool]],
         if result is None:
             # skipped due to skipped category skip or 'tools:' or 'skip_on_env:'
             is_skipped = True
+            skip_reason = 'not run because preconditions were not met'
             skip_as_expected = True
         else:
             # skipped due to test outputting 'MESON_SKIP_TEST'
-            is_skipped = 'MESON_SKIP_TEST' in result.stdo
+            for l in result.stdo.splitlines():
+                if test_emits_skip_msg(l):
+                    is_skipped = True
+                    offset = l.index('MESON_SKIP_TEST') + 16
+                    skip_reason = l[offset:].strip()
+                    break
+            else:
+                is_skipped = False
+                skip_reason = ''
             if not skip_dont_care(t):
                 skip_as_expected = (is_skipped == t.skip_expected)
             else:
@@ -1295,6 +1310,7 @@ def _run_tests(all_tests: T.List[T.Tuple[str, T.List[TestDef], bool]],
 
         if is_skipped and skip_as_expected:
             f.update_log(TestStatus.SKIP)
+            safe_print(bold('Reason:'), skip_reason)
             current_test = ET.SubElement(current_suite, 'testcase', {'name': testname, 'classname': t.category})
             ET.SubElement(current_test, 'skipped', {})
             continue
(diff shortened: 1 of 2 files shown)


eli-schwartz:
This is a small quality of life improvement I came up with while trying to help debug #10080

I eventually opted for #10137 there because actually making the test fail when a dependency should really be found, provides better error messages. Nevertheless, it would be nice to get *some* information out of a skipped test.

eli-schwartz:
This is a small quality of life improvement I came up with while trying to help debug #10080

I eventually opted for #10137 there because actually making the test fail when a dependency should really be found, provides better error messages. Nevertheless, it would be nice to get *some* information out of a skipped test.

eli-schwartz:
@mensinda it appears that the cmake subproject handling is broken based on this Windows CI failure.

Cmake succeeds, but emits a FATAL_ERROR as a print statement instead of actually failing, and then meson's configure stage succeeds but the test project fails at compile time.

This seems quite wrong, and I presume that real life projects will also successfully configure but then fail to compile?

eli-schwartz:
@mensinda ping?

mensinda:
Sorry for the late response. Sadly the Azure logs seem to be gone and I can't trigger a rerun. Can you rebase and push again?

mensinda:
@eli-schwartz Should be fixed by #10228

eli-schwartz:
Oh ha, that simple is it.

Why does cmake allow that to produce the wrong results? So strange. Is it just because it thinks that's a string instead of a parameter?

mensinda:
Who knows. Stuff like that is why I originally switched from CMake to Meson.

mensinda:
> Is it just because it thinks that's a string instead of a parameter?

I think so. `,` and `'` have no special meaning in CMake

eli-schwartz:
Beautiful. Thanks for tracking this down.

eli-schwartz:
Alright, new, more understandable hurdle. The test fails like this:

```
CMake Error at CMakeLists.txt:19 (message):
  MESON_SKIP_TEST: Unsupported Assembler Platform


-- Configuring incomplete, errors occurred!
See also "D:/a/1/s/b 4236002d13/subprojects/cmTest/__CMake_build/CMakeFiles/CMakeOutput.log".

CMake configuration: FAILED

test cases\cmake\25 assembler\meson.build:5:0: ERROR: Failed to configure the CMake subproject
```

I want to match it as an error message, not random text, but cmake has it on two lines. And I now wonder if it makes sense for Meson to log that final error while including any message that CMake emits. Maybe

```
ERROR: Failed to configure the CMake subproject with message: MESON_SKIP_TEST: Unsupported Assembler Platform
```

@mensinda thoughts?

run_project_tests.py
```diff
@@ -1230,19 +1236,20 @@ def _run_tests(all_tests: T.List[T.Tuple[str, T.List[TestDef], bool]],
...
     tests_canceled = False
 
-    # Optionally enable the tqdm progress bar
+    # Optionally enable the tqdm progress bar, but only if there is at least one TestRunFuture
     global safe_print
     futures_iter: T.Iterable[RunFutureUnion] = futures
-    try:
-        from tqdm import tqdm
-        futures_iter = tqdm(futures, desc='Running tests', unit='test')
+    if len(futures) > 2:
```
dcbaker, line 1242: Comment and code don’t match

eli-schwartz: futures is a RunFutureUnion, which should always have one `LogRunFuture` (either `Not running {name} tests.` or `Running {name} tests.`) and one or more `TestRunFuture`s -- I only care about the latter.

Arguably I could check how many of them are isinstance TestRunFuture, but since I "know" that LogRunFuture is only used once for one header per category, I took a shortcut and hardcoded that there should be more than 2 of the union of types.

Would you prefer switching from `len(futures)` to checking how many of them are specifically TestRunFuture?

eli-schwartz: Or perhaps that assumption could at least be commented better.

dcbaker: That would be good, otherwise looks good to me

eli-schwartz: Fixed.

eli-schwartz:
The tests all passed for 93cff694228e72fe871e29941e15afe64cfea5d7 and the force push only modified a comment line, so there's no need to wait for CI.

PR #10153, opened 2022-03-20
Fix missing ' in cutom_target docs

rusty-snake:


eli-schwartz:
Hmm, there is a missing "s" in the commit message, `cutom_target` -> `custom_target`. :p

Otherwise LGTM.

PR #10160, opened 2022-03-22
ninja backend: do not fatally error on compdb failure

eli-schwartz:
We print a warning if a compilation database isn't successfully generated, which is good, because that gives some visibility in case the user really wanted to use the compdb. But warnings default to being fatal with --fatal-meson-warnings, which is not so good, because this isn't a very important warning at all, and we'd rather not error out in such cases when building works fine and a random bonus IDE feature doesn't work.

Mark this particular warning as non-fatal.

Fixes side issue in https://github.com/mesonbuild/wrapdb/pull/343#issuecomment-1074545609


xclaesse:
LGTM

PR #10164, opened 2022-03-22
CustomTarget: Use option_overrides as Target base class

xclaesse:
commit b120f83b088e4bc30e54721c8d97531fe47003ba

    Replace backend.get_option_for_target() with target.get_option()
    
    That method had nothing specific to the backend, it's purely a Target
    method. This allows to cache the OptionOverrideProxy object on the
    Target instance instead of creating a new one for each option lookup.

commit 8cd3253fe282b4dab47c6a9dfb491acf5e2cfa86

    Pass environment down to base Target class


xclaesse:
@dcbaker quick fix for the per subproject compiler option PR.

xclaesse:
Actually, hold, I've got a better idea that I need for another pr anyway.

dcbaker (approved):

xclaesse:
@dcbaker I changed the implementation of this PR with a bigger refactor that I needed for another PR anyway. IMHO it's more consistent like that.

xclaesse:
@dcbaker Does this look good to you?

dcbaker (approved):

PR #10165, opened 2022-03-22
docs/rust: fix markdown title

dcbaker:
VSCode is perfectly happy with the trailing `---`, github only likes
`...` apparently.


eli-schwartz:
Correction: github doesn't care, but hotdoc does.

PR #10167, opened 2022-03-23
better sanity checks on target output names

eli-schwartz:
I started this journey wondering whether it was possible to create bad ninja targets that clash with builtin target names. Turns out it is possible, but only in very specify circumstances.

While looking at that and inspecting the sanity checking for custom_target's `output:` kwarg, I noticed that it does a lot of useful checks that aren't necessarily being done elsewhere, such as configure_file which turned out to not actually need the first fix.

So yeah, turns out that we need to do custom_target's output validator checks in several more places. Let's do that.

### DIFF
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -442,10 +442,10 @@ class NinjaBackend(backends.Backend):
         'benchmark', etc, and also for RunTargets.
         https://github.com/mesonbuild/meson/issues/1644
         '''
-        if dummy_outfile.startswith('meson-'):
+        if dummy_outfile.startswith('meson-internal__'):
             raise AssertionError(f'Invalid usage of create_phony_target with {dummy_outfile!r}')
 
-        to_name = f'meson-{dummy_outfile}'
+        to_name = f'meson-internal__{dummy_outfile}'
         elem = NinjaBuildElement(all_outputs, dummy_outfile, 'phony', to_name)
         self.add_build(elem)
 
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -434,17 +434,22 @@ class NinjaBackend(backends.Backend):
         self.introspection_data = {}
         self.created_llvm_ir_rule = PerMachine(False, False)
 
-    def create_target_alias(self, to_target):
-        # We need to use aliases for targets that might be used as directory
-        # names to workaround a Ninja bug that breaks `ninja -t clean`.
-        # This is used for 'reserved' targets such as 'test', 'install',
-        # 'benchmark', etc, and also for RunTargets.
-        # https://github.com/mesonbuild/meson/issues/1644
-        if not to_target.startswith('meson-'):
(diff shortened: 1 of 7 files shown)


eli-schwartz:
```
The Meson build system
Version: 0.62.99
Source dir: /home/runner/work/meson/meson/docs
Build dir: /home/runner/work/meson/meson/docs/_build
Build type: native build
Project name: Meson documentation
Project version: 1.0
Host machine cpu: x86_64
Program ./genrefman.py found: YES (/home/runner/work/meson/meson/docs/./genrefman.py)

meson.build:66:0: ERROR: Target names starting with 'meson-' are reserved for Meson's internal use. Please rename.

A full log can be found at /home/runner/work/meson/meson/docs/_build/meson-logs/meson-log.txt
```

@annacrombie lol, `meson-reference.3`...

eli-schwartz:
Trying to decide the best solution here. We're weird, we actually want targets named `meson-*` apparently ¯\_(ツ)_/¯ so, do we:
- allow them in subdirectories where they won't conflict with alias targets
- guarantee that all `meson-` prefixed targets are something even more unique such as `mesoninternal-{hash}` and then we can happily allow stuff with `[_.g-z]` in them?

eli-schwartz:
teeeeeeeechnically that is in fact the target name. It's the ninja target name, not the meson target name which isn't always the same thing.

eli-schwartz:
Okay, getting back to this... I think we can solve it via a multi-pronged approach.

mesonbuild/backend/ninjabackend.py
```diff
@@ -434,18 +434,23 @@ def __init__(self, build: T.Optional[build.Build], interpreter: T.Optional[Inter
...
-        # We need to use aliases for targets that might be used as directory
-        # names to workaround a Ninja bug that breaks `ninja -t clean`.
-        # This is used for 'reserved' targets such as 'test', 'install',
-        # 'benchmark', etc, and also for RunTargets.
-        # https://github.com/mesonbuild/meson/issues/1644
-        if not to_target.startswith('meson-'):
-            raise AssertionError(f'Invalid usage of create_target_alias with {to_target!r}')
-        from_target = to_target[len('meson-'):]
-        elem = NinjaBuildElement(self.all_outputs, from_target, 'phony', to_target)
+    def create_phony_target(self, all_outputs, dummy_outfile, rulename, phony_infilename, implicit_outs=None):
```
dcbaker, line 437: is there actually a case where `phony_infilename` isn't going to be `'PHONY'`?

dcbaker (approved):
One question where we might be able to drop a parameter name (or give it a default value?) otherwise LGTM

mesonbuild/backend/ninjabackend.py
eli-schwartz: I did this to have the same API for both `create_phony_target` and `NinjaBuildElement`, but it's probably nonsensical to use it for anything other than "phony", yes.

Ultimately this hack feels terrible, ninja is flawed. Someday I'd like to drop this function and just use NinjaBuildElement here...

PR #10168, opened 2022-03-23
Prefer cython3 over cython which could be python2

xclaesse:


tristan957:
This doesn't exist on Fedora.

dcbaker:
It's a debian-ism I think. This might break machien files if you have `[programs]cython=...` but also have cython3 installed, right?

eli-schwartz:
What's debian doing here exactly?

xclaesse:
> It's a debian-ism I think. This might break machien files if you have `[programs]cython=...` but also have cython3 installed, right?

Good question, would have to try. I can also invert the order, I don't care that much myself.

xclaesse:
> What's debian doing here exactly?

cython package that provide the `cython` executable depends on python2. There is a `cython3` package that installs `cython3` executable and depends on python3, but meson does not try to use that executable ATM.

xclaesse:
> This might break machien files if you have `[programs]cython=...` but also have cython3 installed, right?

@dcbaker Tested, if you add `[binaries] cython='/some/cython'` it will use that one still, regardless if you have cython or cython3 installed. Which makes sense.

eli-schwartz:
That sounds like something debian should solve via a machine file, and debian users should solve by avoiding debian's cython package like the plague and using cython from pip.

They should not be renaming every CLI application in existence with the digit of the python version it happens to be implemented in. Cython does not care which version of python it happens to be running with -- it can generate python2 code just fine from a python3 runtime, and probably do the reverse as well except no one cares.

If debian provided two versions of cython, one installed with python2 and one installed with python3, that's a debian bug.

It's a CLI application, the version of python it depends on is totally irrelevant.

Moreover, cython3 is a namespace collision and should refer to version 3 of cython instead of version 0.29 of cython. So they are breaking their own ability to offer two versions of cython, in order to instead offer two "builds" of cython.

xclaesse:
Blame Debian all you want, that's not going to change anything. It's not like it's a big deal for us, let's move on.

eli-schwartz:
Well, it's pretty maddening for the same reason as the other distro that doesn't provide `valac`, only `valac-0.54` and suchlike... maybe we should hardcode a list of all point releases of valac.

Also since it is a CLI tool we should not care whether it is python2, we should look for the canonical name first, and only after that look for maddening debianisms and alternate names.

xclaesse:
Switched the order since `cython` is indeed the canonical name.

PR #10170, opened 2022-03-23
Make compilers list per subproject

xclaesse:
Previously subprojects inherited languages already added by main
project, or any previous subproject. This change to have a list of
compilers per interpreters, which means that if a subproject does not
add 'c' language  it won't be able to compile .c files any more, even if
main project added the 'c' language.

This delays processing list of compilers until the interpreter adds the
BuildTarget into its list of targets. That way the interpreter can add
missing languages instead of duplicating that logic into BuildTarget for
the cython case.


nirbheek:
This is a good change, even though it _can_ cause breakage. The likely candidates are:

1. Projects that are _only_ used as subprojects, such as [libgd](https://gitlab.gnome.org/GNOME/libgd)
2. Projects that exercise platform-specific build code (such as Objective-C support) only when used as a subproject

In either case, it's been working purely due to bugs, and breakage is going to be rare. We should not worry about it.

dcbaker:
Does this mean that we run the compiler detection logic for every subproject?

xclaesse:
> This is a good change, even though it _can_ cause breakage.

It is indeed possible we break some projects. I hope that it's rare enough to get away with this.

eli-schwartz:
Obviously this will need release notes then! We should document that projects which were "broken but happened to usually work" might break, and how to detect and solve the problem.

xclaesse:
> Obviously this will need release notes then!

Good point, done.

dcbaker (approved):

PR #10171, opened 2022-03-23
Add support for cython_args

tristan957:
cython_args was previoously ignored by Meson.


xclaesse:
I think that commit fix it in a more generic way: https://github.com/mesonbuild/meson/pull/9379/commits/42daf32a8621e10416be6f74859a7927a09d9498. It's part of my PR to add NASM: https://github.com/mesonbuild/meson/pull/9379.

However, we still need `args += target.get_extra_args('cython')` and the unit test from this PR.

tristan957:
This PR _might_ be able to go in a 0.62.1...ping @jpakkane for clarification. Otherwise you should just steal what you need.

PR #10174, opened 2022-03-23
gnome module: properly fallback to gtk-update-icon-cache

ajacoutot:
Commit [a0cade8f](https://github.com/mesonbuild/meson/commit/a0cade8f1d1470cf3203baae16d181eb4df8e9fe) introduced a typo and wrongly checks for gtk4-update-icon-cache twice.
If gtk4-update-icon-cache (gtk4) is not found, look for gtk-update-icon-cache (gtk3) instead.

### DIFF
--- a/mesonbuild/modules/gnome.py
+++ b/mesonbuild/modules/gnome.py
@@ -371,7 +371,7 @@ class GnomeModule(ExtensionModule):
             prog = state.find_program('gtk4-update-icon-cache', required=False)
             found = isinstance(prog, build.Executable) or prog.found()
             if not found:
-                prog = state.find_program('gtk4-update-icon-cache')
+                prog = state.find_program('gtk-update-icon-cache')
             icondir = os.path.join(datadir_abs, 'icons', 'hicolor')
             script = state.backend.get_executable_serialisation([prog, '-q', '-t', '-f', icondir])
             script.skip_if_destdir = True


eli-schwartz (approved):
Oops, sorry...

xclaesse:
LGTM

lazka:
looks like the typo was also backported to 0.61.4, which means gtk3 now fails to build with that.

eli-schwartz:
We should, yes.

It is worth noting, that this isn't actually a problem on all systems. Particularly, it was never a problem on *my* system where I tested the original commit that introduced this regression...

... because gtk2, gtk3, and gtk4 all provide an update-icon-cache program, and they are all the *same* program with no meaningful difference. (Other than low-level code cleanups and minor bug fixes.) One might well ask why it's named something different at all -- the answer is almost certainly "there's no good reason for it, but we wanted people to be able to install gtk3 and gtk4 at the same time without having to remove a conflicting filename".

And my system has a distro package "gtk-update-icon-cache" which is a split package provided by gtk4, which provides gtk4-update-icon-cache (naturally), as well as gtk-update-icon-cache (a symlink to the gtk4 version). And gtk2, gtk3, gtk4 all depend on that standalone package (and gtk2/gtk3 have gtk-update-icon-cache removed from the package entirely).

So if either program name is installed, they both are. I actually do not have gtk4 installed at all, only gtk3, but my local tests of the typoed functionality worked because I had the gtk4 version of the update-icon-cache program...

eli-schwartz:
That's essentially what I said.

(Note as well that distros still have pain, because they need to distribute a single post_install package manager hook that triggers for all 3 versions of gtk, which means having the other 2 versions of gtk depend on the first version's program and strip out the redundant versions, rename or not. Some distros package the gtk4 version only, some the gtk3 version only, and some the gtk2 version only.)


Personally I'd add a feature option for whether to build and install it, and in the auto case, check if `find_program(prefix / bindir / 'gtk-update-icon-cache').found()` to determine whether to build a copy. Regular users won't end up with file conflicts because they will simply get the first one they installed, and distros can manually control which gtk version must provide it.

I guess it doesn't much matter, that boat has long since sailed anyway.

PR #10176, opened 2022-03-24
install_symlink: Handle $DESTDIR case for links with absolute path

3v1n0:
In case a link is pointing_to an absolute path and we are using $DESTDIR
we fail in case the target is missing.

This is incorrect because we may need to use an absolute path to an
already installed file that is in $DESTDIR.

So if an absolute target is not existing, check if we have such file in
$DESTDIR before failing for real.

### DIFF
--- a/mesonbuild/minstall.py
+++ b/mesonbuild/minstall.py
@@ -431,10 +431,12 @@ class Installer:
         append_to_log(self.lf, to_file)
         return True
 
-    def do_symlink(self, target: str, link: str, full_dst_dir: str, allow_missing: bool) -> bool:
+    def do_symlink(self, target: str, link: str, destdir: str, full_dst_dir: str, allow_missing: bool) -> bool:
         abs_target = target
         if not os.path.isabs(target):
             abs_target = os.path.join(full_dst_dir, target)
+        elif not os.path.exists(abs_target) and not allow_missing:
+            abs_target = destdir_join(destdir, abs_target)
         if not os.path.exists(abs_target) and not allow_missing:
             raise MesonException(f'Tried to install symlink to missing file {abs_target}')
         if os.path.exists(link):
@@ -604,7 +606,7 @@ class Installer:
             full_dst_dir = get_destdir_path(destdir, fullprefix, s.install_path)
             full_link_name = get_destdir_path(destdir, fullprefix, s.name)
             dm.makedirs(full_dst_dir, exist_ok=True)
-            if self.do_symlink(s.target, full_link_name, full_dst_dir, s.allow_missing):
+            if self.do_symlink(s.target, full_link_name, destdir, full_dst_dir, s.allow_missing):
                 self.did_install_something = True
 
     def install_man(self, d: InstallData, dm: DirMaker, destdir: str, fullprefix: str) -> None:


mesonbuild/minstall.py
```diff
@@ -431,10 +431,12 @@ def do_copyfile(self, from_file: str, to_file: str,
         append_to_log(self.lf, to_file)
         return True
 
-    def do_symlink(self, target: str, link: str, full_dst_dir: str, allow_missing: bool) -> bool:
+    def do_symlink(self, target: str, link: str, destdir: str, full_dst_dir: str, allow_missing: bool) -> bool:
         abs_target = target
         if not os.path.isabs(target):
             abs_target = os.path.join(full_dst_dir, target)
+        elif not os.path.exists(abs_target) and not allow_missing:
+            abs_target = os.path.join(destdir, abs_target[len(Path(abs_target).anchor):])
```
eli-schwartz, line 439: This looks unfortunate(ly complex) and should consider using destdir_join instead.

eli-schwartz (changes requested):

mesonbuild/minstall.py
3v1n0: Ok, done.
But also rewritten `destdir_join` to use `pathlib` as it was suggested long time ago

mesonbuild/scripts/__init__.py
```diff
@@ -12,10 +12,12 @@
...
 def destdir_join(d1: str, d2: str) -> str:
     # c:\destdir + c:\prefix must produce c:\destdir\prefix
-    if len(d1) > 1 and d1[1] == ':' \
-            and len(d2) > 1 and d2[1] == ':':
-        return d1 + d2[2:]
-    return d1 + d2
+    if d1 and os.path.isabs(d2):
+        # TODO: use removeprefix() when depending on Python 3.9
+        d2 = d2[len(Path(d2).anchor):]
+    return os.path.join(d1, d2)
```
xclaesse, line 23: I've been doing that in other places, seems cleaner? Or do you think it's not always what we want?
```
p = Path(d2)
os.path.join(d1, d2.relative_to(d2.root))
```

xclaesse: Dunno if that's better, just like that it's not playing with path as string, but your code is fine too I guess.

eli-schwartz:
> But also rewritten `destdir_join` to use `pathlib` as it was suggested long time ago

Please don't. We already offer a high level abstraction for this, whole using pathlib pays the cost of the pathlib abstraction while... simultaneously solving none of the ugliness of string slicing.

3v1n0:
> Please don't. We already offer a battle-tested high level abstraction for this, while using pathlib pays the cost of the pathlib abstraction and makes for an abstraction of an abstraction while... simultaneously solving none of the ugliness of string slicing.

Well, the comment on the code was mentioning the idea of using it, so I tried. But I'm happy to revert if there's no consesus.

eli-schwartz:
Well, "consider". :p

As it turns out, after consideration I don't think pathlib has a straightforward way to do what actually turns out to not be path-like semantics (at least as pathlib defines it). And it shows in the fact that the proposed implementation still has to drop back down to strings and do string slicing.

Anyway, the first patch is useful on its own and seems correct. Thanks.

3v1n0:
@eli-schwartz ok I've dropped the 2nd commit, I can propose it in another PR in case.

mesonbuild/scripts/__init__.py
3v1n0: Ok, moving the discussion in https://github.com/mesonbuild/meson/pull/10180

mesonbuild/minstall.py
```diff
@@ -431,10 +431,12 @@ def do_copyfile(self, from_file: str, to_file: str,
         append_to_log(self.lf, to_file)
         return True
 
-    def do_symlink(self, target: str, link: str, full_dst_dir: str, allow_missing: bool) -> bool:
+    def do_symlink(self, target: str, link: str, destdir: str, full_dst_dir: str, allow_missing: bool) -> bool:
```
eli-schwartz, line 434: Actually I wonder. We use both target and link, but we don't "use" full_destdir, we just need it to define the abs_target. I am not sure whether here or at the call site is maybe a better place to calculate this?
(I genuinely am not sure.)

eli-schwartz (approved):

mesonbuild/minstall.py
3v1n0: I was thinking about that, but `full_dst_dir` is still used by parent to create the directory, 

We could definitely do that under symlink, but we'd still need to bring the `DirMaker` with us

3v1n0:
Wondering: In such cases, and so when the absolute path is something that meson installed, shouldn't we try to convert the symlink to be relative instead?

eli-schwartz:
I would say in such a case, if the project developers specified an absolute path how do we know they want it to be turned into a relative one?

Probably the only case that would ever be true, they really want `fs_mod.relative_to()`.

eli-schwartz:
CI failure on Cygwin is caused by https://github.com/gcovr/gcovr/pull/576 -- yippee for strange Windows envs that cannot be baked into a CI image which is regularly updated and tested before being released as a new frozen image. Not only is it slow to repeatedly install everything, it also periodically changes the versions of everything installed, then breaks sometimes. ;)

All the rest of the CI passes, and the breakage is known and shortly to be fixed. Merging.

PR #10177, opened 2022-03-24
string: Add length() method

3v1n0:
While it was possible to access to the last elements of a string via
negative indexing, it was not possible easily iterate through it or to
check nicely whether it was empty or of an expected size (without comparisons
with another expected string).

Also it was not possible to iterate through it, now with the `range()`
method it would be possible to go char-by-char, but strings doesn't
expose their length so add a method to fill this gap.


3v1n0:
> As a matter of curiosity, what would you use this for?

Well, meson is getting quite flexible now and it allows to do lots of string manipulations, that can be even more complex when you use `fs.read()`, where you may want to go through the string and do manipulations that are almost impossible to do without being able to know what are your limits.

jpakkane:
> Well, meson is getting quite flexible now and it allows to do lots of string manipulations, that can be even more complex when you use fs.read()

As a rule of thumb, whenever you find yourself doing "coding" in your Meson script, the recommended approach is to not do that and instead write a Python script that has the actual logic and just call into that. Parsing the contents of files is _very strongly discouraged_.

What specific thing are you doing and is there some higher level functionality we could add instead to avoid doing bitbanging inside Meson files?

3v1n0:
In my case it's not related to _coding_, even though it could be, but I had cases in which I wanted go through a string (path) to do manipulations that were not possible using `Fs` methods (like searching for the common prefix of two of them).

And in general be able to iterate through them, thus the addition that I found to be relevant in many other cases that I couldn't recall now.

3v1n0:
> Just missing a release notes snippet, add a .md file in doc/markdown/snippets.

Done.

jpakkane:
> but I had cases in which I wanted go through a string (path) to do manipulations that were not possible using Fs methods (like searching for the common prefix of two of them)

So for this particular case what you actually needed was a function for computing the common prefix? Anything else?

PR #10178, opened 2022-03-24
string: Add missing FeatureNew tags to methods

3v1n0:
There are some new(er) methods that have not version reference, so add
the missing ones in order to be properly notified when targetting older
meson versions.

### DIFF
--- a/mesonbuild/interpreter/primitives/string.py
+++ b/mesonbuild/interpreter/primitives/string.py
@@ -108,6 +108,7 @@ class StringHolder(ObjectHolder[str]):
         return self.held_object.join(args[0])
 
     @noKwargs
+    @FeatureNew('str.replace', '0.58.0')
     @typed_pos_args('str.replace', str, str)
     def replace_method(self, args: T.Tuple[str, str], kwargs: TYPE_kwargs) -> str:
         return self.held_object.replace(args[0], args[1])
@@ -120,9 +121,12 @@ class StringHolder(ObjectHolder[str]):
     @noKwargs
     @typed_pos_args('str.strip', optargs=[str])
     def strip_method(self, args: T.Tuple[T.Optional[str]], kwargs: TYPE_kwargs) -> str:
+        if args[0]:
+            FeatureNew.single_use('str.strip with stripped characters string', '0.43.0', self.subproject)
         return self.held_object.strip(args[0])
 
     @noKwargs
+    @FeatureNew('str.substring', '0.56.0')
     @typed_pos_args('str.substring', optargs=[int, int])
     def substring_method(self, args: T.Tuple[T.Optional[int], T.Optional[int]], kwargs: TYPE_kwargs) -> str:
         start = args[0] if args[0] is not None else 0


mesonbuild/interpreter/primitives/string.py
```diff
@@ -108,6 +108,7 @@ def join_method(self, args: T.Tuple[T.List[str]], kwargs: TYPE_kwargs) -> str:
         return self.held_object.join(args[0])
 
     @noKwargs
+    @FeatureNew('str.replace', '0.58.0')
     @typed_pos_args('str.replace', str, str)
```
tristan957, line 112: my bad :)

mesonbuild/interpreter/primitives/string.py
```diff
@@ -118,11 +119,13 @@ def split_method(self, args: T.Tuple[T.Optional[str]], kwargs: TYPE_kwargs) -> T
         return self.held_object.split(args[0])
 
     @noKwargs
+    @FeatureNew('str.strip', '0.43.0')
```
eli-schwartz, line 122: This was actually added in commit 5e2278d6d4e2337c05b2f4e583dc8620b90e9add with version 0.5.0, however in version 0.43.0 it began to accept optional posargs.

Please remove the FeatureNew. Unlike typed_kwargs, typed_pos_args does not directly support "since" checks. So you'll need to manually code it as an if block inside the function definition.

mesonbuild/interpreter/primitives/string.py
```diff
@@ -108,6 +108,7 @@ def join_method(self, args: T.Tuple[T.List[str]], kwargs: TYPE_kwargs) -> str:
         return self.held_object.join(args[0])
 
     @noKwargs
+    @FeatureNew('str.replace', '0.58.0')
```
eli-schwartz, line 111: This actually used to have a check but in commit d93d01b6c5c98a77e057b914e8d66b01d2a10771 while refactoring to create primitives/string.py, it got accidentally dropped. Thanks for the catch!

eli-schwartz (changes requested):
Mostly lgtm, but one of these is wrong. See below.

mesonbuild/interpreter/primitives/string.py
eli-schwartz: No no no, see my review below. This was correct when you implemented it in commit c83106ee38048acbe737ef112e8d51c9b5bd42d7.

mesonbuild/interpreter/primitives/string.py
3v1n0: Ok aded single use, let me know.

mesonbuild/interpreter/primitives/string.py
```diff
@@ -120,9 +121,12 @@ def split_method(self, args: T.Tuple[T.Optional[str]], kwargs: TYPE_kwargs) -> T
     @noKwargs
     @typed_pos_args('str.strip', optargs=[str])
     def strip_method(self, args: T.Tuple[T.Optional[str]], kwargs: TYPE_kwargs) -> str:
+        if args[0]:
+            FeatureNew.single_use('str.strip with characters strip string-list', '0.43.0', self.subproject)
```
eli-schwartz, line 125: As a user I don't think I'd really understand what this message is trying to convey. What about "str.strip with positional arguments"?

eli-schwartz (changes requested):

mesonbuild/interpreter/primitives/string.py
xclaesse: As a developer, I don't understand it neither :D

eli-schwartz: As a developer I see "str.strip with characters", ignore the second half, and assume "characters" means "something as opposed to nothing". I can do that since I already expect that from the "str.strip" part. :D

eli-schwartz: I think it's unlikely we'll add another one, but if we did then the FeatureNew could be changed to state "with optional first/second positional parameter" so I don't think it's *that* confusing. :)

eli-schwartz: @xclaesse I am still not entirely sure what to think of this wording, so what do you think about it?

The PR currently uses:
```diff
-str.strip with characters strip string-list
+str.strip with stripped characters string
```

xclaesse: I would have said simply `str.strip with a positional argument`.

xclaesse: You don't really need to explain what the arg does, there is a doc for that. And if user have put an arg they know what it does already.

xclaesse: To be fair, this is nitpicking, I'm fine with whatever, not like anyone is going to hit that warning anyway, who uses 0.43?

eli-schwartz: Yeah, that's the wording I suggested too.

At the moment what I really just want is some sort of consensus on what (if anything) to suggest to @3v1n0.

eli-schwartz: @3v1n0

Is this PR ever going to switch to "with positional argument"? I would love to merge it someday. :)

Issue #10179, opened 2022-03-24
`c_compiler.get_define()` returns a string containing a whitespace even though the define is not defined


HaSa1002:
**Describe the bug**
`cc.get_define()` is expected to return an empty string if the define is not defined. However, using specific MSVC version this doesn't seem to be guaranteed.
This issue is based on the findings in https://github.com/mesonbuild/wrapdb/pull/350

MSVC seems to print random spaces. I think instead of using the current code:
```c
        #ifndef DEFINE
        # define DEFINE
        #endif
        "MESON_GET_DEFINE_DELIMITER"
DEFINE
```
```c
        #ifndef DEFINE
        "MESON_GET_DEFINE_NDEFINE"
        #endif
        "MESON_GET_DEFINE_DELIMITER"
DEFINE
```
may be more robust? (This is just something I noticed while reading logs.)

**To Reproduce**
Build the libpng wrap using msvc via the x64 native tools command prompt for VS 2019.

**Expected behavior**
`c_compiler.get_define()` should return an empty string if the define is not defined.

**system parameters**
* Native Build
* Windows 10
* Py: 3.10.2
* 0.61.2

I have to leave my computer now but I wanted to post this issue beforehand. I am happy to provide more info and improve the ticket when I am back.

nyorain:
I'm running into the same problem (meson 0.61.1 and msvc 2022)

PR #10182, opened 2022-03-25
Fix `c_compiler`.get_define parsing unintentional whitespace as defined variable

HaSa1002:
This is my attempt to fix https://github.com/mesonbuild/meson/issues/10179
I don't use python often and I had trouble setting up my dev environment for meson.

I feel like using this more explicit path should be robuster against wrong whitespaces.

### DIFF
--- a/mesonbuild/compilers/mixins/clike.py
+++ b/mesonbuild/compilers/mixins/clike.py
@@ -665,12 +665,15 @@ class CLikeCompiler(Compiler):
                    dependencies: T.Optional[T.List['Dependency']],
                    disable_cache: bool = False) -> T.Tuple[str, bool]:
         delim = '"MESON_GET_DEFINE_DELIMITER"'
+        not_found_delim = '"MESON_GET_DEFINE_NOT_DEFINED"'
         code = f'''
         {prefix}
         #ifndef {dname}
-        # define {dname}
+        {not_found_delim}
+        #else
+        {delim}\n{dname}
         #endif
-        {delim}\n{dname}'''
+        '''
         args = self.build_wrapper_args(env, extra_args, dependencies,
                                        mode=CompileCheckMode.PREPROCESS).to_native()
         func = functools.partial(self.cached_compile, code, env.coredata, extra_args=args, mode='preprocess')
@@ -680,6 +683,8 @@ class CLikeCompiler(Compiler):
             cached = p.cached
             if p.returncode != 0:
                 raise mesonlib.EnvironmentException(f'Could not get define {dname!r}')
+        if not_found_delim in p.stdout:
+            return '', cached
         # Get the preprocessed value after the delimiter,
         # minus the extra newline at the end and
         # merge string literals.


tristan957:
It would be nice to get a minimal reproducer to throw in `test cases` that way it doesn't regress.

HaSa1002:
I am on holiday for 1,5 weeks now. I know, I have seen some info about how to make tests but that is nothing I can do on the fly. Anyway, I *think* this should be enough to reproduce this behaviour:
```python
cc = meson.get_compiler('c')
assert(cc.get_define('THIS_IS_PROBABLY_NOT_DEFINED') == '')
assert(cc.get_define('__DATE__) != '')
```

tristan957:
Would that test fail on Meson master?

HaSa1002:
Well... I hope so. The first use case is basically the issue I encountered to begin with (see the linked issue and the linked pr in the linked issue). It *may* fail on some MSVC instances (it did reliably when meson was invoked using the msvc dev command prompt). Instead of redefining and relying on the delimiter to check if the define is defined, I just added another text that can be checked against. So the check itself is more robust. This is basically juggling with standard output.

tristan957:
I fixed the bugs in that snippet and neither of the asserts failed for me.

HaSa1002:
Do you invoked meson using the x64 native command prompt? I only reliably reproduced it there. One with GH Actions and using the dev command prompt. Alternately you can try to build the libpng wrap in a project (again if invoked using the dev tools).

xclaesse:
I can't reproduce this issue neither, something is weird with @HaSa1002 setup, as far as I know he's the only one who ever had that issue. Not saying he did anything wrong, but I feel we haven't found the real root cause here. It could be his Windows language has an effect, or a bug in a very specific MSVC version, some extra cflag... who knows...

gregsskyles:
FWIW, I can't reproduce this bug with the latest release either; I'm guessing there's something else besides just VS 2019 needed to cause the issue.
Here's a log of what I did:
```
C:\Users\GregSkyles>cd "C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\"

C:\Program Files (x86)\Microsoft Visual Studio\2019\Community>%comspec% /k "C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build\vcvars64.bat"
**********************************************************************
** Visual Studio 2019 Developer Command Prompt v16.10.2
** Copyright (c) 2021 Microsoft Corporation
**********************************************************************
[vcvarsall.bat] Environment initialized for: 'x64'

C:\Program Files (x86)\Microsoft Visual Studio\2019\Community>cd \Users\GregSkyles\source\test-vs-env

C:\Users\GregSkyles\source\test-vs-env>meson init test-vs-env
Using "test-vs-env" (name of current directory) as project name.
Using "test-vs-env" (project name) as name of executable to build.
Defaulting to generating a C language project.
Sample project created. To build it run the
following commands:

meson setup builddir
meson compile -C builddir


C:\Users\GregSkyles\source\test-vs-env>rem  edit meson.build to add asserts

C:\Users\GregSkyles\source\test-vs-env>type meson.build
project('test-vs-env', 'c',
  version : '0.1',

exe = executable('test_vs_env', 'test_vs_env.c',
  install : true)

cc = meson.get_compiler('c')
assert(cc.get_define('THIS_IS_PROBABLY_NOT_DEFINED') == '')
assert(cc.get_define('__DATE__') != '')

test('basic', exe)

C:\Users\GregSkyles\source\test-vs-env>meson setup build .
The Meson build system
Version: 0.62.0
Source dir: C:\Users\GregSkyles\source\test-vs-env
Build dir: C:\Users\GregSkyles\source\test-vs-env\build
Build type: native build
Project name: test-vs-env
Project version: 0.1
C compiler for the host machine: cl (msvc 19.29.30038.1 "Microsoft (R) C/C++ Optimizing Compiler Version 19.29.30038.1 for x64")
C linker for the host machine: link link 14.29.30038.1
Host machine cpu: x86_64
Fetching value of define "THIS_IS_PROBABLY_NOT_DEFINED" :
Fetching value of define "__DATE__" : "Mar 28 2022"
Build targets in project: 1

Found ninja-1.10.2 at "C:\Program Files\Meson\ninja.EXE"

C:\Users\GregSkyles\source\test-vs-env>
```

xclaesse:
That said, I think this PR is a good idea because it distinguish between undefined and defined with empty value which could be generally useful. At least I prefer that over the original idea of stdout.strip().

xclaesse:
CI fails with:

```
test cases\common\132 get define\meson.build:15:4: ERROR: Assert failed: _WIN32 value is 1
        
         instead of 1
```

That's weird... There is definitely something going on with whitespaces. I think it's because the value is inside the `#else` case now. Wondering if we should have a delimiter before **and after** the value to be sure?

xclaesse:
To be fair, nothing forbid the preprocessor to add whitespaces there if it wants, we are probably relying on UB here anyway.

PR #10184, opened 2022-03-25
Fix CMake deprecation warning generated from interpreter

tristan957:


jpakkane:
Fine by me. @mensinda what do you think?

mesonbuild/cmake/interpreter.py
```diff
@@ -784,7 +784,6 @@ def __init__(self, build: 'Build', subdir: Path, src_dir: Path, install_prefix:
         self.languages         = []  # type: T.List[str]
         self.targets           = []  # type: T.List[ConverterTarget]
         self.custom_targets    = []  # type: T.List[ConverterCustomTarget]
-        self.trace             = CMakeTraceParser('', Path('.'), self.env)  # Will be replaced in analyse
```
mensinda, line 787: Maybe `self.trace: T.Optional[CMakeTraceParser] = None` instead? I am not a huge fan of introducing `self.X` variables in methods other than `__init__` (though this would not be a deal breaker if my suggestion introduces other problems).

tristan957: Try latest change on for size.

mensinda (approved):
That's even better! LGTM once the CI is green

tristan957:
@dcbaker taught me that trick.

Issue #5202, opened 2019-04-02
gtk-doc fails with Darwin frameworks


jtojnar:
When building Pango 1.43.0 on Darwin under Nix and Meson 0.49.1 with ninja backend, it fails with

```
[0/1] Installing files.
Building documentation for pango
ERROR: Error in gtkdoc helper script:

ERROR: ['gtkdoc-scangobj', '--types=/private/tmp/nix-build-pango-1.43.0.drv-0/pango-1.43.0/build/docs/pango.types', '--module=pango', '--run=', '--cflags=…', '--ldflags=… -framework CoreFoundation ApplicationServices', …] failed with status 1
clang-5.0: error: no such file or directory: 'ApplicationServices'
```

Perhaps the `-framework` is missing before `ApplicationServices`.

Full build log: [ykppzv3j4vch369k9mb00zaj4mnwzvbv-pango-1.43.0.drv.txt](https://github.com/mesonbuild/meson/files/3035623/ykppzv3j4vch369k9mb00zaj4mnwzvbv-pango-1.43.0.drv.txt)

PR #10186, opened 2022-03-26
gnome: Fix gtk-doc when using multiple Apple frameworks

jtojnar:
The `-framework Foundation -framework CoreFoundation` ended up de-duplicated by OrderedSet into `-framework Foundation CoreFoundation`.

Fixes: https://github.com/mesonbuild/meson/issues/5202

### DIFF
--- a/mesonbuild/modules/gnome.py
+++ b/mesonbuild/modules/gnome.py
@@ -635,21 +635,19 @@ class GnomeModule(ExtensionModule):
             link_command.append('-l' + lib.name)
         return link_command, new_depends
 
-    def _get_dependencies_flags(
+    def _get_dependencies_flags_raw(
             self, deps: T.Sequence[T.Union['Dependency', build.BuildTarget, build.CustomTarget, build.CustomTargetIndex]],
             state: 'ModuleState',
             depends: T.Sequence[T.Union[build.BuildTarget, 'build.GeneratedTypes', 'FileOrString', build.StructuredSources]],
-            include_rpath: bool = False,
-            use_gir_args: bool = False,
-            separate_nodedup: bool = False
-            ) -> T.Tuple[OrderedSet[str], OrderedSet[str], OrderedSet[str], T.Optional[T.List[str]], OrderedSet[str],
+            include_rpath: bool,
+            use_gir_args: bool,
+            ) -> T.Tuple[OrderedSet[str], OrderedSet[T.Union[str, T.Tuple[str, str]]], OrderedSet[T.Union[str, T.Tuple[str, str]]], OrderedSet[str],
                          T.List[T.Union[build.BuildTarget, 'build.GeneratedTypes', 'FileOrString', build.StructuredSources]]]:
         cflags: OrderedSet[str] = OrderedSet()
-        internal_ldflags: OrderedSet[str] = OrderedSet()
-        external_ldflags: OrderedSet[str] = OrderedSet()
         # External linker flags that can't be de-duped reliably because they
-        # require two args in order, such as -framework AVFoundation
-        external_ldflags_nodedup: T.List[str] = []
+        # require two args in order, such as -framework AVFoundation will be stored as a tuple.
+        internal_ldflags: OrderedSet[T.Union[str, T.Tuple[str, str]]] = OrderedSet()
+        external_ldflags: OrderedSet[T.Union[str, T.Tuple[str, str]]] = OrderedSet()
         gi_includes: OrderedSet[str] = OrderedSet()
         deps = mesonlib.listify(deps)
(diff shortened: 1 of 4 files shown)


test cases/frameworks/10 gtk-doc/meson.build
```diff
@@ -31,9 +31,15 @@ libfoo = shared_library('foo', 'foo.c',
   dependencies: gobject,
 )
 
+deps = []
+if host_machine.system() == 'darwin'
+  deps += dependency('appleframeworks', modules : ['Foundation', 'CoreFoundation'])
```
jtojnar, line 36: This does not seem to reproduce the issue on CI. And since I do not have access to a Darwin machine, I am not sure how to make fail.

jtojnar:
It seems to work on aarch64-linux, at least when [backported to Meson 0.60.3](https://github.com/NixOS/nixpkgs/pull/160343/commits/0e78d578e643f353d6db74a8514bab28099760dc).

test cases/frameworks/10 gtk-doc/meson.build
eli-schwartz: If the test is being skipped, this is probably due to the macOS CI not having gtkdoc installed?

The truth is, we install surprisingly little in `.github/workflows/macos.yml`

eli-schwartz:
Okay, so the proof that this test fails on macOS before the fix is here: https://github.com/mesonbuild/meson/runs/5709205241?check_suite_focus=true

test cases/frameworks/10 gtk-doc/meson.build
jtojnar: Yeah, installing gtkdoc allows me to reproduce it on CI: https://github.com/mesonbuild/meson/runs/5709205241?check_suite_focus=true#step:5:1331

jtojnar:
For some reason gtk-doc [cannot find the XSL stylesheets](https://github.com/mesonbuild/meson/runs/5709331021?check_suite_focus=true#step:5:1181), even though it is [added to catalog](https://github.com/mesonbuild/meson/runs/5709331021?check_suite_focus=true#step:3:185).

jtojnar:
All green now.

eli-schwartz:
> Brew seems to install stuff to /usr/local, trying setting XML_CATALOG_FILES.

TBH that seems like a brew bug, it should be built to know how to find its own default files automatically. Probably worth a bug report to homebrew.

And on an unrelated note, I see that this now has conflicts, because apparently I touched the same code in another PR.

mesonbuild/modules/gnome.py
```diff
@@ -667,22 +665,20 @@ def _get_dependencies_flags(
...
+                        gi_includes.update(libdepflags[3])
+                        depends.extend(libdepflags[4])
+                extdepflags = self._get_dependencies_flags_raw(dep.ext_deps, state, depends, include_rpath,
+                                                               use_gir_args)
                 cflags.update(extdepflags[0])
                 internal_ldflags.update(extdepflags[1])
                 external_ldflags.update(extdepflags[2])
-                external_ldflags_nodedup += extdepflags[3]
-                gi_includes.update(extdepflags[4])
-                depends.extend(libdepflags[5])
```
jtojnar, line 685: Assuming this was a typo supposed to be `extdepflags`.

jtojnar:
> > Brew seems to install stuff to /usr/local, trying setting XML_CATALOG_FILES.
> 
> TBH that seems like a brew bug, it should be built to know how to find its own default files automatically. Probably worth a bug report to homebrew.

I opened an discussion about it: https://github.com/Homebrew/discussions/discussions/3122

mesonbuild/modules/gnome.py
eli-schwartz: Yes, copy paste error. We already added that above... We want to add these new ones too and need this typo fix in order to do it.

eli-schwartz:
Wait, so they already had two big reports for this bug and are now saying for a *third* time that it's the user's responsibility to set environment variables to configure a homebrew package into working out of the box?

Weird.

jtojnar:
Anything else to do here? The Windows CI failure seems unrelated (brownout).

xclaesse:
Meson master gets OOM killed when configuring GStreamer, git bisect is blaming this commit: 39623a36ca0dc757ea36960e0589a06ca798e5c1.

xclaesse:
Could it be that it extends a list with itself, or something?

jtojnar:
Sorry I did not get to this sooner, looks like you managed to fix it in the meanwhile https://github.com/mesonbuild/meson/pull/10216

PR #10189, opened 2022-03-26
Dub dependency configuration

rtbo:
Add the possibility to specify a configuration for Dub dependencies.
Requires to add new kwarg to `dependency`.

### DIFF
--- a/ci/run.ps1
+++ b/ci/run.ps1
@@ -91,6 +91,15 @@ python --version
 echo ""
 python -m pip --disable-pip-version-check install --upgrade pefile pytest-xdist pytest-subtests jsonschema coverage
 
+# Needed for some D tests
+if ($dmd) {
+  if ($env:arch -eq 'x64') { $dmdArch = 'x86_64' } else { $dmdArch = 'x86_mscoff' }
+  echo ""
+  echo "Installing some dependencies for D tests"
+  & dub run dub-build-deep --yes -- xlsx --compiler dmd --arch $dmdArch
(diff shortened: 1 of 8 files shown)


anadon:
Hello @rtbo, I'm new here and trying to help out a bit.  One of those things is kicking the tires on PRs that haven't updated in a bit.  Is this PR ready for review?

rtbo:
> Is this PR ready for review?

The code is ready, but I'm struggling a bit with the tests.
The main reason is that I can't add new Dub packages to the tests images, so I try to install Dub deps directly from the test meson scripts, which is at best an awful hack.

anadon:
Can you tell me how I could test this locally?

rtbo:
I reverted code that skips the failing tests. It was only failing on Windows, I think due to racing of different tests running Dub commands at the same time. (tests `d/14 dub with deps` and `d/16 dub config`).
You can run tests locally with `./run_project_tests.py --only d`

rtbo:
@dcbaker @eli-schwartz  This PR was stalled for a few month, but it should be OK for review now.
Main use case: use Dub dependencies that are applications, but have a library configuration (e.g. dub, dfmt ...).
Currently we can't use such dependencies because `application` is the default configuration. Now we can choose `library`.

rtbo:
Note: the tests only run on Windows.
On linux they are skipped either because Dub is not installed, or because GDC is used as a compiler.
To run the tests on Linux we should have Dub and Dmd/Ldc at the same time.
Each time I try to modify the image scripts I get weird unrelated errors though. The tests will kick-in as soon as those conditions are met (and will fail because uninstalled dub packages).
I'm not sure how I can act on this.

PR #10191, opened 2022-03-27
packaging: rework how pyinstaller gets its instructions

eli-schwartz:
Make use of pyinstaller hooks by creating a hook that updates how the `mesonbuild` import functions.

This is more or less the same as passing a bajillion arguments to pyinstaller's CLI, but allows the logic to be self-contained (and reusable). It becomes more obvious what parts of the process pertain to pyinstaller, and which parts pertain to MSI/pkg creation.

@jpakkane please confirm this still works. I was only able to perform... limited testing of this on Linux.

### DIFF
--- a/packaging/createmsi.py
+++ b/packaging/createmsi.py
@@ -39,49 +39,6 @@ def gen_guid():
     '''
     return str(uuid.uuid4()).upper()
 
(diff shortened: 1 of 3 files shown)


jpakkane:
Tested this and it still works.

PR #10194, opened 2022-03-27
i18n: don't let our maintainer targets run via wrapped env

eli-schwartz:
They are RunTargets because they are one-shot commands without outputs. But we implement them purely via our internal wrapper for gettext, so there is no reason to wrap them *again* in our internal wrapper for meson_exe and set a bunch of environment variables we know we absolutely do not need, use, or want.

This avoids the ugly "wrapped due to env" status, and allows users to directly see the command being run without going into despair at obscure pickled nonsense.

It also offers a tiny defense against upgrading Meson without reconfiguring. People should not do that, and we error out about this in a bunch of places, but `--internal gettext` has a perfectly stable interface just like most build tools that aren't part of Meson internals, since it uses command line arguments instead of pickling.

### DIFF
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -1445,11 +1445,12 @@ class Backend:
 
     def get_run_target_env(self, target: build.RunTarget) -> build.EnvironmentVariables:
         env = target.env if target.env else build.EnvironmentVariables()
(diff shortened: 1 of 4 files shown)


xclaesse:
LGTM

eli-schwartz:
Rebased for conflicts, no changes.

Issue #10172, opened 2022-03-23
`compiler.find_library` should always search `-L` dirs from `c*_link_args`


blue42u:
`compiler.find_library('x')` uses the compiler to find a matching library by passing `-lx` along with any flags specified by the user in `c_link_args` (or `cpp_link_args` etc. as appropriate). In contrast `compiler.find_library('x', static: ...)` manually searches for a matching library. In the former case `-L` flags in `c*_link_args` affect the search (as per the compiler), in the latter case `-L` flags do not.

This is a confusing inconsistency, and is a significant issue for dependencies that do not (and never will) provide `pkg-config` files.
Currently users cannot build projects if they cannot easily add such dependencies the system search paths (eg. they aren't packaged by the distro and root access is unavailable). While it's possible to mitigate this within individual projects (project-specific options or long support Meson code), IMHO it makes more sense for Meson to provide a globally-applicable solution.

Previous discussions on similar topics:
 - https://github.com/mesonbuild/meson/issues/109
 -  https://github.com/mesonbuild/meson/issues/2472#issuecomment-542559297 (and following discussion)

Those discussions suggested a new built-in option to add global search paths to `compiler.find_library`. Since the user would already need to synchronize such an option with `c*_args` anyway to provide include paths, IMHO it makes more sense to use `c*_link_args` where this requirement and behavior is already well-known and well-understood.

**Failing example**
```meson
# meson.build
project('test', 'c')

cc = meson.get_compiler('c')

foo_either = cc.find_library('foo', required: false)
foo_shared = cc.find_library('foo', static: false, required: false)
foo_static = cc.find_library('foo', static: true, required: false)

assert(foo_either.found() and foo_shared.found() and foo_static.found())
```
```c
// foo.c
void foo() {}
```
```console
$ gcc -o libfoo.so foo.c -shared -fPIC
$ gcc -c -o foo.o foo.c
$ ar rcs libfoo.a foo.o
$ meson -Dc_link_args="-L`pwd`" b
The Meson build system
Version: 0.61.2
Source dir: /tmp/foo
Build dir: /tmp/foo/b
Build type: native build
Project name: test
Project version: undefined
C compiler for the host machine: ccache cc (gcc 11.2.0 "cc (Debian 11.2.0-18) 11.2.0")
C linker for the host machine: cc ld.bfd 2.38
Host machine cpu family: x86_64
Host machine cpu: x86_64
Library foo found: YES
Library foo found: NO
Library foo found: NO

[...]

PR #10195, opened 2022-03-27
compiler.find_library: Also search lib dirs from link_args

blue42u:
`compiler.find_library` without a `static:` argument searches paths from `<lang>_link_args` (via the compiler), however adding `static:` removes these paths from the search. This inconsistency is confusing and complicates using external libraries without pkg-config/CMake support.

Make the `static:` case search paths from `*_link_args` as well for the C-like compilers.

Resolves mesonbuild/meson#10172


blue42u:
Rebased onto current master, no other change.

Volker-Weissmann:
@dcbaker Did you forget this one?

PR #10197, opened 2022-03-28
backends: Remove /Zi arg if equested to use /Z7

cwe-skov-dk:
If the user requests to use /Z7 debug information embedded in
object files, msvc generates a warning: (D9025 : overriding '/Zi' with '/Z7')

### DIFF
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -959,15 +959,15 @@ class Backend:
         # Add compile args added using add_global_arguments()
         # These override per-project arguments
         commands += self.build.get_global_args(compiler, target.for_machine)
-        # Using both /ZI and /Zi at the same times produces a compiler warning.
-        # We do not add /ZI by default. If it is being used it is because the user has explicitly enabled it.
-        # /ZI needs to be removed in that case to avoid cl's warning to that effect (D9025 : overriding '/ZI' with '/Zi')
-        if ('/ZI' in commands) and ('/Zi' in commands):
-            commands.remove('/Zi')
         # Compile args added from the env: CFLAGS/CXXFLAGS, etc, or the cross
         # file. We want these to override all the defaults, but not the
         # per-target compile args.
         commands += self.environment.coredata.get_external_args(target.for_machine, compiler.get_language())
+        # Using both /Z7 or /ZI and /Zi at the same times produces a compiler warning.
+        # We do not add /Z7 or /ZI by default. If it is being used it is because the user has explicitly enabled it.
+        # /Zi needs to be removed in that case to avoid cl's warning to that effect (D9025 : overriding '/Zi' with '/ZI')
+        if ('/Zi' in commands) and (('/ZI' in commands) or ('/Z7' in commands)):
+            commands.remove('/Zi')
         # Always set -fPIC for shared libraries
         if isinstance(target, build.SharedLibrary):
             commands += compiler.get_pic_args()


mesonbuild/backend/backends.py
```diff
@@ -959,10 +959,10 @@ def generate_basic_compiler_args(self, target: build.BuildTarget, compiler: 'Com
...
         # These override per-project arguments
         commands += self.build.get_global_args(compiler, target.for_machine)
-        # Using both /ZI and /Zi at the same times produces a compiler warning.
-        # We do not add /ZI by default. If it is being used it is because the user has explicitly enabled it.
-        # /ZI needs to be removed in that case to avoid cl's warning to that effect (D9025 : overriding '/ZI' with '/Zi')
-        if ('/ZI' in commands) and ('/Zi' in commands):
+        # Using both /Z7 or /ZI and /Zi at the same times produces a compiler warning.
+        # We do not add /Z7 or /ZI by default. If it is being used it is because the user has explicitly enabled it.
+        # /Zi needs to be removed in that case to avoid cl's warning to that effect (D9025 : overriding '/Zi' with '/ZI')
+        if (('/ZI' in commands) or ('/Z7' in commands)) and ('/Zi' in commands):
```
anarazel, line 965: I'd personally check /Zi first, given that the other part is a tad more expensive, but maybe that's my micro-optimizing database programmer side talking...

anarazel (commented):
Hm. Something doesn't quite seem to work - I still end up with /Zi in arguments, resulting in warnings like
```
[224/1623] Compiling C object src/port/libpgport_shlib.a.p/dirent.c.obj
cl : Command line warning D9025 : overriding '/Zi' with '/Z7'
```
Perhaps the debug arguments are added later than generate_basic_compiler_args()?

mesonbuild/backend/backends.py
cwe-skov-dk: I have fixed the order of comparisons.

For me this works fine, when I remove my changes, I get the compiler warnings. With my changes these warnings are quiet. Are you sure you are using the changed version of meson and did you build from scratch?

anarazel: > For me this works fine, when I remove my changes, I get the compiler warnings. With my changes these warnings are quiet. Are you sure you are using the changed version of meson and did you build from scratch?

Yes. The reason it doesn't work is that Z7 is only added in the subsequent line, via `self.environment.coredata.get_external_args(target.for_machine, compiler.get_language())`

cwe-skov-dk: Thanks for the info, your use pattern seems to differ to mine...

I have updated the pull request accordingly.

cwe-skov-dk: Is there any news on reviewing this change?

anarazel: I just tested this, and the issue looks to be fixed. Sorry that it took a while, I don't use windows much.

anarazel:
@eli-schwartz Seems this could be merged now?

eli-schwartz:
Small typo in the commit message: "equested" should have an "r" at the beginning.

Could you mention in the commit message why you are shifting the order around? That context is currently only available in the github discussion and commits should be able to stand on their own.

cwe-skov-dk:
I have updated the commit message.

cwe-skov-dk:
What do I do about the failing linux/Arch linux check?
Looking at the failure, it seems the test is feeding c-code to gfortran. I have seen this failure once before during working on this pull request, the Arch linux check seems to be unstable.

eli-schwartz:
It's a cuda bug and a known build flake so you just wait for me to kick the CI until it restarts. :)

Issue #6259, opened 2019-11-28
Regression breaking build of glib


xclaesse:
This PR in Meson master breaks build of glib: https://github.com/mesonbuild/meson/pull/6179 beacuse it adds many `\` in its `config.h`.

xclaesse:
The reason is glib does this:
```meson
gl_extern_inline = '''
/*some comment*/
#if
#define blablabla
#endif
'''
glib_conf.set('gl_extern_inline', gl_extern_inline)
```

It assumes that the multi line string will be added verbatim in the config.h

xclaesse:
@Jehan could you check if it can be fixed please?

xclaesse:
as far as I understand, it's the intended behaviour of the PR, so the only thing we can do is reverse the commit. Or assume we are not backward compatible and provide a way to support glib's use-case.

xclaesse:
I would recommend adding something like `set('var', value, multiline : true)`.

Jehan:
Can I get the accurate example of what glib was going and what they were hoping to get?

Because from the above example:
```
gl_extern_inline = '''
/*some comment*/
#if
#define blablabla
#endif
'''
glib_conf.set('gl_extern_inline', gl_extern_inline)
```

I think it would generate this in `config.h` (without the commit adding the `\`):

```
#define gl_extern_inline
/*some comment*/
#if
#define blablabla
#endif
```

Is it what they wanted to generate? If so, it looks like a very hack misuse of the `set()` function (where we want to actually set a key-value pair; and here they would use the wrong newlines as a trick to output random other code).

Jehan:
P.S.: if I understood well what glib is trying to do, I think a new function should be created for what they do. Something like:

glib_conf.add('''
#if bla
#define random code
with(multiline);
/*whatever*/
#endif
''')
```

But then keep the `set()` function for what it is (a key-value pair, which means in the particular case when this becomes a C macro that we have to add `\` in newlines).

xclaesse:
> I think it would generate this in `config.h` (without the commit adding the `\`):
> 
> ```
> #define gl_extern_inline
> /*some comment*/
> #if
> #define blablabla
> #endif
> ```

Exactly, it's a ugly hack to inject random code into the configured header. I agree ideally we should have a dedicated function for this, the problem now is backward compatibility. What I would recommend:
1) Add new function `conf.write(text)`
2) Add new kwarg `conf.set('var', 'foo\nbar', newline : true)`
[...]

PR #10199, opened 2022-03-28
Add snippets.symbol_visibility_header() method

xclaesse:
Defining public API in a cross platform library is painful, especially
on Windows. Since every library have to define pretty much the same
macros, better do it in Meson.

### DIFF
--- a/docs/markdown/Snippets-module.md
+++ b/docs/markdown/Snippets-module.md
@@ -1,0 +1,111 @@
+---
+short-description: Code snippets module
+...
+
+# Snippets module
+
+*(new in 1.10.0)*
+
+This module provides helpers to generate commonly useful code snippets.
+
+## Functions
+
+### symbol_visibility_header()
+
+```meson
+snippets.symbol_visibility_header(header_name,
+  namespace: str
+  api: str
+  compilation: str
+  static_compilation: str
+  static_only: bool
+)
+```
+
+Generate a header file that defines macros to be used to mark all public APIs
+of a library. Depending on the platform, this will typically use
+`__declspec(dllexport)`, `__declspec(dllimport)` or
+`__attribute__((visibility("default")))`. It is compatible with C, C++,
+ObjC and ObjC++ languages. The content of the header is static regardless
+of the compiler used.
+
+The first positional argument is the name of the header file to be generated.
+It also takes the following keyword arguments:
+
+- `namespace`: Prefix for generated macros, defaults to the current project name.
+  It will be converted to upper case with all non-alphanumeric characters replaced
+  by an underscore `_`. It is only used for `api`, `compilation` and
+  `static_compilation` default values.
+- `api`: Name of the macro used to mark public APIs. Defaults to `<NAMESPACE>_API`.
+- `compilation`: Name of a macro defined only when compiling the library.
+   Defaults to `<NAMESPACE>_COMPILATION`.
+- `static_compilation`: Name of a macro defined only when compiling or using
+  static library. Defaults to `<NAMESPACE>_STATIC_COMPILATION`.
+- `static_only`: If set to true, `<NAMESPACE>_STATIC_COMPILATION` is defined
+  inside the generated header. In that case the header can only be used for
+  building a static library. By default it is `true` when `default_library=static`,
+  and `false` otherwise. [See below for more information](#static_library)
+
+Projects that define multiple shared libraries should typically have
+one header per library, with a different namespace.
+
+The generated header file should be installed using `install_headers()`.
+
+`meson.build`:
+```meson
+project('mylib', 'c')
+subdir('mylib')
+```
+
+`mylib/meson.build`:
+```meson
+snippets = import('snippets')
+apiheader = snippets.symbol_visibility_header('apiconfig.h')
+install_headers(apiheader, 'lib.h', subdir: 'mylib')
+lib = library('mylib', 'lib.c',
+  gnu_symbol_visibility: 'hidden',
+  c_args: ['-DMYLIB_COMPILATION'],
+)
+```
+
+`mylib/lib.h`
+```c
+#include <mylib/apiconfig.h>
+MYLIB_API int do_stuff();
+```
+
+`mylib/lib.c`
+```c
+#include "lib.h"
+int do_stuff() {
+  return 0;
+}
+```
+
+#### Static library
+
+When building both static and shared libraries on Windows (`default_library=both`),
+`-D<NAMESPACE>_STATIC_COMPILATION` must be defined only for the static library,
+using `c_static_args`. This causes Meson to compile the library twice.
+
+```meson
+if host_system == 'windows'
+  static_arg = ['-DMYLIB_STATIC_COMPILATION']
+else
+  static_arg = []
+endif
+lib = library('mylib', 'lib.c',
+  gnu_symbol_visibility: 'hidden',
+  c_args: ['-DMYLIB_COMPILATION'],
+  c_static_args: static_arg
+)
+```
+
+`-D<NAMESPACE>_STATIC_COMPILATION` C-flag must be defined when compiling
+applications that use the library API. It typically should be defined in
+`declare_dependency(..., compile_args: [])` and
+`pkgconfig.generate(..., extra_cflags: [])`.
+
+Note that when building both shared and static libraries on Windows,
+applications cannot currently rely on `pkg-config` to define this macro.
+See https://github.com/mesonbuild/meson/pull/14829.
--- a/docs/markdown/snippets/symbol_visibility_header.md
+++ b/docs/markdown/snippets/symbol_visibility_header.md
@@ -1,0 +1,10 @@
+## New method to handle GNU and Windows symbol visibility for C/C++/ObjC/ObjC++
+
+Defining public API of a cross platform C/C++/ObjC/ObjC++ library is often
+painful and requires copying macro snippets into every projects, typically using
+`__declspec(dllexport)`, `__declspec(dllimport)` or
+`__attribute__((visibility("default")))`.
+
+Meson can now generate a header file that defines exactly what's needed for
+all supported platforms:
+[`snippets.symbol_visibility_header()`](Snippets-module.md#symbol_visibility_header).
--- a/mesonbuild/modules/snippets.py
+++ b/mesonbuild/modules/snippets.py
@@ -1,0 +1,101 @@
+# Copyright 2025 The Meson development team
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+import textwrap
+import typing as T
+
+from pathlib import Path
+
+from . import NewExtensionModule, ModuleInfo
+from ..interpreterbase import KwargInfo, typed_kwargs, typed_pos_args
+from ..interpreter.type_checking import NoneType
+from .. import mesonlib
+
+if T.TYPE_CHECKING:
+    from typing_extensions import TypedDict
+    from . import ModuleState
+
+    class SymbolVisibilityHeaderKW(TypedDict):
+        namespace: T.Optional[str]
+        api: T.Optional[str]
+        compilation: T.Optional[str]
+        static_compilation: T.Optional[str]
+        static_only: bool
+
+
+class SnippetsModule(NewExtensionModule):
+
+    INFO = ModuleInfo('snippets', '1.10.0')
+
+    def __init__(self) -> None:
+        super().__init__()
+        self.methods.update({
+            'symbol_visibility_header': self.symbol_visibility_header_method,
+        })
+
+    @typed_kwargs('snippets.symbol_visibility_header',
+                  KwargInfo('namespace', (str, NoneType)),
+                  KwargInfo('api', (str, NoneType)),
+                  KwargInfo('compilation', (str, NoneType)),
+                  KwargInfo('static_compilation', (str, NoneType)),
+                  KwargInfo('static_only', (bool, NoneType)))
+    @typed_pos_args('snippets.symbol_visibility_header', str)
+    def symbol_visibility_header_method(self, state: ModuleState, args: T.Tuple[str], kwargs: 'SymbolVisibilityHeaderKW') -> mesonlib.File:
+        header_name = args[0]
+        namespace = kwargs['namespace'] or state.project_name
+        namespace = mesonlib.underscorify(namespace).upper()
+        if namespace[0].isdigit():
+            namespace = f'_{namespace}'
+        api = kwargs['api'] or f'{namespace}_API'
+        compilation = kwargs['compilation'] or f'{namespace}_COMPILATION'
+        static_compilation = kwargs['static_compilation'] or f'{namespace}_STATIC_COMPILATION'
+        static_only = kwargs['static_only']
+        if static_only is None:
+            default_library = state.get_option('default_library')
+            static_only = default_library == 'static'
+        content = textwrap.dedent('''\
+            // SPDX-license-identifier: 0BSD OR CC0-1.0 OR WTFPL OR Apache-2.0 OR LGPL-2.0-or-later
+            #pragma once
+            ''')
+        if static_only:
+            content += textwrap.dedent(f'''
+                #ifndef {static_compilation}
+                #  define {static_compilation}
+                #endif /* {static_compilation} */
+                ''')
+        content += textwrap.dedent(f'''
+            #if (defined(_WIN32) || defined(__CYGWIN__)) && !defined({static_compilation})
+            #  define {api}_EXPORT __declspec(dllexport)
+            #  define {api}_IMPORT __declspec(dllimport)
+            #elif __GNUC__ >= 4
+            #  define {api}_EXPORT __attribute__((visibility("default")))
+            #  define {api}_IMPORT
+            #else
+            #  define {api}_EXPORT
+            #  define {api}_IMPORT
+            #endif
+
+            #ifdef {compilation}
+            #  define {api} {api}_EXPORT extern
+            #else
+            #  define {api} {api}_IMPORT extern
+            #endif
+            ''')
+        header_path = Path(state.environment.get_build_dir(), state.subdir, header_name)
+        header_path.write_text(content, encoding='utf-8')
+        return mesonlib.File.from_built_file(state.subdir, header_name)
+
+def initialize(*args: T.Any, **kwargs: T.Any) -> SnippetsModule:
+    return SnippetsModule()
--- a/docs/sitemap.txt
+++ b/docs/sitemap.txt
@@ -56,6 +56,7 @@ index.md
 			Qt6-module.md
 			Rust-module.md
 			Simd-module.md
+			Snippets-module.md
 			SourceSet-module.md
 			Windows-module.md
 			i18n-module.md
--- a/docs/theme/extra/templates/navbar_links.html
+++ b/docs/theme/extra/templates/navbar_links.html
@@ -26,6 +26,7 @@
             ("Qt6-module.html","Qt6"), \
             ("Rust-module.html","Rust"), \
             ("Simd-module.html","Simd"), \
+            ("Snippets-module.html","Snippets"), \
             ("SourceSet-module.html","SourceSet"), \
             ("Wayland-module.html","Wayland"), \
             ("Windows-module.html","Windows")]:
--- a/docs/yaml/functions/_build_target_base.yaml
+++ b/docs/yaml/functions/_build_target_base.yaml
@@ -255,7 +255,9 @@ kwargs:
       `default`, `internal`, `hidden`, `protected` or `inlineshidden`, which
       is the same as `hidden` but also includes things like C++ implicit
       constructors as specified in the GCC manual. Ignored on compilers that
-      do not support GNU visibility arguments.
+      do not support GNU visibility arguments. See also
+      [`snippets.symbol_visibility_header()`](Snippets-module.md#symbol_visibility_header)
+      method to help with defining public API.
 
   d_import_dirs:
     type: array[inc | str]
--- a/mesonbuild/interpreter/primitives/string.py
+++ b/mesonbuild/interpreter/primitives/string.py
@@ -7,7 +7,7 @@ import os
 
 import typing as T
 
-from ...mesonlib import version_compare, version_compare_many
+from ...mesonlib import version_compare, version_compare_many, underscorify
 from ...interpreterbase import (
     InterpreterObject,
     MesonOperator,
@@ -151,7 +151,7 @@ class StringHolder(ObjectHolder[str]):
     @noPosargs
     @InterpreterObject.method('underscorify')
     def underscorify_method(self, args: T.List[TYPE_var], kwargs: TYPE_kwargs) -> str:
-        return re.sub(r'[^a-zA-Z0-9]', '_', self.held_object)
+        return underscorify(self.held_object)
 
     @noKwargs
     @InterpreterObject.method('version_compare')
--- a/mesonbuild/utils/universal.py
+++ b/mesonbuild/utils/universal.py
@@ -152,6 +152,7 @@ __all__ = [
     'set_meson_command',
     'split_args',
     'stringlistify',
+    'underscorify',
     'substitute_values',
     'substring_is_in_list',
     'typeslistify',
@@ -1692,6 +1693,8 @@ def typeslistify(item: 'T.Union[_T, T.Sequence[_T]]',
 def stringlistify(item: T.Union[T.Any, T.Sequence[T.Any]]) -> T.List[str]:
     return typeslistify(item, str)
 
+def underscorify(item: str) -> str:
+    return re.sub(r'[^a-zA-Z0-9]', '_', item)
 
 def expand_arguments(args: T.Iterable[str]) -> T.Optional[T.List[str]]:
     expended_args: T.List[str] = []
--- a/run_mypy.py
+++ b/run_mypy.py
@@ -70,6 +70,7 @@ modules = [
     'mesonbuild/modules/qt6.py',
     'mesonbuild/modules/rust.py',
     'mesonbuild/modules/simd.py',
+    'mesonbuild/modules/snippets.py',
     'mesonbuild/modules/sourceset.py',
     'mesonbuild/modules/wayland.py',
     'mesonbuild/modules/windows.py',
--- a/run_project_tests.py
+++ b/run_project_tests.py
@@ -79,7 +79,7 @@ ALL_TESTS = ['cmake', 'common', 'native', 'warning-meson', 'failing-meson', 'fai
              'keyval', 'platform-osx', 'platform-windows', 'platform-linux', 'platform-android',
              'java', 'C#', 'vala', 'cython', 'rust', 'd', 'objective c', 'objective c++',
              'fortran', 'swift', 'cuda', 'python3', 'python', 'fpga', 'frameworks', 'nasm', 'wasm', 'wayland',
-             'format',
+             'format', 'snippets',
              ]
 
 
@@ -355,15 +355,15 @@ def setup_commands(optbackend: str) -> None:
 def platform_fix_name(fname: str, canonical_compiler: str, env: environment.Environment) -> str:
     if '?lib' in fname:
         if env.machines.host.is_windows() and canonical_compiler == 'msvc':
-            fname = re.sub(r'lib/\?lib(.*)\.', r'bin/\1.', fname)
+            fname = re.sub(r'lib/\?lib(.*)$', r'bin/\1', fname)
             fname = re.sub(r'/\?lib/', r'/bin/', fname)
         elif env.machines.host.is_windows():
-            fname = re.sub(r'lib/\?lib(.*)\.', r'bin/lib\1.', fname)
+            fname = re.sub(r'lib/\?lib(.*)$', r'bin/lib\1', fname)
             fname = re.sub(r'\?lib(.*)\.dll$', r'lib\1.dll', fname)
             fname = re.sub(r'/\?lib/', r'/bin/', fname)
         elif env.machines.host.is_cygwin():
             fname = re.sub(r'lib/\?lib(.*)\.so$', r'bin/cyg\1.dll', fname)
-            fname = re.sub(r'lib/\?lib(.*)\.', r'bin/cyg\1.', fname)
+            fname = re.sub(r'lib/\?lib(.*)$', r'bin/cyg\1', fname)
             fname = re.sub(r'\?lib(.*)\.dll$', r'cyg\1.dll', fname)
             fname = re.sub(r'/\?lib/', r'/bin/', fname)
         else:
@@ -1145,6 +1145,7 @@ def detect_tests_to_run(only: T.Dict[str, T.List[str]], use_tmp: bool) -> T.List
         TestCategory('wasm', 'wasm', shutil.which('emcc') is None or backend is not Backend.ninja),
         TestCategory('wayland', 'wayland', should_skip_wayland()),
         TestCategory('format', 'format'),
+        TestCategory('snippets', 'snippets'),
     ]
 
     categories = [t.category for t in all_tests]
--- a/test cases/snippets/1 symbol visibility header/main-static-only.c
+++ b/test cases/snippets/1 symbol visibility header/main-static-only.c
@@ -1,0 +1,3 @@
+#include <mylib/lib-static-only.h>
+
+int main(void) { return do_stuff(); }
--- a/test cases/snippets/1 symbol visibility header/main.c
+++ b/test cases/snippets/1 symbol visibility header/main.c
@@ -1,0 +1,3 @@
+#include <mylib/lib.h>
+
+int main(void) { return do_stuff(); }
--- a/test cases/snippets/1 symbol visibility header/meson.build
+++ b/test cases/snippets/1 symbol visibility header/meson.build
@@ -1,0 +1,13 @@
+project('symbol visibility header', 'c')
+
+sta_dep = dependency('mylib-sta', fallback: 'sub')
+exe = executable('exe-sta', 'main.c', dependencies: sta_dep)
+test('test-sta', exe)
+
+sha_dep = dependency('mylib-sha', fallback: 'sub')
+exe = executable('exe-sha', 'main.c', dependencies: sha_dep)
+test('test-sha', exe)
+
+static_only_dep = dependency('static-only', fallback: 'sub')
+exe = executable('exe-static-only', 'main-static-only.c', dependencies: static_only_dep)
+test('test-static-only', exe)
--- a/test cases/snippets/1 symbol visibility header/subprojects/sub/meson.build
+++ b/test cases/snippets/1 symbol visibility header/subprojects/sub/meson.build
@@ -1,0 +1,5 @@
+project('my lib', 'c')
+
+pkg = import('pkgconfig')
+
+subdir('mylib')
--- a/test cases/snippets/1 symbol visibility header/subprojects/sub/mylib/lib-static-only.c
+++ b/test cases/snippets/1 symbol visibility header/subprojects/sub/mylib/lib-static-only.c
@@ -1,0 +1,3 @@
+#include "lib-static-only.h"
+
+int do_stuff(void) { return 0; }
--- a/test cases/snippets/1 symbol visibility header/subprojects/sub/mylib/lib-static-only.h
+++ b/test cases/snippets/1 symbol visibility header/subprojects/sub/mylib/lib-static-only.h
@@ -1,0 +1,3 @@
+#include <mylib/apiconfig-static-only.h>
+
+MY_LIB_API int do_stuff(void);
--- a/test cases/snippets/1 symbol visibility header/subprojects/sub/mylib/lib.c
+++ b/test cases/snippets/1 symbol visibility header/subprojects/sub/mylib/lib.c
@@ -1,0 +1,3 @@
+#include "lib.h"
+
+int do_stuff(void) { return 0; }
--- a/test cases/snippets/1 symbol visibility header/subprojects/sub/mylib/lib.h
+++ b/test cases/snippets/1 symbol visibility header/subprojects/sub/mylib/lib.h
@@ -1,0 +1,3 @@
+#include <mylib/apiconfig.h>
+
+MY_LIB_API int do_stuff(void);
--- a/test cases/snippets/1 symbol visibility header/subprojects/sub/mylib/meson.build
+++ b/test cases/snippets/1 symbol visibility header/subprojects/sub/mylib/meson.build
@@ -1,0 +1,39 @@
+snippets = import('snippets')
+
+lib_incdir = include_directories('..')
+lib_args = ['-DMY_LIB_COMPILATION']
+lib_static_args = ['-DMY_LIB_STATIC_COMPILATION']
+
+h = snippets.symbol_visibility_header('apiconfig.h')
+install_headers(h, 'lib.h', subdir: 'mylib')
+mylib = both_libraries('mylib', 'lib.c',
+  include_directories: lib_incdir,
+  gnu_symbol_visibility: 'hidden',
+  c_args: lib_args,
+  c_static_args: lib_static_args,
+  install: true)
+mylib_sta_dep = declare_dependency(link_with: mylib.get_static_lib(),
+  include_directories: lib_incdir,
+  compile_args: lib_static_args)
+mylib_sha_dep = declare_dependency(link_with: mylib.get_shared_lib(),
+  include_directories: lib_incdir)
+meson.override_dependency('mylib-sta', mylib_sta_dep)
+meson.override_dependency('mylib-sha', mylib_sha_dep)
+pkg.generate(mylib,
+  extra_cflags: lib_static_args,
+)
+
+# When using static_only, we don't need lib_static_args because
+# MY_LIB_STATIC_COMPILATION gets defined in the generated header.
+h = snippets.symbol_visibility_header('apiconfig-static-only.h',
+  static_only: true)
+install_headers(h, 'lib-static-only.h', subdir: 'mylib')
+libstaticonly = static_library('static-only', 'lib-static-only.c',
+  include_directories: lib_incdir,
+  gnu_symbol_visibility: 'hidden',
+  c_args: lib_args,
+  install: true)
+static_only_dep = declare_dependency(link_with: libstaticonly,
+  include_directories: lib_incdir)
+meson.override_dependency('static-only', static_only_dep)
+pkg.generate(libstaticonly)
--- a/test cases/snippets/1 symbol visibility header/test.json
+++ b/test cases/snippets/1 symbol visibility header/test.json
@@ -1,0 +1,15 @@
+{
+  "installed": [
+    {"type": "file", "file": "usr/include/mylib/apiconfig-static-only.h"},
+    {"type": "file", "file": "usr/include/mylib/apiconfig.h"},
+    {"type": "file", "file": "usr/include/mylib/lib-static-only.h"},
+    {"type": "file", "file": "usr/include/mylib/lib.h"},
+    {"type": "file", "file": "usr/lib/libmylib.a"},
+    {"type": "expr", "file": "usr/lib/?libmylib?so"},
+    {"type": "implib", "file": "usr/lib/libmylib"},
+    {"type": "pdb", "file": "usr/bin/mylib"},
+    {"type": "file", "file": "usr/lib/libstatic-only.a"},
+    {"type": "file", "file": "usr/lib/pkgconfig/mylib.pc"},
+    {"type": "file", "file": "usr/lib/pkgconfig/static-only.pc"}
+  ]
+}


xclaesse:
Cases I'm unsure: clang-cl? cygwin? msys2 gcc?

dcbaker:
Question, why not put this in the windows module?

xclaesse:
> Question, why not put this in the windows module?

I did that initially, then saw that we have gnu visibility to deal with too. The pain is not limited to Windows.

tristan957:
@dcbaker this also defines the GCC/clang attributes.

docs/yaml/objects/compiler.yaml
```diff
@@ -586,3 +586,68 @@ methods:
...
+  description: |
+    Generate a header file that define a macro to be used to mark all public APIs
+    of a library. Depending on the type of compiler and the platform, this will
+    typically use `__declspec(dllexport)`, `__declspec(dllimport)` or
+    `__attribute__((visibility("default")))`. It is only supported for C and C++
+    compilers.
+
+    If this header is installed it should be included at the beginning of every
+    public headers. If it is not being installed, it should be included at the
+    beginning of every source files, and it is the responsability of every
```
jpakkane, line 619: After "every" you need to use singular rather than plural case. So "every public header", "every source file" and "every application".

xclaesse: Thanks, fixed.

docs/yaml/objects/compiler.yaml
```diff
@@ -586,3 +586,68 @@ methods:
...
+
+- name: symbol_visibility_header
+  returns: File
+  since: 0.63.0
+  posargs:
+    name:
+      type: str
+      description: Header file name.
+  kwargs:
+    api_variable:
```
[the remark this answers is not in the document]
xclaesse: I did not like _variable neither, so I just removed it. It's now `api`, `compilation` and `static_compilation`. Short and simple IMHO.

dcbaker:
Okay, then this should really live in the clike compiler mixin not in the base compiler

xclaesse:
> This is such a notable feature that it should get a mention in the proper docs as well, not only in the reference part.

Added a mention in `gnu_symbol_visibility` kwarg doc, do you have other places in mind?

> Okay, then this should really live in the clike compiler mixin not in the base compiler

I wasn't sure it's worth moving implementation into Compiler class, but since it needs different code for at least msvc and gnu, that would make sense indeed.

tristan957:
Thinking about naming consistency...I added `java.generate_native_headers()`. Wondering if I should have named it `java.native_headers()` instead similar to how this PR is `cc.symbol_visibility_header()`. No verb. Naming is hard :(

xclaesse:
Added `static_only` kwarg, when set to true _STATIC_COMPILATION gets defined in the header, so no need to define it manually in cflags and pkgconfig file any more. That's how glib does it.

Also I fixed the doc, installing the generated header is always required. I was wrong to think it was a project choice.

xclaesse:
> No verb. Naming is hard :(

I think most of the time we indeed avoid verbs in Meson API, there are very few get_foo(). But that's not entirely consistent.

tristan957:
Oh well, can always be fixed in a future release.

mesonbuild/interpreter/compiler.py
```diff
@@ -743,3 +747,68 @@ def get_supported_function_attributes_method(self, args: T.Tuple[T.List[str]], k
...
+        content = ''
+
+        if static_only:
+            content += textwrap.dedent(f'''\
+                #ifndef {static_compilation}
+                #  define {static_compilation}
+                #endif /* {static_compilation} */
+                ''')
+
+        if self.compiler.info.is_windows():
```
eli-schwartz, line 780: This is something I found really annoying about the cmake version, because you generate a file that isn't comprehensibly useful in the general case. What's the point in saving a few bytes for cross-platform defines?

mesonbuild/interpreter/compiler.py
```diff
@@ -743,3 +747,68 @@ def get_supported_function_attributes_method(self, args: T.Tuple[T.List[str]], k
...
+                ''')
+        elif self._has_argument_impl('-fvisibility=hidden'):
+            content += textwrap.dedent(f'''\
+                #ifndef {api}
+                #  ifdef {compilation}
+                #    define {api} __attribute__((visibility("default")))
+                #  else /* !{compilation} */
+                #    define {api}
+                #  endif /* !{compilation} */
+                #endif /* {api} */
```
eli-schwartz, line 802: This is incorrect, because if the compiler supports visibility=hidden then there is *never* any circumstance, period, end of story, in which it's bad or wrong to define it.

You do not need to leave it empty when using it as a public header for linking against. You SHOULD NOT leave it empty, because it is confusing to change cases around. Also, it perpetuates the erroneous belief that you need to do so, which means Meson is teaching bad behavior. :)

eli-schwartz (commented):
This is essentially cmake's `GENERATE_EXPORT_HEADER()`and incidentally shares my frustration with cmake's implementation.

The block of code in question can be statically determined in a cross-platform manner and checked into git, so generating it at build time seems silly... but worse, it teaches bad behavior to people who would check this sort of thing into git.

And the test suite even bends over backwards in order to avoid the real problem here, which is that MSVC doesn't support default_library=both because you cannot define the API as two different things entirely.

docs/yaml/objects/compiler.yaml
```diff
@@ -586,3 +586,95 @@ methods:
...
+    `meson.build`:
+    ```meson
+    project('mylib', 'c')
+    subdir('mylib')
+    ```
+
+    `mylib/meson.build`:
+    ```meson
+    default_library = get_option('default_library')
+    if host_machine.system() == 'windows' and default_library == 'both'
```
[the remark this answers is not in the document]
xclaesse: I think it is a DLL think whatever is the compiler, but could be wrong.

mesonbuild/interpreter/compiler.py
xclaesse: I'm not sure it is possible to have a static file that works in all cases tbh. How do you detect you're building a DLL, regardless of the compiler, from within a macro?

mesonbuild/interpreter/compiler.py
xclaesse: That's entirely subjective to judge if it's bad or not. You don't *need* to define the attribute, but you *can* define it if you want. Both are just equivalently good IMHO.

I copied this from how glib does it, but I don't mind changing it to always define the attribute if it's confirmed that it has no impact.

mesonbuild/interpreter/compiler.py
eli-schwartz: This review comment is responding to the fact that the file contents are generated by checking in meson code whether you're building for Windows or `GNUC`, instead of generating both everywhere and using preprocessor checks to determine which one is used.

Generating both everywhere and using preprocessor checks to determine which one is used, results in a header that serves as a good teaching example and can be extracted from any OS and embedded into a third-party derivation.

I have been previously annoyed that cmake generates Windows-specific snippets, which means that the example file it generates -- and which people try to use in PRs to the WrapDB -- cannot be reused in a Meson port.

mesonbuild/interpreter/compiler.py
eli-schwartz: No, it's objectively bad for exactly the social reasons I explained.

xclaesse:
> This is essentially cmake's `GENERATE_EXPORT_HEADER()`

I did not know cmake has this, do you have an example of project that uses that? I'm currious of what it contains.

eli-schwartz:
https://github.com/mesonbuild/wrapdb/pull/309

mesonbuild/interpreter/compiler.py
xclaesse: "social reasons" that's by definition subjective :D

xclaesse:
> [mesonbuild/wrapdb#309](https://github.com/mesonbuild/wrapdb/pull/309)

Lol, cmake does it even weirder, they define it twice the same way...

```
#    ifdef libhighs_EXPORTS
        /* We are building this library */
#      define LIBHIGHS_EXPORT __attribute__((visibility("default")))
#    else
        /* We are using this library */
#      define LIBHIGHS_EXPORT __attribute__((visibility("default")))
#    endif
```

mesonbuild/interpreter/compiler.py
xclaesse: Ah, I see, makes more sense, I'll update.

mesonbuild/interpreter/compiler.py
```diff
@@ -743,3 +747,68 @@ def get_supported_function_attributes_method(self, args: T.Tuple[T.List[str]], k
...
+    @FeatureNew('compiler.symbol_visibility_header', '0.63.0')
+    def symbol_visibility_header_method(self, args: T.List[str], kwargs: 'TYPE_kwargs') -> mesonlib.File:
+        if self.compiler.language not in {'c', 'cpp'}:
+            raise InterpreterException('compiler.symbol_visibility_header is supported only for C and C++ compilers')
+
+        header_name = args[0]
+
+        default_prefix = mesonlib.underscorify(self.interpreter.active_projectname.upper())
+        api = kwargs['api'] or f'{default_prefix}_API'
+        compilation = kwargs['compilation'] or f'{default_prefix}_COMPILATION'
```
tristan957, line 767: use kwargs.get() with a fallback instead

mesonbuild/interpreter/compiler.py
```diff
@@ -743,3 +747,68 @@ def get_supported_function_attributes_method(self, args: T.Tuple[T.List[str]], k
...
+                #  else /* !{static_compilation} */
+                #    ifdef {compilation}
+                #      define {api} extern __declspec(dllexport)
+                #    else /* !{compilation} */
+                #      define {api} extern __declspec(dllimport)
+                #    endif /* !{compilation} */
+                #  endif /* !{static_compilation} */
+                #endif /* {api} */
+                ''')
+        elif self._has_argument_impl('-fvisibility=hidden'):
```
tristan957, line 794: I think this check is better written as callling the implementation of `compiler.has_function_attribute('visibility:hidden')`

mesonbuild/interpreter/compiler.py
```diff
@@ -743,3 +747,68 @@ def get_supported_function_attributes_method(self, args: T.Tuple[T.List[str]], k
...
+        )
+    @typed_pos_args('compiler.symbol_visibility_header', str)
+    @FeatureNew('compiler.symbol_visibility_header', '0.63.0')
+    def symbol_visibility_header_method(self, args: T.List[str], kwargs: 'TYPE_kwargs') -> mesonlib.File:
+        if self.compiler.language not in {'c', 'cpp'}:
+            raise InterpreterException('compiler.symbol_visibility_header is supported only for C and C++ compilers')
+
+        header_name = args[0]
+
+        default_prefix = mesonlib.underscorify(self.interpreter.active_projectname.upper())
```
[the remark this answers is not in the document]
xclaesse: Fixed, I prepend an underscore in that case.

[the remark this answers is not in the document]
tristan957: Aren't identifiers starting with an `_` reserved for the C implementation?

mesonbuild/interpreter/compiler.py
xclaesse: I changed the code to have a common block for all platforms, but Meson still have to define `MESON_HOST_IS_WINDOWS` or `MESON_HAS_GNU_SYMBOL_VISIBILITY`. I don't think the latter can be determined by preprocessor, can it? The former could be replaced by something like `#if defined(WIN32) || defined(_WIN32) || defined(__WIN32__) || defined(__NT__)` but I'm actually not entirely sure what's the right case where we actually need dllexport/dllimport, clang-cl, cygwin?

xclaesse: What, actually there is an example there: https://gcc.gnu.org/wiki/Visibility.

xclaesse: Ok updated again, now without any check at meson level. Is that what you had in mind?

mesonbuild/interpreter/compiler.py
xclaesse: No, that's a trap of typed_kwargs, it sets None as default value.

mesonbuild/interpreter/compiler.py
xclaesse: I replaced that check with `__GNUC__ >= 4` proprocessor, which is what https://gcc.gnu.org/wiki/Visibility suggests.

mesonbuild/interpreter/compiler.py
xclaesse: Bah, changed to always set the attribute, that's what cmake does too.

mesonbuild/interpreter/compiler.py
eli-schwartz: IIRC we discussed this in the wrapdb at one point @xclaesse because a wrap you wrote turned out to be broken with has_argument, but worked with has_function_attribute.

docs/yaml/objects/compiler.yaml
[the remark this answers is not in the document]
xclaesse: Doc suggests that's indeed needed for GNU compilers too: https://gcc.gnu.org/wiki/Visibility. I replaced that Meson check by preprocessor check now anyway.

mesonbuild/interpreter/compiler.py
tristan957: Thanks for the lesson :)

mesonbuild/interpreter/compiler.py
eli-schwartz: Anyway, the `__GNUC__` preprocessor check is accurate as it declares whether the compiler supports the GNU C dialect including visibility attributes, so no worries there.

mesonbuild/interpreter/compiler.py
```diff
@@ -743,3 +747,59 @@ def get_supported_function_attributes_method(self, args: T.Tuple[T.List[str]], k
...
+            #    ifdef {static_compilation}
+            #      define {api} extern
+            #    else /* !{static_compilation} */
+            #      ifdef {compilation}
+            #        define {api} extern __declspec(dllexport)
+            #      else /* !{compilation} */
+            #        define {api} extern __declspec(dllimport)
+            #      endif /* !{compilation} */
+            #    endif /* !{static_compilation} */
+            #  elif __GNUC__ >= 4
```
tristan957, line 794: Feels like a trap to me when all compilers supported by Meson can tell you if they support this or not, but if others don't agree then it can just be fixed in a future bug report.

xclaesse:
> Okay, then this should really live in the clike compiler mixin not in the base compiler

@dcbaker done now.

mesonbuild/interpreter/compiler.py
xclaesse: That's what says GNU's own documentation...

xclaesse:
Finally passed the CI, ready for review.

tristan957:
I like the idea of this function existing in a `clike` module rather than on the compiler itself.

xclaesse:
Anyone for a review please?

docs/yaml/objects/compiler.yaml
```diff
@@ -586,3 +586,95 @@ methods:
...
+      default: false
+      description: |
+        If set to true, '<PROJECT_NAME>_STATIC_COMPILATION' is defined inside the
+        generated header.
+  description: |
+    Generate a header file that define a macro to be used to mark all public APIs
+    of a library. Depending on the type of compiler and the platform, this will
+    typically use `__declspec(dllexport)`, `__declspec(dllimport)` or
+    `__attribute__((visibility("default")))`. It is only supported for C and C++
+    compilers.
```
jpakkane, line 624: Instead of producing different files for different compilers, could we not make it always do the same thing with the appropriate ifdefs? It's not terribly complicated after all. Wit that change this would not need to live inside the compiler object, which is a slightly unintuitive place for it.

eli-schwartz: I said the same thing in my review above. @xclaesse acceded to my suggestion.

The actual implementation in `compiler.py` now writes the same file everywhere, after taking into account the fact that it inserts the string values of the method kwargs into that file.

Effectively, it is now `configure_file()` with a builtin template.

P.S. 

> It's not terribly complicated after all.

This is also why I believe it's best described in a FAQ rather than implemented as a Meson function. My personal feelings on the matter are that users should never use this -- it does pointless file I/O at build time for something that could be checked into git.

But apparently people really want to "have meson guarantee it is correct", for the same reason that when using cmake they want cmake to generate a similar file "to have cmake guarantee it is correct".

xclaesse: Yes, everyone gets it wrong, today I discovered that GLib got it wrong forever too. And it seems that even this PR got it wrong according to conversations I had with various people on the subject.

docs/yaml/objects/compiler.yaml
```diff
@@ -586,3 +586,95 @@ methods:
...
+      description: |
+        Name defined only when compiling or using static library.
+    static_only:
+      type: bool
+      default: false
+      description: |
+        If set to true, '<PROJECT_NAME>_STATIC_COMPILATION' is defined inside the
+        generated header.
+  description: |
+    Generate a header file that define a macro to be used to mark all public APIs
```
bugaevc, line 620:
```suggestion
    Generate a header file that defines a macro to be used to mark all public APIs
```

docs/yaml/objects/compiler.yaml
```diff
@@ -586,3 +586,95 @@ methods:
...
+    typically use `__declspec(dllexport)`, `__declspec(dllimport)` or
+    `__attribute__((visibility("default")))`. It is only supported for C and C++
+    compilers.
+
+    `<PROJECT_NAME>` in default values is the current project name in upper case
+    with all non-alphanumeric characters replaced by an underscore `_`.
+
+    The generated header file should be installed using `install_headers()`.
+
+    `static_only: true` only has an affect when building for Windows platform.
```
bugaevc, line 631:
```suggestion
    `static_only: true` only has an effect when building for Windows platform.
```

mesonbuild/interpreter/compiler.py
```diff
@@ -734,3 +743,57 @@ def get_supported_function_attributes_method(self, args: T.Tuple[T.List[str]], k
...
+    @typed_kwargs('compiler.symbol_visibility_header',
+        KwargInfo('api', (str, NoneType)),
+        KwargInfo('compilation', (str, NoneType)),
+        KwargInfo('static_compilation', (str, NoneType)),
+        KwargInfo('static_only', bool, default=False),
+        )
+    @typed_pos_args('compiler.symbol_visibility_header', str)
+    @FeatureNew('compiler.symbol_visibility_header', '0.63.0')
+    def symbol_visibility_header_method(self, args: T.Tuple[str], kwargs: 'SymbolVisibilityHeaderKW') -> mesonlib.File:
+        if self.compiler.language not in {'c', 'cpp'}:
```
bugaevc, line 756: Will this also work for Objective-C / Objective-C++?

xclaesse: I don't know how it works with objc, but if someone who knows can confirm it works, it can be added there, yes.

ePirat: It works the same, Objective-C is a superset of C, so all you can do there, you can in Objective-C too.

xclaesse: Ok, updated.

docs/yaml/objects/compiler.yaml
xclaesse: fixed

docs/yaml/objects/compiler.yaml
xclaesse: fixed

docs/yaml/objects/compiler.yaml
xclaesse: I now integrated the changes we made in glib since then, the macro is slightly different.

mesonbuild/interpreter/compiler.py
```diff
@@ -797,3 +806,54 @@ def preprocess_method(self, args: T.Tuple[T.List['mesonlib.FileOrString']], kwar
...
+
+        content = '#pragma once\n'
+        if static_only:
+            content += textwrap.dedent(f'''
+                #ifndef {static_compilation}
+                #  define {static_compilation}
+                #endif /* {static_compilation} */
+                ''')
+        content += textwrap.dedent(f'''
+            #if (defined(_WIN32) || defined(__CYGWIN__)) && !defined({static_compilation})
```
tristan957, line 839: So I would like to merge this given every project that ever exists need to do a configure_file() to get it right given the existence of static_compilation and compilation in this format string.

mesonbuild/interpreter/compiler.py
```diff
@@ -797,3 +806,54 @@ def preprocess_method(self, args: T.Tuple[T.List['mesonlib.FileOrString']], kwar
...
+
+        default_prefix = mesonlib.underscorify(self.interpreter.active_projectname.upper())
+        if default_prefix[0].isdigit():
+            default_prefix = f'_{default_prefix}'
+        api = kwargs['api'] or f'{default_prefix}_API'
+        compilation = kwargs['compilation'] or f'{default_prefix}_COMPILATION'
+        static_compilation = kwargs['static_compilation'] or f'{default_prefix}_STATIC_COMPILATION'
+        static_only = kwargs['static_only']
+
+        content = '#pragma once\n'
```
tristan957, line 831: I would rather not use `#pragma once` here. I think we should use include guards instead since they are much more universal.

mesonbuild/interpreter/compiler.py
tristan957: Hmm I guess I am an idiot, and these would be constants and not require configure_file() :D

mesonbuild/interpreter/compiler.py
```diff
@@ -797,3 +806,54 @@ def preprocess_method(self, args: T.Tuple[T.List['mesonlib.FileOrString']], kwar
...
+        if static_only:
+            content += textwrap.dedent(f'''
+                #ifndef {static_compilation}
+                #  define {static_compilation}
+                #endif /* {static_compilation} */
+                ''')
+        content += textwrap.dedent(f'''
+            #if (defined(_WIN32) || defined(__CYGWIN__)) && !defined({static_compilation})
+            #  define _{api}_EXPORT __declspec(dllexport)
```
tristan957, line 840: Why are these prefixed with underscore? Underscore-prefixed macros are reserved for the compiler

mesonbuild/interpreter/compiler.py
xclaesse: I never understood why people use header guards, as far as I know pragma once is universal and faster because it does not have to preprocess the whole file. It reads only the first line the first time it's included and next time it does not even open the file at all.

That said, I'm not against changing that here, should not make a real difference for a small file like that.

mesonbuild/interpreter/compiler.py
xclaesse: Right, you can typically commit that file into your project. I personally still believe it has value in meson because it can serve as reference. You can generate it with meson initially then commit the file. It can also be useful for meson unit tests, many of them has some variant of this, it's boring to copy that file everywhere. I'm yet to see a single project who does that correctly, delegating that to meson means every dev does not have to know all the details, they can just trust meson.

mesonbuild/interpreter/compiler.py
dnicolodi: I though that the double leading underscore is reserved, not the single one.

xclaesse: I think that as well, but if in doubt I can use triple underscore lol

mesonbuild/interpreter/compiler.py
tristan957: No sense in changing it until there is more consensus on this PR

mesonbuild/interpreter/compiler.py
tristan957: https://stackoverflow.com/a/17307796/7572728 has some good info

mesonbuild/interpreter/compiler.py
eli-schwartz: Since we already produce pragma once for configure_file, it makes sense to produce it for anything else too as there is no advantage to doing so *inconsistently*.

mesonbuild/interpreter/compiler.py
eli-schwartz: Why do they even need to be private? :D

xclaesse: > Why do they even need to be private? :D

Fair enough, removed the underscore. There is even a niche use-case for using `MYPROJECT_API_EXPORT`macro, glib does it for functions declarations that are exposed by modules: https://gitlab.gnome.org/GNOME/glib/-/blob/main/gio/giomodule.h#L130.

mesonbuild/interpreter/compiler.py
xclaesse: I didn't know there is precedent for using `#pragma once` in code generated by Meson. Since nobody ever complained, we can assume that's fine.

xclaesse:
Updated this PR for 1.3.0. AFAIK there is no strong opposition, just slight concern that it might not be as useful as it seems?

eli-schwartz:
> Updated this PR for 1.3.0. AFAIK there is no strong opposition, just slight concern that it might not be as useful as it seems?

The objection was more than slight concern, but it was also about the belief that this is badly suited to being a compiler attribute and better suited to being e.g. a module named something like "codesnippets".

mesonbuild/interpreter/compiler.py
eli-schwartz: I'm not certain whether anyone complained on the meson bugtracker.

Certainly, people complain on general C/C++ discussion sites that it isn't "portable" despite their refusal to use any compiler other than GCC, clang, or MSVC on account of the lot of the remainder being "half-baked garbage that doesn't implement standards from the last decade" and in some cases "is working towards c99 compliance but we're not quite there yet, please send patches".

Of course all 3 of those support pragma once.

Meanwhile, the standards committee doesn't want to standardize pragma once because they are concerned that the language specification shouldn't mandate that source code come inside files, because filesystems are mysterious beasts that are considered way too complicated for Serious Business™ like a programming language standard to depend on. What should the programming language describe as the correct way to handle headers that are a symlink? Or even worse, what about hardlinks??? No, no, this "filesystem" technology is too unpredictable.

xclaesse:
@eli-schwartz what other method would you see in that module? I never saw a concrete proposal, and making a module for a single method seems overkill. IMHO a compiler method is by far the most intuitive place, I really don't why that's a problem.

mesonbuild/interpreter/compiler.py
xclaesse: It's impossible to write C without hitting a million UB. At some point when all serious implementation does the same thing, who cares what the spec says...

eli-schwartz: I find it hard to out into words, how utterly upset I was to (recently) discover that the spec says if you don't include an end of line character at the end of your file (the posix definition of a text, rather than binary file, is that it has posix line terminators)...

... Then because C/C++ source code is *text files*, and a file that isn't a posix text file is technically not C/C++ source code, the spec declares this, too, is UB. And not something sane like I don't know, an input processing "error: an input file was specified that isn't a source code file". Or even just saying "an implementation is permitted to accept inputs other than source files as an implementation extension, as long as it successfully parses as valid C/C++".

But no, they outright declare that compilers have a full license to generate UB and make demons fly out of your nose instead of doing what your code's C/C++ mapping says.

Nasal demons for your line endings. It's obscene. But they don't want to standardize pragma once ***because filesystems***.

dnicolodi:
> making a module for a single method seems overkill

Well, technically, the `python` module exposes only one function... 😄

docs/markdown/snippets/symbol_visibility_header.md
```diff
@@ -0,0 +1,10 @@
+## New method to handle GNU and Windows symbol visibility
+
+Defining public API of a cross platform C/C++ library is often painful and
+require to copy macros snippets into every projects, typically using
+`__declspec(dllexport)`, `__declspec(dllimport)` or
+`__attribute__((visibility("default")))`.
+
+Meson can know generate a header file that defines exactly what's needed for
```
dnicolodi, line 8: `s/know/now/`

docs/yaml/objects/compiler.yaml
```diff
@@ -659,3 +659,105 @@ methods:
...
+        Name defined only when compiling or using static library.
+    static_only:
+      type: bool
+      default: false
+      description: |
+        If set to true, '<NAMESPACE>_STATIC_COMPILATION' is defined inside the
+        generated header.
+  description: |
+    Generate a header file that defines a macro to be used to mark all public APIs
+    of a library. Depending on the type of compiler and the platform, this will
```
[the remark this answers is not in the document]
xclaesse: > Typically this would be `__declspec(dllexport)`, `__declspec(dllimport)` for MSVC

I think that's a common misconception to think this is a MSVC thing. This is a **Windows** thing, even with GCC. At least as far as I understood.

https://gcc.gnu.org/onlinedocs/gcc/Microsoft-Windows-Function-Attributes.html
> On systems that support the visibility attribute, this attribute also implies “default” visibility.

It implies default visibility, but is not synonym to it, that's what I understand from that doc at least.

Removed `the type of compiler and` in my sentence, because it depends more on the platform than compiler.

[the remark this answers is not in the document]
dnicolodi: I indeed thought so, but your sentence indeed cause me to doubt my recollections. I do almost no development on Windows. However, if the header does not depend on the compiler, someone could ask why this is a compiler object method, and we are back to the discussion about where this facility belongs...

The new sentence is clearer, but still gives the impression that different headers are generated on different platforms. I think it is important to communicate that the generated header is static. That's what my reformulation was trying to achieve.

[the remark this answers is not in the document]
xclaesse: > think it is important to communicate that the generated header is static.

Added a sentence saying that explicitly.

> However, if the header does not depend on the compiler, someone could ask why this is a compiler object method, and we are back to the discussion about where this facility belongs...

It does not have to be a compiler method, but I personally think it's the most natural place. OTOH, it's not a hill I'm ready to die on, if @eli-schwartz really wants a module just for that, let's do it. Could be `meson.symbol_visibility_header()` too. But it's not clear he's going to accept it in that form, feels like he's going to reject it in any shape of form, but I could be wrong.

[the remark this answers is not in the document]
eli-schwartz: I said I'd be happy with it as a module, and I mean it. :)

I do think that a module makes some sense, because there's other forms of snippets we may very well want too. Another possible use case is to generate test code snippets for some of the fancier autotools macros that projects sometimes find themselves using in order to test which of several different API conventions are used for the same meta-concept. Stuff like PAM_STRERROR_TWO_ARGS.

Why do you think that only visibility headers are going to end up being widely used/needed? :)

(This is also why I think the `meson` builtin object isn't any better of a location than the compiler object.)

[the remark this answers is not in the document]
bugaevc: Just a drive-by note, potentially languages other than C may need visibility headers, which is why it makes sense to put it into a C compiler method.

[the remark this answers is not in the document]
xclaesse: > Why do you think that only visibility headers are going to end up being widely used/needed? :)

Because I haven't seen requests for anything else yet, but I might very well just have missed them.

> potentially languages other than C may need visibility headers

Any specific language in mind? Currently this PR covers C/C++ and ObjC/ObjC++ which all use the same header file.

xclaesse:
Updated this PR to add `snippets` module.

mesonbuild/modules/snippets.py
```diff
@@ -0,0 +1,98 @@
...
+
+    INFO = ModuleInfo('snippets', '1.3.0')
+
+    def __init__(self) -> None:
+        super().__init__()
+        self.methods.update({
+            'symbol_visibility_header': self.symbol_visibility_header_method,
+        })
+
+    @typed_kwargs('compiler.symbol_visibility_header',
```
tristan957, line 48: Looks like this didn't get renamed

test cases/common/262 symbol visibility header/subprojects/sub/meson.build
tristan957, line 1: Maybe the tests should move into a snippets module subfolder?

mesonbuild/modules/snippets.py
xclaesse: done

test cases/common/262 symbol visibility header/subprojects/sub/meson.build
xclaesse: good point, done

tristan957:
May want to change the PR title too!

ePirat:
> Should we add a license header to the generated header? Something like Unlicense or CC0? Or add a preamble/prefix kwarg of some sort?

IMO the best solution to this and the `#pragma` vs classic include guards issue is to add the possibility to allow to just get the string so projects and configure_file their own header with their own include guards, license and whatever else they might want to add.

eli-schwartz:
We don't add a license for generated config.h files because they're just a list of constants that have been format-shifted into C preprocessor statements. I can see the argument that this produced header is a bit more, well, creative work.

What's the purpose of the license in this case, then? Do we want to explicitly disavow any copyright claim against people using the produced header? Probably, that's the usual reason to go with a license that claims to be equivalent to a public domain dedication.

The frustrating thing is that there's a lot of disagreement about what kinds of public domain dedication "count". Various groups will insist that CC0 isn't free enough, or that unlicense isn't "clear", etc. For example, Google says its employees cannot contribute to projects under either of those two licenses as part of its internal policy guidance on:

> Forbidden patches
> 
> Our general philosophy is that we do not allow patches to projects that Google cannot use, either because of the terms of the license or Google policy.

xclaesse:
AFAIK if we want a license it should be Meson's Apache 2.0, because that's basically copying a string from Meson source into a file. I wouldn't mind adding `/* SPDX-License-Identifier: Apache-2.0 */` line on top.

But then raise the question of where did I found that code snippet... it is _inspired_ from what I saw in glib/gstreamer (LGPL2+), but not exactly (because I had to fix them in the process). It is also _inspired_ from GCC documentation https://gcc.gnu.org/wiki/Visibility (GPL?).

ePirat:
> AFAIK if we want a license it should be Meson's Apache 2.0, because that's basically copying a string from Meson source into a file. I wouldn't mind adding `/* SPDX-License-Identifier: Apache-2.0 */` line on top.

This will be unusable at least in VLC then, as it would force the rest of the code to (L)GPL 3 to be compatible and I am not even sure if that compatibility extends to source-level and not just linking…

xclaesse:
Let's just do nothing and everyone will be happy. As the author of that snippet I hereby claim that everyone can do what the fuck they want with it:  `SPDX-License-Identifier: WTFPL` :D

xclaesse:
More seriously, I think it's fine for GPL3: https://en.wikipedia.org/wiki/Apache_License#Compatibility

eli-schwartz:
> Let's just do nothing and everyone will be happy. As the author of that snippet I hereby claim that everyone can do what the fuck they want with it: `SPDX-License-Identifier: WTFPL` :D

That too is on the Google "do not use" list. :P My understanding is that Google is mad that it isn't actually Public Domain, so the code is still provided under a "license", but the license says you can "*do* whatever you want" but doesn't say you can "*have* a warranty disclaimer". :facepalm:

Their documentation on it appears to have vanished but [the Internet Archive](https://web.archive.org/web/20181022125422/https://opensource.google.com/docs/thirdparty/licenses/#wtfpl-not-allowed) does not forget:
> Code released under the [WTFPL](http://www.wtfpl.net/) cannot be used at Google. This license has a large number of issues (lack of warranty disclaimer, very vague rights grant), and was also rejected as an open source license by OSI. We also do not allow contribution to projects under the WTFPL.

Perhaps its disappearance is a sign that it became acceptable to use, but I doubt it since it is still listed as one of the licenses that Googlers aren't allowed to contribute to: https://opensource.google/documentation/reference/patching#forbidden

BSD0 is however on the approved list as an explicit exception to "we at Google do not like Public Domain dedications", and is commonly considered to be "not actually a Public Domain dedication but effectively equivalent".

eli-schwartz:
> Let's just do nothing and everyone will be happy.

And in particular this: I am not overly worried about whether or not people can freely use a function to generate code snippets, but "do nothing and everyone will be happy" is an *incredibly* humorous thought, because this is a topic that people make a very good livelihood off of convincing the entire tech world to be deeply unhappy about it.

xclaesse:
Anyway, let me know what you want, I couldn't care less. I'm not a lawyer.

tristan957:
> IMO the best solution to this and the `#pragma` vs classic include guards issue is to add the possibility to allow to just get the string so projects can configure_file their own header with their own include guards, license and whatever else they might want to add.

I really like this idea. @xclaesse what if we made this more similar to vcs_tag()?

xclaesse:
Returning the string to be able to add it into an existing `configure_file()` would make sense, that's the kind of things many project already have in their `config.h`. IMHO that would require to add at least `fs.write()` method to be able to write that string verbatim into a new file, and also have proper solution for https://github.com/mesonbuild/meson/issues/6259.

I don't think that solves the license question, the problem is exactly the same, are you allowed to paste that code into a file with your project license. But unless one of us is secretly a lawyer specialized on that matter, I think we are just talking without knowing. If a big company wants to pay a few millions their legal team to debate on this we'll probably have a answer in a decade or two.

tristan957:
What I am thinking is more along the lines of:

```c
// header.h.in

// preable...

@SYMBOL_VISIBILITY_MACROS@

// function decls, etc.
```

```meson
header = snippets.symbole_visibility_macros(input: 'header.h.in', output: 'header.h')
```

xclaesse:
That makes API more complicated to use for no real benefit IMHO. Hooking it with configure_file() does have the benefit to make the snippet part of an already existing file in the project instead of adding an extra one.

tristan957:
The benefit I see is that in the current approach all projects now have to install a `visibility.h` into the includedir. While the proposed approach allows them to embed in any file they want.

xclaesse:
Sure, but then doing it through `configure_file()` seems much more flexible because you can substitute other keys in a single step.

tristan957:
> Sure, but then doing it through `configure_file()` seems much more flexible because you can substitute other keys in a single step.

Right. This is similar the vcs_tag() idea I talked about on Matrix

docs/markdown/Snippets-module.md
```diff
@@ -0,0 +1,118 @@
...
+
+- `namespace`: Prefix for generated macros, defaults to the current project name.
+  It will be converted to upper case with all non-alphanumeric characters replaced
+  by an underscore `_`. It is only used for `api`, `compilation` and
+  `static_compilation` default values.
+- `api`: Name of the macro used to mark public APIs. Defaults to `<NAMESPACE>_API`.
+- `compilation`: Name of a macro defined only when compiling the library.
+   Defaults to `<NAMESPACE>_COMPILATION`.
+- `static_compilation`: Name of a macro defined only when compiling or using
+  static library. Defaults to `<NAMESPACE>_STATIC_COMPILATION`.
```
bonzini, line 43: Are these needed? I'd prefer less flexibility but a very simple to use API.

docs/markdown/Snippets-module.md
```diff
@@ -0,0 +1,118 @@
...
+- `static_only`: If set to true, `<NAMESPACE>_STATIC_COMPILATION` is defined
+  inside the generated header. In that case the header can only be used for
+  building a static library and thus cannot be used with `both_libraries()` or
+  `library()` when `default_library=both`. If set to `false` (the default) and
+  the generated header is used in a static library,
```
bonzini: Would it make sense to instead pass a string that is `'static', 'shared', 'both'`, and default to `get_option('default_library')`? Changing the name of the argument of course.

xclaesse:
@tristan957 I updated this PR, and added `symbol_visibility_string()` that returns the string that can be used with `configure_file()` instead of writing a file.

mesonbuild/modules/snippets.py
```diff
@@ -0,0 +1,112 @@
...
+                  KwargInfo('compilation', (str, NoneType)),
+                  KwargInfo('static_compilation', (str, NoneType)),
+                  KwargInfo('static_only', bool, default=False))
+    @typed_pos_args('snippets.symbol_visibility_header', str)
+    def symbol_visibility_header_method(self, state: ModuleState, args: T.Tuple[str], kwargs: 'SymbolVisibilityHeaderKW') -> mesonlib.File:
+        header_name = args[0]
+        content = '#pragma once\n'
+        content += self.symbol_visibility_string(state, kwargs)
+        header_path = Path(state.environment.get_build_dir(), state.subdir, header_name)
+        header_path.write_text(content, encoding='utf-8')
```
bonzini, line 97: Would it make sense to have a `fs.write()` method instead? Otherwise any future similar methods would almost definitely come in pairs.

mesonbuild/modules/snippets.py
```diff
@@ -0,0 +1,112 @@
...
+        return content
+
+    @typed_kwargs('snippets.symbol_visibility_string',
+                  KwargInfo('namespace', (str, NoneType)),
+                  KwargInfo('api', (str, NoneType)),
+                  KwargInfo('compilation', (str, NoneType)),
+                  KwargInfo('static_compilation', (str, NoneType)),
+                  KwargInfo('static_only', bool, default=False))
+    @noPosargs
+    def symbol_visibility_string_method(self, state: ModuleState, args: T.Tuple[str], kwargs: 'SymbolVisibilityHeaderKW') -> str:
```
bonzini: Maybe add `c_` here so that it's clear what language is generated?

If my proposal below to add `fs.write()` is accepted, I'd also remove `_string`, and call this for example `c_symbol_visibility_macros`.

docs/markdown/Snippets-module.md
xclaesse: It all has default value, the flexibility is intended mostly for wrapdb ports where you don't control how the project names those macros.

xclaesse: Typically cmake projects using [GenerateExportHeader](https://cmake.org/cmake/help/latest/module/GenerateExportHeader.html).

mesonbuild/modules/snippets.py
xclaesse: Hm, I'm not sure. I like generating the file directly because it's a simple one-liner and matches what cmake does. Also the string variant you need to prepend `#pragma once`.

mesonbuild/modules/snippets.py
xclaesse: It is c, c++, objc and objc++. Who knows, maybe a language kward could be added for other langues, dunno if there are any candidates.

mesonbuild/modules/snippets.py
bonzini: Ah I missed the `#pragma once` but it is usually not necessary. Compilers should not warn if you use the exact same definition, and if the definition changes it'd be a bug and `#pragma once` hides it.

This doesn't seem too bad:

```
file = fs.write('apiconfig.h', snippets.whatever_name_is_chosen(
    ...
)
```

docs/markdown/Snippets-module.md
xclaesse: When I wrote that Meson couldn't build both shared and static for Windows, but now it can thanks to `c_static_args`. There is less of a use-case for `static_only` now. I think a boolean value is cleaner than literal strings.

bonzini: No problem but please make the default based on `default_library`.

mesonbuild/modules/snippets.py
bonzini: I am just worried that this becomes a dumping ground with no rhythm and rhyme. Sure `c` is not hundred percent accurate but it's better than nothing. I will let others judge the color that they think is best for the bike shed.

docs/markdown/Snippets-module.md
```diff
@@ -0,0 +1,118 @@
...
+  static_compilation: str
+  static_only: bool
+)
+```
+
+Generate a header file that defines macros to be used to mark all public APIs
+of a library. Depending on the platform, this will typically use
+`__declspec(dllexport)`, `__declspec(dllimport)` or
+`__attribute__((visibility("default")))`. It is compatible with C, C++,
+OBJC and OBJC++ languages. The content of the header is static regardless
```
bugaevc: Why is this uppercase?

docs/markdown/Snippets-module.md
```diff
@@ -0,0 +1,118 @@
...
+```
+
+`mylib/meson.build`:
+```meson
+default_library = get_option('default_library')
+snippets = import('snippets')
+apiheader = snippets.symbol_visibility_header('apiconfig.h',
+  static_only: default_library == 'static')
+install_headers(apiheader, 'lib.h', subdir: 'mylib')
+lib = library('mylib', 'lib.c',
```
bugaevc, line 65: Does this not need to mention `apiheader` in the list of sources?

mesonbuild/modules/snippets.py
```diff
@@ -0,0 +1,112 @@
...
+                #  define {static_compilation}
+                #endif /* {static_compilation} */
+                ''')
+        content += textwrap.dedent(f'''
+            #if (defined(_WIN32) || defined(__CYGWIN__)) && !defined({static_compilation})
+            #  define {api}_EXPORT __declspec(dllexport)
+            #  define {api}_IMPORT __declspec(dllimport)
+            #elif __GNUC__ >= 4
+            #  define {api}_EXPORT __attribute__((visibility("default")))
+            #  define {api}_IMPORT
```
bugaevc, line 84: Doesn't this also need default visibility? Suppose we're including this library's headers into another library, which is also being built with `-fvisibility=hidden`. The other library should still reference these symbols as external ones.

mesonbuild/modules/snippets.py
```diff
@@ -0,0 +1,112 @@
...
+                #ifndef {static_compilation}
+                #  define {static_compilation}
+                #endif /* {static_compilation} */
+                ''')
+        content += textwrap.dedent(f'''
+            #if (defined(_WIN32) || defined(__CYGWIN__)) && !defined({static_compilation})
+            #  define {api}_EXPORT __declspec(dllexport)
+            #  define {api}_IMPORT __declspec(dllimport)
+            #elif __GNUC__ >= 4
+            #  define {api}_EXPORT __attribute__((visibility("default")))
```
bugaevc, line 83: Consider making this `protected` rather if the toolchain supports it.

(See https://maskray.me/blog/2021-01-09-copy-relocations-canonical-plt-entries-and-protected for discussion.)

mesonbuild/modules/snippets.py
dcbaker: I would prefer the module to have a name related to the C language family. We have both `rust` and `dlang` modules already, so there's good precedent for that kind of name, a `c` or `clike` (or whatever) module would be a good conceptual grouping.

mesonbuild/modules/snippets.py
eli-schwartz: Only Microsoft requires you to specify when linking against a library that the external library's symbols are importable. The purpose of "import" attributes is so that when utilizing a header with the import attribute, the compiler produces a function call to a different (mangled) object

Visibility attributes for unix happen solely when compiling a symbol implementation, not when compiling a function call to that symbol. There is no utility in having visibility attributes in the installable headers, but it causes no harm and is necessary for single-sourcing (that is, avoiding having different header files used internally for compiling the application versus linking to it).

True single-sourcing would be if Microsoft behaved like Unix, but unfortunately it doesn't and therefore Microsoft compilers depend on both the contents of the header as well as CFLAGS, to determine whether the you are building the library or linking against it.

docs/markdown/Snippets-module.md
ePirat: Should be Objective-C and Objective-C++

mesonbuild/modules/snippets.py
eli-schwartz: I think that adding a generic `fs.write()` method probably needs @jpakkane's sign-off either way. I seem to recall it was intentionally excluded, originally.

mesonbuild/modules/snippets.py
xclaesse: Initially I wanted this to be a Compiler function, `cc.symbol_visibility_header()` but it was specifically required to move that to a `snippets` module. I still think it's better on the Compiler object, but the idea was to have more code generator helpers in this module. TBH, I think it's fine as it is.

eli-schwartz:
> Anyway, let me know what you want, I couldn't care less. I'm not a lawyer.

I offer you an interesting novelty solution: generate the code snippet using a dual license:

```
// SPDX-license-identifier: 0BSD OR CC0-1.0 OR WTFPL OR Apache-2.0 OR LGPL-2.0-or-later
```

docs/markdown/Snippets-module.md
xclaesse: It's generated at configure time, so it's not needed. You add generated headers in sources when it's generated at build-time, so meson needs to know the build order.

mesonbuild/modules/snippets.py
xclaesse: TBH, I don't think we need a string version. You can just `#include` the generated header into your header. And with @eli-schwartz insisting on putting a licence on the snippet, how will that work if you write it in another file...

mesonbuild/modules/snippets.py
xclaesse: Hm, I've never seen it used anywhere, I'm not sure we can do it by default. Maybe it can be a boolean kwarg to use it.

mesonbuild/modules/snippets.py
eli-schwartz: Point of order: I didn't insist on adding a license, @tristan957 is the one that brought that up.

I believe if we are going to add a license (and I can see the arguments for doing so, and I can see the arguments for not doing so), then it should be done correctly, because "adding a license header that is incompatible with some projects" is worse than "omitting a license header and leaving it up to projects to decide whether they think the result is copyrightable".

I do suspect it is better to add one than to omit one, but that's a long way away from insisting.

mesonbuild/modules/snippets.py
eli-schwartz: The GCC manpage claims it is entirely useless, and the linked blog post says doing so is fraught with risk.

I can't imagine why we would want to consider this suggestion, honestly.

mesonbuild/modules/snippets.py
bugaevc: I really don't understand how you suppose a compiler would access a global variable (or a pointer to a function, for purposes other than calling the function, because calling is done through a PLT proxy, which is essentially a local function, hence there's no need to differentiate) inside a compilation unit that is being built with hidden visibility.

On e.g. x86_64, loading a locally defined global variable looks like
```
movl the_global(%rip), %eax
```
vs loading an externally defined global:
```
movq the_global@GOTPCREL(%rip), %rax
movl (%rax), %eax
```

And unless the compiler sees `__attribute__((visibility("default")))`, it's going to generate the former (assuming `-fvisibility=hidden`).

docs/markdown/Snippets-module.md
xclaesse: Made it true by default when default_library=static. Refactored a bit the documentation for that.

mesonbuild/modules/snippets.py
xclaesse: Reverted to just be a file generator. It's simpler and I'm not convinced we really need a string version. Can always be added later.

docs/markdown/Snippets-module.md
xclaesse: fixed

mesonbuild/modules/snippets.py
xclaesse: Let's not block on that, it can be done later.

xclaesse: Let's not block on that, it can be done later.

xclaesse:
> > Anyway, let me know what you want, I couldn't care less. I'm not a lawyer.
> 
> I offer you an interesting novelty solution: generate the code snippet using a dual license:
> 
> ```
> // SPDX-license-identifier: 0BSD OR CC0-1.0 OR WTFPL OR Apache-2.0 OR LGPL-2.0-or-later
> ```

Added that in the generated file.

mesonbuild/modules/snippets.py
thesamesam: This matters for protected visibility. Without protected visibility being involved, the only way I see a difference with GCC is `-mno-direct-extern-access` vs `-mdirect-extern-access`.

Without protected visibility, you have the following cases:
* Symbol is exported in a shared library via `default` => must be accessed via the GOT
* Symbol is not exported in a shared library (so `hidden`) => can't be accessed at all

With protected visibility, you have:
* The above and..
* Symbol is exported in a shared library via `protected` and access has `protected` attribute visible => GOT or copy relocation
* Symbol is exported in a shared library via `protected` and access is undecorated => failure to link

For static libraries, visibility makes no difference, of course.

Am I missing another case? If so, could you show me testcase so I can play with it? As a disclaimer, I'm a bit under-the-weather and should not be working right now but wanted to chime in now because I agree this should be handled.

bugaevc: Basically, this: https://godbolt.org/z/jrr5sqd7E

But you're right that `-fvisibility=hidden` alone doesn't seem enough to trigger it; I thought it did.

> * Symbol is exported in a shared library via `protected` and access has `protected` attribute visible => GOT or copy relocation
> * Symbol is exported in a shared library via `protected` and access is undecorated => failure to link

Rather: if a symbol is exported as `protected` and accessed as if `default` => GOT; anything else and you get a failure. So basically:
* `hidden` means defined in this DSO, not exported
* `protected` means defined in this DSO, exported
* `public` means defined in another (or this same) DSO, and exported from there

`hidden` and `protected` are accessed PC-relatively without GOT, `public` is accessed through GOT. `protected` and `public` end up generating entries in the dynamic symbol table.

Historically people have not been using `protected`, and also used copy relocations, which are incompatible with `protected`. Now there is motion to stop using copy relocations, which enables libraries to use `protected` for their symbols.

> For static libraries, visibility makes no difference, of course.

Well, static libraries get linked into some DSOs eventually, and are otherwise subject to the same rules.

docs/markdown/snippets/symbol_visibility_header.md
```diff
@@ -0,0 +1,10 @@
+## New method to handle GNU and Windows symbol visibility
+
+Defining public API of a cross platform C/C++ library is often painful and
+require to copy macros snippets into every projects, typically using
```
tristan957:
```suggestion
requires copying macro snippets into every projects, typically using
```

docs/markdown/snippets/symbol_visibility_header.md
```diff
@@ -0,0 +1,10 @@
+## New method to handle GNU and Windows symbol visibility
```
tristan957: Maybe we should add "for C/C++/ObjC/ObjC++" in the title?

tristan957 (commented):
This looks awesome. Thanks for sticking with it!

docs/markdown/snippets/symbol_visibility_header.md
xclaesse: done

docs/markdown/snippets/symbol_visibility_header.md
xclaesse: done

PR #10202, opened 2022-03-28
mdist: allow passing options to setup, allow specifying options to setup in meson.build

### LINKED ISSUE #10181: Python module broken on 0.62.0 for meson dist
Running `meson dist` on 0.62.0 produces this error:
```
meson-private/dist-unpack/test-1/meson.build:4:0: ERROR: python.platlibdir and python.install_env are mutually exclusive
```
Required `meson.build` file:
```
project('test', 'c', version: '1')
pymod = import('python')
python = pymod.find_installation('python3', required: true)
```
And then just:
```
$ meson setup builddir/
$ meson dist -C builddir/
Testing distribution package /home/whot/tmp/2022-03-25-Fri/test/build/meson-dist/test-1.tar.xz
The Meson build system
Version: 0.62.0
Source dir: /home/whot/tmp/2022-03-25-Fri/test/build/meson-private/dist-unpack/test-1
Build dir: /home/whot/tmp/2022-03-25-Fri/test/build/meson-private/dist-build
Build type: native build
Project name: test
Project version: 1
C compiler for the host machine: ccache cc (gcc 12.0.1 "cc (GCC) 12.0.1 20220308 (Red Hat 12.0.1-0)")
C linker for the host machine: cc ld.bfd 2.37-24
Host machine cpu: x86_64

build/meson-private/dist-unpack/test-1/meson.build:4:0: ERROR: python.platlibdir and python.install_env are mutually exclusive

A full log can be found at /home/whot/tmp/2022-03-25-Fri/test/build/meson-private/dist-build/meson-logs/meson-log.txt
Running Meson on distribution package failed
Dist check build directory was /home/whot/tmp/2022-03-25-Fri/test/build/meson-private/dist-build
```
There isn't really anything in the suggested meson-log.txt file, it's just the standard checks up to `Target machine cpu: x86_64` and then the error message. Reproduced on a Ubuntu 20.04 podman container (Python 3.8.10) and an up-to-date Fedora 36 (Python 3.10.3).

Bisected to commit 78945fb9832e989453fbabc471d45bb71805eca8, cc @eli-schwartz.

eli-schwartz:
Also reduce the number of arguments manually specified when reading currently active config.

Fixes #10181

### DIFF
--- a/docs/markdown/Creating-releases.md
+++ b/docs/markdown/Creating-releases.md
@@ -60,6 +60,13 @@ for example when done in CI that already does its own testing.
 So with `--no-tests` you can tell Meson "Do not build and test generated
 packages.".
 
+## Run the build and test with custom configurations
+
+The `meson dist` comand accepts arbitrary arguments after `--`, which will be
+forwarded to `meson setup` when invoking the build and test step of the
+generated packages. It can be used to perform a distcheck for specific
+configurations.
+
 ## Use `--allow-dirty` to override error when git repository contains uncommitted changes
 
 *Since 0.62.0* Instead of emitting a warning when a repository contains
--- a/docs/markdown/snippets/add_dist_args.md
+++ b/docs/markdown/snippets/add_dist_args.md
@@ -1,0 +1,6 @@
+## It is now possible to customize the setup arguments to ninja dist
+
+A new function has been added: [[meson.add_dist_args]]. This allows specifying
+arguments which will be used in the generated `ninja dist` target to customize
+how the distcheck stage will configure the build. By default a distcheck simply
+uses the same configuration as the main build.
--- a/docs/markdown/snippets/dist-accepts-setup-args.md
+++ b/docs/markdown/snippets/dist-accepts-setup-args.md
@@ -1,0 +1,6 @@
+## dist now accepts arguments to be passed to setup
+
+When performing a distcheck, it is sometimes desirable to check various
+configurations. Previously, Meson ran the distcheck using the same
+configuration as the current build. Now it is possible to override this in
+`meson dist` without first reconfiguring the main build.
--- a/docs/yaml/builtins/meson.yaml
+++ b/docs/yaml/builtins/meson.yaml
@@ -5,6 +5,18 @@ description: |
   system. This object is always mapped in the `meson` variable.
 
 methods:
+  - name: add_dist_args
+    returns: void
+    since: 0.63.0
+    description: |
+      Passes additional arguments to `meson dist` to be used when building and
+      testing the generated tarball.
+
+    varargs:
+      name: arg
+      type: str
+      description: the args to pass
+
   - name: add_dist_script
     returns: void
     since: 0.48.0
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -3208,9 +3208,13 @@ https://gcc.gnu.org/bugzilla/show_bug.cgi?id=47485'''))
         return sorted(cmds)
 
     def generate_dist(self):
+        if self.build.dist_args_exclusive:
+            dist_args = self.build.dist_args_exclusive
+        else:
+            dist_args = ['--reuse-setup-args'] + self.build.dist_args
         elem = NinjaBuildElement(self.all_outputs, 'meson-dist', 'CUSTOM_COMMAND', 'PHONY')
         elem.add_item('DESC', 'Creating source packages')
-        elem.add_item('COMMAND', self.environment.get_build_command() + ['dist'])
+        elem.add_item('COMMAND', self.environment.get_build_command() + ['dist'] + dist_args)
         elem.add_item('pool', 'console')
         self.add_build(elem)
         # Alias that runs the target defined above
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -236,6 +236,8 @@ class Build:
         self.subproject_dir = ''
         self.install_scripts: T.List['ExecutableSerialisation'] = []
         self.postconf_scripts: T.List['ExecutableSerialisation'] = []
+        self.dist_args: T.List[str] = []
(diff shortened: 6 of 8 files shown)


eli-schwartz:
First commit is a bugfix for stable.

xclaesse:
I'm not convinced add_dist_args is needed, why would you invoke meson through Ninja? IMHO those Ninja targets are mostly legacy.

eli-schwartz:
The ninja targets aren't legacy, they exist to implement and alias good default behavior. In this case, upstream's preferred default dist behavior.

xclaesse:
> In this case, upstream's preferred default dist behavior

Hmm, makes sense to define default dist options indeed. But I always found it weird to invoke meson via ninja for things that are purely a meson job, I really think those rules are legacy for when we didn't have proper meson command for install/dist/test/etc. 

`meson dist` already pickle the build data file, so we don't need to pass those options via ninja at all, we could also make bare "meson dist" use them too which would be more consistent. I definitely do not want to force invoking dist via ninja for this feature.

An alternative is to let user override the `dist` target with their own run_target('dist', command: ['meson', 'dist', '--', '-Dfoo']). That way we don't need `add_dist_args` method at all. But then we are back at forcing using ninja which I would like to avoid.

xclaesse:
Wait, what about adding instead: `project(..., dist_options: [])` that way it's nicely together with `default_options` instead of randomly hidden anywhere in your build definition?

Btw, you did not document `overwrite` kwarg and I'm not sure to nderstand what it does.

eli-schwartz:
I am quite happy to invoke meson via ninja. In fact, the alternative is to invoke ninja via meson. So I don't know what the problem is supposed to be.

`meson dist` is for when you want fine-grained control over the options. This PR is already frustratingly ugly because I have to hack around the fact that its *historic* behavior is to recant the original setup arguments, which I could see as reasonable for the ninja wrapper but is wrong for running it directly and requires even more flags to disable it.

eli-schwartz:
> Wait, what about adding instead: `project(..., dist_options: [])` that way it's nicely together with `default_options` instead of randomly hidden anywhere in your build definition?

Alternatively, now you cannot adapt it to different needs. This is exactly why despite having languages in the project() varargs, there's also a "randomly hidden anywhere" function to add more.

The point of having default_options in the project kwargs is ***not*** to "have it nicely instead of randomly hidden anywhere", it is to read it immediately because project options must take effect before any other statements are run. It's never reasonable to define the default options in an if statement that can e.g. depend on the existing value of that option.

> Btw, you did not document `overwrite` kwarg and I'm not sure to nderstand what it does.

Oops.

xclaesse:
@eli-schwartz IMHO either we add `add_dist_args()` API and it must also work when using `meson dist` instead of `ninja dist`, or you really want it to be a ninja specific thing, and then we don't need `add_dist_args()` but rather override `dist` target with a run_target(). I personally prefer the 1st option.

eli-schwartz:
> IMHO either we add `add_dist_args()` API and it must also work when using `meson dist` instead of `ninja dist`,

I have already explained why I believe this would be hostile and surprising behavior. `meson dist` already has surprising and hostile behavior, but I initially opted to leave that in place for backwards compatibility.

> rather override `dist` target with a run_target()

I'm not even sure how that's conceptually supposed to work, you can't define targets using the meson command line because you don't know what `meson` is. It might be meson.py, it might be `python -m mesonbuild.mesonmain`, it might not be on PATH...

xclaesse:
> I'm not even sure how that's conceptually supposed to work, you can't define targets using the meson command line because you don't know what `meson` is. It might be meson.py, it might be `python -m mesonbuild.mesonmain`, it might not be on PATH...

It has been requested many times in the past already to special case `find_program('meson')` and/or `command: ['meson', ...]` to use currently running meson. Meson already knows how to invoke itself, we do that everywhere.

> `meson dist` already has surprising and hostile behavior, but I initially opted to leave that in place for backwards compatibility.

And not respecting `add_dist_args()` would make that even worse. Using `add_dist_args()` can be seen as an opt-in for a new behavior, so that could on the contrary be the occasion to fix it being "hostile".

xclaesse:
To be clear, adding API that make `ninja dist` and `meson dist` less consistent in a way that it's not even possible to get the same behavior with `meson dist` is a strong NO from me. Now, others could always overrule my judgment.

xclaesse:
Note that I agree with you that reusing setup args has always seemed wrong to me too.

My suggestion:
- In mdist.py if `b.dist_args` is not empty it implies `--no-reuse-setup-args`.
- Append `b.dist_args` to `extra_meson_args`.
- Remove `--reuse-setup-args`, it's bad enough to reuse setup args by default, I don't see why you want to use both setup args AND extra args.

jpakkane:
Making it possible to pass extra arguments with `meson dist` is good. I'm not too fond of the API addition, though.

Typically we have tried to be minimal. That is, add the `meson dist` bit first as it is better isolated and then only add the other bits as needed. I suspect that people have a command that runs `meson dist <whatever>` in their CI build pipeline anyway so they might well be happy with just the command addition.

jpakkane:
The problem here is that this MR both fixes an issue and adds new functionality. For point releases we only do bugfixes, no new features. Would it be possible to split out the bugfix so we can merge that now and get the point release out?

jpakkane:
I created a regression test for the actual issue which is in this commit: bfca4c90de5fc98e25fefdd08f8cf3a624e118c9

eli-schwartz:
Manually merged that one commit to master as 1420d0daceb10cafb52a7405f157032a5cc811a5, to avoid wasting CI time

jpakkane:
That won't get into the point release then, because the script only takes commits from merges that have the milestone tag applied.

eli-schwartz:
It had already gotten manually applied to the backport branch 13 days ago -- before we realized that the new features would necessitate more discussion and possibly rethinking, it was supposed to be all one PR anyway, so that would require manual work to prevent backporting too much.

PR #10203, opened 2022-03-29
fix regression in propagating depends in gtkdoc

eli-schwartz:
In commit 68e684d51f1e469e0d9f4b499ffda15146cad98a the function signature was changed, but several places did not adapt. Additionally, we now totally dropped the in-place update of gtkdoc's sole source of dependencies, but didn't propagate them upward to assign the newly collected dependencies anywhere.

Fixes building gtkdoc with internal dependencies and failing when specified directly (when building the 'all' target with sufficiently random parallelism, deps may be built on time).

Fixes:
  https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1008382
  https://gitlab.gnome.org/GNOME/libmediaart/-/issues/4

### DIFF
--- a/mesonbuild/modules/gnome.py
+++ b/mesonbuild/modules/gnome.py
@@ -674,6 +674,7 @@ class GnomeModule(ExtensionModule):
                         external_ldflags.update(libdepflags[2])
                         external_ldflags_nodedup += libdepflags[3]
                         gi_includes.update(libdepflags[4])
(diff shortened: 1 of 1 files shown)


dcbaker (approved):

eli-schwartz:
One day. One day it will *finally* be possible to edit the gnome module without regressing. :pray:

PR #10206, opened 2022-03-29
interpreter: add a new function add_project_dependencies()

bonzini:
This function can be used to add fundamental dependencies such as `libm`, `dependency('threads')` or glib to all build products in one fell swoop.  This can be useful whenever, due to a project's coding conventions, it is not really possible to compile any source file without including the dependency.

This function also makes it possible to add an `include_directories()` object to the project, removing the need to list both source and global directories. An example from QEMU is this:

```
  add_project_arguments('-isystem', meson.current_source_dir() / 'linux-headers',
                        '-isystem', 'linux-headers',
                        language: ['c', 'cpp'])
```

which might become:

```
   linux_headers = declare_dependency(include_directories: 'linux-headers').as_system('system')
   add_project_dependencies(linux_headers, language: ['c', 'cpp'])
```

### DIFF
--- a/docs/markdown/snippets/add_project_dependencies.md
+++ b/docs/markdown/snippets/add_project_dependencies.md
@@ -1,0 +1,11 @@
+## `add_project_dependencies()` function
+
+Dependencies can now be added to all build products using
+`add_project_dependencies()`.  This can be useful in several
+cases:
+
+* with special dependencies such as `dependency('threads')`
+* with system libraries such as `find_library('m')`
+* with the `include_directories` keyword argument of
+`declare_dependency()`, to add both source and build
+directories to the include search path
--- a/mesonbuild/dependencies/base.py
+++ b/mesonbuild/dependencies/base.py
@@ -23,7 +23,7 @@ import itertools
 import typing as T
 from enum import Enum
 
-from .. import mlog
+from .. import mlog, mesonlib
 from ..compilers import clib_langs
 from ..mesonlib import LibType, MachineChoice, MesonException, HoldableObject
 from ..mesonlib import version_compare_many
@@ -459,6 +459,22 @@ class ExternalLibrary(ExternalDependency):
         if not link_args:
             new.link_args = []
         return new
+
+
+def get_leaf_external_dependencies(deps: T.List[Dependency]) -> T.List[Dependency]:
+    if not deps:
+        # Ensure that we always return a new instance
+        return deps.copy()
+    final_deps = []
+    while deps:
+        next_deps = []
+        for d in mesonlib.listify(deps):
+            if not isinstance(d, Dependency) or d.is_built():
+                raise DependencyException('Dependencies must be external dependencies')
+            final_deps.append(d)
+            next_deps.extend(d.ext_deps)
+        deps = next_deps
+    return final_deps
 
 
 def sort_libpaths(libpaths: T.List[str], refpaths: T.List[str]) -> T.List[str]:
--- a/mesonbuild/dependencies/base.py
+++ b/mesonbuild/dependencies/base.py
@@ -165,6 +165,9 @@ class Dependency(HoldableObject):
         else:
             return 'unknown'
 
+    def get_include_dirs(self) -> T.List['IncludeDirs']:
+        return []
+
     def get_include_type(self) -> str:
         return self.include_type
 
@@ -297,6 +300,9 @@ class InternalDependency(Dependency):
             self.version, final_includes, final_compile_args,
             final_link_args, final_libraries, final_whole_libraries,
             final_sources, final_deps, self.variables, [], [])
+
+    def get_include_dirs(self) -> T.List['IncludeDirs']:
+        return self.include_directories
 
     def get_variable(self, *, cmake: T.Optional[str] = None, pkgconfig: T.Optional[str] = None,
                      configtool: T.Optional[str] = None, internal: T.Optional[str] = None,
--- a/test cases/common/251 add_project_dependencies/meson.build
+++ b/test cases/common/251 add_project_dependencies/meson.build
@@ -1,0 +1,22 @@
+project('zlib system dependency', 'c')
+
+cc = meson.get_compiler('c')
+
+m = cc.find_library('m', required: false)
+add_project_dependencies(m, language: ['c'])
+
+z = dependency('zlib', method: 'system', required: false)
+if not z.found()
+  error('MESON_SKIP_TEST zlib not present')
+endif
+
+z_c_args = z.partial_dependency(compile_args: true, includes: true)
+add_project_dependencies(z_c_args, language: 'c', native: false)
+
+global_dep = declare_dependency(include_directories: include_directories('inc'),
+                                compile_args: '-DDEFINED')
+add_project_dependencies(global_dep, language: 'c', native: false)
+
+lib = static_library('rary', 'lib.c')
+exe = executable('prog', 'main.c', link_with: lib, dependencies: z)
+test('test', exe)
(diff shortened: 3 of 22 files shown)


eli-schwartz:
This would also spare @rgommers the need to check if `find_library('m')` is found and then add `-lm` as project link args.

Seems reasonable enough, anyway.

eli-schwartz:
add_project_arguments and suchlike cannot be used after build targets have been defined, IIRC.

bonzini:
> add_project_arguments and suchlike cannot be used after build targets have been defined, IIRC.

I think you could use `override_dependency` on a superproject and `add_project_dependency()` on a subproject, so I preferred to reuse the check on `is_built()`.

eli-schwartz:
No, I mean this:

https://github.com/mesonbuild/meson/blob/80192aa577578a094c7da2c3696edbfbd38052cf/mesonbuild/interpreter/interpreter.py#L2670-L2682

rgommers:
> This would also spare @rgommers the need to check if `find_library('m')` is found and then add `-lm` as project link args.

For context, that code can be found [here](https://github.com/scipy/scipy/blob/main/meson.build#L32) and looks like:
```meson
# We need -lm for all C code (assuming it uses math functions, which is safe to
# assume for SciPy). For C++ it isn't needed, because libstdc++/libc++ is
# guaranteed to depend on it. For Fortran code, Meson already adds `-lm`.
m_dep = cc.find_library('m', required : false)
if m_dep.found()
  add_project_link_arguments('-lm', language : 'c')
endif
```

That is't too bad in the end, but figuring out the details and how every language treats `mlib` differently was a little nontrivial. So I guess it's the code comment rather than the `add_project_link_arguments` above that took the work here; if that logic could be built in I think that would make sense.

mesonbuild/dependencies/base.py
```diff
@@ -467,6 +473,22 @@ def get_partial_dependency(self, *, compile_args: bool = False,
...
+    final_deps = []
+    while deps:
+        next_deps = []
+        for d in mesonlib.listify(deps):
+            if not isinstance(d, Dependency) or d.is_built():
+                raise DependencyException('Dependencies must be external dependencies')
+            final_deps.append(d)
+            next_deps.extend(d.ext_deps)
+        deps = next_deps
+    return final_deps
```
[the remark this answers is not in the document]
bonzini: That looks nice, however note that all of the uses I could find start from a _list_ of dependencies. So you would need something like `itertools.chain.from_iterable((iter_deps(d) for d in deps))` which is not pretty either.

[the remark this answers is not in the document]
bonzini: Another update, I jotted something that takes a list of dependencies and returns a generator of dependencies, but it cannot really be used as much as I hoped:
* it can be used in BuildTarget.add_deps
* it cannot be used in HotdocTargetBuilder.process_dependencies because it accepts pretty much anything and not just dependencies
* it cannot be used very well in GnomeModule._get_dependencies_flags because it recurses on build.SharedLibrary's `lib.get_external_deps()` in addition to dependencies. I made `iter_deps()` could optionally iterate on the external dependencies of an InternalDependency's `dep.libraries` (but I don't understand why not do so with static libraries).
* the pkgconfig module's DependencyHelper._process_libs is also a complicated beast doing mutual recursion with DependencyHelper's other methods _add_lib_dependencies, add_pub_libs and add_priv_libs. I didn't touch it.

All in all, adding iter_deps() ended up adding 6 lines of code. The only file where the lines are reduced is mesonbuild/build.py.

tristan957:
Would be pretty interesting to see an `add_project_include_directories()` too.

xclaesse:
> Would be pretty interesting to see an `add_project_include_directories()` too.

Could be nice, but that could already be done easily with this PR:
```
incdir = include_directories('include')
dep = declare_dependency(include_directories: incdir)
add_project_dependency(dep)
```

Implementation looks good to me and I like the idea as well. LGTM.

test cases/common/251 add_project_dependencies/meson.build
```diff
@@ -0,0 +1,19 @@
+project('zlib system dependency', 'c')
+
+cc = meson.get_compiler('c')
+
+z = dependency('zlib', method: 'system', required: false)
+if not z.found()
+  error('MESON_SKIP_TEST zlib not present')
+endif
```
xclaesse, line 8: I actually would like to see in this unit test  adding `m_dep = cc.find_library('m')` to validate it works with library dependencies and also with not-found dependencies. From the code I'm pretty sure it works, but just to be sure.

xclaesse:
> > add_project_arguments and suchlike cannot be used after build targets have been defined, IIRC.
> 
> I think you could use `override_dependency` on a superproject and `add_project_dependency()` on a subproject, so I preferred to reuse the check on `is_built()`.

Using `is_built()` is stricter that relying on `add_project_arguments()` that checks we have no built targets in current project, because we also want to forbid using built targets from other subprojects. So I think we are all good on this.

tristan957:
> > Would be pretty interesting to see an `add_project_include_directories()` too.
> 
> Could be nice, but that could already be done easily with this PR:
> 
> ```
> incdir = include_directories('include')
> dep = declare_dependency(include_directories: incdir)
> add_project_dependency(dep)
> ```
> 
> Implementation looks good to me and I like the idea as well. LGTM.

Yep good point. My motivation for making that comment was essentially that we now have global options for all the important kwargs in build_target(): add_project_link_args(), add_project_args(), add_project_dependencies(). So adding add_project_include_directories() would finish it out in my opinion.

test cases/common/251 add_project_dependencies/meson.build
bonzini: Done.

bonzini:
@dcbaker, does it look good now?

bonzini:
Also ping @jpakkane.

jpakkane:
There needs to be a test that you can't call this function after any build targets have been defined.

```meson
project(...)
executable(...)
add_project_dependencies(...) # <- Must error out.
```

This is in line with how other such functions work.

eli-schwartz:
This does dispatch to `self._add_project_arguments()` so it should be enforcing that rule, although I guess in theory a refactoring could break that.

bonzini:
> This does dispatch to self._add_project_arguments() so it should be enforcing that rule

On top of that, it is also enforced by the fact that the dependency cannot be a "built" one. See also the new `test cases/failing/124 override and add_project_dependency`.

eli-schwartz:
That doesn't really assert anything as the restriction jpakkane mentioned is about forbidding any dependencies at all, not built ones specifically.

bonzini:
Ping.

docs/markdown/snippets/add_project_dependencies.md
```diff
@@ -0,0 +1,12 @@
+## `add_project_dependencies()` function
+
+Dependencies can now be added to all build products using
+`add_project_dependencies()`.  This can be useful in several
+cases:
+
+* with special dependencies such as `dependency('threads')`
+* with system libraries such as `find_library('m')`, or
+core dependencies such as `glib`
```
xclaesse, line 9: nitpicking: I don't think glib is a good example here because it will often have subproject fallback and in that case it won't be allowed in `add_project_dependencies()`, unless you do `if glib_dep.type_name() != 'internal'`.

xclaesse:
@bonzini it conflicts now, but looks good to me.

docs/markdown/snippets/add_project_dependencies.md
bonzini: Ok, I'll remove it.

xclaesse:
LGTM when CI is green, thanks!

PR #10207, opened 2022-03-29
docs: correct documentation of shared_library soversions

eli-schwartz:
We have always accepted an int here as an alternative to a string, but the initial documentation thought it was only a string.


eli-schwartz:
Personally, I don't find the use of an int here to be super interesting, but apparently it is supported?

annacrombie:
It is already relied on in several projects so I don't think it will be easy to drop support for.

PR #10208, opened 2022-03-29
Add new debug() function

ePirat:
Adds a new debug() function that can be used in the meson.build to
log messages to the meson-log.txt that will not be printed to stdout
when configuring the project.

### DIFF
--- a/unittests/platformagnostictests.py
+++ b/unittests/platformagnostictests.py
@@ -70,3 +70,16 @@ class PlatformAgnosticTests(BasePlatformTests):
     def test_python_dependency_without_pkgconfig(self):
         testdir = os.path.join(self.unit_test_dir, '102 python without pkgconfig')
         self.init(testdir, override_envvars={'PKG_CONFIG': 'notfound'})
+
+    def test_debug_function_outputs_to_meson_log(self):
+        testdir = os.path.join(self.unit_test_dir, '104 debug function')
+        log_msg = 'This is an example debug output, should only end up in debug log'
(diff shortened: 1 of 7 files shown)


unittests/platformagnostictests.py
```diff
@@ -70,3 +70,16 @@ def write_file(code: str):
...
+        testdir = os.path.join(self.unit_test_dir, '104 debug function')
+        log_msg = 'This is an example debug output, should only end up in debug log'
+        output = self.init(testdir)
+
+        # Check if message is not printed to stdout while configuring
+        assert(log_msg not in output)
+
+        # Check if message is written to the meson log
+        mesonlog = os.path.join(self.builddir, 'meson-logs/meson-log.txt')
+        with open(mesonlog, mode='r') as file:
```
xclaesse, line 84: I think this needs `encoding='utf-8'`

ePirat: Thanks, added.

xclaesse:
LGTM when CI is green, thanks!

xclaesse:
ah, new functions must be added into data/syntax-highlighting/vim/syntax/meson.vim and mesonbuild/ast/interpreter.py

xclaesse:
And if you're motivated, you can add it there too: https://github.com/mesonbuild/vscode-meson.

ePirat:
@xclaesse I think I added it to all needed places now, I hope.

xclaesse:
Looks good, thanks!

PR #10212, opened 2022-03-30
unittests: correctly use test asserts instead of regular ones

ePirat:


xclaesse:
Oops missed that during review. THanks.

PR #10213, opened 2022-03-30
compilers/gnu: use Popen_safe to prevent resource leaks

ePirat:
Fixes the following ResourceWarnings:

```
ResourceWarning: subprocess 25556 is still running
  _warn("subprocess %s is still running" % self.pid,
ResourceWarning: Enable tracemalloc to get the object allocation traceback

mesonbuild/compilers/mixins/gnu.py:195: ResourceWarning: unclosed file <_io.BufferedReader name=4>
  return gnulike_default_include_dirs(tuple(self.exelist), self.language).copy()
ResourceWarning: Enable tracemalloc to get the object allocation traceback
```


xclaesse:
LGTM when CI is green.

PR #10216, opened 2022-03-30
fix continued breakage in gnome module API

eli-schwartz:
In commit 823da3990947a8f4a2152826f0d7229f8a7a0159 we tried to fix disappearing dependencies. Instead, we appended the replacement dependencies to the existing ones. But this, too, was wrong. The function doesn't return new dependencies... it returns a copied list of all the dependencies, then alone of all parts of that API, expects to overwrite the existing variable.

(Sadly, part of the internals actually uses the entire list for something.)

As a result, we produced a repeatedly growing list, which eventually scaled really badly and e.g. OOMed on gstreamer.

Instead, let's just replace the dependencies with the updated copy.

### DIFF
--- a/mesonbuild/modules/gnome.py
+++ b/mesonbuild/modules/gnome.py
@@ -675,14 +675,14 @@ class GnomeModule(ExtensionModule):
                         internal_ldflags.update(libdepflags[1])
                         external_ldflags.update(libdepflags[2])
                         gi_includes.update(libdepflags[3])
-                        depends.extend(libdepflags[4])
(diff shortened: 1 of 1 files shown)


dcbaker:
That is... a terrible bit of code. My purpose was to limit side effects in called functions. But again, we're dealing with the fact that this whole module has the insane assumption that you're going to build all of Gnome as a unit test.

eli-schwartz:
Perhaps we need a CI runner just to test every commit touching the gnome module, with building all of gstreamer.

PR #5297, opened 2019-04-21
Add support for the Xtensa toolchain

greenfoo:
From (almost) all points of view, the Xtensa toolchain can be treated as
a regular GCC toolchain.

This patch adds very basic support so that, at least, meson does not
fail when trying to use "xt-xcc" (which makes it possible to use it
without problems).

PR #10218, opened 2022-03-31
Check for clang before guessing gcc or lcc

kraj:
clang --version can yield a string like below when its installed into
such a directory

clang version 14.0.0 (https://github.com/llvm/llvm-project 3f43d803382d57e3fc010ca19833077d1023e9c9)
Target: aarch64-yoe-linux
Thread model: posix
InstalledDir: /mnt/b/yoe/master/build/tmp/work/cortexa72-yoe-linux/gnome-text-editor/42.0-r0/recipe-sysroot-native/usr/bin/aarch64-yoe-linux

as you can see InstallDir has 'xt-' subtring and this trips the check to
guess gcc

if 'Free Software Foundation' in out or 'xt-' in out:

Therefore, check if compiler is clang then there is no point of running
this check anyway.

Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/mesonbuild/compilers/detect.py
+++ b/mesonbuild/compilers/detect.py
@@ -334,7 +334,7 @@ def _detect_c_or_cpp_compiler(env: 'Environment', lang: str, for_machine: Machin
         version = search_version(out)
 
         guess_gcc_or_lcc: T.Optional[str] = None
-        if 'Free Software Foundation' in out or 'xt-' in out:
+        if 'Free Software Foundation' in out or out.startswith('xt-'):
             guess_gcc_or_lcc = 'gcc'
         if 'e2k' in out and 'lcc' in out:
             guess_gcc_or_lcc = 'lcc'


tristan957:
I am not an experienced individual when it comes to this sort of thing. Is there any reason you can't just grep for clang in the version output? I assume some other compilers masquerade as clang without actually being clang?

kraj:
> I am not an experienced individual when it comes to this sort of thing and this part of the code. Is there any reason you can't just grep for clang in the version output? I assume some other compilers masquerade as clang without actually being clang?

sure you can but to what effect will it be better ?
grepping for stuff from compiler version output can be fragile as seen in this pull itself.

eli-schwartz:
> sure you can but to what effect will it be better ?
> grepping for stuff from compiler version output can be fragile as seen in this pull itself.

At least, grepping for very anonymous strings like "xt-" can be fragile. Looking for "clang" is probably a lot less fragile -- any version of clang should have that in the output, and any version of gcc or lcc should NOT.

The question then is, what's a better way to detect clang emitting a line with `InstalledDir: .*xt-.*`?
- seeing that the binary basename contains the string 'clang'
- seeing that 'clang' is in stdout
- seeing that 'InstalledDir:' is in stdout

Any of these options would suffice for actual clang, but the first option would NOT work for any compiler that demonstrates the same issue like:
```
 InstalledDir: /mnt/b/yoe/master/build/tmp/work/cortexa72-yoe-linux/gnome-text-editor/42.0-r0/recipe-sysroot-native/usr/bin/aarch64-yoe-linux
```

but is a fork of clang with a different binary name.

bruchar1:
What about checking only the first line of the output?

eli-schwartz:
That I don't know. The purpose of checking for xt- is because of the Xtensa toolchain, which is called `xt-xcc` and `xt-xc++` but I guess doesn't use the string "Free Software Foundation".

Is it always on the first line? Maybe, I don't have one to check.

xref #5297 so perhaps @greenfoo can answer that question.

eli-schwartz:
Fantastic, thank you for clarifying.

@kraj, can you switch to that approach?

bruchar1 (approved):

eli-schwartz (approved):

Issue #3653, opened 2018-06-01
[feature request] enable usage on pkg-config on MSYS2 (Windows)


franko:
Hi all,

I stumbled on a problem using meson on q MSYS2 system. On this system one can use unix-like path like /usr/lib on the MSYS2 bash shell.

If you install pkg-config using pacman, the standard packaging system, you gen a functional pkg-config. The problem arise when you invoke Meson and what happens is that Meson is not able to recognize and use the MSYS2-provided pkg-config.

By doing some investigations the problem seems to be the path:

```
> which pkg-config
/usr/bin/pkg-config
```

Otherwise the pkg-config is a native windows executable that can be ran without any problem.

I did some investigations and I think that Meson is confused by the fact that pkg-config is located in /usr/bin and this is not a Windows-like PATH.

So the paradox is that you have a working pkg-config from your shell, it is a native windows executable, but Meson is unable to use it.

I confess that I don't what could be the best approach to fix this problem but I use and I like very much Meson so I would love to have this case, usage inside MSYS2 bash, fixed.

Thank you in advance for any help.

Francesco

lazka:
Do you run meson with mingw Python or the cygwin one?

If the mingw one you need to install the mingw pkg-config: `pacman -S mingw-w64-i686-pkg-config`

nirbheek:
> The problem arise when you invoke Meson and what happens is that Meson is not able to recognize and use the MSYS2-provided pkg-config.

What's the return value of `shutil.which('pkg-config')` inside a python 3 REPL?

Like @lazka said, `mingw32/mingw-w64-i686-pkg-config` and `mingw64/mingw-w64-x86_64-pkg-config` in MSYS2 work fine. Are you sure you're not accidentally using the `msys/pkg-config` package instead? Confusingly, that's not supposed to be used for building with mingw ;)

franko:
Hi all,

actually I was using the pkg-config from msys but I tried using the one from mingw64 as suggested and I have the same problem as before.

Now, with the package from mingw64, I get from python3's shutil the following:
```
>>> shutil.which('pkg-config')
'/mingw64/bin/pkg-config'
```

and from the bash shell:
```
> which pkg-confi
/mingw64/bin/pkg-config
```

and still Meson obstinately refuse to use it! :-)

Here the output:
```
$ meson --buildtype=debug build-debug
The Meson build system
Version: 0.46.1
Source dir: /home/utente/dev/libcanvas
[...]

PR #10224, opened 2022-04-01
main: Make the msys/python check work again, in some cases

### LINKED ISSUE #8726: [Feature request, enhancement, minor] Please consider adding the pacman command after error "ERROR: This python3 seems to be msys/python on MSYS2 Windows, which is known to have path semantics incompatible with Meson"
Hey meson-folks,

Allow me to explain the issue first a bit. Today I compiled my first .exe on windows \o/ via msys2. And it worked!
(I am a linux person, so ... I have tons of ruby scripts that I use for batch-compiling everything.)

Anyway. Meson is very popular and for this we need python; python3.

The default install I did on windows msys2 was via pacman.

The message I get when I try to use meson 0.58.0 is this:

"ERROR: This python3 seems to be msys/python on MSYS2 Windows, which is known to have path semantics incompatible with Meson
ERROR: Please install and use mingw-w64-i686-python3 and/or mingw-w64-x86_64-python3 with Pacman"

See also here:

https://github.com/mesonbuild/meson/issues/8173

Obviously the best would be to get rid of it, and have meson just work as-is. But alright. ;)

Anyway - I found the command:

"pacman -S mingw-w64-i686-meson mingw-w64-i686-pkg-config"

I actually don't know pacman. :P

So that helped me.

Could this be included as an example, after the ERROR please? It may help future people who run
into this issue.

Perhaps show both variants, the other for 64 bit too, so people can pick. But I think this may
be better, because people can just copy/paste that command. So perhaps two more sentences
after the current ERROR message, would be appropriate, without becoming too overly verbose.

PS: You can state that it is not the job of meson to show this. Fair enough. But! Why does it then
mention the error "ERROR: Please install and use mingw-w64-i686-python3 and/or mingw-w64-x86_64-python3 with Pacman"?
So if it already does so, a copy/paste-able command would be better, IMO, after the explanation.

lazka:
msys/python in MSYS2 pretends to be cygwin in all cases for some time
now, so this check was impossible to hit.

The underlying confusion it tried to prevent is still there, namely trying
to build with mingw but wrongly using a msys/cygwin python/meson.

We can use the MSYSTEM env var to detect if we are in a mingw shell, and
in case the Python doesn't match we suggest installing mingw variants of both
python and meson.

Using msys/python + meson in a MSYS environment works fine on the other hand,
so no need to error out in that case.

Fixes #8726

Also addresses the concern raised in
https://github.com/mesonbuild/meson/issues/3653#issuecomment-474122564

### DIFF
--- a/mesonbuild/mesonmain.py
+++ b/mesonbuild/mesonmain.py
@@ -29,7 +29,6 @@ from . import mesonlib
 from . import mlog
 from . import mconf, mdist, minit, minstall, mintro, msetup, mtest, rewriter, msubprojects, munstable_coredata, mcompile, mdevenv
 from .mesonlib import MesonException, MesonBugException
-from .environment import detect_msys2_arch
 from .wrap import wraptool
 from .scripts import env2mfile
 
@@ -235,13 +234,9 @@ def run(original_args, mainfile):
     ensure_stdout_accepts_unicode()
 
     # https://github.com/mesonbuild/meson/issues/3653
-    if sys.platform.lower() == 'msys':
-        mlog.error('This python3 seems to be msys/python on MSYS2 Windows, which is known to have path semantics incompatible with Meson')
-        msys2_arch = detect_msys2_arch()
-        if msys2_arch:
-            mlog.error('Please install and use mingw-w64-i686-python3 and/or mingw-w64-x86_64-python3 with Pacman')
-        else:
-            mlog.error('Please download and use Python as detailed at: https://mesonbuild.com/Getting-meson.html')
+    if sys.platform == 'cygwin' and os.environ.get('MSYSTEM', '') not in ['MSYS', '']:
+        mlog.error('This python3 seems to be msys/python on MSYS2 Windows, but you are in a MinGW environment')
+        mlog.error('Please install and use mingw-w64-x86_64-python3 and/or mingw-w64-x86_64-meson with Pacman')
         return 2
 
     # Set the meson command that will be used to run scripts and so on


lazka:
Alternative is to remove that whole block, as it's a no-op right now.

mesonbuild/mesonmain.py
```diff
@@ -235,13 +234,9 @@ def run(original_args, mainfile):
...
-    if sys.platform.lower() == 'msys':
-        mlog.error('This python3 seems to be msys/python on MSYS2 Windows, which is known to have path semantics incompatible with Meson')
-        msys2_arch = detect_msys2_arch()
-        if msys2_arch:
-            mlog.error('Please install and use mingw-w64-i686-python3 and/or mingw-w64-x86_64-python3 with Pacman')
-        else:
-            mlog.error('Please download and use Python as detailed at: https://mesonbuild.com/Getting-meson.html')
+    if sys.platform == 'cygwin' and os.environ.get('MSYSTEM', '') not in ['MSYS', '']:
+        mlog.error('This python3 seems to be msys/python on MSYS2 Windows, but you are in a MinGW environment')
+        mlog.error('Please install and use mingw-w64-x86_64-python3 and/or mingw-w64-x86_64-meson with Pacman')
```
xclaesse, line 239: I don't think you can do 2 mlog.error() calls, the first one is boing to abort, no?

xclaesse (commented):
You know that better than me, so LGTM.

mesonbuild/mesonmain.py
lazka: hm, it did two `mlog.error` calls before this change too. And I just tested, and it works. I don't mind changing it though if you want.

xclaesse: My bad, I thought `mlog.error()` was always fatal, like `error()` in a `meson.build`.

lazka:
thanks!

PR #10231, opened 2022-04-03
cmake: Better error message when configuring a CMake subproject fails.

mensinda:
Implementation of https://github.com/mesonbuild/meson/pull/10152#issuecomment-1086336886


eli-schwartz:
Neat, thanks.

PR #10236, opened 2022-04-04
subprojects: Create `subprojects` directory if missing

otherJL0:
* File formatted with `black`
* Added new method `get_project_root` to find meson's project root directory
* Change behavior of `install` method to create `subprojects` directory in meson project root if missing instead of raising exception
* Replaced uses of `os.path` module with equivalent `pathlib` functions


eli-schwartz:
> File formatted with `black`

Please remove this commit, this project does not use black and changing the coding style of a project is not ideal.

anadon:
@otherJL0 Hello, I'm new here and trying to help out a bit.  Would you be able to provide a reproducer project that I can examine and use to verify that the change works?

mesonbuild/wrap/wraptool.py
```diff
@@ -98,14 +99,24 @@ def get_latest_version(name: str, allow_insecure: bool) -> T.Tuple[str, str]:
     version, revision = latest_version.rsplit('-', 1)
     return version, revision
 
+def get_project_root() -> pathlib.Path:
+    cwd = pathlib.Path.cwd()
+    while str(cwd) != cwd.root:
+        meson_build = cwd / "meson.build"
+        if meson_build.is_file():
+            return cwd
```
eli-schwartz, line 107: This functionality permits running `meson wrap install` from somewhere other than the project root, which I'm not sure should be allowed.

But it only does so if you are in a directory that doesn't have its own `subdir()` meson.build? This seems unreliable.

Currently the presence of the subprojects directory is a poor man's marker that you're in the project root and can act on things in there. I'd prefer any replacement of this to be something that detects if the user is already in the project root, and creates the subprojects directory as needed.

eli-schwartz (changes requested):
This could probably be a single commit. Also uses double quoted strings when general Meson style is to use single ones.

mesonbuild/wrap/wraptool.py
xclaesse: If I remember correctly, this has been requested and rejected in the past. We did not want to create files at random places if the user just ran the command in the wrong directory.

I personally don't have strong opinion against, but would prefer to not search in parent directories if current directory does not have meson.build file. Also could maybe uses mparser.py to check that the meson.build file in current directory starts with `project()`?

otherJL0: I agree with it being unreliable when considering the possibility of being in a sub-directory with its own `meson.build` file. And it does seem like the `subprojects` directory is a makeshift landmark for the project root, which maybe makes this attempt fruitless. Instead this approach can be applied to the [`update`](https://github.com/mesonbuild/meson/blob/7954a4c9cbf8355d8c8ea9b3d98df45d9f96f66e/mesonbuild/wrap/wraptool.py#L159) method further below and look for `subprojects` instead of a `meson.build` file.

I still think it would be beneficial to create a `subprojects` directory if nonexistent, though there are no clear paths forward. Some possible alternatives could be to look for landmarks like `README.*` or `.gitignore` but those can also exist in subdirectories like `meson.build`. There could be a growing list of project root files (like `.clang-format` or `.envrc`) that is used to determine the project root. But maybe the best indicator would be the presence of a `.git` directory. Of course this approach would fail if the user did not add any of these optional files but it adds functionality to those who do.

otherJL0: @xclaesse ah I see, that seems to be the most deterministic approach. I had not refreshed my page when I started writing my reply and did not see your two comments until just now.

eli-schwartz: As xclaesse elaborated upon, the correct approach broadly speaking is to check that there is a meson.build in the current directory and that the first non-whitespace non-comment line begins with `project(`.

There is code somewhere in the AST interpreter to do this, which may be easier than open-coding it using `open().read().startswith()`.


I do think this would be a nice improvement.

PR #10238, opened 2022-04-04
wayland: Mark module stable

xclaesse:
There is no point in printing warning about unstable module, in the
worst case we can just depracate and add new API. It has been tested in
a few projects already, and this warning is a blocker on wider adoption.

### DIFF
--- a/docs/markdown/Wayland-module.md
+++ b/docs/markdown/Wayland-module.md
@@ -1,13 +1,10 @@
-# Unstable Wayland Module
+# Wayland Module
 
-This module is available since version 0.62.0.
+This module is available since version 0.62.0, and has been stable since version
+1.8.0.
 
 This module provides helper functions to find wayland protocol
 xmls and to generate .c and .h files using wayland-scanner
-
-**Note**: this module is unstable. It is only provided as a technology
-preview. Its API may change in arbitrary ways between releases or it
-might be removed from Meson altogether.
 
 ## Quick Usage
 
@@ -15,7 +12,7 @@ might be removed from Meson altogether.
 project('hello-wayland', 'c')
 
 wl_dep = dependency('wayland-client')
-wl_mod = import('unstable-wayland')
+wl_mod = import('wayland')
 
 xml = wl_mod.find_protocol('xdg-shell')
 xdg_shell = wl_mod.scan_xml(xml)
--- a/mesonbuild/modules/wayland.py
+++ b/mesonbuild/modules/wayland.py
@@ -35,7 +35,7 @@ if T.TYPE_CHECKING:
 
 class WaylandModule(ExtensionModule):
 
-    INFO = ModuleInfo('wayland', '0.62.0', unstable=True)
+    INFO = ModuleInfo('wayland', '0.62.0', stabilized='1.8.0')
 
     def __init__(self, interpreter: Interpreter) -> None:
         super().__init__(interpreter)


xclaesse:
Wayland project refuse to use `meson.override_find_program()` which makes this module not work when wayland is a subproject. Until that gets fixed we cannot consider this module stable. https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/224.

fooishbar:
Yes, what a hopelessly frustrating experience. And the MR closed before I could comment (have been stuck in workshops lately), oh well.

Here's the issue:
* telling people 'use the Wayland module' is not helpful when it is marked as unstable and subject to constant change
* even the unstable module has only been released for about 2 or 3 weeks, and not everyone upgrades Meson every week
* regardless of that, pointing to GNOME and GStreamer using `find_program('wayland-scanner')` does not help when wayland-protocols, Weston, wlroots, and many others (mostly copied from these projects) do _not_ use them
* the wayland-scanner version used to generate the source files must be version-matched to the libwayland version used to build those source files
* this was previously ensured by getting the executable path via pkg-config, meaning that users only had to configure their pkg-config path to have this working (e.g. not finding a random `wayland-scanner` binary from the host which had no relation to their libwayland)
* the Wayland module now requires that modules using Meson to build change their build system, _and_ that users configure the `wayland-scanner` path in their native/cross files at the same time

Subjectively, I think it's a technically inferior solution, but whatever. I would've been happy to work through things in order to have a usable Wayland module.

Objectively, this has been a really horrible experience trying to get something done. Being **SHOUTED AT IN ALL CAPS** about how Wayland is doing Wayland wrong (because GNOME and GStreamer did it differently and never bothered to consult with Wayland - great start), being repeatedly told that we have to use a module which can't actually be used, being told that Meson's choice to screw up types by mixing up strings and paths is totally immutable and will never be changed but we were required to change us and all our downstream users (no doubt bearing all the pain and wrath of people having to make those changes) to work around this, MRs being randomly opened and closed, and pretty much just being arrogantly lectured to with no justification.

All my prior experiences with Meson have been great, but this one isn't.

If it really is down to doing exactly what you say or nothing happening - because Wayland is doing Wayland wrong and we haven't immediately changed our mind when presented with a fait accompli - then just delete this module entirely and maybe we can have a proper native module when someone else wants to work on it. Which is a massive shame for both projects.

xclaesse:
https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/224 got merged now, so I think we are all good to merge this PR too. Reopening.

xclaesse:
@eli-schwartz I think we landed everything needed to mark this module stable.

markbolhuis:
Before this is maybe stabilized I'd like to discuess the current behavior of `scan_xml`. 
Defaulting to client side is fine, but the way it works right now leads to misleading meson code:
```
xdg_shell = wl_mod.scan_xml(server : true)
```
Somebody reading that would not realize that it also generates client side code.

The easy fix is to set the default value of `client` to `False`, and we require the user to set either option explicitly.
We can also eliminate any error state by generating client side code when both are false.
```
scan_xml(...) # Client side only
scan_xml(client : true) # Client side only
scan_xml(server : true) # Server side only
scan_xml(client : true, server : true) # Both client and server side
```

That is a lot more intuitive.

xclaesse:
@markbolhuis I agree that's not optimal indeed. I was wondering, is there really a downside in always generating both? Other than wasting at most 42 nanoseconds of CPU time?

dcbaker:
> We can also eliminate any error state by generating client side code when both are false.

That sounds like a bad idea. setting both to false should be an error state.

I guess we could switch to a `mode : Literal[client | server | both] = client` model

xclaesse:
> > We can also eliminate any error state by generating client side code when both are false.
> 
> That sounds like a bad idea. setting both to false should be an error state.

I agree

> I guess we could switch to a `mode : Literal[client | server | both] = client` model

IIRC it was using a litteral in the first iteration of the module, but maybe with list[str] instead of having `both`. To be honest I don't mind either way, but before trying to find the best solution for that option, I would like to understand if we if need an option at all. Hate to quote him, but "the best part is no part" said Elon.

markbolhuis:
> is there really a downside in always generating both?

> I would like to understand if we if need an option at all. 

There is no downside other than wasting CPU time and cluttering the build directory with unused files. I don't like that personally, but if there is no way to provide an option cleanly then I guess It wouldn't hurt.

jadahl:
This was brought up on the Wayland IRC channel not long ago. Is it time?

orowith2os:
I would also like to make use of the wayland module in my own compositor, without relying on an unstable feature. How's it looking over here?

dcbaker:
@fooishbar, How are you feeling about this from the Wayland side at this point?

From my perspective on the Meson side we pretty much need to stabilize this if there isn't any upstream concerns, since it's got a longstanding API and people using it.

fooishbar:
Nope, no concerns. I’d love to go ahead.

dcbaker:
Updated to the newer mechanism for marking stable

dcbaker (approved):

PR #10239, opened 2022-04-04
Revert "wayland: Also lookup scanner in pkgconfig"

eli-schwartz:
Reverts mesonbuild/meson#10041

This did not get an approving review, and it did get a disapproving review.

Immediately after I clarified that I still disapprove the PR, it was merged by the original submitter.

@jpakkane

### DIFF
--- a/mesonbuild/interpreter/interpreter.py
+++ b/mesonbuild/interpreter/interpreter.py
@@ -1493,12 +1493,12 @@ external dependencies (including libraries) must go to "dependencies".''')
                           required: bool = True, silent: bool = True,
                           wanted: T.Union[str, T.List[str]] = '',
                           search_dirs: T.Optional[T.List[str]] = None,
-                          version_func: T.Optional[T.Callable[[T.Union['ExternalProgram', 'build.Executable', 'OverrideProgram']], str]] = None,
-                          depname: T.Optional[str] = None, varname: T.Optional[str] = None) -> T.Union['ExternalProgram', 'build.Executable', 'OverrideProgram']:
+                          version_func: T.Optional[T.Callable[[T.Union['ExternalProgram', 'build.Executable', 'OverrideProgram']], str]] = None
+                          ) -> T.Union['ExternalProgram', 'build.Executable', 'OverrideProgram']:
         args = mesonlib.listify(args)
 
         extra_info: T.List[mlog.TV_Loggable] = []
-        progobj = self.program_lookup(args, for_machine, required, search_dirs, depname, varname, extra_info)
+        progobj = self.program_lookup(args, for_machine, required, search_dirs, extra_info)
         if progobj is None:
             progobj = self.notfound_program(args)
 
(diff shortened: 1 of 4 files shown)


eli-schwartz:
> Instead of reverting, let's just print the warning

No, you misunderstand.

I am -0 on the entire PR as a whole, for reasons that don't really have to do with the warning per se, but my plan was to see what other people think about it rather than voice strong opinions.

I am -1 specifically on the warnings thing, and fixing the warnings would get us back to "Eli is -0 but that's okay, the real question is whether anyone does in fact approve it".

So printing the warning doesn't solve my concern that this should be reverted and discussed properly and approved by someone other than the submitter just like everyone else's PRs.

xclaesse:
I don't think we need to be too dramatic, if there were serious issues (other than the warning) someone would said already.

For the record, here is how I would like to use that as public API: https://github.com/mesonbuild/meson/pull/10242.

eli-schwartz:
> I don't think we need to be too dramatic, if there were serious issues (other than the warning) someone would said already.

Someone did say something -- I did. It's not just about the warning. But apparently because I *wasn't* dramatic about it, that got dismissed and then overridden via unreviewed merge.

But apparently this is all a moot point, because wayland upstream doesn't like that PR, or any PR (they would rather never use find_program at all), and no one agrees on what wayland should be doing, and they maybe want to have the unstable_wayland module removed from meson to boot. So who is made happy by the unreviewed PR being merged? Wayland doesn't like it. I don't like it. You don't like it anymore, and would rather doom unstable_wayland to be unstable forever.

*gah. sigh*

xclaesse:
> It's not just about the warning. But apparently because I _wasn't_ dramatic about it, that got dismissed and then overridden via unreviewed merge.

What is it about then? You prefer doing it in ModuleState instead of Interpreter? Sorry I thought that was a slight preference and not a real objection. Maybe you should have been more dramatic afterall :D

xclaesse:
@eli-schwartz Can you please clarify what you did not like with the PR (besides the warning) so we can move forward?

eli-schwartz:
Sorry, holiday preparation has been very hectic recently and I meant to respond earlier, and then this dropped off my plate.

- I would prefer it in ModuleState, yes, unless we actually intend to expose it in the meson.build DSL
  - I'm not sure it's very useful there, as looking up the dependency yourself gives you much more control.
  - if it were added, we'd want to duplicate quite a bit of functionality from `dependency()` itself. A dedicated version kwarg, because programs can have different versions than the framework that provides them -- this does happen, even if it is rare -- as well as the ability to check multiple names in order, or select different variable names for pkgconfig or cmake or config-tool etc, and maybe (I'm not sure, but I think there are likely use cases for this) also the native kwarg so that people can find cross-compiled tools in combination with an exe_wrapper for introspecting cross-compile properties
- we would want to check the version, but that has been implemented in a separate PR.
- I think we can probably do an upfront lookup of the dependency, see if it is found, and if so, invoke `state.find_program([depvar, 'progname'])` which means we no longer need to partially reimplement find_program_impl and the entire implementation of `state.get_native_binary` becomes pretty simple. Which in turn begs the question why add a public API?

xclaesse:
> * I think we can probably do an upfront lookup of the dependency, see if it is found, and if so, invoke `state.find_program([depvar, 'progname'])` which means we no longer need to partially reimplement find_program_impl and the entire implementation of `state.get_native_binary` becomes pretty simple. Which in turn begs the question why add a public API?

The issue with that is the program could have been overridden but not the dependency because `meson.override_find_program()` is older than `meson.override_dependency()`, in that case you would be using system program via pkgconfig instead of overridden one. I think that was the case for gnome tools, it is also the case of wayland-scanner.

In theory user could also set a different path in their machine file [binaries] and that should take precedence, but that's really corner case that does not really matter, I think.

xclaesse:
> * A dedicated version kwarg, because programs can have different versions than the framework that provides them

No, because we don't care about the version of the dependency, it does a version check on the program only, that was already covered. On the contrary, by doing it in ModuleState I have to either duplicate that logic, or rely on the pkgconfig version to be the same as program version, which is the case in current use-cases.

xclaesse:
Done in https://github.com/mesonbuild/meson/pull/10349

PR #10240, opened 2022-04-04
find_program: Print warning if variable is not found in pkgconfig file

xclaesse:


xclaesse:
Here is what gets logged in the case `wayland-scanner` pkgconfig file does not define `wayland_scanner` variable (it does, but I manually edited the pc file to test). Note that it still continues and find it in PATH.
```
Build-time dependency wayland-scanner found: YES 1.20.0
WARNING: pkgconfig variable 'wayland_scanner' not defined for dependency wayland-scanner.
Program wayland-scanner found: YES (/usr/bin/wayland-scanner)
```

xclaesse:
Done in https://github.com/mesonbuild/meson/pull/10349

PR #10242, opened 2022-04-04
find_program: Add from_dependency kwarg

xclaesse:
This allows searching for programs using pkgconfig when the pc file
    has a variable containing the path to the installed tool which could not
    always be in PATH.

TODO:
- unit test
- release snippet


bonzini:
> I have no strong preference either way, other than this MR is easier from an implementation POV. Anyone has an opinions?

The only qualm I have is the hardcoding of `native: true`. I am not sure it is correct 100% of the time, especially if for example Wine or 64bit->32bit compilation enters the picture. Allowing variables in internal dependencies would sidestep the issue.

PR #10243, opened 2022-04-04
Remove cython and vala before adding langs for host machine

tristan957:
Transpilers need to run on the build machine in order to generate their
output, which can then be taken by a cross-compiler to create the final
output.


tristan957:
Would appreciate feedback here. I definitely having this tuple here is stupid. There should probably be a list of transpiled languages somewhere or there should be a mesonlib function for checking if a language is transpiled.

I opening this because I don't think it is valuable to get a vala and cython compiler for the host machine, when all that matters is if the host machine post-transpilation compiler for transpiled output.

tristan957:
Almost wondering if cython and vala should be marked as required for the build machine where the other languages would be optional. All in all I think this issue is larger than this PR really since I may only want C for the build machine and not the host machine.

Having a more granular syntax for languages in project() would be nice.

eli-schwartz:
I wonder if, instead, in mesonbuild/compilers/detect.py, we should just have detect_cython_compiler and detect_vala_compiler ignore the for_machine argument and always do `_get_compilers(env, 'this-compiler', MachineChoice.BUILD)`?

The difference would be that if we do add_languages() we would still look up the language, but attempting to look up the cython cross compiler would always return the build executable anyway.

tristan957:
I quite like that idea. @jpakkane would you have any objections?

xclaesse:
I also prefer @eli-schwartz suggestion.

The side effect is you would now defined valac and cython path in native file instead of cross file `[binaries]`. This is not strictly backward compat but I think we can get away with that.

anadon:
@tristan957 Hey there.  I'm trying to earn my keep here a little and learn enough so that I can add transpiling for various Antlr4 grammars which looks like it would be affected by this work.  I have a 4-day weekend here in case you could throw a few bits my way to help out.

tristan957:
No idea how I could help you though! Best bet is to hop into the matrix channel.

anadon:
I figure I should help out here before I go consume people's time just for my stuff.  Besides, it is too steep for me to just add a new transpilation target without some smaller tasks in the system.  So I'm looking to help out with some easy stuff so I can earn my keep then be able to make the changes I need.

tristan957:
I would recommend checking out the issue tracker for easy issues then.

tristan957:
So I am still a little confused by Java, but should Java also receive the same fate as Vala and Cython? I have never heard of a Java cross compiler.

tristan957:
That would just be the C/C++ compiler though, right? Wouldn't need javac

dcbaker:
But the resulting binaries would require a host Java, right?

dcbaker (approved):

dcbaker:
Okay, we can deal with Java later if we need to

PR #10246, opened 2022-04-06
Add new message kwargs for status-message logging

ePirat:
This adds two new kwargs to the `message()`  function that allows printing messages just
like mesons own check-messages:

```
Found Foo: YES (0.10.0)
```

This can be incredibly useful when doing complex checks that meson does not
cover with a simple function with nice output but you still want a user friendly message
in the log/stdout about the check.

The new kwargs are:

- `status`: Boolean to indicate success or failure, will be printed as YES or NO.
- `extra_info`: Additional information to be printed after the status message.


mesonbuild/interpreter/interpreter.py
```diff
@@ -1234,15 +1234,31 @@ def func_add_languages(self, node: mparser.FunctionNode, args: T.Tuple[T.List[st
...
+                raise InvalidArguments('`extra_info` must be used together with `status`.')
+            mlog.log(mlog.bold('Message:'), *args)
+        else:
+            if len(''.join(args)) == 0:
+                raise InvalidArguments('message must not be empty when `status` kwarg is used.')
+            status_str = mlog.green('YES') if status else mlog.red('NO')
+            args[-1] += ':'
+            args.append(status_str)
+            if extra_info:
+                args.append(f'({extra_info})')
```
xclaesse, line 1260: I think it should use mlog.blue(), no?

mesonbuild/interpreter/interpreter.py
```diff
@@ -1234,15 +1234,31 @@ def func_add_languages(self, node: mparser.FunctionNode, args: T.Tuple[T.List[st
...
+    def message_impl(self, args, status: T.Optional[bool] = None, extra_info: T.Optional[str] = None):
+        if status is None:
+            if extra_info is not None:
+                raise InvalidArguments('`extra_info` must be used together with `status`.')
+            mlog.log(mlog.bold('Message:'), *args)
+        else:
+            if len(''.join(args)) == 0:
+                raise InvalidArguments('message must not be empty when `status` kwarg is used.')
+            status_str = mlog.green('YES') if status else mlog.red('NO')
+            args[-1] += ':'
```
xclaesse, line 1257: I don't this it is guaranteed that last arg is a string. It would be weird, but techically they could be int/bool, no? I think you can print the first part of the message with mlog.log(*args, end='') then the status with mlog.log(status_str, extra_info). Would that work?

xclaesse (commented):
Looks good overall. Missing release snippet and a unit test. Thanks :D

mesonbuild/interpreter/interpreter.py
ePirat: As far as I can see all cases that would ever end up here do `stringifyUserArguments` first, so it seems it would always be a string, no?

xclaesse: oh no, forget me, that code is AFTER stringifyUserArguments(). All good.

mesonbuild/interpreter/interpreter.py
ePirat: Indeed.

mesonbuild/interpreter/interpreter.py
ePirat: (I agree its not super nice to read but after this PR I plan to refactor some of this anyway in mlog.py itself that hopefully makes this much easier)

ePirat:
@xclaesse Regarding unit tests it would probably be wise to wait for  !10215 so I can use `get_meson_log_raw()` without then having to clean-up after me again if this is merged before !10215 is…

eli-schwartz:
> The new kwargs are:
> 
>     * `status`: Boolean to indicate success or failure, will be printed as YES or NO.
> 
>     * `extra_info`: Additional information to be printed after the status message.

This feels like awkward UX to pass multiple bits of information here. I am reminded that `summary()` was enhanced to accept, as an alternative to `foo_dep.found()`, `foo_dep` itself. This then uses `Dependency.summary_value()` to print information -- it internally checks if the dependency was found, and then produces a return value of `mlog.AnsiText(mlog.green('YES'), ' ', mlog.cyan(self.version))`.

So if `status: foo_dep` was supported, you would not need to be passing extra_info, merely to have something "everyone" agrees is a good `summary_value()` for the passed object.

It would then work OOTB for such a use case.

I'm not entirely sure under which circumstances it's important to put version info after (not before) the "YES" / "NO" status, especially the one in the example is hardcoded anyway.

ePirat:
The primary intention for this API is to cover complex checks that are not and can not properly be displayed by meson, so I don't think accepting a dep for example makes much sense here.

While I hardcoded the values for simplicity that's not how it will be used most of the time, of course:

```meson
mingw_version_major = ''
mingw_check = '''
    #ifndef __MINGW32__
    # error Not compiling with mingw
    #endif
'''

# Check if MinGW is used at all
has_mingw = cc.compiles(mingw_check)
if has_mingw
    # Check which kind of MinGW
    mingw_version_major = cc.get_define('__MINGW64_VERSION_MAJOR',
        prefix : '#include <_mingw.h>', silent: true)
endif

message('Has MinGW-w64 5.0 or higher', status: (mingw_version_major != '' and mingw_version_major.to_int() >= 5), extra_info: mingw_version_major)
if mingw_version_major == ''
    error('Cannot compile with MinGW, use MinGW-w64 >= 5.0 instead.')
endif

# Check that MinGW w64 is at least 5.0
if mingw_version_major.to_int() < 5
    error('MinGW-w64 5.0 or higher required!')
endif
```

I am open to suggestions for how to improve the API, of course. This was just the easiest I could come up with without introducing a new function.

(I will mark it as draft until tests are implemented covering all relevant cases)

bonzini:
I understand the API is intended for complex cases, but most of the time the result of complex tests can be summarized in a dependency or program object, so the status argument should accept a dependency or program.

If the status is not a boolean, the extra_info argument should be rejected and `message()` would print the same information as `summary()`. But then I wonder if adding extra_info is really necessary, and which usecases are not satisfied by passing a dependency or program objects.

ePirat:
@bonzini Thanks for the suggestions.

We discussed on IRC that approach but the main issue with accepting a dependency is, how would you. get it? The only reasonable case where you get a dependency that is not already logged by meson would be `declare_dependency` and for that you have the issue that you would have none in the failure case, because how would you create one if you have nothing to put in it. And even if you were to create a empty `dependency('', required: false)` one for that case, which feels quite awkward to me, there is not enough info for proper logging in there, so you would anyway need to provide a full reasonable message and all you would safe here would be to type `.found()`.

Another issue I see with that, take my example I gave above, in this case it would be completely pointless to create a dependency as it would only ever be used for the logging. In fact most cases where I want to use this API do not have either a dependency of program involved, so I am curious on your use-cases here, so please elaborate.

bonzini:
No problem, I don't oppose in principle having `extra_info` if you have a candidate for using it. But it should accept everything that `summary` gets, too.

xclaesse:
I'm sorry to contradict I said previously, but I'm starting to think this should be its own `status()` (or `status_message()`?) function because it seems we have ideas to grow it with more functionalities than I first anticipated, which is a good thing.

That said, I personally don't think supporting `dependency()` is a blocker to get this merged first. We can always extend over time. `summary()` itself did not support that initially.

PR #10250, opened 2022-04-08
cmake: Always use all compilers for LLVM (fixes #10249)

### LINKED ISSUE #10249: [frameworks: 15 llvm] fails with llvm/clang 14.0.0
As pointed out here: https://github.com/mesonbuild/meson/pull/10248#issuecomment-1092454817

The error I see in the logs (`check_source_compiles: C: needs to be enabled before use.`) suggest it is https://github.com/llvm/llvm-project/issues/53950

And indeed

```diff
diff --git a/mesonbuild/dependencies/data/CMakeListsLLVM.txt b/mesonbuild/dependencies/data/CMakeListsLLVM.txt
index 9d3e41234..4084597d3 100644
--- a/mesonbuild/dependencies/data/CMakeListsLLVM.txt
+++ b/mesonbuild/dependencies/data/CMakeListsLLVM.txt
@@ -2,6 +2,8 @@ cmake_minimum_required(VERSION ${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION}.${C
 
 set(PACKAGE_FOUND FALSE)
 
+project(MesonTemp LANGUAGES C CXX)
+
 while(TRUE)
   find_package(LLVM REQUIRED CONFIG QUIET)
```

seems to make it work here.

I see that the generic CMakeLists file has some logic to inject the language, so I'm not sure if that should be extended or if the above patch is enough.

based on the file history -> pinging @mensinda

mensinda:


eli-schwartz:
Maybe I'm missing the way this works, but it sounded to me like we specifically need a C compiler, which is probably not guaranteed to be in the global compilers?

So how does this guarantee that the problem is fixed?

mensinda:
By default, Meson adds all known compilers to the CMake project call that it can map to CMake. Without this change, the LLVM compiler had previously only added a C++ compiler.

Hardcoding a C compiler would not really solve the problem since it is still required to have Meson know a (or rather the correct) C compiler it can then give to CMake.

I have added a check to make sure that both compilers must be present.

mesonbuild/dependencies/dev.py
```diff
@@ -404,7 +404,19 @@ class LLVMDependencyCMake(CMakeDependency):
...
-        super().__init__(name, env, kwargs, language='cpp')
+
+        compilers = None
+        if kwargs.get('native', False):
+            compilers = env.coredata.compilers.build
+        else:
+            compilers = env.coredata.compilers.host
+        if not compilers and not all(x in compilers for x in ('c', 'cpp')):
+            if kwargs.get('required', True):
+                raise self._gen_exception(f'The CMake LLVM dependency needs both a C and C++ compiler')
```
eli-schwartz, line 415: Is an exception the right thing to do here? It's possible to fall back on another method, right?

eli-schwartz: @mensinda ping

I think if we just print a warning here (and return early with `self.found = False`), we will anyways error out when the dependency is not found, depending on the value of `required` *and* whether or not other methods were tried and succeeded/failed.

mesonbuild/dependencies/dev.py
```diff
@@ -404,7 +404,18 @@ class LLVMDependencyCMake(CMakeDependency):
...
         self.llvm_modules = stringlistify(extract_as_list(kwargs, 'modules'))
         self.llvm_opt_modules = stringlistify(extract_as_list(kwargs, 'optional_modules'))
-        super().__init__(name, env, kwargs, language='cpp')
+
+        compilers = None
+        if kwargs.get('native', False):
+            compilers = env.coredata.compilers.build
+        else:
+            compilers = env.coredata.compilers.host
+        if not compilers and not all(x in compilers for x in ('c', 'cpp')):
```
eli-schwartz, line 413: Should this be:

```suggestion
        if not compilers or not all(x in compilers for x in ('c', 'cpp')):
```

mensinda: Good catch.

PR #10253, opened 2022-04-09
Validate that introspection command exited cleanly.

jpakkane:
In Debian sid the script actually fails with this:

```
/root/./pyscript.py:6: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
  File "/root/./pyscript.py", line 6, in <module>
ModuleNotFoundError: No module named 'distutils.command'
```

which gets written to `stderr`. `stdout` is empty which apparently is valid JSON so it will not cause any errors.

### DIFF
--- a/mesonbuild/modules/python.py
+++ b/mesonbuild/modules/python.py
@@ -415,6 +415,8 @@ class PythonExternalProgram(ExternalProgram):
         cmd = self.get_command() + [tmpfilename]
         p, stdout, stderr = mesonlib.Popen_safe(cmd)
         os.unlink(tmpfilename)
+        if p.returncode != 0:
+            raise mesonlib.MesonException('Trying to introspect Python configuration failed.\n\n'  + stdout + stderr)
         try:
             info = json.loads(stdout)
         except json.JSONDecodeError:


eli-schwartz:
>`stdout` is empty which apparently is valid JSON so it will not cause any errors.

For me this returns: `json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)`

A valid empty json document is `'{}'`

On debian, distutils is erroneously not included in the stdlib, which is a Debian bug that Debian itself calls a "feature".

This is supposed to `raise mesonlib.MesonException(f'{python} is not a valid python or it is missing distutils')` if the dependency cannot be found, but that got broken in commit c5c02b72e132a03ae0284bdd5e15d1675301a37d, cc @xclaesse 

The dropped logic here seems... unfortunate.

eli-schwartz:
Re-added the missing logging in #10257

eli-schwartz:
Taking a closer look, this commit is wrong for another reason. It unconditionally raises, but we used to treat it as a non-fatal warning as long as it is `required: false`.

My PR that I referenced above returns it to being a warning instead of silently ignored...

jpakkane:
All tests pass (the two are unrelated). I'd stay this still makes a useful sanity check addition.

eli-schwartz:
But it actively breaks existing use cases, as I said. We can't make this fail, because if the python installation isn't required then having a broken one should not be worse than having none at all.

Furthermore, it should be impossible to ever hit this because an empty string ***isn't*** valid JSON and python tracebacks if you try to parse it... Which means we already hit the error condition, log it as not found, and mlog.debug() some relevant context about how it failed.

Furthermore, we require a "version" key in the returned json in order to successfully process anything here.

So I don't believe this is a net positive sanity check, and I also don't understand how to reproduce the problem you say you had.

jpakkane:
With the latest trunk and sid the issue does not occur any more. Closing.

PR #10254, opened 2022-04-09
Use a temp file to invoke the introspection command.

jpakkane:


eli-schwartz:
Can you elaborate on what this is trying to fix? I would natively assume there should be no difference. Is there a bug report somewhere that explains under what circumstances the current approach fails?

jpakkane:
In #10253 things fail on certain platforms. Probably because the command line is too long.

eli-schwartz:
That doesn't seem to have anything to do with the command line being too long. It happens in the check that `dir` should correctly report a bad python. The `dir` command fails because it expects a filename, not an arbitrary string, and because it does additional validation there, the command itself reports that it successfully executed, then errored out in its own argument parsing logic when given something that it "knows" can't be a disk filename.

jpakkane:
Thinking further, this should be a script that can be invoked via `meson --internal` or something. This makes it easier to develop and debug as you can run it standalone from the command line (I had to develop this by copypasting the script to its own file, edit and run that and then copypaste it back.

jpakkane:
Doesn't help with running it via an external Python executable, though..

eli-schwartz:
It can't be run via `meson --internal` because it needs to execute successfully under python2, python 3.4, and arbitrary pythons that aren't the one meson uses.

Now that we rely on python 3.7 we no longer need to assume that we cannot access files from the mesonbuild package (we used to do the "write a file" trick for cmakelists.txt in the cmake module too) so we could use importlib.resources to access it as a data file.

I'm not sure how well that plays with PyInstaller though. Or alternatively, I'm not sure that this actually helps run the script for debugging. In the best case scenario you still have to manually pass the filename from a nested directory in site-packages/.

jpakkane:
Ignoring the sidetracks, there is a reason for this on its own. Invoking via `-c` has a failure mode where things start failing if the command script becomes too long (it might not be yet, but as the code is altered in the future, it might be). Writing to a file is more reliable and can't fail unless the hard drive is 100% full. This is unlikely and, if it happens, many other things break first.

So even if nothing is broken yet, this is a future-proofing thing.

eli-schwartz:
Sorry, yeah, I 100% agree about that. And it's not just about the command line, but the env, which can be depleted in a couple of ways.

We probably only used -c because when this was originally implemented having it as a data file was unreliable without pkg_resources. ;)

(I think importlib.resources, which is the same thing but in the stdlib, is a bit more elegant than writing a temp file manually.)

Either way I think it would be neat if the commit message could explain that rationale ("future proofing" is good enough).

jpakkane:
Fixed commit message and pushed directly to master to save CI resources.

nirbheek:
FTR, master branch commit was https://github.com/mesonbuild/meson/commit/3c80f8f96591a03207b58e8201700bfcbcfa95a1

PR #10257, opened 2022-04-10
python module: restore logging for broken python

eli-schwartz:
This check was erroneously removed in commit c5c02b72e132a03ae0284bdd5e15d1675301a37d


xclaesse:
Good catch thanks, LGTM.

PR #10258, opened 2022-04-10
Fix universal builds on Darwin PPC: add ppc7400 definition

barracuda156:
Solves the problem on Darwin PowerPC, when `meson` fails to identify `ppc7400` as `ppc`, which causes universal build failure.
Example: https://trac.macports.org/ticket/64787

P. S. Should these be added or not? I did not encounter such cases myself so far, but I seldom use cpu-specific optimizations.
```
stdo = stdo.replace('ppc7450', 'ppc')
stdo = stdo.replace('ppc970', 'ppc')
stdo = stdo.replace('ppc970-64', 'ppc64')
```
Ref: https://opensource.apple.com/source/cctools/cctools-836/cctools-836/libmacho/arch.c

### DIFF
--- a/mesonbuild/mesonlib/universal.py
+++ b/mesonbuild/mesonlib/universal.py
@@ -710,6 +710,7 @@ def darwin_get_object_archs(objpath: str) -> 'ImmutableListProtocol[str]':
     # Convert from lipo-style archs to meson-style CPUs
     stdo = stdo.replace('i386', 'x86')
     stdo = stdo.replace('arm64', 'aarch64')
+    stdo = stdo.replace('ppc7400', 'ppc')
     # Add generic name for armv7 and armv7s
     if 'armv7' in stdo:
         stdo += ' arm'
--- a/mesonbuild/mesonlib/universal.py
+++ b/mesonbuild/mesonlib/universal.py
@@ -711,6 +711,7 @@ def darwin_get_object_archs(objpath: str) -> 'ImmutableListProtocol[str]':
     stdo = stdo.replace('i386', 'x86')
     stdo = stdo.replace('arm64', 'aarch64')
     stdo = stdo.replace('ppc7400', 'ppc')
+    stdo = stdo.replace('ppc970', 'ppc')
     # Add generic name for armv7 and armv7s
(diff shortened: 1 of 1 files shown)


eli-schwartz:
Please rebase to remove the merge commit.

barracuda156:
> Please rebase to remove the merge commit.
> 
> See the second paragraph of https://mesonbuild.com/Contributing.html#keeping-pull-requests-up-to-date

I think I have done it now.

eli-schwartz:
> P. S. Should these be added or not? I did not encounter such cases myself so far, but I seldom use cpu-specific optimizations.

This seems plausible, but I admit to not really knowing anything about darwin here. I guess the logic for adding those would be the same as for adding ppc7400?

Note: "Update universal.py" seems a bit obscure as a commit message title, perhaps it would make sense to amend the commit and use the same title as the title of the PR?

barracuda156:
> > P. S. Should these be added or not? I did not encounter such cases myself so far, but I seldom use cpu-specific optimizations.
> 
> This seems plausible, but I admit to not really knowing anything about darwin here. I guess the logic for adding those would be the same as for adding ppc7400?

@eli-schwartz Yes, the logic is correct. Just adding `ppc7400` is crucial since otherwise universal builds are broken on PPC, while `ppc970` would perhaps be a rather exotic case. I need to check on G5 if building something with `-mtune=970` or `-mcpu=970` will result in binaries which lipo gonna see as `ppc970`, but looks like yes: https://groups.google.com/g/hugin-ptx/c/uWSPhl1MjCc/m/gYyxZemL6bsJ

barracuda156:
> Note: "Update universal.py" seems a bit obscure as a commit message title, perhaps it would make sense to amend the commit and use the same title as the title of the PR?

@eli-schwartz Fixed that.

@jpakkane Is it good as it is, or should I add `ppc970`? (Other arch variants, I think, are unneeded.)

jpakkane:
If it is actually used in the wild, do add it.

barracuda156:
> If it is actually used in the wild, do add it.

Thank you, added.

PR #10265, opened 2022-04-11
wayland: Make sure wayland-scanner has same version as wayland libraries

xclaesse:


ppaalanen:
Wayland-scanner may make use of newly added ABI in libwayland, so libwayland cannot be older than the scanner.

That means the minimum libwayland version you can run with is affected by the scanner version used to generate the code.

OTOH, wayland-scanner could be older than libwayland, ABI stability requires that to work. But, sometimes bug fixes require a new scanner, so if stuff was built with old scanner, it might miss some fixes. (There were some threading problems at some point, that needed a new scanner and new ABI to fix.)

So I think requiring an exact version match is justified.

xclaesse:
@ppaalanen ok, fixed.

ppaalanen:
The version check looks good to me now, for what I understand of the code, but is it ok to hardcode `wayland-client` there?

Someone might be building a compositor and need only `wayland-server`. How determined does one need to be in shooting their own foot to have different versions of `wayland-client` and `wayland-server`?

xclaesse:
> How determined does one need to be in shooting their own foot to have different versions of `wayland-client` and `wayland-server`?

Wayland does not allow building one without the other, and at least Ubuntu package them together. Seems pretty safe to assume we can check either of them, and it won't matter. IMHO at some point it become user's problem if they are asking for trouble.

nirbheek:
I guess this should also go into the stable release?

xclaesse:
> I guess this should also go into the stable release?

It could but not essential. The module is marked unstable and I don't think it has anyone using it yet.

nirbheek:
> It could but not essential. The module is marked unstable and I don't think it has anyone using it yet.

that just makes it all the more reason to get this in, IMO. Unstable modules can get more risky bugfixes in point releases.

PR #10266, opened 2022-04-11
Add regression test for Python dist.

jpakkane:


unittests/pythontests.py
```diff
@@ -43,3 +48,15 @@ def test_bad_versions(self):
...
             self.init(testdir, extra_args=['-Dpython=dir'])
         self.wipe()
+
+    def test_dist(self):
+        with TemporaryDirectoryWinProof() as dirstr:
+            dirobj = pathlib.Path(dirstr)
+            mesonfile = dirobj / 'meson.build'
+            mesonfile.write_text('''project('test', 'c', version: '1')
+pymod = import('python')
+python = pymod.find_installation('python3', required: true)
```
tristan957, line 58: True is probably the default for required, so maybe remove it

jpakkane:
I basically agree with you but I'll merge this directly since the changes are not really necessary to reduce the load on CI.

tristan957:
Makes sense

Issue #9503, opened 2021-10-30
Yield options cause configure output to not be in sync


tristan957:
**Describe the bug**

When putting a message(get_option()) for a option which has been marked as yielding, the output in `meson configure` becomes inconsistent with the true value printed the by the aforementioned message().

**To Reproduce**

```meson
# root meson.build
project('yo')

subproject('sub')

message(get_option('lol'))
```

```meson
# for both parent project and subproject
option('lol', type: 'boolean', value: true, yield: true)
```

```meson
# sub meson.build
project('yoyo')

message(get_option('lol'))
```

```text
meson build
meson configure build
meson configure build -Dlol=false
meson configure build
```

Notice output in the configure operations.

**Expected behavior**

`meson configure` should be consistent with actual state.

**system parameters**
* Is this a [cross build](https://mesonbuild.com/Cross-compilation.html) or just a plain native build (for the same computer)? No
* what operating system (e.g. MacOS Catalina, Windows 10, CentOS 8.0, Ubuntu 18.04, etc.): Fedora 34
* what Python version are you using e.g. 3.8.0: 3.9
* what `meson --version` 0.58.2
* what `ninja --version` if it's a Ninja build: 1.10.2.git.kitware.jobserver-1

tristan957:
I should probably confirm this on master or even latest 0.60.1

xclaesse:
> option('lol', type: 'boolean', value: true)
It's not yielding, is that a typo in your big report?

eli-schwartz:
It's not a typo; add the yield kwarg and you still get this:

```
$ meson setup builddir/ -Dlol=false
The Meson build system
Version: 0.60.0
Source dir: /tmp
Build dir: /tmp/builddir
Build type: native build
Project name: yo
Project version: undefined
Host machine cpu family: x86_64
Host machine cpu: x86_64

Executing subproject sub 

sub| Project name: yoyo sub
sub| Project version: undefined
sub| Message: lol is false
sub| Build targets in project: 0
sub| Subproject sub finished.

Message: lol is false
Build targets in project: 0

yo undefined

  Subprojects
    sub: YES

  User defined options
    lol: false

Found ninja-1.10.2 at /usr/bin/ninja

$ meson configure builddir/
Core properties:
  Source dir /tmp
  Build dir  /tmp/builddir

Main project options:
  
[...]
  
  Project options                 Current Value                     Possible Values                   Description
  ---------------                 -------------                     ---------------                   -----------
[...]

PR #10270, opened 2022-04-12
mconf: Fix printing <inherited from main project> for yielding options

xclaesse:
In print_options() k was a string instead of OptionKey, but
self.yielding_options expects OptionKey. Not sure how this has not been
catched by mypy.

Fix by keeping k as OptionKey which makes self.yielding_options useless.

Fixes: #9503


xclaesse:
This is a regression, but I don't know since when it has been broken.

PR #10271, opened 2022-04-12
coredata: Abort when setting value on yielding option

xclaesse:


thiblahute:
I think I do not agree about that behavior, and instead, if a user want to override an option that was yielded it gets overriden.

xclaesse:
> I think I do not agree about that behavior, and I think that instead, if a user wants to override an option that was yielded it gets overriden.

I agree with that. The reason I made this PR is because whatever decision we want, current status of silently ignoring it is unacceptable. Making it work as you said would require more refactoring (maybe not as hard as I fear), but it is always possible to change from forbidden to allowed later, but not the other way around. 

That said, I should probably downgrade to a warning for now and get that into 0.62.1?

jpakkane:
> I think I do not agree about that behavior, and I think that instead, if a user wants to override an option that was yielded it gets overriden.

No. The semantics of yielding is "always use the value in the parent project". That is the entire point of it.

xclaesse:
Another reason for allowing setting value on yielding subproject option is that's how it works for per-subproject builtin options. They inherit the value from main project but user can set a different value explicitly. Currently built-in and project options are inconsistent.

See https://mesonbuild.com/Builtin-options.html#specifying-options-per-subproject.

eli-schwartz:
Intuitively it makes sense to me that a yielding option should set its default by yielding, not error out if for some reason you really want to fine-tune it per subproject.

xclaesse:
> we might have a project with a given option name and a subproject having the same option name which are not related.

In that case the subproject should not set `yielding: true` on that option and Meson will keep them completely separated.

dabrain34:
https://gitlab.freedesktop.org/gstreamer/gstreamer/-/merge_requests/4128#note_1815317

PR #10272, opened 2022-04-12
Add include-what-you-use usage info to FAQ

alex-tee:


docs/markdown/FAQ.md
```diff
@@ -683,3 +683,15 @@ executable(
   link_language : 'c',
 )
 ```
+
+## How do I use include-what-you-use with meson?
```
tristan957, line 687: Put `include-what-you-use` in backticks.

docs/markdown/FAQ.md
```diff
@@ -683,3 +683,21 @@ executable(
...
+
+```console
+$ /path/to/iwyu_tool.py -p build
+```
+
+See `--help` for more options.
+
+A run target can also be added with the following meson code:
+```meson
+iwyu_tool = find_program('iwyu_tool.py')
```
tristan957, line 701: I don't think `.py` exists if you install the package for instance. @xclaesse might be nice to be able to specify multiple program names similar to multiple dependency names for times like this.

docs/markdown/FAQ.md
```diff
@@ -683,3 +683,21 @@ executable(
...
+run_target('iwyu', command: [iwyu_tool, '-p', meson.project_build_dir()])
```
tristan957, line 702: I think it is `project_build_root()`

docs/markdown/FAQ.md
alex-tee: it seems to be installed as `iwyu-tool`. updated the patch

docs/markdown/FAQ.md
```diff
@@ -683,3 +683,21 @@ executable(
...
+
+Following the steps from
+[Using with a compilation database](https://github.com/include-what-you-use/include-what-you-use#using-with-a-compilation-database),
+configure the build directory (e.g., with `meson setup build`), and run:
+
+```console
+$ iwyu-tool -p build
+```
+
+See `--help` for more options.
```
tristan957, line 697: Maybe "See `iwyu_tool --help`"

docs/markdown/FAQ.md
```diff
@@ -683,3 +683,21 @@ executable(
...
 ```
+
+## How do I use `include-what-you-use` with meson?
+
+Following the steps from
+[Using with a compilation database](https://github.com/include-what-you-use/include-what-you-use#using-with-a-compilation-database),
+configure the build directory (e.g., with `meson setup build`), and run:
+
+```console
+$ iwyu-tool -p build
```
tristan957, line 694: Should be iwyu_tool, right?

docs/markdown/FAQ.md
xclaesse: find_program() has always accepted multiple names ;-)

docs/markdown/FAQ.md
tristan957: Hmm. I think on ubuntu the binaries are iwyu and iwyu_tool

alex-tee: oh you're right https://packages.ubuntu.com/bionic/amd64/iwyu/filelist

docs/markdown/FAQ.md
```diff
@@ -683,3 +683,21 @@ executable(
...
+```
+
+See `iwyu-tool --help` for more options.
+
+A run target can also be added with the following meson code:
+```meson
+iwyu_tool = find_program(['iwyu-tool', 'iwyu_tool'])
```
tristan957, line 701: This shouldn't need to be in an array.

docs/markdown/FAQ.md
```diff
@@ -683,3 +683,21 @@ executable(
   link_language : 'c',
 )
 ```
+
+## How do I use `include-what-you-use` with meson?
```
tristan957, line 687: Capitalize Meson

eli-schwartz:
Cross posting from IRC...

I wonder if instead we should do like we do already for:
- scan-build
- clang-format
- clang-tidy
- ctags

and try to detect the presence of the program on the build machine. If it's there, we generate a rule automatically to do this -- unless the meson.build already defines a run_target for it.

docs/markdown/FAQ.md
```diff
@@ -683,3 +683,21 @@ executable(
...
+```console
+$ iwyu-tool -p build
+```
+
+See `iwyu-tool --help` for more options.
+
+A run target can also be added with the following meson code:
+```meson
+iwyu_tool = find_program('iwyu-tool', 'iwyu_tool')
+run_target('iwyu', command: [iwyu_tool, '-p', meson.project_build_root()])
```
eli-schwartz, line 702: This can't be the project build root, as that will be wrong inside a subproject

tristan957: I don't think run_targets() get generated for subprojects, so it doesn't really matter. I have never really gotten confirmation on this though

jpakkane: There's `meson.build_root()`. But maybe it would be better to have an ifdef so the target is only created if the current project is the toplevel project (which is what most people probably expect).

jpakkane: If memory serves, subproject run targets are generated as `subprojname:targetname`.

tristan957: *meson.global_build_root() probably

tristan957:
```console
$ ninja -C build/ hse:shellcheck
ninja: Entering directory `build/'
ninja: error: unknown target 'hse:shellcheck', did you mean 'hse@@shellcheck'?
```
That `@@` is gross. Is that right?

docs/markdown/FAQ.md
```diff
@@ -683,3 +683,21 @@ executable(
...
+
+```console
+$ iwyu-tool -p build
+```
+
+See `iwyu-tool --help` for more options.
+
+A run target can also be added with the following meson code:
+```meson
+iwyu_tool = find_program('iwyu-tool', 'iwyu_tool')
```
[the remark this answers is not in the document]
tristan957: Maybe `iwyu_tool.py` should be added to the list as well. I am under the impression that the distros that rename it to `iwyu-tool` are the ones that hate users :)

tristan957:
> Cross posting from IRC...
> 
> I wonder if instead we should do like we do already for:
> 
> * scan-build
> * clang-format
> * clang-tidy
> * ctags
> 
> and try to detect the presence of the program on the build machine. If it's there, we generate a rule automatically to do this -- unless the meson.build already defines a run_target for it.

If Meson intends to keep these, then I think `iwyu` is on this level, but my opinion is that these are not really in scope for Meson and it isn't hard to write these yourself. Others may have different opinions though.

docs/markdown/FAQ.md
eli-schwartz: It's an illegal filename character, and this was the rationale in commit 02bea7d5bf5c98c586e59e60357fc5ca11826172.

Granted, a run_target isn't a "real" filename, as ninja never expects to actually *create* it... but it is written to build.ninja as though it were one.

`meson test` doesn't even pretend to take filenames.

tristan957: > It's an illegal filename character

Never knew this. Good to know then.

eli-schwartz: (Technically it's not an illegal filename character, it's a *sometimes* illegal one. It's illegal on Windows, but not Unix.)

docs/markdown/FAQ.md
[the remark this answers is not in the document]
alex-tee: updated all references to `iwyu_tool.py` and added all 3 in the `find_program()` invocation

tristan957:
So I think there are 3 things left for this PR:

1. Change meson.project_build_root() to meson.global_build_root()
2. Potentially wrap the run_target() snippent is an `if not meson.is_subproject() ... endif`
3. Determine if we want an iwyu builtin target.

alex-tee:
> So I think there are 3 things left for this PR:
> 
>     1. Change meson.project_build_root() to meson.global_build_root()
> 
>     2. Potentially wrap the run_target() snippent is an `if not meson.is_subproject() ... endif`
> 
>     3. Determine if we want an iwyu builtin target.

done 1 and 2

docs/markdown/FAQ.md
```diff
@@ -683,3 +683,23 @@ executable(
...
+```console
+$ iwyu_tool.py -p build
+```
+
+See `iwyu_tool.py --help` for more options.
+
+A run target can also be added with the following meson code:
+```meson
+if not meson.is_subproject()
+iwyu_tool = find_program('iwyu_tool.py', 'iwyu_tool', 'iwyu-tool')
```
tristan957, line 702: Indent these two lines within the if statement

tristan957:
Thanks for your work on this so far :).

PR #10275, opened 2022-04-13
gnome: Make sure g-ir-scanner can use pkg-config properly

xclaesse:
We need to setup the environment we pass to g-ir-scanner because it will
try to use pkg-config to find dependencies, and that must respect user
settings from machine file. Also make it use uninstalled pc files Meson
generated in the case dependencies, such as glib, have been built as
subproject.

### DIFF
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -550,7 +550,7 @@ class EnvironmentVariables(HoldableObject):
         curr = env.get(name)
         return separator.join(values if curr is None else values + [curr])
 
-    def get_env(self, full_env: T.Dict[str, str]) -> T.Dict[str, str]:
+    def get_env(self, full_env: T.MutableMapping[str, str]) -> T.Dict[str, str]:
         env = full_env.copy()
         for method, name, values, separator in self.envvars:
             env[name] = method(env, name, values, separator)
--- a/mesonbuild/dependencies/pkgconfig.py
+++ b/mesonbuild/dependencies/pkgconfig.py
@@ -13,6 +13,8 @@
 # limitations under the License.
 from __future__ import annotations
 
+from pathlib import Path
+
 from .base import ExternalDependency, DependencyException, sort_libpaths, DependencyTypeName
 from ..mesonlib import OptionKey, OrderedSet, PerMachine, Popen_safe
 from ..programs import find_external_program, ExternalProgram
@@ -27,6 +29,7 @@ if T.TYPE_CHECKING:
     from ..environment import Environment
     from ..mesonlib import MachineChoice
     from .._typing import ImmutableListProtocol
+    from ..build import EnvironmentVariables
 
 class PkgConfigDependency(ExternalDependency):
     # The class's copy of the pkg-config path. Avoids having to search for it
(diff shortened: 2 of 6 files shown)


nirbheek:
CI failure is relevant:

```
  File "/__w/meson/meson/mesonbuild/modules/gnome.py", line 977, in _make_gir_target
    Path(state.environment.get_build_dir(), 'meson-uninstalled').as_posix())
NameError: name 'Path' is not defined
```

LGTM otherwise.

eli-schwartz:
> Drop test_pkgconfig_duplicate_path_entries because it's not clear what it's trying to test. `111 pkgconfig duplicate path entries` does not exist. `PkgConfigDependency.setup_env({}, env, MachineChoice.HOST, pkg_dir)` has no effect on the option value, so it really just set a value then read it back.

I didn't take a close look at what it does, but this code comes from #8458 and apparently before that PR, the test failed, also you're the one that merged it. :p

jpakkane:
Those Python extension module failure tests are not looking healthy...

eli-schwartz:
```
'utf-8' codec can't decode byte 0x90 in position 2: invalid start byte
Unusable script 'D:\\a\\_temp\\msys64\\mingw64\\bin\\python3.exe'
```

What???

jpakkane:
Maybe it is running `python3 python3 script.py`?

xclaesse:
Was hoping for a glitch and pushed a rebase, but still same issue, so that's definitely related to this PR somehow.

Note that I did try to change PKG_CONFIG_PATH to use `;` separator on Windows a long time ago and got weird CI issue too, had gave up. So it's not entirely surprising, but we need to find the root cause once and for all.

Something definitely fishy in those logs, somehow `D:` became `D;`.
```
env[PKG_CONFIG_PATH]: D;\a\_temp\msys64\mingw64\lib\pkgconfig;D;\a\_temp\msys64\mingw64\share\pkgconfig
```

xclaesse:
Hmm, gave a try to msys2 on my machine I get that in the env by default:
```
$ echo $PKG_CONFIG_PATH
/usr/lib/pkgconfig:/usr/share/pkgconfig:/lib/pkgconfig
```

So they use unix path separator, interesting... Here we split on `:` which probably cause it to split `D:/foo` -> `['D', '/foo']`: https://github.com/mesonbuild/meson/blob/master/mesonbuild/environment.py#L656.

xclaesse:
@lazka Thanks, that helps a lot. I did not know msys2 rewrite env like that, and meson was splitting PKG_CONFIG_PATH on `:` instead of using `os.pathsep` which caused to split the drive as if it was a separate path.

Added a commit to fix that as well, and that seems to pass CI now.

xclaesse:
@nirbheek @eli-schwartz This is ready for review now. However I'm not sure it's safe for 0.62.x.

xclaesse:
@eli-schwartz squashed the commit as you suggested on IRC. Good to go now?

xclaesse:
Can someone do a review please?

jpakkane:
LGTM.

xclaesse:
Thanks

ylatuya:
@xclaesse There is one last bit missing for that to work correctly with a pkg-config binary that is not in the default search paths (eg: pkg-config.exe provided through a wrap file on Windows). For g-ir-scanner to use the pkg-config binary found we must also export PKG_CONFIG with the path to the binary: https://gitlab.gnome.org/GNOME/gobject-introspection/-/blob/main/giscanner/pkgconfig.py#L30

xclaesse:
@ylatuya good point could you give that a try please? https://github.com/mesonbuild/meson/pull/11007

PR #10276, opened 2022-04-13
vcs_tag: properly handle non-str commands

hg:
Hi,

Currently `vcs_tag` does not accept anything but string as its first `command` argument, which forces you to write things like this:

https://github.com/gsliepen/tinc/blob/1.1/meson.build#L38-L40

https://github.com/gsliepen/tinc/blob/1.1/src/include/meson.build#L3-L4

Docs were updated to match annotated types on top of the function. Not sure how you're supposed to call `vcs_tag` with `custom_target`, so there's no test for it.

Could it be an error in type annotation, or is there something obvious I'm not seeing?

### DIFF
--- a/docs/yaml/functions/vcs_tag.yaml
+++ b/docs/yaml/functions/vcs_tag.yaml
@@ -22,13 +22,15 @@ description: |
 
 kwargs:
   command:
-    type: list[str]
+    type: list[str | file]
     description: |
       The command to execute, see [[custom_target]] for details
       on how this command must be specified.
 
       This parameter is optional. If it is absent, Meson will try
       its best to find a suitable default command.
+
+      *(since 0.62.0)* [[@file]] is accepted.
 
   input:
     type: str
--- a/docs/yaml/functions/vcs_tag.yaml
+++ b/docs/yaml/functions/vcs_tag.yaml
@@ -22,7 +22,7 @@ description: |
 
 kwargs:
   command:
-    type: list[str | file]
+    type: list[exe | external_program | custom_tgt | file | str]
     description: |
       The command to execute, see [[custom_target]] for details
       on how this command must be specified.
@@ -31,6 +31,8 @@ kwargs:
       its best to find a suitable default command.
 
       *(since 0.62.0)* [[@file]] is accepted.
+
+      *(since 0.63.0)* [[@custom_tgt]], [[@exe]], and [[@external_program]] are accepted.
 
   input:
     type: str
--- a/mesonbuild/interpreter/interpreter.py
+++ b/mesonbuild/interpreter/interpreter.py
@@ -1752,11 +1752,14 @@ external dependencies (including libraries) must go to "dependencies".''')
         vcs_cmd = kwargs['command']
         source_dir = os.path.normpath(os.path.join(self.environment.get_source_dir(), self.subdir))
         if vcs_cmd:
-            if isinstance(vcs_cmd[0], mesonlib.File):
-                FeatureNew.single_use('vcs_tag with file as the first argument', '0.62.0', self.subproject, location=node)
-            maincmd = self.find_program_impl(vcs_cmd[0], required=False)
-            if maincmd.found():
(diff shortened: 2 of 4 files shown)


eli-schwartz:
> Docs were updated to match annotated types on top of the function. Not sure how you're supposed to call `vcs_tag` with `custom_target`, so there's no test for it.
> 
> Could it be an error in type annotation, or is there something obvious I'm not seeing?

Could what be an error in type annotation? What is or is not calling what with custom_target?

docs/yaml/functions/vcs_tag.yaml
```diff
@@ -22,7 +22,7 @@ description: |
 
 kwargs:
   command:
-    type: list[str]
+    type: list[exe | external_program | custom_tgt | file | str]
```
eli-schwartz, line 25: The `description` field must specify that e.g.

`*(since 0.63.0)* [[@custom_tgt]], [[@exe]], [[@external_program]], and [[@file]] are accepted.`

mesonbuild/interpreter/interpreter.py
```diff
@@ -1752,9 +1752,10 @@ def func_vcs_tag(self, node: mparser.BaseNode, args: T.List['TYPE_var'], kwargs:
         vcs_cmd = kwargs['command']
         source_dir = os.path.normpath(os.path.join(self.environment.get_source_dir(), self.subdir))
         if vcs_cmd:
-            maincmd = self.find_program_impl(vcs_cmd[0], required=False)
-            if maincmd.found():
-                vcs_cmd[0] = maincmd
+            if isinstance(vcs_cmd[0], (str, mesonlib.File)):
```
eli-schwartz, line 1755: This should emit a `FeatureNew.single_use()` if it isn't a string or File.

Also it turns out that we "accidentally" started accepting a File in commit b402817fb6f0392812bfa272bdbc05c9c30139fa (/cc @dcbaker) so that should emit a FeatureNew since 0.62, I guess.

eli-schwartz:
BTW:

```
  command: [python_path, src_root / 'version.py'],
```

could just be `command: [src_root / 'version.py']` as meson will then internally detect a shebang referring to python3, and use its own `sys.executable` as a python implementation in order to guarantee that it can be run. Alternatively, for generic backwards compatibility, if you use `command: ['python3', ...]` it is identical to `command: [find_program('python3'), ...]`, it will just internally run find_program() again. It's a lot more interesting to use built executables here as that's downright impossible right now.

hg:
Thanks for the review. Sorry for such basic omissions, meson insides are pretty overwhelming at first because of how much stuff it's doing for you.

> It's a lot more interesting to use built executables here as that's downright impossible right now.

This seems to work, although it's not pretty:

```meson
exe = executable('tagprog', 'tagprog.c', version_src)

tag = vcs_tag(
    input : 'vcstag.c.in',
    output : 'exe.c',
    command : [exe.full_path()]
)

executable('tagprog-exe', 'tagprog.c', tag)
```

Older versions work fine if `exe`/`file`/etc are used as a second argument onwards, so we're only checking the first one.

eli-schwartz:
They're not basic omissions :) my review suggestions were not about the functionality, just about making sure the functionality communicates the API level at which it can be used.

As far as `exe.full_path()` goes, that will not guarantee the exe is built first so it will unpredictably fail depending on the random order in which ninja schedules ready nodes on the build graph.

eli-schwartz:
Functionality looks good. Can I bother you to split it into two commits, though? One bugfix to document and emit FeatureNew for file objects, and a second commit to implement the new feature.

This also allows us to potentially backport the first part into 0.62.1

hg:
Sure. I pushed the first commit [here][file] to run CI checks on it.

[file]: https://github.com/hg/meson/actions?query=branch%3Avcs_tag_file

eli-schwartz:
@nirbheek the first commit should probably be backported to 0.62.1 as it only adds a FeatureNew in a palace that was overlooked.

Issue #10269, opened 2022-04-11
[Proposal] relocatable pkgconfig files


plattfot:
# Problem

At work we install our software a bit differently than a standard linux distribution. Closest to our system would be [rez](https://github.com/nerdvegas/rez). Due to that, we need to have our pkgconfig files generated by meson to be relocatable. Which right now is not possible. As a workaround we run a `sed` command after `meson install` to fix it. E.g.

```sh
sed -i -E "s|$prefix|\${pcfiledir}/../..|" $prefix/lib/pkgconfig/*.pc
```
Where `prefix` is the path given to `--prefix` when running `meson setup`.

# Proof of concept
The extra step has annoyed me for a while now so thought I might try and see how hard it would be to add that to `meson`. Here is what I came up with:

```diff
diff -urB a/mesonbuild/modules/pkgconfig.py b/mesonbuild/modules/pkgconfig.py
--- a/mesonbuild/modules/pkgconfig.py	2022-03-15 02:28:57.000000000 -0700
+++ b/mesonbuild/modules/pkgconfig.py	2022-04-11 12:32:40.430481547 -0700
@@ -328,7 +328,8 @@
 
     def _generate_pkgconfig_file(self, state, deps, subdirs, name, description,
                                  url, version, pcfile, conflicts, variables,
-                                 unescaped_variables, uninstalled=False, dataonly=False):
+                                 unescaped_variables, uninstalled=False, dataonly=False,
+                                 pkgroot=None):
         coredata = state.environment.get_coredata()
         referenced_vars = set()
         optnames = [x.name for x in BUILTIN_DIR_OPTIONS.keys()]
@@ -373,7 +374,12 @@
             srcdir = PurePath(state.environment.get_source_dir())
         else:
             outdir = state.environment.scratch_dir
-            prefix = PurePath(coredata.get_option(mesonlib.OptionKey('prefix')))
+            if pkgroot:
+                prefix = PurePath('${pcfiledir}',
+                                  *('..' for _ in PurePath(pkgroot).parts))
+                mlog.debug(f"using relative path {prefix} for prefix")
+            else:
+                prefix = PurePath(coredata.get_option(mesonlib.OptionKey('prefix')))
         fname = os.path.join(outdir, pcfile)
         with open(fname, 'w', encoding='utf-8') as ofile:
             for optname in optnames:
@@ -484,6 +490,7 @@
             if cflags and not dataonly:
[...]

PR #10277, opened 2022-04-13
Relocatable pkgconfig

plattfot:
Add `relocatable` module option for the `pkgconfig` module. When set to `true` the `prefix` of the generated pkgconfig files will be relative to their `install_dir`. 

By default this is set to `false`.

For more detail see the discussion in  #10269.

### DIFF
--- a/mesonbuild/modules/pkgconfig.py
+++ b/mesonbuild/modules/pkgconfig.py
@@ -328,7 +328,8 @@ class PkgConfigModule(ExtensionModule):
 
     def _generate_pkgconfig_file(self, state, deps, subdirs, name, description,
                                  url, version, pcfile, conflicts, variables,
-                                 unescaped_variables, uninstalled=False, dataonly=False):
+                                 unescaped_variables, uninstalled=False, dataonly=False,
+                                 pkgroot=None):
         coredata = state.environment.get_coredata()
         referenced_vars = set()
         optnames = [x.name for x in BUILTIN_DIR_OPTIONS.keys()]
@@ -374,6 +375,15 @@ class PkgConfigModule(ExtensionModule):
         else:
             outdir = state.environment.scratch_dir
             prefix = PurePath(coredata.get_option(mesonlib.OptionKey('prefix')))
+            if pkgroot:
+                pkgroot = PurePath(pkgroot)
+                if not pkgroot.is_absolute():
+                    pkgroot = prefix / pkgroot
+                elif prefix not in pkgroot.parents:
+                    raise mesonlib.MesonException('Pkgconfig prefix cannot be outside of the prefix '
+                                                  'when pkgconfig.relocatable=true. '
+                                                  f'Pkgconfig prefix is {pkgroot.as_posix()}.')
+                prefix = PurePath('${pcfiledir}', os.path.relpath(prefix, pkgroot))
         fname = os.path.join(outdir, pcfile)
         with open(fname, 'w', encoding='utf-8') as ofile:
             for optname in optnames:
@@ -591,9 +601,11 @@ class PkgConfigModule(ExtensionModule):
                 pkgroot_name = os.path.join('{libdir}', 'pkgconfig')
         if not isinstance(pkgroot, str):
             raise mesonlib.MesonException('Install_dir must be a string.')
+        relocatable = state.get_option('relocatable', module='pkgconfig')
         self._generate_pkgconfig_file(state, deps, subdirs, name, description, url,
                                       version, pcfile, conflicts, variables,
-                                      unescaped_variables, False, dataonly)
+                                      unescaped_variables, False, dataonly,
+                                      pkgroot=pkgroot if relocatable else None)
         res = build.Data([mesonlib.File(True, state.environment.get_scratch_dir(), pcfile)], pkgroot, pkgroot_name, None, state.subproject, install_tag='devel')
         variables = self.interpreter.extract_variables(kwargs, argname='uninstalled_variables', dict_new=True)
         variables = parse_variable_list(variables)
--- a/test cases/failing/123 pkgconfig not relocatable outside prefix/meson.build
+++ b/test cases/failing/123 pkgconfig not relocatable outside prefix/meson.build
@@ -1,0 +1,22 @@
+project(
+  'pkgconfig-not-relocatable-outside-prefix',
+  version : '1.0',
+  default_options: [
+    'pkgconfig.relocatable=true',
+  ])
+
+pkgg = import('pkgconfig')
+
+# A drive letter is needed on windows for this to be an absolute path.
+if host_machine.system() == 'windows'
+  install_dir = 'C:/opt/lib/pkgconfig'
+else
+  install_dir = '/opt/lib/pkgconfig'
+endif
+
+pkgg.generate(
+  name : 'libsimple',
+  version : '1.0',
+  description : 'A simple pkgconfig file.',
+  install_dir: install_dir,
+)
--- a/test cases/unit/105 pkgconfig relocatable with absolute path/meson.build
+++ b/test cases/unit/105 pkgconfig relocatable with absolute path/meson.build
@@ -1,0 +1,15 @@
+project(
+  'pkgconfig-relocatable-with-absolute-path',
+  version : '1.0',
+  default_options: [
+    'pkgconfig.relocatable=true',
+  ])
+
+pkgg = import('pkgconfig')
+
+pkgg.generate(
+  name : 'libsimple',
+  version : '1.0',
+  description : 'A simple pkgconfig.',
+  install_dir: get_option('prefix')/'share/misc/pkgconfig',
+)
--- a/docs/markdown/Builtin-options.md
+++ b/docs/markdown/Builtin-options.md
@@ -275,6 +275,27 @@ Since 0.56.0 `warning_level` can also be defined per subproject.
 Some Meson modules have built-in options. They can be set by prefixing the option
 name with the module name: `-D<module>.<option>=<value>` (e.g. `-Dpython.platlibdir=/foo`).
 
+### Pkgconfig module
+
+| Option      | Default value | Possible values | Description                                                |
+|-------------|---------------|-----------------|------------------------------------------------------------|
+| relocatable | false         | true, false     | Generate the pkgconfig files as relocatable (Since 0.63.0) |
+
+*Since 0.63.0* The `pkgconfig.relocatable` option is used by the
+pkgconfig module, namely [`pkg.generate()`](Pkgconfig-module.md) and affect how the
+`prefix` in the generated pkgconfig file is set (not to be confused
+with the [install prefix](#directories)). When it is `true` the `prefix` will be
+relative to the `install_dir`. This allows the pkgconfig file to be
+moved around and still work, as long as the relative path is not
+broken. In general this allows for the whole installed package to be
+placed anywhere on the system and still work as a dependency. When it
+is set to `false` the `prefix` will be the same as the install prefix.
+
+An error will be raised if `pkgconfig.relocatable` is `true` and the
+`install_dir` for a generated pkgconfig file points outside the
+install prefix. For example if the install prefix is `/usr` and the
+`install_dir` for a pkgconfig file is `/var/lib/pkgconfig`.
+
 ### Python module
 
 | Option           | Default value | Possible values             | Description |
--- a/docs/markdown/snippets/pkgconfig-relocatable.md
+++ b/docs/markdown/snippets/pkgconfig-relocatable.md
@@ -1,0 +1,18 @@
+## Installed pkgconfig files can now be relocatable
+
+The pkgconfig module now has a module option `pkgconfig.relocatable`.
+When set to `true`, the pkgconfig files generated will have their
+`prefix` variable set to be relative to their `install_dir`.
+
+For example to enable it from the command line run:
+
+```sh
+meson setup builddir -Dpkgconfig.relocatable=true …
+```
+
+It will only work if the `install_dir` for the generated pkgconfig
+files are located inside the install prefix of the package. Not doing
+so will cause an error.
+
+This should be useful on Windows or any other platform where
+relocatable packages are desired.
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -1225,6 +1225,10 @@ BUILTIN_CORE_OPTIONS: 'MutableKeyedOptionDictType' = OrderedDict([
     (OptionKey('wrap_mode'),       BuiltinOption(UserComboOption, 'Wrap mode', 'default', choices=['default', 'nofallback', 'nodownload', 'forcefallback', 'nopromote'])),
     (OptionKey('force_fallback_for'), BuiltinOption(UserArrayOption, 'Force fallback for those subprojects', [])),
 
+    # Pkgconfig module
+    (OptionKey('relocatable', module='pkgconfig'),
+     BuiltinOption(UserBooleanOption, 'Generate pkgconfig files as relocatable', False)),
+
     # Python module
     (OptionKey('install_env', module='python'),
      BuiltinOption(UserComboOption, 'Which python environment to install to', 'prefix', choices=['auto', 'prefix', 'system', 'venv'])),
--- a/test cases/failing/123 pkgconfig not relocatable outside prefix/test.json
+++ b/test cases/failing/123 pkgconfig not relocatable outside prefix/test.json
@@ -1,0 +1,8 @@
+{
+  "stdout": [
+    {
+      "line": "test cases/failing/123 pkgconfig not relocatable outside prefix/meson\\.build:17:5: ERROR: Pkgconfig prefix cannot be outside of the prefix when pkgconfig\\.relocatable=true. Pkgconfig prefix is (C:)?/opt/lib/pkgconfig.",
+      "match": "re"
+    }
+  ]
+}
--- a/unittests/allplatformstests.py
+++ b/unittests/allplatformstests.py
@@ -1636,6 +1636,39 @@ class AllPlatformTests(BasePlatformTests):
         # pkg-config and pkgconf does not respect the same order
         self.assertEqual(sorted(foo_dep.get_compile_args()), sorted(cargs))
 
+    @skipIfNoPkgconfig
+    def test_pkgconfig_relocatable(self):
+        '''
+        Test that it generates relocatable pkgconfig when module
+        option pkgconfig.relocatable=true.
+        '''
+        testdir_rel = os.path.join(self.common_test_dir, '44 pkgconfig-gen')
+        self.init(testdir_rel, extra_args=['-Dpkgconfig.relocatable=true'])
+
+        def check_pcfile(name, *, relocatable, levels=2):
+            with open(os.path.join(self.privatedir, name), encoding='utf-8') as f:
+                pcfile = f.read()
+                # The pkgconfig module always uses posix path regardless of platform
+                prefix_rel = PurePath('${pcfiledir}', *(['..'] * levels)).as_posix()
+                (self.assertIn if relocatable else self.assertNotIn)(
+                    f'prefix={prefix_rel}\n',
+                    pcfile)
+
+        check_pcfile('libvartest.pc', relocatable=True)
+        check_pcfile('libvartest2.pc', relocatable=True)
+
+        self.wipe()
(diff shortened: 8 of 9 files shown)


mesonbuild/modules/pkgconfig.py
```diff
@@ -374,6 +375,16 @@ def _generate_pkgconfig_file(self, state, deps, subdirs, name, description,
...
+            if pkgroot:
+                pkgroot = PurePath(pkgroot)
+                if pkgroot.is_absolute():
+                    try:
+                        pkgroot = pkgroot.relative_to(prefix)
+                    except ValueError:
+                        raise mesonlib.MesonException('Pkgconfig prefix cannot be outside of the prefix '
+                                                      'when pkgconfig.relocatable=true. '
+                                                      f'Pkgconfig prefix is {pkgroot}.')
+                prefix = PurePath('${pcfiledir}', *('..' for _ in pkgroot.parts))
```
eli-schwartz, line 387:
```python
if not pkgroot.is_absolute():
    pkgroot = prefix / pkgroot
elif prefix not in pkgroot.parents:
    raise mesonlib.MesonException('...')
prefix = PurePath('${pcfiledir}', os.path.relpath(prefix, pkgroot))
```

The use of one `'..'` per part in pkgroot feels wrong to me, but PurePath().relative_to() is broken by design as it doesn't allow you to compute the relative position of two absolute paths. This is one more example of how pathlib's best feature is introducing an os.PathLike interface, but their actual implementation decisions are... often bad.

Fortunately, the os.path.* family of methods still work fine, and they accept PathLike objects. In this case, `os.path.relpath` > `PurePath.relative_to`.

(Don't let me get started on how utterly catastrophic the entire set of concrete Path() methods are, with their insistence that .exists() should raise a traceback if a path cannot exist, or doesn't seem to exist, due to stat errors... but it's totally fine to check if an invalid pathname exists, that just immediately returns `False` lolwut.)

eli-schwartz (changes requested):
This generally looks okay to me, but it needs documentation and release notes, and particularly the testsuite fails in `DataTests::test_builtin_options_documented`.

unittests/linuxliketests.py
```diff
@@ -256,6 +256,36 @@ def test_pkgconfig_gen_deps(self):
...
+        option pkgconfig.relocatable=true.
+        '''
+        testdir_rel = os.path.join(self.common_test_dir, '44 pkgconfig-gen')
+        self.init(testdir_rel, extra_args=['-Dpkgconfig.relocatable=true'])
+
+        def check_pcfile(name, *, relocatable, levels=2):
+            with open(os.path.join(self.privatedir, name), encoding='utf-8') as f:
+                pcfile = f.read()
+                (self.assertIn if relocatable else self.assertNotIn)(
+                    f'prefix={os.path.join("${pcfiledir}", *(".." for _ in range(levels)))}\n',
```
eli-schwartz, line 271:
```suggestion
                    f'prefix={os.path.join("${pcfiledir}", *([".."] *levels))}\n',
```
is much faster, per timeit.

Aside: I don't like the use of extremely complex expressions inside f-strings as it is extremely hard to read.

unittests/linuxliketests.py
```diff
@@ -256,6 +256,36 @@ def test_pkgconfig_gen_deps(self):
...
+        check_pcfile('libvartest2.pc', relocatable=True)
+
+        self.wipe()
+        self.init(testdir_rel, extra_args=['-Dpkgconfig.relocatable=false'])
+
+        check_pcfile('libvartest.pc', relocatable=False)
+        check_pcfile('libvartest2.pc', relocatable=False)
+
+        self.wipe()
+        testdir_abs = os.path.join(self.common_test_dir, '251 pkgconfig relocatable with absolute path')
```
eli-schwartz, line 284: Is there any reason to test this as a project test *and* as a unittest? That will run it once and simply check that it successfully configures and builds, and run it again inside unittest in order to check the contents of files.

unittests/linuxliketests.py
```diff
@@ -256,6 +256,36 @@ def test_pkgconfig_gen_deps(self):
             content = f.read()
             self.assertNotIn('-lstat2', content)
 
+    def test_pkgconfig_relocatable(self):
```
eli-schwartz, line 259: Why is this only in linuxliketests? Do we not want to check that it works on Windows?

mesonbuild/modules/pkgconfig.py
```diff
@@ -591,9 +602,11 @@ def parse_variable_list(vardict):
                 pkgroot_name = os.path.join('{libdir}', 'pkgconfig')
         if not isinstance(pkgroot, str):
             raise mesonlib.MesonException('Install_dir must be a string.')
+        relocatable = state.environment.coredata.get_option(mesonlib.OptionKey('relocatable', module='pkgconfig'))
```
xclaesse, line 605: There is `state.get_option('relocatable', module='pkgconfig')` shortcut.

eli-schwartz:
Actually when I originally mentioned it is failing the datatests, I overlooked that you did add docs for this, but... I thought I fixed this, apparently not... in #9669 I unbroke the datatests so that it is capable of parsing multiple module subsections.

unittests/linuxliketests.py
plattfot: Sorry, I was a bit confused with how testing is structured in meson. Realized now that I should move this to `test cases/unit`.

unittests/linuxliketests.py
plattfot: That's where I found other tests for pkgconfig, so I just assumed meson did not use pkgconfig on windows (I haven't used a windows box in years). Should I move this to common? Or where is a good place for it?

unittests/linuxliketests.py
plattfot: I'll move that expression outside the f-strings.

mesonbuild/modules/pkgconfig.py
plattfot: Yeah, it could be fragile if `pkgroot` for some reason contains `'..'`. The `os.path.relpath` implementation looks a lot cleaner. I'll be switching to that. And I should have listen to @xclaesse.

Thanks for the input.

unittests/linuxliketests.py
eli-schwartz: As a unittest (rather than a project test) it should likely go in `unittests/allplatformstests.py`.

As for whether you can use pkg-config on Windows, you definitely *can*, and some people do. Meson supports it. There's a culture on Windows of refusing to use pkg-config because Windows users think that as tech that originated in Unix land, pkg-config must have cooties or something.

cmake is actually very popular on Windows for exactly that reason -- it both uses its own in-house format + config parser, and supports relocatable installs (much to the dismay of Linux distro packagers! This relocation support actually causes Linux distro bugs if you invoke /bin/cmake instead of /usr/bin/cmake, and as a result the includedir is "relocated" to /include which obviously does not exist).

So cmake users think that cmake supports relocatable installs and pkg-config doesn't. For that matter, I am not much of a Windows user, so I didn't realize pkg-config supported pcfiledir either -- but...

ironically, pcfiledir was *added* to the pkg-config spec for the sake of Windows support, and this is the reason why I felt a PR for this feature is worth it, and particularly, why it should be a module option rather than used exclusively inside one or two projects that expect to be relocatable. Because I suspect that Windows users will want to set this option as a matter of course. :)

plattfot: > As a unittest (rather than a project test) it should likely go in unittests/allplatformstests.py.

Sounds good, I'll put it there.

> As for whether you can use pkg-config on Windows, you definitely can, and some people do. Meson supports it. There's a culture on Windows of refusing to use pkg-config because Windows users think that as tech that originated in Unix land, pkg-config must have cooties or something.…

Thanks for the explanation, I learned something new today. :+1: Both pkg-config on Windows and cmake. Luckily I haven't encountered the `/bin/cmake` vs `/usr/bin/cmake` but a fair deal of cmake based packages trying to vendor all the deps instead of using the deps already installed. I assume I have Windows to thank for that. :smile:

mesonbuild/modules/pkgconfig.py
plattfot: Should be all good now.

test cases/failing/123 pkgconfig not relocatable outside prefix/meson.build
```diff
@@ -0,0 +1,19 @@
+project(
+  'pkgconfig-not-relocatable-outside-prefix', 'c',
+  version : '1.0',
+  default_options: [
+    'pkgconfig.relocatable=true',
+  ])
+
+pkgg = import('pkgconfig')
+lib = shared_library('simple', 'simple.c')
```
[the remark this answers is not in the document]
plattfot: Removed in my latest commit.

test cases/common/251 pkgconfig relocatable with absolute path/meson.build
```diff
@@ -0,0 +1,19 @@
+project(
+  'pkgconfig-relocatable-with-absolute-path', 'c',
+  version : '1.0',
+  default_options: [
+    'pkgconfig.relocatable=true',
+  ])
+
+pkgg = import('pkgconfig')
+lib = shared_library('simple', 'simple.c')
```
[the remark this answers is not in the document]
plattfot: Same for this.

unittests/linuxliketests.py
plattfot: This is now in `test cases/unit/105` and should only be run for the unittest `AllPlatformTests.test_pkgconfig_relocatable`.

unittests/linuxliketests.py
plattfot: In the latest commit I moved the expression outside the f-string and switched to `['..'] * levels`. Hopefully it's a bit more readable now.

mesonbuild/modules/pkgconfig.py
plattfot: Latest commit I changed to the shortcut.

test cases/failing/123 pkgconfig not relocatable outside prefix/meson.build
```diff
@@ -0,0 +1,15 @@
+project(
+  'pkgconfig-not-relocatable-outside-prefix', 'c',
```
eli-schwartz, line 2: As we aren't building anything anymore, we can avoid declaring that we need the C language, and skip doing a compiler lookup/sanity check.

test cases/unit/105 pkgconfig relocatable with absolute path/meson.build
```diff
@@ -0,0 +1,15 @@
+project(
+  'pkgconfig-relocatable-with-absolute-path', 'c',
```
eli-schwartz, line 2: Likewise.

test cases/failing/123 pkgconfig not relocatable outside prefix/meson.build
plattfot: Removed in the latest push.

test cases/unit/105 pkgconfig relocatable with absolute path/meson.build
plattfot: Same for this one.

plattfot:
> Actually when I originally mentioned it is failing the datatests, I overlooked that you did add docs for this, but... I thought I fixed this, apparently not... in #9669 I unbroke the datatests so that it is capable of parsing multiple module subsections. Apparently it just never got merged...

How should I deal with the CI that are failing? Is this due to my changes or #9669? I haven't figured out how to interpret the logs on Azure…

eli-schwartz:
> How should I deal with the CI that are failing? Is this due to my changes or #9669? I haven't figured out how to interpret the logs on Azure…

I just merged a PR that contained that change standalone. Please:
- rebase on git master
- in `unittests/datatests.py`, change `for idx, mod in enumerate(['Python']):` to `['Python', 'Pkgconfig']`

You may also consider sorting the sections to be alphabetical (both in this list and in `docs/markdown/Builtin-options.md`).

eli-schwartz:
The Azure logging frankly sucks, you need to scroll up past the codecov.

plattfot:
> > How should I deal with the CI that are failing? Is this due to my changes or #9669? I haven't figured out how to interpret the logs on Azure…
> 
> I just merged a PR that contained that change standalone. Please:
> 
>     * rebase on git master
> 
>     * in `unittests/datatests.py`, change `for idx, mod in enumerate(['Python']):` to `['Python', 'Pkgconfig']`
> 
> 
> You may also consider sorting the sections to be alphabetical (both in this list and in `docs/markdown/Builtin-options.md`).

Thanks, I changed so the modules are now in alphabetical order and rebase on master.

> The Azure logging frankly sucks, you need to scroll up past the codecov.

Ah, yeah the codecov really confused me. Thanks, now I can see that my failing test `123 pkgconfig not relocatable outside prefix` is failing. Is there a way to get the traceback of that test? Right now all I see is:

```
[ERROR]    failing: 123 pkgconfig not relocatable outside prefix
During: configure
Reason: Unexpected output while configuring.
```

I'm guessing it has something to do with the `/opt/lib/pkgconfig` path I'm using for the `install_dir`. But I'm not sure if the output is different so the `test.json` that's way it's failing or it is the path itself. I don't have any windows machine to test this on.

eli-schwartz:
The logs contain this at the bottom of the testsuite run:

```
======================= failing: 123 pkgconfig not relocatable outside prefix ======================
 

Build started at 2022-04-14T23:59:23.811597
 

Build started at 2022-04-14T23:59:23.811597
Main binary: C:\hostedtoolcache\windows\Python\3.7.9\x64\python.exe
Build Options: -Dprefix=x:/usr -Dlibdir=lib -Dbackend=ninja
Python system: Windows
The Meson build system
Project name: pkgconfig-not-relocatable-outside-prefix
Project version: 1.0
Build machine cpu: x86_64
Host machine cpu: x86_64
Target machine cpu family: x86_64
Target machine cpu: x86_64
Build targets in project: 0
WARNING: Project specifies a minimum meson_version '' but uses features which were added in newer versions:
 * 0.49.0: {'/ with string arguments'}
 * 0.60.0: {'list.<plus>'}
NOTICE: Future-deprecated features used:
 * 0.56.0: {'meson.source_root'}

pkgconfig-not-relocatable-outside-prefix 1.0

  User defined options
    backend: ninja
    libdir : lib
    prefix : x:/usr

Found ninja.EXE-1.10.0 at D:\a\ci_data\ninja.EXE

!meson_ci!/ci_include "D:\a\1\s\b b7642b3b40\build.ninja"
expected "test cases\failing\123 pkgconfig not relocatable outside prefix\meson.build:10:5: ERROR: Pkgconfig prefix cannot be outside of the prefix when pkgconfig.relocatable=true. Pkgconfig prefix is /opt/lib/pkgconfig." not found in stdout 
```

eli-schwartz:
Windows does not consider `PureWindowsPath('/opt/lib/pkgconfig')` to be an absolute path. It has no drive letter...

plattfot:
Updated the `test.json` for `test case/failing/123` to use regex instead to be able to handle the differences in output between Windows and the rest.

mesonbuild/modules/pkgconfig.py
```diff
@@ -374,6 +375,15 @@ def _generate_pkgconfig_file(self, state, deps, subdirs, name, description,
...
             outdir = state.environment.scratch_dir
             prefix = PurePath(coredata.get_option(mesonlib.OptionKey('prefix')))
+            if pkgroot:
+                pkgroot = PurePath(pkgroot)
+                if not pkgroot.is_absolute():
+                    pkgroot = prefix / pkgroot
+                elif prefix not in pkgroot.parents:
+                    raise mesonlib.MesonException('Pkgconfig prefix cannot be outside of the prefix '
+                                                  'when pkgconfig.relocatable=true. '
+                                                  f'Pkgconfig prefix is {pkgroot}.')
```
eli-schwartz, line 385: This currently prints an error in CI pointing at either:
```
C:\opt\lib\pkgconfig
```
or
```
/opt/lib/pkgconfig
```

This is challenging to detect in a regex match in test.json, because you need to change lots of things. You can simplify it quite a bit by printing out an error message of the form 
```suggestion
                                                  f'Pkgconfig prefix is {pkgroot.as_posix()}.')
```

plattfot: Thanks! I was going a bit insane trying to escape `\` for the path in the `test.json`. Changed it to transform the path into posix in the error message.

Hopefully if will go through :crossed_fingers:

plattfot: Nope :disappointed:, looks like I somehow need to deal with the drive letter. I'll see if I get some time working on it the following days otherwise I'll continue on Monday.

eli-schwartz: Yes, your previous regex approach seems reasonable there.

What as_posix() does is ensures you don't need to *also* handle the difference between \ and / and particularly getting the former to be correctly json-escaped. Handling `C:` via regex is comparatively trivial.

plattfot: Ah, ok. I thought there was more than just the `\` and `/` when using the regex. Rolling back to using that with the `as_posix()`. We'll see how that goes.

plattfot:
Question, what path syntax does pkgconfig use on windows? I'm seeing it failing on `msys2/gccx86ninja` with the pkgconfig prefix using `/` but the unit test expects `\`. Wondering which one I should fix.

eli-schwartz:
The unittest uses os.path.join to define the path it is looking for in the .pc file, so it uses \ rather than matching the as_posix() version that got written out. You could use `'/'.join()` instead, I suppose.

plattfot:
Is pkg-config on Windows expecting the `prefix` path in the .pc file to be posix?

plattfot:
Never mind, I found my answer in [mesonbuild/modules/pkgconfig.py](https://github.com/mesonbuild/meson/blob/ab5ed4db09331d8d182bcdda4a44fa88b9265acb/mesonbuild/modules/pkgconfig.py#L312). It always uses posix. I'll update the unittest to reflect that.

eli-schwartz:
More specifically, it uses PurePath with the as_posix() function I mentioned above in the discussion around https://github.com/mesonbuild/meson/pull/10277#discussion_r851050102, which has different semantics from PurePosixPath. Specifically, it treats \ as a path separator and parses a leading C: as a drive letter. It's not really relevant to the extremely limited use case here, but for correctness I guess we ought to use `PurePath.as_posix()` anyway.

plattfot:
Oh, I thought `PurePath.as_posix()` just returned a `PurePosixPath`. But looking at the docs, I don't see it any mention about that. *sigh* I'll change it to call `PurePath.as_posix()`.

plattfot:
> Oh, I thought `PurePath.as_posix()` just returned a `PurePosixPath`. But looking at the docs, I don't see it any mention about that. _sigh_ I'll change it to call `PurePath.as_posix()`.

Done

plattfot:
Done

eli-schwartz:
Nope, it's an alternative `str()` conversion. :)

eli-schwartz:
Can you fold up the last 3 doc commits into one? Particularly commit 6 fixes the tests which commit 4 breaks.

plattfot:
> Nope, it's an alternative `str()` conversion. :)

Haha, ok. That module seems to be full of surprises.

eli-schwartz:
Tell me about it!

The Path() methods as opposed to the PurePath() methods, that is to say anything that does filesystem I/O, have this highly nasty habit of being exactly identical to `os.path.*` functions except that they will randomly raise tracebacks under conditions such as "PermissionError for .exists() but it doesn't exist because the current user lacks permission to access that path", or "Incorrect Function for .resolve() but paths that don't exist will resolve correctly, while paths that use certain types of Windows shares that return certain types of stat failures for unmounted/inaccessible locations".

plattfot:
> Tell me about it!
> 
> The Path() methods as opposed to the PurePath() methods, that is to say anything that does filesystem I/O, have this highly nasty habit of being exactly identical to `os.path.*` functions except that they will randomly raise tracebacks under conditions such as "PermissionError for .exists() but it doesn't exist because the current user lacks permission to access that path (but you don't try accessing it, merely checking if it's accessible)", or "Incorrect Function for .resolve() but paths that don't exist will resolve correctly, while paths that use certain types of Windows shares that return certain types of stat failures for unmounted/inaccessible locations are just incurably broken".

Yikes… seems there are a lot of foot guns in that module. I still can't believe the `relative_to` function is broken by design.

PR #10278, opened 2022-04-14
Misc fixes

eli-schwartz:
Just a boatload of random fixes that don't fit into any theme, which I've been carrying locally.

### DIFF
--- a/.github/workflows/images.yml
+++ b/.github/workflows/images.yml
@@ -25,6 +25,8 @@ on:
 
 jobs:
   build:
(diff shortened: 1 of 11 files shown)


xclaesse:
Not sure if CI failure is related. If not, LGTM.

eli-schwartz:
> Not sure if CI failure is related. If not, LGTM.

Those were the CI image builder checks, which are known to fail due to docker issues. They just trigger on any updates to the CI image builder file, even though in this case all the change does is modify how the schedule runs.

rebased, merged

PR #10281, opened 2022-04-15
gtkdochelper.py: Use os.pathsep for --path argument

fanc999:
Hi,

When attempting to build GLib's documentation using Meson, the build failed as the `--path` argument for `gtkdoc-mkhtml` passed paths that were separated by `:`, which is not valid on Windows `cmd.exe` consoles.  Fix this by using `os.pathsep` instead.

With blessings, thank you!

### DIFF
--- a/mesonbuild/scripts/gtkdochelper.py
+++ b/mesonbuild/scripts/gtkdochelper.py
@@ -198,7 +198,7 @@ def build_gtkdoc(source_root: str, build_root: str, doc_subdir: str, src_subdirs
 
     # Make HTML documentation
     mkhtml_cmd = [options.gtkdoc_mkhtml,
-                  '--path=' + ':'.join((doc_src, abs_out)),
+                  '--path=' + os.pathsep.join((doc_src, abs_out)),
                   module,
                   ] + html_args
     if main_file:


eli-schwartz:
AFAICT gtkdoc-mkhtml wants to see `--path foo --path bar`, hmm.

fanc999:
Hi @eli-schwartz,

Unfortunately, no, using `--path=foo --path=bar` didn't work, but using `-path=foo;bar` on a cmd.exe command prompt did.

With blessings, and cheers!

eli-schwartz:
I must confess, I'm incredibly curious what gtkdoc-mkhtml is doing there!

xclaesse:
> I must confess, I'm incredibly curious what gtkdoc-mkhtml is doing there!

You don't want to know, trust me. But if you insist: https://gitlab.gnome.org/GNOME/gtk-doc/-/blob/master/gtkdoc/mkhtml.py#L34.

xclaesse:
LGTM, ideally it should get fixed in gtkdoc, but then we would need to wait for new release, etc... Since we were already concatenating paths anyway, let's just keep doing that.

For reference: https://gitlab.gnome.org/GNOME/gtk-doc/-/commit/350fbc5adaf5f996287ce9bbdd6b73000de9a401.

eli-schwartz:
I agree, we can easily be backwards-compatible with old gtkdoc here. However, someone should fix it there too...

PR #10284, opened 2022-04-15
Add support for multiline f-strings

selavy:
+ Extend the parser to recognize the multiline f-strings, which the
documentation already implies will work.

The syntax is like:
```
x = 'hello'
y = 'world'

msg = f'''This is a multiline string.

Sending a message: '@x@ @y@'
'''
```

which produces:
```
This is a multiline string.

Sending a message: 'hello world'

```

+ Added some f-string tests cases to "62 string arithmetic" to exercise
the new behavior.

### DIFF
--- a/docs/markdown/snippets/support-multiline-fstring.md
+++ b/docs/markdown/snippets/support-multiline-fstring.md
@@ -1,0 +1,22 @@
+## Added support for multiline fstrings
+
+Added support for multiline f-strings which use the same syntax as f-strings
+for string substition.
+
+```meson
+x = 'hello'
+y = 'world'
+
+msg = f'''Sending a message...
+"@x@ @y@"
+'''
+```
+
+which produces:
+
+```
+Sending a message....
+
+"hello world"
+
+```
--- a/mesonbuild/interpreterbase/interpreterbase.py
+++ b/mesonbuild/interpreterbase/interpreterbase.py
@@ -217,7 +217,10 @@ class InterpreterBase:
         elif isinstance(cur, mparser.TernaryNode):
             return self.evaluate_ternary(cur)
         elif isinstance(cur, mparser.FormatStringNode):
-            return self.evaluate_fstring(cur)
+            if isinstance(cur, mparser.MultilineFormatStringNode):
+                return self.evaluate_multiline_fstring(cur)
+            else:
+                return self.evaluate_fstring(cur)
         elif isinstance(cur, mparser.ContinueNode):
             raise ContinueRequest()
         elif isinstance(cur, mparser.BreakNode):
@@ -366,6 +369,10 @@ class InterpreterBase:
             return self.evaluate_statement(node.trueblock)
         else:
             return self.evaluate_statement(node.falseblock)
+
+    @FeatureNew('multiline format strings', '0.63.0')
+    def evaluate_multiline_fstring(self, node: mparser.MultilineFormatStringNode) -> InterpreterObject:
+        return self.evaluate_fstring(node)
 
     @FeatureNew('format strings', '0.58.0')
     def evaluate_fstring(self, node: mparser.FormatStringNode) -> InterpreterObject:
--- a/mesonbuild/mparser.py
+++ b/mesonbuild/mparser.py
@@ -114,6 +114,7 @@ class Lexer:
         self.token_specification = [
             # Need to be sorted longest to shortest.
             ('ignore', re.compile(r'[ \t]')),
+            ('multiline_fstring', re.compile(r"f'''(.|\n)*?'''", re.M)),
             ('fstring', re.compile(r"f'([^'\\]|(\\.))*'")),
             ('id', re.compile('[_a-zA-Z][_0-9a-zA-Z]*')),
             ('number', re.compile(r'0[bB][01]+|0[oO][0-7]+|0[xX][0-9a-fA-F]+|0|[1-9]\d*')),
@@ -203,9 +204,17 @@ class Lexer:
                             value = ESCAPE_SEQUENCE_SINGLE_RE.sub(decode_match, value)
                         except MesonUnicodeDecodeError as err:
                             raise MesonException(f"Failed to parse escape sequence: '{err.match}' in string:\n  {match_text}")
-                    elif tid == 'multiline_string':
-                        tid = 'string'
-                        value = match_text[3:-3]
+                    elif tid in {'multiline_string', 'multiline_fstring'}:
+                        # For multiline strings, parse out the value and pass
+                        # through the normal string logic.
+                        # For multiline format strings, we have to emit a
+                        # different AST node so we can add a feature check,
+                        # but otherwise, it follows the normal fstring logic.
+                        if tid == 'multiline_string':
+                            value = match_text[3:-3]
+                            tid = 'string'
+                        else:
+                            value = match_text[4:-3]
                         lines = match_text.split('\n')
                         if len(lines) > 1:
                             lineno += len(lines) - 1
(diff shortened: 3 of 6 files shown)


selavy:
Since this feature was already implied to work by the documentation, can someone please advise on whether to add it to the FeatureCheck framework or not.

eli-schwartz:
If it's never worked before, then that's a documentation issue. Please correct the documentation to indicate that multiline strings can only be combined with the .format() method of string formatting, and not the `f'string'` syntax.

This is not a review comment on adding support for it. If you'd like to combine that documentation fix with implementing multiline format strings, we're open to that suggestion, but...

... it would be a new feature, and be documented as a new feature, and it would need a FeatureCheck. (Because the feature check is supposed to tell people when they can use the feature, it doesn't make a lot of sense to say you can use it in versions where the documentation implied it was possible, but it actually wasn't possible.)

eli-schwartz:
AFAICT it was just neglected to even be thought about in #8485.
I don't really see a reason to not include support, personally.

test cases/common/62 string arithmetic/meson.build
```diff
@@ -14,3 +14,56 @@ b = 'b'
...
+# ------------------------------------------------------------------------------
+sub1 = 'the'
+sub2 = ' quick\n'
+sub3 = '\tbrown'
+sub4 = '\nfox'
+x = f'@sub1@@sub2@@sub3@@sub4@'
+
+if x != sub1 + sub2 + sub3 + sub4
+  error('string concat with f-strings is broken.')
+endif
```
eli-schwartz, line 29: We have an `assert()` function.

mesonbuild/mparser.py
```diff
@@ -203,9 +204,16 @@ def lex(self, filename: str) -> T.Generator[Token, None, None]:
...
+                    elif tid in {'multiline_string', 'multiline_fstring'}:
+                        # for multiline strings, parse the value out of the
+                        # quotes, update the line numbers, then funnel into
+                        # the normal string/fstring logic
+                        prefix, tid = {
+                            'multiline_string': ("'''", 'string'),
+                            'multiline_fstring': ("f'''", 'fstring'),
+                        }[tid]
+                        postfix = "'''"
+                        value = match_text[len(prefix):-len(postfix)]
```
eli-schwartz, line 216: Shouldn't we just hardcode postfix as 3, and hardcode prefix as either 3 or 4?

eli-schwartz (changes requested):

mesonbuild/mparser.py
selavy: I'm fine going either way on this. I was trying to make it a little clearer to read, but I can change it to be hardcoded.

eli-schwartz: To be clear, I don't have an issue with the dictionary getter, but `len("f'''")` can just be `4` inside the dictionary.

mesonbuild/interpreterbase/interpreterbase.py
```diff
@@ -367,6 +370,10 @@ def evaluate_ternary(self, node: mparser.TernaryNode) -> T.Optional[InterpreterO
         else:
             return self.evaluate_statement(node.falseblock)
 
+    @FeatureNew('multiline format strings', '0.63.0')
+    def evaluate_multiline_fstring(self, node: mparser.MultilineFormatStringNode) -> InterpreterObject:
+        return self.evaluate_fstring(node)
```
eli-schwartz, line 375: You can check in evaluate_fstring whether it's a multiline string and use `FeatureNew.single_use()` with a manually constructed message string.

This might save the need to add another AST token, but the question is whether that can be reliably detected. We can't have a literal newline character in a regular single single-quote string as written in meson.build, but is there any other way to end up with one and have it tokenized as a FormatStringNode?

e.g. if you format the newline escape sequence into a string, that should just result in a StringNode, not a FormatStringNode. To differentiate between a string and a multiline string we would need to know whether the newline char came from a literal or via string building and formatting, but an f-string doesn't have that concern... I think.

test cases/common/62 string arithmetic/meson.build
```diff
@@ -1,16 +1,49 @@
 project('string arithmetic', 'c')
```
eli-schwartz, line 1: We can add a minimum version here, and check in the file test.json whether a FeatureNew actually gets printed (and on which line of course).

eli-schwartz (changes requested):

eli-schwartz:
A bunch of these commits should eventually be squashed into one.

In "62 string arithmetic":

I'd also like to see the switch from if -> error() to use of assert in existing checks, be done as a distinct preparatory commit rather than grouped in with the new multiline f-string checks.

run_format_tests.py
```diff
@@ -63,6 +63,7 @@ def check_format() -> None:
         'work area',
         '.eggs', '_cache',              # e.g. .mypy_cache
         'venv',                         # virtualenvs have DOS line endings
+        'test cases',
```
eli-schwartz, line 66: This is effectively the same as removing `'.build'` from the list of suffixes to check, please don't do this
 The check exists for a reason.

eli-schwartz (changes requested):

run_format_tests.py
selavy: The reason I added it was because the test case has a tab character in it. I can back out that change and remove the tab from the string.

selavy: And as a note, I still need to update the docs from the other commit to reflect the version in which they became supported.

mesonbuild/interpreterbase/interpreterbase.py
selavy: I'm inclined to leave it with the extra AST node so that the code to detect a multline format string is in one location (the parser). Happy to change though if you feel otherwise.

eli-schwartz: I'm not really picky, I guess. It would probably be easier to implement without the extra token type, but on the other hand it's already implemented...

eli-schwartz:
Minor nit:

>  Convert if...error checks to debug() in string arithmetic tests

The commit message references debug() but uses assert().

Also you switched from debug() to message() in the second commit, dunno whether you intended that or not (or why you might prefer either one over the other).

selavy:
> Minor nit:
> 
> > Convert if...error checks to debug() in string arithmetic tests
> 
> The commit message references debug() but the diff uses assert().
> 
> Also you switched from debug() to message() in the second commit, dunno whether you intended that or not (or why you might prefer either one over the other).
> 
> One test failure but that is a flaky test, will restart it once github allows it...

Oh good catch on the commit message. I switched to `message` because `debug` was introduced in the latest version so it adds another warning to the output.

I forgot to include the updates to the docs so will do that once the other tests are finished (as well as fixing the commit message)

eli-schwartz:
> because `debug` was introduced in the latest version so it adds another warning to the output.

Eh, I suppose. I personally wouldn't much care, the stdout checks in test.json don't care either. :D

selavy:
Sorry for the force-push spam... I believe this now covers all of the changes you requested if you're okay with having the new AST node.

eli-schwartz (approved):

docs/markdown/snippets/support-multiline-fstring.md
```diff
@@ -0,0 +1,22 @@
+## Added support for `multiline fstrings`
```
eli-schwartz, line 1: One teeny, tiny formatting nit. Why is this an inline code segment?

selavy:
Not sure why some of the checks were cancelled, but I believe that is spurious. The only change in the force-push was removing the inline code formatting from the title of the docs.

eli-schwartz:
The canceled checks happen because we are using a deprecated windows-2016 runner image. They have been doing regular brownouts to warn people to stop using it, and we're aware of the issue... but actually switching over to newer images has proven challenging. Also we actually *do* want to test that Meson works correctly with older versions of Visual Studio. :(

I went ahead and restarted them now that the brownout is over. Had I noticed last night I would simply merge it regardless, but if we can run the check today, why not...

eli-schwartz:
Thanks!

PR #10290, opened 2022-04-20
Fix purelib and platlib validation in Python3 module.

jpakkane:
Match what the `python` module does.

We really should kill this module, though.


jpakkane:
So it all works. Anyone have issues?

PR #10292, opened 2022-04-20
Add NetBSD support in symbolextractor.

0-wiz-0:
Choose FreeBSD backend (OpenBSD backend would also work).


eli-schwartz:
FreeBSD backend is probably a better idea if possible as it assumes a functional nm --defined-only rather than trying to hack around it. ;)

I assume given your association with NetBSD you're best equipped to verify that this change actually works... thanks for the contribution, which lacking a test system I shall take on faith as correct.

PR #10282, opened 2022-04-15
improve help for -Ddebug

bonzini:
A trivial tweak to docs/help; "debug" is not a very helpful help message.

PR #10259, opened 2022-04-10
Look for Qt 6 tools in libexec/

stefanha:
This is @jlindgren90's "Add support for Qt 6.1+" #9792 with the following changes:
1. Initialize `_QtBase.libexecdir` to `None` so earlier versions of Qt that didn't report a libexec/ directory still work.
2. Handle the non-libexec/ case in `get_qmake_host_libexecs()` too.
3. Add a new commit "tests: fix incomplete Qt 6 support" that makes test "frameworks/4 qt" really work with Qt 6.

The goal is to detect the moc, rcc, uic Qt tools on Qt 6 where they moved from the bin/ directory to the libexec/ directory.

@eli-schwartz: Since @jlindren90 has not been active lately I reworked his pull request from earlier this year. This obsoletes #9792 and #10255.

PR #10214, opened 2022-03-30
Fix typos in Xcode backend. Closes #10193.

jpakkane:

PR #10295, opened 2022-04-21
Patch collection for 0.61.5

bonzini:
* #10168
* #10204
* #10184
* #10213 
* #10259
* #10270 
* #9988 
* #10268 
* #10266
* #10258 
* #10176 
* #10209
* #10214 
* #10228
* #10250 
* #10191
* #10181
* #10254
* #10282
* #10324
* #10331
* #10129
* #10290 
* #10297


bonzini:
@nirbheek that's for you...

eli-schwartz:
The install_symlink change is incompatible as-is with 0.61 because it changes the calling API of a function that is used in multiple places in 0.61 but since 01e92dc543fb7443fba8a33687ffd726f11433e8 in fewer places.

The other caller would need to be adapted (or that commit would need to be selected to backport).

nirbheek:
We should also backport CI fixes to make sure that the CI is green here.

Also we should include these (see 0.62.1 milestone description):

https://github.com/mesonbuild/meson/pull/10191 (will make it easier to backport future changes)

https://github.com/mesonbuild/meson/commit/1420d0daceb10cafb52a7405f157032a5cc811a5

https://github.com/mesonbuild/meson/commit/3c80f8f96591a03207b58e8201700bfcbcfa95a1

https://github.com/mesonbuild/meson/commit/33aa803521b2aa62688833842ea7eddb828e0e25

eli-schwartz:
That last one for the vcs_tag changes is only applicable to the 0.62.x series as it backports a FeatureNew from `master -> 0.62.1` to specify that 0.62.0 started accepting a new type of value. It didn't work in 0.61.x and so shouldn't be backported to 0.61.5

mesonbuild/dependencies/data/CMakeListsLLVM.txt
```diff
@@ -1,4 +1,3 @@
-cmake_minimum_required(VERSION ${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION}.${CMAKE_PATCH_VERSION} )
```
eli-schwartz, line 1: You're going to need to manually patch this in the old version of meson, since one of the changes that broke python 3.6 support was beginning to depend on importlib.resources to use bundled data files instead of syncing their contents into an inline string.

There's a script to update this for you.

eli-schwartz: Basically just add an additional commit "run tools/gen_data.py and commit the results".

eli-schwartz:
@nirbheek is this good to go now?

PR #10297, opened 2022-04-21
Strip "local" bits from system paths.

jpakkane:
Starting with Python 3.10 Debian changed how it reports paths via sysconfig.
They now have a 'local' path segment in them, presumably to match that
when installed system-wide things to go "/usr/local". When you install in
a custom dir, though, the "local" part is not actually used. So don't use that
then.

### DIFF
--- a/run_meson_command_tests.py
+++ b/run_meson_command_tests.py
@@ -24,17 +24,24 @@ from pathlib import Path
 from mesonbuild.mesonlib import windows_proof_rmtree, python_command, is_windows
 from mesonbuild.coredata import version as meson_version
 
+# Handle the scheme that Debian patches in the as default
+import sysconfig
+# This function was renamed and made public in Python 3.10
+if hasattr(sysconfig, 'get_default_scheme'):
+    scheme = sysconfig.get_default_scheme()
+else:
+    scheme = sysconfig._get_default_scheme()
+if scheme == 'posix_local':
+    scheme = 'posix_prefix'
 
 def get_pypath():
-    import sysconfig
-    pypath = sysconfig.get_path('purelib', vars={'base': ''})
+    pypath = sysconfig.get_path('purelib', scheme=scheme, vars={'base': ''})
     # Ensure that / is the path separator and not \, then strip /
     return Path(pypath).as_posix().strip('/')
 
 def get_pybindir():
-    import sysconfig
     # 'Scripts' on Windows and 'bin' on other platforms including MSYS
-    return sysconfig.get_path('scripts', vars={'base': ''}).strip('\\/')
+    return sysconfig.get_path('scripts', scheme=scheme, vars={'base': ''}).strip('\\/')
 
 class CommandTests(unittest.TestCase):
     '''


run_meson_command_tests.py
```diff
@@ -27,14 +27,22 @@
 
 def get_pypath():
     import sysconfig
-    pypath = sysconfig.get_path('purelib', vars={'base': ''})
+    pypath = sysconfig.get_path('purelib', vars={'base': ''}).replace('dist-packages', 'site-packages')
     # Ensure that / is the path separator and not \, then strip /
+    # Starting with Python 3.10 the Debian installation returns paths like
+    # '/usr/local', even though it does the installation to just ''/usr'.
```
tristan957, line 33:
```suggestion
    # '/usr/local', even though it does the installation to just '/usr'.
```

nirbheek (commented):
Instead of all this, you can also just set the scheme:

```
>>> sysconfig.get_path('purelib', vars={'base': ''})
'/local/lib/python3.10/dist-packages'
>>> sysconfig.get_scheme_names()
('deb_system', 'nt', 'nt_user', 'osx_framework_user', 'posix_home', 'posix_local', 'posix_prefix', 'posix_user')
>>> sysconfig.get_path('purelib', scheme='posix_prefix', vars={'base': ''})
'/lib/python3.10/site-packages'
```

nirbheek:
Ah, you also need to make it conditional on this being Debian. For that, I'd recommend doing this:

```python
scheme = sysconfig.get_default_scheme()
if scheme == 'posix_local':
   scheme = 'posix_prefix'
sysconfig.get_path('purelib', scheme=scheme, vars={'base': ''})
```

`'posix_local'` is a custom scheme defined by Debian (see `debian/patches/sysconfig-debian-schemes.diff` in the [source package](https://packages.debian.org/sid/python3.10)). No other platform has it.

run_meson_command_tests.py
```diff
@@ -24,17 +24,20 @@
 from mesonbuild.mesonlib import windows_proof_rmtree, python_command, is_windows
 from mesonbuild.coredata import version as meson_version
 
+# HAndle the cusomt scheme that Debian patches in.
```
nirbheek, line 27: Since CI failed on Cygwin, maybe you want to fix these typos? :grin:

nirbheek:
CI failure is relevant:

```
  File "/cygdrive/d/a/meson/meson/run_meson_command_tests.py", line 29, in <module>
    scheme = sysconfig.get_default_scheme()
AttributeError: module 'sysconfig' has no attribute 'get_default_scheme'
```

nirbheek:
> Changed in version 3.10: This function was previously named _get_default_scheme() and considered an implementation detail.

https://docs.python.org/3/library/sysconfig.html

So I guess we have to check both...

nirbheek:
I updated the PR, waiting for CI now.

jpakkane:
It be green. Not sure what happened with that cygwin one, though.

Issue #14, opened 2014-10-20
Feature: install_* to respect path of files


RobinMcCorkell:
It would be of great benefit if Meson supported something similar to GNU Automake's `nobase` option for installing files (see http://www.gnu.org/software/automake/manual/html_node/Alternative.html).

tristan957:
Why was this closed? `install_subdir` is just a work around. `no_base` is a feature that should be implemented. I don't want Meson flattening my headers on install. It is currently throwing off my entire library because the headers themselves use the nested paths of my project, but when they install, they are flat and thus the includes don't resolve. This is a huge bug in Meson in my opinion. Following the directory structure should be the default. Please reopen the issue.

FFY00:
My workaround:
```meson
# Install files from sources array
# Loop needed as meson doesn't preserve the path for some reason
# See: https://github.com/mesonbuild/meson/issues/3371
include_subdir = 'mylib'
foreach file : mylib_sources
    file_path = file.split('/')

    # Extract folder
    folder_path = []
    foreach path : file_path
        if path != file_path[-1]
            folder_path += path
        endif
    endforeach

    folder = join_paths(folder_path)
    install_headers(file, subdir: join_paths(include_subdir, folder))
endforeach
```

nirbheek:
> I don't want Meson flattening my headers on install.

That sounds like a bug. I don't believe that's the intention of the call. Could you please open a new issue about that?

tristan957:
@nirbheek I believe #3371 is about this behavior.

dzhang-b:
I think there should be two ways of install headers, similar to what described here.
[nobase_headers](https://www.gnu.org/software/automake/manual/html_node/Headers.html)

Currently we have `install_subdir`, but it is not obvious that it should be used to install header if I am not looking at this github issue.

eli-schwartz:
This ticket was originally about having a nobase-style approach to the install_headers() function. Despite adding install_subdir, this did not actually answer the request, even if it could be used that way (in fact, many, many install_subdir uses have nothing to do with headers and people don't always realize it can be used for this).

It also doesn't work when there are some private headers and some public headers.

Anyway, it is now implemented in #10299 and will be available in the next version of Meson as `install_headers('list', 'of', 'headers', preserve_path: true)`

PR #10299, opened 2022-04-21
implemented `preserve_path` for install_headers

### LINKED ISSUE #3371: install_headers(preserve_path)
```meson
# proj/kola.h -> /usr/include/myproj/kola.h
install_headers('proj/kola.h', subdir : 'myproj')

# proj/kola.h -> /usr/include/myproj/proj/kola.h
install_headers('proj/kola.h', subdir : 'myproj', preserve_path : true)
```

```meson
# you can write
install_headers(
    'foo/bar.h',
    'baz/bob.h',
    subdir : 'proj',
    preserve_path : true,
)

# instead of
install_headers('foo/bar.h', subdir : 'proj/foo')
install_headers('baz/bob.h', subdir : 'proj/baz')
```

One possible alternative to consider is `preserve_path_from : directory`, but it introduces invalid uses of API and by calling `install_headers` from subdirectory the same behavior can be achieved with proposal above, so I don't think it's worth implementing.

@textshell also mentioned:
My usecase was a list of headers to install. Where the headers might not be all in one directory. Allowing that usecase is esp. important for an IDE integration that can modify file lists, but is not smart enough to actually write and auto maintain more general elements in a meson.build file.
And i belive editing file lists should be the primary write interaction of an IDE with the meson.build files. Because meson files are really not easy to auto maintain because they allow all kinds of different styles and usages.
Also i think IDEs automatically editing target definitions to add files is not what most projects expect. I think most somewhat bigger projects have lists of files instead of a list in the target definition call.

sp1ritCS:
This allows for structures like
```meson
headers = [
  'header1.h',
  'header2.h',
  'header3.h',
  'inner/header1.h',
  'inner/header2.h',
  'inner/header3.h'
]
```
with headers living alongside source files in a dir with multiple child-dirs.

to be passed to `install_headers(headers, subdir: 'mylib', preserve_path: true)` while keeping the original meaning.

This was requested in #14 and again in #3371

Fixes #3371

### DIFF
--- a/docs/markdown/snippets/install_headers_preserve_path_arg.md
+++ b/docs/markdown/snippets/install_headers_preserve_path_arg.md
@@ -1,0 +1,35 @@
+## Added preserve_path arg to install_headers
+
+The [[install_headers]] function now has an optional argument `preserve_path`
+that allows installing multi-directory headerfile structures that live
+alongside sourcecode with a single command.
+
+For example, the headerfile structure
+
+```meson
+headers = [
+  'one.h',
+  'two.h',
+  'alpha/one.h',
+  'alpha/two.h',
+  'alpha/three.h'
+  'beta/one.h'
+]
+```
+
+can now be passed to `install_headers(headers, subdir: 'mylib', preserve_path: true)`
+and the resulting directory tree will look like
+
+```
+{prefix}
+└── include
+    └── mylib
+        ├── alpha
+        │   ├── one.h
+        │   ├── two.h
+        │   └── three.h
+        ├── beta
+        │   └── one.h
+        ├── one.h
+        └── two.h
+```
--- a/mesonbuild/interpreter/interpreter.py
+++ b/mesonbuild/interpreter/interpreter.py
@@ -2065,6 +2065,7 @@ external dependencies (including libraries) must go to "dependencies".''')
     @typed_kwargs(
         'install_headers',
         KwargInfo('install_dir', (str, NoneType)),
+        KwargInfo('preserve_path', bool, default=False, since='0.63.0'),
         KwargInfo('subdir', (str, NoneType)),
         INSTALL_MODE_KW.evolve(since='0.47.0'),
     )
@@ -2078,12 +2079,25 @@ external dependencies (including libraries) must go to "dependencies".''')
                 raise InterpreterException('install_headers: cannot specify both "install_dir" and "subdir". Use only "install_dir".')
             if os.path.isabs(install_subdir):
                 mlog.deprecation('Subdir keyword must not be an absolute path. This will be a hard error in the next release.')
+        else:
+            install_subdir = ''
 
-        h = build.Headers(source_files, install_subdir, kwargs['install_dir'],
-                          kwargs['install_mode'], self.subproject)
-        self.build.headers.append(h)
+        dirs = collections.defaultdict(list)
+        ret_headers = []
+        if kwargs['preserve_path']:
+            for file in source_files:
+                dirname = os.path.dirname(file.fname)
+                dirs[dirname].append(file)
+        else:
+            dirs[''].extend(source_files)
 
-        return h
+        for childdir in dirs:
+            h = build.Headers(dirs[childdir], os.path.join(install_subdir, childdir), kwargs['install_dir'],
+                              kwargs['install_mode'], self.subproject)
+            ret_headers.append(h)
+            self.build.headers.append(h)
+
+        return ret_headers
 
     @typed_pos_args('install_man', varargs=(str, mesonlib.File))
     @typed_kwargs(
--- a/test cases/common/9 header install/meson.build
+++ b/test cases/common/9 header install/meson.build
@@ -1,11 +1,12 @@
 project('header install')
 
-as_array = ['subdir.h']
+as_array = ['subdir.h', 'child/childdir.h']
 
 subdir('vanishing_subdir')
 subdir('sub')
 
 h1 = install_headers('rootdir.h')
 h2 = install_headers(as_array, subdir : 'subdir')
-h3 = install_headers(subheader)
-h4 = install_headers(disabler())
+h3 = install_headers(as_array, subdir : 'subdir', preserve_path : true)
+h4 = install_headers(subheader)
+h5 = install_headers(disabler())
--- a/docs/yaml/functions/install_headers.yaml
+++ b/docs/yaml/functions/install_headers.yaml
@@ -29,6 +29,13 @@ example: |
   install_headers('common.h', 'proj/kola.h', install_dir : 'cust', subdir : 'myproj')
   ```
 
+  This will install `common.h` into `/{prefix}/include` and `kola.h`
+  into `/{prefix}/include/proj/`:
+
+  ```meson
+  install_headers('common.h, 'proj/kola.h', preserve_path : true)
+  ```
+
 varargs:
   name: file
   type: file | str
@@ -54,3 +61,12 @@ kwargs:
       and optionally the owner/uid and group/gid for the installed files.
 
       See the `install_mode` kwarg of [[install_data]] for more information.
+
+  preserve_path:
+    type: bool
+    since: 0.63.0
+    default: false
+    description: |
+      Disable stripping child-direcories from header files when installing.
+
+      This is equivalent to GNU Automake's `nobase` option.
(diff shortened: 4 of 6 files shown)


mesonbuild/interpreter/interpreter.py
```diff
@@ -2065,6 +2065,7 @@ def add_test(self, node: mparser.BaseNode, args: T.List, kwargs: T.Dict[str, T.A
     @typed_kwargs(
         'install_headers',
         KwargInfo('install_dir', (str, NoneType)),
+        KwargInfo('nobase', (bool, NoneType), since='0.63.0'),
```
eli-schwartz, line 2068: This should just be a bool, defaulting to false.

mesonbuild/interpreter/interpreter.py
```diff
@@ -2079,8 +2080,9 @@ def func_install_headers(self, node: mparser.BaseNode,
             if os.path.isabs(install_subdir):
                 mlog.deprecation('Subdir keyword must not be an absolute path. This will be a hard error in the next release.')
 
-        h = build.Headers(source_files, install_subdir, kwargs['install_dir'],
-                          kwargs['install_mode'], self.subproject)
+        h = build.Headers(source_files, os.path.join(self.environment.source_dir, self.subdir + "/"),
+                install_subdir, kwargs['install_dir'],
+                    kwargs['install_mode'], self.subproject, kwargs['nobase'])
```
eli-schwartz, line 2085: Instead of moving this parsing logic into the `build.Headers` class, the backend, and minstall itself, what about having the interpreter group these files by leading path and generate a build.Headers per destination subdir?

eli-schwartz (changes requested):
This should use github closing keywords, i.e. on its own line in the commit message and PR description

```
Fixes #3371
```

I would also suggest using the imperative mood in the commit message summary, i.e. not "implemented foo" but "implement foo".

The commit message should also give a basic description of what that actually does, not dissimilar to the PR description (often they are the same thing in fact). This is useful to people looking at the git logs, but not tracking down and consulting the PR corresponding to each commit.

Lastly, this needs to come with:
- documentation, in `docs/yaml/functions/install_headers.yaml`
- release notes, in `docs/markdown/snippets/*.md`
- a test case demonstrating the functionality, i.e. expand `test cases/common/9 header install/` to demonstrate the usage of `nobase: true` and check that it installs the files as expected in test.json

sp1ritCS:
@eli-schwartz One thing I'm unsure about is the return value of `func_install_headers`. The meson docs describe the `install` function as returning void, while this one clearly doesn't. Do I get away with returning an Array<Headers>, while keeping the non-nobase return value as a single Headers, or does that lead to issues somewhere else?

jpakkane:
`nobase` is not really a great term. Maybe use something more descriptive like `preserve_subdirs`?

sp1ritCS:
> `nobase` is not really a great term. Maybe use something more descriptive like `preserve_subdirs`?

Blame autotools :) - I've decided to go with `preserve_path`, like @sarum9in suggested in #3371.

eli-schwartz:
> One thing I'm unsure about is the return value of `func_install_headers`. The meson docs describe the `install` function as returning void, while this one clearly doesn't.

Many functions are documented as returning void, but internally they actually return the data they operate on. In theory, they could return (implicit?) None. Some modules do that. Attempting to assign such functions returns a fatal `ERROR: Can not assign None to variable.`

The install_*() functions don't skip returning a value, or even explicitly return None, though. They return objects that are useless in the meson.build DSL, as they cannot ever be used as input to anything other than, idk, `'@0@'.format(...)` which has undefined behavior of formatting a string with the python `__repr__()` of any object that doesn't have a documented string value.

So it really doesn't make any difference to users what the return value after your changes are. Internally, if e.g. modules want to use `self.interpreter.func_install_headers()` they might find it valuable to do something with those build.Header objects, so we should probably return the entire array of them, and continue to document the Meson DSL as returning void (in this case, I guess "undocumented nonsense which always produces errors if you try using it").

docs/markdown/snippets/install_headers_preserve_path_arg.md
```diff
@@ -0,0 +1,35 @@
...
+  'one.h',
+  'two.h',
+  'alpha/one.h',
+  'alpha/two.h',
+  'alpha/three.h'
+  'beta/one.h'
+]
+```
+
+can now be passed to `install_header(headers, subdir: 'mylib', preserve_path: true)`
```
eli-schwartz, line 20: "install_headers" is misspelled here (and twice in the commit message).

docs/markdown/snippets/install_headers_preserve_path_arg.md
```diff
@@ -0,0 +1,35 @@
+## Added nobase arg to install_headers
+
+The `install_headers` function now has an optional argument `preserve_path`
```
eli-schwartz, line 3: https://mesonbuild.com/Yaml-RefMan.html#linking-to-the-reference-manual

You can reference the function name as `[[install_headers]]` and it will do a cross-reference for you.

mesonbuild/interpreter/interpreter.py
```diff
@@ -2079,11 +2080,30 @@ def func_install_headers(self, node: mparser.BaseNode,
...
+        if kwargs['preserve_path']:
+            dirs = {}
+            ret_headers = []
+            for file in source_files:
+                dirname = os.path.dirname(file.fname)
+                dirfiles = dirs.get(dirname)
+                if dirfiles is None:
+                    dirs[dirname] = [file]
+                else:
+                    dirfiles.append(file)
```
eli-schwartz, line 2092: This would be more elegantly written if `dirs = collections.defaultdict(list)`, then you can simply append to it without regard to whether it exists.

test cases/common/9 header install/meson.build
```diff
@@ -1,11 +1,12 @@
...
+as_array = ['subdir.h', 'child/childdir.h', 'child/otherchild.h']
 
 subdir('vanishing_subdir')
 subdir('sub')
 
 h1 = install_headers('rootdir.h')
 h2 = install_headers(as_array, subdir : 'subdir')
-h3 = install_headers(subheader)
-h4 = install_headers(disabler())
+h3 = install_headers(as_array, subdir : 'subdir', preserve_path : true)
```
eli-schwartz, line 10: I'm actually slightly surprised that this worked, since it seems you are installing `includedir/subdir.h` twice now (as that file doesn't have a leading directory component to preserve).

sp1ritCS: I'm aware of that. I didn't think this was an issue, since meson doesn't complain about installing the same file twice. Unless this is destined to change, I don't see why to change that behaviour.

sp1ritCS:
> if e.g. modules want to use self.interpreter.func_install_headers() they might find it valuable to do something with those build.Header objects, so we should probably return the entire array of them

I've made it so, that the no-preserve_path also returns a list (of the single Headers object it used to return). That's probably better if any module implementation wants to somehow use the return value without doing a typecheck first.

test cases/common/9 header install/meson.build
eli-schwartz: Right, that was me commenting to myself.

QuLogic:
> Blame autotools :) - I've decided to go with `preserve_path`, like @sarum9in suggested in #3371.

Don't forget the PR title...

docs/markdown/snippets/install_headers_preserve_path_arg.md
```diff
@@ -0,0 +1,35 @@
+## Added nobase arg to install_headers
```
eli-schwartz, line 1: Oops, this still refers to the old name.

mesonbuild/interpreter/interpreter.py
```diff
@@ -2079,11 +2080,26 @@ def func_install_headers(self, node: mparser.BaseNode,
...
-        return h
+            for childdir in dirs:
+                h = build.Headers(dirs[childdir], os.path.join(install_subdir, childdir), kwargs['install_dir'],
+                                  kwargs['install_mode'], self.subproject)
+                ret_headers.append(h)
+                self.build.headers.append(h)
+
+            return ret_headers
+        else:
+            h = build.Headers(source_files, install_subdir, kwargs['install_dir'],
```
eli-schwartz, line 2098: Since this code is essentially the same as the other branch, what are your thoughts on making `dirs` without a preserce_path kwarg be `{'': source_files}`?

eli-schwartz: Hmm, does it pass CI if you leave the dot out?

sp1ritCS: yes, seems to windows related. didn't occur on linux

PR #10303, opened 2022-04-23
CMake: fix custom target: directories are no longer used as inputs

mensinda:


mesonbuild/cmake/interpreter.py
```diff
@@ -714,7 +714,7 @@ def ensure_absolute(x: Path) -> Path:
             # targets, etc. This reduces the chance of misdetecting input files
             # as outputs from other targets.
             # See https://github.com/mesonbuild/meson/issues/6632
-            if not raw.is_absolute() and (self.current_src_dir / raw).exists():
+            if not raw.is_absolute() and (self.current_src_dir / raw).exists() and (self.current_src_dir / raw).is_file():
```
eli-schwartz, line 717:
```suggestion
            if not raw.is_absolute() and (self.current_src_dir / raw).is_file():
```

We don't need to check both.

eli-schwartz (changes requested):

mesonbuild/cmake/interpreter.py
eli-schwartz: ping?

jpakkane:
Ping indeed, this is the last MR needed for 0.62.2.

mesonbuild/cmake/interpreter.py
mensinda: Forgot about that one...

PR #10304, opened 2022-04-23
Fix handling of install_tag and extend its usage in gnome module 

susnux:
Purpose of the pull request:

1. There is a regression: `CustomTarget` does not set `install_tag` correctly if one tag should be used for all outputs (see comment https://github.com/mesonbuild/meson/pull/10304#issuecomment-1107636710).
2. Currently there is no way to specify to only install `help` generated by `gnome.yelp`. So this changes allow to use the `man` `install_tag` for installing `help` files.

### DIFF
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -2450,13 +2450,9 @@ class CustomTarget(Target, CommandBase):
         self.install = install
         self.install_dir = list(install_dir or [])
         self.install_mode = install_mode
-        _install_tag: T.List[T.Optional[str]]
-        if not install_tag:
-            _install_tag = [None] * len(self.outputs)
-        elif len(install_tag) == 1:
-            _install_tag = list(install_tag) * len(self.outputs)
-        else:
-            _install_tag = list(install_tag)
+        _install_tag: T.List[T.Optional[str]] = [None] if not install_tag else stringlistify(install_tag)
+        if len(_install_tag) == 1:
+            _install_tag = list(_install_tag) * len(self.outputs)
         self.install_tag = _install_tag
         self.name = name if name else self.outputs[0]
 
--- a/docs/markdown/Installing.md
+++ b/docs/markdown/Installing.md
@@ -168,6 +168,7 @@ time, please help extending the list of well known categories.
   * `install_man()`.
 - `doc`:
   * `gnome.gtkdoc()`,
+  * `gnome.yelp()`,
   * `hotdoc.generate_doc()`.
 - `i18n`:
   * `i18n.gettext()`,
--- a/mesonbuild/modules/gnome.py
+++ b/mesonbuild/modules/gnome.py
@@ -1281,7 +1281,7 @@ class GnomeModule(ExtensionModule):
         install_dir = os.path.join(state.environment.get_datadir(), 'help')
         c_install_dir = os.path.join(install_dir, 'C', project_id)
         c_data = build.Data(sources_files, c_install_dir, c_install_dir,
-                            mesonlib.FileMode(), state.subproject)
+                            mesonlib.FileMode(), state.subproject, install_tag='doc')
         targets.append(c_data)
 
         media_files: T.List[mesonlib.File] = []
@@ -1291,7 +1291,7 @@ class GnomeModule(ExtensionModule):
             media_files.append(f)
             m_install_dir = os.path.join(c_install_dir, os.path.dirname(m))
             m_data = build.Data([f], m_install_dir, m_install_dir,
-                                mesonlib.FileMode(), state.subproject)
+                                mesonlib.FileMode(), state.subproject, install_tag='doc')
             targets.append(m_data)
 
         pot_file = os.path.join('@SOURCE_ROOT@', state.subdir, 'C', project_id + '.pot')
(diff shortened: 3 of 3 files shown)


xclaesse:
Shouldn't it be `doc` like `gnome.gtkdoc()`?

susnux:
> Shouldn't it be `doc` like `gnome.gtkdoc()`?

That's a good point, I would say `doc` is developer documentation whereas `man` is user manuals and `yelp` produces mostly user documentation.

xclaesse:
Hm, maybe I shouldn't have made different tags for them in the first place. Bah, LGTM.

susnux:
**Do not accept yet!**
I tested it with a different project now (gtk-doc), and for that project `gnome.yelp` fails using this changes, I will have to look into that...

```
  File "/usr/lib/python3.8/site-packages/mesonbuild/mesonmain.py", line 153, in run
  File "/usr/lib/python3.8/site-packages/mesonbuild/msetup.py", line 301, in run
    app.generate()
  File "/usr/lib/python3.8/site-packages/mesonbuild/msetup.py", line 185, in generate
    self._generate(env)
  File "/usr/lib/python3.8/site-packages/mesonbuild/msetup.py", line 247, in _generate
    intr.backend.generate()
  File "/usr/lib/python3.8/site-packages/mesonbuild/backend/ninjabackend.py", line 558, in generate
    self.generate_install()
  File "/usr/lib/python3.8/site-packages/mesonbuild/backend/ninjabackend.py", line 1124, in generate_install
    self.create_install_data_files()
  File "/usr/lib/python3.8/site-packages/mesonbuild/backend/backends.py", line 1525, in create_install_data_files
    pickle.dump(self.create_install_data(), ofile)
  File "/usr/lib/python3.8/site-packages/mesonbuild/backend/backends.py", line 1512, in create_install_data
    self.generate_target_install(d)
  File "/usr/lib/python3.8/site-packages/mesonbuild/backend/backends.py", line 1562, in generate_target_install
    assert len(t.install_tag) == num_out
AssertionError
```

xclaesse:
@susnux Our CI catched that too :)

susnux:
@xclaesse 

Can it be that this is a regression in `CustomTarget`? Because `self.install_tag` is set to `['m', 'a', 'n']`.
I guess the lines regarding the tag are wrong here:
https://github.com/mesonbuild/meson/blob/a2079939de2d1d271e5ea7f23c3656a73e325728/mesonbuild/build.py#L2456

I think it should be:
```py
if not install_tag:
    _install_tag = [None] * len(self.outputs)
elif isinstance(nstall_tag, str):
    _install_tag = list(install_tag) * len(self.outputs)
else:
     _install_tag = list(install_tag)
```

Or am I wrong?

xclaesse:
@susnux I don't think that's a regression, CustomTarget() expects a list of strings, you should pass `['man']`. Passing a single string is allowed from the public API `custom_target()` (it will listify user value) but not when you use internal API. See the type the `__init__` function expects: `install_tag: T.Optional[T.Sequence[T.Optional[str]]] = None`.

susnux:
> See the type the `__init__` function expects: `install_tag: T.Optional[T.Sequence[T.Optional[str]]] = None`.

Seems reasonable, got confused by `BuildTarget` as it allows passing a single string (because it uses `stringlistify`).
Should I leave the changes on the `CustomTarget` class? Nevertheless I will change the `yelp` part to pass `install_tag` as `[...]`.

eli-schwartz:
> That's a good point, I would say `doc` is developer documentation whereas `man` is user manuals and `yelp` produces mostly user documentation.

I would say "doc" is unspecified documentation formats, while "man" is troff-formatted `man(1)` pages.

If there was an intention to distinguish between developer docs and user docs, then perhaps the tags should be "devdoc" and "userdoc".

Or "yelpdoc" and assume people interested in multiple kinds of docs will specify multiple install tags, idk.

susnux:
@xclaesse Fixed it to pass the tag as list`[...]`. I leave the changes to the `__init__` as it adds some validation, or should I revert that too?

@eli-schwartz Seems reasonable, I guess multiple users would be confused if man is more than man-pages. `devdoc` and `userdoc` would have been be a more general solution.

mesonbuild/build.py
```diff
@@ -2450,13 +2450,9 @@ def __init__(self,
...
-        _install_tag: T.List[T.Optional[str]]
-        if not install_tag:
-            _install_tag = [None] * len(self.outputs)
-        elif len(install_tag) == 1:
-            _install_tag = list(install_tag) * len(self.outputs)
-        else:
-            _install_tag = list(install_tag)
+        _install_tag = [None] if not install_tag else stringlistify(install_tag)
+        if len(_install_tag) == 1:
+            _install_tag = list(_install_tag) * len(self.outputs)
```
xclaesse, line 2455: I don't think this change is needed any more.

xclaesse:
mypy is not happy: `mesonbuild/backend/backends.py:1541:24: error: Argument 1 to "len" has incompatible type "object"; expected "Sized"`. I think it's because it's too dumb to understand the type of `_install_tag` because you removed `_install_tag: T.List[T.Optional[str]]`.

susnux:
> mypy is not happy

That is fixed, but for the one negative CI: Is that a false positive? Not sure why only this Ubuntu version fails.

xclaesse:
I retried the job and it passed. LGTM. Thanks!

PR #10305, opened 2022-04-23
Do not use 'echo' as it is not available on plain cmd.exe

jpakkane:


xclaesse:
LGTM

jpakkane:
Mac failure is unrelated.

PR #10306, opened 2022-04-24
Qt compile translations errorchecking

### LINKED ISSUE #10301: qt5.compile_translations regression
**Describe the bug**
Since a recent meson version, the attached meson.build doesn't build anymore but fails with the following error
```
ERROR: Target 'qt5-compile-de.ts' has 1 outputs: ['de.qm'], but only 12 "install_dir"s were found.
Pass 'false' for outputs that should not be installed and 'true' for
using the default installation directory for an output.
```
git bisect leads to the following commit:
```
11f96380351a88059ec55f1070fdebc1b1033117 is the first bad commit
commit 11f96380351a88059ec55f1070fdebc1b1033117
Author: Dylan Baker <dylan@pnwbakers.com>
Date:   Wed Jan 19 12:42:25 2022 -0800

    build: replace kwargs in CustomTarget initializer
    
    Because we don't want to pass the Interpreter kwargs into the build
    layer. This turned out to be a mega commit, as there's really on elegant
    way to make this change in an incremental way. On the nice side, mypy
    made this change super easy, as nearly all of the calls to
    `CustomTarget` are fully type checked!
    
    It also turns out that we're not handling install_tags in custom_target
    correctly, since we're not converting the boolean values into Optional
    values!
```

**To Reproduce**
```
project('foobar', 'c',
        version : '0.0.1',
        meson_version : '>=0.44.0')

qt5 = import('qt5')
qt5.compile_translations(ts_files : ['de.ts'],
                         install : true,
                         install_dir : join_paths(get_option('datadir'), 'foobar'),
                         build_by_default: true)
```

**Expected behavior**
meson should understand this, as before.

**system parameters**
  * plain native build
  * arch linux
  * Python 3.10.4
* what `meson --version`
  * Tried both 0.62.0 and git master
  * 1.10.2

eli-schwartz:
Since the conversion of func_custom_target to take over errorchecking from build.CustomTarget, compile_translations has been incapable of installing anything properly.

Fixes #10301

### DIFF
--- a/mesonbuild/modules/qt.py
+++ b/mesonbuild/modules/qt.py
@@ -546,6 +546,8 @@ class QtBaseModule(ExtensionModule):
         if any(isinstance(s, (build.CustomTarget, build.CustomTargetIndex, build.GeneratedList)) for s in ts_files):
             FeatureNew.single_use('qt.compile_translations: custom_target or generator for "ts_files" keyword argument',
                                   '0.60.0', state.subproject, location=state.current_node)
+        if kwargs['install'] and not kwargs['install_dir']:
+            raise MesonException('qt.compile_translations: "install_dir" keyword argument must be set when "install" is true.')
         qresource = kwargs['qresource']
         if qresource:
             if ts_files:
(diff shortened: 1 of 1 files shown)


z3ntu:
Thanks, my project builds again with this fix!

punytroll:
I can confirm that meson commands _setup_ and _install_ now work properly.

With this change, providing an "install_dir" becomes mandatory, when using "install". I have never used that, so I don't know whether there really is no sensible "default" install directory. Also on the Qt module documentation, it could be noted explicitly that both parameters are optional **together**.

jpakkane:
> I don't know whether there really is no sensible "default" install directory

If that turns out to be the case, we can add that in a followup pr.

PR #10308, opened 2022-04-24
Remove deprecated python3 module.

jpakkane:


eli-schwartz:
As I pointed out in https://github.com/mesonbuild/meson/pull/10290#issuecomment-1104254136 there may be projects still using this because they want to support Ubuntu Bionic. It's not even out of *standard* support until next year. :(

jpakkane:
> there may be projects still using this because they want to support the version of Meson distributed with Ubuntu

But is that really realistic? I'd suspect that there are other things that already don't work with both versions.

eli-schwartz:
I'm not sure what you mean by that. Do you mean that code written and tested for meson 0.45, which is known to work there, is realistically suspected to not work with meson 0.62?

If that's the case, shouldn't that be a bug we should fix?

And either way, why does deleting a low-maintenance module whose only fault is having a better alternative, make that *better*?

jpakkane:
> If that's the case, shouldn't that be a bug we should fix?

Yes. But this was mostly speculative. I don't know of any actual bugs, but I suspect that they are hidden somewhere.

> And either way, why does deleting a low-maintenance module whose only fault is having a better alternative, make that better?

The reason I started doing this is because it turned out to _not_ be low maintenance. Debian's Python changes in 3.10 broke stuff that I had to fix.

eli-schwartz:
> The reason I started doing this is because it turned out to _not_ be low maintenance. Debian's Python changes in 3.10 broke stuff that I had to fix.

Low maintenance != no maintenance

The only changes needed were to the testsuite meson.build files, and moreover that just involved syncing existing logic from the python module test files.

Furthermore, those changes also broke the `python3 setup.py install` command tests which you also had to fix, which is highly unfortunate and annoying and also not really a justification to remove the python3 module IMO.

The changes are over and done with, they were fairly simple to do AFAICT, and I think we can probably safely forget about this module for the next 2 or 3 years and eventually remove it when:
- no one, not even Ubuntu LTS or the other LTS suspects, is still using such old versions
- I can no longer find github search results for this module's use


Alternatively, if we actually have issues with it again then I could hear the argument that it's easier to remove it than to fix it. But we are not at that point today.

eli-schwartz:
Aside: is it possible to test Debian in CI somehow as we clearly didn't have complete coverage there? It's unfortunate that you're the only one who has to deal with Debian issues because you're the only one who tests it.

jpakkane:
You'd need to create a sid container and keep it up to date and it would still break every now and then as that is just the nature of Debian Sid. But it can be done.

Maybe it would be smarter to have a set of platforms that are run only weekly on trunk. It could include Debian Sid, Fedora Rawhide and whatnot.

eli-schwartz:
Yeah, running the CI only as a scheduled job on git master seems like the right balance given the idea is to check that unmodified Meson still passes all tests on updated OSes.

PR #10309, opened 2022-04-25
Windows: Improve Python 3.8+ module check on Windows

fanc999:
Hi,

From the commit message:

<i>On Python 3.8.x and later, if the imported module requires non-system DLLs that are not installed nor bundled with the module package, `os.add_dll_directory()` must be called on every path that contains the required DLLs, so that the module
can be imported successfully by Python.</i>

<i>Make things easier for people by calling `os.add_dll_directory()` on the valid directories in %PATH%, so that such module checks can be carried out successfully with much less manual intervention.</i>

Note that for running applications, application authors and/or packagers are still responsible for bundling the needed DLLs and/or using `os.add_dll_directory()` as needed.

With blessings, thank you.

### DIFF
--- a/mesonbuild/modules/python.py
+++ b/mesonbuild/modules/python.py
@@ -727,10 +727,22 @@ class PythonModule(ExtensionModule):
         found_modules: T.List[str] = []
         missing_modules: T.List[str] = []
         if python.found() and want_modules:
+            # Python 3.8.x or later require add_dll_directory() to be called on Windows if
+            # the needed modules require external DLLs that are not bundled with the modules.
+            # Simplify things by calling add_dll_directory() on the paths in %PATH%
+            add_paths_cmd = ''
+            if hasattr(os, 'add_dll_directory'):
+                add_paths_cmds = []
+                paths = os.environ['PATH'].split(os.pathsep)
+                for path in paths:
+                    if path != '' and os.path.isdir(path):
+                        add_paths_cmds.append(f'os.add_dll_directory({path!r})')
+                add_paths_cmd = 'import os;' + ';'.join(reversed(add_paths_cmds)) + ';'
+
             for mod in want_modules:
                 p, *_ = mesonlib.Popen_safe(
                     python.command +
-                    ['-c', f'import {mod}'])
+                    ['-c', f'{add_paths_cmd}import {mod}'])
                 if p.returncode != 0:
                     missing_modules.append(mod)
                 else:


mesonbuild/modules/python.py
```diff
@@ -727,10 +727,27 @@ def find_installation(self, state: 'ModuleState', args: T.Tuple[T.Optional[str]]
...
+            # Python 3.8.x or later require add_dll_directory() to be called on Windows if
+            # the needed modules require external DLLs that are not bundled with the modules.
+            # Simplify things by calling add_dll_directory() on the paths in %PATH%
+            add_paths_cmd = None
+            if hasattr(os, 'add_dll_directory'):
+                paths = os.environ['PATH'].split(os.pathsep)
+                add_paths_cmds = []
+                for path in paths:
+                    if path != '' and os.path.isdir(path):
+                        add_paths_cmds.append("os.add_dll_directory(r'%s')" % path.strip(os.sep))
```
eli-schwartz, line 739: Why does this need to have the path stripped of leading `\\` ?

Also:
```suggestion
                        add_paths_cmds.append('os.add_dll_directory({!r}).format(path.strip(os.sep)))
```

mesonbuild/modules/python.py
```diff
@@ -727,10 +727,27 @@ def find_installation(self, state: 'ModuleState', args: T.Tuple[T.Optional[str]]
...
             for mod in want_modules:
-                p, *_ = mesonlib.Popen_safe(
-                    python.command +
-                    ['-c', f'import {mod}'])
+                if add_paths_cmd is not None:
+                    p, *_ = mesonlib.Popen_safe(
+                        python.command +
+                        ['-c', f'import os;{add_paths_cmd};import {mod}'])
+                else:
+                    p, *_ = mesonlib.Popen_safe(
```
eli-schwartz, line 748: Rather than having add_paths_cmd be None or a string, why not have it be an empty string and then if there is nothing to add, nothing meaningful gets inserted into the command to run?

eli-schwartz (changes requested):

mesonbuild/modules/python.py
fanc999: Oh, yeah, UNC paths...

Thanks for the notes.

eli-schwartz: Okay. Why does it need to be stripped of trailing \\ either?

fanc999: Hi @eli-schwartz,

Sometimes people may have a trailing `\` in one or more of the directories that is included in `%PATH%` (you never know how people setup their `%PATH%` envvar), and that will cause trouble, one will get a Python error for an unescaped `\` when calling `os.add_dll_directory()`, so that `.rstrip(os.sep)` became needed.

eli-schwartz: This doesn't seem correct to me at all.

```python
>>> path = 'C:\\foo\\bar\\'
>>> print(path)
C:\foo\bar\
>>> func = 'print({!r})'.format(path)
>>> func
"print('C:\\\\foo\\\\bar\\\\')"
>>> subprocess.run[sys.executable, '-c', func])
C:\foo\bar\
CompletedProcess(args=['/usr/bin/python', '-c', "print('C:\\\\foo\\\\bar\\\\')"], returncode=0)
```

eli-schwartz: Although that is probably just a bug which my review suggestion to use `{!r}` has solved, since the repr of a string is guaranteed to be formatted correctly for reuse, very unlike the str() surrounded by quotes.

fanc999: OK, I will give it a shot.  Maybe 3-5 minutes?

mesonbuild/modules/python.py
```diff
@@ -727,10 +727,23 @@ def find_installation(self, state: 'ModuleState', args: T.Tuple[T.Optional[str]]
...
+            # Simplify things by calling add_dll_directory() on the paths in %PATH%
+            add_paths_cmd = ''
+            if hasattr(os, 'add_dll_directory'):
+                add_paths_cmds = []
+                paths = os.environ['PATH'].split(os.pathsep)
+                for path in paths:
+                    if path != '' and os.path.isdir(path):
+                        add_paths_cmds.append('os.add_dll_directory({!r})'.format(path))
+                add_paths_cmds.reverse()
+                add_paths_cmd = 'import os;' + ';'.join(add_paths_cmds) + ';'
```
eli-schwartz, line 741:
```suggestion
                add_paths_cmd = 'import os;' + ';'.join(reversed(add_paths_cmds)) + ';'
```

Or alternatively you might wish to do `for path in reversed(paths):`  up above.

mesonbuild/modules/python.py
```diff
@@ -727,10 +727,23 @@ def find_installation(self, state: 'ModuleState', args: T.Tuple[T.Optional[str]]
...
+            # Python 3.8.x or later require add_dll_directory() to be called on Windows if
+            # the needed modules require external DLLs that are not bundled with the modules.
+            # Simplify things by calling add_dll_directory() on the paths in %PATH%
+            add_paths_cmd = ''
+            if hasattr(os, 'add_dll_directory'):
+                add_paths_cmds = []
+                paths = os.environ['PATH'].split(os.pathsep)
+                for path in paths:
+                    if path != '' and os.path.isdir(path):
+                        add_paths_cmds.append('os.add_dll_directory({!r})'.format(path))
```
eli-schwartz, line 739: Since this is no longer a complex expression, this can become:

```suggestion
                        add_paths_cmds.append(f'os.add_dll_directory({path!r})')
```

eli-schwartz:
Okay, this PR looks correct to me from an implementation point of view.

I will leave it to others to determine whether this is something that "should" be done, since I'm not familiar enough with the problem to guess for myself. (Really, this whole PATH thing is terrible, Windows should learn how to do rpaths. And I'm completely lost on the ramifications of the python interaction here.)

fanc999:
Agreed on the `%PATH%` business, oh well... it's facts of life :)

PR #10323, opened 2022-04-27
compilers/c++: Add MSVC option to make the __cplusplus define accurate

eli-schwartz:
Otherwise it always returns the value for c++98, starting with MSVC 2017 15.7 or later. Earlier versions are not affected by this mis-feature

This was originally applied as 0b97d585480e973d8b149618901f7a4ddfa1a906 but later reverted because it made the CI red. Try it again, now.


xclaesse:
LGTM, if you get the ci green

eli-schwartz:
Correct.

There's a CI failure anyway, but I have a separate PR for that. It does solve the problem, but I don't know why, because it seems like it should be a mypy bug and not require code changes on our part...

PR #10324, opened 2022-04-27
another day, another mypy update becomes stricter about typing

eli-schwartz:
Move GlobalState to a runtime T.NamedTuple, use it for constructing the tuple we are passing around rather than expecting mypy to detect that the one we already have matches.

### DIFF
--- a/run_project_tests.py
+++ b/run_project_tests.py
@@ -82,20 +82,6 @@ if T.TYPE_CHECKING:
         no_unittests: bool
         only: T.List[str]
 
-    # In previous python versions the global variables are lost in ProcessPoolExecutor.
-    # So, we use this tuple to restore some of them
-    class GlobalState(T.NamedTuple):
-        compile_commands:   T.List[str]
-        clean_commands:     T.List[str]
-        test_commands:      T.List[str]
-        install_commands:   T.List[str]
-        uninstall_commands: T.List[str]
-
-        backend:      'Backend'
-        backend_flags: T.List[str]
-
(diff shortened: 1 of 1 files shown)


eli-schwartz:
(Why is this necessary?)

eli-schwartz:
I still really do not understand why this is necessary. mypy used to be able to tell that the tuple fulfilled the type contract of a NamedTuple's contents, but now it requires an honest to goodness *instance* of the NamedTuple.

@dcbaker any idea?

Is this expected mypy usage or a mypy bug or what?

bonzini:
> mypy used to be able to tell that the tuple fulfilled the type contract of a NamedTuple's contents

But it doesn't, you cannot do `compile_commands = state.compile_commands` if you pass a basic tuple. `T.NamedTuple` satisfies `T.Tuple1, but not vice versa.

eli-schwartz:
> you cannot do compile_commands = state.compile_commands if you pass a basic tuple

Am I missing the part where that's literally ever relevant?

bonzini:
It's always relevant to mypy. Static typing doesn't care if you use it, it doesn't like you passing a tuple because you *might* be using it.

It is just like you cannot pass a set for a `typing.List`, even if all that the function does is iterate on it. Either you relax the constraint in the callee or you use the expected type in the caller.

jpakkane:
I makes things go green so merging. If there are tweaks people want to do, feel free to open new PRs.

eli-schwartz:
@nirbheek in order to successfully pass CI for the point release, we need this too.

PR #10325, opened 2022-04-27
minit: refuse to allow creating broken projects

### LINKED ISSUE #10321: Exclude `test` as project name for `meson init`.
**Describe the bug**
Please see "To Reproduce"

**To Reproduce**
Just execute the following steps:
```
$ mkdir test; cd test
$ meson init
Using "test" (name of current directory) as project name.
Using "test" (project name) as name of executable to build.
Defaulting to generating a C language project.
Sample project created. To build it run the
following commands:

meson setup builddir
meson compile -C builddir

$ meson setup builddir
The Meson build system
Version: 0.62.1
Source dir: /home/gerion/test
Build dir: /home/gerion/test/builddir
Build type: native build
Project name: test
Project version: 0.1
C compiler for the host machine: ccache cc (gcc 10.2.1 "cc (Debian 10.2.1-6) 10.2.1 20210110")
C linker for the host machine: cc ld.bfd 2.35.2
Host machine cpu: x86_64

meson.build:5:0: ERROR: Target name 'test' is reserved for Meson's internal use. Please rename.

A full log can be found at /home/gerion/test/builddir/meson-logs/meson-log.txt
```

**Expected behavior**
`meson init` should rename the project if the folder is `test`.

**system parameters**
* what operating system: Debian 11
* what Python version are you using: 3.9.0
* what `meson --version`: 0.62.1

eli-schwartz:
Some executable names are invalid, and while it's unlikely anyone will specify such a thing using the --executable argument, it's not unlikely that people experimenting will attempt to use meson init in a directory named "test".

This then defaults to that for both the project name and the sample target name, and the latter produces errors when you try to build it.

Fixes #10321

### DIFF
--- a/mesonbuild/minit.py
+++ b/mesonbuild/minit.py
@@ -23,6 +23,7 @@ import os
 import re
 from glob import glob
 from mesonbuild import mesonlib
+from mesonbuild.coredata import FORBIDDEN_TARGET_NAMES
 from mesonbuild.environment import detect_ninja
 from mesonbuild.templates.samplefactory import sameple_generator
 import typing as T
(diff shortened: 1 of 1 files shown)


xclaesse:
LGTM

PR #10326, opened 2022-04-28
[skip ci] Setting Up Learning Module docs for Meson

michael-gene-brockus:
Got around to working on this PR, tutorial related pages are under the new root page **Learning Modules** not sure if you guys have any ideas on how to better name this page, also think that the learning content should probably be updated.


eli-schwartz:
I don't know what this is trying to do.

Also why does it take 12 commits to rename a few files? This seems wrong and should, on general principle, be squashed.

PR #10327, opened 2022-04-28
coverage: be clever and detect config files for gcovr/lcov

### LINKED ISSUE #3287: Allow excluding directories from coverage report

### LINKED ISSUE #4628: Support project-provided lcovrc script for lcov coverage
lcov has various options for excluding certain lines, directories and functions. In particular, I would like to be able to exclude certain function calls from branch coverage since they are assertions which it makes no sense to test. This uses the `lcov_excl_br_line` option ([example in GLib](https://gitlab.gnome.org/GNOME/glib/blob/master/.gitlab-ci/lcovrc)).

Since that kind of configuration is quite highly project dependent, and needs to be repeated in all the `lcov` commands, I think it might make sense to allow the project to provide an `lcovrc` file which would get passed to all the `coverage.py` `lcov` commands using `--config-file path/to/lcovrc`.

Failing that, it would be useful to be able to pass custom arguments to `lcov` so that projects could at least pass `--config-file` (or excl options) themselves.

eli-schwartz:
gcovr will read this file anyway, but if it exists we don't need to assume that the project wishes to exclude subprojects/ -- they can determine that themselves.

Fixes #3287
Closes #9761

lcov doesn't read the config file by default, but we can do the smart thing here.

Fixes #4628

### DIFF
--- a/docs/markdown/snippets/coverage_config_files.md
+++ b/docs/markdown/snippets/coverage_config_files.md
@@ -1,0 +1,12 @@
+## Coverage targets now respect tool config files
+
+gcovr >= 4.2 supports `gcovr.cfg` in the project source root to configure how
+coverage is generated. If Meson detects that gcovr will load this file, it no
+longer excludes the `subprojects/` directory from coverage. It's a good default
+for Meson to guess that projects want to ignore it, but not all projects prefer
+that and it is assumed that if a gcovr.cfg exists then it will manually
+include/exclude desired paths.
+
+lcov supports `.lcovrc`, but only as a systemwide or user setting. This is
+non-ideal for projects, so Meson will now detect one in the project source root
+and, if present, manually tell lcov to use it.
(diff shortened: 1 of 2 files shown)


eli-schwartz:
"150 reserved targets" is not testing coverage, it's only testing whether you can create custom_targe() output in directories with certain names that are also target names.

unittests/allplatformstests.py does have actual tests for the coverage target, but all it does is check that the output files are created. Checking that gcovr correctly uses the config file would involve somehow parsing the coverage output, which seems fraught with risk.

PR #10332, opened 2022-04-28
c_std, cpp_std: Change to a list of desired versions in preference order

xclaesse:
Projects that prefer GNU C but can fallback to ISO C can now set for
    example `default_options: 'c_std=gnu11,c11'` and it will use gnu11 when
    available, fallback to c11 otherwise. It is an error only if none of the
    values are supported by the current compiler.
    
    This allows to deprecate gnuXX values from MSVC compiler, that means
    that `default_options: 'c_std=gnu11'` will now print warning with MSVC
    but still fallback to 'c11' value. No warning is printed if at least one
    of the values is valid, i.e. `default_options: 'c_std=gnu11,c11'`.
    
    In the future that deprecation warning will become an hard error because
    `c_std=gnu11` should mean GNU is required, for projects that cannot be
    built with MSVC for example.

### DIFF
--- a/docs/markdown/Builtin-options.md
+++ b/docs/markdown/Builtin-options.md
@@ -296,6 +296,21 @@ is inherited from the main project. This is useful, for example, when the main
 project requires C++11, but a subproject requires C++14. The `cpp_std` value
 from the subproject's `default_options` is now respected.
 
+Since *1.3.0* `c_std` and `cpp_std` options now accept a list of values.
+Projects that prefer GNU C, but can fallback to ISO C, can now set, for
+example, `default_options: 'c_std=gnu11,c11'`, and it will use `gnu11` when
+available, but fallback to c11 otherwise. It is an error only if none of the
+values are supported by the current compiler.
+Likewise, a project that can take benefit of `c++17` but can still build with
+`c++11` can set `default_options: 'cpp_std=c++17,c++11'`.
+This allows us to deprecate `gnuXX` values from the MSVC compiler. That means
+that `default_options: 'c_std=gnu11'` will now print a warning with MSVC
+but fallback to `c11`. No warning is printed if at least one
+of the values is valid, i.e. `default_options: 'c_std=gnu11,c11'`.
+In the future that deprecation warning will become an hard error because
+`c_std=gnu11` should mean GNU is required, for projects that cannot be
+built with MSVC for example.
+
 ## Specifying options per machine
 
 Since *0.51.0*, some options are specified per machine rather than
--- a/docs/markdown/snippets/cstd.md
+++ b/docs/markdown/snippets/cstd.md
@@ -1,0 +1,18 @@
+## `c_std` and `cpp_std` options now accepts a list of values
+
+Projects that prefer GNU C, but can fallback to ISO C, can now set, for
+example, `default_options: 'c_std=gnu11,c11'`, and it will use `gnu11` when
+available, but fallback to `c11` otherwise. It is an error only if none of the
+values are supported by the current compiler.
+
+Likewise, a project that can take benefit of `c++17` but can still build with
+`c++11` can set `default_options: 'cpp_std=c++17,c++11'`.
+
+This allows us to deprecate `gnuXX` values from the MSVC compiler. That means
+that `default_options: 'c_std=gnu11'` will now print a warning with MSVC
+but fallback to `c11`. No warning is printed if at least one
+of the values is valid, i.e. `default_options: 'c_std=gnu11,c11'`.
+
+In the future that deprecation warning will become an hard error because
+`c_std=gnu11` should mean GNU is required, for projects that cannot be
+built with MSVC for example.
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -256,7 +256,8 @@ class UserArrayOption(UserOption[T.List[str]]):
         self.allow_dups = allow_dups
         self.set_value(value)
 
-    def listify(self, value: T.Union[str, T.List[str]]) -> T.List[str]:
+    @staticmethod
+    def listify_value(value: T.Union[str, T.List[str]], shlex_split_args: bool = False) -> T.List[str]:
         if isinstance(value, str):
             if value.startswith('['):
                 try:
@@ -266,7 +267,7 @@ class UserArrayOption(UserOption[T.List[str]]):
             elif value == '':
                 newvalue = []
             else:
-                if self.split_args:
+                if shlex_split_args:
                     newvalue = split_args(value)
                 else:
                     newvalue = [v.strip() for v in value.split(',')]
@@ -275,6 +276,9 @@ class UserArrayOption(UserOption[T.List[str]]):
         else:
             raise MesonException(f'"{value}" should be a string array, but it is not')
         return newvalue
+
+    def listify(self, value: T.Any) -> T.List[T.Any]:
+        return self.listify_value(value, self.split_args)
 
     def validate_value(self, value: T.Union[str, T.List[str]]) -> T.List[str]:
         newvalue = self.listify(value)
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -247,23 +247,16 @@ class UserComboOption(UserOption[str]):
 
 class UserArrayOption(UserOption[T.List[str]]):
     def __init__(self, description: str, value: T.Union[str, T.List[str]],
-                 split_args: bool = False, user_input: bool = False,
+                 split_args: bool = False,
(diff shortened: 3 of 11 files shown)


xclaesse:
@dcbaker @eli-schwartz @jpakkane I know there are been recurrent requests for more flexible way of defining the std. What do you think about this proposal? Are there use-cases not covered?

eli-schwartz:
The first thing I thought about when reading this is, it seems like it might break any meson.build that does `get_option('c_std')`.

On second look, it seems like at evaluation time it always yields whichever one gets selected in the end? Clever...

nirbheek:
I don't see any blockers for merging this? Other than resolving conflicts and fixing mypy.

dcbaker:
It conflicts pretty bad #11280… I think getting rid of the individual functions is going the wrong way

dcbaker:
But I think that this is the right language change, I’m just not sure about the implementation

eli-schwartz:
> It conflicts pretty bad #11280… I think getting rid of the individual functions is going the wrong way

And merged, so this needs fixing.

xclaesse:
Rebased.

tristan957:
@dcbaker looks like you had a concern here

xclaesse:
> @dcbaker looks like you had a concern here

I solved the conflicts and kept the functions in optinterpreter.

mesonbuild/coredata.py
```diff
@@ -258,6 +259,9 @@ def listify(self, value: T.Union[str, T.List[str]]) -> T.List[str]:
             raise MesonException(f'"{value}" should be a string array, but it is not')
         return newvalue
 
+    def listify(self, value: T.Any) -> T.List[T.Any]:
```
tristan957, line 262: Any reason to keep the non-static one around?

docs/markdown/Builtin-options.md
```diff
@@ -275,6 +275,21 @@ is inherited from the main project. This is useful, for example, when the main
 project requires C++11, but a subproject requires C++14. The `cpp_std` value
 from the subproject's `default_options` is now respected.
 
+Since *1.2.0* `c_std` and `cpp_std` options now accepts a list of values.
+Projects that prefer GNU C but can fallback to ISO C can now set for
+example `default_options: 'c_std=gnu11,c11'` and it will use gnu11 when
```
tristan957, line 280: > , but can fallback to ISO C, can now set, for example, `default_options: 'c_std=gnu11,c11'`, and it will use `gnu11` when available, but fallback to `c11` otherwise.

docs/markdown/Builtin-options.md
```diff
@@ -275,6 +275,21 @@ is inherited from the main project. This is useful, for example, when the main
...
+available, fallback to c11 otherwise. It is an error only if none of the
+values are supported by the current compiler.
+Likewise, a project that can take benefit of `c++17` but can still build with
+`c++11` can set `default_options: 'cpp_std=c++17,c++11'`.
+This allows to deprecate gnuXX values from MSVC compiler, that means
+that `default_options: 'c_std=gnu11'` will now print warning with MSVC
```
tristan957, line 286: print a warning

docs/markdown/Builtin-options.md
```diff
@@ -275,6 +275,21 @@ is inherited from the main project. This is useful, for example, when the main
...
+but still fallback to 'c11' value. No warning is printed if at least one
```
tristan957, line 287: fallback to `c11`.

docs/markdown/snippets/cstd.md
```diff
@@ -0,0 +1,18 @@
+## c_std and cpp_std options now accepts a list of values
+
+Projects that prefer GNU C but can fallback to ISO C can now set for
```
tristan957, line 3: Same comments for some of the paragraphs since they seem like copy pastes.

mesonbuild/coredata.py
```diff
@@ -302,6 +302,59 @@ def is_disabled(self) -> bool:
...
         return self.value == 'auto'
 
+class UserStdOption(UserComboOption):
+    '''
+    UserOption specific to c_std and cpp_std options. User can set a list of
+    STDs in preference order and it selects the first one supported by current
+    compiler.
+
+    For historical reasons, some compilers (msvc) allowed setting a GNU std and
+    silently fallback to C std. This is now deprecated, projects that support
```
tristan957, line 312: s/fallback/fell back

tristan957, line 312: Period after deprecated to prevent run-on sentence

mesonbuild/coredata.py
```diff
@@ -302,6 +302,59 @@ def is_disabled(self) -> bool:
...
+    both GNU and MSVC compilers should set e.g. c_std=gnu11,c11.
+
+    This is not using self.deprecated mechanism we already have for project
+    options because we want to print warning if ALL values are deprecated, not
```
tristan957, line 316: print a warning

mesonbuild/coredata.py
```diff
@@ -302,6 +302,59 @@ def is_disabled(self) -> bool:
...
+        for std in candidates:
+            if std in self.choices:
+                return std
+        # Fallback to a deprecated std if any
+        for std in candidates:
+            newstd = self.deprecated_stds.get(std)
+            if newstd is not None:
+                mlog.deprecation(
+                    f'None of the values {candidates} are supported by the {self.lang} compiler.\n' +
+                    f'However, the deprecated {std} std currently fallbacks to {newstd}.\n' +
```
tristan957, line 351: s/falls back/fallbacks

mesonbuild/coredata.py
xclaesse: Yes, it overrides the base class implementation. We do `opt.listify()` in `coredata.set_option()`. The reason I add a staticmethod version here is to be able to reuse that code elsewhere: It's used in mcompile and the next commit uses it in UserStdOption.

tristan957: Ok. Thanks for the explanation. I wonder if there is a better name for the static function, but as-is is also fine.

mesonbuild/optinterpreter.py
```diff
@@ -266,6 +266,11 @@ def integer_parser(self, description: str, args: T.Tuple[bool, _DEPRECATED_ARGS]
     def string_array_parser(self, description: str, args: T.Tuple[bool, _DEPRECATED_ARGS], kwargs: StringArrayArgs) -> coredata.UserOption:
         choices = kwargs['choices']
         value = kwargs['value'] if kwargs['value'] is not None else choices
+        if isinstance(value, str):
+            if value.startswith('['):
+                FeatureDeprecated('String value for array option', '1.2.0').use(self.subproject)
```
tristan957: 1.3.0

docs/markdown/Builtin-options.md
```diff
@@ -296,6 +296,21 @@ is inherited from the main project. This is useful, for example, when the main
...
+values are supported by the current compiler.
+Likewise, a project that can take benefit of `c++17` but can still build with
+`c++11` can set `default_options: 'cpp_std=c++17,c++11'`.
+This allows us to deprecate gnuXX values from the MSVC compiler. That means
+that `default_options: 'c_std=gnu11'` will now print a warning with MSVC
+but fallback to 'c11'. No warning is printed if at least one
+of the values is valid, i.e. `default_options: 'c_std=gnu11,c11'`.
+In the future that deprecation warning will become an hard error because
+`c_std=gnu11` should mean GNU is required, for projects that cannot be
+built with MSVC for example.
```
tristan957, line 312: Make sure that the monospace is consistent

mesonbuild/optinterpreter.py
xclaesse: Already done :)

docs/markdown/Builtin-options.md
xclaesse: What do you mean?

tristan957: you have c++17 in monospace, but c11 not in monospace for instance.

jpakkane:
The obvious question to this is whether this works only for this option or should it work for all options with choices?

I don't think it should work with anything except this, but OTOH it does add a behavioural inconsistency (though a mild and arguably good in any case).

xclaesse:
@jpakkane I don't think we can do it for project options. The trick here is Meson selects which of the choices is used, so get_option() only return one value, but for project option you would have to return all values because only the project can select... Which is no different from a regular array option.

I think it is a specific behavior for a very specific problem, I don't currently see how it could be generalized.

jpakkane:
I agree this should not be a generally available thing, but I suspect people are going to ask for it anyway, so it's better to be prepared.

PR #10335, opened 2022-04-30
Some documentation fix

soda92:
the old image doesn't match source.
the installation step is too complicated.


jpakkane:
The image fix is good, but the installation changes less so. We intentionally use official upstream installation instructions rather than thirdparty ones like Scoop (or vcpkg, or brew or ...).

Please remove the third commit and squash the two image changes into one. Thanks.

jpakkane:
You have merge commits in your code, which is something you [should never have in pull requests](https://mesonbuild.com/Contributing.html#keeping-pull-requests-up-to-date). You should use rebase instead and in addition fix the original commits instead of adding new commits. For this one I can fix it by squashing but in the future please keep your branch up to date properly, otherwise it will not be considered for merging.

PR #10339, opened 2022-05-01
compilers: remove opinionated c++ warning flag

eli-schwartz:
-Wnon-virtual-dtor is not what people think of as a standard warning flag. It was previously removed from -Wall in https://gcc.gnu.org/bugzilla/show_bug.cgi?id=16190 on the grounds that people didn't like it and were refusing to use -Wall at all because it forced this warning. Instead, it is enabled by -Weffc++ which is typically not enabled and even comes with GCC documentation warnings stating that the standard library doesn't obey it, and you might need to `grep -v` and filter out warnings. (!!!)

It doesn't fit into the typical semantics of Meson's warning_level option, which usually aligns with compiler standard warning levelsrather than a niche ideological warning level.

It was originally added in commit 22af56e05aa9cba4740d2ff303d876bb0c3cfb2b, but without any specific rationale included, and has gone unquestioned since then -- except by the Meson users who see it, assume there is a finely crafted design behind it, and quietly opt out by rolling their own warning options with `add_project_arguments('-Wall', ...)`.

/cc @kjellahl

### DIFF
--- a/docs/markdown/snippets/cxx_warn_args_dtor.md
+++ b/docs/markdown/snippets/cxx_warn_args_dtor.md
@@ -1,0 +1,3 @@
+## Flags removed from cpp/objcpp warning level 1
+
+`-Wnon-virtual-dtor` is no longer implied by `meson setup -Dwarning_level=1`.
--- a/mesonbuild/compilers/cpp.py
+++ b/mesonbuild/compilers/cpp.py
@@ -190,7 +190,7 @@ class ClangCPPCompiler(ClangCompiler, CPPCompiler):
         CPPCompiler.__init__(self, ccache, exelist, version, for_machine, is_cross,
                              info, exe_wrapper, linker=linker, full_version=full_version)
         ClangCompiler.__init__(self, defines)
-        default_warn_args = ['-Wall', '-Winvalid-pch', '-Wnon-virtual-dtor']
+        default_warn_args = ['-Wall', '-Winvalid-pch']
         self.warn_args = {'0': [],
                           '1': default_warn_args,
                           '2': default_warn_args + ['-Wextra'],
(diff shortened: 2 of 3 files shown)


jpakkane:
The reason I added it was that it is actually very useful as opposed to the rest of `-Weffc++`, which are inconvenient. IMHO it should be in `-Wall` but sadly isn't. There are a bunch of other such flags like `-Wshadow` but I did not add those at the time because I did not know that they existed.

eli-schwartz:
This heavily implies the question: what exactly *is* a warning_level as far as Meson is concerned?

dotnwat:
This was one of the first frustrating things I encountered with meson. The first was figuring out that `implicit_include_directories` was available. Now I'm having to explicit disable warnings `-Wno-non-virtual-dtor`. At the very least it would be nice to be able to turn off the opinionated warning flags.

dcbaker (approved):
We add Wno-non-virtual-dtor in mesa because of this, I think it makes sense to remove

jwakely:
> The reason I added it was that it is actually very useful

It really isn't though. `-Wdelete-non-virtual-dtor` catches all the real bugs (i.e. deleting a polymorphic type through a pointer to a base class without a virtual dtor). The `-Wnon-virtual-dtor` warning complains even if you never do that (e.g. because the destructor is deleted and can't ever be called!) It's broken by design and has been superseded by a much better option.

> as opposed to the rest of `-Weffc++`, which are inconvenient.

I agree with this part. I've been arguing for deprecating or removing `-Weffc++` from GCC for years.

> IMHO it should be in `-Wall` but sadly isn't.

You're welcome to that opinion, but it's not shared by G++ devs and I don't see why your opinion should be forced on your users as policy.

PR #10346, opened 2022-05-03
dependencies: handle one more case of subproject installed files

eli-schwartz:
Some projects treat meson.project_source_root() as the root of the dependency files, because the project itself merely wraps a bunch of datafiles. Our validation to make sure this doesn't point to another subproject, made use of pathlib.Path's generator for all component paths, which... did not include the path itself. So go ahead and explicitly check that too. Add a test case to verify it while we are at it.

Fixes https://github.com/mesonbuild/meson/pull/10103#issuecomment-1114901033


xclaesse:
It fixes the issue seen in gtk, LGTM.

xclaesse:
It fixes the issue seen in gtk, LGTM.

PR #10347, opened 2022-05-03
Preparatory work for C++ modules in GCC

jpakkane:
C++ modules with GCC can not be currently supported in Ninja due to:

https://github.com/ninja-build/ninja/issues/1962

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105467

These are just minor changes that are eventually going to be needed so get this in so I don't have to dangle this branch around.


jpakkane:
Mac failure is unrelated.

jpakkane:
Mac failure is unrelated.

PR #10349, opened 2022-05-03
wayland: Lookup for wayland-scanner using pkgconfig

xclaesse:
Just like some of glib tools, wayland-scanner can be defined in the
pkgconfig dependency variables. Share code between gnome and wayland
modules into ModuleState.

### DIFF
--- a/mesonbuild/modules/__init__.py
+++ b/mesonbuild/modules/__init__.py
@@ -20,14 +20,16 @@ import typing as T
 from .. import mesonlib
 from ..mesonlib import relpath, HoldableObject, MachineChoice
 from ..interpreterbase.decorators import noKwargs, noPosargs
+from ..programs import ExternalProgram
 
 if T.TYPE_CHECKING:
     from ..interpreter import Interpreter
     from ..interpreter.interpreterobjects import MachineHolder
     from ..interpreterbase import TYPE_var, TYPE_kwargs
-    from ..programs import ExternalProgram
+    from ..programs import OverrideProgram
     from ..wrap import WrapMode
-    from ..build import EnvironmentVariables
+    from ..build import EnvironmentVariables, Executable
+    from ..dependencies import Dependency
 
 class ModuleState:
     """Object passed to all module methods.
@@ -86,6 +88,38 @@ class ModuleState:
                      for_machine: MachineChoice = MachineChoice.HOST) -> 'ExternalProgram':
         return self._interpreter.find_program_impl(prog, required=required, version_func=version_func,
                                                    wanted=wanted, silent=silent, for_machine=for_machine)
+
+    def find_tool(self, name: str, depname: str, varname: str, required: bool = True,
+                  wanted: T.Optional[str] = None) -> T.Union['Executable', ExternalProgram, 'OverrideProgram']:
+        # Look in overrides in case it's built as subproject
+        progobj = self._interpreter.program_from_overrides([name], [])
+        if progobj is not None:
+            return progobj
+
+        # Look in machine file
+        prog_list = self.environment.lookup_binary_entry(MachineChoice.HOST, name)
+        if prog_list is not None:
+            return ExternalProgram.from_entry(name, prog_list)
+
+        # Check if pkgconfig has a variable
+        dep = self.dependency(depname, native=True, required=False, wanted=wanted)
+        if dep.found() and dep.type_name == 'pkgconfig':
+            value = dep.get_variable(pkgconfig=varname)
+            if value:
+                return ExternalProgram(name, [value])
+
+        # Normal program lookup
+        return self.find_program(name, required=required, wanted=wanted)
+
+    def dependency(self, depname: str, native: bool = False, required: bool = True,
+                   wanted: T.Optional[str] = None) -> 'Dependency':
+        kwargs = {'native': native, 'required': required}
+        if wanted:
+            kwargs['version'] = wanted
+        # FIXME: Even if we fix the function, mypy still can't figure out what's
+        # going on here. And we really dont want to call interpreter
+        # implementations of meson functions anyway.
+        return self._interpreter.func_dependency(self.current_node, [depname], kwargs) # type: ignore
 
     def test(self, args: T.Tuple[str, T.Union[build.Executable, build.Jar, 'ExternalProgram', mesonlib.File]],
              workdir: T.Optional[str] = None,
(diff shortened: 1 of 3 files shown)


xclaesse:
@eli-schwartz I thought this was merged already. Does it look good now?

mesonbuild/modules/unstable_wayland.py
```diff
@@ -47,10 +47,9 @@ def scan_xml(self, state, args, kwargs):
         if self.scanner_bin is None:
             # wayland-scanner from BUILD machine must have same version as wayland
             # libraries from HOST machine.
-            dep = self.interpreter.func_dependency(state.current_node, ['wayland-client'], {})
-            self.scanner_bin = state.find_program('wayland-scanner',
-                                                  for_machine=MachineChoice.BUILD,
```
eli-schwartz, line 52: This PR switches from using the BUILD version of the program, to not specifying that and thus falling back on the HOST version.

This may or may not be desirable... technically if there is some difference in the generated output between versions, dropping this is more reliable. The cost is that after merging this, in a cross-compile we will need to have an exe_wrapper.

It's very much not clear to me when we choose to look up a native or a host tool anyway, so :shrug:

mesonbuild/modules/gnome.py
```diff
@@ -311,36 +311,6 @@ def _print_gdbus_warning() -> None:
                      mlog.bold('https://github.com/mesonbuild/meson/issues/1387'),
                      once=True)
 
-    def _get_dep(self, state: 'ModuleState', depname: str, native: bool = False,
-                 required: bool = True) -> Dependency:
-        kwargs = {'native': native, 'required': required}
-        # FIXME: Even if we fix the function, mypy still can't figure out what's
-        # going on here. And we really dont want to call interpreter
-        # implementations of meson functions anyway.
```
eli-schwartz, line 319: We should probably preserve this comment when moving code around?

mesonbuild/modules/gnome.py
```diff
@@ -311,36 +311,6 @@ def _print_gdbus_warning() -> None:
...
-        return self.interpreter.func_dependency(state.current_node, [depname], kwargs)  # type: ignore
-
-    def _get_native_binary(self, state: 'ModuleState', name: str, depname: str,
-                           varname: str, required: bool = True) -> T.Union[ExternalProgram, OverrideProgram, 'build.Executable']:
-        # Look in overrides in case glib/gtk/etc are built as subproject
```
eli-schwartz, line 324: This comment is a bit self-explanatory, but so are the comments below. This specific one didn't get copied over.

mesonbuild/modules/gnome.py
```diff
@@ -311,36 +311,6 @@ def _print_gdbus_warning() -> None:
...
-    def _get_dep(self, state: 'ModuleState', depname: str, native: bool = False,
-                 required: bool = True) -> Dependency:
-        kwargs = {'native': native, 'required': required}
-        # FIXME: Even if we fix the function, mypy still can't figure out what's
-        # going on here. And we really dont want to call interpreter
-        # implementations of meson functions anyway.
-        return self.interpreter.func_dependency(state.current_node, [depname], kwargs)  # type: ignore
-
-    def _get_native_binary(self, state: 'ModuleState', name: str, depname: str,
-                           varname: str, required: bool = True) -> T.Union[ExternalProgram, OverrideProgram, 'build.Executable']:
```
eli-schwartz, line 323: In the modulestate function, we drop most of these from the return value type, and declare that it only returns ExternalProgram. This seems wrong; the function basically exists to possibly return one of the other two.

xclaesse: Indeed, copied state.find_program() but it was wrong too. Fixed both now.

mesonbuild/modules/gnome.py
xclaesse: Added back.

mesonbuild/modules/gnome.py
xclaesse: Added back.

mesonbuild/modules/unstable_wayland.py
xclaesse: The way it was done for GNOME tools is a bit of mix of both native and build machines:
1. Looks in cross file, not native file
2. Checks in native pkgconfig dep
3. Check in build machine PATH

Better be consistent in Wayland module.

mesonbuild/modules/__init__.py
```diff
@@ -83,10 +84,42 @@ def get_include_args(self, include_dirs: T.Iterable[T.Union[str, build.IncludeDi
     def find_program(self, prog: T.Union[str, T.List[str]], required: bool = True,
                      version_func: T.Optional[T.Callable[['ExternalProgram'], str]] = None,
                      wanted: T.Optional[str] = None, silent: bool = False,
-                     for_machine: MachineChoice = MachineChoice.HOST) -> 'ExternalProgram':
+                     for_machine: MachineChoice = MachineChoice.HOST) -> T.Union[ExternalProgram, OverrideProgram, 'build.Executable']:
```
eli-schwartz, line 87: build.Executable needs to be imported in the TYPE_CHECKING block.

xclaesse: That file is not checked by mypy and changing that cause a billion errors. I leave that to someone else who want to typecheck that module.

eli-schwartz: I don't care about mypy right now, but this is easy to fix, and what it fixes is this:

```
$ flake8 mesonbuild/modules/__init__.py 
mesonbuild/modules/__init__.py:67:70: F821 undefined name 'build'
mesonbuild/modules/__init__.py:87:99: F821 undefined name 'OverrideProgram'
mesonbuild/modules/__init__.py:87:116: F821 undefined name 'build'
mesonbuild/modules/__init__.py:92:79: F821 undefined name 'OverrideProgram'
mesonbuild/modules/__init__.py:92:96: F821 undefined name 'build'
mesonbuild/modules/__init__.py:123:47: F821 undefined name 'build'
mesonbuild/modules/__init__.py:123:65: F821 undefined name 'build'
mesonbuild/modules/__init__.py:126:38: F821 undefined name 'build'
mesonbuild/modules/__init__.py:126:58: F821 undefined name 'build'
mesonbuild/modules/__init__.py:222:62: F821 undefined name 'build'
mesonbuild/modules/__init__.py:225:54: F821 undefined name 'build'
```

Not all of these are your fault :) a bunch of them come from a2def550c586aeba4269588e79a1a308467f2582.

This is what happens when flake8 doesn't pass the entire project and we don't enforce it...

xclaesse: ah I see, should pass now.

PR #10350, opened 2022-05-04
interpreter: more feature methods

dcbaker:
I was looking at using features again in Mesa instead of Combos, and realized that we simply don't have the methods necissary for mesa's use casess. There are two specific shortfalls:

- no `enable_auto_if` method to compliement the `disable_auto_if`
- No way to error if `disable_auto_if` doesn't return a disabled object, other than to manually check and error.

Additionally, mesa would benefit from an option to force an auto to enabled or disabled in one call, since we have many features that are "auto on with !windows, auto off with windows" (or vice versa).

In testing this code I ran into the annoyance of writing failure tests, I would either have to resort to python unit tests, or I would have to write an individual test for each thing, or run the same test multiple times with different options. Instead, I opted to add a new module specifically for adding testing helpers. It is only available for import if an environment variable is set, and otherwise will raise an `ImportError`, and Meson will treat it just like a non-existant module. I've added a method here that can catch `MesonExceptions` and report that the exception with the required message was raised, which can then be passed to assert.

The functional changes are threefold:
1. Add a `strict` keyword argument to `disable_auto_if`, which causes an error to be raised if the returned option would not be `disabled`
2. Add an `enable_auto_if` method that works just like `disable_auto_if`, but in reverse.
3. Add a `force` method, which returns an `enabled` feature if passed `true` or a `disabled` feature if passed `false`, and always acts in `strict` mode.

I've tested this series with Mesa, and it allows me to remove a significant mount of `if error()` checking, and use features more and booleans less.


docs/markdown/snippets/feature_force.md
```diff
@@ -0,0 +1,17 @@
+## Add `FeatureOption.force` to force a feature to enabled or disabled
+
+Sometimes we want to force a feature to be enabled or disabled, take for example
+when a feature should be automatically turned off for one platform, but off for
+another:
+
+```meson
+feat = get_option('feat')
```
eli-schwartz, line 8: "feat: an achievement that requires great courage, skill, or strength."

Weird shortening if you ask me -- it derails my train of thought while trying to read this. Any reason not to just spell the whole thing out?

eli-schwartz (changes requested):
I'm not sure I understand these feature methods. 

> no `enable_auto_if` method to compliement the `disable_auto_if`

Fair enough.

> No way to error if `disable_auto_if` doesn't return a disabled object, other than to manually check and error.

So this is a feature option where it's (depending on platform?) not permitted for users to configure it as `=enabled`? It really isn't very verbose to manually check and error, and I suspect people actually using this are going to end up breaking the kwargs across multiple lines *anyway*...

This seems like it really only saves on the grammar overhead of `if / endif`, by letting you inline `if cond` as `strict: cond`. 

> Additionally, mesa would benefit from an option to force an auto to enabled or disabled in one call, since we have many features that are "auto on with !windows, auto off with windows" (or vice versa).

So if I'm reading this correctly, the value of the option defined by the user is completely irrelevant, there is only one valid state on Windows and that is "don't do the thing" and only one valid state elsewhere, which is "do the thing". It's outright deceptive to allow the user to specify anything, and the only reason you list the option in meson_options.txt at all is because that is the only way to create a feature object.

You don't want "auto on with !windows, auto off with windows", you want "use auto to seed a feature option that can be kludged into enabled with !windows, disabled with windows"


We have had discussion about a function to instantiate feature objects. e.g.

```meson
do_thing = feature_option(is_windows ? 'disabled' : 'enabled', name: 'thing')
```

And if that is the use case (user configuration is undesirable) then I strongly disagree with *any* proposed solution that involves encouraging projects to add it as user configuration to meson_options.txt

> Instead, I opted to add a new module specifically for adding testing helpers. It is only available for import if an environment variable is set

This makes me very nervous.

bonzini:
> no enable_auto_if method to compliement the disable_auto_if

What would you use it for? The typical use is to either use the feature as a `required` kwarg, or go from a feature to a boolean using `.allowed()`.

> No way to error if disable_auto_if doesn't return a disabled object, other than to manually check and error.

That's done with e.g. `.require(targetos == 'linux')` (either before or after the disable_auto_if).

> Additionally, mesa would benefit from an option to force an auto to enabled or disabled in one call, since we have many features that are "auto on with !windows, auto off with windows" (or vice versa).

As above, if a boolean is enough you can get it with `get_option('name').require(targetos == 'windows').allowed()` . A couple examples from QEMU:

    # Build tools by default if have_system is true
    have_tools = get_option('tools') \
      .disable_auto_if(not have_system) \
      .allowed()

    # Build qemu-ga by default if have_system or -Dtools=enabled, but only on Solaris/Linux/Windows
    have_ga = get_option('guest_agent') \
      .disable_auto_if(not have_system and not have_tools) \
      .require(targetos in ['sunos', 'linux', 'windows'],
               error_message: 'unsupported OS for QEMU guest agent') \
      .allowed()

    # Check dependency requirements that are stricter than the rest of QEMU
    dbus_display = get_option('dbus_display') \
      .require(gio.version().version_compare('>=2.64'),
               error_message: '-display dbus requires glib>=2.64') \
      .allowed()

bonzini:
```python
with_shared_glapi = get_option('shared-glapi') \
  .disable_auto_if(host_machine.system() == 'windows').allowed()
with_gles1 = get_option('gles1').require(not with_shared_glapi, error_message : ...).allowed()
with_gles2 = get_option('gles2').require(not with_shared_glapi, error_message : ...).allowed()
```

It should be without the "not" and well it makes perfect sense to me :) It's a 1-to-1 translation of your English description. The point of `require()` is to apply a... requirement, no matter if it's a necessity or an implementation detail.

The reason why I avoided enable_auto_if is that "enabled" means "error if not found", and that is a behavior that you will never get from auto. It seemed to me that people would think it is just disable_auto_if with a negated argument, but it's not.

dcbaker:
no, that won't get me what I want at all. I *need* an error if `disabled() == true`, `require` *does nothing* if the condition is true. so:
```meson
gles = get_option('gles1')
assert(gles1.disabled())
gles1.require(true)  # should error here, but does not, it just returns the original feature
```

dcbaker:
I always that it was strange to only have disable_auto_if() as well, there are cases where If you enable one feature, I need to enable a coresponding feature, or fail. like shared glapi, if you enable gles1 I *have* to enable shared-glapi or fail to configure, so I really need an `enable_auto_if`, because I really need to force a feature to be enabled.

dcbaker:
I think part of the disconnect @bonzini is that I always think of features as a boolean, but with a default value that can't be determined until configure time (in many cases as an artificial restriction in the meson_options.txt file). I really don't want to muck with `allowed`, I just want to get rid of `auto` and deal with exclusively `enabled` or `disabled` as quickly as possible.

bonzini:
I also find that `require` and `disable_auto_if` often end up being followed by `.allowed()`, but it was basically the price to pay for an API that is small and widely applicable.

The truth table of `require().allowed()`, for the Linux case with gles1 (E, A, D) on the side, the booleanized shapi (T, F) on the top and the outcome (T, F, error) in the middle is

```
     | T     F
-----+------------
Ena  | T     err
Auto | T     F
Dis  | F     F
```

What would you want instead?

eli-schwartz:
> I think part of the disconnect @bonzini is that I always think of features as a boolean, but with a default value that can't be determined until configure time

This doesn't really capture the nuance that "can't be determined until configure time" is what "boolean false" means in all contexts where feature options are valid.

bonzini:
Ok, I think I understand the difference. If shapi is disabled you want gles1=auto/gles2=auto to cause an error:

```
     | T     F
-----+------------
Ena  | T     err
Auto | T     err
Dis  | F     F
```

Why isn't it okay to disable them instead? That would be my expectation based on the definition of auto.

bonzini:
I don't understand if what you want is this?

```
     | T     F
-----+------------
Ena  | T     err
Auto | T     F
Dis  | err    F
```

Or this:

```
     | T     F
-----+------------
Ena  | T     err
Auto | T     err
Dis  | F      F
```

If the latter, I think you should just use require() and change the meaning of gles1=auto/shapi=disabled. For the former I guess it makes sense to add a force() method that takes a boolean + an error message, and returns a feature.

bonzini:
@dcbaker, ping. If you answer the question [here](https://github.com/mesonbuild/meson/pull/10350#issuecomment-1123243033) I'm happy to do the implementation.

PR #10355, opened 2022-05-06
Require CMake 3.14 for all tests that use the cmake module

smcv:
Older versions are not supported by the cmake module since 0.62.

This avoids having to hard-code the linux-bionic-gcc CI job as being
unable to run these tests, which leaves other older environments like
Debian 10 still trying to run them (and failing).

---

I was looking into this as part of a backport to a Debian-10-based environment. I ended up backporting CMake instead of disabling these test-cases, so I'm not actually using this patch; but it still seems maybe better to do a capability check, rather than hard-coding the knowledge that one specific CI runner has an outdated CMake?


eli-schwartz:
Is this still a draft?

smcv:
> Is this still a draft?

Only because I was waiting for CI. Now undrafted.

PR #10357, opened 2022-05-06
Make `meson wrap update` command update all wraps in parallel

xclaesse:


xclaesse:
See WrapDB PR https://github.com/mesonbuild/wrapdb/pull/425 to write `wrapdb_version` field in future releases.

xclaesse:
> Why are we moving wrapdb functionality from `meson wrap` which is intended to interact with the wrapdb, to `meson subprojects` which is not? I don't understand the goal here.

Because "meson wrap" command is super old and ugly code I want to ditch. More specifically, msubprojects.py already has all the infrastructure to update all wraps in parallel which speed up significantly when you have lots of wraps like gstreamer does.

eli-schwartz:
That has nothing to do with what I said, which is about the user-visible organization of the commands.

If you want to propose rewriting wraptool to "be less super old and ugly code", fine, that sounds fantastic. If you want to give `meson wrap update` an `--all` flag, that sounds reasonable enough too, and I don't object to making it operate in parallel even though I suspect it doesn't actually matter in practice for utterly tiny files.

> More specifically, msubprojects.py already has all the infrastructure

This is NOT an excuse to put functionality in the wrong place, instead it should be copied or refactored to be reusable and then plugged into `meson wrap update`.

We should not implement bad/inconsistent UX because "it's easier".

xclaesse:
It's pretty subjective to say it's inconsistent UX, could also just consolidated everything inside a single `subprojects` subcommand and consider `wrap` subcommand deprecated, that would be even more consistent IMHO.

xclaesse:
IMHO only `install` and `search` still make sense in the `wrap` subcommand, maybe I could just move them to subprojects and deprecate `wrap` subcommand completely?

jpakkane:
There is a clear conceptual difference between the two. `subprojects` is about subprojects in general and `wrap` is about dealing with the WrapDB in particular. Or at least that is how things should be. It would be good to keep these separate somehow.

eli-schwartz:
> It's pretty subjective to say it's inconsistent UX,

Subjectively it seems like jpakkane and I agree that there's a difference which this violates and makes confusing, so I have to reiterate -- why is it so important to you that `update-wrapdb` is a subcommand of `meson subprojects`?

Because from your initial wording it seemed like the answer is really just as simple as "I don't really care, this was just easier", potentially with a side serving of "now that I am called upon to defend this as a UX vision, XXX".

There are two things going on with this PR:
- make the act of updating wraps from the wrapdb be parallel
- make the act of updating wraps be entered from a different sub-tool

And I simply don't understand why the latter is required.

If nothing else, by separating out those concerns I think we would *all* agree on implementing the former, which fulfills your practical desires for gstreamer, while we could defer the latter to a hotly debated UX question in which you may or may not get another core committer to share your vision on.

At the moment I'm opposed to giving my LGTM on the former if requires me to give a LGTM on the latter. But I'd like to give my LGTM on the former.

xclaesse:
I still think it would make sense to consolidate all subprojects related commands into a single CLI, but I don't care enough to pick up a fight on that.

I updated this PR to make "meson wrap update" actually implemented into msubprojects.py to benefit from its infra for parallel. It's hackish but at least the UX is preserved.

tristan957:
> IMHO only `install` and `search` still make sense in the `wrap` subcommand, maybe I could just move them to subprojects and deprecate `wrap` subcommand completely?

Not that my opinion really matters, but I am failing to understand why this can't just be part of the `meson subprojects update` command. As a user, I want to update my subprojects, and for any subprojects from the wrapdb, please download the latest revision.

eli-schwartz:
> As a user, I want to update my subprojects, and for any subprojects from the wrapdb, please download the latest revision.

Other users don't want to mix "update the subproject to the most recent version specified in the currently available wrap" with "download new stable releases with new checksummed source code from the wrapdb".

Note that the current implementation of the `meson subprojects update` command is exclusively intended to be used for wrap-git, to:
- re-initialize it when switching from wrap-file
- for wraps that track an upstream branch e.g. master, do a git pull

So making it do a bunch of things it was never meant to do and in the wrong subcommand is obviously controversial.

tristan957:
I mean it sounds like `meson subprojects update` is meant to update subprojects. The fact that it currently only does the two functions you outlined, doesn't mean it couldn't be extended.

Whether it is the `wrong` subcommand is the entire point of the discussion! I don't think Meson should be in the business of pulling a whole new version from the wrapdb ala updating xxhash from 0.8.0 to 0.8.1. Just increment the revision of the wrap. Just my 2 cents.

When I think about updating dependencies using other tooling, there is `yarn upgrade`, `poetry upgrade`, etc. All are just a single command. Seems wrong for Meson to require two commands to update subprojects. `meson subprojects update && meson wrap update`. Poor UX in my opinion.

eli-schwartz:
> I don't think Meson should be in the business of pulling a whole new version from the wrapdb ala updating xxhash from 0.8.0 to 0.8.1. Just increment the revision of the wrap. Just my 2 cents.

So you want a `wrap update` command, but one which only updates if the `tag, rev = upstream_version.split('-')` has a matching tag and an updated rev?

That would definitely be functionality that doesn't currently exist in Meson at all.

xclaesse:
@tristan957 taking GStreamer use-case here: We commit into gstreamer git repository every .wrap files we need, including some that comes from WrapDB. All those wraps are pinned to a specific version we tested, so we want people to use that version.

When user does git pull in the gstreamer repository, it could pull updates for those wraps. In that case user needs to also upgrade their subprojects checkouts to the version specified in the wrap file, but NOT to whatever version is latest in WrapDB. For that use-case, users needs to run `meson subprojects update`. Note that in addition to users, our CI also runs that command, because we cache subprojects checkouts between runs to avoid downloading from scratch every time, so `meson subprojects update --reset` guarantees to get the version specified in the .wrap file, but we don't want to also pull random version from WrapDB at the same time.

Now as a GStreamer maintainer I want to be able to pull latest wraps from wrapdb, test that, and if all is good, commit them into git. For that I use `meson wrap update`.

xclaesse:
@eli-schwartz @jpakkane does it look good now?

xclaesse:
ping?

xclaesse:
@eli-schwartz @jpakkane kind reminder to review to PR please :)

xclaesse:
ping please?

xclaesse:
Any object against this PR? :pray:

jpakkane:
Conceptually I like this. I haven't looked at the implementation because of a death flu.

PR #10359, opened 2022-05-06
is_dragonflybsd() returns 'dragonflybsd'

Gio:
A small change that makes it easier to sort out BSD operating systems; as it stands currently, DragonFlyBSD is the only outlier on the reference table.

### DIFF
--- a/docs/markdown/Reference-tables.md
+++ b/docs/markdown/Reference-tables.md
@@ -136,7 +136,7 @@ These are provided by the `.system()` method call.
 | android             | By convention only, subject to change |
 | cygwin              | The Cygwin environment for Windows |
 | darwin              | Either OSX or iOS |
-| dragonfly           | DragonFly BSD |
+| dragonflybsd        | DragonFly BSD |
 | emscripten          | Emscripten's Javascript environment |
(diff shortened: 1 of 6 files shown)


eli-schwartz:
Unfortunately I am nervous about this change because changing it means that if anyone currently checks for dragonfly using the current logic, then they need it to return the version without "bsd" at the end or suffer failure to match.

jpakkane:
We have also committed to the fact that system names are immutable. This is why `darwin` is still a value rather than `macos`, `ios` et al.

eli-schwartz:
One possibility would be if we had a `.system_family()` that returns "bsd" for all BSD's including "dragonfly". It would probably return the same thing as `.system()` everywhere else, but there may be other systems that could be logically grouped together? Not sure.

Gio:
> One possibility would be if we had a `.system_family()` that returns "bsd" for all BSD's including "dragonfly". It would probably return the same thing as `.system()` everywhere else, but there may be other systems that could be logically grouped together? Not sure.

`.system_family()` would certainly simplify the BSDs, and might be useful in the future. I'll close this PR and open an issue instead.

PR #10361, opened 2022-05-06
test cases: use best practices method to find the python3 program

eli-schwartz:
We do not need the python module's find_installation() for this, as this does various things to set up building and installing python modules (pure python and C-API). This functionality is already tested in the python tests.

Elsewhere, when we just need an interpreter capable of running python scripts in order to guarantee a useful scripting language for custom commands, it suffices to use find_program(), which does not run an introspection script or do module imports, and is thus faster and a bit cleaner.

Either way, both methods are guaranteed to find the python3 interpreter, deferring to mesonlib.python_command for that guarantee.

### DIFF
--- a/docs/markdown/Generating-sources.md
+++ b/docs/markdown/Generating-sources.md
@@ -53,7 +53,7 @@ is generated and that the proper include paths are created for the
 target:
 
 ```meson
-prog_python = import('python').find_installation('python3')
+prog_python = [[#find_program]]('python3')
 
 foo_c = custom_target(
     'foo.c',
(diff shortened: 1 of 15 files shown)


jpakkane:
```
E               AssertionError: '    existing prog  : D:\\a\\_temp\\msys64\\mingw32\\bin\\python3.EXE' != '    existing prog  : D:\\a\\_temp\\msys64\\mingw32\\bin\\python3.exe'
```

eli-schwartz:
I really do not know what the difference is between uppercase and lowercase `.exe` extensions or why sys.executable is one and find_program_impl() is the other. :/

jpakkane:
Well a simple `.lower()` call when on Windows should fix it.

xclaesse:
@eli-schwartz it is also wrongly used in `docs/markdown/Generating-sources.md` I think.

PR #10369, opened 2022-05-08
python module: default extensions to hidden symbol visibility

eli-schwartz:
python compiled extensions should never need to expose any symbol other than `PyInit_*` which is declared with default visibility via `PyMODINIT_FUNC` on supported compilers.

Thus, a reasonably sane default is to mark any other symbols as hidden, while still respecting any manually specified visibility.

gate this on the version of python itself, as not all versions decorate `PyMODINIT_FUNC` properly.

/cc @rgommers

### DIFF
--- a/docs/markdown/Python-module.md
+++ b/docs/markdown/Python-module.md
@@ -93,6 +93,12 @@ the addition of the following:
   it will be appended to that location. This keyword argument is
   mutually exclusive with `install_dir`
 
+Additionally, the following diverge from [[shared_module]]'s default behavior:
+
+- `gnu_symbol_visibility`: if unset, it will default to `'hidden'` on versions
+  of Python that support this (the python headers define `PyMODINIT_FUNC` has
+  default visibility).
+
 `extension_module` does not add any dependencies to the library so
 user may need to add `dependencies : py_installation.dependency()`,
 see [[dependency]].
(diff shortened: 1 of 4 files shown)


xclaesse:
LGTM.

rgommers:
LGTM too, thanks @eli-schwartz. It did uncover an issue (see https://github.com/scipy/scipy/pull/16165), but that's a good thing I'd say.

eli-schwartz:
Huh. That's also the thing which marks the init function via declspec on MSVC as `__declspec(dllexport)`, and since MSVC defaults to the equivalent of what this PR did for GCC/clang...

... I would expect that failure to use `PyMODINIT_FUNC` would break the build on MSVC by having it not provide the one symbol that *is* supposed to be exported.

Interesting to note: exactly one of the 19 files you fixed, manually coded this (reimplementing PyMODINIT_FUNC but without cygwin support and without support for unix compilers other than gcc/clang):
```c

#if defined(WIN32) || defined(_WIN32)
#  define MODULE_EXPORT __declspec(dllexport)
#else
#  define MODULE_EXPORT __attribute__((visibility("default")))
#endif

extern "C" MODULE_EXPORT PyObject * PyInit__uarray(void) {
```

eli-schwartz:
@nirbheek

The second commit here is a new feature and cannot be backported.

The first commit fixes a possible compiler warning and may be worth backporting. Thoughts?

nirbheek:
> The first commit fixes a possible compiler warning and may be worth backporting. Thoughts?

I agree.

PR #10370, opened 2022-05-09
dependencies/boost.py: ignore unknown files instead of erroring out on them

### LINKED ISSUE #8325: PDB files shipped with Boost/MSVC releases from sourceforge trigger ERROR when using dependency('boost')
Binary Boost releaases downloaded from https://sourceforge.net/projects/boost/files/boost-binaries/ include debbuging information in separete .pdb files stored alongside .lib and .dll files in `lib<32|64>-msvc-<version>` directories. 

The error visible in the log looks like this:

`meson.build:13:0: ERROR: Unable to process library extension "pdb" (C:\opt\moonlight-deps\boost_1_75_0\lib32-msvc-14.1\boost_atomic-vc141-mt-gd-x32-1_75.pdb)`

which hints at the following code location:

https://github.com/mesonbuild/meson/blob/2a8d6690f73ec6415c2ee592e856b3b6de7cfb02/mesonbuild/dependencies/boost.py#L155

I think the official binary releases for MSVC should be consumable in the form they are shipped.
Workaround: currently I need to move .pdb files to separate directory, which requires write access to the lib... directory - not always available on CI systems with boost preinstalled.

fghzxm:
Fixes #8325.

### DIFF
--- a/mesonbuild/dependencies/boost.py
+++ b/mesonbuild/dependencies/boost.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import re
+import dataclasses
 import functools
 import typing as T
 from pathlib import Path
@@ -83,6 +84,10 @@ if T.TYPE_CHECKING:
 #     2.4 Ensure that all libraries have the same boost tag (and are thus compatible)
 #   3. Select the libraries matching the requested modules
 
+@dataclasses.dataclass(eq=False, order=False)
+class UnknownFileException(Exception):
+    path: Path
+
 @functools.total_ordering
 class BoostIncludeDir():
     def __init__(self, path: Path, version_int: int):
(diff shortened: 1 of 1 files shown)


mesonbuild/dependencies/boost.py
```diff
@@ -152,7 +159,7 @@ def __init__(self, path: Path):
         elif self.nvsuffix in ['a', 'lib']:
             self.static = True
         else:
-            raise DependencyException(f'Unable to process library extension "{self.nvsuffix}" ({self.path})')
+            raise UnknownFileException(self.path)
```
mensinda, line 162: Also, there should be a warning (or at least a debug log) here to indicate that there was an unexpected file. This way, debugging further issues will be easier since there is a specific reason printed in the `meson-log.txt`.

mensinda (changes requested):
First of all, thanks for working on this, but I am not 100% happy with the current solution:

My main problem is that it ignores **all** unexpected files, instead of just ignoring the ones we know we don't need. In short, I would prefer if there was a specific check to silently ignore `PDB` files, but warn for other unexpected files.

mensinda (approved):
LGTM assuming the CI is green.

Also, could you please squash your two commits into a single commit to keep the git history clean?

mensinda (approved):

fghzxm:
Why am I getting these errors about nonexistent headers and identifiers?

PR #10375, opened 2022-05-09
docs: simplify SimpleStart apt commands

Tachi107:
- Install `g++` instead of `build-essential`. Citing [`build-essential`](https://packages.debian.org/stable/build-essential)'s description: _"If you do not plan to build Debian packages, you don't need this package"_. `g++` pulls in everything that is required to build C and C++ programs, while `build-essential` installs unneeded stuff like `make` and `dpkg-dev`.

- Remove explicit `ninja-build` when installing `meson`. The `meson` package also installs `ninja-build`, so specifying it isn't needed (just like in the `pacman` command)

### DIFF
--- a/docs/markdown/SimpleStart.md
+++ b/docs/markdown/SimpleStart.md
@@ -28,7 +28,7 @@ All Linux distributions provide easy access to development tools.
 Typically you need to open a terminal and execute one command, which
 depends on your distro.
 
- - Debian, Ubuntu and derivatives: `sudo apt install build-essential`
+ - Debian, Ubuntu and derivatives: `sudo apt install g++`
  - Fedora, Centos, RHEL and derivatives: `sudo dnf install gcc-c++`
  - Arch: `sudo pacman -S gcc`
 
@@ -66,7 +66,7 @@ install more files that are needed for compilation.
 
 Installing Meson is just as simple as installing the compiler toolchain.
 
- - Debian, Ubuntu and derivatives: `sudo apt install meson ninja-build`
+ - Debian, Ubuntu and derivatives: `sudo apt install meson`
  - Fedora, Centos, RHEL and derivatives: `sudo dnf install meson ninja-build`
  - Arch: `sudo pacman -S meson`
 


Tachi107:
libc6-dev is pulled in by gcc (only recommended) and g++ (g++ depends on libstdc++, that depends on libc6-dev)

Tachi107:
> In this case, apparently, somewhere 5 dependencies deep into the dependency chain for build-essential it already depends on libc6-dev -- but only via g++, if you install gcc instead then it doesn't??? No, recommends is not good enough. And if we document to install g++ instead of build-essential, people _**will**_ replace that with gcc if they think that's all they need, and then be missing headers.

Then the guide should say `apt install g++ libc-dev`, installing `build-essential` is still not right. If Debian moves to not requiring a C++ compiler to compile its packages then the guide will contain wrong instructions (and Debian is free to do whatever it wants to its own internal packages). If you need to install a C or C++ compiler and the standard library, well, you should install a compiler and the library. And by the way, for a "SimpleStart" guide, Recommends is enough. Recommended packages gets automatically installed unless you explicitly block their installation (and there are scenarios where not wanting the standard C library might be wanted). Still, to be sure that users with broken systems might run into issues, explicitly listing `libc-dev` (or `libc6-dev`) in the documented command could be appropriate.

> Of course this is basically impossible to find out by reading the website index. Debian's package index website only shows what a package depends on and not which other packages depend on it, so it's completely impossible to ask the question "libc6-dev, how can I be sure I have it? Which other packages require it and thus guarantee it is installed?"

Yeah I'm not aware of any web interface that does it, but if you have access to a Debian machine the command is `apt rdepends`.

> If debian actually expects people to not use `build-essential` as a user-friendly virtual package for "the compiler plus whatever the compiler should depend on but doesn't", then my formal recommendation to the debian policy group is to rename it to "dh-build-essential" and make a new build-essential package that _**guarantees**_ that basic dependencies of the compiler are actually installed, but doesn't include dpkg-dev.

"dh" is a tool that helps building Debian packages, but the policy isn't really related to it. If you want a working C compiler just install the compiler, "build-essential" doesn't imply C anywhere (even though `deb-build-essential` would make it more obvious).

> people unfortunate enough to be Debian users

It's not that bad after all :)

jpakkane:
A problem of sorts with this change is that sooner or later people will start sending merge requests to change the command to something else to "minimize install sizes" or "so you don't install a C++ compiler when you don't need it" and so on. The change itself is fine by me, but the text should explain that we are using `g++` as a simple shorthand that pulls in the necessary packages for even for plain C development.

eli-schwartz:
> And by the way, for a "SimpleStart" guide, Recommends is enough. Recommended packages gets automatically installed unless you explicitly block their installation (and there are scenarios where not wanting the standard C library might be wanted). Still, to be sure that users with broken systems might run into issues, explicitly listing `libc-dev` (or `libc6-dev`) in the documented command could be appropriate.

And then we end up having to keep track of how Debian currently breaks the C compiler and document this, which is pain and sadness and grief.

Also: NO IT IS NOT enough. It is not enough for the SimpleStart guide to *Meson* to assume that people haven't configured apt 8 months ago to not install unrequired recommendations. Or that people aren't, say, using docker.

Perhaps if this was the SimpleStart guide to *Debian* that would be a reasonable assumption (but see again, docker).

Debian and their YOLO "it's in recommends, that should be good enough for everyone to have a baseline functional package"...

> Yeah I'm not aware of any web interface that does it, but if you have access to a Debian machine the command is `apt rdepends`.

I don't have access to a Debian machine, I'm aware that rdepends exists.

> "dh" is a tool that helps building Debian packages, but the policy isn't really related to it. If you want a working C compiler just install the compiler, "build-essential" doesn't imply C anywhere (even though `deb-build-essential` would make it more obvious).

Semantics, whatever. This is a meta-package that helps build debian packages. :shrug:

And again no, you cannot just install the compiler, the compiler might be broken out of the box because "who needs the headers for libc in order to build programs using libc".

For that matter, I'm shocked that the C++ compiler depends on libstdc++ (and dev), surely someone might want to use `g++` with a custom spec file and their own standard library. Or just use libc++. This is completely inconsistent with `gcc` and surely a bug in the gcc-10 source package. 

In contrast, Debian is much less likely to break their own internal processes i.e. building debs. Fedora aside, there are many, many people out there using build-essential to build... debs for private repositories.

Apparently the maintainer of build-essential considers it irrationally dangerous to assume that merely having g++ is sufficient to have libc6-dev as a transitive recursive dependency, even though it currently is one, and feels the need to manually specify it in order to reduce risk to people building debs.

I'd be willing to bet good money that in the not unlikely scenario where g++-10 drops libstdc++ to a recommends for the sake of people using `-nostdinc++`, build-essential will coordinate with that change and ensure it is installed regardless.

> It's not that bad after all :)

It's worse. Shall we discuss how Debian breaks `python3` in ways that you cannot fix ever under any circumstances, no matter how many additional packages you install?

It's so bad that elements of the Python community want to pull a Mozilla and send lawyers to forbid Debian to use the name "Python" in order to prevent damage to the python.org reputation.

Some, but not all, of those badnesses affect Meson e.g. our janky hacks to make `import('python')` halfway functional on Debian.

Tachi107:
> A problem of sorts with this change is that sooner or later people will start sending merge requests to change the command to something else to "minimize install sizes" or "so you don't install a C++ compiler when you don't need it" and so on. The change itself is fine by me, but the text should explain that we are using `g++` as a simple shorthand that pulls in the necessary packages for even for plain C development.

But this is already done by the Fedora command, should that be addressed as well?

> In contrast, Debian is much less likely to break their own internal processes i.e. building debs. Fedora aside, there are many, many people out there using build-essential to build... debs for private repositories.

I wouldn't rely on that.

> It's worse. Shall we discuss how Debian breaks `python3` in ways that you cannot fix ever under any circumstances, no matter how many additional packages you install?
> 
> It's so bad that elements of the Python community want to pull a Mozilla and send lawyers to forbid Debian to use the name "Python" in order to prevent damage to the python.org reputation.

Oh wow, how can it be _this_ bad? And how is it not fixed yet?

Anyway, I still think that if you want to be sure that you can compile a C/C++ program you should tell users to install the compiler and libc... Relying on whatever Debian uses to build packages doesn't seem right in my opinion. But of course it's not a big deal at all, it just bugs me that the `apt` line is inconsistent with the other two

Tachi107:
Closing as not that important and (surprisingly :)) controversial

Issue #10366, opened 2022-05-08
EFI Application Windows Subsystem


LensPlaysGames:
For some reason only two of the many windows subsystems are allowed through. This disallows me from building a UEFI bootloader application using meson, as it FORCES either console or windows subsystem when attempting to use a MinGW compiler to generate UEFI-compatible PE32+ executables.

https://github.com/mesonbuild/meson/blob/c16fdaeecafe94461d6048b90bd337a04c25ad3c/mesonbuild/linkers/linkers.py#L702-L712

---

### Simple fix

```python
elif 'efi_application' in value: 
         args = ['--subsystem,10'] 
```

---

### Non-simple fix

https://github.com/mesonbuild/meson/blob/ca77adea89b3f102ca3e62465f651811516071f5/mesonbuild/build.py#L1260-L1264

The regexp above could take into account straight subsystem numbers, although that would definitely allow for things to break (which seems fine if you are messing with the windows subsystem, imo).

---

I'll probably end up making a pull request once I figure out how to build/use meson from source.

Tachi107:
> build/use meson from source

It should be as easy as editing the source files and running running [`path/to/meson.py`](https://github.com/mesonbuild/meson/blob/master/meson.py)

eli-schwartz:
https://mesonbuild.com/Getting-meson.html#downloading-meson

We document that running from source in the manner @Tachi107 described shall work, and it's tested in CI too (in addition to being the primary way I run meson myself).

PR #10378, opened 2022-05-09
linkers/MinGW: Allow `efi_application` value for `win_subsystem`

LensPlaysGames:
Previously, any value other than `windows` or `console` caused an exception. 
This change allows for `efi_application` to be passed as the `win_subsystem` to MinGW based linkers.

This means Meson can now be used to build EFI applications that can be run from a UEFI shell.

I've `grep`ed the whole repository for "subsystem", and I didn't find any docs that needed updated, or any other places that 
might need changed. Please let me know if there's anything I'm missing.

fixes https://github.com/mesonbuild/meson/issues/10366

### DIFF
--- a/mesonbuild/linkers/linkers.py
+++ b/mesonbuild/linkers/linkers.py
@@ -700,14 +700,32 @@ class GnuLikeDynamicLinkerMixin:
         return (args, rpath_dirs_to_remove)
 
     def get_win_subsystem_args(self, value: str) -> T.List[str]:
-        if 'windows' in value:
-            args = ['--subsystem,windows']
-        elif 'console' in value:
-            args = ['--subsystem,console']
-        else:
-            raise MesonException(f'Only "windows" and "console" are supported for win_subsystem with MinGW, not "{value}".')
+        # MinGW only directly supports a couple of the possible
+        # PE application types. The raw integer works as an argument
+        # as well, and is always accepted, so we manually map the
+        # other types here. List of all types:
+        # https://github.com/wine-mirror/wine/blob/3ded60bd1654dc689d24a23305f4a93acce3a6f2/include/winnt.h#L2492-L2507
+        subsystems = {
+            "native": "1",
+            "windows": "windows",
+            "console": "console",
+            "posix": "7",
+            "efi_application": "10",
+            "efi_boot_service_driver": "11",
+            "efi_runtime_driver": "12",
+            "efi_rom": "13",
+            "boot_application": "16",
+        }
+        versionsuffix = None
         if ',' in value:
-            args[-1] = args[-1] + ':' + value.split(',')[1]
+            value, versionsuffix = value.split(',', 1)
+        newvalue = subsystems.get(value)
+        if newvalue is not None:
+            if versionsuffix is not None:
+                newvalue += f':{versionsuffix}'
+            args = [f'--subsystem,{newvalue}']
+        else:
+            raise mesonlib.MesonBugException(f'win_subsystem: {value!r} not handled in MinGW linker. This should not be possible.')
 
         return self._apply_prefix(args)
 


mesonbuild/linkers/linkers.py
```diff
@@ -704,8 +704,10 @@ def get_win_subsystem_args(self, value: str) -> T.List[str]:
             args = ['--subsystem,windows']
         elif 'console' in value:
             args = ['--subsystem,console']
+        elif 'efi_application' in value:
+            args = ['--subsystem,10']
```
LensPlaysGames, line 708: It looks like `objcopy` allows a subsystem of `efi-app`, but `ld` doesn't share the same capability, so we have to use the number directly.

Tachi107: A code comment would make it easier to understand without having to look at this PR

Tachi107:
> I've `grep`ed the whole repository for "subsystem", and I didn't find any docs that needed updated, or any other places that might need changed. Please let me know if there's anything I'm missing.

I honestly have almost no idea how these Windows subsystems work, but isn't the `win_subsystem` kwarg of [`build_target()`](https://mesonbuild.com/Reference-manual_functions.html#build_target) related?

mesonbuild/linkers/linkers.py
eli-schwartz: Does `ld` allow any direct numeric value?

Perhaps the correct solution would be to have a mapping table here, something like

```python
    def get_win_subsystem_args(self, value: str) -> T.List[str]:
        subsystems = {
            "windows": "windows",
            "console": "console",
            "efi_application": "10",
            # ....
        }
        newvalue = subsystems.get(value)
        if newvalue is not None:
            args = [f'--subsystem,{newvalue}']
        else:
            raise MesonException(f'Only {set(subsystems)} are supported for win_subsystem with MinGW, not "{value}".')
            ## this prints:
            ### Only {'efi_application', 'console', 'windows'} are supported for win_subsystem with MinGW, not "{value}".
            ## joining with "," except for the last item which shall be joined with the word "and" isn't conveniently done, hmm
        if ',' in value:
            args[-1] = args[-1] + ':' + value.split(',')[1]

        return self._apply_prefix(args)
```

eli-schwartz:
> I honestly have almost no idea how these Windows subsystems work, but isn't the `win_subsystem` kwarg of [`build_target()`](https://mesonbuild.com/Reference-manual_functions.html#build_target) related?

It's not related... it *is* what is being modified here. :)

> Typical values include console for text mode programs and windows for gui apps.

This is not documentation on "allowable" values.

mesonbuild/linkers/linkers.py
eli-schwartz: On second thought, this is incomplete... value -> newvalue would need to be split on "," and joined later.

LensPlaysGames:
> > I've `grep`ed the whole repository for "subsystem", and I didn't find any docs that needed updated, or any other places that might need changed. Please let me know if there's anything I'm missing.
> 
> I honestly have almost no idea how these Windows subsystems work, but isn't the `win_subsystem` kwarg of [`build_target()`](https://mesonbuild.com/Reference-manual_functions.html#build_target) related?

I'm brand new to contributing, so I'm not sure what build_target is or how it's used, but I'll take a look.

It seems to me like `build_target.process_kwargs` already accepts a value of `efi_application` for the `win_subsystem` kwarg by means of `validate_win_subsystem()` (when it is an executable).
- [`win_subsystem` is set here](https://github.com/mesonbuild/meson/blob/master/mesonbuild/build.py#L1184)
- [`validate_win_subsystem()` definition](https://github.com/mesonbuild/meson/blob/master/mesonbuild/build.py#L1260-L1264)

The MinGW linker implementation reacts to the kwarg differently, but the kwarg is already present and validated.

It seems to me like no change is necessary, but please let me know if I am mistaken. 

With the current changes, I have a Meson build system generating working EFI applications tested in three virtual machines: QEMU + VirtualBox + VMWare Workstation Player emulating x86_64 CPUs.

mesonbuild/linkers/linkers.py
LensPlaysGames: > Does `ld` allow any direct numeric value?

It appears to be based on whether `ld` is targeting `PE` or not (search for `--subsystem` in the [`ld` man page](https://www.man7.org/linux/man-pages/man1/ld.1.html)).

LensPlaysGames: > On second thought, this is incomplete... value -> newvalue would need to be split on "," and joined later.

Is this acceptable? If so, I'll commit it to this PR.

```python
def get_win_subsystem_args(self, value: str) -> T.List[str]:
        subsystems = {
            "windows": "windows",
            "console": "console",
            "efi_application": "10",
            # ....
        }
        versionsuffix = None
        if ',' in value:
            value, versionsuffix = value.split(',', 1)
        newvalue = subsystems.get(value)
        if newvalue is not None:
            if versionsuffix is not None:
                newvalue += f',{versionsuffix}'
            args = [f'--subsystem,{newvalue}']
        else:
            raise MesonException(f'Only {set(subsystems)} are supported for win_subsystem with MinGW, not "{value}".')

        return self._apply_prefix(args)
```

eli-schwartz: That looks reasonable, yes.

Should we map some other values too?

> Numeric values are also accepted for which.

Doesn't seem to be any restriction there, my interpretation is that `ld` allows any number and simply assumes it must be a valid subsystem.

So we *could* enable all of the ones MSVC accepts.

LensPlaysGames: > That looks reasonable, yes.

Good to hear, and thank you :^). It's been committed.

> Should we map some other values too?

This sounds like something worth doing, in order to help anyone down the line making other applications. I'll look into it.

mesonbuild/linkers/linkers.py
```diff
@@ -700,18 +700,22 @@ def build_rpath_args(self, env: 'Environment', build_dir: str, from_dir: str,
         return (args, rpath_dirs_to_remove)
 
     def get_win_subsystem_args(self, value: str) -> T.List[str]:
-        if 'windows' in value:
-            args = ['--subsystem,windows']
-        elif 'console' in value:
-            args = ['--subsystem,console']
-        elif 'efi_application' in value:
-            # objcopy allows a subsystem of efi-app, but ld doesn't share
-            # the same capability, so we must use the number directly.
```
eli-schwartz: This comment should be moved to the mapping table.

eli-schwartz (changes requested):
BTW the commits could be squashed. ;)

LensPlaysGames:
> BTW the commits could be squashed. ;)

I actually have never used this feature before; thank you for introducing me to it!

mesonbuild/linkers/linkers.py
LensPlaysGames: I've looked into it: I couldn't find "concrete" documentation on what the numbers are, but I did find some [rust code](https://docs.rs/object/0.25.3/src/object/pe.rs.html#487) for PE executables that defines the subsystem constants as follows:

- `native` = 0
- `windows` = 1
- `console` = 2
- `posix` = 7
- `efi_application` = 10
- `efi_boot_service_driver` = 11
- `efi_runtime_driver` = 12
- `efi_rom` = 13
- `xbox` = 14
- `boot_application` = 16

If we are happy using this source of magic numbers, then I'll go ahead and commit + squash these changes as well.

eli-schwartz: That rust crate documents:
> //! This module is based heavily on "winnt.h" (10.0.17763.0).

With that in mind, we can quickly find a couple other sources of magic numbers, specifically, the well-known WINE project:
https://github.com/wine-mirror/wine/blob/3ded60bd1654dc689d24a23305f4a93acce3a6f2/include/winnt.h#L2492-L2507

And the mingw-w64 sources as well:
https://github.com/mingw-w64/mingw-w64/blob/2c202976797b80703720a2bb19841d304205c936/mingw-w64-headers/include/winnt.h#L6572-L6585

These corroborate the values.

Naturally and understandably, https://docs.microsoft.com/en-us/windows/win32/api/winnt/ contains absolutely no reference to these constants at all.

mesonbuild/linkers/linkers.py
```diff
@@ -700,14 +700,31 @@ def build_rpath_args(self, env: 'Environment', build_dir: str, from_dir: str,
...
     def get_win_subsystem_args(self, value: str) -> T.List[str]:
-        if 'windows' in value:
-            args = ['--subsystem,windows']
-        elif 'console' in value:
-            args = ['--subsystem,console']
-        else:
-            raise MesonException(f'Only "windows" and "console" are supported for win_subsystem with MinGW, not "{value}".')
+        # Subsystem affects a single entry at the beginning of PE executables.
+        # It is used to verify that an executable is of the expected
+        # application type before attempting to execute it.
```
eli-schwartz, line 705: This is not an ideal comment, since we don't much care what it does, especially not as some mingw-specific thing.

Instead we should say something like "mingw only directly supports a couple of the possible PE application types. The raw integer works as an argument as well, and is always accepted, so we manually map the other types here. See LINK for the list of all types".

mesonbuild/linkers/linkers.py
```diff
@@ -700,14 +700,31 @@ def build_rpath_args(self, env: 'Environment', build_dir: str, from_dir: str,
...
+        subsystems = {
+            "native": "1",
+            "windows": "2",
+            "console": "3",
```
eli-schwartz, line 709: Since windows/console are supported without using integers, IMHO we should return the english name (resulting in the same value we used before this PR).

mesonbuild/linkers/linkers.py
```diff
@@ -700,14 +700,31 @@ def build_rpath_args(self, env: 'Environment', build_dir: str, from_dir: str,
...
+            "posix": "7",
+            "efi_application": "10",
+            "efi_boot_service_driver": "11",
+            "efi_runtime_driver": "12",
+            "efi_rom": "13",
+            "xbox": "14",
```
eli-schwartz, line 715: xbox is not a valid value in https://github.com/mesonbuild/meson/blob/7650f328284f24c2f70dcfb1689ae8fc22c04cf1/mesonbuild/build.py#L1260-L1264

mesonbuild/linkers/linkers.py
LensPlaysGames: I knew something should be said, but not exactly what as I'm not familiar with the codebase. 
Thanks for the great example, I'll update the comment.
 I'm going to link to the same MSDN article the docs link to, but if you want me to link to somewhere in the meson docs I'll change it.

eli-schwartz: Since the code comment is mainly concerned with "what are these magic values anyway" I would probably go with the wine source code link.

mesonbuild/linkers/linkers.py
```diff
@@ -700,14 +700,32 @@ def build_rpath_args(self, env: 'Environment', build_dir: str, from_dir: str,
...
         if ',' in value:
-            args[-1] = args[-1] + ':' + value.split(',')[1]
+            value, versionsuffix = value.split(',', 1)
+        newvalue = subsystems.get(value)
+        if newvalue is not None:
+            if versionsuffix is not None:
+                newvalue += f',{versionsuffix}'
+            args = [f'--subsystem,{newvalue}']
+        else:
+            raise MesonException(f'Only {set(subsystems)} are supported for win_subsystem with MinGW, not "{value}".')
```
eli-schwartz, line 728: Since we now map all types that are supported in the interpreter's validator (i.e. checked immediately when processing the `executable()` kwargs, not dependent on which linker is used), we can get rid of this MesonException as it cannot be reached.

Alternatively, we can change it to a `mesonlib.MesonBugException(f'win_subsystem: {value!r} not handled in MinGW linker. This should not be possible.')`

MesonBugException will generate a clever error message that includes the statement "this is a Meson bug and must be reported", thus encouraging people to report it as a bug because clearly the mapping dict needs to be updated to re-sync with the interpreter's validator.

LensPlaysGames: I actually ran into one of these same `MesonBugException`s while playing with the program and it helped me track down what I did wrong, so I am going to go with that option to help future developers.

eli-schwartz:
The commit message is very... information dense. I think all the "Also in this commit" content isn't really required for a permanent `git log`.

Other than that this LGTM.

LensPlaysGames:
> The commit message is very... information dense.

When squashing, it automatically concatenates the commit messages. I simply added a `-` before each commit message and called it good. I've cleaned it up now.

mesonbuild/linkers/linkers.py
```diff
@@ -700,14 +700,32 @@ def build_rpath_args(self, env: 'Environment', build_dir: str, from_dir: str,
...
+            "boot_application": "16",
+        }
+        versionsuffix = None
         if ',' in value:
-            args[-1] = args[-1] + ':' + value.split(',')[1]
+            value, versionsuffix = value.split(',', 1)
+        newvalue = subsystems.get(value)
+        if newvalue is not None:
+            if versionsuffix is not None:
+                newvalue += f',{versionsuffix}'
```
eli-schwartz, line 725: Wait, I just realized...

Before, we were joining the version suffix back with `:` only for mingw. So arguments looked like this for MSVC:

```
/SUBSYSTEM:CONSOLE,VERSION
```

And for MinGW:

```
--subsystem,console:version
```

This change seems to have overlooked that sneaky nuance, and causes it to reassemble the arguments to:

```
--subsystem,console,version
```

LensPlaysGames: Great catch, that slipped right by me. 

It's been updated.

eli-schwartz:
And just to confirm, this *does* get tested in CI even though I don't recall actually checking the CI status. Before the fixup for , to : the msys2 test failed like such:


```
[14/14] "cc"  -o gui_prog.exe gui_prog.exe.p/gui_prog.c.obj "-Wl,--allow-shlib-undefined" "-Wl,--subsystem,windows,6.0" "-Wl,--start-group" "-lkernel32" "-luser32" "-lgdi32" "-lwinspool" "-lshell32" "-lole32" "-loleaut32" "-luuid" "-lcomdlg32" "-ladvapi32" "-Wl,--end-group"
FAILED: gui_prog.exe 
"cc"  -o gui_prog.exe gui_prog.exe.p/gui_prog.c.obj "-Wl,--allow-shlib-undefined" "-Wl,--subsystem,windows,6.0" "-Wl,--start-group" "-lkernel32" "-luser32" "-lgdi32" "-lwinspool" "-lshell32" "-lole32" "-loleaut32" "-luuid" "-lcomdlg32" "-ladvapi32" "-Wl,--end-group"
D:/a/_temp/msys64/mingw32/bin/../lib/gcc/i686-w64-mingw32/11.3.0/../../../../i686-w64-mingw32/bin/ld.exe: cannot find 6.0: No such file or directory
collect2.exe: error: ld returned 1 exit status
```

This means that the PR does not need to add a new test case.

eli-schwartz (approved):

PR #10380, opened 2022-05-10
Add optional compress

joshcangit:
Compress files under source directory using Deflate method.
By default, files are not compressed in the archive.
Compression is active only when this option is defined.

[--compress](https://docs.python.org/3/library/zipapp.html?highlight=zipapp#cmdoption-zipapp-c) has no effect when copying an archive.

### DIFF
--- a/packaging/create_zipapp.py
+++ b/packaging/create_zipapp.py
@@ -11,6 +11,7 @@ parser = argparse.ArgumentParser()
 parser.add_argument('source', nargs='?', default='.', help='Source directory')
 parser.add_argument('--outfile', default='meson.pyz', help='Output file for the zipapp')
 parser.add_argument('--interpreter', default='/usr/bin/env python3', help='The name of the Python interpreter to use')
+parser.add_argument('--compress', action='store_true', default=False, help='Compress files')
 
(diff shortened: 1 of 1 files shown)


eli-schwartz:
Hmm, now I'm curious how the tradeoff between disk size and decompression overhead takes effect in practice. Did you try this out and it made a significant improvement?

joshcangit:
I've done a bit of testing on them.

### Size difference
```
-rwxrw-r-- 1 joshua 3225818 May 11 12:28 meson.pyz
-rwxrw-r-- 1 joshua  847527 May 11 12:28 meson-zip.pyz
```

> It's about 73.7% in size reduction.

### Testing execution time
Testing **uncompressed** file with `time meson.pyz --version`
```
0.62.1

real	0m1.212s
user	0m1.118s
sys	0m0.095s
```

Testing **compressed** file with `time meson-zip.pyz --version`:
```
0.62.1

real	0m1.276s
user	0m1.172s
sys	0m0.105s
```

> I think it may be anywhere from 0.01 to 0.1 seconds slower than if uncompressed.
> I think it's only a little slower since the original size, 3MB, is already not very big to begin with.

PR #10386, opened 2022-05-11
mtest: stop disrespecting the gdb config file

eli-schwartz:
This was added in commit 01be50fdd90851f17de5499537f10b5b62c9fb49 with zero explanation as a side effect of moving code around. It seems like a really bad idea and it causes people to view debugging Meson projects on e.g. debuginfod systems as "painful".

/cc @tristan957


tristan957:
If this is accepted, please backport it to whatever versions are currently supported (0.61?). So thankful for this PR.

Issue #5399, opened 2019-05-16
Add a werror: (true|false) kwarg to cc.compiles


nirbheek:
Sometimes we need to test if a particular piece of code spews a compiler warning that will be an error with `-Werror` or `/WX` (msvc). Would be nice to have a kwarg for this so that build files don't need to have compiler-specific code.

Was discussed at https://gitlab.gnome.org/GNOME/glib/issues/1777#note_512409.

jpakkane:
Makes sense.

xclaesse:
Made a PR: https://github.com/mesonbuild/meson/pull/10387

anarazel:
This is awesome! I've had to write pretty ugly versions of this for postgres...

PR #10387, opened 2022-05-12
Compiler: Add werror kwarg to compiles(), links() and run() methods

xclaesse:
Fixes: #5399

### DIFF
--- a/docs/markdown/snippets/compile_werror.md
+++ b/docs/markdown/snippets/compile_werror.md
@@ -1,0 +1,5 @@
+## Treat warnings as error in compiler checks
+
+Compiler check methods `compiler.compiles()`, `compiler.links()` and `compiler.run()`
+now have a new `werror: true` keyword argument to treat compiler warnings as error.
+This can be used to check if code compiles without warnings.
--- a/mesonbuild/interpreter/compiler.py
+++ b/mesonbuild/interpreter/compiler.py
@@ -45,20 +45,20 @@ if T.TYPE_CHECKING:
         args: T.List[str]
         dependencies: T.List[dependencies.Dependency]
 
-    class CompileKW(TypedDict):
+    class BaseCompileKW(TypedDict):
+        no_builtin_args: bool
+        include_directories: T.List[build.IncludeDirs]
+        args: T.List[str]
+
+    class CompileKW(BaseCompileKW):
 
         name: str
-        no_builtin_args: bool
-        include_directories: T.List[build.IncludeDirs]
-        args: T.List[str]
         dependencies: T.List[dependencies.Dependency]
+        werror: bool
 
-    class CommonKW(TypedDict):
+    class CommonKW(BaseCompileKW):
 
         prefix: str
-        no_builtin_args: bool
-        include_directories: T.List[build.IncludeDirs]
-        args: T.List[str]
         dependencies: T.List[dependencies.Dependency]
 
     class ComputeIntKW(CommonKW):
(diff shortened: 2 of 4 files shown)


mesonbuild/interpreter/compiler.py
```diff
@@ -223,20 +225,20 @@ def version_method(self, args: T.List['TYPE_var'], kwargs: 'TYPE_kwargs') -> str
...
                 args.extend(self.compiler.get_include_args(idir, False))
-        if not nobuiltins:
+        if not kwargs['no_builtin_args']:
             opts = self.environment.coredata.options
             args += self.compiler.get_option_compile_args(opts)
             if mode is CompileCheckMode.LINK:
                 args.extend(self.compiler.get_option_link_args(opts))
-        args.extend(extra_args)
+        if kwargs.get('werror', False):
+            args.extend(self.compiler.get_werror_args())
```
eli-schwartz, line 240: Is there some reason to use `kwargs.get('werror', False)` here instead of `kwargs['werror']` like we do for all the other kwargs?

typed_kwargs guarantees this is always set to something, in this case defaulting to False anyway.

eli-schwartz (changes requested):

mesonbuild/interpreter/compiler.py
xclaesse: That function is called by other methods that don't have that kwarg, like has_header().

eli-schwartz: Ah right, that makes sense -- I should have looked closer.

eli-schwartz:
Test failure is relevant:

```
D:\a\1\s\b 13e1d34133\meson-private\tmpe28q5siv\testfile.c(1): fatal error C1021: invalid preprocessor command 'warning'

Compiler stderr:
 
Checking if "should succeed with warning" compiles: NO 

test cases\common\28 try compile\meson.build:32:4: ERROR: Problem encountered: Compiler msvc returned false on valid code.
```

MSVC uses `#pragma message("informational text")` but that is printed as informational text even if you happen to include the English word "warning" in the informational text, so that won't actually test anything.

eli-schwartz:
We had a similar issue here: https://github.com/mesonbuild/meson/pull/6795#discussion_r394572370

In that case we did indeed make the tests only test anything on gcc/clang.

> The CI fails with MSVC that doesn't support #warning. Since this change is compiler agnostic, it doesn't matter, IMO.

Seems to me like this change is compiler agnostic too, so it's not the end of the world if we don't test anything on MSVC.

docs/markdown/snippets/compile_werror.md
```diff
@@ -0,0 +1,5 @@
+## Treat warnings as error in compiler checks
+
+Compiler check methods `compiler.compiles()`, `compiler.links()` and `compiler.run()`
+now have a new `werror: true` keyword argument to treat compiler warnings as error.
+This can be used to check of code compilers without warnings.
```
nirbheek, line 5: *to check if code compiles

xclaesse:
> > The CI fails with MSVC that doesn't support #warning. Since this change is compiler agnostic, it doesn't matter, IMO.
> 
> Seems to me like this change is compiler agnostic too, so it's not the end of the world if we don't test anything on MSVC.

Tried a few things to get a code that warns on both MSVC and GCC, but turns out it's not that easy. I'm wondering if we should also be able to define the `warning_level` option to make this useful in practice. But then I'm wondering if we should actually add the `override_options` kwarg instead just like on build targets? OTOH, having separate `werror` and `warning_level` kwargs is a nicer syntax but less generic IMHO.

eli-schwartz:
Holy cow.

https://en.cppreference.com/w/c/preprocessor/error

They added `#warning` to C23, which means that MSVC is now compelled to implement it in order to be conformant.

docs/markdown/snippets/compile_werror.md
xclaesse: done.

xclaesse:
> > The CI fails with MSVC that doesn't support #warning. Since this change is compiler agnostic, it doesn't matter, IMO.
> 
> Seems to me like this change is compiler agnostic too, so it's not the end of the world if we don't test anything on MSVC.

Skipped that part on MSVC, it's agnostic indeed.

tristan957:
```
mesonbuild/wrap/__init__.py:54:9: error: Returning Any from function declared to return "str"
Found 1 error in 1 file (checked 176 source files)
```

xclaesse:
Lol, it's completely unrelated and mypy pass fine locally... I think mypy hates me more than I hate it.

tristan957:
Is this still good to go.

tristan957:
I don't see a FeatureNew anywhere

xclaesse:
I think it's still desired feature. Rebased.

tristan957:
@eli-schwartz wanna provide a re-review when you get a chance since you "requested changes"

xclaesse:
> I don't see a FeatureNew anywhere

It's done by KwargInfo()

tristan957:
Knew that, but I totally missed that line of code for whatever reason.

mesonbuild/interpreter/compiler.py
```diff
@@ -251,20 +253,20 @@ def version_method(self, args: T.List['TYPE_var'], kwargs: 'TYPE_kwargs') -> str
     def cmd_array_method(self, args: T.List['TYPE_var'], kwargs: 'TYPE_kwargs') -> T.List[str]:
         return self.compiler.exelist
 
-    def _determine_args(self, nobuiltins: bool,
-                        incdirs: T.List[build.IncludeDirs],
-                        extra_args: T.List[str],
+    def _determine_args(self, kwargs: 'BaseCompileKW',
```
tristan957, line 256: no need for quoting

test cases/common/28 try compile/meson.build
```diff
@@ -31,4 +34,15 @@ foreach lang : ['c', 'cpp']
   if compiler.compiles(files('invalid.c'), name : 'file should fail')
     error('Compiler ' + compiler.get_id() + ' returned true on broken code.')
   endif
+
+  # MSVC does not support #warning instruction
```
tristan957, line 38: I think we should add similar tests for links and runs and then this is good.

xclaesse: Expanded tests to check the full matrix werror=true,false and compiles,links,runs.

mesonbuild/interpreter/compiler.py
xclaesse: done

tristan957 (approved):

PR #10403, opened 2022-05-17
fix incorrect FeatureNew for separator kwarg in environment object methods

### LINKED ISSUE #10402: Separator arg warning with meson version >= 0.62.0
**Describe the bug**
For a long time, I have used the `separator` kwarg when setting an environment variable in meson. Now, [v0.62.0](https://mesonbuild.com/Release-notes-for-0-62-0.html#new-method-and-separator-kwargs-on-environment-and-mesonadd_devenv) says this is a new feature...

A meson.build file that I've been using since before meson version 0.58.0 includes the line:
```
env.append('BAR', '0', separator: ':')
```

This has worked fine before, but after upgrading my meson version to 0.62.1, I now get this warning after running `meson build`:
```
meson.build:4: WARNING: Project targeting '>=0.58.0' but tried to use feature introduced in '0.62.0': separator arg in environment.append.
```

**To Reproduce**
1. With meson >= 0.62.0 installed, take the following meson.build:
```
project('foo', 'c', meson_version: '>=0.58.0')

env = environment()
env.append('BAR', '0', separator: ':')
```

2. Run `meson build`
3. See the warning

This warning doesn't occur with meson < 0.62.0

**Expected behavior**
I would not expect to receive a warning for a "new" feature that works with older meson versions.

eli-schwartz:
In commit 6acfe48f32110671a0adf80ad3497a35105b265d, the kwarg was added to environment() in addition to the env object methods. As part of the associated refactor, a shared KwargInfo was used, and evolved to be new since 0.62.0 in the two cases where it is in fact new.

However, it *also* set the base KwargInfo for the exact same newness, which is wrong as it was present ever since the initial introduction in 0.34.0

As usual for anything that predates 0.37.0 we simply don't tag FeatureNew. Revert this back to the same KwargInfo definition from before the refactoring commit.

Fixes #10402

### DIFF
--- a/mesonbuild/interpreter/type_checking.py
+++ b/mesonbuild/interpreter/type_checking.py
@@ -336,4 +336,4 @@ DEFAULT_OPTIONS: KwargInfo[T.List[str]] = KwargInfo(
 ENV_METHOD_KW = KwargInfo('method', str, default='set', since='0.62.0',
     validator=in_set_validator({'set', 'prepend', 'append'}))
 
-ENV_SEPARATOR_KW = KwargInfo('separator', str, default=os.pathsep, since='0.62.0')
+ENV_SEPARATOR_KW = KwargInfo('separator', str, default=os.pathsep)


xclaesse:
Thanks, LGTM.

PR #10404, opened 2022-05-18
ninja backend: generate additional meta-rules for test/benchmarks targets

eli-schwartz:
'meson-test-prereq' now depends on any targets that were formerly added directly to 'all'. Behavior is not changed -- the all target still depends on this other meta-rule, and thus indirectly depends on all targets it used to depend on.

It is now possible to build just the targets needed for the testsuite and then e.g. run `meson test --no-rebuild`.

### DIFF
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from __future__ import annotations
 
 from collections import OrderedDict
 from dataclasses import dataclass, InitVar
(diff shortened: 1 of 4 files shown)


lazka:
(those MSYS2 errors could be because we updated to ninja 1.11.0 yesterday... let me know if I should investigate that somehow..)

lazka:
> (those MSYS2 errors could be because we updated to ninja 1.11.0 yesterday... let me know if I should investigate that somehow..)

ah, never mind; master CI ran through, sorry for the noise.

lazka:
ah, I missed it, let me check IRC logs.

lazka:
* ninja downgrade doesn't change anything
* "ninja explain" is written to stderr, while "ninja: no work to do." is written to stdout, so imo this depends on buffering/timing what you get first if you pipe them together.

If there is anything else I should look into/try, let me know.

eli-schwartz:
> "ninja explain" is written to stderr, while "ninja: no work to do." is written to stdout, so imo this depends on buffering/timing what you get first if you pipe them together.

That is... a good point. The platform tests invoke all commands with `stderr=subprocess.STDOUT` which in cases where you explicitly only want stdout seems a bit unfortunate...

eli-schwartz:
Alright, I tentatively added two preparatory commits to make the unittest handler be nicer about this.

dcbaker (changes requested):
One small nit, but otherwise this looks good to me.

dcbaker (approved):

PR #10408, opened 2022-05-19
Add sentinel and section function attributes

tristan957:


tristan957:
Still needs a docs update.

mesonbuild/compilers/mixins/clike.py
```diff
@@ -187,8 +187,8 @@ def get_compile_only_args(self) -> T.List[str]:
     def get_no_optimization_args(self) -> T.List[str]:
         return ['-O0']
 
-    def get_output_args(self, target: str) -> T.List[str]:
-        return ['-o', target]
+    def get_output_args(self, outputname: str) -> T.List[str]:
+        return ['-o', outputname]
```
eli-schwartz, line 191: I dislike this, but it's not your fault.

https://peps.python.org/pep-0570/

What we actually want is:
```python
    def get_output_args(self, target: str, /) -> T.List[str]:
        return ['-o', target]
```

```
>>> instance.get_output_args(target='foo')
TypeError: get_output_args() got some positional-only arguments passed as keyword arguments: 'target'
```

... sadly that takes python 3.8 and we still have to be source-compatible with 3.7 so...

This isn't a bad change, I merely regret the need for it.

Issue #5468, opened 2019-06-10
Extracting directories from include_directories()


Snektron:
When using an unsupported compiler in a generator (such as nvcc in my case), it would be useful to be able to add include directories. For example one could write:
```
nvcc = generator(
    find_program('nvcc'),
    output: '@PLAINNAME@.o',
    arguments: [
        '-I' + subproject('thing').get_variable('inc').directories()[0],
        '@INPUT@', '-c', '-o', '@OUTPUT@'
    ]
)
```
Currently i see no way do to this without hardcoding the include path of the subproject.

zackw:
I would also like to have this feature.  In my case, it is because I need to invoke the compiler by hand from inside a program that's part of my test suite (it is testing properties of the ABI of a static library).

xclaesse:
Not exactly what you want, but https://github.com/mesonbuild/meson/pull/4436 could help you actually. You compiler wrapper script that set PKG_CONFIG_LIBDIR to `<builddir>/meson-uninstalled` then use `pkg-config --cflags` to get include flags.

I've been using that lately to build Rust code using cargo from within a meson project.

denizzzka:
+1 for provide access to directories list of `include_directories`

My case same as initial, but instead of `-I` my generator needs `--include-path=` (this is for d++, tool to convert C/C++ headers to D)

PR #10409, opened 2022-05-19
Add subproject.build_root()

### LINKED ISSUE #8302: Add a subproject.build_root() function
I am doing some funky stuff where my subproject is a Cython project and I want to know about the location of the build output so I can appropriately add to my `PYTHONPATH`.

tristan957:
Imagine a C project with a subproject that is its Python bindings. The C
project makes use of the Python bindings to use a higher level language
for writing tests. In order to properly set PYTHONPATH in the tests'
environments, you need to know the subprojects build root, so that the
generated module is visibile when the test imports it. By exposing the
subproject's build root, the parent project can easily set the
PYTHONPATH to resolve the import.

Fixes #8302


tristan957:
Assuming this gets accepted, I wouldn't mind also having a source_root()

tristan957:
I am going to advocate for this again. In some Go bindings for a C library that I have been working on, I need to set the `CGO_CFLAGS` so that header files are visible. In the case of the installed C library, I can use `dep.get_variable('includedir')`, but in the case of the subproject being used when the installed library isn't available, I cannot just add an `includedir` variable to the `dep` because I have some public headers being generated and others in the source tree, so I need to pass two paths to `CGO_CFLAGS`. I have worked around it by doing the following:

```meson
cgo_env.set('CGO_CFLAGS', '-I@0@ -I@1@'.format(hse_dep.get_variable('includedir'), meson.global_build_root() / 'subprojects/hse/include')
```

Obviously this is not ideal because `subprojects/hse` is an implementation detail.

cc @eli-schwartz

eli-schwartz:
I thought I already commented on this but apparently not...

I don't really understand the rationale for this. It can inherently be done with `subp.get_variable('build_root')` instead, plus an accompanying variable assignment in the subproject.

And, in general this involves poking at subproject internals, which we would usually say means the subproject should collaborate with this use case by exporting whatever variables are necessary.

As for the specificity of this case exactly, this feels very much a solution to the wrong problem. The actual problem is that you want to get `include_directories()` attributes from a dep object and pass them as gcc-compatible flags to a custom target (or an env object applying to a custom target) in order to work around the fact that golang doesn't have built-in support.

So, hard NACK. If we are going to add API here, I'm strongly of the opinion that subp.get_variable() is wrong and you actually want idk, `dep.cflags_as_string()`.

eli-schwartz:
Relevant: #5468

tristan957:
Makes sense. Essentially, I want access to the cflags and libflags for a dependency. I need them for both internal and pkgconfig dependency types.

Issue #10401, opened 2022-05-16
Meson sets PATH rather than WINEPATH in run_exe()


hughsie:
**Describe the bug**

I have a subproject that defines a binary:

    gcab = executable(
      'gcab',
      install : true,
      ...
    )
    meson.override_find_program('gcab', gcab)

and defines it in the `gcab.wrap` file:

    [provide]
    program_names = gcab

and then a toplevel project (fwupd, but unimportant) that wants to use that binary:

    gcab = find_program('gcab')

...which is used like this:

    custom_target('foo',
      input : [
        'firmware.bin',
      ],
      output : 'colorhug-als-3.0.2.cab',
      command : [
        gcab, '--create', '--nopath', '@OUTPUT@', '@INPUT@',
      ],
    )

...I'm also then using a cross build file, with needs_exe_wrapper = true:

    [binaries]
    c = '/usr/bin/x86_64-w64-mingw32-gcc'
    cpp = '/usr/bin/x86_64-w64-mingw32-g++'
    pkgconfig = '/usr/bin/x86_64-w64-mingw32-pkg-config'
    exe_wrapper = '/usr/bin/wine'

    [properties]
    root = '/usr/x86_64-w64-mingw32/sys-root/mingw'
    needs_exe_wrapper = true

    [host_machine]
    system = 'windows'
    cpu_family = 'x86_64'
    cpu = 'x86_64'
    endian = 'little'

[...]

PR #10411, opened 2022-05-19
Do not call wine if the WINEPATH is already short enough

hughsie:
We call 'wine cmd' to shorten the wine paths so that we do not exceed
the 2048-char limit for the WINEPATH environment variable.

In the most common case where the path is well below this limit, just
use the unadulterated paths.
This speeds up generating exe_wrapper outputs, and also seems to work
correctly in more cases.

Fixes https://github.com/mesonbuild/meson/issues/10401

### DIFF
--- a/mesonbuild/mesonlib/universal.py
+++ b/mesonbuild/mesonlib/universal.py
@@ -1893,6 +1893,11 @@ def get_wine_shortpath(winecmd: T.List[str], wine_paths: T.Sequence[str]) -> str
 
     wine_paths = list(OrderedSet(wine_paths))
 
+    # do we even need to bother shortening the path?
+    wine_path = ';'.join(wine_paths)
+    if len(wine_path) < 2048:
+        return wine_path
+
     getShortPathScript = '%s.bat' % str(uuid.uuid4()).lower()[:5]
     with open(getShortPathScript, mode='w', encoding='utf-8') as f:
         f.write("@ECHO OFF\nfor %%x in (%*) do (\n echo|set /p=;%~sx\n)\n")
--- a/mesonbuild/mesonlib/universal.py
+++ b/mesonbuild/mesonlib/universal.py
@@ -1913,9 +1913,9 @@ def get_wine_shortpath(winecmd: T.List[str], wine_paths: T.Sequence[str]) -> str
         wine_path = ';'.join(wine_paths)
     finally:
         os.remove(getShortPathScript)
-    if len(wine_path) > 2048:
+    if len(wine_path) >= 2048:
         raise MesonException(
-            'WINEPATH size {} > 2048'
+            'WINEPATH size {} >= 2048'
             ' this will cause random failure.'.format(
                 len(wine_path)))
 


mesonbuild/mesonlib/universal.py
```diff
@@ -1893,6 +1893,11 @@ def get_wine_shortpath(winecmd: T.List[str], wine_paths: T.Sequence[str]) -> str
 
     wine_paths = list(OrderedSet(wine_paths))
 
+    # do we even need to bother shortening the path?
+    wine_path = ";".join(wine_paths)
```
eli-schwartz, line 1897: I think we generally use single-quotes.

In general, this LGTM.

mesonbuild/mesonlib/universal.py
```diff
@@ -1893,6 +1893,11 @@ def get_wine_shortpath(winecmd: T.List[str], wine_paths: T.Sequence[str]) -> str
...
+    if len(wine_path) < 2048:
```
eli-schwartz, line 1898: I think it can be 2048 exactly without problems? Down below we check for `> 2048` in order to error out if (after the bat script) it is still too long. So maybe this should be `<= 2048`?

mesonbuild/mesonlib/universal.py
hughsie: Fixed, thanks. I'm used to `black`. :)

mesonbuild/mesonlib/universal.py
hughsie: I was worried about the need of a trailing NUL, and I can't find the limit in the wine codebase. I did wonder if `winepath` might be a better way to shorten those paths anyway...

eli-schwartz: Either way, shouldn't we be consistent? and allow `< 2048` in both places, or `<= 2048` in both places?

hughsie: I've added a follow up patch to limit the WINEPATH to 2048-1 chars.

jpakkane:
LGTM.

eli-schwartz:
I'd actually like to know which one is correct...

eli-schwartz:
> I've added a follow up patch to limit the WINEPATH to 2048-1 chars.

Assuming we do limit it to 2047 instead of 2048, shouldn't that commit come *first*, and then add the commit to shortcut if it's already inside the 2047 limit?

eli-schwartz:
The limit seems a lot more complicated.

```
WINEPATH='C:\aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa' wine cmd /c echo %PATH%
```
produces
```
'C:\aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa;C:\windows\system32;C:\windows;C:\windows\system32\wbem;C:\windows\system32\WindowsPowershell\v1.0'
```

371 characters long.

If I try to grow WINEPATH by a single byte, I get a segmentation fault. Grow it a bit more, so that the resulting path is 374 characters, and I get
```
wine: Unhandled page fault on read access to FFFFFFFF at address 64A50003 (thread 01ac), starting debugger...
```

Eventually at 620 I start hitting this instead:
```
0324:err:virtual:virtual_setup_exception stack overflow 836 bytes in thread 0324 addr 0x610061 stack 0x440cbc (0x440000-0x441000-0x640000)
```

How do I test this PR?

hughsie:
> If I try to grow WINEPATH by a single byte, I get a segmentation fault

This all feels really fragile. Does anyone know where the WINEPATH is handled in wine? It seems like working around the 2048 limit is causing all kinds of problems, when it might be best to "fix" wine so allow any size environment variable.

mesonbuild/mesonlib/universal.py
xclaesse: In my testing for devenv, the limit is actually 1024.

xclaesse: Also devenv has a smarter code to shorten path, using relative to builddir paths.

xclaesse:
I worked on this recently for devenv, wrote similar code there: https://github.com/mesonbuild/meson/blob/master/mesonbuild/mdevenv.py#L31. I did not know we already had code for this when I wrote that, so the first step is to refactor to remove that code duplication.

I tested by cross compiling GStreamer:
```
$ meson builddir -Dauto_features=disabled -Dtools=enabled --cross-file data/cross-files/mingw_w64_x86-64.txt
$ ninja -C builddir
$ meson devenv -C builddir
$ echo $WINEPATH
subprojects/glib/glib;subprojects/gstreamer/libs/gst/controller;subprojects/gst-plugins-base/gst-libs/gst/pbutils;subprojects/gst-plugins-base/gst-libs/gst/sdp;subprojects/gstreamer/tools;subprojects/glib/gthread;subprojects/proxy-libintl;subprojects/gst-plugins-base/gst-libs/gst/rtsp;subprojects/gst-plugins-base/gst-libs/gst/app;subprojects/gst-plugins-base/gst-libs/gst/rtp;subprojects/gstreamer/gst;subprojects/gst-plugins-base/gst-libs/gst/video;subprojects/glib/gobject;subprojects/libffi/src;subprojects/gst-plugins-base/gst-libs/gst/tag;subprojects/gst-plugins-base/gst-libs/gst/allocators;subprojects/gstreamer/libs/gst/net;subprojects/gst-plugins-base/gst-libs/gst/riff;subprojects/gst-plugins-base/tools;subprojects/glib/gio;subprojects/gst-plugins-base/gst-libs/gst/audio;subprojects/gst-plugins-base/gst-libs/gst/fft;subprojects/glib/gmodule;subprojects/zlib-1.2.11;subprojects/gstreamer/libs/gst/base;Z:\usr\X86_~FWL\bin;Z:\usr\X86_~FWL\lib;Z:\usr\lib\gcc\X86_~FWL\10-posix;Z:\usr\lib\gcc\x86_64-w64-mingw32
$ echo $WINEPATH | wc -c
1023
$ wine subprojects/gstreamer/tools/gst-launch-1.0.exe --version
gst-launch-1.0 version 1.21.0
GStreamer 1.21.0 (GIT)
Unknown package origin
```

As you can see, by default I got super lucky and WINEPATH is exactly 1023 chars and it works fine. What were the odds...

Now lets append a few chars:
```
$ export WINEPATH="$WINEPATH;a"
$ wine subprojects/gstreamer/tools/gst-launch-1.0.exe --version
gst-launch-1.0 version 1.21.0
GStreamer 1.21.0 (GIT)
Unknown package origin
```
1025 (including the extra `;`) worked.

```
$ export WINEPATH="$WINEPATHa"
$ wine subprojects/gstreamer/tools/gst-launch-1.0.exe --version
00f0:err:module:import_dll Library libglib-2.0-0.dll (which is needed by L"Z:\\home\\xclaesse\\programmation\\gstreamer\\builddir\\subprojects\\gstreamer\\tools\\gst-launch-1.0.exe") not found
00f0:err:module:import_dll Library libgobject-2.0-0.dll (which is needed by L"Z:\\home\\xclaesse\\programmation\\gstreamer\\builddir\\subprojects\\gstreamer\\tools\\gst-launch-1.0.exe") not found
00f0:err:module:import_dll Library libgstreamer-1.0-0.dll (which is needed by L"Z:\\home\\xclaesse\\programmation\\gstreamer\\builddir\\subprojects\\gstreamer\\tools\\gst-launch-1.0.exe") not found
00f0:err:module:import_dll Library libintl-8.dll (which is needed by L"Z:\\home\\xclaesse\\programmation\\gstreamer\\builddir\\subprojects\\gstreamer\\tools\\gst-launch-1.0.exe") not found
00f0:err:module:LdrInitializeThunk Importing dlls for L"Z:\\home\\xclaesse\\programmation\\gstreamer\\builddir\\subprojects\\gstreamer\\tools\\gst-launch-1.0.exe" failed, status c0000135
```
1026 failed.

I have no idea what's the rational for that... There is some discussion there: https://bugs.winehq.org/show_bug.cgi?id=45810

xclaesse:
Note that I have wine 6.0.3 (Ubuntu 22.04 LTS), and last command in the [bug report](https://bugs.winehq.org/show_bug.cgi?id=45810#c9)  suggests it has been fixed in 6.4. If someone can test that version.

eli-schwartz:
Alright, so under the same circumstances, cmd /c will segfault or whatever, but gst-launch-1.0.exe --version works for me. Fun!

gst-launch-1.0.exe --version in fact works for me even given this:
```
$ printf %s "$WINEPATH" | wc -c
1065
$ wine subprojects/gstreamer/tools/gst-launch-1.0.exe --version
gst-launch-1.0 version 1.21.0
GStreamer 1.21.0 (GIT)
Unknown package origin
```

It works if I grow it to 2282 as well, which is over the 2048 limit. In fact, I grew it to 8152 before getting bored due to it continuing to work.

I have wine 7.x installed.

> As you can see, by default I got super lucky and WINEPATH is exactly 1023 chars and it works fine. What were the odds...

P.S. Since you used `echo` instead of `printf`, there was an extra char -- a newline character -- at the end of the stream of characters which you counted. So you're off by one.

xclaesse:
Given that it can work in at least some circumstances, I would downgrade the exception to a simple warning, which is what the code in mdevenv.py already does.

eli-schwartz:
> You mean the command we run to shorten paths segfault when it gets 2048-ish paths in its argv?

No, I mean that given a specific WINEPATH value, and running `wine someprogram`:
 - the cmd.exe shell will segfault
 - gst-launch-1.0.exe will work fine

This is presumably a bug in the cmd.exe shell, which is very fun because I had used it at one point to try to test the behavior of the variable. Luckily, the actual meson-built programs in the devenv don't have this bug, so all is well...

hughsie:
> I can confirm that using wine 7.9 from winehq repository fixes this issue completely

Fedora 36 just got wine 7.9 too -- maybe we should just rip out the patch shortening completely?

xclaesse:
Ubuntu 22.04 LTS is there for a while still, I think it's better to keep that work around.

jpakkane:
⚔️ conflicts.

eli-schwartz:
This was obsoleted by #10457 (hence the conflicts).

Issue #10032, opened 2022-02-25
mtest TAP parser parses fails on Test::More and node-tap subtests


anarazel:
Hi,

CC: @bonzini 

Test::More and node-tap generate subtest output like this:
```
# Subtest: some subject
    ok 1 - one property
    ok 2 - another property
    1..2
ok 1 - some subject
1..1

```

which mtest doesn't like: "TAP parsing error: unexpected input at line 4"

This is supposed to be backward compatible with tools that don't understand subtests. Apparently based on the following language in the tap spec (v12, not 13, which appears to be a draft or something?):

> Any output line that is not a version, a plan, a test line, a diagnostic or a bail out is considered an “unknown” line. A TAP parser **is required to not consider an unknown line as an error** but may optionally choose to capture said line and hand it to the test harness, which may have custom behavior attached. This is to allow for forward compatability. Test::Harness silently ignores incorrect lines, but will become more stringent in the future. TAP::Harness reports TAP syntax errors at the end of a test run.

I don't quite understand how "required to not consider an unknown line as an error", "silently ignores incorrect lines" and "TAP::Harness reports TAP syntax errors at the end of a test run" fit together...

Given that common producers of TAP generate output like this, it may be worth ignoring such lines unless v13 is specified? 

meson.build:
```meson
project('test', 'c')

perl = find_program('perl')
test('subtest', perl,
     protocol: 'tap',
     args: [files('subtest.pl')])
```

subtest.pl:
```perl
use Test::More;

sub run_sub
{
    my $arg = shift;
    ok($arg, "one property");
    ok($arg, "another property");
}

subtest "some subject" => \&run_sub, 1;
done_testing();
```

Regards,

Andres

ilmari:
The subtest syntax has now been formally specified in TAP version 14: https://testanything.org/tap-version-14-specification.html#subtests

anarazel:
Thanks @bonzini and @jpakkane!

atorik:
Hi,

I tried to add subtest, but current HEAD(d7b25662ce5c7f9831b003581054e174b97c4865) doesn't accept subtest outputs.
Below is the result when I ran meson test on meson.build and subtest.pl, which were pasted on this issue by Andres.

```
1/1 subtest        OK              0.08s   1 subtests passed

stdout: 2: UNKNOWN:     ok 1 - one property
stdout: 3: UNKNOWN:     ok 2 - another property
stdout: 4: UNKNOWN:     1..2
[...]

PR #10418, opened 2022-05-23
mtest: ignore invalid input

bonzini:
TAP version 14 introduced subtests, that are supposedly backward compatible because "TAP13 specifies that non-TAP output should be ignored".  Meson reported TAP syntax errors based on behavior of "prove" at the time, but it seems that now "prove" has become a lot more lenient; it even accepts the following completely bogus input just fine:

```
ok 1
    ok 2
x
1..1
```

So do the same and make Meson's parser accept invalid TAP input silently.

Fixes: #10032

### DIFF
--- a/mesonbuild/mtest.py
+++ b/mesonbuild/mtest.py
@@ -426,11 +426,6 @@ class TAPParser:
                 else:
                     yield self.Version(version=self.version)
                 return
-
-            if not line:
-                return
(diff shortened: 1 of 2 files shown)


eli-schwartz:
Thinking about this more, I think it's probably wrong to completely ignore it: #10964

Issue #10394, opened 2022-05-14
custom targets don't handle extraced objects input containing custom target generated sources


anarazel:
Hi,

When a custom target has build_tgt.extract_all_objects() as the source and the build_tgt has another custom target as the input, meson fails with a backtrace.

Repro:

```meson
project('foo', ['c'])

test2_c = custom_target(output: 'test.c', command: ['echo', 'int foo;'])

lib1 = static_library('l1', test2_c)

custom_target(
  input: lib1.extract_all_objects(recursive: false),
  output: 'test.txt',
  command: ['echo', '@INPUT@'],
  capture: true,
)
```

results in:
```
The Meson build system
Version: 0.62.99
Source dir: /tmp/repro
Build dir: /tmp/repro/build
Build type: native build
Project name: foo
Project version: undefined
C compiler for the host machine: ccache cc (gcc 11.3.0 "cc (Debian 11.3.0-1) 11.3.0")
C linker for the host machine: cc ld.bfd 2.38
Host machine cpu family: x86_64
Host machine cpu: x86_64
Traceback (most recent call last):
  File "/home/andres/src/meson/mesonbuild/mesonmain.py", line 151, in run
    return options.run_func(options)
  File "/home/andres/src/meson/mesonbuild/msetup.py", line 303, in run
    app.generate()
  File "/home/andres/src/meson/mesonbuild/msetup.py", line 186, in generate
    self._generate(env)
  File "/home/andres/src/meson/mesonbuild/msetup.py", line 230, in _generate
    intr.run()
  File "/home/andres/src/meson/mesonbuild/interpreter/interpreter.py", line 2766, in run
    super().run()
  File "/home/andres/src/meson/mesonbuild/interpreterbase/interpreterbase.py", line 149, in run
    self.evaluate_codeblock(self.ast, start=1)
  File "/home/andres/src/meson/mesonbuild/interpreterbase/interpreterbase.py", line 174, in evaluate_codeblock
    raise e
  File "/home/andres/src/meson/mesonbuild/interpreterbase/interpreterbase.py", line 167, in evaluate_codeblock
    self.evaluate_statement(cur)
  File "/home/andres/src/meson/mesonbuild/interpreterbase/interpreterbase.py", line 180, in evaluate_statement
    return self.function_call(cur)
  File "/home/andres/src/meson/mesonbuild/interpreterbase/interpreterbase.py", line 462, in function_call
    res = func(node, func_args, kwargs)
[...]

PR #10419, opened 2022-05-23
Fix custom_target whose input is extract_objects of generated sources

bonzini:
Fix yet another extract_objects corner case.  This time however the fix is generic and removes some ugly hooking from build.py to the backend. Crossing fingers that it works on Xcode and VS...

### DIFF
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -456,6 +456,10 @@ class Backend:
         obj_list = self._flatten_object_list(target, target.get_objects(), proj_dir_to_build_root)
         return list(dict.fromkeys(obj_list))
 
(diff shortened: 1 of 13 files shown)


anarazel:
I suspect this would fix #10394

bonzini:
Yes it's the same backtrace. Are Postgres and QEMU discovering the same tricks at the same time? :)

anarazel:
> Yes it's the same backtrace. Are Postgres and QEMU discovering the same tricks at the same time? :)

Or maybe rather hitting the same meson limitations, requiring workaround like extract_objects...

bonzini:
I think it's different based on your bug report, QEMU is using the .o file names just as a key for compile_commands.json lookup. But in general it does look like static_library()+extract_objects() is a common escape hatch.

bonzini:
Failure is due to Win2k16 brownout

bonzini:
Ping @jpakkane.

PR #10422, opened 2022-05-24
Various typing fixes split out from the pkg-config typing PR

dcbaker:
As requested, here are the non-pkgconfig fixes from the pkg-config typing PR (plus another small one from another branch I have), split out for easier review. Most of thse came from #10108

### DIFF
--- a/mesonbuild/dependencies/base.py
+++ b/mesonbuild/dependencies/base.py
@@ -317,8 +317,19 @@ class InternalDependency(Dependency):
         raise DependencyException(f'Could not get an internal variable and no default provided for {self!r}')
 
     def generate_link_whole_dependency(self) -> Dependency:
+        from ..build import SharedLibrary, CustomTarget, CustomTargetIndex
         new_dep = copy.deepcopy(self)
-        new_dep.whole_libraries += new_dep.libraries
+        for x in new_dep.libraries:
+            if isinstance(x, SharedLibrary):
+                raise MesonException('Cannot convert a dependency to link_whole when it contains a '
+                                     'SharedLibrary')
+            elif isinstance(x, (CustomTarget, CustomTargetIndex)) and x.links_dynamically():
+                raise MesonException('Cannot convert a dependency to link_whole when it contains a '
+                                     'CustomTarget or CustomTargetIndex which is a shared library')
+
+        # Mypy doesn't understand that the above is a TypeGuard
+        new_dep.whole_libraries += T.cast('T.List[T.Union[StaticLibrary, CustomTarget, CustomTargetIndex]]',
+                                          new_dep.libraries)
         new_dep.libraries = []
         return new_dep
 
--- a/mesonbuild/dependencies/base.py
+++ b/mesonbuild/dependencies/base.py
@@ -35,7 +35,10 @@ if T.TYPE_CHECKING:
     from ..compilers.compilers import Compiler
     from ..environment import Environment
     from ..interpreterbase import FeatureCheckBase
-    from ..build import BuildTarget, CustomTarget, IncludeDirs
+    from ..build import (
+        CustomTarget, IncludeDirs, CustomTargetIndex, SharedLibrary,
+        StaticLibrary
+    )
     from ..mesonlib import FileOrString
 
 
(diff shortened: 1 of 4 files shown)


mesonbuild/interpreter/interpreterobjects.py
```diff
@@ -510,7 +510,12 @@ def as_system_method(self, args: T.Tuple[T.Optional[str]], kwargs: TYPE_kwargs)
     def as_link_whole_method(self, args: T.List[TYPE_var], kwargs: TYPE_kwargs) -> Dependency:
         if not isinstance(self.held_object, InternalDependency):
             raise InterpreterException('as_link_whole method is only supported on declare_dependency() objects')
-        new_dep = self.held_object.generate_link_whole_dependency()
+        try:
+            new_dep = self.held_object.generate_link_whole_dependency()
+        except MesonException as e:
+            raise MesonException(
+                e.args, self.current_node.filename, self.current_node.lineno,
+                self.current_node.colno)
```
eli-schwartz, line 518: I don't think we need this, because interpreterbase's `evaluate_codeblock()` catches all exceptions, checks if it already has these properties, and if not, assigns filename/lineno/colno. Then it reraises the exception.

mesonbuild/dependencies/base.py
```diff
@@ -317,8 +317,14 @@ def get_variable(self, *, cmake: T.Optional[str] = None, pkgconfig: T.Optional[s
         raise DependencyException(f'Could not get an internal variable and no default provided for {self!r}')
 
     def generate_link_whole_dependency(self) -> Dependency:
+        from ..build import SharedLibrary
         new_dep = copy.deepcopy(self)
-        new_dep.whole_libraries += new_dep.libraries
+        # XXX: we need to check that CustomTargets are not SharedLibs...
+        if any(isinstance(x, SharedLibrary) for x in new_dep.libraries):
+            raise MesonException(f'Cannot convert a dependency to link_whole with shared libraries')
```
eli-schwartz: I guess technically this could be an InvalidArguments, but it may not particularly matter?

mesonbuild/dependencies/base.py
```diff
@@ -317,11 +317,11 @@ def get_variable(self, *, cmake: T.Optional[str] = None, pkgconfig: T.Optional[s
...
     def generate_link_whole_dependency(self) -> Dependency:
-        from ..build import SharedLibrary
+        from ..build import SharedLibrary, CustomTarget, CustomTargetIndex
         new_dep = copy.deepcopy(self)
-        # XXX: we need to check that CustomTargets are not SharedLibs...
-        if any(isinstance(x, SharedLibrary) for x in new_dep.libraries):
-            raise MesonException(f'Cannot convert a dependency to link_whole with shared libraries')
+        for x in new_dep.libraries:
+            if isinstance(x, SharedLibrary) or (isinstance(x, (CustomTarget, CustomTargetIndex)) and x.links_dynamically()):
+                raise MesonException(f'Cannot convert a dependency to link_whole with shared libraries')
```
eli-schwartz, line 324: Can we reorder this slightly and merge this commit with the one two commits before?

mesonbuild/interpreter/interpreter.py
```diff
@@ -687,10 +687,19 @@ def func_declare_dependency(self, node, args, kwargs):
...
                 raise InterpreterException('Invalid dependency')
+
+        dep_err = 'declare_dependency: Entries in "{}" may only be self-built targets, external dependencies (including libraries) must go to "dependencies".'
+
         for l in libs:
             if isinstance(l, dependencies.Dependency):
-                raise InterpreterException('''Entries in "link_with" may only be self-built targets,
-external dependencies (including libraries) must go to "dependencies".''')
+                raise InvalidArguments(dep_err.format('link_with'))
+        for l in libs_whole:
```
eli-schwartz, line 696: Despite the commit message, it seems we're also checking that anything we disallowed in link_with is also disallowed in link_whole, which is a whole 'nother failure mode we're catching now.

I'd say this is worth mentioning in the commit message.

mesonbuild/interpreter/interpreter.py
```diff
@@ -687,10 +687,19 @@ def func_declare_dependency(self, node, args, kwargs):
...
+            if isinstance(l, dependencies.Dependency):
+                raise InvalidArguments(dep_err.format('link_whole'))
+            if isinstance(l, build.SharedLibrary):
+                raise InvalidArguments('declare_dependency: SharedLibrary objects are not allowed in "link_whole"')
+            if isinstance(l, (build.CustomTarget, build.CustomTargetIndex)) and l.links_dynamically():
+                raise InvalidArguments('declare_dependency: CustomTarget and CustomTargetIndex returning shared libaries are not allowed in "link_whole"')
```
eli-schwartz, line 702: The previous commit treats both of these the same and just calls them "shared libraries" without respect to what represented type it is. We should probably either change this to not distinguish them, or change the previous commit to distinguish them.

eli-schwartz (changes requested):
In general, this series looks really nice, and comes with some great fixes too.

Just a couple of tiny nitpicks below.

mesonbuild/interpreter/interpreter.py
dcbaker: I think being more specific is better than being less, so I'll convert the previous case

dcbaker:
@eli-schwartz I think I addressed everything except for the InvalidArguments vs MesonException. I'm not really sure there. I feel like we want to get to the point that InvalidArguments is all handled in the interpreter level, and not in the mid level, but…

eli-schwartz:
> everything except for the InvalidArguments vs MesonException. I'm not really sure there. I feel like we want to get to the point that InvalidArguments is all handled in the interpreter level, and not in the mid level, but...

I agree, that was the one part that I was also just very iffy about and only mentioned because it was a random thought and I didn't really expect a change.

PR #10424, opened 2022-05-25
Fix optimization level 's' for the TI CL2000 compiler

aobergschwandtner:
A wrong command line option is passed to the TI CL2000 compiler if the optimization level is set to 's' (0 instead of O).


xclaesse:
LGTM

PR #10427, opened 2022-05-25
Make a copy of auto_features options when changing its name

xclaesse:
This fixes bogus messages "skipped: feature foo disabled" when
auto_features=disabled. It was reporting the name of the latest
get_option() call instead of the name of the current feature option.

This is especially visible in GStreamer summary where it should show a
different option name for every subproject but instead shows "tools"
everywhere:
```
  Subprojects
    gst-devtools              : NO Feature 'tools' disabled
    gst-editing-services      : NO Feature 'tools' disabled
    ...
```


xclaesse:
@nirbheek You may want to backport this fix, it has probably been wrong for ages.

Issue #8651, opened 2021-04-13
generated "-uninstalled.pc" files missing `CFlags`


nanthony21:
I have a large collection of pre-built libraries which are dependencies for project `A`. Rather than  include all the logic for selecting linker/include paths in project `A` I would like to create a subproject `B` which does no compiling but instead just generates the appropriate `.pc` files to so that project `A` can find it's dependencies using `dependency('my3rdPartySDK')`.

I had initially hoped to be able to do this as something like:
```
pkg = import('pkgconfig')
SDK_lib = compiler.find_library('MySDKLibFileName`, dirs: meson.current_source_dir() / 'pathtolib')
SDK_dep = declare_dependency(include_directories: SDK_include_dir, dependencies: SDK_lib)
pkg.generate(SDK_dep)
```
but as far as I can tell `generate` will only take objects from `library()`, `shared_library()`, etc. so there is no way to do this for libraries that are pre-built.

I am now doing this instead:
```
SDK_lib = compiler.find_library('MySDKLibFileName`, dirs: meson.current_source_dir() / 'pathtolib')
pkg.generate(
    name: 'MySDKName',
    libraries: SDK_lib,
    description: 'My SDK Description',
    version: '0.1')
```

The resulting `-uninstalled.pc` file looks like this:
```
prefix=C:/absolute_path_to_meson_build_dir
srcdir=C:/absolute_path_to_meson_source_dir
libdir=${prefix}/lib
includedir=${prefix}/include

Name: MySDKName
Description: My SDK Description.
Version: 0.1
Libs: absolute_path_to_SDK_lib/MySDKLibFileName.lib
```

This obviously causes problems at compile time in project `A` because the header files are not found.

I tried adding something to the `subdirs` kwarg of `pkg.generate` but it had no effect on the final output. The only way I got the `Cflags` section to appear was with the `extra_cflags` kwarg, so I could potentially manually insert the correct include path with this kwarg.

It's also worth noting that the generated `includedir` variable doesn't point to the correct location. Is there a possibility of adding an `includes` kwarg to the generator that will set the `includedir` variable and then use it in `CFlags`?

Also, if there's a smarter way for me to achieve what I'm trying to do please let me know.

eli-schwartz:
I don't understand the goal here. pkg.generate is used to export information about your build outputs to non-meson build systems, but this doesn't seem to be a build output plus you're using it to export information out of a subproject to the superproject.

[...]

PR #10428, opened 2022-05-25
find_library: Add has_headers include dirs if available

nirbheek:
If the headers specified in `has_headers:` require `include_directories` to be usable, the dependency should also contain those directories in `compile_args`. Otherwise the user has to add it manually using `declare_dependency()` or `include_directories:` to each target.


eli-schwartz:
> Otherwise the user has to add it manually using `declare_dependency()` or `include_directories:` to each target.

I agree it would be neat if this could be easier, and TBH I'm not sure why the original design didn't do this.

That being said, this *does* constitute a behavior/semantics change.

It could be argued the change is a bugfix...

nirbheek:
> TBH I'm not sure why the original design didn't do this.

Yeah, I expected it to work this way and was surprised that it didn't. I think it's just one of those things that is obvious only once the feature is already implemented and you try to use it.

> That being said, this _does_ constitute a behavior/semantics change.
> 
> It could be argued the change is a bugfix...

I also don't think this will break any projects. At best it'll add duplicate include flags.

eli-schwartz:
> I also don't think this will break any projects. At best it'll add duplicate include flags.

Yeah, the question is whether anyone would ever *not* want that include directory. I suppose if they looked for it in the dependency lookup... probably not.

On the other hand, I just found a [muon](https:/muon.build) bug that had extra include directories added by source headers in violation of `implicit_include_directories: false` so I guess the general topic is just on my mind...

eli-schwartz:
I wonder if there's any way to add a FeatureNew to this?

xclaesse:
Issue https://github.com/mesonbuild/meson/issues/8651 is kind of related to this. @nirbheek please make sure that the include directory you add into ExternalLibrary also gets into the -uninstalled.pc file. Probably will need to use _add_uninstalled_incdirs() I add in https://github.com/mesonbuild/meson/pull/10452.

nirbheek:
> @nirbheek please make sure that the include directory you add into ExternalLibrary also gets into the -uninstalled.pc file

I don't understand. This is already handled there by this conditional?

```python
            elif isinstance(obj, dependencies.Dependency):
                if obj.found():
                    processed_libs += obj.get_link_args()
                    processed_cflags += obj.get_compile_args()
```

I'm just setting `compile_args` which was previously `[]`.

xclaesse:
That works in the case headers and on the system. Was wondering if we should deal with the case where the library is prebuilt in source tree, something like: `lib = cc.find_library('prebuilt', dirs: meson.current_source_dir(), had_headers: 'prebuilt.h', header_include_directories: '.')`. But that's probably silly to check for header that we have in source tree.

nirbheek:
> Was wondering if we should deal with the case where the library is prebuilt in source tree, something like:

That will also be covered, won't it?

```python
            for incdirs in kwargs['header_include_directories']:
                for arg in incdirs.to_string_list(self.environment.get_source_dir()):
                    compile_args.append(f'-I{arg}')
```

nirbheek:
> Was wondering if we should deal with the case where the library is prebuilt in source tree, something like:

That will also be covered, won't it?

```python
            for incdirs in kwargs['header_include_directories']:
                for arg in incdirs.to_string_list(self.environment.get_source_dir()):
                    compile_args.append(f'-I{arg}')
```

xclaesse:
Hmmm, wait, if you do that the installed pc file will point to source tree. What if those headers gets installed with install_headers() ?

nirbheek:
> Hmmm, wait, if you do that the installed pc file will point to source tree

You're thinking of the case where there's a prebuilt library + headers in the source tree that the project links to, and on install it installs everything into the prefix?

Yeah, I guess this will be wrong in that case. I think it would be too error-prone to poke at the `InstallData` etc structures in backends to do the right thing automatically; especially with subdir installation. 

I think we should filter ExternalLibrary paths pointing to the source / build trees for installed pc files, and require users to add those manually. The flags can stay for uninstalled pc files.

nirbheek:
> You're thinking of the case where there's a prebuilt library + headers in the source tree that the project links to, and on install it installs everything into the prefix?
> 
> Yeah, I guess this will be wrong in that case.

Hm, actually this is already behaving weirdly for this use-case, because we are generating pc files pointing to the builddir / srcdir for the library itself.

I am hesitant to change this. I think we should let this behaviour be as-is.

xclaesse:
> You're thinking of the case where there's a prebuilt library + headers in the source tree that the project links to, and on install it installs everything into the prefix?

Exactly

> Yeah, I guess this will be wrong in that case. I think it would be too error-prone to poke at the `InstallData` etc structures in backends to do the right thing automatically; especially with subdir installation.

I think if you store the include paths as `IncludeDirs` objects instead of `compile_args` it should be possible to check in pkgconfig if the path is in source/build tree and in that case only use it for -uninstalled.pc file. AFAIK ExternalLibrary can just implement `get_include_dirs()` from its baseclass for that.

> Hm, actually this is already behaving weirdly for this use-case, because we are generating pc files pointing to the builddir / srcdir for the library itself.

Ok, I'm fine with considering that use-case as already broken and could be fixed in another PR then.

mesonbuild/interpreter/compiler.py
```diff
@@ -591,6 +591,14 @@ def find_library_method(self, args: T.Tuple[str], kwargs: 'FindLibraryKW') -> 'd
             if not self._has_header_impl(h, has_header_kwargs):
                 return self.notfound_library(libname)
 
+        # If has_headers is enabled and header_include_directories were passed,
+        # then construct compile_args from those include dirs.
+        compile_args = []
+        if kwargs['has_headers']:
+            for incdirs in kwargs['header_include_directories']:
+                for arg in incdirs.to_string_list(self.environment.get_source_dir()):
+                    compile_args.append(f'-I{arg}')
```
xclaesse, line 600: I think we should store them as `IncludeDirs` objects on `ExternalProgram` object. Its base class already has `get_include_dirs()` that we can implement to return that list.

PR #10432, opened 2022-05-26
fix custom_target crash if boolean true is specified in install_dir

eli-schwartz:
We accept boolean false to indicate "do not install this one particular output", but the type checking simply checked if it is a bool. We do this correctly for configure_file, so copy the same validator from there.


dcbaker:
There's typing changes that go along with this

eli-schwartz:
> There's typing changes that go along with this: [f8bfaa0](https://github.com/mesonbuild/meson/commit/f8bfaa018bc1bb28231cc1b82e230559afbc212a)

Those typing changes are wrong for i18n by the way, see https://github.com/mesonbuild/meson/pull/10167/commits/80155ec909dc97c7cadb10db0cef8b69290c1a9b

I really need to make that PR work sanely though.

dcbaker (approved):

PR #10433, opened 2022-05-26
[skip ci] remove repetition and stiff language

guijan:


eli-schwartz:
> [[skip ci] remove repetition and stiff language](https://github.com/mesonbuild/meson/pull/10433/commits/8a519d31d83aa273adf6df807f0ae49f7e9b6c70)

These days we actually have file triggers, which means that changes which only touch docs/ only trigger the website update workflow. Which we probably should run here in this case. :)

Can you amend the PR to remove that?

PR #10435, opened 2022-05-27
linkers: try a bit harder to autodetect the correct linker id

eli-schwartz:
mingw GCC using ld.bfd emits diagnostics that include "-plugin-opt=-pass-through=-lmoldname" and this triggers a match for
mold when it should not.

Instead, always check the very beginning of the output for the linker id. This is pretty consistent:

- it is always on stdout
- GCC may put additional things on stderr we don't care about
- swift is bizarre and on some OSes redirects the linker stdout to swiftc's stderr, but it will still be the only stderr; we didn't even check stderr at all until commit 712cbe056811ebdf0d7358ba07a874717a1c736f

For gold/bfd linkers, the linker id is always the *second* word, after the legally mandated "GNU" we already check for.

### DIFF
--- a/mesonbuild/linkers/detect.py
+++ b/mesonbuild/linkers/detect.py
@@ -194,9 +194,11 @@ def guess_nix_linker(env: 'Environment', compiler: T.List[str], comp_class: T.Ty
         linker = AppleDynamicLinker(compiler, for_machine, comp_class.LINKER_PREFIX, override, version=v)
     elif 'GNU' in o or 'GNU' in e:
         cls: T.Type[GnuDynamicLinker]
-        if 'gold' in o or 'gold' in e:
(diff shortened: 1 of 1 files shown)


dcbaker (approved):

eli-schwartz:
>  LGTM.com / LGTM analysis: Python
> Started 3h 22m 15s ago 
> 
> The analysis is in progress and results will be available soon on [LGTM.com](https://lgtm.com/projects/g/mesonbuild/meson/rev/pr-3b586b4f796eee1258b760c5021864bdf70d3e09).

Good stuff. Even cygwin goes faster.

anarazel:
Can confirm that this fixes cross-compile issue I had.

Issue #10351, opened 2022-05-05
meson @0.62.0 fails to link correct arch when building on Rosetta


barracuda156:
Any idea how to fix the following? `meson` system fails to link the correct arch when building `gobject-introspection` on 10.6.8 Rosetta:
```
ld: warning: in /opt/local/lib/libgio-2.0.dylib, file was built for ppc7400 which is not the architecture being linked (x86_64)
ld: warning: in /opt/local/lib/libgobject-2.0.dylib, file was built for ppc7400 which is not the architecture being linked (x86_64)
ld: warning: in /opt/local/lib/libgmodule-2.0.dylib, file was built for ppc7400 which is not the architecture being linked (x86_64)
ld: warning: in /opt/local/lib/libglib-2.0.dylib, file was built for ppc7400 which is not the architecture being linked (x86_64)
ld: warning: in /opt/local/lib/libintl.dylib, file was built for ppc7400 which is not the architecture being linked (x86_64)
Undefined symbols:
…
ld: symbol(s) not found
collect2: ld returned 1 exit status
```
Of course it must link the build arch, `ppc` (=ppc7400), and not the arch of a physical cpu, but it does not.

I have tried specifying `--host-arch=ppc`, I have tried using cross file with:
```
[built-in options]
c_link_args = '-arch ppc'
```
Nothing helps. Somehow this works with other ports, at least so far, but in this instance I can’t figure what’s the trick to fix it.

tristan957:
@barracuda156 I don't think that is how you should be passing the c_link_args. Pretty sure it needs to be `c_link_args = ['-arch, 'ppc']`

barracuda156:
@tristan957 Thank you for responding! I am not an expert in `meson` at all, but IMO `-arch ppc` is a single arg (either won't make sense on its own). But I will try your suggestion when I return to the Intel machine.

tristan957:
`-arch` seems like the option and `ppc` seems like the value to me, so two separate args.

Think about these args from the perspective of `main()` in C. If a C program were given what you originally had, `a.out "-arch ppc"`, argc would equal 2 and argv[1] would equal "-arch ppc".

barracuda156:
> `-arch` seems like the option and `ppc` seems like the value to me, so two separate args.
> 
[...]

PR #10442, opened 2022-05-28
mesonbuild: fix ppc64 case for 10.5 and bring ppc/ppc64 trials to consistency

barracuda156:
This PR:

1. Fixes a build error on Leopard for `ppc64`. An example, where `gobject-introspection` fails to build for `ppc`+`ppc64` due to `mesonbuild` error: https://github.com/macports/macports-ports/pull/14909#issuecomment-1134971290
2. Brings trials for `ppc` and `ppc64` into consistency, so that we don’t get “power macintosh” as a cpu value.

Tested with the case of `gobject-introspection` build cited above: with these changes `meson` works as it should.

### DIFF
--- a/mesonbuild/environment.py
+++ b/mesonbuild/environment.py
@@ -316,9 +316,12 @@ def detect_cpu_family(compilers: CompilersDict) -> str:
         trial = 'aarch64'
     elif trial.startswith('arm') or trial.startswith('earm'):
         trial = 'arm'
-    elif trial.startswith(('powerpc64', 'ppc64')):
-        trial = 'ppc64'
-    elif trial.startswith(('powerpc', 'ppc')) or trial in {'macppc', 'power macintosh'}:
+    elif trial.startswith(('powerpc', 'ppc')):
+        if '64' not in trial:
+            trial = 'ppc'
+        else:
+            trial = 'ppc64'
+    elif trial in {'macppc', 'power macintosh'}:
         trial = 'ppc'
     elif trial in ('amd64', 'x64', 'i86pc'):
         trial = 'x86_64'
@@ -382,6 +385,13 @@ def detect_cpu(compilers: CompilersDict) -> str:
             trial = 'aarch64'
     elif trial.startswith('earm'):
         trial = 'arm'
+    elif trial.startswith(('powerpc', 'ppc')):
+        if '64' not in trial:
+            trial = 'ppc'
+        else:
+            trial = 'ppc64'
+    elif trial in {'macppc', 'power macintosh'}:
+        trial = 'ppc'
     elif trial == 'e2k':
         # Make more precise CPU detection for Elbrus platform.
         trial = platform.processor().lower()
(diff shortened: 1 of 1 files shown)


barracuda156:
@jpakkane Could you please help me a bit?

Original problem which I solve in my PR is that `meson` identifies `power macintosh` with `ppc` arch, and that leads to errors when the build is for `ppc64`: `mesonbuild` fails, thinking that it needs `exe_wrapper`. I added a condition specifying that macOS PPC is able to run `ppc64` binaries. That solved the original problem (a particular example of it see in the linked Macports PR).

However this does not help with `unittests` – and now some of the bots report a test failure:
```
        with mock.patch('mesonbuild.environment.any_compiler_has_define', mock.Mock(return_value=True)):
            for test, expected in [('x86_64', 'i686'), ('aarch64', 'arm'), ('ppc', 'ppc64')]:
                with self.subTest(test, has_define=True), mock_trial(test):
                    actual = mesonbuild.environment.detect_cpu({})
>                   self.assertEqual(actual, expected)
E                   AssertionError: 'ppc' != 'ppc64'
E                   - ppc
E                   + ppc64
E                   ?    ++

unittests/internaltests.py:1573: AssertionError
```
Of course we could just disable it for PPC, but maybe there is another way.

jpakkane:
Is ppc like the case of x86 and x86_64 where both live at the same time in the same userland or are there separate OS installs for each of them?

barracuda156:
> Is ppc like the case of x86 and x86_64 where both live at the same time in the same userland or are there separate OS installs for each of them?

@jpakkane On Darwin/MacOS, they co-exist. The last official release of macOS for PPC, 10.5.x (Leopard) supports building and running both natively.

Normally this does not cause any problem, since you can specify `powerpc-apple-darwin` and `powerpc64-apple-darwin`. With `meson` there is an issue, since it uses some obscure machine-related info, and identifies the build machine as `power macintosh`, which then falls into `ppc` cpu family. So when you try to build for `ppc64`, it thinks binaries require `exe_wrapper`.

The hack that `meson` definitions use for AIX does not work in this case. What I eventually did and submitted here fixes the issue in practical terms – `meson` does not fail when building `+universal`. This may not be an optimal solution of course.

P. S. I have to notice that `gobject-introspection` + `meson` combo is generally problematic. It took me quite a while until I was able to build it for `ppc` under Rosetta – again `meson` kept failing for every reason (wrong arch, wrong linker etc.): https://github.com/mesonbuild/meson/issues/10351

barracuda156:
@jpakkane Let me clarify a bit. Running ppc64 natively on MacOS requires two things: 10.5.8 (which supports ppc64 on OS level) and ppc64-capable hardware (G5, or ppc970). This sounds restrictive, but it’s perhaps the most common case, since G4 are collectors items now, while G5 are still capable.

jpakkane:
The proper fix would probably be the same as with x86_64, that is, detecting if the current OS has a 64 bit kernel (or however it is detected) and returning ppc64 in that case (I'm assuming that that is the toolchain default, of course).

jpakkane:
If you are building for the non-native(st) case then you need to specify a cross build.That is also how 32 bit compilation happens on x86_64.

barracuda156:
> If you are building for the non-native(st) case then you need to specify a cross build.That is also how 32 bit compilation happens on x86_64.

@jpakkane Yes, but that’t not the issue. The problem is that `meson` thinks it needs `exe_wrapper`, while in fact it does not.

PR #10444, opened 2022-05-29
fix incorrectly allowed kwarg for custom_target

eli-schwartz:
override_options makes no sense for custom_target as we don't use it for anything. Also, this was added in commit c3c30d4b060239654c9b848092692ab346ebed9d despite not being allowed in permittedKwargsc3c30d4b0.

For inexplicable reasons, we had a known_kwargs for custom_target that looped over kwargs and issued a warning, not an error, for unknown kwargs. It was impossible to ever hit that check to begin with, though, ever since commit e08d73510552fa4a3a087af1a9e5fded8f5749fd which added permittedKwargs and obsoleted those manual checks with real errors.

So at one point override_options was specially permitted to be used without emitting a warning, and then for about half a decade it was an error, and then based on some dead code it was allowed again for a bit. But through all this it doesn't do anything and isn't documented.


eli-schwartz:
One more minor fix in the same theme of mismatched kwargs types.

Issue #8640, opened 2021-04-11
Relative paths for include_directories is enforced on Ubuntu but not on Windows


nanthony21:
I have a large collection of 3rd party libraries that various parts of my project need to link to. I've been doing this like so:

At the top level of my project I have an absolute path to the collection of libraries: `DEPS_REPO = join_paths(meson.current_source_dir(), '../../../3rdPartyRepo)`

Having this as an absolute path is nice since:
 - `find_library` requires an absolute path for its `dirs` argument.
 - The path remains valid even when using it from nested subdirectories in my project. 

I then do this:
```
SDK_dir = join_paths(DEPS_REPO, 'path_to_specific_sdk')
SDK_lib = meson.compiler('cpp').find_library('sdklibname', dirs: join_paths(SDK_dir, 'lib64'), static: true)
SDK_inc = include_directories(join_paths(SDK_dir, 'include'))
library('MyDriver', sources, include_directories: SDK_inc, dependencies: SDK_dep, install: true)
```

This works fine for me on Windows with Meson 0.57.1, however with the same version of Meson on Ubuntu 20.04 I get `ERROR: Tried to form an absolute path to a source dir. You should not do that but use relative paths instead.`

Aside from the issue of getting inconsistent behavior between OS's I don't understand why these restrictions are in place. It would be nice if both `include_directories` and `find_library(..., dir:` would allow the usage of both relative and absolute paths. 

Is there an entirely different way that I should go about this? I had considered using `declare_dependency` since it would allow me to bundle the linker path and the include directory into a single dependency object but it seemed like this was going against the intended behavior since these aren't dependencies which will be used in other parts of the project, as soon as they are created they are passed to `library(..., dependencies:` and never used again.

As  final note, I think it would be really useful if the "FS" module had functionality similar to Python's `os.path.abspath` and `os.path.relpath` for converting paths as needed.

tristan957:
Seems like we should close this

PR #10445, opened 2022-05-29
Improve error message for `include_directories(abs_path_intree)`

rgommers:
This error message was quite confusing when triggered by use of an absolute path to the include dir of an external dependency (numpy in my case). Changing that to a relative dir also isn't a solution, because Meson will *not* do the "busywork to make paths work" that the error message says it will.

[ci skip]

This came up in the SciPy build in https://github.com/scipy/scipy/issues/16312. 

Note that gh-8640 is related. I think the issue description there is probably a little confused, but it looks very similar - no follow-up there, so probably can be closed?


rgommers:
@eli-schwartz would you mind having a look at this? It's probably going to save other Python devs some head-scratching.

eli-schwartz:
(I would squash the two commits together.)

rgommers:
> (I would squash the two commits together.)

Sorry, I forgot about that guideline - I'd normally squash-merge. There's one test failure which doesn't look related but also doesn't happen on other CI runs right now (it may be a flake, [log](https://github.com/mesonbuild/meson/runs/6908573056?check_suite_focus=true)):
```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 166: invalid start byte
```
So let me squash the commits and re-run tests.

eli-schwartz:
Yeah. I'm pretty sure I've seen that exact issue quite a few times and it's a flake.

dcbaker (approved):

PR #10449, opened 2022-06-01
Fix sandbox violation when using subproject as a symlink

### LINKED ISSUE #8827: subproject symbol link triggers Sandbox violation: outside current (sub) project.
**Describe the bug**
A subproject implemented as a symbolic link to another directory causes a false Sandbox violation.

**To Reproduce**
Create a symbolic link under subprojects/ which links to a directory containing a typical subproject with source files.

For example:

```
/top/main_project/subdir/meson.build
/top/main_project/subdir/src/main.cpp
/top/main_project/subprojects/mysubproject --> /top/main_project/subdir
```

Run 'meson setup <builddir>'
Observe:

`subprojects/mysubproject/meson.build:30:0: ERROR: Sandbox violation: Tried to grab file main.cpp outside current (sub)project.
`
The code that triggers the error is
in mesonbuild/interpreter/interpreter.py:

```
  2681         if project_root not in norm.parents:
  2682             raise InterpreterException(f'Sandbox violation: Tried to grab {inputtype} {norm.name} outside current (sub)project.')
```
based on the following values:

```
mysubproject| srcdir: <path to top-level dir> e.g  /top/main_project
mysubproject| norm: <absolute path to source file> e.g. /top/main_project/subdir/src/main.cpp
mysubproject| project_root: <absolute path to subproject mysubproject symbolic link> e.g  /top/main_project/subprojects/mysubproject
```
Note that the (sub) project_root, which is a symbolic link is not ".resolve()" 'd like the norm path is.
And the following change allows this to work as expected.

```
< 2680         project_root = Path(srcdir, self.root_subdir)
> 2680         project_root = Path(srcdir, self.root_subdir).resolve()

```

**Expected behavior**
I expect to be able to use symbolic links to subproject directories without a Sandbox violation.

**system parameters**
*  plain native build (for the same computer)
* Debian 7
* Python 3.6
* `meson --version` 0.58.0, but started with at least 0.57.2
* `ninja --version`1.8.2

Thanks
-tim

vvainola:
Fix "Tried to grab file outside current (sub)project" error when subproject exists within
a source tree but it is used through a symlink. Using subprojects as symlinks is very useful
feature when migrating an existing codebase to meson that all sources do not need to be
immediately moved to subprojects folder.

fixes #8827


unittests/allplatformstests.py
```diff
@@ -4263,3 +4263,16 @@ def test_custom_target_name(self):
...
             self.assertIn('Generating subdir/file.txt with a custom command', out)
+
+    def test_symlinked_subproject(self):
+        testdir = os.path.join(self.unit_test_dir, '106 subproject symlink')
+        subproject = os.path.join(testdir, 'symlinked_subproject')
+        symlinked_subproject = os.path.join(testdir, 'subprojects', 'symlinked_subproject')
+        if os.path.exists(symlinked_subproject):
+            os.remove(symlinked_subproject)    
+        os.symlink(subproject, symlinked_subproject)
+
```
dcbaker, line 4274: The better thing to do here is:
```python
os.symlink(subproject, symlinked_subproject)
self.addCleanup(os.remove, symlinked_subproject)
```

Which is guaranteed to get called even if there's an exception in `self.init` or `self.build`, unlike the current code which needs the check to ensure that the symlink was removed in the last run.

test cases/unit/106 subproject symlink/symlinked_subproject/src.c
```diff
@@ -0,0 +1,4 @@
+int foo()
```
dcbaker, line 1: I think one of the CI passes is giong to get grump that this isn't `int foo(void)`

dcbaker (changes requested):
Thank you so much for looking at this, and for creating a test!

I've got a couple of little things, but overall looks great.

dcbaker (approved):
Thanks, LGTM!

mesonbuild/interpreter/interpreter.py
```diff
@@ -2851,7 +2851,7 @@ def validate_installable_file(fpath: Path) -> bool:
             # subproject files, as long as they are scheduled to be installed.
             if validate_installable_file(norm):
                 return
-        norm = Path(srcdir, subdir, fname).resolve()
+        norm = Path(os.path.abspath(Path(srcdir, subdir, fname)))
```
eli-schwartz, line 2854: Yet again, pathlib proves that it sucks horribly at anything other than PurePath.

eli-schwartz:
That's supposed to not work.

What's supposed to work is a subproject invoking files() on files that belong to the subproject.

You've never been allowed to invoke files() on a string that points to a ***different*** subproject.

Apparently the same bug that this PR fixed, also *allowed* the latter case until this PR made the check operate correctly? That's a good thing, then.

eli-schwartz:
On second thought, does your setup fail with meson 0.62.2 or with git master?

eli-schwartz:
So then that has nothing to do with this PR, which was backported to 0.62.2

In git master, the standard sandbox violation checks which formerly only validated e.g. kwargs to a build target /custom_target, now validate the files() function as well.

This is the change that is affecting you.

...

...

> Imagine e.g. the subproject containing a directory full of files. From those files, that root project will pick and choose some to be used e.g. for [...]

The *reason* someone chooses to do a sandbox violation doesn't really matter unless you're trying to argue in favor of the sandbox being inherently wrong and that it should be wholly removed.

As of today, Meson still considers the sandbox to be useful and thus it is present.

> Another thing that worries me is is that if this subproject's contents are installed on the build host the directory and the files within are now accessed via pkgconfig files and we are thus using strings for paths.

https://mesonbuild.com/Dependencies.html#dependencies-that-provide-resource-files

I'm the one who closed the overlooked `files()` loophole. I took great care to ensure beforehand, that pkg-config/subproject fallback files still work. I added test cases for it and everything.

That too is backported to 0.62.2, so passing a subproject fallback dependency file directly as a kwarg to a custom_target now *started* working and it is no longer necessary to evade the sandbox by wrapping kwargs in `file()` (although the latter still works until 0.63.0).

PR #10456, opened 2022-06-03
Try to clarify what modules are found/missing

chiroptical:
I recently used the meson build system in a relatively simple project that used python optionally. I received the following output and found it confusing. Here is the relevant section of `meson.build`

```
testing_py = py.find_installation('python3',
  modules: ['cffi', 'hypothesis', 'numpy', 'pandas', 'pytest_benchmark'],
  required: false
  )
```

I am missing `hypothesis` and `pytest_benchmark`, running `meson build` I get the following:

```
...
Program python3 (cffi, hypothesis, numpy, pandas, pytest_benchmark) found: NO modules: cffi, numpy, pandas
...
```

I interpreted this as `cffi, numpy, pandas` were not found. I then installed them and found they were already installed and eventually realized that I was actually missing `hypothesis`/`pytest_benchmark`.

@eli-schwartz in another repository suggested that I make a PR. So, I modified the output a bit to try to clarify missing/found and whether or not the installation is even required.

## Output Examples

These outputs are generating with a section like the following,

```python
testing_py = py.find_installation('python3.9',    
  modules: ['cffi', 'hypothesis', 'numpy', 'pandas', 'pytest_benchmark'],     
  # modules: [],  
  required: false,    
  )    
```

- `python3.9` does **not** exist on my system, `python3.10` does exist
- I ask for no dependencies or install/remove modules for various outputs

### python3.9/no modules

```
Runtime dependency python3.9 found: NO
```

### python3.9/with modules

```
Runtime dependency python3.9 found: NO
```

### python3.10/no modules

```
Runtime dependency python3.10 found: YES (/usr/bin/python3.10)
```

### python3/no modules (this will be 3.10 on my system)

```
Runtime dependency python3 found: YES (/usr/bin/python3)
```

### python3.10/some modules missing

```
Runtime dependency python3.10 (found: cffi, numpy, pandas | missing: hypothesis, pytest_benchmark) found: NO
```

### python3.10/all modules found

```
Runtime dependency python3.10 (found: cffi, hypothesis, numpy, pandas, pytest_benchmark) found: YES (/usr/bin/python3.10)
```

### DIFF
--- a/mesonbuild/modules/python.py
+++ b/mesonbuild/modules/python.py
@@ -755,13 +755,18 @@ class PythonModule(ExtensionModule):
         msg: T.List['mlog.TV_Loggable'] = ['Program', python.name]
         if want_modules:
             msg.append('({})'.format(', '.join(want_modules)))
-        msg.append('found:')
+        msg.append('required:')
+        if required:
+            msg.append(mlog.green('YES'))
+        else:
+            msg.append(mlog.red('NO'))
+        msg.append('configured:')
         if python.found() and not missing_modules:
             msg.extend([mlog.green('YES'), '({})'.format(' '.join(python.command))])
         else:
-            msg.append(mlog.red('NO'))
+            msg.extend([mlog.red('NO'), 'missing: {}'.format(', '.join(missing_modules))])
         if found_modules:
-            msg.append('modules:')
+            msg.append('found:')
             msg.append(', '.join(found_modules))
 
         mlog.log(*msg)
--- a/mesonbuild/modules/python.py
+++ b/mesonbuild/modules/python.py
@@ -755,18 +755,13 @@ class PythonModule(ExtensionModule):
         msg: T.List['mlog.TV_Loggable'] = ['Program', python.name]
         if want_modules:
             msg.append('({})'.format(', '.join(want_modules)))
-        msg.append('required:')
-        if required:
-            msg.append(mlog.green('YES'))
-        else:
-            msg.append(mlog.red('NO'))
-        msg.append('configured:')
+        msg.append('found:')
         if python.found() and not missing_modules:
             msg.extend([mlog.green('YES'), '({})'.format(' '.join(python.command))])
         else:
             msg.extend([mlog.red('NO'), 'missing: {}'.format(', '.join(missing_modules))])
         if found_modules:
-            msg.append('found:')
+            msg.append('modules:')
             msg.append(', '.join(found_modules))
 
         mlog.log(*msg)
--- a/mesonbuild/modules/python.py
+++ b/mesonbuild/modules/python.py
@@ -752,17 +752,22 @@ class PythonModule(ExtensionModule):
                 else:
                     found_modules.append(mod)
 
-        msg: T.List['mlog.TV_Loggable'] = ['Program', python.name]
-        if want_modules:
-            msg.append('({})'.format(', '.join(want_modules)))
+        msg: T.List['mlog.TV_Loggable'] = ['Runtime dependency', display_name]
+
+        found_modules_msg = ', '.join(found_modules)
+        missing_modules_msg = ', '.join(missing_modules)
+        if found_modules_msg and missing_modules_msg:
+            msg.append("(found: {} | missing: {})".format(found_modules_msg, missing_modules_msg))
+        elif found_modules_msg:
+            msg.append("(found: {})".format(found_modules_msg))
+        elif missing_modules_msg:
+            msg.append("(missing: {})".format(missing_modules_msg))
+
         msg.append('found:')
         if python.found() and not missing_modules:
             msg.extend([mlog.green('YES'), '({})'.format(' '.join(python.command))])
         else:
-            msg.extend([mlog.red('NO'), 'missing: {}'.format(', '.join(missing_modules))])
-        if found_modules:
-            msg.append('modules:')
-            msg.append(', '.join(found_modules))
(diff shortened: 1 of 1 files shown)


eli-schwartz:
Thanks for looking at this!

Elsewhere we use "found" to indicate whether the program was found, so I'm not completely sure about changing that here to "configured" in order to use "found" for modules. Maybe it's sufficient to just list which modules were missing? e.g.

```
Program python3 (cffi, hypothesis, numpy, pandas, pytest_benchmark) required: NO found: NO missing: hypothesis, pytest_benchmark
```

Also there's no need to specify whether it was required, I think. This isn't directly interesting information, and if it isn't found then the build will fail, which demonstrates that it is required well enough.

mesonbuild/modules/python.py
```diff
@@ -759,7 +759,7 @@ def find_installation(self, state: 'ModuleState', args: T.Tuple[T.Optional[str]]
         if python.found() and not missing_modules:
             msg.extend([mlog.green('YES'), '({})'.format(' '.join(python.command))])
         else:
-            msg.append(mlog.red('NO'))
+            msg.extend([mlog.red('NO'), 'missing: {}'.format(', '.join(missing_modules))])
```
chiroptical, line 762: @eli-schwartz reduced this down to the simplest change.

eli-schwartz: I think we only want to extend it with the "missing" text if `missing_modules`, but this branch is also taken when `python.found()` is true.

So perhaps:
```python
msg.append(mlog.red('NO'))
if not missing_modules:
    msg.append('missing: {}'.format(', '.join(missing_modules)))
```

eli-schwartz:
While we are at it, I decided to take a look at the consistency of the current state, and the proposed new state, across various parts of Meson.

```meson
import('python').find_installation(modules: ['html5lib', 'regex'], required: false)
import('python').find_installation(modules: ['html5lib', 'regex', 'gooo', 'flop'], required: false)


dependency('qt5', modules: ['Core', 'Xml'], required: false)
dependency('qt5', modules: ['Core', 'Xml', 'gargle'], required: false)

dependency('boost', modules : ['thread', 'regex'], required: false)
dependency('boost', modules : ['thread', 'regex', 'utility'], required: false)

dependency('llvm', modules: ['amdgpu', 'core'], required: false, method: 'config-tool')
dependency('llvm', modules: ['amdgpu', 'core', 'puzzle'], required: false, method: 'config-tool')
```
produces:
```
Program python3 (html5lib, regex) found: YES (/usr/bin/python) modules: html5lib, regex
Program python3 (html5lib, regex, gooo, flop) found: NO modules: html5lib, regex


Run-time dependency qt5 (modules: Core, Xml) found: YES 5.15.2 (pkg-config)
Run-time dependency qt5 (modules: Core, Xml, gargle) found: NO (tried pkgconfig and config-tool)

Run-time dependency Boost (found: regex, thread) found: YES 1.78.0 (/usr)
Run-time dependency Boost (found: regex, thread | missing: utility) found: NO (tried system)

Run-time dependency LLVM (modules: amdgpu, core) found: YES 13.0.1
Run-time dependency LLVM (modules: amdgpu, core, puzzle(missing)) found: NO (tried config-tool)
```

We have four different ways (at least) to message about partially found dependencies.

- python: list requested modules at beginning, list found modules at end
- qt5: list requested modules at beginning, and merely state whether the overall lookup was successful
- boost: list `found: ... | missing: ...` at beginning
- llvm: list requested modules at beginning, with missing ones replaced by `XXX -> XXX(missing)`

This PR makes the following change to python:

```
# old
Program python3 (html5lib, regex) found: YES (/usr/bin/python) modules: html5lib, regex
Program python3 (html5lib, regex, gooo, flop) found: NO modules: html5lib, regex

# new
Program python3 (html5lib, regex) found: YES (/usr/bin/python3) modules: html5lib, regex
Program python3 (html5lib, regex, gooo, flop) found: NO missing: gooo, flop modules: html5lib, regex
```

I think we badly need some consistency. The state of python before this PR, and the state of Qt5, are both subpar, and boost/llvm both seem pretty decent though I'm not sure which is better.

The state of python after this PR could use some tweaking, I think. It ends up listing every module twice in the log message, while both boost/llvm have interesting ideas for deduplicating that.

Any thoughts on a preference?

gregsskyles:
FWIW, as a non-involved third-party interloper, I would rate the Boost method of reporting the found and not found modules as the most readable and understandable.

chiroptical:
While messing with the boost style of printing I wanted to try a case where `python` didn't exist on the system. I was able to generate the following wonky message,

mesonbuild/modules/python.py
```diff
@@ -752,17 +752,22 @@ def find_installation(self, state: 'ModuleState', args: T.Tuple[T.Optional[str]]
                 else:
                     found_modules.append(mod)
 
-        msg: T.List['mlog.TV_Loggable'] = ['Program', python.name]
-        if want_modules:
-            msg.append('({})'.format(', '.join(want_modules)))
+        msg: T.List['mlog.TV_Loggable'] = ['Runtime dependency', display_name]
```
chiroptical, line 755: See comment: https://github.com/mesonbuild/meson/pull/10456#issuecomment-1158809399 for the reasoning behind `s/python.name/display_name`

chiroptical: Using style from Boost a la https://github.com/mesonbuild/meson/pull/10456#issuecomment-1152874592

PR #10457, opened 2022-06-03
Improve WINEPATH reduction

xclaesse:
- Remove duplicated code in mdevenv.py
- Change the limit to 1024 instead of 2048 which is what has been
  tested.
- Skip shortening if it is already short enough.
- Skip shortening with wine >= 6.4 which does not seems to have that
  limitation any more.
- Downgrade exception to warning in the case WINEPATH cannot be
  shortened under 1024 chars, it is possible that it will still work.

### DIFF
--- a/mesonbuild/mdevenv.py
+++ b/mesonbuild/mdevenv.py
@@ -6,7 +6,7 @@ import itertools
 
 from pathlib import Path
 from . import build, minstall, dependencies
-from .mesonlib import MesonException, RealPathAction, is_windows, setup_vsenv, OptionKey, quote_arg
+from .mesonlib import MesonException, RealPathAction, is_windows, setup_vsenv, OptionKey, quote_arg, get_wine_shortpath
 from . import mlog
 
 import typing as T
@@ -28,43 +28,6 @@ def get_windows_shell() -> str:
     result = subprocess.check_output(command)
     return result.decode().strip()
 
-def get_wine_shortpath(build_dir: str, winecmd: str, wine_paths: T.List[str]) -> T.List[str]:
-    '''
-    WINEPATH size is limited to 1024 bytes which can easily be exceeded when
-    adding the path to every dll inside build directory. See
-    https://bugs.winehq.org/show_bug.cgi?id=45810.
-
-    To shorthen it as much as possible we use path relative to builddir
-    where possible and convert absolute paths to Windows shortpath (e.g.
-    "/usr/x86_64-w64-mingw32/lib" to "Z:\\usr\\X86_~FWL\\lib").
-    '''
-    rel_paths = []
-    abs_paths = []
-    builddir = Path(build_dir)
-    for p in wine_paths:
-        try:
-            rel = Path(p).relative_to(builddir)
-            rel_paths.append(str(rel))
-        except ValueError:
-            abs_paths.append(p)
-    if not abs_paths:
-        return rel_paths
-    with tempfile.NamedTemporaryFile('w', suffix='.bat', encoding='utf-8', delete=False) as bat_file:
-        bat_file.write('''
-        @ECHO OFF
-        for %%x in (%*) do (
-            echo|set /p=;%~sx
-        )
-        ''')
-    try:
-        stdout = subprocess.check_output([winecmd, 'cmd', '/C', bat_file.name] + abs_paths,
-                                         encoding='utf-8', stderr=subprocess.DEVNULL)
-        return rel_paths + [p for p in set(stdout.split(';')) if p]
-    except subprocess.CalledProcessError:
-        return rel_paths + abs_paths
-    finally:
-        os.unlink(bat_file.name)
-
 def reduce_winepath(build_dir: str, env: T.Dict[str, str]) -> None:
     winepath = env.get('WINEPATH')
     if not winepath:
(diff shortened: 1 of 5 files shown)


mesonbuild/mesonlib/universal.py
```diff
@@ -1887,34 +1886,57 @@ def __call__(self, parser: argparse.ArgumentParser, namespace: argparse.Namespac
...
+
+    # Check if we have wine >= 6.4
     try:
-        with open(os.devnull, 'w', encoding='utf-8') as stderr:
-            wine_path = subprocess.check_output(
-                winecmd +
-                ['cmd', '/C', getShortPathScript] + wine_paths,
-                stderr=stderr).decode('utf-8')
+        stdout = subprocess.check_output([winecmd, '--version'], encoding='utf-8')
+        match = re.search(r'([0-9][0-9\.]+)', stdout)
```
eli-schwartz, line 1911: `programs.ExternalProgram('wine', winecmd).get_version()`  adds a much nicer UX for checking versions, including a single unified regex.

xclaesse: But it takes an interpreter in args which we don't have. Would have to refactor that further.

eli-schwartz: Yeah. :(

xclaesse: Made interpreter optional and use Popen_safe() in the case it's not provided.

eli-schwartz: Can we split that into its own commit? :)

xclaesse: done.

mesonbuild/programs.py
```diff
@@ -100,16 +100,25 @@ def description(self) -> str:
...
+                res = interpreter.run_command_impl(interpreter.current_node, (self, ['--version']),
+                                                   {'capture': True,
+                                                    'check': True,
+                                                    'env': build.EnvironmentVariables()},
+                                                   True)
+                o, e = res.stdout, res.stderr
+            else:
+                p, o, e = mesonlib.Popen_safe(raw_cmd)
+                if p.returncode != 0:
+                    cmd_str = ' '.join(raw_cmd)
```
eli-schwartz, line 117: I would use `mesonlib.join_args()` here.

mesonbuild/mdevenv.py
```diff
@@ -28,54 +28,14 @@ def get_windows_shell() -> str:
...
-    abs_paths = []
-    builddir = Path(build_dir)
-    for p in wine_paths:
-        try:
-            rel = Path(p).relative_to(builddir)
-            rel_paths.append(str(rel))
-        except ValueError:
-            abs_paths.append(p)
-    if not abs_paths:
-        return rel_paths
```
eli-schwartz, line 51: Is this no longer needed? It doesn't seem to be present in the consolidated version anymore.

mesonbuild/programs.py
xclaesse: It's just for logging the error, but better use it indeed. Done.

mesonbuild/mdevenv.py
xclaesse: It was using paths relative to top builddir for devenv which has the annoying effect that you can only run wine from that directory. I was fine with that limitation because I thought we had no other choice, but now that we found that newer wine got fixed, I think it's better to not keep such fragile workaround. Also, we cannot rely on that hack for other use-cases because they can have a different workdir.

eli-schwartz: I was actually thinking it would work better in the non-devenv case since there's a set workdir. e.g. maybe it could take a `cwd=None` kwarg and make it relative to that?

xclaesse: Good point, did that now. But I noticed we set `Z:` prefix on all paths. That's probably going to break the relative path detection. Not sure why we have that, seems to work fine without `Z:` in devenv.

xclaesse: Yes, I think it's better to remove that, it creates issues that users won't understand.

xclaesse: Removed from devenv now.

mesonbuild/mesonlib/universal.py
```diff
@@ -1888,35 +1887,71 @@ def __call__(self, parser: argparse.ArgumentParser, namespace: argparse.Namespac
...
-        print("Could not get short paths: %s" % e)
-        wine_path = ';'.join(wine_paths)
-    finally:
-        os.remove(getShortPathScript)
-    if len(wine_path) > 2048:
-        raise MesonException(
-            'WINEPATH size {} > 2048'
-            ' this will cause random failure.'.format(
-                len(wine_path)))
+    This limitation reportedly have been fixed with wine >= 6.4
```
eli-schwartz, line 1901:
```suggestion
    This limitation reportedly has been fixed with wine >= 6.4
```

eli-schwartz (approved):
LGTM with a couple extremely minor typo nits.

mesonbuild/mesonlib/universal.py
```diff
@@ -1888,35 +1887,71 @@ def __call__(self, parser: argparse.ArgumentParser, namespace: argparse.Namespac
...
-    """
+def get_wine_shortpath(winecmd: T.List[str], wine_paths: T.Sequence[str],
+                       workdir: T.Optional[str] = None) -> str:
+    '''
+    WINEPATH size is limited to 1024 bytes which can easily be exceeded when
+    adding the path to every dll inside build directory. See
+    https://bugs.winehq.org/show_bug.cgi?id=45810.
 
-    wine_paths = list(OrderedSet(wine_paths))
+    To shorthen it as much as possible we use path relative to `workdir`
```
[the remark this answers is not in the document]
xclaesse: done

PR #10458, opened 2022-06-03
Here is another set of stand-alone changes pulled out of my `BuildTarget` annotations series

dcbaker:
I'm starting to get down to the nitty gritty, but here is a set of standa lone changes that can be reviewed and landed before the main changes, which, again, should make such a review easier by reducing the number of tangentally related pathces in the main one. This is *mosly* typoing stuff, though there are some helpers patches and some things that mypy catches in a fully typed build.py, but can be fixed without fully typing it.


mesonbuild/build.py
```diff
@@ -2451,10 +2463,7 @@ def __init__(self,
         self.install = install
         self.install_dir = list(install_dir or [])
         self.install_mode = install_mode
-        _install_tag: T.List[T.Optional[str]] = [None] if not install_tag else stringlistify(install_tag)
```
eli-schwartz, line 2454: We seem to be dropping the stringlistify here, and instead offering a typing violation by saying that `install_tag: install_tag: T.Optional[...]` is passed to a function that accepts `T.List[T.Optional[str]]`.

eli-schwartz (commented):
I wish this had waited until I had time to review it properly...

There's a couple of small issues here, and one thing that confuses me.

mesonbuild/backend/vs2010backend.py
```diff
@@ -270,12 +270,15 @@ def get_obj_target_deps(self, obj_list):
...
+    def get_target_deps(self, t: T.Dict[T.Any, build.Target], recursive=False):
+        all_deps: T.Dict[str, build.Target] = {}
         for target in t.values():
             if isinstance(target, build.CustomTarget):
                 for d in target.get_target_dependencies():
-                    all_deps[d.get_id()] = d
+                    # FIXME: this isn't strictly correct, as the target doesn't
+                    # Get dependencies on non-targets, such as Files
+                    if isinstance(d, build.Target):
+                        all_deps[d.get_id()] = d
```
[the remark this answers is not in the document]
eli-schwartz: Excellent point, I think it confused me because the typing fix came after the change I commented on. :)

mesonbuild/build.py
dcbaker: Yeah... that `stringlistify` is totally wrong. the interpreter accepts `List[str | bool]` for booleans it converts it to `None`. So you can get a list containing a mixture of str and None, and that's supposed to be valid. (you wrote that code).

eli-schwartz: What the interpreter does is completely irrelevant, what matters is what various code that creates a `build.CustomTarget(..., install_tag=)` does.

In that sense, the typing we now have is also a lie. We say `T.Optional[T.Sequence[T.Optional[str]]]`, but we actually accept

```python
T.Union[T.Optional[T.Sequence[T.Optional[str]], str]
```

Because someone in a PR tried to do an API violation and create a `build.CustomTarget(..., install_tag='man')` when the API demanded `build.CustomTarget(..., install_tag=['man'])`, and solved that by changing the API to stringlistify and allow passing a string instead of a list of strings.

This did NOT come with accompanying type annotation changes. Sorry. :p

I pointed this out at the time in this comment, which directly pinged you: https://github.com/mesonbuild/meson/pull/10304#issuecomment-1111041646

Now the change we are discussing today has broken four different functions in the gnome module, I think, which of course do not actually result in CI errors because G-d help us when it comes to testing the gnome module?

eli-schwartz: Actually hmm, we may not actually do that, as noted on IRC only build.Data gets a string value?

It remains true that the reason for the stringlistify was by someone trying to "fix" the API, and the intention was to make it *actually* accept strings, which this now reverts. I'm okay with reverting it since the type annotations do in fact forbid it and I was never sure why we want to do that at all.

So the dropping of stringlistify is fine as far as I'm concerned, and as it turns out we aren't actually testing the thing that that other PR tried to "fix", so :shrug:

eli-schwartz: Yay! Fantastic. So that other PR broke stuff.

I've gone from lukewarm apathy to a strong agreement that dropping it was absolutely the right thing to do and we really, really should not be adding more validation back into the build layer, which we removed it from for good reason.

PR #10460, opened 2022-06-04
MVSC: Allow path separator to be also forward slash after drive letter

vvainola:
Ccache normalizes the include paths from compiler stdout to use
forward slashes and rewrites the paths to relative if they are within
CCACHE_BASEDIR because ninja uses this information to detect
changes.

Arguably this should be changed in ccache that
the rewritten include paths use backwards slashes but allowing
either seems quite harmless.


vvainola:
Same change implemented in https://github.com/mesonbuild/meson/pull/10493

PR #10461, opened 2022-06-04
Change minstall and mtest use mcompile to fix rebuild for VS backend

vvainola:
- Change mtest and minstall to call mcompile for rebuild so that automatic
 rebuild works for VS backend as well.
- Explicitly call REGEN target first in msbuild compile because otherwise if
 build files have been changed, they are not properly regenerated before
 building other targets. This often lead to build errors if files were added
 to build because first build after changing used old VS projects.
 For some reason this worked without calling REGEN if the maxCpuCount
 argument is removed

- Change tests to build VS backend through 'meson compile' because this 
simplified the test scripts and improves test coverage for the command line functions.
Duplicating the build logic e.g. how to build single target with msbuild
for tests does not seem wise.

- Fix mcompile suggesting targets that do not work since correct format
is PATH/TARGET:TYPE and not PATH/FILENAME:TYPE
- Fix that run targets could not be built for any other buildtype than debug
- Fix that targets not built by default could not be built with 'meson compile' at all
- Fix custom target with object inputs on VS backend.

### DIFF
--- a/mesonbuild/backend/vs2010backend.py
+++ b/mesonbuild/backend/vs2010backend.py
@@ -1468,7 +1468,7 @@ class Vs2010Backend(backends.Backend):
         ET.SubElement(midl, 'InterfaceIdentifierFilename').text = '%(Filename)_i.c'
         ET.SubElement(midl, 'ProxyFileName').text = '%(Filename)_p.c'
         # FIXME: No benchmarks?
-        test_command = self.environment.get_build_command() + ['test', '--no-rebuild']
+        test_command = self.environment.get_build_command() + ['test']
         if not self.environment.coredata.get_option(OptionKey('stdsplit')):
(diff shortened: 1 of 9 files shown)


eli-schwartz:
I'm not sure what most of this is doing, but it makes me extremely nervous because it uses `meson compile` which IMHO should be deleted without a replacement.

Also, it seems to actually *execute* `meson compile` from python code? Which seems wrong no matter what, due to recursion.

vvainola:
Calling meson compile from python is probably wrong indeed, I will fix that. Otherwise this fixes many of the issues wrong with meson compile for VS backend. Removing the meson compile command means that the issues of handling VS building from CMD line is left to the user and it is not very convenient to figure out how to compile again when switching backend.

PR #10462, opened 2022-06-04
Fix numerous type inconsistencies in docs

annacrombie:
I wrote a [script](https://github.com/annacrombie/meson/blob/introspect_type_info/signatures.py) that uses `@typed_posargs`, `@typed_kwargs`, and mypy type annotations to extract the actual type signatures of most meson functions and methods.  I then output function and method signatures from the docs in the same format.  By diffing the output of these two sources, I found numerous inconsistencies in the types of functions.  Some of these are understandable, e.g. `project(version:)` technically accepts `list[file]` as well as the documented `file | str`, but this is just to accommodate passing `files()` to that keyword.  In other cases, I discovered entire undocumented methods, like `build_tgt.get_id()`.

I made a [repo](https://github.com/annacrombie/meson-types-diff) where you can see the diffs online.

Also, the dep.as_system() fix is unrelated, but I had it hanging around for awhile.


docs/yaml/builtins/meson.yaml
```diff
@@ -88,7 +88,7 @@ methods:
 
     posargs:
       script_name:
-        type: str | file | external_program | exe | custom_tgt | custom_idx
+        type: str | file | external_program | exe
```
jpakkane, line 91: This seems incorrect. Generating a custom script to run seems like a thing we should support and the comment below this even explicitly says that custom targets are accepted.

eli-schwartz:
> In other cases, I discovered entire undocumented methods, like build_tgt.get_id().

IIRC there has been discussion about the purpose of that method, particularly, we aren't sure why it exists or whether it should. I honestly wonder what one would wish to do with it.

docs/yaml/builtins/meson.yaml
annacrombie: I tested it, and meson throws an error in this case.  Maybe the actual type annotation needs updating?

annacrombie:
> > In other cases, I discovered entire undocumented methods, like build_tgt.get_id().
> 
> IIRC there has been discussion about the purpose of that method, particularly, we aren't sure why it exists or whether it should. I honestly wonder what one would wish to do with it.

The docs could mark the function as deprecated then.  I don't think having it totally undocumented is a good idea.

docs/yaml/builtins/meson.yaml
dcbaker: `add_install_script` should accept `CustomTarget` and `CustomTargetIndex`, if it doesn't that's a bug (and probably my fault :/)

PR #10464, opened 2022-06-04
add D generated files to order-only deps

### LINKED ISSUE #10463: Missing implicit dependency to generated code
**Describe the bug**
When using the result of `custom_target` as source to a build target, an implicit dependency (or order-only dependency?) should be added to each object file to ensure that the generated code is built before the other sources, because the other sources may depend on the generated code.

**To Reproduce**
Try to build the project https://github.com/rtbo/meson-dep-test (meson.build copied hereafter).
Compilation of `lib.d` races with compilation of `generated.d` and cannot find module `generated`.
If I manually add `generated.o` as implicit dependency of `lib.o` in `ninja.build`, everything is file.

```meson
project('meson-dep-test', 'd')

generator = executable('generator', 'generator.d')

generated = custom_target('generated',
    capture: true,
    output: 'generated.d',
    input: 'input.txt',
    command: [
        generator, '@INPUT@'
    ]
)

lib = static_library('lib', generated, 'lib.d',
    include_directories: include_directories('.'),
)
```

**Expected behavior**
Reliable build without race

**system parameters**
Arch Linux
meson master branch (0.62.99)
Python 3.10.4
Ninja 1.11.0

rtbo:
fixes #10463

### DIFF
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -833,6 +833,11 @@ class NinjaBackend(backends.Backend):
                 # people generate files with weird suffixes (.inc, .fh) that they then include
                 # in their source files.
                 header_deps.append(raw_src)
+
+        # For D language, the object of generated source files are added
+        # as order only deps because other files may depend on them
+        d_generated_deps = []
+
         # These are the generated source files that need to be built for use by
         # this target. We create the Ninja build file elements for this here
         # because we need `header_deps` to be fully generated in the above loop.
@@ -845,6 +850,8 @@ class NinjaBackend(backends.Backend):
             compiled_sources.append(s)
             source2object[s] = o
             obj_list.append(o)
+            if s.endswith('.d'):
+                d_generated_deps.append(o)
 
         use_pch = self.environment.coredata.options.get(OptionKey('b_pch'))
         if use_pch and target.has_pch():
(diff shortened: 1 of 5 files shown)


mesonbuild/backend/ninjabackend.py
```diff
@@ -845,6 +850,8 @@ def generate_target(self, target):
             compiled_sources.append(s)
             source2object[s] = o
             obj_list.append(o)
+            if s.endswith('.d'):
```
rtbo, line 853: Is there a predefined routine to identify the language of a source file?

eli-schwartz: https://github.com/mesonbuild/meson/blob/29c26d5b26397eaa3606d22d71309ecd1eb6b223/mesonbuild/compilers/compilers.py#L54-L61

It's not a routine, but it is a predefined lookup table.

dcbaker: or you could do `s.split('.')[1] in compilers.lang_suffixes['d']`

rtbo: yes OK. I used `s.split('.')[-1]` as there could be another point in the path

dcbaker (approved):

tristan957:
@dcbaker is this good to merge? Maybe you were waiting for CI to pass and just forgot about it.

PR #10470, opened 2022-06-08
visualstudio.py: Don't error out when removing /utf-8 (fix regression)

fanc999:
Hi,

From the commit message:

<i>In PR 10263, we didn't account for that we may initialize the Visual Studio-like compiler two times, once for a C compiler and once for the C++ compiler, so we end up with Meson breaking on Visual Studio 2013 or earlier, such as when building GLib.</i>

<i>Fix this by checking for a ValueError, just like what we did when /[source|execution]-charset:... is specified, , so that we do not error out when we have already removed /utf-8 from the set of always-used compiler arguments.</i>

With blessings, thank you!

### DIFF
--- a/mesonbuild/compilers/mixins/visualstudio.py
+++ b/mesonbuild/compilers/mixins/visualstudio.py
@@ -101,9 +101,7 @@ class VisualStudioLikeCompiler(Compiler, metaclass=abc.ABCMeta):
 
     # /showIncludes is needed for build dependency tracking in Ninja
     # See: https://ninja-build.org/manual.html#_deps
-    # Assume UTF-8 sources by default, but self.unix_args_to_native() removes it
-    # if `/source-charset` is set too.
-    always_args = ['/nologo', '/showIncludes', '/utf-8']
+    always_args = ['/nologo', '/showIncludes']  # type: T.List[str]
     warn_args = {
         '0': [],
         '1': ['/W2'],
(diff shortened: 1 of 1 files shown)


eli-schwartz:
> we didn't account for that we may initialize the Visual Studio-like compiler two times, once for a C compiler and once for the C++ compiler

This indicates that the original implementation was wrong because it mutated state shared between multiple compilers. The solution here isn't to catch and ignore the exception, but rather to fix this leaky behavior.

fanc999:
Hi @eli-schwartz,

OK, so I guess I will try my best in doing this.  So, maybe instead of removing the '/utf-8' flag, I only add it to `always_args` when we are using CLang or Visual Studio 2015+, when it is not already in `always_args`.

eli-schwartz:
Test failures seem relevant. In general, the underlying problem is the same whether you add them when wanted, or remove them when not wanted.

Namely, that one way or another a shared class attribute is getting modified in a manner that affects other instances, for example this could result in bad command lines where one instance *does* support /utf-8 and one does not.

fanc999:
Hi @eli-schwartz,

Thanks for the lesson in Python on shared objects...!  Let's see how CI looks at the code.

eli-schwartz:
It may be possible to set it instead as `self.always_args = self.always_args + ['/utf-8']`

fanc999:
Hi @eli-schwartz,

Indeed it is, interesting things in Python... :)

PR #10475, opened 2022-06-09
Allow for 3-component ACfL versions

stepannassyr:
Current version of ACfL is 22.0.2, which has 3 components. Current version detection assumes 2 components (and a build number), so maybe something like this could work?

### DIFF
--- a/mesonbuild/compilers/detect.py
+++ b/mesonbuild/compilers/detect.py
@@ -477,11 +477,20 @@ def _detect_c_or_cpp_compiler(env: 'Environment', lang: str, for_machine: Machin
                 exe_wrap, linker=linker, full_version=full_version)
 
         if 'Arm C/C++/Fortran Compiler' in out:
+            version = ''
             arm_ver_match = re.search(r'version (\d+)\.(\d+) \(build number (\d+)\)', out)
-            arm_ver_major = arm_ver_match.group(1)
-            arm_ver_minor = arm_ver_match.group(2)
-            arm_ver_build = arm_ver_match.group(3)
-            version = '.'.join([arm_ver_major, arm_ver_minor, arm_ver_build])
+            if arm_ver_match:
+                arm_ver_major = arm_ver_match.group(1)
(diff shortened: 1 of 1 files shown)


mesonbuild/compilers/detect.py
```diff
@@ -477,11 +477,20 @@ def sanitize(p: str) -> str:
...
+                arm_ver_minor = arm_ver_match.group(2)
+                arm_ver_build = arm_ver_match.group(3)
+                version = '.'.join([arm_ver_major, arm_ver_minor, arm_ver_build])
+            else:
+                arm_ver_match = re.search(r'version (\d+)\.(\d+)\.(\d+) \(build number (\d+)\)', out)
+                arm_ver_major = arm_ver_match.group(1)
+                arm_ver_minor = arm_ver_match.group(2)
+                arm_ver_patch = arm_ver_match.group(3)
+                arm_ver_build = arm_ver_match.group(4)
+                version = '.'.join([arm_ver_major, arm_ver_minor, arm_ver_patch, arm_ver_build])
```
dcbaker, line 493: With a bit of regex finagaling and a list comprehension you can get this to:
```python
arm_ver_match = re.search(r'version (\d+)\.(\d+)\.?(\d+)? \(build number (\d+)\)', out)
assert arm_ver_match is not None, 'for mypy'  # because mypy *should* be complaning that this could be None
version = '.'.join([x for x in arm_ver_match.groups() if x is not None])
```
with no need for an if/else and the second regex search.

I'm cheating a little with the `?` here, since in theory this would match somethign like `1.2.` (no third number), but in practice that is unlikley and it makes the list comprehension simpler.

dcbaker (changes requested):

mesonbuild/compilers/detect.py
stepannassyr: Alright, changed

dcbaker (approved):

dcbaker:
Thanks!

Issue #10311, opened 2022-04-25
[Feature Request] Support ThinLTO cache


ishitatsuyuki:
ThinLTO accepts a cache directory as linker plugin option. When cache is enabled, it's able to perform incremental optimization, accelerating iteration times for small source changes. We should make it easy to enable cache in Meson.

The options are documented at: https://clang.llvm.org/docs/ThinLTO.html#incremental

The most relevant option is `--thinlto-cache-dir`, but there are also other options to control cache pruning behavior.

Other notes:
- The way to configure cache differs between linker, so we need to detect and adapt in Meson.
- This is useful for developers but not much when you just want to get an one-off build (distro packaging etc), so it should be opt-in.

tristan957:
@ishitatsuyuki wouldn't this work better just passed as `-D<lang>_args` or `-D<lang>_link_args`?

ishitatsuyuki:
> This doesn't sound like a concept which exists in other compilers. At least I have never heard of such a thing in GCC

Yes, ThinLTO is LLVM/Clang only. But we already have `-Db_lto=thin` in Meson. It can be substituted with `-flto=thin` in `{C,CXX,LD}FLAGS`, but maybe Meson's option is worth the convenience.

The reason you would want it, is that the cache is somewhat tied to the build directory (it's rare for the emitted bitcode to be common across build configurations, let alone different code bases). Specify `-Dlang_args` with a directory different for each build would be rather annoying as a personal setup. It's also different per linker - handling this within a utility would be nicer than doing it manually.

tristan957:
I see, maybe Meson could check if -Db_lto is this and then set the cache dir to somewhere is Meson's private target directory. I don't think a PR would be very hard for that if you want to look into it.

ishitatsuyuki:
I did not mean that it should be a user function --- I meant Meson as the "command-line utility". A b_ option is likely good for this.

tristan957:
I mistyped....should be option

ishitatsuyuki:
We just need an option so that it can be opt-in --- Meson is also called from packaging routines where you only fire off the build once and the cache would be just waste of space.

tristan957:
good point! Sounds like a PR would be warranted then if you would like to submit one. I don't think it would be too difficult to create this.

PR #10481, opened 2022-06-10
Add support for incremental ThinLTO + ThinLTO with mold

ishitatsuyuki:
Closes: #10311


mesonbuild/compilers/mixins/clang.py
```diff
@@ -148,17 +149,25 @@ def get_coverage_link_args(self) -> T.List[str]:
...
-            # Thin LTO requires the use of gold, lld, ld64, or lld-link
-            if not isinstance(self.linker, (AppleDynamicLinker, ClangClDynamicLinker, LLVMDynamicLinker, GnuGoldDynamicLinker)):
-                raise mesonlib.MesonException(f"LLVM's thinLTO only works with gnu gold, lld, lld-link, and ld64, not {self.linker.id}")
+            # ThinLTO requires the use of gold, lld, ld64, lld-link or mold 1.1+
+            if isinstance(self.linker, (MoldDynamicLinker)):
+                # https://github.com/rui314/mold/commit/46995bcfc3e3113133620bf16445c5f13cd76a18
+                if not mesonlib.version_compare(self.linker.version, '>=1.1'):
+                    raise mesonlib.MesonException(f"LLVM's ThinLTO requires mold 1.1+")
+            elif not isinstance(self.linker, (AppleDynamicLinker, ClangClDynamicLinker, LLVMDynamicLinker, GnuGoldDynamicLinker)):
+                raise mesonlib.MesonException(f"LLVM's ThinLTO only works with gnu gold, lld, lld-link, and ld64, not {self.linker.id}")
```
tristan957, line 158: Maybe remove the mention of GNU here? Or capitalize it.

tristan957:
Sucks that a lot of functionality exists only in the ninja backend due to the way the backends are abstracted.

Thanks for submitting the PR! You requested the feature and you put in the work. Awesome stuff.

mesonbuild/compilers/mixins/clang.py
ishitatsuyuki: Message updated.

ishitatsuyuki:
Thanks for the kind words! I've addressed your comment.

jpakkane:
Instead of adding a new feature, could we make it always automatically work?

ishitatsuyuki:
As mentioned in the issue, I made this an opt-in because it’s pure overhead for one-time builds (e.g. packaging).

eli-schwartz:
This should probably get release notes, particularly for adding a cache.

Hmm, ccache does work even after a --wipe because it inherently uses a per-user cache, not a per-build cache. Would it make sense to accept a path for thinlto in addition to true/false?

ishitatsuyuki:
> This should probably get release notes, particularly for adding a cache.

Will do.

> Hmm, ccache does work even after a --wipe because it inherently uses a per-user cache, not a per-build cache. Would it make sense to accept a path for thinlto in addition to true/false?

Sure, I'll get another option in to specify that.

docs/markdown/Builtin-options.md
```diff
@@ -123,23 +123,24 @@ no options.
...
+| b_coverage      | false          | true, false                                                   | Enable coverage tracking                                                       |
+| b_lundef        | true           | true, false                                                   | Don't allow undefined symbols when linking                                     |
+| b_lto           | false          | true, false                                                   | Use link time optimization                                                     |
+| b_lto_threads   | 0              | Any integer*                                                  | Use multiple threads for lto. *(Added in 0.57.0)*                              |
+| b_lto_mode      | default        | default, thin                                                 | Select between lto modes, thin and default. *(Added in 0.57.0)*                |
+| b_thinlto_cache | default        | true, false                                                   | Enable LLVM's ThinLTO cache for faster incremental builds. *(Added in 0.63.0)* |
```
Tachi107, line 136: Here the default value is "default". Did you mean to write "false"?

ishitatsuyuki:
Sorry for the delay, I've now added an option to specify the cache directory. (Plus a version bump for "added in" docs to compensate for the delay :stuck_out_tongue:)

One thing I doesn't understand is that the `-Wl,--thinlto-cache-dir=<dir>` option is only updated by `--wipe`, not `--reconfigure`. Any idea what I can do to fix that?

ishitatsuyuki:
It has been a week since I last updated the patch. Could we make progress on getting a review consensus? Any idea about the `--reconfigure` issue I mentioned?

eli-schwartz:
I don't know what the reconfigure issue is and I can't investigate right now, but have you tried printf debugging along the call route you added and checking what changes during a reconfigure?

tristan957:
@dcbaker maybe you know something about this?

ishitatsuyuki:
Do we have any concerns apart from the general reconfigure stuff (which seems to be an old thing)? I'd like to get this merged, so that I don't need to source my venv everytime I want to reconfigure my local Mesa build.

eli-schwartz:
> This should probably get release notes, particularly for adding a cache.

ishitatsuyuki:
I added an changelog entry.

PR #10489, opened 2022-06-14
move various unused typing-only imports into type-checking blocks

eli-schwartz:


mesonbuild/compilers/compilers.py
```diff
@@ -24,7 +25,7 @@
 from .. import mesonlib
 from ..mesonlib import (
     HoldableObject,
-    EnvironmentException, MachineChoice, MesonException,
+    EnvironmentException, MesonException,
```
eli-schwartz, line 28: *cries*

```
  File "/__w/meson/meson/mesonbuild/compilers/cs.py", line 22, in <module>
    from .compilers import Compiler, MachineChoice, mono_buildtype_args
ImportError: cannot import name 'MachineChoice' from 'mesonbuild.compilers.compilers' (/__w/meson/meson/mesonbuild/compilers/compilers.py)
```

mesonbuild/ast/interpreter.py
```diff
@@ -14,8 +14,11 @@
 
 # This class contains the basic functionality needed to run any interpreter
 # or an interpreter-based tool.
+from __future__ import annotations
+
+import os, sys
```
dcbaker, line 19: one import per line, please :)

mesonbuild/compilers/compilers.py
dcbaker: _cries too_

dcbaker (approved):
One small nit, otherwise this looks good to me

mesonbuild/ast/interpreter.py
eli-schwartz: :p and here I just moved existing blocks of code around.

Issue #10225, opened 2022-04-01
error when trying to enable vaapi: Tried to get unknown builtin option gstreamer-vaapi:default_library


gdesmott:
**Describe the bug**

I get this error when trying to enable vaapi in [GStreamer](https://gitlab.freedesktop.org/gstreamer/gstreamer):

```
../subprojects/gstreamer-vaapi/gst/vaapi/meson.build:36:0: ERROR: Tried to get unknown builtin option gstreamer-vaapi:default_library
```

**To Reproduce**

- meson build-test
- meson configure build-test -D vaapi=enabled
- ninja -C build-test/

```
(...)
Executing subproject gstreamer-vaapi 

gstreamer-vaapi| Project name: gstreamer-vaapi
gstreamer-vaapi| Project version: 1.21.0.1
gstreamer-vaapi| C compiler for the host machine: ccache cc (gcc 10.3.1 "cc (GCC) 10.3.1 20210422 (Red Hat 10.3.1-1)")
gstreamer-vaapi| C linker for the host machine: cc ld.bfd 2.35-18
gstreamer-vaapi| Compiler for C supports link arguments -Wl,-Bsymbolic-functions: YES (cached)
gstreamer-vaapi| Compiler for C supports arguments -fvisibility=hidden: YES (cached)
gstreamer-vaapi| Compiler for C supports arguments -fno-strict-aliasing: YES (cached)
gstreamer-vaapi| Library m found: YES
gstreamer-vaapi| Dependency gstreamer-1.0 found: YES 1.21.0.1 (overridden)
gstreamer-vaapi| Dependency gstreamer-base-1.0 found: YES 1.21.0.1 (overridden)
gstreamer-vaapi| Dependency gstreamer-pbutils-1.0 found: YES 1.21.0.1 (overridden)
gstreamer-vaapi| Dependency gstreamer-allocators-1.0 found: YES 1.21.0.1 (overridden)
gstreamer-vaapi| Dependency gstreamer-video-1.0 found: YES 1.21.0.1 (overridden)
gstreamer-vaapi| Dependency gstreamer-codecparsers-1.0 found: YES 1.21.0.1 (overridden)
gstreamer-vaapi| Dependency gstreamer-gl-1.0 found: YES 1.21.0.1 (overridden)
gstreamer-vaapi| Message: GStreamer debug system is enabled
gstreamer-vaapi| Dependency libva found: YES 1.9.0 (cached)
[...]

PR #10490, opened 2022-06-14
Fix crash when a reconfigure adds a new subproject

xclaesse:
When a subproject is disabled on the initial configuration we should not
add it into self.coredata.initialized_subprojects because that will
prevent calling self.coredata.init_builtins() on a reconfigure if the
subproject gets enabled.

Fixes: #10225.


xclaesse:
Note that there is another subtle case: `subproject('foo', required: false)` that fails on initial invocation. In that case we already initialized builtin options. On a reconfigure we don't have to initialize builtin options again. That's probably why we were setting that in disabled_subproject(). But it's better to set it in the only place we actually use self.coredata.initialized_subprojects, just after actually initializing builtin options.

eli-schwartz (approved):

PR #10492, opened 2022-06-14
Fix issue where the -j flag would sometimes not be added to certain external projects

stazio:
Sometimes the make command will be `/usr/bin/make` instead of just `make`, which will prevent the multi-threading flag from being added. This PR addresses that by removing the test entirely, as it seems to me that `self.gnu_make()` is sufficient.

Thanks to @xclaesse for finding the issue!

### DIFF
--- a/mesonbuild/scripts/externalproject.py
+++ b/mesonbuild/scripts/externalproject.py
@@ -65,10 +65,7 @@ class ExternalProject:
 
         install_cmd = self.make.copy()
         install_env = {}
-        if is_make:
-            install_cmd.append(f'DESTDIR={self.install_dir}')
-        else:
-            install_env['DESTDIR'] = self.install_dir
+        install_env['DESTDIR'] = self.install_dir
         install_cmd.append('install')
         rc = self._run('install', install_cmd, install_env)
         if rc != 0:
--- a/mesonbuild/scripts/externalproject.py
+++ b/mesonbuild/scripts/externalproject.py
@@ -48,16 +48,15 @@ class ExternalProject:
         with open(self.stampfile, 'w', encoding='utf-8'):
             pass
 
-    def gnu_make(self) -> bool:
+    def supports_jobs_flag(self) -> bool:
         p, o, e = Popen_safe(self.make + ['--version'])
-        if p.returncode == 0 and 'GNU Make' in o:
+        if p.returncode == 0 and ('GNU Make' in o or 'waf' in o):
             return True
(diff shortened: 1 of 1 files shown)


mesonbuild/scripts/externalproject.py
```diff
@@ -55,9 +55,8 @@ def gnu_make(self) -> bool:
         return False
 
     def build(self) -> int:
-        is_make = self.make[0] == 'make'
         make_cmd = self.make.copy()
-        if is_make and self.gnu_make():
```
xclaesse, line 60: is_make is used further down too. Probably should use os.path.basename instead?

eli-schwartz:
> This PR addresses that by removing the test entirely, as it seems to me that `self.gnu_make()` is sufficient.

The External Project module supports:
- `configure && make`
- `waf && waf build`

In the latter case, we could run waf build --version and notice that it doesn't claim to be GNU make, but currently we optimize that out by checking for the literal text "make", and checking that it *contains* make may be more desirable.

(Either that or adding a command-line option to the internal script to specify the "flavor" as either make or waf.)

We do want to handle whatever `state.find_program('make')` can give us, which may be a machine file specifying `gmake` on FreeBSD. Fortunately, `'make' in self.make[0]` still detects that and handily avoids finding `'make' in '/path/to/waf'`.

stazio:
So I added the same checks for `waf()` when testing for multi-threading, since it does support `-j` as well. Does this satisfy the requirement as well? Checking both the basename and gnu_make() seems unnecessary to me...

eli-schwartz:
I tweaked the commit message slightly because the description was a bit long: https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

Thanks for updating it to cover waf too.

mesonbuild/scripts/externalproject.py
```diff
@@ -48,27 +48,23 @@ def write_stampfile(self) -> None:
...
         if rc != 0:
             return rc
 
         install_cmd = self.make.copy()
         install_env = {}
-        if is_make:
-            install_cmd.append(f'DESTDIR={self.install_dir}')
-        else:
-            install_env['DESTDIR'] = self.install_dir
+        install_env['DESTDIR'] = self.install_dir
```
xclaesse, line 67: I'm not so sure about this change, IIRC the reason I made it like that for make is because it does not always work with DESTDIR in env instead of in args.

PR #10493, opened 2022-06-14
Fix finding of Visual studio path in mingw64

TyBalduf:
Ninja backend will fail to find the vs dep dependency
prefix string in a mingw64 environment. This change
simply updates the regex to be able to capture mingw64's unique
file separation pattern.


eli-schwartz:
Please rebase instead of adding merge commits to a PR. :(

TyBalduf:
@eli-schwartz sorry about that. Wasn't thinking the other day and clicked the "merge upstream" button. I have rebased to just include my actual commit.

PR #10494, opened 2022-06-14
devenv: Add support for PowerShell 7 on Windows

seungha-yang:
Checks "pwsh.exe" in addition to "powershell.exe" and "cmd.exe"
to support PowerShell 7 on Windows. The Powershell 7 support was added
in GStreamer (which is the origin of this script) already via
https://gitlab.freedesktop.org/gstreamer/gstreamer/-/merge_requests/2545


eli-schwartz (approved):
Makes sense.

PR #10495, opened 2022-06-15
 migrate declare_dependency to typed_kwargs 

eli-schwartz:
Plus misc preparatory cleanup.

### DIFF
--- a/mesonbuild/interpreter/type_checking.py
+++ b/mesonbuild/interpreter/type_checking.py
@@ -328,7 +328,7 @@ CT_BUILD_ALWAYS_STALE: KwargInfo[T.Optional[bool]] = KwargInfo(
 INSTALL_DIR_KW: KwargInfo[T.Optional[str]] = KwargInfo('install_dir', (str, NoneType))
 
 INCLUDE_DIRECTORIES: KwargInfo[T.List[T.Union[str, IncludeDirs]]] = KwargInfo(
-    'include_dirs',
+    'include_directories',
     ContainerTypeInfo(list, (str, IncludeDirs)),
     listify=True,
     default=[],
--- a/mesonbuild/interpreter/type_checking.py
+++ b/mesonbuild/interpreter/type_checking.py
@@ -9,8 +9,10 @@ import typing as T
 
 from .. import compilers
 from ..build import (EnvironmentVariables, EnvInitValueType, CustomTarget, BuildTarget,
-                     CustomTargetIndex, ExtractedObjects, GeneratedList, IncludeDirs)
+                     CustomTargetIndex, ExtractedObjects, GeneratedList, IncludeDirs,
+                     BothLibraries, SharedLibrary, StaticLibrary, Jar)
 from ..coredata import UserFeatureOption
+from ..dependencies import Dependency, InternalDependency
 from ..interpreterbase.decorators import KwargInfo, ContainerTypeInfo
 from ..mesonlib import File, FileMode, MachineChoice, listify, has_path_sep, OptionKey
 from ..programs import ExternalProgram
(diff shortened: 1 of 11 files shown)


mesonbuild/interpreter/type_checking.py
```diff
@@ -347,3 +380,52 @@ def _output_validator(outputs: T.List[str]) -> T.Optional[str]:
...
 ENV_SEPARATOR_KW = KwargInfo('separator', str, default=os.pathsep)
+
+DEPENDENCIES_KW: KwargInfo[T.List[Dependency]] = KwargInfo(
+    'dependencies',
+    ContainerTypeInfo(list, (Dependency, ExternalLibrary, InternalDependency)),
+    listify=True,
+    default=[],
+)
+
+link_with_error = '''can only be self-built targets, external dependencies (including libraries) must go in "dependencies".'''
```
tristan957, line 391: If this is a global constant, which it looks like it is, and probably only used in this file, preferably this should be `__LINK_WITH_ERROR`.

dcbaker: Just `_LINK_WITH_ERROR` is fine. `__` instructus python to mangle the symbol in a way that prevents collissions in class heirarchies

eli-schwartz: It's a two-functions-local constant, really. :)

Feels weird to name a variable in shouty-caps when its purpose is not to be "read-only, available to the entire file" but rather "the same string is used in two places 8 lines apart and I didn't want to type it twice, but it should go out of brain-scope as soon as it scrolls off the screen".

eli-schwartz:
From a quick check, looks like all test failures are test.json output line matches that are covering an error message with slightly changed wording, will fix that tomorrow.

mesonbuild/interpreter/type_checking.py
```diff
@@ -347,3 +380,59 @@ def _output_validator(outputs: T.List[str]) -> T.Optional[str]:
     validator=in_set_validator({'set', 'prepend', 'append'}))
 
 ENV_SEPARATOR_KW = KwargInfo('separator', str, default=os.pathsep)
+
+DEPENDENCIES_KW: KwargInfo[T.List[Dependency]] = KwargInfo(
+    'dependencies',
+    ContainerTypeInfo(list, (Dependency, ExternalLibrary, InternalDependency)),
```
dcbaker, line 386: `ExternalLibrary` and `InternalDependency` are both `Dependency` subclasses, so it doesn't seem useful to list them here?

mesonbuild/interpreter/type_checking.py
```diff
@@ -347,3 +380,59 @@ def _output_validator(outputs: T.List[str]) -> T.Optional[str]:
...
+    for l in values:
+        if isinstance(l, (CustomTarget, CustomTargetIndex)) and l.links_dynamically():
+            return f'{type(l).__name__} returning a shared library is not allowed'
+        if isinstance(l, Dependency):
+            return _link_with_error
+    return None
+
+LINK_WHOLE_KW: KwargInfo[T.List[T.Union[SharedLibrary, StaticLibrary, CustomTarget, CustomTargetIndex]]] = KwargInfo(
+    'link_whole',
+    ContainerTypeInfo(list, (StaticLibrary, CustomTarget, CustomTargetIndex, Dependency)),
```
dcbaker, line 419: This should allows `BothLibraries` as well, right?

mesonbuild/interpreterbase/decorators.py
```diff
@@ -505,18 +505,24 @@ def check_value_type(types_tuple: T.Tuple[T.Union[T.Type, ContainerTypeInfo], ..
...
+            def emit_feature_change(values: T.Dict[_T, T.Union[str, T.Tuple[str, str]]], feature: T.Union[T.Type['FeatureDeprecated'], T.Type['FeatureNew']]) -> None:
                 for n, version in values.items():
-                    if isinstance(value, (dict, list)):
+                    warn = False
+                    if isinstance(version, tuple):
+                        version, msg = version
+                    else:
+                        msg = None
+
+                    if n in (dict, list):
```
dcbaker, line 516:
```suggestion
                    if n in {dict, list}:
```
always use a set for `in` checks when the types are hashable, it's faster

mesonbuild/interpreterbase/decorators.py
```diff
@@ -505,18 +505,24 @@ def check_value_type(types_tuple: T.Tuple[T.Union[T.Type, ContainerTypeInfo], ..
...
+                        assert(isinstance(n, type)), 'for mypy'
```
dcbaker, line 517:
```suggestion
                        assert isinstance(n, type), 'for mypy'
```
no paranes with assert, please

dcbaker (changes requested):
I noticed a few things, but otherwise this looks fine.

mesonbuild/interpreter/type_checking.py
eli-schwartz: This code predates d413dedf2a6a6f893e110f7a03d9e3468fdf1fc3 and was never modified to sync with the exact types from interpreter.py

On the other hand, KwargInfo will actually use this information when emitting an error message stating the types you are allowed to use. So, I don't know. Maybe it makes sense to list them all? Maybe it doesn't?

mesonbuild/interpreterbase/decorators.py
eli-schwartz: I have absolutely no earthly clue why I did that.

mesonbuild/interpreter/type_checking.py
dcbaker: Yeah, I guess leave them, or at least leave ExternalLibrary which is not an obvious `Dependency` type. But probably add a comment to explain why we're leaving them.

dcbaker (approved):

PR #10498, opened 2022-06-15
Add @BASENAMES@ and @PLAINNAMES@ to custom_target

dcbaker:
There are currently projects in the wild (I don't want to shame them), who are recommending this:
```meson
custom_target(
  ...
  input : ['a.in', 'b.in', 'c.in'],
  output : '.'
  command : [find_program('cmd'), '@OUTPUT@', '@INPUT@'],
)
```
of course, this isn't what they actually want, they just don't want to have to list `['a.out', 'b.out', 'c.out']` in their output lines. So lets make it easy to do the right thing:
```meson
custom_target(
  ...
  input : ['a.in', 'b.in', 'c.in'],
  output : '@PLAINNAMES@.out',
  command : [find_program('cmd'), 'batch', '@OUTDIR@', '@INPUT@'],
)
```

With the plural form passed (it is allowed only to be passed as an exclusve output argument), meson will replace that with the @PLAINNAME@ applied to *each* output, making it simple to handle a custom_target with multiple outputs that follow the @BASENAME@ or @PLAINNAME@ scheme.

still needs tests, but I figure we're going to bikeshed about the naming for a while, so...

### DIFF
--- a/docs/markdown/Generating-sources.md
+++ b/docs/markdown/Generating-sources.md
@@ -106,6 +106,22 @@ executable('myexe', ['main.c', foo_ch[1]], link_with : libfoo)
 In this case `libfoo` depends on both `foo.c` and `foo.h` but `myexe`
 only depends on `foo.h`, the second output.
 
+Since 0.63, if a generate creates multiple files using a `@BASENAME@` or
+`@PLAINNAME@` scheme, the `@BASENAMES@` and `@PLAINNAMES@` value may be used:
+
+```meson
+
+procd = custom_target(
+    'processed sources',
+    output : @BASENAMES@,
+    input : ['foo.c.in', 'bar.c.in']
+    command : [find_program('processor'), '@INPUT@', '@OUTPUT@'],
+)
+
+executable('myexe', ['main.c', procd])
+```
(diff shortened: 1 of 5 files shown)


eli-schwartz:
> of course, this isn't what they actually want, they just don't want to have to list `['a.out', 'b.out', 'c.out']` in their output lines. So lets make it easy to do the right thing:

Actually, these projects just want to place output files in subdirectories. It's not technically required, but the transpiler cmd in question simultaneously documents the meson snippet using `output: '.'` and makes the tool emit the files in directory structure. The example project using the cmd, then makes use of file aliases in the compiler to expose all those files without a directory structure at all.

In theory, the real solution here is `generator(...).process(..., preserve_path_from: meson.current_source_dir())`, in practice generators are totally useless unless you're outputting `*.c *.cpp *.h`.

dcbaker:
I thought more about this, I think instead of adding a new `@BASENAMES@` template, allowing `@BASENAME@` to contain the array of outputs like `@INPUT@` and `@OUTPUT@` do would be more consistant and natural, which would also allow for `@BASENAME1@`, which is probably useful. This shouldn't cause any regressions because currently passing `@BASENAME@` or `@PLAINNAME@` with multiple outputs is an error.

PR #10500, opened 2022-06-16
external-project: Add depends kwarg

xclaesse:
The unit test was racy but surprisingly never failed on CI. The reason
is we need to ensure ninja build somelib.so before running `make` into
the external project.


mesonbuild/modules/unstable_external_project.py
```diff
@@ -55,7 +57,8 @@ def __init__(self,
                  configure_options: T.List[str],
                  cross_configure_options: T.List[str],
                  env: build.EnvironmentVariables,
-                 verbose: bool):
+                 verbose: bool,
+                 extra_depends: T.List[T.Union[BuildTarget, CustomTarget]]):
```
eli-schwartz, line 61: typing-only imports need to either be quoted at the time of use or use future annotations.

eli-schwartz (changes requested):

mesonbuild/modules/unstable_external_project.py
xclaesse: ah right, done.

PR #10501, opened 2022-06-16
docs: only list the latest build of a given stable tag for each wrapdb entry

eli-schwartz:
If we increment the build revision and re-release a wrap, it's a bugfix of an old version, so we can simply not bother to list it in the table. Just list the latest and greatest wrap for each tagged release.

The rest are obviously still available if you look up the relevant tag manually on github, but they aren't very interesting to show here.


xclaesse:
Agreed, LGTM.

PR #10503, opened 2022-06-16
fix: `add_dep` function accessing `dep` attrs before checking if it's a valid dependency

### LINKED ISSUE #10468: AttributeError: 'SubprojectHolder' object has no attribute 'd_features'
**Describe the bug**
Attempting to use `libjpeg-turbo` from `wrapdb` as a dependency for my project. During compilation, I get told that the above-mentioned error should be reported. As a law-abiding developer, I humbly oblige.

**To Reproduce**
```meson
project(
  'meson_wrap_reproduct',
  'c',
  version: '0.1',
  default_options: [
    'c_std=18',
    'warning_level=3',
    'werror=true'
  ],
)

libjpeg = subproject('libjpeg-turbo')

executable(
  'meson_wrap_reproduct',
  'meson_wrap_reproduct.c',
  dependencies: [
    libjpeg,
  ]
)
```

with `libjpeg-turbo` as a subprojects dependency.

**Expected behavior**
Using the dependency from `wrapdb` should work.

**system parameters**
  * Plain
  * Arch Linux @ 5.17.9
  * 3.10.4
* what `meson --version`
  * 0.62.2
  * 1.10.2

This seems to be the case for `libpng` as well, so I'm assuming this is a wider problem.

fxxf:
```python
[...]
            dep_d_features = dep.d_features

            for feature in ('versions', 'import_dirs'):
                if feature in dep_d_features:
                    self.d_features[feature].extend(dep_d_features[feature])
[...]
 ```
Code in this snippet tries to access the `d_features` attribute on `dep` before the if-else chain can ensure that `dep` is valid.

I wasn't quite sure if this has to be executed where it currently is, so I just moved it to the end of the for loop. If that's not the case, this snippet can be made into its own function to avoid an extra call to `isinstance` and get called in the first two branches of the if-else statement.

current-output

Fixes #10468

### DIFF
--- a/test cases/failing/127 subproject object as a dependency/test.json
+++ b/test cases/failing/127 subproject object as a dependency/test.json
@@ -1,0 +1,7 @@
+{
+  "stdout": [
+    {
+      "line": "test cases/failing/127 subproject object as a dependency/meson.build:3:0: ERROR: Argument is of an unacceptable type 'SubprojectHolder'."
+    }
+  ]
+}
--- a/test cases/failing/127 subproject object as a dependency/test.json
+++ b/test cases/failing/127 subproject object as a dependency/test.json
@@ -1,7 +1,7 @@
 {
   "stdout": [
     {
-      "line": "test cases/failing/127 subproject object as a dependency/meson.build:3:0: ERROR: Argument is of an unacceptable type 'SubprojectHolder'."
+      "line": "test cases/failing/127 subproject object as a dependency/meson.build:3:0: ERROR: Tried to use subproject object as a dependency."
     }
   ]
 }
--- a/mesonbuild/build.py
+++ b/mesonbuild/build.py
@@ -1373,6 +1373,9 @@ You probably should put it in link_with instead.''')
                 # This is a bit of a hack. We do not want Build to know anything
                 # about the interpreter so we can't import it and use isinstance.
                 # This should be reliable enough.
+                if hasattr(dep, 'held_object'):
+                    # FIXME: subproject is not a real ObjectHolder so we have to do this by hand
+                    dep = dep.held_object
                 if hasattr(dep, 'project_args_frozen') or hasattr(dep, 'global_args_frozen'):
                     raise InvalidArguments('Tried to use subproject object as a dependency.\n'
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by eli-schwartz** [comment]
This fixes #10468 and should be marked as such, thanks for looking into why it happens.

As it's apparent that we try to specially report some kinds of error here, particularly (as indicated by the mentioned ticket, people trying to use a subproject in the wrong place) "Tried to use subproject object as a dependency", I'd like to ensure that this continues to perform as expected... can you add a `test cases/failing/` project test that sets up a simple subproject and tries to pass it as a dependency to a build target? Then ensure that `test.json` checks for the expected error message on stdout.

> I wasn't quite sure if this has to be executed where it currently is

@rtbo

File: test cases/failing/127 subproject object as a dependency/test.json [hunk 1]
```diff
@@ -0,0 +1,7 @@
+{
+  "stdout": [
+    {
+      "line": "test cases/failing/127 subproject object as a dependency/meson.build:3:0: ERROR: Argument is of an unacceptable type 'SubprojectHolder'."
```
eli-schwartz, line 4: This... should not pass. It's not what I wanted to see, there's a specific check to emit this instead:

> ERROR: Tried to use subproject object as a dependency.
> You probably wanted to use a dependency declared in it instead.
> Access it by calling get_variable() on the subproject object.

This has been broken for quite some time and shows the generic error instead. It's not related to your PR, but actually checking it correctly requires your PR, so... I'm pushing a third commit on top of yours which fixes this properly.

fxxf: Ah, okay. I see what you meant, thank you for the clarification.

eli-schwartz: Yeah, as you can see we should have had a test for this a looooooong time ago.

rtbo:
> I wasn't quite sure if this has to be executed where it currently is

The relevant test for this snippet is `d/13 declare dep`. 
If it still passes we should be OK

eli-schwartz:
It does still pass.

Also the snippet merely reads a `dep` property and extends a dictionary attribute on `self`, so the other dependency processing isn't relevant.

PR #10506, opened 2022-06-17
interpreter: fix a subproject check with symlinks

trhd:
As discussed with @eli-schwartz in #10449  I ran into some problems with subprojects and symlinks earlier. Couldn't get it to work without the attached change. I think the current behavior is unintended but perhaps @eli-schwartz could verify I haven't misunderstood something here.

### DIFF
--- a/mesonbuild/interpreter/interpreter.py
+++ b/mesonbuild/interpreter/interpreter.py
@@ -684,7 +684,7 @@ class Interpreter(InterpreterBase, HoldableObject):
             else:
                 if not self.is_subproject() and srcdir / self.subproject_dir in p.parents:
                     continue
-                if p.is_absolute() and p.is_dir() and srcdir / self.root_subdir in [p] + list(p.resolve().parents):
+                if p.is_absolute() and p.is_dir() and srcdir / self.root_subdir in [p] + list(Path(os.path.abspath(p)).parents):
                     variables[k] = P_OBJ.DependencyVariableString(v)
         for d in deps:
             if not isinstance(d, dependencies.Dependency):
--- a/test cases/unit/106 subproject symlink/meson.build
+++ b/test cases/unit/106 subproject symlink/meson.build
@@ -1,8 +1,15 @@
 project('foo', 'c')
 
-symlinked_subproject = subproject('symlinked_subproject')
+symlinked_subproject = subproject('symlinked_subproject').get_variable('dep')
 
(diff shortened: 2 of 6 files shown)


test cases/unit/106 subproject symlink/meson.build
```diff
@@ -1,8 +1,14 @@
...
 executable('foo',
     sources : 'main.c',
-    dependencies : symlinked_subproject.get_variable('dep')
+    dependencies : symlinked_subproject
+)
+
+custom_target(
+    input : symlinked_subproject.get_variable('datadir') / 'datafile',
+    output : 'data',
+    command : [ find_program(meson.current_source_dir() / 'cat'), '@INPUT@', '@OUTPUT' ]
```
eli-schwartz, line 13: We would typically use `find_program('cp.py')` here

test cases/unit/106 subproject symlink/cat
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env python3
+
+import sys
+
+with open(sys.argv[0]) as in_handle:
+    with open(sys.argv[1]) as out_handle:
+        out_handle.write(in_handle.read())
```
eli-schwartz, line 7: shutil.copyfile?

mesonbuild/interpreter/interpreter.py
```diff
@@ -684,7 +684,7 @@ def func_declare_dependency(self, node, args, kwargs):
             else:
                 if not self.is_subproject() and srcdir / self.subproject_dir in p.parents:
                     continue
-                if p.is_absolute() and p.is_dir() and srcdir / self.root_subdir in [p] + list(p.resolve().parents):
+                if p.is_absolute() and p.is_dir() and srcdir / self.root_subdir in [p] + list(Path(os.path.abspath(p)).parents):
```
eli-schwartz, line 687: Sigh, once again pathlib's concrete Path sucks and os.path is infinitely superior. :)

test cases/unit/106 subproject symlink/meson.build
eli-schwartz: No need for meson.current_source_dir()

eli-schwartz: find_program always searches the current directory first no matter what.

eli-schwartz:
Thanks!

PR #10508, opened 2022-06-17
gnome module: fix regression that broke using built xml files as gresources

### LINKED ISSUE #10367: Regression when passing custom target as input file to gnome.compile_resources
**Describe the bug**
Sometime between Meson 0.59 and 0.61, passing a custom target as the XML file to gnome.compile_resources broke. It was probably sketchy to do that in the first place, so it wouldn't be so bad, except that there no longer seems to be an obvious way to achieve the same thing.
(This bug is kind of the reverse of #5017.)

**To Reproduce**
```meson
project('example', 'c')
gnome = import('gnome')
generated_gresource_xml = custom_target('generated-gresource-xml',
    output: 'my.gresource.xml', capture: true,
    command: ['echo', '<?xml version="1.0" encoding="UTF-8"?><gresources/>'])
resources_generated = gnome.compile_resources('resources-generated',
    generated_gresource_xml)
```
This logs the following error:
> ../meson.build:6:0: ERROR: gnome.compile_resources argument 2 was of type "CustomTarget" but should have been one of: "str", "File"

I've tried to work around it by using the custom target's `full_path()` and adding the custom target to the dependencies keyword arg. (i.e., `gnome.compile_resources('resources-generated', generated_gresource_xml.full_path(), dependencies: generated_gresource_xml)`) This logs the following error:
> ninja: error: '/var/home/ptomato/workspace/scratch/res/_build/my.gresource.xml', needed by 'resources-generated.c', missing and no known rule to make it

So despite it being included in `dependencies`, it's not made first. However, if you ask for it specifically with `ninja my.gresource.xml`, it does get built.

**Expected behavior**
On running the build, `my.gresource.xml` should first be generated, then subsequently used as input for the `compile_resources` call. This worked correctly under 0.59.

**system parameters**
* Is this a [cross build](https://mesonbuild.com/Cross-compilation.html) or just a plain native build (for the same computer)? native build
* what operating system? Fedora Silverblue 35, Meson installed via `pip3 install --user`
* what Python version are you using? 3.10.2
* what `meson --version`? 0.61.0
* what `ninja --version` if it's a Ninja build? 1.10.2

eli-schwartz:
In commit 3dcc7125833cae138987aa4535c88dbd4dbd960d we moved to typed_pos_args. In the process, we deleted some code to specifically raise an error if you use custom_target or generator outputs, instead leaving it out of the typed pos args.

However, that support was specifically supposed to be there. It was onlyan error in part of an if statement for handling old versions of glib-compile-resources. The specific error it calls out is that we need to manually parse the depfile at configure time, due to an external bug; obviously this is impossible if the gresource is only created at build time.

Reinstate the original error message check, and allow built outputs to be used as compile_resources() inputs.

Fixes #10367

### DIFF
--- a/mesonbuild/modules/gnome.py
+++ b/mesonbuild/modules/gnome.py
@@ -371,7 +371,7 @@ class GnomeModule(ExtensionModule):
             rv.append(script)
         return ModuleReturnValue(None, rv)
 
-    @typed_pos_args('gnome.compile_resources', str, (str, mesonlib.File))
+    @typed_pos_args('gnome.compile_resources', str, (str, mesonlib.File, build.CustomTarget, build.CustomTargetIndex, build.GeneratedList))
     @typed_kwargs(
         'gnome.compile_resources',
         _BUILD_BY_DEFAULT,
@@ -426,6 +426,13 @@ class GnomeModule(ExtensionModule):
                     ifile = os.path.join(state.environment.get_build_dir(), input_file.subdir, input_file.fname)
                 else:
                     ifile = os.path.join(input_file.subdir, input_file.fname)
+
+            elif isinstance(input_file, (build.CustomTarget, build.CustomTargetIndex, build.GeneratedList)):
+                raise MesonException('Resource xml files generated at build-time cannot be used with '
+                                     'gnome.compile_resources() in the current version of glib-compile-resources '
+                                     'because we need to scan the xml for dependencies due to '
+                                     '<https://bugzilla.gnome.org/show_bug.cgi?id=774368>\nUse '
+                                     'configure_file() instead to generate it at configure-time.')
             else:
                 ifile = os.path.join(state.subdir, input_file)
 


eli-schwartz:
I'm admittedly not entirely sure when it's desirable to generate a gresource manifest anyway. Inherently, it's a list of files to be bundled; how does Meson know when to regenerate that? Globbing in a python script?

dcbaker:
https://github.com/dcbaker/meson/tree/for-eli has docs changes and unittess for custom_target and custom_target index from my branch. but your actual patch is better than mine.

dcbaker:
I think we need to add the .subdirs from the targets to the `subdirs` list as well. I should patch the test for that...

eli-schwartz:
I got delayed attaching the test case from @dcbaker (it didn't pass because e.g. the file was not specified as ../input, and test.json wasn't updated with the installed files).

Manually pushed as ea952966b41e1ffb72304d317b76848f74fc8d15

PR #10509, opened 2022-06-17
add type annotations to `declare_dependency`

dcbaker:
I realized after I reviewed the previous PR that it didn't have type annotations. Surprise, surprise, mypy found some issues, both with internal annotations, ~and some actual bugs~, nope just annotation issues.

### DIFF
--- a/mesonbuild/dependencies/base.py
+++ b/mesonbuild/dependencies/base.py
@@ -36,9 +36,8 @@ if T.TYPE_CHECKING:
     from ..interpreterbase import FeatureCheckBase
     from ..build import (
         CustomTarget, IncludeDirs, CustomTargetIndex, LibTypes,
-        StaticLibrary, StructuredSources, ExtractedObjects
+        StaticLibrary, StructuredSources, ExtractedObjects, GeneratedTypes
     )
-    from ..mesonlib import FileOrString
 
 
 class DependencyException(MesonException):
@@ -109,7 +108,7 @@ class Dependency(HoldableObject):
         # Raw -L and -l arguments without manual library searching
         # If None, self.link_args will be used
         self.raw_link_args: T.Optional[T.List[str]] = None
-        self.sources: T.List[T.Union['FileOrString', 'CustomTarget', 'StructuredSources']] = []
+        self.sources: T.List[T.Union[mesonlib.File, GeneratedTypes, 'StructuredSources']] = []
         self.extra_files: T.List[mesonlib.File] = []
         self.include_type = self._process_include_type_kw(kwargs)
         self.ext_deps: T.List[Dependency] = []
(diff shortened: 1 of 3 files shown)


mesonbuild/interpreter/interpreter.py
```diff
@@ -674,19 +674,12 @@ def extract_variables(self, kwargs: T.Dict[str, T.Union[T.Dict[str, str], T.List
...
-        sources = self.source_strings_to_files(kwargs['sources'])
-        compile_args = kwargs['compile_args']
-        link_args = kwargs['link_args']
         variables = kwargs['variables']
         version = kwargs['version']
+        sources = self.source_strings_to_files(kwargs['sources'])
         if version is None:
             version = self.project_version
-        d_module_versions = kwargs['d_module_versions']
-        d_import_dirs = self.extract_incdirs(kwargs, 'd_import_dirs')
```
eli-schwartz, line 688: Intermediate variables being useless is kind of a matter of opinion, it can in fact be argued it makes the InternalDependency() easier to read.

However, one thing is definitely outright broken. You stop using extract_incdirs on d_import_dirs, then you add it back in in the next commit when "mypy spots that it is an issue".

mesonbuild/interpreter/interpreter.py
```diff
@@ -674,19 +674,12 @@ def extract_variables(self, kwargs: T.Dict[str, T.Union[T.Dict[str, str], T.List
...
-        deps = kwargs['dependencies']
         incs = self.extract_incdirs(kwargs)
-        libs = kwargs['link_with']
-        libs_whole = kwargs['link_whole']
-        sources = self.source_strings_to_files(kwargs['sources'])
-        compile_args = kwargs['compile_args']
-        link_args = kwargs['link_args']
         variables = kwargs['variables']
         version = kwargs['version']
+        sources = self.source_strings_to_files(kwargs['sources'])
```
eli-schwartz, line 681: Is there a reason this gets deleted from one hunk and added lower down in a different hunk?

mesonbuild/dependencies/base.py
```diff
@@ -246,8 +246,8 @@ def __init__(self, version: str, incdirs: T.List['IncludeDirs'], compile_args: T
         self.include_directories = incdirs
         self.compile_args = compile_args
         self.link_args = link_args
-        self.libraries = libraries
-        self.whole_libraries = whole_libraries
+        self.libraries = list(libraries)
+        self.whole_libraries = list(whole_libraries)
```
eli-schwartz, line 250: If this is already a T.List, why do you need to convert it into one again?

eli-schwartz (changes requested):
This PR leaves me feeling confused.

mesonbuild/interpreter/interpreter.py
eli-schwartz: To add to the confusion, you also misspelled it in the double refactor. `d_import_dirs` -> `d_module_dirs`

mesonbuild/dependencies/base.py
dcbaker: It's not, i've changed it to a Sequence. List is invariant, so a `List[a| b | c | d| e | f]` is *not* fulfilled by `List[a]`. Sequence, however, is co-variant. So `List[a]` does fulfill `Sequence[ a | b | c]`. And this means that every place that wants to pass a list doesn't have to cast or define their list as a n `[a | b | c| d | e | f]`. In this case that's like 8 places in the gnome module.

dcbaker:
I've removed the use of `T.Sequence` and shifted it back to `T.List`, and made the changes elsewhere to make that work.

tristan957 (approved):

xclaesse:
LGTM when CI passed.

PR #10512, opened 2022-06-19
python module: implicitly add python dep to extensions

eli-schwartz:
If there isn't a preexisting dependency on python, append one. It's almost assuredly needed, so just do the right thing out of the box.

/cc @rgommers

See modified unittest. After removing the passed-around dep from `dependencies: ...` it still builds. Quickly tested with SciPy as well -- seems to build, does NOT build if you use `py3_dep.partial_dependency()` as the loophole in the release notes indicates.

As discussed elsewhere, I don't know why one would ever *not* want to have this dep, but... if you really really don't, then the partial_dependency lets you skip out on it. :shrug:

### DIFF
--- a/docs/markdown/Python-module.md
+++ b/docs/markdown/Python-module.md
@@ -99,9 +99,9 @@ Additionally, the following diverge from [[shared_module]]'s default behavior:
   of Python that support this (the python headers define `PyMODINIT_FUNC` has
   default visibility).
 
-`extension_module` does not add any dependencies to the library so
-user may need to add `dependencies : py_installation.dependency()`,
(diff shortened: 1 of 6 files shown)


eli-schwartz:
Made some minor tweaks to the first commit to use the coredata cache with proper identifiers.

rgommers:
Thanks @eli-schwartz, this is useful!

> As discussed elsewhere, I don't know why one would ever _not_ want to have this dep

It's not impossible that there's some very niche use case like a freezing tool that may need this, so the escape hatch seems like a good idea just in case. I've never code that actually did remove the dependency in practice though.

Issue #10211, opened 2022-03-30
Meson 0.6.2 configure error building libadwaita on i686 (regression from 0.61.3)


satmandu:
Error when building libadwaita on linux `i686` (ChromeOS/Chromebrew) **(This is a regression from 0.61.3, which works fine.)**

```
The Meson build system
Version: 0.62.0
Source dir: /usr/local/tmp/crew/libadwaita.20220330014121.dir
Build dir: /usr/local/tmp/crew/libadwaita.20220330014121.dir/builddir
Build type: native build
Project name: libadwaita
Project version: 1.0.3
C compiler for the host machine: cc (gcc 11.2.1 "cc (GCC) 11.2.1 20220305")
C linker for the host machine: cc ld.gold 2.38
Host machine cpu family: x86
Host machine cpu: i686
Compiler for C supports arguments -Wcast-align: YES 
Compiler for C supports arguments -Wdate-time: YES 
Compiler for C supports arguments -Wdeclaration-after-statement: YES 
Compiler for C supports arguments -Werror=format-security -Werror=format=2: YES 
Compiler for C supports arguments -Wendif-labels: YES 
Compiler for C supports arguments -Werror=incompatible-pointer-types: YES 
Compiler for C supports arguments -Werror=missing-declarations: YES 
Compiler for C supports arguments -Werror=overflow: YES 
Compiler for C supports arguments -Werror=return-type: YES 
Compiler for C supports arguments -Werror=shift-count-overflow: YES 
Compiler for C supports arguments -Werror=shift-overflow=2: YES 
Compiler for C supports arguments -Werror=implicit-fallthrough=3: YES 
Compiler for C supports arguments -Wformat-nonliteral: YES 
Compiler for C supports arguments -Wformat-security: YES 
Compiler for C supports arguments -Winit-self: YES 
Compiler for C supports arguments -Wmaybe-uninitialized: YES 
Compiler for C supports arguments -Wmissing-field-initializers: YES 
Compiler for C supports arguments -Wmissing-include-dirs: YES 
Compiler for C supports arguments -Wmissing-noreturn: YES 
Compiler for C supports arguments -Wnested-externs: YES 
Compiler for C supports arguments -Wno-missing-field-initializers: YES 
Compiler for C supports arguments -Wno-sign-compare: YES 
Compiler for C supports arguments -Wno-strict-aliasing: YES 
Compiler for C supports arguments -Wno-unused-parameter: YES 
[...]

PR #10513, opened 2022-06-19
fix: prevent `meson configure` from exiting abnormally when the terminal width (or $COLUMNS) is equal or less than 60

fxxf:
```python
        total_width = shutil.get_terminal_size(fallback=(160, 0))[0]
        _col = max(total_width // 5, 20)
        four_column = (_col, _col, _col, total_width - (3 * _col))
```
In the case where `total_width <=60` is true, `textwrap.wrap(line[3], four_column[3])` raises a `ValueError` which in turn causes meson to error out and exit.

I see that this has been brought up on #10211 and believe that instead of leaving it as-is, falling back to `_col` seems to be the right thing to do.

### DIFF
--- a/mesonbuild/mconf.py
+++ b/mesonbuild/mconf.py
@@ -112,7 +112,8 @@ class Conf:
         """
         total_width = shutil.get_terminal_size(fallback=(160, 0))[0]
         _col = max(total_width // 5, 20)
-        four_column = (_col, _col, _col, total_width - (3 * _col))
+        last_column = total_width - (3 * _col)
+        four_column = (_col, _col, _col, last_column if last_column > 1 else _col)
         # In this case we don't have the choices field, so we can redistribute
         # the extra 40 characters to val and desc
         three_column = (_col, _col * 2, total_width // 2)


eli-schwartz:
I'm not completely sure what the best thing to do here is, because we also want the output to behave sanely.... and with a terminal that narrow it won't be, even if the ValueError is avoided.

fxxf:
While that's fair, with this PR I at least can resize my terminal to get a somewhat saner output shown below.

gif

Another option could be just telling the user that their terminal's width is too small to output something that would make sense, but doing that doesn't sound quite right. Either way, I'd be happy to see this addressed since I'm usually using meson from a relatively small terminal window.

fxxf:
I'm assuming that this PR isn't the wanted way forward. Should I just close this and create an issue so it can be tracked from there?

@eli-schwartz

eli-schwartz:
I'm not going to overthink this. Displaying tabular output on too-small terminals is fraught with risk and may not be possible to do well, users can pay the consequences on their own. At least we can avoid producing a traceback instead of terribly-formatted text, and as you point out, resizing the terminal fixes the previous lines...

PR #10515, opened 2022-06-20
Update Unit-tests.md

CLanguagePurist:
Adding options information for timeout in test cases in meson.

### DIFF
--- a/docs/markdown/Unit-tests.md
+++ b/docs/markdown/Unit-tests.md
@@ -234,8 +234,8 @@ other useful information as the environmental variables. This is
 useful, for example, when you run the tests on Travis-CI, Jenkins and
 the like.
 
-### Timeout
-To disable timeout in test cases, add `timeout: 0` or negative value to allow infinite duration for test case to completes. Alternatively, `timeout_multiplier: 0` accomplish the same functionality as `timeout: 0`. For running test, you can specify command line argument for overriding timeout as well by supplying `--timeout-multiplier 0` argument. The `timeout` option of test case is specified in a number of seconds where 60 seconds equals to 1 minute and go on. The `timeout-multiplier`multiplies the specified timeout duration by a given factors for the timeout-multiplier option.
+**Timeout**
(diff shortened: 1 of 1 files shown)


docs/markdown/Unit-tests.md
```diff
@@ -234,6 +234,9 @@ other useful information as the environmental variables. This is
 useful, for example, when you run the tests on Travis-CI, Jenkins and
 the like.
 
+### Timeout
+To disable timeout in test cases, add `timeout: 0` or negative value to allow infinite duration for test case to completes. Alternatively, `timeout_multiplier: 0` accomplish the same functionality as `timeout: 0`. For running test, you can specify command line argument for overriding timeout as well by supplying `--timeout-multiplier 0` argument. The `timeout` option of test case is specified in a number of seconds where 60 seconds equals to 1 minute and go on. The `timeout-multiplier`multiplies the specified timeout duration by a given factors for the timeout-multiplier option.
```
Ongy, line 238: > [...] and go on.

Should that be "so on" instead? To indicate that it's 60min in 1 hour etc.

> [...] 60 seconds equals to 1 minute [...]

shouldn't it be either "60 seconds equal 1 minute" or "60 seconds are equal to 1 minute"

eli-schwartz, line 238: - should be broken up into lines, since it is markdown a line break doesn't begin a new paragraph
- timeout-multiplier is not a kwarg, only a CLI option.

eli-schwartz (changes requested):

docs/markdown/Unit-tests.md
eli-schwartz: Honestly, this feels like giving people lessons in units of time measurement.

CLanguagePurist: I'll remove it.

CLanguagePurist: Got it, I cleaned up the stub and rearrange the sentences so it make more sense.

CLanguagePurist: Sorry about that, when I made the PR, I haven't slept since. I noticed that you have went out of your way to fix that, thanks.

PR #10520, opened 2022-06-21
Set project(meson_version : ) to default to 0.37

dcbaker:
This is the oldest version we have feature new/deprecation warnings for,
and is also more then 5 years old. By setting a default new feature and
deprecation warnings will be triggered, which will help as we consider
removing deprecated features around the 1.0 release.


dcbaker:
I'm hoping to avoid that, since setting to 0.37 *now* means that people will get warnings immediately of "hey, you're using a feature deprecated in 0.55, and that's getting removed in 1.0"

eli-schwartz:
This seems reasonable to do for a while before 1.0.

Can you squash the first 3 commits? They aren't meaningfully different... if it was useful to split them out depending on which test category they were in, you could just as easily split them out per individual test, or per affected file. But I don't think either of those make sense -- it's one semantic change, and it's category-agnostic.

eli-schwartz:
Hmm, pondering this again. I'm not sure if it makes sense, but...

One alternative to setting the minimum version would be to make `FeatureCheckBase.use()` not skip if the version is `''` but instead have messages such as `Project targets "{tv}'" but uses feature deprecated since ...` show, instead, `Project has no minimum requirements but uses very old feature deprecated since ...`.

The only meaningful difference, probably, is that we wouldn't claim that the project targets such and such a version, if it's actually Meson assuming it should act that way anyway.

dcbaker:
Well, my though was that at some point after 1.0 we would change the assumed base version to 1.x, say when 1.10 comes out that the base changes to 1.0 if unset, since by that point we're assuming that no ones cases about 10 year old versions of meson. But I don't see a ton of difference otherwise 🤷🏻

jpakkane:
Seems reasonable. Needs a rebase, though.

eli-schwartz:
I think that based on my previous review comment I'd rather do this pretty differently UX-wise. Trigger warnings that state "no minimum specified" and broadly warn about using features which aren't universally guaranteed to work. This also would allow warning for any feature which is deprecated by *any* version of meson, as well as any feature which is too new for the current semver slot.

The thing is that this is hardly urgent to get merged, unless we're preparing for 1.0 which didn't end up happening. The purpose of the change is in order to give users fair warning before we remove things. So I'd like to sit down and polish up the messaging a bit and land it in 1.2.0 instead.

PR #10524, opened 2022-06-22
Move python version check as first thing

xclaesse:


xclaesse:
This fix this error when running latest Meson with python 3.5:
```
16:00:26 Traceback (most recent call last):
16:00:26   File "/usr/local/bin/meson", line 7, in <module>
16:00:26     from mesonbuild.mesonmain import main
16:00:26   File "/usr/local/lib/python3.5/dist-packages/mesonbuild/mesonmain.py", line 205
16:00:26     mlog.error(f'Error in {script_name} helper script:')
16:00:26                                                       ^
16:00:26 SyntaxError: invalid syntax
```

meson.py
```diff
@@ -15,6 +15,15 @@
 # limitations under the License.
 
 import sys
+
+# Check python version before importing anything else, we might have an older
+# Python that would error on f-string syntax for example.
+if sys.version_info < (3, 7):
```
eli-schwartz, line 21: Huh, wait a minute.

For some reason when I first looked at this PR, I overlooked that this isn't just moving the check earlier in mesonmain... it is moving it to another file.

The problem is that meson.py is the in-tree loader (and also the zipapp loader) but it's not the loader for building and installing a wheel.

That being said... building and installing a wheel generally means using pip install. pip install (any halfway recent version of it) respects this from setup.cfg: `python_requires = >= 3.7`.

Do we care about giving a proper error message to people that somehow managed to *install* (to site-packages and everything) the wrong (incompatible) version of Meson? Or is this error message only here to handle people that try to run Meson without installing it? ;)

dcbaker: @eli-schwartz i don’t care about people using ancient versions of setuptools, personally

PR #10526, opened 2022-06-22
docs: make the refman linkification slightly prettier for codeblocks

eli-schwartz:
If we link to

```meson
[[#function]]('posarg')
```

then the ideal linkification would operate on "function" in the formatted text:

```
function('posarg')
```

Instead, it operated on "function()" in the formatted text:

```
function()('posarg')
```

Fix this by detecting the next character after the reference, and skipping the usual "automatically append the () for you" stage if it already has that opening parenthesis.


eli-schwartz:
@mensinda it seems to work locally, but this code is yours so I'd really like you to confirm it does what I think it does. :)

mensinda:
Good idea, but I tried it locally and it doesn't seem to work with the `Yaml-RefMan.html`

eli-schwartz:
In the currently deployed version, that page has an example:

source:
```
[[#@str]] [[#executable]]('main', [
    'file_@0@.cpp'.format([[#meson.version]])
])
```

website:

```meson
str executable()('main', [
    'file_@0@.cpp'.format(meson.version())
])
```

This is correct for str, as it doesn't ever get () anyway, and correct for `meson.version()` which has no arguments.

For executable, which does have arguments, we end up with `executable()(...)`.

In my local test, this exact page is in fact fixed. meson.version gets automatic parentheses, since it does not have manually added ones, but executable does not get automatic parentheses (since it has a manual pair). My assumption is that normally we would bother to add manual parentheses only when something needs to go in between, but something like `[[#meson.version]]()` should still detect manual ones, and avoid adding more.

eli-schwartz:
Oh. WAIT.

Haha.

I also fixed this locally but did not commit it:
```diff
diff --git a/docs/markdown/Yaml-RefMan.md b/docs/markdown/Yaml-RefMan.md
index b59e4e6a9..0e59bf51c 100644
--- a/docs/markdown/Yaml-RefMan.md
+++ b/docs/markdown/Yaml-RefMan.md
@@ -48,7 +48,7 @@ Examples:
 Now the same in a code block:
 
 ```meson
-[[#@str]] [[executable]]('main', [
+[[#@str]] [[#executable]]('main', [
     'file_@0@.cpp'.format([[#meson.version]])
 ])
```

PR #10527, opened 2022-06-23
wrap: meson should continue if it is a directory.

skypexu:
Code indent problem causes meson to not apply patches.
This fixes famous problem:
    "Subproject exists but has no meson.build file"

### DIFF
--- a/mesonbuild/wrap/wrap.py
+++ b/mesonbuild/wrap/wrap.py
@@ -363,25 +363,25 @@ class Resolver:
         if os.path.exists(self.dirname):
             if not os.path.isdir(self.dirname):
                 raise WrapException('Path already exists but is not a directory')
-        else:
-            if self.wrap.type == 'file':
-                self.get_file()
             else:
-                self.check_can_download()
-                if self.wrap.type == 'git':
-                    self.get_git()
-                elif self.wrap.type == "hg":
-                    self.get_hg()
-                elif self.wrap.type == "svn":
-                    self.get_svn()
+                if self.wrap.type == 'file':
+                    self.get_file()
                 else:
-                    raise WrapException(f'Unknown wrap type {self.wrap.type!r}')
-            try:
-                self.apply_patch()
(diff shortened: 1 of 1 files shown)


xclaesse:
Actually no, in the case applying patch fails we don't want to `windows_proof_rmtree(self.dirname)` for a directory that already existed (or was a git submodule).

eli-schwartz:
> Code indent problem causes meson to not apply patches.

It's not a code indent problem, it's a design decision. :)

Specifically, we ***must*** perform the download if the subproject path does not exist. Your patch breaks this by removing it from the else clause of `os.path.exists`.

We ***choose*** not download anything in the else clause of `not os.path.isdir`, because we assume it is properly set up. Your patch changes this. to always perform full download/patch overlay, even if it exists already.

Why does a subproject exist without a meson.build file in it? This should not happen.

It used to be possible to have network errors result in successfully downloading and extracting a wrap, but failing to apply the patch overlay, and then the source code got left behind, caused Meson to think it is properly set up, and then fail due to missing meson.build. I fixed this in commit 7deeb293b684e6361fee28e1ec34adce693969b4.

eli-schwartz:
I disagree with the decision to re-download and re-extract a subproject even when it already exists. I also disagree with the decision to re-extract even a patch overlay every time.

If there are cases where meson.build is supposed to exist, but due to a Meson bug it does not, we should fix that bug.

NACK.

eli-schwartz:
> This fixes famous problem:
> "Subproject exists but has no meson.build file"

This PR causes a bunch of CI failures, which say they failed due to "Subproject exists but has no meson.build file"

:(

skypexu:
strange thing is before download, the directory wa already created, but I didn't created it

denizzzka:
> strange thing is before download, the directory was already created, but I didn't created it

It is look like what is described here? https://github.com/mesonbuild/meson/issues/9085

PR #10531, opened 2022-06-24
Fix destdir_join

akihikodaki:
The old implementation assumed a path is of Windows iff the second character is a colon. However, that is not always true. First, a colon can be included in a non-Windows path. For example, it is totally fine to have a directory named `:` on Linux 5.17.0 tmpfs. Second, a Windows path may start with `\\` for UNC or extended length.

Use pathlib to handle all of these cases.

The problem was found during QEMU development:
https://github.com/patchew-project/qemu/commit/a923ba2db1bd396ecf98cc74d7ed6e2398f77fe1#diff-cb6cd6656e23b30f170ce2692ea2f8582de1cb46c92c3b659b34dcbd2ee83b19R10-R18

### DIFF
--- a/mesonbuild/scripts/__init__.py
+++ b/mesonbuild/scripts/__init__.py
@@ -12,10 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from pathlib import PurePath
+
 def destdir_join(d1: str, d2: str) -> str:
     if not d1:
         return d2
     # c:\destdir + c:\prefix must produce c:\destdir\prefix
-    if len(d2) > 1 and d2[1] == ':':
-        return d1 + d2[2:]
-    return d1 + d2
+    return str(PurePath(d1, *PurePath(d2).parts[1:]))


eli-schwartz:
> Second, a Windows path may start with \\ for UNC or extended length.

After destdir_join wouldn't this just be an embedded directory separator, which is correct for this case? It's the same reason on Linux we would just blindly prepend DESTDIR, because we don't need to remove an intervening drive letter.

> For example, it is totally fine to have a directory named : on Linux 5.17.0 tmpfs.

An interesting perspective. I suspect this has never once been done by anyone in the history of meson though.

Checking first for Windows might make sense in that case (and is what pathlib will anyways do underneath umpteen layers of abstraction).

I dislike the idea of using pathlib in a case where it is, unusually, inferior. Pathlib cannot do destdir joining, and using pathlib but then resorting to manual bit banging pays the cost of the pathlib abstraction while not, in the end, making the code any more readable.

akihikodaki:
@eli-schwartz
> After destdir_join wouldn't this just be an embedded directory separator, which is correct for this case? It's the same reason on Linux we would just blindly prepend DESTDIR, because we don't need to remove an intervening drive letter.

Not really. For example, an extended length path starts with `\\?\`, and `?` is not a valid character for a path segment. Also, a user is unlikely to expect it to include e.g. UNC server name in the resulting path where a drive letter is not included.

> Checking first for Windows might make sense in that case (and is what pathlib will anyways do underneath umpteen layers of abstraction).

> I dislike the idea of using pathlib in a case where it is, unusually, inferior. Pathlib cannot do destdir joining, and using pathlib but then resorting to manual bit banging pays the cost of the pathlib abstraction while not, in the end, making the code any more readable.

In this case, it removes the "root" and a following `/`. The concept "root" is described at: https://docs.microsoft.com/en-us/dotnet/standard/io/file-path-formats.
pathlib cannot do destdir joining, but understands the "root" although it confusingly calls Windows "root" as `drive` and the following `/` as `root` (probably for compatibility with POSIX systems). I guess you are referring `parts` access as manual bit banging, but the functionality of `parts` to remove `root` and `drive` is well-documented: https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.parts Python developers just decided it is a good interface for that.

mesonbuild/scripts/__init__.py
```diff
@@ -12,10 +12,14 @@
...
         return d2
+    d2_path = Path(d2)
+    d2_parts = d2_path.parts
     # c:\destdir + c:\prefix must produce c:\destdir\prefix
-    if len(d2) > 1 and d2[1] == ':':
-        return d1 + d2[2:]
-    return d1 + d2
+    if d2_path.drive or d2_path.root:
+        d2_parts = d2_parts[1:]
+    return str(Path(d1, *d2_parts))
```
eli-schwartz, line 25: The API contract of destdir_join is that `d2` must always be an abspath, and if any installation path is not an abspath, it is prefixed with the core option "--prefix" (which itself is a fatal error if it isn't an abspath) before being passed into destdir_join.

So `d2_path.root` cannot ever be false; this check is not needed.

```suggestion
    d2_parts = Path(d2).parts[1:]
    return str(Path(d1, *d2_parts))
```

mesonbuild/scripts/__init__.py
```diff
@@ -12,10 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from pathlib import Path
```
eli-schwartz, line 15: I'd prefer to explicitly use PurePath instead of Path.

The only difference between the two is that a concrete `Path()` exposes a bunch of additional, but broken, I/O methods that aren't even used.

eli-schwartz (commented):
> Not really. For example, an extended length path starts with `\\?\`, and `?` is not a valid character for a path segment. Also, a user is unlikely to expect it to include e.g. UNC server name in the resulting path where a drive letter is not included.

That's... a good point. Although actually I have no idea what users are likely to expect -- I personally would expect that a networked drive share would be the kind of thing someone would stick into DESTDIR, not --prefix, but maybe I'm just clueless about how people use DESTDIR on Windows.

>  pathlib cannot do destdir joining, but understands the "root" although it confusingly calls Windows "root" as `drive` and the following `/` as `root` (probably for compatibility with POSIX systems). I guess you are referring `parts` access as manual bit banging, but the functionality of `parts` to remove `root` and `drive` is well-documented: https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.parts Python developers just decided it is a good interface for that.

Well, tbh this is not the first suggestion to move to pathlib here. Formerly someone (who did not even suggest that there was an edge case to fix) tried this using `d2 = d2[len(Path(d2).anchor):]`. I guess I just feel a bit burned, here.

Using `parts` does not have the same cost to readability. I think I'm a bit more comfortable with that.

akihikodaki:
19048e281109bccf78e3225843e3568086791030 applies suggested changes.



> > Not really. For example, an extended length path starts with `\\?\`, and `?` is not a valid character for a path segment. Also, a user is unlikely to expect it to include e.g. UNC server name in the resulting path where a drive letter is not included.
> 
> That's... a good point. Although actually I have no idea what users are likely to expect -- I personally would expect that a networked drive share would be the kind of thing someone would stick into DESTDIR, not --prefix, but maybe I'm just clueless about how people use DESTDIR on Windows.

--prefix should be used instead of DESTDIR to:
- Install files into different drives. e.g. different drive for `localstatedir`
- Let `meson install` work with particular paths. It is convenient if you repeat installing.

I think these situation would arise equally for a network drive and a drive with a drive letter although a network drive is equally less unlikely to be used.

PR #10534, opened 2022-06-24
Fixed cmake2meson.py multiline strings

Volker-Weissmann:
Without my patch, running `cmake2meson.py` on a CMakeLists.txt that includes
```
set(myvar "abc\
def")
```
will produce a `ValueError: Lexer got confused`.

### DIFF
--- a/tools/cmake2meson.py
+++ b/tools/cmake2meson.py
@@ -38,7 +38,7 @@ class Lexer:
         self.token_specification = [
             # Need to be sorted longest to shortest.
             ('ignore', re.compile(r'[ \t]')),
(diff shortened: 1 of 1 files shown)


dcbaker:
The cmake module has much better code for passing cmake files, and it generates a Meson.build file when you use a cmake subproject. It probably Maude’s more sense to just replace cmake2meson with calls into that module, which is much more sophisticated anyway

Volker-Weissmann:
Your probably right.
I did not know that the cmake module existed and just wasted two days.
*frustrated german noises*

PR #10535, opened 2022-06-24
RFC: meson: Add .get_subprojects() method

killekulla:
Useful to automatically extract variables from all subprojects previously
registered inside the active one, including those of their children.

**TBD:** 
If you think this is usefull, I'am willing to add test case, feature check and documentation.

**Example:**
```
    subs = meson.get_subprojects()
    all_test_deps = []
    foreach name, sub: subs
        all_test_deps += sub.get_variable('test_deps', [])
    endforeach
```

**Behavior:**
```
    meson.build:
        project('example')
        subproject('sub1')
        meson.get_subprojects()
        # returns sub1, sub2.1, sub2.2.1, sub2.2, sub2
        subproject('sub2')
        subprojects/sub1/meson.build
        meson.get_subprojects()
        # returns sub1, sub2.1, sub2.2.1, sub2.2, sub2

    subprojects/sub1/meson.build:
        project('sub1')
        meson.get_subprojects()
        # returns []

    subprojects/sub2/meson.build:
        project('sub2')
        subproject('sub2.1')
        subproject('sub2.2')
        meson.get_subprojects()
        # returns sub2.1, sub2.2.1, sub2.2

    subprojects/sub2.1/meson.build:
        project('sub2.1')

    subprojects/sub2.2/meson.build:
        project('sub2.2')
        subproject('sub2.2.1')
        meson.get_subprojects()
        # returns sub2.2.1

    subprojects/sub2.2.1/meson.build:
        project('sub2.2.1')
```

### DIFF
--- a/mesonbuild/interpreter/interpreter.py
+++ b/mesonbuild/interpreter/interpreter.py
@@ -305,6 +305,7 @@ class Interpreter(InterpreterBase, HoldableObject):
         self.project_args_frozen = False
         self.global_args_frozen = False  # implies self.project_args_frozen
         self.subprojects: T.Dict[str, SubprojectHolder] = {}
+        self.my_subprojects: T.Dict[str, SubprojectHolder] = {}
         self.subproject_stack: T.List[str] = []
         self.configure_file_outputs: T.Dict[str, int] = {}
         # Passed from the outside, only used in subprojects.
@@ -889,6 +890,7 @@ class Interpreter(InterpreterBase, HoldableObject):
                 wanted = kwargs['version']
                 if pv == 'undefined' or not mesonlib.version_compare_many(pv, wanted)[0]:
                     raise InterpreterException(f'Subproject {subp_name} version is {pv} but {wanted} required.')
+            self.my_subprojects[subp_name] = subproject;
             return subproject
 
         r = self.environment.wrap_resolver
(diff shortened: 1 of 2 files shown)


jpakkane:
While this seems somewhat useful functionality it seems a bit like you are using this to implement by hand some higher level functionality and maybe it would be better to provide that directly instead.

Could you explain what the actual issue is that you are solving with this and why you can't do it with meson's existing functionality?

killekulla:
Some Background:

The application we are developing consists of many modules, represented in sub-projects. These modules in turn depend on libraries, usually also subprojects. All modules/libraries contain unit tests.

Executing the unit tests on the Linux development computers with `meson test` works without any problems, the registered test cases (`test(...)`) are automatically collected and executed by meson :+1:.

But for our embedded bare-metal target, for practical reasons, we need to link all unit tests (or at least as many as possible) into one single executable.

One possibility would be to iterate through all sub-projects and automatically collect special test_deps of the sub-projects to use them accordingly in the top-level meson file. Collecting them manually would be error-prone, since the sub-subprojects would not necessarily be visible in the top-level meson.build.

I don't know of any interpreter feature that lets me automatically iterate through registered subprojects. `meson subprojects foreach` is the wrong tool for that.

Any ideas how this could be done without `.get_subprojects()`?

jpakkane:
> But for our embedded bare-metal target, for practical reasons, we need to link all unit tests (or at least as many as possible) into one single executable.

One way of doing that is to replicate exactly what dependencies do. That is, each project is responsible for reporting its own things as well as its dependencies. So something like:

```
test_things = [<this projects's things>, sp1.get_variable('test_things'), sp2.get_variable('test_things')]
```

PR #10538, opened 2022-06-26
chore: Included githubactions in the dependabot config

naveensrinivasan:
This should help with keeping the GitHub actions updated on new releases. This will also help with keeping it secure.

Dependabot helps in keeping the supply chain secure https://docs.github.com/en/code-security/dependabot

GitHub actions up to date https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

https://github.com/ossf/scorecard/blob/main/docs/checks.md#dependency-update-tool
Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"


eli-schwartz:
Maybe when dependabot stops spamming forks and refusing to group updates together, etc. Personally, I find it all rather gross.

Supposedly renovate solves both of these problems. The OpenSSF scorecard even seems to subtly suggest a preference for renovate.

Also nice: renovate is open source and reduces vendor lockin. This is both nice, in the sense that OSS projects should support OSS solutions, and also probably the reason that they actually manage to improve the state of affairs vs. dependabot.

> This will also help with keeping it secure.

Well, really there are much better ways. You keep things secure by not running actions as trusted, that way the only thing they can do is burn CI minutes (which is free via OSS sponsorship).

We have 10 workflow files, 8 of which run on every PR and number some 30 workflow runs. These *never* need to do anything trusted, even when merged to master.

The remaining 2 are for:
- checking documentation builds on updates to `doc/`, and automatically pushing to gh-pages when merged to master
- checking CI image builds (6 jobs) on updates to `ci/ciimage/`, and automatically pushing to Docker Hub when merged to master

The former needs a configured secret for docker, the latter needs a configured secret for cross-repo pushes. Neither use the default GITHUB_TOKEN. (I think this can be configured in the org dashboard to be read-only by default).

eli-schwartz:
> Neither use the default GITHUB_TOKEN. (I think this can be configured in the org dashboard to be read-only by default).

Ah, and regardless of the org setting, I see someone already submitted #10537 to set it in the workflow files directly.

naveensrinivasan:
I will close this. Thanks

PR #10540, opened 2022-06-26
Fix the wrong detection of Win32 platform in MSVC backend

### LINKED ISSUE #10539: Unsupported Visual Studio target machine: Win32
**Describe the bug**
meson throw exception `Unsupported Visual Studio target machine: Win32`

**To Reproduce**
This problem was happened on my build work of glib. There are the steps to reproduce this issue.

* Download glib with tag [v2.72.1](https://gitlab.gnome.org/GNOME/glib/-/tags/2.72.1).  
* Open **x86 Native** Tools Command Prompt for VS 2019. Btw, my machine is on x64 platform.  
* Use following command to execute meson:  
```
cd /d root_of_glib
mkdir _buildprj
cd _buildprj
meson .. --buildtype=release --prefix=D:/CppLib/GLib/Win32 --backend=vs2019
```

The build of glib for x64 platform was perfect. But when I was trying building this for Win32 platform. It threw this exception.

**Expected behavior**
meson should generate correct Visual Studio project files, rather than throwing exception.

I have done some work with my view. I traced this exception and I noticed this line: https://github.com/mesonbuild/meson/blob/5f8b0487abf501a09c51387791b578a148c973e2/mesonbuild/backend/vs2010backend.py#L1315  
It was pretty weird according to my understanding, so I changed it to `targetplatform = platform.lower()`. Then, everything work as my expect. I don't know whether this modification is suit for this project, so I open an issue rather than opening a PR.

**System parameters**
* Maybe this is a cross build. Because I try building an x86 project in a x64 computer.  
* OS: Windows 10 (21H1)
* Visual Studio version: Visual Studio Community 2019 (16.11.3)
* Python version : 3.9.0
* Meson version: 0.62.2
* Machine Platform: x64

yyc12345:
Add a missing `lower()` for MSVCbackend platform detection. This PR will fix #10539.  

I am not familiar with meson. So I hope someone who knows MSVC backend well can check my code. Hope my changes will not affect the build of any other architectures.

### DIFF
--- a/mesonbuild/backend/vs2010backend.py
+++ b/mesonbuild/backend/vs2010backend.py
@@ -1312,7 +1312,7 @@ class Vs2010Backend(backends.Backend):
             pdb.text = f'$(OutDir){target_name}.pdb'
         targetmachine = ET.SubElement(link, 'TargetMachine')
         if target.for_machine is MachineChoice.BUILD:
-            targetplatform = platform
+            targetplatform = platform.lower()
         else:
             targetplatform = self.platform.lower()
         if targetplatform == 'win32':


eli-schwartz:
The code looks reasonable, though I tweaked the commit message a teeny bit.

PR #10542, opened 2022-06-27
dependencies: update llvm versions

dcbaker:
15 is the current snapshot version, 13 and 14 are released.


dcbaker:
@nirbheek, I think this should be backported to any new stable releases

PR #10544, opened 2022-06-28
pkgconfig: do not add valgrind as requirement

elmarco:
For the same reasons commit 7aa28456d ("Add dependency type for
Valgrind") removed linking with valgrind, pkgconfig shouldn't generate
"Requirements" for it, in general.

This solves dbus meson port question/issue from:
https://gitlab.freedesktop.org/dbus/dbus/-/merge_requests/303#note_1444819


dcbaker:
This would also fix an issue we have with libdrm

dcbaker (approved):

PR #10545, opened 2022-06-28
Cmake Module: Nicer error messages

Volker-Weissmann:
I made some error messages nicer. Hopefully the next one debugging will appreciate this.

I can also tell you why the error message sometimes gets triggered:

In `_cmake_add_executable`, we find the lines

```python
if 'IMPORTED' not in args:
    return self._gen_exception('add_executable', 'non imported executables are not supported', tline)
```

I am going to (try to) add support for non imported executables.


mesonbuild/cmake/interpreter.py
```diff
@@ -664,6 +664,7 @@ def ensure_absolute(x: Path) -> Path:
         commands = []  # type: T.List[T.List[T.Union[str, ConverterTarget]]]
         for curr_cmd in self._raw_target.command:
             assert isinstance(curr_cmd, list)
+            assert curr_cmd[0] != '', "An empty string is not a valid executable"
```
eli-schwartz, line 667: Is this supposed to:
- catch programming errors in the cmake module
- tell mypy something about type checking that it cannot infer automatically
- emit user errors for a cmake project that for whatever reason cannot be parsed

eli-schwartz (changes requested):
This needs to be rebased instead of having merge commits. Also please `git commit --amend` if the original commit accidentally dropped some code, rather than re-adding it in a follow-up commit.

mesonbuild/cmake/generator.py
```diff
@@ -66,6 +67,7 @@ def target_property(arg: str) -> str:
 
     def target_file(arg: str) -> str:
         if arg not in trace.targets:
+            mlog.warning(f"Somewhere in your CMakeLists.txt you have '$<TARGET_FILE:{arg}>'. In cmake, this evaluates to the path to '{arg}'. If '{arg}' does not exist, cmake errors out. We think that '{arg}' does not exist, so we do not know its path and just return an empty string.")
```
tristan957, line 70: s/to the path to/to the path of/

tristan957:
I don't really love the new messages. They are too conversational in my opinion. Make them succinct, but descriptive.

Volker-Weissmann:
> I don't really love the new messages. They are too conversational in my opinion. Make them succinct, but descriptive.

I applied your suggestions.

mesonbuild/cmake/traceparser.py
```diff
@@ -231,7 +231,13 @@ def parse(self, trace: T.Optional[str] = None) -> None:
 
         for ctgt in self.custom_targets:
             ctgt.outputs = pathlist_gen(ctgt._outputs_str)
+            temp = ctgt.command
             ctgt.command = [strlist_gen(x) for x in ctgt.command]
+            for command, src in zip(ctgt.command, temp):
+                if command[0] == "":
+                    raise CMakeException(
+                        "We evaluated the cmake variable '{}' to an empty string, which is not a valid path to an executable.".format(src[0])
```
eli-schwartz, line 239: Maybe:

"The cmake variable '{}' is an empty string, which is not a valid path to an executable."

Volker-Weissmann: I think its important to stress here that meson made the variable evaluation, not cmake, because it is likely, that a variable is not an empty string if it is evaluated by cmake. i.e. correctly, but it is an empty string it is evaluated by meson i.e. incorrectly.

Volker-Weissmann:
Do I need to request a new review or something?

eli-schwartz:
It's already assigned, but the cmake module maintainer isn't always available.

dcbaker (approved):

mensinda (approved):
LGTM

PR #10550, opened 2022-06-30
gnome module: Use --quiet to supress printing the link command line

hughsie:
This removes one line of stderr output per GObject Introspection file
processed, e.g.

g-ir-scanner: link: gcc -o Fwupd-2.0 Fwupd-2.0.o -L. -Wl,-rpath...

### DIFF
--- a/mesonbuild/modules/gnome.py
+++ b/mesonbuild/modules/gnome.py
@@ -1149,6 +1149,7 @@ class GnomeModule(ExtensionModule):
         gir_inc_dirs: T.List[str] = []
 
         scan_command: T.List[T.Union[str, build.Executable, 'ExternalProgram', 'OverrideProgram']] = [giscanner]
+        scan_command += ['--quiet']
         scan_command += ['--no-libtool']
         scan_command += ['--namespace=' + ns, '--nsversion=' + nsversion]
         scan_command += ['--warn-all']


dcbaker (approved):

tristan957:
Could this not just go in 0.63? Not like it adds a feature or anything.

eli-schwartz (changes requested):
I'm a bit skeptical of this change in general, because "quiet" is not the same as "don't print status messages" and there are indications in the codebase that it's used for more than that. I would like to double-check what all this does.

As far as the timing of 0.63.0 / 0.64.0, this also isn't a bugfix (the fact that the scanner is noisy doesn't result in any negative build effects) so I'm not particularly concerned about making it into the pending release when we are in the middle of rc2. In general, during an rc, we want to minimize the number of changes, minimize having to publish yet another rc, and as much as possible focus on catching regressions.

Passing the "quiet" argument to a tool doesn't feel super urgent.

dcbaker:
That was my logic as well, it's not a bug fix, it's a quality of life improvement. That can wait for 0.64

hughsie:
> That can wait for 0.64

Ack, no time pressure on my side. The console output has been there since 2016, so two weeks more isn't going to make any difference :)

dcbaker:
@eli-schwartz I went and read through the g-ir-scanner code, and it looks like --quiet is safe, and just turns off printing: https://gitlab.gnome.org/search?search=.quiet&nav_source=navbar&project_id=659&group_id=8&search_code=true&repository_ref=main

eli-schwartz:
Okay, so, it gets used in a couple cases:
- it determines whether to `print()` the above-mentioned log line directly in g-ir-scanner
- it determines whether to pass `--silent` to any invocation of libtool
- in the case of `elif warning_count > 0 and options.warn_all is False and options.quiet is False:` it will print directly in g-ir-scanner the message: "g-ir-scanner: %s: warning: %d warnings suppressed (use --warn-all to see them)"


Fortunately, Meson does pass --warn-all, so no warnings are suppressed. And I think it would be fair enough to stamp on any messages libtool emits, but we also pass --no-libtool. ;)

So in the current state of g-ir-scanner, 2 of the 3 effects of this flag are in code we disable, and the 3rd effect of this flag is what this PR specifically wants to turn off (and it is reasonable to do so). I think it's probably safe to assume g-ir-scanner won't add more prints that we actually do want. (They will hopefully refactor to allow Meson to directly take charge of compiling/linking first...)

Let's ship it.

PR #10552, opened 2022-06-30
interpreter: fix both libraries with `pic : false` and generators

dcbaker:
If we have a call to `both_libraries()` with `pic : false` (or are using
the xcode backend), then we are forced to compile all of the sources
twice, once for the static library and once for the shared library. If
there is a generator in the sources, then we will create two rules that
produce the same output, and configuration will fail.

To avoid that, when this happens, rename the static library so that the
static and shared libraries will get separate folder in the build tree,
and no collision will happen.


dcbaker:
@jpakkane, @nirbheek, here's a fun bug fix that we should probably backport

eli-schwartz:
This looks suspiciously like it results in the wrong publicly installed name?

eli-schwartz:
Also without looking at the implementation, I thought that these generated sources *also* would just be generated into the libfoo.a.p/ and libfoo.so.p/ corresponding private directories? So why do they collide?

dcbaker:
Adn now I simply cannot get it to reproduce. I have no idea what's going on. I'll close it for now and if I see it again re-open

test cases/common/178 bothlibraries/gen.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env python3
+
+import argparse
+import shutil
+
+
+def main() -> None:
```
eli-schwartz, line 7: Ooor we could just not bother type-annotating the least important part of an extremely trivial copy.py script. Which we should just be copying from one of the other 19 cp.py / copyfile.py implementations in `test cases/`, which generally do not go to the overhead of importing argparse for a non-user-facing script...

PR #10554, opened 2022-07-01
Add warning_level=everything

drobilla:
Adds a new highest warning level that is roughly equivalent to "all warnings".
This adds a way to use `/Wall` with MSVC (without the previous broken warning),
`-Weverything` with clang, and almost all general warnings in GCC with
strictness roughly equivalent to clang's `-Weverything`.

For objectivity, all warnings are included except those that require specific
values or are related to language level compatibility.  This warning level is
mainly intended for new code, and it is expected that projects will need to add
some suppressions.

Since the warnings enabled at this level are inherently unstable with respect
to compiler versions, it is intended for use by developers and not to be set as
the default.

### DIFF
--- a/mesonbuild/compilers/mixins/gnu.py
+++ b/mesonbuild/compilers/mixins/gnu.py
@@ -92,6 +92,238 @@ gnu_color_args = {
     'never': ['-fdiagnostics-color=never'],
 }  # type: T.Dict[str, T.List[str]]
 
+# Warnings collected from the GCC source and documentation.  This is an
+# objective set of all the warnings flags that apply to general projects: the
+# only ones omitted are those that require a project-specific value, or are
+# related to non-standard or legacy language support.  This behaves roughly
+# like -Weverything in clang.  Warnings implied by -Wall, -Wextra, or
+# higher-level warnings already enabled here are not included in these lists to
+# keep them as short as possible.  History goes back to GCC 3.0.0, everything
+# earlier is considered historical and listed under version 0.0.0.
+
+# GCC warnings for all C-family languages
+# Omitted non-general warnings:
+#   -Wabi=
+#   -Waggregate-return
+#   -Walloc-size-larger-than=BYTES
+#   -Walloca-larger-than=BYTES
+#   -Wframe-larger-than=BYTES
+#   -Wlarger-than=BYTES
+#   -Wstack-usage=BYTES
+#   -Wsystem-headers
+#   -Wtrampolines
+#   -Wvla-larger-than=BYTES
+#
+# Omitted warnings enabled elsewhere in meson:
+#   -Winvalid-pch (GCC 3.4.0)
+gnu_common_warning_args = {
+    "0.0.0": [
+        "-Wcast-qual",
+        "-Wconversion",
+        "-Wfloat-equal",
+        "-Wformat=2",
+        "-Winline",
+        "-Wmissing-declarations",
+        "-Wredundant-decls",
+        "-Wshadow",
+        "-Wundef",
+        "-Wuninitialized",
+        "-Wwrite-strings",
+    ],
+    "3.0.0": [
+        "-Wdisabled-optimization",
+        "-Wpacked",
+        "-Wpadded",
+    ],
+    "3.3.0": [
+        "-Wmultichar",
+        "-Wswitch-default",
+        "-Wswitch-enum",
+        "-Wunused-macros",
+    ],
+    "4.0.0": [
+        "-Wmissing-include-dirs",
+    ],
+    "4.1.0": [
+        "-Wunsafe-loop-optimizations",
+        "-Wstack-protector",
+    ],
+    "4.2.0": [
+        "-Wstrict-overflow=5",
+    ],
+    "4.3.0": [
+        "-Warray-bounds=2",
+        "-Wlogical-op",
+        "-Wstrict-aliasing=3",
+        "-Wvla",
+    ],
+    "4.6.0": [
+        "-Wdouble-promotion",
+        "-Wsuggest-attribute=const",
+        "-Wsuggest-attribute=noreturn",
+        "-Wsuggest-attribute=pure",
+        "-Wtrampolines",
+    ],
+    "4.7.0": [
+        "-Wvector-operation-performance",
+    ],
+    "4.8.0": [
+        "-Wsuggest-attribute=format",
+    ],
+    "4.9.0": [
+        "-Wdate-time",
+    ],
+    "5.1.0": [
+        "-Wformat-signedness",
+        "-Wnormalized=nfc",
+    ],
+    "6.1.0": [
+        "-Wduplicated-cond",
+        "-Wnull-dereference",
+        "-Wshift-negative-value",
+        "-Wshift-overflow=2",
+        "-Wunused-const-variable=2",
+    ],
+    "7.1.0": [
+        "-Walloca",
+        "-Walloc-zero",
+        "-Wformat-overflow=2",
+        "-Wformat-truncation=2",
+        "-Wstringop-overflow=3",
+    ],
+    "7.2.0": [
+        "-Wduplicated-branches",
+    ],
+    "8.1.0": [
+        "-Wattribute-alias=2",
+        "-Wcast-align=strict",
+        "-Wsuggest-attribute=cold",
+        "-Wsuggest-attribute=malloc",
+    ],
+    "10.1.0": [
+        "-Wanalyzer-too-complex",
+        "-Warith-conversion",
+    ],
+    "12.1.0": [
+        "-Wbidi-chars=ucn",
+        "-Wopenacc-parallelism",
+        "-Wtrivial-auto-var-init",
+    ],
+}  # type: T.Dict[str, T.List[str]]
+
+# GCC warnings for C
+# Omitted non-general or legacy warnings:
+#   -Wc11-c2x-compat
+#   -Wc90-c99-compat
+#   -Wc99-c11-compat
+#   -Wdeclaration-after-statement
+#   -Wtraditional
+#   -Wtraditional-conversion
+gnu_c_warning_args = {
+    "0.0.0": [
+        "-Wbad-function-cast",
+        "-Wmissing-prototypes",
+        "-Wnested-externs",
+        "-Wstrict-prototypes",
+    ],
+    "3.4.0": [
+        "-Wold-style-definition",
+        "-Winit-self",
+    ],
+    "4.1.0": [
+        "-Wc++-compat",
+    ],
+    "4.5.0": [
+        "-Wunsuffixed-float-constants",
+    ],
+}  # type: T.Dict[str, T.List[str]]
+
+# GCC warnings for C++
+# Omitted non-general or legacy warnings:
+#   -Wc++0x-compat
+#   -Wc++1z-compat
+#   -Wc++2a-compat
+#   -Wctad-maybe-unsupported
+#   -Wnamespaces
+#   -Wtemplates
+gnu_cpp_warning_args = {
+    "0.0.0": [
+        "-Wctor-dtor-privacy",
+        "-Weffc++",
+        "-Wnon-virtual-dtor",
+        "-Wold-style-cast",
+        "-Woverloaded-virtual",
+        "-Wsign-promo",
+    ],
+    "4.0.1": [
+        "-Wstrict-null-sentinel",
+    ],
+    "4.6.0": [
+        "-Wnoexcept",
+    ],
+    "4.7.0": [
+        "-Wzero-as-null-pointer-constant",
+    ],
+    "4.8.0": [
+        "-Wabi-tag",
+        "-Wuseless-cast",
+    ],
+    "4.9.0": [
+        "-Wconditionally-supported",
+    ],
+    "5.1.0": [
+        "-Wsuggest-final-methods",
+        "-Wsuggest-final-types",
+        "-Wsuggest-override",
+    ],
+    "6.1.0": [
+        "-Wmultiple-inheritance",
+        "-Wplacement-new=2",
+        "-Wvirtual-inheritance",
+    ],
+    "7.1.0": [
+        "-Waligned-new=all",
+        "-Wnoexcept-type",
+        "-Wregister",
+    ],
+    "8.1.0": [
+        "-Wcatch-value=3",
+        "-Wextra-semi",
+    ],
+    "9.1.0": [
+        "-Wdeprecated-copy-dtor",
+        "-Wredundant-move",
+    ],
+    "10.1.0": [
+        "-Wcomma-subscript",
+        "-Wmismatched-tags",
+        "-Wredundant-tags",
+        "-Wvolatile",
+    ],
+    "11.1.0": [
+        "-Wdeprecated-enum-enum-conversion",
+        "-Wdeprecated-enum-float-conversion",
+        "-Winvalid-imported-macros",
+    ],
+}  # type: T.Dict[str, T.List[str]]
+
+# GCC warnings for Objective C and Objective C++
+# Omitted non-general or legacy warnings:
+#   -Wtraditional
+#   -Wtraditional-conversion
+gnu_objc_warning_args = {
+    "0.0.0": [
+        "-Wselector",
+    ],
+    "3.3": [
+        "-Wundeclared-selector",
+    ],
+    "4.1.0": [
+        "-Wassign-intercept",
+        "-Wstrict-selector-match",
+    ],
+}  # type: T.Dict[str, T.List[str]]
+
 
 @functools.lru_cache(maxsize=None)
 def gnulike_default_include_dirs(compiler: T.Tuple[str, ...], lang: str) -> 'ImmutableListProtocol[str]':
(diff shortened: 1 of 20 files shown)


mesonbuild/compilers/mixins/gnu.py
```diff
@@ -91,6 +91,169 @@
...
+# Walloca-larger-than=BYTES
+# Wframe-larger-than=BYTES
+# Wlarger-than=BYTES
+# Wstack-usage=BYTES
+# Wsystem-headers
+# Wtraditional
+# Wtraditional-conversion
+# Wtrampolines
+# Wvla-larger-than=BYTES
+gnu_common_warning_args = [
```
tristan957, line 108: I live `-Weverything`-esque changes. Don't like this. Seems unmaintainable. Would rather compiler developers include these with their `-Weverything` equivalent.

drobilla: Fair enough, but in GCC's case, they don't, and won't.  In reality, the problem remains.

I don't see a maintainability issue for meson - if future versions add warnings, meson _can_ add them, or it can't, and that's fine.  The point is that it's globally much more maintainable.  Doing this (via mass copy/paste) in countless projects is a nightmare - I personally have _23_ copies of this at the moment, and from a user POV, this very much smells like a thing my build system should be able to do for me.

eli-schwartz: > Don't like this. Seems unmaintainable. Would rather compiler developers include these with their `-Weverything` equivalent.

So instead of implementing something useful now, we should ask GCC to implement it instead even though they've been steadfastly not caring for quite a while now?

On a meta level, I approve this feature using the current design for 0.64.0, hardcoding compiler flags is perfectly fine. We might discuss using something other than "4", as per discussion on IRC, but that's a separate topic.

tristan957: Multiple C compilers exist for a reason. clang has access to this option while gcc does not. Use both compilers for your CI environment.

All Meson is going to do is field issues and PRs to add warnings to these arrays. Seems like pointless noise to me.

If gcc doesn't have an equivalent, `4` should just be the equivalent of `3`. We aren't making up equivalents for compilers who don't have a `3` from what I can tell in the context of this PR. Doesn't make sense to start doing it now.

drobilla: > All Meson is going to do is field issues and PRs to add warnings to these arrays. Seems like pointless noise to me.

I made the set as objective as possible for this reason.  The only reason to add warnings here is on new major releases of GCC, which are infrequent, easy to track (new warnings are in the release notes), and don't often add warnings that aren't a part of `-Wall` or `-Wextra` anyway.  On top of that, it's not necessary for meson to update them anyway - with GCC this is _not_ like `-Weverything`, the flags are stable.

> If gcc doesn't have an equivalent, 4 should just be the equivalent of 3.

3 doesn't do what I need.

> We aren't making up equivalents for compilers who don't have a 3 from what I can tell in the context of this PR. Doesn't make sense to start doing it now.

GCC doesn't have numeric warning levels to begin with, i.e., _all_ the warning levels are already made up.

drobilla:
Regarding the option value:

It's come up on IRC that there may be a future desire for more warning levels, in which case this would be in the way at level 4.  it's also true that this isn't _every_ warning flag in GCC (because some are language-level-specific, and some require values although those don't really count here).  I don't want to discuss additional subjective warning levels here, but think leaving space for them as a possibility is a good idea.

Some ideas:

 * Stick with numbers but go all the way to 9 ("surely 9 warning levels ought to be enough for everybody"): `warning_level=9`
 
 * Use symbolic values, inventing enumerators to avoid confusing with well-known compiler options:
   * `warning_level=most` - The current patch ("almost every warning that applies to general projects")
   * `warning_level=max` - Literally every warning it is possible for meson to enable.

Using `warning_level=everything` instead of `max` might be fine because that's unambiguous.  I think `all` and `extra` should be avoided here, though.

mesonbuild/compilers/mixins/gnu.py
tristan957: I think it is pretty well understood in the compiler world that GCC levels of warnings go from "-Wall" to "-Wextra" to "-pedantic".

drobilla: `-Wpedantic` is more or less orthogonal to the others.

Clang and MSVC have more levels, so it already isn't true that meson warning levels exactly reflect those in compilers.  It wouldn't make sense to add this level for them, but have it do nothing for GCC because <something users don't care about>.

In any case, the whole point of this PR is to add an additional warning level.  If your argument is essentially "you don't want this", well, I promise that I do, and others do as well, and that we know what we're doing and have our reasons.

Concretely, my motivations for this are rooted in real world problems.  At the moment, I have _4767_ lines of mostly duplicated meson code that this would allow me to delete.  Meson supporting this would allow projects to use the same "all warnings then explicit suppressions" pattern with all compilers.  This seems fully in-line with the point of a build system to me: papering over differences between platforms and compilers and allowing the same patterns to work with all of them.

drobilla:
The only tricky part here to me is the warnings related to language levels.  To get roughly equivalent behaviour to clang, the warnings would need to be based on the current language level (which I don't think is possible?).

This leaves a few options:

  * Leave them out entirely (current proposal)
    * Inconsistency: projects would need to add things like `-Wno-c++98-compat` for clang
    * or, maintenance burden: meson would need to add new suppressions for clang as new language versions come out
    * Unfortunate: these warnings aren't unreasonable in general
  
  * Leave them all enabled
    * Inconsistency: GCC would warn for future languages (those warnings don't care about the language level), clang would not
    * Maintenance burden: new ones would need to be added for GCC as new language versions come out

Since this isn't possible to do perfectly without knowing the language level, I'm okay with the current solution of just leaving them out for GCC, although the inconsistency smells a bit.  The suppressions could be maintained for clang, but given the maintenance concerns, I think I prefer just keeping all changing and/or subjective things out of this (and meson) entirely.

(In practice, I have been suppressing `-Wno-c++98-compat` and `-Wno-c++98-compat-pedantic` everywhere with clang, although this is subjective and other projects might not care about anything pre-c++-17 or whatever so I don't think it would be appropriate to bake this into meson)

mesonbuild/compilers/mixins/gnu.py
tristan957: I never said I don't want this. I said maintaining this huge list of flags is a burden. clang has `-Weverything` which corresponds well to `warning_level=4`. gcc doesn't. I am not convinced that maintaining these lists is Meson's burden to bare, but I am also not the person that needs to be convinced.

drobilla:
ObjC: d8979c647

eli-schwartz:
> The only tricky part here to me is the warnings related to language levels. To get roughly equivalent behaviour to clang, the warnings would need to be based on the current language level (which I don't think is possible?).

Without looking at implementation details:
- you definitely cannot set it up as class-level attributes
- you may be able to handle it in `__init__()`? I think at the latest, some time during `__init__()` any compiler will define and load `<lang>_*` args, so you'd just have to do it later than that I guess.
- if nothing else I don't see why you cannot handle this late during `get_warn_args`.

drobilla:
It might be possible in `__init__()`, I just didn't see a glaringly obvious case of doing it that I could steal (I'm not familiar with meson internals).  `get_warn_args` is a good pointer, thanks.

I might look into it, but after looking over the flags and compiler behaviours, I'm not sure what the end goal even would be.  The language-level warnings are really complicated and work differently between clang and gcc.  For C, trying for "don't warn about older languages" seems something like:

  * clang: `-Weverything` plus:
    * `-Wno-c99-compat -Wno-c99-extensions`, for `c99`
    * `-Wno-c11-extensions`, for `c11` and newer
    * `-Wno-c14-extensions`, for `c14` and newer
    * etc
  * gcc: all flags plus:
    * `-Wc90-c99-compat` for `c90`
    * `-Wc99-c11-compat` for `c99`
    * etc

C++ is different still, and even more irregular and complicated.  I have no idea what analogues there are in MSVC or other compilers.

This seems like a messy maintenance nightmare with little utility to me.  If you want to test compatibility with a newer language, then you can set `c_std=c11` or whatever.  Using this warning level always means signing up for suppressing a bunch of things, so I'm happy for meson to just punt on the language level warnings.

That said, we could make the ubiquitous `Wc++98-compat` and `Wc++98-compat-pedantic`, as well as the various prehistoric `Wtraditional` flags go away by being just the right amount of opinionated: draw a line in the sand and say that meson's `warning_level=everything` is intended for "modern" code and suppresses anything related to "legacy" languages (pre-C99 or pre-C++11) for convenience.  This seems like an obvious line for "modern" here: both languages changed significantly then, but there's been over a decade of adoption, and they have changed less dramatically since.

I think incompatibility issues after that are uncommon enough that leaving the rest as the project's problem is fine.  "More or less everything relevant in the last decade, and I'll deal with tidying that up" is inherent to this `warning_level` anyway (zero suppressions is usually impossible, even for the strictest code).  If someone feels like the language warning suppression spam is too much, they can make meson smarter about it later.

... of course, that's all just an argument from laziness, but I think it's a pretty good one anyway :)

mesonbuild/compilers/mixins/gnu.py
drobilla: I've tried to explain why I don't think it's much of a burden.  Why do you think it will be?

I understand that it's a large intimidating-looking list, and curating it was certainly a huge burden, but that's done now.  These lists only contain warnings that aren't already enabled by `-Wall -Wextra -pedantic`.  I could probably cobble together concrete statistics about that, but from skimming the release notes (and personal experience), I don't think it happens very often.

This is actually even more rare than that: meson would only "need" to update the lists if the new warning wasn't already covered by one of the existing ones.  Many of the existing flags actually enable whole groups of warnings (the set was minimized by pruning any redundancies lower in the tree, so to speak).

Anyway, worst case scenario, gcc adds some new extreme warning flags and meson doesn't update them.  Whatever.  Eager projects can add them if they want, or not.  Someone can update meson if they want, or not.  The version these come from is known and exceptions are exhaustively documented, so it's easy enough to do so.

There's no promises that meson could possibly make about this anyway.  Meson being slightly out of date relative to the current compiler is about the same as the current compiler being newer than meson.  By using `warning_level=everything`, you're explicitly signing up for all of this, so I don't think it's really a problem if meson lags or outright fails to update these lists.  Warnings at this level are not stable in any way on any compiler.

> I am not convinced that maintaining these lists is Meson's burden to bare

It's either meson, or countless other places eventually.  Meson generally encourages changes to be incorporated as special support in meson itself (often discouraging being more general in the process).  For better or worse, that's just how meson is.  It's here or nowhere.

tl;dr: It's not much of a burden, and it's not a big deal if meson fails to carry it for a while, or forever, anyway.

drobilla: I waded through the release notes back to GCC 9.  These are the only such warnings I can find:

-Wmismatched-tags
-Wredundant-tags

Both are a C++ consistency / MS portability thing, more "language stuff" if you squint right.

So, about two flags (give or take) over about 4 years, which nobody would miss.  I'd Say It's Fine™?

drobilla:
I went with `-Dwarning_level=everything`, and just ignoring the language level thing entirely so meson doesn't have to deal with it.  They get added sometimes, and most projects are going to suppress some, but that's what you're signing up for by using this warning level (in general, including with clang and MSVC), so... whatever.

drobilla:
Updated to organize the flags by version and only include the ones supported by the current compiler version.  At the one-time cost of a very tedious few hours, now `warning_level=everything` should do the right thing for any vaguely reasonable version of GCC without spamming the console / taking forever doing flag checks.

I'd say this is obviously the best thing from a user perspective (checking this many flags really sucks).  It's about the same to maintain: not a big sorted list, which can be inconvenient, but broken down by version, which makes forward maintenance easier and just seems like a more sensible/useful data set in general.

History stops here at 3.0.0, everything earlier is considered "historical" and ever-present as far as `warning_level=everything` is concerned.  Realistically this could be cut off much later, but it doesn't really cost anything to keep full fidelity here so I went back as far as seems reasonable (things get messy in the GCC 2 days in general, the EGCS fork was a thing, etc).

Code-wise, `GnuCompiler.supported_warn_args` might be weird.  Reworking `get_warn_args` seemed like unnecessary churn and this was the most convenient/localized place to put this code that I could figure out.

eli-schwartz:
Fantastic work, will review when I get a chance.

drobilla:
> This looks more or less correct.

Great, thanks.

> For the dict of versions-to-args, I did a spot check against my GCC manual and found a bunch of args which are already enabled by -Wall or -Wextra. I have not checked if they only got enabled by those in a given version...
> Should those be removed?

Hm.  Pobody's nerfect, but finding "a bunch" in a "spot check" surprises me, pretty sure I went over all of them.  Example?  Occasionally the documentation is a lie and/or misleading though, I'll go over them again.  This part unfortunately isn't cleanly machine readable from the definition files.

drobilla:
Thanks, but I did the same thing about 10 minutes ago :)

I missed `Wextra` mostly.

mesonbuild/compilers/mixins/gnu.py
```diff
@@ -92,6 +92,275 @@
...
+#   -Wctad-maybe-unsupported
+#   -Wnamespaces
+#   -Wtemplates
+gnu_cpp_warning_args = {
+    "0.0.0": [
+        "-Wctor-dtor-privacy",
+        "-Wdeprecated",
+        "-Weffc++",
+        "-Wimport",
+        "-Wnon-virtual-dtor",
```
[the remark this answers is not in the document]
drobilla: I think I am going to leave this one on purpose.  It's the only weird case like this, but since `-Weffc++` is so widely turned off, I think omitting it would mean people accidentally disable this (more useful and less sketchy) warning.

mesonbuild/compilers/mixins/gnu.py
```diff
@@ -92,6 +92,275 @@
...
+    ],
+    "3.3.0": [
+        "-Wmultichar",
+        "-Wswitch-default",
+        "-Wswitch-enum",
+        "-Wunused-macros",
+    ],
+    "3.4.0": [
+        "-Wendif-labels",
+        "-Winit-self",
```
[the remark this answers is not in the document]
drobilla: https://github.com/mesonbuild/meson/pull/10554/commits/d28bc77bbb615f0d97c0ac6feaba34338d1a0858

drobilla:
> Can you add a release notes snippet in docs/markdown/snippets/ mentioning the new warning level?

https://github.com/mesonbuild/meson/pull/10554/commits/36e321b68387345d8e2b05e3575fa7b8612a7d31

First ditch effort, perhaps needs more fleshing out.  I'll look at the previous notes for guidance.

drobilla:
> This part unfortunately isn't cleanly machine readable from the definition files.

For the record, this was wrong.  I did the `Wall` culling pseudo-automatically from the opt files but just forgot to do `Wextra`, which is there as well.

mesonbuild/compilers/mixins/gnu.py
```diff
@@ -92,6 +92,238 @@
...
+        "-Wpadded",
+    ],
+    "3.3.0": [
+        "-Wmultichar",
+        "-Wswitch-default",
+        "-Wswitch-enum",
+        "-Wunused-macros",
+    ],
+    "3.4.0": [
+        "-Winvalid-pch",
```
eli-schwartz, line 142: Meson sets this as part of default warning args (at warning_level=1) already.

drobilla: 7e796deda

eli-schwartz (approved):
LGTM once `fixup!` commits are applied via rebase --autosquash.

amitdo:
https://quuxplusone.github.io/blog/2018/12/06/dont-use-weverything/

eli-schwartz:
That blog post agrees that it's valid to use in one-off developer builds, which is why we added support in the first place.

drobilla:
Indeed, that is specifically about _defaults_.  If projects setting this warning level as the default becomes a thing, meson can and should strictly prevent that (along with `-Werror`), but this seems unlikely so I didn't bother.

eli-schwartz:
> If projects setting this warning level as the default becomes a thing, meson can and should strictly prevent that (along with` -Werror`)

Actually, a reasonably large number of people do set `default_options: ['werror=true']`, but we still don't want to emit a meson warning about this.

It's sufficient that there's a standard configurable way to set this back to false when the user disagrees with the developer. We really want to avoid cases where people say "it's a configure warning either way, I might as well just set it with `add_project_arguments`", or find some way to spell it that the compiler recognizes but meson doesn't.

This is in keeping with a theme of Meson -- there are some things projects really want to do, that maybe aren't the greatest general thing to do. Meson doesn't forbid doing it, and instead finds ways to compromise, by making it so that the easiest way to do it is via tuneable builtins. Projects can be happy that they tuned their builds to be the way they like them (sometimes considerably easier than if they had to hardcode it manually) and as a side effect users can trivially tune it how *they* like it.

See for example "the best way to vendor and carefully version your dependencies is to use regular system dependencies and then inject a wrap". And yes, there are people who use `default_options: ['wrap_mode=forcefallback']` too -- we'd rather they do that than use `subproject().get_variable()` which cannot be overridden by distros. :D

drobilla:
Fair enough.  I don't particularly care either way.  Unless meson had a somewhat enforced "default optionless builds should actually work for people as-is" philosophy everywhere, which it clearly doesn't (and even if it did, `--auto-features` would ruin it in practice anyway), there wouldn't be much of a win anyway.

PR #10556, opened 2022-07-02
Make --help output more consistent and narrower

keszybz:


eli-schwartz:
> Argparse doesn't use dots at the end, so some --help lines had a
> dot, and some didn't. So drop them everywhere for consistency.

Hmm, but why not add them everywhere for consistency? e.g. now you end up with some --help lines containing multiple sentences where only the first sentence has a dot.

keszybz:
> > Argparse doesn't use dots at the end, so some --help lines had a
> > dot, and some didn't. So drop them everywhere for consistency.
> 
> Hmm, but why not add them everywhere for consistency? e.g. now you end up with some --help lines containing multiple sentences where only the first sentence has a dot.

Two reasons: 1. there's an autogenerated messages for `--help` itself which use a dot and is provided implicitly by argparse. 2. the dots are not that useful here. They take up space and make the text *feel" longer. Help outputs like this are almost always without help. The few places where there are multiple sentences still read OK without the dot.

eli-schwartz:
> there's an autogenerated messages for `--help` itself which use a dot and is provided implicitly by argparse.

But this:
```
-h, --help                               show this help message and exit
```

also refrains from capitalizing the first word, while you went ahead and capitalized all the manually defined option entries, so the autogenerated --help is still inconsistent. So I don't see the problem with making this one autogenerated message be inconsistent about the dot as well.

keszybz:
Yes, that's true. But consistency is just one of the reasons. The other reason is that it's just better to not have the dots, as described above. Most --help outputs don't.

PR #10559, opened 2022-07-03
[skip ci] Update bug_report.md

trilobite-coder:
Small change for questions at the lower part of the issue template


eli-schwartz:
This seems more confusing than it originally was.

Can you clarify the problem you are trying to solve?

PR #10560, opened 2022-07-03
Update deps page

dreamer-coding-555:
I moved thread library info from its page to the dependencies page with the rest of the libraries.


troglobyte-coder:
And forgot the skip_ci... oh boi.

eli-schwartz:
The standalone wording on the dedicated Threads.md page feels awkward when simply moved into Dependencies.md.

eli-schwartz:
> And forgot the skip_ci

We do not use this anymore, we have file-based CI triggers.

PR #10573, opened 2022-07-06
Ignore encoding errors when scanning. Closes #10571.

### LINKED ISSUE #10571: Stripping non-ASCII characters
**Describe the bug**
We have some Fortran files that contain non-ASCII characters in comments. They compile fine but when used with meson, I get errors in `depscan.py`:
```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xd6 in position 3329: invalid continuation byte
```

Instead of writing a custom script that strips out these non-ASCII characters, I wonder if there's a way to do it from within meson. Alternatively, is it possible to just bypass this check or make it skip comments?

**To Reproduce**
I'm working off of this pyOptSparse build at mdolab/pyoptsparse#300. The offending Fortran source file is unfortunately not publicly available, but any Fortran file with a non-ASCII character in a comment line should do.

**Expected behavior**
First, I think better debugging info would be helpful. It's not clear from the message which file was the issue, and I had to check with `chardetect` to figure it out.

I think the best thing would be for meson to allow these characters to exist, since the compiler has no problems with it. If that's not possible, then I was wondering if meson provides any utilities in stripping characters from files, as that seems to be fairly useful.

**system parameters**
* what operating system (e.g. MacOS Catalina, Windows 10, CentOS 8.0, Ubuntu 18.04, etc.) Manjaro 21.2.6
* what Python version are you using? Python 3.9
* what `meson --version` 0.63.0
* what `ninja --version` if it's a Ninja build  `1.10.2.git.kitware.jobserver-1`

jpakkane:
In Fortran and C++ all the bits we care about are in ASCII. 8-bit
characters can only occur in comments and string literals and we don't
parse those.

Closes #10571

### DIFF
--- a/mesonbuild/scripts/depscan.py
+++ b/mesonbuild/scripts/depscan.py
@@ -62,7 +62,7 @@ class DependencyScanner:
     def scan_fortran_file(self, fname: str) -> None:
         fpath = pathlib.Path(fname)
         modules_in_this_file = set()
-        for line in fpath.read_text(encoding='utf-8').split('\n'):
+        for line in fpath.read_text(encoding='utf-8', errors='ignore').split('\n'):
             import_match = FORTRAN_USE_RE.match(line)
             export_match = FORTRAN_MODULE_RE.match(line)
             submodule_export_match = FORTRAN_SUBMOD_RE.match(line)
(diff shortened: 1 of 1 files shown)


dcbaker (approved):
It might be worth adding a comment for why we're ignoring errors, but otherwise LGTM

PR #10576, opened 2022-07-06
Add typing annotations to the wayland module

dcbaker:
This ends up being pretty small since they're already using the typed_*args helpers.

### DIFF
--- a/mesonbuild/interpreter/interpreter.py
+++ b/mesonbuild/interpreter/interpreter.py
@@ -1652,7 +1652,7 @@ class Interpreter(InterpreterBase, HoldableObject):
     @disablerIfNotFound
     @permittedKwargs(permitted_dependency_kwargs)
     @typed_pos_args('dependency', varargs=str, min_varargs=1)
-    def func_dependency(self, node, args, kwargs):
+    def func_dependency(self, node, args, kwargs) -> Dependency:
         # Replace '' by empty list of names
         names = [n for n in args[0] if n]
         if len(names) > 1:
--- a/mesonbuild/modules/unstable_wayland.py
+++ b/mesonbuild/modules/unstable_wayland.py
@@ -12,7 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import annotations
 import os
+import typing as T
 
 from . import ExtensionModule, ModuleReturnValue
 from ..build import CustomTarget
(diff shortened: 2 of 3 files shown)


mesonbuild/interpreter/interpreter.py
```diff
@@ -1651,7 +1651,7 @@ def func_find_library(self, node, args, kwargs):
     @disablerIfNotFound
     @permittedKwargs(permitted_dependency_kwargs)
     @typed_pos_args('dependency', varargs=str, min_varargs=1)
-    def func_dependency(self, node, args, kwargs):
+    def func_dependency(self, node, args, kwargs) -> Dependency:
```
tristan957, line 1654: Looks like you are missing annotations for the parameters, but overall looks good!

mesonbuild/modules/unstable_wayland.py
```diff
@@ -12,24 +12,47 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import annotations
 import os
+import typing as T
+
+from mesonbuild.programs import OverrideProgram
```
eli-schwartz, line 19: Shouldn't this be grouped with the other imports from mesonbuild? And why does this use absolute imports, while all the others are relative ones?

eli-schwartz: Also this isn't used at runtime, only as an annotation.

mesonbuild/modules/unstable_wayland.py
```diff
@@ -12,24 +12,47 @@
...
-    def __init__(self, interpreter):
+    def __init__(self, interpreter: Interpreter) -> None:
         super().__init__(interpreter)
 
-        self.protocols_dep = None
-        self.pkgdatadir = None
-        self.scanner_bin = None
+        self.protocols_dep: T.Optional[Dependency] = None
+        self.pkgdatadir: T.Optional[str] = None
+        self.scanner_bin: T.Optional[T.Union[ExternalProgram, OverrideProgram, Executable]] = None
```
eli-schwartz, line 55: You don't need to mention OverrideProgram, since it's a type of ExternalProgram anyway.

mesonbuild/interpreter/interpreter.py
eli-schwartz: That's fine, because that is just the minimum annotations needed in the interpreter to get the wayland module typed. :p

Issue #10577, opened 2022-07-07
GJS tests get stuck since Meson 0.63.0


heftig:
**Describe the bug**

Some of GJS 1.72.1's tests time out since Meson 0.63.0. Meson 0.62.2 is fine.
GJS issue report at https://gitlab.gnome.org/GNOME/gjs/-/issues/490.

Stderr is no longer properly consumed for TAP tests, so the test gets stuck when the pipe is full. It seems #10471 broke this. 

**To Reproduce**
```
git clone https://gitlab.gnome.org/GNOME/gjs.git
cd gjs
git checkout 1.72.1
meson . b
meson compile -C b
meson test -C b --print-errorlogs
```

**Expected behavior**
Tests pass.

**system parameters**
* Native build
* Arch Linux
* Python 3.10.5
* Meson 0.63.0
* Ninja 1.11.0

eli-schwartz:
This was also reported to @jpakkane in Debian as https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1014417

PR #10580, opened 2022-07-07
fix mtest regression

### LINKED ISSUE #10563: testlog.txt etc don't contain output of tap tests anymore
Starting with 702f3fe15 meson-logs/testlog.txt don't contain output from tap test failures anymore. And not just when the test succeed, even when the tap test failed. @trhd I assume this was not intentional?

Reproducer:
```
project('repro', ['c'])

test('ex', find_program('echo'),
     args: ['ok 1 - frak
# diag
not ok 2 - blarg
1..2
'],
     protocol: 'tap',
    )
```
with 0.62 that'll include 
```
----------------------------------- output -----------------------------------
ok 1 - frak
# diag
not ok 2 - blarg
1..2
```
in the log, but with 702f3fe15 that's not included anymore.

Obviously testlog.txt (etc) need to include the test output...

trhd:
This is a cherry-picked commit from #9930. Fixes #10563 and #10577.

### DIFF
--- a/test cases/common/253 long output/meson.build
+++ b/test cases/common/253 long output/meson.build
@@ -1,0 +1,5 @@
+project('long-stderr', 'c')
+
+dumper = executable('dumper', 'dumper.c')
+test('dump-test', dumper)
+test('dump-test-TAP', dumper, protocol : 'tap')
--- a/mesonbuild/mtest.py
+++ b/mesonbuild/mtest.py
@@ -1108,7 +1108,9 @@ def decode(stream: T.Union[None, bytes]) -> str:
     except UnicodeDecodeError:
         return stream.decode('iso-8859-1', errors='ignore')
 
-async def read_decode(reader: asyncio.StreamReader, console_mode: ConsoleUser) -> str:
+async def read_decode(reader: asyncio.StreamReader,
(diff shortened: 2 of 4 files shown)


anarazel (commented):
Looks like a pretty big change for a bugfix that needs to be backpatched - could this be done in a narrower fashion?

Would be good to add a test for the currently broken logging.

eli-schwartz:
For fixing the stable release, having small changes that are easily vetted in scope is crucial.

test cases/common/253 long output/meson.build
```diff
@@ -0,0 +1,5 @@
+project('long-stderr', 'c')
+
+dumper = executable('dumper', 'dumper.c')
+test('dump-test', dumper)
+test('dump-test-TAP', dumper, protocol : 'tap')
```
anarazel, line 5: Hm - nothing in this test verifies that the test log etc contains relevant output, correct?

jpakkane:
Is this now good enough that it fixes all known regressions? IOW if we merge this can we do a point release and fix the remaining regressions in the next one?

anarazel:
I don't know the code in mtest well enough to fully judge, but I can confirm that this fixes the problem. The expanded test in the last revision also increases coverage of the facility by a good bit, I think...

PR #10582, opened 2022-07-08
Make more uses of `mlog.warning` non-fatal

dcbaker:
At least, that's how it started. Then I noticed that I wasn't get completions, and realized that I was lazy when I annotated the mlog module, and used lots of `**kwargs: T.Any`, which is less than ideal. In cleaning tha tup I found some annoying mypy issues, as well as a few actual bugs or shortcomings in the exiting code. Finally we get to adding `fatal=False` in a bunch of places where the end user cannot reasonably be expected to fix the warning (old versions of software, the need to support old versions of meson, etc). I noticed this when there was a suggestion to bump the required c++ version of a wrap to c++14 because of a warning that MSVC doesn't support c++11, and we just set it to c++14 instead.

### DIFF
--- a/mesonbuild/backend/backends.py
+++ b/mesonbuild/backend/backends.py
@@ -284,7 +284,7 @@ class Backend:
         if isinstance(t, build.CustomTarget):
             if warn_multi_output and len(t.get_outputs()) != 1:
                 mlog.warning(f'custom_target {t.name!r} has more than one output! '
-                             'Using the first one.')
+                             f'Using the first one. Consider using `{t.name}[0]`.')
             filename = t.get_outputs()[0]
         elif isinstance(t, build.CustomTargetIndex):
             filename = t.get_outputs()[0]
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -782,7 +782,8 @@ class CoreData:
                 try:
                     value.set_value(oldval.value)
                 except MesonException:
-                    mlog.warning(f'Old value(s) of {key} are no longer valid, resetting to default ({value.value}).')
+                    mlog.warning(f'Old value(s) of {key} are no longer valid, resetting to default ({value.value}).',
+                                 fatal=False)
 
     def is_cross_build(self, when_building_for: MachineChoice = MachineChoice.HOST) -> bool:
         if when_building_for == MachineChoice.BUILD:
--- a/mesonbuild/dependencies/dev.py
+++ b/mesonbuild/dependencies/dev.py
@@ -534,8 +534,11 @@ class JNISystemDependency(SystemDependency):
         modules: T.List[str] = mesonlib.listify(kwargs.get('modules', []))
         for module in modules:
             if module not in {'jvm', 'awt'}:
-                log = mlog.error if self.required else mlog.debug
-                log(f'Unknown JNI module ({module})')
+                msg = f'Unknown JNI module ({module})'
+                if self.required:
+                    mlog.error(msg)
+                else:
+                    mlog.debug(msg)
                 self.is_found = False
                 return
 
@@ -553,8 +556,11 @@ class JNISystemDependency(SystemDependency):
                     res = subprocess.run(['/usr/libexec/java_home', '--failfast', '--arch', m.cpu_family],
                                          stdout=subprocess.PIPE)
                     if res.returncode != 0:
-                        log = mlog.error if self.required else mlog.debug
-                        log('JAVA_HOME could not be discovered on the system. Please set it explicitly.')
+                        msg = 'JAVA_HOME could not be discovered on the system. Please set it explicitly.'
+                        if self.required:
+                            mlog.error(msg)
+                        else:
+                            mlog.debug(msg)
                         self.is_found = False
                         return
                     self.java_home = pathlib.Path(res.stdout.decode().strip())
--- a/mesonbuild/dependencies/dev.py
+++ b/mesonbuild/dependencies/dev.py
@@ -422,7 +422,7 @@ class LLVMDependencyCMake(CMakeDependency):
         # cmake if dynamic is required
         if not self.static:
             self.is_found = False
-            mlog.warning('Ignoring LLVM CMake dependency because dynamic was requested')
+            mlog.warning('Ignoring LLVM CMake dependency because dynamic was requested', fatal=False)
             return
 
         if self.traceparser is None:
@@ -455,7 +455,7 @@ class LLVMDependencyCMake(CMakeDependency):
                 if required:
                     raise self._gen_exception(f'LLVM module {mod} was not found')
                 else:
-                    mlog.warning('Optional LLVM module', mlog.bold(mod), 'was not found')
+                    mlog.warning('Optional LLVM module', mlog.bold(mod), 'was not found', fatal=False)
                     continue
             for i in cm_targets:
                 res += [(i, required)]
--- a/mesonbuild/environment.py
+++ b/mesonbuild/environment.py
@@ -455,7 +455,7 @@ class Environment:
                 # If we stored previous command line options, we can recover from
                 # a broken/outdated coredata.
                 if os.path.isfile(coredata.get_cmd_line_file(self.build_dir)):
-                    mlog.warning('Regenerating configuration from scratch.')
+                    mlog.warning('Regenerating configuration from scratch.', fatal=False)
                     mlog.log('Reason:', mlog.red(str(e)))
                     coredata.read_cmd_line_file(self.build_dir, options)
                     self.create_new_coredata(options)
@@ -551,7 +551,8 @@ class Environment:
         if bt in self.options and (db in self.options or op in self.options):
             mlog.warning('Recommend using either -Dbuildtype or -Doptimization + -Ddebug. '
                          'Using both is redundant since they override each other. '
-                         'See: https://mesonbuild.com/Builtin-options.html#build-type-options')
+                         'See: https://mesonbuild.com/Builtin-options.html#build-type-options',
+                         fatal=False)
 
         exe_wrapper = self.lookup_binary_entry(MachineChoice.HOST, 'exe_wrapper')
         if exe_wrapper is not None:
--- a/mesonbuild/interpreter/interpreter.py
+++ b/mesonbuild/interpreter/interpreter.py
@@ -2905,10 +2905,12 @@ class Interpreter(InterpreterBase, HoldableObject):
             return
         if (self.coredata.options[OptionKey('b_lundef')].value and
                 self.coredata.options[OptionKey('b_sanitize')].value != 'none'):
-            mlog.warning('''Trying to use {} sanitizer on Clang with b_lundef.
-This will probably not work.
-Try setting b_lundef to false instead.'''.format(self.coredata.options[OptionKey('b_sanitize')].value),
-                         location=self.current_node)
+            value = self.coredata.options[OptionKey('b_sanitize')].value
+            mlog.warning(textwrap.dedent(f'''\
+                    Trying to use {value} sanitizer on Clang with b_lundef.
+                    This will probably not work.
+                    Try setting b_lundef to false instead.'''),
+                location=self.current_node)  # noqa: E128
 
     # Check that the indicated file is within the same subproject
     # as we currently are. This is to stop people doing
--- a/mesonbuild/mlog.py
+++ b/mesonbuild/mlog.py
@@ -495,3 +495,14 @@ def stop_pager() -> None:
             pass
         log_pager.wait()
         log_pager = None
+
+
+def code_line(text: str, line: str, colno: int) -> str:
+    """Print a line with a caret pointing to the colno
+
+    :param text: A message to display before the line
+    :param line: The line of code to be pointed to
+    :param colno: The column number to point at
+    :return: A formatted string of the text, line, and a caret
+    """
+    return f'{text}\n{line}\n{" " * colno}^'
--- a/mesonbuild/mlog.py
+++ b/mesonbuild/mlog.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 from __future__ import annotations
+import enum
 import os
 import io
 import sys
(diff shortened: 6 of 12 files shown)


mesonbuild/mlog.py
```diff
@@ -326,25 +336,49 @@ def _log_error(severity: str, *rargs: TV_Loggable,
...
-def notice(*args: TV_Loggable, **kwargs: T.Any) -> None:
-    return _log_error('notice', *args, **kwargs, is_error=False)
+def error(*args: TV_Loggable,
+          once: bool = False, fatal: bool = True,
+          location: T.Optional[BaseNode] = None,
+          nested: bool = True, sep: T.Optional[str] = None,
+          end: T.Optional[str] = None,
+          is_error: bool = True) -> None:
+    return _log_error('error', *args, once=once, fatal=fatal, location=location,
+                      nested=nested, sep=sep, end=end, is_error=is_error)
```
eli-schwartz, line 362: Before this change, it didn't accept an is_error kwarg, but always forced it to `True`.

mesonbuild/backend/backends.py
```diff
@@ -300,7 +300,7 @@ def get_target_filename(self, t: T.Union[build.Target, build.CustomTargetIndex],
         if isinstance(t, build.CustomTarget):
             if warn_multi_output and len(t.get_outputs()) != 1:
                 mlog.warning(f'custom_target {t.name!r} has more than one output! '
-                             'Using the first one.')
+                             f'Using the first one. Consider using `{t.name}[0]`.')
```
eli-schwartz, line 303: This is an unrelated change and deserves its own commit.

mesonbuild/backend/backends.py
```diff
@@ -1475,7 +1475,7 @@ def create_install_data(self) -> InstallData:
         strip_bin = self.environment.lookup_binary_entry(MachineChoice.HOST, 'strip')
         if strip_bin is None:
             if self.environment.is_cross_build():
-                mlog.warning('Cross file does not specify strip binary, result will not be stripped.')
+                mlog.warning('Cross file does not specify strip binary, result will not be stripped.', fatal=False)
```
eli-schwartz, line 1478: Tbh this is a bad cross file configuration which causes your configure settings to be ignored, it's absolutely a user thing to fix.

mesonbuild/interpreter/compiler.py
```diff
@@ -659,7 +659,7 @@ def get_supported_arguments_method(self, args: T.Tuple[T.List[str]], kwargs: 'Ge
             if not self._has_argument_impl([arg]):
                 msg = f'Compiler for {self.compiler.get_display_language()} does not support "{arg}"'
                 if checked == 'warn':
-                    mlog.warning(msg)
+                    mlog.warning(msg, fatal=False)
```
eli-schwartz, line 662: This seems like something that should be a fatal warning?

mesonbuild/interpreter/interpreter.py
```diff
@@ -2845,10 +2847,12 @@ def check_clang_asan_lundef(self) -> None:
...
-            mlog.warning('''Trying to use {} sanitizer on Clang with b_lundef.
-This will probably not work.
-Try setting b_lundef to false instead.'''.format(self.coredata.options[OptionKey('b_sanitize')].value),
-                         location=self.current_node)
+            mlog.warning(textwrap.dedent(f'''\
+                    Trying to use
+                    {self.coredata.options[OptionKey('b_sanitize')].value }
+                    sanitizer on Clang with b_lundef.  This will probably not work.
+                    Try setting b_lundef to false instead.'''),
+                location=self.current_node, fatal=False)
```
eli-schwartz, line 2855: If it won't work, why isn't it a fatal warning to make you stop doing it? ISTR that "won't work" means linker errors, not "clang just doesn't set up sanitizer support".

mesonbuild/mcompile.py
```diff
@@ -224,7 +224,7 @@ def get_parsed_args_vs(options: 'argparse.Namespace', builddir: Path) -> T.Tuple
         cmd.append('-maxCpuCount')
 
     if options.load_average:
-        mlog.warning('Msbuild does not have a load-average switch, ignoring.')
+        mlog.warning('Msbuild does not have a load-average switch, ignoring.', fatal=False)
```
eli-schwartz, line 227: This CLI subtool doesn't have a `--fatal-meson-warnings` flag.

mesonbuild/mesonlib/universal.py
```diff
@@ -1950,7 +1950,7 @@ def get_wine_shortpath(winecmd: T.List[str], wine_paths: T.Sequence[str],
             os.unlink(bat_file.name)
     wine_path = ';'.join(rel_paths + wine_paths)
     if len(wine_path) > 1024:
-        mlog.warning('WINEPATH exceeds 1024 characters which could cause issues')
+        mlog.warning('WINEPATH exceeds 1024 characters which could cause issues', fatal=False)
```
eli-schwartz, line 1953: Should the user fix this by moving the worktree, and thus maybe want it to error out?

mesonbuild/interpreter/compiler.py
dcbaker: oops, you're right. I misread that as "compiler does not support *checking* for arg", I'll change that back.

mesonbuild/interpreter/interpreter.py
dcbaker: We've had reports of it working on some OSes and failing on others. I don't konw what the difference between them is, but I can drop this change

mesonbuild/mesonlib/universal.py
dcbaker: I have no idea, TBH. The comment is "this could cause issues" seems to imply that in some cases it's fine, in others it's not and we don't konw, so if it fails you can figure it out.

dcbaker: I'll just leav it as fatal and we can change it later if we need to

mesonbuild/backend/backends.py
dcbaker: I'm not sure I agree, I never specify strip in my cross files because I simply don't care about stripping them (I'm basically just compile testing, I'm not actually going to use them). but It's easy enough for someone to fix if they want fatal warnings I guess. Dropped

dcbaker:
I split all of the changes out of the `fatal=False` that are not specifically `fatal=False` additions, and addressed the other comments.

mesonbuild/mparser.py
```diff
@@ -193,10 +193,9 @@ def lex(self, filename: str) -> T.Generator[Token, None, None]:
                     elif tid in {'string', 'fstring'}:
                         # Handle here and not on the regexp to give a better error message.
                         if match_text.find("\n") != -1:
-                            msg = ParseException("Newline character in a string detected, use ''' (three single quotes) "
-                                                 "for multiline strings instead.\n"
-                                                 "This will become a hard error in a future Meson release.",
-                                                 self.getline(line_start), lineno, col)
+                            msg = ("Newline character in a string detected, use ''' (three single quotes) "
```
eli-schwartz, line 196: Taking a closer look at this...

The ParseException generates a message string that includes the actual line which triggered the issue, with a caret pointing at the exact position of the issue.

The commit message is wrong to claim that passing the message directly is the same if we don't raise the exception.

To quell mypy, what we could do is `mlog.warning(str(msg), ...)`

mesonbuild/coredata.py
```diff
@@ -782,7 +782,8 @@ def update_project_options(self, options: 'MutableKeyedOptionDictType') -> None:
                 try:
                     value.set_value(oldval.value)
                 except MesonException:
-                    mlog.warning(f'Old value(s) of {key} are no longer valid, resetting to default ({value.value}).')
+                    mlog.warning(f'Old value(s) of {key} are no longer valid, resetting to default ({value.value}).',
+                                 fatal=False)
```
eli-schwartz, line 786: I'm conflicted about this. On the one hand, the user absolutely can and really should do something about this, unlike the MSVC std case where the user cannot fix Microsoft.

On the other hand, this may be triggered by `ninja` updating build.ninja by running a reconfigure, which drops --fatal-meson-warnings anyway.

On the other other hand, in that case the value of fatal= is irrelevant...

Do we really want to avoid failing here?

mesonbuild/environment.py
```diff
@@ -568,7 +568,8 @@ def __init__(self, source_dir: T.Optional[str], build_dir: T.Optional[str], opti
         if bt in self.options and (db in self.options or op in self.options):
             mlog.warning('Recommend using either -Dbuildtype or -Doptimization + -Ddebug. '
                          'Using both is redundant since they override each other. '
-                         'See: https://mesonbuild.com/Builtin-options.html#build-type-options')
+                         'See: https://mesonbuild.com/Builtin-options.html#build-type-options',
+                         fatal=False)
```
eli-schwartz, line 572: This is something users can presumably fix, as we're nagging them about the CLI options they pass, on the other hand idk if I care.

mesonbuild/wrap/wrap.py
```diff
@@ -81,7 +81,7 @@ def open_wrapdburl(urlstring: str, allow_insecure: bool = False, have_opt: bool
             msg = f'WrapDB connection failed to {urlstring} with error {excp}.'
             if isinstance(excp.reason, ssl.SSLCertVerificationError):
                 if allow_insecure:
-                    mlog.warning(f'{msg}\n\n    Proceeding without authentication.')
+                    mlog.warning(f'{msg}\n\n    Proceeding without authentication.', fatal=False)
```
eli-schwartz, line 84: This CLI subtool doesn't have a `--fatal-meson-warnings` flag.

eli-schwartz (changes requested):

eli-schwartz:
"wrap/warp" in a commit message. ;)

mesonbuild/wrap/wrap.py
dcbaker: Oh, right. For some reason I thought this was a common path when fetching sources as well

dcbaker:
Already fixed the wrap/wrap locally, but didn’t want to trigger another ci run good that unless it was good otherwise :)

mesonbuild/wrap/wrap.py
eli-schwartz: Hold on, that's a good point and we do. We auto-retrieve at least patch_url from the wrapdb when doing subproject fallback.

eli-schwartz: And triple-checking, we cannot actually hit the SSL codepath in subproject fallback, because we don't support `--allow-insecure` except in the wrap subtool. Subproject fallback will always error out instead of emitting this `mlog.warning()`.

eli-schwartz:
Can we get a reroll of this series?

eli-schwartz:
ping?

mesonbuild/environment.py
dcbaker: This is one people complain about all the time, since they want to set their buildtype to something, and their optimization level to something else, like `-Dbuildtype=release -Doptimization=3` (or 2) seems like a legitimate thing to do, ie: "I'm building release, but optlevel 2 isn't fast enough or optlevel 3 breaks." Making that a hard error seems not very nice.

mesonbuild/mparser.py
dcbaker: I've split the code for generating that message out into a helper function and used that here instead.

dcbaker:
Respun with the obvious comments fixed

mesonbuild/mlog.py
```diff
@@ -495,3 +495,17 @@ def stop_pager() -> None:
...
+
+    :param text: A message to display before the line
+    :param line: The line of code to be pointed to
+    :param colno: The column number to point at
+    :return: A formatted string of the text, line, and a caret
+    """
+    return f'''\
+        {text}
+        {line}
+        {' ' * colno}^'''
```
eli-schwartz, line 511: This intersperses whitespace into the printed string, unlike the single-line version this is copied from.

eli-schwartz: Is it really better to dedent this? Why not just use the pretty readable original string?

This is still broken, so maybe it's not all that readable using dedent. :)

mesonbuild/mparser.py
```diff
@@ -194,11 +193,10 @@ def lex(self, filename: str) -> T.Generator[Token, None, None]:
...
                         if match_text.find("\n") != -1:
-                            msg = ParseException("Newline character in a string detected, use ''' (three single quotes) "
-                                                 "for multiline strings instead.\n"
-                                                 "This will become a hard error in a future Meson release.",
-                                                 self.getline(line_start), lineno, col)
-                            mlog.warning(msg, location=BaseNode(lineno, col, filename))
+                            msg = ("Newline character in a string detected, use ''' (three single quotes) "
+                                   "for multiline strings instead.\n"
+                                   "This will become a hard error in a future Meson release.")
+                            mlog.warning(mlog.code_line(msg, self.getline(lineno), col))
```
eli-schwartz, line 199: Alright, just tried testing this and it no longer even prints the correct line.

```suggestion
                            mlog.warning(mlog.code_line(msg, self.getline(line_start), col), location=BaseNode(lineno,col, filename))
```

mesonbuild/mlog.py
```diff
@@ -457,3 +496,17 @@ def stop_pager() -> None:
...
+def code_line(text: str, line: str, colno: int) -> str:
+    """Print a line with a caret pointing the colno
+
+    :param text: A message to display before the line
+    :param line: The line of code to be pointed to
+    :param colno: The column number to point at
+    :return: A formatted string of the text, line, and a caret
+    """
+    return textwrap.dedent(f'''\
+        {text}
```
eli-schwartz, line 510: text can contain lines without any indentation. textwrap.dedent then doesn't indent anything.

eli-schwartz (changes requested):
I would just drop these two commits, they're not worth it and it's not obviously better.

eli-schwartz (commented):
niggling grammar nit

eli-schwartz (approved):

mesonbuild/dependencies/dev.py
```diff
@@ -534,8 +534,11 @@ def __init__(self, environment: 'Environment', kwargs: JNISystemDependencyKW):
         modules: T.List[str] = mesonlib.listify(kwargs.get('modules', []))
         for module in modules:
             if module not in {'jvm', 'awt'}:
-                log = mlog.error if self.required else mlog.debug
-                log(f'Unknown JNI module ({module})')
+                msg = f'Unknown JNI module ({module})'
+                if self.required:
+                    mlog.error(msg)
+                else:
+                    mlog.debug(msg)
```
eli-schwartz, line 541: > There should be a way to make mypy happy without casting, but I can't figure it out, since the mlog.error and mlog.debug actually have different signatures.

Hmm, mypy still fails, because in commit 7254db36a7adfa3f5e3e6cbce5dc1e3037893eb9 we added more `log = mlog.error if self.required else mlog.debug`.

eli-schwartz (changes requested):

PR #10587, opened 2022-07-11
add test for global variable

ailin-nemui:
source: https://lists.gnu.org/archive/html/libtool/2002-07/msg00025.html


eli-schwartz:
I would put this in "common/5 linkstatic"

eli-schwartz:
This test case was included in #10628 alongside the actual fix.

PR #10588, opened 2022-07-11
add intel oneAPI ifort compiler to azure and run fortran tests

eli-schwartz:
And fix the regression that totally broke it.

@Hofer-Julian this should fix the issue we discussed.

@dcbaker thoughts?

...

Actually running this correctly is a bag of mixed results. I don't understand why the `vc2019x64vs` job has 13 failing tests in which, randomly, I get errors about how it cannot find some command (e.g. rc.exe, but also cmd.exe)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,3 +1,17 @@
+# intel oneapi installation is based on the examples at
+# https://github.com/oneapi-src/oneapi-ci
+#
+# and are copyrighted:
+#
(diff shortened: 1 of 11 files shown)


dcbaker:
I’m in favor of not breaking things on windows, so this seems like a good idea to me. I didn’t look at the implantation closely, but it seems ok

dcbaker (approved):

Hofer-Julian:
Thanks!

PR #10595, opened 2022-07-12
cmake: allow dynamic linking with LLVM

### LINKED ISSUE #10592: Is llvm with CMake is always static and nothing changed from 0.51x?
**Describe the bug**
I want to cross-compile with CMake-based LLVM detection (without LLVM-config). LLVM should be dynamic lib (at best, it should find libLLVM.so) with CMake. We cannot use llvm-config from cross-file due to qemu requirement (to run mipsel binaries on amd64), so, I do not provide it in cross file.

**To Reproduce**
Make native mipsel debian chroot via debootstrap
Try to compile some llvm-dependent package with sysroot of this chroot. Non-llvm dependent library works fine.

**Expected behavior**
LLVM should be found without using qemy bu CMake usage.

**system parameters**
* Cross build, Ubuntu 20.04 -> debian bullseye mipsel
* 3.8.10
* 0.63.0
* 1.10.0
Cross-file:

```
[constants]
fs = '/mnt/chroot/debian'
common_flags = ['--sysroot=' + fs ]

[host_machine]
system = 'linux'
cpu_family = 'mips'
cpu = 'mips32'
endian = 'little'

[properties]
sizeof_int = 4
sizeof_wchar_t = 4
sizeof_void* = 4

alignment_char = 1
alignment_void* = 4
alignment_double = 4

has_function_printf = true

sys_root = fs
pkg_config_libdir = [fs + '/usr/lib/mipsel-linux-gnu/pkgconfig', fs + '/usr/share/pkgconfig']

c_args = common_flags + ['-I' + fs + '/usr/include', '-L' + fs + '/usr/lib/mipsel-linux-gnu']

[binaries]
c = 'mipsel-linux-gnu-gcc'
cpp = 'mipsel-linux-gnu-g++'
strip = 'mipsel-linux-gnu-strip'
pkgconfig = 'pkg-config'
```

rilian-la-te:
This patch allows to link dynamically with LLVM when CMake requested.

For now, only linking with libLLVM.so is supported (not linking to individual libraries).
There are certainly some bugs.

Fixes #10592

### DIFF
--- a/mesonbuild/dependencies/data/CMakeListsLLVM.txt
+++ b/mesonbuild/dependencies/data/CMakeListsLLVM.txt
@@ -1,8 +1,24 @@
+# fail noisily if attempt to use this file without setting:
+# cmake_minimum_required(VERSION ${CMAKE_VERSION})
+# project(... LANGUAGES ...)
+
+cmake_policy(SET CMP0000 NEW)
 
 set(PACKAGE_FOUND FALSE)
 
+list(REMOVE_DUPLICATES LLVM_MESON_VERSIONS)
+
 while(TRUE)
-  find_package(LLVM REQUIRED CONFIG QUIET)
+  #Activate CMake version selection
+  foreach(i IN LISTS LLVM_MESON_VERSIONS)
+    find_package(LLVM ${i}
+      CONFIG
+      NAMES ${LLVM_MESON_PACKAGE_NAMES}
+      QUIET)
+    if(LLVM_FOUND)
+      break()
+    endif()
+  endforeach()
 
   # ARCHS has to be set via the CMD interface
   if(LLVM_FOUND OR "${ARCHS}" STREQUAL "")
@@ -13,9 +29,70 @@ while(TRUE)
   list(REMOVE_AT ARCHS 0)
 endwhile()
 
-if(LLVM_FOUND)
-  set(PACKAGE_FOUND TRUE)
+function(meson_llvm_cmake_dynamic_available mod out)
+  # Check if we can only compare LLVM_DYLIB_COMPONENTS, because
+  # we do not need complex component translation logic, if all
+  # is covered by one variable
+  if(mod IN_LIST LLVM_DYLIB_COMPONENTS)
+    set(${out} TRUE PARENT_SCOPE)
+    return()
+  elseif((NOT (mod IN_LIST LLVM_DYLIB_COMPONENTS))
+      AND (NOT("${LLVM_DYLIB_COMPONENTS}" STREQUAL "all")))
+    set(${out} FALSE PARENT_SCOPE)
+    return()
+  endif()
 
+  # Complex heurisic to filter all pseudo-components and skip invalid names
+  # LLVM_DYLIB_COMPONENTS will be 'all', because in other case we returned
+  # in previous check. 'all' is also handled there.
+  set(llvm_pseudo_components "native" "backend" "engine" "all-targets")
+  is_llvm_target_specifier(${mod} mod_spec INCLUDED_TARGETS)
+  string(TOUPPER "${LLVM_AVAILABLE_LIBS}" capitalized_libs)
+  string(TOUPPER "${LLVM_TARGETS_TO_BUILD}" capitalized_tgts)
+  if(mod_spec)
+    set(${out} TRUE PARENT_SCOPE)
+  elseif(mod IN_LIST capitalized_tgts)
+    set(${out} TRUE PARENT_SCOPE)
+  elseif(mod IN_LIST llvm_pseudo_components)
+    set(${out} TRUE PARENT_SCOPE)
+  elseif(LLVM${mod} IN_LIST capitalized_libs)
+    set(${out} TRUE PARENT_SCOPE)
+  else()
+    set(${out} FALSE PARENT_SCOPE)
+  endif()
+endfunction()
+
+function(is_static target ret)
+  if(TARGET ${target})
+    get_target_property(target_type ${target} TYPE)
+    if(target_type STREQUAL "STATIC_LIBRARY")
+      set(${ret} TRUE PARENT_SCOPE)
+      return()
+    endif()
+  endif()
+  set(${ret} FALSE PARENT_SCOPE)
+endfunction()
+
+# Concatenate LLVM_MESON_REQUIRED_MODULES and LLVM_MESON_OPTIONAL_MODULES
+set(LLVM_MESON_MODULES ${LLVM_MESON_REQUIRED_MODULES} ${LLVM_MESON_OPTIONAL_MODULES})
+
+
+# Check if LLVM exists in dynamic world
+# Initialization before modules checking
+if(LLVM_FOUND)
+  if(LLVM_MESON_DYLIB AND TARGET LLVM)
+    set(PACKAGE_FOUND TRUE)
+  elseif(NOT LLVM_MESON_DYLIB)
+    # Use LLVMSupport to check if static targets exist
+    set(static_tg FALSE)
+    is_static(LLVMSupport static_tg)
+    if(static_tg)
+      set(PACKAGE_FOUND TRUE)
+    endif(static_tg)
+  endif()
+endif()
+
+if(PACKAGE_FOUND)
   foreach(mod IN LISTS LLVM_MESON_MODULES)
     # Reset variables
     set(out_mods)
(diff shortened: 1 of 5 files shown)


dcbaker:
If you want to test this locally `./run_single_test.py "test cases/framework/15 llvm" -- -Dmethod=cmake -Dlink-static=(true|false)` will do what you want.

dcbaker:
Also, I don't want to support linking to individual `.so`s, unless upstream LLVM has changed their policy and support that (last I knew they didn't support that, and I don't want to take on supporting something they don't)

rilian-la-te:
@dcbaker this PR does not want to support to linking to individual .so files. This PR want to support linking to libLLVM.so, and check if requested components is present in libLLVM.so. 

No individual component linking should be involved.

dcbaker:
Right, sorry, you had mentiond that wasn't supported by your PR, and I just wanted to clarify we don't want that anyway :)

rilian-la-te:
@dcbaker do you know how to disable single test on Ubuntu Bionic? It have LLVM 6.0, which cannot find dynamically linked modules by CMake dep (LLVM limitation).

dcbaker:
You’ll probably need to add a check inside the meson.build file that checks the llvm version and calls error() with `MESON_SKIP_TEST message`. There’s other tests that do this you can look at. You may need to do something like just call dependency(llvm) to get the version info

rilian-la-te:
@dcbaker I want to remove UNEXPSKIP on Ubuntu Bionic and make skip expected via test.json, but I cannot find how to do it properly and not skip tests on other targets.

rilian-la-te:
@dcbaker and what should I do to pass a CI? I tried to install llvm-7 on bionic, but images cannot be created. And for llvm < 7 we cannot support dynamic linking with LLVM < 7 with cmake without random bugs.

dcbaker:
I started looking at the testing code, and I I can add support for skipping specific ci configurations, I just don’t get it done Friday

rilian-la-te:
@dcbaker I think it is better just to install llvm-7 to bionic CI image.

rilian-la-te:
@mensinda is my approach to make CMake with LLVM working is correct?

kanavin:
I support this. Running llvm-config really does not work when cross-compiling.

dcbaker:
since bionic has been dropped from our CI this may Just Work™ if you rebase it.

rilian-la-te:
@dcbaker I rebased it, but bionic is still here, and ArchLinux MPI unrelated failure too.

rilian-la-te:
@dcbaker can you remove Bionic from meson CI at last or just update LLVM in its image?

rilian-la-te:
@mensinda will this be addresed? I think dynamic LLVM is cruical for cross-compilation. For example, Mesa can greatly benefit from it)

dcbaker:
@rilian-la-te Can you rebase this, and I’ll have a look at reviewing it again?

rilian-la-te:
@dcbaker already rebased. One failing CI check - skipping (UNEXPSKIP) LLVM tests on Bionic (because cmake dynamic llvm can be used only on LLVM 7+ and Bionic have 6.0).

dcbaker:
Ohhh that’s right. I started working on some infrastructure changes for dealing with that ash’s never finished…

rilian-la-te:
@dcbaker can you review this changes? It seems than this PR was forgotten, but it can be very useful for cross-compilation with distros' builtin LLVM (I tested it in Debian - it is usable for cross-compiling there with their multiarch approach).

mesonbuild/dependencies/data/CMakeListsLLVM.txt
```diff
@@ -1,8 +1,24 @@
...
+list(REMOVE_DUPLICATES LLVM_MESON_VERSIONS)
+
 while(TRUE)
-  find_package(LLVM REQUIRED CONFIG QUIET)
+  #Activate CMake version selection
+  foreach(i IN LISTS LLVM_MESON_VERSIONS)
+    find_package(LLVM ${i}
+    CONFIG
+    NAMES ${LLVM_MESON_PACKAGE_NAMES}
+    QUIET)
```
mensinda, line 17: This could be indented a bit for better readability

mensinda (commented):
Sorry for the late review, but better late than never :)

LGTM overall, just two minor points from my side:

You should be able to fix the errors in the bionic CI by adding `linux-bionic-gcc` to `skip_on_jobname` in the `test.json`.

mesonbuild/dependencies/data/CMakeListsLLVM.txt
rilian-la-te: > You should be able to fix the errors in the bionic CI by adding `linux-bionic-gcc` to `skip_on_jobname` in the `test.json`.

@mensinda I tried, but then some tests with LLVM static linking shows UNEXPPASS on Bionic.

rilian-la-te: @mensinda indentation done.

eli-schwartz:
I think we're currently unable to mark a single cell in a matrix as skip_on_jobname. Only an entire row or column.

mesonbuild/dependencies/data/CMakeListsLLVM.txt
mensinda: You are right, the current matrix implementation does not support this specific use-case...

It looks like the best solution would be to add (or even better move) the `skip_on_*` fields section. Ideally the `skip_on_*` would work similar to the [`exclude`](https://github.com/mesonbuild/meson/blob/master/run_project_tests.py#L880-L896) code, where single cells in the matrix can be excluded. See https://mesonbuild.com/Contributing.html#configuring-project-tests

Format proposal for skipping:

```json
{
  "env": {
    "VAR": "VAL"
  },
  "installed": [
    { "type": "exe", "file": "usr/bin/testexe" },
    { "type": "pdb", "file": "usr/bin/testexe" },
    { "type": "shared_lib", "file": "usr/lib/z", "version": "1.2.3" },
  ],
  "matrix": {
    "options": {
      "opt1": [
        { "val": "abc"   },
        { "val": "qwert" },
        { "val": "bad"   }
      ],
      "opt2": [
        { "val": null    },
        { "val": "true"  },
        { "val": "false" },
      ]
    },
    "skip": [
      {
        "options": { "opt1": "qwert", "opt2": "false" },
        "skip_on_jobname": "foo"
      },
      {
        "options": { "opt1": "bad" },
        "skip_on_env": "VAR1",
        "skip_on_os": "unix"
      }
    ]
  },
  "tools": {
    "cmake": ">=3.11"
  }
}
```

rilian-la-te:
@mensinda can it be upstreamed before CI fix will landed? 

Most simple fix - just install llvm 7 or 8 on Bionic instead of 6. I tried, but was unsuccessful in generating a new image.

kanavin:
Ubuntu Bionic (18.04) goes out of support in about two months. I would just drop testing with it altogether, or replace with a more recent release.

mensinda:
Since bionic is EOL soon anyway, I agree with @kanavin that we could just skip this test on bionic (if installing llvm-7 does not work out).

I have tested it locally with the docker image and this should work:

```diff
diff --git a/test cases/frameworks/15 llvm/test.json b/test cases/frameworks/15 llvm/test.json
index b39844d73..e88419e84 100644
--- a/test cases/frameworks/15 llvm/test.json
+++ b/test cases/frameworks/15 llvm/test.json
@@ -12,5 +12,8 @@
       ]
     }
   },
-  "skip_on_jobname": ["azure", "cygwin"]
+  "skip_on_jobname": ["azure", "cygwin"],
+  "tools": {
+    "llvm-config": ">=7.0"
+  }
 }
```

rilian-la-te:
@mensinda it seems than installing llvm-7 works. I think it can be merged after CI passes.

rilian-la-te:
@mensinda do I need a separate PR for updating a CI image?

mensinda:
Looking at the history, the sheduled CI image builder has been failing for some time now: https://github.com/mesonbuild/meson/actions/runs/3977183901

Ideally a separate PR would be better to keep things separated.

rilian-la-te:
@mensinda is Meson capable to generate CMake toolchain file from cross file?

rilian-la-te:
@mensinda @eli-schwartz CI is passed. Can it finally be merged?

mensinda:
Yes, LGTM

eli-schwartz:
> Looking at the history, the sheduled CI image builder has been failing for some time now: https://github.com/mesonbuild/meson/actions/runs/3977183901

The scheduled CI builder fails pretty often, but not always on the same images. Each image is updated individually, if that one image passes then that one image gets updated and uploaded.

Aside: I actually semi-periodically look at these and try to kick them to get them moving again, if there are recurring problems that cannot just be solved by waiting for the distro to fix their nonsense -- most recently 921ddb1980e1a8d08f8706a1556896793dafccbe which was an actual incompatibility with upstream libpcap.

The Ubuntu bionic image passes in the run that you linked, though.

Ubuntu *rolling* on the other hand is dumb because Debian has fixed the currently broken LLVM but Ubuntu is effectively unmaintained, they auto-sync packages from Debian unstable and then bake those, unchanging, into LTS releases no matter how broken -- and the universe repository is explicitly lacking in official Canonical Ltd. support. I think the libpcap issue was cropping up there, though. :)

eli-schwartz:
> Ubuntu Bionic (18.04) goes out of support in about two months. I would just drop testing with it altogether, or replace with a more recent release.

You wish. :P The world will be stuck with Bionic until 2028. That being said, I'm not attached to using it for our "fairly old systems" CI, but I think we're currently using it to test that Meson supports Python 3.7.

kanavin:
> > Ubuntu Bionic (18.04) goes out of support in about two months. I would just drop testing with it altogether, or replace with a more recent release.
> 
> You wish. :P The world will be stuck with Bionic until 2028. That being said, I'm not attached to using it for our "fairly old systems" CI, but I think we're currently using it to test that Meson supports Python 3.7.

'Security updates until 2028' is only available to ubuntu customers with a commercial support contract. General support  will end in April 2023. Even if your CI would somehow move onto a version of Ubuntu under such a contract, it would not be reasonable to ask contributors to fix issues on a effectively proprietary system that is not available to them.

Python 3.7 itself goes EOL in June this year FWIW.

eli-schwartz:
Oh we absolutely won't attempt to support security updates, but it's not like those should change anything. And you absolutely will still be able to download the final public release of Bionic as well as the final release of all packages in their archives, and people will keep using it even though it is unsupported. Since Github no longer supports those as hosted images for Actions, we are even using it ourselves via that same "download it yourself" approach.

But as I said, I'm not strongly motivated one way or the other. I'm not going to personally switch the CI over, at least.

> Python 3.7 itself goes EOL in June this year FWIW.

I know. As early as January of *last* year, I was pondering about the timeframe for removing 3.7 support for Meson -- but I've mostly figured there is no real need. We have people who are unhappy in 2023 that we removed 3.6 support shortly after it was EOL 13 months ago, because "it is still in use by really old Linux distros" but we aren't overly concerned about that because we had good reason for dropping 3.6. There were a bunch of useful features in the 3.7 release that we made heavy advantage of. But 3.8 just feels less compelling... I am very against dropping support for any version of python, even the EOL ones, if we don't actually need to / take advantage of newer functionality.

PR #10598, opened 2022-07-13
Improvements to logging commands during compiler detection

eli-schwartz:
meson-log.txt should contain information on what commands were run, exactly, to get some classes of error.

/cc @rgommers @mkoeppe As promised.

Inspired by https://github.com/scipy/scipy/issues/16536

The following error message:

```
C compiler for the host machine: ccache cc (clang 13.1.6 "Apple clang version 13.1.6 (clang-1316.0.21.2.5)")
C linker for the host machine: cc ld64 764

../../meson.build:1:0: ERROR: Unable to get clang pre-processor defines:
error: invalid argument '-std=gnu++11' not allowed with 'C'


A full log can be found at /Users/mkoeppe/s/sage/scipy/.mesonpy-029irg41/build/meson-logs/meson-log.txt
```

Now becomes:

```
C compiler for the host machine: clang (clang 13.0.1 "clang version 13.0.1")
C linker for the host machine: clang ld.bfd 2.36.1

meson.build:1:0: ERROR: Unable to get clang pre-processor defines:
Compiler stdout:

-----
Compiler stderr:
error: invalid argument '-std=c++11' not allowed with 'C'

-----


A full log can be found at /tmp/testproj/builddir/meson-logs/meson-log.txt
```

This is mostly the same thing we already printed to the console, but now we say that it's "Compiler stderr:".
The log file also records the actual command that was run, which is key (because we will tell people to go look in that file for more info).

We do the same thing for --version checks when trying to detect what vendor/version the compiler is, though that only logs to the logfile, not the console.

### DIFF
--- a/mesonbuild/compilers/detect.py
+++ b/mesonbuild/compilers/detect.py
@@ -417,7 +417,12 @@ def _detect_c_or_cpp_compiler(env: 'Environment', lang: str, for_machine: Machin
 
         cmd = compiler + [arg]
         try:
+            mlog.debug('-----')
+            mlog.debug(f'Detecting compiler via: {join_args(cmd)}')
             p, out, err = Popen_safe(cmd)
+            mlog.debug(f'compiler returned {p}')
+            mlog.debug(f'compiler stdout:\n{out}')
+            mlog.debug(f'compiler stderr:\n{err}')
         except OSError as e:
             popen_exceptions[join_args(cmd)] = e
             continue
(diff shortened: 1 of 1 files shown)


eli-schwartz:
The commit message is a bit clearer, I've updated the PR description a bit. Rewording suggestions are always welcome either way.

dcbaker (approved):

Issue #10584, opened 2022-07-09
`meson help foo` KeyError


baedert:
**Describe the bug**
`meson help <arg>` needs an existing command as `<arg>`. Passing anything else leads to a `KeyError` instead of a nice error message.

**To Reproduce**
```
~ ❯ meson --version
0.62.2
~ ❯ meson help foo
Traceback (most recent call last):
  File "/usr/lib/python3.10/site-packages/mesonbuild/mesonmain.py", line 153, in run
    return options.run_func(options)
  File "/usr/lib/python3.10/site-packages/mesonbuild/mesonmain.py", line 118, in run_help_command
    self.commands[options.command].print_help()
KeyError: 'foo'

ERROR: Unhandled python exception

    This is a Meson bug and should be reported!

```

**Expected behavior**
Some error message like "unknown command 'foo'" would be better, and/or list the available valid `<arg>` values.

theyashl:
Hi, I want to work on this issue.
I have the modified code ready with me in which if unknown command is passed to help module, meson will just print the basic meson usage help. If any other approach is needed, I'm ready to modify it.

dcbaker:
The best way to get code review is to open a merge request :)

theyashl:
Just opened a MR! Hope it does the expected work. Thank you! :)

PR #10600, opened 2022-07-13
Resolve KeyError while executing help command with unknown param

theyashl:
Fixed the code for reported bug raised in issues (#10584 ).
Now code will not display any python error for help command with unknown param, but instead it will show basic meson help module.

### DIFF
--- a/mesonbuild/mesonmain.py
+++ b/mesonbuild/mesonmain.py
@@ -61,8 +61,6 @@ class CommandLineParser:
                          help_msg='Wrap tools')
         self.add_command('subprojects', msubprojects.add_arguments, msubprojects.run,
                          help_msg='Manage subprojects')
-        self.add_command('help', self.add_help_arguments, self.run_help_command,
-                         help_msg='Print help of a subcommand')
         self.add_command('rewrite', lambda parser: rewriter.add_arguments(parser, self.formatter), rewriter.run,
                          help_msg='Modify the project definition')
         self.add_command('compile', mcompile.add_arguments, mcompile.run,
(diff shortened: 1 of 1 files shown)


mesonbuild/mesonmain.py
```diff
@@ -112,8 +112,9 @@ def add_help_arguments(self, parser):
         parser.add_argument('command', nargs='?')
 
     def run_help_command(self, options):
-        if options.command:
-            self.commands[options.command].print_help()
+        command = self.commands.get(options.command)
+        if command:
+            command.print_help()
```
eli-schwartz, line 117: This means that if someone runs `meson help gargle` meson will simply return the help for `meson help` because it treats it identical to no arguments.

I'm not sure that's optimal, e.g. if you pass an invalid option to a different subcommand it returns an error.

Maybe instead we could build a new set of command line arguments, which passes the command (if anything) followed by `--help` into `self.parser.parse_args()`?

theyashl: Yes ofcourse, we can add choices option to display available commands.
I'll refactor the code for the same. Thanks.

mesonbuild/mesonmain.py
```diff
@@ -109,7 +110,7 @@ def run_runpython_command(self, options):
         return 0
 
     def add_help_arguments(self, parser):
-        parser.add_argument('command', nargs='?')
+        parser.add_argument('command', nargs='?', choices=self.commands.keys())
```
eli-schwartz, line 113: This is a dict_keys object by default, which means that when the dict is updated this reference will reflect that. So I think we need this too:

```suggestion
        parser.add_argument('command', nargs='?', choices=list(self.commands.keys()))
```

eli-schwartz (changes requested):
Can you please squash the 3 commits into one? You can do this with `git rebase -i origin/master` and follow the instructions to change the second/third commits from "pick" to "fixup" / "squash".

mesonbuild/mesonmain.py
theyashl: Yeah right! We should leave out Hidden commands. I'll commit the suggestion.

eli-schwartz:
Thanks!

PR #10602, opened 2022-07-14
wrap: Fix diff_files not finding the patch to apply

nirbheek:
The path to the patch must be provided relative to the working directory used to run the command: `patch` or `git`.

The feature doesn't work when using `meson subprojects update` without this fix.


xclaesse:
Don't we have a unit test for this feature?

nirbheek:
> Don't we have a unit test for this feature?

It works in the test because `wrap.filesdir` is an absolute path, probably because `interpreter.py` passes it an absolute path to the source directory. If you use `meson subprojects update`, the source directory is relative, and things fail.

Fixing merge request description...

eli-schwartz:
Can we fix the commit description too? :)

nirbheek:
Done.

PR #10603, opened 2022-07-14
Do a better job of checking module imports

dcbaker:
There are aseveral problems with the way we handle module imports at the moment:
1) Each moduie needs an appropriate FeatureNew and/or FeatureDeprecated, duplicating a good deal of information
2) We handle stability of modules based on name, which means stabalization requires a `git mv`, and we have no way to track *when* the module was stabalized
3) by design unstable modules are supposed to be imported as `unstable-module`, however, we have an implementation leak that allows them to be imported as `unstable_module`.
4) because of the `unstable_module.py` situation, we have to try to import a module twice, once for `unstable_module`, and once for `module`.

This series does away with all of that by encoding module information in the the `ExtensionModule` object, using a small dataclass. The `Interpreter` uses that information at import time to emit proper warnings and errors, including the information about the stability of the module. This means we can have proper warnings about modules being stable or unstable, when they were stabilizied, as well as deprecation and newness.

As an added bonus we now properly emit warnings about stability for modules imported as `unstable_module`, since the check in the interpreter specifically looked for `unstable-`. And finally, all modules are just `module.py`, the module just adds an additional field to the `ModuleInfo` class declaring that it's unstable, or when it was stabilized.


mesonbuild/interpreter/interpreter.py
```diff
@@ -603,35 +623,25 @@ def _import_module(self, modname: str, required: bool, node: mparser.BaseNode) -
...
+            FeatureDeprecated.single_use(
+                f'module {ext_module.INFO.name} has been stabilized',
+                ext_module.INFO.stabilized, self.subproject,
+                'drop "unstable-" prefix from the module name',
+                location=node)
+            FeatureNew.single_use(
+                f'module {ext_module.INFO.name} as stable module',
+                ext_module.INFO.stabilized, self.subproject,
+                f'Consider either adding "unstable-" to the module name, or updating the meson required version to ">= {ext_module.INFO.stabilized}"',
+                location=node)
```
eli-schwartz, line 641: This needs to be if/else expect_unstable?

mesonbuild/interpreter/interpreter.py
```diff
@@ -604,6 +604,9 @@ def func_import(self, node: mparser.BaseNode, args: T.Tuple[str],
         # Some tests use "unstable_" instead of "unstable-", and that happens to work because
         # of implementation details
         if modname.startswith('unstable'):
+            if modname.startswith('unstable_'):
+                FeatureDeprecated.single_use('Importing unstable modules as "unstable_*"', '0.64.0', self.subproject,
+                                             'use "unstable-" instead.', node)
```
eli-schwartz, line 609: This is not a Feature, just emit 

```
mlog.deprecation('Importing unstable modules as "unstable_*" was never supposed to work, '
                 'use "unstable-*" instead. This will be removed in a future Meson release.')
```

eli-schwartz (changes requested):

mesonbuild/interpreter/interpreter.py
dcbaker: These both need to be triggered in the stablized case, the first applies when the targetted meson is 0.60, the module was stabilized in 0.62, but it's being imported as `import('module')` instead of `import('unstable-module')`, since that would fail with 0.60.

The second is needed when your require 0.62, but are still importing as `import('unstable-module')`.

Do either of those need to be  triggered in additonal cases?

eli-schwartz: expect_unstable is not the same as stabilized though.

The current version afaict will print a warning, if:
- you use `import('unstable-module')`, that you should "add unstable-", and
- you use `import('module'), that it's been stabilized and you should drop the "unstable-"

This would be fixed by only printing the featurenew if stabilized ***and*** not expect_unstable, and only printing the featuredeprecated if stabilized ***and*** expect_unstable.

mesonbuild/modules/hotdoc.py
```diff
@@ -38,6 +38,9 @@ def ensure_list(value):
 
 
 class HotdocTargetBuilder:
+
+    INFO = ModuleInfo('hotdoc', '0.48.0')
```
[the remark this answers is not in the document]
eli-schwartz: There's still some whitespace independently added here.

mesonbuild/interpreter/interpreter.py
dcbaker: Right, I've fixed this in the most recent version.

mesonbuild/interpreter/interpreter.py
```diff
@@ -614,20 +598,54 @@ def func_import(self, node: mparser.BaseNode, args: T.Tuple[str],
...
-            plainname = modname.split('-', 1)[1]
-            try:
-                # check if stable module exists
-                mod = self._import_module(plainname, required)
-                # XXX: this is actually not helpful, since it doesn't do a version check
-                mlog.warning(f'Module {modname} is now stable, please use the {plainname} module instead.')
-                return mod
-            except InvalidArguments:
-                mlog.warning(f'Module {modname} has no backwards or forwards compatibility and might not exist in future releases.', location=node)
-                modname = 'unstable_' + plainname
```
eli-schwartz, line 629: After this PR, the warning references "simd" instead of "unstable-simd", which in turn means that a test regex no longer matches.

eli-schwartz (changes requested):
Pondering this a bit more, I think we may now be allowing `import('simd')`, which formerly did not work.

mesonbuild/interpreter/interpreter.py
dcbaker: Yup, I added a test and that does happen. I've changed the behavior to allow if required is false (but you get a notfound module), or to error if required and added a test.

eli-schwartz: In the not-found case we still print a warning about its stability?

dcbaker: Yup, fixed that too.

xclaesse:
I like this change, +1.

mesonbuild/interpreter/interpreter.py
```diff
@@ -617,6 +617,9 @@ def func_import(self, node: mparser.BaseNode, args: T.Tuple[str],
         # Some tests use "unstable_" instead of "unstable-", and that happens to work because
         # of implementation details
         if modname.startswith('unstable'):
+            if modname.startswith('unstable_'):
+                mlog.deprecation(f'Importing unstable modules as "{modname}" instead of "{modname.replace("_", "-", 1)}"',
+                                 location=node)
             modname = modname[len('unstable') + 1:]  # + 1 to handle the - or _
             expect_unstable = True
```
dnicolodi, line 624: Doesn't this match anything that starts with `unstable`, including `unstable` and `unstableee`?

dcbaker: Yeah, it would, though we don't have any such modules

dnicolodi: Sure, but because of the name mangling (removal of the prefix) wouldn't the error message be a bit surprising?

eli-schwartz:
```python
if modname.replace('_', '-', 1).startswith('unstable-'):
    if modname.startswith('unstable_'):
        ...
    ...
```

dcbaker: I've reworked it in the latest version.

mesonbuild/interpreter/interpreter.py
```diff
@@ -616,7 +616,10 @@ def func_import(self, node: mparser.BaseNode, args: T.Tuple[str],
         expect_unstable = False
         # Some tests use "unstable_" instead of "unstable-", and that happens to work because
         # of implementation details
-        if modname.startswith('unstable'):
+        if modname.startswith(('unstable_', 'unstable-')):
```
eli-schwartz, line 619: This hunk is in the wrong commit. It belongs in "move handling of module stability to interpreter" but it's in "deprecate the ability import unstable modules as `unstable_*`"

It's correct to check both "unstable-" and "unstable_" in the original implementation, rather than switching to that later on from "unstable".

mesonbuild/interpreter/interpreter.py
```diff
@@ -627,20 +611,61 @@ def func_import(self, node: mparser.BaseNode, args: T.Tuple[str],
...
+            else:
+                FeatureNew.single_use(
+                    f'module {ext_module.INFO.name} as stable module',
+                    ext_module.INFO.stabilized, self.subproject,
+                    f'Consider either adding "unstable-" to the module name, or updating the meson required version to ">= {ext_module.INFO.stabilized}"',
+                    location=node)
+        elif ext_module.INFO.unstable:
+            if not expect_unstable:
+                if required:
+                    raise InvalidArguments(f'Module "{ext_module.INFO.name}" has not been stabalized, and must be imported as unstable-{ext_module.INFO.name}')
```
eli-schwartz, line 662: typo:
```suggestion
                    raise InvalidArguments(f'Module "{ext_module.INFO.name}" has not been stabilized, and must be imported as unstable-{ext_module.INFO.name}')
```

eli-schwartz (changes requested):
Typo nits in commit messages:

> interpreter: move handling of module stability to interpreter

"allo modules"

> move handling of module stability to interpreter

"stabalized"

PR #10604, opened 2022-07-15
Only reconfigure if configure options actually changed

DaanDeMeyer:
Currently, if we run "meson configure -Doption=value", meson will
do a reconfigure when running "ninja build" afterwards, even if
the new value is the same one that was already configured previously.

To avoid this unnecessary reconfigure, we can keep track of whether
any option values actually changed when running configure, and only
save the conf file if any options actually changed. If none changed,
we don't save the file.

### DIFF
--- a/mesonbuild/coredata.py
+++ b/mesonbuild/coredata.py
@@ -97,8 +97,11 @@ class UserOption(T.Generic[_T], HoldableObject):
     def validate_value(self, value: T.Any) -> _T:
         raise RuntimeError('Derived option class did not override validate_value.')
 
-    def set_value(self, newvalue: T.Any) -> None:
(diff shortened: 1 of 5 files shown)


DaanDeMeyer:
Another approach is to check if the new and old file are equal in coredata's save() method and not replacing the old file if the new one is the same. Not sure which one would be preferred.

DaanDeMeyer:
Closing in favor of #10605 with the second approach

eli-schwartz:
> Closing in favor of #10605 with the second approach

Wrong ticket number.

Also you could have just updated this PR...

eli-schwartz:
Due to reimplementing the PR on another branch, the unit test directory got forgotten (never checked into git) and isn't present in the new PR.

PR #10606, opened 2022-07-15
Only reconfigure if configure options actually changed

DaanDeMeyer:
Currently, if we run "meson configure -Doption=value", meson will
do a reconfigure when running "ninja build" afterwards, even if
the new value is the same one that was already configured previously.

To avoid this unnecessary reconfigure, let's use replace_if_different()
instead of unconditionally replacing the conf file in coredata's save()
function.


DaanDeMeyer:
Ping @eli-schwartz @tristan957

tristan957:
Looks good to me, but I am not a maintainer

PR #10607, opened 2022-07-17
Making $<TARGET_FILE:some_target> work for non imported targets

Volker-Weissmann:
This PR makes that the generator expression $<TARGET_FILE:some_target> gets correctly translated to /path/to/some/target, even if some_target is not imported, i.e. the cmake command `add_executable` was called without the IMPORTED argument...

Note: [This other PR from me](https://github.com/mesonbuild/meson/pull/10545) made an error message nicer, and this PR here makes that this error does not occur (at least in the cases I tested for).


Volker-Weissmann:
The tests failed, because what I wrote does not work when cross-compiling. I can reproduce it locally and will fix it.

mesonbuild/cmake/traceparser.py
```diff
@@ -354,16 +357,16 @@ def _cmake_add_executable(self, tline: CMakeTraceLine) -> None:
...
-        self.targets[args[0]] = CMakeTarget(args[0], 'EXECUTABLE', {}, tline=tline, imported=is_imported)
+        if 'IMPORTED' in args:
+            args.remove('IMPORTED')
+            if len(args) < 1:
+                return self._gen_exception('add_executable', 'requires at least 1 argument', tline)
+            self.targets[args[0]] = CMakeTarget(args[0], None, 'EXECUTABLE', {}, tline=tline, imported=True)
+        else:
+            if len(args) < 1:
+                return self._gen_exception('add_executable', 'requires at least 1 argument', tline)
+            build_path = self.build_dir.parent / args[0]
```
mensinda, line 368: This looks a bit too fragile to me (and is probably also the cause of the current test failures). Ideally, we could use [`Executable.get_filename()`](https://github.com/mesonbuild/meson/blob/master/mesonbuild/build.py#L1296) here, but constructing a new `Executable` looks too expensive for something that won't be needed in 90% of all cases.

The best solution I can come up with right now is to defer the construction of the `build_path` by passing a lambda instead to `CMakeTarget` and have it lazily evaluate (and cache) the result if required.

mensinda (changes requested):
I am happy with this change, except for the one comment and the red CI.

Volker-Weissmann:
There is a deeper problem:
If you are crosscompiling, you need to build all generators for the build machine, not the host machine. In my test case, I used
```
add_executable(mygen gen.c)
```
to build this generator. This is [broken, even if you use just cmake instead of cmake+meson](https://cmake.org/cmake/help/book/mastering-cmake/chapter/Cross%20Compiling%20With%20CMake.html#running-executables-built-in-the-project). Meson could try to fix this by checking which executable is used as a generator, then build those for the build machine and use them. But this would be a lot of work for something that isn't even supported by cmake itself. So I'm just going to exclude my test from the cross-compiling tests. They failed because ninja tried to execute an arm binary on an x86 machine. (Build machine was x86, host machine was arm.)

Volker-Weissmann:
My plan to fix this is as follows (and no, I don't know what I'm doing):

Make `parse_generator_expressions`  return an object of a new class, `CmakeGenExpr`. This class will have a method called `as_string`.
If TARGET_NAME or similar occurs in the generator expression this method will terminate meson with a message saying "missing feature in meson you cannot build this"
It TARGET_NAME or similar things do not occur in the generator expression, `as_string` will return what `parse_generator_expressions` returns now.

Then I put `as_string()` calls everywhere the result of `parse_generator_expressions` is currently used with the exception of `pretend_to_be_meson`.

`pretend_to_be_meson` currently returns, among other things, something like
```
custom_tgt(
    ...,
    command: ['something', 'something', 'str returned by parse_generator_expressions']
)
```
In other words, it puts a `StringNode(Token('string', some/path, 0, 0, 0, None, 'str returned by parser_generator_expressions'))` in its output.  If this StringNode would be replaced by the result of `as_meson_ast`, a new method of `CmakeGenExpr`, then the ast that pretend_to_be_meson generates could look like this
```
custom_tgt(
    ...,
    command: ['something', 'something', some_target.full_path()]
)
```

This should be a robust and extensible solution. The only cause for future problems is that the "missing feature in meson you cannot build this" might occur.

Volker-Weissmann:
The failing tests also fail on the master branch.
It's ready to merge.

Volker-Weissmann:
Will this get merged eventually?

eli-schwartz:
@mensinda ping?

mesonbuild/cmake/generator.py
```diff
@@ -67,8 +66,7 @@ def target_property(arg: str) -> str:
 
     def target_file(arg: str) -> str:
         if arg not in trace.targets:
-            mlog.warning(f"Unable to evaluate the cmake variable '$<TARGET_FILE:{arg}>'.")
-            return ''
+            return arg
```
mensinda, line 69: Technically, just returning arg here is wrong. (Since it will return `cm_exe_local` instead of the actual location of the newly built exe). This works in this case, since `add_custom_command` also accepts a target name as the first argument.

So, while returning `arg` is technically wrong, I would argue that it is "less wrong" than returning an empty string, since it works in *some* cases (but will still potentially fail for others).

However, the warning should not be removed, since we are still guessing and future issues will be easier to identify (mayebe reword it to `Unable to evaluate the...` --> `Unable to correctly evaluate the...`).

mensinda (commented):
I have only one minor issue. Once this is resolved, this PR is ready to merge (assuming the CI is green).

Sorry for the late review.

test cases/cmake/26 target file/meson.build
```diff
@@ -0,0 +1,37 @@
...
+
+# gen_src = custom_target('gen-output',
+#                         output : ['out.c'],
+#                         command : [volker_inter_exe, '@OUTPUT@'], install: true, install_dir: '.')
+
+# exe = executable('volker_exe', gen_src)
+x = 'abc' + 'def' + 'ghi'
+
+if meson.is_cross_build()
+    message('Test skipped, see https://github.com/mesonbuild/meson/pull/10607#issuecomment-1188117276 for an explanation.')
```
eli-schwartz, line 24: This should be an `error('MESON_SKIP_TEST: explanation here')` probably? The test harness will then log it as `[SKIPPED]` instead of `[SUCCESS]`.

eli-schwartz (changes requested):

mesonbuild/cmake/generator.py
Volker-Weissmann: Done

test cases/cmake/26 target file/meson.build
Volker-Weissmann: Done

mensinda:
The vs2019 test failures look relevant: https://dev.azure.com/jussi0947/jussi/_build/results?buildId=16963&view=results

Volker-Weissmann:
> The vs2019 test failures look relevant: https://dev.azure.com/jussi0947/jussi/_build/results?buildId=16963&view=results

Yes it does. 
I think we pause this PR until I find the inner strength to work on windows.

PR #10609, opened 2022-07-17
ci: try to update CI to latest packages

rilian-la-te:
For adequately testing dynamic LLVM I need to install llvm-7 on bionic and having multiple LLVM versions on other Ubuntu versions. This PR is for make sure than all images will build.


eli-schwartz:
This conflicts, and it's not clear what's actually necessary here.

PR #10612, opened 2022-07-18
qt: fix qt tools detection for Qt6.3+ by pkg-config

### LINKED ISSUE #10611: ArchLinux meson cannot detect Qt Tools with pkg-config.
**Describe the bug**
ArchLinux meson cannot detect Qt Tools with pkg-config.

**To Reproduce**
Run ./run_project_tests.py --only frameworks/4 on ArchLinux

**Expected behavior**
Qt 6 Tools should be detectable.

**system parameters**
*Native
*ArchLinux
* Python 3.10.5
* master
* ninja 1.11.0

rilian-la-te:
Fixes #10611

### DIFF
--- a/mesonbuild/dependencies/qt.py
+++ b/mesonbuild/dependencies/qt.py
@@ -433,6 +433,13 @@ class Qt5PkgConfigDependency(QtPkgConfigDependency):
 
 class Qt6PkgConfigDependency(QtPkgConfigDependency):
 
+    def __init__(self, name: str, env: 'Environment', kwargs: T.Dict[str, T.Any]):
+        super().__init__(name, env, kwargs)
+        if not self.libexecdir:
+            mlog.debug(f'detected Qt6 {self.version} pkg-config dependency does not '
+                       'have proper tools support, ignoring')
+            self.is_found = False
+
     @staticmethod
     def get_pkgconfig_host_bins(core: PkgConfigDependency) -> str:
         return core.get_pkgconfig_variable('bindir', [], None)
--- a/mesonbuild/dependencies/qt.py
+++ b/mesonbuild/dependencies/qt.py
@@ -223,9 +223,16 @@ class QtPkgConfigDependency(_QtBase, PkgConfigDependency, metaclass=abc.ABCMeta)
             if prefix:
                 self.bindir = os.path.join(prefix, 'bin')
 
+        self.libexecdir = self.get_pkgconfig_host_libexecs(self)
+
     @staticmethod
     @abc.abstractmethod
     def get_pkgconfig_host_bins(core: PkgConfigDependency) -> T.Optional[str]:
+        pass
+
+    @staticmethod
+    @abc.abstractmethod
+    def get_pkgconfig_host_libexecs(core: PkgConfigDependency) -> T.Optional[str]:
         pass
 
     @abc.abstractmethod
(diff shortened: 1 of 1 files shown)


eli-schwartz:
This really should not have anything to do with Arch Linux, it's a general transition in Qt6 which we fixed the generic way in the qmake handler and if this is relying on Arch-specific quirks then it needs to not be.

(For a long while the pkg-config detection would simply not be found, because Qt6 only recently re-added pkg-config support, sigh.)

rilian-la-te:
@eli-schwartz fixed in generic way. This is also works in debian experimental too.

mesonbuild/dependencies/qt.py
```diff
@@ -420,7 +427,8 @@ class Qt6PkgConfigDependency(QtPkgConfigDependency):
 
     @staticmethod
     def get_pkgconfig_host_bins(core: PkgConfigDependency) -> str:
-        return core.get_pkgconfig_variable('host_bins', [], None)
+        #Qt6 pkg-config for Qt defines bindir, and not host_bins
```
tristan957, line 430: Put a space after the octothorpe (`#`)

eli-schwartz:
Right so as I was concerned, this actually is not generic and does hardcode Arch Linux and I guess Debian too. It's not actually going to work correctly.

The solution for now is probably just use make...

eli-schwartz:
> But the problem is than Qt6 pkg-config do not define libexecdir, but use it in build, so, we need to guess it like we will do in qt4 with bindir, I does not figured how to do it even more general way (path can vary).

Maybe for 6.3.2 we can use this: https://codereview.qt-project.org/c/qt/qtbase/+/422416

rilian-la-te:
@eli-schwartz I created issue with a similar patch, will post link later today.

eli-schwartz:
The patch is merged to dev and cherry-picked to at least 6.3 and 6.4, so we have a good solution lined up and I'd like to use that.

Hmm, when is the next scheduled point release for Qt?

rilian-la-te:
@eli-schwartz will rewotk my PR this evening)

eli-schwartz:
Already evening, huh?

eli-schwartz:
Overall this looks good, though it would be nice to have confirmation it works.

One thing is that I'm not sure we should allow finding Qt6 via pkg-config for versions that don't have libexec defined, as the result will likely be broken in practice. We should fall back on qmake.

eli-schwartz:
Also BTW, a couple days ago Qt6 landed (masked) in ::gentoo and the path is /usr/lib64/qt6/libexec so indeed the original version of this patch would not have worked there. :)

eli-schwartz:
I added a commit that causes the pkg-config dependency to be reported as not-found, if libexecdir is not available. This should make it fall back to qmake in such a case.

thesamesam (approved):
This works for both broken Qt 6.3.1 (no libexecdir in .pc files) and patched Qt 6.3.1 (applying https://codereview.qt-project.org/c/qt/qtbase/+/422416) on Gentoo, which was broken beforehand.

It falls back correctly to qmake if lrelease can't be detected for unpatched Qt as well, so older Qt 6 series should be fine.

Thank you!

eli-schwartz:
CI is broken due to weird Ubuntu issues we're still trying to debug, plus one macOS flake which will sort itself out when rerun during merge. Merged.

PR #10616, opened 2022-07-19
Add cross-compile support for Microsoft GDK platforms

TheSpydog:
This PR adds Visual Studio backend support for cross-compiling to platforms included in the [Microsoft Game Development Kit](https://github.com/microsoft/GDK).

There are 3 flavors of GDK targets (one for PC, one for Xbox Series X|S, and one for Xbox One), but the platforms all share a common naming scheme: `Gaming.<Target>.x64`. We can catch all of these by checking for a target system whose name begins with "Gaming" and setting the platform name accordingly.

An example cross file for the Windows GDK platform would look like:

```
[target_machine]
system = 'Gaming.Desktop.x64'
cpu_family = 'x86_64'
cpu = 'x86_64'
endian = 'little'
```

### DIFF
--- a/mesonbuild/backend/vs2010backend.py
+++ b/mesonbuild/backend/vs2010backend.py
@@ -21,6 +21,7 @@ import xml.etree.ElementTree as ET
 import uuid
 import typing as T
 from pathlib import Path, PurePath
+import re
 
 from . import backends
 from .. import build
@@ -96,6 +97,8 @@ def split_o_flags_args(args):
 def generate_guid_from_path(path, path_type):
     return str(uuid.uuid5(uuid.NAMESPACE_URL, 'meson-vs-' + path_type + ':' + str(path))).upper()
 
+def detect_microsoft_gdk(platform: str) -> bool:
+    return re.match(r'Gaming\.(Desktop|Xbox.XboxOne|Xbox.Scarlett)\.x64', platform, re.IGNORECASE)
 
 class Vs2010Backend(backends.Backend):
     def __init__(self, build: T.Optional[build.Build], interpreter: T.Optional[Interpreter]):
@@ -182,7 +185,11 @@ class Vs2010Backend(backends.Backend):
         target_machine = self.interpreter.builtin['target_machine'].cpu_family_method(None, None)
         if target_machine in {'64', 'x86_64'}:
             # amd64 or x86_64
-            self.platform = 'x64'
+            target_system = self.interpreter.builtin['target_machine'].system_method(None, None)
+            if detect_microsoft_gdk(target_system):
+                self.platform = target_system
+            else:
+                self.platform = 'x64'
         elif target_machine == 'x86':
             # x86
             self.platform = 'Win32'
@@ -1321,7 +1328,7 @@ class Vs2010Backend(backends.Backend):
             targetplatform = self.platform.lower()
         if targetplatform == 'win32':
             targetmachine.text = 'MachineX86'
-        elif targetplatform == 'x64':
+        elif targetplatform == 'x64' or detect_microsoft_gdk(targetplatform):
             targetmachine.text = 'MachineX64'
         elif targetplatform == 'arm':
             targetmachine.text = 'MachineARM'


tristan957:
It would probably be better to use a regex in my opinion. It would be a fairly easy regex too.

```
Gaming\.(differences...)\.x64
```

TheSpydog:
> It would probably be better to use a regex in my opinion. It would be a fairly easy regex too.
> 
> ```
> Gaming\.(differences...)\.x64
> ```
> 
> Just a thought. No need to reserve all `Gaming.*` prefixes for the GDK.

Sure, makes sense. Just pushed a commit to explicitly check for the supported platform names.

tristan957:
@eli-schwartz thoughts?

mesonbuild/backend/vs2010backend.py
```diff
@@ -1317,7 +1322,7 @@ def gen_vcxproj(self, target, ofname, guid):
             targetplatform = self.platform.lower()
         if targetplatform == 'win32':
             targetmachine.text = 'MachineX86'
-        elif targetplatform == 'x64':
+        elif targetplatform == 'x64' or re.match(r'gaming\.(desktop|xbox.xboxone|xbox.scarlett)\.x64', targetplatform):
```
xclaesse, line 1325: You repeat that regex, would be better to factor that in a single place.

eli-schwartz:
This seems reasonable enough. I guess we should probably document it somewhere in https://mesonbuild.com/Reference-tables.html though?

mesonbuild/backend/vs2010backend.py
```diff
@@ -96,6 +97,8 @@ def split_o_flags_args(args):
 def generate_guid_from_path(path, path_type):
     return str(uuid.uuid5(uuid.NAMESPACE_URL, 'meson-vs-' + path_type + ':' + str(path))).upper()
 
+def detectGDK(platform, ignore_case):
```
tristan957, line 100: detect_Microsoft_GDK would probable be a better function name and please add types to to the function

eli-schwartz:
Can you squash these into one commit? There's only one effective change.

mesonbuild/backend/vs2010backend.py
```diff
@@ -96,6 +97,8 @@ def split_o_flags_args(args):
...
+    return re.match(r'Gaming\.(Desktop|Xbox.XboxOne|Xbox.Scarlett)\.x64', platform, re.IGNORECASE if ignore_case else 0)
```
xclaesse, line 101: nitpicking: we use underscore in method calls: `detect_gdk()`. Also, any reason for not always ignoring case?

TheSpydog: I had the ignore case toggle because I thought the GDK toolchain in Visual Studio relied on the target name matching a string exactly. Just tested and that is not the case, so I'll remove the parameter.

TheSpydog:
Squashed and applied the suggested changes. Still pondering how to handle documentation, since this is a niche target and doesn't behave like other platforms.

mesonbuild/backend/vs2010backend.py
```diff
@@ -96,6 +97,8 @@ def split_o_flags_args(args):
 def generate_guid_from_path(path, path_type):
     return str(uuid.uuid5(uuid.NAMESPACE_URL, 'meson-vs-' + path_type + ':' + str(path))).upper()
 
+def detect_Microsoft_GDK(platform: str):
```
tristan957, line 100: Add a return type please

mesonbuild/backend/vs2010backend.py
```diff
@@ -96,6 +97,8 @@ def split_o_flags_args(args):
 def generate_guid_from_path(path, path_type):
     return str(uuid.uuid5(uuid.NAMESPACE_URL, 'meson-vs-' + path_type + ':' + str(path))).upper()
 
+def detect_Microsoft_GDK(platform: str) -> bool:
```
xclaesse, line 100: Sorry for being picky, but we don't use capitals in our function names.

PR #10617, opened 2022-07-19
fix Popen_safe write

### LINKED ISSUE #10614: Potential regression: incorrect selinux context after install
**Describe the bug**
Potential regression of https://github.com/mesonbuild/meson/issues/1967
The selinux context is wrong after a ninja install. 

**To Reproduce**
```
# meson.build
project('test', 'c', version : '0')

systemd = dependency('systemd')
systemd_unit_dir = systemd.get_variable(pkgconfig: 'systemdsystemunitdir')

install_data(
  'test.service',
  install_dir: systemd_unit_dir # /usr/lib/systemd/system
)
```
```
$ touch test.service
$ meson --prefix=/usr build
# ninja -C build install
```
**Current behaviour**
```
$ ls -lZ /usr/lib/systemd/system | grep test
-rw-r--r--. 1 root root unconfined_u:object_r:user_home_t:s0                     0 Jul 19 04:38 test.service
```
As shown here, the selinux context is `unconfined_u:object_r:user_home_t:s0`.

**Expected behavior**
Ideally the installed file should have `system_u:object_r:systemd_unit_file_t:s0` (this is what is set if using `/usr/bin/install` or `restorecon -F` afterwards).
Otherwise, at least `unconfined_u:object_r:systemd_unit_file_t:s0` (using `restorecon` afterwards)

**system parameters**
Fedora 36 with default selinux settings
Python 3.10.5
Meson 0.62.2
Ninja 1.10.2
```
$ selinuxenabled
$ which restorecon
/usr/sbin/restorecon
```

zhangyuannie:
Fixes #10614

Popen object must be created with with `stdin=PIPE` so `Popen.communicate(write)` can send data to the process’s stdin.

### DIFF
--- a/mesonbuild/mesonlib/universal.py
+++ b/mesonbuild/mesonlib/universal.py
@@ -1394,20 +1394,22 @@ def partition(pred: T.Callable[[_T], object], iterable: T.Iterable[_T]) -> T.Tup
 
 
 def Popen_safe(args: T.List[str], write: T.Optional[str] = None,
+               stdin: T.Union[T.TextIO, T.BinaryIO, int] = subprocess.DEVNULL,
                stdout: T.Union[T.TextIO, T.BinaryIO, int] = subprocess.PIPE,
                stderr: T.Union[T.TextIO, T.BinaryIO, int] = subprocess.PIPE,
                **kwargs: T.Any) -> T.Tuple['subprocess.Popen[str]', str, str]:
     import locale
     encoding = locale.getpreferredencoding()
-    # Redirect stdin to DEVNULL otherwise the command run by us here might mess
+    # Stdin defaults to DEVNULL otherwise the command run by us here might mess
     # up the console and ANSI colors will stop working on Windows.
-    if 'stdin' not in kwargs:
-        kwargs['stdin'] = subprocess.DEVNULL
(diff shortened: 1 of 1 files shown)


zhangyuannie:
Hi, just a friendly bump. Is there anything that is blocking this bug fix?

**What does this PR do?**
`Popen` object are currently created with `stdin` set to `DEVNULL`. This PR make it so that if the `write` parameter is provided, then `stdin` will default to `PIPE` instead. `stdin` is still set to `DEVNULL` if `write` is not present.

`stdin` must be set to `PIPE` so that `Popen.communicate` can pass the data to the process.

This shouldn't affect anything else since `write` is only used by `restorecon` and `compilers/detect.py`. `restorecon`'s bug will be fixed by this PR and `compilers/detect.py` has stdin manually set to `PIPE` already so nothing will change there.

**How this change fix the bug with `restorecon`?**
Currently, the files that restorecon should touch is passed via stdin. But as  `stdin` is set to `DEVNULL`, the files are not sent to restorecon.

**What commit introduced this regression?**
7e8d5207a7c662beeae41b375e0af5c96f57e7db where raw `subprocess.Popen` is replaced by `Popen_safe`.

PR #10618, opened 2022-07-19
Use collapsable markdown

jpr42:
Helps readability, and avoids scrolling for users who will most likely only care about their target platform.

### DIFF
--- a/docs/markdown/SimpleStart.md
+++ b/docs/markdown/SimpleStart.md
@@ -22,7 +22,8 @@ There are three phases to getting a development environment running.
 
 ## Installing a compiler toolchain
 
-### Linux
+<details>
+ <summary>Linux</summary>
 
 All Linux distributions provide easy access to development tools.
 Typically you need to open a terminal and execute one command, which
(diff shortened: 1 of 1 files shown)


eli-schwartz:
Have you verified that this works when built by hotdoc?

(I haven't checked, I just know that we're building the HTML ourselves, so it isn't guaranteed that whatever github.com supports is also supported by our documentation builder.)

jpr42:
> Have you verified that this works when built by hotdoc?

No. I wasn't aware how the website was being generated. Are there instructions on how to do this locally? I didn't see anything in the [contributing guide regarding the docs](https://mesonbuild.com/Contributing.html)

eli-schwartz:
https://github.com/mesonbuild/meson/tree/master/docs#building-the-documentation

eli-schwartz:
I built it locally. It works, e.g.

```html
<h2 id="installing-a-compiler-toolchain">Installing a compiler toolchain</h2>
<details>
 <summary>Linux</summary>
<p>All Linux distributions provide easy access to development tools.
Typically you need to open a terminal and execute one command, which
depends on your distro.</p>
<ul>
<li>Debian, Ubuntu and derivatives: <code>sudo apt install build-essential</code>
</li>
<li>Fedora, Centos, RHEL and derivatives: <code>sudo dnf install gcc-c++</code>
</li>
<li>Arch: <code>sudo pacman -S gcc</code>
</li>
</ul>
</details>
```

But there's no indication that it is clickable.

PR #10619, opened 2022-07-20
ar linker: detect the "osx ld" case (thin archives not OK) based on the host OS

UncombedCoconut:
Hi! I just got to the bottom of a regression cross-compiling a project for macOS (using the Linux-based "osxcross" toolchain).
It bisected to 0a3a9fa0c3ebf45c94d9009a59cead571cbecf7b.
The Meson behavior change that killed us (in case of LTO builds) was invoking llvm-ar with the options "csrDT" instead of "csrD"... which had been thoughtfully prevented already, but based on a condition that unfortunately doesn't work for cross builds.

Please feel free to rewrite the patch if it's an awkward fit.
The heart of the matter is replacing a `mesonlib.is_osx()` check in linkers.py.
My real-life test case is pretty horrible in terms of download size and time to test, but I can provide details if helpful.
I've tried it with the below patch successfully, and I can retest if further changes are needed.

Thanks for your time.

### DIFF
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -2862,7 +2862,7 @@ https://gcc.gnu.org/bugzilla/show_bug.cgi?id=47485'''))
             if target.import_filename:
                 commands += linker.gen_import_library_args(self.get_import_filename(target))
         elif isinstance(target, build.StaticLibrary):
-            commands += linker.get_std_link_args(not target.should_install())
+            commands += linker.get_std_link_args(self.environment, not target.should_install())
         else:
             raise RuntimeError('Unknown build target type.')
         return commands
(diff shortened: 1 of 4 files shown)


dcbaker (approved):
This looks correct to me, thanks for getting to the bottom of this!

UncombedCoconut:
Oops, pylint caught me making an unwanted substitution.

eli-schwartz:
> Oops, pylint caught me making an unwanted substitution.

The changes should really be squashed together. :)

UncombedCoconut:
> > Oops, pylint caught me making an unwanted substitution.
> 
> The changes should really be squashed together. :)

If you're OK with the churn it will create, I'm happy to `git rebase -i` and force-push.

eli-schwartz:
We do prefer it, yes. It reduces the churn in git log, and we don't really care about CI churn. :)

dcbaker:
@nirbheek , this is probably stable branch material

PR #10622, opened 2022-07-21
Add static analyzer support for gnu.

AmitShukla87622:


docs/markdown/Configuring-a-build-directory.md
```diff
@@ -55,6 +55,7 @@ a sample output for a simple project.
       b_pgo       off           [off, generate, use]                                          Use profile guided optimization
       b_sanitize  none          [none, address, thread, undefined, leak, memory, address,undefined] Code sanitizer to use
       b_staticpic true          [true, false]                                                 Build static libraries as position independent
+      b_sanalyzer false         [true, false]                                                 Build time static code analyzer
```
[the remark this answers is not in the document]
AmitShukla87622: > Write out static. No need to abbreviate
Done as part of next commit. Thanks.

AmitShukla87622:
Ping, for review or feedback for the patch.

eli-schwartz:
As far as I can tell, this adds an option but doesn't actually use it anywhere?

Additionally, as a general commit policy:

Can you please squash changes from review feedback into the commit that contains the original logical change? You can update the most recent commit with `git commit --amend` instead of making a new commit, or squash two existing commits with `git rebase -i origin/master` and reordering the list of commits, following the instructions in the editor to change the word "pick" to "squash".

AmitShukla87622:
Thanks @eli-schwartz for code review suggestions. I have incorporated your suggestions as part of latest commit.

mesonbuild/compilers/compilers.py
```diff
@@ -1275,6 +1281,13 @@ def get_no_warn_args(self) -> T.List[str]:
     def needs_static_linker(self) -> bool:
         raise NotImplementedError(f'There is no static linker for {self.language}')
 
+    def analyzer_compile_args(self) -> T.List[str]:
+        if self.get_id() == 'gcc':
+            return ['-fanalyzer']
```
eli-schwartz, line 1286: The base compilers class should return `[]` and the method should be overridden in the gnu and clang mixins, rather than doing ID checks in the base class.

eli-schwartz (changes requested):
I would still like these commits to be squashed... ;)

eli-schwartz:
https://mesonbuild.com/Contributing.html#special-procedure-for-new-features

It would also be good to have tests for this, and it needs a release notes snippet. Since this isn't a feature of the meson.build DSL, only a user option, it doesn't need a FeatureCheck.

mesonbuild/compilers/compilers.py
AmitShukla87622: Done the changes as part of latest commit. Thanks.

AmitShukla87622:
> I would still like these commits to be squashed... ;)

My bad. Have squashed all commits. Thanks.

AmitShukla87622:
Thanks @eli-schwartz for review suggestions. I have made the suggested changes as part of latest commit.

AmitShukla87622:
> https://mesonbuild.com/Contributing.html#special-procedure-for-new-features
> 
> It would also be good to have tests for this, and it needs a release notes snippet. Since this isn't a feature of the meson.build DSL, only a user option, it doesn't need a FeatureCheck.

Done, added a test case and release notes snippet. Thanks.

mesonbuild/compilers/mixins/gnu.py
```diff
@@ -397,3 +397,6 @@ def use_linker_args(cls, linker: str, version: str) -> T.List[str]:
         if linker == 'mold' and mesonlib.version_compare(version, '>=12.0.1'):
             return ['-fuse-ld=mold']
         return super().use_linker_args(linker, version)
+
+    def analyzer_compile_args(self) -> T.List[str]:
```
tristan957, line 401: Seems like you need to protect this with a version check of the compiler. Same thing for clang. Otherwise this looks like it would always return the argument as supported even if it is not.

docs/markdown/snippets/static-analyzer.md
```diff
@@ -0,0 +1,5 @@
+## Added support for static analyzer
+
+User can use gcc(>= 10.0.0) and clang(>= 9.0.1) compiler, static analyzer feature with
```
tristan957, line 3: The static analyzer feature is available as `-Db_static_analyzer=true/false` on `gcc >= 10.0.0` and `clang >= 9.0.1`. In the event that `b_static_analyzer` is `true`, but the compiler doesn't support the flag, the value of `b_static_analyzer` is ignored.

unittests/linuxliketests.py
```diff
@@ -1793,3 +1793,25 @@ def test_install_strip(self):
         self._run(install_cmd + ['--strip'], workdir=self.builddir)
         stdout = self._run(['file', '-b', lib])
         self.assertNotIn('not stripped', stdout)
+
+    @skip_if_not_base_option('b_static_analyzer')
```
tristan957, line 1797: Does this belong in a different file (not `linuxliketests.py`). I'll let someone smarter than me about test organization comment on that.

Issue #10443, opened 2022-05-28
test cases/frameworks/4 qt/meson.build:48:4: ERROR: Unhandled python exception


barracuda156:
I got the following error when running test suite:
```
test cases/frameworks/4 qt/meson.build:48:4: ERROR: Unhandled python exception
with This is a Meson bug and should be reported!
```
Environment:

10.6 PPC (10A190)
Xcode 3.2
`meson` 0.62.1
`ninja` 1.11.0

Error log:
```
test_qtdependency_pkgconfig_detection (unittests.linuxliketests.LinuxlikeTests)
Test that qt4 and qt5 detection with pkgconfig works. ... ERROR

Stdout:
The Meson build system
Version: 0.62.1
Source dir: /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_devel_meson/meson/work/meson-0.62.1/test cases/common/12 data
Build dir: /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_devel_meson/meson/work/meson-0.62.1/tmp0u6mhmyg
Build type: native build
Project name: data install test
Project version: undefined
C compiler for the host machine: /usr/bin/gcc-4.2 (gcc 4.2.1 "powerpc-apple-darwin10-gcc-4.2.1 (GCC) 4.2.1 (Apple Inc. build 5621) (dot 3)")
C linker for the host machine: /usr/bin/gcc-4.2 ld64 87.3.1
Host machine cpu family: ppc
Host machine cpu: ppc
Build targets in project: 0

data install test undefined

  User defined options
    backend: ninja
    libdir : lib
    prefix : /usr

Found ninja-1.11.0 at /opt/local/bin/ninja

ninja explain: output PHONY of phony edge with no inputs doesn't exist
ninja explain: PHONY is dirty
ninja explain: output all of phony edge with no inputs doesn't exist
ninja explain: all is dirty
ninja explain: meson-install is dirty
[0/1] Installing files.
Installing /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_devel_meson/meson/work/meson-0.62.1/test cases/common/12 data/datafile.dat to /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_devel_meson/meson/work/meson-0.62.1/tmp0u6mhmyg/install/usr/share/progname
[...]

PR #10623, opened 2022-07-22
qt module: correctly compare program name for lrelease version

eli-schwartz:
(b, name) will always have the canonical tool name, not just a potentially weird bin path. We need to check stderr for qt4 tools other than lrelease, but we keyed off of the tool bin path, so matches were by no means guaranteed, and specifically, would fail if the tool is looked up via bindir.

This really should always have checked the canonical name, which is guaranteed to be correct.

Partial fix for #10443


barracuda156:
@eli-schwartz Thanks!

dcbaker (approved):

PR #10628, opened 2022-07-25
call ranlib -c after static linking on macos

dcbaker:
Apple, for some reason, decided that ranlib should not, by default, include extern'd variables in the symbol table when running ar with the `-s` option (run ranlib). So we must additionally call `ranlib -c $out` when linking a static archive, so that behavior is consistant between the Apple archiver and every other archvier other (including llvm-ar).

Incidently we never caught this because we have code to specifically chose llvm-ar when using any variaty of clang if it's available, and apple's clang is a variety of clang, and we install llvm for running llvm tests. Because of that I've modifed a test for apple that forces the use of the `ar` archiver, even when `llvm-ar` or `gcc-ar` are installed, and don't have this behavior.

### DIFF
--- a/mesonbuild/backend/ninjabackend.py
+++ b/mesonbuild/backend/ninjabackend.py
@@ -42,7 +42,7 @@ from ..compilers import (
     PGICCompiler,
     VisualStudioLikeCompiler,
 )
-from ..linkers import ArLinker, RSPFileSyntax
+from ..linkers import ArLinker, AppleArLinker, RSPFileSyntax
 from ..mesonlib import (
     File, LibType, MachineChoice, MesonException, OrderedSet, PerMachine,
     ProgressBar, quote_arg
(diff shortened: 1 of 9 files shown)


eli-schwartz (approved):

test cases/osx/9 global variable ar/nativfile.ini
```diff
@@ -0,0 +1,2 @@
+[binaries]
```
jpakkane, line 1: This file shoule be called`nativefile.ini`, not `nativfile.ini`.

dcbaker:
@nirbheek this should get backported

eli-schwartz:
This broke other things, and was fixed in #10699

That in turn broke #10745 which we still don't have a fix for.

@nirbheek We should revert this backport for 0.63.2, I think.

PR #10633, opened 2022-07-27
mtest: Inline try_wait_one using asyncio.wait_for

QuLogic:
Although the former accepts multiple awaitables, it is only ever called with a single one, so just use `wait_for` instead.

Additionally, the `try_wait_one` fails in Python 3.11, as `Process.wait()` returns a coroutine, and `asyncio.wait` only accepts tasks, so it errors out.

### DIFF
--- a/mesonbuild/mtest.py
+++ b/mesonbuild/mtest.py
@@ -16,6 +16,7 @@
 
 from pathlib import Path
 from collections import deque
+from contextlib import suppress
 from copy import deepcopy
 import argparse
 import asyncio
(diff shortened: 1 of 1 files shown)


eli-schwartz:
@bonzini does this look right to you?

bonzini:
Yes, it does.

bonzini:
> Additionally, the try_wait_one fails in Python 3.11, as Process.wait() returns a coroutine, and asyncio.wait only accepts tasks, so it errors out

@QuLogic is this a documented change? If not we should report a bug.

QuLogic:
It was deprecated in 3.8, and scheduled for removal in 3.11: https://docs.python.org/3/library/asyncio-task.html#asyncio.wait

bonzini:
Ah I interpreted that as "Process.wait() returns a coroutine in 3.11 instead of another kind of future", thanks.

eli-schwartz:
Hmm, I tried to run the Meson testsuite with a warning filter that turns deprecations into errors, and I got a bunch of failures. But I never made it as far as this, I don't think.

```
[0/1] Running all tests.
  File "/home/eschwartz/git/meson/mesonbuild/mesonmain.py", line 153, in run
  File "/home/eschwartz/git/meson/mesonbuild/mtest.py", line 2049, in run
    return th.doit()
  File "/home/eschwartz/git/meson/mesonbuild/mtest.py", line 1726, in doit
    self.run_tests(runners)
  File "/home/eschwartz/git/meson/mesonbuild/mtest.py", line 1864, in run_tests
    loop = asyncio.get_event_loop()
DeprecationWarning: There is no current event loop

```

PR #10634, opened 2022-07-27
mesonbuild/mtest.py: filter more invalid googletest JUnit4 attributes

xnox:
googletest 1.12.1 generates new JUnit4 invalid attributes file and line.

Maybe all gtest "invalid" attributes are actually valid JUnit5 attributes, and maybe schema should be upgraded to JUni5.

### DIFF
--- a/mesonbuild/mtest.py
+++ b/mesonbuild/mtest.py
@@ -774,6 +774,10 @@ class JunitBuilder(TestLogger):
                     del case.attrib['result']
                 for case in suite.findall('.//testcase[@timestamp]'):
                     del case.attrib['timestamp']
+                for case in suite.findall('.//testcase[@file]'):
+                    del case.attrib['file']
+                for case in suite.findall('.//testcase[@line]'):
+                    del case.attrib['line']
                 self.root.append(suite)
             return
 


jpakkane:
This is the same as

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1015132

jpakkane:
This seems a bit suspicious. The correct fix would seem to be either to fix Google test to produce valid XML output or update our schema file to allow those entries. Deleting entries from XML like this does not seem right, especially if the file and line entries are valid and we remove them. That is a different bug report waiting to happen.

smcv:
> Deleting entries from XML like this does not seem right, especially if the file and line entries are valid and we remove them.

As far as I can see, Meson is already deleting some of the entries from the googletest output to get it to match the required format, with comment `# GTest can inject invalid attributes`: this MR is just expanding that with a few new attributes.

> The correct fix would seem to be either to fix Google test to produce valid XML output

It's Meson that is asserting in its test suite that googletest will produce XML output that conforms to a particular schema, not googletest. If googletest doesn't promise to produce XML output according to a required schema, and if Meson promises that its output *will* conform to that schema, then Meson presumably shouldn't trust googletest to do that, either by postprocessing its XML output, doing a parse-and-reserialize to take only the parts that are understood, or (as a last resort) not picking up this extra information from googletest at all.

> or update our schema file to allow those entries

If https://github.com/junit-team/junit5/blob/main/platform-tests/src/test/resources/jenkins-junit.xsd is accurate, then `@result`, `@file` and `@line` are not allowed in JUnit 5's version of this format either.

However, https://junit.org/junit5/docs/snapshot/user-guide/#advanced-topics says:

> [LegacyXmlReportGeneratingListener](https://junit.org/junit5/docs/snapshot/api/org.junit.platform.reporting/org/junit/platform/reporting/legacy/xml/LegacyXmlReportGeneratingListener.html) generates a separate XML report for each root in the [TestPlan](https://junit.org/junit5/docs/snapshot/api/org.junit.platform.launcher/org/junit/platform/launcher/TestPlan.html). Note that the generated XML format is compatible with the de facto standard for JUnit 4 based test reports that was made popular by the Ant build system.
...
> [OpenTestReportGeneratingListener](https://junit.org/junit5/docs/snapshot/api/org.junit.platform.reporting/org/junit/platform/reporting/open/xml/OpenTestReportGeneratingListener.html) writes an XML report for the entire execution in the event-based format specified by [Open Test Reporting](https://github.com/ota4j-team/open-test-reporting) which supports all features of the JUnit Platform such as hierarchical test structures, display names, tags, etc.

The format Meson is producing seems to be the "de facto standard for JUnit 4 based test reports", so perhaps Meson shouldn't be promising that it won't contain additional attributes, because it's a de facto standard rather than a formal standard? I don't really know XML Schema, but perhaps inserting some `<xs:anyAttribute/>` into the schema would be appropriate?

At the same time, producing this de facto standard seems more useful than Open Test Reporting, because it's the JUnit-reporting de facto standard that interoperates with things like [Gitlab-CI test result collection](https://docs.gitlab.com/ee/ci/yaml/artifacts_reports.html#artifactsreportsjunit), and Open Test Reporting doesn't.

PR #10635, opened 2022-07-27
Handle garbage in environment more gracefully when setting vsenv

DFOVIT:
Sometime people have garbage in their environment and it make meson crash when setting visual studio environment.
With this patch garbage in => garbage out

### DIFF
--- a/mesonbuild/mesonlib/vsenv.py
+++ b/mesonbuild/mesonlib/vsenv.py
@@ -96,8 +96,13 @@ def _setup_vsenv(force: bool) -> bool:
             continue
         if not bat_line:
             continue
-        k, v = bat_line.split('=', 1)
-        os.environ[k] = v
+        try:
+            k, v = bat_line.split('=', 1)
+        except ValueError:
+            # there is no "=", ignore junk data
+            pass
+        else:
+            os.environ[k] = v
     return True
 
 def setup_vsenv(force: bool = False) -> bool:


mesonbuild/mesonlib/vsenv.py
```diff
@@ -96,8 +96,8 @@ def _setup_vsenv(force: bool) -> bool:
             continue
         if not bat_line:
             continue
-        k, v = bat_line.split('=', 1)
-        os.environ[k] = v
+        k, *v = bat_line.split('=', 1)
+        os.environ[k] = ''.join(v)
```
eli-schwartz, line 100: It took me a bit of thinking to figure out what your problem was. Can you please update the commit message to mention that this fixes the case where there isn't a `=` and the split cannot assign to a second element?

I also think we don't need to bother with join, because the split can *only* produce a single value (k) or two values, never three or more.

Maybe we should just wrap this in a try/except.

```
try:
    k, v = bat_line.split('=', 1)
except ValueError:
    # there is no "=", ignore junk data
    pass
else:
    os.environ[k] = v
```

DFOVIT: I added a detailed description in the commit message.
I'm open to make the try/except change but i would argue that it is safer for meson not to decide what is garbage and what is not.

eli-schwartz: If it doesn't have "=" in it, does it even constitute an assignment? Seems fairly clear to me, TBH.

eli-schwartz: (I'm not even sure I understand under what circumstances garbage data can be there, but perhaps I'm missing something.)

DFOVIT: Ok, i updated the patch.

dcbaker (approved):

PR #10636, opened 2022-07-28
Improve logging of issues during linker detection

eli-schwartz:
Also, when the linker is borked and we try to detect it as Apple ld, we should not do so, so don't.

### DIFF
--- a/mesonbuild/linkers/detect.py
+++ b/mesonbuild/linkers/detect.py
@@ -186,18 +186,24 @@ def guess_nix_linker(env: 'Environment', compiler: T.List[str], comp_class: T.Ty
             v = search_version(o)
 
         linker = LLVMDynamicLinker(compiler, for_machine, comp_class.LINKER_PREFIX, override, version=v)
-    # first is for apple clang, second is for real gcc, the third is icc
+    # first might be apple clang, second is for real gcc, the third is icc
     elif e.endswith('(use -v to see invocation)\n') or 'macosx_version' in e or 'ld: unknown option:' in e:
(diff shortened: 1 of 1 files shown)


mesonbuild/linkers/detect.py
```diff
@@ -179,18 +186,23 @@ def guess_nix_linker(env: 'Environment', compiler: T.List[str], comp_class: T.Ty
...
             if 'PROJECT:ld' in line:
                 v = line.split('-')[1]
                 break
         else:
-            v = 'unknown version'
+            # did we get here because a generally broken ld?
+            if not 'macosx_version' in e and not 'ld: unknown option:' in e:
+                __failed_to_detect_linker(compiler, check_args, o, e)
+            else:
+                v = 'unknown version'
```
eli-schwartz, line 205: @dcbaker this dates back to commit 3fbb45016610deba863630b20e95f60d2d49a487 and I do not understand it. Why do we detect "unknown version" at all, actually?

dcbaker: I think I basically copied what was happening elsewhere, not anything specific

eli-schwartz: Well, I would dearly like to burn it to the ground as we hit this code path for "error occurred, use -v to see invocation. What do you think?

dcbaker: Seems fine to me to make that an error

eli-schwartz: Alright, it's an error now.

PR #10642, opened 2022-07-29
Add option to generate profile information

### LINKED ISSUE #3659: Could you please help with compile under profiler support like gpref
I create some application and want understand profile information over gprof.
Tight now I add
```
c_args = [ '-pg' ]
link_args = ['-pg']
```
As result I receive gmon.out but gprof does not provide any result. Do u know where problem may bey exists?

andy5995:
(closes #3659)

For something like this, which tests do you recommend I should run? I see several scripts for testing in the root directory....


andy5995:
I don't understand why the unit test is failing

https://github.com/mesonbuild/meson/runs/7588018175?check_suite_focus=true#step:4:1044

When I run the unit tests locally, I can manually verify the gmon.out file was created.

(P.S. I ain't so very good with python)

docs/markdown/Builtin-options.md
```diff
@@ -134,6 +134,7 @@ available on all platforms or with all compilers:
 | b_ndebug      | false          | true, false, if-release                                          | Disable asserts                                                               |
 | b_pch         | true           | true, false                                                      | Use precompiled headers                                                       |
 | b_pgo         | off            | off, generate, use                                               | Use profile guided optimization                                               |
+| b_profile     | off            | off, prof, gprof                                                 | Generate extra code to write profile information                              |
```
andy5995, line 137: Should I get rid of the 'prof' option and just make this a Boolean option (gprof|True/False)? I couldn't find 'prof' available in any Debian or Arch packages. I couldn't find docs online for it either. Any search I tried only returned results for using 'gprof'.

I only thought to add 'prof' because it was mentioned in the gcc docs:

>Generate extra code to write profile information suitable for the analysis program prof (for -p) or gprof (for -pg).

eli-schwartz: I think prof only in theory exists on proprietary systems that shipped their own, but the GNU version more or less won everywhere.

mesonbuild/compilers/mixins/gnu.py
```diff
@@ -314,6 +314,9 @@ def use_linker_args(cls, linker: str, version: str) -> T.List[str]:
                 f'Unsupported linker, only bfd, gold, and lld are supported, not {linker}.')
         return [f'-fuse-ld={linker}']
 
+    def get_gprof_args(self) -> T.List[str]:
+        return ['-pg', '-no-pie']
```
dcbaker, line 318: I'm uncomfortable putting -no-pie here. The other cases we have where options don't interact correctly (undef + asan with clang) we simply warn "this might now work correctly".

eli-schwartz:
When fixing up a commit to remove something that shouldn't have been added (like no-pie) that change needs to be squashed back into the first commit.

docs/markdown/howtox.md
```diff
@@ -263,6 +263,20 @@ $ meson compile
 
 After these steps the resulting binary is fully optimized.
 
+## Generate extra code to write profile information
```
eli-schwartz, line 266: This How-to entry doesn't actually tell users anything meaningful above and beyond the fact that the option exists. I think it should be removed altogether.

mesonbuild/compilers/mixins/gnu.py
andy5995: Ah, ok, so besides removing 'no-pie' I'll need to figure out how to add that text...

mesonbuild/compilers/compilers.py
```diff
@@ -287,6 +287,7 @@ class CompileCheckMode(enum.Enum):
     OptionKey('b_pgo'): coredata.UserComboOption('Use profile guided optimization',
                                                  ['off', 'generate', 'use'],
                                                  'off'),
+    OptionKey('b_profile'): coredata.UserBooleanOption('Generate extra code to write profile information suitable for gprof', False),
```
anarazel, line 290: The documentation above documents this as '[off, prof, gprof]' but here you define it as a boolean option.

mesonbuild/compilers/compilers.py
```diff
@@ -349,6 +350,11 @@ def get_base_compile_args(options: 'KeyedOptionDictType', compiler: 'Compiler')
             args.extend(compiler.get_profile_use_args())
     except KeyError:
         pass
+    try:
+        if options[OptionKey('b_profile')].value:
+            args += compiler.get_gprof_args()
```
anarazel, line 355: Given you documented it as an option with multiple values, shouldn't it be passed to get_gprof_args()?

andy5995: Thanks @anarazel

I made the correction in the docs. It's intended to be Boolean.

mesonbuild/compilers/mixins/gnu.py
andy5995: > I'm uncomfortable putting -no-pie here. The other cases we have where options don't interact correctly (undef + asan with clang) we simply warn "this might now work correctly".

Is that supposed to go in the docs somewhere or output by meson at some point

anarazel: Why is -pg incompatible with pie in the first place?

mesonbuild/compilers/compilers.py
andy5995: @anarazel It's not up to me. They asked me above to remove it.

andy5995:
If I fix the conflicts and rebase, might this be reviewed soon?

andy5995:
Rebased

mesonbuild/compilers/compilers.py
```diff
@@ -359,6 +360,11 @@ def get_base_compile_args(options: 'KeyedOptionDictType', compiler: 'Compiler')
...
```
anarazel, line 365: Given the general name (b_profile), shouldn't the function be named more generically as well?

unittests/allplatformstests.py
```diff
@@ -4092,6 +4092,18 @@ def test_devenv(self):
...
+        env = get_fake_env()
+        cc = detect_c_compiler(env, MachineChoice.HOST)
+        if cc.get_id() == 'clang' or is_windows() or is_cygwin():
+            raise SkipTest('-pg not supported by compiler')
+        testdir = os.path.join(self.unit_test_dir, '90 devenv')
+        self.init(testdir, extra_args=['-Db_profile=True'])
+        self.build()
+        self._run('./app', workdir = self.builddir)
+        gmon_out_path = os.path.join(self.builddir, 'gmon.out')
+        self.assertPathExists(gmon_out_path)
```
anarazel, line 4105: Perhaps also test the inverse, i.e. if profiling is disabled, gmon.out doesn't exist?

unittests/allplatformstests.py
```diff
@@ -4092,6 +4092,18 @@ def test_devenv(self):
         self.assertIn(f'TEST_C="{expected}"', o)
         self.assertIn('export TEST_C', o)
 
+    def test_create_gmon_out(self):
+        env = get_fake_env()
+        cc = detect_c_compiler(env, MachineChoice.HOST)
+        if cc.get_id() == 'clang' or is_windows() or is_cygwin():
```
anarazel, line 4098: clang supports -pg, no?

PR #10649, opened 2022-07-31
mcompile: provide user visibility into what actually happens

eli-schwartz:
This command is magical and I hate it. Mostly because it seems people have no clue what it does, and what it doesn't do.

Provide informational messages to the user indicating how it works, e.g. for debugging. Point out if we ran vsenv before shelling out to the backend.

### DIFF
--- a/mesonbuild/mcompile.py
+++ b/mesonbuild/mcompile.py
@@ -22,11 +22,12 @@ import shutil
 import typing as T
 from collections import defaultdict
 from pathlib import Path
+from time import sleep
 
 from . import mlog
 from . import mesonlib
 from . import coredata
-from .mesonlib import MesonException, RealPathAction, setup_vsenv
+from .mesonlib import MesonException, RealPathAction, join_args, setup_vsenv
 from mesonbuild.environment import detect_ninja
 from mesonbuild.coredata import UserArrayOption
 from mesonbuild import build
@@ -336,13 +337,16 @@ def run(options: 'argparse.Namespace') -> int:
 
     cdata = coredata.load(options.wd)
     b = build.load(options.wd)
-    setup_vsenv(b.need_vsenv)
+    vsenv_active = setup_vsenv(b.need_vsenv)
+    if vsenv_active:
+        mlog.log(mlog.green('INFO:'), 'automatically activated MSVC compiler environment')
 
     cmd = []    # type: T.List[str]
     env = None  # type: T.Optional[T.Dict[str, str]]
 
     backend = cdata.get_option(mesonlib.OptionKey('backend'))
     assert isinstance(backend, str)
+    mlog.log(mlog.green('INFO:'), 'autodetecting backend as', backend)
     if backend == 'ninja':
         cmd, env = get_parsed_args_ninja(options, bdir)
     elif backend.startswith('vs'):
@@ -353,6 +357,8 @@ def run(options: 'argparse.Namespace') -> int:
         raise MesonException(
             f'Backend `{backend}` is not yet supported by `compile`. Use generated project files directly instead.')
 
+    mlog.log(mlog.green('INFO:'), 'calculating backend command to run:', join_args(cmd))
+    sleep(2)
     p, *_ = mesonlib.Popen_safe(cmd, stdout=sys.stdout.buffer, stderr=sys.stderr.buffer, env=env)
 
     return p.returncode


dcbaker (approved):

mesonbuild/mcompile.py
```diff
@@ -353,6 +357,8 @@ def run(options: 'argparse.Namespace') -> int:
         raise MesonException(
             f'Backend `{backend}` is not yet supported by `compile`. Use generated project files directly instead.')
 
+    mlog.log(mlog.green('INFO:'), 'calculating backend command to run:', join_args(cmd))
+    sleep(2)
```
dnicolodi, line 361: Maybe I am not reading this right, but doesn't this this introduce a 2 seconds pause for each invocation of `meson compile`?

eli-schwartz: This is intentional in order to guarantee that the user actually sees the informational messages before they scroll off the screen.

The overhead of invoking python and importing the entirety of the meson codebase in order to access utility functions, argument parsing, vsenv handling, and more, is already significant. If you want to get down to the metal and eke out as much speed as you can, you were already running ninja directly -- ninja gives instant results, mcompile never did.

anarazel: This is a bad idea. Particularly when using the vs backend there's not really a good alternative to mcompile due to the odd project names meson generates (e.g. "22e3565@@postgres@exe.vcxproj"). And even if those get fixed, people won't be able to rely on it in docs etc for a few years. What's the point here, except to annoy people that use a feature that you're not a fan of?

eli-schwartz: To put this into perspective, on my current potato, to perform a no-op build.

running `meson compile` can take 31 seconds to complete on a cold cache, vs. 2 seconds to run ninja.

On a warm cache, running the same command twice in a row, for the second run, meson compile takes 1 second vs. ninja taking 0.005 seconds.

ninja utterly destroys meson compile in performance, because no-op builds are *actually* instant, they don't take an entire second to produce results. It's not especially unlikely that Meson gets evicted from the cache and starts taking longer, either.

What are your thoughts on immediately proceeding specifically during the vs backend due to the complexity of msbuild names decreasing the usefulness of knowing the command?

xclaesse:
I really don't see why we need that sleep at all, by default the backend is not verbose, so user has plenty of time to see those INFO message while it compiles. I'm not even sure why those info are even useful at all, but that does not hurt either, just remove the arbitrary sleep, IMHO.

eli-schwartz:
> by default the backend is not verbose, so user has plenty of time to see those INFO message while it compiles

The compilations themselves can be far, far more verbose than the backend, especially for C++. This actually makes it reasonably likely that a *single* build edge pushes messages off the screen due to warnings quite fast.

Also, not all backends are ninja. :p

mesonbuild/mcompile.py
anarazel: > ninja utterly destroys meson compile in performance, because no-op builds are actually instant, they don't take an entire second to produce results. It's not especially unlikely that Meson gets evicted from the cache and starts taking longer, either.

Sure. I agree that using ninja in hack-compile-test loops is better. But that's not really related to mcompile adding a 2s sleep.

> What are your thoughts on immediately proceeding specifically during the vs backend due to the complexity of msbuild names decreasing the usefulness of knowing the command?

Not favorable. When writing getting-started docs needing to branch on the used backend increases complexity unnecessarily. And people unnecessarily waiting for 2s longer just makes meson look bad.

anarazel:
> I'm not even sure why those info are even useful at all, but that does not hurt either, just remove the arbitrary sleep, IMHO.

I like that the command that gets invoked is printed, that makes it easier for the user to migrate to the "lower level" command. I don't care much about / mind the rest.

mesonbuild/mcompile.py
eli-schwartz: Well, I guess we're screwed, better quickly rewrite it in rust so people don't have to wait an eyeblink.

eli-schwartz:
> I like that the command that gets invoked is printed, that makes it easier for the user to migrate to the "lower level" command. I don't care much about / mind the rest.

They literally won't ever see it. I guess we can just revert this commit then.

anarazel:
> They literally won't ever see it. I guess we can just revert this commit then.

Why? The perf difference is irrelevant with msbuild, given how atrociously slow it is. With ninja it'll be a line away.

dnicolodi:
If running `meson compile` is such a bad interface and no one should be using it, why Meson own infrastructure uses it?
https://github.com/mesonbuild/wrapdb/blob/master/tools/sanity_checks.py#L292

eli-schwartz:
@dnicolodi

Is this supposed to be some sort of "haha I caught you out in a falsehood" power move?

Perhaps you'd like to take a look at commit 3c7ab542c0c4770241eae149b0d4cd8de329aee0 in which we deprecated something that not only is a bad interface no one should be using, but actively crippled development.

"Oh, but surely no one actually uses `meson <builddir>` without the setup subcommand".

Oh yes we do, and not just for internal stuff that one person who happens to like mcompile decided to use in a place no one routinely looks for best practices.

Nosirree, the official end user facing docs are full of it! 25b0988d4e40d22d907169b4f73ff07b4cb7fd0e We also use it for the tests. Also also, we use a command in build.ninja which hits this case: 2fe07adc0eee3920f825b536bf053ab3788cdaef

Should I take heed of the fact that it's used all over the place and decide "gosh, I was wrong after all, let's not deprecate implicit setup because we use it ourselves and therefore I'm proven wrong"? I think not. We cannot add new subcommands safely because of this implicit subcommand. Even mcompile was originally supposed to be called "meson build", but... we couldn't do that... because people were using that exact command with implicit setup and a `<builddir>` called "build".

Can we please not be facetious with the whys? Thanks.

dnicolodi:
@eli-schwartz I apologize, my comment definitely didn't deliver the intended message (and I don't want to sound like an armchair maintainer of someone else project).

I was trying to make the point that `meson compile` is there and it is a convenient interface when someone (mainly for being concise) does not want to bother with handling the different possible compilation backends explicitly, and because of this convenience is used in documentation and scripts. I don't question the decision of deprecating it (if there are more reasons to it than the fact that it is slower than invoking the backend directly). I question the decision of moving users away from it adding an arbitrary 2 seconds pause in its execution without adding a deprecation warning (where the point is really the delay without the deprecation warning, the warning alone would be best, but I would not be questioning the delay as much if it would come with a warning). By the way, I find the deprecation of `meson <builddir>` perfectly reasonable.

The reason put forward for adding the pause is to allow users to see the log messages and that the execution of the command is already slow, thus making it slower does not hurt much. However, on my system running `meson compile` adds just about half a second runtime to the direct invocation of ninja, thus on a no-op rebuild the added pause introduces a 4x slowdown, which is not very nice. The same command also produces just two lines of output, making the pause completely unnecessary.

It may require a creative phrasing of the log message, but what about moving the log message to after the compilation ended, to make sure it is visible? Something like:
```
INFO: compilation ended in {duration:.3f} seconds: {cmd}
```

mesonbuild/mcompile.py
xclaesse: > running meson compile can take 31 seconds to complete on a cold cache, vs. 2 seconds to run ninja.

Really? On my laptop no-op rebuild of the giant GStreamer project with all its subprojects takes 200ms with ninja, 700ms with `meson compile`. That's a significant increase but barely noticable in practice. I'm sure we can improve that, for example IIRC mcompile reads introspection json instead of pickle meson internal data. However now it takes 2.7s which is in my opinion unacceptable just for printing INFO that I fail to understand why it is even needed in the first place. In the rare case you care about the command it ran, just scroll up...

PR #10654, opened 2022-08-04
linkers: detection should invoke the linker with lang_link_args only

eli-schwartz:
Currently we invoke it with lang_args only, which is wrong and probably useless.

Fixes misdetecting the linker when overridden as -fuse-ld, which led to Meson trying to pass incompatible flags, outright failing if CFLAGS contained flags that only work with a non-default linker, or in the most benevolent case, having the status log report the wrong linker.

/cc @unhappy-ending @thesamesam


unhappy-ending:
I can confirm this fixes the issue. Thank you so much for your time and work on this!

PR #10613, opened 2022-07-18
cmake: Change assertion int a more useful error (fixes #9925)

mensinda:

PR #10656, opened 2022-08-04
cmake: Add rule relaxations for CMake subprojects (fixes #10613)

mensinda:
It ended up being more intrusive than I would have liked, but this was the only way to implement exceptions, without allowing projects to access files in the build directory.

This still won't address the issue of being able to address files in the build directory in general if the build directory is not located anywhere inside the meson project.

### DIFF
--- a/mesonbuild/interpreter/interpreter.py
+++ b/mesonbuild/interpreter/interpreter.py
@@ -92,6 +92,7 @@ from .type_checking import (
 from . import primitives as P_OBJ
 
 from pathlib import Path
+from enum import Enum
 import os
 import shutil
 import uuid
@@ -240,6 +241,16 @@ TEST_KWARGS: T.List[KwargInfo] = [
     KwargInfo('verbose', bool, default=False, since='0.62.0'),
 ]
 
+class InterpreterRuleRelaxation(Enum):
+    ''' Defines specific relaxations of the Meson rules.
+
+    This is intended to be used for automatically converted
+    projects (CMake subprojects, build system mixing) that
+    generate a Meson AST via introspection, etc.
+    '''
+
+    ALLOW_BUILD_DIR_FILE_REFFERENCES = 1
+
 permitted_dependency_kwargs = {
     'allow_fallback',
     'cmake_args',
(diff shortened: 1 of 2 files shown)


eli-schwartz:
So this allows cmake to access any build files, yes? Is the general idea here, something that could also be used to solve... I seem to recall previous issues with CMake subprojects running into sandbox violations if they tried to use pkg-config dependencies from another subproject? Does this help with that?

eli-schwartz:
The commit message is wrong, and tries to claim it's fixing the wrong bug.

Aside: I think it's better to have the "fixes #NNNN" on its own line.

mensinda:
> So this allows cmake to access any build files, yes?

Yes. I could narrow it down to only allow it to access build files from the current subproject, but I am not sure if this is worth it at this point.

> Is the general idea here, something that could also be used to solve...

I could either have added a boolean flag just for this case, or a more general solution (which is what I have chosen). I don't know of any other problem (currently) that might need another Meson rule relaxation, but I **really** want to avoid a scenario where another exemption is needed and the flags keep piling up.

Regarding the pkg-config thing, I think there were other problems, but it has been a long time and I should look into that again.

mensinda:
> The commit message is wrong, and tries to claim it's fixing the wrong bug.

Thanks, it is fixed now.

jpakkane:
> I could narrow it down to only allow it to access build files from the current subproject, but I am not sure if this is worth it at this point.

Most definitely yes. Any hole in the sandbox will be exploited by end users. We have seen that many, many, many times. In fact it would be better if the fix did not add new relaxation permissions but instead would verify absolute paths and allow them if they point to the current subproject build dir.

eli-schwartz:
@jpakkane

There's context here. In the linked bug report I already insisted we must not permit, in the general case, to use files in the build dir at all.

This PR therefore permits it only for CMakeLists.txt, which is basically wild west stuff. Our only goal is to try as best we can to merge the build edges "somehow", and not necessarily to enforce stuff like sandboxes.

It doesn't seem unreasonable to me to therefore say that the sandbox is out of scope for cmake subprojects. That's also the reason I wondered if it was useful to enable getting *-uninstalled.pc in a cmake subproject by adding the meson-uninstalled directory to cmake's pkg_config_path (because in the end we only get the full paths cmake writes out, not annotated objects like an Internal dependency).

tristan957:
Thanks for your work on this issue @mensinda. Very much appreciated.

mesonbuild/interpreter/interpreter.py
```diff
@@ -240,6 +241,16 @@ def dump_value(self, arr, list_sep, indent):
...
 
+class InterpreterRuleRelaxation(Enum):
+    ''' Defines specific relaxations of the Meson rules.
+
+    This is intended to be used for automatically converted
+    projects (CMake subprojects, build system mixing) that
+    generate a Meson AST via introspection, etc.
+    '''
+
+    ALLOW_BUILD_DIR_FILE_REFFERENCES = 1
```
FichteFoll, line 252: *RE~~F~~FERENCES*, probably?

PR #10668, opened 2022-08-09
add compilers from extracted objects directly to build targets

### LINKED ISSUE #10579: No host machine compiler for subproject linking to dependency in other subproject
**Describe the bug**
When using c-subproject in the dependencies of cpp-subproject, I get an error about no host machine compiler for the c language. The base project is also cpp.

```text
subprojects/animationwindow/meson.build:7:0: ERROR: No host machine compiler for 'subprojects/SDL2-2.0.20/src/SDL.c'
```

This **doesn't** happen if:

* I define `c` as one of the languages in the cpp-subproject
* I remove animationwindow, and have SDL as the only subproject

It does not however matter whether or not i define `c` as one of the languages in base project, neither if animationwindow is present or not

**To Reproduce**
structure:

├── subprojects
│   ├── cpp-subproject (animationwindow
│   └── c-subproject (sdl)
├── main.cpp
└── meson.build

E.g. SDL and animationwindow are parallell, not nested.

meson.build
```meson
project('testproject', 'cpp',
  version : '0.1',
  default_options : ['warning_level=2',
                     'cpp_std=c++17',
                     'c_std=c17',
                     'default_library=static'])

sdl2_dep = dependency('sdl2')
animationwindow_dep = dependency('animationwindow', fallback: ['animationwindow', 'animationwindow_dep'])

src = ['main.cpp']

exe = executable('program', src, dependencies : [animationwindow_dep, sdl2_dep])
```

subprojects/animationwindow.cpp
```meson
project('animationwindow', 'cpp', default_options: ['cpp_std=c++17', 'default_library=static', 'buildtype=debugoptimized'])

sdl2_dep = dependency('sdl2')

build_files = ['src/notfinished.cpp']
incdir = include_directories('include')
animationwindow = static_library('animationwindow', build_files, include_directories: incdir, dependencies: [sdl2_dep], install: true)
install_subdir('include', install_dir: '.')
install_subdir('src', install_dir: '.')

animationwindow_dep = declare_dependency(link_with: animationwindow, include_directories: incdir)
```

subprojects/sdl2.wrap
```ini
[wrap-file]
directory = SDL2-2.0.20
source_url = https://libsdl.org/release/SDL2-2.0.20.tar.gz
source_filename = SDL2-2.0.20.tar.gz
source_hash = c56aba1d7b5b0e7e999e4a7698c70b63a3394ff9704b5f6e1c57e0c16f04dd06
patch_filename = sdl2_2.0.20-5_patch.zip
patch_url = https://wrapdb.mesonbuild.com/v2/sdl2_2.0.20-5/get_patch
patch_hash = b0a3700820b2553ae228963f5a6168525404e65d579528957e872c19235deb22
wrapdb_version = 2.0.20-5

[provide]
sdl2 = sdl2_dep
```

**Expected behavior**
Expected to `meson setup debug` without any issues.

**system parameters**
* This is a native build
* on a m1 Mac running macos monterey 12.4
* meson 0.63.0 from homebrew
* python 3.8.9

eli-schwartz:
In order to reliably link to static libraries or individual object files, we need to take their languages into account as well. For static
libraries this is easy: we just add the static library's list of compilers to the build target. For extracted objects, we need to only add the ones for the objects we use.

But we did this really inefficiently -- in fact, downright terribly. We iterated over all source files from the extracted objects, then tried to look up a new compiler for them. Even though the extracted objects already had a list of compilers! This broke once compilers were made per-subproject, because while the extracted objects have a reference to all the compilers it needs (just like static archives do, actually) we might not actually be able to look up that compiler from scratch inside the current subproject.

Fix this by asking the extracted objects to categorize all its own sources and return the compilers we want.

Fixes #10579


mesonbuild/build.py
```diff
@@ -950,12 +950,15 @@ def process_compilers(self) -> T.List[str]:
         for o in self.objects:
             if not isinstance(o, ExtractedObjects):
                 continue
-            for s in o.srclist:
+            compsrcs = o.classify_all_sources(o.srclist, [])
+            for comp in compsrcs.keys():
```
dcbaker, line 954: Don’t use .keys() here

mesonbuild/build.py
```diff
@@ -950,12 +950,15 @@ def process_compilers(self) -> T.List[str]:
...
                 # Don't add Vala sources since that will pull in the Vala
                 # compiler even though we will never use it since we are
                 # dealing with compiled C code.
-                if not s.endswith(lang_suffixes['vala']):
-                    sources.append(s)
+                if comp.language == 'vala':
```
dcbaker, line 958: Shouldn’t we also check for cython?

dcbaker (changes requested):
Sorry it took me so long to get to looking at this. One small nit, and one question, but otherwise looks good

PR #10670, opened 2022-08-09
pkg-config: support multiple variables in `pkgconfig_define` kwarg

jtojnar:
The `pkgconfig_define` kwarg of [[dep.get_variable]] and `define_variable` kwarg of [[dep.get_pkgconfig_variable]] now accept multiple variables, consistent with pkg-config’s handling of `--define-variable` flag. The latter errored out but the former produced junk argument, breaking e.g. bubblewrap build.


jtojnar:
Verified that this fixes Nix build of bubblewrap.

mesonbuild/interpreter/interpreterobjects.py
```diff
@@ -448,7 +448,6 @@ def name_method(self, args: T.List[TYPE_var], kwargs: TYPE_kwargs) -> str:
             ContainerTypeInfo(list, str, pairs=True),
             default=[],
             listify=True,
-            validator=lambda x: 'must be of length 2 or empty' if len(x) not in {0, 2} else None,
         ),
     )
     def pkgconfig_method(self, args: T.Tuple[str], kwargs: 'kwargs.DependencyPkgConfigVar') -> str:
```
dcbaker, line 453:
```suggestion
    def pkgconfig_method(self, args: T.Tuple[str], kwargs: 'kwargs.DependencyPkgConfigVar') -> str:
        if len(kwargs['define_variable'] > 2):
            FeatureNew.single_use('multiple pkg-config define overrides', '0.64', self.subproject, location=self.current_node)
```

And also in the `get_variable()` method

dcbaker (changes requested):
There's two places we need a `FeatureNew` call, otherwise this looks good to me. Thanks!

eli-schwartz:
Maybe we should just require it to be a dict instead, if defining multiple replacements?

dcbaker:
I'd like to move to a dict, but we already have projects in the wild defining multipe values, so we'll need to use a deprecation cycle

eli-schwartz:
I thought defining multiple values produces configure-time errors due to passing invalid arguments to pkg-config? That's literally why we are adding a FeatureNew here?

jtojnar:
From the point of pkg-config, the arguments are valid:

```
$ pkg-config gio-2.0 --variable datadir --define-variable=prefix=/=datadir=/usr/share
/=datadir=/usr/share/share
```

I think the `FeatureNew` is just to add a proper warning that people should either raise their `meson_version` or stop using multiple args, if they want to keep supporting old Meson versions.

eli-schwartz:
So again, if anyone did it, it was buggy and led to incorrect build results, although it might not actually lead to a configuration failure.

If we simply verify that exactly two values are used, it is ***correct*** behavior otherwise known as "sanity checks", and I would consider that to be a bugfix (but not a candidate for backporting) and would not require a deprecation cycle.

By the same token, I would not require a deprecation cycle if we first add that sanity check, then add dictionary support. But I don't feel strongly about it.

jtojnar:
> So again, if anyone did it, it was buggy and led to incorrect build results, although it might not actually lead to a configuration failure.

Right.
 
> If we simply verify that exactly two values are used, it is _**correct**_ behavior otherwise known as "sanity checks", and I would consider that to be a bugfix (but not a candidate for backporting) and would not require a deprecation cycle.

I agree that what you write would make sense if we were introducing the verification that exactly two values are used, and we should not need to formally deprecate passing more arguments – we could just forbid it straight away.

But we are doing the opposite (relaxing the two items limit) so there is no need to deprecate anything.

Though, at least for `dep.get_pkgconfig_variable(define_variable)`, multiple arguments were previously a configure failure so we do in fact need to introduce a `FeatureNew` there to avoid developers using Meson 0.64 accidentally breaking projects targetting older versions.

> By the same token, I would not require a deprecation cycle if we first add that sanity check, then add dictionary support. But I don't feel strongly about it.

Again, I think `FeatureNew` is in order here.

dcbaker (approved):

Issue #10641, opened 2022-07-29
sincos, sincosf detection as used by gtk misdetects sincos, sincosf


0-wiz-0:
**Describe the bug**
gtk uses meson to detect sincos and sincosf.  NetBSD does not provide these functions.
gtk first tries to detect them without any libraries, this fails as expected; then it tries using libm and succeeds. However, the second test is broken.

**To Reproduce**
Here's the code that gtk uses to detect sincos and sincosf:
https://gitlab.gnome.org/GNOME/gtk/-/blob/4.6.6/meson.build#L188-208

or to quote it directly:
```
check_functions = [
  'dcgettext',
  'getpagesize',
  'getresuid',
  'lstat',
  'memfd_create',
  'mmap',
  'posix_fallocate',
  '_lock_file',
  'flockfile',
  'mkstemp',
  'mallinfo2',
  'sincos',
  'sincosf',
]

foreach func : check_functions
  if cc.has_function(func, dependencies: libm)
    cdata.set('HAVE_' + func.underscorify().to_upper(), 1)
  endif
endforeach
```

**Expected behavior**
sincos and sincosf should not be found, but they are.

I looked in `output/meson-logs/meson-log.txt` to check what happens and found that it first checks for `sincos()` and fails:

```
[...]

Issue #4501, opened 2018-11-09
RFC: Do not check for builtins in has_function, add has_builtin


ePirat:
Checking for builtins in `has_function` has causes quite some trouble and is still broken in meson. IMO there is no easy way to get both right in one kwarg and it would be much cleaner and sensible to do not check at all for builtins in `has_function` and instead introduce a kwarg `has_builtin` to check for compiler builtins.

dcbaker:
I don't have a problem with adding a `has_builtin` method, but we'll need a backwards compatibility story. I actually like this idea because as long as the compiler doesn't change we should be able to cache whether we have builtins.

PR #10674, opened 2022-08-10
Also test linking when checking for a builtin function

lb90:
A builtin function may resolve to a function call, but it's not mandatory
for a libc implementation to provide such function. When testing for the
presence of a builtin function, also check if linking succeeds or not.

Fixes https://github.com/mesonbuild/meson/issues/10641


mesonbuild/compilers/mixins/clike.py
```diff
@@ -863,6 +863,19 @@ def has_function(self, funcname: str, prefix: str, env: 'Environment', *,
         #ifdef __has_builtin
             #if !__has_builtin({__builtin_}{func})
                 #error "{__builtin_}{func} not found"
+            #else
+                /* Builtins, depending on the target architecture, may be
+                 * substitued inline by the compiler, or may just produce
```
eli-schwartz, line 868:
```suggestion
                 * substituted inline by the compiler, or may just produce
```

mesonbuild/compilers/mixins/clike.py
```diff
@@ -863,6 +863,19 @@ def has_function(self, funcname: str, prefix: str, env: 'Environment', *,
...
+                 * a call into the c standard library. Some libc may not
+                 * have an implementation, though. As an example, sincos()
+                 * is a GCC built-in, but the NetBSD libc doesn't provide
+                 * sincos (). Here we check if linking actually works. /*
```
eli-schwartz, line 872:
```suggestion
                 * sincos (). Here we check if linking actually works. */
```

eli-schwartz (commented):
In addition to the CI failures that indicate this logic is flawed, there are a couple of typos.

lb90:
This is tricky to get right. I made a test using autotools:

```M4Sugar
AC_INIT([testcheck], [1.0])
AM_INIT_AUTOMAKE([foreign])

AC_CHECK_LIB(m, sincos)
AC_CHECK_FUNC(alloca)

AC_PROG_CC
AC_CONFIG_FILES([Makefile])
AC_OUTPUT
```

```make
bin_PROGRAMS=testcheck
```

```c
int main()
{
  return 0;
}
```

Running

```bash
autoreconf -fiv
./configure
```

Gives:

```
configure:4209: result: gcc3
configure:4225: checking for sincos in -lm
configure:4248: gcc -o conftest -g -O2   conftest.c -lm   >&5
conftest.c:15:6: warning: conflicting types for built-in function 'sincos'; expected 'void(double,  double *, double *)' [-Wbuiltin-declaration-mismatch]
   15 | char sincos ();
      |      ^~~~~~
conftest.c:1:1: note: 'sincos' is declared in header '<math.h>'
    1 | /* confdefs.h */
configure:4248: $? = 0
configure:4258: result: yes
configure:4269: checking for alloca
configure:4269: gcc -o conftest -g -O2   conftest.c -lm  >&5
conftest.c:28:6: warning: conflicting types for built-in function 'alloca'; expected 'void *(long unsigned int)' [-Wbuiltin-declaration-mismatch]
   28 | char alloca ();
      |      ^~~~~~
/usr/bin/ld: /tmp/cck1SkYn.o: in function `main':
/home/luca/test-autoconf/conftest.c:39: undefined reference to `alloca'
configure:4269: $? = 1
configure: failed program was:
| /* confdefs.h */
| #define PACKAGE_NAME "testcheck"
| #define PACKAGE_TARNAME "testcheck"
| #define PACKAGE_VERSION "1.0"
| #define PACKAGE_STRING "testcheck 1.0"
| #define PACKAGE_BUGREPORT ""
| #define PACKAGE_URL ""
| #define PACKAGE "testcheck"
| #define VERSION "1.0"
| #define HAVE_LIBM 1
| /* end confdefs.h.  */
| /* Define alloca to an innocuous variant, in case <limits.h> declares alloca.
|    For example, HP-UX 11i <limits.h> declares gettimeofday.  */
| #define alloca innocuous_alloca
| 
| /* System header to define __stub macros and hopefully few prototypes,
|    which can conflict with char alloca (); below.  */
| 
| #include <limits.h>
| #undef alloca
| 
| /* Override any GCC internal prototype to avoid an error.
|    Use char because int might match the return type of a GCC
|    builtin and then its argument prototype would still apply.  */
| #ifdef __cplusplus
| extern "C"
| #endif
| char alloca ();
| /* The GNU C library defines this for functions which it implements
|     to always fail with ENOSYS.  Some functions are actually named
|     something starting with __ and the normal name is an alias.  */
| #if defined __stub_alloca || defined __stub___alloca
| choke me
| #endif
| 
| int
| main (void)
| {
| return alloca ();
|   ;
|   return 0;
| }
configure:4269: result: no
```

lb90:
To know if a built-in may not be inlined, I guess we can test for

```c
__has_builtin (func) && __has_builtin (__builtin_func)
```

matthiasclasen:
All this seems much to complicated to me. What we want to know is:

1) Will a program that calls sincos(x) link ?
2) If not, will it link if we add -lm to LIBS ?

I don't see why there is any need to fiddle with builtins here.

lb90:
Yeah, see also:

* https://github.com/mesonbuild/meson/issues/7113#issuecomment-626619683
* https://github.com/mesonbuild/meson/issues/4501

Now that compilers have `__has_builtin` there's even less need to do checks from the build system

lb90:
@lazka could you let us know more about the non-functioning builtins in mingw GCC?

lazka:
> @lazka could you let us know more about the non-functioning builtins in mingw GCC?

that's all I know: https://github.com/mesonbuild/meson/pull/7116/files#r424014885

lb90:
Anyway consider compiling a math library using sincos(), sqrt() etc. on netBSD. If the architecture has machine instruction for sincos, we don't actually care that the netBSD libc doesn't implement sincos(), because GCC uses the machine instruction anyway.

matthiasclasen:
> Anyway consider compiling a math library using sincos(), sqrt() etc. on netBSD. If the architecture has machine instruction for sincos, we don't actually care that the netBSD libc doesn't implement sincos(), because GCC uses the machine instruction anyway. So yes, I'd say the function should be reported as available. Avoiding that could hurt performance considerably.

But if the compiler doesn't use a library function, then compiling a trivial test program using sincos(x) will succeed, no? Why do we need to fiddle with builtins?

lb90:
> then compiling a trivial test program using sincos(x) will succeed, no?

Unfortunately no, it will fail with GCC (not so sure about CLang)

A build system writes a simple test program to check for availability of a function. It doesn't know how it should call the function, though, because it doesn't have any knowledge of the argument list. Thus the function is prototyped in a generic way: `char func();`. This is how a test program looks like:

```c
/* generic prototype */
char func();

int main()
{
  return func();
}
```

If that compiles and links, the function is available.

Now consider what happens when `func` is a builtin. The compiler encounters an incompatible prototype of one of its builtin functions. What should it do? That's not standardized, because builtins are not part of the C or C++ language specs. GCC, probably for compatibility with existing codebases, will consider sincos a normal function, not a builtin. Thus it will try to resolve func at link time, and will fail on netBSD.

It can be tested on Linux:

* This compiles with `gcc test-sincos.c`
   ```c
   void sincos(double x, double *s, double *c);
   
   int main()
   {
     double s, c;
     sincos(3.14 / 4, &s, &c);
   
     return (int)((s*s + c*c)*1000);
   }
   ```
* This requires linking with the math library: `gcc test-sincos.c -lm`
   ```c
   char sincos();
   
   int main()
   {
     return sincos();
   }
   ```

Another example is:

```c
/* remove me */
char __builtin_trap();

int main()
{
  __builtin_trap();
  return 0;
}
```

Finally, another issue that might get in the way is explained in https://gcc.gnu.org/pipermail/gcc/2022-August/239262.html

matthiasclasen:
I would be quite ok if the check determines that sincos is not available on netbsd.

matthiasclasen:
I would say the 'generic prototype' is part of the problem here.

Maybe something like this would work  better:

```#define _GNU_SOURCE

#include <math.h>

typedef char (* func_ptr) (void);

int main ()
{
  func_ptr f;

  f = (func_ptr) sincos;

  return 0;
}
```

lb90:
Here's something very intersting:

`__has_builtin ()` in clang evaluates to 1 only for builtins that are _always_ inlined. Consider the following sample:

```c
#if __has_builtin(sincos)
#error "sincos is a builtin"
#else
#error "sincos is NOT a builtin"
#endif

int main()
{
  return 0;
}
```

Here's the outcome of compilation:

```
[luca@ArchLinux ~]$ gcc test-has-builtin.c 
test-has-builtin.c:2:2: error: #error "sincos IS a builtin"
    2 | #error "sincos IS a builtin"
      |  ^~~~~
[luca@ArchLinux ~]$ clang test-has-builtin.c 
test-has-builtin.c:4:2: error: "sincos NOT a builtin"
#error "sincos NOT a builtin"
 ^
1 error generated.
```

And the same happens for `__builtin_sincos`.

lb90:
IMO the behavior of clang makes much more sense. Also, `__has_builtin` was introduced in gcc as a consequence of its addition in clang, so it makes sense to stay compatible

lb90:
See https://github.com/llvm/llvm-project/commit/cefe472c51fbcd1aed4d4a090709f25a12a8bc2c, https://reviews.llvm.org/D125829

